PR #1019, opened 2015-12-04
Install and start one etcd server before the others.

dgoodwin:
This might need some discussion but I believe we have a bug with etcd cluster setup in faster environments, all the cluster members appear to come up at roughly the same time when using all local vms (within a second or two) and one or two of my three will fail to make it to the cluster. 

When this bug happens a failed member's etcd logs will look like:

http://fpaste.org/296258/89844131/

Whereas one of the member's that made it up will look like: 

http://fpaste.org/296239/14489825/

Logging into those members and removing the etcd data dir and restarting the service will get that member to cluster up.

Adding members one at a time appears to be recommended best practice at least for cluster recovery, which is essentially the same thing: https://github.com/coreos/etcd/blob/master/Documentation/runtime-configuration.md#restart-cluster-from-majority-failure

Serializing the etcd setup to one server at a time appears to correct this problem, the cluster comes up healthy.

I suspect this hasn't surfaced in cloud environments just due to the delays in communicating with them and the nodes have time to come up individually.

### DIFF
--- a/playbooks/common/openshift-etcd/config.yml
+++ b/playbooks/common/openshift-etcd/config.yml
@@ -68,8 +68,32 @@
       validate_checksum: yes
     with_items: etcd_needing_server_certs
 
-- name: Configure etcd hosts
-  hosts: oo_etcd_to_config
+# Configure a first etcd host to avoid conflicts in choosing a leader
+# if other members come online too quickly.
+- name: Configure first etcd host
+  hosts: oo_first_etcd
+  vars:
+    sync_tmpdir: "{{ hostvars.localhost.g_etcd_mktemp.stdout }}"
+    etcd_url_scheme: https
+    etcd_peer_url_scheme: https
+    etcd_peers_group: oo_etcd_to_config
+  pre_tasks:
+  - name: Ensure certificate directory exists
+    file:
+      path: "{{ etcd_cert_config_dir }}"
+      state: directory
+  - name: Unarchive the tarball on the etcd host
+    unarchive:
+      src: "{{ sync_tmpdir }}/{{ etcd_cert_subdir }}.tgz"
+      dest: "{{ etcd_cert_config_dir }}"
+    when: etcd_server_certs_missing
+  roles:
+  - etcd
+  - role: nickhammond.logrotate
+
+# Configure the remaining etcd hosts, skipping the first one we dealt with above.
+- name: Configure remaining etcd hosts
+  hosts: oo_etcd_to_config:!oo_first_etcd
   vars:
     sync_tmpdir: "{{ hostvars.localhost.g_etcd_mktemp.stdout }}"
     etcd_url_scheme: https


detiber:
@dgoodwin would it make more sense to do the following steps:
- configure and start oo_first_etcd
- configure and start remaining etcd hosts? oo_etcd_to_config should be sufficient here, since running against oo_first_etcd should be a noop at that point.

It seems that the error is related to the one host thinking it was the cluster leader without any peers, while the other two hosts agreed on a different leader.

detiber:
@dgoodwin to be clear my only concern is fully serializing this to avoid increasing ansible runtimes too much.

dgoodwin:
In testing the downgrade / restore procedure it appeared important that each host was brought up one at a time, I.E. I couldn't add 2 members on the first running member, then go start each one up, it had to be add one and start it, then the other. The linked docs have a little note indicating this as well but it's for the backup restore case, though I don't see why it would be different.

Based on the fact that this has been working in other environments I would guess that you are correct and we can probably just get a first member up and then let all the others come in. Can give it a shot anyhow. Closing for now, will re-open if this approach seems the safest. I will test this other theory now as I need a fix to proceed with the restart work.

dgoodwin:
Ok doing one etcd host first and then the others in parallel does seem to solve the problem.  PR updated.

detiber:
Does defining the hosts for the configure play to oo_first_etcd:oo_etcd_to_config do the same thing? I'd just like to avoid duplication if possible.

dgoodwin:
Unfortunately seems to re-introduce the problem. Is it worth refactoring it out into a task or is that too heavy handed?

dgoodwin:
Per @sdodson 's request to get the first etcd master out of the list on the second pass, PR has been updated.

I tried a variety of options including "hosts: oo_etcd_to_config[1:]" (which matched no hosts) as well as "hosts: oo_etcd_to_config:-oo_first_etcd" (which matched all hosts again). Defining a group does work however, does this look ok?

sdodson:
LGTM

detiber:
The host pattern you want is most likely:

```
hosts: oo_etcd_to_config:!oo_first_etcd
```

proceeding a host pattern with a '-' indicates a regex: http://docs.ansible.com/ansible/intro_patterns.html

dgoodwin:
That worked, thanks!

PR updated.

abutcher:
aos-ci-test

detiber:
@twiest @wshearn please review

twiest:
@dgoodwin Sorry for the delay on this, this looks good to me.

@detiber is this still something we want merged?

detiber:
aos-ci-test

detiber:
@twiest let me do a validation run first.

detiber:
aos-ci-test

dgoodwin:
Yes please @twiest @detiber, thanks!

detiber:
@brenton passed my smoke testing and the aos tests.

PR #1020, opened 2015-12-04
Default 3.0.x dnsIP

abutcher:
QE found an issue with the default getting set when `dnsIP` is missing in 3.0.x so it should be set to the first master ip. I've also removed the original variables from scale up.

https://bugzilla.redhat.com/show_bug.cgi?id=1246458

### DIFF
--- a/playbooks/common/openshift-cluster/scaleup.yml
+++ b/playbooks/common/openshift-cluster/scaleup.yml
@@ -3,6 +3,4 @@
 
 - include: ../openshift-node/config.yml
   vars:
-    osn_cluster_dns_domain: "{{ hostvars[groups.oo_first_master.0].openshift.dns.domain }}"
-    osn_cluster_dns_ip: "{{ hostvars[groups.oo_first_master.0].openshift.dns.ip }}"
     openshift_deployment_type: "{{ deployment_type }}"
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -158,8 +158,10 @@
   vars:
     sync_tmpdir: "{{ hostvars.localhost.mktemp.stdout }}"
     openshift_node_master_api_url: "{{ hostvars[groups.oo_first_master.0].openshift.master.api_url }}"
+    # TODO: Prefix flannel role variables.
     etcd_urls: "{{ hostvars[groups.oo_first_master.0].openshift.master.etcd_urls }}"
     embedded_etcd: "{{ hostvars[groups.oo_first_master.0].openshift.master.embedded_etcd }}"
+    openshift_node_first_master_ip: "{{ hostvars[groups.oo_first_master.0].openshift.common.ip }}"
   pre_tasks:
   - name: Ensure certificate directory exists
     file:
--- a/roles/openshift_node/tasks/main.yml
+++ b/roles/openshift_node/tasks/main.yml
@@ -14,7 +14,10 @@
       hostname: "{{ openshift_hostname | default(none) }}"
       public_hostname: "{{ openshift_public_hostname | default(none) }}"
       deployment_type: "{{ openshift_deployment_type }}"
-      dns_ip: "{{ openshift_dns_ip | default(openshift_master_cluster_vip | default(None, true), true) }}"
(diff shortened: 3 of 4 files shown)


playbooks/common/openshift-node/config.yml
```diff
@@ -160,6 +160,7 @@
     openshift_node_master_api_url: "{{ hostvars[groups.oo_first_master.0].openshift.master.api_url }}"
     etcd_urls: "{{ hostvars[groups.oo_first_master.0].openshift.master.etcd_urls }}"
     embedded_etcd: "{{ hostvars[groups.oo_first_master.0].openshift.master.embedded_etcd }}"
+    first_master_ip: "{{ hostvars[groups.oo_first_master.0].openshift.common.ip }}"
```
detiber: we'll want to prefix this var for the openshift_node role...  old style was to use openshift_node_<var_name>, newer style started using osn_<var_name>...  It seems that openshift_node_<var_name> is more consistent with galaxy recommendations though...

On an aside, I think variables above (embedded_etcd, etcd_urls) is used when configuring flannel on nodes, and should probably have a prefix as well.

roles/openshift_node/tasks/main.yml
```diff
@@ -14,7 +14,9 @@
       hostname: "{{ openshift_hostname | default(none) }}"
       public_hostname: "{{ openshift_public_hostname | default(none) }}"
       deployment_type: "{{ openshift_deployment_type }}"
-      dns_ip: "{{ openshift_dns_ip | default(openshift_master_cluster_vip | default(None, true), true) }}"
+      dns_ip: "{{ openshift_dns_ip
+                  | default(openshift_master_cluster_vip
+                  | default(None if openshift.common.version_greater_than_3_1_or_1_1 | bool else first_master_ip | default(None, true), true), true) }}"
```
detiber: At this point we are starting to approach lookup_plugin or filter_plugin territory...  either that or pushing the logic into openshift_facts.

detiber:
@abutcher variable scoping is covered a bit in the style guide: https://github.com/openshift/openshift-ansible/blob/master/docs/style_guide.adoc

The exception here is that I would say it is okay to use <role_name>_<var_name> in addition to the short prefixes mentioned there.

Also the style guide doesn't cover temporary variables used in playbooks, which I would suggest should be tmp_<playbook_id>_<var_name> with playbook_id being a short playbook descriptor based on where the playbook is located...  something like c_oc_config could be used for playbooks/common/openshift-cluster/config.yml.

roles/openshift_node/tasks/main.yml
abutcher: Hmmm, what about an `oo_pick...(var1, var2)` that grabs the first thing defined. The default result could be passed in or `None`.

playbooks/common/openshift-node/config.yml
abutcher: I'll rename all of them while I'm here.

roles/openshift_node/tasks/main.yml
detiber: @abutcher you sir are a genius :) Even just returning None on no items defined and allowing the use of the default filter afterwards would be acceptable.

I would like to see it accept the jinja2 'dot' notation for variable names though.

playbooks/common/openshift-cluster/scaleup.yml
```diff
@@ -3,6 +3,4 @@
 
 - include: ../openshift-node/config.yml
   vars:
-    osn_cluster_dns_domain: "{{ hostvars[groups.oo_first_master.0].openshift.dns.domain }}"
-    osn_cluster_dns_ip: "{{ hostvars[groups.oo_first_master.0].openshift.dns.ip }}"
     openshift_deployment_type: "{{ deployment_type }}"
```
sdodson: These two show up in another file, should we remove them from there too?

https://github.com/openshift/openshift-ansible/blob/master/playbooks/gce/openshift-cluster/join_node.yml#L48-L49

abutcher: @sdodson Yepppp good catch

sdodson:
@brenton LGTM

Issue #1002, opened 2015-11-30
Making Examples Optional


jtslear:
Good Morning,  would it be worthwhile to provide the capability to optionally disable the loading of the openshift_examples?  In our environment, we aren't using them and are rather useless.  I'm merely trying to remove a bit of cruft from our overall installation.

sdodson:
@jtslear For now, you can stop them from being installed into the namespace by setting the following, however I agree we should add a variable that disables the entire role.

```
openshift_examples_load_db_templates: false
openshift_examples_load_xpaas: false
openshift_examples_load_quickstarts: false
```

If you also want to disable the image streams
```
openshift_examples_load_rhel: false
openshift_examples_load_centos: false
```

abutcher:
@brenton This one can be closed. ( Or @jtslear can close it if he gets here first :smiley_cat: )

PR #1030, opened 2015-12-06
Make the install of openshift_examples optional

jtslear:
* Allows us to make it optional to run the role `openshift_examples`
* Defaults to true
* I would definitely recommend testing to ensure this goes down the path you guys would normally for something like this
* Addresses #1002

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -349,7 +349,8 @@
   roles:
   - role: openshift_master_cluster
     when: openshift_master_ha | bool and openshift.master.cluster_method == "pacemaker"
-  - openshift_examples
+  - role: openshift_examples
+    when: openshift.common.install_examples | bool
   - role: openshift_cluster_metrics
     when: openshift.common.use_cluster_metrics | bool
   - role: openshift_manageiq
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1057,6 +1057,7 @@ class OpenShiftFacts(object):
         common['client_binary'] = 'oc' if os.path.isfile('/usr/bin/oc') else 'osc'
         common['admin_binary'] = 'oadm' if os.path.isfile('/usr/bin/oadm') else 'osadm'
         common['dns_domain'] = 'cluster.local'
+        common['install_examples'] = True
         defaults['common'] = common
 
         if 'master' in roles:
--- a/inventory/byo/hosts.aep.example
+++ b/inventory/byo/hosts.aep.example
@@ -21,6 +21,9 @@ ansible_ssh_user=root
 # deployment type valid values are origin, online, atomic-enterprise, and openshift-enterprise
 deployment_type=atomic-enterprise
 
+# Install the openshift examples
+#openshift_install_examples=true
+
 # Enable cluster metrics
 #use_cluster_metrics=true
 
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -21,6 +21,9 @@ ansible_ssh_user=root
 # deployment type valid values are origin, online, atomic-enterprise and openshift-enterprise
 deployment_type=origin
 
+# Install the openshift examples
+#openshift_install_examples=true
+
 # Enable cluster metrics
 #use_cluster_metrics=true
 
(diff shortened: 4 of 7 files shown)


roles/openshift_facts/library/openshift_facts.py
```diff
@@ -381,6 +381,23 @@ def set_master_selectors(facts):
...
+            facts (dict): existing facts
+        Returns:
+            dict: the facts dict updated with the generated openshift_examples
+            facts if they were not already present
+
+    """
+    if 'common' in facts:
+        if 'install_openshift_examples' not in facts['common']:
+            install_openshift_examples = True
+            facts['common']['install_openshift_examples'] = install_openshift_examples
```
detiber: I'd combine lines 397 and 398:
```
facts['common']['install_openshift_examples'] = True
```

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1024,6 +1041,7 @@ def generate_facts(self, local_facts, additive_facts_to_overwrite):
         facts = set_flannel_facts_if_unset(facts)
         facts = set_node_schedulability(facts)
         facts = set_master_selectors(facts)
+        facts = set_openshift_examples_facts_if_unset(facts)
```
detiber: I don't think there is a need to set the default through a method, that is really only needed for values that don't have a simple default.  Instead, you should be able to set the default in get_defaults() instead.

playbooks/common/openshift-master/config.yml
```diff
@@ -349,7 +349,8 @@
   roles:
   - role: openshift_master_cluster
     when: openshift_master_ha | bool and openshift.master.cluster_method == "pacemaker"
-  - openshift_examples
+  - role: openshift_examples
+    when: openshift.common.install_openshift_examples | bool
```
detiber: I don't see where the value provided on the inventory is persisted.  To persist the user provided value, roles/openshift_common/tasks/main.yml will need to be updated to set the value through the local_facts parameter to openshift_facts.

inventory/byo/hosts.example
```diff
@@ -21,6 +21,9 @@ ansible_ssh_user=root
 # deployment type valid values are origin, online and enterprise
 deployment_type=atomic-enterprise
 
+# Install the openshift examples
+#install_openshift_examples=true
```
detiber: This should probably be named openshift_install_examples.

I would also rename the fact to openshift.common.install_examples

detiber:
@jtslear thanks for the PR, it'll be nice to close out the associated issue for this.

@sdodson do we want/need to treat the base imagestreams separately from the rest of the examples?

I just wonder if there is a use case to not install the base imagestream (outside of Atomic Enterprise Platform)

jtslear:
@detiber I think I've addressed everything you've noted in the latest update.

jtslear:
I was originally using the `use_cluster_metrics` as a template towards implementing this.  I can see how it's been greatly simplified now.  Though, I would recommend it go through a gauntlet of testing.

playbooks/common/openshift-master/config.yml
```diff
@@ -349,7 +349,8 @@
   roles:
   - role: openshift_master_cluster
     when: openshift_master_ha | bool and openshift.master.cluster_method == "pacemaker"
-  - openshift_examples
+  - role: openshift_examples
+    when: openshift.common.openshift_install_examples | bool
   - role: openshift_cluster_metrics
```
detiber: This should be openshift.common.install_examples since you are setting install_examples as a local_fact for the common role.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1057,6 +1057,7 @@ def get_defaults(self, roles):
         common['client_binary'] = 'oc' if os.path.isfile('/usr/bin/oc') else 'osc'
         common['admin_binary'] = 'oadm' if os.path.isfile('/usr/bin/oadm') else 'osadm'
         common['dns_domain'] = 'cluster.local'
+        common['openshift_install_examples'] = True
```
detiber: common['install_examples']

detiber:
@jtslear this is looking really good.

@abutcher can you give this a smoke test after the variable naming is updated?

abutcher:
Certainly

jtslear:
Updated, please enjoy.

abutcher:
@brenton Good to merge

brenton:
@sdodson, is this similar to the approach you wanted to take for AEP?

sdodson:
This is the PR that spurred the discussion about what we should be
installing in AEP. Right now this just provides the admin one variable to
completely disable examples. I'll put a card in for cleaning up what we
install into AEP by default.

+1 from me to this PR

PR #1032, opened 2015-12-07
Require pyOpenSSL

sdodson:
required by oo_filters

Fixes
```
import OpenSSL.crypto
No module named OpenSSL.crypto
```

### DIFF
--- a/openshift-ansible.spec
+++ b/openshift-ansible.spec
@@ -211,6 +211,7 @@ BuildArch:     noarch
 Summary:       Openshift and Atomic Enterprise Ansible filter plugins
 Requires:      %{name}
 BuildArch:     noarch
+Requires:      pyOpenSSL
 
 %description filter-plugins
 %{summary}.


sdodson:
@thoraxe FYI

@detiber PTAL

detiber:
:+1:

sdodson:
Hmm, should we also ensure that it's installed on the remote host via openshift_facts? I don't know if the filters are run on the ansible host or the target host.

abutcher:
@sdodson They run on the ansible host.

Similar to netaddr requirement for the ipaddr filter http://docs.ansible.com/ansible/playbooks_filters_ipaddr.html.

"To use this filter in Ansible, you need to install netaddr Python library on a computer on which you use Ansible (it is not required on remote hosts). It can usually be installed either via your system package manager, or using pip."

sdodson:
@brenton ready to go

PR #1036, opened 2015-12-08
Fix absent state

joelddiaz:


kwoodson:
@joelddiaz, good fine.  Copy pasta error.  :+1:

PR #1037, opened 2015-12-08
Bump ansible requirement to 1.9.4

sdodson:
We'll be shipping Ansible 1.9.4 with the updated installer. We may as well require it to reduce the matrix of possible environments.


sdodson:
@brenton thoughts on this one? I know we'll be including ansible-1.9.4 in the installer update. It's also available in EPEL already for those not using OSE.

brenton:
I think this is ok.  I know Jason has been pushing for it.  @wshearn, is this ok with you?

detiber:
+1000

detiber:
We should add an issue for removing pre-1.9.4 workarounds after this is merged.

brenton:
I talked with Ops.  They're already using 1.9.4

PR #1038, opened 2015-12-08
atomic-openshift-installer: Error handling on yaml loading

smunilla:
This addresses the stack trace that has been plaguing recent demos. In the
case of an error with callback_facts.yaml the program output is much clearer
and a course of action is suggested.


abutcher:
:tada: :smile: :tada:

PR #1039, opened 2015-12-09
Fix version dependent image streams

sdodson:
While we'd created separate trees of files for v1.1 and v1.0 content we unfortunately failed to reset the v1.0 content back to the v1.0 spec.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1289043


sdodson:
@brenton we should include this in the installer update

PR #1040, opened 2015-12-09
Squash pcs install into one task.

abutcher:
Use `action: "{{ ansible_pkg_mgr }}"` to squash pcs install into a single task. We should probably do this in all places where we have a yum task and a dnf task but I'm making the change for this particular task because the `install_result` for dnf is interfering with setting the cluster password for pacemaker.

On skipped tasks, variables are overwritten with a value of `{'skipped': True}`.

https://bugzilla.redhat.com/show_bug.cgi?id=1288481

@sdodson @brenton

### DIFF
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -267,13 +267,8 @@
   when: openshift_master_ha | bool and openshift.master.cluster_method == 'native'
 
 - name: Install cluster packages
-  yum: pkg=pcs state=present
-  when: (ansible_pkg_mgr == "yum") and openshift_master_ha | bool and openshift.master.cluster_method == 'pacemaker'
-  register: install_result
-
-- name: Install cluster packages
-  dnf: pkg=pcs state=present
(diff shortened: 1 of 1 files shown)


brenton:
This looks good to me.

sdodson:
Agreed, I'll probably work on a followup to my containerization work today where I move all the RPM (yum and dnf) installation work into one file and containerized installation into another file. Included in that I was going to do what you've done here for all the yum/dnf duplication across the entire repo.

detiber:
We probably want to address the rest of the installs asap. Otherwise the restart handlers for etcd, master and node will be kicked off on the initial config run. 

This can cause errors when the first startup is interrupted with the restart. The restart handler was put in place because of this issue in the first place.

sdodson:
Lets go ahead and squash all the yum/dnf now then. I'll rebase the containerized stuff, and then do a followup to split out rpm and containerized install methods.

sdodson:
#1046 in for this, I may cull this back just to instances where we're registering the result of the installation but so far it's testing well for me.

Issue #1042, opened 2015-12-09
Error in session secret generation


sdodson:
@spinolacastro ran into this when they had set `openshift_master_session_auth_secrets` and `openshift_master_session_encryption_secrets` in their hosts file. It skips the generation but then for some reason it's falling into the default() where we expect the output from the openssl commands.

```
TASK: [Install OpenSSL package] ***********************************************
ok: [master0]

TASK: [Generate session authentication key] ***********************************
skipping: [master0] => (item=1)

TASK: [Generate session encryption key] ***************************************
skipping: [master0] => (item=1)

TASK: [set_fact ] *************************************************************
fatal: [master0] => Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/ansible/runner/__init__.py", line 586, in _executor
    exec_rc = self._executor_internal(host, new_stdin)
  File "/usr/lib/python2.7/site-packages/ansible/runner/__init__.py", line 789, in _executor_internal
    return self._executor_internal_inner(host, self.module_name, self.module_args, inject, port, complex_args=complex_args)
  File "/usr/lib/python2.7/site-packages/ansible/runner/__init__.py", line 1013, in _executor_internal_inner
    complex_args = template.template(self.basedir, complex_args, inject, fail_on_undefined=self.error_on_undefined_vars)
  File "/usr/lib/python2.7/site-packages/ansible/utils/template.py", line 140, in template
    d[k] = template(basedir, v, templatevars, lookup_fatal, depth, expand_lists, convert_bare, fail_on_undefined, filter_fatal)
  File "/usr/lib/python2.7/site-packages/ansible/utils/template.py", line 124, in template
    varname = template_from_string(basedir, varname, templatevars, fail_on_undefined)
  File "/usr/lib/python2.7/site-packages/ansible/utils/template.py", line 382, in template_from_string
    res = jinja2.utils.concat(rf)
  File "<template>", line 12, in root
  File "/home/azureuser/openshift-ansible/filter_plugins/oo_filters.py", line 83, in oo_collect
    retval = [FilterModule.get_attr(d, attribute) for d in data]
  File "/home/azureuser/openshift-ansible/filter_plugins/oo_filters.py", line 42, in get_attr
    ptr = ptr[attr]
KeyError: 'stdout'

FATAL: all hosts have already failed -- aborting
```

https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-master/config.yml#L238-L267

[...]

PR #1018, opened 2015-12-03
Clean up idempotency issues with session secrets.

abutcher:
* Session secrets are now cached.
* Playbooks will generate session secrets if they were not passed in via inventory variables and were not found in the first master's cache.
* Use the generated or overridden secrets on the other masters.
* Session secrets file is a template. Overriding after an initial install with generated secrets will rewrite the session secrets file.

Fixes #953

PR #1043, opened 2015-12-09
Simplify session secrets overrides.

abutcher:
#1042

@sdodson @brenton @spinolacastro

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -249,22 +249,14 @@
   - name: Generate session authentication key
     command: /usr/bin/openssl rand -base64 24
     register: session_auth_output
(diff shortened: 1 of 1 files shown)


abutcher:
Noting that I am reworking this section in https://github.com/openshift/openshift-ansible/pull/1018. In order to replace secrets you would currently need to delete `/etc/origin/master/session-secrets.yaml` on all masters.

sdodson:
LGTM, I'll test this later today unless @spinolacastro gets to it first. I haven't deployed an HA setup in a bit and I need to anyway.

Fixes #1042  (so github does the needful when this merges)

abutcher:
@brenton @spinolacastro verified in the attached issue!

PR #1046, opened 2015-12-09
Remove yum / dnf duplication

sdodson:
@maxamillion FYI

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
@@ -54,8 +54,8 @@
 - name: Verify upgrade can proceed
   hosts: oo_masters_to_config:oo_nodes_to_config
   tasks:
-  - name: Clean yum cache
-    command: yum clean all
+  - name: Clean package cache
+    command: "{{ ansible_pkg_mgr }} clean all"
 
   - set_fact:
       g_new_service_name: "{{ 'origin' if deployment_type =='origin' else 'atomic-openshift' }}"
@@ -149,9 +149,7 @@
     when: (embedded_etcd | bool) and (etcd_disk_usage.stdout|int > avail_disk.stdout|int)
 
   - name: Install etcd (for etcdctl)
-    yum:
-      pkg: etcd
-      state: latest
+    action: "{{ ansible_pkg_mgr }} name=etcd state=latest"
 
   - name: Generate etcd backup
     command: >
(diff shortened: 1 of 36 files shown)


playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
```diff
@@ -54,8 +54,8 @@
 - name: Verify upgrade can proceed
   hosts: oo_masters_to_config:oo_nodes_to_config
   tasks:
-  - name: Clean yum cache
-    command: yum clean all
+  - name: Clean yum/dnf cache
```
brenton: can `{{ ansible_pkg_mgr }}` be used in the name?

abutcher: Vars in task names won't be evaluated

brenton: Makes sense.

brenton:
@sdodson, looks like we need a rebase.  Is this something you want to have merged for our async update?

playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
detiber: "Clear package cache"?

detiber:
@brenton async update would be good, the separate tasks for install leave the ability for etcd, master or node corruption in environments with slow storage or network latency.

abutcher:
@dgoodwin FYI

sdodson:
@detiber from the other PR you commented
> We probably want to address the rest of the installs asap. Otherwise the restart handlers for etcd, master and node will be kicked off on the initial config run.
> This can cause errors when the first startup is interrupted with the restart. The restart handler was put in place because of this issue in the first place.

The two cases where I see that we were actually paying attention to the registered result were the iptables and firewalld installation. In all the other cases in the master and node roles we were `register: install_result` but not actually using it so I've removed those.

Am I missing some other way that we're conditionally restarting?

detiber:
Ah, nevermind me, we are testing the service result and not the install result.

sdodson:
> Ah, nevermind me, we are testing the service result and not the install result.

Cool, thanks for checking. I think we're good to merge now.

@brenton looking at the openshift-ansible that's released currently it doesn't have any of the yum/dnf duplication. The iptables/firewalld bits have potential to be a problem so I'd say we merge this for the async release.

PR #983, opened 2015-11-25
Adding ManageIQ service account

enoodle:
cc @simon3z

PR #1050, opened 2015-12-10
Propose adding ManageIQ service account by default

enoodle:
Signed-off-by: enoodle <efreiber@redhat.com>

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1121,7 +1121,7 @@ class OpenShiftFacts(object):
 
         common = dict(use_openshift_sdn=True, ip=ip_addr, public_ip=ip_addr,
                       deployment_type='origin', hostname=hostname,
-                      public_hostname=hostname, use_manageiq=False)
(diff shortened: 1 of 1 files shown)


detiber:
:+1: from me.

@twiest @wshearn

twiest:
I'll let @wshearn weigh in here, I'm not sure how this would affect his process.

enoodle:
For easy reference, ManageIQ service accounts were added in https://github.com/openshift/openshift-ansible/pull/983

simon3z:
@wshearn can you review?

wshearn:
It shouldn't be too much of an impact for us.

Though shouldn't this be set to true inside openshift_facts.py instead for consistency?(It is where we enable sdn, flannel, fluentd, etc)

enoodle:
@wshearn You are right, I fixed that.

detiber:
ok to test

detiber:
@wshearn @twiest This should be ready for merge.

simon3z:
@enoodle please rebase.

detiber:
@wshearn please merge.

PR #1052, opened 2015-12-11
Change controllers service type to simple.

abutcher:
* Controllers which don't get a lease don't `SDNotify` so use `Type=simple`.
* Template unit files every run so that these changes will be picked up.
* Remove `ignore_errors: yes` on controllers start.

@sdodson @dgoodwin

### DIFF
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -135,13 +135,11 @@
   template:
     src: atomic-openshift-master-api.service.j2
     dest: /usr/lib/systemd/system/{{ openshift.common.service_type }}-master-api.service
-    force: no
   when: openshift_master_ha | bool and openshift_master_cluster_method == "native"
 - name: Create the controllers service file
   template:
     src: atomic-openshift-master-controllers.service.j2
     dest: /usr/lib/systemd/system/{{ openshift.common.service_type }}-master-controllers.service
-    force: no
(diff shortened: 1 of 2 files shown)


sdodson:
LGTM. Lets make sure there's an origin issue created to fix the controllers and link it to this issue.

dgoodwin:
Looks good to me as well, I assume this will dramatically cut down the time we spend waiting for controller services that never come up in an HA install. 90 second timeout, I think we hit it twice per non-functioning master so it'll shave a few good minutes off install time.

abutcher:
@dgoodwin It speeds things up significantly, yep.

abutcher:
Created https://github.com/openshift/origin/issues/6272

dgoodwin:
yes please! I'm at almost an hour without this for just a 7 system cluster. :)

sdodson:
> Looks good to me as well, I assume this will dramatically cut down the time we spend waiting for controller services that never come up in an HA install. 90 second timeout, I think we hit it twice per non-functioning master so it'll shave a few good minutes off install time.

I'd expect them to be started and timed out in parallel, so roughly 90 seconds across all masters, no?

abutcher:
@sdodson Masters are serial

sdodson:
LGTM, @wshearn mind merging this while Brenton is afk?

abutcher:
@brenton heads up

PR #1055, opened 2015-12-11
Expose custom api url

spinolacastro:
We can leverage openshift_master_named_certificates and use valid certs for console and api.


abutcher:
@spinolacastro This can be closed right?

spinolacastro:
Yes! 
Thank you!

PR #1059, opened 2015-12-14
sync master -> prod branch

twiest:


twiest:
This was +1d in STG PR #1058

PR #1060, opened 2015-12-14
Sync prod -> int branch.

twiest:


twiest:
This was +1d in stg pr #1058

PR #1064, opened 2015-12-14
Adding option for cache reading and specify openshift version.

kwoodson:


bin/ohi
```diff
@@ -85,23 +92,27 @@ class Ohi(object):
...
-        parser.add_argument('-l', '--user', action='store', default=None,
-                               help='username')
+        parser.add_argument('-l', '--user', action='store', default=None, help='username')
 
+        parser.add_argument('-c', '--cache-only', action='store_true', default=False, 
+                            help='Retrieve the host inventory by cache only. Default is false.')
 
-        self.args = parser.parse_args()
+        parser.add_argument('-o', '--openshift-version', action='store', required=True,
+                            help='Specify the openshift account version.')
```
twiest: "account version" ?

I see what you are going for, but I don't think account version is correct.

I think just openshift version is probably good.

kwoodson:
ok to test

twiest:
:+1:

twiest:
:+1:

PR #1065, opened 2015-12-14
Use inventory value when checking for flannel

lebauce:


abutcher:
:+1: @brenton

abutcher:
aos-ci-test

arilivigni:
aos-ci-test

brenton:
I'm going to defer to @detiber on this one.  I think https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_common/tasks/main.yml#L25 should handle setting `use_flannel` if it's set in the inventory but it could be that we aren't properly resetting the openshift facts in some case.

I think the whole reason why these facts exist is so that the openshift playbooks can safely be reused.  For example, say you run with `openshift_use_flannel` in the inventory and it correctly sets the fact and configures flannel.  However, say I wrote an adhoc playbook for doing something administrative.  You then run it and maybe the inventory you're using at the time does not have `openshift_use_flannel`.  It's fairly common to create an adhoc inventory when running adhoc playbooks since you may only want it to run on a subset of hosts.  However, my playbook loads the Node playbook in a way you didn't expect and since `openshift_use_flannel` isn't set it changes the network configuring and breaks the environment.  If using facts this won't happen because even if  `openshift_use_flannel` isn't set the fact should still be set correctly.

lebauce:
@brenton You're right.

lebauce:
@brenton Right. Obviously, the openshift fact `use_flannel` was not set, I updated my branch with this fix.

detiber:
@lebauce hmm, we should not have to set use_flannel in roles/openshift_node, since we are setting it in roles/openshift_common.

Do you have a copy of an inventory and the output of running playbooks/byo/openshift_facts.yml on a cluster that exhibits the error?

lebauce:
@detiber `openshift.common.use_flannel` is used at the very top of `playbooks/common/openshift-node/config.yml`. At this step, the `openshift_node` role hasn't been used, and therefore `openshift_common`.

detiber:
@lebauce :+1: thanks, I missed that when focusing on the diff.

We are currently digging into the issue we've seen previously with using delegate_to, so hopefully we'll be able to find resolution there and remove a lot of the complexity from the node config playbook.

detiber:
@brenton lgtm

PR #1067, opened 2015-12-15
Fix registry modification for new deployment types.

dgoodwin:
New deployment types were missed when installing with additional docker
registries.


abutcher:
:+1: @twiest could you merge while @brenton is afield

twiest:
@detiber I'm not familiar with the deployment type part of openshift-ansible. Would you mind reviewing this? If you +1 it, I'll merge it?

abutcher:
@detiber is on PTO. Summoning @sdodson

sdodson:
@twiest looks good to me

PR #1070, opened 2015-12-16
Fix alternative bind address

spinolacastro:


roles/openshift_master/tasks/main.yml
```diff
@@ -201,7 +201,7 @@
     line: "{{ item.line }}"
   with_items:
     - regex: '^OPTIONS='
-      line: "OPTIONS=--loglevel={{ openshift.master.debug_level }} --listen=https://0.0.0.0:8443 --master=https://{{ openshift.common.ip }}:8443"
+      line: "OPTIONS=--loglevel={{ openshift.master.debug_level }} --listen=https://{{ openshift.master.bind_addr }}:{{ openshift.master.api_port }} --master=https://{{ openshift.common.ip }}:{{ openshift.master.api_port }}"
```
detiber: @sdodson Should we also update this for the master controller settings as well? I can't remember what address the controllers service is registered with in etcd.

abutcher: Controllers:
```
line: "OPTIONS=--loglevel={{ openshift.master.debug_level }} --listen=https://0.0.0.0:8444"
```

We'd need to use `openshift.master.bind_addr` here and probably add an `openshift.master.controllers_port` to facts. Also looks like our firewall rules always use defaults.

detiber: Yeah, the default firewall rules are a known issue. iirc, we couldn't rely on the openshift_fact generated defaults to feed into os_firewall module based on the role dependencies.

I've mostly been dragging my feet on it until we can move to only supporting firewalld.

abutcher:
aos-ci-test

detiber:
ok to test

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -463,6 +463,7 @@ def set_url_facts_if_unset(facts):
     if 'master' in facts:
         api_use_ssl = facts['master']['api_use_ssl']
         api_port = facts['master']['api_port']
+        controllers_port = facts['master']['controllers_port']
```
detiber: This line isn't needed.

roles/openshift_master/defaults/main.yml
```diff
@@ -6,7 +6,9 @@ os_firewall_allow:
 - service: etcd embedded
   port: 4001/tcp
 - service: api server https
-  port: 8443/tcp
+  port: "{{ openshift.master.api_port }}/tcp"
```
detiber: I'm actually surprised this works, but the last time I tested I don't think we were setting openshift_facts for masters in the playbooks for the certificate generation.

spinolacastro: Yeah, i tested in a already deployed install, i'll run a properly test.

spinolacastro:
Closing since @detiber cherrypicked these commits whitin https://github.com/openshift/openshift-ansible/pull/1175

PR #1073, opened 2015-12-16
split zagg.server.processor.errors into separate heartbeat and metrics error items (needed since the scripts are split now).

twiest:


mwoodson:
:+1

mwoodson:
:+1:

PR #1074, opened 2015-12-16
change ovs mount to /var/run/openvswitch

joelddiaz:
will not require a container restart if openvswitch service is restarted


twiest:
:+1:

PR #1057, opened 2015-12-14
Fix typos that cause master/node restarts which break openshift SDN setup

EricMountain-1A:
Fixes typo when setting facts to record whether master/node has been restarted already, to decide whether notify handler should restart services or not.

Currently, this causes random SDN network setup failures as openshift-node gets restarted while the setup script is running, and the subsequent start fails to configure the SDN because it thinks it's already done.

For instance:
```
##### Ansible starts openshift-node
Dec 11 08:06:39 192.168.105.4 ansible-service[12218]: Invoked with name=openshift-node pattern=None enabled=True state=started sleep=None arguments= runlevel=default
...
##### We see traces that the openshift-sdn setup script is running
Dec 11 08:06:43 192.168.105.4 ovs-vsctl[12549]: ovs|00001|vsctl|INFO|Called as ovs-vsctl add-port br0 vovsbr -- set Interface vovsbr ofport_request=9
...
##### Ansible does a restart of openshift-node
Dec 11 08:06:44 192.168.105.4 ansible-service[12675]: Invoked with name=openshift-node pattern=None enabled=None state=restarted sleep=None arguments= runlevel=default
...
##### openshift-sdn script is relaunched, however it thinks setup is done:
Dec 11 08:06:45 192.168.105.4 openshift-node[12773]: I1211 08:06:45.824998   12773 kube.go:29] Output of setup script:
Dec 11 08:06:45 192.168.105.4 openshift-node[12773]: + lock_file=/var/lock/openshift-sdn.lock
...
Dec 11 08:06:45 192.168.105.4 openshift-node[12773]: + return 1
Dec 11 08:06:45 192.168.105.4 openshift-node[12773]: + echo 'SDN setup not required.'
Dec 11 08:06:45 192.168.105.4 openshift-node[12773]: SDN setup not required.
Dec 11 08:06:45 192.168.105.4 openshift-node[12773]: + exit 140
```

Before this change:
```
TASK: [openshift_node | Start and enable node] ********************************
changed: [eric-node-compute-c93b9]
changed: [eric-node-infra-87f90]
changed: [eric-node-compute-273b0]

TASK: [openshift_node | set_fact node_service_status_changed = start_result | changed] ***
ok: [eric-node-compute-c93b9]
ok: [eric-node-infra-87f90]
ok: [eric-node-compute-273b0]

...

NOTIFIED: [openshift_node | restart node] *************************************
changed: [eric-node-compute-c93b9]
changed: [eric-node-compute-273b0]
[...]

Issue #780, opened 2015-11-01
Openshift v1.0.7: Installation failure. Docker 1.8.x dependency issue


sureddy:
when i tried to install openshift origin v1.0.7 release, it failed with the following message:

[...]

PR #808, opened 2015-11-04
Containerized install with SDN support

sdodson:
Supersedes #620 includes SDN support based on https://github.com/openshift/origin/pull/5442

This work adds containerized installation support including OpenvSwitch SDN enabled nodes. Suitable for use on Atomic Host and tested on RHEL Atomic Host 7.1.6 or later. I imagine it works on Fedora Atomic Host as well but I haven't personally tested there.

Images used
 - Origin 
  - openshift/origin
  - openshift/node (node + openshift-sdn + openvswitch rpm for client tools)
  - openshift/openvswitch (centos7 + openvswitch rpm, runs ovsdb ovsctl processes)
  - registry.access.redhat.com/rhel7/etcd
 - OpenShift Enterprise
  - openshift3/ose
  - openshift3/node
  - openshift3/openvswitch
  - registry.access.redhat.com/rhel7/etcd
 - Atomic Enterprise Platform
  - aep3/aep
  - aep3/node
  - aep3/openvswitch
  - registry.access.redhat.com/rhel7/etcd

This work should be considered a 'Technical Preview'. Whenever possible we have attempted to minimize the differences between containerized and RPM based installation methods. One point where things differ is in the installation of etcd. Currently etcd is installed in atomic ostree however there are plans to remove it in the next 3-6 months so I have disabled etcd in atomic-host and installed the containerized version of etcd as `etcd_container`. Another point where things differ is that RPM installs commonly write to /usr/lib which is read-only on atomic hosts so equivalent paths have been found in /etc.

Outstanding work to be completed before merging this PR
- [x] Add master HA services
- [x] Add check to ensure we're on atomic-host 7.1.6 or later which includes iptables (Fedora equivalent?)

Followup items for future PRs
- [x] Allow overriding master, node, openvswitch, and etcd image names
- [x] Allow overriding versions of images, currently latest tag is hard-coded. (just specify image name with version tag)
- [x] Consider what to do about deploying HA master load balancer in containerized environments. For now out of scope to install haproxy in containerized form

PR #1081, opened 2015-12-16
Refactors Docker setup to own role to avoid race in SDN setup

EricMountain-1A:
Refactors Docker setup to its own role to avoid race between start/stop and SDN setup that also restarts Docker.

What o-a does:

* docker role
    * tasks start docker
    * handler restarts docker and udevd
* openshift_node role
    * tasks start node
    * handler:
        * restarts node (if wasn't started by tasks, per #1057)
        * restarts docker

openshift-node runs openshift-sdn-kube-subnet-setup.sh on startup.  Amongst other things it:

* runs systemctl daemon-reload
* restarts docker

The docker restart above races with the one in the openshift_node role's handler, leading to the setup script exiting prematurely due to systemctl cancelling conflicting jobs relating to intertwined docker start/stop actions:

```
[root@192 ~]# journalctl | grep -E "Docker Application|ansible-service|Starting OpenShift|Output of setup script|Job for"
Dec 16 08:52:07 acs-e2e-test-eric-10340-node-compute-0.novalocal ansible-service[10372]: Invoked with name=iptables pattern=None enabled=True state=started sleep=None arguments= runlevel=default
Dec 16 08:52:37 192.168.243.4 ansible-service[11531]: Invoked with name=docker pattern=None enabled=True state=started sleep=None arguments= runlevel=default
Dec 16 08:52:37 192.168.243.4 systemd[1]: Starting Docker Application Container Engine...
Dec 16 08:52:37 192.168.243.4 systemd[1]: Started Docker Application Container Engine.
Dec 16 08:52:49 192.168.243.4 ansible-service[12281]: Invoked with name=openshift-node pattern=None enabled=True state=started sleep=None arguments= runlevel=default
Dec 16 08:52:49 192.168.243.4 systemd[1]: Starting OpenShift Node...
Dec 16 08:52:52 192.168.243.4 systemd[1]: Stopping Docker Application Container Engine...
Dec 16 08:52:52 192.168.243.4 ansible-service[12576]: Invoked with name=docker pattern=None enabled=None state=restarted sleep=None arguments= runlevel=default
Dec 16 08:52:52 192.168.243.4 systemd[1]: Starting Docker Application Container Engine...
Dec 16 08:52:52 192.168.243.4 openshift-node[12348]: I1216 08:52:52.371431   12348 kube.go:29] Output of setup script:
Dec 16 08:52:52 192.168.243.4 openshift-node[12348]: Job for docker.service canceled.
Dec 16 08:52:52 192.168.243.4 systemd[1]: Stopped Docker Application Container Engine.
Dec 16 08:52:52 192.168.243.4 openshift-node[12348]: Job for docker.service canceled.
Dec 16 08:52:52 192.168.243.4 systemd[1]: Starting Docker Application Container Engine...
Dec 16 08:52:52 192.168.243.4 systemd[1]: Started Docker Application Container Engine.
Dec 16 08:52:52 192.168.243.4 systemd[1]: Starting OpenShift Node...
Dec 16 08:52:53 192.168.243.4 openshift-node[12711]: I1216 08:52:53.138953   12711 kube.go:29] Output of setup script:
```

On the 2nd start of openshift-node, the setup script aborts claiming nothing needs to be done.

Other folks are seeing this too it seems - c.f. #780.

Objective of this PR is to refactor the docker config that is done in the openshift_node role, to the docker role.  This should eliminate the race.

With this patch:

```
openshift@eric-node-compute-8048a ~]$ sudo journalctl | grep -E "Docker Application|ansible-service|Starting OpenShift|Output of setup script|Job for"
Dec 16 16:01:38 eric-node-compute-8048a.example.com ansible-service[11134]: Invoked with name=iptables pattern=None enabled=True state=started sleep=None arguments= runlevel=default
Dec 16 16:01:55 eric-node-compute-8048a.example.com ansible-service[12383]: Invoked with name=docker pattern=None enabled=True state=started sleep=None arguments= runlevel=default
Dec 16 16:01:55 eric-node-compute-8048a.example.com systemd[1]: Starting Docker Application Container Engine...
Dec 16 16:01:56 eric-node-compute-8048a.example.com systemd[1]: Started Docker Application Container Engine.
Dec 16 16:01:59 eric-node-compute-8048a.example.com ansible-service[13051]: Invoked with name=openshift-node pattern=None enabled=True state=started sleep=None arguments= runlevel=default
Dec 16 16:02:00 eric-node-compute-8048a.example.com systemd[1]: Starting OpenShift Node...
Dec 16 16:02:03 eric-node-compute-8048a.example.com systemd[1]: Stopping Docker Application Container Engine...
Dec 16 16:02:03 eric-node-compute-8048a.example.com systemd[1]: Starting Docker Application Container Engine...
Dec 16 16:02:04 eric-node-compute-8048a.example.com systemd[1]: Started Docker Application Container Engine.
Dec 16 16:02:04 eric-node-compute-8048a.example.com openshift-node[13118]: I1216 16:02:04.092320   13118 kube.go:29] Output of setup script:
```

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -1,10 +1,88 @@
 ---
 # tasks file for docker
+
+- name: Set node facts applicable to docker
+  openshift_facts:
+    role: "{{ item.role }}"
+    local_facts: "{{ item.local_facts }}"
+  with_items:
+  - role: common
+    local_facts:
+      deployment_type: "{{ openshift_deployment_type }}"
+  - role: node
+    local_facts:
+      docker_log_driver:  "{{ lookup( 'oo_option' , 'docker_log_driver'  )  | default('',True) }}"
+      docker_log_options: "{{ lookup( 'oo_option' , 'docker_log_options' )  | default('',True) }}"
+      portal_net: "{{ openshift_master_portal_net | default(None) }}"
+
 - name: Install docker
   action: "{{ ansible_pkg_mgr }} name=docker state=present"
-  
+
+- stat: path=/etc/sysconfig/docker
+  register: docker_check
+
+  # TODO: Enable secure registry when code available in origin
+- name: Secure Registry and Logs Options
+  lineinfile:
+    dest: /etc/sysconfig/docker
+    regexp: '^OPTIONS=.*$'
+    line: "OPTIONS='--insecure-registry={{ openshift.node.portal_net }} \
+{% if ansible_selinux and ansible_selinux.status == '''enabled''' %}--selinux-enabled{% endif %} \
+{% if openshift.node.docker_log_driver is defined %} --log-driver {{ openshift.node.docker_log_driver }}  {% endif %} \
+{% if openshift.node.docker_log_options is defined %}   {{ openshift.node.docker_log_options |  oo_split() | oo_prepend_strings_in_list('--log-opt ') | join(' ')}}  {% endif %} '"
+  when: docker_check.stat.isreg
+  notify:
+    - restart docker
+
+- set_fact:
+    docker_additional_registries: "{{ lookup('oo_option', 'docker_additional_registries')
+                                      | oo_split() | union(['registry.access.redhat.com'])
+                                      | difference(['']) }}"
+  when: openshift.common.deployment_type == 'enterprise'
+
+- set_fact:
+    docker_additional_registries: "{{ lookup('oo_option', 'docker_additional_registries')
+                                      | oo_split() | difference(['']) }}"
+  when: openshift.common.deployment_type != 'enterprise'
+
+- name: Add personal registries
+  lineinfile:
+    dest: /etc/sysconfig/docker
+    regexp: '^ADD_REGISTRY=.*$'
+    line: "ADD_REGISTRY='{{ docker_additional_registries
+                            | oo_prepend_strings_in_list('--add-registry ') | join(' ') }}'"
+  when: docker_check.stat.isreg and docker_additional_registries
+  notify:
+    - restart docker
+
+- name: Block registries
+  lineinfile:
+    dest: /etc/sysconfig/docker
+    regexp: '^BLOCK_REGISTRY=.*$'
+    line: "BLOCK_REGISTRY='{{ lookup('oo_option', 'docker_blocked_registries') | oo_split()
+                              | oo_prepend_strings_in_list('--block-registry ') | join(' ') }}'"
+  when: docker_check.stat.isreg and
+        lookup('oo_option', 'docker_blocked_registries') != ''
+  notify:
+    - restart docker
+
+- name: Grant access to additional insecure registries
+  lineinfile:
+    dest: /etc/sysconfig/docker
+    regexp: '^INSECURE_REGISTRY=.*'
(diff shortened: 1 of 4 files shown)


roles/docker/tasks/main.yml
```diff
@@ -1,10 +1,88 @@
 ---
 # tasks file for docker
+
+- name: Set node facts applicable to docker
```
EricMountain-1A: I think this might be contentious: if we ever want to e.g. configure Docker on a master, and not only on nodes, this might conflict and require rework.

Thing is, it seems right to do all the Docker config in one place, but there is at least portal_net that is node-specific.

If you see this as a problem, then the only alternative I see is to create a docker_node role.  Keeping the docker config in the openshift_node simply doesn't fly due to the fact you can't guarantee docker restart (in the handler) before node start (in the tasks), and you can't guarantee node SDN setup will have completed before docker restart.

EricMountain-1A:
/cc abutcher twiest lhuard1A

This is the follow on from #1057.

roles/docker/tasks/main.yml
```diff
@@ -1,10 +1,88 @@
...
+  openshift_facts:
```
detiber: I'd like to avoid adding a requirement on openshift_facts to the docker role, since it is meant to be a generic docker role and not OpenShift specific.

Instead, I would prefer to see the OpenShift specific config managed through the playbooks or through an openshift_docker role.

That said, I do think that the log options and registry settings should belong in the docker role, just not the references to openshift_facts varaibles, openshift_* variables, or any of the filters defined in oo_filters.

detiber:
@EricMountain-1A thanks for digging into this issue! In general I really like conceptual changes, I'd just like to see the docker role avoid any openshift specific dependencies.

@sdodson I believe your containerization changes included moving some changes similar to this as well.

EricMountain-1A:
@detiber fair enough.  What solution should we implement to fix this bug then?  If there's a consensus on the way to go, and it's within my Ansible coding abilities (newbie), I'll take a bash at it.

e.g. would you agree to the creation of an openshift_docker role that uses openshift_facts? (the last part of your line note wasn't clear to me)

Note this is a major issue for us, causing significant time-wasting (I really should have dug into this sooner): whether a node comes up in a usable state or not is basically the flip of a coin.  To illustrate: we have some automated tests that spawn a 2-node cluster on the fly, and we have something like 1-in-4 chances (optimistically) the cluster will actually work and our tests pass.  Setting up "real" clusters with more nodes has a higher probability of ending up with at least 1 node that spoils the cluster, though we can mitigate by making it unschedulable and evacuating it - however this requires manual intervention.

detiber:
@EricMountain-1A I have no issues with putting the openshift specific config in an openshift_docker role. Alternatively, using the playbooks/common/openshift-node/config.yaml playbook to inject the openshift specific values into the docker role would be sufficient as well.

We've seen reports of similar errors, we just have not had not been able to reproduce it to date.

sdodson:
Yeah, #808 moves much of the docker config into openshift_docker role. We should get that merged soon.

EricMountain-1A:
OK, I see #808 was merged a few hours ago.  I'll take a look over the coming week to see if anything related to this issue is still needed and go from there.

EricMountain-1A:
Had a look at #808, and indeed this refactoring does what's needed to fix this issue.  #1105 opened to deal with a small remaining issue around Docker restarts.

PR #1087, opened 2015-12-17
Allow libvirt cluster to be restarted after host is rebooted

mwburdeos:
This change is specific to running OpenShift on a cluster of VMs managed through **libvirt**.  Currently, if you restart your host machine and attempt to restart your VMs the the **libvirt** Virtual Machine Manager UI, the VMs fail to start stating that the network is missing.

I created a YML file for Ansible that configures libvirt and allows the VMs to be restarted.  The YML file is called rehydrate.yml, because the purpose of the file is to allow the cluster to be restarted (rehydration).

It would have been easier for the user to run this command if it was integrated into the **bin/cluster** script; however I didn't do that because rehydration is specific to the **libvirt** provider and not applicable to the other providers.

Once the host is rebooted and the VMs no longer start, run the following from the same place you would run bin/cluster:

```
    ansible-playbook playbooks/libvirt/openshift-cluster/rehydrate.yml
```

The YML file re-executes only the LIBVIRT Configuration tasks, allowing the VMs to be restarted through the libvirt Virtual Machine Manager.

### DIFF
--- a/playbooks/libvirt/openshift-cluster/rehydrate.yml
+++ b/playbooks/libvirt/openshift-cluster/rehydrate.yml
@@ -1,0 +1,17 @@
+---
+- name: Rehydrate dormant OpenShift cluster
+  hosts: localhost
+  gather_facts: no
+  vars_files:
+  - vars.yml
+  vars:
+    os_libvirt_storage_pool: "{{ libvirt_storage_pool | default('images') }}"
+    os_libvirt_storage_pool_path: "{{ libvirt_storage_pool_path | default('/var/lib/libvirt/images') }}"
+    os_libvirt_network: "{{ libvirt_network | default('default') }}"
+    image_url: "{{ deployment_vars[deployment_type].image.url }}"
+    image_sha256: "{{ deployment_vars[deployment_type].image.sha256 }}"
+    image_name: "{{ deployment_vars[deployment_type].image.name }}"
+  tasks:
+  - include: tasks/configure_libvirt.yml
+
+


detiber:
@mwburdeos instead of having a separate re-hydrate command, would it make more sense to have the libvirt network created persist through reboots?

mwburdeos:
I'm not an expert in this.. and I only did this to get it working.  If the
libvirt network can be created to persist through reboots, that would be
much better.

On Thu, Dec 17, 2015 at 2:51 PM, Jason DeTiberus <notifications@github.com>
wrote:

> @mwburdeos <https://github.com/mwburdeos> instead of having a separate
> re-hydrate command, would it make more sense to have the libvirt network
> created persist through reboots?
>
> —
> <https://github.com/openshift/openshift-ansible/pull/1087#issuecomment-165578753>
> .
>



-- 
*Michael Burdeos*
Senior Software Developer
On Center Software, Inc.

P: 713.240.6854
W: https://www.oncenter.com

detiber:
@mwburdeos I believe all that is necessary is to change playbooks/libvirt/openshift-cluster/tasks/configure_libvirt_network.yml:
Instead of:
```
- name: Create libvirt network for openshift
  command: "virsh -c {{ libvirt_uri }} net-create {{ mktemp.stdout }}/network.xml"
  when: net_info_result.rc == 1
```
You would want to use net-define, net-start and net-autostart in place of the single net-create step.
net-define creates a persistent network instead of the transient network started by net-create, however persistent networks aren't started by default, so the net-start is needed to start the network and net-autostart is needed to start the network on subsequent restarts of libvirt.

mwburdeos:
Thanks Jason

On Thu, Dec 17, 2015 at 3:10 PM, Jason DeTiberus <notifications@github.com>
wrote:

> @mwburdeos <https://github.com/mwburdeos> I believe all that is necessary
> is to change
> playbooks/libvirt/openshift-cluster/tasks/configure_libvirt_network.yml:
> Instead of:
>
> - name: Create libvirt network for openshift
>   command: "virsh -c {{ libvirt_uri }} net-create {{ mktemp.stdout }}/network.xml"
>   when: net_info_result.rc == 1
>
> You would want to use net-define, net-start and net-autostart in place of
> the single net-create step.
> net-define creates a persistent network instead of the transient network
> started by net-create, however persistent networks aren't started by
> default, so the net-start is needed to start the network and net-autostart
> is needed to start the network on subsequent restarts of libvirt.
>
> —
> <https://github.com/openshift/openshift-ansible/pull/1087#issuecomment-165583001>
> .
>



-- 
*Michael Burdeos*
Senior Software Developer
On Center Software, Inc.

P: 713.240.6854
W: https://www.oncenter.com

detiber:
@sdodson can you take a look at this

tbielawa:
> @sdodson | jdetiber: tbielawa: interesting but no way i'll have time to review it, i think it should move to openshift-ansible-contrib ?
> @tbielawa | sdodson: that sounds right to me.
> jdetiber | tbielawa: it's kind of specific to bin/cluster, until we can move bin/cluster to -contrib, it wouldn't necessarily make sense to move that script

detiber:
@mwburdeos if this playbook still works for you, I'm happy to merge, since I don't think it has any negative impact on existing users.

mtnbikenc:
Closing this PR as it is stale and `libvirt` is no longer a part of the project.

PR #1088, opened 2015-12-17
Allow compression option to be set to empty for non compressed images

akram:
Enteprise or custom images deployment may not use a compressed image and hence the playbook crashes when the compression attribute is not defined.
This change allows to have a compression defined to empty string. And thus, the stored image will have no additional extension.

At the same time, support for tgz compression has been added for other images.

### DIFF
--- a/playbooks/libvirt/openshift-cluster/tasks/launch_instances.yml
+++ b/playbooks/libvirt/openshift-cluster/tasks/launch_instances.yml
@@ -13,15 +13,27 @@
   get_url:
     url: '{{ image_url }}'
     sha256sum: '{{ image_sha256 }}'
-    dest: '{{ os_libvirt_storage_pool_path }}/{{ [image_name, image_compression] | join(".") }}'
+    dest: '{{ os_libvirt_storage_pool_path }}/{{ [image_name, image_compression] | reject("equalto", "") | join(".") }}'
   when: '{{ ( lookup("oo_option", "skip_image_download") | default("no", True) | lower ) in ["false", "no"] }}'
   register: downloaded_image
 
-- name: Uncompress Base Cloud image
+- name: Uncompress xz compressed base cloud image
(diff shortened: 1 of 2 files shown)


akram:
@lhuard1A @detiber could you please review ?

playbooks/libvirt/openshift-cluster/tasks/launch_instances.yml
```diff
@@ -23,6 +23,12 @@
     creates: '{{ os_libvirt_storage_pool_path }}/{{ image_name }}'
   when: image_compression in ["xz"] and downloaded_image.changed
 
+- name: Uncompress TGZ Base Cloud image
+  command: 'tar zxvf {{ os_libvirt_storage_pool_path }}/{{ [image_name, image_compression] | join(".") }}'
+  args:
+    creates: '{{ os_libvirt_storage_pool_path }}/{{ image_name }}'
+  when: image_compression in ["tgz"] and downloaded_image.changed
```
detiber: Do we need to handle more compressed file formats than .tgz?

detiber:
@akram outside of the comment about compression format extensions, this change looks good to me (albeit untested).

playbooks/libvirt/openshift-cluster/tasks/launch_instances.yml
akram: I guess we should also handle gz format. For example, centos atomic is provided as .qcow.gz
I am adding it.

akram:
@dtieber added gz

detiber:
@twiest

twiest:
@akram sorry for the delay on this. Could you rebase it to the latest master please?

akram:
@twiest done

twiest:
@detiber are you still a +1 for this?

detiber:
@twiest :+1: from me :shipit:

PR #1090, opened 2015-12-17
atomic-openshift-installer: Improvements to wizard mode

smunilla:


utils/src/ooinstall/install_type.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def __init__(self, lower, upper):
+        self.lower = lower
+        self.upper = upper
+
+    def in_range(self, number):
+        return number >= self.lower and number <= self.upper
+
+
+SUPPORTED_INSTALL_TYPES = {
```
detiber: From a selfish perspective, it would be nice to have a way to get separate etcd hosts as well.

utils/src/ooinstall/oo_config.py
```diff
@@ -38,23 +38,14 @@ def __init__(self, **kwargs):
...
         if self.connect_to is None:
-            raise OOConfigInvalidHostError("You must specify either an ip " \
-                "or hostname as 'connect_to'")
+            raise OOConfigInvalidHostError(
+                "You must specify either an ip or hostname as 'connect_to'.")
 
-        if self.master is False and self.node is False and self.master_lb is False:
+        if self.is_master is False and self.is_node is False and self.is_master_lb is False:
             raise OOConfigInvalidHostError(
                 "You must specify each host as either a master or a node.")
```
detiber: Error message does not match the condition.  Also, are we assuming that is_master_lb will also host storage as well? if so, we should probably rename it.

utils/src/ooinstall/oo_config.py
```diff
@@ -76,30 +67,50 @@ def to_dict(self):
...
+    def is_master_lb(self):
+        """ Is this host a load balancer for all masters. """
+        return 'master_lb' in self.roles
+
+    def is_storage(self):
+        """ Will this host be NFS storage for a registry. """
+        return 'storage' in self.roles
+
 
 class OOConfig(object):
```
detiber: I don't see any conversion from v1 to v2 happening, we'll need to make sure to handle that case.

utils/src/ooinstall/openshift_ansible.py
```diff
@@ -131,6 +131,9 @@ def write_host(host, inventory, schedulable=None):
     elif not schedulable:
         facts += ' openshift_schedulable=False'
 
+    if host.is_infra_node:
+        facts += ' region=Infra'
```
detiber: I almost wonder if handling region in a better way would be a good idea... 
If allowing arbitrary config, the user could be specifying the regions. It'd also be nice to eventually support the concept of regions/zones in the guided config as well.

utils/src/ooinstall/oo_config.py
smunilla: Updated to convert from v1 to v2

PR #1092, opened 2015-12-18
Add the new metrics and logging urls

git001:
When you add the 2 new possibility

https://docs.openshift.org/latest/install_config/cluster_metrics.html
https://docs.openshift.org/latest/install_config/aggregate_logging.html

and then run the playbook again the both urls are gone.
With this PR you can add both urls in the hosts file and don't need to add in the new file the urls.

### DIFF
--- a/roles/openshift_master/templates/master.yaml.v1.j2
+++ b/roles/openshift_master/templates/master.yaml.v1.j2
@@ -8,6 +8,12 @@ assetConfig:
   logoutURL: ""
   masterPublicURL: {{ openshift.master.public_api_url }}
   publicURL: {{ openshift.master.public_console_url }}/
+{% if openshift.master.loggingPublicURL %}
+  loggingPublicURL: {{ openshift.master.loggingPublicURL }}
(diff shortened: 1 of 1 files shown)


roles/openshift_master/templates/master.yaml.v1.j2
```diff
@@ -8,6 +8,12 @@ assetConfig:
   logoutURL: ""
   masterPublicURL: {{ openshift.master.public_api_url }}
   publicURL: {{ openshift.master.public_console_url }}/
+{% if openshift.master.loggingPublicURL %}
```
detiber: This will use the value of loggingPublicURL set by openshift_facts, I do not see where that value is being set from this diff though.

Most likely you will also want to update roles/openshift_master/tasks.main.yml to include setting this value in the `openshift_facts` task as well.

I would also suggest using underscores rather than camelCase for the ansible variable names (logging_public_url instead of loggingPublicURL).

git001:
Okay. I will open a new pull request

PR #1094, opened 2015-12-18
Reset Type=notify for controllers service

abutcher:
This can be reset once https://github.com/openshift/origin/pull/6275 has been released. Also removing the ignore from the restart handler.

### DIFF
--- a/roles/openshift_master/templates/atomic-openshift-master-controllers.service.j2
+++ b/roles/openshift_master/templates/atomic-openshift-master-controllers.service.j2
@@ -7,7 +7,7 @@ Before={{ openshift.common.service_type }}-node.service
 Requires=network.target
 
 [Service]
-Type=simple
+Type=notify
 EnvironmentFile=/etc/sysconfig/{{ openshift.common.service_type }}-master-controllers
 Environment=GOTRACEBACK=crash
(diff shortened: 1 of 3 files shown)


abutcher:
Tested w/ the latest puddle @sdodson got me for Christmas

roles/openshift_master/templates/atomic-openshift-master-controllers.service.j2
```diff
@@ -7,7 +7,7 @@ Before={{ openshift.common.service_type }}-node.service
 Requires=network.target
 
 [Service]
-Type=simple
+Type=notify
```
detiber: We will probably need to conditionalize this based on the version, right? If version < supported version then Type=simple else type=notify

abutcher: Oh yeah for sure!

abutcher:
@detiber PTAL this one should be good to go with the conditional

detiber:
:+1:

sdodson:
LGTM

abutcher:
aos-ci-test

abutcher:
@brenton this one should be good to merge

PR #1097, opened 2015-12-21
Always pull images

sdodson:
Remove local image list inspection for now.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1292679

### DIFF
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -86,25 +86,9 @@
   action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-master{{ openshift_version  }} state=present"
   when: not openshift.common.is_containerized | bool
 
-- name: Get docker images
-  command: docker images
-  changed_when: false
-  when: openshift.common.is_containerized | bool
-  register: docker_images
-
 - name: Pull master image
   command: >
     docker pull {{ openshift.master.master_image }}
-  when: openshift.common.is_containerized | bool and openshift.master.master_image not in docker_images.stdout
-  
-- name: Wait for master image
-  command: >
-      docker images
-  register: docker_images
-  until: openshift.master.master_image in docker_images.stdout
-  retries: 30
-  delay: 10
-  changed_when: false
   when: openshift.common.is_containerized | bool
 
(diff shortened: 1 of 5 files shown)


roles/openshift_master/tasks/main.yml
```diff
@@ -86,7 +86,7 @@
   when: not openshift.common.is_containerized | bool
 
 - name: Get docker images
-  command: docker images
+  shell: docker images | awk '{  print $1 "\:" $2 }'
```
detiber: I think it would be better to have create a filter to parse the output of docker images rather than using shell/awk.

sdodson:
I hope to work on #1136 tomorrow but if I don't get to that I believe we should merge this instead before delivering the next installer build.

sdodson:
@detiber mind taking a look, this removes all the conditional image pull due to issues associated with how `docker images` formats the output not matching how we were looking for it. Based on bugzilla discussion it's better to always attempt to pull the image than fail right now. If the image is already there it should be a very quick operation. I'll work on a proper image list filter when I get back from PTO.

This is linked to a blocker bug.

playbooks/common/openshift-node/config.yml
```diff
@@ -215,6 +215,12 @@
                          | oo_collect('openshift.common.hostname') }}"
     openshift_node_vars: "{{ hostvars | oo_select_keys(groups['oo_nodes_to_config']) }}"
   pre_tasks:
-
+  - name: Wait for master to become available before proceeding
+    wait_for:
+      host: "{{ hostvars[groups.oo_first_master.0].openshift.common.ip }}"
+      port: 8443
```
detiber: We should pull the api port from facts as well, otherwise this will fail when overriding the ports.

detiber:
@sdodson outside of my nitpicking on the wait_for, I see no issues with this

playbooks/common/openshift-node/config.yml
sdodson: oh, sorry, this slipped in i'll pull it out of this PR

sdodson:
@detiber thanks for the quick review, dropped the extraneous commit from this PR
@brenton good to merge

PR #1099, opened 2015-12-22
Fix version interrogation on containerized installs

sdodson:
* get_openshift_version now searches the PATH for 'openshift'
 * client_binary and admin_binary also search the PATH for 'oc' and 'oadm'
 * openshift_cli reloads facts after installing cli wrappers

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -835,6 +835,24 @@ def get_current_config(facts):
 
     return current_config
 
+def search_path(filename, path=None):
+    """
+    Searches PATH for a file
+
+    Args:
+      name: the filename to search for
+      path: the optional path string (default: os.environ['PATH'])
+
+    Returns:
+      The abspath to the file or None if not found.
+    """
+    path = path or os.environ['PATH']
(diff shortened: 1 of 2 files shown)


abutcher:
aos-ci-test

abutcher:
LGTM tagging @detiber for a second look

abutcher:
@sdodson needs rebased

sdodson:
@detiber rebased, mind giving this a second pair of eyes?

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -842,13 +860,11 @@ def get_openshift_version():
             version: the current openshift version
     """
     version = None
-
-    if os.path.isfile('/usr/bin/openshift'):
-        _, output, _ = module.run_command(['/usr/bin/openshift', 'version'])
+    o_path = search_path('openshift')
```
detiber: Instead of searching the path, we could always just invoke the command and catch the exception if it isn't found. I'm not sure if module.run_command honors the environment though.

Also, we might run into a  situation where an atomic host has a modified search path, we may just want to check certain known locations instead.

PR #1100, opened 2015-12-22
Install all xpaas streams when enabled

sdodson:
We now ship multiple xpaas image stream files so install all of them when xpaas examples are enabled.


sdodson:
@brenton ready to merge

PR #1103, opened 2015-12-23
Remove lb from docker hosts.

abutcher:
@sdodson PTAL, I removed the `is_containerized | bool` from the haproxy package install since we assume it will never be containerized. Do you think that will be sufficient or should we guard the lb hosts by setting `is_containerized: false` when we gather facts for lbs?

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -232,6 +232,7 @@
       balance: source
       servers: "{{ hostvars.localhost.haproxy_backend_servers }}"
   roles:
+  - role: openshift_facts
   - role: haproxy
     when: groups.oo_masters_to_config | length > 1
 
--- a/playbooks/common/openshift-docker/config.yml
+++ b/playbooks/common/openshift-docker/config.yml
@@ -1,5 +1,5 @@
 - name: Configure docker hosts
-  hosts: oo_masters_to_config:oo_nodes_to_config:oo_etcd_to_config:oo_lb_to_config
+  hosts: oo_masters_to_config:oo_nodes_to_config:oo_etcd_to_config
   vars:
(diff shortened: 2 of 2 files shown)


roles/haproxy/tasks/main.yml
```diff
@@ -1,7 +1,6 @@
 ---
 - name: Install haproxy
   action: "{{ ansible_pkg_mgr }} name=haproxy state=present"
-  when: not openshift.common.is_containerized | bool
```
sdodson: I'd just leave this for now. I'm doing a sweep where I'll be differentiating between is_containerized and is_atomic and I'll get this there unless this is critical to fixing the docker config on lb hosts.

playbooks/common/openshift-master/config.yml
```diff
@@ -232,6 +232,7 @@
       balance: source
       servers: "{{ hostvars.localhost.haproxy_backend_servers }}"
   roles:
+  - role: openshift_facts
```
sdodson: Not sure this is necessary?

roles/haproxy/tasks/main.yml
abutcher: Okay I'll remove it, it'll work with the `openshift_facts` addition in `playbooks/common/openshift-master/config.yml`

playbooks/common/openshift-master/config.yml
abutcher: These hosts aren't gathering facts prior to this play (now that `oo_lb_to_config` is absent from docker hosts) so `is_containerized` won't be set for `haproxy` and `os_firewall` roles.

sdodson: Ahh, makes sense, thanks.

sdodson:
:+1: LGTM

abutcher:
aos-ci-test

abutcher:
@brenton this one should be good to merge

PR #1095, opened 2015-12-18
Fix restart handlers

abutcher:
These were never firing because `start_result | changed` wasn't templated. Also added `| bool` filters to the conditionals.

PR #1105, opened 2015-12-29
Fix Docker restart handling

EricMountain-1A:
* Fixes typo assigning docker_service_status_changed which leads to misinterpretation in handler.
* Fixes Docker restart handling to ensure openshift_docker role does restart Docker on change.

The docker role has a typo setting `docker_service_status_changed` (c.f. #1057 where this was fixed for master/node restarts).

If we just fix this typo bluntly, the configuration changes (e.g. additional registries) made subsequently in the `openshift_docker` role will only take effect by chance thanks to the openshift-node SDN setup script performing a docker restart (and therefore only on nodes I suppose).  Indeed, the `docker` role having started the service, it sets `docker_service_status_changed`, so any `restart docker` notifications signalled in `openshift_docker` are skipped.

I have 2 proposals to fix this:

1. Change the docker service start+enable task to only enable it, notify `restart docker` and remove the `docker_service_status_changed` business.  I dislike this option as it means we no longer guarantee that docker is in state started at some point.
2. Introduce a 2nd `restart docker` handler (with a different name since "[If two handler tasks have the same name, only one will run. *](http://docs.ansible.com/ansible/playbooks_intro.html)") specifically to be used by the openshift_docker role.

This PR implements option 2.

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -12,7 +12,7 @@
   register: start_result
 
 - set_fact:
-    docker_service_status_changed = start_result | changed
+    docker_service_status_changed: start_result | changed
 
 - include: udev_workaround.yml
   when: docker_udev_workaround | default(False)
--- a/roles/openshift_docker/handlers/main.yml
+++ b/roles/openshift_docker/handlers/main.yml
@@ -1,0 +1,6 @@
+---
+
+- name: restart openshift_docker
+  service:
+    name: docker
+    state: restarted
--- a/roles/openshift_docker/tasks/main.yml
+++ b/roles/openshift_docker/tasks/main.yml
@@ -18,7 +18,7 @@
 
 - stat: path=/etc/sysconfig/docker
   register: docker_check
-  
+
 - name: Set registry params
   lineinfile:
     dest: /etc/sysconfig/docker
@@ -36,7 +36,7 @@
     reg_fact_val: "{{ openshift.common.docker_insecure_registries }}"
     reg_flag: --insecure-registry
   notify:
-  - restart docker
+  - restart openshift_docker
 
 # TODO: Enable secure registry when code available in origin
 # TODO: perhaps move this to openshift_docker?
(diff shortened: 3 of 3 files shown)


EricMountain-1A:
/cc @lhuard1A 

@sdodson PTAL at this small fix on top of #808, thanks!

roles/docker/tasks/main.yml
```diff
@@ -12,7 +12,7 @@
   register: start_result
 
 - set_fact:
-    docker_service_status_changed = start_result | changed
+    docker_service_status_changed: start_result | changed
```
abutcher: I think this line will need to be templated based on my testing with #1095.

```
docker_service_status_changed: "{{ start_result | changed }}"
```

EricMountain-1A: @abutcher, taking the 1st example from #1095:

```
- set_fact:
    etcd_service_status_changed = "{{ start_result | changed }}"
```
(Note the `=` you used in #1095, vs `:` I've used.)

is in fact parsed by YAML as:

```
- 
  set_fact: "etcd_service_status_changed = \"{{ start_result | changed }}\""
```

Whereas:

```
- set_fact:
    docker_service_status_changed: start_result | changed
```

is parsed as 

```
- 
  set_fact: 
    docker_service_status_changed: "start_result | changed"
```

So I don't think it needs to be templated.  And according to my tests, it is firing.  The case that does not fire is:

```
- set_fact:
    etcd_service_status_changed = start_result | changed
```

abutcher: Good deal if it's firing on changes. I had checked a `debug: var=some_status_changed` and it was literally storing `"start_result | changed"` rather than `True` or `False`.

abutcher: Also good catch on that stray `=` fixing that now

abutcher:
aos-ci-test

abutcher:
aos-ci-test

sdodson:
LGTM, thanks @EricMountain-1A and @abutcher 

@brenton good to go

Issue #1111, opened 2016-01-04
I believe AMI ID needs changed


cooktheryan:
< TASK: fail msg="Could not find requested ami" >
 -----------------------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

failed: [localhost] => {"failed": true}
msg: Could not find requested ami

FATAL: all hosts have already failed -- aborting

***** I believe it should be 
     # centos-7, requires marketplace
"-    image: ami-96a818fe"
"+    image: ami-61bbf104"

cooktheryan:
Created PR  Incorrect AMI id #1112

detiber:
@twiest @brenton this can be closed now that https://github.com/openshift/openshift-ansible/issues/1112

PR #1112, opened 2016-01-04
Incorrect AMI id

cooktheryan:
I believe the ami ID changed since the initial documentation was created for AWS deployment


detiber:
ok to test

detiber:
@cooktheryan thanks for the PR!

@twiest @wshearn should be good to merge. Resolves https://github.com/openshift/openshift-ansible/issues/1111

PR #1113, opened 2016-01-05
Fix RHN subscription by explicitly attaching to the right pool

lhuard1A:
Depending on the user RHN account, attempting to spawn an OpenShift cluster can result in the following error:

```
TASK: [rhel_subscribe | Enable RHEL repositories] ***************************** 
failed: [lenaic-master-e8f6d] => {"changed": true, "cmd": ["subscription-manager", "repos", "--enable=rhel-7-server-rpms", "--enable=rhel-7-server-extras-rpms", "--enable=rhel-7-server-ose-3.0-rpms"], "delta": "0:00:12.185557", "end": "2016-01-05 11:07:45.918421", "rc": 1, "start": "2016-01-05 11:07:33.732864", "warnings": []}
stdout: Error: rhel-7-server-ose-3.0-rpms is not a valid repository ID. Use --list option to see valid repositories.
Repository 'rhel-7-server-rpms' is enabled for this system.
Repository 'rhel-7-server-extras-rpms' is enabled for this system.
failed: [lenaic-node-compute-224b8] => {"changed": true, "cmd": ["subscription-manager", "repos", "--enable=rhel-7-server-rpms", "--enable=rhel-7-server-extras-rpms", "--enable=rhel-7-server-ose-3.0-rpms"], "delta": "0:00:12.342549", "end": "2016-01-05 11:07:46.796444", "rc": 1, "start": "2016-01-05 11:07:34.453895", "warnings": []}
stdout: Error: rhel-7-server-ose-3.0-rpms is not a valid repository ID. Use --list option to see valid repositories.
Repository 'rhel-7-server-rpms' is enabled for this system.
Repository 'rhel-7-server-extras-rpms' is enabled for this system.

```

It seems that “auto-subscription” is not always magically working, so I want to be more explicit about the pool to attach to.

I tried the `pool` attribute of the `redhat_subscription` ansible module this way:
```
- name: RedHat subscriptions
  redhat_subscription:
    username: "{{ rhel_subscription_user }}"
    password: "{{ rhel_subscription_pass }}"
    pool: "^OpenShift Enterprise, Premium"
```
but I got the following error:
```
TASK: [rhel_subscribe | RedHat subscriptions] ********************************* 
failed: [lenaic-master-842f4] => {"failed": true}
msg: Failed to register with 'subscription.rhn.redhat.com': 'RhsmPool' object has no attribute 'PoolId'
failed: [lenaic-node-compute-4184b] => {"failed": true}
msg: Failed to register with 'subscription.rhn.redhat.com': 'RhsmPool' object has no attribute 'PoolId'
```

That’s why I had to “manually” invoke `subscription_manager`.

### DIFF
--- a/roles/rhel_subscribe/tasks/main.yml
+++ b/roles/rhel_subscribe/tasks/main.yml
@@ -4,6 +4,7 @@
 #       to make it able to enable repositories
 
 - set_fact:
+    rhel_subscription_pool: "{{ lookup('oo_option', 'rhel_subscription_pool') | default(rhsub_pool, True) | default('OpenShift Enterprise, Premium*', True) }}"
     rhel_subscription_user: "{{ lookup('oo_option', 'rhel_subscription_user') | default(rhsub_user, True) | default(omit, True) }}"
     rhel_subscription_pass: "{{ lookup('oo_option', 'rhel_subscription_pass') | default(rhsub_pass, True) | default(omit, True) }}"
     rhel_subscription_server: "{{ lookup('oo_option', 'rhel_subscription_server') | default(rhsub_server) }}"
@@ -30,7 +31,14 @@
   redhat_subscription:
     username: "{{ rhel_subscription_user }}"
     password: "{{ rhel_subscription_pass }}"
-    autosubscribe: yes
+
+- name: Retrieve the OpenShift Pool ID
+  command: subscription-manager list --available --matches="{{ rhel_subscription_pool }}" --pool-only
+  register: openshift_pool_id
+  changed_when: False
+
+- name: Attach to OpenShift Pool
+  command: subscription-manager subscribe --pool {{ openshift_pool_id.stdout_lines[0] }}
 
 - include: enterprise.yml
   when: deployment_type == 'enterprise'


roles/rhel_subscribe/tasks/main.yml
```diff
@@ -30,7 +30,14 @@
   redhat_subscription:
     username: "{{ rhel_subscription_user }}"
     password: "{{ rhel_subscription_pass }}"
-    autosubscribe: yes
+
+- name: Retrieve the OpenShift Pool ID
+  command: subscription-manager list --available --matches="OpenShift Enterprise, Premium*" --pool-only
```
detiber: We should probably have the Product name to match be configurable, there will be some that will want to use `OpenShift Enterprise, Standard`, demo subscriptions, beta subscriptions and even employee subscriptions instead.

Otherwise, I see no issues with handling the pool attachment outside of the module, since it is a bug in ansible < 2.0 and even when the pool subscriptions work, if given a regex it will subscribe to all of them rather than a single sub that matches.

lhuard1A: @detiber: I have just added an option to configure the Pool name to attach to: `rhel_subscription_pool`. Its default value is `OpenShift Enterprise, Premium*`.

I wish I could have used a more complex regex like `OpenShift Enterprise, (Standard|Premium).*` but it seems the `--matched` switch of `subscription_manager` doesn’t support such regex.

detiber: @lhuard1A I like this approach better anyway, there could be some accounts that have multiple entitlements and they may want to specify which one specifically. Also the potential of host consuming multiple entitlements would be bad as well.

detiber:
:+1: on these changes from me.

@brenton @twiest PTAL

brenton:
I'm definitely in favor of this.  I've never trusted auto-attaching because I have so many different subscriptions.

twiest:
merging based on comments above.

PR #1114, opened 2016-01-05
Fix update_repos_and_packages playbook which now needs openshift_facts

lhuard1A:
The `update_repos_and_packages` playbook is failing with the following error:
```
TASK: [rhel_subscribe | set_fact ] ******************************************** 
fatal: [lenaic-master-6b224] => error while evaluating conditional: deployment_type == "enterprise" and ansible_distribution == "RedHat" and lookup('oo_option', 'rhel_skip_subscription') | default(rhsub_skip, True) | default('no', True) | lower in ['no', 'false'] and not openshift.common.is_atomic | bool
fatal: [lenaic-node-compute-351cd] => error while evaluating conditional: deployment_type == "enterprise" and ansible_distribution == "RedHat" and lookup('oo_option', 'rhel_skip_subscription') | default(rhsub_skip, True) | default('no', True) | lower in ['no', 'false'] and not openshift.common.is_atomic | bool
```

I suspected that it was the openshift facts which were missing in that case and indeed, adding `openshift_facts` solved this issue.

### DIFF
--- a/playbooks/common/openshift-cluster/update_repos_and_packages.yml
+++ b/playbooks/common/openshift-cluster/update_repos_and_packages.yml
@@ -8,6 +8,5 @@
           ansible_distribution == "RedHat" and
           lookup('oo_option', 'rhel_skip_subscription') | default(rhsub_skip, True) |
             default('no', True) | lower in ['no', 'false']
-          and not openshift.common.is_atomic | bool
   - openshift_repos
   - os_update_latest
--- a/roles/rhel_subscribe/tasks/main.yml
+++ b/roles/rhel_subscribe/tasks/main.yml
@@ -41,4 +41,5 @@
   command: subscription-manager subscribe --pool {{ openshift_pool_id.stdout_lines[0] }}
 
 - include: enterprise.yml
-  when: deployment_type in [ 'enterprise', 'atomic-enterprise', 'openshift-enterprise' ]
+  when: deployment_type in [ 'enterprise', 'atomic-enterprise', 'openshift-enterprise' ] and
+        not openshift.common.is_atomic | bool
--- a/roles/rhel_subscribe/meta/main.yml
+++ b/roles/rhel_subscribe/meta/main.yml
@@ -1,0 +1,2 @@
+dependencies:
+  - openshift_facts


detiber:
@lhuard1A instead of adding the role to the playbook, I would rather see us add it to the role dependencies for openshift_repos. In roles/openshift_repos/meta/main.yml replacing:
```
dependencies: []
```
With:
```
dependencies:
- { role: openshift_common }
```
Should do the trick.

lhuard1A:
Thanks @detiber for your review.

In fact, [`roles/openshift_repos/meta/main.yml`](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_repos/meta/main.yml#L15) already contains
```
dependencies:
- { role: openshift_facts }
```

As the problematic expression is the `when` clause attached to the `rhel_subscribe` role, I tried to create the `openshift_facts` dependency in `roles/rhel_subscribe/meta/main.yaml` instead.

Doing so led to the following error:

```
TASK: [openshift_facts | Verify Ansible version is greater than 1.8.0 and not 1.9.0 and not 1.9.0.1] *** 
fatal: [lenaic-master-0636e] => error while evaluating conditional: deployment_type == "enterprise" and ansible_distribution == "RedHat" and lookup('oo_option', 'rhel_skip_subscription') | default(rhsub_skip, True) | default('no', True) | lower in ['no', 'false'] and not openshift.common.is_atomic | bool                                                                                                                                                                       
fatal: [lenaic-node-compute-eb8dc] => error while evaluating conditional: deployment_type == "enterprise" and ansible_distribution == "RedHat" and lookup('oo_option', 'rhel_skip_subscription') | default(rhsub_skip, True) | default('no', True) | lower in ['no', 'false'] and not openshift.common.is_atomic | bool                                                                                                                                                        
```

where 
* `Verify Ansible version is greater than 1.8.0 and not 1.9.0 and not 1.9.0.1` is a task of the `openshift_facts` role and
* `deployment_type == "enterprise" and […] and not openshift.common.is_atomic` is the `when` clause attached to the `rhel_subscribe` role.

My understanding is that when `roles/A/meta/main.yml` references a dependency on role `B`, a playbook containing:
```
- role: A
  when: C
```
is internally evaluated as if it was:
```
- role: B
  when: C
- role: A
  when: C
```

  * A = `rhel_subscribe`
  * B = `openshift_facts`
  * C = `[…] and not openshift.common.is_atomic`

But the evaluation of expression C needs role B to be evaluated first and the `when` clause is evaluated before the role it is attached to.

detiber:
@lhuard1A indeed, thanks for the clarification. I wonder if pushing the test for `not openshift.common.is_atomic` down into the rhel_subscribe role is a better option, then we could extend that role to support atomic hosts in the future.

detiber:
@sdodson thoughts on above?

sdodson:
Yeah, I think the is_atomic check should be pushed down into rhel_subscribe probably here https://github.com/openshift/openshift-ansible/blob/master/roles/rhel_subscribe/tasks/main.yml#L36 though I'm worried about installing of the satellite so possible here too https://github.com/openshift/openshift-ansible/blob/master/roles/rhel_subscribe/tasks/main.yml#L27

We'd still need to add openshift_facts to the rhel_subscribe role, correct?

detiber:
I'd say that we'd even want to refactor out the non-containerized bits into an included tasks file, so after running set_fact and then the tests, I would do an:
```
- include: non-containerized.yml
  when: not is_atomic
```
and push all of the tasks that change things down into non-containerized.yml

detiber:
@sdodson correct on needing to add openshift_facts as a dependency to rhel_subscribe

I'd probably say even move the variables that are being set by set_fact to default/main.yml as well, I can't see a reason to prefer set_fact over making them default variables.

sdodson:
Discussed with @cgwalters whether or not satellite certificate installation is possible on atomic host and it looks like it should work because it installs into /etc. Additionally, there's value in installing the cert if you're using Satellite 6.1 as a docker registry. So only the inclusion of enterprise.yml (where we manipulate repo subscription) needs to be skipped when is_atomic.

detiber:
@sdodson thanks for digging into that.

lhuard1A:
@detiber @sdodson Thanks for your investigation and your advice.
I now move the `is_atomic` check from `update_repos_and_packages.yml` to `rhel_subscribe`.
Could you check?

roles/rhel_subscribe/tasks/main.yml
```diff
@@ -41,4 +41,5 @@
   command: subscription-manager subscribe --pool {{ openshift_pool_id.stdout_lines[0] }}
 
 - include: enterprise.yml
-  when: deployment_type == 'enterprise'
+  when: deployment_type == 'enterprise' and
```
detiber:
```
     when: deployment_type in ["enterprise", "openshift-enterprise", "atomic-enterprise"] and
```

playbooks/common/openshift-cluster/update_repos_and_packages.yml
```diff
@@ -8,6 +8,5 @@
           ansible_distribution == "RedHat" and
           lookup('oo_option', 'rhel_skip_subscription') | default(rhsub_skip, True) |
             default('no', True) | lower in ['no', 'false']
```
detiber: Looking at the test for rhel_skip_subscription, we should probably update it while we are in here:
```
     when: deployment_type == "enterprise" and
            ansible_distribution == "RedHat" and
            lookup('oo_option', 'rhel_skip_subscription') | default(rhsub_skip, True) |
              default('no', True) | lower in ['no', 'false']
```
should probably be:
```
     when: deployment_type in ["enterprise", "openshift-enterprise", "atomic-enterprise"] and
            ansible_distribution == "RedHat" and
            not (lookup('oo_option', 'rhel_skip_subscription') | default(rhsub_skip, True) | bool)
```

lhuard1A: @detiber I have rebased this PR and the deployment_type check is now checking the 3 values as expected.

roles/rhel_subscribe/tasks/main.yml
lhuard1A: Done during the rebase.

detiber:
:+1: @twiest

lhuard1A:
Hi @twiest ,

Could you please have a look?

detiber:
@brenton ptal

PR #1115, opened 2016-01-05
Fix multi-word arguments & cli wrapper stdin plumbing

sdodson:
Fixes https://bugzilla.redhat.com/1295677
Fixes `echo "foo" | oc create -f -` by removing -t option entirely

### DIFF
--- a/roles/openshift_cli/templates/openshift.j2
+++ b/roles/openshift_cli/templates/openshift.j2
@@ -6,11 +6,4 @@ cmd=`basename $0`
 user=`id -u`
 group=`id -g`
 
-# docker can only split stderr and stdin when run without -t
-# https://github.com/docker/docker/issues/725
-# ansible checks various streams DO NOT CROSS THE STREAMS
-if [ -z $TERM ]; then
-  $t = '-it'
-fi
-
-docker run ${t} -a STDERR -a STDOUT -a STDIN --privileged --net=host --user=${user}:${group} -v ~/.kube:/root/.kube -v /tmp:/tmp -v {{ openshift.common.config_base}}:{{ openshift.common.config_base }} -e KUBECONFIG=/root/.kube/config --entrypoint ${cmd} --rm {{ openshift.common.cli_image }} ${@}
+docker run -i -a STDERR -a STDOUT -a STDIN --privileged --net=host --user=${user}:${group} -v ~/.kube:/root/.kube -v /tmp:/tmp -v {{ openshift.common.config_base}}:{{ openshift.common.config_base }} -e KUBECONFIG=/root/.kube/config --entrypoint ${cmd} --rm {{ openshift.common.cli_image }} "${@}"


sdodson:
@ncdc mind taking a look at this to make sure the wrapper script is as versatile as possible? The purpose is to provide /usr/local/bin/openshift wrapper script that can be deployed on atomic hosts invoking openshift inside a container where it's otherwise impossible to install a local binary.

detiber:
@sdodson if the wrapper script is going to be a pain to maintain, maybe we shouldn't even bother with it...  I don't see a reason that we couldn't tell users to run docker exec against the openshift-master image instead.

We'd have to do some contortions ourselves to invoke the oc/oadm commands for containerized installs, but we'd have more control over what the inputs are there though.

sdodson:
I hope that it's not a pain to maintain. I'd really like for running `oc`, etc to be as easy on atomic host as it is elsewhere. I may revisit the `atomic install` option having that extract the oc binary from the image into /usr/local/bin/ but that seems more error prone.

detiber:
@sdodson another option would be to ship oc/oadm with atomic host as well...  Though really, I'm not completely convinced that the convenience of having the client tooling available on the host itself as being necessary outside of a demo system (provided we have sufficient documentation around setting up a remote administrative client).

sdodson:
I hope this is it, it works for every permutation I can think of so far.

 - `oc create -f /etc/origin/foo.json` (we note that you need to place json files in /etc/origin as it's mounted)
 - `echo "json" | oc create -f -`
 - `oc create -f - < /any/path.json`

Right now unfortunately containerized installs are broken because the manageiq stuff was enabled and relies on the second form above.

detiber:
The first example above kind of highlights my concern a bit...  we are still requiring users to have to workaround the containerized oc...  Documenting something along the lines of "It just works, except a, b, and c" is worse than just saying "This isn't available on an Atomic Host installation, instead do this..."

roles/openshift_cli/templates/openshift.j2
```diff
@@ -6,11 +6,4 @@ cmd=`basename $0`
...
-# docker can only split stderr and stdin when run without -t
-# https://github.com/docker/docker/issues/725
-# ansible checks various streams DO NOT CROSS THE STREAMS
-if [ -z $TERM ]; then
-  $t = '-it'
-fi
-
-docker run ${t} -a STDERR -a STDOUT -a STDIN --privileged --net=host --user=${user}:${group} -v ~/.kube:/root/.kube -v /tmp:/tmp -v {{ openshift.common.config_base}}:{{ openshift.common.config_base }} -e KUBECONFIG=/root/.kube/config --entrypoint ${cmd} --rm {{ openshift.common.cli_image }} ${@}
\ No newline at end of file
+docker run -i -a STDERR -a STDOUT -a STDIN --privileged --net=host --user=${user}:${group} -v ~/.kube:/root/.kube -v /tmp:/tmp -v {{ openshift.common.config_base}}:{{ openshift.common.config_base }} -e KUBECONFIG=/root/.kube/config --entrypoint ${cmd} --rm {{ openshift.common.cli_image }} "${@}"
```
ncdc: afaik you shouldn't need to specify the 3 `-a` args.

Why does this need to be privileged?

sdodson:
@detiber Yeah, I agree it's not ideal but I think it's necessary to have something readily accessible on the master to get them to the point that they can add another admin. At least right now oadm isn't part of the client download but I think they're working to roll oadm into the oc binary.

How about emitting a banner like this?

```
ATTENTION: You are running $cmd via a wrapper around `docker run $img`. This
wrapper is intended only to be used to bootstrap an environment. Please install
the full client tools on another host once you have granted cluster-admin
privileges to a user. See $docs_url for more information.
```

sdodson:
@brenton @abutcher This is the PR I mentioned during standup where we need to decide whether or not to deliver the wrapper script for containerized installs.

roles/openshift_cli/templates/openshift.j2
sdodson: I think I found that ansible, which relies on differentiating the stdio streams, it wasn't working properly without all three `-a` args but I'll go through and test again.

Yeah, I don't think privileged is necessary, testing without out it now.

abutcher:
Since we're currently deploying the wrapper script (and it's broke) I think we should at least merge this fix while we're working on/discussing an alternative. +1 on including the banner too.

detiber:
:+1: on fixing the outstanding bug and adding the banner.
We should discuss the pro/cons of the wrapper script at some point before 3.2 though.

roles/openshift_cli/templates/openshift.j2
sdodson: Thanks, I found that -a args weren't necessary. --privileged is however because we're often writing files to host mounted paths like /etc/origin/master/ca.crt  We probably need to revisit this when we stop running the master as root though.

PR #1118, opened 2016-01-05
Add ability to disable os_firewall

detiber:
@abutcher ptal

### DIFF
--- a/roles/os_firewall/defaults/main.yml
+++ b/roles/os_firewall/defaults/main.yml
@@ -1,2 +1,3 @@
 ---
+os_firewall_enabled: True
 os_firewall_use_firewalld: True
--- a/roles/os_firewall/tasks/main.yml
+++ b/roles/os_firewall/tasks/main.yml
@@ -1,6 +1,6 @@
 ---
 - include: firewall/firewalld.yml
-  when: os_firewall_use_firewalld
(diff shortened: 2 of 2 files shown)


abutcher:
aos-ci-test

abutcher:
@detiber Should we call this out in the example inventory files?

detiber:
@abutcher possibly, I'm thinking about going back in and making it the default for cloud based installs as well...

detiber:
@brenton @sdodson @abutcher @twiest: Looking for some feedback on the though of disabling the firewall by default when deploying to cloud instances, since the hosts are already managed by security_groups.

The user would also have the choice to override the defaults by setting os_firewall_enabled

detiber:
@brenton @abutcher: still looking for a second opinion about whether it would be 'intuitive' to the users to default the local firewall to 'off' (wide open) for cloud-based deployments?

abutcher:
@detiber Everyone will probably end up doing things differently wrt using host firewalls or fine grained NACLs vs security groups but I think it's fine to steer users towards security groups for cloud environments with the option to enable os firewall...

detiber:
@brenton I think I'd like to go ahead and get this merged as an undocumented change for now, and add a card for addressing the dynamic setting of the firewall config.

PR #1121, opened 2016-01-06
Rolling restart playbook for masters

abutcher:
Builds on @dgoodwin's work and supports single master, native ha and pacemaker ha configurations. Allows users to configure `openshift_rolling_restart_mode` to either `"system"` or `"services"` (defaults to `"services"`).

In the case of pacemaker ha and restart mode = system, we
1. Restart non active masters
2. Wait for non-active masters to restart
3. Fail over to first non-active master
4. Wait for API to be available
5. Restart the previously active master
6. Wait for the master to start

Other configurations will restart masters or services serially while checking services after each restart.

@detiber


playbooks/common/openshift-cluster/restart.yml
```diff
@@ -0,0 +1,161 @@
...
+      - role: common
+        local_facts:
+          rolling_restart_mode: "{{ openshift_rolling_restart_mode | default('services') }}"
+      - role: master
+        local_facts:
+          cluster_method: "{{ openshift_master_cluster_method | default(None) }}"
+
+# Creating a temp file on localhost, we then check each system that will
+# be rebooted to see if that file exists, if so we know we're running
+# ansible on a machine that needs a reboot, and we need to error out.
```
detiber: Do we want to error out here, or pause with a message stating that the localhost will be skipped (or performed last and not verified)?

That said, we should probably break out restarting the different host types into separate playbooks (and put them under the associated playbooks/common directories for those components) and have this playbook include those playbooks to perform the restart tasks...  maybe even flags to skip restarting certain host types or to only restart a single host type?

detiber: Grr, just saw this only restarts masters...  that said, I'd move it to playbooks/common/openshift-master

abutcher: Yeah, the name is misleading +1

abutcher:
@detiber The entry point for the master rolling restart playbook is now `playbooks/byo/openshift-master/restart.yml`. I haven't re-organized `evaluate_groups` but I can chop that up if you think it's worth it here.

playbooks/byo/openshift-master/restart.yml
```diff
@@ -0,0 +1,8 @@
+---
+- include: ../../common/openshift-master/restart.yml
+  vars:
```
detiber: You should be able to use vars_files with playbooks/byo/openshift-cluster/cluster_hosts.yml instead of defining g_*_hosts

playbooks/common/openshift-master/restart.yml
```diff
@@ -0,0 +1,163 @@
...
+  tasks:
+  - file: path="{{ hostvars.localhost.mktemp.stdout }}" state=absent
+    changed_when: false
+
+- name: Fail if restarting the system where ansible is running
+  hosts: oo_masters_to_config
+  any_errors_fatal: true
+  tasks:
+  - fail: msg="Cannot run playbook on a host that will be restarted."
+    when: exists.stat.exists
```
detiber: I still don't lilke the idea of failing here...  In my mind, we should probably warn the user with the option to cancel out (using the pause module).

If they continue through then we can make sure that the current master is the last one that is restarted.

We would just have to make sure that the warning tells the user that the restart of the current master will cause the playbook to stop executing and they will have to verify that the host and services will need to be manually verified.

playbooks/common/openshift-master/restart.yml
```diff
@@ -0,0 +1,163 @@
...
+
+- name: Restart masters (native)
+  hosts: oo_masters_to_config
+  serial: 1
+  tasks:
+  - name: Restart master system
```
detiber: Okay, I'm starting to think that we should have a task file that we include that does the actual restart, we can conditionally include a task file for restarting the host or one for restarting the services for each type of cluster_method

Then we could do something like this:
```
- name: Restart masters
  hosts: oo_masters_to_config
  serial: 1
  tasks:
  - include: restart_host.yml
    when: (openshift.common.rolling_restart_mode == 'system' and openshift.master.cluster_method == 'native') or (openshift.master.cluster_method == 'pacemaker' and not is_active)

  - include: restart_native_services.yml
    when: openshift.master.cluster_method == 'native'

  - include: restart_pacemaker_services.yml
    when: openshift.common.rolling_restart_mode == 'service' and openshift.master.cluster_method == 'pacemaker' and not is_active
```
post_tasks:
- include: wait_for_native_service.yml
  when: openshift.master.cluster_method == 'native'
- include: wait_for_pacemaker_services.yml
  when: openshift.master.cluster_method == 'pacemaker' and not is_active

We would also need to shift around the pacemaker active logic and still perform the restart on the active host last. 

Think,ing about it a bit more, we should probably just use add_host to create 3 groups and have a playbook that does the above for each...
first group would be all the hosts we want to just roll through (native ha hosts + pacemaker hosts that aren't active) 
- the current host
second group would be pacemaker hosts that are active
third group would consist of the current host.

Then we could push all the restart and validation logic into a playbook that we include 3 times and pass in the list of hosts that comprise the appropriate group.

Man, I'm longing for ansible v2 and being able to loop over a block...

playbooks/common/openshift-master/restart_hosts.yml
```diff
@@ -0,0 +1,28 @@
...
+# When cluster_method is pacemaker we can only ensure that the host
+# restarted successfully.
+- name: Wait for master to start
+  become: no
+  local_action:
+    module: wait_for
+      host="{{ inventory_hostname }}"
+      state=started
+      delay=10
+  when: openshift.master.cluster_method == 'pacemaker'
```
detiber: Should we also check the cluster status from pcs and ensure that the current host is showing as available?

detiber: We should probably check the cluster status from pcs before starting the reboot and bomb out if the cluster is degraded or unhealthy.

abutcher: Adding a pacemaker cluster validation that will run on `oo_active_masters` before we restart anything and then I will ensure that the rebooted master becomes available afterwards.

abutcher:
@detiber Added some pacemaker validations

detiber:
@brenton :+1:

PR #1123, opened 2016-01-06
We can't enable manageiq for installations less than OSE 3.1 or Origi…

brenton:
…n 1.1

This is because 'oadmn policy add-scc-to-user' didn't exist in 3.0

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -715,6 +715,26 @@ def set_version_facts_if_unset(facts):
 
     return facts
 
+def set_manageiq_facts_if_unset(facts):
+    """ Set manageiq facts. This currently includes common.use_manageiq.
+
+        Args:
+            facts (dict): existing facts
+        Returns:
+            dict: the facts dict updated with version facts.
+        Raises:
+            OpenShiftFactsInternalError:
+    """
+    if 'common' not in facts:
+        if 'version_greater_than_3_1_or_1_1' not in facts['common']:
+            raise OpenShiftFactsInternalError(
+                "Invalid invocation: The required facts are not set"
+            )
+
+    facts['common']['use_manageiq'] = facts['common']['version_greater_than_3_1_or_1_1']
+
+    return facts
+
 def set_sdn_facts_if_unset(facts, system_facts):
     """ Set sdn facts if not already present in facts dict
 
@@ -1021,6 +1041,11 @@ def set_container_facts_if_unset(facts):
     return facts
 
 
+class OpenShiftFactsInternalError(Exception):
+    """Origin Facts Error"""
+    pass
+
+
 class OpenShiftFactsUnsupportedRoleError(Exception):
     """Origin Facts Unsupported Role Error"""
     pass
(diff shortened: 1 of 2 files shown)


brenton:
@detiber, PTAL

playbooks/common/openshift-master/config.yml
```diff
@@ -351,7 +351,7 @@
   - role: openshift_cluster_metrics
     when: openshift.common.use_cluster_metrics | bool
   - role: openshift_manageiq
-    when: openshift.common.use_manageiq | bool
+    when: openshift.common.version_greater_than_3_1_or_1_1 | bool and openshift.common.use_manageiq | bool
```
detiber: I would probably skip the version_greater_than_3_1_or_1_1 test here and instead have the openshift_manageiq role fail if version_greater_than_3_1_or_1_1 is false.

It might require adding the openshift_facts role as a dependency in the openshift_manageiq role if it isn't already (meta/main.yml in the role dir).

roles/openshift_manageiq/tasks/main.yaml
```diff
@@ -1,4 +1,8 @@
 ---
+- fail:
+    msg: "use_manageiq requires OpenShift Enterprise 3.1 or Origin 1.1."
+  when: openshift.common.version_greater_than_3_1_or_1_1 | bool == False
```
brenton: @detiber, does this look better?  I'm not entirely certain I have the syntax correct.

detiber: I'd probably just go with:
```
when: not openshift.common.version_greater_than_3_1_or_1_1 | bool
```
but that is just being pedantic :)

Outside of that, I would update the message to be "The openshift_manageiq role requires..."

abutcher: `use_manageiq` is enabled by default right?

detiber: Not with the openshift_fact changes @brenton made above.

abutcher: Oof... we disable for 3.0 and this message is for overrides, gotcha.

brenton:
cc @enoodle, @simon3z

abutcher:
aos-ci-test

brenton:
@detiber gave me his +1 on irc.... merging

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -715,6 +715,26 @@ def set_version_facts_if_unset(facts):
...
+        Raises:
+            OpenShiftFactsInternalError:
+    """
+    if 'common' not in facts:
+        if 'version_greater_than_3_1_or_1_1' not in facts['common']:
+            raise OpenShiftFactsInternalError(
+                "Invalid invocation: The required facts are not set"
+            )
+
+    facts['common']['use_manageiq'] = facts['common']['version_greater_than_3_1_or_1_1']
```
simon3z: @detiber @brenton about `version_greater_than_3_1_or_1_1` (probably just about the naming), shouldn't be "equal or greater than"? (e.g. what about 3.1.0)

simon3z:
@brenton thanks for getting this! IIUC `use_manageiq` it is still automatically set to `True` when `version_greater_than_3_1_or_1_1`, isn't it? cc @enoodle

roles/openshift_facts/library/openshift_facts.py
detiber: Yes, it really should be gte, but the variable actually tests against version 3.0.2.900, since that was the version number for our first 3.1 RC :)

enoodle:
@simon3z Yes, it does.

PR #1126, opened 2016-01-06
Setting relative paths in the upgrade playbooks wasn't working

brenton:


brenton:
@detiber, PTAL

detiber:
As much as I'd love to try to work around this, I'm drawing blanks right now. It seems like the issue is that it is attempting to resolve the vars_files from multiple different working directories (hence the conflicting errors with ../../ and ../../../../

:+1: on this

PR #1128, opened 2016-01-07
Make bin/cluster able to spawn an OSE 3.1 cluster

lhuard1A:
This patch contains the changes required to ask `bin/cluster` to create an OSE cluster with the `openshift-enterprise` deployment type.

### DIFF
--- a/bin/cluster
+++ b/bin/cluster
@@ -294,11 +294,8 @@ if __name__ == '__main__':
         meta_parser.add_argument('cluster_id', help='prefix for cluster VM names')
 
     meta_parser.add_argument('-t', '--deployment-type',
-                             choices=['origin', 'online', 'enterprise'],
+                             choices=['origin', 'online', 'enterprise', 'atomic-enterprise', 'openshift-enterprise'],
                              help='Deployment type. (default: origin)')
-    meta_parser.add_argument('-T', '--product-type',
-                             choices=['openshift', 'atomic-enterprise'],
-                             help='Product type. (default: openshift)')
     meta_parser.add_argument('-o', '--option', action='append',
                              help='options')
 
--- a/playbooks/aws/openshift-cluster/vars.yml
+++ b/playbooks/aws/openshift-cluster/vars.yml
@@ -1,5 +1,19 @@
 ---
 debug_level: 2
+
+deployment_rhel7_ent_base:
+  # rhel-7.1, requires cloud access subscription
+  image: ami-10663b78
+  image_name:
+  region: us-east-1
+  ssh_user: ec2-user
+  sudo: yes
+  keypair: libra
+  type: m4.large
+  security_groups: [ 'public' ]
+  vpc_subnet:
+  assign_public_ip:
+
 deployment_vars:
   origin:
     # centos-7, requires marketplace
@@ -25,15 +39,6 @@ deployment_vars:
     security_groups: [ 'public' ]
     vpc_subnet:
     assign_public_ip:
-  enterprise:
-    # rhel-7.1, requires cloud access subscription
-    image: ami-10663b78
-    image_name:
-    region: us-east-1
-    ssh_user: ec2-user
-    sudo: yes
-    keypair: libra
-    type: m4.large
-    security_groups: [ 'public' ]
-    vpc_subnet:
-    assign_public_ip:
+  enterprise: "{{ deployment_rhel7_ent_base }}"
+  openshift-enterprise: "{{ deployment_rhel7_ent_base }}"
+  atomic-enterprise: "{{ deployment_rhel7_ent_base }}"
--- a/playbooks/common/openshift-cluster/update_repos_and_packages.yml
+++ b/playbooks/common/openshift-cluster/update_repos_and_packages.yml
@@ -4,7 +4,7 @@
     openshift_deployment_type: "{{ deployment_type }}"
   roles:
   - role: rhel_subscribe
-    when: deployment_type == "enterprise" and
+    when: deployment_type in ["enterprise", "atomic-enterprise", "openshift-enterprise"] and
           ansible_distribution == "RedHat" and
           lookup('oo_option', 'rhel_skip_subscription') | default(rhsub_skip, True) |
             default('no', True) | lower in ['no', 'false']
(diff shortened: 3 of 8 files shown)


bin/cluster
```diff
@@ -294,7 +294,7 @@ if __name__ == '__main__':
         meta_parser.add_argument('cluster_id', help='prefix for cluster VM names')
 
     meta_parser.add_argument('-t', '--deployment-type',
-                             choices=['origin', 'online', 'enterprise'],
+                             choices=['origin', 'online', 'enterprise', 'openshift-enterprise'],
```
detiber: We should probably also add 'atomic-enterprise' to the list of deployment-types and the product-type flag below should probably be removed, since it's been removed in the ansible playbooks.

playbooks/aws/openshift-cluster/vars.yml
```diff
@@ -37,3 +37,15 @@ deployment_vars:
     security_groups: [ 'public' ]
     vpc_subnet:
     assign_public_ip:
+  openshift-enterprise:
```
detiber: I'm wondering if we should go ahead and create a new variable that can be used for setting multiple deployment types here to avoid duplication.
Something like:
```
rhel7_ent_base:
    # rhel-7.1, requires cloud access subscription
    image: ami-10663b78
    image_name:
    region: us-east-1
    ssh_user: ec2-user
    sudo: yes
    keypair: libra
    type: m4.large
    security_groups: [ 'public' ]
    vpc_subnet:
    assign_public_ip:

enterprise: "{{ rhel7_ent_base }}"
openshift-enterprise: "{{ rhel7_ent_base }}"
atomic-enterprise: "{{ rhel7_ent_base }}"
```

playbooks/common/openshift-cluster/update_repos_and_packages.yml
```diff
@@ -4,7 +4,7 @@
     openshift_deployment_type: "{{ deployment_type }}"
   roles:
   - role: rhel_subscribe
-    when: deployment_type == "enterprise" and
+    when: deployment_type in ["enterprise", "openshift-enterprise"] and
```
detiber: atomic-enterprise should be tested here as well.

playbooks/gce/openshift-cluster/vars.yml
```diff
@@ -19,3 +19,8 @@ deployment_vars:
     machine_type: n1-standard-1
     ssh_user:
     sudo: yes
+  openshift-enterprise:
+    image: rhel-7
+    machine_type: n1-standard-1
+    ssh_user:
+    sudo: yes
```
detiber: See my comment on playbooks/aws/openshift-cluster/vars.yml

playbooks/libvirt/openshift-cluster/vars.yml
```diff
@@ -25,18 +25,25 @@ deployment_vars:
...
+    ssh_user: openshift
+    sudo: yes
+  openshift-enterprise:
+    image:
+      url:    "{{ lookup('oo_option', 'image_url') |
+                  default('https://access.cdn.redhat.com//content/origin/files/sha256/25/25f880767ec6bf71beb532e17f1c45231640bbfdfbbb1dffb79d2c1b328388e0/rhel-guest-image-7.2-20151102.0.x86_64.qcow2', True) }}"
+      name:   "{{ lookup('oo_option', 'image_name') |
+                  default('rhel-guest-image-7.2-20151102.0.x86_64.qcow2', True) }}"
+      sha256: "{{ lookup('oo_option', 'image_sha256') |
+                  default('25f880767ec6bf71beb532e17f1c45231640bbfdfbbb1dffb79d2c1b328388e0', True) }}"
```
detiber: See my comment on playbooks/aws/openshift-cluster/vars.yml

playbooks/openstack/openshift-cluster/vars.yml
```diff
@@ -30,6 +30,10 @@ deployment_vars:
     ssh_user: root
     sudo: no
   enterprise:
-    image: "{{ lookup('oo_option', 'image_name') | default('rhel-guest-image-7.1-20150224.0.x86_64', True) }}"
+    image: "{{ lookup('oo_option', 'image_name') | default('rhel-guest-image-7.2-20151102.0.x86_64', True) }}"
+    ssh_user: openshift
+    sudo: yes
+  openshift-enterprise:
```
detiber: See my comment on playbooks/aws/openshift-cluster/vars.yml

roles/rhel_subscribe/tasks/openshift-enterprise.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Disable all repositories
+  command: subscription-manager repos --disable="*"
+
+- name: Enable RHEL repositories
+  command: subscription-manager repos \
+               --enable="rhel-7-server-rpms" \
+               --enable="rhel-7-server-extras-rpms" \
+               --enable="rhel-7-server-ose-3.1-rpms"
```
detiber: This will definitely work for now, however we might want to think about having a way to specify the version, since going forward we will not be updating deployment_types along with versions.

detiber:
@lhuard1A thanks for the PR, outside of my mostly nitpicky comments I'm all for getting this merged.

playbooks/aws/openshift-cluster/vars.yml
lhuard1A: I tried something like:
```
deployment_vars:
  rhel7_ent_base:
    …
  enterprise: "{{ rhel7_ent_base }}"
```
but this produces:
```
fatal: [localhost] => One or more undefined variables: 'rhel7_ent_base' is undefined
```

So, I tried:
```
deployment_vars:
  rhel7_ent_base:
    …
  enterprise: "{{ deployment_vars.rhel7_ent_base }}"
```
but this produces:
```
ERROR: recursive loop detected in template string
```

That’s why I had to define the “common settings” variable at the top level.
Would you know if I missed another Ansible trick?

roles/rhel_subscribe/tasks/openshift-enterprise.yml
lhuard1A: Ok.
I tried to make something a little bit more generic.

lhuard1A:
@detiber Thanks for your review ! Could you check my changes ?

playbooks/aws/openshift-cluster/vars.yml
```diff
@@ -1,5 +1,19 @@
 ---
 debug_level: 2
+
+deployment_rhel7_ent_base:
```
detiber: Nice! I didn't think about having to have this be outside of the deployment_vars dict, but it makes sense. Thanks for cleaning that up!

detiber:
@lhuard1A outstanding work once again! :+1: 

@twiest

twiest:
:+1:

PR #1129, opened 2016-01-07
Bug 1296388 - The playbook still configure ManageIQ when openshift_us…

brenton:
…e_manageiq is false

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -730,8 +730,8 @@ def set_manageiq_facts_if_unset(facts):
             raise OpenShiftFactsInternalError(
                 "Invalid invocation: The required facts are not set"
             )
-
-    facts['common']['use_manageiq'] = facts['common']['version_greater_than_3_1_or_1_1']
+    if 'use_mangeiq' not in facts['common']:
+        facts['common']['use_manageiq'] = facts['common']['version_greater_than_3_1_or_1_1']
 
     return facts
 


brenton:
aos-ci-test

brenton:
@detiber, @abutcher PTAL

abutcher:
LGTM - I've seen aos-ci-test not kick off if I comment right after opening a PR. Curious if it'll grab it in the middle of a comment too

brenton:
aos-ci-test

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -730,6 +730,9 @@ def set_manageiq_facts_if_unset(facts):
             raise OpenShiftFactsInternalError(
                 "Invalid invocation: The required facts are not set"
             )
+    if 'use_manageiq' in facts['common']:
+        if facts['common']['use_manageiq'] is not None:
+            return facts
```
detiber: I think I would rather see something like this instead of returning early:
```
if 'use_mangeiq' not in facts['common']:
    facts['common']['use_manageiq'] = facts['common']['version_greater_than_3_1_or_1_1']
```

Since we are not setting a default for `use_manageiq` before `set_manageiq_facts_if_unset()` then we would only set the default if it is not currently set.

@abutcher added some groundwork for validation with `validate_local_facts()`, it probably wouldn't hurt to test that  facts['common']['use_manageiq'] is indeed a boolean like we expect there instead of testing against None in this method.

brenton: I see it as two cases and I was thinking `set_manageiq_facts_if_unset` should handle both.  If `use_manageiq` doesn't even exist in the common facts then it should set it.  If it does exist in the common facts but is set to the default of None then it should be set.

However, thinking more about how we're setting defaults I think we have logic in two places:

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_common/tasks/main.yml#L26

There we set the default to `None` and here our real goal is for the default to be based on `version_greater_than_3_1_or_1_1`.  Is the logic in `openshift_common` actually needed there?

detiber: So...  any facts passed in to the openshift_facts module with a value of None are actually discarded. The main purpose was to give us a way to conditionally set the fact based on the presence/lack of presence of the variable.

https://github.com/brenton/openshift-ansible/blob/manageiq/roles/openshift_facts/library/openshift_facts.py#L1286-L1293 is where we discard items that are set to None or to an empty string.

brenton: Ahh, that makes a lot more sense now.  I'll update the PR.

brenton: updated!

brenton:
aos-ci-test

PR #1131, opened 2016-01-07
Fix terminate.yml on OpenStack

lhuard1A:
Trying to invoke `bin/cluster terminate […] openstack […]` causes the following error:
```
PLAY [Terminate instance(s)] ************************************************** 

TASK: [add_host ] ************************************************************* 
fatal: [localhost] => with_items expects a list or a set

```

It looks like the `with_items` clause is bogus. If my understanding is correct, we’re looking for the hosts that mach both the `environment` and the `clusterid`.


detiber:
:+1: @twiest ptal

twiest:
:+1:

PR #1132, opened 2016-01-07
Rename env into clusterid and add environment in the OpenStack VMs tags

lhuard1A:
The meta-tags attached to VMs that are used to generate Ansible groups have changed:
* `env` has been renamed `clusterid`;
* `environment` has been introduced.

This patch applies those changes on the OpenStack VMs.


detiber:
:+1: @twiest

twiest:
:+1:

PR #1133, opened 2016-01-07
Increase OpenStack stack creation/deletion timeout

lhuard1A:
The OpenStack instance I have access to is a little bit too slow for the timeout values that are currently used.
In practice, each time it takes more than 30 seconds to create a HEAT stack or more than 60 seconds to delete it, `bin/cluster` exits in failure saying that the OpenStack stack is still in `CREATE_IN_PROGRESS`/`DELETE_IN_PROGRESS`.

Increasing the `sleep` value has 2 effects:
* effectively increasing the timeout value;
* reducing the polling frequency at which Ansible polls OpenStack so that it reduces the stress on the OpenStack controllers.

### DIFF
--- a/playbooks/openstack/openshift-cluster/launch.yml
+++ b/playbooks/openstack/openshift-cluster/launch.yml
@@ -29,6 +29,7 @@
 
   - name: Create or Update OpenStack Stack
     command: 'heat {{ heat_stack_action }} -f {{ openstack_infra_heat_stack }}
+             --timeout 3 --enable-rollback
              -P cluster_env={{ cluster_env }}
              -P cluster_id={{ cluster_id }}
              -P cidr={{ openstack_network_cidr }}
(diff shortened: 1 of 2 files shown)


playbooks/openstack/openshift-cluster/launch.yml
```diff
@@ -55,7 +55,7 @@
     register: stack_show_status_result
```
detiber: Is it possible to have the heat command block on creating the stack? If so, then it might be better to run that as an async task and poll for the completion/result instead of performing the retries here...

lhuard1A: I 100% agree with your point @detiber !

Unfortunately, it was not possible to ask `heat stack-create` to be synchronous in version 0.4.0. The latest version I just tried (0.8.0) learnt the `--poll` switch which aims at making `heat stack-create` synchronous.
I tried it and with this option, `heat stack-create --poll` never returns. It seems there is an issue in the exit condition. So, it is not usable.

Moreover, in heat 0.8.0, the `--poll` switch exists only on `stack-create` and not on `stack-delete` so that there are still no way to make the stack deletion synchronous.

detiber: @lhuard1A thanks for the background info on the issue. Considering the limitations of the tools, I am good with the proposed changes.

detiber:
:+1: @twiest

detiber:
@twiest @wshearn PTAL

lhuard1A:
Hi @twiest ,

Could you please have a look?

detiber:
@brenton ptal

PR #1134, opened 2016-01-07
Add a banner to CLI wrapper instructing users that it's only for boot…

sdodson:
…strapping

Also remove -a STDIN,STDOUT,STDERR that's the default
The banner is sent to STDERR so it doesn't pollute STDIN when used to dump json/yaml then import it later.


sdodson:
@detiber As discussed in #1115

detiber:
:+1: @brenton

PR #1135, opened 2016-01-07
atomic-openshift-installer: Add support for installing OpenShift Orgin

smunilla:


detiber:
@maxamillion @ryanj are either of you guys able to give this a run through.

detiber:
@dgoodwin this might be a good candidate to work on while @smunilla is out. This and the config file format refactoring to handle arbitrary ansible variables are both requirements for the ManageIQ integration work being done.

detiber:
I don't think we need to worry about versions of Origin for now, since we do not have previous versions of the packages available. We should just always install the 'latest' version.

After we officially ship RPM packages through the CentOS SIG work that @tdawson is doing we can re-evaluate the approach.

dgoodwin:
No problem, just looking to give it a run-through and touchup anything needed?

detiber:
@dgoodwin indeed, thanks!

dgoodwin:
Looks like I'm running into a 1.1 issue: https://lists.openshift.redhat.com/openshift-archives/users/2016-January/msg00327.html

I don't know of a way to bypass examples in atomic-openshift-installer directly, I modified the generated hosts file instead, not a perfect test, but hopefully good enough. Aside from that bug this PR correctly generates a hosts file and installs origin:

```
[OSEv3:vars]
ansible_ssh_user=root
deployment_type=origin
```

We can change the variant version to "latest" which alters what the user sees for option 4, will be written to installer.cfg as "variant_version", and has no effect on the inventory file for origin. (see example of what was written above) 

I'll put in a PR to Sam's PR for him to merge when he gets back, if we're in a rush we can close this PR and I'll submit a new one. 

Otherwise everything LGTM.

dgoodwin:
Switching to latest as the display string: https://github.com/smunilla/openshift-ansible/pull/5

herlo:
While attempting to install origin with atomic-openshift-installer, I get an error saying that it can't find atomic-openshift. My configuration looks similar to the attached 
[aos-installer.cfg.yml.txt](https://github.com/openshift/openshift-ansible/files/245978/aos-installer.cfg.yml.txt). The results are included in the attached 
[installer.log.txt](https://github.com/openshift/openshift-ansible/files/245980/installer.log.txt)

Steps to repeat are as follows:

1) git clone git://github.com/openshift/openshift-ansible
2) cd openshift-ansible && tito build --test --rpm --output=/srv/www/oarepo/
3) createrepo_c /srv/www/oarepo/noarch/
4) create repo file for all nodes / masters pointing to repository above
5) generate installer config
6) run installer as described above

Expected result is the installer will use the origin rpms
Actual result is the installer attempts to look for atomic-openshift rpm which is intentionally not available.

detiber:
@herlo you will want to update the following variables:
```
variant: 'origin'
```

@smunilla is it possible to leave variant_version unset for unattended? We probably just want to target the latest available for origin (for now)

utils/src/ooinstall/variants.py
```diff
@@ -49,9 +49,15 @@ def latest_version(self):
     ]
 )
 
-# Ordered list of variants we can install, first is the default.
-SUPPORTED_VARIANTS = (OSE, AEP)
+origin = Variant('origin', 'OpenShift Origin',
+    [
+        Version('1.1', 'origin')
```
abutcher: This can be 1.2 now, yeah?

tdawson: Yep.
1.2 is now in the CentOS OpenShift Origin repo.

maxamillion:
@detiber apologies, this somehow got lost in my end ... I'll look into this asap. Thanks!

sdodson:
aos-ci-test

utils/src/ooinstall/variants.py
```diff
@@ -49,9 +49,16 @@ def latest_version(self):
     ]
 )
 
-# Ordered list of variants we can install, first is the default.
-SUPPORTED_VARIANTS = (OSE, AEP)
+origin = Variant('origin', 'OpenShift Origin',
+    [
+        Version('1.2', 'origin')
```
abutcher: Need a comma on the end of this guy.

abutcher:
aos-ci-test

tdawson:
[test]

tdawson:
aos-ci-test

utils/src/ooinstall/variants.py
```diff
@@ -49,9 +49,16 @@ def latest_version(self):
     ]
 )
 
-# Ordered list of variants we can install, first is the default.
-SUPPORTED_VARIANTS = (OSE, AEP)
+origin = Variant('origin', 'OpenShift Origin',
+    [
+        Version('1.2', 'origin'),
+        Version('1.1', 'origin')
```
detiber: Is there any value in keeping 1.1 around at this point? I believe 1.2 is the first origin release that we shipped to the CentOS channels which is the only place that I would expect to find older versions at this point.

utils/src/ooinstall/variants.py
```diff
@@ -49,9 +49,16 @@ def latest_version(self):
...
+    ]
+)
 
+# Ordered list of variants we can install, first is the default.
+SUPPORTED_VARIANTS = (OSE, AEP, origin)
```
detiber: s/AEP//g

utils/src/ooinstall/variants.py
tdawson: CentOS released origin 1.1.6, which was basically a 1.2.0 preview.  We released it in CentOS because we were setting things up and needed to release something. 
I'm fine either way.  As I write up my CentOS Documentation I'm pretending 1.2 is the first.

detiber: @tdawson ah, I had forgotten about that. 

@smunilla I would prefer that we avoid 1.1 here, since the documentation isn't going to mention it and I would like to avoid people installing and asking for support for 1.1.6 :)

smunilla:
Easy enough change. Updated.

detiber:
aos-ci-test

sdodson:
aos-ci-test

utils/src/ooinstall/variants.py
sdodson: :+1:

detiber:
lgtm

sdodson:
aos-ci-test

sdodson:
unit tests pass, the other tests are failing with a known issue.

PR #1137, opened 2016-01-07
Add a wait before node registration to ensure the master is available

sdodson:
This is necessary because on containerized installs restarting the node often triggers restarting the master and there's up to 60 seconds of delay before systemd restarts the master.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1293826

### DIFF
--- a/roles/openshift_master/handlers/main.yml
+++ b/roles/openshift_master/handlers/main.yml
@@ -2,11 +2,20 @@
 - name: restart master
   service: name={{ openshift.common.service_type }}-master state=restarted
   when: (not openshift_master_ha | bool) and (not (master_service_status_changed | default(false) | bool))
+  notify: Verify API Server
 
 - name: restart master api
   service: name={{ openshift.common.service_type }}-master-api state=restarted
   when: (openshift_master_ha | bool) and (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  notify: Verify API Server
 
 - name: restart master controllers
   service: name={{ openshift.common.service_type }}-master-controllers state=restarted
   when: (openshift_master_ha | bool) and (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+
+- name: Verify API Server
+  wait_for:
+    host: "{{ openshift.common.ip }}"
+    port: "{{ openshift.master.api_port }}"
+    state: started
+    timeout: 180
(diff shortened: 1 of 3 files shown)


detiber:
Man, that is ugly...  But I guess it makes sense, since if docker is being restarted the master will go poof. Can you conditionalize it for just containerized installs with a comment for why?

As an aside, I've hit issues with the api server not being available and bombing out at openshift_examples...  Maybe we should add a check at the end of the master role to validate that the public and private api endpoints are working before proceeding.

sdodson:
It looks like we can chain handlers, so the `restart master` handler could call a `Veryify API server` handler which waits for the API to become available after a restart. How's it look now?

roles/openshift_master/handlers/main.yml
```diff
@@ -2,11 +2,21 @@
...
 - name: restart master api
   service: name={{ openshift.common.service_type }}-master-api state=restarted
   when: (openshift_master_ha | bool) and (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  notify: Verify API Server
 
 - name: restart master controllers
   service: name={{ openshift.common.service_type }}-master-controllers state=restarted
   when: (openshift_master_ha | bool) and (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+
+- name: Verify API Server
```
detiber: This handler won't trigger on the initial service start...  for that you'd need to add a notify to the service calls in tasks/main.yml.

roles/openshift_master/handlers/main.yml
```diff
@@ -2,11 +2,21 @@
...
+  wait_for:
+    host: "{{ openshift.common.ip }}"
+    port: "{{ openshift.master.api_port }}"
+    state: started
+    timeout: 180
+    delay: 10
```
detiber: A 10 second delay here seems a bit excessive, I can see it for the containerized test in the node playbook though.

roles/openshift_master/handlers/main.yml
sdodson: Thanks, added notifies on initial start ups. I've reduced it to 5 seconds.

sdodson:
@detiber think i'm done with this one, discussed with @abutcher a bit how similar this is to some of the problems he's working on.

detiber:
@sdodson: @abutcher has been tweaking the method used to verify the api server, I'd like to see both use cases use the same method.

sdodson:
Sounds good to me, @abutcher want to roll these changes into your PR as I'm
going to be out for a while?

abutcher:
@sdodson Sure thing

sdodson:
Superseded by https://github.com/openshift/openshift-ansible/pull/1159

PR #1139, opened 2016-01-08
Add local machine name in /etc/hosts on OpenStack

lhuard1A:
The OpenStack instance I am using doesn’t provide any DNS. As a consequence, my OpenStack VMs are using external DNS servers which are not fed by the DHCP servers that allocated the IPs to the VMs.

The result is that the VMs have names which are not solvable by DNS. Ex:
```
[root@lenaic-master-0 ~]# hostname
lenaic-master-0.novalocal
[root@lenaic-master-0 ~]# dig +short lenaic-master-0.novalocal
[root@lenaic-master-0 ~]#
```

This could be fixed by migrating to an OpenStack version which implements DNS or by having `openshift-ansible` deploy a DNS server and its own DHCP server replacing the one provided by OpenStack but I would rather stick with a simpler solution.

What is done so far is that we use IPs rather than DNS names.

Still, with OpenShift 3.1, I have the following error:
```
Jan 08 06:01:01 lenaic-master-0.novalocal atomic-openshift-master[78306]: E0108 06:01:01.371587   78306 common.go:722] Failed to lookup IP address for node lenaic-master-0.novalocal: lookup lenaic-master-0.novalocal: no such host
Jan 08 06:01:01 lenaic-master-0.novalocal atomic-openshift-master[78306]: F0108 06:01:01.371623   78306 flatsdn.go:27] SDN initialization failed: lookup lenaic-master-0.novalocal: no such host
Jan 08 06:01:01 lenaic-master-0.novalocal systemd[1]: atomic-openshift-master.service: main process exited, code=exited, status=255/n/a
Jan 08 06:01:01 lenaic-master-0.novalocal systemd[1]: Unit atomic-openshift-master.service entered failed state.
Jan 08 06:01:01 lenaic-master-0.novalocal systemd[1]: atomic-openshift-master.service failed.
Jan 08 06:01:01 lenaic-master-0.novalocal systemd[1]: atomic-openshift-master.service holdoff time over, scheduling restart.
```

The `Failed to lookup IP address for […]` error message is coming from [netutils.GetNodeIP](https://github.com/openshift/openshift-sdn/blob/master/pkg/netutils/common.go#L72) which is called from [osdn.BaseInit](https://github.com/openshift/openshift-sdn/blob/master/plugins/osdn/common.go#L65) which is feeding GetNodeIP with the output of `uname -n`.
So, the hostname openshift-sdn tries to resolve here is not coming from a configuration file where we could have put directly an IP instead.

To conclude, in order to have the OVS SDN plugin of OpenShift able to start, it must be possible to do:
```
getent hosts $(uname -n)
```
(This is the bash equivalent of what is done [there](https://github.com/openshift/openshift-sdn/blob/master/plugins/osdn/common.go#L56) and [there](https://github.com/openshift/openshift-sdn/blob/master/pkg/netutils/common.go#L72))

In the absence of DNS able to resolve the OpenStack VMs hostname, amending `/etc/hosts` is a way to make the above commands work.

### DIFF
--- a/playbooks/openstack/openshift-cluster/launch.yml
+++ b/playbooks/openstack/openshift-cluster/launch.yml
@@ -133,6 +133,14 @@
       - parsed_outputs.node_floating_ips
       - parsed_outputs.infra_floating_ips
 
+- name: Add self in /etc/hosts because OpenStack may lack DNS
+  hosts: "tag_clusterid_{{ cluster_id }}"
+  tasks:
+  - name: Update /etc/hosts
+    lineinfile:
+      line: "{{ ansible_default_ipv4.address }} {{ ansible_nodename }}"
+      dest: /etc/hosts
+
 - include: update.yml
 
 - include: list.yml


detiber:
@lhuard1A ah, the issue is because we are no longer setting the nodeIp in the node config by default. When the nodeIp is specified the reverse resolution is skipped.

It is still possible to set the nodeIp (using `openshift_set_node_ip`), however that has led to other issues that prevented us from keeping it as the default. Mainly that when it is set, the node names are overridden by the value of nodeIp, even if a hostname is set in the node config.

Since the openstack playbooks are overriding the hostname with the IP anyway, you should be able to use `openshift_set_node_ip` instead of modifying /etc/hosts

lhuard1A:
@detiber I had a look at this `openshift_set_node_ip` feature but it seems to be specific to nodes whereas the issue I am facing is the master which tries to reverse resolve its own hostname.

So, the master fails to start before we even start provisioning the nodes.

I checked if there was a similar parameter for master and `masterIP` is already properly set:

```
[openshift@lenaic-master-0 ~]$ grep IP /etc/origin/master/master-config.yaml 
  masterIP: 192.168.205.6
```
It is in the `kubernetesMasterConfig` section.

What sounds strange to me is that the SDN is started on the master. I would have expected the SDN to be started only on nodes. Maybe is this coming from the `g_nodeonmaster: true` property ?

detiber:
@lhuard1A That variable is currently configuring the master as a node as well, which is required for accessing pods through the proxy api.

One other thing that I would be concerned about is that the resolvers on the node are injected into the pods, by switching to non-existent hostnames and using /etc/hosts may cause errors there.

I haven't had a chance to dig back in, but I believe that Neutron networking handles the dns names better than Nova networking, but I believe that is located in a different place in the metadata than the hostname values that we are currently using.  We have an item on our backlog to improve the autodetection under OpenStack, but it'll still be a bit before we can get to it.

lhuard1A:
Superseded by #1206 .

PR #883, opened 2015-11-12
Add support for Openstack integration

lebauce:

PR #1140, opened 2016-01-08
a quick fix for providing cloud-provider arguments

LutzLange:
It might not be beautiful, but it works. I'm open for suggestions on improving this.

I used these settings in my inventory. It would be nice to have an example config in the byo inventory directory as that is the playbook that was used to test this.

osm_controller_args={'cloud-provider': ['gce']}
osm_api_server_args={'cloud-provider': ['gce']}
openshift_node_kubelet_args={'cloud-provider': ['gce']}

### DIFF
--- a/roles/openshift_master/templates/master.yaml.v1.j2
+++ b/roles/openshift_master/templates/master.yaml.v1.j2
@@ -87,8 +87,10 @@ kubernetesMasterConfig:
   - v1beta3
   - v1
 {% endif %}
-  apiServerArguments: {{ openshift.master.api_server_args | default(None) | to_json }}
-  controllerArguments: {{ openshift.master.controller_args | default(None) | to_json }}
+  apiServerArguments: 
+    {{ openshift.master.api_server_args | default(None) | to_yaml }}
+  controllerArguments: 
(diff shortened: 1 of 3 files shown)


detiber:
The following PR has the start of what will become a generic base to build upon for multiple cloud providers: https://github.com/openshift/openshift-ansible/pull/883

The downside to that PR though, is that (at the last time I checked) the OpenStack provider did not work.

roles/openshift_master/templates/master.yaml.v1.j2
```diff
@@ -87,8 +87,10 @@ kubernetesMasterConfig:
   - v1beta3
   - v1
 {% endif %}
-  apiServerArguments: {{ openshift.master.api_server_args | default(None) | to_json }}
-  controllerArguments: {{ openshift.master.controller_args | default(None) | to_json }}
+  apiServerArguments: 
+    {{ openshift.master.api_server_args | default(None) | to_yaml }}
+  controllerArguments: 
+    {{ openshift.master.controller_args | default(None) | to_yaml }}
```
detiber: Have you tested this with controller_args that would expand further than one line? The main reason for embedding the json was to avoid having to deal with properly embedding the generated yaml.

PR #1143, opened 2016-01-08
atomic-openshift-installer: Populate new_nodes group

smunilla:
Set the new_nodes group when scaling up additional nodes

### DIFF
--- a/utils/src/ooinstall/cli_installer.py
+++ b/utils/src/ooinstall/cli_installer.py
@@ -72,7 +72,7 @@ def delete_hosts(hosts):
                 click.echo("\"{}\" doesn't coorespond to any valid input.".format(del_idx))
     return hosts, None
 
-def collect_hosts(oo_cfg, masters_set=False, print_summary=True):
+def collect_hosts(oo_cfg, existing_env=False, masters_set=False, print_summary=True):
     """
         Collect host information from user. This will later be filled in using
         ansible.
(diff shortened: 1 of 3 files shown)


utils/src/ooinstall/cli_installer.py
```diff
@@ -72,7 +72,7 @@ def delete_hosts(hosts):
                 click.echo("\"{}\" doesn't coorespond to any valid input.".format(del_idx))
     return hosts, None
 
-def collect_hosts(oo_cfg, masters_set=False, print_summary=True):
+def collect_hosts(oo_cfg, existing_env=False, masters_set=False, print_summary=True):
```
detiber: I almost think we need to maintain some type of state.

Basically, to start, we can have the following:
new_host
installed_host
failed_host

And update the state (and persist to the config file) as state changes for a host.
Then, we could have a proper way to re-run against a set of hosts (any that were in a failed_host state)

It'll also lay the ground work for supporting adding additional host types.

detiber: I should add that once we have a more intelligent callback plugin, then we could add additional states (or substates) based on where a failure happens to make decisions on whether or not a re-run should even be attempted.

smunilla: @detiber That's a good idea and not too hard to do instead. I'll start hacking on doing it that way.

detiber:
@brenton  :+1:

PR #1144, opened 2016-01-08
Remove default disable of SDN for GCE

detiber:
- Update openshift_facts to no longer default use_openshift_sdn  to disabled for GCE
  installs
- Remove workarounds in the GCE playbooks that were forcing SDN configuration

### DIFF
--- a/playbooks/gce/openshift-cluster/config.yml
+++ b/playbooks/gce/openshift-cluster/config.yml
@@ -13,4 +13,3 @@
     openshift_debug_level: "{{ debug_level }}"
     openshift_deployment_type: "{{ deployment_type }}"
     openshift_hostname: "{{ gce_private_ip }}"
-    openshift_use_openshift_sdn: "{{ do_we_use_openshift_sdn }}"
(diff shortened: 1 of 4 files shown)


detiber:
@twiest ptal

twiest:
@detiber Are we documenting what users should do instead of using the SDN?

I know that gce can have many IPs added to each host, but we should probably point that out to them somewhere and tell them how we expect it to be set up.

Also, how does the master know which IPs it can assign out to the PODs?

detiber:
@twiest my comment above was a bit misleading, previously we disabled the sdn for GCE, and there was no documentation around alternative approaches. This PR just removes the default disable and all installations will now default to using openshift_sdn.

When using an sdn plugin, the plugin provides the IPs to the master, so there is no need to configure the subnet directly in the node config.

twiest:
@detiber ah, ok, that makes sense.

twiest:
@menren Please review this change and make sure it doesn't break you guys.

menren:
I will check this ASAP

menren:
@twiest @detiber It looks good to me

detiber:
aos-ci-test

detiber:
@brenton ptal

twiest:
@detiber This looks good to me and I'm willing to merge it if you think it's ready.

Or, if you'd prefer to wait for @brenton , that's fine too.

brenton:
Sorry, I meant to chime in sooner.  LGTM.

PR #1159, opened 2016-01-11
API Verification

abutcher:
Wait until the API becomes available
* before starting the controllers service w/ native ha.
* before node registration.
* after api service restarts

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -313,6 +313,7 @@
 
 - name: Configure master instances
   hosts: oo_masters_to_config
+  any_errors_fatal: true
   serial: 1
   vars:
     sync_tmpdir: "{{ hostvars.localhost.g_master_mktemp.stdout }}"
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -269,6 +269,7 @@
   service: name={{ openshift.common.service_type }}-master enabled=yes state=started
   when: not openshift_master_ha | bool
   register: start_result
+  notify: Verify API Server
 
 - name: Stop and disable non HA master when running HA
   service: name={{ openshift.common.service_type }}-master enabled=no state=stopped
@@ -282,6 +283,7 @@
   service: name={{ openshift.common.service_type }}-master-api enabled=yes state=started
   when: openshift_master_ha | bool and openshift.master.cluster_method == 'native'
   register: start_result
+  notify: Verify API Server
 
 - set_fact:
     master_api_service_status_changed: "{{ start_result | changed }}"
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -287,6 +287,22 @@
     master_api_service_status_changed: "{{ start_result | changed }}"
   when: openshift_master_ha | bool and openshift.master.cluster_method == 'native'
 
+- name: Wait for API to become available
+  # Using curl here since the uri module requires python-httplib2 and
+  # wait_for port doesn't provide health information.
+  command: >
+    curl -k --head --silent {{ openshift.master.api_url }}
+  register: api_available_output
+  until: api_available_output.stdout.find("200 OK") != -1
+  retries: 120
+  delay: 1
+  changed_when: false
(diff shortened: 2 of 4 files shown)


detiber:
@sdodson would this solve your containerized problem if we trigger the test for non-ha installs as well?

It might require a bit of rework to make it so that the ordering is correct for ha and non-ha deployments, but I think it's worth it to notify the user of a failed api server before getting a failure from a later role (usually openshift_examples)

sdodson:
The containerization problem is unique to having to restart docker during node initialization which leads to up to a minute of master downtime while systemd starts the master services again after docker has been restarted. We need to fix both. Should I limit my fix to the node restarting docker in #1137 ?

detiber:
@sdodson hmm, I think considering they are slightly different issues, it probably wouldn't hurt to leave the additional check in place for #1137

sdodson:
:+1: to needing this and the other, LGTM

abutcher:
@wshearn is testing this w/ online

detiber:
:+1:

wshearn:
So the curl part is failing for me, it is like the api_available_output is not getting updated.

    TASK: [openshift_master | Wait for API to become available] ******************* 
    ok: [XXXXXXXXXX]

    TASK: [openshift_master | fail ] ********************************************** 
    failed: [XXXXXXXXXX] => {"failed": true}
    msg: Unable to contact master API at https://nope.com

But the ELB is saying InService around the time it fails.

detiber:
@wshearn how is the elb health check configured? If I'm remembering the default settings right 120s might time out before re-adding a previously failed host.

wshearn:
Ping Target	TCP:443
Timeout	5 seconds
Interval	30 seconds
Unhealthy Threshold	2
Healthy Threshold	2

And I bumped it up to 180 retries.

detiber:
how about increasing the retry interval, I believe I vaguely remember reading a bug about using 0 or 1 as the retry interval?

abutcher:
@wshearn the fail conditional was reversed. Updated and this is working for me locally.

abutcher:
aos-ci-test

wshearn:
:+1: works for me.

abutcher:
Addressing the nosetests failure in #1169

abutcher:
aos-ci-test

abutcher:
Melded @sdodson and my changes together and I'm still testing native ha with the combo.

abutcher:
aos-ci-test

abutcher:
@detiber @wshearn PTAL

roles/openshift_master/tasks/main.yml
```diff
@@ -283,6 +284,20 @@
...
+  # Using curl here since the uri module requires python-httplib2 and
+  # wait_for port doesn't provide health information.
+  command: >
+    curl -k --head --silent {{ openshift.master.api_url }}
+  register: api_available_output
+  until: api_available_output.stdout.find("200 OK") != -1
+  retries: 120
+  delay: 1
+  changed_when: false
+  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native'
```
detiber: Why not move this down below the set_fact and make it conditional on start_result?

playbooks/common/openshift-master/config.yml
```diff
@@ -313,6 +313,7 @@
 
 - name: Configure master instances
   hosts: oo_masters_to_config
+  any_errors_fatal: true
```
abutcher: Calling attention to this line. This will cause any failures in the `openshift_master` role to halt execution on all hosts. I think this makes sense to do for masters but a node failure doesn't seem fatal.

roles/openshift_master/tasks/main.yml
abutcher: Good idea

playbooks/common/openshift-master/config.yml
detiber: Agreed.

abutcher:
aos-ci-test

abutcher:
Closes #1137

abutcher:
@detiber Moved the wait below `set_fact`

detiber:
@brenton :+1:

PR #1160, opened 2016-01-11
Nuage

vishpat:
@detiber Changes made for Nuage plugin, please note this is still a work in progress.

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -336,6 +336,8 @@
   - role: nickhammond.logrotate
   - role: fluentd_master
     when: openshift.common.use_fluentd | bool
+  - role: nuage_master
+    when: openshift.common.use_nuage | bool
   post_tasks:
   - name: Create group for deployment type
     group_by: key=oo_masters_deployment_type_{{ openshift.common.deployment_type }}
--- a/roles/nuage_master/handlers/main.yaml
+++ b/roles/nuage_master/handlers/main.yaml
@@ -1,0 +1,18 @@
+---
+- name: restart nuagekubemon
+  sudo: true
+  service: name=nuagekubemon state=restarted
+
+- name: restart master
+  service: name={{ openshift.common.service_type }}-master state=restarted
+  when: (not openshift_master_ha | bool) and (not master_service_status_changed | default(false))
+
+- name: restart master api
+  service: name={{ openshift.common.service_type }}-master-api state=restarted
+  when: (openshift_master_ha | bool) and (not master_api_service_status_changed | default(false)) and openshift.master.cluster_method == 'native'
+
+# TODO: need to fix up ignore_errors here
+- name: restart master controllers
+  service: name={{ openshift.common.service_type }}-master-controllers state=restarted
+  when: (openshift_master_ha | bool) and (not master_controllers_service_status_changed | default(false)) and openshift.master.cluster_method == 'native'
+  ignore_errors: yes
--- a/roles/nuage_master/tasks/main.yaml
+++ b/roles/nuage_master/tasks/main.yaml
@@ -1,0 +1,34 @@
+---
+- name: Create directory /usr/share/nuagekubemon
+  sudo: true
+  file: path=/usr/share/nuagekubemon state=directory  
+
+- name: Create the log directory
+  sudo: true
+  file: path={{ nuagekubemon_log_dir }} state=directory
+
+- name: Install Nuage Kubemon
+  sudo: true
+  yum: name={{ nuage_kubemon_rpm }} state=present
+
+- name: Run the service account creation script
+  sudo: true
+  script: serviceaccount.sh --server={{ openshift.master.api_url }} --output-cert-dir={{ cert_output_dir }} --master-cert-dir={{ openshift_master_config_dir }}
+
+- name: Download the certs and keys
+  sudo: true
+  fetch: src={{ cert_output_dir }}/{{ item }} dest=/tmp/{{ item }} flat=yes
+  with_items:
+        - ca.crt
+        - nuage.crt
+        - nuage.key
+        - nuage.kubeconfig 
+ 
+- name: Create nuagekubemon.yaml
+  sudo: true
+  template: src=nuagekubemon.j2 dest=/usr/share/nuagekubemon/nuagekubemon.yaml owner=root mode=0644
+  notify:
+    - restart master
+    - restart master api
+    - restart master controllers
+    - restart nuagekubemon
--- a/roles/nuage_node/tasks/main.yaml
+++ b/roles/nuage_node/tasks/main.yaml
@@ -1,0 +1,37 @@
+---
+- name: Install Nuage VRS
+  sudo: true
+  yum: name={{ vrs_rpm }} state=present
+  
+- name: Set the uplink interface 
+  sudo: true
+  lineinfile: dest={{ vrs_config }} regexp=^NETWORK_UPLINK_INTF line='NETWORK_UPLINK_INTF={{ uplink_interface }}'
+
+- name: Set the Active Controller 
+  sudo: true
+  lineinfile: dest={{ vrs_config }} regexp=^ACTIVE_CONTROLLER line='ACTIVE_CONTROLLER={{ vsc_active_ip }}'
+
+- name: Set the Standby Controller 
+  sudo: true
+  lineinfile: dest={{ vrs_config }} regexp=^STANDBY_CONTROLLER line='STANDBY_CONTROLLER={{ vsc_standby_ip }}'
+  when: vsc_standby_ip is defined
+
+- name: Install plugin rpm
+  sudo: true
+  yum: name={{ plugin_rpm }} state=present
+
+- name: Copy the certificates and keys
+  sudo: true
+  copy: src="/tmp/{{ item }}" dest="{{ vsp_k8s_dir }}/{{ item }}"
+  with_items:
+        - ca.crt
+        - nuage.crt
+        - nuage.key
+        - nuage.kubeconfig 
+
+- name: Set the vsp-k8s.yaml 
+  sudo: true
+  template: src=vsp-k8s.j2 dest={{ vsp_k8s_yaml }} owner=root mode=0644 
+  notify:
+    - restart vrs
+    - restart node 
--- a/roles/nuage_node/vars/main.yaml
+++ b/roles/nuage_node/vars/main.yaml
@@ -1,0 +1,9 @@
+---
+vrs_config: /etc/default/openvswitch
+vsp_k8s_dir: /usr/share/vsp-k8s
+vsp_k8s_yaml: "{{ vsp_k8s_dir }}/vsp-k8s.yaml"
+client_cert: "{{ vsp_k8s_dir }}/nuage.crt"
+client_key: "{{ vsp_k8s_dir }}/nuage.key"
+ca_cert: "{{ vsp_k8s_dir }}/ca.crt"
+api_server: "{{ openshift_node_master_api_url }}"
+docker_bridge: "docker0"
--- a/roles/openshift_common/tasks/main.yml
+++ b/roles/openshift_common/tasks/main.yml
@@ -4,6 +4,14 @@
   when: openshift_use_openshift_sdn | default(false) | bool and openshift_use_flannel | default(false) | bool
 
 - fail:
+   msg: Nuage sdn can not be used with openshift sdn
+  when: openshift_use_openshift_sdn | default(false) | bool and openshift_use_nuage | default(false) | bool
+
+- fail:
+   msg: Nuage sdn can not be used with flannel 
+  when: openshift_use_flannel | default(false) | bool and openshift_use_nuage | default(false) | bool
+
+- fail:
     msg: openshift_hostname must be 64 characters or less
   when: openshift_hostname is defined and openshift_hostname | length > 64
 
@@ -22,6 +30,7 @@
       deployment_type: "{{ openshift_deployment_type }}"
       use_fluentd: "{{ openshift_use_fluentd | default(None) }}"
       use_flannel: "{{ openshift_use_flannel | default(None) }}"
+      use_nuage: "{{ openshift_use_nuage | default(None) }}"
       use_manageiq: "{{ openshift_use_manageiq | default(None) }}"
 
   # For enterprise versions < 3.1 and origin versions < 1.1 we want to set the
--- a/roles/openshift_master/templates/master.yaml.v1.j2
+++ b/roles/openshift_master/templates/master.yaml.v1.j2
@@ -102,7 +102,7 @@ masterPublicURL: {{ openshift.master.public_api_url }}
 networkConfig:
   clusterNetworkCIDR: {{ openshift.master.sdn_cluster_network_cidr }}
   hostSubnetLength: {{ openshift.master.sdn_host_subnet_length }}
-{% if openshift.common.use_openshift_sdn %}
+{% if openshift.common.use_openshift_sdn or openshift.common.use_nuage %}
   networkPluginName: {{ openshift.common.sdn_network_plugin_name }}
 {% endif %}
 # serviceNetworkCIDR must match kubernetesMasterConfig.servicesSubnet
--- a/roles/openshift_node/templates/node.yaml.v1.j2
+++ b/roles/openshift_node/templates/node.yaml.v1.j2
@@ -22,7 +22,7 @@ networkPluginName: {{ openshift.common.sdn_network_plugin_name }}
 # deprecates networkPluginName above. The two should match.
 networkConfig:
    mtu: {{ openshift.node.sdn_mtu }}
-{% if openshift.common.use_openshift_sdn %}
+{% if openshift.common.use_openshift_sdn or openshift.common.use_nuage %}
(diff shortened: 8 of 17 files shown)


detiber:
@vishpat thanks, I'll take a look a bit later and offer suggestions.

detiber:
@vishpat didn't quite get to it yesterday, but will do a review today.

vishpat:
@detiber I am still working on these changes, so there is no rush. take your time.

playbooks/common/openshift-master/config.yml
```diff
@@ -372,6 +372,12 @@
   - role: flannel_register
     when: openshift.common.use_flannel | bool
 
+- name: Configure Nuage 
+  hosts: oo_first_master
+  roles:
+  - role: nuage_master 
+    when: openshift.common.use_nuage | bool
```
detiber: Is there a reason this would only be installed on a single master? Ideally, the master component would be able to be installed on all masters to avoid a single point of failure.

roles/nuage_master/handlers/main.yaml
```diff
@@ -0,0 +1,12 @@
...
+  sudo: true
+  service: name=atomic-openshift-master state=restarted
+ 
+- name: restart nuagekubemon
+  sudo: true
+  service: name=nuagekubemon state=restarted
+
+- name: restart atomic-openshift-master
+  sudo: true
+  service: name=atomic-openshift-master state=restarted
```
detiber: Looks like there are two handlers here to restart atomic-openshift-master...  One thing to keep in mind is that on an HA installation, you would most likely want to restart the api and/or the controllers service instead of the master service.

roles/nuage_master/tasks/main.yaml
```diff
@@ -0,0 +1,35 @@
+---
+- name: Install Nuage Kubemon
+  sudo: true
+  yum: name={{ nuage_kubemon_rpm }} state=present
+
+- name: set kubeconfig 
+  sudo: true
+  lineinfile: dest={{ kubemon_yaml }} regexp=^kubeconfig line='kubeconfig{{':'}} {{ admin_config }}'
```
detiber: Ideally, we should use a specially created kubeconfig for this purpose. When we get a bit closer, I can work on getting you the commands needed to create a service account and add the appropriate permissions based on the api objects that need to be interacted with.

roles/nuage_master/tasks/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+- name: set openshift admin 
+  sudo: true
+  lineinfile: dest={{ kubemon_yaml }} regexp=^openshiftAdmin line='openshiftAdmin{{':'}} {{ os_admin }}'
+
+- name: Set Openshift master config path
+  sudo: true
+  lineinfile: dest={{ kubemon_yaml }} regexp=^openshiftMasterConfig line='openshiftMasterConfig{{':'}} {{ master_config_yaml }}'
```
detiber: What is the master config used for?

It seems odd to pass an external process the configuration file for another service...

roles/nuage_master/tasks/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+- name: Set enterprise 
+  sudo: true
+  lineinfile: dest={{ kubemon_yaml }} regexp=^enterpriseName line='enterpriseName{{':'}} {{ enterprise }}'
+
+- name: Set domain
+  sudo: true
+  lineinfile: dest={{ kubemon_yaml }} regexp=^domainName line='domainName{{':'}} {{ domain }}'
+  notify:
+    - restart master
```
detiber: I'm assuming here that we are restarting the master because the config is being modified. I'd rather the config changes be explicitly handled through the template in openshift-master, otherwise subsequent ansible runs will result in the master config showing as changed and restarting the services again.

roles/nuage_node/tasks/main.yaml
```diff
@@ -0,0 +1,52 @@
...
+- name: Set the master API server 
+  sudo: true
+  lineinfile: dest={{ vsp_k8s_yaml }} regexp=^masterApiServer line='masterApiServer{{':'}} {{ api_server }}'
+
+- name: Set the docker bridge name 
+  sudo: true
+  lineinfile: dest={{ vsp_k8s_yaml }} regexp=^dockerBridgeName line='dockerBridgeName{{':'}} {{ docker_bridge }}'
+  notify:
+    - restart vrs
+    - restart node
```
detiber: Is the node restart here simply to "pick up the underlying network changes"? Or is there deeper integration with the kubelet than appears from the tasks listed here?

roles/nuage_node/vars/main.yaml
```diff
@@ -0,0 +1,10 @@
+---
+vrs_config: /etc/default/openvswitch
+uplink_interface: "{{ ansible_default_ipv4.interface }}"
```
detiber: Ideally, we wouldn't assume a specific interface, but rather have it configurable by the user.

roles/nuage_node/vars/main.yaml
```diff
@@ -0,0 +1,10 @@
...
+vsp_k8s_yaml: /usr/share/vsp-k8s/vsp-k8s.yaml
+config_dir: "{{ openshift.common.config_base }}/node"
+client_cert: "{{ config_dir }}/system:node:{{ openshift.common.hostname }}.crt"
+client_key: "{{ config_dir }}/system:node:{{ openshift.common.hostname }}.key"
```
detiber: We really shouldn't be overloading the node cert/key here, but rather generating a new cert/key pair for use by the plugin.

detiber:
@vishpat overall, things look pretty good. I have some ideas on how to refactor a bit to avoid adding a add_nuage flag, but it definitely wouldn't be a blocker to merge.

If I get a chance, I may submit a PR to your branch with the suggested refactoring. I'm listing my initial thoughts below (but like I said, I wouldn't consider this to be a blocker):
- Add a new variable for openshift_sdn_plugin
  - Expose sdn variables as openshift_sdn_plugin_<plugin name>_<var_name>
  - Persist openshift_sdn_plugin and exposed variables to openshift_facts
    - openshift.common.sdn_plugin for the plugin name
    - Validate that it is one of: openshift-sdn, nuage, flannel, none
    - master variables for sdn configured under openshift.master
    - node variables for sdn configured under openshift.node
    - shared variables for sdn configured under openshift.common
- Move the master components of flannel and nuage networking into openshift_master (separate task files that are included based on the setting of openshift.common.sdn_plugin)
- Move the node components of flannel and nuage networking into openshift_node
- Handle doing the right thing when the old variables are specified, at least for a reasonable transition period.

vishpat:
@detiber Jason, many thanks for reviewing. Most of your review comments make sense to me and will address them. I am also adding my colleagues @rfredette27 and @abhat to have a look and address some of the comments.

playbooks/common/openshift-master/config.yml
abhat: Yes, the changes for the master should apply to all the master nodes in HA.

roles/nuage_master/tasks/main.yaml
abhat: @detiber agreed. The master config has the cluster's network specification. Our plugin parses the master config to get this information. Beyond that, it doesn't have any business to read the master config. I wasn't sure what the "rules" are to get the cluster configuration - I was being told by @rchopra to not rely on etcd to get the config information as it is an implementation artifact.

roles/nuage_node/vars/main.yaml
abhat: @rfredette I remember we had some problems reusing the service account we were creating to access the API server from the nodes. Or was it to access the local kubelet?

roles/nuage_master/tasks/main.yaml
detiber: @abhat correct, we wouldn't want to rely on direct access to etcd.

@abhat: You can use openshift.master.sdn_cluster_network_cidr and openshift.master.sdn_host_subnet_length to get the values from openshift_facts instead of querying the master config.

@rajatchopra Is there anything in OpenShift that uses the clusterNetworkCIDR, hostSubnetLength, and networkPluginName within OpenShift outside of openshift-sdn? Or can those be omitted in the case of third-party networking plugins?

roles/nuage_node/vars/main.yaml
detiber: Looking at the authorization policies doc, it looks like the system:sdn-manager and system:sdn-reader roles would probably be the roles you would need for the master and node components.

playbooks/common/openshift-master/config.yml
detiber: In that case, I would suggest moving this role up to the 'Configure master instances' play after the openshift-master role is applied.

roles/nuage_node/vars/main.yaml
rfredette: @abhat We weren't able to use the service account to access the local kubelet, but the service account was able to read from the API server just fine.

detiber: @rfredette which role was the service account given? I would think that the system:sdn-reader might just be able to, otherwise system:cluster-reader definitely should, since that is what heapster is using for gathering metrics on the nodes.

rfredette: The service account was given the `cluster-reader` role since it's being used to monitor creation and deletion of projects, services, and pods.

roles/nuage_master/tasks/main.yaml
abhat: @detiber, pardon the ignoramus. What is the nature of this openshift_facts API that you refer to - is it REST/something else? Note that our plugin on the master is not compiled inside of Openshift master today. Is the openshift_facts available at run-time or installation time?

abhat: If these parameters aren't embedded in Openshift per-se beyond for use by the openshift-sdn plugin, our third-party plugin can define its own perhaps. Having said that, one could make a case for "a consistent user-experience" and want the user facing parameters to be all in one place (the master-config.yaml as it stands today) instead of scattered around.

roles/nuage_node/vars/main.yaml
abhat: @detiber @rfredette I think the problem we had in using just the service account was potentially extracting the auth information from the kubeconfig to connect to the API server. The certificate information embedded within the service account's kubeconfig would have to be read?

roles/nuage_master/tasks/main.yaml
detiber: @abhat my apologies for not providing more context. I should have said the openshift_facts role. Within it, we provide an ansible module called openshift_facts that can be used for setting, persisting, and getting facts from the hosts. It would only be available during installation, reconfiguration (through ansible) and upgrade.

abhat: @detiber thanks - do we expect users to _only_ use ansible to "update" parameters in the master-config? If not, it begs the question as to what hooks are available for plugins to re-read this information?

detiber: I think it depends on the values. The OpenShift api can be used to query the SDN parameters as well.

roles/nuage_node/vars/main.yaml
detiber: The kubeconfig is the general method of supplying the entire config within Kubernetes and OpenShift, yes. But it also contains all of the other information needed for auth as well.

roles/nuage_master/tasks/main.yaml
detiber: @abhat it is not the only expected method of update, the preferred method of tracking changes to the SDN config in particular would be through the api list/watch mechanism.

playbooks/common/openshift-master/config.yml
vishpat: @detiber Done.

roles/nuage_master/handlers/main.yaml
vishpat: @detiber Done

roles/nuage_node/vars/main.yaml
vishpat: @detiber  Done

vishpat:
@detiber I have addressed all of the review comments and we have successfully tested a Nuage deployment with these changes as well. Can you have a look at the latest changes.

roles/openshift_common/tasks/main.yml
```diff
@@ -4,6 +4,10 @@
   when: openshift_use_openshift_sdn | default(false) | bool and openshift_use_flannel | default(false) | bool
 
 - fail:
+    msg: Nuage can not be used with openshift sdn
+  when: openshift_use_openshift_sdn | default(false) | bool and openshift_use_nuage | default(false) | bool
```
detiber: There should probably be another test here to ensure that openshift_use_nuage and openshift_use_flannel are both not set to True

roles/openshift_master/templates/master.yaml.v1.j2
```diff
@@ -102,7 +102,7 @@ masterPublicURL: {{ openshift.master.public_api_url }}
 networkConfig:
   clusterNetworkCIDR: {{ openshift.master.sdn_cluster_network_cidr }}
   hostSubnetLength: {{ openshift.master.sdn_host_subnet_length }}
-{% if openshift.common.use_openshift_sdn %}
+{% if openshift.common.use_openshift_sdn or openshift.common.use_nuage %}
   networkPluginName: {{ openshift.common.sdn_network_plugin_name }}
```
detiber: I don't see anything setting openshift.common.sdn_network_plugin_name when use_nuage is set.

Are there multiple plugin names that can be used with nuage, or is it just a single one that can be set when use_nuage is set to true?

roles/openshift_node/templates/node.yaml.v1.j2
```diff
@@ -15,7 +15,7 @@ kind: NodeConfig
 kubeletArguments: {{ openshift.node.kubelet_args | to_json }}
 {% endif %}
 masterKubeConfig: system:node:{{ openshift.common.hostname }}.kubeconfig
-{% if openshift.common.use_openshift_sdn %}
+{% if openshift.common.use_openshift_sdn or openshift.common.use_nuage %}
 networkPluginName: {{ openshift.common.sdn_network_plugin_name }}
```
detiber: Is networkPluginName still needed for nodes? I don't see anything in the templated config that references the node config file anymore? Or am I just missing where this value is being used.

roles/nuage_master/tasks/main.yaml
```diff
@@ -0,0 +1,34 @@
...
+  sudo: true
+  file: path={{ nuagekubemon_log_dir }} state=directory
+
+- name: Install Nuage Kubemon
+  sudo: true
+  yum: name={{ nuage_kubemon_rpm }} state=present
+
+- name: Run the service account creation script
+  sudo: true
+  script: serviceaccount.sh --server={{ openshift.master.api_url }} --output={{ cert_output_dir }} --master-dir={{ openshift_master_config_dir }}
```
detiber: Nice! I'm glad to see the service account was able to work.

detiber:
@vishpat overall, I think this is looking really good. There are some improvements that I'd like to see:
- the refactoring of the way the networking options are handled
- merging the way you are handling the service account creation and the way we are doing it in openshift_serviceaccounts
- transferring the service accounts from the master to the nodes

However I think those are all things that can be addressed as followup items (and not necessarily your responsibility to handle)

roles/openshift_common/tasks/main.yml
vishpat: Will do

roles/openshift_master/templates/master.yaml.v1.j2
vishpat: @detiber I am specifying the network plugin name in my ansible hosts file as follows

os_sdn_network_plugin_name='nuage/vsp-k8s'

roles/openshift_node/templates/node.yaml.v1.j2
abhat: Not required for the node config files. But perhaps still needed for master-config.yaml.

abhat: @vishpat can you please make the change and quickly test it out?

detiber: @abhat I'm guessing that is because there isn't a better way to query the master to see if it should be the active network plugin?

vishpat: No, I am reverting this change

vishpat:
@detiber I have addressed and tested the latest round of review comments.

detiber:
ok to test

detiber:
aos-ci-test

detiber:
@vispat if this is ready to go, can you squash your commits?

vishpat:
@detiber I think we are ready to merge. However before we do that can you have a look at the commit (https://github.com/vishpat/openshift-ansible/commit/a81ab12403a85edf897a1ee33b8bb5d9461e92c6) I made to this branch this morning. If you are ok with this change. I will squash the commits.

detiber:
@vishpat :+1: go ahead and squash.

vishpat:
@detiber Changes have been squashed.

detiber:
:+1: @brenton This PR adds support for configuring Nuage networking in place of openshift-sdn

Issue #1161, opened 2016-01-11
Generated cert invalid when changing 'openshift_master_portal_net'


timothysc:
We had run into an issue where the certificate that was generated did not allow pods to connect back to the internal kubernetes service endpoint.  

```
[OSEv3:vars]
openshift_master_portal_net=172.24.0.0/14
osm_cluster_network_cidr=172.20.0.0/14
osm_host_subnet_length=8
```

when running e2e-networking tests (https://github.com/openshift/origin/issues/6577#issuecomment-170023459) we uncovered:

``` 2016/01/08 14:34:25 Unable to get server version: Get https://172.24.0.1:443/version: x509: certificate is valid for 10.1.236.12, 172.30.0.1, 192.2.4.238, not 172.24.0.1.```

which basically means it used the default values of 173.30.0.0 and not the user specified one. 

/cc @liggitt @detiber @jayunit100

detiber:
@brenton ready to close now that #1162 merged

timothysc:
closing, thx for the quick fix!

PR #1162, opened 2016-01-11
Set portal net in master playbook

detiber:
Addresses https://github.com/openshift/openshift-ansible/issues/1161


detiber:
@sdodson or @abutcher mind taking a look?

sdodson:
@detiber confirmed that the cert has a non default ip when it should,  LGTM

detiber:
@brenton

detiber:
@timothysc

PR #1163, opened 2016-01-11
re-enable containerize installs

sdodson:
This uncomments the questions about rpm versus container based installation.


sdodson:
@smunilla PTAL I don't know if we wanted to change anything here before re-enabling it or not

smunilla:
:+1: 

That should do it. I wasn't really able to come up with a non-awkward way to ask this question since it has to be asked per host.

sdodson:
Ok, if we're keeping it as is then I guess it's ready to merge @brenton

PR #1165, opened 2016-01-11
use hostnamectl to set the hostname

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1280171

### DIFF
--- a/roles/openshift_common/tasks/main.yml
+++ b/roles/openshift_common/tasks/main.yml
@@ -38,5 +38,6 @@
     set_hostname_default: "{{ not openshift.common.version_greater_than_3_1_or_1_1 }}"
 
 - name: Set hostname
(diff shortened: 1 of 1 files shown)


sdodson:
@detiber you and I had discussed this but I think I misunderstood the problem when I mentioned it. re-reading the bug it looks like QE understands that you can't set the hostname to a name that doesn't resolve, but they have hosts where they'd like to set the hostname to a name that does resolve. This swaps out ansible hostname module for hostnamectl on atomic.

detiber:
@sdodson why not just change both to use hostnamectl?  All supported host targets we support would have it.

sdodson:
> @sdodson why not just change both to use hostnamectl? All supported host targets we support would have it.

Makes sense, done.

detiber:
:+1: @brenton

PR #1168, opened 2016-01-12
Fix for to_padded_yaml filter

detiber:
- Fix issue where None is passed to to_padded_yaml filter and invalid config
  is generated.


detiber:
aos-ci-test

detiber:
@brenton addresses an issue with the recently merged to_padded_yaml filter that was causing broken config when None is passed to the filter.

PR #1172, opened 2016-01-12
atomic-openshift-installer: Error handling for unicode hostnames

smunilla:
Fix error handling for invalid hostnames. Previously we were trying to print
the offending hostname out, which caused errors due to python's handling
of unicode strings. Seeing as how the user's input stays on the screen
directly above the error there is no need to try and print it and we can
avoid this problem entirely.

For https://bugzilla.redhat.com/show_bug.cgi?id=1284086


detiber:
pylint!

detiber:
@smunilla :+1:

PR #1173, opened 2016-01-12
Refactor storage options

abutcher:
* Adds `openshift_env` facts which will persist any inventory variable matching `/^openshift_*/`.
  * Ex: `openshift_hosted_registry_storage_kind=nfs` would be persisted as `facts['openshift']['hosted']['registry']['storage']['kind'] = 'nfs'`
  * Defaults for `facts['hosted'] = {...}` can be overridden without providing the entire dictionary. :fire: :fire: :fire: 
  * This will allow us to generically add `openshift_hosted_metrics_storage_kind=nfs`, etc.

* Adds `openshift_persistent_volumes` role.
  * Accepts a list of persistent volumes and persistent volume claims to create.

```
- name: Create persistent volumes/claims
  hosts: oo_first_master
  vars:
    persistent_volumes:
    - name: "registry-volume"
      capacity: "5Gi"
      access_modes:
      - "ReadWriteMany"
      storage:
        nfs:
          server: "nfs.example.com"
          path: "/var/exports/registry"
    persistent_volume_claims:
    - name: "registry-claim"
      capacity: "5Gi"
      access_modes:
      - "ReadWriteMany"
  roles:
  - role: openshift_persistent_volumes
```

* Registry storage now relies on `oo_openshift_env` facts and the `openshift_persistent_volumes` role.

Inventory vars w/ the `[nfs]` group:
```
[OSEv3:children]
...
nfs
...

[OSEv3:vars]
# _All_ of these will default to these values if an [nfs] group is provided.
openshift_hosted_registry_storage_kind=nfs
openshift_hosted_registry_storage_nfs_directory=/exports
openshift_hosted_registry_storage_nfs_options='*(rw,sync,all_squash)'
openshift_hosted_registry_storage_volume_name=registry
openshift_hosted_registry_storage_volume_size=5Gi

...

[nfs]
nfs.example.com
```

Inventory vars w/ pre-existing nfs host:
```
[OSEv3:vars]
openshift_hosted_registry_storage_kind=nfs
openshift_hosted_registry_storage_host=nfs.example.com
openshift_hosted_registry_storage_nfs_directory=/exports
openshift_hosted_registry_storage_volume_name=registry
openshift_hosted_registry_storage_volume_size=5Gi


```
@detiber

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -556,6 +556,96 @@ class FilterModule(object):
         except Exception as my_e:
             raise errors.AnsibleFilterError('Failed to convert: %s', my_e)
 
+    @staticmethod
+    def oo_openshift_env(hostvars):
+        ''' Return facts which begin with "openshift_"
+            Ex: hostvars = {'openshift_fact': 42,
+                            'theyre_taking_the_hobbits_to': 'isengard'}
+                returns  = {'openshift_fact': 42}
+        '''
+        if not issubclass(type(hostvars), dict):
+            raise errors.AnsibleFilterError("|failed expects hostvars is a dict")
+
+        facts = {}
+        regex = re.compile('^openshift_.*')
+        for key in hostvars:
+            if regex.match(key):
+                facts[key] = hostvars[key]
+        return facts
+
+    @staticmethod
+    # pylint: disable=too-many-branches
+    def oo_persistent_volumes(hostvars, groups, persistent_volumes=None):
+        """ Generate list of persistent volumes based on oo_openshift_env
+            storage options set in host variables.
+        """
+        if not issubclass(type(hostvars), dict):
+            raise errors.AnsibleFilterError("|failed expects hostvars is a dict")
+        if not issubclass(type(groups), dict):
+            raise errors.AnsibleFilterError("|failed expects groups is a dict")
+        if persistent_volumes != None and not issubclass(type(persistent_volumes), list):
+            raise errors.AnsibleFilterError("|failed expects persistent_volumes is a list")
+
+        if persistent_volumes == None:
+            persistent_volumes = []
+        for component in hostvars['openshift']['hosted']:
+            kind = hostvars['openshift']['hosted'][component]['storage']['kind']
+            create_pv = hostvars['openshift']['hosted'][component]['storage']['create_pv']
+            if kind != None and create_pv:
+                if kind == 'nfs':
+                    host = hostvars['openshift']['hosted'][component]['storage']['host']
+                    if host == None:
+                        if len(groups['oo_nfs_to_config']) > 0:
+                            host = groups['oo_nfs_to_config'][0]
+                        else:
+                            raise errors.AnsibleFilterError("|failed no storage host detected")
+                    directory = hostvars['openshift']['hosted'][component]['storage']['nfs']['directory']
+                    volume = hostvars['openshift']['hosted'][component]['storage']['volume']['name']
+                    path = directory + '/' + volume
+                    size = hostvars['openshift']['hosted'][component]['storage']['volume']['size']
+                    access_modes = hostvars['openshift']['hosted'][component]['storage']['access_modes']
+                    persistent_volume = dict(
+                        name="{0}-volume".format(volume),
+                        capacity=size,
+                        access_modes=access_modes,
+                        storage=dict(
+                            nfs=dict(
+                                server=host,
+                                path=path)))
+                    persistent_volumes.append(persistent_volume)
+                else:
+                    msg = "|failed invalid storage kind '{0}' for component '{1}'".format(
+                        kind,
+                        component)
+                    raise errors.AnsibleFilterError(msg)
+        return persistent_volumes
+
+    @staticmethod
+    def oo_persistent_volume_claims(hostvars, persistent_volume_claims=None):
+        """ Generate list of persistent volume claims based on oo_openshift_env
+            storage options set in host variables.
+        """
+        if not issubclass(type(hostvars), dict):
+            raise errors.AnsibleFilterError("|failed expects hostvars is a dict")
+        if persistent_volume_claims != None and not issubclass(type(persistent_volume_claims), list):
+            raise errors.AnsibleFilterError("|failed expects persistent_volume_claims is a list")
+
+        if persistent_volume_claims == None:
+            persistent_volume_claims = []
+        for component in hostvars['openshift']['hosted']:
+            kind = hostvars['openshift']['hosted'][component]['storage']['kind']
+            create_pv = hostvars['openshift']['hosted'][component]['storage']['create_pv']
+            if kind != None and create_pv:
+                volume = hostvars['openshift']['hosted'][component]['storage']['volume']['name']
+                size = hostvars['openshift']['hosted'][component]['storage']['volume']['size']
+                access_modes = hostvars['openshift']['hosted'][component]['storage']['access_modes']
+                persistent_volume_claim = dict(
+                    name="{0}-claim".format(volume),
+                    capacity=size,
+                    access_modes=access_modes)
+                persistent_volume_claims.append(persistent_volume_claim)
+        return persistent_volume_claims
+
     def filters(self):
         """ returns a mapping of filters to methods """
         return {
@@ -578,4 +668,7 @@ class FilterModule(object):
             "oo_generate_secret": self.oo_generate_secret,
             "to_padded_yaml": self.to_padded_yaml,
             "oo_nodes_with_label": self.oo_nodes_with_label,
+            "oo_openshift_env": self.oo_openshift_env,
+            "oo_persistent_volumes": self.oo_persistent_volumes,
+            "oo_persistent_volume_claims": self.oo_persistent_volume_claims,
         }
--- a/inventory/byo/hosts.aep.example
+++ b/inventory/byo/hosts.aep.example
@@ -203,6 +203,40 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # Configure dnsIP in the node config
 #openshift_dns_ip=172.30.0.1
 
+# Persistent Storage Options
+#
+## Registry Storage Options
+##
+## Storage Kind
+## Specifies which storage kind will be used for the registry.
+## "nfs" is the only supported kind at this time.
+##openshift_hosted_registry_storage_kind=nfs
+##
+## Storage Host
+## This variable can be used to identify a pre-existing storage host
+## if a storage host group corresponding to the storage kind (such as
+## [nfs]) is not specified,
+##openshift_hosted_registry_storage_host=nfs.example.com
+##
+## NFS Export Options
+##openshift_hosted_registry_storage_nfs_options='*(rw,root_squash)'
+##
+## NFS Export Directory
+## Specify the root exports directory. This directory will be created
+## if specifying an [nfs] host group.
+## This variable must be supplied if using a pre-existing nfs server.
+##openshift_hosted_registry_storage_nfs_directory=/exports
+##
+## Registry Volume Name
+## Specify the storage volume name. This directory will be created
+## within openshift_hosted_registry_storage_nfs_directory if
+## specifying an [nfs] group. Ex. /exports/registry
+## This variable must be supplied if using a pre-existing nfs server.
+##openshift_hosted_registry_storage_volume_name=registry
+##
+## Persistent Volume Access Mode
+##openshift_hosted_registry_storage_access_modes=['ReadWriteMany']
+
 # Configure node kubelet arguments
 #openshift_node_kubelet_args={'max-pods': ['40'], 'image-gc-high-threshold': ['90'], 'image-gc-low-threshold': ['80']}
 
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -53,6 +53,11 @@
           console_use_ssl: "{{ openshift_master_console_use_ssl | default(None) }}"
           public_console_url: "{{ openshift_master_public_console_url | default(None) }}"
           portal_net: "{{ openshift_master_portal_net | default(None) }}"
+  - openshift_facts:
+      role: hosted
+      openshift_env:
+        openshift_hosted_registry_storage_kind: 'nfs'
+    when: openshift_hosted_registry_storage_kind is not defined and groups.oo_nfs_to_config is defined and groups.oo_nfs_to_config | length > 0
   - name: Check status of external etcd certificatees
     stat:
       path: "{{ openshift.common.config_base }}/master/{{ item }}"
@@ -402,24 +407,20 @@
 
 - name: Configure service accounts
   hosts: oo_first_master
-
   vars:
     accounts: ["router", "registry"]
-
   roles:
   - openshift_serviceaccounts
 
-- name: Create services
+- name: Create persistent volumes and services
   hosts: oo_first_master
   vars:
-    attach_registry_volume: "{{ groups.oo_nfs_to_config | length > 0 }}"
-  pre_tasks:
-  - set_fact:
-      nfs_host: "{{ groups.oo_nfs_to_config.0 }}"
-      registry_volume_path: "{{ hostvars[groups.oo_nfs_to_config.0].openshift.nfs.exports_dir + '/' + hostvars[groups.oo_nfs_to_config.0].openshift.nfs.registry_volume }}"
-    when: attach_registry_volume | bool
+    persistent_volumes: "{{ hostvars[groups.oo_first_master.0] | oo_persistent_volumes(groups) }}"
+    persistent_volume_claims: "{{ hostvars[groups.oo_first_master.0] | oo_persistent_volume_claims }}"
   roles:
+  - role: openshift_persistent_volumes
+    when: persistent_volumes | length > 0 or persistent_volume_claims | length > 0
   - role: openshift_router
     when: openshift.master.infra_nodes is defined
   - role: openshift_registry
-    when: openshift.master.infra_nodes is defined and attach_registry_volume | bool
+    when: openshift.master.infra_nodes is defined and openshift.hosted.registry.storage.kind != None
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -957,12 +957,12 @@ def merge_facts(orig, new, additive_facts_to_overwrite):
                 # Fact is additive so we'll combine orig and new.
                 if isinstance(value, list) and isinstance(new[key], list):
                     new_fact = []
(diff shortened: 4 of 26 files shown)


playbooks/common/openshift-master/config.yml
```diff
@@ -38,6 +38,11 @@
           public_hostname: "{{ openshift_public_hostname | default(None) }}"
           public_ip: "{{ openshift_public_ip | default(None) }}"
           deployment_type: "{{ openshift_deployment_type }}"
+      - role: nfs
+        local_facts:
+          host: "{{ openshift_nfs_host | default(None) if openshift_nfs_host is defined else groups.oo_nfs_to_config.0 if groups.oo_nfs_to_config | length > 0 else None }}"
```
abutcher: This is gross.

inventory/byo/hosts.aep.example
```diff
@@ -180,6 +180,18 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
 # Configure dnsIP in the node config
 #openshift_dns_ip=172.30.0.1
 
+# NFS Options
```
detiber: We need to update this to Persistent storage options and have a subsection for NFS

playbooks/common/openshift-master/config.yml
detiber: Why not move some of this to the play vars?

In fact, it looks like the set_fact in pre_tasks can be moved to the play vars as well.

abutcher: Yep, looks like all of them can go into the play vars

detiber: The fact we should persist, should probably be something like this:
```
storage_providers:
- name: abc 
  type: nfs
  exports_dir: <some dir>
- name: def
  type: <blah>
  <some_attribute>: <some_value>
```
and maybe another fact for mapping storage providers to volumes to use:
```
storage_volumes:
- name: abc1
  provider_name: abc
- name: def1
  provider_name: def
```

detiber: Main goal is not to tie this specifically to NFS...  Eventually we want to be able to easily extend for additional storage providers.

roles/openshift_registry/tasks/main.yml
```diff
@@ -7,22 +7,10 @@
...
-    volume_already_attached: "{{ 'server:' + nfs_host in registry_volumes_output.stdout and 'path:' + registry_volume_path in registry_volumes_output.stdout }}"
-  when: attach_registry_volume | bool
-
-- name: Add nfs volume to dc/docker-registry
-  command: >
-    {{ openshift.common.client_binary }} volume dc/docker-registry
-    --add --overwrite --name=registry-storage --mount-path=/registry
-    --source='{"nfs": {"server": "{{ nfs_host }}", "path": "{{ registry_volume_path }}"}}'
-  when: attach_registry_volume | bool and not volume_already_attached | bool
+- include: nfs.yml
```
detiber: Would it be better to have a openshift_pvs or openshift_persistent_volumes role that handles the importing/creation of the pvs and then just have the openshift_registry (and eventually metrics, logging, etc) just use a consistent name for the created PV?

abutcher: Definitely, it would remove a lot of the duplication that would occur with this implementation

abutcher:
@detiber PTAL at the `openshift_persistent_volumes` role and the new filters when you get a chance.

filter_plugins/oo_filters.py
```diff
@@ -15,6 +15,9 @@
 import yaml
 from ansible.utils.unicode import to_unicode
 
+# Disable pylint too-many-public-methods; most if not all of the methods in this
+# filter module are public.
+# pylint: disable=too-many-public-methods
```
detiber: pylint! oh, how I loathe thee for rules like this.

filter_plugins/oo_filters.py
```diff
@@ -489,6 +492,78 @@ def to_padded_yaml(data, level=0, indent=2, **kw):
...
+    def oo_persistent_volume_claims(hostvars, persistent_volume_claims=None):
+        """ Generate list of persistent volume claims based on oo_openshift_env
+            storage options set in host variables.
+        """
+        if persistent_volume_claims == None:
+            persistent_volume_claims = []
+        if 'hosted' in hostvars['openshift']:
+            hosted_vars = hostvars['openshift']['hosted']
+            # Registry storage
+            if 'registry_storage_type' != None:
```
detiber: We'll need to take into account that s3 based storage (and other docker storage backends may not use PVCs at all).

I'm not sure if it should be an additional variable, or if we should have a mapping of storage_type to use_pvc?

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1102,9 +1104,9 @@ def __init__(self, role, filename, local_facts, additive_facts_to_overwrite=Fals
             )
         self.role = role
         self.system_facts = ansible_facts(module)
-        self.facts = self.generate_facts(local_facts, additive_facts_to_overwrite)
+        self.facts = self.generate_facts(local_facts, additive_facts_to_overwrite, openshift_env)
```
detiber: Instead of passing openshift_env all around, should we just do a self.openshift_env?

filter_plugins/oo_filters.py
abutcher: How about we check `registry_storage_type in ["nfs","whatever"]` (things we know will have pv/pvc) and we can have some other variables associated with `openshift_hosted_registry_storage_type="s3"`.

roles/openshift_facts/library/openshift_facts.py
abutcher: On it

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1187,11 +1189,12 @@ def get_defaults(self, roles):
...
-            nfs = dict(exports_dir='/var/export', registry_volume='regvol',
-                       export_options='*(rw,sync,all_squash)')
-            defaults['nfs'] = nfs
-
+        defaults['hosted'] = dict(registry_storage_type=None,
+                                  registry_storage_host=None,
+                                  registry_storage_volume_dir='/var/exports',
+                                  registry_storage_volume='registry',
+                                  registry_storage_volume_capacity='5Gi',
+                                  registry_storage_export_options='*(rw,sync,all_squash)')
```
detiber: Should we clean this up a bit by making the datastructure look more like this:
````
defaults['hosted'] = dict(
    registry=dict(
        storage=dict(
            type=None,
            volume=dict(
                name='registry',
                size='5Gi'
            ),
            nfs=dict(
                dir='/var/exports',
                options='*(rw,sync,all_squash)'),
                host=None,
                access_modes=[read_write_once, read_write_many, read_only_many],
                create_pv=True
            )
        )
    )
)
```                           
        

```

Comment (abutcher) [inline roles/openshift_facts/library/openshift_facts.py]: Most of our variables are in the form `openshift_$ROLE_$VARIABLE` and this would make them be `openshift_$ROLE_$COMPONENT_$VARIABLE`. I'm okay with it but it restricts what can be passed through `oo_openshift_env`

Comment (detiber) [inline roles/openshift_facts/library/openshift_facts.py]: Thinking a bit more about this...  we might want more of an object model here...

```
class StorageBase(object):
    def __init__(self, type):
        self.type = type


class PersistentVolume(StorageBase):
    def __init__(self, name, size, create_pv, access_modes)
        StorageBase(self, 'pv')
        self.name = name
        self.size = size
        self.create_pv = create_pv
        self.access_modes = access_modes
        
class NfsPersistentVolume(PersistentVolume):
    def __init__(self, name, size, host, dir, options):
        PersistentVolume(self, name, size, True, [read_write_once, read_write_many, read_only_many])
        self.host = host
        self.dir = dir
        self.options = options
```

Granted classes are a bit heavyweight here given that it would only exist in openshift_facts, but a structured dictionary that extends the idea is probably a good idea to make it easier to add additional types later without just creating a ton of new variables to toss around.

Comment (detiber) [inline roles/openshift_facts/library/openshift_facts.py]: Oh, and I'm not suggesting that we expose the dict out to the users in inventory (but we can and probably should do it as well, since it'll be easier to use in the installer wrapper config format), but rather how we use the internal representation of the variables after they have been generated by openshift_facts.

File: roles/openshift_persistent_volumes/tasks/main.yml [hunk 8]
```diff
@@ -0,0 +1,37 @@
...
+
+- name: Deploy PersistentVolume definitions
+  template:
+    dest: "{{ mktemp.stdout }}/persistent-volumes.yml"
+    src: persistent-volume.yml.j2
+  changed_when: False
+
+- name: Create PersistentVolumes
+  command: >
+    {{ openshift.common.client_binary }} create -f {{ mktemp.stdout }}/persistent-volumes.yml
```
detiber: We should get away from using the default kubeconfig..  since you are already creating a temporary directory for the volume defs, copy /etc/origin/master/admin.kubeconfig there and use the temp copy instead.

roles/openshift_facts/library/openshift_facts.py
abutcher: Well, the inventory variables for your structure coming through `oo_openshift_env` would be

```
openshift_hosted_registry_storage_type=
openshift_hosted_registry_storage_volume_name=
openshift_hosted_registry_storage_volume_size=
openshift_hosted_registry_storage_nfs_dir=
openshift_hosted_registry_storage_nfs_options=
openshift_hosted_registry_storage_nfs_host=
openshift_hosted_registry_storage_nfs_access_modes=
openshift_hosted_registry_storage_nfs_create_pv=
```

roles/openshift_registry/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+registry_volume: 'registry-volume'
```
detiber: Is this used anywhere now that the PVs and PVCs are handled in openshift_persistent_volumes?

roles/openshift_registry/templates/nfs/registry-claim.json.j2
```diff
@@ -0,0 +1,15 @@
...
+  },
+  "spec": {
+    "accessModes": [ "ReadWriteMany" ],
+    "resources": {
+      "requests": {
+        "storage": "{{ openshift.hosted.registry_storage_volume_capacity | default('5Gi') }}"
+      }
+    }
+  }
+}
```
detiber: left over from a previous iteration?

roles/openshift_registry/templates/nfs/registry-volume.json.j2
```diff
@@ -0,0 +1,17 @@
...
+    "capacity": {
+        "storage": "{{ openshift.hosted.registry_storage_volume_capacity | default('5Gi') }}"
+        },
+    "accessModes": [ "ReadWriteMany" ],
+    "nfs": {
+        "path": "{{ registry_storage_volume_path }}",
+        "server": "{{ registry_storage_host }}"
+    }
+  }
+}
```
detiber: left over from a previous iteration?

roles/openshift_registry/templates/nfs/registry-claim.json.j2
abutcher: Yep.

roles/openshift_registry/templates/nfs/registry-volume.json.j2
abutcher: Indeed.

roles/openshift_registry/defaults/main.yml
abutcher: It's not used.

roles/openshift_facts/library/openshift_facts.py
detiber: I actually like the idea of a '.' per '_', that way we could in theory support dict based config at any level, or allow setting specific variables on a case by case basis...

filter_plugins/oo_filters.py
abutcher: Note to use `...create_pv=true` or some similar flag.

roles/openshift_facts/library/openshift_facts.py
abutcher: I think it needs to be passed around since we need to merge `openshift_env` facts with `facts_to_set` based on discussion yesterday.

filter_plugins/oo_filters.py
```diff
@@ -491,6 +494,95 @@ def to_padded_yaml(data, level=0, indent=2, **kw):
...
+        for component in hostvars['openshift']['hosted']:
+            kind = hostvars['openshift']['hosted'][component]['storage']['kind']
+            create_pv = hostvars['openshift']['hosted'][component]['storage']['create_pv']
+            if kind != None and create_pv:
+                if kind == 'nfs':
+                    host = hostvars['openshift']['hosted'][component]['storage']['host']
+                    if host == None and len(groups['oo_nfs_to_config']) > 0:
+                        host = groups['oo_nfs_to_config'][0]
+                    else:
+                        raise errors.AnsibleFilterError("|failed no storage host detected")
```
detiber: Do we want to fail here? If they are specifying an existing NFS server and have pre-created the mounts shouldn't we still provision the PVs?

abutcher: No host was configured as an override and no `[nfs]` group was configured so we can't make the PV.

detiber: Ah, I see that now...  What if  the storage host was persisted in openshift_facts when oo_nfs_to_config > 0 prior to this filter being run?

abutcher: If storage host is persisted then it will take precedence over the nfs group.

detiber:
Only thing left from my point of view is to have an example of configuring an existing nfs export for use as router storage.

abutcher:
@detiber Inventory examples incoming

abutcher:
@detiber Added some inventory examples. LMK how those read. This section is going to get complicated as more storage options and components get thrown in.

abutcher:
aos-ci-test

detiber:
@abutcher inventory examples look good.

detiber:
@abutcher I don't seem to be able to actually mount the PVs with the nfsv4 config, the client keeps failing because rpc.statd is failing on the client...  we might need to expose more than just the nfsv4 port.

abutcher:
@detiber Alright, I'll add some configs / new ports.

abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
aos-ci-test

erinboyd:
For NFS and SELinux to work together in OSE, you will also have to:
setsebool -P virt_use_nfs 1

Also make sure each export has:
_/<example_fs> *(rw,root_squash)_
and IP tables is adjusted for the port:
_ iptables -I INPUT 1 -p tcp --dport 2049 -j ACCEPT

detiber:
@erinboyd We are already setting the SELinux boolean here: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_node/tasks/storage_plugins/nfs.yml

The plugin allows users to set the mount options: *(rw,sync,all_squash) is what is referenced currently. Wouldn't we need to ensure uid/gid configuration between the container and the nfs host if we use root_squash (if the container user is not root)?

@abutcher as an aside, should we configure the hosts that can mount the registry to limit mounting to the 'infra' nodes if available?

erinboyd:
@detiber all_squash is not longer the preferred method for the export. This was changed in 3.1 and I note has been added to the 3.0 document to indicate this. 

Yes, we would need to ensure the uid/gid configuration between the container and the NFS host. I believe there is a PR already completed to accomplish this @pmorie

abutcher:
@erinboyd Thanks, I'll update the export options to match.

@detiber Seems like a good idea. I'll add support for that and throw infra_nodes into the config.

roles/openshift_storage_nfs/handlers/main.yml
```diff
@@ -4,3 +4,8 @@
     name: nfs-server
     state: restarted
   when: not (nfs_service_status_changed | default(false))
+
+- name: restart nfs-config
+  service:
+    name: nfs-config
```
detiber: Looks like the handler may not be enough. If there was a failed nfs-server instance, then nfs-config needs to be run before service start of nfs-server in tasks/main.yml

detiber: Might want to just restart nfs-config in the playbook if the config changes rather than waiting for the trigger...

That or not start the nfs-server in tasks/main.yml and just enable it and let the restart handler handle starting the service.

detiber:
@abutcher after fixing my env issues I can confirm that only port 2049 is needed for the firewall config.

abutcher:
@detiber Sweet

abutcher:
Only outstanding bits are:
* Removing unnecessary configuration
* Configuring the registry export only for infra nodes
@detiber sound right?

roles/openshift_storage_nfs/tasks/main.yml
```diff
@@ -1,31 +1,41 @@
...
+  - regex: '^RPCNFSDARGS=.*$'
+    line: 'RPCNFSDARGS="-N 2 -N 3"'
+  - regex: '^RPCMOUNTDOPTS=.*$'
+    line: 'RPCMOUNTDOPTS="-p 20048"'
+  - regex: '^STATDARG=.*$'
+    line: 'STATDARG="-p 50825"'
+
+- name: Restart nfs-config
+  service: name=nfs-config state=restarted
+  when: "True in (nfs_config.results | oo_collect(attribute='changed') | list)"
```
detiber: Soes `nfs_config | changed` not work?

Also, the RPCMOUNTDOPTS and STATDARG should not be needed now

abutcher: It'll work when `nfs_config` isn't being registered in a loop.

abutcher: Or either way 8)

roles/openshift_storage_nfs/tasks/main.yml
```diff
@@ -45,5 +48,5 @@
 
 - set_fact:
     nfs_service_status_changed: "{{ True in (start_result.results
-                                    | map(attribute='changed')
+                                    | oo_collect(attribute='changed')
```
detiber:
```
 - set_fact:
      nfs_service_status_changed: "{{ start_result | changed }}"
```

detiber:
@abutcher outside of the service_status_changed test, :+1: from me

anpingli:
For pre-existing NFS host,  do we support RHEL 6.X now?

detiber:
@abutcher I say we ship this PR as is, and implement the restriction on the export with a followup PR.

abutcher:
@detiber Alright, I'm implementing that right now but I can add it in a followup. Kicking off a final tests run

abutcher:
aos-ci-test

detiber:
@brenton :shipit:

PR #1177, opened 2016-01-13
v3_0_to_v3_1_upgrade: Install storage packages

smunilla:
Include the existing storage plugin configuration during an upgrade.

For: https://bugzilla.redhat.com/show_bug.cgi?id=1281730

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
@@ -398,8 +398,25 @@
   - name: Ensure node service enabled
     service: name="{{ openshift.common.service_type }}-node" state=started enabled=yes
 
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
```diff
@@ -398,6 +398,9 @@
   - name: Ensure node service enabled
     service: name="{{ openshift.common.service_type }}-node" state=started enabled=yes
 
+  - name: Additional storage plugin configuration
+    include: ../../../../../roles/openshift_node/tasks/storage_plugins/main.yml
+
```
detiber: We probably shouldn't rely on a task file from a role within a playbook.

As much as I dislike duplication, I think it is actually warranted here.

detiber:
:+1:

PR #1182, opened 2016-01-13
Playbook for enabling origin-metrics

vnugent:
This PR automates the install steps outlined in origin-metrics [readme](https://github.com/openshift/origin-metrics/blob/master/README.adoc)

### DIFF
--- a/playbooks/adhoc/metrics_setup/playbooks/install.yml
+++ b/playbooks/adhoc/metrics_setup/playbooks/install.yml
@@ -1,0 +1,36 @@
+---
+- include: master_config_facts.yml
+- name: "Install metrics" 
+  hosts: masters
+  vars:
+     metrics_public_url: "https://{{ metrics_external_service }}/hawkular/metrics"
+  tasks:
+  - name: "Add metrics url to master config"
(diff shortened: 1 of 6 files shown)


detiber:
@vnugent would you mind refactoring this into an ansible role (similar to openshift_router or openshift_registry) instead?

playbooks/adhoc/metrics_setup/playbooks/install.yml
```diff
@@ -0,0 +1,36 @@
...
+- name: "Install metrics" 
+  hosts: masters
+  vars:
+     metrics_public_url: "https://{{ metrics_external_service }}/hawkular/metrics"
+  tasks:
+  - name: "Add metrics url to master config"
+    lineinfile: "state=present dest=/etc/origin/master/master-config.yaml regexp='^\ \ metricsPublicURL' insertbefore='^\ \ publicURL' line='\ \ metricsPublicURL: {{ metrics_public_url }}'"
+
+  - name: "Restart master service"
+    service: name=atomic-openshift-master state=restarted
```
chunyunchen: For HA master, the master service name is "atomic-openshift-master-api"

jeremyeder:
Comments from @detiber 

- it doesn't integrate with the existing roles for setting the master config
- I don't see a way to configure the SSL certificate for metrics
- it doesn't integrate with the pv/pvc creation we are doing for persistent storage.

detiber:
ok to test

detiber:
@brenton :+1: for adding this as an adhoc playbook until we have a chance to get this converted to a role.

PR #1185, opened 2016-01-14
Check API prior to starting node

abutcher:
This PR adds an API check prior to starting the node so that we can ensure the node will start properly.

I was able to reproduce [BZ#1293826](https://bugzilla.redhat.com/show_bug.cgi?id=1293826) prior to this change with a 1 master, 2 node containerized install and I'm getting happy nodes consistently with the additional API check.

Also removing a fail that will never occur since the previous task would have failed before we reach it.

@brenton @detiber PTAL

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -229,9 +229,5 @@
     delay: 1
     changed_when: false
     when: openshift.common.is_containerized | bool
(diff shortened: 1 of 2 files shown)


abutcher:
aos-ci-test

brenton:
LGTM

detiber:
:+1: ship it!

PR #1187, opened 2016-01-14
Update ec2.ini

detiber:
- Set rds and elasticache to False, since they are not needed for the
  playbooks and cause issues for users without those services enabled.


abutcher:
@brenton :+1:

PR #1190, opened 2016-01-15
Added anchors for rules in style_guide.adoc in order to make it easier to reference specific rules in PRs.

twiest:


joelddiaz:
:+1:

PR #1192, opened 2016-01-15
Add support for Openstack based persistent volumes

lebauce:
This PR adds an `openshift_storage_cinder` role similar to the `openshift_storage_nfs_lvm`. It creates Openstack Cinder volumes, formats them and creates associated PersistentVolume and PersistentVolumeClaim.

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -693,21 +693,22 @@ class FilterModule(object):
         if 'hosted' in hostvars['openshift']:
             for component in hostvars['openshift']['hosted']:
                 if 'storage' in hostvars['openshift']['hosted'][component]:
-                    kind = hostvars['openshift']['hosted'][component]['storage']['kind']
-                    create_pv = hostvars['openshift']['hosted'][component]['storage']['create_pv']
+                    params = hostvars['openshift']['hosted'][component]['storage']
+                    kind = params['kind']
+                    create_pv = params['create_pv']
                     if kind != None and create_pv:
                         if kind == 'nfs':
-                            host = hostvars['openshift']['hosted'][component]['storage']['host']
+                            host = params['host']
                             if host == None:
                                 if len(groups['oo_nfs_to_config']) > 0:
                                     host = groups['oo_nfs_to_config'][0]
                                 else:
                                     raise errors.AnsibleFilterError("|failed no storage host detected")
-                            directory = hostvars['openshift']['hosted'][component]['storage']['nfs']['directory']
-                            volume = hostvars['openshift']['hosted'][component]['storage']['volume']['name']
+                            directory = params['nfs']['directory']
+                            volume = params['volume']['name']
                             path = directory + '/' + volume
-                            size = hostvars['openshift']['hosted'][component]['storage']['volume']['size']
-                            access_modes = hostvars['openshift']['hosted'][component]['storage']['access_modes']
+                            size = params['volume']['size']
+                            access_modes = params['access_modes']
                             persistent_volume = dict(
                                 name="{0}-volume".format(volume),
                                 capacity=size,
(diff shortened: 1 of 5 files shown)


roles/openshift_storage_cinder/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Setup RDO repositories
+  yum: name=https://rdoproject.org/repos/rdo-release.rpm state=present
```
detiber: This would have to be conditional, we would only want to configure this repo when deployment_type == 'origin'

roles/openshift_storage_cinder/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
...
+
+- name: Install Nova and Cinder clients
+  yum: name=python-novaclient,python-cinderclient state=present
+
+- name: Create the cinder volume creation script
+  template:
+    dest: /usr/bin/create-openshift-cinder-volume
+    src: create-openshift-cinder-volume.j2
+    mode: 0755
```
detiber: Instead of using a script here, this seems like an Ansible module would be a better fit here. Modules can be embedded in roles by adding a `library` directory to the role.

Docs on creating an Ansible module: http://docs.ansible.com/ansible/developing_modules.html

detiber:
@abutcher mind syncing up with @lebauce on this so that we can incorporate this into the persistent storage for infrastructure containers you are doing?

abutcher:
@detiber Correct any of this if it sounds off.

@lebauce I'm still working on implementing this and we just started discussing a few days ago but, the basic idea is that we will use a set of variables for configuring which persistent storage providers are being used for what (registry, metrics, free floating storage etc). 

For example:
```
# [nfs]
# nfs.example.com

[ OSEv3:vars]
openshift_hosted_registry_storage_type=nfs
openshift_hosted_registry_storage_host=nfs.example.com # or {{ groups.nfs.0 }} by default
openshift_hosted_registry_storage_volume_dir=/var/exports
openshift_hosted_registry_storage_volume=registry
...
```

Note that `_hosted_` is kind of arbitrary and may change.

The variables will be processed a little bit differently in `openshift_facts` such that these will get automatically loaded into `facts['hosted']['registry_storage_type'] = 'nfs'` (or something near that) without needing to set them individually within `openshift_facts`.

Each of the various component roles like `openshift_registry` would be able to determine if storage is being used as well as the type of the storage and then the role can get anything else it needs for that type and attach it to the registry. This part will be tricky but I can't see us having more than a handful of storage options.

roles/openshift_storage_cinder/tasks/main.yml
lebauce: Done

roles/openshift_storage_cinder/tasks/main.yml
lebauce: Done

lebauce:
@abutcher Thanks for this work and for the explanation. Is there something I can do in my PR to help ?

abutcher:
@lebauce I think it would be great if we could use the same variable convention so that we could extend `openshift_registry` and eventually metrics to use cinder volumes. I'm not sure what that will look like yet, but my refactor PR is https://github.com/openshift/openshift-ansible/pull/1173.

detiber:
@lebauce the variables in the pr by @abutcher should be in their final form now, we are still working through some final fixes and tweaks for the openshift_fact changes and nfs config though.

lebauce:
@detiber @abutcher I reduced the scope of this PR.

With this PR, you can specify `openshift_hosted_registry_storage_kind: openstack` to deploy a registry using Openstack/Cinder volumes.
You need to specify the ID of the volume in `openshift_hosted_registry_storage_openstack_volumeID` and the filesystem to use with `openshift_hosted_registry_storage_openstack_fs` (defaults to ext4)

detiber:
ok to test

detiber:
aos-ci-test

detiber:
@herlo are these failures known issues?

herlo:
@detiber Not sure. I was running into this issue earlier today on some internal openstack clusters, but they seem to have resolved themselves in the last few minutes.

detiber:
@herlo thanks!

detiber:
aos-ci-test

detiber:
@lebauce pylint isn't a fan of two letter vars (`fs` in this case).

detiber:
It looks like a rebase will be needed as well.

lebauce:
@detiber Done

detiber:
aos-ci-test

detiber:
@abutcher ptal

lebauce:
ping @detiber @abutcher :-)

detiber:
@lebauce it looks like the `openshift_storage_cinder` is missing from the PR

lebauce:
@detiber Indeed, I reduced the scope of this PR. The Cinder volumes need to be created by the administrator. It leverages the awesome @abutcher work on the registry to create the persistent volume and the persistent volume claims.

detiber:
@lebauce ah, makes sense. Could you add some examples of the usage to the inventory/byo/hosts.*.example files?

I think we will want to add on the ability to provision the volumes in OpenStack for users as well (in the future), but using pre-created volumes is sorely needed.

lebauce:
@detiber Done.

I do agree, we still want the ability to automatically provision the volume.

detiber:
aos-ci-tes

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
aos-ci-test

PR #1193, opened 2016-01-15
Configure node-masters before nodes in containerized install

abutcher:
Nodes which are also masters are configured serially before nodes are configured to avoid issues with docker restarting and taking down the API while nodes are starting.

https://bugzilla.redhat.com/show_bug.cgi?id=1293826

@detiber PTAL

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -154,21 +154,15 @@
       validate_checksum: yes
     with_items: nodes_needing_certs
 
-- name: Configure node instances
+- name: Deploy node certificates
   hosts: oo_nodes_to_config
   vars:
     sync_tmpdir: "{{ hostvars.localhost.mktemp.stdout }}"
-    openshift_node_master_api_url: "{{ hostvars[groups.oo_first_master.0].openshift.master.api_url }}"
-    # TODO: Prefix flannel role variables.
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-node/config.yml
```diff
@@ -177,8 +171,44 @@
...
+      ansible_sudo: "{{ g_sudo | default(omit) }}"
+    with_items: "{{ groups.oo_nodes_to_config | default([]) }}"
+    when: hostvars[item].openshift.common.is_containerized | bool and (item in groups.oo_nodes_to_config and item in groups.oo_masters_to_config)
+
+- name: Configure node instances
+  hosts: oo_containerized_master_nodes
+  serial: 1
+  vars:
+    openshift_node_master_api_url: "{{ hostvars[groups.oo_first_master.0].openshift.master.api_url }}"
+    openshift_node_first_master_ip: "{{ hostvars[groups.oo_first_master.0].openshift.common.ip }}"
```
detiber: You can actually set these vars in the add_host task above and not have to specify it separately for the Configure plays.

abutcher: The group is only for containerized master/node combo hosts so wouldn't they need to be set separately?

detiber: doh, you are correct...  carry on...  too much multi-tasking

abutcher:
aos-ci-test

detiber:
:+1: @brenton

PR #1107, opened 2016-01-04
Group name as per hosts.origin.example

donovanmuller:
The example inventory file does not currently work and fails with:

```
PLAY [Gather Cluster facts] ***************************************************
skipping: no hosts matched

PLAY RECAP ********************************************************************
```

this is because the group name is actually `OSEv3` (inline with [hosts.origin.example](https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.origin.example#L4)) and not `OSv3`.

Also added reference to inventory/byo/hosts.origin.example for more complete example.
Perhaps the basic example above should be removed and replaced with a reference to [inventory/byo/hosts.origin.example](https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.origin.example) instead?

PR #1197, opened 2016-01-18
Host group should be OSEv3 not OSv3

donovanmuller:
Following on from #1107 the host group's name is [OSEv3](https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.origin.example#L4) and not OSv3


abutcher:
@brenton

PR #1201, opened 2016-01-18
Use IdentityFile instead of PrivateKey

### LINKED ISSUE #1196: README_AWS refers to PrivateKey, which seems to not exist
The paragraph on "[(Optional) Setup your $HOME/.ssh/config file](https://github.com/openshift/openshift-ansible/blob/master/README_AWS.md#optional-setup-your-homesshconfig-file)" refers to adding the `PrivateKey` entry in your `.ssh/config`.

However, the `PrivateKey` option does not seem to [exist](http://linux.die.net/man/5/ssh_config) and fails with the following error when running the "Wait for user steup" task:

```
stderr: /Users/xxx/.ssh/config: line 3: Bad configuration option: privatekey
/Users/xxx/.ssh/config: terminating, 1 bad configuration options
```

I believe this should be `IdentityFile` instead.
I.e.

```
Host *.compute-1.amazonaws.com
  IdentityFile $HOME/.ssh/my_private_key.pem
```

Happy to make the change to the docs, just wanted to get confirmation?

donovanmuller:
Fixes #1196

### DIFF
--- a/README_AWS.md
+++ b/README_AWS.md
@@ -51,7 +51,7 @@ to setup a private key file to allow ansible to connect to the created hosts.
 To do so, add the the following entry to your $HOME/.ssh/config file and make it point to the private key file which allows you to login on AWS.
 ```
 Host *.compute-1.amazonaws.com
-  PrivateKey $HOME/.ssh/my_private_key.pem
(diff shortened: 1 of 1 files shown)


abutcher:
@brenton

PR #1206, opened 2016-01-18
Add a DNS server on OpenStack clusters

lhuard1A:
On OpenStack, VMs are assigned a name by Nova.
But in Kilo and previous versions of OpenStack, the network provided by Neutron doesn’t embed a DNS server. As a consequence, the VMs names are not resolvable by DNS from other VMs.

Having the VMs name resolvable by DNS from the other VMs is a feature of Liberty:
* https://blueprints.launchpad.net/neutron/+spec/internal-dns-resolution
* https://specs.openstack.org/openstack/neutron-specs/specs/liberty/internal-dns-resolution.html

For older versions of OpenStack, we have to live with the fact that VMs names are not simply resolvable.

We use to workaround this limitation in the openstack provider by overriding the hostname with their IP:
https://github.com/openshift/openshift-ansible/blob/master/playbooks/openstack/openshift-cluster/config.yml#L13
It is not a very nice solution:
* The output of `oc get nodes` shows IPs instead of meaningful node names.
* The machines are booted with their meaningful names (for ex.: `clusterid-master-0` (their domain being `novalocal`) and when they are rebooted, their hostname change. All the machines of the cluster have all the same hostname (`192`) and their are all in different domains (`168.0.1`, `168.0.2`, etc.). This clearly shows that we put the wrong data at the wrong place.
* In addition, this workaround won’t play nice with the activation of cloud provider (#1207) where the OpenShift cloud provider for OpenStack will check the nodeName against the Nova VM names.

In #1139, @detiber, you suggested that I should use `openshift_set_node_ip`, but it was not working well without the above mentioned trick. Would you have any pointer ? Is the issue in OpenShift itself ? Can it be worked around in openshift-ansible ?
Moreover, I tried to both override the hostnames with IPs and enable `openshift_set_node_ip`, but I still have the issue that the ovs sdn plugin on the master tries to resolve its hostname as described in #1139.

In order to have a chance to be able to enable the OpenStack cloud provider, we cannot override the hostnames.

This PR proposes to start a DNS server on a dedicated VM (named `<clusterid>-dns`) fed with the nova names of the VMs with their IPs in order to get rid of the hostname override trick.

All the other VMs are using `<clusterid>-dns` as DNS server. It acts as an authoritative server for the `.novalocal` and `.openstacklocal` domains and as a cache forwarder for everything else.

### DIFF
--- a/roles/dns/README.md
+++ b/roles/dns/README.md
@@ -1,0 +1,43 @@
+dns
+===
+
+Configure a DNS server serving IPs of all the nodes of the cluster
+
+Requirements
+------------
+
+None
+
+Role Variables
+--------------
+
+| Name | Mandatory / Optional | Description |
+|------|----------------------|-------------|
+| `dns_zones` | Mandatory | DNS zones in which we must find the hosts |
+| `dns_forwarders` | If not set, the DNS will be a recursive non-forwarding DNS server | DNS forwarders to delegate the requests for hosts outside of `dns_zones` |
+| `dns_all_hosts` | Mandatory | Exhaustive list of hosts |
+
+Dependencies
+------------
+
+None
+
+Example Playbook
+----------------
+
+    - hosts: dns_hosts
+      roles:
+      - role: dns
+        dns_forwarders: [ '8.8.8.8', '8.8.4.4' ]
+        dns_zones: [ novalocal, openstacklocal ]
+        dns_all_hosts: "{{ g_all_hosts }}"
+
+License
+-------
+
+ASL 2.0
+
+Author Information
+------------------
+
+OpenShift operations, Red Hat, Inc
--- a/roles/dns/tasks/main.yml
+++ b/roles/dns/tasks/main.yml
@@ -1,0 +1,22 @@
+- name: Install Bind
+  action: "{{ ansible_pkg_mgr }} name=bind"
+
+- name: Configure Bind
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    validate: "{{ item.validate }}"
+  with_items:
+    - src: openshift-cluster.zone
+      dest: /var/named/openshift-cluster.zone
+      validate: "named-checkzone {{ dns_zones[0] }} %s"
+    - src: named.conf
+      dest: /etc/named.conf
+      validate: "named-checkconf %s"
+  notify: restart bind
+
+- name: Enable Bind
+  service:
+    name: named
+    state: started
+    enabled: yes
--- a/playbooks/openstack/openshift-cluster/config.yml
+++ b/playbooks/openstack/openshift-cluster/config.yml
@@ -10,7 +10,6 @@
     openshift_cluster_id: "{{ cluster_id }}"
     openshift_debug_level: "{{ debug_level }}"
     openshift_deployment_type: "{{ deployment_type }}"
-    openshift_hostname: "{{ ansible_default_ipv4.address }}"
     openshift_registry_selector: 'type=infra'
     openshift_router_selector: 'type=infra'
     openshift_infra_nodes: "{{ g_infra_hosts }}"
--- a/playbooks/openstack/openshift-cluster/dns.yml
+++ b/playbooks/openstack/openshift-cluster/dns.yml
@@ -1,0 +1,47 @@
+- name: Populate oo_dns_hosts_to_update group
+  hosts: localhost
+  connection: local
+  become: no
+  gather_facts: no
+  vars_files:
+  - vars.yml
+  - cluster_hosts.yml
+  tasks:
+  - name: Evaluate oo_dns_hosts_to_update
+    add_host:
+      name: "{{ item }}"
+      groups: oo_dns_hosts_to_update
+      ansible_ssh_user: "{{ deployment_vars[deployment_type].ssh_user }}"
+      ansible_sudo: "{{ deployment_vars[deployment_type].sudo }}"
+    with_items: "{{ groups[cluster_id ~ '-dns'] }}"
+
+  - name: Evaluate oo_hosts_to_add_in_dns
+    add_host:
+      name: "{{ item }}"
+      groups: oo_hosts_to_add_in_dns
+      ansible_ssh_user: "{{ deployment_vars[deployment_type].ssh_user }}"
+      ansible_sudo: "{{ deployment_vars[deployment_type].sudo }}"
+    with_items: "{{ groups['tag_clusterid_' ~ cluster_id] }}"
+
+- name: Gather facts
+  hosts: oo_hosts_to_add_in_dns
+  vars_files:
+  - vars.yml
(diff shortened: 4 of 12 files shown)


lhuard1A:
@detiber: Could you please give me you thoughts about that?

detiber:
@lhuard1A notifications for this and the associated pr got buried in my email and I just dug them out.

I'll take a closer look at this today, I suspect there may be quite a few assumptions being made on the OpenStack cloud provider side about which metadata values to use and they may not align with the ones in your deployment.

Would it be possible to provide the output for:
```
 curl http://169.254.169.254/openstack/latest/meta_data.json
```
from an instance?

lhuard1A:
Hi @detiber,

Thanks for having a look!
Here is what I have when `playbooks/openstack/openshift-cluster/config.yml` contains:

* neither `openshift_hostname: "{{ ansible_default_ipv4.address }}"`
* nor `openshift_set_node_ip: True`

and there is a DNS configured as in this PR.

This is the state I observe *with this PR merged*:

```
[openshift@lenaic-master-0 ~]$ oc get nodes
```
```
NAME                              LABELS                                                   STATUS                     AGE
lenaic-master-0.novalocal         kubernetes.io/hostname=lenaic-master-0.novalocal         Ready,SchedulingDisabled   30m
lenaic-node-compute-0.novalocal   kubernetes.io/hostname=lenaic-node-compute-0.novalocal   Ready                      30m
lenaic-node-compute-1.novalocal   kubernetes.io/hostname=lenaic-node-compute-1.novalocal   Ready                      30m
```

```
curl http://169.254.169.254/openstack/latest/meta_data.json | jq
```
```json
{
  "random_seed": "LxoXav2+FUKGYzf5V5GlnBigr+dojBXjJ/11b/YFznbkZkwN6D74UyVoe/B3rgBrmI0BU5nrNTNv5Te3I/ZqZITh9olc1ehbVqgaTqz/2ervQKImbWMQhHxOZ0sl3TMkot6kef+pkTxdx+stMPNF443bHYDsiFLGdJV2qbpz7cNJbwg2WvpLGYTHtdp5Oad+2LlEOhS7vlpKZPFoQ7co8eYa+EMwHZRa7OA5a7jtaU7L1p0Q3VYb0amdwu43USjAHtkhHvCkkal6XlTqcTKNVf1nkDaUJ7J2TkqvB/hV+g2+zwHVhKQ1HAb2c1mRoOtK7uweLxfzz8zie0s47lmdGCRB02wYo3M0+rlQikFuiTIUO9Z7JwT7RTGUQ5lYmS+jkDABDNlpdpbCxa/+fHYDnhMk2wi3Xmh6c9wChKvE3NoGOx+59W9SDQlbVdGDH6jNaurmEC5DCU5Viv8SYg52dNGnZP21aovetkAuEFZLjYkp2lZ7Fa1bplZvNOiiMne3OEg3ac4wl8k0Ivbx4cjb5HHCVLHQn6lOSiRAyAtZ7NfjTKBS4sBs360DRDcTx6ZlffDy+yr8a9tm5BPuCrwdX7xNHMaWQMjuGPPCYWsS3qnKrGaBjX87tFetTXLqz5QCjZGRFIyjLjzE8YD8R0GlsauZvfKyzIpJ4qu83z+ccYo=",
  "uuid": "db336ea6-b4d0-425c-9640-488f6f6ed991",
  "availability_zone": "nova",
  "hostname": "lenaic-master-0.novalocal",
  "launch_index": 0,
  "meta": {
    "environment": "dev",
    "clusterid": "lenaic",
    "host-type": "master",
    "sub-host-type": "default"
  },
  "public_keys": {
    "openshift-ansible-lenaic-keypair": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDKOUSztBJFHoNKpvJBVt+oHepl3iv6kIgOMirbxrHPNkY2lcLIJBPUPLEXwfJf4GlgPw5GLgAGBdHbqxXH2I+rnQEnDxpduNKqDD7FhUpwGnfdZ3DL8CjixxPkAlYK1QH3jS6NItKnoAP07WtH8JXSNtjUESfCZrqMLymppJYUOwe3bZS4AUR2ZcKH87yVPlkYv1rqLgMk69SfDepuBcovEf6CITRJkNcvLbNLYkUjjLg1V3jZEHlgp6IhiHw0NjNJBBTcds6M10dN2195XKypKjhULHwGIZ1AAb283ZxqVHm/5SeT5tzrpjcyzNPNmq+dsotroHiIpUKC+cp4vPav lenaic@ncelhuard-Xperiment"
  },
  "name": "lenaic-master-0"
}
```

roles/dns/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+- name: Install Bind
+  action: "{{ ansible_pkg_mgr }} name=bind"
+
+- name: Configure Bind
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    validate: "{{ item.validate }}"
+  with_items:
+    - src: openstack.zone
```
detiber: I can see this role as being more generically applicable than just for openstack, could you add variables in place of the openstack specific references?

detiber:
@lhuard1A can you also add metadata and a README for the dns role?

detiber:
@lhuard1A Looks like a merge conflict popped up. Once that is resolved, I think this is good to go.

lhuard1A:
Hi @detiber,
I have just rebased this PR.
Readme and metadata have been added. Let me know if you have other remarks.

detiber:
@lhuard1A looks good to me, thanks!

When we get ready to do the generic dns config, we'll have some updates to make it a bit more flexible, but we'll take care of that when the time comes.

@twiest ptal

roles/dns/README.md
```diff
@@ -0,0 +1,42 @@
...
+None
+
+Role Variables
+--------------
+
+| Name | Mandatory / Optional | Description |
+|------|----------------------|-------------|
+| `dns_zones` | Mandatory | DNS zones in which we must find the hosts |
+| `dns_forwarders` | If not set, the DNS will be a recursive non-forwarding DNS server | DNS forwarders to delegate the requests for hosts outside of `dns_zones` |
+| `g_all_hosts` | Mandatory | Exhaustive list of hosts |
```
twiest: We try not to use global variables directly in roles.

Instead, we prefer to pass in the global variable as a role variable.

So, something like this (untested):
```yaml
    - hosts: dns_hosts
      roles:
      - role: dns
        dns_forwarders: [ '8.8.8.8', '8.8.4.4' ]
        dns_zones: [ novalocal, openstacklocal ]
        dns_all_hosts: "{{ g_all_hosts }}"
```

That way we can pass in whatever we want to the role (makes it more generic).

lhuard1A: Thanks for the review @twiest.

I’ve created a dedicated `dns_all_hosts` role variable.

twiest:
:+1:

Issue #1106, opened 2015-12-30
Added option for resource-quota-sync-period 


jkroepke:
Hi,

if you want to use quotas, you have to set resource-quota-sync-period on kubernetesMasterConfig/controllerArguments (Documentation: https://docs.openshift.com/enterprise/3.1/dev_guide/quota.html#configuring_quota_sync_period).

It should be configurable via ansible.

abutcher:
Hey @jkroepke this will be configurable via `osm_controller_args` once #1110 merges. I added some examples to the byo inventory files in the PR as well.

PR #1207, opened 2016-01-18
[WIP] Add OpenStack cloud provider

lhuard1A:
Enable the OpenStack cloud provider as described on:
https://docs.openshift.org/latest/install_config/configuring_openstack.html

In order for the cloud provider to work, the `nodeName` of the nodes has to match their name in Nova, otherwise, we get the following error:
```
Jan 18 14:02:02 lenaic-node-infra-0.novalocal atomic-openshift-node[20039]: E0118 14:02:02.708995   20039 kubelet.go:925] Unable to construct api.Node object for kubelet: failed to get external ID from cloud provider: Failed to find object
```
That’s why #1206 is a prerequisite to this PR.

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -14,6 +14,7 @@ import re
 import json
 import yaml
 from ansible.utils.unicode import to_unicode
+import jinja2
 
 class FilterModule(object):
     ''' Custom ansible filters '''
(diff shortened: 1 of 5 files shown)


lhuard1A:
Even with #1106, things are not working because the `nodeName` of a node is, for ex., `<clusterid>-node-infra-0.novalocal` whereas the name retrieved by the OpenStack cloud provider is just `<clusterid>-node-infra-0`.
This leads to the following error:
```
Jan 18 14:02:02 lenaic-node-infra-0.novalocal atomic-openshift-node[20039]: E0118 14:02:02.708995   20039 kubelet.go:925] Unable to construct api.Node object for kubelet: failed to get external ID from cloud provider: Failed to find object
```

If I manually override the `nodeName` in `/etc/origin/node/node-config.yaml` to `<clusterid>-node-infra-0`, the above error disappear, but I still have errors saying:
```
common.go:577] Could not find an allocated subnet for node: lenaic-node-infra-0.novalocal, Waiting...
```

@detiber: Would you know where we expect a hostname without domain and where we expect a fqdn? Or any other idea about how I could fix this?

lebauce:
@lhuard1A There is already a PR for this : https://github.com/openshift/openshift-ansible/pull/883, though it may missing the load balancer bit. I'll add it to the existing PR if it's ok for you

theone4ever:
@lhuard1A and @lebauce , is there any plan to include this fix? I really want to use openstack cinder as persistence storage. Btw, i also tried to add another kubernetes arguments "hostname_override" (the value is the hostname without the postfix ".novalocal" in /etc/origin/node/node-config.yaml,  but it doesn't seem to make any difference.

detiber:
@theone4ever we are still working through the final testing/changes for adding an OpenStack cloud provider config here: https://github.com/openshift/openshift-ansible/pull/883

PR #1235, opened 2016-01-19
Update cluster_hosts.yml for cloud providers

detiber:
- Add g_infra_hosts (nodes with sub-type infra)
- Add g_compute_hosts (nodes with sub-type compute)
- Reduce duplication by re-using previously defined variables

PR #1237, opened 2016-01-19
infra_node fixes

detiber:
- openshift_master role update
  - infra_nodes was previously being set to num_infra, which is an integer
    value when using the cloud providers, added a new variable osm_infra_nodes
    that is expected to be a list of hosts

- Cloud provider config playbook updates
  - override openshift_router_selector for cloud providers to avoid using the
    default of 'region=infra' when deployment_type is not 'online'
  - Set osm_infra_nodes to g_infra_host for cloud providers

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -15,6 +15,9 @@ import json
 import yaml
 from ansible.utils.unicode import to_unicode
 
+# Disabling too-many-public-methods, since filter methods are necessarily
+# public
+# pylint: disable=too-many-public-methods
 class FilterModule(object):
     ''' Custom ansible filters '''
 
@@ -48,7 +51,7 @@ class FilterModule(object):
     def oo_flatten(data):
         ''' This filter plugin will flatten a list of lists
         '''
-        if not issubclass(type(data), list):
+        if not isinstance(data, list):
             raise errors.AnsibleFilterError("|failed expects to flatten a List")
 
         return [item for sublist in data for item in sublist]
@@ -68,14 +71,14 @@ class FilterModule(object):
                 filters   = {'z': 'z'}
                 returns [1, 2, 3]
         '''
-        if not issubclass(type(data), list):
+        if not isinstance(data, list):
             raise errors.AnsibleFilterError("|failed expects to filter on a List")
 
         if not attribute:
             raise errors.AnsibleFilterError("|failed expects attribute to be set")
 
         if filters is not None:
-            if not issubclass(type(filters), dict):
+            if not isinstance(filters, dict):
                 raise errors.AnsibleFilterError("|failed expects filter to be a"
                                                 " dict")
             retval = [FilterModule.get_attr(d, attribute) for d in data if (
@@ -93,10 +96,10 @@ class FilterModule(object):
                 returns [1, 3]
         '''
 
-        if not issubclass(type(data), list):
+        if not isinstance(data, list):
             raise errors.AnsibleFilterError("|failed expects to filter on a list")
 
-        if not issubclass(type(keys), list):
+        if not isinstance(keys, list):
             raise errors.AnsibleFilterError("|failed expects first param is a list")
 
         # Gather up the values for the list of keys passed in
@@ -112,10 +115,10 @@ class FilterModule(object):
                 returns [1, 3]
         '''
 
-        if not issubclass(type(data), dict):
+        if not isinstance(data, dict):
             raise errors.AnsibleFilterError("|failed expects to filter on a dict")
 
-        if not issubclass(type(keys), list):
+        if not isinstance(keys, list):
             raise errors.AnsibleFilterError("|failed expects first param is a list")
 
         # Gather up the values for the list of keys passed in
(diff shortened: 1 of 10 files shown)


detiber:
depends on https://github.com/openshift/openshift-ansible/pull/1235

abutcher:
@detiber BYO would need to set `osm_infra_nodes` in their inventory moving forward? A while back I was thinking about what would be the easiest way to end up with infrastructure nodes:
* With a selector set (or the default) we could determine how many infra nodes were configured by checking hostvars.
* If there was an `[infra_nodes]` or `[infra]` host group, we could just label them automatically.
* Something else.

detiber:
@abutcher There should be no need to set osm_infra_nodes if setting the labels...  I'm just worried about explaining it in the docs for the Advanced Installation guide. I almost think byo would probably be better to create the osm_infra_nodes variable from querying the labels on groups.nodes instead.

Of course, we really need to have a better approach to infrastructure nodes than labels in the long run...  We should be able to add a new label "openshift_infra" and automatically set it to True for infra nodes and False for non-infra nodes. Then it would just be a matter of updating the scheduler to allow the 'default' and 'openshift-infra' projects to schedule pods there. At that point, we should probably set a projectDefaultNodeSelector in the master config so that openshift_infra=False is added to the default prject node selector for all new projects as well.

detiber:
@abutcher btw, thanks for making me take another look, I realized that I missed the fix for replicas when breaking this PR out.

abutcher:
@detiber Gotcha, I like `openshift_infra=True` seems pretty intuitive.

detiber:
@brenton ptal

brenton:
@detiber, where's the handling of `osm_infra_nodes` for byo?  You mentioned that we would query the labels.  Are we already doing that?

What does `osm` stand for?

detiber:
@brenton I'll rename osm_infra_nodes, it should rather be openshift_master_infra_nodes

I neglected to update the replicas setting to take the difference into account, but the selector is set to use the "region" currently for byo installs rather than the type=infra tag.

brenton:
Makes sense.  Is this ready or should someone from Ops review?

detiber:
@brenton I'll update this PR to address the issue.

detiber:
latest changes still need testing, but defaults to 1 replica if openshift.master.infra_nodes is not defined and renames osm_infra_nodes to openshift_infra_nodes.

detiber:
aos-ci-test

detiber:
aos-ci-test

filter_plugins/oo_filters.py
```diff
@@ -273,16 +276,52 @@ def oo_filter_list(data, filter_attr=None):
...
         # Gather up the values for the list of keys passed in
         return [x for x in data if x.has_key(filter_attr) and x[filter_attr]]
 
     @staticmethod
+    def oo_nodes_with_label(nodes, label=None, value=None):
+        ''' Filters a list of nodes by label '''
+        if not isinstance(nodes, list):
+            raise errors.AnsibleFilterError("failed expects to filter on a list")
+        if label is None:
+            return nodes
```
brenton: Why would we want to return all the nodes here?  It seems a little odd for `oo_nodes_with_label` to return any nodes that don't have a label.

filter_plugins/oo_filters.py
```diff
@@ -273,16 +276,52 @@ def oo_filter_list(data, filter_attr=None):
...
+            else:
+                return False
+
+            if isinstance(labels, basestring):
+                labels = yaml.safe_load(labels)
+            if not isinstance(labels, dict):
+                raise errors.AnsibleFilterError(
+                    "failed expected node labels to be a dict or serializable to a dict"
+                )
+            return label in labels and (value is None or labels[label] == value)
```
brenton: Can you explain the "value is None or labels[label] == value" part?

detiber: If called without a value specified, it will return all nodes that have the label.  Will add a comment to clarify.

filter_plugins/oo_filters.py
detiber: I can see it both ways. My thought was that since it's a filter, if no filter criteria are set then the full list us returned.

brenton: Would there be a reason to use the filter if no criteria was set?  It seems to me like it would indicate someone was using it incorrectly.

filter_plugins/oo_filters.py
brenton: Ahh, that makes sense.  I wasn't understanding why there would be labels without values.

filter_plugins/oo_filters.py
detiber: I was thinking of the case where a variable was used to pass into the filter, but I agree it seems like an issue around misuse.  I'll updated labels to make it a required parameter and remove the return if label is None.

detiber:
@brenton updated with better docs, if it looks good, I'll squash for merge.

brenton:
yeah, this looks good.  squash it and then we can merge it in after the CI passes again.

detiber:
aos-ci-test

detiber:
@brenton

PR #1240, opened 2016-01-19
Fix client and admin wrapper pathing for non-root user install.

abutcher:
The `oc` and `oadm` wrappers located in `/usr/local/bin` aren't in `PATH` for non-root user containerized installs. There may be a cleaner way to ensure that our path includes `/usr/local/bin` but this seems like an alright way to go about it... if we know a host is containerized we can use these paths.

https://bugzilla.redhat.com/show_bug.cgi?id=1299742

@brenton @detiber PTAL


brenton:
That makes sense to me.

detiber:
:+1:

abutcher:
aos-ci-test

PR #1251, opened 2016-01-21
Fix 'pcs status' validation filter

abutcher:
Input `data` should be tested for instance of `basestring` (`str` and `unicode`) instead of `str`.

https://bugzilla.redhat.com/show_bug.cgi?id=1298803


abutcher:
@detiber PTAL

detiber:
:+1: @brenton

PR #1266, opened 2016-01-22
Reverse 'is-active' check in restart playbook

abutcher:
Fail when master restart playbook finds no active masters rather than any failed masters. To succeed at this gate, at least one master must report `active` from `systemctl is-active atomic-openshift-master`.

I've found that `failed` is a completely valid state but we want at least one master to be `active`.

Log from a pacemaker restart: http://file.rdu.redhat.com/abutcher/ansible.2.bz1298803.txt

@brenton @detiber ptal


brenton:
LGTM

detiber:
:+1:

Issue #1268, opened 2016-01-23
openssl not installed on etcd?


goern:
openssl seems not to be installed by default on RHEL71
```
TASK: [etcd_ca | copy ] ******************************************************* 
changed: [ose3-etcd-1.test.example.com]

TASK: [etcd_ca | command openssl req -config {{ etcd_openssl_conf }} -newkey rsa:4096 -keyout {{ etcd_ca_key }} -new -out {{ etcd_ca_cert }} -x509 -extensions {{ etcd_ca_exts_self }} -batch -nodes -days {{ etcd_ca_default_days }} -subj /CN=etcd-signer@{{ ansible_date_time.epoch }}
] *** 
failed: [ose3-etcd-1.test.example.com] => {"cmd": "openssl req -config /etc/etcd/ca/openssl.cnf -newkey rsa:4096 -keyout /etc/etcd/ca/ca.key -new -out /etc/etcd/ca/ca.crt -x509 -extensions etcd_v3_ca_self -batch -nodes -days 365 -subj ''", "failed": true, "rc": 2}
msg: [Errno 2] No such file or directory

FATAL: all hosts have already failed -- aborting
```

goern:
here is the https://github.com/goern/openshift-ansible/blob/testing/Vagrantfile and https://github.com/goern/openshift-ansible/blob/testing/inventory/byo/hosts.testing I used

PR #1270, opened 2016-01-25
Ensure openssl present for etcd_ca

detiber:
Previously we assumed that openssl was present on the host machine. Now we at least test the package is present: fixes https://github.com/openshift/openshift-ansible/issues/1268


abutcher:
aos-ci-test

detiber:
@brenton

PR #1274, opened 2016-01-26
Fix checking for update package availability

rassie:
Currently, if `yum list available` returns two versions, for whatever reason, no sorting is imposed. Therefore it's possible that an upgraded package version is available but is not being detected.

This patch sorts the version number list so that most recent version is always picked first.


detiber:
ok to test

detiber:
aos-ci-test

abutcher:
aos-ci-test

detiber:
@brenton

PR #1275, opened 2016-01-26
Expose data_dir

spinolacastro:
Expose data_dir so we can take advantage of ephemeral disks.


detiber:
ok to test

detiber:
aos-ci-test

detiber:
:+1: @brenton

PR #563, opened 2015-09-04
Avoid a recursive loop

sdodson:
Fixes an issue introduced in #558 

```
TASK: [openshift_facts | Verify Ansible version is greater than 1.8.0 and not 1.9.0] *** 
fatal: [ose3-master.example.com] => Failed to template {{ openshift_version | default('') }}: Failed to template {{ openshift_version | default('') }}: Failed to template {{ openshift_version | default('') }}: Failed to template {{ openshift_version | default('') }}: Fail
ed to template {{ openshift_version | default('') }}: Failed to template {{ openshift_version | default('') }}: Failed to template {{ openshift_version | default('') }}: Failed to template {{ openshift_version | default('') }}: Failed to template {{ openshift_version | de
fault('') }}: Failed to template {{ openshift_version | default('') }}: Failed to template {{ openshift_version | default('') }}:
...
recursive loop detected in template string: {{ openshift_version | default('') }}

FATAL: all hosts have already failed -- aborting
```

PR #1276, opened 2016-01-26
Fix pkg_version 

spinolacastro:
Ansible install the latest version despite the value we set at openshift_pkg_version. This patch intend to fix it.

### DIFF
--- a/roles/openshift_common/vars/main.yml
+++ b/roles/openshift_common/vars/main.yml
@@ -5,3 +5,4 @@
 # chains with the public zone (or the zone associated with the correct
 # interfaces)
 os_firewall_use_firewalld: False
+openshift_version: "{{ openshift_pkg_version | default('') }}"


abutcher:
This makes sense to me. We might as well just use `openshift_pkg_version` in the tasks but I'm not partial either way.

@detiber ptal

sdodson:
@spinolacastro :-) thanks for fixing this, just ran into this myself and confirmed that this fixes it.

@abutcher I agree it'd be great to get out of translating host variables into local variables in order to set defaults but I wasn't sure how to do that. When I originally put this together I ran into a recursive loop, is there a better way to fix this? https://github.com/openshift/openshift-ansible/pull/563/files

abutcher:
@sdodson Ah, I see. I think we would just have to remove `openshift_version` from all of the vars files and just rely on it directly from hostvars with a `| default('')` every where we used it.

abutcher:
I think we should keep it the way it is with `openshift_pkg_version` and merge this since it matches the pattern.

sdodson:
LGTM, @brenton ready to merge

abutcher:
:+1:

detiber:
:+1:

PR #1280, opened 2016-01-26
replace yum with dnf

sjenning:
The yum module is not installed following the directions for installation on Fedora (23) and causes the byo playbook to fail.

### DIFF
--- a/roles/openshift_cli/tasks/main.yml
+++ b/roles/openshift_cli/tasks/main.yml
@@ -6,7 +6,7 @@
       cli_image: "{{ osm_image | default(None) }}"
       
 - name: Install clients
-  yum: pkg={{ openshift.common.service_type }}-clients state=installed
+  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-clients state=present"
   when: not openshift.common.is_containerized | bool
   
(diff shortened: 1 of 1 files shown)


roles/openshift_cli/tasks/main.yml
```diff
@@ -6,7 +6,7 @@
       cli_image: "{{ osm_image | default(None) }}"
       
 - name: Install clients
-  yum: pkg={{ openshift.common.service_type }}-clients state=installed
+  dnf: name={{ openshift.common.service_type }}-clients state=installed
```
abutcher: Thanks @sjenning, if we change this to an `ansible_pkg_mgr` action it should work for either.

```
- name: Install clients
  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-clients state=present"
```

sjenning:
@abutcher thanks! updated the PR.

abutcher:
aos-ci-test

abutcher:
ok to test

abutcher:
@brenton LGTM - replaces a yum task with an `ansible_pkg_mgr` action so that it works for Fedora/RHEL/Centos.

PR #1288, opened 2016-01-27
Configure logrotate on atomic.

abutcher:
@sdodson PTAL - logrotate comes standard with atomic so we can just skip the package install afaik

### DIFF
--- a/inventory/byo/hosts.aep.example
+++ b/inventory/byo/hosts.aep.example
@@ -206,6 +206,10 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # Configure node kubelet arguments
 #openshift_node_kubelet_args={'max-pods': ['40'], 'image-gc-high-threshold': ['90'], 'image-gc-low-threshold': ['80']}
 
+# Configure logrotate scripts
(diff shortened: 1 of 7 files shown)


detiber:
lgtm, assuming it works on an atomic host :smile:

abutcher:
@detiber We only set `logrotate_scripts` on ec2, yeah?

abutcher:
^ meaning this would basically be a noop on atomic but the configuration would occur if we had any

detiber:
@abutcher nope, the common playbooks are used by the byo playbook as well

sdodson:
LGTM, as far as I can tell this will only enable logrotate which seems fine. The aws playbook is the only playbook that actually defines some logrotate scripts, we may want to consider moving that definition into the common playbooks?

abutcher:
aos-ci-test

detiber:
@sdodson I'm not sure we want to manage a user's logrotate config (unless they want us to at least).

Maybe we can offer some examples in the docs and let them "opt in"

abutcher:
@sdodson @detiber Added an example to byo inventories

detiber:
:+1:

abutcher:
@brenton

PR #1291, opened 2016-01-28
Check that openshift_hostname resolves to an ip on our host

sdodson:
Check to see if openshift_hostname resolves to an ip on our host in order to avoid problems in public/private IP environments where liveness and readiness probes are assigned to pods utilzing host networking.

See https://bugzilla.redhat.com/show_bug.cgi?id=1293578 for more details

### DIFF
--- a/playbooks/common/openshift-cluster/config.yml
+++ b/playbooks/common/openshift-cluster/config.yml
@@ -1,6 +1,8 @@
 ---
 - include: evaluate_groups.yml
 
+- include: validate_hostnames.yml
+
 - include: ../openshift-docker/config.yml
 
 - include: ../openshift-etcd/config.yml
--- a/playbooks/common/openshift-cluster/validate_hostnames.yml
+++ b/playbooks/common/openshift-cluster/validate_hostnames.yml
@@ -1,0 +1,26 @@
+---
+- include: evaluate_groups.yml
+
+- name: Gather and set facts for node hosts
+  hosts: oo_nodes_to_config
+  roles:
+  - openshift_facts
+  tasks:
+  - openshift_facts:
+      role: "{{ item.role }}"
+      local_facts: "{{ item.local_facts }}"
+    with_items:
+      - role: common
+        local_facts:
+          hostname: "{{ openshift_hostname | default(None) }}"
+          public_hostname: "{{ openshift_public_hostname | default(None) }}"
+  - shell:
+      getent ahostsv4 {{ openshift.common.hostname }} | head -n 1 | awk '{ print $1 }'
+    register: lookupip
+    changed_when: false
+    failed_when: false
+  - name: Warn user about bad openshift_hostname values
+    pause:
+       prompt: "The hostname \"{{ openshift.common.hostname }}\" for \"{{ ansible_nodename }}\" doesn't resolve to an ip address owned by this host. Please set openshift_hostname variable to a hostname that when resolved on the host in question resolves to an IP address matching an interface on this host. This host will fail liveness checks for pods utilizing hostPorts, press CTRL-C to continue."
+       seconds: "{{ 10 if openshift_override_hostname_check | default(false) | bool else omit }}"
+    when: lookupip.stdout not in ansible_all_ipv4_addresses
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ gce.ini
 multi_ec2.yaml
 multi_inventory.yaml
 .vagrant
+.tags*
--- a/inventory/byo/hosts.aep.example
+++ b/inventory/byo/hosts.aep.example
@@ -210,6 +210,13 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # See: https://github.com/nickhammond/ansible-logrotate
 #logrotate_scripts=[{"name": "syslog", "path": "/var/log/cron\n/var/log/maillog\n/var/log/messages\n/var/log/secure\n/var/log/spooler\n", "options": ["daily", "rotate 7", "compress", "sharedscripts", "missingok"], "scripts": {"postrotate": "/bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true"}}]
 
+# openshift-ansible will wait indefinitely for your input when it detects that the
+# value of openshift_hostname resolves to an IP address not bound to any local
+# interfaces. This mis-configuration is problematic for any pod leveraging host
+# networking and liveness or readiness probes. 
+# Setting this variable to true will override that check.
+#openshift_override_hostname_check=true
+
 # host group for masters
 [masters]
 aep3-master[1:3]-ansible.test.example.com
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -220,6 +220,13 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # See: https://github.com/nickhammond/ansible-logrotate
 #logrotate_scripts=[{"name": "syslog", "path": "/var/log/cron\n/var/log/maillog\n/var/log/messages\n/var/log/secure\n/var/log/spooler\n", "options": ["daily", "rotate 7", "compress", "sharedscripts", "missingok"], "scripts": {"postrotate": "/bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true"}}]
 
+# openshift-ansible will wait indefinitely for your input when it detects that the
+# value of openshift_hostname resolves to an IP address not bound to any local
+# interfaces. This mis-configuration is problematic for any pod leveraging host
+# networking and liveness or readiness probes. 
+# Setting this variable to true will override that check.
+#openshift_override_hostname_check=true
+
 # host group for masters
 [masters]
 ose3-master[1:3]-ansible.test.example.com
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -210,6 +210,13 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # See: https://github.com/nickhammond/ansible-logrotate
 #logrotate_scripts=[{"name": "syslog", "path": "/var/log/cron\n/var/log/maillog\n/var/log/messages\n/var/log/secure\n/var/log/spooler\n", "options": ["daily", "rotate 7", "compress", "sharedscripts", "missingok"], "scripts": {"postrotate": "/bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true"}}]
 
+# openshift-ansible will wait indefinitely for your input when it detects that the
+# value of openshift_hostname resolves to an IP address not bound to any local
+# interfaces. This mis-configuration is problematic for any pod leveraging host
+# networking and liveness or readiness probes. 
+# Setting this variable to true will override that check.
+#openshift_override_hostname_check=true
+
 # host group for masters
 [masters]
 ose3-master[1:3]-ansible.test.example.com


detiber:
:+1:

sdodson:
I'm actually finding that default firewall rules seem to be fine. Am I missing something?

```
[root@ose3-master ~]# iptables -L                                                                            
Chain INPUT (policy ACCEPT)                                                                                  
target     prot opt source               destination                                                         
ACCEPT     udp  --  anywhere             anywhere             multiport dports 4789 /* 001 vxlan incoming */ 
ACCEPT     all  --  anywhere             anywhere             /* traffic from docker for internet */         
ACCEPT     all  --  anywhere             anywhere             state RELATED,ESTABLISHED                      
ACCEPT     icmp --  anywhere             anywhere                                                            
ACCEPT     all  --  anywhere             anywhere                                                            
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:ssh                          
OS_FIREWALL_ALLOW  all  --  anywhere             anywhere                                                    
REJECT     all  --  anywhere             anywhere             reject-with icmp-host-prohibited               
                                                                                                             
Chain FORWARD (policy ACCEPT)                                                                                
target     prot opt source               destination                                                         
DOCKER     all  --  anywhere             anywhere                                                            
ACCEPT     all  --  anywhere             anywhere             ctstate RELATED,ESTABLISHED                    
ACCEPT     all  --  anywhere             anywhere                                                            
ACCEPT     all  --  anywhere             anywhere                                                            
DOCKER     all  --  anywhere             anywhere                                                            
ACCEPT     all  --  anywhere             anywhere             ctstate RELATED,ESTABLISHED                    
ACCEPT     all  --  anywhere             anywhere                                                            
ACCEPT     all  --  anywhere             anywhere                                                            
ACCEPT     all  --  10.1.0.0/16          anywhere                                                            
ACCEPT     all  --  anywhere             10.1.0.0/16                                                         
ACCEPT     all  --  10.1.0.0/16          anywhere                                                            
ACCEPT     all  --  anywhere             10.1.0.0/16                                                         
ACCEPT     all  --  10.1.0.0/16          anywhere                                                            
ACCEPT     all  --  anywhere             10.1.0.0/16                                                         
ACCEPT     all  --  10.1.0.0/16          anywhere                                                            
ACCEPT     all  --  anywhere             10.1.0.0/16                                                         
REJECT     all  --  anywhere             anywhere             reject-with icmp-host-prohibited      
                                                                                                    
Chain OUTPUT (policy ACCEPT)                                                                        
target     prot opt source               destination                                                
                                                                                                    
Chain DOCKER (2 references)                                                                         
target     prot opt source               destination                                                
                                                                                                    
Chain OS_FIREWALL_ALLOW (1 references)                                                              
target     prot opt source               destination                                                
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:newoak              
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:2379                
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:2380                
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:pcsync-https        
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:10250               
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:domain              
ACCEPT     udp  --  anywhere             anywhere             state NEW udp dpt:domain              
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:http                
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:https               
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:24224               
ACCEPT     udp  --  anywhere             anywhere             state NEW udp dpt:24224               
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:pcsync-http         
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:efi-mg              
ACCEPT     udp  --  anywhere             anywhere             state NEW udp dpt:hpoms-dps-lstn      
ACCEPT     udp  --  anywhere             anywhere             state NEW udp dpt:netsupport          
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:websm               
ACCEPT     tcp  --  anywhere             anywhere             state NEW tcp dpt:10255               
ACCEPT     udp  --  anywhere             anywhere             state NEW udp dpt:10255               
ACCEPT     udp  --  anywhere             anywhere             state NEW udp dpt:4789                
[root@ose3-master ~]#

[root@ose3-master ~]# curl 192.168.122.134:1936/healthz
<html><body><h1>200 OK</h1>
Service ready.
</body></html>
```

detiber:
@sdodson you are curling from the same host, so its treated as localhost

playbooks/common/openshift-cluster/config.yml
```diff
@@ -1,6 +1,8 @@
 ---
 - include: evaluate_groups.yml
 
+- include: validate_hostnames.yml
```
sdodson: Should we leave this here or only in byo playbooks?

playbooks/common/openshift-cluster/validate_hostnames.yml
```diff
@@ -0,0 +1,28 @@
...
+  - openshift_facts:
+      role: "{{ item.role }}"
+      local_facts: "{{ item.local_facts }}"
+    with_items:
+      - role: common
+        local_facts:
+          hostname: "{{ openshift_hostname | default(None) }}"
+          public_hostname: "{{ openshift_public_hostname | default(None) }}"
+  - shell:
+      getent ahostsv4 {{ openshift.common.hostname }} | head -n 1 | awk '{ print $1 }'
```
sdodson: I couldn't come up with a better solution to cover cases where they had /etc/hosts entries, though now that I think of it if they do have /etc/hosts entries those wouldn't be reflected in containerized installs.

sdodson:
@detiber mind taking another look, this adds the dns resolution to host interfaces checks as well as attempting to connect to port 1936. I think the first is probably sufficient in most cases but the second would find places where IaaS rules deny access if they continue.

playbooks/common/openshift-cluster/validate_hostnames.yml
detiber: If they are using /etc/hosts entries, they are going to have bigger issues.

playbooks/common/openshift-cluster/config.yml
detiber: I think the common playbook is good, we'll want to warn for any usage.

roles/openshift_node/tasks/main.yml
```diff
@@ -124,3 +124,15 @@
...
+  script: check-1936-access.sh {{ openshift.common.hostname }}
+  register: check1936
+  failed_when: false
+  changed_when: false
+
+- name: Warn about restricted access to port 1936
+  pause:
+    prompt: "Your node {{ ansible_host }} cannot access {{ openshift.common.hostname }} on port 1936, this host will fail default router liveness checks. Press CTRL-C to abort or continue"
+    seconds: 30
+  when: check1936.rc != 0
```
detiber: We'll probably want to collect this response and display it back as a localhost task, otherwise the pause task could get lost in the scrollback buffer

roles/openshift_node/files/check-1936-access.sh
```diff
@@ -0,0 +1,30 @@
...
+}
+
+# Check to see if we already have a router and stats port ready to go
+if check-1936; then 
+    echo "OK"
+    exit 0
+fi
+
+# No router running currently, setup netcat to test
+echo "200 OK" | nc -l 1936 -w 5 &
```
detiber: I don't think nc is installed by default on a minimal install.

playbooks/common/openshift-cluster/validate_hostnames.yml
```diff
@@ -0,0 +1,28 @@
...
+    register: lookupip
+    changed_when: false
+    failed_when: false
+  - pause:
+      prompt: "The hostname {{ openshift.common.hostname }} for {{ ansible_nodename }} doesn't resolve to an ip address owned by this host. Please set openshift_hostname variable to a hostname that when resolved on the host in question resolves to an IP address matching an interface on this host. This host will fail liveness checks for pods utilizing hostPorts, (a)bort or (c)ontinue"
+    when: lookupip.stdout not in ansible_all_ipv4_addresses
```
detiber: I'd say we should leave it at this check...  otherwise we are adding an otherwise unnecessary dependency, or creating a tool that will attempt to bind to port 1936 to verify connectivity.

I think this will catch the majority of uses, and then the other cases can be handled based on routers that are failing health checks.

sdodson: Ok, I'll separate out that work and put it in a WIP PR encase anyone wants to pick it up where I've left it.

sdodson:
@detiber thanks for review, I think it's good to go now

detiber:
:+1: @brenton

brenton:
@sdodson, is this the one you said you were going to split in to two PRs?

sdodson:
@brenton Yeah, I split the check for connectivity to `openshift_hostname` on port 1936 off into #1315

brenton:
@sdodson, Unless we can't get https://github.com/openshift/origin/pull/6939 merged I'm thinking that we probably don't want to open the port.  If we opened the port we'd then have to deal with closing it once a real fix was put in.

detiber:
@brenton it's relatively easy to close the port with os_firewall

sdodson:
@brenton I've dropped opening the port from this PR given we know we'll want this regardless of the outcome of the origin PR and discussions happening there. I'll open up a separate PR to open port 1936 based on the outcome of the discussions in origin.

brenton:
@sdodson, cool.  This looks good to me then.   Should we merge it after running the tests?

sdodson:
aos-ci-test

sdodson:
:-) didn't account for when `openshift_hostname` is set to an IP address, i'll fix that.

sdodson:
@arilivigni Can you confirm that the CI jobs are setting openshift_hostname to the public IPs of the openstack instances?

@brenton I believe the CI test is a true positive error.  I can provide a timeout meaning that it will pause for $x seconds and then move on if there's no response but I'm not sure that's the right thing to do.

brenton:
@sdodson, the other way to to provide a setting to ignore warnings.  We could set that in CI.

arilivigni:
@sdodson it is listed in the log we use the public IPs in the inventory:

https://aos-ci.s3.amazonaws.com/openshift-ansible/jenkins-openshift-ansible-run-openshift-ansible-92/ed5ba80f4126241a316ba2635adcbd52e3562657.txt

detiber:
@sdodson is it the lack of timeout that is the issue or that it is attempting to prompt for user input?

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
@detiber adding a seconds value to the pause makes the tests pass but I think we shouldn't include that in the PR as we've added an override variable.

sdodson:
aos-ci-test

sdodson:
@detiber @brenton I've made it so that there's an override variable `openshift_override_hostname_check`, when set this will set the pause timeout to 10 seconds, unfortunately this swallows the prompt, I'm not sure why. I think that's OK because they've deliberately overridden the check so they should know what they're doing. I left the pause in so CI tests would show signs of positive hits on the check in the logs but.

When `openshift_override_hostname_check` is not set the ansible playbook will pause indefinitely waiting for the user to abort or continue.

brenton:
@sdodson, That sounds good to me.  Do you know how to update the inventory for the ci run or do we need ari's help?

brenton:
Actually, looking at the gerrit email I just received it looks like you already updated the ci job. :) @detiber, if this looks good to you I'll merge it in.

sdodson:
@brenton already done, latest ci job looks to have performed as expected

https://aos-ci.s3.amazonaws.com/openshift-ansible/jenkins-openshift-ansible-run-openshift-ansible-98/cf9403fa748e54d35afe7e8416aa736d6ea375d2.txt

detiber:
@brenton :shipit:

PR #1293, opened 2016-01-28
aoi: Ask for osm_default_subdomain in interactive mode

smunilla:
Add a question to ask the user for a default subdomain.


smunilla:
aos-ci-test

PR #1305, opened 2016-01-29
Mount /etc/openshift when starting openshift-* containers

jprovaznik:
Otherwise openshift-* services fail to start because of missing
/etc/openshift/password file.


detiber:
@jprovaznik what if we updated the example inventory files to put the htpasswd file in /etc/origin/master instead?

jprovaznik:
@detiber that sounds like a better idea. I'm closing this.

Issue #1170, opened 2016-01-12
Set nodeName to internal DNS by default for ec2


dak1n1:
The openshift-ansible installer currently creates node-config.yaml using the internal IP as the nodeName. This causes the following error when a node is [configured for AWS.](https://docs.openshift.com/enterprise/3.1/install_config/configuring_aws.html)

Jan 12 12:39:18 ip-172-20-1-226.ec2.internal atomic-openshift-node[32871]: E0112 12:39:18.979571   32871 kubelet.go:925] Unable to construct api.Node object for kubelet: failed to get external ID from cloud provider: instance not found

Instead, the nodeName should be set to the internal DNS name of the host. For example:

nodeName: ip-172-20-1-226.ec2.internal

PR #1310, opened 2016-01-29
Fix hostname for aws cloud provider

detiber:
- No longer set openshift_hostname to the private ip for the instance

resolves https://github.com/openshift/openshift-ansible/issues/1170


detiber:
@dak1n1

dak1n1:
:+1:

PR #1315, opened 2016-02-01
[WIP] Check nodes for access to port 1936

sdodson:
Proof of concept checking a given node for access to port 1936 via it's openshift_hostname
This is an extension of #1291

### DIFF
--- a/roles/openshift_node/tasks/main.yml
+++ b/roles/openshift_node/tasks/main.yml
@@ -124,3 +124,15 @@
 
 - set_fact:
     node_service_status_changed: "{{ start_result | changed }}"
+    
+- name: Check access to port 1936
+  script: check-1936-access.sh {{ openshift.common.hostname }}
+  register: check1936
+  failed_when: false
+  changed_when: false
+
+- name: Warn about restricted access to port 1936
+  pause:
+    prompt: "Your node {{ ansible_host }} cannot access {{ openshift.common.hostname }} on port 1936, this host will fail default router liveness checks. Press CTRL-C to abort or continue"
+    seconds: 30
+  when: check1936.rc != 0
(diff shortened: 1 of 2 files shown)


roles/openshift_node/tasks/main.yml
```diff
@@ -124,3 +124,15 @@
...
+  script: check-1936-access.sh {{ openshift.common.hostname }}
+  register: check1936
+  failed_when: false
+  changed_when: false
+
+- name: Warn about restricted access to port 1936
+  pause:
+    prompt: "Your node {{ ansible_host }} cannot access {{ openshift.common.hostname }} on port 1936, this host will fail default router liveness checks. Press CTRL-C to abort or continue"
+    seconds: 30
+  when: check1936.rc != 0
```
xlr-8: Although you check for an URL in the script, is it really necessary? Wouldn't it have been easier to use the wait_for module of ansible http://docs.ansible.com/ansible/wait_for_module.html for that task?

sdodson: I did this because if a router wasn't deployed on the host the check would fail even if the networking rules were fine because nothing was listening so we need to use `nc` to stub out a listener to verify things. Does that make sense?

xlr-8: Yep, thanks for the clarification!

sdodson:
@xlr-8 BTW, the way we went to address this was to check that the node's hostname resolves to a local ip address then warn if it doesn't. We've also modified the liveness probe created for routers such that it works against localhost rather than against the node's IP address. So I think I'm going to close this PR unless you can think of a really good reason why we still need this?

xlr-8:
Nope, fine with me

PR #1324, opened 2016-02-02
Remove fluentd_master and fluentd_node roles

abutcher:
Remove `fluentd_master` and `fluentd_node` roles as they are not being used and will be replaced with aggregated logging configuration.


roles/fluentd_master/vars/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+embedded_etcd: "{{ openshift.master.embedded_etcd }}"
```
detiber: I'm not sure the related checks are even needed anymore with the test in the openshift_master role to verify that the api server is up before continuing.

As an aside, we should update for the proper package if it is available in the shipped repos as opposed to the public repo that it is currently being pulled from.

abutcher: Kk, I'll take the checks out.

abutcher:
@twiest @wshearn is ops using the `fluentd_master`/`fluentd_node` roles via `openshift_use_fluentd=true`?

twiest:
@abutcher I'll let @wshearn answer that.

wshearn:
We are not using fluentd_{master,node}

abutcher:
@detiber How do you feel about just making this work for now and dropping the roles when we add log aggregation bits

detiber:
@abutcher truthfully, I'd rather yank this out if they aren't using it...  unless there is a valid case to do otherwise?

abutcher:
@detiber Nah, I agree.

abutcher:
aos-ci-test

abutcher:
@detiber PTAL

detiber:
@abutcher should probably kill the use_fluentd references in playbooks/{aws,gce,libvirt,openstack} as well, otherwise, :+1: from me

abutcher:
@detiber Should be clear

```
<abutcher>@(strahd)[~/rhat/openshift-ansible] 16:10:26  (fluentd-fixes) 
$ grep -ir use_fluentd | wc -l
0
```

PR #1326, opened 2016-02-02
Fix infra_node deployment

detiber:
- Do not deploy the router/registry when the infra_nodes variable is present
  but does not contain a list of infra nodes.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1303939

- Reorganize the master playbook a bit
  - Move service account creation near the router/registry roles where they are consumed.
  - Consolidate much of the oo_first_master post-installation config instead of having many separate plays

- Fix missing dependencies on openshift_facts for some roles
- cleanup service account creation to use oadm policy add-scc-to-user instead of dumping, editing, and re-loading the yaml.

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -397,7 +397,7 @@
   hosts: oo_first_master
   vars:
     attach_registry_volume: "{{ openshift.hosted.registry.storage.kind != None }}"
-    deploy_infra: "{{ openshift.master.infra_nodes | default(0) | length > 0 }}"
+    deploy_infra: "{{ openshift.master.infra_nodes | default([]) | length > 0 }}"
     persistent_volumes: "{{ hostvars[groups.oo_first_master.0] | oo_persistent_volumes(groups) }}"
     persistent_volume_claims: "{{ hostvars[groups.oo_first_master.0] | oo_persistent_volume_claims }}"
   roles:
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -360,6 +360,8 @@
 - name: Additional master configuration
   hosts: oo_first_master
   vars:
+    cockpit_plugins: "{{ osm_cockpit_plugins | default(['cockpit-kubernetes']) }}"
+    etcd_urls: "{{ openshift.master.etcd_urls }}"
     openshift_master_ha: "{{ groups.oo_masters_to_config | length > 1 }}"
     omc_cluster_hosts: "{{ groups.oo_masters_to_config | join(' ')}}"
   roles:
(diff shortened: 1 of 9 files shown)


detiber:
aos-ci-test

abutcher:
aos-ci-test

detiber:
I'm expecting this test to fail, found some issues with the role reordering I did

detiber:
aos-ci-test

playbooks/common/openshift-master/config.yml
```diff
@@ -405,22 +393,22 @@
...
     accounts: ["router", "registry"]
-  roles:
-  - openshift_serviceaccounts
-
-- name: Create persistent volumes and services
-  hosts: oo_first_master
-  vars:
+    attach_registry_volume: "{{ openshift.hosted.registry.storage.kind != None }}"
+    deploy_infra: "{{ openshift.master.infra_nodes | default(0) | length > 0 }}"
+    nfs_host: "{{ groups.oo_nfs_to_config.0 }}"
```
abutcher: This var shouldn't be used anymore.

playbooks/common/openshift-master/config.yml
```diff
@@ -405,22 +393,22 @@
...
     persistent_volumes: "{{ hostvars[groups.oo_first_master.0] | oo_persistent_volumes(groups) }}"
     persistent_volume_claims: "{{ hostvars[groups.oo_first_master.0] | oo_persistent_volume_claims }}"
+    registry_volume_path: "{{ hostvars[groups.oo_nfs_to_config.0].openshift.nfs.exports_dir + '/' + hostvars[groups.oo_nfs_to_config.0].openshift.nfs.registry_volume }}"
```
abutcher: Same with this one.

detiber:
aos-ci-test

abutcher:
@brenton :+1:

roles/openshift_serviceaccounts/tasks/main.yml
```diff
@@ -1,36 +1,21 @@
...
+  when: item.rc != 0
+  with_items: account_test.results
+  register: create_accounts
 
-- name: Apply new scc rules for service accounts
-  command: "{{ openshift.common.client_binary }} update -f /tmp/openshift/scc.yaml --api-version=v1"
+- name: Grant the user access to the privileged scc
+  shell: >
+      {{ openshift.common.admin_binary }} policy add-scc-to-user
+      privileged system:serviceaccount:default:{{ item.item.item }}
```
brenton: wow, the with_items nesting is a little hard to read.  Does ansible have any other `with_` looping mechanisms that could make this more obvious?  The `item.item.item` part just looks really odd.

detiber: heh, basically it's because I'm chaining the output to be dependent on the output of the previous command.

It would be nice if you could do alias the variables created when using with_items.

The alternative is to just loop on the first output, which would still work, but if someone later tried to refactor and switched the ordering of the tasks it would break.

brenton: I'm not totally following.  Wouldn't things still break today if the order of the tasks were switched?  Or are you saying that if they switched the ordering now they would get an error about the registered variable not being declared so that they would get a playbook compilation error instead of a runtime error with oadm?

detiber: Thinking about this a bit more...  Since it will still error out, I don't see too much harm in chaining both off of the user check..

I'll be happy when we can make the jump to Ansible 2.0 and can put this type of dependency chain in a block so that we can more clearly express that they are indeed a chain of events that need to be done sequentially.

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
aos-ci-test

retrying due to node registration flake

detiber:
aos-ci-test

detiber:
@abutcher ptal

playbooks/common/openshift-master/config.yml
```diff
@@ -405,22 +393,25 @@
...
-    accounts: ["router", "registry"]
-  roles:
-  - openshift_serviceaccounts
-
-- name: Create persistent volumes and services
+- name: Create persistent volumes and create hosted services
   hosts: oo_first_master
   vars:
+    attach_registry_volume: "{{ openshift.hosted.registry.storage.kind != None }}"
+    deploy_infra: "{{ openshift.master.infra_nodes | default(0) | length > 0 }}"
```
abutcher: Should be `default([])`?

detiber: indeed.

detiber:
aos-ci-test

detiber:
@brenton

brenton:
This breaks 3.0 installs because `oadm policy add-scc-to-user` didn't exist back then.

PR #1327, opened 2016-02-03
Make pod_eviction_timeout configurable from cli

jkhelil:
We need to make it possible to configure openshift_master_pod_eviction_timeout parameter from cli so we can override the default value from bin/cluster

### DIFF
--- a/playbooks/byo/openshift-cluster/cluster_hosts.yml
+++ b/playbooks/byo/openshift-cluster/cluster_hosts.yml
@@ -7,6 +7,8 @@ g_master_hosts: "{{ groups.masters | default([]) }}"
 
 g_node_hosts:   "{{ groups.nodes | default([]) }}"
 
+g_new_node_hosts: "{{ groups.new_nodes | default([]) }}"
+
 g_nfs_hosts:   "{{ groups.nfs | default([]) }}"
 
 g_all_hosts:    "{{ g_master_hosts | union(g_node_hosts) | union(g_etcd_hosts)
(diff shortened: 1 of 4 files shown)


jkhelil:
@lhuard1A Please review

roles/openshift_master/tasks/main.yml
```diff
@@ -56,7 +56,7 @@
       embedded_dns: "{{ openshift_master_embedded_dns | default(None) }}"
       dns_port: "{{ openshift_master_dns_port | default(None) }}"
       bind_addr: "{{ openshift_master_bind_addr | default(None) }}"
-      pod_eviction_timeout: "{{ openshift_master_pod_eviction_timeout | default(None) }}"
+      pod_eviction_timeout: "{{ lookup('oo_option', 'openshift_master_pod_eviction_timeout') | default(openshift_master_pod_eviction_timeout, true) | default(None) }}"
```
detiber: This should probably be:
```
pod_eviction_timeout: "{{ lookup('oo_option', 'openshift_master_pod_eviction_timeout') | default(openshift_master_pod_eviction_timeout | default(None), true) }}"
```

One thing I think I would like to see though, is not pushing the oo_option filter down into the roles. 

Instead, you should be able to add the following to https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-master/config.yml#L19 for the same effect:
```
- set_fact:
    openshift_master_pod_eviction_timeout: {{ lookup('oo_option', 'openshift_master_pod_eviction_timeout') | default(None)
    when: openshift_master_pod_eviction_timeout is not defined
```

jkhelil:
@detiber  I will implement like you suggest, test and submit new pull request

PR #1333, opened 2016-02-04
renamed /etc/openshift to /etc/origin

stenwt:


sdodson:
LGTM, ops only playbooks

mwoodson:
@sdodson @detiber can you look at this one?

The bottom file (oso-rhel7-host-monitoring.service.j2) and this is a good change. 


What about openshift_cluster_metrics/tasks/main.yml?

detiber:
@mwoodson @stenwt lgtm

Issue #1138, opened 2016-01-08
ansible-1.9.4 and yum-3.4.3-125 and yum-utils-1.1.31-29 may skip installing packages


sdodson:
@wshearn and I were looking into an issue where the installer was silently failing to install packages. This problem will manifest itself as an access denied error when ansible attempts to start a service that's provided by an RPM that was being silently skipped. The verbose (-vvv) output will show something similar to this 

```
ok: [ose-node1.example.com] => {"changed": false, "msg": "", "rc": 0, "results": ["Plugin \"search-disabled-repos\" requires API 2.7. Supported API is 2.6. providing atomic-openshift-node is already installed", "Plugin \"search-disabled-repos\" requires API 2.7. Supported API is 2.6. providing tuned-profiles-atomic-openshift-node is already installed"]}
```

This seems to be https://github.com/ansible/ansible-modules-core/issues/2559

After trying a few things we were able to figure out that updating both yum-3.4.3-132.el7 and yum-utils-1.1.31-34.el7 versions which shipped with RHEL 7.2 seems to have resolved the issue. I'm surprised we haven't seen more reports of this. Maybe this only happens when subscription-manager is the latest but yum isn't?

Mostly putting this in encase anyone else runs into it, but if someone else has time to look into this more we should consider enforcing a minimum version of yum as much as I'd rather not do so.

detiber:
@sdodson, I don't think that we can enforce a minimum version of yum and/or subscription-manager, since we are supporting RHEL 7.1 for deployments/upgrades still.

That said, I suspect that if we can track down a way to identify when we have a bad combination of packages that exhibit the issue, then we can fail only when that condition is met.

tbielawa:
According to this comment, the issue should be fixed in a newer ansible version

https://github.com/ansible/ansible-modules-core/issues/3066#issuecomment-219048424

This issue has been inactive for quite some time. Please update and reopen this issue if this is still a priority you would like to see action on.

sbadakhc:
Looks like this is back? https://github.com/openshift/openshift-ansible/issues/7490

PR #1334, opened 2016-02-04
Bug 1304150 - Can't upgrade atomic-openshift to specified version

brenton:


brenton:
Thinking more about this, I think I'm going to see if I can limit this change to the upgrade playbook.

roles/openshift_common/tasks/main.yml
```diff
@@ -36,7 +36,7 @@
       data_dir: "{{ openshift_data_dir | default(None) }}"
 
 - name: Install the base package for versioning
-  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}{{ openshift_version | default('') }} state=present"
+  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}{{ openshift_version | default('') }},{{ openshift.common.service_type }}-clients{{ openshift_version | default('') }},{{ openshift.common.service_type }}-master{{ openshift_version | default('') }},{{ openshift.common.service_type }}-node{{ openshift_version | default('') }},{{ openshift.common.service_type }}-sdn-ovs{{ openshift_version | default('') }},tuned-profiles-{{ openshift.common.service_type }}-node{{ openshift_version | default('') }} state=present"
```
detiber: This role is applied to all masters and nodes, since the issue is with upgrades, should we have a entry in openshift_facts that returns the currently installed packages (possibly minus version number) and use that as a base for the packages to upgrade?

brenton:
This is a bit of a mess.  In the upgrade playbook we call the common role very early to set a bunch of variables.  I suspect that's how `openshift.common.is_containerized` is set.  Without that set it's not really easy to clean this up in the upgrade.  Technically, it would be safe to assume we didn't have any containerized installs prior to 3.1 so I may go that route.

roles/openshift_common/tasks/main.yml
brenton: I'll check that out.

detiber:
@brenton I don't see anywhere in the upgrade playbooks (at least for 3.0 -> 3.1 and 3.1 minor upgrade playbooks that call the common role. They call the playbooks/common/openshift-cluster/evaluage_groups.yml playbook which sets the host groups to use.

The openshift.common.is_containerized flag is coming from the openshift_facts role.

brenton:
@detiber, it's really complicated.  What's happening is if openshift_pkg_version is set this line fails silently https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml#L236.

Then later when we call the `openshift_master_certificates` it loads the master role which loads common which hits the error QE found.

playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
```diff
@@ -235,6 +238,11 @@
 
   - name: Upgrade master packages
     command: "{{ ansible_pkg_mgr}} update -y {{ openshift.common.service_type }}-master{{ openshift_version }}"
+    when: openshift_pkg_version is not defined
+
+  - name: Upgrade packages
+    command: "{{ ansible_pkg_mgr}} install -y {{ openshift.common.installed_product_rpms | oo_31_rpm_rename_conversion(openshift_version) | join (' ')}}"
```
brenton: @detiber, For some reason installed_product_rpms isn't set here.  If I put a debug line at near the top of this playbook it's set.  Any idea what's going on?

detiber: @brenton it's because it's a different play...  openshift_facts is called for the play that starts at 221 and ends at 229, 231 starts a separate play, but it doesn't look like it has to.

detiber:
@brenton is it failing silently or is it the bug with certain versions of yum/ansible 1.9 where the yum module returns success even if the package isn't installed?

sdodson:
> @brenton is it failing silently or is it the bug with certain versions of yum/ansible 1.9 where the yum module returns success even if the package isn't installed?

details of that archived here https://github.com/openshift/openshift-ansible/issues/1138

brenton:
Yum exits 0 whenever you try to update package that doesn't exist.  In this case it seems like yum doesn't handle the obsoletes correctly.  It knows that the atomic-openshift package obsoletes openshift but it doesn't know atomic-openshift-3.1.0.4 obsoletes openshift.  It just exits 0.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1070,6 +1070,26 @@ def set_container_facts_if_unset(facts):
...
 
+def set_installed_variant_rpm_facts(facts):
+    """ Set RPM facts of installed variant
+        Args:
+            facts (dict): existing facts
+        Returns:
+            dict: the facts dict updated with installed_variant_rpms
+                          """
+    installed_rpms = []
+    for base_rpm in ['openshift', 'atomic-openshift', 'origin']:
```
brenton: @detiber, originally this method was using the service_type however that proved to be a little complicated at upgrade time.  In the case of the 3.0->3.1 upgrade the service type can be different than what is actually installed.  If this looks OK I'd like to merge this in for QE to test.

detiber: :+1:

PR #1337, opened 2016-02-04
Support for adding new masters

abutcher:
**Adding Masters and Nodes**

Modify host inventory, adding a host group for `new_masters` and `new_nodes`:

We currently require that masters also be nodes so that masters are part of the OpenShift SDN. For each master added to `new_masters`, that master should be added to `new_nodes`.

```
[OSEv3:children]
masters
nodes
etcd
lb
new_masters
new_nodes

...

[new_masters]
master4.example.com

[new_nodes]
master4.example.com
```

Run the master scaleup playbook:

```
ansible-playbook -i ~/ansible-inventory openshift-ansible/playbooks/byo/openshift-master/scaleup.yml
```

Once scale up is complete, modify inventory to add hosts in the `new_nodes` host group to the `nodes` host group and remove the `new_nodes` host group. Do the same with `new_masters` hosts.

```
[OSEv3:children]
masters
nodes
etcd
lb

...

[masters]
...
master4.example.com

[nodes]
...
master4.example.com
```

**Adding Nodes**

Nodes may be scaled up independently of masters by using the node scale up playbook.

Modify host inventory, adding a host group for `new_nodes`:

```
[OSEv3:children]
masters
nodes
new_nodes

...

[new_nodes]
node4.example.com openshift_node_labels="{'region': 'infra'}"
```

Run the node scaleup playbook:

```
ansible-playbook -i ~/ansible-inventory openshift-ansible/playbooks/byo/openshift-node/scaleup.yml
```

Once scale up is complete, modify inventory to add hosts in the `new_nodes` host group to the `nodes` host group and remove the `new_nodes` host group.

```
[OSEv3:children]
masters
nodes

...

[nodes]
...
node4.example.com openshift_node_labels="{'region': 'infra'}"
```

**Other Stuff**

Adds `modify_yaml` ansible module which can update key/value pairs in a yaml file.

```
- modify_yaml:
    dest: /etc/origin/master/master-config.yaml
    yaml_key: 'kubernetesMasterConfig.masterCount'
    yaml_value: 2
```

Adds protected facts which have conditions for changing such as `openshift.master.master_count` which can only increase and `openshift.master.ha` which cannot change. If `openshift_facts` attempts to make a change that is not allowed the module will fail and a message will be displayed to the user. It is possible to override this behavior by passing the `protected_facts_to_overwite` argument to `openshift_facts`. For example:

```
- openshift_facts:
    role: master
    local_facts:
      master_count: 3
      ha: False
    protected_facts_to_overwrite:
    - "master.master_count"
    - "master.ha"
```

Master certificates are now generated per master rather than copied from the first master. Each master serving certificate will contain SAN entries for: master hostname, master ip, cluster hostname, public cluster hostname, kubernetes service ip, and all relevant service names. For example, if the master hostname was `master1.example.com`  and that master's IP was `192.168.122.96` and a cluster/public_hostname such as `openshift-ansible-cluster.example.com` had been set, then the SAN entries would look like this:

```
X509v3 Subject Alternative Name:
                DNS:kubernetes, DNS:kubernetes.default, DNS:kubernetes.default.svc, DNS:kubernetes.default.svc.cluster.local, DNS:master1.example.com, DNS:openshift, DNS:openshift-ansible-cluster.example.com, DNS:openshift.default, DNS:openshift.default.svc, DNS:openshift.default.svc.cluster.local, DNS:172.30.0.1, DNS:192.168.122.96, IP Address:172.30.0.1, IP Address:192.168.122.96
```


detiber:
I'm not sure that the additional master certs would have a problem. The existing client certs would work for the new master and the master's client certs will work with the old masters.

The new certs would contain more SNI entries than the old certs to include their names, so there should be no issue there (existing masters have certs that are associated with their names, and the new master(s) have certs that are associated with their names).

playbooks/common/openshift-master/config.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: Set master facts and determine if external etcd certs need to be generated
-  hosts: oo_masters_to_config
+  hosts: oo_first_master:oo_masters_to_config
```
detiber: oo_first_master shouldn't need to be in here

playbooks/common/openshift-master/config.yml
```diff
@@ -140,7 +140,7 @@
     when: etcd_client_certs_missing is defined and etcd_client_certs_missing
 
 - name: Determine if master certificates need to be generated
-  hosts: oo_masters_to_config
+  hosts: oo_first_master:oo_masters_to_config
```
detiber: same here.

abutcher:
Yeah this works from a cluster perspective but not if someone added master4.example.com and something was going to talk directly to master4.example.com rather than the cluster/public_hostname. This shouldn't be the case though.

playbooks/common/openshift-master/config.yml
abutcher: I think it might be for the `certs_missing` facts. I'll dig into it.

detiber: certs_missing should just create a list to iterate through in openshift_master_certificates

detiber:
Sure it would, now if all of the hostnames pointed to the LB and they expected them to all work, that is another story. The certificate generated for master4.example.com would included through the all_hostnames variables, and the generated cert that is transferred to it will include master4.example.com

abutcher:
Adding master6.example.com (master[4:5] are my original masters):

```
$ oc login https://master6.example.com:8443/
Unable to connect to the server: x509: certificate is valid for kubernetes, kubernetes.default, kubernetes.default.svc, kubernetes.default.svc.cluster.local, master4.example.com, master5.example.com, openshift, openshift-ansible-cluster.test.example.com, openshift-ansible.test.example.com, openshift.default, openshift.default.svc, openshift.default.svc.cluster.local, 172.30.0.1, 192.168.122.224, 192.168.122.96, not master6.example.com
```

playbooks/common/openshift-master/config.yml
abutcher: Current required here because `openshift_master_ca` uses `master_certs_missing`.

roles/openshift_master_certificates/tasks/main.yml
```diff
@@ -6,48 +6,24 @@
...
+  -
+    - ca.crt
+    - ca.key
+    - ca.serial.txt
 
 - name: Create the master certificates if they do not already exist
   command: >
     {{ openshift.common.admin_binary }} create-master-certs
-      --hostnames={{ item.openshift.common.all_hostnames | join(',') }}
+      --hostnames={{ master_hostnames | join(',') }}
```
detiber: I'm not sure we need master_hostnames here, should item.openshift.common.all_hostnames still fit the bill for each master?

abutcher: It'll work great for initial install but scaleup only hits one master so we need to add all master hostnames somehow.

abutcher:
aos-ci-test

abutcher:
aos-ci-test

playbooks/common/openshift-master/config.yml
```diff
@@ -404,23 +385,3 @@
   tasks:
   - file: name={{ g_master_mktemp.stdout }} state=absent
     changed_when: False
-
```
detiber: Looks like 'Enable cockpit' and 'Configure flannel' should also be in additional_config.yaml as well.

playbooks/byo/openshift-cluster/cluster_hosts.yml
```diff
@@ -5,6 +5,10 @@ g_lb_hosts:     "{{ groups.lb | default([]) }}"
 
 g_master_hosts: "{{ groups.masters | default([]) }}"
 
+g_new_master_hosts: "{{ groups.new_masters | default([]) }}"
+
+g_all_master_hosts: "{{ g_master_hosts | union(g_new_master_hosts) | default([]) }}"
+
```
detiber: make sure these groups are added to the cluster_hosts.yaml for aws, gce, libvirt and openstack

abutcher:
Note: make sure that master configuration isn't updated when only new_nodes are specified

abutcher:
@detiber PTAL

My original paths for the scaleup playbooks would have been `playbooks/byo/openshift-cluster/scaleup/{masters,nodes}.yml` but `playbooks/byo/openshift-{master,node}/scaleup.yml` work better for `vars_files` pathing.

playbooks/common/openshift-cluster/config.yml
```diff
@@ -11,4 +11,6 @@
 
 - include: ../openshift-master/config.yml
 
+- include: ../openshift-master/additional_config.yml
```
detiber: Should additional_config be moved under openshift_cluster instead...  While the items are run on the first master, they are more cluster related configuration.

playbooks/common/openshift-cluster/evaluate_groups.yml
```diff
@@ -29,6 +29,14 @@
       msg: The nfs group must be limited to one host
     when: (groups[g_nfs_hosts] | default([])) | length > 1
 
+  - name: Evaluate oo_masters
+    add_host:
+      name: "{{ item }}"
+      groups: oo_masters
+      ansible_ssh_user: "{{ g_ssh_user | default(omit) }}"
+      ansible_sudo: "{{ g_sudo | default(omit) }}"
+    with_items: "{{ g_all_master_hosts | default([]) }}"
```
detiber: Should we skip the definition of g_all_master_hosts in the cluster_host.yml files, and just set oo_masters to the union of g_master_hosts and g_new_master_hosts?

playbooks/common/openshift-cluster/evaluate_groups.yml
```diff
@@ -37,13 +45,16 @@
       ansible_sudo: "{{ g_sudo | default(omit) }}"
     with_items: "{{ g_etcd_hosts | default([]) }}"
 
+  - set_fact:
+      g_master_hosts_to_config: "{{ g_new_master_hosts | default(g_master_hosts | default([], true), true) }}"
```
detiber: I'd skip the set_fact here and just use `g_new_master_hosts | default(g_master_hosts | default([], true), true)` in the with_items below.

playbooks/common/openshift-cluster/scaleup.yml
```diff
@@ -1,6 +0,0 @@
----
-- include: evaluate_groups.yml
-
-- include: ../openshift-node/config.yml
-  vars:
-    openshift_deployment_type: "{{ deployment_type }}"
```
detiber: We'll need to make sure that the docs get updated for this change (or at least the docs PR/Issue and/or the docs bz)

playbooks/common/openshift-master/additional_config.yml
```diff
@@ -0,0 +1,51 @@
...
+    when: not openshift.common.is_atomic and ( deployment_type in ['atomic-enterprise','openshift-enterprise'] ) and
+      (osm_use_cockpit | bool or osm_use_cockpit is undefined )
+
+- name: Configure flannel
+  hosts: oo_first_master
+  vars:
+    etcd_urls: "{{ openshift.master.etcd_urls }}"
+  roles:
+  - role: flannel_register
+    when: openshift.common.use_flannel | bool
```
detiber: We might want to double check that flannel should only be installed on the first master and not all masters...

playbooks/common/openshift-master/config.yml
```diff
@@ -228,7 +228,7 @@
   gather_facts: no
   tasks:
   - set_fact:
-      haproxy_backend_servers: "{{ hostvars | oo_select_keys(groups['oo_masters_to_config']) | oo_haproxy_backend_masters }}"
+      haproxy_backend_servers: "{{ hostvars | oo_select_keys(groups['oo_masters']) | oo_haproxy_backend_masters }}"
```
detiber: Instead of set_fact here, this should probably be set as a var for the role directly..
```
roles:
- role: haproxy
  haproxy_backend_servers: "{{ hostvars | oo_select_keys(groups['oo_masters']) | oo_haproxy_backend_masters }}"
```

We might want to add some docs somewhere to say not to run ansible against the lb group without also specifying the masters group

playbooks/common/openshift-master/config.yml
```diff
@@ -334,8 +334,8 @@
   serial: 1
   vars:
     sync_tmpdir: "{{ hostvars.localhost.g_master_mktemp.stdout }}"
-    openshift_master_ha: "{{ groups.oo_masters_to_config | length > 1 }}"
-    openshift_master_count: "{{ groups.oo_masters_to_config | length }}"
+    openshift_master_ha: "{{ groups.oo_masters | length > 1 }}"
```
detiber: Should we persist the value of openshift_master_ha (and not setting it to false here) to avoid changing the master configuration too substantially if we run against only a single master?

detiber: Either that or have openshift_facts throw an error if the value is different than the persisted setting?

playbooks/common/openshift-master/config.yml
```diff
@@ -334,8 +334,8 @@
...
+    openshift_master_count: "{{ groups.oo_masters | length }}"
```
detiber: Similar to ^ should we throw an error in openshift_facts if they try to reduce the number of masters? And have a separate playbook for removing masters that would bypass that check?

playbooks/common/openshift-master/scaleup.yml
```diff
@@ -0,0 +1,42 @@
...
+  roles:
+  - openshift_facts
+
+- name: Update master count
+  hosts: oo_masters:!oo_masters_to_config
+  serial: 1
+  vars:
+    openshift_master_count: "{{ groups.oo_masters | length }}"
+  roles:
+  - openshift_facts
```
detiber: We should make sure to persist the new master count as well.

playbooks/common/openshift-master/scaleup.yml
```diff
@@ -0,0 +1,42 @@
...
+  post_tasks:
+  - name: Update master count
+    lineinfile:
+      dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+      regexp: '^  masterCount:.*$'
+      line: "  masterCount: {{ openshift_master_count }}"
```
detiber: Any objection to having a generic-ish Ansible module for modifying a yaml file that would parse the yaml file modify a specify key in jinja2 '.' notation and then write the file back out?

I'm a bit worried that we may have unintentional breakage if the config file is not in a pretty printed format.

playbooks/common/openshift-master/scaleup.yml
```diff
@@ -0,0 +1,42 @@
...
+    register: master_config
+  - name: Restart master api
+    service: name={{ openshift.common.service_type }}-master-api state=restarted
+    when: master_config | changed
```
detiber: Do we actually need to restart the api servers here? or is it just the controllers that use the masterCount setting?

detiber: Also, having the service restarts as handlers might be a better than having them in post_tasks as well.

playbooks/common/openshift-cluster/scaleup.yml
abutcher: I don't think we have docs for the advanced or quick installer process. I'll double check and create an issue.

playbooks/common/openshift-master/scaleup.yml
abutcher: Genius level

playbooks/common/openshift-cluster/config.yml
abutcher: Yeah, I'll move these to openshift-cluster

playbooks/common/openshift-master/scaleup.yml
abutcher: Confirmed it has to be both or the masters will thrash updating kubernetes endpoints. Moving them to handlers

playbooks/common/openshift-cluster/scaleup.yml
abutcher: https://access.redhat.com/solutions/2150381

detiber: https://bugzilla.redhat.com/show_bug.cgi?id=1304954

abutcher:
@detiber Check out the protected facts bits in the last commit. Protected facts follow the same path as additive ones and the logic is in `merge_facts`.

abutcher:
@detiber Warnings from the command module aren't bubbled up and it looks like they're just there for verbose output. In order to get warnings out of merge_facts using debug we'd have to refactor openshift facts a bit. It might be better to have the module fail for now and slate that work for later?

abutcher:
Also, I don't expect anyone will be using --limit since there's a 99% chance the playbooks would explode so I doubt most folks would hit the protected facts

detiber:
@abutcher the limit issue is something we should be working to actively address, since it will cause major issues with any large deployment.

:+1: on using module_fail for now and adding a card for refactoring openshift_facts for outputing warnings.

abutcher:
@detiber Added module fails PTAL

abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
aos-ci-test

detiber:
lgtm

abutcher:
@brenton :rocket:

nak3:
@abutcher @detiber @brenton 
Could you please update the documentation after you guys commit something affects to the current configuration or release?

In this case, `/usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/scaleup.yml` has been obsoleted by this patch.

https://docs.openshift.org/latest/install_config/install/advanced_install.html#adding-nodes-advanced
https://docs.openshift.com/enterprise/3.2/install_config/install/advanced_install.html#adding-nodes-advanced
https://docs.openshift.com/enterprise/3.1/install_config/install/advanced_install.html#adding-nodes-advanced

If it should be updated by doc team, then please solve this by adding `need-doc-update` tag or something. Thanks...

abutcher:
Hey @nak3, this PR corrects the path https://github.com/openshift/openshift-docs/pull/2135

detiber:
@abutcher do we have a bz to track adding the master scaleup to the docs? I didn't see any PRs or issues on openshift-docs for it.

abutcher:
@detiber I don't think there is one, should we make an issue or BZ?

detiber:
@abutcher I'd say an issue and tag in one of the docs folks for eyes.

abutcher:
Issue for documenting how to add masters w/ ansible: https://github.com/openshift/openshift-docs/issues/2210

nak3:
Thank you so much. This saves our heavy workload! (Many users asked us about doc's conflicts and we need to explain it...)

abutcher:
Sure thing @nak3 sorry for the confusion!

PR #1345, opened 2016-02-05
[WIP] Nodes resolve cluster.local via dnsmasq

sdodson:
Resolve cluster dns on nodes by configuring dnsmasq and configuring NetworkManager to use it.

TODO:
- [ ] Add a task to move skydns to 8053 and restart master
- [ ] Either configure /etc/resolv.conf with default interface's ip or configure kubelet to add the node ip to the resolver

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -173,6 +173,15 @@
       dest: "{{ node_cert_dir }}"
     when: certs_missing
 
+- name: Configure node dnsmasq
+  hosts: oo_nodes_to_config:!oo_containerized_master_nodes
+  vars:
+    dns_domain: "{{ openshift.common.dns_domain }}"
+    kube_svc_ip: "{{ hostvars[groups.oo_first_master.0].openshift.master.kube_svc_ip }}"
+    portal_net: "{{ openshift.node.portal_net }}"
+  roles:
+    - openshift_node_dns
+
 - name: Evaluate node groups
   hosts: localhost
   become: no
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -178,7 +178,7 @@
   vars:
     dns_domain: "{{ openshift.common.dns_domain }}"
     kube_svc_ip: "{{ hostvars[groups.oo_first_master.0].openshift.master.kube_svc_ip }}"
-    portal_net: "{{ openshift.node.portal_net }}"
+    portal_net: "{{ openshift.common.portal_net }}"
   roles:
     - openshift_node_dns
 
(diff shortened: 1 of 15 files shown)


roles/flannel_register/defaults/main.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-flannel_network: "{{ openshift.master.portal_net | default('172.30.0.0/16', true) }}"
+flannel_network: "{{ openshift.common.portal_net | default('172.30.0.0/16', true) }}"
```
detiber: If we are going to switch this from master/node to common, we'll need to write a migration script to do it.

I started to implement this here: https://github.com/detiber/openshift-ansible/blob/openshift_facts/roles/openshift_facts/library/openshift_facts.py#L1215-L1241 but never finished cleaning up that PR to go further with it.

roles/openshift_master/tasks/main.yml
```diff
@@ -25,7 +25,13 @@
 
 - name: Set master facts
   openshift_facts:
-    role: master
+    role: "{{ item.role }}"
+    local_facts: "{{ item.local_facts }}"
+  with_items:
+  - role: common
+    local_facts:
+      portal_net: "{{ openshift_master_portal_net | default(None) }}"
```
detiber: This should be pushed down into openshift_common, we should start avoiding setting role facts outside of the role itself.

roles/openshift_node_dns/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
+---
+- name: Install NetworkManager
+  action: "{{ ansible_pkg_mgr }} name=NetworkManager state=present"
+  
+- name: Switch resolver to dnsmasq
+  copy:
+    src: origin-cluster-resolver.conf
```
detiber: Should we be copying this file over as is, or should we just use the ini_file module to ensure it's there and has the values we expect?

playbooks/common/openshift-node/config.yml
```diff
@@ -173,6 +173,15 @@
       dest: "{{ node_cert_dir }}"
     when: certs_missing
 
+- name: Configure node dnsmasq
+  hosts: oo_nodes_to_config:!oo_containerized_master_nodes
+  vars:
+    dns_domain: "{{ openshift.common.dns_domain }}"
+    kube_svc_ip: "{{ hostvars[groups.oo_first_master.0].openshift.master.kube_svc_ip }}"
+    portal_net: "{{ openshift.common.portal_net }}"
```
detiber: Since these are stored in openshift_facts, we should just add a dependency for openshift_facts in openshift_node_dns

We probably want to store the kube_svc_ip var in openshift_facts for all hosts as well. Maybe cluster_dns_ip?

sdodson:
Reworked in #1588

PR #1347, opened 2016-02-08
Fix enabling iptables for latest rhel versions

detiber:
Latest RHEL versions will mask the iptables service if it is installed when firewalld is installed and enabled. This PR reworks the ordering to disable/mask firewalld before installing iptables-services.


detiber:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
LGTM

detiber:
@brenton on further digging, i think this was actually caused by an issue when re-creating my local dev environment, but it does still improve the ordering of the configuration.

PR #1355, opened 2016-02-09
Fix issue when there are no infra nodes

lhuard1A:
It seems that a parenthesis was not at the same place in the definition of `g_{all,etcd,lb,nfs,master,node}_hosts` and in the definition of `g_{infra,compute}_hosts`.
It is problematic when there is no `infra` node at all. As soon as there is one `infra` node somewhere, the error will remain hidden, even if the cluster being spawned has no `infra` node.


detiber:
:+1: @twiest

twiest:
:+1:

PR #1359, opened 2016-02-09
Fix a bug with existing CNAME records

rharrison10:
Realized I wasn't accounting for the trailing . in proper DNS records.


kwoodson:
:+1:

PR #1362, opened 2016-02-09
Fix loopback cluster name, context name, and user

detiber:
This PR fixes the loopback config for a multi-master setup, where previously it was still using the internal clustered DNS name. 

Currently blocked on the following bug: https://bugzilla.redhat.com/show_bug.cgi?id=1306011

Will need to be conditionalized to not use the loopback config unless the version includes the fix from the above BZ.

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -525,10 +525,11 @@ def set_url_facts_if_unset(facts):
                                                                    ports[prefix]))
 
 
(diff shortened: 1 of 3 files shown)


detiber:
Origin PR: https://github.com/openshift/origin/pull/7197

detiber:
@abutcher ptal

detiber:
@aos-ci-test

abutcher:
aos-ci-test

abutcher:
@brenton lgtm

PR #1364, opened 2016-02-09
Add missing atomic- and openshift-enterprise

codificat:
Some checks related to enterprise deployments were still only looking for "enterprise" deployment_type.
Update them to cover "atomic-enterprise" and "openshift-enterprise" for deployment type.


sdodson:
LGTM, I don't think our CI jobs test the subscription bits, but none the less...
aos-ci-test

detiber:
:+1: @brenton

PR #1365, opened 2016-02-09
Specify default namespace when creating router

pat2man:
In case we have switched projects before running ansible.


detiber:
aos-ci-test

detiber:
ok to test

detiber:
@pat2man We are currently working on implementing a module for running these commands to make sure that changes made to the ansible user's client config will not cause conflicts with the commands run by openshift-ansible. This is a good interim step until that is available though. Thanks for the contribution.

@brenton :shipit:

PR #1368, opened 2016-02-10
Add gte check for 3.2, update version checks to gte

detiber:


detiber:
aos-ci-test

detiber:
@sdodson ptal

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -732,16 +732,20 @@ def set_version_facts_if_unset(facts):
...
+                version_gte_3_1_1_or_1_1_1 = LooseVersion(version) >= LooseVersion('3.1.1')
+                version_gte_3_2_or_1_2 = LooseVersion(version) >= LooseVersion('3.2.0')
         else:
-            version_gt_3_1_or_1_1 = True
-            version_gt_3_1_1_or_1_1_1 = True
-        facts['common']['version_greater_than_3_1_or_1_1'] = version_gt_3_1_or_1_1
-        facts['common']['version_greater_than_3_1_1_or_1_1_1'] = version_gt_3_1_1_or_1_1_1
+            version_gte_3_1_or_1_1 = True
+            version_gte_3_1_1_or_1_1_1 = True
+            version_gte_3_2_or_1_2 = True
```
sdodson: @smunilla @detiber The goal of the else case here, where get_openshift_version() returns none is to account for containerized installs where it currently returns None?

sdodson:
@brenton LGTM, my only question is unrelated to this particular PR, so :shipit:

PR #1370, opened 2016-02-10
add iscsi storage_plugin dependency

vanveele:
This adds a storage_plugin task for iSCSI dependency configurable using the existing openshift.node.storage_plugin_deps variable. This does not change the default value for this variable.

### DIFF
--- a/inventory/byo/hosts.aep.example
+++ b/inventory/byo/hosts.aep.example
@@ -142,7 +142,7 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 
 # default storage plugin dependencies to install, by default the ceph and
 # glusterfs plugin dependencies will be installed, if available.
-#osn_storage_plugin_deps=['ceph','glusterfs']
+#osn_storage_plugin_deps=['ceph','glusterfs','iscsi']
 
 # default selectors for router and registry services
 # openshift_router_selector='region=infra'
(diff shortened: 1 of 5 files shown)


detiber:
@vanveele looks good. I think we'll want to enable the iSCSI dependency installation by default, it can be updated here: https://github.com/vanveele/openshift-ansible/blob/iscsi-plugin/roles/openshift_facts/library/openshift_facts.py#L714-L715

Also, looking there, it looks like we should be installing the dependencies by default for the 'origin' and 'online' deployment_types in addition to the 'openshift-enterprise' and 'atomic-enterprise' deployment_types.

sdodson:
Confirmed with storage team that iSCSI is fully supported as of 1.0/3.0 so agree it should be added by default.

detiber:
@sdodson we don't enable storage plugin deps by default for 3.0 installs (we exclude the 'enterprise' deployment_type), so the change would only affect 3.1+

vanveele:
should deployment_type 'origin' be included in this list?

detiber:
@vanveele yes, and also the 'online' deployment_type

detiber:
@vanveele actually, scratch that...  just the 'origin' deployment_type needs to be added.

vanveele:
@detiber added as described.

detiber:
ok to test

detiber:
aos-ci-test

detiber:
:+1:  from me. @brenton

PR #1371, opened 2016-02-10
Don't mask master service on atomic.

abutcher:
@fridim Can you try this out?


detiber:
whoops...  

:+1: from me

fridim:
it's ok with this commit. Thanks

abutcher:
@brenton

PR #1377, opened 2016-02-10
Bug 1302970 - update script does not patch router if name is differen…

brenton:
…t from default

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -670,6 +670,33 @@ class FilterModule(object):
 
         return rpms_31
 
+    @staticmethod
+    def oo_pods_match_component(pods, deployment_type, component):
+        """ Filters a list of Pods and returns the ones matching the deployment_type and component
+        """
+        if not isinstance(pods, list):
+            raise errors.AnsibleFilterError("failed expects to filter on a list")
+        if not isinstance(deployment_type, basestring):
+            raise errors.AnsibleFilterError("failed expects deployment_type to be a string")
+        if not isinstance(component, basestring):
+            raise errors.AnsibleFilterError("failed expects component to be a string")
+
+        image_prefix = 'openshift/origin-'
+        if deployment_type in ['enterprise', 'online', 'openshift-enterprise']:
+            image_prefix = 'openshift3/ose-'
+        elif deployment_type == 'atomic-enterprise':
+            image_prefix = 'aep3_beta/aep-'
+
+        matching_pods = []
+        image_regex = image_prefix + component + r'.*'
+        for pod in pods:
+            for container in pod['spec']['containers']:
+                if re.search(image_regex, container['image']):
+                    matching_pods.append(pod)
+                    break # stop here, don't add a pod more than once
+
+        return matching_pods
+
     def filters(self):
         """ returns a mapping of filters to methods """
         return {
@@ -696,4 +723,5 @@ class FilterModule(object):
             "oo_persistent_volumes": self.oo_persistent_volumes,
             "oo_persistent_volume_claims": self.oo_persistent_volume_claims,
             "oo_31_rpm_rename_conversion": self.oo_31_rpm_rename_conversion,
+            "oo_pods_match_component": self.oo_pods_match_component,
         }
--- a/playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
@@ -74,6 +74,11 @@
 
   - set_fact:
       g_new_version: "{{ g_aos_versions.curr_version.split('-', 1).0 if g_aos_versions.avail_version is none else g_aos_versions.avail_version.split('-', 1).0 }}"
+    when: openshift_pkg_version is not defined
+
+  - set_fact:
+      g_new_version: "{{ openshift_pkg_version | replace('-','') }}"
+    when: openshift_pkg_version is defined
 
   - fail:
       msg: This playbook requires Origin 1.0.6 or later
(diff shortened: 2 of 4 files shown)


brenton:
- [ ]  Figure out a better way to pass in the pods to `with_items`
- [ ]  Update the minor upgrade playbooks
- [ ]  Add the new liveness probes to the 3.1.1 playbook

filter_plugins/oo_filters.py
```diff
@@ -670,6 +670,34 @@ def oo_31_rpm_rename_conversion(rpms, openshift_version=None):
...
 
+    @staticmethod
+    def oo_pods_match_component(pods, deployment_type, component):
+        """ Filters a list of Pods and returns the ones matching the variant.
+        """
+        #pdb.set_trace()
+        if not isinstance(pods, list):
+            raise errors.AnsibleFilterError("failed expects to filter on a list")
+        if not isinstance(deployment_type, basestring):
+            raise errors.AnsibleFilterError("failed expects openshift_version to be a string")
```
detiber: error message does not match failure.

filter_plugins/oo_filters.py
```diff
@@ -670,6 +670,34 @@ def oo_31_rpm_rename_conversion(rpms, openshift_version=None):
...
+        if not isinstance(component, basestring):
+            raise errors.AnsibleFilterError("failed expects openshift_version to be a string")
```
detiber: error message does not match failure.

playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
```diff
@@ -567,46 +572,49 @@
...
 
   - name: Update deployment config to 1.0.4/3.0.1 spec
-    when: _default_router.rc == 0
+    when: _all_routers.rc == 0
     command: >
-      {{ oc_cmd }} patch dc/router -p
+      {{ oc_cmd }} patch dc/{{ item['labels']['deploymentconfig'] }} -p
       '{"spec":{"strategy":{"rollingParams":{"updatePercent":-10},"spec":{"serviceAccount":"router","serviceAccountName":"router"}}}}'
       --api-version=v1
+    with_items: (_all_routers.stdout | from_json)['items'] | oo_pods_match_component(openshift_deployment_type, 'haproxy-router') | oo_select_keys_from_list(['metadata'])
```
brenton: @detiber, @abutcher, I really struggled to make this do the right thing and right now there's a lot of duplicate logic.  What would be the best way to save the matching pods to another variable that could be reused?  I was probably doing something wrong at the time but the last time I tried using set_fact it was messing things up.  It that should work for a data structure like this then I can try again.

playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
```diff
@@ -74,6 +74,11 @@
 
   - set_fact:
       g_new_version: "{{ g_aos_versions.curr_version.split('-', 1).0 if g_aos_versions.avail_version is none else g_aos_versions.avail_version.split('-', 1).0 }}"
+    when: openshift_pkg_version is not defined
+
+  - set_fact:
+      g_new_version: "{{ openshift_pkg_version | replace('-','') }}"
+    when: openshift_pkg_version is defined
```
brenton: This is just another bug fix.  Previously if you set openshift_pkg_version to something less than the latest the correct RPM would be installed but the router would be updated to the latest image.

filter_plugins/oo_filters.py
```diff
@@ -670,6 +670,34 @@ def oo_31_rpm_rename_conversion(rpms, openshift_version=None):
...
+
+        image_prefix = 'openshift/origin-'
+        if deployment_type in ['enterprise', 'online', 'openshift-enterprise']:
+            image_prefix = 'openshift3/ose-'
+        elif deployment_type == 'atomic-enterprise':
+            image_prefix = 'aep3_beta/aep-'
```
detiber: Do we want 'aep3_beta' here or just 'aep3'?

playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
```diff
@@ -567,46 +572,49 @@
   - role: openshift_examples
     openshift_examples_import_command: replace
   pre_tasks:
-  - name: Check for default router
+  - name: Collect all routers
     command: >
-      {{ oc_cmd }} get -n default dc/router
-    register: _default_router
+      {{ oc_cmd }} get pods --all-namespaces -l 'router' -o json
+    register: _all_routers
```
detiber: Variables with leading underscores cause issues with Ansible 2.0

playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
detiber: @brenton set_fact *should* work, you just might have to throw a from_yaml filter in there if it is being treated as a string.

playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
detiber: gotta love those edge cases :)

filter_plugins/oo_filters.py
brenton: I think we'll need to take a pass to swap these out as part of a separate PR.

detiber: Sounds good to me

playbooks/common/openshift-cluster/upgrades/v3_0_to_v3_1/upgrade.yml
brenton: I'll try using set_fact again now that I've changed my approach a bit.  Originally I was trying to do all this without a custom filter but I couldn't get it to work right.

brenton:
There really isn't a good way to update the liveness probes until the cli supports the other patch operations: https://github.com/kubernetes/kubernetes/blob/master/docs/devel/api-conventions.md#patch-operations

Today only strategic merge is available from the cli which doesn't provide a way to replace a liveness probe.  Take this for example:

`oc patch dc/custom-router -p '{"spec":{"template":{"spec":{"containers":[{"name":"router","livenessProbe":{"httpGet":{"path": "/healthz", "port": 1936, "host": "localhost", "scheme": "HTTP"},"initialDelaySeconds":10,"timeoutSeconds":1}}]}}}}'
The DeploymentConfig "custom-router" is invalid.
spec.template.spec.containers[0].livenessProbe: invalid value '<*>(0xc20f13d1d0){Exec:<nil> HTTPGet:<*>(0xc20ee3d4a0){Path:/healthz Port:1936 Host:localhost Scheme:HTTP} TCPSocket:<*>(0xc20c74d400){Port:80}}', Details: exactly 1 handler type is required`

What it's doing is trying to update the liveness probe with the new httpGet probe and still leave the TCPSocket probe in place.  As a workaround I thought I could effectively erase the probe by doing this:

`oc patch dc/custom-router -p '{"spec":{"template":{"spec":{"containers":[{"name":"router","livenessProbe":{}}]}}}}'
"custom-router" patched`

However, that results in no change.  I see upstream kube talking about adding cli support for the additional patch operations so my vote is to wait until then. 

cc @detiber, @sdodson

detiber:
@brenton would it be better to avoid the patch and just update the DC directly? or does that suffer from the same issue?

sdodson:
I'm fine with waiting for this particular change, we'll want to evaluate future changes as some we may not be able to wait for. Depending on how fast kube adds additional patch operations perhaps we should consider including jq (https://stedolan.github.io/jq/) in our repos so we can perform advanced json modification?

brenton:
@sdodson, I like the idea of using jq if it comes down to that.  To @detiber's point I just didn't want to deal with the additional complexity of having to carefully modify the entire json object just yet.  These liveness probes aren't actually causing a real problem right now (at least not once 3.1.1.6 is released) and I'd rather wait until we have either better access to the patch API or something like jq available.

playbooks/common/openshift-cluster/upgrades/v3_1_minor/pre.yml
```diff
@@ -41,7 +41,7 @@
       g_new_service_name: "{{ 'origin' if deployment_type =='origin' else 'atomic-openshift' }}"
 
   - name: Determine available versions
-    script: ../files/versions.sh {{ g_new_service_name }} openshift
+    script: ../files/versions.sh {{ g_new_service_name }}
```
brenton: @sdodson, @detiber, this is another one of those things that I have no idea how anyone's 3.1.1 upgrade actually worked.  This needs to be fixed asap.  In the minor upgrade we only need to know the latest version of the atomic-openshift package.

brenton:
@detiber, PTAL.  I think this is ready for QE.  I'll write up all the various test cases they should try in the BZ.

playbooks/common/openshift-cluster/upgrades/v3_1_minor/pre.yml
sdodson: I think this attempted to handle cases where they're upgrading within 3.0.x. The `sort -r` in versions.sh should make this a non issue assuming that atomic-openshift version is higher than that of the available openshift version. I think this is safe to remove at this point so +1

brenton: Yeah, it's needed for 3.0.z upgrades.  Just not for 3.1.z.

detiber:
I'm not sure we need something like jq, especially since we are talking about wrapping the client/admin commands in a module. At that point we can just manipulate the json within the module if needed.

@brenton I say :shipit: to QE

brenton:
@detiber, That makes sense.

PR #1379, opened 2016-02-10
handle being passed an empty group list

joelddiaz:
Previous version would have the behavior where when an empty list is passed, the zbx API call would return a list of all groups. So an action with no groups defined would default to having all groups added to the operations.
Handle this by iterating one at a time through the list of provided groups. When an empty list is provided, and empty list is returned.


kwoodson:
:+1:

PR #1383, opened 2016-02-11
Bug 1306665 - [metrics] update metrics-deployer template to use lates…

brenton:
…t image versions

### DIFF
--- a/roles/openshift_examples/files/examples/v1.1/infrastructure-templates/enterprise/metrics-deployer.yaml
+++ b/roles/openshift_examples/files/examples/v1.1/infrastructure-templates/enterprise/metrics-deployer.yaml
@@ -85,7 +85,7 @@ parameters:
 -
   description: 'Specify version for metrics components; e.g. for "openshift/origin-metrics-deployer:v1.1", set version "v1.1"'
   name: IMAGE_VERSION
-  value: "3.1.0"
+  value: "3.1.1"
 -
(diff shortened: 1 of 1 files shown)


sdodson:
Interesting, for some reason I was under the impression these wouldn't be updated until 3.2 if not 3.3 so I didn't look for any updates.

detiber:
@sdodson @brenton are we going to be releasing updates for each release? If so, then we should probably tie the version number to the binary version....  My other though is if the newer metrics-deployer template will work on older installations...

brenton:
@detiber, hmm.. I like the idea of setting it to the version the rest of the deployment is using.  Would that mean we'd need to templatize this template?

brenton:
Also, I'm guessing it would mean tagging the images differently in the registry.  The questions we have now for the integration services team are is whether or not the new default of 3.1.1 can safely be used in 3.1.0 environments?

jcantrill:
@brenton from an origin perspective, there is a PR https://github.com/openshift/origin/pull/6825 to tag and build with origin releases to keep them in lockstep.  I think it makes sense to do the same with these releases as well.  @ewolinetz can you verify @brenton question using the e2e script?

ewolinetz:
We were able to get the 3.1.1 images for metrics to install and start up on a 3.1.0 base and confirm that metrics for containers are coming in and able to get status from hawkular.

PR #1385, opened 2016-02-11
Global Proxy Config

sdodson:
Installer implementation for https://github.com/openshift/openshift-docs/pull/1563 and master, node, docker HTTP_PROXY, HTTPS_PROXY, NO_PROXY settings.

No actual implementation yet, just pushing this out to facilitate discussion on desired configuration knobs.

CC: @csrwng @adellape

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -69,6 +69,21 @@
     reg_flag: --insecure-registry
   notify:
   - restart docker
+  
+- name: Set Proxy Settings
+  lineinfile:
+    dest: /etc/sysconfig/docker
+    regexp: '^{{ item.reg_conf_var }}=.*$'
+    line: "{{ item.reg_conf_var }}='{{ item.reg_fact_val }}'"
+  with_items:
+    - reg_conf_var: HTTP_PROXY
+      reg_fact_val: "{{ docker_http_proxy | default('') }}"
+    - reg_conf_var: HTTPS_PROXY
+      reg_fact_val: "{{ docker_https_proxy | default('') }}"
+    - reg_conf_var: NO_PROXY
+      reg_fact_val: "{{ docker_no_proxy | default('') | join(',') }}"
+  notify:
+    - restart docker
 
 - name: Set various docker options
   lineinfile:
--- a/roles/openshift_builddefaults/tasks/main.yml
+++ b/roles/openshift_builddefaults/tasks/main.yml
@@ -1,0 +1,24 @@
+---
+- name: Set builddefaults
+  openshift_facts:
+    role: builddefaults
+    # TODO: add ability to define builddefaults env vars sort of like this
+    # may need to move the config generation to a filter however.
+    # openshift_env: "{{ hostvars[inventory_hostname]
+    #                    | oo_merge_dicts(hostvars)
+    #                    | oo_openshift_env }}"
+    # openshift_env_structures:
+    # - 'openshift.builddefaults.env.*'
+    local_facts:
+      http_proxy: "{{ openshift_builddefaults_http_proxy | default(None) }}"
+      https_proxy: "{{ openshift_builddefaults_https_proxy | default(None) }}"
+      no_proxy: "{{ openshift_builddefaults_no_proxy | default(None) }}"
+      git_http_proxy: "{{ openshift_builddefaults_git_http_proxy | default(None) }}"
+      git_https_proxy: "{{ openshift_builddefaults_git_https_proxy | default(None) }}"
+      
+- name: Set builddefaults config structure
+  openshift_facts:
+    role: builddefaults
+    local_facts:
+      config: "{{ openshift_builddefaults_json | default(builddefaults_yaml) }}"
+        
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1327,6 +1327,23 @@ def get_local_facts_from_file(filename):
 
     return local_facts
 
+def sort_unique(alist):
+    """ Sorts and de-dupes a list
+
+        Args:
+            list: a list
+        Returns:
+            list: a sorted de-duped list
+    """
+
+    alist.sort()
+    out = list()
+    for i in alist:
+        if i not in out:
+            out.append(i)
+
+    return out
+
 def safe_get_bool(fact):
     """ Get a boolean fact safely.
 
@@ -1363,6 +1380,7 @@ def set_proxy_facts(facts):
                     common['no_proxy'].extend(common['no_proxy_internal_hostnames'].split(','))
                 common['no_proxy'].append('.' + common['dns_domain'])
                 common['no_proxy'].append(common['hostname'])
+                common['no_proxy'] = sort_unique(common['no_proxy'])
         facts['common'] = common
 
     if 'builddefaults' in facts:
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1337,6 +1337,57 @@ def safe_get_bool(fact):
     """
     return bool(strtobool(str(fact)))
 
+def set_proxy_facts(facts):
+    """ Set global proxy facts and promote defaults from http_proxy, https_proxy,
+        no_proxy to the more specific builddefaults and builddefaults_git vars.
+           1. http_proxy, https_proxy, no_proxy
+           2. builddefaults_*
+           3. builddefaults_git_*
+
+        Args:
+            facts(dict): existing facts
+        Returns:
+            facts(dict): Updated facts with missing values
+    """
+    if 'common' in facts:
+        common = facts['common']
+        if 'http_proxy' in common or 'https_proxy' in common:
+            if 'generate_no_proxy_hosts' in common and \
+                    common['generate_no_proxy_hosts']:
+                if 'no_proxy' in common and \
+                    isinstance(common['no_proxy'], basestring):
+                    common['no_proxy'] = common['no_proxy'].split(",")
+                else:
+                    common['no_proxy'] = []
+                if 'no_proxy_internal_hostnames' in common:
+                    common['no_proxy'].extend(common['no_proxy_internal_hostnames'].split(','))
+                common['no_proxy'].append('.' + common['dns_domain'])
+                common['no_proxy'].append(common['hostname'])
+        facts['common'] = common
+
+    if 'builddefaults' in facts:
+        facts['master']['admission_plugin_config'] = dict()
+        builddefaults = facts['builddefaults']
+        common = facts['common']
+        if 'http_proxy' not in builddefaults and 'http_proxy' in common:
+            builddefaults['http_proxy'] = common['http_proxy']
+        if 'https_proxy' not in builddefaults and 'https_proxy' in common:
+            builddefaults['https_proxy'] = common['https_proxy']
+        if 'no_proxy' not in builddefaults and 'no_proxy' in common:
+            builddefaults['no_proxy'] = common['no_proxy']
+        if 'git_http_proxy' not in builddefaults and 'http_proxy' in builddefaults:
+            builddefaults['git_http_proxy'] = builddefaults['http_proxy']
+        if 'git_https_proxy' not in builddefaults and 'https_proxy' in builddefaults:
+            builddefaults['git_https_proxy'] = builddefaults['https_proxy']
+        if 'admission_plugin_config' not in builddefaults:
+            builddefaults['admission_plugin_config'] = dict()
+        if 'config' in builddefaults and ('http_proxy' in builddefaults or \
+                'https_proxy' in builddefaults):
+            facts['master']['admission_plugin_config'].update(builddefaults['config'])
+        facts['builddefaults'] = builddefaults
+
+    return facts
+
 # pylint: disable=too-many-statements
 def set_container_facts_if_unset(facts):
     """ Set containerized facts.
@@ -1470,7 +1521,8 @@ class OpenShiftFacts(object):
         Raises:
             OpenShiftFactsUnsupportedRoleError:
     """
-    known_roles = ['cloudprovider',
+    known_roles = ['builddefaults',
+                   'cloudprovider',
                    'common',
                    'docker',
                    'etcd',
@@ -1558,6 +1610,7 @@ class OpenShiftFacts(object):
         facts = set_manageiq_facts_if_unset(facts)
         facts = set_aggregate_facts(facts)
         facts = set_etcd_facts_if_unset(facts)
+        facts = set_proxy_facts(facts)
         if not safe_get_bool(facts['common']['is_containerized']):
             facts = set_installed_variant_rpm_facts(facts)
         return dict(openshift=facts)
--- a/roles/openshift_master/templates/atomic-openshift-master.j2
+++ b/roles/openshift_master/templates/atomic-openshift-master.j2
@@ -10,8 +10,14 @@ AWS_SECRET_ACCESS_KEY={{ openshift.cloudprovider.aws.secret_key }}
 {% endif %}
 
 # Proxy configuration
-# Origin uses standard HTTP_PROXY environment variables. Be sure to set
-# NO_PROXY for your master
-#NO_PROXY=master.example.com
-#HTTP_PROXY=http://USER:PASSWORD@IPADDR:PORT
-#HTTPS_PROXY=https://USER:PASSWORD@IPADDR:PORT
+# See https://docs.openshift.com/enterprise/latest/install_config/install/advanced_install.html#configuring-global-proxy
+{% if 'http_proxy' in openshift.common or 'https_proxy' in openshift.common %}
+HTTP_PROXY='{{ openshift.common.http_proxy | default('') }}'
+HTTPS_PROXY='{{ openshift.common.https_proxy | default('')}}'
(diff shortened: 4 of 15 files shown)


sdodson:
@csrwng how's this look as far as configuration? Should I break out gitHttpProxy and gitHttpsProxy or will 99% of the environments want these to match the HTTP_PROXY and HTTPS_PROXY env vars in BuildDefaults?

sdodson:
@detiber Do you think we should add all known node hostnames and master hostnames to the noproxy list by default with an override to disable that behavior?

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -394,6 +394,31 @@ def set_master_selectors(facts):
...
+            facts['common']['noproxy'] = ''
+    
+    if 'master' in facts:
+        if 'builddefaults_http_proxy' not in facts['master']:
+            facts['master']['builddefaults_http_proxy'] = facts['common']['http_proxy']
+        if 'builddefaults_https_proxy' not in facts['master']:
+            facts['master']['builddefaults_https_proxy'] = facts['common']['https_proxy']
+        if 'builddefaults_noproxy' not in facts['master']:
+            facts['master']['builddefaults_noproxy'] = facts['common']['noproxy']
+    return facts
```
detiber: Since these do not require any computation, we should be adding these to the default facts in get_defaults() instead.

roles/openshift_node/tasks/main.yml
```diff
@@ -98,6 +98,12 @@
       line: "OPTIONS=--loglevel={{ openshift.node.debug_level }}"
     - regex: '^CONFIG_FILE='
       line: "CONFIG_FILE={{ openshift_node_config_file }}"
+    - regex: '^HTTP_PROXY='
+      line: "HTTP_PROXY='{{ openshift.common.http_proxy }}''"
+    - regex: '^HTTPS_PROXY='
+      line: "HTTPS_PROXY='{{ openshift.common.https_proxy }}'"
+    - regex: '^NOPROXY='
+      line: "NOPROXY='{{ openshift.common.noproxy }}'"
```
detiber: At this point, I'm wondering if we should just template /etc/sysconfig/{origin,atomic-openshift}-node

roles/openshift_master/templates/native-cluster/atomic-openshift-master-api.j2
```diff
@@ -4,6 +4,6 @@ CONFIG_FILE={{ openshift_master_config_file }}
 # Proxy configuration
 # Origin uses standard HTTP_PROXY environment variables. Be sure to set
 # NO_PROXY for your master
-#NO_PROXY=master.example.com
-#HTTP_PROXY=http://USER:PASSWORD@IPADDR:PORT
-#HTTPS_PROXY=https://USER:PASSWORD@IPADDR:PORT
+NO_PROXY={{ openshift.common.noproxy }}
+HTTP_PROXY={{ openshift.common.http_proxy }}
+HTTPS_PROXY={{ openshift.common.https_proxy }}
```
detiber: If we create a template for atomic-openshift-master, then we should just have a single template and somehow create the right OPTIONS line depending on which file we are writing out...

detiber:
> @csrwng how's this look as far as configuration? Should I break out gitHttpProxy and gitHttpsProxy or will 99% of the environments want these to match the HTTP_PROXY and HTTPS_PROXY env vars in BuildDefaults?

I think it might be a good idea to have them separate, but to have them default to the HTTP_PROXY and HTTPS_PROXY values by default.

> @detiber Do you think we should add all known node hostnames and master hostnames to the noproxy list by default with an override to disable that behavior?

Node names should not matter, since everything will use the node IPs instead. Having the all_hostnames values for all masters would probably be a good idea though.

sdodson:
> Node names should not matter, since everything will use the node IPs instead. Having the all_hostnames values for all masters would probably be a good idea though.

Not sure I follow on the IP address thing. AFAIK HTTP_PROXY and HTTPS_PROXY affects requests regardless of whether or not they're made via hostnames or IP address.

Regarding population of noproxy, what's the best way to achieve this? Those are burried in openshift.common.all_hostnames but only populated with the full set of hostnames for masters. Should I ensure that facts for the master are set and then set openshift.common.noproxy with the union of openshift_noproxy and hostvars[oo_first_master].openshift.common.all_hostnames in the playbooks?

csrwng:
@sdodson I'm back :) So yes, checked with @bparees and for most environments you'll want gitHTTP(S)Proxy to be the same as HTTP(S)_PROXY.

playbooks/common/openshift-master/config.yml
```diff
@@ -42,6 +42,12 @@
           public_hostname: "{{ openshift_public_hostname | default(None) }}"
           public_ip: "{{ openshift_public_ip | default(None) }}"
           deployment_type: "{{ openshift_deployment_type }}"
+          http_proxy: "{{ openshift_http_proxy | default(None) }}"
+          https_proxy: "{{ openshift_https_proxy | default(None) }}"
+          no_proxy: "{{ openshift_no_proxy | default(None) }}"
+          no_proxy_auto_include_hostnames: "{{ openshift_no_proxy_internal_hosts | default(False) }}"
+          no_proxy_internal_hostnames: "{{ hostvars | oo_select_keys(groups['oo_nodes_to_config'] | union(groups['oo_masters_to_config']) | union(groups['oo_etcd_to_config'] | oo_flatten() ))
+                                          | oo_collect('openshift.common.all_hostnames') | oo_flatten() | join(',')}}"
```
sdodson: @detiber / @abutcher any idea how to clean up this monster? Mainly the | union() | union() stuff  I'm trying to collect all known internal hostnames

sdodson:
Should be functionally complete now and most previous review comments addressed.

playbooks/common/openshift-master/config.yml
abutcher: I think this is okay. I would pipe this through `unique` and I'd be worried about missing names during scale up where we have truncated `oo_masters_to_config` and `oo_nodes_to_config`. I'm adding an `oo_masters` (old+new=total) group and I think a similar `oo_nodes` would be good but I can resolve that when we integrate master scaleup.

sdodson:
aos-ci-test

playbooks/common/openshift-master/config.yml
detiber: We do have the g_all_hosts variable...  We could also add a oo_all_hosts to evaluate_groups for consistency.

 Also, I'm not sure that oo_flatten is needed after the union filters.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1007,6 +1007,41 @@ def get_local_facts_from_file(filename):
 
     return local_facts
 
+def set_proxy_facts(facts):
```
detiber: These values aren't going to be persisted... Also, we should probably configure this as an additive fact, so that we don't accidentally remove entries that are set.

Since we are only relying on common facts for the computed values, we should be able to set all of these values within init_local_facts so that we can ensure that the values are persisted.

roles/openshift_master/templates/atomic-openshift-master.j2
```diff
@@ -2,8 +2,9 @@ OPTIONS=--loglevel={{ openshift.master.debug_level }}
...
-# NO_PROXY for your master
-#NO_PROXY=master.example.com
-#HTTP_PROXY=http://USER:PASSWORD@IPADDR:PORT
-#HTTPS_PROXY=https://USER:PASSWORD@IPADDR:PORT
+# See https://docs.openshift.com/enterprise/latest/install_config/install/advanced_install.html#configuring-global-proxy
+{% if openshift.common.http_proxy or openshift.common.https_proxy %}
+HTTP_PROXY='{{ openshift.common.http_proxy }}'
+HTTPS_PROXY='{{ openshift.common.https_proxy }}'
+NO_PROXY='{{ openshift.common.no_proxy }}'
+{% endif %}
```
detiber: Should we keep the commented examples in the file in the case that we aren't setting the proxy values?

roles/openshift_master/templates/master.yaml.v1.j2
```diff
@@ -96,6 +96,24 @@ kubernetesMasterConfig:
...
+          kind: BuildDefaultsConfig
+          gitHTTPProxy: {{ openshift.master.builddefaults_git_http_proxy }}
+          gitHTTPSproxy: {{ openshift.master.builddefaults_git_https_proxy }}
+          env:
+          - name: HTTP_PROXY
+            value: '{{ openshift.master.builddefaults_http_proxy }}'
+          - name: HTTPS_PROXY
+            value: '{{ openshift.master.builddefaults_https_proxy }}'
+          - name: NO_PROXY
+            value: '{{ openshift.master.builddefaults_no_proxy }}'
```
detiber: Not related to this PR, but it seems a bit redundant to have to set HTTP_PROXY, HTTPS_PROXY and NO_PROXY for the BuildDefaults here...  it would be nice if the values where injected by the Build Controller, since they are already present in the controllers env.

roles/openshift_master/templates/atomic-openshift-master.j2
sdodson: Yeah, I'll add those back in.

roles/openshift_master/templates/master.yaml.v1.j2
sdodson: I'm probably going to need to revisit the defaults we stuff into the build environment. When testing a build today it dawned on me that we're going to be leaking a lot of infrastructure information into the build environment that's probably not necessary. That said, the web-ui already exposes the node's hostname so this can probably slide for now. I'll look at it some more.

detiber: @sdodson alternatively, we could have it added as a secret instead of using the env.

sdodson: @csrwng Can you comment on this? what we should do here? Mainly we're trying to come up with what the default NO_PROXY should be for the builddefaults controller. What endpoints, internal to the cluster, are used during the build phase? I'm wondering if there aren't any since the only thing that requires internal access would be the docker push and that's happening outside of the build environment via the docker daemon on the node.

sdodson:
Massively updated, anyone interested in using this please give it a shot.

sdodson:
@detiber ready for review, @abutcher took a look and i've cleaned up a few things he noticed already

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
Can't reproduce the issues locally, assuming the problem was actually a flake.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1329,6 +1329,53 @@ def safe_get_bool(fact):
     """
     return bool(strtobool(str(fact)))
 
+def set_proxy_facts(facts):
+    """ Set BuildDefaults admission controller facts
```
detiber: comment doesn't match method name

roles/docker/tasks/main.yml
```diff
@@ -69,6 +69,22 @@
...
+    regexp: '^{{ item.reg_conf_var }}=.*$'
+    line: "{{ item.reg_conf_var }}='{{ item.reg_fact_val }}'"
+  with_items:
+    - reg_conf_var: HTTP_PROXY
+      reg_fact_val: "{{ openshift.common.http_proxy | default('') }}"
+      when: "http_proxy in openshift.common"
+    - reg_conf_var: HTTPS_PROXY
+      reg_fact_val: "{{ openshift.common.https_proxy | default('') }}"
+    - reg_conf_var: NO_PROXY
+      reg_fact_val: "{{ openshift.common.no_proxy | default('') | join(',') }}"
```
detiber: We shoudln't be referencing openshift facts in the docker role. Rather there should be docker_ variables for configuring the proxy and openshift_docker_facts should map the openshift fact values to the docker_ variables

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -73,3 +73,9 @@
       oauth_templates: "{{ openshift_master_oauth_templates | default(None) }}"
       oauth_always_show_provider_selection: "{{ openshift_master_oauth_always_show_provider_selection | default(None) }}"
       image_policy_config: "{{ openshift_master_image_policy_config | default(None) }}"
+      builddefaults_http_proxy: "{{ openshift_builddefaults_http_proxy | default(None) }}"
+      builddefaults_https_proxy: "{{ openshift_builddefaults_https_proxy | default(None) }}"
+      builddefaults_no_proxy: "{{ openshift_builddefaults_no_proxy | default(None) }}"
+      builddefaults_git_http_proxy: "{{ openshift_builddefaults_git_http_proxy | default(None) }}"
+      builddefaults_git_https_proxy: "{{ openshift_builddefaults_git_https_proxy | default(None) }}"
+      builddefaults_config: "{{ openshift_builddefaults_json | default(builddefaults_yaml) }}"
```
detiber: Is there a benefit to overriding the builddefaults_config here vs setting the other variables?

I suspect there might be non-proxy options that could be passed to the builddefaults_config, in which case the user provided config would be expected to override the other values. In that case, I would expect it to default to None rather than the value of builddefaults_yaml.

It also seems like there could be a case for a 'builddefaults' role that would allow for setting the individual variables. I'm wondering if it could also leverage openshift_env to make things a bit cleaner...  @abutcher thoughts?

abutcher: +1 Could use something like this to grab and set all of the `builddefaults` facts ie. `openshift.builddefaults.http_proxy` once the role has been added to `openshift_facts`.

```
- name: Set builddefaults facts
  openshift_facts:
    role: builddefaults
    openshift_env: "{{ hostvars[inventory_hostname]
                       | oo_merge_dicts(hostvars)
                       | oo_openshift_env }}" 
    openshift_env_structures:
    - 'openshift.builddefaults.*' 
```

ganhuang:
@sdodson , Added a comment in BZ#1253675 in case if didn't notice it.: )

roles/docker/tasks/main.yml
sdodson: Looks like openshift_docker_facts only depends on openshift_facts which wouldn't have the whole openshift_common facts, correct? Are you ok with me adding openshift_common as a dependency of openshift_docker_facts?

roles/openshift_master_facts/tasks/main.yml
sdodson: Right now the only non proxy options are additional env vars. I'll try to hack something up to allow arbitrary env vars

roles/docker/tasks/main.yml
detiber: @sdodson we don't want to pull in openshift_common to avoid installing the base package for non-masters and non-nodes that are containerized (etcd is the current case, but as we containerize other roles they will factor in as well). We already set some common facts in openshift_facts, we should push additional ones we need here there as well.

sdodson: ok, think i've addressed this one.

roles/openshift_master_facts/tasks/main.yml
sdodson: I didn't get this sorted out today. Will work on this tomorrow when i can sync with either of you on irc.

roles/openshift_builddefaults/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Set builddefaults
+  openshift_facts:
+    role: builddefaults
+    openshift_env: "{{ hostvars[inventory_hostname]
+                       | oo_merge_dicts(hostvars)
+                       | oo_openshift_env }}"
+    openshift_env_structures:
+    - 'openshift.builddefaults.env.*'
```
detiber: Should this just be 'openshift.builddefaults.*' instead?

abutcher: Indeed. Copy pasta.

detiber:
I'm fine with pushing off support for the env vars until a later time.

It looks like pylint isn't too happy though, and that this might need a rebase as well.

sdodson:
Rebased and cleaned up commit message

sdodson:
aos-ci-test

detiber:
aos-ci-test

detiber:
@sdodson looks like pylint isn't happy

sdodson:
Thanks, sorry I missed that.

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
@sdodson looks like we are hitting some flakes with the instance bootstrapping for tests. If you are ready for this to merge (and we are going to be doing another build today), then I am ready to push the merge button.

sdodson:
@detiber yeah, lets go ahead and get it in

andrewklau:
My HTTP_PROXY is being set to HTTP_PROXY='' without any inventory configuration, I don't think that's the expected behaviour? It breaks my services from even starting due to AWS cloud configuration.

sdodson:
> My HTTP_PROXY is being set to HTTP_PROXY='' without any inventory configuration, I don't think that's the expected behaviour? It breaks my services from even starting due to AWS cloud configuration.

If they haven't merged yet, can you try the fixes in #1808 ?

PR #1387, opened 2016-02-12
Change etcd deamon name for atomic-host in playbooks/adhoc/uninstall.yml

talset:
On atomic host the "rm /var/lib/etcd" fail because etcd still running. The "deamon" name is not etcd

```bash
systemctl -l | grep etcd
  etcd_container.service                                                                    loaded active running   The Etcd Server container
```
Updated to add a different name of service between rhel and atomic host

### DIFF
--- a/playbooks/adhoc/uninstall.yml
+++ b/playbooks/adhoc/uninstall.yml
@@ -53,6 +53,13 @@
         - pcsd
       failed_when: false
 
+    - name: Stop additional atomic services
+      service: name={{ item }} state=stopped
+      when: is_atomic | bool
+      with_items:
+        - etcd_container
+      failed_when: false
+
     - name: Remove packages
       action: "{{ ansible_pkg_mgr }} name={{ item }} state=absent"
       when: not is_atomic | bool


playbooks/adhoc/uninstall.yml
```diff
@@ -39,7 +39,7 @@
         - atomic-openshift-master-api
         - atomic-openshift-master-controllers
         - atomic-openshift-node
-        - etcd
+        - "{% if is_atomic %}etcd_container{% else %}etcd{% endif %}"
```
detiber: In this case, we list out all services that could be encountered...  instead of using the conditional here, it would probably be better to just list etcd_container separately.

As an aside, instead of using `{% if is_atomic %}...{% endif %}` here, you could have done `{{ etcd_container if is_atomic | bool else etcd }}` (The use of the bool filter just ensures that is_atomic is cast as a boolean before comparing in case it is a string representation of a boolean)

talset: Thanks for your ansible tip.

@detiber etcd could be a critical service on atomic and it could be used by other docker containers. Should we assume that this playbook have to be used only on dedicated host for openshift ?
If you validate, I will change my PR to keep etcd and separate etcd_container.

sdodson: We disable the atomic host level etcd service during installation when is_containerized to avoid port conflicts so it shouldn't be used for anything already. I'd say disabling it during uninstall is fine.

talset:
Ok thanks, so I changed the commit to just add the "Stop additional atomic services" section. It stop etcd_container when is_atomic.

abutcher:
@brenton

PR #1388, opened 2016-02-12
a-o-i: Prompts to allow minor upgrades

smunilla:
Updates to the upgrade command to allow the user to select a minor
upgrade.

### DIFF
--- a/utils/src/ooinstall/cli_installer.py
+++ b/utils/src/ooinstall/cli_installer.py
@@ -722,14 +722,30 @@ def upgrade(ctx):
         click.echo("No hosts defined in: %s" % oo_cfg.config_path)
         sys.exit(1)
 
-    # Update config to reflect the version we're targetting, we'll write
-    # to disk once ansible completes successfully, not before.
     old_variant = oo_cfg.settings['variant']
     old_version = oo_cfg.settings['variant_version']
-    if oo_cfg.settings['variant'] == 'enterprise':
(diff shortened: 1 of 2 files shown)


utils/src/ooinstall/cli_installer.py
```diff
@@ -722,14 +722,30 @@ def upgrade(ctx):
         click.echo("No hosts defined in: %s" % oo_cfg.config_path)
         sys.exit(1)
 
-    # Update config to reflect the version we're targetting, we'll write
-    # to disk once ansible completes successfully, not before.
     old_variant = oo_cfg.settings['variant']
     old_version = oo_cfg.settings['variant_version']
```
detiber: Instead of trusting the version in the config, would it be better to call out to the openshift_facts playbook and get the version from the host? Or maybe use a combination of both approaches?

smunilla: Totally makes sense. We should probably also query to figure out which versions they have available too so the user doesn't start an upgrade they can't actually finish.

But, this is just a change to our existing model to let them kick off a minor upgrade instead of assuming that they want to do a major upgrade.

brenton:
I'm merging this since we need this for the 3.1.1.6 update asap.  I definitely like the idea of not relying on user input to get the versions.  @smunilla, can you create an issue or update one of the existing installer cards for that?

PR #1390, opened 2016-02-12
Changes for new Nuage RPMS

vishpat:
@detiber These change are required because we renamed the rpms from kubemon to openshift. There is no change in functionality.


detiber:
ok to test

detiber:
@brenton :+1: from me

PR #1391, opened 2016-02-12
Add missing `type` node labels on OpenStack and libvirt

lhuard1A:
This fixes, for example, the fact that the OpenShift router was not starting on OpenStack and libvirt.


detiber:
:+1: @brenton

Issue #1200, opened 2016-01-18
ImportError: No module named yum


donovanmuller:
This seems to be the same error as #855.

Using the [Fedora 23 GP2 HVM AMI](https://getfedora.org/cloud/download/) (ami-518bfb3b) on AWS I'm getting the following *after* successfully running `playbooks/adhoc/bootstrap-fedora.yml`:

```
$ ansible-playbook -i ... playbooks/adhoc/bootstrap-fedora.yml

PLAY [OSEv3] ******************************************************************

TASK: [install python and deps for ansible modules] ***************************
ok: [ec2-54-172-114-234.compute-1.amazonaws.com]
ok: [ec2-54-174-9-0.compute-1.amazonaws.com]
ok: [ec2-52-90-78-173.compute-1.amazonaws.com]

PLAY RECAP ********************************************************************
ec2-52-90-78-173.compute-1.amazonaws.com : ok=1    changed=0    unreachable=0    failed=0
ec2-54-172-114-234.compute-1.amazonaws.com : ok=1    changed=0    unreachable=0    failed=0
ec2-54-174-9-0.compute-1.amazonaws.com : ok=1    changed=0    unreachable=0    failed=0

$ ansible-playbook -i ... playbooks/byo/config.yml

...
TASK: [openshift_facts | set_fact ] *******************************************
ok: [ec2-54-172-114-234.compute-1.amazonaws.com]
ok: [ec2-54-174-9-0.compute-1.amazonaws.com]
ok: [ec2-52-90-78-173.compute-1.amazonaws.com]

TASK: [openshift_facts | Ensure PyYaml is installed] **************************
...
failed: [ec2-54-172-114-234.compute-1.amazonaws.com] => {"failed": true, "parsed": false}
/etc/profile.d/lang.sh: line 19: warning: setlocale: LC_CTYPE: cannot change locale (UTF-8): No such file or directory
BECOME-SUCCESS-alcqqopudzupdufmdzvyulsdxepfilxk
/bin/sh: warning: setlocale: LC_CTYPE: cannot change locale (UTF-8): No such file or directory
Traceback (most recent call last):
  File "/home/fedora/.ansible/tmp/ansible-tmp-1453123922.84-195523226686556/yum", line 27, in <module>
    import yum
ImportError: No module named yum
OpenSSH_6.9p1, LibreSSL 2.1.8
debug1: Reading configuration data /etc/ssh/ssh_config
[...]

PR #1393, opened 2016-02-12
Fix adhoc boostrap fedora playbook

detiber:
Resolves https://github.com/openshift/openshift-ansible/issues/1200


detiber:
@brenton

PR #1394, opened 2016-02-12
Don't automatically give additional permissions to all OAuth users on upgrade

liggitt:
OAuth-authenticated users have the `system:authenticated:oauth` group as of https://github.com/openshift/origin/pull/7054, which is what the bootstrap policy gives permission to request projects (to ensure service accounts cannot create new projects by default).

We should exclude that group from automatically gaining new permissions on upgrades, or a cluster-admin's decision to remove project request access would be reverted.

The exclude is a no-op if the group doesn't exist.


liggitt:
@sdodson @detiber

detiber:
:+1: @brenton

PR #1395, opened 2016-02-12
Updating the 3.1.1 router to match the new liveness probe configuration

brenton:


brenton:
@abutcher gave me the tip on setting the tcpsocket probe to `null`.  One gotcha is that I still want to allow 3.0->3.1.0 upgrades so had to handle that case.

  I didn't have time to make the registry probe improvements.  I think they were going to be more complicated primarily because of having to handle secured registries.

@abutcher, @sdodson, @detiber PTAL

sdodson:
Only concern i have is if they've secured their router. Let me see real quick how we can figure out if that's the case or not.

sdodson:
> Only concern i have is if they've secured their router. Let me see real quick how we can figure out if that's the case or not.

Sorry, this is router not registry. LGTM

PR #1405, opened 2016-02-15
Bug 1308411 - Fail to install OSE 3.0 for no add-scc-to-user command

brenton:


brenton:
aos-ci-test

brenton:
@sdodson, @abutcher, this is ugly but it gets the job done.

sdodson:
Yeah, ugly but LGTM

abutcher:
:+1:

roles/openshift_serviceaccounts/tasks/main.yml
```diff
@@ -27,7 +27,47 @@
...
   - openshift_serviceaccounts_names
   - scc_test.results
+
+####
+#
+# Support for 3.0.z
+#
+####
+
+- name: tmp dir for openshift
```
detiber: Just getting around to taking a look at this...  would it be better to move this into a separate task file in the role, say `tasks/legacy_add_scc_to_user.yml` and then just add a conditional include:
```
- include: legacy_add_scc_to_user.yml
  when: not openshift.common.version_gte_3_1_or_1_1
```

brenton: That makes sense.  I'll fix that up.

PR #1410, opened 2016-02-16
Apply openshift_docker to nodes during scaleup.

### LINKED ISSUE #1407: scaleup playbook does not configure insecure-registry in sysconfig/docker
I added a couple of nodes like this:

```bash
ansible-playbook -i hosts openshift-ansible/playbooks/byo/openshift-cluster/scaleup.yml
```

Afterwards builds on these nodes failed when they could not push to the registry. The following change to `/etc/sysconfig/docker` and restarted of docker fixes it.

```diff
4c4
< OPTIONS='--selinux-enabled'
---
> OPTIONS='--insecure-registry=172.30.0.0/16 --selinux-enabled   '
36a37,38
> BLOCK_REGISTRY=''
> INSECURE_REGISTRY=''
```

Shouldn't `scaleup.yml` apply the `openshift_docker` role which normally makes this change?

abutcher:
Resolves #1407

@sdodson PTAL.

### DIFF
--- a/playbooks/common/openshift-cluster/scaleup.yml
+++ b/playbooks/common/openshift-cluster/scaleup.yml
@@ -1,6 +1,16 @@
 ---
 - include: evaluate_groups.yml
 
+- name: Configure docker hosts
+  hosts: oo_nodes_to_config
+  vars:
+    docker_additional_registries: "{{ lookup('oo_option', 'docker_additional_registries') | oo_split }}"
+    docker_insecure_registries: "{{ lookup('oo_option',  'docker_insecure_registries') | oo_split }}"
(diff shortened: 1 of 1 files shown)


sdodson:
LGTM, @brenton

sdodson:
aos-ci-test

brenton:
Does this behave if the values from the lookup are empty?

abutcher:
Tested with empty values

sdodson:
Yeah it's fine. I looked at this yesterday when adding openshift_docker_options, the tasks check for undefined. I actually rather like this implementation as it doesn't require anything in openshift_facts but it does litter the tasks with a lot of checks for variable definition.

Issue #1412, opened 2016-02-16
Master cannot access to POD SDN network while created with OpenStack provider


jdnieto:
When select OpenStack as provider, it creates 3 security groups master/node/infra. Master need access to PODs SDN network but there is no rule in their OpenStack security group.

PR #1413, opened 2016-02-16
Fix openstack master sdn port

jdnieto:
Pull request to solve issue [#1412](https://github.com/openshift/openshift-ansible/issues/1412)


sdodson:
@jdnieto Lots of extraneous commits here, do you perhaps need to rebase?

PR #110, opened 2015-03-09
Allow for overriding node resources

detiber:
- add variable openshift_node_resources to openshift_node role
- set default value for openshift_node_resources to
  { capacity: { cpu: ,memory: }}
- If cpu is not set, then the default value will be chosen by the
  openshift_register_node module (num logical cpus)
- If memory is not set, then the default value will be chosen by the
  openshift_register_node module (75% MemTotal according to /proc/meminfo)

PR #1414, opened 2016-02-16
added port 4789 to master security group for access to SDN network

jdnieto:


abutcher:
@jdnieto This adds ingress 4789 to the etcd security group. Should be master security group?

jdnieto:
@abutcher repaired, sorry for the mistake :(

sdodson:
A master needs to be a node in order to be on the SDN, is your master not also a node?

jdnieto:
Yes, my master is also a node. oc get nodes shows my master. But in
openshift-ansible it is created without that port in OpenStack, and their
security group is master without 4789 SDN port.

El mar., 16 feb. 2016 19:36, Scott Dodson <notifications@github.com>
escribió:

> A master needs to be a node in order to be on the SDN, is your master not
> also a node?
>
> —
> <https://github.com/openshift/openshift-ansible/pull/1414#issuecomment-184818410>
> .
>
-- 

Un saludo.
-------------------------------------------
José David Martín Nieto
RHCE, RHCJA, RHCJD, RHCoE-PaaS, RHCoE-Hibernate, RHCI, RHCX
#110-442-589

Public Key 2048R/24C474E5 C8B9 5589 9C19 1453 1184  EEAE A642 701E 24C4
74E5 @pool.sks-keyservers.net

sdodson:
@abutcher explained the situation to me, LGTM

abutcher:
If masters and nodes are in their own security groups then we should probably also add 10250 to master ingress. I'm not very familiar with the openstack bits in openshift-ansible so I'll have to defer to @lhuard1A and @menren

jdnieto:
@abutcher the port 10250 is only needed for connection between master and nodes, and it's already enabled in the node-secgrp. So it's not needed. But I modified also the infra-secgrp for the same reason. I'm thinking to create a new pull request configuring a new group called "common-group" for the common ports and to allocate it together with the corresponding groups in instances.
What do you think?

jdnieto:
I close this pull request. Opened a new one with better description. [#1423](https://github.com/openshift/openshift-ansible/pull/1423)

PR #1415, opened 2016-02-16
bind in /etc/origin/node for non-master monitoring to be able to talk…

joelddiaz:
… with master

### DIFF
--- a/roles/oso_host_monitoring/templates/oso-rhel7-host-monitoring.service.j2
+++ b/roles/oso_host_monitoring/templates/oso-rhel7-host-monitoring.service.j2
@@ -57,11 +57,13 @@ ExecStart=/usr/bin/docker run --name {{ osohm_host_monitoring }}                
            -v /var/lib/docker:/var/lib/docker:ro                                             \
            -v /var/run/docker.sock:/var/run/docker.sock                                      \
            -v /var/run/openvswitch:/var/run/openvswitch                                      \
-{% if hostvars[inventory_hostname]['ec2_tag_host-type'] == 'master' %}
+{% if hostvars[inventory_hostname]['oo_hosttype'] == 'master' %}
            -v /etc/origin/master/admin.kubeconfig:/etc/origin/master/admin.kubeconfig  \
            -v /etc/origin/master/master.etcd-client.crt:/etc/origin/master/master.etcd-client.crt \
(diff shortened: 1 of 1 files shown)


roles/oso_host_monitoring/templates/oso-rhel7-host-monitoring.service.j2
```diff
@@ -62,6 +62,8 @@ ExecStart=/usr/bin/docker run --name {{ osohm_host_monitoring }}
            -v /etc/origin/master/master.etcd-client.crt:/etc/origin/master/master.etcd-client.crt \
            -v /etc/origin/master/master.etcd-client.key:/etc/origin/master/master.etcd-client.key \
            -v /etc/origin/master/master-config.yaml:/etc/origin/master/master-config.yaml \
+{% elif hostvars[inventory_hostname]['ec2_tag_host-type'] == 'node' %}
```
twiest: We should start using the oo_ vars as they'll be cross cloud.

So, `oo_hosttype`.

twiest:
:+1:

detiber:
I have a few questions related to this (mostly around improvements that can be made in the future):
- Why does the monitoring on the nodes need to talk to the master?
- What permissions are needed for this communication? It seems like using the master/node kubeconfigs would probably be granting way too much access for monitoring.
- Why not create and use a service account for master/kubelet communication for monitoring purposes?

joelddiaz:
@detiber , there was a desire for the non-master nodes to be able to test and report on the health of the docker registries. The minimum permissions needed are simply to get the IP address of the docker-registry service and POD endpoint IPs.

I'll discus with @twiest about using the service account, and what the implications might be for these dedicated clusters (will customers see and/or object to these accounts?).

joelddiaz:
@detiber I spoke with @twiest . Basically, these monitoring containers are root/super-privileged and are supposed to run the equivalent of root on the bare metal of the instance/node. Using a service account is something we can definitely look into. And it might make sense to consider this change when we move the monitoring containers to be run/deployed by OpenShift itself.

No concerns about customers as they wouldn't have visibility into the existence of the service account/monitoring containers from the customer's namespace.

detiber:
@joelddiaz thanks for the additional info. Using service accounts will definitely be much easier once the containers are running within OpenShift, since then the generated secrets will be provided within the container automatically.

Using serviceaccounts where possible for accessing the api is generally good practice, especially if the admin privileges aren't needed.

Issue #1422, opened 2016-02-17
Ability to configure imageConfig latest value via inventory


jeremyeder:
We have a use-case for eliminating the specific image version requirements from OpenShift.
We'd like it to allow use of images tagged "latest".

Turns out this is hard-coded into the node config template:
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_node/templates/node.yaml.v1.j2#L12

Is there a way to have openshift-ansible set "latest: true" for me via inventory customization ?

@ofthecurerh

jeremyeder:
https://github.com/openshift/openshift-ansible/pull/1424

PR #1424, opened 2016-02-17
add support for toggling imageconfig via inventory

jeremyeder:
This will assist with automated testing by allowing test candidate images (which have no version tags), to be used.

### DIFF
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -87,6 +87,9 @@
       master_count: "{{ openshift_master_count | default(None) }}"
       controller_lease_ttl: "{{ osm_controller_lease_ttl | default(None) }}"
       master_image: "{{ osm_image | default(None) }}"
+    role: common
+    local_facts:
+      imageconfig_latest: "{{ openshift_imageconfig_latest | default(False) }}"
 
 - name: Install Master package
   action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-master{{ openshift_version  }} state=present"
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -77,6 +77,9 @@ deployment_type=origin
 # cli_docker_additional_registries instead
 #oreg_url=example.com/openshift3/ose-${component}:${version}
 
+# Allow OpenShift to use pods that are tagged "latest" when there is no exact version match
+#openshift_imageconfig_latest=true
+
 # Origin copr repo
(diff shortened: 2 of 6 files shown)


sdodson:
aos-ci-test

sdodson:
lgtm just want to review the output of a ci job to see that the defaults work as expected

sdodson:
Fixes #1422

sdodson:
@jeremyeder Need to do the same for openshift_master and master.yaml.j2 template now that I think about it, can you add that?

jeremyeder:
Done, and tested OK.

```

 $ grep imageconfig inventory/byo/hosts.1424
openshift_imageconfig_latest=true

$ ansible-playbook -i /tmp/openshift-ansible/inventory/byo/hosts.1424 /tmp/openshift-ansible/playbooks/byo/config.yml

<snip>

$ ssh root@ip-172-31-4-39.us-west-2.compute.internal grep latest /etc/origin/*/*-config.yaml
/etc/origin/master/master-config.yaml:  latest: true
/etc/origin/node/node-config.yaml:  latest: true
```

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

jeremyeder:
I don't see how that CI failure relates to this patch.

abutcher:
aos-ci-test

abutcher:
aos-ci-test

roles/openshift_master/tasks/main.yml
```diff
@@ -87,6 +87,9 @@
       master_count: "{{ openshift_master_count | default(None) }}"
       controller_lease_ttl: "{{ osm_controller_lease_ttl | default(None) }}"
       master_image: "{{ osm_image | default(None) }}"
+    role: common
+    local_facts:
+      imageconfig_latest: "{{ openshift_imageconfig_latest | default(False) }}"
```
abutcher: The common role addition will need to be separated into a new openshift_facts task or this task can be modified to iterate over the roles like https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-master/config.yml#L34-L59

detiber: The imageconfig_latest fact should be added in the openshift_common role instead of the openshift_master role...  Since openshift_common is a dependency of openshift_master, it will be picked up just the same as if it was added in this role.

detiber: For setting this, we should also be using the bool filter to ensure that we can accept all of the ansible true/false values and that we do not end up with errors around the value being a string:
```
imageconfig_latest: "{{ openshift_imageconfig_latest | default(False) | bool }}"
```

mtnbikenc:

PR #1425, opened 2016-02-17
a-o-i: Change method for counting master_lb as installed

smunilla:


smunilla:
@abutcher PTAL

abutcher:
Talking to @smunilla this identifies the load balancer as the host matching the cluster_hostname with the goal of excluding the load balancer during node scaleup. LGTM @brenton

PR #1439, opened 2016-02-18
Add extra parameters for S3 registry: delete file, create bucket.

xlr-8:
Thought it would be nice to have options to be able to:
* Delete or not the temporary config file - so that it can be check/modified directly
* Create or not the bucket, as you might not have the right to do so
* Configure the path - to avoid issue if not ran as root

This commit allows those things, without changing the default behavior of the playbook.

### DIFF
--- a/playbooks/adhoc/s3_registry/s3_registry.yml
+++ b/playbooks/adhoc/s3_registry/s3_registry.yml
@@ -15,6 +15,8 @@
     aws_secret_key: "{{ lookup('env', 'S3_SECRET_ACCESS_KEY') }}"
     aws_bucket_name: "{{ aws_bucket | default(clusterid ~ '-docker') }}"
     aws_bucket_region: "{{ aws_region | default(lookup('env', 'S3_REGION') | default('us-east-1', true)) }}"
+    aws_create_bucket: "{{ aws_create | default(True) }}"
+    aws_delete_after_run: "{{ aws_delete_temp | default(True) }}"
 
   tasks:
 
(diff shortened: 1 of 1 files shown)


abutcher:
@dak1n1 @wshearn

playbooks/adhoc/s3_registry/s3_registry.yml
```diff
@@ -30,6 +33,7 @@
     command: oc scale --replicas=0 dc/docker-registry
 
   - name: Create S3 bucket
+    when: aws_create_bucket | bool
```
dak1n1: I'm not sure I understand the purpose of this 'when' statement. Without it, the bucket will only be created if it does not already exist. So it serves as a verification that the registry's backend storage actually exists. Is there a case where we would specifically *not* want to verify that the bucket exists?

dak1n1: I suppose if your AWS credentials don't have permissions to do this, that would be one use case. But in that case, S3_SECRET_ACCESS_KEY wouldn't exist either, since that's supposed to be the credentials with access to the S3 bucket. Offhand this doesn't seem like a common case... but I suppose it doesn't hurt anything. That should be ok.

dak1n1:
:+1:

playbooks/adhoc/s3_registry/s3_registry.yml
talset: Hi,
@dak1n1 For example on our project we only have a IAM with a restricted policy. We only have access to an existing bucket and only for few actions.
The S3_SECRET_ACCESS_KEY are required because we need it into config.yml generated from the j2 template.

So I think it is a good thing to have that (or ignore error for this task)

xlr-8:
Thx

PR #1442, opened 2016-02-18
Fix pv/c creation failed_when condition

### LINKED ISSUE #1441: openshift_persistent_volumes | Create PersistentVolumes fails even though "registry-volume" created
I'm running openshift-ansible against a set of centos 7.1 hosts, trying to spin up a byo-origin cluster.
Ansible version: 1.9.4
Python: 2.7.5

    TASK: [openshift_persistent_volumes | Create PersistentVolumes] ***************
    failed: [mdeboer-master-01] => {"changed": true, "cmd": ["/usr/local/bin/oc", "create", "-f", "/tmp/openshift-ansible-f6qpyek/persistent-volumes.yml", "--config=/tmp/openshift-ansible-f6qpyek/admin.kubeconfig"], "delta": "0:00:05.410378", "end": "2016-02-18 17:39:25.999555", "failed": true, "failed_when_result": true, "rc": 0, "start": "2016-02-18 17:39:20.589177", "stdout_lines": ["persistentvolume \"registry-volume\" created"], "warnings": []}
    stderr:
    ================================================================================
    ATTENTION: You are running oc via a wrapper around 'docker run openshift/origin'.
    This wrapper is intended only to be used to bootstrap an environment. Please
    install client tools on another host once you have granted cluster-admin
    privileges to a user.
    See https://docs.openshift.org/latest/cli_reference/get_started_cli.html
    stdout: persistentvolume "registry-volume" created

As you can see above, the rc=0 and stdout lists the volume as created. I believe there is an error in ansible's understanding of the condition:
    `failed_when: ('already exists' not in pv_create_output.stderr if pv_create_output.stderr else False) or ('created' not in pv_create_output.stdout if pv_create_output.stdout else False)`
Although it seems to make logical sense to me, ansible interprets the above output as having failed.

Changing the condition to this simpler construct seems to resolve the issue for me: `failed_when: ('already exists' not in pv_create_output.stderr) and ('created' not in pv_create_output.stdout)`.

abutcher:
@matt-deboer PTAL. I remember working around stderr or stdout not being present for some reason so I'm adding empty string defaults to your fix.

Resolves #1441

### DIFF
--- a/roles/openshift_persistent_volumes/tasks/main.yml
+++ b/roles/openshift_persistent_volumes/tasks/main.yml
@@ -23,7 +23,7 @@
     --config={{ mktemp.stdout }}/admin.kubeconfig
   register: pv_create_output
   when: persistent_volumes | length > 0
-  failed_when: ('already exists' not in pv_create_output.stderr if pv_create_output.stderr else False) or ('created' not in pv_create_output.stdout if pv_create_output.stdout else False)
+  failed_when: ('already exists' not in pv_create_output.stderr) and ('created' not in pv_create_output.stdout)
   changed_when: ('created' in pv_create_output.stdout)
 
 - name: Deploy PersistentVolumeClaim definitions
@@ -40,7 +40,7 @@
     --config={{ mktemp.stdout }}/admin.kubeconfig
   register: pvc_create_output
   when: persistent_volume_claims | length > 0
-  failed_when: ('already exists' not in pvc_create_output.stderr if pvc_create_output.stderr else False) or ('created' not in pvc_create_output.stdout if pvc_create_output.stdout else False)
+  failed_when: ('already exists' not in pvc_create_output.stderr) and ('created' not in pvc_create_output.stdout)
   changed_when: ('created' in pvc_create_output.stdout)
 
 - name: Delete temp directory


matt-deboer:
LGTM. Thanks!

abutcher:
@sdodson ptal

roles/openshift_persistent_volumes/tasks/main.yml
```diff
@@ -23,7 +23,7 @@
     --config={{ mktemp.stdout }}/admin.kubeconfig
   register: pv_create_output
   when: persistent_volumes | length > 0
-  failed_when: ('already exists' not in pv_create_output.stderr if pv_create_output.stderr else False) or ('created' not in pv_create_output.stdout if pv_create_output.stdout else False)
+  failed_when: ('already exists' not in (pv_create_output.stderr | default(''))) and ('created' not in (pv_create_output.stdout | default('')))
```
sdodson: This seems like it'd yield false failures, I wonder if the wrapper script broke stdout / stderr splitting.

detiber: The default filter here should be relatively safe since it is only setting the default if stderr/stdout doesn't exist...  if it was using `| default('', true)`, then I would worry about false failures.

That said, I think stdout should always exist from the command module.

abutcher: @sdodson Yeah the problem is that I was expecting `stderr` to be empty when we were successful.

sdodson: After discussing offline, looks good to me.

abutcher: I'm good with removing the defaults aaand they're gone.

abutcher:
@brenton ^^ gtg review in the diff comments

PR #1445, opened 2016-02-19
Document parallel image pulls suggestion for Docker 1.9+.

aveshagarwal:


README_OSE.md
```diff
@@ -222,3 +223,6 @@ To override the the defaults, you can set the variables in your inventory:
 ose3-master.example.com openshift_ip=1.1.1.1 openshift_hostname=ose3-master.example.com openshift_public_ip=2.2.2.2 openshift_public_hostname=ose3-master.public.example.com
 ...snip...
 ```
+
+## Parallel image pulls with Docker 1.9+
+If you are using Docker 1.9+, you may consider to disable serial image pulls (or to enable parallel image pulls).
```
ncdc: I wouldn't say both disable serial and enable parallel. Just say enable parallel.

Please say there is an issue with data corruption with parallel pulls using Docker before 1.9. And that starting with 1.9, the corruption issue has been resolved, so it's safe to switch to parallel, but that the default if you do nothing is still serial.

ncdc:
cc @smarterclayton @brenton @sdodson @pweil-

detiber:
@ncdc @aveshagarwal: Is this something that we want to enable by default if docker 1.9+ is detected? And just document what the defaults are and how to override?

ncdc:
@detiber we will enable by default once Docker 1.9 is widely available. For now we're doing docs-only. See comments in https://trello.com/c/nOLPqFI6/268-2-enable-parallel-image-pulls-in-ose-3-2-when-running-on-docker-1-9-ops-rfe and https://github.com/openshift/origin/pull/6538.

aveshagarwal:
@detiber No, the default will be same as it is serial pulls for now. Just documenting it is safe to switch to parallel pulls if that want when using docker 1.9.

aveshagarwal:
@detiber Regarding override, probably we could have something in openshift-docs in master-node-configuration explaining how to override. And perhaps a link to that here.

detiber:
@ncdc I meant enabling it by default in openshift-ansible, since we can easily detect the version of docker installed?

ncdc:
> I meant enabling it by default in openshift-ansible, since we can easily detect the version of docker installed?

SGTM

ncdc:
@aveshagarwal please update this PR to include how to disable parallel (config sample)

detiber:
@ncdc that would best be done in the followup PR that provides a variable for setting it.

detiber:
created https://github.com/openshift/openshift-ansible/issues/1448 to track the feature addition.

ncdc:
:+1:

aveshagarwal:
@detiber  
As you suggested which is having a new variable in openshift-ansible, defaulting to serial pulls, and overriding that as needed. Right? And also I understand that it would be global not per node, right?

So does that mean there would be a default stanza in each node's configuration file? And once that default stanza is there, that could be modified per node too as needed?

ncdc:
@aveshagarwal don't worry about the change to openshift-ansible; we'll do that separately. Just doc what it would look like in node-config.yaml for kubeletArguments.

aveshagarwal:
@ncdc I dont think so this is a right place to document kubeletArguments. That seems more appropriate for openshift-docs to me.

detiber:
@aveshagarwal I'll work on a PR to implement this shortly, will tag you in the PR once it is ready.

detiber:
@aveshagarwal agreed that the kubeletArguments should be documented within openshift-docs and not here.

aveshagarwal:
@ncdc, thats why I am asking if there would be a default stanza there or not, so that it would be easier to explain as it is currently done in https://github.com/openshift/openshift-docs/blob/master/install_config/master_node_configuration.adoc

ncdc:
@aveshagarwal my bad, I meant to ask you to send the docs change to openshift-docs.

ncdc:
@aveshagarwal and no, I don't think kubeletArguments is there by default.

aveshagarwal:
@ncdc I know its not there now. But after this dicsussion as @detiber told he is going to create a new variable, and that sounded like there could be a default kubeletArguments stanza. And then it would be in line how things are explained there currently for node-confg.

aveshagarwal:
@ncdc but that does not mean, it can not be explained now in node-config.yaml.

detiber:
We actually already support setting kubeletArguments (through the openshift_node_kubelet_args variable), but I will be adding a separate variable to avoid having to use a json encoded string just to set parallel image pulls

detiber:
@sdodson, looks like this got lost in the backlog a while back, not sure if it is still relevant.

sdodson:
No longer relevant.

PR #1447, opened 2016-02-19
Add byo examples for network cidr and api/console ports.

### LINKED ISSUE #1446: hosts.example for byo doesn't document cluster_network_cidr or api_port or console_port

abutcher:
Resolves #1446

### DIFF
--- a/inventory/byo/hosts.aep.example
+++ b/inventory/byo/hosts.aep.example
@@ -163,6 +163,10 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # Can not be changed after deployment.
 #osm_cluster_network_cidr=10.1.0.0/16
 
+# Configure number of bits to allocate to each host’s subnet e.g. 8
+# would mean a /24 network on the host.
+#osm_host_subnet_length=8
+
 # Configure master API and console ports.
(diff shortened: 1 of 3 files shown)


inventory/byo/hosts.aep.example
```diff
@@ -157,6 +157,16 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
...
 
+# Configure SDN cluster network CIDR block. This network block should
+# be a private block and should not conflict with existing network
+# blocks in your infrastructure that pods may require access to.
+# Can not be changed after deployment.
+#osm_cluster_network_cidr=10.1.0.0/16
+
+# Configure master API and console ports.
+#openshift_master_api_port=8443
+#openshift_master_console_port=8443
```
detiber: Not directly related to this, but should we eliminate the separate variables for api_port and console_port, either that or ensure that if api_port is set and console_port is not set that console_port defaults to api_port?

cc @smarterclayton

liggitt: latter... console is separable from api... probably don't want to lose that

abutcher:
@brenton

abutcher:
@brenton should be gtg

@niqhil added an example for `osm_host_subnet_length`

PR #1449, opened 2016-02-19
Fix 1.2 version check

detiber:


sdodson:
LGTM, I'd also be fine with calling 1.1.3 as 1.2. It's closer to 1.2 than it is 1.1.

detiber:
@sdodson I'm hesitant in calling it 1.2 just because we wouldn't be able to handle any changes between 1.1.3 and 1.2 (and the obvious naming disconnect)

sdodson:
:+1:

detiber:
@brenton

PR #1450, opened 2016-02-19
a-o-i: Redo logic for detecting master_lb

smunilla:
After much research and debugging, it turns out the facts
for nativeha hosts were much less consistent than I hoped.
With this newfound knowledge, I've implemented this method.

In short, you check with the first master to see what it thinks
the master_lb is, then you include that host as "installed".


smunilla:
@abutcher @brenton PTAL

abutcher:
:+1:

smunilla:
aos-ci-tests

smunilla:
aos-ci-test

abutcher:
@brenton should be good to merge but parent CI job never marked complete

PR #1465, opened 2016-02-22
Add ansible.cfg to .gitignore

detiber:


detiber:
@eparis relevant to your interests.

@brenton ptal

eparis:
looks good to me. That plus $3 can get you a cup of coffee.....

PR #1469, opened 2016-02-23
Add missing is_atomic condition in the upgrade playbook

talset:
Update playbooks/common/openshift-cluster/upgrades/v3_1_minor/upgrade.yml

Add when not is_atomic on yum upgrade on master and node to be able to run the playbook on atomic host


sdodson:
LGTM, @brenton has been working on containerized upgrades lately so his work may intersect some here.

brenton:
@talset, I think we want to change it to `openshift.common.is_containerized`.

talset:
@brenton ok thanks, I correctly changed is_atomic in is_containerized

PR #1475, opened 2016-02-23
Fix openshift_repos role to support paths with spaces

davidmat:
The copy tasks in the openshift_repos fail when it is run from a path containing spaces. Surrounding the `{{item}}` argument with quotes fixes the jinja2 expansion

Example: Running `bin/cluster aws create 'clustername' ` in `/Users/dmat/Google Drive/openshift-ansible`.

_Before_

The tasks fail because the jinja template expands

````
src={{ item }} dest=/etc/pki/rpm-gpg/
````

to 

```
src=/Users/dmat/Google Drive/openshift-ansible  dest=/etc/pki/rpm-gpg/
```

which means ansible goes looking in /Users/dmat/Google and doesn't find anything.

_After_

```
src="/Users/dmat/Google Drive/openshift-ansible"  dest=/etc/pki/rpm-gpg/
```

and ansible looks in the correct location

### DIFF
--- a/roles/openshift_repos/tasks/main.yaml
+++ b/roles/openshift_repos/tasks/main.yaml
@@ -46,19 +46,19 @@
   with_fileglob:
   - '*/repos/*'
   when: not (item | search("/files/fedora-" ~ openshift_deployment_type ~ "/repos")) and
-        (ansible_distribution == "Fedora") 
+        (ansible_distribution == "Fedora")
         and not openshift.common.is_containerized | bool
   notify: refresh cache
 
 - name: Configure gpg keys if needed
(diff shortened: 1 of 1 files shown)


detiber:
ok to test

detiber:
aos-ci-test

abutcher:
aos-ci-test

detiber:
@abutcher I suspect the 3.1 failures are related to the channel issues we are currently having.

abutcher:
@detiber AH quite right

detiber:
aos-ci-test

detiber:
:+1: @brenton

PR #918, opened 2015-11-17
Fix ec2 instance type override

detiber:

PR #1476, opened 2016-02-23
Correct parsing of ec2_security_groups env variable

davidmat:
Similar to pull request #918 which fixed the parsing og the ec2_instance_type env. variable.

The `README_AWS.md` specifies that `export ec2_security_groups="['public']"` can be used to declare the security groups the instances will belong to. However, this is currently disregarded in the implementation and instead the default deployment_vars are always used.

This fix aligns the population of the `ec2_security_groups` with the  `ec2_instance_type` environment variable.


abutcher:
ok to test

detiber:
lgtm @brenton

PR #1478, opened 2016-02-23
Increase timeout on Wait for Node Registration

davidmat:
The "Wait for Node Registration task" in the openshift_manage_node task fails often when deploying a cluster to AWS. (using the `bin/cluster` script)

When re-applying the role a few minutes later, there are never issues.

Hence, I would propose to increase the timeout. The setting below works reliably for me.

### DIFF
--- a/roles/openshift_manage_node/tasks/main.yml
+++ b/roles/openshift_manage_node/tasks/main.yml
@@ -4,7 +4,7 @@
   register: omd_get_node
   until: omd_get_node.rc == 0
   retries: 20
-  delay: 5
+  delay: 30
   changed_when: false
   with_items: openshift_nodes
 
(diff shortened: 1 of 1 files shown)


abutcher:
ok to test

abutcher:
aos-ci-test

abutcher:
aos-ci-test

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -4,7 +4,7 @@
   register: omd_get_node
   until: omd_get_node.rc == 0
   retries: 20
-  delay: 5
+  delay: 30
```
detiber: I think I would prefer to see the number of retries increased rather than increasing the delay here.

davidmat:
Ok, I've done some more tests and it works for me keeping the delay on 5 secs, and instead increasing the retries (from 20 to 50), following @detiber's suggestion

detiber:
aos-ci-test

detiber:
:+1: from me. @brenton

PR #1479, opened 2016-02-23
lib_dyn: updates to lib_dyn role

mwoodson:


kwoodson:
:+1:

roles/lib_dyn/library/dyn_record.py
```diff
@@ -187,7 +190,7 @@ def main():
             record_type=dict(required=False, type='str', choices=[
                 'A', 'AAAA', 'CNAME', 'PTR', 'TXT']),
             record_value=dict(required=False, type='str'),
-            record_ttl=dict(required=False, default=0, type='int'),
+            record_ttl=dict(required=False, default=3600, type='int'),
```
rharrison10: We want the default to remain 0 here because the API will use the default ttl configured for the zone in this case. What we need to do is to get the zone default when ttl=0 and compare the record ttl to that as well.

roles/lib_dyn/library/dyn_record.py
```diff
@@ -233,27 +258,48 @@ def main():
...
         # Check to see if the record is already in place before doing anything.
-        if (dyn_node_records and
-                dyn_node_records[value_key][0].ttl == module.params['record_ttl'] and
-                (param_value in values[value_key] or
-                 param_value + '.' in values[value_key])):
-
-            module.exit_json(changed=False)
+        if dyn_node_records and \
+           ((param_value in dyn_values.keys()) or param_value + '.' in dyn_values.keys()):
+           #((param_value in dyn_values[record_type_key]) or param_value + '.' in dyn_values[record_type_key]):
```
rharrison10: Probably remove the commented line here.

roles/lib_dyn/library/dyn_record.py
```diff
@@ -233,27 +258,48 @@ def main():
...
+    elif module.params['state'] == 'present':
+
+        # configure the TTL variable:
+        # if use_zone_ttl, use the default TTL of the account.
+        # if TTL == None, don't check it, set it as 0 (api default)
+        # if TTL > 0, ensure this TTL is set
+        if module.params['use_zone_ttl']: #or isinstance(module.params['record_ttl'], str):
+            user_param_ttl = dyn_zone.ttl
+        elif not module.params['record_ttl']:
+            user_param_ttl = 0
```
rharrison10: This will cause the same always updating case that you're trying to work around. The value of 0 means set the value of the node ttl to be the same as the default for the zone. Its much simpler to leave the default as 0 and simply compare the `dyn_zone.ttl` to the node's ttl when `module.params['record_ttl'] == 0`. The `module.params['use_zone_ttl']` isn't necessary.

kwoodson:
:+1:

rharrison10:
Looking much better. Great improvements. :+1:

PR #1481, opened 2016-02-23
Better diagnostic messages when an OpenStack heat stack creation fails

lhuard1A:
On OpenStack, the resource creation can fail, for example because of quotas.
When it happens, it is cleaner to delete the heat stack in order to not leak unusable IaaS resources.
The inconvenient of that is that it lets the user without the opportunity to check what went wrong.

With that change, if OpenStack fails to create all the resources, openshift-ansible will display the status of the stack explaining the reason of the failure before deleting the stack.


detiber:
lgtm @brenton

PR #1482, opened 2016-02-23
[WIP] openshift_master_certificates: Force symlink

smunilla:
Force the creation of cert symlinks

### DIFF
--- a/roles/openshift_master_certificates/tasks/main.yml
+++ b/roles/openshift_master_certificates/tasks/main.yml
@@ -10,6 +10,7 @@
     src: "{{ openshift_master_config_dir }}/{{ item.1 }}"
     dest: "{{ openshift_generated_configs_dir }}/{{ item.0.master_cert_subdir }}/{{ item.1 }}"
     state: hard
+    force: true
   with_nested:
   - masters_needing_certs
   -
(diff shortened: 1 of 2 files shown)


smunilla:
@abutcher Is there a reason not to do this?

abutcher:
@smunilla Which bug is this for? I just reduced the list of certs we're hard linking to 3 in the master scaleup PR so it may not even be a target

roles/openshift_master_certificates/tasks/main.yml
```diff
@@ -10,6 +10,7 @@
     src: "{{ openshift_master_config_dir }}/{{ item.1 }}"
     dest: "{{ openshift_generated_configs_dir }}/{{ item.0.master_cert_subdir }}/{{ item.1 }}"
     state: hard
+    force: true
```
detiber: Technically, this is forcing a hardlink, not a symlink :grinning: 

That said, I don't see any issue with this. I'm assuming the error crops up when they are attempting to install after having generated_configs_dir already populated with some previous certs.

abutcher: Yeah agree with @detiber but I am very curious why `masters_needing_certs` contains these masters in the first place.

abutcher:
Twas in the branch name https://bugzilla.redhat.com/show_bug.cgi?id=1303859

abutcher:
So ansible is now attempting to create certificates for this master every run and getting no where. I think we need to track down why this is but we can hard link these anyway...

smunilla:
Interesting. So I'll try and track down a better root cause then.

PR #1483, opened 2016-02-23
First past at the upgrade process for 3.2

brenton:


playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
```diff
@@ -0,0 +1,147 @@
...
+    when: node_update_failed | length > 0
+
+###############################################################################
+# Reconcile Cluster Roles, Cluster Role Bindings and Security Context Constraints
+###############################################################################
+- name: Reconcile Cluster Roles and Cluster Role Bindings and Security Context Constraints
+  hosts: oo_masters_to_config
+  roles:
+  - openshift_facts
+  - openshift_cli
```
brenton: @detiber, The cli role needs to be run twice.  The first time it's running correctly and setting the right image tag image used inside the oc wrapper.  The second time we have to run it again after we've upgraded the images so that the various oc and oadm commands run with the latest version.  However, it keeps getting the old version for openshift.common.image_tag.  If I run the openshift_facts.yml playbook by itself it shows the correct output.  I tried explicitly putting the openshift_facts role in this playbook and nothing helped.

detiber: Adding the openshift_facts role would not make any difference here, since openshift_cli has a dependency on openshift_facts through openshift_common. I'll take a closer look at the change set to see if I can figure out what is going on.

playbooks/common/openshift-cluster/upgrades/files/rpm_versions.sh
```diff
@@ -0,0 +1,24 @@
...
+      containerized="TRUE"
+      ;;
+    \?)
+      echo "Invalid option: -$OPTARG" >&2
+      ;;
+  esac
+done
+
+if [ "${containerized}" == "TRUE" ] ; then
+  docker exec atomic-openshift-master rpm -q atomic-openshift
```
detiber: it looks like the `rpm -q` should be using `$@` instead, and this will probably need to be refactored to handle a parameter for the image name to use.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/containerized_upgrade.yml
```diff
@@ -0,0 +1,6 @@
+- name: Update system_units
+  script: ../files/ensure_system_units_have_version.sh {{ openshift.common.service_type }} {{ openshift.common.deployment_type }} {{ g_new_version }}
+
+- name: Ensure python-yaml present for config upgrade
+  action: "{{ ansible_pkg_mgr }} name=PyYAML state=present"
+  when: not openshift.common.is_atomic | bool
```
detiber: This shouldn't be needed here... PyYAML would only be needed for running the config update module.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
```diff
@@ -0,0 +1,220 @@
...
+###############################################################################
+- name: Load openshift_facts
+  hosts: oo_masters_to_config:oo_nodes_to_config:oo_etcd_to_config:oo_lb_to_config
+  roles:
+  - openshift_facts
+
+- name: Load openshift_facts
+  hosts: oo_masters_to_config:oo_nodes_to_config:oo_etcd_to_config:oo_lb_to_config
+  roles:
+  - openshift_facts
```
detiber: Any reason to invoke openshift_facts twice here?

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
```diff
@@ -0,0 +1,220 @@
...
+  vars:
+    openshift_master_ha: "{{ groups.oo_masters_to_config | length > 1 }}"
+    target_version: "{{ '1.2' if deployment_type == 'origin' else '3.1.1.900' }}"
+  gather_facts: no
+  tasks:
+  - fail:
+      msg: >
+        This upgrade is only supported for origin, openshift-enterprise, and online
+        deployment types
+    when: deployment_type not in ['origin','openshift-enterprise', 'online']
```
detiber: Do we want to exclude atomic-enterprise here? There shouldn't be any reason I can think of that upgrading from AEP 3.1 -> AEP/AP 3.2 wouldn't work.

playbooks/common/openshift-cluster/upgrades/files/rpm_versions.sh
brenton: I've actually been meaning to rip this out.  I split the containerized logic into a separate script.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/containerized_upgrade.yml
brenton: Yeah, another thing I've been meaning to rip out.

roles/cockpit/tasks/main.yml
```diff
@@ -6,10 +6,11 @@
     - cockpit-shell
     - cockpit-bridge
     - "{{ cockpit_plugins }}"
-  when: not openshift.common.is_atomic | bool
+  when: not openshift.common.is_containerized | bool
```
detiber: Any reason to switch to is_containerized here? The is_atomic check was put in specifically because we couldn't install the cockpit-kubernetes package on atomic.

roles/cockpit/tasks/main.yml
```diff
@@ -6,10 +6,11 @@
...
 
 - name: Enable cockpit-ws
   service:
     name: cockpit.socket
     enabled: true
     state: started
+  when: not openshift.common.is_containerized | bool
```
detiber: Have the requirements here changed? Previously it was requested that we explicitly start cockpit.

roles/openshift_node/tasks/main.yml
```diff
@@ -72,7 +79,7 @@
 - name: Reload systemd units
   command: systemctl daemon-reload
   when: openshift.common.is_containerized and ( ( install_node_result  | changed )
-    or ( install_ovs_result | changed ) )
+    or ( install_ovs_result | changed ) or ( install_ovs_sysconfig | changed ) )
```
detiber: daemon-reload shouldn't need to be invoked when the sysconfig file is changed, but you will want to trigger a handler  to restart the openvswitch service

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
detiber: Call me crazy, but I don't see anywhere that the master service is restarted after updating the packages and/or the unit files for the versions.

That said, it looks like the openvswitch restart might trigger master/node restarts for containerized installations though. I'm not sure if that has any impact on it or not.

It also looks like we are referencing the image by name, I'm wondering if there is an issue where the name isn't being mapped properly back to the versioned container in all cases.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
brenton: Definitely a poor copy/paste issue.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
brenton: Good catch!

roles/cockpit/tasks/main.yml
brenton: Scott and I discussed keeping containerized installs consistent.  If you perform a containerized OSE install you should get a containerized cockpit.  However, that container doesn't exist yet.  I probably won't ship this change with the upgrade I think it was just something I add because the correct cockpit RPM couldn't even be found and it was blocking my testing.

roles/cockpit/tasks/main.yml
brenton: I just don't want to do anything with cockpit for containerized installs until there actually is a containerized cockpit.

roles/openshift_node/tasks/main.yml
brenton: makes sense.  I didn't look closely enough at what this task was doing. :(

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
brenton: openvswitch restarting does trigger restarting the Master.  That's actually somewhat of a problem because it conflicts with us trying to perform a rolling restart of the masters here: https://github.com/brenton/openshift-ansible/blob/32upgrade/playbooks/byo/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml#L18

brenton:
aos-ci-test

brenton:
aos-ci-test

brenton:
aos-ci-test

brenton:
@detiber, are you generally OK with the state of this PR?  I just added the 1.2 templates so the diff is going to be a little more painful to look at.  I added a new fact to make it easier to select the correct templates to use.  I couldn't figure out how to do a case statement in ansible so this seemed like the cleanest approach.

brenton:
aos-ci-test

brenton:
btw, I don't think the 3.0 install failure was related.  All the other tests passed the last time they ran.

roles/openshift_examples/examples-sync.sh
```diff
@@ -5,8 +5,8 @@
 #
 # This script should be run from openshift-ansible/roles/openshift_examples
 
-XPAAS_VERSION=ose-v1.2.0-1
-ORIGIN_VERSION=v1.1
+XPAAS_VERSION=master
```
detiber: Has master stabilized enough for us to use it?

playbooks/byo/openshift-cluster/upgrades/v3_1_to_v3_2/README.md
```diff
@@ -0,0 +1,17 @@
+# v3.1 to v3.2 upgrade playbook
+
+## Overview
+This playbook currently performs the
+following steps.
+
+**TODO: update for current steps**
```
detiber: heh, wonder how many of the readmes look the same.

roles/openshift_examples/examples-sync.sh
brenton: I didn't see another branch.  In any case, I would expect the latest dev version to pull from master and then when it stabilizes we start downloading to a new v1.3 directory (from master).

playbooks/common/openshift-cluster/upgrades/files/ensure_system_units_have_version.sh
```diff
@@ -0,0 +1,51 @@
...
+    sysconfig_file=/etc/sysconfig/${unit_name}
+
+    if ! grep IMAGE_VERSION ${sysconfig_file}; then
+        sed -i "/CONFIG_FILE/a IMAGE_VERSION=${1}" ${sysconfig_file}
+    else
+        sed -i "s/\(IMAGE_VERSION=\).*/\1${1}/" ${sysconfig_file}
+    fi
+}
+
+add_image_version_to_unit () {
```
detiber: for this should we just lay down the new template? I wouldn't trust us clobbering the sysconfig file, but the unit file should be considered under our ownership, no different than the ones we ship in the RPMs.

playbooks/byo/openshift-cluster/upgrades/v3_1_to_v3_2/README.md
brenton: I'll remove this from this README and the 3.0->3.1 one.

playbooks/common/openshift-cluster/upgrades/files/ensure_system_units_have_version.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+set -e
+
+SERVICE_TYPE=$1
+DEPLOYMENT_TYPE=$2
+VERSION="v${3}"
+
+add_image_version_to_sysconfig () {
```
detiber: the lineinfile ansible module can do this (add a line only if the regex does not match)

playbooks/common/openshift-cluster/upgrades/files/openshift_container_versions.sh
```diff
@@ -0,0 +1,22 @@
...
+elif grep ose $unit_file > /dev/null; then
+    image_name="openshift3/ose"
+fi
+
+docker pull ${image_name} 1>&2
+available=$(docker run --rm ${image_name} version | grep openshift | awk '{ print $2 }' | cut -f1 -d"-" | tr -d 'v')
+
+echo "---"
+echo "curr_version: ${installed}"
+echo "avail_version: ${available}"
```
detiber: I'm starting to wonder if this just belongs in OpenShift_facts...

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/containerized_upgrade.yml
```diff
@@ -0,0 +1,9 @@
+- name: Update system_units
+  script: ../files/ensure_system_units_have_version.sh {{ openshift.common.service_type }} {{ openshift.common.deployment_type }} {{ g_new_version }}
+
+- name: Verifying the correct version was configured
+  shell: grep {{ verify_upgrade_version }} {{ item }}
```
detiber: the command module should work here.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
```diff
@@ -0,0 +1,235 @@
...
+  - set_fact:
+      g_new_version: "{{ openshift_pkg_version | replace('-','') }}"
+    when: openshift_pkg_version is defined
+
+  - name: Update systemd units
+    script: ../files/ensure_system_units_have_version.sh {{ openshift.common.service_type }} {{ openshift.common.deployment_type }} {{ g_aos_versions.curr_version }}
+    when: openshift.common.is_containerized | bool
+
+  # Note: the version number is hardcoded here in hopes of catching potential
+  # bugs in how g_aos_versions.curr_version is set
```
detiber: leftover debug statement?

playbooks/common/openshift-cluster/upgrades/files/ensure_system_units_have_version.sh
brenton: I really struggled with what appeared to be cached values returned by openshift_facts when upgrades were in progress.  I'll have to see if I can come up with a reproducer outside of the upgrade.  Part of me just wants to get this to QE officially as soon as possible so they can start finding serious bugs.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/rpm_upgrade.yml
```diff
@@ -0,0 +1,6 @@
+- name: Upgrade packages
+  command: "{{ ansible_pkg_mgr}} update -y {{ openshift.common.service_type }}-{{ component }}{{ openshift_version }}"
+
+- name: Ensure python-yaml present for config upgrade
```
detiber: are we doing any config upgrade this time?

playbooks/common/openshift-cluster/upgrades/files/ensure_system_units_have_version.sh
detiber: :+1:

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
brenton: I'm actually planning to leave this in here.  This is going to be _really_ hard for QE to test so when they run their upgrades they will set `verify_upgrade_version=<whatever version they think they are installing>` and the extra checks will be run.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/rpm_upgrade.yml
brenton: I haven't completely decided yet.  We could technically add in the oc client download links but I don't view that as a blocker right now all things considered.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/containerized_upgrade.yml
brenton: I'll update this.  Is the command module just faster since isn't running  a shell?

roles/openshift_examples/examples-sync.sh
sdodson: The xpaas team strongly prefers we never allow anything other than a tagged version of their content into the installer. I think we should leave that particular one as is.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/containerized_upgrade.yml
detiber: I'm not sure about faster, but it is safer, since there is no chance of running extraneous commands through the shell.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
detiber: In that case should the version to check against be a variable as well?

detiber: alternatively, we could just output the value of g_aos_versions.curr_version in a debug statement that can be validated.

brenton:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
brenton: That just seemed a little too manual to me.  I'll think through this a bit tomorrow to see if there is something better I could do here.

roles/openshift_examples/examples-sync.sh
brenton: That's fine.  I guess we need to tell them to create the tag for the 1.2 release then.  I didn't see it when I looked.

arilivigni:
aos-ci-test

brenton:
aos-ci-test

brenton:
aos-ci-test

arilivigni:
aos-ci-test

PR #1485, opened 2016-02-24
a-o-i: Better method for identifying master_lb

smunilla:
The previous check was less than reliable thanks to
potential differences between hostname / ip / connect_to.
This check should cover all instances.


abutcher:
:+1:

smunilla:
@brenton The latest in my attempts to fix this persistent bug

Issue #1026, opened 2015-12-04
Style guide - role variable names


detiber:
Browsing galaxy.ansible.com, I came across the following: https://galaxy.ansible.com/intro#variables

The suggestion is that all role variables (including vars/main.yml) should be prefixed with the role name. This contradicts our style guide suggestion to use shortened prefixes.

As much as the long descriptive naming is a pain to work with, I think sticking with the galaxy best practices and more descriptive names for end users is worth the hassle.

detiber:
@twiest

twiest:
@detiber hmm, yeah, this is something we should definitely consider. I always prefer to use the conventions of the community. I'll think about this further.

tbielawa:
This is probably covered under the **Future Plans** doc:

https://docs.google.com/document/d/1uqW2Ze0Hc_2D976k2WTN6TNvAuuOZFPyztQ3d4iGCgQ/edit

PR #1495, opened 2016-02-25
Rename variable to openshift_master_default_subdomain

jstuever:
Updated the openshift_master role to prefer the correctly named openshift_master_default_subdomain variable while maintaining backwards compatibility with osm_default_subdomain. The inventory examples now reference the correctly named variable.

### DIFF
--- a/inventory/byo/hosts.aep.example
+++ b/inventory/byo/hosts.aep.example
@@ -132,7 +132,7 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 #osm_api_server_args={'max-requests-inflight': ['400']}
 
 # default subdomain to use for exposed routes
-#osm_default_subdomain=apps.test.example.com
+#openshift_master_default_subdomain=apps.test.example.com
 
 # additional cors origins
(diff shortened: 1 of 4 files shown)


sdodson:
Was it decided that we were vacating the osm_ and osn_ prefixes for master and node specific variables? I personally prefer osm_ and osn_ over the longer forms.

abutcher:
@sdodson from this issue https://github.com/openshift/openshift-ansible/issues/1026

detiber:
@sdodson basically, if we plan on shipping the roles through galaxy eventually, then we need to use the more verbose naming by default. The verbose naming also will help as we move towards the automatic variable -> fact configuration that @abutcher started work on.

detiber:
ok to test

detiber:
aos-ci-test

detiber:
:+1: @brenton

brenton:
@smunilla I suspect you'll need to update atomic-openshift-installer to match.

detiber:
@brenton @smunilla the old variable is preserved, but yes, we should update the installer.

PR #1496, opened 2016-02-25
Update main.yml

niqhil:
Do not enforce nfs storage plugin. Let it be handled at the configuration storage_plugin_deps. 
This gives error for byo offline install. ( yum repo )


abutcher:
Looks good but I think `'nfs'` should be in the default `storage_plugin_deps` defined in `openshift_facts` but this would allow it to be overridden in inventory.

Are you installing on centos? Want to know where nfs-utils isn't available and maybe we can resolve that or document it better.

abutcher:
Oh, offline install...

detiber:
@niqhil +1 to this change, can you squash the commits?

detiber:
aos-ci-test

detiber:
@niqhil apologies about letting this PR drag on, if you can resolve the conflict, we can re-test and get this merged.

mtnbikenc:

PR #1501, opened 2016-02-25
[WIP] Delegation Refactor

abutcher:


roles/etcd/meta/main.yml
```diff
@@ -18,3 +18,4 @@ galaxy_info:
 dependencies:
 - { role: os_firewall }
 - { role: etcd_common }
+- { role: etcd_server_certificates }
```
detiber: Why break up the etcd_certificates role here? instead you can just call etcd_certificates with a variable for generating either a client cert or a server cert.

Also, now that I think about it with the delegate_to, we could probably also consolidate the etcd_ca and etcd_certificates role (I would suggest they be separate task files within the etcd role and included in main.yml). That consolidation may make it harder to generalize the CA and integrate it with an external cert provider eventually though.

roles/flannel/meta/main.yml
```diff
@@ -14,3 +14,4 @@ galaxy_info:
   - system
 dependencies:
 - { role: openshift_facts }
+- { role: etcd_client_certificates, etcd_generated_certs_dir: /etc/etcd/generated_certs, etcd_cert_prefix: node.etcd- }
```
detiber: Any reason to use the node.etcd- prefix here instead of a certificate specific to flannel here? It would make it easier to have etcd reference a separate etcd instance once we have etcd sharding supported.

detiber: That said, we should also probably make the destination directory for the certificates configurable.

playbooks/common/openshift-master/config.yml
```diff
@@ -1,26 +1,10 @@
 ---
-- name: Set master facts and determine if external etcd certs need to be generated
+- name: Set master facts
   hosts: oo_masters_to_config
   pre_tasks:
-  - name: Check for RPM generated config marker file .config_managed
```
detiber: I think we still need this task in case the openshift package was installed and run prior to the playbook being executed.

detiber:
Just looking at the number of lines removed from the playbooks puts a huge :smile:  on my face.

We are just an openshift_etcd and openshift_etcd_facts role away from just having an openshift-etd/config.yml playbook that just applies roles (and by extension easier for Online to consume).

roles/etcd/meta/main.yml
abutcher: I'm not opposed to joining them back together but I liked the idea of them being separate with this code since they didn't really share any code. If we consolidate CA into a single role then I think it makes sense to combine them again.

roles/flannel/meta/main.yml
abutcher: Nah, I copied this out right from where it was defined in the playbook- I will change to flannel. The destination directory is configurable but is set at the playbook level. I assumed these vars would be static.

playbooks/common/openshift-master/config.yml
abutcher: I thought I moved this but it looks like I just deleted it :smile:

roles/etcd/meta/main.yml
detiber: In that case, should we make the CA role more generic, say `openssl_ca` so that it can be used by an *_certificates role that uses an openssl certificate authority?

detiber: I think I like that idea better, so that other roles that would need/want an independent CA can piggyback off of it (such as the new nuage certificate bits.

abutcher: I like it

detiber: btw, I might need you to submit a PR to implement the certificate work for this pr: https://github.com/openshift/openshift-ansible/pull/1503

abutcher:
aos-ci-test

abutcher:
@detiber Getting closer, PTAL when you have a chance.

abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
```
TASK: [common | Disable all repositories] ************************************* 
changed: [10.8.168.208]
changed: [10.8.168.210]
failed: [10.8.168.206] => {"changed": true, "cmd": "subscription-manager repos --disable=\"*\"", "delta": "0:01:06.129579", "end": "2016-03-08 09:45:13.270110", "rc": 70, "start": "2016-03-08 09:44:07.140531", "warnings": []}
stderr: Unable to verify server's identity: timed out
```

abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
aos-ci-test

arilivigni:
aos-ci-test

arilivigni:
aos-ci-test

arilivigni:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
aos-ci-test

roles/etcd/meta/main.yml
```diff
@@ -17,4 +17,4 @@ galaxy_info:
   - system
 dependencies:
 - { role: os_firewall }
-- { role: etcd_common }
+- { role: etcd_server_certificates, etcd_cert_config_dir: /etc/etcd, etcd_cert_prefix: '', etcd_generated_certs_dir: /etc/etcd/generated_certs }
```
detiber: This can actually be:
```
- role: etcd_server_certificates
  etcd_cert_config_dir: /etc/etcd
  etcd_cert_prefix: ''
  etcd_generated_certs_dir: /etc/etcd/generated_certs
```

roles/etcd_ca/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+openshift_etcd_ca_host: "{{ groups.oo_first_etcd.0 }}"
```
detiber: I'm not sure we want to rely on specific group names at the role level, the roles should really be self-sufficient and testable without needing to expose playbook specific details.

Instead, I would say that we just require that openshfit_etcd_ca_host be set, and pass that in as a variable to the role.

roles/etcd_client_certificates/README.md
```diff
@@ -1,5 +1,5 @@
-OpenShift etcd certificates
-========================
+OpenShift etcd Client Certificates
+==================================
```
detiber: What do you think about consolidating etcd_ca, etcd_client_certificates, and etcd_server_certificates?

tasks/main.yaml could do conditional includes based on a variable describing which action to take.

roles/etcd_ca/tasks/main.yml
```diff
@@ -2,6 +2,8 @@
 - name: Install openssl
   action: "{{ ansible_pkg_mgr }} name=openssl state=present"
   when: not openshift.common.is_atomic | bool
+  delegate_to: "{{ openshift_etcd_ca_host }}"
```
detiber: If this is going to be a generic etcd role (and I think it can and should be), then we need to avoid using openshift_* for variables, and use an openshift_etcd_facts role to translate values from openshift_facts into etcd_* variables.

roles/etcd_client_certificates/tasks/create.yml
```diff
@@ -0,0 +1,94 @@
...
+    mode: 0700
+  delegate_to: "{{ openshift_etcd_ca_host }}"
+
+- name: Create the client csr
+  command: >
+    openssl req -new -keyout {{ etcd_cert_prefix }}client.key
+    -config {{ etcd_openssl_conf }}
+    -out {{ etcd_cert_prefix }}client.csr
+    -reqexts {{ etcd_req_ext }} -batch -nodes
+    -subj /CN={{ openshift.common.hostname }}
```
detiber: same comments here about the openshift.* variables.

roles/etcd_client_certificates/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+  when: groups.oo_etcd_to_config is defined and groups.oo_etcd_to_config
+
+- set_fact:
+    etcd_client_certs_missing: "{{ False in (g_external_etcd_cert_stat_result.results
+                                   | oo_collect(attribute='stat.exists')
+                                   | list) }}"
+  when: groups.oo_etcd_to_config is defined and groups.oo_etcd_to_config
+
+- include: create.yml
+  when: groups.oo_etcd_to_config is defined and groups.oo_etcd_to_config and etcd_client_certs_missing | bool
```
detiber: We should be passing in a variable containing the list of hosts instead of relying on groups.oo_etcd_to_config.

roles/flannel/defaults/main.yaml
```diff
@@ -3,6 +3,6 @@ flannel_interface: "{{ ansible_default_ipv4.interface }}"
 flannel_etcd_key: /openshift.com/network
 etcd_hosts: "{{ etcd_urls }}"
 etcd_conf_dir: "{{ openshift.common.config_base }}/node"
-etcd_peer_ca_file: "{{ etcd_conf_dir }}/{{ 'ca' if (embedded_etcd | bool) else 'node.etcd-ca' }}.crt"
-etcd_peer_cert_file: "{{ etcd_conf_dir }}/{{ 'system:node:' + openshift.common.hostname if (embedded_etcd | bool) else 'node.etcd-client' }}.crt"
-etcd_peer_key_file: "{{ etcd_conf_dir }}/{{ 'system:node:' + openshift.common.hostname if (embedded_etcd | bool) else 'node.etcd-client' }}.key"
+etcd_peer_ca_file: "{{ etcd_conf_dir }}/{{ 'ca' if (embedded_etcd | bool) else 'flannel.etcd-ca' }}.crt"
+etcd_peer_cert_file: "{{ etcd_conf_dir }}/{{ 'system:node:' + openshift.common.hostname if (embedded_etcd | bool) else 'flannel.etcd-client' }}.crt"
+etcd_peer_key_file: "{{ etcd_conf_dir }}/{{ 'system:node:' + openshift.common.hostname if (embedded_etcd | bool) else 'flannel.etcd-client' }}.key"
```
detiber: I don't consider this flannel role to be generic by any means, so we might want to rename it to openshift_flannel and not worry about requiring openshift_facts here.

roles/fluentd_master/meta/main.yml
```diff
@@ -11,5 +11,5 @@ galaxy_info:
     - 7
   categories:
   - monitoring
-  dependencies:
-  - openshift_facts
+dependencies:
+- { role: openshift_facts }
```
detiber: are the fluentd_* roles even in use anymore? part of me wants to say that we should just remove them...

@wshearn is Ops using the fluentd_* roles? If so, any objections to moving them over to openshift_tools?

roles/haproxy/meta/main.yml
```diff
@@ -10,5 +10,6 @@ galaxy_info:
     versions:
```
detiber: Should we rename this role to openshift_loadbalancer?

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -1,22 +1,38 @@
+# Necessary because when you're on a node that's also a master the master will be
+# restarted after the node restarts docker and it will take up to 60 seconds for
+# systemd to start the master again
+- name: Wait for master API to become available before proceeding
+  # Using curl here since the uri module requires python-httplib2 and
+  # wait_for port doesn't provide health information.
+  command: >
+    curl --silent --cacert {{ openshift.common.config_base }}/master/ca.crt
+    {{ openshift_node_master_api_url }}/healthz/ready
```
detiber: should this be the openshift_fact value here instead of openshift_node_master_api_url?

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -1,22 +1,38 @@
...
+  delegate_to: "{{ openshift_manage_node_master_host }}"
 
 - name: Label nodes
   command: >
-    {{ openshift.common.client_binary }} label --overwrite node {{ item.openshift.common.hostname | lower }} {{ item.openshift.node.labels | oo_combine_dict  }}
-  with_items:
-    -  "{{ openshift_node_vars }}"
-  when: "'labels' in item.openshift.node and item.openshift.node.labels != {}"
+    {{ openshift.common.client_binary }} label --overwrite node {{ openshift.common.hostname | lower }}
+    {{ openshift.node.labels | oo_combine_dict }}
```
detiber: I saw a change recently that allowed for setting the node labels in the node config itself, might be worth taking a look at that approach eventually (for versions that support it) instead of relying on openshift_manage_node to manage the labels...

roles/openshift_manage_node/vars/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+openshift_manage_node_master_host: "{{ groups.oo_first_master.0 }}"
```
detiber: Same comment here as with the etcd roles, we shouldn't be relying on specific group names in the roles.

roles/openshift_master_certificates/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
```
detiber: I'm wondering if we should consolidate the openshift_master_ca, openshift_master_certificates and openshift_node_certificates roles. With the delegate_to changes, it seems like it makes more sense to consolidate these roles instead of keeping them separate...

roles/openshift_master_facts/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+openshift_master_ca_host: "{{ groups.oo_first_master.0 }}"
```
detiber: another spot where we are leaking groups into roles.

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+- name: Check for RPM generated config marker file .config_managed
+  stat:
+    path: /etc/origin/.config_managed
+  register: rpmgenerated_config
+
+- name: Remove RPM generated config files if present
+  file:
+    path: "/etc/origin/{{ item }}"
+    state: absent
+  when: rpmgenerated_config.stat.exists == true and deployment_type in ['openshift-enterprise', 'atomic-enterprise']
```
detiber: I'm wondering if 'online' should be in this list as well, since it is using the same binary packaging and could potentially have the generated configs.

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+  - .config_managed
+
+- set_fact:
+    openshift_master_pod_eviction_timeout: "{{ lookup('oo_option', 'openshift_master_pod_eviction_timeout') | default(none, true) }}"
+  when: openshift_master_pod_eviction_timeout is not defined
+
+- set_fact:
+    openshift_master_etcd_port: "{{ (etcd_client_port | default('2379')) if (groups.oo_etcd_to_config is defined and groups.oo_etcd_to_config) else none }}"
+    openshift_master_etcd_hosts: "{{ hostvars
+                                     | oo_select_keys(groups['oo_etcd_to_config']
```
detiber: another group leakage, as much as I hate to push the logic back into the playbook, tying into group names that we are setting in the playbooks creates an artificial dependency on the playbooks.

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+    state: absent
+  when: rpmgenerated_config.stat.exists == true and deployment_type in ['openshift-enterprise', 'atomic-enterprise']
+  with_items:
+  - master
+  - node
+  - .config_managed
+
+- set_fact:
+    openshift_master_pod_eviction_timeout: "{{ lookup('oo_option', 'openshift_master_pod_eviction_timeout') | default(none, true) }}"
+  when: openshift_master_pod_eviction_timeout is not defined
```
detiber: instead of using set_fact here, you can just set this using the default filter in the openshift_facts task below.

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+    openshift_master_etcd_port: "{{ (etcd_client_port | default('2379')) if (groups.oo_etcd_to_config is defined and groups.oo_etcd_to_config) else none }}"
+    openshift_master_etcd_hosts: "{{ hostvars
+                                     | oo_select_keys(groups['oo_etcd_to_config']
+                                                      | default([]))
+                                     | oo_collect('openshift.common.hostname')
+                                     | default(none, true) }}"
+
+- set_fact:
+    openshift_master_debug_level: "{{ lookup('oo_option', 'openshift_master_debug_level') | default(openshift.common.debug_level, true) }}"
+  when: openshift_master_debug_level is not defined
```
detiber: probably want to move this down into the openshift_facts task as well to avoid doing this in a separate task.

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+        mcs_allocator_range: "{{ osm_mcs_allocator_range | default(None) }}"
+        mcs_labels_per_project: "{{ osm_mcs_labels_per_project | default(None) }}"
+        uid_allocator_range: "{{ osm_uid_allocator_range | default(None) }}"
+        router_selector: "{{ openshift_router_selector | default(None) }}"
+        registry_selector: "{{ openshift_registry_selector | default(None) }}"
+        api_server_args: "{{ osm_api_server_args | default(None) }}"
+        controller_args: "{{ osm_controller_args | default(None) }}"
+        infra_nodes: "{{ openshift_infra_nodes | default(None) }}"
+        disabled_features: "{{ osm_disabled_features | default(None) }}"
+        ha: "{{ openshift_master_ha | default(groups.oo_masters | length > 1) }}"
```
detiber: more group leakage

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+        master_count: "{{ openshift_master_count | default(groups.oo_masters | length) }}"
+        controller_lease_ttl: "{{ osm_controller_lease_ttl | default(None) }}"
+        controllers_port: "{{ openshift_master_controllers_port | default(None) }}"
+        master_image: "{{ osm_image | default(None) }}"
+
+- openshift_facts:
+    role: hosted
+    openshift_env:
+      openshift_hosted_registry_storage_kind: 'nfs'
+  when: openshift_hosted_registry_storage_kind is not defined and groups.oo_nfs_to_config is defined and groups.oo_nfs_to_config | length > 0
```
detiber: group leakage.

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+
+- set_fact:
+    openshift_infra_nodes: "{{ hostvars | oo_select_keys(groups['oo_nodes_to_config'])
```
detiber: group leakage

roles/openshift_node_facts/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
+---
+- set_fact:
+    openshift_node_debug_level: "{{ lookup('oo_option', 'openshift_node_debug_level') | default(openshift.common.debug_level, true) }}"
+  when: openshift_node_debug_level is not defined
```
detiber: this can be pushed down into the openshift_facts task.

roles/rhel_subscribe/meta/main.yml
```diff
@@ -1,2 +1,2 @@
 dependencies:
-  - openshift_facts
+- { role: openshift_facts }
```
detiber: not really related to this pr, other than it being in the changeset, but rhel_subscribe probably shouldn't depend on openshift_facts :)

roles/flannel/defaults/main.yaml
abutcher: Renamed openshift_flannel.

roles/haproxy/meta/main.yml
abutcher: Renamed to openshift_loadbalancer.

roles/openshift_manage_node/tasks/main.yml
abutcher: The fact is persisted for masters and relies on other master facts to build so this is passed in.

roles/openshift_master_certificates/defaults/main.yml
abutcher: I'm cool with this, the approach would mirror what I've done with the `etcd_certificates` role.

abutcher:
aos-ci-test

abutcher:
aos-ci-test

detiber:
@abutcher call me crazy, but I just had the thought that we should probably break up the openshift_facts module now that we are introducing openshift_<role>_facts roles.

Would that help out with the group leakage issues?

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1425,6 +1425,13 @@ def get_defaults(self, roles, deployment_type):
                                       auth_token_max_seconds=500,
                                       oauth_grant_method='auto')
 
+        if 'haproxy' in roles:
```
detiber: This role should probably be loadbalancer now with the openshift_haproxy -> openshift_loadbalancer rename.

roles/openshift_manage_node/tasks/main.yml
detiber: :+1:

playbooks/common/openshift-haproxy/config.yml
```diff
@@ -0,0 +1,5 @@
+---
```
detiber: This playbook should probably be renamed to openshift-loadbalancer/config.yml with the role renaming.

abutcher:
Heresy!! I like that idea but we'd have to think about facts that rely on
each other from different roles? I suppose each _whatever_facts could
depend on common facts.

I think I've got all of the leaks patched up sans nuage roles. I need to
comb through variable naming.

detiber:
> Heresy!! I like that idea but we'd have to think about facts that rely on
each other from different roles? I suppose each _whatever_facts could
depend on common facts.

Yeah, it definitely is something that is a huge refactor in and of itself. I suspect we'll need to be a bit more granular as well.

openshift_*_facts would have to depend on openshift_facts for all the generic facts
openshift_{master,node}_facts would have to depend on a openshift_<something>_facts that would bring in the binary for doing version testing (and removing that from openshift_common).

We'd also probably want to split the persisted files out as well, which would add yet another level of complication :)

roles/etcd_certificates/tasks/create_client_certificates.yml
```diff
@@ -0,0 +1,94 @@
...
+  environment:
+    SAN: "IP:{{ etcd_ip }}"
+  delegate_to: "{{ etcd_certificates_ca_host }}"
+
+- name: Sign and create the client crt
+  command: "{{ item.command }}"
+  args: "{{ item.args | default({}) }}"
+  environment: "{{ item.environment | default({}) }}"
+  with_items:
+  - command: "sleep {{ etcd_certificates_etcd_hosts | default([inventory_hostname]) | oo_index(inventory_hostname, 0) }}"
```
abutcher: Working on a solution to this hack :angel:

ashcrow:
Is this still likely to be worked on and merged in the near future?

abutcher:
@ashcrow I've begun chunking this up into smaller PRs, although I recently had to revert the openshift certificates refactor and it looks like we'll have to finish getting to ansible 2.0 support in order to wrap this up.

openshift-certificates https://github.com/openshift/openshift-ansible/pull/1908
openshift-certificates reverted https://github.com/openshift/openshift-ansible/pull/1962
etcd-certificates https://github.com/openshift/openshift-ansible/pull/1953
separate master/loadbalancer configs https://github.com/openshift/openshift-ansible/pull/1896

I'd say go ahead with whatever work this is blocking and we can take care of any subsequent refactors that need to happen.

PR #1503, opened 2016-02-26
Changes required for Nuage Monitor REST server

vishpat:
@detiber Changes for Nuage Openshift Monitor REST server configuration.


roles/nuage_node/vars/main.yaml
```diff
@@ -6,4 +6,10 @@ client_cert: "{{ vsp_openshift_dir }}/nuage.crt"
 client_key: "{{ vsp_openshift_dir }}/nuage.key"
 ca_cert: "{{ vsp_openshift_dir }}/ca.crt"
 api_server: "{{ openshift_node_master_api_url }}"
+nuage_mon_rest_server_port: "{{ nuage_openshift_monitor_rest_server_port | default('8444') }}"
```
detiber: 8444 is the same port that the controllers service listens on, since the nuage mon service and the controllers service will be hosted together on the master hosts, the nuage mon service should default to a different port.

roles/nuage_master/files/serviceaccount.sh
```diff
@@ -59,5 +66,31 @@ oadm create-api-client-config --certificate-authority=${CA_CERT} --client-dir=${
...
+openssl x509 -req -in ${OUTDIR}/restServer.req -CA $REST_CA_CERT -CAkey \
+    $REST_CA_KEY -CAserial $REST_CA_SERIAL \
+    -out $REST_SERVER_CERT
+
+openssl genrsa -out $REST_CLIENT_KEY 4096
+openssl req -key $REST_CLIENT_KEY -new -out ${OUTDIR}/restClient.req -subj "/CN=nuage-client"
+openssl x509 -req -in ${OUTDIR}/restClient.req -CA $REST_CA_CERT -CAkey \
+    $REST_CA_KEY -CAserial $REST_CA_SERIAL \
+    -out $REST_CLIENT_CERT -extensions clientauth \
+    -extfile ${OUTDIR}/openssl.cnf
```
detiber: This will create a CA on each of the nuage_master hosts, instead I believe you only want to create the CA on a single master and distribute the created certs to the other master and node hosts.

We are currently refactoring the way we are generating and distributing certificates. Currently the logic to generate and distribute them is done within the playbooks, but we are pushing that logic down into the roles. You can see the approach we are moving to here: https://github.com/openshift/openshift-ansible/pull/150

In general, you will most likely want the following:

- a nuage_ca role that creates the certificate authority to be used and uses delegate_to and run_once to ensure that the ca is created only on the delegated host (most likely you'll want groups.oo_first_master.0 to be the delegate_to host).
- a nuage_certificate role that creates the server and/or client certs based on which is needed and distributes them to the appropriate host. This role should use delegate_to to create the certificates on the CA host and local_action to facilitate transferring the certificates to the intended host. This role should also have a dependency defined on the nuage_ca role.
- modify the nuage_master and nuage_node roles to have a dependency on the nuage_certificate role, and to set the variable needed to tell it to create and distribute the server/client cert appropriately.

detiber:
@vishpat Does the nuage monitor rest server configuration replace any existing communication for the master/node clients, or is it just an additional communication stream?

I'm just wondering if there are any associated configuration items that can be removed with this change.

vishpat:
@detiber The Nuage monitor rest server is an additional communication stream. Your comments about the certificate creation makes sense, I will implement those changes.

roles/nuage_master/files/serviceaccount.sh
vishpat: For nuage_ca role, Instead of delegate_to can't I just use the host to generate the key and crt. For example.

- name: Create CA key
  host: {{ groups.oo_first_master.0 }}
  command: openssl genrsa -out {{ nuage_ca_key }} 4096
  run_once: true

detiber: If you don't use delegate_to, then each host would have it's own CA and wouldn't be able to validate certificates generated by one of the other CAs.

vishpat: @detiber I am able to generate the certificates for each node on the CA host but I am unable to copy them over. I have pushed my latest changes which exemplify this. 

<pre>
- name: Copy the CA certificate from master to node
  synchronize: src="{{ nuage_ca_crt }}" dest="{{ nuage_plugin_crt_dir }}/nuageMonCA.crt"
  delegate_to: "{{ nuage_ca_master }}" 
</pre>

Obviously this fails because the master node is not setup to do passwordless ssh into the node.
Any idea how do I copy over the certificates ?

detiber:
@abutcher can you review and make sure that the certificate generation aligns with the certificate refactoring you are doing?

abutcher:
@vishpat Where is the `nuage_master_certificate` role being applied or included?

vishpat:
Opening a new PR (https://github.com/openshift/openshift-ansible/pull/1528) for these changes.

@abutcher The changes you looked at were work in progress. I have posted a new PR for this work (since I messed up this branch). Please take a look at those changes.

PR #1504, opened 2016-02-26
libvirt: Upgrade to CentOS-7-x86_64-GenericCloud-1602.

### LINKED ISSUE #1384: [playbooks/libvirt] CentOS-7-x86_64-GenericCloud doesn't match sha256
[followup to #954]
`bin/cluster create libvirt foo` currently fails out of the box due to CentOS-7-x86_64-GenericCloud.qcow2.xz failing sha256sum check.

http://cloud.centos.org/centos/7/images/ dir is presently in a confusing state (which I don't know where to report):

- The current expected checksum, last updated in #954 (9461006300d65172f5668d8875f2aad7b54f7ba4e9c5435d65a84a5a2d66e39b)
matches `CentOS-7-x86_64-GenericCloud-1510.qcow2.xz`.
- According to http://cloud.centos.org/centos/7/images/sha256sum.txt ,
  the no-version-suffix `CentOS-7-x86_64-GenericCloud.qcow2.xz` is *supposed* to have same content as `CentOS-7-x86_64-GenericCloud-1503.qcow2.xz`.
  But the actual content has checksum c2e16bab8b5283cae77deb5bb0aad0a494cd4aabcb5dc5c71363056e88b93c0f, so it matches `CentOS-7-x86_64-GenericCloud-1511.qcow2.xz`.
- The last version mentioned in `sha256sum.txt` and actually appearing in the directory is `1601`.
  To make things more confusing, `sha256sum.txt.asc` has no mention of any `1601`.

Mismatch with the .txt files aside, I think the playbook should point to a specific `-YYMM` version so that it won't break in the future whenever `CentOS-7-x86_64-GenericCloud.qcow2.xz` gets updated.
(I could send a PR but I don't know which version to use.  I've tried both 1510 and 1511, and got unrelated errors, so can't even attest what's "working for me".)

cben:
Closes #1384.
`bin/cluster create libvirt` worked for me with both -1510 and -1602 images.

### DIFF
--- a/README_libvirt.md
+++ b/README_libvirt.md
@@ -118,7 +118,7 @@ The following options can be passed via the `-o` flag of the `create` command or
 * `image_url` (default to `http://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud.qcow2.xz`): URL of the QCOW2 image to download
 * `image_name` (default to `CentOS-7-x86_64-GenericCloud.qcow2`): Name of the QCOW2 image to boot the VMs on
 * `image_compression` (default to `xz`): Source QCOW2 compression (only xz supported at this time)
-* `image_sha256` (default to `9461006300d65172f5668d8875f2aad7b54f7ba4e9c5435d65a84a5a2d66e39b`): Expected SHA256 checksum of the downloaded image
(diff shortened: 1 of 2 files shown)


detiber:
Ok to test

detiber:
lgtm @brenton

PR #1506, opened 2016-02-26
Synchronize master kube configs

abutcher:
Testing with these changes. @detiber PTAL

https://bugzilla.redhat.com/show_bug.cgi?id=1311038

### DIFF
--- a/filter_plugins/openshift_master.py
+++ b/filter_plugins/openshift_master.py
@@ -501,7 +501,36 @@ class FilterModule(object):
                 valid = False
         return valid
 
+    @staticmethod
+    def certificates_to_synchronize(hostvars):
+        ''' Return certificates to synchronize based on facts. '''
+        if not issubclass(type(hostvars), dict):
+            raise errors.AnsibleFilterError("|failed expects hostvars is a dict")
+        certs = ['admin.crt',
+                 'admin.key',
+                 'admin.kubeconfig',
+                 'master.kubelet-client.crt',
+                 'master.kubelet-client.key',
+                 'openshift-registry.crt',
+                 'openshift-registry.key',
+                 'openshift-registry.kubeconfig',
+                 'openshift-router.crt',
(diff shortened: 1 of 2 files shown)


filter_plugins/openshift_master.py
```diff
@@ -501,7 +501,36 @@ def validate_pcs_cluster(data, masters=None):
                 valid = False
         return valid
 
+    @staticmethod
+    def certificates_to_synchronize(hostvars):
+        ''' Return certificates to synchronize based on facts. '''
+        if not issubclass(type(hostvars), dict):
+            raise errors.AnsibleFilterError("|failed expects hostvars is a string or unicode")
```
detiber: error message doesn't match test.

abutcher: derp

filter_plugins/openshift_master.py
```diff
@@ -501,7 +501,36 @@ def validate_pcs_cluster(data, masters=None):
                 valid = False
         return valid
 
+    @staticmethod
+    def certificates_to_synchronize(hostvars):
```
detiber: Should this actually filter for the certificates that are not already present on the host by default? With a param to sync all?

detiber: nm, I see that it is simply hardlinking the existing certs...  lgtm

detiber: I'm now following what you mean by not using the input for the filter plugin... This might be better suited for a lookup plugin instead of a filter plugin.

abutcher: I decided to base this on hostvars since all of the version booleans have already been created in there.

detiber: ah, not sure how I didn't catch that considering I verified the cert names used with the version conditions....

abutcher:
aos-ci-test

detiber:
:+1: @brenton

PR #1507, opened 2016-02-26
Add a check to see if yum or yum-utils is a problematic version

detiber:
Adds fail conditions when yum is < 3.4.3-132 or yum-utils is < 1.1.31-34
Resolves https://github.com/openshift/openshift-ansible/pull/1507
Resolves https://github.com/openshift/openshift-ansible/issues/1138

### DIFF
--- a/roles/openshift_facts/tasks/main.yml
+++ b/roles/openshift_facts/tasks/main.yml
@@ -4,6 +4,28 @@
     msg: "Unsupported ansible version: {{ ansible_version }} found"
   when: ansible_version.full | version_compare('1.9.4', 'lt') or ansible_version.full | version_compare('2.0', 'ge')
 
+- name: Get yum package version
+  command: rpm -q yum
+  when: ansible_os_family == 'RedHat' and ansible_distribution != 'Fedora'
+  changed_when: false
+  register: yum_version
+
+- name: Get yum-utils package version
+  command: rpm -q yum-utils
+  when: ansible_os_family == 'RedHat' and ansible_distribution != 'Fedora'
+  changed_when: false
+  register: yum_utils_version
+
+- name: Verify yum version is greater than or equal to 3.4.3-132
+  fail:
+    msg: "Please upgrade yum to version 3.4.3-132 or greater. The installed version is known to have issues with the Ansible yum module."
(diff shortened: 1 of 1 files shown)


detiber:
@sdodson @abutcher: ptal

detiber:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
:+1:

sdodson:
LGTM

detiber:
hah, ci is failing because it has problematic versions of yum.

detiber:
@arilivigni possible to get yum/yum-utils updated on the jenkins slave hosts running the CI jobs?

arilivigni:
@detiber done

detiber:
aos-ci-test

simon3z:
@detiber can't we attempt to update yum/yum-utils? (After all several packages get updated anyway during the process just by the fact that we install required packages, etc.)

detiber:
@simon3z We *could* but I hesitate to upgrade a core package like yum/yum-utils without notice.

Instead I would expect that to be handled by the conditional system package upgrade we discussed previously (or handled as a bootstrapping step).

arilivigni:
@detiber FYI I updated the package on the slave machine not the cluster machines that get deployed.  That would need to be added to our prep playbook

detiber:
@arilivigni ah, thanks for the heads up.

simon3z:
> @simon3z We could but I hesitate to upgrade a core package like yum/yum-utils without notice.

@detiber just to clarify, I meant to upgrade them only when they don't meet the minimum requirements.

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
@arilivigni any update on getting the slaves updated for this change?

arilivigni:
@detiber slave already has this I don't think this has anything to do with that.

detiber:
@arilivigni oops, I misread the previous comments, looks like it was the cluster machines being installed that were problematic...

arilivigni:
@detiber yes I assume this is another prerequisite so customers will need to do this?

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
aos-ci-test

dgoodwin:
aos-ci-test

dgoodwin:
I will rebase and do some testing with this on 7.1 where we're seeing the problem.

PR #1511, opened 2016-02-26
[WIP] Parallel pulls

detiber:
Does:
- configures parallel pulls by default if docker > 1.9 is present
- gathers and adds facts for versions of the following packages: etcd, docker, openshift
  - Doesn't (yet) support versions for containerized installs
- cleanup a few set if_unset methods in openshift_facts that no longer needed to be computed outside of get_defaults()
- add ability to query on is_containerized, deployment_type and version info within get_defaults() in openshift_facts

~~Depends on: https://github.com/openshift/openshift-ansible/pull/1550~~
Depends on: https://github.com/openshift/openshift-ansible/pull/1601
~~Depends on https://github.com/openshift/openshift-ansible/pull/1603~~

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -575,14 +575,12 @@ class FilterModule(object):
 
     @staticmethod
     # pylint: disable=too-many-branches
-    def oo_persistent_volumes(hostvars, groups, persistent_volumes=None):
+    def oo_persistent_volumes(hostvars, persistent_volumes=None):
         """ Generate list of persistent volumes based on oo_openshift_env
             storage options set in host variables.
         """
         if not issubclass(type(hostvars), dict):
             raise errors.AnsibleFilterError("|failed expects hostvars is a dict")
-        if not issubclass(type(groups), dict):
-            raise errors.AnsibleFilterError("|failed expects groups is a dict")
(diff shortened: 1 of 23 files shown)


detiber:
@abutcher ptal, this is how I envision wrapping the upstream type roles.

playbooks/common/openshift-docker/config.yml
```diff
@@ -1,9 +1,24 @@
...
+    when: openshift_docker_blocked_registries is not defined
+  - set_fact:
+      openshift_docker_options: "{{ lookup('oo_option', 'docker_options') }}"
+    when: openshift_docker_options is not defined
+  - set_fact:
+      openshift_docker_log_driver: "{{ lookup('oo_option', 'docker_log_driver') }}"
+    when: openshift_docker_log_driver is not defined
+  - set_fact:
+      openshift_docker_log_options: "{{ lookup('oo_option', 'docker_log_options') }}"
+    when: openshift_docker_log_options is not defined
```
detiber: One of these days we need to go in and update the oo_option code to no longer prepend 'cli_' to all environments. Either that or provide another way for bin/cluster to provide variables until we get it relegated to just env bootstrapping.

tbielawa:
This PR has been inactive for quite some time. Please update, fix any conflicts, and reopen this PR if this is still a priority you would like to see action on.

PR #1513, opened 2016-02-27
added use_cluster_metrics_hawkular

jdnieto:
I've created a new role to deploy hawkular metrics with openshift-ansible
Also added a new option --use_cluster_metrics_hawkular to deploy it


roles/openshift_cluster_metrics_hawkular/files/cluster-metrics/metrics.yaml
```diff
@@ -0,0 +1,98 @@
+apiVersion: "v1"
+kind: "Template"
+metadata:
+  name: metrics-deployer-template
+  annotations:
+    description: "Template for deploying the required Metrics integration. Requires cluster-admin 'metrics-deployer' service account and 'metrics-deployer' secret."
+    tags: "infrastructure"
```
dak1n1: We'll probably need some way to differentiate between the enterprise and origin metrics-deployer here. Maybe directly using the existing files instead of making a copy.

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_examples/files/examples/v1.1/infrastructure-templates/enterprise/metrics-deployer.yaml

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_examples/files/examples/v1.1/infrastructure-templates/origin/metrics-deployer.yaml

simon3z:
@cben can you review (and test) this?

roles/openshift_cluster_metrics_hawkular/tasks/main.yaml
```diff
@@ -0,0 +1,56 @@
...
+  with_items: "{{hawkular_tasks}}"
+  register: hawkular_perm_task
+  failed_when: "'already exists' not in hawkular_perm_task.stderr and hawkular_perm_task.rc != 0"
+  changed_when: hawkular_perm_task.rc == 0
+
+- name: Create Heapster Services
+  shell: >
+    {{ openshift.common.client_binary }} process -f
+    {{ openshift.common.config_base }}/cluster-metrics/metrics.yaml -v 
+    HAWKULAR_METRICS_HOSTNAME=hawkular-metrics.example.com,USE_PERSISTENT_STORAGE=false |
```
abutcher: Should both of these values be configurable? Note that we have `openshift_master_metrics_public_url` for providing the full metrics URL used in the master config.

simon3z: +1 to make this configurable (with also the `DURATION` of metrics). Then there's also `REDEPLOY` that you may take in consideration (not for exposing as an option, but as something you must add when something changed and you need to re-deploy).

jdnieto:
@dak1n1, @abutcher  required changes done.
@simon3z DURATION change in progress, today I've no more time :(

cben:
Tried the first patch via:

    bin/cluster -v create libvirt metrics2 --option use_cluster_metrics_hawkular=true

which run successfully but seemed to have some problems, notably AFAICT there was no heapster.
Now re-testing last version.

@jdnieto how are you running this?  libvirt?  byo?  something else?

jdnieto:
I run in libvirt and also in openstack.

El mié., 2 mar. 2016 14:51, Beni Cherniavsky-Paskin <
notifications@github.com> escribió:

> Tried the first patch via:
>
> bin/cluster -v create libvirt metrics2 --option use_cluster_metrics_hawkular=true
>
> which run successfully but seemed to have some problems, notably AFAICT
> there was no heapster.
> Now re-testing last version.
>
> @jdnieto <https://github.com/jdnieto> how are you running this? libvirt?
> byo? something else?
>
> —
> <https://github.com/openshift/openshift-ansible/pull/1513#issuecomment-191245593>
> .
>
-- 

Un saludo.
-------------------------------------------
José David Martín Nieto
RHCE, RHCJA, RHCJD, RHCoE-PaaS, RHCoE-Hibernate, RHCI, RHCX
#110-442-589

Public Key 2048R/24C474E5 C8B9 5589 9C19 1453 1184  EEAE A642 701E 24C4
74E5 @pool.sks-keyservers.net

roles/openshift_cluster_metrics_hawkular/vars/main.yaml
```diff
@@ -0,0 +1,17 @@
...
+    apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name: metrics-deployer
+    secrets:
+    - name: metrics-deployer
+
+hawkular_tmp_conf: /tmp/hawkular_admin.kubeconfig
+
+hawkular_hostname: {{ openshift.master.metrics_public_url | default("hawkular-metrics.example.com") }}
```
cben: `"{{ openshift.master.metrics_public_url | default('hawkular-metrics.example.com') }}"`

playbooks/common/openshift-cluster/additional_config.yml
```diff
@@ -20,6 +20,8 @@
     when: openshift.common.install_examples | bool
   - role: openshift_cluster_metrics
     when: openshift.common.use_cluster_metrics | bool
+  - role: openshift_cluster_metrics_hawkular 
+    when: openshift.common.use_cluster_metrics_hawkular | bool and not openshift.common.use_cluster_metrics | bool
   - role: openshift_manageiq
```
detiber: This should be moved down to the "Create persistent volumes and create hosted services" play.

playbooks/common/openshift-cluster/additional_config.yml
```diff
@@ -20,6 +20,8 @@
     when: openshift.common.install_examples | bool
   - role: openshift_cluster_metrics
     when: openshift.common.use_cluster_metrics | bool
+  - role: openshift_cluster_metrics_hawkular 
+    when: openshift.common.use_cluster_metrics_hawkular | bool and not openshift.common.use_cluster_metrics | bool
```
detiber: This should be moved down to the "Create persistent volumes and create hosted services" play.

I'd say the role should also be named openshift_metrics as well.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -393,6 +393,9 @@ def set_metrics_facts_if_unset(facts):
         if 'use_cluster_metrics' not in facts['common']:
             use_cluster_metrics = False
             facts['common']['use_cluster_metrics'] = use_cluster_metrics
+        if 'use_cluster_metrics_hawkular' not in facts['common']:
```
detiber: This can be set in get_defaults instead of here. The set_<something>_if_unset is an anti-pattern we are trying to get away from.

detiber: @abutcher thoughts on persisting values like deploy_<hosted_role> for metrics, registry, router, etc that would allow for forcing creation even if there are no infra hosts defined and/or no storage defined?

roles/openshift_node/templates/partials/kubeletArguments.j2
```diff
@@ -1,4 +1,4 @@
-{% if openshift.common.use_cluster_metrics | bool %}
+{% if openshift.common.use_cluster_metrics | bool or openshift.common.use_cluster_metrics_hawkular | bool %}
```
detiber: The read-only port should not be needed for hawkular, it was a workaround needed for the previous metrics role.

roles/openshift_facts/library/openshift_facts.py
abutcher: @detiber I like that idea - is that something you'd like to see here or an improvement we can make later?

playbooks/openstack/openshift-cluster/files/heat_stack.yaml
```diff
@@ -350,7 +350,6 @@ resources:
           port_range_min: 10250
           port_range_max: 10250
           remote_mode: remote_group_id
-          remote_group_id: { get_resource: master-secgrp }
```
detiber: Does this change the security access to the nodes group?

I'm almost inclined to say that we should limit it to "infra" nodes only.

roles/openshift_cluster_metrics_hawkular/files/cluster-metrics/enterprise-metrics.yaml
```diff
@@ -0,0 +1,98 @@
+apiVersion: "v1"
```
detiber: Instead of having the templates in here, the ones on disk should be used instead (openshift_examples). potentially we may need to land those templates even when the user chooses to skip the other templates.

roles/openshift_cluster_metrics_hawkular/tasks/main.yaml
```diff
@@ -0,0 +1,56 @@
...
+  command: >
+    cp {{ openshift.common.config_base }}/master/admin.kubeconfig {{hawkular_tmp_conf}}
+  changed_when: false
+
+- name: Install cluster metrics templates
+  copy:
+    src: cluster-metrics
+    dest: "{{ openshift.common.config_base }}/"
+
+- name: Create metrics-deployer Service Account
```
detiber: This should be done using the openshift_serviceaccounts role

roles/openshift_facts/library/openshift_facts.py
detiber: Later would be fine for that one. it would at least cut down on user confusion as to what magic creates the services.

roles/openshift_cluster_metrics_hawkular/tasks/main.yaml
abutcher: I'll take a look as soon as I can.

Outside of the scope of this, but I was thinking earlier about making the service account role delegate to the first master and then we could just depend on it where necessary.

detiber: genius!

roles/openshift_cluster_metrics_hawkular/vars/main.yaml
cben: Sorry, I was unclear, I meant to add outer double quotes around the {{...}}:
```
hawkular_hostname: "{{ openshift.master.metrics_public_url | default('hawkular-metrics.example.com') }}"
```
the change to single quote inside was just to not require escaping.
With your a9a6ba commit it still says:
```
ERROR: Syntax Error while loading YAML script, /home/bpaskinc/openshift-ansible/playbooks/common/openshift-cluster/roles/openshift_cluster_metrics_hawkular/vars/main.yaml
Note: The error may actually appear before this position: line 15, column 21


hawkular_hostname: {{ openshift.master.metrics_public_url | default('hawkular-metrics.example.com') }} 
                    ^
start a value. For instance:

    with_items:
      - {{ foo }}

Should be written as:

    with_items:
      - "{{ foo }}"
```

roles/openshift_cluster_metrics_hawkular/tasks/main.yaml
detiber: After discussing this a bit, @abutcher and I believe that this is something that we should handle as a later refactoring (considering the changes needed to the way we are handling service accounts, secrets and permissions).

jdnieto:
I close this PR in order to a best refactor of openshift-metrics.

PR #1519, opened 2016-02-29
Adding a symlink to making loading the examples more convenient

brenton:
This is related to https://bugzilla.redhat.com/show_bug.cgi?id=1300503


brenton:
@sdodson, PTAL.  Diffing symlink creation looks odd.  This is just a symlink so Docs doesn't have to keep updating their links whenever we ship new released. 

cc @tpoitras

sdodson:
@brenton LGTM

PR #1523, opened 2016-02-29
a-o-i: More flexible upgrade mappings

smunilla:
Change the upgrade logic to allow for more than two types of upgrades.


smunilla:
@brenton I was thinking something like this?

brenton:
Is that all that is needed to call the correct playbook?  The new playbook will be `playbooks/byo/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml`.

smunilla:
@brenton Yes. In a rare bit of foresight, I wrote the original code with exactly this in mind :)

smunilla:
Superseded by PR1536

Issue #1512, opened 2016-02-27
DNS issue in Openstack deployment


theone4ever:
Hi all,

I tried to create a cluster on Openstack. I used command below, openstack heat stack has been created without pb. In the created Neurtron subnet, DNS is 8.8.8.8 and 8.8.4.4 which is correct. But the DNS host created in this cluster (hostname is e2e-dns) use itself as nameserver instead of both itself and the two i provided (8.8.8.8 and 8.8.4.4 ), thus it can't access Internet to download package. Is this correct?

bin/cluster create -t origin -n 3 -o image_name=CentOS-7-Generic-Cloud -o external_net=net04_ext -o floating_ip_pool=net04_ext -o master_flavor=m1.medium -o node_flavor=m1.medium -o infra_flavor=m1.medium -o dns=8.8.8.8,8.8.4.4 openstack e2e

theone4ever:
I checked the source code, it seems there is dns forward support if the deployment type is enterprise, atomic-enterprise and so on, also it requires the host os as REHL. my deployment type is origin and using centos, anyway workaround?

lhuard1A:
@theone4ever 
What looks wrong in that setup is:

> In the created Neurtron subnet, DNS is 8.8.8.8 and 8.8.4.4

I expect the DNS of the Neutron subnet to be the IP of the dns VM

> But the DNS host created in this cluster (hostname is e2e-dns) use itself as nameserver instead of both itself and the two i provided (8.8.8.8 and 8.8.4.4 ), thus it can't access Internet to download package.

I expect the dns VM (`e2e-dns`) to use only the DNS that you specified on the `bin/cluster` command line and to ignore what is advertised by DHCP.

Here is what should “force” the DNS setting of the `e2e-test` VM: https://github.com/openshift/openshift-ansible/blob/master/playbooks/openstack/openshift-cluster/files/heat_stack.yaml#L605

theone4ever:
Hi Lenaic,

Thank you for your reply.
I double checked the deployment, actually i was wrong. The DNS server for DNS VM is actually itself, also in the neutron subnet, the DNS server is the DNS VM. 

But the problem comes later when ansible tries to install bind and bind-utils package on DNS VM. Because DNS VM uses itself as DNS server while bind and bind-utils (named service) are not installed, so it can't resolve any domain name on internet, so whole installation failed. 

Did i miss sth?

BR/Keven

theone4ever:
@lhuard1A 
[...]

PR #1526, opened 2016-03-01
Fix DNS problem on DNS server

eesprit:
Hi,

I think that is a solution to the #1512 issue.

With the actual interface configuration, the DNS server is using the dns resolvers in /etc/resolv.conf, which are obtained from DHCP, and are... itself... the problem is that the dns server need to install bind/named but is unable to resolv anything with such a configuration (chicken and egg problem).

With this change, the dns server only use the DNS servers which are defined in the interface configuration (which is obtained via cloud-init on boot), those are the DNS forwarders asked by the user (or 8.8.8.8 and 8.8.4.4 by default), so no problem to install the bind package and the rest of the installation process can go on.

I don't know if the DNS server should be able to resolv the other hosts after that, if that is the case the ansible tasks should change this parameter after the installation (and starting) of bind/named  and reload the network service.

Regards,

### DIFF
--- a/playbooks/openstack/openshift-cluster/files/heat_stack.yaml
+++ b/playbooks/openstack/openshift-cluster/files/heat_stack.yaml
@@ -604,7 +604,7 @@ resources:
                       BOOTPROTO="dhcp"
                       DNS1="$dns1"
                       DNS2="$dns2"
-                      PEERDNS="no"
+                      PEERDNS="yes"
                       ONBOOT="yes"
                 runcmd:
                   - [ "/usr/bin/systemctl", "restart", "network" ]


detiber:
ok to test

detiber:
@lhuard1A Can you review this PR? I do not see any issues with the change.

lhuard1A:
Hi @eesprit ,

It seems that what you observe is the exact opposite of what I am expecting.

What I expect is that:
— Neutron subnet is configured, as DNS, with the IP of the dns server.
— All the VMs but the dns one get from DHCP the IP of the dns VM as DNS.
— The dns VM ignores the DNS advertised by DHCP (to avoid the loop), hence the `PEERDNS="no"`
— The dns VM uses the DNS provided by the user, hence the `DNS1="$dns1"` and `DNS2="$dns2"` lines.
— bind is using the DNS provided by the user as forwarders.

What seems to be wrong is:

> With the actual interface configuration, the DNS server is using the dns resolvers in /etc/resolv.conf, which are obtained from DHCP, and are... itself...

I expect the DHCP to advertise the dns VM IP as DNS, but I want the dns VM to ignore what the DHCP advertises. And that behavior corresponds to [PEERDNS="no"](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Deployment_Guide/s1-networkscripts-interfaces.html)

```
PEERDNS=answer
    where answer is one of the following:

        yes — Modify /etc/resolv.conf if the DNS directive is set, if using DHCP, or if using Microsoft's RFC 1877 IPCP extensions with PPP. In all cases yes is the default.
        no — Do not modify /etc/resolv.conf. 
```

So, I have the feeling that this PR implements the opposite of what we need.

@eesprit could you double check?

eesprit:
I tried it 2 times with PEERDNS=no, it always failed on installing the named/bind package, connecting and interactively running yum install show that the server can't resolv DNS names.

After changing to PEERDNS=yes, all worked as expected.

The explanation from the redhat doc is quite fuzzy, with PEERDNS=yes, the /etc/resolv.conf is populated like that :

lhuard1A:
@eesprit : Could you check the content of `/etc/resolv.conf` on all the VMs ?
What is expected is:
— on the dns VM: `8.8.8.8` and `8.8.4.4` and nothing else;
— on all the other VMs: the IP on the dns VM and nothing else.

eesprit:
Sure, this is what I have after having set PEERDNS to yes :

[root@openshift-lab-master-0 master]# cat /etc/resolv.conf 
; generated by /usr/sbin/dhclient-script
search openstacklocal
nameserver 192.168.50.2

[root@openshift-lab-node-infra-0 ~]# cat /etc/resolv.conf 
; generated by /usr/sbin/dhclient-script
search openstacklocal
nameserver 192.168.50.2

[root@openshift-lab-node-compute-0 ~]# cat /etc/resolv.conf 
; generated by /usr/sbin/dhclient-script
search openstacklocal
nameserver 192.168.50.2

[root@openshift-lab-node-compute-1 ~]# cat /etc/resolv.conf 
; generated by /usr/sbin/dhclient-script
search openstacklocal novalocal
nameserver 192.168.50.2

[root@openshift-lab-dns named]# cat /etc/resolv.conf
; generated by /usr/sbin/dhclient-script
search openstacklocal
nameserver 8.8.8.8
nameserver 8.8.4.4
[root@openshift-lab-dns named]# cat /etc/sysconfig/network-scripts/ifcfg-eth0 
DEVICE="eth0"
BOOTPROTO="dhcp"
DNS1="8.8.8.8"
DNS2="8.8.4.4"
PEERDNS="yes"
ONBOOT="yes"

By the way, all the other nodes are using PEERDNS=yes (which seems logical if you thought it worked the either way) but are not impacted as they don't define DNS1 and DNS2 :

[root@openshift-lab-node-compute-1 ~]# cat /etc/sysconfig/network-scripts/ifcfg-eth0
DEVICE="eth0"
BOOTPROTO="dhcp"
ONBOOT="yes"
TYPE="Ethernet"
USERCTL="yes"
PEERDNS="yes"
IPV6INIT="no"
PERSISTENT_DHCLIENT="1"

lhuard1A:
So, if we focus on `openshift-lab-dns`, which is the machine which is malfunctioning, where bind installation fails,
* `/etc/resolv.conf` contains only `8.8.8.8` and `8.8.4.4`. So, DNS resolution should work. I don’t understand why yum fails.
* `ifcfg-eth0` contains `PEERDNS="yes"` which should cause `/etc/resolv.conf` to contain `192.168.50.2` in addition to `8.8.8.8` and `8.8.4.4` and it is not the case.

eesprit:
Perhaps I was not clear.

The configuration I gave is what I have after applying my change (PEERDNS=yes), so yes, with this conf, yum works.

With PEERDNS=no (for the dns instance only, via cloud-init), the server has the same config than the other nodes, only 192.168.50.2 (itself), so yum can't work.

I understand it is counter-inituitive and opposite to what the doc says (but the doc is not really clear), but this is what I am seeing here.

With PEERDNS=no, I can't see how the resolv.conf would contain the DNS1/DNS2 entries, as the goal is to have nothing editing the resolv.conf file. It is useful for manually overriding this file, but I didn't see anything in the provisionning code to modify it anyway ?

My understanding :
- PEERDNS=no => DHCP does not update the resolv.conf (but DNS1/2 are not used if defined)
- PEERDNS=yes and no DNS1/2 => resolv.conf use the DHCP obtained DNS servers
- PEERDNS=yes + DNS1/2 => the resolv.conf is updated with DNS1/2

I could be wrong, but this is what I am seeing there...

eesprit:
From the RH manual :

> DNS{1,2}=address
> where address is a name server address to be placed in /etc/resolv.conf **provided that the PEERDNS directive is not set to no**.

lhuard1A:
My understanding and the observations that I made on a RHEL 7.2 are:
* `PEERDNS=no` and `DNS[12]` set ⇒ `/etc/resolv.conf` contains `$DNS[12]`
* `PEERDNS=yes` and `DNS[12]` not set ⇒ `/etc/resolv.conf` contains the DNS advertised by DHCP
* `PEELDNS=yes` and `DNS[12]` set ⇒ `/etc/resolv.conf` contains both `$DNS[12]` and the DNS advertised by DHCP. In this case, whether DNS resolution worked on not was depending on the order and in my case, Murphy’s law applied: it was in the order that made things not working.

Maybe we are facing an issue of different behaviours between my RHEL 7.2 and your CentOS ?

eesprit:
Maybe that's it, but I am quite surprised of such a difference. I don't have RedHat subscriptions or images to test so. Perhaps the base images have differents behaviours (are your RedHat OS images using NetworkManager ?).

Perhaps a better way to deal with that is setting PEERDNS=no, do not define DNS1/2 in ifcfg-eth0 and to add an ansible task to edit the /etc/resolv.conf file for the dns instance (before installing the bind package) ?
That would made the behaviour consistent between CentOS/RH.

eesprit:
Alternatively, cloud-init allow to manage the resolv.conf file :

#    #cloud-config
#    manage_resolv_conf: true
#
#    resolv_conf:
#      nameservers: ['8.8.4.4', '8.8.8.8']
#      searchdomains:
#        - foo.example.com
#        - bar.example.com
#      domain: example.com
#      options:
#        rotate: true
#        timeout: 1

It is recommended to use it in conjunction with PEERDNS=no.

I'll give that a try and let you know.

detiber:
@eesprit where you able to get things working without this PR?

detiber:
Closing due to age, please rebase and reopen if it is still an issue.

PR #1533, opened 2016-03-02
Bug 1313169 - Ansible installer tries to enable etcd_container servic…

brenton:
…e even though containerized=false

### DIFF
--- a/roles/etcd/defaults/main.yaml
+++ b/roles/etcd/defaults/main.yaml
@@ -1,5 +1,5 @@
 ---
-etcd_service: "{{ 'etcd' if not openshift.common.is_containerized else 'etcd_container' }}"
+etcd_service: "{{ 'etcd' if not openshift.common.is_containerized | bool else 'etcd_container' }}"
 etcd_interface: "{{ ansible_default_ipv4.interface }}"
 etcd_client_port: 2379
 etcd_peer_port: 2380


brenton:
@detiber, @abutcher, PTAL.

This would have the side effect of always setting 'containerized' in the openshift fact json.

brenton:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
I think the problem may be here https://github.com/openshift/openshift-ansible/blob/master/roles/etcd/defaults/main.yaml#L2. Any value of `openshift.common.is_containerized` would result in `etcd_service = etcd_container` getting set since `is_containerized` isn't passed through `|  bool`.

abutcher:
Side note: CI failed here last few runs. We may need to change the failed condition for that task.

```
TASK: [common | Subscribe to OpenShift Enterprise subscription] *************** 
failed: [10.8.54.55] => {"changed": true, "cmd": "subscription-manager attach --pool=8a85f9823e3d5e43013e3ddd4e9509c4", "delta": "0:00:08.169204", "end": "2016-03-02 09:40:21.859763", "rc": 1, "start": "2016-03-02 09:40:13.690559", "warnings": []}
stdout: This unit has already had the subscription matching pool ID '8a85f9823e3d5e43013e3ddd4e9509c4' attached.
```

brenton:
@abutcher, does this look better?

abutcher:
@brenton I think that'll do it

detiber:
:+1:

nak3:
@brenton

To fix bz#1313169, I think this PR will work. But after I tried workaround this https://bugzilla.redhat.com/show_bug.cgi?id=1313169#c7, I hit one more issue as below.

~~~
TASK: [openshift_master_ca | Create the master certificates if they do not already exist] ***
failed: [master.example.com] => {"cmd": "/usr/local/bin/oadm create-master-certs --hostnames=kubernetes.default,10.67.112.120,kubernetes.default.svc.cluster.local,kubernetes,openshift.default....}
msg: [Errno 2] No such file or directory
~~~

`/usr/local/bin/oadm` is binary for the containerized installation. Should I file another report?

PR #1534, opened 2016-03-02
Fix logging infra template version mismatch.

dgoodwin:
Metrics was updated to 3.1.1, logging was not.


dgoodwin:
https://bugzilla.redhat.com/show_bug.cgi?id=1306678 probably where the mismatch occurred.

detiber:
lgtm (assuming we actually shipped a 3.1.1 version of logging).
@brenton

PR #1536, opened 2016-03-02
a-o-i: Support for unattended upgrades

smunilla:
Adds the ability to perform upgrades without any user input through:

atomic-openshift-installer --unattended upgrade --next-major
or
atomic-openshift-installer --unattended upgrade --latest-minor

### DIFF
--- a/utils/src/ooinstall/cli_installer.py
+++ b/utils/src/ooinstall/cli_installer.py
@@ -783,8 +783,10 @@ def uninstall(ctx):
 
 
 @click.command()
+@click.option('--latest-minor', '-l', is_flag=True, default=False)
+@click.option('--next-major', '-n', is_flag=True, default=False)
 @click.pass_context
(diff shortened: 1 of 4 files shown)


smunilla:
@brenton

brenton:
Could you rebase this on top of https://github.com/openshift/openshift-ansible/pull/1523 ?

detiber:
Now if only we could just pass in a master fqdn and have it discover the
hosts in the cluster from the master.

smunilla:
@brenton and…done

brenton:
It doesn't seem like this can handle upgrading to 3.1.1.90x versions.  Could you take a look at that?  It will be really helpful if you can install the GA version and then run the upgrade command to install that version.

smunilla:
I'll take a look to see why that would be. As is, it's just calling the v3_1_minor playbook.

PR #1537, opened 2016-03-03
Check for is_containerized value when setting binary locations.

abutcher:
Followup to #1533.

@nak3 @brenton

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1090,7 +1090,7 @@ def set_container_facts_if_unset(facts):
         if 'ovs_image' not in facts['node']:
             facts['node']['ovs_image'] = ovs_image
 
-    if facts['common']['is_containerized']:
+    if bool(strtobool(str(facts['common']['is_containerized']))):
         facts['common']['admin_binary'] = '/usr/local/bin/oadm'
         facts['common']['client_binary'] = '/usr/local/bin/oc'
 


abutcher:
aos-ci-test

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1090,7 +1090,7 @@ def set_container_facts_if_unset(facts):
         if 'ovs_image' not in facts['node']:
             facts['node']['ovs_image'] = ovs_image
 
-    if facts['common']['is_containerized']:
+    if facts['common']['is_containerized'] in ['true', 'True']:
```
detiber: it would probably be better to use strtobool when setting this value, similar to: https://github.com/abutcher/openshift-ansible/blob/bz1313169-oadm/roles/openshift_facts/library/openshift_facts.py#L769

abutcher:
aos-ci-test

roles/openshift_facts/library/openshift_facts.py
brenton: Wouldn't there be other places we'd need this same check for `is_containerized` in openshift_fact.py?  I feel like it would be better to make sure it's set correctly the fact is first set.  That's why I initially thought we should fix the problem in https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_facts/tasks/main.yml#L28 so that everywhere else we could assume that the value was a bool.  The side effect of that is that `is_containerized` would now be set on every host to either true or false instead of the assumption that only a missing value was false.

abutcher: When we don't set `is_containerized` it currently takes on the value of `is_atomic` which would be a boolean value in this section of the code. If set by the user then it will be a string instead so I think we have to treat it like a boolean everywhere we use it.

detiber: It might be better to enforce it as a boolean here: https://github.com/abutcher/openshift-ansible/blob/bz1313169-oadm/roles/openshift_facts/library/openshift_facts.py#L1419-L1426

The tricky part is that we'd need to make sure that if we've persisted the fact previously, we'll need to take that into account as well.

PR #1539, opened 2016-03-03
package up lib_zabbix into its own subpackage

joelddiaz:


joelddiaz:
@kwoodson this subpackage is the one python-openshift-tools-ansible would/should depend on

kwoodson:
:+1:

smunilla:
Doing a `tito build --test --rpm --install` now yield the error:

Command output: error: Failed dependencies:
        python-openshift-tools-zbxapi is needed by openshift-ansible-zabbix-3.0.52-1.git.6.7461853.fc22.noarch

where does the python-openshift-tools-zbxapi dep come from?

joelddiaz:
@smunilla it comes from here: https://github.com/openshift/openshift-tools/tree/master/openshift_tools
The packages should make it out to copr as well.

Issue #1497, opened 2016-02-25
HA-Proxy Config doesn't work for modestly large clusters ootb


timothysc:
On provisioning a 300 node cluster we see a number of connection issues during installation( xref: #1492 ). These issues occur for two reasons; 1 a flood of new connections, 2 haproxy config is not set to handle a reasonable # of connections.  

Currently we average ~20 connections per node on an idle cluster (ref: https://github.com/kubernetes/kubernetes/issues/21081) 

So we need to update our default config ~20000 max connections.  

Also, we need to tune our defaults due to our ssl-pass-through handling.  **Honestly I don't know why we don't do ssl-termination at the load-balancer ootb.**  This would drastically reduce load on the master components, and increase system responsiveness. (but we may want to open another issue on that one)

/cc @ncdc @danmcp @rrati @jeremyeder @eparis @liggitt

timothysc:
Another side note, the ulimit needs to be set on the service file.

jeremyeder:
Has anyone determined what those 20 connections actually are?  20k connections in the load balancer is asking for trouble.

timothysc:
In keeping score that's 6000 connections with ssl-passthrough, with bursts up to 10k when idle.

rrati:
That also isn't including the connections to the apiservers, which may also count towards the connection limit.  We are seeing ~1000 with busts of ~1700 connections per apiserver.

detiber:
> Also, we need to tune our defaults due to our ssl-pass-through handling. Honestly I don't know why we don't do ssl-termination at the load-balancer ootb. This would drastically reduce load on the master components, and increase system responsiveness. (but we may want to open another issue on that one)

The main reason that we don't currently do ssl-termination for the load balancer is that there is a chicken/egg scenario around creating the certificates and creating the load balancer that would require a pretty large refactoring to make work correctly. That said, I'm not sure we currently would even support disabling SSL to the api servers, since it would break ssl client cert authentication.

timothysc:
On large deployments I'd be pretty surprised to have such a large number of ssl-pass-throughs.  It's a computationally expensive mode of operation.

detiber:
@liggitt do we have a way to maintain ssl client based auth while terminating ssl at a load balancer for the api server?

liggitt:
like having the API trust a remote-user header from the proxy over a non-TLS connection? no...

detiber:
[...]

PR #1540, opened 2016-03-03
Add setting for configuring nofile limit for haproxy

detiber:
Untested so far, but should address nofile setting mentioned in https://github.com/openshift/openshift-ansible/issues/1497#issuecomment-191992788


detiber:
aos-ci-test

detiber:
@brenton I've just finished testing this, and I can verify that the drop-in file is being picked up by systemd and that systemctl show haproxy shows the LimitNOFILE setting from the drop-in file.

PR #1541, opened 2016-03-04
Add cockpit-docker package by default

nak3:


nak3:
Without `cockpit-docker` package, `container` button in the menu bar will not show up. And it is hard to find the issue and most of the OpenShift user this function.

sdodson:
LGTM, @brenton

PR #1542, opened 2016-03-04
Update README_origin.md

trondvh:
Added etcd and lb to the OSEv3:children group in the example host inventory. The playbook fails with a SSH exception if you're running the playbook with other users than root.


abutcher:
@brenton

PR #1548, opened 2016-03-04
Bug 1314645 - Upgrade failed with "One or more undefined variables 'd…

brenton:
…ict object' has no attribute 'stdout'"


brenton:
@detiber, @abutcher, PTAL

The way I was registering the variable wasn't working.  Whenver it was skipped it was overwriting the original one.  I incorrectly assumed nothing would be registered.

detiber:
@brenton yeah, the variable is registered regardless of the state of the task. 

lgtm

abutcher:
:+1:

PR #1553, opened 2016-03-07
Set node IMAGE_VERSION to use image_tag instead

andrewklau:
@brenton This broke my install for (CentOS 7.2 atomic), I don't think the openshift.common.version is set in some cases? I had to update the line for my install to work

Logs:
Mar 07 02:11:52 docker[19687]: Unable to find image 'openshift/node:v{# openshift.common.version.split('-')[0] #}' locally
Mar 07 02:11:57 docker[19687]: Trying to pull repository docker.io/openshift/node ... not found
Mar 07 02:11:57 docker[19687]: Tag v{# openshift.common.version.split('-')[0] #} not found in repository docker.io/openshift/node

### DIFF
--- a/roles/openshift_node/tasks/main.yml
+++ b/roles/openshift_node/tasks/main.yml
@@ -110,7 +110,7 @@
     - regex: '^CONFIG_FILE='
       line: "CONFIG_FILE={{ openshift_node_config_file }}"
     - regex: '^IMAGE_VERSION='
-      line: "IMAGE_VERSION=v{{ openshift.common.version.split('-')[0] }}"
+      line: "IMAGE_VERSION={{ openshift.common.image_tag }}"
   notify:
   - restart node
 


brenton:
This change looks reasonable to me.  I would like to know why `version` wasn't set.  I'll look through this today.

sdodson:
I need to checkout the latest code to see if you've already solved this via another means, but I had done some work to fix version introspection on containerized installs here https://github.com/openshift/openshift-ansible/pull/1099  I just never got around to addressing the feedback.

brenton:
@sdodson, pick up that PR when I start looking at this.

talset:
I tested with line: "IMAGE_VERSION={{ openshift.common.image_tag }}"
and it fix the deployment. I was able to setuped my ose on atomic

brenton:
@talset, @andrewklau, would you mind letting me know if the PR I pushed an hour or so ago resolves this problem?

andrewklau:
@brenton yep that seems to fix it. I'll close this.

brenton:
@andrewklau, thanks for the help!

PR #1559, opened 2016-03-07
BZ1315151: Support openshift_image_tag

brenton:
This is the containerized openshift_pkg_version equivalent.  Originally I was hoping to reuse openshift_pkg_version for containerized installs but the fact that it's very coupled to yum made that pretty ugly.                                          
                                                                             
However, I did opt to rely on the previously existing 'openshift_version' variable.  Containerized and RPM installs can both use that variable and it will be set appropriately if either openshift_pkg_version or openshift_image_tag are set.  I suspect someday containerized installs will be the only option and I didn't can to have thinkgs like openshift_pkg_version and openshift_image_tag in the playbooks anymore the necessary.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
@@ -40,6 +40,12 @@
         valid version for a {{ target_version }} upgrade
     when: openshift_pkg_version is defined and openshift_pkg_version.split('-',1).1 | version_compare(target_version ,'<')
 
+  - fail:
+      msg: >
+        openshift_image_tag is {{ openshift_image_tag }} which is not a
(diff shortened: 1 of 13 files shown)


brenton:
aos-ci-test

brenton:
@abutcher, @detiber PTAL

brenton:
This should handle the fix from https://github.com/openshift/openshift-ansible/pull/1553 as well.

brenton:
aos-ci-test

abutcher:
lgtm

detiber:
@brenton to be able to let users set openshift_version, it will need to be defined in defaults/main.yml instead of vars/main.yml. The variable precedence will force anything defined in vars/main.yml to override anything set by the user.

brenton:
@detiber, users would set either `openshift_pkg_version` or `openshift_image_tag`

detiber:
@brenton in that case, lgtm

PR #1572, opened 2016-03-09
Make sure install docker version 1.8.2. Fix issue can not start docker

hieuhuynh3001:


sdodson:
The real problem is we had a bad docker-1.9.1 build in our repo. I don't think we should do this.

sdodson:
@hieuhuynh3001 This should have already been addressed, can you try the latest code and close this if it's no longer a problem?

sdodson:
This is no longer relevant.

PR #1575, opened 2016-03-09
ManageIQ SA: Adding image-puller role

enoodle:
This will allow ManageIQ to run image-inspector when the registry requires authentication.

### DIFF
--- a/roles/openshift_manageiq/tasks/main.yaml
+++ b/roles/openshift_manageiq/tasks/main.yaml
@@ -18,9 +18,20 @@
   failed_when: "'already exists' not in osmiq_create_mi_project.stderr and osmiq_create_mi_project.rc != 0"
   changed_when: osmiq_create_mi_project.rc == 0
 
-- name: Create Service Account
+- name: Create Admin Service Account
   shell: >
     echo {{ manageiq_service_account | to_json | quote }} |
+    {{ openshift.common.client_binary }} create
+    -n management-infra
+    --config={{manage_iq_tmp_conf}}
+    -f -
+  register: osmiq_create_service_account
+  failed_when: "'already exists' not in osmiq_create_service_account.stderr and osmiq_create_service_account.rc != 0"
+  changed_when: osmiq_create_service_account.rc == 0
+
+- name: Create Image Inspector Service Account
+  shell: >
+    echo {{ manageiq_image_inspector_service_account | to_json | quote }} |
     {{ openshift.common.client_binary }} create
     -n management-infra
     --config={{manage_iq_tmp_conf}}
--- a/roles/openshift_manageiq/vars/main.yml
+++ b/roles/openshift_manageiq/vars/main.yml
@@ -15,6 +15,12 @@ manageiq_service_account:
     metadata:
       name: management-admin
 
+manageiq_image_inspector_service_account:
+    apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name: inspector-admin
+
 manage_iq_tmp_conf: /tmp/manageiq_admin.kubeconfig
 
 manage_iq_tasks:
@@ -22,3 +28,5 @@ manage_iq_tasks:
     - policy add-role-to-user -n management-infra management-infra-admin -z management-admin
     - policy add-cluster-role-to-user cluster-reader system:serviceaccount:management-infra:management-admin
     - policy add-scc-to-user privileged system:serviceaccount:management-infra:management-admin
+    - policy add-cluster-role-to-user system:image-puller system:serviceaccount:management-infra:inspector-admin
+    - policy add-scc-to-user privileged system:serviceaccount:management-infra:inspector-admin


enoodle:
@simon3z

simon3z:
cc @deads2k @detiber

roles/openshift_manageiq/vars/main.yml
```diff
@@ -22,3 +22,4 @@ manage_iq_tasks:
     - policy add-role-to-user -n management-infra management-infra-admin -z management-admin
     - policy add-cluster-role-to-user cluster-reader system:serviceaccount:management-infra:management-admin
     - policy add-scc-to-user privileged system:serviceaccount:management-infra:management-admin
+    - policy add-cluster-role-to-user system:image-puller system:serviceaccount:management-infra:management-admin
```
deads2k: It seems like you might want to start subdividing the power of your service accounts.  There's probably no reason for the pod running openscap to be able to read every deploymentconfig in the system, right?

simon3z:
@detiber do we need to do anything special to add it to the update from 3.1 to 3.2?

detiber:
ok to test

detiber:
aos-ci-test

detiber:
@simon3z assuming that @deads2k's concern is addressed related to the service account, we should be able to get this into the next puddle for QE.

arilivigni:
aos-ci-test

detiber:
aos-ci-test

enoodle:
I think that if we will add another SA here then I will also take the opportunity to refactor this whole role to use the new ```oc sa``` command. (The manageiq feature that requires this will need some time to be integrated any way..)

simon3z:
:+1 to separate this to an additional service account.

enoodle:
@detiber after 3.2 will be out, will we still want to preserve this backward compatibility? I wanted to have a simpler logic, not an even more complicated one..

detiber:
On Mar 15, 2016 9:36 AM, "Erez Freiberger" <notifications@github.com> wrote:
>
> @detiber after 3.2 will be out, will we still want to preserve this
backward compatibility? I wanted to have a simpler logic, not an even more
complicated one..

Understood, the goal is to still support earlier releases until the are
EOL. Otherwise we would have to maintain parallel release branches for all
supported versions.

simon3z:
@enoodle please try to split the permissions on a different SA (and address @detiber comment if relevant).

enoodle:
@simon3z done :)

detiber:
aos-ci-test

simon3z:
@deads2k to create privileged pods you need a service account that is allowed to do that... then can you specify a different `serviceAccount` for the pod has less permissions?

If not then we have to turn also `management-image-scanner` into a privileged service account.

arilivigni:
aos-ci-test

deads2k:
> If not then we have to turn also management-image-scanner into a privileged service account.

You'll have to make it privileged.  There's no other reasonable way to achieve this.

arilivigni:
aos-ci-test

arilivigni:
aos-ci-test

arilivigni:
aos-ci-test

arilivigni:
aos-ci-test

detiber:
@arilivigni We probably want to skip the e2e tests for 3.0. Also, the error messages point to the change in configuration directory between 3.0 and 3.1 (/etc/origin/* instead of /etc/openshift/*).

arilivigni:
@detiber Yes I assumed they would run on 3.0 I am changing that no.  I am not sure what you mean by the other issue but i assume once I stop running for 3.0 this won't be an issue.

arilivigni:
aos-ci-test

arilivigni:
aos-ci-test

detiber:
@arilivigni correct, it won't be an issue for testing against 3.1+.

For version 3.0 we used to deploy the configuration files to /etc/openshift/master and /etc/openshift/node. With version 3.1+ we started using /etc/origin/master and /etc/origin/node instead.

arilivigni:
aos-ci-test

simon3z:
@detiber let's not merge this yet. I think we may have to address what @deads2k mentioned above (privileged).

arilivigni:
aos-ci-test

arilivigni:
aos-ci-test

arilivigni:
@detiber I hope you don't mind I am using this PR to test the updates to the CI contexts

detiber:
@arilivigni I don't mind, but you may want to switch over to a PR that doesn't have external (to the OpenShift team) contributors on it :)

roles/openshift_manageiq/tasks/main.yaml
```diff
@@ -29,6 +29,17 @@
   failed_when: "'already exists' not in osmiq_create_service_account.stderr and osmiq_create_service_account.rc != 0"
   changed_when: osmiq_create_service_account.rc == 0
 
+- name: Create Image Scanner Service Account
```
simon3z: I don't like much "image scanner" ...after all... why inventing another term (like "scanner") when we already have "inspector".

Although I am not sure I like `management-image-inspector`... any other suggestion? @enoodle @detiber @deads2k

detiber: I'd personally be okay with miq-<name> instead of management-<name> management almost seems like it is a built in SA and not an integration.

enoodle: maybe miq-inspector ?

arilivigni:
aos-ci-test

arilivigni:
running this one more time to cleanup the mess

simon3z:
Few comments:

1. @enoodle I prefer to change the name of the service account to `inspector-admin`

2. @deads2k I checked and it seems that the creating service account needs to have the `privileged` scc, but it's not required in the specificed `serviceAccount`. E.g.

```
# oc whoami
system:serviceaccount:management-infra:management-admin

# oc get scc privileged -o yaml
...
- system:serviceaccount:openshift-infra:build-controller
- system:serviceaccount:management-infra:management-admin
- system:serviceaccount:default:router
- system:serviceaccount:default:registry

# oc create -f - <<EOF
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  serviceAccount: inspector-admin
  containers:
  - name: nginx
    image: nginx
    securityContext:
      privileged: true
EOF
pod "nginx" created

# oc get pods -o yaml
...
  spec:
    containers:
    - image: nginx
      ...
      securityContext:
        privileged: true
    ...
    serviceAccount: inspector-admin
    serviceAccountName: inspector-admin


# oc exec -ti nginx whoami
root
```

This seems to contradict [your previous comment](https://github.com/openshift/openshift-ansible/pull/1575#issuecomment-197447754), or anyway there was a misunderstanding (or there's a bug).

deads2k:
> @deads2k I checked and it seems that the creating service account needs to have the privileged scc, but it's not required in the specificed serviceAccount

In order to use this `PodSpec` in a `Job` or other structure that causes a system controller to create the `Pod` on your behalf, the SA you use in the `PodSpec` has to have sufficient power.  Creating a one-off Pod, can be done with a caller's rights, but we generally recommend against it because it prevents you from using higher level constructs.

In the case of the ImageInspector, a `Job` seems like a natural fit in the future, so unless unless there's something else in play, I'd have my `PodSpec` specify an SA with sufficient SCC privileges.

roles/openshift_manageiq/tasks/main.yaml
```diff
@@ -29,6 +29,17 @@
...
+- name: Create Image Inspector Service Account
+  shell: >
+    echo {{ manageiq_image_inspector_service_account | to_json | quote }} |
+    {{ openshift.common.client_binary }} create
+    -n management-infra
+    --config={{manage_iq_tmp_conf}}
+    -f -
+  register: osmiq_create_service_account
+  failed_when: "'already exists' not in osmiq_create_service_account.stderr and osmiq_create_service_account.rc != 0"
+  changed_when: osmiq_create_service_account.rc == 0
```
detiber: I'm not going to ask for you to refactor this now, since we are already doing this in multiple places, just wanted to add a comment and get feedback from @abutcher and @sdodson on trying to move away from this pattern and separating these into two separate tasks, one that checks if the account exists and another that attempts to create the account if it doesn't exist.

Thoughts?

roles/openshift_manageiq/vars/main.yml
```diff
@@ -15,6 +15,12 @@ manageiq_service_account:
     metadata:
       name: management-admin
 
+manageiq_image_inspector_service_account:
+    apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name: miq-inspector
```
detiber: Again, not asking for changes, just gathering general feedback for future improvements. @abutcher @sdodson thoughts on having a filter that builds up the service_account template instead of templating it out all the time?

simon3z:
LGTM :+1: and ready for merge unless there are critical issues brought up in the comments to be addressed cc @detiber

detiber:
aos-ci-test

detiber:
lgtm pending the test results. @brenton

detiber:
@brenton should be ready to go.

simon3z:
> In the case of the ImageInspector, a Job seems like a natural fit in the future, so unless unless there's something else in play, I'd have my PodSpec specify an SA with sufficient SCC privileges.

Ah thanks @deads2k

PR #1576, opened 2016-03-09
Bug 1315564 - upgrade to ose3.2 failed on Atomic Hosts

brenton:
Previously I was greping for 'ose' in the systemd units.  That was only working
on my machine because my Nodes were also Masters.  It's safer to grep for
openshift3 since that would be present for Masters or Nodes.


detiber:
lgtm

brenton:
Thx.  I'm going to include another bug fix in this PR shortly.

brenton:
@detiber, ok this should be ready for review.  I've successfully run it in QE's environment.  The situation was that `openshift.common.is_atomic` was blank because openshift_facts hadn't been run.  I changed the docker_upgrade.yml to a list of tasks instead of a playbook so that I could apply the `openshift_facts` role properly.  I also did it for the Master and Node upgrades.

detiber:
@brenton :+1:

PR #1582, opened 2016-03-10
added new openshift-metrics service

jdnieto:
It replaces PR #1513

### DIFF
--- a/playbooks/common/openshift-cluster/additional_config.yml
+++ b/playbooks/common/openshift-cluster/additional_config.yml
@@ -52,3 +52,5 @@
   - role: openshift_registry
     registry_volume_claim: "{{ openshift.hosted.registry.storage.volume.name }}-claim"
     when: deploy_infra | bool and attach_registry_volume | bool
+  - role: openshift_metrics
+    when: openshift.hosted.metrics.deploy | bool
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -35,6 +35,14 @@
   - set_fact:
       openshift_master_debug_level: "{{ t_oo_option_master_debug_level }}"
     when: openshift_master_debug_level is not defined and t_oo_option_master_debug_level != ""
+    
+  - set_fact:
+      openshift_master_default_subdomain: "{{ lookup('oo_option', 'openshift_master_default_subdomain') | default(None) }}"
+    when: openshift_master_default_subdomain is not defined
+  - set_fact:
+      openshift_hosted_metrics_deploy: "{{ lookup('oo_option', 'openshift_hosted_metrics_deploy') | default(false, true) }}"
+    when: openshift_hosted_metrics_deploy is not defined
+    
   roles:
   - openshift_facts
   post_tasks:
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1576,6 +1576,22 @@ class OpenShiftFacts(object):
 
         if 'hosted' in roles or self.role == 'hosted':
             defaults['hosted'] = dict(
+                metrics=dict(
+                    deploy=False,
+                    storage=dict(
+                        kind=None,
+                        volume=dict(
+                            name='metrics',
+                            size='10Gi'
+                        ),
+                        nfs=dict(
+                            directory='/exports',
+                            options='*(rw,root_squash)'),
+                        host=None,
+                        access_modes=['ReadWriteMany'],
+                        create_pv=True
+                    )
+                ),
                 registry=dict(
                     storage=dict(
                         kind=None,
--- a/roles/openshift_metrics/README.md
+++ b/roles/openshift_metrics/README.md
@@ -1,0 +1,51 @@
+OpenShift Metrics with Hawkular
+====================
+
+OpenShift Metrics Installation
+
+Requirements
+------------
+It requires subdomain fqdn to be set.
+If persistence is enabled, then it also requires NFS 
+
+Role Variables
+--------------
+
+From this role:
+| Name                                            | Default value         |                                                             |
+|-------------------------------------------------|-----------------------|-------------------------------------------------------------|
+| openshift_hosted_metrics_deploy                 | False                 | If metrics should be deployed                               |
+| openshift_hosted_metrics_storage_nfs_directory  | /exports              | Root export directory.                                      |
+| openshift_hosted_metrics_storage_volume_name    | metrics               | Metrics volume within openshift_hosted_metrics_volume_dir   |
+| openshift_hosted_metrics_storage_volume_size    | 10Gi                  | Metrics volume size                                         |
+| openshift_hosted_metrics_storage_nfs_options    | *(rw,root_squash)     | NFS options for configured exports.                         |
+
+
+From openshift_common:
+| Name                                  | Default Value  |                                        |
+|---------------------------------------|----------------|----------------------------------------|
+| openshift_master_default_subdomain    | null           | Subdomain FQDN (Mandatory)             |
+
+
+Dependencies
+------------
+openshift_facts
+openshift_examples
+
+Example Playbook
+----------------
+
+- name: Configure openshift-metrics
+  hosts: oo_first_master
+  roles:
+  - role: openshift_metrics
+
+License
+-------
+
+Apache License, Version 2.0
+
+Author Information
+------------------
+
+Jose David Martín (j.david.nieto@gmail.com)
--- a/roles/openshift_metrics/meta/main.yaml
+++ b/roles/openshift_metrics/meta/main.yaml
@@ -1,0 +1,3 @@
+dependencies:
+- { role: openshift_examples }
+- { role: openshift_facts }
--- a/roles/openshift_metrics/tasks/main.yaml
+++ b/roles/openshift_metrics/tasks/main.yaml
@@ -1,0 +1,55 @@
+---
+- name: Copy Configuration to temporary conf
+  command: >
+    cp {{ openshift.common.config_base }}/master/admin.kubeconfig {{hawkular_tmp_conf}}
+  changed_when: false
+
+- name: Create metrics-deployer Service Account
+  shell: >
+    echo {{ deployer_service_account | to_json | quote }} |
+    {{ openshift.common.client_binary }} create
+    -n openshift-infra
+    --config={{hawkular_tmp_conf}}
+    -f -
+  register: deployer_create_service_account
+  failed_when: "'already exists' not in deployer_create_service_account.stderr and deployer_create_service_account.rc != 0"
+  changed_when: deployer_create_service_account.rc == 0
+
+- name: Create metrics-deployer Secret
+  command: >
+    {{ openshift.common.client_binary }}
+    secrets new metrics-deployer
+    nothing=/dev/null
+    -n openshift-infra
+  register: deployer_create_secret
+  failed_when: "'already exists' not in deployer_create_secret.stderr and deployer_create_secret.rc !=0"
+  changed_when: deployer_create_secret.rc == 0
+
+- name: Configure role/user permissions
+  command: >
+    {{ openshift.common.admin_binary }} {{item}}
+    --config={{hawkular_tmp_conf}}
+  with_items: "{{hawkular_permission_oc_commands}}"
+  register: hawkular_perm_task
+  failed_when: "'already exists' not in hawkular_perm_task.stderr and hawkular_perm_task.rc != 0"
+  changed_when: hawkular_perm_task.rc == 0
+
+- name: Check openshift_master_default_subdomain
+  fail: 
+    msg: "Default subdomain should be defined"
+  when: openshift.master.default_subdomain is not defined
+
+- name: Create Heapster and Hawkular/Cassandra Services
+  shell: >
+   {{ openshift.common.client_binary }} process -f \
+   /usr/share/openshift/examples/infrastructure-templates/{{ hawkular_type }}/metrics-deployer.yaml -v \
+    HAWKULAR_METRICS_HOSTNAME=hawkular-metrics.{{ openshift.master.default_subdomain }},USE_PERSISTENT_STORAGE={{ hawkular_persistence }} | \
+    {{ openshift.common.client_binary }} create -n openshift-infra -f - 
+  register: oex_heapster_services
+  failed_when: "'already exists' not in oex_heapster_services.stderr and oex_heapster_services.rc != 0"
+  changed_when: false
+
+- name: Clean temporary config file
+  command: >
+    rm -rf {{hawkular_tmp_conf}}
+  changed_when: false
--- a/roles/openshift_metrics/vars/main.yaml
+++ b/roles/openshift_metrics/vars/main.yaml
@@ -1,0 +1,19 @@
+hawkular_permission_oc_commands:
+    - policy add-role-to-user edit system:serviceaccount:openshift-infra:metrics-deployer -n openshift-infra
+    - policy add-cluster-role-to-user cluster-admin system:serviceaccount:openshift-infra:heapster
+
+deployer_service_account:
+    apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name: metrics-deployer
+    secrets:
+    - name: metrics-deployer
+
+
+hawkular_tmp_conf: /tmp/hawkular_admin.kubeconfig
+
+hawkular_persistence: "{% if openshift.hosted.metrics.storage.kind != None %}true{% else %}false{% endif %}"
+
+hawkular_type: "{{ 'origin' if deployment_type == 'origin' else 'enterprise' }}"
+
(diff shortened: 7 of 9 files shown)


playbooks/common/openshift-cluster/additional_config.yml
```diff
@@ -54,3 +54,5 @@
   - role: openshift_registry
     registry_volume_claim: "{{ openshift.hosted.registry.storage.volume.name }}-claim"
     when: deploy_infra | bool and attach_registry_volume | bool
+  - role: openshift_metrics
+    when: use_openshift_metrics | bool
```
detiber: @abutcher can you make recommendations to tie this into the persistent storage work...  then this condition could be:
```
when: deploy_infra | bool and attach_metrics_volumes | bool
```

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -395,6 +395,23 @@ def set_metrics_facts_if_unset(facts):
...
+            facts (dict): existing facts
+        Returns:
+            dict: the facts dict updated with the generated openshift metrics
+            facts if they were not already present
+
+    """
+    if 'common' in facts:
+        if 'use_cluster_metrics_hawkular' not in facts['common']:
+            use_cluster_metrics_hawkular = False
+            facts['common']['use_cluster_metrics_hawkular'] = use_cluster_metrics_hawkular
```
detiber: A few suggestions here:
- Since this is a static "fact", it should be set in the get_defaults method. set_*_if_unset methods were originally introduced to solve setting facts based on the runtime value of other facts and has since been used for simple facts (which we are working on cleaning up).
- From the previous use, it looks like use_cluster_metrics_hawkular should be use_openshift_metrics. I'm not a huge fan of that naming though...  possibly openshift_hosted_metrics_deploy=True/False for the user exposed variable, and the fact could be openshift.hosted.metrics.deploy ? @abutcher

roles/openshift_metrics/meta/main.yaml
```diff
@@ -0,0 +1,2 @@
+dependencies:
+- role: openshift_examples
```
detiber: Will need a dependency on openshift_facts as well.

detiber:
ok to test

detiber:
@dak1n1 @tstclair @jerms @simon3z  relevant to your interests

@abutcher ptal

roles/openshift_facts/library/openshift_facts.py
jdnieto: I'm not sure I understand well the first suggestion.

abutcher: @jdnieto Rather than creating a `set_openshift_metrics_facts_if_unset` function, @detiber is saying that we should move the fact setting to `get_defaults`.

simon3z:
@cben can you give this a try?

cben:
Already looking at it.

playbooks/common/openshift-cluster/additional_config.yml
abutcher: To integrate metrics with persistent storage, take a look at the variables we can set in the [BYO inventories](https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.ose.example#L224-L256) and also take a look at this PR https://github.com/openshift/openshift-ansible/pull/1173.

We would want to provide a set of metrics variables in the example inventories similar to the registry variables. If present in the inventory, anything beginning with `openshift_hosted*` variable is persisted by [this call to openshift_facts](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_common/tasks/main.yml#L41-L46). [Defaults for these variables](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_facts/library/openshift_facts.py#L1332-L1348) are set in `openshift_facts`.


For example,

```
openshift_hosted_metrics_storage_kind=nfs
openshift_hosted_metrics_storage_nfs_directory=/exports
openshift_hosted_metrics_storage_volume_name=metrics
openshift_hosted_metrics_storage_access_modes=['ReadWriteMany']
```

These variables are persisted in the facts dictionary like this:

```
facts['openshift']['hosted']['metrics']['storage']['kind'] = 'nfs'
facts['openshift']['hosted']['metrics']['storage']['nfs']['directory'] = '/exports'
...
```

An nfs host is configured via the `openshift_storage_nfs` role which makes use of these variables to create volumes. This role currently only creates a registry volume and would need to be extended to create a metrics volume.

In the [additional cluster config playbook](https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/additional_config.yml), we process hostvars with `oo_persistent_volumes` and `oo_persistent_volume_claims` filters to return a list of persistent volumes and claims to create. Both filters are component agnostic and look at `openshift_hosted*` persisted facts to determine which volumes and claims to make. Using the variables above, we'd end up with a `metrics-volume` PV and `metrics-claim` PVC and we can use these when deploying metrics.

cben:
Tried this.  Full details: https://gist.github.com/cben/6101fbb918d2a751796c
- Good: The services are all running.
- Bad: `metricsPublicURL` in `master-config.yaml` was missing.  Easy to fix.
- Bad: https://metrics4-master-d780a.example.com/hawkular/metrics and https://metrics4-master-d780a.example.com/hawkular/metrics/status don't respond.  Got stuck here, don't know how to debug.

I assume this is the relevant route?
```
[openshift@metrics4-master-d780a ~]$ oc describe route -n openshift-infra
Name:           hawkular-metrics
Created:        4 hours ago
Labels:         metrics-infra=support
Annotations:        <none>
Requested Host:     hawkular-metrics.example.com
              exposed on router router 3 hours ago
Path:           <none>
TLS Termination:    passthrough
Insecure Policy:    <none>
Service:        hawkular-metrics
Endpoint Port:      <all endpoint ports>
Endpoints:      10.1.2.2:8443
```
Is it suspicious that `hawkular-metrics.example.com` doesn't resolve on DNS?
Sending `Host: hawkular-metrics.example.com` to the master doesn't help - master is not even listening on port 443 (nor 80).

roles/openshift_metrics/vars/main.yaml
```diff
@@ -0,0 +1,19 @@
+hawkular_tasks:
```
cben: To a naive reader (e.g. me), "tasks" sounds like these would be ansible tasks.
Perhaps rename to "hawkular_permission_oc_commands"?

roles/openshift_metrics/tasks/main.yaml
```diff
@@ -0,0 +1,50 @@
...
+- name: Configure role/user permissions
+  command: >
+    {{ openshift.common.admin_binary }} {{item}}
+    --config={{hawkular_tmp_conf}}
+  with_items: "{{hawkular_tasks}}"
+  register: hawkular_perm_task
+  failed_when: "'already exists' not in hawkular_perm_task.stderr and hawkular_perm_task.rc != 0"
+  changed_when: hawkular_perm_task.rc == 0
+
+- name: Create Heapster Services
```
cben: Perhaps call this "Create Heapster and Cassandra services"?
IIUC "Heapster Services" doesn't necessarily imply cassandra, eg. origin-metrics README documents "Deploying Only Heapster" vs "Deploying all of the Metrics Components".

jdnieto:
@cben it's supposed that "metricsPublicUrl" points to hawkular-metrics route:
`[openshift@flynn-master-0 ~]$ oc get route
NAME               HOST/PORT                      PATH      SERVICE            LABELS                  INSECURE POLICY   TLS TERMINATION
hawkular-metrics   hawkular-metrics.example.com             hawkular-metrics   metrics-infra=support                     passthrough
`
This route will redirect to hawkular-metrics service which have the HTTPS port
`[openshift@flynn-master-0 ~]$ oc get svc
NAME                       CLUSTER_IP       EXTERNAL_IP   PORT(S)                               SELECTOR                  AGE
hawkular-cassandra         172.30.246.19    <none>        9042/TCP,9160/TCP,7000/TCP,7001/TCP   type=hawkular-cassandra   6h
hawkular-cassandra-nodes   None             <none>        9042/TCP,9160/TCP,7000/TCP,7001/TCP   type=hawkular-cassandra   6h
hawkular-metrics           172.30.249.125   <none>        443/TCP                               name=hawkular-metrics     6h
heapster                   172.30.145.253   <none>        80/TCP                                name=heapster             6h
`

playbooks/common/openshift-cluster/additional_config.yml
jdnieto: Working on it

roles/openshift_metrics/meta/main.yaml
jdnieto: Done! Will be reflected in next commit

roles/openshift_metrics/vars/main.yaml
jdnieto: Ok, changed in future commit

roles/openshift_metrics/tasks/main.yaml
jdnieto: Ok, changed in future commit

detiber:
@jdnieto @cben is example.com also the same domain used for the master/node fqdns? If so, that would be a problem. You will want the application domain used by the router to be disjoint from the one used by the host infrastructure, otherwise there is a conflict when using the wildcard on the router.

jdnieto:
@detiber, do you consider is good to change default "hawkular-metrics.example.com" to "hawkular-metrics.metricsample.com"?

detiber:
@jdnieto works for me, I just know that example.com will conflict with a lot of test environments, especially ones that were originally based off of https://github.com/openshift/training

abutcher:
Would we not want to require that the metrics hostname be provided by the user so there aren't any surprises @detiber @jdnieto ?

jdnieto:
The metrics hostname is optional, can be configured passing an option, but defaults relay on hawkular-metrics.example.com. Should be changed to be required instead of optional?

abutcher:
The [docs say that the hostname is required](https://docs.openshift.com/enterprise/3.1/install_config/cluster_metrics.html#deployer-template-parameters) and changing the hostname would mean redeploying metrics. I think we should make sure that it has been explicitly set up front.

detiber:
+1 to setting it upfront...  that or if openshift_master_default_subdomain is set then we could default to `"hawkular_metrics.{{ openshift.master.default.subdomain }}"`

cben:
> is example.com also the same domain used for the master/node fqdns?

Yes, my hosts are:
```
metrics4-master-d780a.example.com
metrics4-node-infra-48f49.example.com
metrics4-node-compute-0e7b2.example.com
metrics4-node-compute-0e7b2.example.com
```
as created by `bin/cluster -v create libvirt metrics4 --option use_openshift_metrics=true`.
(They're resolvable both inside and outside thanks to
```
> cat /etc/NetworkManager/dnsmasq.d/libvirt_dnsmasq.conf
server=/example.com/192.168.55.1
```
as recommended in README_libvirt.md.)

Gonna test with `openshift_master_default_subdomain`.  BTW, use `hawkular-metrics.{{...}}` rather than `hawkular_metrics.{{...}}` (underscore is somewhat-working-but-not-quite-legal in host names).

jdnieto:
Reactor almost done.

jdnieto:
@detiber It looks good now?

playbooks/common/openshift-master/config.yml
```diff
@@ -69,6 +73,11 @@
       openshift_env:
         openshift_hosted_registry_storage_kind: 'nfs'
     when: openshift_hosted_registry_storage_kind is not defined and groups.oo_nfs_to_config is defined and groups.oo_nfs_to_config | length > 0
+  - openshift_facts:
```
detiber: you probably want to push this openshift_facts call down into the openshift_metrics role instead of doing it in this playbook.

you will however want to do a set_fact for openshift_hosted_metrics_deploy using the oo_option lookup if openshift_hosted_metrics_deploy is not defined still....

playbooks/common/openshift-master/config.yml
```diff
@@ -69,6 +73,11 @@
       openshift_env:
         openshift_hosted_registry_storage_kind: 'nfs'
     when: openshift_hosted_registry_storage_kind is not defined and groups.oo_nfs_to_config is defined and groups.oo_nfs_to_config | length > 0
```
detiber: @abutcher we probably want to set the default for openshift_hosted_registry_storage_kind somewhere outside of this playbook....   possibly in the openshift_storage_nfs role when creating the pvs?

roles/openshift_storage_nfs/defaults/main.yml
```diff
@@ -8,6 +8,14 @@ openshift:
           options: "*(rw,root_squash)"
         volume:
           name: "registry"
+    metrics:
```
detiber: should probably set `kind: nfs` here as well.

detiber:
@abutcher can you review the storage related parts of this PR?

playbooks/common/openshift-master/config.yml
abutcher: @detiber I was moving it into openshift_master_facts in the delegate refactor since we're interacting with masters to create these

detiber: @abutcher that works, I'm assuming you are running those steps using the oo_masters_to_config group and doing a limit of 1 then?

abutcher: @detiber Yep and this future group leak will use a var representing the list of nfs hosts | default([]).

abutcher: We probably want to set `openshift_hosted_metrics_storage_kind: 'nfs'` when an nfs host group is defined.

abutcher:
Storage related changes are heading in the right direction. `roles/openshift_storage_nfs/tasks/main.yml` contains a few registry specific directory creates that will need to be duplicated/extended to include metrics. I'm also not sure that we'll end up using the correct persistent volumes when we do the deploy unless we attach them ourselves?

jdnieto:
@detiber @abutcher  updated.
About persistent-volumes, in openshift-metrics, I think is better not to be automatic deployed. I think is better to create manually pv and pvc to ensure the use of the correct pv.

jdnieto:
@detiber Can be reviewed?

detiber:
@abutcher ptal

roles/openshift_metrics/tasks/main.yaml
```diff
@@ -0,0 +1,55 @@
...
+    {{ openshift.common.admin_binary }} {{item}}
+    --config={{hawkular_tmp_conf}}
+  with_items: "{{hawkular_permission_oc_commands}}"
+  register: hawkular_perm_task
+  failed_when: "'already exists' not in hawkular_perm_task.stderr and hawkular_perm_task.rc != 0"
+  changed_when: hawkular_perm_task.rc == 0
+
+- name: Check openshift_master_default_subdomain
+  fail: 
+    msg: "Default subdomain should be defined"
```
enoodle: Shouldn't this fail? Or print a more erroneous message to help debug this situation?

jdnieto: @enoodle Yes, it should fail if openshift_master_default_subdomain is not defined.
I think error message is self-explanatory, but if you want it will be more clearly can be added needed variable.

enoodle: Thank you, its ok. I didn't understand that msg belonged to fail.

simon3z:
@detiber @enoodle IIUC we can tweak `USE_PERSISTENT_STORAGE` but not `METRIC_DURATION` which is critical for us.
Do you think it will be possible to be added?

Also the (soon?) coming `METRIC_RESOLUTION` is also really important: https://github.com/openshift/origin-metrics/issues/101

roles/openshift_metrics/README.md
```diff
@@ -0,0 +1,50 @@
...
+------------
+For NFS storage:
+
+Example Playbook
+----------------
+
+- name: Configure nfs hosts
+  hosts: oo_nfs_to_config
+  roles:
+  - role: openshift_storage_nfs
```
abutcher: Please update README for this role in "Dependencies" and "Example Playbook"

abutcher:
@detiber @jdnieto Once this has been rebased and the README has been updated, I think this will be ready to test/merge.

enoodle:
@simon3z ```METRIC_DURATION``` can be added like ```USE_PERSISTENT_STORAGE```, Unless I missed something. We will probably want to set it from the inventory file.

detiber:
@simon3z @enoodle We can do a followup PR to address those settings.

simon3z:
> @simon3z @enoodle We can do a followup PR to address those settings.

:+1: but let's keep note of that (maybe open an issue?)

jdnieto:
Rebased

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1576,6 +1576,22 @@ def get_defaults(self, roles, deployment_type):
...
+                        size='10Gi'
+                    ),
+                    nfs=dict(
+                        directory='/exports',
+                        options='*(rw,root_squash)'),
+                    host=None,
+                    access_modes=['ReadWriteMany'],
+                    create_pv=True
+                )
+            ),
```
abutcher: Jenkins is reporting indentation issues in this area.

jdnieto:
@abutcher Solved

abutcher:
aos-ci-test

detiber:
@brenton

detiber:
@simon3z issue for tracking the additional features is here: https://github.com/openshift/openshift-ansible/issues/1780

Issue #1314, opened 2016-02-01
DNS out-of-the-box configuration. 


timothysc:
Post deployment our default configuration leaves DNS in a state where pods can not communicate with the outside world.   

When running upstreams networking conformance tests the pods fail to resolve google.com.  In order to rectify, I had to setup dnsmasq on the master.  http://developerblog.redhat.com/2015/11/19/dns-your-openshift-v3-cluster/

This is not spelled out as a separate step post installation, but imho there is really no reason the installer can not properly configure dnsmasq on the master. 

/cc @rrati @jayunit100 @jeremyeder @mattf

detiber:
Trello card associated with this issue: https://trello.com/c/uIvwy5Yj

sdodson:
> Post deployment our default configuration leaves DNS in a state where pods can not communicate with the outside world.

The article you've linked to solves many problems but it shouldn't be directly related to this unless google.com overlaps with your default subdomain, overlaps with your host's search path, or your hosts themselves cannot resolve google.com. Can you post the contents of a pod's `/etc/resolv.conf`? 

Below you can see that the pod has skydns listed as its first resolver and then the host's resolvers are next. SkyDNS returns a SERVFAIL so the resolver moves on to the resolvers which are successful.

```
[root@hello-openshift1 /]# cat /etc/resolv.conf 
nameserver 172.30.0.1
nameserver 192.168.122.1
search default.svc.cluster.local svc.cluster.local cluster.local example.com
options ndots:5

[root@hello-openshift1 /]# dig +nofail google.com            
;; Got SERVFAIL reply from 172.30.0.1, trying next server

; <<>> DiG 9.9.4-RedHat-9.9.4-29.el7_2.2 <<>> +nofail google.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 21539
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 6, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;google.com.                    IN      A

;; ANSWER SECTION:
google.com.             27      IN      A       173.194.68.139
google.com.             27      IN      A       173.194.68.101
google.com.             27      IN      A       173.194.68.102
google.com.             27      IN      A       173.194.68.138
google.com.             27      IN      A       173.194.68.100
google.com.             27      IN      A       173.194.68.113

;; Query time: 0 msec
[...]

PR #1550, opened 2016-03-05
Docker role refactor

detiber:
- refactors the docker role to push generic config into docker role and wrap
  openshift specific variables into an openshift_docker role and it's
  dependent openshift_docker_facts role
- adds support for setting --confirm-def-push flag (Resolves
  https://github.com/openshift/openshift-ansible/issues/1014)
- moves docker related facts from common/node roles to a new docker role
- renames cli_docker_* role varialbes to openshift_docker-* (maintaining
  backward compatibility)
- update role dependencies to pull in openshift_docker conditionally based on
  is_containerized
- remove playbooks/common/openshift-docker since the docker role is now
  conditionally included

PR #1588, opened 2016-03-10
Cluster dns provided by dnsmasq

sdodson:
This pull request installs dnsmasq on each node for two purposes 
 - Provide resolution of resources stored in skydns on the node itself which facilitates docker daemon pushing to the integrated registry via hostname rather than IP address
 - Ensures that the pod's first resolver works for all hostnames while not relying on skydns for recursion

### DIFF
--- a/filter_plugins/openshift_node.py
+++ b/filter_plugins/openshift_node.py
@@ -1,0 +1,43 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# vim: expandtab:tabstop=4:shiftwidth=4
+'''
+Custom filters for use in openshift-node
+'''
+from ansible import errors
+
+class FilterModule(object):
+    ''' Custom ansible filters for use by openshift_node role'''
+
+    @staticmethod
+    def get_dns_ip(openshift_dns_ip, hostvars):
+        ''' Navigates the complicated logic of when to set dnsIP
+
+            In all situations if they've set openshift_dns_ip use that
+            For 1.0/3.0 installs we use the openshift_master_cluster_vip, openshift_node_first_master_ip, else None
+            For 1.1/3.1 installs we use openshift_master_cluster_vip, else None (product will use kube svc ip)
+            For 1.2/3.2+ installs we set to the node's default interface ip
+        '''
+
+        if not issubclass(type(hostvars), dict):
+            raise errors.AnsibleFilterError("|failed expects hostvars is a dict")
+
+        # We always use what they've specified if they've specified a value
+        if openshift_dns_ip != None:
+            return openshift_dns_ip
+
+        if bool(hostvars['openshift']['common']['version_gte_3_2_or_1_2']):
+            return hostvars['ansible_default_ipv4']['address']
+        elif bool(hostvars['openshift']['common']['version_gte_3_1_or_1_1']):
+            if 'openshift_master_cluster_vip' in hostvars:
+                return hostvars['openshift_master_cluster_vip']
+        else:
+            if 'openshift_master_cluster_vip' in hostvars:
+                return hostvars['openshift_master_cluster_vip']
+            elif 'openshift_node_first_master_ip' in hostvars:
+                return hostvars['openshift_node_first_master_ip']
+        return None
+
+    def filters(self):
+        ''' returns a mapping of filters to methods '''
+        return {'get_dns_ip': self.get_dns_ip}
--- a/playbooks/adhoc/uninstall.yml
+++ b/playbooks/adhoc/uninstall.yml
@@ -232,6 +232,9 @@
         - /usr/local/bin/oadm
         - /usr/local/bin/oc
         - /usr/local/bin/kubectl
+        - /etc/NetworkManager/dispatcher.d/99-origin-dns.sh
+        - /etc/dnsmasq.d/origin-dns.conf
+        - /etc/dnsmasq.d/origin-upstream-dns.conf
 
     # Since we are potentially removing the systemd unit files for separated
     # master-api and master-controllers services, so we need to reload the
@@ -244,3 +247,5 @@
   tasks:
     - name: restart docker
       service: name=docker state=restarted
+    - name: restart NetworkManager
+      service: name=NetworkManager state=restarted
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -53,7 +53,6 @@
         console_url: "{{ openshift_master_console_url | default(None) }}"
         console_use_ssl: "{{ openshift_master_console_use_ssl | default(None) }}"
         public_console_url: "{{ openshift_master_public_console_url | default(None) }}"
-        portal_net: "{{ openshift_master_portal_net | default(None) }}"
         ha: "{{ openshift_master_ha | default(groups.oo_masters | length > 1) }}"
         master_count: "{{ openshift_master_count | default(groups.oo_masters | length) }}"
   - openshift_facts:
--- a/roles/openshift_common/README.md
+++ b/roles/openshift_common/README.md
@@ -20,6 +20,7 @@ Role Variables
 | openshift_ip              | UNDEF             | Internal IP address to use for this host    |
 | openshift_public_hostname | UNDEF             | Public hostname to use for this host        |
 | openshift_public_ip       | UNDEF             | Public IP address to use for this host      |
+| openshift_portal_net      | UNDEF             | Service IP CIDR |
 
 Dependencies
 ------------
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -58,10 +58,13 @@ def migrate_docker_facts(facts):
         facts['docker']['hosted_registry_network'] = facts['node'].pop('portal_net')
     return facts
 
+# TODO: We should add a generic migration function that takes source and destination
+# paths and does the right thing rather than one function for common, one for node, etc.
 def migrate_common_facts(facts):
     """ Migrate facts from various roles into common """
     params = {
-        'node': ( 'portal_net' )
+        'node': ('portal_net'),
+        'master': ('portal_net')
     }
     if 'common' not in facts:
         facts['common'] = {}
@@ -72,11 +75,26 @@ def migrate_common_facts(facts):
                     facts['common'][param] = facts[role].pop(param)
     return facts
 
+def migrate_node_facts(facts):
+    """ Migrate facts from various roles into node """
+    params = {
+        'common': ('dns_ip'),
+    }
+    if 'node' not in facts:
+        facts['node'] = {}
+    for role in params.keys():
+        if role in facts:
+            for param in params[role]:
+                if param in facts[role]:
+                    facts['node'][param] = facts[role].pop(param)
+    return facts
+
 def migrate_local_facts(facts):
     """ Apply migrations of local facts """
     migrated_facts = copy.deepcopy(facts)
     migrated_facts = migrate_docker_facts(migrated_facts)
     migrated_facts = migrate_common_facts(migrated_facts)
+    migrated_facts = migrate_node_facts(migrated_facts)
     return migrated_facts
 
 def migrate_hosted_facts(facts):
@@ -462,6 +480,27 @@ def set_metrics_facts_if_unset(facts):
         if 'use_cluster_metrics' not in facts['common']:
             use_cluster_metrics = False
             facts['common']['use_cluster_metrics'] = use_cluster_metrics
+    return facts
+
+def set_dnsmasq_facts_if_unset(facts):
+    """ Set dnsmasq facts if not already present in facts
+    Args:
+        facts (dict) existing facts
+    Returns:
+        facts (dict) updated facts with values set if not previously set
+    """
+
+    if 'common' in facts:
+        if 'use_dnsmasq' not in facts['common'] and facts['common']['version_gte_3_2_or_1_2']:
+            facts['common']['use_dnsmasq'] = True
+        else:
+            facts['common']['use_dnsmasq'] = False
+        if 'master' in facts and 'dns_port' not in facts['master']:
+            if facts['common']['use_dnsmasq']:
+                facts['master']['dns_port'] = 8053
+            else:
+                facts['master']['dns_port'] = 53
+
     return facts
 
 def set_project_cfg_facts_if_unset(facts):
@@ -1507,6 +1546,7 @@ class OpenShiftFacts(object):
         facts = build_controller_args(facts)
         facts = build_api_server_args(facts)
         facts = set_version_facts_if_unset(facts)
+        facts = set_dnsmasq_facts_if_unset(facts)
         facts = set_manageiq_facts_if_unset(facts)
         facts = set_aggregate_facts(facts)
         facts = set_etcd_facts_if_unset(facts)
@@ -1564,7 +1604,7 @@ class OpenShiftFacts(object):
                                       etcd_hosts='', etcd_port='4001',
                                       portal_net='172.30.0.0/16',
                                       embedded_etcd=True, embedded_kube=True,
-                                      embedded_dns=True, dns_port='53',
+                                      embedded_dns=True,
                                       bind_addr='0.0.0.0',
                                       session_max_seconds=3600,
                                       session_name='ssn',
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -58,10 +58,26 @@ def migrate_docker_facts(facts):
         facts['docker']['hosted_registry_network'] = facts['node'].pop('portal_net')
     return facts
 
+def migrate_common_facts(facts):
+    """ Migrate facts from various roles into common """
+    params = {
+        'node': ( 'portal_net' )
+    }
+    if 'common' not in facts:
+        facts['common'] = {}
+    for role in params.keys():
+        if role in facts:
+            for param in params[role]:
+                if param in facts[role]:
+                    facts['common'][param] = facts[role].pop(param)
+    return facts
+
 def migrate_local_facts(facts):
     """ Apply migrations of local facts """
     migrated_facts = copy.deepcopy(facts)
-    return migrate_docker_facts(migrated_facts)
+    migrated_facts = migrate_docker_facts(migrated_facts)
+    migrated_facts = migrate_common_facts(migrated_facts)
+    return migrated_facts
 
 def migrate_hosted_facts(facts):
     """ Apply migrations for master facts """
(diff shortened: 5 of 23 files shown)


roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1303,7 +1304,7 @@ def get_defaults(self, roles):
 
         common = dict(use_openshift_sdn=True, ip=ip_addr, public_ip=ip_addr,
                       deployment_type='origin', hostname=hostname,
-                      public_hostname=hostname)
+                      public_hostname=hostname, portal_net='172.30.0.0/16')
```
jeremyeder: does ansible have a variable to store portal_net or do we have to hard-code this.

detiber: @jeremyeder this is just a default setting that is overriden when the user specifies the user facing variable.

playbooks/adhoc/uninstall.yml
```diff
@@ -236,3 +239,5 @@
   tasks:
     - name: restart docker
       service: name=docker state=restarted
+    - name: restart NetworkManager
+      service: name=NetworkManager state=restarted
```
detiber: Is a restart needed here, or is it possible to reload/reconfigure this through nmcli?

roles/openshift_facts/library/openshift_facts.py
sdodson: With this pr it's `openshift_portal_net` prior to this pr it's `openshift_master_portal_net` assuming people are ok with that change.

playbooks/common/openshift-master/config.yml
```diff
@@ -47,6 +47,7 @@
           public_hostname: "{{ openshift_public_hostname | default(None) }}"
           public_ip: "{{ openshift_public_ip | default(None) }}"
           deployment_type: "{{ openshift_deployment_type }}"
+          portal_net: "{{ openshift_portal_net | default(openshift_master_portal_net) | default(None) }}"
```
detiber: Youl'll want to migrate the previously set fact similar to what I've done in https://github.com/openshift/openshift-ansible/pull/1550: https://github.com/detiber/openshift-ansible/blob/docker_refactor/roles/openshift_facts/library/openshift_facts.py#L31-L58

roles/openshift_node/templates/atomic-openshift-node.j2
```diff
@@ -0,0 +1,23 @@
+# This file templated by openshift-ansible, re-running
```
detiber: I don't see any changes to openshift_node to lay down this template.

sdodson: I'm not sure why this file was in here. I'll remove it.

playbooks/common/openshift-master/config.yml
sdodson: Are we going to merge that shortly, should I simply add things to your maps there if yours merges first?

roles/openshift_facts/library/openshift_facts.py
detiber: That PR should be ready to go, just need to get Brenton back for the merge.

jayunit100:
To clarify the scope of this PR,  skyDNS is working OK.
This is just the cluster dns that works for the outside world and is needed for scenarios like this http://developerblog.redhat.com/2015/11/19/dns-your-openshift-v3-cluster/.

sdodson:
This PR does *not* add wildcard dns as dnsmasq is only available on the
cluster nodes it wouldn't work for external wildcard use.

filter_plugins/openshift_node.py
```diff
@@ -0,0 +1,46 @@
...
+'''
+Custom filters for use in openshift-node
+'''
+from ansible import errors
+
+class FilterModule(object):
+    ''' Custom ansible filters for use by openshift_node role'''
+    
+    @staticmethod
+    def get_dns_ip(openshift_dns_ip, hostvars):
```
sdodson: @abutcher mind checking that this accurately replicates the what was previously in ansible?

abutcher: @sdodson Looks good :+1:

filter_plugins/openshift_node.py
```diff
@@ -0,0 +1,46 @@
...
+        
+        if bool(hostvars['openshift']['common']['version_gte_3_2_or_1_2']):
+            return hostvars['ansible_default_ipv4']['address']
+        elif bool(hostvars['openshift']['common']['version_gte_3_1_or_1_1']):
+            if hostvars['openshift_master_cluster_vip'] != None:
+                return hostvars['openshift_master_cluster_vip']
+            else:
+                return None
+        else:
+            if hostvars['opesnhift_master_cluster_vip'] != None:
```
ibotty: typo: opesnhift

sdodson:
Fixes #1314

filter_plugins/openshift_node.py
sdodson: Thanks!

playbooks/adhoc/uninstall.yml
sdodson: No, unfortunately neither the unit nor nmcli can load dispatcher scripts.

sdodson:
@detiber I believe I'm done and ready for review

detiber:
@sdodson pylint doesn't seem to be unhappy with the openshift_node filter plugin and some of the changes in the openshift_facts module.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -58,12 +58,26 @@ def migrate_docker_facts(facts):
         facts['docker']['hosted_registry_network'] = facts['node'].pop('portal_net')
     return facts
 
+def migrate_common_facts(facts):
+    """ Migrate facts from various roles into common """
+    params = {
+        'node': ( 'portal_net' )
```
detiber: Should this also migrate from openshift.master.portal_net as well?

roles/openshift_node/tasks/main.yml
```diff
@@ -32,6 +25,7 @@
       ovs_image: "{{ osn_ovs_image | default(None) }}"
       proxy_mode: "{{ openshift_node_proxy_mode | default('iptables') }}"
       local_quota_per_fsgroup: "{{ openshift_node_local_quota_per_fsgroup | default(None) }}"
+      dns_ip: "{{ openshift_dns_ip | default(none) | get_dns_ip(hostvars[inventory_hostname])}}"
```
detiber: We might want to migrate this fact in openshift_facts as well.

playbooks/common/openshift-node/config.yml
```diff
@@ -109,6 +109,15 @@
...
     when: hostvars[item].openshift.common.is_containerized | bool and (item in groups.oo_nodes_to_config and item in groups.oo_masters_to_config)
 
+- name: Configure node dnsmasq
+  hosts: oo_nodes_to_config
+  vars:
+    dns_domain: "{{ openshift.common.dns_domain }}"
+    kube_svc_ip: "{{ openshift.common.kube_svc_ip }}"
+  roles:
+    - role: openshift_node_dnsmasq
+      when: openshift_node_dnsmasq | default(True if openshift.common.version_gte_3_2_or_1_2 | bool else False) | bool
```
detiber: Would it make more sense to include this as a dependency of openshfit_node through meta/main.yml instead? That would also eliminate the need to invoke openshift_facts from within the openshift_node_dnsmasq role, since it can be ordered after the openshift_common role has been run.

roles/openshift_node_dnsmasq/tasks/network-manager.yml
```diff
@@ -0,0 +1,23 @@
...
+  
+- name: Restart Network Manager
+  service: 
+    name: NetworkManager
+    state: restarted
+  when: dispatch_script.changed | bool
+
+- name: Test dns resolution
+  command: >
+    getent hosts google.com
```
detiber: Do we want to test against google.com? Won't this cause issues in disconnected environments?

sdodson: Yeah, I can't come up with a decent alternative however. If I were to check for kubernetes.default that'll fail if the node and SDN haven't been started. What do you think about removing it entirely? It was a nice safety net for me while developing but I'm not certain it's necessary long term once QE has finished testing.

detiber: I'd say I'm partial to removing it.

sdodson:
aos-ci-test

roles/openshift_node/tasks/main.yml
sdodson: ack, got it

sdodson:
@detiber rebase and I think I've addressed your comments

sdodson:
aos-ci-test

sdodson:
aos-ci-test

detiber:
aos-ci-test

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -58,10 +58,42 @@ def migrate_docker_facts(facts):
...
+    return facts
+
+def migrate_node_facts(facts):
+    """ Migrate facts from various roles into node """
+    params = {
+        'common': ('dns_ip'),
+    }
+    if 'node' not in facts:
+        facts['node'] = {}
+    for role in params.keys():
```
detiber: I wonder if we need to break out this common logic into a separate method...

sdodson: Yeah I meant to at least put in a TODO to make this more generic. It'd take me longer to do than I'm happy to admit though. OK with a TODO for now?

detiber:
@etsauer mind taking a look and tagging anyone else that may be interested

sdodson:
I'm going to temporarily enable everything on 1.1/3.1 to get a run through jenkins with this functionality enabled.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

etsauer:
cc: @oybed @timothyh

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
@herlo any idea why  aos-ci-jenkins/OS_3.1_NOT_containerized is failing to install some rpms?

```
Error downloading packages:
  tuned-profiles-atomic-openshift-node-3.1.1.6-4.git.32.adf8ec9.el7aos.x86_64: [Errno 256] No more mirrors to try.
  socat-1.7.2.2-5.el7.x86_64: [Errno 256] No more mirrors to try.
  atomic-openshift-node-3.1.1.6-4.git.32.adf8ec9.el7aos.x86_64: [Errno 256] No more mirrors to try.
```

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
Failing because dns_port conflicts, fixed now.
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

detiber:
aos-ci-test

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -30,10 +30,9 @@
       embedded_etcd: "{{ openshift_master_embedded_etcd | default(None) }}"
       embedded_kube: "{{ openshift_master_embedded_kube | default(None) }}"
       embedded_dns: "{{ openshift_master_embedded_dns | default(None) }}"
-      dns_port: "{{ openshift_master_dns_port | default(None) }}"
+      dns_port: "{{ openshift_master_dns_port | default('53' if not openshift.common.version_gte_3_2_or_1_2 else None ) }}"
```
detiber: We probably want to set this is as a default openshift_facts based on the whether dnsmasq is configured rather than using the version fact here.

roles/openshift_node/meta/main.yml
```diff
@@ -15,3 +15,8 @@ dependencies:
 - role: openshift_docker
 - role: openshift_cloud_provider
 - role: openshift_common
+- role: openshift_node_dnsmasq
+  dns_domain: "{{ openshift.common.dns_domain }}"
+  kube_svc_ip: "{{ openshift.common.kube_svc_ip }}"
+  when: openshift_node_dnsmasq | default(True if openshift.common.version_gte_3_2_or_1_2 | bool else False) | bool
```
detiber: There should be a fact value added for `openshift.node.config_dnsmasq` or similar and have the value defaulted in openshift_facts based on the version number rather than computing it here.

This will require pushing the node facts into an openshift_node_facts role, but @abutcher has a PR that could be used as a guide.

roles/openshift_node_dnsmasq/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+dns_domain: 'cluster.local'
+kube_svc_ip: '172.30.0.1'
```
detiber: We probably don't want to use defaults here, otherwise we might end up with a case where a refactoring would end up with these values not being overridden properly.

roles/openshift_node_dnsmasq/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
+---
+- name: Set requisite facts
+  openshift_facts:
+    role: common
+    local_facts:
+      portal_net: "{{ openshift_portal_net | default(openshift_master_portal_net) | default(None) }}"
```
detiber: We should be relying on this being set from openshift_facts (or openshift_common) rather than setting it here.

roles/openshift_node_dnsmasq/tasks/network-manager.yml
```diff
@@ -0,0 +1,13 @@
...
+    src: networkmanager/99-origin-dns.sh
+    dest: /etc/NetworkManager/dispatcher.d/
+    mode: 0755
+  register: dispatch_script
+  
+- name: Restart Network Manager
+  service: 
+    name: NetworkManager
+    state: restarted
+  when: dispatch_script.changed | bool
```
detiber: This should probably be a handler instead

roles/openshift_node/meta/main.yml
sdodson: Since I need to know if we're using dnsmasq in master role too should I just move this to openshift.common.use_dnsmasq ?

detiber: :+1:

roles/openshift_node_dnsmasq/tasks/network-manager.yml
sdodson: Good idea, should be fixed now.

roles/openshift_node_dnsmasq/tasks/main.yml
sdodson: fixed

roles/openshift_common/README.md
```diff
@@ -20,6 +20,7 @@ Role Variables
 | openshift_ip              | UNDEF             | Internal IP address to use for this host    |
 | openshift_public_hostname | UNDEF             | Public hostname to use for this host        |
 | openshift_public_ip       | UNDEF             | Public IP address to use for this host      |
+| openshift_portal_net      | 172.30.0.0/16     | Service IP CIDR |
```
detiber: Default here should now be UNDEF

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1568,7 +1620,7 @@ def get_defaults(self, roles, deployment_type):
                                       etcd_hosts='', etcd_port='4001',
                                       portal_net='172.30.0.0/16',
                                       embedded_etcd=True, embedded_kube=True,
-                                      embedded_dns=True, dns_port='53',
+                                      embedded_dns=True, dns_port='8053',
```
detiber: dns_port should probably be set in `set_dnsmasq_facts_if_unset` setting 8053 only if `use_dnsmasq` is `True`.

roles/openshift_common/README.md
sdodson: Do these README generally refer to the actual variable or the fact that they alter?

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -30,10 +30,10 @@
       embedded_etcd: "{{ openshift_master_embedded_etcd | default(None) }}"
       embedded_kube: "{{ openshift_master_embedded_kube | default(None) }}"
       embedded_dns: "{{ openshift_master_embedded_dns | default(None) }}"
-      dns_port: "{{ openshift_master_dns_port | default(None) }}"
+      # defaults to 8053 when using dnsmasq in 1.2/3.2
+      dns_port: "{{ openshift_master_dns_port | default('53' if not openshift.common.use_dnsmasq else None ) }}"
```
detiber: See my comment on the openshift_facts module, I'd rather see the default handled there. openshift_common is pulled in prior to openshift_master_facts being run (through openshift_cli).

roles/openshift_node_dnsmasq/defaults/main.yml
detiber: Instead of using defaults here `openshift_node_dnsmasq` should just use the `openshift.node.dns_domain` and `openshift.node.kube_svc_ip` values, since it has a dependency on `openshift_common`.

roles/openshift_node_dnsmasq/files/networkmanager/99-origin-dns.sh
```diff
@@ -0,0 +1,55 @@
...
+  def_route_int=$(/sbin/ip route get to ${def_route} | awk '{print $3}')
+  def_route_ip=$(/sbin/ip route get to ${def_route} | awk '{print $5}')
+  if [[ ${DEVICE_IFACE} == ${def_route_int} ]]; then
+    if [ ! -f /etc/dnsmasq.d/origin-dns.conf ]; then
+      cat << EOF > /etc/dnsmasq.d/origin-dns.conf
+strict-order
+no-resolv
+domain-needed
+server=/cluster.local/172.30.0.1
+server=/30.172.in-addr.arpa/172.30.0.1
```
detiber: ^ This should be a template and those values should be adjusted for the user selected values.

detiber: Also, shouldn't the reverse  record here be `1.0.30.172.in-addr.arpa`?

roles/openshift_common/README.md
detiber: I would consider these the actual variable and not the fact altered.

roles/openshift_node_dnsmasq/files/networkmanager/99-origin-dns.sh
sdodson: This file is actually copied directly from https://github.com/openshift/origin/pull/7598 where it should work without any modification for default kube svc cidr and cluster.local dns_domain. That PR hasn't merged yet due to merge flakes unfortunately. We actually do provide the config file that's created in a template that adds some additional config.

I could add a quick hack to configure this for classed subnets, but I'd have to generate them for CIDR ranges that don't map to /8 /16 /24 as that's required for reverse dns. I didn't see reverse dns as a huge win so I just skipped that. Should we add a followup issue for it?

detiber: Yeah, we'll probably have to find a library to use. I'm good with it being a followup issue. My worry is that we might redirect requests to a valid host on the user's network.

sdodson:
Comments addressed.

sdodson:
aos-ci-test

detiber:
lgtm pending tests

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
@alivigni or @herlo any idea why the bot is ignoring me?

sdodson:
aos-ci-test

arilivigni:
@sdodson looks like a github issue

arilivigni:
@sdodson actually could be a token issue. Is it possible someone change the bot api token?

sdodson:
@brenton ready to merge

PR #1589, opened 2016-03-10
openshift-ansible: Wrap boolean facts

smunilla:
Centralize the very weird process for converting strings to booleans
to help keep the code consistent.

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -775,7 +775,7 @@ def set_sdn_facts_if_unset(facts, system_facts):
     if 'common' in facts:
         use_sdn = facts['common']['use_openshift_sdn']
         if not (use_sdn == '' or isinstance(use_sdn, bool)):
-            use_sdn = bool(strtobool(str(use_sdn)))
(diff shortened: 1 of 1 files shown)


roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1032,7 +1032,7 @@ def merge_facts(orig, new, additive_facts_to_overwrite, protected_facts_to_overw
                 # ha (bool) can not change unless it has been passed
                 # as a protected fact to overwrite.
                 if key == 'ha':
-                    if bool(value) != bool(new[key]):
+                    if safe_get_bool(value) != safe_get_bool(new[key]):
```
smunilla: @detiber This was the only other obvious cast to bool I could find.

detiber:
aos-ci-test

detiber:
lgtm

brenton:
It's not obvious to me how this patch is going to fix https://bugzilla.redhat.com/show_bug.cgi?id=1313169.  There's definitely some way `is_containerized` is set to a string and I don't see any places where we're this patch is addressing that.  It's monday morning and I could be overlooking something though.

PR #1594, opened 2016-03-11
First attempt at openshift api secrets module

kwoodson:


roles/lib_openshift_api/library/oc_secrets.py
```diff
@@ -0,0 +1,364 @@
...
+#      delete_after: False
+
+#   # To create a secret:
+#   # This module expects the user to place the files on the remote server and pass them in.
+#  - name: create a secret from content
+#    oca_secrets:
+#      state: present
+#      namespace: default
+#      name: mysecret
+#      content:
```
twiest: We might want to call this "contents" (plural) since it's an array.

twiest:
:+1:

roles/lib_openshift_api/library/oc_secrets.py
```diff
@@ -0,0 +1,360 @@
+#!/usr/bin/env python
+'''
+module for openshift cloud secrets
+'''
+#   Examples:
+#
+#  # to initiate and use /etc/origin/master/admin.kubeconfig file for auth
```
detiber: Is this comment right? Or should it be to list all secrets for the default namespace?

roles/lib_openshift_api/library/oc_secrets.py
```diff
@@ -0,0 +1,360 @@
...
+                fds.write(sfile['content'])
+            files.append(sfile['path'])
+
+        # Register cleanup when module is done
+        atexit.register(OpenShiftOC.cleanup, files)
+        return files
+
+    def update_secret(self, files):
+        '''run update secret '''
+        result = self.delete_secret()
```
detiber: Would it be better to use `oc replace` here instead?

roles/lib_openshift_api/library/oc_secrets.py
```diff
@@ -0,0 +1,360 @@
+#!/usr/bin/env python
+'''
+module for openshift cloud secrets
+'''
+#   Examples:
+#
+#  # to initiate and use /etc/origin/master/admin.kubeconfig file for auth
+#  - name: list secrets
+#    oca_secrets:
```
detiber: s/oca_secrets/oc_secrets/ ?

roles/lib_openshift_api/library/oc_secrets.py
kwoodson: @detiber,  You can specify a namespace for the secrets so I believe this is accurate.

roles/lib_openshift_api/library/oc_secrets.py
kwoodson: Removed oca.  oca is what we use for our admin command. oc + admin.  I have updated it to be oc in the examples.

roles/lib_openshift_api/library/oc_secrets.py
kwoodson: update_secret is extremely clean in this method.  call delete and call secret_new.  2 code paths.  By introducing the replace command this is what it would look like:
-----
        secret = self.prep_secret(files)
        if secret['returncode'] != 0:
            return secret

        sfile_path = '/tmp/%s' % secret['results']['metadata']['name']
        with open(sfile_path, 'w') as sfd:
            sfd.write(json.dumps(secret['results']))

        cmd = ['replace', 'secrets', '-f', sfile_path]

        atexit.register(OpenShiftOC.cleanup, [sfile_path])
        return self.oc_cmd(cmd)
-----

detiber: @kwoodson What is the effect of running the delete/create on an existing secret that is used and referenced in Kubernetes already? oc replace secrets should be an atomic command on the secret object.

kwoodson: @detiber I'm comfortable either way at this point.  This takes a separate code path then the delete and new paths which is one of my concerns.  It also requires a generation of a yaml or json resource.

If we go this route, would we want to '--force' on the update?  I'm thinking we would to avoid any merging issues.  Thoughts?

detiber: @kwoodson I would almost say that we might want to have --force be optional...  I can see this being used semi-regularly for something like updating deployed certificates for the registry, metrics, logging, etc...  In those cases we should be able to update the certs without needing --force, but for other use cases we may very well want to force an update.

kwoodson: Alright, I added --force option as an optional parameter to update.

PR #1597, opened 2016-03-14
Bug 1316761 - Skip the available version check if openshift_image_tag…

brenton:
… is defined.

We already have a check in pre.yml to make sure openshift_image_tag is set to a
range that is allowed.  This is an advanced setting and should be used to
override whatever is returned by the 'latest' image in a given registry.


brenton:
@detiber, @sdodson One of you guys mind taking a look?  My solution to this workaround for registries with bad 'latest' tags is simply not to perform this check.  `openshift_image_tag` would allow them to workaround these registries if needed but this is more of an advanced situation.

brenton:
@detiber, I'm adding another commit that renames the `cli_docker` variables in atomic-openshift-installer to use the new name.

detiber:
lgtm

PR #1599, opened 2016-03-14
Bug 1315564 - Containerized installs require a running environment

brenton:
If the master or node aren't running we can't determine the correct version
that is currently installed.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
@@ -47,6 +47,26 @@
     when: openshift_image_tag is defined and openshift_image_tag.split('v',1).1 | version_compare(target_version ,'<')
 
 - name: Verify upgrade can proceed
+  hosts: oo_masters_to_config
+  tasks:
+  - name: Ensure Master is running
+    service:
+      name: "{{ openshift.common.service_type }}-master"
+      state: started
+      enabled: yes
+    when: openshift.common.is_containerized | bool
+
+- name: Verify upgrade can proceed
+  hosts: oo_nodes_to_config
+  tasks:
+  - name: Ensure Node is running
+    service:
+      name: "{{ openshift.common.service_type }}-node"
+      state: started
+      enabled: yes
+    when: openshift.common.is_containerized | bool
+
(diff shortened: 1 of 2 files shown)


brenton:
@sdodson, @detiber, PTAL

This isn't going to solve QE's problem in the BZ but it should stop things earlier in the process if there's a reason their services won't start.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
```diff
@@ -47,6 +47,26 @@
     when: openshift_image_tag is defined and openshift_image_tag.split('v',1).1 | version_compare(target_version ,'<')
 
 - name: Verify upgrade can proceed
+  hosts: oo_masters_to_config
+  tasks:
+  - name: Ensure Master is running
+    service:
+      name: "{{ openshift.common.service_type }}-master"
```
detiber: Probably need to break this down separately for ha/non-ha installs as well.

brenton: dang... you're right.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
```diff
@@ -47,6 +46,42 @@
...
+  hosts: oo_masters_to_config
+  vars:
+    openshift_master_ha: "{{ groups.oo_masters_to_config | length > 1 }}"
+  tasks:
+  - name: Ensure Master is running
+    service:
+      name: "{{ openshift.common.service_type }}-master"
+      state: started
+      enabled: yes
+    when: not openshift_master_ha and openshift.common.is_containerized | bool
```
brenton: how about this?

detiber: probably want:
```
when: not openshift_master_ha | bool and openshift.common.is_containerized | bool
```

brenton: yeah, I just caught that myself.  updated!

detiber:
:+1:

PR #1600, opened 2016-03-14
OSE/Origin < 3.2/1.2 should not get Docker 1.9

sdodson:
RHEL and Centos docker-1.9 RPMs will have a Conflicts for origin and
atomic-openshift < 3.2. If we left yum to its own devices we'd get a mess when
trying to install these packages with earlier versions of Origin or OSE.
So avoid installing docker-1.9 on Origin/OSE < 1.2/3.2.
See: https://bugzilla.redhat.com/show_bug.cgi?id=1304038

Fedora doesn't keep old packages in repos so we just get the latest there.
Docker 1.10 should be fine if and when it becomes available.

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -1,7 +1,31 @@
 ---
 # tasks file for docker
+
+# Avoid docker 1.9 when installing origin < 1.2 or OSE < 3.2 on RHEL/Centos and
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=1304038
+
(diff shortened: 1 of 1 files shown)


sdodson:
@brenton @detiber mind taking a look?

sdodson:
@adellape relevant to your interests too

detiber:
@sdodson lgtm

detiber:
I didn't catch this on the first review, but we will probably need to refactor this at some point based on my comment here: https://github.com/openshift/openshift-ansible/pull/1656#discussion_r57238058

PR #1601, opened 2016-03-15
NFS Storage changes

detiber:
- tweak the way we are setting openshift.hosted vars and only set them when
  the role is present on a host.
- do not persist empty values from openshift_env
- change openshift_storage_nfs to depend on openshift_facts rather than
  openshift_common (so that we do not pull in the openshift binary).
- better handle the case where the nfs server is not the master server.

Depends on: https://github.com/openshift/openshift-ansible/pull/1603

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -575,14 +575,12 @@ class FilterModule(object):
 
     @staticmethod
     # pylint: disable=too-many-branches
-    def oo_persistent_volumes(hostvars, groups, persistent_volumes=None):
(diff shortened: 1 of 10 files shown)


detiber:
aos-ci-test

detiber:
@abutcher ptal

detiber:
aos-ci-test

abutcher:
LGTM assuming this passes with the dependent change?

abutcher:
@detiber needs a rebase and a new test tag

detiber:
aos-ci-test

detiber:
man, looks like I might have messed up something when breaking the changes out...  will take a closer look tomorrow.

PR #1603, opened 2016-03-15
Move common common facts to openshift_facts

detiber:
- Prevents roles that need common facts from needing to require
  openshift_common, which pulls in the openshift binary.
- Add dependency on openshift_facts to os_firewall, since it uses
  openshift.common facts


detiber:
@abutcher ptal

detiber:
aos-ci-test

abutcher:
:+1: @brenton

PR #1606, opened 2016-03-15
Buildapp monitor

BlueShells:
add 3 item ,build time ,create time ,create if ok,and 2 triggers for the app create

### DIFF
--- a/roles/os_zabbix/vars/template_openshift_master.yml
+++ b/roles/os_zabbix/vars/template_openshift_master.yml
@@ -301,8 +301,23 @@ g_template_openshift_master:
     - 'Openshift Master process not running on {HOST.NAME}'
     priority: avg
 
+  - name: 'Application creation with build has failed on {HOST.NAME}'
+    expression: '{Template Openshift Master:openshift.master.app.build.create.last(#1)}=1 and {Template Openshift Master:openshift.master.app.build.create.last(#2)}=1'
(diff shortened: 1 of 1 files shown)


roles/os_zabbix/vars/template_openshift_master.yml
```diff
@@ -303,6 +328,14 @@ g_template_openshift_master:
     description: The application create loop has failed 4 or more times in the last hour
     priority: avg
 
+  - name: 'Application with build creation has failed multiple times in the last hour on {HOST.NAME}'
+    expression: '{Template Openshift Master:openshift.master.app.build.create.sum(1h)}>3'
```
kwoodson: The cron job runs every 30 minutes.  Therefore this trigger will never fail more than 3 times an hour.
Either change this to '{Template Openshift Master:openshift.master.app.create.sum(2h)}>3'.  The problem with this is that we want to know that it is slow before the 1.5 hour mark.

Or update the cron to run every 10-15 minutes and then we can see if it fails more than 3 times.



 */30 * * * * export ZAGG_CLIENT_HOSTNAME=ops-master-f58e0; ops-runner -f -s 15 -n csca.openshift.master.app.build.create /usr/bin/cron-send-build-app

kwoodson:
Let's give it a try. :+1:

PR #1613, opened 2016-03-15
Only mask etcd service for containerized installls when it's installed

sdodson:
Proposed fix for https://bugzilla.redhat.com/show_bug.cgi?id=1317734
I'm assuming they're installing containerized onto a host that does not have etcd installed.

### DIFF
--- a/roles/etcd/tasks/main.yml
+++ b/roles/etcd/tasks/main.yml
@@ -36,8 +36,12 @@
     state: stopped
     enabled: no
 
+- name: Check for etcd service presence
+  command: systemctl show etcd.service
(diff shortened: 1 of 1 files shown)


sdodson:
@detiber mind taking a look, hopefully QE will test this overnight. I'm making some assumptions that this is what they're running into but I'm not certain. Uncertain if we need the masking or not, so removing it is another possible solution. I just wasn't sure what sort of issues you ran into without masking it.

detiber:
lgtm @brenton

sdodson:
QE marked the bug verified.

PR #1614, opened 2016-03-16
Bug 1317755 - Set insecure-registry for internal registry by default

detiber:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1317755

### DIFF
--- a/roles/openshift_node/tasks/main.yml
+++ b/roles/openshift_node/tasks/main.yml
@@ -23,7 +23,6 @@
       iptables_sync_period: "{{ openshift_node_iptables_sync_period | default(None) }}"
       kubelet_args: "{{ openshift_node_kubelet_args | default(None) }}"
       labels: "{{ lookup('oo_option', 'openshift_node_labels') | default( openshift_node_labels | default(none), true) }}"
-      portal_net: "{{ openshift_master_portal_net | default(None) }}"
       registry_url: "{{ oreg_url | default(none) }}"
       schedulable: "{{ openshift_schedulable | default(openshift_scheduleable) | default(None) }}"
       sdn_mtu: "{{ openshift_node_sdn_mtu | default(None) }}"
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -115,6 +115,11 @@
   vars:
     openshift_node_master_api_url: "{{ hostvars[groups.oo_first_master.0].openshift.master.api_url }}"
     openshift_node_first_master_ip: "{{ hostvars[groups.oo_first_master.0].openshift.common.ip }}"
+    # TODO: configure these based on
+    # hostvars[groups.oo_first_master.0].openshift.hosted.registry instead of
+    # hardcoding
+    openshift_docker_hosted_registry_insecure: True
+    openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.master.portal_net }}"
   roles:
   - openshift_node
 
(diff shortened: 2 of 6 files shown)


detiber:
aos-ci-test

detiber:
Ok, so this is presenting issues for containerized installations currently...  I'm seeing an issue where the image streams are failing to import, and I believe it's related to the way that services are getting restarted. 

One thing I noticed is that with the dependency chains pulling in roles like docker, we have the interesting side-effect of handlers not being kicked off until after the top level role has been processed. 

My initial thought is to not rely on handlers at least for the docker role, but a better option may be to just  adding additional checks on the services following the restart of a dependent service.

Maybe we could have the handler in the role dependency trigger a handler in the dependent role to validate that the service comes back up appropriately before continuing?

@abutcher @sdodson thoughts?

roles/openshift_node/tasks/main.yml
```diff
@@ -23,7 +23,6 @@
       iptables_sync_period: "{{ openshift_node_iptables_sync_period | default(None) }}"
       kubelet_args: "{{ openshift_node_kubelet_args | default(None) }}"
       labels: "{{ lookup('oo_option', 'openshift_node_labels') | default( openshift_node_labels | default(none), true) }}"
-      portal_net: "{{ openshift_master_portal_net | default(None) }}"
```
sdodson: You're removing this because it was really only needed by docker? I need this for the dns work but I can add it back in after this merges.

sdodson:
> One thing I noticed is that with the dependency chains pulling in roles like docker, we have the interesting side-effect of handlers not being kicked off until after the top level role has been processed.

Could you flush them within the docker role?
- meta: flush_handlers

It's not entirely clear to me if this only works in playbooks or within a role.
http://docs.ansible.com/ansible/playbooks_intro.html#handlers-running-operations-on-change

roles/openshift_node/tasks/main.yml
detiber: Is the whole portal_net needed, or just the DNS IP? I'm fine with persisting either though, I just didn't see a need to keep persisting a value on the node that was not needed.

abutcher:
> It's not entirely clear to me if this only works in playbooks or within a role.
http://docs.ansible.com/ansible/playbooks_intro.html#handlers-running-operations-on-change

Looks like either would work.

roles/openshift_node/tasks/main.yml
sdodson: What I really need in that work is the kube_svc_ip which is computed from portal_net. But I guess we could just persist that.

detiber: +1 on just persisting the kube service ip, since it will also provide the default for the dns_ip (which we are currently only persisting if the user overrides it, otherwise we are performing a bit of acrobatics to determine it.

detiber:
`meta: flush_handlers` did the trick, thanks!

@abutcher as we start implementing the delegate_to pattern and nesting these roles a bit more through dependencies, we'll probably need to make sure to flush the handlers at the end of each tasks/main.yml for the dependent roles.


@abutcher @sdodson: as an aside, one thing that I was thinking about, was instead of doing service started/enabled and bypassing the handlers at install time, we *could* instead just enable the service and allow the handlers to kick off a restart...  we may need to add a task to notify the handler if the service is presently stopped, though.

detiber:
aos-ci-test

detiber:
@brenton

sdodson:
Confirmed this fixes it in my own testing. LGTM

PR #1578, opened 2016-03-09
Libvirt provider fixes

detiber:
- Use difference filter instead of reject/equalto
- add ability to override vcpu and memory config for vms
- add a dedicated drive for docker storage

PR #1626, opened 2016-03-16
libvirt cluster variables cleanup

codificat:
An install of a libvirt cluster was failing if the `libvirt_instance_memory_mib` option introduced in #1578 (d4fa9b2) was not specified.

While investigating this I saw that there were several `os_libvirt_*` variables that were apparently redundant: they seemed to be meant to have default values, but they were actually always getting the values of their corresponding `libvirt_*` counterparts. Moreover, these ones were hard-coded with different values than the intended default values for `os_libvirt_*`.

This patch cleans that up by removing the various `os_libvirt_*` and having all the `libvirt_*` in `vars.yaml` configurable.

It's unclear to me what the purpose of these `os_libvirt_*` was, so please let me know if that was the wrong approach.

### DIFF
--- a/README_libvirt.md
+++ b/README_libvirt.md
@@ -10,7 +10,7 @@ This makes `libvirt` useful to develop, test and debug OpenShift and openshift-a
 Install dependencies
 --------------------
 
-1.      Install [ansible](http://www.ansible.com/)
+1.	Install [ansible](http://www.ansible.com/)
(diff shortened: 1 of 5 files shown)


detiber:
lgtm @lhuard1A any objections?

lhuard1A:
@detiber no. lgtm as well.

detiber:
@wshearn

wshearn:
lgtm :+1:

PR #1628, opened 2016-03-16
Add default deployment_type to openshift_facts.

abutcher:
`openshift_docker_deployment` will not be set when not using the playbooks so we should default to using `deployment_type` for compat. ~~If neither are available then set `deployment_type = 'origin'`.~~

@detiber PTAL. We could instead use `deployment_type` throughout playbooks/roles and add a fail task to the openshift_facts role when `deployment_type` is missing.

### DIFF
--- a/roles/openshift_facts/tasks/main.yml
+++ b/roles/openshift_facts/tasks/main.yml
@@ -25,7 +25,8 @@
   openshift_facts:
     role: common
     local_facts:
-      deployment_type: "{{ openshift_deployment_type }}"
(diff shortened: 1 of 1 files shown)


kwoodson:
:+1:

detiber:
@abutcher I would probably prefer to deprecate deployment_type all together and just use openshift_deployment_type as the variable.

abutcher:
@detiber that will take a bit more finesse -- how do you feel about pulling this in to make it work sans defaulting to 'origin'

detiber:
:+1:

PR #1629, opened 2016-03-16
Fix for docker role when variables are undefined.

kwoodson:


abutcher:
LGTM @detiber  PTAL

roles/docker/tasks/main.yml
```diff
@@ -49,16 +49,16 @@
     dest: /etc/sysconfig/docker
     regexp: '^{{ item.reg_conf_var }}=.*$'
     line: "{{ item.reg_conf_var }}='{{ item.reg_fact_val | oo_prepend_strings_in_list(item.reg_flag ~ ' ') | join(' ') }}'"
-  when: item.reg_fact_val is defined and docker_check.stat.isreg
+  when: item.reg_fact_val is defined and item.reg_fact_val != '' and  docker_check.stat.isreg
```
detiber: Can skip the `is defined` check here.

+1 on the change though.

PR #1635, opened 2016-03-18
[WIP] Option to install ntpd

jstuever:
Allows ntpd to be installed, enabled, and started. Gives a choice to fail if ntpd is unable to synchronize clocks. Can be enabled by setting openshift_manage_clock_method to ntpd or auto. Can be configured to fail if unable to synchronize clocks by setting openshift_require_clock_sync=True. Cannot install on container hosts and will automatically skip if is_containierized. Does not currently maintain ntp configuration file. Can easily be extended to do so and to allow for other methods such as chrony.

### DIFF
--- a/playbooks/common/openshift-cluster/additional_config.yml
+++ b/playbooks/common/openshift-cluster/additional_config.yml
@@ -1,3 +1,13 @@
+- name: Configure ntp
+  hosts:
+    - oo_nodes_to_config
+    - oo_etcd_to_config
+  vars:
+    ntp_require_clock_sync: "{{ openshift_require_clock_sync | default(false) | bool }}"
+  roles:
+    - role: ntp
+      when: not openshift.common.is_containerized | bool and openshift_manage_clock_method is defined and
+        ( openshift_manage_clock_method == "ntpd" or openshift_manage_clock_method == "auto" )
 - name: Configure flannel
   hosts: oo_first_master
   vars:
--- a/roles/ntp/tasks/main.yml
+++ b/roles/ntp/tasks/main.yml
@@ -1,0 +1,33 @@
+---
+# tasks file for ntp
+- name: Install ntp
+  action: "{{ ansible_pkg_mgr }} name=ntp state=present"
+
+- name: Enable and (re)start the ntpd service
+  service:
+    name: ntpd
+    enabled: yes
+    state: restarted
+
+- name: Pause to give ntpd a chance to start and sync
+  pause: seconds=5
+  when: (ntp_require_clock_sync is defined) and ntp_require_clock_sync | bool
+
+- name: Restart ntpd again to recalculate sync status.
+  service:
+    name: ntpd
+    state: restarted
+  when: (ntp_require_clock_sync is defined) and ntp_require_clock_sync | bool
+
+- name: Wait for ntp to become synchronized
+  shell: ntpstat
+  ignore_errors: True
+  register: ntp_ntpstat_result
+  until: ntp_ntpstat_result.rc == 0
+  retries: 5
+  delay: 1
+  when: (ntp_require_clock_sync is defined) and ntp_require_clock_sync | bool
+
+- fail:
+    msg: "ntpd failed to synchronize clock."
+  when: (ntp_ntpstat_result is defined) and ntp_ntpstat_result.rc != 0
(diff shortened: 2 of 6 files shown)


playbooks/common/openshift-cluster/additional_config.yml
```diff
@@ -1,3 +1,13 @@
+- name: Configure ntp
+  hosts:
+    - oo_nodes_to_config
+    - oo_etcd_to_config
+  vars:
+    ntp_require_clock_sync: "{{ openshift_require_clock_sync | default(false) | bool }}"
+  roles:
+    - role: ntp
+      when: not openshift.common.is_containerized | bool and openshift_manage_clock_method is defined and
```
detiber: I don't think we'd want to skip ntp configuration on containerized hosts (though we may want to fail if is_containerized is true and is_atomic is true and openshift_manage_clock_method isn't available).

playbooks/common/openshift-cluster/additional_config.yml
```diff
@@ -1,3 +1,13 @@
...
+        ( openshift_manage_clock_method == "ntpd" or openshift_manage_clock_method == "auto" )
```
detiber: I'd rename the openshfit_require_clock_sync and openshift_manage_clock_method variables to be a bit more consistent naming wise. Maybe openshift_clock_sync_required and openshift_clock_sync_method?

@abutcher thoughts on naming for these vars?

playbooks/common/openshift-cluster/additional_config.yml
```diff
@@ -1,3 +1,13 @@
+- name: Configure ntp
+  hosts:
+    - oo_nodes_to_config
+    - oo_etcd_to_config
+  vars:
+    ntp_require_clock_sync: "{{ openshift_require_clock_sync | default(false) | bool }}"
+  roles:
+    - role: ntp
```
detiber: We'll want to sync the clocks prior to the additional_config.yml playbook being run.

One of the patterns that we've started to use lately (to clean up the steps required in playbooks as well as make it easier to consume the roles outside of playbooks) is to configure things like this as a role dependency.

The way we are now pulling in the docker role is one example (and @abutcher is working on implementing it as part of his delegate_to refactoring: https://github.com/openshift/openshift-ansible/pull/1501)

An example of the way it works is as follows:
- openshift_node defines a dependency on the openshift_docker role: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_node/meta/main.yml
- the openshift_docker role only consists of a metadata file which describes  dependency on openshift_docker_facts and docker
- the openshift_docker_facts role persists local facts related to openshift docker configuration and uses set_fact to map the openshift provided variables/fact defaults to variables that the docker role understands.

The main reason for having openshift_docker and openshift_docker_facts is to make sure that we are setting the variables before the docker role is run, since the role dependencies are executed before the dependent role. That type of wrapping wouldn't be needed unless consuming an upstream ntp/chrony role that was meant to be generic.

For ntp config, I'd say we want to add it as a dependency to at least the following roles: openshift_etcd (coming with Andrew's delegate_to changes), openshift_master and openshift_node to start.

One approach that may work is a role called openshift_clock_sync or even just openshift_clock (to align naming with openshift_env set variables) that would conditionally set dependencies on an ntp or the chrony role accordingly (the when clause works in the dependency field of the meta/main.yml file). It might even be worth it to use an upstream ntp role like https://galaxy.ansible.com/geerlingguy/ntp/ or https://gitlab.com/MathieuMD/ansible-role-ntp

roles/ntp/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+  when: (ntp_require_clock_sync is defined) and ntp_require_clock_sync | bool
+
+- name: Wait for ntp to become synchronized
+  shell: ntpstat
+  ignore_errors: True
+  register: ntp_ntpstat_result
+  until: ntp_ntpstat_result.rc == 0
+  retries: 5
+  delay: 1
+  when: (ntp_require_clock_sync is defined) and ntp_require_clock_sync | bool
```
detiber: It might be better to handle this series of tasks using handlers instead of ordered tasks.

handlers are actually capable of notifying other handlers.

playbooks/common/openshift-cluster/additional_config.yml
jstuever: This is extremely helpful... using wrapper roles and dependencies are the pieces of the puzzle I was failing to realize. Crystal clear now.

jstuever:
I'm going to rewrite this and submit again later.

PR #1636, opened 2016-03-18
Adding openshift_efk role

JayKayy:
This role is used to setup the internal aggregated logging solution within openshift. 
When including the role in a playbook the required variables needed to be passed in are:

-   kibana_hostname: name to use to acces the kibana console. (kibana.apps.example.com)
-   master_url: Url of the openshift master (https://master.example.com:84443)'
-   es_cluster_size: The number of elasticsearch instances in the cluster (3)

cc: @detiber @etsauer

### DIFF
--- a/roles/openshift_efk/README.md
+++ b/roles/openshift_efk/README.md
@@ -6,3 +6,4 @@
 #kibana_hostname: kibana.example.com
 #es_cluster_size: 1
 #master_url: https://localhost:8443
+deployer_secret_vars: kibana.crt=/etc/origin/master/ca.crt kibana.key=/etc/origin/master/ca.key ca.crt=/etc/origin/master/ca.crt ca.key=/etc/origin/master/ca.key
--- a/roles/openshift_efk/README.md
+++ b/roles/openshift_efk/README.md
@@ -1,0 +1,8 @@
+
+
+
+ Required vars:
+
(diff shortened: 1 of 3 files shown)


roles/openshift_efk/README.md
```diff
@@ -0,0 +1,9 @@
+
+
+
+ Required vars:
+
+#kibana_hostname: kibana.example.com
+#es_cluster_size: 1
+#master_url: https://localhost:8443
+deployer_secret_vars: kibana.crt=/etc/origin/master/ca.crt kibana.key=/etc/origin/master/ca.key ca.crt=/etc/origin/master/ca.crt ca.key=/etc/origin/master/ca.key
```
etsauer: @JayKayy is deployer_secret_vars required? should be optional.

roles/openshift_efk/tasks/main.yaml
```diff
@@ -0,0 +1,180 @@
+---
+
+  - fail: msg="This roel requires the following vars to be defined: master_url, kibana_hostname, es_cluster_size"
```
etsauer: role is misspelled

roles/openshift_efk/README.md
JayKayy: um its required to provide something in the secret yes. I should just have it default to `noting=/dev/null`

PR #1637, opened 2016-03-18
Yaml edit ansible module

kwoodson:


roles/lib_yaml_editor/library/yedit.py
```diff
@@ -0,0 +1,218 @@
...
+                return Yedit.get_entry(data[key], rest)
+
+        else:
+            return data.get(keys, None)
+
+
+    def write(self):
+        ''' write to file '''
+        with open(self.filename, 'w') as yfd:
+            yfd.write(yaml.dump(self.yaml_dict, default_flow_style=False))
```
twiest: Below it says that I'm able to specify "json" as the file type.

But here it always only writes yaml.

Is that on purpose?

twiest:
:+1:

PR #1638, opened 2016-03-18
Adding openshift_efk role

JayKayy:
Automates setting up the aggregated container logs in Openshift. This automation deploys the EFK(Elasticsearch, Fluentd, Kibana) within an openshift install.

cc: @etsauer @detiber


roles/openshift_efk/README.md
```diff
@@ -0,0 +1,9 @@
+###Required vars:
```
detiber: I would suggest renaming to openshift_logging, and if there are ever alternatives available, then selecting the alternatives can come down to specifying some different variables.

roles/openshift_efk/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+- kibana_hostname: kibana.example.com
+- es_cluster_size: 1
+- master_url: https://localhost:8443
```
detiber: Recommended best practices for role variable naming is to use <role_name>_<var_name> to avoid variable name collisions.

I would suggest the following names instead (assuming a rename of the role to openshift_logging):
```
openshift_logging_hostname
openshift_logging_elasticsearch_cluster_size
openshift_logging_master_public_url
openshift_logging_secret_vars
openshift_logging_fluentd_replicas

File: roles/openshift_efk/tasks/main.yaml [hunk 3]
```diff
@@ -0,0 +1,179 @@
...
+    when: "kibana_hostname is not defined or
+          es_cluster_size is not defined or
+          master_url is not defined"
+
+  - name: "Checking for logging project"
+    command: oc get project logging
+    register: logging_project
+    failed_when: "'FAILED' in logging_project.stderr"
+    tags:
+      - cleanup
```
detiber: I can see where you are going with the cleanup/build tags, however it would cause a lot of pain to consume this role with the tags being configured in this way.

Instead I would suggest the following:
- in tasks/main.yaml
  - test that the required variables exist.
  - add an optional variable for openshift_hosted_logging_cleanup
  - conditionally include a deploy/cleanup task file based on the value of the above variable:
```
- name: Cleanup logging deployment
  include: cleanup.yml
  when: openshift_hosted_logging_cleanup | default(false) | bool

- name: Deploy logging
  include: deploy.yml
  when: not openshift_hosted_logging_cleanup | default(false) | bool
```

This would still give you both behaviors, you would just need to create a playbook to get the same tagging functionality back:
```
- hosts: my_master_host
  tags:
  - build
  roles:
  - role: openshift_hosted_logging

- hosts: my_master_host
  tags:
  - cleanup
  roles:
  - role: openshift_hosted_logging
    openshift_hosted_logging_cleanup: yes
```

roles/openshift_efk/README.md
detiber: The additional benefit to this naming scheme is that we can more easily add the ability to persist these facts with openshift_facts down the line.

roles/openshift_efk/tasks/main.yaml
```diff
@@ -0,0 +1,179 @@
...
+    failed_when: "fluentd2_output.rc == 1 and 'exists' not in fluentd2_output.stderr"
+    tags:
+      - build
+
+  - name: "Make sure to remove stale deployer template"
+    command: oc delete template logging-deployer-template -n openshift
+    register: delete_ouput
+    failed_when: delete_ouput.rc == 1 and 'exists' not in delete_ouput.stderr
+    tags:
+      - build
```
detiber: Should this be part of the cleanup tasks instead of a build task?

roles/openshift_efk/tasks/main.yaml
```diff
@@ -0,0 +1,179 @@
...
+    tags:
+      - cleanup
+
+  - name: "Create logging project"
+    command: oadm new-project logging
+    when: logging_project.rc != 0
+    tags:
+      - build
+  - name: "Changing projects"
+    command: oc project logging
```
detiber: Since this mutates the default user config, I would recommend that instead you should copy the admin kubeconfig to a temporary location and then use that configuration file for the oc/oadm commands instead.

The openshift_persistent_volumes role has a good example of this. temp file creation: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_persistent_volumes/tasks/main.yml#L2-L10 temp file cleanup: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_persistent_volumes/tasks/main.yml#L46-L50

detiber: You will also want to add a role dependency on the openshift_common role, so that you can take advantage of the openshift.common.admin_binary and openshift.common.client_binary variables in case the oc/oadm clients get renamed in the future. See https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_persistent_volumes/tasks/main.yml#L19-L27

roles/openshift_efk/tasks/main.yaml
```diff
@@ -0,0 +1,179 @@
...
+    command: oc create -f /usr/share/openshift/examples/infrastructure-templates/enterprise/logging-deployer.yaml -n openshift
+    register: template_output
+    failed_when: "template_output.rc == 1 and 'exists' not in template_output.stderr"
+    tags:
+      - build
+
+  - name: "Clear out any previous pods"
+    command: oc delete pods --all
+    tags:
+      - build
```
detiber: Seems like this should be part of the cleanup task as well.

roles/openshift_efk/tasks/main.yaml
```diff
@@ -0,0 +1,179 @@
...
+    shell: oc process logging-deployer-template -n openshift -v KIBANA_HOSTNAME={{ kibana_hostname | quote }},ES_CLUSTER_SIZE={{ es_cluster_size | quote }},PUBLIC_MASTER_URL={{ master_url | quote }},IMAGE_PREFIX={{ target_registry | quote }}/  | oc create -f -
+    when: target_registry is defined
+    tags:
+      - build
+
+  - name: "Process the default deployer template"
+    shell: oc process logging-deployer-template -n openshift -v KIBANA_HOSTNAME={{ kibana_hostname | quote }},ES_CLUSTER_SIZE={{ es_cluster_size | quote }},PUBLIC_MASTER_URL={{ master_url | quote }}  | oc create -f -
+    when: target_registry is not defined
+    tags:
+      - build
```
detiber: You can consolidate the two process deployer template tasks as follows:
```
- name: Process the deployer template with an registry other than registry.access.redhat.com
  shell: >
    oc process logging-deployer-template -n openshift -v
    KIBANA_HOSTNAME={{ kibana_hostname | quote }},ES_CLUSTER_SIZE={{ es_cluster_size | quote }},PUBLIC_MASTER_URL={{ master_url | quote }},{{ 'IMAGE_PREFIX=' ~ target_registry | quote if target_registry is defined else '' }}
    | oc create -f -
```

It can be cleaned up a bit further:
```
- name: Process the deployer template with an registry other than registry.access.redhat.com
  shell: oc process logging-deployer-template -n openshift -v {{ oc_process_values }} | oc create -f -
```
in vars/main.yml:
```
kh_kv: KIBANA_HOSTNAME={{ kibana_hostname | quote }}
es_cs_kv: ES_CLUSTER_SIZE={{ es_cluster_size | quote }}
pmu_kv: PUBLIC_MASTER_URL={{ master_url | quote }}
ip_kv: "{{ 'IMAGE_PREFIX=' ~ target_registry | quote if target_registry is defined else '' }}"
oc_process_values: "{{ kh_kv }} {{ es_cs_kv }} {{ pmu_kv }} {{ ip_kv }}"
```

roles/openshift_efk/tasks/main.yaml
```diff
@@ -0,0 +1,179 @@
...
+
+  - name: "Wait for image pull and deployer pod"
+    action: shell oc get pods | grep logging-deployer.*Completed
+    register: result
+    until: result.rc == 0
+    retries: 15
+    delay: 10
+    tags:
+      - build
```
detiber: This is an older style ansible syntax that sadly hasn't been completely eliminated from the docs, the following will work just the same:
```
- name: "Wait for image pull and deployer pod"
   shell: oc get pods | grep logging-deployer.*Completed
   register: result
   until: result.rc == 0
   retries: 15
   delay: 10
```

roles/openshift_efk/tasks/main.yaml
```diff
@@ -0,0 +1,179 @@
...
+
+  - name: "Wait for imagestreams to become available"
+    action: shell oc get is | grep logging-fluentd
+    register: result
+    until: result.rc == 0
+    failed_when: result.rc == 1 and 'not found' not in result.stderr
+    retries: 15
+    delay: 5
+    tags:
+      - build
```
detiber: It's not clear to me why there is a need to wait for the imagestreams to be available here.

detiber: nm my previous comment, it looks like this should actually be moved above "Set insecured registry"

roles/openshift_efk/tasks/main.yaml
```diff
@@ -0,0 +1,179 @@
...
+
+  - name: "Scale fluentd replication controller"
+    command: oc scale rc/logging-fluentd-1 --replicas={{ fluentd_replicas | default('1') }}
+    tags:
+      - build
```
detiber: I haven't played around with logging much, but is there a dc associated with logging-fluentd? if so, you'd want to scale the dc and not the rc.

detiber: nm my previous comment, it appears you are already doing that above, so you shouldn't need to do that here. Instead do you need to wait for the DC to be created above before scaling the DC there?

detiber:
@sosiouxme are some of these steps no longer needed for 1.2 deployments of logging? It looks like the deamonset addition would remove the need to scale fluentd.

sosiouxme:
@detiber the install diverges significantly at the point where we introduce daemonsets and curator. However those innovations aren't included in OSE 3.2, so Origin procedure doesn't really represent OSE procedure at this point - not sure which is targeted here?

detiber:
Why wouldn't daemonsets be supported for 3.2, especially since the 1.2 release hasn't been cut upstream yet.

I understand 3.1 and below since they are already released. Anything we accept here doesn't have to support previously released versions.

Ideally anything accepted here would support both Origin and Enterprise for 1.2/3.2+.

sosiouxme:
@ewolinetz remind me here, was there a reason other than "feature freeze came too soon" for not including the daemonset implementation in our 3.2 image/templates?

ewolinetz:
@sosiouxme correct, that was the reason that we didn't manage to get the daemonset changes in for 3.2.

detiber:
@ewolinetz @sosiouxme Considering there are still fixes coming in for 3.2 bugs, it might not be too late.  Do we currently have a workaround for not having a daemonset in a deployment that uses regions?

sosiouxme:
There are bug fixes still coming in, but this is a substantial feature change that hasn't gone through QE and requires a more complicated upgrade path.

detiber:
@sosiouxme okay, well the main thing that we need is a way to deploy the logging infrastructure for Enterprise 3.2, Online and the latest Origin. I'm not sure the best way to handle that at this point.

Do we need to schedule a working session to discuss the best path forward?

sosiouxme:
@detiber BTW I don't believe there is a problem deploying with regions. fluentd does not have a service and RegionAffinity is based on service selecting the pods.

roles/openshift_efk/tasks/main.yaml
JayKayy: Would you recommend the copied config file over simply adding the `-n <namespace> flag`?

roles/openshift_efk/tasks/main.yaml
JayKayy: After splitting up these into two files, are the tags really even necessary?

JayKayy: Also can you explain to me what this line is actually saying:
`not openshift_hosted_logging_cleanup | default(false) | bool`

Im confused as to where bool coming from?

detiber: @JayKayy I was just indicating a way that there could still be a playbook utilizing the roles that would still allow for tags, it's not something that is needed by any means.

detiber: > Also can you explain to me what this line is actually saying: 
> `not openshift_hosted_logging_cleanup | default(false) | bool`

Yes, first it is taking the value of `openshift_hosted_logging_cleanup` and applying the `default` filter, which will set the value to `false` if `openshift_hosted_logging_cleanup` is not defined. Then it is applying the `bool` filter which treats the resulting value of applying the `default` filter to `openshift_hosted_logging_cleanup` as a yaml/ansible boolean value (yes/no, true/false, 1/0, True/False, TRUE/FALSE, etc). Then finally it negates the result.

roles/openshift_efk/tasks/main.yaml
detiber: Yes, in general you will want to copy directly from the admin config in /etc/origin/master/ (there is an openshift.common variable that I can't recall off the top of my head that provides the path without hardcoding it), and use the copied version. You can still use the `-n` flag as needed, but the `-n` flag using the default kubeconfig does not guarantee that you will be able to perform the operation.

For example, a local user may have done an `oc login` using the default kubeconfig and then you would be attempting to operate as that other user.

JayKayy:
@detiber Hey i took your feedback which was very helpful thank you. I refactored the role based off your suggestions and just completed testing of it. Please review my latests commits. Any additional feedback is appreciated.

detiber:
@abutcher ptal

detiber:
@dak1n1 It's still missing persistent storage, but can you see if this is closer to something that Ops can use?

roles/openshift_hosted_logging/tasks/cleanup_logging.yaml
```diff
@@ -0,0 +1,63 @@
+---
+
+  - name: "Checking for logging project"
+    command: oc get project logging
```
abutcher: We should make use of `{{ openshift.common.client_binary }}` here.

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -0,0 +1,105 @@
...
+    register: mktemp
+    changed_when: False
+
+  - name: Copy the admin client config(s)
+    command: >
+      cp {{ openshift_master_config_dir }}/admin.kubeconfig {{ mktemp.stdout }}/admin.kubeconfig
+    changed_when: False
+
+  - name: "Create logging project"
+    command: oadm new-project logging
```
abutcher: `{{ openshift.common.admin_binary}}`

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -0,0 +1,105 @@
...
+    register: deployer_output
+    failed_when: "deployer_output.rc == 1 and 'exists' not in deployer_output.stderr"
+
+  - name: "Set permissions for logging-deployer service account"
+    command:  "{{ openshift.common.client_binary }} policy add-role-to-user edit system:serviceaccount:logging:logging-deployer"
+    register: permiss_output
+    failed_when: "permiss_output.rc == 1 and 'exists' not in permiss_output.stderr"
+
+  - name: "Set permissions for fluentd"
+    command: oadm policy add-scc-to-user privileged system:serviceaccount:logging:aggregated-logging-fluentd
```
abutcher: `{{ openshift.common.admin_binary}}`

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -0,0 +1,105 @@
...
+    register: fluentd_output
+    failed_when: "fluentd_output.rc == 1 and 'exists' not in fluentd_output.stderr"
+
+  - name: "Set additional permissions for fluentd"
+    command: oadm policy add-cluster-role-to-user cluster-reader system:serviceaccount:logging:aggregated-logging-fluentd
```
abutcher: `{{ openshift.common.admin_binary}}`

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
abutcher: Each `oadm` call could also use the temp copy of the admin client config.

```
{{ openshift.common.admin_binary}} --config={{ mktemp.stdout }}/admin.kubeconfig new-project logging
```

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -0,0 +1,105 @@
...
+  - name: Copy the admin client config(s)
+    command: >
+      cp {{ openshift_master_config_dir }}/admin.kubeconfig {{ mktemp.stdout }}/admin.kubeconfig
+    changed_when: False
+
+  - name: "Create logging project"
+    command: oadm new-project logging
+
+  - name: "Changing projects"
+    command:  "{{ openshift.common.client_binary }} project logging"
```
abutcher: Rather than change projects should we use `-n logging` throughout? Otherwise we'd have to switch back to the previous project?

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
JayKayy: @abutcher Thanks. Good catch. All the `oc` and `oadm` commands should use `--config={{ mktemp.stdout }}/admin.kubeconfig` correct?

abutcher: @JayKayy Yessir

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
abutcher: If we use the temporary config we won't mess with the current namespace so this should be cool.

JayKayy: Thanks for the feedback @abutcher I have made the changes based on your feedback. Let me know if you think i missed something.

dak1n1:
@detiber I think it's close. I asked @wshearn to take a look and he pointed out that it's missing the ability to set the selector. He said we need to be able to limit running ES and Kibana to our infra nodes. I didn't do any work on the logging automation myself (and I've only installed it by hand once), so I don't have any useful input myself. But I'd be willing to try it if I ever get the time.

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
abutcher: Thanks @JayKayy, everything from my feedback lgtm.

detiber:
@brenton We will need to followup still to add PV support and Selector support, but this is a good base to start on.

PR #1639, opened 2016-03-18
Adding openshift_metrics role

JayKayy:
This role automates the container metrics setup within openshift. This will setup the Hawkular metrics within an installation of Openshift so that metrics can be viewed in the openshift web console. See roles/openshift_metrics/README.md for required variables.   

cc: @etsauer @detiber


detiber:
@JayKayy can you coordinate with @jdnieto on his existing PR for this: https://github.com/openshift/openshift-ansible/pull/1582

PR #1644, opened 2016-03-21
Lock down permissions on named certificates

elyscape:
The `named_certificates` directory is created with mode `0755`, and all the files inside it are given mode `0644`, because these are the defaults. SSL private keys, however, are sensitive and should not be world-readable. This changes the permissions of the `named_certificates` directory to `0700` and the private keys to `0600`. It does not change the permissions of the SSL certificates themselves, as those are non-sensitive and provided by the web server anyway.

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -311,13 +311,14 @@
     file:
       path: "{{ named_certs_dir }}"
       state: directory
+      mode: 0700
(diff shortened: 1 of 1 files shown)


sdodson:
Thanks, this looks good.

aos-ci-test

sdodson:
aos-ci-test

abutcher:
@brenton

abutcher:
@brenton PTAL

PR #1645, opened 2016-03-22
Update the container facts before using it

krsacme:
Moved set_container_facts_if_unset before calling set_version_facts_if_unset, which will call get_openshift_version, which in-turn fetches the version based on the is_containarized flag.

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1350,11 +1350,11 @@ class OpenShiftFacts(object):
         facts = set_identity_providers_if_unset(facts)
         facts = set_sdn_facts_if_unset(facts, self.system_facts)
         facts = set_deployment_facts_if_unset(facts)
+        facts = set_container_facts_if_unset(facts)
         facts = set_version_facts_if_unset(facts)
         facts = set_manageiq_facts_if_unset(facts)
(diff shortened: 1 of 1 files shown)


abutcher:
aos-ci-test

sdodson:
@brenton lgtm

brenton:
@krsacme, can you tell me a little more about the problem you were seeing?  This definitely sounds like a good fix but I'm worried something in our upgrade process was relying on version being unset.  We have a number of blocking bugs to work on right now but I'd like to test this PR during an upgrade before merging.

krsacme:
@brenton, thanks for your comments. I am facing issues while deploying a atomic-host as a node in the cluster. I have few bunch of issues related to atomic-host, which I am working on. This particular issue is an obvious mistake and an independent fix, so I have raised it. If you could let me know the upgrade scenario that you are concerned, I could validate and give feedback.

brenton:
@krsacme, I should be able to test the upgrade myself tomorrow and get this merged in.

krsacme:
@brenton I can see this change has been done by you on commit [First pass at systemd unit refactor](https://github.com/openshift/openshift-ansible/commit/1e53057c40fcc6648e4c69d83d0cebc33789a893#diff-8c83287a125220bc46f58c79f26c6670R1353). Thanks for your support.

PR #1646, opened 2016-03-22
[WIP] openshift_clock role with ntp option

jstuever:
- openshift_clock openshift_clock_facts roles  to manage system clock
- Option to choose which role to use to manage the clocks: none, auto, ntpd, chrony.
- Includes upstream ntp role; not available for atomic hosts.
- Allows setting of timezone and option to fail if clock is not synchronized.
- Easily extendable to enable chrony options.
- Disabled by default.


roles/openshift_clock_facts/tasks/main.yaml
```diff
@@ -0,0 +1,40 @@
+---
+- fail:
+    msg: "openshift_clock_method must be set to either 'none', 'auto', 'ntpd', or 'chrony'"
+  when:  openshift_clock_method is defined and
+    openshift_clock_method not in ["none", "auto", "ntpd", "chrony"]
+
+- set_fact:
+    openshift_clock_method: chrony
+  when: openshift_clock_method is defined and openshift_clock_method == "auto" and
+    openshift.common.is_containerized and openshift.common.is_atomic
```
detiber: hmm, I'm wondering if this test is going to be sufficient. We know that when `is_atomic == True` that chrony is the only available option, but `is_containerized` could also mean a RHEL7 host with ntp installed but `containerized=yes` is set in the host inventory.

For now, I would say to just use chrony by default only if `is_atomic == True`

roles/openshift_clock_facts/tasks/main.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+- set_fact:
+    openshift_clock_method: ntpd
+  when: openshift_clock_method is defined and openshift_clock_method == "auto"
```
detiber: This will probably not work as intended. Ansible doesn't allow for re-assigning variables that have already been set. You can only reliably do this if testing that the variable you are setting with set_fact `is not defined`.

Instead, you will probable want to just persist 'auto' to openshift_facts, test that for the invalid conditions, and then set the dependency role variables based on something like this:
```
when: openshift.clock.method is defined and (openshift.clock.method == 'ntpd' or (openshift.clock.method == 'auto'  and default_clock_method == 'ntpd'))
```

detiber: You could also push the logic for determining whether 'auto' should be 'ntpd' or 'chrony' into openshift_facts.

roles/openshift_node/meta/main.yml
```diff
@@ -12,5 +12,6 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
+- role: openshift_clock
```
detiber: We'll probably want to coordinate this with @abutcher's PR here: https://github.com/openshift/openshift-ansible/pull/1501, since depending on which role goes first, the openshift_etcd role he introduces should have a dependency on openshift_clock as well.

inventory/byo/hosts.ose.example
```diff
@@ -271,6 +271,19 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
 # Setting this variable to true will override that check.
 #openshift_override_hostname_check=true
 
+# Specify which method to use to synchronize system clocks.
+# none   - Do not manage clocks with openshift-ansible.
+# auto   - Automatically choose method based on host type [atomic=chrony, default=ntpd].
```
detiber: On thinking about this a bit more...  I'm wondering if there should not be an auto option at all, and openshift_clock_method being undefined should default to host detection in openshift_facts...

You might even want to go as far as testing if ntpd or chrony are already installed and use the pre-installed package if present as the default.

roles/openshift_node/meta/main.yml
abutcher: I expect this will make it in first and I can fix things up in my rebase. Already a few rebases behind anyway.

roles/openshift_clock_facts/tasks/main.yaml
jstuever: I'm a little confused here... this worked 100% of the time in my testing. When I set to auto, atomic hosts saved their openshift_fact as "chrony" while non-atomic hosts save as "ntpd". Are you saying there are one-off situations where this does not work as expected?

detiber: @jstuever I think the behavior is dependent on the version of ansible in use and not guaranteed to stay the same...  Though, they have supposedly documented the expected variable precedence rules with 2.x, I haven't dug too deeply into them since then (and we haven't quite the 2.x support working yet).

PR #1647, opened 2016-03-22
Support setting local storage perFSGroup quota in node config.

dgoodwin:
Adds a new inventory var "openshift_node_local_quota_per_fsgroup", which will
set the new perFSGroup quota in node-config.yaml.

This will leave the property empty if not specified in your inventory, origin interprets this correctly as "do not enable this feature".

To test, your volumeDirectory must be on an XFS filesystem mounted with grpquota option. 

NOTE: On first run for a fresh node, the node will fail to start with latest builds due to an unanticipated bug with doing some checks on the volumeDirectory's filesystem, before it's actually been created. This will be addressed in https://github.com/openshift/origin/pull/8193.

To workaround this just create /var/lib/origin/openshift.local.volumes, and re-run the config playbook.

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1407,6 +1407,7 @@ class OpenShiftFacts(object):
         if 'node' in roles:
             defaults['node'] = dict(labels={}, annotations={},
                                     iptables_sync_period='5s',
+                                    local_quota_per_fsgroup="",
                                     set_node_ip=False)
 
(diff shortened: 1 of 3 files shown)


detiber:
:+1: @brenton

brenton:
aos-ci-test

dgoodwin:
3.0 failing due to stage subscription management server timeout. I'd like a word with the clowns responsible for writing that code. 
dohmonkey

detiber:
aos-ci-test

PR #1650, opened 2016-03-23
Enable Ansible ssh pipelining to speedup deployment

lhuard1A:
https://docs.ansible.com/ansible/intro_configuration.html#pipelining

This requires `sudo` to not `requiretty`. As the `requiretty` option is in the default RHEL `sudoers` file, it has to be explicitly disabled.

This was already the case for [libvirt](https://github.com/openshift/openshift-ansible/blob/master/playbooks/libvirt/openshift-cluster/templates/user-data#L25) and [AWS](https://github.com/openshift/openshift-ansible/blob/master/playbooks/aws/openshift-cluster/templates/user_data.j2#L48). So, just did the same thing for OpenStack.

### DIFF
--- a/bin/cluster
+++ b/bin/cluster
@@ -34,6 +34,8 @@ class Cluster(object):
             os.environ['ANSIBLE_HOST_KEY_CHECKING'] = 'False'
             # TODO: A more secure way to proceed would consist in dynamically
             # retrieving the ssh host public keys from the IaaS interface
+        if 'ANSIBLE_SSH_PIPELINING' not in os.environ:
+            os.environ['ANSIBLE_SSH_PIPELINING'] = 'True'
 
     def get_deployment_type(self, args):
         """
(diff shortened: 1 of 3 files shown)


detiber:
@lhuard1A I'm fine with this change. I will mention that I have seen odd and difficult to troubleshoot behavior when enabling pipelining. You may want to consider a flag to disable it for troubleshooting purposes.

lhuard1A:
Hi @detiber,

It is in fact already possible to disable it by explicitly set the `ANSIBLE_SSH_PIPELINING` environment variable to false.
The advantage is that it is the same environment variable than Ansible itself.

I have just amended the PR to describe in the help of `bin/cluster` which environment variables are overridden by `bin/cluster`.

Do you think the additional piece of documentation in `bin/cluster --help` output is enough or would you like to have an additional `bin/cluster` option?

detiber:
@lhuard1A lgtm

@brenton

PR #1653, opened 2016-03-23
a-o-i: OSE/AEP 3.2 product option

smunilla:
Adds the option to select 3.2 in the quick-installer


detiber:
I could probably find out by reading the code, but do we do anything with the version currently to make sure that the installation results in the proper version being installed?

brenton:
@detiber, we don't have that today but I think we definitely need it.  I don't think it should be part of this PR though.  My thought was that atomic-openshift-installer could set some variables that the playbooks would use to verify the expected version was installed.  We have some of this today.  For RPM installed I'm betting if you set `openshift_pkg_version=-3.2` it would do the right thing.  It would take a little bit of work to make `openshift_image_tag` work the same way.  My vote would be to file an UpcomingRelease bug for this feature.

PR #1654, opened 2016-03-23
Add --gen-inventory command to atomic-openshift-installer.

dgoodwin:
This will just generate the Ansible inventory once all information is gathered,
inform the user where to find it, and exit.

Includes some test utility updates to make what broke as a result of this
change less fragile.

Ping @smunilla


utils/src/ooinstall/cli_installer.py
```diff
@@ -871,7 +872,16 @@ def install(ctx, force):
...
+    # Write quick installer config file to disk:
     oo_cfg.save_to_disk()
+    # Write ansible inventory file to disk:
+    inventory_file = openshift_ansible.generate_inventory(hosts_to_run_on)
+
+    if gen_inventory:
+        click.echo()
+        click.echo("Generated ansible inventory: %s" % inventory_file)
+        click.echo()
+        sys.exit(0)
```
detiber: Any objections to displaying the path to the generated inventory for all deployments and just having the new flag only generate the inventory?

dgoodwin: Sounds good to me. I'll adjust tomorrow morn.

We should probably get it out of .ansible too someday.

detiber: :+1: on moving it out of .ansible

dgoodwin:
Modified to always write out the inventory location, slight text edits to make it consistent with similar line for the quick installer config write, added help for the new gen-inventory option.

Added another commit to move the inventory out of .ansible.

detiber:
aos-ci-test

detiber:
lgtm

brenton:
@dgoodwin, would you mind rebasing this?  For some reason whenever I try to merge it github reports that the merge failed.  It seems stuck in 'Checking for ability to automatically merge…' now.

dgoodwin:
@brenton not sure what happened here, I went to rebase, it looks like the commit is already in master so I think your merge attempt worked.

@andrewklau is the above meant for another PR?

andrewklau:
@dgoodwin hmm I left the comment on this commit https://github.com/openshift/openshift-ansible/commit/67fda43df08a260571eb9b1d7f1ecdab5e323c41 not sure why it ended up here too :/

dgoodwin:
Nice little github bug there.

PR #1655, opened 2016-03-23
Downgrade to docker 1.8.2 if installing < 1.2/3.2

sdodson:
This ignores the situation of Origin by way of only working when version is 3.0 or 3.1. The current version of origin should be fine with docker-1.9.

I've tested
- [x] Installing 3.1.x without docker installed, you get docker-1.8.2
- [x] Installing 3.1.x with docker-1.9 installed, downgrades to 1.8.2
- [x] Upgrading to 3.2 using playbooks/byo/openshift-cluster/upgrades/...
- [x] Installing 3.2 without docker installed, it installs docker-1.9.1 as that's the latest.
- [x] Installing 3.2 with docker-1.8.2 installed, upgrades to 1.9.1 due to RPM deps
- [x] Installing 3.2 with docker-1.9 installed

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -1,32 +1,29 @@
 ---
 # tasks file for docker
 
-# Avoid docker 1.9 when installing origin < 1.2 or OSE < 3.2 on RHEL/Centos and
-# See: https://bugzilla.redhat.com/show_bug.cgi?id=1304038
+- name: Get current installed version if docker_version is specified
+  command: "{{ repoquery_cmd }} --installed --qf '%{version}' docker"
+  when:  not openshift.common.is_atomic | bool and docker_version != ''
+  register: docker_version_result
+  changed_when: false
 
-- name: Default to latest docker for 1.2/3.2 or Fedora
-  set_fact:
-    docker_version: ''
-  when: openshift.common.version_gte_3_2_or_1_2 | bool or ansible_distribution == 'Fedora'
-
-- name: Gather latest version of docker
-  shell: >
-    yum list available -e 0 -q "docker" 2>&1 | tail -n +2 | awk '{ print $2 }' | sort -r | tr '\n' ' ' | tail -n 1
-  register: latest_docker
-  when: not openshift.common.version_gte_3_2_or_1_2 | bool and ansible_distribution != 'Fedora'
-
-- name: Check if Docker 1.9 is the latest
-  set_fact:
-    docker19_is_latest: "{{ True if '1.9' in latest_docker.stdout else False }}"
-  when: not openshift.common.version_gte_3_2_or_1_2 | bool and ansible_distribution != 'Fedora'
-
-- set_fact:
-    docker_version: "{{ '-1.8.2' if docker19_is_latest | bool else ''}}"
-  when: not openshift.common.version_gte_3_2_or_1_2 | bool and ansible_distribution != 'Fedora'
+- name: Downgrade docker if necessary
+  command: "{{ ansible_pkg_mgr }} downgrade -y docker-{{ docker_version }}"
+  register: docker_downgrade_result
+  when: not docker_version_result | skipped and docker_version_result | default('0.0', True) | version_compare(docker_version, 'gt')
 
 - name: Install docker
-  action: "{{ ansible_pkg_mgr }} name=docker{{ docker_version }} state=present"
-  when: not openshift.common.is_atomic | bool
+  action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version != '' else '' }} state=present"
+  when: not openshift.common.is_atomic | bool and not docker_downgrade_result | changed
+
+- stat: path=/etc/sysconfig/docker
+  register: docker_check
+  when: docker_downgrade_result | changed
+
+- name: Remove deferred deletion for downgrades from 1.9
+  command: >
+    sed -i 's/--storage-opt dm.use_deferred_deletion=true//' /etc/sysconfig/docker-storage
+  when: docker_downgrade_result | changed and docker_check.stat.exists | bool and docker_version_result | default('0.0', True) | version_compare('1.9', '>=') and docker_version | version_compare('1.9', '<')
 
 - name: enable and start the docker service
   service:
--- a/roles/openshift_docker_facts/tasks/main.yml
+++ b/roles/openshift_docker_facts/tasks/main.yml
@@ -37,3 +37,19 @@
 - set_fact:
     docker_options: "{{ openshift.docker.options | default(omit) }}"
   when: not openshift.docker.hosted_registry_insecure | default(False) | bool
+
+# Avoid docker 1.9 when installing origin < 1.2 or OSE < 3.2 on RHEL/Centos and
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=1304038
+- name: Gather common package version
+  command: >
+    {{ repoquery_cmd }} --qf '%{version}' "{{ openshift.common.service_type}}"
+  register: common_version
+  failed_when: false
+  changed_when: false
+  when: not openshift.common.is_atomic | bool
+
+- name: Set docker version to be installed
+  set_fact:
+    docker_version: "{{ '1.8.2' }}"
+  when: " ( common_version.stdout | default('0.0', True) | version_compare('3.2','<') and openshift.common.service_type == 'atomic-openshift' ) or
+          ( common_version.stdout | default('0.0', True) | version_compare('1.2','<') and openshift.common.service_type == 'origin' )"
--- a/roles/openshift_master/meta/main.yml
+++ b/roles/openshift_master/meta/main.yml
@@ -12,6 +12,7 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
+- role: openshift_docker
 - role: openshift_cli
 - role: openshift_cloud_provider
 - role: openshift_master_facts
--- a/roles/docker/defaults/main.yml
+++ b/roles/docker/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+docker_version: ''
--- a/roles/openshift_cli/meta/main.yml
+++ b/roles/openshift_cli/meta/main.yml
@@ -13,6 +13,5 @@ galaxy_info:
   - cloud
 dependencies:
 - role: openshift_docker
-  when: openshift.common.is_containerized | bool
 - role: openshift_common
 - role: openshift_cli_facts
--- a/roles/openshift_docker/meta/main.yml
+++ b/roles/openshift_docker/meta/main.yml
@@ -12,5 +12,6 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
+- role: openshift_repos
 - role: openshift_docker_facts
 - role: docker
--- a/roles/openshift_docker_facts/vars/main.yml
+++ b/roles/openshift_docker_facts/vars/main.yml
@@ -1,3 +1,2 @@
 ---
 repoquery_cmd: "{{ 'dnf repoquery --latest-limit 1 -d 0' if ansible_pkg_mgr == 'dnf' else 'repoquery' }}"
-udevw_udevd_dir: /etc/systemd/system/systemd-udevd.service.d
--- a/roles/openshift_facts/tasks/main.yml
+++ b/roles/openshift_facts/tasks/main.yml
@@ -15,7 +15,7 @@
 - set_fact:
     l_is_atomic: "{{ ostree_output.rc == 0 }}"
 - set_fact:
-    l_is_containerized: "{{ l_is_atomic or containerized | default(false) | bool }}"
+    l_is_containerized: "{{ (l_is_atomic | bool) or (containerized | default(false) | bool) }}"
 
 - name: Ensure PyYaml is installed
   action: "{{ ansible_pkg_mgr }} name=PyYAML state=present"


sdodson:
@brenton @detiber Thoughts on if this is safe to do? Currently if they have 1.9 installed and run the installer to install 3.0/3.1 they'll get the following. If they don't have docker installed already we do the right thing because we addressed that in #1600

```
TASK: [openshift_common | Install the base package for versioning] ************
failed: [ose3-master.example.com] => {"changed": true, "rc": 1, "results": ["Loaded plugins: search-disabled-repos\nResolving Dependencies\n--> Running transaction check
\n---> Package atomic-openshift.x86_64 0:3.1.1.6-4.git.27.478f750.el7 will be installed\n--> Processing Dependency: atomic-openshift-clients = 3.1.1.6-4.git.27.478f750.e
l7 for package: atomic-openshift-3.1.1.6-4.git.27.478f750.el7.x86_64\n--> Running transaction check\n---> Package atomic-openshift-clients.x86_64 0:3.1.1.6-4.git.27.478f
750.el7 will be installed\n--> Processing Conflict: docker-1.9.1-19.el7.x86_64 conflicts atomic-openshift < 3.2\n--> Processing Conflict: atomic-openshift-3.1.1.6-4.git.
27.478f750.el7.x86_64 conflicts docker = 1.9.1\n--> Processing Conflict: atomic-openshift-clients-3.1.1.6-4.git.27.478f750.el7.x86_64 conflicts docker = 1.9.1\n--> Finis
hed Dependency Resolution\n You could try using --skip-broken to work around the problem\n You could try running: rpm -Va --nofiles --nodigest\n"]}
msg: Repository 'bob' is missing name in configuration, using id
Repository 'rhel-7-extras-nightly' is missing name in configuration, using id
Error: docker conflicts with atomic-openshift-3.1.1.6-4.git.27.478f750.el7.x86_64
Error: atomic-openshift conflicts with docker-1.9.1-19.el7.x86_64
Error: atomic-openshift-clients conflicts with docker-1.9.1-19.el7.x86_64
```

brenton:
I think QE would be able to test and verify if it's actually safe to do or not.  It's worth sanity checking with Lokesh though.  I'm worried mostly about the advanced user that will be frustrated by this if they somehow really wanted to use 1.9.

detiber:
Doh, just realized I commented on the wrong PR. My comment here: https://github.com/openshift/openshift-ansible/pull/1656#discussion_r57237907 applies

sdodson:
:-) that's fine the other PR is just cherry-picking things atop the latest version we released. I'll check both later tonight, thanks for looking.

sdodson:
@detiber ok, i've moved the version specification from the docker role to openshift_docker_facts. The downgrade when running openshift_common still remains. I couldn't find a way to do that without introducing dependency loops.

detiber:
@sdodson what we probably need to do is to make sure the openshift-master and openshift-node both conditionally require openshift-docker prior to openshift-common, then the dependency loop shouldn't be an issue.

sdodson:
aos-ci-test

sdodson:
@detiber not sure i did what you were suggesting or not but I think this is ready to review

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

roles/openshift_common/meta/main.yml
```diff
@@ -15,4 +15,5 @@ dependencies:
 - { role: os_firewall }
 - { role: openshift_facts }
 - { role: openshift_repos }
+- { role: openshift_docker }
```
detiber: We don't want to make openshift_docker a dependency on openshift_common, otherwise we'll pull it in for things that aren't dockerized yet.

detiber: From latest master:
```
jdetiber@jdetiber:~/git/github/openshift/v3/openshift-ansible git:(imagePolicyConfig⚡)➤ git grep openshift_docker | grep meta
roles/openshift_cli/meta/main.yml:- role: openshift_docker
roles/openshift_docker/meta/main.yml:- role: openshift_docker_facts
roles/openshift_etcd/meta/main.yml:- role: openshift_docker
roles/openshift_node/meta/main.yml:- role: openshift_docker
```

detiber:
There might be a chicken/egg here, but I'm not sure.

sdodson:
Ok, I can clean that up. I have some other work that I need to forward port to this, should have that done after lunch.

detiber:
@arilivigni @brenton should we wait for this PR first, or should we add a workaround in the CI jobs for docker-1.9 in the meantime?

arilivigni:
@detiber what is the workaround?  Will this eventually be resolved?

brenton:
How long of a wait are we talking about?

brenton:
One workaround is just to make sure docker 1.8.2 is already installed before starting a 3.0 or 3.1 installation.

arilivigni:
@brenton I would hate to hack in a change that has to be removed.  When will this be fixed?

sdodson:
I'm working on syncing this with my backport work now. If review of this PR
is good I think we should merge this later today.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -15,6 +15,9 @@
       disable_push_dockerhub: "{{ openshift_disable_push_dockerhub | default(None) }}"
       hosted_registry_insecure: "{{ openshift_docker_hosted_registry_insecure | default(None) }}"
       hosted_registry_network: "{{ openshift_docker_hosted_registry_network | default(None) }}"
+  - role: common
+    local_facts:
+      openshift_deployment_type: "{{ deployment_type }}"
```
detiber: Why would we persist this as openshift.common.openshift_deployment_type? shouldn't it just be openshift.common.deployment_type?  

Also, this should be pushed down into openshift_facts, which is a dependency of openshift_docker_facts.

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +40,46 @@
...
+  register: available_common_version
+  failed_when: false
+  changed_when: false
+
+- name: Gather installed version of docker
+  shell: >
+    rpm -q docker --queryformat '%{version}' | cut -d '.' -f1,2 | grep -v 'not installed'
+  register: installed_docker_version
+  failed_when: false
+  changed_when: false
```
detiber: This is already available as a fact if docker is installed

detiber: openshift.docker.version

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +40,46 @@
...
+  set_fact:
+    docker_version: ''
+    docker_downgrade: False
+  
+- name: Gather installed version of common package
+  shell: >
+    rpm -q {{ openshift.common.service_type }} --queryformat '%{version}' | cut -d '.' -f1,2 | grep -v 'not installed'
+  register: installed_common_version
+  failed_when: false
+  changed_when: false
```
detiber: On a side note, we should probably persist the installed version of OpenShift somewhere (currently we are only doing this in openshift.common.image_tag when is_containerized is true).

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +40,46 @@
...
+  
+- name: Gather available version of common package
+  shell: >
+    yum list available -e 0 -q "{{ openshift.common.service_type}}" 2>&1 | tail -n +2 | awk '{ print $2 }' | sort -r | tr '\n' ' ' | tail -n 1 | cut -d '.' -f1,2
+  register: available_common_version
+  failed_when: false
+  changed_when: false
```
detiber: Does this work on containerized installs?

roles/openshift_docker_facts/tasks/main.yml
sdodson: I think it can be removed. I was having a really hard time in the backbort
branch ensuring that this was set early on but I don't think master suffers
from those issues. Testing without it now.

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +40,46 @@
...
+  register: installed_docker_version
+  failed_when: false
+  changed_when: false
+  
+- name: Gather available version of docker
+  shell: >
+    yum list available -e 0 -q "docker" 2>&1 | tail -n +2 | awk '{ print $2 }' | sort -r | tr '\n' ' ' | tail -n 1
+  register: available_docker_version
+  failed_when: false
+  changed_when: false
```
detiber: probably want to add `when: not is_containerized | bool` here.

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +40,46 @@
...
+
+- set_fact:
+    docker_version: "{{ '1.8.2' }}"
+  when: " installed_common_version.stdout in ['3.0','3.1','1.0','1.1'] or available_common_version.stdout in ['3.0','3.1','1.0','1.1']"
```
detiber: Do we want to set this separately for Origin, so that we only downgrade if origin < 1.4?

detiber: Also, I'm not sure how much we can trust using set_fact to set the same variable twice without ensuring that the conditions are mutually exclusive.

roles/openshift_master/meta/main.yml
```diff
@@ -12,6 +12,7 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
+- role: openshift_docker
```
detiber: openshift_cli pulls in openshift_docker, so this shouldn't be needed.

roles/openshift_storage_nfs/meta/main.yml
```diff
@@ -11,5 +11,4 @@ galaxy_info:
     - 7
 dependencies:
 - { role: os_firewall }
-- { role: openshift_common }
```
detiber: If openshift_storage_nfs relies on facts set by openshift_common this will still be needed (or maybe it just needs openshift_facts)?

roles/openshift_docker_facts/tasks/main.yml
sdodson: The centos docker-1.9 package unfortunately `Conflicts: origin < 1.2` so I think we should leave it as is.

You mean docker_version and docker_downgrade? Earlier I just set them to default values and then only set them otherwise as needed. I may be misunderstanding your question though.

detiber: Yes, I'm not sure if overwriting the default values is something that we can rely on being consistent with Ansible.

roles/openshift_storage_nfs/meta/main.yml
sdodson: sorry, I didn't mean for this to slip in. Fixed

roles/openshift_master/meta/main.yml
sdodson: I moved openshift_cli back to it's original state of only including openshift_docker when openshift.common.is_containerized  do you think I should always include it there instead?

roles/openshift_docker_facts/tasks/main.yml
sdodson: thanks, fixed

sdodson:
aos-ci-test

roles/openshift_master/meta/main.yml
detiber: openshift_master only requires docker if containerized as well.

detiber:
I'm wondering how this would work out for containerized installs...  we can't determine what version is available until we can pull a docker image, and by then we have docker installed, so the openshift_docker role has already run....  but if docker is already installed and we run openshift_facts, then we should be able to get the containerized version... So maybe the downgrade should be handled in openshift_docker instead....

so, openshift_docker_facts would run, followed by docker, then in openshift_docker, we could check openshift.docker.version and the version of openshift available (openshift.common.image_tag for containerized installs) and then do the downgrade if openshift available < 3.2/1.2, docker >= 1.9, and not is_atomic?

sdodson:
Wow that's a great question and I'm surprised it didn't occur to me to test that scenario. Honestly, as long as there's no RPM conflicts I'm not worried about ensuring that we downgrade to docker-1.8.2 for containerized (but not atomic) installs. I say this because if they're running on atomic they have no ability to downgrade docker anyway so we have to ensure that containerized installs work on docker-1.9.1. I could make all of the downgrade codepaths openshift.common.is_containerized instead of openshift.common.is_atomic.

I'm testing a few other things right now, I'll take a look at this particular aspect of it tomorrow morning.

sdodson:
aos-ci-test

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +37,52 @@
...
+  shell: >
+    yum list available -e 0 -q "docker" 2>&1 | tail -n +2 | awk '{ print $2 }' | sort -r | tr '\n' ' ' | tail -n 1
+  register: available_docker_version
+  failed_when: false
+  changed_when: false
+  when: not openshift.common.is_atomic | bool
+  
+- set_fact:
+    docker_version: "{{ '1.8.2' }}"
+  when: " installed_common_version.stdout in ['3.0','3.1','1.0','1.1'] or available_common_version.stdout in ['3.0','3.1','1.0','1.1']
```
detiber: Why not use the version_compare filter here? It would also avoid having to use the cut command

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +37,52 @@
...
+          and not openshift.common.is_atomic | bool"
+  
+- set_fact:
+    docker_downgrade: True
+  when: " docker_version < installed_docker_version.stdout and docker_version != ''
```
detiber: This should be using the version_compare filter

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +37,52 @@
...
+# Avoid docker 1.9 when installing origin < 1.2 or OSE < 3.2 on RHEL/Centos and
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=1304038
+- name: Default to latest docker
+  set_fact:
+    docker_version: ''
+    docker_downgrade: False
+  
+- name: Gather installed version of common package
+  shell: >
+    rpm -q {{ openshift.common.service_type }} --queryformat '%{version}' | cut -d '.' -f1,2 | grep -v 'not installed'
```
detiber: this can be `rpm --quiet -q {{ openshift.common.service_type }} --queryformat '%{version}'` if using the version_compare filter

detiber: Also, if using the command I commented above, you can switch to using the command module instead of the shell module.

detiber: We probably need an alternative dnf command for fedora as well...  @maxamillion any suggestions?

detiber: nevermind on the `--quiet` it removes the output of the version as well

roles/openshift_docker_facts/tasks/main.yml
detiber: I think this can be replaced with `repoquery --nvr {{ openshift.common.service_type }}` with give you the nvr version of the latest available in configured repos and can be used with version_compare to test against rpm -q

roles/openshift_docker_facts/tasks/main.yml
detiber: This command will do the trick though: `repoquery --nvr --installed {{ openshift.common.service_type }}`

roles/openshift_master/meta/main.yml
detiber: This openshift_docker dependency is still not needed.

detiber:
For fedora, I think we can use `dnf repoquery` http://dnf-plugins-core.readthedocs.org/en/latest/repoquery.html

sdodson:
aos-ci-test

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +37,43 @@
...
+  
+- name: Set docker version to be installed
+  set_fact:
+    docker_version: "{{ '1.8.2' }}"
+  when: " ( common_version.stdout | default('0.0') | version_compare('3.2','<') and openshift.common.service_type == 'atomic-openshift' ) or
+          ( common_version.stdout | default('0.0') | version_compare('1.2','<') and openshift.common.service_type == 'origin' )"
+
+# for some reason couldn't use installed_docker_version.stdout directly below
+- set_fact:
+    _installed: installed_docker_version.stdout | default('0.0')
```
detiber: I'd move this up before 'Set docker version to be installed' and add another fact for:
```
l_available: common_version.stdout | default('0.0')
```
Also, change `_installed` to `l_installed` to avoid breaking ansible 2.0

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +37,43 @@
 - set_fact:
     docker_options: "{{ openshift.docker.options | default(omit) }}"
   when: not openshift.docker.hosted_registry_insecure | default(False) | bool
+
+# Avoid docker 1.9 when installing origin < 1.2 or OSE < 3.2 on RHEL/Centos and
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=1304038
+- name: Set docker defaults
+  set_fact:
+    docker_version: ''
+    docker_downgrade: False
```
detiber: Instead of setting docker_downgrade to False by default, you can no set it and then test later on:
```
docker_downgrade is defined
```

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +37,43 @@
...
+  command: >
+    {{ repoquery_cmd }} --installed --qf '%{version}' docker
+  register: installed_docker_version
+  failed_when: false
+  changed_when: false
+  when: not openshift.common.is_atomic | bool
+  
+- name: Set docker version to be installed
+  set_fact:
+    docker_version: "{{ '1.8.2' }}"
```
detiber: I'm still a bit uneasy about setting docker_version above and here without the when clauses being mutually exclusive...  I know it works for 1.9.4, I'm just not sure if it will work for future versions of Ansible

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +37,43 @@
...
     docker_options: "{{ openshift.docker.options | default(omit) }}"
   when: not openshift.docker.hosted_registry_insecure | default(False) | bool
+
+# Avoid docker 1.9 when installing origin < 1.2 or OSE < 3.2 on RHEL/Centos and
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=1304038
+- name: Set docker defaults
+  set_fact:
+    docker_version: ''
+    docker_downgrade: False
+    repoquery_cmd: "{{ 'dnf repoquery --latest-limit 1 -d 0' if ansible_pkg_mgr == 'dnf' else 'repoquery' }}"
```
detiber: repoquery_cmd can be set in vars/main.yml instead of using set_fact.

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -37,3 +37,43 @@
...
+  when: " ( common_version.stdout | default('0.0') | version_compare('3.2','<') and openshift.common.service_type == 'atomic-openshift' ) or
+          ( common_version.stdout | default('0.0') | version_compare('1.2','<') and openshift.common.service_type == 'origin' )"
+
+# for some reason couldn't use installed_docker_version.stdout directly below
+- set_fact:
+    _installed: installed_docker_version.stdout | default('0.0')
+  
+- name: Set docker downgrade fact
+  set_fact:
+    docker_downgrade: True
```
detiber: It might make sense to move docker_downgrade to the docker role based on the version of docker installed and the version requested...

roles/docker/tasks/main.yml
```diff
@@ -1,32 +1,19 @@
 ---
 # tasks file for docker
 
-# Avoid docker 1.9 when installing origin < 1.2 or OSE < 3.2 on RHEL/Centos and
-# See: https://bugzilla.redhat.com/show_bug.cgi?id=1304038
+- name: Downgrade docker if necessary
+  command: >
+    {{ ansible_pkg_mgr }} downgrade -y docker{{ '-' + docker_version if docker_version != '' else '' }}
+  when: not openshift.common.is_atomic and docker_downgrade | bool
```
detiber: Expanding on previous comments, I would say that something like the following may work:

- Add `docker_version = ''` to roles/docker/defaults/main.yml

```
- name: Get current installed version if docker_version is specified
  command: "{{ repoquery_command }}  --installed --qf '%{version}' docker"
  when:  not openshift.common.is_atomic | bool and docker_version != ''
  register: docker_version_result
  changed_when: false

- name: Downgrade docker if necessary
  command: "{{ ansible_pkg_mgr }} downgrade -y docker-{{ docker_version }}"
  register: docker_downgrade_result
  when: not docker_version_result | skipped and docker_version_result.stdout | default('0.0') | version_compare(docker_version, 'gt')

- name: Install docker
  action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version != '' else '' }} state=present"
  when: not openshift.common.is_atomic | bool and not docker_downgrade_result | changed

- name: Remove deferred deletion for downgrades from 1.9
  command: >
    sed -i 's/--storage-opt dm.use_deferred_deletion=true//' /etc/sysconfig/docker-storage
  when: docker_downgrade_result | changed and docker_version_result.stdout | default('0.0') | version_compare('1.9', 'gt') and docker_version | default('9999') | version_compare('1.9', 'lt')
```

roles/openshift_docker_facts/tasks/main.yml
sdodson: Should I move the gathering of installed version there too?

sdodson:
aos-ci-test

detiber:
@sdodson @arilivigni @herlo looks like a bug in one of the playbooks for the CI tests.

herlo:
Yessir, I already have a fix in place.

Sorry for the noise.

arilivigni:
aos-ci-test

roles/docker/tasks/main.yml
sdodson: Ok, this should be done now.

sdodson:
aos-ci-test

detiber:
@sdodson I think I know the issue with the version check...

instead of `| default('0.0')`, it should probably be `| default('0.0', true)`, I didn't think about that earlier, but without adding the second param the default filter only checks if a variable is defined, and the stdout will always be defined for the task if it runs. setting the second param to true causes the default filter to treat the value of the variable as a boolean as well, so that the empty string will evaluate to false and the specified default value will be used.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

detiber:
@herlo @arilivigni What version of docker is installed in these environments? Also, is the rhel7 pulp repo config also configuring the extras repo (which is where we should be pulling in docker from).

herlo:
I just talked with @sdodson on irc about this, and we are not doing extras,
but I'm fixing that up right now. I'm still wanting to watch where we set
these repos so we can do the repo_from_source builds. I'll let you know how
this part works out.

Cheers!

arilivigni:
aos-ci-test

arilivigni:
@sdodson @herlo reverted to use SM until more testing is done with the pulp repo

herlo:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
Fixes #1694

arilivigni:
aos-ci-test

arilivigni:
@sdodson rerunning to see if it gets paused somewhere

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
@detiber final review? The last aos-ci-test passed all green but I squashed a commit and pushed again.

arilivigni:
@sdodson looks like we can add comments through the API pretty easily.  I'll come up with a layout where we update the commit id and the test that ran for it when they are completed

detiber:
lgtm @brenton

PR #1656, opened 2016-03-23
Docker19 conflicts backport

sdodson:
Backports #1655 and #1600 to `openshift-ansible-3.0.47-1`


roles/docker/tasks/main.yml
```diff
@@ -1,7 +1,31 @@
 ---
 # tasks file for docker
+
+# Avoid docker 1.9 when installing origin < 1.2 or OSE < 3.2 on RHEL/Centos and
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=1304038
+
+- name: Default to latest docker for 1.2/3.2 or Fedora
+  set_fact:
+    docker_version: ''
+  when: openshift.common.version_gte_3_2_or_1_2 | bool or ansible_distribution == 'Fedora'
```
detiber: We don't want to expose openshift_facts values to the generic roles. Instead there should be a docker role specific variable that is set in the openshift_docker_facts role.

roles/openshift_common/tasks/main.yml
```diff
@@ -34,6 +34,26 @@
...
+  register: latest_package
+  failed_when: false
+  changed_when: false
+
+- name: Downgrade docker to 1.8.2 if we're installing a version older than 3.2
+  command: >
+    {{ ansible_pkg_mgr }} -y downgrade docker-1.8.2
+  notify: restart docker
+  when: " '1.9' in installed_docker.stdout and latest_package.stdout in ['3.0','3.1'] "
+
```
detiber: I'm not sure we want to handle management of the docker package in openshift_common, instead ensuring the version we expect should probably happen in the docker role, and the openshift_docker_facts role should contain the logic for telling the docker role which version we want installed.

sdodson: The problem I ran into is that openshift_common requires the base package and that comes before we bother with docker role, which is fine if docker hasn't been installed, however if it has been installed already and it's version 1.9 then the base package will fail to install if it's < 3.2.

If openshift_common were to depend on openshift_docker I could teach the docker role to downgrade if necessary.

roles/docker/tasks/main.yml
sdodson: openshift_docker_facts is newer than the point at which I've backported this change to.

detiber: @sdodson doh, What if we left the backport as is and addressed any issues in master instead?

sdodson: +1 lets get master right and then i'll backport as necessary.

sdodson:
Backport worksforme, i've handed this code off to QE to test. See https://bugzilla.redhat.com/show_bug.cgi?id=1320695

sdodson:
These items are in https://github.com/openshift/openshift-ansible/releases/tag/openshift-ansible-3.0.47-6

Issue #1346, opened 2016-02-06
oauthConfig.templates.error - custom error page template


liggitt:
See docs changes at https://github.com/openshift/openshift-docs/pull/1552 and code changes at https://github.com/openshift/origin/pull/7057

Admin can set a custom error page template for rendering errors during the authentication flow in 

```
oauthConfig:
  templates:
    error: templates/error-template.html
```

tbielawa:
@liggitt are you requesting a specific change here?

This issue has been inactive for quite some time. Please update and reopen this issue if this is still a priority you would like to see action on.

liggitt:
because ansible templates the server config, it is unclear to me which parts of the config can be driven through ansible via custom properties. if this can already be set in the config by ansible via custom properties provided by the user, then no change is needed

Issue #1180, opened 2016-01-13
New master config options for identity provider selection page


jwforres:
See docs changes here https://github.com/openshift/openshift-docs/issues/1422  and code changes here https://github.com/openshift/origin/pull/6596

The page isn't super useful yet unless you want to make sure you have a branded page the user sees before being redirected off to some login flow, e.g. if you want to see an "OpenShift" branded page before the user gets taken to GitHub's login page.  Eventually if we support multiple identity providers then the idea is that this page would always appear if the admin configured multiple providers with login support.

jwforres:
Also the related oauth error page, configuration is basically the same just a different key in the config.

detiber:
@dgoodwin can you verify if this is fully supported now?

dgoodwin:
I'm not positive exactly which config options this issue is regarding. Digging through the code PR I see alwaysShowProviderSelection and providerSelection, are there any others?

alwaysShowProviderSelection can be configured via openshift_master_oauth_always_show_provider_selection in inventory. 

I don't think anything was added for providerSelection. Is this required as well @jwforres ?

detiber:
@dgoodwin actually, according to https://github.com/openshift/online/issues/29 it looks like this is fully supported in openshift-ansible now.

Closing.

PR #1660, opened 2016-03-24
Add support for templating master admissionConfig.

dgoodwin:
Adds two new inventory variables for setting sections in admissionConfig.

openshift_master_admission_plugin_order allows configuring the list of
admission controller plugins to enable and what order to run them in. This must
be a JSON formatted list of strings:

openshift_master_admission_plugin_order=["RunOnceDuration", "NamespaceLifecycle", "OriginPodNodeEnvironment", "ClusterResourceOverride", "LimitRanger", "ServiceAccount", "SecurityContextConstraint", "ResourceQuota", "SCCExecRestrictions"]

openshift_master_admission_plugin_config allows setting free-form configuration stanzas that match up with enabled admission controller plugins. This must be a JSON formatted hash:

openshift_master_admission_plugin_config={"RunOnceDuration":{"configuration":{"apiVersion":"v1","kind":"RunOnceDurationConfig","activeDeadlineSecondsOverride":3600}},"ClusterResourceOverride":{"configuration":{"apiVersion":"v1","kind":"ClusterResourceOverrideConfig","limitCPUToMemoryPercent":200,"cpuRequestToLimitPercent":6,"memoryRequestToLimitPercent":60}}}

If neither is specified the template will generate an empty admissionConfig section:

```
kubernetesMasterConfig:
  admissionConfig:
```

If these variables are set output will be:

```
kubernetesMasterConfig:
  admissionConfig:
    pluginOrderOverride:
      - RunOnceDuration
      - NamespaceLifecycle
      - OriginPodNodeEnvironment
      - ClusterResourceOverride
      - LimitRanger
      - ServiceAccount
      - SecurityContextConstraint
      - ResourceQuota
      - SCCExecRestrictions
    pluginConfig:
      ClusterResourceOverride:
        configuration:
          apiVersion: v1
          cpuRequestToLimitPercent: 6
          kind: ClusterResourceOverrideConfig
          limitCPUToMemoryPercent: 200
          memoryRequestToLimitPercent: 60
      RunOnceDuration:
        configuration:
          activeDeadlineSecondsOverride: 3600
          apiVersion: v1
          kind: RunOnceDurationConfig
```

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1118,12 +1118,21 @@ def merge_facts(orig, new, additive_facts_to_overwrite, protected_facts_to_overw
     """
     additive_facts = ['named_certificates']
     protected_facts = ['ha', 'master_count']
+
+    # Facts we do not ever want to merge. These originate in inventory variables
+    # and typically contain JSON dicts. We don't ever want to trigger a merge
+    # here, just completely overwrite with the new if they are present there.
+    overwrite_facts = ['admission_plugin_config',
+                       'kube_admission_plugin_config']
+
     facts = dict()
     for key, value in orig.iteritems():
         # Key exists in both old and new facts.
         if key in new:
+            if key in overwrite_facts:
+                facts[key] = copy.deepcopy(new[key])
             # Continue to recurse if old and new fact is a dictionary.
-            if isinstance(value, dict) and isinstance(new[key], dict):
+            elif isinstance(value, dict) and isinstance(new[key], dict):
                 # Collect the subset of additive facts to overwrite if
                 # key matches. These will be passed to the subsequent
                 # merge_facts call.
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -837,6 +837,25 @@ def set_sdn_facts_if_unset(facts, system_facts):
 
     return facts
 
+def migrate_oauth_template_facts(facts):
+    """
+    Migrate an old oauth template fact to a newer format if it's present.
+
+    The legacy 'oauth_template' fact was just a filename, and assumed you were
+    setting the 'login' template.
+
+    The new pluralized 'oauth_templates' fact is a dict mapping the template
+    name to a filename.
+
+    Simplify the code after this by merging the old fact into the new.
+    """
+    if 'master' in facts and 'oauth_template' in facts['master']:
+        if 'oauth_templates' not in facts['master']:
+            facts['master']['oauth_templates'] = {"login": facts['master']['oauth_template']}
+        elif 'login' not in facts['master']['oauth_templates']:
+            facts['master']['oauth_templates']['login'] = facts['master']['oauth_template']
+    return facts
+
 def format_url(use_ssl, hostname, port, path=''):
     """ Format url based on ssl flag, hostname, port and path
 
@@ -1450,6 +1469,7 @@ class OpenShiftFacts(object):
                             local_facts,
                             additive_facts_to_overwrite,
                             protected_facts_to_overwrite)
+        facts = migrate_oauth_template_facts(facts)
         facts['current_config'] = get_current_config(facts)
         facts = set_url_facts_if_unset(facts)
         facts = set_project_cfg_facts_if_unset(facts)
--- a/roles/openshift_master/templates/master.yaml.v1.j2
+++ b/roles/openshift_master/templates/master.yaml.v1.j2
@@ -1,3 +1,10 @@
+admissionConfig:
+{% if 'admission_plugin_order' in openshift.master %}
+  pluginOrderOverride:{{ openshift.master.admission_plugin_order | to_padded_yaml(level=2) }}
+{% endif %}
+{% if 'admission_plugin_config' in openshift.master %}
+  pluginConfig:{{ openshift.master.admission_plugin_config | to_padded_yaml(level=2) }}
+{% endif %}
 apiLevels:
 {% if not openshift.common.version_gte_3_1_or_1_1 | bool %}
 - v1beta3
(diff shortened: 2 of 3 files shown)


dgoodwin:
@csrwng does the above config generated look ok to you?

csrwng:
@dgoodwin it looks fine to me. Not sure if it makes a difference, but by default when there is no config, the created config includes a ```null```:

```
admissionConfig:
  pluginConfig: null
```

The other thing to note is that ```admissionConfig``` appears in 2 places in the ```master-config.yaml```: 1) at the root of the config which applies to OpenShift resources (projects, build configs, deployment configs, etc.), and 2) under kubernetesMasterConfig which applies to Kubernetes objects (pods, replication controllers, etc.)

So it seems 4 variables would be needed.

dgoodwin:
@csrwng that is unfortunately. Do you know if the example in the online issue is correct for what is needed? https://github.com/openshift/online/issues/62#issuecomment-199369802

@detiber what do you think, should we implement both possible sections even if just one is needed? Things are getting a little messy.

csrwng:
@dgoodwin that example only includes the admissionConfig under kubernetesMasterConfig. However, it does look ok. For an example of the admissionConfig at the root level, take a look at https://github.com/openshift/online/issues/61

dgoodwin:
Thanks Cesar, I will add support for both.

dgoodwin:
PR has been updated, it now supports four new variables:

openshift_master_admission_plugin_order
openshift_master_admission_plugin_config
openshift_master_kube_admission_plugin_order
openshift_master_kube_admission_plugin_config

WARNING: This involves a tricky change to merge_facts, because the _config variables contain a JSON dict, if you attempted to modify any keys in the dict and re-run, you end up with a very incorrect merge of the old + new dicts. 

I don't believe this matches the concept of protected facts in match_facts, as this is not about protecting an old value, rather always overwriting with a new.

Added a new hardcoded set of variables to always overwrite, I've tested this with a variety of changes to the config sections, as well as removing a pre-existing variable entirely and everything appears to be working correctly.

Should be ready for review now.

dgoodwin:
[aos-ci-test]

dgoodwin:
I have added an additional commit for a similar feature, which requires the merge_facts fix. These are both for ops related bugzillas. Details in commit messages.

roles/openshift_master/tasks/main.yml
```diff
@@ -46,7 +46,9 @@
       extension_scripts: "{{ openshift_master_extension_scripts | default(None) }}"
       extension_stylesheets: "{{ openshift_master_extension_stylesheets | default(None) }}"
       extensions: "{{ openshift_master_extensions | default(None) }}"
-      oauth_template: "{{ openshift_master_oauth_template | default(None) }}"
+      oauth_template: "{{ openshift_master_oauth_template | default(None) }}" # deprecated
+      oauth_templates: "{{ openshift_master_oauth_templates | default(None) }}"
```
detiber: Should these be set the same way to cover backwards/forwards compatibility for now?

We should probably mark the version the old options are deprecated in so that we know when they can be officially removed.

detiber:
I believe this now addresses: https://github.com/openshift/openshift-ansible/issues/1180

detiber:
aos-ci-test

dgoodwin:
@detiber updated per the above and our chat on IRC. Tested the oauth with old value, new, old and new, and removing both after the fact.

detiber:
@dgoodwin, we need to keep the old singular value in the config around for backwards compatibility with 3.0 and 3.1.

Instead of doing the full migration, I would just add the value from the singular variable to the values in the plural variable if the login key doesn't exist.

dgoodwin:
Ok updated again, if the legacy fact is there it becomes the new fact if new fact isn't set. If it is, we'll add it in only if "login" is not already in the new fact.

detiber:
aos-ci-test

detiber:
lgtm pending tests

detiber:
This also addresses: https://github.com/openshift/openshift-ansible/issues/1346

dgoodwin:
aos-ci-test

dgoodwin:
@sdodson @brenton I've got a failure here on  aos-ci-jenkins/OS_3.1_containerized, Ari thought you guys might know if it's legit or not? Boils down to a node service not starting, says to check journalctl but those logs are lost unfortunately. Afaict PR has no changes for nodes but I might be missing something.

brenton:
We're still working on containerized node problems.  This PR wouldn't behave any differently for containzerized installs so I vote for merging it once the aos-ci-jenkins/OS_3.1 passes.

dgoodwin:
@brenton I think this is ok to go. I've tested an atomic install, node service starts fine, and indeed there are no changes in here that should affect nodes. 

Ari informed me that the OS_3.1 timeout is not a relevant failure, it's the other builds that we're interested in. (not sure why)

brenton:
@dgoodwin, what timeout are you talking about? The containerized 3.1 install failed for a known reason.  The rpm 3.1 install should pass.

dgoodwin:
In the checks you'll see there's a OS_3.1 check that never completed. Ari
said this is expected though I need to check why.

dgoodwin:
Ari just flipped it to failed so it's not just dangling.

VIA IRC: 

12:12pm [ari(~ari@vpn-63-16.rdu2.redhat.com)] the context changed from when we added the containerized installs so the NOT_containerized, containerized, e2e, and os_unit_tests are the ones that matter

Update for the merge conflicts coming shortly.

dgoodwin:
aos-ci-test

dgoodwin:
Hitting the docker conflict bug in the tests. I have no idea if this should be merged or not now. :)

brenton:
unless this is blocking Ops from making progress this afternoon my vote would be to wait to merge this until @sdodson's PR lands.  He's working on the Docker conflict problem.

dgoodwin:
Sounds good, no immediate issue I'm aware of.

brenton:
aos-ci-test

brenton:
aos-ci-test

brenton:
@arilivigni, how could aos-ci-jenkins/OS_3.0_NOT_containerized fail and aos-ci-jenkins/OS_3.0 pass?  All 3.0 installs on RPM based so I'm not exactly sure how to interpret the failure.

detiber:
Looks like an unrelated issue anyway:
```
failed: [10.8.168.156] => {"changed": false, "failed": true, "rc": 1}
msg: Package at http://download.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm could not be installed
```

arilivigni:
@brenton that was the old context so this didn't run recently  OS_3.0_NOT_containerized is failing because of a playbook issue on our end, but I believe I fixed it.  I had to move things over to pulp since sm is down for maintenance.
You can also track this in the new sidebar you wouldn't see a OS_3.0 there since it is old.  I am thinking the best way to handle this in the future is to find a way to delete any left over statuses from previous runs so then there is no confusion.

arilivigni:
aos-ci-test

arilivigni:
I just Scott's PR pass 1715 so yours should be fine on this run

brenton:
@arilivigni, there's probably something I still don't totally understand.  I had submitted a test run an hour earlier.

arilivigni:
@brenton not sure what you are missing this issue was recent on our end since we switched to pulp ping me on irc and we can discuss it.

PR #1666, opened 2016-03-28
Reusing the systemd templates for the upgrade

brenton:


playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/containerized_upgrade.yml
```diff
@@ -1,8 +1,10 @@
-- name: Update system_units
-  script: ../files/ensure_system_units_have_version.sh {{ openshift.common.service_type }} {{ openshift.common.deployment_type }} {{ g_new_version }}
+- include_vars: ../../../../../roles/openshift_node/vars/main.yml
+
+- name: Update systemd units
+  include: ../../../../../roles/openshift_node/tasks/systemd_units.yml openshift_version=g_aos_versions.avail_version
 
 - name: Verifying the correct version was configured
-  command: grep {{ verify_upgrade_version }} {{ item }}
+  shell: grep {{ verify_upgrade_version }} {{ item }}
```
detiber: Is shell actually needed here?

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
```diff
@@ -16,14 +45,20 @@
 ###############################################################################
 - name: Upgrade master
   hosts: oo_masters_to_config
+  vars:
+    openshift_master_ha: "{{ groups.oo_masters_to_config | length > 1 }}"
```
detiber: @abutcher did we introduce a fact for this? Or am I thinking about a pending change?

detiber:
@brenton I didn't think about breaking those out to a separate task file and including them like that, it keeps it a lot cleaner (and possibly less error prone) thank just calling out to the template module.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/containerized_upgrade.yml
brenton: Yes, I'm pretty sure one of the items I'm passing in uses a glob.

detiber: ah, carry on then.

brenton:
One update I'm going to make is to actually remove ensure_system_units_have_version.sh now that it's no longer needed.

brenton:
@detiber, updated.  I also disabled a too-many-statements pylint check in openshift_facts.  It's definitely true that the lines I added set it over the threshhold but I don't feel like there is any complexity in that function that makes it hard to manage.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
abutcher: `openshift.master.ha` came in from the master scaleup work...

brenton: I'll try this out!

detiber:
@brenton those types of checks in pylint drive me absolutely batty...  oh, you added one too many statements to that method, now you must refactor it into multiple methods...  oh, now you have too many class methods...  <three hours later and the file diff looking like a horror show> ...test passed

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
detiber: chances are you may need to fall back to computing it yourself if it isn't available, but if it is there we should respect it (since it would technically allow for a master to be treated as an HA master).

brenton:
aos-ci-test

roles/openshift_master/tasks/systemd_units.yml
```diff
@@ -0,0 +1,61 @@
+# This file is included both in the openshift_master role and in the upgrade
+# playbooks.  For that reason the ha_svc variables are use set_fact instead of
+# the vars directory on the role.
+
+- debug: var=openshift_version
```
detiber: might want to yank this prior to merge.

roles/openshift_master/tasks/systemd_units.yml
```diff
@@ -0,0 +1,61 @@
...
+
+- name: Init HA Service Info
+  set_fact:
+    ha_svc_svc_dir: "/usr/lib/systemd/system"
```
detiber: Thinking about this a bit, we may want to just go with /etc/systemd/system instead of /usr/lib/systemd/system, since /usr/lib/systemd/system should only contain units related to installed packages.

Of course, that would require a migration for existing installs, but the 3.2 upgrade is probably the best time to do that :)

roles/openshift_master/tasks/systemd_units.yml
brenton: YANKED

roles/openshift_master/tasks/systemd_units.yml
brenton: I feel like there's been some discussion on how we could properly ship all of the unit files in the RPMs.  @sdodson, do you remember anything about that?

I'd like to get this merged today if there aren't any major blockers.

sdodson: I'd like to ship them in the RPMs, I don't see any technical blockers to doing so. I just haven't made time to do the work.

sdodson: Well, as part of that work I was wanting to find a clever way to make 'systemctl start atomic-openshift-master' do the right thing so that starting services is common whether HA or not. Actually the shortest path to this may just be to always run the split services no matter whether we're HA or not.

brenton: Yeah, I'm starting to feel the same way given all the proxy/scaling/unknown challenges we have.

detiber: If we do that, we'll need to make sure to do it for 3.2+ only, and it will require a migration in the 3.1 -> 3.2 upgrade.

brenton:
aos-ci-test

brenton:
@detiber, PTAL at my last commit.  Is this ready for merge if the tests pass?

detiber:
@brenton lgtm

PR #1671, opened 2016-03-28
make docker service want ose containerized services

sjenning:
There are points where the docker service is restarted during
the install.  Sometimes the services that are PartOf docker.service
do not get restarted when docker is restarted.

https://bugzilla.redhat.com/show_bug.cgi?id=1318948

Systemd documentation recommends using "wants" as the perferred
means of creating startup linkages between units.  This
patch makes the ose services wanted by the docker service rather
than multi-user.target.

This creates a downward link from the docker service to the ose
containerized services.

### DIFF
--- a/roles/etcd/templates/etcd.docker.service
+++ b/roles/etcd/templates/etcd.docker.service
@@ -13,4 +13,4 @@ SyslogIdentifier=etcd_container
 Restart=always
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=docker.service
--- a/roles/openshift_master/templates/docker-cluster/atomic-openshift-master-api.service.j2
+++ b/roles/openshift_master/templates/docker-cluster/atomic-openshift-master-api.service.j2
@@ -22,5 +22,5 @@ SyslogIdentifier={{ openshift.common.service_type }}-master-api
 Restart=always
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=docker.service
 WantedBy={{ openshift.common.service_type }}-node.service
--- a/roles/openshift_master/templates/docker-cluster/atomic-openshift-master-controllers.service.j2
+++ b/roles/openshift_master/templates/docker-cluster/atomic-openshift-master-controllers.service.j2
@@ -21,4 +21,4 @@ SyslogIdentifier={{ openshift.common.service_type }}-master-controllers
 Restart=on-failure
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=docker.service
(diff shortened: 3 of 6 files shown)


sdodson:
@arilivigni you try the containerized version of the CI jobs running with this PR? This should fix the long standing issue of containerized installs failing.

I'll test this on my own regardless.

detiber:
aos-ci-test

sdodson:
@brenton LGTM and QE has verified this addresses the bug, lets merge this

sdodson:
@arilivigni you should be good to go on containerized jobs now if you're using master

arilivigni:
@sdodson I ran after this merge and updated and got a cacert failure

brenton:
I saw the same cacert issue.  The master was definitely stopped which was why the curl command from the Node didn't work.

detiber:
gotta love edge cases...  too bad runc isn't available now :)

sdodson:
The only other thing I thought of is we could have 'docker restart' notify 'containerized master restart' 'containerized node restart' 'containerized openvswitch restart' 'containerized etcd restart'  and have those only fire when containerized. 

I'm curious, if in your situation if all services are brought up if `systemctl restart docker` works as expected. As far as I know it always does, but for some reason it fails when ansible does it.

arilivigni:
@sdodson even with your other PR I get the following:

TASK: [openshift_node | Sleep 10 seconds when starting the node failed] ******* 
fatal: [10.8.168.37] => error while evaluating conditional: node_start_result.state != 'started'

sdodson:
Read through this bit of code, I think it may be calling stop then start as separate actions which would probably cause issues. For some reason yesterday when i looked I was under the impression it was just running restart.

https://github.com/ansible/ansible-modules-core/blob/stable-1.9/system/service.py#L843-L904

sdodson:
https://github.com/ansible/ansible-modules-core/issues/1836

brenton:
Could we drop files in /etc/systemd/system/docker.service.d/ that would make 'systemctl stop/start docker' behave the way we want?

detiber:
It actually might be easier to ship a fixed up version of the service module in openshift-ansible to cause restart to actually run systemctl restart

sdodson:
Why not just change the 'restart docker' handler to be command: systemctl
restart docker ? that'd be a quick test in an environment where this
readily reproduces. I can't reproduce the issue fwiw :-(

brenton:
That would work too.  I was trying putting BindsTo=docker.server in the master unit and then the following in cat /etc/systemd/system/docker.service.d/atomic-openshift-master.conf

```
[Unit]
Wants=atomic-openshift-master.service
```

brenton:
It seemed to do the job.  I think it's the BindsTo that would stop the master is someone manually stopped Docker.

PR #1672, opened 2016-03-28
Add openshift_clock role

### LINKED ISSUE #961: validate that timestamps on servers are within reasonable amounts (seconds)
A user reported an issue where the installer happily marched along even with the servers having timestamps that were more than 10 minutes out.

Should we validate that server TZs and times are reasonably in sync?

### LINKED ISSUE #1041: ntp
I saw no references in the manual on setting up clock synchronization.  

Does the installer setup and configure some sane defaults?  If not, it really should.  

e.g. chrony on atomic ntpclient on EL7

/cc @scollier

### LINKED ISSUE #1675: Howto use chrony module
Hello, I want to use the chrony module available in openshift-ansible (https://github.com/openshift/openshift-ansible/tree/master/roles/chrony) but I don't know how to enable this one.
What I already tried is that:
add the line the "chrony_ntp_servers" in my ansible/host file like that:  
nod001.example.com openshift_hostname=nod001.example.com openshift_ip=1.1.1.1 chrony_ntp_servers='ntp1.example.com,ntp2.example.com,ntp3.example.com' openshift_node_labels="{'env': 'test', 'role': 'node'}"

but it seems not working; 
Best regards,

jstuever:
- openshift_clock role to manage system clock
- Can be enabled in inventory: openshift_clock_enabled=true
- Installs ntp when enabled and chrony is not installed
- Starts and enables ntpd or chronyd via: timedatectl set-ntp true
- Enabled by default.

resolves #1675
resolves #1041
resolves #961

### DIFF
--- a/roles/openshift_clock/meta/main.yml
+++ b/roles/openshift_clock/meta/main.yml
@@ -1,0 +1,15 @@
+---
+galaxy_info:
+  author: Jeremiah Stuever
+  description: OpenShift Clock
+  company: Red Hat, Inc.
+  license: Apache License, Version 2.0
+  min_ansible_version: 1.9
+  platforms:
+  - name: EL
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- { role: openshift_facts }
--- a/roles/openshift_clock/tasks/main.yaml
+++ b/roles/openshift_clock/tasks/main.yaml
@@ -1,0 +1,14 @@
+---
+- name: Set clock facts
+  openshift_facts:
+    role: clock
+    local_facts:
+      enabled: "{{ openshift_clock_enabled | default(None) }}"
+
+- name: Install ntp package
+  action: "{{ ansible_pkg_mgr }} name=ntp state=present"
+  when: openshift.clock.enabled | bool and not openshift.clock.chrony_installed | bool
+
+- name: Start and enable ntpd/chronyd
+  shell: timedatectl set-ntp true
+  when: openshift.clock.enabled | bool
--- a/inventory/byo/hosts.aep.example
+++ b/inventory/byo/hosts.aep.example
@@ -408,6 +408,8 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # masterConfig.volumeConfig.dynamicProvisioningEnabled, configurable as of 1.2/3.2, enabled by default
 #openshift_master_dynamic_provisioning_enabled=False
 
+# Configure usage of openshift_clock role.
+#openshift_clock_enabled=true
 
 # host group for masters
 [masters]
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -413,6 +413,9 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # masterConfig.volumeConfig.dynamicProvisioningEnabled, configurable as of 1.2/3.2, enabled by default
 #openshift_master_dynamic_provisioning_enabled=False
 
+# Configure usage of openshift_clock role.
+#openshift_clock_enabled=true
+
 # host group for masters
 [masters]
 ose3-master[1:3]-ansible.test.example.com
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -409,6 +409,9 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # masterConfig.volumeConfig.dynamicProvisioningEnabled, configurable as of 1.2/3.2, enabled by default
 #openshift_master_dynamic_provisioning_enabled=False
 
+# Configure usage of openshift_clock role.
+#openshift_clock_enabled=true
+
 # host group for masters
 [masters]
 ose3-master[1:3]-ansible.test.example.com
--- a/roles/openshift_etcd/meta/main.yml
+++ b/roles/openshift_etcd/meta/main.yml
@@ -13,6 +13,7 @@ galaxy_info:
   - cloud
 dependencies:
 - role: openshift_etcd_facts
+- role: openshift_clock
 - role: openshift_docker
   when: openshift.common.is_containerized | bool
 - role: etcd
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1549,6 +1549,7 @@ class OpenShiftFacts(object):
             OpenShiftFactsUnsupportedRoleError:
     """
     known_roles = ['builddefaults',
+                   'clock',
                    'cloudprovider',
                    'common',
                    'docker',
@@ -1718,6 +1719,16 @@ class OpenShiftFacts(object):
                 docker['api_version'] = version_info['api_version']
                 docker['version'] = version_info['version']
             defaults['docker'] = docker
+
+        if 'clock' in roles:
+            exit_code, _, _ = module.run_command(['rpm', '-q', 'chrony'])
+            if exit_code == 0:
+                chrony_installed = True
+            else:
+                chrony_installed = False
+            defaults['clock'] = dict(
+                enabled=True,
+                chrony_installed=chrony_installed)
 
         if 'cloudprovider' in roles:
             defaults['cloudprovider'] = dict(kind=None)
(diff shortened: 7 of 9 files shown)


jstuever:
New pull request because I messed up my other branch. Refer to #1646 for older comments. Here's a rehash....
@detiber: you asked that I remove the re-assigned variable logic and so I have done this by using a second variable to store the desired auto setting and storing the desired setting into the openshift_fact.

@detiber: you asked that I remove is_containerized from the logic, which I have also done.

@detiber: you questioned if we really want to have an auto option. While I agree with you that we could default to "auto"....  this may impact existing clusters. Specifically, the default timezone is set to "America/Chicago" in the upstream ntp role. As such, existing clusters will all default to this value if the admin does not set their openshift_clock_timezone variable. While we could do a lookup on the system's current settings, this will require additional work. At this time, I propose we default to None an plan to expand so that we can default to auto later.

@abutcher: As Jason mentioned, we'll want to hook this into etcd hosts.

roles/openshift_clock_facts/tasks/main.yaml
```diff
@@ -0,0 +1,41 @@
+---
+- fail:
+    msg: "openshift_clock_method must be set to either 'none', 'auto', 'ntpd', or 'chrony'"
+  when:  openshift_clock_method is defined and
+    openshift_clock_method not in ["none", "auto", "ntpd", "chrony"]
```
detiber: picking nits here, but instead of using fail/when here, these could be validated in the openshift_facts module. See the validate_local_facts method: https://github.com/openshift/openshift-ansible/blob/135f37c77145b08521a921dca81b843591b82c29/roles/openshift_facts/library/openshift_facts.py#L1595

roles/openshift_clock_facts/tasks/main.yaml
```diff
@@ -0,0 +1,41 @@
...
+- name: Set clock facts
+  openshift_facts:
+    role: "{{ item.role }}"
+    local_facts: "{{ item.local_facts }}"
+  with_items:
+  - role: clock
+    local_facts:
+      method: "{{ openshift_clock_auto_method | default(openshift_clock_method) | default(None) }}"
+      require_sync: "{{ openshift_clock_require_sync | default(None) }}"
+      timezone: "{{ openshift_clock_timezone | default(None) }}"
```
detiber: It might be good to set defaults for these fact values in the openshift_facts module, it would allow you to avoid having to do the openshift.clock.method is defined test after this role has been run.

Also, it looks like you can use the `timedatectl` command to query if NTP is enabled, NTP is synchronized and the current configured timezone.

It also looks like timedatectl can be used to configure ntp generically regardless of whether chrony or ntpd is installed: https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/System_Administrators_Guide/chap-Configuring_the_Date_and_Time.html

jstuever:
@detiber: I greatly simplified this based on our conversations regarding timedatectl.

roles/openshift_clock/meta/main.yml
```diff
@@ -0,0 +1,15 @@
...
+  license: Apache License, Version 2.0
+  min_ansible_version: 1.9
+  platforms:
+  - name: EL
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- { role: openshift_clock_facts }
```
detiber: Since there is no longer an upstream role, I think it might make sense to consolidate openshift_clock_facts and openshift_clock.

roles/openshift_clock/tasks/main.yaml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Install ntp package
+  action: "{{ ansible_pkg_mgr }} name=ntp state=present"
+  when: openshift.clock.enabled | bool and not openshift.common.is_atomic
```
detiber: `when: openshift.clock.enabled | bool and not openshift.common.is_atomic | bool`

roles/openshift_clock_facts/tasks/main.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Set clock facts
+  openshift_facts:
+    role: "{{ item.role }}"
+    local_facts: "{{ item.local_facts }}"
+  with_items:
+  - role: clock
+    local_facts:
+      enabled: "{{ openshift_clock_enabled | default(False) }}"
```
detiber: No need for with_items here, I would change this to:
```
- name: Set clock facts
  openshift_facts:
    role: clock
    local_facts:
      enabled: "{{ openshift_clock_enabled | default(False) }}"
```

detiber: Also, it would probably be better to use default(None) here and set the default for openshift.clock.enabled to False in openshift_facts.

jstuever:
@detiber All fair points, and done.

detiber:
@jstuever In the description for this PR you'll want to add the following:
resolves https://github.com/openshift/openshift-ansible/issues/1041
resolves https://github.com/openshift/openshift-ansible/issues/961

You'll also want to add a dependency for openshift_clock in the openshift_etcd role and I think we should probably go ahead and default openshift.clock.enabled to True

detiber:
@abutcher ptal, this is just about ready to merge

detiber:
@jstuever you can add the following to the description as well:
resolves https://github.com/openshift/openshift-ansible/issues/1675

abutcher:
needs rebase but lgtm

detiber:
@brenton

brenton:
aos-ci-test

jstuever:
Not sure if and/or how installing and enabling ntp caused this to fail....

> TASK: [openshift_node | Start and enable node] ******************************** 
> changed: [10.8.168.178]
> failed: [10.8.168.182] => {"failed": true}
> msg: Job for atomic-openshift-node.service failed because the control process exited with error code. See "systemctl status atomic-openshift-node.service" and "journalctl -xe" for details.
> 
> failed: [10.8.168.180] => {"failed": true}
> msg: Job for atomic-openshift-node.service failed because the control process exited with error code. See "systemctl status atomic-openshift-node.service" and "journalctl -xe" for details.

brenton:
@jstuever, it's likely unrelated but if you are able to reproduce and debug it would be a huge help.  Running the tests again just to make sure it's reproduceable.

brenton:
aos-ci-test

jstuever:
@brenton, I may have managed to reproduce the issue. Looks like timedatectl fails to start chronyd.service or ntpd.service if both are installed at the same time. This caused a node to refuse the master's not-yet-valid certificate. I'll need to modify to not install ntp if chrony is already installed.

jstuever:
@detiber ptal

jstuever:
@detiber bump

abutcher:
aos-ci-test

detiber:
@abutcher ptal, lgtm from a quick review.

abutcher:
@detiber LGTM

detiber:
@jstuever did you validate applying this role on a host that already has ntp or chrony configured outside of timedatectl?

jstuever:
@detiber I tested the following combination (below) for 1) date = current, and 2) date = days behind. This covers all scenarios except when chrony is previously enabled (which is the desired state). Furthermore, subsequent runs do not produce any errors. By deduction, this should cover all scenarios. I encountered two known issues. A)when NTP/Chrony was running on a system with a good sync, and then the date/time is changed to a bad value. In my testing, datetimectl is not able to recognize that the service has been stopped, or that the datetime is no longer in sync. This situation should not exist under normal conditions, I always had to fabricate it. B) If the date is set to the future, yum has trouble installing ntp. Again, this is not a realistic scenario.

Type: Prior state => end state
RHEL7: Neither NTP nor Chrony installed => NTP enabled
RHEL7: NTP installed / disabled => NTP enabled
RHEL7: NTP installed / enabled => NTP enabled
RHEL7: Chrony installed / disabled => Chrony enabled
Atomic: Chrony installed / disabled => Chrony enabled

detiber:
ok to test

detiber:
@jstuever pylint errors in default tests:
```
C:1726, 0: Exactly one space required around assignment
                chrony_installed=True
                                ^ (bad-whitespace)
C:1728, 0: Exactly one space required around assignment
                chrony_installed=False
                                ^ (bad-whitespace)
```

jstuever:
@detiber resolved.

detiber:
aos-ci-test

PR #1673, opened 2016-03-29
fix the key name for the dynamic item of avalable

BlueShells:
fix the key name for the dynamic item of avalable


mwoodson:
:+1:

PR #1674, opened 2016-03-29
Bug 1320829 - Ensure docker installed for facts

detiber:
- gather facts requiring docker only if docker is present and running
- Update reference to etcd role in playbooks/common/openshift-etcd/config.yml
  to use openshift_etcd


detiber:
@abutcher ptal
@brenton this change might have unintended consequences, we may want to get QE to do some regression testing prior to merging.

detiber:
aos-ci-test

brenton:
aos-ci-test

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1214,8 +1261,10 @@ def set_container_facts_if_unset(facts):
     if safe_get_bool(facts['common']['is_containerized']):
         facts['common']['admin_binary'] = '/usr/local/bin/oadm'
         facts['common']['client_binary'] = '/usr/local/bin/oc'
-        base_version = get_openshift_version(facts, cli_image).split('-')[0]
-        facts['common']['image_tag'] = "v" + base_version
+        openshift_version = get_openshift_version(facts, cli_image)
+        if openshift_version is not None:
+            base_version = get_openshift_version(facts, cli_image).split('-')[0]
```
brenton: probably better to reuse openshift_version here instead of calling get_openshift_version twice.

brenton:
What part of this PR actually makes sure Docker is installed in a minimal RHEL environment?

brenton:
The 3.1 tests failed:

```
TASK: [openshift_common | Set common Cluster facts] *************************** 
failed: [10.8.168.24] => {"failed": true, "parsed": false}
  File "/root/.ansible/tmp/ansible-tmp-1459253622.65-135016735101257/openshift_facts", line 6879, in <module>
    main()
  File "/root/.ansible/tmp/ansible-tmp-1459253622.65-135016735101257/openshift_facts", line 1746, in main
    protected_facts_to_overwrite)
  File "/root/.ansible/tmp/ansible-tmp-1459253622.65-135016735101257/openshift_facts", line 1352, in __init__
    protected_facts_to_overwrite)
  File "/root/.ansible/tmp/ansible-tmp-1459253622.65-135016735101257/openshift_facts", line 1383, in generate_facts
    defaults = self.get_defaults(roles, deployment_type)
  File "/root/.ansible/tmp/ansible-tmp-1459253622.65-135016735101257/openshift_facts", line 1464, in get_defaults
    version_info = get_docker_version_info()
  File "/root/.ansible/tmp/ansible-tmp-1459253622.65-135016735101257/openshift_facts", line 960, in get_docker_version_info
```

detiber:
> What part of this PR actually makes sure Docker is installed in a minimal RHEL environment?

Basically, it is the openshift_docker role dependency that ensures that docker is installed. The issue we're running into is that openshift_facts is trying to run docker before the package has been installed, so this is mainly attempting to shortcut the use of docker in openshift_facts.

brenton:
OK, so these changes just make it so certain facts won't be gathered yet if docker isn't installed.  They would simply be set the next time the facts are retrieved, right?

detiber:
@brenton that is the goal, yes. I still have some more testing to do to verify that is the case though.

detiber:
aos-ci-test

sdodson:
I think dbus-python is required by minimal install, I don't have a clean minimal install right now to verify. But should we add it to openshift-ansible.spec and/or as an openshift_facts task anyway? I've confirmed it is installed on atomic so should be good to go there.

detiber:
+1 for adding a dependency in the spec.

detiber:
aos-ci-test

brenton:
@arilivigni, containerized installs we not supported for 3.0 so we can remove the  aos-ci-jenkins/OS_3.0${CONTAINERIZED}  job.  Is that failure blocking the other jobs?

arilivigni:
It shouldn't be running for 3.0.  This is why I was hesitant to turn this
on.   I am going to revert the change to CI when I get in and you guys can
run containerized installs and trouble shoot.   Is that ok?

-== @ri ==-
On Mar 31, 2016 8:10 AM, "Brenton Leanhardt" <notifications@github.com>
wrote:

@arilivigni <https://github.com/arilivigni>, containerized installs we not
supported for 3.0 so we can remove the
aos-ci-jenkins/OS_3.0${CONTAINERIZED} job. Is that failure blocking the
other jobs?

—
Reply to this email directly or view it on GitHub
<https://github.com/openshift/openshift-ansible/pull/1674#issuecomment-203902670>

brenton:
@arilivigni, I'm fine with whatever you need to do to make sure the 3.0 and 3.1 RPM installs run.  It doesn't bother me at all that the 3.1 containerized jobs are failing because we're still fixing problems.  As soon as you think you can get the RPM installs running again I would recommend reenabling the containerized jobs in whatever state they are in.

arilivigni:
It will have to wait until I get in.   Have you fibre to the jenkins
instance and the jobs themselves?

-== @ri ==-

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
@brenton this should be ready to go now.

arilivigni:
@brenton so containerized install is fixed?

brenton:
@arilivigni, I don't think so.  The containerized install definitely flakes out most of the time.  I've seen it passed a couple times at least though.  I'm hoping to continue debugging it today.

PR #1678, opened 2016-03-29
Retry starting the node once if it failed the first time

sdodson:
We should also wait for master api regardless of whether or not we're containerized.

### DIFF
--- a/roles/openshift_node/tasks/main.yml
+++ b/roles/openshift_node/tasks/main.yml
@@ -133,11 +133,20 @@
   retries: 120
   delay: 1
   changed_when: false
-  when: openshift.common.is_containerized | bool
 
 - name: Start and enable node
   service: name={{ openshift.common.service_type }}-node enabled=yes state=started
   register: node_start_result
-
+  failed_when: false
+  
+- name: Sleep 10 seconds when starting the node failed
+  pause: secs=10
+  when: node_start_result.state != 'started'
+  
+- name: Try starting the node again
+  service: name={{ openshift.common.service_type }}-node enabled=yes state=started
+  register: node_start_result
+  when: node_start_result.state != 'started'
+  
 - set_fact:
     node_service_status_changed: "{{ node_start_result | changed }}"


sdodson:
aos-ci-test

roles/openshift_node/tasks/main.yml
```diff
@@ -133,11 +133,20 @@
...
+  failed_when: false
+  
+- name: Sleep 10 seconds when starting the node failed
+  pause: secs=10
+  when: node_start_result.state != 'started'
+  
+- name: Try starting the node again
+  service: name={{ openshift.common.service_type }}-node enabled=yes state=started
+  register: node_start_result
+  when: node_start_result.state != 'started'
```
detiber: Is there a reason that the node is failing to start on the first try?

Also, it may make more sense to just set a notify for the node restart handler instead of enabling/starting the node again.

sdodson: The environment has been blown away but from what I remember of the logs I saw ansible-service start it, the node starts, restarts docker during SDN initialization at which point the service is marked failed, docker and master restart, systemd restarts the node but by this point ansible already exited.

detiber: @sdodson in that case, would it make sense to make the node start/enable an async task and validate it afterwards?

sdodson:
aos-ci-test

roles/openshift_node/tasks/main.yml
brenton: We'd still the sleep call though, right?  It's hard to validate something that may disappear randomly in a 10 second window.  Whatever we decide we should debug this a little more and write up a good comment.  This definitely reads weird.

detiber: I'm thinking of having it be an async task, and then following up on checking the service status using a do-until loop. It might require some tweaking of the delay value to make sure it doesn't check in too early, though.

brenton: Can we reliably use delegate_to now?  I ask because I think it would be required.  This role is running on the Node yet the only way I know of to call the Node healthz url is with the admin cert and key which is only on the Master.  I think I'd be happy if this PR simply had some comments for now and then we continue working to find a way to have systemd do the correct thing and not kill the Node on startup.

detiber: @brenton yes, delegate_to is working reliably now.

brenton:
@sdodson how are the 3.1 containerized installs passing without this workaround?  Was there something done in another PR?

sdodson:
I'm not sure. They also seem to be passing without your PR as far as I can tell. That has me worried even more than if it were still failing.

sdodson:
The only thing I can think of is that we drove installation and configuration of docker into the install process at a much earlier point in time. It's quite possible that the changes necessary to downgrade docker mean that now docker is no longer restarted after any of the containerized services are created.  With the exception of the restart that's triggered when the node re-configures docker networking, but as far as I know that always worked.

PR #1679, opened 2016-03-29
Refactor of openshiftcli to be more generic.

kwoodson:


roles/lib_openshift_api/build/src/base.py
```diff
@@ -1,7 +1,14 @@
-#!/usr/bin/env python
+# pylint: skip-file
 '''
   OpenShiftCLI class that wraps the oc commands in a subprocess
+
+
+#################################
+#  THIS MODULE IS AUTOGENARATED #
+#################################
```
twiest: I want the autogenerated message to be more "in your face."

Consider doing something like this:
https://github.com/openshift/openshift-tools/blob/master/docker/oso-ops-base/centos7/build.sh#L2-L9

roles/lib_openshift_api/build/src/base.py
```diff
@@ -1,7 +1,17 @@
...
 '''
   OpenShiftCLI class that wraps the oc commands in a subprocess
+
+     ___ ___ _  _ ___ ___    _ _____ ___ ___
+    / __| __| \| | __| _ \  /_\_   _| __|   \
+   | (_ | _|| .` | _||   / / _ \| | | _|| |) |
+    \___|___|_|\_|___|_|_\/_/_\_\_|_|___|___/_ _____
+   |   \ / _ \  | \| |/ _ \_   _| | __|   \_ _|_   _|
+   | |) | (_) | | .` | (_) || |   | _|| |) | |  | |
+   |___/ \___/  |_|\_|\___/ |_|   |___|___/___| |_|
```
twiest: Could we change the generator to add this?

It's strange to have this in a file that isn't generated. I know that the purpose is so that it's eventually in the generated file, but it feels odd to be in this file directly.

I'm thinking right after the `#!/usr/bin/env python` line is added in the generator.

twiest:
:+1:

PR #1681, opened 2016-03-29
Fixes for containerized installs

brenton:


brenton:
@arilivigni @sdodson, @detiber, this resolves the containerized installation problems for me.

sdodson:
aos-ci-test

sdodson:
LGTM

arilivigni:
@brenton @sdodson testing now in my env (fingers crossed)

arilivigni:
@brenton no luck did you do an all-in-one node 

https://paste.fedoraproject.org/347273/30457314/

@sdodson

brenton:
Weird.  I'm not seeing this.  My environment had two hosts.  One master/node and one node.

sdodson:
I think that's fixed by #1678 but I don't understand why that wouldn't happen in all environments. I'd be interested in seeing what happens in CI with both PRs.

arilivigni:
Has anyone tried this in a 3 node environment?  Or would someone like to actually use the CI environment where we create the nodes to root cause what is happening?  I don't think enabling this at this point gives us any value if it fails.

brenton:
I don't think there would be anything unique about 3 nodes.  The issue PR that scott referenced seems mostly likely.

arilivigni:
@sdodson is this any different than the fix you provided yesterday that I already tried?

brenton:
@arilivigni, yes.  We're saying you need this PR and #1678.

arilivigni:
@brenton what I am saying is I already tried this from @sdodson yesterday besides your change and I was trying to find out if an additional change was made to this PR.  Maybe one of you guys should give it a try in the environment I can just give you the IPs and then you can debug it

sdodson:
> @sdodson is this any different than the fix you provided yesterday that I already tried?

I believe I fixed the ansible syntax error after you pastebinned it, I don't know that my PR will fix your install but it should fix the error you pastebinned. It is however a bit hard for me to test because I'm not hitting the failure path you are, I tried to force the error but I failed.

brenton:
Is there any reason we don't enable the containerized CI job now with aos-ci-test?  That way we could add all the fixes we know of to a signle PR and debug it there.

arilivigni:
I don't think it is worth enabling something until we have a fix so I am basically running what we have in CI and seeing a failure and these PRs so what is the difference?

brenton:
@arilivigni, the difference would be we wouldn't have to bother you anymore.  I really think what you're hitting are real ansible problems and not CI job problems.

arilivigni:
@brenton it will take me some cycles to get this in then.  I have to add another context for this in the jobs.

brenton:
aos-ci-test

brenton:
aos-ci-test

brenton:
aos-ci-test

brenton:
aos-ci-test

brenton:
This is so crazy.  I'm still hitting cases where I need this commits yet is doesn't seem that the CI environments hit it.

sdodson:
Where exactly is it failing for you?

brenton:
It's failing the same way.  My environment isn't anything really special.  I am running ansible on one of the masters but I don't exactly see how that would cause this.

detiber:
could it be that the node is coming up before the master is and that is what is causing the node start to fail the first time?

tbielawa:
@detiber @brenton still relevant issue to you guys?

PR #1682, opened 2016-03-30
Bug 1322338 - The upgrade should keep the option insecure-registry=17…

brenton:
…2.30.0.0/16


brenton:
openshift_cli pulls in the docker roles and needs the variables.

@detiber, PTAL

detiber:
:+1:

PR #1685, opened 2016-03-31
Switch to using CentOS SIG repos for Origin installs

detiber:


detiber:
@tdawson here is the branch I'm using for testing the CentOS SIG repo builds

sdodson:
Was speaking with Lokesh yesterday, he's suggested that we use the centos virt sig's testing repo for docker. 
http://cbs.centos.org/repos/virt7-docker-common-testing/x86_64/os/

Other than that openvswitch should be in extras now, I think?

detiber:
@sdodson I'm not seeing openvswitch in http://mirror.centos.org/centos/7/extras/x86_64/Packages/

We also discussed during a previous Cent OS PaaS SIG meeting whether we wanted to use packages from other SIG repos and we decided that it would be better to ship those in our repos to avoid package version related issues.

sdodson:
@detiber sorry, you're right about openvswitch.
For docker can we tag something from another SIG into our targets? I guess it's fine to build it on our own but in the short time I've been building docker in the COPR it's been an incredible time sink for me.

detiber:
@tdawson what are your thoughts? do you have to rebuild docker each time, or are you able to tag in an existing build?

sdodson:
Wondering if we should do this now as 1.2.0 is about to be released. Centos PaaS could track stable builds while the origin-next copr could track alpha/beta builds.

@tdawson @maxamillion thoughts?

detiber:
On May 13, 2016 1:16 PM, "Scott Dodson" <notifications@github.com> wrote:
>
> Wondering if we should do this now as 1.2.0 is about to be released.
Centos PaaS could track stable builds while the origin-next copr could
track alpha/beta builds.
>
> @tdawson @maxamillion thoughts?

I've missed the meeting the past couple of weeks, but if we have a release
in the stable repo for the CentOS SIG, then I am all for it.

@maxamillion we'll probably want to do something similar for Fedora when an
official Fedora and/or Fedora SIG package is available.

detiber:
@sdodson Should we go ahead and merge this, or wait until we have a release package to enable gpg?

tdawson:
Sorry, I didn't see this due to a filter.
Stable OpenShift Origin packages are now published.  They are at the URL in this pull request.
You can also get the repo by installing centos-release-openshift-origin, from centos-extras.
That stable repo has openvswitch, and the patched docker, that allows origin to install in it.

detiber:
@tdawson @sdodson updated for using the release packages. I haven't tested yet, and I'm not sure if installing the release package automatically enables the repos or not...

tdawson:
centos-openshift-origin yum repo is enabled by default when you install centos-release-openshift-origin

detiber:
A quick test shows this appears to be working for me, so I'm going to go ahead and merge.

PR #1690, opened 2016-03-31
Bug 1322788 - The IMAGE_VERSION wasn't added to atomic-openshift-mast…

brenton:
…er-api and atomic-openshift-master-controllers


brenton:
@abutcher, PTAL.  From looking at https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-master/config.yml#L57 it seems like it's safe to ensure `openshift.master.ha` is set on upgrade for all masters.  The bug was due to it not being defined.

abutcher:
:+1:

PR #1691, opened 2016-03-31
Check consumed pools prior to attaching.

### LINKED ISSUE #1670: Unable to run rhel_subscribe playbook multiple times
It fails because once you run it the first time, the pool becomes part of the consumed pools for the system, so subsequent runs of `Retrieve the OpenShift Pool ID` fail because the pool doesn't show up in `subscription-manager list --available --matches="{{ rhel_subscription_pool }}" --pool-only`

abutcher:
Also fail when we don't get a match in either available or consumed.

Resolves #1670

### DIFF
--- a/roles/rhel_subscribe/tasks/main.yml
+++ b/roles/rhel_subscribe/tasks/main.yml
@@ -37,8 +37,19 @@
   register: openshift_pool_id
   changed_when: False
 
+- name: Determine if OpenShift Pool Already Attached
+  command: subscription-manager list --consumed --matches="{{ rhel_subscription_pool }}" --pool-only
+  register: openshift_pool_attached
+  changed_when: False
+  when: openshift_pool_id.stdout == ''
+
+- fail:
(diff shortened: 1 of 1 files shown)


ncdc:
Appears to work, although I think it ended up reattaching me to the pool instead of skipping over it

abutcher:
If more than 1 pool matches `rhel_subscription_pool` the role will keep tacking them on.

sdodson:
The error this is fixing matches this?
```
TASK: [rhel_subscribe | Attach to OpenShift Pool] *****************************                  
fatal: [ose3-master.example.com] => One or more undefined variables: list object has no element 0
fatal: [ose3-node2.example.com] => One or more undefined variables: list object has no element 0 
fatal: [ose3-node1.example.com] => One or more undefined variables: list object has no element 0 
```

abutcher:
@sdodson Yeah, this should fix that.

detiber:
@abutcher any issues with this merging?

abutcher:
@detiber Nope. The role needs more work wrt attaching the right pool when multiple match but this is okay.

PR #1692, opened 2016-03-31
Fix openshift_node_lables for aws bin/cluster

detiber:


bitwalker:
:+1 thanks again for the help!

detiber:
@lhuard1A can you review this change? it looks like we were failing to set some variables we were using in the config playbook after launching the instances, causing the config playbook to fail on newly provisioned hosts.

PR #1693, opened 2016-03-31
Adding list editing support to yaml editor

kwoodson:


roles/lib_openshift_api/build/generate.py
```diff
@@ -27,18 +28,23 @@
...
+                        'src/edit.py',
+                        'ansible/edit.py',
+                       ],
         }
 
 
 def main():
     ''' combine the necessary files to create the ansible module '''
-    openshift_ansible = ('../library/')
+    library = os.path.join(OPENSHIFT_ANSIBLE_PATH, '../library/')
```
twiest: Are you sure this is the right way to do this?

I thought it should be something more like this (so that the library puts the path together for us):
```
library = os.path.join(OPENSHIFT_ANSIBLE_PATH, '..', 'library')
```

roles/lib_yaml_editor/build/src/yedit.py
```diff
@@ -47,20 +75,52 @@ def add_entry(data, keys, item):
...
+            elif isinstance(data, dict):
+                if arr:
+                    keys = keys.split('[')[0]
+                    data[keys][int(arr[0])] = item
+
+                else:
+                    data[keys] = item
 
 
+    # pylint: disable=too-many-branches,too-many-return-statements
```
twiest: I'm not a fan of disabling too-many-branches. We should make smaller, more cohesive functions. This makes it so that they can be more easily too.

roles/lib_yaml_editor/build/src/yedit.py
```diff
@@ -47,20 +75,52 @@ def add_entry(data, keys, item):
...
     @staticmethod
     def get_entry(data, keys):
```
twiest: get_entry, delete_entry and add_entry container a lot of the same functionality to find the specific part of the dictionary to act on. That code needs to be shared. Here it's repeated over and over.

roles/lib_openshift_api/build/ansible/edit.py
```diff
@@ -0,0 +1,78 @@
+# pylint: skip-file
+
+# pylint: disable=too-many-branches
```
twiest: Why is this needed?

roles/lib_openshift_api/library/oc_secret.py
```diff
@@ -248,21 +279,24 @@ def check_def_equal(user_def, result_def, debug=False):
 
         return True
 
+#import os, yaml, re
```
twiest: remove

twiest:
:+1:

PR #1695, opened 2016-04-01
Fixing bugs 1322788 and 1323123

brenton:


playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/containerized_upgrade.yml
```diff
@@ -1,7 +1,7 @@
 - include_vars: ../../../../../roles/openshift_node/vars/main.yml
 
 - name: Update systemd units
-  include: ../../../../../roles/openshift_node/tasks/systemd_units.yml openshift_version=g_aos_versions.avail_version
+  include: ../../../../../roles/openshift_node/tasks/systemd_units.yml openshift_version=v{{ g_aos_versions.avail_version }}
```
brenton: @detiber, this is the thing I mentioned to you that was probably never working correctly but was masked by variable precedence previously.

brenton:
@detiber, PTAL.

When I was digging into these bugs it made the most sense to move `openshift_version` to the `defaults/main.yml`.  My understanding is that those variables have the lowest of all precedence.  Given that I would need to very carefully override `openshift_version` during the upgrade I wanted to make it absolutely clear that the value I'm passing in on the role overrides whatever possible default could be set.

brenton:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/containerized_upgrade.yml
detiber: hah, nice!

roles/openshift_common/defaults/main.yml
```diff
@@ -1,3 +1,4 @@
 ---
 openshift_cluster_id: 'default'
 openshift_debug_level: 2
+openshift_version: "{{ openshift_pkg_version | default('') }}"
```
detiber: Do we want to allow this to be overridden by users?

detiber: just fully read the comment you tagged me in, nevermind...

brenton: I don't really see why not.  It would definitely be an advanced scenario.

sdodson:
aos-ci-test

brenton:
aos-ci-test

arilivigni:
aos-ci-test

brenton:
aos-ci-test

PR #1703, opened 2016-04-04
a-o-i: Update README

smunilla:
Update to the readme file to reflect the new, better method for installing
the quick-installer from source


detiber:
I would suggest that we also add a set of instructions for how to run directly from the source tree, similar to: https://github.com/ftl-toolbox/ansible-swagger-generator/blob/master/README.md#running-from-source

By using the virtualenv and running `python setup.py develop` it allows you to better test changes as you are developing.

PR #1705, opened 2016-04-04
Add support for creating secure router

abutcher:
* Adds option for specifying an `openshift_hosted_router_certificate` cert and key pair.
* Removes dependency on node label variables and retrieves the node list from the API s.t. this role can be applied to any cluster with existing nodes. I've added an openshift_hosted playbook that occurs after node install to account for this.
* Infrastructure nodes are selected using `openshift_hosted_router_selector` which is based on deployment type by default; openshift-enterprise -> "region=infra" and online -> "type=infra".

Inventory options
```
# OpenShift Router Options
# 
# An OpenShift router will be created during install if there are
# nodes present with labels matching the default router selector,
# "region=infra". Set openshift_node_labels per node as needed in
# order to label nodes.
# 
# Example:
# [nodes]
# node.example.com openshift_node_labels="{'region': 'infra'}"
# 
# Router selector (optional)
# Router will only be created if nodes matching this label are present.
# Default value: 'region=infra'
#openshift_hosted_router_selector='region=infra'
# 
# Router replicas (optional)
# Unless specified, openshift-ansible will calculate the replica count
# based on the number of nodes matching the openshift router selector.
#openshift_hosted_router_replicas=2
#
# Router certificate (optional)
# Provide local certificate paths which will be configured as the
# router's default certificate.
#openshift_hosted_router_certificate={"certfile": "/path/to/router.crt", "keyfile": "/path/to/router.key"}
```

Example playbook
```
- name: Create hosted resources
  hosts: oo_first_master
  roles:
  - role: openshift_hosted
    openshift_hosted_router_certificate:
      certfile: /path/to/my-router.crt
      keyfile: /path/to/my-router.key
    openshift_hosted_router_registryurl: 'registry.access.redhat.com/openshift3/ose-haproxy-router:v3.0.2.0'
    openshift_hosted_router_selector: 'type=infra'
```

@dak1n1 @detiber ptal


playbooks/common/openshift-cluster/config.yml
```diff
@@ -31,6 +31,6 @@
 
 - include: ../openshift-master/config.yml
 
-- include: additional_config.yml
-
 - include: ../openshift-node/config.yml
+
+- include: additional_config.yml
```
detiber: I'm not sure I'm a fan of this...  having additional config happen before nodes was kind of nice.

roles/openshift_router/tasks/main.yml
```diff
@@ -1,10 +1,66 @@
...
+  command: >
+    {{ openshift.common.client_binary }} --api-version='v1' -o json
+    get nodes -n default --config={{ openshift.common.config_base }}/master/admin.kubeconfig
+  register: openshift_router_nodes_json
+
+- name: Collect nodes matching router selector
+  set_fact:
+    openshift_router_nodes: >
+      {{ (openshift_router_nodes_json.stdout|from_json)['items']
+         | oo_oc_nodes_matching_selector(openshift.router.selector) }}
```
detiber: What if we have an optional variable to pass in for a list of nodes? That would allow for initial installation to still run this role prior to nodes being installed.

detiber:
Overall, outside of wanting to avoid pushing additional config to after the nodes I like where this is going.

One alternative I can think of, is to separate out things like router/registry/logging/metrics and doing those in a post-node step instead of moving all of additional_config

roles/openshift_router/tasks/main.yml
abutcher: Yeah I'm cool with that, I was also thinking some folks might just want to tell us the replicas without us checking labels...

playbooks/common/openshift-cluster/config.yml
abutcher: I'm okay with separating router/registry/etc from additional_config but what is nice about running additional config before nodes?

detiber: There is the flannel config, the service account creation, the manageiq integraion, openshift_examples, etc.

Also, at some point we want to declare failure if < x% of nodes fail to install, at which point none of the additional_config would be completed.

roles/openshift_router/tasks/main.yml
detiber: yeah, good call.

playbooks/common/openshift-cluster/config.yml
abutcher: Oy, yeah as things stand some of that will need to be separated. I'd like to see service accounts become dependencies on the roles that needs them and maybe flannel be a conditional dep for master.

detiber: Agreed on the refactoring, I just don't want to hold up this PR for that large of a change

abutcher:
aos-ci-test

abutcher:
aos-ci-test

playbooks/common/openshift-cluster/config.yml
```diff
@@ -34,3 +34,5 @@
 - include: additional_config.yml
 
 - include: ../openshift-node/config.yml
+
+- include: openshift_router.yml
```
detiber: Should this playbook be called openshift_hosted instead?

I'm also wondering if we should take this opportunity to rename the openshift_router role to openshift_hosted to start the consolidation of the "hosted" roles into a openshift_hosted role...

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -403,16 +403,18 @@ def set_master_selectors(facts):
...
     """
+    deployment_type = facts['common']['deployment_type']
+    if deployment_type == 'online':
+        selector = "type=infra"
+    else:
+        selector = "region=infra"
+
+    if 'router' in facts:
+        if 'selector' not in facts['router']:
+            facts['router']['selector'] = selector
```
detiber: There should probably be a migration to move any existing value of openshift.master.router_selector to openshift.router.selector as well.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1411,7 +1413,8 @@ class OpenShiftFacts(object):
                    'etcd',
                    'hosted',
                    'master',
-                   'node']
+                   'node',
+                   'router']
```
detiber: router should probably live under the hosted role, instead of a new role.

roles/openshift_router/meta/main.yml
```diff
@@ -12,4 +12,4 @@ galaxy_info:
   categories:
   - cloud
   dependencies:
-  - openshift_facts
+  - openshift_common
```
detiber: hmm, I'm wondering if the common role is enough here to make this fully standalone...  But at the same time, I wouldn't want to pull in the openshift_master playbook here either.

abutcher:
aos-ci-test

abutcher:
aos-ci-test

filter_plugins/oo_filters.py
```diff
@@ -601,36 +664,38 @@ def oo_persistent_volumes(hostvars, groups, persistent_volumes=None):
...
-        for component in hostvars['openshift']['hosted']:
-            kind = hostvars['openshift']['hosted'][component]['storage']['kind']
-            create_pv = hostvars['openshift']['hosted'][component]['storage']['create_pv']
-            if kind != None and create_pv:
-                if kind == 'nfs':
-                    host = hostvars['openshift']['hosted'][component]['storage']['host']
-                    if host == None:
-                        if len(groups['oo_nfs_to_config']) > 0:
-                            host = groups['oo_nfs_to_config'][0]
+        if 'hosted' in hostvars['openshift']:
```
detiber: We may want to see if we can clean this up at some point in the future, this is going to get big and hairy as we continue to add storage types.

abutcher: It sure is

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -63,7 +63,16 @@ def migrate_local_facts(facts):
     migrated_facts = copy.deepcopy(facts)
     return migrate_docker_facts(migrated_facts)
 
-
+def migrate_master_facts(facts):
```
detiber: Should this be migrate_hosted_facts instead?

filter_plugins/oo_filters.py
detiber: I'm actually thinking that a template may be a better fit, especially for the hosted components.

detiber: Either way, just something to think about, no reason to block this PR on it :)

roles/openshift_facts/library/openshift_facts.py
abutcher: Renamed

detiber:
lgtm

detiber:
aos-ci-test

abutcher:
@brenton ptal

PR #1707, opened 2016-04-05
Commenting post-install change implicatations for the sample inventories

brenton:


brenton:
Once we know my comments are accurate I'll copy them to the other sample inventory files.

inventory/byo/hosts.ose.example
```diff
@@ -225,10 +260,14 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
 #
 # Signing secrets, used to authenticate sessions using
 # HMAC. Recommended to use secrets with 32 or 64 bytes.
+#
+# Changing this post deployment will invalidate all tokens.
```
brenton: @liggitt Are these two comments about session secrets accurate?

liggitt: it invalidates login sessions (in progress logins), not OAuth tokens

inventory/byo/hosts.ose.example
```diff
@@ -201,6 +233,9 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
 # set RPM version for debugging purposes
 #openshift_pkg_version=-3.1.0.0
 
+# Explicitly set the tag that should be used for containerized installs
+#openshift_image_tag=v3.1.1.6
```
brenton: @dgoodwin, is this still the correct way to force a given image version?

dgoodwin:
Still the correct way yes, just not needed anymore for most users.

brenton:
I reviewed the current example hosts file and the current docs are even better than what this PR was originally trying to do.

PR #1710, opened 2016-04-05
iptables packages need to be same version

### LINKED ISSUE #1709: unreliable iptables-services installation
It looks like `iptables` package not upgraded if exists but that can break `iptables-services` installation which is newer in repo. I think `iptables-services` should use same version as `iptables`  or have both packages forced to version `latest`.

```
TASK: [os_firewall | Install iptables packages] ******************************* 
ok: [ec2-52-23-215-161.compute-1.amazonaws.com] => (item=iptables) => {"changed": false, "item": "iptables", "msg": "", "rc": 0, "results": ["iptables-1.4.21-14.el7.x86_64 providing iptables is already installed"]}
failed: [ec2-52-23-215-161.compute-1.amazonaws.com] => (item=iptables-services) => {"changed": true, "item": "iptables-services", "rc": 1, "results": ["Loaded plugins: amazon-id, rhui-lb, search-disabled-repos\nResolving Dependencies\n--> Running transaction check\n---> Package iptables-services.x86_64 0:1.4.21-16.el7 will be installed\n--> Finished Dependency Resolution\n\nDependencies Resolved\n\n================================================================================\n Package                 Arch         Version               Repository     Size\n================================================================================\nInstalling:\n iptables-services       x86_64       1.4.21-16.el7         rhel-72        50 k\n\nTransaction Summary\n================================================================================\nInstall  1 Package\n\nTotal download size: 50 k\nInstalled size: 24 k\nDownloading packages:\nRunning transaction check\nERROR with transaction check vs depsolve:\niptables = 1.4.21-16.el7 is needed by iptables-services-1.4.21-16.el7.x86_64\n You could try running: rpm -Va --nofiles --nodigest\nYour transaction was saved, rerun it with:\n yum load-transaction /tmp/yum_save_tx.2016-04-05.15-50.02Wuyy.yumtx\n"]}

```

akostadinov:
My local testing shows this fix #1709

### DIFF
--- a/roles/os_firewall/tasks/firewall/iptables.yml
+++ b/roles/os_firewall/tasks/firewall/iptables.yml
@@ -21,7 +21,7 @@
   ignore_errors: yes
 
 - name: Install iptables packages
-  action: "{{ ansible_pkg_mgr }} name={{ item }} state=present"
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=latest"
   with_items:
   - iptables
   - iptables-services


sdodson:
Yeah, I think this would fix it.
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

detiber:
lgtm @brenton

detiber:
actually, I'm not sure we want to do  this...  if someone is using this for ongoing CM, we could inadvertently upgrade iptables and iptables-service on them during a regular CM run.

roles/os_firewall/tasks/firewall/iptables.yml
```diff
@@ -21,7 +21,7 @@
   ignore_errors: yes
 
 - name: Install iptables packages
-  action: "{{ ansible_pkg_mgr }} name={{ item }} state=present"
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=latest"
```
detiber: Instead of using latest, we would probably want to separate out the iptables and iptables-services installation into separate tasks. We'll probably also have to use repoquery/dnf repoquery to query the locally installed version of iptables and force iptables-services to match that version....

Either that, or would we have a better experience by just shelling out to the native package management tools for this installation (I'm not sure if there is a broken dependency on the iptables/iptables-services packages or if it is related to the underlying ansible yum package module causing the issue).

sdodson: Filed https://bugzilla.redhat.com/show_bug.cgi?id=1324914
Found https://bugzilla.redhat.com/show_bug.cgi?id=1096752

I looked at their spec file and I cannot figure out what's wrong with the package.

akostadinov: Good catch. Too bad maintainer closed original bug instead of routing it properly.

sdodson:
I think we concluded that this only happens when packages were installed from a pre-release repo and that repo is no longer available when ansible runs. Closing this, re-open if I'm wrong.

PR #1715, opened 2016-04-06
Make cockpit require openshift_docker

sdodson:
Cockpit is installed on masters, which
must also be nodes, so it's ok for now to call openshift_docker there

### DIFF
--- a/roles/cockpit/meta/main.yml
+++ b/roles/cockpit/meta/main.yml
@@ -13,3 +13,4 @@ galaxy_info:
   - cloud
 dependencies:
   - { role: os_firewall }
+  - { role: openshift_docker }


sdodson:
aos-ci-test

detiber:
Is this so that we can deploy a containerized cockpit, or to ensure that cockpit doesn't install an unsupported version of docker?

sdodson:
cockpit is installing after atomic-openshift is installed but before docker is installed so it's getting a depsolving issue. I guess another solution would be to move the cockpit install earlier, which would install the latest version of docker then rely on properly downgrading docker.

detiber:
@sdodson I'd rather avoid the downgrade if we don't really need it. :+1: to adding the openshift_docker dependency.

sdodson:
aos-ci-test

arilivigni:
aos-ci-test

arilivigni:
@sdodson subscription manager looks like it had an issue kicking this off again

arilivigni:
aos-ci-test

arilivigni:
@sdodson testing a fix for this in our playbooks so it uses 3.0 repo trying to move away from subscription manager since it is down

arilivigni:
aos-ci-test

sdodson:
aos-ci-test

PR #1716, opened 2016-04-06
Unmask services

brenton:
Unmasking etcd is is important when switching from containerized installs to
RPM installs


brenton:
@detiber PTAL

detiber:
@brenton lgtm

PR #1718, opened 2016-04-06
Containerized installs on RHEL were downgrading docker unnecessarily

brenton:


roles/docker/tasks/main.yml
```diff
@@ -10,7 +10,12 @@
 - name: Downgrade docker if necessary
   command: "{{ ansible_pkg_mgr }} downgrade -y docker-{{ docker_version }}"
   register: docker_downgrade_result
-  when: not docker_version_result | skipped and docker_version_result | default('0.0', True) | version_compare(docker_version, 'gt')
+  when: not docker_version_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare(docker_version, 'gt')
```
brenton: @sdodson, I think not using `.stdout` was just a bug previously.

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -46,10 +46,24 @@
...
   set_fact:
     docker_version: "{{ '1.8.2' }}"
-  when: " ( common_version.stdout | default('0.0', True) | version_compare('3.2','<') and openshift.common.service_type == 'atomic-openshift' ) or
-          ( common_version.stdout | default('0.0', True) | version_compare('1.1.4','<') and openshift.common.service_type == 'origin' )"
+  when: " ( l_common_version | version_compare('3.2','<') and openshift.common.service_type == 'atomic-openshift' ) or
+          ( l_common_version | version_compare('1.1.4','<') and openshift.common.service_type == 'origin' )"
+
+- name: Set docker version to be installed
+  set_fact:
+    docker_version: "{{ '1.9.1' }}"
```
brenton: @detiber, @sdodson, I'm not really excited about having to explicitly declare compatible version numbers.   I suspect our supported versions of docker for each OSE release are only going to get more complicated over time.  I definitely want the ansible install code to be a simple as possible and least likely to fail.  I wanted to set `docker_version` to 1.9 however that would complicate the gt and lt checks.  On OSE 3.1 if you are running 1.9.1 you need to downgrade.  However, if you are running OSE 3.2 you don't need to downgrade or upgrade.

roles/docker/tasks/main.yml
```diff
@@ -10,7 +10,12 @@
...
+
+- name: Upgrade docker if necessary
+  command: "{{ ansible_pkg_mgr }} upgrade -y docker-{{ docker_version }}"
+  register: docker_upgrade_result
+  when: docker_downgrade_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare(docker_version, 'lt')
```
detiber: Is this needed? wouldn't the yum/dnf install step below handle the upgrade step?

brenton:
aos-ci-test

brenton:
aos-ci-test

roles/docker/tasks/main.yml
brenton: I'll keep digging.  I promise it's needed.  Every time I tried to do a containerized install my docker was downgraded.

roles/docker/tasks/main.yml
abutcher: :+1:

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -46,10 +46,24 @@
...
-  when: not openshift.common.is_atomic | bool
+  when: not openshift.common.is_containerized | bool
+
+- set_fact:
+    l_common_version: "{{ openshift.common.image_tag | oo_image_tag_to_rpm_version | default('0.0', True) }}"
+  when: openshift.common.is_containerized | bool
+
+- set_fact:
+    l_common_version: "{{ common_version.stdout | default('0.0', True) }}"
+  when: not openshift.common.is_containerized | bool
```
detiber: I probably have an irrational hate of set_fact...  I can't think of a way to put this into a vars file though, with the use of common_version.stdout though...

it could be limited down to one set_fact if you do the following though:
```
l_common_version: "{{ ( openshift.common.image_tag | oo_image_tag_to_rpm_version if openshift.common.is_containerized | bool else common_version.stdout) | default('0.0', True) }}"

File: roles/openshift_docker_facts/tasks/main.yml (hunk 2 above)
Comment (detiber) [inline roles/openshift_docker_facts/tasks/main.yml]: @brenton I'm wondering if we shouldn't worry about upper bounds on docker version unless we have a known issue like we have with 3.1. Then during upgrades we could handle upgrading docker if necessary.

File: roles/docker/tasks/main.yml (hunk 3 above)
Comment (brenton) [inline roles/docker/tasks/main.yml]: It does appear that I can remove the upgrade logic and leave the other fixes.  Thx!

File: roles/openshift_examples/files/examples/v1.2/xpaas-streams/jboss-image-streams.json [hunk 5]
```diff
@@ -152,8 +152,8 @@
                             "version": "1.3"
                         },
                         "from": {
-                          "Kind": "DockerImage",
-                          "Name": "registry.access.redhat.com/jboss-eap-7-beta/eap70-openshift:1.3"
+                          "kind": "DockerImage",
+                          "name": "registry.access.redhat.com/jboss-eap-7-beta/eap70-openshift:1.3"
                         }
```
brenton: @sdodson, I think the reason these passed testing is because they ran against 3.0 and 3.1.  It looks like for 3.2 it breaks validation.

roles/openshift_docker_facts/tasks/main.yml
brenton: I'd prefer to wait to handle the upper bound issue until we see what the docker 1.9 vs 1.10 packaging looks like.  There are still a lot of possibilities and I don't think we would guess correctly at this point.

roles/openshift_docker_facts/tasks/main.yml
brenton: I'd do whatever you prefer.  I sort of hate inlining if/else like this.  It almost always gets more complicated and I can no longer count how many bugs I've seen because of followup edits that subtly break things.

roles/openshift_examples/files/examples/v1.2/xpaas-streams/jboss-image-streams.json
sdodson: crap, yeah. filed an issue with upstream of us so we don't inadvertantly re-introduce this https://github.com/jboss-openshift/application-templates/issues/144

roles/openshift_docker_facts/tasks/main.yml
sdodson: I wonder if ensure=latest does the right thing if you're specifying version? We could just not specify the version for 3.2 until such a point that we know of an upper bound. Then the install task would ensure it's the latest.

roles/openshift_examples/files/examples/v1.2/xpaas-streams/jboss-image-streams.json
brenton: no worries, I figured I'd bundle it in with some other PR.

brenton:
aos-ci-test

roles/openshift_docker_facts/tasks/main.yml
brenton: I'll test that out.

brenton: Just thinking a little more.  This role gets applied many times throughout an install.  We'd have to be pretty careful how we crafted the `when:` clause on the docker install to make sure we didn't downgrade the first time the role is applied only to upgrade to the latest later in the run.

roles/openshift_docker_facts/tasks/main.yml
detiber: That's fine, if it can't be eliminated by pushing it to vars, it's not really worth it anyway.

roles/openshift_docker_facts/tasks/main.yml
detiber: Every use of the docker role should be through the openshift_docker role which is pulling data from openshift_facts.

arilivigni:
aos-ci-test

roles/openshift_docker_facts/tasks/main.yml
brenton: I was mostly just thinking out loud that the conditional is going to be a little tricky and require a lot of testing.  I think my perception that this role was being applied multiple times was that I had multiple hosts in my environment that are all master+node.

roles/docker/tasks/main.yml
```diff
@@ -10,11 +10,11 @@
...
   command: "{{ ansible_pkg_mgr }} downgrade -y docker-{{ docker_version }}"
   register: docker_downgrade_result
-  when: not docker_version_result | skipped and docker_version_result | default('0.0', True) | version_compare(docker_version, 'gt')
+  when: not docker_version_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare(docker_version, 'gt')
 
 - name: Install docker
-  action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version != '' else '' }} state=present"
-  when: not openshift.common.is_atomic | bool and not docker_downgrade_result | changed
+  action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version_result.stdout == '' }} state=latest"
+  when: not openshift.common.is_atomic | bool and not docker_version_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare(docker_version, 'lt')
```
brenton: @detiber, @sdodson, this is what I'm testing.  It's weird but it seems to handle the 1) no docker installed 2) downgrade and 3) upgrade cases.  The idea is that if docker_version_result.stdout is empty we need to install the minimum required version.  If there is a version installed and we didn't need to downgrade then we can ensure the latest version.

I think there's still an edgecase if a version like 1.8.1 was installed.  This is a mess.

sdodson: Did it not work without changing from `if docker_version != ''` to `docker_version_result.stdout == ''` As you have it we'd always need to set docker_version, which would be the case if we changed from > 3.2/1.2 to >= 3.2/1.2 in the facts.

sdodson: Nevermind, I see you're accounting for that in the when condition.

roles/openshift_docker_facts/tasks/main.yml
detiber: @brenton indeed, for a HA master host, it will be applied 3 times (etcd, master, and node)

detiber: Just wanted to add a comment that we should avoid using `latest` for rpm installation to avoid issues when using these playbooks/roles for ongoing management.

roles/docker/tasks/main.yml
detiber: I'm actually surprised there isn't an error for line 16 and not having a else clause... I suppose that is just a jinja2'ism.

detiber: Again, I really  think we should avoid latest to avoid upgrading docker inadvertently during regular CM operations.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
```diff
@@ -1,5 +1,19 @@
...
+###############################################################################
+# The restart playbook should be run after this playbook completes.
+###############################################################################
+
+- name: Upgrade docker
+  hosts: oo_masters_to_config:oo_nodes_to_config:oo_etcd_to_config
+  roles:
+  - openshift_facts
+  tasks:
+  - include: docker_upgrade.yml
```
detiber: docker_upgrade.yml file missing from PR?

roles/openshift_docker_facts/tasks/main.yml
detiber: I'm not sure why I didn't think about this earlier, but would it make sense to push the complex version logic in openshift_docker_facts into the openshift_facts module directly, then we wouldn't need to run multiple tasks to set the docker_version that is passed to the docker role.

Moving the package version tests into the openshift_facts module would also probably prevent any issues around installing a newer version of docker only to downgrade later as well.

We'd still have some funkiness in the docker role proper to handle downgrade, but I think it'd still be worth it to keep the docker role generic.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
brenton: That file already existed.  I just had to move it earlier due to bugs in docker 1.8.2's --add-registry.  Without docker 1.9 the workaround for authenticated registries doesn't even work.

roles/docker/tasks/main.yml
brenton: I'm fine with reworking this to avoid using 'latest'.  I think one of the problems I was hitting was coming up with some simple logic without reworking what we already had too much.  I think if we don't have 'latest' the way the logic is now it could result in docker-1.9.1 being explicitly installed even though a 1.9.2 package was available.  That isn't the end of the world but it's a spell of overly complicated logic.

sdodson: I've posted this[0] question regarding how we should express requirements of the host os. If we're not worried about containerized installs this is a non issue because installing 3.2 atomic-openshift-node package will ensure that we're at a minimum version so it's really the upper version limits for 3.1 and 3.0 that we require.

0 - http://ask.projectatomic.io/en/question/3641/expressing-requirements-of-the-host-os/

roles/docker/tasks/main.yml
```diff
@@ -7,14 +7,19 @@
   register: docker_version_result
   changed_when: false
 
+- name: Remove deferred deletion for downgrades from 1.9
+  command: >
+    sed -i 's/--storage-opt dm.use_deferred_deletion=true//' /etc/sysconfig/docker-storage
+  when: docker_version_result.stdout | default('0.0', True) | version_compare('1.9', '>=') and docker_version | version_compare('1.9', '<')
```
brenton: @sdodson, what do you think about this?

brenton:
aos-ci-test

roles/docker/tasks/main.yml
```diff
@@ -7,14 +7,19 @@
   register: docker_version_result
   changed_when: false
 
+- name: Remove deferred deletion for downgrades from 1.9
+  command: >
+    sed -i 's/--storage-opt dm.use_deferred_deletion=true//' /etc/sysconfig/docker-storage
+  when: not docker_version_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare('1.9', '>=') and docker_version | version_compare('1.9', '<')
```
brenton: @sdodson, I just made another update to add `not docker_version_result | skipped`.

brenton:
aos-ci-test

roles/docker/tasks/main.yml
sdodson: Maybe it's just github but it looks like you didn't remove the old version of this from below? We also need the check for the file existence moved above this, i'd drop its when clause all together.

sdodson: Because the file may not exist if they haven't started docker before.

```
  - stat: path=/etc/sysconfig/docker
   register: docker_check
```

brenton:
aos-ci-test

detiber:
lgtm

roles/docker/tasks/main.yml
```diff
@@ -7,23 +7,27 @@
...
 - name: Downgrade docker if necessary
   command: "{{ ansible_pkg_mgr }} downgrade -y docker-{{ docker_version }}"
   register: docker_downgrade_result
   when: not docker_version_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare(docker_version, 'gt')
 
 - name: Install docker
-  action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version != '' else '' }} state=present"
-  when: not openshift.common.is_atomic | bool and not docker_downgrade_result | changed
+  action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version_result.stdout == '' }} state=present"
+  when: not openshift.common.is_atomic | bool and not docker_version_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare(docker_version, 'lt')
```
brenton: @detiber, @abutcher do you have any idea how this line could result in this error?

```
One or more undefined variables: the inline if-expression on line 1 evaluated to false and no else section was defined.

```

brenton: This showed that all the variables were actually defined as expected:
```
- debug: var=openshift
- debug: var=docker_version_result
- debug: var=docker_version
```

detiber: @brenton maybe stdout is not in docker_version_result?

abutcher: Line above it is missing an else

```
{{ '-' + docker_version if docker_version_result.stdout == '' else '' }}
```

brenton: From the logs I have it is:

```
2016-04-08 16:51:21,001 p=924 u=root |  ok: [ip-172-18-0-28.ec2.internal] => {
    "var": {
        "docker_version_result": {
            "changed": false,
            "cmd": [
                "repoquery",
                "--installed",
                "--qf",
                "%{version}",
                "docker"
            ],
** SNIP**
            "start": "2016-04-08 16:51:19.957724",
            "stderr": "",
            "stdout": "1.8.2",
            "stdout_lines": [
                "1.8.2"
            ],
            "warnings": []
        }
    }
}
2016-04-08 16:51:21,003 p=924 u=root |  ok: [ip-172-18-15-36.ec2.internal] => {
    "var": {
        "docker_version_result": {
            "changed": false,
**SNIP**
            "rc": 0,
            "start": "2016-04-08 16:51:20.019622",
            "stderr": "",
            "stdout": "1.8.2",
            "stdout_lines": [
                "1.8.2"
            ],
            "warnings": []
        }
    }
}
````

Comment (detiber) [inline roles/docker/tasks/main.yml]: @brenton nm, just did some more testing it's missing the else clause

```
action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version_result.stdout == '' else '' }} state=present"
```

Comment (detiber) [inline roles/docker/tasks/main.yml]: also looks like the docker_version_result.stdout test should be `!=`

Comment (brenton) [inline roles/docker/tasks/main.yml]: @abutcher, yeah, that was it. :(

Comment (brenton) [inline roles/docker/tasks/main.yml]: @detiber, this is pretty hard to read but here's the thinking:

If `docker_version_result.stdout == ''` that means docker was never installed.  Therefore we have to rely on the docker_version and result in something like `yum install docker-1.8.2`.  If there is stdout that means it's installed.  We've already previously downgraded to the correct version in the previous step.  So really we just have to make sure the package is installed to skip that test.  I'll be honest, I would have expected the 'when' clause to prevent that logic from ever running.

Comment (sdodson) [inline roles/docker/tasks/main.yml]: There's gaps in this that have been identified since I did the backport but should we try to be less clever for the sake of clarity? I liked how easy this was to follow in my backport https://github.com/openshift/openshift-ansible/blob/7e39163da55b20e7dc51837c8da6eebcbb68a69b/roles/docker/tasks/main.yml#L10-L65

Comment (brenton) [inline roles/docker/tasks/main.yml]: It's getting late on a Friday.  I'm honestly struggling to see how either of these are easy to follow given all the logic branches for containerized installs.

Comment (detiber) [inline roles/docker/tasks/main.yml]: So, I didn't test with Docker not installed, but with the current logic it failed for the upgrade scenario

**Comment by detiber** [comment]
```
action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version != '' else '' }} state=present"
```

I think the conditional in the action needs to use Docker_version and not Docker_version_result.

**Comment by brenton** [comment]
@detiber, I don't think that handles the case of docker not being installed.  I'll have to check though.

**Comment by brenton** [comment]
@aos-ci-test

**Comment by brenton** [comment]
@detiber, also are you saying the conditional needed to be `docker_version and not docker_version_result` or that we shouldn't use docker_version_result here?

**Comment by brenton** [comment]
I think the logic worked in https://github.com/openshift/openshift-ansible/blob/7e39163da55b20e7dc51837c8da6eebcbb68a69b/roles/docker/tasks/main.yml#L10-L65 because 1.8.2 was set unconditionally.  There was definitely a bug in this logic when I originally started working with it.  I don't doubt I may have broken something else though.

**Comment by brenton** [comment]
I'm testing out just using docker_version now.

**Comment by brenton** [comment]
aos-ci-test

**Comment by brenton** [comment]
@detiber, I updated the check to only use docker_version.

**Comment by brenton** [comment]
aos-ci-test

**Comment by brenton** [comment]
aos-ci-test

**Comment by arilivigni** [comment]
aos-ci-test

**Comment by brenton** [comment]
aos-ci-test

File: playbooks/common/openshift-cluster/upgrades/files/openshift_container_versions.sh [hunk 11]
```diff
@@ -3,19 +3,19 @@
...
     image_name="openshift/origin"
 elif grep aep $unit_file 2>&1 > /dev/null; then
-    image_name="aep3/aep"
+    image_name="aep3/node"
 elif grep openshift3 $unit_file 2>&1 > /dev/null; then
-    image_name="openshift3/ose"
+    image_name="openshift3/node"
 fi
 
+installed=$(docker run --rm --entrypoint=/bin/openshift ${image_name} version 2> /dev/null | grep openshift | awk '{ print $2 }' | cut -f1 -d"-" | tr -d 'v')
```
brenton: @detiber, are you OK with this?  It's true there's an edge case of a running service being a different version from the node image that's already cached, however, if they haven't run the installer yet and restarted their system they would be running that version anyway.  My goal here is to do something failsafe that will never result in setting the version to something that won't even start.  Also, this way the upgrade playbook can safely be re-run.

detiber: @brenton yeah, I think that's fine...  it's funny how I always thought the containerized installs would be "better" and "easier" :)

brenton:
aos-ci-test

brenton:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
```diff
@@ -40,6 +11,20 @@
   - include: docker_upgrade.yml
     when: not openshift.common.is_atomic | bool
 
+# The cli image is used by openshift_facts to determine the currently installed
+# version.  We need to explicitly pull the latest image to handle cases where
+# the locally cached 'latest' tag is older the g_new_version.
```
brenton: @abutcher, PTAL.  This was why Docker was downgraded when upgrading to 3.2.

brenton:
aos-ci-test

brenton:
aos-ci-test

arilivigni:
aos-ci-test

brenton:
aos-ci-test

brenton:
All 7 ci jobs are passing and I did a ton of manual testing today.  I'm merging this to unblock a lot of other testing.

PR #1720, opened 2016-04-06
Test docker_version_result.stdout when determining if docker should be installed/downgraded.

abutcher:
@sdodson ptal


abutcher:
Closing, changes included in #1718

PR #1723, opened 2016-04-06
Replace unused Dockerfile with one used for official builds.

dgoodwin:
A container that can be used to run the atomic-openshift-installer, or
by bypassing the entrypoint can be used to run ansible playbooks
directly.

Expects the user to mount a directory to /var/lib/openshift, where we'll
write out the installer.cfg, ansible.log, and hosts inventory.

WARNING: Don't run this on a host you're installing to, installation
restarts Docker which will kill your installer.

Example of running the quick installer:

docker run -ti -v ~/openshift/:/var/lib/openshift-installer/:Z -v ~/.ssh:/root/.ssh:Z openshift3/installer install

Example of running ansible directly:

docker run --entrypoint=/usr/bin/ansible-playbook -ti -v ~/openshift/:/var/lib/openshift-installer/:Z -v ~/.ssh:/root/.ssh:Z openshift3/installer -i /openshift-installer/install/.ansible/hosts /usr/share/ansible/openshift-ansible/playbooks/adhoc/uninstall.yml

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,23 @@
 FROM rhel7
 
-MAINTAINER Aaron Weitekamp <aweiteka@redhat.com>
+MAINTAINER Troy Dawson <tdawson@redhat.com>
 
-RUN yum -y install http://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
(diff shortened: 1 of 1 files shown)


detiber:
lgtm @brenton

Issue #1661, opened 2016-03-24
Handle additional OpenShift configuration files


simon3z:
Sometimes (especially in the configuration of `openshift_master_identity_providers`) a manual intervention is necessary to copy files on the nodes.

E.g. the `htpasswd` file for `HTPasswdPasswordIdentityProvider` or a certificate file for `LDAPPasswordIdentityProvider`, etc.

It should be possible to specify local files that gets deployed automatically.

cc @abutcher @detiber

detiber:
@simon3z for htpasswd in particular, do you think providing the file would be of more value, or specifying users and passwords (potentially pre-hashed) to be better?

I do agree on the ldap certs though, I hate asking someone to provide the cert as a variable. @abutcher can this be handled similarly to the named certificates?

simon3z:
> @simon3z for htpasswd in particular, do you think providing the file would be of more value, or specifying users and passwords (potentially pre-hashed) to be better?

I thought that asking for that would have been too much :smile: , but indeed it would be definitely better for us.

abutcher:
@detiber Yup, I'm not sure if the variable should be part of the provider dictionary or something adjacent but we can provide a mechanism to land a cert.

detiber:
@abutcher I'm thinking adjacent variables.

PR #1726, opened 2016-04-07
Add support for setting identity provider custom values

detiber:
Add support for setting identity provider custom values
- htpasswd users
- ldap ca file
- openid ca file
- request_header ca file

Resolves https://github.com/openshift/openshift-ansible/issues/1661


detiber:
@simon3z It's not quite ready yet, but wanted to let you know this is here for tracking purposes.

simon3z:
@detiber I assume that by the virtue of being in `openshift_master/templates` it's expanded to all masters, right?

detiber:
That is correct, if the variables are specified they are managed for all masters.

detiber:
aos-ci-test

detiber:
@sdodson @abutcher ptal, this should be about ready to go now.

abutcher:
lgtm

PR #1724, opened 2016-04-06
Bug 1324371: needn't install atomic-openshift packages on nfs server

abutcher:
Removes unnecessary openshift_common dep from openshift_storage_nfs.

https://bugzilla.redhat.com/show_bug.cgi?id=1324371

PR #1733, opened 2016-04-08
Fix hosted facts

abutcher:
I busted these with #1724. I think a better approach is to separate hosted facts into a new role so that we can depend on `openshift_hosted_facts` when we need them.

https://bugzilla.redhat.com/show_bug.cgi?id=1324371

### DIFF
--- a/roles/openshift_hosted_facts/tasks/main.yml
+++ b/roles/openshift_hosted_facts/tasks/main.yml
@@ -2,8 +2,6 @@
 - name: Set hosted facts
   openshift_facts:
     role: hosted
-    openshift_env: "{{ item | oo_openshift_env }}"
-  no_log: true
-  with_items:
-  - "{{ hostvars[inventory_hostname] }}"
-  - "{{ hostvars }}"
+    openshift_env: "{{ hostvars[inventory_hostname]
+                       | oo_merge_dicts(hostvars)
+                       | oo_openshift_env }}"
--- a/roles/openshift_hosted_facts/tasks/main.yml
+++ b/roles/openshift_hosted_facts/tasks/main.yml
@@ -1,0 +1,9 @@
+---
+- name: Set hosted facts
+  openshift_facts:
+    role: hosted
+    openshift_env: "{{ item | oo_openshift_env }}"
+  no_log: true
+  with_items:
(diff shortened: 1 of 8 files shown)


playbooks/common/openshift-cluster/additional_config.yml
```diff
@@ -32,6 +32,11 @@
   - role: os_env_extras
     when: openshift.common.deployment_type == 'online'
 
+- name: Gather hosted facts
+  hosts: oo_first_master
+  roles:
+  - role: openshift_hosted_facts
```
abutcher: This play sets up hosted facts for the following play. An alternative would be to add hosted facts as a dep to the `openshift_master` role.

detiber: I think I like the idea of making it a dependency of the openshift_master role

abutcher: Moved to dep

abutcher:
aos-ci-test

roles/openshift_hosted_facts/tasks/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Set hosted facts
+  openshift_facts:
+    role: hosted
+    openshift_env: "{{ item | oo_openshift_env }}"
+  no_log: true
+  with_items:
+  - "{{ hostvars[inventory_hostname] }}"
+  - "{{ hostvars }}"
```
abutcher: Noting that I updated this to line up with how we collect cloudprovider facts. Top level hostvars must be checked for anyone consuming the roles directly and passing non-inventory vars.

detiber: Is it possible to do a union of hostvars and hostvars[inventory_hostname] instead of the loop?

abutcher: We could merge em

abutcher: Added a simple `oo_merge_dicts`.

detiber:
:+1: pending test results

abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
@brenton :shipit:

PR #1735, opened 2016-04-08
a-o-i: Support openshift_image_tag

smunilla:
Somewhat hacky solution to let users install containerized 3.1.
A more flexible version will be forthcoming once we settle tagging
issues.


smunilla:
@brenton

brenton:
@detiber, fyi, I talked to sam about simply hardcoding v3.1.1.6 in the image tag for now for 3.1 installs.  We don't have any plans to update those tags and we need to work out a more testable QE handoff before we change how tagging works (meaning, start tagging things as v3.1).

brenton:
If anything, this will work and it won't regress anything.  We didn't test the tagging images as v3.1 correctly I could see it breaking minor and maybe even major upgrades.

detiber:
lgtm

PR #1736, opened 2016-04-08
update bootstrap-fedora to include python2-firewall for F24+

maxamillion:


detiber:
lgtm @brenton

PR #1740, opened 2016-04-11
Fix a docker-storage sysconfig bug.

dgoodwin:
On behalf of @sdodson we noticed a bug here, docker_check register was being used twice, the first stat was checking the wrong file and using the same register name when it shouldn't.


sdodson:
This will conflict with #1718 it looks like but I'm +1 on going ahead with getting this in now unless @brenton was about to merge his.

PR #1742, opened 2016-04-12
Fixes For Docker Log Options

dgoodwin:
Legacy cli_docker_log_options (from bin/cluster) was not working due to use of an outdated host group that was since refactored.

Previously we expected openshift_docker_log_options to be a comma separated list, but you'd never really know this without documentation so instead we moved to an explicit JSON list which will make this very clear in examples and docs.

NOTE: This will break backward compat with cli_docker_log_options or anyone using custom docker log options in the wild. 

@detiber keep an eye on this just to make sure I understood you correctly that this was safe. ^^

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -56,6 +56,13 @@ def migrate_docker_facts(facts):
     if 'node' in facts and 'portal_net' in facts['node']:
         facts['docker']['hosted_registry_insecure'] = True
         facts['docker']['hosted_registry_network'] = facts['node'].pop('portal_net')
+
+    # log_options was originally meant to be a comma separated string, but
+    # we now prefer an actual list, with backward compatability:
+    if 'log_options' in facts['docker'] and \
+            isinstance(facts['docker']['log_options'], basestring):
+        facts['docker']['log_options'] = facts['docker']['log_options'].split(",")
+
     return facts
 
 # TODO: We should add a generic migration function that takes source and destination
@@ -1852,6 +1859,10 @@ class OpenShiftFacts(object):
                     if isinstance(val, basestring):
                         val = [x.strip() for x in val.split(',')]
                     new_local_facts['docker'][key] = list(set(val) - set(['']))
+            # Convert legacy log_options comma sep string to a list if present:
+            if 'log_options' in new_local_facts['docker'] and \
+                    isinstance(new_local_facts['docker']['log_options'], basestring):
+                new_local_facts['docker']['log_options'] = new_local_facts['docker']['log_options'].split(',')
 
         new_local_facts = self.remove_empty_facts(new_local_facts)
 
@@ -1875,7 +1886,7 @@ class OpenShiftFacts(object):
             if isinstance(facts[fact], dict):
                 facts[fact] = self.remove_empty_facts(facts[fact])
             else:
-                if value == "" or value is None:
+                if value == "" or value == [""] or value is None:
                     facts_to_remove.append(fact)
         for fact in facts_to_remove:
             del facts[fact]
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -259,8 +259,11 @@ class FilterModule(object):
 
     @staticmethod
     def oo_split(string, separator=','):
-        """ This splits the input string into a list
+        """ This splits the input string into a list. If the input string is
+        already a list we will return it as is.
         """
+        if isinstance(string, list):
+            return string
         return string.split(separator)
 
     @staticmethod
--- a/inventory/byo/hosts.aep.example
+++ b/inventory/byo/hosts.aep.example
@@ -75,7 +75,7 @@ deployment_type=atomic-enterprise
 #openshift_docker_options="-l warn --ipv6=false"
 # Deprecated methods to set --log-driver and --log-opts flags, use openshift_docker_options instead
 #openshift_docker_log_driver=json
-#openshift_docker_log_options="tag=mailer"
+#openshift_docker_log_options=["tag=mailer"]
 
 # Alternate image format string. If you're not modifying the format string and
 # only need to inject your own registry you may want to consider
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -76,7 +76,7 @@ deployment_type=origin
 #openshift_docker_options="-l warn --ipv6=false"
 # Deprecated methods to set --log-driver and --log-opts flags, use openshift_docker_options instead
 #openshift_docker_log_driver=json
-#openshift_docker_log_options="tag=mailer"
+#openshift_docker_log_options=["tag=mailer"]
 
 # Alternate image format string. If you're not modifying the format string and
 # only need to inject your own registry you may want to consider
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -75,7 +75,7 @@ deployment_type=openshift-enterprise
 #openshift_docker_options="-l warn --ipv6=false"
 # Deprecated methods to set --log-driver and --log-opts flags, use openshift_docker_options instead
 #openshift_docker_log_driver=json
-#openshift_docker_log_options="tag=mailer"
+#openshift_docker_log_options=["tag=mailer"]
 
 
 # Alternate image format string. If you're not modifying the format string and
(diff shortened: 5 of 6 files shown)


detiber:
@lhuard1A just wanted to give you a heads up about this change coming up.

detiber:
@dgoodwin are we already persisting these values as an array? If not, then we'll need to handle a migration for existing values.

detiber:
@dgoodwin alternatively, I'm now starting to wonder if we might want to check for a comma (or space) separated list and convert it to an array explicitly within openshift_facts...

dgoodwin:
@detiber this is explicitly what I was worried about regarding my interpretation of what you said, I thought you were saying we don't need to worry about supporting legacy plain string values, or converting, and that sounded suspicious. 

I will get this updated to handle upgrading old variants of the fact.

detiber:
@dgoodwin hah, you understood correctly, I just kind of changed my mind :smile:

lhuard1A:
Thanks for the notification @detiber !

dgoodwin:
@detiber ok I updated by making oo_split filter be tolerant when the incoming value is already a list. I had it as a separate filter first but this felt a little better, and means we can support explicit lists for anything that once was intended to contain a comma separated string. The two can be used interchangeably now.

dgoodwin:
aos-ci-test

detiber:
@dgoodwin I do like the filter change. However, we'll probably want to do something similar within openshift_facts so that we handle the conversion when persisting/setting the fact values as well.

dgoodwin:
Updated with a facts upgrade as well, tested by installing with another branch, verifying a string log_options fact in /etc/ansible/facts.d/openshift.fact, then re-running off this branch and seeing it convert to a list.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -56,6 +56,13 @@ def migrate_docker_facts(facts):
     if 'node' in facts and 'portal_net' in facts['node']:
         facts['docker']['hosted_registry_insecure'] = True
         facts['docker']['hosted_registry_network'] = facts['node'].pop('portal_net')
+
+    # log_options was originally meant to be a comma separated string, but
+    # we now prefer an actual list, with backward compatability:
+    if 'log_options' in facts['docker'] and \
+            isinstance(facts['docker']['log_options'], basestring):
+        facts['docker']['log_options'] = facts['docker']['log_options'].split(",")
```
detiber: Looks good for the migration use case...  You'll need to add something similar to `init_local_facts` to handle the conversion from string to array there as well to avoid persisting a string on the first time through the role.

dgoodwin: Just so I understand, would that not be caught by the use of oo_split in:

roles/docker/tasks/main.yml:      {% if docker_log_options is defined %} {{ docker_log_options |  oo_split() | oo_prepend_strings_in_list('--log-opt ') | join(' ')}}{% endif %}

I thought that would be the source for this fact if it wasn't already set.

detiber: @dgoodwin The openshift_docker_facts role is applied prior to the docker role, so in the case of a comma separated list:
- We would persist the comma separated list the first time openshift_docker is run
- We would pass the persisted comma separated list to the docker role as docker_log_options and convert the value using the oo_split filter
- We would convert the value from a comma separated list to an array the next time openshift_facts is run on the host

dgoodwin: init_local_facts looks like it actually called migrate_docker_facts (via migrate_local_facts), does this cover that situation? Just going to do a quick test here to see if I can expose the problem, uninstall, remove facts cache, install with log options, and make sure we get an array in the fact cache.

dgoodwin: First pass through using a comma separated string, no openshift.fact cache in place, it does seem to be coming out ok I think: "log_options": ["max-size=1M", "max-file=3"]

Let me know if I've got anything mixed up here but I think it's getting picked up via init_local_facts calling the migration code.

detiber: Are you testing immediately after the openshift_docker_facts role is run? the next time openshift_facts is run it would perform the migration. The easiest way to validate would be to add the following right after setting the value in openshift_docker_facts:
```
debug: var=openshift.docker.log_options
fail:
```

dgoodwin: Ok no it's definitely not upgraded there and I am very confused. Trying to sort it out.

dgoodwin: Ok I think I see it now, init_local_facts is calling migrate for the pre-existing local facts, but this is coming in via "facts_to_set" in this case and not a part of that migration. I have attempted to update, let me know how that looks.

detiber:
aos-ci-test

detiber:
@brenton lgtm

dgoodwin:
Some kind of a test flake.

aos-ci-test

brenton:
It might be worth investigating if it happens again.  The failure is happening right after docker options are being set.

dgoodwin:
Yikes that's not good, that was not the error in the link I hit but I see it in some of the others. Will investigate.

dgoodwin:
Very legit failure, docker_log_options is coming out as an empty string, not undefined, so I'm setting invalid options in /etc/sysconfig/docker if you DON'T specify log options. (--log-opt with nothing following)

Thank you jenkins!

dgoodwin:
Ok ready to try again, the first commit here caused a behavior change where docker_log_options went from being unset to being an empty string. The facts changes I made were then setting this as [""] if you didn't specify any docker cli options. 

Modified the facts module changes to detect an empty string and just not set it if that is found anywhere, incoming fact or in the local cache.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1774,6 +1782,11 @@ def init_local_facts(self, facts=None,
                     if isinstance(val, basestring):
                         val = [x.strip() for x in val.split(',')]
                     new_local_facts['docker'][key] = list(set(val) - set(['']))
+            # Convert legacy log_options comma sep string to a list if present:
+            if 'log_options' in new_local_facts['docker'] and \
+                    isinstance(new_local_facts['docker']['log_options'], basestring) and \
+                    new_local_facts['docker']['log_options'] != '':
```
detiber: Is this a new regression (from a separate PR)? Previously we threw away empty string values in addition to values that are None.

abutcher: Empty/none values are deleted right below this.

dgoodwin: It's happening because in this case it's a split on an empty string, which ends up returning [""]. Indeed those empty strings do get cleared out right below, I can push the entire "docker" block above down below the cleanup loop, or I can make a new "docker" block below just for this, or we keep it as is with an extra empty string check.

detiber: I'm remembering an issue like this with the docker registry values as well..  maybe we need to add [""] to the things to delete below as well.

dgoodwin:
Updated to remove [""] facts as well as empty string + None. Dropped one of the empty string checks from my new additions but kept the one for the system's fact migration because why not.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -56,6 +56,14 @@ def migrate_docker_facts(facts):
     if 'node' in facts and 'portal_net' in facts['node']:
         facts['docker']['hosted_registry_insecure'] = True
         facts['docker']['hosted_registry_network'] = facts['node'].pop('portal_net')
+
+    # log_options was originally meant to be a comma separated string, but
+    # we now prefer an actual list, with backward compatability:
+    if 'log_options' in facts['docker'] and \
+            isinstance(facts['docker']['log_options'], basestring) and \
+            facts['docker']['log_options'] != '':
```
detiber: Is this test for `!= ''` still needed with the change for removing [''] as an empty value?

dgoodwin: This is on the client side fact cache, it's probably unlikely it could ever slip through but I left it just in case. I will go remove...

dgoodwin:
Updated and rebased on master, moved the [""] into Andrews new function.

detiber:
aos-ci-test

detiber:
lgtm pending tests

dgoodwin:
Should be all set @brenton

PR #1746, opened 2016-04-13
The openshift_docker role must set the version facts for containerize…

brenton:
…d installs

QE found that for fresh installs we were basing the docker version facts of the
images that could be pulled prior to configuring /etc/sysconfig/docker.  This
is an edge case but something we need to fix.


brenton:
aos-ci-test

brenton:
aos-ci-test

roles/openshift_docker/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+- set_fact:
+    l_image_tag: "{{ cli_image_version.stdout_lines[0].split(' ')[1].split('-')[0] }}"
+  when: openshift.common.is_containerized | bool and openshift_image_tag is not defined
+
+- set_fact:
+    l_image_tag: "{{ openshift_image_tag }}"
+  when: openshift.common.is_containerized | bool and openshift_image_tag is defined
+
+- debug: var=openshift
+- debug: var=l_image_tag
```
brenton: I'll make sure to remove this before merging.

roles/openshift_docker/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
...
+- name: Set post docker install facts
+  openshift_facts:
+    role: "{{ item.role }}"
+    local_facts: "{{ item.local_facts }}"
+  with_items:
+  - role: docker
+    local_facts:
+      openshift_image_tag: "{{ l_image_tag }}"
+      openshift_version: "{{ l_image_tag | oo_image_tag_to_rpm_version }}"
+  when: openshift.common.is_containerized | bool
```
brenton: @detiber  This is failing for RPM installs and it doesn't make sense to me.  Even the debug lines I have above are being skipped.  It seems like something is screwed up with the 'openshift' variables for RPM installs.

brenton: It really seems like it's running `l_image_tag | oo_image_tag_to_rpm_version` even though the task is going to be skipped.

brenton:
aos-ci-test

roles/openshift_docker/tasks/main.yml
brenton: Done.

brenton:
aoc-ci-test

brenton:
aos-ci-test

roles/openshift_cli/defaults/main.yml
```diff
@@ -1,2 +1,2 @@
 ---
-openshift_version: "{{ openshift_image_tag | default(openshift.common.image_tag) | default('') }}"
+openshift_version: "{{ openshift_image_tag | default(openshift.docker.openshift_image_tag) | default('') }}"
```
detiber: I actually really like this moving under the 'docker' role, do we need to migrate the existing common.image_tag value to docker.openshift_image_tag in openshift_facts though?

roles/openshift_docker/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
+---
+# It's important that we don't explicitly pull this image here.  Otherwise we
+# could result in upgrading a preinstalled environment.  We'll have to set
+# openshift_image_tag correctly for upgrades.
+- name: Set version when containerized
+  command: >
+    docker run --rm {{ openshift.common.cli_image }}:latest version
+  register: cli_image_version
+  when: openshift.common.is_containerized | bool and openshift_image_tag is not defined
```
detiber: Shouldn't this value be set in openshift_facts and use the facts here? I hate to rely just on passed in values.

roles/openshift_master/defaults/main.yml
```diff
@@ -29,4 +29,4 @@ os_firewall_deny:
 - service: former etcd peer port
   port: 7001/tcp
 
-openshift_version: "{{ openshift_pkg_version | default(openshift_image_tag) | default(openshift.common.image_tag) | default('') }}"
+openshift_version: "{{ openshift_pkg_version | default(openshift_image_tag) | default(openshift.docker.openshift_image_tag) | default('') }}"
```
detiber: I think these defaults actually need to be nested and not chained..   for example if openshift_image_tag is not defined, I would expect this to throw an error and not use the value of the next default filter.

roles/openshift_cli/defaults/main.yml
detiber: similar thing here with the default filters needing to be nested rather than chained.

roles/openshift_docker/tasks/main.yml
brenton: Are you talking about is_containerized?

roles/openshift_cli/defaults/main.yml
brenton: I don't think it existed prior to 3.2 if that's what you're asking.

roles/openshift_docker/tasks/main.yml
detiber: I was referring to openshift_image_tag

roles/openshift_master/defaults/main.yml
brenton: openshift_pkg_version and it's containerized counterpart openshift_image_tag are optional overrides.  They won't be set normally.

roles/openshift_cli/defaults/main.yml
detiber: Good point. As long as we haven't shipped a version that uses that fact, it should be fine to move it without migration.

roles/openshift_master/defaults/main.yml
detiber: right, what I am suggesting is this:
```
openshift_version: "{{ openshift_pkg_version | default(openshift_image_tag | default(openshift.docker.openshift_image_tag | default(''))) }}"
```

brenton: ok, I'll have to try that out.  I feel like I tested the current defaulting but I can't say I know what default is supposed to do if you pass it something that isn't defined.

roles/openshift_docker/tasks/main.yml
brenton: I didn't want to do anything with openshift_image_tag or image tags in openshift facts simply because the only way it can be computed is post docker install.  And I thought it made since to have the place where it was computed close to the place that handles the overriding.  I ripped all the image_tag stuff out of openshift facts.  I think soon we need to fix how openshift_version is used all over the place to be a canonical format and not have to use image_tag_to_rpm_version all over the place.  We have a bit of a mess right now.

detiber: gotcha, makes sense.

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -49,7 +49,7 @@
   when: not openshift.common.is_containerized | bool
 
 - set_fact:
-    l_common_version: "{{ openshift.common.image_tag | default('0.0', True) | oo_image_tag_to_rpm_version }}"
+    l_common_version: "{{ openshift_version | default('0.0', True) | oo_image_tag_to_rpm_version }}"
```
brenton: @detiber, in the end I created a new defaults file for openshift_docker_facts.  It should allow admins to override with the value as well as automatically set it based on openshift.docker.openshift_image_tag.

brenton:
aos-ci-test

brenton:
aos-ci-test

brenton:
aos-ci-test

brenton:
@detiber, we walked through this at scrum and decided to merge it.  One feedback item was the fact that the etcd role should only depend on the openshift_docker role if it's running containerized.  Andrew is going to handle that since it's related to another refactor he's working on.

PR #1749, opened 2016-04-13
Fix issues when docker is not pre-installed.

dgoodwin:
A docker run command was recently guarded to only be run when docker is
installed, this change does the same for an additional docker exec.

Adjusted some logic that might make for inaccurate openshift version
reporting when containerized is true, but for some reason openshift is
installed on the main OS.

Fixed an additional failure where a recent fix to reload the docker
service would run after the docker rpm was just installed, and the
service is not yet loaded. Due to it's nature as a workaround, it does
not look like it can be moved to a better location, so ignoring errors
appears to be the best solution.

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -28,6 +28,8 @@
 # seems a package install/upgrade/downgrade has rebooted docker and crashed it.
 - name: Reset docker service state
   command: systemctl reset-failed docker.service
+  # This will error if the package was just installed and the service is not loaded.
+  ignore_errors: yes
 
 - name: enable and start the docker service
   service:
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1066,36 +1066,40 @@ def get_openshift_version(facts, cli_image=None):
         if 'version' in facts['common'] and facts['common']['version'] is not None:
             return facts['common']['version']
 
-    if os.path.isfile('/usr/bin/openshift'):
-        _, output, _ = module.run_command(['/usr/bin/openshift', 'version'])
-        version = parse_openshift_version(output)
-
+    # If this is a containerized host, we need to get the openshift version
+    # from within a relevant container. It is possible that docker is not yet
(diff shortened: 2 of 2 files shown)


roles/docker/tasks/main.yml
```diff
@@ -28,6 +28,8 @@
 # seems a package install/upgrade/downgrade has rebooted docker and crashed it.
 - name: Reset docker service state
   command: systemctl reset-failed docker.service
+  # This will error if the package was just installed and the service is not loaded.
+  ignore_errors: yes
```
detiber: Is it better to ignore errors here, or to query if it is in a failed state first?

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1066,36 +1066,40 @@ def get_openshift_version(facts, cli_image=None):
         if 'version' in facts['common'] and facts['common']['version'] is not None:
             return facts['common']['version']
 
-    if os.path.isfile('/usr/bin/openshift'):
-        _, output, _ = module.run_command(['/usr/bin/openshift', 'version'])
-        version = parse_openshift_version(output)
-
+    # If this is a containerized host, we need to get the openshift version
```
detiber: These changes may have some overlap with @brenton's PR here: https://github.com/openshift/openshift-ansible/pull/1746

dgoodwin: Yeah looks like a lot of overlap. I'm going to go test running without docker pre-installed on @brenton 's PR and I'll send over a patch if anything is still missing there.

brenton: I'm sure there are still problems on my branch related to docker not being installed.  I think we'll need the refactorings in my PR however late yesterday I noticed they break the 3.2 upgrade fairly severely because of how we're calculating the version of docker to install.

dgoodwin: FWIW your branch actually passed an install without docker there for me. This PR may be unnecessary so I am going to close for now.

dgoodwin:
May be

PR #1751, opened 2016-04-14
fix systemctl reset-failed docker

fridim:
This commit fix the playbook when docker has never been run yet (first run, new install e.g.).
The error is :

```stderr: Failed to reset failed state of unit docker.service: Unit docker.service is not loaded.```


fridim:
Just saw the #1749   closing as duplicate

PR #1752, opened 2016-04-14
Enable docker before potentially resetting the failure

sdodson:
If docker isn't enabled we need it to be enabled, fixes

```
06:45:15 TASK: [docker | Reset docker service state] *********************************** 
06:45:15 failed: [10.8.171.105] => {"changed": true, "cmd": ["systemctl", "reset-failed", "docker.service"], "delta": "0:00:00.015313", "end": "2016-04-14 10:54:52.908223", "rc": 1, "start": "2016-04-14 10:54:52.892910", "warnings": []}
06:45:15 stderr: Failed to reset failed state of unit docker.service: Unit docker.service is not loaded.
```

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -24,12 +24,21 @@
   action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version is defined and docker_version != '' else '' }} state=present"
   when: not openshift.common.is_atomic | bool and not docker_version_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare(docker_version, 'lt')
 
+# Enable docker and record if it was changed, if it was changed then we have no
+# need to reset-failed.
+- name: Enable the docker service
+  service:
+    name: docker
+    enabled: yes
+  register: docker_enabled
+
 # We're getting ready to start docker.  This is a workaround for cases where it
 # seems a package install/upgrade/downgrade has rebooted docker and crashed it.
 - name: Reset docker service state
   command: systemctl reset-failed docker.service
+  when: not docker_enabled | changed
 
-- name: enable and start the docker service
+- name: Start the docker service
   service:
     name: docker
     enabled: yes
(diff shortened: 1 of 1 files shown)


sdodson:
@detiber fixes CI breakage, alternative is just to remove the resetting of the service. We added that because we were afraid that, especially when downgrading, docker may have already been in a failed state. What do you think?

sdodson:
aos-ci-test

sdodson:
@alivigni I believe this will fix the problem you've encountered

roles/docker/tasks/main.yml
```diff
@@ -24,12 +24,21 @@
   action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version is defined and docker_version != '' else '' }} state=present"
   when: not openshift.common.is_atomic | bool and not docker_version_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare(docker_version, 'lt')
 
+# Enable docker and record if it was changed, if it was changed then we have no
+# need to reset-failed.
+- name: Enable the docker service
+  service:
+    name: docker
+    enabled: yes
+  register: docker_enabled
```
detiber: Instead of enabling here, can we just test for failure and only reset the failed state if it is in a failed state?

sdodson: Ok, how's it look now? Not sure the general pattern to use here.

dgoodwin:
Bug also came in for this: https://bugzilla.redhat.com/show_bug.cgi?id=1327063

sdodson:
aos-ci-test

brenton:
@sdodson, @detiber, is this ready?

roles/docker/tasks/main.yml
```diff
@@ -24,17 +24,29 @@
...
 
-# We're getting ready to start docker.  This is a workaround for cases where it
-# seems a package install/upgrade/downgrade has rebooted docker and crashed it.
+- name: Start the docker service
+  service:
+    name: docker
+    enabled: yes
+    state: started
+  register: start_result
+  ignore_errors: yes
```
detiber: oh, I meant query the status of the unit to see if it is in a failed state.

brenton: @detiber  Since scott it out for the rest of the day would we want to commit this now since it will unblock CI and then clean it up tomorrow?

detiber: :+1:

PR #1756, opened 2016-04-15
Configurable admin username and passsword for Openshift enterprise in Nuage.

abhat:
This change is to allow users to configure admin username and password for Nuage enterprise created corresponding to the openshift cluster.


detiber:
@abutcher ptal

roles/nuage_master/vars/main.yaml
```diff
@@ -15,3 +15,6 @@ nuage_ca_master_rest_server_key: "{{ nuage_mon_rest_server_crt_dir }}/nuageMonSe
 nuage_ca_master_rest_server_crt: "{{ nuage_mon_rest_server_crt_dir }}/nuageMonServer.crt" 
 
 nuage_master_crt_dir : /usr/share/nuage-openshift-monitor
+cspadminpasswd: ''
+adminsusername: 'admin'
+adminuserpasswd: 'admin'
```
abutcher: Could we add `nuage_master_` to the front of these variable names so we can quickly know which role makes use of them?

abhat:
Updated the review with changes to address the comments from @abutcher

detiber:
lgtm @brenton 

@abhat as an aside, would it be possible to get some type of documentation added to openshift-ansible or at least a link to the Nuage documentation? I've had a few people ask me about how the integration works and would like to be able to easily point future requests to the right info.

abutcher:
aos-ci-test

abutcher:
@brenton

PR #1758, opened 2016-04-15
Fix use of JSON inventory vars with raw booleans.

dgoodwin:
Thought this was fixed in recent patch but somehow backed out a critical
change to the isinstance, we need to check for basestring to also have
this work with unicode strings.


dgoodwin:
@dak1n1 this was it, not sure how I did that, I swear I remembered this during devel but I think I backed it out somehow.

dgoodwin:
@brenton probably an easy merge here IMO.

detiber:
lgtm

dak1n1:
@dgoodwin Thanks! I'll end up using this today actually.

detiber:
aos-ci-test

abutcher:
aos-ci-test

dgoodwin:
@brenton look ok to merge this one?

PR #1761, opened 2016-04-18
openshift_image_tag changes

smunilla:
A few things to support openshift_image_tag for rpm based installations.

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -800,7 +800,7 @@ class FilterModule(object):
         """
         if not isinstance(version, basestring):
             raise errors.AnsibleFilterError(
-                "| {} failed expects a string or unicode".format(version))
+                "| {0} failed expects a string or unicode".format(version))
 
         if version.startswith("v"):
             version = version.replace("v", "")
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -799,7 +799,8 @@ class FilterModule(object):
             Ex. v3.2.0.10 -> -3.2.0.10
         """
         if not isinstance(version, basestring):
-            raise errors.AnsibleFilterError("|failed expects a string or unicode")
+            raise errors.AnsibleFilterError(
+                "| {} failed expects a string or unicode".format(version))
 
         if version.startswith("v"):
             version = version.replace("v", "")
--- a/roles/openshift_node/tasks/main.yml
+++ b/roles/openshift_node/tasks/main.yml
@@ -40,7 +40,7 @@
   when: not openshift.common.is_containerized | bool
 
 - name: Install sdn-ovs package
-  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-sdn-ovs{{ openshift_version | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
+  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-sdn-ovs{{ openshift_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
   when: openshift.common.use_openshift_sdn and not openshift.common.is_containerized | bool
 
 - name: Pull node image


roles/openshift_cli/defaults/main.yml
```diff
@@ -1,2 +1,2 @@
 ---
-openshift_version: "{{ openshift_image_tag | default(openshift.docker.openshift_image_tag | default('')) }}"
+openshift_version: "{{ openshift_image_tag | default(openshift.docker.openshift_image_tag) | default('') }}"
```
detiber: Why the switch from nested default filters? If openshift.docker.openshift_image_tag and openshift_image_tag are both undefined here you will get an error when evaluating openshift.docker.openshift_image_tag when applying the second default filter here.

roles/openshift_docker/tasks/main.yml
```diff
@@ -6,15 +6,18 @@
...
-  when: openshift.common.is_containerized is defined and openshift.common.is_containerized | bool and openshift_image_tag is not defined
+  when: openshift.common.is_containerized | bool and openshift_image_tag is not defined
 
 - set_fact:
     l_image_tag: "{{ openshift_image_tag }}"
-  when: openshift.common.is_containerized is defined and openshift.common.is_containerized | bool and openshift_image_tag is defined
+  when: openshift.common.is_containerized | bool and openshift_image_tag is defined
+
+- debug: var=openshift
+- debug: var=l_image_tag
```
detiber: These debug tasks should be removed prior to merge.

roles/openshift_docker/tasks/main.yml
```diff
@@ -24,5 +27,5 @@
   - role: docker
     local_facts:
       openshift_image_tag: "{{ l_image_tag }}"
-      openshift_version: "{{ l_image_tag if l_image_tag is defined else '' | oo_image_tag_to_rpm_version }}"
```
detiber: l_image_tag is only defined when openshift.common.is_containerized is true you probably want:

```
openshift_image_tag: "{{ l_image_tag | default(None) }}"
```

roles/openshift_master/defaults/main.yml
```diff
@@ -29,4 +29,4 @@ os_firewall_deny:
 - service: former etcd peer port
   port: 7001/tcp
 
-openshift_version: "{{ openshift_pkg_version | default(openshift_image_tag | default(openshift.docker.openshift_image_tag | default(''))) }}"
+openshift_version: "{{ openshift_pkg_version | default(openshift_image_tag) | default(openshift.docker.openshift_image_tag) | default('') }}"
```
detiber: These default filters should be nested, otherwise there will be an error for any variable used in a default filter that is undefined.

roles/openshift_node/defaults/main.yml
```diff
@@ -13,4 +13,4 @@ os_firewall_allow:
 - service: OpenShift OVS sdn
   port: 4789/udp
   when: openshift.node.use_openshift_sdn | bool
-openshift_version: "{{ openshift_pkg_version | default(openshift_image_tag | default(openshift.docker.openshift_image_tag | default(''))) }}"
+openshift_version: "{{ openshift_pkg_version | default(openshift_image_tag) | default(openshift.docker.openshift_image_tag) | default('') }}"
```
detiber: Same here for the default filters needing to be nested.

filter_plugins/oo_filters.py
```diff
@@ -799,7 +799,8 @@ def oo_image_tag_to_rpm_version(version, include_dash=False):
             Ex. v3.2.0.10 -> -3.2.0.10
         """
         if not isinstance(version, basestring):
-            raise errors.AnsibleFilterError("|failed expects a string or unicode")
+            raise errors.AnsibleFilterError(
+                "| {} failed expects a string or unicode".format(version))
```
abutcher: I think we'll need to number this placeholder for python 2.6

detiber: We *could* always drop python 2.6 support, RHEL 7 and the latest versions of OS X ship with 2.7. I don't see an issue with dropping support for RHEL 6 clients (especially since we only ship a package for RHEL 7 anyway)

roles/openshift_docker/tasks/main.yml
```diff
@@ -6,15 +6,15 @@
   command: >
     docker run --rm {{ openshift.common.cli_image }}:latest version
   register: cli_image_version
-  when: openshift.common.is_containerized is defined and openshift.common.is_containerized | bool and openshift_image_tag is not defined
+  when: openshift.common.is_containerized | bool and openshift_image_tag is not defined
```
brenton: Why was this change undone?  I'm worried this was due to a bad rebase.

roles/openshift_docker/tasks/main.yml
```diff
@@ -23,6 +23,6 @@
   with_items:
   - role: docker
     local_facts:
-      openshift_image_tag: "{{ l_image_tag }}"
-      openshift_version: "{{ l_image_tag if l_image_tag is defined else '' | oo_image_tag_to_rpm_version }}"
-  when: openshift.common.is_containerized is defined and openshift.common.is_containerized | bool
+      openshift_image_tag: "{{ l_image_tag | default(None) }}"
+      openshift_version: "{{ l_image_tag | oo_image_tag_to_rpm_version }}"
+  when: openshift.common.is_containerized | bool
```
brenton: Same question here.  If `openshift.common.is_containerized is defined` isn't need then we should remove it.  I just don't see how it's related to the bug in question.

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -49,7 +49,7 @@
   when: not openshift.common.is_containerized | bool
 
 - set_fact:
-    l_common_version: "{{ openshift_version | default('0.0', True) | oo_image_tag_to_rpm_version }}"
+    l_common_version: "{{ openshift.docker.openshift_image_tag | default('0.0', True) | oo_image_tag_to_rpm_version }}"
```
brenton: I'm pretty sure `openshift_version` is set in the defaults for the openshift_docker_facts role to be based on openshift.docker.openshift_image_tag.

roles/openshift_node/tasks/main.yml
```diff
@@ -40,7 +40,7 @@
   when: not openshift.common.is_containerized | bool
 
 - name: Install sdn-ovs package
-  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-sdn-ovs{{ openshift_version | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
+  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-sdn-ovs{{ openshift_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
```
brenton: I would expect the default for openshift_version to be handled in the defaults/main.yml for this role.

roles/openshift_docker/tasks/main.yml
smunilla: Yeah, I think this was a bad rebase. Let me see if I can un-undo it.

smunilla:
@brenton, @detiber Results of the previous bad rebase are now removed.

roles/openshift_node/tasks/main.yml
brenton: @smunilla, this what what I'm talking about https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_node/defaults/main.yml#L16

PR #1762, opened 2016-04-18
Bug 1320829 - Handle OSE 3.0 installs

brenton:


brenton:
@sdodson, @detiber, PTAL

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -56,6 +56,12 @@
     l_common_version: "{{ common_version.stdout | default('0.0', True) }}"
   when: not openshift.common.is_containerized | bool
 
+# This can be removed when OSE 3.0 is no longer supported.
+- name: Set docker version to be installed
+  set_fact:
+    docker_version: "{{ '1.8.2' }}"
+  when: " ( l_common_version | version_compare('3.2','<') and openshift.common.service_type == 'openshift' )"
```
sdodson: openshift.common.service_type != 'origin' instead? so they're mutually exclusive and we cover online?

brenton: Online should already be covered since we're talking about service_types here.  I did update the PR based on the change we discussed before scrum.

brenton:
aos-ci-test

sdodson:
lgtm

brenton:
@dgoodwin, I think this will be the last change needed for https://bugzilla.redhat.com/show_bug.cgi?id=1320829 unless you know of something else.

brenton:
@arilivigni, for the 3.0 RPM install do you remember why we're installing the epel repos?

```
TASK: [nodes | install epel rpm] ********************************************** 
failed: [10.8.171.15] => {"changed": false, "failed": true, "rc": 1}
msg: Package at http://download.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm could not be installed
failed: [10.8.171.18] => {"changed": false, "failed": true, "rc": 1}
msg: Package at http://download.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm could not be installed
changed: [10.8.171.92]
```

arilivigni:
We always install it from when we did 3.1 so never changed it for 3.0

-== @ri ==-

arilivigni:
@brenton I think actually clint added this for the ose stuff

brenton:
@arilivigni, I can't think of any reason why we'd need epel.

arilivigni:
@brenton check with clint he added it doesn't matter to me

dgoodwin:
My testing seemed to indicate this branch worked fine without docker pre-installed, no pending issues I'm aware of.

PR #1763, opened 2016-04-18
Use 'openshift3/' as IMAGE_PREFIX for enterprise logging and metrics

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1326446

Minimal fix, still need to address upstream enterprise content source for metrics or we'll just re-introduce the problem at a later date.

### DIFF
--- a/roles/openshift_examples/files/examples/v1.2/infrastructure-templates/enterprise/metrics-deployer.yaml
+++ b/roles/openshift_examples/files/examples/v1.2/infrastructure-templates/enterprise/metrics-deployer.yaml
@@ -79,7 +79,7 @@ parameters:
 -
   description: 'Specify prefix for metrics components; e.g. for "openshift/origin-metrics-deployer:latest", set prefix "openshift/origin-"'
   name: IMAGE_PREFIX
-  value: "openshift/origin-"
+  value: "registry.access.redhat.com/openshift3/"
 -
   description: 'Specify version for metrics components; e.g. for "openshift/origin-metrics-deployer:latest", set version "latest"'
   name: IMAGE_VERSION


sdodson:
@mwringe @sosiouxme mind taking a quick look?

sosiouxme:
LGTM

sdodson:
@brenton ready to merge

mwringe:
The IMAGE_PREFIX is a template parameter that can be used when using the deployer. You wouldn't need to actually modify the template file, just specifying the parameter would be enough.

So if I create a user in docker hub called 'openshift3' and for some reason registry.access.redhat.com is down, does that mean it could potentially bring in a compromised container instead?

sdodson:
Yes, I think that's the case. I'd have to test to be sure.

detiber:
@sdodson we may want to start thinking about how to make the templates more customizable on deployment. This will be more of an issue when it comes to disconnected installs.

sdodson:
I've always assumed that people would utilize the --additional-registry flag for such environments.

brenton:
I think we're going to have move away from relying on --add-registry in a number of places actually.  This will require more of the product relying on image streams and then making sure they are using the fqdn's of the registry so that the Kubelet docker auth keyring works correctly.

sdodson:
I'll revert back to the fully qualified IMAGE_PREFIX for now.

sdodson:
@brenton reverted back to fully qualified image names, only the metrics IMAGE_PREFIX needed to be fixed.

brenton:
If this missed today's puddle I'll tag/build the package and give the brew link to QE.

PR #1765, opened 2016-04-18
a-o-i: UI additions for proxies

smunilla:


utils/src/ooinstall/cli_installer.py
```diff
@@ -562,6 +569,10 @@ def get_missing_info_from_user(oo_cfg):
         oo_cfg.settings['master_routingconfig_subdomain'] = get_master_routingconfig_subdomain()
         click.clear()
 
+    if not oo_cfg.settings.get('master_routingconfig_subdomain', None):
+        oo_cfg.settings['master_routingconfig_subdomain'] = get_proxy_hostname()
```
brenton: Isn't the `master_routingconfig_subdomain` something totally different from the proxy?  The latest docs PR shows the inventory values that should be set https://github.com/openshift/openshift-docs/pull/1607/files

For a-o-i I think we'd only want to set a single proxy hostname for all the proxy values and then allow them to set excludes.

utils/src/ooinstall/cli_installer.py
```diff
@@ -516,6 +516,13 @@ def error_if_missing_info(oo_cfg):
     if missing_info:
         sys.exit(1)
 
+def get_proxy_hostname_and_excludes():
+    message = "Specify the hostname for your proxy? (ENTER for none)"
```
brenton: I think we should put a little more information here about why they might need a proxy.  Maybe something like this:

```
If a proxy is needed to reach HTTP and HTTPS traffic please enter the name below.  This proxy will be configured by default for all processes needing to reach systems outside the cluster.  More advanced configuration is possible if using ansible directly:

<insert link to official docs>
```

PR #1766, opened 2016-04-18
Bug 1328119 - router selector not obeyed

abutcher:
* Fix router selector fact migration.
* Nodes matching selector filter wasn't addressing multiple selectors ie. `region=infra,zone=router`
* Fix empty fact removal for nested dictionaries.

@akostadinov @detiber PTAL

https://bugzilla.redhat.com/show_bug.cgi?id=1328119

### DIFF
--- a/roles/openshift_hosted_facts/tasks/main.yml
+++ b/roles/openshift_hosted_facts/tasks/main.yml
@@ -1,10 +1,4 @@
 ---
-- name: Set legacy hosted facts
-  openshift_facts:
-    role: hosted
-    openshift_env:
-      openshift_hosted_router_selector: "{{ openshift_router_selector | default(None) }}"
-
 - name: Set hosted facts
   openshift_facts:
     role: hosted
--- a/roles/openshift_hosted_facts/tasks/main.yml
+++ b/roles/openshift_hosted_facts/tasks/main.yml
@@ -1,4 +1,10 @@
 ---
+- name: Set legacy hosted facts
+  openshift_facts:
+    role: hosted
+    openshift_env:
+      openshift_hosted_router_selector: "{{ openshift_router_selector | default(None) }}"
+
 - name: Set hosted facts
   openshift_facts:
     role: hosted
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -650,7 +650,9 @@ class FilterModule(object):
 
     @staticmethod
     def oo_openshift_env(hostvars):
-        ''' Return facts which begin with "openshift_"
+        ''' Return facts which begin with "openshift_" and translate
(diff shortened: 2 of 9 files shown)


abutcher:
aos-ci-test

detiber:
@abutcher are you leaving the setting of openshift.master.router_selector in openshift_master_facts for backwards compat? If so, I think I'd rather see that handled outside of the master facts (maybe handled in the `oo_openshift_env` filter. Otherwise facts will consistently set/migrate the data if the user is using the old variable.

abutcher:
@detiber Yes, I'm leaving it where it is so that the migration can handle it in that case. My original diff for this change added an additional `oo_openshift_env` to `openshift_hosted_facts` which we can do if you prefer.

detiber:
@abutcher I'm just worried a bit about the way it functions when using a migration and setting the facts. the migration only works on persisted values, so the first time through it would be set as the old value and then it would be migrated on the second run through openshift_facts. This would also lead to the appearance of idempotency issues for subsequent runs with the old variable.

abutcher:
@detiber Definitely, changing incoming :rocket:

roles/openshift_hosted_facts/tasks/main.yml
```diff
@@ -1,4 +1,10 @@
 ---
+- name: Set legacy hosted facts
+  openshift_facts:
+    role: hosted
+    openshift_env:
+      openshift_hosted_router_selector: "{{ openshift_router_selector | default(None) }}"
```
detiber: Would it make sense to just to do something like this instead:
```
- name: Set legacy hosted facts
  openshift_facts:
    role: hosted
    openshift_env:
      openshift_hosted_router_selector: "{{ openshift_router_selector  }}"
  when: openshift_hosted_router_selector is not defined and openshift_router_selector is defined
```

abutcher: @detiber I imagined more things being tacked onto that forever... the registry is the next obvious thing

abutcher: So, one bulk legacy `openshift_facts` call rather than individual ones

detiber: yeah, I'd just like to see us limit the number of calls into openshift_facts at some point... Is the oo_openshift_env filter a better place to try to tackle this?

abutcher: openshift_env fact migrations are now occurring in the `oo_openshift_env` filter.

detiber:
aos-ci-test

detiber:
lgtm pending tests

abutcher:
@brenton

brenton:
Now with 30% more obedience!

PR #1777, opened 2016-04-19
Pod evacuation

brenton:


playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/node_upgrade.yml
```diff
@@ -0,0 +1,31 @@
...
+  vars:
+     component: "node"
+     openshift_version: "{{ openshift_pkg_version | default('') }}"
+  when: not openshift.common.is_containerized | bool
+
+- include: containerized_upgrade.yml
+  when: openshift.common.is_containerized | bool
+
+# This will restart the node
+- name: Restart openvswitch service
```
detiber: Task name here is wrong

brenton: I'm honestly not sure if it's the tasks name that is wrong or the actual action.  I'll have to investigate.

detiber:
@brenton overall looks good. We may want to do some type of a host limit though. And possibly treat nodes hosting the router/registry differently.

roles/openshift_docker/tasks/main.yml
```diff
@@ -4,7 +4,7 @@
 # openshift_image_tag correctly for upgrades.
 - name: Set version when containerized
   command: >
-    docker run --rm {{ openshift.common.cli_image }}:latest version
+    docker run --rm {{ openshift.common.cli_image }} version
```
brenton: This was needed for a docker 1.8.2 bug.  It seems like if you run <some image>:latest explicitly and the latest tag is not already pulled you hit and error about the image ID missing.  If you leave it off it's assumed and will correctly pull the image if needed.

brenton:
@detiber, PTAL.  This is really all I had time to test.

detiber:
@brenton the steps look good to me. I am still worried about having the node upgrade step be fully parallel now that we are evacuating nodes now.

Also, thinking about it...  we would probably want to check the schedulability of the node at the start of the node_update.yml step to avoid making an unschedulable master schedulable after the upgrade.

brenton:
It only makes Node schedulable `when: openshift.node.schedulable | bool`.  Am I missing something?

I agree there's some risk of running this in parallel.  Would we want to make the Node upgrade serial for now?

detiber:
@brenton indeed, but currently we can't be sure that the user didn't set the node as unschedulable manually post installation.

brenton:
That's a really good point. @dgoodwin is going to help out with this tomorrow while I'm out.

dgoodwin:
Endless trouble trying to test this today across employee sub problems, prod subscription service, AWS and local. I got through one upgrade on 2/3 of my nodes but my running pods did not redeploy and from the log it looks like the evac playbook did not run at all. Not yet sure how I did that, but working on another attempt.

sdodson:
This was the PR I mentioned during standup today that we said we wanted for 3.2.  BTW should we create priority/p# labels so we can tag issues accordingly?

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/node_upgrade.yml
```diff
@@ -0,0 +1,24 @@
+- name: Prepare for Node evacuation
+  command: >
+    {{ openshift.common.admin_binary }} manage-node {{ openshift.common.hostname | lower }} --schedulable=false
+  delegate_to: "{{ groups.oo_first_master.0 }}"
```
dgoodwin: This came out as /usr/local/bin/oadm in an rpm install, which does not exist. I am confused as to why that fact is wrong.

dgoodwin: Found it, bad inventory, disregard.

detiber:
@sdodson agree on creating the priority labels, I just haven't gotten there yet. Do you have the ability to create them?

brenton:
I didn't see how to create labels.  I think some of my access to this repository may have been removed.

brenton:
I rebased this from master btw.

detiber:
ok, I went ahead and created some labels to match the origin labels. We will still probably need to add some more, but the priorities are there at least.

dgoodwin:
Some observations from testing an rpm based upgrade, ansible log available here: 

Firstly as you can see in the logs docker is restarted prior to the node evac, effectively killing all current pods, is this expected / acceptable? 

I was casually watching the status of "oc get pods" and I noticed that shortly before the node evac, all pods had gone to Error state, with no information in the pod logs themselves. I suspect this was related to the docker restart for this reason.

Once the node evac plays were reached the pods were all re-created. I

After upgrade my registry failed to fully deploy. The registry actually came up but then the deployer later failed. this could be my environment or something more sinister:

```
(root@master1 ~) $ oc get pods
NAME                        READY     STATUS    RESTARTS   AGE
docker-registry-1-pbsrn     1/1       Running   5          6m
docker-registry-2-deploy    0/1       Error     0          4m
jenkins-1-centos7-1-9f5pu   1/1       Running   0          6m
(root@master1 ~) $ oc logs docker-registry-2-deploy
F0425 13:36:14.295063       1 deployer.go:70] couldn't get deployment default/docker-registry-2: Get https://172.30.0.1:443/api/v1/namespaces/default/replicationcontrollers/docker-registry-2: dial tcp 172.30.0.1:443: i/o timeout
(root@master1 ~) $ 
```

So depending on the answers to some of the above, we may not need to do serial upgrade if all the pods are coming down regardless, and nothings running anyhow.

brenton:
Just to be a little more precise in the terminology, Docker restarting kills the running containers.  It doesn't kill or delete any Pods.

I suspect it would be more of a refactoring than we can afford right now to properly evacuate the Nodes before restarting Docker in all cases.  We'll probably just have to live with that for now.  Let me know if you need help debugging what happened during the registry deployment.

dgoodwin:
So what do you think about dropping the serialized node upgrade and letting them all go at once? Would there be any negative impact is we accept that docker was already restarted just prior.

dgoodwin:
I know this is merged I just wanted to double check if that serial upgrade is buying us anything, that could be a noticeable problem in a larger env and I don't know as it's buying us anything if all containers are coming down beforehand anyhow. Is there a benefit to serializing it I'm unaware of?

dgoodwin:
I follow, thanks!

PR #1778, opened 2016-04-20
Arbitrary installer yaml

smunilla:
Initial build of new, more flexible installer config file format.

### DIFF
--- a/utils/src/ooinstall/cli_installer.py
+++ b/utils/src/ooinstall/cli_installer.py
@@ -10,7 +10,7 @@ import click
 from ooinstall import openshift_ansible
 from ooinstall import OOConfig
 from ooinstall.oo_config import OOConfigInvalidHostError
-from ooinstall.oo_config import Host
+from ooinstall.oo_config import Host, Role
 from ooinstall.variants import find_variant, get_variant_version_combos
 
 DEFAULT_ANSIBLE_CONFIG = '/usr/share/atomic-openshift-utils/ansible.cfg'
@@ -121,21 +121,23 @@ http://docs.openshift.com/enterprise/latest/architecture/infrastructure_componen
     click.echo(message)
 
     hosts = []
+    roles = set(['master', 'node', 'storage'])
     more_hosts = True
     num_masters = 0
     while more_hosts:
         host_props = {}
+        host_props['roles'] = []
         host_props['connect_to'] = click.prompt('Enter hostname or IP address',
                                                 value_proc=validate_prompt_hostname)
 
         if not masters_set:
             if click.confirm('Will this host be an OpenShift Master?'):
-                host_props['master'] = True
+                host_props['roles'].append('master')
                 num_masters += 1
 
                 if oo_cfg.settings['variant_version'] == '3.0':
                     masters_set = True
-        host_props['node'] = True
+        host_props['roles'].append('node')
 
         host_props['containerized'] = False
         if oo_cfg.settings['variant_version'] != '3.0':
@@ -152,6 +154,7 @@ http://docs.openshift.com/enterprise/latest/architecture/infrastructure_componen
 
         hosts.append(host)
 
+
         if print_summary:
             print_installation_summary(hosts, oo_cfg.settings['variant_version'])
 
@@ -163,11 +166,12 @@ http://docs.openshift.com/enterprise/latest/architecture/infrastructure_componen
 
     if num_masters >= 3:
         collect_master_lb(hosts)
+        roles.add('master_lb')
 
     if not existing_env:
         collect_storage_host(hosts)
 
-    return hosts
+    return hosts, roles
 
 
 def print_installation_summary(hosts, version=None):
@@ -184,9 +188,9 @@ def print_installation_summary(hosts, version=None):
     for host in hosts:
         print_host_summary(hosts, host)
 
-    masters = [host for host in hosts if host.master]
-    nodes = [host for host in hosts if host.node]
-    dedicated_nodes = [host for host in hosts if host.node and not host.master]
+    masters = [host for host in hosts if host.is_master()]
+    nodes = [host for host in hosts if host.is_node()]
+    dedicated_nodes = [host for host in hosts if host.is_node() and not host.is_master()]
     click.echo('')
     click.echo('Total OpenShift Masters: %s' % len(masters))
     click.echo('Total OpenShift Nodes: %s' % len(nodes))
@@ -226,26 +230,26 @@ deployment."""
 
 def print_host_summary(all_hosts, host):
     click.echo("- %s" % host.connect_to)
-    if host.master:
+    if host.is_master():
         click.echo("  - OpenShift Master")
-    if host.node:
+    if host.is_node():
         if host.is_dedicated_node():
             click.echo("  - OpenShift Node (Dedicated)")
         elif host.is_schedulable_node(all_hosts):
             click.echo("  - OpenShift Node")
         else:
             click.echo("  - OpenShift Node (Unscheduled)")
-    if host.master_lb:
+    if host.is_master_lb():
         if host.preconfigured:
             click.echo("  - Load Balancer (Preconfigured)")
         else:
             click.echo("  - Load Balancer (HAProxy)")
-    if host.master:
+    if host.is_master():
         if host.is_etcd_member(all_hosts):
             click.echo("  - Etcd Member")
         else:
             click.echo("  - Etcd (Embedded)")
-    if host.storage:
+    if host.is_storage():
         click.echo("  - Storage")
 
 
@@ -279,7 +283,7 @@ hostname.
 
         # Make sure this host wasn't already specified:
         for host in hosts:
-            if host.connect_to == hostname and (host.master or host.node):
+            if host.connect_to == hostname and (host.is_master() or host.is_node()):
                 raise click.BadParameter('Cannot re-use "%s" as a load balancer, '
                                          'please specify a separate host' % hostname)
         return hostname
@@ -289,9 +293,7 @@ hostname.
     install_haproxy = \
         click.confirm('Should the reference haproxy load balancer be installed on this host?')
     host_props['preconfigured'] = not install_haproxy
-    host_props['master'] = False
-    host_props['node'] = False
-    host_props['master_lb'] = True
+    host_props['roles'] = ['master_lb']
     master_lb = Host(**host_props)
     hosts.append(master_lb)
 
@@ -309,14 +311,14 @@ Note: Containerized storage hosts are not currently supported.
     click.echo(message)
     host_props = {}
 
-    first_master = next(host for host in hosts if host.master)
+    first_master = next(host for host in hosts if host.is_master())
 
     hostname_or_ip = click.prompt('Enter hostname or IP address',
                                             value_proc=validate_prompt_hostname,
                                             default=first_master.connect_to)
     existing, existing_host = is_host_already_node_or_master(hostname_or_ip, hosts)
-    if existing and existing_host.node:
-        existing_host.storage = True
+    if existing and existing_host.is_node():
+        existing_host.roles.append('storage')
     else:
         host_props['connect_to'] = hostname_or_ip
         host_props['preconfigured'] = False
@@ -331,14 +333,14 @@ def is_host_already_node_or_master(hostname, hosts):
     existing_host = None
 
     for host in hosts:
-        if host.connect_to == hostname and (host.master or host.node):
+        if host.connect_to == hostname and (host.is_master() or host.is_node()):
             is_existing = True
             existing_host = host
 
     return is_existing, existing_host
 
 def confirm_hosts_facts(oo_cfg, callback_facts):
-    hosts = oo_cfg.hosts
+    hosts = oo_cfg.deployment.hosts
     click.clear()
     message = """
 A list of the facts gathered from the provided hosts follows. Because it is
@@ -414,19 +416,20 @@ Edit %s with the desired values and run `atomic-openshift-installer --unattended
 
 def check_hosts_config(oo_cfg, unattended):
     click.clear()
-    masters = [host for host in oo_cfg.hosts if host.master]
+    masters = [host for host in oo_cfg.deployment.hosts if host.is_master()]
 
     if len(masters) == 2:
         click.echo("A minimum of 3 Masters are required for HA deployments.")
         sys.exit(1)
 
     if len(masters) > 1:
-        master_lb = [host for host in oo_cfg.hosts if host.master_lb]
+        master_lb = [host for host in oo_cfg.deployment.hosts if host.is_master_lb()]
+
         if len(master_lb) > 1:
             click.echo('ERROR: More than one Master load balancer specified. Only one is allowed.')
             sys.exit(1)
         elif len(master_lb) == 1:
-            if master_lb[0].master or master_lb[0].node:
+            if master_lb[0].is_master() or master_lb[0].is_node():
                 click.echo('ERROR: The Master load balancer is configured as a master or node. ' \
                            'Please correct this.')
                 sys.exit(1)
@@ -440,7 +443,8 @@ https://docs.openshift.org/latest/install_config/install/advanced_install.html#m
             click.echo(message)
             sys.exit(1)
 
-    dedicated_nodes = [host for host in oo_cfg.hosts if host.node and not host.master]
+    dedicated_nodes = [host for host in oo_cfg.deployment.hosts \
+                        if host.is_node() and not host.is_master()]
     if len(dedicated_nodes) == 0:
         message = """
 WARNING: No dedicated Nodes specified. By default, colocated Masters have
@@ -481,7 +485,7 @@ def confirm_continue(message):
 
 def error_if_missing_info(oo_cfg):
     missing_info = False
-    if not oo_cfg.hosts:
+    if not oo_cfg.deployment.hosts:
         missing_info = True
         click.echo('For unattended installs, hosts must be specified on the '
                    'command line or in the config file: %s' % oo_cfg.config_path)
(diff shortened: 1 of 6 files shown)


detiber:
Example config file is here: https://gist.github.com/detiber/da042623b26522fcd5767825eafe97a0

utils/src/ooinstall/cli_installer.py
```diff
@@ -164,8 +165,8 @@ def collect_hosts(oo_cfg, existing_env=False, masters_set=False, print_summary=T
             more_hosts = click.confirm('Do you want to add additional hosts?')
 
     if num_masters == 1:
-        master = next((host for host in hosts if host.master), None)
-        master.storage = True
+        master = next((host for host in hosts if host.is_master()), None)
+        master['roles'].append('storage')
```
detiber: Might want to have the ability to split out storage and lb hosts separate from the master, especially since they are not currently containerized.

utils/src/ooinstall/oo_config.py
```diff
@@ -73,41 +61,75 @@ def __repr__(self):
...
                 d[prop] = getattr(self, prop)
         return d
 
+    def is_master(self):
+        return 'master' in self.roles
+
+    def is_node(self):
+        return 'node' in self.roles
+
     def is_etcd_member(self, all_hosts):
```
detiber: This should check for etcd in roles, and we should handle the default co-location by adding both roles to the master hosts if no etcd role is defined.

utils/src/ooinstall/oo_config.py
```diff
@@ -73,41 +61,75 @@ def __repr__(self):
...
-        if not self.master:
+        if not self.is_master():
             return False
-        masters = [host for host in all_hosts if host.master]
+        masters = [host for host in all_hosts if host.is_master()]
         if len(masters) > 1:
             return True
         return False
 
     def is_dedicated_node(self):
```
detiber: Not directly related to this PR, but can't say I'm a fan of using 'dedicated' here.

detiber:
I don't want to block this change on the following items, but some things that we might want to address through follow ups::
- I think it would be good to break out OOConfig into different modules
  - I'm thinking that Role should be a Python Mixin rather than a class and there should be a Mixin for each role that extends the Role Mixin.
    - each host can then have the correct role mixins added depending on the role configuration.
    - Host can introspect itself to determine which roles are applied an call the appropriate methods from the role mixin.
- legacy config migration should probably be done based on the absence of the version field.
- Switch from using a single inventory file to generating an inventory based on the hosts/groups only and generating host_vars and group_vars for the variables (though we might want to consider still generating a combined inventory for users to take elsewhere).

utils/src/ooinstall/oo_config.py
```diff
@@ -73,41 +61,75 @@ def __repr__(self):
...
         """ Will this host be a dedicated node. (not a master) """
-        return self.node and not self.master
+        return self.is_node() and not self.is_master()
 
     def is_schedulable_node(self, all_hosts):
```
detiber: this method doesn't appear to be called anywhere

dgoodwin:
From the example config, should hosts and roles be nested under deployment? The description of the two looks a bit like they should all be at the same level. (deployment = global to the deployment, hosts being the opposite of that)

dgoodwin:
Might be helpful to get an initial test in place to help speed up the devel/test cycle, and make the WIP PR a little easier to just view and see how everything fits together.

detiber:
@dgoodwin In my mind nodes and roles both belong to a deployment, but I'm not married to the idea.

dgoodwin:
The description of it in the sample file might be what's throwing me. If it's a container for global'ish vars, and there's just one deployment per cfg file, I would expect the hosts/roles to not be nested beneath it and the file itself is the container for the overall deployment.

If however it's models something larger than just some global vars, and/or there might be multiple deployments in the file, then nesting it def makes more sense.

No major objection either way just some thoughts.

smunilla:
@detiber This is working well enough to at least attempt run throughs with now

utils/src/ooinstall/cli_installer.py
```diff
@@ -120,24 +120,26 @@ def collect_hosts(oo_cfg, existing_env=False, masters_set=False, print_summary=T
     click.echo(message)
 
     hosts = []
+    roles = set(['master', 'node', 'storage'])
```
detiber: Should `master_lb` be added to this list as well?

utils/src/ooinstall/cli_installer.py
```diff
@@ -576,29 +594,32 @@ def get_missing_info_from_user(oo_cfg):
...
+        oo_cfg.deployment.variables['ansible_ssh_user'] = get_ansible_ssh_user()
         click.clear()
 
-    if oo_cfg.settings.get('variant', '') == '':
+    if not oo_cfg.deployment.variables.get('variant', ''):
         variant, version = get_variant_and_version()
-        oo_cfg.settings['variant'] = variant.name
-        oo_cfg.settings['variant_version'] = version.name
+        oo_cfg.deployment.variables['variant'] = variant.name
+        oo_cfg.deployment.variables['variant_version'] = version.name
```
detiber: variant and variant_version should not be attributes of the deployment. In the example config they are top level attributes of the config file since they drive the installer workflow and not deployment configuration.

utils/src/ooinstall/cli_installer.py
```diff
@@ -576,29 +594,32 @@ def get_missing_info_from_user(oo_cfg):
...
+        oo_cfg.deployment.hosts, roles = collect_hosts(oo_cfg)
+
+        for role in roles:
+            oo_cfg.deployment.roles.append(Role(name=role))
         click.clear()
 
-    if not oo_cfg.settings.get('master_routingconfig_subdomain', None):
-        oo_cfg.settings['master_routingconfig_subdomain'] = get_master_routingconfig_subdomain()
+    if not oo_cfg.deployment.variables.get('master_routingconfig_subdomain', None):
+        oo_cfg.deployment.variables['master_routingconfig_subdomain'] = get_master_routingconfig_subdomain()
```
detiber: master_routingconfig_subdomain should be under the master role rather than the deployment.

Also, there should also be a short name mapping for it.

utils/src/ooinstall/cli_installer.py
```diff
@@ -773,23 +794,23 @@ def cli(ctx, unattended, configuration, ansible_playbook_directory, ansible_conf
 
     # If no playbook dir on the CLI, check the config:
     if not ansible_playbook_directory:
-        ansible_playbook_directory = oo_cfg.settings.get('ansible_playbook_directory', '')
+        ansible_playbook_directory = oo_cfg.deployment.variables.get('ansible_playbook_directory', '')
```
detiber: ansible_playbook_directory should be at the root of the config rather than under deployment.

utils/src/ooinstall/cli_installer.py
```diff
@@ -773,23 +794,23 @@ def cli(ctx, unattended, configuration, ansible_playbook_directory, ansible_conf
...
-    oo_cfg.settings['ansible_playbook_directory'] = ansible_playbook_directory
+    oo_cfg.deployment.variables['ansible_playbook_directory'] = ansible_playbook_directory
     oo_cfg.ansible_playbook_directory = ansible_playbook_directory
     ctx.obj['ansible_playbook_directory'] = ansible_playbook_directory
 
     if ctx.obj['ansible_config']:
-        oo_cfg.settings['ansible_config'] = ctx.obj['ansible_config']
-    elif 'ansible_config' not in oo_cfg.settings and \
+        oo_cfg.deployment.variables['ansible_config'] = ctx.obj['ansible_config']
+    elif 'ansible_config' not in oo_cfg.deployment.variables and \
```
detiber: ansible_config and ansible_logpath should also be root level and not under deployments.

utils/src/ooinstall/oo_config.py
```diff
@@ -40,31 +43,24 @@ def __init__(self, **kwargs):
         self.public_ip = kwargs.get('public_ip', None)
         self.public_hostname = kwargs.get('public_hostname', None)
         self.connect_to = kwargs.get('connect_to', None)
+
         self.preconfigured = kwargs.get('preconfigured', None)
+        self.schedulable = kwargs.get('preconfigured', None)
```
detiber: copypasta error? should this be` kwargs.get('schedulable', None` instead?

utils/src/ooinstall/openshift_ansible.py
```diff
@@ -11,30 +11,46 @@
 
 CFG = None
 
+ROLES_TO_GROUPS_MAP = {
+    'master': 'masters',
+    'node': 'nodes',
+    'storage': 'nfs'
```
detiber: should master_lb be in here as well?

utils/src/ooinstall/openshift_ansible.py
```diff
@@ -11,30 +11,46 @@
...
+}
+
+VARIABLES_MAP = {
```
detiber: Having a single mapping for variables seems a bit odd. I would  think there should be at minimum:
- mapping dict for deployment level vars
- mapping dict for role level vars
- mapping dict for host level vars

And potentially have sets of mappings that could appear at more than one level as well.

utils/src/ooinstall/openshift_ansible.py
```diff
@@ -11,30 +11,46 @@
...
+    'ansible_ssh_user':'ansible_ssh_user',
+    'ansible_config':'ansible_config',
+    'ansible_log_path':'ansible_log_path',
```
detiber: Any real reason to have the above ansible_* vars as part of a mapping, since the names are the same for input and output?

utils/src/ooinstall/cli_installer.py
smunilla: I don't think so. All-in-one and Single Master installations won't have it.

utils/src/ooinstall/openshift_ansible.py
smunilla: Mostly I was putting them there to indicate that they are "mapped variables" and thus supported. This would let us later know if they passed in unmapped variables and we could print messages or trigger other behavior based on that.

But I'm not overly attached to the idea.

utils/src/ooinstall/cli_installer.py
detiber: As long as the additional roles get added as needed (or when pulled in from config), then no objections.  I made my comments as I read top-down.

utils/src/ooinstall/openshift_ansible.py
detiber: Gotcha, maybe a separate list for supported unmapped vars then?

utils/src/ooinstall/openshift_ansible.py
detiber: We need to take into account the etcd group here as well. Maybe ROLES_TO_GROUPS_MAP should be dict of role -> [groups] instead of role -> group?

smunilla: It depends on our approach. It would be `'master_lb': 'master_lb'`. This is mostly here to handle the pluralization that the groups expected.

detiber: As long as the implementation handles roles/groups that are not defined in the installer, I think we'll be good.

For example, I think it would be perfectly valid to break out etcd hosts separately using the arbitrary config, where we would not support that through the guided installer.

smunilla:
@detiber All nosetests fixed and I think I addressed all your comments. Mind taking another look?

detiber:
@smunilla now pylint is unhappy :)

smunilla:
@detiber Stupid pylint making me write clean code! Fixed :)

detiber:
aos-ci-test

detiber:
@abutcher ptal when you get a chance.

@sdodson how hard would it be to generate a puddle with this change included for testing by the CFME team?

sdodson:
> @sdodson how hard would it be to generate a puddle with this change included for testing by the CFME team?

Not hard, only this change atop openshift-ansible-3.0.94-1 or current state of master after this merges?

detiber:
@sdodson just this change on top of 3.0.94-1 would be great.

sdodson:
> @sdodson just this change on top of 3.0.94-1 would be great.

cherry-picked those changes into https://github.com/openshift/openshift-ansible/releases/tag/openshift-ansible-3.0.95-1

ganhuang:
Hello, I suspect it can't be testable after trying with latest code.

* 1. Inventory file is generated not as what we expect. 

Create test_installer.cfy.yaml :
```
---
version: v3
variant: openshift-enterprise
variant_version: '3.2'
ansible_ssh_user: root 
 
deployment:
  ansible_config: /usr/share/atomic-openshift-utils/ansible.cfg
  ansible_log_path: /tmp/ansible.log
 
  openshift_set_hostname: true
  openshift_hosted_registry_storage_host: 10.x.x.x 
  openshift_hosted_registry_storage_kind: nfs
  openshift_hosted_registry_storage_nfs_options: '*(rw,root_squash)'
  openshift_hosted_registry_storage_nfs_directory: /exports
  openshift_hosted_registry_storage_volume_name: registry
  openshift_hosted_registry_storage_access_modes: ['ReadWriteMany']
  
  hosts:
  - connect_to: 10.x.x.x 
    hostname: 10.x.x.x 
    ip: 10.x.x.x 
    public_ip: 10.x.x.x 
    public_hostname: 10.x.x.x 
    
    node_labels:
      region: master
      zone: default   
    roles:
    - node 
    - master
    - storage 
  roles:    
    master:
      containerized: true
    node:
      containerized: true
    storage:
```
Check the inventory hosts after running "atomic-openshift-installer -c test_installer.cfy.yaml -u install":
```
[OSEv3:children]
nodes
nfs
masters

[OSEv3:vars]
None=/usr/lib/python2.7/site-packages/ooinstall/ansible_plugins
ansible_log_path=/tmp/ansible.log
None=v3
None=/.ansible
None=/.ansible/callback_facts.yaml
None=/usr/share/ansible/openshift-ansible/
ansible_config=/usr/share/atomic-openshift-utils/ansible.cfg
None=openshift-enterprise
None=/hosts
ansible_ssh_user=root
None={'roles': {'node': {'containerized': True}, 'master': {'containerized': True}, 'storage': {'containerized': True}}, 'ansible_log_path': '/tmp/ansible.log', 'openshift_hosted_registry_storage_volume_name': 'registry', 'ansible_config': '/usr/share/atomic-openshift-utils/ansible.cfg', 'openshift_set_hostname': True, 'openshift_hosted_registry_storage_nfs_options': '*(rw,root_squash)', 'hosts': [{'public_ip': '10.x.x.x', 'roles': ['node', 'master', 'storage'], 'node_labels': {'region': 'master', 'zone': 'default'}, 'connect_to': '10.x.x.x', 'ip': '10.x.x.x', 'public_hostname': '10.x.x.x', 'hostname': '10.x.x.x'}], 'openshift_hosted_registry_storage_kind': 'nfs', 'openshift_hosted_registry_storage_host': '10.x.x.x', 'openshift_hosted_registry_storage_access_modes': ['ReadWriteMany'], 'openshift_hosted_registry_storage_nfs_directory': '/exports'}
None=3.2
containerized=True
containerized=True

[masters]
10.x.x.x  openshift_ip=10.x.x.x openshift_public_ip=10.x.x.x openshift_hostname=10.x.x.x openshift_public_hostname=10.x.x.x

[nodes]
10.x.x.x  openshift_ip=10.x.x.x openshift_public_ip=10.x.x.x openshift_hostname=10.x.x.x openshift_public_hostname=10.x.x.x openshift_schedulable=True

[nfs]
10.x.x.x  openshift_ip=10.x.x.x openshift_public_ip=10.x.x.x openshift_hostname=10.x.x.x openshift_public_hostname=10.x.x.x
```

* 2. The installer config generated by current work flow (atomic-openshift-installer install) is not compatible with the arbitrary config. 

installer.cfg.yml generated by "atomic-openshift-installer install":
```
ansible_callback_facts_yaml: /home/ghuang/.config/openshift/.ansible/callback_facts.yaml
ansible_config: /usr/share/atomic-openshift-utils/ansible.cfg
ansible_inventory_path: /home/ghuang/.config/openshift/hosts
ansible_log_path: /tmp/ansible.log
ansible_ssh_user: root
deployment:
  hosts:
  - connect_to: 10.x.x.x
    containerized: true
    hostname: 10.x.x.x
    ip: 10.x.x.x
    public_hostname: 10.x.x.x
    public_ip: 10.x.x.x
    roles:
    - master
    - node
    - storage
  master_routingconfig_subdomain: ''
  proxy_exclude_hosts: ''
  proxy_http: ''
  proxy_https: ''
  roles: {}
variant: openshift-enterprise
variant_version: '3.2'
version: v2
```
Trigger the unattended mode with installer.cfg.yml above
```
# atomic-openshift-installer -u install
Any roles assigned to hosts must be defined.
```

I don't think this is the expected result  if the way of my installs is correct.

Btw, from the example config, two points confused me:

- Seems the parameter of "containerized" under node role is applied to all nodes, could every host variables (such as label, containerized) with node role be configurable individually?
- The deployment keys (execpt hosts and roles) and configuration values can be set  just like "OSEv3:vars" group in inventory file? If yes,  provider configs can be set in both the deployment keys and master roles?

detiber:
@ganhuang I'll leave @smunilla to respond to the majority of your comment, but I will comment on this small bit:

> Seems the parameter of "containerized" under node role is applied to all nodes, could every host variables (such as label, containerized) with node role be configurable individually?

Yes, setting it under the node role is equivalent to setting it on all nodes. It could also be set under deployment, which would cause all hosts defined to be containerized, or also under each host definition, which would cause a single host to be containerized:

```
deployment:
  hosts:
  - connect_to: 10.x.x.x
    containerized: true
    hostname: 10.x.x.x
    ip: 10.x.x.x
    public_hostname: 10.x.x.x
    public_ip: 10.x.x.x
    roles:
    - master
    - node
    - storage
```

> The deployment keys (execpt hosts and roles) and configuration values can be set just like "OSEv3:vars" group in inventory file? If yes, provider configs can be set in both the deployment keys and master roles?

Yes, setting a variable at the deployment level is equivalent to setting them in `[OSEv3:vars]`, setting it on the master role is equivalent to setting it in `[masters:vars]`, and setting it at the host level is equivalent to defining it on the same line as the host definition.

smunilla:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
@smunilla No longer WIP, right?

sdodson:
aos-ci-test

abutcher:
Needs rebase.

sdodson:
Oh noz!

nak3:
@smunilla @detiber and All 
Please please please update documentation if you changed configuration schema. Current doc's example[1] doesn't work at all nor I cannot find any PR in openshift-docs repository[2]

[1] https://docs.openshift.com/enterprise/3.2/install_config/install/quick_install.html#defining-an-installation-configuration-file

[2] https://github.com/openshift/openshift-docs/pulls

smunilla:
@nak3 I'm working on some migration code for the old schema and will get the docs updated asap.

nak3:
@smunilla Tahnk you so much! That really helps us.

PR #1781, opened 2016-04-20
Fix use of older image tag version during upgrade.

dgoodwin:
If doing an upgrade with a 3.2.0.x version that was older than the
latest, the upgrade would actually use the latest in the systemd unit
files and thus the actual containers that get used. (despite pulling
down the correct version first)

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1326642


detiber:
lgtm @brenton

PR #1783, opened 2016-04-21
Fix bug after portal_net move from master to common role.

dgoodwin:
Use of the variable was still pointing to the master role location
causing the default insecure registry subnet to be used rather than the
expected one.

Looks like an addition required for #1588 

CC @sdodson for verification this is right


sdodson:
LGTM, Thanks!

dgoodwin:
aos-ci-test

PR #1784, opened 2016-04-21
Only add new sccs

brenton:


brenton:
cc @detiber, just a minor update based on feedback from paul weil.

detiber:
:+1:

PR #1776, opened 2016-04-19
Image stream sync

sdodson:
- Switch to xpaas ose-v1.3.0-1
 - Switch to metrics to new upstream enterprise source
 - Sync all the things

PR #1786, opened 2016-04-21
openshift-metrics: adding duration and resolution options

### LINKED ISSUE #1767: Allow configuration of more options in openshift-metrics
The options ```METRIC_DURATION``` and ```METRIC_RESOLUTION``` should be configurable to allow better control of the hawkular-metrics.

#1582 is a reference to this issue.

enoodle:
Fixes #1767

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1644,6 +1644,8 @@ class OpenShiftFacts(object):
             defaults['hosted'] = dict(
                 metrics=dict(
                     deploy=False,
+                    duration=7,
+                    resolution=10,
                     storage=dict(
                         kind=None,
                         volume=dict(
--- a/roles/openshift_metrics/tasks/main.yaml
+++ b/roles/openshift_metrics/tasks/main.yaml
@@ -20,6 +20,7 @@
     {{ openshift.common.client_binary }}
     secrets new metrics-deployer
     nothing=/dev/null
+    --config={{hawkular_tmp_conf}}
     -n openshift-infra
   register: deployer_create_secret
   failed_when: "'already exists' not in deployer_create_secret.stderr and deployer_create_secret.rc !=0"
@@ -43,8 +44,9 @@
   shell: >
    {{ openshift.common.client_binary }} process -f \
    /usr/share/openshift/examples/infrastructure-templates/{{ hawkular_type }}/metrics-deployer.yaml -v \
-    HAWKULAR_METRICS_HOSTNAME=hawkular-metrics.{{ openshift.master.default_subdomain }},USE_PERSISTENT_STORAGE={{ hawkular_persistence }} | \
-    {{ openshift.common.client_binary }} create -n openshift-infra -f - 
+    HAWKULAR_METRICS_HOSTNAME=hawkular-metrics.{{ openshift.master.default_subdomain }} USE_PERSISTENT_STORAGE={{ hawkular_persistence }} \
+    METRIC_DURATION={{ openshift.hosted.metrics.duration }} METRIC_RESOLUTION={{ openshift.hosted.metrics.resolution }} |
+    {{ openshift.common.client_binary }} create -n openshift-infra --config={{hawkular_tmp_conf}} -f -
   register: oex_heapster_services
   failed_when: "'already exists' not in oex_heapster_services.stderr and oex_heapster_services.rc != 0"
   changed_when: false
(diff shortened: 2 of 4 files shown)


enoodle:
@simon3z
I am not sure what should be the default values of duration and resolution, I put in 7 as a place holder.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1646,6 +1646,8 @@ def get_defaults(self, roles, deployment_type):
                     deploy=False,
                     storage=dict(
                         kind=None,
+                        duration=7,
```
jdnieto: duration and resolution options should be at the same level that deploy or storage

roles/openshift_metrics/tasks/main.yaml
```diff
@@ -43,7 +43,8 @@
   shell: >
    {{ openshift.common.client_binary }} process -f \
    /usr/share/openshift/examples/infrastructure-templates/{{ hawkular_type }}/metrics-deployer.yaml -v \
-    HAWKULAR_METRICS_HOSTNAME=hawkular-metrics.{{ openshift.master.default_subdomain }},USE_PERSISTENT_STORAGE={{ hawkular_persistence }} | \
+    HAWKULAR_METRICS_HOSTNAME=hawkular-metrics.{{ openshift.master.default_subdomain }},USE_PERSISTENT_STORAGE={{ hawkular_persistence }} \
+    METRIC_DURATION=openshift_hosted_metrics_duration,METRIC_RESOLUTION=openshift_hosted_metrics_resolution | \
```
jdnieto: Should be METRIC_DURATION={{ openshift.hosted.metrics.duration }}
The tamplate doesn't support yet METRIC_RESOLUTION see [metrics-deployer.yaml](https://github.com/jdnieto/openshift-ansible/blob/ea80a6cb7fc4309e5c5c4210b2ce8504e8008c56/roles/openshift_examples/files/examples/v1.2/infrastructure-templates/enterprise/metrics-deployer.yaml)

roles/openshift_facts/library/openshift_facts.py
enoodle: Yes, Thanks!

roles/openshift_metrics/tasks/main.yaml
enoodle: Thanks! Ill see what I can do to add support for it.

simon3z:
> @simon3z
> I am not sure what should be the default values of duration and resolution, I put in 7 as a place holder.

@enoodle default for duration is 7 (days) and for resolution we have to see what @mwringe is going to use.
(I think it's 30, but double check).

enoodle:
This is still WIP as I need to test it correctly add add support for METRIC_RESOLUTION.

mwringe:
@simon3z The default metric resolution for Heapster upstream is 30 seconds, for the previous version we were using it was 10 seconds. The latest in origin-metrics is set to use 10 seconds (to make it more backwards compatible with what we had before). We may set this to 30 seconds though as having this value too low can cause performance issues.

simon3z:
> @simon3z The default metric resolution for Heapster upstream is 30 seconds, for the previous version we were using it was 10 seconds. The latest in origin-metrics is set to use 10 seconds (to make it more backwards compatible with what we had before). We may set this to 30 seconds though as having this value too low can cause performance issues.

@mwringe are you sure about 10 seconds? I think it was 5. We had 4 samples every 20 seconds in CFME, which means that the resolution was 5 seconds in the past.

mwringe:
@simon3z Ah, maybe it was 5 seconds in the past. For CFME would you need this to still be specified as 5 seconds? I can easily change the default in our template to go back to 5 seconds.

simon3z:
@mwringe we have 20 seconds samples so anything higher that that is problematic (e.g. 30 seconds).

detiber:
@sosiouxme @mwringe Which versions of the metrics deployer support METRICS_RESOLUTION and METRICS_DURATION?

mwringe:
@detiber : 
METRICS_DURATION is supported by all versions of origin metrics and OSE Metrics

METRICS_RESOLUTION is much newer (as in less than a week old) and is only supported in the more recent origin-metrics v1.2.0-rc1

roles/openshift_metrics/tasks/main.yaml
enoodle: Apparently it was already done #1776

enoodle:
Changed the default value of METRICS_RESOLUTION to 10. IIUC it is passed as an environment variable so if an earlier version of openshift-ansible is used it wont raise a problem.

detiber:
ok to test

detiber:
aos-ci-test

detiber:
aos-ci-test

PR #1788, opened 2016-04-21
Fix route name for jenkins example templates.

ikatson:
Before this change the route name was hard-coded, and if you try to create multiple instances of jenkins, the route creation would fail.


sdodson:
@ikatson We really just pull things from upstream repos for inclusion in the installer, has this been fixed upstream in https://github.com/jboss-openshift/application-templates ?

sdodson:
Sorry, wrong upstream... these come from https://github.com/openshift/origin/tree/master/examples/jenkins

ikatson:
@sdodson sorry, I had no idea, will do there instead. Does not seem to be fixed in master there.

sdodson:
@ikatson thanks, feel free to mention me in that PR, I'm happy to merge this once it's accepted upstream. I just want to make sure that it doesn't regress the next time we sync.

PR #1789, opened 2016-04-21
Remove /dev mount for node container

### LINKED ISSUE #1779: PTY allocation request failed on channel 0
openshift-ansible fails when running on atomic host (containerized deployment). During openshift-ansible run selinux setting on master&openshift nodes is changed which causes that any ssh connection to master/openshift nodes fails with error:

$ ssh centos@172.24.4.204
PTY allocation request failed on channel 0

In permissive mode this AVC is logged on the master node:
Apr 20 15:27:06 atomic2-origin-master-0.example.com kernel: type=1400 audit(1461166026.440:42): avc:  denied  { open } for  pid=3978 comm="sshd" path="/dev/pts/ptmx" dev="devpts" ino=2 scontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:devpts_t:s0 tclass=chr_file

used env:
- centos atomic host
- openshift-ansible master branch (last commit afff74593f6de2492e3ed1cbe63f68a2dfb732f3)

sdodson:
This change breaks some storage plugins but leaving ti as is breaks non root logins. A fix for
docker is coming in a future updated to docker-1.9.1 at which time we'll revert
this once we have a method to ensure we're running a fixed version.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1329326
Fixes #1779


sdodson:
To be merged once we have final decision that docker will not fix this for 7.2.4.
CC: @eparis @brenton

sdodson:
No longer necessary. This is fixed in docker-1.10 and a yet to be released version of docker-1.9.1 in RHEL7.

PR #1791, opened 2016-04-22
Reconcile roles in additive-only mode on upgrade

liggitt:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1328822 by changing the automatic role reconcile up upgrade to leave existing permissions alone. Cluster admins can still run `oadm policy reconcile-cluster-roles --additive-only=false` to tighten permissions to the current defaults post-upgrade, if they want.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_0_minor/upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_0_minor/upgrade.yml
@@ -42,10 +42,10 @@
 - name: Update cluster policy
   hosts: oo_first_master
   tasks:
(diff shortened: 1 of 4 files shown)


liggitt:
@sdodson PTAL

liggitt:
verified `--additive-only` option existed since commit https://github.com/openshift/ose/commit/e837f1584d37f6308c4d204f5a2f2218cde50223#diff-fd19f0bea8ffa99f2153e5aaa1278b32

sdodson:
aos-ci-test

detiber:
aos-ci-test

Issue #1790, opened 2016-04-22
docker_version is never found causing it to be skipped and never installed.


jmashburn:
Deploying on a Mac using Vagrant/VirtualBox and deployment_type=origin.   I've found that while setting the openshift_docker_facts a docker_version is never set do to the conditionals comparing the l_common_version in roles/openshift_docker_facts/tasks/main.yml [Set docker version to be installed].

```
TASK: [openshift_docker_facts | debug var=l_common_version] *******************
ok: [master1] => {
    "var": {
        "l_common_version": "l_common_version"
    }
}

TASK: [openshift_docker_facts | set_fact ] ************************************
ok: [master1]

TASK: [openshift_docker_facts | debug var=l_common_version] *******************
ok: [master1] => {
    "var": {
        "l_common_version": "1.1.6"
    }
}

TASK: [openshift_docker_facts | Set docker version to be installed] ***********
skipping: [master1]

TASK: [openshift_docker_facts | debug var=docker_version] *********************
ok: [master1] => {
    "var": {
        "docker_version": ""
    }
}

TASK: [openshift_docker_facts | Set docker version to be installed] ***********
skipping: [master1]

TASK: [openshift_docker_facts | debug var=docker_version] *********************
ok: [master1] => {
    "var": {
        "docker_version": ""
    }
}

TASK: [docker | Get current installed version if docker_version is specified] ***
skipping: [master1]

TASK: [docker | stat path=/etc/sysconfig/docker-storage] **********************
ok: [master1]

TASK: [docker | Remove deferred deletion for downgrades from 1.9] *************
skipping: [master1]

TASK: [docker | Downgrade docker if necessary] ********************************
skipping: [master1]

TASK: [docker | Install docker] ***********************************************
skipping: [master1]

TASK: [docker | Check that docker hasn't entered failed state] ****************
ok: [master1]

TASK: [docker | Reset docker service state] ***********************************
skipping: [master1]

TASK: [docker | Start the docker service] *************************************
failed: [master1] => {"failed": true}
msg: Error when trying to enable docker: rc=1 Failed to execute operation: Access denied

FATAL: all hosts have already failed -- aborting
```

jmashburn:
Ran again. Successful.   l_common_version = 1.2.x which assigned docker_version = 1.9.1

PR #1792, opened 2016-04-22
Fix for docker not present

detiber:
Fixes https://github.com/openshift/openshift-ansible/issues/1790

@sdodson ptal

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -22,7 +22,7 @@
 
 - name: Install docker
   action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version is defined and docker_version != '' else '' }} state=present"
-  when: not openshift.common.is_atomic | bool and not docker_version_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare(docker_version, 'lt')
+  when: not openshift.common.is_atomic | bool and docker_downgrade_result | skipped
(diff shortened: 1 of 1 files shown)


sdodson:
LGTM but i'm surprised it was as complicated as it was without good reason.

sdodson:
aos-ci-test

roles/docker/tasks/main.yml
```diff
@@ -22,7 +22,7 @@
 
 - name: Install docker
   action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version is defined and docker_version != '' else '' }} state=present"
-  when: not openshift.common.is_atomic | bool and not docker_version_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare(docker_version, 'lt')
+  when: not openshift.common.is_atomic | bool and docker_downgrade_result | skipped
```
detiber: As far as I can tell this covers the previous case where we only want to do the install if the version available is less than the version we want and it fixes the issue where docker wasn't being installed when docker_version is not present (since docker_downgrade_result will only be skipped if docker_version_result is skipped or the installed version is not greater than the current version). Now it will not skip this task when the version installed is equal to the version wanted, but that is a noop.

detiber:
@brenton merging since this is impacting installations where docker is not already present on hosts and docker_version is not being set.

Issue #686, opened 2015-10-12
Ansible 2.0 support


detiber:
If possible we should ensure that we can run the playbooks using ansible 2.x as well as 1.9.z

one issue we have is that variables with leading underscores are not valid with ansible 2.0

jtslear:
Or organization has started to use version 2 for newer feature/modules ansible is providing.  So this had started to hold us up.  Unfortunately I'm not well into the world of python.  I took a stab at this and ended up here:

```
TASK [Generate etcd instance names(s)] *****************************************
/Users/jskarbek/projects/openshift-ansible/lookup_plugins/sequence.py:18: RuntimeWarning: Parent module 'ansible.plugins.lookup' not found while handling absolute import
  from ansible.errors import AnsibleError
/Users/jskarbek/projects/openshift-ansible/lookup_plugins/sequence.py:19: RuntimeWarning: Parent module 'ansible.plugins.lookup' not found while handling absolute import
  import ansible.utils as utils
/Users/jskarbek/projects/openshift-ansible/lookup_plugins/sequence.py:20: RuntimeWarning: Parent module 'ansible.plugins.lookup' not found while handling absolute import
  from re import compile as re_compile, IGNORECASE
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: __init__() takes exactly 2 arguments (1 given)
fatal: [localhost]: FAILED! => {"failed": true, "stdout": ""}
```

See my branch above for the  changes made thus far.  Aside from the error above, doesn't look too far away.

detiber:
@jtslear thanks for digging into this.

The error you are seeing is because the sequence plugin in our repo (which we added as a workaround to a specific bug in one of the 1.9.x releases) is not compatible with 2.0.

Since the latest 1.9.4 release no longer needs this fix, I think we should just set the minimum supported version to 1.9.4 and then we can remove the sequence lookup plugin from our repo.

benbarclay:
Ansible 2.0 support would be awesome. My team has to use SAML auth (which was never released into any Ansible 1.x branches) to deploy against a number of AWS accounts. 

As it is, we are running Ansible from a git checkout that exists somewhere 1.9.x and 2.0, that has the SAML auth additions, but not the plugin breaking changes.

detiber:
@benbarclay you may want to give https://github.com/openshift/openshift-ansible/pull/991 a try to see if it works for you.

I suspect there will be other issues along the way, but any testing we can get the better.

[...]

PR #1799, opened 2016-04-25
port filter_plugins to ansible2

ibotty:


abutcher:
aos-ci-test

abutcher:
ok to test

detiber:
@ibotty pylint is a bit unhappy:
```
************* Module openshift_master
I: 41, 0: Locally disabling too-many-instance-attributes (R0902) (locally-disabled)
W: 14, 0: No exception type(s) specified (bare-except)
E: 15, 4: No name 'plugins' in module 'ansible' (no-name-in-module)
F: 15, 4: Unable to import 'ansible.plugins.filter.core' (import-error)
```
Can you add pylint disables for those lines?

ibotty:
I did not expect that you pay attention to pylint. there are so many warnings in all the files. I think I fixed the pylint warnings I introduced.

detiber:
aos-ci-test

filter_plugins/oo_filters.py
```diff
@@ -128,14 +128,14 @@ def oo_select_keys(data, keys):
                 returns [1, 3]
         """
 
-        if not isinstance(data, dict):
-            raise errors.AnsibleFilterError("|failed expects to filter on a dict")
+        if not isinstance(data, dict) and not isinstance(data, object):
+            raise errors.AnsibleFilterError("|failed expects to filter on a dict or object")
```
abutcher: @ibotty could you explain this change?

detiber: After a bit of digging, we believe we better understand the issue here...  It looks like in Ansible 2.0+ passing `hostvars` without referencing a specific host passes a HostVars object into the filter rather than a dict. HostVars inherits from Mapping, so instead of comparing against object here, you should be able to just update the test as below:
```
if not isinstance(data, collections.Mapping):
```

@abutcher is doing some more testing around this as well.

abutcher: Ah, we're passing `hostvars` into `oo_select_keys` which is now an object (`collections.Mapping`). Since the `hostvars` object and dictionaries provide mapping interfaces, we can probably just check for `isinstance(data, collections.Mapping)` and that should work for python 2.6+.

ibotty: You are right, that hostvars is now an object. I updated the patch to check whether we are dealing with a Mapping.

detiber:
aos-ci-test

abutcher:
Linking to #686

PR #1804, opened 2016-04-25
Replace deprecated sudo with become.

abutcher:
@detiber PTAL


abutcher:
aos-ci-test

detiber:
@abutcher I'm a little torn on using true/false in the playbooks here. Ansible docs generally use yes/no. 

That said, I think using true/false in the inventory makes more sense to more users.

abutcher:
aos-ci-test

abutcher:
@detiber Updated for `yes`/`no`

PR #1805, opened 2016-04-25
For s3_registry backing, only generate secret on one master

bjpritch:
Running secret creation on multiple masters simultaneously causes those actions after the first secret is created to fail since it already exists.  Adding "run_once" forces ansible to only create the registry secret on the first master and move along.

### DIFF
--- a/playbooks/adhoc/s3_registry/s3_registry.yml
+++ b/playbooks/adhoc/s3_registry/s3_registry.yml
@@ -5,8 +5,10 @@
 #
 # The AWS access/secret keys should be the keys of a separate user (not your main user), containing only the necessary S3 access role.
 # The 'clusterid' is the short name of your cluster.
+#
+# This playbook should be limited to one openshift master
 
-- hosts: tag_clusterid_{{ clusterid }}:&tag_host-type_openshift-master
+- hosts: tag_clusterid_{{ clusterid }}:&tag_host-type_openshift-master[0]
   remote_user: root
   gather_facts: False
 
@@ -57,7 +59,6 @@
 
   - name: Add secrets to registry service account
     command: oc secrets add serviceaccount/registry secrets/dockerregistry
-    run_once: true
     when: "'dockerregistry' not in serviceaccount.stdout"
 
   - name: Determine if deployment config contains secrets
--- a/playbooks/adhoc/s3_registry/s3_registry.yml
+++ b/playbooks/adhoc/s3_registry/s3_registry.yml
@@ -8,7 +8,16 @@
 #
 # This playbook should be limited to one openshift master
 
-- hosts: tag_clusterid_{{ clusterid }}:&tag_host-type_openshift-master[0]
+- hosts: tag_clusterid_{{ clusterid }}:&tag_host-type_openshift-master
+  remote_user: root
+  tasks:
+  - add_host:
+      name: hostvars[item].name
+      group: oo_first_master
+    with_items: "{{ play_hosts[0] }}"
+
+
+- hosts: oo_first_master
   remote_user: root
   gather_facts: False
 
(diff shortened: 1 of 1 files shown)


playbooks/adhoc/s3_registry/s3_registry.yml
```diff
@@ -57,6 +57,7 @@
 
   - name: Add secrets to registry service account
     command: oc secrets add serviceaccount/registry secrets/dockerregistry
+    run_once: true
```
detiber: If I'm reading this correctly, all of this really just needs to be done once and not just this single step.

It might be better to have a separate play at the beginning of this file to create a new group that contains just a single master host.  Something like this should work:
```
hosts: localhost
become: no
sudo: no
tasks:
- add_host:
    name: hostvars[item].name
    group: oo_masters
  with_items: groups['tag_clusterid_' ~ cluster_id] | default([]) | intersect(groups['tag_host-type_master'] | default([]))
```

And then for this play, you could just specify:
```
hosts: oo_masters[0]
```

bjpritch:
Correct, this playbook looks to only be needed for one master.  I've updated the host group pattern directly to limit this run to the first master.

playbooks/adhoc/s3_registry/s3_registry.yml
```diff
@@ -5,8 +5,10 @@
 #
 # The AWS access/secret keys should be the keys of a separate user (not your main user), containing only the necessary S3 access role.
 # The 'clusterid' is the short name of your cluster.
+#
+# This playbook should be limited to one openshift master
 
-- hosts: tag_clusterid_{{ clusterid }}:&tag_host-type_openshift-master
+- hosts: tag_clusterid_{{ clusterid }}:&tag_host-type_openshift-master[0]
```
detiber: I admit I haven't played around too much with the more advanced ansible pattern matching here, but it looks to me like this might take the first host out of all with tag_host-type_openshift-master and then 'and' it with tag_cluster_id{{ clusterid }}, which would present issues with trying to run this with multiple clusters defined.

bjpritch:
You are correct that specifying host index in openshift-master group first would not use the correct logic when working with multiple clusters.  I've included a patch similar to your first suggestion that calculates the host outside of the s3 registry play.

playbooks/adhoc/s3_registry/s3_registry.yml
```diff
@@ -5,9 +5,20 @@
...
+#
+# This playbook should be limited to one openshift master
 
 - hosts: tag_clusterid_{{ clusterid }}:&tag_host-type_openshift-master
   remote_user: root
+  tasks:
+  - add_host:
+      name: hostvars[item].name
+      group: oo_first_master
+    with_items: "{{ play_hosts[0] }}"
```
detiber: The way you are currently doing this, you are running the add_host task against all hosts specified. In order to limit this to a single execution you could do the following:
```
- hosts: localhost
  gather_facts: no
  become: no
  tasks:
  - add_host:
      name: "{{ (groups['tag_clusterid_' ~ clusterid] | union(groups['tag_host-type_openshift-master']))[0] }}"
      group: oo_first_master
```
or:
```
- hosts: tag_clusterid_{{ clusterid }}:&tag_host-type_openshift-master
  remote_user: root
  become: no
  tasks:
  - add_host:
      name: "{{ inventory_hostname }}"
      group: oo_first_master
    run_once: yes
```

As an aside, the way you are specifying the name attribute to add_host, it will actually set the name to 'hostvars[item].name', you will need to use an explicit jinja2 template there to evaluate the value.

The with_items on the other hand should use an implicit jinja2 template, but you need to pass it a list, so you could do this instead:
```
with_items: [play_hosts[0]] 
```

bjpritch:
@detiber 
Thanks for the feedback.  I've reworked according to your suggestions.

detiber:
@etsauer I know I've discussed this playbook with you guys, but I'm not sure if you are making use of it yet. If so, any issues with the change?

tbielawa:
This issue has been inactive for quite some time. Please update and reopen this issue if this is still a priority you would like to see action on.

PR #1806, opened 2016-04-25
In s3_registry playbook, modified the host group for tagged master instances

bjpritch:
According to http://docs.ansible.com/ansible/intro_dynamic_inventory.html#example-aws-ec2-external-inventory-script under "Tags", host group names would have dashes converted to underscores

### DIFF
--- a/playbooks/adhoc/s3_registry/s3_registry.yml
+++ b/playbooks/adhoc/s3_registry/s3_registry.yml
@@ -6,7 +6,7 @@
 # The AWS access/secret keys should be the keys of a separate user (not your main user), containing only the necessary S3 access role.
 # The 'clusterid' is the short name of your cluster.
 
-- hosts: tag_clusterid_{{ clusterid }}:&tag_host-type_openshift-master
+- hosts: tag_clusterid_{{ clusterid }}:&tag_host_type_openshift_master
   remote_user: root
   gather_facts: False
 


playbooks/adhoc/s3_registry/s3_registry.yml
```diff
@@ -6,7 +6,7 @@
 # The AWS access/secret keys should be the keys of a separate user (not your main user), containing only the necessary S3 access role.
 # The 'clusterid' is the short name of your cluster.
 
-- hosts: tag_clusterid_{{ clusterid }}:&tag_host-type_openshift-master
+- hosts: tag_clusterid_{{ clusterid }}:&tag_host_type_openshift_master
```
detiber: This one is a bit tricky, we have some users that have run this playbook that have the 'replace_dash_in_groups' variable in ec2.ini set to False, which means that this change would cause this script to no longer work for those users.

It might be better to add a variable for setting this instead.

That, or we would need to update the inventory/aws/hosts/ec2.ini file in this repo and all of the references with '-''s in them under playbooks/aws/

bjpritch:
I did not see this option when reviewing our dynamic inventory (it's listed in Ansible's ec2.py but not in their example ec2.ini).  This pull request can be canceled as it's simpler to update our inventory to match the expected host group.

PR #1812, opened 2016-04-26
Add masterConfig.volumeConfig.dynamicProvisioningEnabled

### LINKED ISSUE #1759: Allow opt-out of Dynamic Provisioning
Dynamic provisioning is on by default in Kube and Origin.  A feature flag was recently added to disable the default provisioning functionality.

The new feature flag is "DynamicProvisioningEnabled" and a missing config value defaults to True.

https://github.com/openshift/origin/pull/8426

sdodson:
Fixes #1759


sdodson:
@markturansky please take a look, thanks

markturansky:
I don't know ansible at all, but the flag is correct and from the filenames, I think the defaults look correct.

OSO and Dedicated have their own set of ansible scripts where this would be disabled?

sdodson:
> OSO and Dedicated have their own set of ansible scripts where this would be disabled?

It's my understanding that they pull our ansible code. They'd want to set openshift_master_dynamic_provisioning_enabled=False in their inventories. They should know how to do that. As long as the config file templates are correct i'm happy.

sdodson:
aos-ci-test

detiber:
@dak1n1 this will require a new variable be set to disable the dynamic provisioning for Ops:
```
openshift_master_dynamic_provisioning_enabled=False
```

markturansky:
@detiber got an issue in the private repo to track that:  https://github.com/openshift/online/issues/109

dak1n1:
:+1:

PR #1816, opened 2016-04-27
how to install ansible in RHEL 7

ReSearchITEng:
how to install ansible in RHEL 7

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,12 @@ This repo contains Ansible code for OpenShift and Atomic Enterprise.
   ```
     dnf install -y ansible-1.9.4 rubygem-thor rubygem-parseconfig util-linux pyOpenSSL libffi-devel python-cryptography
   ```
+  - RHEL 7:
+  ```
+  yum install http://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+  rpm --import http://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-7
+  yum install ansible1.9-1.9.4-2.el7 rubygem-thor rubygem-parseconfig util-linux pyOpenSSL libffi-devel python-cryptography
+  ```
    - OSX:
   ```
     # Install ansible 1.9.4 and python 2


README.md
```diff
@@ -8,6 +8,12 @@ This repo contains Ansible code for OpenShift and Atomic Enterprise.
   ```
     dnf install -y ansible-1.9.4 rubygem-thor rubygem-parseconfig util-linux pyOpenSSL libffi-devel python-cryptography
   ```
+  - RHEL 7:
+  ```
+  yum install http://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+  rpm --import http://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-7
+  yum install ansible1.9-1.9.4-2.el7 rubygem-thor rubygem-parseconfig util-linux pyOpenSSL libffi-devel python-cryptography
```
detiber: I think the only thing that should be needed here is to install just the ansible package. I can't think of a reason why the python dependencies for ansible wouldn't pull in properly or why we would need the non-python deps at all here.

ReSearchITEng: Running the `yum install ansible-1.9.4` does not end with success on RHEL 7.
(and, openshift-ansible is not compatible with ansible 2.0 ).
The only option I could find to install 1.9.4 on RHEL is adding the epel repo, so I can install: ansible1.9-1.9.4-2.el7.

detiber: @ReSearchITEng ah, I should have said that I onlyl see the need to install EPEL and the Ansible package.  I was referring to the other packages listed on line 15 in my previous comment.

ReSearchITEng: Ok, so it won't be a complete guide for RHEL (like it is for Fedora), but only a note, something like: "For RHEL, install epel and in the Fedora list replace ansible-1.9.4 with ansible1.9-1.9.4-2.el7 ". It might be little confusing for the reader. Any phrasing you choose, is good, as long as ppl will have at least some hints how to proceed.

detiber:
@sdodson ptal

README.md
ingvagabund: I would not mention specific ansible version since we have already `ansible-2.3.0.0-3.el7` in epel7.

mtnbikenc:

PR #1818, opened 2016-04-27
Fix inventory properties with raw booleans, again...

dgoodwin:
The fix earlier in this affected method was only being applied when
merging old and new facts. In a first run on a clean system with no
pre-existing openshift facts cache, the yaml safe load was not applied
resulting in the same broken master config. (which would fix itself if
you just reran the config playbook)

Apply the same check on new facts not previously applied on the system.


dgoodwin:
Found it @dak1n1 , it was just on the first clean run, a re-run would correct the problem due to the previous attempt at a fix, which caused me to think it was fully fixed.

sdodson:
aos-ci-test

detiber:
Test failure looks like it's related to node labeling, which causes occasional issues...  I'm thinking we probably need to address that one sooner rather than later. Not a blocker for this PR though.

dak1n1:
:+1:

detiber:
aos-ci-test

PR #1828, opened 2016-04-27
Fix openshift_generate_no_proxy_hosts boolean

sdodson:
- Fix example hosts files
- Fix openshift_generate_no_proxy_hosts boolean handling
  - When this was fixed it became obvious that we should add the current host to the no_proxy list regardless of whether or not they've disabled the generation of the full host list because without this a master cannot talk to etcd on the same host and I don't believe admins should be expected to account for that.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1330934
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1330920

I'd say this is a P2, both bugs have been marked upcoming release. The first is more substantial than the second.


sdodson:
aos-ci-test

sdodson:
@detiber this should be good to go

detiber:
:+1:

PR #1830, opened 2016-04-27
a-o-i: Populate groups for openshift_facts

smunilla:
openshift_facts is currently failing because it doesn't properly set up groups
after the proxy changes we made. This fixes that.

### DIFF
--- a/playbooks/byo/openshift_facts.yml
+++ b/playbooks/byo/openshift_facts.yml
@@ -1,4 +1,14 @@
 ---
+- name: Cluster hosts
+  hosts: localhost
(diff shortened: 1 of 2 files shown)


smunilla:
@detiber What do you think of this approach?

abutcher:
@detiber RE: recent discussions, this goes with the additional play route as opposed to `group_vars` at the `/playbooks/byo` level to avoid adding another `vars_files` to an include.

detiber:
lgtm

detiber:
aos-ci-test

PR #1831, opened 2016-04-27
Update repos per inventory before upgrading

sdodson:
Useful for me when i have to include pre-release repos during upgrades. PErhaps we should just make openshift_repos a dependency of openshift_facts instead?


detiber:
@sdodson hmm, I don't see an issue with moving the dependency to openshift_facts.

sdodson:
Hmm, openshift_repos depends on openshift_facts so that'd create a circular dependency. How do you like it as is?

PR #1834, opened 2016-04-28
update tenand_id typo in example file

jianlinliu:
@brenton could you help have a review.


brenton:
LGTM, @detiber 

This is obviously safe to merge now but if we want to wait until after we're certain the release blockers are merged that would be fine with me.

PR #1835, opened 2016-04-28
BuildDefaults are a kube admission controller not an openshift admission

sdodson:
controller

Also, improve commenting

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1389,9 +1389,9 @@ def set_proxy_facts(facts):
         facts['common'] = common
 
     if 'builddefaults' in facts:
-        facts['master']['admission_plugin_config'] = dict()
         builddefaults = facts['builddefaults']
         common = facts['common']
+        # Copy values from common to builddefaults
         if 'http_proxy' not in builddefaults and 'http_proxy' in common:
             builddefaults['http_proxy'] = common['http_proxy']
         if 'https_proxy' not in builddefaults and 'https_proxy' in common:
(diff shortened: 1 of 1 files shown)


sdodson:
@dgoodwin mind looking this over?

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1402,11 +1402,14 @@ def set_proxy_facts(facts):
...
-            builddefaults['admission_plugin_config'] = dict()
+        # If we're actually defining a proxy config then create kube_admission_plugin_config
+        # if it doesn't exist, then merge builddefaults[config] structure
+        # into kube_admission_plugin_config
+        if 'kube_admission_plugin_config' not in facts['master']:
+            facts['master']['kube_admission_plugin_config'] = dict()    
         if 'config' in builddefaults and ('http_proxy' in builddefaults or \
                 'https_proxy' in builddefaults):
-            facts['master']['admission_plugin_config'].update(builddefaults['config'])
+            facts['master']['kube_admission_plugin_config'].update(builddefaults['config'])
```
dgoodwin: Should we pull 'config' out of builddefaults here to avoid duplicating it in the facts?

dgoodwin:
:+1: from me otherwise!

roles/openshift_facts/library/openshift_facts.py
sdodson: Oh, that's the sort of messy thing. So builddefaults.config yaml structure is defined here 
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_builddefaults/vars/main.yml
and it depends on other builddefaults values being set, so we run the builddefaults openshift_facts role twice here
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_builddefaults/tasks/main.yml

I guess I could move the yaml structure into openshift_facts.py but I really dislike having that embedded this deeply. What do you think? Or maybe I'm misunderstanding what you're suggesting.

dgoodwin: I might be a little bit confused, but I was referring to a "del builddefaults['config']" so we don't double serialize that structure to the fact cache on the system. (as it would also appear in master.kube_admission_plugin_config)

sdodson: I understand now. It looks like master.kube_admission_plugin_config isn't actually being persisted FWIW.

dgoodwin: Ahh I follow. I often think I've figured this module out and soon realize I'm still getting there. :)

sdodson:
aos-ci-test

sdodson:
aos-ci-test

PR #1841, opened 2016-04-29
Add openshift_docker_hosted_registry_insecure option

### LINKED ISSUE #1832: secure registry considerations
There seems to be a default no possible way to not have the --insecure-registry line added. It appears to be hardcoded.

Also, what are thoughts on adding the ca.crt into the the root bundle: eg.
cp /etc/origin/node/ca.crt /etc/pki/ca-trust/source/anchors; update-ca-trust 

This could avoid having to worry about having to make changes to all nodes should the docker-registry be re-deployed etc. (ie. not worrying about /etc/docker/certs.d/docker-registry-ip:5000 )

The above is all assuming that people are using the default ca.crt for securing their registry..

andrewklau:
lets me override the hardcoded the insecure registry 

Resolves #1832

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -118,7 +118,7 @@
     # TODO: configure these based on
     # hostvars[groups.oo_first_master.0].openshift.hosted.registry instead of
     # hardcoding
-    openshift_docker_hosted_registry_insecure: True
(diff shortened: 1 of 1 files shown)


sdodson:
aos-ci-test

detiber:
ok to test

detiber:
@abutcher ptal

abutcher:
@detiber :shipit: 

@mwoodson FYI

PR #1842, opened 2016-04-29
Fix inventory syntax

talset:
Issue : 

```
ERROR: /etc/ansible/hosts:...: variables assigned to group must be in key=value form
```

Changed to match to the key=value syntax


abutcher:
@detiber

PR #1845, opened 2016-04-29
Make NetworkManager failure friendlier

sdodson:


sdodson:
@detiber just makes the error message friendlier

cben:
Q: why check if NM is installed instead of just installing it?
OK, I suppose when you install on existing machines which chose not to have NM you don't want to mess them.
Is there a place I can put `installed` and `enabled`, `started` that would only run on new-machine installs, such as `bin/cluster create libvirt`?

detiber:
@cben I don't see why we couldn't add that to the provisioning playbooks for the different providers, but I don't think we necessarily want to bring that in to the official playbooks/roles, since it could break someone's existing setup.

That said, those provisioning playbooks are community supported, so I can't guarantee a quick turnaround for implementing such a feature.  libvirt, openstack and aws should be a bit easier than gce, though, since cloud-init is available there.

Issue #1632, opened 2016-03-17
msg: Failed to stop atomic-openshift-master.service: Unit atomic-openshift-master.service not loaded.


jfcoz:
I'm trying to install origin with three masters

In playbooks/adhoc/metrics_setup/playbooks/install.yml I get error : 

```
PLAY [Install metrics] ******************************************************** 

GATHERING FACTS *************************************************************** 
ok: [master0.paas.dev.xxx]
ok: [master1.paas.dev.xxx]
ok: [master2.paas.dev.xxx]

TASK: [Add metrics url to master config] ************************************** 
changed: [master0.paas.dev.xxx]
changed: [master1.paas.dev.xxx]
changed: [master2.paas.dev.xxx]

TASK: [Restart master service] ************************************************ 
failed: [master0.paas.dev.xxx] => {"failed": true}
msg: Failed to stop atomic-openshift-master.service: Unit atomic-openshift-master.service not loaded.
Failed to start atomic-openshift-master.service: Unit atomic-openshift-master.service failed to load: No such file or directory.

failed: [master2.paas.dev.xxx] => {"failed": true}
msg: Failed to stop atomic-openshift-master.service: Unit atomic-openshift-master.service not loaded.
Failed to start atomic-openshift-master.service: Unit atomic-openshift-master.service failed to load: No such file or directory.

failed: [master1.paas.dev.xxx] => {"failed": true}
msg: Failed to stop atomic-openshift-master.service: Unit atomic-openshift-master.service not loaded.
Failed to start atomic-openshift-master.service: Unit atomic-openshift-master.service failed to load: No such file or directory.


FATAL: all hosts have already failed -- aborting

```

I use : 
```
product_type=openshift
deployment_type=origin
openshift_master_cluster_method=native
```

detiber:
@jfcoz the 'adhoc' playbooks are considered community supported and do not necessarily respect the variable names specified in the inventory example files. In this case the service name is hardcoded to atomic-openshift-master: https://github.com/openshift/openshift-ansible/blob/master/playbooks/adhoc/metrics_setup/playbooks/install.yml#L12

For origin it should be attempting to restart the origin-master service. Also for ha installations it should be attempting to restart the atomic-openshift-master-api service (or origin-master-api service for origin).

As an aside, the following PR has been submitted for deploying the metrics using an ansible role: https://github.com/openshift/openshift-ansible/pull/1582

jfcoz:
Thanks

PR #1849, opened 2016-05-02
using variable for deployment_type

jfcoz:
https://github.com/openshift/openshift-ansible/issues/1632


playbooks/adhoc/metrics_setup/playbooks/install.yml
```diff
@@ -9,8 +9,11 @@
     lineinfile: "state=present dest=/etc/origin/master/master-config.yaml regexp='^\ \ metricsPublicURL' insertbefore='^\ \ publicURL' line='\ \ metricsPublicURL: {{ metrics_public_url }}'"
 
   - name: "Restart master service"
-    service: name=atomic-openshift-master state=restarted
+    service: name={{deployment_type}}-master-api state=restarted
```
sdodson: {{ openshift.common.service_type }}  is what we use for this

detiber:
@sdodson I suspect the need for this will be negated with your upcoming changes. Please close if that is the case.

detiber:
@sdodson can you confirm this is no longer needed?

sdodson:
Yes, no longer necessary.

PR #1850, opened 2016-05-02
Fix tables in roles/openshift_metrics/README.md

cben:
Empty line needed before table or it won't render.  UPDATE: Also made example literal.
After: https://github.com/cben/openshift-ansible/blob/patch-1/roles/openshift_metrics/README.md


sdodson:
LGTM @detiber

PR #1853, opened 2016-05-02
Replace hostvars with vars for openshift env facts when ansible >= v2.

abutcher:
Working towards #686

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -7,10 +7,12 @@ Custom filters for use in openshift-ansible
 
 from ansible import errors
 from collections import Mapping
+from distutils.version import LooseVersion
 from operator import itemgetter
 import OpenSSL.crypto
 import os
 import pdb
+import pkg_resources
 import re
 import json
 import yaml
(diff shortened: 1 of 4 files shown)


abutcher:
aos-ci-test

roles/openshift_builddefaults/tasks/main.yml
```diff
@@ -5,7 +5,9 @@
     # TODO: add ability to define builddefaults env vars sort of like this
     # may need to move the config generation to a filter however.
     # openshift_env: "{{ hostvars[inventory_hostname]
-    #                    | oo_merge_dicts(hostvars)
+    #                    | oo_merge_dicts(vars if (ansible_version.full
+    #                                              | version_compare('2.0', 'ge')
+    #                                              | bool) else hostvars)
```
detiber: Would it be better to push the logic into a filter (say oo_merge_hostvars and pass it both hostvars and vars)?

It seems like it would be less error prone than doing the version compare multiple times in roles.

abutcher:
aos-ci-test

abutcher:
aos-ci-test

filter_plugins/oo_filters.py
```diff
@@ -71,6 +72,44 @@ def oo_merge_dicts(first_dict, second_dict):
         return merged
 
     @staticmethod
+    def oo_merge_hostvars(hostvars, variables, inventory_hostname, ansible_version):
```
detiber: Can ansible_version be accessed through the ansible api rather than being passed in?

abutcher:
aos-ci-test

filter_plugins/oo_filters.py
abutcher: Grabbing via `pkg_resources`

abutcher:
@detiber PTAL

detiber:
lgtm

PR #1854, opened 2016-05-02
V2 iptables overrides

abutcher:
Working towards #686 

* Default to using iptables in `os_firewall` and remove overrides from roles which depend on `os_firewall` (defaults won't override other defaults in ansible v2).
* Check and unmask iptables/firewalld if masked.


abutcher:
aos-ci-test

detiber:
:shipit:

PR #1773, opened 2016-04-19
Cloud provider cleanup

detiber:
Fixes up the way we are using some variables like ec2_region

Fixes https://github.com/openshift/openshift-ansible/issues/1760
Fixes https://github.com/openshift/openshift-ansible/issues/1826
Fixes https://github.com/openshift/openshift-ansible/issues/1825
Fixes #1933

Also fixes issues with using vars_files on an include.

PR #1856, opened 2016-05-03
Added extra install dependency on OSX

glnds:


detiber:
@glnds Thanks for the contribution.

I do want to make sure to point out that the bin/cluster scripts are not officially supported and are owned by the community.

I am working on some changes that fix issues with the bin/cluster scripts here: https://github.com/openshift/openshift-ansible/pull/1773

PR #1858, opened 2016-05-04
Fixed the credentials

vishpat:
@detiber These changes let a user override the default credentials for Nuage VSP.

### DIFF
--- a/roles/nuage_master/vars/main.yaml
+++ b/roles/nuage_master/vars/main.yaml
@@ -28,7 +28,3 @@ nuage_service_account_config:
 
 nuage_tasks:
     - policy add-cluster-role-to-user cluster-reader {{ nuage_service_account }} 
-
-nuage_master_cspadminpasswd: ''
-nuage_master_adminusername: 'admin'
-nuage_master_adminuserpasswd: 'admin'
--- a/roles/nuage_master/defaults/main.yaml
+++ b/roles/nuage_master/defaults/main.yaml
@@ -1,0 +1,4 @@
+---
+nuage_master_cspadminpasswd: ""
+nuage_master_adminusername: admin
+nuage_master_adminuserpasswd: admin 


roles/nuage_master/vars/main.yaml
```diff
@@ -29,6 +29,6 @@ nuage_service_account_config:
 nuage_tasks:
     - policy add-cluster-role-to-user cluster-reader {{ nuage_service_account }} 
 
-nuage_master_cspadminpasswd: ''
-nuage_master_adminusername: 'admin'
-nuage_master_adminuserpasswd: 'admin'
+nuage_master_csp_admin_passwd: "{{nuage_master_cspadminpasswd | default('') }}"
+nuage_master_admin_username: "{{nuage_master_adminusername | default('admin') }}"
+nuage_master_admin_userpasswd: "{{nuage_master_adminuserpasswd | default('admin') }}"
```
detiber: You want to move these to roles/nuage_master/vars/defaults.yaml if you want users to be able to override these values. It would look like this:
```
nuage_master_csp_admin_passwd: ""
nuage_master_admin_username: admin
nuage_master_admin_userpasswd: admin
```

If the variables are set in a playbook or inventory, then they would take precedence over the ones defined in the defaults.

vishpat: @detiber Makes sense. Thanks.

vishpat: @detiber I tried the suggested approach but the variable resolution in the template file fails

https://github.com/vishpat/openshift-ansible/blob/credentials/roles/nuage_master/templates/nuage-openshift-monitor.j2#L19

detiber: Man, I need to stop commenting from my phone...  The correct location is `roles/nuage_master/defaults/main.yaml`

vishpat:
@detiber Jason, have implemented and tested the suggestion

detiber:
@vishpat looks great, thanks! Will merge pending test results

detiber:
ok to test

detiber:
aos-ci-test

vishpat:
@detiber let me know if I need do something here. Thanks.

detiber:
aos-ci-test

detiber:
@vishpat re-running the tests, I don't suspect that the failure was related.

detiber:
aos-ci-test

arilivigni:
aos-ci-test

PR #1859, opened 2016-05-05
Containerized Upgrade Fixes

dgoodwin:
Fixes three issues: containers are upgraded during normal config playbook re-run, openvswitch will now be correctly restarted after upgrade (which is required to get the new container version), and a follow-up issue where the node is left down after this due to the ordering of the notification handlers. 

Some more details in the commits.

My testing procedure has been:

- launch 2 AWS atomic VMs.
- register and subscribe them
- install 3.1 prod
- add a custom registry that has 3.2 uploaded to it via Troy's scripts to the inventory, re-run config playbook
- run upgrade playbook

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
@@ -102,6 +102,7 @@
     target_version: "{{ '1.2' if deployment_type == 'origin' else '3.1.1.900' }}"
     openshift_docker_hosted_registry_insecure: True
     openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
+    upgrading: True
   handlers:
   - include: ../../../../../roles/openshift_master/handlers/main.yml
   - include: ../../../../../roles/openshift_node/handlers/main.yml
@@ -110,7 +111,7 @@
   # are modified to use the correct image tag.  However, this can trigger a
   # docker restart if new configuration is laid down which would immediately
   # pull the latest image and defeat the purpose of these tasks.
-  - openshift_cli
+  - { role: openshift_cli }
   pre_tasks:
   - name: Clean package cache
     command: "{{ ansible_pkg_mgr }} clean all"
--- a/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
@@ -130,6 +130,7 @@
     ent_reconcile_bindings: true
     openshift_docker_hosted_registry_insecure: True
     openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
+    upgrading: True
   tasks:
   - name: Verifying the correct commandline tools are available
     shell: grep {{ verify_upgrade_version }} {{ openshift.common.admin_binary}}
--- a/roles/openshift_cli/meta/main.yml
+++ b/roles/openshift_cli/meta/main.yml
@@ -12,6 +12,6 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
-- role: openshift_docker
+- { role: openshift_docker, upgrading: "{{ upgrading | default(False) }}" }
 - role: openshift_common
 - role: openshift_cli_facts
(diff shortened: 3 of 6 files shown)


brenton:
LGTM.  @detiber, mind taking a quick look?

sdodson:
aos-ci-test-extended

abutcher:
aos-ci-test-extended

dgoodwin:
Ok 3.1 containerized install has failed twice, I did extensive tests installing 3.1 from prod on atomic as that was an integral part of the upgrade testing. There's no way to see the logs from why it's failing as far as I can tell, I'm going to re-verify I can do one in AWS on atomic again just to be sure.

Beyond that I'm a bit lost on how to fix, anyone else seeing that error or have ideas?

sdodson:
@dgoodwin seeing the same, I've dumped some debugging ansible code into #1870 I'm trying to grok what's going wrong but I'm not sure. direct link to the log https://aos-ci.s3.amazonaws.com/openshift/openshift-ansible/jenkins-openshift-ansible-run-openshift-ansible-1033/4fcd7a3716e8dfef3e66decd580c5bf03f2f76b5.txt

sdodson:
@dgoodwin my working theory is that we're installing 3.2 but using 3.1 playbooks.

sdodson:
aos-ci-test-extended

detiber:
aos-ci-test

sdodson:
aos-ci-test-extended

dgoodwin:
Ok 3.1 upgrade test is failed because the recent CI fix is requesting openshift_image_tag=v3.1.1.6 for all tests now, which is going to fail if you try to upgrade with that. 

@abutcher could you take a quick look at this, I think if you're onboard we'll request a merge.

dgoodwin:
Ran into a noteworthy master conflict, branch updated.

aos-ci-test-extended

dgoodwin:
@abutcher @brenton what do you guys think? Hoping to get this merged soon for next build.

brenton:
I'm fine with this.  I'm a little bit worried that changing `openshift_image_tag` in the inventory will not longer have an effect unless it's an upgrade.  That's probably the right direction we want to go though so we'll just have to make it clear in documentation.

abutcher:
lgtm, I think we could rename `upgrading` to `openshift_docker_upgrading` or something in this case so we know what it impacts but a general `upgrading` variable might come in handy elsewhere.

abutcher:
aos-ci-test-extended

abutcher:
@detiber PTAL -- we should get this in for the update if we can.

detiber:
I think we should probably discuss this a bit after standup on Monday.

My concern is that the upgrading variable introduces counter-intuitive behavior for ongoing configuration management. It also goes against upstream recommended best practice of defining role interactions to be declarative (yes, we have done a poor job of ensuring this in the past, but I'd rather not build up more tech debt in this area if we can help it).

I think we need to either:
- make a change to the image_tag fact drive the upgrade determination rather than pushing that outside of the roles.
- rip out the logic to manage the image tag versions and put it move it into an ad-hoc playbook that is invoked on installation, scale-up, upgrade, etc.

dgoodwin:
I lieu of scrum I'll try to keep the discussion going here as we may want to consider getting something out the door ASAP for this problem, right now with 3.2 in the wild, anyone on a 3.1 containerized installation who re-runs config playbook is going to get their containers all bumped to 3.2. (seems quite serious) As such I think we should consider pushing something ASAP (if it's functionally valid) and then (a) I can go to work on the alternate approach immediately, or (b) we can roll it into the planned cleanup for upgrade. (a little more risky tech debt wise)

For the actual issues in the PR itself I could use some more info in the hopes of avoiding mistakes in the future, could you clarify the counter intuitive part? In english the PR makes it such that you must explicitly state that you wish to upgrade container versions, any normal use of the openshift_cli (and subsequently openshift_docker) role will leave current versions be if they are already there. 

Similar question around going against upstream ansible recommendation to be declarative. If by declarative we mean "describe what we want" instead of "how to get there", I'm not fully following where this is going against that. Arguably the "upgrading" variable sounds a bit too stateful, would "upgrade" = True be a better fit?

brenton:
I'm in favor of anything that unbreaks containerized upgrades right now.  Any surprises around changes in how `openshift_pkg_version` or `openshift_img_tag` potentially be printed as warnings in the short term.  I'm definitely liking the idea of soon only changing the installed version of a system during an upgrade playbook.  If this gets us even a step closer I'd like to have it tested and released while we work replacing the need for the `upgrading` variable.

@detiber, do you have any estimate of how much this PR would need to change for your suggestions?

abutcher:
@brenton Please merge per convo with @detiber. @dgoodwin and I are thinking about an alternative approach that takes our previous discussions on image_tag into account.

PR #1860, opened 2016-05-05
Allow ansible >= 1.9.4

abutcher:
Working towards #686.


detiber:
If you add a check on the upgrade to fail for 2.0+, then we can go ahead and get this merged.

abutcher:
aos-ci-test

sdodson:
LGTM

abutcher:
@detiber To show a clean "v2 is not supported for upgrades" message we would have to separate the `vars_files` changes from #1862, otherwise we'd end up seeing the `vars_files` error first thing. I'm leaning towards doing both at once if we can swing it.

PR #1867, opened 2016-05-09
Unmask atomic-openshift-master on uninstall

sdodson:
If you'd installed HA cluster this service would've been masked. If you then
uninstalled and installed a non HA cluster you run into atomic-openshift-master
service not being loaded.


detiber:
:+1:

PR #1868, opened 2016-05-09
WIP: Add new registries during upgrade playbook.

dgoodwin:
The situation with upgrades and pre.yml vs upgrade.yml is a bit
convoluted. Previously we were running some pre tasks, which included
updating the systemd units, which have automatic notifiers to restart
the services, and would attempt to pull down the new images immediately.
This should not be happening in pre.yml, but it also meant that we could
not actually pull the images if we're adding a new registry for this run
of the upgrade playbook. (i.e. you would have to re-run config playbook
first to get the new registry added)

This change drops the use of systemd unit file updates in pre.yml
completely, they were already re-applied in the actual upgrade plays.

We also stop using pre_tasks and then openshift_cli (which modifies the
wrappers around openshift tooling, i.e. a write operation) in pre.yml.

Instead we use the openshift_docker role which is a little safer in this
regard, and will update /etc/sysconfig/docker so we get our additional
registries and can safely lookup the new version.

openshift_cli role is now run early in the actual upgrade plays to
ensure the wrappers are updated appropriately.


dgoodwin:
@brenton when you get a chance could you let me know what you think about this experimental change. The situation relates to a comment you had in the pre.yml and I think the exact scenario you described as something we have to avoid was what is now happening. (services getting auto-restarted and pulling down new image despite the fact we're in pre.yml)

Commit details what I changed and why, I'm not 100% sure it's safe yet, it looks good on atomic but I need to verify RHEL + containerized as well as rpm based still.

brenton:
This is a much needed simplification.  :+1: 

I think part of the problem I was dealing with was related to the oc cli wrapper for containerized installs.  I suspect that was why it was running the tasks in pre and then running openshift_cli afterwards.  I remember seeing cases where the cli wrapper was running and pulling down a newer ose image which effectively upgraded the masters.  If you don't see anything like that happening then I'd say this is a much better route.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
```diff
@@ -110,8 +110,8 @@
   # are modified to use the correct image tag.  However, this can trigger a
   # docker restart if new configuration is laid down which would immediately
   # pull the latest image and defeat the purpose of these tasks.
-  - openshift_cli
-  pre_tasks:
+  - openshift_docker
```
detiber: Is the goal here to modify the docker config, or just to gather the fact data? If just the fact data, then openshift_docker_facts should be sufficient.

dgoodwin: Yeah goal was to modify docker config, I wanted to get new registries applied during upgrade so you don't need a re-run of config.yml to do that.

detiber: Okay, might want to push the modification outside of the pre-upgrade steps. Ideally, we would bail out if things do not look right before making any changes to the systems.

dgoodwin:
@brenton where would my best bet to see that ose image pull problem? If I were to bail out at the end of pre.yml, and check the images on the system, and I saw a 3.2.0.x image there, is this the issue to be on the lookout for?

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
dgoodwin: This is where things are getting very gnarly, we have a pre-upgrade step that figures out what new version to use for the container image, that makes perfect sense, but we can't do that if we don't have this new registry in /etc/sysconfig/docker.

dgoodwin: I could just yank that feature, it would be nice, but maybe it's best if you have to re-run config playbook to add a new registry prior to upgrade. Then pre.yml can lookup latest version, it won't need to run the openshift_docker role anymore, or update systemd unit files.

brenton:
@dgoodwin, I eventually added some optional checks to the playbook to verify things.  If the tests still look sane you could try them out.  I think you just need to set 'verify_upgrade_version=<whatever 3.2 version you should get'.

This PR is actually removing one of the sanity checks.  I think you'll want to add something equivalent back to verify that you're still running 3.1 until the upgrade plays are actually run.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
detiber: @dgoodwin What about this...

- oo_first_master:
  - Pre-upgrade checking prior to updating the docker config
  - apply openshift-docker role
  - Query new versions
- oo_all_hosts - oo_first_master
  - Pre-upgrade checking prior to updating the docker config
  - apply openshift_docker role
  - Validate new versions from oo_first_master are available
...

That could limit the potential issues that could arise from attempting to modify the docker configs all around.

dgoodwin: Hmm, am I following you correctly:

All of this would still take place in pre.yml. We already have a play for first master, all masters, all nodes, and all nodes + all masters. 

I'm assuming this would be adding a second, new play for just the first master, it would immediately follow the existing one for first master, it would apply docker openshift role and query new versions.

Then add a second new play for all hosts minus first master, which follows all existing verify steps, includes openshift_docker role, and then uses the version from the master.

So we'd still be modifying docker related things in pre.yml, just later in the process, and we'd gain a little more confidence that everybody will be trying to use the same version?

detiber: > Hmm, am I following you correctly:

> All of this would still take place in pre.yml. We already have a play for first master, all masters, all nodes, and all nodes + all masters.

Without breaking this into additional stages, yes (which I think is a really bad idea right now, but probably is a good idea for future upgrade work).

> I'm assuming this would be adding a second, new play for just the first master, it would immediately follow the existing one for first master, it would apply docker openshift role and query new versions.

It probably could be merged into the same play. The following oo_nodes_to_config play can be consolidated into the oo_masters_to_config:oo_nodes_to_config play, since we require masters be nodes ( to be extra safe, could conditionalize it on `when: 'node' in openshift`. 

> Then add a second new play for all hosts minus first master, which follows all existing verify steps, includes openshift_docker role, and then uses the version from the master.

If it's easier to just run it against all hosts, then that is fine as well, but it would save a small amount of time to exclude oo_first_master

> So we'd still be modifying docker related things in pre.yml, just later in the process, and we'd gain a little more confidence that everybody will be trying to use the same version?

Yeah, we would gain a canary for the docker changes on the master and we will avoid checking available versions on all hosts.

One thing that we might have to keep in mind is if a user is re-running a failed upgrade. I know @brenton and @sdodson have done some discussion here, but I'm not sure what the current state of handling for re-running upgrades is in place.

dgoodwin:
I've been attempting to implement the ideas above, in some places it's feeling like things are getting more complex rather than less, so here is the idea I half pitched to Jason on IRC for making pre-upgrade safer:

A remaining problem here is the use of openshift_docker, which looks up the version of openshift available by executing docker commands (which we need pre-upgrade), but depends on docker role, which may completely upgrade/downgrade/install docker. 

I am now thinking we should break the dependence between the two. 

Introduce a new role that handles /etc/sysconfig/docker called "docker_sysconfig", it will expect this file to exist, and docker to be installed and running already, if not it should fail. If it makes changes it will restart docker.

The docker role should then run the docker_sysconfig role at the end. (need to determine how this is possible)

openshift_docker role will no longer depend on "docker" role, rather the new "docker_sysconfig" role. I will examine all uses of openshift_docker and add in a dep on "docker" role for all of them. (excluding pre-upgrade)

In pre-upgrade, we now can use docker_sysconfig to ensure docker is present, running, and config is up to date with latest inventory settings, without running any risk of upgrading/downgrading it prior to real upgrade. We'll do this early in the playbook similar to how we depend on openshift_repos. 

I think after this we can leave most of the checking for available rpm/container version exactly as it is today in master, run on every master/node. Alternative would be to still break it out, check versions on first master, then make sure that version is available on all other hosts, but this will imply new shell scripts be written for both rpm/containers, and I'm not sure if there's much benefit there vs just checking for latest as we do now.

@brenton perhaps a quick one for you, you mention docker pulling a new image can actually upgrade the running container, could you point me to where that happens? My impression was it only boots the container referenced in /etc/sysconfig/atomic-openshift-master IMAGE_VERSION.

detiber:
@dgoodwin one thing to keep in mind. docker_sysconfig will need an openshift_docker_sysconfig wrapper in order to expose the openshift_* variables externally.

I don't see an issue with this, I just want to make sure that we don't expose openshift_* variables from a non-openshift role.

brenton:
@dgoodwin, I feel like it had something to do with `roles/openshift_cli/templates/openshift.j2`.  I don't see us ever doing any explicit docker pulls in that template.  I would just made sure there is some sort of sanity check test run pre-upgrade that makes sure the version we started with is still the current state.

dgoodwin:
Thanks guys, I will try to put this together.

dgoodwin:
@detiber when you get a chance, would you or anyone else be able to brain dump on your comment about exposing of openshift_ variables and wrapping with an openshift role. I'm not entirely aware how roles expose variables, or what our conventions are for this and why we do it. Is there anything out there I could read up on for that? AFAIK the sysconfig role shouldn't need to expose any variables. 

I see we don't really ever reference the docker role, only openshift_docker. If I break the dependency I need to then go add in use of the docker role most places we include openshift_docker. Is this ok for our conventions?

Is it more logical to just add openshift_docker_sysconfig instead of two levels of roles? This is sysconfig driven by openshift_variables.

dgoodwin:
WIP of the role split is pushed. I'm not sure it's looking so good. I have to go add inclusion of the docker role in places where we include the openshift_docker role, then discovered it has an implied dependency on openshift_docker_facts to know the version of docker to configure, so that dependency has to be explicitly added everywhere as well. (I don't think I can get away with having docker depend on openshift_docker_facts can I?)

Pondering just passing a flag into openshift_docker and then to docker, protect_docker_version.

detiber:
@dgoodwin I think at this point we should probably just table the improvements for the 3.3 upgrade.

dgoodwin:
Ok I will close this for now and return to bug queue.

PR #1869, opened 2016-05-09
Fix cli_docker_additional_registries being erased during upgrade.

dgoodwin:
Legacy options (cli_*) were not being migrated during upgrade. Add the
oo_all_hosts group, and migrate the facts as we do in the normal cluster
playbooks.


detiber:
@dgoodwin lgtm, is this ready to go?

dgoodwin:
@detiber I think so yeah.

PR #1863, opened 2016-05-06
Fix unsafe boolean handling on use_dnsmasq

sdodson:
Some installs seem to be getting openshift.master.dns_port = 8053 here https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_master/defaults/main.yml#L12-L15 but in the template openshift.master.dns_port = 53 https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_master/templates/master.yaml.v1.j2#L57

Before this change see that iptables and config mismatch
```
$ iptables -nL | grep 53 && grep -e 53 -e dns /etc/origin/master/master-config.yaml /etc/ansible/facts.d/openshift.fact"
master1
  ACCEPT     tcp  --  0.0.0.0/0            0.0.0.0/0            state NEW tcp dpt:8053
  ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            state NEW udp dpt:8053
  /etc/origin/master/master-config.yaml:dnsConfig:
  /etc/origin/master/master-config.yaml:  bindAddress: 0.0.0.0:53
master2
  ACCEPT     tcp  --  0.0.0.0/0            0.0.0.0/0            state NEW tcp dpt:8053
  ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            state NEW udp dpt:8053
  /etc/origin/master/master-config.yaml:dnsConfig:
  /etc/origin/master/master-config.yaml:  bindAddress: 0.0.0.0:53
master3
  ACCEPT     tcp  --  0.0.0.0/0            0.0.0.0/0            state NEW tcp dpt:53
  ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            state NEW udp dpt:53
  /etc/origin/master/master-config.yaml:dnsConfig:
  /etc/origin/master/master-config.yaml:  bindAddress: 0.0.0.0:53
```

After this change (apply patch, re-run config playbook)
```
master1
    ACCEPT     tcp  --  0.0.0.0/0            0.0.0.0/0            state NEW tcp dpt:8053
    ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            state NEW udp dpt:8053
    ACCEPT     tcp  --  0.0.0.0/0            0.0.0.0/0            state NEW tcp dpt:53
    ACCEPT     udp  --  0.0.0.0/0            0.0.0.0/0            state NEW udp dpt:53
    /etc/origin/master/master-config.yaml:dnsConfig:
[...]

PR #1870, opened 2016-05-09
Fix master firewall rules by deferring them

sdodson:
Items in defaults are evaluated prior to dependencies so we cannot rely on many of the facts being set properly until after openshift_common role has been called. Defer firewall rule configuration until that point.

Cleans up #1863 hack


sdodson:
WIP'ish need to do a bit more testing on my own

roles/openshift_master/meta/main.yml
```diff
@@ -18,3 +18,30 @@ dependencies:
...
+    port: 24224/tcp
+  - service: Fluentd td-agent udp
+    port: 24224/udp
+  - service: pcsd
+    port: 2224/tcp
+  - service: Corosync UDP
+    port: 5404/udp
+  - service: Corosync UDP
+    port: 5405/udp
+  os_firewall_deny:
```
detiber: We can probably remove the os_firewall_deny definition at this point, we haven't configured a cluster that way since pre-GA.

sdodson: cool, removed them

sdodson:
aos-ci-test

sdodson:
aos-ci-test

detiber:
@sdodson is this ready to go? changes lgtm

sdodson:
@detiber I just added a commit that moves os_firewall out of openshift_common and adds it to openshift_node which was the only role that configures firewall rules but doesn't call os_firewall on its own. I think it's better to have os_firewall included directly in each role that requires it rather than by way of openshift_common.

sdodson:
aos-ci-test

detiber:
@sdodson agree on moving the requires.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

PR #1875, opened 2016-05-10
Use IP4_NAMESERVERS rather than DHCP4_DOMAIN_NAME_SERVERS

### LINKED ISSUE #1882: No upstream DNS for dnsmasq for static IP configuration
In the NetworkManager dispatcher 99-origin-dns.sh there is an issue with configuring upstream DNS resolvers in case of static IP configuration.

If the IP address and DNS servers are configured statically, the variable DHCP4_DOMAIN_NAME_SERVERS is empty and no upstream servers are stored in /etc/dnsmasq.d/origin-upstream-dns.conf as upstream DNS resolvers. This leads to no DNS name resolution for statically configured network settings.

I've created a patch for the dispatcher that looks also for the DNS servers in the variable IP4_NAMESERVERS. With this change, DNS resolution is working in both cases - DHCP configured network settings and static configuration.

sdodson:
- Switches to IP4_NAMESERVERS
- Does not update resolv.conf if IP4_NAMESERVERS is undefined as it's not safe to do so

Fixes #1882

### DIFF
--- a/roles/openshift_node_dnsmasq/files/networkmanager/99-origin-dns.sh
+++ b/roles/openshift_node_dnsmasq/files/networkmanager/99-origin-dns.sh
@@ -8,10 +8,12 @@
 # a pod would fail.
 #
 # To use this,
-# Drop this script in /etc/NetworkManager/dispatcher.d/
-# systemctl restart NetworkManager
-# Configure node-config.yaml to set dnsIP: to the ip address of this
-# node
+# - If this host is also a master, reconfigure master dnsConfig to listen on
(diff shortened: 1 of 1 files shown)


sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

detiber:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

PR #1879, opened 2016-05-11
a-o-i: Better error handling

smunilla:
Some better error checking to handle input that previously generated
stack traces. Additionally, lay out some groundwork for upcoming 3.2
upgrades.

### DIFF
--- a/utils/src/ooinstall/cli_installer.py
+++ b/utils/src/ooinstall/cli_installer.py
@@ -510,7 +510,7 @@ def error_if_missing_info(oo_cfg):
         sys.exit(1)
 
     # Lookup a variant based on the key we were given:
-    if not oo_cfg.settings['variant']:
+    if oo_cfg.settings.get('variant', None) is None:
         click.echo("No variant specified in configuration file.")
         sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


smunilla:
pylint is failing. I'll see what I can do.

smunilla:
aoi-ci-test

detiber:
aos-ci-test

smunilla:
aos-ci-test

smunilla:
@sdodson When you get a free moment :smile:

utils/src/ooinstall/cli_installer.py
```diff
@@ -510,7 +510,7 @@ def error_if_missing_info(oo_cfg):
         sys.exit(1)
 
     # Lookup a variant based on the key we were given:
-    if not oo_cfg.settings['variant']:
+    if not 'variant' in oo_cfg.settings:
```
detiber: maybe use `if oo_cfg.settings.get('variant') is not None' here, so that you will also catch the case where the variant key is present in the config, but does not have a value set.

smunilla: Good point!

PR #1880, opened 2016-05-11
Fixes for openshift_docker_hosted_registry_insecure var.

dgoodwin:
Fixes a failure on masters if you explicitly set
openshift_docker_hosted_registry_insecure=true. This is the default but
if you tried to set it an error would trigger as a relevant variable was
not passed in the master playbooks.

Fixes setting the variable to false being ignored.
master/node playbooks were referencing the docker fact, which was not
set at that point and thus we were always getting the default of true,
regardless what was in your inventory.

It felt like in playbooks we should instead reference inventory vars,
rather than rely on more fact loading. This however led to a circular
reference because the variable was being passed as the same name as the
inventory variable. As such the internal variable name has been changed
to avoid this.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
@@ -100,7 +100,6 @@
   hosts: oo_masters_to_config:oo_nodes_to_config
   vars:
     target_version: "{{ '1.2' if deployment_type == 'origin' else '3.1.1.900' }}"
-    openshift_docker_hosted_registry_insecure: True
     openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
   handlers:
   - include: ../../../../../roles/openshift_master/handlers/main.yml
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -115,10 +115,6 @@
   vars:
     openshift_node_master_api_url: "{{ hostvars[groups.oo_first_master.0].openshift.master.api_url }}"
     openshift_node_first_master_ip: "{{ hostvars[groups.oo_first_master.0].openshift.common.ip }}"
-    # TODO: configure these based on
-    # hostvars[groups.oo_first_master.0].openshift.hosted.registry instead of
-    # hardcoding
-    openshift_docker_hosted_registry_insecure: "{{ openshift.docker.hosted_registry_insecure | default(True) }}"
     openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
   roles:
   - openshift_node
@@ -128,10 +124,6 @@
   vars:
     openshift_node_master_api_url: "{{ hostvars[groups.oo_first_master.0].openshift.master.api_url }}"
     openshift_node_first_master_ip: "{{ hostvars[groups.oo_first_master.0].openshift.common.ip }}"
-    # TODO: configure these based on
-    # hostvars[groups.oo_first_master.0].openshift.hosted.registry instead of
-    # hardcoding
-    openshift_docker_hosted_registry_insecure: "{{ openshift.docker.hosted_registry_insecure | default(True) }}"
     openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
   roles:
   - openshift_node
--- a/roles/openshift_docker_facts/tasks/main.yml
+++ b/roles/openshift_docker_facts/tasks/main.yml
@@ -35,7 +35,7 @@
     docker_options: >
       --insecure-registry={{ openshift.docker.hosted_registry_network }}
       {{ openshift.docker.options | default ('') }}
-  when: openshift.docker.hosted_registry_insecure | default(False) | bool
+  when: openshift.docker.hosted_registry_insecure | default(False) | bool and openshift.docker.hosted_registry_network is defined
 
 - set_fact:
     docker_options: "{{ openshift.docker.options | default(omit) }}"
--- a/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
@@ -27,6 +27,8 @@
   hosts: oo_masters_to_config:oo_nodes_to_config
   roles:
   - { role: openshift_docker_facts }
+  vars:
+    openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
   tasks:
   - name: Pull Images
     command: >
@@ -128,7 +130,6 @@
   vars:
     origin_reconcile_bindings: "{{ deployment_type == 'origin' and g_new_version | version_compare('1.0.6', '>') }}"
     ent_reconcile_bindings: true
-    openshift_docker_hosted_registry_insecure: True
     openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
   tasks:
   - name: Verifying the correct commandline tools are available
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -204,6 +204,7 @@
                                | oo_collect('openshift.common.all_hostnames')
                                | oo_flatten | unique }}"
     sync_tmpdir: "{{ hostvars.localhost.g_master_mktemp.stdout }}"
+    openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
   roles:
   - openshift_master_certificates
   post_tasks:
@@ -346,6 +347,7 @@
     openshift_master_count: "{{ openshift.master.master_count }}"
     openshift_master_session_auth_secrets: "{{ hostvars[groups.oo_first_master.0].openshift.master.session_auth_secrets }}"
     openshift_master_session_encryption_secrets: "{{ hostvars[groups.oo_first_master.0].openshift.master.session_encryption_secrets }}"
+    openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
   pre_tasks:
   - name: Ensure certificate directory exists
     file:
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1710,7 +1710,7 @@ class OpenShiftFacts(object):
                                     set_node_ip=False)
 
         if 'docker' in roles:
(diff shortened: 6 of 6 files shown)


dgoodwin:
@abutcher for your consideration if you get a chance to look, this is the result of several questions to you on IRC today. :)

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
```diff
@@ -100,8 +100,8 @@
   hosts: oo_masters_to_config:oo_nodes_to_config
   vars:
     target_version: "{{ '1.2' if deployment_type == 'origin' else '3.1.1.900' }}"
-    openshift_docker_hosted_registry_insecure: True
-    openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
+    docker_hosted_registry_insecure: True
```
abutcher: Do we need to hardcode this setting or can we yank the inventory variable here as well?

dgoodwin: I can't think of a reason we would want to ignore the inventory setting here. I will adjust.

dgoodwin:
aos-ci-test-extended

detiber:
aos-ci-test

dgoodwin:
aos-ci-test-extended

dgoodwin:
Upgrade test failures are expected right now and fixing is on hold until we can spend some time with Ari / Clint.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
```diff
@@ -100,8 +100,8 @@
   hosts: oo_masters_to_config:oo_nodes_to_config
   vars:
     target_version: "{{ '1.2' if deployment_type == 'origin' else '3.1.1.900' }}"
-    openshift_docker_hosted_registry_insecure: True
-    openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
+    docker_hosted_registry_insecure: "{{ openshift_docker_hosted_registry_insecure | default(True) }}"
```
sdodson: This wasn't working as is? I'd expect these to get passed down to openshift_docker as `openshift_docker_hosted_registry_insecure` and `openshift_docker_hosted_registry_network` where they're actually used. I think just switching from hardcoded to the defaulting is enough.

Need opinion of someone more versed in ansible than myself though. @detiber / @abutcher ?

abutcher: If we tried to do something like this here we'd get into a template loop BUT we require that both of these variables be set in playbooks prior to including a role that includes `openshift_docker`.

```
openshift_docker_hosted_registry_insecure: "{{ openshift_docker_hosted_registry_insecure | default(True) }}"
```

I think we could remove all `openshift_docker_hosted_registry_insecure` settings from playbooks and let `openshift_docker_facts` default the variable to `True` OR we could default to `True` wherever the value is accessed if the intent was to not persist this fact when the user didn't provide it.

We will continue to require passing `openshift_docker_hosted_registry_insecure` since it's driven by the first master's `portal_net` but this isn't a variable that is exposed to users so renaming here makes sense to me.

abutcher: At the same time, I like prefixing variable names with the role name...

abutcher: IMO we should remove the registry insecure setting from playbooks and maintain the `openshift_docker_hosted_registry_network` setting in playbooks with the `openshift_docker*` prefix.

detiber: > I think we could remove all openshift_docker_hosted_registry_insecure settings from playbooks and let openshift_docker_facts default the variable to True OR we could default to True wherever the value is accessed if the intent was to not persist this fact when the user didn't provide it.

Or we could default it to false in openshift_facts. I believe we need to make sure that this setting doesn't get exposed directly in the docker role, since it is an OpenShift specific configuration item, the hosted registry belongs to OpenShift (openshift_docker) and not Docker.

abutcher: > Or we could default it to false in openshift_facts.

I like this even more. +1

detiber: > IMO we should remove the registry insecure setting from playbooks and maintain the openshift_docker_hosted_registry_network setting in playbooks with the openshift_docker* prefix.

We still need a way to compute it, and prior to the deployment of the registry. If we defer the setting of it, then we are looking at an additional docker restart on provisioning.

detiber: Another approach is to use conditionally use set_fact when the variable is not already defined in pre_tasks.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
```diff
@@ -100,8 +100,8 @@
...
+    docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
```
detiber: I don't think we want to rename this from openshift_docker, since the registry network is an OpenShift concept and not a general docker concept.

playbooks/common/openshift-node/config.yml
```diff
@@ -131,8 +131,8 @@
     # TODO: configure these based on
     # hostvars[groups.oo_first_master.0].openshift.hosted.registry instead of
     # hardcoding
-    openshift_docker_hosted_registry_insecure: "{{ openshift.docker.hosted_registry_insecure | default(True) }}"
-    openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
+    docker_hosted_registry_insecure: "{{ openshift_docker_hosted_registry_insecure | default(True) }}"
+    docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
```
detiber: It seems really painful that we are setting this in so many different places. Granted we have copypasta all over the upgrade playbooks.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
abutcher: Can we not continue to set it like we have been? Everywhere we're including a role that includes `openshift_docker` from a playbook, set

```
openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
```

WRT registry insecure, we would set like below within `openshift_docker_facts` and within `openshift_facts` library default the value to `True` allowing the `openshift_docker_facts` to override.

```
openshift_facts:
  role: docker
  local_facts:
    hosted_registry_insecure: "{{ openshift_docker_hosted_registry_insecure | default(None) }}"
```

dgoodwin: > I believe we need to make sure that this setting doesn't get exposed directly in the docker role, since it is an OpenShift specific configuration item, the hosted registry belongs to OpenShift (openshift_docker) and not Docker.

By "exposed directly in the docker role" do you mean "used in" the docker role? (slightly confused on the terminology) My understanding, playbook includes a role, we pass it settings, those go into a semi-global namespace accessible to all dependent roles included so it would therefore be exposed to the docker role. Wondering if that's right and if so how we avoid that.

abutcher: > By "exposed directly in the docker role" do you mean "used in" the docker role?

Believe @detiber means that we don't want to use `openshift_docker_*` variables within the `docker` role to keep it openshift agnostic.

dgoodwin:
How does this look? Stopped passing in the registry insecure setting from playbooks, Andrew informed me that openshift_facts role can read it anyhow so we can just reference it directly there and avoid the whole circular dep in the template.

Added a default in openshift_facts as well.

dgoodwin:
Tested an existing cluster, no setting left the insecure registry option, setting to false removed it, setting explicitly to true brought it back.

abutcher:
LGTM pending a broken and sad CI run

abutcher:
aos-ci-test-extended

sdodson:
LGTM2

abutcher:
Non containerized upgrade caught:

```
TASK: [openshift_docker_facts | set_fact ] ************************************ 
fatal: [10.8.171.211] => One or more undefined variables: 'dict object' has no attribute 'hosted_registry_network'
fatal: [10.8.171.216] => One or more undefined variables: 'dict object' has no attribute 'hosted_registry_network'
fatal: [10.8.171.214] => One or more undefined variables: 'dict object' has no attribute 'hosted_registry_network'
```

dgoodwin:
aos-ci-test-extended

dgoodwin:
@sdodson I brought your commit over here, looks like your branch picked up a previous iteration of mine. Re-running tests now, thank you for the fix!

dgoodwin:
Hmm showing all checks have passed but upgrade ones are not in the list?

arilivigni:
aos-ci-test-extended

dgoodwin:
Problem is still there with @sdodson's patch, I will investigate asap.

dgoodwin:
aos-ci-test-extended

sdodson:
aos-ci-test-extended

dgoodwin:
The same error is surfacing out of an etcd role if you run an HA install with separate etcd, and at this point I'm thinking we need to guard that set_fact that's failing and make sure the hosted_registry_network is defined before we use it. Added this to the PR.

roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -35,7 +35,7 @@
     docker_options: >
       --insecure-registry={{ openshift.docker.hosted_registry_network }}
       {{ openshift.docker.options | default ('') }}
-  when: openshift.docker.hosted_registry_insecure | default(False) | bool
+  when: openshift.docker.hosted_registry_insecure | default(False) | bool and openshift.docker.hosted_registry_network is defined
```
abutcher: Shouldn't `openshift.docker.hosted_registry_network` always be set? We may need to make sure that the roles (master, node, etcd) that need it are getting a default set within `openshift_facts` lib.

abutcher: And if not set, there may be another place where we weren't passing from playbooks.

abutcher:
aos-ci-test-extended

roles/openshift_docker_facts/tasks/main.yml
dgoodwin: Discussed on IRC but the playbook still failing was etcd, and it didn't feel right requiring that one to pass in the registry network, so started looking at handling it more gracefully internally.

abutcher:
@brenton :shipit:

PR #1885, opened 2016-05-12
Fix mistaken quotes on proxy sysconfig variables.

dgoodwin:
Only need double quotes if preserving whitespace, and never singles.

CC @sdodson who looked at this too and discovered the rules regarding quoting in sysconfig env var files.


dgoodwin:
aos-ci-test-extended

sdodson:
LGTM

sdodson:
aos-ci-test

PR #1888, opened 2016-05-13
Update default max-pods parameter

maci0:
This updates the openshift_node_kubelet_args max-pods parameter from 40 to 110 which is the new number of maximum supported pods per node in OSE 3.2.
The inventory file should reflect this.


sdodson:
@maci0 can you update the other examples too?

maci0:
ok, done

PR #1897, opened 2016-05-16
Include sdn-ovs in upgrade commands to avoid pulling in newer packages

### LINKED ISSUE #1435: upgrade/v3_1_minor ignoring requested package version
Tried to upgrade from origin 1.1.0.1 to 1.1.2 using the byo/upgrade/v3_1_minor playbook. I had set in my inventory the lines:

[OSEv3:vars]
openshift_pkg_version=-1.1.2

which works fine for normal installation. But the openshift_pkg_version seems to have been ignored by the upgrade scripts, and lead to an upgrade to verison 1.1.3, not 1.1.2.

sdodson:
If the package isn't installed yum just ignores it.

Fixes #1435

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_1_minor/upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_1_minor/upgrade.yml
@@ -12,7 +12,7 @@
     openshift_version: "{{ openshift_pkg_version | default('') }}"
   tasks:
   - name: Upgrade master packages
-    command: "{{ ansible_pkg_mgr}} update -y {{ openshift.common.service_type }}-master{{ openshift_version }}"
+    command: "{{ ansible_pkg_mgr}} update-to -y {{ openshift.common.service_type }}-master{{ openshift_version }} {{ openshift.common.service_type }}-sdn-ovs{{ openshift_version }}"
     when: not openshift.common.is_containerized | bool
 
   - name: Ensure python-yaml present for config upgrade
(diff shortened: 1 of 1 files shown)


sdodson:
Need to fix 3.1 to 3.2 upgrade playbooks too.

detiber:
@sdodson lgtm, were you going to add the other upgrades to this PR as well?

sdodson:
> @sdodson lgtm, were you going to add the other upgrades to this PR as well?

I had planned to but it's a bit more involved and right now there's only one 3.2 release so it has no immediate benefit.

zeroem:
Was the `3.1 -> 3.2` upgrade ever addressed? We're attempting to upgrade from `origin 1.1.6` to `origin 1.2.1` and we're seeing this exact issue. Instead of upgrading to the desired version, it's upgrading to the latest released version which is currently `origin 1.4.1`.  Additionally, this only appears to happen on the nodes, the masters upgrade as expected.

sdodson:
@zeroem Hmm, the 3.2 playbooks were significantly refactored. This now happens in playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml and a component is passed in as a variable. We'd need to refactor that to have logic similar to what was introduced in this PR when component=='node'. Is that something you'd be interested in opening a pull request against the release-1.2 branch for?

sdodson:
Also, this problem is avoided in OCP because minor versions are delivered via separate repos. That will soon be the case for Origin 1.6 too, but I don't know that that will be done for 1.1-1.5.

PR #1899, opened 2016-05-16
Fix 'recursive loop detected in template string' for upgrading variable.

### LINKED ISSUE #1902: openshift_docker : set_fact ==> FAIL
I have tried to deploy openshift cluster. 
But I get this error on the master:
```
TASK [openshift_docker : set_fact] *********************************************
fatal: [172.17.0.6]: FAILED! => {"failed": true, "msg": "The conditional check 'upgrading is not defined' failed. The error was: recursive loop detected in template string: {{ upgrading | default(False) }}

The error appears to have been in '/home/lunik/paas/openshift-ansible/roles/openshift_docker/tasks/main.yml': line 6, column 3, but may

- set_fact:
  ^ here
"}
```

abutcher:
Always default `updrading` to `False` within the `openshift_docker` role. The `upgrading` playbook variable will take precedence when upgrading.

@dgoodwin PTAL

Resolves #1902

### DIFF
--- a/roles/openshift_cli/meta/main.yml
+++ b/roles/openshift_cli/meta/main.yml
@@ -12,6 +12,6 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
-- { role: openshift_docker, upgrading: "{{ upgrading | default(False) }}" }
+- role: openshift_docker
 - role: openshift_common
 - role: openshift_cli_facts
--- a/roles/openshift_docker/defaults/main.yml
+++ b/roles/openshift_docker/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+upgrading: False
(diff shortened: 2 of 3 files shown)


abutcher:
Also, I think this error is specific to v2.

roles/openshift_cli/meta/main.yml
```diff
@@ -12,6 +12,6 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
-- { role: openshift_docker, upgrading: "{{ upgrading | default(False) }}" }
```
abutcher: This is the recursive loop.

abutcher:
aos-ci-test-extended

dgoodwin:
Seems good to me. We really gotta get some 2.0 CI coverage, I will switch over myself as well.

abutcher:
aos-ci-test-extended

roles/openshift_cli/meta/main.yml
dgoodwin: Knowing what I know now about how we pass vars in from playbooks, and then they're visible to all roles, is an explicit passthrough of this variable needed here? I'm wondering now if it's necessary at all...

abutcher: No, not necessary. Would  have been nice for humans to follow though.

abutcher:
@detiber ptal

PR #1905, opened 2016-05-17
Use the cluster hostname while generating certificates for the master

vishpat:
@detiber These changes use the openshift cluster hostname while generating SSL certificates on the master. This change is required for HA.

### DIFF
--- a/utils/src/ooinstall/cli_installer.py
+++ b/utils/src/ooinstall/cli_installer.py
@@ -11,6 +11,7 @@ from ooinstall import OOConfig
 from ooinstall.oo_config import OOConfigInvalidHostError
 from ooinstall.oo_config import Host
 from ooinstall.variants import find_variant, get_variant_version_combos
+from distutils.version import LooseVersion
 
 DEFAULT_ANSIBLE_CONFIG = '/usr/share/atomic-openshift-utils/ansible.cfg'
 DEFAULT_PLAYBOOK_DIR = '/usr/share/ansible/openshift-ansible/'
@@ -594,7 +595,8 @@ https://docs.openshift.com/enterprise/latest/admin_guide/install/prerequisites.h
         oo_cfg.settings['master_routingconfig_subdomain'] = get_master_routingconfig_subdomain()
         click.clear()
 
-    if not oo_cfg.settings.get('openshift_http_proxy', None):
+    if not oo_cfg.settings.get('openshift_http_proxy', None) and \
+       LooseVersion(oo_cfg.settings.get('variant_version', '0.0')) >= LooseVersion('3.2'):
         http_proxy, https_proxy, proxy_excludes = get_proxy_hostnames_and_excludes()
         oo_cfg.settings['openshift_http_proxy'] = http_proxy
         oo_cfg.settings['openshift_https_proxy'] = https_proxy


roles/nuage_master/tasks/certificates.yml
```diff
@@ -10,7 +10,7 @@
 
 - name: Create the req file
   command: >
-    openssl req -key "{{ nuage_ca_master_rest_server_key }}" -new -out "{{ nuage_mon_rest_server_crt_dir }}/restServer.req" -subj "/CN={{ ansible_nodename }}"
+    openssl req -key "{{ nuage_ca_master_rest_server_key }}" -new -out "{{ nuage_mon_rest_server_crt_dir }}/restServer.req" -subj "/CN={{ openshift_master_cluster_hostname }}"
```
abutcher: You may want to pass in `openshift.common.cluster_hostname` from the playbooks as a nuage specific variable incase `openshift_master_cluster_hostname` is unset in the inventory.

detiber: Do we set cluster_hostname for non-ha installs? should there be a fallback to using public_hostname?

abutcher: You're right, I don't think we do.

vishpat: In our documentation we were planning to mandate the setting of the openshift_master_cluster_hostname for both HA and non-HA mode. In case of HA it would be the load balancer host while in non-HA it would be the single master hostname.

But I would do what is right. I am still not clear what variable should I use ?

abutcher: You could go that route OR in the case of HA use the `openshift.master.cluster_hostname` variable (assuming internal traffic) and fall back to using the first master's hostname (`groups.oo_first_master.0`) when not HA.

What do you think @detiber?

detiber: Since nuage_master requires openshift_facts and should not be run on a host that openshift_master has not been applied to, you can use: `openshift.master.cluster_hostname | default(openshift.common.hostname)`

vishpat:
@detiber @abutcher Have made changes as per the review comments.

vishpat:
@abutcher @detiber My commit history for this change is screwed up. Should I created a new PR ?

abutcher:
@vishpat You should be able to do a soft reset (`git reset --soft upstream/master`) and re-commit the added changes. Then you could force push to this branch and maintain the PR.

abutcher:
@vishpat I noticed that the `nuage_node` role makes use of `openshift_master_cluster_hostname` as well. You can use something like this to accomplish the same defaulting for `nuage_node`.

In playbooks/common/openshift-node/config.yml
```
- role: nuage_node                                                                                                                                                                                               
  nuage_mon_rest_server_host: "{{ hostvars[groups.oo_first_master.0].openshift.master.cluster_hostname | default(hostvars[groups.oo_first_master.0].openshift.common.hostname) }}"
  when: openshift.common.use_nuage | bool
```

In roles/nuage_node/vars/main.yaml
```
nuage_mon_rest_server_url: "https://{{ nuage_mon_rest_server_host }}:{{ nuage_mon_rest_server_port }}"
```

PR #1909, opened 2016-05-18
Fix playbooks/openshift-master/library move to symlink

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1336647

### DIFF
--- a/openshift-ansible.spec
+++ b/openshift-ansible.spec
@@ -38,6 +38,7 @@ popd
 mkdir -p %{buildroot}%{_datadir}/%{name}
 mkdir -p %{buildroot}%{_datadir}/ansible/%{name}
 mkdir -p %{buildroot}%{_datadir}/ansible_plugins
+cp -rp library %{buildroot}%{_datadir}/ansible/%{name}/
 
 # openshift-ansible-bin install
 mkdir -p %{buildroot}%{_bindir}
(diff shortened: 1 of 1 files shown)


sdodson:
@tdawson Any thoughts on simplifying packaging? roles and playbooks are very intertwined. I'd almost prefer just shipping one 'openshift-ansible' and perhaps an 'openshift-ansible-docs' package if anything else. I found several dangling symlinks because we've sliced packaging up so much we lost some content.

detiber:
So, think the intertwinedness is something that we can address with the Ansible 2.0+ migration.

We'll be able to more easily embed filter plugins directly into roles, rather than having to have them defined at the top level and the playbooks themselves will contain less logic in them.

I like the idea of keeping the roles as a separate deliverable, since it allows for a way to consume the officially released roles with Ansible Tower without having to pull in the playbooks as well.

sdodson:
I just realized the dangling symlinks isn't an issue because the ansible.cfg loads them from another path. I've removed that change. Good to review now.

openshift-ansible.spec
```diff
@@ -78,6 +79,8 @@ popd
 %files
 %doc LICENSE.md README*
 %dir %{_datadir}/ansible/%{name}
+%{_datadir}/ansible/%{name}/library
+%ghost %%{_datadir}/ansible/%{name}/playbooks/common/openshift-master/library.rpmmoved
```
tdawson: typo - one too many % before datadir, should be  
+%ghost %{_datadir}/ansible/%{name}/playbooks/common/openshift-master/library.rpmmoved

tdawson:
It's a little wordy, but I think it's very good.  I would much rather have comments and documentation showing why something like that is put in, than not having any explanation.

sdodson:
@tdawson thanks, fixed now

sdodson:
@detiber ready to go CI jobs don't build rpms currently so not running them

detiber:
lgtm

PR #1913, opened 2016-05-18
docker-current was missing from the containerized atomic-openshift-no…

maci0:
…de.service file

This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1336780


sdodson:
@pmorie or @eparis This fixes a containerized node issue due to the introduction of double docker where /usr/bin/docker is actually a shell script that runs /usr/bin/docker-current.

I'm wondering if a better fix for this is to actually just install docker in the node container image rather than subjecting ourselves to potential package refactorings in the future.  Downside of that would be that we'd also potentially have docker client vs server version mismatch whenever docker on the host is upgraded. Thoughts?

maci0:
@sdodson simply removing the `-v /usr/bin/docker:/usr/bin/docker:ro ` works too..  because the image ships its own version of docker.. as long as its compatible with the host's docker API this should work too.
both approaches should be validated

sdodson:
@rhatdan or @lsm5 can you weigh in on the risks of mounting /usr/bin/docker-current into the node container? Using the version of docker inside the container will fall down whenever the host has a newer version of docker than that which is installed inside the container.

I'm going to go ahead and merge this but if that's not safe lets figure out a proper fix for this.

sdodson:
aos-ci-test

arilivigni:
aos-ci-test

rhatdan:
docker-current should work fine, even in the case where docker-daemon is newer then the docker-current.  Docker has worked to make sure these work together.

sdodson:
@detiber ready to go pending test results I'm actually surprised e2e tests haven't been failing for all containerized installs.

sdodson:
Nevermind, we're currently installing OSE 3.1 in our containerized tests which would be pinned to docker-1.8.2. It all makes sense now why CI didn't find this problem.

detiber:
@sdodson doh, you are right. We need to add tests for 3.2

detiber:
ok to test

Issue #1904, opened 2016-05-17
v3_1_to_v3_2 upgrade playbooks use default kubeconfig when running oadm/oc commands


detiber:
We should always be copying admin.kubeconfig to a temporary location and using that rather than relying on the users ~/.kube/config file, which may be set to the wrong context.

detiber:
@dgoodwin not sure if this is still an issue or not.

PR #1914, opened 2016-05-18
Task "Update router image to current version" failed, if router not in default namespace

jkroepke:
Also added hardcoded -n default to registry patch, since oc get has -n default, too.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/post.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/post.yml
@@ -37,7 +37,7 @@
   - name: Update router image to current version
     when: all_routers.rc == 0
     command: >
-      {{ oc_cmd }} patch dc/{{ item['labels']['deploymentconfig'] }} -p
+      {{ oc_cmd }} patch dc/{{ item['labels']['deploymentconfig'] }} -n {{ item['namespace'] }} -p
       '{"spec":{"template":{"spec":{"containers":[{"name":"router","image":"{{ router_image }}","livenessProbe":{"tcpSocket":null,"httpGet":{"path": "/healthz", "port": 1936, "host": "localhost", "scheme": "HTTP"},"initialDelaySeconds":10,"timeoutSeconds":1}}]}}}}'
       --api-version=v1
     with_items: haproxy_routers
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/post.yml
```diff
@@ -37,7 +37,7 @@
   - name: Update router image to current version
     when: all_routers.rc == 0
     command: >
-      {{ oc_cmd }} patch dc/{{ item['labels']['deploymentconfig'] }} -p
+      {{ oc_cmd }} patch dc/{{ item['labels']['deploymentconfig'] }} -n {{ item['namespace'] }} -p
```
detiber: I'm not sure this is the intended behavior, the above get uses `--all-namespaces` and not `-n default`.

This probably needs to be updated to somehow get the namespace from `haproxy_routers` instead.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/post.yml
```diff
@@ -52,7 +52,7 @@
   - name: Update registry image to current version
     when: _default_registry.rc == 0
     command: >
-      {{ oc_cmd }} patch dc/docker-registry -p
+      {{ oc_cmd }} patch dc/docker-registry -p -n default
```
detiber: No objections to this change, since the get for the registry is using `-n default`.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/post.yml
detiber: @dgoodwin this is probably a pretty big issue considering Cloud Forms integration relies on a router that is not in the default namespace.

playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/post.yml
jkroepke: Does openshift-ansible execute oc project default to enforce the correct context? What happens, if I execute ``oc project dev`` before I execute the ansible?

detiber: @jkroepke it does not. I previously created this issue: https://github.com/openshift/openshift-ansible/issues/1904 to address the use of the default user kubeconfig for these operations.

detiber:
ok to test

detiber:
aos-ci-test-extended

detiber:
@sdodson @abutcher any objections to this change?

abutcher:
@detiber No objections.

PR #1866, opened 2016-05-09
Fix docker downgrade

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1334187

```
TASK: [docker | Downgrade docker if necessary] ******************************** 
failed: [openshift-x.com] => {"changed": true, "cmd": ["yum", "downgrade", "-y", "docker-1.8.2"], "delta": "0:00:02.177351", "end": "2016-05-08 15:44:15.269851", "rc": 1, "start": "2016-05-08 15:44:13.092500", "warnings": ["Consider using yum module rather than running yum"]}
stderr: 

Transaction check error:
  file /etc/sysconfig/docker from install of docker-1.8.2-10.el7.x86_64 conflicts with file from package docker-common-1.9.1-40.el7.x86_64
  file /usr/bin/docker from install of docker-1.8.2-10.el7.x86_64 conflicts with file from package docker-common-1.9.1-40.el7.x86_64
```

PR #1916, opened 2016-05-18
Specify docker_version before calling docker role so we don't downgra…

sdodson:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1337253

We're not passing in docker_version so we're downgrading it. I think maybe we should be calling openshift_docker here instead?

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
@@ -17,6 +17,7 @@
     with_items:
     - role: docker
       local_facts:
+        docker_version: '1.9'
         openshift_image_tag: "v{{ g_new_version }}"
         openshift_version: "{{ g_new_version }}"
 


detiber:
@sdodson I think you are correct, openshift_docker should be used here instead of docker.

sdodson:
#1866 would make it so that the playbooks run but they'll upgrade, downgrade, upgrade, which is a mess. So I could see this as either P1 or P0.

sdodson:
I'm actually not able to reproduce the problem cited in the bug, in the bug they upgrade, then remove everything, then either run the installer or upgrade playbook and then ansible fails. I'm less sure this is an actual problem now.

detiber:
@sdodson did they by chance have a 3.1 system with docker 1.9 already installed somehow?

dgoodwin:
I'm seeing similar to what this customer is reporting in the branch I'm working on, unnecessary downgrades, as well as the file conflict. Some of it may be related to what I'm working on but perhaps not entirely.

My solution for the unnecessary downgrade was to modify the 'when' in the two tasks in openshift_docker_facts that set docker_version to run only if l_common_version != '0.0'. It's fairly easy for that var to come out as 0.0 I think at least in my branch and for containerized envs, and I specifically want docker_version unset for an initial run of the docker role so if docker is already installed, it will just use what's there to do the lookup of what version to install. (this is for when no image tag / pkg version / release is defined in inventory) Then we're only downgrading if it's necessary after that.

sdodson:
@dgoodwin you found that the root of your problems were the improper creation /usr/bin/docker-current as a directory or do you still think there's a problem here?

dgoodwin:
That clears up the rpm problem, I will have fixes in my branch that makes the unnecessary downgrade even less likely, for this fix I'd say safe to take but it's also spreading version number assumptions even further, if you want I could just tackle it with the refactor and keep the logic around what version number is required per release in just one place. (openshift_docker_facts)

sdodson:
Given I haven't been able to reproduce the problem I'm fine with waiting for your more complete solution.

PR #1917, opened 2016-05-18
Use cluster hostname while generating certificate on the master nodes

vishpat:
@detiber @abutcher Changes as per our discussion in the PR1905


detiber:
ok to test

PR #1918, opened 2016-05-18
Use systemctl restart docker instead of ansible service.

dgoodwin:
Ansible is doing a full service stop and then start, which does not
allow systemd to keep the containerized services up and running.
Switching to this will cause the related services to come back within a
few seconds.


dgoodwin:
aos-ci-test-extended

detiber:
lgtm pending test results.

abutcher:
aos-ci-test-extended

detiber:
aos-ci-test-extended

abutcher:
haha

sdodson:
aos-ci-test-extended

detiber:
I think them thar tests are stuck

sdodson:
:-( was hoping it was just a timing issue before but it appears that way.

PR #1927, opened 2016-05-20
add unit in seconds for metrics resolution

jfcoz:
unit was missing in METRIC_RESOLUTION={{ openshift.hosted.metrics.resolution }}s

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -46,7 +46,7 @@
       openshift_hosted_metrics_duration: "{{ lookup('oo_option', 'openshift_hosted_metrics_duration') | default(7) }}"
     when: openshift_hosted_metrics_duration is not defined
   - set_fact:
-      openshift_hosted_metrics_resolution: "{{ lookup('oo_option', 'openshift_hosted_metrics_resolution') | default(10) }}"
+      openshift_hosted_metrics_resolution: "{{ lookup('oo_option', 'openshift_hosted_metrics_resolution') | default('10s', true) }}"
     when: openshift_hosted_metrics_resolution is not defined
   roles:
   - openshift_facts
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1727,7 +1727,7 @@ class OpenShiftFacts(object):
                 metrics=dict(
                     deploy=False,
                     duration=7,
-                    resolution=10,
+                    resolution='10s',
                     storage=dict(
                         kind=None,
                         volume=dict(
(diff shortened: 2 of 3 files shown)


detiber:
ok to test

detiber:
aos-ci-test

abutcher:
Should we add the unit to the value instead (and update the default in openshift_facts)?

```
openshift_hosted_metrics_resolution="10s"
```

detiber:
@abutcher I agree, I think updating the default to have the unit and requiring the user to specify the unit would give more flexibility when setting the value.

roles/openshift_metrics/tasks/main.yaml
```diff
@@ -44,7 +44,7 @@
   shell: >
    {{ openshift.common.client_binary }} process -f \
    /usr/share/openshift/examples/infrastructure-templates/{{ hawkular_type }}/metrics-deployer.yaml -v \
-    HAWKULAR_METRICS_HOSTNAME=hawkular-metrics.{{ openshift.master.default_subdomain }},USE_PERSISTENT_STORAGE={{ hawkular_persistence }},METRIC_DURATION={{ openshift.hosted.metrics.duration }},METRIC_RESOLUTION={{ openshift.hosted.metrics.resolution }} \
+    HAWKULAR_METRICS_HOSTNAME=hawkular-metrics.{{ openshift.master.default_subdomain }},USE_PERSISTENT_STORAGE={{ hawkular_persistence }},METRIC_DURATION={{ openshift.hosted.metrics.duration }},METRIC_RESOLUTION={{ openshift.hosted.metrics.resolution }}s \
```
abutcher: Can remove the trailing `s` here.

playbooks/common/openshift-master/config.yml
```diff
@@ -46,7 +46,7 @@
       openshift_hosted_metrics_duration: "{{ lookup('oo_option', 'openshift_hosted_metrics_duration') | default(7) }}"
     when: openshift_hosted_metrics_duration is not defined
   - set_fact:
-      openshift_hosted_metrics_resolution: "{{ lookup('oo_option', 'openshift_hosted_metrics_resolution') | default(10) }}"
+      openshift_hosted_metrics_resolution: "{{ lookup('oo_option', 'openshift_hosted_metrics_resolution') | default('10s', true) }}"
```
abutcher: I don't think these set_facts are necessary? The default should be coming from [openshift_facts](https://github.com/openshift/openshift-ansible/blob/00eba039c9312fbd04cc05a8a890ef48f2311769/roles/openshift_facts/library/openshift_facts.py#L1725-L1764).

detiber: The set_fact here is for providing compatibility with bin/cluster and the way additional config can be specified.

abutcher: Got it.

detiber:
aos-ci-test

detiber:
bumping this back up. @jfcoz could you update the default in openshift_facts as well? Then I think this will be ready to go.

detiber:
aos-ci-test

abutcher:
aos-ci-test

PR #1936, opened 2016-05-24
add metricspublicURL in master-config template (same url as metrics d…

jfcoz:
When using openshift_hosted_metrics_deploy=true we can add metricsPublicURL: https://hawkular-metrics.{{ openshift.master.default_subdomain }}/hawkular/metrics in master-config

### DIFF
--- a/playbooks/adhoc/uninstall.yml
+++ b/playbooks/adhoc/uninstall.yml
@@ -25,11 +25,6 @@
     - set_fact:
         is_containerized: "{{ is_atomic or containerized | default(false) | bool }}"
 
-    - name: Remove br0 interface
-      shell: ovs-vsctl del-br br0
-      changed_when: False
-      failed_when: False
-
     - name: Stop services
       service: name={{ item }} state=stopped
       with_items:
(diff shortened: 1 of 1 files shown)


roles/openshift_master/templates/master.yaml.v1.j2
```diff
@@ -18,8 +18,8 @@ assetConfig:
 {% if 'logging_public_url' in openshift.master %}
   loggingPublicURL: {{ openshift.master.logging_public_url }}
 {% endif %}
-{% if 'metrics_public_url' in openshift.master %}
-  metricsPublicURL: {{ openshift.master.metrics_public_url }}
+{% if openshift_hosted_metrics_deploy | bool %}
+  metricsPublicURL: https://hawkular-metrics.{{ openshift.master.default_subdomain }}/hawkular/metrics
```
detiber: @jfcoz I think I would rather still allow a user to override the metrics_public_url if they wish.

@abutcher do you think it would be enough to set the default for openshift.master.metrics_public_url to the above URL if metrics_public_url is not set? Only other thing it would affect (that I can think of), is that we may need to inject the metrics_public_url value into the metrics deployer if deploying metrics.

abutcher: The `openshift_metrics` role is setting the URL like the above changes. Both configs should feed from the same variable while providing compatibility with `openshift_master_metrics_public_url` (if changed). I expect we'd want the variable to be `openshift_hosted_metrics_public_url` if we allowed overriding in either case and we could default to this value.


Within the `openshift_metrics` role:
```
HAWKULAR_METRICS_HOSTNAME=hawkular-metrics.{{ openshift.master.default_subdomain }}
```

PR #1937, opened 2016-05-24
adhoc playbook, docker-registry swift storage

maci0:
this sets up the docker registry to use swift as a backend,
based on the s3_registry playbook,
super hacky and unclean but works for me.

### DIFF
--- a/playbooks/adhoc/swift_registry/swift_registry.j2
+++ b/playbooks/adhoc/swift_registry/swift_registry.j2
@@ -1,0 +1,23 @@
+version: 0.1
+log:
+  level: debug
+http:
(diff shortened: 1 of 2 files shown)


detiber:
@maci0 I think I would rather this go into https://github.com/openshift/openshift-ansible-contrib where we are planning on moving community supported content. Eventually we will be moving the bulk of the "adhoc" playbooks over that way.

Ideally, swift integration would be rolled into the `openshift_registry` role, which *eventually* will be merged into the `openshift_hosted` role.

sdodson:
closing this pr lets open against openshift-ansible-contrib

PR #1939, opened 2016-05-24
add registry ip to no_proxy line adhoc playbook

maci0:
add registry ip to no_proxy line adhoc playbook
This needs to be done after deploying the registry, because internally OSE doesn't use the DNS name to reach the registry but the ip address
requires oc login first


detiber:
@maci0 I'm not a huge fan of this approach, since it breaks using the roles for ongoing configuration management. That said, I would be open to accepting this into https://github.com/openshift/openshift-ansible-contrib

detiber:
@sdodson thoughts?

mtnbikenc:
Closing this PR as it is stale.  Please reopen if you feel this patch is still required.

PR #1945, opened 2016-05-25
openshift_release / version / upgrade improvements

dgoodwin:
These changes bring the addition of a new openshift_release="3.2" variable for inventory, giving customers a more generic version they can use to install a specific release instead of having to know 3.1.1.6. 

Includes a significant refactor for how we handle versions throughout all roles, and safer behavior.

Changes for users:

- Can now specify openshift_release="3.2" to install a generic release. 
  - This value is treated as a seed, we only really use it for cluster setup to lookup a more specific version (i.e. 3.2.0.44), and then use that as the exact version to configure on all hosts.
  - This is the recommended path for choosing which version to configure.
  - RPM installations on RHEL still require you to configure the correct repositories beforehand for this to work correctly.
- Once a host is installed we try to never modify the running version, but there are some exceptions.
- openshift_image_tag and openshift_pkg_version are treated as overrides, if you specify these values in your inventory, we will make sure that is the version configured when you install and/or maintain your environment with the config playbooks. If you modify those values and re-run the playbooks, this could cause an upgrade and downtime.
- Added new "docker_version=1.9.1" inventory variable for specifying the version of Docker you wish to install, or upgrade to. Default will be to install the latest, and then leave it alone. This value is respected both in the normal config playbooks, as well as the 3.2 and Docker upgrade playbooks.
- Added new "docker_upgrade=False" inventory variable for performing a 3.2 upgrade without upgrading Docker.
- Docker upgrade playbook is now less 1.10 centric and can be re-used to upgrade to an arbitrary Docker version. Images will only be cleared on the system when crossing the Docker 1.10 barrier.

Changes within the roles/playbooks:

- Introduced new openshift_version role, responsible for making sure all related version variables are *always* set for a host. 
  - Role's default behavior is to protect the installed version if there is one, however a playbook can override this with a specific version, a flag to ignore the current version, and by specifying openshift_release, image_tag, and pkg_version.
  - If no version is installed already and no version variables were specified, the role will lookup the latest version available.
- Improved openshift.common.openshift_version fact, should now be a more reliable way to always detect the version of openshift installed on a host, if there is one.
  - Removed other unnecessary and redundant version facts.
- Roles use openshift_pkg_version for RPM installations, and openshift_image_tag for pulling images. We now can assume that these values are always set. 
- Playbooks do version initialization on first master only, then pass this value along to all other hosts.
- Upgrade has become dramatically simpler, no longer requires service file updates and unsafe restarts during the "pre" stage of the upgrade, effectively upgrading a containerized machine just to check if we can proceed with upgrade.
- Old upgrade paths are removed. (3.0 -> 3.1, 3.1 minor)
- Virtually all instances of pulling latest, or pulling images possibly causing unexpected side effects, have been eliminated. The openshift_version role makes sure all required variables are set, and after that we always pull and configure the use of images with an exact tag, never latest.
- Fixed unsafe Docker restarts on nodes. (without evac) Docker upgrade is now handled together with node upgrade so we only evac once for both operations. Each step is run only if the host is actually a node or a host running docker.


playbooks/common/openshift-master/config.yml
```diff
@@ -337,12 +361,54 @@
     with_items: openshift_master_named_certificates
     when: named_certs_specified | bool
 
-- name: Configure master instances
-  hosts: oo_masters_to_config
+- name: Configure first master instance
```
detiber: I'm not sure I'm a fan of introducing the duplication here. Granted quite a bit of it will be removed with Andrew's delegate_to changes, but I think the real issue is coming from the attempted use of openshift_version as both an inventory variable and a role variable in this case.

If there is a distinction between the inventory variable and the role variable, then there should be less of an issue with variable dependence and the need to break out the plays separately.

Maybe this is a good place to re-introduce the requested_version variable, only make it the variable that we document and expose through the inventory only. Then it should be possible to handle both the first master and subsequent master use-cases within the same play.

dgoodwin: This is actually on my todo list to clean up I just couldn't figure out any other way to do it to get going.

If I'm following, openshift_version isn't used as an inventory variable, it's only set in the playbooks/roles and it meant to be the calculated variable we actually use, after taking release / pkg_version / image_tag into account.

openshift_release would be the misleadingly similarly named inventory variable. 

So how can we just use one play here? It looked to me like I must (a) calculate openshift_version based on all other inputs for first master, (b) install first master so it has a openshift.common.version fact, and then (c) install all other masters. I can't think of a way to do this without separation.

I also just noticed the subsequent masters are not properly protecting their installed versions, i.e. if first master somehow got a newer version, a re-run of config playbook would upgrade the subsequent masters. Adding that to TODO as well.

I'm thinking I need another play in between these to sort out the openshift_version fact for non-first masters, but still requires separate copied plays.

dgoodwin: Hmm, although, if I calculated *precise* openshift_version for first master in the playbook, perhaps just by running the openshift_docker role and not much else, I could then set that on all masters, and then run configure on all masters just once. We'd be down to just one block.

detiber: > So how can we just use one play here? It looked to me like I must (a) calculate openshift_version based on all other inputs for first master, (b) install first master so it has a openshift.common.version fact, and then (c) install all other masters. I can't think of a way to do this without separation.

You could do this by using set_fact to conditionally set it based on whether the host is the "first master". The template in the set_fact task will be evaluated at run time, and as long as the play is still serialized we would know that we have already run on the first master.

Also, since the act of running openshift_facts updates the local facts for the host, by the time the subsequent masters are run, the fact variable would already be set.

> I also just noticed the subsequent masters are not properly protecting their installed versions, i.e. if first master somehow got a newer version, a re-run of config playbook would upgrade the subsequent masters. Adding that to TODO as well.

I'm not completely convinced they should, either. if the config playbooks are intended to keep the cluster configured and in sync, the version from the first master should be the starting point, I believe.

dgoodwin: > You could do this by using set_fact to conditionally set it based on whether the host is the "first master". The template in the set_fact task will be evaluated at run time, and as long as the play is still serialized we would know that we have already run on the first master.
>
> Also, since the act of running openshift_facts updates the local facts for the host, by the time the subsequent masters are run, the fact variable would already be set.

I was already mid-way on iterating on this setting of openshift_version and eliminating the duplicated blocks. The current implementation has a play for first master that just determines the initial openshift_version if possible in the playbook (depending on what you set in inventory), then calls the openshift_docker role (which is likely due for a rename soon), which will make absolutely sure openshift_version is then set to something fully specified. (i.e. it will look up latest if the inventory doesn't specify anything, and it will convert 3.2 to 3.2.0.44) When this play completes, we *know* openshift_version is set.

Then I just run another play for the other masters that sets openshift_version to openshift.common.version if it exists (i.e. protect what's already installed), otherwise the value calculated above for first master.

I think this should get us there, most of that code is live in openshift-master/config.yml if you'd like to take a look. (haven't removed the duplication yet but I could very soon) 



>> I also just noticed the subsequent masters are not properly protecting their installed versions, i.e. if first master somehow got a newer version, a re-run of config playbook would upgrade the subsequent masters. Adding that to TODO as well.

> I'm not completely convinced they should, either. if the config playbooks are intended to keep the cluster configured and in sync, the version from the first master should be the starting point, I believe.

Hmm, so you're saying for subsequent masters, always key them off the current version from the first master? I was operating on the "protect what is installed unless explicitly upgrading" goal, but that may have been node specific.

I assume for nodes, we still will preserve current version unless upgrading though correct?

dgoodwin: I would propose we make the goal a consistent "protect currently installed unless upgrading", rather than "keep cluster in sync", and make the latter the job of a specific playbook for updating the cluster to latest version in the current stream. Seem reasonable?

detiber: > then calls the openshift_docker role (which is likely due for a rename soon)

there will still be a need for the openshift_docker role, but splitting out the OpenShift version handling, etc into a separate role would make a lot of sense.

> I would propose we make the goal a consistent "protect currently installed unless upgrading", rather than "keep cluster in sync", and make the latter the job of a specific playbook for updating the cluster to latest version in the current stream. Seem reasonable?

I think my hangup here is what determines we are upgrading? This becomes even murkier when you consider the idealized stand-alone Ansible role use case:
- The input to the role specifies a specific version (or release):
  - If that version or release is already installed:
    - If the version matches, then do nothing
    - If the version is an upgrade, then initiate upgrade for role
    - If the version is a downgrade, error
- No version specified to role:
  - If already installed, do nothing
  - If not already installed, install latest

In the situations we've discussed so far, we want to lock the version installed across the cluster in some way to avoid drift. If we do this by keying off the first master, then we can ensure that deployments managed by the stand-alone playbooks/roles would eventually converge on a single version. For this case, and ideal scenario would be that simply changing the version or release for their deployment would trigger an upgrade that would start with the first master and roll through the rest of the cluster.

Obviously, we are quite far from that being reality, but having to use some externally controlled toggle to control the behavior outside of that seems a bit off.

What if there was some comparison done between the user provided variable and the one derived from the master? If the host is getting release 3.2 from the user provided variable and the master is telling us that we should be installing 3.3, then maybe some other behavior is triggered? Possibly failure or do skip package changes?

dgoodwin: > there will still be a need for the openshift_docker role, but splitting out the OpenShift version handling, etc into a separate role would make a lot of sense.

Can you clarify what the need is for openshift_docker? I think there's literally nothing left in my branch beyond setting this version. Do we need it as a wrapper around docker for some reason,  or should I just flip all the callers to just require "docker" and rename it to openshift_version.

> I think my hangup here is what determines we are upgrading? This becomes even murkier when you consider the idealized stand-alone Ansible role use case:

Per our discussions the playbook (i.e. orchestrator) has full control over this. Roles will do what they're told, or if they're not told any version, they'll install latest. For config playbooks it behaves as described above, but the major upgrade playbook will specify openshift_version=3.3 for first master, run openshift_docker role to convert that to 3.3.y.z, and then re-use that value for all the remaining hosts. A minor upgrade playbook can do similar.

Does that help at all? 

If not I can ditch the "don't mess with running versions unless upgrading" goal we planned, and instead focus on keeping the cluster in sync with first master. (possibly with some version comparison and related actions, I'll have to think on that a bit more)

detiber: > Can you clarify what the need is for openshift_docker? I think there's literally nothing left in my branch beyond setting this version. Do we need it as a wrapper around docker for some reason, or should I just flip all the callers to just require "docker" and rename it to openshift_version.

So, openshift_docker is what forces openshift_docker_facts to run before openshift_docker (through the meta/main.yml dependencies). While another role could fill that need, I think it makes it a bit less intuitive what the relationship is between openshfit_docker_facts and docker is.

> Per our discussions the playbook (i.e. orchestrator) has full control over this. Roles will do what they're told, or if they're not told any version, they'll install latest. For config playbooks it behaves as described above, but the major upgrade playbook will specify openshift_version=3.3 for first master, run openshift_docker role to convert that to 3.3.y.z, and then re-use that value for all the remaining hosts. A minor upgrade playbook can do similar.

Yes, that is the desired behavior.

> Does that help at all?

> If not I can ditch the "don't mess with running versions unless upgrading" goal we planned, and instead focus on keeping the cluster in sync with first master. (possibly with some version comparison and related actions, I'll have to think on that a bit more)

I guess I'm not seeing where the version locking happens in the above situation, and how we can ensure that the locking behavior is triggered under the right conditions only.

If the role only does what it's told with respect to version, then version locking would have to happen external to the role somehow, which seems overly complex.

If we introduce an `upgrade` variable into the role, then we make the stand-alone role interface odd to use outside of our blessed playbooks.

dgoodwin: Yeah it's presently the playbook/orchestrator's role to determine openshift_version from inventory vars, or protect the installed version by passing in the current version from pre-existing fact. You've hit upon the reason why that is, we can't have the role protect the installed version fact or we'd need an upgrade flag, which we're already trying to avoid.

Is there another option? 

What if I consider openshift_version an override instead of something that's almost always set, and push more of the logic into openshift_docker role.

- config playbook for first master passes no openshift_version
- openshift_docker role inherits responsibility of sorting out the various inventory vars and setting openshift_version:
  - if openshift_version passed use it (still converts generic 3.2 to specific 3.2.0.44 if necessary)
  - else use installed version if present
  - else use inventory vars if present (release / pkg_version / image_tag)
    - only openshift_docker will reference these, everything else uses openshift_version
  - else use lookup latest and use that
- config playbook for subsequent masters and nodes will pass openshift_version = first master installed version fact
- upgrade playbook will need to determine always set openshift_version.

Now the role defaults to protecting installed version, and installing latest if nothing is requested. It involves no upgrade flag, but orchestrator can request a specific version that takes over everything for upgrades and/or pinning to master version.

We do have to choose one of (a) protect installed versions and (b) sync everything to first master. Which one will customers prefer? 

I believe the above can solve both cases, we'd just have the config playbook *also* check for pre-existing version and pass it as openshift_version, otherwise pass first master.

detiber: > What if I consider openshift_version an override instead of something that's almost always set, and push more of the logic into openshift_docker role.

As far as the roles are concerned, I believe this should always be the case.

> config playbook for first master passes no openshift_version
openshift_docker role inherits responsibility of sorting out the various inventory vars and setting openshift_version:
if openshift_version passed use it (still converts generic 3.2 to specific 3.2.0.44 if necessary)
else use installed version if present
else use inventory vars if present (release / pkg_version / image_tag)
only openshift_docker will reference these, everything else uses openshift_version
else use lookup latest and use that

What is the plan for looking up the latest with containerized installs? Just pull the `latest` tag? If so, I'm good with that approach, we'd just need to keep in mind that we'd need to take care when testing this internally where the latest tags are broken (which I'm fine with, we'd just need to make sure it's documented somewhere).

> config playbook for subsequent masters and nodes will pass openshift_version = first master installed version fact
upgrade playbook will need to determine always set openshift_version.
Now the role defaults to protecting installed version, and installing latest if nothing is requested. It involves no upgrade flag, but orchestrator can request a specific version that takes over everything for upgrades and/or pinning to master version.

> We do have to choose one of (a) protect installed versions and (b) sync everything to first master. Which one will customers prefer?

> I believe the above can solve both cases, we'd just have the config playbook also check for pre-existing version and pass it as openshift_version, otherwise pass first master.

I can live with that, though doing that in a way that wouldn't require pre-loading the host facts may prove difficult. The easiest way I can think of to do that would be to use `vars` on the role invocation, in which case the template would be evaluated at the time it is used by a task within the role.

The alternative could be that to get 'a', the user would need to explicitly set the version variable.

gravis:
Guys, this is really a must-have, as the current playbook is installing 1.3.0, which is in *alpha*.
Thanks

detiber:
@gravis https://github.com/openshift/openshift-ansible/pull/1685 is a better short term fix. I don't want to rush this PR and risk have to tackle another major refactoring for this in another couple of months

playbooks/common/openshift-master/config.yml
dgoodwin: > What is the plan for looking up the latest with containerized installs? Just pull the latest tag? If so, I'm good with that approach, we'd just need to keep in mind that we'd need to take care when testing this internally where the latest tags are broken (which I'm fine with, we'd just need to make sure it's documented somewhere).

Yeah just pulling latest tag, if we're ruling out hard coding in openshift-ansible I can't think of another reasonable option. Internal testers will have several options for overriding at least.

> I can live with that, though doing that in a way that wouldn't require pre-loading the host facts may prove difficult. The easiest way I can think of to do that would be to use vars on the role invocation, in which case the template would be evaluated at the time it is used by a task within the role.

I think facts are already loaded at the point we start applying these roles but I'll keep an eye out for trouble here in case I missed something.

dgoodwin: Oh I think I see the problem you were alluding to now, how do I not pass openshift_version for first master, but then pass for everyone else if they're all handled by the same play after removing the duplication. Looking into your suggestions now.

dgoodwin: This has been reworked, the duplication is gone, the openshift_version is calculated for first master and set for all other hosts *very* early in the process via a playbook include. I believe this is required when you start installing containerized etcd envs and don't want to end up installing docker, then downgrading later.

filter_plugins/oo_filters.py
```diff
@@ -881,14 +881,13 @@ def oo_image_tag_to_rpm_version(version, include_dash=False):
         """
         if not isinstance(version, basestring):
             raise errors.AnsibleFilterError("|failed expects a string or unicode")
-        # TODO: Do we need to make this actually convert v1.2.0-rc1 into 1.2.0-0.rc1
-        # We'd need to be really strict about how we build the RPM Version+Release
```
detiber: Man, the new origin tagging is going to throw a wrench into this...  Maybe we can tackle that in packaging, where we if upstream specifies something that would look like a 'release' we use it as the 'release' in the subsequent packaging. 

I can't think of a reason that we would need to track a separate release than upstream, and if so, maybe we can just tack on a second "release", so 1.2.0-rc1-0 ?  I must say that I'm pretty ignorant of t he edge cases of RPM versioning, though.

@sdodson @tdawson thoughts?

filter_plugins/oo_filters.py
```diff
@@ -881,14 +881,13 @@ def oo_image_tag_to_rpm_version(version, include_dash=False):
...
         if version.startswith("v"):
-            version = version.replace("v", "")
+            version = version[1:]
+            # Strip release from requested version, we no longer support this.
             version = version.split('-')[0]
```
detiber: Of course, with the nonsense I'm talking about above, release would once again come into play, at least in a subset of cases.

filter_plugins/oo_filters.py
dgoodwin: I have not gotten to the origin use case much here, I assumed you can specify openshift_release="1.2", and get "v1.2", whatever that is tagged at. If you want to install container tags for an rc you'd specify openshift_image_tag=v1.2.0-rc1-0.

playbooks/common/openshift-cluster/initialize_openshift_version.yml
```diff
@@ -0,0 +1,32 @@
...
+# NOTE: We set this even on etcd hosts as they may also later run as masters,
+# and we don't want to install wrong version of docker and have to downgrade
+# later.
+- name: Set openshift_version for all hosts
+  hosts: oo_all_hosts
+  tasks:
+  - debug: var=hostvars[groups.oo_first_master.0].openshift.common.version
+  - debug: var=openshift.common.version
+  - debug: var=openshift_version
+    # TODO: Should we use the first master's "openshift_version" var instead of a fact? Could go to just openshift_docker role above, and skip CLI config this early.
```
detiber: It should be possible to use the openshift_version from the first master, and avoiding set_fact if possible is always nice. One catch is that you'll need to pass in the data from the playbook, rather than trying to introspect the data from the group membership within the roles.

playbooks/common/openshift-master/config.yml
```diff
@@ -310,12 +310,19 @@
...
   any_errors_fatal: true
   serial: 1
   vars:
     sync_tmpdir: "{{ hostvars.localhost.g_master_mktemp.stdout }}"
+
+    # Do not pass a version if this is first master. (openshift_docker will sort it out)
+    # For subsequent masters we pass either pre-existing version for the master (if already installed),
+    # otherwise the first master version.
+    openshift_version: "{{ openshift.common.version if openshift.common.version is defined else oo_first_master.openshift.common.version.split('-')[0] }}"
+    when: inventory_hostname != groups.oo_first_master.0
```
detiber: Does this work (and if so, does it work for both ansible 2.0+ and 1.9)? I didn't think it was possible to use a when clause in the vars section. I have been under the impression that the only supported way to do this type of thing is with set_fact.

playbooks/common/openshift-cluster/initialize_openshift_version.yml
dgoodwin: I actually implemented this TODO yesterday and forgot to push, it's up now.

roles/openshift_ca/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
+---
```
detiber: rebase needed?

roles/openshift_common/meta/main.yml
```diff
@@ -13,4 +13,4 @@ galaxy_info:
   - cloud
 dependencies:
 - role: openshift_facts
-- role: openshift_repos
+- role: openshift_docker
```
detiber: I'm not sure I'm a fan of making openshift_docker a dependency of openshift_common. This is going to force docker to be installed on hosts that may not need it. The existing conditional dependencies should handle the case of installing docker where necessary only.

playbooks/common/openshift-master/config.yml
dgoodwin: It appears to be working yes, I got the example from openshift_no_proxy_internal_hostnames just below this. I'll keep an eye out and make sure I'm not assuming it's working on my next HA run.

roles/openshift_common/tasks/main.yml
```diff
@@ -29,17 +29,6 @@
       data_dir: "{{ openshift_data_dir | default(None) }}"
       use_dnsmasq: "{{ openshift_use_dnsmasq | default(None) }}"
 
-# Using oo_image_tag_to_rpm_version here is a workaround for how
-# openshift_version is set.  That value is computed based on either RPM
-# versions or image tags.  openshift_common's usage requires that it be a RPM
-# version and openshift_cli expects it to be an image tag.
-- name: Install the base package for versioning
-  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}{{ openshift_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
-  when: not openshift.common.is_containerized | bool
```
detiber: Of course, this was another issue of installing something where it wasn't necessary as well...  openshift_common is used by components that don't strictly need OpenShift channel access (haproxy, storage, etcd). Some of these have been moved to relying on openshift_facts directly, but I'm not sure all have.

roles/openshift_common/meta/main.yml
dgoodwin: This was done because I now needed the base package for versioning installation in openshift_docker (now called openshift_version). I could make this a dep on openshift_version now instead which should dodge useless docker installation, or just duplicate the two tasks involved.

dgoodwin: Nevermind, gotta duplicate the tasks, openshift_version depends on openshift_docker.

roles/openshift_version/meta/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+galaxy_info:
+  author: Jason DeTiberus
+  description: OpenShift Docker
```
detiber: author, description fields need updating.

roles/openshift_version/tasks/set_version_rpm.yml
```diff
@@ -0,0 +1,23 @@
...
+    # Expects a leading "-" in inventory, strip it off here, and ignore a trailing release,
+    # openshift_version should always just be "3.2" or "3.2.0.44"
+    openshift_version: "{{ openshift_pkg_version[1:].split('-')[0] }}"
+  when: openshift_pkg_version is defined
+
+# We do not have as fine grained control over version to install presently with rpms,
+# we just assume the user has configured the correct repos and install the rpm requested.
+- name: Install latest rpm available in configured repos to check version
+  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}{{ openshift_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
+  when: openshift_version is not defined
```
detiber: Would it be better here to just do a repoquery rather than an install?

playbooks/common/openshift-master/config.yml
detiber: @sdodson have you checked on whether the when clause is fully supported here? I don't want us to fall into another situation like we did with the vars_files on includes :)

roles/openshift_version/tasks/set_version_rpm.yml
dgoodwin: Yes I think it might, good thinking. Then I could move that back to common and just leave it alone.

roles/openshift_common/meta/main.yml
detiber: Do we actually need to install the package? or can we get away with just doing a repoquery for what is available?

I know there are situations where we need the version reported from the client for getting the associated image versions, but if that logic is separated from the package installation logic, then we should be good.

playbooks/common/openshift-master/config.yml
sdodson: Yes, sounds like this isn't supported and it likely actually just sets a variable named 'when'

detiber: We'll need to push these into set_fact tasks then.

roles/openshift_common/meta/main.yml
dgoodwin: Probably not, per other comment we can likely just query in version, and leave this guy as he was.

playbooks/common/openshift-master/config.yml
dgoodwin: Added a todo, will do that once I can get back to this.

filter_plugins/oo_filters.py
dgoodwin: Added a task for this, will need more details on these tags you're referring to.

detiber: I'm not even sure if we are doing anything with them currently, but the tags would be the ones used for pulling containers used by openshift, for example the router and/or registry. I think they may be defaulting to using the templates defined in the master config (or the version from the client itself) now, though.

playbooks/common/openshift-master/config.yml
dgoodwin: This sort of leads back to the need to duplicate the master config blocks. We need to setup first master without passing a version in, let it figure it out, then set_fact on other masters, then the config block for other masters.

I'm thinking now about just passing in a new var for the "primary_master_version" or something like this, then the subsequent masters can see if that's set, and use it if they don't have a version already.

dgoodwin: NOTE: This still implies I'm targetting "do not modify an already installed version unless this is an upgrade". We can change that to "always sync to first master version" very easily.

roles/openshift_common/meta/main.yml
dgoodwin: Hmm, I'm not seeing how we can avoid the dep. Even if I restore things back the way they were in common, openshift_common requires the openshift_version variable to know which package version to install, which means we must depend on the new openshift_version role, which depends on docker. 

Previously openshift_common did it's own assumptions about what openshift_version to use, which I am trying to avoid and use the role to calculate it.

I guess I could make the role dependencies conditional, so openshift_version would depend on openshift_docker_facts and docker roles only if it's a containerized install.

playbooks/common/openshift-master/config.yml
detiber: Hmm, I'm not sure how I feel about that, a separate variable seems a bit wonky to me, but if it simplifies the correct behavior, them I am good with it.

roles/openshift_common/meta/main.yml
detiber: The conditional makes sense to me here, we want to avoid installing something where it isn't needed

playbooks/common/openshift-master/config.yml
dgoodwin: I think we're ok, I think I discovered this morning that the whole vars passing here is unnecessary, we sort it all out with the initialize_openshift_version.yml stuff, it wasn't even being used here.

roles/openshift_common/meta/main.yml
dgoodwin: Will do.

dgoodwin:
My plan was to have *only* the openshift_version variable, it would be set by the new openshift_version role, it would only ever contain a fully qualified version like 3.2.0.44 (with exception of openshift_release where it's briefly "3.2" until we lookup the full latest version of that release), and we would use it everywhere for pulling images, in systemd IMAGE_VERSION variables, and when installing rpms.

Then I saw origin is now tagging images with 'v1.3.0-alpha.0'. The above would imply openshift_version of 1.3.0 and we we wouldn't be able to use it correctly anywhere.

My plan now is to have the openshift_version role resolve all possible inputs, (openshift_release, openshift_image_tag, and openshift_pkg_version), determine the full openshift_version, and then ensure all three of openshift_version/openshift_image_tag/openshift_release are set at all times after that. We'll use the image tag when pulling images, and the pkg_version when installing rpms. (I'm not sure if openshift_version var is useful anymore after this but I'll check) 

This should allow for mixed containerized/rpm environments, though I do not commit that this is working or tested yet, it just opens the door. It will also allow fine grained control over rpm release or image tag to be used.

dgoodwin:
@detiber do you have any thoughts on commit https://github.com/openshift/openshift-ansible/pull/1945/commits/8c9ed6719bdbc0ce7574fa9261c60e4978c203e6? 

I'm finding it extremely easy to accidentally use a role, not pass some variable required by one of it's dependencies, and thus change docker config triggering a restart without any evac, multiple times through a single playbook run.

In that commit I'm ensuring the value is set as soon as common facts are loaded and in general, I was thinking this is the safest thing, use set_fact not vars for these things. I saw this in master with some proxy vars and filed a bug on it there as well, it's a little bit too easy.

Any thoughts on what the right thing to do for this is?

filter_plugins/oo_filters.py
dgoodwin: I believe this concern is now covered, I've tested with the latest 1.3 alpha tags and openshift_image_tag, and it should be installable.

dgoodwin:
Removed WIP tag, ready for full review.

playbooks/common/openshift-cluster/initialize_facts.yml
```diff
@@ -8,3 +8,5 @@
       role: common
       local_facts:
         hostname: "{{ openshift_hostname | default(None) }}"
+  - set_fact:
+      openshift_docker_hosted_registry_network: "{{ hostvars[groups.oo_first_master.0].openshift.common.portal_net }}"
```
dgoodwin: @abutcher @detiber when review kicks off I wanna call this out specifically for attention. I ran into the problem where docker restarts multiple times in a config re-run because in using the docker role, if you forget to pass this var explicitly, the config changes. 

I don't really know what the right solution here is, I want defaults with a little logic for inventory vars, if they're not defined ideally they should be set to a default so we can just assume they're usable. That default might be programatic, and it might be based off another system like the first master, so role defaults don't feel quite right. 

My thought was we should have snippets we can include in playbooks that do all variable initialization, this facts file seemed like a decent first start.

I'd like to scale something like this up to cover a lot of other vars. I've found it really helpful in this branch to know based on my behavior of openshift_version role, that openshift_version + openshift_pkg_version + openshift_image_tag are all set and correct and can be used accordingly.

Maybe it's even a role, openshift_defaults...

inventory/byo/hosts.aep.example
```diff
@@ -24,6 +24,23 @@ debug_level=2
...
+
+# Specify an exact container image tag to install or configure.
+# WARNING: This value will be used for all hosts in containerized environments, even those that have another version installed.
+# This could potentially trigger an upgrade and downtime, so be careful with modifying this value after the cluster is set up.
+#openshift_image_tag=v3.2.0.46
+
+# Specify an exact rpm version to install or configure.
+# WARNING: This value will be used for all hosts in RPM based environments, even those that have another version installed.
+# This could potentially trigger an upgrade and downtime, so be careful with modifying this value after the cluster is set up.
+#openshift_image_tag=-3.2.0.46
```
abutcher: Should this be `openshift_pkg_version`?

playbooks/common/openshift-cluster/initialize_facts.yml
abutcher: We default `openshift.common.portal_net` in `openshift_facts` `get_defaults()`, could we default `openshift.docker.hosted_registry_network` in the same place?

playbooks/common/openshift-cluster/initialize_facts.yml
```diff
@@ -8,3 +8,8 @@
...
+  - set_fact:
+      repoquery_cmd: "{{ 'dnf repoquery --latest-limit 1 -d 0' if ansible_pkg_mgr == 'dnf' else 'repoquery' }}"
```
abutcher: Depending on the future of `openshift_facts` this may need to become a common fact also since we're using it in multiple places rather than setting it here.

Also calling attention to the updates to repoquery commands in https://github.com/openshift/openshift-ansible/pull/2101

roles/openshift_version/tasks/main.yml
```diff
@@ -0,0 +1,85 @@
...
+- set_fact:
+    openshift_version: "{{ openshift_version }}"
+  when: openshift_version is defined
+
+# Protect the installed version by default unless explicitly told not to, or given an
+# openshift_version already.
+- name: Use openshift.common.version fact as version to configure if already installed
+  set_fact:
+    openshift_version: "{{ openshift.common.version }}"
+  when: openshift.common.version is defined and openshift_version is not defined and openshift_protect_installed_version
```
abutcher: `... and openshift_protect_installed_version | bool`

inventory/byo/hosts.aep.example
dgoodwin: Yup, good catch.

playbooks/common/openshift-cluster/initialize_facts.yml
dgoodwin: I don't think we can primarily because it's the first master's fact we need, I think that implies we'd have to pass in the fact from the first master for all other systems, then set the fact on in openshift_facts on those systems. At that point I'm thinking it makes the most sense to be defined in the playbook/orchestration layer.

I've been thinking about this more since and I'm still feeling like we need a centralized place that sorts out *our* inventory vars, and makes sure they're set sanely throughout the roles. Roles should still have sensible defaults but in this case, it's a problem with that sensible default accidentally getting used when we don't want it to, because passing it individually for each role use is so fragile. This, or something like it, would then become a centralized yml you can include in any of our playbooks to initialize, evaluate groups would be a good candidate to merge into here too likely.

roles/openshift_version/tasks/main.yml
dgoodwin: Good eye thanks.

playbooks/common/openshift-cluster/initialize_facts.yml
dgoodwin: Hoping to apply this approach to fix additional issues such as:

https://bugzilla.redhat.com/show_bug.cgi?id=1347303

https://bugzilla.redhat.com/show_bug.cgi?id=1345804

detiber: So, I think this is tricky...  ideally anything we would want to coordinate at that level would be managed by a shared/common role that is required by all of the other roles involved...  We've had a bit of dysfunction around defining what that role was in the past, generally falling on openshift_facts, since openshift_common (at least in the past) pulled in deps that we weren't comfortable pulling in for auxiliary roles. 

In reality, it should probably be openshift_common that holds this centralized logic/validation for this type of stuff with a more specific role handling pulling in shared binaries for master, node, etc.

I really hate the idea of managing this type of complexity at the playbook level, since I think it actually makes it more difficult to enforce. That said, if the problem is solely that the vars are being set at too narrow of a scope, then we should address that and do what we can to set them at the proper scope.

As an additional aside, I think evaluate_groups is probably a candidate for re-evaluation in the not too distant future, as we split out the community supported bits into openshift-ansible-contrib and better defining the model for using openshift-ansible from Ansible and/or Ansible Tower directly. It served the purpose of bringing together the bin/cluster playbooks with the byo playbooks, but I think in the long run it's use adds complexity for very little real benefit, especially if we can better document and define the pure-ansible management model.

playbooks/common/openshift-cluster/initialize_openshift_version.yml
```diff
@@ -0,0 +1,29 @@
+---
+# NOTE: requires openshift_facts be run
+- name: Determine openshift_version to configure on first master
```
detiber: Ideally, I would like to see this pushed down into a role dependency, where we use delegate_to for getting the data for the first master (if we aren't the first master)

detiber: Granted that is something that can be addressed as a follow-up, since we are still struggling to break backwards compat for 1.9.4 currently.

playbooks/common/openshift-cluster/upgrades/docker/upgrade.yml
```diff
@@ -0,0 +1,44 @@
...
+  service: name={{ item }} state=stopped
+  with_items:
+    - "{{ openshift.common.service_type }}-master"
+    - "{{ openshift.common.service_type }}-master-api"
+    - "{{ openshift.common.service_type }}-master-controllers"
+    - "{{ openshift.common.service_type }}-node"
+    - etcd_container
+    - openvswitch
+  failed_when: false
+  when: openshift.common.is_containerized | bool
```
detiber: I would like to see if you could put some thought into how we could potentially get this pushed back down into the docker/openshift_docker role at some point. As an interim step, I see us completely needing a separate playbook, but for the long run, I think we will need to incorporate the docker upgrade scenario back into the roles to not sacrifice the pure-ansible usability.

playbooks/common/openshift-cluster/initialize_facts.yml
dgoodwin: Ok this sounds really good to me, I will carry this idea through to openshift_common instead.

dgoodwin: Oh right, except for the complication of the openshift_docker_hosted_registry_network which has to be based on the first master. I will leave that item here and we can address with delegate_to perhaps in the future.

dgoodwin:
@smunilla I ended up fixing that bug and applying it to this branch for a couple reasons, firstly this is where I removed the 3.1 minor upgrade, and also where I know for certain that 3.2 minor/major upgrade both use the same playbook. Could you take a quick look at commit 6ec757e4449fbc59f509f980bd9423bdb35412eb and let me know if it looks ok?

playbooks/common/openshift-cluster/upgrades/docker/upgrade.yml
dgoodwin: Yeah so with this PR, technically you could upgrade Docker by setting an explicit docker_version in inventory, it just wouldn't be a "safe" upgrade with node evac. That seems pretty tough right now, though I guess openshift_docker *could* handle logic across docker/node roles and determine if an evac is necessary, however if we wait for 1.12 (can restart daemon without losing containers) or system containers, then the problem goes away entirely I think. So I think this would be very possible soon, and I'm pretty intrigued by your idea to push ALL upgrade into config.yml runs. Let the roles handle it just by interpreting your inventory, so upgrade would become a process of just setting inventory var, likely openshift_release.

Should be able to do this but it feels like we should just wait until 1.12 or sys containers.

dgoodwin: As far as actual implementation goes (for triggering upgrades via config.yml playbook runs), I would be thinking about having the new openshift_version role do comparisons of currently running version to the requested openshift_release (or image tag / pkg version) and check if this looks like the user asked for an upgrade. openshift_version would then set whatever facts we need to signal that an upgrade is underway, other roles would key off those, otherwise just protect installed versions.

i.e. docker role would detect that an upgrade is underway and assume to go to latest docker if this is a docker deployment. (unless docker_upgrade=False)

node role would detect an upgrade requested and handle all evac tasks.

When it comes to one off things like the steps required for image upgrade when crossing the Docker 1.10 boundary, we'll just have to embed that logic into the roles for as long as that is within the realm of reason for a supported path, and kill those tasks when it's not.

roles/openshift_version/tasks/main.yml
```diff
@@ -0,0 +1,88 @@
...
+
+- fail:
+    msg: openshift_version role was unable to set openshift_pkg_version
+  when: openshift_pkg_version is not defined
+
+# We can't map an openshift_release to full rpm version like we can with containers, make sure
+# the rpm version we looked up matches the release requested and error out if not.
+- fail:
+    msg: "Detected openshift version {{ openshift_version }} does not match requested openshift_release {{ openshift_release }}. You may need to adjust your yum repositories or specify an exact openshift_pkg_version."
+  when: not is_containerized | bool and openshift_release is defined and not openshift_version | oo_startswith(openshift_release) | bool
```
abutcher: Really nit picky but we can use `.startswith()` in this expression w/o a filter :sun_with_face:

dgoodwin: Are you positive? I tried that every which way I could think of, ansible 1.9.4 at the time, and I couldn't find an example of it with searching. I will re-try.

abutcher: Yeah, try this -- ofc using 2.1 here

```
---
- hosts: localhost
  gather_facts: no
  vars:
    some_variable: "12345"
  tasks:
  - debug:
      msg: "YAY"
    when: some_variable.startswith('1')
  - debug:
      msg: "OH NO"
    when: some_variable.startswith('2')
```

dgoodwin: "ofc" = ? Is it 2.1 specific? If so I will make a note to remove that filter later as I'm hoping to get this in for next 3.2 build.

abutcher: ofc = of course

abutcher: Works with 1.9.4 and 2.1 with a quick test.

detiber: nice!

sdodson:
aos-ci-test-extended

roles/openshift_version/tasks/main.yml
dgoodwin: So confused, I spent the better part of an hour trying to get that to work and couldn't, but you are correct and it's fine now. Must have botched something else in the line. I've updated and remove oo_startswith.

dgoodwin:
aos-ci-test-extended

dgoodwin:
aos-ci-test-extended

@detiber @abutcher how are you guys feeling about this PR? I've fixed everything that's come out of QE so far. Still no clean test runs due to CI problems but trying again.

dgoodwin:
aos-ci-test-extended

dgoodwin:
aos-ci-test-extended

abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
@dgoodwin I'm good with merging pending some tests. /me kicks jenkins

dgoodwin:
@abutcher ok thanks, I just pushed my last fixes for known issues.

abutcher:
@dgoodwin There are also a few pylint warnings.

This one is fixed by https://github.com/openshift/openshift-ansible/pull/2146
```
E:1621,32: No value for argument 'gather_subset' in function call (no-value-for-parameter)
```

dgoodwin:
On it.

dgoodwin:
@abutcher I *hope* that will take care of them, probably some ones coming in future pylint versions but I tried not to get into other files I didn't hit in the PR.

detiber:
@dgoodwin looks like there is still an outstanding pylint error: `E: 63, 0: Bad option value 'redefined-variable-type' (bad-option-value)`

dgoodwin:
aos-ci-test

dgoodwin:
Trying a fix, caused by fixing pylint errors locally that are for a newer version of pylint than CI uses, leading to unknown keys I was trying to disable.

dgoodwin:
aos-ci-test

dgoodwin:
Fixed a couple small errors reported by QE. Two bugs pending that I cannot reproduce and waiting for more info, I don't think either would be a blocker to merging though.

sdodson:
Remove WIP from title? So far my testing looks good. Working on containerized now.

sdodson:
aos-ci-test-extended

sdodson:
Manual testing checks out.

PR #1948, opened 2016-05-25
[WIP][DO NOT MERGE] Use docker 1.10 for 3.2

sdodson:


roles/openshift_docker_facts/tasks/main.yml
```diff
@@ -65,6 +65,6 @@
 
 - name: Set docker version to be installed
   set_fact:
-    docker_version: "{{ '1.9.1' }}"
+    docker_version: "{{ '1.10.3' }}"
```
detiber: Do we want to mandate a specific docker version for 3.2, or would it make sense to just ensure gte 1.9.1 (at least until we have a ceiling on supported version)?

sdodson: to be determined, this is mostly just to get QE a version that installs with docker-1.10 so we can come to a conclusion on whether or not we're going to ship docker-1.0

sdodson: that said, my preference would be to implement gte, but I've yet to come up with a clean way to do that. I guess we could do bounds checking in ansible, if the to-be-installed version fits within those bounds then just run yum without a version. If to-be is outside of those bounds then set version to the upper bound.

detiber: True, extending that logic into ansible is something we should try to avoid. Which would mean having to deal with some yum shell shenanigans or such...

The complication here seems to be around containerized deployments, though...  Maybe we need to segregate these types of checks there and let the RPM dependencies manage it for us in the case of non-containerized installations?

PR #1949, opened 2016-05-26
Add requirements.txt

jae2:
For those of us who prefer to use ansible in a virtualenv this makes things nice and easy for us as we can just do pip install -r requirements.txt.

Steps to test:

```
virtualenv venv 
source venv/bin/activate
pip install -r requirements.txt
ansible-playbook -i your_inventory playbooks/path/to/playbook
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,17 @@
+Jinja2==2.8
+MarkupSafe==0.23
+PyYAML==3.11
+ansible==1.9.4
+cffi==1.5.0
+cryptography==1.2.1
+ecdsa==0.13
+enum34==1.1.2
+idna==2.0
+ipaddress==1.0.16
+paramiko==1.16.0
+pyOpenSSL==0.15.1
+pyasn1==0.1.9
+pycparser==2.14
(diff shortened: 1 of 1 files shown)


detiber:
@abutcher ptal

requirements.txt
```diff
@@ -0,0 +1,17 @@
...
+enum34==1.1.2
+idna==2.0
+ipaddress==1.0.16
+paramiko==1.16.0
+pyOpenSSL==0.15.1
+pyasn1==0.1.9
+pycparser==2.14
+pycrypto==2.6.1
+six==1.10.0
+wsgiref==0.1.2
```
abutcher: Should we leave versions out of the requirements except for the ansible version? Most of these are dependencies of ansible. Can we limit the requirements to ansible and pyopenssl?

DanyC97:
@jae2 are you still working on this? seems is a trivial change following @abutcher review . Happy to take it if you don't have time to do so

sdodson:
@DanyC97 looks like no response from OP so i've closed this.

PR #1964, opened 2016-05-31
Preserve proxy config if it exists in the sysconfig files but is undefined in our config

sdodson:
Preserves existing PROXY variables in /etc/sysconfig/docker and /etc/sysconfig/*master-* whenever `openshift.common.http_proxy` and `openshift.common.https_proxy` are not defined.

This will gratuitously restart master/docker services if the template applies no changes other than removing proxy settings and then we restore the proxy settings.

### DIFF
--- a/roles/openshift_master/tasks/systemd_units.yml
+++ b/roles/openshift_master/tasks/systemd_units.yml
@@ -38,6 +38,13 @@
   when: create_ha_unit_files | changed
 # end workaround for missing systemd unit files
 
+- name: Preserve Master API Proxy Config options
+  command: grep PROXY /etc/sysconfig/{{ openshift.common.service_type }}-master-api
+  register: master_api_proxy
+  when: openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
+  failed_when: false
+  changed_when: false
+
 - name: Create the master api service env file
   template:
     src: "{{ ha_svc_template_path }}/atomic-openshift-master-api.j2"
@@ -46,6 +53,21 @@
   when: openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
   notify:
   - restart master api
+
+- name: Restore Master API Proxy Config Options
+  when: openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
+      and master_api_proxy.rc == 0 and 'http_proxy' not in openshift.common and 'https_proxy' not in openshift.common
+  lineinfile:
+    dest: /etc/sysconfig/{{ openshift.common.service_type }}-master-api
+    line: "{{ item }}"
+  with_items: "{{ master_api_proxy.stdout_lines | default([]) }}"
+
+- name: Preserve Master Controllers Proxy Config options
+  command: grep PROXY /etc/sysconfig/{{ openshift.common.service_type }}-master-api
+  register: master_controllers_proxy
+  when: openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
+  failed_when: false
+  changed_when: false
 
 - name: Create the master controllers service env file
   template:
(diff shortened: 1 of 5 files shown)


roles/openshift_master/tasks/systemd_units.yml
```diff
@@ -38,6 +38,11 @@
   when: create_ha_unit_files | changed
 # end workaround for missing systemd unit files
 
+- name: Preserve Master API Proxy Config options
+  command: grep PROXY /etc/sysconfig/{{ openshift.common.service_type }}-master-api
+  register: master_api_proxy
+  when: openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
```
detiber: If we are going this route, should we then store these values in openshift_facts?

sdodson: I think it's safest to gather them each time encase something has changed them since we last ran?

sdodson:
@detiber PTAL, i feel like the only other way to address this is to go back to the master configs all being linein implementations rather than templates.

sdodson:
aos-ci-test-extended

roles/openshift_master/tasks/systemd_units.yml
detiber: So, one of the things that I keep coming back to lately is the thought that openshift_facts should really just return the state of the system as configured rather than some supposed "desired state".

My original thought was that we needed to persist these settings on the hosts somewhere in the event that someone ran the installer (or playbooks) without having the proper variables set.  Instead, I think the only time we should be worrying about "desired config" rather than current config is when someone is managing their environment using ansible, in which case the ansible variables will already be set.

In this case, openshift-facts should be returning the data on disk for these values, and we should feed in something like this to the templates (pseudocode ahead):
```
MY_PROXY_VAR={{ openshift_my_proxy_var | default(openshift.common.my_proxy_var | default(None)) }}"
```

sdodson: You mean on disk value for openshift.common.http_proxy and friends or on disk value by parsing /etc/sysconfig/docker and /etc/sysconfig/atomic-openshift-master* ? Those two will actually be slightly different as the docker config doesn't know how to handle CIDR addresses and the master config does.

detiber: Well, isn't that fun...  In short, no, yes and yes. I would like to see us start getting away from persisting values in openshift_facts where it makes sense...  maybe it doesn't make sense in this case, which would mean that parsing and re-adding it each time is the only solution, but it still seems very error prone that way.

sdodson:
aos-ci-test-extended

roles/openshift_master/tasks/systemd_units.yml
sdodson: I think this particular implementation keeps us from having to parse them. If it's there and we don't have these variables defined, either on disk or from inventory, then we just restore them exactly as they are after making whatever other changes we need to make.

detiber: I'm just not a fan of doing this within the config playbooks...  If it is something that we do only in the upgrade playbooks, then it doesn't concern me...  Doing this in the config playbooks breaks idempotency.

sdodson:
aos-ci-test-extended

roles/openshift_master/tasks/systemd_units.yml
sdodson: ok, added conditions to only preserve these settings on upgrades

roles/openshift_master/tasks/systemd_units.yml
```diff
@@ -38,32 +38,81 @@
...
+      and upgrading is defined and upgrading | bool
```
detiber: I can't say I'm a fan of the upgrade variable here...  I don't want to hold this change up too much, since it is affecting users, but I'd like to make a point of cleaning this up at a later point...  maybe that can be rolled into @dgoodwin's upgrade changes?

sdodson: @dgoodwin are you ok with my use of upgrading variable here? That's not going away in your in-progress work is it?

dgoodwin: Unfortunately it is, @detiber explained the reasons why we don't want to
require that state in the roles. It went into master temporarily due to
time constraints. Can we find another way to tell the role what we want
here?

brenton:
You may have mentioned this at scrum and I forgot, is this still planned for 3.2.1?

sdodson:
Sorry I forgot to bring this up, we should try to get this in 3.2.1. I'll work on it tomorrow.

sdodson:
aos-ci-test

brenton:
LGTM

PR #1965, opened 2016-05-31
Ensure registry url evaluated when creating router.

abutcher:
https://bugzilla.redhat.com/show_bug.cgi?id=1340410

### DIFF
--- a/playbooks/common/openshift-cluster/openshift_hosted.yml
+++ b/playbooks/common/openshift-cluster/openshift_hosted.yml
@@ -1,5 +1,7 @@
 - name: Create Hosted Resources
   hosts: oo_first_master
+  pre_tasks:
+  - set_fact:
+      openshift_hosted_router_registryurl: "{{ hostvars[groups.oo_first_master.0].openshift.master.registry_url }}"
   roles:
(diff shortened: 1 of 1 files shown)


abutcher:
@detiber PTAL - this is impacting 1.9.4 in my testing so we should get this included quickly if possible. /cc @sdodson

abutcher:
aos-ci-test

sdodson:
LGTM

abutcher:
aos-ci-test

playbooks/common/openshift-cluster/openshift_hosted.yml
```diff
@@ -1,5 +1,7 @@
 - name: Create Hosted Resources
   hosts: oo_first_master
+  pre_tasks:
+  - set_fact:
+      openshift_hosted_router_registryurl: "{{ hostvars[groups.oo_first_master.0].openshift.master.registry_url }}"
```
detiber: I suspect this is similar to the issue when things are being set in vars...  The templates are being evaluated when they are used rather than when they are defined (as in the case of set_fact).

PR #1968, opened 2016-06-01
[v2] Move os_firewall_allow from defaults to role dependencies.

abutcher:
* Move `os_firewall_allow` from defaults to role dependencies for all roles. These defaults were not being acted on.
* Fix issue with openshift-loadbalancer refactor, ensuring that `haproxy_frontend_port` is fed from the playbook to drive firewall configuration.

The loadbalancer refactor changes were merged post cut so, if we're doing backports for critical fixes, these changes shouldn't be included in the update since the update won't be impacted by them.

https://bugzilla.redhat.com/show_bug.cgi?id=1339431

@detiber PTAL

### DIFF
--- a/roles/openshift_node/meta/main.yml
+++ b/roles/openshift_node/meta/main.yml
@@ -19,4 +19,17 @@ dependencies:
 - role: openshift_node_dnsmasq
   when: openshift.common.use_dnsmasq
 - role: os_firewall
-
+  os_firewall_allow:
+  - service: Kubernetes kubelet
+    port: 10250/tcp
+  - service: http
+    port: 80/tcp
+  - service: https
+    port: 443/tcp
+  - service: Openshift kubelet ReadOnlyPort
+    port: 10255/tcp
+  - service: Openshift kubelet ReadOnlyPort udp
+    port: 10255/udp
+  - service: OpenShift OVS sdn
+    port: 4789/udp
+    when: openshift.node.use_openshift_sdn | bool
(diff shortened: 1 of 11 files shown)


abutcher:
@sdodson also PTAL

abutcher:
aos-ci-test

roles/openshift_node/meta/main.yml
```diff
@@ -19,4 +19,17 @@ dependencies:
...
+    port: 80/tcp
+  - service: https
+    port: 443/tcp
+  - service: Openshift kubelet ReadOnlyPort
+    port: 10255/tcp
+  - service: Openshift kubelet ReadOnlyPort udp
+    port: 10255/udp
+  - service: OpenShift OVS sdn
+    port: 4789/udp
+    when: openshift.node.use_openshift_sdn | bool
```
detiber: hmm, does this when actually work, or just add a 'when' key to the dict?

detiber:
outside of the nitpicky comment about an existing when clause, lgtm.

roles/openshift_node/meta/main.yml
abutcher: It applies -- not sure about the negative case but I can confirm.

abutcher: I'd assume it just adds this no matter what with a 'when' key in the dict.

detiber: That's what I was thinking as well...  Unless I added some special hackery to os_firewall, which isn't out of the question either.

sdodson: Probably doesn't work as expected, though it's not a change from current implementation.

sdodson:
LGTM

PR #1969, opened 2016-06-01
Libvirt provider fixes

detiber:


detiber:
This will probably also impact other bin/cluster uses, but I haven't verified, yet.

detiber:
@abutcher This seems to fix libvirt for me, I have a suspicion that the other providers will need a similar fix.

bachp:
I can confirm that the change also fixes aws. However I'm not sure this is really the correct solution.

moortimis:
This fixed my AWS deployment. Thanks

detiber:
@bachp I agree on this not necessarily being the correct solution, however since it resolves the issue and the affected playbooks are community-supported only, I'm going to merge.

I'm happy to review any PRs that might have a more comprehensive fix, though.

PR #1971, opened 2016-06-02
Fix error with stopping services that may not exist.

dgoodwin:
Causes a hard failure due to missing etcd/atomic-openshift-master
services. Instead, check that the services exist before ensuring they're
stopped/disabled.

### DIFF
--- a/roles/etcd/tasks/main.yml
+++ b/roles/etcd/tasks/main.yml
@@ -28,17 +28,17 @@
     state: directory
     mode: 0700
 
-- name: Disable system etcd when containerized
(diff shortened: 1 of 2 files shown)


dgoodwin:
@abutcher @detiber please take a look. HA installs are busted seemingly 3 different ways (so far) right now, this is a fix for 2 of them.

abutcher:
aos-ci-test

dgoodwin:
3rd breakage alluded to above looked like it was fixed yesterday (firewall issue)

abutcher:
lgtm

PR #1973, opened 2016-06-02
Add flag to manage htpasswd, or not.

ibotty:
Setting `openshift_master_manage_htpasswd` falsy will disable managing
the htpasswd file. It won't get overwritten/generated.

### DIFF
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -93,7 +93,7 @@
     src: htpasswd.j2
     mode: 0600
     backup: yes
-  when: item.kind == 'HTPasswdPasswordIdentityProvider'
+  when: item.kind == 'HTPasswdPasswordIdentityProvider' and openshift.master.manage_htpasswd | bool
   with_items: "{{ openshift.master.identity_providers }}"
 
 - name: Create the ldap ca file if needed
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -91,9 +91,17 @@
   template:
     dest: "{{ item.filename }}"
     src: htpasswd.j2
-    mode: 0600
     backup: yes
   when: item.kind == 'HTPasswdPasswordIdentityProvider' and openshift.master.manage_htpasswd | bool
+  with_items: "{{ openshift.master.identity_providers }}"
+
+- name: Ensure htpasswd file exists
+  copy:
+    dest: "{{ item.filename }}"
+    force: no
+    content: ""
+    mode: 0600
+  when: item.kind == 'HTPasswdPasswordIdentityProvider'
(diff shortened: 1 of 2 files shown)


sdodson:
aos-ci-test

sdodson:
bot, test pull request

roles/openshift_master/tasks/main.yml
```diff
@@ -93,7 +93,7 @@
     src: htpasswd.j2
     mode: 0600
     backup: yes
-  when: item.kind == 'HTPasswdPasswordIdentityProvider'
+  when: item.kind == 'HTPasswdPasswordIdentityProvider' and openshift.master.manage_htpasswd | bool
```
detiber: I think we also need to ensure that the file exists in the case that manage_htpasswd is false, otherwise the master service will fail to startup if the file isn't present, something like this should be sufficient:
```
- name: Ensure htpasswd file exists
  copy:
    dest: "{{ item.filename }}"
    content: ""
    mode: 0600
    force: no
  when: item.kind == 'HTPasswdPasswordIdentityProvider' and openshift.master.manage_htpasswd | bool
  with_items: "{{ openshift.master.identity_providers }}"
```

ibotty: I made a commit that addresses the concern. BTW: force defaults to "no".

detiber:
ok to test

detiber:
aos-ci-test

roles/openshift_master/tasks/main.yml
```diff
@@ -91,8 +91,15 @@
...
-    mode: 0600
     backup: yes
+  when: item.kind == 'HTPasswdPasswordIdentityProvider' and openshift.master.manage_htpasswd | bool
+  with_items: "{{ openshift.master.identity_providers }}"
+
+- name: Ensure htpasswd file exists
+  copy:
+    dest: "{{ item.filename }}"
+    state: touch
+    mode: 0600
```
detiber: Does this task show as changed every time? From the docs it appears that would be the case. I'd prefer that we didn't update the file time stamps or report changed on every run.

ibotty: I concur. The amended commit uses the copy module with force=no now.

detiber:
aos-ci-test

detiber:
ok to test

detiber:
@sdodson lgtm, do we want to backport this to the 3.2 branch as well?

detiber:
3.1 containerized test failures are unrelated, and I don't see anything that would behave differently here vs containerized.

sdodson:
> @sdodson lgtm, do we want to backport this to the 3.2 branch as well?

Sure, how should we track those items?

detiber:
@sdodson sounds good to me, will need one for backport/3.1 as well. I'll create them now.

PR #1974, opened 2016-06-02
WIP: Docker 1.10 Upgrade Playbook / Installer Changes

dgoodwin:
Initial draft up of just the playbook.

Changes to config playbooks coming next.


playbooks/byo/openshift-cluster/upgrades/docker/files/nuke_images.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Stop all running containers
+docker stop $(docker ps -q)
+
+# Delete all containers
+docker rm $(docker ps -a -q)
+
+# Deleate all images (forcefully)
```
detiber: s/Deleate/Delete/

playbooks/byo/openshift-cluster/upgrades/docker/upgrade.yml
```diff
@@ -0,0 +1,170 @@
+# Playbook to upgrade Docker to the max allowable version for an OpenShift cluster.
+#
+# Currently only supports upgrading 1.9.x to >= 1.10.x.
+---
+- hosts: localhost
+  connection: local
+  become: no
+  gather_facts: no
+  tasks:
+  - name: Verify Ansible version is greater than or equal to 1.9.4
```
detiber: Do we need to test this separately here, or can we rely on the test in openshift_facts?

playbooks/byo/openshift-cluster/upgrades/docker/upgrade.yml
```diff
@@ -0,0 +1,170 @@
...
+    when: openshift_docker_blocked_registries is not defined
+  - set_fact:
+      openshift_docker_options: "{{ lookup('oo_option', 'docker_options') }}"
+    when: openshift_docker_options is not defined
+  - set_fact:
+      openshift_docker_log_driver: "{{ lookup('oo_option', 'docker_log_driver') }}"
+    when: openshift_docker_log_driver is not defined
+  - set_fact:
+      openshift_docker_log_options: "{{ lookup('oo_option', 'docker_log_options') }}"
+    when: openshift_docker_log_options is not defined
```
detiber: We shouldn't need to worry about oo_option lookups here. They are an artifact of bin/cluster and I would expect that to be handled in a different way by bin/cluster dependent playbooks.

playbooks/byo/openshift-cluster/upgrades/docker/upgrade.yml
```diff
@@ -0,0 +1,170 @@
...
+      docker_upgrade: true
+    when: not openshift.common.is_atomic | bool and pkg_check.rc == 0 and g_docker_version.curr_version | version_compare('1.10','<')
+
+# TODO: what if running 3.1 and 1.8?
+#
+  - fail:
+      msg: This playbook requires access to Docker 1.10 or later
+    when: g_docker_version.avail_version | default(g_docker_version.curr_version, true) | version_compare('1.10','<')
+
+# TODO: By definition this will cover all masters too, what about containerized etcd hosts that may not be included but are running docker?
```
detiber: I'm not sure we really care about containerized etcd hosts...  We could leave that to system patching to handle (with the exception of etcd hosts that are masters obviously).

detiber: I'm also a bit weary about relying on oo_nodes_to_config to catch containerized masters...  We *could* have someone that installed masters without nodes.  That said, I'm not sure we actually care about what version of docker those masters are.

dgoodwin: It seems like if we run a playbook to upgrade docker in our environment, it should upgrade all docker hosts. I'm happy to skip separate etcd hosts if you're sure though, but it's also probably possible to hit them without too much trouble. Think we should just go for it?

For masters without nodes, I can probably handle that with more conditionals on the evac stuff to check if inventory hostname is in the nodes group. This won't get tested obviously.

Doing that would help with separate etcd hosts too, so why don't we go for that and kill both birds.

detiber: Sounds good.

playbooks/byo/openshift-cluster/upgrades/docker/upgrade.yml
```diff
@@ -0,0 +1,170 @@
...
+    register: nuke_images_result
+    when: docker_upgrade is defined and docker_upgrade | bool
+
+# TODO: Re-enable once process is more locked down.
+  #- name: Upgrade Docker
+    #command: "{{ ansible_pkg_mgr}} update -y docker"
+    #register: docker_upgrade_result
+    #when: docker_upgrade is defined and docker_upgrade | bool
+
+#   Don't think this is necessary, the update seems to do it...
```
detiber: Hmm, really? Is there some craziness happening in the RPM post install that is triggering the restart?

playbooks/byo/openshift-cluster/upgrades/docker/upgrade.yml
```diff
@@ -0,0 +1,170 @@
...
+    service: name={{ item }} state=started
+    with_items:
+      - etcd
+      - openvswitch
+      - "{{ openshift.common.service_type }}-master"
+      - "{{ openshift.common.service_type }}-master-api"
+      - "{{ openshift.common.service_type }}-master-controllers"
+      - "{{ openshift.common.service_type }}-node"
+    failed_when: false
+    when: docker_upgrade is defined and docker_upgrade | bool and openshift.common.is_containerized | bool
```
detiber: I don't think this should be necessary anymore, but better safe then sorry.

dgoodwin:
More changes pushed, the 3.2 upgrade integration hasn't been run yet but I'm working on that now.

playbooks/byo/openshift-cluster/upgrades/docker/files/nuke_images.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Stop all running containers
+docker stop $(docker ps -q)
+
+# Delete all containers
+docker rm $(docker ps -a -q)
```
rassie: This will leave container volumes behind, the correct line should be `docker rm -v ...`

dgoodwin: Thanks, I will have this in the next push.

dgoodwin:
More updates and rebased off master.

playbooks/byo/openshift-cluster/upgrades/docker/upgrade.yml
```diff
@@ -0,0 +1,137 @@
...
+- hosts: localhost
+  connection: local
+  become: no
+  gather_facts: no
+  tasks:
+  - include_vars: ../../../../../byo/openshift-cluster/cluster_hosts.yml
+  - add_host:
+      name: "{{ item }}"
+      groups: l_oo_all_hosts
+    with_items: g_all_hosts | default([])
```
detiber: One thing I just noticed yesterday (and haven't taken the time to see if it occurs with ansible-1.9.x, but these add_hosts tasks show as 'changed', we should probably add `changed_when: false` to these to avoid that.

playbooks/byo/openshift-cluster/upgrades/docker/upgrade.yml
```diff
@@ -0,0 +1,137 @@
...
+
+# Currently only supports upgrading 1.9.x to >= 1.10.x.
+- hosts: l_oo_all_hosts
+  gather_facts: no
+  tasks:
+  - include_vars: ../../../../byo/openshift-cluster/cluster_hosts.yml
```
detiber: could this not be `../../cluster_hosts.yml`?

detiber:
lgtm outside of the nitpicky comments I left, nothing I would consider blockers.

dgoodwin:
Fixed several issues with paths, I had to get a little crazy to avoid issues on Docker 1.9.4, but realized they were caused by including files that upgrade had already run, so I separated the bits we actually want into something we can include more reliably.

playbooks/byo/openshift-cluster/upgrades/docker/upgrade.yml
dgoodwin: I will add that here.

playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
```diff
@@ -0,0 +1,111 @@
...
+  - name: Upgrade Docker
+    command: "{{ ansible_pkg_mgr}} update -y docker"
+    register: docker_upgrade_result
+    when: docker_upgrade is defined and docker_upgrade | bool
+
+#   Don't think this is necessary, the update seems to do it...
+  #   - name: Restart Docker
+  #    command: systemctl restart docker
+  #    when: not openshift.common.is_atomic | bool and docker_upgrade_result | changed
+  #
```
detiber: Man, we should probably talk to upstream about this...  The thought of a package upgrade restarting a service seems like a horrible idea to me.

detiber: @brenton ^

detiber:
aos-ci-test-extended

detiber:
lgtm, haven't been tracking if the extended tests were working, but it doesn't hurt to try.

playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
brenton: It was probably never needed.  As best I can tell by looking at the spec files this docker has restart on upgrade since 1.8.2.

dgoodwin:
@brenton @detiber ok this is squashed and I think as good as we can get for this week. It's a rush job but it has passed an upgrade on 3.2 HA env, and a lot of local testing on single master+node.

dgoodwin:
Doesn't look like a relevant error to PR. However these may fail if dockers
1.10 isn't available on the test systems.

dgoodwin:
aos-ci-test-extended

dgoodwin:
Unsurprising failures as a result of making master 1.10+ only and requiring separate upgrade to run the config playbooks.

TASK: [docker | Fail if Docker upgrade is required] *************************** 
failed: [10.8.168.252] => {"failed": true}
msg: Docker 1.9.1 must be upgraded to Docker 1.10 or greater

Are we headed for trouble here, do we know with certainty that Docker 1.10 will be available when this PR lands in a build? 

What can we do about the CI environment? If master is going to now only support fresh 3.2, and upgrading pre-existing 3.1, I think we need to drop almost all of our existing CI jobs, 3.0, 3.1, and 3.1 -> 3.2 upgrade now can't really work. :(

CC @brenton

brenton:
@dgoodwin, I definitely think we can drop the 3.0 job at this point.  Now the question is what to do with the 3.1 jobs.  Since we aren't going to ship these changes to 3.1 we don't have to gate on those tests passing for now.  Essentially we're going to have to make sure the 3.1 jobs only fire for the < docker 1.10 branch.  The upgrade job will likely need to be changed to be:

* install with the 3.1 GA playbooks
* build the 3.2 playbooks and upgrade with that

@arilivigni, what would be our easiest option?  The short of this is that 'master' can no longer support 3.0 and 3.1 and we're going to maintain a branch for versions that support less than docker 1.10.

arilivigni:
@brenton We should probably discuss it outside of this PR since this isn't something currently on the continuous-infra backlog.

brenton:
Does this PR address https://bugzilla.redhat.com/show_bug.cgi?id=1338726 ?

dgoodwin:
@brenton it does not and I'm quite unsure how I was able to successfully install without hitting that. I will investigate asap.

detiber:
So, I think the new branches should rectify issues with testing here..  As we discussed earlier, we are missing coverage for 3.2 and for 'master', which I think is another issue.

I agree with brenton that we should drop 3.0.

We'll need to update the other branches as follows:
- 3.1 should use docker 1.8, associated 3.1 branch of openshift-ansible, 3.1 prod repos
- 3.2 should use docker 1.10, associated 3.2 branch of openshift-ansible, 3.2 prod repos (the docker 1.10 change would need to be coordinated with this PR)
- puddle should use docker 1.10, master branch of openshift-ansible, latest devel puddle.

At some point, we'll probably have to think about running 3.1/3.2 against associated puddles as well.

dgoodwin:
@brenton added another commit to solve that bug, I won't squash that one as it is very separate.

PR #1976, opened 2016-06-02
Configure default docker logging options.

### LINKED ISSUE #1887: Parameterize some defaults for Logging to prevent cluster collaps
Lets see if we can prevent https://github.com/openshift/origin/issues/8868 at the installer level, at least temporarily if not permanantly.

TL;DR :We don't want `docker logs` from noisy containers to fill up the disks and kill your cluster!

abutcher:
* Adds `"--log-driver=json-file --log-opt max-size=50m"` to docker options by default ([docs](https://docs.docker.com/engine/admin/logging/overview/#json-file-options)).
* Removes deprecated `openshift_docker_log_driver` and `openshift_docker_log_options` from example inventories to encourage use of `openshift_docker_options`. Also show the default value in example inventories.

The [advanced installation docs](https://docs.openshift.com/enterprise/3.2/install_config/install/advanced_install.html#configuring-host-variables) do not mention `openshift_docker_options` but do mention `openshift_docker_log_options` so an issue should be created to swap them.

Resolves #1887
https://bugzilla.redhat.com/show_bug.cgi?id=1335939

### DIFF
--- a/inventory/byo/hosts.aep.example
+++ b/inventory/byo/hosts.aep.example
@@ -72,10 +72,8 @@ deployment_type=atomic-enterprise
 # Disable pushing to dockerhub
 #openshift_docker_disable_push_dockerhub=True
 # Items added, as is, to end of /etc/sysconfig/docker OPTIONS
+# Default value: "--log-driver=json-file --log-opt max-size=50m"
 #openshift_docker_options="-l warn --ipv6=false"
-# Deprecated methods to set --log-driver and --log-opts flags, use openshift_docker_options instead
-#openshift_docker_log_driver=json
-#openshift_docker_log_options=["tag=mailer"]
 
 # Alternate image format string. If you're not modifying the format string and
 # only need to inject your own registry you may want to consider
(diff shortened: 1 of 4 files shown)


abutcher:
@richm Would you recommend these defaults until we can use the journald log driver by default?

detiber:
@timothysc ptal as well.

jayunit100:
I think The models in openshift_facts.py appears to solve the issue for us, thanks !, otherwise not qualified to review this comprehensively .

detiber:
@jayunit100 that is the confirmation we needed, thanks!

detiber:
aos-ci-test

richm:
@timothysc @detiber would using the `--log-opt max-size=50m` be an acceptable workaround to the problem of docker logs filling up the container?  As opposed to using the `--log-driver=journald` option?

detiber:
@richm I think the goal was to use this as a stopgap until the journald log driver can be used for 3.3+

detiber:
@abutcher e2e looks like a flake, going to retrigger.

detiber:
aos-ci-test

richm:
@detiber ack - LGTM

timothysc:
> @richm I think the goal was to use this as a stopgap until the journald log driver can be used for 3.3+

+1 lgtm

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
Merging since the one failed test was a flake and passed during the previous run where the e2e test flaked.

PR #1979, opened 2016-06-03
Hardcoded values in "launch_instances" - isue # 1970

dumdan:
I suppose this would just correct an omission in the installer.


detiber:
ok to test

detiber:
Skipping other tests as they do not test this playbook.

detiber:
@dumdan many thanks!

PR #1984, opened 2016-06-03
[WIP] Only run node specific bits on nodes

### LINKED ISSUE #1966: Uninstall should restart docker on hosts where docker is installed
Instead of all hosts.

sdodson:
Fixes #1966

### DIFF
--- a/playbooks/adhoc/uninstall.yml
+++ b/playbooks/adhoc/uninstall.yml
@@ -25,11 +25,6 @@
     - set_fact:
         is_containerized: "{{ is_atomic or containerized | default(false) | bool }}"
 
-    - name: Remove br0 interface
-      shell: ovs-vsctl del-br br0
-      changed_when: False
-      failed_when: False
-
     - name: Stop services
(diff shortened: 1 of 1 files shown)


detiber:
lgtm

sdodson:
I think this probably fails to clean up docker modifications like additional registries and proxy on a standalone containerized etcd host as is but I can do a  followup to get that.

detiber:
@sdodson @brenton: I think at some point we need to discuss what the purpose of the uninstallation playbook is...  If it's meant to be a helper script that allows for easy testing, or if we want a fully supported un-installation tool that leaves no traces behind...

I'd be happy to relegate it to community supported status. There are a ton of permutations of things that we are doing currently and will be doing over time. I don't think it's quite reasonable to think that this tool will ever reach 100% full, leave no trace behind functionality.

brenton:
There are two goals:

* facilitate development
* provide a way for atomic-openshift-installer to wipe a system in hopes of the next run running flawlessly.

PR #1986, opened 2016-06-06
Load openshift ca.crt into ca-trust

andrewklau:
This avoids me having to load the openshift cert manually on every node host after securing the docker registry. I'm not sure how relevant this may be to other configurations..

### DIFF
--- a/roles/openshift_node/handlers/main.yml
+++ b/roles/openshift_node/handlers/main.yml
@@ -7,3 +7,8 @@
   service: name={{ openshift.common.service_type }}-node state=restarted
   when: not (node_service_status_changed | default(false) | bool)
 
+- name: restart docker
+  service:
+    name: docker
+    state: restarted
+  when: default(false)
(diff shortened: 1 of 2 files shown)


detiber:
@andrewklau I like what you are trying to accomplish with this change, unfortunately I don't think we want to modify docker config or trigger a docker restart directly from the node role.

@abutcher could you work with Andrew on getting this aligned with your delegate_to changes, so that it can be handled through dependencies? I would suspect that we may want to also make sure that there can be additional CAs specified in addition to the OpenShift CA

abutcher:
I think we want the docker role to manage updating ca-trust here so the CA(s) would become a parameter for the docker role via openshift_docker. My certificate changes in https://github.com/openshift/openshift-ansible/pull/1990 make openshift_node_certificates a dependency of openshift_node, so openshift_docker needs to be depended on after openshift_node_certificates in order for the CA to exist.

andrewklau:
Ok great! I'll close this in favour of your #1990

PR #1995, opened 2016-06-07
roles/openshift_docker_facts: add missing --installed option to repoquery

ingvagabund:
Running on RHEL7.2 cluster the ``repoquery --qf '%{version}' origin`` returns empty line. With ``--installed`` option, it returns version of installed origin rpm.

Not sure if this is intended or not. Thought, this fixes the docker installation for me.

### DIFF
--- a/roles/openshift_docker_facts/tasks/main.yml
+++ b/roles/openshift_docker_facts/tasks/main.yml
@@ -43,7 +43,7 @@
 # See: https://bugzilla.redhat.com/show_bug.cgi?id=1304038
 - name: Gather common package version
   command: >
-    {{ repoquery_cmd }} --qf '%{version}' "{{ openshift.common.service_type}}"
+    {{ repoquery_cmd }} --installed --qf '%{version}' "{{ openshift.common.service_type}}"
   register: common_version
   failed_when: false
   changed_when: false


detiber:
@dgoodwin ptal

dgoodwin:
I haven't done a lot of origin install testing lately but could you clarify the situation a little bit @ingvagabund , is this a fresh installation? Or are you running the playbooks for ongoing maintenance.

This would seem to indicate you don't have the origin rpm available in any configured repos, but you do have it installed, so I'm a little confused about how it got there and where it came from. I don't think we want to flip it to --installed as a fresh install wouldn't have anything installed yet afaik.

I will do some poking around first chance I get, please provide any additional info that might be relevant. Thanks!

ingvagabund:
@dgoodwin I have built the origin on Koji, installed relevant rpms on nodes and then run the ansible playbook. So you are saying the use case for installing nonrepository rpms (i.e. scratch builds) are not supported?

dgoodwin:
I can't say for sure regarding supported or not, we've actually been talking about this lately as we're swamped with the number of install paths we're attempting to support.

Your explanation does clarify how this happens though, the installer does assume you have yum repositories with the rpms required available. The best workaround would likely be to publish a yum repo with these custom rpms you built and then distribute that repo file to all the systems. This would be a lot closer to the normal supported path, no guarantees something else wouldn't break but it's worth a shot.

dgoodwin:
Out of curiosity, why not install origin via the installer and skip all the custom building of rpms?

ingvagabund:
@dgoodwin what you propose is easier to automatize. I need custom building to test the latest origin code to reproduce a bug (or confirm the issue is gone).

dgoodwin:
I understand, ok lets give that a shot. 

I'm doing some work that will hopefully allow you to install from the latest origin 1.3 alpha tags quite easily via a containerized install (no rpms), this should be available in the coming weeks.

ingvagabund:
@dgoodwin thanks for that. Meantime, I will take the "custom repo" way. Close this PR anytime you find it "closable".

PR #1978, opened 2016-06-03
[Do not merge] Keynote demo tweaks

detiber:
Tracking PR for changes made to support or improve deployment for the keynote demo.

Changes so far:
- hack workaround for disabling openshfit_no_proxy_internal_hostnames 
- Addition of openshift_projects role, allowing the ability to set default node selectors on a project level
- Remove injecting of credentials file for router
- Tweaks to router deployment
- Move registry config into openshift_hosted role, remove injecting of credentials file
- Introduction of openshift_hosted_infra_selector to set the selector for openshift_hosted_*_selector with one variable

PR #2001, opened 2016-06-08
set default metrics_url if not set

jfcoz:


abutcher:
I think #1978 addresses the issue with metrics_public_url. I'll be separating those changes into separate pull requests as I can but could you take a look and comment there if you notice any problems?

detiber:
@sdodson I think this is not going to be needed with your upcoming changes.

detiber:
Closing due to age and metrics refactoring.

PR #2005, opened 2016-06-08
Bug 1338726 - never abort install if the latest version of docker is already installed

brenton:
This includes https://github.com/openshift/openshift-ansible/pull/1974


roles/docker/tasks/main.yml
```diff
@@ -1,41 +1,42 @@
...
-
-- name: Downgrade docker if necessary
-  command: "{{ ansible_pkg_mgr }} swap -y docker-* docker-*{{ docker_version }}"
-  register: docker_downgrade_result
-  when: not docker_version_result | skipped and docker_version_result.stdout | default('0.0', True) | version_compare(docker_version, 'gt')
+    {{ repoquery_cmd }} --qf '%{version}' "docker"
+  register: avail_docker_version
+  failed_when: false
+  changed_when: false
+  when: not curr_docker_version.stdout | default('0.0', True) | version_compare('1.10', '>=') and not openshift.common.is_atomic | bool
```
brenton: @abutcher, @detiber you might want to see my commit to see what I changed on top of Devan's.  The basic idea is that we don't need to check the latest avail version if we're already running it.  Then the check to 'Verify Docker >= 1.10 is available' will be skipped as well which was what the bug was about.

cc @dgoodwin

brenton:
@abutcher, @detiber, besides this Devan's PR passed QE's testing.

abutcher:
lgtm

playbooks/byo/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
```diff
@@ -4,10 +4,6 @@
   become: no
   gather_facts: no
   tasks:
-  - name: Verify Ansible version is greater than or equal to 1.9.4 and less than 2.0
```
detiber: Do we still want to check for docker > 1.9.4?

roles/openshift_master/tasks/main.yml
```diff
@@ -172,14 +172,9 @@
...
 
-- name: Check for non-HA master service presence
-  command: systemctl show {{ openshift.common.service_type }}-master.service
-  register: master_svc_show
-  changed_when: false
-
-- name: Stop and disable non-HA master when running HA
+- name: Stop and disable non HA master when running HA
   service: name={{ openshift.common.service_type }}-master enabled=no state=stopped
-  when: openshift_master_ha | bool and 'LoadState=not-found' not in master_svc_show.stdout
```
abutcher: Was this accidentally removed in a rebase or intentionally removed?

roles/docker/tasks/main.yml
```diff
@@ -1,41 +1,42 @@
...
+  when: not avail_docker_version | skipped and avail_docker_version.stdout | default('0.0', True) | version_compare('1.10', '<')
 
-# If docker were enabled and started before we downgraded it may have entered a
-# failed state. Check for that and clear it if necessary.
-- name: Check that docker hasn't entered failed state
-  command: systemctl show docker
-  register: docker_state
-  changed_when: False
+- stat: path=/etc/sysconfig/docker-storage
+  register: docker_storage_check
```
detiber: Is this used anywhere anymore?

roles/etcd/tasks/main.yml
```diff
@@ -28,18 +28,18 @@
     state: directory
     mode: 0700
 
-- name: Check for etcd service presence
-  command: systemctl show etcd.service
-  register: etcd_show
-  changed_when: false
-
 - name: Disable system etcd when containerized
-  when: etcd_is_containerized | bool and 'LoadState=not-found' not in etcd_show.stdout
```
detiber: echoing @abutcher's comments about the master role here as well.

roles/openshift_master/tasks/main.yml
brenton: we might have to wait for @dgoodwin to chime in.

playbooks/byo/openshift-cluster/upgrades/v3_1_to_v3_2/upgrade.yml
dgoodwin: I think I removed this as something else was replicating this check, IIRC openshift_facts but I can check tomorrow.

roles/openshift_master/tasks/main.yml
abutcher: I remembered a PR adding these recently and it was @dgoodwin's so this is probably part of the plan

dgoodwin: It was intentional but possibly mistaken, I think at the time I had reasoned that the service would be guaranteed to be there. Can dig more tomorrow but let me know if anyone knows otherwise. Harmless to restore it in any case.

dgoodwin: Yikes, not sure how this happened but this is explicitly undo'ing the fix that went in with https://github.com/openshift/openshift-ansible/pull/1971/files. I botched something with submitting it separately and then syncing the branches. Removing this from this PR.

brenton: If you simply want to cherry pick my commit in to your original PR it would be fine with me.

brenton:
@dgoodwin, does your latest PR address all of @abutcher and @detiber's comments?

dgoodwin:
@brenton everything except the question about whether or not docker storage check is still used, I'm not sure on the answer to that but it's a task that's just being shuffled around (not added new) so I think we can defer.

roles/docker/tasks/main.yml
sdodson: Looks like no.

abutcher:
LGTM I'll add removing the upgrade restriction here to https://github.com/openshift/openshift-ansible/pull/1894

sdodson:
aos-ci-test-extended

sdodson:
Tests fail because docker-1.10 is unavailable. I just ran through an rpm install on my environment and everything looks fine.

PR #2007, opened 2016-06-09
Add lower case proxy variables

bachp:
Some applications expect the *_PROXY variables to be lower case.
To support this too inject them in addition to the upper case ones.

Signed-off-by: Pascal Bach <pascal.bach@siemens.com>
Reviewed-by: Fabio Huser <fabio.huser@siemens.com>

### DIFF
--- a/roles/openshift_builddefaults/vars/main.yml
+++ b/roles/openshift_builddefaults/vars/main.yml
@@ -13,3 +13,9 @@ builddefaults_yaml:
         value: "{{ openshift.builddefaults.https_proxy | default('', true) }}"
       - name: NO_PROXY
         value: "{{ openshift.builddefaults.no_proxy | default('', true) | join(',') }}"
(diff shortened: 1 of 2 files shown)


sdodson:
@csrwng For some reason I thought downstream components would handle automatically populating the lowercase variants of these builddefaults env-vars? Does this change seem necessary?

csrwng:
@sdodson the builders don't do the conversion for you (that I know of), so I think this change is good. 
@bparees can you confirm?

bparees:
yeah we don't do anything, those are just user provided envs so they need to provide whatever the thing inside the image needs.

sdodson:
Cool, thanks!

sdodson:
aos-ci-test

sdodson:
bot, test pull request

abutcher:
ok to test

PR #2010, opened 2016-06-09
Bug 1338726 - never abort install if the latest version of docker is already installed

sdodson:


sdodson:
I've tested this on my vms using both RPMs and Containerized installs.

PR #2020, opened 2016-06-10
Fix docker 1.10 upgrade on embedded etcd masters.

dgoodwin:
The tasks were attempting to stop/start etcd on containerized hosts, but the actual service is etcd_container, and yet the etcd service can also still exist. When we stopped services everything was fine, but attempting to start any services that ar e present would kick off the 'etcd' service that is not expected to be running. It would claim a port, and then when the master came up it's embedded etcd was unable to run taking the master out.

Using the correct etcd_container service should cover this, as it does not exist on an embedded master host.

In future however we might have to be more careful with start/stopping services here, and only attempt to do so based on facts and a lot of logic.

### DIFF
--- a/playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
+++ b/playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
@@ -59,7 +59,7 @@
       - "{{ openshift.common.service_type }}-master-api"
       - "{{ openshift.common.service_type }}-master-controllers"
       - "{{ openshift.common.service_type }}-node"
-      - etcd
+      - etcd_container
       - openvswitch
     failed_when: false
     when: docker_upgrade is defined and docker_upgrade | bool and openshift.common.is_containerized | bool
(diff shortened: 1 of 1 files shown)


sdodson:
LGTM

sdodson:
aos-ci-test-extended

sdodson:
aos-ci-test-extended

sdodson:
aos-ci-test-extended

sdodson:
@dgoodwin was there a BZ associated with this?

dgoodwin:
There is: https://bugzilla.redhat.com/show_bug.cgi?id=1344731

@dgoodwin <https://github.com/dgoodwin> was there a BZ associated with this?

—
<https://github.com/openshift/openshift-ansible/pull/2020#issuecomment-225708622>,
or mute the thread
<https://github.com/notifications/unsubscribe/AADIQXw6lDRNkZ01H_3blBOnJztr6D-Qks5qLcZggaJpZM4IzNFE>
.

PR #2021, opened 2016-06-10
Remove Docker 1.10 requirement temporarily.

dgoodwin:
Docker 1.10 is not widely available in RHEL / CentOS yet, lets remove
the restriction for users of master / origin. We can revert as soon as
it's available publicly.

@sdodson @brenton @detiber @abutcher what do you guys think? This reverts part of the Docker 1.10 stuff and would allow master users to still install and upgrade while we wait for 1.10. We can revert as soon as it's available.

This is actually pretty close to what @sdodson suggested:

- just require docker to be installed in the docker role
- 3.2 upgrade will always upgrade docker to newest available if we're below 1.9.
- docker 1.10 upgrade playbook is still there for anyone who wants it

No real version checking left. Thoughts?

(NOTE: testing still underway)


sdodson:
+1 from me, even if it's only temporary so we can get tests passing again

sdodson:
aos-ci-test-extended

sdodson:
aos-ci-test-extended

detiber:
I think I'd like to see the ability to install a specific version (only handling the upgrade case, fail if downgrade required) and just not enforce a minimum version.

It'd be nice if we could give the users a warning about potential issues, but we don't have a way to bubble that up yet. There is a facility for returning warnings from an Ansible module, but that would get list in the noise.

We could potentially have a validation playbook that runs and gives useful output...

sdodson:
Merged as is, the failed tests aren't related to this change. We'll follow up to add the ability to specify desired docker version.

PR #2023, opened 2016-06-10
Various hosted component improvements

### LINKED ISSUE #1551: RFE: put nodeSelector label in default namespace
When using labels for region and zones (as shown in https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.origin.example#L156 ) the default namespace will never run a registry or router. 

Fixing it using https://lists.openshift.redhat.com/openshift-archives/users/2016-February/msg00036.html is easy and could be part of a playbook.

abutcher:
Pulled from https://github.com/openshift/openshift-ansible/pull/1978.

**Router**
* Infrastructure node count greatly simplified. Removes the need for the `oo_oc_nodes_matching_selector` and `oo_oc_nodes_with_label` filters.
* Admin binary uses temporary kubeconfig instead of the admin kubeconfig.
* Adds option for router `--force-subdomain`.

For example:
```
openshift_hosted_router_force_subdomain='${name}-${namespace}.apps.example.com'
```

**Registry**
- Registry moved to the `openshift_hosted` role to align with router deployment.
- A registry will be deployed with an emptydir volume by default if infrastructure nodes are present.
- Registry selector changed from `openshift_registry_selector` to `openshift_hosted_registry_selector`.
- New object storage options for s3.

For example:
```
openshift_hosted_registry_storage_kind=object
openshift_hosted_registry_storage_provider=s3
openshift_hosted_registry_storage_s3_accesskey=aws_access_key_id
openshift_hosted_registry_storage_s3_secretkey=aws_secret_access_key
openshift_hosted_registry_storage_s3_bucket=bucket_name
openshift_hosted_registry_storage_s3_region=bucket_region
```

**Service accounts**
- Service accounts have been added as a dependency to the `openshift_hosted` role.
- Users are now granted access to the specified sccs (`openshift_serviceaccounts_sccs`) instead of the privileged scc.

**New role: openshift_projects**
- Adds support for creating additional projects with specified default node selectors.

For example:
```
openshift_additional_projects={'my-project': {'default_node_selector': 'label=value'}}
```

- Standard project node selectors set to an empty string in order to override cluster wide default node selector settings.

Fixes #1551
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1344302
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1328464
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1348518


andrewklau:
Could we add the chunksize as an option to the s3 config file, I've seen better performance by increasing the chunksize.

https://docs.docker.com/registry/storage-drivers/s3/
The default is 10485760

sdodson:
@andrewklau is there a blocksize you've used that seems to work best?

detiber:
I went ahead and took a quick stab at this in my branch. I tested with double the default, but haven't really done any performance testing yet.

andrewklau:
@sdodson Using 26214400 I saw about a 10-20% speed increase for pushes. It will also reduce the number of PUT requests, reducing costs marginally..

sdodson:
CC: @mffiedler for references to optimal chunk size tuning

abutcher:
@tbielawa Could you peep the example inventory changes and let me know if they make sense? Do these need more explanation?

abutcher:
@sdodson @detiber @tbielawa PTAL

abutcher:
@dgoodwin

inventory/byo/hosts.aep.example
```diff
@@ -307,56 +320,46 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
...
+# External NFS Host
+# NFS volume must already exist with path "nfs_directory/_volume_name" on
+# the storage_host. For example, the remote volume path using these
+# options would be "nfs.example.com:/exports/registry"
+#openshift_hosted_registry_storage_kind=nfs
+#openshift_hosted_registry_storage_access_modes=['ReadWriteMany']
+#openshift_hosted_registry_storage_host=nfs.example.com
+#openshift_hosted_registry_storage_nfs_directory=/exports
+#openshift_hosted_registry_storage_volume_name=registry
+#openshift_hosted_registry_storage_volume_size=10Gi
```
dgoodwin: Just curious, any idea if / how size is used here?

abutcher: Volume size is bound by the underlying disk partition size and our NFS role just creates a directory on the root disk so size is just tacked on here and doesn't impact anything. I'm thinking it makes sense to remove these.

abutcher: Same story with external except the size could be set to the size of the volume.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -114,6 +114,12 @@ def migrate_hosted_facts(facts):
             if 'router' not in facts['hosted']:
                 facts['hosted']['router'] = {}
             facts['hosted']['router']['selector'] = facts['master'].pop('router_selector')
+        if 'registry_selector' in facts['master']:
+            if 'hosted' not in facts:
+                facts['hosted'] = {}
+            if 'registry' not in facts['hosted']:
+                facts['hosted']['registry'] = {}
+            facts['hosted']['registry']['selector'] = facts['master'].pop('registry_selector')
```
dgoodwin: Is this a situation where we can just skip openshift_facts entirely? From what I can see this isn't really a fact *about* the system, we're configuring a desired state based either on what's in the inventory, or a default. Would it be possible to just ditch this entirely in openshift_facts, avoid transfering the variable to fact, upgrading old facts, and instead just use openshift_hosted_registry_selector, or a default.

roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -0,0 +1,40 @@
+---
+- name: Retrieve list of openshift nodes matching registry selector
+  command: >
+    {{ openshift.common.client_binary }} --api-version='v1' -o json
+    get nodes -n default --config={{ openshift_hosted_kubeconfig }}
+    --selector={{ openshift.hosted.registry.selector | default('') }}
```
dgoodwin: This would go to openshift_hosted_registry_selector if my above proposal is sane.

roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -0,0 +1,40 @@
...
+  register: openshift_hosted_registry_nodes_json
+  changed_when: false
+  when: openshift.hosted.registry.replicas | default(none) is none
```
dgoodwin: Arguably we could do the same here?

dgoodwin: Let me know if that's infeasible or just outright a terrible idea, I'd just love to see us get away from some situations where the same little bit of data is desired state, and we refer to it by 3-4 different variables/locations throughout.

roles/openshift_facts/library/openshift_facts.py
abutcher: @detiber thoughts on this? Only benefit I can think of is managing the default in one place instead of 5.

inventory/byo/hosts.aep.example
sdodson: If it were using dynamic volume provisioning it would define how large of a volume to create.

abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
aos-ci-test

sdodson:
LGTM, I think we should reach consensus on Devan's questions regarding defaulting in openshift_facts and document that. I struggle with the same questions most of the time I work with openshift-ansible. We can figure that out later though.

abutcher:
aos-ci-test

abutcher:
aos-ci-test

roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -0,0 +1,40 @@
...
+    {% endif -%}
+  register: openshift_hosted_registry_results
+  changed_when: "'service exists' not in openshift_hosted_registry_results.stdout"
+  failed_when: "openshift_hosted_registry_results.rc != 0 and 'service exists' not in openshift_hosted_registry_results.stdout and 'deployment_config' not in openshift_hosted_registry_results.stderr and 'service' not in openshift_hosted_registry_results.stderr"
+
+- include: storage/object_storage.yml
+  when: openshift.hosted.registry.storage.kind | default(none) == 'object'
+
+- include: storage/persistent_volume.yml
+  when: openshift.hosted.registry.storage.kind | default(none) in ['nfs', 'openstack']
```
abutcher: @sdodson re: the issue you hit where we were attempting to attach the PV when there wasn't one.. This now checks if the storage kind is nfs or openstack which are the only two kinds that introduce PVs.

abutcher:
aos-ci-test

playbooks/common/openshift-cluster/openshift_hosted.yml
```diff
@@ -1,30 +1,33 @@
...
-    when: openshift.hosted.metrics.deploy | bool
 
 - name: Create Hosted Resources
   hosts: oo_first_master
+  vars:
+    l_projects:
+      default:
+        default_node_selector: "{{ openshift_hosted_router_selector
+                                   | default(openshift_hosted_infra_selector
+                                   | default(none)) }}"
```
abutcher: @detiber Do you see any issues with setting the default project node selector to whatever is common among router and registry selectors via a filter? If they don't share any `label=value`'s then we can exit with a message.

Also, if `osm_default_node_selector` has been configured I think we need to require that either `openshift_hosted_infra_selector` or each individual selector has been set for this bug https://bugzilla.redhat.com/show_bug.cgi?id=1348518

abutcher: If a default node selector is set cluster wide, these need to default to whatever the default component selector is; `region=infra` in all cases.

roles/openshift_facts/library/openshift_facts.py
detiber: I actually am leaning towards @dgoodwin's pov here. We've discussed leveraging openshift_facts less for "desired state" and more for just retrieving info about the configured system instead. I'd be good with completely skipping openshift_facts here and potentially moving the default into a filter or lookup plugin to allow for re-use.

playbooks/common/openshift-cluster/openshift_hosted.yml
detiber: @abutcher as we discussed, I think the sane default for the default project node selector would be '' rather than trying to determine what the value should be derived from the router/registry node selector. We should be passing in the router/registry node selector values into the router/registry creation commands anyway.

playbooks/common/openshift-cluster/openshift_hosted.yml
```diff
@@ -1,30 +1,33 @@
...
-  - role: openshift_registry
-    registry_volume_claim: "{{ openshift.hosted.registry.storage.volume.name }}-claim"
-    when: deploy_infra | bool and attach_registry_volume | bool
-  - role: openshift_metrics
-    when: openshift.hosted.metrics.deploy | bool
 
 - name: Create Hosted Resources
   hosts: oo_first_master
+  vars:
+    l_projects:
```
detiber: I'm wondering if it makes sense to push this down into the openshift_hosted role now rather than having the definition sitting out here in the playbook.

roles/openshift_hosted/meta/main.yml
```diff
@@ -12,5 +12,23 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
-- openshift_common
-- openshift_hosted_facts
+- role: openshift_common
```
detiber: I'm wondering if it makes sense to put a dependency on openshift_cli, since that is really a requirement here.

roles/openshift_hosted/tasks/main.yml
```diff
@@ -1,3 +1,22 @@
 ---
+- name: Create temp directory for kubeconfig
+  command: mktemp -d /tmp/openshift-ansible-XXXXXX
+  register: mktemp
+  changed_when: False
```
detiber: I was going to say that I can't wait until we can get rid of this, but router/registry will definitely not be covered for the first iteration of the upstream kube/openshift modules.

roles/openshift_hosted/tasks/registry/storage/persistent_volume.yml
```diff
@@ -1,13 +1,8 @@
...
-    {{ openshift.common.admin_binary }} registry
-    --create --replicas={{ openshift.master.infra_nodes | length }}
-    --service-account=registry {{ oreg_selector }}
-    --credentials={{ openshift_master_config_dir }}/openshift-registry.kubeconfig {{ oreg_images }}
-  register: oreg_results
-  changed_when: "'service exists' not in oreg_results.stdout"
+- set_fact:
+    registry_volume_claim: "{{ openshift.hosted.registry.storage.volume.name }}-claim"
 
+# TODO: ensure api version here
```
detiber: If we are targeting 3.3 here, we should definitely use the 1.3/3.3 method of specifying an api version here to be safe.

roles/openshift_hosted/tasks/registry/storage/s3.yml
```diff
@@ -0,0 +1,105 @@
...
+
+- fail:
+    msg: >
+      openshift_hosted_registry_storage_s3_bucket and
+      openshift_hosted_registry_storage_s3_region are required
+  when: openshift.hosted.registry.storage.s3.bucket | default(none) is none or openshift.hosted.registry.storage.s3.region | default(none) is none
+
+# Currently this is relying on the boto config (or env vars) on the master host, it
+# should be a local_action and use separate credentials for creating the
+# bucket (vs the ones for accessing the bucket)
```
detiber: The above comment is only partially correct...  it is a local_action, but I think we should definitely add the ability to specify the aws creds used for this task...

roles/openshift_hosted/tasks/registry/storage/s3.yml
```diff
@@ -0,0 +1,105 @@
...
+
+- set_fact:
+    registry_config: "{{ lookup('template', '../templates/registry_config.j2') | b64encode }}"
+
+- set_fact:
+    registry_config_secret: "{{ lookup('template', '../templates/registry_config_secret.j2') | from_yaml }}"
+
+# TODO: update config secret if different
+#       need to make sure update will not update backend storage. since that
+#       is destructive
```
detiber: I'm wondering if it is worth adding a filter here to address this comment, I worry that there will be people that want to go back to add s3 storage after the fact, or want to update things like the chunk size and the behavior being counterintuitive.

roles/openshift_hosted/tasks/registry/storage/s3.yml
```diff
@@ -0,0 +1,105 @@
...
+# bucket (vs the ones for accessing the bucket)
+- name: Create the S3 bucket
+  become: no
+  local_action:
+    module: s3
+    bucket: "{{ openshift.hosted.registry.storage.s3.bucket }}"
+    region: "{{ openshift.hosted.registry.storage.s3.region }}"
+    mode: create
+  when: openshift.hosted.registry.storage.s3.create | default(false) | bool
+
```
detiber: Everything below this point is most likely common to all object storage configuration and not just s3.

roles/openshift_hosted_facts/tasks/main.yml
```diff
@@ -1,7 +1,22 @@
 ---
+- set_fact:
+    openshift_hosted_router_selector: "{{ openshift_hosted_infra_selector }}"
+  when: openshift_hosted_router_selector is not defined and openshift_hosted_infra_selector is defined
+- set_fact:
+    openshift_hosted_registry_selector: "{{ openshift_hosted_infra_selector }}"
+  when: openshift_hosted_registry_selector is not defined and openshift_hosted_infra_selector is defined
+- set_fact:
+    openshift_hosted_logging_selector: "{{ openshift_hosted_infra_selector }}"
+  when: openshift_hosted_logging_selector is not defined and openshift_hosted_infra_selector is defined
```
detiber: openshift_hosted_{logging,metrics}_selector should probably be dropped from this PR.

roles/openshift_serviceaccounts/tasks/main.yml
```diff
@@ -24,11 +24,11 @@
...
   with_items: "{{ openshift_serviceaccounts_sccs }}"
 
-- name: Grant the user access to the privileged scc
+- name: Grant the user access to the appropriate scc
   command: >
       {{ openshift.common.admin_binary }} policy add-scc-to-user
-      privileged system:serviceaccount:{{ openshift_serviceaccounts_namespace }}:{{ item.0 }}
-  when: "openshift.common.version_gte_3_1_or_1_1 and item.1.rc == 0 and 'system:serviceaccount:{{ openshift_serviceaccounts_namespace }}:{{ item.0 }}' not in {{ (item.1.stdout | from_yaml).users }}"
+      {{ item.1.item }} system:serviceaccount:{{ openshift_serviceaccounts_namespace }}:{{ item.0 }}
+  when: "openshift.common.version_gte_3_1_or_1_1 and item.1.rc == 0 and 'system:serviceaccount:{{ openshift_serviceaccounts_namespace }}:{{ item.0 }}' not in {{ (item.1.stdout | from_yaml).users | default([]) }}"
```
detiber: Not really applicable to this PR, but I'm wondering if this needs to be more flexible than just applying sccs, there are valid use cases for applying clusterroles or roles within a namespace to a service account and not just sccs.

roles/openshift_facts/library/openshift_facts.py
detiber: To expand on this a bit more. If we do switch to not using openshift_facts here for the desired state, we would need to add introspection for the current state from the existing configured project node selectors and/or the node selectors defined on the DCs for the router/registry in particular.

roles/openshift_hosted/tasks/registry/storage/persistent_volume.yml
abutcher: Adding `--output-version`, we'll need a follow up PR to address 3.3 object.vX across the board.

roles/openshift_hosted/tasks/registry/storage/s3.yml
abutcher: This is also a local_action with become which requires ansible v2.

detiber: doh, indeed. I'd suggest that we either require that the s3 bucket be pre-created for the time being, or just delay these changes until we ship Ansible 2.1

abutcher: Let's require pre-created. I'll create an issue to address creating the bucket.

roles/openshift_facts/library/openshift_facts.py
abutcher: I'd like to tackle this as a follow up PR if that is agreeable with y'all

abutcher:
Comments addressed for the most part. @detiber @sdodson PTAL

roles/openshift_facts/library/openshift_facts.py
sdodson: I'm fine with that, needs to be sorted out as we decide how to proceed with openshift_facts in general, not directly related to this PR.

roles/openshift_hosted/meta/main.yml
```diff
@@ -12,5 +12,24 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
-- openshift_common
-- openshift_hosted_facts
+- role: openshift_cli
+- role: openshift_hosted_facts
+- role: openshift_projects
+  openshift_projects: "{{ openshift_additional_projects | default({}) | oo_merge_dicts(hosted_projects | default({})) }}"
```
abutcher: Ran into problems accessing `hosted_projects` here from `roles/openshift_hosted/vars/main.yml` with ansible-1.9.4. `hosted_projects` likely needs to be defined here similar to how we do `os_firewall_allow` with other roles.

abutcher: Moved definitions to openshift_hosted/meta for now.

roles/openshift_facts/library/openshift_facts.py
detiber: :+1: on waiting for any openshift_facts behavioral changes.

roles/openshift_hosted/tasks/registry/registry.yml
detiber: OpenStack introduces a PV? I'm not sure we would want that...  Ideally OpenStack would leverage the swift object store config rather than a PV.

detiber: Granted we would need to actually make the swift support work :)

detiber:
based on a quick review, lgtm

sdodson:
LGTM

abutcher:
aos-ci-test

abutcher:
aos-ci-test

roles/openshift_hosted/templates/registry_config.j2
```diff
@@ -0,0 +1,70 @@
...
+{%   if 'rootdirectory' in openshift.hosted.registry.storage.gcs %}
+    rootdirectory: {{ openshift.hosted.registry.storage.gcs.rootdirectory }}
+{%   endif -%}
+{% endif -%}
+auth:
+  openshift:
+    realm: openshift
+middleware:
+  repository:
+  - name: openshift
```
jianlinliu: Is it possible to enable pullthrough by default like the following?
middleware:
  repository:
    - name: openshift
      options:
        pullthrough: true

abutcher: https://github.com/openshift/openshift-ansible/pull/2138

Issue #2018, opened 2016-06-10
Allow the ability to pass through an environment variables from the installer to the sysconfig or unit-env files.


timothysc:
Kubernetes has a series of environment variables which allow some features to be turned on and off.  These features are typically centralized behaviors that would hit all the components.  

An example of this is: https://github.com/kubernetes/kubernetes/pull/25280/files#diff-2bcb1660b642c577632c4fc884f1d411R81 which will allow the system to enable http2 across the system.   

For large scale customers this would decrease their connection counts to a manageable level. 

/cc @rrati @jeremyeder @mffiedler

timothysc:
awesome, thx abutcher..

PR #2025, opened 2016-06-10
Per-service environment variables

abutcher:
```
# OpenShift Per-Service Environment Variables
# Environment variables are added to /etc/sysconfig files for
# each OpenShift service: node, master (api and controllers).
# API and controllers environment variables are merged in single
# master environments.
#openshift_master_api_env_vars={"ENABLE_HTTP2": "true"}
#openshift_master_controllers_env_vars={"ENABLE_HTTP2": "true"}
#openshift_node_env_vars={"ENABLE_HTTP2": "true"}
```

@detiber @timothysc PTAL
Resolves: #2018

### DIFF
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -410,6 +410,15 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # Configure usage of openshift_clock role.
 #openshift_clock_enabled=true
 
+# OpenShift Per-Service Environment Variables
+# Environment variables are added to /etc/sysconfig files for
+# each OpenShift service: node, master (api and controllers).
+# API and controllers environment variables are merged in single
+# master environments.
+#openshift_master_api_env_vars={"ENABLE_HTTP2": "true"}
+#openshift_master_controllers_env_vars={"ENABLE_HTTP2": "true"}
+#openshift_node_env_vars={"ENABLE_HTTP2": "true"}
+
 # host group for masters
 [masters]
 ose3-master[1:3]-ansible.test.example.com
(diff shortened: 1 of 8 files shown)


inventory/byo/hosts.ose.example
```diff
@@ -410,6 +410,11 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
 # Configure usage of openshift_clock role.
 #openshift_clock_enabled=true
 
+# Global OpenShift Environment Variables
+# Environment variables are added to /etc/sysconfig files for
+# each OpenShift service (node, master, api, controllers).
+#openshift_global_env_vars={"ENABLE_HTTP2": "true"}
```
detiber: I don't really like this approach with just a single var for setting things across all services.

That said, I think there is room to have a var that sets things at a high level.

Thoughts on a top level var (OPENSHIFT_SERVICE_ENV_VARS), and then one for each service (maybe a var for applying to both master services). It would require merging the vars before applying, but it would allow for big and small hammers and allow for overriding specific vars on a specific service (or a subset of hosts with that service depending on where the vars are defined).

detiber:
Copying my line comment here, since it wasn't really a line comment:

I don't really like this approach with just a single var for setting things across all services.

That said, I think there is room to have a var that sets things at a high level.

Thoughts on a top level var (OPENSHIFT_SERVICE_ENV_VARS), and then one for each service (maybe a var for applying to both master services). It would require merging the vars before applying, but it would allow for big and small hammers and allow for overriding specific vars on a specific service (or a subset of hosts with that service depending on where the vars are defined).

abutcher:
@detiber Updated to include per-service environment variables. I don't think we should have global vars for all services or the single master service if we can accomplish everything with these and merge api/controllers for single master.

timothysc:
I don't really have strong opinions on the matter.  I do know that upstream only really uses env-vars for features that cut across many components, otherwise they prefer command line options and ConfigData.

abutcher:
aos-ci-test

detiber:
lgtm

PR #2033, opened 2016-06-13
[WIP][DO NOT MERGE] Add /etc/sysconfig/docker to node mounts

sdodson:
The origin image doesn't have docker installed in the container, or at least not
a version that has /etc/sysconfig/docker so when running docker inside that
container it will fail to load /etc/sysconfig/docker. This is an additional
workaround for the fallout of docker and docker-latest.


sdodson:
Origin v1.2.0 does not have the chroot fix that made this un-necessary. Until we ship v1.3.0 this is necessary.

sdodson:
Duped of #2037

PR #2034, opened 2016-06-13
Fix docker 1.10 upgrade on embedded etcd masters.

sdodson:
The tasks were attempting to stop/start etcd, which would be fine on the
stop but on start could actually kick the non-containerized etcd service
which happens to be layed down even though it's unused.

When the service was requested to start again it would claim the port
embedded etcd needs and the master would then fail to come up.

Instead use the correct etcd_container service.


sdodson:
Backports #2020

PR #2037, opened 2016-06-14
also volume-mount /etc/sysconfig/docker

ibotty:
for recent RHEL (or CentOS) atomic hosts with docker-recent/docker-latest split, it's neccessary to mount /etc/sysconfig/docker, because the /usr/bin/docker shell script reads this file and fails, if it does not exist.


sdodson:
LGTM

sdodson:
We'll rip all this out once 1.3 images ship which include the running `chroot /rootfs docker`

ibotty:
I hope so!

sdodson:
Thanks BTW!

PR #2040, opened 2016-06-14
a-o-i: Remove 3.2 options for 3.1 quick installer

smunilla:
Remove any option for installing 3.2 in this version and add some logic to
instruct the user when they need to update the installer before attempting
a major upgrade of their cluster.


sdodson:
Disable the pylint check for too many lines?

smunilla:
aoi-ci-test

sdodson:
Does this need to be rebased? looks like there's a lot of unassociated changes.

smunilla:
@sdodson That should be much better now. Not sure what was up there.

sdodson:
Should we also remove playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/ ?

sdodson:
I think this is sufficient, though I do wonder if we should remove the other v3_1_to_v3_2 playbooks.

smunilla:
Might as well remove those while we're here. I'll update.

sdodson:
QE has reported https://bugzilla.redhat.com/show_bug.cgi?id=1355926 against this PR

PR #2041, opened 2016-06-14
Fix no proxy hostnames during upgrade.

dgoodwin:
This value not being set was causing missing hostnames in the sysconfig
files with NO_PROXY.

This is not the same way we set it during config playbooks, they use
vars definitions but this is too difficult in upgrade as there are too
many roles that might need it set.


dgoodwin:
@sdodson ping.

sdodson:
LGTM, going to give this a quick test.

sdodson:
tested locally, fixes issue described

sdodson:
Backported in #2043

PR #2052, opened 2016-06-16
Metrics improvements

sdodson:
- Metrics deployment is currently disabled by default. Enable it by setting `openshift_hosted_metrics_deploy=true`
- Metrics currently only supports one single cassandra node. This has been shown to be sufficient for up to 1000 pods. If you require multiple cassandra nodes you need to deploy metrics manually and ensure that you have enough persistent volumes to match your cassandra node count.
- If you're using the nfs role to automatically provision exports you may add the following
 - `openshift_hosted_metrics_storage_kind=nfs`
 - `openshift_hosted_metrics_storage_volume_name=metrics`
 - `openshift_hosted_metrics_storage_volume_size=10Gi`

### DIFF
--- a/roles/openshift_storage_nfs/tasks/main.yml
+++ b/roles/openshift_storage_nfs/tasks/main.yml
@@ -20,13 +20,14 @@
 
 - name: Ensure export directories exist
   file:
-    path: "{{ openshift.hosted.registry.storage.nfs.directory }}/{{ item }}"
+    path: "{{ item.storage.nfs.directory }}/{{ item.storage.volume.name }}"
     state: directory
     mode: 0777
     owner: nfsnobody
     group: nfsnobody
   with_items:
-  - "{{ openshift.hosted.registry.storage.volume.name }}"
+  - "{{ openshift.hosted.registry }}"
+  - "{{ openshift.hosted.metrics }}"
 
 - name: Configure exports
   template:
--- a/roles/openshift_storage_nfs/tasks/main.yml
+++ b/roles/openshift_storage_nfs/tasks/main.yml
@@ -30,8 +30,23 @@
 
 - name: Configure exports
   template:
-    dest: /etc/exports
+    dest: /etc/exports.d/openshift-ansible.exports
     src: exports.j2
+  notify:
+  - restart nfs-server
+
+# Now that we're putting our exports in our own file clean up the old ones
+- name: register exports
+  command: cat /etc/exports.d/openshift-ansible.exports
+  register: exports_out
+
+- name: remove exports from /etc/exports
+  lineinfile:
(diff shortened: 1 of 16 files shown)


sdodson:
[test]

roles/openshift_storage_nfs/tasks/main.yml
```diff
@@ -30,11 +30,24 @@
...
+#  command: cat /etc/exports.d/openshift-ansible.exports
+#  register: exports_out
+#- name: remove exports
+#  line_in_file:
+#    src: /etc/exports
+#    line: {{ item }}
+#    state: absent
+#  with_items: exports_out.stdout.lines
+#  when: exports_out.rc == 0
+#
```
sdodson: @abutcher i imagine there's a simpler way to do this, right?

abutcher: Not unless we just remove the ones we're adding?

detiber: @sdodson @abutcher considering that we don't really consider the nfs host to be mean for production, I don't really see an issue with not cleaning up after ourselves here.

roles/openshift_storage_nfs/tasks/main.yml
```diff
@@ -30,11 +30,24 @@
 
 - name: Configure exports
   template:
-    dest: /etc/exports
+    dest: /etc/exports.d/openshift-ansible.exports
```
detiber: nice!

roles/openshift_storage_nfs/tasks/main.yml
sdodson: I got it figured out, I'll deploy /etc/exports.d/openshift-ansible.exports and remove any entries that exist there from /etc/exports. That way the uninstall playbooks can nuke /etc/exports.d/openshift-ansible.exports without fear of deleting something else even though nothing else will exist in most cases. I'm just running into this because i'm installing, uninstalling, installing...

sdodson:
@detiber ok i think i'm done with all functional changes.

sdodson:
aos-ci-test

abutcher:
Pylint errors:
```
************* Module openshift_facts
C:469, 0: Exactly one space required around assignment
    metrics_selector="region=metrics"
                    ^ (bad-whitespace)
C:470, 0: Exactly one space required around assignment
    logging_selector="region=logging"
                    ^ (bad-whitespace)
W:469, 4: Unused variable 'metrics_selector' (unused-variable)
W:470, 4: Unused variable 'logging_selector' (unused-variable)
```

sdodson:
aos-ci-test

sdodson:
aos-ci-test

PR #2054, opened 2016-06-16
[WIP] Preserve registry ip on upgrade

sdodson:
If a registry exists when upgrading then figure out what its IP address is and add it to the no_proxy list for docker.
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1347431

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/pre.yml
@@ -7,16 +7,29 @@
   roles:
   - openshift_facts
   - openshift_repos
+  
+- name: Obtain registry service-ip
+  hosts: oo_first_master
+  tasks:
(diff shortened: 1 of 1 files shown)


sdodson:
aos-ci-test-extended

sdodson:
@abutcher is there a better way to do this than setting a fact on oo_first_master?

detiber:
@sdodson I don't think there is a good way until we have the upstream modules in place.

sdodson:
Is this acceptable for now or should I close this? If we keep this open I'd like to add this logic to the install playbooks such that if someone were to re-run them then the registry ip, if it exists, is included in the no_proxy.

detiber:
I see no issues with adding this...  The one question I have regarding it for installation is how do we handle the chicken/egg scenario of configuring NO_PROXY prior to deploying the registry?

sdodson:
I wasn't planning on handling that, just the re-running the installer scenario.

PR #2055, opened 2016-06-16
Stop dumping debug output, re-try startng the node once

sdodson:
When we dump the logs from a failed node it contains control characters that break the quick installer. While dumping this information is useful for CI jobs where it's often hard to determine why something has failed it's more important that if a percentage of the nodes fail to install with the quick installer that things progress normally on those nodes that haven't failed.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1347209

### DIFF
--- a/roles/openshift_node/tasks/main.yml
+++ b/roles/openshift_node/tasks/main.yml
@@ -118,14 +118,10 @@
   service: name={{ openshift.common.service_type }}-node enabled=yes state=started
   register: node_start_result
   ignore_errors: yes
-
-- name: Check logs on failure
-  command: journalctl -xe
-  register: node_failure
-  when: node_start_result | failed
-
-- name: Dump failure information
-  debug: var=node_failure
+  
+- name: Start and enable node again
+  service: name={{ openshift.common.service_type }}-node enabled=yes state=started
+  register: node_start_result
   when: node_start_result | failed
 
 - set_fact:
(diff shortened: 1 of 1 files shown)


sdodson:
aos-ci-test

sdodson:
@dgoodwin PTAL, how often have you seen the node fail the first time on containerized installs? For me it seems to happen in frequently but for CI it happens more frequently.

dgoodwin:
Very rare to see that in my testing, what do we know about why it happens?

sdodson:
All of the recent CI jobs that've failed were due to node-dep not being
started, which you fixed. Otherwise I do think it's incredibly rare.

dgoodwin:
Are we fairly confident there's no way for the quick installer to NOT break on the control characters? If there's no reasonable way to do that then this LGTM, but it will be a little unfortunate to loose insight into why a CI node failed to start. We could play it by ear and merge this,  and see how often that bites us if at all.

sdodson:
@smunilla any quick fix to keep the installer from breaking on control
characters?

dgoodwin:
If we go ahead with this how about a little sleep in between if the first attempt failed. I've seen this happen in AWS when Docker is starting up for the first time.

sdodson:
aos-ci-test

smunilla:
@sdodson I can't find any quick fix. Disabling it seems good to me.

sdodson:
aos-ci-test

PR #2065, opened 2016-06-20
[WIP] Merge recent changes into enterprise-3.2

sdodson:
Since we've not yet introduced any code that *requires* ansible 2.0/2.1 we'll go ahead and pull everything into enterprise-3.2


sdodson:
@abutcher as soon as we merge something into master that requires 2.x lets bring this up to date and merge it.

sdodson:
@dgoodwin this will remove the docker-1.10 requirement for 3.2 since it was removed in master. Before we merge this we should decide what the right thing to do for 3.2 is.

dgoodwin:
Hard one, I like the state of master personally, but we can't ignore that Docker 1.10 really should be in place going forward. It's probably correct to keep the 1.10 requirement for 3.2 if we know it will be available by then.

sdodson:
meh, i over-wrote my branch, closing this

PR #2067, opened 2016-06-21
Make /var/lib/origin mounted rslave

sdodson:
When docker-1.10 is in use set the rslave mount option on /var/lib/origin

https://trello.com/c/FGEj11M6/243-3-allow-containerized-kubelet-to-support-docker-s-per-mount-propagation-mode
mirrors origin changes here https://github.com/openshift/origin/pull/9459


ncdc:
@pmorie

ncdc:
@sdodson what is openshift.common.data_dir - is that /var/lib/origin?

sdodson:
@ncdc yes

ncdc:
I'm pretty sure this LGTM. @smarterclayton can you think of any problems this might create?

pmorie:
LGTM cc @rootfs

smarterclayton:
@csrwng do we need to change cluster up to handle this case?

roles/openshift_node/templates/openshift.docker.node.service
```diff
@@ -15,7 +15,7 @@ After={{ openshift.common.service_type }}-node-dep.service
...
+ExecStart=/usr/bin/docker run --name {{ openshift.common.service_type }}-node --rm --privileged --net=host --pid=host --env-file=/etc/sysconfig/{{ openshift.common.service_type }}-node -v /:/rootfs:ro -e CONFIG_FILE=${CONFIG_FILE} -e OPTIONS=${OPTIONS} -e HOST=/rootfs -e HOST_ETC=/host-etc -v {{ openshift.common.data_dir }}:{{ openshift.common.data_dir }}{{ ':rslave' if openshift.docker.gte_1_10 | default(False) | bool else '' }} -v {{ openshift.common.config_base }}/node:{{ openshift.common.config_base }}/node {% if 'cloudprovider' in openshift and 'kind' in openshift.cloudprovider and openshift.cloudprovider.kind != '' -%} -v {{ openshift.common.config_base }}/cloudprovider:{{ openshift.common.config_base}}/cloudprovider {% endif -%} -v /etc/localtime:/etc/localtime:ro -v /etc/machine-id:/etc/machine-id:ro -v /run:/run -v /sys:/sys:ro -v /usr/bin/docker:/usr/bin/docker:ro -v /var/lib/docker:/var/lib/docker -v /lib/modules:/lib/modules -v /etc/origin/openvswitch:/etc/openvswitch -v /etc/origin/sdn:/etc/openshift-sdn -v /etc/systemd/system:/host-etc/systemd/system -v /var/log:/var/log -v /dev:/dev $DOCKER_ADDTL_BIND_MOUNTS {{ openshift.node.node_image }}:${IMAGE_VERSION}
```
sdodson: `{{ ':rslave' if openshift.docker.gte_1_10 | default(False) | bool else '' }}`

@abutcher should i move this out to a variable in the task or leave it here?

detiber: Since the conditional is temporary (just needed for 3.2, since 3.3 will require 1.10), I would say we just need it for the 3.2 branch and for master we can just set ':rslave'

sdodson: I always lean towards less branch divergence but I can do that only on enterprise-3.2 if you'd like.

abutcher: It'll be safe for either -- cool with leaving it here.

csrwng:
@smarterclayton yes 'cluster up' should do the same thing, I'll create an issue.

PR #2074, opened 2016-06-22
NetworkManager service never changes

tbielawa:
* Don't ever report the NetworkManager service as changed when
  preparing to set the 'network_manager_active' fact

Also clean up a bunch of trailing whitespace


sdodson:
aos-ci-test

PR #2085, opened 2016-06-27
Make image stream munging optional

sdodson:
Because we don't currently support also modifying image streams to add insecure annotation or secrets required to pull from a private authenticated registry we should make this an opt-in change.


sdodson:
@abutcher PTAL, this is necessary for us to merge master into enterprise-3.2

abutcher:
lgtm

PR #2087, opened 2016-06-27
Speed up copying OpenShift examples

tbielawa:
This is a work in progress of a change to try and speed up the example copying process:

* The `copy` module does not scale to hundreds of files
* Ultimately we'll replace `copy` with `unarchive`

Testing with the `profile_tasks` callback showed that copying the ~80 files from `roles/openshift_examples/files/examples/{{ content_version }}` was taking aroud 70 seconds+ on a speedy broadband connection to an EC2 instance.

Switching to the `unarchive` approach reduced the total time down to ~2 seconds. Testing with the [synchronize module](http://docs.ansible.com/ansible/synchronize_module.html) attempted first, however due to behavior changes introduced in Ansible 2.x+ we had to ditch that approach (issues with `become`, google for it if you're interested).

**WIP** Because this is still incomplete (need to erase the temp dir, etc) and I would like feedback on my approach to some references. For example, when creating a tarball of the examples this change uses the `-C` option and tell it to switch into the examples directory as

    tar -C "roles/openshift_examples/files/examples/{{ content_version }}/" ...

Is there a better way to determine that path?

### DIFF
--- a/roles/openshift_examples/tasks/main.yml
+++ b/roles/openshift_examples/tasks/main.yml
@@ -1,8 +1,45 @@
 ---
-- name: Copy openshift examples
-  copy:
-    src: "examples/{{ content_version }}/"
+######################################################################
+# Copying Examples
+#
+# We used to use the copy module to transfer the openshift examples to
+# the remote. Then it started taking more than a minute to transfer
+# the files. As noted in the module:
+#
+#   "The 'copy' module recursively copy facility does not scale to
+#   lots (>hundreds) of files."
+#
+# The `synchronize` module is suggested as an alternative, we can't
+# use it either due to changes introduced in Ansible 2.x.
+- name: Create local temp dir for OpenShift examples copy
+  local_action: command mktemp -d /tmp/openshift-ansible-XXXXXXX
+  become: False
+  register: copy_examples_mktemp
+  run_once: True
+
+- name: Create tar of OpenShift examples
+  local_action: command tar -C "{{ role_path }}/files/examples/{{ content_version }}/" -cvf "{{ copy_examples_mktemp.stdout }}/openshift-examples.tar" .
+  become: False
+  register: copy_examples_tar
+
+- name: Create the remote OpenShift examples directory
+  file:
+    dest: "{{ examples_base }}"
+    state: directory
+    mode: 0755
+
+- name: Unarchive the OpenShift examples on the remote
+  unarchive:
+    src: "{{ copy_examples_mktemp.stdout }}/openshift-examples.tar"
     dest: "{{ examples_base }}/"
+
+- name: Cleanup the OpenShift Examples temp dir
+  become: False
+  local_action: file dest="{{ copy_examples_mktemp.stdout }}" state=absent
+
+# Done copying examples
+######################################################################
+# Begin image streams
 
 - name: Modify registry paths if registry_url is not registry.access.redhat.com
   shell: >


detiber:
Do you know if the synchronize issue is fixed in 2.1.1? If so, it may be worth it to work on getting that pulled in asap, since it fixes having to specify become on local_action and delegate_to as well.

detiber:
Determining the path can be a bit tricky, you can dump the environment variable to see if there is a variable that will work.

tbielawa:
@detiber I experimented with the `synchronize` module for over an hour trying to get it to work.

I tried following suggestions from this site http://cheat.readthedocs.io/en/latest/ansible/synchronize.html but nothing worked out. I thought there would be some hope in this upstream ansible issue https://github.com/ansible/ansible/issues/13825 but after reading that thread I couldn't find any new work-around I had not tried yet.

In the end the error always wound up being about rsync and sudo not having a tty available. I'd love to use sync instead but I just couldn't get the dang thing to work.

roles/openshift_examples/tasks/main.yml
```diff
@@ -1,8 +1,45 @@
...
+# use it either due to changes introduced in Ansible 2.x.
+- name: Create local temp dir for OpenShift examples copy
+  local_action: command mktemp -d /tmp/openshift-ansible-XXXXXXX
+  become: False
+  register: g_copy_examples_mktemp
+  changed_when: False
+  run_once: True
+
+- name: Create tar of OpenShift examples
+  local_action: command tar -C "roles/openshift_examples/files/examples/{{ content_version }}/" -cvf "{{ g_copy_examples_mktemp.stdout }}/openshift-examples.tar" .
```
abutcher: Ansible 1.8 added a `role_path` variable that might be cool to use here.

```
TASK [openshift_examples : debug] **********************************************
Monday 27 June 2016  10:44:40 -0400 (0:00:00.283)       0:00:22.261 ***********
ok: [master4.example.com] => {
    "role_path": "/home/abutcher/rhat/openshift-ansible/roles/openshift_examples"
}
```

tbielawa: So then the original command:

    +  local_action: command tar -C "roles/openshift_examples/files/examples/{{ content_version }}/" -cvf "{{ g_copy_examples_mktemp.stdout }}/openshift-examples.tar" .


Would turn into:

    +  local_action: command tar -C "{{ role_path }}/files/examples/{{ content_version }}/" -cvf "{{ g_copy_examples_mktemp.stdout }}/openshift-examples.tar" .

? Sounds reasonable to me.

abutcher: Yep, +1 from me.

tbielawa:
aos-ci-test

tbielawa:
Updating with some more tweaks. Also, fixing the cleanup task. It needs to run locally, not on remote.

tbielawa:
aos-ci-test

tbielawa:
aos-ci-test

abutcher:
aos-ci-test

tbielawa:
Squashed down to one commit.

tbielawa:
aos-ci-test

tbielawa:
@detiber @abutcher good to merge? :+1: ?

abutcher:
lgtm

sdodson:
aos-ci-test

PR #2082, opened 2016-06-24
Don't upgrade docker on non-containerized etcd.

abutcher:
```
TASK [Upgrade Docker] **********************************************************
Friday 24 June 2016  11:57:43 -0400 (0:00:00.444)       0:02:00.898 ***********
skipping: [master6.example.com]
skipping: [master4.example.com]
fatal: [master5.example.com]: FAILED! => {"failed": true, "msg": "The conditional check 'pkg_check.rc == 0 and g_docker_version.curr_version | version_compare('1.9','<')' failed. The error was: error while evaluating conditional (pkg_check.rc == 0 and g_docker_version.curr_version | version_compare('1.9','<')): 'g_docker_version' is undefined\n\nThe error appears to have been in '/home/abutcher/rhat/openshift-ansible/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2/docker_upgrade.yml': line 7, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Upgrade Docker\n  ^ here\n"}
```

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1348127

@sdodson @dgoodwin PTAL

PR #2089, opened 2016-06-27
Don't upgrade docker on non-containerized etcd.

sdodson:
Backports #2082 -- needs a bit of re-factoring to work on enterprise-3.2 branch.


sdodson:
@abutcher mind taking a look since this was less than straight forward?

sdodson:
Actually, as is it's going to attempt to evacuate etcd hosts I think. :-) need to fix that

PR #2090, opened 2016-06-27
Don't upgrade docker on non-containerized etcd.

abutcher:
Backports #2082 -- needed refactored to work with enterprise-3.2. I've brought in only the changes necessary for upgrades to work with external etcd.


sdodson:
LGTM

PR #2092, opened 2016-06-27
Insert iptables rules before DROP or REJECT

### LINKED ISSUE #2084: [BUG] iptable insert
The following doesn't appear to handle a situation where the last rule in the INPUT chain is a DROP and we've witnessed behavior where the installer is appending the etcd 8443 rule after the DROP. 

"os_firewall_manage_iptables.py" line 155 of 273

                # Naively assume that if the last row is a REJECT rule, then
                # we can add insert our rule right before it, otherwise we
                # assume that we can just append the rule.
                if (last_rule_num and last_rule_target
                        and last_rule_target == 'REJECT'):
                    # insert rule
                    cmd = self.cmd + ['-I', self.jump_rule_chain,
                                      str(last_rule_num)]
                else:
                    # append rule
                    cmd = self.cmd + ['-A', self.jump_rule_chain]

Suggested Fix

               # Naively assume that if the last row is a REJECT rule, then
                # we can add insert our rule right before it, otherwise we
                # assume that we can just append the rule.
                if (last_rule_num and last_rule_target
                        and ( last_rule_target == 'REJECT' || last_rule_target == 'DROP')):
                    # insert rule
                    cmd = self.cmd + ['-I', self.jump_rule_chain,
                                      str(last_rule_num)]
                else:
                    # append rule
                    cmd = self.cmd + ['-A', self.jump_rule_chain]
                cmd += ['-j', self.chain]

abutcher:
Resolves #2084

### DIFF
--- a/roles/os_firewall/library/os_firewall_manage_iptables.py
+++ b/roles/os_firewall/library/os_firewall_manage_iptables.py
@@ -37,14 +37,14 @@ class IpTablesSaveError(IpTablesError):
 
 
 class IpTablesCreateChainError(IpTablesError):
-    def __init__(self, chain, msg, cmd, exit_code, output): # pylint: disable=too-many-arguments, line-too-long
+    def __init__(self, chain, msg, cmd, exit_code, output): # pylint: disable=too-many-arguments, line-too-long, redefined-outer-name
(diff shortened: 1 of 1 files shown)


abutcher:
aos-ci-test

sdodson:
LGTM

PR #2094, opened 2016-06-28
Set any_errors_fatal for etcd facts play

abutcher:
This PR sets `any_errors_fatal` for etcd facts play so that if the facts play fails for any reason, playbook execution will halt and we won't attempt to use those facts later.

https://bugzilla.redhat.com/show_bug.cgi?id=1348846


abutcher:
aos-ci-test

abutcher:
@sdodson ptal

abutcher:
Actually, we should set `any_errors_fatal` for the initial facts play too

abutcher:
aos-ci-test

sdodson:
LGTM

PR #2101, opened 2016-06-29
Switch to repoquery, enable plugins for satellite support

sdodson:


dgoodwin:
@sdodson is this use of repoquery --plugins safe to use everywhere we define repoquery_cmd I assume?

sdodson:
Yeah it should be.

PR #2095, opened 2016-06-29
Enable additional 'virt_sandbox_use_nfs' seboolean as per documentation.

georgegoh:
Enable additional 'virt_sandbox_use_nfs' seboolean as per documentation:

https://docs.openshift.org/latest/install_config/persistent_storage/persistent_storage_nfs.html#nfs-selinux

PR #2102, opened 2016-06-29
Enable additional 'virt_sandbox_use_nfs' seboolean as per documentation:

sdodson:
https://docs.openshift.org/latest/install_config/persistent_storage/persistent_storage_nfs.html#nfs-selinux


sdodson:
Backports #2095

PR #2104, opened 2016-06-30
No Package matching 'atomic-openshift3.2.0.46'

cw-aleks:
I need to install a dedicated version `openshift_version=3.2.0.46`

I tried it and got the following error.

```
TASK: [openshift_common | Install the base package for versioning] ************
failed: [<IP>] => {"changed": false, "failed": true, "rc": 0, "results": []}
msg: No Package matching 'atomic-openshift3.2.0.46' found available, installed or updated
```

To be able to install a dedicated version you need to use the following syntax

`atomic-openshift-3.2.0.46`

### DIFF
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -24,7 +24,7 @@
   when: openshift_master_ha | bool and openshift_master_cluster_method == "pacemaker" and openshift.common.is_containerized | bool
 
 - name: Install Master package
-  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-master{{ openshift_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
+  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-master-{{ openshift_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
   when: not openshift.common.is_containerized | bool
 
 - name: Pull master image


roles/openshift_master/tasks/main.yml
```diff
@@ -24,7 +24,7 @@
   when: openshift_master_ha | bool and openshift_master_cluster_method == "pacemaker" and openshift.common.is_containerized | bool
 
 - name: Install Master package
-  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-master{{ openshift_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
+  action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-master-{{ openshift_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
```
sdodson: I think oo_image_tag_to_rpm_version should be adding in the dash. can you check what's happening there?

cw-aleks:
@sdodson Well no this does not helps.

# NOK

With this settings I was not able to install openshift 3.2.0

```
TASK: [openshift_common | Install the base package for versioning] ************
failed: [<IP>] => {"changed": false, "failed": true, "rc": 0, "results": []}
msg: No Package matching 'atomic-openshift-' found available, installed or updated

```

The ansible call
```
/bin/ansible-playbook --inventory-file=/root/.config/openshift/hosts_with_version \
    /usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/config.yml
```
The var section from **hosts_with_version**
```
[OSEv3:vars]
ansible_ssh_user=root
openshift_master_cluster_method=native
openshift_master_cluster_hostname=<hostname>
openshift_master_cluster_public_hostname=<hostname>
openshift_version=3.2.0.46
oo_image_tag_to_rpm_version={{ openshift_version }}
```

# OK
With this settings I was able to install openshift 3.2.0

I have also added in `/usr/share/ansible/openshift-ansible/roles/openshift_node/tasks/main.yml` some dashes

```
[OSEv3:vars]
ansible_ssh_user=root
openshift_master_cluster_method=native
openshift_master_cluster_hostname=<hostname>
openshift_master_cluster_public_hostname=<hostname>
openshift_version=3.2.0.46
openshift_pkg_version={{ openshift_version }}
```

roles/openshift_master/tasks/main.yml
detiber: @sdodson I'm wondering if it's the use of:
> `oo_image_tag_to_rpm_version={{ openshift_version }}`

There seems to be a lot going on here, it could also potentially be that depending on the version of ansible that `include_dash=True` is not being evaluated properly within the `oo_image_tag_to_rpm_version` filter, maybe we need to attempt to cast the variable rather than assuming it is a boolean value.

abutcher: The configurable is `openshift_pkg_version` -- `openshift_version` shouldn't be set from the inventory.

sdodson:
openshift_pkg_version requires leading - but it's preferred to use openshift_version which is used by both containerized and rpm installs.

PR #2108, opened 2016-06-30
Prepare for v1.3

sdodson:
Not ready for merge, also it looks like the image streams and templates are now embedded into the binary. We may want to investigate having the openshift binary create them if that's possible. See https://github.com/openshift/origin/commit/2b11e952d7b01fd20cb1d25f78984e4de41e0960


sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
@sosiouxme FYI, this merged including the latest logging changes and is built into openshift-ansible-3.3.1-1.git.2.57a2f34.el7

detiber:
@sdodson I imagine the embedding of the templates/imagestreams is related to oc cluster up, maybe we need to investigate whether we can leverage that work to handle the management of the templates, imagestreams and possibly some of the other post config steps as well.

PR #2126, opened 2016-07-07
Clean up some deprecation warnings

### LINKED ISSUE #2125: There are many deprecation warnings in a playbook run
I noticed many deprecation warnings like such

> [DEPRECATION WARNING]: Using bare variables is deprecated. Update your playbooks so that the environment value uses the full variable syntax ('{{g_all_hosts}}').

This is a simple fix. PR incoming presently.

tbielawa:
* "Using bare variables is deprecated. Update your playbooks so that
  the environment value uses the full variable syntax"
* Closes #2125

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -121,7 +121,7 @@
                                                     | union(groups['oo_etcd_to_config'] | default([])))
                                                 | oo_collect('openshift.common.hostname') | default([]) | join (',')
                                                 }}"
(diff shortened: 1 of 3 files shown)


tbielawa:
Note that some of those lines changed are just cleaned up trailing whitespace.

detiber:
@sdodson @abutcher: how far off are we from dropping 1.9.4 support?

abutcher:
@detiber we're a few days away at best. These changes look safe with both 1.9.4/2.x, though.

abutcher:
aos-ci-test

abutcher:
Tested with 1.9.4 :+1: lgtm

PR #2131, opened 2016-07-08
Fix up some broken markdown formatting (mostly tables)

tbielawa:


tbielawa:
I was trying to make a unified view of all the various roles README files and I ran into a couple with tables that couldn't be parsed by the md2html tool I was using.

This PR fixes the formatting in those broken READMEs

tbielawa:
aos-ci-test

detiber:
@tbielawa lgtm, but most of those readmes are probably in a very bad state :)

PR #2134, opened 2016-07-08
Ansible role graph script

abutcher:
First take on an ansible role graph script which has been useful for debugging.

Example graphs:
- internal svg http://file.rdu.redhat.com/abutcher/generated-ansible-role-graph.svg
- external png http://i.imgur.com/nWVUnWZ.png

```
./bin/ansible-role-graph --format pdf --engine dot --roles-path ~/openshift-ansible/roles/
```

Depends on #2132

PR #2136, opened 2016-07-09
Colorize role graph

tbielawa:
Depends on @abutcher 's changes in https://github.com/openshift/openshift-ansible/pull/2134


tbielawa:
resubmitting to @abutcher's fork directly

PR #2138, opened 2016-07-11
Enable pullthrough by default in registry config for object storage.

abutcher:


detiber:
This option should probably be toggle-able.

abutcher:
@detiber Sounds good. I'll make this disabled by default with a toggle on var.

detiber:
@abutcher I think having the default being enabled is fine, we just should have a way for someone to disable it.

abutcher:
aos-ci-test

abutcher:
@sdodson ptal

sdodson:
LGTM

PR #2139, opened 2016-07-11
Remove cleanup code from 1.0 to 1.1 upgrade era

sdodson:
There were a batch of templates and image streams that needed to be remove when upgrading from 1.0 to 1.1. Since the current version of the installer only needs to support 1.2 and later we can remove this rather expensive bit of code.


sdodson:
@tbielawa this shall make you happy
@rcernich there weren't any similar removal of 1.1 content necessary when upgrading to 1.2 or later was there?

sdodson:
aos-ci-test

detiber:
:+1:

rcernich:
@sdodson I don't believe so.

PR #2140, opened 2016-07-12
ops-docker-loopback-to-direct-lvm.yml: fix typo on the variable name "cli_name vs cli_host"

gaelL:
ops-docker-loopback-to-direct-lvm.yml: fix typo on the variable name `cli_name vs cli_host`


detiber:
@gaelL Thanks for the fix

PR #2141, opened 2016-07-12
Fix bin/cluster openstack related error

lhuard1A:
It looks like spawning a cluster with `bin/cluster` on OpenStack is suffering from the same issue as the one that @detiber fixed in #1969.
So, I just applied the same fix.


detiber:
@lhuard1A Thanks, I don't currently have a good environment to test bin/cluster against for OpenStack.

PR #2145, opened 2016-07-13
Default sdn_cluster_network_cidr changed to 10.128.0.0/14

jtudelag:
Documentation says the default sdn cluster network cidr in OSE 3.2.X has changed to 10.128.0.0/14.
https://docs.openshift.com/enterprise/3.2/install_config/install/advanced_install.html#configuring-cluster-variables 

There is also a open bug: https://bugzilla.redhat.com/show_bug.cgi?id=1320952

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -909,7 +909,7 @@ def set_sdn_facts_if_unset(facts, system_facts):
 
     if 'master' in facts:
         if 'sdn_cluster_network_cidr' not in facts['master']:
-            facts['master']['sdn_cluster_network_cidr'] = '10.1.0.0/16'
+            facts['master']['sdn_cluster_network_cidr'] = '10.128.0.0/14'
         if 'sdn_host_subnet_length' not in facts['master']:
             facts['master']['sdn_host_subnet_length'] = '8'
 


detiber:
@abutcher ptal, we will need to handle a migration, to avoid issues where an existing install doesn't get the new default value.

abutcher:
@detiber Yeah, I was thinking about how we could do this since we aren't persisting the value. It'll be funky.

detiber:
@abutcher might be a good spot for the new facts logic we've talked about...  If the value appears on disk in the config (as returned from openshift_facts), use it. Otherwise, set the default in the template.

We would need to add comments to avoid confusion in the different behavior, or possibly even introduce a separate facts module to handle the different logic...

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -909,7 +909,7 @@ def set_sdn_facts_if_unset(facts, system_facts):
 
     if 'master' in facts:
         if 'sdn_cluster_network_cidr' not in facts['master']:
-            facts['master']['sdn_cluster_network_cidr'] = '10.1.0.0/16'
+            facts['master']['sdn_cluster_network_cidr'] = '10.128.0.0/14'
         if 'sdn_host_subnet_length' not in facts['master']:
             facts['master']['sdn_host_subnet_length'] = '8'
```
danwinship: This needs to be updated as well (from 8 to 9)

danwinship:
I think this PR got forgotten about? The openshift docs (as of 3.2) claim that the default clusterNetworkCIDR is 10.128.0.0/14 and the default hostSubnetLength is 9, but openshift-ansible is still overriding that to the old values. (https://github.com/openshift/openshift-docs/issues/1700#issuecomment-252794842)

abutcher:
@danwinship Thanks for the reminder. The problem here is that we need a way to ensure that existing clusters don't get updated `sdn_cluster_network_cidr` on re-runs with these new defaults.

danwinship:
Can't ansible just leave the field blank unless the user is overriding it?

Alternatively, you could get the current value from the "default" ClusterNetwork record

detiber:
@danwinship If we left the field blank and a user updated from a version that had the old default to a version that had the new default, would they not be in a situation where they have incompatible network settings for the SDN?

danwinship:
ah, yes

detiber:
@abutcher considering the potential issues, we probably need to get a card added to the backlog to provide a playbook that would allow for updating either the service network cidr or the sdn network cidr/host length.

abutcher:
Updated in https://github.com/openshift/openshift-ansible/pull/2652. Closing.

PR #2146, opened 2016-07-13
Fix erroneous pylint error

smunilla:
pylint is throwing up a no-value-for-parameter error here. Testing shows there is no
real issue, so adding that to the ignores.


abutcher:
aos-ci-test

PR #2147, opened 2016-07-13
Make libvirt’s VM use virtio-scsi insteal of virtio-blk

lhuard1A:
The main advantage is that virtio-scsi honors TRIM so that deleting files
and/or docker images inside the VM now frees disk space on the host by
resparsifying the qcow2 file.


sdodson:
Thanks.

PR #2148, opened 2016-07-13
Install bash-completion package for the oc/oadm tools

tbielawa:
I'd like the bash-completion package installed by default on masters after an installation. Adding bash-completion to the default install enables use of the completion scripts already installed by the master packages.

### DIFF
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -27,6 +27,9 @@
   action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-master{{ openshift_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
   when: not openshift.common.is_containerized | bool
 
+- name: Install bash completion for oc tools
+  action: "{{ ansible_pkg_mgr }} name=bash-completion state=present"
+
 - name: Pull master image
   command: >
     docker pull {{ openshift.master.master_image }}{{ ':' + openshift_version if openshift_version is defined and openshift_version != '' else '' }}


abutcher:
👍

roles/openshift_master/tasks/main.yml
```diff
@@ -27,6 +27,9 @@
   action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-master{{ openshift_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
   when: not openshift.common.is_containerized | bool
 
+- name: Install bash completion for oc tools
+  action: "{{ ansible_pkg_mgr }} name=bash-completion state=present"
```
detiber: Should this be pulled in with openshift_cli rather than openshift_master?

tbielawa: I am uncertain. I put it into master because I only ever imagined running
the oc/oadm commands on a master node.

On Wed, Jul 13, 2016 at 1:35 PM, Jason DeTiberus <notifications@github.com>
wrote:

> In roles/openshift_master/tasks/main.yml
> <https://github.com/openshift/openshift-ansible/pull/2148#discussion_r70671770>
> :
>
> > @@ -27,6 +27,9 @@
> >    action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-master{{ openshift_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
> >    when: not openshift.common.is_containerized | bool
> >
> > +- name: Install bash completion for oc tools
> > +  action: "{{ ansible_pkg_mgr }} name=bash-completion state=present"
>
> Should this be pulled in with openshift_cli rather than openshift_master?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/openshift/openshift-ansible/pull/2148/files/192a5c2e770325cc95deb5071b99ed7373968e38#r70671770>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AACBYeTQUwgd_NJJFBMIf_faNmZs_Ih9ks5qVSHvgaJpZM4JLh0m>
> .
>



-- 
Tim Bielawa, ${::title}
Cell: 919.332.6411  | IRC: tbielawa (#openshift)
1BA0 4FAB 4C13 FBA0 A036  4958 AD05 E75E 0333 AE37

detiber: @tbielawa That is currently the case, but the openshift_cli role is responsible for installing the client tools. I would also suggest to add a `when not openshift.common.is_containerized | bool` so that it doesn't attempt to install bash-completion for a containerized install.

sdodson: Sorry for the hastey merge, i pondered master versus openshift_cli but decided master made sense as that's where the kubeconfig was. I missed the containerized check though.

detiber: kubeconfig is in master only because the admin kubeconfig does not exist on a host that isn't a master, it is a convenience that only makes sense on masters...  having bash-completion on client hosts makes sense on all client hosts :)

PR #2153, opened 2016-07-14
Move the bash completion into the cli role. Only add when not contain…

tbielawa:
…erized

* Moving the task and adding the `when` statement came from conversation in the original PR: #2148


sdodson:
LGTM, CI is bustificated right now so merging manually.

PR #2156, opened 2016-07-15
Two small upgrade related fixes

dgoodwin:
Quick install 3.2 is pointing to wrong location for the playbooks, and adjusting an error message at QE's request.


sdodson:
aos-ci-extended-test

PR #2159, opened 2016-07-15
Symlink ansible 2.x locations to ansible 1.9 locations

sdodson:
Ansible 2.x has moved the plugin path from /usr/share/ansible_plugins/{filter,lookup}_plugins to /usr/share/plugins/{filter,lookup,...} we will try to accommodate both versions.


sdodson:
@abutcher ptal

abutcher:
lgtm

sdodson:
@tdawson Mind giving this a review? I originally thought about trying to symlink the old path to the new path but that seemed to provide no value over symlinking the new path to the old path but added the hassle of writing the pretrans lua scripts.

tdawson:
Looks good.  I like it this way.

detiber:
:+1:

PR #2160, opened 2016-07-15
Convert openshift_release and openshift_version to strings for starts…

sdodson:
…with

Fixes : startswith first arg must be str, unicode, or tuple, not float
```
TASK [openshift_version : fail] ************************************************
fatal: [ose3-master.example.com]: FAILED! => {"failed": true, "msg": "The conditional check 'not is_containerized | bool and openshift_release is defined and not openshift_version.startswith(openshift_release) | bool' failed. The error was: Unexpected templating type error occurred on ({% if not is_containerized | bool and openshift_release is defined and not openshift_version.startswith(openshift_release) | bool %} True {% else %} False {% endif %}): startswith first arg must be str, unicode, or tuple, not float\n\nThe error appears to have been in '/home/rdu/sdodson/git/Openshift/openshift-ansible/roles/openshift_version/tasks/main.yml': line 77, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# the rpm version we looked up matches the release requested and error out if not.\n- fail:\n  ^ here\n"}
```

### DIFF
--- a/roles/openshift_version/tasks/main.yml
+++ b/roles/openshift_version/tasks/main.yml
@@ -17,6 +17,10 @@
   when: openshift_release is defined and openshift_release[0] == 'v'
 
 - set_fact:
+    openshift_release: "{{ openshift_release | string }}"
+  when: openshift_release is defined
+
+- set_fact:
     openshift_image_tag: "{{ 'v' + openshift_image_tag }}"
   when: openshift_image_tag is defined and openshift_image_tag[0] != 'v'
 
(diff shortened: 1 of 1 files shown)


sdodson:
@dgoodwin PTAL

dgoodwin:
Shouldn't be required for openshift_version afaik as that should only be set programatically in the openshift_version role. (though it's harmless to keep)

I hit this when I used openshift_release=3.2 as well and couldn't figure out how to convert to a string, so I set it up to allow either v3.2 or "3.2" if you quoted it in inventory. Seeing how to convert though looks much better, but could we do a blanket set_fact on openshift_release to openshift_release.string, in the openshift_version role, if openshift_release is defined. Then we know from then on it's always going to be a string if it's defined and can use it anywhere as such.

If we keep the openshift_version.string we should do the same for that somewhere too.

sdodson:
look better now?

sdodson:
aos-ci-test

PR #2166, opened 2016-07-17
Check for existence of sebooleans prior to setting.

### LINKED ISSUE #2161: virt_sandbox_use_nfs boolean may not be available, breaks install
The virt_sandbox_use_nfs has been removed from docker-selinux and was to be moved to the selinux-policy package, however, it appears it may now be deprecated and just virt_use_nfs is used instead. 

See https://github.com/fedora-cloud/docker-selinux/issues/18

The `openshift-ansible/roles/openshift_node/tasks/storage_plugins/nfs.yml` playbook assumes that this boolean is available, and if not, the install will fail with an error. Presumably, there should be a check for the boolean with it being set if, and only if, it is present.

abutcher:
Closes #2161


sdodson:
LGTM

sdodson:
aos-ci-test

PR #2168, opened 2016-07-18
Copy openshift binary instead of using wrapper script.

dgoodwin:
For containerized masters, we previously create a wrapper script to run
the docker CLI image and clean up afterward, but this approach is much
slower than having the binary on the native system.

Instead whenever the config playbook is run we copy the binary out of the
CLI image onto the host itself where it can use the kube config already
being set up. Commands now run instantly on the host.

Substantial speed improvement for containerized installs which requires
many openshift/oc commands. If images are pre-pulled on the system, a single master and single node in vms on my local machine installs in 9 minutes vs 18 previously.


roles/openshift_cli/tasks/main.yml
```diff
@@ -8,11 +8,12 @@
...
-  template:
-    src: openshift.j2
-    dest: /usr/local/bin/openshift
-    mode: 0755
+- set_fact:
+    container_name: "openshift-cli-{{ 100000 | random }}"
+  when: openshift.common.is_containerized | bool
+
+- name: Copy openshift binary out of CLI image for use on the host
+  shell: docker create --name {{ container_name }} {{ openshift.common.cli_image }}:{{ openshift_image_tag }} && docker cp {{ container_name }}:/usr/bin/openshift /usr/local/bin/openshift && docker rm {{ container_name }}
```
detiber: Isn't oc a separate binary from openshift now? If we can get away with just copying openshift and using that for both oc/oadm I'm all for it, though.

detiber: Also, do we want to check for the existence of the binary and the version of the binary to determine if we need to update the client? I suspect this is getting quite a bit more complex, but for ongoing config it would seem bad to just consistently overwrite the binary.

dgoodwin: Hmm looking at how that worked it's possible oc was a different binary. I will update to copy each binary separately instead of symlinking, the old wrapper looks like it knew how to handle that better.

Is it the performance (time to copy binaries) you're concerned about?

detiber: @dgoodwin it's less about performance and more about the idempotency of the role.

dgoodwin: @detiber I am not following how it violates idempotency, you can re-run and the end state will be consistent each time provided the image version hasn't changed. If it has then we sync it up, but that sounds desired from a config management POV. Let me know if there's a specific concern not addressed.

detiber: Right now we will mark this step as changed at ever invocation of this role rather than just when the binary changes.

dgoodwin:
oc indeed does appear to be a different binary. kubectl should be a symlink to is, and oadm should be a symlink to openshift. Updated to accommodate all of these, end result is:

```
[root@m1 bin]# ll /usr/local/bin
total 150448
lrwxrwxrwx. 1 root root         2 May 16 16:53 kubectl -> oc
lrwxrwxrwx. 1 root root        24 Jul 18 13:53 oadm -> /usr/local/bin/openshift
-rwxr-xr-x. 1 root root  46489767 May 16 12:18 oc
-rwxr-xr-x. 1 root root 107558985 May 16 12:19 openshift
```

roles/openshift_cli/tasks/main.yml
dgoodwin: Aha! Ok I follow, will think on this and come up with something.

dgoodwin:
@detiber how do you feel about a Python script for this? It would 

- create the temp container
- create a temp dir on host
- copy binaries to temp dir on host
- compare checksums with host, check what's a symlink and what isn't
- copy files or create symlinks as appropriate
- return some output if any change was made
- cleanup the container

The logic is a bit to gnarly for Ansible IMO, as oc is a symlink in 3.2, but a separate binary now, and yet I don't really want to abandon the change as it would save a lot of time for us especially.

sdodson:
> The logic is a bit to gnarly for Ansible IMO, as oc is a symlink in 3.2, but a separate binary now, and yet I don't really want to abandon the change as it would save a lot of time for us especially.

oc is a separate binary in 3.2 too, it's just that on atomic oc is a symlink to the wrapper script.

detiber:
@dgoodwin sounds like you are proposing a custom ansible module to manage the binary, and I am good with that.

roles/openshift_cli/tasks/main.yml
```diff
@@ -8,24 +8,35 @@
...
   when: openshift.common.is_containerized | bool
 
+- name: Create kubectl client symlink
+  file:
+    path: /usr/local/bin/kubectl
+    state: link
+    src: /usr/local/bin/oc
+  when: openshift.common.is_containerized | bool
+
+- fail:
```
tbielawa: This `fail` is intended to be taken out later right? I only noticed it because I was testing this PR and it 'unexpectedly failed' :)


```
TASK [openshift_cli : fail] ****************************************************
task path: /home/tbielawa/rhat/os/openshift-ansible-test/roles/openshift_cli/tasks/main.yml:38
Tuesday 19 July 2016  12:08:03 -0400 (0:00:02.165)       0:04:08.472 **********
fatal: [54.....149]: FAILED! => {"changed": false, "failed": true, "msg": "Failed as requested from task"}
```

dgoodwin: Whoops, yes you are correct :)

dgoodwin:
Ok updated to use an ansible module to handle the tricky logic of getting the binaries and possibly symlinks out and set up, and correctly reporting changed or not. 

Also modified all uses of docker pull to correctly report changed or not.

dgoodwin:
aos-ci-test-extended

roles/openshift_cli/library/container_binary_sync.py
```diff
@@ -0,0 +1,129 @@
...
+class BinarySyncError(Exception):
+    def __init__(self, msg, cmd, exit_code, output):
+        super(BinarySyncError, self).__init__(msg)
+        self.msg = msg
+        self.cmd = cmd
+        self.exit_code = exit_code
+        self.output = output
+
+
+def run_cmd(args):
```
detiber: There is actually a helper method in ansible module_utils for this already: https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1951

roles/openshift_cli/library/container_binary_sync.py
```diff
@@ -0,0 +1,129 @@
...
+        self.changed = False
+        self.output = []
+        self.bin_dir = '/usr/local/bin'
+        self.image = image
+        self.tag = tag
+        self.temp_dir = None # TBD
+
+    def sync(self):
+        container_name = "openshift-cli-%s" % random.randint(1, 100000)
+        self.output.append(run_cmd(['docker', 'create', '--name',
```
detiber: just wondering if it would make sense to use docker-py as opposed to wrapping the command line?

roles/openshift_cli/library/container_binary_sync.py
```diff
@@ -0,0 +1,129 @@
...
+            shutil.move(src_path, dest_path)
+            self.output.append("Moved %s to %s." % (src_path, dest_path))
+            self.changed = True
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            image=dict(required=True),
+            tag=dict(required=True),
```
detiber: Would it make sense to make this module a bit more generic? Allowing the user to specify which binaries to copy out and which symlinks to create to each binary?

Maybe a bit overkill, but seems like outside of the specific binaries and symlinks everything else is pretty generic.

roles/openshift_cli/library/container_binary_sync.py
dgoodwin: This would imply installing a new dependency on each system? If so I'd be inclined to just call the command, but if there's some guarantee it's already there then that sounds good to me. What do you think?

roles/openshift_cli/library/container_binary_sync.py
dgoodwin: I considered this but ruled it out because of the mess around "oc" where it might be a binary or it might be a symlink, and the fact that some of the symlinks come out with wonky full paths to /usr/bin instead of just relative links to same dir. 

However it might be possible, I would have to make sure it remains just one call as we do not want to make and delete this container 4x, it's quite expensive. I will poke around.

dgoodwin: I could do 

```
image_copy:
  image: X
  tag: Y
  dest_dir: /usr/local/bin
  files: ['/usr/local/bin/openshift', '/usr/local/bin/oc']

But the module has to assume that if oc is a symlink, we get the linked files basename (which would be openshift) and link to that in the dest_dir.

oadm and kubectl could be the same, or go back to being standard ansible symlink tasks like they were before.

File: roles/openshift_cli/library/container_binary_sync.py (hunk 4 above)
Comment (detiber) [inline roles/openshift_cli/library/container_binary_sync.py]: Good point, but if we plan to leverage the docker modules for ensuring images for containerized at some point, we will need the dependency anyway.

Adding the dep now or then doesn't really make a big difference to me.

File: roles/openshift_cli/library/container_binary_sync.py (hunk 5 above)
Comment (detiber) [inline roles/openshift_cli/library/container_binary_sync.py]: good point, I say just update the naming so it is obvious that it is OpenShift specific.

File: roles/openshift_cli/library/container_binary_sync.py (hunk 4 above)
Comment (dgoodwin) [inline roles/openshift_cli/library/container_binary_sync.py]: Would also need to run on Atomic where we can't easily install an rpm dep, not positive but I doubt it's installed by default there. Will skip the dep for now.

File: roles/openshift_cli/library/container_binary_sync.py (hunk 5 above)
Comment (dgoodwin) [inline roles/openshift_cli/library/container_binary_sync.py]: Will do.

**Comment by dgoodwin** [comment]
Updated to use ansible run_command, renamed module to include openshift, caught a bug checking the checksums.

File: roles/openshift_cli/library/openshift_container_binary_sync.py [hunk 6]
```diff
@@ -0,0 +1,131 @@
...
+import shutil
+import os.path
+
+# pylint: disable=redefined-builtin,wildcard-import,unused-wildcard-import
+from ansible.module_utils.basic import *
+
+
+DOCUMENTATION = '''
+---
+module: container_binary_sync
```
detiber: openshift_container_binary_sync

dgoodwin: Oops, fixed now.

detiber:
aos-ci-test

detiber:
lgtm

roles/openshift_cli/tasks/main.yml
```diff
@@ -6,24 +6,14 @@
...
-    src: /usr/local/bin/openshift
-  with_items:
-    - /usr/local/bin/oadm
-    - /usr/local/bin/oc
-    - /usr/local/bin/kubectl
+- name: Copy client binaries/symlinks out of CLI image for use on the host
+  openshift_container_binary_sync:
+    image: "{{ openshift.common.cli_image }}"
+    tag: "{{ openshift_image_tag }}"
   when: openshift.common.is_containerized | bool
```
sdodson: Maybe in the future only do this when pull_result | changed or the paths we care about don't exist. I don't think we should do that now though, it may not be worth it.

sdodson:
LGTM, thanks for looking into this.

PR #2170, opened 2016-07-18
a-o-i: Looser facts requirements for unattended

smunilla:
Loosen the facts requirements for unattended installs to allow the user to install
with only 'connect_to' defined.

### DIFF
--- a/utils/src/ooinstall/cli_installer.py
+++ b/utils/src/ooinstall/cli_installer.py
@@ -512,13 +512,6 @@ def error_if_missing_info(oo_cfg):
         sys.exit(1)
     oo_cfg.settings['variant_version'] = version.name
 
-    missing_facts = oo_cfg.calc_missing_facts()
-    if len(missing_facts) > 0:
-        missing_info = True
-        click.echo('For unattended installs, facts must be provided for all masters/nodes:')
-        for host in missing_facts:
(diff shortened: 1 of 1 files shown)


utils/src/ooinstall/cli_installer.py
```diff
@@ -512,13 +512,6 @@ def error_if_missing_info(oo_cfg):
         sys.exit(1)
     oo_cfg.settings['variant_version'] = version.name
 
-    missing_facts = oo_cfg.calc_missing_facts()
-    if len(missing_facts) > 0:
-        missing_info = True
-        click.echo('For unattended installs, facts must be provided for all masters/nodes:')
-        for host in missing_facts:
-            click.echo('Host "%s" missing facts: %s' % (host, ", ".join(missing_facts[host])))
-
```
dgoodwin: Are we validating connect_to is set in unattended install?

smunilla: Interestingly, we weren't ever checking connect_to for unattended installs. I suppose we should though, I'll update.

smunilla: Actually, looking at it closer, a host object can't be created without connect_to set. So I think we're good here.

smunilla:
aos-ci-test

PR #2173, opened 2016-07-18
Resolve some deprecation warnings.

abutcher:


abutcher:
aos-ci-test

abutcher:
The only remaining deprecation warning should be this one -- which will be ugly to resolve.

```
[DEPRECATION WARNING]: Skipping task due to undefined Error, in the future this will be a fatal error.: 'dict object' has no attribute 'cloudprovider'.
This feature will be removed in a future 
release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

roles/openshift_hosted/tasks/router/router.yml
```diff
@@ -9,10 +9,15 @@
...
+  # defined and has all keys (certfile, keyfile, cafile) which we
+  # check above.
   with_items:
-  - "{{ openshift_hosted_router_certificate.certfile }}"
-  - "{{ openshift_hosted_router_certificate.keyfile }}"
-  - "{{ openshift_hosted_router_certificate.cafile }}"
+  - "{{ (openshift_hosted_router_certificate | default({'certfile':none})).certfile }}"
+  - "{{ (openshift_hosted_router_certificate | default({'keyfile':none})).keyfile }}"
+  - "{{ (openshift_hosted_router_certificate | default({'cafile':none})).cafile }}"
   when: openshift_hosted_router_certificate is defined
```
detiber: heh, this is hilarious that the when clause doesn't insulate us from this error.

detiber:
lgtm

roles/openshift_hosted/tasks/router/router.yml
abutcher: Right!? It's gonna be bad when these are fatal.

detiber: Can you check to see if there is an upstream issue for this? I suspect this should be considered a bug.

abutcher: I remember seeing something nearly identical to what we're doing. I'll follow up with it.

PR #2174, opened 2016-07-19
Skip docker upgrades on Atomic.

### LINKED ISSUE #2172: upgrade playbook fails on atomic hosts
There is no repoquery on atomic hosts, so `playbooks/common/openshift-cluster/upgrades/docker/upgrade_check.yml` will fail.

If the rpm_q module #963 is merged, at least checking the current version won't need repoquery anymore which will help a bit.

I can see two ways to check for docker upgrades on atomic hosts, both are pretty invasive.

1. Also do the atomic host upgrade:
 * Check whether there is a new upgrade available,
 * Check docker versions (current and in the next atomic release)
 * proceed as usual

2. Install the new docker rpm on top of the atomic release using rpm-ostree pkg-add.

Both will need very thorough documentation as well!

dgoodwin:
Fixes #2172


dgoodwin:
Tested in AWS with RHEL Atomic, 3.2 upgrade succeeds now.

Issue #2182, opened 2016-07-20
ansible playbook play fails


ekuric:
with latest git pull of this repository, ansible play fails with below error message. narrowed to 
openshift-ansible/playbooks/byo/openshift-cluster/cluster_hosts.yml and ```g_all_hosts:``` section in it
https://github.com/openshift/openshift-ansible/blob/master/playbooks/byo/openshift-cluster/cluster_hosts.yml#L16 

```

# ansible-playbook -i playbooknodes_srijeda.yml openshift-ansible/playbooks/byo/config.yml 
 [WARNING]: It is unnecessary to use '{{' in conditionals, leave variables in
loop expressions bare.

PLAY [localhost] ************************************************************** 

TASK: [include_vars ../../byo/openshift-cluster/cluster_hosts.yml] ************ 
ok: [localhost]

TASK: [add_host ] ************************************************************* 
fatal: [localhost] => Failed to template {{ g_all_hosts | default([]) }}: Failed to template {{ g_master_hosts | union(g_node_hosts) | union(g_etcd_hosts) | union(g_lb_hosts) | union(g_nfs_hosts) | union(g_new_node_hosts)| union(g_new_master_hosts) | default([]) }}: an unexpected type error occurred. Error was unsupported operand type(s) for +: 'set' and 'unicode'

FATAL: all hosts have already failed -- aborting

PLAY RECAP ******************************************************************** 
           to retry, use: --limit @/root/config.retry

localhost                  : ok=1    changed=0    unreachable=0    failed=1   

```

abutcher:
Hey @ekuric, we recently updated the master branch to require ansible-2.1. I made some changes to silence 2.1 deprecation warnings but they are preventing the playbooks from reaching the ansible version check / error message because they are fatal with ansible-1.9.x.

https://github.com/openshift/openshift-ansible/pull/2184 should allow the playbooks to reach the version check.

ekuric:
@abutcher I tried this with ansible 2.x. ( had before this error related to checking ansible version, which I overcome with updating ansible )

abutcher:
With the linked change I hit the version check error message when using 1.9.6 and 2.0.2.0 but install is successful with 2.1.0.0

ekuric:
hi @abutcher ok, I will get latest ansible version on my box, and test it thank you

PR #2184, opened 2016-07-20
Bring back g_all_hosts deprecation warnings.

abutcher:
Resolving these deprecation warnings prevents ansible 1.9.x users from seeing the ansible version error message.

Fixes https://github.com/openshift/openshift-ansible/issues/2182


detiber:
@abutcher do we need to put the ansible check in the playbooks?

abutcher:
@detiber I like that idea..

PR #2185, opened 2016-07-20
a-o-i: Write Role variable groups

smunilla:
Correct an error in the way we're writing out the variable group tags
for host roles.


smunilla:
aos-ci-test

sdodson:
LGTM

PR #2187, opened 2016-07-20
Fix libvirt provider for Ansible 2.1.0.0

lhuard1A:
* The change of Ansible API requires to update the `oo_option` lookup plugin;
* Some variables now need to be enclosed inside `{{…}}`.

### DIFF
--- a/lookup_plugins/oo_option.py
+++ b/lookup_plugins/oo_option.py
@@ -33,15 +33,6 @@ except ImportError:
             def get_basedir(self, variables):
                 return self.basedir
 
-# pylint: disable=no-name-in-module,import-error
-try:
-    # ansible-2.0
-    from ansible import template
-except ImportError:
-    # ansible 1.9.x
-    from ansible.utils import template
-
-
 # Reason: disable too-few-public-methods because the `run` method is the only
 #     one required by the Ansible API
 # Status: permanently disabled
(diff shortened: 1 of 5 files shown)


detiber:
@lhuard1A This will most likely break compatibility with Ansible 1.9.4. If you are okay with that, then so am I.

There also appears to be a pylint issue: W: 39, 4: Unused template imported from ansible (unused-import)

lhuard1A:
Hello @detiber ,

The extra `{{…}}` around variables shouldn’t hurt for Ansible 1.9, but the change in the `oo_option` lookup plugin is indeed not backward compatible.
Making the `oo_option` work for both Ansible 1.9 and Ansible 2.1 shouldn’t be that difficult.
The reason why I didn’t make the effort is because I had the feeling that `openshift-ansible` was already checking that it is run under Ansible 2.1+ (See [roles/openshift_facts/tasks/main.yml:2](https://github.com/abutcher/openshift-ansible/blob/856e47e3a0dc21d680c0e6289c2c16575fd64bee/roles/openshift_facts/tasks/main.yml#L2))
So, making this specific piece backward compatible would make the code uselessly more complex because, anyhow, there is an explicit check that forbids execution with Ansible 1.9.

But if you think there is, nevertheless, an added value in making `oo_option` work for both Ansible 1.9 and 2.1, I can make it. What do you think?

detiber:
@lhuard1A no objections from me for keeping Ansible 2.1+, just wanted to make sure it was known. If you can remove the template import (since it is no longer used), I can go ahead and merge.

lhuard1A:
@detiber Useless template import removed.
Thanks.

PR #2192, opened 2016-07-21
Bug 1358723 - openshift_hosted_router_replicas option didn't work

abutcher:
Default list of nodes matching selector when not retrieved.

@detiber PTAL. ~~I also noticed that we will create a router/registry with replicas=0 when there are no infrastructure nodes. Should we base creation on replicas >= 1?~~

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1358723

### DIFF
--- a/roles/openshift_hosted/tasks/registry/registry.yml
+++ b/roles/openshift_hosted/tasks/registry/registry.yml
@@ -9,7 +9,7 @@
   when: openshift.hosted.registry.replicas | default(none) is none
 
 - set_fact:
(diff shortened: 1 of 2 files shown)


detiber:
@abutcher yes, we should probably default to 1 when no infra nodes are defined or if the selector is not defined.

abutcher:
@detiber I take it back -- there is a conditional in the creation template.

abutcher:
aos-ci-test

PR #2194, opened 2016-07-21
a-o-i: Persist Roles Variables

smunilla:
Previously, we we're saving the roles variables set during a run.


smunilla:
aos-ci-test

PR #2199, opened 2016-07-22
hack to make callback runner_on_ok work with both v1 and v2 api

richm:
This is a hack I use to make oo_install work using ansible 2.1.0 on Fedora 23.

### DIFF
--- a/utils/src/ooinstall/ansible_plugins/facts_callback.py
+++ b/utils/src/ooinstall/ansible_plugins/facts_callback.py
@@ -1,6 +1,6 @@
 # TODO: Temporarily disabled due to importing old code into openshift-ansible
 # repo. We will work on these over time.
-# pylint: disable=bad-continuation,missing-docstring,no-self-use,invalid-name,no-value-for-parameter
+# pylint: disable=bad-continuation,missing-docstring,no-self-use,invalid-name,no-value-for-parameter,protected-access
 
 import os
 import yaml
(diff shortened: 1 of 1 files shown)


sdodson:
@smunilla mind taking a look

utils/src/ooinstall/ansible_plugins/facts_callback.py
```diff
@@ -27,9 +27,19 @@ def on_any(self, *args, **kwargs):
     def runner_on_failed(self, host, res, ignore_errors=False):
         pass
```
smunilla: pylint is unhappy with with why things are being accessed. Until we can get a better fix though, I suggest we just add `#pylint: disable=protected-access` and get this merged in to unblock things.

smunilla:
I just spent the day struggling with what turned out to be exactly what this is addressing. Other than pylint being unhappy, this looks good to me. :+1:

@brenton This looks to be a safer version of the patch we were just talking about, doesn't it?

utils/src/ooinstall/ansible_plugins/facts_callback.py
richm: Well, it is only blocking people who are using ansible 2.1 and/or ansible on Fedora 23, of whom I seem to be the only one . . .

utils/src/ooinstall/ansible_plugins/facts_callback.py
```diff
@@ -27,9 +27,19 @@ def on_any(self, *args, **kwargs):
...
 
-    def runner_on_ok(self, host, res):
-        if res['invocation']['module_args'] == 'var=result':
-            facts = res['var']['result']['ansible_facts']['openshift']
+    def runner_on_ok(self, *args):
```
smunilla: pylint is unhappy with with why things are being accessed. Until we can get a better fix though, I suggest we just add `#pylint: disable=protected-access` and get this merged in to unblock things.

richm: This is a bad hack - there is probably a more elegant way to have the callback class work with both ansible 1.x/2.0/2.1

brenton:
@smunilla, if this works then I'm fine with it.  It's not clear to me when the v1 api will be dropped but please create a card in the 'New' list to officially port this plugging once the old api is removed.

sdodson:
aos-ci-test

abutcher:
Superseded by https://github.com/openshift/openshift-ansible/pull/2266

PR #2201, opened 2016-07-22
add 3.3 to installer

richm:


utils/src/ooinstall/variants.py
```diff
@@ -44,6 +45,7 @@ def latest_version(self):
 
 AEP = Variant('atomic-enterprise', 'Atomic Enterprise Platform',
     [
+        Version('3.3', 'atomic-enterprise'),
```
sdodson: 3.3 doesn't exist for atomic-enterprise

sdodson:
@smunilla FYI

PR #2202, opened 2016-07-23
initial support for v1.3 with logging v1.3

richm:
Support for the new way to deploy logging in origin-aggregated-logging v1.3
* daemonset for fluentd
* configmap for deployer config
* support for using journald log source


sdodson:
Oh wow, thanks, will review next week.

richm:
@sdodson new and improved version

portante:
@richm, @sdodson, what is the state of this work?

richm:
@portante It needs reviews

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -28,21 +28,20 @@
...
     command: >
-      {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig create -f  /tmp/logging-deployer-sa.yaml
-    register: deployer_output
-    failed_when: "deployer_output.rc == 1 and 'exists' not in deployer_output.stderr"
+      {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig create -f /usr/share/openshift/examples/infrastructure-templates/enterprise/logging-deployer.yaml
+    register: template_output
+    failed_when: "template_output.rc == 1 and 'exists' not in template_output.stderr"
+
+  - name: "Process the logging accounts template"
+    shell:  "{{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig process logging-deployer-account-template |  {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig create -f -"
```
enoodle: why not using `new-app` like here https://github.com/openshift/origin-aggregated-logging/tree/master/deployer#create-supporting-service-accounts-and-permissions
for me it fails here because `logging-deployer-account-template` is in `openshift` namespace.

richm: > for me it fails here because logging-deployer-account-template is in openshift namespace

Are you redeploying on top of an earlier deployment?  Because doing a fresh deployment does this:
```
  - name: "Create logging project"
    command: >
      {{ openshift.common.admin_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig new-project logging
    when: logging_project_result.stdout == ""

  - name: "Changing projects"
    command: >
      {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig project logging
...
  - name: "Create templates for logging accounts and the deployer"
    command: >
      {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig create -f /usr/share/openshift/examples/infrastructure-templates/enterprise/logging-deployer.yaml
    register: template_output
    failed_when: "template_output.rc == 1 and 'exists' not in template_output.stderr"
```
AFAICT, this will create the `logging-deployer-account-template` in the `logging` namespace.

jeremyeder:
@rflorenc

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
enoodle: Yes, It seems to be a problem with my installation for not copying the right file there. Thank you.

enoodle:
This works for me. My only problem is that its not easy to run this role with the normal installation, but I guess its not part of this PR's scope.
LGTM :+1:

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -77,9 +76,14 @@
   - name: "Process support template"
     shell:  "{{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig process logging-support-template |  {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig create -f -"
 
+  - name: "Process imagestream template"
+    shell:  "{{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig process logging-imagestream-template -v {{ oc_process_values }} |  {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig create -f -"
```
enoodle: I would suggest adding `-v` to `oc_process_values` and remove it above. When the user doesn't specify image version or prefix it will be empty and this will fail.

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -77,9 +76,14 @@
   - name: "Process support template"
     shell:  "{{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig process logging-support-template |  {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig create -f -"
```
enoodle: IIUC this is not needed now because the deployer runs it. https://github.com/openshift/origin-aggregated-logging/blob/master/deployer/scripts/install.sh#L113

richm:
@enoodle you made an earlier comment:

> This works for me. My only problem is that its not easy to run this role with the normal installation, but I guess its not part of this PR's scope. LGTM

What do you mean "it's not easy to run this role with the normal installation."?

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
sdodson: looks like `oc process KEY1=VALUE1 KEY2=VALUE2` works too so maybe just `{{ ip_kv }} {{ iv_kv }}` and then get rid of oc_process_values all together so we don't have to deal with whether or not to use -v

roles/openshift_hosted_logging/vars/main.yaml
```diff
@@ -1,6 +1,24 @@
-kh_kv: "KIBANA_HOSTNAME={{ openshift_hosted_logging_hostname | quote }}"
-es_cs_kv: "ES_CLUSTER_SIZE={{ openshift_hosted_logging_elasticsearch_cluster_size | string | quote }}"
-pmu_kv: "PUBLIC_MASTER_URL={{ openshift_hosted_logging_master_public_url | quote }}"
 ip_kv: "{{ 'IMAGE_PREFIX=' ~ target_registry | quote if target_registry is defined else '' }}"
```
sdodson: I don't know how widely used this role is so far, we haven't officially supported this yet so I'd prefer if we went ahead and changed `target_registry` to `openshift_hosted_logging_image_prefix` here.

sdodson:
> @enoodle you made an earlier comment:
> 
>     This works for me. My only problem is that its not easy to run this role with the normal installation, but I guess its not part of this PR's scope. LGTM
> 
> What do you mean "it's not easy to run this role with the normal installation."?

I think he just means that no playbooks currently call this role. For metrics the install playbooks include the role when `openshift_hosted_metrics_deploy=true` and it's currently defaulted to false. I'll put together a follow up PR to integrate this into the playbooks and we'll have QE test it then we can decide whether or not to enable it by default.

roles/openshift_hosted_logging/vars/main.yaml
sdodson: @JayKayy thoughts on impact of changing this? we're looking to get this role included in default installs

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
sdodson: According to Clayton `-v KEY1=VALUE1,KEY2=VALUE2` is a bug and will be removed in the future, hopefully not soon as it's currently documented in the --help output, but anyway my proposal should work long term.

roles/openshift_hosted_logging/vars/main.yaml
JayKayy: I honestly dont think many people are using this role at all. This impact would be very small based on my estimations.

richm:
@sdodson @enoodle PTAL

sdodson:
Looks good to me, thanks.

PR #2210, opened 2016-07-25
Stagger HA master service start

abutcher:
@detiber thoughts on an approach like this?


sdodson:
Have we discussed why this is necessary with origin folks? While we can work around it here, if master services cannot be started in parallel that's something that needs to be fixed in the product for situations where an entire environment is restarted outside of our control (power outage, etc).

CC: @smarterclayton

abutcher:
We haven't but this may not be necessary. All of my testing has been good
so far.

sdodson:
aos-ci-test

sdodson:
Confirmed to have resolved problems with at least one cluster install.

sdodson:
aos-ci-test

PR #2211, opened 2016-07-25
1.3 / 3.3 Upgrades

dgoodwin:
Refactored the 3.2 upgrade common files out to a path that does not
indicate they are strictly for 3.2.

3.3 upgrade then becomes a relatively small copy of the byo entry point,
all calling the same code as 3.2 upgrade.

Thus far there are no known 3.3 specific upgrade tasks. In future we
will likely want to allow hooks out to version specific pre/upgrade/post
tasks.

Also fixes a bug where the handlers were not restarting
nodes/openvswitch containers doing upgrades, due to a change in Ansible
2+.


dgoodwin:
aos-ci-test-extended

dgoodwin:
aos-ci-test

dgoodwin:
aos-ci-test

playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
```diff
@@ -61,5 +61,7 @@
 - include: ../../../../common/openshift-cluster/upgrades/upgrade.yml
   vars:
     openshift_deployment_type: "{{ deployment_type }}"
+    master_config_hook: "v3_3/master_config_upgrade.yml"
+    node_config_hook: "v3_3/node_config_upgrade.yml"
```
dgoodwin: This is my "hooks" idea. Unfortunately the path is relative to the common directory where it actually gets evaluated, not byo.

library/modify_yaml.py
```diff
@@ -20,6 +20,23 @@
...
 '''
 
+
+def set_key(yaml_data, yaml_key, yaml_value):
+    changes = []
+    ptr = yaml_data
+    for key in yaml_key.split('.'):
+        if key not in ptr and key != yaml_key.split('.')[-1]:
+            ptr[key] = {}
+            ptr = ptr[key]
```
dgoodwin: This line above was the bugfix, needed to re-assign the ptr after creating a new section.

playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
```diff
@@ -0,0 +1,67 @@
...
+      openshift_upgrade_min: "{{ '1.2' if deployment_type == 'origin' else '3.2' }}"
+
+- include: ../../../../common/openshift-cluster/upgrades/pre.yml
+  vars:
+    openshift_deployment_type: "{{ deployment_type }}"
+- include: ../../../../common/openshift-cluster/upgrades/upgrade.yml
+  vars:
+    openshift_deployment_type: "{{ deployment_type }}"
+    master_config_hook: "v3_3/master_config_upgrade.yml"
+    node_config_hook: "v3_3/node_config_upgrade.yml"
```
dgoodwin: This was my hooks idea, unfortunately these paths are relative to the common upgrade files where they get used, rather than here where defined.

dgoodwin:
Added some additional commits to address issue #2214 and deploy/upgrade protobuf config.

sdodson:
aos-ci-test

sdodson:
LGTM, will test an upgrade or two tomorrow.

sdodson:
Ran through a few upgrades of various configs, all went well, opened a PR against your branch to migrate ca.crt to ca-bundle.crt.

test/modify_yaml_tests.py
```diff
@@ -0,0 +1,37 @@
+""" Tests for the modify_yaml Ansible module. """
+# pylint: disable=missing-docstring,invalid-name
+
+import os
+import sys
+import unittest
+
+sys.path = [os.path.abspath(os.path.dirname(__file__) + "/../library/")] + sys.path
+
+# pylint: disable=import-error,wrong-import-position
```
abutcher: Pylint doesn't like `wrong-import-position` for some reason. Passes locally for me.

dgoodwin: Someone, in their infinite wisdom, decided they should constantly change the pylint keys, and make it fail if it doesn't recognize one. Must not be present in whatever env we're running pylint for CI. I'll remove it.

abutcher:
aos-ci-test

abutcher:
aos-ci-test

PR #2222, opened 2016-07-27
a-o-i: Write missing openshift_node_labels

dkorn:
Currently, the node_labels aren't written to the inventory, which affects different features including the router (requires node label `"{'region': 'infra'}"`).

### DIFF
--- a/utils/src/ooinstall/openshift_ansible.py
+++ b/utils/src/ooinstall/openshift_ansible.py
@@ -205,6 +205,8 @@ def write_host(host, inventory, schedulable=None):
         facts += ' openshift_public_hostname={}'.format(host.public_hostname)
     if host.containerized:
         facts += ' containerized={}'.format(host.containerized)
+    if host.node_labels:
+        facts += ' openshift_node_labels="{}"'.format(host.node_labels)
 
     # Distinguish between three states, no schedulability specified (use default),
     # explicitly set to True, or explicitly set to False:


simon3z:
Thanks @dkorn :+1: cc @detiber @sdodson @smunilla

smunilla:
My original plan was to be able to specify node labels on hosts using a syntax like `node_labels: 'infra'`, which would result in something like `region=infra` getting written to the inventory. However, after talking with @dkorn and thinking about it overnight, we might want to allow labels other than region for futureproofing. This would lead me to want a format more like:
`
  node_labels:
    - region: infra
`

This would require a little more tweaking. If that makes sense to everyone, @dkorn and I can work together to make those changes.

smunilla:
I talked more with @dkorn and I now like this method a lot :) We just have to set `node_labels: {'region': 'infra'}` in the quick installer config. So LGTM :+1:

smunilla:
aos-ci-test

smunilla:
Just for fun, here is the inventory we get out of this when setting `node_labels: {'region': 'infra'}` on a host: 
```
[OSEv3:children]
nodes
nfs
masters

[OSEv3:vars]
ansible_config=/usr/share/atomic-openshift-utils/ansible.cfg
ansible_log_path=/tmp/ansible.log
ansible_ssh_user=openshift
ansible_become=yes

[masters]
192.168.1.x  openshift_node_labels="{'region': 'infra'}"

[nodes]
192.168.1.x  openshift_node_labels="{'region': 'infra'}" openshift_schedulable=True

[nfs]
192.168.1.x  openshift_node_labels="{'region': 'infra'}" 
```

yaacov:
How do you specify different labels to the master node ?

For example, how do you do:
```
[masters]
192.168.1.x openshift_scheduleable=True

# Host group for nodes
[nodes]
192.168.1.x openshift_node_labels="{'region': 'infra', 'zone': 'default'}"
192.168.1.y
```

sdodson:
> How do you specify different labels to the master node ?

node_labels and openshift_schedulable are only relevant to the node roles, however a host that's got multiple roles will merge the variables from all the applied roles. I'm not certain which would take precedence over the other, probably whatever comes last, but in your example 192.168.1.x would have labels and schedulable=True.

yaacov:
@sdodson :+1:  thanks
my question: 
a. Is the `node_labels:` key per node or for all the nodes ?
b. Can I set different labels for different nodes ? (for example, can 192.168.1.1 have different labels than 192.168.1.2  )

sdodson:
@yaacov #2228 added support for per host variables, again i'm not certain of the behavior if you had defined them at multiple levels, I'd avoid setting things that likely vary by host at the global or role level. node_labels is one in particular I'd suspect you want to set per host.

smunilla:
`node_labels` is host right now. You could possibly set a role level variable in a clever way to get labels on all nodes for instance if you really wanted to do that.

smunilla:
#2228 Can definitely allow you to get yourself into trouble. It's very under-the-hood-i-know-what-i-am-doing.

yaacov:
@sdodson @smunilla Thanks :exclamation: I never seen a `.yml ` configuration file, it's a newbie question :flushed:
I wanted to know that I can set the master-node labels, and not set the other nodes.

p.s.
Do you have some basic examples ?

sdodson:
Rebased in #2229

smunilla:
@yaacov Basic examples are in progress. I'm PM what I have so far on mojo.

yaacov:
@smunilla :+1:

PR #2224, opened 2016-07-27
Fix registry/router being created despite no infra nodes.

dgoodwin:
Fix a bug with determining the correct number of replicas to use, and
guarding the creation of the router/registry on whether or not replicas
is > 0.


dgoodwin:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
LGTM

PR #2226, opened 2016-07-27
a-o-i: Move inventory vars to the correct location

smunilla:
Several variables such as 'deployment_type' and 'ansible_config' were
being set under a variable group for the last defined role instead of
under OSEv3:vars.


sdodson:
aos-ci-test

PR #2228, opened 2016-07-27
a-o-i: Support for arbitrary host-level variables

smunilla:
This allows the user to set a block containing any variables they want to set per-host
instead of per-role.


smunilla:
aos-ci-test

sdodson:
LGTM

smunilla:
aos-ci-test

PR #2232, opened 2016-07-28
a-o-i: Set roles on standalone storage

smunilla:
Sets the roles appropriately when using a standalone storage host.

Fixes BZ#1353152


smunilla:
aos-ci-test

PR #2233, opened 2016-07-28
Fix HAProxy logging

sdodson:
- Fix haproxy logging -- probably more do to here to get useful logs but at least this will show us when frontends/backends are initialized

### DIFF
--- a/roles/openshift_loadbalancer/templates/haproxy.cfg.j2
+++ b/roles/openshift_loadbalancer/templates/haproxy.cfg.j2
@@ -7,6 +7,7 @@ global
     user        haproxy
     group       haproxy
     daemon
+    log         /dev/log local0 info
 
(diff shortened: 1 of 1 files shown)


sdodson:
@abutcher you removed the serial stuff, will re-introducing this break your work?

sdodson:
@jeremyeder pretty sure this will fix your deployments

sdodson:
aos-ci-test

abutcher:
@sdodson Re-adding serial is fine but may be slower.

liggitt:
https://github.com/openshift/origin/pull/10099 will make the policy setup more robust, but letting the masters start one at a time would still be ideal (bootstrap stuff will still compete unnecessarily in etcd if all the masters start at once)

PR #2238, opened 2016-07-29
Don't set a networkPluginName in 3.3 installs

### LINKED ISSUE #2066: Do not set networkPluginName in node config when openshift network plugin is used
With this PR https://github.com/openshift/origin/pull/9147, node configuration networkConfig.networkPluginName is optional for openshift network plugin. We recommend users not to set this param in this case. OpenShift node will auto detect network plugin from OpenShift master and this will avoid any plugin mismatches (more details: https://trello.com/c/Lr0wa4hr/228-5-detect-master-and-node-sdn-plugin-mismatch-sdn)

I don't know if we need any ansible changes, please verify and do the needful.
@brenton @sdodson

sdodson:
Fixes #2066 

For now, conditionalize it, remove it entirely in the near future.

### DIFF
--- a/roles/openshift_node/templates/node.yaml.v1.j2
+++ b/roles/openshift_node/templates/node.yaml.v1.j2
@@ -13,14 +13,14 @@ imageConfig:
 kind: NodeConfig
 kubeletArguments: {{ openshift.node.kubelet_args | default(None) | to_padded_yaml(level=1) }}
 masterKubeConfig: system:node:{{ openshift.common.hostname }}.kubeconfig
-{% if openshift.common.use_openshift_sdn %}
+{% if openshift.common.use_openshift_sdn | bool and not openshift.common.version_gte_3_3_or_1_3 | bool %}
 networkPluginName: {{ openshift.common.sdn_network_plugin_name }}
 {% endif %}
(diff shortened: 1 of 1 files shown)


sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
@abutcher ptal i'd like to keep master compat with 3.2 but I think we may've merged other things into master already that aren't so perhaps I should just remove it entirely.

abutcher:
lgtm

roles/openshift_node/templates/node.yaml.v1.j2
```diff
@@ -13,14 +13,14 @@ imageConfig:
 kind: NodeConfig
 kubeletArguments: {{ openshift.node.kubelet_args | default(None) | to_padded_yaml(level=1) }}
 masterKubeConfig: system:node:{{ openshift.common.hostname }}.kubeconfig
-{% if openshift.common.use_openshift_sdn %}
+{% if openshift.common.use_openshift_sdn | bool and not openshift.common.version_gte_3_3_or_1_3 | bool %}
 networkPluginName: {{ openshift.common.sdn_network_plugin_name }}
 {% endif %}
```
detiber: At this point, we could probably remove this block completely.

PR #2243, opened 2016-07-30
a-o-i: Fix broken uninstall

smunilla:
The uninstall method was looking for the host list according to the
old quick-installer config file format. Updated to match the new
arbitrary yaml config format.

Fixes BZ#1359427


smunilla:
aos-ci-test

sdodson:
Will this work if they have an old config file? How are we handling old config file migrations?

smunilla:
Good question… I think for this patch it'd be good to cover both cases. I'll work on that.

sdodson:
aos-ci-test

PR #2259, opened 2016-08-04
Correct openshift ca missing check

abutcher:
No need to delegate and reference from first master. Facts will be available on the host that delegated.

Allows master scaleup playbook to work with the refactored certificate roles.


abutcher:
aos-ci-test

abutcher:
aos-ci-test

PR #2260, opened 2016-08-04
Bug 1347650 - New master isn't added to haproxy.cfg file of LB when scaling masters

abutcher:
https://bugzilla.redhat.com/show_bug.cgi?id=1347650

### DIFF
--- a/playbooks/common/openshift-master/scaleup.yml
+++ b/playbooks/common/openshift-master/scaleup.yml
@@ -58,4 +58,6 @@
 
 - include: ../openshift-master/config.yml
 
+- include: ../openshift-loadbalancer/config.yml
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-master/scaleup.yml
```diff
@@ -56,6 +56,8 @@
   - openshift_facts
   - openshift_docker
 
+- include: ../openshift-loadbalancer/config.yml
+
 - include: ../openshift-master/config.yml
```
sdodson: I assume the load balancer health checks will prevent a bad host from being put into rotation, but should we update the load balancer after the master has been deployed?

Issue #2262, opened 2016-08-07
ansible_connection=local is added automatically by the installer when using hostnames via the quick installer


alongoldboim:
@simon3z @dkorn @smunilla @sdodson 
We are using ssh connection to execute the playbook on remote vm and then disconnecting hence we need the vm to create an ssh connection to itself and supply tty to the process.
Using hostnames in the installer causes it to append ansible_connection=local to the local vm, this behavior should be optional not mandatory.
http://pastebin.com/W0cNth4a (inventories, ansible.log), the only difference in the config files that were used to create those inventories is the  "connect_to: 10.35.48.24" to
 "connect_to: vm-48-24.eng.lab.tlv.redhat.com".

alongoldboim:
@smunilla waiting for alternative fix, this is important to get off our list :) thanks.

smunilla:
@alongoldboim I'm working on it. I just want to make sure all our bases are covered for interactive installs.

detiber:
@alongoldboim So, the main reason that we are updating the connection to be a local connection is that we've had intermittent issues where the ssh connection back to itself is severed during the iptables configuration step. Can you provide info on the error you are seeing when the connection back to itself is local rather than over ssh?

alongoldboim:
@detiber Yea, we are using the openshift-installer in the following way:
```
SSH_AUTH_SOCK=socket nohup sudo -b -E atomic-openshift-installer -u -c /usr/share/ansible/openshift-ansible/inventory.yaml install 1> ANSIBLE_LOG 2> #ANSIBLE_ERROR_LOG < /dev/null
```
When we don't connect through ssh then there is no tty supplied, and then playbook fails.

detiber:
@alongoldboim is it not possible to allocate a ptty when invoking the installer? I can't say I'm familiar with using SSH_AUTH_SOCK in that way

alongoldboim:
@detiber any suggestions on how I can allocate it?

detiber:
@alongoldboim not using SSH_AUTH_SOCK, I've never seen using that env var for executing a binary instead of just passing host auth through to the client.

Assuming you are using net-ssh to make the remote connection, it is possible to call the request_pty method on the Channel, more info is available here: http://net-ssh.github.io/net-ssh/Net/SSH/Connection/Channel.html

alongoldboim:
[...]

PR #2263, opened 2016-08-07
removing auto ansible_connection = local attr from ooinstaller

alongoldboim:
@smunilla @sdodson 
suggested fix for:
https://github.com/openshift/openshift-ansible/issues/2262
We will loose the few optimizations because we connecting via ssh to the same machine, but if a user will specify it it will still work.


smunilla:
We can't really simply cut this out of the quick installer. I can see cases where the user might not want this, but certainly as the default behavior in interactive mode I'd want this added. 

A better solution might be to write out local_connection as a host-level variable in the installer config and then use that to determine what to write in the inventory. This would allow for more advanced users to have a choice.

alongoldboim:
@smunilla That's the other option i was thinking of, I'm good with that also, just need to add the ability to ignore that auto addition of local.
I'll close this pr.
But we need the fix asap.

PR #2264, opened 2016-08-07
Fix GCE Launch

bdurrow:
The Ansible GCE module (documentation here: http://docs.ansible.com/ansible/gce_module.html) requires a comma separated list when you pass an array here (even with a single element) the argument has square brackets around it and the instance doesn't get launched.  Testing shows that joining with ', ' (comma space) works with one instance but breaks with two so I used ',' (comma no space).

### DIFF
--- a/playbooks/gce/openshift-cluster/tasks/launch_instances.yml
+++ b/playbooks/gce/openshift-cluster/tasks/launch_instances.yml
@@ -1,7 +1,7 @@
 ---
 - name: Launch instance(s)
   gce:
-    instance_names: "{{ instances }}"
+    instance_names: "{{ instances|join(',') }}"
     machine_type: "{{ gce_machine_type | default(deployment_vars[deployment_type].machine_type, true) }}"
     image: "{{ gce_machine_image | default(deployment_vars[deployment_type].image, true) }}"
     service_account_email: "{{ lookup('env', 'gce_service_account_email_address') }}"


sdodson:
@menren mind taking a look at this?

menren:
Hello,

I do not have my workstation as I am on holiday, can't check now.
As far as I know it was working well ...

According to this [gce.py](https://github.com/ansible/ansible-modules-core/blob/devel/cloud/google/gce.py#L597) we can either pass a list or comma separated string for instance_names.

For instance  `master_names` seems is created in [playbooks/common/openshift-cluster/tasks/set_master_launch_facts.yme](https://github.com/openshift/openshift-ansible/blob/5a0a35b7d3ece8c1c4b60c12d4ddbfe2a3c070a2/playbooks/common/openshift-cluster/tasks/set_master_launch_facts.yml#L11). It should be recognized as a list by gce.py

@bdurrow With which type of instances did you get error ?

bdurrow:
```
bst:tmp brad$ git clone git@github.com:openshift/openshift-ansible.git --depth=1  --single-branch
Cloning into 'openshift-ansible'...
remote: Counting objects: 1142, done.
remote: Compressing objects: 100% (716/716), done.
remote: Total 1142 (delta 313), reused 718 (delta 190), pack-reused 0
Receiving objects: 100% (1142/1142), 545.48 KiB | 0 bytes/s, done.
Resolving deltas: 100% (313/313), done.
Checking connectivity... done.
bst:tmp brad$ cd openshift-ansible/
bst:openshift-ansible brad$ bin/cluster create gce -t origin test04
================================================================================
ATTENTION: You are running a community supported utility that has not been
tested by Red Hat. Visit https://docs.openshift.com for supported installation
instructions.
================================================================================


PLAY [Launch instance(s)] ******************************************************

TASK [fail] ********************************************************************
skipping: [localhost]

TASK [set_fact] ****************************************************************
ok: [localhost]

TASK [Generate etcd instance names(s)] *****************************************

TASK [set_fact] ****************************************************************
ok: [localhost]

TASK [Launch instance(s)] ******************************************************
skipping: [localhost]

TASK [set_fact] ****************************************************************
skipping: [localhost]

TASK [set_fact] ****************************************************************
skipping: [localhost]

TASK [Add new instances to groups and set variables needed] ********************
[DEPRECATION WARNING]: Using bare variables is deprecated. Update your playbooks
 so that the environment value uses the full variable syntax
('{{gce.instance_data | default([], true)}}').
This feature will be removed in a
 future release. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.

TASK [Wait for ssh] ************************************************************
[DEPRECATION WARNING]: Using bare variables is deprecated. Update your playbooks
 so that the environment value uses the full variable syntax
('{{gce.instance_data | default([], true)}}').
This feature will be removed in a
 future release. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.

TASK [Wait for user setup] *****************************************************
[DEPRECATION WARNING]: Using bare variables is deprecated. Update your playbooks
 so that the environment value uses the full variable syntax
('{{gce.instance_data | default([], true)}}').
This feature will be removed in a
 future release. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.

TASK [set_fact] ****************************************************************
ok: [localhost]

TASK [Generate master instance names(s)] ***************************************
ok: [localhost] => (item=1)

TASK [set_fact] ****************************************************************
ok: [localhost]

TASK [Launch instance(s)] ******************************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Unexpected error attempting to create instance ['test04-master-c89d3'], error: {'domain': 'global', 'message': \"Invalid value for field 'resource.name': '['test04-master-c89d3']'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)'\", 'reason': 'invalid'}"}

	to retry, use: --limit @playbooks/gce/openshift-cluster/launch.retry

localhost                  : ok=5    changed=0    unreachable=0    failed=1

ACTION [create] failed: Command 'ansible-playbook  -i inventory/gce/hosts -e 'num_masters=1 num_nodes=2 cluster_id=test04 cluster_env=dev num_etcd=0 num_infra=1 deployment_type=origin' playbooks/gce/openshift-cluster/launch.yml' returned non-zero exit status 1
```

menren:
This is caused by a bug in ansible 2.1 that converts complex types to str when passed to module : https://github.com/ansible/ansible/issues/16057

This bug would impact all modules which try to handle complex types.

So for now we can pass all instances names as string to gce module as @bdurrow did.
I would add a mention to this bug in the code next to the fix

:+1:

detiber:
@abutcher what are your thoughts on this? Does this behavior still exist with the 2.2 pre-release build we are using?

bdurrow:
While the issue linked by @menren may be relevant, the documentation for the GCE module specifically calls for a comma separated string.  I don't think that we should have any expectation that it would work with a list.  I know the code implies a list will work but using an undocumented feature is unlikely to have good long term outcome.

PR #2265, opened 2016-08-08
Fix little mistake in openshift_master_htpasswd_users value .

jmferrer:
Fix little mistake in openshift_master_htpasswd_users value .


sdodson:
Thanks!

PR #2268, opened 2016-08-08
Ansible 2.2 Support

abutcher:


roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1599,11 +1604,13 @@ def __init__(self, role, filename, local_facts,
...
+            # pylint: disable=too-many-function-args,invalid-name
             self.system_facts = ansible_facts(module, ['hardware', 'network', 'virtual', 'facter'])
-        except TypeError:
-            # ansible-1.9.x,ansible-2.0.x
-            self.system_facts = ansible_facts(module)
+            for (k, v) in self.system_facts.items():
+                self.system_facts["ansible_%s" % k.replace('-', '_')] = v
+        except UnboundLocalError:
+            # ansible-2.2
+            self.system_facts = get_all_facts(module)['ansible_facts']
```
abutcher: @smunilla ^ what I was blathering about earlier regarding "ansible_" prefixing on system facts.

abutcher:
aos-ci-test

roles/openshift_facts/library/openshift_facts.py
smunilla: @abutcher Looking through them, I'm still pretty sure the quick installer doesn't really use any of the ones that changed

abutcher: Tested with unattended all in one and it looks alright.

PR #2269, opened 2016-08-08
Set become=no for etcd server certificates temporary directory.

abutcher:


sdodson:
LGTM, CI jobs won't touch this. Though we should add an item to do CI jobs with etcd hosts.

sdodson:
https://trello.com/c/z5UHFbmd/271-ci-jobs-should-deploy-ha-configs

fahmifahim:
Hello, 
I got almost the same error as yours. 
In my case, the error happened on the task of "Retrieve etcd cert tarball". 
Error message: Unable to create local directories(/tmp/{Master-hostname}/etc/etcd/generated_certs):[Errno 13] Permission denied /tmp/{Master-hostname}

I executed the deploy-cluster.yaml playbook on my Bastion/Boot node. (This node is separated from other OpenShift cluster node). 
I checked the permission of my OpenShift node (3Master, 3Infra, 3Worker), and they are showing the expected permission as 777. 
But, I noticed the /tmp directory on my Bastion/Boot node was changed to 755 (root:root). Other user can't write this folder. 
During the OpenShift installation, Ansible creates temporary directory (for the etcd certificates) on /tmp. Since there is no permission for other user to write the /tmp, the process terminated. 
(In my case, the Ansible playbook is executed on Bastion/Boot node as Ansible server) 

I fixed the problem by giving write permission to Other user by setting it to 777. 
Re-run the deploy-cluster.yaml and it should be fine. 


There is still one big question, why the /tmp on my Bastion/Boot node changed to 755. I am sure my /tmp directory was previously set as 777, but it changed (misteriously?) to 755. (which is the cause of the Permission denied)

PR #2275, opened 2016-08-09
a-o-i: Labeling nodes only

ganhuang:
If node_labels is defined in installer.cfg.yml, it's not resonable to apply the label to nfs/master/etcd hosts.


smunilla:
I agree with this logic. I don't think it hurts anything to apply node_labels everywhere, but we might as well not write it to avoid confusion.

:+1:

Issue #784, opened 2015-11-02
Proposal for the /utils installer wrapper config format


detiber:
I would suggest that the config format should really be something along the lines of (using some variables that we don't allow setting yet with the quick installer to show why I think it'll be valuable):
```
version: v1 # might as well conform with the OpenShift/Kubernetes api version naming scheme
deployment:
  variant: openshift-enterprise
  variant_version: 3.1
  ansible_ssh_user: root  #in reality we could even allow this to be set on a per host basis as well
  ansible_become: no      # same as above
  use_openshift_sdn: yes
  sdn_network_plugin_name: redhat/openshift-ovs-subnet
  masters:
    default_subdomain: apps.test.example.com
    master_identity_providers:
    - name: htpasswd_auth
      login: true
      challenge: true
      kind: HTPasswdPasswordIdentityProvider
      filename: /etc/openshift/htpasswd
    hosts:
    - ip: 10.0.0.1
      hostname: master-private.example.com
      public_ip: 24.222.0.1
      public_hostname: master.example.com
      containerized: true
  nodes:
    kubelet_args:
      max_pods:
      - "100"
    storage_plugin_deps:
    - ceph
    - glusterfs
    hosts:
    - ip: 10.0.0.1
      hostname: master-private.example.com
      public_ip: 24.222.0.1
      public_hostname: master.example.com
      schedulable: no
      containerized: true
    - ip: 10.0.0.2
      hostname: node1-private.example.com
      public_ip: 24.222.0.2
      public_hostname: node1.example.com
      node_labels:
        region: primary
        zone: default
    - ip: 10.0.0.3
      hostname: node2-private.example.com
      public_ip: 24.222.0.3
      public_hostname: node2.example.com
      node_labels:
        region: secondary
        zone: default
  etcd:
    initial_cluster_token: etcd-cluster-1
    hosts:
    - ip: 10.0.0.4
      hostname: etcd1-private.example.com
      etcd_interface: eth0
```

My goal here is to bring the config as close to something that we can easily serialize out to an inventory script (and a set of group/host vars that will allow us to avoid having to json encode variables into a ini file), we should be able to build a very simple inventory file and then serialize the different parameters according to the group or host that they belong.

We should also be able to more easily support features that aren't yet exposed in the cli through the unattended installation this way.

The resulting inventory would be:
```
[OSEv3:children]
masters
nodes
etcd

[masters]
10.0.0.1

[nodes]
10.0.0.1
10.0.0.2
[...]

PR #2276, opened 2016-08-09
a-i-o: Relocate the variables related ansible

ganhuang:
ansible_log_path and ansible_config should be the env variables used by ansible, they needn't to be written to inventory hosts. 
ansible_ssh_user is a variable under [OSEv3:vars] in inventory hosts, it's better to be moved to development in installer.cfg.yml.


smunilla:
This also matches the original proposal #784 . I must have misread that in the first place.

:+1:

smunilla:
#2278 opened with this + nosetest updates

PR #2280, opened 2016-08-10
docker-registry's middleware stanza should contain 'registry' and 'storage' by default

ganhuang:
Fix: https://bugzilla.redhat.com/show_bug.cgi?id=1356823

### DIFF
--- a/roles/openshift_hosted/templates/registry_config.j2
+++ b/roles/openshift_hosted/templates/registry_config.j2
@@ -58,6 +58,8 @@ auth:
   openshift:
     realm: openshift
 middleware:
+  registry:
+  - name: openshift
   repository:
   - name: openshift
     options:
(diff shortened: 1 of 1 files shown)


ganhuang:
That's will be cool if bz#1365730 can be fixed in the same time so that our testing using s3 backend storage isn't blocked. :smile:

sdodson:
LGTM, thanks

sdodson:
aos-ci-test

roles/openshift_hosted/templates/registry_config.j2
```diff
@@ -58,6 +58,8 @@ auth:
   openshift:
     realm: openshift
 middleware:
+  registry:
+  - name: openshift
```
sdodson: After discussing with @miminar this config is only valid for 2.4.1 registry, so we need to make this conditional at least on 3.3 for now. Wrap this in 
```
{% if openshift.common.version_gte_3_3_or_1_3 | bool %}
{% endif -%}
```

roles/openshift_hosted/templates/registry_config.j2
```diff
@@ -69,4 +71,7 @@ middleware:
       baseurl: {{ openshift.hosted.registry.storage.s3.cloudfront.baseurl }}
       privatekey: {{ openshift.hosted.registry.storage.s3.cloudfront.privatekeyfile }}
       keypairid: {{ openshift.hosted.registry.storage.s3.cloudfront.keypairid }}
+{% else %}
+  storage:
+  - name: openshift
```
sdodson: Same here.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

PR #2281, opened 2016-08-10
a-o-i: flexible to set openshift_hostname and openshift_public_hostname

ganhuang:
In my previous testing on OpenStack (Neutron deployment), we're bothered that 'openshift_hostname' and 'openshift_public_hostname' is set to [instance_name].novalocal by quick-installer, this hostname can't be resloved in Neutron deployment (only host-192-168-2-xx can be resloved by internal dns). Thus we have to modify installer.cfg.yml manually, this is really not convenient. 

(Note: in nova-network deployment, [instance_name].novalocal can be resloved successfully, but nova-network has been deprecated in recent OpenStack releases)

So if users input a hostname instead of ip address, it's better to apply the hostname to the values of "openshift_hostname" and "openshift_public_hostname" directly.

### DIFF
--- a/utils/src/ooinstall/openshift_ansible.py
+++ b/utils/src/ooinstall/openshift_ansible.py
@@ -5,6 +5,7 @@ import subprocess
 import sys
 import os
 import yaml
+import re
 from ooinstall.variants import find_variant
 
 CFG = None
@@ -172,6 +173,10 @@ def write_host(host, role, inventory, schedulable=None):
 
     if host.preconfigured:
         return
+    
+    # Assuming host.connect_to is a hostname if it's not all-numeric
+    if not re.match(r"[\d.]+$", host.connect_to):
+        host.public_hostname = host.hostname = host.connect_to
 
     facts = ''
     if host.ip:


ganhuang:
@smunilla , could you help review when you get a chance?

tbielawa:
@ganhuang I looked up this `novalocal` suffix you mentioned and found a lot of people having issues with that. It looks like this is a side effect of your specific Neutron deployment.

https://www.redhat.com/archives/rdo-list/2015-March/msg00055.html

> The hostname inside of the guest is typically set via the metadata service, which is native to Nova, and not the Neutron service. Inside of your Nova configuration file you’ll see this configuration option-

> [root stack-node1 ~(keystone_demo)]# grep dhcp_domain /etc/nova/nova.conf
> #dhcp_domain=novalocal

> Which is why when you launch an instance, the cloud-init (or similar) service configures the local hostname based on this, and not what the DHCP service is giving out. You can test this inside of your instance:

> $ curl http://169.254.169.254/latest/meta-data/local-hostname
> test1.novalocal


Also https://bugs.launchpad.net/tripleo/+bug/1449520

> Nova adds a .novalocal suffix to its guests hostname unless dhcp_domain in nova.conf is set to something else OR to an empty string; this causes all guests (undercloud, overcloud and overcloud's guests) to have an unmanaged (and unknown to dns) domain part into their hostname.

> Neutron will provide as well, via dhcp, its own (managed) domain part, set to openstacklocal by default, causing the contents of /etc/hostname , /etc/hosts and /etc/resolv.conf to be inconsistent and as such issues where different tooling assumes different FQDNs for the same host.

I don't think it's right to change how the openshift_hostname (and assorted other vars) are set across the board for all users. You may want to look into configuring your deployment such that hosts can resolve their own hostnames correctly first.

One more thing from that Launchpad issue, it looks like that `novalocal` auto-appending behavior was changed back in early 2015. Are you running an older version?

I looked around for how to change the domain name settings and came across this openstack question:

https://ask.openstack.org/en/question/3185/how-to-set-domain-name-in-quantum/

It shows how some users were able to reconfigure their deployment to not include .novalocal in the resolved hostname.

Let me know if this isn't applicable to your use case. If it's not then we can continue looking for options.

ganhuang:
> https://www.redhat.com/archives/rdo-list/2015-March/msg00055.html

After removing "set_hostname", "update_hostname" and "update_etc_hosts" in cloud-config, the hostname of my vm becomes "host-172-16-120-187". 
```
cloud_init_modules:
 - rh_subscription
 - migrator
 - bootcmd
 - write-files
 - growpart
 - resizefs
# - set_hostname
# - update_hostname
# - update_etc_hosts
 - rsyslog
 - users-groups
 - ssh
```
But "host-172-16-120-187" was still failed to reslove due to https://openstack.nimeyo.com/51153/openstack-neutron-subnets-dns-nameserver-doesnt-order-input. This issue isn't fixed in OSP7, what I could think of is to set up a instance running dnsmasq in the subnet for internal and external DNS resolution.

>One more thing from that Launchpad issue, it looks like that novalocal auto-appending behavior was changed back in early 2015. Are you running an older version?

No, I'm using OSP7. Actually the novalocal suffix doesn't matter to me as "host-172-16-120-187" can be resloved regardless of including it or not. (at lease it's not a issue in OSP7, looks like "novalocal" in Neutron is for the compatibility with Nova-network)

>I don't think it's right to change how the openshift_hostname (and assorted other vars) are set across the board for all users. You may want to look into configuring your deployment such that hosts can resolve their own hostnames correctly first

Yeah, now I agree with you. According to our investigations, it's a OpenStack DNS resolution issue indeed. Current logic to get the hostname should be ok.
This issue has been fixed in Mitaka: https://specs.openstack.org/openstack/nova-specs/specs/mitaka/implemented/internal-dns-resolution.html. Thanks for your help, @tbielawa !

tbielawa:
@ganhuang is this issue OK to close then?

ganhuang:
It's most likely to be an OpenStack issue and should have been fixed in OpenStack Mitaka, so close it.

tbielawa:
Thanks @ganhuang !

PR #2282, opened 2016-08-10
Update docker-registry middleware config

andrewklau:
From https://github.com/openshift/origin/issues/10324


sdodson:
I believe this is covered in #2280

PR #2285, opened 2016-08-11
Improvements for Docker 1.10+ Upgrade Image Nuking

dgoodwin:
Shutdown docker before the rpm upgrade, this prevents the image migration during an rpm transaction which does a docker restart if it's already running, thus preventing the two versions of docker installed bug that was seen.

Switched to using the more common command to clear out all images, this seems less error prone than the variant I was using.

Added a parallel step prior to upgrade (if we're crossing Docker 1.10 boundary) to delete all *unused* images on Docker hosts. This should be mostly safe to interrupt and retry as no upgrade steps have been performed yet.

Added some logging of image counts before/after image cleanup.


sdodson:
LGTM, thanks for digging into this.

PR #2235, opened 2016-07-28
a-o-i: Add support for Atomic Registry Installs

smunilla:
Add the Registry deployment type as an option in the quick installer.

Supersedes #2318

PR #2286, opened 2016-08-11
a-o-i: Automatically Label Nodes as Infra

smunilla:
In interactive mode, automatically label nodes as infrastructure nodes. Two cases are covered:

1) If all nodes are masters, all hosts are labeled infra
2) If dedicated nodes are defined, the first two (or one) nodes are labeled as infra


smunilla:
@abutcher Pulling this out of #2235 as discussed.

smunilla:
aos-ci-test

detiber:
Shouldn't we prompt users for identifying infra nodes or require unattended installs to provide the labels explicitly?

I worry about automatically determining what nodes should be infra nodes for a user without knowing their intent.

If we are worried about not deploying "infra" components automatically, I would rather we default to not setting a selector for those deployments rather than just selecting nodes to use, since distinct "infra" hosts should be not be able to accept user-created pods.

PR #2287, opened 2016-08-11
qps typo

deads2k:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1364431

This typo causes all of our core controllers to share a QPS of the default: 5 (not a typo).  Shockingly, our system mostly runs, but verrrrry poorly.

@sdodson ptal

### DIFF
--- a/roles/openshift_node/templates/node.yaml.v1.j2
+++ b/roles/openshift_node/templates/node.yaml.v1.j2
@@ -16,8 +16,8 @@ kubeletArguments: {{ openshift.node.kubelet_args | default(None) | to_padded_yam
 masterClientConnectionOverrides:
   acceptContentTypes: application/vnd.kubernetes.protobuf,application/json
   contentType: application/vnd.kubernetes.protobuf
-  burst: 40
-  ops: 20
+  burst: 200
+  qps: 100
 {% endif %}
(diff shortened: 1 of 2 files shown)


sdodson:
@dgoodwin FYI

dgoodwin:
OMG, my bad. It was correct in the example I was given. Sorry dudes.

roles/openshift_node/templates/node.yaml.v1.j2
```diff
@@ -16,8 +16,8 @@ kubeletArguments: {{ openshift.node.kubelet_args | default(None) | to_padded_yam
 masterClientConnectionOverrides:
   acceptContentTypes: application/vnd.kubernetes.protobuf,application/json
   contentType: application/vnd.kubernetes.protobuf
-  burst: 40
-  ops: 20
+  burst: 200
+  qps: 100
```
sdodson: @deads2k  I should've asked before merging, but wanted to verify this value change was intentional.

deads2k: > @deads2k I should've asked before merging, but wanted to verify this value change was intentional.

Yes.  These were values we default to in the code.

PR #2290, opened 2016-08-12
a-o-i: Fix broken upgrades

smunilla:
Upgrade was looking for the hosts in the old location. Quick fix to get it looking in the
correct location.

Fixes BZ#1357627.


sdodson:
aos-ci-test

PR #2291, opened 2016-08-12
Bug 1366595 - installer failed at task "Validate permissions on certificate files"

abutcher:
Ensure etcd user exists in etcd_server_certificates by installing etcd.

@sdodson PTAL

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1366595

### DIFF
--- a/roles/etcd_server_certificates/tasks/main.yml
+++ b/roles/etcd_server_certificates/tasks/main.yml
@@ -1,4 +1,8 @@
 ---
+- name: Install etcd
+  action: "{{ ansible_pkg_mgr }} name=etcd state=present"
+  when: not etcd_is_containerized | bool
+
 - name: Check status of etcd certificates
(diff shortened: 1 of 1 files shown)


sdodson:
LGTM

roles/etcd_server_certificates/tasks/main.yml
```diff
@@ -1,4 +1,8 @@
 ---
+- name: Install etcd
+  action: "{{ ansible_pkg_mgr }} name=etcd state=present"
+  when: not etcd_is_containerized | bool
+
```
detiber: with us now requiring 2.1+, couldn't we remove ansible_pkg_mgr invocations and replace with the generic package module?

abutcher: We sure could

PR #2292, opened 2016-08-12
a-o-i: Bug1329455, [quick-install] Cannot add new nodes in pre-existing env

ganhuang:
It's enough to identify installed/uninstalled hosts by mastr and node roles instead of all kins of hosts.

### DIFF
--- a/utils/src/ooinstall/cli_installer.py
+++ b/utils/src/ooinstall/cli_installer.py
@@ -721,9 +721,9 @@ def get_hosts_to_run_on(oo_cfg, callback_facts, unattended, force, verbose):
             for uninstalled_host in uninstalled_hosts:
                 click.echo("{} is currently uninstalled".format(uninstalled_host))
             # Fall through
-            click.echo('\nUninstalled hosts has been detected in your environment. ' \
-                       'Please make sure your environment was been installed successfully ' \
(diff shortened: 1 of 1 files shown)


detiber:
> It's enough to identify installed/uninstalled hosts by mastr and node roles instead of all kins of hosts.

@ganhuang I'm not sure that is the case, in the case of containerized installs, the lb and storage hosts cannot be on a master or a node (and they cannot be containerized currently).

@smunilla Would it make more sense here to derive installed/uninstalled as sets based on iterating through the defined roles and the hosts that are assigned to each role?

ganhuang:
>
@ganhuang I'm not sure that is the case, in the case of containerized installs, the lb and storage hosts cannot be on a master or a node (and they cannot be containerized currently).

 @detiber , yeah, i see. In the old code, LB will be added to installed_hosts automatically even if LB wasn't installed successfully, and it doesn't handle separate nfs host either. From openshift_fact.py, we could hardly identify if LB/NFS were actually be installed. 

In the new code, I thought in most of cases, only by master and node roles should we be able to identify if the cluster is installed successfully.

detiber:
@abutcher ptal

abutcher:
Fixed the pylint problem and squashed in #2297.

PR #2295, opened 2016-08-15
Allow overriding the Docker 1.10 requirement for upgrade.

dgoodwin:
Respect an explicit docker_version, and the use of docker_upgrade=False.


dgoodwin:
Bump, I think this got lost in the shuffle.

PR #2301, opened 2016-08-15
a-o-i: Fix broken tests from installed hosts check

smunilla:
With the latest changes to behavior around uninstalled hosts during a scaleup,
nosetests were broken. This updates the tests to verify that we exit when we are
expecting.


sdodson:
aos-ci-test

PR #2198, opened 2016-07-22
Various logging enhancements

richm:
These are various enhancements to make deploying and re-deploying logging a bit easier.

PR #2305, opened 2016-08-16
Fix hosted_logging

enoodle:
This is fixing some problems with the hosted logging role, probably the logging deployment project has changed and this role didn't keep up:
- using ```logging-deployer-account-template``` to create the service accounts
- processing the ```imagestream``` template instead of ```support```
- ```fluentd``` is deployed using daemon set so there are no replication controllers for it
- ```loggging-deployer``` SA need the oauth-editor role too

I also added a fact to allow the user to deploy the hosted logging from the inventory file.

[1] https://github.com/openshift/origin-aggregated-logging/tree/master/deployer

### DIFF
--- a/playbooks/common/openshift-cluster/openshift_hosted.yml
+++ b/playbooks/common/openshift-cluster/openshift_hosted.yml
@@ -20,7 +20,7 @@
   - role: openshift_projects
     # TODO: Move standard project definitions to openshift_hosted/vars/main.yml
     # Vars are not accessible in meta/main.yml in ansible-1.9.x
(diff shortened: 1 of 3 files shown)


enoodle:
cc @simon3z @detiber @sdodson

sdodson:
aos-ci-test

richm:
Does this overlap with or supersede https://github.com/openshift/openshift-ansible/pull/2202 and https://github.com/openshift/openshift-ansible/pull/2198 ?

detiber:
@enoodle @richm I suspect there is overlap, we will be taking a closer look at logging early next week.

richm:
@detiber Great!  Please let me know how I can help.

enoodle:
I think #2202 includes this one and is more thorough. 
@richm  I couldn't find something like that when I started making this one, my only suggestions are to add ```hosted_logging``` somewhere in the patch description and to add a way to include this role in a regular installation (something like I did here with ```openshift_hosted_logging_deploy```. Anyway I am closing this one.

PR #2307, opened 2016-08-16
Bug 1367323 - the "OpenShift Container Platform 3.2" variant is still listed when quick install ose-3.3

smunilla:
Remove 3.2 from the install options in 3.3 installer.


detiber:
:+1:

PR #2309, opened 2016-08-16
a-o-i: Fix nosetests after removing 3.2 from installer

smunilla:
Update the tests to match 3.3 being the only version.


sdodson:
aos-ci-test

PR #2310, opened 2016-08-16
Reconcile roles after master upgrade, but before nodes.

dgoodwin:
Prevents the network egress bug causing node restart to fail during 3.3
upgrade. (even though a separate fix is incoming for this)

Only catch is preventing the openshift_cli role, which requires docker,
from triggering a potential upgrade, which we still don't want at this
point. To avoid we use the same variable to protect docker installed
version as we use in pre.yml.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1366722


detiber:
lgtm

sdodson:
LGTM2

PR #2316, opened 2016-08-17
Fix errors in docker role

mkumatag:
In https://github.com/openshift/openshift-ansible/blob/master/roles/docker/tasks/main.yml#L111 and many places in the file docker_check.stat.isreg variable has been referred directly without check it whether variable got value or not.

In latest docker release like 1.11 and 1.12, I don't see /etc/sysconfig/docker present in docker package, because of this test scripts will message like - 

The error was: error while evaluating conditional (docker_check.stat.isreg): 'dict object' has no attribute 'is_reg'

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -67,7 +67,7 @@
     dest: /etc/sysconfig/docker
     regexp: '^{{ item.reg_conf_var }}=.*$'
     line: "{{ item.reg_conf_var }}='{{ item.reg_fact_val | oo_prepend_strings_in_list(item.reg_flag ~ ' ') | join(' ') }}'"
-  when: item.reg_fact_val != '' and  docker_check.stat.isreg
+  when: item.reg_fact_val != '' and  (docker_check.stat.isreg is define and docker_check.stat.isreg)
   with_items:
   - reg_conf_var: ADD_REGISTRY
     reg_fact_val: "{{ docker_additional_registries | default(None, true)}}"
@@ -96,7 +96,8 @@
       reg_fact_val: "{{ docker_no_proxy | default('') | join(',') }}"
   notify:
     - restart docker
-  when: "{{ 'http_proxy' in openshift.common or 'https_proxy' in openshift.common and docker_check.stat.isreg }}"
+  when:
+    - (docker_check.stat.isreg is define and docker_check.stat.isreg) and "{{ 'http_proxy' in openshift.common or 'https_proxy' in openshift.common }}"
 
 - name: Set various Docker options
   lineinfile:
@@ -108,7 +109,7 @@
       {% if docker_log_options is defined %} {{ docker_log_options |  oo_split() | oo_prepend_strings_in_list('--log-opt ') | join(' ')}}{% endif %}\
       {% if docker_options is defined %} {{ docker_options }}{% endif %}\
       {% if docker_disable_push_dockerhub is defined %} --confirm-def-push={{ docker_disable_push_dockerhub | bool }}{% endif %}'"
-  when: docker_check.stat.isreg
+  when: docker_check.stat.isreg is define and docker_check.stat.isreg
   notify:
     - restart docker
 
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -67,7 +67,7 @@
     dest: /etc/sysconfig/docker
     regexp: '^{{ item.reg_conf_var }}=.*$'
     line: "{{ item.reg_conf_var }}='{{ item.reg_fact_val | oo_prepend_strings_in_list(item.reg_flag ~ ' ') | join(' ') }}'"
-  when: item.reg_fact_val != '' and  (docker_check.stat.isreg is define and docker_check.stat.isreg)
+  when: item.reg_fact_val != '' and  (docker_check.stat.isreg is defined and docker_check.stat.isreg)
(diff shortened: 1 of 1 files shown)


roles/docker/tasks/main.yml
```diff
@@ -67,7 +67,7 @@
     dest: /etc/sysconfig/docker
     regexp: '^{{ item.reg_conf_var }}=.*$'
     line: "{{ item.reg_conf_var }}='{{ item.reg_fact_val | oo_prepend_strings_in_list(item.reg_flag ~ ' ') | join(' ') }}'"
-  when: item.reg_fact_val != '' and  docker_check.stat.isreg
+  when: item.reg_fact_val != '' and  (docker_check.stat.isreg is define and docker_check.stat.isreg)
```
abutcher: We may want to check for the existence of the `is_reg` key here rather than using `is defined`.

```
when: item.reg_fact_val != '' and  ('is_reg' in docker_check.stat and docker_check.stat.isreg)
```

Side note: Where did you get your docker package? Assuming it came from centos/rhel/fedora it should create `/etc/sysconfig/docker`.

detiber: @abutcher wouldn't the default filter work for the test as well? might be a bit cleaner.

abutcher: The default filter should work in this case this `docker_check.stat` will always be present.

That would make it something like:
```
when: item.reg_fact_val != '' and  (docker_check.stat.isreg | default(false))
```

mkumatag: @abutcher I followed just the example from [ansible stat documentation](http://docs.ansible.com/ansible/stat_module.html).
If you follow the docker install instructions from [docker site](https://docs.docker.com/engine/installation/linux/rhel/) you will hit this issue

mkumatag:
@abutcher , @detiber IMO using `docker_check.stat.isreg is defined` is better solution and even [ansible stat documenation](http://docs.ansible.com/ansible/stat_module.html) says to use `is defined` in the examples. Please let me know if any other comments.

abutcher:
ok to test

abutcher:
aos-ci-test

abutcher:
I'm cool with `is defined` :+1:

mkumatag:
Why CI is taking so much of time to finish?

detiber:
@mkumatag glitch in the matrix.

detiber:
aos-ci-test

roles/docker/tasks/main.yml
```diff
@@ -96,7 +96,8 @@
       reg_fact_val: "{{ docker_no_proxy | default('') | join(',') }}"
   notify:
     - restart docker
-  when: "{{ 'http_proxy' in openshift.common or 'https_proxy' in openshift.common and docker_check.stat.isreg }}"
+  when:
+    - (docker_check.stat.isreg is defined and docker_check.stat.isreg) and "{{ 'http_proxy' in openshift.common or 'https_proxy' in openshift.common }}"
```
detiber: @mkumatag @abutcher taking a closer look at this, I'm not sure this will work, it is mixing an implicit jinja2 template with an explicit template.

This should probably be:
```
docker_check.stat.isreg is defined and docker_check.stat.isreg and  ('http_proxy' in openshift.common or 'https_proxy' in openshift.common)
```
or maybe even:
```
docker_check.stat.isreg is defined and docker_check.stat.isreg and openshift.common.http_proxy | default(openshift.common.https_proxy| default(none)) is not none
```

roles/docker/tasks/main.yml
```diff
@@ -67,7 +67,7 @@
     dest: /etc/sysconfig/docker
     regexp: '^{{ item.reg_conf_var }}=.*$'
     line: "{{ item.reg_conf_var }}='{{ item.reg_fact_val | oo_prepend_strings_in_list(item.reg_flag ~ ' ') | join(' ') }}'"
-  when: item.reg_fact_val != '' and  docker_check.stat.isreg
+  when: item.reg_fact_val != '' and  (docker_check.stat.isreg is defined and docker_check.stat.isreg)
```
detiber: parenthesis shouldn't be needed here, but not something I'd consider a blocker.

roles/docker/tasks/main.yml
detiber: that said, my brain has now entered PTO-mode, so I can't remember if the first suggestion can just be wrapped in double quotes or if it does indeed need to be an explicit template.

mkumatag:
@detiber fixed your review comments in laytest commit, can you please review. Ran the playbooks/byo/config.yml playbook, it is working fine.

detiber:
aos-ci-test

detiber:
@abutcher ptal

mkumatag:
Not sure what tests are failed here!

abutcher:
aos-ci-test

abutcher:
LGTM. The overall test is not updated by our CI following a passing test run and it's something we need to fix.

abutcher:
aos-ci-test

mkumatag:
@abutcher can you invoke tests again? still no results!

abutcher:
aos-ci-test

mkumatag:
@abutcher @detiber seems like tests are stuck again somewhere!

abutcher:
@mkumatag Nope, we're all good to merge here! Overall isn't reported but all of the required checks are green.

PR #2318, opened 2016-08-17
Add registry console template

aweiteka:
See deployment notes: https://gist.github.com/aweiteka/12a59f01c03219b3c92f2f99b48638c3

Dependency for #2235 

cc @smunilla @abutcher

### DIFF
--- a/roles/openshift_examples/files/examples/v1.3/infrastructure-templates/origin/registry-console.yaml
+++ b/roles/openshift_examples/files/examples/v1.3/infrastructure-templates/origin/registry-console.yaml
@@ -1,0 +1,135 @@
+kind: Template
+apiVersion: v1
+metadata:
+  name: "registry-console"
+  annotations:
+    description: "Template for deploying registry web console. Requires cluster-admin."
+    tags: infrastructure
+labels:
+  createdBy: "registry-console-template"
+objects:
+  - kind: DeploymentConfig
+    apiVersion: v1
+    metadata:
+      name: "registry-console"
+      labels:
+        name: "registry-console"
+    spec:
+      triggers:
+      - type: ConfigChange
+      replicas: 1
+      selector:
+        name: "registry-console"
+      template:
+        metadata:
+          labels:
+            name: "registry-console"
+        spec:
+          containers:
+            - name: registry-console
+              image: cockpit/kubernetes
+              ports:
+                - containerPort: 9090
+                  protocol: TCP
+              livenessProbe:
+                failureThreshold: 3
+                httpGet:
+                  path: /ping
+                  port: 9090
+                  scheme: HTTP
+                initialDelaySeconds: 10
+                periodSeconds: 10
+                successThreshold: 1
+                timeoutSeconds: 5
+              readinessProbe:
+                failureThreshold: 3
+                httpGet:
+                  path: /ping
+                  port: 9090
+                  scheme: HTTP
+                periodSeconds: 10
+                successThreshold: 1
+                timeoutSeconds: 5
+              env:
+                - name: OPENSHIFT_OAUTH_PROVIDER_URL
+                  value: "${OPENSHIFT_OAUTH_PROVIDER_URL}"
+                - name: OPENSHIFT_OAUTH_CLIENT_ID
+                  value: "${OPENSHIFT_OAUTH_CLIENT_ID}"
+                - name: KUBERNETES_INSECURE
+                  value: "${KUBERNETES_INSECURE}"
+                - name: KUBERNETES_CA_DATA
+                  value: "${KUBERNETES_CA_DATA}"
+                - name: COCKPIT_KUBE_INSECURE
+                  value: "${COCKPIT_KUBE_INSECURE}"
+                - name: REGISTRY_ONLY
+                  value: "true"
+                - name: REGISTRY_HOST
+                  value: "${REGISTRY_HOST}"
+  - kind: Service
+    apiVersion: v1
+    metadata:
+     name: "registry-console"
+     labels:
+       name: "registry-console"
+    spec:
+      type: ClusterIP
+      ports:
+        - name: registry-console
+          protocol: TCP
+          port: 9000
+          targetPort: 9090
+      selector:
+        name: "registry-console"
+  - kind: ImageStream
+    apiVersion: v1
+    metadata:
+      name: registry-console
+      annotations:
+        description: Atomic Registry console
+    spec:
+      tags:
+        - annotations: null
+          from:
+            kind: DockerImage
+            name: ${IMAGE_NAME}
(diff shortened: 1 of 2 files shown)


roles/openshift_examples/files/examples/v1.3/infrastructure-templates/origin/registry-console.yaml
```diff
@@ -0,0 +1,135 @@
...
+    name: OPENSHIFT_OAUTH_CLIENT_SECRET
+    from: "user[a-zA-Z0-9]{64}"
+    generate: expression
+  - description: "Oauth client id"
+    name: OPENSHIFT_OAUTH_CLIENT_ID
+    value: "cockpit-oauth-client"
+  - description: "Skip kubernetes CA verification"
+    name: KUBERNETES_INSECURE
+    value: "false"
+  - description: "PEM Encoded certificate to use for CA verification"
```
smarterclayton: Shouldn't this just use the mounted service account token and ca?

roles/openshift_examples/files/examples/v1.3/infrastructure-templates/origin/registry-console.yaml
```diff
@@ -0,0 +1,135 @@
...
+    required: true
+  - description: "Console will auto-generate self-signed certificates if secure is false."
+    name: COCKPIT_KUBE_INSECURE
+    value: "false"
+  - description: "Oauth client secret"
+    name: OPENSHIFT_OAUTH_CLIENT_SECRET
+    from: "user[a-zA-Z0-9]{64}"
+    generate: expression
+  - description: "Oauth client id"
+    name: OPENSHIFT_OAUTH_CLIENT_ID
```
smarterclayton: Now that we have service accounts able to be oauth clients, can this just use that?

roles/openshift_examples/files/examples/v1.3/infrastructure-templates/origin/registry-console.yaml
```diff
@@ -0,0 +1,135 @@
...
+    name: OPENSHIFT_OAUTH_PROVIDER_URL
+    required: true
+  - description: "The registry console URL, e.g. https://registry-console-default.example.com"
+    name: COCKPIT_KUBE_URL
+    required: true
+  - description: "Console will auto-generate self-signed certificates if secure is false."
+    name: COCKPIT_KUBE_INSECURE
+    value: "false"
+  - description: "Oauth client secret"
+    name: OPENSHIFT_OAUTH_CLIENT_SECRET
```
smarterclayton: If you use service account oauth clients, you can simply use your token file.

roles/openshift_examples/files/examples/v1.3/infrastructure-templates/origin/registry-console.yaml
```diff
@@ -0,0 +1,135 @@
...
+  - description: 'Specify image version; e.g. for "cockpit/kubernetes:latest", set version "latest"'
+    name: IMAGE_VERSION
+    value: latest
+  - description: "The public URL for the Openshift OAuth Provider"
+    name: OPENSHIFT_OAUTH_PROVIDER_URL
+    required: true
+  - description: "The registry console URL, e.g. https://registry-console-default.example.com"
+    name: COCKPIT_KUBE_URL
+    required: true
+  - description: "Console will auto-generate self-signed certificates if secure is false."
```
smarterclayton: You should be able to use the new service serving cert generator to get certs

sdodson:
@aweiteka Is there an upstream source we could pull this from? If so can you add it to roles/openshift_examples/example-sync.sh so we retrieve updates each time we update the installer?

aweiteka:
> Is there an upstream source we could pull this from? If so can you add it to roles/openshift_examples/example-sync.sh so we retrieve updates each time we update the installer?

@petervo what do you think? It does seem better to maintain a single template for the registry console. My work here just cleans a few things up, changing defaults and it's YAML. Hard to diff, I know. :(

roles/openshift_examples/files/examples/v1.3/infrastructure-templates/origin/registry-console.yaml
petervo: The variable only controls if cockpit servers with SSL or not. When serving with ssl, you can provide the certificate however you want. It just needs to be mounted in ```/etc/cockpit/ws-certs.d```. In the format documented here. http://cockpit-project.org/guide/latest/https.html#https-certificates. A new certificate will only be generated if nothing is present in that directory.

roles/openshift_examples/files/examples/v1.3/infrastructure-templates/origin/registry-console.yaml
petervo: We do use the mounted service account CA when present.

roles/openshift_examples/files/examples/v1.3/infrastructure-templates/origin/registry-console.yaml
petervo: @smarterclayton does this work to log users in across projects? My reading of the docs (https://docs.openshift.org/latest/architecture/additional_concepts/authentication.html)  seems to indicate that the resulting token will be restricted to within the serviceaccounts namespace. Which doesn't work as the users needs to be able view and manage multiple projects as their permissions allow. FWIW we are using the implicit grant oauth flow.

roles/openshift_examples/files/examples/v1.3/infrastructure-templates/origin/registry-console.yaml
aweiteka: I'm thinking we should just hard-code this in the template. I don't see a need to expose as a param since cockpit makes it so easy to use SSL.

petervo:
The one in cockpit is meant to be used for both the registry and kubernetes so things are named a little differently. If we want a version that is easier for someone that only cares about the registry to read and understand we are always going to have two. I don't mind that being here, but if you prefer we have it in cockpit and this pulls it from there, I don't mind that either.

roles/openshift_examples/files/examples/v1.3/infrastructure-templates/origin/registry-console.yaml
petervo: A lot of these variables are needed for different scenarios the openshift registry users won't really ever hit. So if that's the target audience for this template it's probably ok to leave some of these out.

aweiteka:
We'll maintain a registry-specific template in this repo.

@smunilla [updated gist](https://gist.github.com/aweiteka/12a59f01c03219b3c92f2f99b48638c3) with deployment instructions.

sdodson:
Said to be superseded by #2235

PR #2323, opened 2016-08-18
Mount /sys rw

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1367937


sdodson:
Mirrors https://github.com/openshift/origin/pull/10516

PR #2325, opened 2016-08-18
Due to problems with with_fileglob lets avoid using it for now

sdodson:
This shouldn't be necessary for OSE installs as repos will be managed and provided via RHSM.

### DIFF
--- a/roles/openshift_repos/tasks/main.yaml
+++ b/roles/openshift_repos/tasks/main.yaml
@@ -37,12 +37,12 @@
   - "fedora-openshift-enterprise/repos/*"
   - "fedora-origin/repos/*"
   - "online/repos/*"
-  - "openshift-enterprise/repos/*"
   - "origin/repos/*"
   - "removed/repos/*"
   when: not openshift.common.is_containerized | bool
         and not (item | search("/files/" ~ openshift_deployment_type ~ "/repos"))
         and (ansible_os_family == "RedHat" and ansible_distribution != "Fedora")
+        and not openshift_deployment_type == 'openshift-enterprise'
   notify: refresh cache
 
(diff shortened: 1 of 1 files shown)


sdodson:
@detiber for your consideration

sdodson:
aos-ci-test

detiber:
I see no issues with this, do we think the hangup was being caused by a missing directory?

sdodson:
aos-ci-test

sdodson:
@detiber I'm unsure of the cause and I don't see any issues upstream that seem relevant. This is mainly me avoiding code that I think is unnecessary on openshift-enterprise deployment types because we've seen repeatable failures there when working on more than a handful of hosts. This doesn't improve anything for non openshift-enterprise deployment types so we should still try to get to the root of this.

sdodson:
aos-ci-test

roles/openshift_repos/tasks/main.yaml
```diff
@@ -37,12 +37,12 @@
...
-  - "openshift-enterprise/repos/*"
   - "origin/repos/*"
   - "removed/repos/*"
   when: not openshift.common.is_containerized | bool
         and not (item | search("/files/" ~ openshift_deployment_type ~ "/repos"))
         and (ansible_os_family == "RedHat" and ansible_distribution != "Fedora")
+        and not openshift_deployment_type == 'openshift-enterprise'
   notify: refresh cache
 
 - name: Remove any yum repo files for other deployment types Fedora
```
detiber: it looks this second loop isn't necessary, since fedora is included above.

Another option that we have is removing the use of with_fileglob and just hardcoding the paths to the files on disk as well.

PR #2327, opened 2016-08-19
Fix warnings in OpenStack provider with ansible 2.1

lhuard1A:
Some expressions now need to be enclosed inside `{{…}}`.


sdodson:
LGTM thanks.

PR #2329, opened 2016-08-19
Fix the “node on master” feature

lhuard1A:
What we want to do is to add the master as a node if:
* `g_nodeonmaster` is set to true, and
* we are not in the case where we want to add new nodes.

The second test was done by only checking whether `g_new_node_hosts` was defined.
This was wrong because, in all cloud-provider setups, this variable was set
with the default value of “empty list” (`[]`).
The test has been changed to use the `bool` filter so that it correctly evaluates
to false (and hence, effectively add the master as a node) when `g_new_node_hosts`
is the empty list.


sdodson:
LGTM

sdodson:
aos-ci-test

PR #2333, opened 2016-08-19
a-o-i: Fix ansible_ssh_user question

smunilla:
Since we moved 'ansible_ssh_user' under 'deployment', we need to update the criteria for when
to ask the user to reenter that information.


smunilla:
aos-ci-test

PR #2334, opened 2016-08-19
Bug 1358951 - Error loading config, no such key: 'deployment' when using ing previously valid answers file

smunilla:
Update the quick installer to automatically convert from the old config file format to
the new format.

### DIFF
--- a/utils/src/ooinstall/oo_config.py
+++ b/utils/src/ooinstall/oo_config.py
@@ -167,16 +167,23 @@ class OOConfig(object):
 
 # pylint: disable=too-many-branches
     def _read_config(self):
+        def _print_read_config_error(error, path='the configuration file'):
+            message = """
+Error loading config. {}.
+
+See https://docs.openshift.com/enterprise/latest/install_config/install/quick_install.html#defining-an-installation-configuration-file
+for information on creating a configuration file or delete {} and re-run the installer.
+"""
+            print message.format(error, path)
+
         try:
             if os.path.exists(self.config_path):
                 with open(self.config_path, 'r') as cfgfile:
                     loaded_config = yaml.safe_load(cfgfile.read())
 
-                # Use the presence of a Description as an indicator this is
-                # a legacy config file:
-                if 'Description' in self.settings:
-                    self._upgrade_legacy_config()
-
+                if not 'version' in loaded_config:
+                    _print_read_config_error('Legacy configuration file found', self.config_path)
+                    sys.exit(0)
 
                 if loaded_config.get('version', '') == 'v1':
                     loaded_config = self._upgrade_v1_config(loaded_config)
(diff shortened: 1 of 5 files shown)


smunilla:
aos-ci-test

utils/src/ooinstall/oo_config.py
```diff
@@ -176,6 +177,10 @@ def _read_config(self):
                 if 'Description' in self.settings:
```
detiber: A bit unrelated to this PR, but would the legacy upgrade make more sense to key off the version field rather than 'Description'?

utils/src/ooinstall/oo_config.py
```diff
@@ -236,9 +241,36 @@ def _upgrade_legacy_config(self):
         self.settings['variant'] = 'openshift-enterprise'
         self.settings['variant_version'] = '3.0'
```
detiber: Would this default version from the legacy config even work, or would it just error out?

utils/src/ooinstall/oo_config.py
```diff
@@ -236,9 +241,36 @@ def _upgrade_legacy_config(self):
...
+        for host in config['hosts']:
+            host_props = {}
+            host_props['roles'] = []
+            host_props['connect_to'] = host['connect_to']
+
+            for prop in ['ip', 'public_ip', 'hostname', 'public_hostname', 'containerized', 'preconfigured']:
+                host_props[prop] = host.get(prop, None)
+
+            for role in ['master', 'node', 'master_lb', 'storage', 'etcd']:
+                if host.get(role, False):
```
detiber: I might be mis-remembering the old format, but weren't the 'roles' determined by `is_master`, `is_node`, etc?

utils/src/ooinstall/oo_config.py
smunilla: If I remember right, we added the version field specifically to make this easier to find in the future.

utils/src/ooinstall/oo_config.py
detiber: I was wrong, I do see `master` and `node` in the example config here: https://github.com/openshift/openshift-ansible/blob/master/utils/docs/config.md, but did we annotate master_lb, storage, and etcd as well? or did we make some assumptions?

detiber: Probably should update that example config I linked above, as well.

utils/src/ooinstall/oo_config.py
detiber: At some point we need to think about removing support for older versions. If we've supported v1 since at least 3.1, then I would say we can error out and point to the docs if the version field isn't present.

utils/src/ooinstall/oo_config.py
smunilla: I think we just made some assumptions. I'll get that example updated though.

utils/src/ooinstall/oo_config.py
smunilla: I'm all for this. I'm not super confident in the legacy upgrade working at all.

utils/src/ooinstall/oo_config.py
detiber: @smunilla if we were just making assumptions, then I wouldn't expect the migration to work, since the etcd group wouldn't be populated...  or do we continue to make those assumptions with the v2 format if they are not present?

smunilla: @detiber In my testing it does the right thing for embedded etcd, which means the migration should work in most cases. This at least puts us ahead of where we are now, which just fails if they have the old format.

PR #2336, opened 2016-08-19
Fix etcd uninstall

### LINKED ISSUE #2317: Uninstall playbook fails to clean up /var/lib/etcd when it has its own mount point
We often recommend mounting /var/lib/etcd separately from /var for disk write stability. When we do this, the uninstall playbook fails on running `rm -rf /var/lib/etcd` because it's a mounted disk. You'll see an error like 'device or resource busy'.

Is there any reason not to simply change this to `rm -rf /var/lib/etcd/*` or should there be more error checking?

cc: @abutcher @detiber

sdodson:
Really fixes #2317. The other pr not so much :-(


sdodson:
@detiber ok working now, thanks for the tip on shell vs command

Issue #2326, opened 2016-08-19
openshift-ansible 3.2 fails with flannel


jprovaznik:
With current RHEL stable rpms openshift-ansible fails when using flannel:

```
TASK: [etcd_ca | copy ] ******************************************************* 
<flannel-openshift-master-0.example.com> ESTABLISH CONNECTION FOR USER: cloud-user
[...]

Issue #2322, opened 2016-08-18
Installation fails when we select flannel network to deploy with etcd host


mkumatag:
Installation is failing with following error:

```
ok: [jupiter-vm1144.pok.stglabs.ibm.com -> jupiter-vm1108.pok.stglabs.ibm.com] => {"changed": false, "invocation": {"module_args": {"checksum_algorithm": "sha1", "follow": false, "get_checksum": true, "get_md5": true, "mime": false, "path": "/etc/origin/node/ca/ca.crt"}, "module_name": "stat"}, "stat": {"exists": false}}

TASK [etcd_client_certificates : fail] *****************************************
task path: /root/openshift-ansible/openshift-ansible/roles/etcd_client_certificates/tasks/main.yml:9
fatal: [jupiter-vm1144.pok.stglabs.ibm.com]: FAILED! => {"changed": false, "failed": true, "invocation": {"module_args": {"msg": "CA certificate /etc/origin/node/ca/ca.crt doesn't exist on CA host jupiter-vm1108.pok.stglabs.ibm.com. Apply 'etcd_ca' role to jupiter-vm1108.pok.stglabs.ibm.com.\n"}, "module_name": "fail"}, "msg": "CA certificate /etc/origin/node/ca/ca.crt doesn't exist on CA host jupiter-vm1108.pok.stglabs.ibm.com. Apply 'etcd_ca' role to jupiter-vm1108.pok.stglabs.ibm.com.\n"}

NO MORE HOSTS LEFT *************************************************************
	to retry, use: --limit @playbooks/byo/config.retry

PLAY RECAP *********************************************************************
jupiter-vm1108.pok.stglabs.ibm.com : ok=104  changed=36   unreachable=0    failed=0   
jupiter-vm1144.pok.stglabs.ibm.com : ok=352  changed=89   unreachable=0    failed=1   
jupiter-vm931.pok.stglabs.ibm.com : ok=152  changed=42   unreachable=0    failed=0   
localhost                  : ok=13   changed=8    unreachable=0    failed=0   

[root@jupiter-vm1108 openshift-ansible]#
```

The content of /etc/ansible/hosts file

```
# Create an OSEv3 group that contains the masters and nodes groups
[OSEv3:children]
masters
nodes
etcd

# Set variables common for all OSEv3 hosts
[OSEv3:vars]
# SSH user, this user should allow ssh based auth without requiring a password
ansible_ssh_user=root

[...]

PR #2337, opened 2016-08-21
Remove duplicate flannel registration

mkumatag:
In playbooks/common/openshift-cluster/additional_config.yml, flannel_register role is called twice. So remove one entry which is not really required.

### DIFF
--- a/playbooks/common/openshift-cluster/additional_config.yml
+++ b/playbooks/common/openshift-cluster/additional_config.yml
@@ -24,7 +24,3 @@
   - role: cockpit
     when: not openshift.common.is_atomic and ( deployment_type in ['atomic-enterprise','openshift-enterprise'] ) and
       (osm_use_cockpit | bool or osm_use_cockpit is undefined )
-  - role: flannel_register
-    when: openshift.common.use_flannel | bool
-
-
--- a/playbooks/common/openshift-cluster/additional_config.yml
+++ b/playbooks/common/openshift-cluster/additional_config.yml
@@ -1,11 +1,3 @@
-- name: Configure flannel
-  hosts: oo_first_master
-  vars:
-    etcd_urls: "{{ openshift.master.etcd_urls }}"
-  roles:
-  - role: flannel_register
-    when: openshift.common.use_flannel | bool
-
 - name: Additional master configuration
   hosts: oo_first_master
   vars:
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/additional_config.yml
```diff
@@ -24,7 +24,3 @@
   - role: cockpit
     when: not openshift.common.is_atomic and ( deployment_type in ['atomic-enterprise','openshift-enterprise'] ) and
       (osm_use_cockpit | bool or osm_use_cockpit is undefined )
-  - role: flannel_register
```
detiber: I think I would prefer to keep flannel_register in the 'Additional master configuration' play rather than having a separate 'Configure flannel' play. If it needs to run prior to any of the roles here, we should be able to just push it up in the order.

detiber: One question that I've had with the flannel config, is that it appears that there is a single point of failure, with just the "first master" running the flannel_register role, is this something that can also be configured on the other masters, or is there additional work needed to make the flannel integration ha?

mkumatag: @detiber okay, then I'll remove the "Configure flannel" play.

In flannel_register role we are populating the networking template in the etcd db. This holds good even if we run on one master assuming all the masters are linked with same etcd api(cluster)

detiber: > In flannel_register role we are populating the networking template in the etcd db. This holds good even if we run on one master assuming all the masters are linked with same etcd api(cluster)

Ah, good to know, thanks!

mkumatag:
@detiber addressed your comments in latest commit, please review.

detiber:
ok to test

detiber:
aos-ci-test

detiber:
@jprovaznik do you see any issues with this change?

jprovaznik:
This LGTM but don't take me too seriously because of my shallow ansible knowledge and because I couldn't successfully test this patch - I'm still hitting #2322 either with or without this patch.

mkumatag:
This PR is nothing to do with #2322, that will still fail. This PR will fix only creating network entry in etcd db twice. For #2322 and #2326 still I need to dig deeper to understand what is really happening there!

mkumatag:
@detiber Please let me know if any other comments?

abutcher:
lgtm

PR #2339, opened 2016-08-22
Add warning at end of 3.3 upgrade if pluginOrderOverride is found.

### LINKED ISSUE #2335: Stop promoting `pluginOrderOverride` and use `pluginConfig` instead
In 3.3, we've created a warning for people about using the `pluginOrderOverride` value in their `master-config.yaml` because it causes them to "miss" new admission plugins.

The "proper" way to enable plugins in 3.3 is to create a stanza like this:
```yaml
admissionConfig:
  pluginConfig:
    AlwaysPullImages:
      configuration:
        kind: DefaultAdmissionConfig
        apiVersion: v1
```

Right now, I think that's the only one that is enabled that way, but others may follow.  The rest in the list (https://github.com/openshift/origin/blob/master/pkg/cmd/server/start/admission.go#L71) are all activated by specifying their particular configs.

In retrospect, you may want a separate list between "simple on" plugins and "hard on" plugins.  Do you need that?

@sdodson

dgoodwin:
Fixes #2335 but I still need a URL for the docs, @deads2k any idea what that would be yet?

Output looks like:

```
TASK [command] *****************************************************************
changed: [m1.aos.example.com]

TASK [Warn if pluginOrderOverride is in use in master-config.yaml] *************
ok: [m1.aos.example.com] => {
    "msg": "WARNING pluginOrderOverride is being deprecated in master-config.yaml, please see URLHERE for more information."
}

localhost                  : ok=17   changed=9    unreachable=0    failed=0   
m1.aos.example.com         : ok=167  changed=16   unreachable=0    failed=0   
n1.aos.example.com         : ok=77   changed=1    unreachable=0    failed=0   

ansible-playbook -i ./hosts   30.93s user 11.32s system 21% cpu 3:15.42 total
```

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/post.yml
+++ b/playbooks/common/openshift-cluster/upgrades/post.yml
@@ -56,3 +56,16 @@
       {{ oc_cmd }} patch dc/docker-registry -n default -p
       '{"spec":{"template":{"spec":{"containers":[{"name":"registry","image":"{{ registry_image }}"}]}}}}'
       --api-version=v1
+
+# Check for warnings to be printed at the end of the upgrade:
+- name: Check for warnings
+  hosts: oo_masters_to_config
+  tasks:
+  # Check if any masters are using pluginOrderOverride and warn if so, only for 1.3/3.3 and beyond:
+  - command: >
+      grep pluginOrderOverride {{ openshift.common.config_base }}/master/master-config.yaml
+    register: grep_plugin_order_override
+    when: openshift.common.version_gte_3_3_or_1_3 | bool
+  - name: Warn if pluginOrderOverride is in use in master-config.yaml
+    debug: msg="WARNING pluginOrderOverride is being deprecated in master-config.yaml, please see https://docs.openshift.com/enterprise/latest/architecture/additional_concepts/admission_controllers.html for more information."
+    when: not grep_plugin_order_override | skipped and grep_plugin_order_override.rc == 0


deads2k:
@dgoodwin docs being updated here: https://github.com/openshift/openshift-docs/pull/2695 .  Since you're a user, mind giving them a read?

dgoodwin:
Ok updated with a link to the correct URL where the docs will appear and reviewed @deads2k 's change.

playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
```diff
@@ -58,10 +58,10 @@
...
-    openshift_deployment_type: "{{ deployment_type }}"
-    master_config_hook: "v3_3/master_config_upgrade.yml"
-    node_config_hook: "v3_3/node_config_upgrade.yml"
-- include: ../../../openshift-master/restart.yml
+#- include: ../../../../common/openshift-cluster/upgrades/upgrade.yml
+  #vars:
+    #openshift_deployment_type: "{{ deployment_type }}"
+    #master_config_hook: "v3_3/master_config_upgrade.yml"
+    #node_config_hook: "v3_3/node_config_upgrade.yml"
+#- include: ../../../openshift-master/restart.yml
```
sdodson: Intentional or did something slip into the PR?

dgoodwin: Something very serious and definitely not intentional slipped into the PR. My bad, fixing now.

playbooks/common/openshift-cluster/upgrades/post.yml
```diff
@@ -56,3 +56,16 @@
...
+- name: Check for warnings
+  hosts: oo_masters_to_config
+  tasks:
+  # Check if any masters are using pluginOrderOverride and warn if so, only for 1.3/3.3 and beyond:
+  - command: >
+      grep pluginOrderOverride {{ openshift.common.config_base }}/master/master-config.yaml
+    register: grep_plugin_order_override
+    when: openshift.common.version_gte_3_3_or_1_3 | bool
+  - name: Warn if pluginOrderOverride is in use in master-config.yaml
+    debug: msg="WARNING pluginOrderOverride is being deprecated in master-config.yaml, please see https://docs.openshift.com/enterprise/latest/architecture/additional_concepts/admission_controllers.html for more information."
```
sdodson: We've used `pause: prompt="warning message" seconds="10"` elsewhere. I think we should standardize on one or the other in the future. I'm a bit surprised there's not a standard warning message module in ansible.

playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
dgoodwin: Also 
doge

playbooks/common/openshift-cluster/upgrades/post.yml
sdodson: Given this is at the end, I'm fine with it as is.

sdodson:
LGTM, ready to merge?

dgoodwin:
Good with me.

playbooks/common/openshift-cluster/upgrades/post.yml
detiber: We really should avoid the use of `pause` unless we are working around a timing issue. Relying on a user to see the output is too risky for things that need action.

Ansible does support sending back warnings from modules, but that is part of a particular module, not general feedback to the user.

What we really need is a filter plugin that tracks these things along the way and then outputs additional info to the user after the run has completed.

PR #2340, opened 2016-08-22
Replace some virsh commands by native virt_XXX ansible module

lhuard1A:


playbooks/libvirt/openshift-cluster/tasks/configure_libvirt_network.yml
```diff
@@ -1,27 +1,10 @@
...
-  file:
-    path: "{{ mktemp.stdout }}"
-    state: absent
-  when: net_info_result.rc == 1
+- name: Create the libvirt network for OpenShift
+  virt_net:
+    name: '{{ libvirt_network }}'
+    state: '{{ item }}'
+    xml: "{{ lookup('template', 'network.xml') }}"
+    uri: '{{ libvirt_uri }}'
```
detiber: Should `autostart` be set as well? 

The docs also suggest that `state: active` should be sufficient, but I'm guessing that it isn't considering the with_items.

playbooks/libvirt/openshift-cluster/tasks/configure_libvirt_storage_pool.yml
```diff
@@ -6,22 +6,24 @@
...
-
-- name: Create the libvirt storage pool for openshift
-  command: 'virsh -c {{ libvirt_uri }} pool-create-as {{ libvirt_storage_pool }} dir --target {{ libvirt_storage_pool_path }}'
-  when: pool_info_result.rc == 1
+- name: Create the libvirt storage pool for OpenShift
+  virt_pool:
+    name: '{{ libvirt_storage_pool }}'
+    state: '{{ item }}'
+    xml: "{{ lookup('template', 'storage-pool.xml') }}"
+    uri: '{{ libvirt_uri }}'
```
detiber: Same comments as above with virt_net.

detiber:
lgtm outside of the comments I made inline.

playbooks/libvirt/openshift-cluster/tasks/configure_libvirt_network.yml
lhuard1A: When I tried with only the `state: active`, it failed:

```
$ cat playbooks/libvirt/openshift-cluster/tasks/configure_libvirt_network.yml
---
- name: Create the libvirt network for OpenShift
  virt_net:
    name: '{{ libvirt_network }}'
    state: active
    xml: "{{ lookup('template', 'network.xml') }}"
    uri: '{{ libvirt_uri }}'
```
leads to:
```
TASK [Create the libvirt network for OpenShift] ********************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "network openshift-ansible not found"}
```

In the [example section of the doc](https://docs.ansible.com/ansible/virt_net_module.html#examples), it is suggested that the network must be defined before being “activated”:
```
# Ensure that a network is active (needs to be defined and built first)
- virt_net: state=active name=br_nat
```

For the `autostart`, it depends whether we want the libvirt OpenShift cluster to be automatically restarted when the workstation is rebooted.

playbooks/libvirt/openshift-cluster/tasks/configure_libvirt_storage_pool.yml
lhuard1A: Indeed. It has the same behavior.

[Example section of the documentation:](https://docs.ansible.com/ansible/virt_pool_module.html#examples)
```
# Ensure that a pool is active (needs to be defined and built first)
- virt_pool: state=active name=vms
```

playbooks/libvirt/openshift-cluster/tasks/configure_libvirt_network.yml
detiber: @lhuard1A thanks for the clarification, the present/active issue sounds like an unfortunate implementation decision.

I don't think it would be reasonable for the entire cluster to be restarted when the workstation is rebooted, but without the network and storage pool being enabled, I think it would make it a bit confusing to a user if they try to start the machines after a reboot.

lhuard1A: Indeed, makes sense.
Note that I had to quote the `yes` in `autostart: 'yes'` because without the quotes (i.e.: `autostart: yes`), I got the following error:

```
TASK [Create the libvirt storage pool for OpenShift] ***************************
failed: [localhost] (item=present) => {"failed": true, "item": "present", "msg": "value of autostart must be one of: yes,no, got: True"}
failed: [localhost] (item=active) => {"failed": true, "item": "active", "msg": "value of autostart must be one of: yes,no, got: True"}
```

which seems to have already been fixed upstream by https://github.com/ansible/ansible-modules-extras/pull/2593/files#diff-00d13769f5d06664a824c39ec04d393c . My understanding of the error message is that the yaml parser interprets the `yes` as a boolean and the ansible module expected the string `yes`.

detiber: @lhuard1A of course it does... Thanks for working through that.

PR #2355, opened 2016-08-24
remove fedora origin copr (it's in mainline fedora now), some dnf/yum clean up

maxamillion:


maxamillion:
@detiber +1 - thanks!

PR #2356, opened 2016-08-24
don't set virt_sandbox_use_nfs on Fedora, it was replaced by virt_use_nfs

maxamillion:
Note that virt_use_nfs is already set in a previous play in the role's tasks


detiber:
aos-ci-test

detiber:
@sdodson any concerns with merging this prior to 3.3 freeze?

sdodson:
@detiber no that's fine, though it was my understanding that selinux in fedora had added an alias and rhel was going to do so as well ensuring that the old ones still worked which would be the best solution as I think getting sebooleans is somewhat slow or at least someone has complained about it in a profiling report at some point.

tl;dr -- LGTM

PR #2360, opened 2016-08-24
Repo rework

detiber:
With fedora no longer needing special repo configuration, it only leaves the config needed for the CentOS PaaS SIG, so simplifying for the particular use case needed.

Also removing cleanup of repos on changing deployment_type, since it was mainly used in dev and introduced a lot of complexity.


detiber:
aos-ci-test

detiber:
@sdodson @abutcher ptal

sdodson:
LGTM

detiber:
@maxamillion

PR #2362, opened 2016-08-25
Set openshift_clock_enabled=True by default

sdodson:
Remove openshift_facts dep too, we don't depend on any facts that we're not
setting within the role tasks.

FYI -- I think we should hold this until post 3.3.

### DIFF
--- a/inventory/byo/hosts.aep.example
+++ b/inventory/byo/hosts.aep.example
@@ -442,8 +442,8 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # masterConfig.volumeConfig.dynamicProvisioningEnabled, configurable as of 1.2/3.2, enabled by default
 #openshift_master_dynamic_provisioning_enabled=False
 
-# Configure usage of openshift_clock role.
-#openshift_clock_enabled=true
+# Configure usage of openshift_clock role, on by default.
+#openshift_clock_enabled=false
 
 # OpenShift Per-Service Environment Variables
 # Environment variables are added to /etc/sysconfig files for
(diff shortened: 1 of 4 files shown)


sdodson:
aos-ci-test

inventory/byo/hosts.aep.example
```diff
@@ -442,8 +442,8 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
 # masterConfig.volumeConfig.dynamicProvisioningEnabled, configurable as of 1.2/3.2, enabled by default
 #openshift_master_dynamic_provisioning_enabled=False
 
-# Configure usage of openshift_clock role.
-#openshift_clock_enabled=true
+# Configure usage of openshift_clock role, on by default.
+#openshift_clock_enabled=False
```
detiber: lowercase 'false' or 'no' fits more with upstream Ansible conventions

roles/openshift_clock/tasks/main.yaml
```diff
@@ -3,7 +3,7 @@
   openshift_facts:
     role: clock
     local_facts:
-      enabled: "{{ openshift_clock_enabled | default(None) }}"
+      enabled: "{{ openshift_clock_enabled | default(True) }}"
```
detiber: This is still using openshift_facts. See `openshift.clock.*` below

sdodson: ahh, missed the chrony one.

detiber: I'm good with removing the dependency on openshift_facts, just need to make sure that we replace the existing functionality.

sdodson:
aos-ci-test

sdodson:
This is already the status quo through openshift_facts. Nevermind.

PR #2363, opened 2016-08-25
Pep8 fixes for quick installer

tbielawa:
Many pep8 violations were discovered in the quick installer codebase. Before this patch there were 139 violations. Now there are none:

* **BEFORE**: https://gist.github.com/tbielawa/59d9ca24b3f1c07afb8f421eabb20170#file-before
* **AFTER**: https://gist.github.com/tbielawa/59d9ca24b3f1c07afb8f421eabb20170#file-after

How to verify:

1. Check out this branch
1. Switch into the `utils` dir
1. Run the command `make ci`

### DIFF
--- a/utils/src/ooinstall/openshift_ansible.py
+++ b/utils/src/ooinstall/openshift_ansible.py
@@ -22,15 +22,17 @@ ROLES_TO_GROUPS_MAP = {
 VARIABLES_MAP = {
     'ansible_ssh_user': 'ansible_ssh_user',
     'deployment_type': 'deployment_type',
-    'master_routingconfig_subdomain':'openshift_master_default_subdomain',
-    'proxy_http':'openshift_http_proxy',
+    'master_routingconfig_subdomain': 'openshift_master_default_subdomain',
+    'proxy_http': 'openshift_http_proxy',
     'proxy_https': 'openshift_https_proxy',
     'proxy_exclude_hosts': 'openshift_no_proxy',
 }
 
+
 def set_config(cfg):
     global CFG
     CFG = cfg
+
 
 def generate_inventory(hosts):
     global CFG
@@ -50,8 +52,7 @@ def generate_inventory(hosts):
 
     write_inventory_vars(base_inventory, multiple_masters, lb)
 
-
-    #write_inventory_hosts
+    # write_inventory_hosts
     for role in CFG.deployment.roles:
         # write group block
         group = ROLES_TO_GROUPS_MAP.get(role, role)
@@ -70,14 +71,16 @@ def generate_inventory(hosts):
     base_inventory.close()
     return base_inventory_path
 
+
 def determine_lb_configuration(hosts):
     lb = next((host for host in hosts if host.is_master_lb()), None)
     if lb:
-        if lb.hostname == None:
+        if lb.hostname is None:
             lb.hostname = lb.connect_to
             lb.public_hostname = lb.connect_to
 
     return lb
+
 
 def write_inventory_children(base_inventory, scaleup):
     global CFG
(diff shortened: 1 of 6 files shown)


tbielawa:
Rebased and squashed down into fewer commits. Also fixed the commit messages for some that were referencing the wrong file names.

I can squash this down more if requested.

utils/src/ooinstall/openshift_ansible.py
```diff
@@ -153,17 +156,17 @@ def write_inventory_vars(base_inventory, multiple_masters, lb):
 def write_proxy_settings(base_inventory):
     try:
         base_inventory.write("openshift_http_proxy={}\n".format(
-                                                            CFG.settings['openshift_http_proxy']))
+            CFG.settings['openshift_http_proxy']))
```
detiber: Any reason to not align on the parenthesis like previous changes?

detiber:
```
         base_inventory.write("openshift_http_proxy={}\n"
                                          .format(CFG.settings['openshift_http_proxy']))
```

tbielawa: tl;dr: I think it's fine just the way it is.

@detiber any of the first three examples in the PEP8 spec section for indentation should be preferred:

https://www.python.org/dev/peps/pep-0008/#indentation

```python
# Aligned with opening delimiter.
foo = long_function_name(var_one, var_two,
                         var_three, var_four)

# More indentation included to distinguish this from the rest.
def long_function_name(
        var_one, var_two, var_three,
        var_four):
    print(var_one)

# Hanging indents should add a level.
foo = long_function_name(
    var_one, var_two,
    var_three, var_four)
```

I wrote up this:
```python
    try:
        base_inventory.write("openshift_http_proxy={}\n".format(
            CFG.settings['openshift_http_proxy']))
```

and you're suggesting this:

```python
    try:
         base_inventory.write("openshift_http_proxy={}\n"
                              .format(CFG.settings['openshift_http_proxy']))
```

In my approach, *technically*, I think there shouldn't be any arguments on the first line. Quoth the PEP:

> When using a hanging indent the following should be considered; there should be no arguments on the first line

However, since both of these pass the PEP8 test I favor my original suggestion because it is more compact visually and the alternative looks more confusing:

```python
    try:
        base_inventory.write(
            "openshift_http_proxy={}\n".format(
                CFG.settings['openshift_http_proxy']))
```

And as always, not all rules are hard and fast:

> know when to be inconsistent -- sometimes style guide recommendations just aren't applicable. When in doubt, use your best judgment ... Some other good reasons to ignore a particular guideline ... When applying the guideline would make the code less readable, even for someone who is used to reading code that follows this PEP.

https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds

detiber: I'm not really partial either way, just noticed the general inconsistency with the other changes being made which favored aligning on the parenthesis.

tbielawa: I just realized this. I think this is two different things we're discussing that's causing the perceived inconsistency.

The other indentation changes you're thinking of are for functions with **multiple** parameters provided, or a **single** parameter *spanning multiple lines*. See the `collect_storage_host` function change in cli_installer.py, ~line 356:

```python
    # ...
    hostname_or_ip = click.prompt('Enter hostname or IP address',
                                  value_proc=validate_prompt_hostname,
                                  # ... more params)

```

In that case we **do** align the parameters up with the opening paren's.

Whereas in the line you pointed out in your original comment the `write` method for the `base_inventory` variable is only given 1 parameter, a formatted string:

```python
    base_inventory.write("openshift_http_proxy={}\n".format(CFG.settings['openshift_http_proxy']))
```

If we aligned the line continuation with the opening `write` method paren we would be visually hinting to the code reader that what they're seeing is another parameter.

Instead, we do a line-continuation style indent for the arguments to the `format` function, indicating this is not another parameter, just a continuation of the preceding line.

tbielawa:
aos-ci-test

tbielawa:
@sdodson @detiber unit tests passed, good to merge :+1: ?

sdodson:
As a python newb, I defer LGTM to someone else.

utils/src/ooinstall/openshift_ansible.py
detiber: Do you want to apply that to lines 134 and 137 as well, those where the lines that I was comparing to from the original comment.

tbielawa:
@detiber @sdodson  A fix for Jason's comment about lines 134 and 137 has been added. This is ready 2 merge.

PR #2364, opened 2016-08-25
Remove openshift_env cloudprovider facts.

### LINKED ISSUE #2349: Templated variables not being substituted in facts
I have been trying to pass environment variables through to vars using the lookup env function like this:

```
openshift_cloudprovider_aws_access_key: "{{ lookup('env','AWS_ACCESS_KEY_ID') }}"
openshift_cloudprovider_aws_secret_key: "{{ lookup('env','AWS_SECRET_ACCESS_KEY') }}"
```

which is mentioned in the example byo hosts file. I've tried it in group_vars/all, hosts and direct in vars passed to a playbook.

A simple debug playbook shows that these variables are being correctly set:

```- debug: var=openshift_cloudprovider_aws_access_key```

However when run with the openshift playbooks, the variables end up like this:

```{# lookup('env','AWS_ACCESS_KEY_ID') #}```

From what I've been able to work out, this is being caused by the openshift_facts module merging unparsed template values over the top of the hostvars (which do contain the correct values):

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_cloud_provider/tasks/main.yml#L6

Looking at the related function here:

https://github.com/openshift/openshift-ansible/blob/master/filter_plugins/oo_filters.py#L83

it seems to be merging the play vars with the host specific section of the hostvars, and the play vars take priority. I'm guessing that this was introduced as a way of supporting multiple versions of ansible, but it seems a bit odd and from what I can see is not fully working. My understanding is that ansible versions prior to 2.0 are not supported, so is this code obsolete?

I've reproduced the problem on latest master, and both ansible v2.1 and v2.2.

abutcher:
Fixes #2349
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1342028


abutcher:
ok to test

abutcher:
ok to test

detiber:
:+1:

abutcher:
aos-ci-test

PR #2366, opened 2016-08-26
add run_once to repeatable actions

Mangirdaz:
These actions need to be done only once on any master. No need to run multiple times.


detiber:
ok to test

detiber:
@Mangirdaz I'm going to go ahead and merge this, however metrics are now able to be installed as part of the install. @sdodson can you provide a link to more info.

Mangirdaz:
I know.. Will submit other commits to fix install as soon as I be able to confirm on my end

PR #2367, opened 2016-08-26
Fix PyLint errors discovered when upgrading to newer version

tbielawa:
Fix PyLint errors discovered when upgrading to newer version
    
    * Fixes PyLint to run in the virtualenv used for all tests
    
    * Replaced 'LooseVersion' with 'parse_version' from setuptools
    - This is a work around for the issue in
      https://github.com/PyCQA/pylint/issues/73 in which pylint can not
      import disutils.version correctly in a virtualenv.
    
    * Removed the unused function 'delete_hosts' which was causing a
      pylint error as well
    
    * Removed a deprecated pylint pragma option, 'bad-builtin'
    
    * Fixed some import ordering issues it was picky about
    
    * Added another disable for a case where the PyLint suggestion would
      have us altering the container we would be iterating over
    
    * Add code-coverage reports to the unittests with the MINIMUM coverage
      percentage for success set to 70%
    - Current test coverage is at 76%


abutcher:
ok to test

abutcher:
lgtm

tbielawa:
To anyone reading over this PR later: This PR introduced the `make ci` testing system to the jenkins worker.

For this specific PR we made a one-off change and switched the Jenkins **default** test from running `git/pylint.sh` to running the `make ci` target in the `utils` directory. The first attempt failed in c959f9d, but after a little debugging by @abutcher and I we got it working with results *consistent* across Jenkins and local workstations in 615b42f

sdodson:
aos-ci-test

tbielawa:
aos-ci-test

tbielawa:
aos-ci-test

utils/Makefile
```diff
@@ -35,29 +35,31 @@ clean:
...
 
+viewcover:
+	xdg-open cover/index.html
+
 virtualenv:
 	@echo "#############################################"
 	@echo "# Creating a virtualenv"
 	@echo "#############################################"
 	virtualenv $(NAME)env
+	. $(NAME)env/bin/activate && pip install setuptools==17.1.1
```
detiber: any reason for the specific version of setuptools?

utils/requirements.txt
```diff
@@ -1 +1,11 @@
-
+enum
+configparser
+pylint
+pep8
+nose
+coverage
+mock
+flake8
```
detiber: shouldn't the test dependencies be in a separate requirements file?

utils/src/ooinstall/cli_installer.py
```diff
@@ -920,7 +902,7 @@ def uninstall(ctx):
 @click.option('--latest-minor', '-l', is_flag=True, default=False)
 @click.option('--next-major', '-n', is_flag=True, default=False)
 @click.pass_context
-# pylint: disable=bad-builtin,too-many-statements
+# pylint: disable=too-many-statements
```
detiber: Should we specify a specific version of pylint to avoid newer versions changing this type of behavior on us unexpectedly?

utils/requirements.txt
detiber: Also, it looks like the install_requires and the test_requires in setup.py no longer matches the list here.

utils/Makefile
abutcher: This is the current version on RHEL 7.2, I believe.

utils/src/ooinstall/cli_installer.py
abutcher: Each test run will install latest so everyone (including jenkins) will test with $latest.

abutcher: and the option deprecations will be logged so we can at least track those changing.

utils/Makefile
detiber: Just seems odd, unless it is working around a specific bug.

abutcher: The version changes below (from previous commit) were using `setuptools.version` which isn't available in the setuptools version on RHEL so we now use `pkg_resources import parse_version`. I think we'll want the local tests to line up when we notice things like this.

abutcher: However, now that we've worked around it this version lock isn't really necessary.

utils/src/ooinstall/oo_config.py
```diff
@@ -325,6 +326,10 @@ def _set_defaults(self):
         self.settings['ansible_inventory_path'] = \
             '{}/hosts'.format(os.path.dirname(self.config_path))
 
+        # pylint: disable=consider-iterating-dictionary
+        # Disabled because we shouldn't alter the container we're
+        # iterating over
+        #
         # clean up any empty sets
         for setting in self.settings.keys():
             if not self.settings[setting]:
```
detiber: @tbielawa @smunilla: not related to this PR, but this doesn't seem right to me...  It also looks like we are removing the ansible_ssh_user default we are setting above here, since an empty string is false...  It also seems to exclude the possibility of adding keys with boolean values.

utils/src/ooinstall/oo_config.py
```diff
@@ -325,6 +326,10 @@ def _set_defaults(self):
         self.settings['ansible_inventory_path'] = \
             '{}/hosts'.format(os.path.dirname(self.config_path))
 
+        # pylint: disable=consider-iterating-dictionary
+        # Disabled because we shouldn't alter the container we're
+        # iterating over
+        #
         # clean up any empty sets
         for setting in self.settings.keys():
```
detiber: In python3 this would return an iterator. I haven't checked out python3 compatibility and not sure we even care, but this is a potential issue there.

utils/Makefile
detiber: Yeah, at some point we'll need to make sure we get fedora-based coverage for this as well.

sdodson:
@tbielawa does this supersede #1949

tbielawa:
@sdodson it could supersede that PR if I change it to have two different requirements.txt files. One including testing deps, and one sans-deps.

tbielawa:
> any reason for the specific version of setuptools?

Yes, so that local tests are 
> Should we specify a specific version of pylint to avoid newer versions changing this type of behavior on us unexpectedly?

Answered by @abutcher 

>

tbielawa:
> any reason for the specific version of setuptools?

Yes, so that local tests run with versions as close to the upstream test platform as possible. In a previous commit we ran into errors on Jenkins because it was coded against functionality I found in the setuptools version installed on my local machine. That functionality wasn't available on the upstream jenkins.


> shouldn't the test dependencies be in a separate requirements file?

I think that's a good idea. I'll update the PR with that.

> Also, it looks like the install_requires and the test_requires in setup.py no longer matches the list here.

Yup. `setup.py` has been out of date for a while. The install/testing docs were out of date as well until I fixed them in an older PR. I **could** try to update `tests_require=` in setup.py to read from `test-requirements.txt` but I think that's only used for projects with `setup.py` integrated with a test runner. That is to say, we list `tests_require` in our setup.py, but we **don't** have functionality written to exercise tests from the setup file. So that line is kind of extra and unused right now.

> Should we specify a specific version of pylint to avoid newer versions changing this type of behavior on us unexpectedly?

Answered by @abutcher 

> @tbielawa @smunilla: not related to this PR, but this doesn't seem right to me... It also looks like we are removing the ansible_ssh_user default we are setting above here, since an empty string is false... It also seems to exclude the possibility of adding keys with boolean values.

Yep. That's unrelated to this PR.

> In python3 this would return an iterator. I haven't checked out python3 compatibility and not sure we even care, but this is a potential issue there.

I made no code changes here, I simply added a comment explaining why I disabled the pylint warning. Not going to make any more changes

> @tbielawa does this supersede #1949 

On second thought, no this doesn't supersede #1949. That person wanted a completely versioned-pinned requirements file so they could run ansible in a virtual env. This PR is only modifying test requirements and won't be creating a `requirements.txt` file as was submitted in the referenced PR.

tbielawa:
ok to test

tbielawa:
aos-ci-test

arilivigni:
aos-ci-test

tbielawa:
aos-ci-test

PR #2368, opened 2016-08-26
Preserve AWS options in sysconfig files.

dgoodwin:
These could exist from 3.1 releases or manual edits, and were getting
wiped out in all variants of the master sysconfig files as we template
these.

This change uses the established pattern of grepping them out if they're
there, then reapplying them if the user has not configured the
cloudprovider framework already.

I have only tested the non-HA use case by manually adding the vars and then running a 3.3 upgrade.

### DIFF
--- a/roles/openshift_master/tasks/systemd_units.yml
+++ b/roles/openshift_master/tasks/systemd_units.yml
@@ -45,6 +45,13 @@
   failed_when: false
   changed_when: false
 
+- name: Preserve Master API AWS options
+  command: grep AWS_ /etc/sysconfig/{{ openshift.common.service_type }}-master-api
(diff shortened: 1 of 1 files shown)


dgoodwin:
I did get an HA env up, upgrade is running now after I embedded these vars into all 4 service files, including node.

abutcher:
ok to test

sdodson:
LGTM, this code path isn't tested by CI so going ahead with merge.

sdodson:
well, actually it is since it's not upgrade only.

sdodson:
aos-ci-test

dgoodwin:
HA test passed, preserved in all files.

PR #2375, opened 2016-08-29
Fix flannel check

mkumatag:
As ovs is enabled by default and will get installed all the rpms and configuration enabled even though flannel is enabled. So this PR is for not installing and configuring if flannel enabled. Enhanced the info message.

### DIFF
--- a/roles/openshift_common/tasks/main.yml
+++ b/roles/openshift_common/tasks/main.yml
@@ -1,7 +1,7 @@
 ---
 - fail:
-    msg: Flannel can not be used with openshift sdn
-  when: openshift_use_openshift_sdn | default(false) | bool and openshift_use_flannel | default(false) | bool
+    msg: Flannel can not be used with openshift sdn, set openshift_use_openshift_sdn=false if you want to use flannel
+  when: openshift_use_openshift_sdn | default(true) | bool and openshift_use_flannel | default(false) | bool
 
 - fail:
    msg: Nuage sdn can not be used with openshift sdn


roles/openshift_node/tasks/main.yml
```diff
@@ -36,7 +36,7 @@
 
 - name: Install sdn-ovs package
   action: "{{ ansible_pkg_mgr }} name={{ openshift.common.service_type }}-sdn-ovs{{ openshift_pkg_version | oo_image_tag_to_rpm_version(include_dash=True) }} state=present"
-  when: openshift.common.use_openshift_sdn and not openshift.common.is_containerized | bool
+  when: openshift.common.use_openshift_sdn and not openshift.common.use_flannel | bool and not openshift.common.is_containerized | bool
```
detiber: @abutcher ptal. If we do this for flannel, we would also have to do it for the nuage flag, and I'm not sure this creates the best interface. It might be confusing for a user to know which "use_<network plugin>" overrides the others.

abutcher: @detiber Right, which is why we currently require setting both `openshift_use_openshift_sdn=false` and `openshift_use_flannel=true`. What if we approached this as an "sdn kind" sort of variable -- defaults to openshift, can be set to nuage, flannel or X?

mkumatag: @abutcher its always hard to remember what all options we need to set to make it work. To make it more friendly can we have one check in `roles/openshift_common/tasks/main.yml` to display a message saying make `openshift_use_openshift_sdn=false if you want to use flannel` or something?

abutcher: @mkumatag That sounds great. It looks like the checks that exist within `openshift_common` need to reference variable names in their messages, correct?

detiber: @abutcher with the changes coming down, I don't know if it makes sense to refactor this now...  that said, we do need to make sure that we are sanity checking the values...  I thought we were already doing this.

abutcher: @detiber We're sanity checking the values but we aren't saying, "this variable needs to be set".

mkumatag: this is sanity checking code I see in openshift_common
```
- fail:
    msg: Flannel can not be used with openshift sdn
  when: openshift_use_openshift_sdn | default(false) | bool and openshift_use_flannel | default(false) | bool
```

Is this skipping the check because we are feeding the default value for `openshift_use_openshift_sdn` as false?

abutcher: @mkumatag Yep, that is skipping because we assume it is false. Nice catch!

mkumatag: Let me make it true and add more information.

mkumatag:
@detiber @abutcher corrected the check in openshift_common and enhanced the message as well. Please review.

abutcher:
LGTM

abutcher:
ok to test

abutcher:
aos-ci-test

PR #2377, opened 2016-08-29
Default to port 80 when deploying cockpit-ui

smunilla:
This should save first-time users some pain.

Fixes BZ#1371031


aweiteka:
LGTM

sdodson:
not currently tested by CI, merging without tests.

detiber:
Do we not configure SSL by default? Or are we handling it with an edge termination route?

liggitt:
also, does this mean config will need undoing once the registry is secured?

PR #2378, opened 2016-08-29
Reload units after node container service modified.

dgoodwin:
@sdodson this should do it, the reload makes sense here as we want it in both upgrade and config playbook, it also matches what we're doing for the master systemd_units.yml.

Node container was restarting but no service changes were taking effect due to missing reload.


sdodson:
LGTM

sdodson:
aos-ci-test

sdodson:
aos-ci-test

PR #2384, opened 2016-08-30
Change how we set master's metricsPublicURL

sdodson:
- If we're deploying metrics configure master during master role execution for metrics rather than only in openshift_metrics role
 - Ignore failed metrics deployers -- there's not much we can do to recover from this and the admin will need to debug what went wrong.

Fixes BZ1371179
Fixes BZ1371103

### DIFF
--- a/roles/openshift_master_facts/tasks/main.yml
+++ b/roles/openshift_master_facts/tasks/main.yml
@@ -80,3 +80,4 @@
       api_env_vars: "{{ openshift_master_api_env_vars | default(None) }}"
       controllers_env_vars: "{{ openshift_master_controllers_env_vars | default(None) }}"
       audit_config: "{{ openshift_master_audit_config | default(None) }}"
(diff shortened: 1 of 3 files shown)


sdodson:
@abutcher PTAL

sdodson:
aos-ci-test

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -80,3 +80,4 @@
       api_env_vars: "{{ openshift_master_api_env_vars | default(None) }}"
       controllers_env_vars: "{{ openshift_master_controllers_env_vars | default(None) }}"
       audit_config: "{{ openshift_master_audit_config | default(None) }}"
+      metrics_public_url: "{% if openshift_hosted_metrics_deploy | default(false) %}https://{{ metrics_hostname }}/hawkular/metrics{% endif %}"
```
abutcher: Is `openshift.master.metrics_public_url` used anywhere?

abutcher: Found it disregard.

abutcher:
lgtm

PR #2387, opened 2016-08-30
Reload docker facts after upgrading docker

sdodson:
The containerized node unit files depend on docker version facts so we should update them after we update docker.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1358197


sdodson:
@dgoodwin PTAL

dgoodwin:
LGTM

PR #2396, opened 2016-09-01
Persist net.ipv4.ip_forward sysctl entry for openshift nodes

tbielawa:
Reference:
https://bugzilla.redhat.com/show_bug.cgi?id=1372388


tbielawa:
aos-ci-test

PR #2400, opened 2016-09-02
Fix ops/qps typo

liggitt:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1372638


liggitt:
@detiber @sdodson PTAL

PR #2404, opened 2016-09-02
Logging fixes

sdodson:
Round 2 of improvements

Things fixed
 - Calls the role when openshift_hosted_logging_deploy=true
 - Provisions 1 NFS volume for ES cluster of 1 pod
 - Sets sane defaults for `openshift_hosted_logging_hostname` `openshift_hosted_logging_ops_hostname` `openshift_hsoted_logging_master_public_url` `openshift_hosted_logging_elasticsearch_cluster_size` `openshift_hosted_logging_elasticsearch_ops_cluster_size` `openshift_hosted_logging_elasticsearch_pvc_dynamic` `openshift_hosted_logging_elasticsearch_pvc_size` `openshift_hosted_logging_elasticsearch_pvc_prefix`
 - Basic inventory examples

These are the major gaps that remain as far as I know
 - Does not provision ops cluster nfs volume
 - NFS storage in all cases is limited to one ES pod
 - Currently attempts to re-run everything even if logging has already been deployed, though in my testing this doesn't result in anything other than the deployer having failed and the installation accurately reports overall success or failure.


sdodson:
@richm FYI additional tweaks to logging

richm:
LGTM

When running oo-install - how do you specify that you want to install logging?

sdodson:
> When running oo-install - how do you specify that you want to install logging?

`atomic-openshift-utils` aka "quick-installer"? Currently you'd have to provide it a custom config file and perform an unattended install. My last commit here adds some items to the sample inventories for use via ansible aka "advanced install" not all of them work right now but i'll work towards parity with metrics where it makes sense

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -138,7 +138,7 @@
     delay: 10
 
   - name: "Deploy fluentd by labeling the node"
-    shell:  "{{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig label node {{ openshift_hostname }} {{ openshift_hosted_logging_fluentd_nodeselector if openshift_hosted_logging_fluentd_nodeselector is defined else 'logging-infra-fluentd=true' }}"
+    shell:  "{{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig label node --all {{ openshift_hosted_logging_fluentd_nodeselector_label if openshift_hosted_logging_fluentd_nodeselector_label is defined else 'logging-infra-fluentd=true' }}"
```
sdodson: @richm Thoughts on what makes the most sense here? I think we should make the label for the daemon set immutable, or at least an advanced setting that 99% of users wouldn't touch only if it conflicts with a label they're alread using for something else. But we should add a selector to the determine which nodes we apply that label to. The default would be no selector meaning that all nodes would get fluentd label applied to them.

richm: @sdodson yeah - I think most users won't care what the label is, as long as it isn't problematic @sosiouxme wdyt?

sdodson:
aos-ci-test

richm:
```
TASK [openshift_hosted_logging : fail] *****************************************
fatal: [10.8.168.131]: FAILED! => {"changed": false, "failed": true, "msg": "This role requires the following vars to be defined. openshift_hosted_logging_master_public_url, openshift_hosted_logging_hostname, openshift_hosted_logging_elasticsearch_cluster_size"}
	to retry, use: --limit @/home/jenkins/workspace/openshift-ansible-3-test-matrix/CONTAINERIZED/_containerized/OSE_VER/3.2/PYTHON/System-CPython-2.7/TOPOLOGY/openshift-cluster-containerized/TargetBranch/master/nodes/openshift-ansible-slave/openshift-ansible/playbooks/byo/config.retry
```
How is the test run?  Where can openshift_hosted_logging_master_public_url, openshift_hosted_logging_hostname, openshift_hosted_logging_elasticsearch_cluster_size be defined?

sdodson:
> How is the test run? Where can openshift_hosted_logging_master_public_url, openshift_hosted_logging_hostname, openshift_hosted_logging_elasticsearch_cluster_size be defined?

We don't have a great way to set arbitrary variables for CI runs right now. I'll work on setting default values for those today based on other cluster config values, for elastic search @sosiouxme and I discussed defaulting to one pod.

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
sosiouxme: At this time we don't offer a way for users to have the daemonset without the selector; you get our default if none is specified. I think we want a selector so that admins have a way to undeploy/redeploy fluentd without deleting the daemonset. I don't think (m)any users are going to care about changing the default selector. It seems unlikely to conflict.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
ci jobs are regrettably 3.2 only right now, so enable logging by default in 3.2 for now.

sdodson:
aos-ci-test

sdodson:
This all works locally for me and the only difference is that CI jobs use 3.2 and I'm running 3.3 so I'm going to work on getting 3.3 CI jobs.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

richm:
For aos-ci-jenkins/OS_3.3_NOT_containerized - is there some way I can see the log from the deployer pod?

sdodson:
Yeah, testing a solution to dumping the logs now. Looks like the details link gets truncated, we'll have to fix that, it's supposed to be a link to the same item that gets posted as a comment. As far as the 3.3 containerized install, it's just skipping logging all together, sigh.

sdodson:
aos-ci-test

sdodson:
openshift_master_default_subdomain is null, so kibana-hostname is being set to 'kibana.'. Working on a fix.

sdodson:
aos-ci-test

sdodson:
@detiber or @abutcher mind reviewing this. This is the last of the logging work i intend to complete before the rewrite/refactor previously discussed.

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -138,7 +138,7 @@
     delay: 10
 
   - name: "Deploy fluentd by labeling the node"
-    shell:  "{{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig label node {{ openshift_hostname }} {{ openshift_hosted_logging_fluentd_nodeselector if openshift_hosted_logging_fluentd_nodeselector is defined else 'logging-infra-fluentd=true' }}"
+    shell:  "{{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig label node --overwrite=true {{ '-l openshift_hosted_logging_fluentd_nodeselector' if openshift_hosted_logging_fluentd_nodeselector is defined else '--all' }} {{ openshift_hosted_logging_fluentd_nodeselector_label if openshift_hosted_logging_fluentd_nodeselector_label is defined else 'logging-infra-fluentd=true' }}"
```
sdodson: @enoodle what about this rather than #2516 ?

enoodle: I will close #2516 then.

playbooks/common/openshift-cluster/openshift_hosted.yml
```diff
@@ -20,8 +20,8 @@
       openshift_hosted_registry_registryurl: "{{ hostvars[groups.oo_first_master.0].openshift.master.registry_url }}"
     when: "'master' in hostvars[groups.oo_first_master.0].openshift and 'registry_url' in hostvars[groups.oo_first_master.0].openshift.master"
   - set_fact:
-      logging_hostname: "{{ openshift_hosted_logging_hostname | default('kibana.' ~ openshift_master_default_subdomain) }}"
-      logging_ops_hostname: "{{ openshift_hosted_logging_ops_hostname | default('kibana-ops.' ~ openshift_master_default_subdomain) }}"
+      logging_hostname: "{{ openshift_hosted_logging_hostname | default('kibana.' ~ (openshift.master.default_subdomain | default('router.default.svc.cluster.local', true))) }}"
+      logging_ops_hostname: "{{ openshift_hosted_logging_ops_hostname | default('kibana-ops.' ~ (openshift.master.default_subdomain | default('router.default.svc.cluster.local', true))) }}"
```
richm: The kibana hostname must be something that can be resolved externally, outside of the cluster.

playbooks/common/openshift-cluster/openshift_hosted.yml
```diff
@@ -57,6 +57,10 @@
     openshift_hosted_logging_master_public_url: "{{ logging_master_public_url }}"
     openshift_hosted_logging_elasticsearch_cluster_size: "{{ logging_elasticsearch_cluster_size }}"
     openshift_hosted_logging_elasticsearch_ops_cluster_size: "{{ logging_elasticsearch_ops_cluster_size }}"
+    openshift_hosted_logging_elasticsearch_pvc_dynamic: "{{ 'true' if openshift.hosted.logging.storage_kind | default(none) == 'dynamic' else 'false' }}"
+    openshift_hosted_logging_elasticsearch_pvc_size: "{{ openshift.hosted.logging.storage.volume.size if openshift.hosted.logging.storage_kind | default(none) == 'dynamic' else ''  }}"
+    openshift_hosted_logging_elasticsearch_pvc_prefix: "{{ 'logging-es' if openshift.hosted.logging.storage_kind | default(none) is not none else '' }}"
```
richm: Do you need corresponding `ops` parameters for the ops cluster e.g. `openshift_hosted_logging_elasticsearch_ops_pvc_dynamic` ?

playbooks/common/openshift-cluster/openshift_hosted.yml
sdodson: Yeah, I was surprised to find that's actually the default routing subdomain if one is not set. So while it may not resolve outside of the cluster I don't know how to come up with a better default.

playbooks/common/openshift-cluster/openshift_hosted.yml
sdodson: Yeah, I'll do that.

playbooks/common/openshift-cluster/openshift_hosted.yml
richm: yeah, not sure what else we could use for a default value
LGTM

Issue #2401, opened 2016-09-02
Deploy PersistentVolume definitions FAILS


deromka:
```
TASK [roles/openshift_persistent_volumes : Deploy PersistentVolume definitions] 
fatal: [10.57.50.187]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleFilterError: Failed to convert: %s"}
```

Here is the call to PersistentVolume role:

```
- name: Create persistent volumes/claims
  hosts: masters[0]
  vars:
    openshift_deployment_type: origin
    persistent_volumes:
    - name: "persistent-log-storage"
      capacity: "10Gi"
      access_modes:
      - "ReadWriteMany"
      storage:
        nfs:
          server: "www.server.com"
          path: "/drm-persistent-storage"
    persistent_volume_claims:
    - name: "persistent-log-storage-claim"
      capacity: "5Gi"
      access_modes:
      - "ReadWriteMany"
  roles:
  - role: roles/openshift_persistent_volumes`
```

TAG: openshift-ansible-3.3.16-1
Destination machine: CentOS Linux release 7.2.1511 (Core)
3.10.0-327.28.2.el7.x86_64 #1 SMP

tbielawa:
@deromka I can't tell what's wrong from the output you provided (though I do appreciate it!)

> FAILED! => {"changed": false, "failed": true, "msg": "AnsibleFilterError: Failed to convert: %s"}

You did just help me uncover an error in our filter plugin. See that `%s` in there? That **should** be the error message that was raised when the Filter Error happened, but we have a typo in our filter module

We're doing this:
> raise errors.AnsibleFilterError('Failed to convert: %s', my_e)

However, according to the Ansible source code:

> def __init__(self, message="", obj=None, show_content=True, suppress_extended_error=False): 

the second parameter to the `AnsibleFilterError` exception should be an `obj` which is *"some subclass of ansible.parsing.yaml.objects.AnsibleBaseYAMLObject, which should be returned by the DataLoader() class"*.

What we should be doing is this:
> raise errors.AnsibleFilterError('Failed to convert: %s' % my_e)

So that the string substitution actually happens. I'm going to file a PR for this error and then ask you to test again.

In the mean time, can you try applying this patch manually and re-running your playbook?

https://gist.github.com/tbielawa/73a41214cb02954f54bfc0858fa2ecb8

You can apply it in one command like this:

[...]

PR #2405, opened 2016-09-02
Fix string substitution error in the to_padded_yaml filter

tbielawa:
Discovered in issue **Deploy PersistentVolume definitions FAILS** #2401

Failed calls to the `to_padded_yaml` function will not print the actual exception message text. This is because the string substitution which would show the error is not actually happening.

Patch fixes call to `errors.AnsibleFilterError` so that the original exception text is properly substituted into the raised Ansible exception.


tbielawa:
aos-ci-test

abutcher:
lgtm

PR #2406, opened 2016-09-02
Record schedulability of node prior to upgrade and re-set it to that

sdodson:
Previously we were setting schedulability to the state defined in the inventory
without regard to whether or not it was manually made schedulable or
unschedulable. The right thing seems to be to record the state prior to upgrade
and set it back.

Also, the bug reports that masters were being set schedulable when they shouldn't have been but I couldn't replicate that behavior but I think this would fix that too.

Based on https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/redeploy-certificates.yml#L202-L236
Fixes Bug 1372594

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade.yml
@@ -173,6 +173,18 @@
   # TODO: To better handle re-trying failed upgrades, it would be nice to check if the node
   # or docker actually needs an upgrade before proceeding. Perhaps best to save this until
   # we merge upgrade functionality into the base roles and a normal config.yml playbook run.
+  - name: Determine if node is currently scheduleable
+    command: >
+      {{ openshift.common.client_binary }} get node {{ openshift.common.hostname | lower }} -o json
+    register: node_output
+    delegate_to: "{{ groups.oo_first_master.0 }}"
+    changed_when: false
+    when: inventory_hostname in groups.oo_nodes_to_config
+
+  - set_fact:
+      was_schedulable: "{{ 'unschedulable' not in (node_output.stdout | from_json).spec }}"
+    when: inventory_hostname in groups.oo_nodes_to_config
+
   - name: Mark unschedulable if host is a node
     command: >
       {{ openshift.common.admin_binary }} manage-node {{ openshift.common.hostname | lower }} --schedulable=false
@@ -203,7 +215,7 @@
     command: >
       {{ openshift.common.admin_binary }} manage-node {{ openshift.common.hostname | lower }} --schedulable=true
     delegate_to: "{{ groups.oo_first_master.0 }}"
-    when: inventory_hostname in groups.oo_nodes_to_config and openshift.node.schedulable | bool
+    when: inventory_hostname in groups.oo_nodes_to_config and was_schedulable | bool
 
 
 ##############################################################################


sdodson:
@abutcher @dgoodwin PTAL

playbooks/common/openshift-cluster/upgrades/upgrade.yml
```diff
@@ -203,7 +215,7 @@
     command: >
       {{ openshift.common.admin_binary }} manage-node {{ openshift.common.hostname | lower }} --schedulable=true
     delegate_to: "{{ groups.oo_first_master.0 }}"
-    when: inventory_hostname in groups.oo_nodes_to_config and openshift.node.schedulable | bool
```
dgoodwin: This was coming from openshift.node.schedulable, am I correct that this is from openshift_facts and accurately reflects what the system was when facts were last loaded? (provided we don't reload facts in the middle of this, which I can't see anywhere but is a little bit of a risk)

If this is correct you could likely simplify the check, was_schedulable = openshift.node.schedulable.

Or I might be missing something here.

dgoodwin:
My understanding, openshift.node.schedulable is coming from openshift_facts, and provided we don't explicitly reload that in the middle of this upgrade block, this *should* be working. It seems to be for me as well, I can't really reproduce the bug yet, not sure why.

This might be safer though given that "don't reload facts" assumption.

playbooks/common/openshift-cluster/upgrades/upgrade.yml
abutcher: I think that fact is representative of what is configured in the inventory but does not represent current state of the node.

sdodson:
Yeah, that's right, the problem probably comes from reloading facts at the wrong time. I think setting everything back to where it was when we started is the least surprising and we can avoid problems with openshift_facts.

playbooks/common/openshift-cluster/upgrades/upgrade.yml
```diff
@@ -173,6 +173,18 @@
   # TODO: To better handle re-trying failed upgrades, it would be nice to check if the node
   # or docker actually needs an upgrade before proceeding. Perhaps best to save this until
   # we merge upgrade functionality into the base roles and a normal config.yml playbook run.
+  - name: Determine if node is currently scheduleable
+    command: >
+      {{ openshift.common.client_binary }} get node {{ openshift.common.hostname | lower }} -o json
+    register: node_output
+    delegate_to: "{{ groups.oo_first_master.0 }}"
+    changed_when: false
+    when: inventory_hostname in groups.oo_nodes_to_config
```
dgoodwin: Ok with this in mind, should we simplify down to just grabbing the value of openshift.node.schedulable at the time we enter this block?

dgoodwin:
Sounds good, were you able to reproduce @sdodson ? If not there's a very very slight possibility we don't have the cause pinned down yet.

playbooks/common/openshift-cluster/upgrades/upgrade.yml
dgoodwin: Doesn't appear to be that way, I have no openshift_schedulable set in inventory on any hosts and I'm getting debug's on openshift.node.schedulable coming out correct.

abutcher: After manually changing one to unschedulable? I guess for the test we'd want to run install, manually change one node to unscheduleable and then run upgrade.

playbooks/common/openshift-cluster/upgrades/upgrade.yml
dgoodwin: Ok definitely do not do what I said. openshift.node.schedulable looks like it is coming from openshift_facts but *that* is not reading anything from the system, it looks like it's just assuming master = false, node = true. True state of the node is not considered so your check here is much safer/better.

playbooks/common/openshift-cluster/upgrades/upgrade.yml
dgoodwin: Yeah that's probably the part I'm missing, I was watching master (although that's what the bug reports on), I think I can see how it would happen if you move a node to unschedulable. 

I'm genuinely not sure how masters are winding up schedulable in his example...

Still doing some testing.

sdodson: > After manually changing one to unschedulable? I guess for the test we'd want to run install, manually change one node to unscheduleable and then run upgrade.

I agree that would definitely be re-set to schedulable, though I don't think that's what's described in the bug.

So, in short, I can't reproduce what they're seeing in the bug but I think this is the right thing to do and I hope we can convince QE of that too.

dgoodwin: This change LGTM to me and fixes *a* bug. I'm not 100% sure we've got the bug in the report, I'm going to keep experimenting looking for a way to flip a master schedulable after an upgrade.

sdodson: Maybe they had overriden the master default schedulability in their inventory, then manually set it unschedulable, and were expecting that to be preserved. I'll follow up on the bug asking for more info and explaining what we've done here.

dgoodwin:
Some testing results without this fix:

```
(root@m1 ~) $ oc version
oc v3.2.1.13-1-gc2a90e1
kubernetes v1.2.0-36-g4a3f9c5
(root@m1 ~) $ oc get nodes
NAME                 STATUS                     AGE
m1.aos.example.com   Ready,SchedulingDisabled   3m
n2.aos.example.com   Ready                      3m
(root@m1 ~) $ oadm manage-node m1.aos.example.com --schedulable=true
NAME                 STATUS    AGE
m1.aos.example.com   Ready     3m
(root@m1 ~) $ oadm manage-node n2.aos.example.com --schedulable=false                                                                                                                 
NAME                 STATUS                     AGE
n2.aos.example.com   Ready,SchedulingDisabled   3m
(root@m1 ~) $ oc get nodes
NAME                 STATUS                     AGE
m1.aos.example.com   Ready                      3m
n2.aos.example.com   Ready,SchedulingDisabled   3m
(root@m1 ~) $ oc version
oc v3.3.0.29
kubernetes v1.3.0+52492b4
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://m1.aos.example.com:8443
openshift v3.3.0.29
kubernetes v1.3.0+52492b4
(root@m1 ~) $ oc get nodes
NAME                 STATUS                     AGE
m1.aos.example.com   Ready,SchedulingDisabled   14m
n2.aos.example.com   Ready                      14m
```

So openshift_facts is being weird and using openshift.node.schedulable is a bad idea.
Testing with your fix now.

dgoodwin:
Status is preserved with your change @sdodson. LGTM.

PR #2413, opened 2016-09-06
Fix for objects which can't be converted by the YAML parser

### LINKED ISSUE #2401: Deploy PersistentVolume definitions FAILS

tbielawa:
* Ansible parsed vars into special Ansible helper objects
* Manually cast each var into a normal string as applicable

Closes #2401

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -16,6 +16,7 @@ import pkg_resources
 import re
 import json
 import yaml
+from ansible.parsing.yaml.dumper import AnsibleDumper
 from ansible.utils.unicode import to_unicode
 from urlparse import urlparse
 
@@ -621,7 +622,9 @@ class FilterModule(object):
             return ""
 
         try:
-            transformed = yaml.safe_dump(data, indent=indent, allow_unicode=True, default_flow_style=False, **kw)
+            transformed = yaml.dump(data, indent=indent, allow_unicode=True,
+                                    default_flow_style=False,
+                                    Dumper=AnsibleDumper, **kw)
             padded = "\n".join([" " * level * indent + line for line in transformed.splitlines()])
             return to_unicode("\n{0}".format(padded))
         except Exception as my_e:


tbielawa:
aos-ci-test

filter_plugins/oo_filters.py
```diff
@@ -621,11 +621,23 @@ def to_padded_yaml(data, level=0, indent=2, **kw):
...
+                for d in data:
+                    _data.append(unicode(d))
+            elif isinstance(data, dict):
+                # Assumes all values are simple strings
+                _data = {}
+                for k, v in data.iteritems():
+                    _data[unicode(k)] = unicode(v)
+            else:
+                _data = data
+
```
tbielawa: After sleeping on this, I wonder if it makes sense to generalize this logic into some kind of utility function? It might end up being reusable in the future in other template situations.

abutcher: I think this needs to be done recursively. I tested with some options that get passed through the `to_padded_yaml` filter.

```ini
openshift_node_kubelet_args={'max-pods': ['40'], 'image-gc-high-threshold': ['90'], 'image-gc-low-threshold': ['80']}
osm_controller_args={'resource-quota-sync-period': ['10s']}
osm_api_server_args={'max-requests-inflight': ['400']}
```

Lists nested within dictionaries are added as strings to the master config.
```yaml
  apiServerArguments:
    max-requests-inflight: '[u''400'']'
  controllerArguments:
    resource-quota-sync-period: '[u''10s'']'
```

detiber: I commented on the linked issue, but can use AnsibleDumper from lib/ansible/parsing/yaml/dumper.py

tbielawa:
aos-ci-test

tbielawa:
As noted in the updated branch:

* Previously we used `yaml.safe_dump`
* Now we use `yamp.dump` with the `Dumper` parameter set to the `AnsibleDumper` class.
* `AnsibleDumper` subclasses `yaml.SafeDumper`, so we aren't losing any safety nets

This addresses the point made by @abutcher

> I think this needs to be done recursively. I tested with some options that get passed through the to_padded_yaml filter.

 regarding nested datastructures which would have required recursion to fix in the previously proposed patch.

filter_plugins/oo_filters.py
tbielawa: Addressed in updated commit.

tbielawa:
aos-ci-test

tbielawa:
@abutcher @detiber PTAL.

detiber:
lgtm

PR #2414, opened 2016-09-07
commented roles on installing docker

leegonz:


sdodson:
@leegonz what are you trying to achieve here?

detiber:
@leegonz closing this, since it looks like a development branch and not a proposed change.

PR #2416, opened 2016-09-07
Use vars instead of facts for openshift.hosted.registry.storage.s3

### LINKED ISSUE #2415: Registry variables not being interpolated in config template
A bug I encountered with cloudprovider variables not being interpolated properly (#2349) also affects the docker registry config among other things. Setting a variable like this:

openshift_hosted_registry_storage_s3_accesskey: "{{ openshift_cloudprovider_aws_access_key }}"

Results in the string:

{# openshift_cloudprovider_aws_access_key #}

being added to the docker registry config.

Replacing the facts with variables in the template fixes the problem.

jam13:
Use vars instead of facts for openshift.hosted.registry.storage.s3 to allow variable interpolation. Resolves #2415.


abutcher:
ok to test

abutcher:
aos-ci-test

roles/openshift_hosted/templates/registry_config.j2
```diff
@@ -66,13 +66,13 @@ middleware:
   - name: openshift
     options:
       pullthrough: {{ openshift.hosted.registry.pullthrough | default(true) }}
-{% if openshift.hosted.registry.storage.provider == 's3' and 'cloudfront' in openshift.hosted.registry.storage.s3 %}
+{% if openshift.hosted.registry.storage.provider == 's3' and openshift_hosted_registry_storage_s3_cloudfront is defined %}
```
abutcher: `openshift_hosted_registry_storage_s3_cloudfront` will never be defined on its own so we should probably check for `openshift_hosted_registry_storage_s3_cloudfront_baseurl` as an alternative for now.

We can remove the fact for `openshift.hosted.registry.storage.provider` as long as we define a default for the comparison.

```
{% if openshift_hosted_registry_storage_provider | default('') == 's3' and openshift_hosted_registry_storage_s3_cloudfront_baseurl is defined %}
```

detiber: We should probably add a comment that *when* re-configuration is supported that we'll need to introspect these values from the running config if the variables are not defined.

sdodson:
@jam13 i think these changes have already been merged in another PR, should this be closed?

jam13:
@sdodson yes, it looks like it's obsolete. Thanks.

PR #2417, opened 2016-09-07
Bug 1327409 - scaleup playbook uses current oc login which may not have enough permissions

abutcher:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1327409

Added an extra commit to avoid looping over entire hostvar blocks when setting node schedulability.

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -159,7 +159,6 @@
     openshift_nodes: "{{ hostvars
                          | oo_select_keys(groups['oo_nodes_to_config'])
                          | oo_collect('openshift.common.hostname') }}"
-    openshift_node_vars: "{{ hostvars | oo_select_keys(groups['oo_nodes_to_config']) }}"
   pre_tasks:
   # Necessary because when you're on a node that's also a master the master will be
   # restarted after the node restarts docker and it will take up to 60 seconds for
--- a/roles/openshift_manage_node/tasks/main.yml
+++ b/roles/openshift_manage_node/tasks/main.yml
@@ -1,6 +1,22 @@
+---
+- name: Create temp directory for kubeconfig
+  command: mktemp -d /tmp/openshift-ansible-XXXXXX
+  register: mktemp
+  changed_when: False
+
+- set_fact:
+    openshift_manage_node_kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
+
+- name: Copy the admin client config(s)
+  command: >
+    cp {{ openshift.common.config_base }}/master/admin.kubeconfig {{ openshift_manage_node_kubeconfig }}
+  changed_when: False
+
 - name: Wait for Node Registration
   command: >
-      {{ openshift.common.client_binary }} get node {{ item | lower }}
+    {{ openshift.common.client_binary }} get node {{ item | lower }}
+    --config={{ openshift_manage_node_kubeconfig }}
+    -n default
   register: omd_get_node
   until: omd_get_node.rc == 0
   retries: 50
@@ -11,6 +27,8 @@
 - name: Set node schedulability
   command: >
     {{ openshift.common.admin_binary }} manage-node {{ item.openshift.common.hostname | lower }} --schedulable={{ 'true' if item.openshift.node.schedulable | bool else 'false' }}
+    --config={{ openshift_manage_node_kubeconfig }}
+    -n default
   with_items:
     -  "{{ openshift_node_vars }}"
   when: item.openshift.common.hostname is defined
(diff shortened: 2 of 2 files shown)


abutcher:
aos-ci-test

abutcher:
@tbielawa @sdodson PTAL

sdodson:
LGTM

playbooks/common/openshift-node/config.yml
```diff
@@ -159,7 +159,6 @@
     openshift_nodes: "{{ hostvars
                          | oo_select_keys(groups['oo_nodes_to_config'])
                          | oo_collect('openshift.common.hostname') }}"
-    openshift_node_vars: "{{ hostvars | oo_select_keys(groups['oo_nodes_to_config']) }}"
```
tbielawa: After seeing the visual effects of this, I don't think there are enough `+`'s for how much I want to :+1:  this.

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -1,6 +1,22 @@
...
+  changed_when: False
+
+- set_fact:
+    openshift_manage_node_kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
+
+- name: Copy the admin client config(s)
+  command: >
+    cp {{ openshift.common.config_base }}/master/admin.kubeconfig {{ openshift_manage_node_kubeconfig }}
+  changed_when: False
+
```
tbielawa: OK, we're making a copy of the master config to follow an existing pattern where we protect the /etc/ config from changes if we're switching around contexts. Got it.

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -1,6 +1,22 @@
...
 - name: Wait for Node Registration
   command: >
-      {{ openshift.common.client_binary }} get node {{ item | lower }}
+    {{ openshift.common.client_binary }} get node {{ item | lower }}
+    --config={{ openshift_manage_node_kubeconfig }}
+    -n default
```
tbielawa: And here on down we're just pointing to the new temporary config file (following the established pattern, such as in the manageiq role) and specifying the default namespace for all following commands. 

ACK

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -10,14 +26,24 @@
...
 - name: Label nodes
   command: >
-    {{ openshift.common.client_binary }} label --overwrite node {{ item.openshift.common.hostname | lower }} {{ item.openshift.node.labels | oo_combine_dict  }}
+    {{ openshift.common.client_binary }} label --overwrite node {{ hostvars[item].openshift.common.hostname | lower }} {{ hostvars[item].openshift.node.labels | oo_combine_dict  }}
+    --config={{ openshift_manage_node_kubeconfig }}
+    -n default
   with_items:
-    -  "{{ openshift_node_vars }}"
-  when: item.openshift.common.hostname is defined and 'labels' in item.openshift.node and item.openshift.node.labels != {}
+    -  "{{ openshift_nodes }}"
```
tbielawa: Not looking at all of the node vars anymore (per changes in the other file) so switching to just `openshift_nodes`. Still labeling nodes, just using a more concise variable to iterate over. 

ACK

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -10,14 +26,24 @@
...
+  when: hostvars[item].openshift.common.hostname is defined and 'labels' in hostvars[item].openshift.node and hostvars[item].openshift.node.labels != {}
+
+- name: Delete temp directory
+  file:
+    name: "{{ mktemp.stdout }}"
+    state: absent
+  changed_when: False
```
tbielawa: Yeah. Get ridda that. Burn it with fire. We don't like you anyway, temporary config file.

tbielawa:
No objections in the line-comments, just me working through understanding the changes.

One notable caveat, which isn't something you can work around, this only works if a user hasn't fudged around their /etc/ admin.kubeconfig already. You can't work around that. 

The benefit of this is that we're no longer depending on the user having a "correct" context set in their user's  `~/.kube/config`. 

LGTM

playbooks/common/openshift-node/config.yml
detiber: :+1: on cleaning it up...  One thing I wonder is if we need to have some type of a filter to provide a more concise mapping of node -> label/scheduleable values. Right now we assume all of the node objects/vars are present, but maybe we should have a more strict definition of what we are acting on and enforce that at the playbook level.

This would also prevent transient node errors from potentially causing issues with the role as well.

PR #2418, opened 2016-09-07
Explain our branching strategy

sdodson:
Explain out branching strategy

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,15 @@ that it installs currently require a package that provides `docker`. Currently
 the RPMs provided from dockerproject.org do not provide this requirement, though
 they may in the future.
 
+##Branches
+The master branch tracks our current work and should be compatible with both
+Origin master branch and the current Origin stable release, currently that's
+v1.3 and v1.2. Enterprise branches exist where we coordinate with internal Red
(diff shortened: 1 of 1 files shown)


sdodson:
@elyscape does this help?

README.md
```diff
@@ -7,6 +7,15 @@ that it installs currently require a package that provides `docker`. Currently
...
 they may in the future.
 
+##Branches
+The master branch tracks our current work and should be compatible with both
+Origin master branch and the current Origin stable release, currently that's
+v1.3 and v1.2. Enterprise branches exist where we coordinate with internal Red
+Hat Quality Assurance teams. Fixes and backwards compatible feature improvements
+are often backported to the more current enterprise branches.  The enterprise
+branches should also be compatible with the corresponding Origin release for
+users who are looking for more conservative rate of change.
```
detiber: We should probably delete off some of the old branches that are not version related as well.

elyscape:
Looks good to me, though it might also be worthwhile to explicitly call out the releases as well.

maxamillion:
This is great info, thanks!

README.md
sdodson: cleaned up stale branches.

PR #2420, opened 2016-09-08
a-o-i: Do not display version number in quick installer

smunilla:
As we now require the version of the installer to match what will be installed in
the target environment and to avoid any potential confusion, remove the version
number from the variant choice question.


smunilla:
aos-ci-test

tbielawa:
Built and tested this branch.

LGTM

PR #2422, opened 2016-09-08
Fix node label / schedulability when inventory_hostname != openshift.common.hostname

abutcher:
Fixes problem introduced by 75d4f13ad39a770a8954aef9507e5d136ecb3306 in which node labels / schedulability would not be applied when node `inventory_hostname` differed from `openshift.common.hostname`.

Previously, tasks in the `openshift_manage_node` role looped over `openshift.common.hostname` for all nodes but will now loop over `inventory_hostname` for all nodes s.t. `item == inventory_hostname` and can be used to access `hostvars` within each task.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1374474

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -156,9 +156,7 @@
 - name: Set schedulability
   hosts: oo_first_master
   vars:
-    openshift_nodes: "{{ hostvars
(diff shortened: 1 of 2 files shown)


abutcher:
aos-ci-test

sdodson:
LGTM

abutcher:
aos-ci-test

PR #2424, opened 2016-09-09
the kubelet example line fails on releases prior to 3.3, so put a comment there.

jeremyeder:
Prefer anything more extensive than this?


sdodson:
That's fine.

sdodson:
Actually, I think we're making a move to keep things in this repo more oriented towards origin when we speak of things like versions. For now can you say 1.3/3.3? Also, when you say fail is it silently ignored or does it prevent the node from starting?

jeremyeder:
I can add the 1.3/3.3.  The node process fails to start because pods-per-core isn't valid.  The error message is

```
openshift-node service failing to start because of "Invalid flag pods-per-core".
```

sdodson:
@jeremyeder bump

jeremyeder:
Better now?

PR #2425, opened 2016-09-09
Flush handlers before marking a node schedulable after upgrade.

dgoodwin:
Handlers normally only trigger at the end of the play, but in this case
we just set our node schedulable again resulting in it immediately
getting taken down again.


sdodson:
LGTM

sdodson:
aos-ci-test-extended

PR #2426, opened 2016-09-09
a-o-i: Allow better setting of host level variables

smunilla:
Allow the user to set arbitrary variables directly under the host item instead of
in under an 'other_variables' item.

### DIFF
--- a/utils/src/ooinstall/oo_config.py
+++ b/utils/src/ooinstall/oo_config.py
@@ -26,6 +26,19 @@ DEPLOYMENT_VARIABLES_BLACKLIST = [
     'roles',
 ]
 
+HOST_VARIABLES_BLACKLIST = [
+    'ip',
+    'public_ip',
+    'hostname',
+    'public_hostname',
+    'node_labels',
+    'containerized',
+    'preconfigured',
+    'schedulable',
+    'other_variables',
+    'roles',
+]
+
 DEFAULT_REQUIRED_FACTS = ['ip', 'public_ip', 'hostname', 'public_hostname']
 PRECONFIGURED_REQUIRED_FACTS = ['hostname', 'public_hostname']
 
@@ -66,7 +79,7 @@ class Host(object):
         self.containerized = kwargs.get('containerized', False)
         self.node_labels = kwargs.get('node_labels', '')
 
-        # allowable roles: master, node, etcd, storage, master_lb, new
+        # allowable roles: master, node, etcd, storage, master_lb
         self.roles = kwargs.get('roles', [])
 
         self.other_variables = kwargs.get('other_variables', {})
@@ -86,11 +99,13 @@ class Host(object):
         d = {}
 
         for prop in ['ip', 'hostname', 'public_ip', 'public_hostname', 'connect_to',
-                     'preconfigured', 'containerized', 'schedulable', 'roles', 'node_labels',
-                     'other_variables']:
+                     'preconfigured', 'containerized', 'schedulable', 'roles', 'node_labels', ]:
             # If the property is defined (not None or False), export it:
             if getattr(self, prop):
                 d[prop] = getattr(self, prop)
+        for variable, value in self.other_variables.iteritems():
+            d[variable] = value
+
         return d
 
     def is_master(self):
@@ -202,7 +217,6 @@ class OOConfig(object):
                     role_list = loaded_config['deployment']['roles']
                 except KeyError as e:
                     print_read_config_error("No such key: {}".format(e), self.config_path)
-                    print "Error loading config, required key missing: {}".format(e)
                     sys.exit(0)
 
                 for setting in CONFIG_PERSIST_SETTINGS:
@@ -237,6 +251,10 @@ class OOConfig(object):
 
                 # Parse the hosts into DTO objects:
                 for host in host_list:
+                    host['other_variables'] = {}
+                    for variable, value in host.iteritems():
+                        if variable not in HOST_VARIABLES_BLACKLIST:
+                            host['other_variables'][variable] = value
                     self.deployment.hosts.append(Host(**host))
 
                 # Parse the roles into Objects
--- a/utils/src/ooinstall/openshift_ansible.py
+++ b/utils/src/ooinstall/openshift_ansible.py
@@ -31,6 +31,15 @@ VARIABLES_MAP = {
     'proxy_exclude_hosts': 'openshift_no_proxy',
 }
 
+HOST_VARIABLES_MAP = {
+    'ip': 'openshift_ip',
+    'public_ip': 'openshift_public_ip',
+    'hostname': 'openshift_hostname',
+    'public_hostname': 'openshift_public_hostname',
+    'node_labels': 'openshift_node_labels',
+    'containerized': 'containerized',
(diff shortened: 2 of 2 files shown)


smunilla:
@tbielawa PTAL :)

utils/src/ooinstall/oo_config.py
```diff
@@ -238,6 +252,10 @@ def _read_config(self):
 
                 # Parse the hosts into DTO objects:
                 for host in host_list:
+                    host['other_variables'] = {}
+                    for variable, value in host.iteritems():
+                        if variable not in HOST_VARIABLES_BLACKLIST:
+                            host['other_variables'][variable] = value
```
detiber: Any reason to keep them in other_variables here?

utils/src/ooinstall/openshift_ansible.py
```diff
@@ -175,30 +184,20 @@ def write_proxy_settings(base_inventory):
...
-        facts += ' openshift_ip={}'.format(host.ip)
-    if host.public_ip:
-        facts += ' openshift_public_ip={}'.format(host.public_ip)
-    if host.hostname:
-        facts += ' openshift_hostname={}'.format(host.hostname)
-    if host.public_hostname:
-        facts += ' openshift_public_hostname={}'.format(host.public_hostname)
-    if host.containerized:
-        facts += ' containerized={}'.format(host.containerized)
+    for prop in HOST_VARIABLES_MAP:
```
detiber: seems like it might be easier to push this into the Host object itself, then you could just pass in the dictionary from the config, mutate it as needed and then handle the mapping internally there. You'd have to also handle data migration within the host object itself, but seems like it would keep the logic more self-contained.

utils/src/ooinstall/oo_config.py
smunilla: Yes. The reason is that the Host object has a number of attributes (connect_to, public_ip, containerized, etc) and any other variables specified in the YAML file need to go somewhere. In this case, that somewhere is a dictionary called `other_variables`. I suppose we could accomplish something similar by adding attributes dynamically, but I like having them all in once place for easy in writing the ansible inventory later

utils/src/ooinstall/openshift_ansible.py
smunilla: I have the mapping between quickinstaller variables and ansible variables all in one place intentionally. Any changes to what ansible is expecting can then all happen in one place.

tbielawa: I took a swing at refactoring that stuff into the `Host` class during the code freeze, I got it working 1:1 w/ current behavior (verified via new unit test)

https://github.com/tbielawa/openshift-ansible/commit/3d76b487b098bb30bdf136a7c6ff0ddbab3cb3bc#diff-3e7671fdf1d178a3b0619520545c963eR164

Diff against master: https://github.com/openshift/openshift-ansible/compare/master...tbielawa:RefactorHostClass

tbielawa: But you probably don't want to work on integrating that right now. It'd be scope creep taken to the max.

smunilla: Yeah, I think this gives us what we need right now. It'll all probably end up getting re-written whenever we switch away from writing the inventory file out this way anyways.

detiber: I'm good with deferring improvements here until later...  It sounds like we may need some type of shared base class (or even a python mixin) to handle attribute mapping, but I definitely want to see us more more towards self-managing classes for a lot of this stuff to avoid having spaghetti code. Ideally, we should be able to take a top level deployment object and serialize it either into a config object or to a inventory (or inventory and group_vars/host_vars) easily.

utils/src/ooinstall/openshift_ansible.py
```diff
@@ -175,30 +184,20 @@ def write_proxy_settings(base_inventory):
...
+        if getattr(host, prop):
+            facts += ' {}={}'.format(HOST_VARIABLES_MAP.get(prop), getattr(host, prop))
+
     if host.other_variables:
         for variable, value in host.other_variables.iteritems():
             facts += " {}={}".format(variable, value)
-    if host.node_labels:
-        if role == 'node':
-            facts += ' openshift_node_labels="{}"'.format(host.node_labels)
```
tbielawa: I think that removing this specific conditional fact append (`host.node_labels`) is causing errors for me.

Running RPM from latest master I get a pretty `hosts` file like this

```ini
[nodes]
192.168.0.14  openshift_ip=192.168.0.14 openshift_public_ip=192.168.0.14 \
    openshift_hostname=deepfryer.redhat.com openshift_public_hostname=deepfryer.redhat.com \
    openshift_node_labels="{'region': 'infra'}" openshift_schedulable=True
```

See how `openshift_node_labels` sets its value wrapped in double quotes?

Then, running from [smunilla/BZ1368414](https://github.com/smunilla/openshift-ansible/commit/5fa778d12eb223bab9c9ad33c8a14cd9f1846f37)

```ini
[nodes]
192.168.0.14  openshift_node_labels={'region': 'infra'} \
    openshift_schedulable=True
```

I suspect this is what's causing me to get this (misleading) error:

```
*** Installation Summary***

...
Gathering information from hosts...
ERROR! Syntax Error while loading YAML.


The error appears to have been in '/home/tbielawa/.config/openshift/hosts': line 2, column 7, but may

[OSEv3:children]
      ^ here

/home/tbielawa/.config/openshift/hosts:13: Expected key=value host variable assignment, got: infra}
There was a problem fetching the required information. Please see /tmp/ansible.log for details.
```

tbielawa: I made a simple patch for this. Built the rpm and installed it. Now I get correct non-exploding output.

~~We may want to consider adding a unit test case for this in the future. The work I did refactoring before included a new unittest to verify current vs. my new method output for writing the hosts file. It actually does 99% of the work to implement a unittest to verify things like this. ~~

I added a unittest to my fork of smunilla/BZ1368414 

* https://github.com/tbielawa/openshift-ansible/commit/2733b8f9cfaca3a4b1c73a5e9df6f757b8581d99 is the actual code fix
* https://github.com/tbielawa/openshift-ansible/commit/a41d7232d665bd126b6ea525e0869c5fd28ef33e adds a unittest that verifies the node labels are written out in double quotes.

detiber: @tbielawa I would suggest a PR against the forked feature branch for suggesting changes rather than linking to remote commits.

tbielawa: Smunilla. PR incoming to your fork for merge into BZ1368414

tbielawa:
@smunilla PR created in your repo which includes my noted fixes and tests.

tbielawa:
ok to test

smunilla:
aos-ci-test

smunilla:
aos-ci-test

utils/src/ooinstall/openshift_ansible.py
sdodson: Was this line of conversation ever resolved?

PR #2427, opened 2016-09-09
Bug 1298336 - Rerunning the installer fails when not in default namespace

abutcher:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1298336


detiber:
lgtm

PR #2428, opened 2016-09-09
Bug 1367284 - rootdirectory configuration is hardcode when installer is using s3 as registry storage

abutcher:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1367284
Bug 1367284


detiber:
lgtm

PR #2429, opened 2016-09-09
Pause after restarting openvswitch in containerized upgrade.

dgoodwin:
The openvswitch restart appears to hang the node container running
briefly, if you try to restart node too quickly it will fail complaining
that the node container name is still in use. Pausing after openvswitch
restart allows the existing node container to catch up and then shutdown
cleanly.

### DIFF
--- a/roles/openshift_node/handlers/main.yml
+++ b/roles/openshift_node/handlers/main.yml
@@ -2,6 +2,12 @@
 - name: restart openvswitch
   service: name=openvswitch state=restarted
   when: not (ovs_service_status_changed | default(false) | bool)
+  notify:
+  - restart openvswitch pause
+
+- name: restart openvswitch pause
+  pause: seconds=15
+  when: openshift.common.is_containerized | bool
 
 - name: restart node
   service: name={{ openshift.common.service_type }}-node state=restarted


abutcher:
aos-ci-test

roles/openshift_node/tasks/systemd_units.yml
```diff
@@ -23,6 +23,7 @@
   register: install_ovs_sysconfig
   notify:
   - restart openvswitch
+  - restart openvswitch pause
```
sdodson: A handler can notify another handler. Maybe just have `restart openvswitch` `notify: restart openvswitch pause` so anyone down the line doesn't have to remember to notify both?

sdodson: And if you only want it to happen for containerized environments just add a when to `restart openvswitch pause`

dgoodwin:
Much better, thanks @sdodson, updated now.

sdodson:
LGTM

sdodson:
aos-ci-test

dgoodwin:
Dead link for the errored test. :(

sdodson:
It was a timeout on the setup module. Killed it.

PR #2431, opened 2016-09-09
Always deduplicate detected certificate names

### LINKED ISSUE #2430: Named certificates with SANs cause duplicate name entries when cacert is specified
When deploying custom named certificates, if you don't specify the names, [they will be detected from the certificate's CN and SANs][detect]. Now, the first entry in a certificate's SAN list is always the CN. As such, the resulting list needs to be deduplicated. And, as it turns out, [this is done][dedupe].

Unfortunately, for some reason, this is only done if the `cacert` field on the Ansible certificate object [is unspecified][bug]. I'm not quite sure the purpose of the preceding line but I think the aim is to not use self-signed certificates for internal communications.

In any event, the deduplication of the name list should be performed regardless of whether or not a value for `cacert` is provided. As it stands, if you specify `cacert` but not `names` for a certificate with SANs, you'll get an entry in `master-config.yaml`'s `named_certificates` field that looks like this:
```yaml
- certFile: /etc/origin/master/named_certificates/console.crt
  keyFile: /etc/origin/master/named_certificates/console.key
  names:
  - "subdomain.example.com"
  - "subdomain.example.com"
  - "example.com"
```
Which will then cause the `origin-master` service to fail with the following error:
> servingInfo.namedCertificates[0].names[1]: Invalid value: "subdomain.example.com": this name is already used in another named certificate

[detect]: https://github.com/openshift/openshift-ansible/blob/openshift-ansible-3.4.1-1/filter_plugins/oo_filters.py#L509-L537
[dedupe]: https://github.com/openshift/openshift-ansible/blob/openshift-ansible-3.4.1-1/filter_plugins/oo_filters.py#L533
[bug]: https://github.com/openshift/openshift-ansible/blob/openshift-ansible-3.4.1-1/filter_plugins/oo_filters.py#L531-L533

elyscape:
Fixes #2430

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -529,9 +529,9 @@ class FilterModule(object):
                 raise errors.AnsibleFilterError(("|failed to parse certificate '%s', " % certificate['certfile'] +
                                                  "please specify certificate names in host inventory"))
 
+            certificate['names'] = list(set(certificate['names']))
             if 'cafile' not in certificate:
                 certificate['names'] = [name for name in certificate['names'] if name not in internal_hostnames]
-                certificate['names'] = list(set(certificate['names']))
                 if not certificate['names']:
                     raise errors.AnsibleFilterError(("|failed to parse certificate '%s' or " % certificate['certfile'] +
                                                      "detected a collision with internal hostname, please specify " +


elyscape:
Also, should I backport this patch to any other branches? This issue affects them all.

johnr-spindance:
👍

abutcher:
ok to test

abutcher:
aos-ci-test

abutcher:
LGTM, defer to @sdodson regarding backports

detiber:
lgtm

abutcher:
aos-ci-test

PR #2437, opened 2016-09-12
Move delegated_serial_command to etcd_common.

abutcher:


roles/etcd_client_certificates/meta/main.yml
```diff
@@ -14,3 +14,4 @@ galaxy_info:
   - system
 dependencies:
 - role: etcd_common
+- role: delegated_serial_command
```
detiber: Should this also have a dependency on etcd_ca?

detiber:
Do we need to prefix the role in some manner to indicate that it is a module-only role? Also, do we want to possibly have one or more consolidated roles for these types of custom modules rather than doing a dedicated role for each one?

roles/etcd_client_certificates/meta/main.yml
abutcher: We still have dependencies in the playbooks for most things master/node/etcd. It's probably safe to return those to their `meta/main.yml` files unless we want to provide a little speed boost for ansible-2.1.

abutcher:
I vote separate role per module and not prefixed as module only but I don't feel strongly about it.

roles/etcd_client_certificates/meta/main.yml
detiber: It'd be nice to minimize the dependency on master and node roles, we should pass in any master/node specific values to this role instead, I think. etcd_ca seems like a hard dependency to me, though.

abutcher: Yeah, remember how we cut down on certain dependencies and added checks like ensuring the etcd ca exists in `etcd_client_certificates` to cut down on the inflated run time w/ 2.1? We can undo that but I expect there are still quite a few 2.1 users.

7ab97920 - Role dep migration for master/node/hosted.
3bd5ae2 - `etcd_ca` removed from etcd_client_certificates dep here.

abutcher: I guess with 2.2 in the repos we should undo the changes linked above.

detiber: We probably need to move the majority of the etcd_ca tasks into an included task file as well, which would help with runtimes and still let us have the hard dependency.

abutcher: I'll work those in a separate PR. What's the verdict on:

* Multiple modules housed in a single role.
* Prefixing the name of this role as module only in some way if it exists only for `delegated_serial_command`.

tbielawa: Putting all of our custom modules into a single role sounds smart to me

```
.
└── openshift-ansible
    └── roles
        └── openshift_modules
            ├── library
            │   ├── delegated_serial_command.py
            │   ├── modify_yaml.py
            │   ├── ...
            │   └── rpm_q.py
            ├── meta
            │   └── main.yml
            └── README.md
```

Each other role can then just set a dependency on the `openshift_modules` role then, right?

detiber: > edit note that this could be interpreted as going against the recommended directory layout per http://docs.ansible.com/ansible/playbooks_best_practices.html#directory-layout

That guide doesn't take into account that modules and other plugins can be embedded inside of roles. Ops really likes using roles to allow better re-use and in discussions with some Ansible folks about this, they didn't see any issues with this method. Requiring modules that exist outside of a role would also preclude being able to publish usable roles to Galaxy.

All that said, now that we are on 2.x, we should move our filter plugins into (a) role(s)

detiber:
So, I just realized that delegated_serial_command is *only* used within the etcd_*_certificates roles, with that in mind, if we bring back the etcd_ca role dependency, then maybe the module should just be part of that role?

abutcher:
> So, I just realized that delegated_serial_command is only used within the etcd_*_certificates roles, with that in mind, if we bring back the etcd_ca role dependency, then maybe the module should just be part of that role?

I'd be okay moving the module to either `etcd_common` or `etcd_ca`.

abutcher:
@detiber how do you feel about moving to `etcd_common` for now?

abutcher:
aos-ci-test

PR #2441, opened 2016-09-12
3.4 Upgrade Improvements

dgoodwin:
Large refactor to support running upgrade on only your control plane, or only your nodes. Done by splitting out all portions of upgrade as necessary, and then including relevant snippits in the primary playbooks used as entry points. 

Added pre_upgrade tag for just running pre-upgrade checks.

Added support for serializing node upgrade as desired.

Added support for only upgrading nodes with a given label.

See trello card for all other improvements made: https://trello.com/c/97zpMmYc/282-5-upgrade-upgrade-improvements


dgoodwin:
@detiber @abutcher @sdodson if any of you have time for a quick sanity check I'd love to know if this is on an acceptable track. It's a lot more small files to get through upgrade, just combined in different ways. Very little duplicated yaml though.

playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
```diff
@@ -1,67 +1,83 @@
...
+
+- include: ../../../../common/openshift-cluster/upgrades/pre/verify_control_plane_running.yml
+
+- include: ../../../../common/openshift-cluster/upgrades/pre/verify_upgrade_targets.yml
+
+- include: ../../../../common/openshift-cluster/upgrades/pre/verify_docker_upgrade_targets.yml
+
+- include: ../../../../common/openshift-cluster/upgrades/pre/gate_checks.yml
+
+- include: ../../../../common/openshift-cluster/upgrades/pre/backup_etcd.yml
```
abutcher: Do we want to backup etcd in pre-upgrade checks? Running for check and then upgrade will result in two backups but that's not necessarily a bad thing I guess.

playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
```diff
@@ -1,67 +1,83 @@
...
+
+- name: Exit upgrade if dry-run specified
+  hosts: oo_al_hosts
+  tasks:
+  - fail:
+      msg: "Pre-upgrade checks completed, exiting due to openshift_upgrade_dry_run variable."
+    when: openshift_upgrade_dry_run is defined and openshift_upgrade_dry_run | bool
```
abutcher: What do you think about structuring this playbook as pre-upgrade/upgrade conditionally running so that there is no need to fail and get a non zero exit code.

playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
dgoodwin: I do now have a step at the start of full upgrade which is kind of similar to this where we clean out unused images/containers if we're going to be upgrading docker, this probably would make sense to move there as well, good idea.

playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
dgoodwin: Then again I just saw I typo'd oo_al_hosts in every playbook so this wasn't getting run now anyhow. :)

dgoodwin: Was just about to do this and reading up, what about using tags here? So we document if you want to just run pre-upgrade you run with --tags "pre-upgrade". This eliminates my new var, and the need to always conditionalize all non-pre-upgrade includes.

dgoodwin:
This work is tracked in https://trello.com/c/97zpMmYc/282-5-upgrade-upgrade-improvements

dgoodwin:
Squashed commits and removed WIP, I think this is ready for full review now.

playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
abutcher: +1

dgoodwin:
687474703a2f2f73332e616d617a6f6e6177732e636f6d2f72617067656e6975732f313239313133313638305f74776f2d7468756d62732d75702e6a7067

abutcher:
aos-ci-test

dgoodwin:
Failure was a no packages to download flake. 

aos-ci-test

abutcher:
aos-ci-test

abutcher:
aos-ci-test

dgoodwin:
All clear.

PR #2442, opened 2016-09-13
get quickstarts from origin, not upstream example repos

bparees:


bparees:
@sdodson @brenton with this change, we will no longer be shipping the non-DB quickstarts, eg:
https://github.com/openshift/nodejs-ex/tree/master/openshift/templates

(we'll only get nodejs-mongo.json, not nodejs.json, since nodejs.json isn't in the origin quickstarts directory).

So we may need to discuss if that's acceptable.

@luciddreamz fyi.  if/when this goes in we can merge:
https://github.com/openshift/django-ex/pull/59
https://github.com/openshift/nodejs-ex/pull/64
https://github.com/openshift/rails-ex/pull/61
https://github.com/openshift/cakephp-ex/pull/45
https://github.com/openshift/nodejs-ex/pull/64

luciddreamz:
@bparees Also remove those ~5 quickstarts in this PR so we don't forget they're not getting updated via the sync script anymore.

sdodson:
followup item, for a separate PR, handle removal of the defunct quickstarts from both the filesystem and openshift namespace. Example https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_examples/tasks/main.yml#L100-L113 if you wish to go ahead and do that now.

bparees:
@sdodson updated to remove the old templates we don't want to provide.

sdodson:
LGTM

sdodson:
aos-ci-test

roles/openshift_examples/tasks/main.yml
```diff
@@ -73,6 +73,27 @@
...
+
+- name: Remove defunct quickstart templates from openshift namespace
+  command: "{{ openshift.common.client_binary }} -n openshift delete templates/{{ item }}"
+  with_items:
+    - nodejs-example
+    - cakephp-example
+    - dancer-example
+    - django-example
+  register: oex_delete_defunct_quickstart_templates
+  failed_when: "'not found' not in oex_delete_defuct_quickstart_templates.stderr and oex_delete_defunct_quickstart_templates.rc != 0"
```
sdodson: defuct versus defunct in the first instance here

sdodson:
Also need to remove the files from roles/openshift_examples/files/examples/v1.3/quickstart-templates

I can pick this up tomorrow if you don't have time to deal with it.

bparees:
> Also need to remove the files from roles/openshift_examples/files/examples/v1.3/quickstart-templates

done...I wasn't sure if this was ultimately a 1.4 item in which case the files still belonged in 1.3.

roles/openshift_examples/tasks/main.yml
bparees: fixed, thanks.

sdodson:
aos-ci-test

sdodson:
ansible flake

PR #2447, opened 2016-09-13
Switch to origin-1.x branch names

sdodson:
Prepare to switch from `enterprise-3.x` to `origin-1.x` branch names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,14 +7,18 @@ that it installs currently require a package that provides `docker`. Currently
 the RPMs provided from dockerproject.org do not provide this requirement, though
 they may in the future.
 
-##Branches
+##Branches and tags
+
 The master branch tracks our current work and should be compatible with both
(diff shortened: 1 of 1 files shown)


sdodson:
@maxamillion @elyscape thoughts on this? We decided that a public github project should more closely align with the project than downstream product. Though maybe `stable-1.x` or `release-1.x` (used by origin repo) is better now that I think about it.

elyscape:
Either option sounds good to me!

detiber:
@sdodson release-1.x is preferable to me than stable-1.x.

maxamillion:
This is great! Thanks all involved. :+1:

sdodson:
release-1.x it is, i'll prep our CI jobs for this today and duplicate branches after that's done and then purge the enterprise-3.x branches soon thereafter

PR #2450, opened 2016-09-14
[redeploy-certificates] Correct etcd service name for containerized.

### LINKED ISSUE #2448: redeploy_certificates playbook fails on containerized install
Playbook fails attempting to restart the etcd service. This is because the service is named differently on a containerized environment: `etcd_container.service` instead of `etcd.service`

https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/redeploy-certificates.yml

abutcher:
Template etcd service name in cert redeploy playbook.

I'm also setting `openshift_master_default_subdomain` similarly to [playbooks/common/openshift-master/config.yml#L39-L41](https://github.com/openshift/openshift-ansible/blob/872aafbe7a65388c473d125095ae60fc9ae915a4/playbooks/common/openshift-master/config.yml#L39-L41) to work around an issue I encountered with master facts in the redeploy playbook.

```
TASK [openshift_master_facts : Set master facts] *******************************
Wednesday 14 September 2016  12:05:16 -0400 (0:00:00.068)       0:01:44.888 ***
fatal: [master4.example.com]: FAILED! => {
    "failed": true
}

MSG:

the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: {{ openshift_hosted_metrics_public_url | default('hawkular-metrics.' ~ (openshift.master.default_subdomain | default(openshift_master_default_subdomain ))) | oo_hostname_from_url }}: 'openshift_master_default_subdomain' is undefined

The error appears to have been in '/home/abutcher/rhat/openshift-ansible/roles/openshift_master_facts/tasks/main.yml': line 2, column 3, but may

---
- name: Set master facts
  ^ here
```

Setting the `metrics_public_url` in master facts only seems to work because `openshift_master_default_subdomain` has a value of `""` and can be used as a default. Copying it here mirrors how the byo playbooks function. Will follow up / create issue to fix.

Fixes #2448.
@sdodson PTAL

### DIFF
--- a/playbooks/common/openshift-cluster/redeploy-certificates.yml
+++ b/playbooks/common/openshift-cluster/redeploy-certificates.yml
@@ -52,6 +52,14 @@
     openshift_ca_host: "{{ groups.oo_first_master.0 }}"
     openshift_master_count: "{{ openshift.master.master_count | default(groups.oo_masters | length) }}"
   pre_tasks:
+  # set_fact task copied from playbooks/common/openshift-master/config.yml
+  # so that openshift_master_default_subdomain has a default value of ""
+  # (emptry string). openshift_master_default_subdomain must have a default
+  # value for openshift_master_facts to set metrics_public_url.
+  # TODO: clean this up.
+  - set_fact:
+      openshift_master_default_subdomain: "{{ lookup('oo_option', 'openshift_master_default_subdomain') | default(None, true) }}"
+    when: openshift_master_default_subdomain is not defined
   - stat:
       path: "{{ openshift_generated_configs_dir }}"
     register: openshift_generated_configs_dir_stat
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/redeploy-certificates.yml
```diff
@@ -52,6 +52,14 @@
...
     openshift_master_count: "{{ openshift.master.master_count | default(groups.oo_masters | length) }}"
   pre_tasks:
+  # set_fact task copied from playbooks/common/openshift-master/config.yml
+  # so that openshift_master_default_subdomain has a default value of ""
+  # (emptry string). openshift_master_default_subdomain must have a default
+  # value for openshift_master_facts to set metrics_public_url.
+  # TODO: clean this up.
+  - set_fact:
+      openshift_master_default_subdomain: "{{ lookup('oo_option', 'openshift_master_default_subdomain') | default(None, true) }}"
+    when: openshift_master_default_subdomain is not defined
```
sdodson: whoah, neat

sdodson:
LGTM

PR #2451, opened 2016-09-14
a-o-i: Don't set unschedulable nodes as infra

smunilla:
Make sure we don't set an unschedulable node as infra as that can cause problems.

Fixes: Bug 1375946


smunilla:
aos-ci-test

sdodson:
aos-ci-test

PR #2453, opened 2016-09-15
Add role to manageiq to allow creation of projects

zeari:
This should allow manageiq to:
  1. request a new project, become an admin in that project,
  2. instantiate a template in that project,
  3. bind a role in that project

@deads2k Please review.

@enoodle Thanks for the help
cc @simon3z @moolitayer @zakiva

### DIFF
--- a/roles/openshift_manageiq/vars/main.yml
+++ b/roles/openshift_manageiq/vars/main.yml
@@ -30,6 +30,7 @@ manage_iq_tasks:
     - policy add-scc-to-user privileged system:serviceaccount:management-infra:management-admin
     - policy add-cluster-role-to-user system:image-puller system:serviceaccount:management-infra:inspector-admin
     - policy add-scc-to-user privileged system:serviceaccount:management-infra:inspector-admin
+    - policy add-cluster-role-to-user self-provisioner system:serviceaccount:management-infra:management-admin
 
 manage_iq_openshift_3_2_tasks:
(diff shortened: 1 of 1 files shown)


simon3z:
@zeari @moolitayer @detiber should we handle upgrades?

E.g. `system:image-auditor` was added under `manage_iq_openshift_3_2_tasks`. Should we add this under `manage_iq_openshift_3_3_tasks`?

simon3z:
cc @deads2k

zeari:
As far as I can tell this role allows you to create projects through this process: https://docs.openshift.com/enterprise/3.0/admin_guide/selfprovisioned_projects.html

which isnt quite the same as creating a project as a cluster-admin

deads2k:
That's the correct role and the command looks correct.

detiber:
ok to test

detiber:
aos-ci-test

simon3z:
@abutcher @detiber what about this [comment](https://github.com/openshift/openshift-ansible/pull/2453#issuecomment-247275555). No need for anything related to upgrade?

abutcher:
@simon3z the `openshift_manageiq` role is applied during upgrade in [post.yml](https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/upgrades/post.yml#L13) and each task in this list will be ran. Based on CI running clean in 3.2/3.3, the added task should be okay for either version. I could be missing something...

PR #2475, opened 2016-09-20
Install Registry by Default

smunilla:
Instead of restricting cockpit-ui to Atomic Registry installations, install it by default everywhere. `deployment_subtype=registry` is no longer required to deploy the Atomic Registry but can be used to create a standalone registry deployment (no web console, etc). This requires securing the registry by default.

Any users maintaining their own registry and/or users which don't want to deploy Atomic Registry may disable management of the registry by adding the following variable to their inventory.

```
openshift_hosted_manage_registry=false
```


smunilla:
@abutcher PTAL

smunilla:
aos-ci-test

smunilla:
CI isn't landing the registry right. Time to figure out what's up.

abutcher:
@smunilla Looks like the image may be missing from our configured registries. /cc @sdodson

abutcher:
Nevermind, the template probably wasn't created.

smunilla:
aos-ci-test

smunilla:
aos-ci-test

smunilla:
aos-ci-test

smunilla:
aos-ci-test

smunilla:
aos-ci-test

smunilla:
aos-ci-test

smunilla:
@sdodson Mind looking this over?

smunilla:
aos-ci-test

smunilla:
aos-ci-test

playbooks/common/openshift-cluster/node_docker_ca.yml
```diff
@@ -0,0 +1,128 @@
+---
+- name: Configure CA certificate for secure registry
```
detiber: I'd like to see this logic pushed down into a role, rather than a playbook. Ideally, we'd be able to apply this as part of the node/docker config.

@abutcher maybe we need to deploy the registry prior to configuring the nodes, that would provide the needed info for node configuration.

abutcher: @detiber We only need to know the service IP and route here so that should work.

smunilla:
aos-ci-test

playbooks/common/openshift-cluster/node_docker_ca.yml
abutcher: This playbook is gone now that we're adding the OpenShift CA to the system trusted roots for each node.

abutcher:
@sdodson PTAL when ya can

abutcher:
aos-ci-test

PR #2485, opened 2016-09-21
Add documentation about the openshift_hosted_metrics_public_url option

yaacov:
We now have an option in the `openshift_metrics` role to set the metrics public hostname

This PR adds the missing documentaion.


yaacov:
@simon3z @sdodson @alongoldboim please review

PR #2486, opened 2016-09-21
Clarify openshift_hosted_metrics_public_url

sdodson:
Currently, metrics only works at /hawkular/metrics so if you alter the path
portion of the variable then the master will direct traffic at an endpoint that
won't work.

### DIFF
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -363,6 +363,8 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 #
 # Override metricsPublicURL in the master config for cluster metrics
 # Defaults to https://hawkular-metrics.{{openshift_master_default_subdomain}}/hawkular/metrics
+# Currently, you may only alter the hostname portion of the url, alterting the
+# `/hawkular/metrics` path will break installation of metrics.
 #openshift_hosted_metrics_public_url=https://hawkular-metrics.example.com/hawkular/metrics
 
 
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -364,6 +364,8 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 #
 # Override metricsPublicURL in the master config for cluster metrics
 # Defaults to https://hawkular-metrics.{{openshift_master_default_subdomain}}/hawkular/metrics
+# Currently, you may only alter the hostname portion of the url, alterting the
+# `/hawkular/metrics` path will break installation of metrics.
 #openshift_hosted_metrics_public_url=https://hawkular-metrics.example.com/hawkular/metrics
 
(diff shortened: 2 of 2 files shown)


sdodson:
Followup to #2485

inventory/byo/hosts.ose.example
```diff
@@ -363,6 +363,8 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
 #
 # Override metricsPublicURL in the master config for cluster metrics
 # Defaults to https://hawkular-metrics.{{openshift_master_default_subdomain}}/hawkular/metrics
+# Currently, you may only alter the hostname portion of the url, alterting the
+# `/hawkular/metrics` path will break installation of metrics.
 #openshift_hosted_metrics_public_url=https://hawkular-metrics.example.com/hawkular/metrics
```
yaacov: I checked it with:
`openshift_hosted_metrics_public_url=hawkular-metrics.example.com` no `http://` and no `path`
I think `http://somthing.com/path` will not work

sdodson: And your master was configured properly? ie: you can see metrics tab when viewing a pod in the console?

yaacov: I do not use the console, I use CFME ....
see:
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_metrics/tasks/install.yml#L106
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_metrics/tasks/install.yml#L106

the `schema` and `path` are added

yaacov: I only checked that CFME works :-) not the console.
p.s 
the path `/hawkular/metrics` is hardcoded in CFME so we prefer it fixed.

yaacov: it's actually a hostname and not url ...
maybe rename to `openshift_hosted_metrics_public_hostname` ?

checked in the console and it works :+1: 
checked in CFME and it works :+1: 

my inventory file: https://github.com/yaacov/miq-scripts/blob/master/hosts-eng-lab.local

screenshot:

screenshot-2016-09-22-11-24-45

sdodson: I think that's just a happy accident that it works. I imagine if you were to set it to https://somehost.example.com/not_hawkular/not_metrics the console wouldn't work but I'm not sure.

yaacov: url ( e.g. `https://foo.bur/path` )will not work, it need to be onlye the `hostname`:
`openshift_hosted_metrics_public_url=somehost.example.com`

see:
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_metrics/tasks/install.yml#L106

yaacov: p.s. 
found where they forgot to set the routes path, it's in origin-metrics ( it's not an openshift-ansible thing ) :
[they set the `--hostaname=..` but leave the `--path` missing ... ]
https://github.com/openshift/origin-metrics/blob/master/deployer/scripts/hawkular.sh#L198

PR #2497, opened 2016-09-23
Fix the nodeName of the OpenShift nodes on OpenStack

lhuard1A:
On some setup of OpenStack, the hostname of the VMs as advertised by the meta-data server (http://169.254.169.254/latest/meta-data/hostname) is suffixed by `.novalocal` (or `.openstacklocal`) (e.g.: `clusterid-node-compute-XXX.novalocal`).
On some other setup of OpenStack, the hostname of the VMs as advertised by the meta-data server is just its name without any suffix (e.g.: `clusterid-node-compute-XXX`).

I found this issue reported there: https://bugs.launchpad.net/tripleo/+bug/1449520 .

This makes a difference for [`openshift_facts.py`](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_facts/library/openshift_facts.py#L152) which considers the former as a valid hostname but not the latest.

As a consequence, on a cluster spawned with `bin/cluster … openstack clusterid`, on the first OpenStack setup, the nodeName looks like `clusterid-node-compute-XXX.novalocal` whereas on the second OpenStack setup, the nodeName are the IPs.

This PR addresses this issue by enforcing a controlled domain on the VMs by adding the `hostname:` and `fqdn:` directives inside the cloud-config of the VMs.

This mimics pretty much the [setup we already had for libvirt](https://github.com/openshift/openshift-ansible/blob/master/playbooks/libvirt/openshift-cluster/templates/user-data#L5):
```yaml
hostname: {{ item[0] }}
fqdn: {{ item[0] }}.example.com
```

Both setups (openstack and libvirt) now spawns nodes the nodeName of which is the pretty name suffixed by `.example.com`.


detiber:
@lhuard1A we've had reports of this elsewhere. Do you know if there is a better metadata field we can use to determine the "name" of the instance to match up with what the API returns (to enable cloud integration), or should we just strip .novalocal or .openstacklocal when present?

lhuard1A:
Hi @detiber,

Let me share the results of my experiments on the OpenStack instance which adds the `.novalocal` suffix:
```
lenaic@ncelhuard-Xperiment:~$ openstack server list --name eric-openshift-int-a-master
+--------------------------------------+-----------------------------+--------+-----------------------------------------------+------------------------------------------------+
| ID                                   | Name                        | Status | Networks                                      | Image Name                                     |
+--------------------------------------+-----------------------------+--------+-----------------------------------------------+------------------------------------------------+
| 02548123-8eee-4da3-bfd0-7a82cd780f9c | eric-openshift-int-a-master | ACTIVE | eric-network_mgmt=192.168.0.26, 10.64.108.220 | rhel-7.2-20160114-no_selinux-NewArtifactoryURL |
+--------------------------------------+-----------------------------+--------+-----------------------------------------------+------------------------------------------------+

lenaic@ncelhuard-Xperiment:~$ ssh lhuard@10.64.108.220
Last login: Fri Sep 23 14:50:39 2016 from 172.16.23.253
[lhuard@ose3-int-a-master ~]$ curl -q http://169.254.169.254/latest/meta-data | while read field; do echo; echo "*** $field ***"; curl -q http://169.254.169.254/latest/meta-data/$field; echo; done

*** ami-id ***
ami-0000020a

*** ami-launch-index ***
0

*** ami-manifest-path ***
FIXME

*** block-device-mapping/ ***
ami
root

*** hostname ***
eric-openshift-int-a-master.novalocal

*** instance-action ***
none

*** instance-id ***
i-00038a81

*** instance-type ***
m1.medium

*** kernel-id ***
None

*** local-hostname ***
eric-openshift-int-a-master.novalocal

*** local-ipv4 ***
192.168.0.26

*** placement/ ***
availability-zone

*** public-hostname ***
eric-openshift-int-a-master.novalocal

*** public-ipv4 ***
10.64.108.220

*** public-keys/ ***
0=eric_key

*** ramdisk-id ***
None

*** reservation-id ***
r-2y80egx6
```

Unfortunately, on this OpenStack instance, there is no meta-data that contains just the naked name of the VM without the `.novalocal` suffix. So, we’ll have to strip it.


I haven’t tried to make the cloud-provider activated automatically, but, if my understanding is correct, the `nodeName` (in `/etc/origin/node/node-config.yml`) has to match the name of the VM as known by the IaaS.
So, it should be something like just `eric-openshift-int-a-master` with no suffix/domain.
But, such a name is considered as not valid by [`openshift_facts.py`](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_facts/library/openshift_facts.py#L152) which requires a host name to have at least two dot-separated parts:
```python
    if (not hostname or
            hostname.startswith('localhost') or
            hostname.endswith('localdomain') or
            len(hostname.split('.')) < 2):
        return False

    return True
```

detiber:
@lhuard1A thanks for the metadata info and you are definitely correct about the way we are treating hostnames currently. We need to shortcut that (at least for OpenStack).

@abutcher I believe you were looking at a similar issue recently.

lhuard1A:
This PR aims at just fixing the hostname of the nodes on OpenStack instances which are not appending the `.novalocal` on the hostname advertised by the metadata server.

I opened an other PR (#2501) to have the nodeName match the VM name so that the OpenStack cloudprovider can be enabled.

PR #2501, opened 2016-09-26
Fix openstack cloudprovider

lhuard1A:
There is currently an issue with the OpenStack cloudprovider because OpenShift requires that the nodeName of the nodes matches the name of the VM as reported by OpenStack.
And the hostname of the VMs may not be the name of the VM as reported by OpenStack.

In addition, on some OpenStack instances, the metadata provided by OpenStack appends an extra “.novalocal” suffix to the hostname.
See https://bugs.launchpad.net/tripleo/+bug/1449520

This PR proposes to add a new “openshift fact”: `nodename` which is computed:
* from the `common.hostname | lower` like it used to be if the cloudprovider isn’t openstack;
* from the metadata provided by OpenStack with the extra `.novalocal` removed if the cloudprovider is openstack.


lhuard1A:
@detiber, @abutcher,

Could you please have a look at this change?
This is following the discussion started in #2497 and is based on top of it.

detiber:
@abutcher lgtm from a quick review.

@jprovaznik can you test this against OpenShift on OpenStack?

detiber:
aos-ci-test

jianlinliu:
Seem like this PR is blocking QE's installation on OSP with cloudprovider enabled.

sdodson:
> Seem like this PR is blocking QE's installation on OSP with cloudprovider enabled. Tracking it in BZ#1383015.

Thanks, this is only an problem when specifying the OSP cloudprovider or did you say this was causing problems for other deployments due to problems with 'http_proxy' ?

jianlinliu:
I means this is only an problem when specifying the OSP cloudprovider.

sdodson:
@abutcher and @detiber So this PR makes it so that the masters cannot proxy to nodes because the node name doesn't resolve. Is setting `openshift_node_set_node_ip=true` a valid workaround for this condition?

detiber:
@jianlinliu is it not possible to update the tooling to use valid dns names for the instance names? Even /etc/hosts entries should be sufficient as long as dnsmasq is left enabled.

The issues at play are:
- openshift_set_node_ip is fragile and will not allow a node to be shutdown/restarted with a different ip address.
- .local/.novalocal dns names are only resolvable between hosts on the same hypervisor host.

jianlinliu:
@detiber currently we are installing env using a jenkins job, firstly launch a batch of instance with a prefix name, e.g: qe-jialiu-, if this cluster is a master + nodes env, its instance name would qe-jialiu-master, qe-jialiu-node1, qe-jialiu-node2... When instance get ready, assign a floating IP to it. 
We have valid dns name for each floating IP (e.g: 10.x.x.220 -> openshift-220.lab.sjc.redhat.com), but according to our tools work flow,  I do not find a good way to give a valid dns name when creating instance (metadata hostname return the name when creating the instance, even if change instance name manually after instance is created, still not chance).

PR #2502, opened 2016-09-26
test fix for systemd changes

sdodson:
systemd no longer returns output on `systemctl show foo.service` when foo.service doesn't exist. It returns non-zero and `Unit foo.service could not be found.`


sdodson:
aos-ci-test

sdodson:
@detiber just a quick hack

sdodson:
aos-ci-test

PR #2505, opened 2016-09-27
Update service presence checks to work with systemd in RHEL 7.3

sdodson:
Backport https://github.com/openshift/openshift-ansible/pull/2502

This adds a check on the master non ha service that didn't exist in the
release-1.1 branch but I figured it made sense to add it when backporting these
changes.

Fixes Bug 1379704


sdodson:
aos-ci-test

PR #2507, opened 2016-09-27
Remove the DNS VM on OpenStack

lhuard1A:
The reason why we had a DNS VM in the setups created by `bin/cluster create openstack …` was because of the `.novalocal` suffix that OpenStack adds on the hostname advertised by the metadata server.
This made openshift-ansible think that `clusterid-role-num.novalocal` was a valid hostname usable to construct urls for the nodes to contact the master, etc.
So, we had to have a DNS serving the `.novalocal` zone.

Now that [the `novalocal` suffixed names are not considered as valid hostnames anymore](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_facts/library/openshift_facts.py#L152), this is not needed anymore. openshift-ansible will just not find any valid pretty hostname and will use IPs as hostname.

If the cloud-provider is enabled, the nodeNames will be pretty, but the url used by the nodes to contact the master etc. will use IPs. Which is fine.

The advantages of removing that DNS VM are that creating an OpenShift cluster on OpenStack will require less resources and will be faster.


lhuard1A:
@detiber, @abutcher : What do you think about that ?

detiber:
lgtm

PR #2514, opened 2016-09-28
Make rhel_subscribe role default to OpenShift Container Platform 3.3

lhuard1A:


lhuard1A:
@detiber Could you have a look?

PR #2525, opened 2016-09-29
Add support for the write access option of hawkular mrtrics

yaacov:
CFME use openshift-metrics for metrics collection and for storing of the openshift-ops team add-hock system metrics. We use hawkular-metrics write access option to allow writing metrics into hawkular using the it's restful api.

This PR adds the option to set the write access option of hawkular-metrics, using the ansible inventory file.


simon3z:
@mwringe @yaacov why would we need this? Isn't heapster writing?

yaacov:
> why would we need this? Isn't heapster writing?

@simon3z :+1: just tested, we do not need this ...

closing this pr

PR #2529, opened 2016-09-29
Update play names for consistency.

abutcher:
@tbielawa ptal


playbooks/common/openshift-node/config.yml
```diff
@@ -85,7 +85,7 @@
       when: openshift.node.use_openshift_sdn | bool
   - role: openshift_node
 
-- name: Configure node instances
+- name: Configure nodes
```
sdodson: Maybe make these two different in some way? Configure containerized nodes, Configure nodes
Duplicate play names really trip me up when i debug things, but i may be a bit dense.

abutcher: @sdodson good call

abutcher: bumpdated

PR #2519, opened 2016-09-28
a-o-i: Move oo_config objects out

smunilla:
Move the model objects used in oo_config out to their own models
file for readablity and future expansion.

PR #2535, opened 2016-09-30
Fix openshift_node_labels in generated hosts file

smunilla:


smunilla:
This should have been included in #2426 , but wasn't for some reason.

smunilla:
aos-ci-test

tbielawa:
aos-ci-test

tbielawa:
+1 kudos to @smunilla for including and updating the new unit test

tbielawa:
Once the other PR is handled (#2519) I'll give a review on this for you @smunilla

utils/test/oo_config_tests.py
```diff
@@ -224,3 +226,82 @@ def test_load_host_no_master_or_node_specified(self):
...
+        # Given the `yaml_props` above we should see a line like this:
+        #     openshift_node_labels="{'region': 'infra'}"
+        node_labels_expected = '''openshift_node_labels="{'region': 'infra'}"'''  # Quotes around the hash
+        node_labels_bad = '''openshift_node_labels={'region': 'infra'}'''  # No quotes around the hash
+
+        # The good line is present in the written inventory line
+        self.assertIn(node_labels_expected, legacy_inventory_line)
+        bad_line_present = node_labels_bad in legacy_inventory_line
+        # An unquoted version is not present
+        self.assertFalse(bad_line_present)
```
tbielawa: Consider using `self.assertNotIn(node_labels_bad, legacy_inventory_line)` rather than using the `assertFalse` method for consistency.

Reference: https://docs.python.org/2/library/unittest.html#assert-methods

utils/test/oo_config_tests.py
```diff
@@ -224,3 +226,82 @@ def test_load_host_no_master_or_node_specified(self):
...
+
+
+    # def test_new_write_inventory_same_as_legacy(self):
```
tbielawa: This is good to keep here as a comment block. I think it'll be good for reference for now and then more useful later if we switch out how to we're generating string lines completely.

utils/test/oo_config_tests.py
```diff
@@ -265,7 +265,7 @@ def test_inventory_file_quotes_node_labels(self):
         self.assertIn(node_labels_expected, legacy_inventory_line)
         bad_line_present = node_labels_bad in legacy_inventory_line
```
tbielawa: bad_line_present must be erased too

PR #2548, opened 2016-10-05
Error out if containerized=true for lb host.

### LINKED ISSUE #1980: Containerized Flag Causes Breakage on Load Balancer System in HAProxy Deployments
If you set containerized=true in your inventory vars (for all systems, which is something I've seen QE doing), or directly on the load balancer, the installer fails with an error about missing an haproxy config file. The LB cannot be containerized as I understand it, so we should likely error out if the user has set this var for that system, or ignore it.

dgoodwin:
Fixes #1980.

It's not early in the process as we don't have a great place for cross-playbook pre-flight checks right now, so if a user hits this they will need to just correct the problem in inventory with containerized=false on that host, and then re-run.

### DIFF
--- a/roles/openshift_loadbalancer/tasks/main.yml
+++ b/roles/openshift_loadbalancer/tasks/main.yml
@@ -1,7 +1,9 @@
 ---
+- fail: msg="Cannot use containerized=true for load balancer hosts."
+  when: openshift.common.is_containerized | bool
+
 - name: Install haproxy
   action: "{{ ansible_pkg_mgr }} name=haproxy state=present"
-  when: not openshift.common.is_containerized | bool
 
 - name: Configure systemd service directory for haproxy
   file:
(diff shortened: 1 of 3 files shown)


dgoodwin:
aos-ci-test

roles/openshift_loadbalancer/tasks/main.yml
```diff
@@ -1,7 +1,9 @@
 ---
+- fail: msg="Cannot use containerized=true for load balancer hosts."
```
sdodson: Perhaps suggest adding containerized=false to [lb] in the example hosts file or hosts within it with a note about why its necessary?

sdodson:
I've marked this RFE as CLOSED DEFERRED given we're doing this and we've also archived the trello card requesting containerized LB. https://bugzilla.redhat.com/show_bug.cgi?id=1298787

roles/openshift_loadbalancer/tasks/main.yml
dgoodwin: Updated.

PR #2549, opened 2016-10-05
Cert Expiry for Cluster Ops

tbielawa:
Notify cluster operators of certificates expired or expiring within a warning window.

https://trello.com/c/OpWOZ3rx/124-3-warn-admins-that-their-certificates-will-soon-expire


tbielawa:
Example output from the current HEAD of this branch

https://gist.github.com/tbielawa/76a67e53ff97b8e5ac76c729ac813578

tbielawa:
To whoever checks this out you'll notice that one yaml file has a lot of commented out lines in it. The playbook runs swell and quick with those lines commented out. It also shaves about 40-50 seconds off the run time on my small 3 node cluster.

Do we **need** those lines? They're standard boilerplate, but if they aren't required then I'll cut 'em out right quick.

tbielawa:
ARRRGH. Looks like adding pyOpenSSL as a test dependency broke CI tests. Virtualenv can't compile the module extensions. Looks like system devel library/compiler issues at the core. Will have to debug that tomorrow.

In related news, the module now checks
* master/node service certificates
* master/node kubeconfigs
* etcd certificates

And it generates an HTML output so you can more easily review the results. See attachment.
screenshot from 2016-10-05 15-48-35

tbielawa:
First missing dep explosion:

```
  Running setup.py install for cryptography
    Package libffi was not found in the pkg-config search path.
    Perhaps you should add the directory containing `libffi.pc'
    to the PKG_CONFIG_PATH environment variable
    No package 'libffi' found
    Package libffi was not found in the pkg-config search path.
    Perhaps you should add the directory containing `libffi.pc'
    to the PKG_CONFIG_PATH environment variable
    No package 'libffi' found
    Package libffi was not found in the pkg-config search path.
    Perhaps you should add the directory containing `libffi.pc'
    to the PKG_CONFIG_PATH environment variable
    No package 'libffi' found
    Package libffi was not found in the pkg-config search path.
    Perhaps you should add the directory containing `libffi.pc'
    to the PKG_CONFIG_PATH environment variable
    No package 'libffi' found
    Package libffi was not found in the pkg-config search path.
    Perhaps you should add the directory containing `libffi.pc'
    to the PKG_CONFIG_PATH environment variable
    No package 'libffi' found
    c/_cffi_backend.c:15:17: fatal error: ffi.h: No such file or directory
     #include <ffi.h>
```

This should be provided by the package `libffi-devel`.

tbielawa:
aos-ci-test

tbielawa:
Rebased onto master. Squashed.

I know there is some duplicated code in there, but this is now a working proof of concept. This could use some reviews.

@smunilla @detiber @abutcher @sdodson PTAL

tbielawa:
Example HTML output is uploaded onto here

* http://lnx.cx/~tbielawa/cert-table.html

tbielawa:
Based off of feedback I've received

* I am going to refactor this into a standalone role to make it more modular
* I'm going to improve the options for saving the expiration result data so it can be fed into other system for parsing after collection (zabbix, splunk, etc)

tbielawa:
aos-ci-test

tbielawa:
This feature has been repackaged into a standalone role with 0 dependencies on other openshift-ansible components. The ability to save results in JSON format has been added. I think this PR is ready for review now.

@sdodson @detiber @smunilla PTAL, friends.

tbielawa:
Updating this with a re-fiddled data structure for JSON results and an updated README.

tbielawa:
Rebased onto latest master. Squashed down from 16-ish commits to about 7 now.

tbielawa:
aos-ci-test

tbielawa:
I've created a gist for the QE guys to reference when writing test cases

https://gist.github.com/tbielawa/fde89c7d617e8df55188dc6419bfcd77

tbielawa:
I'm done tweaking this PR now. It's ready for review @sdodson @detiber @smunilla

roles/openshift_certificate_expiry/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+openshift_certificate_expiry_config_base: "/etc/origin"
```
jiajliu: If this mean that we currently can not support check customized certificates which maybe not located in the default path like os.path.join(openshift_base_config_path, "master/master-config.yaml")?
eg: 
I set openshift_certificate_expiry_config_base to "/usr/share/crts", but there is no master/master-config.yaml in that path?

tbielawa: @jiajliu that is correct. The `openshift_certificate_expiry_config_base` key is how you specify where the base **OCP cluster configuration** directory is.

After that, the module I wrote for the role will discover the correct paths to the certificates itself. For example, if you have the following setup:

* OCP config files are installed in `/usr/share/ocp/[master/master-config.yaml, node/node-config.yaml]` 
* Your certificates are in `/usr/share/crts`
* Inside the `master/node-config.yaml` files there are keys like this:

```yaml
servingInfo:
    certFile: /path/to/name-of-cert.crt
    clientCA: /path/to/name-of-ca.crt
```

The module will read the YAML files and then parse the `certFile` and `clientCA` values out from it. If you certificates are in a special location then these two values will say where you certificates are.

In this example if we were reading `/usr/share/ocp/master/master-config.yaml` then the `[servingInfo][certFile]` value would be something like `/usr/share/crts/master.server.crt` and `node-config.yaml` could be something like `/usr/share/crts/server.crt`.

@jiajliu This role is fundamentally different from the proof of concept you saw a few months ago in that where **previously** the playbook would look for `crt` files on the file system, now we read in server config files and then read the certificates that are defined inside of the config files.

That is to say, previously we were just blindly looking for any file that ended in `.crt`, a brute-force approach. This new approach is more precise and only reads files that the service says it explicitly uses.

**Note**: Other certificates are read in different ways. For example, router/registry secrets have to be read from the `oc get secrets` command, there is no file holding the certificate data for those items.

In summary: if your cluster is alive and working then all you have to do is:

* Ensure that the default value (`/etc/origin`) for the `openshift_certificate_expiry_config_base` key is correct
* Or, ensure that you set `openshift_certificate_expiry_config_base` to point to where your actual config base is.

Locating the certificates will happen automatically.

tbielawa:
@sdodson @detiber @smunilla or @abutcher  PTAL, friends

roles/openshift_certificate_expiry/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+    src: cert-expiry-table.html.j2
+    dest: "{{ openshift_certificate_expiry_html_report_path }}"
+  delegate_to: localhost
+  when: "{{ openshift_certificate_expiry_generate_html_report|bool }}"
+
+- name: Generate expiration results JSON
+  become: no
+  run_once: yes
+  template:
+    src: save_json_results.j2
```
detiber: Instead of a template here, it would probably be better to create a filter plugin (with ansible 2.x+ can be embedded in a role) rather than a template.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -0,0 +1,633 @@
...
+# Expiration parsing
+import datetime
+# File path stuff
+import os
+# Config file parsing
+import yaml
+# Certificate loading
+import OpenSSL.crypto
+
+DOCUMENTATION = '''
```
detiber: Not directly related to PR, but I'm wondering if we should start generating module docs for internal use.

tbielawa: I :heart: docs (and comments). They're pretty useful when you need to know what something does :smile:

detiber: https://github.com/ansible/ansible/blob/devel/hacking/module_formatter.py is how Ansible generates their module docs, haven't looked to see if it can be used unmodified or would need to be tweaked for our use.

roles/openshift_certificate_expiry/tasks/main.yml
detiber: @tbielawa main thing I was thinking was the ability to modify as a python dict and serialize directly to json for output to avoid maintenance issues down the line rather than relying on the template to always produce valid json.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
tbielawa: For the modules that we ship I wonder if we should set the `version_added` key to the version of origin/ocp that the module first appeared in. I guess for this case then we'd set it to either `1.3` or `3.3` depending on which you prefer to use.

Also. The doc generator is great. I've ran that on modules I've written before and it works great (it just took a little rejiggering).

detiber: @sdodson thoughts on versioning? I'm not sure 1.3/3.3 is sufficient because we would potentially update mid-stream. Backports would further complicate it.

roles/openshift_certificate_expiry/tasks/main.yml
tbielawa: I think the maintenance costs are the same down the line. If the check results datastructures change then you either have to update a python file or a jinja template file.

I see what you're saying there though. Some of the template logic makes me feel a little gross and the serializing could look prettier w/ some actual python.

I will take a swing at making it into a role filter (I'm **keeping** this role self-contained). I'll also fix some typos I saw in the module code comments which are *killing me* on the inside.

detiber:
@tbielawa love how you typo'd your typo fixing commit message (fix tyops).

filter plugin looks good, too bad it still had to use the template, but at least it isn't relying on the template itself to provide valid json.

tbielawa:
Rebased onto latest master

Squashed new commits down (bye bye tyops).

JSON results are filter-driven now.

Version added set to 1.0, role version set to 1.0.

Added some path checking to avoid potential *file not found* errors.

JSON output is the same as far as data equality is concerned. The only difference from the old output is visual. Indentation and what have you is uniform now.

tbielawa:
aos-ci-test

tbielawa:
@detiber @sdodson CI complete. Any last thoughts on merginating this?

Sidebar: Has anyone actually **tested** this role themselves?

tbielawa:
aos-ci-test

tbielawa:
aos-ci-test

tbielawa:
Rebased onto latest master to get CI to unscrew itself.

tbielawa:
aos-ci-test

tbielawa:
@sdodson @detiber one more time, thoughts on testing/merging?

tbielawa:
Rebased on latest master to pull in the infinite recusion fix. Updated README to use more focused host groups in the example playbooks.

tbielawa:
aos-ci-test

tbielawa:
aos-ci-test

PR #2551, opened 2016-10-05
logging: fix enterprise deployer template

sosiouxme:
It should refer to registry.access, not dockerhub, as the default.


sdodson:
I'm very confused. I fixed this already in https://github.com/openshift/openshift-ansible/pull/2508 on master then https://github.com/openshift/openshift-ansible/pull/2510 on release-1.3 but your PR clearly shows a diff being applied. Am I missing some weird git voodoo?

https://github.com/openshift/openshift-ansible/blob/release-1.3/roles/openshift_examples/files/examples/v1.3/infrastructure-templates/enterprise/logging-deployer.yaml#L203-L209

sosiouxme:
I'm confused too. I actually made my change from a base branch enterprise-3.3 which I don't think is there anymore. When I cherry-pick this commit to the proper branch (release-1.3) I get an empty commit. I don't know how I managed to get a diff out of this. I think we'll just call it good.

sdodson:
Andrew mentioned that maybe it's because your base fork is detiber/openshift-ansible? Anyway, the fix should've shipped to customers on Monday, sorry for the mishap.

PR #2554, opened 2016-10-05
Add support for 3.4 upgrade.

dgoodwin:
This is a direct copy of 3.3 upgrade playbooks, with 3.3 specific hooks
removed and version numbers adjusted appropriately.


dgoodwin:
Yeah upgrade was working but it's been awhile, and I think I might need to sync this with the 3.3 files they originated from for some recent changes. Will update asap.

dgoodwin:
Putting on hold until #2637 comes through, at which point I need to sync that change into the 3.4 files.

Issue #2345, opened 2016-08-23
enable service-serving-cert-signer by default


deads2k:
Commands like `oadm router` want to take advantage of the service-serving-cert-signing controller.  This requires that 
```yaml
controllerConfig:
  serviceServingCert:
    signer:
      certFile: service-signer.crt
      keyFile: service-signer.key
```

in the master-config.yaml be configured with a signer.  This signs for internal DNS names and can be created using a command like:

```bash
oadm ca create-signer-cert --cert=service-signer.crt --key=service-signer.key --serial=service-signer-serial.txt
```

https://github.com/openshift/openshift-docs/pull/2324 has docs about using it.

knobunc:
@pecameron you need this fix.

pecameron:
@knobunc OK

eparis:
changing to p0. unable to deploy 3.3 routers without.

itewk:
For anyone wondering who stumbles on this when you are debuging your OCP 3.3 cluster after an upgrade to 3.2 and wondering why it keeps choking on service-signer.crt missing. Be sure you are using the 3.3.30-1 or later tag of the ansible scripts, anything before that doesn't install the service-signer.crt which is required for a successful upgrade.

itewk:
looks like 3.3.30-1 doesn't work either, trying master.

on 3.3.30-1 the "Upgrade master packages" play happens just before the "Determine if service signer cert must be created" play and that "Upgrade master packages" play fails on the "Restart node service" task because the cert does not exist.

itewk:
master doesn't work either

itewk:
patched: https://github.com/openshift/openshift-ansible/pull/2557

dagwieers:
This bit us when doing the OSE v3.2 tot OCP v3.3 upgrade :-/ The Ansible-driven advanced upgrade process fails to get to the finish successfully. Probably warrants an official fix to Red Hat OCP v3.3 ?

detiber:
@itewk @dagwieers what versions of Ansible are you using? I'm wondering if there is version dependent behavior causing the handlers to be triggered at different times depending on the  version in use.

itewk:
@detiber 

Steps to Reproduce:
[...]

PR #2557, opened 2016-10-06
create service serving certs before upgrading master packages

itewk:
The upgrade master packages will fail if the create service signer certificates needs to be done and has not bee done yet.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -88,17 +88,6 @@
       msg: "Upgrade cannot continue. The following hosts did not complete etcd backup: {{ etcd_backup_failed | join(',') }}"
     when: etcd_backup_failed | length > 0
 
-- name: Upgrade master packages
-  hosts: oo_masters_to_config
-  handlers:
-  - include: ../../../../roles/openshift_master/handlers/main.yml
-    static: yes
-  roles:
-  - openshift_facts
-  tasks:
-  - include: rpm_upgrade.yml component=master
-    when: not openshift.common.is_containerized | bool
-
 - name: Determine if service signer cert must be created
   hosts: oo_first_master
   tasks:
@@ -111,6 +100,17 @@
 # Create service signer cert when missing. Service signer certificate
 # is added to master config in the master config hook for v3_3.
 - include: create_service_signer_cert.yml
+
+- name: Upgrade master packages
+  hosts: oo_masters_to_config
+  handlers:
+  - include: ../../../../roles/openshift_master/handlers/main.yml
+    static: yes
+  roles:
+  - openshift_facts
+  tasks:
+  - include: rpm_upgrade.yml component=master
+    when: not openshift.common.is_containerized | bool
 
 - name: Upgrade master config and systemd units
   hosts: oo_masters_to_config


detiber:
ok to test

detiber:
aos-ci-test

detiber:
@dgoodwin ptal

detiber:
related to https://github.com/openshift/openshift-ansible/issues/2345

dgoodwin:
Hmm would kind of like to see ansible logs with the error and steps to reproduce, the change looks fine but I'd really like to understand why we haven't seen it yet. @itewk could you provide either of those? If not we can likely just merge.

detiber:
@dgoodwin I'm thinking that it might be an issue with handlers being triggered after each play (which I believe is the proper behavior) vs at the playbook level. Could be a but with the 2.2 pre-release packages we are currently using.

dgoodwin:
Also until yesterday there was a weird "node" restart right after we upgraded the master packages, since removed. I'm not clear on how master rpm upgrade could surface any problems with signer certs, so we do probably need a little more info here. 

@itewk could you provide ansible log and steps to reproduce the failure if you're aware of any reason it would surface, and what version of ansible was in use?

itewk:
@dgoodwin and @detiber 

Steps to Reproduce:
The first attempt at upgrade was using the playbooks shipped with the 3.3 RPMs. That version is 3.3.28-1. That version does not contain the Ansible tasks to install the certs if needed. It apeared from that upgrade run that the upgrade worked. But then on a run of the playbooks/byo/config.yml the services attempted to restart and ran into the error.

We then tried using the latest ansible playbooks for 3.3, 3.3.30-1, those scripts have the tasks to install the certs if needed, but not in the correct order. To resolve the issue we had to update the scripts to install the certs before doing the master package upgrade. Patch link below.

1. playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml (3.3.28-1, shipped with 3.3 RPMs)
2. playbooks/byo/update.yml (3.3.28-1, shipped with 3.3 RPMs)
3. run failed due to missing certs
4. download 3.3.30-1 scripts from github
5. playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml (3.3.30-1)
6. Failed for same error because step to install certs is just after the package upgrade which is the step that fails due to restart
7. apply patch (https://github.com/openshift/openshift-ansible/pull/2557)
8. playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml (3.3.30-1 with patch)
9. success
10. playbooks/byo/update.yml (3.3.30-1 with patch)
11. success

dgoodwin:
@itewk do you still have any output from the failure when trying to run upgrade without your patch? I run upgrades daily and haven't seen that so the reproducer must be requiring something additional we're not pinning down yet. I'm curious exactly where and how it fails.

itewk:
@dgoodwin 

Attaching logs for the runs described in the previously provided steps.


[step_10_playbooks_byo_update.yml_3.3.28-1-with-proposed-patch____success.txt](https://github.com/openshift/openshift-ansible/files/529360/step_10_playbooks_byo_update.yml_3.3.28-1-with-proposed-patch____success.txt)
[step_1_playbooks_byo_openshift-cluster_upgradesv3_3_upgrade_yml__3.3.28-1___ansible_success_but_certs_not_installed.txt](https://github.com/openshift/openshift-ansible/files/529362/step_1_playbooks_byo_openshift-cluster_upgradesv3_3_upgrade_yml__3.3.28-1___ansible_success_but_certs_not_installed.txt)
[step_2_playbooks_byo_update.yml_3.3.28-1____run_after_upgrade_failed_result_due_to_missing_certs.txt](https://github.com/openshift/openshift-ansible/files/529361/step_2_playbooks_byo_update.yml_3.3.28-1____run_after_upgrade_failed_result_due_to_missing_certs.txt)
[step_5_playbooks_byo_openshift-cluster_upgrades_v3_3_upgrade.yml_3.3.30-1_____second_run_of_upgrade_script_failed_due_to_missing_scripts.txt](https://github.com/openshift/openshift-ansible/files/529363/step_5_playbooks_byo_openshift-cluster_upgrades_v3_3_upgrade.yml_3.3.30-1_____second_run_of_upgrade_script_failed_due_to_missing_scripts.txt)
[step_8_playbooks_byo_openshift-cluster_upgrades_v3_3_upgrade.yml_3.3.30-1-with-proposed-patch_____third_run_of_upgrade_scripts_with_patch_this_time_success.txt](https://github.com/openshift/openshift-ansible/files/529364/step_8_playbooks_byo_openshift-cluster_upgrades_v3_3_upgrade.yml_3.3.30-1-with-proposed-patch_____third_run_of_upgrade_scripts_with_patch_this_time_success.txt)

dgoodwin:
Excellent thanks @itewk, this sheds a lot of light on what happened. I believe this is the rogue node restart mentioned earlier, in step 5 logfile above we get into "PLAY [Upgrade master packages]", and then "TASK [Restart node service]" where the failure occurs.

This was removed just 2 days ago in 245fef16573757b6e691c448075d8564f5d569f4. 

This PR is very likely safe, but I think unnecessary in this regard, that node restart should not have been happening in the first place and won't until after signer certs have been generated.

Going to close for now. Major thanks @itewk for finding, fixing, and patiently providing all that info.

PR #2569, opened 2016-10-07
[logging] Fix openshift_hosted_logging_fluentd_nodeselector

### LINKED ISSUE #2568: openshift_hosted_logging_fluentd_nodeselector not interpreted when labeling nodes
On line https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted_logging/tasks/deploy_logging.yaml#L141 `openshift_hosted_logging_fluentd_nodeselector` is passed as a literal string to the oc client instead of being replaced with the value of the ansible variable.

##### Version
3.3.30-1

##### Steps To Reproduce
1. Set `openshift_hosted_logging_fluentd_nodeselector` in the inventory

##### Current Result
fluentd labels are not applied, and playbook fails waiting for fluentd to be installed. The executed command is `oc --config=/tmp/openshift-ansible-sLi5bi/admin.kubeconfig label node --overwrite=true -l openshift_hosted_logging_fluentd_nodeselector logging-infra-fluentd=true `

sdodson:
Fixes #2568

### DIFF
--- a/roles/openshift_hosted_logging/tasks/deploy_logging.yaml
+++ b/roles/openshift_hosted_logging/tasks/deploy_logging.yaml
@@ -138,7 +138,7 @@
     delay: 5
 
   - name: "Deploy fluentd by labeling the node"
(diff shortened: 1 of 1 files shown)


sdodson:
@abutcher ptal

PR #2571, opened 2016-10-07
Retry starting master and node services once

sdodson:
Especially for containerized environments the services will fail the first time they're started. We should retry starting them once in order to ensure a higher rate of success.


sdodson:
aos-ci-test

sdodson:
aos-ci-test

PR #2572, opened 2016-10-07
Retry failed master startup once

ironcladlou:
Master startup can fail when ec2 transparently reallocates the block
storage, causing etcd writes to temporarily fail. Retry failures blindly
just once to allow time for this transient condition to to resolve and for
systemd to restart the master (which will eventually succeed).

https://github.com/coreos/etcd/issues/3864
https://github.com/openshift/origin/issues/6065
https://github.com/openshift/origin/issues/6447

### DIFF
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -168,10 +168,21 @@
 - include: set_loopback_context.yml
   when: openshift.common.version_gte_3_2_or_1_2
 
+# TODO: Master startup can fail when ec2 transparently reallocates the block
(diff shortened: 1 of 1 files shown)


ironcladlou:
Still testing this, but wanted to get it ready for discussion.

ironcladlou:
@sdodson @abutcher

sdodson:
aos-ci-test

abutcher:
Will we also want to do this with the controllers service for HA installs?

sdodson:
Yeah, I think we should probably do this for all the cases where we're
starting services for the first time.

sdodson:
I'll get those in my PR.

PR #2573, opened 2016-10-07
Move infrastructure templates into openshift_hosted_templates role.

### LINKED ISSUE #2575: The new secured registry fails on existing installs
- Existing cluster had `openshift_hosted_manage_registry=true`
- Manually add secured registry components by docs
- Re-running the installer with `openshift_hosted_manage_registry=true` after the secured registry

Installation fails due to conflicting secrets volume

abutcher:
Metrics, logging and registry console depend on templates which are created in `openshift_examples`. Examples can be disabled via `openshift_install_examples=false` so roles which depend on these templates should create them outside `openshift_examples`. Adds an `openshift_hosted_templates` role which owns infrastructure/hosted templates.

- [x] registry-console
- [x] metrics
- [x] logging

Fixes #2575.


abutcher:
@sdodson @ironcladlou

abutcher:
aos-ci-test

sdodson:
The only thing that's missing is that openshift_examples performed updates by setting `openshift_examples_import_command: replace` and calling the role. We'd need to decide what to do about that.

abutcher:
aos-ci-test

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -17,7 +17,7 @@
       cp {{ openshift_master_config_dir }}/admin.kubeconfig {{ mktemp.stdout }}/admin.kubeconfig
     changed_when: False
 
-  - name: Check for logging project already exists
+  - name: "Check for logging project already exists"
```
sdodson: This is something we should be doing everywhere?

abutcher: Most of the task names were quoted in this role so I just stuck with the convention.

PR #2590, opened 2016-10-12
add regionendpoint parameter for registry s3

tobru:
This commit adds support for the `regionendpoint` parameter to support any S3 service as registry backend storage. For a description of the parameter, see https://docs.docker.com/registry/storage-drivers/s3/#/parameters

Unfortunately I wasn't able to completely test the change. But as soon as I have the possibility I'll do it. In the meantime if someone of the Ansible experts can have a look into it I would really appreciate it.


roles/openshift_hosted/templates/registry_config.j2
```diff
@@ -13,6 +13,9 @@ storage:
     accesskey: {{ openshift.hosted.registry.storage.s3.accesskey }}
     secretkey: {{ openshift.hosted.registry.storage.s3.secretkey }}
     region: {{ openshift.hosted.registry.storage.s3.region }}
+{%   if openshift.hosted.registry.storage.s3.regionendpoint %}
```
abutcher: Haven't tested but I think this will fail when `regionendpoint` is not set. Checking for the `regionendpoint` key should do the trick.

```
{% if 'regionendpoint' in openshift.hosted.registry.storage.s3 %}
```

tobru: Thanks for the hint, that was that I was looking for... Code is updated and pushed.

inventory/byo/hosts.origin.example
```diff
@@ -351,6 +351,22 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
...
 #openshift_hosted_registry_enforcequota=true
+#
+# Any S3 service (Minio, ExoScale, ...): Basically the same as above
+# but with regionendpoint configured
+# S3 bucket must already exist.
+#openshift_hosted_registry_storage_kind=object
+#openshift_hosted_registry_storage_provider=s3
+#openshift_hosted_registry_storage_s3_accesskey=access_key_id
+#openshift_hosted_registry_storage_s3_secretkey=secret_access_key
+#openshift_hosted_registry_storage_s3_regionendpoint=https;//myendpoint.example.com
```
detiber: should probably be:
```
openshift_hosted_registry_storage_s3_regionendpoint=https://myendpoint.example.com
```
(':' instead of ';')

tobru: Oh, my bad, fixed.

detiber:
ok to test

detiber:
aos-ci-test

detiber:
@cooktheryan if the refarch is using s3 for registry storage, could you validate this pr?

cooktheryan:
@detiber gladly. I've been waiting to try this one

roles/openshift_hosted/templates/registry_config.j2
```diff
@@ -13,6 +13,9 @@ storage:
     accesskey: {{ openshift.hosted.registry.storage.s3.accesskey }}
     secretkey: {{ openshift.hosted.registry.storage.s3.secretkey }}
     region: {{ openshift.hosted.registry.storage.s3.region }}
+{%   if 'regionendpoint' in openshift.hosted.registry.storage.s3 %}
+    regionendpoint: {{ openshift.hosted.registry.storage.s3.regionendpoint }}
+{%   endif -%}
```
cooktheryan: It appears a line spacing issue occurs with this in place. For example

```
    region: us-east-1
bucket: rcook-pr-test
```

@detiber noted that it was the - on line 18

detiber: This should be:
```
{%   endif -%}
```
The '-' is throwing off formatting for the bucket key below by removing leading spaces.

detiber:
@tobru outside of the latest comments around formatting of the processed registry_config template, lgtm.

detiber:
aos-ci-test

PR #2592, opened 2016-10-12
Fix upgrade mappings in quick installer

smunilla:
Update to use the correct playbook for 3.2 to 3.3 upgrades
and some error handling for unsupported upgrades.


smunilla:
aos-ci-test

Issue #2523, opened 2016-09-28
Upgrade error: SDN node startup failed: could not get EgressNetworkPolicies


macedogm:
Hi. I tried to run an ansible playbook upgrade from Origin 1.2.1 to 1.3.0 in a cluster with 01 load balance, 03 masters and 03 nodes over CentOS 7, but it failed while restarting origin-node service.

##### Version
Git repo version: openshift-ansible-3.4.2-1-22-g3391904
Ansible version: ansible 2.2.0

##### Steps To Reproduce
1. Run a playbook upgrade: ansible-playbook openshift-ansible/playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml

2. Upgrade fail with error:
```
TASK [Restart node service] ****************************************************
fatal: [tst-openshift-master01-poa.rbs.com.br]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to restart service origin-node: Job for origin-node.service failed because the control process exited with error code. See \"systemctl status origin-node.service\" and \"journalctl -xe\" for details.\n"}
fatal: [tst-openshift-master03-poa.rbs.com.br]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to restart service origin-node: Job for origin-node.service failed because the control process exited with error code. See \"systemctl status origin-node.service\" and \"journalctl -xe\" for details.\n"}
fatal: [tst-openshift-master02-poa.rbs.com.br]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to restart service origin-node: Job for origin-node.service failed because the control process exited with error code. See \"systemctl status origin-node.service\" and \"journalctl -xe\" for details.\n"}

NO MORE HOSTS LEFT *************************************************************

PLAY RECAP *********************************************************************
localhost                  : ok=18   changed=10   unreachable=0    failed=0   
openshift-tst.rbs.com.br   : ok=56   changed=0    unreachable=0    failed=0   
tst-openshift-master01-poa.rbs.com.br : ok=115  changed=3    unreachable=0    failed=1   
tst-openshift-master02-poa.rbs.com.br : ok=115  changed=3    unreachable=0    failed=1   
tst-openshift-master03-poa.rbs.com.br : ok=115  changed=3    unreachable=0    failed=1   
tst-openshift-node01-poa.rbs.com.br : ok=81   changed=1    unreachable=0    failed=0   
[...]

PR #2593, opened 2016-10-12
Stop restarting node after upgrading master rpms.

dgoodwin:
This sucks a bit but I cannot reproduce the original issue, however several users and customers have hit a problem where node restart causes upgrade to fail. I have one log from such an install and found this, a node restart immediately following a master package upgrade. (note: not a master service restart, which comes later after we setup systemd configs)

In this case the node attempts to use an EgressNetworkPolicy API which does not yet exist because master services are not yet restarted, and the upgrade fails.

The node restart here appears out of place and unnecessary.

This is likely the cause behind #2523. If this does not fix all instances of the problem, there are still workarounds available.


sdodson:
Sounds good to me.

PR #2595, opened 2016-10-12
Fix missing play assignment in a-o-i callback plugin

tbielawa:
If a task fails in the quick installer it calls the `v2_runner_on_failed` method in the callback plugin. I missed setting `self._play` in the first iteration of the callback plugin, which it turns out is critical for the `on_failed` method to work. This fixes that by adding back in the assignment in the `play_start` method.


tbielawa:
aos-ci-test

tbielawa:
@sdodson I can haz review?

PR #2600, opened 2016-10-13
Port openshift_facts to py3

mscherer:
- fix configparser import
- convert bytes to str using to_native
- replace method that did got removed in py3 with six equivalent

With ansible 2.2, python 3 start to be supported enough to
run non trivial playbooks, and so a test run of openshift-ansible on
F24 showed that openshift_facts.py needed to be ported.

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -7,7 +7,13 @@
 
 """Ansible module for retrieving and setting openshift related facts"""
 
-import ConfigParser
+try:
+    # python2
+    import ConfigParser
+except ImportError:
+    # python3
+    import configparser as ConfigParser
+
 import copy
 import io
 import os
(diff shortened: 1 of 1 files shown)


mscherer:
For the record, it then fail on:
```
fatal: [master-1]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Shared connection to 172.25.160.25 closed.\r\n", "module_stdout": "Traceback (most recent call last):\r\n  File \"/tmp/ansible_ilwhwe09/ansible_module_unarchive.py\", line 867, in <module>\r\n    main()\r\n  File \"/tmp/ansible_ilwhwe09/ansible_module_unarchive.py\", line 851, in main\r\n    file_args['path'] = os.path.join(dest, filename)\r\n  File \"/usr/lib64/python3.5/posixpath.py\", line 89, in join\r\n    genericpath._check_arg_types('join', a, *p)\r\n  File \"/usr/lib64/python3.5/genericpath.py\", line 145, in _check_arg_types\r\n    raise TypeError(\"Can't mix strings and bytes in path components\") from None\r\nTypeError: Can't mix strings and bytes in path components\r\n", "msg": "MODULE FAILURE"}
```

I am gonna fix that upstream and continue testing.

mscherer:
So others tests permit to found 2 bugs in ansible, https://github.com/ansible/ansible-modules-core/pull/5265 and https://github.com/ansible/ansible-modules-core/pull/5266 .

mscherer:
And with all patches applied, I was able to finish to deploy openshift on F24 using python 3 on both side for https://fedorahosted.org/cloud/ticket/153

sdodson:
aos-ci-test

dustymabe:
can we get this change in? I hit this today with Fedora 25 atomic host

sdodson:
aos-ci-test

PR #2605, opened 2016-10-14
[logging] Use inventory variables rather than facts

sdodson:
The facts are only used for defaulting of a few items I could probably get rid of them entirely if we so wished. Anyway, this fixes dynamic volume provisioning, tested on GCE.


sdodson:
@abutcher ptal

PR #2608, opened 2016-10-14
Template with_items for upstream ansible-2.2 compat.

abutcher:


abutcher:
aos-ci-test

abutcher:
aos-ci-test

sdodson:
@abutcher this is compatible with 2.1/2.0?

detiber:
@sdodson explicit jinja templates should be compatible for all 2.x releases of Ansible, only 1.9 should be broken with it.

abutcher:
Yep, we left with_items untemplated in a lot of places so that we could support 1.9 and 2.0 at the same time even though 2.0 outputs a warning.

sdodson:
Flagging this for backport to 3.2 in the event that we use 3.2 playbooks to deliver certificate redeploy for 3.1 playbooks we'll need to do an actual backport there.

PR #2612, opened 2016-10-17
[logging] Fix NFS volume binding

sdodson:
- Use automatic PVC creation
 - Workaround PVCs being set to dynamic even when configmap values are set to false (see https://bugzilla.redhat.com/show_bug.cgi?id=1385824 )


sdodson:
@abutcher PTAL

PR #2615, opened 2016-10-18
Set node-labels in kubeletArguments

tbielawa:
This takes `openshift_node_labels` from your inventory and adds them to the `node-config.yaml` file as array items under the `kubeletArguments.node-labels` subkey.

Given an inventory with a host like this:

`hostname .... openshift_node_labels="{'region': 'infra', 'foo': 'bar'}" .....`

This will produce output in your `node-config.yaml` file as such:

```yaml
kubeletArguments:
  node-labels:
  - region=infra
  - foo=bar
```

**CURRENT BUG**
* If `openshift_node_labels` changes (items added/removed) only **additions** are reflected in the `node-config.yaml` file
* Removing labels from the inventory does not result in the labels being removed from the `node-config.yaml` file

Feature requests comes from the Ops team

* https://bugzilla.redhat.com/show_bug.cgi?id=1359848

### DIFF
--- a/roles/openshift_node/tasks/main.yml
+++ b/roles/openshift_node/tasks/main.yml
@@ -9,6 +9,10 @@
     role: "{{ item.role }}"
     local_facts: "{{ item.local_facts }}"
   with_items:
+  # Reset node labels to an empty dictionary.
+  - role: node
+    local_facts:
+      labels: {}
   - role: node
     local_facts:
       annotations: "{{ openshift_node_annotations | default(none) }}"
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -234,7 +234,7 @@ class FilterModule(object):
            arrange them as a string 'key=value key=value'
         """
         if not isinstance(data, dict):
-            raise errors.AnsibleFilterError("|failed expects first param is a dict")
+            raise errors.AnsibleFilterError("|failed expects first param is a dict [oo_combine_dict]. Got %s. Type: %s" % (str(data), str(type(data))))
 
         return out_joiner.join([in_joiner.join([k, str(v)]) for k, v in data.items()])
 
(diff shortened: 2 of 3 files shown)


tbielawa:
Bugs possibly manifesting due to cached facts on the target hosts in `/etc/ansible/facts.d/openshift.fact`.

```json
[root@n01 facts.d]# less openshift.fact
{"node": {"schedulable": true, "labels": {"region": "infra", "ho": "dor"}, ... }}
```

Changing `openshift_node_labels` in your inventory to remove the `'ho': 'dor'` key will not be reflected in the `node-config.yaml`. Adding another key will be reflected.

tbielawa:
One last note I just observed, **changing** a label value **is** reflected in the `node-config.yaml` file.

tbielawa:
The commit added by @abutcher fixes the issue with the `node-config.yaml` file not being updated. However we've discovered a new problem.

Even with the `--overwrite` option to to the labeling command, labels saved in the nodes object (i.e. `oc get node/foo.example.com`) will not be removed. Only added/updated. The node object also stores kube specific labels so we can no just erase them all.

Erasing a label is done by giving the name of the label to remove and adding a `-` to the end of it `kubectl label pods foo bar-` would remove `bar`.

To utilize this functionality we would effectively need to calculate a diff of the current vs. expected new labels and then conditionally run the label remove command.

abutcher:
This tracks current behavior of not removing labels but we will always set the correct labels in node `kubeletArguments` and in cached facts. Adding (optional?) support for label removal can be done in a follow-up PR, imo.

abutcher:
aos-ci-test

tbielawa:
aos-ci-test

abutcher:
@detiber PTAL

tbielawa:
@detiber @sdodson review/merge plz?

detiber:
@wshearn @dak1n1 ptal

roles/openshift_node/tasks/main.yml
```diff
@@ -9,6 +9,10 @@
     role: "{{ item.role }}"
     local_facts: "{{ item.local_facts }}"
   with_items:
+  # Reset node labels to an empty dictionary.
+  - role: node
+    local_facts:
+      labels: {}
```
sdodson: Why was this necessary?

sdodson: Nevermind, it's necessary to pickup deletions.

PR #2616, opened 2016-10-18
Add pvc=true and pv=true

cw-aleks:
In https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/openshift_hosted.yml#L7-L8 checks openshift if any pv or pvc exists.

When you not add the  pvc=true and pv=true into the hosted sections the setup will fail because the pv's & pvc's are not created.

This can be verified in the code, as long as I have understood python right.

oo_persistent_volumes https://github.com/openshift/openshift-ansible/blob/master/filter_plugins/oo_filters.py#L669
create_pv https://github.com/openshift/openshift-ansible/blob/master/filter_plugins/oo_filters.py#L687

oo_persistent_volume_claims https://github.com/openshift/openshift-ansible/blob/master/filter_plugins/oo_filters.py#L733
create_pvc https://github.com/openshift/openshift-ansible/blob/master/filter_plugins/oo_filters.py#L750


abutcher:
Closing and working in #2617.

PR #2579, opened 2016-10-10
Fix config and namespace for registry volume detection

sdodson:

PR #2617, opened 2016-10-18
Add pvc=true and pv=true

cw-aleks:
In https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/openshift_hosted.yml#L7-L8 checks openshift if any pv or pvc exists.

When you not add the pvc=true and pv=true into the hosted sections the setup will fail because the pv's & pvc's are not created.

This can be verified in the code, as long as I have understood python right.

oo_persistent_volumes https://github.com/openshift/openshift-ansible/blob/master/filter_plugins/oo_filters.py#L669
create_pv https://github.com/openshift/openshift-ansible/blob/master/filter_plugins/oo_filters.py#L687

oo_persistent_volume_claims https://github.com/openshift/openshift-ansible/blob/master/filter_plugins/oo_filters.py#L733
create_pvc https://github.com/openshift/openshift-ansible/blob/master/filter_plugins/oo_filters.py#L750

### DIFF
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -324,6 +324,9 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # the storage_host. For example, the remote volume path using these
 # options would be "nfs.example.com:/exports/registry"
 #openshift_hosted_registry_storage_kind=nfs
+# In case the pv & pvc does not exists let openshift create it for you
+#openshift_hosted_registry_storage_create_pv=true
+#openshift_hosted_registry_storage_create_pvc=true
 #openshift_hosted_registry_storage_access_modes=['ReadWriteMany']
 #openshift_hosted_registry_storage_host=nfs.example.com
 #openshift_hosted_registry_storage_nfs_directory=/exports
@@ -397,6 +400,9 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # the storage_host. For example, the remote volume path using these
 # options would be "nfs.example.com:/exports/metrics"
 #openshift_hosted_metrics_storage_kind=nfs
+# In case the pv & pvc does not exists let openshift create it for you
+#openshift_hosted_metrics_storage_create_pv=true
+#openshift_hosted_metrics_storage_create_pvc=true
 #openshift_hosted_metrics_storage_access_modes=['ReadWriteOnce']
 #openshift_hosted_metrics_storage_host=nfs.example.com
 #openshift_hosted_metrics_storage_nfs_directory=/exports
@@ -435,6 +441,9 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # the storage_host. For example, the remote volume path using these
 # options would be "nfs.example.com:/exports/logging"
 #openshift_hosted_logging_storage_kind=nfs
+# In case the pv & pvc does not exists let openshift create it for you
+#openshift_hosted_logging_storage_create_pv=true
+#openshift_hosted_logging_storage_create_pvc=true
 #openshift_hosted_logging_storage_access_modes=['ReadWriteOnce']
 #openshift_hosted_logging_storage_host=nfs.example.com
 #openshift_hosted_logging_storage_nfs_directory=/exports


abutcher:
Hey @cw-aleks, `create_pv` and `create_pvc` are defaulted per component in the `openshift_facts` module [here](https://github.com/openshift/openshift-ansible/blob/9a64840fc11bd466502ae18e556f322447b11781/roles/openshift_facts/library/openshift_facts.py#L1863-L1864) so they shouldn't need to be set explicitly. Which component did you need to set these variables for?

I ran a quick install to confirm. Here's an inventory that just adds an `[nfs]` group which means I should end up with nfs registry storage by default with a pv and pvc for the nfs registry volume.

```
[OSEv3:children]
masters
nodes
nfs
etcd

[OSEv3:vars]
ansible_ssh_user=openshift
ansible_become=true

openshift_master_identity_providers=[{'name': 'allow_all', 'login': 'true', 'challenge': 'true', 'kind': 'AllowAllPasswordIdentityProvider'}]

[masters]
master4.example.com

[etcd]
master4.example.com

[nodes]
master4.example.com openshift_schedulable=true openshift_node_labels="{'region': 'infra'}"

[nfs]
master4.example.com
```

I can see that the pv and pvc were successfully created.

```
[root@master4 ~]# oc get pv
NAME              CAPACITY   ACCESSMODES   STATUS    CLAIM                    REASON    AGE
registry-volume   5Gi        RWX           Bound     default/registry-claim             4m

[root@master4 ~]# oc get pvc
NAME             STATUS    VOLUME            CAPACITY   ACCESSMODES   AGE
registry-claim   Bound     registry-volume   5Gi        RWX           4m
```

cw-aleks:
@abutcher thanks for the test.
Sorry to say that but this was an none valid test.

The comment was in the use case with external nfs server.

So please remove in `[OSEv3:children]` the nfs server and use a nfs server which is setuped out of ocp-e (OpenShift Container platform 3.3).

There is a case open in customer portal for this use case.

We have also added the metrics and the logging to be setuped at install case.

abutcher:
@cw-aleks I'll give it a try, thanks for clarification.

abutcher:
Alright, here is the new test with external pre-configured NFS host. I re-used master4.example.com as my NFS host and moved the install to master5.example.com.

```
[OSEv3:children]
masters
nodes
etcd

[OSEv3:vars]
ansible_ssh_user=openshift
ansible_become=true

openshift_master_identity_providers=[{'name': 'allow_all', 'login': 'true', 'challenge': 'true', 'kind': 'AllowAllPasswordIdentityProvider'}]

# NFS is pre-configured on master4.example.com
openshift_hosted_registry_storage_kind=nfs
openshift_hosted_registry_storage_host=master4.example.com

[masters]
master5.example.com

[etcd]
master5.example.com

[nodes]
master5.example.com openshift_schedulable=true openshift_node_labels="{'region': 'infra'}"
```

Both the pv and pvc were created without needing to set `create_pv` or `create_pvc` explicitly.

```
[root@master5 ~]# oc get pv
oNAME              CAPACITY   ACCESSMODES   STATUS    CLAIM                    REASON    AGE
registry-volume   5Gi        RWX           Bound     default/registry-claim             2m

[root@master5 ~]# oc get pvc
NAME             STATUS    VOLUME            CAPACITY   ACCESSMODES   AGE
registry-claim   Bound     registry-volume   5Gi        RWX           2m
```

cw-aleks:
interesting.
was the pv already there?

When you add
```
openshift_hosted_metrics_deploy=true
openshift_hosted_metrics_storage_kind=nfs
openshift_hosted_metrics_storage_host=master4.example.com
```
and rerun.
Do you have then a metrics pv?

`oc get pv --all-namespaces`

abutcher:
@cw-aleks This host was completely clean prior to install. I'll give metrics a try next -- one difference between metrics and the registry is that the metrics deployer creates the pvc.

cw-aleks:
@abutcher with **This host** you mean the master or the nfs-server?
What I mean is that the registry pv wasn't already in place?
Do you use enterprise 3.3.x ?

Well maybe the metrics deployer creates the pv & pvc but the check much earlier fails.

https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/openshift_hosted.yml#L7-L11

=> WHICH try to call

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_persistent_volumes/tasks/main.yml

but this role is not called, at least on the current install platform.

I have't the current error message on hand but I can put it here tomorrow when I'm at the customer-site.

abutcher:
@cw-aleks Right, I meant the openshift master was clean and that the pv and pvc were created as a result of install.

Another issue was reported with those filters recently but we haven't been able to reproduce. https://github.com/openshift/openshift-ansible/issues/2553

abutcher:
After adding metrics options I've got a `metrics-volume` pv and the pvc was created as a result of the deploy.

```
[root@master5 ~]# oc get pv
NAME              CAPACITY   ACCESSMODES   STATUS    CLAIM                                 REASON    AGE
metrics-volume    10Gi       RWO           Bound     openshift-infra/metrics-cassandra-1             17h
registry-volume   5Gi        RWX           Bound     default/registry-claim                          17h

[root@master5 ~]# oc get pvc --all-namespaces
NAMESPACE         NAME                  STATUS    VOLUME            CAPACITY   ACCESSMODES   AGE
default           registry-claim        Bound     registry-volume   5Gi        RWX           17h
openshift-infra   metrics-cassandra-1   Bound     metrics-volume    10Gi       RWO           3m
```

cw-aleks:
Okay. So what's next to find the reason for the problems and a solution?
How can I help.

To me looks this "**hosted**" stuff a nice idea but not very well tested.

Today happens this also.

```
TASK [openshift_hosted : Add volume to dc/docker-registry] ********************* 
fatal: [tl-os-master-01.local.ergodirekt.de]: FAILED! => {"changed": true, "cmd": ["oc", "volume", "dc/docker-registry", "--add", "--overwrite", "-t", "persistentVolumeClaim", "--claim-name=registry-claim", "--name=registry-storage"], "delta": "0:00:00.091601", "end": "2016-10-19 15:08:24.525838", "failed": true, "rc": 1, "start": "2016-10-19 15:08:24.434237", "stderr": "Error from server: deploymentconfigs \"docker-registry\" not found", "stdout": "", "stdout_lines": [], "warnings": []}
```
The registry is up and have a volume
A `oc volumes "dc/docker-registry"  -n default` shows

```
deploymentconfigs/docker-registry 
  pvc/registry-claim (allocated 5GiB) as registry-storage 
    mounted at /registry
```
Looks like that this check doesn't worked.

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/tasks/registry/storage/persistent_volume.yml#L5

But this could be different in the OCP-E 3.3 version.

Another thing which I have seen is that I must add the metrics_public_url manually into the master-config.yml because of this.

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_master_facts/tasks/main.yml#L82
```
metrics_public_url: "{% if openshift_hosted_metrics_deploy | default(false) %}https://{{ metrics_hostname }}/hawkular/metrics{% endif %}"
```
This means **NO openshift_hosted_metrics_deploy=true** you must add the metrics url manually into the master-config.yml. I mean really?!

Why not something like this.

pseudo code:
```
if (not openshift_hosted_metrics_deploy ) && \
   (defined openshift_master_metrics_public_url) \
     then use openshift_master_metrics_public_url
```
I'm sure that the same stuff will come to the logging component then we are back in the 3.0 time when you have adopt the master-config.yml manually 8-O.

abutcher:
I'd be happy to debug the persistent volume / claims not being created if you could provide some logs and the exact versions you're testing with. Our testing hasn't uncovered the issue yet but it may be related to #2553 and we can debug there if so.

I suspect that the first error from last comment (`Error from server: deploymentconfigs "docker-registry" not found`) is fixed by https://github.com/openshift/openshift-ansible/pull/2579. The command previously did not specify a namespace (`-n default`) and if the namespace had been changed for the root user then ansible was unable to find the docker-registry dc.

@sdodson can you chime in WRT the metrics URL?

cw-aleks:
The used versions are

```
 rpm -qa "*opensh*" 
openshift-ansible-lookup-plugins-3.3.28-1.git.0.762256b.el7.noarch 
atomic-openshift-utils-3.3.28-1.git.0.762256b.el7.noarch 
atomic-openshift-clients-3.3.0.34-1.git.0.83f306f.el7.x86_64 
openshift-ansible-3.3.28-1.git.0.762256b.el7.noarch 
openshift-ansible-filter-plugins-3.3.28-1.git.0.762256b.el7.noarch 
openshift-ansible-playbooks-3.3.28-1.git.0.762256b.el7.noarch 
atomic-openshift-3.3.0.34-1.git.0.83f306f.el7.x86_64 
tuned-profiles-atomic-openshift-node-3.3.0.34-1.git.0.83f306f.el7.x86_64 
atomic-openshift-node-3.3.0.34-1.git.0.83f306f.el7.x86_64 
atomic-openshift-sdn-ovs-3.3.0.34-1.git.0.83f306f.el7.x86_64 
openshift-ansible-docs-3.3.28-1.git.0.762256b.el7.noarch 
openshift-ansible-roles-3.3.28-1.git.0.762256b.el7.noarch 
atomic-openshift-master-3.3.0.34-1.git.0.83f306f.el7.x86_64 
```

Yes we was in another project not the default one.

I think this was the reason for this error.

````
TASK [openshift_hosted : Add volume to dc/docker-registry]
```

But I don't think that was the reason for the not creating pv/pvc issue.

As for the metrics url I will try to make a pr

PR #2621, opened 2016-10-19
Switch from "oadm" to "oc adm" and fix bug in binary sync.

dgoodwin:
Found bug syncing binaries to containerized hosts where if a symlink was
pre-existing, but pointing to the wrong destination, it would not be
corrected.

Switched to using oc adm instead of oadm.

### DIFF
--- a/playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
+++ b/playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
@@ -25,13 +25,13 @@
   tasks:
   - name: Prepare for Node evacuation
     command: >
-      {{ openshift.common.admin_binary }} manage-node {{ openshift.node.nodename }} --schedulable=false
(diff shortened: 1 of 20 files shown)


dgoodwin:
aos-ci-test

sdodson:
aos-ci-test

dgoodwin:
Failure is a test flake: fatal: [10.8.169.57]: FAILED! => {"changed": true, "failed": true, "msg": "XXXXXXXXXXXXXXXXXXXXXXXXXXX/rcm-guest/puddles/RHAOS/AtomicOpenShift/3.4/latest/x86_64/os/Packages/atomic-openshift-tests-3.4.0.12-1.git.0.9ca893a.el7.x86_64.rpm: [Errno 14] HTTP Error 404 - Not Found\nTrying other mirror.

sdodson:
Yeah, yum metadata is bad probably.

detiber:
@sdodson I'm not sure which version of the oc tooling introduced 'oc adm', so we'll want to be mindful about backporting this change.

sdodson:
Nevermind, i see where you adjusted the symlink.

PR #2624, opened 2016-10-19
Reference full vars for registry object storage.

abutcher:
This is to avoid hosted facts stomping more complex variable usage.


cooktheryan:
+1 Tested in my environment. Code worked as expected

PR #2634, opened 2016-10-20
Add logic for deploying or not infrastructure templates

IshentRas:
Give the opportunity to users not to use the rhel/centos IS but still deploying infrastructure templates.
Before the "when" criteria was mapped to the same value "openshift_examples_load_rhel".
Having a dedicated one for infrastructure templates solves that issue.

### DIFF
--- a/roles/openshift_examples/README.md
+++ b/roles/openshift_examples/README.md
@@ -21,13 +21,14 @@ Facts
 Role Variables
 --------------
 
-| Name                                | Default value                                       |                                          |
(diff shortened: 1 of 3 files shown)


sdodson:
https://github.com/openshift/openshift-ansible/pull/2573 moves the infrastructure templates into the roles that use them, registry-console, metrics, and logging respectively. I think that's a better solution, only install them when those roles are used.

IshentRas:
(y)

IshentRas:
Agree +

PR #2636, opened 2016-10-20
Support 3rd party scheduler

enlinxu:
This work is mainly done by Andrew Butcher, during AnsibleFest when I described to him what the problem is and he pulled out the laptop and helped. You Rock Buddy!

I tested the code with the following in my hosts file:
openshift_node_kubelet_args={'config' : ['/etc/kubernetes/manifest']}

openshift_master_scheduler_args={'scheduler-name' : ['foo']}

The scheduler no longer runs, pods stay at pending state. 3rd party scheduler can be put in the config dir, and start as a mirror pod, which is specified as the first parameter for kubelet config.

### DIFF
--- a/roles/openshift_master/templates/master.yaml.v1.j2
+++ b/roles/openshift_master/templates/master.yaml.v1.j2
@@ -131,6 +131,7 @@ kubernetesMasterConfig:
   proxyClientInfo:
     certFile: master.proxy-client.crt
     keyFile: master.proxy-client.key
+  schedulerArguments: {{ openshift_master_scheduler_args | default(None) | to_padded_yaml( level=3 ) }}
   schedulerConfigFile: {{ openshift_master_scheduler_conf }}
   servicesNodePortRange: ""
   servicesSubnet: {{ openshift.common.portal_net }}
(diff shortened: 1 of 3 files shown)


detiber:
ok to test

detiber:
aos-ci-test

detiber:
@abutcher We'll need to document this behavior, and should probably get some test cases fed to QE for it as well.

abutcher:
@enlinxu Thanks for opening this and sorry for the late response! Do you think it would make sense to also manage the scheduler config placed in the directory we create?

enlinxu:
@abutcher You mean put the scheduler config into kubelet config directory, like /etc/kubernetes/manifest in my example? That makes sense

abutcher:
@enlinxu Yeah, maybe a var that picks up a list of scheduler files or full config object definitions to land? I can take a look at extending soon.

enlinxu:
@abutcher Thanks man, let me know if you want me to add anything else in this PR.

enlinxu:
@abutcher Please let me know if you need anything else from me (docs, use cases, tests), in order to get this PR merged. Thanks

detiber:
aos-ci-test

detiber:
@abutcher ptal

PR #2637, opened 2016-10-20
Fix and reorder control plane service restart.

dgoodwin:
This was missed in the standalone upgrade control plane playbook.

However it also looks to be out of order, we should restart before
reconciling and upgrading nodes. As such moved the restart directly into
the control plane upgrade common code, and placed it before
reconciliation.

### DIFF
--- a/playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
+++ b/playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
@@ -92,10 +92,9 @@
   vars:
     master_config_hook: "v3_3/master_config_upgrade.yml"
 
+- include: ../../../../common/openshift-cluster/upgrades/post_control_plane.yml
+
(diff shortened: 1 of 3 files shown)


playbooks/byo/openshift-master/restart.yml
```diff
@@ -14,5 +14,3 @@
   gather_facts: no
   tasks:
   - include_vars: ../../byo/openshift-cluster/cluster_hosts.yml
-
-- include: ../../common/openshift-master/restart.yml
```
abutcher: This include should remain here for standalone use of the playbook, right?

dgoodwin: Good thinking yes, thanks.

dgoodwin: Updated.

dgoodwin:
@sdodson fyi this one needs to go in before 3.4 upgrade ideally.

PR #2643, opened 2016-10-21
Add VolumeZonePredicate to scheduler predicates

andrewklau:
VolumeZonePredicate ensures pods are placed in the same zone as their corresponding PVC.

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1751,6 +1751,7 @@ class OpenShiftFacts(object):
                 {"name": "PodFitsPorts"},
                 {"name": "NoDiskConflict"},
                 {"name": "NoVolumeZoneConflict"},
+                {"name": "VolumeZonePredicate"},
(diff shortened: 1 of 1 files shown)


abutcher:
aos-ci-test

abutcher:
ok to test

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1751,6 +1751,7 @@ def get_defaults(self, roles, deployment_type, deployment_subtype):
                 {"name": "PodFitsPorts"},
                 {"name": "NoDiskConflict"},
                 {"name": "NoVolumeZoneConflict"},
+                {"name": "VolumeZonePredicate"},
```
detiber: This should be handled by 'NoVolumeZoneConflict'

andrewklau: I think you may be right, I was looking at ways to resolve issues with multizone volumes. Ideal fix seems to be in 1.4.

PR #2630, opened 2016-10-19
Fix maximum recursion depth exceeded in default callback

abutcher:
Hopefully fixes #2622

/cc @tbielawa @jaryn

PR #2518, opened 2016-09-28
Reduce warning output

abutcher:
Especially from the command module since we are using the command module rather than the alternative for a reason in most (if not all) cases.

There are a few warnings not covered here such as from the standalone registry deployment tasks in `openshift_hosted` that are removed in https://github.com/openshift/openshift-ansible/pull/2475.

PR #2651, opened 2016-10-24
Backport ansible-2.2 fixes

### LINKED ISSUE #2479: TASK openshift_node : Configure Proxy Settings - Fails
We got this when deploying 2 Master nodes cluster from Ansible into Atomic Hosts.
Even if we define or not proxy cluster vars, it fails. We don't use proxy in our env.

OUTPUT:

```
TASK [openshift_node : Configure Proxy Settings] *******************************
task path: /drake/openshift-ansible/roles/openshift_node/tasks/systemd_units.yml:51
fatal: [master2.demo.<mydomain>.com]: FAILED! => {
    "failed": true, 
    "msg": "'dict object' has no attribute 'http_proxy'"
}

RUNNING HANDLER [openshift_node : restart openvswitch] *************************

RUNNING HANDLER [openshift_node : restart node] ********************************
    to retry, use: --limit @/drake/openshift-ansible/playbooks/byo/config.retry

localhost                  : ok=12   changed=7    unreachable=0    failed=0   
master1.demo.<mydomain>.com : ok=169  changed=20   unreachable=0    failed=1   
master2.demo.<mydomain>.com : ok=220  changed=41   unreachable=0    failed=1
```

sdodson:
Fixes Bug 1386654

Fixes #2479 
Backports #2518 
Backports #2630

Per https://bugzilla.redhat.com/show_bug.cgi?id=1386654#c5 "with_ clauses ALWAYS
evaluate before the when clause, otherwise we cannot use item and condition loop
iterations. So this is not a bug, just how ansible works. The default and
ternary filters can be used to avoid the undefined error."

### DIFF
--- a/roles/openshift_node/tasks/systemd_units.yml
+++ b/roles/openshift_node/tasks/systemd_units.yml
@@ -56,12 +56,12 @@
     create: true
   with_items:
     - regex: '^HTTP_PROXY='
-      line: "HTTP_PROXY={{ openshift.common.http_proxy }}"
+      line: "HTTP_PROXY={{ openshift.common.http_proxy | default('') }}"
     - regex: '^HTTPS_PROXY='
-      line: "HTTPS_PROXY={{ openshift.common.https_proxy }}"
+      line: "HTTPS_PROXY={{ openshift.common.https_proxy | default('') }}"
     - regex: '^NO_PROXY='
-      line: "NO_PROXY={{ openshift.common.no_proxy | join(',') }},{{ openshift.common.portal_net }},{{ hostvars[groups.oo_first_master.0].openshift.master.sdn_cluster_network_cidr }}"
-  when: "{{ openshift.common.http_proxy is defined and openshift.common.http_proxy != '' }}"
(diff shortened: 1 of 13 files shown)


sdodson:
Probably need to rebase this onto master then backport to 3.3 and 3.2, but running CI jobs real quick.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
@abutcher PTAL, also any other Ansible 2.2 compat items we may have missed?

abutcher:
@sdodson Templated with_items already backported so I think the only thing left would be #2630.

sdodson:
aos-ci-test

sdodson:
`make ci` doesn't exist on this branch, that's why this one is failing.  @abutcher or @detiber PTAL need to get this in today if possible.

roles/openshift_node/tasks/systemd_units.yml
```diff
@@ -56,12 +56,12 @@
     create: true
   with_items:
     - regex: '^HTTP_PROXY='
-      line: "HTTP_PROXY={{ openshift.common.http_proxy }}"
+      line: "HTTP_PROXY={{ openshift.common.http_proxy | default('') }}"
```
gundalow: Does this fix the issues reported in https://bugzilla.redhat.com/show_bug.cgi?id=1386654

PR #2652, opened 2016-10-24
Update defaults for clusterNetworkCIDR & hostSubnetLength

smunilla:
Per https://github.com/openshift/openshift-docs/issues/1700:

The default values for pod networking have changed:
    - clusterNetworkCIDR now defaults to 10.128.0.0/14 (10.128.0.0 - 10.131.255.255)
      rather than 10.1.0.0/16.
    - hostSubnetLength now defaults to 9 rather than 8
      (meaning each node will be assigned a /23 subnet rather than a /24)

Fixes Bug 1320952

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -901,10 +901,24 @@ def set_sdn_facts_if_unset(facts, system_facts):
             facts['common']['sdn_network_plugin_name'] = plugin
 
     if 'master' in facts:
-        if 'sdn_cluster_network_cidr' not in facts['master']:
-            facts['master']['sdn_cluster_network_cidr'] = '10.1.0.0/16'
-        if 'sdn_host_subnet_length' not in facts['master']:
-            facts['master']['sdn_host_subnet_length'] = '8'
+        # set defaults for sdn_cluster_network_cidr and sdn_host_subnet_length
+        # these might be overridden if they exist in the master config file
+        facts['master']['sdn_cluster_network_cidr'] = '10.128.0.0/14'
+        facts['master']['sdn_host_subnet_length'] = '9'
+
+        master_cfg_path = os.path.join(facts['common']['config_base'],
+                                       'master/master-config.yaml')
+        if os.path.isfile(master_cfg_path):
+            with open(master_cfg_path, 'r') as master_cfg_f:
+                config = yaml.safe_load(master_cfg_f.read())
+
+            if 'networkConfig' in config:
+                if 'clusterNetworkCIDR' in config['networkConfig']:
+                    facts['master']['sdn_cluster_network_cidr'] = \
+                        config['networkConfig']['clusterNetworkCIDR']
+                if 'hostSubnetLength' in config['networkConfig']:
+                    facts['master']['sdn_host_subnet_length'] = \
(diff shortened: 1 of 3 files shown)


smunilla:
aos-ci-test

smunilla:
@sdodson PTAL

abutcher:
See #2145 original change for this w/ some discussion.

smunilla:
@abutcher informs me there are some potential complications. Marking this WIP until I work those out.

smunilla:
After much assistance from @abutcher this is tested and ready to go. We'll set the new defaults on new environments, but leave them alone if they are already set.

smunilla:
aos-ci-test

smunilla:
The CI failure looks like it might be unrelated to this change. Giving it another run.

smunilla:
aos-ci-test

smunilla:
aos-ci-test

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -901,10 +901,24 @@ def set_sdn_facts_if_unset(facts, system_facts):
...
+            with open(master_cfg_path, 'r') as master_cfg_f:
+                config = yaml.safe_load(master_cfg_f.read())
+
+            if 'networkConfig' in config:
+                if 'clusterNetworkCIDR' in config['networkConfig']:
+                    facts['master']['sdn_cluster_network_cidr'] = \
+                        config['networkConfig']['clusterNetworkCIDR']
+                if 'hostSubnetLength' in config['networkConfig']:
+                    facts['master']['sdn_host_subnet_length'] = \
+                        config['networkConfig']['hostSubnetLength']
```
sdodson: @smunilla -- After we merged @abutcher astutely pointed out that this loses the ability to set the value via inventory variables.  Can we re-introduce that? Something like this

```
        sdn_cluster_network_cidr = '10.128.0.0/14'
        sdn_host_subnet_length= '9'

        master_cfg_path = os.path.join(facts['common']['config_base'],
                                       'master/master-config.yaml')
        if os.path.isfile(master_cfg_path):
            with open(master_cfg_path, 'r') as master_cfg_f:
                config = yaml.safe_load(master_cfg_f.read())

            if 'networkConfig' in config:
                if 'clusterNetworkCIDR' in config['networkConfig']:
                    sdn_cluster_network_cidr = \
                        config['networkConfig']['clusterNetworkCIDR']
                if 'hostSubnetLength' in config['networkConfig']:
                    sdn_host_subnet_length = \
                        config['networkConfig']['hostSubnetLength']

        if 'sdn_cluster_network_cidr' not in facts['master']:
            facts['master']['sdn_cluster_network_cidr'] = sdn_cluster_network_cidr
        if 'sdn_host_subnet_length' not in facts['master']:
            facts['master']['sdn_host_subnet_length'] = sdn_host_subnet_length
```

sdodson: With this implementation there's a chance that the config file has different values from what's in the inventory. We could try to throw an exception but I think it's alright to just accept what's set in inventory variables.

sdodson: https://github.com/openshift/openshift-ansible/pull/2752

smunilla:
@sdodson oh yeah, totally an easy fix

PR #2654, opened 2016-10-24
Remove Vagrantfile

detiber:
- Remove Vagrantfile
- Update README_vagrant.md to point to openshift-ansible-contrib for a
  vagrant-based installation


detiber:
@sdodson ptal

PR #2663, opened 2016-10-25
Add support for 3.4 upgrade.

dgoodwin:
This is a direct copy of 3.3 upgrade playbooks, with 3.3 specific hooks
removed and version numbers adjusted appropriately.

Wasn't able to re-open my previous PR for this (#2554) as I force pushed after closing. The change yesterday was to sync up with the changes in #2637 which are now a dependency of this and should be merged first.


dgoodwin:
cc @dav1x had to reopen here.

dav1x:
@dgoodwin locally merged this PR and tested. works great! Thanks!

PR #2611, opened 2016-10-17
Build full node config path in systemd_units tasks.

abutcher:
The `openshift_node_config_file` var isn't available from `openshift_node` role vars when included in upgrade playbooks with ansible 2.2 rc1.

@sdodson ptal

PR #2664, opened 2016-10-25
[upgrades] Fix containerized node

sdodson:
This file was removed and no longer used in #2611

Fixes Bug 1387529


sdodson:
@dgoodwin PTAL

dgoodwin:
Looks good to me.

PR #2666, opened 2016-10-25
ansible playbooks continue running after get error

tbielawa:
Fail on any errors when configuring nodes.

Fixes:
- BZ1377619 - https://bugzilla.redhat.com/show_bug.cgi?id=1377619

Adds the `any_errors_fatal` key to the node configuration plays. This causes a complete abort of the installation if any task under those plays fails. This is in the spirit of the "fail fast" system of design.

Without this parameter set to a boolean true value then the installation continues until the master attempts to register the failed node. That results in a waste of time for the user.

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -47,6 +47,7 @@
 
 - name: Configure containerized nodes
   hosts: oo_containerized_master_nodes
+  any_errors_fatal: true
   serial: 1
   vars:
     openshift_node_master_api_url: "{{ hostvars[groups.oo_first_master.0].openshift.master.api_url }}"
@@ -87,6 +88,7 @@
 
 - name: Configure nodes
   hosts: oo_nodes_to_config:!oo_containerized_master_nodes
+  any_errors_fatal: true
   vars:
     openshift_node_master_api_url: "{{ hostvars[groups.oo_first_master.0].openshift.master.api_url }}"
     openshift_node_first_master_ip: "{{ hostvars[groups.oo_first_master.0].openshift.common.ip }}"


tbielawa:
aos-ci-test

tbielawa:
PTAL @sdodson, @detiber, @abutcher (since I took the original bug from you)

tbielawa:
Example of how the behavior changes from the user perspective is shown in comment 3 in https://bugzilla.redhat.com/show_bug.cgi?id=1377619#c3

tbielawa:
Obvious next question

* Where else, if anywhere, would we want to apply this parameter in the future?

detiber:
Does this cause the entire install to fail if a single node fails to install properly? If so, then we don't want to do this, since larger scale installs are likely to have some failures.

tbielawa:
@detiber 
> Does this cause the entire install to fail if a single node fails to install properly? If so, then we don't want to do this, since larger scale installs are likely to have some failures.

That is exactly what this PR does. It addresses the concern from the original bug, but what you pointed out also concerned me when I was writing this.

The reporter was doing a single master single node deployment and ran into an issue with the node getting screwed up mid configuration. I can understand why they'd want the install to quit earlier, but like you said, for larger clusters that may not be ideal behavior.

I reverted the patch and tried a 3 host install, 1 master, 2 nodes, and then I forced the second node into a state where it will error out during the node configuration play.

In this test the second node failed like it did previously, and the first node continued with its configuration (as per usual). Eventually we get to "wait for node reg" and it attempts to find the broken node 50 times. Eventually the master gives up trying to find the node. That's when this happens:

> FAILED - RETRYING: TASK: openshift_manage_node : Wait for Node Registration (1 retries left).
> failed: [m01.example.com] (item=n02.example.com) => {"attempts": 50, "changed": false, "cmd": ["oc", "get", "node", "n02.example.com", "--config=/tmp/openshift-ansible-l6TSVf/admin.kubeconfig", "-n", "default"], "delta": "0:00:00.094149", "end": "2016-10-25 15:47:01.351489", "failed": true, "item": "n02.example.com", "rc": 1, "start": "2016-10-25 15:47:01.257340", "stderr": "Error from server: nodes \"n02.example.com\" not found", "stdout": "", "stdout_lines": [], "warnings": []}

And then the installation quits.

This is the same final result **as if the patch had been applied**.

The difference is, with the patch applied we would have saved at least 4 minutes (50 retries * 5 second delay) + the time in between when the node failed configuration to when the master tries to find the nodes.

tbielawa:
@detiber @sdodson 

I think on IRC we talked about this and the consensus was... I can't remember the consensus... Something to do with what Jason said, we don't want to kill huge installs. But then we observed that the entire install fails anyway on the next steps if a node fails. Then we started talking about concurrency and forks.

So in conclusion, we still need an actual conclusion/consensus for what to do with this PR. Please advise.

sdodson:
Can we make it so that the task that waits for nodes to register moves on after removing failed nodes from the group so that subsequent tasks are skipped?

abutcher:
@tbielawa ^ yeah, what @sdodson said. These tasks need to ensure the variable they are accessing exists in a when statement. The second task looks like it's covered. https://github.com/openshift/openshift-ansible/blob/9e90b0085b1203830164c7ea515e69236f16de76/roles/openshift_manage_node/tasks/main.yml#L15-L33

tbielawa:
this pr is deprecated. See the referenced PR above.

PR #2669, opened 2016-10-26
Change HA master controller service to restart always.

dgoodwin:
Restarts of systemd-journald can cause the master controllers service to
die in HA environments, due to a SIGPIPE and how it's handled.
(seemingly by default in golang)

on-failure does not trigger in this situation, instead we set to
Restart=always in the systemd unit, which matches the non-HA service
definitions as well.

See https://bugzilla.redhat.com/show_bug.cgi?id=1378929 for more info.


sdodson:
Any reason not to get `roles/openshift_master/templates/docker-cluster/atomic-openshift-master-controllers.service.j2` too?

dgoodwin:
I had it changed there originally but then got to thinking I should limit changes and it should be unaffected as I believe it just logs through docker. No real reason not to though I suppose, I will enable it.

dgoodwin:
Updated.

sdodson:
aos-ci-test

sdodson:
Known bug in 3.4 prevenging e2e from working properly, going to try to find out what's up with that.

PR #2673, opened 2016-10-27
Add node redeploy playbook for (re)provisioning on top of gold images

andrewklau:
Deploying with gold images cuts the install process significantly. 

This playbook removes a lot of the installation processes/checks from the scaleup playbook so new nodes can be added in about 1 minute compared to 5-15 minutes with the scaleup playbook.

This has worked in my test setup, but not sure if I missed anything.

### DIFF
--- a/playbooks/byo/openshift-node/redeploy.yml
+++ b/playbooks/byo/openshift-node/redeploy.yml
@@ -1,0 +1,18 @@
+---
+- hosts: localhost
+  connection: local
+  become: no
+  gather_facts: no
+  tasks:
+  - include_vars: ../../byo/openshift-cluster/cluster_hosts.yml
+  - add_host:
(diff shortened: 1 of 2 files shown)


andrewklau:
Node labels seem to be part part of the node-config now and I believe they are schedulable by default, so I have not included the `openshift_manage_node` role

The node gold images were created by:

- running the node scaleup.yaml
- stopping the node service
- removing config from /etc/origin/node/* and /etc/origin/openvswitch/*
- remove random-seed, ssh keys etc.
- power off and snapshot

andrewklau:
Perhaps an alternative could be to use tags to skip tasks. But all the skips takes time, this so far lets us scale up a new node in about 2 min (1 min being this playbook).

Issue #2152, opened 2016-07-14
Introduce concept of pre-flight sanity playbook


jeremyeder:
The use-case is laid out in the trello card [1], but I'll copy it here:
The openstack performance and scale team has developed a set of
pre-flight sanity checks for RHEL OSP.
We could do the same for openshift (does something already exist?)

https://github.com/openstack/browbeat/tree/master/ansible/check

A pre-flight playbook could do things like: - verify nodes are reachable in the first place (sometimes in the real
world, not all nodes are equal. timeouts for ansible may have to be
increased).
- verify subscriptions or access to repos
- verify supported versions of dependent packages
- check against known bugs
- check against known configuration issues (oo-diagnostics ?)
- verify your planned inventory file is sane
- support an install "dry-run"?

[1] https://trello.com/c/3Zb2d9zl/98-3-cfme-integration-prerequisites-playbook

timothysc:
> verify your planned inventory file is sane

- check hosts are up and running 
- walk-through or hand-lint inventory file settings (ip-addresses, collisions, etc.) 
...

detiber:
We definitely want to do some sanity checking for conflicting variable settings to avoid discovering those down the line.

Also I like the idea of doing some testing around detected host IP addresses and the service/pod networks to make sure that there isn't a conflict there. Obviously we can't get all use cases, but if we can catch the obvious ones it would be a huge win.

jeremyeder:
And lint ansible.cfg against known limitations such as # of forks.

detiber:
@jeremyeder validating ansible.cfg may be a difficult task. Some of the values are bubbled up in a way that they can be tested, but others like the value of forks, do not get used outside of the ansible-playbook command itself. Challenge will be to see if there is a way to get the value without diving too deep into internal-use only code that might be fragile with different ansible verisons going forward.

jeremyeder:
There really should be a way for playbooks to introspect the ansible environment they're working in.  Surprised that there isn't one :-( ... hopefully we can do something short term and perhaps file something with Ansible to provide that guaranteed interface?

timothysc:
Gather state and dump before upgrade.

tpoitras:
Hey, so I have a few more days left as a writer on the OpenShift documentation team before I leave the company, but this issue echoes an idea I've been chewing on. Initially I was thinking about a script that would be run pre-install, but this pre-flight sanity playbook is a similar idea. I'd hope something could be scripted for the quick install method as well.

[...]

PR #2674, opened 2016-10-27
ensure/enable dependencies needed for FC24

chrrrles:
These packages are required in order to install on Fedora Cloud 24 -- as well, NetworkManager needs to be enabled and running.

### DIFF
--- a/roles/openshift_facts/tasks/main.yml
+++ b/roles/openshift_facts/tasks/main.yml
@@ -10,12 +10,19 @@
 - set_fact:
     l_is_containerized: "{{ (l_is_atomic | bool) or (containerized | default(false) | bool) }}"
 
-- name: Ensure PyYaml is installed
-  action: "{{ ansible_pkg_mgr }} name=PyYAML state=present"
+- name: Ensure required packages are installed
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=present"
+  with_items:
+    - PyYAML
+    - yum-utils
+    - dbus-python
+    - libselinux-python
+    - libsemanage-python
+    - NetworkManager
   when: not l_is_atomic | bool
 
-- name: Ensure yum-utils is installed
-  action: "{{ ansible_pkg_mgr }} name=yum-utils state=present"
+- name: Enable NetworkManager
+  service: name=NetworkManager enabled=yes state=started
   when: not l_is_atomic | bool
 
 - name: Gather Cluster facts and set is_containerized if needed


roles/openshift_facts/tasks/main.yml
```diff
@@ -10,12 +10,19 @@
...
+    - dbus-python
+    - libselinux-python
+    - libsemanage-python
+    - NetworkManager
   when: not l_is_atomic | bool
 
-- name: Ensure yum-utils is installed
-  action: "{{ ansible_pkg_mgr }} name=yum-utils state=present"
+- name: Enable NetworkManager
+  service: name=NetworkManager enabled=yes state=started
```
chrrrles: Hmm - I agree, but on that point it might be good to move all package / distribution oriented tasks out of openshift_facts and into a pre-req playbook/task - this would make it easier to adapt openshift-ansible for Debian, etc...

sdodson: Yeah, we have a near term goal of moving that sort of stuff out of openshift_facts into a pre-req playbook so it's not called a dozen times or more. There's also #2152 that's related to this.

roles/openshift_facts/tasks/main.yml
```diff
@@ -10,12 +10,19 @@
...
     l_is_containerized: "{{ (l_is_atomic | bool) or (containerized | default(false) | bool) }}"
 
-- name: Ensure PyYaml is installed
-  action: "{{ ansible_pkg_mgr }} name=PyYAML state=present"
+- name: Ensure required packages are installed
+  action: "{{ ansible_pkg_mgr }} name={{ item }} state=present"
+  with_items:
+    - PyYAML
+    - yum-utils
+    - dbus-python
```
mscherer: That will install python 2 version of the dependencies, so I would suggest doing something like this:
```
    with_items:
    - "libsemanage-python{{ ansible_python['version']['major']}}"
    - "python{{ ansible_python['version']['major']}}-dbus"
```

So this can be safely run on python 3 (I am currently testing ansible on python 3, and I was able to deploy on F24 with python 3 installed and used with ansible 2.2)

mscherer:
Also related https://github.com/openshift/openshift-ansible/pull/2840 (since F25 systemd is a bit less forgiving when NetworkManager is not installed)

mscherer:
I also took a look at F25, and there is no NetworkManager, and you can't add dispatcher script to systemd-networkd (by design).

roles/openshift_facts/tasks/main.yml
detiber: There is also this PR that was recently merged that provides a byo playbook for installing and enabling NetworkManager: https://github.com/openshift/openshift-ansible/pull/2771

detiber:
@chrrrles @mscherer is this still relevant?

chrrrles:
@detiber Hello.  I don't feel this PR is relevant anymore -- as I am setting up the prequisites outside of openshift-ansible.  Closing...

PR #2677, opened 2016-10-27
Docker daemon is started prematurely.

EricMountain-1A:
Docker service is started prior to configuration changes being applied.
The service is then not restarted by the handlers, so configuration
changes are not applied.

We now start the docker service only once all config changes have been
made.

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -54,17 +54,6 @@
     dest: "{{ docker_systemd_dir }}/custom.conf"
     src: custom.conf.j2
 
-- name: Start the Docker service
-  systemd:
-    name: docker
-    enabled: yes
-    state: started
-    daemon_reload: yes
-  register: start_result
-
-- set_fact:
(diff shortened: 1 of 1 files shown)


roles/docker/tasks/main.yml
```diff
@@ -113,4 +103,14 @@
   notify:
     - restart docker
```
stevekuznetsov: Wouldn't this have restarted the daemon after configuration?

EricMountain-1A: Not the very first time round, because the task `Start the Docker service` starts docker, registers the result in `start_result` and then we have:

```
- set_fact:
    docker_service_status_changed: start_result | changed
```

And the `restart docker` handler only runs

```
  when: not docker_service_status_changed | default(false) | bool
```

EricMountain-1A:
Thanks @abutcher , rebased.

abutcher:
aos-ci-test

PR #2679, opened 2016-10-27
Add requirements, fix a small formatting issue.

erinn:
Ansible can be run with jinja <= 2.7 but jinja 2,7 is required for some of the iptables modules (at least), specifically the map filter: http://jinja.pocoo.org/docs/dev/changelog/#version-2-7


abutcher:
@sdodson ptal

PR #2681, opened 2016-10-27
Bug 1388191 - Default docker log driver should be journald

abutcher:
Fixes Bug 1388191
https://bugzilla.redhat.com/show_bug.cgi?id=1388191

### DIFF
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -79,7 +79,7 @@ openshift_release=v1.4
 # Disable pushing to dockerhub
 #openshift_docker_disable_push_dockerhub=True
 # Items added, as is, to end of /etc/sysconfig/docker OPTIONS
-# Default value: "--log-driver=json-file --log-opt max-size=50m"
+# Default value: "--log-driver=journald"
 #openshift_docker_options="-l warn --ipv6=false"
 
 # Specify exact version of Docker to configure or upgrade to.
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -79,7 +79,7 @@ openshift_release=v3.4
 # Disable pushing to dockerhub
 #openshift_docker_disable_push_dockerhub=True
 # Items added, as is, to end of /etc/sysconfig/docker OPTIONS
-# Default value: "--log-driver=json-file --log-opt max-size=50m"
+# Default value: "--log-driver=journald"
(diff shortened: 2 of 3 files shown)


abutcher:
aos-ci-test

abutcher:
@tbielawa PTAL

tbielawa:
Discussing potential docker restart issue on IRC now.

abutcher:
Changing the default here will cause a node-wide restart of docker and we're considering what options we have to reorganize this.

detiber:
@abutcher @tbielawa: I think we'll want to preserve the existing value for existing users, new users should get the journald logging and we should handle the transition as part of the 1.4/3.4 upgrade (@dgoodwin)

dgoodwin:
Upgrade is a very safe place to apply it. 

We just need a way to roll it out for new 3.4 clusters as well, but *not* on re-run's of config.yml. I can't think of any place we do that today offhand.

dgoodwin:
02:41pm <jdetiber> dgoodwin: we could do it at a release boundary...  all new 3.4 clusters will have it by default
02:41pm <jdetiber> 3.3 and earlier would be the old value
02:41pm <jdetiber> since we are forking
02:42pm <jdetiber> err branching

abutcher:
@dgoodwin So, this change along with some upgrade logic would only go into master + 3.4 branches?

richm:
LGTM - I've been testing with `openshift_docker_options: '--log-driver=journald'` for a few months now.

sdodson:
@richm -- From the duped PR you said
> If you change this on an existing cluster, you will need to restart fluentd so that it knows to use journald as the input source. That is, on the node, when you restart docker to use journald, you should restart fluentd on that node too. When upgrading to 3.4, will all of the logging components be restarted/redeployed?

We don't really have logging upgrades automated so we'd instruct people to follow the documentation to upgrade logging. I assume fluentd is covered there? If not would simply deleting all fluentd pods after the upgrade be sufficient?

richm:
@sdodson I'm not sure what logging upgrade does - it may restart fluentd.  But at any rate, yes, simply deleting all of the fluentd pods would work.

ewolinetz:
@sdodson the upgrade script will delete the daemonset (if it exists) and recreate it afterwards. So as long as the nodes are [still] labeled and the matchSelector hasn't changed it redeploys Fluentd.

sdodson:
aos-ci-test

abutcher:
aos-ci-test

PR #2683, opened 2016-10-28
Update master_lb vs cluster_hostname workflow

smunilla:
Previously, the messaging around setting cluster_hostname
for single master setups was confusing. This should help with
that as well as removing unecessary values from the cfg file.


smunilla:
aos-ci-test

smunilla:
aos-ci-test

PR #2685, opened 2016-10-29
Set cloudprovider kind with openshift_facts.

ganhuang:
Fix https://bugzilla.redhat.com/show_bug.cgi?id=1389928


ganhuang:
Close it as the bug has already been fixed.

PR #2686, opened 2016-10-31
Touch all ini_file files before using them

### LINKED ISSUE #2668: Ansible 2.2 bug in AWS cloud provider ini file
We are testing a new version of Ansible (ansible-2.2.0.0-0.70.rc3.el7.noarch).

In the new version of Ansible, the ini module will die if the ini file is NOT there.  According to the ini module docs, it looks like there is a new option called "create:"  However, I'm not sure if create is backwards compatible.

I see 2 options. 1) "touch" the file before it's created.  2) use "create: true"

Here is where it's bombing in the ansible code:

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_cloud_provider/tasks/aws.yml

Here is the ini module doc.  Please see the "create" statement.

http://docs.ansible.com/ansible/ini_file_module.html

sdodson:
Ansible 2.2 adds 'create' flag, however we'd like to preserve backwards
compatability.

Fixes bug 1390169
Fixes BZ1390169
Fixes #2668

### DIFF
--- a/roles/openshift_cloud_provider/tasks/aws.yml
+++ b/roles/openshift_cloud_provider/tasks/aws.yml
@@ -1,4 +1,14 @@
-- name: Create cloud config
+# Work around ini_file create option in 2.2 which defaults to no
+- name: Create cloud config file
+  file:
+    dest: "{{ openshift.common.config_base }}/cloudprovider/aws.conf"
+    state: touch
+    mode: 0660
+    owner: root
+    group: root
+    changed_when: false
+
+- name: Configure AWS cloud provider
   ini_file:
     dest: "{{ openshift.common.config_base }}/cloudprovider/aws.conf"
     section: Global
--- a/roles/openshift_loadbalancer/tasks/main.yml
+++ b/roles/openshift_loadbalancer/tasks/main.yml
@@ -10,6 +10,16 @@
     path: /etc/systemd/system/haproxy.service.d
     state: directory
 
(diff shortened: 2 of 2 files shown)


sdodson:
aos-ci-test

sdodson:
@tbielawa PTAL

roles/openshift_cloud_provider/tasks/aws.yml
```diff
@@ -1,4 +1,13 @@
-- name: Create cloud config
+# Work around ini_file create option in 2.2 which defaults to no
+- name: Create cloud config file
+  file:
+    dest: "{{ openshift.common.config_base }}/cloudprovider/aws.conf"
+    state: touch
```
tbielawa: I double checked the docs to be sure about this one

> If `touch` (new in 1.4), an empty file will be created if the path does not exist, while an existing file or directory will receive updated file access and modification times (similar to the way `touch` works from the command line).

So in effect this means our playbook will always say something `changed`. I know we have a lot of tasks that do that already, but IMHO that's misleading behavior and we should avoid it. What about adding

```yaml
- name: Create cloud config file
  ...
  changed_when: false
```

Since this is just a work-around for the missing `create` option in older versions?

sdodson: OK, sounds good.

tbielawa:
aos-ci-test

sdodson:
ci flake on something crazy

PR #2690, opened 2016-10-31
Only restart dnsmasq if the DNS servers have changed

tbielawa:
https://bugzilla.redhat.com/show_bug.cgi?id=1374170


tbielawa:
aos-ci-test

sdodson:
This will work but I was envisioning going the other way. Create the new temp file, compare it to the existing file, if there's no diff in md5sum don't do anything but remote the temp file. If the file is changed replace the existing file with the new file and restart the service. I think this has the added benefit of preserving timestamp information of the original file when nothing has changed. But you'd have to deal with what to do when the file doesn't exist already.

I'm fine with taking this as is, your call.

sdodson:
aos-ci-test

PR #2692, opened 2016-11-01
Fix iptables dependency problems.

dgoodwin:
In situations where iptables is already installed but is not the latest
version, and iptables-services is *not* installed, this task would fail
as ansible tries to install the newest services rpm, but it requires a
change to iptables version. (which is not handled gracefully) This
situation will very likely trigger on the release of 7.3, if user is still running 7.2.

Instead, switch to installing "latest" versions of each package, and
handle in two separate tasks so the iptables rpm is already latest
before we proceed to services.

Fixes #1389722

### DIFF
--- a/roles/os_firewall/tasks/firewall/iptables.yml
+++ b/roles/os_firewall/tasks/firewall/iptables.yml
@@ -24,17 +24,21 @@
   when: pkg_check.rc == 0
   ignore_errors: yes
 
-- name: Install iptables packages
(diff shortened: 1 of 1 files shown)


dgoodwin:
In bug Jason objected to the use of latest.

I discovered the problem is reproducible just with yum if you confirm the transaction, it for some reason cannot update iptables. Tempted to re-assign to RHEL but I need confirmation.

PR #2697, opened 2016-11-01
Consider lb/preconfigured hosts when verifying mixed environments

tbielawa:
* Preconfigured hosts (specifically load balancers) need to be checked
  so that the installed_hosts length assertion passes

* https://bugzilla.redhat.com/show_bug.cgi?id=1390064


tbielawa:
aos-ci-test

tbielawa:
aos-ci-test

PR #2702, opened 2016-11-02
Prevent useless master restart by reworking template for master service enf file

jkhelil:
We are running openshift-ansible every x minutes(among the reasons detecting and running openshift ansible for new nodes joining the cluster ex: preemptible vms), and for every run, master is restarting because the master service env file is changing. this is because of the lineinfile statement added after creating the template of the file,
I think in general manner, it would be good to ensure that all changes are in the template, and hence the master will restart only if needed.

### DIFF
--- a/roles/openshift_master/tasks/systemd_units.yml
+++ b/roles/openshift_master/tasks/systemd_units.yml
@@ -127,15 +127,21 @@
 
 - name: Preserve Master Proxy Config options
   command: grep PROXY /etc/sysconfig/{{ openshift.common.service_type }}-master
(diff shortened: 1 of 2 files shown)


jkhelil:
@detiber @lhuard1A PTAL

sdodson:
Thanks, I like your proposed fix. I'd like to run through it once to test it. If anyone else gets to that first feel free to approve and merge.

abutcher:
ok to test

abutcher:
aos-ci-test

PR #2707, opened 2016-11-02
Where we use curl force it to use tlsv1.2

sdodson:
curl, prior to RHEL 7.2, did not properly negotiate up the TLS protocol, so
force it to use tlsv1.2 as origin 1.4 has bumped the TLS requirement to v1.2.

See https://bugzilla.redhat.com/show_bug.cgi?id=1170339 and https://github.com/openshift/origin/issues/11702
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1390869


sdodson:
Alternatively, we could just upgrade curl but that'd quickly become another 'Check that PyYAML is installed'.

PR #2711, opened 2016-11-03
Add hawkular admin cluster role to management admin

simon3z:
cc @mwringe @detiber

### DIFF
--- a/roles/openshift_manageiq/vars/main.yml
+++ b/roles/openshift_manageiq/vars/main.yml
@@ -9,6 +9,20 @@ manageiq_cluster_role:
       verbs:
       - '*'
 
+manageiq_metrics_admin_clusterrole:
+  apiVersion: v1
+  kind: ClusterRole
+  metadata:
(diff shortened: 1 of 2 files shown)


roles/openshift_manageiq/vars/main.yml
```diff
@@ -9,6 +9,20 @@ manageiq_cluster_role:
       verbs:
       - '*'
 
+manageiq_metrics_admin_clusterrole:
```
simon3z: @mwringe @detiber not sure if this belongs here... also how do you remember to keep this in sync these with `origin-metrics`?

detiber:
@sdodson ptal

sdodson:
aos-ci-test

sdodson:
@simon3z which releases do you need this in? 3.4 and 3.3?

simon3z:
> @simon3z which releases do you need this in? 3.4 and 3.3?

@sdodson only 3.4 thanks!

PR #2713, opened 2016-11-03
Update quick installer upgrade mappings for 3.4

smunilla:
Just a quick addition for the upgrade playbooks for 3.4

Fixes Bug 1384291


sdodson:
aos-ci-test

PR #2715, opened 2016-11-03
ansible 2.2 template fixes for conditional/loops

detiber:


detiber:
@sdodson @abutcher this is still incomplete, There are still quite a few instances I haven't gotten a chance to fix, but this is as I was able to get prior to having to take off.

Rough query I'm using to find all the conditionals:
```
git grep -e "\swhen:" | grep -v "{{" | less
```

sdodson:
aos-ci-test

sdodson:
Test upgrades with this and ansible-2.2.0.0, worked for me.

PR #2721, opened 2016-11-04
Image stream sync

sdodson:
- Syncs latest content for both examples and hosted templates
 - Add sync script for hosted templates
 - Switches FIS image stream to 'GA' branch which doesn't have FIS 2.0 content as it's pre-release.

Fixes Bug 1391815
Fixes BZ1391815


sdodson:
aos-ci-test

PR #2728, opened 2016-11-04
default to multizone GCE config

sjenning:
Nodes created in zones other than the master are being deleted by the node controller because the master is not running in multizone mode when `openshift_cloudprovider_kind=gce`

This PR assumes we want to run in multizone for all configurations since there is no downside as far as I know of for doing so.

xref https://bugzilla.redhat.com/show_bug.cgi?id=1390160

@derekwaynecarr @detiber

### DIFF
--- a/roles/openshift_cloud_provider/tasks/gce.yml
+++ b/roles/openshift_cloud_provider/tasks/gce.yml
@@ -1,0 +1,16 @@
+# Work around ini_file create option in 2.2 which defaults to no
+- name: Create cloud config file
+  file:
+    dest: "{{ openshift.common.config_base }}/cloudprovider/gce.conf"
+    state: touch
+    mode: 0660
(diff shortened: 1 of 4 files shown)


sdodson:
aos-ci-test

sdodson:
I can fix those errors for you later tonight.

sjenning:
@sdodson I thought I had added that before :-/  I updated `roles/openshift_cloud_provider/vars/main.yml` to define `cloudprovider_is_gce` if you want to kick off the tests again.

sdodson:
aos-ci-test

sjenning:
@sdodson `aos-ci-jenkins/OS_3.4_NOT_containerized` flaked on `Install cockpit-ws` because the node wasn't registered with subscription manager.

PR #2730, opened 2016-11-04
Fix HA etcd upgrade when facts cache has been deleted.

dgoodwin:
Upgrade will fail if you do not have a
/etc/ansible/facts.d/openshift.fact cache on the system, usually from
running original config.yml for installation.

To fix we check if running embedded etcd more logically, however the
fact for etcd_data_dir still needs to be populated, and this is not
possible without running the openshift_master_facts role, which involves
a huge number of variables and thus we need to include this common
config.yml to get them all populated.

@abutcher or @detiber could really use your feedback on whether or not this include is safe. It works, but it brings in quite a bit of yaml. Here is the link to what's included: https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-master/config.yml

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -13,10 +13,23 @@
       groups: etcd_hosts_to_backup
     with_items: "{{ groups.oo_etcd_to_config if groups.oo_etcd_to_config is defined and groups.oo_etcd_to_config | length > 0 else groups.oo_first_master }}"
 
+# If facts cache were for some reason deleted, this fact may not be set, and if not set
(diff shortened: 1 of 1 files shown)


dgoodwin:
@abutcher noticed this will rewrite their master config, which is a non-starter. Re-evaluating and trying another approach now.

dgoodwin:
Fixed thanks to help from @abutcher, the embedded_etcd master fact was defaulting to True, we now set it explicitly during the etcd backup procedure. 

Problem is really only affecting co-located masters+etcds (rpm and containerized both). Embedded etcd, or separate etcd hosts, are not affected as far as I can tell. I have tested each.

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1391608

Issue #2731, opened 2016-11-05
[OpenStack method] ImportError: No module named shade 


dougbtv:
I get a `ImportError: No module named shade` and will propose a documentation fix to install the python module.

##### Version
openshift-ansible-3.4.17-1

##### Steps To Reproduce
1. Run `bin/cluster ... openshift clustername` for openstack on CentOS (additionally see example run command in the "Current Result" section)

##### Current Result

I'm told I don't have the python module shade.

Short version

```
ImportError: No module named shade
```

Long version / example 
```
[stack@undercloud openshift-ansible]$ bin/cluster create \
>   -o image_name=centos-atomic \
>   -o external_net=ext-net \
>   -o floating_ip_pool=ext-net \
>   -o net_cidr=40.0.0.0/24 \
>   -o public_key=~/atomic_key.pub \
>   openstack test_cluster
================================================================================
ATTENTION: You are running a community supported utility that has not been
tested by Red Hat. Visit https://docs.openshift.com for supported installation
instructions.
================================================================================

ERROR! The file inventory/openstack/hosts/openstack.py is marked as executable, but failed to execute correctly. If this is not supposed to be an executable script, correct this with `chmod -x inventory/openstack/hosts/openstack.py`.
Inventory script (inventory/openstack/hosts/openstack.py) had an execution error: Traceback (most recent call last):
  File "/home/stack/openshift-ansible/inventory/openstack/hosts/openstack.py", line 61, in <module>
    import shade
ImportError: No module named shade
 
inventory/openstack/hosts/openstack.py:48: Expected key=value host variable assignment, got: argparse

```

##### Expected Result

Continue onto running playbook... 

##### Proposed Fix

I think this is just a documentation error and I'll submit a PR. We just need to 

```
sudo pip install shade
```

##### Additional Information
Exact command above.
OS: CentOS 7.2
Inventory: localhost

PR #2732, opened 2016-11-05
[openstack] updates documentation to show that you need to install shade

dougbtv:
Documentation fix to show that you need the shade module install in CentOS in order to continue when using the OpenStack method.

Error detailed in issue #2731


sdodson:
@dougbtv thanks, BTW, if you put "Fixes #number" in the description of the PR then that issue will auto close when the PR merges.

PR #2745, opened 2016-11-07
[upgrades] Use etcd3 to backup embedded etcd if possible

sdodson:
If RHEL, detect if etcd is installed and is < 3.0. If so then swap etcd3 for
etcd.

If Fedora, install latest etcd and attempt to backup. If it fails advise the
user they may need to download etcd from github to attempt a backup.

Fixes Bug 1382634
Fixes BZ1382634

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -56,9 +56,9 @@
         {{ avail_disk.stdout }} Kb available.
     when: (embedded_etcd | bool) and (etcd_disk_usage.stdout|int > avail_disk.stdout|int)
 
-  - name: Install etcd (for etcdctl)
-    action: "{{ ansible_pkg_mgr }} name=etcd state=latest"
-    when: not openshift.common.is_atomic | bool
+  - name: Install etcdctl for embedded environments
+    include: embedded_etcdctl.yml
+    when: embedded_etcd
 
   - name: Generate etcd backup
     command: >
--- a/playbooks/common/openshift-cluster/upgrades/embedded_etcdctl.yml
+++ b/playbooks/common/openshift-cluster/upgrades/embedded_etcdctl.yml
@@ -1,0 +1,30 @@
+- name: Install etcd for etcdctl on non RHEL
+  action: "{{ ansible_pkg_mgr }} name=etcd state=latest"
+  when: not openshift.common.is_atomic | bool and ansible_distribution != 'RedHat'
+
+# etcd3 in RHEL 7.3.0 does not Obsolstes: etcd, however etcd3-3.0.12 or later will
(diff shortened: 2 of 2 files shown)


sdodson:
@dgoodwin PTAL I believe this fixes the 3.3 async upgrade failures where there's an errant file in the wal or snap directories. I suspect these problems would also happen on a 3.3 to 3.4 upgrade but we'll ship etcd3 versions that obsolete etcd prior to 3.4 so I'm only making this fix on release-1.3.

I have *not* been able to reproduce the problem on my own but I have successfully tested that using etcd3 to back up a 3.3 embedded etcd environment worked where as using etcd-2.3.7 failed.

sdodson:
Yeah, though I just realized that this would also upgrade non embedded etcd instances to 3.0 and it'll do so without the careful upgrade stepping we're working on in the 3.4 playbooks. I feel like we're stuck between a rock and a hard place.

sdodson:
@dgoodwin mind checking that last commit to only switch to etcd3 when embedded?

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -56,17 +56,52 @@
...
+  - name: Swap etcd3 for etcd
+    command: "yum swap etcd etcd3 -y"
+    args:
+      warn: no
+    when: not openshift.common.is_atomic | bool and ansible_distribution == 'RedHat' and etcd2_installed | bool and embedded_etcd
+
+  - name: Installed latest etcd
     action: "{{ ansible_pkg_mgr }} name=etcd state=latest"
-    when: not openshift.common.is_atomic | bool
+    when: not openshift.common.is_atomic | bool and ansible_distribution != 'RedHat'
```
sdodson: Probably should get rid of ansible_distribution and add `not embedded_etcd` to preserve the current behavior even though that's currently upgrading etcd all the time.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -56,17 +56,48 @@
...
+  # etcd3 in RHEL 7.3.0 does not Obsolstes: etcd, however etcd3-3.0.12 or later will
+  # so until that ships we need to carefully install etcd 3.0 to ensure we get fixes
+  # for backing up embedded etcd in 2.3
+  # see https://github.com/coreos/etcd/pull/4952 and https://bugzilla.redhat.com/show_bug.cgi?id=1382634
+  - name: Check for etcd package
+    rpm_q:
+      name: etcd
+      state: present
+    failed_when: false
+    register: etcd_rpmq
```
dgoodwin: This should probably be conditionalized as should the etcd2_installed set_fact below. For safety and readability I almost think we should push all of these embedded specific statements out to something like "embedded_etcd_backup_hacks.yml" and slap a conditional on it. This would help readibility IMO and might actually save us from some bugs in the future.

PR #2749, opened 2016-11-07
set AWS creds task with no_logs

sallyom:
@detiber @sdodson I notice the creds are printed here, is it best to use no_log?


abutcher:
aos-ci-test

PR #2750, opened 2016-11-07
set defaults for debug_level in template and task

sdodson:


sdodson:
@tbielawa fix this on release-1.2 too

PR #2752, opened 2016-11-08
Refactor default sdn_cluster_network_cidr and sdn_host_subnet_length

sdodson:
Inventory values highest priority, then anything in master-config.yaml, then
defaults last.

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -903,8 +903,8 @@ def set_sdn_facts_if_unset(facts, system_facts):
     if 'master' in facts:
         # set defaults for sdn_cluster_network_cidr and sdn_host_subnet_length
         # these might be overridden if they exist in the master config file
-        facts['master']['sdn_cluster_network_cidr'] = '10.128.0.0/14'
-        facts['master']['sdn_host_subnet_length'] = '9'
+        sdn_cluster_network_cidr = '10.128.0.0/14'
+        sdn_host_subnet_length = '9'
 
         master_cfg_path = os.path.join(facts['common']['config_base'],
                                        'master/master-config.yaml')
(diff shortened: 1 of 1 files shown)


sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
@smunilla PTAL

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -914,12 +914,17 @@ def set_sdn_facts_if_unset(facts, system_facts):
...
                 if 'clusterNetworkCIDR' in config['networkConfig']:
-                    facts['master']['sdn_cluster_network_cidr'] = \
+                    sdn_cluster_network_cidr = \
                         config['networkConfig']['clusterNetworkCIDR']
                 if 'hostSubnetLength' in config['networkConfig']:
-                    facts['master']['sdn_host_subnet_length'] = \
+                    sdn_host_subnet_length = \
                         config['networkConfig']['hostSubnetLength']
 
+        if 'sdn_cluster_network_cidr' not in facts['master']:
```
abutcher: This will allow vars to be overridden but the config file values will never overwrite user provided because they will already be set.

I think that the config section should set directly to facts (overwriting) and the defaults should be set when no value is already set.

```diff
         # set defaults for sdn_cluster_network_cidr and sdn_host_subnet_length
         # these might be overridden if they exist in the master config file
-        facts['master']['sdn_cluster_network_cidr'] = '10.128.0.0/14'
-        facts['master']['sdn_host_subnet_length'] = '9'
+        if 'sdn_cluster_network_cidr' not in facts['master']:
+            facts['master']['sdn_cluster_network_cidr'] = '10.128.0.0/14'
+        if 'sdn_host_subnet_length' not in facts['master']:
+            facts['master']['sdn_host_subnet_length'] = '9'
```

sdodson: Or are you suggesting that master-config.yaml should override inventory variables?

abutcher: Any values set by the user will be set before we call `set_sdn_facts_if_unset` and since we have conditionals on the defaults and config values, the config values won't override user values.

abutcher: Right

sdodson: Oh, I was thinking we always wanted inventory vars to win.

abutcher: Kk, I'm on board with letting inventory take precedence.

abutcher:
aos-ci-test

sdodson:
Cockpit shenanigans.

PR #2753, opened 2016-11-08
Always add local dns domain to no_proxy

jkhelil:
Even if we set generate_no_proxy_hosts to false, we want to have .cluster.local in the no_proxy by default


sdodson:
aos-ci-test

PR #2758, opened 2016-11-08
Re-allow setting clusterNetworkCIDR & hostSubnetLength

smunilla:
A previous change prevented setting clusterNetworkCIDR &
hostSubnetLength though inventory variables. This addresses
that.


smunilla:
aos-ci-test

Issue #2698, opened 2016-11-01
Need generic way to deploy with a specific CNI plugin


yfauser:
I am working on integrating our network virtualization product (NSX) with OpenShift using CNI.
I'm deploying OpenShift in my Lab using the BYO playbooks.

When deploying OpenShift with BYO, I don't want to install OpenShift SDN, I only want to the network plug-in name in the master and node config to be 'cni':

```
os_sdn_network_plugin_name=cni
openshift_use_openshift_sdn=false
```

I found that this is not working and added a quick fix to get around this:
https://github.com/yfauser/openshift-ansible/commit/ee2f9daacea2f84d9f7917bcbae38648bd359191

While this is a quick fix, I would rather want to suggest a generic way to handle cases like ours.
Instead of having a separate flag per network plugin vendor, like used in the Nuage and Flannel case, it would be great to be able to specify the cni config file in the [OSEv3:vars] e.g. like:

```
openshift_use_openshift_sdn=false

os_sdn_network_plugin_name=cni
os_sdn_network_cni_netconf=<file or template name>
os_sdn_network_cni_netbinary=<file or template name>
```

or

```
openshift_use_openshift_sdn=false

os_sdn_network_plugin_name=cni
os_sdn_network_cni_package=<rpm to install>
```

What do you guys think?

Also, let me know if you want to receive a pull request for the above 'quick fix'

Thanks!

##### Version
▶ git describe
openshift-ansible-3.4.14-1-5-gee2f9da

▶ ansible --version
ansible 2.1.1.0

Centos7.2 used as base OS

detiber:
> Instead of having a separate flag per network plugin vendor, like used in the Nuage and Flannel case, it would be great to be able to specify the cni config file in the [OSEv3:vars] e.g. like:

```
openshift_use_openshift_sdn=false

os_sdn_network_plugin_name=cni
os_sdn_network_cni_netconf=<file or template name>
os_sdn_network_cni_netbinary=<file or template name>
```
> or

```
openshift_use_openshift_sdn=false

os_sdn_network_plugin_name=cni
os_sdn_network_cni_package=<rpm to install>
```

> What do you guys think?


I love the idea. Currently the flannel and nuage configurations we support do not leverage CNI, but I think the addition of generic CNI configuration would be a great addition.

[...]

PR #2763, opened 2016-11-08
[#2698] Change to allow cni deployments without openshift SDN

yfauser:
The roles/openshift_facts main task did not pass the cni plugin variable to the later role playbooks.
The master.yaml and node.yaml templates did not allow for a cni configuration without either installing openshift sdn or nuage.

This change will allow to use os_sdn_network_plugin_name=cni and set openshift_use_openshift_sdn=false for deployments that use a cni plugin that doesn't need and want openshift sdn to be installed

See Issue #2698


abutcher:
ok to test

abutcher:
aos-ci-test

PR #2767, opened 2016-11-09
Update default log driver to journald

smunilla:
Change the default log driver from josn-file to journald to
improve logging functionality.

Fixes Bug 1393178


smunilla:
@sdodson Is there any reason not to do this I'm not thinking of?

sdodson:
I can't think of anything.
@richm Are there any other tasks that need to be done when changing this on an existing cluster?

richm:
If you change this on an existing cluster, you will need to restart fluentd so that it knows to use journald as the input source.  That is, on the node, when you restart docker to use journald, you should restart fluentd on that node too.
When upgrading to 3.4, will all of the logging components be restarted/redeployed?

smunilla:
Looks like @abutcher had this covered in #2767

PR #2770, opened 2016-11-09
Resolve docker and iptables service dependencies

mtnbikenc:
The docker service adds rules to the iptables configuration to support proper
network functionality for running containers. If the service is started prior
to iptables, these rules are not properly created.

* Ensure iptables is started prior to docker

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1390835

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -46,11 +46,23 @@
   action: "{{ ansible_pkg_mgr }} name=docker{{ '-' + docker_version if docker_version is defined else '' }} state=present"
   when: not openshift.common.is_atomic | bool
 
+- name: Ensure docker.service.d directory exists
+  file:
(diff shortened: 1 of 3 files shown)


roles/os_firewall/tasks/firewall/iptables.yml
```diff
@@ -54,6 +54,13 @@
   - ip6tables
   when: "'masked' in os_firewall_iptables_masked_output.results | map(attribute='stdout')"
 
+# Configure iptables to save config on STOP and RESTART
+- name: Update default iptables configuration
+  replace:
+    dest: /etc/sysconfig/iptables-config
+    regexp: '^(IPTABLES_SAVE_ON_.*)no(.*)$'
+    replace: '\1yes\2'
+
```
abutcher: @dcbw

mtnbikenc:
aos-ci-test

PR #2771, opened 2016-11-09
Added a BYO playbook for configuring NetworkManager on nodes

stevekuznetsov:
In order to do a full install of OpenShfit using the byo/config.yml
playbook, it is currently required that NetworkManager be installed
and configured on the nodes prior to the installation. This playbook
introduces a very simple default configuration that can be used to
install, configure and enable NetworkManager on their nodes.

@sdodson PTAL

### DIFF
--- a/playbooks/byo/openshift-node/network_manager.yml
+++ b/playbooks/byo/openshift-node/network_manager.yml
@@ -1,0 +1,36 @@
+---
+- hosts: localhost
+  connection: local
+  become: no
+  gather_facts: no
+  tasks:
+  - include_vars: ../../byo/openshift-cluster/cluster_hosts.yml
+  - add_host:
+      name: "{{ item }}"
+      groups: l_oo_all_hosts
+    with_items: "{{ g_all_hosts }}"
+
+- hosts: l_oo_all_hosts
+  become: yes
+  tasks:
+    - name: install NetworkManager
+      package:
+        name: 'NetworkManager'
+        state: present
+
+    - name: configure NetworkManager
+      lineinfile:
+        dest: "/etc/sysconfig/network-scripts/ifcfg-{{ ansible_default_ipv4['interface'] }}"
+        regexp: '^{{ item }}='
+        line: '{{ item }}=yes'
+        state: present
+        create: yes
+      with_items:
+        - 'USE_PEERDNS'
+        - 'NM_CONTROLLED'
+
+    - name: enable and start NetworkManager
+      service:
+        name: 'NetworkManager'
+        state: started
+        enabled: yes


playbooks/byo/openshift-node/network_manager.yml
```diff
@@ -0,0 +1,46 @@
...
+        name: 'NetworkManager'
+        state: present
+
+    - name: determine the default route
+      shell: "/sbin/ip route list match 0.0.0.0/0 | awk '{print $3 }'"
+      register: default_route
+
+    - name: determine the default interface
+      shell: "/sbin/ip route get to {{ default_route.stdout }} | awk '{print $3}'"
+      register: default_interface
```
mscherer: Isn't it simpler to use ansible_default_ipv4['interface'] ?

stevekuznetsov: I think I copied this step from where-ever else this project was looking up the default interface. If that variable is what we want (why not IPv6?) then I will switch it over, sure.

mscherer: We could get ipv6 as well, but as "/sbin/ip route list match 0.0.0.0/0" will return ip v4, I assumed that's what was required.

stevekuznetsov: I see, understood. @sdodson does using the `ansible_default_ipv4['interface']` fact seem reasonable here for you as well?

playbooks/byo/openshift-node/network_manager.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    - name: configure NetworkManager
+      lineinfile:
+        dest: '/etc/sysconfig/network-scripts/ifcfg-{{ default_interface.stdout }}'
+        regexp: '{{ item.regexp }}'
+        line: '{{ item.line }}'
+        state: present
+        create: yes
+      with_items:
+        - regexp: '^USE_PEERDNS='
```
mscherer: What about: 
```
     lineinfile:
        dest: '/etc/sysconfig/network-scripts/ifcfg-{{ default_interface.stdout }}'
        regexp: '^{{ item }}='
        line: '^{{ item }}=yes'
        state: present
        create: yes
      with_items:
      - 'USE_PEERDNS'
      - 'NM_CONTROLLED'
 ```

So one can avoid repetition, and directly abstract the configuration we want to be at 'yes' ?

stevekuznetsov: Sure, that collapses nicely

stevekuznetsov:
@mscherer addressed your comments

playbooks/byo/openshift-node/network_manager.yml
detiber: @stevekuznetsov it does.

PR #2562, opened 2016-10-06
etcd upgrade playbooks

sdodson:
Done:
 - Stepwise RPM upgrades from 2.0 to 2.1 to 2.2 to 2.3 to 3.0 on RHEL/Centos
 - Stepwise containerized upgrades from 2.1 to 2.2 to 2.3 to 3.0 (2.0 was never available as an image)
 - Blind upgrade to latest on fedora
 - Backups before and after
 - Drops bash functions 'etcdctl2' and 'etcdctl3' which have reasonable flags for invoking etcdctl against v2 or v3 apis respectively

Tangential note - etcd3 package was added to RHEL 7.3.0. In the yet to be released RHEL 7.3.1 etcd3-3.0.12 or newer will have `Obsoletes: etcd` meaning that if you install etcd after that version ships you'll actually get etcd3. As such I've not added conditional logic to install 'etcd3' and I'm assuming that the Obsoletes will provide that for clean installs of OCP3.4.

PR #2773, opened 2016-11-09
Don't upgrade etcd on backup operations

sdodson:
We need to install etcd for etcdctl but if it's already installed we shouldn't upgrade it.

Fixes Bug 1393187
Fixes BZ1393187

We'll be providing a comprehensive etcd upgrade playbook in #2562 ASAP


sdodson:
@dgoodwin PTAL

dgoodwin:
A thousand times yes. That was a big oversight.

PR #2779, opened 2016-11-10
Update override cluster_hostname

smunilla:
Update the default for the override prompt and set both
openshift_master_cluster_hostname and
openshift_master_cluster_public_hostname


tbielawa:
aos-ci-test

PR #2781, opened 2016-11-10
Add Travis integration

rhcarvalho:
We can run simple tests in every PR, while triggering our internal CI
tests on manually selected PRs.

This is a bit better than the all-or-nothing we have today.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+sudo: false
+
+language: python
+python:
+  - "2.7"
+
+install:
+  - pip install -r requirements.txt
+
+script:
+  # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
+  - ansible-playbook --syntax-check playbooks/byo/config.yml
+  # TODO(rhcarvalho): update make ci to pick up these tests
+  - nosetests --tests=test
+  - cd utils && make ci
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![Join the chat at https://gitter.im/openshift/openshift-ansible](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/openshift/openshift-ansible)
+[![Build Status](https://travis-ci.org/openshift/openshift-ansible.svg?branch=master)](https://travis-ci.org/openshift/openshift-ansible)
 
 # OpenShift Ansible
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+ansible>=2.1
+pyOpenSSL


sdodson:
@rhcarvalho What's the advantage to running these in travis versus in the 'default' job and just allowing the default job to run unabated whenever a PR is opened? currently the user has to be whitelisted. Is travis seen as safer to do that in because it's not our infra?

rhcarvalho:
@sdodson 

> What's the advantage to running these in travis versus in the 'default' job and just allowing the default job to run unabated whenever a PR is opened?

One advantage of Travis is that the CI configuration lives in this repository, and it is transparent to contributors what tests are being run.

The internal CI is a black box. One needs to have access to https://ci.openshift.redhat.com/jenkins/job/openshift-ansible/ to be able to see what is configured there and what tests are executed. (Unless I'm wrong? I believe build results are public, build config is private)

> currently the user has to be whitelisted.

Aha, so it explains why it was already automatically triggered in this PR.

I've seen some PRs that got merged and apparently no tests were run.
Not that our tests do miracles today, but there is no harm in running the simple and quick ones always.

I didn't see any way to trigger just the `default` tests today, is it possible?
I've only learned about one command that triggers all tests (which I will not mention to don't trigger the tests ;) ), and in many cases those tests take time and infra and are unrelated to the changes (so we often choose not to run them).

> Is travis seen as safer to do that in because it's not our infra?

I wouldn't use that as an argument, but it might be the case. Perhaps transparency is the strongest advantage for me.

sdodson:
Ok, I think I'd prefer merging what you're suggesting with `make ci` and having travis run that, then do away with the `default` test job. Do you think that's possible?

rhcarvalho:
@sdodson that's possible.

I'm trying out `make ci` on Travis now. Seems that that target it is opinionated about creating a virtualenv, don't know how that will behave in Travis.

Also worth nothing that `make ci` doesn't run the tests under `test/`, only those from `util/test/`.

rhcarvalho:
> then do away with the default test job

@sdodson I'd be okay either way. Travis and default run in different environments. The Jenkins job can be useful for times when Travis is slow like right now...

sdodson:
@tbielawa thoughts on the matter?

tbielawa:
On my other project I use travis and the same 'make ci' system I integrated into openshift-ansible. My travis file for that project calls 'make ci' just like ours does, and that system uses an virtualenv, just like ours does.

https://github.com/tbielawa/bitmath/blob/master/.travis.yml

The config options there are what I use with great success!

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+  - "2.7"
+
+install:
+  - pip install -r requirements.txt
+
+script:
+  # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
+  - ansible-playbook --syntax-check playbooks/byo/config.yml
+  # TODO(rhcarvalho): update make ci to pick up these tests
+  - nosetests test/
```
rhcarvalho: @sdodson FYI I've made a change here to run only the tests in `test/`, while `make ci` picks up other tests (same as what runs in the 'default' Jenkins job).

I plan to improve this later when I'm back from PTO, for now it is already better than before :-)

rhcarvalho: Ah, and the reason for that is that just `nosetests` + `make ci` would run 37 tests twice. The tests are quick, but installing their dependencies is not. So no need to wait for dependency installation twice.

rhcarvalho:
Added the badge, thanks @sdodson.

PR #2782, opened 2016-11-10
[WIP] Merge kube_admission_plugin_config with admission_plugin_config

smunilla:
Move the values in kube_admission_plugin_config up one level per
the new format from 1.3:

"The kubernetesMasterConfig.admissionConfig.pluginConfig should be moved
and merged into admissionConfig.pluginConfig."


playbooks/common/openshift-cluster/upgrades/v3_3/master_config_upgrade.yml
```diff
@@ -48,3 +48,7 @@
     dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
     yaml_key: 'controllerConfig.servicesServingCert.signer.keyFile'
     yaml_value: service-signer.key
+- modify_yaml:
+    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+    yaml_key: 'admissionConfig.pluginConfig'
+    yaml_value: "{{ openshift.master.kube_admission_plugin_config | to_padded_yaml(level=2) }}"
```
dgoodwin: Is this for 3.4? If so you'll need to add a "hook" file for 3.4 similar to this one, and link it similar to how this one is for the upgrade process. (just grep for master_config_upgrade)

smunilla: The docs seem to indicate we should have been doing this starting with 1.3 (https://docs.openshift.org/latest/architecture/additional_concepts/admission_controllers.html#admission-controllers-general-admission-rules) so I thought it would be good to do on minor upgrades too. Though I guess that does increase the chances of something breaking…

dgoodwin: In that case you can get it out of this file (which is only run on 3.3 upgrades) and somewhere up a level in common. Keep an eye out for configs that shouldn't have it at all though in your testing.

smunilla: @dgoodwin Is there somewhere common that already exists or would the easy thing be to just copy-paste this somewhere else as well?

dgoodwin: @smunilla yeah just look for where the master config hooks are run.

abutcher: `modify_yaml` should take care of the `to_padded_yaml` bits so we can remove that piece.

```
- modify_yaml:
    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
    yaml_key: 'admissionConfig.pluginConfig'
    yaml_value: "{{ openshift.master.kube_admission_plugin_config }}"
  when: "'kube_admission_plugin_config' in openshift.master"
```

But we're also completely overwriting the original `admissionConfig.pluginConfig` which may have had different config values.

Testing with these inventory vars you'll see the `ProjectRequestLimit` disappear after the upgrade.

```
openshift_master_admission_plugin_config={"ProjectRequestLimit":{"configuration":{"apiVersion":"v1","kind":"ProjectRequestLimitConfig","limits":[{"selector":{"admin":"true"}},{"maxProjects":"1"}]}},"PodNodeConstraints":{"configuration":{"apiVersion":"v1","kind":"PodNodeConstraintsConfig"}}}
openshift_master_kube_admission_plugin_config={"RunOnceDuration":{"configuration":{"apiVersion":"v1","kind":"RunOnceDurationConfig","activeDeadlineSecondsOverride":"3600"}},"ClusterResourceConfig":{"configuration":{"apiVersion":"v1","kind":"ClusterResourceOverrideConfig","limitCPUToMemoryPercent":"200","cpuRequestToLimitPercent":"6","memoryRequestToLimitPercent":"60"}},"PodNodeConstraints":{"configuration":{"apiVersion":"v1","kind":"PodNodeConstraintsConfig"}}}
```

Running byo/config after upgrade will restore the original `admissionConfig.pluginConfig` but removes the `kubernetesMasterConfig.admissionConfig` that moved over during upgrade.

PR #2786, opened 2016-11-10
Begin requiring Docker 1.12.

dgoodwin:
Building off the work done for Docker 1.10, we now require Docker 1.12
by default.

The upgrade process was already set to ensure you are running the latest
docker during upgrade, and the standalone docker upgrade playbook can
also be used if desired.

As before, you can override this Docker 1.12 requirement by setting a
docker_version=1.10.3 (or similar), or you can skip upgrading docker entirely with docker_upgrade=false.

### DIFF
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -84,7 +84,7 @@ openshift_release=v1.4
 
 # Specify exact version of Docker to configure or upgrade to.
 # Downgrades are not supported and will error out. Be careful when upgrading docker from < 1.10 to > 1.10.
-# docker_version="1.10.3"
+# docker_version="1.12.1"
 
 # Skip upgrading Docker during an OpenShift upgrade, leaves the current Docker version alone.
 # docker_upgrade=False
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -84,7 +84,7 @@ openshift_release=v3.4
 
 # Specify exact version of Docker to configure or upgrade to.
 # Downgrades are not supported and will error out. Be careful when upgrading docker from < 1.10 to > 1.10.
-# docker_version="1.10.3"
+# docker_version="1.12.1"
 
 # Skip upgrading Docker during an OpenShift upgrade, leaves the current Docker version alone.
 # docker_upgrade=False
--- a/playbooks/byo/openshift-cluster/upgrades/docker/upgrade.yml
+++ b/playbooks/byo/openshift-cluster/upgrades/docker/upgrade.yml
@@ -1,6 +1,4 @@
 # Playbook to upgrade Docker to the max allowable version for an OpenShift cluster.
-#
-# Currently only supports upgrading 1.9.x to >= 1.10.x.
 - hosts: localhost
   connection: local
   become: no
--- a/playbooks/common/openshift-cluster/upgrades/docker/upgrade_check.yml
+++ b/playbooks/common/openshift-cluster/upgrades/docker/upgrade_check.yml
@@ -28,9 +28,9 @@
   changed_when: false
 
 - fail:
-    msg: This playbook requires access to Docker 1.10 or later
-  # Disable the 1.10 requirement if the user set a specific Docker version
-  when: docker_version is not defined and (docker_upgrade is not defined or docker_upgrade | bool == True) and (pkg_check.rc == 0 and (avail_docker_version.stdout == "" or avail_docker_version.stdout | version_compare('1.10','<')))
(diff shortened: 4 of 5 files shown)


dgoodwin:
In my testing this seems to be working great, thankfully the situation is very consistent with 1.10. I've tested with no Docker 1.12 available, the standalone docker upgrade, and docker upgrade within regular cluster upgrade.

Note that during upgrade, docker is handled within the "node" upgrade, if you just upgrade control plane this is not where the upgrade will happen.

sdodson:
Sigh, the only problem is logistics. Centos/RHEL users won't have access to docker-1.12 immediately. I'll pick this into an ocp-3.4 branch and leave this PR until docker-1.12 is available in centos. That way QE can test tonight and we can discuss what to do tomorrow.

dgoodwin:
For what it's worth it can be overridden with the vars mentioned in pr,
though there will be questions.

alikhajeh1:
Any update on this? Is it possible to use openshift-ansible to update docker on an existing openshift 1.4 cluster to pickup fixes for things like https://bugzilla.redhat.com/show_bug.cgi?id=1409531?
If not, I was thinking of doing it manually as follows (based on OC docs):
1. Disable node for scheduling and evacuate pods
2. `yum update docker` then restart docker and origin-node (on masters also restart api/controllers)
3. Enable node for scheduling

sdodson:
@alikhajeh1 that should work. RHEL7 just shipped docker-1.12, you may have to clear your yum repo cache. I'd expect Centos to ship that update shortly.

sdodson:
aos-ci-test

camilamacedo86:
This PR is affecting the installation of the open shift. This version of docker is not compatible and we should come back this changes. 

Error:

Issue: Message error "SetupNetwork" is found and the default router is not created in the installation of versions 3.2 and 3.4 ( probably 3.3 too )
Issue found when openshift-ansible[1] is used to install and the machine/server don't have a docker installed. This script is suggested to be used in OpenShift documents[2].
 $ oc get pods

NAME                       READY     STATUS              RESTARTS   AGE
docker-registry-1-deploy   0/1       DeadlineExceeded    0          17m
docker-registry-2-deploy   0/1       ContainerCreating   0          16m
router-1-deploy            0/1       ContainerCreating   0   
$ oc describe pod/router-1-deploy
WarningFailedSync	Error syncing pod, skipping: failed to "SetupNetwork" for "router-1-deploy_default" with SetupNetworkError: "Failed to setup network for pod \"router-1-deploy_default(b43bdd83-0a59-11e7-bfa4-02acbbb94784)\" using network plugins \"redhat/openshift-ovs-subnet\": Error running network setup script: + brctl delif lbr0\nIncorrect number of arguments for command\nUsage: brctl delif <bridge> <device>\tdelete interface from bridge\n; Skipping pod"
Root Cause:
The openshift-ansible[1] is installing a version not compatible with OpenShift in this situation. Version 1.12.6 was installed and it should be 1.10.3.
Solution suggested:
Fix the openshift-ansible[1] to install the compatibles versions or the fixed version 1.10.3 which we checked that worked with 3.2 and 3.4.

eparis:
I don't see how the docker version would be relevant here.... Seems more like the new ansible version is incompatible with the sdn scripts from the old kubernetes you are using. Is that supposed to work? And I supposed to be able to get brand new ansible and install old versions of OpenShift?

 @dcbw or @danwinship know x100 more than me...

dgoodwin:
OpenShift Ansible master cannot currently be used to install older versions, but you can use the relevant release-3.x branch to do so.

Docker 1.12 is required for 3.4+ but per the PR description you can override if necessary.

camilamacedo86:
@dgoodwin We tried with 3.4. The same error is happening.

sdodson:
@camilamacedo86 We need to debug your scenario with the networking team. Docker 1.12 is a requirement for OCP 3.4 and Origin 1.4 and we have many many deployments using it and QE tested docker-1.12 extensively so there must be something unique to your environment.

Also, as Devan mentioned, you cannot use the master branch of openshift-ansible with OCP / Origin versions older than 3.4 / 1.4. The master branch always supports the latest development version and the current latest GA version only.

Lets open an origin issue regarding your particular error and make sure that it gets assigned to the networking team for diagnosis.

vanloswang:
@sdodson I meet 'Failed to "SetupNetwork"' error on OCP 3.5 latest with docker 1.12.6-16.

sdodson:
That's a networking issue we need the networking team to debug can you please open an issue in the origin repo and tag @openshift/networking

camilamacedo86:
@sdodson it is already open Bugzila 1436371

PR #2788, opened 2016-11-11
Use six rather than try/except import for configparser

mscherer:


mscherer:
aos-ci-test

mscherer:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
@mscherer thanks, LGTM but we're trying to wrap up a release and minimize churn. Without this fix are Fedora or other Python 3 environments broken?

mscherer:
Oh no, that's just a cleanup as I was reading the code, nothing urgent at all.

mscherer:
Seems to no longer be needed since be949e0a

PR #2789, opened 2016-11-11
NFS on Atomic Host

cgwalters:
The middle patch is obviously the biggest and may have upgrade implications...I can split it into a separate PR if you prefer.

### DIFF
--- a/roles/kube_nfs_volumes/tasks/nfs.yml
+++ b/roles/kube_nfs_volumes/tasks/nfs.yml
@@ -3,6 +3,13 @@
   action: "{{ ansible_pkg_mgr }} name=nfs-utils state=present"
   when: not openshift.common.is_containerized | bool
 
+- name: Work around https://bugzilla.redhat.com/show_bug.cgi?id=1394395
+  file: path=/var/lib/nfs/{{ item }} state=touch
+  with_items:
+    - etab
+    - rmtab
+  when: openshift.common.is_atomic | bool
+
 - name: Start rpcbind on Fedora/Red Hat
   service: name=rpcbind state=started enabled=yes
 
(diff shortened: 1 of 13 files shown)


roles/kube_nfs_volumes/tasks/nfs.yml
```diff
@@ -3,6 +3,13 @@
   action: "{{ ansible_pkg_mgr }} name=nfs-utils state=present"
   when: not openshift.common.is_containerized | bool
 
+- name: Work around https://bugzilla.redhat.com/show_bug.cgi?id=1394395
+  file: path=/var/lib/nfs/{{ item }}
+  with_items:
+    - etab
+    - rmtab
+  when: openshift.common.is_atomic | bool
+
```
sdodson: I remember discussing this bug with either you or someone else and they said that we'd never support running NFS server on atomic host. Has that changed?

cgwalters: Maybe...that sounds somewhat vaguely familiar.  But why wouldn't we?

cgwalters:
For background here, this is part of working to have the Vagrant environment work on AH: https://github.com/openshift/openshift-ansible-contrib/pull/104

roles/kube_nfs_volumes/tasks/nfs.yml
sdodson: https://bugzilla.redhat.com/show_bug.cgi?id=1293933

cgwalters: Thanks.  I am trying to make Atomic Host a fully general system and not just a kube node, for various reasons, so hosting NFS is definitely something I want to do.  The workaround is pretty small, and we'll fix it upstream ideally soon.

sdodson:
@abutcher Do you mind picking this work up and getting it merged? It'll need a rebase and some way to take into account legacy config. Perhaps if /exports/* exists create /srv/exports/* and move the contents there then complete the rest of the tasks?

abutcher:
@sdodson Sure thing

michaelgugino:
This patch is stale and never going to ship.

PR #2793, opened 2016-11-14
Added dependency of os_firewall to docker role

mtnbikenc:
The docker role requires iptables-services to be installed. Added
dependency on so_firewall role to ensure the iptables service is
installed first. Currently this will only work with iptables and
not with firewalld.

* Added allow_duplicates to os_firewall role meta
* Removed unused task from docker/tasks
* Corrected os_firewall Defaults in README


sdodson:
aos-ci-test

PR #2794, opened 2016-11-14
Fix HA upgrade when fact cache deleted.

dgoodwin:
This variable is referenced in the systemd unit templates, this seems
like the easiest and most consistent fix.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -25,6 +25,7 @@
       role: master
       local_facts:
         embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -25,6 +25,7 @@
       role: master
       local_facts:
         embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
+        debug_level: "{{ openshift_master_debug_level | default(openshift.common.debug_level) }}"
```
sdodson: Are we sure that common facts have been run here? If not add | default(2) too?

dgoodwin: Updated but please double check my syntax, I didn't run a full HA upgrade after the change.

PR #2804, opened 2016-11-15
Bug 1393716 - OpenShift containerized install should mount openshift binary inside the node container

pravisankar:
'oadm diagnostics NetworkCheck' uses openshift running on the node to peform network checks.
So mount the openshift binary inside the node container so that the network diagnostic pod can use it.


pravisankar:
@sdodson @openshift/networking PTAL

sdodson:
I don't think this will fix the problem. /usr/bin/openshift already exists inside that container image. Take a look,
`docker run -it --entrypoint /bin/bash openshift/node`

Lets find some time to bluejeans so you guys can fill me in on how this should work.

pravisankar:
This won't fix the issue. Closing this pr in favor of https://github.com/openshift/origin/pull/12127

PR #2813, opened 2016-11-16
Optimize the cloud-specific list.yml playbooks

lhuard1A:
by removing the need to gather facts on all VMs in order to list them.

And prettify the output of AWS list the same way it is done for other cloud providers.

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -608,8 +608,8 @@ class FilterModule(object):
                           host_type=_get_tag_value(host['group_names'], 'host-type'),
                           sub_host_type=_get_tag_value(host['group_names'], 'sub-host-type'),
                           host={'name': host['inventory_hostname'],
-                                'public IP': host['ansible_ssh_host'],
-                                'private IP': host['ansible_default_ipv4']['address']})
+                                'public IP': host['oo_public_ipv4'],
+                                'private IP': host['oo_private_ipv4']})
             except KeyError:
                 pass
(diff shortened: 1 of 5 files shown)


dougbtv:
Any insight on the error I'm having in issue #2856 -- it's failing on the new lines added that look like:

```
      oo_public_ipv4: "{{ hostvars[item].openstack.public_v4 }}"
      oo_private_ipv4: "{{ hostvars[item].openstack.private_v4 }}"
```

in the `playbooks/openstack/openshift-cluster/list.yml` playbook

lhuard1A:
@dougbtv 
Those variables are properly set by the OpenStack dynamic inventory.
Launching the `list.yml` playbook in “stand-alone mode” with `bin/cluster list openstack ''` should work.
The problem is when `list.yml` is invoked as part of `launch.yml` because in that case, I forgot to check the `add_host` task to check that it was properly setting those variables.
Sorry for that regression.
If your `bin/cluster create` failed at that point, it means that it was successfully created. Only the list failed. But if you relaunch a `bin/cluster list openstack ''` at that point, it should work.

Anyhow, I’ve submitted a fix for that issue #2858. Could you validate it?

dougbtv:
Thanks for the excellent detail, I'm happy to rerun to validate it, should
be able to later today or tomorrow morning -- appreciate the quick
submission a bunch! Thanks

dougbtv:
Fix looks good, also commented on the new PR. Thanks again!

PR #2817, opened 2016-11-16
Refactor os_firewall role

mtnbikenc:
* Remove unneeded tasks duplicated by new module functionality
* Ansible systemd module has 'masked' and 'daemon_reload' options
* Ansible firewalld module has 'immediate' option

### DIFF
--- a/roles/os_firewall/README.md
+++ b/roles/os_firewall/README.md
@@ -31,7 +31,6 @@ Use iptables and open tcp ports 80 and 443:
 ---
 - hosts: servers
   vars:
-    os_firewall_use_firewalld: false
     os_firewall_allow:
     - service: httpd
       port: 80/tcp
@@ -46,6 +45,7 @@ Use firewalld and open tcp port 443 and close previously open tcp port 80:
 ---
 - hosts: servers
   vars:
+    os_firewall_use_firewalld: true
     os_firewall_allow:
     - service: https
       port: 443/tcp
--- a/roles/os_firewall/meta/main.yml
+++ b/roles/os_firewall/meta/main.yml
@@ -6,11 +6,11 @@ galaxy_info:
   license: Apache License, Version 2.0
   min_ansible_version: 1.7
   platforms:
-  - name: EL
-    versions:
-    - 7
+    - name: EL
+      versions:
+        - 7
   categories:
-  - system
+    - system
 allow_duplicates: yes
 dependencies:
-- { role: openshift_facts }
+  - role: openshift_facts
--- a/roles/os_firewall/tasks/firewall/firewalld.yml
+++ b/roles/os_firewall/tasks/firewall/firewalld.yml
@@ -2,87 +2,44 @@
 - name: Install firewalld packages
   action: "{{ ansible_pkg_mgr }} name=firewalld state=present"
   when: not openshift.common.is_containerized | bool
-  register: install_result
-
-- name: Check if iptables-services is installed
-  command: rpm -q iptables-services
-  register: pkg_check
-  failed_when: pkg_check.rc > 1
-  changed_when: no
 
 - name: Ensure iptables services are not enabled
-  service:
+  systemd:
     name: "{{ item }}"
     state: stopped
     enabled: no
+    masked: yes
   with_items:
-  - iptables
-  - ip6tables
-  when: pkg_check.rc == 0
-
-- name: Reload systemd units
-  command: systemctl daemon-reload
-  when: install_result | changed
-
-- name: Determine if firewalld service masked
-  command: >
-    systemctl is-enabled firewalld
-  register: os_firewall_firewalld_masked_output
-  changed_when: false
-  failed_when: false
-
-- name: Unmask firewalld service
-  command: >
-    systemctl unmask firewalld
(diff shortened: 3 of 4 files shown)


mtnbikenc:
@detiber PTAL

detiber:
@mtnbikenc looks good. We'll probably want to hold of merging until after 3.4 ships.

We'll also need to figure out when we want to change the default, since we might need to handle migration during an upgrade. @sdodson @dgoodwin thoughts?

dgoodwin:
How does it behave if you had the old style firewall, then re-ran config and picked up this change? 

Definitely makes you think about upgrade, if we can just run the role though it's probably no big deal, but where is the line on what we should do during upgrade what we should preserve if they're not using our config playbooks for on-going maintenance.

detiber:
@dgoodwin upgrade/re-run is definitely an issue we need to consider here.

Would it be worth it to test for the existence of the chain we create to assist in setting the default? That way if the chain already exists then we would prevent using firewalld as the default?

Biggest thing would be a re-run that switches the firewall provider would result in availability issues as the iptables-services is stopped and before firewalld is running with the proper rules.

Another concern is that we would need to ensure proper documentation around the default to avoid unexpected issues with users that did not expect us to disable/mask iptables (users that may be managing iptables rules using other configuration management for example).

mtnbikenc:
@detiber This PR is to change the tasks to use `systemd` instead of `service`.  The change to the docs for `os_firewall_use_firewalld: true` is only to correctly match the current defaults in the code.  The default is currently `os_firewall_use_firewalld: false` so the examples in the README needed to be switched around.  Is there more that is needed for this PR?

Rebased and squashed.

detiber:
@mtnbikenc thanks for the clarification.

detiber:
aos-ci-test

detiber:
@sdodson any objections to this merging prior to 3.4 being wrapped up?

sdodson:
Fine with me.

dustymabe:
This breaks on atomic host where firewalld is not installed. You removed the "Check if firewalld is installed" step and now we get this:

```
TASK [os_firewall : Ensure firewalld service is not enabled] *******************
fatal: [ec2-54-161-88-56.compute-1.amazonaws.com]: FAILED! => {"changed": false, "failed": true, "failed_when_result": true, "msg": "failure 1 running systemctl show for 'firewalld': Unit firewalld.service could not be found.\n"}
```

detiber:
@dustymabe thanks for the heads up. @mtnbikenc could you re-add that check?

mtnbikenc:
That check was incorporated into the failed_when for the systemd task, however it appears that the resulting failure message is different between rhel and atomic host.  Investigating.

dustymabe:
I think shouldn't run the `systemctl show firewalld` task if firewalld is not installed, otherwise you get the error.

mtnbikenc:
We are not running `systemctl show firewalld` directly, the systemd module is.  Also, the systemd module should gracefully exit if the module is not found, however in this instance the task is exiting the module at this point, https://github.com/ansible/ansible-modules-core/blob/devel/system/systemd.py#L278

I tested this successfully on a local atomic host 7.3 install.

@dustymabe Could you provide more details/versions of your test environment?

dustymabe:
```
-bash-4.3# systemctl show firewalld.service
Unit firewalld.service could not be found.
-bash-4.3# echo $?
1
-bash-4.3# cat /etc/os-release 
NAME=Fedora
VERSION="25 (Atomic Host)"
ID=fedora
VERSION_ID=25
PRETTY_NAME="Fedora 25 (Atomic Host)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:25"
HOME_URL="https://fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=25
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=25
PRIVACY_POLICY_URL=https://fedoraproject.org/wiki/Legal:PrivacyPolicy
VARIANT="Atomic Host"
VARIANT_ID=atomic.host
```

mtnbikenc:
There is a difference in the way systemctl show exits for Fedora 25 or RHEL 7 based systems.  This is causing the systemd module to fail on Fedora 25 systems due to the rc=1 for `systemctl show`.

I've filed a bug with Ansible: https://github.com/ansible/ansible-modules-core/issues/5710

In the meantime, I'll put a workaround in place until this is fixed.

mtnbikenc:
Looks like systemd v232 could resolve this as well.
https://github.com/systemd/systemd/commit/bd5b9f0a12dd9c1947b11534e99c395ddf44caa9

dustymabe:
> Looks like systemd v232 could resolve this as well.
> systemd/systemd@bd5b9f0

yeah. I'm not sure when we'll get there though. I have a workaround now that is working. Would you like me to submit a PR or are you already working on one?

mtnbikenc:
@dustymabe I submitted a PR as noted above.  Thanks for reporting this problem.

PR #2524, opened 2016-09-28
ip forwarding for Nuage SDN

vishpat:
@detiber @abutcher 

Enable iptable forward rules required for Nuage SDN.

PR #2819, opened 2016-11-16
Added ip forwarding for nuage

vishpat:
@detiber @sdodson

### DIFF
--- a/roles/nuage_node/handlers/main.yaml
+++ b/roles/nuage_node/handlers/main.yaml
@@ -6,3 +6,7 @@
 - name: restart node
   become: yes
   service: name={{ openshift.common.service_type }}-node state=restarted
+
+- name: save iptable rules
+  become: yes
+  command: iptables-save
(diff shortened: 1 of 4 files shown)


sdodson:
aos-ci-test

sdodson:
Merging based on green CI tests and LGTM in #2524

roles/nuage_node/handlers/main.yaml
```diff
@@ -6,3 +6,7 @@
 - name: restart node
   become: yes
   service: name={{ openshift.common.service_type }}-node state=restarted
+
+- name: save iptable rules
+  become: yes
+  command: iptables-save
```
mtnbikenc: This command is not actually saving the iptables rules.  iptables-save just dumps the loaded tables to stdout.  Additionally, a wholesale saving of iptables rules has been discouraged by the OpenShift networking team.  A possible alternative is using the os_firewall role to create rules.  cc: @detiber @sdodson

PR #2827, opened 2016-11-18
Allow ansible to continue when a node is unaccessible or fails.

abutcher:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1377619
Supersedes #2666

### DIFF
--- a/roles/openshift_manage_node/tasks/main.yml
+++ b/roles/openshift_manage_node/tasks/main.yml
@@ -23,6 +23,7 @@
   delay: 5
   changed_when: false
   with_items: "{{ openshift_nodes }}"
+  when: hostvars[item].openshift.node.nodename is defined
 
 - name: Set node schedulability
   command: >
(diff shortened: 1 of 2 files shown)


abutcher:
aos-ci-test

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -19,10 +19,12 @@
     -n default
   register: omd_get_node
   until: omd_get_node.rc == 0
-  retries: 50
+  retries: 10
   delay: 5
   changed_when: false
   with_items: "{{ openshift_nodes }}"
+  when: hostvars[item].openshift.node.nodename is defined
+  ignore_errors: true
```
detiber: I'm not sure I'm a fan of this approach, seems like it'll lead to other issues if we just allow all of these to fail.

What if before openshift_manage_node is applied we had some task that attempts to poll for the availability similar to above, and then if the retries is reached, we only run o_m_n against the set of hosts that are available from `oc get nodes`?

abutcher: Agreed. There is no `any_errors_fatal: true` so I'm not yet sure why this is causing the whole play to error out, rather than continue with nodes that succeeded.

tbielawa:
aos-ci-test

PR #2831, opened 2016-11-21
Fix rare failure to deploy new registry/router after upgrade.

dgoodwin:
Router/registry update and re-deploy was recently reordered to
immediately follow control plane upgrade, right before we proceed to
node upgrade.

In some situations (small or single host clusters) it appears possible
that the deployer pods are running when the node in question is
evacuated for upgrade. When the deployer pod dies the deployment is
failed and the router/registry continue running the old version, despite
the deployment config being updated correctly.

This change re-orderes the router/registry upgrade to follow node
upgrade. However for separate control plane upgrade, the router/registry
still occurs at the end. This is because router/registry seems like they
should logically be included in a control plane upgrade, and presumably
the user will not manually launch node upgrade so quickly as to trigger
an evac on the node in question.

Workaround for this problem when it does occur is simply to:

oc deploy docker-registry --latest

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1395081

### DIFF
--- a/playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
+++ b/playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
@@ -92,9 +92,9 @@
   vars:
     master_config_hook: "v3_3/master_config_upgrade.yml"
 
-- include: ../../../../common/openshift-cluster/upgrades/post_control_plane.yml
-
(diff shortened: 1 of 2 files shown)


dgoodwin:
@sdodson this is a pretty rare edge case with an easy workaround, I'm leaving the bug off blocker list and IMO no need to bring back to 3.3.

sdodson:
oh you said not a 3.4 blocker, so removing that label.

PR #2832, opened 2016-11-21
Simplify OpenShift service debug levels.

dgoodwin:
Stop using openshift_facts for debug levels entirely as this is not a fact about
the system, rather a configuration parameter of the playbooks/inventory.

Deprecate debug_level and rename to openshift_debug_level. Default is
2 if undefined, and can be specified separately for master/node if
desired.

Moved all definitions/defaulting straight to systemd_units.yml for
master/node where the values are actually used. Here we check if
openshift_debug_level is set and default it if not, and then set the
master/node variant if they are not set.

NOTE: Vars can't be used in the roles here as the systemd_units.yml
files are used separtely during upgrade.

@detiber @abutcher @sdodson could use one of your reviews here if possible.

@sdodson this is not explicitly for 3.4, I'm not sure it should be included at all. Based on end result it probably looks safe but it could be disruptive.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -25,7 +25,6 @@
       role: master
       local_facts:
         embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
-        debug_level: "{{ openshift_master_debug_level | default(openshift.common.debug_level | default(2)) }}"
 
 - name: Backup etcd
   include: ./etcd/backup.yml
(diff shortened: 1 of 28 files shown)


playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -25,10 +25,6 @@
       role: master
       local_facts:
         embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
-        debug_level: "{{ openshift_master_debug_level | default(openshift.common.debug_level | default(2)) }}"
-
-- name: Backup etcd
-  include: ./etcd/backup.yml
```
mtnbikenc: Did you intend to remove Backup etcd?

dgoodwin: No! Thanks, restoring now.

mtnbikenc:

PR #2833, opened 2016-11-21
Only run tuned-adm if tuned exists.

dustymabe:
Fedora Atomic Host does not have tuned installed.


detiber:
aos-ci-test

detiber:
@sdodson we'll need to figure out what we'll need to do about tuned profiles for newer builds of atomic host, but I'm merging this since it breaks f25 installs today.

PR #2836, opened 2016-11-21
Merge admission plugin configs

### LINKED ISSUE #2462: changes to openshift_master_kube_admission_plugin_config in 1.3
Using an old 1.2 inventory file which had `openshift_master_kube_admission_plugin_config` fails to start the controller. 

It appears to be related to changes in the admission controller https://docs.openshift.org/latest/install_config/master_node_configuration.html

Merging the kube admission config with `openshift_master_admission_plugin_config` resolves this.

abutcher:
Fixes Bug 1393645
Fixes #2462
Supersedes #2782 

This change deprecates the following inventory variables:

```
openshift_master_admission_plugin_order
openshift_master_kube_admission_plugin_order
openshift_master_kube_admission_plugin_config
```

Based on the [documentation](https://docs.openshift.com/container-platform/latest/architecture/additional_concepts/admission_controllers.html), plugin order should no longer be set and all admission plugin configuration should be moved to `admissionConfig.pluginConfig` in `/etc/origin/master/master-config.yaml`. Admission plugin configuration set within `openshift_master_kube_admission_plugin_config` will be merged into configuration set within `openshift_master_admission_plugin_config`.

Since inventories may not be updated to reflect this change, it is still possible to set `openshift_master_kube_admission_plugin_config` which will always be merged into `openshift_master_admission_plugin_config`

https://bugzilla.redhat.com/show_bug.cgi?id=1393645

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/v3_3/master_config_upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_3/master_config_upgrade.yml
@@ -48,3 +48,18 @@
     dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
     yaml_key: 'controllerConfig.servicesServingCert.signer.keyFile'
     yaml_value: service-signer.key
+
+- modify_yaml:
+    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+    yaml_key: 'admissionConfig.pluginConfig'
+    yaml_value: "{{ openshift.master.admission_plugin_config }}"
+
+- modify_yaml:
+    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+    yaml_key: 'admissionConfig.pluginOrderOverride'
+    yaml_value:
+
+- modify_yaml:
+    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+    yaml_key: 'kubernetesMasterConfig.admissionConfig'
+    yaml_value:
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -619,6 +619,9 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # masterConfig.volumeConfig.dynamicProvisioningEnabled, configurable as of 1.2/3.2, enabled by default
 #openshift_master_dynamic_provisioning_enabled=False
 
+# Admission plugin config
+#openshift_master_admission_plugin_config={"ProjectRequestLimit":{"configuration":{"apiVersion":"v1","kind":"ProjectRequestLimitConfig","limits":[{"selector":{"admin":"true"}},{"maxProjects":"1"}]}},"PodNodeConstraints":{"configuration":{"apiVersion":"v1","kind":"PodNodeConstraintsConfig"}}}
+
 # Configure usage of openshift_clock role.
 #openshift_clock_enabled=true
 
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -619,6 +619,9 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # masterConfig.volumeConfig.dynamicProvisioningEnabled, configurable as of 1.2/3.2, enabled by default
 #openshift_master_dynamic_provisioning_enabled=False
 
+# Admission plugin config
+#openshift_master_admission_plugin_config={"ProjectRequestLimit":{"configuration":{"apiVersion":"v1","kind":"ProjectRequestLimitConfig","limits":[{"selector":{"admin":"true"}},{"maxProjects":"1"}]}},"PodNodeConstraints":{"configuration":{"apiVersion":"v1","kind":"PodNodeConstraintsConfig"}}}
+
 # Configure usage of openshift_clock role.
 #openshift_clock_enabled=true
 
--- a/playbooks/common/openshift-cluster/upgrades/v3_4/master_config_upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_4/master_config_upgrade.yml
@@ -1,0 +1,15 @@
+---
+- modify_yaml:
+    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+    yaml_key: 'admissionConfig.pluginConfig'
+    yaml_value: "{{ openshift.master.admission_plugin_config }}"
+
+- modify_yaml:
+    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+    yaml_key: 'admissionConfig.pluginOrderOverride'
+    yaml_value:
+
+- modify_yaml:
+    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+    yaml_key: 'kubernetesMasterConfig.admissionConfig'
+    yaml_value:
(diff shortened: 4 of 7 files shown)


detiber:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/v3_3/master_config_upgrade.yml
```diff
@@ -48,3 +48,13 @@
...
+
+- modify_yaml:
+    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+    yaml_key: 'admissionConfig.pluginConfig'
+    yaml_value: "{{ openshift.master.admission_plugin_config }}"
+
+- modify_yaml:
+    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+    yaml_key: 'kubernetesMasterConfig.admissionConfig'
+    yaml_value:
```
dgoodwin: Is this also supposed to happen on 3.3 upgrades?

If this is something we want to do every upgrade going forward if release is gte 3.3, we probably don't want hooks and instead should plug it into https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml#L57 and have it run if the gta 3_3 var is set.

If it's one time 3.4 only, then the hooks look correct but suspect this 3.3 block could go away.

If it's literally just run for 3.3 and 3.4 upgrades only, then this is all correct.

playbooks/byo/openshift-cluster/upgrades/v3_4/upgrade_control_plane.yml
```diff
@@ -94,5 +94,7 @@
   - include: ../../../../common/openshift-cluster/upgrades/cleanup_unused_images.yml
 
 - include: ../../../../common/openshift-cluster/upgrades/upgrade_control_plane.yml
+  vars:
+    master_config_hook: "v3_4/master_config_upgrade.yml"
```
dgoodwin: This needs to be also be set when we include upgrade_control_plane.yml in the regular upgrade.yml, same dir as this file.

abutcher: Thanks, fixed!

abutcher:
@dak1n1 Heads up on this change.

playbooks/common/openshift-cluster/upgrades/v3_3/master_config_upgrade.yml
abutcher: Since the facts are being removed/migrated I'd like to limit this to 3.3 and 3.4 upgrades.

playbooks/common/openshift-cluster/upgrades/v3_3/master_config_upgrade.yml
```diff
@@ -48,3 +48,18 @@
...
+
+- modify_yaml:
+    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+    yaml_key: 'admissionConfig.pluginConfig'
+    yaml_value: "{{ openshift.master.admission_plugin_config }}"
+
+- modify_yaml:
+    dest: "{{ openshift.common.config_base}}/master/master-config.yaml"
+    yaml_key: 'admissionConfig.pluginConfig.pluginOrderOverride'
+    yaml_value:
```
detiber: Would we always want to override pluginOrderOverride and set it to null? That appears to be what is happening here.

playbooks/common/openshift-cluster/upgrades/v3_3/master_config_upgrade.yml
detiber: If this affects 3.3 clusters, then we should probably do it for 3.2 -> 3.3 upgrades. Wouldn't 3.3 systems that have this set already be broken prior to the 3.3 -> 3.4 upgrade?

detiber:
@abutcher will need to get some test cases together for QE for this as well.

playbooks/common/openshift-cluster/upgrades/v3_3/master_config_upgrade.yml
abutcher: Right, based on the docs the order override should no longer be set.

playbooks/common/openshift-cluster/upgrades/v3_3/master_config_upgrade.yml
abutcher: Starting in 3.3 they should exist at the same level so I believe that is correct. These changes cause 3.3 and 3.4 upgrades to fix the config.

abutcher:
Testing by setting the following environment variables:

```
openshift_master_admission_plugin_config={"ProjectRequestLimit":{"configuration":{"apiVersion":"v1","kind":"ProjectRequestLimitConfig","limits":[{"selector":{"admin":"true"}},{"maxProjects":"1"}]}},"PodNodeConstraints":{"configuration":{"apiVersion":"v1","kind":"PodNodeConstraintsConfig"}}}
openshift_master_kube_admission_plugin_config={"RunOnceDuration":{"configuration":{"apiVersion":"v1","kind":"RunOnceDurationConfig","activeDeadlineSecondsOverride":"3600"}},"ClusterResourceConfig":{"configuration":{"apiVersion":"v1","kind":"ClusterResourceOverrideConfig","limitCPUToMemoryPercent":"200","cpuRequestToLimitPercent":"6","memoryRequestToLimitPercent":"60"}},"PodNodeConstraints":{"configuration":{"apiVersion":"v1","kind":"PodNodeConstraintsConfig"}}}
```

Installing from master branch yields master config sections:

```
admissionConfig:
  pluginConfig:
    PodNodeConstraints:
      configuration:
        apiVersion: v1
        kind: PodNodeConstraintsConfig
    ProjectRequestLimit:
      configuration:
        apiVersion: v1
        kind: ProjectRequestLimitConfig
        limits:
        - selector:
            admin: 'true'
        - maxProjects: '1'
```

```
kubernetesMasterConfig:
  admissionConfig:
    pluginConfig:
      ClusterResourceConfig:
        configuration:
          apiVersion: v1
          cpuRequestToLimitPercent: '6'
          kind: ClusterResourceOverrideConfig
          limitCPUToMemoryPercent: '200'
          memoryRequestToLimitPercent: '60'
      PodNodeConstraints:
        configuration:
          apiVersion: v1
          kind: PodNodeConstraintsConfig
      RunOnceDuration:
        configuration:
          activeDeadlineSecondsOverride: '3600'
          apiVersion: v1
          kind: RunOnceDurationConfig
```

After upgrading with this branch the combined config is:

```
admissionConfig:
  pluginConfig:
    ClusterResourceConfig:
      configuration:
        apiVersion: v1
        cpuRequestToLimitPercent: '6'
        kind: ClusterResourceOverrideConfig
        limitCPUToMemoryPercent: '200'
        memoryRequestToLimitPercent: '60'
    PodNodeConstraints:
      configuration:
        apiVersion: v1
        kind: PodNodeConstraintsConfig
    ProjectRequestLimit:
      configuration:
        apiVersion: v1
        kind: ProjectRequestLimitConfig
        limits:
        - selector:
            admin: 'true'
        - maxProjects: '1'
    RunOnceDuration:
      configuration:
        activeDeadlineSecondsOverride: '3600'
        apiVersion: v1
        kind: RunOnceDurationConfig
  pluginOrderOverride:
```

abutcher:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/v3_3/master_config_upgrade.yml
abutcher: #2462 points to this being a problem in 1.3 so this change impacts 3.2->3.3 upgrade as well as 3.3->3.4.

PR #2840, opened 2016-11-22
Make the role work on F25 Cloud

mscherer:
On F24 and earlier, systemctl show always returned 0. On F25, it
return 1 when a service do not exist, and thus the role fail
on Fedora 25 cloud edition.


mscherer:
aos-ci-test

sdodson:
Is NetworkManager not used on Fedora 25 cloud images? We require it, so fixing the check just means that we'll later fail the run indicating that NM is required.

mscherer:
Yeah, it is not used, at least not on the F25 I did test. And yes, it fail after saying it must be installed and or running. I just sent the patch out of love for correctness.

PR #2845, opened 2016-11-22
Fix issues encountered in mixed environments

abutcher:


sdodson:
Looks good enough to solve the cert issue, so LGTM. But probably not quite enough to solve https://bugzilla.redhat.com/show_bug.cgi?id=1364160  perhaps we should create a bug that's more limited in scope for that?

abutcher:
@sdodson I've updated this to cover the upgrade case as well. I'm not sure what the backports will look like but I think we've got the BZ covered now.

sdodson:
aos-ci-test

abutcher:
Reproduced and verified that this fixes the install case in my testing. Master is rpm based and I had one node which was containerized.

PR #2853, opened 2016-11-23
Set nameservers on DHCPv6 event

alexcern:
A dhcp6-change event may happen on nodes running dual stack
IPv4/IPv6 and DHCP, even if Openshift itself doesn't use IPv6.
/etc/resolv.conf needs be adjusted as well in this case.


detiber:
ok to test

detiber:
aos-ci-test

detiber:
@sdodson any objections to this change?

sdodson:
LGTM, thanks for the PR!

Issue #2714, opened 2016-11-03
scheduler template is outdated


liggitt:
Current defaults are at https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/plugin/pkg/scheduler/algorithmprovider/defaults/defaults.go#L113-L206

Need to audit differences between those defaults and what ansible lays down, and figure out how to prevent drift in the future

PR #2855, opened 2016-11-23
Update scheduler defaults

detiber:
resolves: https://github.com/openshift/openshift-ansible/issues/2714

### DIFF
--- a/roles/openshift_master/vars/main.yml
+++ b/roles/openshift_master/vars/main.yml
@@ -1,17 +1,14 @@
 ---
-openshift_master_config_dir: "{{ openshift.common.config_base }}/master"
-openshift_master_config_file: "{{ openshift_master_config_dir }}/master-config.yaml"
 openshift_master_loopback_config: "{{ openshift_master_config_dir }}/openshift-master.kubeconfig"
 loopback_context_string: "current-context: {{ openshift.master.loopback_context_name }}"
-openshift_master_scheduler_conf: "{{ openshift_master_config_dir }}/scheduler.json"
 openshift_master_session_secrets_file: "{{ openshift_master_config_dir }}/session-secrets.yaml"
 openshift_master_policy: "{{ openshift_master_config_dir }}/policy.json"
 
 scheduler_config:
   kind: Policy
   apiVersion: v1
-  predicates: "{{ openshift.master.scheduler_predicates }}"
-  priorities: "{{ openshift.master.scheduler_priorities }}"
+  predicates: "{{ openshift_master_scheduler_predicates }}"
+  priorities: "{{ openshift_master_scheduler_priorities }}"
 
 openshift_master_valid_grant_methods:
 - auto
--- a/roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py
+++ b/roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py
@@ -1,0 +1,96 @@
+# pylint: disable=missing-docstring
+
+import re
+from ansible.errors import AnsibleError
+from ansible.plugins.lookup import LookupBase
+
+
+class LookupModule(LookupBase):
+    # pylint: disable=too-many-branches,too-many-statements
+
+    def run(self, terms, variables=None, regions_enabled=True, **kwargs):
+        if 'openshift' not in variables:
+            raise AnsibleError("This lookup module requires openshift_facts to be run prior to use")
+        if 'master' not in variables['openshift']:
+            raise AnsibleError("This lookup module is meant to be run against an OpenShift master host only")
+
+        if 'openshift_master_scheduler_predicates' in variables:
+            return variables['openshift_master_scheduler_predicates']
+        elif 'scheduler_predicates' in variables['openshift']['master']:
+            return variables['openshift']['master']['scheduler_predicates']
+        else:
+            predicates = []
+
+            if 'deployment_type' not in variables['openshift']['common']:
+                raise AnsibleError("This lookup module requires that the deployment_type be set")
+
+            deployment_type = variables['openshift']['common']['deployment_type']
+
+            if 'short_version' in variables['openshift']['common']:
+                short_version = variables['openshift']['common']['short_version']
+            elif 'openshift_release' in variables:
+                release = variables['openshift_release']
+                if release.startswith('v'):
+                    short_version = release[1:]
+                else:
+                    short_version = release
+            elif 'openshift_version' in variables:
+                version = variables['openshift_version']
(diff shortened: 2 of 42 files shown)


detiber:
aos-ci-test

detiber:
@dgoodwin @abutcher ptal. I still need to work on retrofitting this into upgrades, but this should handle the new install use case.

roles/openshift_master/vars/main.yml
```diff
@@ -1,17 +1,14 @@
...
 openshift_master_session_secrets_file: "{{ openshift_master_config_dir }}/session-secrets.yaml"
 openshift_master_policy: "{{ openshift_master_config_dir }}/policy.json"
 
 scheduler_config:
   kind: Policy
   apiVersion: v1
-  predicates: "{{ openshift.master.scheduler_predicates }}"
-  priorities: "{{ openshift.master.scheduler_priorities }}"
+  predicates: "{{ openshift_master_scheduler_predicates }}"
+  priorities: "{{ openshift_master_scheduler_priorities }}"
```
dgoodwin: 👍

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py
```diff
@@ -0,0 +1,83 @@
...
+                        'argument': {
+                            'serviceAffinity': {
+                                'labels': ['region']
+                            }
+                        }
+                }
+                predicates.append(region_predicate)
+
+
+            return predicates
```
dgoodwin: What do you think about adding some unit tests for these modules? They look like great candidates for it, just need a little import magic for the test/ module and it should work fine.

detiber:
aos-ci-test

detiber:
aos-ci-test

sdodson:
I'm not surprised that the 3.3 tests are failing as I suspect we're using defaults that wouldn't work there, but I wonder what's up with the 3.4 tests.

sdodson:
aos-ci-test

detiber:
@sdodson I'm still validating the pylint and flak8e fixes, wouldn't be surprised if something is off from those. That said, the defaults *should* work right for all of these, but I might be handling openshift_version/openshift_release wrong

detiber:
@dgoodwin adding those tests sent me down a huge rabbit hole, which made me realize we weren't testing a lot of the custom code outside of utils/

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
@sdodson I believe this is good to go for new installs? I can follow up with the upgrade work in another PR.

utils/setup.cfg
```diff
@@ -3,3 +3,16 @@
 # 3. If at all possible, it is good practice to do this. If you cannot, you
 # will need to generate wheels for each Python version that you support.
 universal=1
+
+[nosetests]
+tests=../,../roles/openshift_master_facts/test/,test/
```
rhcarvalho: @detiber hi there. When rebasing a local branch I noticed you removed `/setup.cfg` and moved the contents here, however this means that running `nosetests` from the repository root doesn't use the config file and run just 2 tests. Why did you move it? Do we expect developers to `cd utils` to run the tests, or maybe I'm missing something in my local setup?

```console
openshift-ansible$ nosetests
..
----------------------------------------------------------------------
Ran 2 tests in 0.040s

OK
openshift-ansible$ (cd utils && nosetests)
Usage: nosetests [options]

nosetests: error: Error reading config file 'setup.cfg': no such option 'with_coverage'
```

(looks like all options should have a `-` and not a `_`, I can fix that)

```console
openshift-ansible$ vim utils/setup.cfg   # replace _ with -
openshift-ansible$ (cd utils && nosetests)
...
Verify is_valid_hostname can detect None or too-long hostnames ... ok
Verify is_valid_hostname can parse regular hostnames ... ok

Name                             Stmts   Miss  Cover
----------------------------------------------------
ooinstall.py                         0      0   100%
ooinstall/cli_installer.py         602    169    72%
ooinstall/oo_config.py             262     68    74%
ooinstall/openshift_ansible.py     203     84    59%
ooinstall/utils.py                  14      0   100%
ooinstall/variants.py               37      1    97%
----------------------------------------------------
TOTAL                             1118    322    71%
----------------------------------------------------------------------
Ran 117 tests in 0.715s

OK

Comment (rhcarvalho) [inline utils/setup.cfg]: note: the `-` x `_` failed in one environment but worked in a different one for me. Possibly due to differing versions of deps.

Comment (detiber) [inline utils/setup.cfg]: @rhcarvalho I moved it, since we run the tests from the Makefile in utils and I was unable to get the tests running from the Makefile properly without having the setup.cfg in the same directory.

I'd be okay if we move the Makefile up to the root, this was just the path of least resistance to get the jenkins and travisci jobs to run all of the tests.  I'll fix the '-' issue' with a PR, but would need someone else to contribute the changes needed to move the Makefile to the root.

Comment (rhcarvalho) [inline utils/setup.cfg]: thanks @detiber! I just wanted to understand which direction we're  going. I can help with the Makefile changes, in fact had yet another Makefile in the root to run some new tests we're adding.

**Comment by detiber** [comment]
@rhcarvalho I created the following PR to address the setup.cfg issue (and improve coverage for other tested files): https://github.com/openshift/openshift-ansible/pull/2927

File: roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_priorities.py [hunk 4]
```diff
@@ -0,0 +1,87 @@
...
+            elif deployment_type == 'openshift_enterprise':
+                if short_version not in ['3.1', '3.2', '3.3', '3.4']:
+                    raise AnsibleError("Unknown short_version %s" % short_version)
+
+            if deployment_type == 'openshift-enterprise':
+                # convert short_version to origin short_version
+                short_version = re.sub('^3.', '1.', short_version)
+
+            if short_version == '1.4':
+                priorities.append({'name': 'NodePreferAvoidPodsPriority', 'weight': 10000})
```
jayunit100: for the future i assume these priorities/predicate weights would be best if externalized, and commented/explained ?

detiber: @jayunit100 agreed, currently they are copied directly from the origin codebase: https://github.com/openshift/origin/blob/release-1.4/vendor/k8s.io/kubernetes/plugin/pkg/scheduler/algorithmprovider/defaults/defaults.go#L113-L206

PR #2859, opened 2016-11-24
Cleanup ovs file and restart docker on every upgrade.

dgoodwin:
In 3.3 one of our services lays down a systemd drop-in for configuring
Docker networking to use lbr0. In 3.4, this has been changed but the
file must be cleaned up manually by us.

However, after removing the file docker requires a restart. This had big
implications particularly in containerized environments where upgrade is
a very fragile series of upgrading and service restarts.

To avoid double docker restarts, and thus double service restarts in
containerized environments, this change does the following:

- Skip restart during docker upgrade, if it is required. We will restart
  on our own later.
- Skip containerized service restarts when we upgrade the services
  themselves.
- Clean shutdown of all containerized services.
- Restart Docker. (always, previously this only happened if it needed an
  upgrade)
- Ensure all containerized services are restarted.
- Restart rpm node services. (always)
- Mark node schedulable again.

At the end of this process, docker0 should be back on the system.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1395069

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -29,6 +29,7 @@
 
 - name: Backup etcd
   include: ./etcd/backup.yml
+  when: openshift_upgrade_skip_etcd_backup | default(false) | bool
 
 - name: Upgrade master packages
   hosts: oo_masters_to_config
--- a/playbooks/common/openshift-cluster/upgrades/containerized_node_upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/containerized_node_upgrade.yml
@@ -1,9 +1,14 @@
+---
+# This is a hack to allow us to use systemd_units.yml, but skip the handlers which
+# restart services. We will unconditionally restart all containerized services
+# because we have to unconditionally restart Docker:
+- set_fact:
+    skip_node_svc_handlers: True
+
 - name: Update systemd units
   include: ../../../../roles/openshift_node/tasks/systemd_units.yml openshift_version={{ openshift_image_tag }}
 
-- name: Verifying the correct version was configured
-  shell: grep {{ verify_upgrade_version }} {{ item }}
-  with_items:
-    - /etc/sysconfig/openvswitch
-    - /etc/sysconfig/{{ openshift.common.service_type }}*
-  when: verify_upgrade_version is defined
+# This is a no-op because of skip_node_svc_handlers, but lets us trigger it before end of
+# play when the node has already been marked schedulable again. (this would look strange
+# in logs otherwise)
+- meta: flush_handlers
--- a/playbooks/common/openshift-cluster/upgrades/docker/restart.yml
+++ b/playbooks/common/openshift-cluster/upgrades/docker/restart.yml
@@ -1,0 +1,29 @@
+---
+- name: Restart docker
+  service: name=docker state=restarted
+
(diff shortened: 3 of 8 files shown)


dgoodwin:
This was extremely tricky to get right and quite unfortunate to have to be doing at this point in a release cycle. I've spent several days trying to get it pinned down and testing on rpm and container environments, at this point I think the upgrade is clean and docker0 always is restored on both masters and nodes. 

CC @dcbw please keep an eye out here and make sure this matches your expectations for what should happen during an ovs upgrade. (both rpm and containerized)

danwinship:
> Restart Docker. (always, previously this only happened if it needed an upgrade)

That's "always (when doing an upgrade)", not "always (even for new installs)", right? Assuming so, that sounds right.

openshift-ansible *shouldn't* have to restart docker in the non-containerized case; OpenShift is supposed to be doing that itself (and it works in local testing... maybe there's an selinux issue or something tripping it up in the rpm-ified install). So if getting rid of the docker restart in the non-containerized case would make your patch simpler, we could try to figure that out so you don't have to do it here. But if it's just as easy for you to restart docker in both cases then we might as well leave it as is.

sdodson:
I think just a point of clarification is that openshift will restart docker when the node restarts after having been upgraded to 3.4. That's also a fairly recent change that may not have been in the builds prior to last Wednesday.

dgoodwin:
@danwinship 

> That's "always (when doing an upgrade)", not "always (even for new installs)", right? Assuming so, that sounds right.

Correct this is just a step in the upgrade playbook, not run in normal install/maintenance.

> openshift-ansible shouldn't have to restart docker in the non-containerized case; OpenShift is supposed to be doing that itself (and it works in local testing... maybe there's an selinux issue or something tripping it up in the rpm-ified install). So if getting rid of the docker restart in the non-containerized case would make your patch simpler, we could try to figure that out so you don't have to do it here. But if it's just as easy for you to restart docker in both cases then we might as well leave it as is.

Fairly positive I saw in testing last week that if you did not do that docker restart even for rpm based clusters, you would not get docker0 without an explicit docker restart. QE reported similar here: https://bugzilla.redhat.com/show_bug.cgi?id=1395069#c24

danwinship:
Right, it's not currently working correctly, but the fact that it doesn't work for RPM installs is potentially a *separate* bug from the fact that it doesn't work for containerized installs, and if necessary, it might be possible to fix that part on the origin end rather than here. (But if that wouldn't make this patch any simpler then there isn't any reason to do that.)

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -29,6 +29,7 @@
 
 - name: Backup etcd
   include: ./etcd/backup.yml
+  when: openshift_upgrade_skip_etcd_backup is not defined or not openshift_upgrade_skip_etcd_backup | bool
```
sdodson: Is this part of some other change or just additional control you added while debugging this?
I think we've decided that we prefer the following rather than checking for is not defined
`when: openshift_upgrade_skip_etcd_backup | default(true) | bool`

sdodson: Yeah I'm fine leaving it in.

sdodson:
Needs a rebase btw.

dgoodwin:
Rebased.

sdodson:
aos-ci-test

PR #2861, opened 2016-11-25
Fix access_modes initialization

luisfdez:
This PR fixes how the `access_modes` are initialized from the parameters.

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -708,7 +708,7 @@ class FilterModule(object):
                             volume = params['volume']['name']
                             path = directory + '/' + volume
                             size = params['volume']['size']
-                            access_modes = params['access_modes']
+                            access_modes = params['access']['modes']
                             persistent_volume = dict(
                                 name="{0}-volume".format(volume),
                                 capacity=size,
@@ -721,7 +721,7 @@ class FilterModule(object):
                         elif kind == 'openstack':
                             volume = params['volume']['name']
                             size = params['volume']['size']
-                            access_modes = params['access_modes']
+                            access_modes = params['access']['modes']
                             filesystem = params['openstack']['filesystem']
                             volume_id = params['openstack']['volumeID']
                             persistent_volume = dict(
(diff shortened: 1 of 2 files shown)


sdodson:
Hmm, i would've expected it to be fine how it was. @abutcher thoughts?

sdodson:
bot, test pull request

abutcher:
LGTM but we'll also need to nest the `access_modes` in the default facts dict in the `openshift_facts` module https://github.com/openshift/openshift-ansible/blob/14debfe6456ee6185e77c160da0b8b18641b4694/roles/openshift_facts/library/openshift_facts.py#L1981-L2036. Thanks @luisfdez

luisfdez:
Thanks @abutcher for the comments (sorry for the delay but I missed the notifications).

I just updated the code with your suggestions and rebased the code against master.

Please let me know if now is fined or further modifications are needed.

Cheers,
Luis

abutcher:
aos-ci-test

abutcher:
ok to test

rhcarvalho:
> Hmm, i would've expected it to be fine how it was.

I also failed to understand what was fixed here.

[`PersistentVolumeSpec.AccessModes`](https://github.com/kubernetes/kubernetes/blob/16f7cd189c1a1ec9fc2ef2fd2e4e29b2b333deb9/pkg/api/v1/types.go#L452) is a list of strings, as per the Kubernetes API, there isn't a concept of "access" with a "mode" inside.

This is different than e.g. `params['volume']` that has a `'name'` and `'size'`. The code strangely never uses `params['access']` alone, but instead always access two levels `params['access']['modes']`.

For the sake of learning something new, is there something more to access modes that require this change?

abutcher:
Hey @rhcarvalho, the problem was that parameters passed in like `openshift_hosted_registry_storage_access_modes=['ReadWriteOnce']` get automatically split by `openshift_facts` into a dictionary like `{"openshift": "hosted": {"registry": {"storage": "access": {"modes": ["ReadWriteOnce"]}}}}` and overriding access modes wasn't possible since we were attempting to access the key `"access_modes"`. Variable splitting like this is only done for a subset of parameters and will likely be phased out piece by piece because it's kind of magical and hard to deal with.

rhcarvalho:
> the problem was that parameters passed in like openshift_hosted_registry_storage_access_modes=['ReadWriteOnce'] get automatically split by openshift_facts into a dictionary like {"openshift": "hosted": {"registry": {"storage": "access": {"modes": ["ReadWriteOnce"]}}}}

@abutcher aha, got it. Yeah, I've seen this type of magic in Rails... is this some feature of Ansible or custom code we have. Where can one find more about it? Thanks!

Issue #2860, opened 2016-11-25
Advanced install fails on firewalld.yml


deanpeterson:
I pulled down the latest origin openshift-ansible playbooks and ran the Advanced install. I get the following error:

```
ansible-playbook ~/openshift-ansible/playbooks/byo/config.yml
statically included: /root/openshift-ansible/roles/os_firewall/tasks/firewall/firewalld.yml
ERROR! no action detected in task. This often indicates a misspelled module name, or incorrect module path.

The error appears to have been in '/root/openshift-ansible/roles/os_firewall/tasks/firewall/firewalld.yml': line 6, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: Ensure iptables services are not enabled
  ^ here
```

The error appears to have been in '/root/openshift-ansible/roles/os_firewall/tasks/firewall/firewalld.yml': line 6, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

- name: Ensure iptables services are not enabled
  ^ here

##### Master
[if you're operating from a git clone provide the output of `git describe`]
openshift-ansible-3.4.17-1-158-g90fb07f

[Your version of ansible, `ansible --version`]
ansible 2.1.2.0

##### Steps To Reproduce
1. # cd ~
    # git clone https://github.com/openshift/openshift-ansible
    # cd openshift-ansible
2. ansible-playbook ~/openshift-ansible/playbooks/byo/config.yml

##### Current Result
```
ERROR! no action detected in task. This often indicates a misspelled module name, or incorrect module path.

The error appears to have been in '/root/openshift-ansible/roles/os_firewall/tasks/firewall/firewalld.yml': line 6, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: Ensure iptables services are not enabled
  ^ here


The error appears to have been in '/root/openshift-ansible/roles/os_firewall/tasks/firewall/firewalld.yml': line 6, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: Ensure iptables services are not enabled
  ^ here

```
##### Expected Result
Successful installation
##### Additional Information
[The exact command you ran]
RHEL 7.2

luizvasconceloss:
I have the same problem.

ansible --version
ansible 2.1.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides

[...]

PR #2865, opened 2016-11-28
Updating docs for Ansible 2.2 requirements

mtnbikenc:
Ansible 2.2 is required due to the use of the Ansible `systemd` module.

#2860


mtnbikenc:
aos-ci-test

PR #2868, opened 2016-11-28
Refactored to use Ansible systemd module

mtnbikenc:
* Ansible systemd module used in place of service module
* Refactored command tasks which are no longer necessary
* Applying rules from openshift-ansible Best Practices Guide


mtnbikenc:
aos-ci-test

PR #2870, opened 2016-11-28
Gracefully handle OpenSSL module absence

### LINKED ISSUE #2869: error because pyOpenSSL is not installed
I get an error because pyOpenSSL is not installed when I run this from an f25 system (the host where ansible is installed). 

```
TASK [Verify Ansible version is greater than or equal to 2.1.0.0] **************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "The conditional check 'not ansible_version.full | version_compare('2.1.0.0', 'ge')' failed. The error was: No module
 named OpenSSL.crypto\n\nThe error appears to have been in '/sharedfolder/code/github.com/openshift/openshift-ansible/playbooks/common/openshift-cluster/verify_ansible_vers
ion.yml': line 8, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  tasks:\n  - name: Verify Ansi
ble version is greater than or equal to 2.1.0.0\n    ^ here\n"}
```

After I install pyOpenSSL everything works fine. Maybe that is a dep that should get installed by the rpm. Maybe not. If not, we should give a better error message to the users of openshift-ansible. 

##### Version
Working from 90fb07f and using `ansible-2.2.0.0-3.fc25.noarch.rpm` on fedora 25. No pyOpenSSL is installed.

mscherer:
Should fix #2869

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -10,7 +10,14 @@ from collections import Mapping
 from distutils.util import strtobool
 from distutils.version import LooseVersion
 from operator import itemgetter
-import OpenSSL.crypto
+
+HAS_OPENSSL=False
+try:
+    import OpenSSL.crypto
+    HAS_OPENSSL=True
+except ImportError:
+    pass
+
 import os
 import pdb
 import pkg_resources
@@ -515,6 +522,9 @@ class FilterModule(object):
 
         if not isinstance(internal_hostnames, list):
             raise errors.AnsibleFilterError("|failed expects internal_hostnames is list")
+
+        if not HAS_OPENSSL:
+            raise errors.AnsibleFilterError("|missing OpenSSL python bindings")
 
         for certificate in certificates:
             if 'names' in certificate.keys():


mscherer:
aos-ci-test

dustymabe:
This is not working for me. I don't have the module and it is not complaining.

mscherer:
It should only complain when we use oo_parse_named_certificates, which is done around openshift master deployment. 

How did you tested, with a full deployment ?

detiber:
aos-ci-test

dustymabe:
@detiber I reported earlier that the code in this PR actually didn't result in an "error" being shown to the user, can someone else confirm the behavior of this PR when the pyOpenSSL rpm does not exist on your system?

dustymabe:
I guess the reason I'm not seeing an "error" message any longer is because when I run the code I don't ever hit the `oo_parse_named_certificates()` function which is where the exception would get raised. I don't know why my code path would not hit this function. I guess I don't have any issue with this PR.

PR #2873, opened 2016-11-28
Move Metrics to ansible from deployer

jcantrill:
@bbguimaraes I found it easier to copy in all your work based on master instead of trying to rebase.  Please review to see if it reflects your changes


jcantrill:
cc @ewolinetz @sdodson @detiber

bbguimaraes:
Looks good, just a few comments:

- This is WIP, currently supporting green-field deployment and cleanup.

- The files `pvc.j2`, `rolebinding.j2`, `serviceaccount.j2` and `service.j2` on
  the `templates` directory are the same used in logging (#2640).

  We were using symlinks for that, but there's probably a better way to
  handle this situation, specially because these templates could potentially
  be used anywhere these objects need to be created.

- The files `handlers/main.yaml`, `tasks/main.yaml.org`, `tasks/install.yml`,
  and all but the last variable in `vars/main.yaml` are left-overs from the
  previous implementation and should be removed.

jcantrill:
@bbguimaraes This PR copies the referenced templates into this PR instead of using links.  I would expect us to reconcile that in the future by having a single source.  I think there are other places where links are used so it is appropriate.

Prior to removing all the files you reference, is there a need for us to keep the handler?  There looks to be some logic where I presume its validating the the cluster is up and the API server is reachable.  It might be useful to reuse.

ewolinetz:
per the discussion with jason and scott, we should update the role name to not be the same as the current role so that this can be merged in, not remove the ability to use the current role and give QE time to switch over to using the new role.

jcantrill:
@bbguimaraes updated based on your comments.  please verify i properly captured the items you identified

bbguimaraes:
@jcantrill: LGTM.

roles/openshift_metrics/templates/heapster.j2
```diff
@@ -0,0 +1,98 @@
...
+kind: "ReplicationController"
+metadata:
+  name: heapster
+  labels:
+    metrics-infra: heapster
+    name: heapster
+spec:
+  selector:
+    name: heapster
+  replicas: 0
```
bbguimaraes: This (and the same for the corresponding lines in `heapster.j2` and `hawkular_metrics_rc.j2`) creates a problem when the role is applied again.  When we `oc apply` this file again, the rc's will be scaled to 0 and then back (in `scale.yaml`), restarting the pods, even if nothing changes.

jcantrill: @ewolinetz  thoughts?  do we see this as an issue if we are scaling down and then scaling up anyway?

jcantrill:
currently blocked by #2881

roles/openshift_metrics/defaults/main.yaml
```diff
@@ -0,0 +1,48 @@
...
+openshift_metrics_cassandra_pv_size: 10Gi
+openshift_metrics_cassandra_limits_memory: 2G
+openshift_metrics_cassandra_limits_cpu: null
+openshift_metrics_cassandra_requests_memory: 1G
+openshift_metrics_cassandra_requests_cpu: null
+
+openshift_metrics_heapster_standalone: False
+openshift_metrics_heapster_limits_memory: 3.75G
+openshift_metrics_heapster_limits_cpu: null
+openshift_metrics_heapster_requests_memory: 0.9375G
```
detiber: I'm sure there's a valid reason for requests_memory to be 0.9375G, but man does it just look odd, instead of 1G or even 0.9G

roles/openshift_metrics/defaults/main.yaml
```diff
@@ -0,0 +1,48 @@
...
+openshift_metrics_heapster_requests_cpu: null
```
detiber: null looks a bit weird here as well, would it be possible to use `None` here instead and translate `None` to `null` where it is used?

roles/openshift_metrics/meta/main.yaml
```diff
@@ -1,17 +1,2 @@
----
```
detiber: There should be more metadata other than just the dependencies here. Also, the `---` header should be retained.

roles/openshift_metrics/tasks/generate_certificates.yaml
```diff
@@ -0,0 +1,26 @@
+---
```
detiber: @abutcher any suggestions on using the openshift CA with OpenSSL here instead of a custom CA cert?

roles/openshift_metrics/tasks/main.yaml
```diff
@@ -1,85 +1,24 @@
...
-  command: mktemp -d /tmp/openshift-ansible-XXXXXX
+- name: Create temp directory for doing work in
+  command: mktemp -td openshift-metrics-ansible-XXXXXX
   register: mktemp
   changed_when: False
 
-- name: Record kubeconfig tmp dir
-  set_fact:
-    openshift_metrics_kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
+- debug: msg="Created temp dir {{mktemp.stdout}}"
```
detiber: This debug statement should be removed prior to merging.

roles/openshift_metrics/tasks/main.yaml
```diff
@@ -1,85 +1,24 @@
...
-    --namespace openshift-infra
-    get pods -l {{ item }} | grep -q Running
-  register: metrics_pods_status
-  with_items:
-    - metrics-infra=hawkular-metrics
-    - metrics-infra=heapster
-    - metrics-infra=hawkular-cassandra
-  failed_when: false
-  changed_when: false
+- include: "{{role_path}}/tasks/install_metrics.yaml"
```
detiber: `include install_metrics.yaml` should work here without having to use the `role_path` var.

roles/openshift_metrics/tasks/scale.yaml
```diff
@@ -0,0 +1,27 @@
+---
+- shell: >
+    {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig get {{object}}
+    --template='{{ '{{.spec.replicas}}' }}' -n {{openshift_metrics_project}}
```
detiber: You should be able to use `command` here instead of `shell`

roles/openshift_metrics/tasks/scale.yaml
```diff
@@ -0,0 +1,27 @@
...
+  register: replica_count
+  failed_when: "replica_count.rc == 1 and 'exists' not in replica_count.stderr"
+  when: not ansible_check_mode
+
+- shell: >
```
detiber: You should be able to use `command` here instead of `shell`

roles/openshift_metrics/tasks/scale.yaml
```diff
@@ -0,0 +1,27 @@
...
+    --replicas={{desired}} -n {{openshift_metrics_project}}
+  register: scale_result
+  failed_when: scale_result.rc == 1 and 'exists' not in scale_result.stderr
+  when:
+  - replica_count.stdout != desired
+  - not ansible_check_mode
+
+- name: Waiting for {{object}} to scale to {{desired}}
+  shell: >
+    {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig describe {{object}} -n {{openshift_metrics_project}} | awk -v statusrx='Pods Status:' '$0 ~ statusrx {print $3}'
```
detiber: openshift_metrics_project should probably be passed through the `quote` filter here:
```
{{ openshift_metrics_project | quote }}
```

roles/openshift_metrics/tasks/setup_certificate.yaml
```diff
@@ -0,0 +1,53 @@
...
+  shell: >
+    p=$(< {{ openshift_metrics_certs_dir }}/{{ component }}-keystore.pwd)
+    &&
+    keytool -v -importkeystore
+    -srckeystore '{{ openshift_metrics_certs_dir }}/{{ component }}.pkcs12'
+    -srcstoretype PKCS12
+    -destkeystore '{{ openshift_metrics_certs_dir }}/{{ component }}.keystore'
+    -deststoretype JKS
+    -deststorepass "$p"
+    -srcstorepass "$p"
```
detiber: component should probably be passed through the quote filter here...  If openshift_metrics_certs_dir is user settable, it should probably be as well.

roles/openshift_metrics/tasks/start_metrics.yaml
```diff
@@ -0,0 +1,52 @@
+---
+- shell: >
```
detiber: command can be used here instead of shell.

roles/openshift_metrics/tasks/start_metrics.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+- name: Start Hawkular Cassandra
+  include: scale.yaml
+  vars:
+    desired: 1
+  with_items: "{{metrics_cassandra_rc.stdout_lines}}"
+  loop_control:
+    loop_var: object
+
+- shell: >
```
detiber: command can be used instead of shell.

roles/openshift_metrics/tasks/start_metrics.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+- name: Start Hawkular Metrics
+  include: scale.yaml
+  vars:
+    desired: "{{openshift_metrics_hawkular_replicas}}"
+  with_items: "{{metrics_metrics_rc.stdout_lines}}"
+  loop_control:
+    loop_var: object
+
+- shell: >
```
detiber: s/shell/command/

roles/openshift_metrics/tasks/stop_metrics.yaml
```diff
@@ -0,0 +1,56 @@
+---
+- shell: >
```
detiber: s/shell/command/

roles/openshift_metrics/tasks/stop_metrics.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+- name: Stop Heapster
+  include: scale.yaml
+  vars:
+    desired: 0
+  with_items: "{{metrics_heapster_rc.stdout_lines}}"
+  loop_control:
+    loop_var: object
+
+- shell: >
```
detiber: s/shell/command/

roles/openshift_metrics/tasks/stop_metrics.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+- name: Stop Hawkular Metrics
+  include: scale.yaml
+  vars:
+    desired: 0
+  with_items: "{{metrics_hawkular_rc.stdout_lines}}"
+  loop_control:
+    loop_var: object
+
+- shell: >
```
detiber: s/shell/command/

roles/openshift_metrics/templates/hawkular_cassandra_rc.j2
```diff
@@ -0,0 +1,125 @@
...
+          mountPath: "/cassandra_data"
+        - name: hawkular-cassandra-secrets
+          mountPath: "/secret"
+{% if ((openshift_metrics_cassandra_limits_cpu is defined and openshift_metrics_cassandra_limits_cpu is not none) 
+   or (openshift_metrics_cassandra_limits_memory is defined and openshift_metrics_cassandra_limits_memory is not none)
+   or (openshift_metrics_cassandra_requests_cpu is defined and openshift_metrics_cassandra_requests_cpu is not none)
+   or (openshift_metrics_cassandra_requests_memory is defined and openshift_metrics_cassandra_requests_memory is not none)) 
+%}
+        resources:
+{% if (openshift_metrics_cassandra_limits_cpu is not none
```
detiber: One trick I've seen for helping with these nested ifs is to add indentation inside of the template start:
```
{%   if (openshift_metrics_cassandra_limits_cpu ...
```

detiber:
@kwoodson ptal, this is an interim step until we get some core modules in place to make this work cleaner. The openshift_metrics role is the new work for this PR, with the openshift_hosted_metrics role being the previous approach that is being phased out with this work.

roles/openshift_metrics/defaults/main.yaml
jcantrill: Straight copy from the deployer.  Not sure the reason.  I'll reach out the perf team to see if these defaults should b adjusted based on their lresults.

roles/openshift_metrics/meta/main.yaml
jcantrill: added

roles/openshift_metrics/tasks/main.yaml
jcantrill: removed

roles/openshift_metrics/tasks/main.yaml
jcantrill: removed

roles/openshift_metrics/tasks/scale.yaml
jcantrill: changed

roles/openshift_metrics/tasks/scale.yaml
jcantrill: changed

roles/openshift_metrics/tasks/scale.yaml
jcantrill: added...but what is the justification?

roles/openshift_metrics/tasks/setup_certificate.yaml
jcantrill: updated in all places where shell was used

roles/openshift_metrics/tasks/start_metrics.yaml
jcantrill: updated all uses of shell to command

roles/openshift_metrics/tasks/start_metrics.yaml
jcantrill: ack

roles/openshift_metrics/tasks/stop_metrics.yaml
jcantrill: updated all uses of shell to command

roles/openshift_metrics/tasks/stop_metrics.yaml
jcantrill: updated all uses of shell to command

roles/openshift_metrics/tasks/stop_metrics.yaml
jcantrill: updated all uses of shell to command

jcantrill:
@detiber updated based on review comments.  also incorporated some of the changes from logging

detiber:
@jcantrill looks like the PR that fixed yamllint testing has raised some errors: https://travis-ci.org/openshift/openshift-ansible/jobs/191116469

roles/openshift_metrics/tasks/install_heapster.yaml
```diff
@@ -0,0 +1,13 @@
+---
+- shell: >
+    {{ openshift.common.client_binary }} -n {{ openshift_metrics_project | quote }}
+    --config={{ mktemp.stdout }}/admin.kubeconfig
+    get rc heapster -o jsonpath='{.spec.replicas}' || echo 0
```
detiber: using `|| echo 0` here seems a bit odd, instead I would set `failed_when: false` here and then do the following below:
```
  replica_count: "{{ heapster_replica_count.stdout | default(0) }}"
```

roles/openshift_metrics/tasks/setup_certificate.yaml
```diff
@@ -0,0 +1,58 @@
...
+    --signer-key='{{ openshift_metrics_certs_dir }}/ca.key'
+    --signer-serial='{{ openshift_metrics_certs_dir }}/ca.serial.txt'
+  when: not '{{ openshift_metrics_certs_dir }}/{{ component }}.key'|exists
+
+- name: generate {{ component }} certificate
+  shell: >
+    cat
+    '{{ openshift_metrics_certs_dir | quote }}/{{ component|quote }}.key'
+    '{{ openshift_metrics_certs_dir | quote }}/{{ component|quote }}.crt'
+    > '{{ openshift_metrics_certs_dir | quote }}/{{ component|quote }}.pem'
```
detiber: I'm not sure if it's any better, but instead of using shell/cat, you could use slurp and copy to create the pem from the contents of the key and cert.

roles/openshift_metrics/tasks/setup_certificate.yaml
```diff
@@ -0,0 +1,58 @@
...
+  when: not '{{ openshift_metrics_certs_dir }}/{{ component }}.pem'|exists
+
+- name: generate random password for the {{ component }} keystore
+  shell: >
+    tr -dc _A-Z-a-z-0-9 < /dev/urandom | head -c15
+    > '{{ openshift_metrics_certs_dir | quote }}/{{ component|quote }}-keystore.pwd'
```
detiber: @ewolinetz would your ansible native solution for generating a random value work here?

roles/openshift_metrics/tasks/start_metrics.yaml
```diff
@@ -0,0 +1,54 @@
...
+    -l metrics-infra=hawkular-cassandra 
+    -o name 
+    -n {{openshift_metrics_project}}
+  register: metrics_cassandra_rc
+  changed_when: no
+
+- name: Start Hawkular Cassandra
+  include: scale.yaml
+  vars:
+    desired: 1
```
detiber: should this be hardcoded to 1 here? Are we expecting petsets to be usable in the future for Cassandra?

roles/openshift_metrics/tasks/start_metrics.yaml
```diff
@@ -0,0 +1,54 @@
...
+    -l metrics-infra=heapster
+    -o name 
+    -n {{openshift_metrics_project}}
+  register: metrics_heapster_rc
+  check_mode: no
+
+- name: Start Heapster
+  include: scale.yaml
+  vars:
+    desired: 1
```
detiber: Is 1 supposed to be  hardcoded here?

roles/openshift_metrics/tasks/install_heapster.yaml
jcantrill: Updated where I could.  I have one case where register produces a list.  A following task pulls values by index from the list so this method doesnt quite work so I had to leave the echo

roles/openshift_metrics/tasks/start_metrics.yaml
jcantrill: we expect statefulesets (petsets) to be the path forward which will work like a traditional replicaset but for now we are limited to a single RC per Cassandra Node.  I don't see any other way to control this for now.  It just means in future when statefulsets come into play we will have to modify the ansible tasks.  I would expect this anyway since we would move from RC->SS

roles/openshift_metrics/tasks/setup_certificate.yaml
jcantrill: updated with slurp and copy

roles/openshift_metrics/tasks/generate_certificates.yaml
abutcher: None yet but hopefully something that could be done soon.

roles/openshift_metrics/tasks/start_metrics.yaml
jcantrill: This is supposed to be one since heapster can not scale beyond 1 per cluster.  This will be obsoleted in future metrics work when heapster is deprecated.

roles/openshift_metrics/tasks/setup_certificate.yaml
jcantrill: updated to use our random_word filter.  Moved a copy of into repo filters as oo_random_word which defaults the set of chars

jcantrill:
@detiber updated again.  im inclined to squash... but suggestions were good and i have it back to a working state

roles/openshift_metrics/defaults/main.yaml
jcantrill: To confirm our IRC discussion, I worked for a while to remove nulls and wasnt fully able to resolve issues with evaluation of None as a string in the jinja template.

jcantrill:
@detiber also just made changes to properly report changed when resource are updated(or not)

jcantrill:
@sdodson @detiber rebased

detiber:
@jcantrill looks like travis tests are failing with some issues related to some whitespace nits.

PR #2874, opened 2016-11-28
etcd upgrade playbook is not currently applicable to embedded etcd in…

sdodson:
…stalls

Fixes Bug 1395945


sdodson:
Need to make sure that when or if we automatically run the etcd upgrade playbooks as part of an upgrade we forego upgrading etcd for embedded environments until we add the migration work.

sdodson:
@dgoodwin PTAL

PR #2875, opened 2016-11-28
Fix invalid embedded etcd fact in etcd upgrade playbook.

sdodson:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1398549

Was getting a different failure here complaining that openshift was not
in the facts, as we had not loaded facts for the first master during
playbook run. However this check was used recently in
upgrade_control_plane and should be more reliable.


sdodson:
backports #2872

PR #2880, opened 2016-11-29
Remove duplicate when key

mtnbikenc:
Minor fix to remove a duplcate `when` key in the docker_upgrade.yml playbook.

cc: @dgoodwin


mtnbikenc:
aos-ci-test

PR #2882, opened 2016-11-29
allow 'latest' origin_image_tag

sjenning:
Right now, we expose the `origin_image_tag` variable to users but normalize it, assuming it should begin with the letter 'v'.  However, in my use case, I'm wanting to pull the `latest` tag for the origin images.  This is currently not possible, even though the variable that should allow it is exposed.

This PR creates an exception for `latest` and doesn't prepend a 'v' to the tag in that case.

@dgoodwin @sdodson


detiber:
aos-ci-test

PR #2888, opened 2016-11-30
openshift_node_dnsmasq - Remove strict-order option from dnsmasq

sdodson:
strict-order forces dnsmasq to iterate through nameservers in order. If one of
the nameservers is down this will slow things down while dnsmasq waits for a
timeout. Also, this option prevents dnsmasq from querying other nameservers if
the first one returns a negative result. While I think it's odd to have a
nameserver that returns negative results for a query that another returns
positive results for this does seem to fix the issue in testing.

Fixes Bug 1399577

### DIFF
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -588,7 +588,8 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # be used with 1.0 and 3.0.
 #openshift_use_dnsmasq=False
 # Define an additional dnsmasq.conf file to deploy to /etc/dnsmasq.d/openshift-ansible.conf
-# This is useful for POC environments where DNS may not actually be available yet.
+# This is useful for POC environments where DNS may not actually be available yet or to set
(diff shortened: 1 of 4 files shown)


sdodson:
@detiber mind taking a look at this change, we've relatively recently made a change to make dnsmasq the only resolver and that change along with `strict-order` being in place seems to break some configurations where one of the nameservers returns no answers but another does. Also, with `strict-order` dnsmasq will iterate through nameservers in order even if it knows that one of them is down which adds potential for timeouts and queries to back up. This particular problem was reported in https://bugzilla.redhat.com/show_bug.cgi?id=1399756 which i've marked as a dupe

sdodson:
aos-ci-test

sdodson:
That can always be included in the contents of `openshift_node_dnsmasq_additional_config_file`

detiber:
Docs around how to do it using openshift_node_dnsmasq_additional_config_file should be sufficient.

sdodson:
Documented that the additional config can be used to provide options to alter dnsmasq configuration.

Issue #2588, opened 2016-10-12
multi-master setup : metricspublicURL set only in first master


jfcoz:
When using byo playbook to install a multi master cluster, with metrics component, the metricspublicURL is set only on the first master master-config.yaml
##### Version

atomic-openshift-utils-3.3.28-1.git.0.762256b.el7.noarch
openshift-ansible-3.3.28-1.git.0.762256b.el7.noarch
##### Steps To Reproduce

run playbook
`- include: /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml`
ansible-hosts at bottom
##### Current Result

The metricspublicURL is updated only in the first master master-config.yaml : on the web ui, metrics display only for this master.
##### Expected Result

The metricspublicURL should be updated in all masters master-config.yaml .
##### Additional Information

Inventory file :

```
[all]
master-01-priv.example.com
master-02-priv.example.com
master-03-priv.example.com
node-01-priv.example.com
node-02-priv.example.com
node-03-priv.example.com
etcd-01-priv.example.com
etcd-02-priv.example.com
etcd-03-priv.example.com
nfs-priv.example.com

[OSEv3:children]
masters
nodes
etcd
nfs

[OSEv3:vars]
ansible_ssh_user=ec2-user
ansible_become=yes

debug_level=2

product_type=openshift
deployment_type=openshift-enterprise
openshift_deployment_type=openshift-enterprise
openshift_release=v3.3

openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 'challenge': 'true', 'kind': 'HTPasswdPasswordIdentityProvider', 'filename': '/etc/origin/master/htpasswd'}]
openshift_master_htpasswd_users={'admin': 'changed'}

osm_default_node_selector='region=primary'

osm_default_subdomain=apps.example.com

openshift_master_cluster_hostname=master.example.com
openshift_master_cluster_public_hostname=master.example.com
openshift_master_cluster_method=native
openshift_master_overwrite_named_certificates=true
os_sdn_network_plugin_name=redhat/openshift-ovs-multitenant

openshift_hosted_router_selector='region=infra'

openshift_hosted_registry_selector='region=infra'

openshift_hosted_registry_storage_kind=nfs
openshift_hosted_registry_storage_access_modes=['ReadWriteMany']
openshift_hosted_registry_storage_host=nfs-priv.example.com
openshift_hosted_registry_storage_nfs_options='*(rw,root_squash)'
[...]

PR #2891, opened 2016-11-30
Define master config dir in metrics role vars.

dgoodwin:
Reported as a problem in github issue #2588. Ensure the common var is
defined, the metrics role has a dependency on openshift_facts so this
should be safe and correct the problem.

### DIFF
--- a/roles/openshift_metrics/vars/main.yaml
+++ b/roles/openshift_metrics/vars/main.yaml
@@ -18,3 +18,4 @@ hawkular_persistence: "{% if openshift.hosted.metrics.storage.kind != None %}tru
 hawkular_type: "{{ 'origin' if deployment_type == 'origin' else 'enterprise' }}"
 
 metrics_upgrade: openshift.hosted.metrics.upgrade | default(False)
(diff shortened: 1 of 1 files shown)


roles/openshift_metrics/vars/main.yaml
```diff
@@ -18,3 +18,4 @@ hawkular_persistence: "{% if openshift.hosted.metrics.storage.kind != None %}tru
 hawkular_type: "{{ 'origin' if deployment_type == 'origin' else 'enterprise' }}"
 
 metrics_upgrade: openshift.hosted.metrics.upgrade | default(False)
+openshift_master_config_dir: "{{ openshift.common.config_base }}/master"
```
detiber: I wonder if this is because I moved this definition to `openshift_master_facts` from `openshift_master`?

dgoodwin: Not sure, I don't think either are required by openshift_metrics, and if something is defined in a roles vars.yml, am I correct that it's only valid there within that role and undefined elsewhere?

In general it looks like roles that need this define it themselves.

detiber:
Closing due to age and refactored metrics implementation

PR #2902, opened 2016-12-01
Fix metricsPublicURL only being set correctly on first master.

### LINKED ISSUE #2778: openshift_host_metrics_public_url is not set correctly in master-config.yaml
[provide a description of the issue]
On a multi master, 3, installation it was observed on 2 of the masters have the setting metricsPublicURL correctly set.
The ansible setting in the ansible hosts file is
   - openshift_master_metrics_public_url=https://hawkular-metrics.{{ mysubdomain }}/hawkular/metrics

the resulted in one of the masters this resulted in the following config set
 - metricsPublicURL: https://hawkular-metrics.apps.{/hawkular/metrics

While the other two were correctly set with the place holder been used {{ mysubdomain }} 
##### Version
[if you're operating from a git clone provide the output of `git describe`]
HEAD detached at openshift-ansible-3.3.38-1
[if you're running from playbooks installed via RPM or atomic-openshift-utils `rpm -q atomic-openshift-utils openshift-ansible`]
[Your version of ansible, `ansible --version`]

##### Steps To Reproduce
1. [step 1]
Create a multi master environment 
2. [step 2]
Set the metricsPublicUrl to use a variable 

##### Current Result
1 or more of the masters will have a incorrect setting 

##### Expected Result
All the masters have the same config 

##### Additional Information
[The exact command you ran]
[Your operating system and version, ie: RHEL 7.2, Fedora 23]
2.1.2.0
[Your inventory file]
[visit https://docs.openshift.org/latest/welcome/index.html]

dgoodwin:
Problem was caused by facts not being set for that master. To fix this
patch cleans up the calculation of metricsPublicURL in general. Because
this value is used in openshift_master to template into the master
config file, we now define these facts more clearly in
openshift_master_facts, and add a dependency on this to
openshift_metrics.

The calculation of default sub-domain is also changed to remove it from
system facts (as neither of these are facts about the system) and
instead use plain variables.

Fixes #2778.

### DIFF
--- a/roles/openshift_master/templates/master.yaml.v1.j2
+++ b/roles/openshift_master/templates/master.yaml.v1.j2
@@ -15,8 +15,8 @@ assetConfig:
 {% if 'logging_public_url' in openshift.master %}
   loggingPublicURL: {{ openshift.master.logging_public_url }}
 {% endif %}
-{% if 'metrics_public_url' in openshift.master %}
-  metricsPublicURL: {{ openshift.master.metrics_public_url }}
+{% if openshift_hosted_metrics_deploy | default(false) | bool %}
+  metricsPublicURL: {{ openshift_hosted_metrics_public_url }}
 {% endif %}
 {% if 'extension_scripts' in openshift.master %}
   extensionScripts: {{ openshift.master.extension_scripts | to_padded_yaml(1, 2) }}
@@ -202,7 +202,7 @@ projectConfig:
     mcsLabelsPerProject: {{ openshift.master.mcs_labels_per_project }}
     uidAllocatorRange: "{{ openshift.master.uid_allocator_range  }}"
 routingConfig:
-  subdomain:  "{{ openshift.master.default_subdomain | default("") }}"
+  subdomain:  "{{ openshift_master_default_subdomain | default("") }}"
 serviceAccountConfig:
   limitSecretReferences: false
   managedNames:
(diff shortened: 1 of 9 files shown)


dgoodwin:
cc @sdodson @detiber

roles/openshift_master/templates/master.yaml.v1.j2
```diff
@@ -15,8 +15,8 @@ assetConfig:
 {% if 'logging_public_url' in openshift.master %}
   loggingPublicURL: {{ openshift.master.logging_public_url }}
 {% endif %}
-{% if 'metrics_public_url' in openshift.master %}
-  metricsPublicURL: {{ openshift.master.metrics_public_url }}
+{% if openshift_hosted_metrics_deploy | default(false) | bool %}
```
sdodson: In lgith of the other pr maybe this instead?
`{% if openshift_hosted_metrics_public_url | default("") != "" %}`

dgoodwin: As it stands here it's always going to be set after openshift_master_facts role runs, which loses us the ability to tell if it was set by user or defaulted. I did post a suggestion for the other PR though that would build on top of this to let us detect here and write this if deploy = true, *or* this var was set in inventory. Proposing this as a followup.

sdodson: ok, lets take this as is and we can follow up later.

sdodson:
aos-ci-test

PR #2909, opened 2016-12-02
Enable firewalld by default

mtnbikenc:


mtnbikenc:
aos-ci-test

roles/docker/templates/custom.conf.j2
```diff
@@ -1,5 +0,0 @@
-# {{ ansible_managed }}
-
-[Unit]
-Requires=iptables.service
-After=iptables.service
```
detiber: It looks like the firewalld unit file sets Before=network.target, which ensures the correct ordering. I suspect this workaround would still be needed for the case where os_firewall_use_firewalld is explicitly set to False.

mtnbikenc:
aos-ci-test

mtnbikenc:
firewalld is currently not supported on Atomic Host
https://bugzilla.redhat.com/show_bug.cgi?id=1403331

mtnbikenc:
aos-ci-test

mtnbikenc:
@detiber I've added a task to override the use of firewalld if running on Atomic Host.  README updated as well.

roles/os_firewall/defaults/main.yml
```diff
@@ -1,9 +1,5 @@
 ---
 os_firewall_enabled: True
-# TODO: Upstream kubernetes only supports iptables currently
-# TODO: it might be possible to still use firewalld if we wire up the created
-# chains with the public zone (or the zone associated with the correct
-# interfaces)
-os_firewall_use_firewalld: False
+os_firewall_use_firewalld: True
```
detiber: Why not set this to `"{{ False if openshift.common.is_atomic | bool else True }}"

Then, you could replace the set_fact in tasks/main.yml with:
```
- assert: 
    that:
    - "{{ openshift.common.is_atomic | bool is true and os_firewall_use_firewalld | bool is true }}"
```
instead of using set_fact

roles/docker/tasks/main.yml
```diff
@@ -47,12 +47,14 @@
...
     path: "{{ docker_systemd_dir }}"
     state: directory
+  when: not os_firewall_use_firewalld | default(True) | bool
 
 # Extend the default Docker service unit file
 - name: Configure Docker service unit file
   template:
     dest: "{{ docker_systemd_dir }}/custom.conf"
     src: custom.conf.j2
+  when: not os_firewall_use_firewalld | default(True) | bool
```
detiber: Lets get these two tasks moved into a block and apply the conditional to the block instead of duplicating the when clause.

mtnbikenc:
@detiber I've update the PR with your suggestions.

detiber:
aos-ci-test

PR #2914, opened 2016-12-05
Conditionalize master config update for admission_plugin_config.

abutcher:
Fixes https://github.com/openshift/origin/issues/12136

### DIFF
--- a/roles/openshift_master_facts/defaults/main.yml
+++ b/roles/openshift_master_facts/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+openshift_master_default_subdomain: "{{ lookup('oo_option', 'openshift_master_default_subdomain') | default(None, true) }}"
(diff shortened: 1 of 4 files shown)


roles/openshift_master_facts/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+openshift_master_default_subdomain: "{{ lookup('oo_option', 'openshift_master_default_subdomain') | default(None, true) }}"
```
abutcher: Required after pulling in https://github.com/openshift/openshift-ansible/pull/2908 to have `openshift_master_default_subdomain` set in upgrade (for `openshift_master_facts` role). We're [setting this in a few places as a workaround](https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/redeploy-certificates.yml#L55-L62) so I've added to defaults so that it's always there.

sdodson:
aos-ci-test

roles/openshift_master_facts/defaults/main.yml
sdodson: Do we need to pull only this change into release-1.3? I think the other changes in this PR aren't relevant to release-1.3.

sdodson: nm, i think not, right?

PR #2915, opened 2016-12-05
node_dnsmasq - restart dnsmasq if it's not currently running

sdodson:
Make sure that if our dispatcher script runs before dnsmasq has started that we start it otherwise we'll fail to update /etc/resolv.conf because we check for dnsmasq to have started.

### DIFF
--- a/roles/openshift_node_dnsmasq/files/networkmanager/99-origin-dns.sh
+++ b/roles/openshift_node_dnsmasq/files/networkmanager/99-origin-dns.sh
@@ -80,6 +80,10 @@ EOF
       NEEDS_RESTART=1
     fi
 
+    if ! `systemctl -q is-active dnsmasq.service`; then
+      NEEDS_RESTART=1
+    fi
+
     ######################################################################
     if [ "${NEEDS_RESTART}" -eq "1" ]; then
       systemctl restart dnsmasq


sdodson:
aos-ci-test

sdodson:
@tbielawa ptal

roles/openshift_node_dnsmasq/files/networkmanager/99-origin-dns.sh
```diff
@@ -80,6 +80,10 @@ EOF
       NEEDS_RESTART=1
     fi
 
+    if ! `systemctl -q is-active dnsmasq.service`; then
```
detiber: Does this work consistently between all "supported" releases? RHEL 7.1, Fedora 24/25?

sdodson: Looks like it was added in systemd v198 prior to v209 (RHEL 7.1) and has defined rules regarding exit codes; 0 only if the service is active.

sdodson:
CI job failed in post job cleanup tasks. Also, need to figure out why the bot isn't posting useful information any longer.

detiber:
@sdodson +1, I wonder if we can use this as opposed to the dbus stuff in openshift-facts.

PR #2920, opened 2016-12-05
Scheduler var fix

detiber:


detiber:
@sdodson @abutcher ptal, previously, I re-used the same vars and it was causing issues in differentiating where the values where coming from (affecting upgrade), this separates it out and makes it more explicit.

abutcher:
aos-ci-test

sdodson:
LGTM

PR #2923, opened 2016-12-05
Backport 1.2: scheduler var fix

abutcher:


roles/openshift_master_facts/tasks/main.yml
```diff
@@ -80,3 +80,31 @@
       api_env_vars: "{{ openshift_master_api_env_vars | default(None) }}"
       controllers_env_vars: "{{ openshift_master_controllers_env_vars | default(None) }}"
       audit_config: "{{ openshift_master_audit_config | default(None) }}"
+      scheduler_args: "{{ openshift_master_scheduler_args | default(None) }}"
```
detiber: I'm not sure if scheduler_args is supported in 1.2/3.2

PR #2928, opened 2016-12-06
Fix etcd upgrades to etcd 3.x

sdodson:
- Fix variable trampling between containerized and rpm based upgrades
 - Switch back to using etcd image, a future update will push etcd 3.x into registry.access.redhat.com/rhel7/etcd


sdodson:
aos-ci-test

PR #2934, opened 2016-12-07
etcd_upgrade: Simplify package installation

sdodson:
Further simplification, requires rhel 7.3.2 or fedora packaging of etcd-3.x rather than RHEL 7.3.0-7.3.1 packaging as etcd3-3.x

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/etcd/rhel_tasks.yml
+++ b/playbooks/common/openshift-cluster/upgrades/etcd/rhel_tasks.yml
@@ -2,13 +2,10 @@
 - name: Verify cluster is healthy pre-upgrade
   command: "etcdctl --cert-file /etc/etcd/peer.crt --key-file /etc/etcd/peer.key --ca-file /etc/etcd/ca.crt -C https://{{ openshift.common.hostname }}:2379 cluster-health"
 
-- name: Update etcd package but exclude etcd3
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/upgrades/etcd/rhel_tasks.yml
```diff
@@ -3,12 +3,7 @@
   command: "etcdctl --cert-file /etc/etcd/peer.crt --key-file /etc/etcd/peer.key --ca-file /etc/etcd/ca.crt -C https://{{ openshift.common.hostname }}:2379 cluster-health"
 
 - name: Update etcd package but exclude etcd3
```
mtnbikenc: Maybe update the name since exclude is no long used.  Also, possibly switch to `package` module since the ``--exclude` flag is no longer necessary.

sdodson: Good idea, updated.

sdodson:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/etcd/rhel_tasks.yml
sdodson: Also switched to package module now, thanks.

sdodson:
aos-ci-test

PR #2937, opened 2016-12-07
Pre-pull master/node/ovs images during upgrade.

dgoodwin:
We did this for install but not upgrade, leading to situations where the
service restarts after upgrade could take much longer than expected as
docker pulls down the new image. Now the images are present when we
restart services and should allow them to come back online much more
quickly, equivalent to rpm service restarts.


sdodson:
aos-ci-test

PR #2938, opened 2016-12-07
Allow to set openshift_hosted_metrics_public_url without deploying me…

AlbertoPeon:
Set `assetConfig.metricsPublicURL` in the `master-config.yaml`  to the value of `openshift_hosted_metrics_public_url` even if `openshift_hosted_metrics_deploy=false`.

This is useful for us as we cannot deploy the metrics directly with Ansible but we want to manage the `master-config.yaml` config file with it

### DIFF
--- a/roles/openshift_master_facts/tasks/main.yml
+++ b/roles/openshift_master_facts/tasks/main.yml
@@ -75,7 +75,7 @@
       api_env_vars: "{{ openshift_master_api_env_vars | default(None) }}"
       controllers_env_vars: "{{ openshift_master_controllers_env_vars | default(None) }}"
       audit_config: "{{ openshift_master_audit_config | default(None) }}"
-      metrics_public_url: "{% if openshift_hosted_metrics_deploy | default(false) %}https://{{ metrics_hostname }}/hawkular/metrics{% endif %}"
+      metrics_public_url: "{{ openshift_hosted_metrics_public_url | default(None) }}"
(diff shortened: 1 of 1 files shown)


sdodson:
@dgoodwin assigning you because you recently worked on metrics public url. My only concern here is that I don't want to make setting openshift_hosted_metrics_public_url mandatory which I'm afraid this change currently does.

AlbertoPeon:
@sdodson I don't think this is making `openshift_hosted_metrics_public_url` mandatory, does it? 

In any case, I'm happy to change so it keeps being backwards compatible!

dgoodwin:
I have a pending cleanup for metrics public URL in #2902, this patch actually won't apply cleanly there as the URL moves to just being used to set g_metrics_hostname once, then we copy a recalculated value back into openshift_hosted_metrics_public_url.

For your use case we want to include the master template value *if* deploy is true, *or* the user explicitly set the public_url in inventory.

@AlbertoPeon if you'd like to take another try at this could you hold off until the other PR merges, and I think then you'd just have to change how I create g_metrics_hostname and then copy back into openshift_hosted_metrics_public_url and instead make that copy into another new variable, which would then we the one we actually use in templates. Then in the template we could tell if you actually set a public_url or not.

AlbertoPeon:
@dgoodwin Sure, I can wait and do it the way you propose

AlbertoPeon:
Closing in favor of #2973

PR #2939, opened 2016-12-07
Speed up 'make ci' and trim the output

### LINKED ISSUE #2933: make ci is too slow
Running tests locally with `make ci` is disconcertingly slow, and there seems to be ways of improving it:

For a local development flow:
- [ ] do not recreate a virtualenv in every run
- [ ] print clearly what failed instead of wall-of-text (is it possible? pylint, I'm looking at you)
- [ ] run `flake8`, `pylint`, `nosetests`, etc, even if a previous step failed (maybe)
- [ ] perhaps we should run tests first, `pylint` later, because tests take a second to run, and `pylint` takes a much longer time.

For remote CI, the current flow is possibly fine. I intend to work on it myself, adding the issue for visibility / tracking.

tbielawa:
The virtualenv is conditionally rebuilt now if the test requirements
file is modified. This will save upwards of 30+ seconds in iterative
'make ci' runs.

The pylint output is now trimmed to exclude disabled tests.

The order of the 'ci' target prerequisites has been changed to run the
fastest tests first.

Closes #2933

### DIFF
--- a/utils/Makefile
+++ b/utils/Makefile
@@ -59,13 +59,18 @@ docs: $(MANPAGES)
 viewcover:
 	xdg-open cover/index.html
 
-virtualenv:
+# Conditional virtualenv building strategy taken from this great post
+# by Marcel Hellkamp:
+# http://blog.bottlepy.org/2012/07/16/virtualenv-and-makefiles.html
+venv: oo-installenv/bin/activate
+oo-installenv/bin/activate: test-requirements.txt
 	@echo "#############################################"
 	@echo "# Creating a virtualenv"
 	@echo "#############################################"
-	virtualenv $(NAME)env
+	test -d venv || virtualenv $(NAME)env
 	. $(NAME)env/bin/activate && pip install setuptools==17.1.1
 	. $(NAME)env/bin/activate && pip install -r test-requirements.txt
+	touch $(NAME)env/bin/activate
 #       If there are any special things to install do it here
 #       . $(NAME)env/bin/activate && INSTALL STUFF
 
@@ -80,7 +85,7 @@ ci-pylint:
 	@echo "#############################################"
 	@echo "# Running PyLint Tests in virtualenv"
 	@echo "#############################################"
-	. $(NAME)env/bin/activate && python -m pylint --rcfile ../git/.pylintrc $(shell find ../ -name $(NAME)env -prune -o -name test -prune -o -name "*.py" -print)
+	. $(NAME)env/bin/activate && python -m pylint --rcfile ../git/.pylintrc $(shell find ../ -name $(NAME)env -prune -o -name test -prune -o -name "*.py" -print) 2>&1 | grep -E -v '(locally-disabled|file-ignored)'
 
 ci-list-deps:
 	@echo "#############################################"
@@ -93,7 +98,11 @@ ci-flake8:
 	@echo "# Running Flake8 Compliance Tests in virtualenv"
 	@echo "#############################################"
 	. $(NAME)env/bin/activate && flake8 --config=setup.cfg ../ --exclude="utils,../inventory"
-	. $(NAME)env/bin/activate && python setup.py flake8 
+	. $(NAME)env/bin/activate && python setup.py flake8
 
-ci: clean virtualenv ci-list-deps ci-flake8 ci-pylint ci-unittests
-	:
+ci: venv ci-list-deps ci-unittests ci-flake8 ci-pylint
+	@echo
+	@echo "##################################################################################"
+	@echo "VIEW CODE COVERAGE REPORT WITH 'xdg-open cover/index.html' or run 'make viewcover'"
+	@echo "To clean your test environment run 'make clean'"
+	@echo "Other targets you may run with 'make': 'ci-pylint', 'ci-unittests', 'ci-flake8'"


tbielawa:
aos-ci-test

tbielawa:
@detiber @rhcarvalho @smunilla  PTAL buds

tbielawa:
Derp. I accidentally included the `-U` option in the `pip install` commands. That could have created inconsistent test conditions. I've removed the option flag and force pushed a new commit.

tbielawa:
aos-ci-test

tbielawa:
Last derp. I updated the PR again to exclude `file-ignored` pylint messages as well. Now the pylint check only outputs actual problems.

tbielawa:
aos-ci-test

rhcarvalho:
> I updated the PR again to exclude file-ignored pylint messages as well. Now the pylint check only outputs actual problems.

That's perhaps the most expedient solution indeed, but IMHO filtering out `pylint` output feels like putting dust under the carpet; it's hiding the fact that we need to add way too many comments to "satisfy" the tooling of choice.

rhcarvalho:
> aos-ci-test

My doubt:
As a matter of understanding the existing processes, why do we run these tests when changing only that Makefile?

AFAIU those tests exercise calls to `ansible-playbook` and have no interaction whatsoever with make? The Makefile is used by the `default` test job (which I think runs automatically for whitelisted authors), and also in Travis (run automatically for everyone).

utils/Makefile
```diff
@@ -93,7 +98,11 @@ ci-flake8:
...
 	. $(NAME)env/bin/activate && flake8 --config=setup.cfg ../ --exclude="utils,../inventory"
-	. $(NAME)env/bin/activate && python setup.py flake8 
+	. $(NAME)env/bin/activate && python setup.py flake8
 
-ci: clean virtualenv ci-list-deps ci-flake8 ci-pylint ci-unittests
-	:
+ci: venv ci-list-deps ci-unittests ci-flake8 ci-pylint
+	@echo
+	@echo "##################################################################################"
+	@echo "VIEW CODE COVERAGE REPORT WITH 'xdg-open cover/index.html' or run 'make viewcover'"
```
rhcarvalho: This is already printed by `ci-unittests`?

I'd prefer this to be documented in CONTRIBUTING.md or similar than to be printed in every run. How often to you need to read this message and use the information to open the coverage report?

utils/Makefile
```diff
@@ -93,7 +98,11 @@ ci-flake8:
...
+	@echo "To clean your test environment run 'make clean'"
+	@echo "Other targets you may run with 'make': 'ci-pylint', 'ci-unittests', 'ci-flake8'"
```
rhcarvalho: I'd say these lines in the Travis or Jenkins output add no value as we don't pick the commands we run in those environments.

I think we want to see problems, errors, and adding boilerplate just gives us more mental work to skip that when we have a test failure and go read the logs?
I'd be really interested in streamlining the output to make it concise and straight to the point.

Information wise, if a developer was able to figure out there is a Makefile inside of `utils` and can run `make ci`, she can probably figure out the other targets as well without having them printed all the time?

tbielawa: @rhcarvalho the whole reason the 'make ci' system was added was to enable contributors to run the tests on their local machines before having to wait on jenkins results to report in. I.e., so you only have to git push commits that you know pass the basic essential tests. That's why you can run them all individually and why the information is printed out with such verbosity. To make it easier on **people** using the ci system.

utils/Makefile
```diff
@@ -59,13 +59,18 @@ docs: $(MANPAGES)
 viewcover:
 	xdg-open cover/index.html
 
-virtualenv:
+# Conditional virtualenv building strategy taken from this great post
+# by Marcel Hellkamp:
+# http://blog.bottlepy.org/2012/07/16/virtualenv-and-makefiles.html
+venv: oo-installenv/bin/activate
```
tbielawa: That is true. References to the virtualenv are inconsistent right now.

utils/Makefile
```diff
@@ -59,13 +59,18 @@ docs: $(MANPAGES)
...
+oo-installenv/bin/activate: test-requirements.txt
 	@echo "#############################################"
 	@echo "# Creating a virtualenv"
 	@echo "#############################################"
-	virtualenv $(NAME)env
+	test -d venv || virtualenv $(NAME)env
```
tbielawa: That's true. The `test` should be testing `$(NAME)env`. I failed to update that part after I copied the example from the original source (linked in the comment block).

tbielawa:
@rhcarvalho I'm going to submit more PRs to address the bugs/inconsistencies you pointed out and backport them to the necessary branches.

utils/Makefile
tbielawa: Yes it is but I duplicated it in the `ci` target because we changed the test running order. Unit tests run earlier now, their output is lost on the screen. If users are running `make ci` locally I want to make it as easy as possible for them to review their own code. Hence why the message is printed at the end of `make ci`.

tbielawa:
@rhcarvalho 

> aos-ci-test
>
> My doubt:
> As a matter of understanding the existing processes, why do we run these tests when changing only that Makefile?

You're correct in that only the first (two?) checks need to run for changes like this. This isn't the answer I'd **like** to give, but right now that's just how things work :confused: The merge system in this repo is configured to require that the tests are ran before the merge button is activated. **Unless** it's a special case and a repo super admin has reason to override the default checks. 

> AFAIU those tests exercise calls to ansible-playbook and have no interaction whatsoever with make?
> The Makefile is used by the default test job (which I think runs automatically for whitelisted authors), and also in Travis (run automatically for everyone).

Correct, the **rest** of the tests in Jenkins exercise the actual playbooks. They have no actual relationship to the make system, but the code quality checks (`make ci`) do.

Yes the first checks (default jenkins and travis) run automatically. I just habitually run `aos-ci-test` to get the per-merge checks ran and passing so we can merge PRs without requiring special repo-admin overrides. Obviously this is not ideal behavior for PRs like this one.

Issue #2925, opened 2016-12-06
Playbooks do not fail gracefully when run with Ansible < 2.2


dtpest:
[provide a description of the issue]
Using libvirtd description

##### Version
git: openshift-ansible-3.4.17-1-212-ga3ce4c1
ansible 2.0.0.2

##### Steps To Reproduce
bin/cluster list libvirt '' <- does work
bin/cluster create libvirt lenaic <- fails

##### Current Result
ERROR! no action detected in task

The error appears to have been in '/home/labuser/openshift-ansible/roles/docker/tasks/main.yml': line 114, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

- name: Start the Docker service
  ^ here

##### Expected Result
Should start cluster

abutcher:
Hey @dtpest, the playbooks should have [verified that your ansible version is 2.2](https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/verify_ansible_version.yml). This task uses the systemd module which was introduced in ansible 2.2.

detiber:
@abutcher I wonder if we are hitting the error prior to the task being run.

abutcher:
@detiber You're right, makes sense that we'd get the syntax error rather than run.

jberkus:
Looks like none of the playbooks are verifying ansible version.  I just hit this with the BYO playbook as well.

detiber:
@jberkus I believe that is because the systemd module doesn't exist in 2.2. @mtnbikenc any ideas on how we could better test for ansible version?

dtpest:
Hey,
i'm using ansible 2.0.0.2 and not 2.2

dtpest:
Updated to 2.2 and ansible could start successfully

mtnbikenc:
@detiber It looks like `bin/cluster create` uses the `launch.yml` playbooks.  In looking at those playbooks, they are not including the `verify_ansible_version.yml` playbook where the version verification is done.

detiber:
@mtnbikenc @jberkus was hitting it with the byo playbooks as well.

@abutcher any thoughts?

Re-opening, since we should fail more gracefully for Ansible < 2.2

dustymabe:
can we change the title of this issue to more accurately reflect the problem ?

detiber:
@dustymabe done

mtnbikenc:
https://github.com/openshift/openshift-ansible/pull/2942 should address this.

PR #2942, opened 2016-12-07
Minimum Ansible version check

mtnbikenc:
Resolves: https://github.com/openshift/openshift-ansible/issues/2925

Moved version check to callback pluging to allow checking the version
prior to playbook syntax checking. This prevents errors on
syntax-checking with older versions of Ansible.


mtnbikenc:
@detiber Here's a first shot at a new version check. I figure we would also remove the `verify_ansible_version.yml` playbook.

callback_plugins/ansible_version_check.py
```diff
@@ -0,0 +1,33 @@
...
+import ansible
+
+try:
+    # Version 2.0+
+    from ansible.plugins.callback import CallbackBase
+except ImportError:
+    CallbackBase = object
+
+
+def print_red_bold(text):
```
tbielawa: There's a built-in with the callback system for printing 'fail' type text, using 'print' directly should be avoided for this. I'll come back with a proper code example in a few minutes.

tbielawa: Set the message string and then use the built-in `_display` (it's a part of callback objects). Here's an example where they detect a task failed and then use the `C` constants to print with the correct red color:

https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/callback/default.py#L55-L57

You'll have to make sure you initialize the parent class though: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/callback/default.py#L41

to make sure that the `_display` gets set: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/callback/__init__.py#L55

Or add the kwarg yourself to the ``__init__`` definition (I think).

mtnbikenc: In looking through trying to implement this I've found that the more we rely on Ansible, while trying to check the current Ansible version, we would break on older versions of Ansible.  In this instance `C.COLOR_ERROR` does not exist in Ansible 2.0.0.2 and bombs. :(  So, it seems we should very lightly rely on Ansible dependencies.  Currently I'm using `__version__` which has been around since the beginning.  Thoughts?

tbielawa: from `constants.py` it defines `COLOR_ERROR` with a default value of the string `red`. I think you could get away with calling `_display.display(...., color='red')`. The difference being that we are overriding any custom user configs (which we were doing anyway using the shell escape codes :) )

mtnbikenc:
@tbielawa I've updated the code and this works only on post 2.0.0.0 versions of Ansible, which is not what we want.  On ansible-1.9.6 I get:

```
AttributeError: 'CallbackModule' object has no attribute '_display'
```

mtnbikenc:
I've updated the code and tested back to Ansible v1.7 successfully.  Additionally, I have removed the verify_ansible_version playbook and references to the same.

tbielawa:
aos-ci-test

sdodson:
aos-ci-test

detiber:
@mtnbikenc one issue that I see is that there is no license on the code from the blog, and the blog lists 'All rights reserved' with the copywrite notice. I've emailed the author to see if he is willing to release a copy of the code with a permissive license.

mtnbikenc:
aos-ci-test

mtnbikenc:
aos-ci-test

PR #2950, opened 2016-12-08
Drop 3.2 upgrade playbooks.

### LINKED ISSUE #2949: .../v3_2/upgrade.yml references files that do not exist
[playbooks/byo/openshift-cluster/upgrades/v3_2/upgrade.yml#L58-L65](https://github.com/openshift/openshift-ansible/blob/d4771b1a9a331393f5f4bf6c1cdf130637546c43/playbooks/byo/openshift-cluster/upgrades/v3_2/upgrade.yml#L58-L65) includes:

- include: ../../../../common/openshift-cluster/upgrades/pre.yml
- include: ../../../../common/openshift-cluster/upgrades/upgrade.yml
- include: ../../../../common/openshift-cluster/upgrades/post.yml

None of these 3 files exist.

@dgoodwin assigning it to you, as I think this was a missing part of the refactor in #2441.

I looked for an easy fix myself, but seems that the changes were more complex than I could follow in a reasonable amount of time.

dgoodwin:
These were supposed to be cleaned up and are currently broken as the large control plane / node refactor was not applied here.

Fixes #2949.


rhcarvalho:
@dgoodwin can we delete https://github.com/openshift/openshift-ansible/tree/master/playbooks/common/openshift-cluster/upgrades/v3_1_to_v3_2 as well? It has just a symlink in there.

rhcarvalho:
And FWIW looks like we have old references here too:
https://github.com/openshift/openshift-ansible/blob/master/utils/src/ooinstall/cli_installer.py#L28-L45

`git grep 3_1` reveals even more.

dgoodwin:
@rhcarvalho I added the nuke_images symlink to this commit. 

There are likely a lot of conditionals floating around in the tree for older versions that could now be cleaned up for sure.

sdodson:
Do we want to wait on this sort of cleanup until we have CI jobs for upgrades?

rhcarvalho:
> Do we want to wait on this sort of cleanup until we have CI jobs for upgrades?

As per #2949, these playbooks cannot be run today, as they refer to files that do not exist.

dgoodwin:
Yeah should be safe @sdodson, quite confident this is just dead broken code.

PR #2951, opened 2016-12-08
Integration tests stub

rhcarvalho:
In the past few weeks, @brenton @sosiouxme and I have been working on several pre-flight checks: playbook(s) that verify certain bits of the environment that folks might run before an installation, upgrade or to debug an existing cluster.

We've been also experimenting with an additional approach to integration testing of such playbooks.

I'd like to start bringing our work gradually to make it easier to review, and get better feedback.

I've cherry-picked commits from our [work-branch](https://github.com/rhcarvalho/openshift-ansible/commits/pre-flight-checks) related to the integration tests, for I think this is an easier piece to grasp and part of the foundation of our tests.
More tests (here is just a demo of the concept) and the playbooks will come in separate PRs.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,6 +3,7 @@ minversion=2.3.1
 envlist =
     py{27,35}-{flake8,pylint,unit}
     py27-{yamllint,ansible_syntax,generate_validation}
+    integration
 skipsdist=True
 skip_missing_interpreters=True
 
@@ -13,6 +14,8 @@ deps =
     -rtest-requirements.txt
     py35-flake8: flake8-bugbear
 
+whitelist_externals = env
+
 commands =
     unit: pip install -e utils
     unit: pytest {posargs}
@@ -22,3 +25,8 @@ commands =
     generate_validation: python setup.py generate_validation
     # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
     ansible_syntax: ansible-playbook --syntax-check playbooks/byo/config.yml
+
+    # Unset GOPATH because tests use relative imports. This should be removed if
+    # we require openshift-ansible to live in a Go work space and use absolute
+    # imports in tests (desirable).
+    integration: env -u GOPATH go test -v ./test/integration/...
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -9,3 +9,4 @@ coverage
 mock
 pytest
 pytest-cov
+docker-py
--- a/test/integration/README.md
+++ b/test/integration/README.md
@@ -1,0 +1,12 @@
+# Integration tests
+
+Integration tests exercise the OpenShift Ansible playbooks by performing
+simulated installations in Docker containers.
+
+## Running the tests
+
+From the repository root, run with:
+
+```
+tox -e integration
+```
--- a/test/integration/openshift_health_checker/common.go
+++ b/test/integration/openshift_health_checker/common.go
@@ -1,0 +1,99 @@
+package test
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"testing"
+)
+
+// A PlaybookTest executes a given Ansible playbook and checks the exit code and
+// output contents.
+type PlaybookTest struct {
+	// inputs
+	Path string
+	// expected outputs
+	ExitCode int
+	Output   []string // zero or more strings that should be in the output
+}
+
+// Run runs the PlaybookTest.
+func (p PlaybookTest) Run(t *testing.T) {
+	// A PlaybookTest is intended to be run in parallel with other tests.
+	t.Parallel()
+
+	cmd := exec.Command("ansible-playbook", p.Path)
+	cmd.Env = append(os.Environ(), "ANSIBLE_FORCE_COLOR=1")
+	b, err := cmd.CombinedOutput()
+
+	// Check exit code.
+	if (err == nil) && (p.ExitCode != 0) {
+		p.checkExitCode(t, 0, p.ExitCode, cmd, b)
+	}
+	if (err != nil) && (p.ExitCode == 0) {
+		got, ok := getExitCode(err)
+		if !ok {
+			t.Logf("unexpected error (%T): %[1]v", err)
+			p.logCmdAndOutput(t, cmd, b)
+			t.FailNow()
+		}
+		p.checkExitCode(t, got, p.ExitCode, cmd, b)
+	}
+
+	// Check output contents.
+	var missing []string
+	for _, s := range p.Output {
+		if !bytes.Contains(b, []byte(s)) {
+			missing = append(missing, s)
+		}
+	}
+	if len(missing) > 0 {
+		t.Logf("missing in output: %q", missing)
+		p.logCmdAndOutput(t, cmd, b)
+		t.FailNow()
+	}
+}
+
+// getExitCode returns an exit code and true if the exit code could be taken
+// from err, false otherwise.
+// The implementation is GOOS-specific, and currently only supports Linux.
+func getExitCode(err error) (int, bool) {
+	exitErr, ok := err.(*exec.ExitError)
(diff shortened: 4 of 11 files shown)


rhcarvalho:
@abutcher @detiber PTAL

@brenton and I would love your feedback!

.travis.yml
```diff
@@ -11,3 +11,4 @@ script:
   # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
   - ansible-playbook --syntax-check playbooks/byo/config.yml
   - cd utils && make ci
+  - make test-integration
```
rhcarvalho: We have already some integration with vagrant-openshift to run these tests in our Jenkins, but before we transition it to point to openshift/openshift-ansible, Travis can demo a test run.

utils/setup.cfg
```diff
@@ -5,7 +5,9 @@
 universal=1
 
 [nosetests]
-tests=../,../roles/openshift_master_facts/test/,test/
+tests = ../test/unit,
+        .,
+        ../roles/openshift_master_facts/test,
```
rhcarvalho: Could have been a separate PR... I realized we were missing 2 tests via `cd utils; make ci`.

Makefile
```diff
@@ -0,0 +1,7 @@
+.PHONY: test
+test:
+	nosetests
```
rhcarvalho: @detiber @tbielawa when we move utils/Makefile into the top-level, we can have `make test` be the one to give very quick feedback during development, while keeping the `make ci` target doing all the things it does today (setup venv, unit tests, flake8, pylint).

detiber:
I'll take a closer look at this tonight/tomorrow, but I wanted to share this article I came across the other day talking about testing with Ansible: https://medium.com/@sebinjohn/testing-ansible-roles-using-molecule-and-vagrant-a15c22af23ab#.8yu427q47

Makefile
detiber: I would probably argue that we should be running unit-tests and flake8 for make test, since both ideally will give quick feedback.

.travis.yml
detiber: make test-integration is running from the utils directory, probably want to do something like the following instead:
```
- pushd utils
- make ci
- popd
- make test-integration
```

setup.cfg
```diff
@@ -0,0 +1,4 @@
+[nosetests]
+tests = test/unit,
+        utils,
+        roles/openshift_master_facts/test,
```
detiber: should `roles/openshift_master_facts/test/*` be moved to `roles/openshift_master_facts/test/unit/` as well?

test/integration/preflight/preflight_test.go
```diff
@@ -0,0 +1,120 @@
+package preflight
```
detiber: Can't say I'm a huge fan of using golang for testing here, seems like it is going to be difficult to get external contributors to participate if the tests are written in golang.

I'd much prefer that we move towards a framework like https://molecule.readthedocs.io instead, since it also allows for much more comprehensive testing against multiple machines, etc.

.travis.yml
rhcarvalho: Problem solved.

Makefile
rhcarvalho: For now we're still using (cd utils; make ...); let's move utils/Makefile -> Makefile in a separate PR and we can have a more focused discussion of what runs when.

setup.cfg
rhcarvalho: Possibly not. I moved the Python files under `test/` just to have a better organization and reuse the `test/` top-level dir.

@detiber, I would even consider moving `test/unit/modify_yaml_tests.py` -> `library/test_modify_yaml.py`, so that test files live close to what they test.

rhcarvalho: > consider moving test/unit/modify_yaml_tests.py -> library/test_modify_yaml.py, so that test files live close to what they test

@detiber what do you think about that idea?

detiber: @rhcarvalho I'm not sure how well that would work, since then the test would be in the Ansible module lookup path.

test/integration/preflight/preflight_test.go
detiber: >>Can't say I'm a huge fan of using golang for testing here, seems like it is going to be difficult to get external contributors to participate if the tests are written in golang.

>Why do you think so?

Most external contributors will not be familiar with golang, If they are already advanced Ansible users, then we can generally assume they are familiar with Python.

>>I'd much prefer that we move towards a framework like https://molecule.readthedocs.io

>Thanks for the link. I see how it could be useful for us in end-to-end testing of the many playbooks in openshift-ansible (installation, upgrades, one-off actions). IIUC it's a YAML-centric framework for a particular flow of testing playbooks? I see it can do check syntax, convergence, idempotence, verify final state.

The verify final state is the one that allows for arbitrary validations to be run.

>For testing the preflight checks, what we need is a way to run openshift-ansible someplaybook.yml and check the exit code and output. Each test playbook will setup a dynamic inventory composed of one or more containers, and then will include another playbook from the preflight checks, to ensure that the check work correctly under certain environments (e.g. system with missing subscriptions, or incompatible versions of packages).

I'll play around with it some, I'm pretty sure this use case is baked in, even if the docs are lacking in how to do it.

>We wanted something that could give us fast feedback to iterate on the checks. Thus, we use Docker images instead of VMs. Running the tests in parallel seemed desirable, and Go's testing let us do that without much hassle.

Molecule is able to use Docker as a backend as well. I'll have to play around with the parallelization, but I think we can accomplish something similar.

.travis.yml
```diff
@@ -15,8 +15,14 @@ python:
...
   - pip install tox-travis coveralls
 
+env:
+  - TEST_SUITE=unit
+  - TEST_SUITE=integration
+
 script:
-  - tox
+  # REVIEW(rhcarvalho): if our Travis account doesn't support more than one
+  # build at a time, there is no point in parallelizing unit/integration tests.
```
rhcarvalho: Our account can do 2 concurrent builds across all openshift/* repos.

rhcarvalho:
Updated to use `tox` and have some examples of what tests could look like:
- automate setting up containers for a dynamic test inventory
- run a playbook
- make assertions about exit code and output

tox.ini
```diff
@@ -22,3 +25,8 @@ commands =
     generate_validation: python setup.py generate_validation
     # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
     ansible_syntax: ansible-playbook --syntax-check playbooks/byo/config.yml
+
+    # Unset GOPATH because tests use relative imports. This should be removed if
+    # we require openshift-ansible to live in a Go work space and use absolute
+    # imports in tests (desirable).
+    integration: env -u GOPATH go test -v ./test/integration/...
```
detiber: This doesn't seem like it should be needed: http://tox.readthedocs.io/en/latest/config.html#confval-passenv=SPACE-SEPARATED-GLOBNAMES

rhcarvalho: Hmm, I failed to see how that would be used?

Did you mean something like

```
[testenv:integration]
setenv=
    GOPATH=
```
?

But then, that looks like a lot more to unset an envvar than using `env -u`?

I'm unsetting `GOPATH` just to not require the code to be within GOPATH, and use relative imports. Relative imports are notably a bad idea in Go, I just wanted to experiment and see how it would compare. This basically affects CI in that we need to decide if the checkout path is within a Go work space or not. And we need to decide between Travis and Jenkins...

detiber: If I'm reading the tox docs correctly, the GOPATH env var shouldn't be available in the test environment with the default config. Outside of a handful of whitelisted vars, you should have to specify which env cars you want to make available using `passenv` or by explicitly defining them with `setenv`.

rhcarvalho: Ah, okay, now I understand.
I'll update it.

rhcarvalho:
This was merged through #3816.

PR #2955, opened 2016-12-08
Fixes to 'make ci'

tbielawa:
* Fixed: A grep filter was capturing the actual pylint check rc's

* Changed: pylint used to print out messages for locally-disabled and
  file-ignored items

* Changed: pylint output format is now 'parseable'

* Cleaned up: Pylint was emitting deprecation messages for
  'required-attributes' and 'ignore-iface-methods'. They have been
  removed from the pylintrc file

### DIFF
--- a/git/.pylintrc
+++ b/git/.pylintrc
@@ -71,7 +71,7 @@ confidence=
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
 # w0511 - fixme - disabled because TODOs are acceptable
-disable=E1608,W1627,E1601,E1603,E1602,E1605,E1604,E1607,E1606,W1621,W1620,W1623,W1622,W1625,W1624,W1609,W1608,W1607,W1606,W1605,W1604,W1603,W1602,W1601,W1639,W1640,I0021,W1638,I0020,W1618,W1619,W1630,W1626,W1637,W1634,W1635,W1610,W1611,W1612,W1613,W1614,W1615,W1616,W1617,W1632,W1633,W0704,W1628,W1629,W1636,W0511,R0801
+disable=E1608,W1627,E1601,E1603,E1602,E1605,E1604,E1607,E1606,W1621,W1620,W1623,W1622,W1625,W1624,W1609,W1608,W1607,W1606,W1605,W1604,W1603,W1602,W1601,W1639,W1640,I0021,W1638,I0020,W1618,W1619,W1630,W1626,W1637,W1634,W1635,W1610,W1611,W1612,W1613,W1614,W1615,W1616,W1617,W1632,W1633,W0704,W1628,W1629,W1636,W0511,R0801,locally-disabled,file-ignored
 
 
 [REPORTS]
@@ -79,7 +79,7 @@ disable=E1608,W1627,E1601,E1603,E1602,E1605,E1604,E1607,E1606,W1621,W1620,W1623,
 # Set the output format. Available formats are text, parseable, colorized, msvs
 # (visual studio) and html. You can also give a reporter class, eg
 # mypackage.mymodule.MyReporterClass.
-output-format=text
+output-format=parseable
 
 # Put messages in a separate file for each module / package specified on the
 # command line instead of printing them on stdout. Reports (if any) will be
@@ -113,9 +113,6 @@ logging-modules=logging
 
 
 [BASIC]
-
-# Required attributes for module, separated by a comma
-required-attributes=
 
 # List of builtins function names that should not be used, separated by a comma
 bad-functions=map,filter,input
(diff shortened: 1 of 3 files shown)


tbielawa:
aos-ci-test

utils/Makefile
```diff
@@ -62,45 +63,45 @@ viewcover:
...
 #       If there are any special things to install do it here
-#       . $(NAME)env/bin/activate && INSTALL STUFF
+#       . $(VENV)/bin/activate && INSTALL STUFF
 
 ci-unittests:
 	@echo "#############################################"
 	@echo "# Running Unit Tests in virtualenv"
 	@echo "#############################################"
-	. $(NAME)env/bin/activate && python setup.py nosetests --cover-erase
+	. $(VENV)/bin/activate && python setup.py nosetests --cover-erase
```
detiber: I would suggest adding `cover-erase` to setup.cfg rather than adding it to the command line here.

utils/Makefile
```diff
@@ -62,45 +63,45 @@ viewcover:
...
-	touch $(NAME)env/bin/activate
+	test -d $(VENV) || virtualenv $(VENV)
+	. $(VENV)/bin/activate && pip install setuptools==17.1.1
+	. $(VENV)/bin/activate && pip install -r test-requirements.txt
+	touch $(VENV)/bin/activate
 #       If there are any special things to install do it here
-#       . $(NAME)env/bin/activate && INSTALL STUFF
+#       . $(VENV)/bin/activate && INSTALL STUFF
 
 ci-unittests:
```
detiber: With this re-org, should we make `$(VENV)` a dependency for all the `ci-*` tasks? Then we should be able to drop the `$(VENV)` dependency from `ci` and make running the individual `ci-*` tasks a bit easier.

utils/Makefile
tbielawa: done

utils/Makefile
tbielawa: good idea. done

tbielawa:
aos-ci-test

tbielawa:
@detiber PTAL

utils/Makefile
```diff
@@ -62,45 +63,45 @@ viewcover:
...
+$(VENV)/bin/activate: test-requirements.txt
 	@echo "#############################################"
 	@echo "# Creating a virtualenv"
 	@echo "#############################################"
-	test -d venv || virtualenv $(NAME)env
-	. $(NAME)env/bin/activate && pip install setuptools==17.1.1
-	. $(NAME)env/bin/activate && pip install -r test-requirements.txt
-	touch $(NAME)env/bin/activate
+	test -d $(VENV) || virtualenv $(VENV)
+	. $(VENV)/bin/activate && pip install setuptools==17.1.1
```
rhcarvalho: Out of curiosity, do you know why are we pinning the `setuptools` version to this particular version?

utils/Makefile
```diff
@@ -62,45 +63,45 @@ viewcover:
...
+	. $(VENV)/bin/activate && pip install -r test-requirements.txt
+	touch $(VENV)/bin/activate
 #       If there are any special things to install do it here
-#       . $(NAME)env/bin/activate && INSTALL STUFF
+#       . $(VENV)/bin/activate && INSTALL STUFF
 
-ci-unittests:
+ci-unittests: $(VENV)
```
rhcarvalho: :cry:  this means that if I want to use `make ci-unittests` I cannot manage my own env/virtualenv...
For my Python projects I use `virtualenvwrapper`, and my envs are not a subdirectory of the repo. I wish I could still use that :)

(the fact that we activate the env in every target already prevents me from managing my own env)

git/.pylintrc
```diff
@@ -71,15 +71,15 @@ confidence=
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
 # w0511 - fixme - disabled because TODOs are acceptable
-disable=E1608,W1627,E1601,E1603,E1602,E1605,E1604,E1607,E1606,W1621,W1620,W1623,W1622,W1625,W1624,W1609,W1608,W1607,W1606,W1605,W1604,W1603,W1602,W1601,W1639,W1640,I0021,W1638,I0020,W1618,W1619,W1630,W1626,W1637,W1634,W1635,W1610,W1611,W1612,W1613,W1614,W1615,W1616,W1617,W1632,W1633,W0704,W1628,W1629,W1636,W0511,R0801
+disable=E1608,W1627,E1601,E1603,E1602,E1605,E1604,E1607,E1606,W1621,W1620,W1623,W1622,W1625,W1624,W1609,W1608,W1607,W1606,W1605,W1604,W1603,W1602,W1601,W1639,W1640,I0021,W1638,I0020,W1618,W1619,W1630,W1626,W1637,W1634,W1635,W1610,W1611,W1612,W1613,W1614,W1615,W1616,W1617,W1632,W1633,W0704,W1628,W1629,W1636,W0511,R0801,locally-disabled,file-ignored
```
rhcarvalho: :+1: this is much better than grep!

utils/Makefile
tbielawa: We pin to that version because it makes sure that we get consistent tests across platforms. It was a while ago when we had to pin that down so my memory is a little fuzzy, but... IIRC we were getting inconsistent versions installed across platforms (clients vs. jenkins). There was a function missing/renamed across a specific version boundary and we settled on using this version. No real technical deliberation, that's just the decision we made to get things working consistently.

detiber: I want to say that we hit some type of an issue with syntax between versions, and we pinned the specific version to know that it would work, instead of fighting with trying to get the syntax exactly right. At some point, we probably want to revisit that.

PR #2956, opened 2016-12-08
[WIP] Wait after disabling firewalld

smunilla:


detiber:
@smunilla were you able to create a reproducer for this?

leifmadsen:
I don't think this is required anymore, as I see the same code in both `firewalld.yml` and `iptables.yml` now.

sdodson:
@smunilla what's the latest on this? I'm fine with a one time 10 second pause if we think it will avoid problems.

PR #2961, opened 2016-12-09
Limit node certificate SANs to node hostnames/ips.

abutcher:
Node serving certificate hostnames are currently pulled from `openshift.common.all_hostnames` which [can include many names if the node is also a master](https://github.com/openshift/openshift-ansible/blob/ce39c9084bc20df242bbbef3a2a5b1c8060cbe9d/roles/openshift_facts/library/openshift_facts.py#L721-L732). This change limits those hostnames to node hostnames and IP addresses.

### DIFF
--- a/roles/openshift_node_certificates/tasks/main.yml
+++ b/roles/openshift_node_certificates/tasks/main.yml
@@ -64,13 +64,13 @@
 - name: Generate the node server certificate
   command: >
     {{ hostvars[openshift_ca_host].openshift.common.client_binary }} adm ca create-server-cert
-      --cert={{ openshift_node_generated_config_dir }}/server.crt
-      --key={{ openshift_generated_configs_dir }}/node-{{ openshift.common.hostname }}/server.key
-      --overwrite=true
-      --hostnames={{ openshift.common.all_hostnames |join(",") }}
-      --signer-cert={{ openshift_ca_cert }}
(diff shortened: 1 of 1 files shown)


abutcher:
aos-ci-test

roles/openshift_node_certificates/tasks/main.yml
```diff
@@ -64,13 +64,13 @@
...
-      --key={{ openshift_generated_configs_dir }}/node-{{ openshift.common.hostname }}/server.key
-      --overwrite=true
-      --hostnames={{ openshift.common.all_hostnames |join(",") }}
-      --signer-cert={{ openshift_ca_cert }}
-      --signer-key={{ openshift_ca_key }}
-      --signer-serial={{ openshift_ca_serial }}
+    --cert={{ openshift_node_generated_config_dir }}/server.crt
+    --key={{ openshift_generated_configs_dir }}/node-{{ openshift.common.hostname }}/server.key
+    --overwrite=true
+    --hostnames={{ openshift.common.hostname }},{{ openshift.common.public_hostname }},{{ openshift.common.ip }},{{ openshift.common.public_ip }}
```
detiber: hmm, I don't think public_hostname or public_ip is ever used by the node. That said, it probably doesn't hurt to include it.

PR #2962, opened 2016-12-09
upgrades - Fix logic error about when to backup etcd

sdodson:
We want backups on by default. When we add etcd upgrade to the default upgrade playbooks we'll want to reconsider this.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -29,7 +29,7 @@
 
 - name: Backup etcd
   include: ./etcd/backup.yml
-  when: openshift_upgrade_skip_etcd_backup | default(false) | bool
+  when: not openshift_upgrade_skip_etcd_backup | default(false) | bool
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -29,7 +29,7 @@
 
 - name: Backup etcd
   include: ./etcd/backup.yml
-  when: openshift_upgrade_skip_etcd_backup | default(false) | bool
+  when: not openshift_upgrade_skip_etcd_backup | default(false) | bool
```
detiber: Should this be named `openshift_upgrade_backup_etcd` instead, with a default value of `true`. We'd have to support a renaming at least temporarily, but at least the future logic won't continue to make my head hurt.

sdodson: It's only recently been added and was never documented. I'd be in favor of making it easier to read. I'll include that in my work to include etcd upgrades by default.

PR #2963, opened 2016-12-09
Upgrade etcd by default

sdodson:
- Move etcd backup inclusion into playbooks/common/openshift-cluster/upgrades/etcd/main.yml
 - Include etcd/main.yml in control plane upgrade
 - Create a group for upgrading etcd which excluded embedded etcd hosts
 - Create a group for backing up etcd which includes embedded etcd hosts
 - Backup etcd
 - Upgrade etcd

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -27,9 +27,8 @@
         embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
         debug_level: "{{ openshift_master_debug_level | default(openshift.common.debug_level | default(2)) }}"
 
-- name: Backup etcd
-  include: ./etcd/backup.yml
-  when: not openshift_upgrade_skip_etcd_backup | default(false) | bool
+- name: Upgrade and backup etcd
+  include: ./etcd/main.yml
 
 - name: Upgrade master packages
   hosts: oo_masters_to_config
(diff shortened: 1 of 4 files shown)


sdodson:
@dgoodwin PTAL doing some final testing, so far i've tested embedded and non embedded RPM environments.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -27,9 +27,8 @@
         embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
         debug_level: "{{ openshift_master_debug_level | default(openshift.common.debug_level | default(2)) }}"
 
-- name: Backup etcd
-  include: ./etcd/backup.yml
-  when: not openshift_upgrade_skip_etcd_backup | default(false) | bool
+- name: Upgrade and backup etcd
+  include: ./etcd/main.yml
```
dgoodwin: Should we bake another variable in for customers to skip this step and not upgrade etcd, or should we consider this a requirement.

sdodson: Actually, at first i tried something like this and I think I came to the conclusion you cannot conditionally include a playbook? Tasks of a play can conditionally include task files however.

(slightly simplified)

```
- name: Backup etcd
  include: ./etcd/backup.yml
  when: embedded_etcd

- name: Upgrade etcd
  include: ./etcd/main.yml
  when: not embedded_etcd
```

However, I think we should consider it to be a requirement to upgrade etcd.

sdodson: Hmm, nevermind, given we just fixed the logic error in the other PR then I think this had to have worked. Perhaps it just fails based on hostvars? I believe I had

when: "{{ hostvars[groups.oo_first_master.0].openshift.master.embedded_etcd }}"

PR #2964, opened 2016-12-09
YAML Linting with CI checking

mtnbikenc:
* Added checks to `make ci` for yaml linting
* Modified y(a)ml files to pass lint checks


mtnbikenc:
aos-ci-test

git/.yamllint
```diff
@@ -0,0 +1,23 @@
+extends: default
```
tbielawa: Does this file allow for comment lines? I love that we can configure everything this way, but descriptions of what, and more importantly **why** a non-intuitive parameter is set such as it is are quite useful.

For example, why is `truthy` set to `disable`? Another quick question about that, is there a difference between `0` and `1` and `disable` and `enable`?

Follow up to that, which `yamllint` program are you using? On my laptop I actually have a script in `~/bin/` called `yamllint` that is just a crappy ruby interpreter that just tries to load the file given in `$1`.

mtnbikenc: FYI, `truthy` warnings/errors = 509

For example:
```
./roles/os_firewall/tasks/firewall/firewalld.yml
  10:14     warning  truthy value is not quoted  (truthy)
  11:13     warning  truthy value is not quoted  (truthy)
  22:14     warning  truthy value is not quoted  (truthy)
  23:13     warning  truthy value is not quoted  (truthy)
  24:20     warning  truthy value is not quoted  (truthy)
```

mtnbikenc:
@tbielawa I've added comments to the .yamllint config and updated a few files to meet more default requirements.  Please comment on my comments as to what you think should and should not be style defaults.  Thanks!

tbielawa:
@mtnbikenc :+1: to the the rules you proposed. I wouldn't change any of those.

## Comments

I made a few comments/editor hints in my checkout. See them in the gist diff at the end of this comment.

## Newlines

I did find an interesting problem when I ran `make ci-yamllint` from my checkout though. Without setting...

```yaml
  empty-lines:
    max-end: 1
```
...I received a warning about `playbooks/byo/config-rhel.yml` : 

> ../playbooks/byo/config-rhel.yml:3:1: [error] too many blank lines (1 > 0) (empty-lines)

According to the project website:

> The POSIX standard requires the last line to end with a new line character. All UNIX tools expect a new line at the end of files. Most text editors use this convention too.

http://yamllint.readthedocs.io/en/latest/rules.html#module-yamllint.rules.new_line_at_end_of_file

So I would propose that we add a section to enable trailing newlines.

## Makefile

My editor was telling me that the `Skipping all '/files/' folders due to conflicting yaml file definitions` comment was mis-indented. I think comments are supposed to start at the beginning of the line. Or at least, that's what my editor wants me to believe.

screenshot from 2016-12-12 09-11-39

I also noticed `$(NAME)env` still used in the yamllint target. Changing that for consistency to `$(VENV)` is desired.

We're slipping in two very long `find` shell commands now. I'd like to see those moved into recursive variables in the Makefile header.

```Makefile
YAMLFILES = $(shell find ../ -name $(VENV) -prune -o \( -name '*.yml' -o -name '*.yaml' \) ! -path "*/files/*" 2>&1)                                                                                                 
PYFILES = $(shell find ../ -name $(VENV) -prune -o -name ooinstall.egg-info -prune -o -name test -prune -o -name "*.py" -print)                                                                                      
...
ci-yamllint: $(VENV)
    . $(VENV)/bin/activate && yamllint -c ../git/.yamllint $(YAMLFILES)
ci-pylint: $(VENV)
    . $(VENV)/bin/activate && python -m pylint --rcfile ../git/.pylintrc $(PYFILES)
```

One last **question** - Obviously the `yamllint` target discovers a LOT of files (about 356 when I test it). Do we want to quiet down the Makefile output there? If we prefix the `. $(VENV)/bin/activate && yamllint...` line like so:

> @. $(VENV)/bin/activate && yamllint -c ../git/.yamllint $(YAMLFILES)

Then all 356+ yaml file names are not printed to the screen. This hides information from people, but also limits output to just results and what have you. On the left in this screenshot is the output when the command is silenced (`@`) and the right side is without the `@` silencer applied:

yamllint-hush-please

## Lemme patch ya up

I captured all of my suggestions into a gist diff for you:

* https://gist.github.com/tbielawa/1509bc137fa399bf054f1fdca80c27e6

tbielawa:
aos-ci-test

tbielawa:
Errata for my last review comment

> ...I received a warning about playbooks/byo/config-rhel.yml :

This was a verified level-1 PEBKAC error (I created that file myself and had it included in `.git/info/exclude`...).

mtnbikenc:
aos-ci-test

PR #2965, opened 2016-12-09
Updated OpenShift Master iptables rules

mtnbikenc:
* Removed unneeded rules
* Moved etcd rule to conditional based on usage of embedded etcd

https://bugzilla.redhat.com/show_bug.cgi?id=1386329

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -156,8 +156,6 @@
   - role: openshift_builddefaults
   - role: os_firewall
     os_firewall_allow:
-    - service: etcd embedded
-      port: 4001/tcp
     - service: api server https
       port: "{{ openshift.master.api_port }}/tcp"
     - service: api controllers https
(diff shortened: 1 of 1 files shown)


sdodson:
aos-ci-test

playbooks/common/openshift-master/config.yml
```diff
@@ -166,16 +164,11 @@
...
-      port: 2224/tcp
-    - service: Corosync UDP
-      port: 5404/udp
-    - service: Corosync UDP
-      port: 5405/udp
+  - role: os_firewall
+    os_firewall_allow:
+    - service: etcd embedded
+      port: 4001/tcp
+    when: embedded_etcd | bool
```
abutcher: We could check this as `groups.oo_etcd_to_config | default([]) | length == 0`. There's also an `openshift.master.embedded_etcd` fact which is equivalent and arrived at similarly.

mtnbikenc:
aos-ci-test

mtnbikenc:
@abutcher thanks for the tip on the var
@detiber All the tests look good now

PR #2973, opened 2016-12-13
Set metrics url even if metrics_deploy is false

AlbertoPeon:
Follow up of  #2938 

Allow to set the `metricsPublicUrl` in the master configuration without the need of deploying the metrics with Ansible.

@dgoodwin I've done it in the way you proposed, using an extra variable that is only set when `openshift_hosted_metrics_deploy`is `true` or when `openshift_hosted_metrics_public_url` is explicitly set in the inventory file. Could you verify?

Thanks!

### DIFF
--- a/roles/openshift_hosted_metrics/tasks/install.yml
+++ b/roles/openshift_hosted_metrics/tasks/install.yml
@@ -128,5 +128,5 @@
   modify_yaml:
     dest: "{{ openshift.common.config_base }}/master/master-config.yaml"
     yaml_key: assetConfig.metricsPublicURL
-    yaml_value: "{{ openshift_hosted_metrics_public_url }}"
+    yaml_value: "{{ openshift_hosted_metrics_deploy_url }}"
   notify: restart master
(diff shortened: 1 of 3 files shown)


dgoodwin:
bot, test pull request

sdodson:
aos-ci-test

AlbertoPeon:
It is not clear to me why the test failed to complete. The error I can see is:
```
PLAY [nodes] *******************************************************************

fatal: [10.8.170.51]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: ssh: connect to host 10.8.170.51 port 22: No route to host\r\n", "unreachable": true}
ok: [10.8.170.6]
```
In any case, it doesn't seem relevant to this pull request. 

Could someone review why the test didn't pass? @sdodson 

Thanks!

AlbertoPeon:
Hi @dgoodwin @sdodson,

Could the test be run once again? The patch looks good IMHO.

ingvagabund:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
ok to test

PR #2975, opened 2016-12-13
Partial uninstall functionality

sjug:
When working with large clusters we would like to save time in "resetting" the cluster state. Rather than running a full uninstall followed by a full install, this partial uninstall is useful and saves time. By default there is no functionality change, however with parameter `openshift_remove_all=False` we neuter the uninstall. The file deletions have been split up to ensure the RPM owned folders still exist.

Internal ticket: [Cluster Soft Reset - Trello](https://trello.com/c/5AfB75Dp/204-tools-cluster-soft-reset)

/cc @detiber

### DIFF
--- a/playbooks/adhoc/uninstall.yml
+++ b/playbooks/adhoc/uninstall.yml
@@ -75,6 +75,10 @@
 
 - hosts: nodes
   become: yes
+  vars:
+    node_dirs:
+    - "/etc/origin"
+    - "/var/lib/origin"
   tasks:
   - name: unmask services
     command: systemctl unmask "{{ item }}"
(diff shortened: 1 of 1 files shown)


sjug:
aos-ci-test

playbooks/adhoc/uninstall.yml
```diff
@@ -182,19 +185,35 @@
...
+    when: openshift_remove_all | default(True) | bool
     file:
       path: /run/openshift-sdn
       state: absent
 
+  - name: Remove generated files
+    file: path={{ item }} state=absent
+    with_fileglob:
+    - /etc/origin/master/*
+    - /etc/origin/node/*
```
detiber: with_fileglob operates on the local host, This would only work if the playbook is run on a host that is both a master and a node. It would still potentially miss some files as well.

playbooks/adhoc/uninstall.yml
```diff
@@ -85,7 +85,7 @@
 
   - name: Remove packages
     package: name={{ item }} state=absent
-    when: not is_atomic | bool
+    when: not is_atomic | bool and openshift_remove_all | default(True) | bool
```
detiber: Instead of applying all of these conditionals to each task, what about grouping them within a block?

Something like this, might clean things up a bit:
```
- block:
  - block:
    - name: Remove packages
      package: name={{ item }} state=absent
      with_items:
      ...
    - name: Remove flannel
      package: name=flannel state=absent
      when: openshift_use_flannel | default(false)
    when: "{{ not is_atomic | bool }}"

  - shell: systemctl reset-failed
    changed_when: False
 
  - shell: systemctl daemon-reload
    changed_when: False
  
  - name: Remove br0 interface
    shell: ovs-vsctl del-br br0
    when: openshift_remove_all | default(True) | bool
    changed_when: False
    failed_when: False
  
  - name: Remove linux interfaces
    shell: ip link del "{{ item }}"
    when: openshift_remove_all | default(True) | bool
    changed_when: False
    failed_when: False
    with_items:
    - lbr0
    - vlinuxbr
    - vovsbr
  when: "{{ openshift_remove_all | default(true) | bool }}"
```

playbooks/adhoc/uninstall.yml
detiber: I think you should be safe to just remove /etc/origin/master and /etc/origin/node, without issues. The installer at least should create those if absent.

playbooks/adhoc/uninstall.yml
```diff
@@ -182,19 +185,35 @@
...
+
+  - name: Remove files owned by RPMs
+    file: path={{ item }} state=absent
+    when: openshift_remove_all | default(True) | bool
+    with_items:
+    - /etc/origin
+    - /etc/sysconfig/openshift-node
+    - /etc/sysconfig/openvswitch
+    - /run/openshift-sdn
+    - /var/lib/origin
```
detiber: It might be easier to just remove /etc/origin and /var/lib/origin and re-create them if openshift_remove_all is false.

sjug: I considered it, but wasn't sure if you were okay with creating things during uninstallation. I'll do that instead.

playbooks/adhoc/uninstall.yml
sjug: This would mean we'd still need multiple blocks, is that still preferred? I'm open to it, no problem with me.

detiber: I'm good with multiple blocks, and even the conditions being applied to individual tasks where needed for ordering.

playbooks/adhoc/uninstall.yml
sjug: Can we talk to the ansible guys and ask them to bump the priority on https://github.com/ansible/ansible-modules-core/pull/3095**

playbooks/adhoc/uninstall.yml
abutcher: @sjug Right

detiber: The example I gave above had nested blocks, so the inside when is applied to the innermost block.

sjug: I've added a few blocks that seem logical.

playbooks/adhoc/uninstall.yml
detiber: @sjug yes, I'll reach out. The workaround stated in https://github.com/ansible/ansible/issues/18910#issuecomment-265544149 looks like a better approach than we've used in the past as a stopgap.

sjug: Should I use find instead of re-creating the directories?

sjug:
@detiber What do you think of the find based deletions?

detiber:
@sjug looks good, any issues with testing?

sjug:
@detiber No issues so far on 10-20 node HA clusters, but I will hopefully tomorrow be able to test this in the scale lab.

timothysc:
Are we waiting on anything else with this PR?  

@sjug @detiber

sjug:
@timothysc I thought that we could test this out on a larger scale pre-merge. Other than that functionally it is good to go.

detiber:
@sjug it looks like there is a merge conflict, otherwise I'm good with merging these changes.

sjug:
@detiber Rebased

PR #2979, opened 2016-12-14
Update os_firewall_manage_iptables.py

ebalsumgo:
changing from /usr/libexec/iptables/iptables.init to /sbin/iptables-save allows saving to somewhere other than /etc/sysconfig/iptables which could be under outside source control and using iptables.d/ is a better location for rules that need to persist over reboots in such situations.


ebalsumgo:
so as @detiber noted this might be better off using os.chdir("/etc/sysconfig/iptables.d") than what I am doing there.

mtnbikenc:

PR #2985, opened 2016-12-16
[uninstall] Remove excluder packages

sdodson:
You will lose hours of your life trying to install again if you don't do this.


sdodson:
@tbielawa ptal

Issue #2970, opened 2016-12-12
check local docker images before pull from docker hub


vanloswang:
[provide a description of the issue]
I specified the openshift_image_tag to 1.4.0 on the installation and build the images with the tag of v1.4.0, it will always notice me docker pull failed and no such tag of v1.4.0. Finally I found out that the tag of v1.4.0 is not exists now, there is only v1.4.0-alpha.1 and v1.4.0-rc1 on docker hub, so this is a misconfiguration. But I specify the right tag and have the right container images locally, it should install ok for me. So Should we check local docker images before pull from docker hub?

##### Version
openshift-ansible-3.4.37 

##### Steps To Reproduce
1. sepcify the openshift_image_tag to 1.4.0 on the installation and build container images with the tag of v1.4.0 locally
2. install an all in one cluster by container installation

##### Current Result
it failed when pull image

##### Expected Result
pass pulling images when the images is stored locally.

sdodson:
We could add an option to avoid pulling images if the admin believes they're already there. We've added the pre-pulling of images because without it the installation would often fail.

vanloswang:
@sdodson maybe the following code will be helpful
```bash
cat chkimage.yml
- hosts : localhost
  remote_user : root
  tasks :
  - name: Check local image
    command: >
      echo `docker images openshift/origin:v1.4.0-rc1` | grep -v grep | grep openshift/origin | grep v1.4.0-rc1
    register: check_result

  - name: master image exists
    command: echo "Image exits..."
    when: check_result | bool

  - name: master image does exit
    command: echo "Image does exit..."
    when: not check_result | bool
```

sdodson:
Yeah, something like that. I'd add it immediately before the steps where we attempt to pull the images, if it exists, don't pull it. We're always happy to receive pull requests.

PR #2992, opened 2016-12-18
check local image before pull it

vanloswang:
add this for https://github.com/openshift/openshift-ansible/issues/2970


vanloswang:
@sdodson Could you take a look at it. Thanks.

PR #2993, opened 2016-12-19
Add RPM checks as a byo playbook

rhcarvalho:
- Playbook to run RPM checks
- Custom callback plugin to summarize errors
- Custom modules to interact with yum


rhcarvalho:
Authored by @sosiouxme, @brenton and I, squashed into a single commit for simplicity.

@detiber @abutcher PTAL

rhcarvalho:
@detiber @abutcher PTAL

callback_plugins/default_plus_summary.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class CallbackModule(CallbackModule_default):
+    '''
+    This is like the default callback plugin, but also stores results and
+    summarizes failures.
+    '''
+
+    CALLBACK_VERSION = 2.0
+    CALLBACK_TYPE = 'stdout'
```
detiber: If you set CALLBACK_TYPE to 'aggregate', then you wouldn't need to replace the default stdout callback plugin.

Also, this and the module should live in a role.

brenton: If set to aggregate would the normal stdout plugin still run?  Would the custom one simply run afterwards?  The idea of this callback plugin was to have reduced stdout that includes better error message aggregation.

detiber: @brenton Yes, the default plugin would still run, aggregate plugins are meant to provide supplemental output.

Thinking from a practical use case, it seems like being able to collect and aggregate output would be useful independent of a separate 'tool' that would just provide limited output. Maybe having two callback plugins, one that would provide the aggregation and summarization and one that could replace the default callback plugin would be the way to go to achieve both use cases?

rhcarvalho: Trying to get it to work as an 'aggregate' plugin, got this error.
@detiber have you seen this before?

While I understand the Python part of the error, I fail to understand what Ansible is doing...


```console
$ ANSIBLE_DEBUG=1 ansible-playbook -i hosts.ini playbooks/adhoc/preflight/check.yml -vvv
  9855 1483635797.56724: starting run
Using /home/vagrant/src/github.com/openshift/openshift-ansible/ansible.cfg as config file
  9855 1483635797.69092: Loading CacheModule 'memory' from /usr/lib/python2.7/site-packages/ansible/plugins/cache/memory.py
  9855 1483635797.84638: Loading CallbackModule 'default_plus_summary' from /home/vagrant/src/github.com/openshift/openshift-ansible/callback_plugins/default_plus_summary.py
  9855 1483635797.84912: Loading CallbackModule 'aa_version_requirement' from /home/vagrant/src/github.com/openshift/openshift-ansible/callback_plugins/aa_version_requirement.py (found_in_cache=False, class_only=True)
  9855 1483635797.84947: Loading CallbackModule 'actionable' from /usr/lib/python2.7/site-packages/ansible/plugins/callback/actionable.py (found_in_cache=False, class_only=True)
  9855 1483635797.84981: Loading CallbackModule 'context_demo' from /usr/lib/python2.7/site-packages/ansible/plugins/callback/context_demo.py (found_in_cache=False, class_only=True)
  9855 1483635797.85005: Loading CallbackModule 'debug' from /usr/lib/python2.7/site-packages/ansible/plugins/callback/debug.py (found_in_cache=False, class_only=True)
  9855 1483635797.85207: Loading CallbackModule 'default' from /home/vagrant/src/github.com/openshift/openshift-ansible/callback_plugins/default.py (found_in_cache=False, class_only=True)
  9855 1483635797.85252: Loading CallbackModule 'default' from /usr/lib/python2.7/site-packages/ansible/plugins/callback/default.py (found_in_cache=False, class_only=True)
  9855 1483635797.85268: Loading CallbackModule 'default_plus_summary' from /home/vagrant/src/github.com/openshift/openshift-ansible/callback_plugins/default_plus_summary.py (found_in_cache=False, class_only=True)
  9855 1483635797.85284: RUNNING CLEANUP
ERROR! Unexpected Exception: super(type, obj): obj must be an instance or subtype of type
the full traceback was:

  File "/usr/bin/ansible-playbook", line 103, in <module>
    exit_code = cli.run()
  File "/usr/lib/python2.7/site-packages/ansible/cli/playbook.py", line 159, in run
    results = pbex.run()
  File "/usr/lib/python2.7/site-packages/ansible/executor/playbook_executor.py", line 89, in run
    self._tqm.load_callbacks()
  File "/usr/lib/python2.7/site-packages/ansible/executor/task_queue_manager.py", line 195, in load_callbacks
    self._callback_plugins.append(callback_plugin())
  File "/home/vagrant/src/github.com/openshift/openshift-ansible/callback_plugins/default_plus_summary.py", line 22, in __init__
    super(CallbackModule, self).__init__()
  File "/usr/lib/python2.7/site-packages/ansible/plugins/callback/default.py", line 41, in __init__
    super(CallbackModule, self).__init__()
TypeError: super(type, obj): obj must be an instance or subtype of type
```


The diff to the code here is just:

```diff
diff --git a/callback_plugins/default_plus_summary.py b/callback_plugins/default_plus_summary.py
index 610a3e9..77452dc 100644
--- a/callback_plugins/default_plus_summary.py
+++ b/callback_plugins/default_plus_summary.py
@@ -15,7 +15,7 @@ class CallbackModule(CallbackModule_default):
     '''
 
     CALLBACK_VERSION = 2.0
-    CALLBACK_TYPE = 'stdout'
+    CALLBACK_TYPE = 'aggregate'
     CALLBACK_NAME = 'default_plus_summary'
 
     def __init__(self):
```

rhcarvalho: Nevermind, I got it working.

I can't find much documentation around plugins... my best friend has been the ansible source code.

It seems that [apart from `stdout` and `tree`, the ansible code ignores the CALLBACK_TYPE completely](https://github.com/ansible/ansible/blob/ad1e0665401dd436b2043ed64b8ffcad49e205b7/lib/ansible/executor/task_queue_manager.py#L185-L190)... it seems to be of just an informational character, and maybe a gentlemen's agreement not written down anywhere?!

FWIW, setting `CALLBACK_TYPE = 'lorem-ipsum'` produces the same result as `'aggregate'` :smile:

rhcarvalho:
@brenton @detiber one more detail I'm noticing with using callback plugins to control output is that they are loaded based on the [sorted order](https://github.com/ansible/ansible/blob/ad1e0665401dd436b2043ed64b8ffcad49e205b7/lib/ansible/plugins/__init__.py#L384) of the plugin file names.

E.g. if I add the line below to my `ansible.cfg` file:

```
callback_whitelist = timer
```

Then the output of my `ansible-playbook` run becomes:


```
...

Failure summary:

  1. ...
  ...

  4. Host:     openshift_ansible_test_preflight_fail_all
     Play:     verify check results
     Task:     ensure all checks succeed
     Message:  Failed as requested from task
Playbook run took 0 days, 0 hours, 0 minutes, 7 seconds
```

If we rename the plugin file name to `zz_failure_summary.py`, then it would print after the output from the `timer` plugin.

I see we already name a plugin `callback_plugins/aa_version_requirement.py` to make it load first...
I don't want to be paranoid about people having whatever other plugins enabled, but that could possibly mean unpredictable output.

Do you have any opinion on this?
Should we have some prefix on the file name to try to make it load first or last?

rhcarvalho:
> this (the callback plugin) and the module should live in a role.

@detiber would you have a suggestion for a name / naming pattern for the role?

Last year I've spent some time thinking how to approach breaking it into roles, but there is no clear naming pattern coming up. For now it would be simply a single "roles/openshift_preflight_check", with all the code there... perhaps you had a better idea?

detiber:
> If we rename the plugin file name to zz_failure_summary.py, then it would print after the output from the timer plugin.

> I see we already name a plugin callback_plugins/aa_version_requirement.py to make it load first...
I don't want to be paranoid about people having whatever other plugins enabled, but that could possibly mean unpredictable output.

> Do you have any opinion on this?

I don't think we need to be overly concerned about user-defined or user-enabled plugins. If users are savy enough to be playing around with plugin config, they should be savy enough to deal with the results.

> Should we have some prefix on the file name to try to make it load first or last?

I'm good with prefixing with zz_ to ensure that it is loaded after any other plugins we use withing openshift_ansible.

detiber:
>> this (the callback plugin) and the module should live in a role.

> @detiber would you have a suggestion for a name / naming pattern for the role?

> Last year I've spent some time thinking how to approach breaking it into roles, but there is no clear naming pattern coming up. For now it would be simply a single "roles/openshift_preflight_check", with all the code there... perhaps you had a better idea?

That works for me. The callback plugin may have more generic use, so maybe a roles/openshift_check_common role or something similar?

rhcarvalho:
Right now, the playbook looks like this:

```yaml
---
- hosts: OSEv3
  gather_facts: no
  tasks:
  - set_fact:
      ...

- hosts: OSEv3
  name: check content available on all hosts
  gather_facts: no
  ignore_errors: yes
  tasks:
  - ...

- hosts: masters
  name: determine if yum install of master pkgs will work
  gather_facts: no
  ignore_errors: yes
  tasks:
  - ..

- hosts: nodes
  name: determine if yum install of node pkgs will work
  gather_facts: no
  ignore_errors: yes
  tasks:
  - ...

- hosts: OSEv3
  name: verify check results
  gather_facts: no
  tasks:

  - set_fact:
      oo_preflight_check_failures: "{{ oo_preflight_check_results | select('failed', 'equalto', True) | list }}"

  - name: ensure all checks succeed
    action: fail
    when: oo_preflight_check_failures
```

How to break this into role(s)?

IIUC a role cannot have a `hosts:` declaration, therefore a role and its tasks must be made to run on a single host/group.

I think the 5 plays we have are reasonable. The first sets variables and errors are not ignored there. The next 3 plays execute checks in a) all hosts b) masters c) nodes. The last play determines the exit status of `ansible-playbook` by failing if any check failed.

That brings us to:

```yaml
---
- hosts: OSEv3
  gather_facts: no
  roles:
  - openshift_preflight_facts

- hosts: OSEv3
  name: check content available on all hosts
  gather_facts: no
  ignore_errors: yes
  roles:
  - openshift_preflight_check_all_hosts

- hosts: masters
  name: determine if yum install of master pkgs will work
  gather_facts: no
  ignore_errors: yes
  roles:
  - openshift_preflight_check_masters

- hosts: nodes
  name: determine if yum install of node pkgs will work
  gather_facts: no
  ignore_errors: yes
  roles:
  - openshift_preflight_check_nodes

- hosts: OSEv3
  name: verify check results
  gather_facts: no
  roles:
  - openshift_preflight_verify_status
```

@detiber is this along the lines of what you would expect?

The set_fact'ing to store check status across multiple roles smells like a bad pattern.

detiber:
@rhcarvalho With the custom callback plugin, I see leveraging custom modules that would be able to do the tests and then you wouldn't need to use ignore_errors.

For handling role-based checks, I see there being a few different ways to handle it:
- Break each set of checks into a different role, using the playbook to enforce ordering.
  - This would not require passing in role-related info into the role, but could cause a proliferation of roles
- Use a variable to indicate which roles a host is a member of
  - Conditionalize certain checks based on the roles a host is a member of
- Combination of the above.

rhcarvalho:
> @rhcarvalho With the custom callback plugin, I see leveraging custom modules that would be able to do the tests and then you wouldn't need to use ignore_errors.

Does this mean making modules that always return 0, and relying on parsing the output to know if it failed or succeed?

detiber:
>> @rhcarvalho With the custom callback plugin, I see leveraging custom modules that would be able to do the tests and then you wouldn't need to use ignore_errors.

> Does this mean making modules that always return 0, and relying on parsing the output to know if it failed or succeed?

Yes, since the modules are about diagnostics/tests, the callback plugin could parse the result and store the relevant errors/warnings from the output. There would still be a need for using register and possibly a filter plugin to provide a pass/fail at the end, for the purposes of failing to prevent installation/etc.

rhcarvalho:
@detiber PTAL. If you're fine with the current shape we can get this in and iterate as we add more checks.

library/aos_version.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/python
```
detiber: Minor nit, instead of these living in <repo root>/library and being symlinked into the role, they should simply live in the role.

playbooks/adhoc/preflight/README.md
```diff
@@ -0,0 +1,43 @@
...
+    ```console
+    $ git clone https://github.com/openshift/openshift-ansible
+    $ cd openshift-ansible
+    ```
+
+2. Run the playbook:
+
+    ```console
+    $ ansible-playbook -i <inventory file> playbooks/adhoc/preflight/check.yml
+    ```
```
detiber: Instead of playbooks/adhoc, I would suggest either playbooks/common or playbooks/byo for the playbook. Since it looks like you are using the byo groups, playbooks/byo is probably the best location.

playbooks/adhoc/preflight/check.yml
```diff
@@ -0,0 +1,32 @@
+---
+- hosts: OSEv3
+  gather_facts: no
+  roles:
+    - openshift_preflight/facts
```
detiber: Instead of nesting these roles under openshift_preflight, they should really be broken out (openshift_preflight_facts, openshift_preflight_all, openshift_preflight_masters, openshift_preflight_nodes, openshift_preflight_verify, etc)

roles/openshift_preflight/common/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+- when:
+    - not containerized
+  block:
+
+    - name: determine if yum update will work
+      action: check_yum_update
+      register: r
+
+    - set_fact:
+        oo_preflight_check_results: "{{ oo_preflight_check_results + [r|combine({'_task': 'check content available on all hosts'})] }}"
```
detiber: I'm not sure we can reliably use this method of updating variables in this manner.

Instead of redefining and updating oo_preflight_check_results in all playbooks, what if there was a single variable per role to track the results and the playbook aggregated the result variables and determined pass/fail?

roles/openshift_preflight/facts/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: set common variables
+  set_fact:
+    deployment_type: "{{ deployment_type | default('openshift-enterprise') }}"
+    containerized: "{{ containerized | default('no') | bool }}"
```
detiber: It might be better to set containerized by default if running on atomic host, we do this with l_containerized in openshift_facts/tasks/main.yml

roles/openshift_preflight/facts/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: set common variables
+  set_fact:
+    deployment_type: "{{ deployment_type | default('openshift-enterprise') }}"
```
detiber: In general, we usually want to default to origin if not specified. It might be better to expect the user to set this (and if we use a container or some other method of distribution we override the value there).

roles/openshift_preflight/facts/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
...
+    containerized: "{{ containerized | default('no') | bool }}"
+    openshift_release: "{{ openshift_release | default('3.3') }}"
```
detiber: Same comment as with deployment_type.

roles/openshift_preflight/masters/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+  block:
+
+    - name: main packages for enterprise
+      when:
+        - deployment_type == "openshift-enterprise"
+      check_yum_update:
+        packages:
+          - atomic-openshift
+          - atomic-openshift-clients
+          - atomic-openshift-master
```
detiber: I would suggest using a variable to track the 'package prefix' here, with the prefix being 'origin' for 'origin' deployments and 'atomic-openshift' for ocp deployments. Then there can be a single task for the main packages instead of having a task for each deployment_type (or even just a single task in total).

roles/openshift_preflight/nodes/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+  block:
+
+    - name: main packages for enterprise
+      when:
+        - deployment_type == "openshift-enterprise"
+      check_yum_update:
+        packages:
+          - atomic-openshift
+          - atomic-openshift-node
+          - atomic-openshift-sdn-ovs
```
detiber: Same comment here about using a variable for the package name prefix to avoid having separate tasks for each deployment_type.

roles/openshift_preflight/nodes/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+      register: r
+
+    - set_fact:
+        oo_preflight_check_results: "{{ oo_preflight_check_results + [r|combine({'_task': 'main packages for origin'})] }}"
+
+    - name: other node packages
+      check_yum_update:
+        packages:
+          - docker
+          - PyYAML
```
detiber: PyYAML is a common dependency because of openshift_facts, shouldn't be node-specific.

roles/openshift_preflight/nodes/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+          - firewalld
+          - iptables
+          - iptables-services
```
detiber: firewalld and iptables-services should be mutually exclusive, might need to test for those depending on the value of `os_firewall_use_firewalld`

roles/openshift_preflight/nodes/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+          - nfs-utils
+          - ntp
```
detiber: ntp should be a common dependency.

roles/openshift_preflight/nodes/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+          - yum-utils
```
detiber: yum-utils should also be a common dependency

roles/openshift_preflight/nodes/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+          - dnsmasq
+          - libselinux-python
+          - ceph-common
+          - glusterfs-fuse
+          - iscsi-initiator-utils
+          - pyparted
+          - python-httplib2
+          - openssl
```
detiber: I'm not completely sure here, but I suspect openssl is a dependency for the ca roles, which are limited to etcd/masters.

roles/openshift_preflight/nodes/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+          - nfs-utils
+          - ntp
+          - yum-utils
+          - dnsmasq
+          - libselinux-python
+          - ceph-common
+          - glusterfs-fuse
+          - iscsi-initiator-utils
+          - pyparted
+          - python-httplib2
```
detiber: not sure what python-httplib2 is used for, but I suspect it's used in the tasks we use to verify that a port is listening, which would most likely be limited to masters

roles/openshift_preflight/nodes/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+          - iptables-services
+          - nfs-utils
+          - ntp
+          - yum-utils
+          - dnsmasq
+          - libselinux-python
+          - ceph-common
+          - glusterfs-fuse
+          - iscsi-initiator-utils
+          - pyparted
```
detiber: I think pyparted is used by one of the unsupported nfs storage roles, I don't think it should be listed here.

roles/openshift_preflight/nodes/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+          - python-httplib2
+          - openssl
+          - flannel
```
detiber: flannel is only required if `openshift_use_flannel` is set to true.

roles/openshift_preflight/nodes/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+          - bind
```
detiber: I suspect that bind should only be a conditional dependency, it shouldn't be required for most installs.

library/aos_version.py
rhcarvalho: I don't have an opinion here, but a bit confused.

A previous suggestion was exactly on the other direction, to move this to the top-level so that it could be "potentially used" by other roles.

Looking at the repository today one can see cases of both: symlinks and modules "hidden" inside of roles. What is the practice we're generally trying to follow? IOW, what should I be thinking about when making this decision myself next time?

rhcarvalho:
@detiber thanks for the review.
Do you know of a better way to get the list of package dependencies instead of enumerating packages as we have now, certainly duplicating the ansible code we have for actually installing or conditionally installing those dependencies?

playbooks/adhoc/preflight/README.md
rhcarvalho: > Since it looks like you are using the byo groups, playbooks/byo is probably the best location.

Let me take the chance to learn something more here.

Ideally the checks should be equally valid for a number of installation paths, not limited to BYO.

Among the choices (in `/playbooks`): adhoc, aws, byo, common, gce, libvirt, openstack; I thought `adhoc` was appropriate for its name not implying any specific target infrastructure, and also that you can run the checks independent of doing an installation.

Definitely `common` would be an option as well, but it's subdirectory structure and the many `include: ../common/.../...yml` references led me to think `common` plays a different role: being included in other playbooks for code reuse.

Could we document in the repo what each thing under `/playbooks` is meant for? I volunteer to create a PR for that if I can get an initial explanation :)

playbooks/adhoc/preflight/check.yml
rhcarvalho: I followed [this idea](https://groups.google.com/d/topic/ansible-project/2as7VmaoHXY/discussion) from the Ansible mailing-list to avoid growing arbitrarily the number of top level roles, while relying on just a common prefix to group them together.

In particular, I don't see these "roles" being usable by themselves, so it seemed more like a sub-role idea than multiple _reusable roles_ (what I understand to be the idea behind Ansible Roles).

TBH I'm not even super comfortable that the current format can scale well and be maintainable. So I'd think to be wise to allow for some imperfection, and embrace refactors instead of trying to get everything right on the first shot. We do need to write more checks and understand better the requirements and constraints of what we're doing to be able to steer the implementation.

roles/openshift_preflight/common/tasks/main.yml
rhcarvalho: I'm quite new to Ansible. I feel it is hard to use the word *reliable* and this type of "code" together, it feels like a hack to me :)
There are many ways to introduce subtle mistakes.

From a programming stand point, I think the current approach is already quite crazy and hard to maintain... But was the best we've got working with Ansible so far, after experimenting with different things.

Having different variable names for each role means that introducing a new check requires changes in at least two different places -- the new role and changing the code that checks these variables at the end of the playbook run. And it would be very very easy to miss this second part and assume things are working.

library/aos_version.py
detiber: Having modules live in the root of the repo reduce re-usability, since they will not be in the lookup path if called from a playbook that exists outside of openshift-ansible.

If there is a module that needs to be used by more than one role, it should be handled by role dependencies.

playbooks/adhoc/preflight/README.md
detiber: >Ideally the checks should be equally valid for a number of installation paths, not limited to BYO.

>Among the choices (in /playbooks): adhoc, aws, byo, common, gce, libvirt, openstack; I thought adhoc was appropriate for its name not implying any specific target infrastructure, and also that you can run the checks independent of doing an installation.

adhoc was started for tasks that really don't have a home, existing playbooks living there are generally unsupported and not maintained.

common was meant as a generic set of playbooks that would run against different "providers" when we were supporting the bin/cluster tooling. playbooks in common use generic group names and do not line up with the actual groups used by the byo playbooks or bin/cluster derived playbooks.

aws, gce, libvirt, openstack playbook directories are meant to be driven by  bin/cluster, which is community supported and most use is considered deprecated.

>Definitely common would be an option as well, but it's subdirectory structure and the many include: ../common/.../...yml references led me to think common plays a different role: being included in other playbooks for code reuse.

Yes, if the playbook were to live in common, it would be necessary to provide a byo playbook that included the common playbook with the correct group mappings, etc.

>Could we document in the repo what each thing under /playbooks is meant for? I volunteer to create a PR for that if I can get an initial explanation :)

Definitely something that is needed. Let me know if you need any additional detail/description.

playbooks/adhoc/preflight/check.yml
detiber: @rhcarvalho I'm definitely open to adding a concept of sub-roles, I just wanted to avoid adding a new concept without having further discussion on it. I do worry a bit about it introducing some confusion, but I can definitely see it helping to clean up the directory structure and helping with role organization.

That said, I would expect that these roles would be able to be fully integrated into a 100% automated workflow without having to be called from distinct playbook. The mechanics of how that is accomplished is something that would need to be worked out, though.

roles/openshift_preflight/facts/tasks/main.yml
rhcarvalho: I agree.

@brenton gave an interesting suggestion which I think we could implement by checking variables defined in the std facts:

> <bleanhar> if we detect RHEL or Atomic Host we default to running the OCP checks
> <bleanhar> anything else we run the Origin checks
> <bleanhar> and then let it be overridden as needed for advanced cases

Do you think that is reasonable? Is there a standard way to check for RHEL or Atomic Host?

roles/openshift_preflight/common/tasks/main.yml
detiber: > I'm quite new to Ansible. I feel it is hard to use the word reliable and this type of "code" together, it feels like a hack to me :)
There are many ways to introduce subtle mistakes.

My concern here is that this is relying on undocumented behavior by relying on redefining a variable that already exists.

> From a programming stand point, I think the current approach is already quite crazy and hard to maintain... But was the best we've got working with Ansible so far, after experimenting with different things.

I agree, it is definitely not the ideal solution to the problem.

> Having different variable names for each role means that introducing a new check requires changes in at least two different places -- the new role and changing the code that checks these variables at the end of the playbook run. And it would be very very easy to miss this second part and assume things are working.

I think I see the issue here. I'm assuming that the only change would need to be in the new role, and that the check at the end of the playbook would simply be looking for the existence of any errors in the known variables. The only case I see there needing to be a change at the end of the playbook for verification is when a new role is added for checks, which would still require changes to the playbook for applying that role. Since the callback plugin is responsible for tracking the actual warnings/errors. I'm thinking something like this for a test (assuming the modules returned the number of errors found):

role a:
```
openshift_preflight_a_error_count = "{{ result_1.errors + result_2.errors + result_3.errors }}"
```

role b:
```
openshift_preflight_b_error_count = "{{ result_1.errors + result_2.errors }}"
```

role c:
```
openshift_preflight_c_error_count = "{{ result_1.errors + result_2.errors + result_3.errors + result_4.errors }}"
```

preflight playbook:
```
...
- hosts: ...
  roles:
  - a
  - b
  - c
  post_tasks:
  - fail:
    msg: "Preflight errors exist"
    when: "{{ (openshift_preflight_a_error_count + openshift_preflight_b_error_count + openshift_preflight_c_error_count) > 0 }}"
```

detiber:
> Do you know of a better way to get the list of package dependencies instead of enumerating packages as we have now, certainly duplicating the ansible code we have for actually installing or conditionally installing those dependencies?

Not accurately, no. I think it would make sense to refactor the package installation tasks so that they can be called in a way that could handle both the preflight testing and the actual package installation. I wouldn't expect that to be tied to this PR, though.

roles/openshift_preflight/facts/tasks/main.yml
detiber: @rhcarvalho for testing for atomic host, it seems that you can simply test for the existence of `/run/ostree-booted`.

I'm not sure that testing for a RHEL host is sufficient for testing for Origin/OCP, though. One could install Origin on RHEL.

brenton: @detiber, we're open to ideas.  Our preference is for options that don't result in users having to answer questions or supply arguments when.  Somehow we want to guess what they are likely to want to install simply by running the tool.  Obviously, there are less common edge-cases and for those they can use settings.  Right now we're thinking a single pre-flight container image can exist.

detiber: @brenton any reason we couldn't bake some of the variables into the container image? We could default deployment_type easily during image build, with the OCP images defaulting to `openshift_enterprise` and the Origin images defaulting to `origin`? We could use this for defaulting version numbers as well (pull the image/tag related to the version you want to install)

brenton: We really just don't want to have more images than we need right now.  I'd say it's on the table for the future but until we start shipping something that people find useful we want to multiple options for them.

detiber: @brenton in that case, I don't see an issue with assuming OCP if using RHEL as an interim workaround.

roles/openshift_preflight/common/tasks/main.yml
rhcarvalho: Thanks for the snippets, it makes it easier to comprehend.
I find that pattern unsatisfactory to work with.

Every time a task is added or removed or moved, one needs to (unless otherwise automated with templates / YAML generation) manually update the definition of `openshift_preflight_a_error_count`. It is so easy to make a mistake an use `result_7` twice, etc. We should not ask humans to type and maintain those enumerated summations.

Similarly, every time a role is added/removed, etc, the `post_tasks` in the playbook needs to change. If I add a role `d` and forget to add `openshift_preflight_d_error_count` we have a silent bug...

Put those things together, and there is room for a complex web of bugs.


---

I didn't come up with the idea myself, but can't remember/find where I took it from... I would also prefer an explicitly documented feature. Here is what I find in the docs:

http://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

> A lot of folks may ask about how variables override another. (...)
> Avoid defining the variable “x” in 47 places and then ask the question “which x gets used”. Why? Because that’s not Ansible’s Zen philosophy of doing things. (...)
> Figure out where to define a variable, and don’t make it complicated.

We clearly go against the philosophy here by implementing a workaround that contradicts Ansible's ideal (fail on first error).

But:

> If multiple variables of the same name are defined in different places, they get overwritten in a certain order. (...)

Doesn't that count as making the variable overriding well-defined and documented?

Variable overriding is a thing, and even the precedence order is listed in the docs (we only use `set_fact`).

detiber: > Variable overriding is a thing, and even the precedence order is listed in the docs (we only use set_fact).

Yes, variable overriding is supported and predictable. The thing that I'm not sure is officially supported and reliable is referencing the overridden variable when overriding the variable.

rhcarvalho: If the "language" allows for referencing variables and for setting variables (the `set_fact` module), I don't see why it would be an exception to reference a variable you are overriding?!

In pseudo-code:

```yaml
- set_fact:
    result: []

- set_fact:
    tmp: '{{ result }}'

- set_fact:
    result: '{{ tmp + [2] }}'
``` 

It would be a breaking change and probably a step backwards to require a `tmp` variable like that. That's not how Python works, not how Ansible works...

detiber: @rhcarvalho In general I agree, however it didn't work in the past with Ansible.

That said, the Ansible Core team has said that if it works now, then it should be supported in the future. So I don't see an issue with using that pattern here.

library/aos_version.py
rhcarvalho: Two questions:

The two modules are used in 3 roles (`common`, `masters` and `nodes`). If the idea is to avoid symlinks, what would you propose -- create yet another role just to hold the modules and then add it as a dependency of each of the other 3?

Should we consider the existing modules in `/library` and their symlinks a "bug/techdebt" or `modify_yaml.py` and `rpm_q.py` are in their right place?

playbooks/adhoc/preflight/README.md
rhcarvalho: https://github.com/openshift/openshift-ansible/pull/3083

rhcarvalho: Playbook moved to `playbooks/byo/openshift-preflight`

roles/openshift_preflight/facts/tasks/main.yml
rhcarvalho: I made `roles/openshift_preflight/facts` depend on `openshift_facts`, so that we can reuse all variables from there.

roles/openshift_preflight/facts/tasks/main.yml
rhcarvalho: I think the easier to implement and better to maintain path is to reuse `openshift_facts`, and that's what we're doing now.

Any improvements to the defaulting in `openshift_facts` will be absorbed by us automatically, but I'll leave it out of this PR.

roles/openshift_preflight/masters/tasks/main.yml
rhcarvalho: Done using `openshift.common.service_type`.

roles/openshift_preflight/nodes/tasks/main.yml
rhcarvalho: Done using `openshift.common.service_type`.

roles/openshift_preflight/nodes/tasks/main.yml
rhcarvalho: Is there an easy way to conditionally add items to a list?

library/aos_version.py
detiber: I would suggest an additional role that is included via dependency in meta/main.yml from common, masters, and nodes roles.

roles/openshift_preflight/nodes/tasks/main.yml
detiber: The simplest way that I can think of is through the use of a filter plugin, which would give you access to the variables that are set and the ability to return a limited list...  possibly a filter that takes a dictionary like this:
```
'pkg_a':
  'condition': True
'pkg_b':
  'condition': "{{ openshift_use_flannel | bool }}"
```

library/aos_version.py
rhcarvalho: > I would suggest an additional role that is included via dependency in meta/main.yml from common, masters, and nodes roles.

I'm trying that. Apparently there is a side-effect that the dependency gets loaded/executed 3 times... is that expected behavior? It makes the playbook run slower.

roles/openshift_preflight/nodes/tasks/main.yml
rhcarvalho: Could you please point me to existing code that does that? (e.g. in the installation playbooks)

library/aos_version.py
detiber: @rhcarvalho yes, I expect that the dependency could be invoked multiple times, is there a considerable slowdown even when the included role does not actually contain any tasks, but just includes the required module(s)?

Alternatively, since you are now requiring openshift_facts, there is potential for the module to exist there instead.

roles/openshift_preflight/nodes/tasks/main.yml
detiber: @rhcarvalho I can't think of an instance where we are currently doing this pattern, but the core filter plugin would do something similar to this:

```
def filter_packages_by_condition(data):
    filtered_list = []
    for key in data:
        if data[key]['condition']:
            filtered_list.append(key)
    return filtered_list
```

obviously some type checking and validation is needed, but I believe that would do accomplish the goal.

library/aos_version.py
rhcarvalho: > Alternatively, since you are now requiring openshift_facts, there is potential for the module to exist there instead.

That's along the lines of what I did. I have a `openshift_preflight/base` (renamed from `facts`) that depends on `openshift_facts` AND includes the modules AND sets the initial value for `oo_preflight_check_results`.

`openshift_facts` takes at least 10s on an inventory with a single docker-based node. This makes the run time go from 7s to about 40s. (A single inclusion/run of `openshift_facts` in https://github.com/openshift/openshift-ansible/pull/2993/commits/c4aec0fbd112fd1c8503e935ac954da268a2fdff increased the time to 17s).

Those 10s times a multiplier is more than what I would like :(

Is there a way to avoid running the same thing twice? I definitely would not want to compose slowdowns in this way. Adding to my notes that role dependencies have some potentially dark side-effects...

rhcarvalho: Maybe I'm hitting this bug https://github.com/ansible/ansible/issues/20169

rhcarvalho: Well, the Ansible docs say:

http://docs.ansible.com/ansible/playbooks_roles.html
> Roles dependencies are always executed before the role that includes them, and are recursive. By default, roles can also only be added as a dependency once - if another role also lists it as a dependency it will not be run again. This behavior can be overridden by adding allow_duplicates: yes to the meta/main.yml file.

I did not set `allow_duplicates`, so I understand `openshift_facts` should be run only once...

Puzzled :confused:

roles/openshift_preflight/nodes/tasks/main.yml
rhcarvalho: Where do we list the packages that need to be installed in each case, for each type of host? Don't we need to pass in a conditional list?

brenton: We don't really have a list.  You could search for 'ansible_pkg_mgr' in openshift-ansible to see where packages are being installed.  I suspect some of these may be dependencies and not even explicitly installed in openshift ansible.

library/aos_version.py
rhcarvalho: Moved to a dumb role, should be less of a problem if loaded multiple times.

rhcarvalho:
@detiber @brenton while we have a review of what packages shouldn't be in the lists, should be moved or conditionally checked, we don't have a good way to tell we're not omitting anything. I'd prefer to leave it for now as @sosiouxme did originally and move on to iterating on the checks and the interface how users will run them.

Ideally I want a sane way to go from what openshift-ansible does/requires to what we check. But this can be done in a follow up.

While I'm positive on learning Ansible and openshift-ansible, I'm also eager to get this merged and start iterating on user feedback.

rhcarvalho:
aos-ci-test

PR #2998, opened 2016-12-20
Use tox and molecule for Ansible testing

detiber:
To test an individual role (openshift_facts and openshift_master_facts currently have a molecule definition):
```
cd roles/<rolename>
molecule test
```

To test all roles that have a molecule definition:
```
cd <repo root>
python setup.py molecule_tests
```

To use tox to test all roles:
```
cd <repo root>
tox -e py27-ansible22-functional
```

As currently configured, the molecule tests defined using testinfra: https://testinfra.readthedocs.io

Molecule docs are: http://molecule.readthedocs.io/en/v2

Currently, molecule is being pulled from the v2_tweaks branch of my fork, since we are relying on the currently in development v2 version of molecule to enable Ansible-native provisioning support and support for running multiple scenarios.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,26 @@
 ---
 sudo: false
 
-cache:
-  - pip
-
 language: python
+
 python:
   - "2.7"
   - "3.5"
 
+cache:
+  - pip
+
+before_cache:
+  - rm ~/.cache/pip/log/debug.log
+
+env:
+  - TEST_DIR=root TEST_TYPE=lint
+  - TEST_DIR=root TEST_TYPE=unit
+  - TEST_DIR=utils
+
 install:
-  - pip install -r requirements.txt
   - pip install tox-travis
 
 script:
-  # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
-  - ansible-playbook --syntax-check playbooks/byo/config.yml
-  - tox
-  - cd utils && tox
+  - if [[ "${TEST_DIR}" == "root" ]]; then tox; fi
+  - if [[ "${TEST_DIR}" == "utils" ]]; then cd utils && tox; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 ---
-sudo: false
+sudo: required
+
+services:
+  - docker
 
 language: python
 
@@ -16,7 +19,13 @@ before_cache:
 env:
   - TEST_DIR=root TEST_TYPE=lint
   - TEST_DIR=root TEST_TYPE=unit
+  - TEST_DIR=root TEST_TYPE=functional
   - TEST_DIR=utils
+
+matrix:
+  exclude:
+    - python: "3.5"
+      env: TEST_DIR=root TEST_TYPE=functional
 
 install:
   - pip install tox-travis
--- a/molecule_common/base/Dockerfile
+++ b/molecule_common/base/Dockerfile
@@ -1,0 +1,6 @@
+FROM centos/systemd:latest
+
+RUN yum makecache fast && yum update -y && \
+    yum install -y sudo yum-plugin-ovl && \
+    sed -i 's/^plugins=0$/plugins=1/g' /etc/yum.conf && \
+    yum clean all
--- a/molecule_common/origin-latest/Dockerfile
+++ b/molecule_common/origin-latest/Dockerfile
@@ -1,0 +1,6 @@
+FROM openshift/origin:latest
+
+RUN yum makecache fast && yum update -y && \
+    yum install -y sudo yum-plugin-ovl && \
+    sed -i 's/^plugins=0$/plugins=1/g' /etc/yum.conf &&\
+    yum clean all
--- a/roles/lib_openshift/molecule/origin-latest/molecule.yml
+++ b/roles/lib_openshift/molecule/origin-latest/molecule.yml
@@ -1,0 +1,35 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: docker
+  options:
+    privileged: yes
+lint:
+  name: ansible-lint
+platforms:
+- name: oa-role-lib-openshift-test
+  groups:
+  - test_group
+provisioner:
+  name: ansible
+  config_options:
+    defaults:
+      callback_whitelist: profile_tasks
+  group_vars:
+    test_group:
+      deployment_type: origin
+      openshift_deployment_type: "{{ deployment_type }}"
+  host_vars:
+    localhost:
+      molecule_docker_config_subdir: 'origin-latest'
+scenario:
+  name: origin-latest
+  setup: ../../../../molecule_common/create.yml
+  teardown: ../../../../molecule_common/destroy.yml
+verifier:
+  name: testinfra
+  env:
+    ANSIBLE_LIBRARY: ../../library
+  options:
+    n: 4
--- a/roles/lib_openshift/molecule/origin-latest/tests/test_oc_secret.py
+++ b/roles/lib_openshift/molecule/origin-latest/tests/test_oc_secret.py
@@ -1,0 +1,330 @@
+import base64
+import pytest
+import testinfra.utils.ansible_runner
+import uuid
+import yaml
+
+testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
+    '.molecule/ansible_inventory.yml').get_hosts('test_group')
+
+
+@pytest.fixture()
+def namespace(Command, tmpdir):
+    namespace = 'test-' + str(uuid.uuid4())
+
+    Command.run_test("mkdir -p {}".format(tmpdir))
+    Command.run_test("cp /etc/origin/master/admin.kubeconfig {}/admin.kubeconfig".format(tmpdir))
+    Command.run_test("oc new-project {} --config {}/admin.kubeconfig".format(namespace, tmpdir))
+
+    yield namespace
+
+    Command.run_test("oc delete project {} --config {}/admin.kubeconfig".format(namespace, tmpdir))
+    Command.run_test("rm -rf {}".format(tmpdir))
+
+
+@pytest.fixture()
+def test_config_file():
+    return {
+        'name': 'config.yml',
+        'contents': yaml.dump({'value': True})
+    }
+
+
+@pytest.fixture()
+def test_passwd_file():
+    return {
+        'name': 'passwords',
+        'contents': "test1\ntest2\ntest3\ntest4"
+    }
+
+
+@pytest.fixture()
+def secret(namespace, tmpdir, Ansible, test_config_file, test_passwd_file):
+    secret_name = 'test-{}'.format(uuid.uuid4())
+
+    config_file = tmpdir.join(test_config_file['name'])
+    with config_file.open(mode='w') as f:
+        f.write(test_config_file['contents'])
+
+    passwd_file = tmpdir.join(test_passwd_file['name'])
+    with passwd_file.open(mode='w') as f:
+        f.write(test_passwd_file['contents'])
+
+    Ansible('file', {'dest': str(tmpdir), 'state': 'directory'}, check=False)
+    Ansible('copy', {'src': str(config_file), 'dest': str(config_file)}, check=False)
+    Ansible('copy', {'src': str(passwd_file), 'dest': str(passwd_file)}, check=False)
+
+    secret_create_params = {
+        'name': secret_name,
+        'namespace': namespace,
+        'state': 'present',
+        'files': [
+            {
+                'name': 'config.yml',
+                'path': str(config_file)
+            },
+            {
+                'name': 'passwords',
+                'path': str(passwd_file)
+            }
+        ]
+    }
+    secret_create = Ansible('oc_secret', secret_create_params, check=False)
+    assert secret_create['changed'] is True
+    assert secret_create['results']['returncode'] == 0
+
+    yield secret_create_params
+
+    secret_delete_params = {
+        'name': secret_name,
+        'namespace': namespace,
+        'state': 'absent'
+    }
+    secret_delete = Ansible('oc_secret', secret_delete_params, check=False)
+    assert secret_delete['changed'] is True
+    assert secret_delete['results']['returncode'] == 0
+
+    Ansible('file', {'dest': str(tmpdir), 'state': 'absent'}, check=False)
+
+
+def test_decode(secret, Ansible, test_config_file, test_passwd_file):
+    secret_list_params = {
+        'name': secret['name'],
+        'namespace': secret['namespace'],
+        'decode': 'yes',
+        'state': 'list'
+    }
+    secret_list = Ansible('oc_secret', secret_list_params)
+    assert secret_list['changed'] is False
+    assert secret_list['results']['exists'] is True
+    assert 'decoded' in secret_list['results']
+    assert test_config_file['name'] in secret_list['results']['decoded']
+    assert test_passwd_file['name'] in secret_list['results']['decoded']
+    assert len(secret_list['results']['decoded']) == 2
+    assert secret_list['results']['decoded'][test_config_file['name']] == test_config_file['contents']
+    assert secret_list['results']['decoded'][test_passwd_file['name']] == test_passwd_file['contents']
+
+
+def test_delete_after(tmpdir, namespace, Ansible, test_config_file, test_passwd_file, File):
+    secret_name = 'test-{}'.format(uuid.uuid4())
+
+    config_file = tmpdir.join(test_config_file['name'])
+    with config_file.open(mode='w') as f:
+        f.write(test_config_file['contents'])
+
+    passwd_file = tmpdir.join(test_passwd_file['name'])
+    with passwd_file.open(mode='w') as f:
+        f.write(test_passwd_file['contents'])
+
+    Ansible('file', {'dest': str(tmpdir), 'state': 'directory'}, check=False)
+    Ansible('copy', {'src': str(config_file), 'dest': str(config_file)}, check=False)
+    Ansible('copy', {'src': str(passwd_file), 'dest': str(passwd_file)}, check=False)
+
+    secret_create_params = {
+        'name': secret_name,
+        'namespace': namespace,
+        'state': 'present',
+        'delete_after': 'yes',
+        'files': [
+            {
+                'name': 'config.yml',
+                'path': str(config_file)
+            },
+            {
+                'name': 'passwords',
+                'path': str(passwd_file)
+            }
+        ]
+    }
+    secret_create = Ansible('oc_secret', secret_create_params, check=False)
+    assert secret_create['changed'] is True
+    assert secret_create['results']['returncode'] == 0
+    assert not File(str(config_file)).exists
+    assert not File(str(passwd_file)).exists
+
+
+@pytest.mark.skip(msg="Need to find a way to test force arg for oc_secret")
+def test_force():
+    assert False
+
+
+@pytest.mark.skip(msg="Need to find a way to test kubeconfig arg for oc_secret")
+def test_kubeconfig(tmpdir, Ansible, Command, namespace):
+    assert False
+
+
+def test_modify_secret(secret, Ansible, test_config_file, test_passwd_file, tmpdir):
+    secret_list_params = {
+        'name': secret['name'],
+        'namespace': secret['namespace'],
+        'state': 'list'
+    }
+    secret_list = Ansible('oc_secret', secret_list_params)
+    assert secret_list['changed'] is False
+    assert secret_list['results']['returncode'] == 0
+    assert secret_list['results']['exists'] is True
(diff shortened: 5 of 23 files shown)


detiber:
@abutcher @sdodson @dgoodwin @smunilla @tbielawa @rhcarvalho ptal

detiber:
I still need to fix up existing tests and wire everything into the Makefile, but wanted to get some feedback on the current state. Also, the failure output from the setuptools molecule_tests command needs to be colorized and improved.

dgoodwin:
Really interesting, I assume we can do multiple containers to test against, is it possible we could use this for a more meaty role like openshift_master?

detiber:
@dgoodwin indeed, multiple containers is supported, and one of the reasons I went with Molecule (in addition to the Vagrant support, when Docker is too limiting for testing needs).  After I work out some of the rough edges, I plan on documenting a more substantial role (possibly etcd to start, since it isn't as complex as openshift_master).

detiber:
@mtnbikenc just realized I forgot to tag you earlier

detiber:
@rhcarvalho yes, I've been working on splitting out the tox-related changes from the molecule-related changes, see: #3007

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -118,6 +114,7 @@ def load_and_handle_cert(cert_string, now, base64decode=False):
 A 3-tuple of the form: (certificate_common_name, certificate_expiry_date, certificate_time_remaining)
 
     """
+    # pylint: disable=too-many-locals
```
detiber: Indeed, basically, I've discovered that at some point we stopped running pylint against some of these files, so these disables were mostly to unblock test failures that were blocking further testing.

roles/openshift_facts/molecule.yml
```diff
@@ -0,0 +1,20 @@
...
+  group_vars:
+    test_group:
+    - deployment_type: origin
+dependency:
+  name: galaxy
+driver:
+  name: docker
+docker:
+  containers:
+  - name: test_host
```
detiber: :+1: sounds good, as I get this futher along I will add a common openshift-ansible related prefix.

roles/openshift_facts/molecule.yml
```diff
@@ -0,0 +1,20 @@
...
+    image: centos/systemd
```
detiber: I wasn't able to get the tests that managed service manipulation working without using the systemd image.

roles/openshift_master_facts/tests/test_default.py
```diff
@@ -0,0 +1,11 @@
...
+
+testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner('.molecule/ansible_inventory').get_hosts('all')
+
+
+def test_hosts_file(File):
+    f = File('/etc/hosts')
+
+    assert f.exists
+    assert f.user == 'root'
+    assert f.group == 'root'
```
detiber: Yes, they were just placeholders to demonstrate how the testing works rather than real tests.

utils/test/openshift_ansible_tests.py
```diff
@@ -0,0 +1,77 @@
+# TODO: Temporarily disabled due to importing old code into openshift-ansible
+# repo. We will work on these over time.
+# pylint: disable=bad-continuation,missing-docstring,no-self-use,invalid-name
```
detiber: copy pasta, will clean it up.

utils/Makefile
```diff
@@ -91,19 +91,8 @@ ci-pylint: $(VENV)
...
-	@echo "#############################################"
-	@. $(VENV)/bin/activate && yamllint -c ../git/.yamllint $(YAMLFILES)
-
-ci-list-deps: $(VENV)
-	@echo "#############################################"
-	@echo "# Listing all pip deps"
-	@echo "#############################################"
-	. $(VENV)/bin/activate && pip freeze
+	. $(VENV)/bin/activate && tox -e py27-pylint
+	. $(VENV)/bin/activate && tox -e py35-pylint
```
detiber: Pylint does display different results in 2.7 vs 3.5. We could use detox and specify both environments at the same time to parallelize the tests.

utils/tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 minversion=2.3.1
 envlist =
-    py{27,35}-{flake8,unit}
+    py{27,35}-{flake8,unit,pylint}
```
detiber: Main reason was to allow for calling out the tests separately like we currently allow in the Makefile.

That said, I think it would be valuable to have many of the tests re-use the same virtualenv vs a virutalenv for each and every environment target.

detiber: I tested out re-using virtualenvs between envs and it causes issues when using detox to run tests in parallel. I think the benefits of the parallel runs outweighs the drawbacks of the extra virtualenvs.

I'm good with consolidating the envs to have a single test run that runs multiple tests, but then that is a feature regression from the existing Makefile.

detiber:
@kwoodson @mwoodson I've updated this PR to add in the molecule_common deduplication from https://github.com/openshift/openshift-ansible-contrib/pull/234 and also started to implement molecule/testinfra testing for the oc_secret module in lib_openshift. Would like your feedback on if I should continue down this path.

kwoodson:
@detiber,

Would love to see an implementation of molecule for testing lib_openshift.  On a side node, `pip install ruamel.yaml` should work in environments where we are pip installing other items.  If we have access to brew or the li repository then the package is called ruamel.yaml.  We are using it in our ops environment from an rpm.

detiber:
@kwoodson molecule testing is now working for oc_secret in lib_openshift, ptal and let me know your thoughts. You should be able to just checkout this branch and run `tox -e py27-ansible22-functional` and replicate locally.

detiber:
@tbielawa @rhcarvalho would like your opinions on testing changes in this PR, it introduces molecule testing, but also works on breaking out the travis config to be more parallel.

There are still some optimizations that need to be worked out for it, to bring down runtimes, but wanted to get feedback on the approach.

setup.py
```diff
@@ -226,6 +256,69 @@ def run(self):
...
+
+        molecule_dirs = find_dirs(starting_dir, self.excludes, self.roles, 'molecule')
+
+        print("Found:\n{0}".format(yaml.dump(molecule_dirs, default_flow_style=False)))
+        base_dir = os.getcwd()
+
+        errors = ""
+        warnings = ""
+
+        for role in molecule_dirs:
```
rhcarvalho: One of the reasons #2951 introduced Go to the repo is that it makes it so much easier and reliable to parallelize this loop of slow tests with clean reporting of errors, all based in Go's stdlib and `go test`.

setup.py
```diff
@@ -18,6 +20,34 @@
 from yamllint import linter
 
 
+def find_dirs(base_dir, exclude_dirs, include_dirs, dir_name):
```
rhcarvalho: What's the use case of such complex `find_dirs` function?
Looking through the existing code it seems we default to empty excludes and include the `roles` subdir, which doesn't seem to justify this much code.

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -6,7 +6,8 @@
     openshift_master_default_subdomain: "{{ osm_default_subdomain | default(None) }}"
   when: openshift_master_default_subdomain is not defined
 
-- fail:
+- name: Verify required variables are set
+  fail:
```
rhcarvalho: Could we please move the commits unrelated to molecule to a separate PR? E.g. adding names to tasks could easily be reviewed and merged independent of molecule, and would also make it easier to read and focus on what is necessary to add the molecule-based tests.

.travis.yml
```diff
@@ -9,12 +12,18 @@ python:
...
   - pip install -r requirements.txt
   - pip install tox-travis
 
 script:
   # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
   - ansible-playbook --syntax-check playbooks/byo/config.yml
-  - tox
-  - cd utils && tox
+  - if [[ "${TEST_DIR}" == "root" ]]; then tox; fi
+  - if [[ "${TEST_DIR}" == "utils" ]]; then cd utils && tox; fi
```
rhcarvalho: Isn't this getting too confusing? I liked better to run `make ci` locally. Using `env` might be good, but at the same time this is making it harder for people to understand what is running in Travis and being able to reproduce it locally before commit/push.

The `utils` feels like it wants to be it's own repo... can't we run the tests for `aos-installer` (utils) from the repository root?

molecule_common/base/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM centos/systemd:latest
```
rhcarvalho: Who is maintaining https://hub.docker.com/r/centos/systemd/? As a good practice, it would be good to depend on a tag other than `:latest` (e.g. `:7`), though it seems nothing other than `:latest` exist.

molecule_common/base/Dockerfile
```diff
@@ -0,0 +1,5 @@
...
+
+RUN yum makecache fast && yum update -y && \
+    yum install -y python sudo yum-plugin-ovl && \
```
rhcarvalho: `python` is already part of the base image.

molecule_common/base/Dockerfile
```diff
@@ -0,0 +1,5 @@
...
+    sed -i 's/plugins=0/plugins=1/g' /etc/yum.conf
```
rhcarvalho: I would suggest a more conservative regexp, `sed -i 's/^plugins=0$/plugins=1/g' /etc/yum.conf`, noting that sed will return 0 even if nothing is replaced.

If we'd want to fail the build if the replacement doesn't happen, then we can replace `sed` with `patch`, or `sed` + `grep`, or some other thing.

molecule_common/origin-latest/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM openshift/origin:latest
+
+RUN yum makecache fast && yum update -y && \
+    yum install -y python sudo yum-plugin-ovl && \
+    sed -i 's/plugins=0/plugins=1/g' /etc/yum.conf
```
rhcarvalho: Same as with the other Dockerfile:

- python is already in the base image;
- `'s/^plugins=0$/plugins=1/g'`
- missing a `yum clean all`?

molecule_common/base/Dockerfile
rhcarvalho: Unless we have a reason to keep the yum cache in the image, let's add a `yum clean all` at the end. E.g. https://github.com/sclorg/mongodb-container/blob/master/3.2/Dockerfile#L29

roles/lib_openshift/molecule/origin-latest/tests/test_oc_secret.py
```diff
@@ -0,0 +1,157 @@
+import pytest
```
detiber: pytest is brought in because of molecule, which uses testinfra as the default testing backend and is a plugin for pytest.

I do like the idea of changing to use pytest instead of nose for the test runner, but iirc we would need to refactor some of the existing tests and I didn't want to go down that road *yet*

test-requirements.txt
```diff
@@ -9,3 +9,6 @@ PyYAML
 yamllint
 nose
 coverage
+git+https://github.com/detiber/molecule@v2_tweaks
+pytest-xdist
```
detiber: I believe it is good enough for this purpose, it actually  helped uncover a bug with the ansible modules where we weren't using a copy of the kubeconfig.

I think there are some other improvements that could speed up test runs, and if that solves the runtime issue then I will remove the pytest-xdist dependency.

tox.ini
```diff
@@ -1,19 +1,27 @@
...
+    ansible22: ansible==2.2.0.0
 
 commands =
-    flake8: flake8
-    pylint: python setup.py lint
-    yamllint: python setup.py yamllint
+    lint: flake8
+    lint: python setup.py lint
+    lint: python setup.py yamllint
+    lint: python setup.py generate_validation
```
detiber: @rhcarvalho yes, generate_validation does indeed validate generated code is up to date.  I'm open to renaming.

I'm working on trying to do two things with the env reworking:
- improve travis parallelization
- improve general runtimes, I found that by adding wheel as a dependency then we are able to share pre-built packages from the pip cache.

tox.ini
```diff
@@ -1,19 +1,27 @@
...
     unit: nosetests
-    generate_validation: python setup.py generate_validation
+    py27-functional: python setup.py molecule_tests
+    py35-functional: echo "no py35 functional tests currently"
```
detiber: For now, yes. I haven't had a chance to contribute changes to make it py35 compatible. Even then, I suspect we might hit some Ansible issues, since they are still fixing some remaining py3 compat issues.

setup.py
detiber: As of now there is little purpose, but eventually I think we'd want to extend molecule testing to playbooks, and also to segregate tests into more consumable chunks. I can rip this out for now and we can re-introduce something when it's needed.

setup.py
```diff
@@ -226,6 +256,69 @@ def run(self):
...
+            print("Excludes:\n{0}".format(yaml.dump(self.excludes, default_flow_style=False)))
+
+        starting_dir = '.'
+        if self.roles is not None:
+            starting_dir = 'roles'
+
+        molecule_dirs = find_dirs(starting_dir, self.excludes, self.roles, 'molecule')
+
+        print("Found:\n{0}".format(yaml.dump(molecule_dirs, default_flow_style=False)))
+        base_dir = os.getcwd()
```
detiber: Because of the way I'm currently wrapping molecule it does indeed depend on a particular working directory. I expect this requirement to change in the future.

setup.py
detiber: It definitely needs to be cleaned up and parallelized better. I'd be hesitant to use a wrapper built with Go, unless we could assure that it would work under both Linux and Mac OSX.

roles/openshift_master_facts/tasks/main.yml
detiber: Yes, I'll work on getting that separated out today. This was mainly to unblock further testing.

.travis.yml
detiber: Agreed, mainly was trying to cleanup travis output to avoid the utils tox tests being the last result summary shown.

I still want to break them up into separate runs, but I agree it can be done much cleaner.

molecule_common/base/Dockerfile
detiber: Agreed, if we can find another (more stable) tag to use, I'll be happy to switch to it.

molecule_common/base/Dockerfile
detiber: Outside of the base image, the rest of the file is verbatim from Molecule. I will work on addressing the comments here and also upstream.

.travis.yml
detiber: We probably need to work out exactly how to properly test utils vs the rest of the repo. I agree that it feels like utils wants to be it's own repo, and it probably should be at some point.

@tbielawa would be interested in your thoughts on this...  Not really sure if we want to lump utils tests back in with the repo root tests, but not quite sure of a better way to make sure we are testing both properly through travis.

roles/lib_openshift/tasks/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
 - name: lib_openshift ensure python-ruamel-yaml package is on target
   package:
-    name: python-ruamel-yaml
+    name: python2-ruamel-yaml
```
detiber: This is a workaround until a proper python-ruamel-yaml image is available for CentOS.

roles/openshift_master_facts/tasks/main.yml
detiber: Created https://github.com/openshift/openshift-ansible/pull/3364 and https://github.com/openshift/openshift-ansible/pull/3365 to separate out the changes unrelated to molecule.

roles/lib_openshift/molecule/origin-latest/tests/test_oc_secret.py
rhcarvalho: Done in #3419 ;-)

setup.py
rhcarvalho: > I'd be hesitant to use a wrapper built with Go, unless we could assure that it would work under both Linux and Mac OSX.

Surely Go runs in Linux / macOS / Windows. Perhaps you just need to play a bit with Go... https://stevemilner.org/2014/07/05/in-my-mind-its-time-for-go/

molecule_common/base/Dockerfile
rhcarvalho: https://github.com/CentOS/CentOS-Dockerfiles/issues/124

setup.py
detiber: Indeed, but it is also compiled and the binaries are not portable between platforms.

If leveraging Go was common in the greater Ansible community it would also be less of an issue.

I also think there is a difference in leveraging Go in some places to provide tooling vs using Go for core parts of the implementation or the actual tests. I think straying away from python for those things will limit the ability for contributions from the greater community.

molecule_common/origin-latest/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM openshift/origin:latest
```
rhcarvalho: Perhaps we should be targeting an specific release version instead of `:latest`.

rhcarvalho:
Reading about Molecule again, and after `molecule init` on a role to see the scaffolding it generates, it seems that its test model is to have a [`single playbook.yml per role`](https://molecule.readthedocs.io/en/stable-1.20/configuration.html).

@detiber is that your understanding as well?

I'm wondering how to test roles that have different behavior depending on variables using Molecule. Even if in the playbook we call the roles multiple times, there are at least two problems:

- Tests are forcefully serialized (after all, it would all be part of a single playbook run).
- After the first role usage, the state of the images is dirty, and subsequent "tests" share state.

@detiber did you consider this case already? The support for "multiple tests" in Molecule seems to be after the playbook is run, not to verify multiple runs of the playbook with different system states (images) or variables/configuration.

Again, I might be wrong in all that and just not have understood Molecule well enough yet.

detiber:
> Reading about Molecule again, and after molecule init on a role to see the scaffolding it generates, it seems that its test model is to have a single playbook.yml per role.

> @detiber is that your understanding as well?

For v1, yes. v2 is under active development and has the concept of scenarios: http://molecule.readthedocs.io/en/v2/configuration.html#molecule.scenario.Scenario

> I'm wondering how to test roles that have different behavior depending on variables using Molecule. Even if in the playbook we call the roles multiple times, there are at least two problems:

> Tests are forcefully serialized (after all, it would all be part of a single playbook run).
After the first role usage, the state of the images is dirty, and subsequent "tests" share state.
@detiber did you consider this case already?

Yes, that was one of the big drivers behind jumping on board with v2. That said, scenarios are still currently serialized and is something that I want to work with upstream on improving, but in the meantime it is possible to test different configurations under the same scenario.

molecule.yml:
```
---
dependency:
  name: galaxy
driver:
  name: docker
lint:
  name: ansible-lint
platforms:
- name: test1
  groups:
  - common
  - orange
- name: test2
  groups:
  - common
  - red
provisioner:
  name: ansible
  group_vars:
    common:
      # common variables
      type: widget
    orange:
      color: orange
    red:
      color: red
scenario:
  name: default
verifier:
  name: testinfra
```

common_tests:
```
import testinfra.utils.ansible_runner


testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
    '.molecule/ansible_inventory.yml').get_hosts('common')


def test_type():
    # validate type is widget
    pass
```

orange_tests:
```
import testinfra.utils.ansible_runner


testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
    '.molecule/ansible_inventory.yml').get_hosts('orange')


def test_color():
    # validate color is orange
    pass
```

red_tests:
```
import testinfra.utils.ansible_runner


testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
    '.molecule/ansible_inventory.yml').get_hosts('red')


def test_color():
    # validate color is red
    pass
```

The example is a bit contrived, since it would be possible to test these differences without needing to have separate test files for each `color`, but it would be helpful for separate tests for more divergent configurations.

rhcarvalho:
@detiber, any ETA for this PR moving forward / Molecule v2?

I would like to use the same solution for testing the checks in openshift_health_checker.

detiber:
@rhcarvalho I'll get it updated in the next day or two.

roles/lib_openshift/molecule/origin-latest/molecule.yml
```diff
@@ -0,0 +1,35 @@
+---
```
detiber: Need to add ansible-doc testing for modules, not just in lib_openshift but for all modules.

michaelgugino:
Closing due to inactivity.

PR #3004, opened 2016-12-21
Do not hardcode openshift_image_tag

smunilla:
It is no longer necessary for us to hardcode openshift_image_tag to
'3.1.1.6' and we can now set it to 3.1.

Fixes Bug 1406713

### DIFF
--- a/utils/src/ooinstall/openshift_ansible.py
+++ b/utils/src/ooinstall/openshift_ansible.py
@@ -127,8 +127,7 @@ def write_inventory_vars(base_inventory, lb):
             "openshift_master_cluster_public_hostname={}\n".format(lb.public_hostname))
 
     if CFG.settings.get('variant_version', None) == '3.1':
-        # base_inventory.write('openshift_image_tag=v{}\n'.format(CFG.settings.get('variant_version')))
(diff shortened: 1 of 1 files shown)


sdodson:
@dgoodwin now that we have v3.1 tags available this is safe to do, correct?

utils/src/ooinstall/openshift_ansible.py
```diff
@@ -127,8 +127,7 @@ def write_inventory_vars(base_inventory, lb):
             "openshift_master_cluster_public_hostname={}\n".format(lb.public_hostname))
 
     if CFG.settings.get('variant_version', None) == '3.1':
-        # base_inventory.write('openshift_image_tag=v{}\n'.format(CFG.settings.get('variant_version')))
-        base_inventory.write('openshift_image_tag=v{}\n'.format('3.1.1.6'))
+        base_inventory.write('openshift_image_tag=v{}\n'.format(CFG.settings.get('variant_version')))
```
dgoodwin: Doesn't look quite right, I think you might want to be setting "openshift_release=v3.1", the image tag is meant to be precise.

PR #3018, opened 2016-12-26
Support openshift_node_port_range for configuring service NodePorts

smarterclayton:

Discovered while testing e2e runs against a real cluster.


DanyC97:
@smarterclayton i guess this PR can be closed as duplicate of #3023 ?

smarterclayton:
Moving to #3096

PR #3030, opened 2017-01-03
Ansible module for modifying yaml idempotently.

kwoodson:


kwoodson:
@twiest @abutcher @sdodson,

Here is a yaml editor module that our operations team uses.  We have recently updated this module to use an external library for yaml.  We packaged it and placed it in our channels.  It is ruamel.yaml.  There is a longer explanation as to why we want to use this version and not the stdlib's yaml editor.  Here are just a couple of the reasons:
- PyYaml only supports yaml 1.1 which came out in [2004.] ( [pyyaml issue](http://pyyaml.org/ticket/139) )
- Round trip loading and dumping (preserves spaces, comments, etc)
- ruamel.yaml is actively being updated and developed

Please let me know what you guys think.

This is also an example of our build pattern.  Edits are made to the files under build/ and then build/generate.py is run to produce library/yedit.py.  This allows us to use the source files of yedit in other ansible code without having to install them on the hosts we manage.

This pull request is a result of our discussion in mid December.

tbielawa:
@kwoodson can you elaborate on, or describe with an example, of how people are expected to use this contribution? I see that is does things, I don't fully understand why and how it does things. Example, I don't understand if I could/should migrate an existing role to use this file-parts pattern or not. What's the suggested uses? How to use?

Thanks Kenny

kwoodson:
@tbielawa,

Sure thing Tim.

The purpose of the module is very simple as it attempts to modify yaml files.  We use it in lots different places inside of our operations team.  The context in which we use it depends on exactly what is needed.

Here are a few example playbooks that use it:
- https://github.com/openshift/openshift-tools/blob/294d6bcf9a80df73dfb9eb2fd6a541cd2f2e0ac8/ansible/roles/openshift_gcp_set_zone/tasks/main.yml#L34-L39
- https://github.com/openshift/openshift-tools/blob/b81b248660cce89cf8c097b1558eefc60fcba8d6/ansible/roles/openshift_aws_iam_cert/tasks/main.yml#L31-L35

This is the basis to a few of our ansible modules that we have written for v3.  I used yedit as a base class to model other objects like deploymentconfig, route, secret, etc.  Since they are just a yaml representation under the covers, it made sense to me to model them as a yedit object.  Examples of that are here:

https://github.com/openshift/openshift-tools/blob/stg/ansible/roles/lib_openshift_3.2/build/lib/route.py#L57

Now the build pattern shown here is something that I have used on our operations team's modules in order to facilitate building, deploying, sharing module code.  If you needed to use the src/yedit.py code (only the yaml modifying code and not the ansible bits) in another ansible module then you could easily include it and generate your module with it.

Is it required?  No but I would like to see teams build on shared code so that modules can share the useful bits and obey the DRY principle.   This pattern is something we can further discuss.  The other way we can get this same functionality is by packaging these utilities.  We have done this in places where it makes sense but backed away since our future plans include moving to atomic.

tbielawa:
> Is it required? No but I would like to see teams build on shared code so that modules can share the useful bits and obey the DRY principle. This pattern is something we can further discuss. The other way we can get this same functionality is by packaging these utilities. We have done this in places where it makes sense but backed away since our future plans include moving to atomic.

@kwoodson :+1:

roles/lib_utils/build/ansible/yedit.py
```diff
@@ -0,0 +1,147 @@
...
+        try:
+            inc_value = yaml.load(inc_value)
+        except Exception as _:
+            raise YeditException('Could not determine type of incoming value. value=[%s] vtype=[%s]' \
+                                 % (type(inc_value), vtype))
+
+    return inc_value
+
+# pylint: disable=too-many-branches
+def main():
```
twiest: Change to use the unit testing pattern and include the current set of yedit unit tests.

roles/lib_utils/build/ansible/yedit.py
```diff
@@ -0,0 +1,147 @@
+#pylint: skip-file
```
twiest: Re-do the hierarchy to be:

```
src/
    ansible/
    lib/
    class/
    test/
    doc/
    generate.py
    
library/yedit.py
```

roles/lib_utils/build/generate.py
```diff
@@ -0,0 +1,44 @@
...
+          "#    / __| __| \| | __| _ \  /_\_   _| __|   \\\n"        + \
+          "#   | (_ | _|| .` | _||   / / _ \| | | _|| |) |\n"        + \
+          "#    \___|___|_|\_|___|_|_\/_/_\_\_|_|___|___/_ _____\n"  + \
+          "#   |   \ / _ \  | \| |/ _ \_   _| | __|   \_ _|_   _|\n" + \
+          "#   | |) | (_) | | .` | (_) || |   | _|| |) | |  | |\n"   + \
+          "#   |___/ \___/  |_|\_|\___/ |_|   |___|___/___| |_|\n"
+
+OPENSHIFT_ANSIBLE_PATH = os.path.dirname(os.path.realpath(__file__))
+
+FILES = {'yedit.py':  ['doc/license', 'doc/yedit', 'src/base.py', 'src/yedit.py', 'ansible/yedit.py'],
```
tbielawa: This here, respectfully, no no no.

Let's move the **data** that the generator code uses to build artifacts out of the **code**. The way I see it, this should be a simple data (yaml) file that the generator implicitly defaults to reading, or is provided as an option or other form of input (Most likely just defaulting to reading the source data)

For example: https://gist.github.com/tbielawa/abc2dde72cc51eb2f1c6ff01b6c780b0#file-generator-sources-yaml

Another thought about this, some of those file selections in the array could be implicitly determined from the to-generate-file name if you define a specific pattern for doing so. That is to say, if your `{'yedit.py'}` entry is basically empty then all the default files are automatically included.

Ex: You're generating 'yedit.py' with no cat-files included in the list, then the license file is included automatically, if a 'doc/yedit' named doc file exists then it is included. The base.py is included automanually. The 'src/yedit.py' and ansible/yedit.py' files are included automatically.

You would basically have an easy-mode for generating new artifacts that meet a standard concatenation pattern. See gist below for example:

https://gist.github.com/tbielawa/abc2dde72cc51eb2f1c6ff01b6c780b0#file-easy-mode-generator-yaml

roles/lib_utils/src/generate.py
```diff
@@ -0,0 +1,44 @@
...
+          "#    / __| __| \| | __| _ \  /_\_   _| __|   \\\n"        + \
+          "#   | (_ | _|| .` | _||   / / _ \| | | _|| |) |\n"        + \
+          "#    \___|___|_|\_|___|_|_\/_/_\_\_|_|___|___/_ _____\n"  + \
+          "#   |   \ / _ \  | \| |/ _ \_   _| | __|   \_ _|_   _|\n" + \
+          "#   | |) | (_) | | .` | (_) || |   | _|| |) | |  | |\n"   + \
+          "#   |___/ \___/  |_|\_|\___/ |_|   |___|___/___| |_|\n"
+
+OPENSHIFT_ANSIBLE_PATH = os.path.dirname(os.path.realpath(__file__))
+
+FILES = {'yedit.py':  ['doc/license', 'doc/yedit', 'class/base.py', 'class/yedit.py', 'ansible/yedit.py'],
```
tbielawa: re-commenting because I somehow commented on an out-dated version before...

----

This here, respectfully, no no no.

Let's move the **data** that the generator code uses to build artifacts out of the **code**. The way I see it, this should be a simple data (yaml) file that the generator implicitly defaults to reading, or is provided as an option or other form of input (Most likely just defaulting to reading the source data)

For example: https://gist.github.com/tbielawa/abc2dde72cc51eb2f1c6ff01b6c780b0#file-generator-sources-yaml

Another thought about this, some of those file selections in the array could be implicitly determined from the to-generate-file name if you define a specific pattern for doing so. That is to say, if your `{'yedit.py'}` entry is basically empty then all the default files are automatically included.

Ex: You're generating 'yedit.py' with no cat-files included in the list, then the license file is included automatically, if a 'doc/yedit' named doc file exists then it is included. The base.py is included automanually. The 'src/yedit.py' and ansible/yedit.py' files are included automatically.

You would basically have an easy-mode for generating new artifacts that meet a standard concatenation pattern. See gist below for example:

https://gist.github.com/tbielawa/abc2dde72cc51eb2f1c6ff01b6c780b0#file-easy-mode-generator-yaml

roles/lib_utils/build/generate.py
tbielawa: Also, updated the gist w/ an example of using extra licenses, etc.
https://gist.github.com/tbielawa/abc2dde72cc51eb2f1c6ff01b6c780b0#file-generate-special-license-yaml

roles/lib_utils/src/generate.py
kwoodson: @tbielawa,

Thanks for the feedback.  I'm fine with moving the generated files into a yaml file.  I'm a fan of auto-generating but in this case I'm not because of a couple of reasons.  

1.   I'd like for everyone who comes in to understand the order in which the files are generated and attempt to be transparent about what is happening.

1. I want it to be flexible.  If there are certain libraries that require certain sources, I don't want all files getting them.  Thus being explicit is what I'd prefer. 

The overall suggestion of putting them into a yaml file is fine.  I put them into a dictionary structure so ultimately putting them into a yaml is better for readability and keeps the dictionary format.

tbielawa: @kwoodson Totally! That's perfectly reasonable, the explicit vs. implicit decision. :+1: 

Re: `i put them into a dictionary structure`, we're on the same page with this one. It just makes sense to abstract that data from the code :smiley_cat:

kwoodson:
@tbielawa @abutcher @sdodson @twiest 
All requested changes have been made.  Please update on your status.  Is this good to merge?

tbielawa:
@kwoodson the `py27-flake8` target is failing on (pretty much?) everything under `roles/lib_utils/`

https://travis-ci.org/openshift/openshift-ansible/jobs/189624227#L1148

May want to consider adding pylint overrides or ignore directives for file fragments and only run the tests on the compiled modules. Thoughts on that, folks?

kwoodson:
@tbielawa,

Thanks, I'll take a look.  We use pylint with different settings.  This passes our pylint.

Are these flake rules hardset?  Are we really only allowed 79 characters?  

As an operations team we set ours at 120 and voted on it as a team.  I'll follow your conventions but I would propose that this be visited at some point.

tbielawa:
@kwoodson I'm not seeing any errors (did I miss it?) about you having any lines which are too long. I double checked [our length setting](https://github.com/openshift/openshift-ansible/blob/master/git/.pylintrc#L283-L286) for pylint and it is set to 120 characters.

As a helpful hint, I *strongly urge you* to move to using the `make ci` command under the `utils` dir to test the python code. That will execute our CI tests exactly as Jenkins and Travis execute them.

I just did that on the `kwoodson/yaml_editor` branch, checked out on my local machine, and the pylint errors came back exactly as the results show in travis/jenkins. You can do that yourself in the future to save yourself the waiting period during the turn-around from when you push your commits to the PR and when CI finally executes them. You'll get the results back instantly w/out ever having to push anything to GitHub.

Also @kwoodson, we're all on the same team, so if you think that some of the pylint rules we have now could be changed we can talk/vote about it just like you did before. Our automated testing right now is in a lot of churn so it's the best time to get settings worked out :smiley_cat: 

I'm going to do a quick test to see if I can get the lint check to exclude the class fragments and only examine the compiled bits.

tbielawa:
I added two excludes to the `utils/tox.ini` file and it took the reported number of lint errors down from 167 to a more realistic 86:

```diff
diff --git a/utils/tox.ini b/utils/tox.ini
index 747d79d..825818e 100644
--- a/utils/tox.ini
+++ b/utils/tox.ini
@@ -12,6 +12,6 @@ deps =
     py35-flake8: flake8-bugbear
 
 commands =
-    flake8: flake8 --config=setup.cfg ../ --exclude="../utils,.tox,../inventory"
+    flake8: flake8 --config=setup.cfg ../ --exclude="../utils,.tox,../inventory,../roles/lib_utils/src/class,../roles/lib_utils/src/test/"
     flake8: python setup.py flake8
     unit: python setup.py nosetests
```

https://gist.github.com/tbielawa/f058d1f6f1060ad139623fdfa80e1516

tbielawa:
Now as for why it still reports this error

> ../roles/lib_utils/library/yedit.py:458:28: F841 local variable '_' is assigned to but never used

I must assume it's some wacky lint setting we have now. As you know, `_` is how you TELL python that something is expanding from a set and **is never used**. But a quick peek at the code is telling me that we're catching exceptions and using the `as FOO` syntax without a reason to.

```
[~/rhat/openshift/openshift-ansible/roles/lib_utils/library] 9:01:19  (yaml_editor⚡)
$ grep -E 'except \w+ as .*' yedit.py
        except Exception as err:
        except KeyError as _:
        except KeyError as _:
        except KeyError as _:
        except KeyError as _:
        except KeyError as _:
        except KeyError as _:
        except KeyError as _:
        except Exception as _:
```

A quick little `sed` will fix those up for you:

> sed -i -r 's/(except \w+)( as _)/\1/' yedit.py

kwoodson:
@tbielawa @abutcher,

I have fixed up the flake8 and the pylint messages.

abutcher:
@kwoodson could you rebase / push to trigger the tests to rekick? This branch doesn't currently have a `tox.ini` so we might see some inconsistencies post merge. Rebasing and running `tox` outputs these errors/warnings:

```
roles/lib_utils/src/test/integration/yedit_test.yml:
  18:6      error    wrong indentation: expected 6 but found 5  (indentation)
  22:1      warning  comment not indented like content  (comments-indentation)
  40:1      warning  comment not indented like content  (comments-indentation)
  62:1      warning  comment not indented like content  (comments-indentation)
  88:1      warning  comment not indented like content  (comments-indentation)
  109:1     warning  comment not indented like content  (comments-indentation)
  130:1     warning  comment not indented like content  (comments-indentation)
  153:1     warning  comment not indented like content  (comments-indentation)
  173:1     warning  comment not indented like content  (comments-indentation)
```

kwoodson:
@abutcher,

Rebased and pushed.

abutcher:
ok to test

Issue #3026, opened 2017-01-02
Scaleup not working if the first master is broken


jkhelil:
[provide a description of the issue]
On a multi master install, if we loose the first master, we will no be able to scaleup the cluster by adding new nodes or new masters.

##### Version
[if you're operating from a git clone provide the output of `git describe`]
[if you're running from playbooks installed via RPM or atomic-openshift-utils `rpm -q atomic-openshift-utils openshift-ansible`]
[Your version of ansible, `ansible --version`]
ansible 2.2.0.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides

##### Steps To Reproduce
- Install multi  master cluster
-  destroy master-1 and update the inventory for the master list
-  Try to scaleup the cluster, I got this error
fatal: [ose3-int-a-master-2.node.xxx-1a-eu-central-1.acs]: FAILED! => {"changed": false, "failed": true, "msg": "openshift_facts received a lower value for openshift.master.master_count"}
ansible/playbooks/byo/openshift-master/scaleup.retry'.         [Errno 13]

If I comment the related code on openshift_facts module, I got an error relative to etcd_ca role
ASK [etcd_client_certificates : fail] *****************************************
fatal: [ose3-int-a-master-3.node.xxxx]: FAILED! => {"changed": false, "failed": true, "msg": "CA certificate /etc/etcd/ca/ca.crt doesn't exist on CA host ose3-int-a-master-2.node.xxxx. Apply 'etcd_ca' role to ose3-int-a-master-2.node.xxxx.\n"}

##### Current Result
fatal: [ose3-int-a-master-2.node.xxx-1a-eu-central-1.acs]: FAILED! => {"changed": false, "failed": true, "msg": "openshift_facts received a lower value for openshift.master.master_count"}
ansible/playbooks/byo/openshift-master/scaleup.retry'.         [Errno 13]
##### Expected Result
Successful Scaleup of the cluster
##### Additional Information
[The exact command you ran]
[Your operating system and version, ie: RHEL 7.2, Fedora 23]
[Your inventory file]
[visit https://docs.openshift.org/latest/welcome/index.html]

PR #2671, opened 2016-10-26
Restructure certificate redeploy playbooks

abutcher:
Restructure certificate redeploy playbooks to support redeploying certificates for control plane and nodes separately to give admins more control over certificate redeploy. This mainly allows node certificate redeploy to be ran individually against sets of nodes or nodes which may have failed as a result from original certificate redeployment.

~~These changes also replace the currently separate etcd CA certificate with the OpenShift CA following a redeploy and as the default for new clusters.~~

### OpenShift CA

Redeploys the OpenShift CA certificate by generating a new CA certificate and distributing an updated bundle to all components including client kubeconfigs and node CA trust.

Serial restarts: etcd, master services, node service and docker.

```
ansible-playbook -i inventory playbooks/byo/openshift-cluster/redeploy-openshift-ca.yml
```

### OpenShift Certificates

Redeploys etcd, master, node, router and registry certificates. This playbook does not redeploy the OpenShift CA certificate and uses the current CA certificate to sign new certificates.

Serial restarts: etcd, master services, and node service.

```
ansible-playbook -i inventory playbooks/byo/openshift-cluster/redeploy-certificates.yml
```

### Etcd Certificates

This playbook only redeploys etcd certificates including master client certificates.

Serial restarts: etcd and master services.

```
ansible-playbook -i inventory playbooks/byo/openshift-cluster/redeploy-etcd-certificates.yml
```

### Master Certificates

This playbook only redeploys master certificates.

Serial restarts: master services.

```
ansible-playbook -i inventory playbooks/byo/openshift-cluster/redeploy-master-certificates.yml
```

### Node Certificates

This playbook only redeploys node certificates.

Serial restarts: node services.

```
ansible-playbook -i inventory playbooks/byo/openshift-cluster/redeploy-node-certificates.yml
```

### Registry and Router Certificates

These playbooks replace installer created certificates for the router and registry. If custom certificates are in use for these components they must be replaced manually.

```
ansible-playbook -i inventory playbooks/byo/openshift-cluster/redeploy-router-certificates.yml
ansible-playbook -i inventory playbooks/byo/openshift-cluster/redeploy-registry-certificates.yml
```

PR #3031, opened 2017-01-03
Sync etcd ca certs from etcd_ca_host to other etcd hosts

jkhelil:
When the etcd_ca_host (In most cases this is the first master) is broken, we can not use scaleup playbook, it exits with this error 
TASK [etcd_client_certificates : fail] *****************************************
fatal: [ose3-int-a-master-3.node.xxxx]: FAILED! => {"changed": false, "failed": true, "msg": "CA certificate /etc/etcd/ca/ca.crt doesn't exist on CA host ose3-int-a-master-2.node.xxxx. Apply 'etcd_ca' role to ose3-int-a-master-2.node.xxxx.\n"}
This is also true for node scaleup playbook
This is troublesome  and, we should ensure etcd ca certs are replicated to other etcd hosts
related to https://github.com/openshift/openshift-ansible/issues/3026

### DIFF
--- a/roles/etcd_server_certificates/tasks/main.yml
+++ b/roles/etcd_server_certificates/tasks/main.yml
@@ -142,6 +142,38 @@
     dest: "{{ etcd_cert_config_dir }}"
   when: etcd_server_certs_missing | bool
 
+- name: Create a tarball of the etcd ca certs
+  command: >
+    tar -czvf {{ etcd_generated_certs_dir }}/{{ etcd_ca_name }}.tgz
+      -C {{ etcd_ca_dir }} .
+  args:
+    creates: "{{ etcd_generated_certs_dir }}/{{ etcd_ca_name }}.tgz"
+    warn: no
+  when: etcd_server_certs_missing | bool
+  delegate_to: "{{ etcd_ca_host }}"
+
+- name: Retrieve etcd ca cert tarball
+  fetch:
+    src: "{{ etcd_generated_certs_dir }}/{{ etcd_ca_name }}.tgz"
+    dest: "{{ g_etcd_server_mktemp.stdout }}/"
+    flat: yes
+    fail_on_missing: yes
+    validate_checksum: yes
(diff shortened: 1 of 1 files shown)


abutcher:
Hey @jkhelil, I'm investigating using the OpenShift CA for signing etcd certificates as opposed to maintaining a separate CA in https://github.com/openshift/openshift-ansible/pull/2671. I'm still dealing with serial issues but it should be possible and would take care of this problem if it works out.

jkhelil:
Hi @abutcher , That's good. Can you confirm plz that in case of broken master 0, we will not have an issue when we'd want to scaleup nodes or master,  because this test https://github.com/openshift/openshift-ansible/blob/master/roles/etcd_client_certificates/tasks/main.yml#L9   is failing whenever it does not find the ca in the ''new" master 0 ?

abutcher:
@jkhelil Right, if we're using the OpenShift CA then the etcd CA host will be a master and each master should have everything it needs to sign new certificates. I'll be sure to test this.

jkhelil:
@abutcher, Hi Andrew, We need etcd certificates sync on other hosts, Do you know if you know when you work on certificates will lands ?
Do you think is possible to integrate my PR,a and removed once you work is done, if you think it will not lands soon ?

abutcher:
@jkhelil It may be a bit longer so I'm open to merging. Could you take a look into the travis failure?

jkhelil:
@abutcher fixed, PTAL

jkhelil:
@abutcher Do you think you can review and merge the PR?

detiber:
aos-ci-test

abutcher:
@detiber I'm not sure how we can deal with the serial drift other than making sure that the serial file is replicated from the first etcd host during cert redeploy/scaleup.

detiber:
@abutcher I like that idea,that should solve the automated use-cases at least...  would still potentially hit issues if users manually create certs, but we can call that out in the docs.

PR #3036, opened 2017-01-04
Fix jsonpath expected output when checking registry volume secrets

soltysh:
In k8s 1.5 (see rebase PR: origin 12143) jsonpath output is slightly different than previously. Causing this line to fail. We need to pick `secret` or `secretName` when checking this task failure condition.

@dgoodwin mind checking if this is the right approach, this is needed for the rebase PR https://github.com/openshift/origin/pull/12143

@ncdc @smarterclayton fyi

### DIFF
--- a/roles/openshift_hosted/tasks/registry/secure.yml
+++ b/roles/openshift_hosted/tasks/registry/secure.yml
@@ -34,13 +34,12 @@
 - name: Create registry certificates if they do not exist
   command: >
     {{ openshift.common.client_binary }} adm ca create-server-cert
-    --signer-cert=/etc/origin/master/ca.crt
-    --signer-key=/etc/origin/master/ca.key
-    --signer-serial=/etc/origin/master/ca.serial.txt
+    --signer-cert={{ openshift_master_config_dir }}/ca.crt
+    --signer-key={{ openshift_master_config_dir }}/ca.key
+    --signer-serial={{ openshift_master_config_dir }}/ca.serial.txt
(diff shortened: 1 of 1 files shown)


dgoodwin:
LGTM @soltysh

smarterclayton:
So this isn't backwards compatible with old clients?

smarterclayton:
Ansible needs to be able to work across 'oc' versions

smarterclayton:
I guess this will work for both.

smarterclayton:
Can you also backport this to the 1.4 branch

soltysh:
> So this isn't backwards compatible with old clients?

It's written as or, so it'll expect either `secretName not found` or `secret not found` and should work with older clients.

> Can you also backport this to the 1.4 branch

On it.

liggitt:
isn't the right fix to filter the list of inspected volumes to ones that reference secrets, not tolerate jsonpath errors on other volume types?

liggitt:
```
{{ openshift.common.client_binary }} get dc/docker-registry
     {{ openshift.common.client_binary }} get dc/docker-registry
    -o jsonpath='{.spec.template.spec.volumes[?(@.secret)].secret.secretName}
```

soltysh:
Hmmm... that would make sense and would be much more readable (sort of 😉) but that'd require some change in this task so that it works as expected. I need to dig a bit deeper into ansible to fix it properly, in that case. Will work on in tomorrow morning.

soltysh:
I almost forgot, I've also changed hard coded paths to `ca.crt` to use variable.

soltysh:
Hmmm... the failure doesn't look like it's related to my changes, @sdodson is it?

sdodson:
Nope, CI infra problem.

Issue #3022, opened 2016-12-28
Install default CNI plugins when os_sdn_network_plugin_name=cni?


ozdanborne:
According to the [k8s docs](http://kubernetes.io/docs/admin/network-plugins/#cni):

>Kubernetes requires the standard CNI `lo` plugin, at minimum version 0.2.0

I'm thinking it would make sense for openshift-ansible to download and install the `loopback` plugin when `os_sdn_network_plugin_name=cni`. Or even better - download all the default plugins?

sdodson:
@djosborne sorry for the delay in getting to this, but loopback plugin is a standard kubernetes CNI plugin?
@dcbw Is this something we should be shipping in our openshift packages?

djosborne:
@sdodson Yes. For example, `registry-console-1-deploy` fails to deploy when networking mode is set to CNI:

`oc describe pod registry-console-1-deploy` shows the following:
```
Error syncing pod, skipping: failed to "SetupNetwork" for "registry-console-1-deploy_default" with SetupNetworkError: "Failed to setup network for pod \"registry-console-1-deploy_default(e89d0b14-0391-11e7-a4da-42010af0003e)\" using network plugins \"cni\": failed to find plugin \"loopback\" in path [/opt/loopback/bin /opt/cni/bin]; Skipping pod"
```

PR #3038, opened 2017-01-04
Add Calico as a Networking Option

djosborne:
Looking for feedback on this PR to add Calico as an optional networking option for Openshift.

There are still a few missing pieces and some bugs, so I don't recommend anyone attempt to run this yet.

Isolation isn't working in this yet as the policy controller is not started. I will work towards adding it but just wanted to make sure my general approach was acceptable before doing so.

### DIFF
--- a/roles/calico/tasks/main.yml
+++ b/roles/calico/tasks/main.yml
@@ -1,0 +1,74 @@
+---
+- include: ../../../roles/etcd_client_certificates/tasks/main.yml
+  vars:
+    etcd_cert_prefix: calico.etcd-
+    etcd_cert_config_dir: "{{ openshift.common.config_base }}/calico"
+    embedded_etcd: "{{ hostvars[groups.oo_first_master.0].openshift.master.embedded_etcd }}"
+    etcd_ca_host: "{{ groups.oo_etcd_to_config.0 }}"
+    etcd_cert_subdir: "openshift-calico-{{ openshift.common.hostname }}"
+
+- name: Assure the calico certs have been generated
+  stat:
+    path: "{{ item }}"
+  with_items:
+  - "{{ calico_etcd_ca_cert_file }}"
+  - "{{ calico_etcd_cert_file}}"
+  - "{{ calico_etcd_key_file }}"
+
+- name: Configure Calico service unit file
+  template:
+    dest: "/lib/systemd/system/calico.service"
+    src: calico.service.j2
+
+- name: Enable calico
+  become: yes
+  systemd:
+    name: calico
+    daemon_reload: yes
+    state: started
+    enabled: yes
+  register: start_result
+
+- name: Assure CNI conf dir exists
+  become: yes
+  file: path="{{ cni_conf_dir }}" state=directory
+
+- name: Generate Calico CNI config
+  become: yes
+  template:
+    src: "calico.conf.j2"
+    dest: "{{ cni_conf_dir }}/10-calico.conf"
+
+- name: Assures Kuberentes CNI bin dir exists
+  become: yes
+  file: path="{{ cni_bin_dir }}" state=directory
+
+- name: Download Calico CNI Plugin
+  become: yes
+  get_url:
+    url: https://github.com/projectcalico/cni-plugin/releases/download/v1.5.5/calico
+    dest: "{{ cni_bin_dir }}"
+    mode: a+x
+
+- name: Download Calico IPAM Plugin
+  become: yes
+  get_url:
+    url: https://github.com/projectcalico/cni-plugin/releases/download/v1.5.5/calico-ipam
+    dest: "{{ cni_bin_dir }}"
+    mode: a+x
+
+- name: Download and unzip standard CNI plugins
+  become: yes
+  unarchive:
+    remote_src: True
+    src: https://github.com/containernetworking/cni/releases/download/v0.4.0/cni-amd64-v0.4.0.tgz
+    dest: "{{ cni_bin_dir }}"
+
+- name: Assure Calico conf dir exists
+  become: yes
+  file: path=/etc/calico/ state=directory
+
+- name: Set calicoctl.cfg
+  template:
+    src: calico.cfg.j2
+    dest: "/etc/calico/calicoctl.cfg"
--- a/roles/calico/templates/calico.service.j2
+++ b/roles/calico/templates/calico.service.j2
@@ -1,0 +1,29 @@
+[Unit]
+Description=calico
+After=docker.service
+Requires=docker.service
+
+[Service]
+Restart=always
+ExecStartPre=-/usr/bin/docker rm -f calico-node
+ExecStart=/usr/bin/docker run --net=host --privileged \
+ --name=calico-node \
+ -e WAIT_FOR_DATASTORE=true \
+ -e FELIX_DEFAULTENDPOINTTOHOSTACTION=ACCEPT \
+ -e CALICO_IPV4POOL_IPIP=always \
+ -e FELIX_IPV6SUPPORT=false \
+ -e ETCD_ENDPOINTS={{ etcd_endpoints }} \
+ -v /etc/origin/calico:/etc/origin/calico \
+ -e ETCD_CA_CERT_FILE={{ calico_etcd_ca_cert_file }} \
+ -e ETCD_CERT_FILE={{ calico_etcd_cert_file }} \
+ -e ETCD_KEY_FILE={{ calico_etcd_key_file }} \
+ -e NODENAME={{ openshift.common.hostname }} \
+ -v /var/log/calico:/var/log/calico \
+ -v /lib/modules:/lib/modules \
+ -v /var/run/calico:/var/run/calico \
+ calico/node:v1.1.0
+
+ExecStop=-/usr/bin/docker stop calico-node
+
+[Install]
+WantedBy=multi-user.target
--- a/roles/calico_master/tasks/main.yml
+++ b/roles/calico_master/tasks/main.yml
@@ -1,0 +1,41 @@
+---
+- name: Assure the calico certs have been generated
+  stat:
+    path: "{{ item }}"
+  with_items:
+  - "{{ calico_etcd_ca_cert_file }}"
+  - "{{ calico_etcd_cert_file}}"
+  - "{{ calico_etcd_key_file }}"
+
+- name: Create temp directory for policy controller definition
+  command: mktemp -d /tmp/openshift-ansible-XXXXXXX
+  register: mktemp
+  changed_when: False
+
+- name: Write Calico Policy Controller definition
+  template:
+    dest: "{{ mktemp.stdout }}/calico-policy-controller.yml"
+    src: calico-policy-controller.yml.j2
+
+- name: Launch Calico Policy Controller
+  command: >
+    {{ openshift.common.client_binary }} create
+    -f {{ mktemp.stdout }}/calico-policy-controller.yml
+    --config={{ openshift.common.config_base }}/master/admin.kubeconfig
+  register: calico_create_output
+  failed_when: ('already exists' not in calico_create_output.stderr) and ('created' not in calico_create_output.stdout)
+  changed_when: ('created' in calico_create_output.stdout)
+
+- name: Delete temp directory
+  file:
+    name: "{{ mktemp.stdout }}"
+    state: absent
+  changed_when: False
+
+
+- name: oc adm policy add-scc-to-user privileged system:serviceaccount:kube-system:calico
+  oc_adm_policy_user:
+    user: system:serviceaccount:kube-system:calico
+    resource_kind: scc
+    resource_name: privileged
+    state: present
--- a/roles/etcd_client_certificates/tasks/main.yml
+++ b/roles/etcd_client_certificates/tasks/main.yml
@@ -51,7 +51,7 @@
     creates: "{{ etcd_generated_certs_dir ~ '/' ~  etcd_cert_subdir ~ '/'
                  ~ etcd_cert_prefix ~ 'client.csr' }}"
   environment:
-    SAN: "IP:{{ etcd_ip }}"
+    SAN: "IP:{{ etcd_ip }},DNS:{{ etcd_hostname }}"
   when: etcd_client_certs_missing | bool
   delegate_to: "{{ etcd_ca_host }}"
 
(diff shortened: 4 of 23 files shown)


sdodson:
FYI @rajatchopra

rajatchopra:
As per a review meeting, the following three items need attention (in no particular order of severity/priority):

1. Make sure the use_calico option is mutually exclusive
2. Create a separate task for calico-master for the policy+controller service
3. Eventually move off to having a separate etcd instance (do not use the one that openshift creates)

@djosborne Thanks for the PR. Looking forward to getting this merged. Let me know if I missed anything.

djosborne:
@rajatchopra I'm trying to implement the 2nd suggestion (split into master / common role). But for some reason when I run with [the 5th commit in this PR](https://github.com/openshift/openshift-ansible/pull/3038/commits/8b98e4ded0060817294c8c6be731a7d9204826e3), the etcd certs become empty directories instead of ... certificates. I think it has to do with [the variables set in master/config.yml](https://github.com/openshift/openshift-ansible/pull/3038/commits/8b98e4ded0060817294c8c6be731a7d9204826e3#diff-385d17203ec610a738ee3ec5938ce287R177), but I'm not sure what I'm doing wrong.

I'm running this on a cluster where the host is both a master and node.

roles/etcd_client_certificates/tasks/main.yml
```diff
@@ -1,4 +1,24 @@
 ---
+- name: DEBUG etcd_cert_config_dir
```
jcpowermac: @djosborne not sure if you caught this but you brought over some of my debug / pause actions.

djosborne:
@rajatchopra This PR should be ready for you to take a look. Here's some information on some  caveats:

### Sample Inventory.ini
Tested with Openshift origin v1.3
```
[OSEv3:children]
masters
nodes
etcd

[OSEv3:vars]
os_sdn_network_plugin_name=cni
openshift_use_openshift_sdn=false
openshift_use_calico=true

[masters]
master1 ansible_host=104.196.229.127

[nodes]
master1 openshift_node_labels="{'region': 'infra'}" 
node2 ansible_host=104.198.3.155

[etcd]
db1 ansible_host=104.196.253.167
```

### External etcd
This calico playbook only supports explicitly defined etcd hosts. I modeled it after the flannel playbook role which has the same requirement.

### Generating etcd Client Certs
The newly introduced Calico playbook role lists `openshift_etcd_client_certificates` as a dependency. But (with the help of @jcpowermac), I discovered that on master hosts, generation of the certs wasn't happening until _after_ calico's playbook role was executed. Because of this, Calico was volume mounting non-existent certs into it's docker container. Docker's default behavior is to then create empty directories on the host in this scenario.

To circumvent this, I modified the role to have the following behavior:

- Node-only hosts successfully generate etcd client certs which Calico will use.
- Master-only and master+node hosts configure Calico to use the existing etcd client certs on the host. Calico certs are still generated for these hosts afterwards, but are not used.

The correct fix here is probably to fix etcd client cert generation so that it happens before the Calico role. But I don't want to waste too much time on this since long term we will likely switch to Calico's "etcdless" backend which uses the kubernetes apiserver to store data, removing the dependency on etcd altogether.

### Etcd cert SAN field
[According to digicert](https://www.digicert.com/subject-alternative-name-compatibility.htm): 
>If a SSL Certificate has a Subject Alternative Name (SAN) field, then SSL clients are supposed to ignore the Common Name value and seek a match in the SAN list. This is why DigiCert always repeats the common name as the first SAN in our certificates.

I found that the SSL libraries Calico uses follows this same paradigm, and was skipping the SN of the etcd cert and jumping directly to the SAN, which only specified the IP of etcd. I configured Calico to connect to etcd using its hostname, so I added the hostname in the SAN field. 

### Other
- I haven't tested with multi-master yet. 
- I'm not completely satisfied with how the `calico` and `calico_master` role are split up / depend on each other. I think I'd like to introduce `calico_certs` and `calico_common` roles to better divy up, but figured since this "works" it'd be best to hold off for now.

zhwenh:
hello , I "git clone https://github.com/djosborne/openshift-ansible/tree/calico  " this version, and setup show this error:
TASK [openshift_ca : Create the master certificates if they do not already exist] ***
fatal: [10.1.130.15 -> 10.1.130.15]: FAILED! => {"changed": true, "cmd": ["/usr/local/bin/oc", "adm", "create-master-certs", "--hostnames=kubernetes.default,kubernetes.default.svc.cluster.local,kubernetes,openshift.default,openshift.default.svc,172.30.0.1,10.1.130.15,openshift.default.svc.cluster.local,kubernetes.default.svc,openshift", "--master=https://10.1.130.15:8443", "--public-master=https://10.1.130.15:8443", "--cert-dir=/etc/origin/master", "--overwrite=false"], "delta": "0:00:03.154188", "end": "2017-02-06 16:29:33.655000", "failed": true, "rc": 1, "start": "2017-02-06 16:29:30.500812", "stderr": "Command \"create-master-certs\" is deprecated, Use 'oc adm ca' instead.\nopen /etc/origin/master/master.etcd-client.crt: is a directory", "stdout": "Generated new key pair as /etc/origin/master/serviceaccounts.public.key and /etc/origin/master/serviceaccounts.private.key", "stdout_lines": ["Generated new key pair as /etc/origin/master/serviceaccounts.public.key and /etc/origin/master/serviceaccounts.private.key"], "warnings": []}

        to retry, use: --limit @/root/ansible-calico/openshift-ansible-calico/playbooks/byo/config.retry

**my hosts :**
[OSEv3:children]
masters
nodes
etcd
# Set variables common for all OSEv3 hosts
[OSEv3:vars]
# SSH user, this user should allow ssh based auth without requiring a password
openshift_master_hostnames=10.1.130.15
ansible_ssh_user=root
openshift_clock_enabled=true
openshift_install_examples=true
#deployment_subtype=registry
deployment_type=origin

openshift_master_default_subdomain=svc.cluster.local
osm_cluster_network_cidr=172.88.0.0/16
openshift_release=v1.3.1
openshift_image_tag=v1.3.1
openshift_use_dnsmasq=true
openshift_pkg_version=-1.3.1
openshift_dns_ip=172.30.0.1
openshift_docker_options="--insecure-registry 172.30.0.0/16 --insecure-registry 172.88.0.0/16 --insecure-registry  10.1.0.0/16 --ipv6=false"
openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 'challenge': 'true', 'kind': 'HTPasswdPasswordIdentityProvider', 'filename': '/etc/origin/master/htpasswd'}]
openshift_additional_repos=[{'id': 'origin-release', 'name': 'origin-release', 'baseurl': 'http://cbs.centos.org/repos/paas7-openshift-origin-release/x86_64/os/','enabled': 1, 'gpgcheck': 0}]
os_sdn_network_plugin_name=cni
openshift_use_openshift_sdn=false
openshift_use_calico=true

# host group for masters
[masters]
10.1.130.15

[etcd]
10.1.130.15
# host group for nodes, includes region info
[nodes]
10.1.130.15 openshift_node_labels="{'nodetype': 'db'}" 
10.1.130.16 openshift_node_labels="{'nodetype': 'normal'}"
10.1.130.14 openshift_node_labels="{'nodetype': 'normal'}"

DanyC97:
@zhwenh you mind if you edit your last comment and format the output into block code so it can be easily read?

DanyC97:
will be nice to make merge this work, not sure what else is needed though

djosborne:
Rebasing onto latest head of `openshift-ansible` is causing connectivity issues when the cluster comes up with Calico. I'm investigating this week.

djosborne:
Hey all, I've fixed up this PR and believe it should work. Tested with Openshift 1.4 and it seems to be working.

One of the fixes required for 1.4 was ensuring the CNI loopback plugin was installed. I added this commit (https://github.com/openshift/openshift-ansible/pull/3038/commits/b183d971f693eb6c62a0da94c3a6862cb7c41ebe) but we should really come up with a more sensible way to resolve this correctly. I raised an issue at #3022 )

roles/calico_master/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+  changed_when: ('created' in calico_create_output.stdout)
+
+- name: Delete temp directory
+  file:
+    name: "{{ mktemp.stdout }}"
+    state: absent
+  changed_when: False
+
+
+- name: Set permissions for calico
```
jcpowermac: @djosborne can you replace this `command` with the new `lib_openshift` module actions?
Example here: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/tasks/registry/registry.yml#L60

https://github.com/openshift/openshift-ansible/tree/master/roles/lib_openshift/library

jcpowermac:
@djosborne once we get closer to approving this PR can this be squashed?

smarterclayton:
Is this using the master etcd?  If so the only access allowed will be from the master nodes - we don't let other clients talk directly to etcd - a separate etcd would be required.

djosborne:
@smarterclayton - This does access master etcd from the nodes. @rajatchopra originally OK'd this as a short term solution. Instead of doing the work to roll our own etcd, I plan to jump directly to using the Kuberentes Apiserver as a backend, removing Calico's dependency on etcd altogether. Planning this as follow-up work.

I'm doing some final testing but would love to see this get in as a first increment this week. Can someone provide some insight on why the travis-ci build is failing? I can't understand what is causing it.

I've live-demoed to @jcpowermac and addressed his PR feedback. What else is needed to get this in?

smarterclayton:
I don't know that we can recommend that as anything other than a trial or demo setup.  We don't allow that for other components, and so it's pretty important that this be:

a) impossible to enable without understanding the security implications
b) not be copied to other places in the code until it is removed within a reasonably short time window 
c) not listed as a recommended or supported configuration

@liggitt @sdodson Scott can you make sure that there is the appropriate security and controls around this so that it does not get kept as is or allowed to escape the very limited circumstances here?  Also a timeframe for this to move to not having direct access.

jcpowermac:
@djosborne looks like the reason travis is failing is:
`./roles/openshift_facts/library/openshift_facts.py:469:1: E302 expected 2 blank lines, found 1`

djosborne:
@smarterclayton 

>I don't know that we can recommend that as anything other than a trial or demo setup.

That is fine. The intent here is to get a base integration working. This work has been going on since before 1.4 was released and has taken unexpetedly long to get working. I'd really like to see this merged as a base working case with the intent of immediately iterating on and improving it.

jcpowermac:
@djosborne maybe remove WIP so the maintainers know this is ready.
@sdodson or @abutcher can you please review.

rajatchopra:
@djosborne The PR looks fine to me. Two things:

1. We are planning to restructure the ansible repo for plugins. So the way this is done is likely going to be rejigged. This PR should nevertheless get merged as the re-architecture starts, but expect to work on the new model soon.
2. The etcd being re-used is a non-flyer for production as you have understood. Could you add a README to the calico folder indicating that its not intended for production. Also link to some docs summarizing calico plugin so that any community user has at least some clues on what/how/why things work with the calico plugin.

djosborne:
@rajatchopra I've added the warning, and opened an issue at Calico Docs to add OpenShift documentation resources. Will move that along ASAP

roles/calico/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+    etcd_ca_host: "{{ groups.oo_etcd_to_config.0 }}"
+    etcd_cert_subdir: "openshift-calico-{{ openshift.common.hostname }}"
+
+- name: Assure the calico certs have been generated
+  stat:
+    path: "{{ item }}"
+  with_items:
+  - "{{ calico_etcd_ca_cert_file }}"
+  - "{{ calico_etcd_cert_file}}"
+  - "{{ calico_etcd_key_file }}"
```
sdodson: You're not actually creating these, you're requiring that they exist prior to running ansible? Perhaps document how you'd expect to create them in README.md?

roles/calico_master/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+- name: Assure the calico certs have been generated
+  stat:
+    path: "{{ item }}"
+  with_items:
+  - "{{ calico_etcd_ca_cert_file }}"
+  - "{{ calico_etcd_cert_file}}"
+  - "{{ calico_etcd_key_file }}"
```
sdodson: Likewise, document how to create these?

roles/calico/templates/calico.service.j2
```diff
@@ -0,0 +1,29 @@
...
+ -e ETCD_ENDPOINTS={{ etcd_endpoints }} \
+ -v /etc/origin/calico:/etc/origin/calico \
+ -e ETCD_CA_CERT_FILE={{ calico_etcd_ca_cert_file }} \
+ -e ETCD_CERT_FILE={{ calico_etcd_cert_file }} \
+ -e ETCD_KEY_FILE={{ calico_etcd_key_file }} \
+ -e NODENAME={{ openshift.common.hostname }} \
+ -v /var/log/calico:/var/log/calico \
+ -v /lib/modules:/lib/modules \
+ -v /var/run/calico:/var/run/calico \
+ calico/node:v1.1.0
```
sdodson: Is this on quay.io like calico-policy-controller image? If so you'll need to specify it as it's not included in our default docker config.

sdodson:
aos-ci-test

sdodson:
[test]

sdodson:
At some point in the future origin will support arm64 and ppc64le architectures. Which architectures is calico built for?

sdodson:
[merge]

zhwenh:
calico network failed to "SetupNetwork" 
 
origin v1.5.0-rc.0 Version + Calico + cni

Steps To Reproduce

router-3-deploy_default pod run show error message

Current Result

Expected Result

Additional Information

Error syncing pod, skipping: failed to "SetupNetwork" for "router-3-deploy_default" with SetupNetworkError: "Failed to setup network for pod "router-3-deploy_default(799d976f-1502-11e7-b3c1-fa163e10ef43)" using network plugins "cni": failed to send CNI request: Post http://dummy/: dial unix /var/run/openshift-sdn/cni-server.sock: connect: no such file or directory; Skipping pod"	39 times in the last 0 seconds

sdodson:
@zhwenh Can you please open an issue for that and mention @djosborne ? Your complete inventory would be helpful too.

PR #3040, opened 2017-01-04
Enable repos defined in openshift_additional_repos by default

### LINKED ISSUE #3025: openshift_additional_repos does not appear to default correctly
Looking at the template for `openshift_additional_repos`, it appears the intent would be that if enabled is not present in the array that it will be turned on (`| default('1')`).  However, in practice:

```
openshift_additional_repos: [{'id': 'origin-pr', 'baseurl': '{{ lookup("env", "PR_REPO_URL") }}', 'gpgcheck': 0}]
```

results in enabled=0 on disk.

sdodson:
Fixes #3025

### DIFF
--- a/roles/openshift_repos/templates/yum_repo.j2
+++ b/roles/openshift_repos/templates/yum_repo.j2
@@ -2,9 +2,9 @@
 [{{ repo.id }}]
 name={{ repo.name | default(repo.id) }}
 baseurl={{ repo.baseurl }}
-{% set enable_repo = repo.enabled | default('1') %}
+{% set enable_repo = repo.enabled | default(1,True) %}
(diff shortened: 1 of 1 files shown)


sdodson:
aos-ci-test

roles/openshift_repos/templates/yum_repo.j2
```diff
@@ -2,7 +2,7 @@
 [{{ repo.id }}]
 name={{ repo.name | default(repo.id) }}
 baseurl={{ repo.baseurl }}
-{% set enable_repo = repo.enabled | default('1') %}
+{% set enable_repo = repo.enabled | default(1,True) %}
```
detiber: Should gpgcheck be updated in the same way?

sdodson: yeah, good catch, updated

PR #3041, opened 2017-01-04
[WIP] Forward requests for PTR records to kube as well

### LINKED ISSUE #3017: dnsmasq does not serve PTR records, making nodes not compliant with Kube DNS
The default dnsmasq configuration on the node does not appear to support PTR records - I discovered this while debugging e2e failures against a GCE cluster.

```
# on the master, direct to skydns
$ dig @localhost -p 8053 +tcp +search 89.65.30.172.in-addr.arpa. PTR
...
;; ANSWER SECTION:
89.65.30.172.in-addr.arpa. 30	IN	PTR	test-service-2.e2e-tests-dns-h04r3.svc.cluster.local.

# on the master, direct to dnsmasq
$ dig @10.128.0.2 +tcp +search 89.65.30.172.in-addr.arpa. PTR
; <<>> DiG 9.9.4-RedHat-9.9.4-38.el7_3 <<>> @10.128.0.2 +tcp +search 89.65.30.172.in-addr.arpa. PTR
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 48864
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;89.65.30.172.in-addr.arpa.	IN	PTR

;; AUTHORITY SECTION:
30.172.in-addr.arpa.	30	IN	SOA	ns.global.gcedns-prod.internal. cloud-dns-hostmaster.google.com. 2015030600 7200 3600 24796800 5
```

This means we are not spec compliant with upstream.  It looks like we need to specify a ptr-address record in dnsmasq, but not sure (I would have expected it to pass that question along to the master, though).

sdodson:
SkyDNS returns SERVFAIL for non existant records and dnsmasq forwards the
requests on to the remaining servers so this should be safe.

Fixes #3017

### DIFF
--- a/roles/openshift_node_dnsmasq/files/networkmanager/99-origin-dns.sh
+++ b/roles/openshift_node_dnsmasq/files/networkmanager/99-origin-dns.sh
@@ -51,7 +51,7 @@ if [[ $2 =~ ^(up|dhcp4-change|dhcp6-change)$ ]]; then
 no-resolv
 domain-needed
 server=/cluster.local/172.30.0.1
-server=/30.172.in-addr.arpa/172.30.0.1
+server=/in-addr.arpa/172.30.0.1
 EOF
       # New config file, must restart
       NEEDS_RESTART=1
@@ -62,6 +62,7 @@ EOF
     for ns in ${IP4_NAMESERVERS}; do
       if [[ ! -z $ns ]]; then
         echo "server=${ns}"
+        echo "server=/in-addr.arpa/${ns}"
       fi
     done > $UPSTREAM_DNS_TMP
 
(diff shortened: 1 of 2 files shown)


sdodson:
aos-ci-test

sdodson:
Hmm, testing this a bit more dnsmasq seems to not be querying all the servers and taking whatever positive answer comes back. Need to test this more.

sdodson:
@smarterclayton here's the initial work, we'd want to update this to use the node's ip everywhere we're using the kube svc ip.

PR #3043, opened 2017-01-05
Add etcd scaleup playbook

jkhelil:
An etcd scaleup playbook is needed for these reasons:
- master scaleup playbook does not scaleup etcd
- if we broke a master vm and we loose it, the master recovery does not recover the etcd cluster size, each master rebuild results on decreasing etcd cluster size, this is particularly troubelsome for upgrades whenever someone wants to upgrades master nodes by destroying/recreating each one progressively

### DIFF
--- a/playbooks/common/openshift-cluster/evaluate_groups.yml
+++ b/playbooks/common/openshift-cluster/evaluate_groups.yml
@@ -5,10 +5,10 @@
   become: no
   gather_facts: no
   tasks:
-  - name: Evaluate groups - g_etcd_hosts required
+  - name: Evaluate groups - g_etcd_hosts or g_new_etcd_hosts required
     fail:
-      msg: This playbook requires g_etcd_hosts to be set
-    when: g_etcd_hosts is not defined
+      msg: This playbook requires g_etcd_hosts or g_new_etcd_hosts to be set
+    when: "{{ g_etcd_hosts is not defined and g_new_etcd_hosts is not defined}}"
 
   - name: Evaluate groups - g_master_hosts or g_new_master_hosts required
     fail:
@@ -65,6 +65,15 @@
       ansible_ssh_user: "{{ g_ssh_user | default(omit) }}"
       ansible_become: "{{ g_sudo | default(omit) }}"
     when: g_master_hosts|length > 0
+    changed_when: no
+
+  - name: Evaluate oo_new_etcd_to_config
+    add_host:
+      name: "{{ item }}"
+      groups: oo_new_etcd_to_config
+      ansible_ssh_user: "{{ g_ssh_user | default(omit) }}"
+      ansible_become: "{{ g_sudo | default(omit) }}"
+    with_items: "{{ g_new_etcd_hosts | default([]) }}"
     changed_when: no
 
   - name: Evaluate oo_masters_to_config
--- a/playbooks/common/openshift-etcd/scaleup.yml
+++ b/playbooks/common/openshift-etcd/scaleup.yml
@@ -1,0 +1,30 @@
+---
+- name: Configure etcd
+  hosts: oo_new_etcd_to_config
+  serial: 1
+  any_errors_fatal: true
+  vars:
+    etcd_ca_host: "{{ groups.oo_etcd_to_config.0 }}"
+  pre_tasks:
+  - name: Add new etcd members to cluster
+    command: >
+      /usr/bin/etcdctl  --cert-file {{ etcd_peer_cert_file }}
+                        --key-file {{ etcd_peer_key_file }}
+                        --ca-file {{ etcd_peer_ca_file }}
+                        -C {{ etcd_peer_url_scheme }}://{{ etcd_ca_host }}:{{ etcd_client_port }}
+                        member add {{ inventory_hostname }} {{ etcd_peer_url_scheme }}://{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}:{{ etcd_peer_port }}
+    delegate_to: "{{ etcd_ca_host }}"
+    register: etcd_add_check
+  roles:
+  - role: openshift_etcd
+    when: etcd_add_check.rc == 0
+    etcd_peers: "{{ groups.oo_etcd_to_config | union(groups.oo_new_etcd_to_config)| default([], true) }}"
+    etcd_ca_host: "{{ groups.oo_etcd_to_config.0 }}"
+    etcd_certificates_etcd_hosts: "{{ groups.oo_etcd_to_config | default([], true) }}"
+    etcd_initial_cluster_state: "existing"
+    initial_etcd_cluster: "{{ etcd_add_check.stdout_lines[3] | regex_replace('ETCD_INITIAL_CLUSTER=','') }}"
+    etcd_hostname: "{{ inventory_hostname }}"
+    etcd_ca_setup: False
+    r_etcd_common_etcd_runtime: "{{ openshift.common.etcd_runtime }}"
+  - role: nickhammond.logrotate
+    when: etcd_add_check.rc == 0
--- a/playbooks/aws/openshift-cluster/cluster_hosts.yml
+++ b/playbooks/aws/openshift-cluster/cluster_hosts.yml
@@ -4,6 +4,8 @@ g_all_hosts: "{{ groups['tag_clusterid_' ~ cluster_id] | default([])
 
 g_etcd_hosts: "{{ g_all_hosts | intersect(groups['tag_host-type_etcd'] | default([])) }}"
 
+g_new_etcd_hosts: "{{ g_all_hosts | intersect(groups['tag_host-type_new_etcd'] | default([])) }}"
+
 g_lb_hosts: "{{ g_all_hosts | intersect(groups['tag_host-type_lb'] | default([])) }}"
 
 g_nfs_hosts: "{{ g_all_hosts | intersect(groups['tag_host-type_nfs'] | default([])) }}"
--- a/playbooks/byo/openshift-cluster/cluster_hosts.yml
+++ b/playbooks/byo/openshift-cluster/cluster_hosts.yml
@@ -1,6 +1,8 @@
 ---
 g_etcd_hosts: "{{ groups.etcd | default([]) }}"
 
+g_new_etcd_hosts: "{{ groups.new_etcd | default([]) }}"
+
 g_lb_hosts: "{{ groups.lb | default([]) }}"
 
 g_master_hosts: "{{ groups.masters | default([]) }}"
@@ -18,6 +20,6 @@ g_glusterfs_hosts: "{{ groups.glusterfs | default([]) }}"
 g_glusterfs_registry_hosts: "{{ groups.glusterfs_registry | default(g_glusterfs_hosts) }}"
 
 g_all_hosts: "{{ g_master_hosts | union(g_node_hosts) | union(g_etcd_hosts)
-                 | union(g_lb_hosts) | union(g_nfs_hosts)
+                 | union(g_new_etcd_hosts) | union(g_lb_hosts) | union(g_nfs_hosts)
                  | union(g_new_node_hosts)| union(g_new_master_hosts)
                  | default([]) }}"
--- a/playbooks/byo/openshift-etcd/scaleup.yml
+++ b/playbooks/byo/openshift-etcd/scaleup.yml
@@ -1,0 +1,22 @@
+---
+- hosts: localhost
+  connection: local
+  become: no
+  gather_facts: no
+  tasks:
+  - include_vars: ../../byo/openshift-cluster/cluster_hosts.yml
+  - add_host:
+      name: "{{ item }}"
+      groups: l_oo_all_hosts
+    with_items: "{{ g_all_hosts }}"
+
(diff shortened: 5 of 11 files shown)


sdodson:
bot, test pull request

sdodson:
aos-ci-test

jkhelil:
This is not complete yet, I found some troubles with it, need to validate it before

jkhelil:
I updated the PR with playbooks working now

jkhelil:
@sdodson PTAL

jkhelil:
travis Erros fixed

jkhelil:
@ sdodson PTAL

sdodson:
Thanks, assigned someone to review who is more familiar with the tricky bits of cert work around etcd.

abutcher:
@jkhelil Started testing this afternoon and will be reviewing soon! Sorry for the delay on this one

jkhelil:
@abutcher 
Hi Andrew, Did you have time to take a lookm Do you have any feedback about this ?

abutcher:
@jkhelil I've tested the playbook starting with multiple etcd instances and it works great. I also think the changes here look good.

I didn't have success starting with a single etcd node and moving to multiple due to the way we configure advertised client urls. I haven't dug into this too deeply but it appears that with a single external etcd instance, the advertised client urls only contain `localhost` which causes the `member add` operation to fail. I expect there is some configuration we can change to make the advertised client urls contain the `protocol:ip:port` for a single node etcd cluster. Have you investigated this or would you expect starting with a single etcd instance to work?

```
[root@master1 ~]# /usr/bin/etcdctl --cert-file /etc/etcd/peer.crt --key-file /etc/etcd/peer.key --ca-file /etc/etcd/ca.crt -C https://master1.abutcher.com:2379 member add master2.abutcher.com https://192.168.122.53:2380
2017-03-08 16:48:09.107878 I | warning: ignoring ServerName for user-provided CA for backwards compatibility is deprecated
client: etcd cluster is unavailable or misconfigured; error #0: dial tcp [::1]:2379: getsockopt: connection refused
```

```
[root@master1 ~]# curl --cacert /etc/etcd/ca.crt --cert /etc/etcd/peer.crt --key /etc/etcd/peer.key -X GET https://192.168.122.186:2379/v2/members
{"members":[{"id":"16d01a6198b0837b","name":"default","peerURLs":["http://192.168.122.186:2380"],"clientURLs":["https://localhost:2379"]}]}
```

As opposed to:

```
[root@master1 ~]# curl --cacert /etc/etcd/ca.crt --cert /etc/etcd/peer.crt --key /etc/etcd/peer.key -X GET https://192.168.122.186:2379/v2/members
{"members":[{"id":"f2bf02afcb3a87c","name":"master1.abutcher.com","peerURLs":["https://192.168.122.186:2380"],"clientURLs":["https://192.168.122.186:2379"]},{"id":"81c1d1344f4a5c06","name":"master2.abutcher.com","peerURLs":["https://192.168.122.53:2380"],"clientURLs":["https://192.168.122.53:2379"]}]}
```

sdodson:
We probably need to do something to convert the single etcd instance from 'localhost' to hostnames and then re-initialize the cluster with that config before scaling up.

abutcher:
I was able to start with a single instance by making these changes to the etcd config in order to set a single node up for future clustering.

```diff
diff --git a/roles/etcd/templates/etcd.conf.j2 b/roles/etcd/templates/etcd.conf.j2
index 64c14a0..1095e76 100644
--- a/roles/etcd/templates/etcd.conf.j2
+++ b/roles/etcd/templates/etcd.conf.j2
@@ -8,12 +8,8 @@
 {% endfor -%}
 {% endmacro -%}
 
-{% if etcd_peers | default([]) | length > 1 %}
 ETCD_NAME={{ etcd_hostname }}
 ETCD_LISTEN_PEER_URLS={{ etcd_listen_peer_urls }}
-{% else %}
-ETCD_NAME=default
-{% endif %}
 ETCD_DATA_DIR={{ etcd_data_dir }}
 #ETCD_SNAPSHOT_COUNTER=10000
 ETCD_HEARTBEAT_INTERVAL=500
@@ -23,7 +19,6 @@ ETCD_LISTEN_CLIENT_URLS={{ etcd_listen_client_urls }}
 #ETCD_MAX_WALS=5
 #ETCD_CORS=
 
-{% if etcd_peers | default([]) | length > 1 %}
 #[cluster]
 ETCD_INITIAL_ADVERTISE_PEER_URLS={{ etcd_initial_advertise_peer_urls }}
 {% if initial_etcd_cluster is defined and initial_etcd_cluster %}
@@ -37,7 +32,6 @@ ETCD_INITIAL_CLUSTER_TOKEN={{ etcd_initial_cluster_token }}
 #ETCD_DISCOVERY_SRV=
 #ETCD_DISCOVERY_FALLBACK=proxy
 #ETCD_DISCOVERY_PROXY=
-{% endif %}
 ETCD_ADVERTISE_CLIENT_URLS={{ etcd_advertise_client_urls }}
 
 #[proxy]
```

jkhelil:
@abutcher @sdodson 
I tested these cases and it is working:
- Install single master with embeded etcd : ok
-> no ability to scale (I consider this as normal case, since we are not specifying we want a cluster)
- Install single master with option openshift_master_cluster_method=native, Ok (etcd has it own process)
  * Use etcd scaleup playbook : OK
-Install multiple masters with etcd in same hosts as masters: ok
  * Use etcd scaleup playbook : ok

I refactored a bit the thirdparty logic which seems to be still incomplete

playbooks/common/openshift-etcd/scaleup.yml
```diff
@@ -0,0 +1,31 @@
...
+  serial: 1
+  any_errors_fatal: true
+  vars:
+    etcd_ca_host: "{{ groups.oo_etcd_to_config.0 }}"
+    etcd_peer_url_scheme: https
+    etcd_peer_port: 2380
+  pre_tasks:
+  - name: Add new etcd members to cluster
+    command: >
+      /usr/bin/etcdctl  --cert-file /etc/etcd/peer.crt
```
ingvagabund: This should go inside some of the etcd roles. If the certs has different location than the default, this will fail.

abutcher: We could also add the `etcd_common` role here prior to tasks and rely on these vars https://github.com/openshift/openshift-ansible/blob/master/roles/etcd_common/defaults/main.yml

abutcher:
With the changes in the PR I'm able to go from one etcd instance to multiple using the scaleup playbook :+1: but not if the cluster was created prior to introducing these config changes. There may be a way to reconfig with these changes and then scaleup. I will test this today.

ingvagabund:
RFE for adding/removing etcd members here: https://github.com/openshift/openshift-ansible/issues/1772

playbooks/common/openshift-etcd/scaleup.yml
ingvagabund: Actually:
```yaml
TASK [Add new etcd members to cluster] *****************************************
fatal: [10.8.174.254 -> None]: FAILED! => {
    "changed": true, 
    "cmd": [
        "/usr/bin/etcdctl", 
        "--cert-file", 
        "/etc/etcd/peer.crt", 
        "--key-file", 
        "/etc/etcd/peer.key", 
        "--ca-file", 
        "/etc/etcd/ca.crt", 
        "-C", 
        "https://10.8.175.78:2379", 
        "member", 
        "add", 
        "10.8.174.254", 
        "https://172.16.186.31:2380"
    ], 
    "delta": "0:00:00.021797", 
    "end": "2017-05-04 06:56:48.534822", 
    "failed": true, 
    "rc": 1, 
    "start": "2017-05-04 06:56:48.513025", 
    "warnings": []
}

STDERR:

2017-05-04 06:56:48.524513 I | warning: ignoring ServerName for user-provided CA for backwards compatibility is deprecated
client: etcd cluster is unavailable or misconfigured; error #0: x509: certificate is valid for 172.16.186.182, not 10.8.175.78
```

as `/etc/etcd` directory does not exist yet. Since the ``openshift_etcd`` role that is run later creates the directory and its dependencies the certs.

playbooks/common/openshift-cluster/evaluate_groups.yml
```diff
@@ -67,6 +67,15 @@
     when: g_master_hosts|length > 0
     changed_when: no
 
+  - name: Evaluate oo_new_etcd_to_config
+    add_host:
```
ingvagabund: You could move this task into `playbooks/common/openshift-cluster/evalute_groups.yml` file which calls `add_host` for various groups.

playbooks/common/openshift-etcd/scaleup.yml
ingvagabund: I see there is ``delegate_to: "{{ etcd_ca_host }}"`` so it should run on the already configure etcd member host. Maybe my inventory is not correctly set.

ingvagabund: My bad, I had to change the `-C https://{{ etcd_ca_host }}:2379` to `-C https://{{ hostvars[etcd_ca_host]['ansible_default_ipv4']['address'] }}:2379`. Now it is running and the etcd reports:
```
etcdctl --cert-file /etc/etcd/peer.crt --key-file /etc/etcd/peer.key --ca-file /etc/etcd/ca.crt -C https://172.16.186.182:2379 cluster-health
2017-05-04 07:52:53.508934 I | warning: ignoring ServerName for user-provided CA for backwards compatibility is deprecated
2017-05-04 07:52:53.509571 I | warning: ignoring ServerName for user-provided CA for backwards compatibility is deprecated
member 6f73297e86157910 is unreachable: no available published client urls
member 8e9e05c52164694d is unhealthy: got unhealthy result from https://172.16.186.182:2379
cluster is unhealthy
```
as expected

ingvagabund:
use cases tested:
* 1-member rpm based cluster: it failed as expected, the etcd cluster was not able to elect a leader
* 3-member rpm based cluster: the cluster got scaled up and was heathy
* 1-member container based cluster: the same as the rpm based one, failed
* 3-member container based cluster: the new etcd_container service failed and the new member hangs out in [unstarted] state indefinitely.

ingvagabund:
1-node member related issue: https://github.com/coreos/etcd/issues/7820

playbooks/common/openshift-etcd/scaleup.yml
jkhelil: fixed

jkhelil:
@abutcher Hi Andrew, were you able to do this 
""With the changes in the PR I'm able to go from one etcd instance to multiple using the scaleup playbook 👍 but not if the cluster was created prior to introducing these config changes. There may be a way to reconfig with these changes and then scaleup. I will test this today.""

abutcher:
> @abutcher Hi Andrew, were you able to do this
""With the changes in the PR I'm able to go from one etcd instance to multiple using the scaleup playbook :+1: but not if the cluster was created prior to introducing these config changes. There may be a way to reconfig with these changes and then scaleup. I will test this today.""

Yes, tested this but not successfully.

jkhelil:
I am working on it, I keep you informed

jkhelil:
I manage to have it working with a cluster created with old conf, and I was able to scaleup etcd on new node using rpm based config, not dockerized one.
but I am encoutring this error 
TASK [etcd_common : Fail if invalid r_etcd_common_action provided] *************
task path: /home/ansible/openshift-ansible/roles/etcd_common/tasks/main.yml:2
fatal: [ose3-int-a-node1.node.new-gen-1a-eu-central-1.acs]: FAILED! => {
    "failed": true, 
    "msg": "The conditional check 'o' failed. The error was: error while evaluating conditional (o): 'o' is undefined\n\nThe error appears to have been in '/home/ansible/openshift-ansible/roles/etcd/tasks/main.yml': line 123, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- include_role:\n  ^ here\n"
}


which seems not related to etcd scaleup, it is reported here 
https://github.com/openshift/openshift-ansible/issues/4121

jkhelil:
@abutcher can you give me explicit detail about what is not working exactly, I was able to add an etcd node using thes playbooks on a stack created with 3.2 playbooks. except the error i mentioned, which seems to be a bug on last openshift ansible

abutcher:
@jkhelil The only case that isn't working for me is to create a cluster with a single instance in the [etcd] group and then attempt to scale it up with this branch. However, if the cluster was initially created with this branch then scaling up from a single instance works fine.

I also encountered some issues with overridden hostnames which I was able to work around with this commit https://github.com/abutcher/openshift-ansible/commit/54d3b459aa36276961dfea7af8295335b9bf70df.

sdodson:
aos-ci-test

sdodson:
[merge]

abutcher:
bot, retest this please

abutcher:
aos-ci-test

ganhuang:
@sdodson Do you think we would need a card to track the feature? I'm afraid it would be missed by QE if it's intended to be in OCP, as well as the document.

sdodson:
@ganhuang there's one already, moved it to complete https://trello.com/c/EESwIsuW/171-5-support-for-scaling-up-etcd

PR #3044, opened 2017-01-05
Add new option 'docker_selinux_enabled'

mtnbikenc:
Allows controlling 'selinux-enabled' docker options
Can be set to true or false

https://bugzilla.redhat.com/show_bug.cgi?id=1392742

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -96,7 +96,7 @@
     dest: /etc/sysconfig/docker
     regexp: '^OPTIONS=.*$'
     line: "OPTIONS='\
-      {% if ansible_selinux and ansible_selinux.status == '''enabled''' %} --selinux-enabled{% endif %}\
+      {% if ansible_selinux.status | default(None) == '''enabled''' and docker_selinux_enabled | default(true) %} --selinux-enabled {% endif %}\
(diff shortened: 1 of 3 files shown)


mtnbikenc:
aos-ci-test

mtnbikenc:
@detiber PTAL

roles/docker/tasks/main.yml
```diff
@@ -102,7 +102,7 @@
     dest: /etc/sysconfig/docker
     regexp: '^OPTIONS=.*$'
     line: "OPTIONS='\
-      {% if ansible_selinux and ansible_selinux.status == '''enabled''' %} --selinux-enabled{% endif %}\
+      {% if docker_selinux_enabled is defined %} --selinux-enabled={{ docker_selinux_enabled }} {% elif ansible_selinux and ansible_selinux.status == '''enabled''' %} --selinux-enabled{% endif %}\
```
detiber: @mtnbikenc hah, it looks like I previously started a review, but never submitted :)

mtnbikenc:
aos-ci-test

mtnbikenc:
@detiber Updated the PR based on your suggestion.  I've tested this with both no option defined as well as setting `openshift_docker_selinux_enabled=false`.

PR #3047, opened 2017-01-05
Change wording in the quick installer callback plugin

tbielawa:
* Old wording only considered skipped plays (reducing the play count)
* New wording accounts for dynamically included plays (increasing the
  play count)

[[quick install]the number of completed plays is bigger than the expected number](https://bugzilla.redhat.com/show_bug.cgi?id=1388739)


tbielawa:
aos-ci-test

PR #3049, opened 2017-01-05
logging-deployer fixes from origin-aggregated-logging/#317

sdodson:
https://github.com/openshift/origin-aggregated-logging/pull/317


sdodson:
@ewolinetz pulled from your PR, mind checking the v1.3 content isn't missing some other fixes? It's missing one of the roles from v1.4

ewolinetz:
@sdodson that is correct, that new role was created as part of 1.4 since we updated the `openshift-elasticsearch-plugin` to read cluster role bindings to determine if a user was a cluster-admin|cluster-reader for that version.

PR #3057, opened 2017-01-09
Serial master upgrade

dgoodwin:
This was the start of my 3.5 upgrade work which will need to be set aside. For ops needs the goal was to get more of master upgrade done serially instead of a bunch of modifications in parallel, then serially restarting services/servers later. 

To accomplish it looked best and preferrable to move the validation of the restart policy to pre-upgrade, and update the standalone restart playbooks accordingly.


ashcrow:
aos-ci-test

sdodson:
This sure sounds like an ansible regression in 2.2.1.0.

```
TASK [openshift_master : set_fact] *********************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: RepresenterError: cannot represent an object: deny_all
fatal: [10.8.171.155]: FAILED! => {"failed": true, "msg": "Unexpected failure during module execution.", "stdout": ""}
```

sdodson:
downgraded to 2.2.0.0 and re-running

sdodson:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -27,16 +27,8 @@
...
-  include: ./etcd/main.yml
-
-- name: Upgrade master packages
-  hosts: oo_masters_to_config
-  roles:
-  - openshift_facts
-  tasks:
-  - include: rpm_upgrade.yml component=master
-    when: not openshift.common.is_containerized | bool
+  #- name: Upgrade and backup etcd
```
ashcrow: Is including ``etcd/main.yml`` for backing up no longer needed?

dgoodwin: Very bad mistaken commit. :) My bad, I will yank it.

ashcrow:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
sdodson: In the future we should consider splitting etcd upgrade out of control plane, it's really more like "Database". But fine to keep it for now until we find that it creates problems.

mtnbikenc:
aos-ci-test

ashcrow:
The changes make sense to me.

ingvagabund:
The patch looks good (given my incomplete knowledge of the code). How does the change impact performance? Though this impacts only the control plane, the complexity of the upgrade shifts from constant to linear (given a number of master nodes). Given 20 nodes (why make 20 master nodes anyway?) this could take some time. What is the number of masters in common/expected HA cluster deployment?

playbooks/common/openshift-master/restart_services.yml
```diff
@@ -8,19 +8,19 @@
   service:
     name: "{{ openshift.common.service_type }}-master-api"
     state: restarted
-  when: openshift_master_ha | bool and openshift.master.cluster_method != 'pacemaker'
```
dgoodwin: Yeah pacemaker is no longer supported.

playbooks/byo/openshift-cluster/upgrades/v3_4/upgrade.yml
```diff
@@ -66,6 +66,10 @@
   tags:
   - pre_upgrade
 
+- include: ../../../../common/openshift-master/validate_restart.yml
```
dgoodwin: It's some logic that was pre-existing but was deep into the upgrade, it made sense on two fronts to push this further up into the upgrade before we make any real changes to validate the state of things, rather than letting it fail midway. 

The most valuable thing it does is make sure you're not trying to restart entire systems, while running ansible on one of them. (which would kill the whole upgrade)

dgoodwin:
Performance wise it will be a little slower as the master package updates go from parallel to serial, but indeed as far as I know virtually nobody ever goes beyond 3 masters and would have very little reason to do so, so I think this is probably a worthy tradeoff. The package updates indeed could trigger a lot of things happening so it's much safer to have it happen per master when that system is ideally taken out of the load balancer.

playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
```diff
@@ -66,6 +66,10 @@
   tags:
   - pre_upgrade
 
+- include: ../../../../common/openshift-master/validate_restart.yml
```
dgoodwin: Moved both of these to the prior commit.

ashcrow:
Odd that CI hasn't reported status ...

sdodson:
I believe that's because commits have been pushed since the last run started. I think github invalidates tests as soon as new commits are pushed.

sdodson:
aos-ci-test

ashcrow:
aos-ci-test

PR #3059, opened 2017-01-09
Use registry.access.redhat.com/rhel7/etcd instead of etcd3

sdodson:


andrewklau:
@sdodson I'm just curious, does this mean 1.4 won't be using etcd3 yet?

sdodson:
It will use etcd v2 API but we're trying to get everyone to the latest
packages. The upgrade playbook will upgrade to the latest version of etcd
image which will soon be 3.0.15. The etcd3 image will never be updated
again. I was trying to avoid implementing [short lived] conditional logic
but that meant I had to go with a version that won't be available until
next week.

On Jan 13, 2017 8:43 PM, "Andrew Lau" <notifications@github.com> wrote:

@sdodson <https://github.com/sdodson> I'm just curious, does this mean 1.4
won't be using etcd3 yet?

—
<https://github.com/openshift/openshift-ansible/pull/3059#issuecomment-272591564>,
or mute the thread

andrewklau:
Makes sense, thanks!

PR #2240, opened 2016-07-29
Speed Improvements

abutcher:

PR #3060, opened 2017-01-09
Return role dependencies to roles for openshift_{hosted,master,node}

abutcher:
Undoes #2240 in which role dependencies were moved to playbooks to improve run speed w/ ansible 2.1.


abutcher:
aos-ci-test

ashcrow:
I did some smoke testing on this as well. Worked fine :+1:.

detiber:
@abutcher what is the performance diff with this vs the current approach?

abutcher:
@detiber about the same

```
Dependencies in playbook: 11:06.71
Dependencies in role meta: 10:54.38
```

PR #3064, opened 2017-01-10
use etcdctl from the container when containerized=True

giuseppe:
found this out while investigating an upgrade on Atomic Host, where etcd 2 is installed on the host.

If etcd runs from a container, then use etcdctl from there.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
+++ b/playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
@@ -3,6 +3,7 @@
   vars:
     embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
     timestamp: "{{ lookup('pipe', 'date +%Y%m%d%H%M%S') }}"
+    etcdctl_command: "{{ 'etcdctl' if not openshift.common.is_containerized or embedded_etcd else 'docker exec etcd_container etcdctl' }}"
   roles:
   - openshift_facts
   tasks:
@@ -47,7 +48,7 @@
 
   - name: Generate etcd backup
     command: >
-      etcdctl backup --data-dir={{ openshift.etcd.etcd_data_dir }}
+      {{ etcdctl_command }} backup --data-dir={{ openshift.etcd.etcd_data_dir }}
       --backup-dir={{ openshift.common.data_dir }}/etcd-backup-{{ backup_tag | default('') }}{{ timestamp }}
 
   - set_fact:


playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
```diff
@@ -49,6 +49,13 @@
     command: >
       etcdctl backup --data-dir={{ openshift.etcd.etcd_data_dir }}
       --backup-dir={{ openshift.common.data_dir }}/etcd-backup-{{ backup_tag | default('') }}{{ timestamp }}
+    when: not openshift.common.is_containerized | bool
+
+  - name: Generate etcd backup
+    command: >
+      docker exec etcd_container etcdctl backup --data-dir={{ openshift.etcd.etcd_data_dir }}
+      --backup-dir={{ openshift.common.data_dir }}/etcd-backup-{{ backup_tag | default('') }}{{ timestamp }}
+    when: openshift.common.is_containerized | bool
```
detiber: Instead of two different tasks, it might make sense to set the etcdctl command using a variable:

```
vars:
  ...
  etcdctl_command: "{{ 'etcdctl' if not openshift.common.is_containerized else 'docker exec etcd_container etcdctl' }}"
tasks:
  ...
  - name Generate etcd backup
    command: >
       {{ etcdctl_command }} backup --data-dir={{ openshift.etcd.etcd_data_dir }}
        --backup-dir={{ openshift.common.data_dir }}/etcd-backup-{{ backup_tag | default('') }}{{ timestamp }}

**Comment by sdodson** [comment]
This is a longstanding TODO item, only gap I see right now is we need to either complete the migration from all scenarios where 'embedded' etcd is in use. Perhaps the right way to do that is to pull the latest etcd container image and use that to perform the backup?

File: playbooks/common/openshift-cluster/upgrades/etcd/backup.yml [hunk 2]
```diff
@@ -3,6 +3,7 @@
   vars:
     embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
     timestamp: "{{ lookup('pipe', 'date +%Y%m%d%H%M%S') }}"
+    etcdctl_command: "{{ 'etcdctl' if not openshift.common.is_containerized else 'docker exec etcd_container etcdctl' }}"
```
sdodson: We can accommodate embedded environments here too.
`etcdctl_command: "{{ 'etcdctl' if not openshift.common.is_containerized or etcd_embedded else 'docker exec etcd_container etcdctl' }}"`

giuseppe: should it be `embedded_etcd`?

detiber: I'm not sure that would work for embedded_etcd use for containerized installs, seems like a particular edge case that might cause problems.

sdodson: Yeah, it's still problematic to use etcdctl from the host on embedded containerized installs because the versioning can be different but I think this does the best thing for all the "supported" scenarios and is no worse than status quo for embedded containerized etcd instances.

giuseppe:
how to get the checks done?

sdodson:
bot, add author to whitelist

sdodson:
bot, test pull request

giuseppe:
I was added to the openshift organization

giuseppe:
aos-ci-test

giuseppe:
I've no write access to the repository.  Could someone merge it?

sdodson:
Done, I'll try to figure out why you couldn't merge that on your own and correct the permissions.

PR #3065, opened 2017-01-10
Workaround for dnf+docker version race condition

ashcrow:
In some versions of ``dnf`` the rpm database can require maintenance after
usage. Unfortunately, this trips up ``get_version_output`` in the
openshift_facts library as the maintenance may interrupt the ``STDOUT``
stream. This makes the resulting data returned from ``get_version_output``
malformed and unparsable. This change adds an ``rpm --rebuilddb`` command
before requesting docker version output when the ansible package manager
is ``dnf``. This causes the ``STDOUT`` maintenance messages to no longer
interrupt the expected yaml stream as well as releases locks in the rpm
database.

For example of people reporting this issue see See https://bugzilla.redhat.com/show_bug.cgi?id=918184#c18

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1256,6 +1256,13 @@ def is_service_running(service):
     return service_running
 
 
+def rpm_rebuilddb():
+    """
+    Runs rpm --rebuilddb to ensure the db is in good shape.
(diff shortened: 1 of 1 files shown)


ashcrow:
cc @abutcher

abutcher:
aos-ci-test

PR #3075, opened 2017-01-11
[WIP] Correct return code compairison for yamllint

mtnbikenc:
This PR corrects a problem where yamllint check failures were not being reported as failed by tox.

The problem.level is text based and was being compared against integer based values, therefore never raising an exception on issues.

Additionally, the use of `assert` on the results felt wrong with the resulting traceback.  The code is not failing, we are just wanting to report that something bad happened.  However, there are probably better ways to do this than what I have here.


mtnbikenc:
@detiber @tbielawa PTAL

PR #3077, opened 2017-01-11
Correct tox to run on Travis

mtnbikenc:
tox-travis needs a python env descriptor in envlist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,8 +1,7 @@
 [tox]
 minversion=2.3.1
 envlist =
-    py{27,35}-ansible22-{pylint,unit,flake8}
-    yamllint
+    py{27,35}-ansible22-{pylint,unit,flake8,yamllint}
 skipsdist=True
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -2,7 +2,7 @@
 minversion=2.3.1
 envlist =
     py{27,35}-ansible22-{pylint,unit,flake8}
-    yamllint
+    py27-yamllint
```
detiber: Would it be worth it to add an env for py35 as well?

mtnbikenc: We could, but didn't know how much difference there would be between the two for yaml parsing.  I was unsure as to why you didn't just have it like this:

`py{27,35}-ansible22-{pylint,unit,flake8,yamllint}`

mtnbikenc:
Looks like we are now getting yamllint in both 2.7 and 3.5 envs.  :+1:

PR #3085, opened 2017-01-12
Do not default registry storage kind to 'nfs' when 'nfs' host group exists.

### LINKED ISSUE #3063: when NFS group exists, registry storage configuration becomes complicated
In the case of wanting to use something like GCS for the registry, where NFS is being used for other things, there's no easy way to get a registry up and going.

* If you define the bare minimum information you get a default configuration of NFS for the registry
* If you define GCS right now it is unsupported so you get an error/failure
* If you try to use "emptydir" and clean up later this is considered an invalid option, even though it is valid.

I would think that the installer should not wait until it's 3/4 done to bomb on an unsupported registry storage type. It would be nice if the "default" was actually emptydir unless something else explicit was suggested, or at least make emptydir an option.

abutcher:
This makes it possible to use openshift-ansible configured nfs for
hosted components other than the registry while also configuring
empty-dir registry storage (no `openshift_hosted_registry_storage_kind`
set).

Fixes #3063

@thoraxe

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -74,11 +74,6 @@
         public_console_url: "{{ openshift_master_public_console_url | default(None) }}"
         ha: "{{ openshift_master_ha | default(groups.oo_masters | length > 1) }}"
         master_count: "{{ openshift_master_count | default(groups.oo_masters | length) }}"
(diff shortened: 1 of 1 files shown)


abutcher:
aos-ci-test

Issue #2984, opened 2016-12-16
dedicated build nodes feature should be surfaced in the installer


jeremyeder:
I heard that 3.4 has a dedicated build node capability, but that it is a manual procedure.  Please surface this capability in the installer.

sdodson:
Link to docs describing that functionality?

jeremyeder:
I heard this from Steve Speicher.  Will ping him.

jeremyeder:
Got an out of office reply from Steve.

jeremyeder:
@sdodson @detiber docs are here: https://docs.openshift.org/latest/install_config/build_defaults_overrides.html

/cc @bparees 

I believe at minimum, the installer needs to be taught to set a taint on nodes somehow.

bparees:
for configuring the build defaulter/override behavior via ansible, we're going to do this:
https://trello.com/c/LsaA9CN2

but setting up appropriate taints on the nodes falls outside the scope of that card (hence @jeremyeder's ask)

bparees:
@jeremyeder it looks like you should already be able to do this by setting the openshift_builddefaults_json:

https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.origin.example#L619

PR #3086, opened 2017-01-12
add configuration for build default+overrides settings

bparees:
fixes https://github.com/openshift/openshift-ansible/issues/2984

### DIFF
--- a/roles/openshift_builddefaults/tasks/main.yml
+++ b/roles/openshift_builddefaults/tasks/main.yml
@@ -15,6 +15,7 @@
       no_proxy: "{{ openshift_builddefaults_no_proxy | default(None) }}"
       git_http_proxy: "{{ openshift_builddefaults_git_http_proxy | default(None) }}"
       git_https_proxy: "{{ openshift_builddefaults_git_https_proxy | default(None) }}"
+      git_no_proxy: "{{ openshift_builddefaults_git_no_proxy | default(None) }}"
 
 - name: Set builddefaults config structure
   openshift_facts:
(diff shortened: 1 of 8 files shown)


playbooks/common/openshift-master/config.yml
```diff
@@ -154,6 +154,7 @@
   - role: openshift_clock
   - role: openshift_cloud_provider
   - role: openshift_builddefaults
+  - role: openshift_buildoverrides
```
detiber: should this role actually be a dependency of the openshift_master role (or openshift_master_facts) rather than a separate role? It seems like it is a dependency, but is only being enforced by the playbook and not the role metadata.

bparees: probably?  I just copied the openshift_builddefaults role.  Should that also be changed?

bparees:
@sdodson @detiber ptal.

playbooks/common/openshift-master/config.yml
bparees: @detiber bump.

sdodson: This was a very recent change, rebase and move this to roles/openshift_master/meta/main.yml after openshift_builddefaults. Not sure why github isn't flaging this for needing a rebase.

sdodson: d113f03968476c4bf899e20317bfa644ee539560 moved it

bparees:
@sdodson @detiber ok i've moved the role to be a dependency of the openshift-master role.  ptal.

sdodson:
aos-ci-test

bparees:
@sdodson ok travis is happy, looks like a bunch of other things are kinda hung, should i be expecting them to complete at some point?

sdodson:
@bparees they only fire on command
aos-ci-test

sdodson:
aos-ci-test

PR #3087, opened 2017-01-12
Added setup.py to flake8 tests

ashcrow:
Added setup.py to ``flake8`` tests since it is code. Also minor updates to
slightly modernize.

Note: In the future ``{#}`` should probably be ported to ``{}`` as numbering is
no longer needed (and disliked by many).


detiber:
@ashcrow :+1: to removing numbering in format, but it'll probably require an update to the flake8 and/or pylint excludes to avoid throwing an error/warning.

PR #3088, opened 2017-01-12
Fix role openshift_storage_nfs_lvm

jfchevrette:
This role would always fail at the first task when testing whether we're running atomic even when applied to a non-atomic host.

- Fixed the 'when' statement that was wrongly indented at the task options level
- Added openshift_facts as a dependency in order to ensure the openshift.common.is_atomic is set prior to running this role

### DIFF
--- a/roles/openshift_storage_nfs_lvm/meta/main.yml
+++ b/roles/openshift_storage_nfs_lvm/meta/main.yml
@@ -14,4 +14,5 @@ galaxy_info:
     - all
   categories:
   - openshift
-dependencies: []
+dependencies:
+- role: openshift_facts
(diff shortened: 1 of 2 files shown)


jfchevrette:
Not sure why the tests are failing. The openshift_storage_nfs_lvm role is not included in any of the cluster playbooks and only included in the setupnfs.yml playbook. No traces of the openshift_storage_nfs_lvm role in the test log.

I'm fairly new to this, but it doesn't look like the tests are failing because of my changes.

sdodson:
bot, test pull request

sdodson:
aos-ci-test

abutcher:
ok to test

PR #3090, opened 2017-01-12
Fix role openshift_storage_nfs_lvm keys should be lowercase in the pv spec template

jfchevrette:
According to the spec, the server and path keys should be lower case.

```
The PersistentVolume "pv-192-168-173-18-os-pv1g0001" is invalid.

* spec.nfs.server: Required value
* spec.nfs.path: Required value
* spec.nfs.path: Invalid value: "": must be an absolute path
```


abutcher:
ok to test

PR #3092, opened 2017-01-13
lib_openshift.  This is the initial commit that has oc_route.

kwoodson:


kwoodson:
@tbielawa, @abutcher, @twiest, @sdodson 

Here is the first couple of modules from the lib_openshift collection.  Please review.

kwoodson:
aos-ci-test

kwoodson:
aos-ci-test

kwoodson:
aos-ci-test

roles/lib_openshift/src/generate.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+'''
+  Generate the openshift-ansible/roles/lib_openshift_cli/library/ modules.
+'''
+
+import argparse
+import os
+import StringIO
```
detiber: This is causing a python3-related test failure in travis. Instead of this, you can use:
```
from six import StringIO
```
and change later `StringIO.StringIO` references to just `StringIO`.

detiber:
@kwoodson Since these modules aren't currently in use, you can probably skip the `aos-ci-tests` for now.

roles/lib_openshift/src/generate.py
kwoodson: Fixed and pushed.

roles/lib_utils/library/yedit.py
```diff
@@ -163,13 +162,15 @@
 #   b:
```
tbielawa: The `separator` option is missing from the `DOCUMENTATION` string.

tbielawa:
aos-ci-test

sdodson:
When reviewing we need to make sure to collect any python dependencies and make sure they're in Fedora and make sure that their either in RHEL/Centos or we get them ready to ship in the OCP/Origin repos.

roles/lib_openshift/src/lib/base.py
```diff
@@ -0,0 +1,456 @@
...
+                "results": results,
+                "cmd": ' '.join(cmds)}
+
+        if proc.returncode == 0:
+            if output:
+                if output_type == 'json':
+                    try:
+                        rval['results'] = json.loads(stdout)
+                    except ValueError as err:
+                        if "No JSON object could be decoded" in str(err):
```
ashcrow: nit: replace ``str(err)`` with ``err.message``. Casting the ``ValueError``. The ``message`` attribute will always be the a ``basestring`` message. While true that casting the ``Exception`` to a ``str`` should always return just the ``message`` attribute it's a bit harder to read and could change between versions.

## Example:
```python
    # ...
    except ValueError as err:
        if err.message == "No JSON object could be decoded":
            # ...
```

roles/lib_openshift/src/lib/base.py
```diff
@@ -0,0 +1,456 @@
...
+                            err = str(err)
```
ashcrow: nit: same as above.

roles/lib_openshift/src/lib/base.py
```diff
@@ -0,0 +1,456 @@
...
+                elif output_type == 'raw':
+                    rval['results'] = stdout
+
+            if self.verbose:
+                print(stdout)
+                print(stderr)
```
ashcrow: nit: Unless ``stderr`` is sacred in this script it would be a good idea to print out to ``stderr`` or note that this is ``stderr`` output.

## Example: Print to stderr
```python
from __future__ import print_function
# ...
import sys
#...
            print(stderr, file=sys.stderr)
```

## Example: Specify Stream
```python
        print("STDOUT: {0}".format(stdout))
        print("STDERR: {0}".format(stderr))
```

kwoodson:
@sdodson 

The current modules needed for lib_utils and lib_openshift that are non-standard:
- ruamel.yaml

This is because lib_openshift uses yedit under the covers.

ashcrow:
Note: None of my nits should be considered blocking.

PR #3093, opened 2017-01-13
Correct consistency between upgrade playbooks

mtnbikenc:
The control_plane only playbook should be consistent with the 'all' playbook as referenced below.

https://github.com/openshift/openshift-ansible/blob/78b948edb0c30e3ec876916a8bbe08db5f055ea7/playbooks/byo/openshift-cluster/upgrades/v3_4/upgrade.yml#L92-L93


mtnbikenc:
cc: @dgoodwin

mtnbikenc:
aos-ci-test

PR #3096, opened 2017-01-16
Support openshift_node_port_range for configuring service NodePorts

abutcher:

Supersedes #3018


abutcher:
aos-ci-test

smarterclayton:
LGTM (to the limits of my knowledge)

abutcher:
Release 1.4 backport PR is on the way.

PR #3097, opened 2017-01-16
Backport 1.4: Support openshift_node_port_range for configuring service NodePorts

abutcher:

Backports #3096, #3060


abutcher:
aos-ci-test

abutcher:
aos-ci-test

sdodson:
aos-ci-test

PR #3101, opened 2017-01-17
Implement simple hooks pre/post master upgrade.

dgoodwin:
Implement simple hook variables for pre-post master upgrade, if defined we'll include the referenced file. Absolute paths seems best, otherwise it needs to be relative to where the include is actually used. 

This will allow users/ops to do things like pull masters out of load balancers and put them back in, perform system updates, etc.

### DIFF
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -89,6 +89,21 @@ openshift_release=v1.4
 # Skip upgrading Docker during an OpenShift upgrade, leaves the current Docker version alone.
 # docker_upgrade=False
 
+
+# Upgrade Hooks
+#
+# Hooks are available to run custom tasks at various points during a cluster
+# upgrade. Each hook should point to a file with Ansible tasks defined. Suggest using
+# absolute paths, if not the path will be treated as relative to the file where the
+# hook is actually used.
+#
+# Tasks to run before each master is upgraded:
+# openshift_upgrade_pre_master_hook=/usr/share/custom/pre_master.yml
+#
+# Tasks to run after each master is upgraded:
+# openshift_upgrade_post_master_hook=/usr/share/custom/post_master.yml
+
+
 # Alternate image format string, useful if you've got your own registry mirror
 #oreg_url=example.com/openshift3/ose-${component}:${version}
 # If oreg_url points to a registry other than registry.access.redhat.com we can
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -97,11 +97,15 @@ openshift_release=v1.4
 # absolute paths, if not the path will be treated as relative to the file where the
 # hook is actually used.
 #
-# Tasks to run before each master is upgraded:
-# openshift_upgrade_pre_master_hook=/usr/share/custom/pre_master.yml
+# Tasks to run before each master is upgraded.
+# openshift_master_upgrade_pre_hook=/usr/share/custom/pre_master.yml
 #
-# Tasks to run after each master is upgraded:
-# openshift_upgrade_post_master_hook=/usr/share/custom/post_master.yml
+# Tasks to run to upgrade the master. These tasks run after the main openshift-ansible
+# upgrade steps, but before we restart system/services.
+# openshift_master_upgrade_hook=/usr/share/custom/master.yml
+#
+# Tasks to run after each master is upgraded and system/services have been restarted.
+# openshift_master_upgrade_post_hook=/usr/share/custom/post_master.yml
 
 
 # Alternate image format string, useful if you've got your own registry mirror
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -89,6 +89,21 @@ openshift_release=v3.4
 # Skip upgrading Docker during an OpenShift upgrade, leaves the current Docker version alone.
 # docker_upgrade=False
 
+
+# Upgrade Hooks
+#
+# Hooks are available to run custom tasks at various points during a cluster
+# upgrade. Each hook should point to a file with Ansible tasks defined. Suggest using
+# absolute paths, if not the path will be treated as relative to the file where the
+# hook is actually used.
+#
+# Tasks to run before each master is upgraded:
+# openshift_upgrade_pre_master_hook=/usr/share/custom/pre_master.yml
+#
+# Tasks to run after each master is upgraded:
+# openshift_upgrade_post_master_hook=/usr/share/custom/post_master.yml
+
+
 # Alternate image format string, useful if you've got your own registry mirror
 #oreg_url=example.com/openshift3/ose-${component}:${version}
 # If oreg_url points to a registry other than registry.access.redhat.com we can
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -97,11 +97,15 @@ openshift_release=v3.4
 # absolute paths, if not the path will be treated as relative to the file where the
 # hook is actually used.
 #
-# Tasks to run before each master is upgraded:
-# openshift_upgrade_pre_master_hook=/usr/share/custom/pre_master.yml
+# Tasks to run before each master is upgraded.
+# openshift_master_upgrade_pre_hook=/usr/share/custom/pre_master.yml
 #
-# Tasks to run after each master is upgraded:
-# openshift_upgrade_post_master_hook=/usr/share/custom/post_master.yml
+# Tasks to run to upgrade the master. These tasks run after the main openshift-ansible
+# upgrade steps, but before we restart system/services.
+# openshift_master_upgrade_hook=/usr/share/custom/master.yml
+#
+# Tasks to run after each master is upgraded and system/services have been restarted.
+# openshift_master_upgrade_post_hook=/usr/share/custom/post_master.yml
 
 
 # Alternate image format string, useful if you've got your own registry mirror
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -108,14 +108,15 @@
       state: link
     when: ca_crt_stat.stat.isreg and not ca_bundle_stat.stat.exists
 
+  # NOTE: We run user provided hooks prior to restarting the system or services.
+  - include: "{{ openshift_upgrade_post_master_hook }}"
+    when: openshift_upgrade_post_master_hook is defined
+
   - include: ../../openshift-master/restart_hosts.yml
     when: openshift.common.rolling_restart_mode == 'system'
 
   - include: ../../openshift-master/restart_services.yml
     when: openshift.common.rolling_restart_mode == 'services'
-
-  - include: "{{ openshift_upgrade_post_master_hook }}"
-    when: openshift_upgrade_post_master_hook is defined
 
   - set_fact:
       master_update_complete: True
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -51,6 +51,8 @@
   roles:
   - openshift_master_facts
 
+# The main master upgrade play. Should handle all changes to the system in one pass, with
+# support for optional hooks to be defined.
 - name: Upgrade master
   hosts: oo_masters_to_config
   vars:
@@ -62,6 +64,10 @@
   roles:
   - openshift_facts
   post_tasks:
+
+  - include: "{{ openshift_upgrade_pre_master_hook }}"
+    when: openshift_upgrade_pre_master_hook is defined
+
   - include: rpm_upgrade.yml component=master
     when: not openshift.common.is_containerized | bool
 
@@ -107,6 +113,9 @@
 
   - include: ../../openshift-master/restart_services.yml
     when: openshift.common.rolling_restart_mode == 'services'
+
+  - include: "{{ openshift_upgrade_post_master_hook }}"
+    when: openshift_upgrade_post_master_hook is defined
 
   - set_fact:
       master_update_complete: True
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -65,6 +65,9 @@
   - openshift_facts
   post_tasks:
 
+  - debug: msg="Running pre-master upgrade hook {{ openshift_upgrade_pre_master_hook }}"
+    when: openshift_upgrade_pre_master_hook is defined
+
   - include: "{{ openshift_upgrade_pre_master_hook }}"
     when: openshift_upgrade_pre_master_hook is defined
 
@@ -109,6 +112,9 @@
     when: ca_crt_stat.stat.isreg and not ca_bundle_stat.stat.exists
 
   # NOTE: We run user provided hooks prior to restarting the system or services.
+  - debug: msg="Running post-master upgrade hook {{ openshift_upgrade_post_master_hook }}"
+    when: openshift_upgrade_post_master_hook is defined
+
   - include: "{{ openshift_upgrade_post_master_hook }}"
     when: openshift_upgrade_post_master_hook is defined
 
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -65,11 +65,12 @@
   - openshift_facts
   post_tasks:
 
-  - debug: msg="Running pre-master upgrade hook {{ openshift_upgrade_pre_master_hook }}"
-    when: openshift_upgrade_pre_master_hook is defined
+  # Run the pre-upgrade hook if defined:
+  - debug: msg="Running master pre-upgrade hook {{ openshift_master_upgrade_pre_hook }}"
+    when: openshift_master_upgrade_pre_hook is defined
 
-  - include: "{{ openshift_upgrade_pre_master_hook }}"
-    when: openshift_upgrade_pre_master_hook is defined
+  - include: "{{ openshift_master_upgrade_pre_hook }}"
+    when: openshift_master_upgrade_pre_hook is defined
 
   - include: rpm_upgrade.yml component=master
     when: not openshift.common.is_containerized | bool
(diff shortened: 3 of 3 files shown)


inventory/byo/hosts.origin.example
```diff
@@ -89,6 +89,21 @@ openshift_release=v1.4
...
+
+# Upgrade Hooks
+#
+# Hooks are available to run custom tasks at various points during a cluster
+# upgrade. Each hook should point to a file with Ansible tasks defined. Suggest using
+# absolute paths, if not the path will be treated as relative to the file where the
+# hook is actually used.
+#
+# Tasks to run before each master is upgraded:
+# openshift_upgrade_pre_master_hook=/usr/share/custom/pre_master.yml
```
detiber: Probably need to start defining an interface, so that we can set expectations on what variables/state are available in the hooks.

First questions I have:
- What host(s) are the hook run against
- Is the playbook run as a full playbook include, or a task include?
- What variables are set/supported?
- If there is a hook failure in the middle of a run, how does a user recover?
- Do we expect there to be data returned from the hooks (and possibly passed on to later hooks)?

dgoodwin: The hook runs against each master in serial. It can contain only tasks as it needs to plug into the middle of our serial upgrade.

I don't think we can set an interface for what variables are set at this point, the caller is going to have to do their own thing. If they choose to use something of ours, there are presently no guarantees it will continue working going forward.

Hook failure = upgrade failure, you will have to solve and re-run. Hooks will need to be able to accommodate that as upgrade does.

I do not expect hooks to return data, but they can set variables on hosts usable by your later hooks, that seems fine to me.

Does this sound ok, and if so where should I document it? At this point I think we have to keep it simple and expand on things later as need arises. I kind of need to see how it plays out for Ops needs as well. (working on that now)

dgoodwin:
This currently contains two commits from pr #3057 as they are dependent. Once that merges this PR will look cleaner.

inventory/byo/hosts.origin.example
detiber: > The hook runs against each master in serial. It can contain only tasks as it needs to plug into the middle of our serial upgrade.

I think that definitely needs to be documented. I can see some issues with this approach, since the master might not be able to reach the lb management interface or may not have access to the credentials needed. Of course that can be solved by using delegated tasks and possibly by setting variables prior to running the upgrade playbook, but I think it is important to document the behavior.

> I don't think we can set an interface for what variables are set at this point, the caller is going to have to do their own thing. If they choose to use something of ours, there are presently no guarantees it will continue working going forward.

I see this as a potential problem, but as long as we document this limitation, then at least the user has been warned.

> Hook failure = upgrade failure, you will have to solve and re-run. Hooks will need to be able to accommodate that as upgrade does.

Keep in mind that we'll need to provide some examples, so that users understand how to achieve that behavior. It would be bad if their hook consistently fails to remove a host from the LB that is already removed for example.

More importantly, how would it be handled if the 3rd host out of 3 was where the failure occurred after the 3rd host was removed from the LB? Would the 1st and 2nd host be removed from the LB prior to getting back to the 3rd host (which is already removed from the LB)?

I'm thinking we are going to have to store some state on the hosts (using ansible local facts) so that we can track the state of the upgrade. If we go that route, we may need to allow the hooks to have some say in setting their state, to allow for users to tweak hooks part way through a run and have that affect the system state.

> I do not expect hooks to return data, but they can set variables on hosts usable by your later hooks, that seems fine to me.

At some point we may have a need for hooks to inject data back into the upgrade. Simple LB operations may not require it, but something more complex like a SDN hook might. At the very least we should document that hooks are not able to affect the upgrade state today.

> Does this sound ok, and if so where should I document it? At this point I think we have to keep it simple and expand on things later as need arises. I kind of need to see how it plays out for Ops needs as well. (working on that now)

I think documenting it in the upgrade README would be a good place until we can get it ironed out and added to the upstream docs.

ashcrow: > I'm thinking we are going to have to store some state on the hosts (using ansible local facts) so that we can track the state of the upgrade. If we go that route, we may need to allow the hooks to have some say in setting their state, to allow for users to tweak hooks part way through a run and have that affect the system state.

:+1: 

> At some point we may have a need for hooks to inject data back into the upgrade. [...]

This makes sense to me, but I'd like to push it out to another chunk of work if possible.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -62,6 +64,10 @@
   roles:
   - openshift_facts
   post_tasks:
+
+  - include: "{{ openshift_upgrade_pre_master_hook }}"
```
ashcrow: :+1: to the simplicity of this.

I'm a paranoid guy when I run things. It may be nice to have an output statement letting the user know t heir hook is being executed. Obviously not a blocker.

dgoodwin: Sounds like a good idea to me. I have a number of other suggestions for present and future from a long chat with @detiber I need to add in here as well.

Just doing some testing to see how this works for ops right now, and then will return to finalize.

ashcrow: Sounds good.

dgoodwin:
@detiber suggested a few things, possibly for a future iteration.

(1) Use some kind of filter plugin to enforce some "hook API versioning". Idea is we would eventually commit to providing certain variables and extracting certain return values from hooks/plugins, and version this. The filter plugin would let us make sure hooks to be run are compatible with the current API supported in openshift-ansible.

(2) Some kind of library path support so various locations are checked for where to load the hooks from.

inventory/byo/hosts.origin.example
wester55: Hello and thank you for quick response!
We just finished successfully upgrade errata for latest 3.2->3.2 version. I want share real world scenario we faced.

We have 12 byo servers running HA Openshift 3.2. All 3 masters also host pods. Each one of 12 nodes has GlusterFS pod mounting local disks on server. When we run upgrade on staging environment each master upgrade also restarted node service, so after all 3 masters were upgraded, we lost whole gluster cluster since auto-heal was not so fast as upgrade moved on (we have replication factor of 3 in gluster). Additionally, upgrade decided to leave master nodes unscheduled even before we started all of them were scheduled and running pods. Also, upgrade detected HA and serialized master upgrade - in our case its problem too. So what we did before upgrade:

```
- hosts: primary_master
  tasks:
  - name: Alter stock update script1
    blockinfile:
      dest: /usr/share/ansible/openshift-ansible/playbooks/common/openshift-cluster/upgrades/upgrade.yml
      insertafter: '\s\swhen\:\ (|\"\{\{\s)inventory_hostname\ in\ groups\.oo_nodes_to_config\ and\ (openshift\.node\.|was_)schedulable\ \|\ bool(|\s\}\}\")$'
      marker: "#{mark}#INSERTED MANAGED BLOCK###"
      content: "  - command: >\n      {{ '{{' }} openshift.common.admin_binary {{ '}}' }} manage-node {{ '{{' }} openshift.common.hostname | lower {{ '}}' }} --schedulable=true\n    delegate_to: \"{{ '{{' }} groups.oo_first_master.0 {{ '}}' }}\"\n  - include: {{ ansible_scripts_dir }}/ansible-openshift/vars/custom_block.yml"
      state: present
    register: insert_result1

  - assert:
      that:
        - 'insert_result1.changed'
        - 'insert_result1.msg == "Block inserted"'
      msg: "Will not continue since we unable to alter original upgrade script1"

  - name: Alter stock update script2
    blockinfile:
      dest: /usr/share/ansible/openshift-ansible/playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
      marker: "#{mark}#INSERTED MANAGED BLOCK###"
      content: "- command: >\n      {{ '{{' }} openshift.common.admin_binary {{ '}}' }} manage-node {{ '{{' }} openshift.common.hostname | lower {{ '}}' }} --schedulable=true\n    delegate_to: \"{{ '{{' }} groups.oo_first_master.0 {{ '}}' }}\"\n- include: {{ ansible_scripts_dir }}/ansible-openshift/vars/custom_block.yml"
      state: present
    register: insert_result2

  - assert:
      that:
        - 'insert_result2.changed'
        - 'insert_result2.msg == "Block inserted"'
      msg: "Will not continue since we unable to alter original upgrade script2"

  - name: Alter stock update script3
    blockinfile:
      dest: /usr/share/ansible/openshift-ansible/playbooks/common/openshift-cluster/upgrades/upgrade.yml
      marker: "#{mark}#INSERTED MANAGED BLOCK2###"
      content: "  serial: 1"
      insertafter: '\s\s\s\swhen\:\ (|\"\{\{\s)ca_crt_stat\.stat\.isreg\ and\ not ca_bundle_stat\.stat\.exists(|\s\}\}\")$'
      state: present
    register: insert_result3

  - assert:
      that:
        - 'insert_result3.changed'
        - 'insert_result3.msg == "Block inserted"'
      msg: "Will not continue since we unable to alter original upgrade script3"
```
custom_block.yml has tasks we run after each gluster pod is restarted to make sure cluster heal completed on all pods. We put long retry and this helped us because we had major OS problem after one of nodes was upgraded. We even restarted this server several times and reinstalled few packages to solve the problem. Eventually when pod on this server was up, the retry succeeded and whole upgrade finished as expected.
 Just in any case here are contents of custom_block.yml:

```
- name: Delay to make sure nodes status updated
  pause: minutes=1

- set_fact: project="storage-project"

# This will check all nodes are ready
- name: Check Openshift nodes
  shell: oc get nodes | grep NotReady | wc -l
  register: ose_nodes_up
  retries: 43200
  delay: 60
  failed_when: "ose_nodes_up.stdout != '0'"
  until: ose_nodes_up.stdout == "0"

# This will check if one node is SchedulingDisabled
- name: Check Openshift node SchedulingDisabled
  shell: oc get nodes | grep SchedulingDisabled | wc -l
  register: ose_scheduling_disabled
  retries: 43200
  delay: 60
  failed_when: "ose_scheduling_disabled.rc != 0"
  until: ose_scheduling_disabled.rc == 0

# This will check all Gluster pods are ready
- name: Check Gluster pods
  shell: oc get pods --namespace={{ project }} | grep "0/1" | wc -l
  register: ose_pods_up
  retries: 43200
  delay: 60
  failed_when: "ose_pods_up.stdout != '0'"
  until: ose_pods_up.stdout == "0"
  when: "ose_scheduling_disabled.stdout == '0'"

- name: Get GlusterFS pods
  shell: oc get pods --namespace={{ project }} -o name | grep glusterfs | grep -v Completed | grep -v Terminating | cut -d'/' -f2
  register: pods
  when: "ose_scheduling_disabled.stdout == '0'"

- name: Build dictionary of pods
  set_fact:
    pods_dict: "{{ pods_dict|default({}) | combine( {item.split('.').0: item} ) }}"
  with_items: "{{ pods.stdout_lines | default([]) }}"
  when: "ose_scheduling_disabled.stdout == '0'"

- name: Get master-pod-volume pairs to check
  shell: oc exec {{ pods_dict[item] }} gluster volume list --namespace={{ project }}
  register: volumes
  with_items:
    - "{{ pods_dict.keys() | default([]) }}"
  retries: 43200
  delay: 60
  until: volumes.rc == 0
  when: "ose_scheduling_disabled.stdout == '0'"

# This will check that all nodes connected to Gluster
- name: Check Gluster peers
  shell: oc exec {{ pods_dict[item] }} gluster peer status --namespace={{ project }} | grep "(Disconnected)" | wc -l
  with_items:
    - '{{ pods_dict.keys() | default([]) }}'
  register: gluster_peers
  failed_when: gluster_peers.stdout != '0'
  retries: 43200
  delay: 60
  until: gluster_peers.stdout == '0'
  when: "ose_scheduling_disabled.stdout == '0'"

- name: Build dictionary of volumes
  set_fact:
    vols_dict: "{{ vols_dict|default({}) | combine( {item.item: { 'pod': item.item, 'volumes': item.stdout_lines} } ) }}"
  with_items: "{{ volumes.results | default([]) }}"
  when: 'ose_scheduling_disabled.stdout == "0" and item.stdout != ""'

# This will check if there is any ongoing task on any of volumes on any of Gluster pods. If it finds task it will loop for a month until task completes
- name: Check Gluster tasks
  shell: oc exec {{ pods_dict[item.0.pod] }} gluster volume status {{ item.1 }} tasks --namespace={{ project }} | grep "There are no active volume tasks" | wc -l
  with_subelements:
    - '{{ vols_dict | default([]) }}'
    - volumes
  register: gluster_tasks
  failed_when: gluster_tasks.stdout != "1"
  retries: 43200
  delay: 60
  until: gluster_tasks.stdout == "1"
  when: "ose_scheduling_disabled.stdout == '0'"

# This will check if any Gluster bricks are healing
- name: Check bricks healing
  shell: oc exec {{ pods_dict[item.0.pod] }} gluster volume heal {{ item.1 }} info --namespace={{ project }} | grep "Number of entries" | grep -v " 0" | wc -l
  with_subelements:
    - '{{ vols_dict | default([]) }}'
    - volumes
  register: healing_tasks
  failed_when: healing_tasks.stdout != "0"
  retries: 43200
  delay: 60
  until: healing_tasks.stdout == "0"
  when: "ose_scheduling_disabled.stdout == '0'"
```

dgoodwin: Good info thanks @wester55 

I think a few of these have been improved in 3.4 and more in 3.5 but boiling this down a bit:

- fix bug where schedulable masters are left unschedulable
- upgrade masters serially (just committed for 3.5, previously we did a bunch in parallel and then restarted services serially)
- implement hooks before and after each master is upgraded (this PR)
- similar hooks for each node

Additionally we plan to make master/control plane upgrade *also* handle the node/docker on the masters, and then omit these later when we're just upgrading nodes. This will make the process of upgrading a master more atomic.

dgoodwin:
Dropping WIP, seems to be working pretty good for what ops needs atm.

ashcrow:
aos-ci-test

detiber:
@dgoodwin just copying my comment from the docs PR. If we are running the post hook before we are restarting the services/system, should we add an additional hook that is triggered after the system/service restart? I don't think we'd want to put the host back into rotation in the LB prior to the service/system restart.

dgoodwin:
I dare say we absolutely must yeah, that is a good point. Lets hold off on merging this.

dgoodwin:
@mwoodson what do you think here, the point was raised that my post hook (excuted before system restart) is not an appropriate place to put a system back into an ELB. 

Would you prefer three hooks, pre, post, and some kind of "mid" or "pre-restart"?

Or should we do the full system yum update in pre (before even the atomic rpm update), and move post to after the system restart?

dgoodwin:
Back to WIP as I need to do a little more testing here Monday.

I moved the post hook to *after* system restart.

Instead I added just a regular "upgrade" hook to be run prior to service restart. This leaves us with pre upgrade hook, upgrade hook, and post upgrade hook. I feel this parallels ansible pre/post/tasks and should be familiar to users. If not I assume they'll be working off documentation in any case.

I also re-prefixed variables from openshift_upgrade_pre_master_hook to openshift_master_upgrade_pre_hook. This is because I forsee a openshift_master_upgrade role and it feels like this is the most logical variable prefixing given what's going on.

Please let me know what you think.

inventory/byo/hosts.origin.example
detiber: This needs to be updated for new variable naming

inventory/byo/hosts.origin.example
```diff
@@ -89,6 +89,21 @@ openshift_release=v1.4
...
+#
+# Tasks to run after each master is upgraded:
+# openshift_upgrade_post_master_hook=/usr/share/custom/post_master.yml
```
detiber: Same here, also looks like pre-service restart hook is missing.

inventory/byo/hosts.ose.example
```diff
@@ -89,6 +89,21 @@ openshift_release=v3.4
...
+# Hooks are available to run custom tasks at various points during a cluster
+# upgrade. Each hook should point to a file with Ansible tasks defined. Suggest using
+# absolute paths, if not the path will be treated as relative to the file where the
+# hook is actually used.
+#
+# Tasks to run before each master is upgraded:
+# openshift_upgrade_pre_master_hook=/usr/share/custom/pre_master.yml
+#
+# Tasks to run after each master is upgraded:
+# openshift_upgrade_post_master_hook=/usr/share/custom/post_master.yml
```
detiber: same as hosts.origin.example

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -62,6 +64,14 @@
   roles:
   - openshift_facts
   post_tasks:
+
+  # Run the pre-upgrade hook if defined:
+  - debug: msg="Running pre-master upgrade hook {{ openshift_master_upgrade_pre_hook }}"
```
detiber: s/pre-master upgrade hook/master pre-upgrade hook/

detiber: Instead of adding debug here, I think just adding `name: Running master pre-upgrade hook` to the below should be fine.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -102,12 +112,26 @@
       state: link
     when: ca_crt_stat.stat.isreg and not ca_bundle_stat.stat.exists
 
+  # Run the upgrade hook prior to restarting services/system if defined:
+  - debug: msg="Running upgrade hook {{ openshift_master_upgrade_hook }}"
```
detiber: I almost wonder if we need to still make this more descriptive. If for no other reason it would make it less confusing if we need to add additional hooks at other parts of the master upgrade process.

ashcrow: @detiber as in ``Running hook openshift_master_upgrade_hook: {{ openshift_master_upgrade_hook  }}``?

detiber: @ashcrow The variable interpolation wouldn't work in the task name. I'm good with having the debugs in place to ensure that we output the hook location in the output.

dgoodwin: I think the current state is about the best we can do but I'm open to more suggestions if people have them. We tried to come up with other naming on IRC and this was my preferred solution. I can't forsee needing more granularity right now, the only major things between "pre" and "upgrade" right now is systemd units and packages, the rest are predominantly one off upgrade tasks that will likely never run again. As it stands today I think we're pretty well covered, and we can adapt in the future if necessary. We can add new options in future and if necessary, deprecate one of these.

inventory/byo/hosts.origin.example
dgoodwin: Update incoming.

dgoodwin:
Updated per last comments.

dgoodwin:
Removed WIP, ready for merge or more review.

ashcrow:
aos-ci-test

ashcrow:
Travis failure is due to [this upstream issue](https://gitlab.com/pycqa/flake8/issues/301) and can be ignored.

dgoodwin:
Test clear but still needs an approval.

PR #3106, opened 2017-01-18
Add containzerized haproxy option

andrewklau:
Add containerised option for the openshift_loadbalancer

Simply uses the official Docker haproxy image and passes the same config file to the container `-v /etc/haproxy/haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro`

### DIFF
--- a/roles/openshift_loadbalancer/tasks/main.yml
+++ b/roles/openshift_loadbalancer/tasks/main.yml
@@ -1,14 +1,31 @@
 ---
-- fail: msg="Cannot use containerized=true for load balancer hosts."
-  when: openshift.common.is_containerized | bool
-
 - name: Install haproxy
   package: name=haproxy state=present
+  when: not openshift.common.is_containerized | bool
+
+- name: Pull haproxy image
+  command: >
+    docker pull {{ openshift.common.router_image }}:{{ openshift_image_tag }}
+  when: openshift.common.is_containerized | bool
+
+- name: Create config directory for haproxy
+  file:
+    path: /etc/haproxy
+    state: directory
+  when: openshift.common.is_containerized | bool
+
+- name: Create the systemd unit files
+  template:
+    src: "haproxy.docker.service.j2"
+    dest: "{{ containerized_svc_dir }}/haproxy.service"
+  when: openshift.common.is_containerized | bool
+  notify: restart haproxy
 
 - name: Configure systemd service directory for haproxy
   file:
     path: /etc/systemd/system/haproxy.service.d
     state: directory
+  when: not openshift.common.is_containerized | bool
 
 # Work around ini_file create option in 2.2 which defaults to no
 - name: Create limits.conf file
@@ -19,6 +36,7 @@
     owner: root
     group: root
   changed_when: false
+  when: not openshift.common.is_containerized | bool
 
 - name: Configure the nofile limits for haproxy
   ini_file:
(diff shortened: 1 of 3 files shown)


detiber:
@sdodson probably need to think about how to handle this for the OCP use case. I don't see an issue with using the docker hub image for Origin, but we would need to either figure out an alternative image for OCP (possibly an SCL haproxy image), or limit this to just Origin deployments.

sdodson:
Yeah, was going to ask for feedback internally. For now I think limiting it to origin is probably the best option, we really don't encourage use of the load balancer for production environments.

andrewklau:
It should be possible to use the origin-haproxy-router too as I just overwrite the config

andrewklau:
I updated it to use the router_image instead

detiber:
aos-ci-test

roles/openshift_loadbalancer/tasks/main.yml
```diff
@@ -1,14 +1,44 @@
...
+  when: openshift.common.is_containerized | bool
+  notify: restart haproxy
+
+- name: Configure systemd service directory for haproxy
+  file:
+    path: /etc/systemd/system/haproxy.service.d
+    state: directory
+  when: not openshift.common.is_containerized | bool
+
+- name: Create the systemd unit files
```
sdodson: This task seems un-necessary?

roles/openshift_loadbalancer/templates/haproxy.docker.service.j2
```diff
@@ -0,0 +1,17 @@
+[Unit]
+After=docker.service
+Requires=docker.service
+PartOf=docker.service
+
+[Service]
+ExecStartPre=-/usr/bin/docker rm -f openshift_loadbalancer
+ExecStart=/usr/bin/docker run --rm --name openshift_loadbalancer -p 8443:8443 -v /etc/haproxy/haproxy.cfg:/etc/haproxy/haproxy.cfg:ro --entrypoint="haproxy -f /etc/haproxy/haproxy.cfg" {{ openshift.common.router_image }}:{{ openshift_image_tag }}
```
abutcher: The port here should be based on `openshift_master_api_port` and default to 8443.

roles/openshift_loadbalancer/tasks/main.yml
andrewklau: this was from the original file https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_loadbalancer/tasks/main.yml

roles/openshift_loadbalancer/templates/haproxy.docker.service.j2
andrewklau: done

roles/openshift_loadbalancer/tasks/main.yml
andrewklau: oops..removed

detiber:
aos-ci-test

detiber:
@sdodson ptal, lgtm

andrewklau:
hmm I don't think the tests include the loadbalancer role

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1298787

sdodson:
We'll have QE test this. LGTM ,thanks!

PR #3107, opened 2017-01-18
Adding tox tests for generated module code.

kwoodson:


setup.py
```diff
@@ -145,6 +145,46 @@ def with_project_on_sys_path(self, func, func_args, func_kwargs):
...
+                                     'test',
+                                     'playbooks',
+                                     'utils'],
+                                    None,
+                                    'generate.py')
+
+        # call them with --verify
+        for gen in generate_files:
+            print('Checking generated module code: {0}'.format(gen))
+            subprocess.call([gen, '--verify'])
```
detiber: Would it make sense to capture the output/error code here and aggregate it, instead of failing on the first error?

kwoodson: @detiber,

I can add that.

kwoodson:
@detiber 

I have updated to process all files before failing.  I had to change the way the generate test worked as tox instantiates a virtualenv.  This does not work well with subprocessing.  Now I put the generate in the python path, import it, and run the verify step.  This works well.

PR #3109, opened 2017-01-18
System containers

giuseppe:
I am submitting these patches just to get some early feedbacks.

First of all, some custom images must be created to install OpenShift as system container:

https://github.com/giuseppe/atomic-openshift-system-containers

This is the command I am using for creating these images, where `localhost:5000` is a Docker registry running on my machine.

`$ SRC_REGISTRY=brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888 sudo -E ./generate_images.sh localhost:5000`

Once these images are built, which are just the original images for ose, node and openvswitch, I add the system container configuration files for running in a runc container.
Once these files get into the official builds, we don't need this step anymore (which is quite annoying).

For installing the system containers I am using this inventory file:
```
[OSEv3:children]
masters
nodes
etcd

[OSEv3:vars]

ansible_ssh_user=cloud-user
ansible_become=yes

#openshift_docker_additional_registries=brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888
#openshift_docker_insecure_registries=brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888
openshift_use_dnsmasq=False

#deployment_subtype=registry
#deployment_type=origin

openshift_router_selector='router=true'
openshift_registry_selector='registry=true'

openshift_image_tag=v3.3.0.26

openshift_master_default_subdomain=localorigin.io

openshift_docker_additional_registries=brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888
openshift_docker_insecure_registries=brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888

######SYSTEM CONTAINERS#######################################
system_images_registry=192.168.125.1:5000  # Docker registry with the custom images
use_system_containers=True
#use_openvswitch_system_container=True
#use_node_system_container=True
#use_master_system_container=True
##########################################################

registry_url=brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888

containerized=True
openshift_uninstall_images=False

# enable htpasswd auth
openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 'challenge': 'true', 'kind': 'HTPasswdPasswordIdentityProvider', 'filename': '/etc/origin/master/htpasswd'}]
openshift_master_htpasswd_users={'admin': '$apr1$zgSjCrLt$1KSuj66CggeWSv.D.BXOA1', 'user': '$apr1$.gw8w9i1$ln9bfTRiD6OwuNTG5LvW50'}

# host group for masters
[masters]
192.168.125.89 openshift_schedulable=true 

[etcd]
192.168.125.89

# host group for nodes
[nodes]
192.168.125.89  openshift_schedulable=true openshift_node_labels="{'router':'true','registry':'true'}"
```

Is it possible to switch single components to be system containers.  `use_system_containers=True` will use system containers for any component supported.

The idempotent update is left to "atomic update --containers (atomic containers update in the upstream version)" that reinstall the container only when a different IMAGE ID is available, so if a container is rebuilt this will trigger the update.

An uninstall/install is done when the version is different or the container is not installed.

### DIFF
--- a/playbooks/adhoc/uninstall.yml
+++ b/playbooks/adhoc/uninstall.yml
@@ -164,9 +164,12 @@
     - atomic-enterprise
     - origin
 
-  - shell: atomic uninstall openvswitch
+  - shell: atomic uninstall "{{ item }}"
     changed_when: False
     failed_when: False
+    with_items:
+    - etcd
+    - openvswitch
 
   - shell: find /var/lib/origin/openshift.local.volumes -type d -exec umount {} \; 2>/dev/null || true
     changed_when: False
--- a/playbooks/adhoc/uninstall.yml
+++ b/playbooks/adhoc/uninstall.yml
@@ -148,6 +148,26 @@
       - vovsbr
     when: "{{ openshift_remove_all | default(true) | bool }}"
 
+  - shell: atomic uninstall "{{ item }}"-master
+    changed_when: False
+    failed_when: False
+    with_items:
+    - openshift-enterprise
+    - atomic-enterprise
+    - origin
+
+  - shell: atomic uninstall "{{ item }}"-node
+    changed_when: False
+    failed_when: False
+    with_items:
+    - openshift-enterprise
+    - atomic-enterprise
+    - origin
+
+  - shell: atomic uninstall openvswitch
+    changed_when: False
+    failed_when: False
+
   - shell: find /var/lib/origin/openshift.local.volumes -type d -exec umount {} \; 2>/dev/null || true
     changed_when: False
 
@@ -262,6 +282,9 @@
     - /etc/systemd/system/origin-node.service.wants
     - /var/lib/atomic-enterprise
     - /var/lib/openshift
+
+  - shell: systemctl daemon-reload
+    changed_when: False
 
   - name: restart docker
     service: name=docker state=restarted
--- a/roles/etcd/tasks/main.yml
+++ b/roles/etcd/tasks/main.yml
@@ -14,13 +14,17 @@
   command: docker pull {{ openshift.etcd.etcd_image }}
   register: pull_result
   changed_when: "'Downloaded newer image' in pull_result.stdout"
-  when: etcd_is_containerized | bool
+  when:
+  - etcd_is_containerized | bool
+  - not openshift.common.is_etcd_system_container | bool
 
 - name: Install etcd container service file
   template:
     dest: "/etc/systemd/system/etcd_container.service"
     src: etcd.docker.service
-  when: etcd_is_containerized | bool
+  when:
+  - etcd_is_containerized | bool
+  - not openshift.common.is_etcd_system_container | bool
 
 - name: Ensure etcd datadir exists when containerized
   file:
@@ -36,9 +40,21 @@
     enabled: no
     masked: yes
     daemon_reload: yes
-  when: etcd_is_containerized | bool
+  when:
+  - etcd_is_containerized | bool
+  - not openshift.common.is_etcd_system_container | bool
   register: task_result
   failed_when: "task_result|failed and 'could not' not in task_result.msg|lower"
+
+- name: Install etcd container service file
+  template:
+    dest: "/etc/systemd/system/etcd_container.service"
+    src: etcd.docker.service
+  when: etcd_is_containerized | bool and not openshift.common.is_etcd_system_container | bool
+
+- name: Install Etcd system container
+  include: system_container.yml
+  when: etcd_is_containerized | bool and openshift.common.is_etcd_system_container | bool
 
 - name: Validate permissions on the config dir
   file:
@@ -54,7 +70,7 @@
     dest: /etc/etcd/etcd.conf
     backup: true
   notify:
-    - restart etcd
+  - restart etcd
 
 - name: Enable etcd
   systemd:
--- a/roles/etcd/tasks/system_container.yml
+++ b/roles/etcd/tasks/system_container.yml
@@ -1,0 +1,63 @@
+---
+- name: Pull etcd system container
+  command: atomic pull --storage=ostree {{ openshift.etcd.etcd_image }}
+  register: pull_result
+  changed_when: "'Pulling layer' in pull_result.stdout"
+
+- name: Check etcd system container package
+  command: >
+    atomic containers list --no-trunc -a -f container=etcd
+  register: result
+
+- name: Set initial Etcd cluster
+  set_fact:
+    etcd_initial_cluster: >
+      {% for host in etcd_peers | default([]) -%}
+      {% if loop.last -%}
+      {{ hostvars[host].etcd_hostname }}={{ etcd_peer_url_scheme }}://{{ hostvars[host].etcd_ip }}:{{ etcd_peer_port }}
+      {%- else -%}
+      {{ hostvars[host].etcd_hostname }}={{ etcd_peer_url_scheme }}://{{ hostvars[host].etcd_ip }}:{{ etcd_peer_port }},
+      {%- endif -%}
+      {% endfor -%}
+
+- name: Update Etcd system container package
+  command: >
+    atomic containers update
+    --set ETCD_LISTEN_PEER_URLS={{ etcd_listen_peer_urls }}
+    --set ETCD_NAME={{ etcd_hostname }}
+    --set ETCD_INITIAL_CLUSTER={{ etcd_initial_cluster | replace('\n', '') }}
+    --set ETCD_LISTEN_CLIENT_URLS={{ etcd_listen_client_urls }}
+    --set ETCD_INITIAL_ADVERTISE_PEER_URLS={{ etcd_initial_advertise_peer_urls }}
+    --set ETCD_INITIAL_CLUSTER_STATE={{ etcd_initial_cluster_state }}
+    --set ETCD_INITIAL_CLUSTER_TOKEN={{ etcd_initial_cluster_token }}
+    --set ETCD_ADVERTISE_CLIENT_URLS={{ etcd_advertise_client_urls }}
+    --set ETCD_CA_FILE={{ etcd_system_container_conf_dir }}/ca.crt
+    --set ETCD_CERT_FILE={{ etcd_system_container_conf_dir }}/server.crt
+    --set ETCD_KEY_FILE={{ etcd_system_container_conf_dir }}/server.key
+    --set ETCD_PEER_CA_FILE={{ etcd_system_container_conf_dir }}/ca.crt
+    --set ETCD_PEER_CERT_FILE={{ etcd_system_container_conf_dir }}/peer.crt
+    --set ETCD_PEER_KEY_FILE={{ etcd_system_container_conf_dir }}/peer.key
+    etcd
+  when:
+  - ("etcd" in result.stdout)
+
+- name: Install Etcd system container package
+  command: >
+    atomic install --system --name=etcd
+    --set ETCD_LISTEN_PEER_URLS={{ etcd_listen_peer_urls }}
+    --set ETCD_NAME={{ etcd_hostname }}
+    --set ETCD_INITIAL_CLUSTER={{ etcd_initial_cluster | replace('\n', '') }}
+    --set ETCD_LISTEN_CLIENT_URLS={{ etcd_listen_client_urls }}
+    --set ETCD_INITIAL_ADVERTISE_PEER_URLS={{ etcd_initial_advertise_peer_urls }}
+    --set ETCD_INITIAL_CLUSTER_STATE={{ etcd_initial_cluster_state }}
+    --set ETCD_INITIAL_CLUSTER_TOKEN={{ etcd_initial_cluster_token }}
+    --set ETCD_ADVERTISE_CLIENT_URLS={{ etcd_advertise_client_urls }}
+    --set ETCD_CA_FILE={{ etcd_system_container_conf_dir }}/ca.crt
+    --set ETCD_CERT_FILE={{ etcd_system_container_conf_dir }}/server.crt
+    --set ETCD_KEY_FILE={{ etcd_system_container_conf_dir }}/server.key
+    --set ETCD_PEER_CA_FILE={{ etcd_system_container_conf_dir }}/ca.crt
+    --set ETCD_PEER_CERT_FILE={{ etcd_system_container_conf_dir }}/peer.crt
+    --set ETCD_PEER_KEY_FILE={{ etcd_system_container_conf_dir }}/peer.key
+    {{ openshift.etcd.etcd_image }}
+  when:
+  - ("etcd" not in result.stdout)
--- a/roles/etcd_common/defaults/main.yml
+++ b/roles/etcd_common/defaults/main.yml
@@ -1,6 +1,7 @@
 ---
 # etcd server vars
-etcd_conf_dir: /etc/etcd
+etcd_conf_dir: "{{ '/etc/etcd' if not openshift.common.is_etcd_system_container else '/var/lib/etcd/etcd.etcd/etc'  }}"
+etcd_system_container_conf_dir: /var/lib/etcd/etc
 etcd_ca_file: "{{ etcd_conf_dir }}/ca.crt"
 etcd_cert_file: "{{ etcd_conf_dir }}/server.crt"
 etcd_key_file: "{{ etcd_conf_dir }}/server.key"
--- a/roles/openshift_facts/tasks/main.yml
+++ b/roles/openshift_facts/tasks/main.yml
@@ -9,6 +9,9 @@
     l_is_atomic: "{{ ostree_booted.stat.exists }}"
 - set_fact:
     l_is_containerized: "{{ (l_is_atomic | bool) or (containerized | default(false) | bool) }}"
+    l_is_openvswitch_system_container: "{{ (use_openvswitch_system_container | default(use_system_containers) | bool) }}"
+    l_is_node_system_container: "{{ (use_node_system_container | default(use_system_containers) | bool) }}"
+    l_is_master_system_container: "{{ (use_master_system_container | default(use_system_containers) | bool) }}"
 
 - name: Ensure various deps are installed
   package: name={{ item }} state=present
@@ -27,6 +30,10 @@
       hostname: "{{ openshift_hostname | default(None) }}"
       ip: "{{ openshift_ip | default(None) }}"
       is_containerized: "{{ l_is_containerized | default(None) }}"
+      is_openvswitch_system_container: "{{ l_is_openvswitch_system_container | default(false) }}"
+      is_node_system_container: "{{ l_is_node_system_container | default(false) }}"
+      is_master_system_container: "{{ l_is_master_system_container | default(false) }}"
+      system_images_registry: "{{ system_images_registry | default('') }}"
       public_hostname: "{{ openshift_public_hostname | default(None) }}"
       public_ip: "{{ openshift_public_ip | default(None) }}"
       portal_net: "{{ openshift_portal_net | default(openshift_master_portal_net) | default(None) }}"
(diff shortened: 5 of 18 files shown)


giuseppe:
@ashcrow here it is :)

roles/openshift_node/tasks/systemd_units.yml
```diff
@@ -28,7 +43,42 @@
...
-  when: openshift.common.is_containerized | bool
+  when: openshift.common.is_containerized | bool and not openshift.common.is_node_system_container | bool
+  notify:
+    - restart node
+
+# System container
+- name: Check Node system container package
+  command: >
+    atomic containers list --no-trunc -a -f container={{ openshift.common.service_type }}-node
+  register: result
```
ashcrow: nit: Same generic register note as above.

roles/openshift_node/tasks/systemd_units.yml
```diff
@@ -7,14 +7,29 @@
     docker pull {{ openshift.node.node_image }}:{{ openshift_image_tag }}
   register: pull_result
   changed_when: "'Downloaded newer image' in pull_result.stdout"
-  when: openshift.common.is_containerized | bool
+  when: openshift.common.is_containerized | bool and not openshift.common.is_node_system_container | bool
```
detiber: I'd suggest either separating out system container/non-system container tasks into separate included files or into blocks, to make the separation of tasks clearer.

detiber:
Copying the travis ci failure here:
```
roles/openshift_node/tasks/systemd_units.yml:
  90:3      error    wrong indentation: expected 4 but found 2  (indentation)
  105:3     error    wrong indentation: expected 4 but found 2  (indentation)
  113:3     error    wrong indentation: expected 4 but found 2  (indentation)
  154:3     error    wrong indentation: expected 4 but found 2  (indentation)
  161:3     error    wrong indentation: expected 4 but found 2  (indentation)
  170:3     error    wrong indentation: expected 4 but found 2  (indentation)
  178:3     error    wrong indentation: expected 4 but found 2  (indentation)
  184:3     error    wrong indentation: expected 4 but found 2  (indentation)
yammlint issues found
ERROR: InvocationError: '/home/travis/build/openshift/openshift-ansible/.tox/py27-ansible22-yamllint/bin/python setup.py yamllint'
```

roles/openshift_facts/tasks/main.yml
```diff
@@ -9,6 +9,10 @@
     l_is_atomic: "{{ ostree_booted.stat.exists }}"
 - set_fact:
     l_is_containerized: "{{ (l_is_atomic | bool) or (containerized | default(false) | bool) }}"
+    l_is_system_containers: "{{ (use_system_containers | default(None) | bool) }}"
```
giuseppe: I thought that was a bit too "aggressive" and still have the possibility to install as Docker containers if needed.  Maybe we should keep both for a while?

playbooks/adhoc/uninstall.yml
```diff
@@ -55,6 +55,10 @@
     - pcsd
     failed_when: false
 
+  - shell: "atomic uninstall atomic-openshift-master"
+    changed_when: False
```
detiber: Instead of setting changed_when  to false, might be better to have failed when parse the rc or output indicates that a container was actually uninstalled.

giuseppe:
@detiber Thanks for the feedback, we already have a system container for etcd, could we use that?

I've changed the container that is now enough to restart the service to read the modified environment variables so there won't be need to reinstall it again.

detiber:
> @detiber Thanks for the feedback, we already have a system container for etcd, could we use that?

I do not see why not. As far as I know the existing etcd image we use is pretty generic.

> I've changed the container that is now enough to restart the service to read the modified environment variables so there won't be need to reinstall it again.

:+1:

playbooks/adhoc/uninstall.yml
ashcrow: I'm good with @detiber's suggestion.

giuseppe: I've used the existing pattern that is already used for Docker containers.  It looks like uninstall.yml is trying to do "remove as much as possible and never fail":

For example, this is done with Docker containers:
```
  - shell: docker rm -f "{{ item }}"-master "{{ item }}"-node
    changed_when: False
    failed_when: False
    with_items:
    - openshift-enterprise
    - atomic-enterprise
    - origin
```

ashcrow: @abutcher it looks like you were the fella to add this in ``playbooks/adhoc/uninstall.yml``. Do you recommend reusing ``changed_when: False`` in this kind of case?

giuseppe:
where should the configuration file for running these images as system containers be added?

I am currently maintaining them here:

https://github.com/giuseppe/atomic-openshift-system-containers

detiber:
@giuseppe They should be added to: https://github.com/openshift/origin/tree/master/images

giuseppe:
@detiber thanks, I had a look at that images, but I cannot figure out which image is used by `master` (i.e. the `openshift3/ose` image)

detiber:
@giuseppe `origin` is the equivalent image for origin. You may want to sync up with @smunilla or @tdawson for how the origin builds differ from ocp builds.

playbooks/adhoc/uninstall.yml
abutcher: :+1: to @detiber's suggestion

giuseppe: @abutcher thanks for the review.  Still it is not clear to me why this should be different than what we do with the docker containers.  We just delete them.  In fact, uninstall.yml doesn't even check if the containers exist before trying to remove them.  Who would use this information?

In my local development version I made some changes for it to work with the "origin" images as well.  This is what I have now (same thing done for -node):

```
  - shell: atomic uninstall "{{ item }}"-master
    changed_when: False
    failed_when: False
    with_items:
    - openshift-enterprise
    - atomic-enterprise
    - origin
```

abutcher: @giuseppe That makes sense and, in sticking with the current convention, I don't think this needs to block this change but I do like seeing changed status when things have actually changed (if it's feasible to do).

I'm 99% sure the uninstall playbook is unsupported but perhaps we could follow up on this wrt the idempotentcy work that is being done. Uninstall won't be ran consecutively (no need to confirm that the playbook is doing unnecessary work on the N+1th run) so I'd consider this extremely low priority.

tbielawa: > I do like seeing changed status when things have actually changed (if it's feasible to do).

:+1: · 1,000,000

> I don't think this needs to block this change

:+1: · 1,000,000

giuseppe:
I've done some changes so that it works with Origin as well.  I've opened a PR to add the files to Origin so that we can start using origin, node, openvswitch as system containers:

https://github.com/openshift/origin/pull/12738

giuseppe:
I've pushed another version.  I've fixed some issues and now the scripts should work also when not installing as system containers.  Can we run the tests on this PR?

ashcrow:
Looks like ``yamllint`` is not passing. Here are the errors:

```
roles/etcd/tasks/main.yml:
  1:1       error    too many blank lines (1 > 0)  (empty-lines)
...
roles/openshift_master/vars/main.yml:
  20:1      error    wrong indentation: expected 2 but found 0  (indentation)
```

Update and test locally with ``tox -e py27-ansible22-yamllint``.

ashcrow:
aos-ci-test

giuseppe:
@ashcrow thanks for the info, I've fixed these issues and dropped some patches from this series that were not relevant for the system containers stuff itself

giuseppe:
aos-ci-test

giuseppe:
I've moved one patch to a new PR: https://github.com/openshift/openshift-ansible/pull/3262

ashcrow:
aos-ci-test

ashcrow:
Failure in https://github.com/openshift/openshift-ansible/pull/3109#issuecomment-277718130 is not related to the code. It looks like a problem with the CI environment

```
2017-02-06 10:30:27,204 - tasks.nodefactory - INFO - No floating IP to release found.
2017-02-06 10:30:27,204 - taskrunner - ERROR - Caught exception while running 'GetNodes'
```

giuseppe:
https://github.com/openshift/origin/pull/12738 got merged.  It should be possible now to use images from Docker Hub to try this out

ashcrow:
aos-ci-test

roles/openshift_node/tasks/systemd_units.yml
ashcrow: @giuseppe Do you want to do what @detiber suggested in this PR or would it make sense to do it later?

giuseppe: @ashcrow no, it is fine, let's fix it now.  I have some WIP for moving the Docker container related stuff into blocks, I am going to clean it up, test and post it later

giuseppe:
@ashcrow I added two patches to refactor the Docker containers code in a block (I have tested locally only the system containers part).  Also for using origin:latest, https://github.com/openshift/openshift-ansible/pull/3287 is required.

ashcrow:
Let's try this again ...

aos-ci-test

giuseppe:
aos-ci-test

giuseppe:
aos-ci-test

sdodson:
```
TASK [openshift_master : Create the systemd unit files] ************************
fatal: [10.8.169.235]: FAILED! => {"changed": true, "failed": true, "msg": "unsupported parameter for module: when"}
```

giuseppe:
@sdodson thanks.  I dropped the "refactor master" patch.  I'll have to do this in a different way for support HA in system containers.  I'll get to that once we get this first part done.

ashcrow:
aos-ci-test

roles/etcd/tasks/main.yml
```diff
@@ -36,10 +40,21 @@
...
   failed_when: "task_result|failed and 'could not' not in task_result.msg|lower"
 
+- name: Install etcd container service file
+  template:
+    dest: "/etc/systemd/system/etcd_container.service"
+    src: etcd.docker.service
+  when: etcd_is_containerized | bool and not openshift.common.is_etcd_system_container | bool
+
+- name: Install Etcd system container
+  include: system_container.yml
```
detiber: instead of having all of the tasks in system_container.yml test if they should run, you can apply the when condition to the include and have it apply to all tasks.

roles/etcd/tasks/system_container.yml
```diff
@@ -0,0 +1,74 @@
...
+  - etcd_is_containerized | bool
+  - openshift.common.is_etcd_system_container | bool
+
+- name: Check etcd system container package
+  command: >
+    atomic containers list --no-trunc -a -f container=etcd
+  register: result
+  when:
+  - openshift.common.is_containerized | bool
+  - openshift.common.is_node_system_container | bool
```
detiber: s/node/etcd/

roles/etcd_common/defaults/main.yml
```diff
@@ -1,6 +1,7 @@
 ---
 # etcd server vars
-etcd_conf_dir: /etc/etcd
+etcd_conf_dir: "{{ '/etc/etcd' if not openshift.common.is_etcd_system_container else '/var/lib/etcd/etcd.etcd/etc'  }}"
+etcd_system_container_conf_dir: /var/lib/etcd/etc
```
detiber: These variables seem to conflict a bit.

Any reason not to just use /etc/etcd for the system container? I worry about using alternative paths if we do not absolutely need to.

roles/openshift_facts/tasks/main.yml
```diff
@@ -7,8 +7,16 @@
 # Locally setup containerized facts for now
 - set_fact:
     l_is_atomic: "{{ ostree_booted.stat.exists }}"
+
+- set_fact:
+    l_is_system_containers: "{{ use_system_containers | default(false) }}"
```
detiber: Any reason not to add `use_system_containers: false` to defaults/main.yml and then just use `use_system_containers` below?

roles/openshift_facts/tasks/main.yml
```diff
@@ -26,7 +34,12 @@
       cluster_id: "{{ openshift_cluster_id | default('default') }}"
       hostname: "{{ openshift_hostname | default(None) }}"
       ip: "{{ openshift_ip | default(None) }}"
-      is_containerized: "{{ l_is_containerized | default(None) }}"
+      is_containerized: "{{ l_is_containerized | default(l_is_system_containers) }}"
```
detiber: l_is_containerized is always going to be set because of the set_fact above, should be able to just use `l_is_containerized or l_is_systemcontainers` here.

Only time default is really needed is if the value might not be set, or if we are trying to make sure it isn't any empty string, list, etc.

roles/openshift_master/tasks/main.yml
```diff
@@ -131,6 +131,9 @@
 - name: Install the systemd units
   include: systemd_units.yml
 
+- name: Install Master system container
+  include: system_container.yml
```
detiber: would suggest putting the common when clauses in system_container.yml on this include.

roles/openshift_master/tasks/systemd_units.yml
```diff
@@ -20,14 +20,14 @@
     docker pull {{ openshift.master.master_image }}:{{ openshift_image_tag }}
   register: pull_result
   changed_when: "'Downloaded newer image' in pull_result.stdout"
-  when: openshift.common.is_containerized | bool
+  when: openshift.common.is_containerized | bool and not openshift.common.is_master_system_container | bool
```
detiber: I'd push these when conditions up to the include task.

roles/etcd_common/defaults/main.yml
giuseppe: unfortunately the rhel/etcd system container reads the `/etc` from the container itself.  I thought of using this for now and then change the etcd system container to accept optionally a bind mount from the host `/etc`

detiber: @sdodson what are your thoughts here?

giuseppe:
@detiber thanks for the review, I've addressed your comments in the version here ⬆️ (except the etcd `/etc` mount point which requires changes in `rhel/etcd`)

roles/openshift_node/tasks/system_container.yml
```diff
@@ -0,0 +1,83 @@
+---
+- name: Pre-pull node system container image
```
detiber: I think either using blocks or breaking this out into two separate task includes would help with the readability here

giuseppe: I've splitted it into two different files

giuseppe:
aos-ci-test

giuseppe:
I've added a commit to get HA master with system containers.

@ashcrow I saw that in the meanwhile you approved the changes, thanks! :)

Is it fine to leave the patch or do I move it to another PR?

ashcrow:
@giuseppe It's fine to leave the patch but I'll need to rereview and it will need to pass the CI checks :smile:. However, I'd do any other additions in other PRs to keep this one "small" and mergable.

giuseppe:
aos-ci-test

roles/openshift_node/tasks/node_system_container.yml
```diff
@@ -0,0 +1,34 @@
...
+    atomic containers list --no-trunc -a -f container={{ openshift.common.service_type }}-node
+  register: result
+
+- name: Update Node system container package
+  command: >
+    atomic containers update {{ openshift.common.service_type }}-node
+  register: update_result
+  changed_when: "'Extracting' in update_result.stdout"
+  when:
+  - (openshift.common.version is defined) and (openshift.common.version == openshift_version) and ("node" in result.stdout) | bool
```
detiber: For a followup, consider switching this to use a list of conditions:
```
when:
- openshift.common.version is defined
- openshift.common.version == openshift_version
- "node" in result.stdout | bool

File: roles/openshift_node/tasks/node_system_container.yml [hunk 14]
```diff
@@ -0,0 +1,34 @@
...
+  when:
+  - (openshift.common.version is not defined) or (openshift.common.version != openshift_version) and ("node" in result.stdout) | bool
+
+- name: Install Node system container package
+  command: >
+    atomic install --system --name={{ openshift.common.service_type }}-node {{ openshift.common.system_images_registry }}/{{ openshift.node.node_system_image }}:{{ openshift_image_tag }}
+  register: install_node_result
+  changed_when: "'Extracting' in pull_result.stdout"
+  when:
+  - (openshift.common.version is not defined) or (openshift.common.version != openshift_version) or ("node" not in result.stdout) | bool
```
detiber: Potential followup: consider a block for the above two command

roles/openshift_node/tasks/openvswitch_system_container.yml
```diff
@@ -0,0 +1,36 @@
...
+    atomic pull --storage=ostree {{ openshift.common.system_images_registry }}/{{ openshift.node.ovs_system_image }}:{{ openshift_image_tag }}
+  register: pull_result
+  changed_when: "'Pulling layer' in pull_result.stdout"
+
+- name: Check OpenvSwitch system container package
+  command: >
+    atomic containers list --no-trunc -a -f container=openvswitch
+  register: result
+  when:
+  - openshift.common.is_openvswitch_system_container | bool
```
detiber: Followup: This condition is not needed, since it is applied on the include.

roles/openshift_node/tasks/openvswitch_system_container.yml
```diff
@@ -0,0 +1,36 @@
...
+
+- name: Update OpenvSwitch system container package
+  command: >
+    atomic containers update openvswitch
+  register: update_result
+  changed_when: "'Extracting' in update_result.stdout"
+  when:
+  - (openshift.common.version is defined) and (openshift.common.version == openshift_version) and ("openvswitch" in result.stdout) | bool
```
detiber: Followup: consider using a list of conditions to increase readability.

roles/openshift_node/tasks/openvswitch_system_container.yml
```diff
@@ -0,0 +1,36 @@
...
+  when:
+  - (openshift.common.version is not defined) or (openshift.common.version != openshift_version) and ("openvswitch" in result.stdout) | bool
+
+- name: Install OpenvSwitch system container package
+  command: >
+    atomic install --system --name=openvswitch {{ openshift.common.system_images_registry }}/{{ openshift.node.ovs_system_image }}:{{ openshift_image_tag }}
+  when:
+  - (openshift.common.version is not defined) or (openshift.common.version != openshift_version) or ("openvswitch" not in result.stdout) | bool
+  notify:
+  - restart docker
```
detiber: Followup: consider using a block to avoid duplicating the conditions on both tasks.

It also looks like this pattern is used in multiple places, It might be a good idea to break it out into a role and use include_role to do this in a re-usable parameterized way.

That said, a module should probably be considered the preferred approach. I know there is an upstream module for atomic, maybe we should consider having it extended to support system containers, and then we could vendor it in our repo until we can require a version of Ansible that ships with support.

roles/etcd/tasks/system_container.yml
```diff
@@ -0,0 +1,63 @@
...
+- name: Set initial Etcd cluster
+  set_fact:
+    etcd_initial_cluster: >
+      {% for host in etcd_peers | default([]) -%}
+      {% if loop.last -%}
+      {{ hostvars[host].etcd_hostname }}={{ etcd_peer_url_scheme }}://{{ hostvars[host].etcd_ip }}:{{ etcd_peer_port }}
+      {%- else -%}
+      {{ hostvars[host].etcd_hostname }}={{ etcd_peer_url_scheme }}://{{ hostvars[host].etcd_ip }}:{{ etcd_peer_port }},
+      {%- endif -%}
+      {% endfor -%}
```
detiber: Followup: Why not put this in a separate template (say etcd_initial_cluster.j2) and then use `lookup('template', 'etcd_initial_cluster.j2)`?

roles/openshift_node/tasks/openvswitch_system_container.yml
giuseppe: yes, I agree, I'd like to clean this up in a module, I am also doing some work upstream for system containers so that it is possible to simplify this even more.  I want to remove the uninstall/install logic completely.

https://github.com/projectatomic/atomic/pull/883

Once that get merged, it will be possible to update to a different tag in the same update command, so we can simply install when not already installed and update in any other case.

PR #3110, opened 2017-01-18
update examples to cover build default/override configuration

bparees:


bparees:
aos-ci-test

@sdodson ptal

PR #3112, opened 2017-01-18
kubelet must have rw to cgroups for pod/qos cgroups to function

derekwaynecarr:
Related:
https://github.com/openshift/origin/pull/12545


derekwaynecarr:
fyi @sdodson

sdodson:
aos-ci-test

PR #3125, opened 2017-01-19
HOOKS.md added documenting new hooks

ashcrow:
Follow on documentation for https://github.com/openshift/openshift-ansible/pull/3101

### DIFF
--- a/HOOKS.md
+++ b/HOOKS.md
@@ -1,0 +1,70 @@
+# Hooks
+
+The ansible installer allows for operators to execute custom tasks during
+specific operations through a system called hooks. Hooks allow operators to
+provide files defining tasks to execute before and/or after specific areas
+during installations and upgrades. This can be very helpful to validate
+or modify custom infrastructure when installing/upgrading OpenShift.
+
+It is important to remember that when a hook fails the operation fails. This
+means a good hook can run multiple times and provide the same results. A great
+hook is idempotent.
+
+**Note**: There is currently not a standard interface for hooks. In the future
+a standard interface will be defined and any hooks that existed previously will
+need to be updated to meet the new standard.
+
+## Using Hooks
+
+Hooks are defined in the ``hosts`` inventory file under the ``OSEv3:vars``
+section.
+
+Each hook should point to a yaml file which defines Ansible tasks. This file
+will be used as an include meaning that the file can not be a playbook but
+a set of tasks. Best practice suggests using absolute paths to the hook file to avoid any ambiguity.
+
+### Example
+```ini
+[OSEv3:vars]
+# <snip>
+openshift_master_upgrade_pre_hook=/usr/share/custom/pre_master.yml
+openshift_master_upgrade_hook=/usr/share/custom/master.yml
+openshift_master_upgrade_post_hook=/usr/share/custom/post_master.yml
+# <snip>
+```
+
+Hook files must be a yaml formatted file that defines a set of Ansible tasks.
+The file may **not** be a playbook.
+
+### Example
+```yaml
+---
+# Trivial example forcing an operator to ack the start of an upgrade
+# file=/usr/share/custom/pre_master.yml
+
+- name: note the start of a master upgrade
+  debug:
+      msg: "Master upgrade of {{ inventory_hostname }} is about to start"
+
+- name: require an operator agree to start an upgrade
+  pause:
+      prompt: "Hit enter to start the master upgrade"
+```
+
+## Upgrade Hooks
+
+### openshift_master_upgrade_pre_hook
+- Runs **before** each master is upgraded.
+- This hook runs against **each master** in serial.
+- If a task needs to run against a different host, said task will need to use [``delegate_to`` or ``local_action``](http://docs.ansible.com/ansible/playbooks_delegation.html#delegation).
+
+### openshift_master_upgrade_hook
+- Runs **after** each master is upgraded but **before** it's service/system restart.
+- This hook runs against **each master** in serial.
+- If a task needs to run against a different host, said task will need to use [``delegate_to`` or ``local_action``](http://docs.ansible.com/ansible/playbooks_delegation.html#delegation).
+
+
+### openshift_master_upgrade_post_hook
+- Runs **after** each master is upgraded and has had it's service/system restart.
+- This hook runs against **each master** in serial.
+- If a task needs to run against a different host, said task will need to use [``delegate_to`` or ``local_action``](http://docs.ansible.com/ansible/playbooks_delegation.html#delegation).
--- a/README.md
+++ b/README.md
@@ -74,6 +74,12 @@ you are not running a stable release.
       - [OpenShift Enterprise](https://docs.openshift.com/enterprise/latest/install_config/install/advanced_install.html)
       - [OpenShift Origin](https://docs.openshift.org/latest/install_config/install/advanced_install.html)
 
+
+## Installer Hooks
+
+See the [hooks documentation](HOOKS.md).
+
+
 ## Contributing
 
 See the [contribution guide](CONTRIBUTING.md).


HOOKS.md
```diff
@@ -0,0 +1,59 @@
...
+during installations and upgrades. This can be very helpful to validate
+or modify custom infrastructure when installing/upgrading OpenShift.
+
+It is important to remember that when a hook fails the operation fails. This
+means a good hook can run multiple times and provide the same results. A great
+hook is idempotent.
+
+**Note**: There is currently not a standard interface for hooks. In the future
+a standard interface will be defined and any hooks that existed previously will
+be updated to meet the new standard.
```
detiber: s/will be updated/will need to be updated/ ?

HOOKS.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Using Hooks
+
+Hooks are defined in the ``hosts`` inventory file under the ``OSEv3:vars``
+section.
+
+Each hook should point to a yaml file which defines Ansible tasks. Best
+practice suggests using absolute paths to the hook file to avoid any ambiguity.
```
detiber: Probably needs to be clarified to state that the hooks are an included tasks file rather than a playbook.

Also, needs to say that they are run against each of the master hosts, so tasks that would need to be run against a different host will need to use `delegate_to` or `local_action`

HOOKS.md
ashcrow: I think either is OK. I'd think we'd define the standard and bring the existing hooks up to it. However, I'll update it :-)

HOOKS.md
ashcrow: Will do.

ashcrow:
@ingvagabund Once you and @dgoodwin have reviewed as well I think this will be ready to go post https://github.com/openshift/openshift-ansible/pull/3101 merge.

HOOKS.md
```diff
@@ -0,0 +1,64 @@
...
+### openshift_upgrade_pre_master_hook
+- Runs before each master is upgraded.
+- This hook runs against **each master** in serial.
+- If as task needs to run against a different host, said task will need to use [``delegate_to`` or ``local_action``](http://docs.ansible.com/ansible/playbooks_delegation.html#delegation).
+
+
+### openshift_upgrade_post_master_hook
+- Runs after each master is upgraded.
+- This hook runs against **each master** in serial.
+- If as task needs to run against a different host, said task will need to use [``delegate_to`` or ``local_action``](http://docs.ansible.com/ansible/playbooks_delegation.html#delegation).
```
dgoodwin: For the post it's worth noting that it will run before we initiate a service or full system restart (which depends on your config). Ops needed this to do a full yum update, normal users presumably might have config etc they want to get updated, so before restarts seemed like the best place.

HOOKS.md
```diff
@@ -0,0 +1,64 @@
...
+  pause:
+      prompt: "Hit enter to start the master upgrade"
+```
+
+## Upgrade Hooks
+
+### openshift_upgrade_pre_master_hook
+- Runs before each master is upgraded.
+- This hook runs against **each master** in serial.
+- If as task needs to run against a different host, said task will need to use [``delegate_to`` or ``local_action``](http://docs.ansible.com/ansible/playbooks_delegation.html#delegation).
```
ingvagabund: s/as task/a task/

HOOKS.md
ingvagabund: s/as task/a task/

sdodson:
This looks really good, I appreciate the distilling my crazy ideas into lucid technical details.

ashcrow:
Will have updates based on @dgoodwin's review shortly.

ashcrow:
:arrow_up:

ingvagabund:
Is there a flag or some command to skip tests? Cause this is "just" a documentation PR.

ashcrow:
@ingvagabund not that I'm aware of. However, when the tests are not required I've seen people merge without letting them run.

HOOKS.md
```diff
@@ -0,0 +1,64 @@
...
+## Upgrade Hooks
+
+### openshift_upgrade_pre_master_hook
+- Runs before each master is upgraded.
+- This hook runs against **each master** in serial.
+- If a task needs to run against a different host, said task will need to use [``delegate_to`` or ``local_action``](http://docs.ansible.com/ansible/playbooks_delegation.html#delegation).
+
+
+### openshift_upgrade_post_master_hook
+- Runs after each master is upgraded and **before** service/system restart.
```
detiber: @dgoodwin (more of a comment on #3101) If this is run after the master is upgraded, but before service/system restart do we need to add an additional hook for running after system/service restart? I wouldn't think that we would want to re-add the host to the load balancer before we do the system/service restart.

detiber:
@ingvagabund @ashcrow: I'll be happy to mash the merge button on this PR after #3101 has been merged without the tests passing.

ashcrow:
@tbielawa added.

ashcrow:
@dgoodwin please see the above changes.

ashcrow:
Updated per https://github.com/openshift/openshift-ansible/pull/3101#issuecomment-274167506

detiber:
Looks like this PR will need to be updated to add the `openshift_master_upgrade_hook` specified in #3101, but there is still some discussion around naming.

ashcrow:
@dgoodwin can you verify that the variable names I'm using here are correct based off the latest changes?

ashcrow:
Updating for latest changes.

ashcrow:
Updated :arrow_up:

ashcrow:
@dgoodwin / @detiber  Since the parent PR has been merged can this one as well?

dgoodwin:
@sdodson @ashcrow @detiber on a call yesterday we discussed taking this to official documentation repo, is everyone ok with this? 

CC @adellape

detiber:
I think before we document this officially we need to have version support for the hooks

sdodson:
> I think before we document this officially we need to have version support for the hooks

What do you mean by that? People may wish to have different hooks based on either the pre or post upgrade versions?

dgoodwin:
I'll try to paraphrase but @detiber proposed a versioning/API system for the hooks themselves to prevent situations where users might have outdated hooks. Right now we offer no guarantees as to what variables we make available to the hooks, we have no error handling, and we read no data out of the hook. Jason proposed a versioned API for the hooks where we could detect and warn when someone was about to run an upgrade against a hook that likely would no longer work.

sdodson:
I think we need to document the hooks as is, explain that it's best to keep them simple and understand that whatever they do in the hooks can affect the outcome of an upgrade so it's best to keep it simple and only rely on the fact that we're passing you the host to work on.

detiber:
@sdodson my main concern is if we need to change the api for the hooks, at some point we are going to hit the point where there will need to at least be data passed in for the hooks to use, not to mention that there may be data that needs to be ferried between hooks in a stable manner.

My biggest concern is that we introduce some of these changes, and then a user tries to do the next upgrade using the same hooks as the previous upgrade and things end up blowing up.

I'm not suggesting that we implement the full API, but rather just enforce that the users somehow specify a "hooks version" in their hooks, so that when we do need to introduce a change, we can fail until the user updates their hook to the new version.

I'm not sure how we'd do it yet, but ideally, we'd be able to just parse the hooks file for this value that way we can do the check at the beginning of the upgrade prior to actually calling the hook.

dgoodwin:
We could treat the absence of a declared version in a hook as the indicator we need to ask the user to go update their hooks.

detiber:
@dgoodwin I'm not a fan of that approach, but i will not block the change if you and @sdodson agree to go that way.

ashcrow:
Why could call it a preview or a v0 and explicitly let the user know in official documentation that the API **will** change.

detiber:
> Why could call it a preview or a v0 and explicitly let the user know in official documentation that the API will change.

I agree that is better than just documenting the feature. I do still have concerns on relying for documentation for these things, though. 

Even if we tell users something is preview and that the interface is not guaranteed to be stable, it we are talking about a feature that is invoked as part of a lengthy upgrade process, we can't expect that a user is going to be actively monitoring and interacting with the terminal to resolve issues that may be encountered or that unexpected breakage is even more painful for support/customers when it happens during the middle of a limited maintenance window during an upgrade.

PR #3126, opened 2017-01-19
Cleaning repo cache earlier when upgrading

mtnbikenc:
When performing an upgrade, facts are set for the currently available version of OpenShift through repoquery.  However, if the repos have been changed manually, for instance to switch from 3.3 to 3.4 repos, and the repo cache was not manually cleaned, the fact would be set based on the repo cache.  A step was introduced previously to ensure all repo cache was clean, however the task was being run after the openshift_version fact was set.

This PR moves the task to clean repo cache early in execution to ensure the openshift_version fact is set according to what is actually available.  This PR was tested on a 3.3 to 3.4 upgrade.

Original Error:
```
TASK [openshift_version : fail] ************************************************
fatal: [ec2-54-208-186-136.compute-1.amazonaws.com]: FAILED! => {"changed": false, "failed": true, "msg": "Detected OpenShift version 3.3.1.9 does not match requested openshift_release 3.4. You may need to adjust your yum repositories, inventory, or run the appropriate OpenShift upgrade playbook."}
```
cc: @dgoodwin

### DIFF
--- a/playbooks/byo/openshift-cluster/upgrades/v3_4/upgrade.yml
+++ b/playbooks/byo/openshift-cluster/upgrades/v3_4/upgrade.yml
@@ -6,8 +6,8 @@
   tags:
   - pre_upgrade
 
-# Configure the upgrade target for the common upgrade tasks:
-- hosts: l_oo_all_hosts
+- name: Configure the upgrade target for the common upgrade tasks
+  hosts: l_oo_all_hosts
   tags:
   - pre_upgrade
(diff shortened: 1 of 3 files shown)


dgoodwin:
Looks like it will run for nodes as well, +1 from me.

mtnbikenc:
aos-ci-test

mtnbikenc:
@detiber PTAL

PR #3129, opened 2017-01-19
BZ 1414477 missing import jks

jcantrill:
fixes BZ141447 by:

* Running JKS container directly on the master via 'docker run'

cc @sdodson @ewolinetz


roles/openshift_metrics/tasks/import_jks_certs.yaml
```diff
@@ -87,32 +34,23 @@
...
-    -o name
-  register: podoutput
-  check_mode: no
-  when: not metrics_keystore.stat.exists or
-        not metrics_truststore.stat.exists or
-        not cassandra_keystore.stat.exists or
-        not cassandra_truststore.stat.exists or
-        not jgroups_keystore.stat.exists
-
-- command: >
```
ewolinetz: I we can't use `oc` here, but shouldn't we wait until the docker container terminates still?

jcantrill: I removed it because it did not appear to an asynchronus call

jcantrill:
closing in favor of another approach

PR #3132, opened 2017-01-19
g_master_mktemp in openshift-master conflicts with openshift_master_certificates

richm:
This prevents running ansible -c local to install against the local
host.  The problem is that both
roles/openshift_master_certificates/tasks/main.yml
and
playbooks/common/openshift-master/config.yml
register the variable g_master_mktemp.  I'm not sure how it happens, but
in some cases openshift_master_certificates will overwrite the value
defined in openshift-master, which will cause the task
`name: Delete temporary directory on localhost` to fail.
```
MSG:

the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'stdout'

The error appears to have been in '/usr/share/ansible/openshift-ansible/playbooks/common/openshift-master/config.yml': line 198, column 5, but may
```

### DIFF
--- a/roles/openshift_master_certificates/tasks/main.yml
+++ b/roles/openshift_master_certificates/tasks/main.yml
@@ -105,7 +105,7 @@
 
 - name: Create local temp directory for syncing certs
   local_action: command mktemp -d /tmp/openshift-ansible-XXXXXXX
-  register: g_master_mktemp
+  register: g_master_certs_mktemp
   changed_when: False
   when: master_certs_missing | bool
   delegate_to: localhost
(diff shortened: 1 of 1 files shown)


abutcher:
aos-ci-test

richm:
@abutcher How would I go about getting this patch into the release-1.4 branch and the ocp-3.4 branch?  I would like to get this patch into the next updates of 1.4/3.4.  Do I just submit new PRs against those branches?

abutcher:
@richm Yeah, create branch based on `release-1.4` and pick this commit for the PR

sdodson:
The ocp-3.4 branch was only maintained to carry the docker-1.12 requirement until docker-1.12 was available in RHEL7. We're building OCP releases from release-1.4 again.

PR #3133, opened 2017-01-19
Template update for Hawkular Metrics 0.23

mwringe:
For the Hawkular Metrics 0.23 release we need a few system properties added to its command.


ingvagabund:
aos-ci-test

ingvagabund:
@mwringe  What is the version currently containerized in https://hub.docker.com/r/openshift/origin-metrics-hawkular-metrics? Latest does not say much about it.

Are there any tests that actually test the Hawkular Metrics is functioning properly?

mwringe:
@jcantrill how do people check that metrics is working properly? If we don't have proper CI tests in the place which can automatically run this, its going to be a problem.

@ingvagabund the version of our images does not correspond to the version of the component running, but the version of OpenShift to be run on.

We are going to have a lot of problems if we cannot get our metrics PRs merged in a timely fashion with CI tests in place. I think it might be have been premature to have metric as part of the ansible install without these tests in place.

There is also problems where we are potentially breaking latest if our images are expecting certain templates to be used, but the latest ansible does not include those updates yet. Not being able to run a single CI to automate that updates to our docker images and updates to ansible all work properly is also not ideal.

We already have these changes done in the origin metrics project, which have passed out CI there (which uses the older deployer and does have CI tests in place): https://github.com/openshift/origin-metrics/commit/4f1e670095989688ce08c2d97481cebd5a1ca43c & https://github.com/openshift/origin-metrics/commit/e5d98d17d3cc7c1c42026e6153732e2d681febe7

sdodson:
@jcantrill @ewolinetz What's your opinion on whether or not we should include logging and metrics in our default configuration when running CI jobs? I feel much more confident in including it in CI jobs now that you guys have rewritten it and are more involved in ownership.

jcantrill:
@mwringe Could we not still use the deployers 'validate' mode?  It does the same thing. The writing is on the wall and Ansible is the path forward; the deployer will cease to exist once we can tie up the loose ends

sdodson:
And, are there any e2e tests that we can run afterwards to verify that they've been deployed correctly?

ingvagabund:
$ docker inspect docker.io/openshift/origin-metrics-hawkular-metrics:latest | grep "HAWKULAR_METRICS_VERSION"
                 "HAWKULAR_METRICS_VERSION=0.23.2.Final",
                 "HAWKULAR_METRICS_VERSION=0.23.2.Final",

Version in the image looks good.

mwringe:
@ingvagabund with this particular PR, the version of Hawkular Metrics doesn't actually matter too much. We are just setting properties values, you can run this with an older version and it will still continue to function properly. And its the same thing if you were to run this without the property values, it will still appear to work properly in most situations, but be affected by some corner cases.

sdodson:
@jcantrill what's the final word on this?

PR #3135, opened 2017-01-20
Updating to use local_action script to generate jks certs for logging

ewolinetz:
cc @jcantrill 

Attempt to resolve on logging
based on https://github.com/openshift/openshift-ansible/pull/3129/commits/1d40a9342a49205e700fe043759e0d06a459f9b1

Need to do full test before removing [WIP]

### DIFF
--- a/roles/openshift_logging/tasks/generate_certs.yaml
+++ b/roles/openshift_logging/tasks/generate_certs.yaml
@@ -85,40 +85,8 @@
   loop_control:
     loop_var: node_name
 
-- name: Copy JKS generation script
-  copy:
-    src: generate-jks.sh
-    dest: "{{generated_certs_dir}}/generate-jks.sh"
-  check_mode: no
-
-# check if pod generated files exist -- if they all do don't run the pod
-- name: Checking for elasticsearch.jks
-  stat: path="{{generated_certs_dir}}/elasticsearch.jks"
-  register: elasticsearch_jks
-  check_mode: no
-
-- name: Checking for logging-es.jks
-  stat: path="{{generated_certs_dir}}/logging-es.jks"
-  register: logging_es_jks
-  check_mode: no
-
-- name: Checking for system.admin.jks
-  stat: path="{{generated_certs_dir}}/system.admin.jks"
-  register: system_admin_jks
-  check_mode: no
-
-- name: Checking for truststore.jks
-  stat: path="{{generated_certs_dir}}/truststore.jks"
-  register: truststore_jks
-  check_mode: no
-
-- name: Run JKS generation script
-  script: generate-jks.sh {{generate_certs_dir}} {{openshift_logging_namespace}}
-  register: script_output
-  check_mode: no
-  become: yes
-  changed_when: script_output.RC == "0"
-  when: not elasticsearch_jks.stat.exists or not logging_es_jks.stat.exists or not system_admin_jks.stat.exists or not truststore_jks.stat.exists
+- name: Creating necessary JKS certs
+  include: generate_jks.yaml
 
 # check for secret/logging-kibana-proxy
 - command: >
--- a/roles/openshift_logging/tasks/generate_certs.yaml
+++ b/roles/openshift_logging/tasks/generate_certs.yaml
@@ -112,20 +112,12 @@
   register: truststore_jks
   check_mode: no
 
-- name: create JKS generation container
-  command: >
-    docker run
-    -u 0
-    -e "PROJECT={{openshift_logging_namespace}}"
-    -e "CERT_DIR={{generated_certs_dir}}"
-    -v "{{generated_certs_dir}}:{{generated_certs_dir}}"
-    --name "jks_gen_{{'abcdefghijklmnopqrstuvwxyz0123456789'|random_word(10)}}"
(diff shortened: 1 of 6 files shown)


roles/openshift_logging/tasks/generate_certs.yaml
```diff
@@ -146,20 +112,30 @@
...
 
 - command: >
-    {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig get {{podoutput.stdout}} -o jsonpath='{.status.phase}' -n {{openshift_logging_namespace}}
+    docker wait "jks_gen_{{jks_container_uuid}}"
   register: result
-  until: result.stdout.find("Succeeded") != -1
-  retries: 5
-  delay: 10
-  changed_when: no
+  changed_when: result.rc == "0"
```
ewolinetz: this may not be the correct check -- doesn't handle errors correctly potentially

jcantrill: The script for metrics appeared to work as expected but had a bunch of logs in stderr which I could not seem to pin down an actual error.  Additionally, the return code was zero so I don't know if there is a good way to determine success of failure.

roles/openshift_logging/tasks/generate_certs.yaml
```diff
@@ -146,20 +112,30 @@
...
+    "{{openshift_logging_image_prefix}}logging-deployer:{{openshift_logging_image_version}}"
+    "{{generated_certs_dir}}/generate-jks.sh"
+  register: container_output
+  check_mode: no
+  become: yes
   when: not elasticsearch_jks.stat.exists or not logging_es_jks.stat.exists or not system_admin_jks.stat.exists or not truststore_jks.stat.exists
 
 - command: >
-    {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig get {{podoutput.stdout}} -o jsonpath='{.status.phase}' -n {{openshift_logging_namespace}}
+    docker wait "jks_gen_{{jks_container_uuid}}"
```
jcantrill: Are you certain you need this wait step?  Best I could tell was that 'docker run' in this context was serial so the next step was executed when the docker run step completed.

jcantrill:
This PR will resolve: https://bugzilla.redhat.com/show_bug.cgi?id=1415057

roles/openshift_logging/tasks/generate_certs.yaml
ewolinetz: I thought it will run in the background otherwise, I'll confirm...

ewolinetz: it looks like it does block, i'll remove this task

sdodson:
Relying on installing java on the control host is probably safe, but as written this would currently run on oo_first_master. If we want to perform this locally we need to do something like https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_examples/tasks/main.yml#L14-L38 where we create an archive on the control host using local_action and then copy it and extract it.

sdodson:
If this is the implementation we plan to ship with we should add the appropriate RPM dependency to the spec file to ensure keytool is installed. Would that be 'java-1.8.0-headless' ?

ewolinetz:
@sdodson ack on it running remotely, i was confused on how we planned to implement it. currenly have a fix for that on the way. We would want 'java-1.8.0-headless", I'll add it as a dependency

roles/openshift_logging/tasks/generate_jks.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+- name: Run JKS generation script
+  local_action: script generate-jks.sh {{local_tmp}} {{openshift_logging_namespace}}
+  register: script_output
+  check_mode: no
+  become: yes
+  changed_when: script_output.RC == "0"
+  when: not elasticsearch_jks.stat.exists or not logging_es_jks.stat.exists or not system_admin_jks.stat.exists or not truststore_jks.stat.exists
+
+- name: Pushing locally generated JKS certs to remote host...
```
jcantrill: Can you replace this and the following with a loop?

ewolinetz: would you mind if we just take that on as later tech debt when cleaning up and increasing idempotency?

jcantrill:
WFM

roles/openshift_logging/tasks/generate_jks.yaml
ewolinetz: Yeah, I'm having trouble getting the when clause to work out correctly... I'd like to get this merged so that we can unblock QE.

ewolinetz:
@sdodson do we want to use tar to push certs back to the remote node? I thought there was a situation where tar may not be available (such as running Ansible within a container?)

sdodson:
We've fixed that by adding tar as a dependency for our playbooks. Whatever method works for you, we use tar for the examples because theres probably 100+ files and copy/sync module doesn't perform well with lots of files but should be fine for a handful of files.

ewolinetz:
@sdodson there's at most 4 files to push or pull at any step, I'll forego using tar for now

PR #3144, opened 2017-01-20
Skip firewalld commands when is_containerized

leifmadsen:
When the system is setup for is_containerized to a value of true, then the deployment will fail due to missing firewalld dependency.

### DIFF
--- a/roles/os_firewall/tasks/firewall/firewalld.yml
+++ b/roles/os_firewall/tasks/firewall/firewalld.yml
@@ -23,6 +23,7 @@
     masked: no
     daemon_reload: yes
   register: result
+  when: not openshift.common.is_containerized | bool
 
 - name: need to pause here, otherwise the firewalld service starting can sometimes cause ssh to fail
   pause: seconds=10
@@ -35,6 +36,7 @@
     immediate: true
     state: enabled
   with_items: "{{ os_firewall_allow }}"
+  when: not openshift.common.is_containerized | bool
 
 - name: Remove firewalld allow rules
   firewalld:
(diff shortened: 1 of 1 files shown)


sdodson:
On Atomic Host or on RHEL/Centos? We should install firewalld on RHEL/Centos but skip it on atomic host so you probably want openshift.common.is_atomic instead?

detiber:
I'm not sure this is the right way to fix this bug...  Are we not properly defaulting os_firewall_use_firewalld to false for atomic host?

leifmadsen:
This is on a CentOS host. I'm mostly following the efforts done in commit afe0064e01d3f6f33f323ffea7aae0eaf00b9ac2

I think you might be right that is_atomic is probably the more appropriate method. I have the `containerized: true` value set to true in my inventory file which causes the firewalld package to be skipped, which then results in a failure on the commands following that which want to use firewalld to manage the firewall.

detiber:
@mtnbikenc ptal, I would expect that someone running on atomic would have os_firewalld_use_firewalld defaulted to false and wouldn't hit this codepath. Maybe we need to have it explicitly fail in that case?

mtnbikenc:
@detiber I've confirmed this on RHEL with `containerized=true`.  I think the better place to fix this is in the defaults for the role and set to false for both `is_atomic` and `is_containerized`.  I'll submit a patch.

FYI: https://bugzilla.redhat.com/show_bug.cgi?id=1413447

detiber:
@mtnbikenc I would think for non-atomic, but containerized we would still want to use firewalld by default.  Is there some other edge case involved here?

mtnbikenc:
https://github.com/openshift/openshift-ansible/pull/3164

mtnbikenc:
Covered by #3164.

PR #3147, opened 2017-01-23
Fix containerized haproxy config

andrewklau:
Follow up for #3106 

- Fixes haproxy config when using the router image
- Properly passes the config file as a parameter

### DIFF
--- a/roles/openshift_loadbalancer/tasks/main.yml
+++ b/roles/openshift_loadbalancer/tasks/main.yml
@@ -17,7 +17,7 @@
 - name: Create the systemd unit files
   template:
     src: "haproxy.docker.service.j2"
-    dest: "{{ containerized_svc_dir }}/haproxy.service"
+    dest: "/etc/systemd/system/haproxy.service"
   when: openshift.common.is_containerized | bool
   notify: restart haproxy
 
--- a/roles/openshift_loadbalancer/templates/haproxy.cfg.j2
+++ b/roles/openshift_loadbalancer/templates/haproxy.cfg.j2
@@ -1,16 +1,20 @@
 # Global settings
 #---------------------------------------------------------------------
 global
+    maxconn     {{ openshift_loadbalancer_global_maxconn | default(20000) }}
+    log         /dev/log local0 info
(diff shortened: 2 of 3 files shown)


andrewklau:
There may be a better way to handle logs, I'm not sure as HaProxy doesn't seem to have a stdout option

detiber:
aos-ci-test

DanyC97:
@andrewklau sorry to hijack this PR, do you mind if i ask you few question about the OpenShift router?
* so it seems the Router is a haproxy pod (single contianer), any idea how this will scale?
* am i right in assuming 1 router pod is going to be mapped to 1 OS node (which is part of lb group)?

thank you and sorry again fro hijack

sdodson:
> @andrewklau sorry to hijack this PR, do you mind if i ask you few question about the OpenShift router?
> 
>     so it seems the Router is a haproxy pod (single contianer), any idea how this will scale?
>     am i right in assuming 1 router pod is going to be mapped to 1 OS node (which is part of lb group)?

An openshift cluster can have multiple HA routers used to service incoming requests for applications hosted on the platform. For HA routers you'd want to put an ELB or other virtual IP in front of the two routers to ensure they're highly available. In order to scale your routing platform you may shard your routers so that different routers service different portions of the cluster. See these docs 
https://docs.openshift.org/latest/install_config/router/default_haproxy_router.html#using-router-shards
https://docs.openshift.org/latest/admin_guide/high_availability.html#admin-guide-high-availability

Though, this PR doesn't deal with routers. This PR re-purposes the existing haproxy router image to provide a load balancer in front of multiple master API servers for containerized installs. The lb group is there mainly for proof of concept environments and we don't expect it to be used for real production environments as without additional work it becomes a single point of failure. We expect people to use ELBs or other ip load balancers to provide true HA.

DanyC97:
Oh i see, much thanks for the detailed info @sdodson !!
(for an outsider is hard to understand the rational for various PRs hence my confusion).

Have a great day

andrewklau:
Fixed config. Tested and implemented in a test cluster.

PR #3150, opened 2017-01-23
Conditionally include metrics start/stop

jcantrill:
fix to BZ-1415447 stop heaster.  This PR:

* Modify main to conditionally include start/stop task based on install/uninstall of metrics.

Not really certain how the bug was generated other then I can guess ansible somehow did not properly evaluate the variable to a boolean.

cc @bbguimaraes @sdodson @detiber


jcantrill:
aos-ci-tests

@sdodson rebased..

sdodson:
I guess this is fine, if we add the role to default install playbooks we'll want to reconsider uninstalling whenever the variable isn't set as that would uninstall metrics in an environment where they had manually installed metrics?

sdodson:
aos-ci-test

PR #3152, opened 2017-01-23
Bug 1415800 - Installer fails to add/check iptables rule due to lock on xtables.

abutcher:
Add `-w` to `os_firewall` iptables invocations to wait for the xtables lock.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1415800


sdodson:
aos-ci-test

PR #3154, opened 2017-01-23
WIP - Use hostvars.localhost.g_master_mktemp

sdodson:
Fixes an error when users have specified a host named localhost.

```
PLAY [Delete temporary directory on localhost] *********************************
 
TASK [file] ********************************************************************
    "failed": true
}
 
MSG:
 
the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'stdout'
 
The error appears to have been in '/usr/share/ansible/openshift-ansible/playbooks/common/openshift-master/config.yml': line 195, column 5, but may
 
  tasks:
  - file: name={{ g_master_mktemp.stdout }} state=absent
    ^ here
start a value. For instance:
 
    with_items:
      - {{ foo }}
 
Should be written as:
 
    with_items:
      - "{{ foo }}"
 
 
        to retry, use: --limit @/usr/share/ansible/openshift-ansible/playbooks/byo/openshift-cluster/config.retry
 
localhost                  : ok=301  changed=78   unreachable=0    failed=1
```

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -151,5 +151,7 @@
   become: no
   gather_facts: no
   tasks:
-  - file: name={{ g_master_mktemp.stdout }} state=absent
+  - file:
+      path: "{{ hostvars.localhost.g_master_mktemp.stdout }}"
+      state: absent
     changed_when: False


sdodson:
aos-ci-test

sdodson:
FYI @abutcher @jtligon

abutcher:
Closing in favor of #3155

PR #3159, opened 2017-01-24
Added oc_secret to lib_openshift.

twiest:


twiest:
aos-ci-test

roles/lib_openshift/src/lib/secret.py
```diff
@@ -0,0 +1,99 @@
...
+        if self._secrets is None:
+            self._secrets = self.get_secrets()
+        return self._secrets
+
+    def get_secrets(self):
+        ''' return cert '''
+        return self.get(Secret.secret_path) or {}
+
+    def add_secret(self, key, value):
+        ''' return cert '''
```
tbielawa: This docstring is not accurate.

roles/lib_openshift/src/doc/secret
```diff
@@ -0,0 +1,115 @@
...
+short_description: Module to manage openshift secrets
+description:
+  - Manage openshift secrets programmatically.
+options:
+  state:
+    description:
+    - Currently present is only supported state.
+    required: true
+    default: present
+    choices: ["present", "absent", "list"]
```
tbielawa: If `present` is only supported state, why is there a list of choices available?

roles/lib_openshift/src/ansible/oc_secret.py
```diff
@@ -0,0 +1,37 @@
...
+
+def main():
+    '''
+    ansible oc module for managing OpenShift Secrets
+    '''
+
+    module = AnsibleModule(
+        argument_spec=dict(
+            kubeconfig=dict(default='/etc/origin/master/admin.kubeconfig', type='str'),
+            state=dict(default='present', type='str',
```
tbielawa: The `DOCUMENTATION` for the `state` key says

> required: true
> default: present

but this parameter is not defined with `required=True`. As stated in the other review comment, if only `present` is supported, why provide a list of `choices`?

roles/lib_openshift/src/doc/secret
```diff
@@ -0,0 +1,115 @@
...
+    - base64 decode the object
+    required: false
+    default: false
+    aliases: []
+author:
+- "Kenny Woodson <kwoodson@redhat.com>"
+extends_documentation_fragment: []
+'''
+
+EXAMPLES = '''
```
tbielawa: More comments on the `state` parameter. Is is defined as `required: true` in `DOCUMENTATION` but only one of the `EXAMPLES` below shows it being set (*get ca from hawkular*) and in that example it is set to `list`. This conflicts with the documentation stating that "Currently `present` is only supported state."

tbielawa:
Looks like there's a little lint in this PR on python3 some how.

```
roles/lib_openshift/library/oc_secret.py:1369: [W1505(deprecated-method), OCSecret.get] Using deprecated method decodestring()
```

roles/lib_openshift/src/doc/secret
twiest: Thanks Tim, you're right, the docs were off. It supports more states than just `present` now. I've updated the docs to reflect this. Please re-review when you get a minute.

roles/lib_openshift/src/ansible/oc_secret.py
twiest: Same, the docs were wrong, state isn't `required`. I've updated the doc.

twiest:
@tbielawa I fixed the deprecation

roles/lib_openshift/src/ansible/oc_secret.py
```diff
@@ -0,0 +1,37 @@
...
+                       choices=['present', 'absent', 'list']),
+            debug=dict(default=False, type='bool'),
+            namespace=dict(default='default', type='str'),
```
tbielawa: This `AnsibleModule` spec says the default for the `namespace` param is the **word** `default` and is of **TYPE** `string`, however the `DOCUMENTATION` says the default value for `namespace` is the **word** `str`:

```
  namespace:
    description:
    - The namespace where the object lives.
    required: false
    default: str
    aliases: []
```

I think the `DOCUMENTATION` just needs to be updated to say `default: default`.

tbielawa: The fragment banners PR was merged earlier today, you may need to pull and rebase on top of that in the update.

twiest:
@tbielawa I fixed the doc as you suggested, rebased and re-generated the module.

PTAL

PR #3164, opened 2017-01-24
Remove is_containerized check for firewalld installs

mtnbikenc:
Need to install firewalld on containerized deployments.

### DIFF
--- a/roles/os_firewall/tasks/firewall/firewalld.yml
+++ b/roles/os_firewall/tasks/firewall/firewalld.yml
@@ -1,7 +1,8 @@
 ---
 - name: Install firewalld packages
-  package: name=firewalld state=present
-  when: not openshift.common.is_containerized | bool
+  package:
+    name: firewalld
+    state: present
 
 - name: Ensure iptables services are not enabled
   systemd:


mtnbikenc:
@detiber Are you saying that we _should_ be installing firewalld when doing a containerized host?  If so, I'll pull out the checks which skip the installs so it installs firewalld correctly and set the defaults back to just false on is_atomic.

mtnbikenc:
Will rebase once lint PR is merged. https://github.com/openshift/openshift-ansible/pull/3166

mtnbikenc:
aos-ci-test

roles/os_firewall/tasks/firewall/firewalld.yml
```diff
@@ -1,7 +1,8 @@
 ---
 - name: Install firewalld packages
-  package: name=firewalld state=present
-  when: not openshift.common.is_containerized | bool
+  package:
+    name: firewalld
+    state: present
```
detiber: wondering if this should have a conditional for `not openshift.common.is_atomic | bool`. otherwise, lgtm

mtnbikenc: We would never hit this code path if `is_atomic` because we already don't use firewalld if `is_atomic`.

leifmadsen: +1, confirmed working on my CentOS deploy now.

PR #3172, opened 2017-01-25
override nodename for gce with cloudprovider

detiber:
Addresses an issue with the default nodename not matching the instance name when using the gce cloudprovider integration.


detiber:
@sabre1041

detiber:
aos-ci-test

smarterclayton:
This doesn't appear to work at all - ".".split(host)[0] gives me ".", not the first chunk.  Also, `facts['provider']['metadata']['hostname']` doesn't return anything on GCE - facts['provider']['metadata']['instance']['hostname']`

smarterclayton:
Opened a PR that contains a fix

PR #3178, opened 2017-01-25
Remove master_count restriction.

### LINKED ISSUE #3175: openshift_facts received a lower value for openshift.master.master_count
I'm testing a cluster environment in which I've scaled it's masters down (from 3 to 2 in native ha), I've updated the masterCount in master-config.yaml but not able to track down why the facts `openshift_facts received a lower value for openshift.master.master_count` error me out

abutcher:
I don't think it makes sense to have this restriction unless there's a way around it.

Fixes #3175.


abutcher:
aos-ci-test

PR #3179, opened 2017-01-25
openshift_logging link pull secret to serviceaccounts

jcantrill:
This PR:

* Allows a user to provide a pull-secret to link to the logging service accounts
* fixes an error when undeploying logging

cc @ewolinetz @sdodson 

aos-ci-test

### DIFF
--- a/roles/openshift_logging/tasks/install_logging.yaml
+++ b/roles/openshift_logging/tasks/install_logging.yaml
@@ -57,6 +57,28 @@
     loop_var: file
   when: ansible_check_mode
 
+  # TODO replace task with oc_secret module that supports
+  # linking when available
+- name: Link Pull Secrets With Service Accounts
+  include: oc_secret.yaml
+  vars:
+    kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
+    subcommand: link
+    service_account: "{{sa_account}}"
+    secret_name: "{{openshift_logging_image_pull_secret}}"
+    add_args: "--for=pull"
+  with_items:
+    - default
+    - aggregated-logging-elasticsearch
+    - aggregated-logging-kibana
+    - aggregated-logging-fluentd
+    - aggregated-logging-curator
+  register: link_pull_secret
+  loop_control:
+    loop_var: sa_account
+  when: openshift_logging_image_pull_secret is defined
+  failed_when: link_pull_secret.rc != 0
+
 - name: Scaling up cluster
   include: start_cluster.yaml
   when: start_cluster | default(true) | bool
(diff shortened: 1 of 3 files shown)


abutcher:
aos-ci-test

roles/openshift_logging/tasks/oc_secret.yaml
```diff
@@ -0,0 +1,7 @@
+---
+- command: >
```
ewolinetz: @kwoodson do we have a timeline on when the oc_secret module will be ported? I didn't have access to the trello card where this work is being tracked...

ewolinetz:
Other than question regarding module usage over new task, LGTM

kwoodson:
@ewolinetz,

oc_secret should be able to be finished today.  Here is the PR https://github.com/openshift/openshift-ansible/pull/3159

roles/openshift_logging/tasks/oc_secret.yaml
ewolinetz: @jcantrill  looks like `oc_secret` was merged. can we try to use that module instead of this task?

jcantrill:
@ewolinetz Looking at the referenced PR, I don't see where it provides options to 'link' a serviceaccount to a secret

kwoodson:
@jcantrill,

We have another module named oc_secret_add that does the linking of secrets to serviceaccounts. It should be coming very soon.

jcantrill:
@ewolinetz any opinions on merging this given we dont have an alternative right now?  Will ops need to specify pull secrets before we can utilize the module?

roles/openshift_logging/tasks/install_logging.yaml
```diff
@@ -51,6 +51,26 @@
     loop_var: file
   when: ansible_check_mode
 
+- name: Link Pull Secrets With Service Accounts
```
ewolinetz: It's pretty inconvenient that we aren't creating service accounts with a module meaning we need to add a pull secret out here...  can you add a TODO comment for this just as a reminder so we don't lose track of that? I'd like to make sure that when we have a module to create a service account idempotently that we would create a pull secret shortly after if applicable.

ewolinetz:
nit and then LGTM I suppose. We'll need to refactor this with modules when they become available

roles/openshift_logging/tasks/install_logging.yaml
jcantrill: added TODO

jcantrill:
@ewolinetz updated and rebased

jcantrill:
aos-ci-test

PR #3182, opened 2017-01-25
Comma separate no_proxy host list in openshift_facts

### LINKED ISSUE #3180: origin-master service wont start with proxy configured
#### Description

On a clean server, the default BYO is not working with proxy configured. 

##### Version

ansible 2.2.0.0
openshift-ansible-3.5.2-1-2-gc5a1e1e

##### Steps To Reproduce

1. Use the following config https://gist.github.com/aerialls/f45abab663233d5f2a63eda37d6f8789
2. Launch ansible with `ansible-playbook playbooks/byo/config.yml`

##### Expected Results

The service origin-master starts successfully.

##### Observed Results

The origin-master is not starting.

```
TASK [openshift_master : Start and enable master] ******************************
FAILED - RETRYING: TASK: openshift_master : Start and enable master (1 retries left).
fatal: [os01.foobar.com]: FAILED! => {"attempts": 1, "changed": false, "failed": true, "msg": "Unable to start service origin-master: Job for origin-master.service failed because the control process exited with error code. See \"systemctl status origin-master.service\" and \"journalctl -xe\" for details.\n"}
```

In the server logs, I can see the following error

```
Jan 25 16:34:13 os01 origin-master: F0125 16:34:13.845990    3399 plugins.go:140] Couldn't init admission plugin "BuildDefaults": error reading config: [pos 197]: json: expect char '"' but got char '['
```

If the lines `openshift_http_proxy` and `openshift_https_proxy` are commented, the service starts. 

##### Additional Information

I'm using `CentOS Linux release 7.3.1611 (Core)`

abutcher:
Proposed change comma separates the `no_proxy` host list upstream in `openshift_facts` so that it appears as a string everywhere it is used. Note that `set_proxy_facts()` converts the `no_proxy` host list to a list and then back to a string.

Fixes #3180.

@sdodson @bparees PTAL

### DIFF
--- a/roles/docker/tasks/main.yml
+++ b/roles/docker/tasks/main.yml
@@ -85,7 +85,7 @@
   - reg_conf_var: HTTPS_PROXY
     reg_fact_val: "{{ docker_https_proxy | default('') }}"
   - reg_conf_var: NO_PROXY
-    reg_fact_val: "{{ docker_no_proxy | default('') | join(',') }}"
+    reg_fact_val: "{{ docker_no_proxy | default('') }}"
   notify:
   - restart docker
   when:
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1641,7 +1641,7 @@ def set_proxy_facts(facts):
             # We always add local dns domain and ourselves no matter what
             common['no_proxy'].append('.' + common['dns_domain'])
             common['no_proxy'].append(common['hostname'])
-            common['no_proxy'] = sort_unique(common['no_proxy'])
+            common['no_proxy'] = ','.join(sort_unique(common['no_proxy']))
         facts['common'] = common
     return facts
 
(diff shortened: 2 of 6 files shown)


bparees:
ugh, my bad.  it looks like a plausible fix to me, but i'm definitely going to defer to @sdodson since my assumptions broke this in the first place.

roles/docker/tasks/main.yml
```diff
@@ -85,7 +85,7 @@
   - reg_conf_var: HTTPS_PROXY
     reg_fact_val: "{{ docker_https_proxy | default('') }}"
   - reg_conf_var: NO_PROXY
-    reg_fact_val: "{{ docker_no_proxy | default('') | join(',') }}"
+    reg_fact_val: "{{ docker_no_proxy | default('') }}"
```
sdodson: I think we need to do the same in all of the templates that include openshift.common.no_proxy too if we're storing it as a string.

sdodson: search for openshift.common.no_proxy

abutcher: Gottem :+1:

abutcher:
aos-ci-test

aerialls:
I can confirm that this PR fixes #3180 👍

sdodson:
@abutcher can you check for relevance to 3.4 and 3.3? we've had s,o,m,e p,e,o,p,l,e report it happens there too

PR #3183, opened 2017-01-25
Adding oc_scale to lib_openshift.

kwoodson:


roles/lib_openshift/src/doc/scale
```diff
@@ -0,0 +1,70 @@
...
+    description:
+    - Name of the object that is being queried.
+    required: false
+    default: None
+    aliases: []
+  namespace:
+    description:
+    - The namespace where the object lives.
+    required: false
+    default: str
```
tbielawa: `namespace` is mis-documented here. The default is `default` the type is `str`.

kwoodson:
@tbielawa, Updating doc string.

PR #3184, opened 2017-01-25
Standardize add_host: with name and changed_when

mtnbikenc:
The add_host: task does not change any data on the host and as practice
has been configured to changed_when: False.  This commit standardizes
that usage in the byo and common playbooks.  Additionally, task names
are added to each task to improve troubleshooting.


mtnbikenc:
aos-ci-test

PR #3187, opened 2017-01-25
rpm_upgrade.yml: Restart the service that was upgraded

zeroem:
Instead of only restarting the `...-node` service

While running through a cluster upgrade, from `3.1` to `3.2`, this section is failing on every attempted run against the master hosts as the `origin-node` service is not installed on them.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
@@ -7,4 +7,4 @@
   when: "{{ not openshift.common.is_atomic | bool }}"
 
 - name: Restart node service
-  service: name="{{ openshift.common.service_type }}-node" state=restarted
+  service: name="{{ openshift.common.service_type }}-{{ component }}" state=restarted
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
```diff
@@ -7,4 +7,4 @@
   when: "{{ not openshift.common.is_atomic | bool }}"
 
 - name: Restart node service
```
detiber: probably should rename this task to 'Restart service' with this change.

playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
```diff
@@ -7,4 +7,4 @@
   when: "{{ not openshift.common.is_atomic | bool }}"
```
detiber: @dgoodwin @tbielawa @ewolinetz Since this seems to be re-used now, PyYAML should probably be installed outside of this task, otherwise we are trying to install it for each package through the upgrade.

dgoodwin:
I believe we hit and fixed this since 1.2, master version here for comparison: https://github.com/zeroem/openshift-ansible/blob/master/playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml

zeroem:
This change won't actually work as the `origin-master` package does not have an `origin-master` service. It has `origin-master-api` and `origin-master-controllers`

PR #3195, opened 2017-01-26
fix 1414625. Additional fix to run password commands on control node

jcantrill:
fixes where we check for tool on control, but were still running on master

@ewolinetz @sdodson

### DIFF
--- a/roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
+++ b/roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
@@ -19,24 +19,52 @@
 - slurp: src={{ openshift_metrics_certs_dir }}/hawkular-metrics-truststore.pwd
   register: hawkular_truststore_password
 
+- stat: path="{{openshift_metrics_certs_dir}}/{{item}}"
+  register: pwd_file_stat
+  with_items:
+  - hawkular-metrics.pwd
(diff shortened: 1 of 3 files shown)


jcantrill:
aos-ci-test

roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
```diff
@@ -19,25 +19,53 @@
...
+    dest: "{{openshift_metrics_certs_dir}}/{{item}}"
+  with_items:
+  - hawkular-metrics.pwd
+  - hawkular-metrics.htpasswd
+  - hawkular-jgroups-keystore.pwd
+  when: "not pwd_files['{{ item }}'].exists"
 
 - include: import_jks_certs.yaml
 
+- file: path="{{local_tmp.stdout}}" state=absent
```
ewolinetz: this should probably be a local_action since we create it as a local_action

ewolinetz:
one nit, otherwise LGTM

roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
jcantrill: fixed

jcantrill:
@sdodson  yes

jcantrill:
aos-ci-test

PR #3197, opened 2017-01-26
Adding oadm_manage_node lib_openshift.

kwoodson:


roles/lib_openshift/src/doc/manage_node
```diff
@@ -0,0 +1,88 @@
...
+    description:
+    - A list of the nodes being managed
+    required: false
+    default: None
+    aliases: []
+  selector:
+    description:
+    - The selector when filtering on node labels
+    required: false
+    default: str
```
tbielawa: According to the Ansible Module `selector` has a default of `None` and type of `str`

> selector=dict(default=None, type='str')

roles/lib_openshift/src/class/oadm_manage_node.py
```diff
@@ -0,0 +1,208 @@
...
+# pylint: disable=too-many-instance-attributes
+class ManageNode(OpenShiftCLI):
+    ''' Class to wrap the oc command line tools '''
+
+    # pylint allows 5
+    # pylint: disable=too-many-arguments
+    def __init__(self,
+                 config,
+                 verbose=False):
+        ''' Constructor for OCVolume '''
```
tbielawa: OCVolume is related in what way?

roles/lib_openshift/src/class/oadm_manage_node.py
```diff
@@ -0,0 +1,208 @@
...
+        if "Listing matched" in results['results']:
+            listing_match = re.compile('\n^Listing matched.*$\n', flags=re.MULTILINE)
+            pods = listing_match.split(results['results'])
+            for pod in pods:
+                if pod:
+                    all_pods.extend(json.loads(pod)['items'])
+
+        # 3.3 specific
+        else:
+            # this is gross but I filed a bug...
```
tbielawa: Got a link so we can check on that later?

kwoodson: I am fixing the current questions.  If drain comes along, I think the plan of action would be to deprecate the old evacuate param in favor of drain. In the code where evacuate is called we could then either check version or we create a lib_openshift that supports the targeted openshift version.

roles/lib_openshift/src/class/oadm_manage_node.py
kwoodson: Will fix.  Copy from a previous file.

roles/lib_openshift/src/doc/manage_node
kwoodson: Will fix doc.

PR #3198, opened 2017-01-26
Correct usage of draining nodes

mtnbikenc:
Verified `oadm drain NODE` applies to 1.3, 1.4 and 1.5.
Verified upgrade from 3.4 to 3.5.


mtnbikenc:
@tbielawa PR updated to remove deprecated code.

detiber:
@mtnbikenc Is drain supported in all shipped versions of 3.3 and 3.4 or only certain versions where it has been backported?

mtnbikenc:
aos-ci-test

mtnbikenc:
@detiber `oadm drain` has been around since 1.3.0. https://github.com/openshift/origin/blob/v1.3.0/docs/man/oadm/oadm-drain.1

Issue #3199, opened 2017-01-26
openshift_logging does not create ES deploymentconfigs on image change


jcantrill:
Trying to install logging using the role 'openshift_logging' does not update the image to the ES DC after installing the logging stack.  This is because of the when statement in 'tasks/install_elasticsearch.yaml' which conditionally runs when the cluster is resized but does not take into account some of the other vars may have changed.

jcantrill:
@ewolinetz I will have to test to confirm if #3202 fixed this issue.  It can be confirmed by:

1.  deploying
2. Modifying the image_version
3. deploy again.

assert new deployment image equals the desired image version.

ewolinetz:
i just verified that the image version will change, closing as resolved by #3202

PR #3202, opened 2017-01-27
Updating to reuse previous ES DC names and always generate DCs

ewolinetz:
Addresses https://github.com/openshift/openshift-ansible/issues/3199

@jcantrill

### DIFF
--- a/roles/openshift_logging/tasks/install_elasticsearch.yaml
+++ b/roles/openshift_logging/tasks/install_elasticsearch.yaml
@@ -1,24 +1,30 @@
 ---
+- name: Getting current ES deployment size
+  set_fact: openshift_logging_current_es_size={{ openshift_logging_facts.elasticsearch.deploymentconfigs.keys() | length }}
+
 - name: Generate PersistentVolumeClaims
   include: "{{ role_path}}/tasks/generate_pvcs.yaml"
   vars:
     es_pvc_names: "{{openshift_logging_facts.elasticsearch.pvcs.keys()}}"
     es_dc_names: "{{openshift_logging_facts.elasticsearch.deploymentconfigs.keys()}}"
-  when:
-    - "{{ openshift_logging_facts.elasticsearch.deploymentconfigs.keys() | length < openshift_logging_es_cluster_size }}"
 
+# we should initialize the es_dc_pool with the current keys
 - name: Init pool of DeploymentConfig names for Elasticsearch
-  set_fact: es_dc_pool={{es_dc_pool | default([]) + [deploy_name]}}
+  set_fact: es_dc_pool={{ es_dc_pool | default([]) + [deploy_name] }}
+  with_items: "{{ openshift_logging_facts.elasticsearch.deploymentconfigs.keys() }}"
+  loop_control:
+    loop_var: deploy_name
+
+# This should be used to generate new DC names if necessary
+- name: Create new DeploymentConfig names for Elasticsearch
+  set_fact: es_dc_pool={{es_dc_pool + [deploy_name]}}
   vars:
     component: es
     es_cluster_name: "{{component}}"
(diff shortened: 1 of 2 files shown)


roles/openshift_logging/tasks/install_elasticsearch.yaml
```diff
@@ -35,14 +41,15 @@
...
     es_node_selector: "{{openshift_logging_es_nodeselector | default({})}}"
   with_indexed_items:
-    - "{{es_dc_pool | default([])}}"
+    - "{{ es_dc_pool }}"
   check_mode: no
-  when:
-    - "{{ openshift_logging_facts.elasticsearch.deploymentconfigs.keys() | length < openshift_logging_es_cluster_size }}"
   changed_when: no
 
 # --------- Tasks for Operation clusters ---------
```
jcantrill: More of an aside, we should be able to refactor this into a single task and run it twice, once for non-ops, and once for ops.

roles/openshift_logging/tasks/install_elasticsearch.yaml
```diff
@@ -65,21 +72,27 @@
...
+- name: Create new DeploymentConfig names for Elasticsearch Ops
+  set_fact: es_ops_dc_pool={{es_ops_dc_pool + [deploy_name]}}
   vars:
     component: es-ops
     es_cluster_name: "{{component}}"
     deploy_name_prefix: "logging-{{component}}"
     deploy_name: "{{deploy_name_prefix}}-{{'abcdefghijklmnopqrstuvwxyz0123456789'|random_word(8)}}"
     cluster_size: "{{openshift_logging_es_ops_cluster_size}}"
-  with_sequence: count={{openshift_logging_es_ops_cluster_size - openshift_logging_facts.elasticsearch_ops.deploymentconfigs.keys() | length}}
+  with_sequence: count={{ openshift_logging_es_ops_cluster_size | int - openshift_logging_current_es_ops_size | int }}
```
jcantrill: Looks to be missing the 'abs' filter like in the similar step above

ewolinetz:
@jcantrill regarding scaling down -- I think you're correct. I think the way to resolve that would be to have a with_sequence on the actual generation of the ES_DCs... then we would handle scale up/down for updating. The other problem is the previously generated DC that isn't used anymore due to scaling down would still be out there... do we want to delete it in that case? Otherwise when we go to start cluster I think we'd grab it and set its replica count to 1....

jcantrill:
@ewolinetz We know you are less likely to scale down so this really is a low probability usecase.  Originally we had logic to only allow you to scale down in 1 node increments.  I think, however, your suggestion is correct in that delete the existing DC's assuming you are above the threshold of shard redundancy (3 members?).  This should mean there is no data loss and the cluster will rebalance.   Maybe there should be some logic like:

1. <3 node. manual scale down, ansible scale up
2. >=3 nodes. ansible scale up or down, delete dc as appropriate

ewolinetz:
Per discussion with @jcantrill we won't allow scale down through Ansible given the required steps that would be needed to prevent data loss. Will need to add documentation on how to correctly scale down a cluster.

roles/openshift_logging/tasks/install_elasticsearch.yaml
```diff
@@ -101,9 +114,8 @@
     openshift_logging_es_recover_after_time: "{{openshift_logging_es_ops_recover_after_time}}"
     es_node_selector: "{{openshift_logging_es_ops_nodeselector | default({})}}"
   with_indexed_items:
-    - "{{es_dc_pool_ops | default([])}}"
+    - "{{ es_ops_dc_pool | default([]) }}"
```
sdodson: Default to empty here because ops cluster is optional but not in non-ops cluster because it's required?

PR #3203, opened 2017-01-27
fixes #3201 by removing immutable resources before creation.

### LINKED ISSUE #3201: openshift_logging role updating route fails because change is immutable
Trying to redeploy logging stack after updating the route fails because the host is an immutable field.  There are 2 issues here:
* Need to better handle immutable fields
* Need to better handle the error condition instead of preceding like there is not a problem
```
TASK [openshift_logging : Checking generation of Route logging-kibana] *********
task path: /home/jeff.cantrill/git/openshift-ansible/roles/openshift_logging/tasks/oc_apply.yaml:2
ok: [openshift] => {"changed": false, "cmd": "/usr/local/bin/oc --config=/tmp/openshift-logging-ansible-u2GlcK/admin.kubeconfig get Route logging-kibana -o jsonpath='{.metadata.resourceVersion}' -n logging || echo 0", "delta": "0:00:00.242171", "end": "2017-01-26 22:03:12.762366", "rc": 0, "start": "2017-01-26 22:03:12.520195", "stderr": "", "stdout": "45276", "stdout_lines": ["45276"], "warnings": []}

TASK [openshift_logging : Applying /tmp/openshift-logging-ansible-u2GlcK/templates/logging-logging-kibana-route.yaml] ***
task path: /home/jeff.cantrill/git/openshift-ansible/roles/openshift_logging/tasks/oc_apply.yaml:12
ok: [openshift] => {"changed": false, "cmd": ["/usr/local/bin/oc", "--config=/tmp/openshift-logging-ansible-u2GlcK/admin.kubeconfig", "apply", "-f", "/tmp/openshift-logging-ansible-u2GlcK/templates/logging-logging-kibana-route.yaml", "-n", "logging"], "delta": "0:00:00.192219", "end": "2017-01-26 22:03:13.244504", "failed": false, "failed_when_result": false, "rc": 1, "start": "2017-01-26 22:03:13.052285", "stderr": "The Route \"logging-kibana\" is invalid: spec.host: Invalid value: \"kibana.192.168.121.194.xip.io\": field is immutable", "stdout": "", "stdout_lines": [], "warnings": []}

TASK [openshift_logging : Determine change status of Route logging-kibana] *****
task path: /home/jeff.cantrill/git/openshift-ansible/roles/openshift_logging/tasks/oc_apply.yaml:21
ok: [openshift] => {"changed": false, "cmd": "/usr/local/bin/oc --config=/tmp/openshift-logging-ansible-u2GlcK/admin.kubeconfig get Route logging-kibana -o jsonpath='{.metadata.resourceVersion}' -n logging || echo 0", "delta": "0:00:00.225473", "end": "2017-01-26 22:03:13.775026", "failed": false, "failed_when_result": false, "rc": 0, "start": "2017-01-26 22:03:13.549553", "stderr": "", "stdout": "45276", "stdout_lines": ["45276"], "warnings": []}
```

jcantrill:


roles/openshift_logging/tasks/install_logging.yaml
```diff
@@ -32,6 +32,17 @@
...
+- oc_obj:
+    state: absent
+    name: "{{item.name}}"
+    kind: "{{item.kind}}"
+    kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
+    namespace: "{{openshift_logging_namespace}}"
+  with_items:
+  - {name: logging-kibana, kind: route}
+  - {name: logging-kibana-ops, kind: route}
+  - {name: kibana-proxy, kind: oauthclient}
```
detiber: Wouldn't this be disruptive if re-running the install?

ewolinetz: Agreed. I think we should only be doing this if we're trying to update an object and the result would be different than the current state.

ewolinetz: Can we use the `oc_route` module specifically for the routes? I'm not sure if it would handle this scenario for us though... it would probably just create a new one and leave the old one dangling

jcantrill: @detiber @ewolinetz This was what I was trying to highlight in the feature; the only way to replace a route where the host has changed is to delete it and recreate it.  The hostname is a fairly easy check and just a little more work; I'll proceed doing something here.  I'm not certain to what degree we want to check the other parts change.  It may not matter if using oc apply but then that will fail if the hostname and certs change.  Do we wish to get that involved?  Is changed that important to introduce the complexity?

The oauthclient produced an error because it needed the resource version for 'apply' to work.  My concern is if the version changes between generation and apply.  Is that a risk we need to accept and just add the resourceversion to the object def?

detiber: @jcantrill I'm good with deleting and re-creating, I just want to make sure that we are maintaining idempotency and only deleting/recreating when necessary.

jcantrill:
closing as an alternate was merged by @ewolinetz in another PR

PR #3205, opened 2017-01-27
Remove v3_3 upgrade playbooks

mtnbikenc:


sdodson:
Lets wait on this to see what shakes out of the multi-version upgrade stuff. I'm sure we can reconstruct this from other branches and history but I'd prefer leaving it for now.

mtnbikenc:
@sdodson Are we going to continue to keep the old upgrade playbooks around?

detiber:
@sdodson @mtnbikenc I would suggest leaving the old, but potentially compatible playbooks in place. That way all bugfixes, etc can be done in master and changes can be backported back to the affected branch. Granted with current work, that would mean backporting the module work. But as long as nothing that is 3.5+ specific is done that should be possible.

mtnbikenc:
Closing this PR.  The work can be completed when deemed necessary in the future.

PR #3214, opened 2017-01-27
Updating openshift_logging role to add kibana public url to loggingPu…

ewolinetz:
…blicURL in master-config

To address https://bugzilla.redhat.com/show_bug.cgi?id=1414706

### DIFF
--- a/roles/openshift_hosted_logging/tasks/deploy_logging.yaml
+++ b/roles/openshift_hosted_logging/tasks/deploy_logging.yaml
@@ -165,12 +165,7 @@
   retries: 20
   delay: 15
 
-- name: Adding Kibana route information to loggingPublicURL
-  modify_yaml:
-    dest: "{{ openshift.common.config_base }}/master/master-config.yaml"
-    yaml_key: assetConfig.loggingPublicURL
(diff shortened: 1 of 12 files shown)


ewolinetz:
aos-ci-test

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -165,6 +165,13 @@
   retries: 20
   delay: 15
 
+- name: Adding Kibana route information to loggingPublicURL
+  modify_yaml:
+    dest: "{{ openshift.common.config_base }}/master/master-config.yaml"
+    yaml_key: assetConfig.loggingPublicURL
+    yaml_value: "https://{{ openshift_logging_kibana_hostname }}"
+  notify: restart master
```
sdodson: We need to do this across all masters right? We probably need to push this up into a separate play in the playbook so that all masters get updated and restarted.

ewolinetz: could we leverage the include_role for this in the playbook and just call the specific task to update the master-configs to do this?

ewolinetz:
@sdodson updated to apply config change to all masters

PR #3215, opened 2017-01-27
Standardize entry-point playbooks with named plays

mtnbikenc:
Standardized the use of `name:` for playbook entry-points to improve troubleshooting and consistency.


detiber:
aos-ci-test

PR #3217, opened 2017-01-29
Added repoquery to lib_utils.

twiest:


sdodson:
I've not reviewed this code, but we should make sure it supports satellite (--plugins for yum) and dnf similar to the logic in playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml

twiest:
@sdodson I actually thought about the dnf case, but I think we want to create a separate, dnf_repoquery specific module for that. 

`dnf repoquery` and `repoquery` have different options (at least not on fedora), so they're two separate things. 

Creating one module for both would be a "least common denominator" type module (only supporting features that are in both), which I doubt we want.

As far as `--plugins` for yum, we're calling repoquery directly so whatever it supports, we support.

You're right that we should test to confirm, I just don't have any idea how or where to test that.

twiest:
@tbielawa I just rebased from master, double checked the docs and made sure everything was generated. I think this is ready for review.

I added blank lines in between generation targets in `roles/lib_openshift/src/sources.yml` to make a clearer delineation. If you guys hate that, I'm willing to remove the blank lines.


@sdodson are you ok with us merging this without the `yum --plugins` testing, or would you prefer that testing be done first?

sdodson:
I'd prefer we ensure --plugins support before we merge this, I'd be worried we'd regress here and wouldn't realize it until we've shipped something to customers who depend on satellite.

tbielawa:
@sdodson said

> I'd prefer we ensure --plugins support before we merge this, I'd be worried we'd regress here and wouldn't realize it until we've shipped something to customers who depend on satellite.

I leave the future of this PR is scott's trusting hands.

twiest:
@sdodson ok, sounds good. How should I test this?

sdodson:
I think just ensuring the option is passed for all invocations is enough, it shouldn't negatively impact anything so all your existing tests should continue passing. I'll see if I can find who did the satellite testing in QE and have them verify too.

twiest:
@sdodson ah, ok, sorry I think I misunderstood your request.

I can definitely do that, and I'll add that to a unit test to ensure we're calling repoquery with `--plugins`.

twiest:
@sdodson Ok, I've added `--plugins` to all invocations of repoquery (it's just 1 code path). You can see that [here](https://github.com/openshift/openshift-ansible/pull/3217/files#diff-f5c092abfffc81525d2215ff42da0052R49).

You might have to hit refresh on your browser if you don't see a highlighted line (I did).


I also added to the unit test to ensure we're always calling repoquery with `--plugins`. You can see that [here](https://github.com/openshift/openshift-ansible/pull/3217/files#diff-65fc3eba676bb383873323459693b4e8R78).

Again, you might have to hit refresh.

@sdodson Is this what you were looking for?

@tbielawa if @sdodson approves of my fix for `--plugins` then I think this is ready to be reviewed.

roles/lib_openshift/src/test/unit/oc_secret.py
```diff
@@ -81,7 +81,7 @@ def test_adding_a_secret(self, mock_openshift_cmd):
 
         # Making sure our mock was called as we expected
         mock_openshift_cmd.assert_has_calls([
-            mock.call(['get', 'secrets', '-o', 'json', 'secretname'], output=True),
+            mock.call(['get', 'secrets', 'secretname', '-o', 'json'], output=True),
```
sdodson: This seems un-related to this PR but inconsequential.

sdodson: As long as it's not an unintentional regression on recent changes I'm fine with it.

twiest: Yeah, a separate PR broke this unit test. This fixes it.

PR #3220, opened 2017-01-30
add persistent versions of quickstarts

bparees:


bparees:
@sdodson @luciddreamz ptal

sdodson:
Did we want to get rid of the ephemeral examples too? I lost track of the intention of the email.

sdodson:
aos-ci-test

bparees:
> Did we want to get rid of the ephemeral examples too? I lost track of the intention of the email.

no, we need to keep ephemeral around because we want something to work out of the box when you may not have persistent volumes available.

PR #3223, opened 2017-01-30
Raise error when unable to parse docker version output

abutcher:
Raise exception when unable to parse the output of 'docker version' in openshift_facts.

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1278,12 +1278,22 @@ def get_docker_version_info():
     """ Parses and returns the docker version info """
     result = None
     if is_service_running('docker'):
-        version_info = yaml.safe_load(get_version_output('/usr/bin/docker', 'version'))
-        if 'Server' in version_info:
-            result = {
-                'api_version': version_info['Server']['API version'],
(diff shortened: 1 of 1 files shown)


roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1278,12 +1278,20 @@ def get_docker_version_info():
...
+                result = {
+                    'api_version': version_info['Server']['API version'],
+                    'version': version_info['Server']['Version']
+                }
+        # PyYAML unable to parse the output of `docker version`
+        except yaml.scanner.ScannerError:
+            raise OpenShiftFactsInternalError('Unable to determine the installed docker version.'
+                                              ' Check the output of `docker version` and ensure that the'
+                                              ' correct version of docker has been installed per the'
+                                              ' host preparation steps in the OpenShift documentation'
```
tbielawa: Can you include the original `yaml.scanner.ScannerError.msg` part in the error message? Save the people some time so they don't have to run it themselves?

tbielawa: Alternatively/in-concert, can you include the actually read output from the command that errored (include the result from the docker version call)

PR #3224, opened 2017-01-30
Adding oc_service to lib_openshift.

kwoodson:


ewolinetz:
LGTM, wouldn't mind a second set of eyes on it since I haven't reviewed our other module PRs
@tbielawa

roles/lib_openshift/src/class/oc_service.py
```diff
@@ -0,0 +1,170 @@
...
+    @property
+    def service(self):
+        ''' property function service'''
+        if not self.svc:
+            self.get()
+        return self.svc
+
+    @service.setter
+    def service(self, data):
+        ''' setter function for yedit var '''
```
tbielawa: yedit says what?

roles/lib_openshift/src/class/oc_service.py
```diff
@@ -0,0 +1,170 @@
...
+        self.svc = data
+
+    def exists(self):
+        ''' return whether a volume exists '''
```
tbielawa: Volumes or services?

roles/lib_openshift/src/class/oc_service.py
```diff
@@ -0,0 +1,170 @@
...
+        if self.service:
+            return True
+
+        return False
+
+    def get(self):
+        '''return volume information '''
```
tbielawa: Volumes or services?

roles/lib_openshift/src/class/oc_service.py
```diff
@@ -0,0 +1,170 @@
...
+            self.service = Service(content=result['results'][0])
+            result['clusterip'] = self.service.get('spec.clusterIP')
+        elif 'services \"%s\" not found' % self.config.name  in result['stderr']:
+            result['clusterip'] = ''
+            result['returncode'] = 0
+
+        return result
+
+    def delete(self):
+        '''delete the object'''
```
tbielawa: object or service? Same thing?

roles/lib_openshift/src/doc/service
```diff
@@ -0,0 +1,121 @@
...
+---
+module: oc_service
+short_description: Create, modify, and idempotently manage openshift services.
+description:
+  - Manage openshift service objects programmatically.
+options:
+  state:
+    description:
+    - State represents whether to create, modify, delete, or list
+    required: true
```
tbielawa: Options with `default`s set do not need `required` set to `true`. For example, look at `kubeconfig` below.

roles/lib_openshift/src/doc/service
```diff
@@ -0,0 +1,121 @@
...
+    aliases: []
+  clusterip:
+    description:
+    - The cluster ip address to use with this service.
+    required: false
+    default: None
+    aliases: []
+  portalip:
+    description:
+    - The portal ip address to use with this service.
```
tbielawa: What is a portal IP address?

roles/lib_openshift/src/lib/service.py
```diff
@@ -0,0 +1,126 @@
...
+        self.cluster_ip = cluster_ip
+        self.portal_ip = portal_ip
+        self.session_affinity = session_affinity
+        self.service_type = service_type
+        self.data = {}
+
+        self.create_dict()
+
+    def create_dict(self):
+        ''' return a service as a dict '''
```
tbielawa: This doesn't actually have a `return` statement in it. Perhaps this is intended to say that it creates a service attribute as a dict?

roles/lib_openshift/src/lib/service.py
```diff
@@ -0,0 +1,126 @@
...
+            self.data['spec']['clusterIP'] = self.cluster_ip
+
+        if self.portal_ip:
+            self.data['spec']['portalIP'] = self.portal_ip
+
+        if self.service_type:
+            self.data['spec']['type'] = self.service_type
+
+# pylint: disable=too-many-instance-attributes,too-many-public-methods
+class Service(Yedit):
```
tbielawa: Why does this subclass `Yedit`? It's not clear to me.

roles/lib_openshift/src/doc/service
kwoodson: Virtual IP of the service, allocated automatically at creation from a pool of internal IPs.

roles/lib_openshift/src/doc/service
kwoodson: fixing.

roles/lib_openshift/src/class/oc_service.py
kwoodson: fixing

roles/lib_openshift/src/class/oc_service.py
kwoodson: fixing

roles/lib_openshift/src/class/oc_service.py
kwoodson: fixing

roles/lib_openshift/src/class/oc_service.py
kwoodson: fixing

roles/lib_openshift/src/lib/service.py
kwoodson: @tbielawa 

This was a design decision that I made when deciding on how I wanted to model these openshift objects.  All of these openshift objects are just basic dictionary data structures.  Yedit is basically a dictionary structure manipulation tool.  It made sense to me to model a service(or any other) oc object on top of a yedit class because it allows you to then set specific paths to items inside of the object's data and build a class around them.

For example (from the PR):
`+    cluster_ip = "spec.clusterIP"`
I can define the path to where attributes live inside of a dictionary and then can code getters/setters on top of it.  Since the data coming out of the oc commands is json or yaml, we can just send the string into the Service(Yedit) class and it can parse and store the data.

I am open to updating the modeling of the classes and I think that we can get more out of the object comparison's when go that route.  This is a longer term discussion and I don't think should block on these.

tbielawa: > I can define the path to where attributes live inside of a dictionary and then can code getters/setters on top of it. Since the data coming out of the oc commands is json or yaml, we can just send the string into the Service(Yedit) class and it can parse and store the data.

That answers my question.

> I am open to updating the modeling of the classes and I think that we can get more out of the object comparison's when go that route. This is a longer term discussion and I don't think should block on these.

Now that I understand it, I am picking up what you are throwing down, as it were. A longer term discussion in the future, I'd be OK with that. But like you said, I wouldn't block on that.

roles/lib_openshift/src/lib/service.py
kwoodson: dang, missed this one.  Will fix.

kwoodson:
aoc-ci-test

kwoodson:
@tbielawa, bump

Just want to keep this on your radar.

PR #3188, opened 2017-01-26
Backport nodename changes for OpenStack and GCE to 3.3

detiber:

PR #3225, opened 2017-01-30
GCE deployment fails due to invalid lookup

smarterclayton:
Data is only available under ['instance'] and split was in the wrong
order.  This deploys against GCE for me.

Change in #3172 did not work for me


smarterclayton:
@sdodson @detiber this works for me while the previous does not

sdodson:
Need to reconcile what's changing here with #3188

sdodson:
aos-ci-test

smarterclayton:
https://ci.openshift.redhat.com/jenkins/job/zz_origin_gce_image/79/console got past the previous failure

smarterclayton:
This is not enough - we also have to ensure that the node name is signed for in the certs (which doesn't look it it is happening).  That's why we used the full name in the first place.

PR #3231, opened 2017-01-31
Fail on Ansible version 2.2.1.0

mtnbikenc:
Due to issues with Ansible 2.2.1.0, the version verification has been updated to fail when attempting to use that version.

### DIFF
--- a/callback_plugins/aa_version_requirement.py
+++ b/callback_plugins/aa_version_requirement.py
@@ -30,7 +30,12 @@ else:
 
 # Set to minimum required Ansible version
 REQUIRED_VERSION = '2.2.0.0'
-DESCRIPTION = "Supported versions: %s or newer" % REQUIRED_VERSION
+DESCRIPTION = "Supported versions: %s or newer (except 2.2.1.0)" % REQUIRED_VERSION
+FAIL_ON_2_2_1_0 = "There are known issues with Ansible version 2.2.1.0 which " \
+                  "are impacting OpenShift-Ansible. Please use Ansible " \
+                  "version 2.2.0.0 or a version greater than 2.2.1.0. " \
+                  "See this issue for more details: " \
+                  "https://github.com/openshift/openshift-ansible/issues/3111"
 
 
 def version_requirement(version):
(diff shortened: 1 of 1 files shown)


callback_plugins/aa_version_requirement.py
```diff
@@ -31,6 +31,10 @@ def display(*args, **kwargs):
 # Set to minimum required Ansible version
 REQUIRED_VERSION = '2.2.0.0'
 DESCRIPTION = "Supported versions: %s or newer" % REQUIRED_VERSION
+FAIL_ON_2_2_1_0 = "There are known issues with Ansible version 2.2.1.0 which " \
+                  "are impacting OpenShift-Ansible. Please downgrade to " \
+                  "Ansible version 2.2.0.0. See this issue for more details: " \
```
detiber: Maybe instead of saying downgrade to version 2.2.0.0, we should say to downgrade to 2.2.0.0 or upgrade to a newer version, that will prevent having to update this once a new version is released that is not broken.

tbielawa:
(First I built && installed an RPM from the `ansible-version` branch)
```
$ rpm -q ansible
ansible-2.2.1.0-1.fc25.noarch

$ ansible-playbook -v -i ~/.config/openshift/hosts ./playbooks/byo/config.yml
Using /home/tbielawa/rhat/os/openshift-ansible/ansible.cfg as config file
FATAL: Current Ansible version (2.2.1.0) is not supported. There are known issues with Ansible version 2.2.1.0 which are impacting OpenShift-Ansible. Please downgrade to Ansible version 2.2.0.0. See this issue for more details: https://github.com/openshift/openshift-ansible/issues/3111
[~/rhat/openshift/openshift-ansible] 8:03:51  (ansible-version)
```

Works as expected!

I tried to do a test w/ an older ansible install but the spec file required newer versions.

callback_plugins/aa_version_requirement.py
tbielawa: Hm. That's smart too.

mtnbikenc: I can update the guidance.  And it would probably be good to update the `or newer` reference so someone doesn't upgrade from an older version only to find out they have to downgrade if they went to 2.2.1.0.

mtnbikenc:
@tbielawa @detiber Updated the version usage guidance.

mtnbikenc:
cc: @abutcher

mtnbikenc:
aos-ci-test

PR #3232, opened 2017-01-31
During node upgrade upgrade openvswitch rpms

sdodson:
Containerized upgrades of openvswitch are already handled by updating
the container image which will have the requisite version of openvswitch already.


sdodson:
Also, specific order of operations for restarts should be ovs, docker, node. But since the node is evacuated this isn't critical.

sdodson:
aos-ci-test

PR #3234, opened 2017-01-31
Replace multi-role checks with action-plugin-based checks

rhcarvalho:
The Customer Success team is writing diagnose checks to identify problems in OpenShift clusters.

In that context, the structure we've added previously using multiple subroles was a pain for maintenance, prone to subtle mistakes, and it was hard to keep adding new checks.

This presents a rewrite of the existing checks using a different approach, based on an action plugin to control what checks are executed, execute them and report back execution status.

New checks can be added with regular Python modules that will typically execute one or more Ansible modules, either builtin or custom.

There's also a simple mechanism for tagging checks and selecting all checks that match a given tag.


rhcarvalho:
@brenton @codificat could you try this in your envs?

@detiber FYI you wanted to see this experiment :-)


TODO:
- [ ] a sane way of smoke testing those checks

roles/openshift_health_checker/openshift_checks/package_availability.py
```diff
@@ -0,0 +1,69 @@
...
+            "{rpm_prefix}-clients".format(rpm_prefix=rpm_prefix),
+            "{rpm_prefix}-master".format(rpm_prefix=rpm_prefix),
+            "bash-completion",
+            "cockpit-bridge",
+            "cockpit-docker",
+            "cockpit-kubernetes",
+            "cockpit-shell",
+            "cockpit-ws",
+            "etcd",
+            "httpd-tools",
```
detiber: Definitely not related to this, but I think we should make sure that there is a card on either CS or CL boards to track adding support for managing these package lists in a consistent way between the checks and the roles that are installing the packages. @sdodson

roles/openshift_health_checker/openshift_checks/package_availability.py
```diff
@@ -0,0 +1,69 @@
...
+    """Check that required RPM packages are available."""
+
+    name = "package_availability"
+    tags = ["preflight"]
+
+    def run(self, tmp, task_vars):
+        try:
+            rpm_prefix = task_vars["openshift"]["common"]["service_type"]
+        except (KeyError, TypeError):
+            raise OpenShiftCheckException("'openshift.common.service_type' is undefined")
```
detiber: I'm wondering if there is a better way to handle this rather than just pulling the value from the task_vars...  

I think it might be better to define an explicit interface for the checks, that way we are not bleeding the variable datastructures all the way down into the module. Especially as we start moving away from openshift_facts, I worry that this will cause some problems down the line.

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -0,0 +1,25 @@
...
+    """Check that available RPM packages match the required versions."""
+
+    name = "package_version"
+    tags = ["preflight"]
+
+    @classmethod
+    def is_active(cls, task_vars):
+        return (
+            super(PackageVersion, cls).is_active(task_vars)
+            and task_vars.get("deployment_type") == "openshift-enterprise"
```
detiber: It might be good to surface up deployment_type up to the base class as well, then you could have checks that are specific to a certain type of deployment_type and is_active could take it into account automatically as well.

rhcarvalho:
@detiber thanks for the review. Will reply inline.

roles/openshift_health_checker/openshift_checks/package_availability.py
rhcarvalho: I agree. This PR is just moving the list that was once in a play into a Python list. How to compute this automatically is still up for investigation.

roles/openshift_health_checker/openshift_checks/package_version.py
rhcarvalho: This particular check looking up `deployment_type` is a bug: https://trello.com/c/NdWoIydb
This check applies equally well to any deployment_type, the package names should not be hard-coded in the `aos_version.py` module...

I would fix it in a separate PR, but if it helps reduce the confusion I'll do it in a new commit.

rhcarvalho: FWIW, another known limitation is that these modules work only with YUM, and not DNF.

roles/openshift_health_checker/openshift_checks/package_availability.py
rhcarvalho: Up to now the interface follows from `OpenShiftCheck`: you need to provide a name and a `run` method with the same signature as an action plugin and a module call (tmp and task_vars are input, the return value is a dict). The semantics are also the same as action plugin/module.

Yes, taking `task_vars` in smells like a violation of the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter), but we're just mimicking how action plugins / modules / Ansible works, I haven't found a better way.

If we would take only the values each check requires, then they would not have a common API, and code would need to know how to instantiate each check with a different set of arguments, and that would put us in a worse situation IMHO.

If we are moving away from `openshift_facts` (are we?) then I expect many things to break, because the whole code base depends on the ad-hoc structure of variables.

roles/openshift_health_checker/openshift_checks/mixins.py
```diff
@@ -0,0 +1,19 @@
+from openshift_checks import OpenShiftCheckException
+
+
+class NotContainerized(object):
```
rhcarvalho: What I have identified so far is that we need to differentiate between `failed`, `passed` or `skipped` (N/A). `is_active` answers whether to skip the check.

What problem is the alternative trying to solve / improve?

I see it increases the API surface and spreads the knowledge about `task_vars` to even more methods.

From a code maintenance point of view, it adds more burden to check authors. Each check would need to overwrite one or more `supports_*` method, compared to adding the right set of mixins.

A new `supports_system_containers` method requires changing multiple places, failing the Open-Closed Principle...

detiber: I worry a bit about mixins causing a bit of confusion with developers that aren't already, but I'm good with keeping it if it is the cleanest approach.

I think we'll need to think of enabling/disabling sets of classes more as a matrix, for example just considering atomic host/non-atomic host and containerized/non-containterized/system-containers:

|   | atomic | non-atomic |
|---|---|---|
| containerized |   |   |
| non-containerized  |   |
| system containers  |   |

I believe there will be similar matrices once we start tackling different options being enabled/disabled for each service type.

roles/openshift_health_checker/openshift_checks/package_availability.py
detiber: One thing that I worry about, is that the checks are being run similar to modules, but they have access to variables that are not available to normal modules because task_vars are being passed through from the action plugin...  

I just worry that we are blending abstractions a bit too much here.

rhcarvalho: I agree with you, in this sense the checks are more like action plugins than modules.
The inherent difficulty is we need to have code somewhere that initializes the check with the proper information it needs, doing it on the loader/runner side makes things hard to deal with. Putting it at the control of the check developer allows for the uniform API.

If we create yet another layer for "check instantiation" (that would feed in to a check information like `openshift.common.service_type` without leaking the whole `task_vars`), then check developers would need to code both the check and the instantiation (probably taking data from `task_vars`, since that's the only input to the action plugin).

roles/openshift_health_checker/HOWTO_CHECKS.md
```diff
@@ -0,0 +1,34 @@
+# OpenShift health checks
```
rhcarvalho: @brenton FYI this is the "how to add a new check"

rhcarvalho:
aos-ci-test

rhcarvalho:
@detiber @abutcher could one of you take a final look please?

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -5,6 +5,13 @@
 import os
 from abc import ABCMeta, abstractmethod, abstractproperty
 from importlib import import_module
+import operator
+
+try:
+    # for Python 3
+    from functools import reduce  # pylint: disable=redefined-builtin
```
detiber: You could use six.moves here instead of try/except.

rhcarvalho: Sure, done.

mtnbikenc:
This now looks and behaves more like a regular Ansible role. LGTM :+1:

rhcarvalho:
@sdodson could I have your help to merge this one please?

PR #3237, opened 2017-01-31
Adding bool filter to ensure that we correctly set ops host for fluentd

ewolinetz:
If the openshift_logging_use_ops=false then it would evaluate incorrectly.
This resolves that issue.

@jcantrill @sdodson


ewolinetz:
aos-ci-test

PR #3238, opened 2017-01-31
[WIP] Pushing a simple dashboard to kibana

enoodle:
Pushing dashboards into Kibana.

### DIFF
--- a/roles/openshift_hosted_logging/files/dashboard.json
+++ b/roles/openshift_hosted_logging/files/dashboard.json
@@ -1,0 +1,13 @@
+{
+  "title": "my-first-dashboard",
+  "hits": 0,
+  "description": "",
+  "panelsJSON": "[{\"id\":\"hostnames\",\"type\":\"visualization\",\"panelIndex\":1,\"size_x\":7,\"size_y\":4,\"col\":1,\"row\":1}]",
+  "optionsJSON": "{\"darkTheme\":true}",
+  "uiStateJSON": "{}",
+  "version": 1,
+  "timeRestore": false,
+  "kibanaSavedObjectMeta": {
+    "searchSourceJSON": "{\"filter\":[{\"query\":{\"query_string\":{\"query\":\"*\",\"analyze_wildcard\":true}}}]}"
+  }
+}
--- a/roles/openshift_hosted_logging/tasks/deploy_logging.yaml
+++ b/roles/openshift_hosted_logging/tasks/deploy_logging.yaml
@@ -167,6 +167,16 @@
 
 - include: update_master_config.yaml
 
+- name: "Copy Dashboards for Kibana"
+  copy:
+    src: roles/openshift_hosted_logging/files/dashboard.json
+    dest: ~/manageiq_ops_dashboard.json
+
+- name: "Push Dashboards to Kibana"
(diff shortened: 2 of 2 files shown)


enoodle:
cc @simon3z @portante

A few problems that I encountered while making this:
- Had to access Elasticsearch through Kibana because we don't have a route to Elasticsearch.
- To access Kibana I used the hostname that was configured, but if it is not in accessible through it this will fail.
- I didn't know where to copy the dashboard file to
- In the final version will need to copy / push many dashboards and visualizations so maybe a new dedicated role is needed for that?

detiber:
@enoodle This should probably be incorporated into the new openshift_logging role instead. @ewolinetz can answer any questions.

ewolinetz:
@enoodle 

A few things:
1) the catch with pushing dashboards into Elasticsearch for Kibana is that we create a kibana 'profile' for each user. So there would likely need to be logic in https://github.com/fabric8io/openshift-elasticsearch-plugin/ to copy any provided dashboards out of the .kibana index for a user's profile index when being created.

2) Unless ES is currently running, you would not be able to push to its indices. Also, if we only seed the dashboard from this role then if the Elasticsearch cluster doesn't have persisted storage backing it, the dashboard would be lost upon subsequent start ups.

I would almost think that this would better belong in the elasticsearch [1] image definition, similar to how index_templates are added to the cluster upon start up.

[1] https://github.com/openshift/origin-aggregated-logging/tree/master/elasticsearch

enoodle:
@ewolinetz 
1. Yes, I forgot to mention that only the user whose credentials are used will be able to access the dashboard that is pushed. Having a way to create global dashboards would be wonderful.
2. The previous tasks in this role assure that ES and Kibana are running before getting to pushing the dashboard.

Adding the dashboard to the container image is also a possible option, we would like to see which is better, because the dashboards may assume things on the common data model used. @simon3z  @portante

roles/openshift_hosted_logging/files/dashboard.json
```diff
@@ -0,0 +1,13 @@
+{
+  "title": "my-first-dashboard",
```
portante: Perhaps the title should be "ManageIQ Ops Overview"?

roles/openshift_hosted_logging/tasks/deploy_logging.yaml
```diff
@@ -167,6 +167,16 @@
 
 - include: update_master_config.yaml
 
+- name: "Copy Dashboards for Kibana"
+  copy:
+    src: roles/openshift_hosted_logging/files/dashboard.json
+    dest: ~/manageiq_ops_dashboard.json
+
+- name: "Push Dashboards to Kibana"
+  shell: "curl -X PUT -k -H \"kbn-version: 4.5.4\" -H \"Authorization: Bearer `{{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig sa get-token management-admin -n management-infra`\" https://{{ openshift_hosted_logging_ops_hostname }}/elasticsearch/.kibana/dashboard/ManageIQOpsDashboard -d @manageiq_ops_dashboard.json"
```
portante: s/ManageIQOpsDashboard/ManageIQOpsOverview/ to avoid any indication this came from the "Department of the Redundancy Department".

Also, is this really only for ManageIQ?  Perhaps we should just name this the Ops Overview dashboard instead of including the ManageIQ name there, since it would be useful to have outside of ManageIQ?

enoodle:
closing in favor of: https://github.com/openshift/origin-aggregated-logging/pull/334

PR #3239, opened 2017-01-31
Adding bool filter to when openshift_logging_use_ops evals and updati…

### LINKED ISSUE #3201: openshift_logging role updating route fails because change is immutable

ewolinetz:
…ng oc_apply to handle trying to update immutable fields

Alternate proposal to fix #3201 

If we are trying to update an object with an immutable field we should recreate it instead.

@jcantrill @detiber  Thoughts?

### DIFF
--- a/playbooks/common/openshift-cluster/openshift_logging.yml
+++ b/playbooks/common/openshift-cluster/openshift_logging.yml
@@ -7,7 +7,8 @@
 - name: Update Master configs
   hosts: masters:!oo_first_master
   tasks:
-  - include_role:
-      name: openshift_logging
-      tasks_from: update_master_config
+  - block:
+    - include_role:
+        name: openshift_logging
+        tasks_from: update_master_config
     when: openshift_logging_install_logging | default(false) | bool
--- a/roles/openshift_logging/tasks/oc_apply.yaml
+++ b/roles/openshift_logging/tasks/oc_apply.yaml
@@ -18,6 +18,26 @@
   failed_when: "'error' in generation_apply.stderr"
   changed_when: no
 
+- name: Removing previous {{file_name}}
+  command: >
+    {{ openshift.common.client_binary }} --config={{ kubeconfig }}
+    delete -f {{ file_name }}
+    -n {{ namespace }}
+  register: generation_delete
+  failed_when: "'error' in generation_delete.stderr"
+  changed_when: no
+  when: "'field is immutable' in generation_apply.stderr"
+
+- name: Applying {{file_name}}
+  command: >
+    {{ openshift.common.client_binary }} --config={{ kubeconfig }}
+    apply -f {{ file_name }}
+    -n {{ namespace }}
+  register: generation_apply
+  failed_when: "'error' in generation_apply.stderr"
+  changed_when: no
+  when: "'field is immutable' in generation_apply.stderr"
+
 - name: Determine change status of {{file_content.kind}} {{file_content.metadata.name}}
   shell: >
     {{ openshift.common.client_binary }} --config={{ kubeconfig }}
--- a/roles/openshift_logging/tasks/oc_apply.yaml
+++ b/roles/openshift_logging/tasks/oc_apply.yaml
@@ -18,26 +18,6 @@
   failed_when: "'error' in generation_apply.stderr"
   changed_when: no
 
-- name: Removing previous {{file_name}}
-  command: >
-    {{ openshift.common.client_binary }} --config={{ kubeconfig }}
-    delete -f {{ file_name }}
-    -n {{ namespace }}
-  register: generation_delete
-  failed_when: "'error' in generation_delete.stderr"
-  changed_when: no
-  when: "'field is immutable' in generation_apply.stderr"
-
-- name: Applying {{file_name}}
-  command: >
-    {{ openshift.common.client_binary }} --config={{ kubeconfig }}
-    apply -f {{ file_name }}
-    -n {{ namespace }}
-  register: generation_apply
-  failed_when: "'error' in generation_apply.stderr"
-  changed_when: no
-  when: "'field is immutable' in generation_apply.stderr"
-
 - name: Determine change status of {{file_content.kind}} {{file_content.metadata.name}}
   shell: >
     {{ openshift.common.client_binary }} --config={{ kubeconfig }}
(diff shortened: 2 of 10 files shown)


jcantrill:
aos-ci-test

roles/openshift_logging/tasks/oc_apply.yaml
```diff
@@ -18,6 +18,26 @@
   failed_when: "'error' in generation_apply.stderr"
   changed_when: no
```
detiber: Doesn't seem right to be setting `changed_when: no` on an action that would be mutating the state.

roles/openshift_logging/tasks/oc_apply.yaml
```diff
@@ -18,6 +18,26 @@
...
 
+- name: Removing previous {{file_name}}
+  command: >
+    {{ openshift.common.client_binary }} --config={{ kubeconfig }}
+    delete -f {{ file_name }}
+    -n {{ namespace }}
+  register: generation_delete
+  failed_when: "'error' in generation_delete.stderr"
+  changed_when: no
+  when: "'field is immutable' in generation_apply.stderr"
```
detiber: `changed_when: no` seems wrong here as well.

roles/openshift_logging/tasks/oc_apply.yaml
```diff
@@ -18,6 +18,26 @@
...
+
+- name: Applying {{file_name}}
+  command: >
+    {{ openshift.common.client_binary }} --config={{ kubeconfig }}
+    apply -f {{ file_name }}
+    -n {{ namespace }}
+  register: generation_apply
+  failed_when: "'error' in generation_apply.stderr"
+  changed_when: no
+  when: "'field is immutable' in generation_apply.stderr"
```
detiber: and here (for `changed_when: no`

roles/openshift_logging/tasks/oc_apply.yaml
ewolinetz: Oh, so this is 'no' here because oc apply will have a rc of 0 even if it didn't actually change anything. So we evaluate the resourceVersion to see if it changed afterwards (lines 2 and then 41)

roles/openshift_logging/tasks/oc_apply.yaml
```diff
@@ -27,3 +47,6 @@
   register: generation_changed
```
detiber: If you are already overriding failed_when, you shouldn't need `|| echo 0` here.

roles/openshift_logging/tasks/oc_apply.yaml
detiber: Is there any way to do the check closer as the immediate next task? I worry about the case where the playbook is interrupted or there is a failure between this task an the change being reported...

Ideally, there would be a way to get the object back from apply to immediately validate the changed status in the same task.

ewolinetz: Makes sense, let me see what I can get back in output when we `oc apply`

ewolinetz: we could only get the name as output with `oc apply` so I moved the check to be immediately after the apply task we are evaluating

roles/openshift_logging/tasks/oc_apply.yaml
ewolinetz: removed `failed_when`

detiber: Could you keep failed_when and use command instead of shell?

ewolinetz: Updated to not use `shell` or `|| echo 0`

ewolinetz:
aos-ci-test

playbooks/common/openshift-cluster/openshift_logging.yml
```diff
@@ -7,7 +7,8 @@
 - name: Update Master configs
   hosts: masters:!oo_first_master
   tasks:
-  - include_role:
-      name: openshift_logging
-      tasks_from: update_master_config
+  - block:
+    - include_role:
+        name: openshift_logging
+        tasks_from: update_master_config
```
detiber: @abutcher do you know if upstream is tracking this issue?

PR #218, opened 2015-05-08
renamed ansible.cfg to ansible.cfg.example to avoid ansible accidentally picking this up as it's config file.

twiest:

PR #3241, opened 2017-02-01
Add default ansible.cfg file

mtnbikenc:
By default the callback_plugins need to be run to ensure proper Ansible version

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -3,15 +3,18 @@
 
 # This config file provides examples for running
 # the OpenShift playbooks with the provided
-# inventory scripts. Only global defaults are
-# left uncommented
+# inventory scripts.
 
 [defaults]
-# Add the roles directory to the roles path
-roles_path = roles/
-
 # Set the log_path
-log_path = /tmp/ansible.log
+#log_path = /tmp/ansible.log
+
+# Additional default options for OpenShift Ansible
+callback_plugins = callback_plugins/
+forks = 20
+host_key_checking = False
+retry_files_enabled = False
+nocows = True
 
 # Uncomment to use the provided BYO inventory
 #hostfile = inventory/byo/hosts
@@ -21,3 +24,7 @@ log_path = /tmp/ansible.log
 
 # Uncomment to use the provided AWS dynamic inventory script
 #hostfile = inventory/aws/ec2.py
+
+# Additional ssh options for OpenShift Ansible
+[ssh_connection]
+pipelining = True
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ gce.ini
 multi_ec2.yaml
 .vagrant
 .tags*
-/ansible.cfg
 *.retry
 .vscode/*
 .cache


ansible.cfg
```diff
@@ -3,13 +3,15 @@
...
-# left uncommented
+# inventory scripts.
 
 [defaults]
 # Add the roles directory to the roles path
 roles_path = roles/
 
+# Add the callback_plugins path
+callback_plugins = callback_plugins/
+
```
detiber: Should we also add some sane defaults? I would suggest the following at minimum:
default:
- forks=20
- host_key_checking=False
- retry_files_enabled=False
- no_cows=True
- log_path=/tmp/ansible.log (commented out)

ssh_connection:
- pipelining=True

detiber:
aos-ci-test

tbielawa:
I am hesitant about this because I actually use an `ansible.cfg` file in my openshift-ansible checkout. This will give me conflicts and what have you :-(

detiber:
@tbielawa it will impact me as well. In my view, I would rather impact the developer workflow to avoid users hitting issues, especially around using the wrong ansible versions.

tbielawa:
@detiber said

> @tbielawa it will impact me as well. In my view, **I would rather impact the developer workflow** to avoid users hitting issues, especially around using the wrong ansible versions.

Yes, ultimately it comes down to that. I'm a developer, I can figure out a way to work around it. Customers must come **first**.

detiber:
@tbielawa btw, simple fix is to use ~/.ansible.cfg (or ansible.cfg in a different working dir) and run from outside the repo root :)

mtnbikenc:
aos-ci-test

rhcarvalho:
One more developer hit by this change.

For the work I've been doing as I learn Ansible, the solution to "ensure a certain callback plugin is loaded" was to have the [callback plugin in a role](https://github.com/openshift/openshift-ansible/tree/15378ed14f5c15d6193a431bd9a6fd1eef0471ad/roles/openshift_preflight/verify_status/callback_plugins) and then load the role.

How about we use the same approach here? Would it work with older versions of Ansible?

Shipping a default `ansible.cfg` would not prevent people to have their own configs that would take precedence over this config in the repo, just like the suggestion above to create a `~/.ansible.cfg` file.

Looking at the project history, `ansible.cfg` became `ansible.cfg.example` long ago, so I suspect it was for a good reason. https://github.com/openshift/openshift-ansible/pull/218

detiber:
@rhcarvalho The issue with relying on the role for the version checking callback plugin is that an issue can be encountered in parsing the playbooks before the role is processed. I agree that we should be shipping it in a role as well, though.

The lookup path for ansible config is as follows:
- ansible.cfg in cwd
- ~/.ansible.cfg
- /etc/ansible/ansible.cfg
using the first found file, so this ansible.cfg would take precedence for users running from the root of the checkout.

We originally moved ansible.cfg to ansible.cfg.example because at the time Ops was running their tooling from the root of the checkout and did not want our settings overriding their ansible.cfg settings, they are no longer doing this.

I think it's a good thing for us to provide a default, known working ansible.cfg file that we can tune/tweak for performance, etc. Being able to check Ansible version before parsing and running the playbooks is just a plus.

rhcarvalho:
Given the lookup order, for whoever was running from the root of the checkout, the behavior will suddenly change (as it happened to me, and then I had to spend some time to realize why things were not working after a rebase -- luckily that also revealed a bug in my playbook ;-) ).

I see there could be a problem before the role is loaded, would it help to make the role the first and foremost thing in the entrypoint playbooks, using a syntax that is broadly compatible?


And what about checking the version with the [`assert`](http://docs.ansible.com/ansible/assert_module.html) module, I see it mentioned in http://stackoverflow.com/a/34811162/4804690, is that of any relevance to us?

mtnbikenc:
One of the challenges we also have with versioning in Ansible, is that internally a `--syntax-check` is run before the playbooks are executed, and will flag any unknown module usage as a 'typo'.  This happens as Ansible introduces new modules, and we use them, which then causes older versions of Ansible break.  This playbook checking precludes the ability to use any type of `assert` task or module in a role because they would never be called before the syntax-check is run.  By putting the version verification in a callback plugin, it will be executed as Ansible is loaded and will fail appropriately before the syntax-check.  We've learned that Ansible version verification is tricky business.

rhcarvalho:
The next thing I'd think about is a wrapper around Ansible... we already had that with the `bin/cluster` tool, but I understand it is unsupported at this point. Different than `bin/cluster`, we could have something much simpler that 1) checks the Ansible version and then 2) calls `ansible-playbook` (potentially forwarding any command line arguments).

Does that have any known drawbacks you guys have come across in the past?

detiber:
> Given the lookup order, for whoever was running from the root of the checkout, the behavior will suddenly change (as it happened to me, and then I had to spend some time to realize why things were not working after a rebase -- luckily that also revealed a bug in my playbook ;-) ).

This highlights another reason why we should actually ship a default ansible.cfg. It will provide a more consistent developer environment going forward :)

> The next thing I'd think about is a wrapper around Ansible... we already had that with the bin/cluster tool, but I understand it is unsupported at this point. Different than bin/cluster, we could have something much simpler that 1) checks the Ansible version and then 2) calls ansible-playbook (potentially forwarding any command line arguments).

> Does that have any known drawbacks you guys have come across in the past?

Yes, it breaks use with Ansible-native workflows. We want to support people integrating into their existing Ansible tooling whether they are using Ansible Core or Ansible Tower.

rhcarvalho:
There are still some points that are unclear to me.

- I see the value in supporting a standard Ansible workflow;
- I understand that having the `ansible.cfg` in the root of the repo may have its advantages;

But,

- I'm still failing to see how that `ansible.cfg` nails the issue of ensuring the callback plugin to check the version gets loaded. It works in "our workflow" because we're running `ansible-playbook` from the root of the repo.

How do people go about integrating with openshift-ansible?
Do they have their own repo and call out e.g. `playbooks/byo/config.yml` from their custom playbook? If that is the case, then they are likely running `ansible-playbook` from a different working directory, and that won't use our `ansible.cfg`.

What about Ansible Tower, I'm not familiar with it, what is the likelihood of an Ansible Tower workflow picking up this config in the root of the repo?

detiber:
It doesn't directly address the integration or Tower use cases directly (I brought those up for why we should discourage a wrapper).

The integration and tower use cases are more advanced, I'm not sure we can fully solve the Ansible version check there and would need to rely on the documented supported version for those use cases.

PR #3242, opened 2017-02-01
Move current node upgrade tasks under openshift_node_upgrade role

ingvagabund:
Moving some tasks from ``playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml`` into separate role. Preparation for running upgrade nodes play as a part of control plane upgrade over masters.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
@@ -5,12 +5,7 @@
   # and is evaluated early. Values such as "20%" can also be used.
   serial: "{{ openshift_upgrade_nodes_serial | default(1) }}"
   any_errors_fatal: true
-  roles:
-  - openshift_facts
-  - docker
-  handlers:
-  - include: ../../../../roles/openshift_node/handlers/main.yml
-    static: yes
+
   pre_tasks:
   # TODO: To better handle re-trying failed upgrades, it would be nice to check if the node
   # or docker actually needs an upgrade before proceeding. Perhaps best to save this until
@@ -21,17 +16,14 @@
     register: node_output
     delegate_to: "{{ groups.oo_first_master.0 }}"
     changed_when: false
-    when: inventory_hostname in groups.oo_nodes_to_upgrade
 
   - set_fact:
       was_schedulable: "{{ 'unschedulable' not in (node_output.stdout | from_json).spec }}"
-    when: inventory_hostname in groups.oo_nodes_to_upgrade
 
   - name: Mark unschedulable if host is a node
     command: >
       {{ hostvars[groups.oo_first_master.0].openshift.common.client_binary }} adm manage-node {{ openshift.node.nodename | lower }} --schedulable=false
     delegate_to: "{{ groups.oo_first_master.0 }}"
-    when: inventory_hostname in groups.oo_nodes_to_upgrade
     # NOTE: There is a transient "object has been modified" error here, allow a couple
     # retries for a more reliable upgrade.
     register: node_unsched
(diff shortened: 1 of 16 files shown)


ingvagabund:
aos-ci-test

roles/openshift_node_upgrade/meta/main.yml
```diff
@@ -0,0 +1,12 @@
...
+  description: OpenShift Node upgrade
+  company: Red Hat, Inc.
+  license: Apache License, Version 2.0
+  min_ansible_version: 2.1
+  platforms:
+  - name: EL
+    versions:
+    - 7
+dependencies:
+  - role: openshift_common
```
dgoodwin: Arguably openshift_version is a dep here as well, which basically just no-ops if it sees that the three core variables are already set.

playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
```diff
@@ -43,83 +35,18 @@
     command: >
       {{ hostvars[groups.oo_first_master.0].openshift.common.admin_binary }} drain {{ openshift.node.nodename | lower }} --force --delete-local-data
     delegate_to: "{{ groups.oo_first_master.0 }}"
```
dgoodwin: How does everyone feel about keeping drain / mark schedulable out of the role? This implementation seems best to me as it's probably better, maybe even necessary, to control the ordering when we get to including it in control plane upgrade.

roles/openshift_node_upgrade/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+Example Playbook
+----------------
+
+Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too:
+
+```
+---
+- name: Upgrade nodes
+  hosts: oo_nodes_to_upgrade
```
dgoodwin: At some point how this group is calculated should be documented here in the readme, specifically how you can run node upgrade on only nodes with a specific label, and with a value for serial.

roles/openshift_node_upgrade/tasks/docker/restart.yml
```diff
@@ -0,0 +1,33 @@
...
+  failed_when: false
+  when: openshift.common.is_containerized | bool
+
+- name: Wait for master API to come back online
+  wait_for:
+    host: "{{ openshift.common.hostname }}"
+    state: started
+    delay: 10
+    port: "{{ openshift.master.api_port }}"
+  when: inventory_hostname in groups.oo_masters_to_config
```
dgoodwin: This must be the duplication you referred to. (with playbooks/common/openshift-cluster/upgrades/docker/restart.yml) Would this be best solved with an openshift_docker_restart role? (not now but in future)

roles/openshift_node_upgrade/tasks/docker/upgrade.yml
```diff
@@ -0,0 +1,49 @@
...
+- debug: var=docker_image_count.stdout
+  when: docker_upgrade_nuke_images is defined and docker_upgrade_nuke_images | bool
+
+- service: name=docker state=stopped
+
+- name: Upgrade Docker
+  package: name=docker{{ '-' + docker_version }} state=present
+
+- include: restart.yml
+  when: not skip_docker_restart | default(False) | bool
```
dgoodwin: Likewise probably a good candidate for a role later on too.

sdodson:
@dgoodwin are you requesting changes or ok to merge as is?

ingvagabund:
aos-ci-test

roles/openshift_node_upgrade/README.md
ingvagabund: This is worth doing for each and every role we currently have.

roles/openshift_node_upgrade/tasks/docker/restart.yml
ingvagabund: * files/nuke_images.sh
* handlers/main.yml
* tasks/containerized_node_upgrade.yml
* tasks/docker/*
* tasks/rpm_upgrade.yml
* tasks/systemd_units.yml
* templates/*

Most of the files is picked from ``playbooks/common/openshift-cluster/upgrades`` directory. Most likely, most of the files can be removed from that. Need to check each place where they are actually used before that. The remaining files are picked from `openshift_node` role. In the former case, the duplicas can be eventually removed. Just the `openshift_node` files will stay.

ingvagabund: `openshift_docker_restart role` makes sense to me. Yet, that is something we need to decide globally as part of a bigger re-factoring.

mtnbikenc:
Upgrade from release-1.4 tested.  :+1:

PR #3243, opened 2017-02-01
Upgrade node part of masters as part of control plane upgrade

ingvagabund:
If a master is also a regular node, run its upgrade node play as part of the control plane upgrade.

Info: this will need a cluster where a master is a regular node as well. Not sure if this is currently covered be the CI.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -229,3 +229,56 @@
   tasks:
   - include: docker/upgrade.yml
     when: l_docker_upgrade is defined and l_docker_upgrade | bool and not openshift.common.is_atomic | bool
+
+- name: Drain and upgrade master nodes
+  hosts: oo_masters_to_config:&oo_nodes_to_upgrade
+  # This var must be set with -e on invocation, as it is not a per-host inventory var
+  # and is evaluated early. Values such as "20%" can also be used.
+  serial: "{{ openshift_upgrade_nodes_serial | default(1) }}"
+  any_errors_fatal: true
+
+  pre_tasks:
+  # TODO: To better handle re-trying failed upgrades, it would be nice to check if the node
+  # or docker actually needs an upgrade before proceeding. Perhaps best to save this until
+  # we merge upgrade functionality into the base roles and a normal config.yml playbook run.
+  - name: Determine if node is currently scheduleable
+    command: >
+      {{ hostvars[groups.oo_first_master.0].openshift.common.client_binary }} get node {{ openshift.node.nodename | lower }} -o json
+    register: node_output
+    delegate_to: "{{ groups.oo_first_master.0 }}"
+    changed_when: false
+
+  - set_fact:
+      was_schedulable: "{{ 'unschedulable' not in (node_output.stdout | from_json).spec }}"
+
+  - name: Mark node unschedulable
+    command: >
+      {{ hostvars[groups.oo_first_master.0].openshift.common.client_binary }} adm manage-node {{ openshift.node.nodename | lower }} --schedulable=false
+    delegate_to: "{{ groups.oo_first_master.0 }}"
+    # NOTE: There is a transient "object has been modified" error here, allow a couple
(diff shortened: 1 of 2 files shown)


ingvagabund:
aos-ci-test

dgoodwin:
No CI coverage exists today afaik.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -229,3 +229,56 @@
   tasks:
   - include: docker/upgrade.yml
     when: l_docker_upgrade is defined and l_docker_upgrade | bool and not openshift.common.is_atomic | bool
+
+- name: Drain and upgrade master nodes
+  hosts: oo_masters_to_config:&oo_nodes_to_upgrade
```
dgoodwin: This may not be necessary, masters are implicitly in the nodes group today, I don't think there's any way around that, so it might be best to keep this just at the masters group. IMO the serial and label vars should probably *not* be used during master upgrade, control plane should remain serial 1 host at a time.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -229,3 +229,56 @@
   tasks:
   - include: docker/upgrade.yml
     when: l_docker_upgrade is defined and l_docker_upgrade | bool and not openshift.common.is_atomic | bool
+
+- name: Drain and upgrade master nodes
```
dgoodwin: Is it possible to integrate this directly into the main play for masters above where we go through them 1 by 1 in serial? This is where the ELB hooks are and after discussions with ops I think we should aim to fully process each master in that loop, as much as possible, so when that system goes back into the ELB it's been fully updated.

As an example if this were a containerized cluster and we were using hooks to put into / out of an ELB, we would be back in the ELB but then later restart docker and bring down our services again.

ingvagabund: I would like to do that. There are still a minor issues with hooks. The `openshift_node_upgrade` restarts docker which given the current `openshift_master_upgrade_hook` position needs to be done after the hook is run. E.g. we can duplicate the hook in the `openshift_node_upgrade` role and run it only if the node is the master as well and run the original hook in the serial master upgrade loop only when the master is not a node. It looks like a hack but given the `openshift_node_upgrade` role is atomic from the point of view of the serial master upgrade loop, I don't see any other way around.

At the same time it would make again a sense to promote the master upgrade loop content into a role (or most of it).

Given the discussion we had today I will keep the PR as it is and move the upgrade node into the master upgrade play in another PR.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
ingvagabund: I would keep it there to stress we are upgrading masters that are nodes as well. Which label variable do you have in mind? The serial should be set to 1 always. I can do that in this PR or make it happen in another PR once I move the upgrade node play into the master upgrade loop.

dgoodwin: oo_nodes_to_upgrade is calculated by processing openshift_upgrade_nodes_serial, and openshift_upgrade_nodes_label, both I think would be an error to apply here as you should never be able to trigger this for more than one host at a time in parallel, or skip some masters due to labels. IMO it would be a bug to process these here. 

You could however use the node group which does *not* apply those variables, which is I believe oo_nodes_to_config, that would give you the clarity but not include any processing of the label var. (and go to serial: 1 as well)

ingvagabund:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
ingvagabund: That will not help. Otherwise there is a way to postpone the restart phase off both master and node upgrades into a single point in time. Optionally, we can skip the restart of openswitch, docker and node service. But then, we would have to do that in the master upgrade play.

PR #3249, opened 2017-02-02
Adding oc_label to lib_openshift.

kwoodson:


roles/lib_openshift/src/doc/label
```diff
@@ -0,0 +1,64 @@
...
+DOCUMENTATION = '''
+---
+module: oc_label
+short_description: Create, modify, and idempotently manage openshift labels.
+description:
+  - Modify openshift labels programmatically.
+options:
+  state:
+    description:
+    - State represents whether to create, modify, delete, or list
```
tbielawa: > State represents whether to create, modify, delete, or list

I see the choices below are ` ["present", "absent", "list", "add"]` so then:

* `create` maps to `add`
* `delete` maps to `absent`
* `list` maps to `list` 
* `modify` maps to `present`?


Maybe like this?

```yaml
    description:
    - State controls what to do with the module
    - An new label is added with the `create` state
    - An existing label is removed (or verified as removed) with the `absent` state
    - Applied labels may be enumerated with the `list` state
    - A specific labels existence is determined with the `present` state
```

But I'm not sure which of those states accounts for modifying objects

roles/lib_openshift/src/doc/label
```diff
@@ -0,0 +1,64 @@
...
+    required: false
+    default: /etc/origin/master/admin.kubeconfig
+    aliases: []
+  debug:
+    description:
+    - Turn on debug output.
+    required: false
+    default: False
+    aliases: []
+  kind:
```
tbielawa: The `AnsibleModule` spec for `kind` defines it as such:

```python
    kind=dict(
        default=None,
        type='str',
        required=True,
        choices=['node', 'pod', 'namespace']
),
```

However the `DOCUMENTATION` does not match

Param |`AnsibleModule` spec | `DOCUMENTATION`
------------ | ------------ | -------------
default | `None` | **node**
required | `True` | ⁿ/ₐ

roles/lib_openshift/src/doc/label
```diff
@@ -0,0 +1,64 @@
...
+    - The kind of object that can be managed.
+    default: node
+    choices:
+    - node
+    - pod
+    - namespace
+    aliases: []
+  labels:
+    description:
+    - A list of labels to for the resource.
```
tbielawa: A list of labels to apply to the resource. Each item in the list is a hash (ex: `region: east`)

roles/lib_openshift/src/doc/label
```diff
@@ -0,0 +1,64 @@
...
+EXAMPLES = '''
+- name: Add label to node
+  oc_label:
+    name: ip-172-31-5-23.ec2.internal
+    state: add
+    kind: node
+    labels:
+      - key: logging-infra-fluentd
+        value: 'true'
+'''
```
tbielawa: Some more examples expanding on those `state` choices would be very useful here.

roles/lib_openshift/src/test/integration/filter_plugins/filters.py
```diff
@@ -0,0 +1,26 @@
...
+'''
+Custom filters for use in testing
+'''
+
+
+class FilterModule(object):
+    ''' Custom ansible filters '''
+
+    @staticmethod
+    def label_dict_to_key_value_list(label_dict):
```
tbielawa: This seems to only be used by the integration tests. Should be add a disclaimer or other annotation in here to indicate this to possible contributors?

roles/lib_openshift/src/doc/label
kwoodson: Will fix.

roles/lib_openshift/src/doc/label
kwoodson: I will give an example.  each has key: <key>, value: <value> In order to key off specific keys or values.

roles/lib_openshift/src/test/integration/filter_plugins/filters.py
kwoodson: Sure, that sounds good.

roles/lib_openshift/src/doc/label
kwoodson: @tbielawa,

This is a good observation.  The 'present' state in declarative means create or update.  Think crud.
create|update = present
delete = absent
read = list
Basically, make-it-so and the 'so' being the state I'm passing in.  For the labeling module, make the labels present on the object.  Note that we did go outside the general pattern here and included an 'add'.  This just adds a label if needed.  That way you don't have to pass _ALL_ labels each time.  
If they already exist, exit.  
If they don't exist, create them.
If they aren't exact, update them.

I understand that the ansible terms for these can be less than ideal but its the same for the base modules.  Take a look at the user module's state:
absent|present - Whether the account should exist or not, taking action if the state is different from what is stated.

Hopefully this clears it up.  I'd be happy to discuss further.

roles/lib_openshift/src/doc/label
kwoodson: fixed to align with doc.  defaults to 'node'

kwoodson:
@tbielawa,  Please approve when you feel its ready.

tbielawa:
aos-ci-test

PR #3251, opened 2017-02-02
Make use of AnsibleDumper in openshift_master filters s.t. we can rep…

sdodson:
…resent AnsibleUnsafeText when dumping yaml.


sdodson:
aos-ci-test

PR #3254, opened 2017-02-02
Temporary hack to skip router/registry upgrade.

dgoodwin:
This is just a stop-gap until we can sync a few minor things with ops, and then flip them over to use this process.

TODOs are covered in https://trello.com/c/U3UsHcNO/348-integrate-registry-router-upgrade-with-ops


sdodson:
@kwoodson @mwoodson This is our temporary solution to address the fact that we currently try to do router and registry upgrades during control plane upgrades. What we're suggesting for now is that ops disables this functionality in openshift-ansible and continues to use their own router/registry upgrade procedures. We'll try to get this re-integrated ASAP though so this is just a stop gap.

Issue #3111, opened 2017-01-18
Failures with ansible 2.2.1.0


sdodson:
We're seeing yaml rendering issues with our code and 2.2.1.0. Need to sort out if we need to fix something or if ansible will be fixing things.

Upstream references https://github.com/ansible/ansible/issues/20290 https://github.com/ansible/ansible/issues/20253

```
    TASK [openshift_master : set_fact] *********************************************
    task path: /home/rdu/sdodson/git/Openshift/openshift-ansible/roles/openshift_master/tasks/main.yml:146
    An exception occurred during task execution. The full traceback is:
    Traceback (most recent call last):
      File "/usr/lib/python2.7/site-packages/ansible/executor/task_executor.py", line 126, in run
        res = self._execute()
      File "/usr/lib/python2.7/site-packages/ansible/executor/task_executor.py", line 443, in _execute
        self._task.post_validate(templar=templar)
      File "/usr/lib/python2.7/site-packages/ansible/playbook/task.py", line 248, in post_validate
        super(Task, self).post_validate(templar)
      File "/usr/lib/python2.7/site-packages/ansible/playbook/base.py", line 373, in post_validate
        value = templar.template(getattr(self, name))
      File "/usr/lib/python2.7/site-packages/ansible/template/__init__.py", line 427, in template
        disable_lookups=disable_lookups,
      File "/usr/lib/python2.7/site-packages/ansible/template/__init__.py", line 383, in template
        disable_lookups=disable_lookups,
      File "/usr/lib/python2.7/site-packages/ansible/template/__init__.py", line 583, in do_template
        res = j2_concat(rf)
      File "<template>", line 9, in root
      File "/home/rdu/sdodson/git/Openshift/openshift-ansible/filter_plugins/openshift_master.py", line 488, in translate_idps
        return yaml.safe_dump([idp.to_dict() for idp in idp_list], default_flow_style=False)
      File "/usr/lib64/python2.7/site-packages/yaml/__init__.py", line 218, in safe_dump
        return dump_all([data], stream, Dumper=SafeDumper, **kwds)
      File "/usr/lib64/python2.7/site-packages/yaml/__init__.py", line 190, in dump_all
        dumper.represent(data)
      File "/usr/lib64/python2.7/site-packages/yaml/representer.py", line 28, in represent
        node = self.represent_data(data)
      File "/usr/lib64/python2.7/site-packages/yaml/representer.py", line 57, in represent_data
[...]

PR #3255, opened 2017-02-02
Add logic to verify patched version of Ansible

mtnbikenc:
A patched version of Ansible is being made available, which should not cause a failure on version checking.

### DIFF
--- a/callback_plugins/aa_version_requirement.py
+++ b/callback_plugins/aa_version_requirement.py
@@ -7,6 +7,7 @@ The plugin is named with leading `aa_` to ensure this plugin is loaded
 first (alphanumerically) by Ansible.
 """
 import sys
+from subprocess import check_output
 from ansible import __version__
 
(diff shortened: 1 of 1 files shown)


sdodson:
https://github.com/ansible/ansible/commit/e85f3f44610fa454c80da6b0f5ae84e3887a6955 is the commit we need, could we just try to import that and if that fails return an error?

detiber:
@sdodson I think the problem with checking for specific implementation details is that they are subject to change with future versions. As terrible as specifying known-bad versions is, I think it will be easier to maintain reliably.

Hopefully we will get a point where we are automating testing against multiple ansible versions including the latest development versions from the Ansible repo to catch these issues before they are an issue :)

mtnbikenc:
Related to #3111

sdodson:
@detiber good point, i agree

mtnbikenc:
aos-ci-test

PR #3257, opened 2017-02-02
fixes 1417261. Points playbooks to the correct 3.5 roles for logging …

jcantrill:
fixes 1417261. Points playbooks to the correct 3.5 roles for logging and metrics

@sdodson This addresses the initial bug I believe but Im having issues testing since the ansible 2.2.2.0 version dependency check.  Also I'm not sure if I should be additionally updating the 'hosted_storage' variable.  I will defer to you recommendation

### DIFF
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -468,18 +468,18 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # pods are deleted
 #
 # Other Logging Options -- Common items you may wish to reconfigure, for the complete
-# list of options please see roles/openshift_hosted_logging/README.md
+# list of options please see roles/openshift_logging/README.md
 #
 # Configure loggingPublicURL in the master config for aggregate logging, defaults
 # to https://kibana.{{ openshift_master_default_subdomain }}
 #openshift_master_logging_public_url=https://kibana.example.com
 # Configure the number of elastic search nodes, unless you're using dynamic provisioning
 # this value must be 1
-#openshift_hosted_logging_elasticsearch_cluster_size=1
-#openshift_hosted_logging_hostname=logging.apps.example.com
+#openshift_logging_es_cluster_size=1
+#openshift_logging_kibana_hostname=logging.apps.example.com
 # Configure the prefix and version for the deployer image
-#openshift_hosted_logging_deployer_prefix=registry.example.com:8888/openshift3/
-#openshift_hosted_logging_deployer_version=3.3.0
+#openshift_logging_image_prefix=registry.example.com:8888/openshift3/
+#openshift_logging_image_version=3.3.0
 
 # Configure the multi-tenant SDN plugin (default is 'redhat/openshift-ovs-subnet')
 # os_sdn_network_plugin_name='redhat/openshift-ovs-multitenant'
--- a/playbooks/adhoc/openshift_hosted_logging_efk.yaml
+++ b/playbooks/adhoc/openshift_hosted_logging_efk.yaml
@@ -1,7 +1,7 @@
 ---
 - hosts: masters[0]
   roles:
-  - role: openshift_hosted_logging
+  - role: openshift_logging
     openshift_hosted_logging_cleanup: no
 
 - name: Update master-config for publicLoggingURL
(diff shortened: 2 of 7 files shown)


playbooks/common/openshift-cluster/openshift_hosted.yml
```diff
@@ -27,21 +27,21 @@
       logging_elasticsearch_ops_cluster_size: "{{ openshift_hosted_logging_elasticsearch_ops_cluster_size | default(1) }}"
   roles:
   - role: openshift_hosted
-  - role: openshift_hosted_metrics
+  - role: openshift_metrics
```
ewolinetz: i dont think you meant to modify this file yet, not until QE has certified both new roles

roles/openshift_logging/defaults/main.yml
```diff
@@ -3,7 +3,7 @@ openshift_logging_image_prefix: docker.io/openshift/origin-
 openshift_logging_image_version: latest
 openshift_logging_use_ops: False
 master_url: "https://kubernetes.default.svc.{{ openshift.common.dns_domain }}"
-public_master_url: "https://{{openshift.common.public_hostname}}:8443"
+openshift_logging_master_public_url: "https://{{openshift.common.public_hostname}}:8443"
```
ewolinetz: can you make the same change for master_url while you're in here?

ewolinetz: edit - do we expect users to change this value given that we're pointing it to the public_hostname gathered by facts?

roles/openshift_logging/README.md
```diff
@@ -25,7 +25,7 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_image_version`: The image version for the logging images to use. Defaults to 'latest'.
 - `openshift_logging_use_ops`: If 'True', set up a second ES and Kibana cluster for infrastructure logs. Defaults to 'False'.
 - `master_url`: The URL for the Kubernetes master, this does not need to be public facing but should be accessible from within the cluster. Defaults to 'https://kubernetes.default.svc.cluster.local'.
-- `public_master_url`: The public facing URL for the Kubernetes master, this is used for Authentication redirection. Defaults to 'https://localhost:8443'.
+- `openshift_logging_master_public_url`: The public facing URL for the Kubernetes master, this is used for Authentication redirection. Defaults to 'https://localhost:8443'.
```
ewolinetz: Do we want users to edit this since we are pulling the value (as a default) from openshift_facts?

playbooks/adhoc/openshift_hosted_logging_efk.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 - hosts: masters[0]
   roles:
-  - role: openshift_hosted_logging
+  - role: openshift_logging
```
ewolinetz: i dont think you meant to modify this file yet, not until QE has certified both new roles

inventory/byo/hosts.origin.example
```diff
@@ -468,18 +468,18 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
...
 #
 # Configure loggingPublicURL in the master config for aggregate logging, defaults
 # to https://kibana.{{ openshift_master_default_subdomain }}
 #openshift_master_logging_public_url=https://kibana.example.com
 # Configure the number of elastic search nodes, unless you're using dynamic provisioning
 # this value must be 1
-#openshift_hosted_logging_elasticsearch_cluster_size=1
-#openshift_hosted_logging_hostname=logging.apps.example.com
+#openshift_logging_es_cluster_size=1
+#openshift_logging_kibana_hostname=logging.apps.example.com
```
ewolinetz: i dont think you meant to modify this file yet, not until QE has certified both new roles

playbooks/common/openshift-cluster/openshift_hosted.yml
jcantrill: Actually, I did intend to modify this entry because of the logging issue; I assume there would be the same problem with metrics.  The user reported an issue installing logging on 3.5 using these playbooks, which point to the wrong playbooks.

roles/openshift_logging/defaults/main.yml
jcantrill: I was wondering about this value along with 'master_url'.  I would expect users to never change these values unless they are deploing outside of the cluster.  That scenario seems like it would be unsupported and users are left to their own devices to make it work.  Regarding master_url, is there any reason we should not make it 'private' and put it int vars?

roles/openshift_logging/README.md
jcantrill: I'm onboard with removing it as a 'public' value if we are in agreement

playbooks/adhoc/openshift_hosted_logging_efk.yaml
jcantrill: The intention is correct, though again, I'm not sure why we have so many playbooks for logging

jcantrill:
aos-ci-test

PR #3259, opened 2017-02-03
Ensure python-ruamel-yaml is installed

sdodson:
need to revisit this, currently python2-ruamel-yaml doesn't provide
'python-ruamel-yaml'. It probably should and that should be what we're
installing.

Fixes Bug 1418911

### DIFF
--- a/roles/openshift_logging/tasks/main.yaml
+++ b/roles/openshift_logging/tasks/main.yaml
@@ -12,10 +12,6 @@
 
 - debug: msg="Created temp dir {{mktemp.stdout}}"
 
-- name: Ensuring ruamel.yaml package is on target
-  command: yum install -y ruamel.yaml
-  check_mode: no
-
 - name: Copy the admin client config(s)
   command: >
     cp {{ openshift_master_config_dir }}/admin.kubeconfig {{ mktemp.stdout }}/admin.kubeconfig
--- a/openshift-ansible.spec
+++ b/openshift-ansible.spec
@@ -20,6 +20,7 @@ Requires:      tar
 Requires:      openshift-ansible-docs = %{version}-%{release}
 Requires:      java-1.8.0-openjdk-headless
 Requires:      httpd-tools
+Requires:      python-ruamel-yaml
 
 %description
 Openshift and Atomic Enterprise Ansible
--- a/roles/lib_openshift/tasks/main.yml
+++ b/roles/lib_openshift/tasks/main.yml
@@ -1,0 +1,5 @@
+---
+- name: lib_openshift ensure python-ruamel-yaml package is on target
+  package:
+    name: python-ruamel-yaml
+    state: present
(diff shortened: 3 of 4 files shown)


roles/openshift_logging/tasks/main.yaml
```diff
@@ -12,8 +12,8 @@
 
 - debug: msg="Created temp dir {{mktemp.stdout}}"
 
-- name: Ensuring ruamel.yaml package is on target
-  command: yum install -y ruamel.yaml
+- name: Ensuring python-ruamel.yaml package is on target
```
jcantrill: should this also say 'python2'

sdodson: Yeah, fixt.

roles/openshift_logging/tasks/main.yaml
```diff
@@ -12,9 +12,10 @@
 
 - debug: msg="Created temp dir {{mktemp.stdout}}"
 
-- name: Ensuring ruamel.yaml package is on target
-  command: yum install -y ruamel.yaml
-  check_mode: no
+- name: Ensuring python2-ruamel-yaml package is on target
+  package:
```
detiber: Is the purpose of this task to provide the dependency for the module? If so, it seems like we should add this task to lib_openshift or perhaps find another way to ensure the package is present rather than requiring all roles that use the modules to ensure the package is installed explicitly.

sdodson: I was hoping to write up a quick pre-requisites playbook and move this along with PyYAML and other openshift_facts dependencies there.

sdodson: I guess we should move it to lib_openshift now though. updating.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

warmchang:
Hi @sdodson , I got the error about "python-ruamel-yaml" when upgrade the origin version.
How to solve this problem? Thanks!

> TASK [lib_openshift : lib_openshift ensure python-ruamel-yaml package is on target] ***
> fatal: [master.example.com]: FAILED! => {
>     "changed": false,
>     "failed": true,
>     "rc": 126,
>     "results": [
>         "No package matching 'python-ruamel-yaml' found available, installed or updated"
>     ]
> }
> 
> MSG:
> 
> No package matching 'python-ruamel-yaml' found available, installed or updated
> 
> PLAY RECAP *********************************************************************
> localhost                  : ok=30   changed=0    unreachable=0    failed=0
> master.example.com         : ok=293  changed=21   unreachable=0    failed=1


The host is CentOS 7.3:

> [root@master openshift-ansible]# cat /etc/redhat-release
> CentOS Linux release 7.3.1611 (Core)
> [root@master openshift-ansible]# origin version
> origin v1.3.1
> kubernetes v1.3.0+52492b4
> etcd 2.3.0+git

warmchang:
mark. https://github.com/openshift/openshift-ansible/issues/3302

PR #3261, opened 2017-02-03
Manage the excluder functionality

sdodson:
Add openshift_excluder role with install, status, exclude, and unexclude tasks.
In config and upgrade playbooks, record status of excluder, disable it, re-enable it if it was previously enabled

### DIFF
--- a/playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
+++ b/playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
@@ -46,6 +46,10 @@
   tags:
   - pre_upgrade
 
+- include: ../../../../common/openshift-cluster/disable_excluder.yml
+  tags:
+  - pre_upgrade
+
 - include: ../../../../common/openshift-cluster/initialize_openshift_version.yml
   tags:
   - pre_upgrade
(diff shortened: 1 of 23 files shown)


sdodson:
aos-ci-test

sdodson:
@mtnbikenc ptal, biggest thing is I need to make sure I'm hooking into all of the right places in upgrade. I think config is pretty straight forward.

sdodson:
aos-ci-test

sdodson:
> Also, I'm not certain about the structure of the openshift_excluder role not having a default entrypoint of tasks/main.yml. I'm not seeing this patter in use so i think further discussion is warranted.

I've added a main.yml that just invokes status, but yeah, definitely open to changes there.

sdodson:
@mtnbikenc Move excluder disablement into control plane and node upgrade sections. Have some additional thoughts on general refactoring, ideally we could move everything that's not byo specific into the common playbooks but we can do that later. I've left the disablement just before the steps that figure out what version is installed and available.

sdodson:
aos-ci-test

sdodson:
Going to go ahead and merge this so we can get QE testing it.

PR #3271, opened 2017-02-06
fix 1406057. Allow openshift_metrics nodeselectors for components

### LINKED ISSUE #2800: [RFE] openshift_hosted_{metrics, logging} roles should support `--host-mount` option for storage
Many customers do not have Persistent Storage options outside of NFS with which they can create PVs, which is not ideal for things like metrics and logging, where we tend to recommend block storage with high i/o for proper performance. While we support a fair number of block storage options, the majority of them are cloud-deployment oriented, which currently leaves non-cloud on-premise customers with two options: Fibre Channle or iSCSI. In the event that neither of those are available, we end up recommending attaching extra disks/LUNs to the nodes that will run these services, using the `--host-mount` option to attach them. We would like to see this option supported in future versions of the installer.

As I see it the needed additions would be the following:

1. Support for a `host-mount` type for var `openshift_hosted_X_storage_kind`.
This would need to result in giving the relevant ServiceAccount access to the `priveleged` SCC to be able to mount a directory on the host, and a patch to the dc to claim the SCC privileges. i.e.:
```
oadm policy add-scc-to-user privileged  \
       system:serviceaccount:logging:aggregated-logging-elasticsearch
oc patch $dc \
       -p '{"spec":{"template":{"spec":{"containers":[{"name":"elasticsearch","securityContext":{"privileged": true}}]}}}}'
```
2. instead of creating a new PV, we would feed the value of 
`openshift_hosted_X_storage_volume_name` to `--mount-host` during deployment, i.e.:
```
oc set volume $dc \
          --add --overwrite --name=elasticsearch-storage \
          --type=hostPath --path={{ openshift_hosted_X_storage_volume_name }}
    oc deploy --latest $dc
    oc scale $dc --replicas=1
```
3. Finally (if not already supported), we need to be able to pass in a node-selector for logging and metrics, as we can currently for the router/registry, i.e. `openshift_hosted_{metrics,logging}_selector`

##### Version
OCP 3.3+

##### Additional Information
Additional docs: https://docs.openshift.com/container-platform/3.3/install_config/aggregate_logging.html#aggregated-elasticsearch

@sdodson @jkupferer @sabre1041 @mwitzenm @themoosman

jcantrill:
This would also partially fix #2800. In regards to https://github.com/openshift/openshift-ansible/issues/2800#issuecomment-268399097  I believe this is resolvable using  affinity/anti-affinity rules.

### DIFF
--- a/roles/openshift_metrics/defaults/main.yaml
+++ b/roles/openshift_metrics/defaults/main.yaml
@@ -13,6 +13,7 @@ openshift_metrics_hawkular_requests_cpu: null
 openshift_metrics_hawkular_cert: ""
 openshift_metrics_hawkular_key: ""
 openshift_metrics_hawkular_ca: ""
+openshift_metrics_hawkular_nodeselector: ""
 
 openshift_metrics_cassandra_replicas: 1
 openshift_metrics_cassandra_storage_type: emptydir
@@ -21,12 +22,14 @@ openshift_metrics_cassandra_limits_memory: 2G
 openshift_metrics_cassandra_limits_cpu: null
 openshift_metrics_cassandra_requests_memory: 1G
 openshift_metrics_cassandra_requests_cpu: null
+openshift_metrics_cassandra_nodeselector: ""
 
 openshift_metrics_heapster_standalone: False
 openshift_metrics_heapster_limits_memory: 3.75G
 openshift_metrics_heapster_limits_cpu: null
 openshift_metrics_heapster_requests_memory: 0.9375G
 openshift_metrics_heapster_requests_cpu: null
+openshift_metrics_heapster_nodeselector: ""
 
 openshift_metrics_duration: 7
 openshift_metrics_resolution: 15s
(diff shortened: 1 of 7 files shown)


roles/openshift_metrics/defaults/main.yaml
```diff
@@ -13,6 +13,7 @@ openshift_metrics_hawkular_requests_cpu: null
 openshift_metrics_hawkular_cert: ""
 openshift_metrics_hawkular_key: ""
 openshift_metrics_hawkular_ca: ""
+openshift_metrics_hawkular_nodeselector: {'metrics-infra-hawkular': 'true'}
```
sdodson: Seems like you should default to "" ?

sdodson:
@mrobson thoughts on the defaults chosen here? We had a fair amount of debate when we originally decided not to assign node selectors out of the box so I imagine we should default to no selector here.

mrobson:
@sdodson Ya, maintaining the current empty default selector makes sense.  Labels and the subsequent nodeSelector settings become a personal choice that will vary per person and per install.

jcantrill:
@sdodson updated defaults to emtpy

jcantrill:
aos-ci-test

PR #3272, opened 2017-02-06
Use service annotations to redeploy router service serving cert signer cert

abutcher:
During certificate redeployment, remove existing `router-certs` secret and replace the `service.alpha.openshift.io/serving-cert-secret-name` service annotation to trigger creation of a new router certificate.


abutcher:
@sdodson It's limited to the case where the `router-certs` secret exists.

abutcher:
and service serving cert signer was added in 3.3

PR #3274, opened 2017-02-07
[RFC][DONTMERGE] scheduler_config: predicates and priorities are lists

giuseppe:
origin:latest fails when I don't use this patch, though it breaks older versions.

### DIFF
--- a/roles/openshift_master/vars/main.yml
+++ b/roles/openshift_master/vars/main.yml
@@ -7,10 +7,12 @@ openshift_master_policy: "{{ openshift_master_config_dir }}/policy.json"
 scheduler_config:
   kind: Policy
   apiVersion: v1
-  predicates: "{{ openshift_master_scheduler_predicates
+  predicates:
+    - "{{ openshift_master_scheduler_predicates
                   | default(openshift_master_scheduler_current_predicates
                             | default(openshift_master_scheduler_default_predicates)) }}"
-  priorities: "{{ openshift_master_scheduler_priorities
+  priorities:
+    - "{{ openshift_master_scheduler_priorities
                   | default(openshift_master_scheduler_current_priorities
                             | default(openshift_master_scheduler_default_priorities)) }}"
 


sdodson:
@mtnbikenc has recently worked in this area too

mtnbikenc:
@giuseppe This change creates a list within a list.
```
    "predicates": [
        [
            {
                "name": "NoDiskConflict"
            },
            {
                "name": "NoVolumeZoneConflict"
            },
            {
                "name": "MaxEBSVolumeCount"
            },
...
```
Can you reference the change in origin latest where this is required?  The example I could shows a list with items.  https://github.com/openshift/origin/blob/85eb37b34f0657631592356d020cef5a58470f8e/vendor/k8s.io/kubernetes/examples/scheduler-policy-config.json

roles/openshift_master/vars/main.yml
```diff
@@ -7,10 +7,12 @@ openshift_master_policy: "{{ openshift_master_config_dir }}/policy.json"
 scheduler_config:
   kind: Policy
   apiVersion: v1
-  predicates: "{{ openshift_master_scheduler_predicates
+  predicates:
+    - "{{ openshift_master_scheduler_predicates
```
detiber: These variables should already be lists, is there a case you are seeing where that is not the case?

giuseppe: without this change, I get master failing with:
```
Feb 07 16:06:53 fedora25 origin-master[6527]: F0207 15:06:53.750381    6536 master.go:276] Unable to start scheduler: invalid scheduler configuration: only encoded map or array can be decoded into a struct
```

and the `scheduler.json` generated is:

```
{
    "apiVersion": "v1", 
    "kind": "Policy", 
    "predicates": {
        "argument": {
            "serviceAffinity": {
                "labels": [
                    "region"
                ]
            }
        }, 
        "name": "Region"
    }, 
    "priorities": {
        "argument": {
            "serviceAntiAffinity": {
                "label": "zone"
            }
        }, 
        "name": "Zone", 
        "weight": 2
    }
}
```

instead, with the patch applied, I get:
```
{
    "apiVersion": "v1", 
    "kind": "Policy", 
    "predicates": [
        {
            "argument": {
                "serviceAffinity": {
                    "labels": [
                        "region"
                    ]
                }
            }, 
            "name": "Region"
        }
    ], 
    "priorities": [
        {
            "argument": {
                "serviceAntiAffinity": {
                    "label": "zone"
                }
            }, 
            "name": "Zone", 
            "weight": 2
        }
    ]
}
```

giuseppe:
I am really not sure why/how this is happening, I am on Fedora 25 and provisioning a F25 VM.  If you have never seen this issue before, then I'll try to dig more and see what is the cause

detiber:
@giuseppe definitely seems like there is a bug, can you create an issue with the details?

giuseppe:
so the reason why this is happening is that `lookup` converts a list with a single element to the single element itself, I reproduced it with:

```
- hosts:
   - localhost
  gather_facts: False
  connection: local
  vars:
    contents: "{{ lookup('foo') }}"
  tasks:
    - set_fact:
        foo: "{{ contents }}"
    - debug: msg="{{ contents }}"
```

and foo.py:
```
from ansible.errors import AnsibleError, AnsibleParserError
from ansible.plugins.lookup import LookupBase

try:
    from __main__ import display
except ImportError:
    from ansible.utils.display import Display
    display = Display()


class LookupModule(LookupBase):

    def run(self, terms, variables=None, **kwargs):
        return ['1']
```

Looking more into this though, I think we should use for '-latest' the same priorities and predicates used for the last version.  I have pushed two patches to do this.

PR #3281, opened 2017-02-07
Reverting commit 3257 and renaming master_url to openshift_logging_ma…

ewolinetz:
…ster_url

I believe the correct approach would be to introduce a way to run either the openshift_hosted_{logging,metrics} roles or the new openshift_{logging,metrics} roles instead of just replacing them outright

### DIFF
--- a/playbooks/common/openshift-cluster/openshift_hosted.yml
+++ b/playbooks/common/openshift-cluster/openshift_hosted.yml
@@ -27,21 +27,21 @@
       logging_elasticsearch_ops_cluster_size: "{{ openshift_hosted_logging_elasticsearch_ops_cluster_size | default(1) }}"
   roles:
   - role: openshift_hosted
-  - role: openshift_metrics
+  - role: openshift_hosted_metrics
     when: openshift_hosted_metrics_deploy | default(false) | bool
-  - role: openshift_logging
+  - role: openshift_hosted_logging
     when: openshift_hosted_logging_deploy | default(false) | bool
-    openshift_logging_kibana_hostname: "{{ logging_hostname }}"
-    openshift_logging_kibana_ops_hostname: "{{ logging_ops_hostname }}"
-    openshift_logging_master_public_url: "{{ logging_master_public_url }}"
-    openshift_logging_es_cluster_size: "{{ logging_elasticsearch_cluster_size }}"
-    openshift_logging_es_pvc_dynamic: "{{ 'true' if openshift_hosted_logging_storage_kind | default(none) == 'dynamic' else '' }}"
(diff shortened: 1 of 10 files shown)


ewolinetz:
Reverts https://github.com/openshift/openshift-ansible/pull/3257

ewolinetz:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1419838 and https://bugzilla.redhat.com/show_bug.cgi?id=1419874

playbooks/common/openshift-cluster/openshift_hosted.yml
```diff
@@ -27,21 +27,21 @@
       logging_elasticsearch_ops_cluster_size: "{{ openshift_hosted_logging_elasticsearch_ops_cluster_size | default(1) }}"
   roles:
   - role: openshift_hosted
-  - role: openshift_metrics
+  - role: openshift_hosted_metrics
```
sdodson: This too? or errant search replace?

playbooks/common/openshift-cluster/openshift_hosted.yml
```diff
@@ -27,21 +27,21 @@
...
     when: openshift_hosted_metrics_deploy | default(false) | bool
-  - role: openshift_logging
+  - role: openshift_hosted_logging
```
sdodson: I think this too should stay the same, right?

playbooks/common/openshift-cluster/openshift_hosted.yml
ewolinetz: This is an intended change.

playbooks/common/openshift-cluster/openshift_hosted.yml
ewolinetz: This is intentional as well.

ewolinetz:
per the discussion we just had we'll be updating the defaults for the new roles instead

ewolinetz:
@sdodson @jcantrill are we okay with just matching on the variables that are used by our current playbooks and then in the example inventories point to the current role READMEs so that users can use the new variable names?

sdodson:
> @sdodson @jcantrill are we okay with just matching on the variables that are used by our current playbooks and then in the example inventories point to the current role READMEs so that users can use the new variable names?

You mean to provide the deprecation warning? Yeah that's fine for inside openshift-ansible, we should update the documentation to use the new variables and put in a release notes item.

ewolinetz:
> You mean to provide the deprecation warning? Yeah that's fine for inside openshift-ansible, we should update the documentation to use the new variables and put in a release notes item.

I mean like this [0] - or would want a dep warning as well? (do you want that added on this PR?)

[0] https://github.com/openshift/openshift-ansible/pull/3281/files#diff-1b06f2ffbea00508e375a7c7ec2d3051R478

sdodson:
That's fine as is, i don't think we need to do the dep warning, it'd likely just be ignored anyway.

PR #3284, opened 2017-02-08
fixes 1419839.  Install only heapster for openshift_metrics when heap…

jcantrill:
…ster standalone flag is set.

### DIFF
--- a/roles/openshift_metrics/tasks/generate_certificates.yaml
+++ b/roles/openshift_metrics/tasks/generate_certificates.yaml
@@ -1,18 +1,4 @@
 ---
-- name: create certificate output directory
-  file:
-    path: "{{ openshift_metrics_certs_dir }}"
-    state: directory
-    mode: 0700
-
-- name: list existing secrets
-  command: >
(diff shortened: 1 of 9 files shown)


jcantrill:
@mwringe Can you comment on the prevalence of installing only heapster?  Does this need to land 3.5 or can it wait for  a point release?

ewolinetz:
LGTM

mwringe:
> @mwringe Can you comment on the prevalence of installing only heapster? Does this need to land 3.5 or can it wait for a point release?

I don't think we even have this in OCP, only in Origin. Its used by users who only want the HPA and do not want the other metrics components. But some people are using this since they only want the HPA

jcantrill:
@sdodson based on https://github.com/openshift/openshift-ansible/pull/3284#issuecomment-278363223 we can wait.  My only reservation might be there are some org change that might make it more challanging to rebase but I suspect that is your reasons to prefer to defer.

sdodson:
:shipit:

smarterclayton:
Heck yeah!

PR #3285, opened 2017-02-08
Fixed ansible module unit and integration tests and added runners.

twiest:
Specifically:
* Added `generate-and-run-tests.sh` scripts in lib_utils and lib_openshift. These make it easy to run the ansible module unit and integration tests in those roles.
   * Note that the lib_openshift `generate-and-run-tests.sh` script also runs the `generate-and-run-tests.sh` from lib_utils. This ensures that lib_utils has been generated and is up to date as well as making it so we can run all tests with a single script.

* Fixed the `oadm_manage_node` itegration test and changed it to only require a master to run. It uses that master and picks the first node returned by oc_obj.

* Fixed the `oc_label` integration test. Also made it run without requiring a node.

* Fixed the `oc_service` integration test.

* Updated the `yedit` integration test to be in line with our new integration testing patterns.

### DIFF
--- a/roles/lib_openshift/src/test/integration/oadm_manage_node.yml
+++ b/roles/lib_openshift/src/test/integration/oadm_manage_node.yml
@@ -1,40 +1,51 @@
 #!/usr/bin/ansible-playbook --module-path=../../../library/
-# ./oadm_manage_node.yml -M ../../../library -e "cli_master_test=$OPENSHIFT_MASTER cli_node_test=$OPENSHIFT_NODE
+#
+# ./oadm_manage_node.yml -e "cli_master_test=$OPENSHIFT_MASTER
 ---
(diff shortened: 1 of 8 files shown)


roles/lib_openshift/src/test/integration/oadm_manage_node.yml
```diff
@@ -1,40 +1,51 @@
...
   tasks:
+  - name: get list of nodes
+    oc_obj:
+      state: list
+      kind: node
+    register: obj_out
+
+  - name: Set the node to work with
+    set_fact:
+      node_to_test: "{{ obj_out['results']['results'][0]['items'][0]['metadata']['name'] }}"
```
kwoodson: This is a better pattern.

PR #3287, opened 2017-02-08
latest: use same priorities and predicates used by the last version

giuseppe:


giuseppe:
aos-ci-test

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py
```diff
@@ -98,7 +98,7 @@ def run(self, terms, variables=None, regions_enabled=True, short_version=None,
                 {'name': 'MatchInterPodAffinity'}
             ])
 
-        if short_version in ['1.5', '1.6']:
+        if short_version in ['1.5', '1.6', 'latest']:
```
detiber: Instead of doing it this way, should we instead hardcode latest to a specific version in these lookup plugins? We already need to audit these for each release, so it seems like having to update the version associated with latest shouldn't be much more of a burden.

giuseppe: would you like if I refactor it to something like this?
```
diff --git a/roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py b/roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py
index d1a08fe..d1cb318 100644
--- a/roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py
+++ b/roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py
@@ -39,15 +39,23 @@ class LookupModule(LookupBase):
             else:
                 # pylint: disable=line-too-long
                 raise AnsibleError("Either OpenShift needs to be installed or openshift_release needs to be specified")
+
+        origin_versions = ['1.1', '1.2', '1.3', '1.4', '1.5', '1.6']
+        ose_versions = ['3.1', '3.2', '3.3', '3.4', '3.5', '3.6']
+
         if deployment_type == 'origin':
-            if short_version not in ['1.1', '1.2', '1.3', '1.4', '1.5', '1.6', 'latest']:
-                raise AnsibleError("Unknown short_version %s" % short_version)
+            versions = origin_versions
         elif deployment_type == 'openshift-enterprise':
-            if short_version not in ['3.1', '3.2', '3.3', '3.4', '3.5', '3.6', 'latest']:
-                raise AnsibleError("Unknown short_version %s" % short_version)
+            versions = ose_versions
         else:
             raise AnsibleError("Unknown deployment_type %s" % deployment_type)
 
+        if short_version == 'latest':
+            short_version = versions[-1]
+
+        if short_version not in versions:
+            raise AnsibleError("Unknown short_version %s" % short_version)
+
         if deployment_type == 'openshift-enterprise':
             # convert short_version to origin short_version
             short_version = re.sub('^3.', '1.', short_version)
@@ -98,7 +106,7 @@ class LookupModule(LookupBase):
                 {'name': 'MatchInterPodAffinity'}
             ])
 
-        if short_version in ['1.5', '1.6', 'latest']:
+        if short_version in ['1.5', '1.6']:
             predicates.extend([
                 {'name': 'NoVolumeZoneConflict'},
                 {'name': 'MaxEBSVolumeCount'},
```

giuseppe: hm.. probably not, I changed idea when I was going to commit it. I will just hardcode "latest" to the version to use.

giuseppe:
@detiber I've pushed two fixup patches

giuseppe:
aos-ci-test

giuseppe:
aos-ci-test

PR #3289, opened 2017-02-08
Modify playbooks to use oadm_manage_node module

mtnbikenc:


mtnbikenc:
Still need to test this on an upgrade but wanted to get the code out here for review.

cc: @sdodson @kwoodson

playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
```diff
@@ -23,11 +23,19 @@
...
     delegate_to: "{{ groups.oo_first_master.0 }}"
-    when: l_docker_upgrade is defined and l_docker_upgrade | bool and inventory_hostname in groups.oo_nodes_to_upgrade
+    retries: 10
+    delay: 5
+    register: node_unschedulable
+    until: node_unschedulable|succeeded
+    when:
+    - l_docker_upgrade is defined
+    - l_docker_upgrade | bool
+    - inventory_hostname in groups.oo_nodes_to_upgrade
```
ingvagabund: nice

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -59,12 +59,16 @@
...
 
 - name: Set node schedulability
-  command: >
-    {{ hostvars[openshift_master_host].openshift.common.client_binary }} adm manage-node {{ openshift.node.nodename }} --schedulable={{ 'true' if openshift.node.schedulable | bool else 'false' }}
-    --config={{ openshift_manage_node_kubeconfig }}
-    -n default
-  when: "'nodename' in openshift.node"
+  oadm_manage_node:
+    node: "{{ openshift.node.nodename | lower }}"
+    schedulable: "{{ 'true' if openshift.node.schedulable | bool else 'false' }}"
```
ingvagabund: maybe change to ``schedulable: True`` and put the condition into ``when``?

playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
mtnbikenc: This is a pattern from the styleguide which is being developed.

roles/openshift_manage_node/tasks/main.yml
mtnbikenc: I had considered that, however I was thinking this task should run either way to set the value as desired, not just skip if schedulable should be false.

detiber:
@sdodson do we have ruamel-yaml available in fedora/CentOS? If not, then we'd want to expedite that with the use of the modules coming in, otherwise we'll start getting issues related to the breakage.

sdodson:
It should be in Centos before 3.5 ships. It's not currently in the Paas SIG repos, I'll check on where we're at there.

ashcrow:
@detiber @sdodson No packages in Fedora https://admin.fedoraproject.org/pkgdb/packages/ruamel-yaml%2A/

sdodson:
It's this, right? https://admin.fedoraproject.org/pkgdb/package/rpms/python-ruamel-yaml/

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -238,29 +238,22 @@
   any_errors_fatal: true
 
   pre_tasks:
+  - name: Load lib_openshift modules
```
ashcrow: Just so I follow, this is done here since there is no ``meta/main.yml`` here correct?

ashcrow:
@sdodson ah, if so, then yes :laughing: 

Looks like it's been in testing for quite a while: https://bodhi.fedoraproject.org/updates/?packages=python-ruamel-yaml

sdodson:
https://bugzilla.redhat.com/show_bug.cgi?id=1386563 is the review request, asked for info on that. I was under the impression it had actually landed in currently supported Fedora releases.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
mtnbikenc: Right, we are not in a role so we don't have meta/main.yml.  Also, this has to be called in both pre_tasks and post_tasks to load the modules.

playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
```diff
@@ -23,11 +23,23 @@
   serial: 1
   any_errors_fatal: true
   tasks:
-  - name: Prepare for Node draining
-    command: >
-      {{ openshift.common.client_binary }} adm manage-node {{ openshift.node.nodename }} --schedulable=false
+  - name: Load lib_openshift modules
+    include_role:
+      name: lib_openshift
```
mtnbikenc: I guess in this case I can add a `roles:` section since that will be run before the tasks.  Will update.

playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
```diff
@@ -23,11 +23,23 @@
...
+
+  - name: Mark node unschedulable
+    oadm_manage_node:
+      node: "{{ openshift.node.nodename | lower }}"
+      schedulable: False
     delegate_to: "{{ groups.oo_first_master.0 }}"
-    when: l_docker_upgrade is defined and l_docker_upgrade | bool and inventory_hostname in groups.oo_nodes_to_upgrade
+    retries: 10
+    delay: 5
```
mtnbikenc: I was following the same pattern here as what I found in the openshift-ansible-ops playbooks. @kwoodson or @mwoodson, can you comment?

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -238,29 +238,22 @@
...
+    include_role:
+      name: lib_openshift
```
mtnbikenc: I agree on not liking the include_role usage here (and elsewhere).  Depending on the amount of work for implementation, it may be something to come back to (tracked on a card) than work on for this PR.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
detiber: I believe you should be able to use any modules defined in `roles` in post_tasks.

playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
mtnbikenc: @ashcrow and @ingvagabund, keep this in mind for your refactoring.  If you have a play, with only a tasks: section, you can add a roles: section to include lib_openshift.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
kwoodson: @mtnbikenc,
I'm a big fan of using include_role.  I disagree that it is code smell.  This is just a question of organization and when the calls to the module should run.  If you need to run something in pre_, then you can include_role in pre.  If the only place where you are running module calls is in post_tasks, then I'd probably place it in among other roles in the roles: section.

playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
kwoodson: @mwoodson, Please reply.  I'm not familiar other than I believe there might have been a bug.

mwoodson: Let me give a bit of history.  This is coming from the oadm command.  We found that at times, we would issue the  "oadm manage-node <nodename> --schedudable=false" (done through the module), but the command would fail.

We were in a rush to get this code working as it is the upgrade code we use on our customers.  This is a bug in openshift oadm command that we were compensating for.

The correct path is to file a bug upstream to get it fixed.  In the meantime, we put in the retry in to actually get the work done.

I agree with @detiber in that the module could be made more resilient, but the question would be what is the right thing for the module to do?  Do we retry 3 times? 5times?  i don't know the answer when you are trying to compensate for a bug.

kwoodson: @mtnbikenc,

https://bugzilla.redhat.com/show_bug.cgi?id=1279303

I think modules can be more resilient.  That is always the goal.  Then the discussion needs to be how long do we wait? How many times do we retry? This is what @mwoodson was mentioning. I'm definitely willing to discuss that as I think we could greatly benefit from guaranteeing state before proceeding.

detiber: @mtnbikenc I'll defer to you on the best approach to take here, but I think that if we leave the retries in then we need at least two followup items to track:
- A bug to fix the upstream client bug
- A card/bug to track moving the workaround to the module, since we will still need to work against released versions of the oc/oadm client.

roles/openshift_manage_node/tasks/main.yml
ingvagabund: I don't see a point of running the tasks unless schedulable is changing from false -> true. What is the point of changing false -> false?

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
ingvagabund: As long is the ``include_role`` does not introduces "hidden" dependencies, it is a useful directive to use.

playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
kwoodson: +1 for readability here.

playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
sdodson: The bug is fixed in the lastest 3.4 release and internal builds of 3.3 and 3.5. So it's safe to expect this to be fixed in all situations where openshift-ansible will use this code.

kwoodson:
@mtnbikenc,

I know that this is part of a refactor but I would like to step back and ask what this role buys us?  If we can make the actual module call to manage node instead of calling a role, I think we could save ourselves some tasks.  For the first pass, this approach would probably be fine.  What do you think?

playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
detiber: @sdodson that assumes that the user is starting with the release of their current version of OCP when they kick off the upgrade. If they are not, they will still hit this bug.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
detiber: I agree that it is useful, especially when you consider the additional parameters that you can pass to override the task file, vars files, etc that are invoked from the role or if you want to have the role task execution happen at a specific time instead of relying on the role dependency order.

My concern is more around simply using it in place of normal role dependencies or the roles attribute of a play.

It may make complete sense to use include_role in pre_tasks here, but I don't think we should be arbitrarily using it everywhere.

roles/openshift_manage_node/tasks/main.yml
ingvagabund: I see, I mixed this role with the control play upgrade. Taking my question back. I see the point now.

ingvagabund:
@kwoodson I am assuming you are referring to ``openshift_manage_node`` role. What you are saying makes perfect sense. AFAIK, the role is used in ``playbooks/common/openshift-node/config.yml`` file only. Based on my fast understanding, the role is used to set additional configuration of a node. It would make sense to get rid of the role, make its task part of the ``Additional node config`` play's task and replace the ``adm manage-node`` command with the module as @mtnbikenc has done.

ingvagabund:
aos-ci-test

mtnbikenc:
aos-ci-test

ingvagabund:
Woud be worth to log what is inside ``"{{ openshift.node.labels | oo_combine_dict | oo_split(separator=' ') }}"`` before the ``Label nodes`` task. E.g.

```yaml
- name: Print something
  debug:
    msg: "Inside: {{ openshift.node.labels | oo_combine_dict | oo_split(separator=' ') }}"
```

ashcrow:
```
fatal: [10.8.169.170 -> 10.8.169.174]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Shared connection to 10.8.169.174 closed.\r\n", "module_stdout": "Traceback (most recent call last):\r\n  File \"/tmp/ansible_FYM_kg/ansible_module_oc_label.py\", line 1605, in <module>\r\n    main()\r\n  File \"/tmp/ansible_FYM_kg/ansible_module_oc_label.py\", line 1597, in main\r\n    results = OCLabel.run_ansible(module.params, module.check_mode)\r\n  File \"/tmp/ansible_FYM_kg/ansible_module_oc_label.py\", line 1505, in run_ansible\r\n    if not oc_label.all_user_labels_exist():\r\n  File \"/tmp/ansible_FYM_kg/ansible_module_oc_label.py\", line 1333, in all_user_labels_exist\r\n    rbool = self.compare_labels(current_host_labels)\r\n  File \"/tmp/ansible_FYM_kg/ansible_module_oc_label.py\", line 1324, in compare_labels\r\n    if label['key'] not in host_labels or \\\r\nTypeError: string indices must be integers, not str\r\n", "msg": "MODULE FAILURE"}
```

Specifically:

```
 File "/tmp/ansible_m4zxDK/ansible_module_oc_label.py", line 1324, in compare_labels

    if label['key'] not in host_labels or

TypeError: string indices must be integers, not str
```

kwoodson:
@ashcrow,

I would be curious to see what the labels getting passed in look like.  If that is understood I would be able to reproduce.

mtnbikenc:
@ashcrow @kwoodson 
The labels come in like:
```
    "openshift.node.labels": {
        "region": "infra"
    }
```
and need to be transformed to
```
- key: region
  value: infra
```

It would be nice if oc_label just took a dictionary.  I've been working on figuring out required outputs/inputs to get them lined up.

mtnbikenc:
FWIW, in this last iteration I was passing labels in as ` - region=infra` but obviously that didn't work once it tried to get `key`.

mtnbikenc:
@detiber Can you comment on if there is already a filter available that would get me close to the above transformation?  I may just have to write a new filter but didn't want to redo work that has already been done.

detiber:
@mtnbikenc I am not aware of one.

mtnbikenc:
aos-ci-test

mtnbikenc:
aos-ci-test

mtnbikenc:
aos-ci-test

mtnbikenc:
PTAL @ashcrow @detiber @ingvagabund

PR #3290, opened 2017-02-08
Added tmpfile for kubeconfig. Fixed tests.

kwoodson:

PR #3250, opened 2017-02-02
Adding oc_process to lib_openshift.

kwoodson:

PR #3291, opened 2017-02-08
[WIP] oc route commands now using the oc_route module

ashcrow:
DO NOT MERGE

### DIFF
--- a/roles/cockpit-ui/tasks/main.yml
+++ b/roles/cockpit-ui/tasks/main.yml
@@ -1,86 +1,58 @@
 ---
-- name: Create temp directory for kubeconfig
-  command: mktemp -d /tmp/openshift-ansible-XXXXXX
-  register: mktemp
-  changed_when: False
+- block:
+  - name: Create passthrough route for docker-registry
+    oc_route:
+      kubeconfig: "{{ openshift_master_config_dir }}/admin.kubeconfig"
+      name: docker-registry
+      namespace: default
+      service_name: docker-registry
+      state: present
+      tls_termination: passthrough
+    register: docker_registry_route
 
-- set_fact:
-    openshift_hosted_kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
+  - name: Create passthrough route for registry-console
+    oc_route:
+      kubeconfig: "{{ openshift_master_config_dir }}/admin.kubeconfig"
+      name: registry-console
+      namespace: default
+      service_name: registry-console
+      state: present
+      tls_termination: passthrough
+    register: registry_console_cockpit_kube
 
-- name: Copy the admin client config(s)
-  command: >
-    cp {{ openshift_master_config_dir }}/admin.kubeconfig {{ openshift_hosted_kubeconfig }}
-  changed_when: False
+  # XXX: Required for items still using command
+  - name: Create temp directory for kubeconfig
+    command: mktemp -d /tmp/openshift-ansible-XXXXXX
+    register: mktemp
+    changed_when: False
 
-- name: Determine if docker-registry service exists
-  command: >
-    {{ openshift.common.client_binary }} get svc/docker-registry
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
-  register: check_docker_registry_exists
-  failed_when: false
-  changed_when: false
+  - set_fact:
+      openshift_hosted_kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
 
-- name: Create passthrough route for docker-registry
-  command: >
-    {{ openshift.common.client_binary }} create route passthrough
-    --service docker-registry
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
-  register: create_docker_registry_route
-  changed_when: "'already exists' not in create_docker_registry_route.stderr"
-  failed_when: "'already exists' not in create_docker_registry_route.stderr and create_docker_registry_route.rc != 0"
-  when: check_docker_registry_exists.rc == 0
+  - name: Copy the admin client config(s)
+    command: >
+      cp {{ openshift_master_config_dir }}/admin.kubeconfig {{ openshift_hosted_kubeconfig }}
+    changed_when: False
 
-- name: Create passthrough route for registry-console
-  command: >
-    {{ openshift.common.client_binary }} create route passthrough
-    --service registry-console
-    --port registry-console
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
-  register: create_registry_console_route
-  changed_when: "'already exists' not in create_registry_console_route.stderr"
-  failed_when: "'already exists' not in create_registry_console_route.stderr and create_registry_console_route.rc != 0"
-  when: check_docker_registry_exists.rc == 0
+  # TODO: Need to fix the origin and enterprise templates so that they both respect IMAGE_PREFIX
+  - name: Deploy registry-console
+    command: >
+      {{ openshift.common.client_binary }} new-app --template=registry-console
+      {% if openshift_cockpit_deployer_prefix is defined  %}-p IMAGE_PREFIX="{{ openshift_cockpit_deployer_prefix }}"{% endif %}
+      {% if openshift_cockpit_deployer_version is defined  %}-p IMAGE_VERSION="{{ openshift_cockpit_deployer_version }}"{% endif %}
+      -p OPENSHIFT_OAUTH_PROVIDER_URL="{{ openshift.master.public_api_url }}"
+      -p REGISTRY_HOST="{{ docker_registry_route.results.results[0].spec.host }}"
+      -p COCKPIT_KUBE_URL="https://{{ registry_console_cockpit_kube.results.results[0].spec.host }}"
+      --config={{ openshift_hosted_kubeconfig }}
+      -n default
+    register: deploy_registry_console
+    changed_when: "'already exists' not in deploy_registry_console.stderr"
+    failed_when: "'already exists' not in deploy_registry_console.stderr and deploy_registry_console.rc != 0"
 
-- name: Retrieve docker-registry route
-  command: >
-    {{ openshift.common.client_binary }} get route docker-registry
-    -o jsonpath='{.spec.host}'
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
-  register: docker_registry_route
-  changed_when: false
-  when: check_docker_registry_exists.rc == 0
-
-- name: Retrieve cockpit kube url
-  command: >
-    {{ openshift.common.client_binary }} get route registry-console
-    -o jsonpath='https://{.spec.host}'
-    -n default
-  register: registry_console_cockpit_kube_url
-  changed_when: false
-  when: check_docker_registry_exists.rc == 0
-
-# TODO: Need to fix the origin and enterprise templates so that they both respect IMAGE_PREFIX
-- name: Deploy registry-console
-  command: >
-    {{ openshift.common.client_binary }} new-app --template=registry-console
-    {% if openshift_cockpit_deployer_prefix is defined  %}-p IMAGE_PREFIX="{{ openshift_cockpit_deployer_prefix }}"{% endif %}
-    {% if openshift_cockpit_deployer_version is defined  %}-p IMAGE_VERSION="{{ openshift_cockpit_deployer_version }}"{% endif %}
-    -p OPENSHIFT_OAUTH_PROVIDER_URL="{{ openshift.master.public_api_url }}"
-    -p REGISTRY_HOST="{{ docker_registry_route.stdout }}"
-    -p COCKPIT_KUBE_URL="{{ registry_console_cockpit_kube_url.stdout }}"
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
-  register: deploy_registry_console
-  changed_when: "'already exists' not in deploy_registry_console.stderr"
-  failed_when: "'already exists' not in deploy_registry_console.stderr and deploy_registry_console.rc != 0"
-  when: check_docker_registry_exists.rc == 0
-
-- name: Delete temp directory
-  file:
-    name: "{{ mktemp.stdout }}"
-    state: absent
-  changed_when: False
+  - name: Delete temp directory
+    file:
+      name: "{{ mktemp.stdout }}"
+      state: absent
+    changed_when: False
+    # XXX: End required for items still using command
+  run_once: true
--- a/roles/openshift_hosted/tasks/registry/secure.yml
+++ b/roles/openshift_hosted/tasks/registry/secure.yml
@@ -1,13 +1,13 @@
 ---
 - name: Create passthrough route for docker-registry
-  command: >
-    {{ openshift.common.client_binary }} create route passthrough
-    --service docker-registry
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
-  register: create_docker_registry_route
-  changed_when: "'already exists' not in create_docker_registry_route.stderr"
-  failed_when: "'already exists' not in create_docker_registry_route.stderr and create_docker_registry_route.rc != 0"
+  oc_route:
+    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
+    name: docker-registry
+    namespace: default
+    service_name: docker-registry
+    state: present
+    tls_termination: passthrough
+  run_once: true
 
 - name: Determine if registry certificate must be created
   stat:
--- a/roles/cockpit-ui/meta/main.yml
+++ b/roles/cockpit-ui/meta/main.yml
@@ -11,3 +11,5 @@ galaxy_info:
     - 7
   categories:
   - cloud
+dependencies:
+- role: lib_openshift


roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -1,13 +1,19 @@
 ---
+- name: Load lib_openshift modules
+  include_role:
+    name: lib_openshift
```
mtnbikenc: This should be moved into meta/main.yml dependencies.

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -1,13 +1,19 @@
...
-    -n default
+  oc_route:
+    name: passthrough
+    namespace: default
+    service_name: docker-registry
+    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
+    state: present
   register: create_docker_registry_route
-  changed_when: "'already exists' not in create_docker_registry_route.stderr"
+  changed_when: "create_docker_registry_route.changed == True"
```
mtnbikenc: Do we need this `register` and `changed_when` clause?  It seems this would be by default from the module to returned `changed: true`

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -1,13 +1,19 @@
...
   failed_when: "'already exists' not in create_docker_registry_route.stderr and create_docker_registry_route.rc != 0"
```
mtnbikenc: And this failed_when wouldn't apply either. There is no .stderr or .rc from the module.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -22,43 +26,51 @@
...
-    -n default
+  oc_route:
+    name: passthrough
+    namespace: default
+    service_name: registry-console
+    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
+    state: present
   register: create_registry_console_route
-  changed_when: "'already exists' not in create_registry_console_route.stderr"
+  changed_when: "create_registry_console_route.changed == True"
```
mtnbikenc: This register, changed_when and failed_when should not be necessary.

roles/openshift_hosted/tasks/registry/secure.yml
ashcrow: @mtnbikenc moved to or also added in ``meta/main.yml``?

roles/cockpit-ui/tasks/main.yml
ashcrow: This was left in as it was already there. However, I'll go ahead and remove it.

ashcrow:
Updated based on @mtnbikenc's review.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -22,43 +22,45 @@
...
-  command: >
-    {{ openshift.common.client_binary }} get route registry-console
-    -o jsonpath='https://{.spec.host}'
-    -n default
+  oc_route:
+    name: registry-console
+    namespace: default
+    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
+    key_path:
+      jsonpath: "https://{.spec.host}"
```
ashcrow: I think this is wrong. Going to double check it shortly.

ashcrow: Yeah, this is wrong. It's not the same kind of key. Updating.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -22,43 +22,45 @@
...
-  command: >
-    {{ openshift.common.client_binary }} create route passthrough
-    --service docker-registry
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
-  register: create_docker_registry_route
-  changed_when: "'already exists' not in create_docker_registry_route.stderr"
-  failed_when: "'already exists' not in create_docker_registry_route.stderr and create_docker_registry_route.rc != 0"
+  oc_route:
+    name: passthrough
```
detiber: I think this should actually be:
```
oc_route:
  kubeconfig: "{{ openshift_hosted_kubeconfig }}"
  name: docker-registry
  namespace: default
  service_name: docker-registry
  state: present
  tls_termination: passthrough
```
The oc command is saying to create a passthrough route without specifying a name, which defaults to the service name.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -22,43 +22,45 @@
...
+    namespace: default
+    service_name: docker-registry
+    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
+    state: present
+  run_once: true
   when: check_docker_registry_exists.rc == 0
```
detiber: You shouldn't need to check if the route exists when using the module.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -22,43 +22,45 @@
...
-    -o jsonpath='{.spec.host}'
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
+  oc_route:
+    name: docker-registry
+    namespace: default
+    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
+    key_path:
+      jsonpath: "{.spec.host}"
+    state: list
```
detiber: Having to run the module a second time to retrieve info about the route seems wrong. I would expect the output of oc_route with state=present to return the created or existing route back.

roles/cockpit-ui/tasks/main.yml
ashcrow: Looks like it would be safe to remove the check then as well. Will do in the next set of updates.

roles/cockpit-ui/tasks/main.yml
ashcrow: Ah ok, thanks!

roles/cockpit-ui/tasks/main.yml
```diff
@@ -41,37 +41,45 @@
   run_once: true
   when: check_docker_registry_exists.rc == 0
 
-- name: Retrieve docker-registry route
+- name: Retrieve docker-registry json
   oc_route:
     name: docker-registry
     namespace: default
     kubeconfig: "{{ openshift_hosted_kubeconfig }}"
```
detiber: With https://github.com/openshift/openshift-ansible/pull/3290 being merged, you should be able to just point to the admin kubeconfig here and avoid the mktemp/delete tasks.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -22,54 +22,64 @@
...
+  run_once: true
 
-- name: Retrieve docker-registry route
-  command: >
-    {{ openshift.common.client_binary }} get route docker-registry
-    -o jsonpath='{.spec.host}'
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
-  register: docker_registry_route
+- name: Retrieve docker-registry json
```
ashcrow: @detiber / @mtnbikenc this seems really clunky to me. I'm positive there is a simpler way to accomplish this but I just have not needed to do it before. Any pointers?

roles/cockpit-ui/tasks/main.yml
```diff
@@ -41,37 +41,45 @@
   run_once: true
   when: check_docker_registry_exists.rc == 0
 
-- name: Retrieve docker-registry route
+- name: Retrieve docker-registry json
```
detiber: I think you can get away with moving `register: docker_registry_route_json` to the create task, and still use the same output. It looks like `state: present` will return the same value as doing a `state: list`

roles/cockpit-ui/tasks/main.yml
```diff
@@ -41,37 +41,45 @@
...
-      jsonpath: "https://{.spec.host}"
     state: list
-  register: registry_console_cockpit_kube_url
+  register: registry_console_cockpit_kube_json
   changed_when: false
   when: check_docker_registry_exists.rc == 0
 
+- name: Transform cockpit kube json to dict
+  set_fact:
+    registry_console_cockpit_kube: "{{ registry_console_cockpit_kube_json|from_json }}"
```
detiber: I'm not a huge fan of using set_fact, I think it would be better to either just move the filter invocation to where the variable is used, or to set the variable on the task using `vars`

roles/cockpit-ui/tasks/main.yml
detiber: I believe you can just move the `register` to the create tasks.

roles/cockpit-ui/tasks/main.yml
ashcrow: Looks like we have some filters for this. I'll try that out and remove the set_fact.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -22,54 +22,50 @@
   changed_when: false
```
detiber: Can remove everything from here up if you switch the oc_route kubeconfig arguments to use `{{ openshift_master_config_dir }}/admin.kubeconfig }}`

roles/cockpit-ui/tasks/main.yml
```diff
@@ -22,54 +22,50 @@
...
-    -n default
+- name: Retrieve docker-registry
+  oc_route:
+    name: docker-registry
+    namespace: default
+    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
+    state: list
   register: docker_registry_route
-  changed_when: false
   when: check_docker_registry_exists.rc == 0
```
detiber: check_docker_registry_exists isn't needed here.

no need to list the route in a separate task, you should be able to just register the output of the task above.

roles/cockpit-ui/tasks/main.yml
ashcrow: Good call.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -1,29 +1,7 @@
...
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
-  register: check_docker_registry_exists
-  failed_when: false
-  changed_when: false
-
 - name: Create passthrough route for docker-registry
   oc_route:
-    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
+    kubeconfig: "{{ openshift_master_config_dir }}/admin.kubeconfig }}"
```
detiber: Add `register: docker_registry_route` here, unless I've read the module wrong, it should return back the same data that you would get with a separate list.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -33,7 +11,7 @@
 
 - name: Create passthrough route for registry-console
   oc_route:
-    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
+    kubeconfig: "{{ openshift_master_config_dir }}/admin.kubeconfig }}"
```
detiber: Same as above, except for using registry_console_cockpit_kube

roles/cockpit-ui/tasks/main.yml
```diff
@@ -45,16 +23,15 @@
...
 
 - name: Retrieve cockpit kube
   oc_route:
     name: registry-console
     namespace: default
-    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
+    kubeconfig: "{{ openshift_master_config_dir }}/admin.kubeconfig }}"
     state: list
   register: registry_console_cockpit_kube
   when: check_docker_registry_exists.rc == 0
```
detiber: No need to delete the temp directory below anymore.

detiber: No need for the `when: check_docker_registry_exists.rc == 0` on the Deploy registry-console task below.

ashcrow:
@detiber thanks for the feedback. This is making the logic so much simpler!!

roles/cockpit-ui/tasks/main.yml
```diff
@@ -1,84 +1,35 @@
...
     -p OPENSHIFT_OAUTH_PROVIDER_URL="{{ openshift.master.public_api_url }}"
-    -p REGISTRY_HOST="{{ docker_registry_route.stdout }}"
-    -p COCKPIT_KUBE_URL="{{ registry_console_cockpit_kube_url.stdout }}"
+    -p REGISTRY_HOST="{{ (docker_registry_route|from_json)['spec']['host'] }}"
+    -p COCKPIT_KUBE_URL="https://{{ (registry_console_cockpit_kube|from_json)['spec']['host'] }}"
     --config={{ openshift_hosted_kubeconfig }}
     -n default
   register: deploy_registry_console
   changed_when: "'already exists' not in deploy_registry_console.stderr"
   failed_when: "'already exists' not in deploy_registry_console.stderr and deploy_registry_console.rc != 0"
```
detiber: Just for consistency, I would suggest adding `run_once: true` here as well.

detiber: I wonder if you could wrap all three tasks in a block and apply `run_once: true` there to avoid the duplication or potential breakage if the role is called from a play with multiple hosts.

ashcrow: I don't see why not,.

ingvagabund:
With the modules the code is more clear. I like it.

ingvagabund:
aos-ci-test

detiber:
Test failures are saying that /etc/origin/master/admin.kubeconfig doesn't exist. I'm wondering if  there is a typo or if the file is being accidentally deleted somehow

ingvagabund:
> Test failures are saying that /etc/origin/master/admin.kubeconfig doesn't exist. I'm wondering if there is a typo or if the file is being accidentally deleted somehow

Could be cause by the ``run_once: true``.

ingvagabund:
Run on ``hosts: oo_first_master`` (`playbooks/common/openshift-cluster/openshift_hosted.yml`) anyway so it can be removed.

ashcrow:
@detiber @ingvagabund it looks like the rest of the run uses the ``tmp`` location (EX: ``/tmp/openshift-ansible-6q7jyy/admin.kubeconfig``). Should I maybe move back to using that?

ingvagabund:
@ashcrow it is worth trying if it helps

ashcrow:
Doing a rebase first. Looks like there was an update that I didn't pull in yesterday that I may have needed.

ashcrow:
aos-ci-test

ashcrow:
Better error with the rebase. Looks like it was a combo: ```IOError: [Errno 2] No such file or directory: '/etc/origin/master/admin.kubeconfig }}'``` :laughing:

ashcrow:
aos-ci-test

ingvagabund:
```yaml
$ cat roles/lib_openshift/tasks/main.yml 
---
  package:
    name: "{{ item }}"
    state: present
  with_items:
  - ruamel.yaml
  - ruamel.yaml
  - ruamel.yaml
  - ruamel.yaml
  - ruamel.yaml
  - ruamel.yaml
```

That is definitely wrong. Caused by https://github.com/openshift/openshift-ansible/pull/3250

ingvagabund:
Reverting the #3250 here https://github.com/openshift/openshift-ansible/pull/3326

ashcrow:
@ingvagabund great find. I'm rebasing now to pull in the fix.

ashcrow:
aos-ci-test

ashcrow:
:confused: 

```
fatal: [10.8.169.26]: FAILED! => {"failed": true, "msg": "Unexpected templating type error occurred on ({{ openshift.common.client_binary }} new-app --template=registry-console {% if openshift_cockpit_deployer_prefix is defined  %}-p IMAGE_PREFIX=\"{{ openshift_cockpit_deployer_prefix }}\"{% endif %} {% if openshift_cockpit_deployer_prefix is defined  %}-p IMAGE_PREFIX=\"{{ openshift_cockpit_deployer_prefix }}\"{% endif %} {% if openshift_cockpit_deployer_version is defined  %}-p IMAGE_VERSION=\"{{ openshift_cockpit_deployer_version }}\"{% endif %} -p OPENSHIFT_OAUTH_PROVIDER_URL=\"{{ openshift.master.public_api_url }}\" -p REGISTRY_HOST=\"{{ (docker_registry_route|from_json)['spec']['host'] }}\" -p COCKPIT_KUBE_URL=\"https://{{ (registry_console_cockpit_kube|from_json)['spec']['host'] }}\" --config={{ openshift_hosted_kubeconfig }} -n default): expected string or buffer"}
```

ashcrow:
aos-ci-test

ashcrow:
aos-ci-test

ingvagabund:
@ashcrow 

``{{ docker_registry_route.results.host }}`` and ``{{ registry_console_cockpit_kube.results.host }}`` should do the trick.

ashcrow:
aos-ci-test

ashcrow:
@ingvagabund I'll try with ``spec`` added.

```
fatal: [10.8.169.239]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'host'
```

ashcrow:
aos-ci-test

ashcrow:
aos-ci-test

ashcrow:
Added debug prints to find out what the data should be per @ingvagabund

ashcrow:
aos-ci-test

ingvagabund:
Ha :D ``results.spec.hosts`` it is.

ashcrow:
aos-ci-test

ingvagabund:
Uhh ``dict object has no element 0``

ashcrow:
I'm going to try: ```results.results[0].spec.host```

ashcrow:
aos-ci-test

ashcrow:
bot, are you there? It's me. @ashcrow.

ashcrow:
aos-ci-test

ingvagabund:
Finally it is a different error: ``error: stat /tmp/openshift-ansible-Y9hPvO/admin.kubeconfig: no such file or directory``

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -1,13 +1,13 @@
...
-  command: >
-    {{ openshift.common.client_binary }} create route passthrough
-    --service docker-registry
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
-  register: create_docker_registry_route
-  changed_when: "'already exists' not in create_docker_registry_route.stderr"
-  failed_when: "'already exists' not in create_docker_registry_route.stderr and create_docker_registry_route.rc != 0"
+  oc_route:
+    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
```
ingvagabund: Removing this line should do the trick. Even if I don't get is why the file does not exist anymore.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -1,86 +1,37 @@
 ---
-- name: Create temp directory for kubeconfig
-  command: mktemp -d /tmp/openshift-ansible-XXXXXX
-  register: mktemp
-  changed_when: False
-
-- set_fact:
-    openshift_hosted_kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
```
ingvagabund: Hmm, maybe caused by this as it is no longer defined, hmmm

roles/openshift_hosted/tasks/registry/secure.yml
ingvagabund: Lines 26, 47 and other in this same file will error the same way most likely. 

Maybe just add here the same mkdir -> copy mechanism:
```yaml
 set_fact:
    openshift_hosted_kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"

- name: Copy the admin client config(s)
  command: >
    cp {{ openshift_master_config_dir }}/admin.kubeconfig {{ openshift_hosted_kubeconfig }}
  changed_when: False
```

roles/cockpit-ui/tasks/main.yml
detiber: This should not be affecting tasks within openshift_hosted.

roles/openshift_hosted/tasks/registry/secure.yml
detiber: This *should* be using openshift_hosted_kubeconfig defined here: https://github.com/ashcrow/openshift-ansible/blob/d4dbf337bdc77fc73d9bcad599205c429cdb91b4/roles/openshift_hosted/tasks/main.yml#L8

That said, there shouldn't be a need to use a copy of the kubeconfig here, since the module will already make a copy of the kubeconfig.

ashcrow: This one I think is fine. But I need to reintroduce the copy in the cockpit-ui module as there is one command that doesn't use a module yet and needs a kubeconfig. Using the same kubeconfig could modify it and cause problems downline. I'll comment around it so it's obvious why the copy is occuring.

ashcrow:
aos-ci-test

roles/cockpit-ui/tasks/main.yml
ashcrow: This is where the actual issue is I believe. ``openshift_hosted`` didn't report an error AFAICT but ``cockpit-ui`` did due to the use of command.

ashcrow:
@ingvagabund :fireworks:

PR #3292, opened 2017-02-08
WIP: replace all relevant oc commands with modules from lib_openshift

ingvagabund:
In progress

### DIFF
--- a/roles/openshift_logging/tasks/generate_secrets.yaml
+++ b/roles/openshift_logging/tasks/generate_secrets.yaml
@@ -56,22 +56,45 @@
   changed_when: no
 
 - name: Generating secrets for elasticsearch
-  command: >
-    {{openshift.common.client_binary}} --config={{ mktemp.stdout }}/admin.kubeconfig secrets new {{secret_name}}
-    key={{generated_certs_dir}}/logging-es.jks truststore={{generated_certs_dir}}/truststore.jks
-    searchguard.key={{generated_certs_dir}}/elasticsearch.jks searchguard.truststore={{generated_certs_dir}}/truststore.jks
-    admin-key={{generated_certs_dir}}/system.admin.key admin-cert={{generated_certs_dir}}/system.admin.crt
-    admin-ca={{generated_certs_dir}}/ca.crt admin.jks={{generated_certs_dir}}/system.admin.jks -o yaml
+  block:
+    - name: Load lib_openshift modules
+      include_role:
+      name: lib_openshift
+    - oc_secret:
+        state: present
+        name: {{secret_name}}
+        kubeconfig: {{ mktemp.stdout }}/admin.kubeconfig
+        files:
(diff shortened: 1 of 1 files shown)


roles/openshift_logging/tasks/generate_secrets.yaml
```diff
@@ -56,22 +56,45 @@
...
     secret_name: logging-elasticsearch
     secret_keys: ["admin-cert", "searchguard.key", "admin-ca", "key", "truststore", "admin-key"]
-  register: logging_es_secret
   when: secret_name not in openshift_logging_facts.elasticsearch.secrets or
         secret_keys | difference(openshift_logging_facts.elasticsearch.secrets["{{secret_name}}"]["keys"]) | length != 0
   check_mode: no
   changed_when: no
 
-- copy: content="{{logging_es_secret.stdout}}" dest={{mktemp.stdout}}/templates/logging-elasticsearch-secret.yaml
+- copy: content="{{logging_es_secret.stdout|to_yaml}}" dest={{mktemp.stdout}}/templates/logging-elasticsearch-secret.yaml
```
detiber: This seems a bit odd that the logging_es_secret var registered by oc_secret would need to be used this way.

I would expect the following:
- The result of oc_secret with state=present to return the created secret
- The result of oc_secret with state=present or state=list would return a datastructure more like:
```
result:
  changed: <changed>
  secret:
    api_version: v1
    data:
      key: <base64 content>
      truststore: <base64 content>
      ...
    kind: Secret
    metadata: <metadata>
    type: <type>
```

roles/openshift_logging/tasks/generate_secrets.yaml
```diff
@@ -56,22 +56,45 @@
...
-  command: >
-    {{openshift.common.client_binary}} --config={{ mktemp.stdout }}/admin.kubeconfig secrets new {{secret_name}}
-    key={{generated_certs_dir}}/logging-es.jks truststore={{generated_certs_dir}}/truststore.jks
-    searchguard.key={{generated_certs_dir}}/elasticsearch.jks searchguard.truststore={{generated_certs_dir}}/truststore.jks
-    admin-key={{generated_certs_dir}}/system.admin.key admin-cert={{generated_certs_dir}}/system.admin.crt
-    admin-ca={{generated_certs_dir}}/ca.crt admin.jks={{generated_certs_dir}}/system.admin.jks -o yaml
+  block:
+    - name: Load lib_openshift modules
+      include_role:
+      name: lib_openshift
```
detiber: lib_openshift should really be a role dependency here, rather than using include_role.

ingvagabund: right, openshift_logging is a role

ingvagabund:
This PR was supposed to replace oc_serviceaccount_secret.

PR #3293, opened 2017-02-08
fix bug 1420204. Default openshift_logging_use_journal to empty so fl…

jcantrill:
…uentd detects and is consistent with deployer

cc @richm

### DIFF
--- a/roles/openshift_logging/README.md
+++ b/roles/openshift_logging/README.md
@@ -51,8 +51,8 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_fluentd_cpu_limit`: The CPU limit for Fluentd pods. Defaults to '100m'.
 - `openshift_logging_fluentd_memory_limit`: The memory limit for Fluentd pods. Defaults to '512Mi'.
 - `openshift_logging_fluentd_es_copy`: Whether or not to use the ES_COPY feature for Fluentd (DEPRECATED). Defaults to 'False'.
-- `openshift_logging_fluentd_use_journal`: Whether or not Fluentd should read log entries from Journal. Defaults to 'False'. NOTE: Fluentd will attempt to detect whether or not Docker is using the journald log driver and may overwrite this value.
-- `openshift_logging_fluentd_journal_read_from_head`: Whether or not Fluentd will try to read from the head of Journal when first starting up, using this may cause a delay in ES receiving current log records. Defaults to 'False'.
+- `openshift_logging_fluentd_use_journal`: NOTE: Fluentd will attempt to detect whether or not Docker is using the journald log driver when using the default of empty.
+- `openshift_logging_fluentd_journal_read_from_head`: If empty, Fluentd will use its internal default, which is false.
 - `openshift_logging_fluentd_hosts`: List of nodes that should be labeled for Fluentd to be deployed to. Defaults to ['--all'].
 
 - `openshift_logging_es_host`: The name of the ES service Fluentd should send logs to. Defaults to 'logging-es'.
--- a/roles/openshift_logging/defaults/main.yml
+++ b/roles/openshift_logging/defaults/main.yml
@@ -39,8 +39,8 @@ openshift_logging_fluentd_nodeselector: {'logging-infra-fluentd': 'true'}
 openshift_logging_fluentd_cpu_limit: 100m
 openshift_logging_fluentd_memory_limit: 512Mi
 openshift_logging_fluentd_es_copy: false
-openshift_logging_fluentd_use_journal: false
-openshift_logging_fluentd_journal_read_from_head: false
+openshift_logging_fluentd_use_journal: ''
+openshift_logging_fluentd_journal_read_from_head: ''
 openshift_logging_fluentd_hosts: ['--all']
 
 openshift_logging_es_host: logging-es


roles/openshift_logging/README.md
```diff
@@ -51,8 +51,8 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_fluentd_cpu_limit`: The CPU limit for Fluentd pods. Defaults to '100m'.
 - `openshift_logging_fluentd_memory_limit`: The memory limit for Fluentd pods. Defaults to '512Mi'.
 - `openshift_logging_fluentd_es_copy`: Whether or not to use the ES_COPY feature for Fluentd (DEPRECATED). Defaults to 'False'.
-- `openshift_logging_fluentd_use_journal`: Whether or not Fluentd should read log entries from Journal. Defaults to 'False'. NOTE: Fluentd will attempt to detect whether or not Docker is using the journald log driver and may overwrite this value.
-- `openshift_logging_fluentd_journal_read_from_head`: Whether or not Fluentd will try to read from the head of Journal when first starting up, using this may cause a delay in ES receiving current log records. Defaults to 'False'.
+- `openshift_logging_fluentd_use_journal`: Whether or not Fluentd should read log entries from Journal. NOTE: Fluentd will attempt to detect whether or not Docker is using the journald log driver and may overwrite this value when using the default of empty.
```
richm: Fluentd doesn't "overwrite" the value.  I would just say this: "NOTE: Fluentd will attempt to detect whether or not Docker is using the journald log driver when using the default of empty."

roles/openshift_logging/README.md
```diff
@@ -51,8 +51,8 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
...
+- `openshift_logging_fluentd_journal_read_from_head`: Whether or not Fluentd will try to read from the head of Journal when first starting up, using this may cause a delay in ES receiving current log records. Defaults to empty.
```
richm: If empty, Fluentd will use its internal default, which is `false`.

ewolinetz:
Pending the doc updates as Rich stated LGTM

jcantrill:
@ewolinetz updated README entries.  please do the needful

roles/openshift_logging/templates/fluentd.j2
```diff
@@ -117,11 +117,11 @@ spec:
         - name: "OPS_COPY_PASSWORD"
           value: "{{ops_copy_password | default('')}}"
         - name: "USE_JOURNAL"
-          value: "{{openshift_logging_fluentd_use_journal|lower}}"
+          value: "{{openshift_logging_fluentd_use_journal|default('')|lower}}"
```
detiber: If you are specifying the default in defaults, you shouldn't need to use the default filter here.

portante:
Why not detect the empty value first, change it to the default of `false`, and then use it as a proper boolean?

richm:
LGTM
@portante - detect where?  change it where?  the use_journal value as '' must be passed all the way through to the fluentd.yml template and the fluentd pod run.sh in order for the autodetect logic there to be triggered

portante:
@richm, so I am probably not understanding the problem given your questions.

If `openshift_logging_fluentd_use_journal` is the empty string (`""`) when I run the playbook, then does fluentd interpret that to mean the environment variable has a `false` value?

If `openshift_logging_fluentd_use_journal` is the string (`"false"`) then does fluentd parse that to mean the environment variable has a `false` value, or does it consider it a `true` value because the `USE_JOURNAL` environment is now present?

If `openshift_logging_fluentd_use_journal` is the string (`"bad"`) then does fluentd parse that to mean the environment variable has a `false` value?

richm:
@portante A picture of code is worth a thousand words - https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/run.sh#L17

The value of `openshift_logging_fluentd_use_journal` is not actually a boolean.  It is a string value which should have one of 3 values: `''`, `true`, or `false`.  This value eventually sets the value of `USE_JOURNAL` in the fluentd pod environment.

> If openshift_logging_fluentd_use_journal is the empty string ("") when I run the playbook, then does fluentd interpret that to mean the environment variable has a false value?

No.  This will cause `USE_JOURNAL` to be set to the empty value, which tells run.sh to autodetect which log-driver docker is using.

> If openshift_logging_fluentd_use_journal is the string ("false") then does fluentd parse that to mean the environment variable has a false value, or does it consider it a true value because the USE_JOURNAL environment is now present?

This causes `USE_JOURNAL` to have the string value `false`, which fluentd interprets as the string value `"false"`.

> If openshift_logging_fluentd_use_journal is the string ("bad") then does fluentd parse that to mean the environment variable has a false value?

The behavior is undefined.  This will likely cause really strange problems where some of the code will interpret `bad` to mean `false` (since `bad != true`) and some of the code will interpret `bad` to mean `true` (since `bad != false`)

jcantrill:
@sdodson @ewolinetz any objections/approvals?

jcantrill:
aos-ci-test

roles/openshift_logging/templates/fluentd.j2
sdodson: yeah lets fix this

roles/openshift_logging/templates/fluentd.j2
```diff
@@ -117,11 +117,11 @@ spec:
...
         - name: "JOURNAL_SOURCE"
           value: "{{fluentd_journal_source | default('')}}"
         - name: "JOURNAL_READ_FROM_HEAD"
-          value: "{{openshift_logging_fluentd_journal_read_from_head|lower}}"
+          value: "{{openshift_logging_fluentd_journal_read_from_head|default('')|lower}}"
       volumes:
```
sdodson: this too

portante:
@richm, okay, so that seems like really confusing behavior:

```
#!/bin/bash

USE_JOURNAL=$1

if [ "${USE_JOURNAL:-}" != false ] ; then
   echo "USE_JOURNAL is true"
else
   echo "USE_JOURNAL is false"
fi
```

Yields:
```
$ ./this.bash
USE_JOURNAL is true
$ ./this.bash bad
USE_JOURNAL is true
$ ./this.bash false
USE_JOURNAL is false
$ ./this.bash true
USE_JOURNAL is true
```

richm:
@portante yes, it is.  feel free to file an issue and/or provide a PR against openshift/origin-aggregated-logging

PR #3294, opened 2017-02-08
use pvc_size instead of pv_size for openshift_metrics since the role …

jcantrill:
…creates claims


jcantrill:
aos-ci-test

jcantrill:
@sdodson  i didnt have a merge option. is this because i had to 'ci' test even though its not hooked up for these roles?

detiber:
@jcantrill yes, we don't currently have a way to differentiate which PRs need ci testing or not.

PR #3296, opened 2017-02-08
[DO NOT MERGE] Move all module deps into playbooks/common/openshift-cluster/config.yml

sdodson:
For science!

### DIFF
--- a/playbooks/common/openshift-cluster/config.yml
+++ b/playbooks/common/openshift-cluster/config.yml
@@ -25,6 +25,25 @@
   - set_fact:
       openshift_docker_selinux_enabled: "{{ lookup('oo_option', 'docker_selinux_enabled') }}"
     when: openshift_docker_selinux_enabled is not defined
+  - name: Detecting Operating System
+    stat:
+      path: /run/ostree-booted
+    register: ostree_booted
+  # Locally setup containerized facts for now
+  - set_fact:
+      l_is_atomic: "{{ ostree_booted.stat.exists }}"
+  - set_fact:
+      l_is_containerized: "{{ (l_is_atomic | bool) or (containerized | default(false) | bool) }}"
+  - name: Ensure various deps are installed
+    package: name={{ item }} state=present
+    with_items:
+    - python-ruamel-yaml
+    - iproute
+    - python-dbus
+    - python-six
+    - PyYAML
+    - yum-utils
+    when: not l_is_atomic | bool
 
 - include: disable_excluder.yml
   tags:
--- a/roles/openshift_facts/vars/main.yml
+++ b/roles/openshift_facts/vars/main.yml
@@ -1,7 +1,1 @@
 ---
-required_packages:
-  - iproute
-  - python-dbus
-  - python-six
-  - PyYAML
-  - yum-utils


sdodson:
aos-ci-test

playbooks/common/openshift-cluster/config.yml
```diff
@@ -25,6 +25,25 @@
...
+  - name: Ensure various deps are installed
+    package: name={{ item }} state=present
+    with_items:
+    - python-ruamel-yaml
+    - iproute
+    - python-dbus
+    - python-six
+    - PyYAML
+    - yum-utils
+    when: not l_is_atomic | bool
```
detiber: I worry about doing this in the playbook here, there are other playbooks that can be called (scaleup, upgrade, cert regen, etc) that would miss installing these dependencies...

sdodson:
Yeah, just trying to get a before and after measure.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

DanyC97:
who is on the list of reviewers and can show this PR some love? :)

sdodson:
@DanyC97 We're trying to decide as a team whether or not to pursue this approach. We have to weigh overall performance and looser coupling versus the reliability of ensuring dependencies are installed whenever pulling the role into scope. I'd also like to measure the impact of this which is what I was hoping to get from this PR by way of jenkins job trends.

What's your opinion on the matter?

sdodson:
aos-ci-test

DanyC97:
@sdodson i'm still new and so i might miss a lot of background but since you asked for my opinion (appreciated) my angle is:

- reliability of ensuring dependencies are installed whenever pulling the role into scope is key
- but then unifying and reducing the overall repo complexity is s'thing which we need to achieve and to me having things aggregated/ located in `common` is easier to follow.

detiber:
@sdodson is this something we are still experimenting with? Personally, I think this is a bad idea. There seems to be too much room for doing something wrong and not having the proper packages needed for different workflows.

I think we could achieve a similar result longer term if we moved package installations into a separate task file in the roles and had a pre-req playbook/role that used `include_role` and `tasks_from`. Then we could use set_fact in the install task files, so we could short-circuit installation by testing for that variable at the include level.

<role name>/tasks/install_packages.yml:
```
<package install steps>

- set_fact:
    <role name>_packages_installed: yes
```

<role name>/tasks/main.yml:
```
<non-package steps>

- include: install_packages.yml
  static: no
  when:
  - not <role name>_packages_installed | default(false)
```

That would also allow someone to explicitly set the vars to shortcut package installs without running the pre-req steps if they desired. Also, by explicitly using a dynamic include we will skip the include step itself when the condition is false, so we wouldn't have the performance hit of skipping the tasks in `install_packages.yml`.

The one catch with this approach (and the playbook pre-req approach as well) is that we couldn't rely on this approach for in-place upgrades, because we wouldn't be able to accurately trigger handlers for restarting services on package updates. We would need to track that with some other method (maybe a local fact we store temporarily on the remote system, since we cannot guarantee that the playbook run was not interrupted and chances of failure are higher than if it was all contained in the same role). If a local fact is used, I think it should be explicit for the service being managed and we should un-set it as part of the service restart, maybe even use the changed_when condition of inserting it to trigger the service restart handler.

PR #3297, opened 2017-02-08
bug 1419962. fix openshift_metrics pwd issue after reinstall where ca…

jcantrill:
…ssandra has incorrect pwd exception

This fixes the case of install/uninstall/install where cassandra is unable to start because of a bad password.  There was some funny business between creating stores and using some password but not using the same one when generating the secrets.

As an aside, I think there are optimizations about where certain password and cert files could be generated, but given cert generation service is in the near future, I don't think its worth to implement these efficiencies.

### DIFF
--- a/roles/openshift_metrics/defaults/main.yaml
+++ b/roles/openshift_metrics/defaults/main.yaml
@@ -39,7 +39,6 @@ openshift_metrics_resolution: 15s
 # overriding the values here
 #####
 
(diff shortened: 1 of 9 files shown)


ewolinetz:
so the basic idea is that we will generate the certs every time for metrics?

jcantrill:
@ewolinetz  yes, per my irc conversation with @mwringe there is no reason we need to keep the certs around as they are all used internally.  I'm not sure why we couldnt do the same with logging.

jcantrill:
aos-ci-tests

PR #3299, opened 2017-02-08
Adding oc_env to lib_openshift.

kwoodson:


roles/lib_openshift/src/class/oc_env.py
```diff
@@ -0,0 +1,147 @@
...
+class OCEnv(OpenShiftCLI):
+    ''' Class to wrap the oc command line tools '''
+
+    container_path = {"pod": "spec.containers[0].env",
+                      "dc":  "spec.template.spec.containers[0].env",
+                      "rc":  "spec.template.spec.containers[0].env",
+                     }
+
+    # pylint allows 5. we need 6
+    # pylint: disable=too-many-arguments
```
detiber: Should we discuss increasing the allowed number of arguments?

roles/lib_openshift/src/class/oc_env.py
```diff
@@ -0,0 +1,147 @@
...
+        result = self._get(self.kind, self.name)
+        if result['returncode'] == 0:
+            if self.kind == 'dc':
+                self.resource = DeploymentConfig(content=result['results'][0])
+                result['results'] = self.resource.get(OCEnv.container_path[self.kind]) or []
+        return result
+
+    def delete(self):
+        '''return all pods '''
+        #yed.put(OCEnv.container_path[self.kind], env_vars_array)
```
detiber: Should this be uncommented or removed?

roles/lib_openshift/src/class/oc_env.py
```diff
@@ -0,0 +1,147 @@
...
+        '''return a environment variables '''
+        result = self._get(self.kind, self.name)
+        if result['returncode'] == 0:
+            if self.kind == 'dc':
+                self.resource = DeploymentConfig(content=result['results'][0])
+                result['results'] = self.resource.get(OCEnv.container_path[self.kind]) or []
+        return result
+
+    def delete(self):
+        '''return all pods '''
```
detiber: comment doesn't match method

roles/lib_openshift/src/class/oc_env.py
```diff
@@ -0,0 +1,147 @@
...
+        #yed.put(OCEnv.container_path[self.kind], env_vars_array)
+        if self.resource.delete_env_var(self.env_vars.keys()):
+            return self._replace_content(self.kind, self.name, self.resource.yaml_dict)
+
+        return {'returncode': 0, 'changed': False}
+
+    # pylint: disable=too-many-function-args
```
detiber: This pylint disable doesn't look needed.

roles/lib_openshift/src/class/oc_env.py
```diff
@@ -0,0 +1,147 @@
...
+    def put(self):
+        '''place env vars into dc '''
+        for update_key, update_value in self.env_vars.items():
+            self.resource.update_env_var(update_key, update_value)
+
+        return self._replace_content(self.kind, self.name, self.resource.yaml_dict)
+
+    # pylint: disable=too-many-return-statements
```
detiber: Should we consider disabling this check?

roles/lib_openshift/src/class/oc_env.py
```diff
@@ -0,0 +1,147 @@
...
+    def resource(self, data):
+        ''' setter function for resource var'''
+        self._resource = data
+
+    def value_exists(self, key, value):
+        ''' return whether a key, value  pair exists '''
+        return self.resource.exists_env_value(key, value)
+
+    def key_exists(self, key):
+        ''' return whether a key, value  pair exists '''
```
detiber: s/key, value pair/key/

roles/lib_openshift/src/class/oc_env.py
```diff
@@ -0,0 +1,147 @@
...
+        return self.resource.exists_env_key(key)
+
+    def get(self):
+        '''return a environment variables '''
```
detiber: s/ a/a/

roles/lib_openshift/src/doc/env
```diff
@@ -0,0 +1,76 @@
...
+  kind:
+    description:
+    - The kind attribute of the object.
+    required: False
+    default: dc
+    choices:
+    - rc
+    - dc
+    - pods
+    aliases: []
```
detiber: env_vars param is not documented.

roles/lib_openshift/src/class/oc_env.py
abutcher: We might as well bump it to a higher value like 10?

detiber: @ashcrow thoughts on this?

roles/lib_openshift/src/class/oc_env.py
kwoodson: @detiber, @abutcher, I'm generally ok with leaving the check.  We have plans to refactor the way the run_ansible function works.  We want to break the functions up by state.  Then the logic in the ansible function would be extremely simple and would be `If state == 'list': return run_state_list()`.  Same goes for present and absent states.  That's on the list to do for the next iteration.

roles/lib_openshift/src/doc/env
kwoodson: good catch.  Adding doc.

roles/lib_openshift/src/class/oc_env.py
kwoodson: @abutcher @detiber, Would also like feedback on number of params.  Its hard when mapping calls with 5+ parameters when self is used as one.  Then you get into an abstraction layer where you pass around a dto/dictionary where the data ultimately lives.  @ashcrow, thoughts on this?

kwoodson:
@detiber @abutcher @tbielawa 

Fixed comments and suggestions.  Ready for review.

roles/lib_openshift/src/class/oc_env.py
detiber: Personally, I find limiting the number of arguments to be an arbitrary metric. I'm good with bumping it up to 10 (or even removing the check).

I'm wondering if we need to find some type of mechanism for tracking and voting on these types of issues?

PR #3300, opened 2017-02-08
WIP: oc secrets now done via oc_secret module

ashcrow:


ashcrow:
@kwoodson Is there a different module for assigning secrets to pods or did I miss the functionality when looking over the doc?

detiber:
@ashcrow I think oc_serviceaccount_secret is what you are looking for.

You shouldn't need it for redeploy, since it is replacing an existing secret.

playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
```diff
@@ -70,13 +72,17 @@
...
+        kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
+        name: registry-certificates
+        namespace: default
+        state: present
+        files:
+        - name: registry.crt
+          path: "{{ openshift.common.config_base }}/master/registry.crt"
+        - name: registry.key
+          path: "{{ openshift.common.config_base }}/master/registry.key"
+      run_once: true
```
ingvagabund: Does it make sense to use ``run_once`` if the hosts is ``oo_first_master`` only?

playbooks/common/openshift-cluster/redeploy-certificates/router.yml
```diff
@@ -44,11 +48,14 @@
...
-      command: >
-        {{ openshift.common.client_binary }} delete secret/router-certs
-        --config={{ mktemp.stdout }}/admin.kubeconfig
-        -n default
+      oc_secret:
+        kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
+        name: router-certs
+        namespace: default
+        state: absent
+        run_once: true
```
ingvagabund: Does it make sense to use `run_once` if the hosts is `oo_first_master` only?

kwoodson:
@ashcrow,

Thanks for the question.  Since these modules are somewhat new I would like to explain.  In Openshift, secrets are their own resource type.  Secrets generally get attached to deployment configs and through their association the resulting replication controller (keeps a copy of the deploymentconfig) which controls the corresponding pod will then ensure the pod receives the secret.  The secret gets attached through volumes.

If you want to attach a secret to a pod, I'm not 100% sure if that works. I've added secrets to deployment configs.  I think the proper way would be to add it to the replication controller. The command would look something like this:
https://docs.openshift.com/enterprise/3.2/dev_guide/volumes.html

According to the docs:
`oc volume rc/r1 --add --name=v1 --type=secret --secret-name='$ecret' --mount-path=/data`

oc_serviceaccount_secret is what maps secrets to service accounts.  What we have that attempts to manage secrets for deployment configs or replication controllers is oc_volume.  It is on my list to port.

Hope this helps.

playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
kwoodson: @ingvagabund, 

I think run_once is fine even if the list of masters is a single.  If that ever changed it would continue to work.

ashcrow:
@kwoodson thanks, that does help though I have one more question. I ported the creation of a secret in this pr ([reference](https://github.com/ashcrow/openshift-ansible/blob/60a45259104a1031f0ba7e75057fb63892290727/playbooks/common/openshift-cluster/redeploy-certificates/registry.yml#L74-L86]) which is based on the ``new`` command. Is there a specific invocation for the ``oc secrets add``?

detiber:
@ashcrow oc secrets add can be replaced with oc_serviceaccount_secret

ashcrow:
@ingvagabund see https://github.com/openshift/openshift-ansible/pull/3300#issuecomment-278673099

ingvagabund:
Already done it as part of the #3307

ashcrow:
aos-ci-test

ashcrow:
Errors are related to ```ruamel.yaml``` not existing. ```fatal: [10.8.169.36]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Shared connection to 10.8.169.36 closed.\r\n", "module_stdout": "Traceback (most recent call last):\r\n  File \"/tmp/ansible_2BE_gm/ansible_module_oc_secret.py\", line 43, in <module>\r\n    import ruamel.yaml as yaml\r\nImportError: No module named ruamel.yaml\r\n", "msg": "MODULE FAILURE"}```

detiber:
@ashcrow was just looking at that, I'm wondering why the task in lib_openshift isn't kicking off to install it.

detiber:
@ashcrow lib_openshift wasn't added as a dependency for the openshift_hosted role, but that doesn't fully explain the error.

playbooks/common/openshift-cluster/redeploy-certificates/router.yml
```diff
@@ -1,4 +1,8 @@
 ---
+- name: Load lib_openshift modules
+  include_role:
```
detiber: Should probably use `roles` on the play below rather than include_role here. I'm not even sure include_role is valid at this level.

ashcrow:
aos-ci-test

ashcrow:
This is odd. The bottom says unittests passed yet it shows failed in the Github notification...

detiber:
@ashcrow I think there was a failure after the tests completed, but before the jenkins job finished.

ashcrow:
aos-ci-test

ashcrow:
I took a look at jenkins and the unittests passed. Jenkins threw a stack trace on something unrelated :roll_eyes:

ingvagabund:
The same error in all tests:

```
TASK [openshift_hosted : Create the secret for the registry certificates] ******
fatal: [10.8.171.151]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Shared connection to 10.8.171.151 closed.\r\n", "module_stdout": "Traceback (most recent call last):\r\n  File \"/tmp/ansible_gXnrAt/ansible_module_oc_secret.py\", line 43, in <module>\r\n    import ruamel.yaml as yaml\r\nImportError: No module named ruamel.yaml\r\n", "msg": "MODULE FAILURE"}
```

Not related to this PR.

detiber:
@ingvagabund I disagree that it is not related, it is saying that the oc_secret module isn't able to find the ruamel.yaml library, which should be installed with the inclusion of the lib_openshift role.

ingvagabund:
aos-ci-test

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -43,15 +43,18 @@
...
+      path: "{{ openshift_master_config_dir }}/registry.crt"
+    - name: registry.key
+      path: "{{ openshift_master_config_dir }}/registry.key"
   register: create_registry_certificates_secret
-  changed_when: "'already exists' not in create_registry_certificates_secret.stderr"
-  failed_when: "'already exists' not in create_registry_certificates_secret.stderr and create_registry_certificates_secret.rc != 0"
+  run_once: true
 
 - name: "Add the secret to the registry's pod service accounts"
   command: >
```
detiber: does the openshift_hosted role already have lib_openshift as a dependency?

ingvagabund:
> @ingvagabund I disagree that it is not related, it is saying that the oc_secret module isn't able to find the ruamel.yaml library, which should be installed with the inclusion of the lib_openshift role.

It was related to https://github.com/openshift/openshift-ansible/pull/3327

roles/openshift_hosted/tasks/registry/secure.yml
ingvagabund: not yet

ashcrow:
aos-ci-test

ashcrow:
Errors don't seem related to the change....
```
    error: package directory 'simplejson' does not exist
```

ashcrow:
Made a harmless change so CI will rerun with request.

ashcrow:
aos-ci-test

ashcrow:
aos-ci-test

PR #3304, opened 2017-02-09
Corrected mistake in container resources

drewhemm:
This is `limits` in both the memcached and postgresql deployment configs,
using `requests` causes an error during deployment.


drewhemm:
Realised the error was being caused by our ResourceQuotas and LimitRanges ;)

PR #3306, opened 2017-02-09
replace 'oc service' command with its lib_openshift equivalent

ingvagabund:
SSIA

### DIFF
--- a/playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
+++ b/playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
@@ -46,12 +46,15 @@
 
   # Replace dc/docker-registry certificate secret contents if set.
   - block:
+    - name: Load lib_openshift modules
+      include_role:
+        name: lib_openshift
+
     - name: Retrieve registry service IP
-      command: >
-        {{ openshift.common.client_binary }} get service docker-registry
-        -o jsonpath='{.spec.clusterIP}'
-        --config={{ mktemp.stdout }}/admin.kubeconfig
-        -n default
+      oc_service:
+        namespace: default
+        name: docker-registry
+        state: list
       register: docker_registry_service_ip
       changed_when: false
 
@@ -65,7 +68,7 @@
         --signer-cert={{ openshift.common.config_base }}/master/ca.crt
         --signer-key={{ openshift.common.config_base }}/master/ca.key
         --signer-serial={{ openshift.common.config_base }}/master/ca.serial.txt
-        --hostnames="{{ docker_registry_service_ip.stdout }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}"
+        --hostnames="{{ docker_registry_service_ip.results.clusterip }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}"
         --cert={{ openshift.common.config_base }}/master/registry.crt
         --key={{ openshift.common.config_base }}/master/registry.key
 
--- a/roles/openshift_hosted/tasks/registry/secure.yml
+++ b/roles/openshift_hosted/tasks/registry/secure.yml
@@ -20,11 +20,10 @@
   failed_when: false
 
 - name: Retrieve registry service IP
-  command: >
-    {{ openshift.common.client_binary }} get service docker-registry
-    -o jsonpath='{.spec.clusterIP}'
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
+  oc_service:
+    namespace: default
+    name: docker-registry
+    state: list
   register: docker_registry_service_ip
   changed_when: false
 
@@ -37,7 +36,7 @@
     --signer-cert={{ openshift_master_config_dir }}/ca.crt
     --signer-key={{ openshift_master_config_dir }}/ca.key
     --signer-serial={{ openshift_master_config_dir }}/ca.serial.txt
-    --hostnames="{{ docker_registry_service_ip.stdout }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}"
+    --hostnames="{{ docker_registry_service_ip.results.clusterip }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}"
     --cert={{ openshift_master_config_dir }}/registry.crt
     --key={{ openshift_master_config_dir }}/registry.key
   when: False in (docker_registry_certificates_stat_result.results | default([]) | oo_collect(attribute='stat.exists') | list)
--- a/roles/openshift_hosted/meta/main.yml
+++ b/roles/openshift_hosted/meta/main.yml
@@ -14,6 +14,7 @@ galaxy_info:
 dependencies:
 - role: openshift_cli
 - role: openshift_hosted_facts
+- role: lib_openshift
 - role: openshift_projects
   openshift_projects: "{{ openshift_additional_projects | default({}) | oo_merge_dicts({'default':{'default_node_selector':''},'openshift-infra':{'default_node_selector':''},'logging':{'default_node_selector':''}}) }}"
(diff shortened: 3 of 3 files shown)


playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
```diff
@@ -46,12 +46,16 @@
 
   # Replace dc/docker-registry certificate secret contents if set.
   - block:
+    - name: Load lib_openshift modules
```
detiber: This can be moved to roles at the play level.

playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
```diff
@@ -46,12 +46,16 @@
...
-      command: >
-        {{ openshift.common.client_binary }} get service docker-registry
-        -o jsonpath='{.spec.clusterIP}'
-        --config={{ mktemp.stdout }}/admin.kubeconfig
-        -n default
+      oc_service:
+        namespace: default
+        name: docker-registry
+        state: list
+        kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
```
detiber: Should be able to omit kubeconfig here, the module should default to making a copy of the admin kubeconfig.

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -20,11 +20,11 @@
...
-  command: >
-    {{ openshift.common.client_binary }} get service docker-registry
-    -o jsonpath='{.spec.clusterIP}'
-    --config={{ openshift_hosted_kubeconfig }}
-    -n default
+  oc_service:
+    namespace: default
+    name: docker-registry
+    state: list
+    kubeconfig: "{{ openshift_hosted_kubeconfig }}"
```
detiber: Can omit kubeconfig here.

playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
```diff
@@ -46,12 +46,16 @@
 
   # Replace dc/docker-registry certificate secret contents if set.
   - block:
+    - name: Load lib_openshift modules
+      include_role:
+      name: lib_openshift
```
mtnbikenc: Needs indented

ingvagabund: done

playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
ingvagabund: What is the right place for the include_role is something we have not figured out yet. I prefer the explicit include than failing when the role is missing.

playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
ingvagabund: Is this a general rule to follow? Is this behaviour documented somewhere?

detiber: It's documented in the module docs. A recent update to the modules added the feature to copy them by default.

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -37,7 +37,7 @@
     --signer-cert={{ openshift_master_config_dir }}/ca.crt
     --signer-key={{ openshift_master_config_dir }}/ca.key
     --signer-serial={{ openshift_master_config_dir }}/ca.serial.txt
-    --hostnames="{{ docker_registry_service_ip.stdout }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}"
+    --hostnames="{{ (docker_registry_service_ip|from_json)["spec"]["clusterIP"] }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}"
```
ashcrow: :+1:

ingvagabund:
aos-ci-test

playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
ingvagabund: The default value for the ``kubeconfig`` if ``/etc/origin/master/admin.kubeconfig``. Given that, we should add additional check at the beginning of the all play making sure ``{{ openshift.common.config_base }}`` equals ``/etc/origin``. Otherwise the module will most likely fail. Unless a master node has both ``/etc/origin/master/admin.kubeconfig`` and ``/SOMEPATH/origin/master/admin.kubeconfig`` and we want the latter case. In which case it will be very hard to debug.

roles/openshift_hosted/tasks/registry/secure.yml
ingvagabund: The same here, can we be sure ``{{ openshift_hosted_kubeconfig }}`` is always ``/etc/origin``?

ingvagabund:
aos-ci-test

playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
detiber: I do not believe that we allow people to override openshift.common.config_base. The only case where config_base would not be '/etc/origin' is for old installs (3.0, maybe 3.1), but during upgrade we symlink /etc/origin to /etc/openshift. Since we wouldn't support the older versions for changes that we land in master, I do not see why we couldn't make the assumption now.

roles/openshift_hosted/tasks/registry/secure.yml
detiber: openshift_hosted_kubeconfig is currently a copy of /etc/origin/master/admin.kubeconfig that we make in roles/openshift_hosted/tasks/main.yml. With the module already making a copy of the kubeconfig provided, it doesn't really make sense to make a copy of a copy.

ingvagabund:
```
TASK [openshift_hosted : Create registry certificates if they do not exist] ****
fatal: [10.8.169.206]: FAILED! => {"failed": true, "msg": "Unexpected templating type error occurred on ({{ openshift.common.client_binary }} adm ca create-server-cert --signer-cert={{ openshift_master_config_dir }}/ca.crt --signer-key={{ openshift_master_config_dir }}/ca.key --signer-serial={{ openshift_master_config_dir }}/ca.serial.txt --hostnames=\"{{ (docker_registry_service_ip|from_json)[\"spec\"][\"clusterIP\"] }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}\" --cert={{ openshift_master_config_dir }}/registry.crt --key={{ openshift_master_config_dir }}/registry.key): expected string or buffer"}
```
Not even related.

ingvagabund:
aos-ci-test

detiber:
@ingvagabund are you sure that error is unrelated?
```
--hostnames=\"{{ (docker_registry_service_ip|from_json)[\"spec\"][\"clusterIP\"] }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}\"
```

ingvagabund:
> @ingvagabund are you sure that error is unrelated?
> ```
> --hostnames=\"{{ (docker_registry_service_ip|from_json)[\"spec\"][\"clusterIP\"] }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}\"
> ```

Oh man, thanks.

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
@detiber removed config from module

PR #3308, opened 2017-02-09
Modify playbooks to use oc_label module

mtnbikenc:
This will need a rebase once #3289 is merged.


ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -67,11 +67,17 @@
...
-  command: >
-    {{ hostvars[openshift_master_host].openshift.common.client_binary }} label --overwrite node {{ openshift.node.nodename }} {{ openshift.node.labels | oo_combine_dict  }}
-    --config={{ openshift_manage_node_kubeconfig }}
-    -n default
-  when: "'nodename' in openshift.node and 'labels' in openshift.node and openshift.node.labels != {}"
+  oc_label:
+    name: "{{ openshift.node.nodename }}"
+    kind: node
+    state: add
+    labels: "{{ openshift.node.kubelet_args.node-labels }}"
```
mtnbikenc: @detiber  What's the different between openshift.node.kublet_args.node-labels and openshift.node.labels?  The first is a list and the second is a dict.  I need a list so would like to use kublet_args.node-labels. Instead of taking the dict from openshift.node.labels and transforming it to a string using oo_combine_dict and then converting it to a list of oc_label, I just wanted to use the list that is already in the form needed. http://pastebin.test.redhat.com/454366

mtnbikenc:
aos-ci-test

roles/openshift_manage_node/tasks/main.yml
detiber: @mtnbikenc so, openshift.node.kubelet_args.node-labels should be a transformation applied to openshift.node.labels in openshift_facts, it was added to try and ensure that if the node is deleted/recreated then it would be recreated with the same node labels.

It was to work around an issue with cloud environments where the node object is deleted if the host is stopped/suspended.

mtnbikenc:
aos-ci-test

mtnbikenc:
aos-ci-test

mtnbikenc:
aos-ci-test

mtnbikenc:
aos-ci-test

PR #3311, opened 2017-02-09
Provisioning of nfs share and PV for logging ops

enoodle:
This remedies the problem of missing PV for the logging-ops cluster when created with nfs storage.


enoodle:
I have some problems here:
* I used the name `loggingops` instead of `logging_ops` that I would have wanted due to the nature of the facts module and the usage of the facts in `oo_filters.py` [1]
* I couldn't test this on master yet due to an unrelated problem with python-ruamel-yaml, I tested this successfully against the release-1.4 branch.

[1]https://github.com/openshift/openshift-ansible/blob/master/filter_plugins/oo_filters.py#L690

sdodson:
Needs a rebase, but LGTM.

enoodle:
@sdodson PTAL

PR #3318, opened 2017-02-09
bug 1420425. Allow setting of public facing certs for kibana in opens…

jcantrill:
…hift_logging role

### DIFF
--- a/roles/openshift_logging/defaults/main.yml
+++ b/roles/openshift_logging/defaults/main.yml
@@ -27,6 +27,18 @@ openshift_logging_kibana_proxy_cpu_limit: null
 openshift_logging_kibana_proxy_memory_limit: null
 openshift_logging_kibana_replica_count: 1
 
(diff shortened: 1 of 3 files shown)


jcantrill:
aos-ci-test

roles/openshift_logging/defaults/main.yml
```diff
@@ -26,6 +26,9 @@ openshift_logging_kibana_proxy_debug: false
 openshift_logging_kibana_proxy_cpu_limit: null
 openshift_logging_kibana_proxy_memory_limit: null
 openshift_logging_kibana_replica_count: 1
+openshift_logging_kibana_cert: ""
+openshift_logging_kibana_key: ""
+openshift_logging_kibana_ca: ""
```
sdodson: Can you add a comment explaining that these should be local file paths on the control host?

jcantrill: fixed.

PR #3319, opened 2017-02-09
make sure cluster_size is an int for arith. ops

richm:
@jcantril @ewolinetz PTAL


richm:
@jcantrill PTAL

jcantrill:
aos-ci-test

PR #3321, opened 2017-02-09
WIP: oc serviceaccount now done via oc_serviceaccount module

ashcrow:


roles/nuage_master/tasks/serviceaccount.yml
```diff
@@ -1,32 +1,19 @@
...
-  changed_when: False
-
-- set_fact:
-    nuage_tmp_conf: "{{ nuage_tmp_conf_mktemp.stdout }}"
-
-- name: Copy Configuration to temporary conf
-  command: >
-    cp {{ openshift.common.config_base }}/master/admin.kubeconfig {{nuage_tmp_conf}}
-  changed_when: false
+- name: Load lib_openshift modules
```
detiber: We probably need to get consensus on this approach sooner rather than later, since it seems like the include_role/role dependency discussions are happening more frequently.

I would suggest this be moved to role dependencies here and in the openshift_manageiq

ashcrow: Fair point. @mtnbikenc / @ingvagabund we talked yesterday about the pattern of using meta when we can, otherwise falling back to include_role. Should we change from the include_role to finding which playbooks include a task file and adding a roles section?

roles/openshift_manageiq/tasks/main.yaml
```diff
@@ -1,50 +1,41 @@
 ---
+- name: Load lib_openshift modules
+  include_role:
```
kwoodson: This include role should be moved into the meta/main.yml dependency list for openshift_manageiq role.  When inside of roles, dependencies should be listed in meta/main.yml.

roles/nuage_master/tasks/serviceaccount.yml
```diff
@@ -1,32 +1,15 @@
...
-    --config={{nuage_tmp_conf}}
-    -f -
-  register: osnuage_create_service_account
-  failed_when: "'already exists' not in osnuage_create_service_account.stderr and osnuage_create_service_account.rc != 0"
-  changed_when: osnuage_create_service_account.rc == 0
+  oc_serviceaccount:
+    kubeconfig: "{{ openshift_master_config_dir }}/admin.kubeconfig"
+    name: nullage
+    namespace: default
+    state: present
```
ingvagabund: Nice catch. As we know what is inside ``nuage_service_account_config`` the generic create command can be replaced.

roles/nuage_master/tasks/serviceaccount.yml
```diff
@@ -1,32 +1,15 @@
...
-    {{ openshift.common.client_binary }} create
-    -n default
-    --config={{nuage_tmp_conf}}
-    -f -
-  register: osnuage_create_service_account
-  failed_when: "'already exists' not in osnuage_create_service_account.stderr and osnuage_create_service_account.rc != 0"
-  changed_when: osnuage_create_service_account.rc == 0
+  oc_serviceaccount:
+    kubeconfig: "{{ openshift_master_config_dir }}/admin.kubeconfig"
+    name: nullage
```
ingvagabund: s/nullage/nuage

roles/openshift_manageiq/tasks/main.yaml
```diff
@@ -1,50 +1,41 @@
...
-    cp {{ openshift.common.config_base }}/master/admin.kubeconfig {{manage_iq_tmp_conf}}
-  changed_when: false
-
 - name: Add Management Infrastructure project
   command: >
     {{ openshift.common.client_binary }} adm new-project
     management-infra
     --description="Management Infrastructure"
-    --config={{manage_iq_tmp_conf}}
+    --config={{ openshift_master_config_dir }}/admin.kubeconfig
```
ingvagabund: What was the reason this change was not done before?

roles/openshift_manageiq/tasks/main.yaml
```diff
@@ -1,50 +1,41 @@
...
-    --config={{manage_iq_tmp_conf}}
-    -f -
-  register: osmiq_create_service_account
-  failed_when: "'already exists' not in osmiq_create_service_account.stderr and osmiq_create_service_account.rc != 0"
-  changed_when: osmiq_create_service_account.rc == 0
+  oc_serviceaccount:
+    kubeconfig: "{{ openshift_master_config_dir }}/admin.kubeconfig"
+    name: management-admin
+    namespace: management-infra
+    state: present
```
ingvagabund: +1

roles/openshift_serviceaccounts/tasks/main.yml
```diff
@@ -1,21 +1,11 @@
...
-       echo {{ lookup('template', '../templates/serviceaccount.j2')
-       | from_yaml | to_json | quote }} | {{ openshift.common.client_binary }}
-       -n {{ openshift_serviceaccounts_namespace }} create -f -
-  when: item.1.rc != 0
-  with_together:
+  oc_serviceaccount:
+    name: "{{ item }}"
+    namespace: "{{ openshift_serviceaccounts_namespace }}"
+    state: present
+  with_items:
```
ingvagabund: Nicely cleaned.

roles/openshift_manageiq/tasks/main.yaml
kwoodson: oc_project.py is in PR form.  Would be great to see it used here.

roles/nuage_master/tasks/serviceaccount.yml
ashcrow: Doh! :sweat_smile:

roles/openshift_manageiq/tasks/main.yaml
ashcrow: Fair enough. There isn't one today so I'll create one.

ashcrow:
aos-ci-test

roles/openshift_manageiq/tasks/main.yaml
detiber: Oh, oc adm new-project shouldn't be run against the admin.kubeconfig, I believe that it will change the current project.

roles/openshift_manageiq/tasks/main.yaml
```diff
@@ -3,48 +3,35 @@
...
+    name: inspector-admin
+    namespace: management-infra
+    state: present
 
 - name: Create Cluster Role
   shell: >
     echo {{ manageiq_cluster_role | to_json | quote }} |
     {{ openshift.common.client_binary }} create
-    --config={{manage_iq_tmp_conf}}
+    --config={{ openshift_master_config_dir }}/admin.kubeconfig
```
detiber: We should only use the admin.kubeconfig directly in the modules where it already handles making a copy of it and using  that copy.

We've been bit in the past by commands that mutate the config and cause issues with later commands :)

roles/openshift_manageiq/tasks/main.yaml
kwoodson: We are currently copying the kubeconfig to a tmp file.  This shouldn't affect the admin.kubeconfig.

roles/openshift_manageiq/tasks/main.yaml
kwoodson: We are copying the kubeconfig into a tmpfile.  Hopefully that doesn't mutate the original.

detiber: @kwoodson indeed, I was speaking to the use of the admin kubeconfig with shell/command, which could.

roles/openshift_manageiq/tasks/main.yaml
detiber: @kwoodson This task is using command, if it were using one of the modules, I would agree.

ashcrow: @detiber in the previous code the ``manage_iq_tmp_conf`` was from this task:

```yaml
- name: Copy Configuration to temporary conf
  command: >
    cp {{ openshift.common.config_base }}/master/admin.kubeconfig {{manage_iq_tmp_conf}}
  changed_when: false
```

Shouldn't ``{{ openshift_master_config_dir }}/admin.kubeconfig`` == `` {{ openshift.common.config_base }}/master/admin.kubeconfig`` or are they different versions of admin configs?

ashcrow: I follow. Adding back in the copy for the shell commands.

ashcrow:
@ingvagabund :arrow_up:

roles/openshift_manageiq/tasks/main.yaml
```diff
@@ -19,26 +19,18 @@
...
-    --config={{manage_iq_tmp_conf}}
-    -f -
-  register: osmiq_create_service_account
-  failed_when: "'already exists' not in osmiq_create_service_account.stderr and osmiq_create_service_account.rc != 0"
-  changed_when: osmiq_create_service_account.rc == 0
+  oc_serviceaccount:
+    kubeconfig: "{{ openshift_master_config_dir }}/admin.kubeconfig"
+    name: inspector-admin
+    namespace: management-infra
+    state: present
```
detiber: Would it make sense to use a loop for these two service accounts, since the only diff is in the serviceaccount name?

ashcrow:
aos-ci-test

ashcrow:
```
[Errno 256] No more mirrors to try.\n  atomic-openshift-3.5.0.18-1.git.0.9a5d1aa.el7.x86_64: [Errno 256] No more mirrors to try.\n\n", 
```

ashcrow:
aos-ci-test

sdodson:
> [Errno 256] No more mirrors to try.\n  atomic-openshift-3.5.0.18-1.git.0.9a5d1aa.el7.x86_64: [Errno 256] No more mirrors to try.\n\n",

^^ is what it looks like when a puddle gets built the middle of a run

PR #3323, opened 2017-02-09
Bug 1420234 - illegal_argument_exception in Kibana UI.

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1420234
The problem is that the fluent.conf is missing the common data model
filter which renames the "time" field to the "@timestamp" field.

@jcantrill @ewolinetz PTAL
[test]
aos-ci-test


jcantrill:
aos-ci-test

sdodson:
CI infra problems and installation of logging isn't included anyway.

PR #3325, opened 2017-02-10
Adding oc_project to lib_openshift.

kwoodson:


roles/lib_openshift/src/class/oc_project.py
```diff
@@ -0,0 +1,183 @@
...
+    # pylint: disable=too-many-return-statements,too-many-branches
+    @staticmethod
+    def run_ansible(params, check_mode):
+        '''run the idempotent ansible code'''
+
+        _ns = None
+        if params['node_selector'] is not None:
+            _ns = ','.join(params['node_selector'])
+
+        pconfig = ProjectConfig(params['name'],
```
ashcrow: Is the use of two names accurate given ``ProjectConfig``'s initializer is

```python
    def __init__(self, rname, namespace, kubeconfig, project_options):
```

roles/lib_openshift/src/class/oc_project.py
```diff
@@ -0,0 +1,183 @@
...
+        ########
+        if state == 'absent':
+            if oadm_project.exists():
+
+                if check_mode:
+                    return {'changed': True, 'msg': 'CHECK_MODE: Would have performed a delete.'}
+
+                api_rval = oadm_project.delete()
+
+                return {'changed': True, 'results': api_rval, 'state': state}
```
ashcrow: The response from ``oadm_project.delete()`` isn't being checked to verify it did delete. I recommend checking before noting that ``changed`` is ``True``.

roles/lib_openshift/src/class/oc_project.py
```diff
@@ -0,0 +1,183 @@
+# pylint: skip-file
+# flake8: noqa
+
+
+# pylint: disable=too-many-instance-attributes
+class OCProject(OpenShiftCLI):
+    ''' Class to wrap the oc command line tools '''
```
detiber: docstring mismatch

roles/lib_openshift/src/class/oc_project.py
```diff
@@ -0,0 +1,183 @@
...
+    kind = 'namespace'
```
detiber: Is there a reason to override kind to namespace here instead of project? 

This will break the ability for non-admin users to manage their project using this module, but that would also require adding support to selectively do `oc new-project` vs `oc adm new-project`

roles/lib_openshift/src/class/oc_project.py
```diff
@@ -0,0 +1,183 @@
...
+
+        result = self.project.find_annotation("description")
+        if result != self.config.config_options['description']['value']:
+            return True
+
+        result = self.project.find_annotation("node-selector")
+        if result != self.config.config_options['node_selector']['value']:
+            return True
+
+        # Check rolebindings and policybindings
```
detiber: Is this comment a TODO or copypasta?

roles/lib_openshift/src/class/oc_project.py
detiber: It does seem odd to use the project name as the namespace considering projects/namespaces do not have a namespace attribute set in the object metadata. Using `None` might be more appropriate if possible.

roles/lib_openshift/src/lib/project.py
```diff
@@ -0,0 +1,84 @@
...
+# pylint: disable=too-many-instance-attributes
+class ProjectConfig(OpenShiftCLIConfig):
+    ''' project config object '''
+    def __init__(self, rname, namespace, kubeconfig, project_options):
+        super(ProjectConfig, self).__init__(rname, rname, kubeconfig, project_options)
+
+class Project(Yedit):
+    ''' Class to wrap the oc command line tools '''
+    annotations_path = "metadata.annotations"
+    kind = 'Service'
```
detiber: s/Service/Project/

roles/lib_openshift/src/lib/project.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+# pylint: disable=too-many-instance-attributes
+class ProjectConfig(OpenShiftCLIConfig):
+    ''' project config object '''
+    def __init__(self, rname, namespace, kubeconfig, project_options):
+        super(ProjectConfig, self).__init__(rname, rname, kubeconfig, project_options)
+
+class Project(Yedit):
+    ''' Class to wrap the oc command line tools '''
```
detiber: docstring mismatch

roles/lib_openshift/src/lib/project.py
```diff
@@ -0,0 +1,84 @@
...
+    annotations_path = "metadata.annotations"
+    kind = 'Service'
+    annotation_prefix = 'openshift.io/'
+
+    def __init__(self, content):
+        '''Service constructor'''
```
detiber: s/Service/Project/

roles/lib_openshift/src/lib/project.py
```diff
@@ -0,0 +1,84 @@
...
+        super(Project, self).__init__(content=content)
+
+    def get_annotations(self):
+        ''' get a list of ports '''
```
detiber: s/ports/annotations/

roles/lib_openshift/src/lib/project.py
```diff
@@ -0,0 +1,84 @@
...
+        return self.get(Project.annotations_path) or {}
+
+    def add_annotations(self, inc_annos):
+        ''' add a port object to the ports list '''
```
detiber: s/ports/annotations/

roles/lib_openshift/src/lib/project.py
```diff
@@ -0,0 +1,84 @@
...
+            self.put(Project.annotations_path, inc_annos)
+        else:
+            for anno in inc_annos:
+                for key, value in anno.items():
+                    annos[key] = value
+
+        return True
+
+    def find_annotation(self, key):
+        ''' find a specific port '''
```
detiber: s/port/annotation/

roles/lib_openshift/src/lib/project.py
```diff
@@ -0,0 +1,84 @@
...
+        for inc_anno in inc_anno_keys:
+            anno = self.find_annotation(inc_anno)
+            if anno:
+                del annos[Project.annotation_prefix + anno]
+                removed = True
+
+        return removed
+
+    def update_annotation(self, key, value):
+        ''' remove an annotation from a project'''
```
detiber: s/remove/update/
s/from/for/

ashcrow:

roles/lib_openshift/src/class/oc_project.py
kwoodson: Fixing

roles/lib_openshift/src/class/oc_project.py
kwoodson: removing.

roles/lib_openshift/src/lib/project.py
kwoodson: Will fix.

roles/lib_openshift/src/lib/project.py
kwoodson: Fixed.

roles/lib_openshift/src/lib/project.py
kwoodson: Fixed.

roles/lib_openshift/src/lib/project.py
kwoodson: will fix

roles/lib_openshift/src/lib/project.py
kwoodson: :Updated project namespace to None.  Will test.

roles/lib_openshift/src/class/oc_project.py
kwoodson: When modifying the project's properties, oc does not allow certain edits when modifying through 'project' but when going through 'namespace' it does allow it.  That's why we chose namespace here.

roles/lib_openshift/src/class/oc_project.py
kwoodson: Added check for failures during delete.

roles/lib_openshift/src/class/oc_project.py
detiber: We should probably get some confirmation that this wouldn't introduce some unintended behavior.

Since `oc adm new-project` is being used to create the project it might be fine. I'm not sure if there would be any other project-specific things that would effect update/delete.

@liggitt @deads2k can either of you confirm if there are any edge cases with update/delete for a project through namespace would cause issues?

deads2k: @enj had a pull to allow those who can mutate a namespace do it by mutating the project.  I don't recall its status.  As I recall, @liggitt held it on performance concerns.

enj: https://github.com/openshift/origin/pull/11647 is the pull - it is stuck in limbo waiting on a review from @deads2k and @liggitt

kwoodson:
aos-ci-test

kwoodson:
@ashcrow, mind reviewing again?

roles/lib_openshift/src/class/oc_project.py
ashcrow: :arrow_up:

kwoodson: @ashcrow  @detiber,

I can change this to None.  Its not used.

kwoodson:
@ashcrow, ran into a few issues with None.  Will fix.

ashcrow:
Looks like a unittest needs updating: https://travis-ci.org/openshift/openshift-ansible/jobs/207397025#L893-L914

kwoodson:
@ashcrow, fixed the tests and the namespacing to 'None'.  The test was an artifact of the locate_oc_binary call.

kwoodson:
aos-ci-test

roles/lib_openshift/src/class/oc_project.py
ashcrow: It looks like ``namespace`` isn't used at all. The class looks to ignore it and use ``None``. I won't block, but is there any reason to require callers to provide ``'None'``?

kwoodson:
@ashcrow, the test passes locally but not remotely.  It has to do with the order of parameters.  I'm confused why the bot doesn't pass.

ashcrow:
@kwoodson it looks like Travis is testing a different hash then your last pushed one. I'm not exactly sure why...

detiber:
@kwoodson @ashcrow: It could be because it is applying on top of master, I'm thinking a rebase might be in order.

kwoodson:
@detiber @ashcrow 

I rebased before I pushed and then generated.  This should be up-to-date.

kwoodson:
aoc-ci-test

kwoodson:
@sdodson,  Something funny going on here with travis.  These tests all pass locally and on py27 on travis.  The error for py35:
`ERROR: InvocationError: '/home/travis/build/openshift/openshift-ansible/.tox/py35-ansible22-unit/bin/pytest'`.

Can we get this merged?  Seems as though our travis needs some resiliency.
.

detiber:
@kwoodson rounding issue for the coverage failure on py35, I'd say just reduce the requirement in .coveragerc to 25 for now.

detiber:
aos-ci-test

kwoodson:
The travis tests pass.  This module is not referenced any where so test cases shouldn't be affected.

roles/lib_openshift/src/test/unit/test_oc_project.py
```diff
@@ -115,9 +115,8 @@ def test_adding_a_project(self, mock_cmd, mock_write, mock_tmpfile_copy, mock_lo
         mock_cmd.assert_has_calls([
             mock.call(['oc', 'get', 'namespace', 'operations', '-o', 'json'], None),
             mock.call(['oc', 'get', 'namespace', 'operations', '-o', 'json'], None),
-            mock.call(['oc', 'adm', 'new-project', 'operations', '--admin-role=admin',
-                       '--display-name=operations project', '--description=All things operations project',
-                       '--node-selector=ops_only=True'], None),
```
detiber: @kwoodson the reason this is failing in Travis is because of the way that `OpenShiftCLIConfig.stringify` is handling converting `_options`, the ordering of `_options` is not guaranteed to be consistent if it is changed: `https://docs.python.org/2/library/stdtypes.html#dict.items`

You could guarantee the order of arguments provided by stringify with something like this:
```
def stringify(self, key=None):
    ''' return the options hash as cli params in a string '''
    rval = []
    config_options = self.config_options
    for name in sorted(list(config_options.keys()), key):
        data = config_options[name]
        if data['include'] \
           and (data['value'] or isinstance(data['value'], int)):
            rval.append('--%s=%s' % (name.replace('_', '-'), data['value']))
    return rval
```
I added the `key=None` param, so that child classes could pass in they key param to [sorted](https://docs.python.org/3.5/library/functions.html#sorted) if they need to ensure a particular ordering of parameters.

kwoodson: @detiber, I like this change.  I think we should create a card and place these types of enhancements there.

PR #3327, opened 2017-02-10
use the correct name for the ruamel-yaml python module

ingvagabund:
SSIA

### DIFF
--- a/roles/lib_openshift/tasks/main.yml
+++ b/roles/lib_openshift/tasks/main.yml
@@ -1,12 +1,5 @@
 ---
 - name: lib_openshift ensure python-ruamel-yaml package is on target
   package:
-    name: "{{ item }}"
+    name: python-ruamel-yaml
     state: present
-  with_items:
-  - ruamel.yaml
-  - ruamel.yaml
-  - ruamel.yaml
-  - ruamel.yaml
-  - ruamel.yaml
-  - ruamel.yaml


ingvagabund:
aos-ci-test

sdodson:
Merging this to unblock other things.

sdodson:
well, nevermind, i can wait 30 minutes for the ci jobs to finish

detiber:
I think there is a package out there that provides ruamel.yaml but does not provide python-ruamel-yaml.

I ran into this when trying to use the python-ruamel-yaml package from here: https://copr.fedorainfracloud.org/coprs/g/OpenShiftOnlineOps/openshift-tools/build/507334/

sdodson:
I think that's a packaging error, I discussed that with Troy and we've since fixed that and the fedora packages don't provide that either.

sdodson:
The Fedora packages in testing provide this
```
python-ruamel.yaml = 0.12.14-3.fc24
python-ruamel.yaml(x86-64) = 0.12.14-3.fc24
python2-ruamel-yaml = 0.12.14-3.fc24
python2-ruamel-yaml(x86-64) = 0.12.14-3.fc24
```

sdodson:
oh wait, hmm

detiber:
It could be a bug with the package/dnf/yum modules as well.

ingvagabund:
The ``python-ruamel-yaml.spec`` does not declare any ``Provides: ruamel.yaml``. I am no python packaging guru to say what is the right way. Though, it seems the package should provide the virtual provides as it is the first name python developers will search for.

PR #3328, opened 2017-02-10
Removing the /usr/bin/ansible-playbook dependency in in the spec file

mwoodson:
When openshift-ansible-roles is created, the requirement of "/usr/bin/ansible-playbook" is inserted into the RPM.  This is specifying the binary, not the RPM package.  This is auto-generated by RPM looking at the file types.

This dependency is already handled by the explicit:

 Requires:      ansible

I followed this guide to attempt to remove that line.

https://fedoraproject.org/wiki/Packaging:AutoProvidesAndRequiresFiltering

### DIFF
--- a/openshift-ansible.spec
+++ b/openshift-ansible.spec
@@ -3,6 +3,10 @@
 %{!?commit:
 %global commit c64d09e528ca433832c6b6e6f5c7734a9cc8ee6f
 }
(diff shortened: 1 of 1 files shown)


tdawson:
Just did a test.  That works great.
It takes out the bad dependency without touching anything else.
:+1:

PR #3332, opened 2017-02-10
system containers: add support for HA deployments

giuseppe:


giuseppe:
aos-ci-test

playbooks/adhoc/uninstall.yml
```diff
@@ -148,6 +148,22 @@
       - vovsbr
     when: "{{ openshift_remove_all | default(true) | bool }}"
 
+  - shell: atomic uninstall "{{ item }}"-master-api
```
ashcrow: Shouldn't there be a ``when`` placed here?

playbooks/adhoc/uninstall.yml
```diff
@@ -148,6 +148,22 @@
...
+    changed_when: False
+    failed_when: False
+    with_items:
+    - openshift-enterprise
+    - atomic-enterprise
+    - origin
+
+  - shell: atomic uninstall "{{ item }}"-master-controllers
+    changed_when: False
```
ashcrow: Shouldn't there be a ``when`` placed here?

roles/openshift_master/tasks/system_container.yml
```diff
@@ -26,11 +27,51 @@
...
+- name: Update Master HA system container package
+  command: >
+    atomic containers update {{ openshift.common.service_type }}-master-{{ item }}
+  register: update_result
+  changed_when: "'Extracting' in update_result.stdout"
+  with_items:
+    - api
+    - controllers
+  when:
+    - ("master" in result.stdout)
```
ashcrow: Not an issue for this PR but ``result`` is a little confusing as it doesn't describe what is actually being checked. No change required here but it would be nice in another PR to rename ``result``.

roles/openshift_master/tasks/system_container.yml
```diff
@@ -26,11 +27,51 @@
...
+    - (openshift.common.version is defined) and (openshift.common.version == openshift_version) | bool
+    - openshift.master.ha is defined and openshift.master.ha | bool
```
ashcrow: nit: Maybe create a ``l_is_master_ha`` or something like that for use since the same check is done a few times.

roles/openshift_master/tasks/systemd_units.yml
```diff
@@ -27,7 +27,9 @@
   template:
     src: "master_docker/master.docker.service.j2"
     dest: "{{ containerized_svc_dir }}/{{ openshift.common.service_type }}-master.service"
-  when: openshift.common.is_containerized | bool and (openshift.master.ha is not defined or not openshift.master.ha | bool and not openshift.common.is_master_system_container | bool)
+  when:
+  - openshift.common.is_containerized | bool and (openshift.master.ha is not defined or not openshift.master.ha) | bool
+  - not openshift.common.is_master_system_container | bool
```
ashcrow: :+1:

roles/openshift_node/tasks/node_system_container.yml
```diff
@@ -16,14 +16,17 @@
   register: update_result
   changed_when: "'Extracting' in update_result.stdout"
   when:
-  - (openshift.common.version is defined) and (openshift.common.version == openshift_version) and ("node" in result.stdout) | bool
+  - openshift.common.version is defined
```
ashcrow: :+1:

roles/openshift_master/tasks/systemd_units.yml
```diff
@@ -37,7 +39,9 @@
   template:
     src: "{{ ha_svc_template_path }}/atomic-openshift-master-{{ item }}.service.j2"
     dest: "{{ containerized_svc_dir }}/{{ openshift.common.service_type }}-master-{{ item }}.service"
-  when: openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
+  when:
+  - openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
+  - not openshift.common.is_master_system_container | bool
```
detiber: Should probably break out the other and conditions to separate list items as well.
```
when:
- openshift.master.ha | default(false) | bool
- openshift_master_cluster_method == "native"
- not openshift.common.is_master_system_container | bool
```

I'm not even sure that we should even need to verify openshift_master_cluster_method, since we only support native now. @abutcher @sdodson: thoughts?

roles/openshift_node/tasks/node_system_container.yml
detiber: It might make sense to use the default filter and another variable to avoid having to test 'is defined'.

in vars/main.yml:
```
l_openshift_common_version: "{{ openshift.common.version | default('v0') }}"
```

roles/openshift_master/tasks/system_container.yml
```diff
@@ -26,11 +27,51 @@
...
+    - openshift.master.ha is not defined or not openshift.master.ha | bool
 
 - name: Install Master system container package
   command: >
     atomic install --system --name={{ openshift.common.service_type }}-master {{ openshift.common.system_images_registry }}/{{ openshift.master.master_system_image }}:{{ openshift_image_tag }}
   when:
     - (openshift.common.version is not defined) or (openshift.common.version != openshift_version) or ("master" not in result.stdout) | bool
+    - openshift.master.ha is not defined or not openshift.master.ha | bool
+  notify:
+    - restart master
```
giuseppe: the problem is that each of them have a different condition, and it seems I cannot `when` inside of a block.  I can try that again.

Anyway, I consider this PR in addition to the previous one a temporary solution just to get it feature complete as soon as possible so that we can figure out if there is still "something big" missing.

At that point, I'd like to rework this file completely (e.g. idempotent updates also between different versions, so that we can drop the uninstall/install logic).

I'd like to not check at all the version here, I want to be able to simply do "atomic update containers $CONTAINER --to=$VERSION" and let atomic understand when there is nothing to do.

playbooks/adhoc/uninstall.yml
ashcrow: OK :+1:

giuseppe:
I pushed two patches to refactor the code in the when conditions

aos-ci-test

giuseppe:
is the new version ok?

In the meanwhile, I opened a PR for Ansible https://github.com/ansible/ansible/pull/21366

so that in future the new module can be used to manage these containers

(The new module though depends from some features that are not part yet of an atomic release)

roles/openshift_master/tasks/system_container.yml
```diff
@@ -17,20 +17,62 @@
...
 
 - name: Uninstall Master system container package
   command: >
     atomic uninstall {{ openshift.common.service_type }}-master
   failed_when: False
   when:
     - ("master" in result.stdout)
-    - (openshift.common.version is not defined) or (openshift.common.version != openshift_version) | bool
+    - l_is_same_version
+    - not l_is_same_version
```
detiber: These conditions contradict each other.

roles/openshift_master/tasks/system_container.yml
```diff
@@ -17,20 +17,62 @@
...
+    - not l_is_ha
 
 - name: Install Master system container package
   command: >
     atomic install --system --name={{ openshift.common.service_type }}-master {{ openshift.common.system_images_registry }}/{{ openshift.master.master_system_image }}:{{ openshift_image_tag }}
   when:
-    - (openshift.common.version is not defined) or (openshift.common.version != openshift_version) or ("master" not in result.stdout) | bool
+    - not l_is_same_version or ("master" not in result.stdout) | bool
+    - not l_is_ha
```
detiber: Would it be better to refactor this to:
```
when:
- "master" in result.stdout
- not l_is_same_version
- not l_is_ha
```

roles/openshift_master/tasks/system_container.yml
```diff
@@ -17,20 +17,62 @@
...
+
+- name: Install Master HA system container package
+  command: >
+    atomic install --system --set COMMAND={{ item }} --name={{ openshift.common.service_type }}-master-{{ item }} {{ openshift.common.system_images_registry }}/{{ openshift.master.master_system_image }}:{{ openshift_image_tag }}
+  with_items:
+    - api
+    - controllers
+  when:
+    - not l_is_same_version or ("master" not in result.stdout) | bool
+    - l_is_ha
```
detiber: I would suggest:
```
when:
- 'master' in result.stdout
- not l_is_same_version
- l_is_ha
```

roles/openshift_node/tasks/node_system_container.yml
```diff
@@ -16,19 +16,21 @@
   register: update_result
   changed_when: "'Extracting' in update_result.stdout"
   when:
-  - (openshift.common.version is defined) and (openshift.common.version == openshift_version) and ("node" in result.stdout) | bool
+  - l_is_same_version
+  - ("node" in result.stdout) | bool
```
detiber: Shouldn't need the bool filter here, you should be able to just use `'node' in result.stdout`

roles/openshift_node/tasks/node_system_container.yml
```diff
@@ -16,19 +16,21 @@
...
 
 - name: Install Node system container package
   command: >
     atomic install --system --name={{ openshift.common.service_type }}-node {{ openshift.common.system_images_registry }}/{{ openshift.node.node_system_image }}:{{ openshift_image_tag }}
   register: install_node_result
   changed_when: "'Extracting' in pull_result.stdout"
   when:
-  - (openshift.common.version is not defined) or (openshift.common.version != openshift_version) or ("node" not in result.stdout) | bool
+  - not l_is_same_version or ("node" not in result.stdout) | bool
```
detiber: I'd suggest:
```
when:
- 'node' in result.stdout
- not l_is_same_version
```

roles/openshift_node/tasks/openvswitch_system_container.yml
```diff
@@ -9,28 +9,28 @@
...
-  - (openshift.common.version is not defined) or (openshift.common.version != openshift_version) and ("openvswitch" in result.stdout) | bool
+  - not l_is_same_version
+  - ("openvswitch" in result.stdout) | bool
 
 - name: Install OpenvSwitch system container package
   command: >
     atomic install --system --name=openvswitch {{ openshift.common.system_images_registry }}/{{ openshift.node.ovs_system_image }}:{{ openshift_image_tag }}
   when:
-  - (openshift.common.version is not defined) or (openshift.common.version != openshift_version) or ("openvswitch" not in result.stdout) | bool
+  - not l_is_same_version or ("openvswitch" not in result.stdout) | bool
```
detiber: Would suggest:
```
when:
- 'openvswitch' in result.stdout
- not l_is_same_version
```

roles/openshift_master/tasks/system_container.yml
```diff
@@ -17,20 +17,62 @@
   changed_when: "'Extracting' in update_result.stdout"
   when:
     - ("master" in result.stdout)
-    - (openshift.common.version is defined) and (openshift.common.version == openshift_version) | bool
+    - l_is_same_version
+    - not l_is_ha
```
detiber: I'd suggest using blocks for the following conditions if possible:
- `not l_is_ha`
- `l_is_ha`
- `'master' in result.stdout`
- `not l_is_same version`

I would also suggest using `check_master_system_container` rather than `result` for the registered variable to avoid confusion.

roles/openshift_node/tasks/node_system_container.yml
detiber: I'd suggest using blocks for the following conditions if possible:
- `'node' in result.stdout`
- `not l_is_same version`
- `openvswitch' in result.stdout`

I would also suggest using `check_master_system_container` rather than `result` for the registered variable to avoid confusion.

Should there also be separate variables for `l_node_is_same_version` and `l_openvswitch_is_same_version` ? It looks like openvswitch containers are being managed by the openshift version and not the openvswitch version.

roles/openshift_master/tasks/system_container.yml
giuseppe: the first two conditions are in OR.  Would that work using a list?

detiber: @giuseppe that is why I switched up the logic, `and 'master' in result.stdout` should be the same as `or 'master' not in result.stdout`

giuseppe: I am still not convinced how this would work.  Do not we miss to negate the two conditions in AND here?

Now it looks that we install the master system container only if "master in result.stdout' (which means 'master is already installed, so it will never install).  Previously the condition was "different version OR master not installed".

ashcrow: @giuseppe agreed. Even though having a ``(X or Y) and Z`` doesn't look very nice in ansible I believe it is technically correct for the logic you are going for.

giuseppe:
[test]

ashcrow:
aos-ci-test

PR #3333, opened 2017-02-10
More fixes for reboot/wait for hosts.

dgoodwin:
See comments in patch for why, this has been a bit messy lately but transitioning back to inventory_hostname is where we were at before and is by far the most tested path. It does technically work for ops needs, even though there is some concern it's not a guarantee. At this point I can't find anything else that will. (CC @mwoodson)

The ansible_ssh_port is not always defined, and has been moved to ansible_port now, so use that and add a default. I think it may only be defined in non-standard use cases, i.e. when the user sets it themselves.

@sdodson Requesting priority merge on this if you can, no build needed, we will carry the patch until the next build.

### DIFF
--- a/playbooks/common/openshift-master/restart_hosts.yml
+++ b/playbooks/common/openshift-master/restart_hosts.yml
@@ -7,14 +7,26 @@
   ignore_errors: true
   become: yes
 
+# WARNING: This process is riddled with weird behavior.
+
+# Workaround for https://github.com/ansible/ansible/issues/21269
+- set_fact:
+    wait_for_host: "{{ ansible_host }}"
+
+# Ansible's blog documents this *without* the port, which appears to now
+# just wait until the timeout value and then proceed without checking anything.
+# port is now required.
+#
+# However neither ansible_ssh_port or ansible_port are reliably defined, likely
+# only if overridden. Assume a default of 22.
 - name: Wait for master to restart
   local_action:
     module: wait_for
-      host="{{ ansible_host }}"
+      host="{{ wait_for_host }}"
       state=started
       delay=10
       timeout=600
-      port="{{ ansible_ssh_port }}"
+      port="{{ ansible_port | default(ansible_ssh_port | default(22,boolean=True),boolean=True) }}"
   become: no
 
 # Now that ssh is back up we can wait for API on the remote system,


playbooks/common/openshift-master/restart_hosts.yml
```diff
@@ -7,14 +7,27 @@
...
+#
+# Using local_action causes ansible_host to be set to "localhost". The bug below
+# documents a workaround but this variable does not appear to exist anymore.
+# For now it appears this must be 'inventory_hostname'. (as it was previously)
+# See: https://github.com/ansible/ansible/issues/16139
 - name: Wait for master to restart
   local_action:
     module: wait_for
-      host="{{ ansible_host }}"
+      host="{{ inventory_hostname }}"
```
detiber: Would: `hostvars.inventory_hostname.ansible_host` work in place of inventory_hostname? inventory_hostname is not guaranteed to be usable in the general case.

playbooks/common/openshift-master/restart_hosts.yml
```diff
@@ -7,14 +7,27 @@
...
       state=started
       delay=10
       timeout=600
-      port="{{ ansible_ssh_port }}"
+      port="{{ ansible_port | default(22,boolean=True) }}"
```
detiber: This is probably because of the same reason that ansible_host is being overridden, if that is the case, then I wouldn't expect the port to be set because it would be using `connection=local`.

Would it also make sense to use `ansible_ssh_port | default(ansible_port(default(22,boolean=True),boolean=True)`?

playbooks/common/openshift-master/restart_hosts.yml
dgoodwin: hostvars[inventory_hostname]['ansible_ssh_host'] is not defined, nor is ansible_host. I saw this recommendation in a bug but it does not appear valid anymore.

playbooks/common/openshift-master/restart_hosts.yml
dgoodwin: Per http://docs.ansible.com/ansible/intro_inventory.html it looks like ansible_ssh_port was deprecated in 2.0, so we should be in the clear, do you think it's still worth it?

detiber: I think it might be good to keep it in the checks, since a user could be overriding the old value. Unless setting ansible_ssh_host automatically sets ansible_host.

playbooks/common/openshift-master/restart_hosts.yml
detiber: Uggh. @sdodson we should probably track a bug on this. I'm not sure we have a really good solution here.

@dgoodwin If facts are available, you might be able to use openshift.common.hostname.

detiber: arrgh, right...  We can't rely on openshift.common.hostname, since we might not have access to the internal network...  openshift.common.public_hostname/openshift.common.public_ip isn't much more reliable there either...

Can we save the values of ansible_port and ansible_host from a run that isn't local_action?  Does this affect `delegate_to: localhost` as well?

dgoodwin:
We found a workaround and filed https://github.com/ansible/ansible/issues/21269 upstream.

PR #3334, opened 2017-02-10
Fix idempotent stuff

tbielawa:
~~For tracking changed status fixes/audits~~

*Summary of notes from down the conversation to save you some time:*

This PR encompasses

* idempotency changes for low hanging fruit I identified last week
* @mtnbikenc's 'combined refactor' branch

Before the PR was created we had numbers like these:

OK | Changed
------------ | -------------
716 | 59

With all the changes in this PR we are now looking at numbers like this:

OK | Changed
------------ | -------------
557 | 43


playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
```diff
@@ -29,12 +29,18 @@
   - name: Check available disk space for etcd backup
     shell: df --output=avail -k {{ openshift.common.data_dir }} | tail -n 1
     register: avail_disk
+    # AUDIT:changed_when: `false` because we are only inspecting
```
detiber: Are these Audit comments meant to be temporary? If they are meant to be permanent, I think I would prefer something more concise, like flake8/pylint/ansible-lint supports.

I also wonder if we can leverage a tool like [ansible-review](https://github.com/willthames/ansible-review) to automate some of this enforcement.

roles/openshift_examples/tasks/main.yml
```diff
@@ -36,6 +36,7 @@
   unarchive:
     src: "{{ copy_examples_mktemp.stdout }}/openshift-examples.tar"
     dest: "{{ examples_base }}/"
+    mode: 0755
```
detiber: Does this impact the permissions on the files unarchived and set them to 0755? if so, it might make sense to use: `mode: u=rwX,g=rX,o=rX` instead.

roles/openshift_manageiq/tasks/main.yaml
```diff
@@ -59,6 +59,9 @@
   register: oshawkular_create_cluster_role
   failed_when: "'already exists' not in oshawkular_create_cluster_role.stderr and oshawkular_create_cluster_role.rc != 0"
   changed_when: oshawkular_create_cluster_role.rc == 0
+  # AUDIT:changed_when_note: Checking the return code is insufficient
+  # here. We really need to verify the if the role even exists before
+  # we run this task.
```
detiber: Any reason why the combination of changed_when and failed_when is not sufficient here? As long as the create isn't trying to create multiple objects, the combination of changed_when/failed_when should be sufficient here.

If creating multiple objects, we would need to be concerned about testing in this way, though.

roles/openshift_examples/tasks/main.yml
tbielawa: I noticed that the previous task was creating a directory with mode `0755`. On a subsequent you would not expect to see that directory created OR CHANGED in the results. However, the directory was actually showing changed every time. The mode starts out as `0755` from that earlier task, but then the `unarchive` task comes up. Here's where it gets interesting.

Git [does not store permissions for directories](http://stackoverflow.com/a/1071274), just executable bits for flat files. When the archive is created the root directory was created with `0765` permissions, that is the mode I see for the examples directory on my local checkout. Probably a system `umask` side-effect.

When `unarchive` ran on the remote side it will change the examples destination directory to the `0765` mode.

You see the cycle now? On the next run the *Create the remote OpenShift examples directory* task runs and changes the mode back to `0755` again. Then the unarchive happens and changes the mode back to `0765`. This update to `unarchive`, the addition of `mode` is intended to prevent the constant flippy-flappy.

Ultimately how we copy the examples over could probably be reevaluated. We stopped using the synchronize module because it's realllllly slow for large numbers of files (like our examples directories). 

In an earlier upgrade run I ran **ara** against you would see results like this:

```json
    "diff": {
        "after": {
            "mode": "0755", 
            "path": "/usr/share/openshift/examples"
        }, 
        "before": {
            "mode": "0765", 
            "path": "/usr/share/openshift/examples"
        }
    }, 
```

That task no longer shows up in the **changed** results any more with the added`mode` setting.

roles/openshift_manageiq/tasks/main.yaml
tbielawa: I suppose that does work.

The audit note isn't so much about checking that the task worked correctly but more about writing the task in a more idempotent way.

The way that task is written now it will **always** run. That's OK because if the role already exists there are no side-effects, **BUT** it's still another task that has to run. We **should** expect to see this task result as `skipped` because nothing has to happen.

In conclusion, @detiber , it's an audit note for future idempotency improvements. The task works fine the way it is but this is not *the ideal way* to do this.

playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
tbielawa: These audit notes are to explain why I added `changed_when` to tasks.

Because we are committed to performing a full `changed_when` audit in the future these notes are temporary. They also help me keep track of which items I have addressed in the play run results. We will eventually remove these comments, but for now I'm not very concerned about making a perfect audit syntax. The only thing that I am sticking to now is that all the audit annotations I am adding now follow these formats:

*Annotating a manually added override:*

```
# AUDIT:changed_when: <THE VALUE> because [EXPLANATION].
```

*Changes to tasks that don't explicitly override `changed_when` while we're getting the repo cleaned up:*

```
# AUDIT:changed_when_note: Testing <THE CHANGES YOU SEE>. <BRIEF EXPLANATION>.
```

ansible.cfg
```diff
@@ -15,6 +15,7 @@ forks = 20
 host_key_checking = False
 retry_files_enabled = False
 nocows = True
+callback_whitelist = profile_tasks
```
tbielawa: I did not intend to commit this change. I'll back it out in a future commit.

tbielawa:
Merged the combined refactor stuff into my branch.

tbielawa:
Rerunning install to get the latest data then I'm going to ask for review on this. Afterwards I'm moving on to breaking the remaining `changed` tasks in the reports out into work we can distribute.

roles/openshift_examples/tasks/main.yml
detiber: @tbielawa what permissions do the files in the directory have, though? My comment about using the symbolic permissions was to prevent the files from potentially getting execute permissions, but still allow setting execute for the directories.

roles/openshift_manageiq/tasks/main.yaml
detiber: @tbielawa Yes this task will *always* run in the same way that a list task would *always* run. If the command/task already possesses some level of idempotency, why not take advantage of it?

roles/openshift_examples/tasks/main.yml
tbielawa: grrrrrrrrrrrrrrrrr. '0755' is getting applied to all directories and files under the `examples/` path. Files should be more like `0644` :confounded:

tbielawa:
PR has been updated:

* Rebased onto latest master and fixed conflicts
* Removed the `mode` changing parameters from the `unarchive` tasks to address the problem @detiber pointed out.

tbielawa:
This PR encompasses

* idempotency changes for low hanging fruit I identified last week
* @mtnbikenc's 'combined refactor' branch

Before the PR was created we had numbers like these:

OK | Changed
------------ | -------------
716 | 59

With all the changes in this PR we are now looking at numbers like this:

OK | Changed
------------ | -------------
557 | 43

When this is reviewed and merged I'll be able to break the remaining 43 changes into distributable workable tasks.

tbielawa:
aos-ci-test

roles/cockpit-ui/tasks/main.yml
```diff
@@ -1,86 +1,38 @@
...
-  changed_when: "'already exists' not in deploy_registry_console.stderr"
-  failed_when: "'already exists' not in deploy_registry_console.stderr and deploy_registry_console.rc != 0"
-  when: check_docker_registry_exists.rc == 0
-
-- name: Delete temp directory
-  file:
-    name: "{{ mktemp.stdout }}"
-    state: absent
-  changed_when: False
+- block:
```
ashcrow: This is from https://github.com/openshift/openshift-ansible/pull/3291 and is still changing to meet testing.

roles/openshift_facts/tasks/main.yml
```diff
@@ -1,52 +1,64 @@
...
-    l_is_atomic: "{{ ostree_booted.stat.exists }}"
-- set_fact:
-    l_is_containerized: "{{ (l_is_atomic | bool) or (containerized | default(false) | bool) }}"
-    l_is_openvswitch_system_container: "{{ (use_openvswitch_system_container | default(use_system_containers) | bool) }}"
-    l_is_node_system_container: "{{ (use_node_system_container | default(use_system_containers) | bool) }}"
-    l_is_master_system_container: "{{ (use_master_system_container | default(use_system_containers) | bool) }}"
-    l_is_etcd_system_container: "{{ (use_etcd_system_container | default(use_system_containers) | bool) }}"
+  # Locally setup containerized facts for now
+  - set_fact:
+      l_is_atomic: "{{ ostree_booted.stat.exists }}"
```
ashcrow: Not that this is specific to this PR but why are these ``set_fact``s separated?

playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
```diff
@@ -70,7 +71,7 @@
         --signer-cert={{ openshift.common.config_base }}/master/ca.crt
         --signer-key={{ openshift.common.config_base }}/master/ca.key
         --signer-serial={{ openshift.common.config_base }}/master/ca.serial.txt
-        --hostnames="{{ docker_registry_service_ip.results.clusterip }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}"
+        --hostnames="{{ (docker_registry_service_ip|from_json)["spec"]["clusterIP"] }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}"
```
ingvagabund: ``docker_registry_service_ip.results.clusterip`` is the correct form

roles/openshift_facts/tasks/main.yml
```diff
@@ -1,52 +1,64 @@
...
+  - name: Set repoquery command
+    set_fact:
+      repoquery_cmd: "{{ 'dnf repoquery --latest-limit 1 -d 0' if ansible_pkg_mgr == 'dnf' else 'repoquery --plugins' }}"
+
+  # This `when` allows us to skip this expensive block of tasks on
+  # subsequent calls to the `openshift_facts` role. You will notice
+  # speed-ups in proportion to the size of your cluster as this will
+  # skip all tasks on the next calls to the `openshift_facts` role.
+  when:
+  - openshift_facts_init is not defined
```
ashcrow: :+1:

roles/cockpit-ui/tasks/main.yml
```diff
@@ -1,86 +1,38 @@
...
+
+  # TODO: Need to fix the origin and enterprise templates so that they both respect IMAGE_PREFIX
+  - name: Deploy registry-console
+    command: >
+      {{ openshift.common.client_binary }} new-app --template=registry-console
+      {% if openshift_cockpit_deployer_prefix is defined  %}-p IMAGE_PREFIX="{{ openshift_cockpit_deployer_prefix }}"{% endif %}
+      {% if openshift_cockpit_deployer_prefix is defined  %}-p IMAGE_PREFIX="{{ openshift_cockpit_deployer_prefix }}"{% endif %}
+      {% if openshift_cockpit_deployer_version is defined  %}-p IMAGE_VERSION="{{ openshift_cockpit_deployer_version }}"{% endif %}
+      -p OPENSHIFT_OAUTH_PROVIDER_URL="{{ openshift.master.public_api_url }}"
+      -p REGISTRY_HOST="{{ (docker_registry_route|from_json)['spec']['host'] }}"
```
ingvagabund: @ashcrow is testing updated version right now

roles/cockpit-ui/tasks/main.yml
```diff
@@ -1,86 +1,38 @@
...
+      -p COCKPIT_KUBE_URL="https://{{ (registry_console_cockpit_kube|from_json)['spec']['host'] }}"
```
ingvagabund: @ashcrow is testing updated version right now

playbooks/common/openshift-cluster/upgrades/etcd/upgrade.yml
```diff
@@ -31,6 +31,15 @@
     # state, not manipulating anything
     changed_when: false
 
+  - name: Record containerized etcd version
+    command: runc exec etcd_container rpm -qa --qf '%{version}' etcd\*
```
ashcrow: nit: This could cause a problem if other packages that start with ``etcd`` are added. Why not: ``rpm -qa --qf '%{version}' etcd``?

detiber:
@tbielawa unless I'm missing something, looks like `mode: 0755` is still being set on the unarchive task in openshift_examples.

roles/cockpit-ui/tasks/main.yml
ashcrow: **Repo**: https://github.com/ashcrow/openshift-ansible.git
**Branch**: oc-route-module

tbielawa:
Pushed a new version. I think removing that `mode` setting got lost in all the rebasing I had to do.

roles/openshift_facts/tasks/main.yml
tbielawa: I have no idea.

playbooks/common/openshift-cluster/upgrades/etcd/upgrade.yml
tbielawa: No idea. It was there when I got here :smile_cat:

tbielawa:
```
git merge ashcrow/oc-route-module
```

Fixed conflict in `IMAGE_PREFIX` cockpit-ui.

Switched back to the correct `--hostnames` syntax with `docker_registry_service_ip.results.clusterip`

roles/openshift_facts/tasks/main.yml
tbielawa: In my testing (1m2n), this simple check increases the `skipped` count from **1123** up to **1309**. That's almost 200 skipped tasks.

tbielawa:
aos-ci-test

roles/openshift_node_upgrade/tasks/main.yml
```diff
@@ -75,3 +75,9 @@
   # so containerized services should restart quickly as well.
   retries: 24
   delay: 5
+  # AUDIT:changed_when: `false` because we are only inspecting the
+  # state of the node, we aren't changing anything (we changed node
+  # service state in the previous task). You could say we shouldn't
+  # override this because something will be changing (the state of a
+  # service), but that should be part of the last task.
+  changed_when: false
```
kwoodson: The module version would report this properly and therefore changed_when would be unnecessary.

roles/openshift_logging/tasks/stop_cluster.yaml
```diff
@@ -9,11 +9,12 @@
...
-  include: label_node.yaml
-  vars:
-    host: "{{fluentd_host}}"
-    label: "{{openshift_logging_fluentd_nodeselector.keys()[0]}}"
-    unlabel: True
+  oc_label:
+    name: "{{ fluentd_host }}"
+    kind: node
+    state: absent
+    label: "{{ openshift_logging_fluentd_nodeselector }}"
```
tbielawa: @mtnbikenc is that updated in your branches?

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -59,19 +59,29 @@
...
-  command: >
-    {{ hostvars[openshift_master_host].openshift.common.client_binary }} label --overwrite node {{ openshift.node.nodename }} {{ openshift.node.labels | oo_combine_dict  }}
-    --config={{ openshift_manage_node_kubeconfig }}
-    -n default
-  when: "'nodename' in openshift.node and 'labels' in openshift.node and openshift.node.labels != {}"
+  oc_label:
+    name: "{{ openshift.node.nodename }}"
+    kind: node
+    state: add
+    labels: "{{ openshift.node.labels | oo_combine_dict | oo_split(separator=' ') }}"
```
tbielawa: @mtnbikenc is that updated in your branches?

roles/openshift_logging/tasks/start_cluster.yaml
```diff
@@ -10,11 +10,12 @@
...
-  include: label_node.yaml
-  vars:
-    host: "{{fluentd_host}}"
-    label: "{{openshift_logging_fluentd_nodeselector.keys()[0]}}"
-    value: "{{openshift_logging_fluentd_nodeselector.values()[0]}}"
+  oc_label:
+    name: "{{ fluentd_host }}"
+    kind: node
+    state: add
+    label: "{{ openshift_logging_fluentd_nodeselector }}"
```
tbielawa: @mtnbikenc is that updated in your branches?

roles/openshift_manage_node/tasks/main.yml
```diff
@@ -59,19 +59,29 @@
...
 - name: Set node schedulability
-  command: >
-    {{ hostvars[openshift_master_host].openshift.common.client_binary }} adm manage-node {{ openshift.node.nodename }} --schedulable={{ 'true' if openshift.node.schedulable | bool else 'false' }}
-    --config={{ openshift_manage_node_kubeconfig }}
-    -n default
-  when: "'nodename' in openshift.node"
+  oadm_manage_node:
+    node: "{{ openshift.node.nodename | lower }}"
+    schedulable: "{{ 'true' if openshift.node.schedulable | bool else 'false' }}"
+    config: "{{ openshift_manage_node_kubeconfig }}"
```
ingvagabund: s/config/kubeconfig, though, it is set to ``/etc/origin/master/admin.kubeconfig`` by default so this line can be removed (unless the path is different)

ingvagabund: set to ``kubeconfig: "{{ openshift.common.config_base }}/master/admin.kubeconfig"`` in @mtnbikenc 's PR

roles/openshift_logging/tasks/start_cluster.yaml
mtnbikenc: Yep, have a new filter plugin to do this for us.

roles/openshift_logging/tasks/stop_cluster.yaml
mtnbikenc: Yep, have a new filter plugin to do this for us.

roles/openshift_manage_node/tasks/main.yml
mtnbikenc: Yep, have a new filter plugin to do this for us.

roles/openshift_manage_node/tasks/main.yml
mtnbikenc: Confirmed this is updated in my PR

PR #3340, opened 2017-02-11
Making router pods scale with infra nodes

etsauer:
This PR compares the current replica count of the router and ensures that it matches the count of nodes with the matching label. This should enable the router to scale as the infra nodes are scaled up and down.

@sdodson @abutcher

### DIFF
--- a/roles/openshift_hosted/tasks/router/router.yml
+++ b/roles/openshift_hosted/tasks/router/router.yml
@@ -86,7 +86,7 @@
     {% else %}
     get dc/router
     {% endif%}
-    --template=\\{\\{.spec.replicas\\}\\} 
+    --template=\\{\\{.spec.replicas\\}\\}
     --namespace={{ openshift.hosted.router.namespace | default('default') }}
   register: current_replicas
   when: replicas | int > 0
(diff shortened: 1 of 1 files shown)


etsauer:
bump. anyone?

detiber:
@abutcher and @kwoodson are in the process of reworking the router/registry config. 

@sdodson It might be worth it to still accept this and backport to <= 1.4/3.4

sdodson:
I'm curious the scenarios in which you've found this to be necessary, if you look at https://github.com/openshift/openshift-ansible/pull/3340/files#diff-fd7332c663fa2d9479983470772ccaf2R37 we're already figuring out the number of nodes that match your selector and then when we create the router we set the replicas. So this code shouldn't be necessary.

detiber:
@sdodson I believe the use case is scaling up/down an existing router based on the number of infra nodes.

etsauer:
@sdodson,  @detiber is correct. This additional code handles post install scale-up events.

sdodson:
aos-ci-test

sdodson:
I think the right solution to this in the future is to use daemonsets for registry and router but that may be a ways off.

detiber:
@sdodson why do you think that is a ways off? The oc adm {router,registry} commands already have support for deploying using daemonsets.

Of course migration of existing deployments to daemonsets might be a bit tricky

etsauer:
agreed. daemonsets is the better way to go. does that fit in with the overall re-work that @abutcher and @kwoodson are doing?

etsauer:
@detiber as far as migration goes, could we make it a configurable option for a while and give operators a choice to transition?

PR #3350, opened 2017-02-13
BZ1421860: increase Heapster's metric resolution to 30s

mwringe:
It looks like we also need this for 1.4

https://bugzilla.redhat.com/show_bug.cgi?id=1421860


mwringe:
@sdodson can you please take a look?

If we need to backport this, do we just need to apply this against the specific template version in master, or backport this also to some other branch?

sdodson:
@mwringe yeah, the change also needs to be made on the release-1.4 branch, can you open a PR against that branch too? There's a chance that installers will all converge back into one branch in the future so it's good that the change is made on master too.

PR #3351, opened 2017-02-13
Fix cases where child classes override OpenShiftCLI values

detiber:


detiber:
@sdodson fixes up some issues where the temporary kubeconfigs were not being used.

detiber:
aos-ci-test

PR #3355, opened 2017-02-14
Idempotency please work

tbielawa:
This is the trimmed down, self contained version of #3334.

### "AUDIT" Syntax

* **"We're cool with this"**
```
grep -r 'AUDIT:changed_when:' roles playbooks
```

* **"General thoughts around and notes"**
```
grep -r 'AUDIT:changed_when_' roles playbooks
```

This **second form** lends itself to extension. For example, you may find items like `changed_when_note`. Whereas, the former form is strict and explicitly ends with a `:` character.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
@@ -51,6 +51,9 @@
       '{"spec":{"template":{"spec":{"containers":[{"name":"router","image":"{{ router_image }}","livenessProbe":{"tcpSocket":null,"httpGet":{"path": "/healthz", "port": 1936, "host": "localhost", "scheme": "HTTP"},"initialDelaySeconds":10,"timeoutSeconds":1}}]}}}}'
       --api-version=v1
     with_items: "{{ haproxy_routers }}"
+    # AUDIT:changed_when_note: `false` not being set here. What we
+    # need to do is check the current router image version and see if
+    # this task needs to be ran.
 
   - name: Check for default registry
     command: >
@@ -66,6 +69,9 @@
       {{ oc_cmd }} patch dc/docker-registry -n default -p
       '{"spec":{"template":{"spec":{"containers":[{"name":"registry","image":"{{ registry_image }}"}]}}}}'
       --api-version=v1
+    # AUDIT:changed_when_note: `false` not being set here. What we
+    # need to do is check the current registry image version and see
+    # if this task needs to be ran.
 
 # Check for warnings to be printed at the end of the upgrade:
 - name: Check for warnings
--- a/roles/openshift_facts/tasks/main.yml
+++ b/roles/openshift_facts/tasks/main.yml
@@ -1,52 +1,64 @@
 ---
-- name: Detecting Operating System
-  stat:
-    path: /run/ostree-booted
-  register: ostree_booted
+- block:
+  - name: Detecting Operating System
+    stat:
+      path: /run/ostree-booted
+    register: ostree_booted
 
-# Locally setup containerized facts for now
-- set_fact:
-    l_is_atomic: "{{ ostree_booted.stat.exists }}"
-- set_fact:
-    l_is_containerized: "{{ (l_is_atomic | bool) or (containerized | default(false) | bool) }}"
-    l_is_openvswitch_system_container: "{{ (use_openvswitch_system_container | default(use_system_containers) | bool) }}"
-    l_is_node_system_container: "{{ (use_node_system_container | default(use_system_containers) | bool) }}"
-    l_is_master_system_container: "{{ (use_master_system_container | default(use_system_containers) | bool) }}"
-    l_is_etcd_system_container: "{{ (use_etcd_system_container | default(use_system_containers) | bool) }}"
+  # Locally setup containerized facts for now
+  - set_fact:
+      l_is_atomic: "{{ ostree_booted.stat.exists }}"
+  - set_fact:
+      l_is_containerized: "{{ (l_is_atomic | bool) or (containerized | default(false) | bool) }}"
+      l_is_openvswitch_system_container: "{{ (use_openvswitch_system_container | default(use_system_containers) | bool) }}"
+      l_is_node_system_container: "{{ (use_node_system_container | default(use_system_containers) | bool) }}"
+      l_is_master_system_container: "{{ (use_master_system_container | default(use_system_containers) | bool) }}"
+      l_is_etcd_system_container: "{{ (use_etcd_system_container | default(use_system_containers) | bool) }}"
 
-- name: Ensure various deps are installed
-  package: name={{ item }} state=present
-  with_items: "{{ required_packages }}"
-  when: not l_is_atomic | bool
+  - name: Ensure various deps are installed
+    package: name={{ item }} state=present
+    with_items: "{{ required_packages }}"
+    when: not l_is_atomic | bool
 
-- name: Gather Cluster facts and set is_containerized if needed
-  openshift_facts:
-    role: common
-    local_facts:
-      debug_level: "{{ openshift_debug_level | default(2) }}"
-      # TODO: Deprecate deployment_type in favor of openshift_deployment_type
-      deployment_type: "{{ openshift_deployment_type | default(deployment_type) }}"
-      deployment_subtype: "{{ openshift_deployment_subtype | default(None) }}"
-      cluster_id: "{{ openshift_cluster_id | default('default') }}"
-      hostname: "{{ openshift_hostname | default(None) }}"
-      ip: "{{ openshift_ip | default(None) }}"
-      is_containerized: "{{ l_is_containerized | default(None) }}"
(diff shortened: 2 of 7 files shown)


playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
```diff
@@ -51,6 +51,9 @@
       '{"spec":{"template":{"spec":{"containers":[{"name":"router","image":"{{ router_image }}","livenessProbe":{"tcpSocket":null,"httpGet":{"path": "/healthz", "port": 1936, "host": "localhost", "scheme": "HTTP"},"initialDelaySeconds":10,"timeoutSeconds":1}}]}}}}'
       --api-version=v1
     with_items: "{{ haproxy_routers }}"
+    # AUDIT:changed_when_note: `false` not being set here. What we
```
ewolinetz: is this audit note correct since we aren't using `changed_when: false` on this task?

playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
```diff
@@ -66,6 +69,9 @@
       {{ oc_cmd }} patch dc/docker-registry -n default -p
       '{"spec":{"template":{"spec":{"containers":[{"name":"registry","image":"{{ registry_image }}"}]}}}}'
       --api-version=v1
+    # AUDIT:changed_when_note: `false` not being set here. What we
```
ewolinetz: same question here regarding `changed_when` note. since this is a `command` task, won't this always return `changed_when: true` ?

roles/openshift_facts/tasks/main.yml
```diff
@@ -1,52 +1,64 @@
...
+      repoquery_cmd: "{{ 'dnf repoquery --latest-limit 1 -d 0' if ansible_pkg_mgr == 'dnf' else 'repoquery --plugins' }}"
+
+  # This `when` allows us to skip this expensive block of tasks on
+  # subsequent calls to the `openshift_facts` role. You will notice
+  # speed-ups in proportion to the size of your cluster as this will
+  # skip all tasks on the next calls to the `openshift_facts` role.
+  when:
+  - openshift_facts_init is not defined
+
+- name: Record that openshift_facts has initialized
```
ewolinetz: should this have a when on it to ensure that we've set the facts correctly first before blocking all future calls to it? or is that a low chance of happening

roles/openshift_examples/tasks/main.yml
```diff
@@ -36,6 +36,7 @@
   unarchive:
     src: "{{ copy_examples_mktemp.stdout }}/openshift-examples.tar"
     dest: "{{ examples_base }}/"
+    mode: 0755
```
tbielawa: How does this keep re-appearing. My business acumen is failing. :man_in_tuxedo:

roles/openshift_facts/tasks/main.yml
tbielawa: This will run once every time, `openshift_facts_init` isn't defined until after the block has been evaluated at least once. Inside of the block is the only place that fact is set.

playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
tbielawa: Correct. The note says we're not but maybe in the future we should look at this task more. There's a longer discussion around it and similar audit comments here: https://github.com/openshift/openshift-ansible/pull/3334#discussion_r100583870

roles/openshift_facts/tasks/main.yml
ashcrow: @ewolinetz as in something like this to act as a canary?

```yaml

when:
  - debug_level is defined
  - deployment_type is defined
  - generate_no_proxy_hosts is defined
```

playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
ewolinetz: i missed the key word 'not'... ack

playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
ashcrow: I think @tbielawa is noting this is an area that needs to be changed and, for now, it is technically a ``changed_when: true`` due to the ``patch``.

playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
ashcrow: I believe this is a note for future logic change but I could be wrong.

roles/openshift_facts/tasks/main.yml
ewolinetz: @ashcrow  yes, or maybe it can be moved up into the block so its the last task for it. that would accomplish the same thing and is probably a little cleaner

roles/openshift_facts/tasks/main.yml
```diff
@@ -1,52 +1,64 @@
...
-    l_is_etcd_system_container: "{{ (use_etcd_system_container | default(use_system_containers) | bool) }}"
+  # Locally setup containerized facts for now
+  - set_fact:
+      l_is_atomic: "{{ ostree_booted.stat.exists }}"
+  - set_fact:
+      l_is_containerized: "{{ (l_is_atomic | bool) or (containerized | default(false) | bool) }}"
+      l_is_openvswitch_system_container: "{{ (use_openvswitch_system_container | default(use_system_containers) | bool) }}"
+      l_is_node_system_container: "{{ (use_node_system_container | default(use_system_containers) | bool) }}"
+      l_is_master_system_container: "{{ (use_master_system_container | default(use_system_containers) | bool) }}"
+      l_is_etcd_system_container: "{{ (use_etcd_system_container | default(use_system_containers) | bool) }}"
```
detiber: `l_is_*_system_container` can probably be moved to the role vars rather than being set with set_fact, or it could be set using the vars attribute on the block.

roles/openshift_facts/tasks/main.yml
```diff
@@ -1,52 +1,64 @@
...
+        no_proxy: "{{ openshift_no_proxy | default(None) }}"
+        generate_no_proxy_hosts: "{{ openshift_generate_no_proxy_hosts | default(True) }}"
+        no_proxy_internal_hostnames: "{{ openshift_no_proxy_internal_hostnames | default(None) }}"
+        sdn_network_plugin_name: "{{ os_sdn_network_plugin_name | default(None) }}"
+        use_openshift_sdn: "{{ openshift_use_openshift_sdn | default(None) }}"
 
-- name: Set repoquery command
+  - name: Set repoquery command
+    set_fact:
+      repoquery_cmd: "{{ 'dnf repoquery --latest-limit 1 -d 0' if ansible_pkg_mgr == 'dnf' else 'repoquery --plugins' }}"
```
detiber: `repoquery_cmd` can be set in role vars or block vars, no need to use set_fact.

tbielawa:
Squashed this down into just 3 commits for sanity.

* Jasons comments about moving set facts out of the tasks file and into vars has been done
* Accidental `mode` attribute setting has been removed

tbielawa:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
sdodson: As long as we're going to sweep all AUDIT: in a future sprint I'm fine with leaving this, but if we're going to leave some in as notes to indicate that we believe the changed_when: false to be benign I think we should add TODO to anything where we're suggesting future fixes.

tbielawa: Yes Scott. I believe it was during the friday call that we agreed we would do a future sweet for all `changed_when` statements because we have tonnnnnnnnnnnnnns of 'em.

tbielawa: + so that's why I added audit notes.

sdodson: Yeah, I'm suggesting that if we're going to live with the AUDIT: messages for a while we need a greppable way to denote which ones mean "we're cool, trust me" versus "fix this real soon now"

tbielawa: @sdodson sounds good to me. We should do that when we begin doing the actual change sweep, IMHO. Right now the grep I use for "we're cool" is 

```
grep -r 'AUDIT:changed_when:' roles playbooks
```

And for ones which I just added thoughts around

```
grep -r 'AUDIT:changed_when_' roles playbooks
```

tbielawa:
Reverted the change that moved stuff into vars. I consider that out of scope for this PR. A new issue was created https://github.com/openshift/openshift-ansible/issues/3360 to followup on this later.

tbielawa:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
sdodson: Works for me, can you update the description of the PR with those two patterns?

PR #3358, opened 2017-02-14
Removed JGroups cert and password generation.

jpkrohling:


jpkrohling:
Unfortunately, I wasn't able to run the tests locally. I installed the packages listed on `README.md`, but it seems I'm still missing some dependency:

```
  Failed cleaning build dir for cryptography
Failed to build pycrypto cryptography
```

If the tests here fail, I'll try to figure out what's missing on my side.

mwringe:
We want to push out an update in origin-metrics so that we no longer need to create the jgroups keystore in ansible/deployer.

This seems like an incomplete PR though, as I don't see the changes to the code where where the password would be generated. If we are going to generate the password in the template, then the template needs to be updated here.

Note that this is also for post 3.5

jpkrohling:
I believe I need some assistance on this PR. I'm not sure where or how to set the `JGROUPS_PASSWORD` in a way that it's idempotent and would allow for a reinstall, at the same time.

The related PR[1] is able to read the environment variable `JGROUPS_PASSWORD`, and there's a wrapper that we can change to read the contents of a file and set this env var, if that's the desired approach. 

The problem as I see it, is that the current approach stores the passwords on a secret, along with the certificates. We'll not have this secret anymore, so, I'd either need to add a new secret to store this password, or have it set to the template only once.

Ideas? Suggestions? Thoughts?

1 - https://github.com/openshift/origin-metrics/pull/295

jpkrohling:
@mwringe , could you please re-review? I added the JGroups password generation and it seems to be working. Here's what I see when I scale up a Metrics pod, on an environment set using the playbooks:

```
2017-02-23 14:29:36,313 INFO  [org.infinispan.remoting.transport.jgroups.JGroupsTransport] (thread-2,ee,hawkular-metrics-fnrh9) ISPN000094: Received new cluster view for channel server: [hawkular-metrics-fnrh9|1] (2) [hawkular-metrics-fnrh9, hawkular-metrics-zvvhh]
```

The password also seems to be correctly generated:

```
$ oc exec -it hawkular-metrics-fnrh9 bash
bash-4.2$ cat /opt/hawkular/server.properties 
javax.net.ssl.keyStorePassword=PGoSwbGCHsu4haG
javax.net.ssl.trustStorePassword=tSFow6RHgjddnt8
jgroups.password=Omm4ppp5sgNXgooup
```

roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
```diff
@@ -24,19 +24,21 @@
...
-- name: generate password for hawkular metrics and jgroups
+- name: generate password for hawkular metrics
   local_action: copy dest="{{ local_tmp.stdout}}/{{ item }}.pwd" content="{{ 15 | oo_random_word }}"
   with_items:
   - hawkular-metrics
-  - hawkular-jgroups-keystore
+
+- name: generate password for hawkular metrics
+  shell: openssl rand -base64 512 | tr -dc A-Z-a-z-0-9 | head -c 17
+  register: jgroups_password
```
mwringe: Is there any reason why you are generating the password here instead of in the template like you have in https://github.com/openshift/origin-metrics/pull/295/files#diff-3354f5717d2fb9fe908abb8e5fd62accR35

its probably better to generate this in the template and to remove as much as possible from the ansible code

jpkrohling: I thought this wasn't a template? The other one is a "kind: Template", and I thought about doing it like we have there, but I had the impression this YAML is deployed directly (`oc create -f ...`), while the other is processed (`oc process ... | oc create -f -`))

jpkrohling: I guess my question then is: is the `hawkular_metrics_rc.j2` merged with another file, to form a template?

mwringe: Ah, I guess ansible is not using templates here anymore. Ok, doing it this way makes sense then

roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
```diff
@@ -24,19 +24,21 @@
...
 
-- name: generate password for hawkular metrics and jgroups
+- name: generate password for hawkular metrics
   local_action: copy dest="{{ local_tmp.stdout}}/{{ item }}.pwd" content="{{ 15 | oo_random_word }}"
   with_items:
   - hawkular-metrics
-  - hawkular-jgroups-keystore
+
+- name: generate password for hawkular metrics
+  shell: openssl rand -base64 512 | tr -dc A-Z-a-z-0-9 | head -c 17
```
ewolinetz: we have a filter module that we can use instead of shell, `oo_random_word`. Can this be updated to use that instead?
I believe the usage would be `"{{ 17 | oo_random_word }}"`

jpkrohling:
PR updated to use `oo_random_word`.

ewolinetz:
aos-ci-test

jpkrohling:
@ewolinetz , is this OK to merge?

sdodson:
He approved it earlier, was just waiting on tests and those are complete so I've merged it.

jpkrohling:
Thanks!

jcantrill:
@jpkrohling @sdodson We need this backed out of 1.5 and merged into 1.6 as its causing: https://bugzilla.redhat.com/show_bug.cgi?id=1428248   @sdodson  can we get a 1.6 label maybe to identify or possibly a 'do not merge' label to make it more obvious 

cc @mwringe

sdodson:
https://github.com/openshift/openshift-ansible/pull/3539

PR #3363, opened 2017-02-14
BZ1421860: increase Heapster's metric resolution to 30s

mwringe:


mwringe:
@sdodson: this is the backported version as requested in https://github.com/openshift/openshift-ansible/pull/3350#issuecomment-279813428

PR #3365, opened 2017-02-15
Update travis/tox testing

detiber:
- Add more parallelization to travis tests
- Reduce number of tox environments by grouping tests into lint/unit tests
- Add wheel to dependencies to improve caching
- Separate out test dependencies based on test type


detiber:
aos-ci-test

tox.ini
```diff
@@ -1,19 +1,31 @@
...
     -rtest-requirements.txt
-    py35-flake8: flake8-bugbear
-    ansible22: ansible~=2.2
+    lint: -rtest-requirements-lint.txt
+    unit: -rtest-requirements-unit.txt
+    py35-lint: flake8-bugbear
+    ansible22: ansible==2.2.0.0
+
+whitelist_externals =
+    echo
```
rhcarvalho: Hmm I see http://tox.readthedocs.io/en/latest/config.html#confval-whitelist_externals=MULTI-LINE-LIST

Where do we use `echo`?

.travis.yml
```diff
@@ -9,12 +9,17 @@ python:
...
   - pip install -r requirements.txt
   - pip install tox-travis
 
 script:
   # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
   - ansible-playbook --syntax-check playbooks/byo/config.yml
-  - tox
-  - cd utils && tox
+  - if [[ "${TEST_DIR}" == "root" ]]; then tox; fi
+  - if [[ "${TEST_DIR}" == "utils" ]]; then pushd utils; tox; popd; fi
```
rhcarvalho: For curiosity's sake, why did you change from `cd` to `pushd`/`popd`?

A potential problem is replacing the `&&` with `;`, pending reading the docs, I suspect the exit code of this line might be that of `popd`, which would hide test failures.

test-requirements.txt
```diff
@@ -1,11 +1,8 @@
...
-flake8-mutable
-flake8-print
-pylint
-setuptools-lint
+# General Deps
+PyOpenSSL
+
+# setup.py Deps
 PyYAML
+sh
```
detiber: Indeed, cleaning it up.

tox.ini
detiber: Missed cleaning this out when I pulled out the molecule changes, fixing.

.travis.yml
detiber: Left over from debugging where I was trying to understand the travis cache size, cleaning up.

detiber:
@rhcarvalho issues addressed.

I think there is still room for improving the test speed/performance, but I think at least breaking out the test results helps in groking the output.

detiber:
Rebased and re-pushed to incorporate lint fix affecting master (https://github.com/openshift/openshift-ansible/pull/3371)

detiber:
aos-ci-test

detiber:
aos-ci-test

detiber:
@rhcarvalho ptal

rhcarvalho:
> I think at least breaking out the test results helps in groking the output

I'll give it a shot here and see how it goes when something needs fix (test / lint / generated code).

rhcarvalho:
As a base line, `detox` took 6'38" on a VM with 2 cores, on a checkout of this PR's branch merged with master on an existing checkout. ~~I suspect it might still get lost on venvs, e.g. in `utils/.tox`~~, no.

```
(env) [vagrant@localhost openshift-ansible]$ rm -rf utils/.tox
(env) [vagrant@localhost openshift-ansible]$ time detox
...
  congratulations :)         

real    6m38.130s
user    3m1.103s
sys     4m0.961s
(env) [vagrant@localhost openshift-ansible]$ find . -path '*/bin/python' | grep -vF .tox
# nothing
```

rhcarvalho:
master (c7f83e208ee433da74b7eb26644b30684a69649b) took 10'14" in the first run (no previous `.tox`), and then 10'27" on a second run (no need to download / deps installed from cache).
Not scientific, but we can say this PR brings a ~35% speedup.

rhcarvalho:
In terms of disk usage, in master:

```
$ du -sh .tox/
671M    .tox/
```

This PR:

```
$ du -sh .tox/                                                                           
263M    .tox/
```

60% less :+1:

tox.ini
```diff
@@ -1,19 +1,30 @@
...
 envlist =
-    py{27,35}-ansible22-{pylint,unit,flake8,yamllint,generate_validation}
+    py{27,35}-ansible22-lint
+    py{27,35}-ansible22-unit
 skipsdist=True
 skip_missing_interpreters=True
 
 [testenv]
 deps =
+    wheel
```
rhcarvalho: Maybe this could also be in `requirements.txt`, so that it broadens the audience benefited by having packages installed via wheel?

tox.ini
```diff
@@ -1,19 +1,30 @@
...
+    lint: python setup.py generate_validation
+    # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
+    lint: ansible-playbook --syntax-check playbooks/byo/config.yml
     unit: nosetests
-    generate_validation: python setup.py generate_validation
+
+[travis:env]
+TEST_TYPE =
+    lint: lint
+    unit: unit
```
rhcarvalho: What do these do?

.travis.yml
```diff
@@ -9,12 +9,17 @@ python:
...
+  - TEST_DIR=root TEST_TYPE=unit
+  - TEST_DIR=utils
+
 install:
-  - pip install -r requirements.txt
   - pip install tox-travis
 
 script:
-  # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
-  - ansible-playbook --syntax-check playbooks/byo/config.yml
```
rhcarvalho: Ah, now I see this was moved inside of the "lint" steps.

.travis.yml
```diff
@@ -9,12 +9,17 @@ python:
   - "2.7"
   - "3.5"
 
+env:
+  - TEST_DIR=root TEST_TYPE=lint
+  - TEST_DIR=root TEST_TYPE=unit
+  - TEST_DIR=utils
+
 install:
-  - pip install -r requirements.txt
```
detiber: @rhcarvalho I'm open to tweaking it a bit more, but the bigger issue at play is that we have ansible in requirements.txt, but we will need the ability to template the version of ansible in tox. We aren't testing multiple versions of Ansible today, but we will definitely want to add that in the not too distant future, especially as the 2.3 release nears (maybe even before 2.2.2 releases).

test-requirements.txt
```diff
@@ -1,11 +1,7 @@
-six
-pyOpenSSL
-flake8
-flake8-mutable
-flake8-print
-pylint
-setuptools-lint
+# General Deps
```
detiber: Agreed, I just don't know of a way to reconcile having ansible in requirements.txt, but also being able to template it in tox.

tox.ini
detiber: Would we consider wheel to be a requirement for general install, though?

.travis.yml
detiber: If we do find a way to add back installing the items from requirements.txt, I think we'd want to do it in tox and not directly in .travis.yml.

tox.ini
detiber: These are used for mapping the travis matrix. tox-travis will use this to invoke the proper test envs based on the value of the TEST_TYPE env var from the test matrix definition.

test-requirements.txt
rhcarvalho:
```
$ head -1 requirements.txt 
ansible>=2.2
```

The current `requirements.txt` allows us to test any version greater than or equal the minimum required version.

There is no problem with specifying more specific versions in `tox.ini`.

tox.ini
rhcarvalho: Hmm, this one is tricky. In several cases this dependency is already installed with `pip` / `virtualenv`:
https://packaging.python.org/current/#id4

So in many cases adding it to `requirements.txt` (or here) is a no-op.

I don't have a strong case for it. We can see at the build log that it is a no-op in Travis:
https://travis-ci.org/openshift/openshift-ansible/jobs/202003785#L167

tox.ini
rhcarvalho: Aha, thanks!

Would you mind me being paranoid and ask us to see a FAILED BUILD (on purpose), just to be sure that we will get build failures, with the expected messages, when there is a error?

After reading http://tox-travis.readthedocs.io/en/stable/toxenv.html#unignore-outcomes, I remember I've seen repos with CI always green even when it should be red.
The transition red->green is a lot more meaningful than seeing all green.

We can force a RED build by adding a dummy commit with mistakes, and then dropping it before merge.

.travis.yml
rhcarvalho: > For travis, we were installing packages outside of the tox venvs, and then also installing them again in each venv.

Yep, we are in agreement that ever since we migrated to `tox` this line is not useful here, because it applies to a different env.

Ideally we would be able to do as I started in the other comment, we could have `Ansible>=MIN_VERSION` in `requirements.txt`, and then have the tox config require more specific versions.

That may not work in the most straight-forward way because of https://github.com/pypa/pip/issues/988

```
$ cat requirements.txt 
Ansible>=2.1
$ pip install -r requirements.txt Ansible==2.2
Double requirement given: Ansible>=2.1 (from -r requirements.txt (line 1)) (already in Ansible==2.2, name='Ansible')
```

But if we call `pip` twice, it should work.

tox.ini
detiber: I'm reworking it to push the version dependent installs to the `commands` section to remove the conflict.

.travis.yml
```diff
@@ -9,12 +9,17 @@ python:
...
 script:
-  # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
-  - ansible-playbook --syntax-check playbooks/byo/config.yml
-  - tox
-  - cd utils && tox
+  - if [[ "${TEST_DIR}" == "root" ]]; then tox; fi
+  - if [[ "${TEST_DIR}" == "utils" ]]; then cd utils && tox; fi
+
+before_cache:
+  - rm ~/.cache/pip/log/debug.log
```
detiber: `cache` was already defined near the top of the file.

test-requirements.txt
detiber: I'm reworking this a bit, not only is specifying the exact version of ansible not really needed for the existing set of tests (it applies to the molecule testing), but even for the molecule testing, I will do a separate pip install within the `commands` section.

.travis.yml
detiber: Indeed, I'm removing the The Ansible install directly from the tox deps for now, but when I add it back for the molecule testing, I will just do a `pip install` from the `commands` section.

detiber:
@rhcarvalho 

> @detiber given that pylint is by far the slowest thing we run, would it make sense to group envs as "fast" and "slow" instead of "lint" and "unit"? 

Possibly, I think once the functional testing comes in with molecule and other functional testing we'll definitely want to figure out a way to group them in a fast/slow way. 

> "unit" is really clear, though "lint" seems like a generic bag of everything else.

I'm not sure it's worth breaking apart the linters. The one exception is `generate_validation`, which isn't really a linter. I just really didn't want to create a separate env for it, and it isn't a unit test.

tox.ini
detiber: @rhcarvalho test failure introduced here: https://travis-ci.org/openshift/openshift-ansible/builds/203025123

rhcarvalho:
> I'm not sure it's worth breaking apart the linters. The one exception is generate_validation, which isn't really a linter. I just really didn't want to create a separate env for it, and it isn't a unit test.

Yeah, thus my suggestion to call it "slow" and "fast" :-)
Not a big deal, but naming does help communicating intent.

tox.ini
rhcarvalho: Thanks. Would you reconsider removing the "verbosity=2" from the config? There are about ~400 lines of output with long and unreadable test names that IMHO are worth nothing. I've been running `nosetests -q` and `pytest -q` locally and that gives just what we need to see - what is bad and needs to be addressed. I'd leave normal verbosity level in CI (no `-q`, no `-v`).

rhcarvalho: I've made the output less verbose in https://github.com/openshift/openshift-ansible/pull/3419

detiber: No issues with reducing the verbosity.

detiber:
At see point we will probably want to see about either consolidating python linting to pylint or flake8, or tweaking configs to reduce the test overlap between the two.

Flake8 is faster and with some additional plugins would probably get us 90-95% of the coverage we have with both tools today, but pylint does provide tests that aren't possible with flake8. Unfortunately, those tests load and evaluate portions of the code, which contributes to the slowness.

detiber:
@rhcarvalho rebased on top of your pytest changes. There are a couple of test failures I haven't dug into yet.

setup.py
```diff
@@ -228,6 +228,14 @@ def run(self):
         print("Unsupported command for openshift-ansible")
 
 
+def extract_requirements(filename):
+    """
+    Extracts requirements from a pip formatted requirements file.
+    """
+    with open(filename, 'r') as requirements_file:
+        return [x.split()[0] for x in requirements_file.readlines()]
```
rhcarvalho: I'm wondering if this function is worth it.
Might work for our specific files now, but doesn't work for arbitrary `requirement.txt` files. They can contain version numbers, `-e ~/path/to/sth`, etc.

utils/tox.ini
```diff
@@ -17,3 +17,18 @@ commands =
     unit: pytest {posargs}
     flake8: python setup.py flake8
     pylint: python setup.py lint
+
+[pytest]
```
rhcarvalho: needs a rebase, this went into `setup.cfg` as you suggested (to keep test/lint related config in one file)

setup.cfg
```diff
@@ -11,7 +11,7 @@ excludes=.tox,utils,files
 lint_disable=fixme,locally-disabled,file-ignored,duplicate-code
 
 [flake8]
-exclude=.tox/*,utils/*,inventory/*
+exclude=.tox/*,utils/*,inventory/*,roles/lib_utils/src/*,roles/lib_openshift/src/*
```
rhcarvalho: This may suffer from the same specificity x maintainability we discussed w.r.t. test discovery.

What is `inventory/*,roles/lib_utils/src/*,roles/lib_openshift/src/*` giving us (in terms of speed, false positives, etc)?

I do understand that getting a flake8 error in both src and then 20-something generated files is unnecessarily confusing. The trouble is if we get it only on the 20 generated files, it is less convenient to find the source of the error and fix it in the source, then regenerate.

Imagine, the report will say a file name and line, but you actually need to go edit a different file in a completely different line... #contributor-experience :-)

setup.py
detiber: We don't even use it for normal testing, since we are doing the following in tox:
- setting skipdist=true without setting usedevelop=true, so we don't do an install step
- we are explicitly installing requirements.txt, setup-requirements.txt and test-requirements.txt

I'm good with dropping this, but I did like the idea of supporting the setuptools workflow if someone tried it. It would be nice to leverage setuptools for managing the dependencies as well, but I was not able to get it to use the cached packages instead of building them.

@ashcrow thoughts?

setup.cfg
detiber: > This may suffer from the same specificity x maintainability we discussed w.r.t. test discovery.

> What is inventory/*,roles/lib_utils/src/*,roles/lib_openshift/src/* giving us (in terms of speed, false positives, etc)?

I added roles/lib_{openshift,utils}/src/* to avoid the need to add `# flake8: noqa` and `# noqa: E302,E301 at the top of every file. 

If we can make flake8 (and ideally pylint as well) work against the src files, I would definitely prefer that than the current situation.

> I do understand that getting a flake8 error in both src and then 20-something generated files is unnecessarily confusing. The trouble is if we get it only on the 20 generated files, it is less convenient to find the source of the error and fix it in the source, then regenerate.

I'm not sure we get away from that, since we are just blanket ignoring the files currently.

> Imagine, the report will say a file name and line, but you actually need to go edit a different file in a completely different line... #contributor-experience :-)

Agreed, which is why I would like to see if we can rework the code generation in a way that the src files could be tested independently.

setup.py
ashcrow: It's fine to drop with no one else needs the workflow. It tripped me up when I joined hence why I added it.

rhcarvalho: :+1: to simplifying and promoting a standard workflow.

tox.ini
```diff
@@ -1,21 +1,26 @@
...
 minversion=2.3.1
 envlist =
-    py{27,35}-ansible22-{pylint,unit,flake8,yamllint,generate_validation}
+    py{27,35}-{lint,unit}
 skipsdist=True
 skip_missing_interpreters=True
 
 [testenv]
 deps =
+    -rrequirements.txt
```
rhcarvalho: This is fine, but it occurred to me that we include `requirements.txt` in `test-requirements.txt`, so that the experience involves a single `pip install`:

```
(tmp-bc3cc5ce42c72db) ✔ ~/.virtualenvs/tmp-bc3cc5ce42c72db
13:34 $ cat requirements.txt 
PyYAML
(tmp-bc3cc5ce42c72db) ✔ ~/.virtualenvs/tmp-bc3cc5ce42c72db 
13:35 $ cat test-requirements.txt 
-r requirements.txt
pytest
(tmp-bc3cc5ce42c72db) ✔ ~/.virtualenvs/tmp-bc3cc5ce42c72db 
13:35 $ pip install -r test-requirements.txt
Collecting PyYAML (from -r requirements.txt (line 1))
Collecting pytest (from -r test-requirements.txt (line 2))
  Using cached pytest-3.0.6-py2.py3-none-any.whl
Requirement already satisfied: setuptools in ./lib/python3.5/site-packages (from pytest->-r test-requirements.txt (line 2))
Collecting py>=1.4.29 (from pytest->-r test-requirements.txt (line 2))
  Using cached py-1.4.32-py2.py3-none-any.whl
Requirement already satisfied: appdirs>=1.4.0 in ./lib/python3.5/site-packages (from setuptools->pytest->-r test-requirements.txt (line 2))
Requirement already satisfied: six>=1.6.0 in ./lib/python3.5/site-packages (from setuptools->pytest->-r test-requirements.txt (line 2))
Requirement already satisfied: packaging>=16.8 in ./lib/python3.5/site-packages (from setuptools->pytest->-r test-requirements.txt (line 2))
Requirement already satisfied: pyparsing in ./lib/python3.5/site-packages (from packaging>=16.8->setuptools->pytest->-r test-requirements.txt (line 2))
Installing collected packages: PyYAML, py, pytest
Successfully installed PyYAML-3.12 py-1.4.32 pytest-3.0.6
```

tox.ini
```diff
@@ -1,21 +1,26 @@
 [tox]
 minversion=2.3.1
 envlist =
-    py{27,35}-ansible22-{pylint,unit,flake8,yamllint,generate_validation}
```
rhcarvalho: :+1: to removing Ansible for now

.travis.yml
```diff
@@ -1,20 +1,26 @@
...
-  - pip install -r requirements.txt
   - pip install tox-travis
 
 script:
-  # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
-  - ansible-playbook --syntax-check playbooks/byo/config.yml
-  - tox
-  - cd utils && tox
+  - if [[ "${TEST_DIR}" == "root" ]]; then tox; fi
+  - if [[ "${TEST_DIR}" == "utils" ]]; then cd utils && tox; fi
```
rhcarvalho: I'm not able to open the Travis build logs now (infra problem), but I was wondering how this will appear.
There are 2 commands to run, and one command will always abort and the other will run tox (because we have TEST_DIR either == root or utils).

Thinking aloud... though I don't like much have a long line with an inline shell script, perhaps the two commands could be factored into one:

```yaml
script:
  - if [[ "${TEST_DIR}" == "root" ]]; then tox; else cd utils && tox; fi
```

Or even:

```yaml
env:
  - TEST_DIR=. TEST_TYPE=lint
  - TEST_DIR=. TEST_TYPE=unit
  - TEST_DIR=utils

script:
  - cd "${TEST_DIR}" && tox
```

tox.ini
ashcrow: Agreed. Shorter environment names :+1:

tox.ini
detiber: Do we really expect many users to come in and try to run and test manually? Tox already treats the deps list as a single pip install.

.travis.yml
detiber: I like the latter, tests were running good up until the last rebase, I still need to sort out why the test is failing.

detiber:
Update on the test failure, it turns out that because we weren't including requirements.txt previously we also weren't installing ruamel.yaml in the test environments. Because of that it looks like a couple of the tests for OCServiceAccountSecretTest are failing because of mocks that were expecting pyYAML.

tox.ini
rhcarvalho: Probably not, I was just thinking aloud. As I said it is good the way it is :+1:

detiber:
@rhcarvalho could use some ideas on how to fix the test failure. I think we are hitting a problem with an unordered dictionary causing the mock assert_has_calls to fail when ruamel.yaml is used: https://github.com/detiber/openshift-ansible/blob/00718703e177a63ab27e866eb2fe27cd85299154/roles/lib_openshift/src/test/unit/test_oc_serviceaccount_secret.py#L192-L196

The order of the dictionary seems to be mutating each run.

rhcarvalho:
> The order of the dictionary seems to be mutating each run.

Yep, dicts have no order. I'll have a look now.

rhcarvalho:
> when ruamel.yaml is used

BTW, now that we're adding rualmel.yaml, should we be running the tests with both that and PyYAML?

detiber:
> BTW, now that we're adding rualmel.yaml, should we be running the tests with both that and PyYAML?

Yes, I neglected to push those changes, will do that when I get back to my computer.

rhcarvalho:
@detiber @ashcrow @tbielawa 

*disclaimer*: rant on developer experience

Trying this PR out on my box, I realized that if tox is our go-to command for running tasks, while I love the changes, it gets harder to run a single thing out of the *-lint envs:

```
$ tox -l
py27-lint
py27-unit
py35-lint
py35-unit
```

Developer experience:

1. Run unit tests:

    ```
    $ tox -e py35-unit
    ```

2. Run flake8:

    ```
    $ ???
    ```

    `tox -e py35-lint` is not a great answer because it will also run `pylint` and other things, making the process much slower.

    My personal answer will be `flake8 path/to/files/I/am/interested/in`.

    Would be nice if `flake8` was exposed as a subcommand from `tox`, while still sharing an env, would that be possible? I haven't investigated.

detiber:
> Would be nice if flake8 was exposed as a subcommand from tox, while still sharing an env, would that be possible? I haven't investigated.

I've been thinking about that myself. Shared envs cause problems with detox, but if we are enabling parallelization of the tests being run, maybe not using detox is a good thing.

I think we can use env vars, but I do like the positional arg idea (if it is possible). I'll poke around with it a bit today.

rhcarvalho:
@detiber I've got a fix for the tests, will commit and push here soon :-)

rhcarvalho:
Ha, we failed on coverage in Py3.5 now:

FAIL Required test coverage of 26% not reached. Total coverage: 25.91%

rhcarvalho:
Float rounding fun, in my box `tox -e py35-unit` passes the coverage bar both on this branch and this + merge with master.

roles/lib_openshift/src/test/unit/test_oc_serviceaccount_secret.py
```diff
@@ -30,6 +30,7 @@
     import ruamel.yaml as yaml  # noqa: EF401
     YAML_TYPE = 'ruamel'
 except ImportError:
+    import yaml
     YAML_TYPE = 'pyyaml'
```
rhcarvalho: @detiber I think we should change this test to exercise both YAML libs, if we support both.
I'd be fine to open and issue and do this in a separate PR.

detiber: I'll push my commit up for it after I rebase it.

ashcrow:
@rhcarvalho / @detiber The tasks could be split back out again with an added task of something like ``tox -e combined-tests`` which would do ``lint``, ``flake8``, and ``unit``.

detiber:
> The tasks could be split back out again with an added task of something like tox -e combined-tests which would do lint, flake8, and unit.

@ashcrow they could indeed, but there is concern about the proliferation of virtualenvs and also with making the travis matrix too large. I'll play around with it a bit to see if I can come up with a good comprimise

rhcarvalho:
This has suddenly grew bigger in scope. The tox/travis improvements were good to be merged pending the coverage rounding problem, but now the diff seems much different than last time I reviewed.

@detiber could we move coveralls to a new PR?

detiber:
I'm breaking this apart into separate PRs, first is here: https://github.com/openshift/openshift-ansible/pull/3566

detiber:
breakout PR number 2: https://github.com/openshift/openshift-ansible/pull/3569

rhcarvalho:
@detiber do you think there's anything here that hasn't been merged through other PRs?

detiber:
@rhcarvalho I can take a closer look next week to verify.

michaelgugino:
Closing due to inactivity.

PR #3367, opened 2017-02-15
Add upgrade job step after the entire upgrade performs

soltysh:
@sdodson ptal

### DIFF
--- a/playbooks/byo/openshift-cluster/upgrades/v3_5/upgrade_control_plane.yml
+++ b/playbooks/byo/openshift-cluster/upgrades/v3_5/upgrade_control_plane.yml
@@ -104,3 +104,5 @@
 - include: ../../../../common/openshift-cluster/upgrades/upgrade_control_plane.yml
 
 - include: ../../../../common/openshift-cluster/upgrades/post_control_plane.yml
+
+- inlcude: ../../../../common/openshift-cluster/upgrades/v3_5/storage_upgrade.yml
--- a/playbooks/common/openshift-cluster/upgrades/v3_5/storage_upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/v3_5/storage_upgrade.yml
@@ -1,0 +1,19 @@
+---
+###############################################################################
+# Post upgrade - Upgrade job storage
+###############################################################################
+- name: Upgrade job storage
+  hosts: oo_first_master
+  roles:
+  - { role: openshift_cli }
+  vars:
+    # Another spot where we assume docker is running and do not want to accidentally trigger an unsafe
+    # restart.
+    skip_docker_role: True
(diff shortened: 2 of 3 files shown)


sdodson:
@mtnbikenc PTAL
@soltysh you said migrate storage is idempotent to start with?

playbooks/common/openshift-cluster/upgrades/v3_5/storage_upgrade.yml
```diff
@@ -0,0 +1,19 @@
...
+  vars:
+    # Another spot where we assume docker is running and do not want to accidentally trigger an unsafe
+    # restart.
+    skip_docker_role: True
+  tasks:
+  - name: Upgrade job storage
+    command: >
+      {{ openshift.common.client_binary }} adm --config={{ openshift.common.config_base }}/master/admin.kubeconfig
+      migrate storage --include=jobs --confirm
+    run_once: true
```
dgoodwin: We need a conditional here to keep it from executing on 3.4 upgrades. Technically we say that master is only 3.5 but as devs and currently in ops we have to use it for 3.4 upgrades, in which case I suspect we don't want this to run.

Example: when: openshift.common.version_gte_3_5_or_1_5

Also if it *shouldn't* be run on 3.6, we need to make that conditional somehow as well. Is this something that should be run on every upgrade going forward?

sdodson: Would ops be running playbooks/byo/openshift-cluster/upgrades/v3_5/upgrade.yml for those upgrades or would they use .../v3_4/upgrade.yml ?

soltysh:
> @soltysh you said migrate storage is idempotent to start with?

That's correct, every following call will not modify anything, so it'll be a no-op.

playbooks/byo/openshift-cluster/upgrades/v3_5/upgrade_control_plane.yml
```diff
@@ -104,3 +104,5 @@
 - include: ../../../../common/openshift-cluster/upgrades/upgrade_control_plane.yml
 
 - include: ../../../../common/openshift-cluster/upgrades/post_control_plane.yml
+
+- inlcude: ../../../../common/openshift-cluster/upgrades/v3_5/storage_upgrade.yml
```
soltysh: I don't care about nodes with this, since I'm talking with master which is responsible for storing objects. So in the situation you're talking about this should be ok.

soltysh:
@dgoodwin any other objections or this is good for merge?

detiber:
@sdodson this PR failed travis tests for yamllint...  working on a PR to fix

sdodson:
sorry about that

PR #3369, opened 2017-02-15
WIP: Register cloudfront privkey when required

ashcrow:
Further updates for https://bugzilla.redhat.com/show_bug.cgi?id=1395168

### DIFF
--- a/roles/openshift_hosted/tasks/registry/storage/s3.yml
+++ b/roles/openshift_hosted/tasks/registry/storage/s3.yml
@@ -21,13 +21,27 @@
       openshift_hosted_registry_storage_s3_cloudfront_keypairid and
       openshift_hosted_registry_storage_s3_cloudfront_privatekeyfile are required
 
-# Copy the cloudfront.pem to the host if the baseurl is given
-- name: Copy cloudfront.pem to the registry
-  copy:
-    src: "{{ openshift_hosted_registry_storage_s3_cloudfront_privatekeyfile }}"
-    dest: /etc/s3-cloudfront/cloudfront.pem
-    backup: true
-    owner: root
-    group: root
-    mode: 0600
+
+# Inject the cloudfront private key as a secret when required
+- block:
+
+    - name: Create registry secret for cloudfront
+      oc_secret:
+        state: present
+        namespace: "{{ openshift.hosted.registry.namespace | default('default') }}"
+        name: docker-registry-s3-cloudfront
+        contents:
+          path: cloudfront.pem
+          data: "{{ lookup('file', openshift_hosted_registry_storage_s3_cloudfront_privatekeyfile) }}"
+
+    - name: Add cloudfront secret to the registry deployment config
+      command: >
+        oc volume dc/docker-registry --add --name=cloudfront-vol
+        --namespace="{{ openshift.hosted.registry.namespace | default('default') }}"
+        -m /etc/origin --type=secret --secret-name=docker-registry-s3-cloudfront
+      register: cloudfront_vol_attach
+      failed_when:
+        - "'already exists' not in cloudfront_vol_attach.stderr"
+        - "cloudfront_vol_attach.rc != 0"
+
   when: openshift_hosted_registry_storage_s3_cloudfront_baseurl | default(none) is not none
--- a/roles/openshift_hosted/meta/main.yml
+++ b/roles/openshift_hosted/meta/main.yml
@@ -4,7 +4,7 @@ galaxy_info:
   description: OpenShift Embedded Router
   company: Red Hat, Inc.
   license: Apache License, Version 2.0
-  min_ansible_version: 1.9
+  min_ansible_version: 2.1
   platforms:
   - name: EL
     versions:
--- a/roles/openshift_hosted/templates/registry_config.j2
+++ b/roles/openshift_hosted/templates/registry_config.j2
@@ -78,7 +78,7 @@ middleware:
   - name: cloudfront
     options:
       baseurl: {{ openshift_hosted_registry_storage_s3_cloudfront_baseurl }}
-      privatekey: {{ openshift_hosted_registry_storage_s3_cloudfront_privatekeyfile }}
+      privatekey: /etc/origin/cloudfront.pem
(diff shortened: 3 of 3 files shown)


ashcrow:
aos-ci-test

ashcrow:
@cooktheryan test at will!

roles/openshift_hosted/tasks/registry/storage/s3.yml
```diff
@@ -21,13 +21,45 @@
...
+        mode: 0600
+
+    - name: Create registry secrets for cloudfront private key file
+      oc_secret:
+        state: present
+        namespace: openshift-infra
+        name: cloudfront
+        files:
+          - name: cloudfront
+            path: "{{ s3_cloudfront_remote_privkey_path }}"
```
detiber: Instead of creating the directory and coping the file over to the remote host, I think it might be better to use the contents parameter for oc_secret: https://github.com/openshift/openshift-ansible/blob/master/roles/lib_openshift/library/oc_secret.py#L144-L160

I think something like this would work:
```
- name: Create registry secret for cloudfront
  oc_secret:
    state: present
    namespace: openshift-infra  # is this namespace right? I think the registry is deployed to the default namespace currently
    name: docker-registry-cloudfront  # probably better to be more descriptive here
    contents:
      path: cloudfront.pem  # this isn't the full file path, just the relative file path from the mounted secret
      data: "{{ lookup('file', openshift_hosted_registry_storage_s3_cloudfront_privatekeyfile) }}"

- name: Add cloudfront secret to the registry deployment config
  command: > 
    oc volume dc/docker-registry --add --name=cloudfront-vol
    -m /etc/origin --type=secret --secret-name=docker-registry-cloudfront
  register: cloudfront_vol_attach
  failed_when: "'already exists' not in cloudfront_vol_attach.stderr and cloudfront_vol_attach.rc != 0"
```

Some other notes:
- I wonder if it is possible to use a list of conditions for `failed_when`, similar to `when`
- If the cloudfront.pem filename is not hardcoded, then you might want to use the basename filter, but I don't see a reason why we wouldn't just hardcode it, since it only exists in the registry config map and secret definitions.
- the set_fact tasks will no longer be needed (but if they were, I would suggest they be moved to role vars or the vars attribute of the block).
- No need to create the remote directory or copy the keyfile to the remote host

ashcrow: Will update based on these notes.

ashcrow: Looks like ```{{ openshift.hosted.registry.namespace | default('default') }}``` is used for the namespace elsewhere. Will run with that.

ashcrow: In theory ``failed_when`` should be able to used as a list but I'll double check (Ref: https://github.com/ansible/ansible/pull/14833)

ashcrow:
@detiber updated based one notes. PTAL.

roles/openshift_hosted/tasks/registry/storage/s3.yml
```diff
@@ -21,13 +21,26 @@
...
+        namespace: "{{ openshift.hosted.registry.namespace | default('default') }}"
+        name: docker-registry-s3-cloudfront
+        contents:
+          path: cloudfront.pem
+          data: "{{ lookup('file', openshift_hosted_registry_storage_s3_cloudfront_privatekeyfile) }}"
+
+    - name: Add cloudfront secret to the registry deployment config
+      command: >
+        oc volume dc/docker-registry --add --name=cloudfront-vol
+        -m /etc/origin --type=secret --secret-name=docker-registry-s3-cloudfront
```
detiber: It would probably be good to add `--namespace={{ openshift.hosted.registry.namespace }}` and potentially use the variable for the oc binary rather than `oc` directly.

Otherwise, lgtm (assuming testing passes)

ashcrow:
aos-ci-test

ashcrow:
@cooktheryan if the CI tests come back :+1: then it should be considered ready to test.

roles/openshift_hosted/tasks/registry/storage/s3.yml
```diff
@@ -21,13 +21,27 @@
...
+        contents:
+          path: cloudfront.pem
+          data: "{{ lookup('file', openshift_hosted_registry_storage_s3_cloudfront_privatekeyfile) }}"
+
+    - name: Add cloudfront secret to the registry deployment config
+      command: >
+        oc volume dc/docker-registry --add --name=cloudfront-vol
+        --namespace="{{ openshift.hosted.registry.namespace | default('default') }}"
+        -m /etc/origin --type=secret --secret-name=docker-registry-s3-cloudfront
+      register: cloudfront_vol_attach
```
jianlinliu: If I read the above code correctly, it will create a secret named "docker-registry-s3-cloudfront" with data "cloudfront.pem: xxxxxx", then attach it to registry. In registry container, the pem file will be located at "/etc/origin/cloudfront.pem". But according to https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/templates/registry_config.j2#L81, in registry config file, it is trying to read key file from openshift_hosted_registry_storage_s3_cloudfront_privatekeyfile, it will fail to read that once openshift_hosted_registry_storage_s3_cloudfront_privatekeyfile is not "/etc/origin/cloudfront.pem". If keep the code in this PR, need modify https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/templates/registry_config.j2#L81 to "privatekey: /etc/origin/cloudfront.pem". Am I right?

detiber: @jianlinliu I believe you are correct.

ashcrow: Updating.

ashcrow:
aos-ci-test

ashcrow:
@jianlinliu / @detiber / @cooktheryan :arrow_up:

sdodson:
Oops, merged WIP, let me know if we should revert or if a followup is needed.

PR #3370, opened 2017-02-15
Fixed issue where upgrade fails when using daemon sets (e.g. aggregated logging)

adbaldi:
The upgrade from origin version 1.3.1 to 1.4.1 fails when a node and/or master uses daemon sets. I'm betting this is true for any version upgrade since daemon sets were introduced.  

It fails as follows:

TASK [Drain Node for Kubelet upgrade] ******************************************
fatal: [master1.example.com -> master1.example.com]: FAILED! => {"changed": true, "cmd": ["oadm", "drain", "master1.example.com", "--force", "--delete-local-data"], "delta": "0:00:00.259678", "end": "2017-02-14 15:51:00.930817", "failed": true, "rc": 1, "start": "2017-02-14 15:51:00.671139", "stderr": "error: DaemonSet-managed pods (use --ignore-daemonsets to ignore): logging-fluentd-mjxkv", "stdout": "node \"master1.example.com\" already cordoned", "stdout_lines": ["node \"master1.example.com\" already cordoned"], "warnings": []}

Adding  --ignore-daemonsets to the task resolves the issue.

### DIFF
--- a/playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
+++ b/playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
@@ -43,7 +43,7 @@
 
   - name: Drain Node for Kubelet upgrade
     command: >
-      {{ openshift.common.admin_binary }} drain {{ openshift.node.nodename }} --force --delete-local-data
+      {{ openshift.common.admin_binary }} drain {{ openshift.node.nodename }} --force --delete-local-data --ignore-daemonsets
     delegate_to: "{{ groups.oo_first_master.0 }}"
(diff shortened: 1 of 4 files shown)


sdodson:
@derekwaynecarr How should we be handling daemonsets when draining nodes during and upgrade? Is the design such that we can just ignore them and then restart the node?

mtnbikenc:
aos-ci-test

sdodson:
IRC conversation confirms this is probably the best we can do right now.

detiber:
Given Derek's feedback, I think it would be a good idea to have a card to track re-assessing this approach for 1.6/3.6.

PR #3373, opened 2017-02-15
Fixing default for logging

ewolinetz:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1421563


ewolinetz:
marking as WIP so that we can also address https://bugzilla.redhat.com/show_bug.cgi?id=1419838

PR #3375, opened 2017-02-15
Revert "Add block+when skip to `openshift_facts` tasks"

abutcher:
This reverts commit 0b1c1c2c524bfb932ab0e15f63f464673c00e651.


abutcher:
aos-ci-test

smarterclayton:
Testing against this it resolved the issue I was having in #3376

sdodson:
@tbielawa FYI, if we go down this path again we should just limit it to the costly things like package installation. The actual facts stuff is very re-entrant

PR #3377, opened 2017-02-15
Adding oc_adm_router and oc_adm_registry to lib_openshift

kwoodson:


kwoodson:
@ashcrow @abutcher, I did not write test cases for these.  I did however test them in our operations playbooks and they were successful.

roles/lib_openshift/src/class/oadm_router.py
```diff
@@ -0,0 +1,437 @@
...
+    ''' Router exception'''
+    pass
+
+class RouterConfig(OpenShiftCLIConfig):
+    ''' RouterConfig is a DTO for the router.  '''
+    def __init__(self, rname, namespace, kubeconfig, router_options):
+        super(RouterConfig, self).__init__(rname, namespace, kubeconfig, router_options)
+
+class Router(OpenShiftCLI):
+    ''' Class to wrap the oc command line tools '''
```
detiber: docstring does not match class

roles/lib_openshift/src/class/oadm_router.py
```diff
@@ -0,0 +1,437 @@
...
+        '''
+        super(Router, self).__init__('default', router_config.kubeconfig, verbose)
+        self.config = router_config
+        self.verbose = verbose
+        self.router_parts = [{'kind': 'dc', 'name': self.config.name},
+                             {'kind': 'svc', 'name': self.config.name},
+                             {'kind': 'sa', 'name': self.config.config_options['service_account']['value']},
+                             {'kind': 'secret', 'name': self.config.name + '-certs'},
+                             {'kind': 'clusterrolebinding', 'name': 'router-' + self.config.name + '-role'},
+                             #{'kind': 'endpoints', 'name': self.config.name},
```
detiber: Why is endpoints commented out?

detiber: Is there a need to also track other components like configmaps?

roles/lib_openshift/src/class/oadm_router.py
```diff
@@ -0,0 +1,437 @@
...
+                            ]
+
+        self.__router_prep = None
+        self.dconfig = None
+        self.svc = None
+        self._secret = None
+        self._serviceaccount = None
+        self._rolebinding = None
```
detiber: Should secret, serviceaccount, and rolebinding be using the properties similar to dconfig and svc?

roles/lib_openshift/src/class/oadm_router.py
```diff
@@ -0,0 +1,437 @@
...
+        self.get()
+
+    @property
+    def router_prep(self):
+        ''' property deploymentconfig'''
+        if self.__router_prep == None:
+            results = self.prepare_router()
```
detiber: Having a router_prep property and a prepare_router method seems a bit confusing.

roles/lib_openshift/src/class/oadm_router.py
```diff
@@ -0,0 +1,437 @@
...
+        # Currently only deploymentconfig needs updating
+        # Verify we got a deploymentconfig
+        if not oc_objects['DeploymentConfig']['obj']:
+            return results
+
+        # results will need to get parsed here and modifications added
+        oc_objects['DeploymentConfig']['obj'] = self.add_modifications(oc_objects['DeploymentConfig']['obj'])
+
+        for oc_type in oc_objects.keys():
+            oc_objects[oc_type]['path'] = Utils.create_tmp_file_from_contents(oc_type, oc_objects[oc_type]['obj'].yaml_dict)
```
detiber: Any particular reason to use temp files here instead of keeping the data in memory and passing it into the oc tooling through stdin?

roles/lib_openshift/src/class/oadm_router.py
```diff
@@ -0,0 +1,437 @@
...
+            if part['returncode'] != 0:
+                if part.has_key('stderr') and 'not found' in part['stderr']:
+                    # the object is not there, continue
+                    continue
+
+                # something went wrong
+                return parts
+
+        # Ugly built in sleep here.
+        time.sleep(15)
```
detiber: instead of just doing a sleep here, it might make more sense to do a watch or a loop with get until the objects no longer exist.

roles/lib_openshift/src/class/oadm_router.py
```diff
@@ -0,0 +1,437 @@
...
+
+        rval = 0
+        for result in results:
+            if result['returncode'] != 0 and not 'already exist' in result['stderr']:
+                rval = result['returncode']
+
+        return {'returncode': rval, 'results': results}
+
+    def update(self):
+        '''run update for the router.  This performs a delete and then create '''
```
detiber: doing delete/create here instead of an update or patch seems like it would cause unnecessary service disruption. Unless I'm missing something even a change to the number of replicas in a router would cause the router components to be deleted and re-created. This would be a regression from current behavior related to scaling in openshift_hosted.

roles/lib_openshift/src/class/oadm_router.py
```diff
@@ -0,0 +1,437 @@
...
+    # pylint: disable=too-many-return-statements,too-many-branches
+    def needs_update(self):
+        ''' check to see if we need to update '''
+        if not self.deploymentconfig or not self.service or not self.serviceaccount or not self.secret:
+            return True
+
+        oc_objects_prep = self.prepare_router()
+
+        # Since the output from oadm_router is returned as raw
+        # we need to parse it.  The first line is the stats_password in 3.1
```
detiber: Any reason to keep support for 3.1? At this point, I highly doubt that we would backport these changes to 3.1 (backports to 3.1 are already very limited since it is not Ansible 2.x compatible).

roles/lib_openshift/src/class/oadm_router.py
```diff
@@ -0,0 +1,437 @@
...
+            if not Utils.check_def_equal(oc_objects_prep['Secret']['obj'].yaml_dict,
+                                         self.secret.yaml_dict,
+                                         skip_keys=skip,
+                                         debug=self.verbose):
+                return True
+
+        # Service:
+        #   Fix the ports to have protocol=TCP
+        for port in oc_objects_prep['Service']['obj'].get('spec.ports'):
+            port['protocol'] = 'TCP'
```
detiber: Is this still needed? Seems like a bug if it is needed and missing. If it is still needed, should it only be done in cases where it is not already set?

roles/lib_openshift/src/class/oadm_router.py
```diff
@@ -0,0 +1,437 @@
...
+            if not port.has_key('protocol'):
+                port['protocol'] = 'TCP'
+
+        #   These are different when generating
+        skip = ['dnsPolicy',
+                'terminationGracePeriodSeconds',
+                'restartPolicy', 'timeoutSeconds',
+                'livenessProbe', 'readinessProbe',
+                'terminationMessagePath', 'hostPort',
+                'defaultMode',
```
detiber: This list seems overly broad, especially considering that there have needed to be changes made to liveness/readiness probes during upgrades.

roles/lib_openshift/src/lib/rolebinding.py
```diff
@@ -0,0 +1,276 @@
+# pylint: skip-file
```
detiber: Is there a standalone module for rolebinding? Should it be added as a pre-req to this PR rather than just including part of it?

roles/lib_openshift/src/lib/rolebinding.py
```diff
@@ -0,0 +1,276 @@
...
+
+# pylint: disable=too-many-instance-attributes
+class RoleBindingConfig(object):
+    ''' Handle route options '''
+    # pylint: disable=too-many-arguments
+    def __init__(self,
+                 sname,
```
detiber: Why `sname`?

roles/lib_openshift/src/lib/volume.py
```diff
@@ -0,0 +1,36 @@
+# pylint: skip-file
+
+class Volume(object):
```
detiber: Is there a standalone module for volume? should it be a pre-req to this PR instead of bringing in just a subset of it?

roles/lib_openshift/src/lib/rolebinding.py
```diff
@@ -0,0 +1,276 @@
+# pylint: skip-file
+
+# pylint: disable=too-many-instance-attributes
+class RoleBindingConfig(object):
+    ''' Handle route options '''
```
detiber: docstring does not match class

roles/lib_openshift/src/lib/rolebinding.py
```diff
@@ -0,0 +1,276 @@
...
+    # pylint: disable=too-many-arguments
+    def __init__(self,
+                 sname,
+                 namespace,
+                 kubeconfig,
+                 group_names=None,
+                 role_ref=None,
+                 subjects=None,
+                 usernames=None):
+        ''' constructor for handling route options '''
```
detiber: docstring does not match class

roles/lib_openshift/src/lib/rolebinding.py
```diff
@@ -0,0 +1,276 @@
...
+        self.group_names = group_names
+        self.role_ref = role_ref
+        self.subjects = subjects
+        self.usernames = usernames
+        self.data = {}
+
+        self.create_dict()
+
+    def create_dict(self):
+        ''' return a service as a dict '''
```
detiber: s/service/rolebinding/

roles/lib_openshift/src/lib/rolebinding.py
```diff
@@ -0,0 +1,276 @@
...
+        self.data['metadata']['namespace'] = self.namespace
+
+        self.data['roleRef'] = self.role_ref
+        self.data['subjects'] = self.subjects
+        self.data['userNames'] = self.usernames
+
+
+# pylint: disable=too-many-instance-attributes,too-many-public-methods
+class RoleBinding(Yedit):
+    ''' Class to wrap the oc command line tools '''
```
detiber: docstring does not match class

roles/lib_openshift/src/lib/volume.py
```diff
@@ -0,0 +1,36 @@
...
+    ''' Class to wrap the oc command line tools '''
```
detiber: docstring does not match class

roles/lib_openshift/src/lib/volume.py
```diff
@@ -0,0 +1,36 @@
...
+                    "dc":  "spec.template.spec.volumes",
+                    "rc":  "spec.template.spec.volumes",
+                   }
+
+    @staticmethod
+    def create_volume_structure(volume_info):
+        ''' return a properly structured volume '''
+        volume_mount = None
+        volume = {'name': volume_info['name']}
+        if volume_info['type'] == 'secret':
```
detiber: This appears to be missing support for `configmap` as a type

detiber:
@kwoodson should these be renamed to oc_adm_router considering the deprecation of oadm?

roles/lib_openshift/src/ansible/oadm_registry.py
```diff
@@ -0,0 +1,48 @@
...
+    module = AnsibleModule(
+        argument_spec=dict(
+            state=dict(default='present', type='str',
+                       choices=['present', 'absent']),
+            debug=dict(default=False, type='bool'),
+            namespace=dict(default='default', type='str'),
+            name=dict(default=None, required=True, type='str'),
+
+            kubeconfig=dict(default='/etc/origin/master/admin.kubeconfig', type='str'),
+            credentials=dict(default='/etc/origin/master/openshift-registry.kubeconfig', type='str'),
```
kwoodson: @detiber, this would be a good fix.  I'll remove it.

roles/lib_openshift/src/ansible/oadm_registry.py
```diff
@@ -0,0 +1,48 @@
...
+            images=dict(default=None, type='str'),
+            latest_images=dict(default=False, type='bool'),
```
kwoodson: @detiber, We can remove but I would ask if customers find this option useful.  I only mapped the oc adm registry api and this was one of the options.  If we find that we aren't using it, I'm fine with removing it.

roles/lib_openshift/src/ansible/oadm_registry.py
```diff
@@ -0,0 +1,48 @@
...
+            labels=dict(default=None, type='list'),
+            ports=dict(default=['5000'], type='list'),
+            replicas=dict(default=1, type='int'),
+            selector=dict(default=None, type='str'),
+            service_account=dict(default='registry', type='str'),
+            mount_host=dict(default=None, type='str'),
+            registry_type=dict(default='docker-registry', type='str'),
```
kwoodson: I can remove this one.

roles/lib_openshift/src/ansible/oadm_registry.py
```diff
@@ -0,0 +1,48 @@
...
+            template=dict(default=None, type='str'),
+            volume=dict(default='/registry', type='str'),
```
kwoodson: Same as above, just supporting the api.

roles/lib_openshift/src/doc/registry
```diff
@@ -0,0 +1,190 @@
+# flake8: noqa
+# pylint: skip-file
+
+DOCUMENTATION = '''
+---
+module: oadm_manage_node
+short_description: Module to manage openshift nodes
+description:
+  - Manage openshift nodes programmatically.
```
kwoodson: Will fix.

roles/lib_openshift/src/doc/registry
```diff
@@ -0,0 +1,190 @@
...
+options:
```
kwoodson: Thanks, updating doc.

kwoodson: Add state.  Removed template.  Added env_vars.  Removed Force.

roles/lib_openshift/src/doc/registry
```diff
@@ -0,0 +1,190 @@
...
+    required: false
+    default: False
+    aliases: []
+  labels:
+    description:
+    - A set of labels to uniquely identify the registry and its components.
+    required: false
+    default: None
+    aliases: []
+  enforce_quota:
```
kwoodson: Added enforce_quota.

roles/lib_openshift/src/class/oadm_registry.py
```diff
@@ -0,0 +1,394 @@
...
+        self.__registry_prep = data
+
+    def force_registry_prep(self):
+        '''force a registry prep'''
+        self.registry_prep = None
+
+    def get(self):
+        ''' return the self.registry_parts '''
+        self.deploymentconfig = None
+        self.service = None
```
kwoodson: openshift_hsoted is a role and it does create the route.  This is the module that manages the creation of the svc and deployment config for registry.   The route will be handled by oc_route in the role.

The oadm registry --dry-run returns 3 objects.  Service, DeploymentConfig, and ServiceAccount.  I'm unsure as to why we aren't keeping track of the registry serviceaccount but the other two objects are managed.  The secrets, config maps, pvcs, etc aren't specific to the module and would be handled in the role.

roles/lib_openshift/src/class/oadm_registry.py
```diff
@@ -0,0 +1,394 @@
...
+            elif result['returncode'] == 0 and part['kind'] == 'svc':
+                self.service = Yedit(content=result['results'][0])
+
+        return (self.deploymentconfig, self.service)
+
+    def exists(self):
+        '''does the object exist?'''
+        self.get()
+        if self.deploymentconfig or self.service:
+            return True
```
kwoodson: Existence is just that.  Are the objects in memory?  Validation happens in a needs_update call.  That is where the idempotency happens.

roles/lib_openshift/src/class/oadm_registry.py
```diff
@@ -0,0 +1,394 @@
...
+        '''return all pods '''
+        parts = []
+        for part in self.registry_parts:
+            if not complete and part['kind'] == 'svc':
+                continue
+            parts.append(self._delete(part['kind'], part['name']))
+
+        return parts
+
+    def prep_registry(self):
```
kwoodson: I agree that it is a little confusing.  The registry_prep is a property.  This represents the output of the prep_registry.  Generally, the output is a dry-run with our desired modifications for it.  I'm open for suggestions.

ashcrow: How about changing the property to ``prepared_registry`` and then renamed ``prep_registry`` to ``_prepare_registry`` as it seems to be called internally only?

roles/lib_openshift/src/class/oadm_registry.py
```diff
@@ -0,0 +1,394 @@
...
+        ''' prepare a registry for instantiation '''
+        # In <= 3.4 credentials are used
+        # In >= 3.5 credentials are removed
```
kwoodson: @detiber, Thanks for this.  Removing credentials as this broke for 3.5 as its no longer supported.

roles/lib_openshift/src/class/oadm_registry.py
```diff
@@ -0,0 +1,394 @@
...
+                        'restartPolicy', 'timeoutSeconds',
+                        'livenessProbe', 'readinessProbe',
+                        'terminationMessagePath',
+                        'rollingParams',
+                        'securityContext',
+                        'imagePullPolicy',
+                        'protocol', # ports.portocol: TCP
+                        'type', # strategy: {'type': 'rolling'}
+                        'defaultMode', # added on secrets
+                        'activeDeadlineSeconds', # added in 1.5 for timeouts
```
kwoodson: @detiber, @abutcher 

Currently the method of determining whether updates are needed is to check all the values in the object.  Generally what happens is the generated source oc adm registry returns values that it has defined or it might not return them at all.  Another reason to skip is that metadata has attributes that are never going to be equal such as dates, object ids, etc.  Status is another section that shouldn't be compared.

These are in place because at one point or another they were either returned / not returned by oc adm --dry-run calls and cause service disruption as they are seen as differences when checking idempotency.  When a difference is found then an update needs to occur.

Our solution is to cover as many as possible but ignore the ones that aren't currently in the dry-run output.

We also provide a way to place edits (this is our work around) inside of the deploymentconfig to match the one in memory.  These come in through the `edits:` parameter.

roles/lib_openshift/src/class/oadm_registry.py
```diff
@@ -0,0 +1,394 @@
...
+            return {'changed': True, 'results': api_rval, 'state': state}
+
+        if state == 'present':
+            ########
+            # Create
+            ########
+            if not ocregistry.exists():
+
+                if check_mode:
+                    return {'changed': True, 'msg': 'CHECK_MODE: Would have performed a delete.'}
```
kwoodson: Will fix.

roles/lib_openshift/src/ansible/oadm_router.py
```diff
@@ -0,0 +1,63 @@
...
+            cert_file=dict(default=None, type='str'),
+            key_file=dict(default=None, type='str'),
+            images=dict(default=None, type='str'), #'openshift3/ose-${component}:${version}'
+            latest_images=dict(default=False, type='bool'),
+            labels=dict(default=None, type='list'),
+            ports=dict(default=['80:80', '443:443'], type='list'),
+            replicas=dict(default=1, type='int'),
+            selector=dict(default=None, type='str'),
+            service_account=dict(default='router', type='str'),
+            router_type=dict(default='haproxy-router', type='str'),
```
kwoodson: @detiber,  The router only passes the string to the oc adm command.  Does the oc adm command support more types?  If not, I'm fine dropping them.

roles/lib_openshift/src/class/oadm_router.py
kwoodson: Endpoints are managed when a service is created or deleted.  We don't explicitly want to manage endpoints as openshift does this for us.  Will remove the commented code.

oc adm router --dry-run -o json does not return any configmaps. It only returns the objects listed and therefore we only manage the ones above.

roles/lib_openshift/src/class/oadm_router.py
kwoodson: I believe they are.

roles/lib_openshift/src/class/oadm_router.py
kwoodson: The router_prep is a property and it saves the output of prepare_router.  The prepare_router puts together the dry-run output with the desired edits into an object.  To limit the number of times needed to call out and modify the output, we store that into the router_prep property.

roles/lib_openshift/src/class/oadm_router.py
kwoodson: @detiber, this was considered at one point.  A few reasons come to mind:
- temporary files make debug much easier
- As I recall, somethings require a file to exist like oc_secret.  This allows us to take in yaml as content, then be able to create those files, and then call oc secret on them.

I like the idea of avoiding files if possible.  Maybe we could add a debug which would write to file if desired.

roles/lib_openshift/src/class/oadm_router.py
kwoodson: I disagree with the regression mentioned.  The current behavior is to do nothing if there are changes.  The modules I'm adding here will at least make the comparisons and attempt to reconcile if there are changes.

I like the update or patch idea.  The difficulty comes in on how do you determine which sections to update.  We can discuss this further as I support the idea of updating instead of delete/recreate.  

As a side note, the delete/recreate is very fast.  It does cause a small window of outage but I think it was probably the quickest and easiest path.

roles/lib_openshift/src/class/oadm_router.py
kwoodson: @detiber, agreed.  Polling would be better.  In the early version of openshift, we found that if we would give openshift a few seconds to perform its actions we would be in a better state.  I think that openshift has a specific amount of time that it listens for events and performs them.  We were just trying to play nice.

roles/lib_openshift/src/class/oadm_router.py
kwoodson: @detiber, I think we can probably remove.

roles/lib_openshift/src/ansible/oadm_router.py
detiber: @kwoodson It also supports the f5 router: https://docs.openshift.com/enterprise/3.0/install_config/install/deploy_router.html#deploying-the-f5-router. I'm just not sure we'd want to expose all of the command line options unless we know that the module handles them properly.

roles/lib_openshift/src/class/oadm_registry.py
detiber: I suspect that the ServiceAccount is only created when --credentials is not specified, which would explain it's absence.

roles/lib_openshift/src/class/oadm_registry.py
detiber: It might be good to document this limitation clearly in the docs, since I think it will cause some confusion.

roles/lib_openshift/src/class/oadm_router.py
detiber: Ah, indeed. Any reason to add leading underscores to svc and dconfig, to discourage direct access without using the properties?

roles/lib_openshift/src/class/oadm_router.py
detiber: >- temporary files make debug much easier

In some cases, I would agree. But I think there are other ways that could facilitate debugging, like just dumping the datastructures back in the fail_json call, or just using a single debug log. For debugging customer issues, we would wouldn't have a good way to automatically gather the remote temp files from the user and would have to instruct them how to find and retrieve them. 

>- As I recall, somethings require a file to exist like oc_secret. This allows us to take in yaml as content, then be able to create those files, and then call oc secret on them.

That limitation is an artifact of using the `oc secrets` command in particular. If the secret was created using `oc create`, then stdin could be used.

I only brought it up because it seems like there is more serialization/deserialization being done than needed.

roles/lib_openshift/src/class/oadm_router.py
```diff
@@ -0,0 +1,437 @@
...
+
+        if edit_results and not any([res[0] for res in edit_results]):
+            return None
+
+        return deploymentconfig
+
+    def prepare_router(self):
+        '''prepare router for instantiation'''
+        # We need to create the pem file
+        router_pem = '/tmp/router.pem'
```
detiber: We need to be careful with hardcoding temporary directories/files for a couple of reasons:
- It creates a security issue (albeit with a low cvss score)
- It will cause issues with users that are required to use a tmp directory other than /tmp

Some more info from the OpenStack security guidelines: https://security.openstack.org/guidelines/dg_using-temporary-files-securely.html

roles/lib_openshift/src/class/oadm_router.py
detiber: > I disagree with the regression mentioned. The current behavior is to do nothing if there are changes. The modules I'm adding here will at least make the comparisons and attempt to reconcile if there are changes.

So, I don't think this is entirely accurate.

That is the case when re-running openshift_hosted. It does seem that I was mistaken about the number of replicas, for some reason I thought we were doing a scale operation to adjust the replica count, but I can see that we actually are not.

For upgrade, we are currently patching the dc of the router/registry to update the required items, so if we handled upgrades using these router/registry modules, it would indeed be a regression from the current behavior. 

> I like the update or patch idea. The difficulty comes in on how do you determine which sections to update. We can discuss this further as I support the idea of updating instead of delete/recreate.

There are a few different ways to tackle it, but if starting with the current implementation then the most expedient approach would depend on whether patch or update is going to be used.

For patch, a method similar to check_def_equal that returns back a minimal datastructure consisting of the `user_def` fields that differ from `result_def` instead of returing True/False. check_def_equal could most likely be updated to wrap this new method, returning False if the patch object is not `{}` or `None`. Since patch would only be called if there is a diff being applied, then the return code of patch could be used to determine success/failure of the patch.

For update, instead of creating a minimal patch you could create a copy of `result_def`. You would also need to strip out read-only fields from metadata (https://kubernetes.io/docs/api-reference/v1/definitions/#_v1_objectmeta) and the object status before applying the update.

> As a side note, the delete/recreate is very fast. It does cause a small window of outage but I think it was probably the quickest and easiest path. We might have also run into some issues with openshift querying or modifying the objects while we were inspecting them. What generally would happen is that the object ids would get incremented and by the time you would attempt a replace it would not accept the version you were adding. Thus, delete / recreate would always work here.

I think you mean resource versions instead of object ids. With the delete/recreate the object uids would change, but update/patch operations would not affect the uids.

Even if the delete/recreate is fast and the outage window is small, I personally would consider this a blocker because the outages will affect the following classes of users:
- Users that are running openshift-ansible periodically for ongoing CM. There was at least one user that was doing this in the past (I'm not sure if they still are), and I know I've helped consultants that have worked on managing their clusters using Ansible Tower and openshift-ansible.
- Users that are expecting blue/green upgrades to incur zero downtime. Since we just delivered this for 1.4/3.4, shipping a regression to this behavior (however small) seems like a step backwards.
- Users expecting certificate-redeploys to incur zero downtime.

ashcrow:

roles/lib_openshift/src/class/oadm_registry.py
kwoodson: @ashcrow, good suggestion.  Will rename.

roles/lib_openshift/src/doc/registry
```diff
@@ -0,0 +1,190 @@
...
+    required: False
+    default: 'registry'
+    aliases: []
+  tls_certificate:
+    description:
+    - An optional path to a PEM encoded certificate (which may contain the private key) for serving over TLS
+    required: false
+    default: None
+    aliases: []
+  tls_key:
```
kwoodson: Will add.

roles/lib_openshift/src/doc/registry
```diff
@@ -0,0 +1,190 @@
...
+    required: false
+    default: '/registry'
+    aliases: []
+  volume_mounts:
+    description:
+    - The volume mounts for the registry.
+    required: false
+    default: None
+    aliases: []
+  daemonset:
```
kwoodson: Will add.

roles/lib_openshift/src/doc/registry
```diff
@@ -0,0 +1,190 @@
...
+    required: False
+    default: None
+    aliases: []
+  service_account:
+    description:
+    - Name of the service account to use to run the registry pod.
+    required: False
+    default: 'registry'
+    aliases: []
+  tls_certificate:
```
kwoodson: will add.

roles/lib_openshift/src/ansible/oadm_registry.py
```diff
@@ -0,0 +1,48 @@
...
+
+def main():
+    '''
+    ansible oc module for registry
+    '''
+
+    module = AnsibleModule(
+        argument_spec=dict(
+            state=dict(default='present', type='str',
+                       choices=['present', 'absent']),
```
kwoodson: Will add a state list.  It will be just a dump of deploymentconfig and service.

roles/lib_openshift/library/oc_adm_registry.py
```diff
@@ -0,0 +1,2508 @@
...
+        curr_env.update({'KUBECONFIG': self.kubeconfig})
+        proc = subprocess.Popen(cmds,
+                                stdin=subprocess.PIPE,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE,
+                                env=curr_env)
+
+        stdout, stderr = proc.communicate(input_data)
+
+        return proc.returncode, stdout, stderr
```
ashcrow: This should be updated wit h``decode``'s for ``stdout`` and ``stderr``.

roles/lib_openshift/library/oc_adm_router.py
```diff
@@ -0,0 +1,2905 @@
...
+        curr_env.update({'KUBECONFIG': self.kubeconfig})
+        proc = subprocess.Popen(cmds,
+                                stdin=subprocess.PIPE,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE,
+                                env=curr_env)
+
+        stdout, stderr = proc.communicate(input_data)
+
+        return proc.returncode, stdout, stderr
```
ashcrow: This should be updated wit h``decode``'s for ``stdout`` and ``stderr``.

roles/lib_openshift/src/class/oadm_registry.py
```diff
@@ -0,0 +1,394 @@
...
+        rval = 0
+        for result in results:
+            if result['returncode'] != 0:
+                rval = result['returncode']
+
+
+        return {'returncode': rval, 'results': results}
+
+    def update(self):
+        '''run update for the registry.  This performs a delete and then create '''
```
kwoodson: Updated this code to do a replace instead of a delete/create.  I did not notice any issue with it.  Also updated the needs_update to save which object to update.  That way if svc didn't receive updates then it will not get replaced.

roles/lib_openshift/src/class/oadm_registry.py
```diff
@@ -0,0 +1,394 @@
...
+        for part in parts:
+            if part['returncode'] != 0:
+                if part.has_key('stderr') and 'not found' in part['stderr']:
+                    # the object is not there, continue
+                    continue
+                # something went wrong
+                return parts
+
+        # Ugly built in sleep here.
+        #time.sleep(10)
```
kwoodson: Removed.

roles/lib_openshift/src/class/oadm_registry.py
kwoodson: Will document.

roles/lib_openshift/tasks/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
 - name: lib_openshift ensure python-ruamel-yaml package is on target
   package:
-    name: python-ruamel-yaml
+    name: python2-ruamel-yaml
```
kwoodson: What are we doing in the meantime?

roles/lib_openshift/src/lib/rolebinding.py
kwoodson: There is not a standalone for rolebinding yet.  The modeling of the object exists but nothing for oc_rolebinding.

roles/lib_openshift/src/lib/rolebinding.py
kwoodson: fixed.

roles/lib_openshift/src/lib/rolebinding.py
kwoodson: I think at one point importing * from ansible had a global 'name' conflict and was causing linters to break.  Will update to name since we no longer import *.

roles/lib_openshift/src/lib/rolebinding.py
kwoodson: will fix

roles/lib_openshift/src/lib/rolebinding.py
kwoodson: will fix

roles/lib_openshift/src/lib/rolebinding.py
kwoodson: fixed

roles/lib_openshift/src/lib/volume.py
kwoodson: There is a stand alone for oc_volume.  It is on the list o things to port.  We are using its library calls under the covers.  Will port when I get a chance.

roles/lib_openshift/src/lib/volume.py
kwoodson: This is probably true.  We will need to add a card for configmap support.

roles/lib_openshift/library/oc_adm_registry.py
kwoodson: @ashcrow, will rebase.

roles/lib_openshift/library/oc_adm_router.py
kwoodson: will rebase.

roles/lib_openshift/src/ansible/oadm_router.py
kwoodson: @detiber, I'm not sure I understand.  The module doesn't do anything aside from calling the oc command and comparing dictionaries.  If oc adm router supports f5, then we should probably allow someone to pass that through to the command line.

roles/lib_openshift/src/doc/router
```diff
@@ -0,0 +1,198 @@
...
+    required: false
+    default: router
+    aliases: []
+  namespace:
+    description:
+    - The namespace where to manage the router.
+    required: false
+    default: default
+    aliases: []
+  credentials:
```
kwoodson: Removing credentials as you mentioned that openshift supports service accounts since 3.2.  Thanks for this one.

roles/lib_openshift/src/ansible/oadm_router.py
sdodson: I think we should build modules in such a way that the list of arguments doesn't change based on the value of one flag. Or put another way, we should have oadm_haproxy_router and oadm_f5_router in the future if we choose to implement that because the list of options between the two different types of routers varies. It's unfortunate that `oadm router` exhibits this behavior but I think building modules that target one implementation will keep our modules simple.

sdodson: or maybe 'openshift_haproxy_router' and 'openshift_f5_router' but that part I'm not that concerned about.

roles/lib_openshift/src/lib/volume.py
kwoodson: fixed

roles/lib_openshift/library/oc_serviceaccount.py
```diff
@@ -1308,7 +1308,7 @@ def __init__(self, sname, namespace, kubeconfig, secrets=None, image_pull_secret
         self.create_dict()
 
     def create_dict(self):
-        ''' return a properly structured volume '''
+        ''' instantiate a properly structured volume '''
```
ashcrow: Is this accurate?

roles/lib_openshift/library/oc_serviceaccount_secret.py
```diff
@@ -1308,7 +1308,7 @@ def __init__(self, sname, namespace, kubeconfig, secrets=None, image_pull_secret
         self.create_dict()
 
     def create_dict(self):
-        ''' return a properly structured volume '''
+        ''' instantiate a properly structured volume '''
```
ashcrow: Same as above.

roles/lib_openshift/src/ansible/oc_adm_router.py
```diff
@@ -0,0 +1,63 @@
...
+            external_host_private_key=dict(default=None, type='str'),
+            # Metrics
+            expose_metrics=dict(default=False, type='bool'),
+            metrics_image=dict(default=None, type='str'),
+            # Stats
+            stats_user=dict(default=None, type='str'),
+            stats_password=dict(default=None, type='str'),
+            stats_port=dict(default=1936, type='int'),
+            # extra
+            cacert_file=dict(default=None, type='str'),
```
ashcrow: Does this map to ``--default-cert=`` which is being used in ``roles/openshift_hosted/tasks/router/router.yml``?

roles/lib_openshift/src/ansible/oc_adm_registry.py
```diff
@@ -0,0 +1,47 @@
+# pylint: skip-file
+# flake8: noqa
+
+def main():
+    '''
+    ansible oc module for registry
+    '''
+
+    module = AnsibleModule(
```
ashcrow: This looks to be missing ``--credentials=`` which is used in ``roles/openshift_hosted/tasks/registry/registry.yml`` when ``not openshift.common.version_gte_3_2_or_1_2``.

roles/lib_openshift/library/oc_adm_router.py
```diff
@@ -0,0 +1,2923 @@
...
+            for sec in self.image_pull_secrets:
+                self.data['imagePullSecrets'].append({"name": sec})
+
+class ServiceAccount(Yedit):
+    ''' Class to wrap the oc command line tools '''
+    image_pull_secrets_path = "imagePullSecrets"
+    secrets_path = "secrets"
+
+    def __init__(self, content):
+        '''ServiceAccount cons
```
ashcrow: This will fail on Python 3. See https://github.com/openshift/openshift-ansible/pull/3424.

ashcrow: This will fail on Python 3. See https://github.com/openshift/openshift-ansible/pull/3424

ashcrow: This will fail on Python 3. Replace it with:

```python
if not 'protocol' in port:
```

roles/lib_openshift/library/oc_adm_registry.py
```diff
@@ -0,0 +1,2524 @@
...
+
+        return removed
+
+    def add_cluster_ip(self, sip):
+        '''add cluster ip'''
+        self.put(Service.cluster_ip, sip)
+
+    def add_portal_ip(self, pip):
+        '''add cluster ip'''
+
```
ashcrow: This will fail on Python 3. Replace it with

```python
...and 'items' in results['results']:
```

roles/lib_openshift/src/doc/registry
```diff
@@ -0,0 +1,191 @@
...
+    default: False
+    aliases: []
+author:
+- "Kenny Woodson <kwoodson@redhat.com>"
+extends_documentation_fragment: []
+'''
+
+EXAMPLES = '''
+- name: create a secure registry
+  oadm_registry:
```
mtnbikenc: Update references to oadm.

roles/lib_openshift/src/doc/router
```diff
@@ -0,0 +1,216 @@
...
+    - sessionAffinity
+    - type
+  - ServiceAccount:
+    - secrets
+    - imagePullSecrets
+'''
+
+EXAMPLES = '''
+- name: create routers
+  oadm_router:
```
mtnbikenc: Update references to oadm.

roles/lib_openshift/src/sources.yml
```diff
@@ -9,6 +9,36 @@ oadm_manage_node.py:
 - class/oadm_manage_node.py
 - ansible/oadm_manage_node.py
```
mtnbikenc: Update references to oadm.

roles/lib_openshift/library/oc_serviceaccount.py
kwoodson: hmm, it setups up all the parameters of what the object should look like.  Its sort of an instantiation

roles/lib_openshift/src/ansible/oc_adm_registry.py
kwoodson: @ashcrow, per @detiber's comments, you can pass in a --credentials or a serviceaccount since 3.2.  Jason mentioned we could just remove credentials in favor of a service account.  If that's not the case, I can re-add them.

@detiber, thoughts?

detiber: The serviceaccount is the preferred approach, since it removes the maintenance required to maintain the certs for the embedded kubeconfig, and there are no concerns with api server hostnames, etc.

detiber: @ashcrow We already have some major issues that prevent backporting changes to 3.1, I see no reason why we would need to maintain <= 3.1 support for this.

roles/lib_openshift/src/ansible/oc_adm_router.py
kwoodson: @ashcrow, good find.  Currently, the operations team's version of this builds the pem file for you since the cert and key are in separate files.  I'm going to add support for this so that we are compatible with both versions.  This should be straight forward since they both result in a single file.

roles/lib_openshift/src/doc/registry
kwoodson: will fix

roles/lib_openshift/src/doc/router
kwoodson: will fix

roles/lib_openshift/src/ansible/oc_adm_router.py
detiber: @ashcrow it does, in `src/class/oc_adm_router.py:_prepare_router()` it looks like cacert_file, cert_file, and key_file are merged to create a combined file that is used for `--default-cert`

roles/lib_openshift/library/oc_adm_registry.py
kwoodson: will fix

roles/lib_openshift/library/oc_adm_router.py
kwoodson: will fix

roles/lib_openshift/src/ansible/oc_adm_router.py
kwoodson: @ashcrow, @detiber, Should I updated to support both methods?  The biggest issue with the way we did this is that the cert and key were already on the box and _not_ in the pem file format.  Building our own was nice.  Is it worth it to provide a way to pass them pem as well?  If so, I'll implement that.

ashcrow: I'll defer to @detiber on that.

detiber: @kwoodson I think that sounds like a good idea. You'll probably want to make sure to make the options mutually_exclusive to avoid confusion.

ashcrow:
https://github.com/openshift/openshift-ansible/pull/3424 merged has been merged. Rebase and regenerate to pull in the ``has_key`` fixes.

roles/lib_openshift/library/oc_serviceaccount.py
ashcrow: I read ``volume`` and think of storage.

roles/lib_openshift/src/ansible/oc_adm_router.py
ashcrow: :+1:

ashcrow:
aos-ci-test

ashcrow:
BTW, @kwoodson thanks for all your work on these modules!

ashcrow:
@kwoodson why change the python files to ``0755``?

kwoodson:
@ashcrow, we run the script in lib_openshfit/src/test/generate-and-run-tests.sh.  This expects them to be executable.  Its a quick way to run all the tests in `unit/` and `integration/`.

mtnbikenc:
Examples still reference oadm_

mtnbikenc:
aos-ci-test

PR #3378, opened 2017-02-15
Adding oc_adm_ca_server_cert to lib_openshift

kwoodson:


kwoodson:
@ashcrow @abutcher, no test cases added but did test this call:
`  oadm_ca:
    cmd: create-server-cert
    ...
`

roles/lib_openshift/src/ansible/oadm_certificate_authority.py
```diff
@@ -0,0 +1,49 @@
...
+# flake8: noqa
+
+def main():
+    '''
+    ansible oadm module for ca
+    '''
+
+    module = AnsibleModule(
+        argument_spec=dict(
+            state=dict(default='present', type='str',
```
ashcrow: I assume this is here for yaml consistency as it is the only option?

roles/lib_openshift/src/doc/certificate_authority
```diff
@@ -0,0 +1,96 @@
...
+    - The directory to place the certificates.
+    required: false
+    default: False
+    aliases: []
+author:
+- "Kenny Woodson <kwoodson@redhat.com>"
+extends_documentation_fragment: []
+'''
+
+EXAMPLES = '''
```
ashcrow: Wrong example :smile:

roles/lib_openshift/src/doc/certificate_authority
```diff
@@ -0,0 +1,96 @@
+# flake8: noqa
+# pylint: skip-file
+
+DOCUMENTATION = '''
+---
+module: oc_secret
```
ashcrow: Wrong docfile?

roles/lib_openshift/src/ansible/oadm_certificate_authority.py
kwoodson: Yeah, I'm fine with removing it as well.  Consistency.  I guess an update is possible so 'present' represents create and update.

roles/lib_openshift/src/class/oadm_certificate_authority.py
```diff
@@ -0,0 +1,110 @@
+# pylint: skip-file
+
+class CertificateAuthorityConfig(OpenShiftCLIConfig):
+    ''' CertificateAuthorityConfig is a DTO for the oadm ca command '''
```
detiber: @kwoodson is the oadm_ naming just a legacy artifact? I would think we would want to use `oc adm` and would suggest the module name probably be `oc_adm_` rather than `oadm_`

detiber: I know there have recently been changes to the cli tooling with deprecating and replacing commands related to CA management. I suspect @abutcher can probably cover those.

roles/lib_openshift/src/class/oadm_certificate_authority.py
```diff
@@ -0,0 +1,110 @@
...
+           If a file exists by the same name in the specified location then the cert exists
+        '''
+        cert = self.config.config_options['cert']['value']
+        if cert and os.path.exists(cert):
+            return open(cert).read()
+
+        return None
+
+    def create(self):
+        '''Create a deploymentconfig '''
```
detiber: docstring does not match method/class

roles/lib_openshift/src/class/oadm_certificate_authority.py
```diff
@@ -0,0 +1,110 @@
...
+        options = self.config.to_option_list()
+
+        cmd = ['ca']
+        cmd.append(self.config.cmd)
+        cmd.extend(options)
+
+        return self.openshift_cmd(cmd, oadm=True)
+
+    def exists(self):
+        ''' check whether the certificate exists and has the clusterIP '''
```
detiber: I'm a bit curious to how this will work in practice, since the actual certs will be distributed amongst the cluster and embedded in deployments, kubeconfigs, etc.

roles/lib_openshift/src/doc/certificate_authority
```diff
@@ -0,0 +1,141 @@
...
+# pylint: skip-file
+
+DOCUMENTATION = '''
+---
+module: oadm_ca
+short_description: Module to manage openshift certificate authority
+description:
+  - Wrapper around the openshift `oc adm ca` command.
+options:
+  state:
```
detiber: Any reason not to support `state: list`, since there is an exists method?

roles/lib_openshift/src/doc/certificate_authority
```diff
@@ -0,0 +1,141 @@
...
+    aliases: []
+author:
+- "Kenny Woodson <kwoodson@redhat.com>"
+extends_documentation_fragment: []
+'''
+
+EXAMPLES = '''
+- name: Create a self-signed cert
+  oadm_ca:
+    cmd: create-server-cert
```
detiber: It seems like this would want to be multiple modules rather than a single module that supports multiple commands.

kwoodson: @detiber, totally agree with that.  Since this is a port over I didn't tack on that work.  I could see a command for each subsection.  There are currently 4.

  - create-key-pair     Create a public/private key pair
  - create-master-certs Create certificates and keys for a master
  - create-server-cert  Create a signed server certificate and key
  - create-signer-cert  Create a signer (certificate authority/CA) certificate and key

roles/lib_openshift/src/doc/certificate_authority
kwoodson: @detiber, it really depends on what it means to exist.  The instance that we use primarily is to generate a signed certificate. In that instance, the exists verifies that the hostnames are in the certificate.  It probably should be broken out into their own modules.

roles/lib_openshift/src/class/oadm_certificate_authority.py
kwoodson: It currently just detects the signed server cert and if it includes the hostnames.

roles/lib_openshift/src/class/oadm_certificate_authority.py
kwoodson: will fix

roles/lib_openshift/src/class/oadm_certificate_authority.py
kwoodson: @detiber, I can do the rename.  Yeah, probably considered legacy.

roles/lib_openshift/src/class/oadm_certificate_authority.py
detiber: Gotcha, so it is more like a validation step after the cert is generated.

roles/lib_openshift/src/doc/certificate_authority
detiber: ack, makes sense. I misread the intention of the exists method in skimming the PR.

roles/lib_openshift/src/doc/certificate_authority
detiber: Out of curiosity, is there a particular use case that is driving the porting of this module now? It doesn't seem like replacing the current calls to generate certificates (that interact with the oc tooling) would buy us much if any improvements to idempotency.

I also know that discussions with @abutcher seemed to indicate that there is going to be a new cert workflow coming that would allow us to do all certificates through the OpenShift CA as opposed to only the subset we are able to do now.

If this is a dependency of the router and registry modules, would it be be better to modify them to accept a pre-created certificates instead as a stopgap?

kwoodson: @detiber @abutcher,

The certificate code that would be replaced by using this module is here https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/tasks/registry/secure.yml#L34-L42.  We do the exact call in our openshift_secure_registry here: https://github.com/openshift/openshift-tools/blob/prod/ansible/roles/openshift_aws_registry/tasks/main.yml#L43-L52.

As to the pre-created certificates, I'm fine if we want to support them in the future.  I think we should have feature parity and then once everyone is happy where things are we can expand everything.

detiber: @kwoodson As long as we are ok limiting the scope of porting to using the module to that particular use case, then I see no issues with pulling it in as is. I'm wondering if we need to somehow document or otherwise track that we wouldn't be able to use it for general certificate management.

ashcrow:

roles/lib_openshift/src/doc/ca_server_cert
```diff
@@ -0,0 +1,90 @@
...
+  - Wrapper around the openshift `oc adm ca create-server-cert` command.
+options:
+  state:
+    description:
+    - Present is the only supported state.  The state present means that `oc adm ca` will generate a certificate
+    - and verify if the hostnames and the ClusterIP exists in the certificate.
+    - When create-server-cert is desired then the following parameters are passed.
+    - ['cert', 'key', 'signer_cert', 'signer_key', 'signer_serial']
+    required: false
+    default: present
```
detiber: Would there be value in deleting the certificate from disk in some cases?

It might also be good to have this module provide a way to 'backup' the certificate file. @abutcher I'm thinking specifically about the cert replacement workflow, thoughts?

roles/lib_openshift/src/ansible/oc_adm_ca_server_cert.py
```diff
@@ -0,0 +1,36 @@
...
+            state=dict(default='present', type='str', choices=['present']),
+            debug=dict(default=False, type='bool'),
+            kubeconfig=dict(default='/etc/origin/master/admin.kubeconfig', type='str'),
+            cmd=dict(default=None, require=True, type='str'),
+            # oadm ca create-server-cert [options]
+            cert=dict(default=None, type='str'),
+            key=dict(default=None, type='str'),
+            signer_cert=dict(default=None, type='str'),
+            signer_key=dict(default=None, type='str'),
+            signer_serial=dict(default=None, type='str'),
```
kwoodson: @detiber,

I can default the signer_* to go to /etc/origin/master/ca.*.

roles/lib_openshift/src/ansible/oc_adm_ca_server_cert.py
```diff
@@ -0,0 +1,36 @@
...
+            hostnames=dict(default=[], type='list'),
+            overwrite=dict(default=False, type='bool'),
+        ),
+        supports_check_mode=True,
+    )
+
+    # pylint: disable=line-too-long
```
kwoodson: will clean up

roles/lib_openshift/src/ansible/oc_adm_ca_server_cert.py
```diff
@@ -0,0 +1,36 @@
...
+    results = CAServerCert.run_ansible(module.params, module.check_mode)
+    if 'failed' in results:
+        return module.fail_json(**results)
+
+    return module.exit_json(**results)
```
kwoodson: @detiber,

I went through this debate with the operations team a couple of times.  We chose this method as we can now use the library/oc*py as python libraries.  The border of what is ansible and what is a python is the module object.  By just passing params into the run_ansible we get a clear separation and the ability to run idempotent code inside of python scripts.

I would love to reduce boilerplate and I'm open to suggestions.

kwoodson: @ashcrow, if you have comments or suggestions here I would be interested.

roles/lib_openshift/src/class/oc_adm_ca_server_cert.py
```diff
@@ -0,0 +1,103 @@
...
+        return self.openshift_cmd(cmd, oadm=True)
+
+    def exists(self):
+        ''' check whether the certificate exists and has the clusterIP '''
+
+        cert_path = self.config.config_options['cert']['value']
+        if not os.path.exists(cert_path):
+            return False
+
+        # Would prefer pyopenssl but is not installed.
```
kwoodson: @detiber, I made the note because I saw the emails with Tim's module.  We also hit this when trying operations with pyopenssl in the operations environment. 

We need a solution for this.  If ssl.SSLContext would work then we could look into that.  What package is that in?  Is that in the stdlib?  If so, we should investigate.

roles/lib_openshift/src/doc/ca_server_cert
kwoodson: @detiber, I'd image a delete would be a `file: state=absent`.  I'm just trying to duplicate the `oc adm ca create-server-cert` command.

Good suggestion on backups.  We could definitely add that support.

roles/lib_openshift/src/ansible/oc_adm_ca_server_cert.py
```diff
@@ -0,0 +1,36 @@
...
+            kubeconfig=dict(default='/etc/origin/master/admin.kubeconfig', type='str'),
+            cmd=dict(default=None, require=True, type='str'),
+            # oadm ca create-server-cert [options]
+            cert=dict(default=None, type='str'),
+            key=dict(default=None, type='str'),
+            signer_cert=dict(default=None, type='str'),
+            signer_key=dict(default=None, type='str'),
+            signer_serial=dict(default=None, type='str'),
+            hostnames=dict(default=[], type='list'),
+            overwrite=dict(default=False, type='bool'),
```
kwoodson: @detiber,
 
I'm intrigued by this option.  The `oc adm ca create-server-cert` doesn't actually return anything.  It does create the cert and key.  We currently don't return the cert and key although we could if that would be useful.

I imagine how that interface would look is you would pass (cert and key) or (pem).  Should we return the artifacts that are created by the oc command?

roles/lib_openshift/src/ansible/oc_adm_ca_server_cert.py
detiber: @kwoodson would there ever be a case for calling into the run_ansible method of a different 'type'? From what I can tell, the only place it is called is from the 'module' implementation.

If that is the case, then I don't think pushing the fail_json/exit_json bits into the class would break that use case. One thing that would probably need to be done, is to change the method signature to just accept a 'module' instance rather than module.params/module.check_mode.

roles/lib_openshift/src/class/oc_adm_ca_server_cert.py
detiber: It is stdlib, but I'm not sure if we can count on python >= 2.7.9. @sdodson could you verify that?

roles/lib_openshift/src/doc/ca_server_cert
detiber: Considering that these tasks create files on disk, maybe it would make sense to use `add_file_common_args=True` to the instantiation of `AnsibleModule`, then you would get all the file-based args, and even the backup for free.

roles/lib_openshift/src/ansible/oc_adm_ca_server_cert.py
ashcrow: To be honest either way works. Personally, I like it in ``main`` as it looks like the ansible module examples and one can quickly look at ``main`` to see what would be considered a failure.

roles/lib_openshift/src/ansible/oc_adm_ca_server_cert.py
detiber: I think it would be useful. Not just for this module, but for others as well. 

A common use case that I'm seeing is where there are two tasks back-to-back to do `state: present` and then `state: list` to retrieve the object.

roles/lib_openshift/src/ansible/oc_adm_ca_server_cert.py
kwoodson: @detiber,

We call into run_ansible from our tests, python scripts that are in our operations monitoring container, and also this allows us to debug in using pdb.

roles/lib_openshift/src/doc/ca_server_cert
kwoodson: @detiber, added backup support.

roles/lib_openshift/library/oc_adm_ca_server_cert.py
```diff
@@ -1345,6 +1351,17 @@ def get(self):
...
+
+        # Added this here as a safegaurd for stomping on the
+        # cert and key files if they exist
+        if self.config.config_options['backup']['value']:
+            if os.path.exists(self.config.config_options['key']['value']):
+                shutil.copy(self.config.config_options['key']['value'],
+                            "%s.orig" % self.config.config_options['key']['value'])
+            if os.path.exists(self.config.config_options['cert']['value']):
+                shutil.copy(self.config.config_options['cert']['value'],
+                            "%s.orig" % self.config.config_options['cert']['value'])
```
detiber: Would date stamped backups be better here? Ideally, I would like to see us use [backup_local](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L2055) from module_utils, but that would require passing module down into CAServerCert/OpenShiftCLI.

Having the date stamped backups will ensure that we wouldn't clobber the backups on a subsequent run.

kwoodson: @detiber, I thought about date but decided not to.  I can do that.  PR inc

ashcrow:
aos-ci-test

kwoodson:
@detiber @ashcrow, Backup feature added with proper naming.  Fixed a bug when checking subject and SAN from x509 output.  Tested with multiple iterations.

I believe this one is ready.

roles/lib_openshift/src/ansible/oc_adm_ca_server_cert.py
```diff
@@ -0,0 +1,35 @@
...
+    ansible oc adm module for ca create-server-cert
+    '''
+
+    module = AnsibleModule(
+        argument_spec=dict(
+            state=dict(default='present', type='str', choices=['present']),
+            debug=dict(default=False, type='bool'),
+            kubeconfig=dict(default='/etc/origin/master/admin.kubeconfig', type='str'),
+            backup=dict(default=True, type='bool'),
+            # oadm ca create-server-cert [options]
```
detiber: s/oadm/oc adm/ - not a blocker by any means, though

roles/lib_openshift/src/class/oc_adm_ca_server_cert.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @staticmethod
+    def run_ansible(params, check_mode):
+        '''run the idempotent ansible code'''
+
+        config = CAServerCertConfig(params['kubeconfig'],
+                                    params['debug'],
+                                    {'cert':          {'value': params['cert'], 'include': True},
+                                     'hostnames':     {'value': ','.join(params['hostnames']), 'include': True},
+                                     'overwrite':     {'value': True, 'include': True},
```
detiber: Just to be safe, should the value of `overwrite` be set to match the value of `force`?  It might make sense to make `overwrite` an alias for `force` as well.

ashcrow:
@kwoodson Looks like regeneration is needed.

```
Checking generated module code: roles/lib_openshift/src/generate.py
('Generated content does not match for /home/travis/build/openshift/openshift-ansible/roles/lib_openshift/src/../library/oc_adm_router.py',)
Checking generated module code: roles/lib_utils/src/generate.py
Found errors while generating module code.
ERROR: InvocationError: '/home/travis/build/openshift/openshift-ansible/.tox/py27-ansible22-generate_validation/bin/python setup.py generate_validation'
```

roles/lib_openshift/src/class/oc_adm_ca_server_cert.py
kwoodson: @detiber, ok, so here's the thought process behind overwrite and force.

Overwrite is a parameter to the `oc adm ca`  command.  It will only update certs if the file exists and overwrite is true.  If you have changes that you know need to take effect, when overwrite is false it will never apply them.  I believe we want overwrite to always be true.

Force is a parameter to allow the user to force an update of the cert file.  Most of the time we won't need to force and it will default to false.  If we detect that the certificate is lacking hostnames, we will want to set overwrite to true, force would remain false, and we would get the update.  If we set overwrite to the value of force, we would only get updates on force since overwrite only writes updates when true.

Force allows the user to get certificate updates no matter what.  I wanted to provide this behavior but only when needed.

roles/lib_openshift/src/ansible/oc_adm_ca_server_cert.py
kwoodson: will fix

roles/lib_openshift/src/class/oc_adm_ca_server_cert.py
detiber: That doesn't seem to be the behavior below, it is only calling create() when it doesn't exist or force is set, which would mean that force/overwrite are linked.

I'm also not sure if there is a reason we would want to set overwrite to true for get operations.

detiber: Ah, I see my confusion now. I had forgotten that exists compares both the cert is on disk and the contents.

Not a blocker, but I think it would be a good idea to break down the concept of create if not present and update if needed.

kwoodson:
@detiber, localized the import time to the class that uses it.

roles/lib_openshift/src/doc/ca_server_cert
```diff
@@ -0,0 +1,96 @@
...
+    aliases: []
+  signer_serial:
+    description:
+    - The signer serial file.
+    required: false
+    default: /etc/origin/master/ca.serial.txt
+    aliases: []
+  hostnames:
+    description:
+    - Every hostname or IP that server certs should be valid for (comma-delimited list)
```
cben: If I read the code correctly, the ansible module takes a real list, it's only comma-delimited string internally in CAServerCertConfig

kwoodson: @cben, thanks for finding this.  I will update.

PR #3380, opened 2017-02-16
initialize oo_nodes_to_upgrade group when running control plane upgrade only

ingvagabund:
The missing piece to initialize ``oo_nodes_up_upgrade`` group in order to upgrade nodes that are masters at the same time.

Before, control plane upgrade was responsible for upgrading master nodes. Recently, node upgrade play was integrated into the control plane upgrade. Unfortunately, as the control plane upgrade did not manage node upgrades, only some groups of hosts were initialized. The ``oo_nodes_up_upgrade`` group was not among them. Which cause the integrated node upgrade to be skipped.


ingvagabund:
aos-ci-test

sdodson:
This fixes a bug, right? If you put 'Bug ####' into the PR description the bot will post a comment in the bug when it merges making it easier to link PR to BZ.

PR #3381, opened 2017-02-16
bug 1420217. Default ES memory to be compariable to 3.4 deployer

jcantrill:
Defaulting memory to be comparible to 3.4


jcantrill:
aos-ci-test

PR #3382, opened 2017-02-16
Dockerfile and docs to run containerized playbooks

codificat:
A Dockerfile based on [playbook2image](https://github.com/aweiteka/playbook2image) to containerize openshift-ansible so we can run an arbitrary playbook from a container.

WIP, want to gather feedback and thoughts around it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,33 @@
-FROM rhel7
+# Using playbook2image as a base
+# See https://github.com/aweiteka/playbook2image for details on the image
+# including documentation for the settings/env vars referenced below
+FROM docker.io/aweiteka/playbook2image:latest
 
-MAINTAINER Troy Dawson <tdawson@redhat.com>
+MAINTAINER OpenShift Team <dev@lists.openshift.redhat.com>
 
-LABEL Name="openshift3/installer"
-LABEL Vendor="Red Hat" License=GPLv2+
-LABEL Version="v3.1.1.901"
-LABEL Release="6"
-LABEL BZComponent="aos3-installation-docker"
-LABEL Architecture="x86_64"
-LABEL io.k8s.description="Ansible code and playbooks for installing Openshift Container Platform." \
-      io.k8s.display-name="Openshift Installer" \
-      io.openshift.tags="openshift,installer"
+LABEL name="openshift-ansible" \
+      summary="OpenShift's installation and configuration tool" \
+      description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain and check an OpenShift cluster" \
+      url="https://github.com/openshift/openshift-ansible" \
+      io.k8s.display-name="openshift-ansible" \
+      io.k8s.description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain and check an OpenShift cluster" \
+      io.openshift.expose-services="" \
+      io.openshift.tags="openshift,install,upgrade,ansible"
 
-RUN INSTALL_PKGS="atomic-openshift-utils" && \
-    yum install -y --enablerepo=rhel-7-server-ose-3.2-rpms $INSTALL_PKGS && \
-    rpm -V $INSTALL_PKGS && \
-    yum clean all
+# The playbook to be run is specified via the PLAYBOOK_FILE env var.
+# This sets a default of openshift_facts.yml as it's an informative playbook
+# that can help test that everything is set properly (inventory, sshkeys)
+ENV PLAYBOOK_FILE=playbooks/byo/openshift_facts.yml \
+    OPTS="-v" \
+    INSTALL_OC=true
 
-# Expect user to mount a workdir for container output (installer.cfg, hosts inventory, ansible log)
-VOLUME /var/lib/openshift-installer/
-WORKDIR /var/lib/openshift-installer/
+# playbook2image's assemble script expects the source to be available in
+# /tmp/src (as per the source-to-image specs) so we import it there
+ADD . /tmp/src
 
-RUN mkdir -p /var/lib/openshift-installer/
+# Running the 'assemble' script provided by playbook2image will install
+# dependencies specified in requirements.txt and install the 'oc' client
+# as per the INSTALL_OC environment setting above
+RUN /usr/libexec/s2i/assemble
 
-ENTRYPOINT ["/usr/bin/atomic-openshift-installer", "-c", "/var/lib/openshift-installer/installer.cfg", "--ansible-log-path", "/var/lib/openshift-installer/ansible.log"]
+CMD [ "/usr/libexec/s2i/run" ]
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -1,0 +1,26 @@
+FROM rhel7
+
+MAINTAINER Troy Dawson <tdawson@redhat.com>
+
+LABEL Name="openshift3/installer"
+LABEL Vendor="Red Hat" License=GPLv2+
+LABEL Version="v3.1.1.901"
+LABEL Release="6"
+LABEL BZComponent="aos3-installation-docker"
+LABEL Architecture="x86_64"
+LABEL io.k8s.description="Ansible code and playbooks for installing Openshift Container Platform." \
+      io.k8s.display-name="Openshift Installer" \
+      io.openshift.tags="openshift,installer"
+
+RUN INSTALL_PKGS="atomic-openshift-utils" && \
+    yum install -y --enablerepo=rhel-7-server-ose-3.2-rpms $INSTALL_PKGS && \
+    rpm -V $INSTALL_PKGS && \
+    yum clean all
+
+# Expect user to mount a workdir for container output (installer.cfg, hosts inventory, ansible log)
+VOLUME /var/lib/openshift-installer/
+WORKDIR /var/lib/openshift-installer/
+
+RUN mkdir -p /var/lib/openshift-installer/
+
+ENTRYPOINT ["/usr/bin/atomic-openshift-installer", "-c", "/var/lib/openshift-installer/installer.cfg", "--ansible-log-path", "/var/lib/openshift-installer/ansible.log"]
--- a/README_CONTAINER_IMAGE.md
+++ b/README_CONTAINER_IMAGE.md
@@ -1,0 +1,41 @@
+# Containerized openshift-ansible to run playbooks
+
+The [Dockerfile](Dockerfile) in this repository uses the [playbook2image](https://github.com/aweiteka/playbook2image) source-to-image base image to containerize `openshift-ansible`. The resulting image can run any of the provided playbooks.
+
+**Note**: at this time there are known issues that prevent to run this image for installation/upgrade purposes from within one of the hosts that is also an installation target at the same time: if the playbook you want to run attempts to manage the docker daemon and restart it (like install/upgrade playbooks do) this would kill the container itself during its operation.
+
+## Build
+
+To build a container image of `openshift-ansible`:
+
+1. Using standalone **Docker**:
+
+        cd openshift-ansible
+        docker build -t openshift-ansible .
+
+1. Using an **OpenShift** build:
+
+        oc new-build docker.io/aweiteka/playbook2image~https://github.com/openshift/openshift-ansible
+        oc describe imagestream openshift-ansible
+
+## Usage
+
+The base image provides several options to control the behaviour of the containers. For more details on these options see the [playbook2image](https://github.com/aweiteka/playbook2image) documentation.
+
+At the very least, when running a container using an image built this way you must specify:
+
+1. The **playbook** to run. This is set using the `PLAYBOOK_FILE` environment variable.
+1. An **inventory** file. This can be mounted inside the container as a volume and specified with the `INVENTORY_FILE` environment variable. Alternatively you can serve the inventory file from a web server and use the `INVENTORY_URL` environment variable to fetch it.
+1. **ssh keys** so that Ansible can reach your hosts. These should be mounted as a volume under `/opt/app-root/src/.ssh`
+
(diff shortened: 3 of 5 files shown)


DanyC97:
@codificat for my own knowledge (sorry if is an obvious answer) if i understand correctly you want to run the `openshift-ansible` playbooks from containers running inside Openshift? if so i am a bit confused what is the benefit.

images/run-playbook/README.md
```diff
@@ -0,0 +1,43 @@
+# Containerized openshift-ansible to run generic playbooks
+
+This image uses the [playbook2image](https://github.com/aweiteka/playbook2image) source-to-image image to containerize `openshift-ansible`. The resulting image can run any of the provided playbooks.
+
+## Build
+
+1. **Docker**:
```
detiber: Is this capable of being run as system container? @giuseppe

giuseppe: the configuration for the container seems easy enough that should not be a problem to make a system container of it we will need to provide the configuration for runC.

We'll probably benefit of running the systemd service as Type=oneshot

codificat:
Some context: the trigger use case behind this was to be able to regularly run a playbook to check expiration dates of internal certificates using the openshift_certificate_expiry role. The idea was indeed to have this run from within OpenShift itself e.g. via a ScheduledJob. There are more use cases that would benefit from the "run a specific playbook from a container" approach.

There's another Dockerfile in the top level dir which has a different focus (i.e. to config/upgrade the cluster). One point to review is if it's worth unifying both types of use cases under a single image.

rhcarvalho:
@DanyC97 another use case and trigger for an earlier work on this that didn't make it into master was to run maintenance and health check playbooks in environments that do not have Ansible 2.2. In some corporate envs there are barries to upgrade/install RPMs, so the Docker image would allow us to ship a "snapshot" of openshift-ansible in a know-to-work configuration -- and that could be run as a Docker container, independent of OpenShift (some playbooks are useful before OpenShift is even installed).

images/run-playbook/README.md
aweiteka: This sounds really interesting. @detiber could you describe the use case a bit?

detiber: @aweiteka The main use case is being able to run the installer on one of the hosts that is part of the installation.

The issue we have today is that as part of the installation we are making changes to the docker config which require a service restart. If the installer is running in a container under docker, then it goes *poof* when the docker restart happens.

detiber:
> There's another Dockerfile in the top level dir which has a different focus (i.e. to config/upgrade the cluster). One point to review is if it's worth unifying both types of use cases under a single image.

I would say the answer to this is: absolutely.

images/run-playbook/README.md
aweiteka: @codificat I'll look at putting the system container files into the playbook2image base so they're inherited for this optional use case. https://github.com/aweiteka/playbook2image/issues/22

DanyC97:
thank you @codificat @rhcarvalho for providing insights, much appreciated !

detiber:
@codificat looks like this PR is raising some yamllint errors: https://travis-ci.org/openshift/openshift-ansible/jobs/202362727#L953

detiber:
travis failures should clean up with a rebase, just fixed in master a few min ago.

Dockerfile
```diff
@@ -1,26 +1,21 @@
...
 
-LABEL Name="openshift3/installer"
-LABEL Vendor="Red Hat" License=GPLv2+
-LABEL Version="v3.1.1.901"
-LABEL Release="6"
-LABEL BZComponent="aos3-installation-docker"
-LABEL Architecture="x86_64"
-LABEL io.k8s.description="Ansible code and playbooks for installing Openshift Container Platform." \
-      io.k8s.display-name="Openshift Installer" \
-      io.openshift.tags="openshift,installer"
```
detiber: Might be good to have relevant metadata for this image, maybe @tdawson can help with recommendations.

codificat:
@detiber thanks for the feedback. yamllint should be happier now and I'll be rebasing, squashing etc as needed as this progresses.

About unifying into a single Dockerfile:
1. I'd move the existing `Dockerfile` to `Dockerfile.rhel7` as I believe it's only a prototype so far and no one is really using it yet, and it seems to be designed for productization (please correct me if that's not the case)
2. Moving the proposed `Dockerfile` that uses `playbook2image` to replace the current Dockerfile. Also moving the doc to the top level dir and not adding an `images` dir.

I'm still thinking how to best handle input/output for the general playbook case. The current Dockerfile uses/requires a VOLUME, not sure if that's the one-and-only way...

jcantrill:
@DanyC97 'oc cluster up' also has need to deploy logging and metrics in order that we may fully deprecate their deployers

Dockerfile
```diff
@@ -1,26 +1,21 @@
...
+# See https://github.com/aweiteka/playbook2image for details on the image
+# including some of the settings below
+FROM docker.io/aweiteka/playbook2image:latest
 
-MAINTAINER Troy Dawson <tdawson@redhat.com>
+# The playbook to be run is specified via the PLAYBOOK_FILE env var
+# Setting a default of openshift_facts.yml as it's an informative playbook
+# that can help test that everything is set properly (inventory, sshkeys)
+ENV PLAYBOOK_FILE=plabyooks/byo/openshift_facts.yml \
+    ANSIBLE_HOST_KEY_CHECKING=False \
```
sosiouxme: I get why this is needed for non-interactive usage, but could there be a comment as to why?

Something to the effect that unless you mount in a pre-verified known_hosts file, there's no security mechanism to verify the host keys at the time this is running. While we could use the method in http://stackoverflow.com/a/35213839/262768 to set up such a file at run-time, it wouldn't offer any actual security.

README_CONTAINER_IMAGE.md
```diff
@@ -0,0 +1,37 @@
...
+The [Dockerfile](Dockerfile) in this repository uses the [playbook2image](https://github.com/aweiteka/playbook2image) source-to-image base image to containerize `openshift-ansible`. The resulting image can run any of the provided playbooks.
+
+## Build
+
+To build a container image of `openshift-ansible`:
+
+1. Using standalone **Docker**:
+
+        cd openshift-ansible
+        docker build -t certcheck -f images/certificate-check/Dockerfile .
```
sosiouxme: errr... the images/certificate-check/Dockerfile doesn't seem to exist...?

README_CONTAINER_IMAGE.md
```diff
@@ -0,0 +1,37 @@
...
+
+Here is an example of how to run a containerized `openshift-ansible` playbook that will check the expiration dates of OpenShift's internal certificates using the [`openshift_certificate_expiry` role](../../roles/openshift_certificate_expiry), using an inventory served by a local HTTP server:
+
+    docker run --rm -u `id -u` \
+           -v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z \
+           -e INVENTORY_URL=http://172.17.0.1/hosts \
+           -e PLAYBOOK_FILE=playbooks/certificate_expiry/default.yaml \
+           certcheck
+
+There [playbook2image examples](https://github.com/aweiteka/playbook2image/tree/master/examples) provide additional information on how to use a built image.
```
sosiouxme: s/There/The/

Dockerfile
detiber: Can we provide a user to run the system container as and automatically mount ~/.ssh for that user?

Then we could provide a flag to opt in for host key checking.

This would also help for ssh proxy config, environment-specific auth, etc.

codificat: Right... for the use cases of running the built image from inside OpenShift itself I was thinking to use a ConfigMap for these settings. For a standalone container I'll update the docs to explain that better.

One thing to keep in mind with mounting `~/.ssh` is that perms/selinux come into play. That's why the example `docker run` in `README_CONTAINER_IMAGE` sets the uid with `id -u` and mounts the ssh key with `:Z` - this has a side effect or leaving your key labelled something like `system_u:object_r:container_file_t:s0:c609,c894`

README_CONTAINER_IMAGE.md
codificat: Rigth, I moved it to the top level dir and forgot to update the doc, will fix, thanks for spotting!

codificat:
Addressed the comments above: added some labels, fixed typos/path references and removed `ANSIBLE_HOST_KEY_CHECKING`.

It turns out that the `ansible.cfg` in the repo contains  `host_key_checking=False`. It's worth documenting this more and explain how to add `known_hosts` to the container. Maybe we should remove that setting from `ansible.cfg`? Not sure why it's there as a default

Will update `Dockerfile.rhel7` after some tests with OSBS.

codificat:
Rebased and squashed.

@detiber PTAL

There are things to be improved (more docs, including ssh key handling; rhel7 Dockerfile update; input/output handling) but these can be done in another PR... having this one in should give us a base to build upon, including having images built and tested.

Dockerfile.rhel7
```diff
@@ -2,16 +2,17 @@ FROM rhel7
...
+LABEL vendor="Red Hat" License=GPLv2+
+LABEL version="v3.4.1.8"
+LABEL release="1"
+LABEL com.redhat.component="aos3-installation-docker"
+LABEL architecture="x86_64"
 LABEL io.k8s.description="Ansible code and playbooks for installing Openshift Container Platform." \
       io.k8s.display-name="Openshift Installer" \
       io.openshift.tags="openshift,installer"
 
+
```
rhcarvalho: nit: extra line?

Dockerfile.rhel7
```diff
@@ -2,16 +2,17 @@ FROM rhel7
...
-LABEL Version="v3.1.1.901"
-LABEL Release="6"
-LABEL BZComponent="aos3-installation-docker"
-LABEL Architecture="x86_64"
+LABEL name="openshift3/installer"
+LABEL vendor="Red Hat" License=GPLv2+
+LABEL version="v3.4.1.8"
+LABEL release="1"
+LABEL com.redhat.component="aos3-installation-docker"
+LABEL architecture="x86_64"
```
rhcarvalho: Good to update these, though we need to make sure this format will already be used by the build tools?
I remember once when we updated labels and then had to revert because the build tool was not accepting the new standard at the time.

Dockerfile
```diff
@@ -1,26 +1,31 @@
-FROM rhel7
+# Using playbook2image as a base
+# See https://github.com/aweiteka/playbook2image for details on the image
+# including documentation for the settings/env vars referenced below
+FROM docker.io/aweiteka/playbook2image:latest
 
-MAINTAINER Troy Dawson <tdawson@redhat.com>
```
rhcarvalho: Shall we declare a MAINTAINER (perhaps a group/team) for this new image?

Dockerfile
```diff
@@ -1,26 +1,31 @@
...
+LABEL name="openshift-ansible" \
+      summary="OpenShift's installation and configuration tool" \
+      description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain, chek, etc. an OpenShift cluster" \
```
rhcarvalho: nit: typo s/chek/check

I'd probably remove the "etc", for it doesn't add information for users/customers.

Dockerfile
```diff
@@ -1,26 +1,31 @@
...
+      url="https://github.com/openshift/openshift-ansible" \
+      io.k8s.display-name="openshift-ansible" \
+      io.k8s.description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain, chek, etc. an OpenShift cluster" \
```
rhcarvalho: Same as above

README_CONTAINER_IMAGE.md
```diff
@@ -0,0 +1,38 @@
...
+        oc describe imagestream openshift-ansible
+
+## Usage
+
+The base image provides several options to control the behaviour of the containers. For more details on these options see the [playbook2image](https://github.com/aweiteka/playbook2image) documentation.
+
+At the very least, when running a container using an image built this way you must specify:
+
+1. The **playbook** to run. This is set using the `PLAYBOOK_FILE` environment variable.
+1. An **inventory** file. One way to provide this is to serve the inventory file from a web server and use the `INVENTORY_URL` environment variable to fetch it.
```
rhcarvalho: I'd suggest to document the use of INVENTORY_FILE instead.

Users will already need to learn/deal with mounts for SSH keys, so I think using that for the inventory would be more natural than have people figure out how to serve their file over HTTP.

Dockerfile.rhel7
codificat: Right - some changes will be necessary to this Dockerfile but there are other steps to do first, most notably to have a base playbook2image in the build system.
I'll revert that commit and leave Dockerfile.rhel7 untouched (just a rename of the previous Dockerfile)

Dockerfile.rhel7
codificat: ditto, will defer changes to the product Dockerfile

Dockerfile
codificat: Will add one and fix the typos :)

README_CONTAINER_IMAGE.md
codificat: Right, will do that - I had problems when I tried this before and that's why I resorted to url-based, leaving that for later, but it works now, so I'll update the docs.

Dockerfile
sosiouxme: nope, you're on the hook forever!

I suggest dev@lists.openshift.redhat.com for all our maintainers.

rhcarvalho: @sosiouxme that's the course of action we took, see the line below :)

rhcarvalho: The original file was moved to `Dockerfile.rhel7`, but because of the name this file is getting compared to the old one.

codificat: Right, I applied the changes and force-pushed, but somehow github was still showing these outdated comments... the latest commit (`c6b43ef`) has dev@lists.openshift.redhat.com as maintainer.

rhcarvalho:
aos-ci-test

PR #3387, opened 2017-02-16
Add SDNValidator Module

enj:
This module should be run as an early stage per-upgrade check for `1.4` to `1.5` to prevent the upgrade from happening if certain SDN objects are invalid.

See openshift/origin/issues/12697, openshift/origin/issues/12959

cc @sdodson @kwoodson 

Signed-off-by: Monis Khan <mkhan@redhat.com>

### DIFF
--- a/roles/lib_openshift/src/ansible/oc_sdnvalidator.py
+++ b/roles/lib_openshift/src/ansible/oc_sdnvalidator.py
@@ -1,0 +1,24 @@
+# pylint: skip-file
+# flake8: noqa
+
+def main():
+    '''
+    ansible oc module for validating OpenShift SDN objects
+    '''
+
+    module = AnsibleModule(
+        argument_spec=dict(
+            kubeconfig=dict(default='/etc/origin/master/admin.kubeconfig', type='str'),
+        ),
+        supports_check_mode=False,
+    )
+
+
+    rval = OCSDNValidator.run_ansible(module.params)
+    if 'failed' in rval:
+        module.fail_json(**rval)
+
+    module.exit_json(**rval)
+
+if __name__ == '__main__':
+    main()
--- a/roles/lib_openshift/src/class/oc_sdnvalidator.py
+++ b/roles/lib_openshift/src/class/oc_sdnvalidator.py
@@ -1,0 +1,58 @@
+# pylint: skip-file
+# flake8: noqa
+
+# pylint: disable=too-many-instance-attributes
+class OCSDNValidator(OpenShiftCLI):
+    ''' Class to wrap the oc command line tools '''
+
+    def __init__(self, kubeconfig):
+        ''' Constructor for OCSDNValidator '''
+        # namespace has no meaning for SDN validation, hardcode to 'default'
+        super(OCSDNValidator, self).__init__('default', kubeconfig)
+
+    def get(self, kind, invalid_filter):
+        ''' return SDN information '''
+
+        rval = self._get(kind)
+        if rval['returncode'] != 0:
+            return False, rval, []
+
+        return True, rval, filter(invalid_filter, rval['results'][0]['items'])
+
+    # pylint: disable=too-many-return-statements
+    @staticmethod
+    def run_ansible(params):
+        ''' run the idempotent ansible code
+
+            params comes from the ansible portion of this module
+        '''
+
+        sdnvalidator = OCSDNValidator(params['kubeconfig'])
+        all_invalid = {}
+        failed = False
+
+        checks = (
+            (
+                'hostsubnet',
+                lambda x: x['metadata']['name'] != x['host'],
+                u'hostsubnets where metadata.name != host',
+            ),
+            (
+                'netnamespace',
+                lambda x: x['metadata']['name'] != x['netname'],
+                u'netnamespaces where metadata.name != netname',
+            ),
+        )
+
+        for resource, invalid_filter, invalid_msg in checks:
+            success, rval, invalid = sdnvalidator.get(resource, invalid_filter)
+            if not success:
+                return {'failed': True, 'msg': 'Failed to GET {}.'.format(resource), 'state': 'list', 'results': rval}
+            if invalid:
+                failed = True
+                all_invalid[invalid_msg] = invalid
+
+        if failed:
+            return {'failed': True, 'msg': 'All SDN objects are not valid.', 'state': 'list', 'results': all_invalid}
+
+        return {'msg': 'All SDN objects are valid.'}
--- a/roles/lib_openshift/src/doc/sdnvalidator
+++ b/roles/lib_openshift/src/doc/sdnvalidator
@@ -1,0 +1,27 @@
+# flake8: noqa
+# pylint: skip-file
+
+DOCUMENTATION = '''
+---
+module: oc_sdnvalidator
+short_description: Validate SDN objects
+description:
+  - Validate SDN objects
+options:
+  kubeconfig:
+    description:
+    - The path for the kubeconfig file to use for authentication
+    required: false
+    default: /etc/origin/master/admin.kubeconfig
+    aliases: []
+author:
+- "Mo Khan <monis@redhat.com>"
+extends_documentation_fragment: []
+'''
+
+EXAMPLES = '''
+oc_version:
+- name: get oc sdnvalidator
+  sdnvalidator:
+  register: oc_sdnvalidator
+'''
--- a/roles/lib_openshift/src/test/unit/oc_sdnvalidator.py
+++ b/roles/lib_openshift/src/test/unit/oc_sdnvalidator.py
@@ -1,0 +1,481 @@
+#!/usr/bin/env python2
+'''
+ Unit tests for oc sdnvalidator
+'''
+# To run
+# ./oc_sdnvalidator.py
+#
+# ....
+# ----------------------------------------------------------------------
+# Ran 4 tests in 0.002s
+#
+# OK
+
+import os
+import sys
+import unittest
+import mock
+
+# Removing invalid variable names for tests so that I can
+# keep them brief
+# pylint: disable=invalid-name,no-name-in-module
+# Disable import-error b/c our libraries aren't loaded in jenkins
+# pylint: disable=import-error
+# place class in our python path
+module_path = os.path.join('/'.join(os.path.realpath(__file__).split('/')[:-4]), 'library')  # noqa: E501
+sys.path.insert(0, module_path)
+from oc_sdnvalidator import OCSDNValidator  # noqa: E402
+
+
+class OCSDNValidatorTest(unittest.TestCase):
+    '''
+     Test class for OCSDNValidator
+    '''
+
+    @mock.patch('oc_sdnvalidator.Utils.create_tmpfile_copy')
+    @mock.patch('oc_sdnvalidator.OCSDNValidator._run')
+    def test_no_data(self, mock_cmd, mock_tmpfile_copy):
+        ''' Testing when both SDN objects are empty '''
+
+        # Arrange
+
+        # run_ansible input parameters
+        params = {
+            'kubeconfig': '/etc/origin/master/admin.kubeconfig',
+        }
+
+        empty = '''{
+    "apiVersion": "v1",
+    "items": [],
+    "kind": "List",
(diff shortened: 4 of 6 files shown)


enj:
As a check I ran the following playbook on a locally running cluster:

```yaml
#!/home/mkhan/mo/originpy/bin/ansible-playbook --module-path=../../../library/

---
- hosts: localhost
  gather_facts: no
  user: mkhan
  tasks:
  - name: Test1
    oc_sdnvalidator:
      kubeconfig: /home/mkhan/.kube/config
    register: oc_sdnvalidator
```

Valid (or no data) data results:

```

TASK [Test1] *******************************************************************
ok: [localhost]

localhost                  : ok=1    changed=0    unreachable=0    failed=0   
```

Invalid data results:

```

TASK [Test1] *******************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "All SDN objects are not valid.", "results": {"hostsubnets where metadata.name != host": [{"apiVersion": "v1", "host": "baz1", "hostIP": "1.1.1.1", "kind": "HostSubnet", "metadata": {"creationTimestamp": "2017-02-16T18:47:49Z", "name": "baz0", "namespace": "", "resourceVersion": "996", "selfLink": "/oapi/v1/hostsubnetsbaz0", "uid": "69f75f87-f478-11e6-aae0-507b9dac97ff"}, "subnet": "1.1.0.0/24"}], "netnamespaces where metadata.name != netname": [{"apiVersion": "v1", "kind": "NetNamespace", "metadata": {"creationTimestamp": "2017-02-16T18:45:52Z", "name": "bar0", "namespace": "", "resourceVersion": "969", "selfLink": "/oapi/v1/netnamespacesbar0", "uid": "245d416e-f478-11e6-aae0-507b9dac97ff"}, "netid": 100, "netname": "bar1"}]}, "state": "list"}
	to retry, use: --limit @/home/mkhan/go/src/github.com/openshift/openshift-ansible/roles/lib_openshift/src/test/integration/oc_sdnvalidator.retry

localhost                  : ok=0    changed=0    unreachable=0    failed=1   
```

`GET` failure results:

```

TASK [Test1] *******************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Failed to GET hostsubnet.", "results": {"cmd": "oc -n default get hostsubnet -o json", "results": [{}], "returncode": 1, "stderr": "Error from server (Forbidden): User \"developer\" cannot list all hostsubnets in the cluster\n", "stdout": "{\n    \"apiVersion\": \"v1\",\n    \"items\": [],\n    \"kind\": \"List\",\n    \"metadata\": {},\n    \"resourceVersion\": \"\",\n    \"selfLink\": \"\"\n}\n"}, "state": "list"}
	to retry, use: --limit @/home/mkhan/go/src/github.com/openshift/openshift-ansible/roles/lib_openshift/src/test/integration/oc_sdnvalidator.retry

localhost                  : ok=0    changed=0    unreachable=0    failed=1   
```

So it seems to be working as expected.

detiber:
@enj looks like there are some pylint and flake8 complaints from the travis run:
- https://travis-ci.org/openshift/openshift-ansible/jobs/202402556#L545
- https://travis-ci.org/openshift/openshift-ansible/jobs/202402556#L946

detiber:
@rhcarvalho Do you think this is something that we could incorporate into the validation work you are doing?

enj:
Not sure why `python setup.py generate_validation` failed on Travis - it works fine locally for me.

detiber:
@kwoodson can you look into the generate_validation error?

roles/lib_openshift/src/test/unit/oc_sdnvalidator.py
```diff
@@ -0,0 +1,481 @@
...
+import unittest
+import mock
+
+# Removing invalid variable names for tests so that I can
+# keep them brief
+# pylint: disable=invalid-name,no-name-in-module
+# Disable import-error b/c our libraries aren't loaded in jenkins
+# pylint: disable=import-error
+# place class in our python path
+module_path = os.path.join('/'.join(os.path.realpath(__file__).split('/')[:-4]), 'library')  # noqa: E501
```
rhcarvalho: Maybe something like this would be cleaner?

```python
os.path.join(os.path.dirname(__file__), os.pardir, os.pardir, ..., "library")
```

roles/lib_openshift/src/test/unit/oc_sdnvalidator.py
```diff
@@ -0,0 +1,481 @@
...
+ Unit tests for oc sdnvalidator
+'''
+# To run
+# ./oc_sdnvalidator.py
+#
+# ....
+# ----------------------------------------------------------------------
+# Ran 4 tests in 0.002s
+#
+# OK
```
rhcarvalho: Hmm, this comment seem unnecessary to me. How to run tests is described in CONTRIBUTING.md.
However, I'm not opposed to having the `if __name__ ...` at the end if that makes it convenient for you to run the tests.

roles/lib_openshift/src/class/oc_sdnvalidator.py
```diff
@@ -0,0 +1,58 @@
...
+
+        return True, rval, filter(invalid_filter, rval['results'][0]['items'])
+
+    # pylint: disable=too-many-return-statements
+    @staticmethod
+    def run_ansible(params):
+        ''' run the idempotent ansible code
+
+            params comes from the ansible portion of this module
+        '''
```
rhcarvalho: The indentation of docstrings seem quite inconsistent, I am surprised our linting tools don't bother.
https://www.python.org/dev/peps/pep-0257/

roles/lib_openshift/src/doc/sdnvalidator
```diff
@@ -0,0 +1,27 @@
+# flake8: noqa
+# pylint: skip-file
```
rhcarvalho: This is a Python file, shouldn't its name end with `.py`?
I see none of the other files in this directory have an extension, @kwoodson was there some idea behind that?

Sad to see that `lib_openshift` simply bypasses both flake8 and pylint...

roles/lib_openshift/src/ansible/oc_sdnvalidator.py
```diff
@@ -0,0 +1,24 @@
...
+    '''
+
+    module = AnsibleModule(
+        argument_spec=dict(
+            kubeconfig=dict(default='/etc/origin/master/admin.kubeconfig', type='str'),
+        ),
+        supports_check_mode=False,
+    )
+
+
```
rhcarvalho: Probably `flake8` would complain about having 2 blank lines instead of one.

roles/lib_openshift/src/ansible/oc_sdnvalidator.py
```diff
@@ -0,0 +1,24 @@
...
+    rval = OCSDNValidator.run_ansible(module.params)
+    if 'failed' in rval:
+        module.fail_json(**rval)
+
+    module.exit_json(**rval)
+
```
rhcarvalho: And here probably `flake8` would complain about having 1 blank line instead of 2.

@detiber are we opening an exception to all of `lib_openshift`?

rhcarvalho:
Hi @enj, I went ahead and did a quick pass review. Sorry for jumping in with questions, I'm looking forward to understanding better what's the approach folks like @detiber and @kwoodson are taking on reviewing the new code that is coming into the repo.

Let me know if I can be of any help.

roles/lib_openshift/src/ansible/oc_sdnvalidator.py
```diff
@@ -0,0 +1,24 @@
...
+# flake8: noqa
+
+def main():
+    '''
+    ansible oc module for validating OpenShift SDN objects
+    '''
+
+    module = AnsibleModule(
+        argument_spec=dict(
+            kubeconfig=dict(default='/etc/origin/master/admin.kubeconfig', type='str'),
```
detiber: There is a differentiation based on product name and version, as of 3.1 or 3.2 (I can't remember) everything standardized on 'etc/origin'. That said, if using `oc cluster up` or the images directly, the location will be different.

roles/lib_openshift/src/ansible/oc_sdnvalidator.py
detiber: I'm not sure if we want to open an exception to all of lib_openshift, but possibly to lib_{utils,openshift}/src.

rhcarvalho: :+1: We could ignore missing imports / names in those files, because they will be concatenated with others, but then we should probably run a more complete check on the generated files.

kwoodson: @rhcarvalho, we ignore all of the files under src/ but enforce all of the linting rules under /library/

roles/lib_openshift/src/test/unit/oc_sdnvalidator.py
```diff
@@ -0,0 +1,481 @@
...
+import mock
+
+# Removing invalid variable names for tests so that I can
+# keep them brief
+# pylint: disable=invalid-name,no-name-in-module
+# Disable import-error b/c our libraries aren't loaded in jenkins
+# pylint: disable=import-error
+# place class in our python path
+module_path = os.path.join('/'.join(os.path.realpath(__file__).split('/')[:-4]), 'library')  # noqa: E501
+sys.path.insert(0, module_path)
```
kwoodson: @rhcarvalho, thanks for this.  We should add a card to track and verify it works.

roles/lib_openshift/src/doc/sdnvalidator
kwoodson: @rhcarvalho, The reason we skip those files is that they are fragments and do not pass linters on their own.  The doc files are just strings.  They could be .py but since they are fragments and only contain doc, I left them without an extension.

> Sad to see that lib_openshift simply bypasses both flake8 and pylint...

I'm not sure I understand this.  `lib_openshift/library` is put through flake8 and pylint as well as `lib_openshift/src/test`.  Its only the fragments that are not.

kwoodson:
@enj,  Thanks for this pull request.  I think this approach to writing validation checks against the sdn is very powerful.  I'm not as familiar with the internals of openshift-ansible yet so I'm a little naive as to which role it will be used in but I think this is great work.  

It makes a lot of sense to me to have a simple call out to python and use our own product to perform validation.  I imagine we will take this pattern and use it for other validators in the future.  I wouldn't mind seeing some playbooks get refactored into modules that perform a series of checks. 

:+1:

roles/lib_openshift/src/doc/sdnvalidator
rhcarvalho: Sorry, I assumed the files were being concatenated as-is, without having looked at the code.
I see now that these lines are filtered out when generating the files: https://github.com/rhcarvalho/openshift-ansible/blob/master/roles/lib_openshift/src/generate.py#L72

rhcarvalho: These files contain variable assignments as well, and pragmas that are related to Python (`noqa`, `skip-file`).
Having the extension `.py` would enable automatic code highlighting on most text editors, make use of the pragmas to skip linting (without the extension I suspect pylint and flake8 don't even touch these files), and make it easier to figure out what the file is, and filter / find in the command line. E.g., we rely on the extension to find all Python files.

sdodson:
@enj can you rebase and re-run roles/lib_openshift/src/generate.py to clear the travis errors?

sdodson:
@kwoodson I can work on adding the playbooks around this module to gate upgrades. Can you do a final review when you get a chance?

kwoodson:
@sdodson sure, let me know.

detiber:
@sdodson @kwoodson I can get the molecule testing PR wrapped up by the end of the week: https://github.com/openshift/openshift-ansible/pull/2998/files

The issue we will have with testing in the current form is having a consistent OpenShift instance to test against. With the molecule PR, it tests against a docker container and will even run in Travis.

So far, I only have oc_secret done in the PR, and the PR needs to be rebased. But, the other tests wouldn't take long to migrate.

sdodson:
aos-ci-test

PR #3391, opened 2017-02-16
Add a configurable 30 second pause after draining a node to allow for rescheduling

sdodson:
This is a quick hack to provide marginal improvement and will require tuning per your environment's needs. 30 seconds may be too low for your environment.

In the future we should replace this with logic that actually evaluates whether or not pods have been rescheduled and are running on their new nodes.

I did some testing today on 3.2. What I found is that on the node being evacuated the pods immediately go terminating. On the nodes where they've been rescheduled how fast they come up depends entirely on how long it takes to pull the images. If the image was already there it was instantaneous. When the image had to be pulled my 400mb nodejs image was ready in a few seconds.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
@@ -29,6 +29,10 @@
       {{ hostvars[groups.oo_first_master.0].openshift.common.admin_binary }} drain {{ openshift.node.nodename | lower }} --force --delete-local-data
     delegate_to: "{{ groups.oo_first_master.0 }}"
 
+  - name: Pause for pod rescheduling during node ugprade
+    pause:
+      seconds: "{{ openshift_node_upgrade_pause | default(30) | int }}"
+
   roles:
   - lib_openshift
   - openshift_facts
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -106,6 +106,10 @@ openshift_release=v1.4
 #
 # Tasks to run after each master is upgraded and system/services have been restarted.
 # openshift_master_upgrade_post_hook=/usr/share/custom/post_master.yml
+#
+# Pause this number of seconds after draining each node to wait for rescheduling, this
+# may be too low for your environment, adjust accordingly
+# openshift_node_upgrade_pause=30
 
 
 # Alternate image format string, useful if you've got your own registry mirror
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -106,6 +106,10 @@ openshift_release=v3.4
 #
 # Tasks to run after each master is upgraded and system/services have been restarted.
 # openshift_master_upgrade_post_hook=/usr/share/custom/post_master.yml
+#
+# Pause this number of seconds after draining each node to wait for rescheduling, this
+# may be too low for your environment, adjust accordingly
+# openshift_node_upgrade_pause=30
 
 
 # Alternate image format string, useful if you've got your own registry mirror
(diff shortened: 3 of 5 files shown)


playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
```diff
@@ -29,6 +29,10 @@
       {{ hostvars[groups.oo_first_master.0].openshift.common.admin_binary }} drain {{ openshift.node.nodename | lower }} --force --delete-local-data
     delegate_to: "{{ groups.oo_first_master.0 }}"
 
+  - name: Pause for pod rescheduling during node ugprade
```
dgoodwin: Small spelling typo.

playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
```diff
@@ -29,6 +29,10 @@
...
+    pause:
+      seconds: "{{ openshift_node_upgrade_pause | default(30) | int }}"
+
```
dgoodwin: I'm trying to think if this is the right place or not. 

If we had say 3 compute nodes, and 2 replicas of some pod on N1 and N2, N3 is idle. We drain N1 for upgrade, kubernetes will kill our N1pod, and shortly thereafter realize it could distribute a replicate to N3. It might however take time to come up on N3 pulling images and such. So the problem arises if we proceed to upgrade N1 too quickly, then move on to N2 and drain it, killing our last running replica because N3 is still busy doing something to get the pod up, and now our app is down. 

Is this what we think happened? Just trying to keep it straight because this stuff can be confusing. 

Pausing after the drain seems reasonable, I'm wondering if it would be better to pause at the end of the node upgrade role instead though, as that would mean we only do this in one place, and could define a default for the seconds in the role. However it does feel like this is an "orchestration" pause and perhaps something that should be handled by the caller playbook. 

PR as is looks good to me, but throwing the above suggestion out there for comment.

sdodson: I wasn't entirely sure if the problem is because we start killing the node before pods have been successfully started elsewhere or if it was that we move on to the next node. I figured by placing it immediately after the drain operation it would address both cases.

ingvagabund: I can image the following scenario: Cluster has `n` nodes `node1`, `node2`, .... Then:
* The first `node1` is upgraded -> pod `important_pod` from `node1` is "rescheduled" to `node2`.
* The second `node2` is upgraded -> pod `important_pod` from `node2` is "rescheduled" to `node3`.
* The third `node3` is upgraded -> pod `important_pod` from `node3` is "rescheduled" to `node4`.
* and so on

Of course, the `important_pod` can be replaced with a group of important pods. So I would rather wait the timeout after the node is schedulable again. So you give a better chance to pending pods to land back on the node. In case the cluster is highly utilized and there is not enough nodes with sufficient resources to receive all drained pods.

sdodson: But the rescheduling happens immediately after you start the evacuation. In my test, which was only two nodes maybe it's different with more nodes, the pods were actually started on the other nodes before they'd fully terminated on the original node.

sdodson:
@derekwaynecarr Can you help us decide the best way to ensure we're performing in place upgrades in the least disruptive manner possible? Assumption is that we have capacity for N+X nodes worth of pods so when a X nodes are evacuated for upgrade all of the pods that were on them are able to be scheduled elsewhere. The problem was that we were evacuating, completing the node upgrade, and moving on to the next nodes before all of the rescheduled pods were back online causing disruption as more and more nodes were evacuated. This is a bit of a blunt instrument to provide some tunable amount of delay before bringing the next nodes down.

PR #3400, opened 2017-02-17
WIP - Ensure that openshift-ansible requires all the subpackages

sdodson:
Fixes Bug 1423444


sdodson:
Reading through the specfile I'm not sure this will actually fix the issue. I've requested more info, lets wait to merge this until we have that info.

Issue #3394, opened 2017-02-17
Fedora (AH): python (core) yaml


cgwalters:
I was looking at a run on current Fedora AH, which fails very quickly with:

```
TASK [openshift_facts : Gather Cluster facts and set is_containerized if needed] ***
task path: /srv/walters/src/github/openshift/openshift-ansible/roles/openshift_facts/tasks/main.yml:20
fatal: [node]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Shared connection to 192.168.121.166 closed.\r\n", "module_stdout": "Traceback (most recent call last):\r\n  File \"/tmp/ansible_SO8QZU/ansible_module_openshift_facts.py\", line 20, in <module>\r\n    import yaml\r\nImportError: No module named yaml\r\n", "msg": "MODULE FAILURE"}
```

Now...I think this is partly a Python 2 vs 3 thing, since on a current CentOS AH, we have:

```
# rpm -q --whatrequires PyYAML
atomic-1.14.1.12-8df71d7c6a33c2f7c95f4d4c72b623546fc82d55.2805e4f21d124fff45ffda5d817d81c80fa41afa.el7.centos.x86_64
cloud-init-0.7.5-10.el7.centos.1.x86_64
```

Yet on Fedora AH:

```
# rpm -q --whatrequires PyYAML
no package requires PyYAML
# rpm -q --whatrequires python3-PyYAML
cloud-init-0.7.8-3.fc25.noarch
atomic-1.14.1-1.fc25.x86_64
```

However, what confuses me is I know we at least had *some* runs since e.g. www.projectatomic.io/blog/2016/12/part1-install-origin-on-f25-atomic-host/  I even tried deploying the version of AH referenced there (`25.42`) thinking that perhaps we *did* previously pull in `PyYAML` but something got ported to Python 3, but AFAICS, nope.

So we'll have to decide a bit - do we carry PyYAML just for openshift-ansible?  How did Dusty's runs work?

dustymabe:
just ran this with latest upstream `openshift-ansible` 6490024 repo and I did [hit an error](https://github.com/openshift/openshift-ansible/issues/3396) but it was way into the install and not this issue at all. Are you running with python3 like I did in the blog post?

```
ansible-playbook -i myinventory playbooks/byo/config.yml -e 'ansible_python_interpreter=/usr/bin/python3'
```

cgwalters:
[...]

PR #3402, opened 2017-02-17
facts/main: Require Python 3 for Fedora, Python 2 everywhere else

cgwalters:
For a few reasons; among them that we currently have a dependency on `PyYAML`
which on Fedora Atomic Host isn't installed for Python 2 by default. Further,
many dependencies are being ported in Fedora to be Python 3.

Conversely, ensure that we're using Python 2 everywhere else (which is really
CentOS/RHEL), since AFAIK we don't test that path at all, and we really don't
need *more* entries in the supported matrix.

Related: https://github.com/openshift/openshift-ansible/issues/3394

### DIFF
--- a/roles/openshift_facts/tasks/main.yml
+++ b/roles/openshift_facts/tasks/main.yml
@@ -14,6 +14,18 @@
     l_is_master_system_container: "{{ (use_master_system_container | default(use_system_containers) | bool) }}"
     l_is_etcd_system_container: "{{ (use_etcd_system_container | default(use_system_containers) | bool) }}"
 
+- name: Validate python version
+  fail:
+    msg: |
+      openshift-ansible requires Python 3 for {{ ansible_distribution }};
+      For information on enabling Python 3 with Ansible, see https://docs.ansible.com/ansible/python_3_support.html
+  when: ansible_distribution == 'Fedora' and ansible_python['version']['major'] != 3
+
+- name: Validate python version
+  fail:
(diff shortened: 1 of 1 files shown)


ashcrow:
Do our e2e tests run with py3 as well?

sdodson:
No, they're currently all RHEL7.

dustymabe:
> No, they're currently all RHEL7.

can we work together to run tests against Fedora Atomic Host as part of the test framework?

detiber:
@dustymabe Is there a Fedora test infrastructure that we could leverage? I'd love to see some automated Fedora testing at least for the master branch.

dustymabe:
> @dustymabe Is there a Fedora test infrastructure that we could leverage? I'd love to see some automated Fedora testing at least for the master branch.

let's set up some time to talk about this. I'm not too worried about the infrastructure as there are a couple of different options and we can go with the one that makes the most sense.

cgwalters:
I was actually starting again on the "test on AH" path and decided to do Fedora, was working through issues.  We have plenty of resources around, we can test against Fedora/CentOS/whatever in CentOS CI or internal Jenkins or whatever.

maxamillion:
LGTM 👍

ashcrow:
aos-ci-test

detiber:
@dustymabe sounds good, feel free to throw something on my calendar. I'd like to discuss how we can work in Fedora and AH testing into the test work I've been doing as well: https://github.com/openshift/openshift-ansible/pull/2998/files

PR #3403, opened 2017-02-17
Compatibility for python2,3.

kwoodson:


kwoodson:
@ashcrow, great suggestion, let me fix those strings on the subprocess so it doesn't propagate through the code.

kwoodson:
@ashcrow, updated.

ashcrow:
aos-ci-test

PR #3404, opened 2017-02-17
post_control_plane.yml: don't fail on grep

giuseppe:
grep returns rc != 0 if the text is not found, the next rule assumes
that.  Do not fail on the check when the line is not found.


giuseppe:
aos-ci-test

sdodson:
that's a new puddle flake

PR #3406, opened 2017-02-17
Install repos on containerized hosts for python dependencies

sdodson:
- Removes dependency on openshift_facts
 - Adds Centos PaaS SIG common repo which is where origin users would get python2-ruamel-yaml
 - Makes lib_openshift and lib_utils depend on openshift_repos
 - Installs repos on containerized hosts for easier puddle testing

### DIFF
--- a/roles/lib_utils/meta/main.yml
+++ b/roles/lib_utils/meta/main.yml
@@ -1,0 +1,15 @@
+---
+galaxy_info:
+  author: TODO
+  description: OpenShift Repositories
+  company: Red Hat, Inc.
+  license: Apache License, Version 2.0
+  min_ansible_version: 1.7
+  platforms:
+  - name: EL
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- { role: openshift_repos }
--- a/roles/openshift_repos/meta/main.yml
+++ b/roles/openshift_repos/meta/main.yml
@@ -11,5 +11,4 @@ galaxy_info:
     - 7
   categories:
   - cloud
-dependencies:
-- { role: openshift_facts }
+dependencies: []
--- a/roles/openshift_repos/tasks/main.yaml
+++ b/roles/openshift_repos/tasks/main.yaml
@@ -1,50 +1,47 @@
 ---
-# TODO: Add flag for enabling EPEL repo, default to false
-
-# TODO: Add subscription-management config, with parameters
-#       for username, password, poolid(name), and official repos to
-#       enable/disable. Might need to make a module that extends the
-#       subscription management module to take a poolid and enable/disable the
-#       proper repos correctly.
+- name: openshift_repos detect ostree
+  stat:
+    path: /run/ostree-booted
+  register: ostree_booted
 
 - assert:
     that: openshift_deployment_type in known_openshift_deployment_types
-  when: not openshift.common.is_containerized | bool
+    msg: "openshift_deployment_type must be one of {{ known_openshift_deployment_types }}"
 
-- name: Ensure libselinux-python is installed
-  package: name=libselinux-python state=present
-  when: not openshift.common.is_containerized | bool
+- block:
+  - name: Ensure libselinux-python is installed
+    package: name=libselinux-python state=present
 
-- name: Create any additional repos that are defined
-  template:
-    src: yum_repo.j2
-    dest: /etc/yum.repos.d/openshift_additional.repo
-  when: openshift_additional_repos | length > 0 and not openshift.common.is_containerized | bool
-  notify: refresh cache
+  - name: Create any additional repos that are defined
+    template:
+      src: yum_repo.j2
+      dest: /etc/yum.repos.d/openshift_additional.repo
+    when:
+    - openshift_additional_repos | length > 0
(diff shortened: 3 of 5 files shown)


sdodson:
Need to sort out python2-ruamel-yaml versus python-ruamel-yaml dependency. I noticed F25 python2-ruamel-yaml package doesn't provide python-ruamel-yaml.

roles/openshift_repos/tasks/main.yaml
```diff
@@ -7,44 +7,51 @@
...
-        and not openshift.common.is_containerized | bool
-        and openshift_enable_origin_repo | default(true) | bool
+- name: openshift_repos detect ostree
+  stat:
+    path: /run/ostree-booted
+  register: ostree_booted
+
+- block:
+  - assert:
+      that: openshift_deployment_type in known_openshift_deployment_types
```
detiber: It almost seems like this assert should be outside of the block, even though the variable isn't used outside of the block.

The assert should probably have a `msg` defined (this wasn't available with Ansible 1.9).

roles/openshift_repos/tasks/main.yaml
```diff
@@ -7,44 +7,51 @@
...
+
+  - name: Configure origin yum repositories RHEL/CentOS
+    copy:
+      src: origin/repos/openshift-ansible-centos-paas-sig.repo
+      dest: /etc/yum.repos.d/
+    notify: refresh cache
+    when:
+    - ansible_os_family == "RedHat" and ansible_distribution != "Fedora"
+    - openshift_deployment_type == 'origin'
+    - openshift_enable_origin_repo | default(true) | bool
```
detiber: I'd condense these two copy tasks into a single task:
```
- name: Configure origin gpg keys if needed
  copy:
    src: origin/gpg_keys/openshift-ansible-CentOS-SIG-PaaS
    dest: /etc/pki/rpm-gpg/
  notify: refresh cache
  with_items:
  - src: origin/gpg_keys/openshift-ansible-CentOS-SIG-PaaS
    dest: /etc/pki/rpm-gpg/
  - src: origin/repos/openshift-ansible-centos-paas-sig.repo
    dest: /etc/yum.repos.d/
  when:
  - ansible_os_family == "RedHat"
  - ansible_distribution != "Fedora"
  - openshift_deployment_type == 'origin'
  - openshift_enable_origin_repo | default(true) | bool
```

I'm not entirely sure that we need to keep `ansible_os_family == "RedHat"`, but it doesn't hurt to leave it in place.

roles/openshift_repos/tasks/main.yaml
```diff
@@ -7,44 +7,51 @@
 #       subscription management module to take a poolid and enable/disable the
 #       proper repos correctly.
```
detiber: I think it's probably safe to delete the above TODO comments at this point

roles/openshift_repos/meta/main.yml
```diff
@@ -12,4 +12,3 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
```
detiber: This should be changed to: `dependencies: []`

roles/openshift_repos/tasks/main.yaml
```diff
@@ -7,44 +7,51 @@
...
+    - openshift_additional_repos | length > 0
+    notify: refresh cache
+
+  - name: Remove the additional repos if no longer defined
+    file:
+      dest: /etc/yum.repos.d/openshift_additional.repo
+      state: absent
+    when:
+    - openshift_additional_repos | length == 0
+    notify: refresh cache
```
sdodson: state doesn't seem to be valid for templates.

roles/openshift_repos/tasks/main.yaml
sdodson: updated

roles/openshift_repos/tasks/main.yaml
sdodson: updated

roles/openshift_repos/meta/main.yml
sdodson: updated

sdodson:
aos-ci-test

roles/lib_utils/meta/main.yml
```diff
@@ -0,0 +1,15 @@
...
+  license: Apache License, Version 2.0
+  min_ansible_version: 1.7
+  platforms:
+  - name: EL
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- { role: openshift_repos }
```
mtnbikenc: roles/lib_utils/meta/main.yml:
  15:28     error    trailing spaces  (trailing-spaces)

roles/openshift_repos/tasks/main.yaml
detiber: doh, of course it isn't.

roles/lib_utils/meta/main.yml
sdodson: thanks! fixed.

sdodson:
aos-ci-test

PR #3407, opened 2017-02-17
Updating node playbooks to use oc_obj

mtnbikenc:


roles/openshift_node_upgrade/tasks/main.yml
```diff
@@ -25,14 +28,19 @@
   when: not openshift.common.is_containerized | bool
 
 - name: Remove obsolete docker-sdn-ovs.conf
-  file: path=/etc/systemd/system/docker.service.d/docker-sdn-ovs.conf state=absent
-  when: (deployment_type == 'openshift-enterprise' and openshift_release | version_compare('3.4', '>=')) or (deployment_type == 'origin' and openshift_release | version_compare('1.4', '>='))
+  file:
+    path: "/etc/systemd/system/docker.service.d/docker-sdn-ovs.conf"
+    state: absent
+  when: (deployment_type == 'openshift-enterprise' and openshift_release | version_compare('3.4', '>='))
+     or (deployment_type == 'origin' and openshift_release | version_compare('1.4', '>='))
```
detiber: Would it make sense to have a filter_plugin for these types of deployment_type/version comparisons?

It's probably not worth it considering that we are going to bring the Origin versioning in line with the OCP versioning in the not to distant future.

roles/openshift_node_upgrade/tasks/main.yml
```diff
@@ -13,7 +13,10 @@
   vars:
     # We will restart Docker ourselves after everything is ready:
     skip_docker_restart: True
-  when: l_docker_upgrade is defined and l_docker_upgrade | bool and not openshift.common.is_atomic | bool
+  when:
+  - l_docker_upgrade is defined
+  - l_docker_upgrade | bool
+  - not openshift.common.is_atomic | bool
```
detiber: Not related to this PR directly, but I'm wondering if we should be skipping the docker restart based on `is_containerized` rather than `is_atomic`...

roles/openshift_node_upgrade/tasks/main.yml
mtnbikenc: If there is a change to versioning coming, it would make sense to me to refactor that at a future time.  I'm changing this line just for readability.

mtnbikenc:
aos-ci-test

PR #3409, opened 2017-02-17
Adding missing handler to resolve error that it was not found

ewolinetz:
To address
https://bugzilla.redhat.com/show_bug.cgi?id=1419811

### DIFF
--- a/roles/openshift_hosted_logging/handlers/main.yml
+++ b/roles/openshift_hosted_logging/handlers/main.yml
@@ -3,3 +3,24 @@
   systemd: name={{ openshift.common.service_type }}-master state=restarted
   when: (openshift.master.ha is not defined or not openshift.master.ha | bool) and (not (master_service_status_changed | default(false) | bool))
   notify: Verify API Server
+
+- name: Verify API Server
+  # Using curl here since the uri module requires python-httplib2 and
+  # wait_for port doesn't provide health information.
+  command: >
+    curl --silent --tlsv1.2
+    {% if openshift.common.version_gte_3_2_or_1_2 | bool %}
+    --cacert {{ openshift.common.config_base }}/master/ca-bundle.crt
+    {% else %}
+    --cacert {{ openshift.common.config_base }}/master/ca.crt
+    {% endif %}
+    {{ openshift.master.api_url }}/healthz/ready
+  args:
+    # Disables the following warning:
+    # Consider using get_url or uri module rather than running curl
+    warn: no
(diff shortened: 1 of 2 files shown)


ewolinetz:
aos-ci-test

roles/openshift_hosted_logging/handlers/main.yml
```diff
@@ -3,3 +3,24 @@
   systemd: name={{ openshift.common.service_type }}-master state=restarted
   when: (openshift.master.ha is not defined or not openshift.master.ha | bool) and (not (master_service_status_changed | default(false) | bool))
   notify: Verify API Server
+
+- name: Verify API Server
+  # Using curl here since the uri module requires python-httplib2 and
```
detiber: It looks like python-httplib2 is no longer a requirement as of Ansible 2.1.

I don't see a way to set the ca cert for the uri module, but I don't see much of an issue with disabling cert validation for this case.

ewolinetz: so, you think its alright to remain using curl, but remove the if block?

ewolinetz:
@detiber I was able to get this to work using the `uri` module, I updated the pr accordingly

roles/openshift_logging/handlers/main.yml
```diff
@@ -3,3 +3,14 @@
...
   when: (openshift.master.ha is not defined or not openshift.master.ha | bool) and (not (master_service_status_changed | default(false) | bool))
   notify: Verify API Server
+
+- name: Verify API Server
+  uri:
+    url: "{{ openshift.master.api_url }}/healthz/ready"
+    return_content: yes
+    status_code: 200
+  register: api_available_output
+  until: "'ok' in api_available_output.content"
```
detiber: @ewolinetz is there a case where the /healthz/ready endpoint would return 200, but not return ok in the content?

I think you could update the until condition to be: `until: api_available_output | succeeded`

ewolinetz: @detiber it might be that the end point is available but for some reason the api cluster isn't ready yet, not sure if it can return a different value.... but i'll make the update since it should be fairly safe

sdodson:
> @detiber I was able to get this to work using the uri module, I updated the pr accordingly

Are we updating all of the cases of API validation handlers to match? I'd prefer they all be the same, and honestly I'd land on the side of minimizing change right now unless there's a compelling advantage to using uri module.

PR #3412, opened 2017-02-18
Remove and document how to find dead Python code

rhcarvalho:
I've used `vulture` some years ago, and it doesn't seem to have evolved much since then.
It does produce false positives and negatives, and, thus, is not suitable for use in CI. Still, it can be useful sometimes during development, so it might be worth sharing its existence as documentation.

The dead test code was spotted manually while rewriting those files to use pytest instead of nose...


rhcarvalho:
aos-ci-test

Issue #3410, opened 2017-02-17
oc binary not in path


kwoodson:
Provide a brief description of your issue here. For example:

On certain installs such as AH, or by running as different users in the ansible inventory the oc binary is not in the PATH.  This is causing errors when attempting to run the modules:

`FileNotFoundError: [Errno 2] No such file or directory: 'oc'\n"`

##### Version

ansible 2.2

##### Steps To Reproduce
1. In the inventory file set your user to vagrant, ec2_user, or a user whose path does not have oc
2. Run any of the ansible lib_openshift modules.

##### Expected Results

I would expect that the modules run successfully.

##### Observed Results
When oc does not exist on the path the user sees: 
```
FileNotFoundError: [Errno 2] No such file or directory: 'oc'\n"
```

##### Additional Information

Currently the expectation for the oc binary is to live in the path of the user.  When that path does not define /usr/local/bin or any other path to where the binary is currently installed then we receive failures.

I think we can make an attempt at being a bit more resilient.  We can try and find the oc binary in the path, and if that fails, we can attempt a few "known" locations.
- ~/bin/
- /usr/local/bin

dustymabe:
yeah the key to this is that `/usr/local/bin/` isn't in the $PATH when running `sudo` so it doesn't work with some random user and `ansible_become=true`.

sdodson:
@thoraxe is the previous comment what you're running into?

detiber:
We can probably leverage the `get_bin_path` method from AnsibleModule to help:
```
oc_bin = module.get_bin_path('oc', opt_dirs=['/usr/local/bin', os.path.expanduser('~/bin/oc')])
```

cgwalters:
I'll work on this.

cgwalters:
Unless you are @kwoodson ?

cgwalters:
PR in https://github.com/openshift/openshift-ansible/pull/3413

thoraxe:
@sdodson if going sudo to root counts (as opposed to running a command with sudo) then yes I suppose it's the same problem. I didn't actually try to su to root.

rhcarvalho:
> ~/bin/

@kwoodson in this case, `~` should expand relative to what user?


---

Perhaps instead of changing the PATH of arbitrary users, or finding executables outside of PATH, wouldn't it be possible to arrange the playbooks in such a way that `oc`, etc, is in the PATH? I think that would be less puzzling, as it corresponds to how Linux works.

[...]

PR #3413, opened 2017-02-18
roles/lib_openshift: Handle /usr/local/bin/oc with sudo

cgwalters:
The real changes here are in `src/lib/{base,import}.py` remember, the rest is
committed to git rather than built for some reason.

This is tested on CentOS AH (python2), I haven't yet tested the Python 3 path
here. I tried the suggestion in the PR for using Ansible's `module` but AFAICS
that would require passing down the `module` variable pretty far down into this
code. This implementation seems OK too.

Closes: https://github.com/openshift/openshift-ansible/issues/3410

### DIFF
--- a/roles/lib_openshift/library/oadm_manage_node.py
+++ b/roles/lib_openshift/library/oadm_manage_node.py
@@ -36,6 +36,7 @@ import atexit
 import copy
 import json
 import os
+import sys
 import re
 import shutil
 import subprocess
@@ -963,10 +964,18 @@ class OpenShiftCLI(object):
     def openshift_cmd(self, cmd, oadm=False, output=False, output_type='json', input_data=None):
         '''Base command for oc '''
         cmds = []
-        if oadm:
-            cmds = ['oadm']
+        basecmd = 'oadm' if oadm else 'oc'
+        # https://github.com/openshift/openshift-ansible/issues/3410
+        # oc can be in /usr/local/bin in some cases, but that may not
+        # be in $PATH due to ansible/sudo
+        if sys.version_info[0] == 3:
+            basepath = shutil.which(basecmd) # pylint: disable=no-member
         else:
-            cmds = ['oc']
+            import distutils.spawn
+            basepath = distutils.spawn.find_executable(basecmd) # pylint: disable=no-name-in-module
+        if basepath is None and os.path.isfile('/usr/local/bin/' + basecmd):
+            basecmd = '/usr/local/bin/' + basecmd
+        cmds.append(basecmd)
 
         if self.all_namespaces:
             cmds.extend(['--all-namespaces'])
@@ -982,7 +991,10 @@ class OpenShiftCLI(object):
         if self.verbose:
             print(' '.join(cmds))
 
-        returncode, stdout, stderr = self._run(cmds, input_data)
+        try:
+            returncode, stdout, stderr = self._run(cmds, input_data)
+        except OSError as ex:
+            returncode, stdout, stderr = 1, '', 'Failed to execute {}: {}'.format(subprocess.list2cmdline(cmds), ex)
 
         rval = {"returncode": returncode,
                 "results": results,
--- a/roles/lib_openshift/library/oc_edit.py
+++ b/roles/lib_openshift/library/oc_edit.py
@@ -36,6 +36,7 @@ import atexit
 import copy
 import json
 import os
+import sys
 import re
 import shutil
 import subprocess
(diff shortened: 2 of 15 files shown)


kwoodson:
@cgwalters,  Thanks for taking a look at this.

I had a couple of thoughts regarding this.  My first thought is that we should probably set the oc command's path during the constructor and do it once per instantiation.  Regardless of which method we choose.  It is not uncommon that we make multiple calls out to openshift_cmd during our execution path.  I think a single call to setup the path might be cleaner and it being abstracted to its own function would make testing and mocking easier.

The other suggestion would be that the oc binary might not be in /usr/local/bin.  It could be somewhere else as is in the case with `oc cluster up`.  My suggestion would be to create a list of known PATH locations and pass them down into the environment:
`self.common_paths = ['/usr/local/bin', os.expand_path('~/bin')]`
`curr_env.update({'PATH': '%s:%s' % (curr_env['PATH'], ':'.join(self.common_paths)})`

_code_ _is_ _untested_

Another option is to add environment variables to ansible.  It can take them at different levels.  At the play level, task level, and the inventory file.

@abutcher, @twiest, @ashcrow, @tbielawa, @detiber 
I'd appreciate it if a few others could voice their opinion.

cgwalters:
Quick side note: Changing the current process' environment isn't threadsafe and will lead to very hard to debug crashes down the line.  See https://github.com/rust-lang/rust/pull/24741 for some links.  While I'm not sure if Ansible uses threads today, it's really better to be safe here.  Two Python projects that went through this pain recently are [Anaconda](https://github.com/rhinstaller/anaconda/commit/93b942e50e63873d2355f1f99d2801853abc978d) and we hit it in the atomic command too.

cgwalters:
Fixed an obvious logic error + some pylint issues ⬆️

cgwalters:
As far as searching other paths...something like this?  Any other dirs to add?

```
diff --git a/roles/lib_openshift/src/lib/base.py b/roles/lib_openshift/src/lib/base.py
index 0018a67..8733cbd 100644
--- a/roles/lib_openshift/src/lib/base.py
+++ b/roles/lib_openshift/src/lib/base.py
@@ -233,9 +233,18 @@ class OpenShiftCLI(object):
         else:
             import distutils.spawn
             basepath = distutils.spawn.find_executable(basecmd) # pylint: disable=no-name-in-module
-        if basepath is None and os.path.isfile('/usr/local/bin/' + basecmd):
-            basecmd = '/usr/local/bin/' + basecmd
-        cmds.append(basecmd)
+        if basepath is None:
+            for known_dir in ['/usr/local/bin/', os.path.expanduser('~/bin/')]:
+                path = known_dir + basecmd
+                if os.path.isfile(path):
+                    basepath = path
+                    break
+        if basepath is None:
+            # We didn't find it...maybe it is there somehow, but if not,
+            # we'll just let the execve below return an error.
+            cmds.append(basecmd)
+        else:
+            cmds.append(basepath)
 
         if self.all_namespaces:
             cmds.extend(['--all-namespaces'])
```

roles/lib_openshift/library/oadm_manage_node.py
```diff
@@ -963,10 +964,18 @@ def _run(self, cmds, input_data):
     def openshift_cmd(self, cmd, oadm=False, output=False, output_type='json', input_data=None):
         '''Base command for oc '''
         cmds = []
-        if oadm:
-            cmds = ['oadm']
+        basecmd = 'oadm' if oadm else 'oc'
```
detiber: Not related to this PR, but @kwoodson is there any reason that we need to keep compat with the `oadm` binary?

kwoodson: I don't think so. It should be `oc adm` since the modules are going that direction.

detiber:
@kwoodson @cgwalters I created https://github.com/openshift/openshift-ansible/pull/3416 that is the result of my working branch for testing some proposed changes to this PR. I pulled in #3414 to allow for better testing of the proposed changes as well.

detiber:
@kwoodson I actually wasn't aware of the env related issues that @cgwalters brought up. I do know that Ansible uses both multi-product and threading, so it is probably a good idea to convert existing env var use to command line options.

On a side note. Have you considered using a tool to wrap the subprocess management? sh provides a nice interface and helps with mocking for tests: https://amoffat.github.io/sh/. I think that would be post-3.5?

ashcrow:
I like @cgwalters code here, but agree it would be nice to have it in a function that could be called by the modules.

FWIW ``ansible`` does use traditional threads as well as multiprocessing:

```
$ grep -ri 'import thread' | wc -l
3
$ grep -ri 'import multiprocess' | wc -l
3
```

@kwoodson / @detiber I'd be happy to take a stab at it combining @detiber's tests, @cgwalters code and  meld it together into a reusable function.

detiber:
@ashcrow I'm working on updating my PR to @cgwalters branch now, since the test PR has been merged

detiber:
@cgwalters @ashcrow @kwoodson if there are no objections, I would like to close this PR in favor of #3416

ashcrow:
Since #3416 includes @cgwalters's changes and properly attributes him I have no objections.

rhcarvalho:
> Quick side note: Changing the current process' environment isn't threadsafe and will lead to very hard to debug crashes down the line.

@cgwalters @detiber 

The above applies to changing the environment of the current process. For the sake of clarity, I understand in this PR the idea revolved around setting up the environment for spawning a new process. That has no problem, neither in Python nor Rust.

Example of copying the current process' env, modifying it and spawning a new process:

https://gist.github.com/rhcarvalho/bc319af47fb505cf1d396c99f5ea9cd1

Notice `oc` is not in my regular PATH.

cgwalters:
Ah, I hadn't realized that `subprocess` would use the child's env for its own PATH lookups.  That seems like a reasonable approach indeed.

sdodson:
Superseded by #3416

PR #3414, opened 2017-02-18
Fix missing unit tests in CI

rhcarvalho:
@detiber @kwoodson while working on tests I realized that we were not running tests from:

- roles/lib_openshift/src/test/unit
- roles/lib_utils/src/test/unit

I've fixed two tests that failed under Python 3, ~~and would like your help to fix yet another 2 tests are are FAILING (roles/lib_openshift/src/test/unit/test_oc_serviceaccount_secret.py)~~ (fixed).

Another consequence of enabling those tests is that the coverage level drop from 86% down to 25%, due to the low coverage in the generated files. We currently require a 70% minimum coverage. ~~Suggestions welcome.~~ (decided to lower to 25% and raise up from there)


detiber:
@rhcarvalho submitted a PR to your fork/branch to separate out the lib_util and lib_openshift unit tests to a separate step with separate coverage report.

I also fixed the some missing dependencies, test file permissions,

rhcarvalho:
For the failing tests we can just mark them as failed and open a separate issue to fix them.

rhcarvalho:
> separate step with separate coverage report

Hmm, thinking a bit about this, the separate reports would help us ensure different bars, and make sure the coverage doesn't go below any of the two thresholds.
However, it adds some extra complexity to the test runner config.

I think it would be more honest to ourselves to acknowledge we had a drop when we brought in `lib_openshift` and `lib_utils`, and work our way to improve it if it is something important to us.

Introducing complexity to report independent coverage reports might just mean we'll not work to raise the bar again -- unless we intent to let the bar for `lib_openshift` and `lib_utils` be intentionally lower.

Thoughts?

rhcarvalho:
aos-ci-test

detiber:
> Hmm, thinking a bit about this, the separate reports would help us ensure different bars, and make sure the coverage doesn't go below any of the two thresholds.
> However, it adds some extra complexity to the test runner config.

Agreed, I only really went down this road to try and stem the bleeding a bit.

> I think it would be more honest to ourselves to acknowledge we had a drop when we brought in lib_openshift and lib_utils, and work our way to improve it if it is something important to us.

I think that is the right thing to do, but I was worried about potentially losing coverage for the previous code unintentionally.

> Introducing complexity to report independent coverage reports might just mean we'll not work to raise the bar again -- unless we intent to let the bar for lib_openshift and lib_utils be intentionally lower.

So, I think with the current testing strategy we are almost doomed to keep the bar lower. Instead of running unit tests against the generated code, I think we need to actually test the source files instead.

I noticed this issue when I tried to add testing for the binary lookup. Since the code was isolated to src/lib/base.py, I first tried to modify the tests to "compile" and load just the composition of src/lib/imports.py and src/lib/base.py, and I was able to test the scenarios that I wanted successfully, but it did not count towards any coverage. In order to get the coverage to increase, I ended up just brute forcing it by adding the tests to each of the existing test classes using copypasta. 

I think in order to get accurate coverage without making the tests a copypasta mess, it'll require reworking the generation in the following ways:
- Remove src/lib/imports.py alltogether
- Add imports directly to the src files using them, including imports for other src files
- Update the generate script logic to gather the imports from all the included files, dedupe, remove imports for other src files and insert the resulting list of imports into the generated file.

That will allow us to run the unit tests directly against the src files and get a more accurate coverage report for the files.

> Thoughts?

In addition to the above changes for unit testing lib_*, I was thinking we could switch to using coveralls.io for our coverage reports, which would give us a better way to track test coverage for individual files over time in addition to overall coverage. Then we could remove the forced coverage percentage and still work towards the desired goals.

@kwoodson interested if you see any issues with this suggested approach?

kwoodson:
@detiber,

This pull request seems fine.  In regards to the comments above I'm a little behind or maybe confused.  I'm not sure how introducing new items lowered our test coverage or the testing bar.  I'm excited to include these in the normal tests that run and feel like that is moving in the right direction.  

As for you thoughts on coveralls.io, I think that could be useful.  I'm not as familiar with it but the more tests we have the more confidence we'll have in our tools so that makes me a 👍.

As for unit testing source files, that is an interesting idea that we could discuss.  I'm not against it but I'd like to see what you had in mind.  The source files are kind of useless without some of the other pieces of the code, like for instance base or any other referenced files inside of lib/.  Again, let's discuss what the goals are and then we can come up with a solution for it.

Thanks for looking at this @detiber.

rhcarvalho:
@detiber @kwoodson thank you, we are in agreement to work on improving `lib_openshift`'s code generation and re-raise the coverage bar gradually. Could you please give your review approval?

roles/lib_utils/library/repoquery.py
```diff
@@ -477,7 +477,7 @@ def format_versions(self, formatted_versions):
         ''' Gather and present the versions of each package '''
 
         versions_dict = {}
-        versions_dict['available_versions_full'] = formatted_versions.keys()
+        versions_dict['available_versions_full'] = list(formatted_versions.keys())
```
rhcarvalho: @detiber @kwoodson I see I missed to update this same line of code in `roles/lib_utils/src/class/repoquery.py`, and CI did not fail... we possibly have a problem with the code generation verification.

rhcarvalho: https://github.com/openshift/openshift-ansible/pull/3426

PR #3416, opened 2017-02-19
lib_openshift - handle cases where oc binary is not in path

detiber:
Handle cases where the oc binary is not in the user path

- Move binary file search to separate method called from constructor
- Use six.PY3 instaed of sys.version_info
- Combine `/usr/local/bin` and `~/bin` with the paths from the environment
  - For py3 pass the combined paths to shutil.which
  - For py2 explictly search for existance of file from combined paths
    instead of using distutils.spawn.find_executable and falling back
- Use 'oc adm' instead of 'oadm'
- Fix generate_validation test issue
- fix tests for oc binary location
- add tests for file lookup

Closes: #3413 
Closes: #3410

### DIFF
--- a/roles/lib_openshift/library/oadm_manage_node.py
+++ b/roles/lib_openshift/library/oadm_manage_node.py
@@ -36,6 +36,7 @@ import atexit
 import copy
 import json
 import os
+import sys
 import re
 import shutil
 import subprocess
@@ -963,10 +964,18 @@ class OpenShiftCLI(object):
     def openshift_cmd(self, cmd, oadm=False, output=False, output_type='json', input_data=None):
         '''Base command for oc '''
         cmds = []
-        if oadm:
-            cmds = ['oadm']
+        basecmd = 'oadm' if oadm else 'oc'
+        # https://github.com/openshift/openshift-ansible/issues/3410
+        # oc can be in /usr/local/bin in some cases, but that may not
+        # be in $PATH due to ansible/sudo
+        if sys.version_info[0] == 3:
+            basepath = shutil.which(basecmd) # pylint: disable=no-member
         else:
-            cmds = ['oc']
+            import distutils.spawn
+            basepath = distutils.spawn.find_executable(basecmd) # pylint: disable=no-name-in-module
+        if basepath is None and os.path.isfile('/usr/local/bin/' + basecmd):
+            basecmd = '/usr/local/bin/' + basecmd
+        cmds.append(basecmd)
 
         if self.all_namespaces:
             cmds.extend(['--all-namespaces'])
@@ -982,7 +991,10 @@ class OpenShiftCLI(object):
         if self.verbose:
             print(' '.join(cmds))
 
-        returncode, stdout, stderr = self._run(cmds, input_data)
+        try:
+            returncode, stdout, stderr = self._run(cmds, input_data)
+        except OSError as ex:
+            returncode, stdout, stderr = 1, '', 'Failed to execute {}: {}'.format(subprocess.list2cmdline(cmds), ex)
 
         rval = {"returncode": returncode,
                 "results": results,
--- a/roles/lib_openshift/library/oadm_manage_node.py
+++ b/roles/lib_openshift/library/oadm_manage_node.py
@@ -36,7 +36,6 @@ import atexit
 import copy
 import json
 import os
-import sys
 import re
 import shutil
 import subprocess
@@ -749,6 +748,32 @@ class OpenShiftCLIError(Exception):
     pass
 
 
+ADDITIONAL_PATH_LOOKUPS = ['/usr/local/bin', os.path.expanduser('~/bin')]
+
+
+def locate_oc_binary():
+    ''' Find and return oc binary file '''
+    # https://github.com/openshift/openshift-ansible/issues/3410
+    # oc can be in /usr/local/bin in some cases, but that may not
+    # be in $PATH due to ansible/sudo
+    paths = os.environ.get("PATH", os.defpath).split(os.pathsep) + ADDITIONAL_PATH_LOOKUPS
+
+    oc_binary = 'oc'
+
+    # Use shutil.which if it is available, otherwise fallback to a naive path search
+    try:
+        which_result = shutil.which(oc_binary, path=os.pathsep.join(paths))
+        if which_result is not None:
+            oc_binary = which_result
+    except AttributeError:
+        for path in paths:
+            if os.path.exists(os.path.join(path, oc_binary)):
+                oc_binary = os.path.join(path, oc_binary)
+                break
+
+    return oc_binary
+
+
 # pylint: disable=too-few-public-methods
 class OpenShiftCLI(object):
     ''' Class to wrap the command line tools '''
@@ -762,6 +787,7 @@ class OpenShiftCLI(object):
         self.verbose = verbose
         self.kubeconfig = Utils.create_tmpfile_copy(kubeconfig)
         self.all_namespaces = all_namespaces
+        self.oc_binary = locate_oc_binary()
 
     # Pylint allows only 5 arguments to be passed.
     # pylint: disable=too-many-arguments
@@ -963,19 +989,10 @@ class OpenShiftCLI(object):
     # pylint: disable=too-many-arguments,too-many-branches
     def openshift_cmd(self, cmd, oadm=False, output=False, output_type='json', input_data=None):
         '''Base command for oc '''
-        cmds = []
-        basecmd = 'oadm' if oadm else 'oc'
-        # https://github.com/openshift/openshift-ansible/issues/3410
-        # oc can be in /usr/local/bin in some cases, but that may not
-        # be in $PATH due to ansible/sudo
-        if sys.version_info[0] == 3:
-            basepath = shutil.which(basecmd) # pylint: disable=no-member
-        else:
-            import distutils.spawn
-            basepath = distutils.spawn.find_executable(basecmd) # pylint: disable=no-name-in-module
-        if basepath is None and os.path.isfile('/usr/local/bin/' + basecmd):
-            basecmd = '/usr/local/bin/' + basecmd
-        cmds.append(basecmd)
+        cmds = [self.oc_binary]
+
+        if oadm:
+            cmds.append('adm')
 
         if self.all_namespaces:
             cmds.extend(['--all-namespaces'])
--- a/roles/lib_openshift/src/lib/base.py
+++ b/roles/lib_openshift/src/lib/base.py
@@ -224,10 +224,18 @@ class OpenShiftCLI(object):
     def openshift_cmd(self, cmd, oadm=False, output=False, output_type='json', input_data=None):
         '''Base command for oc '''
         cmds = []
-        if oadm:
-            cmds = ['oadm']
+        basecmd = 'oadm' if oadm else 'oc'
+        # https://github.com/openshift/openshift-ansible/issues/3410
+        # oc can be in /usr/local/bin in some cases, but that may not
+        # be in $PATH due to ansible/sudo
+        if sys.version_info[0] == 3:
+            basepath = shutil.which(basecmd) # pylint: disable=no-member
         else:
-            cmds = ['oc']
+            import distutils.spawn
+            basepath = distutils.spawn.find_executable(basecmd) # pylint: disable=no-name-in-module
+        if basepath is None and os.path.isfile('/usr/local/bin/' + basecmd):
+            basecmd = '/usr/local/bin/' + basecmd
+        cmds.append(basecmd)
 
         if self.all_namespaces:
             cmds.extend(['--all-namespaces'])
@@ -243,7 +251,10 @@ class OpenShiftCLI(object):
         if self.verbose:
             print(' '.join(cmds))
 
-        returncode, stdout, stderr = self._run(cmds, input_data)
+        try:
+            returncode, stdout, stderr = self._run(cmds, input_data)
+        except OSError as ex:
+            returncode, stdout, stderr = 1, '', 'Failed to execute {}: {}'.format(subprocess.list2cmdline(cmds), ex)
 
         rval = {"returncode": returncode,
                 "results": results,
--- a/roles/lib_openshift/src/lib/base.py
+++ b/roles/lib_openshift/src/lib/base.py
@@ -9,6 +9,32 @@ class OpenShiftCLIError(Exception):
     pass
 
 
+ADDITIONAL_PATH_LOOKUPS = ['/usr/local/bin', os.path.expanduser('~/bin')]
+
+
+def locate_oc_binary():
+    ''' Find and return oc binary file '''
+    # https://github.com/openshift/openshift-ansible/issues/3410
+    # oc can be in /usr/local/bin in some cases, but that may not
+    # be in $PATH due to ansible/sudo
+    paths = os.environ.get("PATH", os.defpath).split(os.pathsep) + ADDITIONAL_PATH_LOOKUPS
+
+    oc_binary = 'oc'
+
+    # Use shutil.which if it is available, otherwise fallback to a naive path search
+    try:
+        which_result = shutil.which(oc_binary, path=os.pathsep.join(paths))
+        if which_result is not None:
+            oc_binary = which_result
+    except AttributeError:
+        for path in paths:
+            if os.path.exists(os.path.join(path, oc_binary)):
+                oc_binary = os.path.join(path, oc_binary)
+                break
+
+    return oc_binary
+
+
 # pylint: disable=too-few-public-methods
 class OpenShiftCLI(object):
     ''' Class to wrap the command line tools '''
@@ -22,6 +48,7 @@ class OpenShiftCLI(object):
         self.verbose = verbose
         self.kubeconfig = Utils.create_tmpfile_copy(kubeconfig)
         self.all_namespaces = all_namespaces
+        self.oc_binary = locate_oc_binary()
 
     # Pylint allows only 5 arguments to be passed.
     # pylint: disable=too-many-arguments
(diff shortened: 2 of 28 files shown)


detiber:
aos-ci-test

rhcarvalho:
@detiber this is a rather large PR, to you intend it to replace the items you mentioned or is this just for testing your changes to #3413?

detiber:
@rhcarvalho It was only meant to show working tests for the proposed changes and to track ensuring that we end up with a working set of tests for the binary lookup.

detiber:
@rhcarvalho I take that back, I'm re-purposing this PR to supersede #3414, but I have rebased it based off of your merged PR.

detiber:
aos-ci-test

roles/lib_openshift/library/oadm_manage_node.py
```diff
@@ -761,6 +762,32 @@ def __init__(self,
...
+        self.oc_binary = self._locate_oc_binary()
+
+    @staticmethod
+    def _locate_oc_binary():
+        ''' Find and return oc binary file '''
+        # https://github.com/openshift/openshift-ansible/issues/3410
+        # oc can be in /usr/local/bin in some cases, but that may not
+        # be in $PATH due to ansible/sudo
+        paths = os.environ.get("PATH", os.defpath).split(os.pathsep)
+        extra_paths = ['/usr/local/bin', os.path.expanduser('~/bin')]
```
ashcrow: No change needed. Just FYI: this could be compressed to:

```python
paths = list(set(os.environ.get(
    "PATH", os.defpath).split(os.pathsep) + ['/usr/local/bin', os.path.expanduser('~/bin')]))
```

or (for easier reading):

```python
paths = os.environ.get("PATH", os.defpath).split(os.pathsep) + ['/usr/local/bin', os.path.expanduser('~/bin')]
paths = list(set(paths))
```

roles/lib_openshift/library/oadm_manage_node.py
```diff
@@ -761,6 +762,32 @@ def __init__(self,
...
+        for path in extra_paths:
+            if path not in paths:
+                paths.append(path)
+
+        oc_binary = 'oc'
+        if six.PY3:
```
ashcrow: Please invert this logic. Python 2 should be the 'catch and modify' since it's legacy.

roles/lib_openshift/library/oadm_manage_node.py
```diff
@@ -963,10 +990,11 @@ def _run(self, cmds, input_data):
     def openshift_cmd(self, cmd, oadm=False, output=False, output_type='json', input_data=None):
         '''Base command for oc '''
         cmds = []
+
+        cmds.append(self.oc_binary)
```
ashcrow: Go ahead and start the ``cmds`` list with ``self.oc_binary``.

roles/lib_openshift/library/oadm_manage_node.py
```diff
@@ -982,7 +1010,10 @@ def openshift_cmd(self, cmd, oadm=False, output=False, output_type='json', input
         if self.verbose:
             print(' '.join(cmds))
 
-        returncode, stdout, stderr = self._run(cmds, input_data)
+        try:
+            returncode, stdout, stderr = self._run(cmds, input_data)
+        except OSError as ex:
+            returncode, stdout, stderr = 1, '', 'Failed to execute {}: {}'.format(subprocess.list2cmdline(cmds), ex)
```
ashcrow: :+1:

roles/lib_utils/src/class/repoquery.py
```diff
@@ -60,7 +60,7 @@ def format_versions(self, formatted_versions):
         ''' Gather and present the versions of each package '''
 
         versions_dict = {}
-        versions_dict['available_versions_full'] = formatted_versions.keys()
+        versions_dict['available_versions_full'] = list(formatted_versions.keys())
```
ashcrow: :+1:

detiber:
@ashcrow comments addressed.

ashcrow:
aos-ci-test

roles/lib_openshift/src/lib/base.py
```diff
@@ -22,6 +22,29 @@ def __init__(self,
...
+        self.oc_binary = self._locate_oc_binary()
+
+    @staticmethod
+    def _locate_oc_binary():
+        ''' Find and return oc binary file '''
+        # https://github.com/openshift/openshift-ansible/issues/3410
+        # oc can be in /usr/local/bin in some cases, but that may not
+        # be in $PATH due to ansible/sudo
+        paths = os.environ.get("PATH", os.defpath).split(os.pathsep) + ['/usr/local/bin', os.path.expanduser('~/bin')]
+        paths = list(set(paths))
```
rhcarvalho: Isn't the order important here? `list(set(seq))` loses the order.

roles/lib_openshift/src/lib/base.py
```diff
@@ -22,6 +22,29 @@ def __init__(self,
...
+
+        oc_binary = 'oc'
+        if six.PY2:
```
rhcarvalho: This type of code raises a red flag on my brain. It is dangerous to carry different implementations for Python 3 and Python 2, specially without good automated tests around it.

Couldn't we resist the temptation to using `shutil.which` from Python 3 and have a single implementation that works for both Py2/3?

roles/lib_openshift/src/lib/base.py
```diff
@@ -22,6 +22,29 @@ def __init__(self,
         self.verbose = verbose
         self.kubeconfig = Utils.create_tmpfile_copy(kubeconfig)
         self.all_namespaces = all_namespaces
+        self.oc_binary = self._locate_oc_binary()
+
+    @staticmethod
```
rhcarvalho: Maybe this could be just a plain function, don't need to be a method?

roles/lib_openshift/src/test/unit/test_oadm_manage_node.py
```diff
@@ -179,6 +180,114 @@ def test_schedulable_false(self, mock_openshift_cmd, mock_tmpfile_copy):
...
+    @mock.patch('os.path.exists')
+    @mock.patch('os.environ.get')
+    def test_binary_lookup_fallback(self, mock_env_get, mock_path_exists):
+        ''' Testing binary lookup fallback '''
+
+        mock_env_get.side_effect = lambda _v, _d: ''
+
+        mock_path_exists.side_effect = lambda _: False
+
+        self.assertEqual(ManageNode._locate_oc_binary(), 'oc')
```
rhcarvalho: I think these tests miss the fact that `oc` might be available in multiple directories in PATH, and the order is important, we should get the first.

roles/lib_openshift/src/test/unit/test_oadm_manage_node.py
```diff
@@ -179,6 +180,114 @@ def test_schedulable_false(self, mock_openshift_cmd, mock_tmpfile_copy):
...
+    def test_binary_lookup_in_usr_local(self, mock_env_get, mock_path_exists):
+        ''' Testing binary lookup in /usr/local/bin '''
+
+        oc_bin = '/usr/local/bin/oc'
+
+        mock_env_get.side_effect = lambda _v, _d: '/bin:/usr/bin'
+
+        mock_path_exists.side_effect = lambda f: f == oc_bin
+
+        self.assertEqual(ManageNode._locate_oc_binary(), oc_bin)
```
rhcarvalho: There's so much code duplication here that pytest's parametrization would come in handy. Every method looks pretty much the same.

roles/lib_openshift/src/test/unit/test_oc_env.py
```diff
@@ -435,6 +451,114 @@ def test_removing_env_vars(self, mock_cmd, mock_tmpfile_copy):
             mock.call(['oc', '-n', 'default', 'get', 'dc', 'router', '-o', 'json'], None),
         ])
 
+    @unittest.skipIf(six.PY3, 'py2 test only')
+    @mock.patch('os.path.exists')
+    @mock.patch('os.environ.get')
+    def test_binary_lookup_fallback(self, mock_env_get, mock_path_exists):
```
rhcarvalho: Ahhh... I think this is the copypasta tests for increasing test coverage? Yeah, I don't like to add more code... more code == harder to maintain. IMHO duplicate tests don't really increase the quality of the project.

roles/lib_openshift/src/lib/base.py
detiber: Ah, that was actually why I took the first approach.

That said, we could always avoid the `list(set(paths))` since the worst case is that we check 2 directories a second time.

roles/lib_openshift/src/lib/base.py
detiber: I do not see why not, I'll update.

roles/lib_openshift/src/lib/base.py
detiber: @rhcarvalho we *could*. 

Ideally, we'd be able to use a library, but there doesn't seem to be one that we can just use without some work.

RHEL/CentOS/Fedora have python-which, but it is old and doesn't appear to be maintained. 

Fedora has [python2-backports-shutil_which](https://koji.fedoraproject.org/koji/packageinfo?packageID=23667), which would give us a backported version of `shutil.which`, but we would need to do the CentOS/RHEL packaging.

Thoughts on going down this path until we could get python2-backports-shutil_which backported?

The code implementation is small, but I'd prefer not to be in the business of managing a which implementation: https://github.com/minrk/backports.shutil_which

roles/lib_openshift/src/test/unit/test_oadm_manage_node.py
detiber: Indeed, but really we don't need to re-test a lot of these methods for each generated module. I was planning on trying to poc some generation changes that would let us test the source independently.

roles/lib_openshift/src/test/unit/test_oc_env.py
detiber: Agreed, it was basically a brute force attempt until a better approach is available. 

I could limit the test to just a single module.

roles/lib_openshift/src/lib/base.py
ashcrow: Agreed. A function would be cleaner, but I won't block on it.

roles/lib_openshift/src/lib/base.py
ashcrow: I don't believe the order matters in this case, but I could be wrong.

roles/lib_openshift/src/test/unit/test_oadm_manage_node.py
ashcrow: Is there a 'if ``oc`` is here it should take precedence" path?

detiber: I think so.

The main purpose of this is to fix the issue with '/usr/local/bin' being dropped from the path when sudo is used.

A secondary purpose (~/bin) is to support common non-admin user and development use cases.

In both cases, I don't see why we would want to override the existing path lookup behavior on the host, so I think we should preserve the order.

detiber:
aos-ci-test

roles/lib_openshift/src/lib/base.py
rhcarvalho: The way `which` and executable lookup works is directly dependent on and influenced by the order of the items in $PATH.

roles/lib_openshift/src/lib/base.py
rhcarvalho: > Ideally, we'd be able to use a library, but there doesn't seem to be one that we can just use without some work.

If 4~6 lines of Python code is the implementation used for Py2, why can't we use that for Py3 as well? We don't need an extra dependency nor maintaining packages.

roles/lib_openshift/src/test/unit/test_oadm_manage_node.py
rhcarvalho: Where's the test testing that the order is relevant? It would have been a failing test.

Now that the purpose is more clear, isn't what we're trying to do augmenting/replacing the root's $PATH with the user's $PATH?

Security issues aside, I think that would've been easier to simple augment/replace env["PATH"] whenever we call a command with `sudo`. Are we talking about calls to external code from Python code? Is it code we control or part of Ansible?

If we need to have this "locate_oc" API for whatever reason, we may use `distutils.spawn.find_executable`, available in Py2/Py3, or simple equivalent code:

http://stackoverflow.com/a/33773405/4804690

ashcrow:
@detiber for cleanliness don you kind squashing both "address comments" commits? Other than that LGTM.

roles/lib_openshift/library/oadm_manage_node.py
```diff
@@ -748,6 +748,29 @@ class OpenShiftCLIError(Exception):
     pass
 
 
+def locate_oc_binary():
```
kwoodson: I'd prefer it if this moved to a static method under the Utils class.  Seems like the perfect place for it.

roles/lib_openshift/library/oadm_manage_node.py
```diff
@@ -748,6 +748,29 @@ class OpenShiftCLIError(Exception):
...
+    ''' Find and return oc binary file '''
+    # https://github.com/openshift/openshift-ansible/issues/3410
+    # oc can be in /usr/local/bin in some cases, but that may not
+    # be in $PATH due to ansible/sudo
+    paths = os.environ.get("PATH", os.defpath).split(os.pathsep) + ['/usr/local/bin', os.path.expanduser('~/bin')]
```
kwoodson: I'd like to see the `['/usr/local/bin', os.path.expanduser('~/bin')]` defined in a variable so we can reference it, update it, etc.  I could see us using that variable in another way at some point in the future.

roles/lib_openshift/library/oadm_manage_node.py
ashcrow: Personally I tend to dislike static methods unless they are absolutely necessary. I'm not against moving it to another location. Could it go in the same module as utils as a function?

roles/lib_openshift/library/oadm_manage_node.py
ashcrow: Good idea! :+1:

roles/lib_openshift/src/lib/base.py
```diff
@@ -9,6 +9,29 @@ class OpenShiftCLIError(Exception):
     pass
 
 
+def locate_oc_binary():
+    ''' Find and return oc binary file '''
+    # https://github.com/openshift/openshift-ansible/issues/3410
+    # oc can be in /usr/local/bin in some cases, but that may not
+    # be in $PATH due to ansible/sudo
+    paths = os.environ.get("PATH", os.defpath).split(os.pathsep) + ['/usr/local/bin', os.path.expanduser('~/bin')]
```
ashcrow: Good point on sudo / root. This would need to move to happen before the change (which I'm guessing is unlikely) or use something like:


```python
os.path.expanduser('~' + os.getlogin()) + os.path.sep + 'bin'
```

detiber: So, I would expect that it would use the user's $HOME if not running under sudo, and root's $HOME if running under sudo. I know that is the default behavior outside of Ansible, but I haven't validated it with Ansible.

roles/lib_openshift/src/lib/base.py
```diff
@@ -243,7 +266,10 @@ def openshift_cmd(self, cmd, oadm=False, output=False, output_type='json', input
         if self.verbose:
             print(' '.join(cmds))
 
-        returncode, stdout, stderr = self._run(cmds, input_data)
+        try:
+            returncode, stdout, stderr = self._run(cmds, input_data)
```
detiber: > - No need for a locate_oc_binary function -- it is too specific, it seems what we want is to update $PATH and let Python/ the shell do its default operation.

@rhcarvalho that was discussed previously and with Ansible using both multi-process and threading, we decided to avoid potential issues with manipulating the env: https://github.com/openshift/openshift-ansible/pull/3413#issuecomment-280856499

> - Change _run to update the PATH, just like we set KUBECONFIG (curr_env.update({'KUBECONFIG': self.kubeconfig}))

We've discussed moving away from manipulating the env for KUBECONFIG as well, and moving towards the command line option instead, because of the discussion linked above.

> And if we're touching _run... there's usually a simpler way than using subprocess.Popen: https://docs.python.org/2/library/subprocess.html#using-the-subprocess-module (see call, check_call and check_output).

+1, though I've also wondered if we should move to something like [sh](https://amoffat.github.io/sh/). It would need to be packaged for RHEL, since it is only in Fedora/EPEL currently.

roles/lib_openshift/library/oadm_manage_node.py
detiber: @ashcrow Utils is defined in src/lib/base.py, which is where `locate_oc_binary()` is already defined.

roles/lib_openshift/src/lib/base.py
rhcarvalho: @detiber the discussion you linked to seems to be related to changing the env of the current process. Here the case is different, we are setting up the env for a process we are about to spawn.

rhcarvalho: I think we can live without a dependency on `sh`. It's not our core business to call external commands. I'd stick with the stdlib.

roles/lib_openshift/library/oadm_manage_node.py
ashcrow: Then I like it as is. But I won't block if others would like it moved.

detiber:
aos-ci-test

roles/lib_openshift/library/oadm_manage_node.py
detiber: @kwoodson @ashcrow updated

roles/lib_openshift/library/oadm_manage_node.py
```diff
@@ -748,6 +748,32 @@ class OpenShiftCLIError(Exception):
     pass
 
 
+ADDITIONAL_PATH_LOOKUPS = ['/usr/local/bin', os.path.expanduser('~/bin')]
```
ashcrow: :+1:

sdodson:
When I merged this travis test had failed to report status but when I viewed the test in travis it was green, but i guess that was wrong. Anyway. #3464 to fix that

detiber:
@ashcrow @rhcarvalho will follow up with a PR to address some of the remaining issues later.

PR #3418, opened 2017-02-19
Remove dead code in installer

rhcarvalho:
Found using `vulture`, confirmed that code that was once used became
unused in later commits after its introduction.


rhcarvalho:
aos-ci-test

PR #3419, opened 2017-02-19
Migrate from nose to pytest

rhcarvalho:
#2998 is introducing a dependency on `pytest` because of the `molecule` framework for integration tests.
Instead of having both `nose` and `pytest`, it is an appropriate time to migrate to `pytest`.

Pytest is more actively maintained than nose, has a vibrant community, plugins, etc.

It also allows us to write more concise test code, and produces good reports on errors.

This PR removes code-level dependencies on nose, rewrite tests to use pytest fixtures, and switch test execution / CI to use pytest, eliminating the dependency on nose for both the top-level and installer wrapper (utils/).

### DIFF
--- a/roles/openshift_master_facts/test/openshift_master_facts_bad_input_tests.py
+++ b/roles/openshift_master_facts/test/openshift_master_facts_bad_input_tests.py
@@ -1,0 +1,58 @@
+import copy
+import os
+import sys
+
+from ansible.errors import AnsibleError
+from nose.tools import raises
+
+sys.path.insert(1, os.path.join(os.path.dirname(__file__), os.pardir, "lookup_plugins"))
+
+from openshift_master_facts_default_predicates import LookupModule  # noqa: E402
+
+
+class TestOpenShiftMasterFactsBadInput(object):
+    def setUp(self):
+        self.lookup = LookupModule()
+        self.default_facts = {
+            'openshift': {
+                'common': {}
+            }
+        }
+
+    @raises(AnsibleError)
+    def test_missing_openshift_facts(self):
+        facts = {}
+        self.lookup.run(None, variables=facts)
+
+    @raises(AnsibleError)
+    def test_missing_deployment_type(self):
+        facts = copy.deepcopy(self.default_facts)
+        facts['openshift']['common']['short_version'] = '10.10'
+        self.lookup.run(None, variables=facts)
+
+    @raises(AnsibleError)
+    def test_missing_short_version_and_missing_openshift_release(self):
+        facts = copy.deepcopy(self.default_facts)
+        facts['openshift']['common']['deployment_type'] = 'origin'
+        self.lookup.run(None, variables=facts)
+
+    @raises(AnsibleError)
+    def test_unknown_deployment_types(self):
+        facts = copy.deepcopy(self.default_facts)
+        facts['openshift']['common']['short_version'] = '1.1'
+        facts['openshift']['common']['deployment_type'] = 'bogus'
+        self.lookup.run(None, variables=facts)
+
+    @raises(AnsibleError)
+    def test_unknown_origin_version(self):
+        facts = copy.deepcopy(self.default_facts)
+        facts['openshift']['common']['short_version'] = '0.1'
+        facts['openshift']['common']['deployment_type'] = 'origin'
+        self.lookup.run(None, variables=facts)
+
+    @raises(AnsibleError)
+    def test_unknown_ocp_version(self):
+        facts = copy.deepcopy(self.default_facts)
+        facts['openshift']['common']['short_version'] = '0.1'
+        facts['openshift']['common']['deployment_type'] = 'openshift-enterprise'
+        self.lookup.run(None, variables=facts)
--- a/roles/openshift_master_facts/test/openshift_master_facts_bad_input_tests.py
+++ b/roles/openshift_master_facts/test/openshift_master_facts_bad_input_tests.py
@@ -3,7 +3,7 @@ import os
 import sys
 
 from ansible.errors import AnsibleError
-from nose.tools import raises
+import pytest
 
 sys.path.insert(1, os.path.join(os.path.dirname(__file__), os.pardir, "lookup_plugins"))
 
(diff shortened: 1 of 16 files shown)


tox.ini
```diff
@@ -12,8 +12,28 @@ deps =
...
     yamllint: python setup.py yamllint
-    unit: nosetests
     generate_validation: python setup.py generate_validation
+
+[pytest]
+testpaths =
+    # TODO(rhcarvalho): enable these after https://github.com/openshift/openshift-ansible/pull/3414 merges.
+    #roles/lib_openshift/src/test/unit
+    #roles/lib_utils/src/test/unit
+    roles/openshift_master_facts/test
```
detiber: I'm wondering if we can improve the autodetection a bit, I hate that we would have to add paths to pick up new tests...  Could we take advantage of norecursedirs to explicitly skip directories that we know are problematic.

roles/openshift_master_facts/test/openshift_master_facts_bad_input_tests.py
```diff
@@ -0,0 +1,57 @@
+import copy
+import os
+import sys
+
+from ansible.errors import AnsibleError
+import pytest
+
```
detiber: Any objections to adding `pytestmark = pytest.mark.unit` to allow for filtering out just the unit tests?

That would also allow to use marks for specifying functional tests, slow tests, etc to filter the tests that are run.

tox.ini
rhcarvalho: Yeah, having to remember to add new directories is problematic, e.g., we missed `lib_openshift` and `lib_utils`.

The purpose of specifying the dirs is making test discovery faster. Maybe we can add some pre-test step that updates the configuration programmatically?!

roles/openshift_master_facts/test/openshift_master_facts_bad_input_tests.py
rhcarvalho: That's a great idea, will do.

rhcarvalho:
aos-ci-test

tox.ini
detiber: How much time are we saving on test discovery? If it's significant then it might not be worth it. If it's only saving a second or less of execution time, then I think the benefit outweighs the cost.

We could also tie into setup tools like we do for yamllint, etc

roles/openshift_master_facts/test/openshift_master_facts_bad_input_tests.py
detiber: I'm good with segregating the tests by directory as well.

I don't think we should centralize tests, though. I think bundling the tests with the roles, playbooks, etc they belong to gives us a way to ensure that they can stand on their own. It also makes it easier to match tests to those components for contributors.

tox.ini
rhcarvalho: I replaced `testpaths` with `omit = utils`, and then run tox/pytest with `--collect-only`.
The discovery alone can take > 15s.

I think that's too much, considering the time the tests actually run.

But I've found a different problem, it seems that test discovery gets lost because of how we have lots of symlinks. The same pytest run I mention failed with errors that include:

```
...
=================== ERRORS ====================
______________ ERROR collecting  ______________
.tox/py35-ansible22-unit/lib/python3.5/site-packages/_pytest/config.py:325: in _getconftestmodules
    return self._path2confmods[path]
E   KeyError: local('/home/rodolfo/openshift/src/github.com/openshift/openshift-ansible/playbooks/aws/openshift-cluster/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/ex
amples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expi
ry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate
_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certif
icate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_c
ertificate_expiry/examples/playbooks/roles/openshift_master_facts/test')

During handling of the above exception, another exception occurred:
.tox/py35-ansible22-unit/lib/python3.5/site-packages/_pytest/config.py:356: in _importconftest
    return self._conftestpath2mod[conftestpath]
E   KeyError: local('/home/rodolfo/openshift/src/github.com/openshift/openshift-ansible/playbooks/aws/openshift-cluster/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/ex
amples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expi
ry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate
_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certif
icate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_certificate_expiry/examples/playbooks/roles/openshift_c
ertificate_expiry/examples/playbooks/roles/openshift_master_facts/test/conftest.py')

During handling of the above exception, another exception occurred:
.tox/py35-ansible22-unit/lib/python3.5/site-packages/py/_path/common.py:366: in visit
    for x in Visitor(fil, rec, ignore, bf, sort).gen(self):
.tox/py35-ansible22-unit/lib/python3.5/site-packages/py/_path/common.py:415: in gen
    for p in self.gen(subdir):
.tox/py35-ansible22-unit/lib/python3.5/site-packages/py/_path/common.py:415: in gen
    for p in self.gen(subdir):
...
```

Looks like some sort of infinite recursion... I should go to sleep ;)

rhcarvalho:
> We could also tie into setup tools like we do for yamllint, etc

I tried to integrate `pytest` with setup.py for utils, but it wasn't very good. In particular, enabling code coverage didn't work, I ran into https://github.com/pytest-dev/pytest-runner/issues/18.

So calling `pytest` alone was much cleaner. I also would like to perhaps have a single config to run all tests without the utils/others segregation in the future, so using the same mechanism is a good start.

What do we lose for not using setup.py?

tox.ini
detiber: I worked up a solution in https://github.com/rhcarvalho/openshift-ansible/pull/10 that I think is much more tenable.

detiber:
I don't think we lose much with not using setup.py.

rhcarvalho:
@detiber ptal

sdodson:
aos-ci-test

sdodson:
Please squash down any un-necessary commits.

detiber:
aos-ci-test

detiber:
@sdodson We should probably update the Jenkins jobs to use tox for the unit tests (or remove them now that Travis handles it)

detiber:
@tbielawa @ashcrow any issues you guys see with migrating to pytest from nose? We are going to end up using pytest with molecule anyway.

ashcrow:
@detiber no issues I can see. It's a pretty common move as well.

PR #3421, opened 2017-02-20
rename registered mktemp command

jhadvig:
When running `playbooks/byo/config.yml` playbook, I've encountered following error:
```
PLAY [Delete temporary directory on localhost] *********************************
TASK [file] ********************************************************************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'stdout'\n\nThe error appears to have been in '/home/ec2-user/openshift-ansible/playbooks/common/openshift-node/config.yml': line 112, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  tasks:\n  - file: name={{ mktemp.stdout }} state=absent\n  ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes.  Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n - {{ foo }}\n\nShould be written as:\n\n    with_items:\n - \"{{ foo }}\"\n"}
```

Fixed by renaming the registered cmd name, since it look like it collides with the build-in commands... but not really sure, not a ansible expert (yet :P)

@abutcher PTAL

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -20,17 +20,6 @@
         annotations: "{{ openshift_node_annotations | default(None) }}"
         schedulable: "{{ openshift_schedulable | default(openshift_scheduleable) | default(None) }}"
 
-- name: Create temp directory for syncing certs
-  hosts: localhost
-  connection: local
-  become: no
-  gather_facts: no
-  tasks:
-  - name: Create local temp directory for syncing certs
-    local_action: command mktemp -d /tmp/openshift-ansible-XXXXXXX
-    register: mktemp
-    changed_when: False
-
 - name: Evaluate node groups
   hosts: localhost
   become: no
@@ -101,13 +90,4 @@
   tasks:
   - name: Create group for deployment type
     group_by: key=oo_nodes_deployment_type_{{ openshift.common.deployment_type }}
-    changed_when: False
-
-- name: Delete temporary directory on localhost
-  hosts: localhost
-  connection: local
-  become: no
-  gather_facts: no
-  tasks:
-  - file: name={{ mktemp.stdout }} state=absent
     changed_when: False


jhadvig:
Also found out that this registration appears in several playbooks, eg:
https://github.com/jhadvig/openshift-ansible/blob/master/roles/openshift_hosted/tasks/main.yml#L3-L5

Shouldn't we update those as well ?

playbooks/common/openshift-node/config.yml
```diff
@@ -28,7 +28,7 @@
   tasks:
   - name: Create local temp directory for syncing certs
     local_action: command mktemp -d /tmp/openshift-ansible-XXXXXXX
-    register: mktemp
+    register: maketemp
```
abutcher: We can remove these two temporary directory plays entirely. I just tested and they are no longer used.

abutcher:
@jhadvig Thanks! We can definitely go ahead and rename those other instances of `mktemp` found. Maybe we should name after the role in which they're used like `hosted_tmpdir` or so?

sdodson:
Is mktemp an ansible internal thing or we're just stomping on things across roles?

detiber:
> Is mktemp an ansible internal thing or we're just stomping on things across roles?
@sdodson I think role scoping is probably saving us from stomping on things in the general use case. It would probably bite us with combined playbook/role use, though.

detiber:
@sdodson @abutcher Since we probably won't be able to completely remove mktemp use in the short term, I wonder if it would be a good idea to vendor the tempfile module until we can require Ansible >= 2.3

jhadvig:
@abutcher so shall I remove those two places and rename rest ? or rename all of it ?
thanks so much !

abutcher:
@jhadvig Right, let's remove the two plays which are currently changed in this PR at least. If the other instances of `mktemp` aren't causing problems then we can leave them for now or rename them in this PR.

jhadvig:
@abutcher I've updated the PR... ptal :)

abutcher:
aos-ci-test

PR #3423, opened 2017-02-20
[openshift_hosted] Router/Registry

mtnbikenc:


roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -1,15 +1,14 @@
...
+  oc_obj:
+    state: list
+    kind: node
+    selector: "{{ openshift.hosted.registry.selector | default('') }}"
+  register: registry_nodes
   when: openshift.hosted.registry.replicas | default(none) is none
 
 - set_fact:
-    l_node_count: "{{ (registry_nodes_json.stdout | default('{\"items\":[]}') | from_json)['items'] | length }}"
+    l_node_count: "{{ registry_nodes.results.results[0]['items'] | length }}"
```
detiber: Any reason we couldn't wrap this all in a block and use vars for this instead of set_fact?

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -1,49 +1,35 @@
...
   oc_service:
     namespace: default
     name: docker-registry
     state: list
   register: docker_registry_service_ip
-  changed_when: false
 
-- set_fact:
+- name: Set fact docker_registry_route_hostname
+  set_fact:
```
detiber: Same comment here about using block/vars instead of set_fact.

roles/openshift_hosted/tasks/router/router.yml
```diff
@@ -1,7 +1,9 @@
 ---
 - fail:
```
detiber: Not related to this PR, but it might be good to switch to use assert for these use cases rather than fail.

mtnbikenc:
aos-ci-test

smarterclayton:
I'm pretty sure this broke origin-gce: 

https://ci.openshift.redhat.com/jenkins/job/zz_origin_gce_image/136/

```
TASK [openshift_hosted : Retrieve list of openshift nodes matching router selector] ***
Friday 24 February 2017  23:17:41 +0000 (0:00:00.032)       0:13:47.120 ******* 
skipping: [ci-primg136-ig-m-8sqp]

TASK [openshift_hosted : set_fact replicas] ************************************
Friday 24 February 2017  23:17:41 +0000 (0:00:00.036)       0:13:47.157 ******* 
fatal: [ci-primg136-ig-m-8sqp]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'results'\n\nThe error appears to have been in '/usr/share/ansible/openshift-ansible/roles/openshift_hosted/tasks/router/router.yml': line 11, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: set_fact replicas\n  ^ here\n"}
```

We set

```
openshift_hosted_router_replicas: 1
openshift_hosted_registry_replicas: 1
```

kwoodson:
@smarterclayton,

My guess is that the search for your selector from the previous step failed.  https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/tasks/router/router.yml#L2-L9

Would need to see what happened in that step.  

@mtnbikenc, looks like we should do some output validation when doing that query for nodes with selector.

smarterclayton:
The Origin GCE setup sets:

```
openshift_registry_selector: "role=infra"
openshift_router_selector: "role=infra"
```

and there is one node with that label.

smarterclayton:
Skipped because openshift_router_selector is set, it looks like.  

```
TASK [openshift_hosted : Retrieve list of openshift nodes matching router selector] ***
task path: /usr/share/ansible/openshift-ansible/roles/openshift_hosted/tasks/router/router.yml:2
Monday 27 February 2017  01:15:05 +0000 (0:00:00.041)       0:16:56.905 *******
skipping: [ci-claytontest-ig-m-r3tx] => {
    "changed": false,
    "skip_reason": "Conditional check failed",
    "skipped": true
}

TASK [openshift_hosted : set_fact replicas] ************************************
task path: /usr/share/ansible/openshift-ansible/roles/openshift_hosted/tasks/router/router.yml:11
Monday 27 February 2017  01:15:05 +0000 (0:00:00.040)       0:16:56.946 *******
fatal: [ci-claytontest-ig-m-r3tx]: FAILED! => {
    "failed": true,
    "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'results'\n\nThe error appears to have been in '/usr/share/ansible/openshift-ansible/roles/openshift_hosted/tasks/router/router.yml': line 11, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: set_fact replicas\n  ^ here\n"
}
```

smarterclayton:
Can we fast track a fix on monday morning?  This is blocking the switch to 3.6 in origin and upgrade test automation.

kwoodson:
@smarterclayton,

Absolutely.  Can you debug of this variable before the set_fact?

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/tasks/router/router.yml#L8

Add this to line 10:
`- debug: var=router_nodes`

smarterclayton:
It's not set - the conditional isn't set because the route selector is.

smarterclayton:
`set_fact replicas` should have a conditional 

```
  when: openshift.hosted.router.replicas | default(none) is none
```

kwoodson:
I'll look at this now.

Issue #3396, opened 2017-02-17
oc_serviceaccount module error on Fedora 25 Atomic Host


dustymabe:
`ansible-2.2.0.0-3.fc24.noarch`
openshift-ansible at 6490024
#### Description

I hit what I think is a python3 error on Fedora Atomic Host commit 5a9de04163a60c5f2771b3d307601fa4e2234f12b662c97e28290b7bcbdb43f0. I'm running using the python3 interpreter like so: `ansible-playbook -i myinventory playbooks/byo/config.yml -e 'ansible_python_interpreter=/usr/bin/python3'`. Here is the error:

```
TASK [openshift_manageiq : Create Admin and Image Inspector Service Account] ***
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: a bytes-like object is required, not 'str'
failed: [192.168.122.101] (item=management-admin) => {
    "failed": true, 
    "item": "management-admin", 
    "module_stderr": "Traceback (most recent call last):\n  File \"/tmp/ansible_xdj8mg5a/ansible_module_oc_serviceaccount.py\", line 1617, in <module>\n    main()\n  File \"/tmp/ansible_xdj8mg5a/ansible_module_oc_serviceaccount.py\", line 1610, in main\n    rval = OCServiceAccount.run_ansible(module.params, module.check_mode)\n  File \"/tmp/ansible_xdj8mg5a/ansible_module_oc_serviceaccount.py\", line 1516, in run_ansible\n    api_rval = oc_sa.get()\n  File \"/tmp/ansible_xdj8mg5a/ansible_module_oc_serviceaccount.py\", line 1452, in get\n    elif '\\\"%s\\\" not found' % self.config.name in result['stderr']:\nTypeError: a bytes-like object is required, not 'str'\n", 
    "module_stdout": ""
}

MSG:

MODULE FAILURE

An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: a bytes-like object is required, not 'str'
failed: [192.168.122.101] (item=inspector-admin) => {
    "failed": true, 
    "item": "inspector-admin", 
[...]

PR #3424, opened 2017-02-20
Replace has_key in new modules

ashcrow:
The dict.has_key has been removed in Python 3. However, both Python 2
and 3 support:

    if needle in dict

See: #3396

### DIFF
--- a/roles/lib_openshift/library/oc_secret.py
+++ b/roles/lib_openshift/library/oc_secret.py
@@ -1429,7 +1429,7 @@ class Secret(Yedit):
     def update_secret(self, key, value):
         ''' update a secret'''
         # pylint: disable=no-member
-        if self.secrets.has_key(key):
+        if key in self.secrets:
             self.secrets[key] = value
         else:
             self.add_secret(key, value)
@@ -1467,7 +1467,7 @@ class OCSecret(OpenShiftCLI):
         if results['returncode'] == 0 and results['results'][0]:
             results['exists'] = True
             if self.decode:
-                if results['results'][0].has_key('data'):
+                if 'data' in results['results'][0]:
                     for sname, value in results['results'][0]['data'].items():
                         results['decoded'][sname] = base64.b64decode(value)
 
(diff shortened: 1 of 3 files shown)


detiber:
@ashcrow I know pylint has a py3 mode that will only test for py3 compatibility issues. I wonder if that would have caught this error. Either that or possibly a flake8 plugin that could do the same.

detiber:
arrgh, it looks like `pylint --py3k` only complains about `no-absolute-import`

detiber:
Even worse than that, it looks like we are disabling pylint tests that would help catch these types of things:
- https://github.com/ashcrow/openshift-ansible/blob/d4a424e51e4cce0e3a5a4618f696800ebe655d25/.pylintrc#L63

ashcrow:
@detiber While this tests I'm going to see if we can use ``2to3`` as part of testing.

ashcrow:
aos-ci-test

ashcrow:
@detiber ``2to3`` looks like it will work :smile:

detiber:
@ashcrow I'm just wondering if there is an existing tool we can leverage instead of adding yet another tool :)

ashcrow:
@detiber good point. Do you know if we have ``python-tools`` installed?

```
$ rpm -q --whatprovides /usr/bin/2to3 
python-tools-2.7.13-1.fc24.x86_64
```

detiber:
@ashcrow it looks like we do have it installed, I'm more concerned with adding additional test invocations.

For 'linting', we currently run:
- pylint
- flake8
- yamllint
- generate_validation

If 2to3 is the best way to do it, then we can always add it to the list, we set the deps for the test environments. It'd be nice to be able to leverage one of the existing tools if at all possible instead, though

detiber:
Really, I would like to see if we could consolidate on either pylint or flake8 instead of running both, or at least tweak our configs to avoid running the same tests with both.

detiber:
@ashcrow ah, looks like pylint would have caught these issues, but we are disabling the `no-member` check which would have caught it.

ashcrow:
@detiber 

```
$ grep -ri 'no-member' . | grep -v tox | wc -l
246
```

Wooo, that will be interesting :sweat_smile:

detiber:
@ashcrow excluding the generated modules, it brings the count down to 23

PR #3426, opened 2017-02-20
Fix validation of generated code

rhcarvalho:
Importing modules with the same name will not reload an already imported
module. Explicitly reloading the module is the quickest fix.


rhcarvalho:
aos-ci-test

PR #3428, opened 2017-02-20
openshift_facts: handle 'latest' version

giuseppe:
do not attempt to compare the version using LooseVersion when 'latest'
is used.  The case openshift_version == None already sets to use the
last version, so simply reuse that.


giuseppe:
aos-ci-test

giuseppe:
I've noticed it with Python3.  Python2 doesn't raise any error.

@sdodson

PR #3429, opened 2017-02-20
Removed unrequired no-members

ashcrow:


roles/lib_utils/src/class/yedit.py
```diff
@@ -1,5 +1,5 @@
 # flake8: noqa
-# pylint: skip-file
+# pylint: disable=undefined-variable,missing-docstring
```
detiber: I think there will be other issues with pylint if we don't skip the src files for now.

ashcrow:
aos-ci-test

roles/lib_utils/src/class/yedit.py
ashcrow: Thankfully it seems fine :fireworks:

ashcrow:
@detiber I'll pull f68bd8844ad9e71f0b247574ff3e7eedd0bac528 if needed. PTAL.

ashcrow:
Added @sdodson as a reviewer as well since both @detiber and I have commits in this PR.

ashcrow:
aos-ci-test

detiber:
@rhcarvalho ptal

detiber:
Doh, I'm looking into the extended test failures

.pylintrc
```diff
@@ -60,7 +60,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=import-star-module-level,old-octal-literal,oct-method,print-statement,unpacking-in-except,parameter-unpacking,backtick,old-raise-syntax,old-ne-operator,long-suffix,dict-view-method,dict-iter-method,metaclass-assignment,next-method-called,raising-string,indexing-exception,raw_input-builtin,long-builtin,file-builtin,execfile-builtin,coerce-builtin,cmp-builtin,buffer-builtin,basestring-builtin,apply-builtin,filter-builtin-not-iterating,using-cmp-argument,useless-suppression,range-builtin-not-iterating,suppressed-message,no-absolute-import,old-division,cmp-method,reload-builtin,zip-builtin-not-iterating,intern-builtin,unichr-builtin,reduce-builtin,standarderror-builtin,unicode-builtin,xrange-builtin,coerce-method,delslice-method,getslice-method,setslice-method,input-builtin,round-builtin,hex-method,nonzero-method,map-builtin-not-iterating
+#disable=
```
rhcarvalho: Wow, this was a long list ;-)

roles/lib_utils/library/yedit.py
```diff
@@ -380,14 +380,16 @@ def write(self):
...
+        except AttributeError:
+            pass
 
-            # pylint: disable=no-member
-            Yedit._write(self.filename, yaml.dump(self.yaml_dict, Dumper=yaml.RoundTripDumper))
-        else:
+        # Try to use RoundTripDumper if supported.
+        try:
+            Yedit._write(self.filename, yaml.safe_dump(self.yaml_dict, Dumper=yaml.RoundTripDumper))
+        except AttributeError:
```
rhcarvalho: :+1:  "beg forgiveness, don't ask permission"

roles/openshift_health_checker/openshift_checks/mixins.py
```diff
@@ -2,17 +2,14 @@
 from openshift_checks import get_var
 
 
-class NotContainerized(object):
+class NotContainerizedMixIn(object):
```
rhcarvalho: Hmm, I had intently opted not for using the word "mixin" here. It will be referenced by folks writing checks, it affects readability.

IMHO this reads better

```python
class PackageAvailability(NotContainerized, OpenShiftCheck):
```

than

```python
class PackageAvailability(NotContainerizedMixIn, OpenShiftCheck):
```

I preferred to "pay the fee" in 1 file by having the pylint pragma, than having a worse name in N files.

ashcrow: I generally prefer having MixIn classes denote themselves by name. How does it affect readability in openshift-ansible land?

ashcrow:
aos-ci-test

roles/openshift_health_checker/openshift_checks/mixins.py
rhcarvalho: I read the first class declaration as "**Package Availability is a Not Containerized OpenShift Check**".
It could have been "My Custom Check is a Enterprise Only Not Containerized OpenShift Check":

```python
class MyCustomCheck(EnterpriseOnly, NotContainerized, OpenShiftCheck):
```


Adding "Mixin" suffixes is just noise.

"MixIn" with camel case is even more strange, the term is a single word AFAIK: https://en.wikipedia.org/wiki/Mixin

---

In Python, perhaps expect for exceptions ;), we don't normally prefix or suffix names with some part of its "type". We don't say FooClass, or BarAbstractInterface, or IReader (I think this might be the convention in other languages like Java, C# or whatnot).

.pylintrc
detiber: Indeed, it was basically the full list of python3 compat tests.

roles/openshift_health_checker/openshift_checks/mixins.py
detiber: @rhcarvalho It looks like [The Hitchhiker's Guide to Python](https://books.google.com/books?id=4yXtDAAAQBAJ&lpg=PT218&ots=XuaM-p9fXR&dq=python%20mixin%20naming%20convention&pg=PT218#v=onepage&q=python%20mixin%20naming%20convention&f=false) is on @ashcrow's side here for naming.

ashcrow: I'll update with ``Mixin`` rather than ``MixIn``.

rhcarvalho: ;) probably we should get rid of the mixins as you suggested originally, @detiber hehe

If we used class hierarchy:

```python
class PackageAvailability(NotContainerizedOpenShiftCheck):
    pass

class NotContainerizedOpenShiftCheck(OpenShiftCheck):
    pass
```

I still think, in this particular context, the suffix adds noise, but won't fight for it.

ashcrow: I'm a fan of going the hierarchy route. I like mixins but not when traditional 1:1 hierarchy works. I'll put together a PR after this to switch it up assuming there is no objections.

ashcrow:
aos-ci-test

roles/openshift_health_checker/openshift_checks/mixins.py
rhcarvalho: That doesn't work for what we want here, because we don't want to create a class for every combination of the ways how one can determine whether a check is active or not.

Right now we only have `NotContainerized` implemented, but the stage is set to have other things like `EnterpriseOnly`, `ContainerizedOnly`, or such.

If not using Mixins, the next level of sophistication I'd go into is having something like a list of `Condition`s, and evaluating them. Something I can think of, thought is in a different level of sophistication and implementation is https://docs.djangoproject.com/en/1.10/ref/models/conditional-expressions/, but more generally a way to combine functions that given some input tell whether a check applies or not...

Dreaming...

```python

class PackageAvailability(OpenShiftCheck):

    name = "package_availability"
    tags = ["preflight"]

    constraints = [NotContainerized, Origin, ..., ...]

    # ...

# in the check runner:

check_class = PackageAvailability
if all(constraint(...) for constraint in check_class.constraints):
     result = check_class(...).run(...)
```

rhcarvalho:
@ashcrow I appreciate you taking the time to review and improve the health check code, thanks!

ashcrow:
```
TASK [openshift_manageiq : Create Admin and Image Inspector Service Account] ***
failed: [10.8.168.178] (item=management-admin) => {"failed": true, "item": "management-admin", "module_stderr": "Shared connection to 10.8.168.178 closed.\r\n", "module_stdout": "Traceback (most recent call last):\r\n  File \"/tmp/ansible_TVYHPG/ansible_module_oc_serviceaccount.py\", line 1640, in <module>\r\n    main()\r\n  File \"/tmp/ansible_TVYHPG/ansible_module_oc_serviceaccount.py\", line 1633, in main\r\n    rval = OCServiceAccount.run_ansible(module.params, module.check_mode)\r\n  File \"/tmp/ansible_TVYHPG/ansible_module_oc_serviceaccount.py\", line 1572, in run_ansible\r\n    api_rval = oc_sa.create()\r\n  File \"/tmp/ansible_TVYHPG/ansible_module_oc_serviceaccount.py\", line 1487, in create\r\n    return self._create_from_content(self.config.name, self.config.data)\r\n  File \"/tmp/ansible_TVYHPG/ansible_module_oc_serviceaccount.py\", line 801, in _create_from_content\r\n    yed.write()\r\n  File \"/tmp/ansible_TVYHPG/ansible_module_oc_serviceaccount.py\", line 327, in write\r\n    Yedit._write(self.filename, yaml.safe_dump(self.yaml_dict, Dumper=yaml.RoundTripDumper))\r\n  File \"/usr/lib64/python2.7/site-packages/ruamel/yaml/main.py\", line 278, in safe_dump\r\n    return dump_all([data], stream, Dumper=SafeDumper, **kwds)\r\nTypeError: dump_all() got multiple values for keyword argument 'Dumper'\r\n", "msg": "MODULE FAILURE"}
```

detiber:
@ashcrow looks like the generated modules are out of date.

roles/openshift_health_checker/openshift_checks/mixins.py
detiber: I agree that going with a hierarchy here is the wrong approach... 

That said, I'm not sure that I'm a fan of the mixin approach here either. Maybe we should go with decorators?

On a more serious note, I wonder if it would make sense to do something crazy with metaclasses, where we can register test "plugins" that match the host profile and then execute them as needed. It does add a bit of "magic", but might make the individual test plugins easier to write and maintain.

ashcrow:
aos-ci-test

roles/openshift_health_checker/openshift_checks/mixins.py
ashcrow: I'd suggest keeping mixins over metaclasses. While metaclasses are great for exposing apis in a easy to use form I feel like adding them where they are not extremely needed would be too magical and would likely cause confusion for those modifying the classes in the future.

ashcrow:
I believe this is ready to go. There is discussion on how to make ```openshift_checks``` even better, but I believe that work should be done in a different PR.

roles/openshift_health_checker/openshift_checks/mixins.py
rhcarvalho: @juanvallejo ^^^ FYI this conversation might be interesting for you, since you'll be writing new checks :)

rhcarvalho:
@ashcrow how is 15bd3c3d24e7f00273767992df2cbf364216b11e related to this PR?

I believe 116c79cf887e0fd93d92a2b07cf92f1c7b5fde27 could be squashed to the commit modifying the \*/src/\* files.

Other than that LGTM :+1:

ashcrow:
Squashed.

ashcrow:
aos-ci-test

PR #3430, opened 2017-02-20
bug 1417261. Quote name and secrets in logging templates

jcantrill:
I am unable to reproduce but this seems to be a reasonable change.


ewolinetz:
I'm not convinced that this would resolve that bz. I'm not able to reproduce it either and it looks similar to this which has been marked as resolved https://bugzilla.redhat.com/show_bug.cgi?id=1419838

jcantrill:
I agree, I'm not convinced either as I am unable to reproduce.  This changes make the template consistent with what we have done in other templates.

PR #3432, opened 2017-02-21
Update openshift_logging role to detect if journald log driver is use…

ewolinetz:
…d and set fluentd setting appropriately if not specified

@richm @sdodson 

This evaluates the docker options openshift_facts picked up and tells fluentd to use journal based on whether or not we are using the journald log driver

### DIFF
--- a/roles/openshift_logging/tasks/install_fluentd.yaml
+++ b/roles/openshift_logging/tasks/install_fluentd.yaml
@@ -5,6 +5,10 @@
 - set_fact: fluentd_ops_port={{ (openshift_logging_use_ops | bool) | ternary(openshift_logging_es_ops_port, openshift_logging_es_port) }}
   check_mode: no
 
+- set_fact: openshift_logging_fluentd_use_journal={{ "journald" in openshift.docker.options }}
+  check_mode: no
+  when: "openshift_logging_fluentd_use_journal == ''"
+
 - name: Generating Fluentd daemonset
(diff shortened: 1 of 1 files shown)


roles/openshift_logging/tasks/install_fluentd.yaml
```diff
@@ -5,6 +5,10 @@
 - set_fact: fluentd_ops_port={{ (openshift_logging_use_ops | bool) | ternary(openshift_logging_es_ops_port, openshift_logging_es_port) }}
   check_mode: no
 
+- set_fact: openshift_logging_fluentd_use_journal={{ "journald" in openshift.docker.options }}
+  check_mode: no
+  when: "openshift_logging_fluentd_use_journal == ''"
```
richm: I would prefer to leave fluentd_use_journal unset, or if it must be set, set to `''`, and just let the fluentd pod internally figure out if journal is used or not.

ewolinetz: Is it worth having it as a configurable option then?
This is to satisfy  https://bugzilla.redhat.com/show_bug.cgi?id=1401417
Effectively we're doing the same logic as would be done within the Fluentd pod, correct?

richm:
In fact I regret even having the option to set USE_JOURNAL.  IMO it should never be used.  It should always be `''`, or not present, and have fluentd figure it out.

ewolinetz:
Per IRC discussion with @richm closing this PR due to the fact that the nature of how/what/from where Fluentd is collecting logs may change in the near future and the purpose of this ENV var will likely change due to it

PR #3434, opened 2017-02-21
System containers module

giuseppe:


giuseppe:
aos-ci-test

giuseppe:
rebased ⬆️ to fix the conflicts with master

aos-ci-test

roles/lib_openshift/library/oc_atomic_container.py
```diff
@@ -0,0 +1,187 @@
...
+'''
+
+# -*- -*- -*- End included fragment: doc/atomic_container -*- -*- -*-
+
+# -*- -*- -*- Begin included fragment: ansible/oc_atomic_container.py -*- -*- -*-
+
+# pylint: disable=wrong-import-position,too-many-branches,invalid-name
+import json
+from ansible.module_utils.basic import AnsibleModule
+
```
ashcrow: nit: I feel like this will fail (or should fail) ``flake8`` and/or ``pep8`` checks. There should be 2 newlines between functions and classes. Same with the rest of the single newline separators in the file.

roles/lib_openshift/library/oc_atomic_container.py
```diff
@@ -0,0 +1,187 @@
...
+def _install(module, container, image, values_list):
+    args = ['atomic', 'install', "--system", '--name=%s' % container] + values_list + [image]
```
ashcrow: nit: docstrings! :smile:

roles/lib_openshift/library/oc_atomic_container.py
```diff
@@ -0,0 +1,187 @@
...
+    rc, out, err = module.run_command(args, check_rc=False)
+    if rc != 0:
+        return rc, out, err, False
+    else:
+        changed = "Extracting" in out
+        return rc, out, err, changed
```
ashcrow: Double check that out and error are proper per Python implementation (IE: ``str`` on 2 and ``bytes`` on 3). If they are not you may want to use ``decode`` before returning.

giuseppe: I've verified that module.run_command returns 'str' on both implementations

roles/lib_openshift/library/oc_atomic_container.py
giuseppe: Thanks.  Going to fix it

roles/lib_openshift/library/oc_atomic_container.py
giuseppe: going to add them :)

giuseppe:
aos-ci-test

giuseppe:
I pushed another version which fixes an issue when installing with more than one master

aos-ci-test

roles/lib_openshift/library/oc_atomic_container.py
```diff
@@ -0,0 +1,196 @@
...
+
+def _uninstall(module, name):
+    ''' uninstall an atomic container by its name. '''
+    args = ['atomic', 'uninstall', name]
+    rc, out, err = module.run_command(args, check_rc=False)
+    return rc, out, err, False
+
```
ashcrow: nit: There are still single newlines between functions. I think ``pylint``/``flake8`` may not be catching this due to disabling errors. However, I recommend following [pep8](https://www.python.org/dev/peps/pep-0008/) as much as possible. I won't block on it though since it's been disabled.

giuseppe:
@detiber does it look fine for you?  I added the module here to not get stuck waiting for an `atomic_container` module in [Ansible](https://github.com/ansible/ansible/pull/21366).  Once the module will be included into Ansible, we can drop this one and use directly the Ansible version.  Hopefully the playbooks will need only minimal changes.

giuseppe:
aos-ci-test

ashcrow:
I think something's up with the tests. The comments suggest that unit tests passed yet the bot said they failed. The comment says containerized e2e failed but the bot said they passed .. .but looking at the log I can't find where it failed.

giuseppe:
@ashcrow thanks!  I am still unable to merge this PR though, I will run the tests again, let's see if it changes something

aos-ci-test

giuseppe:
aos-ci-test

giuseppe:
@ashcrow finally all tests pass, is is fine to merge?

ashcrow:
@giuseppe still looks good! :smile:

PR #3435, opened 2017-02-21
WIP: introduce master upgrade role

ingvagabund:
DO NOT MERGE

### DIFF
--- a/roles/openshift_master_upgrade/handlers/main.yml
+++ b/roles/openshift_master_upgrade/handlers/main.yml
@@ -1,0 +1,35 @@
+---
+- name: restart master
+  systemd: name={{ openshift.common.service_type }}-master state=restarted
+  when: (openshift.master.ha is not defined or not openshift.master.ha | bool) and (not (master_service_status_changed | default(false) | bool))
+  notify: Verify API Server
+
+- name: restart master api
+  systemd: name={{ openshift.common.service_type }}-master-api state=restarted
+  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  notify: Verify API Server
+
+- name: restart master controllers
+  systemd: name={{ openshift.common.service_type }}-master-controllers state=restarted
+  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+
+- name: Verify API Server
+  # Using curl here since the uri module requires python-httplib2 and
+  # wait_for port doesn't provide health information.
+  command: >
+    curl --silent --tlsv1.2
+    {% if openshift.common.version_gte_3_2_or_1_2 | bool %}
+    --cacert {{ openshift.common.config_base }}/master/ca-bundle.crt
+    {% else %}
+    --cacert {{ openshift.common.config_base }}/master/ca.crt
+    {% endif %}
+    {{ openshift.master.api_url }}/healthz/ready
+  args:
+    # Disables the following warning:
+    # Consider using get_url or uri module rather than running curl
+    warn: no
+  register: api_available_output
+  until: api_available_output.stdout == 'ok'
+  retries: 120
+  delay: 1
+  changed_when: false
--- a/roles/openshift_master_upgrade/meta/main.yml
+++ b/roles/openshift_master_upgrade/meta/main.yml
@@ -1,0 +1,13 @@
+---
+galaxy_info:
+  author: your name
+  description: OpenShift Master upgrade
+  company: Red Hat, Inc.
+  license: Apache License, Version 2.0
+  min_ansible_version: 2.1
+  platforms:
+  - name: EL
+    versions:
+    - 7
+dependencies:
+- role: openshift_common
--- a/roles/openshift_master_upgrade/tasks/main.yml
+++ b/roles/openshift_master_upgrade/tasks/main.yml
@@ -1,0 +1,49 @@
+---
+# input variables:
+# - openshift_master_upgrade_pre_hook
+# - master_config_hook
+# - openshift.common.config_base
+# - openshift_master_upgrade_hook
+# - openshift.common.rolling_restart_mode
+# - openshift_master_upgrade_post_hook
+
+# Run the pre-upgrade hook if defined:
+- debug: msg="Running master pre-upgrade hook {{ openshift_master_upgrade_pre_hook }}"
+  when: openshift_master_upgrade_pre_hook is defined
+
+- include: "{{ openshift_master_upgrade_pre_hook }}"
+  when: openshift_master_upgrade_pre_hook is defined
+
+- include: rpm_upgrade.yml component=master
+  when: not openshift.common.is_containerized | bool
+
+- include: upgrade_scheduler.yml
+
+- include: "{{ master_config_hook }}"
+  when: master_config_hook is defined
+
+- name: Update systemd units
+  include: systemd_units.yml
+
+- name: Bundle certificate
+  include: bundle-cert.yml
+
+# Run the upgrade hook prior to restarting services/system if defined:
+- debug: msg="Running master upgrade hook {{ openshift_master_upgrade_hook }}"
+  when: openshift_master_upgrade_hook is defined
+
+- include: "{{ openshift_master_upgrade_hook }}"
+  when: openshift_master_upgrade_hook is defined
+
+- include: restart_hosts.yml
+  when: openshift.common.rolling_restart_mode == 'system'
+
+- include: restart_services.yml
+  when: openshift.common.rolling_restart_mode == 'services'
+
+# Run the post-upgrade hook if defined:
+- debug: msg="Running master post-upgrade hook {{ openshift_master_upgrade_post_hook }}"
+  when: openshift_master_upgrade_post_hook is defined
+
+- include: "{{ openshift_master_upgrade_post_hook }}"
+  when: openshift_master_upgrade_post_hook is defined
--- a/roles/openshift_master_upgrade/tasks/restart_hosts.yml
+++ b/roles/openshift_master_upgrade/tasks/restart_hosts.yml
@@ -1,0 +1,43 @@
+---
+# input variables:
+# - openshift.common.hostname
+# - openshift.master.api_port
+
+- name: Restart master system
+  # https://github.com/ansible/ansible/issues/10616
+  shell: sleep 2 && shutdown -r now "OpenShift Ansible master rolling restart"
+  async: 1
+  poll: 0
+  ignore_errors: true
+  become: yes
+
+# WARNING: This process is riddled with weird behavior.
+
+# Workaround for https://github.com/ansible/ansible/issues/21269
+- set_fact:
+    wait_for_host: "{{ ansible_host }}"
+
+# Ansible's blog documents this *without* the port, which appears to now
+# just wait until the timeout value and then proceed without checking anything.
+# port is now required.
+#
+# However neither ansible_ssh_port or ansible_port are reliably defined, likely
+# only if overridden. Assume a default of 22.
+- name: Wait for master to restart
+  local_action:
+    module: wait_for
+      host="{{ wait_for_host }}"
+      state=started
+      delay=10
+      timeout=600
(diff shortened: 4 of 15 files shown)


roles/openshift_master_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+# - openshift_master_upgrade_post_hook
+
+# Run the pre-upgrade hook if defined:
+- debug: msg="Running master pre-upgrade hook {{ openshift_master_upgrade_pre_hook }}"
+  when: openshift_master_upgrade_pre_hook is defined
+
+- include: "{{ openshift_master_upgrade_pre_hook }}"
+  when: openshift_master_upgrade_pre_hook is defined
+
+- include: rpm_upgrade.yml component=master
```
detiber: Instead of doing in an include to update a single package here, it might make sense to either push this task file into a common parent role, so that it can be included from other upgrade roles as well.

The other thing that I'm wondering about is if we can consolidate this task with the similar tasks needed for updating docker containers (and eventually system containers). If the task file to be included is a variable, we can take advantage of dynamic includes and can avoid the performance hit of the skipped tasks for the deployment scenarios that do not apply. http://docs.ansible.com/ansible/playbooks_roles.html#dynamic-versus-static-includes

roles/openshift_master_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+  when: openshift_master_upgrade_hook is defined
+
+- include: "{{ openshift_master_upgrade_hook }}"
+  when: openshift_master_upgrade_hook is defined
+
+- include: restart_hosts.yml
+  when: openshift.common.rolling_restart_mode == 'system'
+
+- include: restart_services.yml
+  when: openshift.common.rolling_restart_mode == 'services'
```
detiber: See my comment above about dynamic includes, these two include tasks would be a good candidate.

roles/openshift_master_upgrade/tasks/restart_hosts.yml
```diff
@@ -0,0 +1,43 @@
...
+# only if overridden. Assume a default of 22.
+- name: Wait for master to restart
+  local_action:
+    module: wait_for
+      host="{{ wait_for_host }}"
+      state=started
+      delay=10
+      timeout=600
+      port="{{ ansible_port | default(ansible_ssh_port | default(22,boolean=True),boolean=True) }}"
+  become: no
```
detiber: Considering how tricky the above steps have proven, I think it would be a good idea to avoid duplicating it. It might be a good idea to try and pull this in from a common shared dependency role.

roles/openshift_master_upgrade/tasks/restart_services.yml
```diff
@@ -0,0 +1,31 @@
...
+  service:
+    name: "{{ openshift.common.service_type }}-master-api"
+    state: restarted
+  when: openshift_master_ha | bool
+- name: Wait for master API to come back online
+  wait_for:
+    host: "{{ openshift.common.hostname }}"
+    state: started
+    delay: 10
+    port: "{{ openshift.master.api_port }}"
```
detiber: I'm wondering if a do/until loop with the uri module testing for 'healthz/ready' would be better than wait_for here.

roles/openshift_master_upgrade/tasks/restart_services.yml
```diff
@@ -0,0 +1,31 @@
...
+  when: openshift_master_ha | bool
+- name: Restart master controllers
+  service:
+    name: "{{ openshift.common.service_type }}-master-controllers"
+    state: restarted
+  # Ignore errrors since it is possible that type != simple for
+  # pre-3.1.1 installations.
+  ignore_errors: true
```
detiber: At this point, I think we can kill off the 3.1.1 support, since we have many other issues that prevent us from backporting changes that far.

roles/openshift_master_upgrade/tasks/restart_services.yml
```diff
@@ -0,0 +1,31 @@
...
+  when: openshift_master_ha | bool
```
detiber: Using a block to wrap the tasks with this same condition would be good.

roles/openshift_master_upgrade/tasks/systemd_units.yml
```diff
@@ -0,0 +1,182 @@
...
+# - restart node
+
+# This file is included both in the openshift_master role and in the upgrade
+# playbooks.  For that reason the ha_svc variables are use set_fact instead of
+# the vars directory on the role.
+
+- name: Init HA Service Info
+  set_fact:
+    containerized_svc_dir: "/usr/lib/systemd/system"
+    ha_svc_template_path: "native-cluster"
```
detiber: Any reason to not just define the vars in a common parent directory for both roles? There are some times when set_fact is needed, but it does incur a cost that can be avoided if defined in another way.

Alternatively, these variables could be set at the block level as well.

roles/openshift_master_upgrade/tasks/systemd_units.yml
```diff
@@ -0,0 +1,182 @@
...
+  - openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
+  - not openshift.common.is_master_system_container | bool
+  with_items:
+  - api
+  - controllers
+  register: create_ha_unit_files
+
+- command: systemctl daemon-reload
+  when: create_ha_unit_files | changed
+# end workaround for missing systemd unit files
```
detiber: Can use the systemd module instead of command here.

roles/openshift_master_upgrade/tasks/systemd_units.yml
```diff
@@ -0,0 +1,182 @@
...
+  template:
+    src: "master_docker/master.docker.service.j2"
+    dest: "{{ containerized_svc_dir }}/{{ openshift.common.service_type }}-master.service"
+  when:
+  - openshift.common.is_containerized | bool and (openshift.master.ha is not defined or not openshift.master.ha) | bool
+  - not openshift.common.is_master_system_container | bool
+  register: create_master_unit_file
+
+- command: systemctl daemon-reload
+  when: create_master_unit_file | changed
```
detiber: should use the systemd module here

roles/openshift_master_upgrade/tasks/systemd_units.yml
```diff
@@ -0,0 +1,182 @@
...
+  when: openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
+  failed_when: false
+  changed_when: false
+
+- name: Preserve Master API AWS options
+  command: grep AWS_ /etc/sysconfig/{{ openshift.common.service_type }}-master-api
+  register: master_api_aws
+  when: openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
+  failed_when: false
+  changed_when: false
```
detiber: Instead of command/grep, could use slurp here

roles/openshift_master_upgrade/meta/main.yml
```diff
@@ -0,0 +1,13 @@
...
+  description: OpenShift Master upgrade
+  company: Red Hat, Inc.
+  license: Apache License, Version 2.0
+  min_ansible_version: 2.1
+  platforms:
+  - name: EL
+    versions:
+    - 7
+dependencies:
+- role: openshift_common
```
ingvagabund: I think both `openshift_version` and `openshift_master_facts` are run separately and only once. Including them into the roles would cause running both roles every time the `openshift_master_upgrade` role is run inside a play.

roles/openshift_master_upgrade/tasks/systemd_units.yml
```diff
@@ -0,0 +1,182 @@
...
+- name: Restore Master API AWS Options
+  when: openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
+      and master_api_aws.rc == 0 and
+      not (openshift_cloudprovider_kind is defined and openshift_cloudprovider_kind == 'aws' and openshift_cloudprovider_aws_access_key is defined and openshift_cloudprovider_aws_secret_key is defined)
+  lineinfile:
+    dest: /etc/sysconfig/{{ openshift.common.service_type }}-master-api
+    line: "{{ item }}"
+  with_items: "{{ master_api_aws.stdout_lines | default([]) }}"
+
+- name: Preserve Master Controllers Proxy Config options
```
detiber: It looks like this preserve, update, notify, restore is done multiple times over different files. Seems like it would be good to abstract it away in a task file, and use an loop with include to do the tasks for the needed items.

Alternatively, the items could be updated to operate on a list, with the list being built up based on the set variables.

roles/openshift_master_upgrade/vars/main.yml
```diff
@@ -0,0 +1,27 @@
+---
+openshift_master_config_dir: "{{ openshift.common.config_base }}/master"
+openshift_master_config_file: "{{ openshift_master_config_dir }}/master-config.yaml"
+openshift_master_scheduler_conf: "{{ openshift_master_config_dir }}/scheduler.json"
```
detiber: It seems like we should have an common role to set these variables, since both roles already rely on `openshift_master_facts`, maybe that is a good place for it.

roles/openshift_master_upgrade/meta/main.yml
detiber: So, within the context of play, you should not see a role listed as a dependency being run more than once, unless it is being called explicitly with different variables. There is a bug in Ansible that we would need to track down if that is the case.

If we are going to add requirements for implicit dependencies rather than explicit, then the role docs should make that clear. In this case, calling the needed steps out in the Requirements section rather than just relying on the example would probably be good.

roles/openshift_master_upgrade/handlers/main.yml
```diff
@@ -0,0 +1,35 @@
+---
```
ingvagabund: How much crazy is to create another ``openshift_master_restart`` role that would provide the handlers. The ``openshift_master`` and ``openshift_master_upgrade`` would then depend on it.

roles/openshift_master_upgrade/meta/main.yml
ingvagabund: That is ok, that works as expected. Just, if you have two plays each one including the same role, dependency of that role gets run twice, once for each play. I need to check the individual ``openshift*_facts`` roles to see how they work.

roles/openshift_master_upgrade/handlers/main.yml
detiber: It really seems like we need an openshift_master_common that can hold things that are not openshift_master_facts?

roles/openshift_master_upgrade/meta/main.yml
detiber: I'm not sure we can rely on variables to be set by a role from a different play.

roles/openshift_master_upgrade/handlers/main.yml
ingvagabund: Maybe too wild but: https://paste.fedoraproject.org/paste/yYpTUUuwZl~sB7bORsieUl5M1UNdIGYhyRLivL9gydE=

The point:
1) Every handler is actually a task that could be needed at any point of any role. So make it includable via ``include_role: name=ROLE file=FILE``
2) in handlers, actually include handler's body as a task from the same role (thus the symlink)
3) roles that need ROLEH's handlers, create new handlers that run tasks from ROLEH's handlers

Benefits:
* no more handlers duplication
* you know precisely from where a handler of a different role comes from
* role does not include files out of the role -> the role can be still shared on ansible-galaxy

Drawbacks:
* symlinks from handlers to tasks
* duplication of handlers on a level of defining handlers and include[_role] statements
* you can not specify ROLEH as a dependency of your role

ingvagabund: So that does not work either. Still hitting https://github.com/ansible/ansible/issues/20493

detiber: If we didn't have the role dependencies to worry about, I would say that we could just add a tasks/noop.yml file to the role, then you could use
```
- include_role:
    tasks_from: noop
```

If we had a shared common role that housed handlers, variables, default values, etc but didn't have any role dependencies, then we could pull that in instead.

That was kind of the goal of roles like openshift_master_facts, but it has the nasty dependency on openshift_facts currently.

michaelgugino:
Closing due to inactivity.

PR #3436, opened 2017-02-21
Implement fake openssl cert classes

tbielawa:
Implements a set of fake classes to mimic the object returned from the actual `OpenSSL.crypto.load_certificate` function.


tbielawa:
Sweeping up flakes now

roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py
```diff
@@ -0,0 +1,91 @@
...
+import sys
+import unittest
+
+# Removing invalid variable names for tests so that I can
+# keep them brief
+# pylint: disable=invalid-name,no-name-in-module
+# Disable import-error b/c our libraries aren't loaded in jenkins
+# pylint: disable=import-error,wrong-import-position
+# place class in our python path
+module_path = os.path.realpath('./library/')
```
detiber: I think you want this instead:
```
module_path = os.path.join('/'.join(os.path.realpath(__file__).split('/')[:-1]), 'library')
```

I'd also remove the executable bits and just let pytest run it.

ashcrow: I'd also switch the '/' to ``os.path.sep``.

tbielawa: Tried this. Still failed. We talked on IRC and decided to use the `pytest` skip functionality for now. The test is still part of the source tree, but it is not ran at this time.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -66,6 +76,124 @@
...
+
+            elif l.startswith('Not After :'):
+                # Not After : Feb  7 18:19:35 2019 GMT
+                # => strptime(str, '%b %d %H:%M:%S %Y %Z')
+                # => strftime('%Y%m%d%H%M%SZ')
+                # => 20190207181935Z
+                not_after_raw = l.partition(' : ')[-1]
+                # Last item: ('Not After', ' : ', 'Feb  7 18:19:35 2019 GMT')
+                not_after_parsed = datetime.datetime.strptime(not_after_raw, '%b %d %H:%M:%S %Y %Z')
+                self.not_after = not_after_parsed.strftime('%Y%m%d%H%M%SZ')
```
ashcrow: Double checking you want ``Z`` and not ``%Z`` in this line.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -66,6 +76,124 @@
...
+                PARSING_ALT_NAMES = True
+                continue
+
+            elif l.startswith('Subject:'):
+                # O=system:nodes, CN=system:node:m01.example.com
+                self.subject = FakeOpenSSLCertificateSubjects(l.partition(': ')[-1])
+
+    def get_serial_number(self):
+        """Return the serial number of the cert"""
+        return self.serial
```
ashcrow: This should be set to ``None`` (or another default) during initialization to avoid missing members.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -66,6 +76,124 @@
...
+
+    def get_subject(self):
+        """Subjects must implement get_components() and return dicts or
+tuples. An 'openssl x509 -in CERT.cert -text' with 'Subject':
+
+    Subject: Subject: O=system:nodes, CN=system:node:m01.example.com
+
+might return: [('O=system', 'nodes'), ('CN=system', 'node:m01.example.com')]
+        """
+        return self.subject
```
ashcrow: This should be set to ``None`` (or another default) during initialization to avoid missing members.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -66,6 +76,124 @@
...
+
+    def get_extension(self, i):
+        """Extensions must implement get_short_name() and return the string
+'subjectAltName'"""
+        return self.extensions[i]
```
ashcrow: This should have a default of ``[]`` during initialization.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -66,6 +76,124 @@
...
+
+    def get_notAfter(self):
+        """Returns a date stamp as a string in the form
+'20180922170439Z'. strptime the result with format param:
+'%Y%m%d%H%M%SZ'."""
+        return self.not_after
```
ashcrow: This should be set to ``None`` (or another default) during initialization to avoid missing members.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -119,10 +247,28 @@ def load_and_handle_cert(cert_string, now, base64decode=False):
...
+    if 'OpenSSL.crypto' in sys.modules:
+        # No work-around required
+        cert_loaded = OpenSSL.crypto.load_certificate(
+            OpenSSL.crypto.FILETYPE_PEM, _cert_string)
+    else:
+        # Missing library, work-around required. We need to write the
+        # cert out to disk temporarily so we can run the 'openssl'
+        # command on it to decode it
+        _, path = tempfile.mkstemp()
+        with open(path, 'w') as fp:
```
ashcrow: nit: Switch to ``io.open`` for a better Python 2/3 interface.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -119,10 +247,28 @@ def load_and_handle_cert(cert_string, now, base64decode=False):
...
+            fp.write(_cert_string)
+            fp.flush()
+        cmd = 'openssl x509 -in {} -text'.format(path)
+        openssl_decoded = subprocess.Popen(cmd.split(),
```
ashcrow: I feel like there should be a ``try``/``except`` here to catch and notify if the command fails or can not be found.

roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+'''
+ Unit tests for the FakeOpenSSL classes
+'''
+# Run from roles/openshift_certificate_expiry/
```
ashcrow: nit: no longer accurate with the new use of ``pytest``.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
tbielawa: Good thing to point out. I double checked, that's correct. Later down in the code you can see the same format string used in a str**p**time call.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
tbielawa: Oh. Neat. Had to slip a `unicode()` into the `fp.write()` to get it working. No problem.

tbielawa:
Rebased. Squashed. All feedback addressed. Ready for final reviews (I hope) @ashcrow and @detiber

tbielawa:
aos-ci-test

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -100,7 +241,8 @@ def filter_paths(path_list):
     return [p for p in path_list if os.path.exists(os.path.realpath(p))]
 
 
-def load_and_handle_cert(cert_string, now, base64decode=False):
+# pylint: disable=too-many-locals,too-many-branches
```
ashcrow: I won't block on this but generally needing to disable these is a sign of the function needing to be broken into multiple functions.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -119,10 +262,33 @@ def load_and_handle_cert(cert_string, now, base64decode=False):
...
+    if 'OpenSSL.crypto' in sys.modules:
+        # No work-around required
+        cert_loaded = OpenSSL.crypto.load_certificate(
+            OpenSSL.crypto.FILETYPE_PEM, _cert_string)
+    else:
+        # Missing library, work-around required. We need to write the
+        # cert out to disk temporarily so we can run the 'openssl'
+        # command on it to decode it
+        _, path = tempfile.mkstemp()
+        with io.open(path, 'w') as fp:
```
ashcrow: :+1:

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -119,10 +262,33 @@ def load_and_handle_cert(cert_string, now, base64decode=False):
...
+            fp.write(six.u(_cert_string))
+            fp.flush()
+
+        cmd = 'openssl x509 -in {} -text'.format(path)
+        try:
+            openssl_decoded = subprocess.Popen(cmd.split(),
+                                               stdout=subprocess.PIPE)
+        except OSError:
+            ans_module.fail_json(msg="Error: The 'OpenSSL' python library and CLI command were not found on the target host. Unable to parse any certificates. This host will not be included in generated reports.")
```
ashcrow: :+1:

roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py
```diff
@@ -0,0 +1,86 @@
...
+
+import os
+import sys
+import unittest
+import pytest
+
+# Disable import-error b/c our libraries aren't loaded in jenkins
+# pylint: disable=import-error,wrong-import-position
+# place class in our python path
+module_path = os.path.join('/'.join(os.path.realpath(__file__).split('/')[:-1]), 'library')
```
ashcrow: Won't block on it but when you find the need to use file system separators ``os.path.sep`` is generally a good idea.

roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py
```diff
@@ -0,0 +1,86 @@
...
+        """We can read the Subject from a cert with sans"""
+        # Subject: O=system:nodes, CN=system:node:m01.example.com
+        subject = self.fake_san_cert.get_subject()
+        subjects = []
+        for name, value in subject.get_components():
+            subjects.append('{}={}'.format(name, value))
+
+        self.assertEqual('CN=172.30.0.1', ', '.join(subjects))
+
+    def tearDown(self):
```
ashcrow: nit: If this isn't required I'd delete it.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
```diff
@@ -119,10 +262,33 @@ def load_and_handle_cert(cert_string, now, base64decode=False):
...
+        else:
+            openssl_decoded = openssl_decoded.communicate()[0]
+            os.remove(path)
```
ashcrow: One thing that just hit me ... if you write the file and the command doesn't work you still will want to delete the cert!

tbielawa:
aos-ci-test

tbielawa:
All review comments addressed.

roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
detiber: Any specific reason to use a tempfile here instead of passing the cert content to stdin for the subprocess?

PR #3438, opened 2017-02-21
Bz 1425321 default api port

jcantrill:
May or may not be the root cause of reported issues but this is a good idea to source from the facts.


jcantrill:
aos-ci-test

hosts.ose
```diff
@@ -0,0 +1,725 @@
+# This is an example of a bring your own (byo) host inventory
```
ewolinetz: slightly confused why this file is at the project root instead of under inventory/byo.
It might make sense to either remove this file or add to one of the current host examples.

jcantrill: cruft.  will remove

jcantrill:
@ewolinetz updated to remove file

jcantrill:
bug 1425312.  Incorrect in the history

PR #3445, opened 2017-02-21
CloudFront oc_secret contents should be a list

ashcrow:
Fixes "argument contents is of type <type 'dict'> and we were unable to
convert to list"

The error mislead me until @tbielawa pointed out it was the argument **named** contents and not the contents of an argument.


ashcrow:
aos-ci-test

PR #3447, opened 2017-02-22
Fix for no default_cert or cert, key, cacert files

kwoodson:


kwoodson:
@mtnbikenc, I put in a fix for no cert files or default_cert passed which was what your case was.  Please test with this fix.

roles/lib_openshift/library/oc_adm_router.py
```diff
@@ -2937,6 +2945,7 @@ def main():
         mutually_exclusive=[["router_type", "images"],
                             ["key_file", "default_cert"],
                             ["cert_file", "default_cert"],
+                            ["cacert_file", "default_cert"],
                            ],
```
detiber: It might be good to add required_together:
```
        required_together=[
            ['cacert_file', 'cert_file', 'key_file'],
        ],
```

It also looks like the cert arguments aren't covered in the docs.

mtnbikenc:
@kwoodson Starting to test.

roles/lib_openshift/library/oc_adm_router.py
kwoodson: @detiber, thanks, I'll add this.

PR #3448, opened 2017-02-22
Prepare for origin moving to OCP version scheme

smarterclayton:
Origin 1.6 and future releases will be versioned as Origin 3.6+. This
unifies a point of inconsistency and brings Origin in line with the full
release history. Update version numbers to match that 3.x is preferred.

@sdodson @tdawson these are the obvious changes I see.  I need to test
with an RPM repo created from this first.

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -875,7 +875,7 @@ def set_version_facts_if_unset(facts):
                 version_gte_3_3_or_1_3 = version >= LooseVersion('1.3.0')
                 version_gte_3_4_or_1_4 = version >= LooseVersion('1.4.0')
                 version_gte_3_5_or_1_5 = version >= LooseVersion('1.5.0')
-                version_gte_3_6_or_1_6 = version >= LooseVersion('1.6.0')
+                version_gte_3_6_or_1_6 = version >= LooseVersion('3.6.0') or version >= LooseVersion('1.6.0')
             else:
                 version_gte_3_1_or_1_1 = version >= LooseVersion('3.0.2.905')
                 version_gte_3_1_1_or_1_1_1 = version >= LooseVersion('3.1.1')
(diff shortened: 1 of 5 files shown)


roles/openshift_facts/library/openshift_facts.py
```diff
@@ -875,7 +875,7 @@ def set_version_facts_if_unset(facts):
                 version_gte_3_3_or_1_3 = version >= LooseVersion('1.3.0')
                 version_gte_3_4_or_1_4 = version >= LooseVersion('1.4.0')
                 version_gte_3_5_or_1_5 = version >= LooseVersion('1.5.0')
-                version_gte_3_6_or_1_6 = version >= LooseVersion('1.6.0')
+                version_gte_3_6_or_1_6 = version >= LooseVersion('3.6.0') || version >= LooseVersion('1.6.0')
```
detiber: With the unified versioning, we don't even really need the version_gte_* variables. As long as we can access the version, we can move the comparisons to where they are used with the `version_compare` filter: https://docs.ansible.com/ansible/playbooks_tests.html#version-comparison

DanyC97:
nice one @smarterclayton !! i'm still new here, roughly what time frame are we looking for 1.6/3.6 to be out? (sorry for not being up to date on the release cycle)

tbielawa:
New versioning scheme will be great. It should make this part of the README a little more intuitive in the future

https://github.com/openshift/openshift-ansible#getting-the-correct-version

roles/openshift_facts/library/openshift_facts.py
smarterclayton: What should I do then?

smarterclayton:
Ping, what else is necessary here?  This blocks tagging origin.

roles/openshift_facts/library/openshift_facts.py
detiber: I don't think anything is needed for this PR, but something we'll want to address down the line.

detiber:
aos-ci-test

PR #3452, opened 2017-02-22
Bump ansible requirement to ansible-2.2.1.0-2

sdodson:


sdodson:
For community members, this version should be available in the centos paas sig common repo at http://mirror.centos.org/centos/7/paas/x86_64/openshift-origin/common/

PR #3457, opened 2017-02-22
BZ1414276 - Quote ansible_ssh_user when determining group id

sdodson:
ansible_ssh_user

### DIFF
--- a/playbooks/common/openshift-cluster/redeploy-certificates/ca.yml
+++ b/playbooks/common/openshift-cluster/redeploy-certificates/ca.yml
@@ -294,7 +294,7 @@
       client_path: "{{ openshift.common.config_base }}/master/admin.kubeconfig"
       ca_path: "{{ openshift.common.config_base }}/master/ca-bundle.crt"
   - name: Lookup default group for ansible_ssh_user
-    command: "/usr/bin/id -g {{ ansible_ssh_user }}"
+    command: "/usr/bin/id -g {{ ansible_ssh_user | quote }}"
     changed_when: false
     register: _ansible_ssh_user_gid
   - set_fact:
--- a/roles/openshift_master_certificates/tasks/main.yml
+++ b/roles/openshift_master_certificates/tasks/main.yml
@@ -159,7 +159,7 @@
   become: no
 
 - name: Lookup default group for ansible_ssh_user
-  command: "/usr/bin/id -g {{ ansible_ssh_user }}"
+  command: "/usr/bin/id -g {{ ansible_ssh_user | quote }}"
   changed_when: false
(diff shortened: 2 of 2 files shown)


playbooks/common/openshift-cluster/redeploy-certificates/ca.yml
```diff
@@ -294,7 +294,7 @@
       client_path: "{{ openshift.common.config_base }}/master/admin.kubeconfig"
       ca_path: "{{ openshift.common.config_base }}/master/ca-bundle.crt"
   - name: Lookup default group for ansible_ssh_user
-    command: "/usr/bin/id -g {{ ansible_ssh_user }}"
+    command: "/usr/bin/id -g {{ ansible_ssh_user | quote }}"
```
detiber: It looks like you can use the `user` module instead of command here:
```
- name Lookup default group for ansible_ssh_user
  user:
    name: ansible_ssh_user
  register:l _ansible_ssh_user
```
And then access it with: `{{ l_ansible_ssh_user.group }}`

sdodson: ok, testing this. Do you know by chance if that pwd.getpwnam() which user module uses works with domain username format? ie: 'domain\user' The customer had confirmed that quoting the command worked so I was going with that. I can't come up with a good test case.
```
>>> import pwd
>>> pwd.getpwnam('dom\user')
KeyError: 'getpwnam(): name not found: dom\\user'
```

detiber: @scollier @cooktheryan Would either of you be able to provide an environment using AD auth to validate?

detiber: @sdodson I think it might not, it looks like `pwd` is tied to the passwd database...  I'm wondering if there is a better option, though

detiber: Could you use the `real_group_id` fact from: https://github.com/ansible/ansible/tree/devel/lib/ansible/module_utils#L546-L558, it should be available as `ansible_real_group_id`

detiber: I'm not sure if that would return the regular user or the sudo user, though

sdodson:
Thanks for the feedback!
@detiber shall we go ahead with this? I've tested the negative case (non dom\user users).

sdodson:
aos-ci-test

cooktheryan:
@detiber @sdodson was this the item @markllama helped you test or do you still need testing?

PR #3459, opened 2017-02-22
[1.4] Fix node draining for hosts with daemonsets

sdodson:


playbooks/byo/openshift-cluster/upgrades/docker/docker_upgrade.yml
```diff
@@ -31,7 +31,7 @@
 
   - name: Drain Node for Kubelet upgrade
     command: >
-      {{ openshift.common.client_binary }} adm manage-node {{ openshift.common.hostname | lower }} {{ openshift.common.evacuate_or_drain }} --force
+      {{ openshift.common.admin_binary }} drain {{ openshift.common.hostname }} --force --delete-local-data --ignore-daemonsets
```
detiber: @sdodson any reason to use `admin_binary` instead of `client_binary` and adding `adm`?

sdodson: No, I just did this because it brings the branch in line with master.

mtnbikenc: @detiber `admin_binary` is my fault.  IIRC, I was having issues with with `oc adm drain` performing correctly at the time.  But since they are just linked to the `openshift` binary it was not making sense.

PR #3460, opened 2017-02-22
Handle router preparation errors. Also added a second sleep timer for creating openshift objects.

kwoodson:


roles/lib_openshift/library/oc_adm_router.py
```diff
@@ -2665,7 +2668,7 @@ def _prepare_router(self):
         results = self.openshift_cmd(cmd, oadm=True, output=True, output_type='json')
 
         # pylint: disable=no-member
```
detiber: Is this pylint disable still needed? If so, would disabling `maybe-no-member` work instead?

roles/lib_openshift/library/oc_adm_router.py
```diff
@@ -2705,9 +2708,11 @@ def create(self):
         '''Create a deploymentconfig '''
         results = []
 
+        import time
         # pylint: disable=no-member
```
detiber: same comment about the pylint disable

sdodson:
aos-ci-test

PR #3467, opened 2017-02-23
Fixed default variables.  Added a fix to generated secret in env var.

kwoodson:


roles/lib_openshift/src/ansible/oc_adm_registry.py
```diff
@@ -24,8 +24,8 @@ def main():
             service_account=dict(default='registry', type='str'),
             mount_host=dict(default=None, type='str'),
             volume_mounts=dict(default=None, type='list'),
-            env_vars=dict(default=None, type='dict'),
-            edits=dict(default=None, type='list'),
+            env_vars=dict(default={}, type='dict'),
+            edits=dict(default=[], type='list'),
```
detiber: Do we know if it is safe to set these defaults to `{}` and `[]`?

@ashcrow

ashcrow: In terms of Python 2 vs 3 it should be fine.

detiber: I wasn't sure if we might hit issues similar to using '[]' or '{}' as a method param default.

ashcrow: @detiber I think it's OK because it should be redefined every time ``main`` is called. If it was at a higher scope I think we'd hit that problem.

roles/lib_openshift/src/lib/deploymentconfig.py
```diff
@@ -105,6 +105,18 @@ def exists_env_key(self, key):
 
         return False
 
+    def get_env_var(self, key):
+        '''return a environment variables '''
+        results = self.get(DeploymentConfig.env_path) or []
+        if not results:
+           return None
```
mtnbikenc: lint nit: needs indent fix

mtnbikenc:
@kwoodson would you mind rebasing this to master?

PR #3468, opened 2017-02-23
openshift_logging default to 2 replicas of primary shards

jcantrill:
Updated value based on recommendation from Peter Portante

### DIFF
--- a/roles/openshift_logging/templates/elasticsearch.yml.j2
+++ b/roles/openshift_logging/templates/elasticsearch.yml.j2
@@ -8,7 +8,7 @@ script:
 index:
   number_of_shards: 1
   number_of_replicas: 0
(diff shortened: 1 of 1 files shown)


ewolinetz:
I'd like to make sure I understand the intent of this, we want to update so the MAXIMUM number of shard replicas that ES will auto expand to is 2 instead of 3 correct? (I'm guessing there is perf data that supports this decision)

jcantrill:
@portante can you address https://github.com/openshift/openshift-ansible/pull/3468#issuecomment-282088678

portante:
@ewolinetz, see issue #3496.  3 way replication is usually sufficient.  If customers need more, for some reason, we can consider adding the upper bound as a parameter.

PR #3472, opened 2017-02-23
Hacky work-around for rhel-subscribe

tbielawa:
I think this needs some work to get into a better form.

This is what I had to do do make the `rhel_subscribe` stuff work again. It USED to work, but then somehow one day it began requiring running `openshift_facts` first but that fails because PyYAML and friends (`required_packages`) aren't installed yet. So the role that gives you packages can't run because the packages it enables you to install aren't present.

Yes, **it is** just like this:

chicken-or-egg-400x301

### DIFF
--- a/playbooks/byo/rhel_subscribe.yml
+++ b/playbooks/byo/rhel_subscribe.yml
@@ -1,7 +1,33 @@
 ---
-- include: ../common/openshift-cluster/std_include.yml
+- name: Create initial host groups for localhost
+  hosts: localhost
+  connection: local
+  become: no
+  gather_facts: no
   tags:
   - always
+  tasks:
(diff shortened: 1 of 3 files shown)


tbielawa:
@abutcher I believe you and I talked about this a few weeks ago. You mentioned something about the `std_include.yml` stuff before.

detiber:
@tbielawa it might be a good idea to merge in the subscription/repo roles from here: https://github.com/openshift/openshift-ansible-contrib/tree/master/roles

@cooktheryan can speak to the implementation if needed.

abutcher:
@tbielawa I think we'll want to undo this change https://github.com/openshift/openshift-ansible/pull/2671/files#diff-377065f5e66159003edb0fc50ed24ee4

PR #3475, opened 2017-02-23
Fails Installation If Docker Version is Too Old On Atomic Host

ashcrow:
If the docker version on a master or node is under 1.12 and the OS is an
Atomic Host variant the installation should fail.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1425583

### DIFF
--- a/roles/openshift_facts/tasks/main.yml
+++ b/roles/openshift_facts/tasks/main.yml
@@ -26,6 +26,25 @@
     msg: "openshift-ansible requires Python 2 for {{ ansible_distribution }}"
   when: ansible_distribution != 'Fedora' and ansible_python['version']['major'] != 2
 
+# Fail as early as possible if Atomic and old version of Docker
+- block:
+
+  # See https://access.redhat.com/articles/2317361
+  # and https://github.com/ansible/ansible/issues/15892
+  # NOTE: the "'s can not be removed at this level else the docker command will fail
+  # NOTE: When ansible >2.2.1.x is used this can be updated per
+  # https://github.com/openshift/openshift-ansible/pull/3475#discussion_r103525121
+  - name: Determine Atomic Host Docker Version
+    shell: 'CURLY="{"; docker version --format "$CURLY{json .Server.Version}}"'
+    register: l_atomic_docker_version
+
+  - assert:
+      msg: Installation on Atomic Host requires Docker 1.12 or later. Please upgrade and restart the Atomic Host.
+      that:
+      - l_atomic_docker_version.stdout | replace('"', '') | version_compare('1.12','>=')
+
+  when: l_is_atomic | bool
+
 - name: Ensure various deps are installed
   package: name={{ item }} state=present
   with_items: "{{ required_packages }}"


ashcrow:
One thing I'm not sure about is if I placed the reusable set of tasks in the right place. AFAICT it made sense to live in ``openshift_common`` but if it needs to move please let me know.

roles/openshift_common/tasks/atomic_docker_check.yml
```diff
@@ -0,0 +1,17 @@
+---
+# Fail early if Atomic and old version of Docker
+- block:
+
+  - name: Determine available Docker
+    shell: "rpm -q --queryformat '---\ncurr_version: %{VERSION}\navail_version: \n' docker"
```
tbielawa: What's up with the `avail_version: ` being empty there? Looks like that would return yaml like:

```yaml
---
curr_version: x.y.z
avail_version:
```

tbielawa:
> One thing I'm not sure about is if I placed the reusable set of tasks in the right place. AFAICT it made sense to live in openshift_common but if it needs to move please let me know.

I do not know either. But that *seems* like the correct place.

roles/openshift_common/tasks/atomic_docker_check.yml
ashcrow: I pulled this from the docker upgrade steps. I'll fix that up to be easier to read.

ashcrow:
Updated.

tbielawa:
I am not sure if it matters, but the commit you made in #3473 (d24eb2d3e0ae360f031ddf23b2b36dc01f2cf51a) found its way into your PR.

Other than that it looks good to me as long as tests do pass.

filter_plugins/oo_filters.py
```diff
@@ -18,8 +18,8 @@
 from distutils.util import strtobool
 from distutils.version import LooseVersion
 from operator import itemgetter
+from ansible.module_utils.six.moves.urllib.parse import urlparse
```
tbielawa: Not sure why this commit was included.

roles/openshift_common/tasks/atomic_docker_check.yml
```diff
@@ -0,0 +1,16 @@
+---
+# Fail early if Atomic and old version of Docker
+- block:
+
+  - name: Determine available Docker
+    shell: "rpm -q --queryformat '---\ncurr_version: %{VERSION}\n' docker"
```
tbielawa: LGTM

filter_plugins/oo_filters.py
ashcrow: That's required for it to work. It's actually part of another PR open right now. I'll remove it once the other PR merges.

ashcrow:
aos-ci-test

ashcrow:
Rebased on master and updated.

ashcrow:
aos-ci-test

playbooks/common/openshift-node/config.yml
```diff
@@ -7,6 +7,11 @@
   - set_fact:
       openshift_node_debug_level: "{{ t_oo_option_node_debug_level }}"
     when: openshift_node_debug_level is not defined and t_oo_option_node_debug_level != ""
+
+  - name: Fail early if Atomic and old version of Docker
+    include: ../../../roles/openshift_common/tasks/atomic_docker_check.yml
+    when: openshift.common.is_atomic | bool
```
detiber: To me, just the fact that we need to copy these checks between the master and node playbooks is an indication that we need to do a better job of solving this problem in general.

roles/openshift_common/tasks/atomic_docker_check.yml
```diff
@@ -0,0 +1,16 @@
...
+    register: l_atomic_docker_version_result
+
+  - set_fact:
+      l_atomic_docker_version: "{{ l_atomic_docker_version_result.stdout | from_yaml }}"
+
+  - fail:
+      msg: Installation on Atomic Host requires Docker 1.12 or later. Please upgrade and restart the Atomic Host.
+    when: l_atomic_docker_version.curr_version | version_compare('1.12','<')
+
+  when: openshift.common.is_atomic | bool
```
detiber: Should we be limiting this type of check only to AH installs? Wouldn't we also want to fail fast for other types of deployments if the version of docker installed (or available) does not meet the requirements as well?

roles/openshift_common/tasks/atomic_docker_check.yml
```diff
@@ -0,0 +1,16 @@
...
+  - name: Determine available Docker
+    shell: "rpm -q --queryformat '---\ncurr_version: %{VERSION}\n' docker"
+    register: l_atomic_docker_version_result
+
+  - set_fact:
+      l_atomic_docker_version: "{{ l_atomic_docker_version_result.stdout | from_yaml }}"
+
+  - fail:
+      msg: Installation on Atomic Host requires Docker 1.12 or later. Please upgrade and restart the Atomic Host.
+    when: l_atomic_docker_version.curr_version | version_compare('1.12','<')
```
detiber: Instead of using `set_fact` to provide an intermediate step, I would suggest adding moving the definition of  `l_atomic_docker_version` to the task_vars. I also think that the `assert` module might be more appropriate here, especially since we have versions of OpenShift that would not be able to work with newer versions of docker as well.

```
- assert:
    msg: blah, blah, something incompatible verisons found
    that:
    - l_current_docker_version | version_compare('1.12', '>=')
    - l_current_docker_version | version_compare('1.13', '<') # hypothetically
  vars:
    l_current_docker_version: "{{ (l_atomic_docker_version_result.stdout | from_yaml).curr_version }}"
```

That would also make it a bit easier to add assertions for non-atomic hosts where we would need to compare against a combination of installed and available packages.

Downside to this is that it makes it harder to give a pretty error message, but I think it would be better to handle those types of issues in a callback plugin rather than having a ton of `fail` statements.

playbooks/common/openshift-master/config.yml
```diff
@@ -5,6 +5,10 @@
     t_oo_option_master_debug_level: "{{ lookup('oo_option', 'openshift_master_debug_level') }}"
 
   pre_tasks:
+  - name: Fail early if Atomic and old version of Docker
+    include: ../../../roles/openshift_common/tasks/atomic_docker_check.yml
+    when: openshift.common.is_atomic | bool
```
detiber: I'm not sure if we can rely on openshift.common.is_atomic being set at this point, since that is set in openshift_facts.

roles/openshift_common/tasks/atomic_docker_check.yml
giuseppe: on Atomic Host there are currently two versions of docker: docker-current and docker-latest (which I think only incidentally are the same on Atomic Host 7.3.2) and I'd say they can still be different in future:

https://access.redhat.com/articles/2317361

I'd suggest to get directly the version from /usr/bin/docker (that you can see it is a shell script that calls the "real" docker executable) as:

docker version --format '{{json .Server}}'

and get the version from there

roles/openshift_common/tasks/atomic_docker_check.yml
ashcrow: That's a good question. The bug this is for is specifically for failing on AH since a reboot is required. Currently it looks like the installer will 

**File**: ./roles/docker/tasks/main.yml
```yaml
- name: Error out if Docker pre-installed but too old
  fail:
    msg: "Docker {{ curr_docker_version.stdout }} is installed, but >= 1.9.1 is required."
  when: not curr_docker_version | skipped and curr_docker_version.stdout != '' and curr_docker_version.stdout | version_compare('1.9.1', '<') and not docker_version is defined
```

I'm happy to bump that up to ``1.12`` in another PR.

playbooks/common/openshift-master/config.yml
```diff
@@ -5,6 +5,10 @@
     t_oo_option_master_debug_level: "{{ lookup('oo_option', 'openshift_master_debug_level') }}"
 
   pre_tasks:
+  - name: Fail early if Atomic and old version of Docker
+    include: ../../../roles/openshift_common/tasks/atomic_docker_check.yml
```
ashcrow: Is there a better place I can put the tasks without creating a new role?

roles/openshift_common/tasks/atomic_docker_check.yml
ashcrow: https://github.com/ansible/ansible/issues/15892 looks like it may make using braces a bit rough but I'll try.

detiber: @ashcrow I think you might be looking for this: https://docs.ansible.com/ansible/playbooks_advanced_syntax.html#id4

```
!unsafe 'unsafe value'

- name: Determine installed docker version
  command: !unsafe "docker version --format '{{ json.Server}}'"
```
You might be able to use the `{%raw}..{%endraw}` template feature as well.

detiber: Also, this should use `command` and not `shell`

ashcrow: Thanks for the suggestion on advanced syntax. Unfortunately it doesn't work for commands ... even when setting via ``vars``. The result ends up being parsed to ``{#json .Server#}``.

roles/openshift_common/tasks/atomic_docker_check.yml
```diff
@@ -0,0 +1,17 @@
+---
+# Fail early if Atomic and old version of Docker
+- block:
+
+  # See https://access.redhat.com/articles/2317361
+  # and https://github.com/ansible/ansible/issues/15892
+  # NOTE: the "'s can not be removed at this level else the docker command will fail
+  - name: Determine Atomic Host Docker Version
+    shell: 'CURLY="{"; docker version --format "$CURLY{json .Server.Version}}"'
```
ashcrow: Not crazy about this but couldn't find another way to get this done.

playbooks/common/openshift-master/config.yml
ashcrow: What about using ``roles/openshift_{master,node}/tasks/main.yml``? It looks like the variables are defined there.

roles/openshift_common/tasks/atomic_docker_check.yml
detiber: It does look like it only works for variables, and the raw/endraw syntax doesn't seem to be respected in an inline template.

I was able to get the expected result with the following:
```
- hosts: localhost
  become: no
  gather_facts: no
  connection: local
  vars:
  tasks:
  - command: docker version --format "{{ version_format }}"
    register: result
    vars:
      version_format: !unsafe '{{json .Server}}'
  - debug: var=result
```
output:
```
23:43:03 [jdetiber:~/git/github/openshift] % ansible-playbook test.yml
 [WARNING]: provided hosts list is empty, only localhost is available


TASK [command] *****************************************************************
changed: [localhost]

TASK [debug] *******************************************************************
ok: [localhost] => {
    "result": {
        "changed": true, 
        "cmd": [
            "docker", 
            "version", 
            "--format", 
            "{{json .Server}}"
        ], 
        "delta": "0:00:00.027524", 
        "end": "2017-02-24 23:43:04.034446", 
        "rc": 0, 
        "start": "2017-02-24 23:43:04.006922", 
        "stderr": "", 
        "stdout": "{\"Version\":\"1.12.6\",\"ApiVersion\":\"1.24\",\"GitCommit\":\"ae7d637/1.12.6\",\"GoVersion\":\"go1.7.4\",\"Os\":\"linux\",\"Arch\":\"amd64\",\"KernelVersion\":\"4.9.7-201.fc25.x86_64\",\"BuildTime\":\"Mon Jan 30 16:15:28 2017\",\"PkgVersion\":\"docker-common-1.12.6-6.gitae7d637.fc25.x86_64\"}", 
        "stdout_lines": [
            "{\"Version\":\"1.12.6\",\"ApiVersion\":\"1.24\",\"GitCommit\":\"ae7d637/1.12.6\",\"GoVersion\":\"go1.7.4\",\"Os\":\"linux\",\"Arch\":\"amd64\",\"KernelVersion\":\"4.9.7-201.fc25.x86_64\",\"BuildTime\":\"Mon Jan 30 16:15:28 2017\",\"PkgVersion\":\"docker-common-1.12.6-6.gitae7d637.fc25.x86_64\"}"
        ], 
        "warnings": []
    }
}

localhost                  : ok=2    changed=1    unreachable=0    failed=0   
```

playbooks/common/openshift-master/config.yml
detiber: I believe the main issue right now is that the variable needed is defined through openshift_facts, which I wouldn't have expected to be available at that point in the playbook. It turns out that we are running openshift_facts very early in the process by including `playbooks/common/initialize_facts.yml` from `playbooks/common/std_include.yml`, which is included from `playbooks/byo/openshift-cluster/config.yml`.

As for a better place to put the check without creating a new role, I'm not sure there is a good answer. Some of the least worst places I can think of:
- Putting the test in a task/playbook include that is pulled in after facts are initialized in `playbooks/common/openshift-cluster/std_include.yml`
- Putting the test directly in `roles/openshift_facts/tasks/main.yml`
- ???

ashcrow:
aos-ci-test

roles/openshift_facts/tasks/main.yml
```diff
@@ -26,6 +26,23 @@
...
 
+# Fail early if Atomic and old version of Docker
+- block:
+
+  # See https://access.redhat.com/articles/2317361
+  # and https://github.com/ansible/ansible/issues/15892
+  # NOTE: the "'s can not be removed at this level else the docker command will fail
+  - name: Determine Atomic Host Docker Version
+    shell: 'CURLY="{"; docker version --format "$CURLY{json .Server.Version}}"'
+    register: l_atomic_docker_version
```
detiber: @ashcrow local testing seems to indicate that:
```
- hosts: localhost
  become: no
  gather_facts: no
  connection: local
  vars:
  tasks:
  - command: docker version --format "{{ version_format }}"
    register: result
    vars:
      version_format: !unsafe '{{json .Server.Version}}'
  - debug: var=result
```
would work to provide the docker version. It could be that I am running 2.2.1 locally, though.

ashcrow: I'll give it a shot and see what happens.

ashcrow: With ```ansible 2.2.0.0``` result ends up being ``"{#json .Server.Version#}"``. It may be something that was fixed in ```2.2.1.x```.

detiber: Of course it is :disappointed: 

Might be good to add a comment so that we can update it with a post-2.2.1 audit

ashcrow: Good idea, I'll add that!

ashcrow:
aos-ci-test

ashcrow:
Failures look unrelated to changes. Retrying.

ashcrow:
aos-ci-test

ashcrow:
@tbielawa merge at will if you still approve.

PR #3476, opened 2017-02-23
[DO NOT MERGE] Include standard group mappings in logging and metrics playbooks

sdodson:


sdodson:
@ewolinetz FYI, the first commit I think does away with the need to add a host to oo_first_master group because this is a group we dynamically generate. I'll see if I can find a better way to set the tags for enterprise later tonight.

roles/openshift_logging/defaults/main.yml
```diff
@@ -1,6 +1,8 @@
 ---
-openshift_logging_image_prefix: "{{ openshift_hosted_logging_deployer_prefix | default('docker.io/openshift/origin-') }}"
-openshift_logging_image_version: "{{ openshift_hosted_logging_deployer_version | default('latest') }}"
+openshift_logging_image_prefix: "{{ openshift_hosted_logging_deployer_prefix | default(
+  'registry.access.redhat.com/openshift3/ose-' if openshift_deployment_type == 'openshift-enterprise'
+  else 'docker.io/openshift/origin-') }}"
```
detiber: This is probably going to be a common thing we need to do. I'm wondering if we either need to start a new `common` or `defaults` role that can define some of these values, that way we aren't intertwining ourselves more into openshift_facts or openshift_common, but can still avoid having to set these defaults everywhere.

Alternatively, maybe something like this would be better as a filter_plugin. We would still need a common location to host it (but we could potentially leverage lib_openshift, or start a similar filters_openshift role)

ewolinetz:
i'm liking the direction this is going. thanks @sdodson !

roles/openshift_logging/defaults/main.yml
ewolinetz: what about the idea of using two variable files `enterprise` and `origin` and specify the default prefixes/versions in there?

detiber: @ewolinetz I don't see why we couldn't go down that route for the simple use cases. It might get a bit tricky with defaults that depend on more than a single variable, think a combination of deployment type and particular versions. I don't think we'd want to maintain a complex matrix of variable files.

ewolinetz: Agreed, I think we would want to only use this for simple cases where we know values based on a single variable/scenario and we're providing values for versions/package names

sdodson:
@ewolinetz PTAL, rebased with the latest changes

playbooks/common/openshift-cluster/openshift_metrics.yml
```diff
@@ -1,4 +1,11 @@
 ---
+# I believe this violates our rule of either includes or plays.
+# Unfortunately we've documented calling this playbook directly so
+# I'd really like to apply the group mappings here rather than up in byo
+- include: ../../common/openshift-cluster/std_include.yml
```
ewolinetz: are we doing something like this in `playbooks/common/openshift-cluster/openshift_logging.yml` ? Why wouldn't we have this in the `byo` metrics playbook?

sdodson: We're doing it in `playbooks/byo/openshift-cluster/openshift_logging.yml`.

Unfortunately docs.openshift.com/org instructs metrics users to call this playbook rather than `playbooks/byo/openshift-cluster/openshift_metrics.yml` I guess I could just make them the same and we can deal with folks having to define oo_first_master for metrics until we release an errata and also update the docs to tell people to use `byo/....`

ewolinetz: makes sense, if that's the intended entry point then I don't see a reason to update this just yet.

sdodson:
aos-ci-test

sdodson:
[merge]

playbooks/byo/openshift-cluster/openshift-logging.yml
```diff
@@ -1,32 +1,7 @@
...
-# This playbook is a preview of upcoming changes for installing
-# Hosted logging on.  See inventory/byo/hosts.*.example for the
-# currently supported method.
-#
-- name: Create initial host groups for localhost
-  hosts: localhost
-  connection: local
-  become: no
-  gather_facts: no
+- include: ../../common/openshift-cluster/std_include.yml
```
mtnbikenc: FYI, by switching from these two tasks to std_include.yml you are adding 1.5 minutes and  230 tasks.

sdodson: Can we automate this sort of analysis?

sdodson:
Pushes new commits to prevent the merge from happening.

playbooks/byo/openshift-cluster/openshift-logging.yml
mtnbikenc: I think you are wanting to include the byo/openshift-cluster/evaluate_groups.yml created in this PR.

ewolinetz:
@sdodson is this good to merge still?

sdodson:
@ewolinetz @mtnbikenc pointed out that this added 1.5minutes and 230 tasks to your playbook. I tried to defray that a bit but we realized that we were finding a lot of cross byo -> common -> byo playbooks.

We're probably going to wait to do something like this until we've had time to properly define some of the early "playbooks calling playbooks" work to ensure that we have a playbook that only defined group mappings without pulling in the rest.

If you think we should just deal with the additional tasks but not try to clean up the playbooks calling playbooks i can revert this PR back to just that.

ewolinetz:
@sdodson I'd prefer that we didn't add the additional tasks. I do see the benefit however of some of these changes... however I think we can wait on them until the playbooks including playbooks details are hashed out

PR #3484, opened 2017-02-24
Fix Bug 1426705 - Quote ansible_ssh_user when determining group id

sdodson:
ansible_ssh_user


sdodson:
aos-ci-test

PR #3486, opened 2017-02-24
Fix Quick Installer failed due to a Python method failure

tbielawa:
Fixes issues dumping various objects to YAML. This change updates the
quick installer fact caching callback to use the improved
AnsibleDumper when writing out the facts to storage.

* https://bugzilla.redhat.com/show_bug.cgi?id=1421053
* https://bugzilla.redhat.com/show_bug.cgi?id=1421061
* https://bugzilla.redhat.com/show_bug.cgi?id=1421059

**NOTE** usage of the `yaml.safe_dump` function was replaced with just `yaml.dump`. This change is inherently and technically incompatible with the `safe_dump` function. The `safe_dump` function already specifies the specific `Dumper` to use. Use of `Dumper=AnsibleDumper` in the `safe_dump` function will cause the error:

> dump_all() got multiple values for keyword argument 'Dumper'

### DIFF
--- a/utils/src/ooinstall/ansible_plugins/facts_callback.py
+++ b/utils/src/ooinstall/ansible_plugins/facts_callback.py
@@ -5,6 +5,7 @@
 import os
 import yaml
 from ansible.plugins.callback import CallbackBase
+from ansible.parsing.yaml.dumper import AnsibleDumper
 
 
 # pylint: disable=super-init-not-called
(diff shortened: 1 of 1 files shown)


tbielawa:
Given the recent churn around the various supported and unsupported Ansible versions I can not make an educated guess as to which Ansible versions this will work with. I just verified it on `ansible-2.2.1.0-2hax` against an OCP 3.5 environment.

tbielawa:
aos-ci-test

sdodson:
2.2.1.0-2.el7 is what we care most about

tbielawa:
**Reviewer note** this follows the pattern already established in

* https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_master_facts/filter_plugins/openshift_master.py#L489
* https://github.com/openshift/openshift-ansible/blob/master/filter_plugins/oo_filters.py#L661

PR #3492, opened 2017-02-24
Fix Bug 1426705 - Quote ansible_ssh_user when determining group id

sdodson:
ansible_ssh_user

### DIFF
--- a/roles/openshift_master_certificates/tasks/main.yml
+++ b/roles/openshift_master_certificates/tasks/main.yml
@@ -149,7 +149,7 @@
   become: no
 
 - name: Lookup default group for ansible_ssh_user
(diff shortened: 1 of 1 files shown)


sdodson:
@abutcher FYI, I think your cert re-deploy work has this same command, lets make sure when backporting that we use the quote filter too

tbielawa:
Yep, checkkit out here @abutcher 

https://github.com/abutcher/openshift-ansible/blob/master/roles/openshift_master_certificates/tasks/main.yml#L161

```yaml
- name: Lookup default group for ansible_ssh_user
  command: "/usr/bin/id -g {{ ansible_ssh_user | quote }}"
  changed_when: false
  register: _ansible_ssh_user_gid
```

PR #3493, opened 2017-02-24
[WIP] Adding support for a route with reencrypt and certificates.

kwoodson:


roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -2,21 +2,51 @@
...
+    src: "{{ item }}"
+  register: openshift_hosted_registry_certificate_content
+  with_items:
+  - "{{ (openshift_hosted_registry_route_certificates | default({'certfile':none})).certfile }}"
+  - "{{ (openshift_hosted_registry_route_certificates | default({'keyfile':none})).keyfile }}"
+  - "{{ (openshift_hosted_registry_route_certificates | default({'cafile':none})).cafile }}"
+  when: openshift_hosted_registry_route_certificates
+
+- debug: var=openshift_hosted_registry_certificate_content
+  when: openshift_hosted_registry_route_certificates
```
detiber: It looks like this is just for debugging, but you could use `with_file` instead of `slurp`, since it is a `local_action`. This would avoid a task being run and also avoids the need to base64 decode the contents.
```
- debug: var=item
  with_files:
  - "{{ (openshift_hosted_registry_route_certificates | default({'certfile':none})).certfile }}"
  - "{{ (openshift_hosted_registry_route_certificates | default({'keyfile':none})).keyfile }}"
  - "{{ (openshift_hosted_registry_route_certificates | default({'cafile':none})).cafile }}"
```

This would

roles/lib_openshift/src/class/oc_route.py
```diff
@@ -55,7 +55,13 @@ def create(self):
 
     def update(self):
         '''update the object'''
-        # need to update the tls information and the service name
+        # when the host attribute is being updated, we need to delete and recreate
```
detiber: It doesn't seem like this should be necessary, I do not see anything in the api docs or api code to indicate that host is immutable.

roles/openshift_hosted/tasks/registry/secure.yml
kwoodson: @detiber, I'll look at this.  This code was taken from another section of code for route.

roles/lib_openshift/src/class/oc_route.py
kwoodson: @detiber, I agree it shouldn't be.  https://bugzilla.redhat.com/show_bug.cgi?id=1426666

If you could test this and confirm.  I've tested on 3.4 and 3.5 and it is immutable.

detiber: Arrgh, I verified this is the case against the api directly as well.

What about something like this instead:
```
force = False
if self.config.host != self.route.host:
    force = True

return self._replace_content(self.kind, self.config.name, self.config.data, force=force)
```

kwoodson: @detiber, this fails as well.

`[root@ip-172-31-54-78 ~]# oc replace -f droute`
`The Route "docker-registry" is invalid: spec.host: Invalid value: "registry1.<removed>.openshift.com": field is immutable`

The bugzilla was updated on why we have to delete/recreate.

kwoodson:
@mtnbikenc, 

In the operations version, when we modify the service for the registry it requires a service restart.  Should I add the handler calls for that?

roles/lib_openshift/src/class/oc_route.py
detiber: @kwoodson I tested `oc replace --force` prior to suggesting it:

```
> oc version
oc v1.5.0-alpha.2+e4b43ee
kubernetes v1.5.2+43a9be4
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://127.0.0.1:8443
openshift v1.4.1+3f9807a
kubernetes v1.4.0+776c994

> oc create route passthrough test-route --service=test --hostname=test.example.com --port=8080
route "test-route" created

> oc get route
NAME         HOST/PORT           PATH      SERVICES   PORT      TERMINATION
test-route   test.example.com             test       8080      passthrough

> oc get route test-route -o yaml > route.tmp

> sed -i -e 's/test.example.com/test2.example.com/'

> oc replace -f route.tmp --force
route "test-route" deleted
route "test-route" replaced

> oc get route test-route
NAME         HOST/PORT           PATH      SERVICES   PORT      TERMINATION
test-route   test2.example.com             test       8080      passthrough
```

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -2,21 +2,43 @@
...
+  copy:
+    backup: True
+    dest: "/etc/origin/master/named_certificates/{{ item | basename }}"
+    src: "{{ item }}"
+  register: openshift_hosted_registry_certificate_content
+  with_items:
+  - "{{ (openshift_hosted_registry_route_certificates | default({'certfile':none})).certfile }}"
+  - "{{ (openshift_hosted_registry_route_certificates | default({'keyfile':none})).keyfile }}"
+  - "{{ (openshift_hosted_registry_route_certificates | default({'cafile':none})).cafile }}"
+  when: openshift_hosted_registry_route_certificates
```
detiber: Instead of defaulting `openshift_hosted_registry_route_certificates` to `false`, you can just default it to:
```
openshift_hosted_registry_route_certificates:
  certfile: None
  keyfile: None
  cafile: None
```
and then update this task to:
```
- name: Get the certificate contents for registry
  copy:
    backup: True
    dest: "/etc/origin/master/named_certificates/{{ item.value | basename }}"
    src: "{{ item.value }}"
  register: openshift_hosted_registry_certificate_content
  when: item.key in ['certfile', 'keyfile', 'cafile'] and item.value is not None
  with_dict: "{{ openshift_hosted_registry_route_certificates }}"
```

roles/lib_openshift/src/class/oc_route.py
kwoodson: @detiber, I didn't actually pass the `--force` in my example.  Thanks for doing that.

I already have  _replace_content take a force flag so I'm updating to call `--force`.

Good find.

roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -40,6 +40,9 @@
     openshift_hosted_registry_images: "{{ openshift.hosted.registry.registryurl | default('openshift3/ose-${component}:${version}')}}"
     openshift_hosted_registry_volumes: []
     openshift_hosted_registry_env_vars: {}
+    openshift_hosted_registry_route_certificates: "{{ openshift.hosted.registry.route.certificates | default(false) }}"
+    openshift_hosted_registry_route_host: "{{ openshift.hosted.registry.route.host | default(false) }}"
```
kwoodson: @detiber, @mtnbikenc 

This is a good question.  Let's see if I can explain:
The host for the route needs to be different depending on the type of route that we are creating.
docker_registry_route_hostname is a product of https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/tasks/registry/secure.yml#L4

and the variable above( openshift_hosted_registry_route_host ) we want to take precedence if it is set:
`    host: "{{ openshift_hosted_registry_route_host | default(docker_registry_route_hostname) }}"`

roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -40,6 +40,9 @@
...
+    openshift_hosted_registry_route_termination: "{{ openshift.hosted.registry.route.termination | default('passthrough') }}"
```
kwoodson: @detiber @abutcher @mtnbikenc 

I'm not sure if moving these to defaults will result in the correct behavior.  I need the variable to default to the openshift.hosted.registry.route.* variable and if that does not exist then default to a different value.

The other way would be to set_fact and set it to the desired when it is defined.  We get that same behavior here I believe.  In the next version we want to refactor how this works so I think variables getting passed in with their proper settings would greatly help.

roles/openshift_hosted/tasks/registry/secure.yml
kwoodson: @detiber, thanks for the suggestion.  I borrowed that code from the previous playbook and it wasn't quite doing what I wanted with the base64 encoding.  I updated this code to do the copy.

kwoodson:
@mtnbikenc, would you mind testing these latest changes.  I ran up to 10 times and it was idempotent each run.  These were with operations settings.  I would like to know if these work with non reencrypt routes and non secure registry.

roles/openshift_hosted/tasks/registry/registry.yml
detiber: I guess I'm just not sure why we'd expose two different routes...  Wouldn't we just want to expose one and have it use the type of route encryption settings we request?

kwoodson: @detiber, 

We are only exposing a single route based upon settings that get passed in.

roles/openshift_hosted/tasks/router/router.yml
```diff
@@ -71,9 +71,6 @@
     - key: spec.strategy.rollingParams.updatePeriodSeconds
       value: 1
       action: put
-    - key: spec.strategy.activeDeadlineSeconds
-      value: 21600
-      action: put
```
mtnbikenc: Removing spec.strategy.activeDeadlineSeconds causes this to report changed each time.

smarterclayton:
Fails for me in the origin GCE runs:

```
TASK [openshift_hosted : set openshift_hosted facts] ***************************
task path: /usr/share/ansible/openshift-ansible/roles/openshift_hosted/tasks/registry/registry.yml:33
Wednesday 01 March 2017  16:40:31 +0000 (0:00:00.039)       0:11:07.132 *******
fatal: [ci-claytontest-ig-m-7v5q]: FAILED! => {
    "failed": true,
    "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'route'\n\nThe error appears to have been in '/usr/share/ansible/openshift-ansible/roles/openshift_hosted/tasks/registry/registry.yml': line 33, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: set openshift_hosted facts\n  ^ here\n"
}
```

roles/openshift_hosted/tasks/registry/registry.yml
detiber: @kwoodson that's why I'm confused about why we have two different variables. If we are deprecating docker_registry_route_hostname, then I don't see why doing something like this wouldn't work:

in vars/main.yml
```
osh_default_subdomain: router.default.svc.cluster.local
oshr_default_route_host: "{{ 'docker-registry-default.' ~ (openshift_master_default_subdomain | default(default_subdomain, true) }}"
```
in defaults/main.yml
```
openshift_hosted_registry_route_host: "{{ docker_registry_route_hostname | oshr_default_route_host }}"
```

Then tasks/registry/secure.yml could be updated to:
- remove the `set_fact` for `docker_registry_route_hostname`
- use `openshift_hosted_registry_route_host` in place of `docker_registry_route_hostname`

Granted, this doesn't take into account insecure registries, but I think you could conditionalize the default value for that:
```
oshr_secure: "{{ not openshift.docker.hosted_registry_insecure | default(false) }}"
osh_default_subdomain: router.default.svc.cluster.local
oshr_default_route_host_secure: "{{ 'docker-registry-default.' ~ (openshift_master_default_subdomain | default(default_subdomain, true) }}"
oshr_default_route_host: "{{ oshr_default_route_host_secure if oshr_secure else None }}" 
```

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -2,21 +2,54 @@
 - name: Set fact docker_registry_route_hostname
   set_fact:
     docker_registry_route_hostname: "{{ 'docker-registry-default.' ~ (openshift_master_default_subdomain | default('router.default.svc.cluster.local', true)) }}"
+  run_once: true
```
mtnbikenc: run_once is on the include, is this necessary?

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -2,21 +2,54 @@
...
+    dest: "/etc/origin/master/named_certificates/{{ item.value | basename }}"
+    src: "{{ item.value }}"
+  when: item.key in ['certfile', 'keyfile', 'cafile'] and item.value is not None
+  with_dict: "{{ openshift_hosted_registry_routecertificates }}"
+  when: openshift_hosted_registry_routecertificates
+
+# When certificates are defined we will create the reencrypt
+# docker-registry route
+- name: Create a reencrypt route for docker-registry
+  run_once: true
```
mtnbikenc: run_once is on the include, is this necessary?

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -2,21 +2,54 @@
...
+  when: openshift_hosted_registry_routetermination == 'passthrough'
 
 - name: Retrieve registry service IP
   oc_service:
     namespace: "{{ openshift_hosted_registry_namespace }}"
     name: docker-registry
     state: list
   register: docker_registry_service_ip
+  run_once: true
```
mtnbikenc: run_once is on the include, is this necessary?

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -2,21 +2,54 @@
...
 - name: Create passthrough route for docker-registry
   oc_route:
     name: docker-registry
     namespace: "{{ openshift_hosted_registry_namespace }}"
     service_name: docker-registry
-    tls_termination: passthrough
-    host: "{{ docker_registry_route_hostname }}"
+    tls_termination: "{{ openshift_hosted_registry_routetermination }}"
+    host: "{{ openshift_hosted_registry_routehost | ternary(openshift_hosted_registry_routehost, docker_registry_route_hostname) }}"
+  run_once: true
```
mtnbikenc: run_once is on the include, is this necessary?

roles/openshift_hosted/tasks/router/router.yml
kwoodson: Adding this back in.

kwoodson:
@smarterclayton, Would you be able to test this with your setup?  I have run it against 2 separate setups and have deployed the router and registry successfully.

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -3,13 +3,42 @@
...
     docker_registry_route_hostname: "{{ 'docker-registry-default.' ~ (openshift_master_default_subdomain | default('router.default.svc.cluster.local', true)) }}"
 
+- name: Get the certificate contents for registry
+  copy:
+    backup: True
+    dest: "/etc/origin/master/named_certificates/{{ item.value | basename }}"
+    src: "{{ item.value }}"
+  when: item.key in ['certfile', 'keyfile', 'cafile'] and item.value is not None
+  with_dict: "{{ openshift_hosted_registry_routecertificates }}"
+  when: openshift_hosted_registry_routecertificates
```
mtnbikenc: lint: duplicate `when:` condition.

mtnbikenc:
aos-ci-test

smarterclayton:
Testing now

PR #3499, opened 2017-02-27
Don't install python-ruamel-yaml

sdodson:
Now that we fallback to yaml from ruamel.yaml we can safely remove the tasks that were installing this.


sdodson:
aos-ci-test

roles/lib_utils/tasks/main.yml
```diff
@@ -8,4 +8,6 @@
   package:
     name: python-ruamel-yaml
```
kwoodson: This says python-ruamel-yaml.   I placed python2-ruamel-yaml in lib_openshift.  Should they be the same? I know that python2-ruamel versione exists.  The other does not, yet.

I'm out of the loop on what the official one will be.

sdodson: Yes, they should be uniform. Looking through the packages that are in Fedora and in our repo python2-ruamel-yaml seems to be the package that's common amoungst them, however is this the right package to install on a target host that's running Fedora 25? I'm not sure if ansible will use python2 or python3?

```
[root@ose3-master ~]# rpm -q --provides python2-ruamel-yaml
python-ruamel-yaml = 0.12.14-9.el7
python-ruamel.yaml = 0.12.14-9.el7
python2-ruamel-yaml = 0.12.14-9.el7
python2-ruamel-yaml(x86-64) = 0.12.14-9.el7

[root@ose3-master ~]# rpm -qp -provides https://kojipkgs.fedoraproject.org//packages/python-ruamel-yaml/0.13.13/3.fc25/x86_64/python2-ruamel-yaml-0.13.13-3.fc25.x86_64.rpm
python-ruamel.yaml = 0.13.13-3.fc25
python-ruamel.yaml(x86-64) = 0.13.13-3.fc25
python2-ruamel-yaml = 0.13.13-3.fc25
python2-ruamel-yaml(x86-64) = 0.13.13-3.fc25
python2.7dist(ruamel.yaml) = 0.13.13
python2dist(ruamel.yaml) = 0.13.13
```

sdodson: And the python3 package from F25
```
$ rpm -qp --provides https://kojipkgs.fedoraproject.org//packages/python-ruamel-yaml/0.13.13/3.fc25/x86_64/python3-ruamel-yaml-0.13.13-3.fc25.x86_64.rpm
python3-ruamel-yaml = 0.13.13-3.fc25
python3-ruamel-yaml(x86-64) = 0.13.13-3.fc25
python3.5dist(ruamel.yaml) = 0.13.13
python3dist(ruamel.yaml) = 0.13.13
```

detiber: I think it'll probably be a good idea to conditionalize python2-ruamel-yaml/python3-ruamel-yaml depending on not-Fedora/Fedora

sdodson:
aos-ci-test

roles/lib_utils/tasks/main.yml
```diff
@@ -6,6 +6,8 @@
 
 - name: lib_utils ensure python-ruamel-yaml package is on target
   package:
-    name: python-ruamel-yaml
+    name: "{{ 'python2-ruamel-yaml' if ansible_distribution != 'Fedora' else 'python3-ruamel-yaml' }}"
```
detiber: I mentioned it in IRC, but didn't specifically mention you.

You could use: `ansible_python.version.major`, which should let you do this:
```
- name: lib_utils ensure python-ruamel-yaml package is installed
  package:
    name: "python{{ ansible_python.version.major }}-ruamel-yaml"
```

sdodson: ok, updated.

roles/lib_openshift/tasks/main.yml
```diff
@@ -6,6 +6,8 @@
 
 - name: lib_openshift ensure python-ruamel-yaml package is on target
   package:
-    name: python2-ruamel-yaml
+    name: "python{{ ansible_python.version.major }}-ruamel-yaml"
     state: present
-  when: not ostree_booted.stat.exists
+  when:
+  - not containerized | default(false)
```
detiber: Why test for containerized? Unless the remote system is atomic, we should go ahead and install dependencies that affect modules.

sdodson:
> Why test for containerized? Unless the remote system is atomic, we should go ahead and install dependencies that affect modules.

Yes, that's what I had done originally, however QE is saying that we have no other dependency on the OCP channels on containerized hosts so we're effectively introducing a new requirement on those channels. So I'm punting until we have a real solution for python dependencies on containerized hosts.

detiber:
> Yes, that's what I had done originally, however QE is saying that we have no other dependency on the OCP channels on containerized hosts so we're effectively introducing a new requirement on those channels. So I'm punting until we have a real solution for python dependencies on containerized hosts.

Do they not consider PyYAML a dependency?

Currently all of the dependencies in openshift_facts are installed unless `is_atomic` is `True`. I think we are treading on dangerous ground if we try to avoid installing dependencies for non-atomic containerized installs, because we cannot assume that a base set of packages is present. If a user is building a modified minimal base image, then we cannot guarantee packages that we need are present.

sdodson:
> Do they not consider PyYAML a dependency?

I think the difference here is that all of the openshift_facts dependencies are in base channels where as python-ruamel-yaml is only in OCP channels and we have a confirmed fallback. Though I guess this means we should add PyYAML here as a dependency as once we de-couple openshift_facts we cannot assume it to be installed.

sdodson:
I'm going to go ahead and merge this if tests pass. I agree with the argument that we should be installing the dependency but we'll defer that until we have a more firm strategy for handling ansible dependencies on containerized hosts.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

detiber:
@sdodson generation validation not happy...  I can setup email alerts for when validation breaks after a merge, which would help notify us that something is broken without resorting on failed tests in PRs.

sdodson:
> @sdodson generation validation not happy... I can setup email alerts for when validation breaks after a merge, which would help notify us that something is broken without resorting on failed tests in PRs.

I'm confused, the last PR that merged travis was green on generated content as far as I can tell. So I don't get why this is failing but I also see new content when i generate stuff again. Have I screwed up or did travis let something slip in it shouldn't have?

sdodson:
Nevermind, I see what happened. Travis tests on #3527 weren't current when it merged, a new module had been added.

PR #3501, opened 2017-02-27
[1.3] Stop restarting node after upgrading master rpms

sdodson:
This looks to be causing a customer issue where some HA upgrades fail,
due to a missing EgressNetworkPolicy API. We update master rpms, we
don't restart services yet, but then restart node service which tries to
talk to an API that does not yet exist. (pending restart)

Restarting node here is very out of place and appears to not be
required.

Bug 1426888

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
@@ -5,7 +5,3 @@
 - name: Ensure python-yaml present for config upgrade
   action: "{{ ansible_pkg_mgr }} name=PyYAML state=present"
   when: not openshift.common.is_atomic | bool
-
-- name: Restart node service
-  service: name="{{ openshift.common.service_type }}-node" state=restarted
-  when: component == "node"


sdodson:
@dgoodwin mind validating if this is appropriate for release-1.3? As far as I can tell it is.

playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
```diff
@@ -5,7 +5,3 @@
 - name: Ensure python-yaml present for config upgrade
   action: "{{ ansible_pkg_mgr }} name=PyYAML state=present"
   when: not openshift.common.is_atomic | bool
```
detiber: Should these also be updated to use the package module while in here?

detiber: Should the Upgrade packages task above be updated to notify a handler? Or are we forcing a node restart later in the process.

sdodson: I'd rather keep things in line with the majority of the release-1.3 branch which uses this pattern. The only place we've used package module in this branch is when we backported conntrack installation from master.
Service restarts are handled later in the playbook.

PR #3502, opened 2017-02-27
install the latest excluders

ingvagabund:
We want the latest version installed during upgrade

Tested the upgrade, rpms are updating when 3.4 versions installed.

Fixes: 1426070

### DIFF
--- a/roles/openshift_excluder/tasks/status.yml
+++ b/roles/openshift_excluder/tasks/status.yml
@@ -20,6 +20,7 @@
 - name: Update to latest excluder packages
   package:
     name: "{{ openshift.common.service_type }}-excluder"
+    state: latest
   when:
   - "{{ openshift_excluder_installed.installed_versions | default([]) | length > 0 }}"
   - not openshift.common.is_containerized | bool
@@ -27,6 +28,7 @@
 - name: Update to the latest docker-excluder packages
   package:
     name: "{{ openshift.common.service_type }}-docker-excluder"
+    state: latest
   when:
   - "{{ docker_excluder_installed.installed_versions | default([]) | length > 0 }}"
   - not openshift.common.is_containerized | bool


ingvagabund:
aos-ci-test

roles/openshift_excluder/tasks/status.yml
```diff
@@ -20,13 +20,15 @@
 - name: Update to latest excluder packages
   package:
     name: "{{ openshift.common.service_type }}-excluder"
+    state: latest
```
detiber: Should we be updating the excluder package outside of an upgrade?

ingvagabund: If the excluders are enabled, they must be updated during the upgrade. Otherwise, they are of no use.

sdodson: For atomic-openshift-excluder/origin-excluder this doesn't matter as those are unversioned excludes.
For atomic-openshift-docker-excluder it's ok because they're bound by the version of the excluder that's specific to a given release, we should never ship a new excluder to the same repo which would introduce a docker version that's not compatible.
For origin-docker-excluder this will be problematic right now because multiple versions are in one repo however that won't be the case for 3.6 (I think?). However, we've only provided excluders for 1.4 and soon 1.5 which both support docker 1.12.

So I think this is OK.

detiber: @sdodson @ingvagabund Is the openshift_excluder role applied outside of the upgrade playbooks? My concern with updating the excluder packages in `status.yml` is that we would update the package outside of the upgrade workflow.

ingvagabund: The `openshift_excluder` is included in `playbooks/common/openshift-cluster` playbook which is included in both deploy and upgrade playbooks.

detiber: @ingvagabund in that case, I would suggest that the install workflow use `state: present` and having the upgrade workflow invoke tasks that use `state: latest`.

Otherwise, if a user re-runs the install playbook (say they are using the install playbook in Ansible tower for ongoing configuration management) they could end up with newer version of the excluder inadvertently.

Granted, it isn't a huge issue, since we would expect them to pull in a new excluder with system updates, but I'd leave that to the user's discretion.

ingvagabund: Lemme polish the excluders

ingvagabund: I will merge this PR and open another one making the excluders more safe.

ingvagabund: After further analysis:
1) `playbooks/byo/config.yml` includes `playbooks/common/openshift-cluster/config.yml`
2) `playbooks/common/openshift-cluster/config.yml` includes `playbooks/common/openshift-cluster/disable_excluder.yml`
3) `playbooks/common/openshift-cluster/disable_excluder.yml` includes `roles/openshift_excluder/tasks/status.yml`

So the excluders are used during installation as well.

ingvagabund: It does not look nice: https://github.com/openshift/openshift-ansible/pull/3529

sdodson: Unfortunately earlier versions of the excluders lacked a status function so I'd really like to ensure they're up to date so that we can rely on the status function. I've also just realized that we're not actually calling the status function below so we should fix that.

sdodson: https://github.com/openshift/openshift-ansible/pull/3502/files#diff-b6b3dc3686358fd5aacff902dccb98a9R2 explains why i'd updated them in this task file.

detiber: @sdodson @ingvagabund: I still think it'd be better to ensure a minimum version of the excluder rather than doing an update each time.

ingvagabund: Let's move the discussion here https://github.com/openshift/openshift-ansible/pull/3529#discussion_r103940333

PR #3504, opened 2017-02-27
Separate s3 cloudfront set_fact calls

ashcrow:
7cf5cc14 cleaned up how the registry was being created. However the s3
cloudfront calls ended up setting and using a fact in the same block.
This change pulls the set_fact calls apart.

### DIFF
--- a/roles/openshift_hosted/tasks/registry/storage/s3.yml
+++ b/roles/openshift_hosted/tasks/registry/storage/s3.yml
@@ -36,13 +36,14 @@
       - path: cloudfront.pem
         data: "{{ lookup('file', openshift_hosted_registry_storage_s3_cloudfront_privatekeyfile) }}"
 
-  - name: Add cloudfront secret to the registry volumes
+  - name: Append cloudfront secret registry volume to openshift_hosted_registry_volumes
     set_fact:
+      openshift_hosted_registry_volumes: "{{ openshift_hosted_registry_volumes | union(s3_volume_mount) }}"
+    vars:
       s3_volume_mount:
       - name: cloudfront-vol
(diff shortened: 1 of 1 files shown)


roles/openshift_hosted/tasks/registry/storage/s3.yml
```diff
@@ -36,13 +36,16 @@
...
     set_fact:
       s3_volume_mount:
       - name: cloudfront-vol
         path: /etc/origin
         type: secret
         secret_name: docker-registry-s3-cloudfront
+
+  - name: Append cloudfront secret registry volume to openshift_hosted_registry_volumes
+    set_fact:
       openshift_hosted_registry_volumes: "{{ openshift_hosted_registry_volumes | union(s3_volume_mount) }}"
```
detiber: I think there has to be a better way than using set_fact, but that would require quite a bit of refactoring.

This should be able to be brought down to a single set_fact by doing something like this:
```
- name: Append cloudfront secret registry volume to openshift_hosted_registry_volumes
  set_fact:
    openshift_hosted_registry_volumes: "{{ openshift_hosted_registry_volumes | union(s3_volume_mount) }}"
  vars:
    s3_volume_mount:
    - name: cloudfront-vol
      path: /etc/origin
      type: secret
      secret_name: docker-registry-s3-cloudfront
```

ashcrow: @detiber I like that!

ashcrow: Updated.

ashcrow:
aos-ci-test

Issue #3505, opened 2017-02-27
Suggest: add in README if OpenShift Enterprise x to y work for origin version


harobed:
Hi,

in this upgrade [README](https://github.com/openshift/openshift-ansible/tree/master/playbooks/byo/openshift-cluster/upgrades) page I read:

```
Upgrades available

* OpenShift Enterprise 3.4 to 3.5
* OpenShift Enterprise 3.3 to 3.4
```

questions: do this upgrade playbooks work with origin version?

If the answer is yes, can you update with this:

```
Upgrades available

* OpenShift Enterprise 3.4 to 3.5 (works for origin-1.4.x to origin-1.5.x)
* OpenShift Enterprise 3.3 to 3.4 (works for origin-1.3.x to origin-1.4.x)
```

What do you think about this suggestion?

Best regards,
Stéphane

harobed:
I see that: https://github.com/openshift/openshift-ansible/blob/master/playbooks/byo/openshift-cluster/upgrades/v3_5/upgrade.yml#L15

then I think that the answer is yes. So, I'm going to create a pull request.

PR #3506, opened 2017-02-27
Fix issue #3505, add notes about origin upgrade versions support in BYO upgrade README file

harobed:
Fix issue https://github.com/openshift/openshift-ansible/issues/3505


harobed:
Thanks

PR #3509, opened 2017-02-27
Restructure openshift_logging into component-based roles

ewolinetz:
The intent of this is to split up the `openshift_logging` role into smaller component roles based on each of the components: `elasticsearch`, `kibana`, `curator`, and `fluentd`.

This should allow us to more easily provide configuration options based on a specific version and keep the openshift_logging role as more of a composed role with more simplified logic. Given that each component may upgrade differently and their configurations may not change from release to release, they would make sense for the logic, files, and defaults to be confined to their specific component.


ewolinetz:
I may update the dependencies for `openshift_logging` to not include the sub-roles and just commit the `openshift_logging_fluentd` role so that it can be used to resolve some bzs.  I also want to further remove the dependency on `openshift_facts`

roles/openshift_logging_fluentd/tasks/main.yaml
```diff
@@ -0,0 +1,225 @@
...
+- name: Set cluster-reader permissions for Fluentd
+  command: >
+    {{ openshift.common.admin_binary}} --config={{ mktemp.stdout }}/admin.kubeconfig policy
+    add-cluster-role-to-user cluster-reader system:serviceaccount:{{openshift_logging_namespace}}:aggregated-logging-fluentd
+  register: fluentd_cluster_reader
+  failed_when:
+  - fluentd_cluster_reader.rc == 1
+  - 'exists' not in fluentd_cluster_reader.stderr
+  check_mode: no
+  when: cluster_readers_output.stdout.find("system:serviceaccount:{{openshift_logging_namespace}}:aggregated-logging-fluentd") == -1
```
t0ffel: add-cluster-role-to-user seem to be idempotent. It doesn't return error for me if the user already has that role

roles/openshift_logging_fluentd/tasks/main.yaml
```diff
@@ -0,0 +1,225 @@
...
+      data: "{{key_pairs | entry_from_named_pair('ca_file')| b64decode }}"
+    - path: key
+      data: "{{key_pairs | entry_from_named_pair('fluentd_key')| b64decode }}"
+    - path: cert
+      data: "{{key_pairs | entry_from_named_pair('fluentd_cert')| b64decode }}"
+
+# create Fluentd daemonset
+- name: Generate logging-fluentd daemonset definition
+  template:
+    src: {{ openshift_logging_fluentd_version }}/fluentd.j2
```
t0ffel: where is openshift_logging_fluentd_version defined?

roles/openshift_logging_fluentd/tasks/main.yaml
```diff
@@ -0,0 +1,225 @@
...
+  changed_when: no
+
+- name: Set privileged permissions for Fluentd
+  command: >
+    {{ openshift.common.admin_binary}} --config={{ mktemp.stdout }}/admin.kubeconfig policy
+    add-scc-to-user privileged system:serviceaccount:{{openshift_logging_namespace}}:aggregated-logging-fluentd
+  register: fluentd_scc
+  failed_when:
+  - fluentd_scc.rc == 1
+  - 'exists' not in fluentd_scc.stderr
```
ewolinetz: It should exist at this point, I make sure that it's created above.
This should only be executed when its not in the SCC, correct.

roles/openshift_logging_fluentd/tasks/main.yaml
ewolinetz: Agreed. This actually should be updated, IIRC adding `-o name` to the command above will print out the names of objects that change. So we can use that to detect a `changed_when` instead of doing a check first.

roles/openshift_logging_fluentd/tasks/main.yaml
ewolinetz: i had started hashing out logic for versions at the end of the day yesterday. Looks like I used two different variable names. I'll either change this to be `fluentd_version` or update the fact set in `determine_version.yaml` to match the above

ewolinetz:
For the ES portion we'll want to take advantage of https://github.com/openshift/openshift-ansible/pull/3692 in lieu of pvc templates

roles/openshift_logging_fluentd/tasks/main.yaml
ewolinetz: looks like there is this now https://github.com/openshift/openshift-ansible/blob/master/roles/lib_openshift/library/oc_adm_policy_group.py

ewolinetz:
I think i may end up removing the 'secure-forward.conf' section from the logging-fluentd configmap in lieu of being able to just configure it from the deployment type that can be specified in ansible

roles/openshift_logging_fluentd/templates/3_6/fluent.conf.j2
```diff
@@ -0,0 +1,43 @@
...
+## sources
+{% if deploy_type in ['hosted', 'secure-aggregator'] %}
+## ordered so that syslog always runs last...
+@include configs.d/openshift/input-pre-*.conf
+@include configs.d/dynamic/input-docker-*.conf
+@include configs.d/dynamic/input-syslog-*.conf
+@include configs.d/openshift/input-post-*.conf
+##
+{% else %}
+
```
ewolinetz: TODO: fill this section out for secure-host

roles/openshift_logging_fluentd/templates/3_6/fluent.conf.j2
```diff
@@ -0,0 +1,43 @@
...
+##
+
+## matches
+  @include configs.d/openshift/output-pre-*.conf
+  @include configs.d/openshift/output-operations.conf
+  @include configs.d/openshift/output-applications.conf
+  # no post - applications.conf matches everything left
+##
+{% else %}
+
```
ewolinetz: TODO: fill this section out for secure-aggregator

roles/openshift_logging_fluentd/defaults/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+### General logging settings
+openshift_logging_image_prefix: "{{ openshift_hosted_logging_deployer_prefix | default('docker.io/openshift/origin-') }}"
+openshift_logging_image_version: "{{ openshift_hosted_logging_deployer_version | default('latest') }}"
+openshift_logging_image_pull_secret: "{{ openshift_hosted_logging_image_pull_secret | default('') }}"
+openshift_logging_master_url: "https://kubernetes.default.svc.{{ openshift.common.dns_domain }}"
+openshift_logging_master_public_url: "{{ openshift_hosted_logging_master_public_url | default('https://' + openshift.common.public_hostname + ':' ~ (openshift_master_api_port | default('8443', true))) }}"
+openshift_logging_namespace: logging
+openshift_logging_tempdir: ''
+
```
richm: This is a lot of stuff that is duplicated in every logging role - can this be consolidated into a `roles/openshift_logging_common` or the existing `roles/openshift_logging` or something like that?

roles/openshift_logging_fluentd/tasks/main.yaml
```diff
@@ -0,0 +1,258 @@
...
+        src: "{{ fluentd_version }}/secure-forward.conf"
+        dest: "{{ tempdir }}/secure-forward.conf"
+      when: fluentd_securefoward_contents is undefined
+
+      changed_when: no
+
+    - copy:
+        content: "{{ fluentd_config_contents }}"
+        dest: "{{ tempdir }}/fluent.conf"
+      when: fluentd_config_contents is defined
```
richm: This will replace the default fluent.conf created by the fluent.conf.j2 template?

roles/openshift_logging_fluentd/tasks/main.yaml
```diff
@@ -0,0 +1,258 @@
...
+
+- name: Label nodes for Fluentd deployment
+  oc_label:
+    name: "{{ fluentd_host }}"
+    kind: node
+    state: add
+    labels: "{{ openshift_logging_fluentd_nodeselector | oo_dict_to_list_of_dict }}"
+  with_items: "{{ openshift_logging_fluentd_hosts }}"
+  loop_control:
+    loop_var: fluentd_host
```
richm: Is there some way to have a delay in here?  For example, if you have 1000 nodes, it will clobber openshift to label all of those nodes at the same time.

richm:
Why do you need different versions of the config files/templates?  Does this mean you will be able to use openshift-ansible release-1.6 to deploy logging version 1.5?

roles/openshift_logging_fluentd/defaults/main.yml
ewolinetz: The intent is that each component role should be self contained so there will be duplication. The names should likely be more specific in each component role. I will also have the current `openshift_logging` role setting those values when it calls the roles. So the user's inventory could look the same as it does now

roles/openshift_logging_fluentd/tasks/main.yaml
ewolinetz: I'll make sure its doing what I expect, but it should be that if you provide a fluentd config in that variable we will use that instead of the template provided one

roles/openshift_logging_fluentd/tasks/main.yaml
ewolinetz: :+1: i will investigate

ewolinetz:
@richm good point, i forget that there are different releases for the ansible roles. I was thinking I needed to cover that from master. That makes this much more simple :)

richm:
@ewolinetz I know that openshift-ansible carries example files for older versions e.g.
```
./roles/openshift_examples/files/examples/v1.3
./roles/openshift_hosted_templates/files/v1.3
```
but I don't know why.

roles/openshift_logging_fluentd/tasks/label_and_wait.yaml
```diff
@@ -6,11 +6,5 @@
...
-# give each pod 2 minutes to either start or fail
-- command: >
-    /usr/bin/oc get pods -o jsonpath='{.items[?(@.spec.nodeName == "{{ node }}")].status.phase}'
-  register: fluentd_status
-  until: fluentd_status.stdout in ['Running', 'Failed']
-  retries: 12
-  delay: 10
-  changed_when: false
+# wait half a second between labels
+- command: sleep {{ openshift_logging_fluentd_label_delay | default('.5') }}
```
richm: `local_action: command sleep ...`
doesn't need to be executed remotely

richm: otherwise, ack

roles/openshift_logging/tasks/install_logging.yaml
```diff
@@ -2,85 +2,141 @@
...
-  when: not ansible_check_mode
+    generated_certs_dir: "{{openshift.common.config_base}}/logging"
 
-- include: update_master_config.yaml
+## Elasticsearch
+
+# TODO: add more vars
+# We don't allow scaling down of ES nodes currently
+- include_role:
+    name: openshift_logging_elasticsearch
```
ewolinetz: should be able to pass in the intended cluster size

roles/openshift_logging_elasticsearch/defaults/main.yml
```diff
@@ -0,0 +1,55 @@
+---
+### Common settings
+openshift_logging_elasticsearch_image_prefix: "{{ openshift_hosted_logging_deployer_prefix | default('docker.io/openshift/origin-') }}"
+openshift_logging_elasticsearch_image_version: "{{ openshift_hosted_logging_deployer_version | default('latest') }}"
+openshift_logging_elasticsearch_image_pull_secret: "{{ openshift_hosted_logging_image_pull_secret | default('') }}"
+openshift_logging_elasticsearch_namespace: logging
```
ewolinetz: intended cluster size for correctly setting quorum in DCs

roles/openshift_logging/tasks/install_logging.yaml
```diff
@@ -2,85 +2,141 @@
...
+    openshift_logging_curator_nodeselector: "{{ openshift_logging_curator_ops_nodeselector }}"
+  when:
+  - openshift_logging_use_ops | bool
+
+
+## Fluentd
+- include_role:
+    name: openshift_logging_fluentd
+  vars:
+    generated_certs_dir: "{{openshift.common.config_base}}/logging"
```
ewolinetz: use_journal should be set based on docker facts gathered

ewolinetz:
rebased to master, need to cherry pick the different fixes that have been merged into master since starting this work

roles/openshift_logging_fluentd/tasks/main.yaml
```diff
@@ -0,0 +1,194 @@
...
+  when: fluentd_secureforward_contents is defined
+  changed_when: no
+
+- name: Set Fluentd configmap
+  oc_configmap:
+    state: present
+    name: "logging-fluentd"
+    namespace: "{{ openshift_logging_namespace }}"
+    from_file:
+      fluentd.conf: "{{ tempdir }}/fluent.conf"
```
richm: fluent.conf

roles/openshift_logging_mux/tasks/main.yaml
```diff
@@ -0,0 +1,167 @@
...
+  when: fluentd_mux_secureforward_contents is defined
+  changed_when: no
+
+- name: Set Mux configmap
+  oc_configmap:
+    state: present
+    name: "logging-mux"
+    namespace: "{{ openshift_logging_mux_namespace }}"
+    from_file:
+      fluentd.conf: "{{ tempdir }}/fluent-mux.conf"
```
richm: fluent.conf

roles/openshift_logging_elasticsearch/templates/elasticsearch.yml.j2
```diff
@@ -0,0 +1,80 @@
...
+  number_of_replicas: 0
+  auto_expand_replicas: 0-2
+  unassigned.node_left.delayed_timeout: 2m
+  translog:
+    flush_threshold_size: 256mb
+    flush_threshold_period: 5m
+
+node:
+  master: {% if deploy_type in ['data-master', 'master'] %}true{% else %}false{% endif %}
+  data: {% if deploy_type in ['data-master', 'data-client'] %}true{% else %}false{% endif %}
```
richm: not sure why but this is being rendered as
```
node:
  master: true  data: true
```
which causes this problem:
```
Exception in thread "main" SettingsException[Failed to load settings from [elast
icsearch.yml]]; nested: MarkedYAMLException[mapping values are not allowed here
 in 'reader', line 18, column 21:
      master: true  data: true
                        ^
 at [Source: cluster:
  name: ${CLUSTER_NAME}
```

richm:
after manually editing the es, mux, and fluentd configmaps as mentioned in the comments above, I was able to start all of the logging components

ewolinetz:
Missing services

roles/openshift_logging_elasticsearch/templates/es.j2
```diff
@@ -0,0 +1,114 @@
...
+    component: elasticsearch
+    deployment: "{{deploy_name}}"
+    logging-infra: "{{logging_component}}"
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      name: "{{deploy_name}}"
+      labels:
+        logging-infra: "{{logging_component}}"
```
ewolinetz: logging-infra: elasticsearch

roles/openshift_logging_elasticsearch/templates/es.j2
```diff
@@ -0,0 +1,114 @@
...
+        provider: openshift
+        component: elasticsearch
```
ewolinetz: component: {{logging_component}}

richm:
missing the mux service:
```
# oc get svc
NAME                     CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
logging-es               172.30.27.221    <none>        9200/TCP   36m
logging-es-cluster       172.30.177.235   <none>        9300/TCP   36m
logging-es-ops           172.30.203.159   <none>        9200/TCP   36m
logging-es-ops-cluster   172.30.115.197   <none>        9300/TCP   36m
logging-kibana           172.30.68.255    <none>        443/TCP    36m
logging-kibana-ops       172.30.242.251   <none>        443/TCP    36m
```

ewolinetz:
@richm should be generated now -- currently unable to correctly create the external one due to https://github.com/openshift/openshift-ansible/issues/4090

richm:
not sure what changed, but now the es configmap is called logging-elasticsearch, but the es dc is still referring to logging-elasticsearch-data-master as the configmap name - es does not start

ewolinetz:
At this point it seems to be able to correct deploy an out of the box logging solution that all the components are able to connect to and begin sending logs to ES (also tested with ops and mux enabled)

ewolinetz:
todo: refactor and remove unused files from openshift_logging

ewolinetz:
I think there are still some changes that are missing from the current master, but otherwise PTAL
@portante @jcantrill @t0ffel @richm @sdodson

roles/openshift_logging_elasticsearch/tasks/main.yaml
```diff
@@ -0,0 +1,282 @@
...
+    name: "{{ openshift_logging_elasticsearch_pvc_name }}"
+    namespace: "{{ openshift_logging_elasticsearch_namespace }}"
+    files:
+    - "{{ tempdir }}/templates/logging-es-pvc.yml"
+    delete_after: true
+  when:
+  - openshift_logging_elasticsearch_storage_type == "pvc"
+
+- set_fact:
+    es_deploy_name: "logging-{{ es_component }}-{{ openshift_logging_elasticsearch_deployment_type }}-{{ 'abcdefghijklmnopqrstuvwxyz0123456789' | random_word(8) }}"
```
jcantrill: i believe we now have oo_random_word where you dont have to pass in the pattern.  Also, any thoughts on moving us to a 0-N based deployment scheme in anticipation of statfulsets?

roles/openshift_logging_elasticsearch/templates/elasticsearch-logging.yml.j2
```diff
@@ -1,25 +1,14 @@
 # you can override this using by setting a system property, for example -Des.logger.level=DEBUG
 es.logger.level: INFO
-rootLogger: ${es.logger.level}, {{root_logger}}
+rootLogger: ${es.logger.level}, console, file
```
jcantrill: Why are we reverting the root logger change?

roles/openshift_logging_elasticsearch/templates/elasticsearch.yml.j2
```diff
@@ -6,16 +6,16 @@ script:
   indexed: on
 
 index:
-  number_of_shards: {{ es_number_of_shards | default ('1') }}
-  number_of_replicas: {{ es_number_of_replicas | default ('0') }}
+  number_of_shards: 1
```
jcantrill: We decided to not allow tweaks down the line?

roles/openshift_logging_elasticsearch/templates/es.j2
```diff
@@ -95,13 +103,6 @@ spec:
               readOnly: true
             - name: elasticsearch-storage
               mountPath: /elasticsearch/persistent
-          readinessProbe:
```
jcantrill: Did we role back the readyness prob?

roles/openshift_logging_elasticsearch/vars/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+__latest_es_version: "3_5"
+__allowed_es_versions: ["3_5", "3_6"]
+__allowed_es_types: ["data-master", "data-client", "master", "client"]
+
+# TODO: integrate these
+openshift_master_config_dir: "{{ openshift.common.config_base }}/master"
+es_node_quorum: "{{ openshift_logging_elasticsearch_replica_count | int/2 + 1 }}"
+es_min_masters_default: "{{ (openshift_logging_elasticsearch_replica_count | int / 2 | round(0,'floor') + 1) | int }}"
```
jcantrill: This logic was moved into 'es_node_quorum' and the min masters removed...

roles/openshift_logging_elasticsearch/templates/elasticsearch.yml.j2
richm: @ewolinetz whoops, looks like I missed that one, should be
```
number_of_shards: {{ es_number_of_shards | default ('1') }}
```
You'll also need the corresponding change for `roles/openshift_logging_elasticsearch/tasks/main.yaml` where the configmap is created - you'll need to add something like this:
```
vars:
...
       - es_number_of_shards: "{{ openshift_logging_es_number_of_shards | default(1) }}"
```

roles/openshift_logging_curator/templates/curator.j2
```diff
@@ -89,10 +89,15 @@ spec:
             - name: config
               mountPath: /etc/curator/settings
               readOnly: true
+            - name: elasticsearch-storage
```
richm: What is this for?

roles/openshift_logging_curator/templates/curator.j2
```diff
@@ -89,10 +89,15 @@ spec:
...
+              mountPath: /elasticsearch/persistent
+              readOnly: true
       volumes:
         - name: certs
           secret:
             secretName: logging-curator
         - name: config
           configMap:
             name: logging-curator
+        - name: elasticsearch-storage
```
richm: What is this for?

roles/openshift_logging_elasticsearch/templates/elasticsearch-logging.yml.j2
```diff
@@ -1,25 +1,14 @@
...
 logger:
   # log action execution errors for easier debugging
   action: WARN
-
-  # deprecation logging, turn to DEBUG to see them
-  deprecation: WARN, deprecation_log_file
-
```
richm: Why get rid of this?

roles/openshift_logging_elasticsearch/templates/elasticsearch-logging.yml.j2
```diff
@@ -1,25 +1,14 @@
...
   com.amazonaws: WARN
-
   io.fabric8.elasticsearch: ${PLUGIN_LOGLEVEL}
   io.fabric8.kubernetes: ${PLUGIN_LOGLEVEL}
 
-  # aws will try to do some sketchy JMX stuff, but its not needed.
-  com.amazonaws.jmx.SdkMBeanRegistrySupport: ERROR
-  com.amazonaws.metrics.AwsSdkMetrics: ERROR
-
-  org.apache.http: INFO
```
richm: Why get rid of these lines?

roles/openshift_logging_elasticsearch/templates/elasticsearch-logging.yml.j2
```diff
@@ -39,14 +28,13 @@ logger:
...
   index.indexing.slowlog: false
-  deprecation: false
 
 appender:
   console:
     type: console
     layout:
       type: consolePattern
-      conversionPattern: "[%d{ISO8601}][%-5p][%-25c] %.10000m%n"
+      conversionPattern: "[%d{ISO8601}][%-5p][%-25c] %m%n"
```
richm: why change this?

roles/openshift_logging_elasticsearch/templates/elasticsearch-logging.yml.j2
```diff
@@ -39,14 +28,13 @@ logger:
 additivity:
   index.search.slowlog: false
   index.indexing.slowlog: false
-  deprecation: false
```
richm: ?

roles/openshift_logging_elasticsearch/templates/elasticsearch-logging.yml.j2
```diff
@@ -56,13 +44,16 @@ appender:
       type: pattern
       conversionPattern: "[%d{ISO8601}][%-5p][%-25c] %m%n"
 
-  deprecation_log_file:
-    type: dailyRollingFile
-    file: ${path.logs}/${cluster.name}_deprecation.log
-    datePattern: "'.'yyyy-MM-dd"
-    layout:
-      type: pattern
-      conversionPattern: "[%d{ISO8601}][%-5p][%-25c] %m%n"
```
richm: Why get rid of this?

roles/openshift_logging_elasticsearch/templates/elasticsearch-logging.yml.j2
```diff
@@ -56,13 +44,16 @@ appender:
...
+  # Use the following log4j-extras RollingFileAppender to enable gzip compression of log files.
+  # For more information see https://logging.apache.org/log4j/extras/apidocs/org/apache/log4j/rolling/RollingFileAppender.html
+  #file:
+    #type: extrasRollingFile
+    #file: ${path.logs}/${cluster.name}.log
+    #rollingPolicy: timeBased
+    #rollingPolicy.FileNamePattern: ${path.logs}/${cluster.name}.log.%d{yyyy-MM-dd}.gz
+    #layout:
+      #type: pattern
+      #conversionPattern: "[%d{ISO8601}][%-5p][%-25c] %m%n"
```
richm: Why add this commented out?

roles/openshift_logging_elasticsearch/defaults/main.yml
```diff
@@ -0,0 +1,57 @@
...
+# pvc options
+# the name of the PVC we will bind to -- create it if it does not exist
+openshift_logging_elasticsearch_pvc_name: ""
+
+# required if the PVC does not already exist
+openshift_logging_elasticsearch_pvc_size: ""
+openshift_logging_elasticsearch_pvc_dynamic: false
+openshift_logging_elasticsearch_pvc_pv_selector: {}
+openshift_logging_elasticsearch_pvc_access_modes: ['ReadWriteOnce']
+openshift_logging_elasticsearch_storage_group: '65534'
```
richm: Many of these parameters were named `openshift_logging_es_<something>` in 3.5.  Is the new code going to honor those old settings?

richm: Looks like the mapping between the old name and the new name is done in `roles/openshift_logging/tasks/install_logging.yaml` e.g.
```
    openshift_logging_elasticsearch_replica_count: "{{ openshift_logging_es_cluster_size | int }}"
```

richm:
@ewolinetz check https://github.com/richm/openshift-ansible/commits/logging_component_subroles - I recently rebased to master, and synced over a lot of changes from openshift_logging/*

What do we need to do to get this PR merged?  The longer it sits, the harder it will be to merge . . .

ewolinetz:
@richm this had dropped in priority for another deliverable for the prior sprint. I will try to pull in the remaining changes from your branch; with those you see that things are working as they currently are? I think once we are ready with that we can merge unless @sdodson has reservations about doing so -- this would simply require a regression test

ewolinetz:
@richm pulled in missing changes from your branch
PTAL

richm:
lgtm @jcantrill ptal

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

sdodson:
Lots of LGTM, may as well get this in.

sdodson:
I'd started trying to fix travis errors but then I started running into items that I'm not sure were intentionally commented out because it didn't look like they were commented out in the original code.

ewolinetz:
@sdodson i didn't notice the travis errors before, i'll try to address those asap

ewolinetz:
aos-ci-test

sdodson:
[merge]

PR #3511, opened 2017-02-27
Run pylint in more files (incl. tests)

### LINKED ISSUE #3500: Some Python files being ignored by `setup.py lint`
While making changes to `roles/lib_openshift/src/test/unit/test_oadm_manage_node.py`, I noticed that it and other files are not being linted.

Seems that the logic in [`setup.py OpenShiftAnsiblePylint.find_all_modules`](https://github.com/openshift/openshift-ansible/blob/641b7c93b1d5ce5388fce66d737704d00a83ec68/setup.py#L128) excludes directories named "test", what might be an oversight.

I believe the test files should be linted as well (they even have `# pylint: disable` comments), @detiber @tbielawa @mtnbikenc @sdodson thoughts?

rhcarvalho:
Fixes #3500

- Several misc. fixes to code style
- Reset pylint config to default + automatically use all CPUs + disable some extra rules
- Remove custom runner code from `setup.py` -- use `pylint`

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -67,6 +67,13 @@ confidence=
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
 disable=
+    # disabling rules is not to be reported as a violation
+    locally-disabled,
+    # some files are to be ignored
+    file-ignored,
+    # it is expected to have TODO notes
+    fixme,
+    duplicate-code
 
 
 [REPORTS]
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,4 +1,5 @@
 [MASTER]
+
 # Specify a configuration file.
 #rcfile=
 
@@ -8,7 +9,11 @@
 
 # Add files or directories to the blacklist. They should be base names, not
 # paths.
-ignore=CVS,setup.py
+ignore=CVS
+
+# Add files or directories matching the regex patterns to the blacklist. The
+# regex matches against base names, not paths.
+ignore-patterns=
 
 # Pickle collected data for later comparisons.
 persistent=no
@@ -35,7 +40,8 @@ extension-pkg-whitelist=
 # be used to obtain the result of joining multiple strings with the addition
 # operator. Joining a lot of strings can lead to a maximum recursion error in
 # Pylint and this flag can prevent that. It has one side effect, the resulting
-# AST will be different than the one from reality.
+# AST will be different than the one from reality. This option is deprecated
+# and it will be removed in Pylint 2.0.
 optimize-ast=no
 
 
@@ -60,7 +66,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=fixme,locally-disabled,file-ignored,duplicate-code
+disable=
 
 
 [REPORTS]
@@ -72,7 +78,8 @@ output-format=parseable
 
 # Put messages in a separate file for each module / package specified on the
 # command line instead of printing them on stdout. Reports (if any) will be
-# written in a file name "pylint_global.[txt|html]".
+# written in a file name "pylint_global.[txt|html]". This option is deprecated
+# and it will be removed in Pylint 2.0.
 files-output=no
 
 # Tells whether to display a full report or only the messages
@@ -90,25 +97,41 @@ evaluation=10.0 - ((float(5 * error + warning + refactor + convention) / stateme
 #msg-template=
 
 
-[SIMILARITIES]
+[MISCELLANEOUS]
 
-# Minimum lines number of a similarity.
-min-similarity-lines=4
+# List of note tags to take in consideration, separated by a comma.
+notes=FIXME,XXX,TODO
 
-# Ignore comments when computing similarities.
-ignore-comments=yes
 
-# Ignore docstrings when computing similarities.
-ignore-docstrings=yes
+[TYPECHECK]
 
-# Ignore imports when computing similarities.
-ignore-imports=yes
+# Tells whether missing members accessed in mixin class should be ignored. A
+# mixin class is detected if its name ends with "mixin" (case insensitive).
+ignore-mixin-members=yes
+
+# List of module names for which member attributes should not be checked
+# (useful for modules/projects where namespaces are manipulated during runtime
+# and thus existing member attributes cannot be deduced by static analysis. It
+# supports qualified module names, as well as Unix pattern matching.
+ignored-modules=
+
+# List of class names for which member attributes should not be checked (useful
+# for classes with dynamically set attributes). This supports the use of
+# qualified names.
+ignored-classes=optparse.Values,thread._local,_thread._local
+
+# List of members which are set dynamically and missed by pylint inference
+# system, and so shouldn't trigger E1101 when accessed. Python regular
+# expressions are accepted.
+generated-members=
+
+# List of decorators that produce context managers, such as
+# contextlib.contextmanager. Add to this list to register other decorators that
+# produce valid context managers.
+contextmanager-decorators=contextlib.contextmanager
 
 
 [BASIC]
-
-# List of builtins function names that should not be used, separated by a comma
-bad-functions=map,filter,input
 
 # Good variable names which should always be accepted, separated by a comma
 good-names=i,j,k,ex,Run,_
@@ -122,6 +145,10 @@ name-group=
 
 # Include a hint for the correct naming format with invalid-name
 include-naming-hint=no
+
+# List of decorators that produce properties, such as abc.abstractproperty. Add
+# to this list to register other decorators that produce valid properties.
+property-classes=abc.abstractproperty
 
 # Regular expression matching correct function names
 function-rgx=[a-z_][a-z0-9_]{2,30}$
@@ -198,44 +225,25 @@ docstring-min-length=-1
 max-nested-blocks=5
 
 
-[MISCELLANEOUS]
+[SIMILARITIES]
 
-# List of note tags to take in consideration, separated by a comma.
-notes=FIXME,XXX,TODO
+# Minimum lines number of a similarity.
+min-similarity-lines=4
 
+# Ignore comments when computing similarities.
+ignore-comments=yes
 
-[TYPECHECK]
+# Ignore docstrings when computing similarities.
+ignore-docstrings=yes
 
-# Tells whether missing members accessed in mixin class should be ignored. A
-# mixin class is detected if its name ends with "mixin" (case insensitive).
-ignore-mixin-members=yes
-
-# List of module names for which member attributes should not be checked
-# (useful for modules/projects where namespaces are manipulated during runtime
-# and thus existing member attributes cannot be deduced by static analysis. It
-# supports qualified module names, as well as Unix pattern matching.
-ignored-modules=
-
-# List of classes names for which member attributes should not be checked
-# (useful for classes with attributes dynamically set). This supports can work
-# with qualified names.
-ignored-classes=
-
-# List of members which are set dynamically and missed by pylint inference
-# system, and so shouldn't trigger E1101 when accessed. Python regular
-# expressions are accepted.
-generated-members=
+# Ignore imports when computing similarities.
+ignore-imports=no
 
 
 [SPELLING]
 
-# Spelling dictionary name. Available dictionaries: en_ZW (myspell), en_NG
-# (myspell), en_NA (myspell), en_NZ (myspell), en_PH (myspell), en_AG
-# (myspell), en_BW (myspell), en_IE (myspell), en_ZM (myspell), en_DK
-# (myspell), en_CA (myspell), en_GH (myspell), en_IN (myspell), en_BZ
-# (myspell), en_MW (myspell), en_TT (myspell), en_JM (myspell), en_GB
-# (myspell), en_ZA (myspell), en_SG (myspell), en_AU (myspell), en_US
-# (myspell), en_BS (myspell), en_HK (myspell).
+# Spelling dictionary name. Available dictionaries: none. To make it working
+# install python-enchant package.
 spelling-dict=
 
 # List of comma separated words that should not be checked.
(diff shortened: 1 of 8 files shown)


tox.ini
```diff
@@ -14,8 +15,6 @@ deps =
 commands =
     unit: pytest {posargs}
     flake8: flake8 {posargs}
-    pylint: python setup.py lint
+    pylint: sh -c "git ls-files -co --exclude-standard '*.py' | grep -Ev '^(utils|roles/lib_(openshift|utils)/src)/' | xargs pylint"
```
detiber: I like that this is more transparent than the setuptools integration, but I'm not sure this is the right answer. It'll skip over any files the user created/moved but hasn't done a `git add` on. That is what led me to the setuptools integration to be able to do better discovery...  Maybe we can leverage pytest for that instead?

.pylintrc
```diff
@@ -60,7 +66,70 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-#disable=
+disable=
+    # These are disabled by default:
```
rhcarvalho: I'm okay with leaving them enabled, for some reason I had the wrong impression that empty list == default.

Reading the [docs](https://pylint.readthedocs.io/en/latest/reference_guide/features.html#messages-control-options) again, I don't know what made me think that way...

Will remove/update.

tox.ini
rhcarvalho: The `-o` includes untracked files.

rhcarvalho: I don't say this is particularly great, open to find improvements.

- I don't like to have to wrap on `sh -c`
- I don't like using `grep` inline like that -- would prefer an easier-to-read list of excludes

One thing I'd like to be able to do is to pass files as positional arguments to limit the scope during development:

```
tox -e py35-ansible22-pylint -- roles/new_role/library/file.py
```

detiber: > One thing I'd like to be able to do is to pass files as positional arguments to limit the scope during development:

Indeed, that is why I was wondering about the pytest-pylint plugin.

rhcarvalho: I had a look, pytest-pylint doesn't seem to help with discovery: https://github.com/carsongee/pytest-pylint/blob/master/pytest_pylint.py#L98-L115, it uses the basename just like pylint itself. And it doesn't handle gitignored files.

.pylintrc
```diff
@@ -60,7 +66,70 @@ confidence=
...
+    map-builtin-not-iterating,
+    # These are disabled in openshift-ansible:
+    # disabling rules is not to be reported as a violation
+    locally-disabled,
+    # some files are to be ignored
+    file-ignored,
+    # it is expected to have TODO notes
+    fixme,
+    # sometimes it's better to have no docstrings than placeholders or incorrect text
+    missing-docstring,
```
rhcarvalho: > I think we should get more input from the larger group

:+1:

Yes, I agree!

---

While disabling was the quick-fix for getting a green run from pytest, I did fix some other issues like long lines, unused arguments.

For docstrings and names, instead of adding several inline disables, I used the reasoning I've documented.

## missing-docstrings

I've seen throughout the repo several docstrings that just state the obvious, apparently just to satisfy pylint, and I remember PRs fixing copypasta incorrect docstrings.

Looking only at files open in my editor:

- `""" Tests for the modify_yaml Ansible module. """`
- `"""A setuptools based setup module.\n\n"""`

I'd say those are not very useful.

I'm in favor of people giving good names to their functions, classes, methods, variables, and documenting what needs to be documented. If that's something we collectively care about, we need to be more attentive to that during code review.

I feel this pylint check is not enough and often forces us to add a `pylint: disable=missing-docstring` (64 occurrences in 45 files) or to add a non-sense docstring. That's just my opinion, I'm okay to keep it enabled if other people prefer it that way. In that case, we'd need to go fix the missing docstrings or add the inline disable comments.

## invalid-name

These are the invalid names when I remove `invalid-name` from the disabled list:

<details>
<summary>$ tox -e py35-ansible22-pylint</summary>

```
py35-ansible22-pylint installed: ansible==2.2.1.0,appdirs==1.4.2,astroid==1.4.9,cffi==1.9.1,coverage==4.3.4,cryptography==1.7.2,flake8==3.3.0,flake8-mutable==1.1.0,flake8-print==2.0.2,idna==2.2,isort==4.2.5,Jinja2==2.8.1,lazy-object-proxy==1.2.2,MarkupSafe==0.23,mccabe==0.6.1,mock==2.0.0,packaging==16.8,paramiko==2.1.2,pbr==1.10.0,py==1.4.32,pyasn1==0.2.3,pycodestyle==2.3.1,pycparser==2.17,pycrypto==2.6.1,pyflakes==1.5.0,pylint==1.6.5,pyOpenSSL==16.2.0,pyparsing==2.1.10,pytest==3.0.6,pytest-cov==2.4.0,PyYAML==3.12,setuptools-lint==0.5.2,six==1.10.0,wrapt==1.10.8,yamllint==1.6.1
py35-ansible22-pylint runtests: PYTHONHASHSEED='1093142714'
py35-ansible22-pylint runtests: commands[0] | sh -c git ls-files -co --exclude-standard '*.py' | grep -Ev '^(utils|roles/lib_(openshift|utils)/src)/' | xargs pylint
************* Module test_fakeopensslclasses
roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py:14: [C0103(invalid-name), ] Invalid constant name "module_path"
roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py:16: [C0103(invalid-name), ] Invalid constant name "openshift_cert_expiry"
roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py:27: [C0103(invalid-name), TestFakeOpenSSLClasses.setUp] Invalid variable name "fp"
roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py:32: [C0103(invalid-name), TestFakeOpenSSLClasses.setUp] Invalid variable name "fp"
roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py:37: [C0103(invalid-name), TestFakeOpenSSLClasses.test_FakeOpenSSLCertificate_get_serial_number] Invalid method name "test_FakeOpenSSLCertificate_get_serial_number"
roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py:41: [C0103(invalid-name), TestFakeOpenSSLClasses.test_FakeOpenSSLCertificate_get_notAfter] Invalid method name "test_FakeOpenSSLCertificate_get_notAfter"
roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py:46: [C0103(invalid-name), TestFakeOpenSSLClasses.test_FakeOpenSSLCertificate_get_sans] Invalid method name "test_FakeOpenSSLCertificate_get_sans"
roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py:61: [C0103(invalid-name), TestFakeOpenSSLClasses.test_FakeOpenSSLCertificate_get_sans_no_sans] Invalid method name "test_FakeOpenSSLCertificate_get_sans_no_sans"
roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py:66: [C0103(invalid-name), TestFakeOpenSSLClasses.test_FakeOpenSSLCertificate_get_subject] Invalid method name "test_FakeOpenSSLCertificate_get_subject"
roles/openshift_certificate_expiry/test/test_fakeopensslclasses.py:76: [C0103(invalid-name), TestFakeOpenSSLClasses.test_FakeOpenSSLCertificate_get_subject_san_cert] Invalid method name "test_FakeOpenSSLCertificate_get_subject_san_cert"
************* Module openshift_master_facts_default_predicates_tests
roles/openshift_master_facts/test/openshift_master_facts_default_predicates_tests.py:163: [C0103(invalid-name), test_short_version_deployment_type_kwargs] Invalid function name "test_short_version_deployment_type_kwargs"
roles/openshift_master_facts/test/openshift_master_facts_default_predicates_tests.py:163: [C0103(invalid-name), test_short_version_deployment_type_kwargs] Invalid argument name "short_version_deployment_type_kwargs_fixture"
roles/openshift_master_facts/test/openshift_master_facts_default_predicates_tests.py:172: [C0103(invalid-name), short_version_deployment_type_kwargs_fixture] Invalid function name "short_version_deployment_type_kwargs_fixture"
************* Module openshift_master_facts_bad_input_tests
roles/openshift_master_facts/test/openshift_master_facts_bad_input_tests.py:33: [C0103(invalid-name), TestOpenShiftMasterFactsBadInput.test_missing_short_version_and_missing_openshift_release] Invalid method name "test_missing_short_version_and_missing_openshift_release"
************* Module openshift_master_facts_default_priorities_tests
roles/openshift_master_facts/test/openshift_master_facts_default_priorities_tests.py:151: [C0103(invalid-name), test_short_version_deployment_type_kwargs] Invalid function name "test_short_version_deployment_type_kwargs"
roles/openshift_master_facts/test/openshift_master_facts_default_priorities_tests.py:151: [C0103(invalid-name), test_short_version_deployment_type_kwargs] Invalid argument name "short_version_deployment_type_kwargs_fixture"
roles/openshift_master_facts/test/openshift_master_facts_default_priorities_tests.py:160: [C0103(invalid-name), short_version_deployment_type_kwargs_fixture] Invalid function name "short_version_deployment_type_kwargs_fixture"
ERROR: InvocationError: "/usr/bin/sh -c git ls-files -co --exclude-standard '*.py' | grep -Ev '^(utils|roles/lib_(openshift|utils)/src)/' | xargs pylint"
___________________________________________________________ summary ___________________________________________________________
ERROR:   py35-ansible22-pylint: commands failed
```
</details>


I don't see value renaming those things?

They'd add to a list of other 40 occurrences of `disable=invalid-name` in 33 files.

rhcarvalho:
Some legitimate lint errors to be addressed in `utils/` -- https://travis-ci.org/openshift/openshift-ansible/jobs/205928475#L923-L954

.pylintrc
detiber: @rhcarvalho I think we should probably discuss these pylint disables at the architecture meeting this week. Personally, I would like to see us leveraging pytest's doctest support: http://doc.pytest.org/en/latest/doctest.html, but we can do that with or without docstring enforcement, it would just make it a bit more 'standard' if we were enforcing docstrings.

One thing that I'm wondering about is if we should have a different set of standards for tests than we do for other python files. I don't see any reason to enforce docstrings, invalid-name, wrong-import-position, etc for tests.

tox.ini
detiber: arrgh, so much for that.

detiber:
@rhcarvalho this one doesn't look legitimate: https://travis-ci.org/openshift/openshift-ansible/jobs/205928475#L940

I'm wondering if we should enforce some `from __future__` imports. I'm thinking `print_function` and `absolute_import` may be a minimum for py2/py3 compatibility.

rhcarvalho:
> @rhcarvalho this one doesn't look legitimate: https://travis-ci.org/openshift/openshift-ansible/jobs/205928475#L940
> 
> I'm wondering if we should enforce some from __future__ imports. I'm thinking print_function and absolute_import may be a minimum for py2/py3 compatibility.

Yes, the fix for that was indeed adding the _\_future\_\_ import... pytest is not super smart. It's a pity the message is the same as it would be for some expression with superfluous parens.

.pylintrc
rhcarvalho: @detiber was this topic discussed in the meeting?

> I'm wondering about is if we should have a different set of standards for tests than we do for other python files

I'd prefer a single standard, as it would be much easier to reason about.

Similarly, I think the repository would be more accessible if there would be a single configuration to run tests / linters / etc and no distinction between `root` and `utils`.

detiber: I agree about root/utils, I'm removing that distinction here: https://github.com/openshift/openshift-ansible/pull/3365, I'll work on breaking that PR up today

detiber: For tests/non-tests, I think it's a bit trickier. There are going to be some common exclusions we are going to see commonly in tests, and I'm not sure it's worth forcing test writers to have to consistently disable them.

Things like import order, invalid-name, and shadowing names in particular will be common. I'm mainly thinking that we could break out pylint invocations into two steps with the one applying to tests adding a handful of additional disables as command line args.

rhcarvalho: @detiber excellent! I have a WIP to simplify a lot of the unit test code to be more lightweight thanks to pytest, I think those we be great for new contributions/contributors!

rhcarvalho:
> Things like import order, invalid-name, and shadowing names in particular will be common.

Hmm

- Most cases of `import-order` I've seen are because of manipulating `sys.path`. There are other ways of handling that (maybe in `conftest.py`, or otherwise). I'd mark those explicitly, and still pay attention to import order for other cases.

- Why would `invalid-name` be valid in tests and not in non-test code?

- Same with shadowing? Only case I've seen is pylint complaining on fixture names that are used as arguments (thus shadowing the function name).


In any case, I think the use cases of splitting are not clear enough yet. I don't oppose to doing it eventually, but we can definitely wait for when the time comes ;)

detiber:
> Most cases of import-order I've seen are because of manipulating sys.path. There are other ways of handling that (maybe in conftest.py, or otherwise). I'd mark those explicitly, and still pay attention to import order for other cases.

There is that, but we'll also hit it when testing code that behaves differently based on dependent packages being available or not: https://github.com/detiber/openshift-ansible/blob/69e35f493e0ee3b30a2548ac72be99fcb16a56ba/roles/lib_openshift/src/test/unit/test_oc_serviceaccount_secret.py#L25-L33

> Why would invalid-name be valid in tests and not in non-test code?

True, I may have been reaching a bit with that one. I do think that tests do not need to be as approachable as the code they are testing. It wouldn't bother me personally to use more descriptive variable names, I just didn't see a point in forcing it on all contributors (for tests).

> Same with shadowing? Only case I've seen is pylint complaining on fixture names that are used as arguments (thus shadowing the function name).

Yes, which is going to be a common thing we will hit, since we are using pytest :)  

> In any case, I think the use cases of splitting are not clear enough yet. I don't oppose to doing it eventually, but we can definitely wait for when the time comes ;)

To be clear, I am not suggesting that we use two distinct set of rules, where one applies to tests and the other to code, but rather just excluding a small subset of the checks when applying the rules to tests.

rhcarvalho:
Rebased to resolve old conflicts. This would still require a considerable amount of work to fix the pylint issues in test files.

rhcarvalho:
No time to work on this, low priority, closing.

PR #3514, opened 2017-02-27
Do not attempt to override openstack nodename

detiber:


detiber:
@cooktheryan @jianlinliu I don't currently have an environment to test against to validate these changes.

detiber:
aos-ci-test

cooktheryan:
@markllama @tomassedovic wanted to get your guys input on this

detiber:
@sdodson I'm hoping that it reverts enough. I mainly do not want to invalidate testing that has already happened with other platforms using nodename :)

detiber:
aos-ci-test

detiber:
@sdodson this should be good to go now. I only created this against release-1.5, since I wasn't sure if we wanted to try to solve this better for 1.6/3.6.

sdodson:
Yeah, I think it's best if we could come up with a better solution for 1.6.

PR #3515, opened 2017-02-27
BZ1427009: Correct fact creation for pvc

mtnbikenc:
Fixes 1427009

### DIFF
--- a/roles/openshift_hosted/tasks/registry/registry.yml
+++ b/roles/openshift_hosted/tasks/registry/registry.yml
@@ -64,13 +64,14 @@
   when:
   - openshift.hosted.registry.storage.kind | default(none) == 'object'
 
-- name: Set facts for persistent volume
+- name: Update openshift_hosted facts for persistent volumes
   set_fact:
+    openshift_hosted_registry_volumes: "{{ openshift_hosted_registry_volumes | union(pvc_volume_mounts) }}"
+  vars:
     pvc_volume_mounts:
     - name: registry-storage
       type: persistentVolumeClaim
       claim_name: "{{ openshift.hosted.registry.storage.volume.name }}-claim"
-    openshift_hosted_registry_volumes: "{{ openshift_hosted_registry_volumes | union(pvc_volume_mounts) }}"
   when:
   - openshift.hosted.registry.storage.kind | default(none) in ['nfs', 'openstack']
 


roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -64,13 +64,19 @@
...
+  - name: Set facts for persistent volume
+    set_fact:
+      pvc_volume_mounts:
+      - name: registry-storage
+        type: persistentVolumeClaim
+        claim_name: "{{ openshift.hosted.registry.storage.volume.name }}-claim"
+
+  - name: Update openshift_hosted facts for persistent volumes
+    set_fact:
+      openshift_hosted_registry_volumes: "{{ openshift_hosted_registry_volumes | union(pvc_volume_mounts) }}"
```
detiber: I think we should consider refactoring this down the line to avoid needing to use set_fact in the way that we are, but I do think this could be reduced down to a single set_fact instead of two separate set_facts:

```
- name: Update openshift_hosted facts for persistent volumes
  set_fact:
    openshift_hosted_registry_volumes: "{{ openshift_hosted_registry_volumes | union(pvc_volume_mounts) }}"
  vars:
    pvc_volume_mounts:
    - name: registry-storage
      type: persistentVolumeClaim
      claim_name: "{{ openshift.hosted.registry.storage.volume.name }}-claim"
```

mtnbikenc:
aos-ci-test

mtnbikenc:
@detiber PTAL

mtnbikenc:
@kwoodson One thing I just noticed is that the volume class checks for type `pvc` instead of the full type as defined in the `oc` spec, `emptyDir, hostPath, secret, configmap, persistentVolumeClaim`.  Given that I'm using `persistentVolumeClaim`, I'm thinking this will still fail.  It seems the module should use the full name vice just `pvc`.  Thoughts?

roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -64,13 +64,14 @@
...
   - openshift.hosted.registry.storage.kind | default(none) == 'object'
 
-- name: Set facts for persistent volume
+- name: Update openshift_hosted facts for persistent volumes
   set_fact:
+    openshift_hosted_registry_volumes: "{{ openshift_hosted_registry_volumes | union(pvc_volume_mounts) }}"
+  vars:
     pvc_volume_mounts:
     - name: registry-storage
       type: persistentVolumeClaim
```
mtnbikenc: @kwoodson `persistentVolumeClaim` or `pvc`?

kwoodson: @mtnbikenc The interface can be as we want.  Currently it expects type to be pvc `        elif volume_info['type'] == 'pvc':`.  We would just need to add support for that.

For this to work, it needs to be pvc currently.   I'm fine if we want to add support for both.

PR #3517, opened 2017-02-27
Adding changed_whens for role, rolebinding, and scc reconciliation ba…

ewolinetz:
…sed on output from oadm policy command

Encompasses https://trello.com/c/5Yb3cOk1/372-2-3-idempotency-reconcile-role-s-bindings
@tbielawa @detiber

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -176,8 +176,8 @@
       policy reconcile-cluster-roles --additive-only=true --confirm -o name
     register: reconcile_cluster_role_result
     changed_when:
-      - reconcile_cluster_role_result.stdout.length > 0
-      - reconcile_cluster_role_result.rc == 0
+    - reconcile_cluster_role_result.stdout.length > 0
+    - reconcile_cluster_role_result.rc == 0
     run_once: true
 
   - name: Reconcile Cluster Role Bindings
@@ -192,8 +192,8 @@
     when: origin_reconcile_bindings | bool or ent_reconcile_bindings | bool
     register: reconcile_bindings_result
     changed_when:
-      - reconcile_bindings_result.stdout.length > 0
-      - reconcile_bindings_result.rc == 0
+    - reconcile_bindings_result.stdout.length > 0
+    - reconcile_bindings_result.rc == 0
     run_once: true
 
   - name: Reconcile Jenkins Pipeline Role Bindings
@@ -202,8 +202,8 @@
     run_once: true
     register: reconcile_jenkens_role_binding_result
     changed_when:
-      - reconcile_jenkins_role_binding_result.stdout.length > 0
-      - reconcile_jenkins_role_binding_result.rc == 0
+    - reconcile_jenkins_role_binding_result.stdout.length > 0
+    - reconcile_jenkins_role_binding_result.rc == 0
     when: openshift.common.version_gte_3_4_or_1_4  | bool
 
   - name: Reconcile Security Context Constraints
@@ -211,8 +211,8 @@
       {{ openshift.common.client_binary }} adm policy reconcile-sccs --confirm --additive-only=true -o name
     register: reconcile_scc_result
     changed_when:
-      - reconcile_scc_result.stdout.length > 0
-      - reconcile_scc_result.rc == 0
+    - reconcile_scc_result.stdout.length > 0
+    - reconcile_scc_result.rc == 0
     run_once: true
 
   - set_fact:
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
@@ -173,7 +173,11 @@
   - name: Reconcile Cluster Roles
     command: >
       {{ openshift.common.client_binary }} adm --config={{ openshift.common.config_base }}/master/admin.kubeconfig
-      policy reconcile-cluster-roles --additive-only=true --confirm
+      policy reconcile-cluster-roles --additive-only=true --confirm -o name
+    register: reconcile_cluster_role_result
+    changed_when:
+      - reconcile_cluster_role_result.stdout.length > 0
+      - reconcile_cluster_role_result.rc == 0
     run_once: true
 
   - name: Reconcile Cluster Role Bindings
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -173,7 +173,9 @@
   - name: Reconcile Cluster Roles
     command: >
       {{ openshift.common.client_binary }} adm --config={{ openshift.common.config_base }}/master/admin.kubeconfig
-      policy reconcile-cluster-roles --additive-only=true --confirm
+      policy reconcile-cluster-roles --additive-only=true --confirm -o name
+    register: reconcile_cluster_role_result
+    changed_when: reconcile_cluster_role_result.stdout.length > 0
```
tbielawa: Missing part of the acceptance criteria

* The `changed_when` parameter is added to each task such that it is only `true` when there is **output from the task** and the **return code is `0`**

For example:

```yaml
   - name: Reconcile Cluster Roles
      command: >
        {{ openshift.common.client_binary }} adm --config={{ openshift.common.config_base }}/master/admin.kubeconfig
      policy reconcile-cluster-roles --additive-only=true --confirm
      policy reconcile-cluster-roles --additive-only=true --confirm -o name
    register: reconcile_cluster_role_result
    changed_when: 
    - reconcile_cluster_role_result.stdout.length > 0
    - reconcile_cluster_role_result.rc == 0
```

would be more like what we're looking for

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -184,19 +186,25 @@
       --exclude-groups=system:authenticated:oauth
       --exclude-groups=system:unauthenticated
       --exclude-users=system:anonymous
-      --additive-only=true --confirm
+      --additive-only=true --confirm -o name
     when: origin_reconcile_bindings | bool or ent_reconcile_bindings | bool
+    register: reconcile_bindings_result
+    change_when: reconcile_bindings_result.stdout.length > 0
```
tbielawa: same as above

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -184,19 +186,25 @@
...
     run_once: true
 
   - name: Reconcile Jenkins Pipeline Role Bindings
     command: >
-      {{ openshift.common.client_binary }} adm --config={{ openshift.common.config_base }}/master/admin.kubeconfig policy reconcile-cluster-role-bindings system:build-strategy-jenkinspipeline --confirm
+      {{ openshift.common.client_binary }} adm --config={{ openshift.common.config_base }}/master/admin.kubeconfig policy reconcile-cluster-role-bindings system:build-strategy-jenkinspipeline --confirm -o name
     run_once: true
+    register: reconcile_jenkens_role_binding_result
+    changed_when: reconcile_jenkins_role_binding_result.stdout.length > 0
```
tbielawa: same

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -184,19 +186,25 @@
...
     when: openshift.common.version_gte_3_4_or_1_4  | bool
 
   - name: Reconcile Security Context Constraints
     command: >
-      {{ openshift.common.client_binary }} adm policy reconcile-sccs --confirm --additive-only=true
+      {{ openshift.common.client_binary }} adm policy reconcile-sccs --confirm --additive-only=true -o name
+    register: reconcile_scc_result
+    changed_when: reconcile_scc_result.stdout.length > 0
```
tbielawa: same

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
ewolinetz: I must have missed that in the AC, will update

ewolinetz:
@tbielawa updated `changed_when` to also verify that the `rc` == 0

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
detiber: Shouldn't the return value impact failed_when not changed_when? I guess I'm not seeing the reasoning for comparing the rc for changed_when in this case.

ewolinetz: We want to also check the rc for the `changed_when` because `-o name` will tell us what objects need a change, not whether or not they did change (per IRC discussion with liggit).

I would agree that we would also want to have a `failed_when` that examined the `rc != 0`

detiber: Ah, that makes sense.

ewolinetz:
reverted `failed_when` commit

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

tbielawa:
I'm running two tests from this branch. A first test to get a base-line for the timing and number of changes. A second test to verify the results from the first. Hopefully should be done in about 25-30 minutes once I examine the ARA report.

tbielawa:
Test run fails for me @ewolinetz 

```
TASK [openshift_cli : Install bash completion for oc tools] ********************
ok: [m01.example.com] => {"changed": false, "msg": "", "rc": 0, "results": ["bash-completion-1:2.1-6.el7.noarch providing bash-completion is already installed"]}

TASK [Reconcile Cluster Roles] *************************************************
fatal: [m01.example.com]: FAILED! => {"failed": true, "msg": "The conditional check 'reconcile_cluster_role_result.stdout.length > 0' failed. The error was: error while evaluating conditional (reconcile_cluster_role_result.stdout.length > 0): 'ansible.vars.unsafe_proxy.AnsibleUnsafeText object' has no attribute 'length'"}

localhost                  : ok=34   changed=0    unreachable=0    failed=0
m01.example.com            : ok=245  changed=10   unreachable=0    failed=1
n01.example.com            : ok=100  changed=1    unreachable=0    failed=0
n02.example.com            : ok=100  changed=1    unreachable=0    failed=0
```

Maybe try with `reconcile_cluster_role_result.stdout.length != ''` ?

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -192,7 +192,7 @@
     when: origin_reconcile_bindings | bool or ent_reconcile_bindings | bool
     register: reconcile_bindings_result
     changed_when:
-    - reconcile_bindings_result.stdout.length > 0
+    - reconcile_bindings_result.stdout != ''
```
detiber: Just a bit, but I think the default filter might be better here.

detiber: Or maybe even just dropping the `== ''`...

tbielawa:
We talked about this on IRC, just noting it here for transparency:

```
TASK [Reconcile Jenkins Pipeline Role Bindings] ********************************
fatal: [m01.example.com]: FAILED! => {"failed": true, "msg": "The conditional check 'reconcile_jenkins_role_binding_result.stdout != ''' failed. The error was: error while evaluating conditional (reconcile_jenkins
_role_binding_result.stdout != ''): 'reconcile_jenkins_role_binding_result' is undefined"}
```

tbielawa:
Restesting w/ updates

ewolinetz:
aos-ci-test

tbielawa:
Tests are quite satisfying. 

change-result-comparisons

Those 9 result rows were (in order) ran against the following branches

* This PR
* This PR
* Master Branch
* Master Branch
* This PR
* This PR
* Master Branch
* Master Branch
* This PR

Examining the changed tasks in the PR vs. Master branch results side-by-side we can see that the tasks we fixed in this PR no longer appear as changing in the PR branch

tasks-changing-comparison

Confirming this PR reduces `changed` by 5 vs the master branch @ `openshift-ansible-3.5.3-1-466-g68aa6f2`.

:+1:

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
ewolinetz: we would need some way to verify whether or not the output is empty which means that there weren't any roles to update. I'm not sure using default or dropping the comparison would allow us to still accomplish this...

detiber: @ewolinetz I meant something like this:
```
changed_when:
- reconcile_bindings_results.stdout
```
Then empty string would evaluate to False

or if it is possible that stdout is not present:
```
changed_when:
- reconcile_bindings_result.stdout | default(false, boolean=true)
```

ewolinetz: ack, that's pretty cool; i didn't realize that we could do that. So it would be something like 
```
changed_when:
- not reconcile_bindings_results.stdout
```

However I think this makes the check non-obvious for people that don't realize you can do that. It would probably be better if we had a way to test like 
```
changed_when:
- not reconcile_bindings_result.stdout | empty
```

tbielawa: * There are many different ways to ~shave this yack~ write these `when` conditions
* I don't care which one we use but we need to make a decision
* I personally prefer the last one @detiber posted, it handles a lot of edge cases

```yaml
changed_when:
- reconcile_bindings_result.stdout | default(false, boolean=true)
```

If `stdout` is empty for some reason then the default is `false`. Otherwise the result is `true`.
* `stdout == ''` → **Nothing changed** → `default` filter says *no input value, so this is `false`*
* `stdout != ''` → **Something changed changed** → `default` filter says *input value present, so this is `true`*


@detiber @ewolinetz if that translation is correct ^, let's just do that and move on with this the way @detiber wrote it up.

ewolinetz:
I didn't get a chance to work on a test plugin for checking std{out,err}; so we can merge this now and i'll take it on as tech debt

PR #3518, opened 2017-02-27
add ram and non-containerized storage preflight check

juanvallejo:
Related Trello card: https://trello.com/c/cLu85dOf/71-uhg-diagnostics-check-memory-matches-requirements?menu=filter&filter=[diagnostics]

Related Trello card: https://trello.com/c/5pr2WcQH/43-diagnostics-additional-pre-flight-checks?menu=filter&filter=[diagnostics]

cc @brenton @rhcarvalho

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/disk_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/disk_availability.py
@@ -1,0 +1,73 @@
+# pylint: disable=missing-docstring
+from openshift_checks import OpenShiftCheck, OpenShiftCheckException, get_var
+from openshift_checks.mixins import NotContainerizedMixin
+
+
+class DiskAvailability(NotContainerizedMixin, OpenShiftCheck):
+    """Check that recommended disk space is available before a first-time install."""
+
+    name = "disk_availability"
+    tags = ["preflight"]
+
+    # all values are base-10 as they are taken, as is, from
+    # the latest requirements for an OpenShift installation
+    # https://docs.openshift.org/latest/install_config/install/prerequisites.html#system-requirements
+    recommended_diskspace = {
+        "nodes": 15 * 10 ** 9,
+        "masters": 40 * 10 ** 9,
+        "etcd": 20 * 10 ** 9,
+    }
+
+    def run(self, tmp, task_vars):
+        ansible_mounts = get_var(task_vars, "ansible_mounts")
+        self.recommended_diskspace["nodes"] = get_var(task_vars,
+                                                      "min_recommended_diskspace_node",
+                                                      default=self.recommended_diskspace["nodes"])
+        self.recommended_diskspace["masters"] = get_var(task_vars,
+                                                        "min_recommended_diskspace_master",
+                                                        default=self.recommended_diskspace["masters"])
+        self.recommended_diskspace["etcd"] = get_var(task_vars,
+                                                     "min_recommended_diskspace_etcd",
+                                                     default=self.recommended_diskspace["etcd"])
+
+        failed, msg = self.volume_check(ansible_mounts, task_vars)
+        return {"failed": failed, "msg": msg}
+
+    def volume_check(self, ansible_mounts, task_vars):
+        group_names = get_var(task_vars, "group_names", default=[])
+
+        if not set(self.recommended_diskspace).intersection(group_names):
+            msg = "Unable to determine recommended volume size for group_name {group_name}"
+            raise OpenShiftCheckException(msg.format(group_name=group_names))
+
+        recommended_diskspace_bytes = max(self.recommended_diskspace.get(group, 0) for group in group_names)
+        openshift_diskfree_bytes = self.get_openshift_disk_availability(ansible_mounts)
+
+        if openshift_diskfree_bytes < recommended_diskspace_bytes:
+            msg = ("Available disk space ({diskfree} GB) for the volume containing \"/var\" is "
+                   "below recommended storage. Minimum required disk space: {recommended} GB")
+            return True, msg.format(diskfree=self.to_gigabytes(openshift_diskfree_bytes),
+                                    recommended=self.to_gigabytes(recommended_diskspace_bytes))
+
+        return False, ""
+
+    @staticmethod
+    def get_openshift_disk_availability(ansible_mounts):
+        """Iterates through a map of mounted volumes to determine space remaining on the OpenShift volume"""
+        if not ansible_mounts:
+            msg = "Unable to determine existing volume mounts from ansible_mounts"
+            raise OpenShiftCheckException(msg)
+
+        # priority list in descending order
+        supported_mnt_paths = ["/var", "/"]
+        available_mnts = {mnt.get("mount"): mnt for mnt in ansible_mounts}
+
+        for path in supported_mnt_paths:
+            if path in available_mnts:
+                return available_mnts[path].get("size_available")
+
+        return 0
+
+    @staticmethod
+    def to_gigabytes(total_bytes):
+        return total_bytes / 10**9
--- a/roles/openshift_health_checker/openshift_checks/disk_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/disk_availability.py
@@ -9,65 +9,57 @@ class DiskAvailability(NotContainerizedMixin, OpenShiftCheck):
     name = "disk_availability"
     tags = ["preflight"]
 
-    # all values are base-10 as they are taken, as is, from
-    # the latest requirements for an OpenShift installation
+    # Values taken from the official installation documentation:
     # https://docs.openshift.org/latest/install_config/install/prerequisites.html#system-requirements
-    recommended_diskspace = {
-        "nodes": 15 * 10 ** 9,
-        "masters": 40 * 10 ** 9,
-        "etcd": 20 * 10 ** 9,
+    recommended_disk_space_bytes = {
+        "masters": 40 * 10**9,
+        "nodes": 15 * 10**9,
+        "etcd": 20 * 10**9,
     }
 
-    def run(self, tmp, task_vars):
-        ansible_mounts = get_var(task_vars, "ansible_mounts")
-        self.recommended_diskspace["nodes"] = get_var(task_vars,
-                                                      "min_recommended_diskspace_node",
-                                                      default=self.recommended_diskspace["nodes"])
-        self.recommended_diskspace["masters"] = get_var(task_vars,
-                                                        "min_recommended_diskspace_master",
-                                                        default=self.recommended_diskspace["masters"])
-        self.recommended_diskspace["etcd"] = get_var(task_vars,
-                                                     "min_recommended_diskspace_etcd",
-                                                     default=self.recommended_diskspace["etcd"])
-
-        failed, msg = self.volume_check(ansible_mounts, task_vars)
-        return {"failed": failed, "msg": msg}
-
-    def volume_check(self, ansible_mounts, task_vars):
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip hosts that do not have recommended disk space requirements."""
         group_names = get_var(task_vars, "group_names", default=[])
+        has_disk_space_recommendation = bool(set(group_names).intersection(cls.recommended_disk_space_bytes))
+        return super(DiskAvailability, cls).is_active(task_vars) and has_disk_space_recommendation
 
-        if not set(self.recommended_diskspace).intersection(group_names):
-            msg = "Unable to determine recommended volume size for group_name {group_name}"
-            raise OpenShiftCheckException(msg.format(group_name=group_names))
+    def run(self, tmp, task_vars):
+        group_names = get_var(task_vars, "group_names")
+        ansible_mounts = get_var(task_vars, "ansible_mounts")
 
-        recommended_diskspace_bytes = max(self.recommended_diskspace.get(group, 0) for group in group_names)
-        openshift_diskfree_bytes = self.get_openshift_disk_availability(ansible_mounts)
+        min_free_bytes = max(self.recommended_disk_space_bytes.get(name, 0) for name in group_names)
+        free_bytes = self.openshift_available_disk(ansible_mounts)
 
-        if openshift_diskfree_bytes < recommended_diskspace_bytes:
-            msg = ("Available disk space ({diskfree} GB) for the volume containing \"/var\" is "
-                   "below recommended storage. Minimum required disk space: {recommended} GB")
-            return True, msg.format(diskfree=self.to_gigabytes(openshift_diskfree_bytes),
-                                    recommended=self.to_gigabytes(recommended_diskspace_bytes))
+        if free_bytes < min_free_bytes:
+            return {
+                'failed': True,
+                'msg': (
+                    'Available disk space ({:.1f} GB) for the volume containing '
+                    '"/var" is below minimum recommended space ({:.1f} GB)'
+                ).format(float(free_bytes) / 10**9, float(min_free_bytes) / 10**9)
+            }
 
-        return False, ""
+        return {}
 
     @staticmethod
-    def get_openshift_disk_availability(ansible_mounts):
-        """Iterates through a map of mounted volumes to determine space remaining on the OpenShift volume"""
-        if not ansible_mounts:
-            msg = "Unable to determine existing volume mounts from ansible_mounts"
-            raise OpenShiftCheckException(msg)
+    def openshift_available_disk(ansible_mounts):
+        """Determine the available disk space for an OpenShift installation.
 
+        ansible_mounts should be a list of dicts like the 'setup' Ansible module
+        returns.
+        """
         # priority list in descending order
         supported_mnt_paths = ["/var", "/"]
         available_mnts = {mnt.get("mount"): mnt for mnt in ansible_mounts}
 
-        for path in supported_mnt_paths:
-            if path in available_mnts:
-                return available_mnts[path].get("size_available")
+        try:
+            for path in supported_mnt_paths:
+                if path in available_mnts:
+                    return available_mnts[path]["size_available"]
+        except KeyError:
+            pass
 
-        return 0
-
-    @staticmethod
-    def to_gigabytes(total_bytes):
-        return total_bytes / 10**9
+        paths = ''.join(sorted(available_mnts)) or 'none'
+        msg = "Unable to determine available disk space. Paths mounted: {}.".format(paths)
+        raise OpenShiftCheckException(msg)
--- a/roles/openshift_health_checker/openshift_checks/memory_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/memory_availability.py
@@ -1,0 +1,35 @@
+# pylint: disable=missing-docstring
+from openshift_checks import OpenShiftCheck, get_var
+
+
+class MemoryAvailability(OpenShiftCheck):
+    """Check that recommended memory is available."""
+
+    name = "memory_availability"
+    tags = ["preflight"]
+
+    recommended_memory_mb = {
+        "nodes": 8 * 1000,
+        "masters": 16 * 1000,
+        "etcd": 10 * 1000,
+    }
+
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip hosts that do not have recommended memory space requirements."""
+        group_names = get_var(task_vars, "group_names", default=[])
+        has_mem_space_recommendation = bool(set(group_names).intersection(cls.recommended_memory_mb))
+        return super(MemoryAvailability, cls).is_active(task_vars) and has_mem_space_recommendation
+
+    def run(self, tmp, task_vars):
+        group_names = get_var(task_vars, "group_names", default=[])
+        total_memory = get_var(task_vars, "ansible_memtotal_mb")
+
+        recommended_memory_mb = max(self.recommended_memory_mb.get(group, 0) for group in group_names)
+
+        if total_memory < recommended_memory_mb:
+            msg = ("Available memory ({available} MB) below recommended storage. "
+                   "Minimum required memory: {recommended} MB")
+            return {"failed": True, "msg": msg.format(available=total_memory, recommended=recommended_memory_mb)}
+
+        return {"changed": False}
--- a/roles/openshift_health_checker/openshift_checks/memory_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/memory_availability.py
@@ -1,5 +1,5 @@
 # pylint: disable=missing-docstring
-from openshift_checks import OpenShiftCheck, OpenShiftCheckException, get_var
+from openshift_checks import OpenShiftCheck, get_var
 
 
 class MemoryAvailability(OpenShiftCheck):
@@ -8,31 +8,37 @@ class MemoryAvailability(OpenShiftCheck):
     name = "memory_availability"
     tags = ["preflight"]
 
-    recommended_memory_mb = {
-        "nodes": 8 * 1000,
-        "masters": 16 * 1000,
-        "etcd": 10 * 1000,
+    # Values taken from the official installation documentation:
+    # https://docs.openshift.org/latest/install_config/install/prerequisites.html#system-requirements
+    recommended_memory_bytes = {
+        "masters": 16 * 10**9,
+        "nodes": 8 * 10**9,
+        "etcd": 20 * 10**9,
     }
 
     @classmethod
     def is_active(cls, task_vars):
-        """Skip hosts that do not have recommended memory space requirements."""
+        """Skip hosts that do not have recommended memory requirements."""
         group_names = get_var(task_vars, "group_names", default=[])
-        has_mem_space_recommendation = bool(set(group_names).intersection(cls.recommended_memory_mb))
-        return super(MemoryAvailability, cls).is_active(task_vars) and has_mem_space_recommendation
+        has_memory_recommendation = bool(set(group_names).intersection(cls.recommended_memory_bytes))
+        return super(MemoryAvailability, cls).is_active(task_vars) and has_memory_recommendation
 
     def run(self, tmp, task_vars):
-        group_names = get_var(task_vars, "group_names", default=[])
-        total_memory = get_var(task_vars, "ansible_memtotal_mb")
+        group_names = get_var(task_vars, "group_names")
+        total_memory_bytes = get_var(task_vars, "ansible_memtotal_mb") * 10**6
 
-        recommended_memory_mb = max(self.recommended_memory_mb.get(group, 0) for group in group_names)
-        if not recommended_memory_mb:
-            msg = "Unable to determine recommended memory size for group_name {group_name}"
-            raise OpenShiftCheckException(msg.format(group_name=group_names))
+        min_memory_bytes = max(self.recommended_memory_bytes.get(name, 0) for name in group_names)
 
-        if total_memory < recommended_memory_mb:
-            msg = ("Available memory ({available} MB) below recommended storage. "
-                   "Minimum required memory: {recommended} MB")
-            return {"failed": True, "msg": msg.format(available=total_memory, recommended=recommended_memory_mb)}
+        if total_memory_bytes < min_memory_bytes:
+            return {
+                'failed': True,
+                'msg': (
+                    'Available memory ({available:.1f} GB) '
+                    'below recommended value ({recommended:.1f} GB)'
+                ).format(
+                    available=float(total_memory_bytes) / 10**9,
+                    recommended=float(min_memory_bytes) / 10**9,
+                ),
+            }
 
-        return {"changed": False}
+        return {}
--- a/roles/openshift_health_checker/openshift_checks/memory_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/memory_availability.py
@@ -1,5 +1,5 @@
 # pylint: disable=missing-docstring
-from openshift_checks import OpenShiftCheck, get_var
+from openshift_checks import OpenShiftCheck, OpenShiftCheckException, get_var
 
 
 class MemoryAvailability(OpenShiftCheck):
@@ -26,6 +26,9 @@ class MemoryAvailability(OpenShiftCheck):
         total_memory = get_var(task_vars, "ansible_memtotal_mb")
 
         recommended_memory_mb = max(self.recommended_memory_mb.get(group, 0) for group in group_names)
+        if not recommended_memory_mb:
+            msg = "Unable to determine recommended memory size for group_name {group_name}"
+            raise OpenShiftCheckException(msg.format(group_name=group_names))
 
         if total_memory < recommended_memory_mb:
             msg = ("Available memory ({available} MB) below recommended storage. "
--- a/roles/openshift_health_checker/test/memory_availability_test.py
+++ b/roles/openshift_health_checker/test/memory_availability_test.py
@@ -1,0 +1,84 @@
+import pytest
+
+from openshift_checks.memory_availability import MemoryAvailability, OpenShiftCheckException
+
+
+@pytest.mark.parametrize('group_names,is_containerized,is_active', [
+    (['masters'], False, True),
+    # ensure check is skipped on containerized installs
+    (['masters'], True, True),
+    (['nodes'], True, True),
+    (['etcd'], False, True),
+    (['masters', 'nodes'], False, True),
+    (['masters', 'etcd'], False, True),
+    ([], False, False),
+    (['lb'], False, False),
+    (['nfs'], False, False),
+])
+def test_is_active(group_names, is_containerized, is_active):
+    task_vars = dict(
+        group_names=group_names,
+        openshift=dict(common=dict(is_containerized=is_containerized)),
+    )
+    assert MemoryAvailability.is_active(task_vars=task_vars) == is_active
+
+
+@pytest.mark.parametrize("group_name,size_available", [
+    (
+        "masters",
+        17200,
+    ),
+    (
+        "nodes",
+        8200,
+    ),
+    (
+        "etcd",
+        12200,
+    ),
+])
+def test_mem_check_with_recommended_memtotal(group_name, size_available):
+    result = MemoryAvailability(execute_module=NotImplementedError).run(tmp=None, task_vars=dict(
+        group_names=[group_name],
+        ansible_memtotal_mb=size_available,
+    ))
+
+    assert not result.get('failed', False)
+
+
+@pytest.mark.parametrize("group_name,size_available", [
+    (
+        "masters",
+        1,
+    ),
+    (
+        "nodes",
+        2,
+    ),
+    (
+        "etcd",
+        3,
+    ),
+])
+def test_mem_check_with_insufficient_memtotal(group_name, size_available):
+    result = MemoryAvailability(execute_module=NotImplementedError).run(tmp=None, task_vars=dict(
+        group_names=[group_name],
+        ansible_memtotal_mb=size_available,
+    ))
+
+    assert result['failed']
+    assert "below recommended storage" in result['msg']
+
+
+def test_mem_check_with_invalid_groupname():
+    with pytest.raises(OpenShiftCheckException) as excinfo:
+        result = MemoryAvailability(execute_module=NotImplementedError).run(tmp=None, task_vars=dict(
+            openshift=dict(common=dict(
+                service_type='origin',
+                is_containerized=False,
+            )),
+            group_names=["invalid"],
+            ansible_memtotal_mb=1234567,
+        ))
+
+    assert "'invalid'" in str(excinfo.value)
--- a/roles/openshift_health_checker/test/memory_availability_test.py
+++ b/roles/openshift_health_checker/test/memory_availability_test.py
@@ -1,84 +1,91 @@
 import pytest
 
-from openshift_checks.memory_availability import MemoryAvailability, OpenShiftCheckException
+from openshift_checks.memory_availability import MemoryAvailability
 
 
-@pytest.mark.parametrize('group_names,is_containerized,is_active', [
-    (['masters'], False, True),
-    # ensure check is skipped on containerized installs
-    (['masters'], True, True),
-    (['nodes'], True, True),
-    (['etcd'], False, True),
-    (['masters', 'nodes'], False, True),
-    (['masters', 'etcd'], False, True),
-    ([], False, False),
-    (['lb'], False, False),
-    (['nfs'], False, False),
+@pytest.mark.parametrize('group_names,is_active', [
+    (['masters'], True),
+    (['nodes'], True),
+    (['etcd'], True),
+    (['masters', 'nodes'], True),
+    (['masters', 'etcd'], True),
+    ([], False),
+    (['lb'], False),
+    (['nfs'], False),
 ])
-def test_is_active(group_names, is_containerized, is_active):
+def test_is_active(group_names, is_active):
     task_vars = dict(
         group_names=group_names,
-        openshift=dict(common=dict(is_containerized=is_containerized)),
     )
     assert MemoryAvailability.is_active(task_vars=task_vars) == is_active
 
 
-@pytest.mark.parametrize("group_name,size_available", [
+@pytest.mark.parametrize('group_names,ansible_memtotal_mb', [
     (
-        "masters",
+        ['masters'],
         17200,
     ),
     (
-        "nodes",
+        ['nodes'],
         8200,
     ),
     (
-        "etcd",
-        12200,
+        ['etcd'],
+        22200,
+    ),
+    (
+        ['masters', 'nodes'],
+        17000,
     ),
 ])
-def test_mem_check_with_recommended_memtotal(group_name, size_available):
-    result = MemoryAvailability(execute_module=NotImplementedError).run(tmp=None, task_vars=dict(
-        group_names=[group_name],
-        ansible_memtotal_mb=size_available,
-    ))
+def test_succeeds_with_recommended_memory(group_names, ansible_memtotal_mb):
+    task_vars = dict(
+        group_names=group_names,
+        ansible_memtotal_mb=ansible_memtotal_mb,
+    )
+
+    check = MemoryAvailability(execute_module=fake_execute_module)
+    result = check.run(tmp=None, task_vars=task_vars)
 
     assert not result.get('failed', False)
 
 
-@pytest.mark.parametrize("group_name,size_available", [
+@pytest.mark.parametrize('group_names,ansible_memtotal_mb,extra_words', [
     (
-        "masters",
-        1,
+        ['masters'],
+        0,
+        ['0.0 GB'],
     ),
     (
-        "nodes",
-        2,
+        ['nodes'],
+        100,
+        ['0.1 GB'],
     ),
     (
-        "etcd",
-        3,
+        ['etcd'],
+        -1,
+        ['0.0 GB'],
+    ),
+    (
+        ['nodes', 'masters'],
+        # enough memory for a node, not enough for a master
+        11000,
+        ['11.0 GB'],
     ),
 ])
-def test_mem_check_with_insufficient_memtotal(group_name, size_available):
-    result = MemoryAvailability(execute_module=NotImplementedError).run(tmp=None, task_vars=dict(
-        group_names=[group_name],
-        ansible_memtotal_mb=size_available,
-    ))
+def test_fails_with_insufficient_memory(group_names, ansible_memtotal_mb, extra_words):
+    task_vars = dict(
+        group_names=group_names,
+        ansible_memtotal_mb=ansible_memtotal_mb,
+    )
+
+    check = MemoryAvailability(execute_module=fake_execute_module)
+    result = check.run(tmp=None, task_vars=task_vars)
 
     assert result['failed']
-    assert "below recommended storage" in result['msg']
+    for word in 'below recommended'.split() + extra_words:
+        assert word in result['msg']
 
 
-def test_mem_check_with_invalid_groupname():
-    with pytest.raises(OpenShiftCheckException) as excinfo:
-        result = MemoryAvailability(execute_module=NotImplementedError).run(tmp=None, task_vars=dict(
-            openshift=dict(common=dict(
-                service_type='origin',
-                is_containerized=False,
-            )),
-            group_names=["invalid"],
-            ansible_memtotal_mb=1234567,
-        ))
-
-    assert "'invalid'" in str(excinfo.value)
+def fake_execute_module(*args):
+    raise AssertionError('this function should not be called')
--- a/roles/openshift_health_checker/openshift_checks/mixins.py
+++ b/roles/openshift_health_checker/openshift_checks/mixins.py
@@ -1,4 +1,8 @@
-# pylint: disable=missing-docstring
+# pylint: disable=missing-docstring,too-few-public-methods
+"""
+Mixin classes meant to be used with subclasses of OpenShiftCheck.
+"""
+
 from openshift_checks import get_var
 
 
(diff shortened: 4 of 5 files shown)


roles/openshift_health_checker/openshift_checks/ram_diskspace_availability.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # attempt to pull required docker images and fail if
+    # any images are missing, or error occurs during pull
+    def run(self, tmp, task_vars):
+        ansible_mounts = get_var(task_vars, "ansible_mounts")
+        diskfree = self.to_gigabytes(self.get_disk_space(ansible_mounts))
+        memoryfree = get_var(task_vars, "facter_memoryfree")
+
+        recommended_memory = get_var(task_vars, "min_memory_gb")
+        recommended_storage = get_var(task_vars, "min_diskspace_gb")
```
juanvallejo: @brenton not sure where these values should be coming from

roles/openshift_health_checker/openshift_checks/ram_diskspace_availability.py
```diff
@@ -0,0 +1,42 @@
...
+    """Check that recommended memory and disk-space are available."""
+
+    name = "ram_diskspace_availability"
+    tags = ["preflight"]
+
+    # attempt to pull required docker images and fail if
+    # any images are missing, or error occurs during pull
+    def run(self, tmp, task_vars):
+        ansible_mounts = get_var(task_vars, "ansible_mounts")
+        diskfree = self.to_gigabytes(self.get_disk_space(ansible_mounts))
```
juanvallejo: @brenton or @rhcarvalho This is the total value of the `/` filesystem, not sure if I should be checking that, or `openshift-xfs-vol-dir`. I am getting this information from `ansible_mounts`

roles/openshift_health_checker/openshift_checks/ram_diskspace_availability.py
```diff
@@ -0,0 +1,42 @@
+# pylint: disable=missing-docstring
+from openshift_checks import OpenShiftCheck, get_var
+from openshift_checks.mixins import NotContainerizedMixin
+
+
+class RamDiskspaceAvailability(NotContainerizedMixin, OpenShiftCheck):
```
detiber: Should the memory and diskspace checks be separated?

I would think that Memory checks would apply equally to containerized/non-containerized.

Are there plans to also add a disk space check for containerized installs as well?

rhcarvalho: :+1:  to splitting into 2 checks.

roles/openshift_health_checker/openshift_checks/ram_diskspace_availability.py
brenton: You can get the supported values here:
https://docs.openshift.com/container-platform/latest/install_config/install/prerequisites.html#hardware

Even though there may be different requirements for older versions of the product, in general I think we should always encourage customers to meet the requirements of the latest version so they don't hit problems when they upgrade.

roles/openshift_health_checker/openshift_checks/ram_diskspace_availability.py
brenton: @detiber, do you have any idea where `openshift-xfs-vol-dir` came from?

detiber: Is it a disk label or a lv name? In any case, I'm not sure we can rely on specific values.

I think the best we can do is match a device with specific directories we care about.

For containerized `docker info` would work for devicemapper w/ a thin pool, not sure about other backends.

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,44 @@
...
+        if float(diskfree) < recommended_diskspace_gb:
+            return {"failed": True, "msg": "Available disk space for %s (%s) below recommended storage. Minimum required disk space: %s GB" % ("node" if "nodes" in group_names else "master", str(diskfree) + " GB", recommended_diskspace_gb)}
+
+        return {"changed": False}
+
+    @staticmethod
+    def get_disk_space(ansible_mounts, is_containerized):
+        if len(ansible_mounts):
+            for mnt in ansible_mounts:
+                if mnt.get("mount") == "/":
```
sosiouxme: Technically the requirements are for the filesystem containing `/var` (https://docs.openshift.com/container-platform/3.4/install_config/install/prerequisites.html#hardware). So should check if `/var` is mounted separately before using `/`.

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,31 @@
...
+    tags = ["preflight"]
+
+    recommended_node_memory_gb = 8.0
+    recommended_master_memory_gb = 16.0
+
+    def run(self, tmp, task_vars):
+        group_names = get_var(task_vars, "group_names", default=[])
+        memoryfree = get_var(task_vars, "facter_memoryfree")
+
+        recommended_memory_gb = self.recommended_master_memory_gb
```
detiber: Masters *should* be nodes, but in general they are unschedulable nodes, so they do not require additional disk space.

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,44 @@
+# pylint: disable=missing-docstring
+from openshift_checks import OpenShiftCheck, get_var
+from openshift_checks.mixins import NotContainerizedMixin
```
rhcarvalho: This seems to be a leftover, unused import.
Unlike Go, the Python interpreter doesn't complain, but I'm sure this is failing our CI.

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,44 @@
...
+
+import json
```
rhcarvalho: Another case of imported and not used.

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class DiskAvailability(OpenShiftCheck):
+    """Check that recommended disk space is available."""
+
+    name = "disk_availability"
+    tags = ["preflight"]
+
+    recommended_node_diskspace_gb = 15.0
+    recommended_master_diskspace_gb = 40.0
```
rhcarvalho: What about things like etcd?

roles/openshift_health_checker/openshift_checks/memory_availability.py
rhcarvalho: > Pythonistas? :)

I'd go with a `dict` and `max`:

```python
class MemoryAvailability(OpenShiftCheck):
    recommended_memory = {
        "nodes": X,
        "masters": Y,
        "etcd": Z,
        "nfs": M,
        "lb": N,
    }

    def run(...):
        # ...

        recommended_memory = max(self.recommended_memory.get(group, 0) for group in group_names)
```

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,44 @@
...
+        if len(ansible_mounts):
+            for mnt in ansible_mounts:
+                if mnt.get("mount") == "/":
+                    return mnt.get("size_available")
+
+        return None
+
+    @staticmethod
+    def to_gigabytes(total_bytes):
+        return total_bytes / 1073741824
```
juanvallejo: Thanks, fixed!

roles/openshift_health_checker/openshift_checks/ram_diskspace_availability.py
juanvallejo: > For containerized docker info would work for devicemapper w/ a thin pool, not sure about other backends.

Thanks, for now at least, I have added support for checking containerized installs that use devicemapper for storage driver

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,31 @@
...
+
+    name = "memory_availability"
+    tags = ["preflight"]
+
+    recommended_node_memory_gb = 8.0
+    recommended_master_memory_gb = 16.0
+
+    def run(self, tmp, task_vars):
+        group_names = get_var(task_vars, "group_names", default=[])
+        memoryfree = get_var(task_vars, "facter_memoryfree")
```
sosiouxme: :+1: `ansible_memtotal_mb`

juanvallejo:
@sdodson wondering if you could give this patch one last pass? Thanks

roles/openshift_health_checker/library/docker_info.py
```diff
@@ -0,0 +1,23 @@
...
+    )
+
+    client.module.exit_json(
+        changed=False,
+        info=client.info(),
+    )
+
+
+if __name__ == '__main__':
+    main()
```
detiber: Would it make sense to make this available to other roles? I could see this being of value in the docker-related roles as well. It might also be nice to return back at least some of the values as 'facts' instead of just a single `info` dict.

roles/openshift_health_checker/library/docker_info.py
```diff
@@ -0,0 +1,23 @@
...
+Ansible module for determining information about the docker host.
+"""
+
+from ansible.module_utils.docker_common import AnsibleDockerClient
+
+
+def main():
+    client = AnsibleDockerClient(
+        argument_spec=dict(
+            name=dict(type='list'),
```
detiber: `name` doesn't appear to be used anywhere.

I would add `supports_check_mode=True` to params for AnsibleDockerClient.

roles/openshift_health_checker/library/docker_info.py
juanvallejo: Sure, will update this to return its values under `ansible_facts`

juanvallejo: hm, maybe it would be better to rename the module from `docker_info` to simply `docker`? That way, it makes sense for it to return not just info, but docker client info.

detiber: `docker_facts` would probably be better, `docker` would shadow the older (and deprecated) docker module upstream.

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+
+class DiskAvailability(OpenShiftCheck):
+    """Check that recommended disk space is available."""
+
+    name = "disk_availability"
+    tags = ["preflight"]
+
+    recommended_node_space_gb = 15.0
+    recommended_master_space_gb = 40.0
+    minimum_docker_space_percent = 5.0
```
rhcarvalho: These should be configurable in run time.

They may be defined as variables in the `preflight/check.yml` playbook, and their names should have some prefix to avoid name collisions...

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+    supported_docker_dev_driver = "devicemapper"
+
+    def run(self, tmp, task_vars):
+        is_containerized = get_var(task_vars, "openshift", "common", "is_containerized")
+        group_names = get_var(task_vars, "group_names", default=[])
+        ansible_mounts = get_var(task_vars, "ansible_mounts")
+        diskfree = self.to_gibibytes(self.get_disk_space(ansible_mounts))
+
+        # if running containerized installation check
+        # that available space is not below 5% instead
```
rhcarvalho: Comments should not refer to specific values like "5%" that are defined in a variable elsewhere.

(multiple occurrences, check all)

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+        if is_containerized:
+            docker_facts = self.module_executor("docker_facts", {}, task_vars)
+            info = docker_facts.get("docker_facts")
+            status = dict(info.get("DriverStatus"))
+
+            if info.get("Driver", None) == self.supported_docker_dev_driver:
```
rhcarvalho: The behavior of this check doesn't seem to be clearly defined when the driver is not what we consider to be "supported".
We need to clarify the intent.

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+                disktotal = status.get("Data Space Total", None)
```
rhcarvalho: What's the guarantee we have on these key names "Data Space Total", etc? I'm afraid those are not guaranteed APIs to rely on.

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+                diskavail = status.get("Data Space Available", None)
```
rhcarvalho: Unless in a context that would otherwise require `None` for clarity, in Python dictionaries `dict.get` has a default value of `None`, so it is more common to write:

```python
some_dict.get("some key")
```

However, in this case here there is a problem with data types. I'd avoid mixing whatever is that type we expect to find in the dictionary (apparently a `str`) x `NoneType`.

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+            "mb": 1000 ** 2,
+            "gib": 1024 ** 3,
+            "gb": 1000 ** 3,
+            "tib": 1024 ** 4,
+            "tb": 1000 ** 4,
+            "pib": 1024 ** 5,
+            "pb": 1000 ** 5
+        }
+
+        segments = strfmt.lower().split(" ")
```
rhcarvalho: There can easily be an `AttributeError` exception here. We use `strfmt` as if it was a `str`, but in the code above we may clearly be calling this function with `None` as the first argument...

In a dynamically typed language, we have to keep our usage of data types sane.

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,33 @@
...
+        "etcd": 20.0
+    }
+
+    def run(self, tmp, task_vars):
+        group_names = get_var(task_vars, "group_names", default=[])
+        memoryfree = get_var(task_vars, "ansible_memtotal_mb")
+
+        recommended_memory_gb = max(self.recommended_memory_gb.get(group, 0) for group in group_names)
+
+        if self.to_gigabytes(memoryfree) < recommended_memory_gb:
```
rhcarvalho: Why don't we store the recommended values in the unit we want to compare to (MB)?

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,33 @@
...
+class MemoryAvailability(OpenShiftCheck):
+    """Check that recommended memory is available."""
+
+    name = "memory_availability"
+    tags = ["preflight"]
+
+    recommended_memory_gb = {
+        "nodes": 8.0,
+        "masters": 16.0,
+        "etcd": 20.0
```
rhcarvalho: It is good practice to include a comma at the end of every line in multiline definitions like this.
It makes it easier to add / remove / move items without having to make changes to other lines.

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    def run(self, tmp, task_vars):
+        group_names = get_var(task_vars, "group_names", default=[])
+        memoryfree = get_var(task_vars, "ansible_memtotal_mb")
+
+        recommended_memory_gb = max(self.recommended_memory_gb.get(group, 0) for group in group_names)
+
+        if self.to_gigabytes(memoryfree) < recommended_memory_gb:
+            kind = "master" if "masters" in group_names else "node"
```
rhcarvalho: This is incorrect. There are many other possibilities for `group_names`, we can't call everything else a "node" (there are `etcd`, `nfs`, ...).

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+        # that available space is not below 5% instead
+        if is_containerized:
+            docker_facts = self.module_executor("docker_facts", {}, task_vars)
+            info = docker_facts.get("docker_facts")
+            status = dict(info.get("DriverStatus"))
+
+            if info.get("Driver", None) == self.supported_docker_dev_driver:
+                disktotal = status.get("Data Space Total", None)
+                diskavail = status.get("Data Space Available", None)
+                diskleft_percent = self.to_bytes(diskavail) / self.to_bytes(disktotal) * 100
```
rhcarvalho: There could be division by zero here...

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+
+                # fail if less than 5% of available dataspace space left
+                if int(diskleft_percent) < self.minimum_docker_space_percent:
```
rhcarvalho: Be mindful of rounding problems here.

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+                    msg = "The amount of data space remaining (%s%%) for the docker storage driver is below %s%%" \
+                          % (int(diskleft_percent), self.minimum_docker_space_percent)
```
rhcarvalho: We should not use `%s` (formatting of strings) with `int` values.

And we should prefer using `str.format` instead of the `%` operator.
(multiple occurrences, check all)

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+                    return {"failed": True, "msg": msg}
+
+            return {"changed": False}
+
+        recommended_diskspace_gb = self.recommended_master_space_gb
+
+        if "nodes" in group_names:
+            recommended_diskspace_gb = self.recommended_node_space_gb
```
rhcarvalho: This logic is wrong.

- There are more group names that `masters` and `nodes`.
- If `"nodes" in group_names`, we should not overwrite `recommended_diskspace_gb` -- what if `"masters"` is also in `group_names`?!

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if float(diskfree) < recommended_diskspace_gb:
```
rhcarvalho: Why the conversion to `float`?!

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def run(self, tmp, task_vars):
+        is_containerized = get_var(task_vars, "openshift", "common", "is_containerized")
+        group_names = get_var(task_vars, "group_names", default=[])
+        ansible_mounts = get_var(task_vars, "ansible_mounts")
+        diskfree = self.to_gibibytes(self.get_disk_space(ansible_mounts))
+
+        # if running containerized installation check
+        # that available space is not below 5% instead
+        if is_containerized:
```
rhcarvalho: For ease of reading and understanding what this check does, please make the "run" method short and clear on its intents.

For example, we could have `run_containerized` and `run_non_containerized` methods to which `run` delegates depending on `is_containerized`.

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+    def get_disk_space(ansible_mounts):
+        if not ansible_mounts:
+            return None
+
+        root_size_available = None
+        for mnt in ansible_mounts:
+            if mnt.get("mount") == "/":
+                root_size_available = mnt.get("size_available")
+            if mnt.get("mount") == "/var":
+                return mnt.get("size_available")
```
rhcarvalho: Why do we return early on `/var`, but keep iterating when we see `/`?!

We need to clarify the intent here, what is this function promising to return? "get disk space" is not a great name, we can try to improve it once the expected output is clear.

From the recommended available storage, is that 100% towards Docker storage or do we need to look at Docker storage + other mount points?

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if float(diskfree) < recommended_diskspace_gb:
+            msg = "Available disk space for %s (%s) below recommended storage. Minimum required disk space: %s GB" \
+                  % ("node" if "nodes" in group_names else "master", str(diskfree) + " GB", recommended_diskspace_gb)
+            return {"failed": True, "msg": msg}
+
+        return {"changed": False}
+
+    @staticmethod
+    def get_disk_space(ansible_mounts):
```
tbielawa: A docstring here would make it clear what the `ansible_mounts` parameter should be for the function to work properly. It's a list of hashes I guess? 

```python
ansible_mounts = []
ansible_mounts.append({'size_available': 12345, 'mount': '/'})
ansible_mounts.append({'size_available': 34512, 'mount': '/var'})
```

Right? (plus other possible params)

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
juanvallejo: hm, they do not seem to be guaranteed [as far as I have seen](https://docs.docker.com/engine/api/v1.26/#operation/SystemInfo).

Another idea to retrieve these values would be to use the `docker_facts` module to get the value of `DockerRootDir` (`/var/lib/docker` by default), and perform `df` on it:

```
$ df --output=avail -B G  /var/lib/docker | grep "G" -m 1
 20G
```

Simpler yet, we could just use `df` to get the mountpoint for `/var/lib/docker`, then look for that in `ansible_mounts` under `task_vars` and get available and total values from there.

juanvallejo: @rhcarvalho 
Upon looking at documentation for devmapper https://github.com/docker/docker/tree/master/daemon/graphdriver/devmapper#information-on-docker-info, the values `Data Space Total` and `Data Space Available` appear in the linked readme. Although the values underneath the storage driver vary from driver to driver, since we are first checking that the host machine uses devicemapper, I would prefer to rely on these values

roles/openshift_health_checker/openshift_checks/memory_availability.py
juanvallejo: Thanks, will fix. Originally referenced [this check](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_health_checker/openshift_checks/package_availability.py#L18) to determine this logic

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
juanvallejo: Thanks for the feedback. `ansible_mounts` is a map like you described. I updated the method to be more legible, added a docstring

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
juanvallejo: I return immediately if `/var` exists at all. If not, `/` is the fallback mount path. I update this method to be a bit more legible

juanvallejo: > From the recommended available storage, is that 100% towards Docker storage or do we need to look at Docker storage + other mount points?

I meant for this to be a check on the mount point for the docker volume. At least in my case, the mount point for its volume was `/`, so I checked for the overall disk space remaining on that point

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
juanvallejo: Thanks, I now make sure to pass a string type to this method. Corrected this in other places as well

roles/openshift_health_checker/openshift_checks/memory_availability.py
rhcarvalho: The logic there is different, it is additive: start with an empty set and add items based on the group names. Here we are incorrectly assuming that everything that is not master is a node.

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,115 @@
...
+            if failed:
+                return {"failed": True, "msg": msg}
+
+        failed, msg = self.noncontainerized_volume_check(ansible_mounts, task_vars)
+        return {"failed": failed, "msg": msg}
+
+    def containerized_volume_check(self, task_vars):
+        docker_facts = self.module_executor("docker_facts", {}, task_vars)
+        info = docker_facts.get("docker_facts")
+        status = dict(info.get("DriverStatus"))
```
rhcarvalho:
```python
>>> dict(None)
TypeError: 'NoneType' object is not iterable
```

```python
>>> info = None
>>> info.get("DriverStatus")
AttributeError: 'NoneType' object has no attribute 'get'
```

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,115 @@
...
+        is_containerized = get_var(task_vars, "openshift", "common", "is_containerized")
+        ansible_mounts = get_var(task_vars, "ansible_mounts")
+
+        # if running containerized installation check that available
+        # space is not below defined minimum threshold
+        if is_containerized:
+            failed, msg = self.containerized_volume_check(task_vars)
+            if failed:
+                return {"failed": True, "msg": msg}
+
```
rhcarvalho: The way it stands now, we will call both `containerized_volume_check` and `noncontainerized_volume_check` when `is_containerized` is true and `failed` is false...

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,115 @@
...
+            return True, msg
+
+        return False, ""
+
+    def get_recommended_diskspace_gb(self, task_vars):
+        group_names = get_var(task_vars, "group_names", default=[])
+        if "masters" in group_names:
+            return self.recommended_master_space_gb
+
+        return self.recommended_node_space_gb
```
rhcarvalho: This is also assuming that everything that is not a master is a node. If we expect to run this check on non-master-non-node hosts, we should follow the docs and use the proper storage required for each type of host.

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class MemoryAvailability(OpenShiftCheck):
+    """Check that recommended memory is available."""
+
+    name = "memory_availability"
+    tags = ["preflight"]
+
+    recommended_memory_mb = {
+        "nodes": 8000.0,
```
juanvallejo: I think 8000 is fine, unless we prefer MiB instead?

rhcarvalho: Our docs use the decimal prefixes, let's go with powers of 10 then.
https://docs.openshift.org/latest/install_config/install/prerequisites.html

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    def run(self, tmp, task_vars):
+        group_names = get_var(task_vars, "group_names", default=[])
+        memoryfree = get_var(task_vars, "ansible_memtotal_mb")
+
+        recommended_memory_mb = max(self.recommended_memory_mb.get(group, 0) for group in group_names)
+
+        if memoryfree < recommended_memory_mb:
+            msg = "Available memory ({available} GB) below recommended storage. Minimum required memory: {recommended} GB".format(available=memoryfree, recommended=recommended_memory_mb)
```
juanvallejo: Will fix, taking care of linting issues towards the end

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,31 @@
...
+            return {"failed": True, "msg": msg}
+
+        return {"changed": False}
+
+    @staticmethod
+    def to_gigabytes(size_in_mb):
```
juanvallejo: Thanks, will remove

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
juanvallejo: Thanks for catching that :)

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
juanvallejo: hm, will go ahead and use info from https://docs.openshift.com/container-platform/3.4/install_config/install/prerequisites.html#hardware, however it only contains reqs for masters, nodes, and etcd

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,115 @@
...
+            "kib": 1024,
+            "kb": 1000,
+            "mib": 1024 ** 2,
+            "mb": 1000 ** 2,
+            "gib": 1024 ** 3,
+            "gb": 1000 ** 3,
+            "tib": 1024 ** 4,
+            "tb": 1000 ** 4,
+            "pib": 1024 ** 5,
+            "pb": 1000 ** 5
```
juanvallejo: I was not sure what specific units to expect, so this at least covers standard cases. @tbielawa has written [a great python module] that we could use in cases such as this one if we are willing to include it in our openshift-ansible image / make it a requirement on the host machine. wdyt?

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        segments = strfmt.lower().split(" ")
+        if not units.get(segments[1], False):
+            return None
```
juanvallejo: Thanks, will address this throughout the code

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
rhcarvalho: I'd prefer not to do any conversions :-)
"the best code is no code at all"

juanvallejo: While I am certainly for this, I am not sure to what extent we can guarantee that the units returned by devicemapper's `Data Space Total`, `Data Space Available` will be consistent in all cases

roles/openshift_health_checker/openshift_checks/memory_availability.py
tbielawa: I scanned over the output from `docker info` and saw that it reports different measurements in different prefix unit systems.

```
...
Storage Driver: devicemapper
 Pool Blocksize: 524.3 kB
 Base Device Size: 10.74 GB
 Data Space Used: 1.42 GB
 Data Space Total: 44.43 GB
 Data Space Available: 43.01 GB
 Metadata Space Used: 1.507 MB
 Metadata Space Total: 511.7 MB
 Metadata Space Available: 510.2 MB
 Thin Pool Minimum Free Space: 4.443 GB
...
Total Memory: 15.11 GiB
...
```

Sizes are reported using the SI base-10 system (`kB`, `MB`, `GB`) and memory is reported using the NIST base-2 system (`KiB`, `GiB`, etc).

As it was pointed out before, I do [maintain a library, 'bitmath'](http://bitmath.readthedocs.io/en/latest/index.html) (200 unit tests, in RHEL/Fedora, etc etc, etc) that handles all of that automatically (and more).

Here's a quick gist I made that takes code from this PR and shows the bitmath equivalent.
* [parse_docker_info.py](https://gist.github.com/tbielawa/a24316f6aa83a1c841220e4cb71f94ac)

I think that trying to get bitmath available as an RPM in the RHSM channels would be a big effort, and getting it into Atomic is nigh impossible. I'm OK w/ vendoring/slipping the bitmath library into this project.

But that's a bigger conversation we should probably have as a team.

tbielawa:
I'm not sure if it is showing up anymore, due to the folded 'Hide Outdated' segments above, but I did post another comment with more information on bitmath and the problem of converting/parsing inconsistent units. [Direct link to above comment here](https://github.com/openshift/openshift-ansible/pull/3518#discussion_r108471585)

https://github.com/openshift/openshift-ansible/pull/3518#discussion_r108471585

roles/openshift_health_checker/openshift_checks/diskspace_availability.py
```diff
@@ -0,0 +1,121 @@
+# pylint: disable=missing-docstring
+from openshift_checks import OpenShiftCheck, get_var
+
+
+class DiskAvailability(OpenShiftCheck):
```
rhcarvalho: The name of the file is inconsistent with the check name and class name.

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -0,0 +1,121 @@
...
+            "tb": 1000**4,
+            "pib": 1024**5,
+            "pb": 1000**5,
+        }
+
+        segments = strfmt.lower().split(" ")
+        if not units.get(segments[1]):
+            return 0
+
+        byte_size = float(segments[0]) * float(units.get(segments[1]))
```
sosiouxme: why float? bytes are integers...

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -0,0 +1,121 @@
+# pylint: disable=missing-docstring
+from openshift_checks import OpenShiftCheck, get_var
+
+
+class DiskAvailability(OpenShiftCheck):
+    """Check that recommended disk space is available."""
```
sosiouxme: Not sure how else we should annotate it but it should be clear that this is disk space prior to an install. If running preflight checks before an upgrade, they may have used a bunch of the space and that should be fine; we should probably not even run this check in that case. We would do a different kind of calculation for hosts that are already installed and running.

roles/openshift_health_checker/openshift_checks/memory_availability.py
sosiouxme: This is memory though which isn't coming from docker info... comes from ansible setup. Whatever our docs say, nobody measures RAM in GB (only disk manufacturers get to pull that crap), so it seems to me MiB/GiB should be used.

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,28 @@
...
+
+    recommended_memory_mb = {
+        "nodes": 8000.0,
+        "masters": 16000.0,
+        "etcd": 20000.0,
+    }
+
+    def run(self, tmp, task_vars):
+        group_names = get_var(task_vars, "group_names", default=[])
+        memoryfree = get_var(task_vars, "ansible_memtotal_mb")
```
sosiouxme: nit, shouldn't `memoryfree` be named `total_memory` or similar?

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -0,0 +1,121 @@
...
+    tags = ["preflight"]
+
+    recommended_diskspace = {
+        "nodes": 15*10**9,
+        "masters": 40*10**9,
+        "etcd": 20*10**9,
+    }
+
+    minimum_docker_space_percent = 5.0
+    supported_docker_dev_driver = "devicemapper"
```
juanvallejo: Yeah, I was thinking of removing the check for the storage driver from this PR altogether, and just handling it as part of the `docker_storage_check` PR https://github.com/openshift/openshift-ansible/pull/3787/files#diff-c3d0f0c6a0d94ad7d6c5a62da695118dR17

juanvallejo: Actually, remembering a very similar conversation I had with @rhcarvalho about this, it would be better to have the storage driver check be its own preflight check, allowing this PR and https://github.com/openshift/openshift-ansible/pull/3787 to focus on a single thing. WDYT?

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -0,0 +1,121 @@
...
+            failed, msg = self.containerized_volume_check(task_vars)
+            if failed:
+                return {"failed": True, "msg": msg}
+
+            return {"changed": False}
+
+        failed, msg = self.noncontainerized_volume_check(ansible_mounts, task_vars)
+        return {"failed": failed, "msg": msg}
+
+    def containerized_volume_check(self, task_vars):
```
juanvallejo: Thanks, per the comments on  https://github.com/openshift/openshift-ansible/pull/3518#discussion_r110243468, I have decided to leave all aspects of the containerized check to https://github.com/openshift/openshift-ansible/pull/3787. (I will further refine that PR, rename it so the name is consistent with the check in this PR, and split out the storage driver check from it into its own check). Any of your feedback on #3787 is definitely welcome :)

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -0,0 +1,121 @@
...
+        # is known to support. The only supported driver is currently "devicemapper"
+        if info.get("Driver", "") == self.supported_docker_dev_driver:
+            disktotal = self.to_bytes(status.get("Data Space Total", ""))
+            diskavail = self.to_bytes(status.get("Data Space Available", ""))
+
+            if not disktotal or not diskavail:
+                msg = "Unable to retrieve storage driver information for {driver}."
+                return True, msg.format(driver=self.supported_docker_dev_driver)
+
+            diskleft_percent = diskavail / disktotal * 100
```
juanvallejo: Ah, I was not sure if it would be approproate to use the [master/node](https://docs.openshift.org/latest/install_config/install/prerequisites.html#system-requirements) minimum diskpace requirements for the containerized check as well, therefore I resorted to a percentage (check that docker volume storage is not at > 90% full for example)

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -0,0 +1,121 @@
...
+                return True, msg.format(remaining=diskleft_percent,
+                                        min=self.minimum_docker_space_percent)
+
+    def noncontainerized_volume_check(self, ansible_mounts, task_vars):
+        group_names = get_var(task_vars, "group_names", default=[])
+
+        recommended_diskspace_bytes = max(self.recommended_diskspace.get(group, 0) for group in group_names)
+        openshift_diskfree_bytes = self.get_openshift_disk_availability(ansible_mounts)
+
+        if openshift_diskfree_bytes < recommended_diskspace_bytes:
```
juanvallejo: Ah, I am converting these values to gigabytes in the `"".format` method

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,31 @@
...
+class MemoryAvailability(OpenShiftCheck):
+    """Check that recommended memory is available."""
+
+    name = "memory_availability"
+    tags = ["preflight"]
+
+    recommended_memory_mb = {
+        "nodes": 8 * 1024,
+        "masters": 16 * 1024,
+        "etcd": 10 * 1024,
```
rhcarvalho: Why do we use base 10 for disk and base 2 for memory?

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,31 @@
...
+    }
+
```
rhcarvalho: We need to implement `is_active` for this check, similar to DiskAvailability: only run for hosts that we actually have a recommendation to verify.

roles/openshift_health_checker/openshift_checks/memory_availability.py
```diff
@@ -0,0 +1,31 @@
...
+        recommended_memory_mb = max(self.recommended_memory_mb.get(group, 0) for group in group_names)
+        if not recommended_memory_mb:
+            msg = "Unable to determine recommended memory size for group_name {group_name}"
+            raise OpenShiftCheckException(msg.format(group_name=group_names))
+
+        if total_memory < recommended_memory_mb:
+            msg = ("Available memory ({available} GB) below recommended storage. "
+                   "Minimum required memory: {recommended} GB")
+            return {"failed": True, "msg": msg.format(available=total_memory, recommended=recommended_memory_mb)}
```
rhcarvalho: The units are mismatched... we say "GB" in `msg`, but the variables are in MB...

roles/openshift_health_checker/test/memory_availability_test.py
```diff
@@ -0,0 +1,72 @@
...
+        "etcd",
+        12200,
+    ),
+])
+def test_mem_check_with_recommended_memtotal(group_name, size_available):
+    result = MemoryAvailability(execute_module=NotImplementedError).run(tmp=None, task_vars=dict(
+        openshift=dict(common=dict(
+            service_type='origin',
+            is_containerized=False,
+        )),
```
rhcarvalho: The `openshift` key/value is probably irrelevant here, isn't it?

roles/openshift_health_checker/test/memory_availability_test.py
```diff
@@ -0,0 +1,84 @@
...
+        2,
+    ),
+    (
+        "etcd",
+        3,
+    ),
+])
+def test_mem_check_with_insufficient_memtotal(group_name, size_available):
+    result = MemoryAvailability(execute_module=NotImplementedError).run(tmp=None, task_vars=dict(
+        group_names=[group_name],
```
rhcarvalho: When all tests use a single group, we miss the nuance that there is a "max of recommend memories based on roles" going on.
I'm addressing it in a soon-to-be-pushed-here commit, but wanted to let you know ;)

rhcarvalho:
@sosiouxme @juanvallejo please give a final review.
LGTM.

juanvallejo:
[merge]

roles/openshift_health_checker/openshift_checks/disk_availability.py
sosiouxme: Err, belatedly... yes.

juanvallejo:
flake https://github.com/openshift/origin/issues/11292
re[merge]

rhcarvalho:
Flakes... @sdodson @stevekuznetsov PTAL

rhcarvalho:
This looks like what is described in https://github.com/openshift/origin/issues/11777.... @ncdc could you please help determine what's happening in https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/215/?

rhcarvalho:
Hmm, does Rosie delete its own comments? The merge failure comment is gone. Magically running the merge again :-)

rhcarvalho:
flake openshift/origin#11292
re[merge]

rhcarvalho:
[test]

juanvallejo:
re[merge]

ncdc:
@rhcarvalho here's your failure:

```
SUCCESS! -- 22 Passed | 0 Failed | 0 Pending | 701 Skipped Apr 12 19:37:42.577: INFO: Error running cluster/log-dump.sh: fork/exec /data/src/github.com/openshift/origin/vendor/k8s.io/kubernetes/cluster/l│
PASS                                                                                                                                                                                                       │
                                                                                                                                                                                                           │
Ginkgo ran 1 suite in 11m50.503199671s                                                                                                                                                                     │
Test Suite Passed                                                                                                                                                                                          │
[INFO] Dumping container logs to /tmp/openshift/conformance/logs/containers                                                                                                                                │
[ERROR] PID 26302: hack/lib/cleanup.sh:74: `[[ -n "${OS_DEBUG:-}" ]]` exited with status 1.                                                                                                                │
[INFO]          Stack Trace:                                                                                                                                                                               │
[INFO]            1: hack/lib/cleanup.sh:74: `[[ -n "${OS_DEBUG:-}" ]]`                                                                                                                                    ├
[INFO]            2: test/extended/setup.sh:48: os::cleanup::dump_container_logs                                                                                                                           │
[INFO]            3: test/extended/conformance.sh:1: cleanup                                                                                                                                               │
[INFO]   Exiting with code 1.                                                                                                                                                                              │
make: *** [test-extended] Error 1                                                                                                                                                                          ┼
++ export status=FAILURE                                                                                                                                                                                   │
++ status=FAILURE                                                                                                                                                                                          │
+ set +o xtrace                                                                                                                                                                                            │
########## FINISHED STAGE: FAILURE: RUN EXTENDED TESTS ##########                                                                                                                                          │
```

rhcarvalho:
@ncdc thanks for taking a look! I have seen that, but I still do not understand it.

- `INFO: Error running cluster/log-dump.sh...`: looks like an error but is not, according to @stevekuznetsov on email:
  > This is a red-herring message that has been around for a long time. Some part of the Kubernetes e2e framework is trying to run a script that doesn't exist. It does not cause a failure and happens on every e2e run. I think the previous consensus was that we don't want to carry yet another UPSTREAM commit to get rid of the function in our vendored k8s code as it's benign. I can understand how this is confusing, though, and we can revisit that decision.

- `SUCCESS!`, `PASS`, `Test Suite Passed`: that means we're good, right? :-)

- `[ERROR] PID 26302: hack/lib/cleanup.sh:74: `[[ -n "${OS_DEBUG:-}" ]]` exited with status 1.`: what should we make of this?! Failed to dump container logs?

stevekuznetsov:
The stacktrace isn't being helpful here unfortunately, see https://github.com/openshift/origin/issues/13759 -- I'll work on the stacktrace being better soon. The error may have been in container cleanup? Unclear, unfortunately.

rhcarvalho:
aos-ci-test

rhcarvalho:
[merge]

juanvallejo:
merge test flaked on https://github.com/openshift/origin/issues/13271 re[merge]

juanvallejo:
test flaked on https://github.com/openshift/origin/issues/13271
aos-ci-test

rhcarvalho:
Test failure is a bug in the CI job -- https://github.com/openshift/origin/issues/13806

[merge]

rhcarvalho:
@sdodson do we have documented anywhere what are the **required** checks before a PR can be merged? Now in this PR I do not know if we need to re-trigger aos-ci-test.

Would be useful to operate like Origin -- where the \[merge\] automatically runs the tests it needs, without requiring 2 different bot commands, is the plan going in that direction?

stevekuznetsov:
The configuration for branches is [here](https://github.com/openshift/aos-cd-jobs/blob/f449380bad040cb9468bdc5ef3a0ea980d6d098a/test_status_config.yml). We do aim to move more and more tests away from AOS CI and onto the Origin CI, and aid in that effort is greatly appreciated, but it's not trivial and will take some time.

rhcarvalho:
@stevekuznetsov thanks for the pointer. Which branch should we refer to to the latest version of that file / what is currently enforced in in-flight PRs?

stevekuznetsov:
https://github.com/openshift/aos-cd-jobs/blob/master/sjb/test_status_config.yml

sdodson:
re[merge] 
EXPOSE THE KUBECONFIG known issue

rhcarvalho:
Last flake was https://github.com/openshift/origin/issues/13814.

Since the code here is not related to installs nor upgrades, is strictly contained in the `openshift_health_checker` role, merge failed at least 8 times with known flakes linked above, and aos-ci-test and Travis are green, we'll use the temporary exception to merge it manually.

PR #3519, opened 2017-02-28
Add docker-registry.default.svc short name to registry service signing

sdodson:
Today this isn't necessary, but we may need to push to the registry via a short name in the future to facilitate easier cluster dns zone changes.

### DIFF
--- a/roles/openshift_hosted/tasks/registry/secure.yml
+++ b/roles/openshift_hosted/tasks/registry/secure.yml
@@ -53,7 +53,8 @@
     signer_serial: "{{ openshift_master_config_dir }}/ca.serial.txt"
     hostnames:
     - "{{ docker_registry_service_ip.results.clusterip }}"
-    - docker-registry.default.svc.cluster.local
+    - "{{ openshift_hosted_registry_name }}.default.svc"
+    - "{{ openshift_hosted_registry_name }}.default.svc.{{ openshift.common.dns_domain }}"
     - "{{ docker_registry_route_hostname }}"
     cert: "{{ openshift_master_config_dir }}/registry.crt"
     key: "{{ openshift_master_config_dir }}/registry.key"
--- a/playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
+++ b/playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
@@ -66,7 +66,7 @@
         --signer-cert={{ openshift.common.config_base }}/master/ca.crt
         --signer-key={{ openshift.common.config_base }}/master/ca.key
         --signer-serial={{ openshift.common.config_base }}/master/ca.serial.txt
-        --hostnames="{{ docker_registry_service_ip.results.clusterip }},docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}"
+        --hostnames="{{ docker_registry_service_ip.results.clusterip }},docker-registry.default.svc,docker-registry.default.svc.cluster.local,{{ docker_registry_route_hostname }}"
         --cert={{ openshift.common.config_base }}/master/registry.crt
         --key={{ openshift.common.config_base }}/master/registry.key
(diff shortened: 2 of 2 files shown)


roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -25,7 +25,8 @@
     signer_serial: "{{ openshift_master_config_dir }}/ca.serial.txt"
     hostnames:
     - "{{ docker_registry_service_ip.results.clusterip }}"
-    - docker-registry.default.svc.cluster.local
+    - docker-registry.default.svc
+    - "docker-registry.default.svc.{{ openshift.common.dns_domain }}"
```
mtnbikenc: Should we try to keep these openshift_facts vars out of the internals of the role as we've done with other facts for registry?

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -25,7 +25,8 @@
     signer_serial: "{{ openshift_master_config_dir }}/ca.serial.txt"
     hostnames:
     - "{{ docker_registry_service_ip.results.clusterip }}"
-    - docker-registry.default.svc.cluster.local
+    - docker-registry.default.svc
```
kwoodson: @sdodson, @mtnbikenc,

Should we update these to have the name portion variable-ized?

`"{{ openshift_hosted_registry_name }}.default.svc"

roles/openshift_hosted/tasks/registry/secure.yml
detiber: Yes, but it should probably still be a defaulted variable.

kwoodson: Same as above:

`"{{ openshift_hosted_registry_name }}.default.svc.{{ openshift.common.dns_domain }}"

sdodson: This will change in the future (post 3.6 at the earliest), but right now the service is always named 'docker-registry' and in the 'default' namespace.

sdodson: Given it's static i'll drop the variable for now, we'll have to do a good bit of audit to handle this change when it happens anyway.

kwoodson: @sdodson,

I'm not sure it is static as we are creating the service here: 

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/tasks/registry/registry.yml#L56-L67

The variable is defined in registry.yml: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/tasks/registry/registry.yml#L36

I'd prefer if we just used it since it will future proof us.

sdodson:
Updated, we already depend on openshift.common elsewhere in this role so assuming one more won't hurt for now.

sdodson:
@mtnbikenc need this for some 3.6 work

sdodson:
aos-ci-test

mtnbikenc:
[merge]

sdodson:
[merge]

sdodson:
Travis never posted results, kicking that build.

sdodson:
[merge]

sdodson:
aos-ci-test

abutcher:
We'll need to update the registry certificate task in the certificate redeploy playbook as well and potentially use/import the same code to make this more dry. https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/redeploy-certificates/registry.yml#L63-L74

sdodson:
@abutcher Added that, we don't have openshift_hosted_registry_name but as far as I understand it that's always going to be 'docker-registry.default.svc.cluster.local' so i don't feel too bad about it being static.

sdodson:
aos-ci-test

sdodson:
[merge]

PR #3522, opened 2017-02-28
Add 'persistentVolumeClaim' to volume_info type

mtnbikenc:
Adds 'persistentVolumeClaim' to match the standard names for `oc volume`.


mtnbikenc:
aos-ci-test

PR #3527, opened 2017-02-28
raise exceptions when walking through object path

joelddiaz:
if we're given path a.b.c and the existing object is:
a:
  b:
   - item1
raise an exception due to unexpected objects found while traversing the path (ie. b is a list, not a dict)

also, add_entry assumes new dicts for each sub element when creating elements besides the final assignment value.
doing something like a.b.c[0] = 12 where 'c' doesn't exist raises an exception

add test cases to cover:
access path that differs from existing object
create new objects with an embedded list in the path
create new object with a list at the end (define the end list in the passed in 'value' to avoid this exception)

### DIFF
--- a/roles/lib_utils/src/test/unit/test_yedit.py
+++ b/roles/lib_utils/src/test/unit/test_yedit.py
@@ -23,7 +23,7 @@ import unittest
 yedit_path = os.path.join('/'.join(os.path.realpath(__file__).split('/')[:-4]), 'library')  # noqa: E501
 sys.path.insert(0, yedit_path)
 
-from yedit import Yedit  # noqa: E402
+from yedit import Yedit, YeditException  # noqa: E402
 
 # pylint: disable=too-many-public-methods
 # Silly pylint, moar tests!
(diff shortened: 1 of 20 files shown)


kwoodson:
aos-ci-test

roles/lib_utils/src/test/unit/test_yedit.py
```diff
@@ -200,6 +200,7 @@ def test_append_twice_to_list(self):
         yed.append('x:y:z', [5, 6])
         yed.append('x:y:z', [5, 6])
         self.assertTrue(yed.get('x:y:z') == [1, 2, 3, [5, 6], [5, 6]])
+        # pylint: disable=maybe-no-member
```
rhcarvalho: ~~Currently the tests are not going through `pylint` (I'm changing that in #3511).
In our [Best Practices Guide](https://github.com/openshift/openshift-ansible/blob/master/docs/best_practices_guide.adoc#All-PyLint-rule-disables-MUST-be-documented-in-the-code) we say to always document pylint disables.~~

~~I volunteer to document this one if someone explain why it was added?~~

---

The line below, `self.assertTrue(2 == yed.get('x:y:z').count([5, 6]))` seems redundant after the line above `self.assertTrue(yed.get('x:y:z') == [1, 2, 3, [5, 6], [5, 6]])`.

If `x:y:z` is the given list, then we are just testing Python when we ask if the count of `[5, 6]` is two...

I'll go ahead and remove this redundant assert.

rhcarvalho: Removing the redundant assert in #3563

PR #3529, opened 2017-03-01
WIP: update excluders to latest by default, in non-upgrade scenarios do not update

ingvagabund:
The excluders should be updated to the latest only when the upgrade is run. Not during installation.

**TODO**:
  - [x] during installation do not upgrade excluders (keep the rpms as they are)
  - [x] during upgrade update both excluders to the latest
  - [x] during upgrade check if available version of any excluder is not higher than the upgrade target version
  - [x] excluder rpms now provide status command which should be used to check excluder status
  - [x] introduce new variables to disable/enable both excluders and one general variable to disable/enable all excluders

**INFO**
- When the `openshift-excluder` is enabled, the repoquery for openshift version does not work. So we should disabled the excluder before running `openshift_version` role.
- when an excluder is installed, ``exclude`` command is run during the installation

**Changes**:
* Introducing new variables:
  * enable_docker_excluder - if not set, the docker excluder is completely ignored (no matter if it is installed or not), if set, the docker excluder is installed (if not already), status is run and it the status is non-zero, the docker excluder is enabled no matter what
  * enable_openshift_excluder - if not set, the docker excluder is completely ignored (no matter if it is installed or not), If set, the openshift excluder is installed (if not already). status is run and if the status is non-zero, the openshift excluder is disabled no matter what
  * enable_excluders - if set, it overrides ``enable_docker_excluder``, resp. ``enable_openshift_exluder`` if they are not set. If an excluder is set, the ``enable_excluder`` has no effect on the excluder.
* during installation an excluder is installed only if it is enabled
* during upgrade an excluder is installed, resp. updated to the latest if enabled
* if upgrade to version `N` is run and if there is a repos with openshift higher than `N+1` enabled, the upgrade reports excluder validation error and fails.

DO NOT MERGE!!!


roles/openshift_excluder/tasks/install.yml
```diff
@@ -2,15 +2,15 @@
...
   - not openshift.common.is_containerized | bool
 
 - name: Install latest docker excluder
   package:
     name: "{{ openshift.common.service_type }}-excluder"
-    state: latest
+    state: "{{ docker_excluder_package_state }}"
   when:
   - docker_excluder_enabled | default(false) | bool
   - not openshift.common.is_containerized | bool
```
detiber: You could leave `state: latest` and rename this task to `tasks/upgrade.yml` and then just have the install call into `status.yml`.

Then you could update the upgrade to use `tasks_from: upgrade` to upgrade the packages prior to enabling/disabling.

ingvagabund: The problem is not the ``openshift_excluder`` role itself but playbooks that use the role. Both install and upgrade are included by common playbook that is responsible for unexcluding excluders. Part of the unexclude process is running ``openshift_excluder.status`` which installs both excluder rpms. I will need to break the common playbook into two and run install/upgrade task accordingly.

ingvagabund:
aos-ci-test

roles/openshift_excluder/tasks/install.yml
ingvagabund: I have introduce another playbook ``playbooks/common/openshift-cluster/upgrades/disable_excluder.yml`` that is used in upgrades only instead of the ``playbooks/common/openshift-cluster/disable_excluder.yml``

playbooks/common/openshift-cluster/upgrades/disable_excluder.yml
```diff
@@ -0,0 +1,14 @@
...
+- name: Record excluder state and disable
+  hosts: l_oo_all_hosts
+  gather_facts: no
+  tasks:
+  - include_role:
+      name: openshift_excluder
+      tasks_from: status
+      vars:
+        excluder_package_state: latest
+        docker_excluder_package_state: latest
```
ingvagabund: With the upgrade do we want the latest or ``openshift_version`` latest excluders?

roles/openshift_excluder/tasks/status.yml
```diff
@@ -17,37 +17,37 @@
...
-
-- name: Update to the latest docker-excluder packages
-  package:
-    name: "{{ openshift.common.service_type }}-docker-excluder"
-    state: latest
-  when:
-  - "{{ docker_excluder_installed.installed_versions | default([]) | length > 0 }}"
-  - not openshift.common.is_containerized | bool
+- block:
+  - name: Update excluder packages if requested
```
ingvagabund: This currently does not make sense. During the installation if the excludes are not installed, the task is skipped. If the excluders are installed, the are not update. IIRC, @detiber's idea was not to upgrade excludes to 3.5 if we are deploying 3.4 openshift. So either we can check what OS version is installed and based on that decide which excluders version we want to update to.

### Some scenarios
Assume ``3.4.1.7-1`` excluders are installed and other ``3.4.1.8-1`` atomic-openshift rpms are installed as well. Should we update excluders to ``3.4.1.8-1``? Or is that something a user should take care of? I would say we should take care of that if we are re-running installation part of the ansible

Assume ``3.4.1.8-1`` exluders and other atomic-openshifts rpms are installed. At the same time 3.5 repo is enabled. However, ``openshift_release=v3.4``. We should not update excluders to the latest when running the installation again (or for the first time).

Assume ``3.3.1.8-1`` excluders and other atomic-openshifts rpms are installed. At the same time 3.4 and 3.5 repos are enabled. User is running upgrade from 3.3 to 3.4 with ``openshift_release=v3.4``. We should not upgrade excluders to ``3.5.*``. Instead of that, we should update excluders to the latest 3.4.

ingvagabund: The checks for the excluders version should do for now.

sdodson: After discussing with @brenton we came to the conclusion that the chances of someone having the wrong repo enabled is high enough that we'd like to make sure that we only upgrade to the latest for the version corresponding to the given upgrade playbook. ie: playbooks/byo/openshift-cluster/ugprades/v3_4/upgrade.yml would only upgrade to the latest atomic-openshift-excluder-3.4* I'm not certain that the package module can handle a glob though and we'd need to account for origin versus OCP versions.

detiber: @sdodson I still think we should avoid upgrading the excluder outside of initial install, with the exception of ensuring that we have the minimum version required for providing the status output. If someone is running playbooks/byo/openshift-cluster/config.yml for their ongoing configuration management, I'm not sure we'd want to mutate package versions unless absolutely necessary.

It's probably less important today where we have plenty of other tasks that lack idempotency, but it could be an unwelcome surprise.

sdodson: I agree, sorry, there was some offline discussion where we said we'd only upgrade in the upgrade playbooks.

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

playbooks/common/openshift-cluster/disable_excluder.yml
```diff
@@ -3,9 +3,29 @@
...
+  # So no pre-validation check here as the excluders are either to be installed (present = latest)
+  # or they are not going to be updated if already installed
+
+  # Install any excluder that is enabled
+  - include_role:
+      name: openshift_excluder
+      tasks_from: init
+  - include_role:
+      name: openshift_excluder
+      tasks_from: install
```
detiber: Having to use two include_role tasks seems a bit fragile. Would it make sense to conditionally include `install` from within `init` in the role or have `install` just include `init` prior to doing other tasks?

ingvagabund:
aos-ci-test

playbooks/common/openshift-cluster/disable_excluder.yml
```diff
@@ -3,9 +3,29 @@
...
+    vars:
+      # Both docker_excluder_on and openshift_excluder_on are set in openshift_excluder->init task
+      install_docker_excluder: docker_excluder_on
+      install_openshift_excluder: openshift_excluder_on
```
detiber: I don't think we should be referencing internal role variables outside of the role like this. As a side note, this is setting this variables to the strings `docker_exlcuder_on` and `openshift_excluder_on` rather than referencing the variables.

playbooks/common/openshift-cluster/disable_excluder.yml
```diff
@@ -3,9 +3,29 @@
...
+  # Check the current state of all excluders
   - include_role:
       name: openshift_excluder
       tasks_from: status
+
+  # And finally adjust an excluder in order to update host components correctly
   - include_role:
       name: openshift_excluder
-      tasks_from: unexclude
+      tasks_from: adjust
```
detiber: It really feels like there should be an entrypoint task file in openshift_excluder that handles the ordering of these tasks, rather than orchestrating the task files oustide of the role...

Maybe if there was a task file called `start_upgrade.yml` that contains the ordering done in this playbook, but still makes use of the more targeted task files for re-use?

playbooks/common/openshift-cluster/reset_excluder.yml
```diff
@@ -5,4 +5,12 @@
   tasks:
   - include_role:
       name: openshift_excluder
+      tasks_from: init
+  - include_role:
+      name: openshift_excluder
       tasks_from: reset
```
detiber: Similar comment here, maybe an `finish_upgrade.yml` task file that does init/reset?

detiber: The tricky part is going to be persisting the data about whether the host should be re-enabled or not...  For that, it might be good to take advantage of a local fact or persisting the status to a local file on the control host, to avoid issues with re-running the upgrade on a failed upgrade and leaving a node out of rotation.

playbooks/common/openshift-cluster/disable_excluder.yml
ingvagabund: I have restructured the role, some tasks removed, some added, some integrated.

playbooks/common/openshift-cluster/disable_excluder.yml
ingvagabund: fixed

playbooks/common/openshift-cluster/disable_excluder.yml
ingvagabund: I have introduce new enable/disable tasks that should be general and configurable enough to cover the case.

sdodson:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
Testing the upgrade now...

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

roles/openshift_excluder/tasks/unexclude.yml
```diff
@@ -1,12 +1,19 @@
 ---
-- name: disable docker excluder
-  command: "{{ openshift.common.service_type }}-docker-excluder unexclude"
-  when:
-  - docker_excluder_enabled | bool
-  - not openshift.common.is_containerized | bool
+# input variables:
+# - disable_docker_excluder
+# - disable_openshift_excluder
+- block:
```
ashcrow: :+1:

ingvagabund:
Still, do not merge yet.

ingvagabund:
Running last tests of the upgrade.

playbooks/byo/openshift-cluster/upgrades/v3_3/upgrade.yml
```diff
@@ -46,7 +46,7 @@
   tags:
   - pre_upgrade
 
-- include: ../../../../common/openshift-cluster/disable_excluder.yml
+- include: ../../../../common/openshift-cluster/upgrades/disable_excluder.yml
```
ingvagabund: I have been confused by that play for more than two days :)

ingvagabund:
Squashing

ingvagabund:
aos-ci-test

PR #3533, opened 2017-03-01
Fixed error handling when oc adm ca create-server-cert fails.  Fixed a logic error in secure.

kwoodson:


mtnbikenc:
`passthrough` case works as expected.

mtnbikenc:
aos-ci-test

PR #3535, opened 2017-03-02
Add pre-upgrade check for reserved namespaces

enj:
Adds pre-upgrade check for https://github.com/openshift/origin/issues/13147

Since the logic was basically the same as `OCSDNValidator`, I replaced it with a generic `OCObjectValidator`.

@sdodson I am assuming various playbooks will need to be updated.

@deads2k @liggitt PTAL and what version do we want to start running this check for?

@kwoodson please review.

Signed-off-by: Monis Khan <mkhan@redhat.com>


deads2k:
Assuming it does what it looks like it does, then the check is correct.  Though given the distribution of names versus annotations, I'd probably check the names first.

enj:
aos-ci-test

sdodson:
rebased

PR #3549, opened 2017-03-02
Adding support for multiple router shards.

kwoodson:


roles/openshift_hosted/tasks/router/router.yml
```diff
@@ -11,69 +11,34 @@
 - name: set_fact replicas
```
detiber: I suspect there is a bit of a chicken/egg here, but it seems like we should be able to determine the number of replicas  per-router based on the `replicas`/`selector` value for each router, rather than doing it a single time here.

I know there is some concern with using task includes, but I'm wondering if it might simplify the overall logic here, where all of the logic for a given router could be done in a separate task file, or even the logic to drive each individual router could be pushed up to `tasks/main.yml`

defaults/main.yml:
```
# With multiple routers this should probably default to whatever naming scheme is being used for the individual 
# routers, but we would need to provide some type of backward compatibility to avoid breaking users.
registry_volume_claim: 'registry-claim'
 
openshift_hosted_manage_router: True

openshift_hosted_router_edits:
- key: spec.strategy.rollingParams.intervalSeconds
  value: 1
  action: put
- key: spec.strategy.rollingParams.updatePeriodSeconds
  value: 1
  action: put
- key: spec.strategy.activeDeadlineSeconds
  value: 21600
  action: put

# handle using openshift_fact values as defaults if they are not set explictly
openshift_hosted_router_namespace: "{{ openshift.hosted.router.namespace | default('default') }}"
openshift_hosted_router_selector: "{{ openshift.hosted.router.selector | default(omit) }}"
openshift_hosted_router_image: "{{ openshift.hosted.router.registryurl }}"
openshift_hosted_router_certificate: "{{ openshift.hosted.router.certificate | default({}) }}"

```
tasks/main.yml:
```
- name: Retrieve list of openshift nodes matching router selector
  oc_obj:
    state: list
    kind: node
    namespace: "{{ openshift_hosted_router_namespace }}"
    selector: "{{ openshift_hosted_router_selector }}"
  register: router_nodes

- include: router/router.yml
  when: openshift_hosted_manage_router | default(true) | bool
  with_items:
  - openshift_hosted_routers
  loop_control:
    loop_var: router_item  # override variable used from item to router_item.
    label: "{{ router.name }}"  # override the label to avoid printing out the full datastructure for each iteration
  vars:
    openshift_hosted_router_replicas: "{{ openshift.hosted.router.replicas | default(None)
                                                                | get_router_replicas(router_nodes) }}"
    openshift_hosted_routers:
    - name: router
      replicas: "{{ openshift_hosted_router_replicas }}"
      namespace: "{{ openshift_hosted_router_namespace | default('default') }}"
      serviceaccount: "{{ openshift_hosted_router_serviceaccount | default('router') }}"
      # This option is not yet implemented
      # force_subdomain: "{{ openshift.hosted.router.force_subdomain | default(none) }}"
      selector: "{{ openshift_hosted_router_selector }}"
      images: "{{ openshift_hosted_router_image }}"
      edits: "{{ openshift_hosted_router_edits }}"
      stats_port: 1936
      ports:
      - 80:80
      - 443:443
      certificates: "{{ openshift_hosted_router_certificate | default({}) }}"
```

tasks/router.yml:
```
---
- name: Get the certificate contents for router
  copy:
    backup: True
    dest: "/etc/origin/master/{{ item | basename }}"
    src: "{{ item }}"
  with_items: "{{ router_item.certificates | oo_select_keys_from_list(['keyfile', 'certfile', 'cafile']) }}"

- name: Create OpenShift router
  oc_adm_router:
    name: "{{ router_item.name }}"
    replicas: "{{ router_item.replicas }}"
    namespace: "{{ router_item.namespace }}"
    # This option is not yet implemented
    # force_subdomain: "{{ router_item.force_subdomain }}"
    service_account: "{{ router_item.serviceaccount }}"
    selector: "{{ router_item.selector }}"
    images: "{{ router_item.images }}"
    cert_file: "{{ cert_file }}"
    key_file: "{{ key_file }}"
    cacert_file: "{{ ca_file }}"
    edits: "{{ openshift_hosted_router_edits | union(router_item.edits)  }}"
    ports: "{{ router_item.ports }}"
    stats_port: "{{ router_item.stats_port }}"
  register: routerout
  vars:
    cert_file_basename: "{{ router_item.certificate.certfile | default('') | basename }}"
    key_file_basename: "{{ router_item.certificate.keyfile | default('') | basename }}"
    ca_file_basename: "{{ router_item.certificate.cafile | default('') | basename }}"
    cert_file: "{{ '/etc/origin/master/' ~ cert_file_basename if cert_file_basename != '' else omit }}"
    key_file: "{{ '/etc/origin/master/' ~ key_file_basename if key_file_basename != '' else omit }}"
    ca_file: "{{ '/etc/origin/master/' ~ ca_file_basename if ca_file_basename != '' else omit }}"

# This should probably move to module
- name: wait for deploy
  pause:
    seconds: 30
  when: routerout.changed

- name: Ensure router replica count matches desired
  oc_scale:
    kind: dc
    name: "{{router_item.name }}"
    namespace: "{{ router_item.namespace }}"
    replicas: "{{ router_item.replicas }}"
```

roles/openshift_hosted/tasks/router/router.yml
```diff
@@ -11,69 +11,34 @@
...
-    copy:
-      content: "{{ openshift.hosted.router.certificate.contents }}"
-      dest: "{{ openshift_master_config_dir }}/openshift-router.pem"
-      mode: 0600
-
-  when: openshift_hosted_router_certificate is defined
+- name: Get the certificate contents for router
+  copy:
+    backup: True
+    dest: "/etc/origin/master/{{ item | basename }}"
```
abutcher: @detiber Are you cool with dropping the router certs in `/etc/origin/master`? I think we could probably land them in a temporary directory that gets deleted after router creation.

detiber: @abutcher I like that idea...  Is there already something short circuiting the upload process if the router already has the matching cert?

abutcher: @detiber I believe we'd still need to copy them up every time to do the comparison within the oc_adm_router role but we shouldn't need to keep the files after.

stevekuznetsov:
[test_install]

roles/openshift_hosted/tasks/router/router.yml
```diff
@@ -11,69 +11,34 @@
...
-      action: put
+    service_account: "{{ item.serviceaccount | default('router') }}"
+    selector: "{{ item.selector | default(none) }}"
+    images: "{{ item.images | default(omit) }}"
+    cert_file: "{{ ('/etc/origin/master/' ~ (item.certificates.certfile | basename)) if 'certfile' in item.certificates else omit }}"
+    key_file: "{{ ('/etc/origin/master/' ~ (item.certificates.keyfile | basename)) if 'keyfile' in item.certificates else omit }}"
+    cacert_file: "{{ ('/etc/origin/master/' ~ (item.certificates.cafile | basename)) if 'cafile' in item.certificates else omit }}"
+    edits: "{{ openshift_hosted_router_edits | union(item.edits)  }}"
+    ports: "{{ item.ports }}"
+    stats_port: "{{ item.stats_port }}"
```
abutcher: Would be nice to default these parameters so that they don't have to be set for each router:

```
    cert_file: "{{ ('/etc/origin/master/' ~ (item.certificates.certfile | basename)) if 'certfile' in item.certificates | default({}) else omit }}"
    key_file: "{{ ('/etc/origin/master/' ~ (item.certificates.keyfile | basename)) if 'keyfile' in item.certificates | default({}) else omit }}"
    cacert_file: "{{ ('/etc/origin/master/' ~ (item.certificates.cafile | basename)) if 'cafile' in item.certificates | default({}) else omit }}"
    edits: "{{ openshift_hosted_router_edits | union(item.edits | default([]))  }}"
    ports: "{{ item.ports | default(["80:80", "443:443"]) }}"
    stats_port: "{{ item.stats_port | default(1936) }}"
```

kwoodson: @abutcher, I agree that would be nice.  The problem lands in that each router can't share the same ports.  The edits need to happen as the oc adm router command doesn't actually put in the correct ports into the environment variables thus the need to edit them preflight.

I'm open to suggestions but as we know, time is of the essence.

sdodson:
aos-ci-test

roles/openshift_hosted/tasks/router/router.yml
detiber: @abutcher seems like it would be better to either:
- Only do the copy if the contents differ from what is already configured
- Pass the certificate contents into oc_adm_router, with the content coming from either a variable or a local file lookup on the control host (my preference here would be to support both).

roles/openshift_hosted/tasks/router/router.yml
detiber: @kwoodson @abutcher port conflicts may not be an issue for many users, since I think most cases of sharding (outside of Ops) has been for using different routers for different security zones. In that case, the routers are hosted on different sets of nodes, so the port conflicts are not an issue.

We could document that the defaults are the same and need to be overridden explicitly if the sharded routers will be co-located on the same sets of nodes.

PR #3550, opened 2017-03-02
Removing the openshift_master_facts dependency

ewolinetz:
Proposed fix for https://bugzilla.redhat.com/show_bug.cgi?id=1425312


ewolinetz:
Still confirming this locally

PR #3551, opened 2017-03-03
Only set ownership to etcd for thirdparty datadir

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1428532


sdodson:
aos-ci-test

PR #3552, opened 2017-03-03
DO NOT MERGE - Excluder fixes

sdodson:


roles/openshift_excluder/tasks/status.yml
```diff
@@ -34,25 +34,37 @@
...
 - name: Record docker excluder status
-  command: "{{ openshift.common.service_type }}-docker-excluder"
+  command: "{{ openshift.common.service_type }}-docker-excluder status"
   register: docker_excluder_status
   when:
   - "{{ docker_excluder_installed.installed_versions | default([]) | length > 0 }}"
   - not openshift.common.is_containerized | bool
   failed_when: false
 
+# Unfortunately rpm_q doesn't return installed_versions if there aren't any installed
```
detiber: Have we shipped a version of the excluder with the status command yet? If not, I think it would be better to handle this in the excluder rather than working around it here.

roles/openshift_excluder/tasks/status.yml
```diff
@@ -34,25 +34,37 @@
...
+# # atomic-openshift-excluder status ; echo $?
+# unexclude -- At least one package not excluded
+# 1
+
 - name: Set excluder status facts
   set_fact:
-    docker_excluder_enabled: "{{ 'false' if docker_excluder_status.rc | default(0) == 0 or docker_excluder_installed.installed_versions | default(0) | length == 0 else 'true' }}"
-    openshift_excluder_enabled: "{{ 'false' if docker_excluder_status.rc | default(0) == 0 or openshift_excluder_installed.installed_versions | default(0) | length == 0 else 'true' }}"
+    docker_excluder_enabled: "{{ 'false' if docker_excluder_status.rc | default(0) != 0 or docker_excluder_installed.installed_versions | default(0) | length == 0 else 'true' }}"
+    openshift_excluder_enabled: "{{ 'false' if docker_excluder_status.rc | default(0) != 0 or openshift_excluder_installed.installed_versions | default(0) | length == 0 else 'true' }}"
```
ingvagabund: Still confused why the vars are named ``docker|openshift_excluder_enabled``. Why not ``docker|openshift_excluder_status_ok`` or similar?

roles/openshift_excluder/tasks/status.yml
```diff
@@ -34,25 +34,37 @@
   - not openshift.common.is_containerized | bool
 
 - name: Record excluder status
-  command: "{{ openshift.common.service_type }}-excluder"
+  command: "{{ openshift.common.service_type }}-excluder status"
```
ingvagabund: Given the latest 3.4 excluder has the status command, this is an improvement.

roles/openshift_excluder/tasks/status.yml
sdodson: I think all current excluders include status as of a week ago.

roles/openshift_excluder/tasks/status.yml
sdodson: Whatever you think is best is fine with me, I just picked this because it's the computed state whether enabled or disabled based on the package being installed and not having been disabled.

sdodson: Is it the excluder's use of "enabled" and "disabled" what's confusing the situation? It's certainly confused me, maybe we should avoid saying 'eanbled' or 'disabled'?

```
[root@ose3-master ~]# atomic-openshift-excluder 
Unknown Option: 

Usage: /usr/sbin/atomic-openshift-excluder <exclude | unexclude>

exclude | disable
  exclude packages via an exclude= line in /etc/yum.conf
  This disables updates for atomic-openshift atomic-openshift-clients atomic-openshift-clients-redistributable atomic-openshift-dockerregistry atomic-openshift-master atomic-openshift-node atomic-openshift-pod atomic-openshift-recycle atomic-openshift-sdn-ovs atomic-openshift-tests tuned-profiles-atomic-openshift-node packages

unexclude | enable
  removes atomic-openshift atomic-openshift-clients atomic-openshift-clients-redistributable atomic-openshift-dockerregistry atomic-openshift-master atomic-openshift-node atomic-openshift-pod atomic-openshift-recycle atomic-openshift-sdn-ovs atomic-openshift-tests tuned-profiles-atomic-openshift-node from the exclude= line in /etc/yum.conf
  This re-enables updates for packages

status
  returns status either 'exclude' or 'unexclude' and 0 or 1 respectively
```

ingvagabund: I understand the concept that if there is at least one package that should be excluded and is not the excluder can not work properly which means it is not fully enabled. So it is disabled. Still, I find the reasoning quite confusing for the first time. Now it make sense, I would just add additional commend saying why an excluder is not enabled when the RC of the status is non-zero.

ingvagabund: Just wandering what should happend when there is one package missing in the yum's exclude list. Since it leads to an excluder not being enabled. That means during the upgrade there is no guarantee docker or other packages get updated even if they should be. Since if you happen to be updating from 3.5 to 3.6 and you would need newer docker version (for some reason), the docker does not have to get updated (since it can be in the exclude list)). Which may lead to a broken node.

E.g. the status is non-zero since there is ``docker*1.20*`` missing in the exclude list. So the docker excluder is understood is disabled. So the docker does not get updated even if it should. Since we decided we will run 3.6 with docker-1.13. Now we have 3.6 installed but the docker is still 1.12. Plus, assume the node/master component is out of the exclude list as well.

detiber: Ideally, just the return code should matter for tooling purposes.

For user consumption, I would like to see something like:
```
OpenShift Excluder
Status: Enabled
Version: <excluder version>

Rules
========
<list of rules>

Managed packages
========
<list of managed packages in yum/dnf list installed format>
```

Alternatively, I think a yaml could be used as a compromise for user-readable and machine parseable.
```
name: openshift_excluder
enabled: True
version: <excluder version>
rules:
- <rule def>
managed_packages:
- pkg_1:
    installed: True
    version: <version>
```

ingvagabund: Worth putting both ``docker_excluder_enabled`` and ``openshift_excluder_enabled`` to default variables set to False be default. And change to True only if the status is zero. Plus, put the status check under a condition where we are sure an excluder is installed.

ingvagabund: Currently, both excluders are run in pairs. I don't mind breaking that coupling and running general excluder first. Then run the docker one. This way, we can put all the tasks into a block. Or even better, use ``with-items`` if it is possible.

sdodson: Oh, that's the other thing, I intended for these two facts to be used only for determining whether or not the excluders should be re-enabled after the upgrade is complete. I'd expect the upgrade playbooks to disable {{ openshift.common.service_type}}-excluder no matter what and i guess we need to ensure that the upgrade playbooks only upgrade you to the version that they're intended to upgrade you to. ie: playbooks/byo/openshift-cluster/upgrades/v3_5/upgrade.yml should ensure that it doesn't upgrade you to anything but the latest 3.5.

The docker excluder in particular I feel we should leave enabled no matter what. The upgrade playbooks need to update to the correct docker excluder version which will set the proper excludes line and then we perform the actual docker upgrade.

sdodson: Maybe we should change the two facts to be `openshift_excluder_post_state: (exclude|unexclude)` so rather than true or false we're clearly defining the state we'll set them back to?

PR #3554, opened 2017-03-03
openshift_facts: install python3-dbus package on Fedora nodes

### LINKED ISSUE #3544: Install python3-dbus package instead of python-dbus when Python3 is used
I'm using Fedora on my host and Fedora on the managed hosts. I'm using Python3 (see #3543).

Installation fails with:
```
TASK [openshift_facts : Gather Cluster facts and set is_containerized if needed] ***
fatal: [ec2-xx-xxx-xx-xx.compute-1.amazonaws.com]: FAILED! => {
	"changed": false, 
	"failed": true
}

MSG:

This module requires dbus python bindings
```

It turns out (thank you @rhcarvalho!) that installer installs `python-dbus` package on the nodes while it should install `python3-dbus` package in case of using Python3.

php-coder:
Fixes #3544 

PTAL @sdodson

### DIFF
--- a/roles/openshift_facts/vars/Fedora.yml
+++ b/roles/openshift_facts/vars/Fedora.yml
@@ -1,0 +1,6 @@
+---
+required_packages:
+  - iproute
+  - python3-dbus
+  - PyYAML
+  - yum-utils
--- a/roles/openshift_facts/tasks/main.yml
+++ b/roles/openshift_facts/tasks/main.yml
@@ -47,6 +47,12 @@
 
   when: l_is_atomic | bool
 
+- name: Load variables
+  include_vars: "{{ item }}"
+  with_first_found:
+  - "{{ ansible_distribution }}.yml"
+  - "default.yml"
+
 - name: Ensure various deps are installed
   package: name={{ item }} state=present
   with_items: "{{ required_packages }}"
--- a/roles/openshift_facts/vars/default.yml
+++ b/roles/openshift_facts/vars/default.yml
@@ -1,0 +1,6 @@
+---
+required_packages:
+  - iproute
+  - python-dbus
+  - PyYAML
+  - yum-utils
--- a/roles/openshift_facts/vars/main.yml
+++ b/roles/openshift_facts/vars/main.yml
@@ -1,10 +1,4 @@
 ---
-required_packages:
-  - iproute
-  - python-dbus
-  - PyYAML
-  - yum-utils
-
 required_system_containers_packages:
   - atomic
   - ostree


roles/openshift_facts/vars/Fedora.yml
```diff
@@ -0,0 +1,7 @@
+---
+required_packages:
+  - iproute
+  - python3-dbus
+  - python-six
```
php-coder: I suspect that here we want to have `python3-six` package but it works for me without this. Perhaps, because I used non-minimal Fedora template.

roles/openshift_facts/vars/Fedora.yml
```diff
@@ -0,0 +1,7 @@
...
+  - PyYAML
+  - yum-utils
```
php-coder: I faced yet another problem with this package: when `yum-utils` is already installed all the process failed with:
```
TASK [openshift_facts : Ensure various deps are installed] *********************
failed: [ec2-54-83-174-64.compute-1.amazonaws.com] (item=[u'iproute', u'python3-dbus', u'python-six', u'PyYAML', u'yum-utils']) => {
    "failed": true, 
    "item": [
        "iproute", 
        "python3-dbus", 
        "python-six", 
        "PyYAML", 
        "yum-utils"
    ], 
    "module_stderr": "Package iproute-4.4.0-3.fc24.x86_64 is already installed, skipping.\nPackage python3-dbus-1.2.4-1.fc24.x86_64 is already installed, skipping.\nPackage python-six-1.10.0-2.fc24.noarch is already installed, skipping.\nPackage PyYAML-3.11-12.fc24.x86_64 is already installed, skipping.\nTraceback (most recent call last):\n  File \"/tmp/ansible_br2rb44o/ansible_module_dnf.py\", line 375, in <module>\n    main()\n  File \"/tmp/ansible_br2rb44o/ansible_module_dnf.py\", line 369, in main\n    ensure(module, base, params['state'], params['name'])\n  File \"/tmp/ansible_br2rb44o/ansible_module_dnf.py\", line 324, in ensure\n    base.do_transaction()\n  File \"/usr/lib/python3.5/site-packages/dnf/base.py\", line 591, in do_transaction\n    self._trans_error_summary(errstring))\ndnf.exceptions.Error: Transaction check error:\n  package yum-utils-1.1.31-509.fc24.noarch is already installed\n\nError Summary\n-------------\n\n", 
    "module_stdout": ""
}

MSG:

MODULE FAILURE
```
To overcome this I had to remove `yum-utils` from this list.

php-coder:
PTAL @sdodson

sdodson:
[merge]

stevekuznetsov:
FYI you'll continue to see the merge tag fail here until you run the other dependent jobs

sdodson:
aos-ci-test

sdodson:
[merge]

stevekuznetsov:
Stand down, rosie

sdodson:
[merge]

roles/openshift_facts/vars/Fedora.yml
detiber: I think if we are still using yum-utils on a Fedora host, then it should be considered a bug.

stevekuznetsov:
We've not seen that many tests go red all at once unless it is legitimate, also worrying that they're all in the same category. Should be looked at.

php-coder:
> We've not seen that many tests go red all at once unless it is legitimate

These tests expect that there no other namespaces with prefix `e2e-tests` except their own. They're failing because there is such a namespace (`e2e-tests-downward-api-g2t0z` in this case). I couldn't find which test is creating it (I suspect that it's `TestMakeEnvironmentVariables` from `vendor/k8s.io/kubernetes/pkg/kubelet/kubelet_pods_test.go` but I couldn't prove it).

I have 2 theories:
1) tests are running in parallel and at the execution time there were two e2e tests running
2) if tests are executing in random order, it's possible that there was a test that didn't cleanup after itself and it affected these tests

stevekuznetsov:
You should also investigate why we don't see it in Origin, it's the same job logic that handles both repositories.

php-coder:
Let's try to merge yet another time and if it will fail, I give up with it.

php-coder:
@sdodson PTAL

php-coder:
@sdodson ping

php-coder:
@sdodson Sorry for disturbing you but could you merge it?

sdodson:
aos-ci-test

sdodson:
[merge]

php-coder:
@sdodson Thanks!!

PR #3555, opened 2017-03-03
Create ephemeral template for CloudForms 

Jooho:
Description :
  In order to deploy CloudForms on OpenShift with official template, PV is necessary. However, just for testing, PV is a little much preparation. Therefore, I create 'ephemeral template'  that deploy CF without PV.

Changes:
    Name change : cfme-template.yaml   ==> cfme-template-persistent.yaml
    New file : cfme-template.yaml


michaelgugino:
Closing due to inactivity.

PR #3558, opened 2017-03-03
Adding some more logging and metrics defaults

ewolinetz:
Still need to go through the metrics defaults from the previous role and map those to the new role
FYI @sdodson @jcantrill

### DIFF
--- a/roles/openshift_logging/tasks/install_elasticsearch.yaml
+++ b/roles/openshift_logging/tasks/install_elasticsearch.yaml
@@ -32,6 +32,10 @@
   with_sequence: count={{ openshift_logging_es_cluster_size | int - openshift_logging_current_es_size | int }}
   check_mode: no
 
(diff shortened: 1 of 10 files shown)


roles/openshift_logging/tasks/install_elasticsearch.yaml
```diff
@@ -31,6 +32,10 @@
   with_sequence: count={{ openshift_logging_es_cluster_size | int - openshift_logging_current_es_size | int }}
   check_mode: no
 
+- set_fact: nodeselector="{{openshift_logging_es_nodeselector | default({}) }}"
```
ewolinetz: this was a debug statement -- need to remove

roles/openshift_logging/tasks/install_elasticsearch.yaml
```diff
@@ -31,6 +32,10 @@
...
+
+- debug: var=nodeselector
```
ewolinetz: same here

Issue #3559, opened 2017-03-04
Adding roles/lib_openshift/library to ansible_library breaks ansible-doc -l


alvaroaleman:
#### Description

Adding `roles/lib_openshift/library` to ansible_library breaks `ansible-doc -l`

##### Version

openshift-ansible 6ecb86b2fcc36e3383d86395d3be0a443e12981e
ansible 2.2.0.0

##### Steps To Reproduce
1. git clone git@github.com:openshift/openshift-ansible.git
2. export ANSIBLE_LIBRARY=$(pwd)/openshift-ansible/roles/lib_openshift/library
3. ansible-doc -l

##### Expected Results

A list of all available modules

##### Observed Results

```
 [ERROR]: unable to parse /tmp/openshift-ansible/roles/lib_openshift/library/oc_adm_router.py

 [ERROR]: unable to parse /tmp/openshift-ansible/roles/lib_openshift/library/oc_env.py

ERROR! module oc_env has a documentation error formatting or is missing documentation

```

detiber:
@alvaroaleman thanks for reporting this issue, I've created #3560 to address it.

PR #3560, opened 2017-03-05
fixup errors when running ansible-docs against lib_openshift

detiber:
Resolves: #3559

### DIFF
--- a/roles/lib_openshift/library/oc_adm_router.py
+++ b/roles/lib_openshift/library/oc_adm_router.py
@@ -98,8 +98,8 @@ options:
   images:
     description:
     - The image to base this router on - ${component} will be replaced with --type
-    required: 'openshift3/ose-${component}:${version}'
-    default: None
+    required: false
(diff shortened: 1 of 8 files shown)


roles/lib_openshift/library/oc_adm_router.py
```diff
@@ -205,31 +205,31 @@
     aliases: []
 author:
 - "Kenny Woodson <kwoodson@redhat.com>"
-extends_documentation_fragment:
-- There are some exceptions to note when doing the idempotency in this module.
```
detiber: @kwoodson I wasn't quite sure what to do with this content. I tried adding it to the description, but the description field is automatically formatted and wasn't playing nice with the current way of listing the exceptions.

`extends_documentation_fragment` is used for adding additional documentation to an existing doc, not for providing additional content for the current doc.

michaelgugino:
Closing due to inactivity.

PR #3565, opened 2017-03-06
Preserve order of Docker registries

EricMountain-1A:
List of additional/blocked/insecure Docker registries was being converted
to a set, thereby dropping the original order.  Fixed to preserve order
of the registries (this allows prioritising registries).

### DIFF
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -2319,14 +2319,19 @@ class OpenShiftFacts(object):
                                       protected_facts_to_overwrite)
 
         if 'docker' in new_local_facts:
-            # remove duplicate and empty strings from registry lists
+            # remove duplicate and empty strings from registry lists, preserving order
             for cat in ['additional', 'blocked', 'insecure']:
                 key = '{0}_registries'.format(cat)
                 if key in new_local_facts['docker']:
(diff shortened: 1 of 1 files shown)


roles/openshift_facts/library/openshift_facts.py
```diff
@@ -2303,14 +2303,15 @@ def init_local_facts(self, facts=None,
...
+            # remove duplicate and empty strings from registry lists, preserving order
             for cat in ['additional', 'blocked', 'insecure']:
                 key = '{0}_registries'.format(cat)
                 if key in new_local_facts['docker']:
                     val = new_local_facts['docker'][key]
                     if isinstance(val, string_types):
                         val = [x.strip() for x in val.split(',')]
-                    new_local_facts['docker'][key] = list(set(val) - set(['']))
+                    seen = set()
+                    new_local_facts['docker'][key] = [r for r in val if r not in seen and r != '' and (seen.add(r) or True)]
```
detiber: This line is causing the travis test failure.

I think it would be better to use a loop rather than a list comprehension here to improve readability/grokability.

detiber:
@EricMountain-1A outside of the test failure, lgtm

EricMountain-1A:
@detiber changed for a loop as requested.  Travis seems happier ;-)

detiber:
aos-ci-test

PR #3566, opened 2017-03-06
separate out test tool configs from setup.cfg

detiber:
Since we are moving away from setuptools for invoking tests, lets
move the configs for the different test tools into their own configs.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+exclude=.tox,utils,inventory
+max_line_length = 120
+ignore = E501,T003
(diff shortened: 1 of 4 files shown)


.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+exclude=.tox,,utils,inventory
```
rhcarvalho: I noticed the pattern has changed, aren't the `/*` suffixes needed?

detiber: They are not, and neither is the extraneous comma. I'll fix taht.

detiber:
aos-ci-test

PR #3567, opened 2017-03-06
logging needs openshift_master_facts before openshift_facts

richm:
@jcantrill @ewolinetz PTAL
aos-ci-test


roles/openshift_logging/meta/main.yaml
```diff
@@ -13,4 +13,5 @@ galaxy_info:
   - cloud
 dependencies:
 - role: lib_openshift
+- role: openshift_master_facts
 - role: openshift_facts
```
detiber: openshift_master_facts will pull in openshift_facts, no need to list both.

PR #3571, opened 2017-03-06
[WIP] Simplify test code base by leveraging pytest

rhcarvalho:
Replace usage of `unittest` and other forms to unify the test code base into a slimmer form.

**NOTE**: not ready for review yet, needs clean up.

### DIFF
--- a/roles/lib_openshift/src/test/unit/test_oc_adm_manage_node.py
+++ b/roles/lib_openshift/src/test/unit/test_oc_adm_manage_node.py
@@ -103,9 +103,9 @@ class ManageNodeTest(unittest.TestCase):
         results = ManageNode.run_ansible(params, False)
 
         # returned a single node
-        self.assertTrue(len(results['results']['nodes']) == 1)
+        assert len(results['results']['nodes']) == 1
         # returned 2 pods
-        self.assertTrue(len(results['results']['nodes']['ip-172-31-49-140.ec2.internal']) == 2)
+        assert len(results['results']['nodes']['ip-172-31-49-140.ec2.internal']) == 2
 
     @mock.patch('oc_adm_manage_node.Utils.create_tmpfile_copy')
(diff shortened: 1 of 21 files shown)


roles/lib_openshift/src/test/unit/test_oadm_manage_node.py
```diff
@@ -176,11 +166,11 @@ def test_schedulable_false(self, mock_openshift_cmd, mock_tmpfile_copy):
...
-        self.assertEqual(results['results']['nodes'][0]['schedulable'], False)
+        assert results['changed']
+        assert results['results']['nodes'][0]['name'] == 'ip-172-31-49-140.ec2.internal'
+        assert results['results']['nodes'][0]['schedulable'] == False
 
-    @unittest.skipIf(six.PY3, 'py2 test only')
+    @pytest.mark.skipif(six.PY3, reason='py2 test only')
     @mock.patch('os.path.exists')
     @mock.patch('os.environ.get')
     def test_binary_lookup_fallback(self, mock_env_get, mock_path_exists):
```
detiber: I have some fixes lined up for this, it turns the binary lookup into a common fixture for all the files and uses a parametrized test to reduce at least some of the duplication.

stevekuznetsov:
Why prefer bare `assert` over the more specialized ones?

detiber:
> Why prefer bare assert over the more specialized ones?

The 'specialized' ones are provided by nose/unitTest. pytest provides assert introspection to provide similar functionality: http://doc.pytest.org/en/latest/assert.html

rhcarvalho:
> Why prefer bare assert over the more specialized ones?

Because we are using pytest, and it can give good error messages with bare assert using introspection:

http://docs.pytest.org/en/latest/

rhcarvalho:
No time to work on this, low priority, closing.

PR #3572, opened 2017-03-06
do not need openshift_facts with openshift_master_facts

richm:
Do not need to use openshift_facts if using openshift_master_facts
openshift_master_facts will pull in openshift_facts, no need to list both.
@detiber @ewolinetz @jcantrill PTAL
aos-ci-test

### DIFF
--- a/roles/openshift_logging/meta/main.yaml
+++ b/roles/openshift_logging/meta/main.yaml
@@ -14,4 +14,3 @@ galaxy_info:
 dependencies:
 - role: lib_openshift
 - role: openshift_master_facts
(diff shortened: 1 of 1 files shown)


ewolinetz:
IIRC there was an issue with getting the correct api_port if we didn't use openshift_facts (openshift cluster facts pulled in a different value than openshift master facts)

richm:
@ewolinetz do we still need this PR?  Should I just close it since #3644 was merged?

ewolinetz:
@richm i think we can close this, yes

PR #3576, opened 2017-03-07
test fixes

detiber:
- update test_oc_serviceaccount_secret to handle both pyyaml and ruamel.yaml
- update test_oc_objectvalidator to work when ~/bin/oc exists on the
  test system.


rhcarvalho:
There is again a rounding problem on coverage:

> FAIL Required test coverage of 29% not reached. Total coverage: 28.90%

https://travis-ci.org/openshift/openshift-ansible/jobs/208479151 (py27)

> FAIL Required test coverage of 29% not reached. Total coverage: 28.71%

https://travis-ci.org/openshift/openshift-ansible/jobs/208479152 (py35)

@detiber I think we may need to give in a little on the tight coverage requirements to allow for refactoring?
Or is this a case that we're indeed reducing coverage? At first sight I wouldn't expect.

rhcarvalho:
Hmmm 3h ago it was like over 1AM over there? Committing from bed :sleeping_bed: ?
;)

roles/lib_openshift/src/test/unit/test_oc_serviceaccount_secret.py
```diff
@@ -1,414 +1,199 @@
...
+    which_result = binary_lookup_test_data['which_result']
+    binary_expected = binary_lookup_test_data['binary_expected']
+    binaries = binary_lookup_test_data['binaries']
+
+    mocker.patch('os.environ.get', side_effect=lambda _v, _d: path)
+
+    if six.PY2:
+        mocker.patch('os.path.exists', side_effect=lambda f: f in binaries)
+    else:
+        mocker.patch('shutil.which', side_effect=lambda _f, path=None: which_result)
```
rhcarvalho: I tend to not like much this type of test. It is highly coupled to the implementation.
We are in fact testing that monkey-patching works and that a certain function calls certain patched functions/methods.

Changing the implementation goes hand-in-hand with changing the tests.

rhcarvalho: In this particular case, both `os.environ.get` and `os.path.exists` are only needed in PY2.

I'd suggest the simplification to always patch everything, without the distinction of Python version. `shutil.which` is an exception because it is not defined in PY2:

```python
def test_binary_lookup(binary_lookup_test_data, mocker):
    path = binary_lookup_test_data['path']
    binaries = binary_lookup_test_data['binaries']
    which_result = binary_lookup_test_data['which_result']
    binary_expected = binary_lookup_test_data['binary_expected']

    mocker.patch('os.environ.get', side_effect=lambda _v, _d: path)
    mocker.patch('os.path.exists', side_effect=lambda f: f in binaries)

    if six.PY3:
        mocker.patch('shutil.which', side_effect=lambda _f, path=None: which_result)

    assert oc_serviceaccount_secret.locate_oc_binary() == binary_expected
```

roles/lib_openshift/src/test/unit/test_oc_serviceaccount_secret.py
```diff
@@ -1,414 +1,199 @@
...
-        ''' Testing adding a secret to a service account '''
-
-        # Arrange
-
-        # run_ansible input parameters
-        params = {
-            'state': 'present',
+MODULE_PATH = os.path.realpath(os.path.join(os.path.dirname(__file__), os.pardir,
+                                            os.pardir, os.pardir, 'library'))
+sys.path.insert(1, MODULE_PATH)
```
rhcarvalho: We need to spread the knowledge in the team to review this during code reviews, avoid using `'/'` etc. :+1:

roles/lib_openshift/src/test/unit/test_oc_objectvalidator.py
```diff
@@ -4,900 +4,838 @@
...
+    error_results = {
+        'returncode': 1,
+        'stderr': 'Error.',
+        'stdout': '',
+        'cmd': 'oc get hostsubnet -o json -n default',
+        'results': [{}]
+    }
+
+    # Assert
+    assert results['failed'] is True
```
rhcarvalho:
```python
assert results['failed']
```

? Or do we need to check specifically for `True`?

Here and other 2 places.

detiber:
> @detiber I think we may need to give in a little on the tight coverage requirements to allow for refactoring?

I agree, I've actually thought about removing the test failure on coverage % on these tests and move that into coveralls where it can be configured more flexibly (fail on loss of >= % of coverage).

> Or is this a case that we're indeed reducing coverage? At first sight I wouldn't expect.

I think it dropped slightly, but haven't dug in yet. I plan on doing it during my next flight.

> Hmmm 3h ago it was like over 1AM over there? Committing from bed 🛌 ?

Hah, I've been doing that too much lately. In this case, I wanted to get it up for review before the start of the day. Especially since I'm travelling today.

detiber:
> This is quite big for a detailed review, @detiber anything you'd like another pair of eyes in particular?

General approach mostly, also wanted to give you a heads up to avoid stepping on each other's toes, since we are both working in the same area.

roles/lib_openshift/src/test/unit/test_oc_serviceaccount_secret.py
detiber: :+1:

detiber: > We are in fact testing that monkey-patching works and that a certain function calls certain patched functions/methods.

Definitely something I'll take a look at addressing. I didn't update the tests, mainly adding the mock for the oc binary and switching to native pytest.

rhcarvalho:
In terms of approach, I'm not a big fan of the heavy mocking and monkey-patching school.
Often that hides a way of writing code that is more straight-forward to test, but I wouldn't argue on that on a PR ;)

In general, we need to keep in mind:
- We can have green tests, 100% coverage, but still failures in production.
- When tests and code are tightly coupled, reworking the code forces us to change the tests, and that somehow reduces the effectiveness of our safety-net against regressions.

lib_utils and lib_openshift seems to be the next big portion of code in this repo, so we need to keep an eye on having good examples for people to follow.

I do like the simplifications, removal of 200+ LOC, better parametrization of tests with pytest :+1:

detiber:
@rhcarvalho tweaked the mocking a bit and started to push some common setup into common fixtures.

I also added some trickery for common tests across all modules.

roles/lib_openshift/src/test/unit/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+    run_cmd = mocker.patch(patch_method)
+    yield run_cmd
+
+
+@pytest.fixture(params=['PyYAML', 'ruamel.yaml'])
+def yaml_provider(request, monkeypatch):
+    if request.param == 'PyYAML':
+        import yaml as yaml_module
+    else:
+        import ruamel.yaml as yaml_module
```
rhcarvalho: https://docs.python.org/2/library/importlib.html#importlib.import_module ?

```python
# params = ['yaml', 'ruamel.yaml']
module = importlib.import_module(request.param)

mokeypatch.setattr(..., module)
return module
```

roles/lib_openshift/src/test/unit/test_oc_scale.py
```diff
@@ -1,256 +1,110 @@
...
+        }
+    }
+
+    mock_run_cmd.side_effect = [
+        (0, json.dumps(dc), '')
+    ]
+
+    results = CLASS_UNDER_TEST.run_ansible(params, False)
+
+    assert results['changed'] is False
```
rhcarvalho: In some places we say

```python
assert results['changed']
```

So perhaps for symmetry we could replace `... is False` with `not ...`?

```python
assert not results['changed']
```

?

Not a big deal, though. What do you think?

roles/lib_openshift/library/__init__.py
```diff
@@ -0,0 +1,24 @@
+''' provide a way to discover modules for tests '''
```
detiber: I'll test for side effects, but there was nothing glaringly obvious.

roles/lib_openshift/library/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    'oc_obj',
+    'oc_process',
+    'oc_project',
+    'oc_route',
+    'oc_scale',
+    'oc_secret',
+    'oc_serviceaccount',
+    'oc_serviceaccount_secret',
+    'oc_service',
+    'oc_version'
```
detiber: I'd like to handle it through introspection a bit better, this was a stopgap.

roles/lib_openshift/src/lib/base.py
```diff
@@ -264,7 +264,7 @@ def openshift_cmd(self, cmd, oadm=False, output=False, output_type='json', input
 
         rval = {}
         results = ''
-        err = None
+        error = None
```
detiber: Yes, if the exception code path is hit, the exception uses err as the variable name for the exception and tries to set err in the method scope.

roles/lib_openshift/src/test/unit/conftest.py
```diff
@@ -0,0 +1,55 @@
+import inspect
+
+import pytest
+
+from lib_openshift.library import *  # noqa: F401,F403
```
detiber: :+1: I'll take a look at updating, I rabbit holing down that path and went this route out of convenience.

roles/lib_openshift/src/test/unit/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+                        ids.append(key)
+        metafunc.parametrize('binary_lookup_module', lookup_modules, ids=ids)
+
+
+@pytest.fixture
+def mock_tmpfile_copy(request, mocker):
+    module_under_test = request.module.MODULE_UNDER_TEST
+    patch_method = module_under_test.__name__ + '.Utils.create_tmpfile_copy'
+    mocker.patch(patch_method, return_value='/tmp/mocked_kubeconfig')
+    yield
```
detiber: I don't think it's needed anymore, I was using the context manager from mocker originally and left the yield after it was refactored.

roles/lib_openshift/src/test/unit/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+@pytest.fixture
+def mock_locate_binary(request, mocker):
+    module_name = request.module.MODULE_UNDER_TEST.__name__
+    patch_method = module_name + '.locate_oc_binary'
+    mocker.patch(patch_method, return_value='oc')
+    yield
+
+
+@pytest.fixture
+def mock_run_cmd(request, mocker, mock_locate_binary, mock_tmpfile_copy):
```
detiber: I tried to standardize on mocking the _run method, since it simply passes in provided arguments to subprocess and returning the value.

We could instead mock/monkeypatch the subprocess calls for the modules and then it would get us out of mocking methods we've written. If we go that route, then we could return back mocked data based on the call coming in instead of the order of the call.

I think you are right about getting too clever with trying to put this in contest, my main goal was to get standardization on mocking at a lower level than the openshift_cmd method.

roles/lib_openshift/src/test/unit/test_oc_scale.py
detiber: I tried using `not` and I think being a bit more explicit with `is False` reads a bit better. It also prevents explicit type conversion. I think I might update the positive changed tests to be more explicit for that reason as well.

roles/lib_openshift/src/test/unit/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+
+from lib_openshift.library import *  # noqa: F401,F403
+
+
+def pytest_generate_tests(metafunc):
+    if 'binary_lookup_module' in metafunc.fixturenames:
+        lookup_modules = []
+        ids = []
+        for key in list(globals().keys()):
+            value = globals()[key]
```
detiber: I like that approach, we'd still need a hack like this to show coverage of the "Binary locator" until we can require Ansible 2.3 and leverage module_utils.

detiber:
As an aside, the reduction in test coverage is because these changes brought in files that previously weren't being tested at all through unit tests.

roles/lib_openshift/src/test/unit/conftest.py
detiber: @ashcrow any suggestions for implementing?

detiber:
@mtnbikenc changes from https://github.com/openshift/openshift-ansible/pull/3626 have been pulled in here.

roles/lib_openshift/src/test/unit/conftest.py
ashcrow: I agree. ```*``` imports make me worry. If we truly don't know what is being imported then I agree with generating ``__all__`` as @rhcarvalho noted. If we do know what we are importing I'd rather see this explicitly import.

As a side note unittests generally should have these attributes:

- Fast
- Reliable
- Easily readable

I'm a bit worried that things such as [this clever bit of code](https://github.com/detiber/openshift-ansible/blob/32f826c860ef6bb0cf692053e59b607b8fcd66dc/roles/lib_openshift/src/test/unit/conftest.py#L12-L19) could make it hard for someone to debug why a test has failed. But I have not done a review on all of the changes so I I could be off base.

detiber: @ashcrow indeed, I'm trying to strike a balance between showing accurate coverage for the generated code without duplicating tests. I'll try to massage it up a bit more.

mtnbikenc:

PR #3579, opened 2017-03-07
[WIP] add excluder check

sosiouxme:
Adds a preflight check that the user has installed the appropriate excluders on all the hosts.

### DIFF
--- a/roles/openshift_health_checker/library/aos_version.py
+++ b/roles/openshift_health_checker/library/aos_version.py
@@ -47,8 +47,6 @@ def main():  # pylint: disable=missing-docstring,too-many-branches
 
     # determine what level of precision we're expecting for the version
     expected_version = module.params['version']
-    if expected_version.startswith('v'):  # v3.3 => 3.3
-        expected_version = expected_version[1:]
     num_dots = expected_version.count('.')
 
     pkgs_by_name_version = {}
--- a/roles/openshift_health_checker/openshift_checks/__init__.py
+++ b/roles/openshift_health_checker/openshift_checks/__init__.py
@@ -72,6 +72,16 @@ def get_var(task_vars, *keys, **kwargs):
         raise OpenShiftCheckException("'{}' is undefined".format(".".join(map(str, keys))))
     return value
 
+def normalized_release(task_vars):
+    """Helper function to normalize the release format to dotted numbers"""
+    release = str(get_var(task_vars, "openshift_release"))
+    if release.startswith('v'):
+        release = release[1:]  # v3.3 => 3.3
+    return release
+
+def normalized_minor_release(task_vars):
+    """Helper function to normalize the release format into x.y minor release"""
+    return '.'.join(normalized_release(task_vars).split('.')[:2])
 
 # Dynamically import all submodules for the side effect of loading checks.
 
--- a/roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
+++ b/roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
@@ -82,7 +82,15 @@ def _format_failure(failure):
         (u'Message', stringc(msg, C.COLOR_ERROR)),
     )
     if 'checks' in result._result:
-        rows += ((u'Details', stringc(pformat(result._result['checks']), C.COLOR_ERROR)),)
+        failed_check_msgs = ""
+        for check, body in result._result['checks'].items():
+            if 'failed' in body and body['failed']: # only show the failed checks
+                msg = body['msg'] if 'msg' in body else "Failed without returning a message"
+                failed_check_msgs += '\ncheck "%s":\n%s\n' % (check, msg)
+        if failed_check_msgs:
+            rows += ((u'Details', stringc(failed_check_msgs, C.COLOR_ERROR)),)
+        else: # something failed but no checks will admit to it, so dump everything
+            rows += ((u'Details', stringc(pformat(result._result['checks']), C.COLOR_ERROR)),)
     row_format = '{:10}{}'
     return [row_format.format(header + u':', body) for header, body in rows]
 
--- a/roles/openshift_health_checker/library/excluder_usage.py
+++ b/roles/openshift_health_checker/library/excluder_usage.py
@@ -55,7 +55,7 @@ def main():
     if len(errors) > 0: bail(errors)
 
     # test if docker-excluder status returns non-zero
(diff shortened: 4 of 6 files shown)


roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -72,6 +72,16 @@ def get_var(task_vars, *keys, **kwargs):
         raise OpenShiftCheckException("'{}' is undefined".format(".".join(map(str, keys))))
     return value
 
+def normalized_release(task_vars):
```
sosiouxme: @rhcarvalho is there a reason not to just make these (including get_var) static methods on OpenShiftCheck? Then we wouldn't need to import them explicitly...

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -47,8 +47,6 @@ def bail(error):  # pylint: disable=missing-docstring
 
     # determine what level of precision we're expecting for the version
     expected_version = module.params['version']
-    if expected_version.startswith('v'):  # v3.3 => 3.3
-        expected_version = expected_version[1:]
```
rhcarvalho: Missing a call to `normalize_release`?

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: When I added `get_var` I didn't have in mind it would be used in every check, but it turns out to be useful and popular.
Making it a static method of OpenShiftCheck would work for me, though we need to see what is more convenient / less inconvenient.

If you make that change, new code (unless we do something unconventional) would need to write `self.get_var` instead of `get_var`. Not sure what I like the least, the import or `self.`.

Having more of these utilities that are used often could justify making them static methods to simplify imports.

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -72,6 +72,16 @@ def get_var(task_vars, *keys, **kwargs):
...
+    """Helper function to normalize the release format to dotted numbers"""
+    release = str(get_var(task_vars, "openshift_release"))
+    if release.startswith('v'):
+        release = release[1:]  # v3.3 => 3.3
+    return release
+
+def normalized_minor_release(task_vars):
```
rhcarvalho: `normalized_minor_release` and `normalized_release` could take just a release string, not `task_vars`.
It becomes easier to compose and test when functions/objects know as much as they need to know, not more.

rhcarvalho: Though simple, we could add unit tests for these utilities, specially to document what is our behavior on corner cases like bad input, or variations of good input.

roles/openshift_health_checker/library/aos_version.py
sosiouxme: nope, it gets passed in normalized

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: I have a feeling we'll see more of these. I'll fiddle with making these static methods and see what that looks like.

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: I thought about having the normalizers be simple string operations but then every invocation would be

    normalized_minor_release(get_var(task_vars, "openshift_release"))

and I hate to see that kind of repetition. That's why I write methods :) Though obviously encapsulation is important.

Actually I don't like passing around `task_vars` everywhere to begin with. Why not just pass it to the check constructor to be put in an attribute and used later?

rhcarvalho: > Actually I don't like passing around task_vars everywhere to begin with. Why not just pass it to the check constructor to be put in an attribute and used later?

I'm also not in favor of passing `task_vars` everywhere indiscriminately.

Storing `task_vars` in an instance attribute has the potential to make dependencies implicit instead of explicit, what might be more damage than good.

If health checks need to access version information, we could make those read-only properties of OpenShiftCheck.... hmm, I started to sketch some code and realized that `OpenShiftCheck.__init__` doesn't take `task_vars`, only `run` does.

I remember the initial "hope" was that we could load the checks only once per execution, then run it once per host in the inventory. However, in practice the action plugin is executed once per each host... so we could change `OpenShiftCheck` to be a "disposable" instance, to be used only once.

rhcarvalho: Changing `OpenShiftCheck.run`'s signatures brings a similar tradeoff than adding static methods.

In every subclass, calls to a module would become:

```python
class MyCheck(OpenShiftCheck):
    def run(self):
        self.module_executor("module_name", self.tmp, self.task_vars)
```

This loses the symmetry with action plugins, the run method would have a different signature.

If we keep diving that way, we could even assume that `self.tmp, and self.task_vars` are the common case and make a helper:

```python
class MyCheck(OpenShiftCheck):
    def run(self):
        self.execute_module("module_name")
```

Please do experiment and improve the "framework" as we progress and add more checks ;)

roles/openshift_health_checker/library/aos_version.py
rhcarvalho: Correct, now looking at the single commit I see it!
If we wanted a more generic module, it would be interesting if the module itself handled the version format... but that is harder to do code reuse... I wonder if Ansible always copy every file from `library/` or just the file with the same name as the module being invoked?

In any case, this module is very specific, it is fine to expect a particular format of version. Could you please document the expected format in the module, just for our future reference?

sosiouxme: `openshift_release` is used by a lot of other stuff in the installer, I'm just piggy-backing on it here. Actually it's not well-documented what it's supposed to be, and I'm finding it doesn't have to be specified. Worth documenting as you say.

sosiouxme:
The excluder part of this has become moot; install/upgrade playbooks will handle excluder state. Closing this and opening a different PR for the other changes I'm working on.

sdodson:
> The excluder part of this has become moot; install/upgrade playbooks will handle excluder state. Closing this and opening a different PR for the other changes I'm working on.

As many eyes on this implementation as possible would be helpful. It's pretty tricky and it seems like every day we find a new corner case or change our minds about how it should be implemented.

sosiouxme:
https://github.com/openshift/openshift-ansible/pull/3649 is the followup

PR #3580, opened 2017-03-07
Do not use auto_expand_replicas

lukas-vlcek:
Do not use `auto_expand_replicas` feature. It is associated with unnecessary data recovery from other nodes.

See for details: https://github.com/elastic/elasticsearch/issues/1873

### DIFF
--- a/roles/openshift_logging/README.md
+++ b/roles/openshift_logging/README.md
@@ -72,6 +72,8 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_es_recover_after_time`: The amount of time ES will wait before it tries to recover. Defaults to '5m'.
 - `openshift_logging_es_storage_group`: The storage group used for ES. Defaults to '65534'.
 - `openshift_logging_es_nodeselector`: A map of labels (e.g. {"node":"infra","region":"west"} to select the nodes where the pod will land.
+- `openshift_logging_es_number_of_shards`: The number of primary shards for every new index created in ES. Defaults to '1'.
+- `openshift_logging_es_number_of_replicas`: The number of replica shards per primary shard for every new index. Defaults to '0'.
 
 When `openshift_logging_use_ops` is `True`, there are some additional vars. These work the
 same as above for their non-ops counterparts, but apply to the OPS cluster instance:
@@ -88,6 +90,8 @@ same as above for their non-ops counterparts, but apply to the OPS cluster insta
 - `openshift_logging_es_ops_pvc_prefix`: logging-es-ops
 - `openshift_logging_es_ops_recover_after_time`: 5m
 - `openshift_logging_es_ops_storage_group`: 65534
+- `openshift_logging_es_ops_number_of_shards`: The number of primary shards for every new index created in ES. Defaults to '1'.
+- `openshift_logging_es_ops_number_of_replicas`: The number of replica shards per primary shard for every new index. Defaults to '0'. 
 - `openshift_logging_kibana_ops_hostname`: The Operations Kibana hostname. Defaults to 'kibana-ops.example.com'.
 - `openshift_logging_kibana_ops_cpu_limit`: The amount of CPU to allocate to Kibana or unset if not specified.
 - `openshift_logging_kibana_ops_memory_limit`: The amount of memory to allocate to Kibana or unset if not specified.
--- a/roles/openshift_logging/tasks/install_elasticsearch.yaml
+++ b/roles/openshift_logging/tasks/install_elasticsearch.yaml
@@ -53,6 +53,8 @@
     deploy_name: "{{item.1}}"
     es_node_selector: "{{openshift_logging_es_nodeselector | default({}) }}"
     es_storage: "{{openshift_logging_facts|es_storage(deploy_name, pvc_claim)}}"
+    es_number_of_shards: "{{ openshift_logging_es_number_of_shards }}"
+    es_number_of_replicas: "{{ openshift_logging_es_number_of_replicas }}"
   with_indexed_items:
     - "{{ es_dc_pool }}"
   check_mode: no
@@ -134,6 +136,8 @@
     openshift_logging_es_recover_after_time: "{{openshift_logging_es_ops_recover_after_time}}"
     es_node_selector: "{{openshift_logging_es_ops_nodeselector | default({}) }}"
     es_storage: "{{openshift_logging_facts|es_storage(deploy_name, pvc_claim,root='elasticsearch_ops')}}"
+    es_number_of_shards: "{{ openshift_logging_es_ops_number_of_shards }}"
+    es_number_of_replicas: "{{ openshift_logging_es_ops_number_of_replicas }}"
   with_indexed_items:
     - "{{ es_ops_dc_pool | default([]) }}"
   when:
--- a/roles/openshift_logging/templates/elasticsearch.yml.j2
+++ b/roles/openshift_logging/templates/elasticsearch.yml.j2
@@ -6,9 +6,8 @@ script:
   indexed: on
 
 index:
-  number_of_shards: 1
-  number_of_replicas: 0
-  auto_expand_replicas: 0-2
+  number_of_shards: {{ es_number_of_shards | default ('1') }}
+  number_of_replicas: {{ es_number_of_replicas | default ('0') }}
   unassigned.node_left.delayed_timeout: 2m
   translog:
     flush_threshold_size: 256mb
--- a/roles/openshift_logging/defaults/main.yml
+++ b/roles/openshift_logging/defaults/main.yml
@@ -92,6 +92,8 @@ openshift_logging_es_storage_group: "{{ openshift_hosted_logging_elasticsearch_s
 openshift_logging_es_nodeselector: "{{ openshift_hosted_logging_elasticsearch_nodeselector | default('') | map_from_pairs }}"
 # openshift_logging_es_config is a hash to be merged into the defaults for the elasticsearch.yaml
 openshift_logging_es_config: {}
+openshift_logging_es_number_of_shards: 1
+openshift_logging_es_number_of_replicas: 0
 
 # allow cluster-admin or cluster-reader to view operations index
 openshift_logging_es_ops_allow_cluster_reader: False
@@ -111,6 +113,8 @@ openshift_logging_es_ops_pvc_prefix: "{{ openshift_hosted_logging_elasticsearch_
 openshift_logging_es_ops_recover_after_time: 5m
 openshift_logging_es_ops_storage_group: "{{ openshift_hosted_logging_elasticsearch_storage_group | default('65534') }}"
 openshift_logging_es_ops_nodeselector: "{{ openshift_hosted_logging_elasticsearch_ops_nodeselector | default('') | map_from_pairs }}"
+openshift_logging_es_ops_number_of_shards: 1
+openshift_logging_es_ops_number_of_replicas: 0
 
 # storage related defaults
 openshift_logging_storage_access_modes: "{{ openshift_hosted_logging_storage_access_modes | default(['ReadWriteOnce']) }}"


lukas-vlcek:
The following is simple reproduction script:

````bash
# Start two ES nodes and for clarity make sure each node stores data to different location

# ES 5.2.2
./bin/elasticsearch -E path.data=./node1/data -E path.logs=./node1/logs
./bin/elasticsearch -E path.data=./node2/data -E path.logs=./node2/logs

# ES 2.4.4
# ./bin/elasticsearch -Dpath.data=./node1/data -Dpath.logs=./node1/logs
# ./bin/elasticsearch -Dpath.data=./node2/data -Dpath.logs=./node2/logs
````
After nodes are up and cluster is formed:
````bash
ES1=http://localhost:9200
ES2=http://localhost:9201

# curl $ES1/_cluster/health?pretty
# curl $ES2/_cluster/health?pretty

# Prepare index template using the "auto_expand_replicas"
curl -X POST $ES1/_template/auto_index -d '{
  "template": "auto*",
  "settings": {
    "number_of_shards": 1,
    "number_of_replicas": 0,
    "auto_expand_replicas": "0-3"
  }
}'

# Another index template using 1/1 shard/replica schema
curl -X POST $ES1/_template/fixed_index -d '{
  "template": "fixed*",
  "settings": {
    "number_of_shards": 1,
    "number_of_replicas": 1
  }
}'

# Index two documents. Each goes to different index.
# - "auto" index uses 'auto_index_expand'
# - "fixed" index uses 1/1 schema
curl -X POST $ES1/auto/type -d '{ "name": "foo" }'
curl -X POST $ES1/fixed/type -d '{ "name": "foo" }'

# ... and let's make sure the data is flushed to disk
# so we can clearly track Lucene segment files on the FS
curl -X POST "$ES1/_flush?wait_if_ongoing=true&force=true"
````
At this point we can check `_cat/indices`:
````bash
curl $ES1/_cat/indices?v

health status index uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   auto  6OLcLCRhT8GM34aExedFkg   1   1          1            0      6.6kb          3.3kb
green  open   fixed miBt1Om1SsegDb67XNpmIg   1   1          1            0      6.6kb          3.3kb
````
And we can also check data store for each node:
````bash
$ cd ./node1/data/nodes/0/indices
$ du -h
4.0K	./6OLcLCRhT8GM34aExedFkg/0/_state
 16K	./6OLcLCRhT8GM34aExedFkg/0/index
8.0K	./6OLcLCRhT8GM34aExedFkg/0/translog
 28K	./6OLcLCRhT8GM34aExedFkg/0
4.0K	./6OLcLCRhT8GM34aExedFkg/_state
 32K	./6OLcLCRhT8GM34aExedFkg
4.0K	./miBt1Om1SsegDb67XNpmIg/0/_state
 16K	./miBt1Om1SsegDb67XNpmIg/0/index
8.0K	./miBt1Om1SsegDb67XNpmIg/0/translog
 28K	./miBt1Om1SsegDb67XNpmIg/0
4.0K	./miBt1Om1SsegDb67XNpmIg/_state
 32K	./miBt1Om1SsegDb67XNpmIg
 64K	.
````
````bash
$ cd ./node2/data/nodes/0/indices
$ du -h
4.0K	./6OLcLCRhT8GM34aExedFkg/0/_state
 16K	./6OLcLCRhT8GM34aExedFkg/0/index
8.0K	./6OLcLCRhT8GM34aExedFkg/0/translog
 28K	./6OLcLCRhT8GM34aExedFkg/0
4.0K	./6OLcLCRhT8GM34aExedFkg/_state
 32K	./6OLcLCRhT8GM34aExedFkg
4.0K	./miBt1Om1SsegDb67XNpmIg/0/_state
 16K	./miBt1Om1SsegDb67XNpmIg/0/index
8.0K	./miBt1Om1SsegDb67XNpmIg/0/translog
 28K	./miBt1Om1SsegDb67XNpmIg/0
4.0K	./miBt1Om1SsegDb67XNpmIg/_state
 32K	./miBt1Om1SsegDb67XNpmIg
 64K	.
````
Both index data on both indices.

Now, let's **disable allocation**:
````bash
curl -X PUT $ES1/_cluster/settings -d '{
  "transient": {"cluster.routing.allocation.enable": "none"}
}'
# {"acknowledged":true,"persistent":{},"transient":{"cluster":{"routing":{"allocation":{"enable":"none"}}}}}
````

Shutdown the node2.
````bash
# The following will appear in log of node1:
# [CJxI8Hy] updating number_of_replicas to [0] for indices [auto]
# [CJxI8Hy] [auto/6OLcLCRhT8GM34aExedFkg] auto expanded replicas to [0]
````
Start the node2 again:
````bash
./bin/elasticsearch -E path.data=./node2/data -E path.logs=./node2/logs

# its log will stop at:
# [mQvaM96] started
````
Now the data for index `auto` (6OLcLCRhT8GM34aExedFkg) is gone in the node2 store:
````bash
$ du -h
4.0K	./6OLcLCRhT8GM34aExedFkg/_state
4.0K	./6OLcLCRhT8GM34aExedFkg
4.0K	./miBt1Om1SsegDb67XNpmIg/0/_state
 16K	./miBt1Om1SsegDb67XNpmIg/0/index
8.0K	./miBt1Om1SsegDb67XNpmIg/0/translog
 28K	./miBt1Om1SsegDb67XNpmIg/0
4.0K	./miBt1Om1SsegDb67XNpmIg/_state
 32K	./miBt1Om1SsegDb67XNpmIg
 36K	.
````

Now, let's **enable allocation**:
````bash
curl -X PUT $ES1/_cluster/settings -d '{
  "transient": {"cluster.routing.allocation.enable": "all"}
}'
# {"acknowledged":true,"persistent":{},"transient":{"cluster":{"routing":{"allocation":{"enable":"all"}}}}}
````

And the data for index `auto` is recovered from other node:
````bash
$ du -h
4.0K	./6OLcLCRhT8GM34aExedFkg/0/_state
 16K	./6OLcLCRhT8GM34aExedFkg/0/index
8.0K	./6OLcLCRhT8GM34aExedFkg/0/translog
 28K	./6OLcLCRhT8GM34aExedFkg/0
4.0K	./6OLcLCRhT8GM34aExedFkg/_state
 32K	./6OLcLCRhT8GM34aExedFkg
4.0K	./miBt1Om1SsegDb67XNpmIg/0/_state
 16K	./miBt1Om1SsegDb67XNpmIg/0/index
8.0K	./miBt1Om1SsegDb67XNpmIg/0/translog
 28K	./miBt1Om1SsegDb67XNpmIg/0
4.0K	./miBt1Om1SsegDb67XNpmIg/_state
 32K	./miBt1Om1SsegDb67XNpmIg
 64K	.
````

----
For more detailed investigation it is possible to turn on logging like this:
````
  indices.recovery: TRACE
  index.shard: TRACE
````
This brings related low level info into ES logs.
Notice: the logging config changed a bit for ES 5.x, the above tip works for ES 2.x.

**Comment by ewolinetz** [comment]
@lukas-vlcek the intent with using the `auto_expand_replicas` was to remove the amount of administrative work that users needed to do with their EFK stack as deployed by OCP. This way a customer's cluster would be green even if they only have a single node (and then the node wouldn't need to maintain information about replica shards that it may never be able to place).

Are we going to start recommending to customers that they should be updating the number of replicas based on their cluster size and also providing commands/documentation on how they would accomplish this? That is my concern with us no longer using `auto_expand_replicas`

**Comment by lukas-vlcek** [comment]
@ewolinetz I understand and welcome effort to minimize user maintenance work and I am not against it. But this ticket is not about maintenance ease, this ticket tries to describe issues associated with using `auto_expand_replicas`. IMO it introduces serious issues (especially in our logging use case) and I do not see any justification for keeping it in OOB experience any longer. If you want to keep it in then please make sure users are **well informed about associated risks and downsides** so they can make qualified decision.

From what I understand when `auto_expand_replicas` is used then local copy of node data is always dropped when node joins cluster, no matter if shard allocation is turned on or off, no matter if the node shutdown/startup is controlled by user or result of network load, long CG run or whatever. In case of logging (especially given our existing setup, i.e. 1 primary shard and 2 or 3 replica shards) this means that every node joining 2 or 3 node cluster will **always pull all the data** from other nodes. Are associated expenses worth it?

//cc @portante

**Comment by lukas-vlcek** [comment]
@ewolinetz As for the second part of your question:

> Are we going to start recommending to customers that they should be updating the number of replicas based on their cluster size and also providing commands/documentation on how they would accomplish this?

I think unless we can reliably automate this for users then providing tooling sounds like good option to me. What is wrong with cluster in yellow state compared to single node cluster in green state? With single node cluster aren't you in permanent risk of loosing service availability anyway?

**Comment by richm** [comment]
I think we just need to document the cases where having a yellow cluster is OK.  It is really a problem for large deployments if nodes have to be unnecessarily initialized with Elasticsearch data.

**Comment by ewolinetz** [comment]
@lukas-vlcek can you also open a PR for this against release-1.5?

File: roles/openshift_logging/templates/elasticsearch.yml.j2 [hunk 1]
```diff
@@ -7,8 +7,7 @@ script:
 
 index:
   number_of_shards: 1
-  number_of_replicas: 0
-  auto_expand_replicas: 0-2
+  number_of_replicas: 1
```
portante: Let's make this `0` by default, and give the customers a setting to change it to the value they want.

I would think both of these numbers (shard and replica counts) should be templated to allow a variable to change their values.

ewolinetz: Can I make the argument that this is something we can specify as part of the install? 
Or if a customer states they want an ES cluster size of `2` or more we would then change this value?

Or or - if not specified we would change this value based on the intended cluster size (0 - 2) but then give users the ability to overwrite it if they so desire.

jcantrill:
@lukas-vlcek It looks like this PR is to address https://bugzilla.redhat.com/show_bug.cgi?id=1430910.  Could you also update the commit message to 'bug 1430910. MESSAGE HERE' which will add commit info to PR

jcantrill:
@ewolinetz If we are to make this configurable, we should consider how to incorporate these changes as part of a map.  It is going to become exceedingly unweildy to create a var like 'openshift_logging_ADDYOURCONFIGOPTIONHERE' everytime we add these tweeks.  It probably should be something like:

```
openshift_logging_es_settings:
   number_of_shards: 1
   number_of_replicas: 0
```

ewolinetz:
@jcantrill i'd just be hesitant to increase the complexity simply to reduce verbosity

jcantrill:
@ewolinetz I dont see this is being that much more additionally complex.  Ansible already supports the notion of variable files which are yaml as opposed to inventory files.  This seems like a more advanced configuration anyway where we might expect users to have or be able to utilize variable files.

ewolinetz:
@lukas-vlcek can you update the template for the es config to use two variables, one for primary shards one for replicas? Something along the lines of `openshift_logging_es_primary_shards` `openshift_logging_es_replicas` (and then `ops` instances).

The defaults should probably be 1 primary and 0 or 1 replicas (either should be fine, customers are currently used to having yellow state if they only have one node).

lukas-vlcek:
@ewolinetz I will do it on Monday (on PTO today 😎 ).

lukas-vlcek:
@ewolinetz let me know if this is what you meant (`/me` not expert on Ansible and its variables):

    number_of_shards: {{openshift_logging_es_primary_shards | default ('1')}}
    number_of_replicas: {{openshift_logging_es_replicas | default ('1')}}

lukas-vlcek:
@ewolinetz May be I should add used variables also to both [`roles/openshift_logging/defaults/main.yml`](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/defaults/main.yml) and [`roles/openshift_logging/vars/main.yaml`](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/vars/main.yaml)?

portante:
@lukas-vlcek, I think the default for shard replication should be "`0`".

lukas-vlcek:
@portante updated

ewolinetz:
@lukas-vlcek if you could just add to `defaults/main.yaml` that would be good. Entries in `variables/main.yaml` are generally not meant to be changed.

We likely should also have ops instances of these as well, since customers may want to configure them differently. That would mean that we would then pass in the values to the templates in `tasks/install_elasticsearch.yaml` instead. If you wouldn't mind doing so.

Can you also update the README for the role just to explain what these vars do?

Otherwise LGTM

lukas-vlcek:
@ewolinetz I think I need some help figuring out how to modify `roles/openshift_logging/tasks/install_elasticsearch.yaml` to make this complete and shine.

lukas-vlcek:
@ewolinetz I did another update, PTAL

ewolinetz:
aos-ci-test

lukas-vlcek:
@ewolinetz I think we still need documentation? Can use override values in `/defaults/main.yml`. If not then documentation update is not needed but if user can override then we should complete it.

ewolinetz:
@lukas-vlcek correct, a user an override what is in `defaults/main.yaml`. If you could add some brief information in the readme that would be great

richm:
@lukas-vlcek yes, please document here: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/README.md - and yes, user can override the values

lukas-vlcek:
@ewolinetz @richm PTAL & check my wording.

roles/openshift_logging/README.md
```diff
@@ -72,6 +72,8 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_es_recover_after_time`: The amount of time ES will wait before it tries to recover. Defaults to '5m'.
 - `openshift_logging_es_storage_group`: The storage group used for ES. Defaults to '65534'.
 - `openshift_logging_es_nodeselector`: A map of labels (e.g. {"node":"infra","region":"west"} to select the nodes where the pod will land.
+- `openshift_logging_es_number_of_shards`: The number of shards for every new index created in ES. Defaults to '1'.
+- `openshift_logging_es_number_of_replicas`: The number of shard replicas for every new index created in ES. Defaults to '0'.
```
ewolinetz: Just in case people aren't clear, maybe we should specify that this number is 'this many replicas per shard'?

roles/openshift_logging/README.md
```diff
@@ -88,6 +90,8 @@ same as above for their non-ops counterparts, but apply to the OPS cluster insta
 - `openshift_logging_es_ops_pvc_prefix`: logging-es-ops
 - `openshift_logging_es_ops_recover_after_time`: 5m
 - `openshift_logging_es_ops_storage_group`: 65534
+- `openshift_logging_es_ops_number_of_shards`: The number of shards for every new index created in ES. Defaults to '1'.
+- `openshift_logging_es_ops_number_of_replicas`: The number of shard replicas for every new index created in ES. Defaults to '0'.
```
ewolinetz: Same as above

roles/openshift_logging/README.md
```diff
@@ -72,6 +72,8 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_es_recover_after_time`: The amount of time ES will wait before it tries to recover. Defaults to '5m'.
 - `openshift_logging_es_storage_group`: The storage group used for ES. Defaults to '65534'.
 - `openshift_logging_es_nodeselector`: A map of labels (e.g. {"node":"infra","region":"west"} to select the nodes where the pod will land.
+- `openshift_logging_es_number_of_shards`: The number of shards for every new index created in ES. Defaults to '1'.
```
ewolinetz: "The number of primary shards"

roles/openshift_logging/README.md
```diff
@@ -88,6 +90,8 @@ same as above for their non-ops counterparts, but apply to the OPS cluster insta
 - `openshift_logging_es_ops_pvc_prefix`: logging-es-ops
 - `openshift_logging_es_ops_recover_after_time`: 5m
 - `openshift_logging_es_ops_storage_group`: 65534
+- `openshift_logging_es_ops_number_of_shards`: The number of shards for every new index created in ES. Defaults to '1'.
```
ewolinetz: "The number of primary shards"

lukas-vlcek:
@ewolinetz updated

ewolinetz:
aos-ci-test

roles/openshift_logging/tasks/install_elasticsearch.yaml
```diff
@@ -134,6 +136,8 @@
     openshift_logging_es_recover_after_time: "{{openshift_logging_es_ops_recover_after_time}}"
```
portante: Should this be `openshift_logging_es_recover_after_time`? Or should that be `openshift_logging_es_ops_recover_after_time`?

lukas-vlcek: @portante It can make sense, I will look at this tomorrow.

lukas-vlcek: @portante I think it is correct.

There is `{{openshift_logging_es_recover_after_time}}` value set to `${RECOVER_AFTER_TIME}` env variable in [`roles/openshift_logging/templates/es.j2`](
https://github.com/openshift/openshift-ansible/blob/39d078c8b1a17f3e4b827f9962a9f317a6464f3e/roles/openshift_logging/templates/es.j2#L87-L88) which is then used in [`elasticsearch.yml.j2` template](https://github.com/openshift/openshift-ansible/blob/39d078c8b1a17f3e4b827f9962a9f317a6464f3e/roles/openshift_logging/templates/elasticsearch.yml.j2#L37).

_//cc @ewolinetz (Erik see below please)_

However, I do not understand why this value is set only for **DeploymentConfig for Ops** in `tasks/install_elasticsearch.yaml` script. Please compare `vars` sections [for Ops](https://github.com/openshift/openshift-ansible/blob/52726611f182c6db390d7a964f0bf24fed297761/roles/openshift_logging/tasks/install_elasticsearch.yaml#L121-L136) vs `vars` section [for Non-Ops](tasks/install_elasticsearch.yaml).

Why the following four `vars` are set only for Ops DeploymentConfig?

- `es_node_quorum`
- `es_recover_after_nodes`
- `es_recover_expected_nodes`
- `openshift_logging_es_recover_after_time`

Apart from this if there is any issue with `openshift_logging_es_recover_after_time` usage and configuration then I suggest to address it in separated ticket/PR.

lukas-vlcek:
rebased

roles/openshift_logging/tasks/install_elasticsearch.yaml
jcantrill: Per my IRC comment, the difference is when running the 'ops' tasks, we update the variables to be the 'ops' variables.  Technically, we should have written these tasks to run twice with a single set of tasks: once with non-ops and the other with ops.  The four variables in question are found in the vars/main.yaml

ewolinetz:
[test]

ewolinetz:
aos-ci-test

ewolinetz:
[merge]

roles/openshift_logging/templates/elasticsearch.yml.j2
```diff
@@ -6,9 +6,8 @@ script:
   indexed: on
 
 index:
-  number_of_shards: 1
-  number_of_replicas: 0
-  auto_expand_replicas: 0-2
+  number_of_shards: {{ es_number_of_shards | default ('1') }}
+  number_of_replicas: {{ es_number_of_replicas | default ('0') }}
```
richm: not exactly sure why, but because of the way we use this template, the values have to be quoted.  Please change this to be:
```
   number_of_shards: "{{ es_number_of_shards | default ('1') }}"
   number_of_replicas: "{{ es_number_of_replicas | default ('0') }}"
```
You'll have to submit a new PR since this one was merged.

lukas-vlcek: Thanks @richm! I opened new PR #3763

PR #3582, opened 2017-03-07
Augmenting documentation for router sharding.

kwoodson:


kwoodson:
@sdodson, added to the hosts.origin.example inventory doc.

PR #3588, opened 2017-03-07
Bug 1416156: Loosely couple docker to iptables service

mtnbikenc:
The docker service should be loosely coupled to the iptables service to avoid restarting docker when iptables is restarted.

Fixes 1416156


mtnbikenc:
aos-ci-test

PR #3589, opened 2017-03-07
Adding scripts for building and pushing images

brenton:
These scripts follow the same pattern as the origin and origin-metrics repo.
From there the Jenkins job will be very simple.

### DIFF
--- a/hack/build-images.sh
+++ b/hack/build-images.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+STARTTIME=$(date +%s)
+source_root=$(dirname "${0}")/..
+
+prefix="openshift/openshift-ansible"
+version="latest"
+verbose=false
+options=""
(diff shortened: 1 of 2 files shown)


brenton:
The basic idea for now is to have a push-button job to build the openshift-ansible image.  Previously we had called this image openshift-ansible-preflight but I think we might as well just have a generic openshift-ansible image now.

Here's a look at the job.  It's pulling from my fork for now:
https://ci.openshift.redhat.com/jenkins/job/build-and-release-latest-openshift-ansible-images/

@rhcarvalho, @codificat

rhcarvalho:
:shipit:

brenton:
@sdodson, PTAL.  These are scripts that were lifted from the origin repo.  The just build the Dockerfile in the root of this repo as 'openshift-ansible'.  A jenkins job will handle pushing it to the Dockerhub.  We'll work with the CD team on the exact needs for the job.  Right now it's just push-button.

PR #3590, opened 2017-03-07
Fix to OpenshiftCLIConfig to support an ordereddict.  Removed redundant get.

kwoodson:


tbielawa:
@kwoodson gotta fix up those test failures.

kwoodson:
@tbielawa, tests pass now.  I think we are good.

kwoodson:
aos-ci-test

stevekuznetsov:
Why not use an ordered data structure, not a `dict`, if you want ordering? This seems error-prone.

kwoodson:
@stevekuznetsov, I attempted this changed with collections.OrderedDict but found that the test cases continued to fail  https://github.com/openshift/openshift-ansible/pull/3590/commits/e2bad8efff2aaee4906575bc523c57bffa924715.  I swapped out ordereddict with a sort and it seemed more stable.

stevekuznetsov:
yo @abutcher you need to use [merge] tag now-adays

stevekuznetsov:
@kwoodson I would fix your test cases ... but using `sort()` on a `dict` is a poor method in comparison to an actually ordered data structure. Everyone that ever uses it, adds to it, removes from it, will need to make sure they are using `sort()`.

kwoodson:
@stevekuznetsov, I disagree that people using it will need to `sort()`.  The command line arguments can be any order.  The test case was capturing the calling data.  We could update the test cases but we still want to ensure the data gets passed to the test with the parameters.  This seemed like an easier solution than modifying the tests and it does keep some consistency to the command line arguments.

stevekuznetsov:
I'm not sure I understand...
 - if you expect to use your data with an order, store it in an ordered data structure
 - if you are modifying your business logic to make your tests pass, you have bad tests

PR #3598, opened 2017-03-08
WIP: Check the excluders are in a proper state

ingvagabund:
This will warn a user if the openshift excluder is installed but not enabled. To make sure the user runs exclude command before continuing.

The prompt message is not printed from some reason.


ingvagabund:
Closing as the specification changed.

PR #3601, opened 2017-03-08
add ability to specify an etcd version

mmckinst:
I needed this feature when duplicating an existing openshift cluster and needed to match the etcd version. It could also be useful if someone isn't ready to upgrade to etcd 3.x because it changes stuff including how to do backups and restore/disaster recovery.

### DIFF
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -89,6 +89,8 @@ openshift_release=v1.4
 # Skip upgrading Docker during an OpenShift upgrade, leaves the current Docker version alone.
 # docker_upgrade=False
 
+# Specify exact version of etcd to configure or upgrade to.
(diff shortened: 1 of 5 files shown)


sdodson:
Seems alright for install scenarios. However, currently we update etcd in lock step when the upgrade playbooks are executed, this PR doesn't account for that, are you ok with that?

Also, just a general FYI, you must get to etcd-3.1 before upgrading to OpenShift 3.6.

mmckinst:
I only needed it for the install. I'll see if I can add some conditionals to the upgrade script too along with a warning to check the docs that older an etcd might not work with a newer openshift version.

sdodson:
bot, test pull request

sdodson:
aos-ci-test

sdodson:
[test]
[merge]

PR #3605, opened 2017-03-08
Fixing the way policies are found.  The old method was unreliable.  This method searches all and matches on properties.

kwoodson:


abutcher:
[test]
[merge]

abutcher:
@stevekuznetsov Like this?

stevekuznetsov:
@abutcher yeah... merge implies test, and merge will check for the aos-ci- stuff to have reported back as green, as well

kwoodson:
@stevekuznetsov, how do we determine the openshift-bot's merge failure?

abutcher:
aos-ci-test

stevekuznetsov:
@kwoodson if you follow the link to the job, the best bet is to look in the "Test Result" section for jUnit failures. If there's nothing indicated there, you can look at the "Identified problems" section for other possible failures -- if there's nothing in either of those sections, it is probably easiest to ask me to look at the log.

kwoodson:
@stevekuznetsov, seems as though something failed internally.  Unsure about this one.
`13:00:22 
[ERROR] PID 22825: test/extended/conformance.sh:56: `return $out` exited with status 1.`

stevekuznetsov:
Hmm... looking at it now, it looks like a conformance test failed, judging by the [jUnit status](https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/5/testReport/junit/(root)/Extended/deploymentconfigs_rolled_back__Conformance__should_rollback_to_an_older_deployment/)

stevekuznetsov:
@sdodson so which robot did this merge?

sdodson:
@openshift-bot did

stevekuznetsov:
To explain what Rosie did here....

There is an option to run the robot with "pre-test success" where, if a PR is slated to merge, but has already run a set of tests against the latest commit in master, it will just be merged. The reasoning is that if we have tested green against master and nothing else has merged since then, we can be reasonably certain that we don't need to test again. This option doesn't make a lot of sense when the job concerns code coming from multiple repositories, but nevertheless the robot did not *technically* misbehave here.

php-coder:
@kwoodson Looks like `roles/lib_openshift/library/oc_adm_registry.py.good` and `roles/lib_openshift/library/oc_adm_registry.py.bak` were committed by mistake.

kwoodson:
@php-coder, thanks will remove.

PR #3607, opened 2017-03-09
extend the excluders to containerized deployment

ingvagabund:
the docker excluder is needed in containerized deployment as well

bug 1430700

### DIFF
--- a/roles/openshift_excluder/tasks/adjust.yml
+++ b/roles/openshift_excluder/tasks/adjust.yml
@@ -20,4 +20,4 @@
       # disable it if the docker excluder is enabled
       disable_openshift_excluder: "{{ openshift_excluder_on | bool }}"
   when:
-  - not openshift.common.is_containerized | bool
+  - not openshift.common.is_atomic | bool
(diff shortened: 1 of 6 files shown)


ingvagabund:
aos-ci-test

ganhuang:
I personally don't think this is the solution QE wanted. We usually don't have OCP repo for containerized deployment on RHEL. It's happy to see installer failed when no OCP repos on RHEL for containerized installation.

sdodson:
@ganhuang but that's the only location that the excluders exist and we should enforce docker exclusions. How else should we do that?

The documentation actually makes no distinction between containerized and rpm based hosts as far as enabling channels. So I think it's fine to rely on content from the OCP channels. Only after finishing the host preparation does the documentation diverge for containerized versus rpm based installation.

See 
https://docs.openshift.com/container-platform/3.4/install_config/install/host_preparation.html

ganhuang:
@sdodson I forgot to ack that bug. I thought we have made some efforts to not use OCP channel in containerized deployment on RHEL. 

To make the default setting in openshift-ansible more reliable, we can leave the options to users so that they can decide if installer would install *-excluder in containerized installation on RHEL instead of failing directly if they don't have that channels subscribed. Upgrade can follow that options as well in this case.

sdodson:
> @sdodson I forgot to ack that bug. I thought we have made some efforts to not use OCP channel in containerized deployment on RHEL.

We did for the python-ruamel-yaml thing because we had a fall back, but I'm afraid that we're going to have more and more requirements that only exist in the OCP channels so we should become more comfortable with including them.

PR #3610, opened 2017-03-09
re-enable excluders if they are enabled after openshift version detection

ingvagabund:
SSIA

Bug 1430613

### DIFF
--- a/playbooks/common/openshift-cluster/config.yml
+++ b/playbooks/common/openshift-cluster/config.yml
@@ -60,3 +60,7 @@
 - include: openshift_hosted.yml
   tags:
   - hosted
+
+- include: reset_excluder.yml
+  tags:
+  - always
--- a/playbooks/common/openshift-cluster/initialize_openshift_version.yml
+++ b/playbooks/common/openshift-cluster/initialize_openshift_version.yml
@@ -18,12 +18,14 @@
       msg: Incompatible versions of yum and subscription-manager found. You may need to update yum and yum-utils.
     when: "not openshift.common.is_atomic | bool and 'Plugin \"search-disabled-repos\" requires API 2.7. Supported API is 2.6.' in yum_ver_test.stdout"
 
+# TODO(jchaloup): find a different way how to make repoquery --qf '%version` atomic-openshift work without disabling the excluders
 - include: disable_excluder.yml
   vars:
     # the excluders needs to be disabled no matter what status says
     with_status_check: false
   tags:
   - always
+  when: openshift_upgrade_target is not defined
 
 - name: Determine openshift_version to configure on first master
   hosts: oo_first_master
@@ -39,3 +41,9 @@
     openshift_version: "{{ hostvars[groups.oo_first_master.0].openshift_version }}"
   roles:
   - openshift_version
+
+  # Re-enable excluders if they are meant to be enabled (and only during installation, upgrade disables the excluders before this play)
+- include: reset_excluder.yml
+  tags:
+  - always
+  when: openshift_upgrade_target is not defined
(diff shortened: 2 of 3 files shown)


ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

sdodson:
[merge]

playbooks/common/openshift-cluster/config.yml
```diff
@@ -60,3 +60,7 @@
 - include: openshift_hosted.yml
   tags:
   - hosted
+
+- include: reset_excluder.yml
```
ingvagabund: The config is run during the installation only so the excluders are enabled at the end of the installation only.
During the installation I am assuming the cluster is vanilla new and without any excluder modifications (basically empty exclude item in ``/etc/yum.conf``. So it is ok to skip the status check. So basically it does not matter how many times we disable and re-enable the excluders. Cause the excluders must be to make sure affected packages are not updated out of installation/upgrade.

On the other hand, during the upgrade we have to disable and re-enable the excluders only once. Since disabling them means clearing the exclude item in ``/etc/yum.conf``. Which can be modified by a user (which is checked by the status command). So we should enable only those excluders that has not been changed by the user. Excluding docker cause we always want the docker excluder to take affect in order to safe the cluster from breaking apart during different docker version.

playbooks/common/openshift-cluster/initialize_openshift_version.yml
```diff
@@ -18,12 +18,14 @@
...
     when: "not openshift.common.is_atomic | bool and 'Plugin \"search-disabled-repos\" requires API 2.7. Supported API is 2.6.' in yum_ver_test.stdout"
 
+# TODO(jchaloup): find a different way how to make repoquery --qf '%version` atomic-openshift work without disabling the excluders
 - include: disable_excluder.yml
   vars:
     # the excluders needs to be disabled no matter what status says
     with_status_check: false
   tags:
   - always
+  when: openshift_upgrade_target is not defined
```
ingvagabund: So that's why I am skipping the disable task as mention on the comment above.

playbooks/common/openshift-cluster/initialize_openshift_version.yml
```diff
@@ -39,3 +41,9 @@
     openshift_version: "{{ hostvars[groups.oo_first_master.0].openshift_version }}"
   roles:
   - openshift_version
+
+  # Re-enable excluders if they are meant to be enabled (and only during installation, upgrade disables the excluders before this play)
+- include: reset_excluder.yml
+  tags:
+  - always
+  when: openshift_upgrade_target is not defined
```
ingvagabund: So that's why I am skipping the disable task as mention on the comment above.

playbooks/common/openshift-cluster/initialize_openshift_version.yml
```diff
@@ -39,3 +41,9 @@
     openshift_version: "{{ hostvars[groups.oo_first_master.0].openshift_version }}"
   roles:
   - openshift_version
+
+  # Re-enable excluders if they are meant to be enabled (and only during installation, upgrade disables the excluders before this play)
+- include: reset_excluder.yml
```
ingvagabund: And just have the excluders to do their job after successful installation.

sdodson:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
``No package matching 'atomic-openshift-tests' found available, installed or updated``

sdodson:
aos-ci-test

sdodson:
aos-ci-test

PR #3613, opened 2017-03-09
Use six from ansible.module_utils for remote hosts

detiber:
Remote hosts do not actually need python-six, we can use ansible.module_utils.six.

This removes the remote dependency on python-six.

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -18,9 +18,9 @@ from collections import Mapping
 from distutils.util import strtobool
 from distutils.version import LooseVersion
 from operator import itemgetter
-from ansible.module_utils.six.moves.urllib.parse import urlparse
 from ansible.parsing.yaml.dumper import AnsibleDumper
 from six import string_types
+from six.moves.urllib.parse import urlparse
 
 HAS_OPENSSL = False
 try:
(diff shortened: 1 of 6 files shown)


filter_plugins/oo_filters.py
```diff
@@ -18,9 +18,9 @@
 from distutils.util import strtobool
 from distutils.version import LooseVersion
 from operator import itemgetter
-from ansible.module_utils.six.moves.urllib.parse import urlparse
 from ansible.parsing.yaml.dumper import AnsibleDumper
 from six import string_types
+from six.moves.urllib.parse import urlparse
```
detiber: Filter plugins run from the command host, might as well use six rather than ansible.module_utils.six.

sdodson: Hmm, why? wouldn't that be one less point of variation and one less dependency to manage since we'd have to assume ansible is installed anyway?

detiber: Ansible requires python-six already, but if we did want to use the vendored six for control hosts, we should use ansible.compat.six, rather than ansible.module_utils.six

detiber: updated to use ansible.compat.six where possible.

detiber:
aos-ci-test

filter_plugins/oo_filters.py
sdodson: Got it, thanks for the explanation.

detiber:
[merge]

detiber:
[merge]

detiber:
@stevekuznetsov merge failure seems to be related to deprovisioning the instances and not actual test-related issues.

stevekuznetsov:
That's a job config issue, I'll get it when I get in in the morning.

PR #3625, opened 2017-03-10
bug 1430661. Update masterConfig metricsPublicURL on install

jcantrill:


jcantrill:
@sdodson @ewolinetz Should we cherry-pick to 1.5?

sdodson:
Yes please.

roles/openshift_metrics/tasks/update_master_config.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Adding metrics route information to metricsPublicURL
+  modify_yaml:
+    dest: "{{ openshift.common.config_base }}/master/master-config.yaml"
+    yaml_key: assetConfig.metricsPublicURL
+    yaml_value: "https://{{ openshift_metrics_hawkular_hostname}}"
```
ewolinetz: just to make sure, that variable doesn't already include "https://" correct?

ewolinetz:
This looks like it currently only updates for the master we are installing on, we want to make sure we also iterate over all the other masters as well and update their configs

roles/openshift_metrics/tasks/update_master_config.yaml
jcantrill: do we do that in logging? I did not see that.

jcantrill:
@ewolinetz is there example where we iterate masters to update configs?

ewolinetz:
@jcantrill https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/openshift_hosted.yml#L49-L62

jcantrill:
Seems like there should be some optimizations here as in if I'm installing both logging in metrics, there should be no need to bump the master server twice.

ewolinetz:
@jcantrill thats the beauty of a handler, it should only run it once as long as its installing it from the same playbook

sdodson:
aos-ci-test

sdodson:
[merge]

stevekuznetsov:
@openshift-bot, the [last build](https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/22/) failed from the following flakes:
 - [conformance](https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/22/testReport/junit/(root)/Extended/_builds__Conformance__oc_new_app_should_succeed_with_a___name_of_58_characters/) failed on: https://github.com/openshift/origin/issues/13067

re[merge]

PR #3628, opened 2017-03-10
Added the ability to do multiple edits

kwoodson:


kwoodson:
aos-ci-tests

kwoodson:
aos-ci-test

roles/lib_openshift/library/oc_adm_ca_server_cert.py
```diff
@@ -662,114 +670,148 @@ def parse_value(inc_value, vtype=''):
...
-                raise YeditException('Could not determine type of incoming ' +
-                                     'value. value=[%s] vtype=[%s]'
-                                     % (type(inc_value), vtype))
+                raise YeditException('Could not determine type of incoming value. ' +
+                                     'value=[{}] vtype=[{}]'.format(type(inc_value), vtype))
 
         return inc_value
 
+    @staticmethod
+    def process_edits(edits, yamlfile):
```
kwoodson: New method added to process edits.  This provides support for passing in a list of edits or a single edit and doing all of them in a single transaction.  This reduced the number of times the file is open, backed up, and written.

roles/lib_openshift/library/oc_adm_ca_server_cert.py
```diff
@@ -662,114 +670,148 @@ def parse_value(inc_value, vtype=''):
...
+
+            if rval[0]:
+                results.append({'key': edit['key'], 'edit': rval[1]})
+
+        return {'changed': len(results) > 0, 'results': results}
+
     # pylint: disable=too-many-return-statements,too-many-branches
     @staticmethod
-    def run_ansible(module):
+    def run_ansible(params):
```
kwoodson: Removing the module from run_ansible for a couple of reasons.

1. Following the patterns of the rest of the modules
1. Easier to mock
1. Allows a pure python implementation for testing and debugging.

roles/lib_openshift/library/oc_adm_ca_server_cert.py
```diff
@@ -662,114 +670,148 @@ def parse_value(inc_value, vtype=''):
...
                 return {'failed': True,
-                        'msg': 'Error opening file [%s].  Verify that the ' +
-                               'file exists, that it is has correct' +
-                               ' permissions, and is valid yaml.'}
-
-        if module.params['state'] == 'list':
-            if module.params['content']:
-                content = Yedit.parse_value(module.params['content'],
-                                            module.params['content_type'])
+                        'msg': 'Error opening file [{}].  Verify that the '.format(params['src']) +
```
kwoodson: fixed for format string

roles/lib_openshift/library/oc_adm_ca_server_cert.py
```diff
@@ -662,114 +670,148 @@ def parse_value(inc_value, vtype=''):
...
-                if module.params['update']:
-                    # pylint: disable=line-too-long
-                    curr_value = Yedit.get_curr_value(Yedit.parse_value(module.params['curr_value']),  # noqa: E501
-                                                      module.params['curr_value_format'])  # noqa: E501
+            # If we were passed a key, value then
+            # we enapsulate it in a list and process it
+            # Key, Value passed to the module : Converted to Edits list #
+            edits = []
+            _edit = {}
+            if params['value'] is not None:
```
kwoodson: If a single key, value was passed, put it in the edits list and we can then call a single path of execution, process_edits.

roles/lib_openshift/src/class/oc_volume.py
```diff
@@ -157,7 +157,7 @@ def run_ansible(params, check_mode=False):
             if not oc_volume.exists():
 
                 if check_mode:
-                    exit_json(changed=False, msg='Would have performed a create.')
+                    return {'changed': True, 'msg': 'CHECK_MODE: Would have performed a create.'}
```
kwoodson: Bug fix here.  Strange how this passed the original tests.  Somehow this was introduced.

roles/lib_openshift/library/oc_adm_ca_server_cert.py
```diff
@@ -206,13 +204,13 @@ def yaml_dict(self, value):
     def parse_key(key, sep='.'):
         '''parse the key allowing the appropriate separator'''
         common_separators = list(Yedit.com_sep - set([sep]))
-        return re.findall(Yedit.re_key % ''.join(common_separators), key)
+        return re.findall(Yedit.re_key.format(''.join(common_separators)), key)
```
ashcrow: :+1: for ``format()``

roles/lib_openshift/library/oc_adm_ca_server_cert.py
```diff
@@ -604,7 +602,17 @@ def put(self, path, value):
             pass
 
         result = Yedit.add_entry(tmp_copy, path, value, self.separator)
-        if not result:
+        if result is None:
```
ashcrow: :+1: explicit checking

roles/lib_utils/src/class/yedit.py
```diff
@@ -56,13 +55,13 @@ def yaml_dict(self, value):
     def parse_key(key, sep='.'):
         '''parse the key allowing the appropriate separator'''
         common_separators = list(Yedit.com_sep - set([sep]))
-        return re.findall(Yedit.re_key % ''.join(common_separators), key)
+        return re.findall(Yedit.re_key.format(''.join(common_separators)), key)
```
ashcrow: :+1: for ``format()``

roles/lib_utils/src/class/yedit.py
```diff
@@ -454,7 +453,17 @@ def put(self, path, value):
             pass
 
         result = Yedit.add_entry(tmp_copy, path, value, self.separator)
-        if not result:
+        if result is None:
```
ashcrow: :+1: for explicit checking

roles/lib_utils/src/class/yedit.py
```diff
@@ -480,7 +489,7 @@ def create(self, path, value):
                 pass
 
             result = Yedit.add_entry(tmp_copy, path, value, self.separator)
-            if result:
+            if result is not None:
```
ashcrow: :+1:

roles/lib_utils/src/class/yedit.py
```diff
@@ -512,112 +521,146 @@ def parse_value(inc_value, vtype=''):
...
         if isinstance(inc_value, str) and 'bool' in vtype:
             if inc_value not in true_bools and inc_value not in false_bools:
-                raise YeditException('Not a boolean type. str=[%s] vtype=[%s]'
-                                     % (inc_value, vtype))
+                raise YeditException('Not a boolean type. str=[{}] vtype=[{}]'.format(inc_value, vtype))
         elif isinstance(inc_value, bool) and 'str' in vtype:
             inc_value = str(inc_value)
 
+        # There is a special case where '' will turn into None after yaml loading it so skip
+        if isinstance(inc_value, str) and inc_value == '':
```
ashcrow: Not an issue, just curious why check for ``str`` and then check if it is equal to ``''``?

roles/lib_utils/src/class/yedit.py
```diff
@@ -512,112 +521,146 @@ def parse_value(inc_value, vtype=''):
...
+            pass
         # If vtype is not str then go ahead and attempt to yaml load it.
-        if isinstance(inc_value, str) and 'str' not in vtype:
+        elif isinstance(inc_value, str) and 'str' not in vtype:
             try:
-                inc_value = yaml.load(inc_value)
+                inc_value = yaml.safe_load(inc_value)
```
ashcrow: :+1: for ``safe_load()``

roles/lib_utils/src/class/yedit.py
```diff
@@ -512,112 +521,146 @@ def parse_value(inc_value, vtype=''):
...
 
+    @staticmethod
+    def process_edits(edits, yamlfile):
+        '''run through a list of edits and process them one-by-one'''
+        results = []
+        for edit in edits:
+            value = Yedit.parse_value(edit['value'], edit.get('value_type', ''))
+            if 'action' in edit and edit['action'] == 'update':
+                # pylint: disable=line-too-long
+                curr_value = Yedit.get_curr_value(Yedit.parse_value(edit.get('curr_value', None)),  # noqa: E501
```
ashcrow: Not an issue, but ``dict.get()`` no longer requires ``None`` if ``None`` is the default.

```python
In [1]: a = {}
In [2]: print(a.get('key'))
None
```

ashcrow: Not required, but you *may* be able to drop the noqa messages by using newlines to clarify

```python
                curr_value = Yedit.get_curr_value(
                    Yedit.parse_value(edit.get('curr_value', None)),
                    edit.get('curr_value_format', None))
```

roles/lib_utils/src/class/yedit.py
```diff
@@ -512,112 +521,146 @@ def parse_value(inc_value, vtype=''):
...
 
         return inc_value
 
+    @staticmethod
+    def process_edits(edits, yamlfile):
+        '''run through a list of edits and process them one-by-one'''
+        results = []
+        for edit in edits:
+            value = Yedit.parse_value(edit['value'], edit.get('value_type', ''))
+            if 'action' in edit and edit['action'] == 'update':
```
ashcrow: Not required. Could also be:

```python
if edit.get('action') == 'update':
```

ashcrow:
[test]

roles/lib_utils/src/class/yedit.py
kwoodson: @ashcrow, great question.  I would like to maybe visit it at some point as I find this a difficult topic.    The incoming value when passing _something_ to a module is generally defined in the module spec.  The problem here is that I want that value to be anything provided.  

The point of this module is to make a specific cut into a yaml file at a specific path and place a value.  Someone could pass in a dict, list, int, str, null, etc.

The problem comes in when loading the variable.  In ansible 1.9 it is different than ansible 2.x.  So I attempt to load the incoming value through yaml if it matches a certain data type.  If the incoming value is of type string and it is empty, the `yaml.safe_load` turns the `''` into `None`.  This is not the desired behavior.  We would like the string to show up in the yaml file as `''`.

I'm always open to discussions on how to handle incoming variable data that could be different types.  That would be helpful.

roles/lib_utils/src/class/yedit.py
kwoodson: I like this suggestion.  Will update.

roles/lib_utils/src/class/yedit.py
kwoodson: I like it.  I can never get the indent correct.  Will update.

kwoodson:
aos-ci-test

kwoodson:
aos-ci-test

kwoodson:
flake https://github.com/openshift/origin/issues/13067

kwoodson:
aos-ci-test

kwoodson:
[merge]

kwoodson:

kwoodson:
[merge]

ashcrow:
Error seems unrelated. Rebase?

```
"Error: Package: atomic-openshift-sdn-ovs-3.5.5-1.git.0.3f53382.el7.x86_64 (origin)\n           Requires: openvswitch >= 2.3.1\n", "rc": 1, "results": ["Loaded plugins: product-id, search-disabled-repos, subscription-manager\nThis system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.\nResolving Dependencies\n--> Running transaction check\n---> Package atomic-openshift-sdn-ovs.x86_64 0:3.5.5-1.git.0.3f53382.el7 will be installed\n--> Processing Dependency: openvswitch >= 2.3.1 for package: atomic-openshift-sdn-ovs-3.5.5-1.git.0.3f53382.el7.x86_64\n--> Finished Dependency Resolution\n You could try using --skip-broken to work around the problem\n You could try running: rpm -Va --nofiles --nodigest\n"
```

sdodson:
@ashcrow it's a change in the puddles that we install 3.5 from, the repos no longer have openvswitch in them so I've altered the CI jobs to add the fast-datapath repo. Next run should be ok, I hope.

ashcrow:
[merge]

kwoodson:
aos-ci-test

ashcrow:
[merge]

ashcrow:
```
[ERROR] Status 'aos-ci-jenkins/OS_3.5_NOT_containerized' is in the 'error' state. This status must succeed for a merge.
[ERROR] Status 'aos-ci-jenkins/OS_3.5_NOT_containerized_e2e_tests' is required but was not found on this pull request.
[ERROR] Status 'aos-ci-jenkins/OS_unit_tests' is in the 'failure' state. This status must succeed for a merge.
```

This is not the case AFAICT :confused:

sdodson:
[merge]

kwoodson:
woot!

PR #3630, opened 2017-03-11
Add etcd volume size check

juanvallejo:
Related Trello card: https://trello.com/c/S1tuBXjh/70-etcd-health-checks-diagnostics?menu=filter&filter=[diagnostics]

~Work in progress. Just something to begin planning how an etcd storage check could go.
The idea would be to fail the check / warn if the space remaining on an etcd volume is more than 50% full.~

This check ensures the disk size for an etcd host does not exceed a defined limit. Limit defaults to 90% of total disk size, as defined by `ansible_facts`.

While this won't be a direct measure of how much space stored objects in etcd are taking up specifically, 
it would preemptively warn before the volume gets full enough to begin causing issues.


roles/openshift_health_checker/openshift_checks/etcd_volume_check.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @staticmethod
+    def get_etcd_mountpath(ansible_mounts):
+        # priority list - in descending priority order
+        valid_etcd_mount_paths = ["/var/lib/etcd", "/var/lib", "/var", "/"]
+
+        if not ansible_mounts:
+            return None
+
+        etcd_mount_path_idx = None
```
sosiouxme: Finding an index smells wrong. I'd suggest doing what I always do when I'm looking for something, build a hash out of it.

```
mount_for_path = {}
for mnt in ansible_mounts:
    mount_for_path[mnt.get("mount")] = mnt
for path in valid_etcd_mount_paths:
    if path in mount_for_path:  #first one we found wins
        return mount_for_path[path] # note, return the whole thing instead of just the name
return None
```

roles/openshift_health_checker/openshift_checks/etcd_volume_check.py
```diff
@@ -0,0 +1,68 @@
...
+from openshift_checks import OpenShiftCheck, get_var
+
+
+class EtcdImageData(OpenShiftCheck):
+    """Check that etcd volume capacity is not over 50%"""
+
+    name = "etcd_volume_check"
+    tags = ["etcd_volume_check"]
+
+    def run(self, _, task_vars):
```
rhcarvalho: In this case I'd keep consistent with the "interface" we're implementing... in the absence of explicit types or other hints, I'd say this is more idiomatic:

```python
```

This is the same signature as action plugins.

We should be passing `tmp` forward to a module that runs in the target.

roles/openshift_health_checker/openshift_checks/etcd_volume_check.py
```diff
@@ -0,0 +1,68 @@
...
+        ansible_mounts = get_var(task_vars, "ansible_mounts")
+        etcd_mountpath = self.get_etcd_mountpath(ansible_mounts)
+        etcd_diskspace_avail = self.get_etcd_diskspace(etcd_mountpath, ansible_mounts)
+        etcd_totaldiskspace = self.get_etcd_totaldiskspace(etcd_mountpath, ansible_mounts)
```
rhcarvalho: Please correct me if I'm mistaken, these seen to be checking the space in the machine where ansible-playbook is run?

I'd expect we be executing an Ansible module only in the hosts that are part of the `etcd` group.

sosiouxme: Depends, is `ansible_mounts` scoped to each host by the time it gets here? It wouldn't make a lot of sense for it to be the mounts on the machine running ansible.

But this should definitely be limited to only hosts in the etcd group.

rhcarvalho: Hmmm yeah, makes sense. It might be the case that the `setup` module already figured all those facts for the hosts.

detiber: To use the values from the `setup` module you will need to ensure they are passed into the module invocation from the action plugin. Another option, which is probably a bit cleaner would be to leverage `module_utils` to re-gather the needed facts at runtime: https://github.com/ansible/ansible/blob/v2.2.1.0-1/lib/ansible/module_utils/facts.py

sosiouxme: Technically we're not in a module here, though, still in the action plugin, right?

roles/openshift_health_checker/openshift_checks/etcd_volume_check.py
juanvallejo: @sosiouxme Thanks, I like the idea of using a hashmap instead. I believe that even in that case, though, we would still need to loop through all available mount paths. If we find `/` first, we would not want to return `valid_etcd_mount_paths["/"]` right away, if a more specific `valid_etcd_mount_paths["/var/lib/etcd"]` also exists

roles/openshift_health_checker/openshift_checks/etcd_volume_check.py
juanvallejo: @rhcarvalho 
> Please correct me if I'm mistaken, these seen to be checking the space in the machine where ansible-playbook is run?
>
> I'd expect we be executing an Ansible module only in the hosts that are part of the etcd group.

Double checked by comparing the output of `ansible_mounts` with the output of `lsblk` on the remote host. They certainly seem to match up:

Output of `ansible_mounts`:
```
[{device': /dev/xvda2',
fstype': xfs',
mount': /',
options': rw,seclabel,relatime,attr2,inode64,noquota',
size_available': 21110980608,
size_total': 26830942208},
{'device': /dev/mapper/docker-openshift--xfs--vol--dir',
fstype': xfs',
mount': /mnt/openshift-xfs-vol-dir',
options': rw,seclabel,relatime,attr2,inode64,grpquota',
size_available': 8004644864,
size_total': 8038383616}]
```

Output of `lsblk` on remote host:
```
NAME                                MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
xvda                                202:0    0   25G  0 disk
├─xvda1                             202:1    0    1M  0 part
└─xvda2                             202:2    0   25G  0 part /
xvdb                                202:16   0   25G  0 disk
└─xvdb1                             202:17   0   25G  0 part
  ├─docker-openshift--xfs--vol--dir 253:0    0  7.5G  0 lvm  /mnt/openshift-xfs-vol-dir
  ├─docker-docker--pool_tmeta       253:1    0   28M  0 lvm
  │ └─docker-docker--pool           253:3    0    7G  0 lvm
  └─docker-docker--pool_tdata       253:2    0    7G  0 lvm
    └─docker-docker--pool           253:3    0    7G  0 lvm
```

juanvallejo:
@sosiouxme or @rhcarvalho any further comments on this? Thanks!

roles/openshift_health_checker/openshift_checks/etcd_volume_check.py
```diff
@@ -0,0 +1,49 @@
+# pylint: disable=missing-docstring
+"""
+Ansible module for determining information about etcd volume size.
+"""
+
+from openshift_checks import OpenShiftCheck, get_var
+
+
+class EtcdVolumeCheck(OpenShiftCheck):
+    """Check that etcd volume capacity is not over 50%"""
```
rhcarvalho: The comment seems outdated?

roles/openshift_health_checker/openshift_checks/etcd_volume_check.py
```diff
@@ -0,0 +1,49 @@
...
+
+    name = "etcd_volume_check"
```
rhcarvalho: No need for `_check` in the name, otherwise all or most names would have that same suffix.

roles/openshift_health_checker/openshift_checks/etcd_volume_check.py
```diff
@@ -0,0 +1,49 @@
...
+    tags = ["etcd_volume_check", "etcd"]
```
rhcarvalho: - A tag should be something more generic that would describe a group of checks, let's no include the name of the check as a tag.
- `etcd` is good, we may have more checks related to etcd :+1: 

When is this check useful to be run? We could think of a tag based on that.
@sosiouxme suggestions?

roles/openshift_health_checker/openshift_checks/etcd_volume_check.py
```diff
@@ -0,0 +1,49 @@
...
+
+        return {"changed": False}
+
+    @staticmethod
+    def get_etcd_mountpath(ansible_mounts):
+        # priority list - in descending priority order
+        valid_etcd_mount_paths = ["/var/lib/etcd", "/var/lib", "/var", "/"]
+
+        if not ansible_mounts:
+            return None
```
rhcarvalho: By returning `None` here and in the last statement of this method, code that uses the return value can raise `TypeError` -- see `etcd_diskspace_avail = etcd_mountpath.get("size_available", 0)` above.

roles/openshift_health_checker/openshift_checks/etcd_volume_check.py
```diff
@@ -0,0 +1,49 @@
...
+    """Check that etcd volume capacity is not over 50%"""
+
+    name = "etcd_volume_check"
+    tags = ["etcd_volume_check", "etcd"]
+
+    def run(self, tmp, task_vars):
+        ansible_mounts = get_var(task_vars, "ansible_mounts")
+        etcd_mountpath = self.get_etcd_mountpath(ansible_mounts)
+        etcd_diskspace_avail = etcd_mountpath.get("size_available", 0)
+        etcd_totaldiskspace = etcd_mountpath.get("size_total", 0)
```
rhcarvalho: Naming: the usage of `_` is inconsistent in the variables and methods referred above.

roles/openshift_health_checker/openshift_checks/etcd_volume_check.py
```diff
@@ -0,0 +1,49 @@
...
+
+        # max disk usage limit
+        diskspace_usage_lim = get_var(task_vars, "etcd_volume_usage_limit_bytes")
```
rhcarvalho: I gather the usage of this check involves somehow setting the `etcd_volume_usage_limit_bytes` variable in the playbook or somewhere else?
We need to give some thought to the UX.

juanvallejo: Thanks, I was not entirely sure how to address this, but for now, I have defaulted this value to 90% of the total disk space

juanvallejo:
@rhcarvalho thanks, unit tests added

rhcarvalho:
@juanvallejo please update the PR description.

rhcarvalho:
@juanvallejo please squash the commits.

rhcarvalho:
Eventually this is a lot similar to the existing disk availability check, the difference being the path / mount point that is considered?
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_health_checker/openshift_checks/disk_availability.py

There is some opportunity for code reuse here.

rhcarvalho:
Pending updating the description and squashing commits, LGTM.
As per my previous comment, we may share code between checks, but that could be a follow up.

juanvallejo:
@rhcarvalho 
> Pending updating the description and squashing commits, LGTM.

Thanks, squashed commits and updated the PR's description

sdodson:
If the host has one filesystem mounted at / and there's 10GB in use on that file system but not necessarily in /var/lib/etcd won't this incorrectly determine etcd's size to be 10GB?

juanvallejo:
@rhcarvalho or @sosiouxme any further comments on this? Thanks!

roles/openshift_health_checker/test/etcd_volume_test.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+
+    if task_vars["etcd_disk_size_limit_bytes"] is None:
+        task_vars.pop("etcd_disk_size_limit_bytes", None)
+
+    check = EtcdVolume(execute_module=fake_execute_module)
+    result = check.run(tmp=None, task_vars=task_vars)
+
+    assert result['failed']
+    for word in 'greater than the recommended'.split() + extra_words:
```
rhcarvalho: Hmm, looks like a misuse of a pattern here :)

This will look for the strings "greater", "than", "the" and "recommended" in no particular order (and not necessarily individual words, e.g., "there" contains "the"). I don't see how good it is to look for a generic string such as "the" in the output. Sure not harmful, but perhaps intention and actual code behavior may not match.

roles/openshift_health_checker/test/etcd_volume_test.py
```diff
@@ -0,0 +1,131 @@
...
+        assert word in result['msg']
+
+
+def fake_execute_module(*args):
+    raise AssertionError('this function should not be called')
```
rhcarvalho: Missing newline character at the end of the file (UNIX convention). This can be configured in your code editor.

roles/openshift_health_checker/test/etcd_volume_test.py
```diff
@@ -0,0 +1,131 @@
...
+        'size_total': 0,
+    }, {
+        # not enough space on /var/lib ...
+        'mount': '/var/lib',
+        'size_available': 2 * 10**9 + 1,
+        'size_total': 21 * 10**9 + 1,
+    }, {
+        # ... but enough on /var/lib/etcd
+        'mount': '/var/lib/etcd',
+        'size_available': 36 * 10**9 + 1,
```
rhcarvalho: Are the `+ 1`'s in the test cases here actually relevant?

I think in call cases the default of 90% of the total will be used as the limit? The plus ones seem to be just distraction, no?

roles/openshift_health_checker/test/etcd_volume_test.py
```diff
@@ -0,0 +1,131 @@
...
+        'size_total': 40 * 10**9
+    }],
+])
+def test_succeeds_with_recommended_disk_space(ansible_mounts):
```
rhcarvalho: The test below (`test_fails_with_insufficient_disk_space`) takes `size_limit` as argument and test both the default 90% and custom values to consider check failures.
Seems inconsistent that for the positive case we only test with the default of 90%, and no custom `size_limit`.

roles/openshift_health_checker/test/etcd_volume_test.py
rhcarvalho: This actually fails `flake8`: https://travis-ci.org/openshift/openshift-ansible/jobs/225309030#L192

juanvallejo:

juanvallejo:
@rhcarvalho or @sosiouxme wondering if this PR is good to go? Thanks!

aos-ci-test

roles/openshift_health_checker/openshift_checks/etcd_volume.py
```diff
@@ -0,0 +1,54 @@
+# pylint: disable=missing-docstring
```
sosiouxme: [PyLint rules MUST NOT be disabled on a whole file.](https://github.com/openshift/openshift-ansible/blob/master/docs/best_practices_guide.adoc#pylint)

And I suppose I'd rather have the docstrings, or make the methods internal, than have this everywhere.

[Also from the same doc](https://github.com/openshift/openshift-ansible/blob/master/docs/best_practices_guide.adoc#python-source-files):

    # vim: expandtab:tabstop=4:shiftwidth=4

roles/openshift_health_checker/openshift_checks/etcd_volume.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+from openshift_checks import OpenShiftCheck, OpenShiftCheckException, get_var
+
+
+class EtcdVolume(OpenShiftCheck):
+    """Ensure disk size for an etcd host does not exceed a defined limit"""
+
+    name = "etcd_volume"
+    tags = ["etcd"]
```
sosiouxme: should also be a `"health"` check

roles/openshift_health_checker/openshift_checks/etcd_volume.py
```diff
@@ -0,0 +1,54 @@
...
+        etcd_disk_size_available = int(etcd_mount_path["size_available"])
+        etcd_disk_size_total = int(etcd_mount_path["size_total"])
+        etcd_disk_size_used = etcd_disk_size_total - etcd_disk_size_available
+
+        self.etcd_default_size_limit = 0.9 * etcd_disk_size_total
+
+        etcd_disk_size_limit = int(get_var(task_vars, "etcd_max_disk_bytes", default=self.etcd_default_size_limit))
+
+        if etcd_disk_size_used > etcd_disk_size_limit:
+            msg = "Current etcd volume usage ({actual:.2f} GB) is greater than the recommended limit ({limit:.2f} GB)."
```
sosiouxme: We don't really have a "recommended" limit. probably just "the limit" would be good here. contemplating whether this is also a good place to tell them how to configure the limit.

Also this should mention what the etcd volume turned out to be.

roles/openshift_health_checker/openshift_checks/etcd_volume.py
rhcarvalho: > PyLint rules MUST NOT be disabled on a whole file.
> 
> And I suppose I'd rather have the docstrings, or make the methods internal, than have this everywhere.

I started doing this in checks because every check has a `run` method, and it doesn't make sense to 1) make it "internal" `_run`; nor 2) add a docstring for it in every subclass of `OpenShiftCheck`.

I don't really like dirtying the file with pylint comments. Once I proposed we disable `missing-docstring` globally, but it did not get the discussion it required to go forward.

Thoughts?

> Also from the same doc:
> 
> \# vim: expandtab:tabstop=4:shiftwidth=4

This one smells like something we may want to remove from the rules. I am not sure, but I think in some PR I got a sign from @detiber that this is not something we insist in.

In a project where contributors are free to use whatever editor they want, and we have linting tools that verify the proper formatting of files, I'd say that `vim`-specific line is not needed.

rhcarvalho: FWIW I'm proposing removing the vim config line in #4123.

roles/openshift_health_checker/openshift_checks/etcd_volume.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def run(self, tmp, task_vars):
+        ansible_mounts = get_var(task_vars, "ansible_mounts")
+
+        etcd_mount_path = self.get_etcd_mount_path(ansible_mounts)
+        etcd_disk_size_available = int(etcd_mount_path["size_available"])
+        etcd_disk_size_total = int(etcd_mount_path["size_total"])
+        etcd_disk_size_used = etcd_disk_size_total - etcd_disk_size_available
+
+        self.etcd_default_size_limit = 0.9 * etcd_disk_size_total
```
rhcarvalho: The current shape of the code makes it harder to understand at a glance what the default limit is:

```python

class ...:
    etcd_default_size_limit = 0   # §1

    def run(...):
        ...
        self.etcd_default_size_limit = 0.9 * etcd_disk_size_total  # §2
```

Whoever stares at **§1** will wonder "hmm, the default limit is zero".

But, in fact, if one keeps reading through `run` until **§2**, one notices that the "default default" is unconditionally updated at runtime to be 90% of some other value.

## Proposal

Since we are making the default be 90% of the total size, and based on a previous experience doing a similar task, I believe that a perhaps better initial interface would be to set the limit in percentage, instead of absolute values.

Such that we could write:

```python
class ...:
    default_etcd_size_limit_percent = 0.9

    def run(...):
        ...
        etcd_size_limit_percent = float(get_var(task_vars, "etcd_size_limit_percent", default=self.default_etcd_size_limit_percent))

        if etcd_disk_size_used / etcd_disk_size_total > etcd_size_limit_percent:
            ...
```

This would make it easier for readers to see what the default is, and for users to specify values such as "0.85" (or "85"?), instead of number of bytes like "34000000".

juanvallejo:

juanvallejo:
@sosiouxme 
> we need a standard format for variable names we expect to be passed into checks. names the user puts in inventory could be different.

do we have an agreed format to follow for these for the time being?

sosiouxme:
Thanks for updating the best practices guide when removing the vim line :)

While I agree that "run" doesn't really need doc nor to be underscored, by
disabling docstring file-wide as boilerplate it means there's no
encouragement to document (or internalize) other methods either. And I kind
of like having some idea what a method does and/or returns without having
to look down through it. I don't think it would be too ugly to add a short
docstring across all the subclasses for the `run` method, e.g. """The entry
point for running the check. Returns a result hash."""

Or if we really don't want to do this, then at least amend the best
practices to make an exception for `missing-docstring` at the beginning of
the file.

sosiouxme:
I was similarly scratching my head looking at this, and I think this is a
good approach.

BTW users generally prefer to think in terms of percentages, not ratios, so
expect values like "85" or "99.5" and convert. `float(var)/100`

rhcarvalho:
The more I look at this, the more I think this should be just a simple variation of the existing disk check....

rhcarvalho:
> The more I look at this, the more I think this should be just a simple variation of the existing disk check....

Well, a difference being this is dealing with percentage of used space versus the other deals with recommended sizes in bytes...

rhcarvalho:
> BTW users generally prefer to think in terms of percentages, not ratios, so
> expect values like "85" or "99.5" and convert. `float(var)/100`

:+1:

rhcarvalho:
> there's no encouragement to document (or internalize) other methods either

I'd definitely still encourage and ask for relevant and correct documentation where missing. The inconvenience is having to please tools that do half of the work.

Your suggestions are good, let's give it a go.

rhcarvalho:
@sosiouxme looks like we have green travis even without the "pylint disable" && no docstring on `run`.
How about that, good?

rhcarvalho:
@juanvallejo  @sosiouxme sorry for the big changes in the last commit, I wanted to send it as a PR to Juan's branch but the GitHub UI gave me a 404... and in the fear of losing it I went back and committed...
Feel free to discuss / give feedback on it / revert all or part.

roles/openshift_health_checker/openshift_checks/etcd_volume.py
```diff
@@ -0,0 +1,54 @@
...
+    tags = ["etcd", "health"]
+
+    default_etcd_device_usage_threshold_percent = 90
+    # where to find ectd data, higher priority first.
+    supported_mount_paths = ["/var/lib/etcd", "/var/lib", "/var", "/"]
+
+    @classmethod
+    def is_active(cls, task_vars):
+        # TODO: only execute this check on hosts in the 'ectd' group?
+        # Maybe also 'masters' if there are no standalone etcd hosts?
```
rhcarvalho: @juanvallejo @sosiouxme what do you guys think? Originally the docstring of the class hinted that this was only run on "etcd hosts"...

juanvallejo: I agree,  maybe we could check if `task_vars.openshift.master.etcd_hosts` exists and if `"etcd" in task_vars.groups`

juanvallejo:
@rhcarvalho @sosiouxme 
Looks like it is still requiring a doc-string: https://travis-ci.org/openshift/openshift-ansible/jobs/230888805#L214

rhcarvalho:
> Looks like it is still requiring a doc-string: https://travis-ci.org/openshift/openshift-ansible/jobs/230888805#L214

Right...

```
roles/openshift_health_checker/openshift_checks/etcd_volume.py:1: [C0111(missing-docstring), ] Missing module docstring
```

This is where I disagree with the pylint settings... The documentation for the module and the `OpenShiftCheck` subclass it contains should be about the same thing, do we really want to add docstrings just to satisfy the tool?

roles/openshift_health_checker/openshift_checks/etcd_volume.py
sosiouxme: Yes, we should do this, as in the logging stack check.

A quick read of https://docs.openshift.org/latest/install_config/install/advanced_install.html#single-master seems to indicate that etcd hosts will always be defined, but I just ran a single-master install without any and it worked fine.

I think maybe the logic should be something like:

    etcd_hosts = get_var(task_vars, "groups", "etcd", default=[]) or get_var(task_vars, "groups", "masters", default=[]) or []
    if get_var(task_vars, "ansible_ssh_host") in etcd_hosts:
        ...

This PR is not the place, but we could think a little about generalizing this logic of which subset of hosts a check runs against. I think it's likely this is only the place where a task does something like this (as opposed to the play that includes it).

sosiouxme:
I don't like disabling docstring checking for the whole file. So for each
pylint docstring complaint on something that seems unnecessary we can
either have a comment disabling it for that line or just put in a
docstring. Personally I'd prefer a single low-content docstring rather than
the pylint comment.

To me it's pretty much that or just disable `C0111(missing-docstring)`
globally. Of course the main goal to me is to have helpful docstrings on
any non-obvious methods, and there's no pylint for whether the doc is
helpful anyway :)

sosiouxme:
On Tue, May 9, 2017 at 5:23 PM, Juan Vallejo <notifications@github.com>
wrote:
>
> we need a standard format for variable names we expect to be passed into
> checks. names the user puts in inventory could be different.
>
> do we have an agreed format to follow for these for the time being?
>
I asked about this for the ansible arch meeting and there was some lively
discussion... seems we came to a consensus:
https://docs.google.com/document/d/11V3tvTsVVNneFhhhTVGpiZOzqNCQK5GS239ddu4lWjE/edit#heading=h.mqbw6sr1m917

The idea seems to be that roles only use variables specific to them, so
there's a functional separation like an API specification. To accomplish
this, any variable a role uses should be "namespaced" with r_role_name_<>.
Any outside values a role wishes to use should be explicitly set when
invoking the role. Our check action being defined in a role, it would
follow that we do the same.

So e.g. for the logging check, we would not access
`openshift_hosted_logging_deploy` but rather set a variable on the role
`r_openshift_health_checker_logging_deployed:
{{openshift_hosted_logging_deploy}}`. In this way we're being explicit
about what we're using. As far as defining inventory variables for
thresholds, the only requirement is that it begin with `openshift_` and not
conflict with anything existing.

This also seems like it could be really bloated and awkward as I think
about it. But maybe not. Want to be guinea pigs?

roles/openshift_health_checker/openshift_checks/etcd_volume.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def run(self, tmp, task_vars):
+        mount_info = self._etcd_mount_info(task_vars)
+        available = mount_info["size_available"]
+        total = mount_info["size_total"]
+        used = total - available
+
+        threshold = get_var(
+            task_vars,
+            "r_openshift_health_checker_etcd_device_usage_threshold_percent",
```
rhcarvalho: @sosiouxme thanks for discussing the standard! Let's give it a try.

How about we add:
- In the example inventories, add a section to hold a list of variables used by the checks (they can be all commented-out)
- Somewhere, a conversion from `"openshift_etcd_device_usage_threshold_percent"` to `"r_openshift_health_checker_etcd_device_usage_threshold_percent"` -- @sosiouxme was it part of the discussion where this conversion belongs?
  If we do it at the role level, then the role will need to know about all variables of all existing checks. If we do it here, then it defeats the purpose of having the two names and I'd think why not simply use `"openshift_etcd_device_usage_threshold_percent"` directly?!

BTW, should we consider that "we're doing it wrong" when we read inventory variables like `openshift_deployment_type` and others?!

sosiouxme: Example inventories are e.g. https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.ose.example - yes, I think we should start adding lines to these explaining check parameters.

I think it would be cleanest to do the conversion from external variables to internal in a vars file in the role - gives you one place to look to see what's coming in. I think we can skip translating globally available ansible stuff like `groups` and `ansible_ssh_host` and that sort of thing, but should still translate `openshift_deployment_type` and stuff introduced to `openshift.common.*` by `openshift_facts`.

That does mean in the long run we would scour all the checks for what they're using and explicitly name them in that one vars file. I think that would be a separate PR that we open up for discussion so the other ansible devs can see what an actual implementation would look like to confirm this is what we want.

For the purposes of this PR I'd say let's just do that for this check's inventory parameters and get this merged.

juanvallejo:
@sosiouxme 
> For the purposes of this PR I'd say let's just do that for this check's inventory parameters and get this merged.

Added `roles/openshift_health_checker/vars/main.yml` to map internal role vars for this PR, and https://github.com/openshift/openshift-ansible/pull/3660

juanvallejo:
@sosiouxme per our conversation, I am reverting var names back from role-specific format in this PR and in https://github.com/openshift/openshift-ansible/pull/3660

Rather than map external (inventory) vars to internal names, for now, these checks will just use inventory vars directly

juanvallejo:
@sosiouxme wondering if this PR is good to merge, thanks!

sosiouxme:
aos-ci-test

sdodson:
aos-ci-test

sosiouxme:
[merge]

sosiouxme:
also failed on integration test

sosiouxme:
Flake: https://github.com/openshift/origin/issues/14254

[merge]

PR #3631, opened 2017-03-13
validate excluders on non-atomic hosts only

ingvagabund:
Don't check for available excluders in atomic host since we don't support them on AH.


ingvagabund:
aos-ci-test

mtnbikenc:
[merge]

PR #3632, opened 2017-03-13
Update metrics public url 1.5

jcantrill:
@sdodson I intended to squash the source PR into a single commit but it was merged before that happened.  Do you want me to do that hear or leave it as is?


sdodson:
I guess just leave it as is so there's parity between the branches.

jcantrill:
@sdodson merge at your discretion then...

PR #3636, opened 2017-03-13
Missing tlsv1.2 argument to curl 

obuisson:
Add --tlsv1.2 to curl command (installation failed on RHEL 7.1) 
Related to **https://github.com/openshift/openshift-ansible/pull/2707**

### DIFF
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -249,7 +249,7 @@
   # Using curl here since the uri module requires python-httplib2 and
   # wait_for port doesn't provide health information.
   command: >
-    curl --silent
+    curl --silent --tlsv1.2
     {% if openshift.common.version_gte_3_2_or_1_2 | bool %}
     --cacert {{ openshift.common.config_base }}/master/ca-bundle.crt
     {% else %}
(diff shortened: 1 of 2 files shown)


sdodson:
aos-ci-test

sdodson:
[merge]

roles/openshift_master/tasks/main.yml
```diff
@@ -249,7 +249,7 @@
   # Using curl here since the uri module requires python-httplib2 and
   # wait_for port doesn't provide health information.
   command: >
-    curl --silent
+    curl --silent --tlsv1.2
```
detiber: Should this be handled in a more dynamic fashion, especially with users being able to specify tls config now?

sdodson: All OpenShift versions support TLS v1.2 regardless of the minimum version that's now configurable in 3.5. This is working around a bug in Curl from RHEL 7.1 that did not properly negotiate up to v1.2, so I think it's safe.

detiber: @sdodson are there other places where not specifying the tls version will be a problem? This might be a good reason to accelerate a change to using the uri module for testing service health as well.

PR #3639, opened 2017-03-13
Cleanup of openshift_version tasks

ashcrow:
- ``when`` in openshift_version tasks are multiline
- Add openshift_version fact fallback debug messages
- Verify ``openshift_image_tag`` is valid during openshift_version main
- Add debug message when ``openshift_image_tag`` is defined for non container installs
- Remove ``atomic-enterprise`` deployment type from documentation

### DIFF
--- a/roles/openshift_version/tasks/main.yml
+++ b/roles/openshift_version/tasks/main.yml
@@ -9,16 +9,36 @@
 # be used by default. Users must indicate what they want.
 - fail:
     msg: "Must specify openshift_release or openshift_image_tag in inventory to install origin. (suggestion: add openshift_release=\"1.2\" to inventory)"
-  when: is_containerized | bool and openshift.common.deployment_type == 'origin' and openshift_release is not defined and openshift_image_tag is not defined
+  when:
+  - is_containerized | bool
+  - openshift.common.deployment_type == 'origin'
+  - openshift_release is not defined
+  - openshift_image_tag is not defined
 
 # Normalize some values that we need in a certain format that might be confusing:
 - set_fact:
+    openshift_release: "{{ openshift_release[1:] }}"
+  when:
+  - openshift_release is defined
+  - openshift_release[0] == 'v'
+
+- set_fact:
+    openshift_release: "{{ openshift_release | string }}"
+  when: openshift_release is defined
+
+# Add a v to the start of the openshift_image_tag if one was not provided
+- set_fact:
     openshift_image_tag: "{{ 'v' + openshift_image_tag }}"
-  when: openshift_image_tag is defined and openshift_image_tag[0] != 'v' and openshift_image_tag != 'latest'
+  when:
+  - openshift_image_tag is defined
+  - openshift_image_tag[0] != 'v'
+  - openshift_image_tag != 'latest'
 
 - set_fact:
     openshift_pkg_version: "{{ '-' + openshift_pkg_version }}"
-  when: openshift_pkg_version is defined and openshift_pkg_version[0] != '-'
+  when:
+  - openshift_pkg_version is defined
+  - openshift_pkg_version[0] != '-'
 
 # Make sure we copy this to a fact if given a var:
 - set_fact:
@@ -30,7 +50,10 @@
 - name: Use openshift.common.version fact as version to configure if already installed
   set_fact:
     openshift_version: "{{ openshift.common.version }}"
-  when: openshift.common.version is defined and openshift_version is not defined and openshift_protect_installed_version | bool
+  when:
+  - openshift.common.version is defined
+  - openshift_version is not defined
+  - openshift_protect_installed_version | bool
 
 - name: Set openshift_version for rpm installation
   include: set_version_rpm.yml
@@ -67,13 +90,18 @@
 
 - fail:
     msg: "No OpenShift version available, please ensure your systems are fully registered and have access to appropriate yum repositories."
-  when: not is_containerized | bool and openshift_version == '0.0'
+  when:
+  - not is_containerized | bool
+  - openshift_version == '0.0'
 
 # We can't map an openshift_release to full rpm version like we can with containers, make sure
 # the rpm version we looked up matches the release requested and error out if not.
 - fail:
     msg: "Detected OpenShift version {{ openshift_version }} does not match requested openshift_release {{ openshift_release }}. You may need to adjust your yum repositories, inventory, or run the appropriate OpenShift upgrade playbook."
-  when: not is_containerized | bool and openshift_release is defined and not openshift_version.startswith(openshift_release) | bool
+  when:
+  - not is_containerized | bool
+  - openshift_release is defined
+  - not openshift_version.startswith(openshift_release) | bool
 
 # The end result of these three variables is quite important so make sure they are displayed and logged:
 - debug: var=openshift_release
--- a/roles/openshift_version/tasks/main.yml
+++ b/roles/openshift_version/tasks/main.yml
@@ -57,20 +57,6 @@
   - openshift_image_tag is defined
   - openshift_image_tag != "latest"
 
-# Add a v to the start of the openshift_image_tag if one was not provided
-- set_fact:
-    openshift_image_tag: "{{ 'v' + openshift_image_tag }}"
-  when:
-  - openshift_image_tag is defined
-  - openshift_image_tag[0] != 'v'
-  - openshift_image_tag != 'latest'
-
-- set_fact:
-    openshift_pkg_version: "{{ '-' + openshift_pkg_version }}"
-  when:
-  - openshift_pkg_version is defined
-  - openshift_pkg_version[0] != '-'
-
 # Make sure we copy this to a fact if given a var:
 - set_fact:
     openshift_version: "{{ openshift_version | string }}"
--- a/roles/openshift_version/tasks/main.yml
+++ b/roles/openshift_version/tasks/main.yml
@@ -94,6 +94,18 @@
   include: set_version_containerized.yml
   when: is_containerized | bool
 
+# Warn if the user has provided an openshift_image_tag but is not doing a containerized install
+# NOTE: This will need to be modified/removed for future container + rpm installations work.
+- name: Warn if openshift_image_tag is defined when not doing a containerized install
+  debug:
+    msg: >
+      openshift_image_tag is used for containerized installs. If you are trying to
+      specify an image for a non-container install see oreg_url.
+  when:
+  - not is_containerized | bool
+  - openshift_image_tag is defined
+
+
 # At this point we know openshift_version is set appropriately. Now we set
 # openshift_image_tag and openshift_pkg_version, so all roles can always assume
 # each of this variables *will* be set correctly and can use them per their
--- a/roles/openshift_version/tasks/main.yml
+++ b/roles/openshift_version/tasks/main.yml
@@ -26,6 +26,37 @@
     openshift_release: "{{ openshift_release | string }}"
   when: openshift_release is defined
 
+# Verify that the image tag is in a valid format
+- block:
+
+  # Verifies that when the deployment type is origin the version:
+  # - starts with a v
+  # - Has 3 integers seperated by dots
+  # It also allows for optional trailing data which:
+  # - must start with a dash
+  # - may contain numbers, letters, dashes and dots.
+  - assert:
+      that:
+      - "{{ openshift_image_tag|match('(^v?\\d+\\.\\d+\\.\\d+(-[\\w\\-\\.]*)?$)') }}"
+      msg: "openshift_image_tag must be in the format v#.#.#[-optional.#]. Examples: v1.2.3, v3.5.1-alpha.1"
+    when: openshift.common.deployment_type == 'origin'
+
+  # Verifies that when the deployment type is openshift-enterprise the version:
+  # - starts with a v
+  # - Has at least 2 integers seperated by dots
+  # It also allows for optional trailing data which:
+  # - must start with a dash
+  # - may contain numbers
+  - assert:
+      that:
+      - "{{ openshift_image_tag|match('(^v\\d+\\.\\d+[\\.\\d+]*(-\\d+)?$)') }}"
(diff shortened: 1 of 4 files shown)


ashcrow:
Updating to allow for ``-$INFO`` now.

roles/openshift_version/tasks/main.yml
```diff
@@ -9,24 +9,43 @@
...
+  - openshift_release[0] == 'v'
 
 - set_fact:
     openshift_release: "{{ openshift_release | string }}"
   when: openshift_release is defined
 
+# Verify that the image tag is in a valid format
+- assert:
+    that:
+      - "{{ openshift_image_tag|match('(^[v]?\\d+\\.\\d+[\\.\\d+]*-?\\w*$)') }}"
```
stevekuznetsov: Does that regex match e.g. `v1.5.0-alpha.3`? Is there a semver/other version parsing util we can use, to say "starts with `v` and the rest is a valid version" without regex?

stevekuznetsov: Also does this match `latest` ?

roles/openshift_version/tasks/main.yml
```diff
@@ -9,24 +9,43 @@
...
+    msg: "openshift_image_tag must be in the format [v]#.#[.#...]. Examples: 1.2, v3.4.1, 3.5.1.3"
+  when: openshift_image_tag is defined
+
+# Add a v to the start of the openshift_image_tag if one was not provided
```
stevekuznetsov: Either assert or do this, can't do both

roles/openshift_version/tasks/main.yml
```diff
@@ -9,24 +9,43 @@
...
     openshift_image_tag: "{{ 'v' + openshift_image_tag }}"
-  when: openshift_image_tag is defined and openshift_image_tag[0] != 'v' and openshift_image_tag != 'latest'
+  when:
+  - openshift_image_tag is defined
+  - openshift_image_tag[0] != 'v'
+  - openshift_image_tag != 'latest'
 
 - set_fact:
     openshift_pkg_version: "{{ '-' + openshift_pkg_version }}"
-  when: openshift_pkg_version is defined and openshift_pkg_version[0] != '-'
```
stevekuznetsov: It's a little too late for this ... but it seems like exposing a very small surface for these vars would be better than trying to accept the world and deal with it.

roles/openshift_version/tasks/main.yml
```diff
@@ -75,13 +107,18 @@
...
 
 # We can't map an openshift_release to full rpm version like we can with containers, make sure
 # the rpm version we looked up matches the release requested and error out if not.
 - fail:
     msg: "Detected OpenShift version {{ openshift_version }} does not match requested openshift_release {{ openshift_release }}. You may need to adjust your yum repositories, inventory, or run the appropriate OpenShift upgrade playbook."
-  when: not is_containerized | bool and openshift_release is defined and not openshift_version.startswith(openshift_release) | bool
+  when:
+  - not is_containerized | bool
+  - openshift_release is defined
+  - not openshift_version.startswith(openshift_release) | bool
```
stevekuznetsov: Are these the normal `NEVRA` versions? How does the version start with the release?

stevekuznetsov:
I think it may be too late to change this API but it seems to me like there's too much "niceness" in this -- for something as important as the version to be installed, I think it would be better to expect the user to give you the _exact_ version they want and not try to work around stuff, as it will inevitably lead to a weird cluster installed with a version they didn't want.

roles/openshift_version/tasks/main.yml
ashcrow: > Does that regex match e.g. v1.5.0-alpha.3? Is there a semver/other version parsing util we can use, to say "starts with v and the rest is a valid version" without regex?

It wouldn't. It would match v1.5.0-alpha3 though. I can easily add a ``.`` if that is part of the versioning scheme here.

AFAIK there is not another util we can use without pulling in yet another library into the installer which we are trying to avoid.

Good catch on latest. I'll add a ``when`` that will skip this if ``latest`` is used.

roles/openshift_version/tasks/main.yml
ashcrow: Why not? This is code that was already there :smile: All this does is add a ``v`` is it was missing which the regex doesn't require.

roles/openshift_version/tasks/main.yml
ashcrow: Again, this is old code that I just cleaned up. In this case I just put them on multiple lines instead of one long line. I'd have to trace back through the code to answer this. I can do so if you think there is an issue here.

ashcrow:
@stevekuznetsov I totally agree with the niceness comment. It is way to forgiving. However, unless it's worth making it less forgiving right now I'd say let's focus on failing when it's not nice enough. I'd rather not refactor version parsing at this moment (again, unless it's a high priority, etc..).

detiber:
Adding @dgoodwin as a reviewer, since he has experience with the oddities of version handling, especially wrt mixed containerized/non-containerized environments.

roles/openshift_version/tasks/main.yml
stevekuznetsov: Ah, didn't see `v` was optional above

roles/openshift_version/tasks/main.yml
stevekuznetsov: Yeah, I saw it was old versions -- just had a hard time understanding what exactly was going on here

ashcrow:
Updates are incoming

roles/openshift_version/tasks/main.yml
ashcrow: Totally understand :smile:

roles/openshift_version/tasks/main.yml
```diff
@@ -9,24 +9,51 @@
...
+  # - starts with a v
+  # - Has at least 2 integers seperated by dots
+  # It also allows for optional trailing data which:
+  # - must start with a dash
+  # - may contain numbers
+  - assert:
+      that:
+        - "{{ openshift_image_tag|match('(^v\\d+\\.\\d+[\\.\\d+]*(-\\d+)?$)') }}"
+      msg: "openshift_image_tag must be in the format v#.#[.#[.#]]. Examples: v1.2, v3.4.1, v3.5.1.3, v1.2-1, v1.2.3-4"
+    when: openshift.common.deployment_type == 'ose'
```
stevekuznetsov: Aren't there also some other types? Like `openshift-enterprise`?

ashcrow: Updating to match https://github.com/openshift/openshift-ansible/blob/master/DEPLOYMENT_TYPES.md

roles/openshift_version/tasks/main.yml
dgoodwin: openshift_release=v3.4, openshift_version=v3.4.1.8

roles/openshift_version/tasks/main.yml
```diff
@@ -9,24 +9,51 @@
...
+  - openshift_release is defined
+  - openshift_release[0] == 'v'
 
 - set_fact:
     openshift_release: "{{ openshift_release | string }}"
   when: openshift_release is defined
 
-- set_fact:
-    openshift_image_tag: "{{ 'v' + openshift_image_tag }}"
-  when: openshift_image_tag is defined and openshift_image_tag[0] != 'v' and openshift_image_tag != 'latest'
```
dgoodwin: I think we lost support for "latest" here, which has come up in past with origin users.

roles/openshift_version/tasks/main.yml
```diff
@@ -9,24 +9,51 @@
...
+  # - starts with a v
+  # - Has at least 2 integers seperated by dots
+  # It also allows for optional trailing data which:
+  # - must start with a dash
+  # - may contain numbers
+  - assert:
+      that:
+      - "{{ openshift_image_tag|match('(^v\\d+\\.\\d+[\\.\\d+]*(-\\d+)?$)') }}"
+      msg: "openshift_image_tag must be in the format v#.#[.#[.#]]. Examples: v1.2, v3.4.1, v3.5.1.3, v1.2-1, v1.2.3-4"
+    when: openshift.common.deployment_type == 'openshift-enterprise' or openshift.common.deployment_type == 'atomic-enterprise'
```
dgoodwin: Is atomic-enterprise still a supported deployment type? That might be dead now.

roles/openshift_version/tasks/main.yml
```diff
@@ -48,17 +78,35 @@
   include: set_version_containerized.yml
   when: is_containerized | bool
 
+# Stop if the user has provided an openshift_image_tag but is not doing a containerized install
+- assert:
+    that: openshift_image_tag is not defined
+    message: >
+      openshift_image_tag is used for containerized installs. If you are trying to
+      specify an image for a non-container install see oreg_url.
+  when: not is_containerized | bool
```
dgoodwin: This does look like an issue when we get to mixed container + rpm installations, which are probably broken right now anyhow but it has been discussed as a desirable end state and I know there are open bugs to get it working.

roles/openshift_version/tasks/main.yml
ashcrow: @sdodson ^

roles/openshift_version/tasks/main.yml
ashcrow: I'll update and ignore validation if ``openshift_image_tag`` is set to ``latest``.

roles/openshift_version/tasks/main.yml
ashcrow: Right now the ``openshift_image_tag`` is only honored during containerized installs. I'll add a note that this will have to be removed/modified with the updates to container + rpm installs.

roles/openshift_version/tasks/main.yml
sdodson: It's dead.

roles/openshift_version/tasks/main.yml
```diff
@@ -9,24 +9,53 @@
...
+  # - may contain numbers, letters, dashes and dots.
+  - assert:
+      that:
+      - "{{ openshift_image_tag|match('(^v?\\d+\\.\\d+\\.\\d+(-[\\w\\-\\.]*)?$)') }}"
+      msg: "openshift_image_tag must be in the format v#.#.#[-optional.#]. Examples: v1.2.3, v3.5.1-alpha.1"
+    when: openshift.common.deployment_type == 'origin'
+
+  # Verifies that when the deployment type is (openshift|atomic)-enterprise the version:
+  # - starts with a v
+  # - Has at least 2 integers seperated by dots
```
sdodson: @dgoodwin I think this is 2 integers based on the whole v3.4 tagging so that we may pull whatever is the latest 3.4 based on openshift_release but is that used by openshift_image_tag or should openshift_image_tag always be the exact tag?

dgoodwin: The idea is if you want to just specify a generic v3.4, you use openshift_release, that is mostly just a seed for cluster setup. After that we lookup precise openshift_version, and use it for openshift_image_tag and openshift_pkg_version, pinning *everything* to those specific versions and making sure they never change unexpectedly.

If the used chooses to explicitly override openshift_image_tag we should respect whatever they set, technically v3.4 should work there but it opens the door to some risk you might mistakenly pull a newer v3.4 before you were ready.

ashcrow:
@dgoodwin I made updates based on comments. Let me know if those address your concerns or if more changes are needed before I push this into CI.

ashcrow:
aos-ci-test

ashcrow:
I updated with names for the asserts. I'm thinking this gets called multiple times meaning the ``openshift_image_tag`` may always be set eventually.

ashcrow:
aos-ci-test

dgoodwin:
After openshift_version role is run it's expected that openshift_image_tag, openshift_pkg_version, openshift_version are all set precisely and can be used anywhere that follows.

ashcrow:
aos-ci-test

ashcrow:
[test]

ashcrow:
```
10:06:18     "changed": false, 
10:06:18     "failed": true, 
10:06:18     "msg": "No package matching 'origin-docker-excluder' found available, installed or updated", 
10:06:18     "rc": 126, 
10:06:18     "results": [
10:06:18         "No package matching 'origin-docker-excluder' found available, installed or updated"
10:06:18     ]
10:06:18 }
10:06:18 
10:06:18 PLAY RECAP *********************************************************************
10:06:18  [WARNING]: Could not create retry file '/usr/share/ansible/openshift-
10:06:18 ansible/playbooks/byo/config.retry'.         [Errno 13] Permission denied:
```

Unrelated failure.

ashcrow:
Rebased just in case.

ashcrow:
aos-ci-test

ashcrow:
[test]

stevekuznetsov:
uhhhh @sdodson I thought that was fixed?

ashcrow:
```
10:43:31     "changed": false, 
10:43:31     "failed": true, 
10:43:31     "msg": "No package matching 'origin-docker-excluder' found available, installed or updated", 
10:43:31     "rc": 126, 
10:43:31     "results": [
10:43:31         "No package matching 'origin-docker-excluder' found available, installed or updated"
10:43:31     ]
10:43:31 }
10:43:31  [WARNING]: Could not create retry file '/usr/share/ansible/openshift-
10:43:31 ansible/playbooks/byo/config.retry'.         [Errno 13] Permission denied:
10:43:31 u'/usr/share/ansible/openshift-ansible/playbooks/byo/config.retry'
```

I'll hold off for a bit.

ashcrow:
Rebased again.

ashcrow:
aos-ci-test :crossed_fingers:

ashcrow:
aos-ci-test

ashcrow:
:x:  aos-ci-jenkins/OS_3.5_NOT_containerized_e2e_tests — "e2e tests passed"

:confused:

ashcrow:
rebased

ashcrow:
[test]

ashcrow:
aos-ci-test

ashcrow:
@sdodson if this LGTY please merge.

sdodson:
[merge]

PR #3641, opened 2017-03-13
Revert module_utils six for openshift_health_checker

detiber:
Embedded version of six in Ansible 2.2.1 does not support add_metaclass,
so revert to using python-six.

### DIFF
--- a/roles/openshift_facts/vars/main.yml
+++ b/roles/openshift_facts/vars/main.yml
@@ -2,6 +2,7 @@
 required_packages:
   - iproute
   - python-dbus
+  - python-six
   - PyYAML
   - yum-utils
 
--- a/roles/openshift_health_checker/openshift_checks/__init__.py
+++ b/roles/openshift_health_checker/openshift_checks/__init__.py
@@ -8,9 +8,11 @@ import os
 from abc import ABCMeta, abstractmethod, abstractproperty
 from importlib import import_module
 
+# add_metaclass is not available in the embedded six from module_utils in Ansible 2.2.1
+from six import add_metaclass
 # pylint import-error disabled because pylint cannot find the package
 # when installed in a virtualenv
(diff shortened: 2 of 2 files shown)


detiber:
aos-ci-test

detiber:
Re-push to try and retrigger Travis...  known issue with Travis currently.

detiber:
[merge]

roles/openshift_facts/vars/main.yml
```diff
@@ -2,6 +2,7 @@
 required_packages:
   - iproute
   - python-dbus
+  - python-six
```
rhcarvalho: If we have no other dependencies on six, I'd be happy to drop the metaclass and have one less dependency installed by openshift_facts.
Or do the PY2/PY3 metaclass dance manually, without six.

@detiber, thoughts?

detiber: @rhcarvalho either way works for me, I was mainly trying to address the immediate issue with this PR.

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -8,9 +8,11 @@
 from abc import ABCMeta, abstractmethod, abstractproperty
 from importlib import import_module
 
+# add_metaclass is not available in the embedded six from module_utils in Ansible 2.2.1
+from six import add_metaclass
 # pylint import-error disabled because pylint cannot find the package
 # when installed in a virtualenv
-from ansible.module_utils.six.moves import add_metaclass, reduce  # pylint: disable=import-error, redefined-builtin
```
rhcarvalho: It turns out the problem was just a mistake in the import... `add_metaclass` in not part of `six.moves`, but `six`.
Fixed in #3707.

PR #3643, opened 2017-03-13
Add elasticsearch, fluentd, kibana, curator check

juanvallejo:
Adds health checks for the logging stack components.

Based on https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py.

### DIFF
--- a/roles/openshift_health_checker/library/ocutil.py
+++ b/roles/openshift_health_checker/library/ocutil.py
@@ -1,0 +1,74 @@
+#!/usr/bin/python
+"""Interface to OpenShift oc command"""
+
+import os
+import shlex
+import shutil
+import subprocess
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+ADDITIONAL_PATH_LOOKUPS = ['/usr/local/bin', os.path.expanduser('~/bin')]
+
+
+def locate_oc_binary():
+    """Find and return oc binary file"""
+    # https://github.com/openshift/openshift-ansible/issues/3410
+    # oc can be in /usr/local/bin in some cases, but that may not
+    # be in $PATH due to ansible/sudo
+    paths = os.environ.get("PATH", os.defpath).split(os.pathsep) + ADDITIONAL_PATH_LOOKUPS
+
+    oc_binary = 'oc'
+
+    # Use shutil.which if it is available, otherwise fallback to a naive path search
+    try:
+        which_result = shutil.which(oc_binary, path=os.pathsep.join(paths))
+        if which_result is not None:
+            oc_binary = which_result
+    except AttributeError:
+        for path in paths:
+            if os.path.exists(os.path.join(path, oc_binary)):
+                oc_binary = os.path.join(path, oc_binary)
+                break
+
+    return oc_binary
+
+
+def main():
+    """Module that executes commands on a remote OpenShift cluster"""
+
+    module = AnsibleModule(
+        argument_spec=dict(
+            namespace=dict(type="str", required=True),
+            config_file=dict(type="str", required=True),
+            cmd=dict(type="str", required=True),
+            extra_args=dict(type="list", default=[]),
+        ),
+    )
+
+    cmd = [
+        locate_oc_binary(),
+        '--config', module.params["config_file"],
+        '-n', module.params["namespace"],
+    ] + shlex.split(module.params["cmd"])
+
+    failed = True
+    try:
+        cmd_result = subprocess.check_output(list(cmd), stderr=subprocess.STDOUT)
+        failed = False
+    except subprocess.CalledProcessError as exc:
+        cmd_result = '[rc {}] {}\n{}'.format(exc.returncode, ' '.join(exc.cmd), exc.output)
+    except OSError as exc:
+        # we get this when 'oc' is not there
+        cmd_result = str(exc)
+
+    module.exit_json(
+        changed=False,
+        failed=failed,
+        result=cmd_result,
+    )
+
+
+if __name__ == '__main__':
+    main()
--- a/roles/openshift_health_checker/openshift_checks/__init__.py
+++ b/roles/openshift_health_checker/openshift_checks/__init__.py
@@ -66,16 +66,26 @@ class OpenShiftCheck(object):
 LOADER_EXCLUDES = (
     "__init__.py",
     "mixins.py",
+    "logging.py",
 )
 
 
-def load_checks():
+def load_checks(path=None, subpkg=""):
     """Dynamically import all check modules for the side effect of registering checks."""
-    return [
-        import_module(__package__ + "." + name[:-3])
-        for name in os.listdir(os.path.dirname(__file__))
-        if name.endswith(".py") and name not in LOADER_EXCLUDES
-    ]
+    if path is None:
+        path = os.path.dirname(__file__)
+
+    modules = []
+
+    for name in os.listdir(path):
+        if os.path.isdir(os.path.join(path, name)):
+            modules = modules + load_checks(os.path.join(path, name), subpkg + "." + name)
+            continue
+
+        if name.endswith(".py") and name not in LOADER_EXCLUDES:
+            modules.append(import_module(__package__ + subpkg + "." + name[:-3]))
+
+    return modules
 
 
 def get_var(task_vars, *keys, **kwargs):
--- a/roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
@@ -1,0 +1,217 @@
+"""
+Module for performing checks on an Elasticsearch logging deployment
+"""
+
+import json
+import re
+
+from openshift_checks import get_var
+from openshift_checks.logging.logging import LoggingCheck
+
+
+class Elasticsearch(LoggingCheck):
+    """Module that checks an integrated logging Elasticsearch deployment"""
+
+    name = "elasticsearch"
+    tags = ["health", "logging"]
+
+    logging_namespace = None
+
+    def run(self, tmp, task_vars):
+        """Check various things and gather errors. Returns: result as hash"""
+
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
+        es_pods, error = super(Elasticsearch, self).get_pods_for_component(
+            self.execute_module,
+            self.logging_namespace,
+            "es",
+            task_vars,
+        )
+        if error:
+            return {"failed": True, "changed": False, "msg": error}
+        check_error = self.check_elasticsearch(es_pods, task_vars)
+
+        if check_error:
+            msg = ("The following Elasticsearch deployment issue was found:"
+                   "\n-------\n"
+                   "{}".format(check_error))
+            return {"failed": True, "changed": False, "msg": msg}
+
+        # TODO(lmeyer): run it all again for the ops cluster
+        return {"failed": False, "changed": False, "msg": 'No problems found with Elasticsearch deployment.'}
+
+    def _not_running_elasticsearch_pods(self, es_pods):
+        """Returns: list of running pods, list of errors about non-running pods"""
+        not_running = super(Elasticsearch, self).not_running_pods(es_pods)
+        if not_running:
+            return not_running, [(
+                'The following Elasticsearch pods are not running:\n'
+                '{pods}'
+                'These pods will not aggregate logs from their nodes.'
+            ).format(pods=''.join(
+                "  {} ({})\n".format(pod['metadata']['name'], pod['spec'].get('host', 'None'))
+                for pod in not_running
+            ))]
+        return not_running, []
+
+    def check_elasticsearch(self, es_pods, task_vars):
+        """Various checks for elasticsearch. Returns: error string"""
+        not_running_pods, error_msgs = self._not_running_elasticsearch_pods(es_pods)
+        running_pods = [pod for pod in es_pods if pod not in not_running_pods]
+        pods_by_name = {
+            pod['metadata']['name']: pod for pod in running_pods
+            # Filter out pods that are not members of a DC
+            if pod['metadata'].get('labels', {}).get('deploymentconfig')
+        }
+        if not pods_by_name:
+            return 'No logging Elasticsearch pods were found. Is logging deployed?'
+        error_msgs += self._check_elasticsearch_masters(pods_by_name, task_vars)
+        error_msgs += self._check_elasticsearch_node_list(pods_by_name, task_vars)
+        error_msgs += self._check_es_cluster_health(pods_by_name, task_vars)
+        error_msgs += self._check_elasticsearch_diskspace(pods_by_name, task_vars)
+        return '\n'.join(error_msgs)
+
+    @staticmethod
+    def _build_es_curl_cmd(pod_name, url):
+        base = "exec {name} -- curl -s --cert {base}cert --key {base}key --cacert {base}ca -XGET '{url}'"
+        return base.format(base="/etc/elasticsearch/secret/admin-", name=pod_name, url=url)
+
+    def _check_elasticsearch_masters(self, pods_by_name, task_vars):
+        """Check that Elasticsearch masters are sane. Returns: list of error strings"""
+        es_master_names = set()
+        error_msgs = []
+        for pod_name in pods_by_name.keys():
+            # Compare what each ES node reports as master and compare for split brain
+            get_master_cmd = self._build_es_curl_cmd(pod_name, "https://localhost:9200/_cat/master")
+            master_name_str = self._exec_oc(get_master_cmd, [], task_vars)
+            master_names = (master_name_str or '').split(' ')
+            if len(master_names) > 1:
+                es_master_names.add(master_names[1])
+            else:
+                error_msgs.append(
+                    'No master? Elasticsearch {pod} returned bad string when asked master name:\n'
+                    '  {response}'.format(pod=pod_name, response=master_name_str)
+                )
+
+        if not es_master_names:
+            error_msgs.append('No logging Elasticsearch masters were found. Is logging deployed?')
+            return '\n'.join(error_msgs)
+
+        if len(es_master_names) > 1:
+            error_msgs.append(
+                'Found multiple Elasticsearch masters according to the pods:\n'
+                '{master_list}\n'
+                'This implies that the masters have "split brain" and are not correctly\n'
+                'replicating data for the logging cluster. Log loss is likely to occur.'
+                .format(master_list='\n'.join('  ' + master for master in es_master_names))
+            )
+
+        return error_msgs
+
+    def _check_elasticsearch_node_list(self, pods_by_name, task_vars):
+        """Check that reported ES masters are accounted for by pods. Returns: list of error strings"""
+
+        if not pods_by_name:
+            return ['No logging Elasticsearch masters were found. Is logging deployed?']
+
+        # get ES cluster nodes
+        node_cmd = self._build_es_curl_cmd(list(pods_by_name.keys())[0], 'https://localhost:9200/_nodes')
+        cluster_node_data = self._exec_oc(node_cmd, [], task_vars)
+        try:
+            cluster_nodes = json.loads(cluster_node_data)['nodes']
+        except (ValueError, KeyError):
+            return [
+                'Failed to query Elasticsearch for the list of ES nodes. The output was:\n' +
+                cluster_node_data
+            ]
+
+        # Try to match all ES-reported node hosts to known pods.
+        error_msgs = []
+        for node in cluster_nodes.values():
+            # Note that with 1.4/3.4 the pod IP may be used as the master name
+            if not any(node['host'] in (pod_name, pod['status'].get('podIP'))
+                       for pod_name, pod in pods_by_name.items()):
+                error_msgs.append(
+                    'The Elasticsearch cluster reports a member node "{node}"\n'
+                    'that does not correspond to any known ES pod.'.format(node=node['host'])
+                )
+
+        return error_msgs
+
+    def _check_es_cluster_health(self, pods_by_name, task_vars):
+        """Exec into the elasticsearch pods and check the cluster health. Returns: list of errors"""
+        error_msgs = []
+        for pod_name in pods_by_name.keys():
+            cluster_health_cmd = self._build_es_curl_cmd(pod_name, 'https://localhost:9200/_cluster/health?pretty=true')
+            cluster_health_data = self._exec_oc(cluster_health_cmd, [], task_vars)
+            try:
+                health_res = json.loads(cluster_health_data)
+                if not health_res or not health_res.get('status'):
+                    raise ValueError()
+            except ValueError:
+                error_msgs.append(
+                    'Could not retrieve cluster health status from logging ES pod "{pod}".\n'
+                    'Response was:\n{output}'.format(pod=pod_name, output=cluster_health_data)
+                )
+                continue
+
+            if health_res['status'] not in ['green', 'yellow']:
+                error_msgs.append(
+                    'Elasticsearch cluster health status is RED according to pod "{}"'.format(pod_name)
+                )
+
+        return error_msgs
+
+    def _check_elasticsearch_diskspace(self, pods_by_name, task_vars):
+        """
+        Exec into an ES pod and query the diskspace on the persistent volume.
+        Returns: list of errors
+        """
+        error_msgs = []
+        for pod_name in pods_by_name.keys():
+            df_cmd = 'exec {} -- df --output=ipcent,pcent /elasticsearch/persistent'.format(pod_name)
+            disk_output = self._exec_oc(df_cmd, [], task_vars)
+            lines = disk_output.splitlines()
+            # expecting one header looking like 'IUse% Use%' and one body line
+            body_re = r'\s*(\d+)%?\s+(\d+)%?\s*$'
+            if len(lines) != 2 or len(lines[0].split()) != 2 or not re.match(body_re, lines[1]):
+                error_msgs.append(
+                    'Could not retrieve storage usage from logging ES pod "{pod}".\n'
+                    'Response to `df` command was:\n{output}'.format(pod=pod_name, output=disk_output)
+                )
+                continue
+            inode_pct, disk_pct = re.match(body_re, lines[1]).groups()
+
+            inode_pct_thresh = get_var(task_vars, 'openshift_check_efk_es_inode_pct', default='90')
+            if int(inode_pct) >= int(inode_pct_thresh):
+                error_msgs.append(
+                    'Inode percent usage on the storage volume for logging ES pod "{pod}"\n'
+                    '  is {pct}, greater than threshold {limit}.\n'
+                    '  Note: threshold can be specified in inventory with {param}'.format(
+                        pod=pod_name,
+                        pct=str(inode_pct),
+                        limit=str(inode_pct_thresh),
+                        param='openshift_check_efk_es_inode_pct',
+                    ))
+            disk_pct_thresh = get_var(task_vars, 'openshift_check_efk_es_storage_pct', default='80')
+            if int(disk_pct) >= int(disk_pct_thresh):
+                error_msgs.append(
+                    'Disk percent usage on the storage volume for logging ES pod "{pod}"\n'
+                    '  is {pct}, greater than threshold {limit}.\n'
+                    '  Note: threshold can be specified in inventory with {param}'.format(
+                        pod=pod_name,
+                        pct=str(disk_pct),
+                        limit=str(disk_pct_thresh),
+                        param='openshift_check_efk_es_storage_pct',
+                    ))
+
+        return error_msgs
+
+    def _exec_oc(self, cmd_str, extra_args, task_vars):
+        return super(Elasticsearch, self).exec_oc(
+            self.execute_module,
+            self.logging_namespace,
+            cmd_str,
+            extra_args,
+            task_vars,
+        )
--- a/roles/openshift_health_checker/openshift_checks/logging/fluentd.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/fluentd.py
@@ -1,0 +1,170 @@
+"""
+Module for performing checks on an Fluentd logging deployment
+"""
+
+import json
+
+from openshift_checks import get_var
+from openshift_checks.logging.logging import LoggingCheck
+
+
+class Fluentd(LoggingCheck):
+    """Module that checks an integrated logging Fluentd deployment"""
+    name = "fluentd"
+    tags = ["health", "logging"]
+
+    logging_namespace = None
+
+    def run(self, tmp, task_vars):
+        """Check various things and gather errors. Returns: result as hash"""
+
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
+        fluentd_pods, error = super(Fluentd, self).get_pods_for_component(
+            self.execute_module,
+            self.logging_namespace,
+            "fluentd",
+            task_vars,
+        )
+        if error:
+            return {"failed": True, "changed": False, "msg": error}
+        check_error = self.check_fluentd(fluentd_pods, task_vars)
+
+        if check_error:
+            msg = ("The following Fluentd deployment issue was found:"
+                   "\n-------\n"
+                   "{}".format(check_error))
+            return {"failed": True, "changed": False, "msg": msg}
+
+        # TODO(lmeyer): run it all again for the ops cluster
+        return {"failed": False, "changed": False, "msg": 'No problems found with Fluentd deployment.'}
+
+    @staticmethod
+    def _filter_fluentd_labeled_nodes(nodes_by_name, node_selector):
+        """Filter to all nodes with fluentd label. Returns dict(name: node), error string"""
+        label, value = node_selector.split('=', 1)
+        fluentd_nodes = {
+            name: node for name, node in nodes_by_name.items()
+            if node['metadata']['labels'].get(label) == value
+        }
+        if not fluentd_nodes:
+            return None, (
+                'There are no nodes with the fluentd label {label}.\n'
+                'This means no logs will be aggregated from the nodes.'
+            ).format(label=node_selector)
+        return fluentd_nodes, None
+
+    @staticmethod
+    def _check_node_labeling(nodes_by_name, fluentd_nodes, node_selector, task_vars):
+        """Note if nodes are not labeled as expected. Returns: error string"""
+        intended_nodes = get_var(task_vars, 'openshift_logging_fluentd_hosts', default=['--all'])
+        if not intended_nodes or '--all' in intended_nodes:
+            intended_nodes = nodes_by_name.keys()
+        nodes_missing_labels = set(intended_nodes) - set(fluentd_nodes.keys())
+        if nodes_missing_labels:
+            return (
+                'The following nodes are supposed to be labeled with {label} but are not:\n'
+                '  {nodes}\n'
+                'Fluentd will not aggregate logs from these nodes.'
+            ).format(label=node_selector, nodes=', '.join(nodes_missing_labels))
+        return None
+
+    @staticmethod
+    def _check_nodes_have_fluentd(pods, fluentd_nodes):
+        """Make sure fluentd is on all the labeled nodes. Returns: error string"""
+        unmatched_nodes = fluentd_nodes.copy()
+        node_names_by_label = {
+            node['metadata']['labels']['kubernetes.io/hostname']: name
+            for name, node in fluentd_nodes.items()
+        }
+        node_names_by_internal_ip = {
+            address['address']: name
+            for name, node in fluentd_nodes.items()
+            for address in node['status']['addresses']
+            if address['type'] == "InternalIP"
+        }
+        for pod in pods:
+            for name in [
+                    pod['spec']['nodeName'],
+                    node_names_by_internal_ip.get(pod['spec']['nodeName']),
+                    node_names_by_label.get(pod.get('spec', {}).get('host')),
+            ]:
+                unmatched_nodes.pop(name, None)
+        if unmatched_nodes:
+            return (
+                'The following nodes are supposed to have a Fluentd pod but do not:\n'
+                '{nodes}'
+                'These nodes will not have their logs aggregated.'
+            ).format(nodes=''.join(
+                "  {}\n".format(name)
+                for name in unmatched_nodes.keys()
+            ))
+        return None
+
+    def _check_fluentd_pods_running(self, pods):
+        """Make sure all fluentd pods are running. Returns: error string"""
+        not_running = super(Fluentd, self).not_running_pods(pods)
+        if not_running:
+            return (
+                'The following Fluentd pods are supposed to be running but are not:\n'
+                '{pods}'
+                'These pods will not aggregate logs from their nodes.'
+            ).format(pods=''.join(
+                "  {} ({})\n".format(pod['metadata']['name'], pod['spec'].get('host', 'None'))
+                for pod in not_running
+            ))
+        return None
+
+    def check_fluentd(self, pods, task_vars):
+        """Verify fluentd is running everywhere. Returns: error string"""
+
+        node_selector = get_var(task_vars, 'openshift_logging_fluentd_nodeselector',
+                                default='logging-infra-fluentd=true')
+
+        nodes_by_name, error = self.get_nodes_by_name(task_vars)
+
+        if error:
+            return error
+        fluentd_nodes, error = self._filter_fluentd_labeled_nodes(nodes_by_name, node_selector)
+        if error:
+            return error
+
+        error_msgs = []
+        error = self._check_node_labeling(nodes_by_name, fluentd_nodes, node_selector, task_vars)
+        if error:
+            error_msgs.append(error)
+        error = self._check_nodes_have_fluentd(pods, fluentd_nodes)
+        if error:
+            error_msgs.append(error)
+        error = self._check_fluentd_pods_running(pods)
+        if error:
+            error_msgs.append(error)
+
+        # Make sure there are no extra fluentd pods
+        if len(pods) > len(fluentd_nodes):
+            error_msgs.append(
+                'There are more Fluentd pods running than nodes labeled.\n'
+                'This may not cause problems with logging but it likely indicates something wrong.'
+            )
+
+        return '\n'.join(error_msgs)
+
+    def get_nodes_by_name(self, task_vars):
+        """Retrieve all the node definitions. Returns: dict(name: node), error string"""
+        nodes_json = self._exec_oc("get nodes -o json", [], task_vars)
+        try:
+            nodes = json.loads(nodes_json)
+        except ValueError:  # no valid json - should not happen
+            return None, "Could not obtain a list of nodes to validate fluentd. Output from oc get:\n" + nodes_json
+        if not nodes or not nodes.get('items'):  # also should not happen
+            return None, "No nodes appear to be defined according to the API."
+        return {
+            node['metadata']['name']: node
+            for node in nodes['items']
+        }, None
+
+    def _exec_oc(self, cmd_str, extra_args, task_vars):
+        return super(Fluentd, self).exec_oc(self.execute_module,
+                                            self.logging_namespace,
+                                            cmd_str,
+                                            extra_args,
+                                            task_vars)
--- a/roles/openshift_health_checker/openshift_checks/logging/kibana.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/kibana.py
@@ -1,0 +1,229 @@
+"""
+Module for performing checks on a Kibana logging deployment
+"""
+
+import json
+import ssl
+
+try:
+    from urllib2 import HTTPError, URLError
+    import urllib2
+except ImportError:
+    from urllib.error import HTTPError, URLError
+    import urllib.request as urllib2
+
+from openshift_checks import get_var
+from openshift_checks.logging.logging import LoggingCheck
+
+
+class Kibana(LoggingCheck):
+    """Module that checks an integrated logging Kibana deployment"""
+
+    name = "kibana"
+    tags = ["health", "logging"]
+
+    logging_namespace = None
+
+    def run(self, tmp, task_vars):
+        """Check various things and gather errors. Returns: result as hash"""
+
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
+        kibana_pods, error = super(Kibana, self).get_pods_for_component(
+            self.execute_module,
+            self.logging_namespace,
+            "kibana",
+            task_vars,
+        )
+        if error:
+            return {"failed": True, "changed": False, "msg": error}
+        check_error = self.check_kibana(kibana_pods)
+
+        if not check_error:
+            check_error = self._check_kibana_route(task_vars)
+
+        if check_error:
+            msg = ("The following Kibana deployment issue was found:"
+                   "\n-------\n"
+                   "{}".format(check_error))
+            return {"failed": True, "changed": False, "msg": msg}
+
+        # TODO(lmeyer): run it all again for the ops cluster
+        return {"failed": False, "changed": False, "msg": 'No problems found with Kibana deployment.'}
+
+    def _verify_url_internal(self, url, task_vars):
+        """
+        Try to reach a URL from the host.
+        Returns: success (bool), reason (for failure)
+        """
+        args = dict(
+            url=url,
+            follow_redirects='none',
+            validate_certs='no',  # likely to be signed with internal CA
+            # TODO(lmeyer): give users option to validate certs
+            status_code=302,
+        )
+        result = self.execute_module('uri', args, task_vars)
+        if result.get('failed'):
+            return result['msg']
+        return None
+
+    @staticmethod
+    def _verify_url_external(url):
+        """
+        Try to reach a URL from ansible control host.
+        Returns: success (bool), reason (for failure)
+        """
+        # This actually checks from the ansible control host, which may or may not
+        # really be "external" to the cluster.
+
+        # Disable SSL cert validation to work around internally signed certs
+        ctx = ssl.create_default_context()
+        ctx.check_hostname = False  # or setting CERT_NONE is refused
+        ctx.verify_mode = ssl.CERT_NONE
+
+        # Verify that the url is returning a valid response
+        try:
+            # We only care if the url connects and responds
+            return_code = urllib2.urlopen(url, context=ctx).getcode()
+        except HTTPError as httperr:
+            return httperr.reason
+        except URLError as urlerr:
+            return str(urlerr)
+
+        # there appears to be no way to prevent urlopen from following redirects
+        if return_code != 200:
+            return 'Expected success (200) but got return code {}'.format(int(return_code))
+
+        return None
+
+    def check_kibana(self, pods):
+        """Check to see if Kibana is up and working. Returns: error string."""
+
+        if not pods:
+            return "There are no Kibana pods deployed, so no access to the logging UI."
+
+        not_running = self.not_running_pods(pods)
+        if len(not_running) == len(pods):
+            return "No Kibana pod is in a running state, so there is no access to the logging UI."
+        elif not_running:
+            return (
+                "The following Kibana pods are not currently in a running state:\n"
+                "{pods}"
+                "However at least one is, so service may not be impacted."
+            ).format(pods="".join("  " + pod['metadata']['name'] + "\n" for pod in not_running))
+
+        return None
+
+    def _get_kibana_url(self, task_vars):
+        """
+        Get kibana route or report error.
+        Returns: url (or empty), reason for failure
+        """
+
+        # Get logging url
+        get_route = self._exec_oc("get route logging-kibana -o json", [], task_vars)
+        if not get_route:
+            return None, 'no_route_exists'
+
+        route = json.loads(get_route)
+
+        # check that the route has been accepted by a router
+        ingress = route["status"]["ingress"]
+        # ingress can be null if there is no router, or empty if not routed
+        if not ingress or not ingress[0]:
+            return None, 'route_not_accepted'
+
+        host = route.get("spec", {}).get("host")
+        if not host:
+            return None, 'route_missing_host'
+
+        return 'https://{}/'.format(host), None
+
+    def _check_kibana_route(self, task_vars):
+        """
+        Check to see if kibana route is up and working.
+        Returns: error string
+        """
+        known_errors = dict(
+            no_route_exists=(
+                'No route is defined for Kibana in the logging namespace,\n'
+                'so the logging stack is not accessible. Is logging deployed?\n'
+                'Did something remove the logging-kibana route?'
+            ),
+            route_not_accepted=(
+                'The logging-kibana route is not being routed by any router.\n'
+                'Is the router deployed and working?'
+            ),
+            route_missing_host=(
+                'The logging-kibana route has no hostname defined,\n'
+                'which should never happen. Did something alter its definition?'
+            ),
+        )
+
+        kibana_url, error = self._get_kibana_url(task_vars)
+        if not kibana_url:
+            return known_errors.get(error, error)
+
+        # first, check that kibana is reachable from the master.
+        error = self._verify_url_internal(kibana_url, task_vars)
+        if error:
+            if 'urlopen error [Errno 111] Connection refused' in error:
+                error = (
+                    'Failed to connect from this master to Kibana URL {url}\n'
+                    'Is kibana running, and is at least one router routing to it?'
+                ).format(url=kibana_url)
+            elif 'urlopen error [Errno -2] Name or service not known' in error:
+                error = (
+                    'Failed to connect from this master to Kibana URL {url}\n'
+                    'because the hostname does not resolve.\n'
+                    'Is DNS configured for the Kibana hostname?'
+                ).format(url=kibana_url)
+            elif 'Status code was not' in error:
+                error = (
+                    'A request from this master to the Kibana URL {url}\n'
+                    'did not return the correct status code (302).\n'
+                    'This could mean that Kibana is malfunctioning, the hostname is\n'
+                    'resolving incorrectly, or other network issues. The output was:\n'
+                    '  {error}'
+                ).format(url=kibana_url, error=error)
+            return 'Error validating the logging Kibana route:\n' + error
+
+        # in production we would like the kibana route to work from outside the
+        # cluster too; but that may not be the case, so allow disabling just this part.
+        if not get_var(task_vars, "openshift_check_efk_kibana_external", default=True):
+            return None
+        error = self._verify_url_external(kibana_url)
+        if error:
+            if 'urlopen error [Errno 111] Connection refused' in error:
+                error = (
+                    'Failed to connect from the Ansible control host to Kibana URL {url}\n'
+                    'Is the router for the Kibana hostname exposed externally?'
+                ).format(url=kibana_url)
+            elif 'urlopen error [Errno -2] Name or service not known' in error:
+                error = (
+                    'Failed to resolve the Kibana hostname in {url}\n'
+                    'from the Ansible control host.\n'
+                    'Is DNS configured to resolve this Kibana hostname externally?'
+                ).format(url=kibana_url)
+            elif 'Expected success (200)' in error:
+                error = (
+                    'A request to Kibana at {url}\n'
+                    'returned the wrong error code:\n'
+                    '  {error}\n'
+                    'This could mean that Kibana is malfunctioning, the hostname is\n'
+                    'resolving incorrectly, or other network issues.'
+                ).format(url=kibana_url, error=error)
+            error = (
+                'Error validating the logging Kibana route:\n{error}\n'
+                'To disable external Kibana route validation, set in your inventory:\n'
+                '  openshift_check_efk_kibana_external=False'
+            ).format(error=error)
+            return error
+        return None
+
+    def _exec_oc(self, cmd_str, extra_args, task_vars):
+        return super(Kibana, self).exec_oc(self.execute_module,
+                                           self.logging_namespace,
+                                           cmd_str,
+                                           extra_args,
+                                           task_vars)
--- a/roles/openshift_health_checker/openshift_checks/logging/logging.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/logging.py
@@ -1,0 +1,96 @@
+"""
+Util functions for performing checks on an Elasticsearch, Fluentd, and Kibana stack
+"""
+
+import json
+import os
+
+from openshift_checks import OpenShiftCheck, OpenShiftCheckException, get_var
+
+
+class LoggingCheck(OpenShiftCheck):
+    """Base class for logging component checks"""
+
+    name = "logging"
+
+    @classmethod
+    def is_active(cls, task_vars):
+        return super(LoggingCheck, cls).is_active(task_vars) and cls.is_first_master(task_vars)
+
+    @staticmethod
+    def is_first_master(task_vars):
+        """Run only on first master and only when logging is configured. Returns: bool"""
+        logging_deployed = get_var(task_vars, "openshift_hosted_logging_deploy", default=True)
+        # Note: It would be nice to use membership in oo_first_master group, however for now it
+        # seems best to avoid requiring that setup and just check this is the first master.
+        hostname = get_var(task_vars, "ansible_ssh_host") or [None]
+        masters = get_var(task_vars, "groups", "masters", default=None) or [None]
+        return logging_deployed and masters[0] == hostname
+
+    def run(self, tmp, task_vars):
+        pass
+
+    def get_pods_for_component(self, execute_module, namespace, logging_component, task_vars):
+        """Get all pods for a given component. Returns: list of pods for component, error string"""
+        pod_output = self.exec_oc(
+            execute_module,
+            namespace,
+            "get pods -l component={} -o json".format(logging_component),
+            [],
+            task_vars
+        )
+        try:
+            pods = json.loads(pod_output)
+            if not pods or not pods.get('items'):
+                raise ValueError()
+        except ValueError:
+            # successful run but non-parsing data generally means there were no pods in the namespace
+            return None, 'There are no pods in the {} namespace. Is logging deployed?'.format(namespace)
+
+        return pods['items'], None
+
+    @staticmethod
+    def not_running_pods(pods):
+        """Returns: list of pods not in a ready and running state"""
+        return [
+            pod for pod in pods
+            if any(
+                container['ready'] is False
+                for container in pod['status']['containerStatuses']
+            ) or not any(
+                condition['type'] == 'Ready' and condition['status'] == 'True'
+                for condition in pod['status']['conditions']
+            )
+        ]
+
+    @staticmethod
+    def exec_oc(execute_module=None, namespace="logging", cmd_str="", extra_args=None, task_vars=None):
+        """
+        Execute an 'oc' command in the remote host.
+        Returns: output of command and namespace,
+        or raises OpenShiftCheckException on error
+        """
+        config_base = get_var(task_vars, "openshift", "common", "config_base")
+        args = {
+            "namespace": namespace,
+            "config_file": os.path.join(config_base, "master", "admin.kubeconfig"),
+            "cmd": cmd_str,
+            "extra_args": list(extra_args) if extra_args else [],
+        }
+
+        result = execute_module("ocutil", args, task_vars)
+        if result.get("failed"):
+            msg = (
+                'Unexpected error using `oc` to validate the logging stack components.\n'
+                'Error executing `oc {cmd}`:\n'
+                '{error}'
+            ).format(cmd=args['cmd'], error=result['result'])
+
+            if result['result'] == '[Errno 2] No such file or directory':
+                msg = (
+                    "This host is supposed to be a master but does not have the `oc` command where expected.\n"
+                    "Has an installation been run on this host yet?"
+                )
+            raise OpenShiftCheckException(msg)
+
+        return result.get("result", "")
--- a/roles/openshift_health_checker/test/curator_test.py
+++ b/roles/openshift_health_checker/test/curator_test.py
@@ -1,0 +1,68 @@
+import pytest
+
+from openshift_checks.logging.curator import Curator
+
+
+def canned_curator(exec_oc=None):
+    """Create a Curator check object with canned exec_oc method"""
+    check = Curator("dummy")  # fails if a module is actually invoked
+    if exec_oc:
+        check._exec_oc = exec_oc
+    return check
+
+
+def assert_error(error, expect_error):
+    if expect_error:
+        assert error
+        assert expect_error in error
+    else:
+        assert not error
+
+
+plain_curator_pod = {
+    "metadata": {
+        "labels": {"component": "curator", "deploymentconfig": "logging-curator"},
+        "name": "logging-curator-1",
+    },
+    "status": {
+        "containerStatuses": [{"ready": True}],
+        "conditions": [{"status": "True", "type": "Ready"}],
+        "podIP": "10.10.10.10",
+    }
+}
+
+not_running_curator_pod = {
+    "metadata": {
+        "labels": {"component": "curator", "deploymentconfig": "logging-curator"},
+        "name": "logging-curator-2",
+    },
+    "status": {
+        "containerStatuses": [{"ready": False}],
+        "conditions": [{"status": "False", "type": "Ready"}],
+        "podIP": "10.10.10.10",
+    }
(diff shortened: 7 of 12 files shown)


juanvallejo:
@portante since this was largely ported from  https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py, I was wondering if you could give this a pass?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def run(self, _, task_vars):
+        self.get_pods(task_vars)
+
+        status = dict()
+        status['elasticsearch'] = self.check_elasticsearch(task_vars)
+        status['fluentd'] = self.check_fluentd(task_vars)
+        status['kibana'] = self.check_kibana(task_vars)
+
+        return {"changed": True, "msg": status}
```
juanvallejo: @brenton was not sure what would constitute a "failure" for this check. In the case of no pods, routes, etc. present for elasticsearch, kibana, or fluentd, I receive an object that looks like: 

```json
{
    "elasticsearch": {
        "pods": {},
        "single_master": null
    },
    "fluentd": {
        "node_mismatch": 0,
        "number_expected_pods": 0,
        "number_pods": 0,
        "running": 1
    },
    "kibana": {}
}
```

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,270 @@
...
+    """"
+        This is a check for the entire EFK stack shipped with OCP
+        https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py
+    """
+
+    name = "efk_stack_check"
+    tags = ["preflight", "elasticsearch", "efkstack"]
+
+    namespace = "logging"
+    config_file = "/etc/origin/master/admin.kubeconfig"
```
rhcarvalho: I think the config might be coming from different paths? Not sure.

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,270 @@
...
+
+    args = None
+    es_pods = []
+    fluentd_pods = []
+
+    def __init__(self, module_executor):
+        super(EFKStackCheck, self).__init__(module_executor)
+
+        es_cert = '/etc/elasticsearch/secret/admin-'
+        self.es_curl = "curl -s --cert {}cert --key {}key --cacert {}ca -XGET ".format(es_cert, es_cert, es_cert)
```
rhcarvalho: Hmm, shouldn't the `ca` suffix be part of the `es_cert` above?!

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
rhcarvalho: Does this check change the host(s)?

rhcarvalho: > what would constitute a "failure" for this check. In the case of no pods, routes, etc. present for elasticsearch, kibana, or fluentd, ...

That sounds like an important question for us to answer :-)

My guess would be that anything that could indicate that the logging stack is non-functional (might be a long list of possibilities)? Ideally we should include some explanation about what is wrong and how to fix it, just like `oadm diagnostics` does.

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def run(self, _, task_vars):
```
rhcarvalho: As in the other PR, I'd suggest the explicit:

```python
```

I am still to see the case where `tmp` is actually used... we need to dig again into action plugins to find an example of that.

```
$ grep -nr tmp ansible/plugins/action
./async.py:32:    def run(self, tmp=None, task_vars=None):
./async.py:37:        result = super(ActionModule, self).run(tmp, task_vars)
./async.py:45:        if not tmp:
./async.py:46:            tmp = self._make_tmp_path(remote_user)
./async.py:47:            self._cleanup_remote_tmp=True
...
./unarchive.py:33:    def run(self, tmp=None, task_vars=None):
./unarchive.py:38:        result = super(ActionModule, self).run(tmp, task_vars)
./unarchive.py:62:        if not tmp:
./unarchive.py:63:            tmp = self._make_tmp_path(remote_user)
./unarchive.py:64:            self._cleanup_remote_tmp = True
...
```

There are several examples where `tmp` is used to determine where to create temp files.

roles/openshift_health_checker/library/ocutil.py
```diff
@@ -0,0 +1,172 @@
...
+# pylint: disable=bare-except
+def cleanup_file(inc_file):
+    """ clean up """
+    try:
+        os.unlink(inc_file)
+    except:
+        pass
+
+
+class OCUtil(object):
```
portante: If we are going to move this module from [`openshift-tools`](https://github.com/openshift/openshift-tools/blob/630252b138a658bec2ed9305e86edced1688a956/openshift_tools/monitoring/ocutil.py), perhaps we should consider updating the license date, and proposing the monitoring code use it from here?

We don't want two copies, but we need to settle the location where it should live.

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,270 @@
+#!/usr/bin/env python
+'''
+  Send Cluster EFK checks to Zagg
+'''
+# pylint: disable=missing-docstring
+# vim: expandtab:tabstop=4:shiftwidth=4
+#
+#   Copyright 2015 Red Hat Inc.
```
portante: If this is a new file, perhaps an updated copyright date? :)

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,270 @@
...
+import urllib2
+import json
+
+from openshift_checks import OpenShiftCheck
+
+
+class EFKStackCheck(OpenShiftCheck):
+    """"
+        This is a check for the entire EFK stack shipped with OCP
+        https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py
```
portante: And if it is a complete replacement, perhaps we should have that script just invoke this one?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
portante: This seems dangerous to hard-code.  Perhaps this can be derived from the environment?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
portante: Perhaps, but this was just taken from the previous code.

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,270 @@
...
+# pylint: disable=bare-except
+import ssl
+import urllib2
+import json
+
+from openshift_checks import OpenShiftCheck
+
+
+class EFKStackCheck(OpenShiftCheck):
+    """"
```
portante: You appear to have four double quotes here ...

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,270 @@
...
+
+    name = "efk_stack_check"
+    tags = ["preflight", "elasticsearch", "efkstack"]
+
+    namespace = "logging"
+    config_file = "/etc/origin/master/admin.kubeconfig"
+
+    args = None
+    es_pods = []
+    fluentd_pods = []
```
portante: Why did you make the above three variable class level variables?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,270 @@
...
+                if pod['status']['podIP'] == es_master_name:
+                    es_master_name = pod['metadata']['name']
+            except:
+                continue
+
+        # get cluster nodes
+        curl_cmd = "{} 'https://localhost:9200/_nodes'".format(self.es_curl)
+        node_cmd = "exec -ti {} -- {}".format(es_master_name, curl_cmd)
+
+        if not es_master_name:
```
portante: Why was this check added here?

It looks as if the for loop on line 93 is supposed to fill this field, since it is needed by line 116.  Can we move it before line 116?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,270 @@
...
+
+        status = dict()
+        status['elasticsearch'] = self.check_elasticsearch(task_vars)
+        status['fluentd'] = self.check_fluentd(task_vars)
+        status['kibana'] = self.check_kibana(task_vars)
+
+        return {"changed": True, "msg": status}
+
+    def oc(self, method, args, task_vars):
+        args = {
```
portante: Why do you overwrite the `args` variable with a dictionary here?

Why not call the parameter `method_args` and just assign it to `args`?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,270 @@
...
+            "namespace": self.namespace,
+            "config_file": self.config_file,
+            "method": method,
+            "method_args": args,
+        }
+        result = self.module_executor("ocutil", args, task_vars)
```
portante: I don't know what this method does, but I assume that even though you are using a lowercase `ocutil`, it will pull in the `OCUtil` module that you are porting over?

roles/openshift_health_checker/library/ocutil.py
juanvallejo: @rhcarvalho 
> Where are we importing this code from?
 
It came from: https://github.com/openshift/openshift-tools/blob/prod/openshift_tools/monitoring/ocutil.py

@portante 
> If we are going to move this module from openshift-tools, perhaps we should consider updating the license date, and proposing the monitoring code use it from here?
>
> We don't want two copies, but we need to settle the location where it should live.

If it is okay / feasible for the monitoring code to use it from here, that would work. I only copied it on this PR with the intent on making it runnable as an ansible module, without having to write a wrapper for `oc` from scratch. But if we can re-use that file, that would definitely be ideal. @brenton WDYT?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+class EFKStackCheck(OpenShiftCheck):
+    """"
+        This is a check for the entire EFK stack shipped with OCP
+        https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py
+    """
+
+    name = "efk_stack_check"
+    tags = ["preflight", "elasticsearch", "efkstack"]
```
juanvallejo: @rhcarvalho 
> Is this really preflight or does it assume a cluster / logging stack deployed?

Thanks, will remove `preflight` tag. It does assume that a logging stack has already been deployed

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
juanvallejo: @rhcarvalho 
> Does this check change the host(s)?

The only change that is made (as far as I can tell) is when we copy the `kubeconfig` file in `OCUtil.py`. However, that same method also deletes the copy on program exit, so the host is not "changed" by the time this check terminates.

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
juanvallejo: Ah, still haven't fixed auto-quote completion to not intervene when doing triple-quotes :) will fix

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
juanvallejo: @portante 
> I don't know what this method does, but I assume that even though you are using a lowercase ocutil, it will pull in the OCUtil module that you are porting over?

Yeah, this is a method we use to execute local modules (as well as ansible modules). It is not case sensitive

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
juanvallejo: I replicated the same behavior of the original script: https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py#L43-L45, however if it is better to declare them locally in `def run(...` and pass them as arguments to the other methods, I can do that instead

roles/openshift_health_checker/library/ocutil.py
rhcarvalho: It looks like `openshift_tools` is just a set of Python scripts unrelated to Ansible...
Moving the canonical source of the file here doesn't make sense, they shouldn't try to reuse an "Ansible module" (similar but != "Python module") from this repo and bring in all the versioning and bookkeeping headache.

Makes even less sense to go that far to grab code instead of:
1. Keep it simple and use the stdlib unless it pays off to bring in a dependency
2. Use code from the openshift-ansible repo, see: https://github.com/openshift/openshift-ansible/blob/b47708f36d0e45688d4444ca22f6a62d0c5de2a6/roles/lib_openshift/src/lib/base.py#L251

Option no. 1 is highly preferable if all you need is to make simple `oc` calls:

https://docs.python.org/2/library/subprocess.html#using-the-subprocess-module

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,277 @@
...
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# This is not a module, but pylint thinks it is.  This is a command.
+# pylint: disable=invalid-name
+# If a check throws an exception it is failed and should alert
+# pylint: disable=bare-except
```
rhcarvalho: If you're copying these pylint disable comments from somewhere else, we need to review them.

The explanations above don't look like good arguments.

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,277 @@
...
+            return {"failed": True, "msg": "Unable to access Kibana logging url. Invalid server response."}
+
+        return {"changed": False, "msg": status}
+
+    def oc(self, method, method_args, task_vars):
+        config_base = get_var(task_vars, "openshift", "common", "config_base")
+
+        args = {
+            "namespace": self.namespace,
+            "config_file": "".join([config_base, "/master/admin.kubeconfig"]),
```
rhcarvalho: In this case we do not need `str.join`, we could simply write:

```python
config_base + "/master/admin.kubeconfig"
```

If we are to join parts, then we should at least go ahead and do not hard code the path separator...

```python
os.path.join(config_base, "master", "admin.kubeconfig")
```

(the effectiveness of this of course depends on how `config_base` itself is computed)

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
rhcarvalho: > However, that same method also deletes the copy on program exit,

Risky assumption. We cannot guarantee the code runs till the end and the file gets removed at all times. Think about exceptions being raised for example... or network error...

Regarding copying the config file, I'd see how we handle this in other places (grep for "admin.kubeconfig" and you will find some playbooks).

https://github.com/rhcarvalho/openshift-ansible/blob/f017f5afc88e65ce57751c6e6ac99cb9094eb716/playbooks/common/openshift-cluster/redeploy-certificates/registry.yml#L21-L25

Apparently we ignore changes to /tmp... not sure if that's aligned with the current best practices.

@detiber what's the best practice to have idempotent modules / use of "changed" key in module results when we need to write temp files?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
rhcarvalho: > It is not case sensitive

It is case sensitive, and the name refers to the module / file name, not the class name.
The module in `library/ocutil.py` is executed by call telling ansible to run a module named "ocutil".

The name of the class is irrelevant. There doesn't even need to be a class involved...

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
juanvallejo: @rhcarvalho 
> Risky assumption. We cannot guarantee the code runs till the end and the file gets removed at all times. Think about exceptions being raised for example... or network error...

Those are good points. An alternative solution would be to simply default `changed` to `True` every time, or avoid copying kubeconfig altogether (although this might be equally risky).

Also, the same file you linked seems to build the path to kubeconfig in the same manner that this PR currently does; by grabbing the value of `config_base`, and concatenating `/master/admin.kubeconfig` to it: https://github.com/rhcarvalho/openshift-ansible/blob/f017f5afc88e65ce57751c6e6ac99cb9094eb716/playbooks/common/openshift-cluster/redeploy-certificates/registry.yml#L15

juanvallejo:
@portante @rhcarvalho thanks for all of the feedback, review comments addressed

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
rhcarvalho: I asked for I don't know what is precisely Ansible's interpretation of "changed"...

rhcarvalho:
@juanvallejo please squash into meaningful commits as the preparation for merge.

roles/openshift_health_checker/library/ocutil.py
```diff
@@ -0,0 +1,62 @@
...
+            config_file=dict(type="str", required=True),
+            cmd=dict(type="str", required=True)
+        ),
+        supports_check_mode=True
+    )
+
+    config_file = module.params["config_file"]
+
+    cmd = " ".join(['oc --config', config_file, '-n', module.params["namespace"], module.params["cmd"]])
+    cmd_split = shlex.split(cmd)
```
rhcarvalho: Why join and then immediately split?!

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,274 @@
+#!/usr/bin/env python
```
rhcarvalho: This is not meant to be an executable file.
Please remove the shebang and make sure the execution bit is not set, it doesn't need to be.

roles/openshift_health_checker/library/ocutil.py
```diff
@@ -0,0 +1,62 @@
...
+
+def main():
+    """ Handle args and execute OCUtil method """
+    module = AnsibleModule(
+        argument_spec=dict(
+            namespace=dict(type="str", required=True),
+            config_file=dict(type="str", required=True),
+            cmd=dict(type="str", required=True)
+        ),
+        supports_check_mode=True
```
rhcarvalho: Looks like the code does not support check mode?!

roles/openshift_health_checker/library/ocutil.py
```diff
@@ -0,0 +1,62 @@
...
+#
+
+import shlex
+import subprocess
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def main():
+    """ Handle args and execute OCUtil method """
```
rhcarvalho: Outdated comment.

roles/openshift_health_checker/library/ocutil.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python
+# vim: expandtab:tabstop=4:shiftwidth=4
+"""
+  Interface to OpenShift oc command
```
rhcarvalho: Extraneous indentation

https://www.python.org/dev/peps/pep-0257/

roles/openshift_health_checker/library/ocutil.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+#
+#   Copyright 2017 Red Hat Inc.
```
rhcarvalho: Most of the source files in this repo don't have this preamble. We do have a LICENSE file in the root of the repo.
I'm not a lawyer, but I'd say unless we have reasons to go this path I'd suggest we keep it simple and remove this.

roles/openshift_health_checker/library/ocutil.py
```diff
@@ -0,0 +1,62 @@
...
+    try:
+        cmd_result = subprocess.check_output(cmd_split)
+    except subprocess.CalledProcessError as err:
+        failed = True
+        cmd_result = err
+
+    module.exit_json(
+        changed=False,
+        failed=failed,
+        ansible_facts={
```
rhcarvalho: This has some special meaning http://docs.ansible.com/ansible/common_return_values.html#ansible-facts

Why are we using it?

Won't it pollute the host vars?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,274 @@
...
+import ssl
+import urllib2
+
+from openshift_checks import OpenShiftCheck, get_var
+
+
+class EFKStackCheck(OpenShiftCheck):
+    """
+        This is a check for the entire EFK stack shipped with OCP
+        https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py
```
rhcarvalho: The indentation here is off, and our CI doesn't check that.
Dedent one level please.

Also how relevant is it to link to that script? Isn't it all here self-contained?
The comment is not clear as to what is the point of the link.

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,274 @@
...
+    """
+
+    name = "efk_stack_check"
+    tags = ["elasticsearch", "efkstack"]
+
+    namespace = "logging"
```
rhcarvalho: @juanvallejo @jcantrill is the namespace for the logging stack installation user-configurable?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,274 @@
...
+
+    es_pods = []
+    fluentd_pods = []
```
rhcarvalho: It is poor coding style to store these lists as class attributes.
Think about it for a while ;)

The code becomes tied to calling a handful of methods in the right order to mutate / read from those...
It is very similar to global variables, misses encapsulation, promotes mutation of state, makes the whole thing harder to test and so on.

Instead, I suggest we return list of pods where they need to be returned, and pass explicitly as argument a list where there needs to be a list as input.

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,274 @@
...
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+
+        # Verify that the url is returning a valid response
+        kibana_status['site_up'] = 1
+        try:
+            # We only care if the url opens
+            urllib2.urlopen(kibana_url, context=ctx)
+
+        except urllib2.HTTPError, e:
```
rhcarvalho: Use `as` instead of `,`.

That's the Python 2 / Python 3 way. Comma is legacy syntax.

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,274 @@
...
+        route = route.get("host", "")
+        kibana_url = "https://{}/".format(route)
+
+        # Disable SSL to work around self signed clusters
+        ctx = ssl.create_default_context()
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+
+        # Verify that the url is returning a valid response
+        kibana_status['site_up'] = 1
```
rhcarvalho: Why do we use 1 / 0 instead of True / False?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,274 @@
...
+        """ Execute an 'oc' command in the remote host """
+        config_base = get_var(task_vars, "openshift", "common", "config_base")
+        args = {
+            "namespace": self.namespace,
+            "config_file": os.path.join(config_base, "master", "admin.kubeconfig"),
+            "cmd": cmd_str,
+        }
+
+        result = self.module_executor("ocutil", args, task_vars)
+        return result.get("ansible_facts", {}).get("result", "")
```
rhcarvalho: Yeah... we totally don't need to wrap the result into a "ansible_facts" key, do we?!

portante:
Could we consider refactoring the original code to extract the zabbix integration from the core OpenShift controls, and just leverage it from ansible scripts here?  Seems like that would be the better way to move forward, as you already have identified what that looks like.

roles/openshift_health_checker/library/ocutil.py
juanvallejo: > This has some special meaning http://docs.ansible.com/ansible/common_return_values.html#ansible-facts
>
> Why are we using it?
>
> Won't it pollute the host vars?

Hm, thanks for the link. I will remove the nested `ansible_facts` dict.
I took this approach, as I saw that this is what some of the [core ansible modules do](https://github.com/ansible/ansible-modules-core/blob/devel/cloud/docker/docker_container.py#L1672)

rhcarvalho:
> Could we consider refactoring the original code to extract the zabbix integration from the core OpenShift controls, and just leverage it from ansible scripts here? Seems like that would be the better way to move forward, as you already have identified what that looks like.

@portante consuming an external dependency for us (Customer Success team, working on these checks) is not interesting in this particular case. The maintenance cost would be higher than the value delivered.
We do use existing tools when that makes sense, but in this case all @juanvallejo needed was simple calls to `oc`.

In any case, let's keep our eyes open for possible code reuse and avoiding reimplementing functionality.

Thank you for your review and suggestions!

roles/openshift_health_checker/library/ocutil.py
juanvallejo: @rhcarvalho 
> Why join and then immediately split?!

To ensure that the `oc` "command string" that will be executed on the remote host is also split, and not just a single item in the list that is passed to `subprocess.check_output`. Having a list where the different parts of `module.params["cmd"]` are not individual items in the list passed to `subprocess` renders an invalid / empty result.

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
juanvallejo: No, according to https://github.com/openshift/origin-aggregated-logging/tree/master/deployer#choose-a-project, the `logging` namespace is recommended, but not required. I can update this line to

```python
namespace = get_var(task_vars, "openshift_logging_namespace")
```

or something similar. WDYT?

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
juanvallejo: This was kept from when this was ported from https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py, but can update to `True / False`

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
juanvallejo: Will update, since per [your suggestion here](https://github.com/openshift/openshift-ansible/pull/3643#discussion_r106513615), we can read the `results` key directly from the module_executor result

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
juanvallejo: Thanks, will update

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,257 @@
...
+        self.es_curl = es_curl_base.format(es_cert_prefix, es_cert_prefix, es_cert_prefix)
+
+    def run(self, tmp, task_vars):
+        es_pods, fluentd_pods = self.get_pods(task_vars)
+
+        status = dict()
+        status['elasticsearch'] = self.check_elasticsearch(es_pods, task_vars)
+        status['fluentd'] = self.check_fluentd(fluentd_pods, task_vars)
+        status['kibana'] = self.check_kibana(task_vars)
+
```
jcantrill: Status check for curator looks to be missing.  Also need to consider how to handle the 'ops' components as well

roles/openshift_health_checker/library/ocutil.py
```diff
@@ -0,0 +1,41 @@
...
+"""Interface to OpenShift oc command"""
+
+import shlex
+import subprocess
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def main():
+    module = AnsibleModule(
```
jcantrill: Have we considered moving the checks from 'oadm diagnostics AggregatedLogging' to here.  I think they would make more sense to live in something that will change quicker then oadm

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
```diff
@@ -0,0 +1,257 @@
...
+            results['used'] = int(disk_used)
+            results['free'] = int(disk_free)
+        except ValueError:
+            results['used'] = int(0)
+            results['free'] = int(0)
+
+        return results
+
+    def check_fluentd(self, fluentd_pods, task_vars):
+        """Verify fluentd is running"""
```
jcantrill: We have some initial work on the logging repo to check fluentd status with 'how much it has read' and 'how much is left'.  This is to try to address 'logging records are slow to show in kibana' issues

roles/openshift_health_checker/openshift_checks/efk_stack_check.py
rhcarvalho: Let's not make it up if it doesn't exist.
Could you check how this is done in the logging installation?

roles/openshift_health_checker/library/ocutil.py
rhcarvalho: @jcantrill that may happen eventually, but not right now in this PR.

rhcarvalho:
> I think I'd actually like to work directly on this when I have some free time, as I'm reasonably familiar with EFK :)

Hey, that would be a win-win situation for us all :-)

How about you push some commits to this PR?
You can do that by:

```console
$ git fetch https://github.com/juanvallejo/openshift-ansible jvallejo/elastic-search-check:jvallejo/elastic-search-check
$ git checkout jvallejo/elastic-search-check
...
$ git commit
$ git push https://github.com/juanvallejo/openshift-ansible jvallejo/elastic-search-check
```

roles/openshift_health_checker/library/ocutil.py
```diff
@@ -0,0 +1,43 @@
...
+        argument_spec=dict(
+            namespace=dict(type="str", required=True),
+            config_file=dict(type="str", required=True),
+            cmd=dict(type="str", required=True)
+        ),
+    )
+
+    cmd = ['oc',
```
sosiouxme: I just tried running this with a containerized install and of course... no `oc`.

I'm going to look at how other code handles OpenShift API interaction.

sosiouxme: Actually, `oc` is there; it's just not in root's path.

Logging code handles it by consulting `openshift.common.client_binary` for the location of the binary.

rhcarvalho: See also: [roles/lib_openshift/src/lib/base.py#L15-L35](https://github.com/openshift/openshift-ansible/blob/6978114e166ce5fd0737e0c2957a7315d4e5fa41/roles/lib_openshift/src/lib/base.py#L15-L35)

juanvallejo: Thanks, tried using `openshift.common.client_binary` but, at least in my case, it just returned `oc` as its value, no full path. Copied the `locate_oc_binary` func from `base.py`

sosiouxme:
It worked on my containerized instance after I changed it to pass in
`openshift.common.client_binary` from the tasks_vars. Whatever works...

richm:
Can you give me an example of how you install this (if applicable), run this, and what the output looks like?

sosiouxme:
@richm This will run from an Ansible playbook that doesn't exist yet in openshift-ansible. You'll be able to run it via the openshift-ansible Docker image that it will be bundled in. For now, obviously we're hacking using the source. Output will look like a bunch of Ansible garbage followed by a legible summary of any problems. This is undergoing major revision so I don't know that now's a good time for trying it out.

juanvallejo:
@richm 
> Can you give me an example of how you install this (if applicable), run this, and what the output looks like?

Sure, you can first deploy an OpenShift cluster with logging enabled by setting `openshift_hosted_logging_deploy` to `true` in your inventory. Then, a quick way to run this check would be by editing the playbook in `playbooks/byo/openshift-preflight/check.yml`, replacing the word `"@preflight"` with `"@health"` so that it looks similar to:

```
---
- hosts: OSEv3
  name: run OpenShift health checks
  roles:
    - openshift_health_checker
  post_tasks:
    # NOTE: we need to use the old "action: name" syntax until
    # https://github.com/ansible/ansible/issues/20513 is fixed.
      args:
        checks:
          - '@health'
```

In the case of a logging component failure, you should see output similar to:

```
Play:     Run Docker image checks
     Details:  check "efk_stack":
               The following errors occurred:
               The route for the Kibana logging url is not being routed, so we cannot check if it works. Is a router running?
```

richm:
I can't seem to get this to work. The first problem I ran into is the EFKStack class has the wrong arguments:
```
TypeError: __init__() got an unexpected keyword argument 'execute_module'
```
So I changed the __init__ to this:
```
    def __init__(self, **kwargs):
        super(EFKStack, self).__init__(kwargs)
```
This is the output:
```
CHECK [efk_stack : openshift] **************************************************
ok: [openshift] => {
    "changed": false,
    "checks": {
        "efk_stack": {
            "skipped": true
        }
    },
    "invocation": {
        "module_args": {
            "checks": [
                "@health"
            ]
        },
        "module_name": "openshift_health_check"
    }
}
```
Did it actually do anything? It says `"skipped": true`

richm:
this works
```
    def __init__(self, **kwargs):
        super(EFKStack, self).__init__(**kwargs)
```

juanvallejo:
@richm hm, `execute_module` has been renamed to `module_executor`, I am not sure why `__init__` is still passing the old arg name, but this will hopefully be fixed once I add the changes that @sosiouxme is working on. Changing the arg name to `module_executor` in init should fix this for now.

Also, if the check is being skipped, make sure that your ansible remote host has [`masters` as part of its `group_names`](https://github.com/openshift/openshift-ansible/pull/3643/files#diff-d6b5b7f8f9fae50eabe9dd6baaf55d73R31)

sosiouxme:
@richm you might try my branch https://github.com/sosiouxme/openshift-ansible/tree/20170418-efk-check

I'm still cleaning up a few things but it's getting close. I included a playbook so you can run it as:

    ansible-playbook -i hosts playbooks/common/openshift-checks/health.yml

Re skipping it is also normal for non-master hosts to be skipped. Actually it's only supposed to run on the first master but that's one of the things I need to clean up.

rhcarvalho:
> hm, execute_module has been renamed to module_executor

Note: it's the opposite -- the original name was `module_executor`, the new and preferred name is `execute_module`. Either name should work for the time being.

rhcarvalho:
> The first problem I ran into is the EFKStack class has the wrong arguments:
> 
> `TypeError: __init__() got an unexpected keyword argument 'execute_module'`

This is because the EFKStack is overriding `__init__` and only supporting the old name. The code that instantiates the class uses the new name, and thus the `TypeError`.

The fix is to update the argument name to `execute_module`. Alternatively, we don't really need `__init__`, as all it does could be done as a class attribute.

sosiouxme:
What would it look like to redo `__init__` with class attributes?

See
https://github.com/juanvallejo/openshift-ansible/pull/3/files#diff-d6b5b7f8f9fae50eabe9dd6baaf55d73R21

rhcarvalho:
> What would it look like to redo `__init__` with class attributes?

```python
class EFK...:
    ...
    es_curl = 'curl -s --cert /etc/elasticsearch/secret/admin-cert --key /etc/elasticsearch/secret/admin-key --cacert /etc/elasticsearch/secret/admin-ca -XGET '
    ....
```

BTW, what is really `es_curl`? There's probably a better name for it.

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -0,0 +1,348 @@
...
+
+            es_health_check = self.check_es_cluster_health(pod_name, task_vars)
+            es_status['pods'][pod_dc]['elasticsearch_health'] = es_health_check
+
+            es_disk_check = self.check_elasticsearch_diskspace(pod_name, task_vars)
+            es_status['pods'][pod_dc]['disk'] = es_disk_check
+
+            # Compare the master across all ES nodes to see if we have split brain
+            curl_cmd = "{} 'https://localhost:9200/_cat/master'".format(self.es_curl)
+            es_master = "exec -ti {} -- {}".format(pod_name, curl_cmd)
```
rhcarvalho: Suggestion: perhaps we can hide all this into a method, we build the command in multiple places.

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -5,344 +5,601 @@
...
+                for condition in pod['status']['conditions']
+            )
+        ]
+
+    def _running_es_pods(self, es_pods):
+        '''Returns: list of running pods, list of errors about non-running pods'''
+        not_running = self._not_running_pods(es_pods)
+        if not_running:
+            return [pod for pod in es_pods if pod not in not_running], [(
+                'The following Elastic Search pods are not running:\n'
```
richm: elsewhere you use `elasticsearch` - it is one word, not two - in order to be consistent, when printing a message to be displayed to the end user, we should use `elasticsearch` or `Elasticsearch` everywhere

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -5,344 +5,601 @@
...
+    def check_elasticsearch(self, es_pods, task_vars):
+        '''Various checks for elasticsearch. Returns: error string'''
+        running_pods, error_msgs = self._running_es_pods(es_pods)
+        pods_by_name = {
+            pod['metadata']['name']: pod for pod in running_pods
+            # Filter out pods that are not members of a DC
+            if pod['metadata'].get('labels', {}).get('deploymentconfig')
+        }
+        if not pods_by_name:
+            return 'No logging Elastic Search pods were found. Is logging deployed?'
```
richm: see above

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -5,344 +5,601 @@
...
-    def check_fluentd(self, fluentd_pods, task_vars):
-        """Verify fluentd is running"""
-        fluentd_status = dict(
-            number_expected_pods=0,
-            number_pods=0,
-            node_mismatch=0,
-            running=0,
-        )
+                error_msgs.append(
+                    'No master? Elastic Search {pod} returned bad string when asked master name:\n'
```
richm: see above

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -5,344 +5,601 @@
...
+                    '  {response}'.format(pod=pod_name, response=master_name_str)
+                )
 
-        # Get all nodes with fluentd label
-        try:
-            nodes = json.loads(self.exec_oc("get nodes -o json", [], task_vars))
-        except ValueError:
-            return fluentd_status
+        if not es_master_names:
+            error_msgs.append('No logging Elastic Search masters were found. Is logging deployed?')
```
richm: see above

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -5,344 +5,601 @@
...
+            return '\n'.join(error_msgs)
 
-        fluentd_nodes = []
-        for node in nodes['items']:
-            if 'logging-infra-fluentd' in node['metadata']['labels']:
-                if node['metadata']['labels']['logging-infra-fluentd'] == 'true':
-                    fluentd_nodes.append(node)
+        if len(es_master_names) > 1:
+            error_msgs.append(
+                'Found multiple Elastic Search masters according to the pods:\n'
```
richm: see above

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -5,344 +5,601 @@
...
 
-        for pod in fluentd_pods:
-            node_matched = False
+    def _check_elasticsearch_node_list(self, pods_by_name, task_vars):
+        '''Check that reported ES masters are accounted for by pods. Returns: list of error strings'''
 
-            if pod['status']['containerStatuses'][0]['ready'] is False:
-                fluentd_status['running'] = 0
+        if not pods_by_name:
+            return ['No logging Elastic Search masters were found. Is logging deployed?']
```
richm: see above

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -5,344 +5,601 @@
...
-                        internal_ip = address['address']
+        # get ES cluster nodes
+        curl_cmd = "{} 'https://localhost:9200/_nodes'".format(self.es_curl)
+        node_cmd = "exec {} -- {}".format(pods_by_name.keys()[0], curl_cmd)
+        cluster_node_data = self._exec_oc(node_cmd, [], task_vars)
+        try:
+            cluster_nodes = json.loads(cluster_node_data)['nodes']
+        except (ValueError, KeyError):
+            return [
+                'Failed to query Elastic Search for the list of ES nodes. The output was:\n' +
```
richm: see above

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -5,344 +5,601 @@
...
+            ]
+
+        # Try to match all ES-reported node hosts to known pods.
+        error_msgs = []
+        for node in cluster_nodes.values():
+            # Note that with 1.4/3.4 the pod IP may be used as the master name
+            if not any(node['host'] in (pod_name, pod['status'].get('podIP'))
+                       for pod_name, pod in pods_by_name.items()):
+                error_msgs.append(
+                    'The Elastic Search cluster reports a member node "{node}"\n'
```
richm: see above

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -5,344 +5,601 @@
...
 
-                if node['metadata']['labels']['kubernetes.io/hostname'] == pod.get('spec', {}).get('host'):
-                    node_matched = True
-                    break
-                elif internal_ip == pod['spec']['nodeName'] or node['metadata']['name'] == pod['spec']['nodeName']:
-                    node_matched = True
-                    break
+            if health_res['status'] not in ['green', 'yellow']:
+                error_msgs.append(
+                    'Elastic Search cluster health status is RED according to pod "{}"'.format(pod_name)
```
richm: see above

roles/openshift_health_checker/test/efk_stack_test.py
```diff
@@ -1,240 +1,252 @@
...
+        "conditions": [{"status": "True", "type": "Ready"}],
+        "containerStatuses": [{"ready": True}],
+        "podIP": "10.10.10.10",
+    },
+    "_test_master_name_str": "name logging-es-2",
+}
+
+
+def test_check_elasticsearch():
+    assert 'No logging Elastic Search pods' in canned_efkstack().check_elasticsearch([], {})
```
richm: see above

roles/openshift_health_checker/test/efk_stack_test.py
```diff
@@ -1,240 +1,252 @@
...
+
+
+def pods_by_name(pods):
+    return {pod['metadata']['name']: pod for pod in pods}
+
+
+@pytest.mark.parametrize('pods, expect_error', [
+    (
+        [],
+        'No logging Elastic Search masters',
```
richm: see above

roles/openshift_health_checker/test/efk_stack_test.py
```diff
@@ -1,240 +1,252 @@
...
+    ),
+    (
+        [plain_es_pod],
+        None,
+    ),
+    (
+        [plain_es_pod, split_es_pod],
+        'Found multiple Elastic Search masters',
```
richm: see above

roles/openshift_health_checker/test/efk_stack_test.py
```diff
@@ -1,240 +1,252 @@
...
-                "host": "10.10.10.10",
-            },
-        ],
-        {
-            "logging-es": "ABcdEfGHiJk1234 10.10.10.10 10.10.10.10 Node Name",
-        },
-        ["find any elasticsearch pods"],
+        [],
+        {},
+        'No logging Elastic Search masters',
```
richm: see above

roles/openshift_health_checker/test/efk_stack_test.py
```diff
@@ -1,240 +1,252 @@
...
+    ),
+    (
+        [plain_es_pod],
+        [{"no-status": "should bomb"}],
+        'Could not retrieve cluster health status',
+    ),
+    (
+        [plain_es_pod, split_es_pod],
+        [{"status": "green"}, {"status": "red"}],
+        'Elastic Search cluster health status is RED',
```
richm: see above

roles/openshift_health_checker/openshift_checks/efk_stack.py
sosiouxme: Agreed the user messaging should be consistent. Somewhere I got in my head that "Elastic Search" was what Elastic wanted to call it but I see [it's not](https://www.elastic.co/products/elasticsearch). Thanks.

sosiouxme:
Lots of changes. Think this is ready for another round of reviews. @rhcarvalho @codificat @juanvallejo etc

sosiouxme:
aos-ci-test

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -0,0 +1,604 @@
...
+        '''Various checks for elasticsearch. Returns: error string'''
+        running_pods, error_msgs = self._running_es_pods(es_pods)
+        pods_by_name = {
+            pod['metadata']['name']: pod for pod in running_pods
+            # Filter out pods that are not members of a DC
+            if pod['metadata'].get('labels', {}).get('deploymentconfig')
+        }
+        if not pods_by_name:
+            return 'No logging Elasticsearch pods were found. Is logging deployed?'
+        error_msgs += self._check_elasticsearch_masters(pods_by_name, task_vars)
```
rhcarvalho: Why do we run this even if there are no Elasticsearch pods running?

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -0,0 +1,604 @@
...
+
+        # Try to match all ES-reported node hosts to known pods.
+        error_msgs = []
+        for node in cluster_nodes.values():
+            # Note that with 1.4/3.4 the pod IP may be used as the master name
+            if not any(node['host'] in (pod_name, pod['status'].get('podIP'))
+                       for pod_name, pod in pods_by_name.items()):
+                error_msgs.append(
+                    'The Elasticsearch cluster reports a member node "{node}"\n'
+                    'that does not correspond to any known ES pod.'.format(node=node['host'])
```
rhcarvalho: s/ES/Elasticsearch ?

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -0,0 +1,604 @@
...
+        for pod_name in pods_by_name.keys():
+            cluster_health_cmd = self._build_es_curl_cmd(pod_name, 'https://localhost:9200/_cluster/health?pretty=true')
+            cluster_health_data = self._exec_oc(cluster_health_cmd, [], task_vars)
+            try:
+                health_res = json.loads(cluster_health_data)
+                if not health_res or not health_res.get('status'):
+                    raise ValueError()
+            except ValueError:
+                error_msgs.append(
+                    'Could not retrieve cluster health status from logging ES pod "{pod}".\n'
```
rhcarvalho: s/ES/Elasticsearch/g ?

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -0,0 +1,604 @@
...
+        )
+        if not check_error['kibana']:
+            check_error['kibana_route'] = self._check_kibana_route(task_vars)
+
+        errors = [err for err in check_error.values() if err]
+        if errors:
+            errors.insert(0, "The following logging stack issues were found:")
+            return {"failed": True, "changed": False, "msg": "\n-------\n".join(errors)}
+
+        # TODO(lmeyer): run it all again for the ops cluster
```
sosiouxme: The EFK stack can be deployed with an option where system logs are split out to go to a separate deployment of ES/Kibana/Curator, which we call the ops cluster. The ops script we were basing this off does not do anything to check the ops cluster, so I don't think it's as much a priority to check that as the main cluster, but it's an obvious expansion that will certainly be wanted. I would like to think of a reasonably clean way to accomplish it, but I've already taken so much time on this check that I think it can wait for a later time.

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -0,0 +1,604 @@
...
+        return {"failed": False, "changed": False, "msg": 'No problems found with logging stack.'}
+
+    def _get_pods(self, task_vars):
+        '''Get all pods and sort them by EFK component. Returns: pods by component, error string'''
+        pod_output = self._exec_oc("get pods -o json", [], task_vars)
+        try:
+            pods = json.loads(pod_output)
+            if not pods or not pods.get('items'):
+                raise ValueError()
```
sosiouxme: Because whether json.loads throws ValueError or the list just comes back empty I'd like the same response.

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -0,0 +1,604 @@
+# pylint: disable=missing-docstring
+'''
+  Module for performing checks on an elasticseatch, fluentd, and kibana stack
```
sosiouxme: :+1: I have clearly been looking at this too long. I noticed that weeks ago and didn't fix it :)

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -0,0 +1,604 @@
...
+        '''Check various things and gather errors. Returns: result as hash'''
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
+        pods_for, error = self._get_pods(task_vars)
+        if error:
+            return {"failed": True, "changed": False, "msg": error}
+        check_error = dict(
+            elasticsearch=self.check_elasticsearch(pods_for['es'], task_vars),
+            fluentd=self.check_fluentd(pods_for['fluentd'], task_vars),
+            curator=self.check_curator(pods_for['curator']),
+            kibana=self.check_kibana(pods_for['kibana']),
```
sosiouxme: There aren't any dependencies between the methods for the different components at this time, so yeah, I suppose they could be split into separate checks for isolation/concurrency. There are a lot of module executions in this check and making those parallel could speed this up noticeably.

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -0,0 +1,604 @@
...
+            ) or not any(
+                condition['type'] == 'Ready' and condition['status'] == 'True'
+                for condition in pod['status']['conditions']
+            )
+        ]
+
+    def _running_es_pods(self, es_pods):
+        '''Returns: list of running pods, list of errors about non-running pods'''
+        not_running = self._not_running_pods(es_pods)
+        if not_running:
```
sosiouxme: I guess the method could as easily be _running_pods and the logic reversed. I guess I focus too much on the negative.

The method returns one or zero errors. The calling logic is simpler if it doesn't have to reason about that. All the other methods called in `check_elasticsearch` also return a list of errors so it's easier to just roll them all together. I admit it looks a little weird on this method, but it seemed to make sense in context...

roles/openshift_health_checker/openshift_checks/efk_stack.py
```diff
@@ -0,0 +1,604 @@
...
+            if any(
+                container['ready'] is False
+                for container in pod['status']['containerStatuses']
+            ) or not any(
+                condition['type'] == 'Ready' and condition['status'] == 'True'
+                for condition in pod['status']['conditions']
+            )
+        ]
+
+    def _running_es_pods(self, es_pods):
```
sosiouxme: It really is just because I want the messaging to be different for each component. Otherwise we could have a single `_errors_from_non_running_pods` method to run for all of them.

I wanted the top-level `check_elasticsearch` to be really easy to read.

I struggled for a long time about where to put messaging logic (what to say about what is observed) versus the evaluating logic it's based on. I did not want every leaf method to generate user-directed errors and have them cascade up, as then the messaging logic is really scattered around and it clutters up leaf methods and makes them hard to reuse. On the other hand I didn't want on huge messaging logic method at the top and didn't want to develop a system of packaging up results just so they can be evaluated legibly several levels up. So I struggled with the level of coupling between evaluating and messaging, and I don't know that I got the balance right, but it seemed worse other ways.

roles/openshift_health_checker/openshift_checks/efk_stack.py
sosiouxme: What's "this"? If there are no pods running the `return` above aborts any further testing.

roles/openshift_health_checker/openshift_checks/efk_stack.py
sosiouxme: If you want... I think once the full name has been mentioned it's safe to abbreviate

roles/openshift_health_checker/library/ocutil.py
```diff
@@ -0,0 +1,73 @@
...
+import subprocess
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+ADDITIONAL_PATH_LOOKUPS = ['/usr/local/bin', os.path.expanduser('~/bin')]
+
+
+def locate_oc_binary():
+    ''' Find and return oc binary file '''
```
sosiouxme: :+1: OK I didn't check the PEP and I got it wrong.

Should we also agree on regular quoted strings? Although it's not specified in a PEP I thought the standard there was single quotes...

roles/openshift_health_checker/test/curator_test.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@pytest.mark.parametrize('groups, logging_deployed, is_active', [
+    (groups_with_first_master, True, True),
+    (groups_with_first_master, False, False),
+    (groups_not_a_master, True, False),
+    (groups_with_second_master, True, False),
+    (groups_not_a_master, True, False),
+])
+def test_is_active(groups, logging_deployed, is_active):
```
rhcarvalho: We could avoid triplicating/duplicating this test for every check, the only thing that changes is the check class.

roles/openshift_health_checker/openshift_checks/curator.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @classmethod
+    def is_active(cls, task_vars):
+        is_first_master = super(Curator, cls).is_first_master(task_vars)
+        return super(Curator, cls).is_active(task_vars) and is_first_master
+
+    def run(self, tmp, task_vars):
+        """Check various things and gather errors. Returns: result as hash"""
+
+        super(Curator, self).logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
```
rhcarvalho: Why not `self.logging_namespace`?!

roles/openshift_health_checker/library/ocutil.py
rhcarvalho: https://www.python.org/dev/peps/pep-0008/#string-quotes

Indeed PEP8 doesn't make a recommendation, but one common approach is to follow what the Python REPL does: single quotes most of the time, double quotes when the string contains a single quote.

Now that said, it's likely overkill to go and change everything in one shot. And because we have no automated way to apply a standard / enforce it, there's little hope to achieve perfect consistency in this area ;)

My original comment was more in the intention to share a Python practice.

roles/openshift_health_checker/test/curator_test.py
juanvallejo: thanks, will only include this once

juanvallejo:

sosiouxme:
@juanvallejo can I suggest scoping these files a bit. It should be visually obvious they are grouped together. Also the more I think about that I don't like calling this the "EFK" check. EFKC is just the current implementation. What if we call it the "logging" check? And put `logging` in the file names, or even make a `logging` subdirectory and use it as a package name.

juanvallejo:
@sosiouxme 
> can I suggest scoping these files a bit. It should be visually obvious they are grouped together. Also the more I think about that I don't like calling this the "EFK" check. EFKC is just the current implementation. What if we call it the "logging" check? And put logging in the file names, or even make a logging subdirectory and use it as a package name.

Thanks, went ahead and made a `logging` subdirectory and renamed "EFKCheck" -> "LoggingCheck", ptal

roles/openshift_health_checker/openshift_checks/efk_check.py
```diff
@@ -0,0 +1,88 @@
...
+    def get_pods_for_component(self, execute_module, namespace, logging_component, task_vars):
+        """Get all pods for a given component. Returns: list of pods for component, error string"""
+        pod_output = self.exec_oc(execute_module, namespace, "get pods -o json", [], task_vars)
+        try:
+            pods = json.loads(pod_output)
+            if not pods or not pods.get('items'):
+                raise ValueError()
+        except ValueError:
+            # successful run but non-parsing data generally means there were no pods in the namespace
+            return None, 'There are no pods in the {} namespace. Is logging deployed?'.format(namespace)
```
sosiouxme: This was originally in the context of getting all pods in the namespace and collating.

How about `There are no "{component}" pods in the "{namespace}" project.`

"Is logging deployed" will hopefully be something they ask themselves when all four component checks abort for lack of pods.

roles/openshift_health_checker/test/elasticsearch_test.py
```diff
@@ -0,0 +1,208 @@
...
+
+def assert_error(error, expect_error):
+    if expect_error:
+        assert error
+        assert expect_error in error
+    else:
+        assert not error
+
+
+# ES tests -------------------------------------------------------------
```
sosiouxme: redundant :)

roles/openshift_health_checker/test/elasticsearch_test.py
```diff
@@ -0,0 +1,208 @@
...
+    ),
+])
+def test_check_elasticsearch_diskspace(disk_data, expect_error):
+    check = canned_elasticsearch(lambda cmd, args, task_vars: disk_data)
+
+    errors = check._check_elasticsearch_diskspace(pods_by_name([plain_es_pod]), task_vars_config_base)
+    assert_error(''.join(errors), expect_error)
+
+
+# utility tests -------------------------------------------------------------
```
sosiouxme: I think the test below belongs elsewhere...

roles/openshift_health_checker/test/kibana_test.py
```diff
@@ -0,0 +1,237 @@
...
+
+    check = canned_kibana()
+    check._get_kibana_url = lambda task_vars: ('url', None)
+    check._verify_url_internal = lambda url, task_vars: None
+
+    error = check._check_kibana_route({})
+    assert_error(error, expect)
+
+
+# utility tests -------------------------------------------------------------
```
sosiouxme: likewise... test belongs elsewhere?

roles/openshift_health_checker/openshift_checks/efk_check.py
```diff
@@ -0,0 +1,88 @@
...
+import json
+import os
+
+from openshift_checks import OpenShiftCheckException, get_var
+
+
+class EFKCheck(object):
+
+    @staticmethod
+    def is_first_master(task_vars):
```
juanvallejo: Since subclasses also inherit from OpenShiftCheck, naming this method `is_active` would cause it to be shadowed by the one defined in `OpenShiftCheck`. A solution might be to have `EFKCheck` inherit from `OpenShiftCheck`, and avoid multiple inheritance in each of the logging component classes (i.e. `logging/kibana.py` would have `EFKCheck` as its only parent, rather than both `EFKCheck` and `OpenShiftCheck`). This would also mean [altering this method to be recursive so that each check that only inherits from `EFKCheck` actually runs](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_health_checker/action_plugins/openshift_health_check.py#L72)

rhcarvalho:
> naming this method is_active would cause it to be shadowed by the one defined in OpenShiftCheck.

Well, not quite. That's one of the reasons why `super` exists.

See this `file.py`:

```python
class OpenShiftCheck(object):
    def is_active(self):
        print "OpenShiftCheck is_active called"

class EFKCheckMixin(object):
    def is_active(self):
        super(EFKCheckMixin, self).is_active() # parent must implement is_active for this to work
        print "EFKCheckMixin is_active called"

class Kibana(EFKCheckMixin, OpenShiftCheck):
    def is_active(self):
        super(Kibana, self).is_active()
        print "Kibana is_active called"

Kibana().is_active()
```

When we run it:

```console
$ python file.py 
OpenShiftCheck is_active called
EFKCheckMixin is_active called
Kibana is_active called
```

juanvallejo:
@rhcarvalho thanks for the clarification, ended up making `LoggingCheck` inherit from `OpenShiftCheck`, and every component check inherit from `LoggingCheck`, avoiding multiple inheritance altogether

juanvallejo:
aos-ci-test

juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
```diff
@@ -0,0 +1,218 @@
...
+            pod['metadata']['name']: pod for pod in running_pods
+            # Filter out pods that are not members of a DC
+            if pod['metadata'].get('labels', {}).get('deploymentconfig')
+        }
+        if not pods_by_name:
+            return 'No logging Elasticsearch pods were found. Is logging deployed?'
+        error_msgs += self._check_elasticsearch_masters(pods_by_name, task_vars)
+        error_msgs += self._check_elasticsearch_node_list(pods_by_name, task_vars)
+        error_msgs += self._check_es_cluster_health(pods_by_name, task_vars)
+        error_msgs += self._check_elasticsearch_diskspace(pods_by_name, task_vars)
```
portante: It would be great to warn the user if the persistent volume ends up being the same as the root volume.

roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
```diff
@@ -0,0 +1,218 @@
...
+                error_msgs.append(
+                    'Inode percent usage on the storage volume for logging ES pod "{pod}"\n'
+                    '  is {pct}, greater than threshold {limit}.\n'
+                    '  Note: threshold can be specified in inventory with {param}'.format(
+                        pod=pod_name,
+                        pct=str(inode_pct),
+                        limit=str(inode_pct_thresh),
+                        param='openshift_check_efk_es_inode_pct',
+                    ))
+            disk_pct_thresh = get_var(task_vars, 'openshift_check_efk_es_storage_pct', default='90')
```
portante: That default seems kind-of high.  Perhaps consider 85%, or even 75%, as Elasticsearch needs a fair amount of space to merge data in to large indices.

roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
```diff
@@ -0,0 +1,218 @@
...
+                '{pods}'
+                'These pods will not aggregate logs from their nodes.'
+            ).format(pods=''.join(
+                "  {} ({})\n".format(pod['metadata']['name'], pod['spec'].get('host', 'None'))
+                for pod in not_running
+            ))]
+        return not_running, []
+
+    def check_elasticsearch(self, es_pods, task_vars):
+        """Various checks for elasticsearch. Returns: error string"""
```
portante: We need to check that the Kibana template .json files are properly configured for building the base index patterns.

portante: And also that the oadm policy for elasticsearch to access the master API server.

portante: The node quorum and master quorum should be properly set to a quorum value appropriate for the number ((N / 2) + 1).

portante: The auto-replica-count should not be set.

portante: Default replica count value should be `<=` `MIN`(`2`, number of expected data nodes in the cluster - `1`).

roles/openshift_health_checker/openshift_checks/logging/fluentd.py
```diff
@@ -0,0 +1,174 @@
...
+class Fluentd(LoggingCheck):
+    """Module that checks an Fluentd deployment shipped with OCP"""
+
+    name = "fluentd"
+    tags = ["health", "logging"]
+
+    logging_namespace = None
+
+    def run(self, tmp, task_vars):
+        """Check various things and gather errors. Returns: result as hash"""
```
portante: Should check to make sure the common data model conf file is in place.

roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
juanvallejo: thanks, will update default value

juanvallejo:
@portante thanks for the feedback, I have tracked a number of these comments under a [Trello card](https://trello.com/c/rC1SaZZo/114-logging-stack-improvements-additional-checks) to implement in an update to this check

juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
sosiouxme: @portante 

> We need to check that the Kibana template .json files are properly configured for building the base index patterns.

What would be a likely indicator that they're not? Would this mean that the user installed logging on an old version and didn't upgrade config, or something else?

sosiouxme: @portante 

> The auto-replica-count should not be set.
￼
For what? I'm just not sure what you're talking about here, please elaborate...

sosiouxme: @portante 
> Default replica count value should be <= MIN(2, number of expected data nodes in the cluster - 1).

Replica count value for what? Is this for the StatefulSet implementation of ES deployment or... I'm not quite following.

roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
sosiouxme: That would mean that the data volume didn't get mounted at all, not even a temporary EmptyDir. Users would have to manually modify the ES pod template to remove the mount. Have you seen someone actually do this and then act confused about what happens to their data?

roles/openshift_health_checker/openshift_checks/logging/fluentd.py
sosiouxme: Do you have a quick pointer to that? I assume this is a case of "EFK got left behind in an upgrade" if it's not there...

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -66,16 +66,30 @@ def subclasses(cls):
...
+
+    modules = []
+
+    for name in os.listdir(path):
+        if os.path.isdir(os.path.join(path, name)):
+            modules = modules + load_checks(os.path.join(path, name), subpkg + "." + name)
+            continue
+
+        if name.endswith(".py") and name not in LOADER_EXCLUDES:
+            if subpkg:
```
sosiouxme: no need for an `if` here, I think the next line works either way.

roles/openshift_health_checker/openshift_checks/logging/kibana.py
```diff
@@ -0,0 +1,247 @@
...
+
+        return None
+
+    def check_kibana(self, pods):
+        """Check to see if Kibana is up and working. Returns: error string."""
+
+        if not pods:
+            return "There are no Kibana pods deployed, so no access to the logging UI."
+
+        not_running = self._not_running_pods(pods)
```
sosiouxme: shouldn't this be `not_running_pods` from superclass?

sosiouxme:
aos-ci-test had a ... mishap yesterday. Looks like it's fixed today.

roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -0,0 +1,96 @@
+# pylint: disable=missing-docstring
```
juanvallejo: :) will stop adding this, promise

sdodson:
aos-ci-test

juanvallejo:
@sosiouxme thanks for the review, feedback addressed

sosiouxme:
@juanvallejo unit tests failing...

roles/openshift_health_checker/openshift_checks/logging/fluentd.py
```diff
@@ -0,0 +1,174 @@
...
+        except ValueError:  # no valid json - should not happen
+            return None, "Could not obtain a list of nodes to validate fluentd. Output from oc get:\n" + nodes_json
+        if not nodes or not nodes.get('items'):  # also should not happen
+            return None, "No nodes appear to be defined according to the API."
+        return {
+            node['metadata']['name']: node
+            for node in nodes['items']
+        }, None
+
+    def _exec_oc(self, cmd_str, extra_args, task_vars):
```
juanvallejo: Needed to add this on every class that needs to interface with `oc`, in order to replace the method with custom logic during tests

sosiouxme:
python nitpickers are failing...

juanvallejo:
aos-ci-test

sosiouxme:
aos-ci-test

rhcarvalho:
@juanvallejo I've updated the description, PTAL. Is this still fair to say this is based on https://github.com/openshift/openshift-tools/blob/prod/scripts/monitoring/cron-send-logging-checks.py?

roles/openshift_health_checker/library/ocutil.py
```diff
@@ -0,0 +1,74 @@
...
+import shutil
+import subprocess
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+ADDITIONAL_PATH_LOOKUPS = ['/usr/local/bin', os.path.expanduser('~/bin')]
+
+
+def locate_oc_binary():
```
sosiouxme: Would including `lib_openshift` be a problem? I expect lots of other stuff in the codebase does and it's just libraries...

roles/openshift_health_checker/openshift_checks/logging/fluentd.py
sosiouxme: I don't see why. Surely you can monkeypatch an inherited method?

roles/openshift_health_checker/library/ocutil.py
rhcarvalho: The problem is `lib_openshift` is a s*** load of things, and this "helper" function gets copied to several modules. I think it is not even accessible "as-is".
One way I can think of is to refactor such that there's a module that is responsible for finding out the `oc` binary and then we set the result as a fact. Then, no other module would need this code, instead code would just read the ansible variable.

rhcarvalho:
@juanvallejo @sosiouxme is there still some important change to be made here?
I don't like to admit this, but I'd be willing to simply merge what we have now than waiting for more CI cycles...

sosiouxme:
@rhcarvalho I think it can [merge] as-is. Such nitpicks as I still have are not important, and we have a laundry list of things to add in the next iteration anyway.

And it's still fair to say it was based on that script...

PR #3647, opened 2017-03-13
Allow user to specify additions to ES config

jcantrill:
This PRs allow a user to define a hash for the ES config that will be merged into the defaults that are defined for templates/elasticsearcy.yaml so to allow further tweaking of ES.

cc @richm

### DIFF
--- a/roles/openshift_logging/tasks/generate_configmaps.yaml
+++ b/roles/openshift_logging/tasks/generate_configmaps.yaml
@@ -6,8 +6,17 @@
       when: es_logging_contents is undefined
       changed_when: no
 
+    - local_action: >
+        copy content="{{ config_source | combine(override_config,recursive=True) | to_nice_yaml }}"
+        dest="{{local_tmp.stdout}}/elasticsearch-gen-template.yml"
+      vars:
+        config_source: "{{lookup('file','templates/elasticsearch.yml.j2') | from_yaml }}"
+        override_config: "{{openshift_logging_es_config | from_yaml}}"
+      when: es_logging_contents is undefined
+      changed_when: no
+
     - template:
-        src: elasticsearch.yml.j2
+        src: "{{local_tmp.stdout}}/elasticsearch-gen-template.yml"
         dest: "{{mktemp.stdout}}/elasticsearch.yml"
       vars:
         - allow_cluster_reader: "{{openshift_logging_es_ops_allow_cluster_reader | lower | default('false')}}"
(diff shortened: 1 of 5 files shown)


roles/openshift_logging/tasks/main.yaml
```diff
@@ -12,6 +12,14 @@
 
 - debug: msg="Created temp dir {{mktemp.stdout}}"
 
+- name: Create local temp directory for doing work in
```
ewolinetz: im confused, isn't this where this is currently in master?

jcantrill: I did not find it though I thought we made that change LONG ago.  Please correct me if I am basing this off a stale commit.

ewolinetz: its in the release-1.5 branch which was based off master...

jcantrill: @ewolinetz Per our IRC discussion this task moves the local temp directory creation that did not exist off master.  Are we good?  Are there any changes you see that are necessary?

ewolinetz: My only concern is that we are now creating a temp dir on the control host but then we are then doing a remote command to copy the kubeconfig to the created temp dir. This would work if our remote and control host are the same, but I'm not sure it will when they are different...

ewolinetz: Is there a reason we wouldn't keep this as a remote action?

roles/openshift_logging/tasks/generate_configmaps.yaml
```diff
@@ -6,8 +6,17 @@
       when: es_logging_contents is undefined
       changed_when: no
 
+    - local_action: >
```
ewolinetz: same question regarding local action as in tasks/main.yaml

jcantrill: I copied this locally because when running the template step, it failed.  I believe its because the module assumes 'src' is a location on the control node and the template module does not support src from memory.

roles/openshift_logging/tasks/main.yaml
jcantrill: This is a direct move from the jks tasks.  We make it local to the control node because thats the only way we have access to certain generated files, at least that was my experience.  <soapbox> I found this is one of the more confusing parts of ansible.  Seems some things need/do run locally or have local dependencies and others do not and its not obvious what is what. </soapbox>

jcantrill:
aos-ci-test

jcantrill:
[merge]

PR #3649, opened 2017-03-14
Various preflight check improvements

sosiouxme:
[Edit: this PR had a lot of drift from its original purpose]
- ~Improves the output of failed checks so that a customer might make sense of it.
https://trello.com/c/K4dPuCHa/98-improve-ansible-diagnostics-failure-output~ (moved to different PR)
- ~package content checks need to discount the openshift excluders, as they wouldn't be in effect during an install/upgrade~ (moved to different PR)
- ~introduced a version normalizer for openshift_release~ (no longer necessary)
- refactored `aos_version` check module
- https://trello.com/c/NdWoIydb/82-diagnostics-content-checks-should-work-with-origin - they already did, but a minor tweak was needed for version checking
- added `openshift_repos` dependency so origin repos will be set up prior to checks

Note: ~this PR is based on https://github.com/openshift/openshift-ansible/pull/3816 integration test work.~ removed so PR can progress first, can add back in later PR.

### DIFF
--- a/roles/openshift_health_checker/library/aos_version.py
+++ b/roles/openshift_health_checker/library/aos_version.py
@@ -1,91 +1,199 @@
 #!/usr/bin/python
 # vim: expandtab:tabstop=4:shiftwidth=4
 '''
-Ansible module for determining if multiple versions of an OpenShift package are
-available, and if the version requested is available down to the given
-precision.
+Ansible module for yum-based systems determining if multiple releases
+of an OpenShift package are available, and if the release requested
+(if any) is available down to the given precision.
 
-Multiple versions available suggest that multiple repos are enabled for the
-different versions, which may cause installation problems.
+For Enterprise, multiple releases available suggest that multiple repos
+are enabled for the different releases, which may cause installation
+problems. With Origin, however, this is a normal state of affairs as
+all the releases are provided in a single repo with the expectation that
+only the latest can be installed.
+
+Code in the openshift_version role contains a lot of logic to pin down
+the exact package and image version to use and so does some validation
+of release availability already. Without duplicating all that, we would
+like the user to have a helpful error message if we detect things will
+not work out right. Note that if openshift_release is not specified in
+the inventory, the version comparison checks just pass.
+
+TODO: fail gracefully on non-yum systems (dnf in Fedora)
 '''
-
-import yum  # pylint: disable=import-error
 
 from ansible.module_utils.basic import AnsibleModule
 
+IMPORT_EXCEPTION = None
+try:
+    import yum  # pylint: disable=import-error
+except ImportError as err:
+    IMPORT_EXCEPTION = err  # in tox test env, yum import fails
 
-def main():  # pylint: disable=missing-docstring,too-many-branches
+
+class AosVersionException(Exception):
+    '''Base exception class for package version problems'''
+    def __init__(self, message, problem_pkgs=None):
+        Exception.__init__(self, message)
+        self.problem_pkgs = problem_pkgs
+
+
+def main():
+    '''Entrypoint for this Ansible module'''
     module = AnsibleModule(
         argument_spec=dict(
-            prefix=dict(required=True),  # atomic-openshift, origin, ...
-            version=dict(required=True),
+            requested_openshift_release=dict(type="str", default=''),
+            openshift_deployment_type=dict(required=True),
+            rpm_prefix=dict(required=True),  # atomic-openshift, origin, ...?
         ),
         supports_check_mode=True
     )
 
-    def bail(error):  # pylint: disable=missing-docstring
-        module.fail_json(msg=error)
+    if IMPORT_EXCEPTION:
+        module.fail_json(msg="aos_version module could not import yum: %s" % IMPORT_EXCEPTION)
 
-    rpm_prefix = module.params['prefix']
-
+    # determine the packages we will look for
+    rpm_prefix = module.params['rpm_prefix']
     if not rpm_prefix:
-        bail("prefix must not be empty")
-
-    yb = yum.YumBase()  # pylint: disable=invalid-name
-    yb.conf.disable_excludes = ["all"]  # assume the openshift excluder will be managed, ignore current state
-
-    # search for package versions available for aos pkgs
-    expected_pkgs = [
+        module.fail_json(msg="rpm_prefix must not be empty")
+    expected_pkgs = set([
         rpm_prefix,
         rpm_prefix + '-master',
         rpm_prefix + '-node',
-    ]
+    ])
+
+    # determine what level of precision the user specified for the openshift version.
+    # should look like a version string with possibly many segments e.g. "3.4.1":
+    requested_openshift_release = module.params['requested_openshift_release']
+
+    # get the list of packages available and complain if anything is wrong
+    try:
+        pkgs = _retrieve_available_packages(expected_pkgs)
+        if requested_openshift_release:
+            _check_precise_version_found(pkgs, expected_pkgs, requested_openshift_release)
+            _check_higher_version_found(pkgs, expected_pkgs, requested_openshift_release)
+        if module.params['openshift_deployment_type'] in ['openshift-enterprise']:
+            _check_multi_minor_release(pkgs, expected_pkgs)
+    except AosVersionException as excinfo:
+        module.fail_json(msg=str(excinfo))
+    module.exit_json(changed=False)
+
+
+def _retrieve_available_packages(expected_pkgs):
+    # search for package versions available for openshift pkgs
+    yb = yum.YumBase()  # pylint: disable=invalid-name
+
+    # The openshift excluder prevents unintended updates to openshift
+    # packages by setting yum excludes on those packages. See:
+    # https://wiki.centos.org/SpecialInterestGroup/PaaS/OpenShift-Origin-Control-Updates
+    # Excludes are then disabled during an install or upgrade, but
+    # this check will most likely be running outside either. When we
+    # attempt to determine what packages are available via yum they may
+    # be excluded. So, for our purposes here, disable excludes to see
+    # what will really be available during an install or upgrade.
+    yb.conf.disable_excludes = ['all']
+
     try:
         pkgs = yb.pkgSack.returnPackages(patterns=expected_pkgs)
-    except yum.Errors.PackageSackError as e:  # pylint: disable=invalid-name
+    except yum.Errors.PackageSackError as excinfo:
         # you only hit this if *none* of the packages are available
-        bail('Unable to find any OpenShift packages.\nCheck your subscription and repo settings.\n%s' % e)
+        raise AosVersionException('\n'.join([
+            'Unable to find any OpenShift packages.',
+            'Check your subscription and repo settings.',
+            str(excinfo),
+        ]))
+    return pkgs
 
-    # determine what level of precision we're expecting for the version
-    expected_version = module.params['version']
-    if expected_version.startswith('v'):  # v3.3 => 3.3
-        expected_version = expected_version[1:]
-    num_dots = expected_version.count('.')
 
-    pkgs_by_name_version = {}
+class PreciseVersionNotFound(AosVersionException):
+    '''Exception for reporting packages not available at given release'''
+    def __init__(self, requested_release, not_found):
+        msg = ['Not all of the required packages are available at requested version %s:' % requested_release]
+        msg += ['  ' + name for name in not_found]
+        msg += ['Please check your subscriptions and enabled repositories.']
+        AosVersionException.__init__(self, '\n'.join(msg), not_found)
+
+
+def _check_precise_version_found(pkgs, expected_pkgs, requested_openshift_release):
+    # see if any packages couldn't be found at requested release version
+    # we would like to verify that the latest available pkgs have however specific a version is given.
+    # so e.g. if there is a package version 3.4.1.5 the check passes; if only 3.4.0, it fails.
+
     pkgs_precise_version_found = {}
     for pkg in pkgs:
-        # get expected version precision
-        match_version = '.'.join(pkg.version.split('.')[:num_dots + 1])
-        if match_version == expected_version:
+        if pkg.name not in expected_pkgs:
+            continue
+        # does the version match, to the precision requested?
+        # and, is it strictly greater, at the precision requested?
+        match_version = '.'.join(pkg.version.split('.')[:requested_openshift_release.count('.') + 1])
+        if match_version == requested_openshift_release:
             pkgs_precise_version_found[pkg.name] = True
-        # get x.y version precision
-        minor_version = '.'.join(pkg.version.split('.')[:2])
-        if pkg.name not in pkgs_by_name_version:
-            pkgs_by_name_version[pkg.name] = {}
-        pkgs_by_name_version[pkg.name][minor_version] = True
 
-    # see if any packages couldn't be found at requested version
-    # see if any packages are available in more than one minor version
     not_found = []
-    multi_found = []
     for name in expected_pkgs:
         if name not in pkgs_precise_version_found:
             not_found.append(name)
+
+    if not_found:
+        raise PreciseVersionNotFound(requested_openshift_release, not_found)
+
+
+class FoundHigherVersion(AosVersionException):
+    '''Exception for reporting that a higher version than requested is available'''
+    def __init__(self, requested_release, higher_found):
+        msg = ['Some required package(s) are available at a version',
+               'that is higher than requested %s:' % requested_release]
+        msg += ['  ' + name for name in higher_found]
+        msg += ['This will prevent installing the version you requested.']
+        msg += ['Please check your enabled repositories or adjust openshift_release.']
+        AosVersionException.__init__(self, '\n'.join(msg), higher_found)
+
+
+def _check_higher_version_found(pkgs, expected_pkgs, requested_openshift_release):
+    req_release_arr = [int(segment) for segment in requested_openshift_release.split(".")]
+    # see if any packages are available in a version higher than requested
+    higher_version_for_pkg = {}
+    for pkg in pkgs:
+        if pkg.name not in expected_pkgs:
+            continue
+        version = [int(segment) for segment in pkg.version.split(".")]
+        too_high = version[:len(req_release_arr)] > req_release_arr
+        higher_than_seen = version > higher_version_for_pkg.get(pkg.name, [])
+        if too_high and higher_than_seen:
+            higher_version_for_pkg[pkg.name] = version
+
+    if higher_version_for_pkg:
+        higher_found = []
+        for name, version in higher_version_for_pkg.items():
+            higher_found.append(name + '-' + '.'.join(str(segment) for segment in version))
+        raise FoundHigherVersion(requested_openshift_release, higher_found)
+
+
+class FoundMultiRelease(AosVersionException):
+    '''Exception for reporting multiple minor releases found for same package'''
+    def __init__(self, multi_found):
+        msg = ['Multiple minor versions of these packages are available']
+        msg += ['  ' + name for name in multi_found]
+        msg += ["There should only be one OpenShift release repository enabled at a time."]
+        AosVersionException.__init__(self, '\n'.join(msg), multi_found)
+
+
+def _check_multi_minor_release(pkgs, expected_pkgs):
+    # see if any packages are available in more than one minor version
+    pkgs_by_name_version = {}
+    for pkg in pkgs:
+        # keep track of x.y (minor release) versions seen
+        minor_release = '.'.join(pkg.version.split('.')[:2])
+        if pkg.name not in pkgs_by_name_version:
+            pkgs_by_name_version[pkg.name] = {}
+        pkgs_by_name_version[pkg.name][minor_release] = True
+
+    multi_found = []
+    for name in expected_pkgs:
         if name in pkgs_by_name_version and len(pkgs_by_name_version[name]) > 1:
             multi_found.append(name)
-    if not_found:
-        msg = 'Not all of the required packages are available at requested version %s:\n' % expected_version
-        for name in not_found:
-            msg += '  %s\n' % name
-        bail(msg + 'Please check your subscriptions and enabled repositories.')
-    if multi_found:
-        msg = 'Multiple minor versions of these packages are available\n'
-        for name in multi_found:
-            msg += '  %s\n' % name
-        bail(msg + "There should only be one OpenShift version's repository enabled at a time.")
 
-    module.exit_json(changed=False)
+    if multi_found:
+        raise FoundMultiRelease(multi_found)
 
 
 if __name__ == '__main__':
--- a/roles/openshift_health_checker/openshift_checks/package_version.py
+++ b/roles/openshift_health_checker/openshift_checks/package_version.py
@@ -10,11 +10,9 @@ class PackageVersion(NotContainerizedMixin, OpenShiftCheck):
     tags = ["preflight"]
 
     def run(self, tmp, task_vars):
-        rpm_prefix = get_var(task_vars, "openshift", "common", "service_type")
-        openshift_release = get_var(task_vars, "openshift_release")
-
         args = {
-            "prefix": rpm_prefix,
-            "version": openshift_release,
+            "requested_openshift_release": get_var(task_vars, "openshift_release", default=''),
+            "openshift_deployment_type": get_var(task_vars, "openshift_deployment_type"),
+            "rpm_prefix": get_var(task_vars, "openshift", "common", "service_type"),
         }
         return self.execute_module("aos_version", args, tmp, task_vars)
--- a/roles/openshift_health_checker/test/aos_version_test.py
+++ b/roles/openshift_health_checker/test/aos_version_test.py
@@ -1,0 +1,120 @@
+import pytest
+import aos_version
+
+from collections import namedtuple
+Package = namedtuple('Package', ['name', 'version'])
+
+expected_pkgs = set(['spam', 'eggs'])
+
+
+@pytest.mark.parametrize('pkgs, requested_release, expect_not_found', [
+    (
+        [],
+        '3.2.1',
+        expected_pkgs,  # none found
+    ),
+    (
+        [Package('spam', '3.2.1')],
+        '3.2',
+        ['eggs'],  # completely missing
+    ),
+    (
+        [Package('spam', '3.2.1'), Package('eggs', '3.3.2')],
+        '3.2',
+        ['eggs'],  # not the right version
+    ),
+    (
+        [Package('spam', '3.2.1'), Package('eggs', '3.2.1')],
+        '3.2',
+        [],  # all found
+    ),
+    (
+        [Package('spam', '3.2.1'), Package('eggs', '3.2.1.5')],
+        '3.2.1',
+        [],  # found with more specific version
+    ),
+    (
+        [Package('eggs', '1.2.3'), Package('eggs', '3.2.1.5')],
+        '3.2.1',
+        ['spam'],  # eggs found with multiple versions
+    ),
+])
+def test_check_pkgs_for_precise_version(pkgs, requested_release, expect_not_found):
+    if expect_not_found:
+        with pytest.raises(aos_version.PreciseVersionNotFound) as e:
+            aos_version._check_precise_version_found(pkgs, expected_pkgs, requested_release)
+        assert set(expect_not_found) == set(e.value.problem_pkgs)
+    else:
+        aos_version._check_precise_version_found(pkgs, expected_pkgs, requested_release)
+
+
+@pytest.mark.parametrize('pkgs, requested_release, expect_higher', [
+    (
+        [],
+        '3.2.1',
+        [],
+    ),
+    (
+        [Package('spam', '3.2.1')],
+        '3.2',
+        [],  # more precise but not strictly higher
+    ),
+    (
+        [Package('spam', '3.3')],
+        '3.2.1',
+        ['spam-3.3'],  # lower precision, but higher
+    ),
+    (
+        [Package('spam', '3.2.1'), Package('eggs', '3.3.2')],
+        '3.2',
+        ['eggs-3.3.2'],  # one too high
+    ),
+    (
+        [Package('eggs', '1.2.3'), Package('eggs', '3.2.1.5'), Package('eggs', '3.4')],
+        '3.2.1',
+        ['eggs-3.4'],  # multiple versions, one is higher
+    ),
+    (
+        [Package('eggs', '3.2.1'), Package('eggs', '3.4'), Package('eggs', '3.3')],
+        '3.2.1',
+        ['eggs-3.4'],  # multiple versions, two are higher
+    ),
+])
+def test_check_pkgs_for_greater_version(pkgs, requested_release, expect_higher):
+    if expect_higher:
+        with pytest.raises(aos_version.FoundHigherVersion) as e:
+            aos_version._check_higher_version_found(pkgs, expected_pkgs, requested_release)
+        assert set(expect_higher) == set(e.value.problem_pkgs)
+    else:
+        aos_version._check_higher_version_found(pkgs, expected_pkgs, requested_release)
+
+
+@pytest.mark.parametrize('pkgs, expect_to_flag_pkgs', [
+    (
+        [],
+        [],
+    ),
+    (
+        [Package('spam', '3.2.1')],
+        [],
+    ),
+    (
+        [Package('spam', '3.2.1'), Package('eggs', '3.2.2')],
+        [],
+    ),
+    (
+        [Package('spam', '3.2.1'), Package('spam', '3.3.2')],
+        ['spam'],
+    ),
+    (
+        [Package('eggs', '1.2.3'), Package('eggs', '3.2.1.5'), Package('eggs', '3.4')],
+        ['eggs'],
+    ),
+])
+def test_check_pkgs_for_multi_release(pkgs, expect_to_flag_pkgs):
+    if expect_to_flag_pkgs:
+        with pytest.raises(aos_version.FoundMultiRelease) as e:
+            aos_version._check_multi_minor_release(pkgs, expected_pkgs)
+        assert set(expect_to_flag_pkgs) == set(e.value.problem_pkgs)
+    else:
+        aos_version._check_multi_minor_release(pkgs, expected_pkgs)
--- a/roles/openshift_health_checker/meta/main.yml
+++ b/roles/openshift_health_checker/meta/main.yml
@@ -1,3 +1,4 @@
 ---
 dependencies:
   - role: openshift_facts
+  - role: openshift_repos
--- a/roles/openshift_health_checker/test/conftest.py
+++ b/roles/openshift_health_checker/test/conftest.py
@@ -6,5 +6,6 @@ import sys
 openshift_health_checker_path = os.path.dirname(os.path.dirname(__file__))
 sys.path[1:1] = [
     openshift_health_checker_path,
-    os.path.join(openshift_health_checker_path, 'action_plugins')
+    os.path.join(openshift_health_checker_path, 'action_plugins'),
+    os.path.join(openshift_health_checker_path, 'library'),
 ]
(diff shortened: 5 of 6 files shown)


sosiouxme:
@rhcarvalho if you have some time to review...

roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
```diff
@@ -82,7 +82,15 @@ def _format_failure(failure):
         (u'Message', stringc(msg, C.COLOR_ERROR)),
     )
     if 'checks' in result._result:
-        rows += ((u'Details', stringc(pformat(result._result['checks']), C.COLOR_ERROR)),)
+        failed_check_msgs = ""
```
rhcarvalho: Suggestions:

- move this logic into a function
- avoid concatenating an arbitrary number of strings (puts pressure on the GC)

```python
def _format_failure(failure):
    # ...
    if 'checks' in result._result:
        rows += (('Failed checks', _format_failed_checks(result._result['checks'])),)
    # ...

def _format_failed_checks(checks):
    output = []
    for check, result in checks.items():
        if not result.get("failed", False):
            continue  # skip checks that did not fail
        output.append(...)
   return stringc("\n".join(output), C.COLOR_ERROR)
```

If you're feeling like refactoring, we could also change `rows` to be a `list`, then `.append` the failed check results into it.

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -19,6 +19,7 @@ def main():  # pylint: disable=missing-docstring,too-many-branches
         argument_spec=dict(
             prefix=dict(required=True),  # atomic-openshift, origin, ...
             version=dict(required=True),
+            short_version=dict(required=True),
```
rhcarvalho: Added an not used?

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -47,8 +49,6 @@ def bail(error):  # pylint: disable=missing-docstring
 
     # determine what level of precision we're expecting for the version
     expected_version = module.params['version']
-    if expected_version.startswith('v'):  # v3.3 => 3.3
-        expected_version = expected_version[1:]
```
rhcarvalho: Document the expected format of the input?

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -77,6 +77,19 @@ def get_var(task_vars, *keys, **kwargs):
...
+    """Helper function to normalize the release format to dotted numbers"""
+    release = str(get_var(task_vars, "openshift_release"))
+    if release.startswith('v'):
+        release = release[1:]  # v3.3 => 3.3
+    return release
+
+
+def normalized_minor_release(task_vars):
+    """Helper function to normalize the release format into x.y minor release"""
+    return '.'.join(normalized_release(task_vars).split('.')[:2])
```
rhcarvalho: How about we add some unit tests to those helpers?

roles/openshift_health_checker/openshift_checks/package_availability.py
```diff
@@ -11,56 +11,63 @@ class PackageAvailability(NotContainerizedMixin, OpenShiftCheck):
...
+                "cockpit-bridge",
+                "cockpit-docker",
+                "cockpit-kubernetes",
+                "cockpit-shell",
+                "cockpit-ws",
+                "etcd",
+                "httpd-tools",
+            ],
+        }
+        return pkgs[minor_version] if minor_version in pkgs else pkgs["current"]
```
rhcarvalho: Why are we changing this?!

rhcarvalho:
Maybe this could be 2 smaller and independent PRs next time?

roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
```diff
@@ -67,7 +67,7 @@ def _print_failure_summary(self):
 # pylint: disable=protected-access
 def _format_failure(failure):
     '''Return a list of pretty-formatted lines describing a failure, including
-    relevant information about it. Line separators are not included.'''
+    relevant information about it. Lines may include line breaks.'''
```
sosiouxme: Not sure how much better but I think "entries".

Agree the trailing newlines should be consistently left off.

roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
sosiouxme: Thanks, that's a little cleaner.

roles/openshift_health_checker/library/aos_version.py
sosiouxme: Yep...

roles/openshift_health_checker/library/aos_version.py
sosiouxme: I'll do that over in `openshift_checks/__init__.py`

roles/openshift_health_checker/openshift_checks/package_availability.py
sosiouxme: The idea is to vary the list depending on the minor version. It doesn't do that now because I haven't gathered the differences between versions yet.

rhcarvalho: How about we add the extra code when it is needed?

roles/openshift_health_checker/library/aos_version.py
rhcarvalho: If another check decides to use this ansible module, whoever gets to write that should have a clue that there are certain expectations about the version format.

I think the best chance we have of someone finding the information would be to have it live in the module itself?

Other than that, have the module be more defensive and make assertions about the input format.
And tests :-)

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: Should they look like https://github.com/openshift/openshift-ansible/tree/master/roles/lib_openshift/src/test ? I'm literally just not sure of the mechanics of how to get tests run by CI.

roles/openshift_health_checker/library/aos_version.py
rhcarvalho: The documentation for how to use a module should be in the module itself, not in a helper function from a different file that happens to be used in conjunction with the module through yet another file.

roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
```diff
@@ -66,25 +66,37 @@ def _print_failure_summary(self):
...
+    '''Return pretty-formatted text describing checks that failed.'''
+    failed_check_msgs = []
+    for check, body in checks.items():
+        if body.get('failed', False):   # only show the failed checks
+            msg = body.get('msg', u"Failed without returning a message")
+            failed_check_msgs.append('check "%s":\n%s' % (check, msg))
+    if failed_check_msgs:
+        return stringc("\n\n".join(failed_check_msgs), C.COLOR_ERROR)
+    else:    # something failed but no checks will admit to it, so dump everything
+        return stringc(pformat(result._result['checks']), C.COLOR_ERROR)
```
sosiouxme: Ah, good catch. If only I had unit tests...

rhcarvalho: Static analysis also catch that, I think both flake8 and pylint should point at it.

roles/openshift_health_checker/test/openshift_check_test.py
```diff
@@ -38,3 +38,24 @@ def test_get_var_error(task_vars, missing_keys):
 def test_get_var_default(task_vars, missing_keys):
     default = object()
     assert get_var(task_vars, *missing_keys, default=default) == default
+
+
+@pytest.mark.parametrize("release,expected", [
+    ('1.2', '1.2'),
+    (1.2, '1.2'),
+    ('v1.2', '1.2'),
+    ('v1.2.3.4', '1.2.3.4'),
```
rhcarvalho: - These two tests could probably have the same list of inputs (with different expected values, of course). They are now off by one.

- Please add a test for a base case: empty input.

- Please add test(s) for bad input.

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -10,11 +10,9 @@ class PackageVersion(NotContainerizedMixin, OpenShiftCheck):
...
 
     def run(self, tmp, task_vars):
-        rpm_prefix = get_var(task_vars, "openshift", "common", "service_type")
-        openshift_release = get_var(task_vars, "openshift_release")
-
         args = {
-            "prefix": rpm_prefix,
-            "version": openshift_release,
+            "prefix": get_var(task_vars, "openshift", "common", "service_type"),
+            # this normalizes openshift_release (from the inventory, or defaults in openshift_facts):
```
rhcarvalho: Unnecessary comment?
- The `normalized_release` name tells a good part of it (should we name it `normalized_openshift_release` instead? So that it is clear this is not Ansible's version or anything else?).
- The `openshift_release` can be set from a good number of places, the many ways to set variables in Ansible. I'd avoid suggesting where it was defined. Defer it to Ansible's doc.
  Or, if we really want to assist whoever is maintaining this code, then perhaps that would belong in a more central place, like a "developers guide" to health checks?

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -27,13 +28,18 @@ def bail(error):  # pylint: disable=missing-docstring
...
     rpm_prefix = module.params['prefix']
-
     if not rpm_prefix:
         bail("prefix must not be empty")
 
-    yb = yum.YumBase()  # pylint: disable=invalid-name
+    # determine what level of precision the user specified for the openshift version.
+    # should look like a version string with possibly many segments e.g. "3.4.1":
+    expected_version = module.params['version']
+    if not re.match(r'\A\d+(?:\.\d+)+\Z', expected_version):
```
rhcarvalho: [`re.match`](https://docs.python.org/3/library/re.html#re.match) and `\A` in the pattern is redundant.

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -27,13 +28,18 @@ def bail(error):  # pylint: disable=missing-docstring
         module.fail_json(msg=error)
 
     rpm_prefix = module.params['prefix']
-
     if not rpm_prefix:
         bail("prefix must not be empty")
 
-    yb = yum.YumBase()  # pylint: disable=invalid-name
+    # determine what level of precision the user specified for the openshift version.
+    # should look like a version string with possibly many segments e.g. "3.4.1":
```
rhcarvalho: IIUC we require the version to have at least 2 levels, "3" is considered invalid, correct?

roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
```diff
@@ -66,25 +66,37 @@ def _print_failure_summary(self):
 # Status: permanently disabled unless Ansible's API changes.
 # pylint: disable=protected-access
 def _format_failure(failure):
-    '''Return a list of pretty-formatted lines describing a failure, including
-    relevant information about it. Line separators are not included.'''
+    '''Return a list of pretty-formatted text entries describing a failure, including
+    relevant information about it. Line breaks can be included but not trailing newlines.'''
```
rhcarvalho: - :+1: s/lines/text entries is good, we now say that each entry is arbitrary text (in fact, it already was and already contained multiple lines, e.g., the pretty-formatted JSON).
- :-1: the second change doesn't add much clarity for me. There can certainly be trailing newlines, e.g., if we want to add extra space between sections. "Separator" is the term we find in the Python docs for splitting and joining text -- https://docs.python.org/3/library/stdtypes.html#str.split

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -78,7 +82,7 @@ def bail(error):  # pylint: disable=missing-docstring
         for name in not_found:
             msg += '  %s\n' % name
         bail(msg + 'Please check your subscriptions and enabled repositories.')
-    if multi_found:
+    if multi_found and rpm_prefix == 'atomic-openshift':
```
rhcarvalho: What's going on here?

roles/openshift_health_checker/test/openshift_check_test.py
sosiouxme: Sounds good. Ah, the empty input would cause an error elsewhere, but it's still good to check here. I wasn't sure if you thought the input validation/assertions should go here or in the modules that use the methods, but I'm thinking here.

rhcarvalho: I meant here, was just thinking about the functions.

roles/openshift_health_checker/openshift_checks/package_version.py
sosiouxme: Actually one thing I need to pin down today is expectations about specifying the release. None are documented in the advanced install guide, and the ansible code has a mishmash of `openshift_release`, `openshift_version`, and `openshift_image_tag` as well as `deployment_type` (all undocumented).

roles/openshift_health_checker/library/aos_version.py
sosiouxme: yep

roles/openshift_health_checker/library/aos_version.py
sosiouxme: The check for multiple minor versions present is really only relevant for enterprise. I assume this merits a comment... I only explained in the commit msg

roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
sosiouxme: So instead, "The list of text entries will be joined by a newline separator"

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -27,14 +28,18 @@ def bail(error):  # pylint: disable=missing-docstring
...
-    yb.conf.disable_excludes = ["all"]  # assume the openshift excluder will be managed, ignore current state
+    # determine what level of precision the user specified for the openshift version.
+    # should look like a version string with possibly many segments e.g. "3.4.1":
+    expected_version = module.params['version']
+    if not re.match(r'\A\d+(?:\.\d+)+\Z', expected_version):
+        bail('Requested release version "%s" does not look like a version.' % expected_version)
 
     # search for package versions available for aos pkgs
+    yb = yum.YumBase()  # pylint: disable=invalid-name
+    yb.conf.disable_excludes = ["all"]  # assume the openshift excluder will be managed, ignore current state
```
rhcarvalho: Do we need to talk about "openshift excluder" here? I fear it may cause more confusion than help.

roles/openshift_health_checker/library/aos_version.py
rhcarvalho: It's weird to rely on the "rpm prefix" for this. As we've seen, there are multiple prefixes that mean "enterprise", depending on the version.

If we want to explicitly disable this check for Origin, I'd suggest we use a mechanism like we use for containerized / non-containerized.

roles/openshift_health_checker/library/aos_version.py
sosiouxme: Well, disabling excludes is kind of a weird step to take, so I thought it needed _some_ explanation. And the reasoning is kind of weird too...

rhcarvalho: In all honesty, I do not understand the comment.

sosiouxme:
Alright, then it needs a better comment.

roles/openshift_health_checker/library/aos_version.py
sosiouxme: I don't really want to disable the whole check, just this part of it. I agree checking rpm prefix is a bit weird... will see if I can clean this up.

sosiouxme:
Rebased on https://github.com/openshift/openshift-ansible/pull/3816 so integration tests work.

I refactored the `aos_version` module. With `openshift_sanitize_inventory` in place we don't need the release normalization stuff in place, so I ripped that out. `aos_version` is hopefully much more pythonic and legible, and I wrote some unit tests. 

@rhcarvalho Those unit tests work fine if I run:

    $ py.test roles/openshift_health_checker/test/aos_version_test.py

Then they run fine. However under `tox -e py27-unit` it always fails with:

```
____________________________________________________________________ ERROR collecting roles/openshift_health_checker/test/aos_version_test.py _____________________________________________________________________
ImportError while importing test module '/home/lmeyer/go/src/github.com/openshift/openshift-ansible/roles/openshift_health_checker/test/aos_version_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
roles/openshift_health_checker/test/aos_version_test.py:2: in <module>
    import library.aos_version as aos_version
roles/openshift_health_checker/library/aos_version.py:23: in <module>
    import yum  # pylint: disable=import-error
E   ImportError: No module named yum
```

I'm sure there's something simple that's missing here but I can't make out what it is.

sosiouxme:
@rhcarvalho I removed the integration tests so this can move forward separately. Should be more review-able now if you find the time.

rhcarvalho:
> I'm sure there's something simple that's missing here but I can't make out what it is.

`tox` run the tests in a virtual env, and it doesn't share packages with the system (`yum` is only available in the system python).

Since `yum` is not available in PyPI, I don't know how to get it into the virtualenv other than making the env include system packages, which comes with its own disadvantages...

sosiouxme:
I suppose I could try/except the yum import and silently succeed when it
fails? Nothing in the unit tests actually relies on yum.

rhcarvalho:
Looks like you just removed a library/\_\_init\_\_.py? I was writing a comment to your question how to be able to import things from the library:

Suggestion: add `.../library` to `sys.path` (see `.../test/conftest.py`).

rhcarvalho:
> I suppose I could try/except the yum import and silently succeed when it
> fails? Nothing in the unit tests actually relies on yum.

Silently ignore might give us headache later on.

Idea: capture the import exception into a variable; check this variable in the beginning of the module execution, re-raising the exception if not None. In the tests we can patch the variable to always be None.

sosiouxme:
I rebased on your changes and found it already worked.​

sosiouxme:
Much better idea, will do.

playbooks/byo/openshift-preflight/lookup_plugins
```diff
@@ -0,0 +1 @@
+../../../lookup_plugins
```
rhcarvalho: Do we actually use those?

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -1,91 +1,189 @@
...
+    expected_pkgs = set([
+        rpm_prefix,
+        rpm_prefix + '-master',
+        rpm_prefix + '-node',
+    ])
 
-    rpm_prefix = module.params['prefix']
+    # determine what level of precision the user specified for the openshift version.
+    # should look like a version string with possibly many segments e.g. "3.4.1":
+    requested_openshift_release = str(module.params.get('requested_openshift_release', ''))
```
rhcarvalho: `'requested_openshift_release'` is guaranteed to be in `module.params`, I think the second argument to `.get` will never be used.

sosiouxme:
We do not directly.​ At some point in development I ran into a role that
used them, took a while to figure out what was going on, and I didn't want
to have to do that again. Could be nothing in this needs them now however
it's also future-proofing in case dependent roles start using them.

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -1,91 +1,189 @@
...
+    # likely be running outside either. When we attempt to determine what
+    # packages are available via yum they may be excluded. So, for our
+    # purposes here, disable excludes to see what will really be available
+    # during an install or upgrade.
+    yb.conf.disable_excludes = ['all']
+
     try:
         pkgs = yb.pkgSack.returnPackages(patterns=expected_pkgs)
-    except yum.Errors.PackageSackError as e:  # pylint: disable=invalid-name
+    except yum.Errors.PackageSackError as excinfo:  # pylint: disable=invalid-name
```
rhcarvalho: If we rename `e`, we can remove the `pylint` comment.

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -1,91 +1,189 @@
...
+    # should look like a version string with possibly many segments e.g. "3.4.1":
+    requested_openshift_release = str(module.params.get('requested_openshift_release', ''))
+
+    # get the list of packages available and complain if anything is wrong
+    try:
+        pkgs = _retrieve_available_packages(expected_pkgs)
+        if requested_openshift_release:
+            _check_precise_version_found(pkgs, expected_pkgs, requested_openshift_release)
+            _check_higher_version_found(pkgs, expected_pkgs, requested_openshift_release)
+        if module.params['openshift_deployment_type'] != 'origin':
```
rhcarvalho: Wouldn't it be better to whitelist? What could we expect from running the function below for an arbitrary deployment type?

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -1,91 +1,189 @@
...
 
-    rpm_prefix = module.params['prefix']
+    # determine what level of precision the user specified for the openshift version.
+    # should look like a version string with possibly many segments e.g. "3.4.1":
+    requested_openshift_release = str(module.params.get('requested_openshift_release', ''))
+
+    # get the list of packages available and complain if anything is wrong
+    try:
+        pkgs = _retrieve_available_packages(expected_pkgs)
+        if requested_openshift_release:
```
rhcarvalho: And what if `requested_openshift_release` is the empty string?!

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -1,91 +1,189 @@
...
     yb = yum.YumBase()  # pylint: disable=invalid-name
-    yb.conf.disable_excludes = ["all"]  # assume the openshift excluder will be managed, ignore current state
 
-    # search for package versions available for aos pkgs
-    expected_pkgs = [
-        rpm_prefix,
-        rpm_prefix + '-master',
-        rpm_prefix + '-node',
-    ]
+    # The openshift excluder prevents unintended updates to openshift
```
rhcarvalho: Is there a link describing what is an "openshift excluder" that we could reference here?

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -1,91 +1,189 @@
...
+    # packages by setting yum excludes on those packages. Excludes are
+    # then disabled during an install or upgrade, but this check will most
+    # likely be running outside either. When we attempt to determine what
+    # packages are available via yum they may be excluded. So, for our
+    # purposes here, disable excludes to see what will really be available
+    # during an install or upgrade.
```
rhcarvalho: What if by disabling excludes we "uncover" more than one version of packages? Is there a case that the check will report incorrect results whereas an installation would succeed properly?

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -1,91 +1,189 @@
...
+def _check_higher_version_found(pkgs, expected_pkgs, requested_openshift_release):
+    requested_openshift_release = [int(segment) for segment in requested_openshift_release.split(".")]
+    # see if any packages are available in a version higher than requested
+    higher_version_for_pkg = {}
+    for pkg in pkgs:
+        if pkg.name not in expected_pkgs:
+            continue
+        version = [int(segment) for segment in pkg.version.split(".")]
+        if (version[:len(requested_openshift_release)] > requested_openshift_release and
+                version > higher_version_for_pkg.get(pkg.name, [])):
```
rhcarvalho: Suggestion: give "names" to each condition and we don't have problems with indentation and long lines:

```python
cond1 = version[:len(requested_openshift_release)] > requested_openshift_release
cond2 = version > higher_version_for_pkg.get(pkg.name, [])

if cond1 and cond2:
    ...
```

Of course, give `cond1` and `cond2` proper names.

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -1,91 +1,189 @@
...
+            higher_version_for_pkg[pkg.name] = version
+
+    if higher_version_for_pkg:
+        higher_found = []
+        for name, version in higher_version_for_pkg.items():
+            higher_found.append(name + '-' + '.'.join([str(segment) for segment in version]))
```
rhcarvalho: Python tip: you can drop the square brackets. That means we use a generator expression instead of a list comprehension, and no intermediate list is created:

```python
'.'.join(str(segment) for segment in version)
```

Or

```python
'.'.join(map(str, version))
```
(Creates an intermediate list in Python 2 -- not a big deal; doesn't create an intermediate list in Python 3)

roles/openshift_health_checker/test/aos_version_test.py
```diff
@@ -0,0 +1,120 @@
+import pytest
+import library.aos_version as aos_version
```
rhcarvalho: This requires `library` to be a package, i.e, to have a `__init__.py` file...

Since to Ansible that's not seem as a package, but rather a bunch of modules/files, I'd go with manipulating `sys.path` and importing as:

```python
import aos_version
```

That's what we do throughout the repo for what I've seen.

roles/openshift_health_checker/test/aos_version_test.py
```diff
@@ -0,0 +1,120 @@
...
+        [Package('spam', '3.2.1'), Package('spam', '3.3.2')],
+        ['spam'],
+    ),
+    (
+        [Package('eggs', '1.2.3'), Package('eggs', '3.2.1.5'), Package('eggs', '3.4')],
+        ['eggs'],
+    ),
+])
+def test_check_pkgs_for_multi_release(pkgs, expect_multi):
+    if expect_multi:
```
rhcarvalho: What does `expect_multi` mean?

There's no assertion on the `else` branch, that's suspicious :male_detective:

roles/openshift_health_checker/test/aos_version_test.py
```diff
@@ -0,0 +1,120 @@
...
+        ['eggs-3.4'],  # multiple versions, two are higher
+    ),
+])
+def test_check_pkgs_for_greater_version(pkgs, requested_release, expect_higher):
+    if expect_higher:
+        with pytest.raises(aos_version.FoundHigherVersion) as e:
+            aos_version._check_higher_version_found(pkgs, expected_pkgs, requested_release)
+        assert set(expect_higher) == set(e.value.problem_pkgs)
+    else:
+        aos_version._check_higher_version_found(pkgs, expected_pkgs, requested_release)
```
rhcarvalho: no assertion == volkswagen test

roles/openshift_health_checker/test/aos_version_test.py
```diff
@@ -0,0 +1,120 @@
...
+        ['spam'],  # eggs found with multiple versions
+    ),
+])
+def test_check_pkgs_for_precise_version(pkgs, requested_release, expect_not_found):
+    if expect_not_found:
+        with pytest.raises(aos_version.PreciseVersionNotFound) as e:
+            aos_version._check_precise_version_found(pkgs, expected_pkgs, requested_release)
+        assert set(expect_not_found) == set(e.value.problem_pkgs)
+    else:
+        aos_version._check_precise_version_found(pkgs, expected_pkgs, requested_release)
```
rhcarvalho: no assertion == volkswagen test

playbooks/byo/openshift-preflight/lookup_plugins
rhcarvalho: > We do not directly.​ At some point in development I ran into a role that
used them, took a while to figure out what was going on, and I didn't want
to have to do that again. Could be nothing in this needs them now however
it's also future-proofing in case dependent roles start using them.

Hmm, if we don't understand well the problem, I'm afraid adding it could have the opposite effect, plus bring in perhaps unintended side-effects (like Ansible doing things we do not need nor anticipate).

Very likely whenever we run the checks Ansible would unnecessarily look at those dirs and load the plugins? I remember I couple of times I ran `strace ... ansible-playbook ...` to track down where ansible was looking for files.

I'd prefer to avoid the bloat over questionable "future-proofing" -- does such a thing even exist in practice? :D
I still remember when I read Ron Jeffries' book! https://en.wikipedia.org/wiki/YAGNI

roles/openshift_health_checker/library/aos_version.py
sosiouxme: Then the user didn't specify what release they want and we have no basis for telling them it won't work.

roles/openshift_health_checker/library/aos_version.py
sosiouxme: I would expect `openshift_sanitize_inventory` to kick out arbitrary types. I don't really expect types to proliferate, but I suppose whitelisting would make more sense.

roles/openshift_health_checker/test/aos_version_test.py
sosiouxme: Will do. I don't know why it's actually working now...

roles/openshift_health_checker/test/aos_version_test.py
sosiouxme: The assertion is that it runs without exception. Do you have a preferred way to indicate that?

http://stackoverflow.com/a/20275035/262768

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -1,91 +1,189 @@
...
+all the releases are provided in a single repo with the expectation that
+only the latest can be installed.
+
+Code in the openshift_version role contains a lot of logic to pin down
+the exact package and image version to use and so does some validation
+of release availability already. Without duplicating all that, we would
+like the user to have a helpful error message early on if things aren't
+going to work out right.
+
+TODO: fail gracefully on non-yum systems (dnf in Fedora)
```
sosiouxme: Definitely

playbooks/byo/openshift-preflight/lookup_plugins
sosiouxme: Fair enough.

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -1,91 +1,189 @@
...
 '''
 
 import yum  # pylint: disable=import-error
 
 from ansible.module_utils.basic import AnsibleModule
 
 
-def main():  # pylint: disable=missing-docstring,too-many-branches
+class AosVersionException(Exception):
+    '''Base exception class for package version problems'''
```
sosiouxme: You get two choices, either put in a fairly useless docstring, or put in this `# pylint` noise. I prefer a docstring that at least has the potential to provide information to a comment that looks like lint itself.

roles/openshift_health_checker/library/aos_version.py
sosiouxme: Not the way excludes are implemented currently; when enabled, they exclude the openshift packages entirely, not by version.

There's no denying the situation with excluders is a little messy, because during an install or upgrade the state changes. This should work and it's the least mess I can think of.

roles/openshift_health_checker/test/aos_version_test.py
sosiouxme: It's a list of packages for which you expect to see multiple releases. I'll try to improve the name...

roles/openshift_health_checker/library/aos_version.py
sosiouxme: This seems to be the best I can find:

https://wiki.centos.org/SpecialInterestGroup/PaaS/OpenShift-Origin-Control-Updates

Our docs don't say much about them, and may say less in the future as you're not really supposed to have to think about them. They're actually totally simple. It's just the use cases that are complex.

roles/openshift_health_checker/library/aos_version.py
rhcarvalho: I wanted missing-docstring disabled in the main config, but that needs agreement with the rest of the team. Too many config flags to bikeshed on pylint...

roles/openshift_health_checker/test/aos_version_test.py
rhcarvalho: Doesn't seem to be a great SO question/page...

What is testing that calling a function doesn't raise any exception really testing?!


```python
def test_foo():
    foo()   # yay, passes -- GREEN test!

def foo():
    pass
```

If `foo` is there just for side-effects, then it is more like a "procedure" not a "function"... It is easier to test functions by looking at their return values.

roles/openshift_health_checker/library/aos_version.py
rhcarvalho: Aha, so the expected behavior is "if no version is specified, this check succeeds", right? I'm good with that. Since we're already improving the wording in the module docstring, maybe we can specify this behavior there also?

roles/openshift_health_checker/test/aos_version_test.py
sosiouxme: As you say, it's not really a function but a procedure. Its purpose is to raise an exception and halt the check for some inputs and do nothing otherwise. The way this check runs, it just seems cleaner. You could instead have the functions return something on failure, check the return values, and introduce a bunch of if/then logic in the calling part.

roles/openshift_health_checker/library/aos_version.py
rhcarvalho: Thanks, that link was indeed helpful, at least now I feel like I understand it a bit better!

roles/openshift_health_checker/library/aos_version.py
sosiouxme: yep! sure.

roles/openshift_health_checker/test/aos_version_test.py
rhcarvalho: I see. Thanks for explaining, let's go with what we have.

roles/openshift_health_checker/library/aos_version.py
sosiouxme: Seems too little for a card though. And it would be strange to create an issue against code that's not merged yet. Should we have a tech debt card?

sosiouxme:
> Could you please summarize what's behind the code additions to
> aos_version.py -- what can it do now that it didn't before? Or if it is
> written somewhere just point me to it please :)
>
I refactored it mainly to be able to unit test it and to erase evidence of
my embarrassing python code :) As far as concrete improvements... it
handles the multi-version issue correctly for origin now, and expects the
origin repo enabled first in order to be able to check the pkg version. I
may have found a bug or two. It's not a lot but it's happier code...

sosiouxme:
aos-ci-test

sosiouxme:
[merge]

rhcarvalho:
flake? :cry:

sosiouxme:
I assume [test flake](https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/182/)

[merge] again.

sosiouxme:
Recorded apparent test flake at https://github.com/openshift/origin/issues/13705 - [merge] again?

rhcarvalho:
[merge]

rhcarvalho:
@team-project-committers @sdodson @stevekuznetsov 

As per https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_tox/7/console

The merge job is failing because it can't install Python dependencies... we are missing packages such as python-devel, most likely.

Can we actually test this sort of change in CI before affecting the whole organization? It basically messes up with a good deal of my day :(

stevekuznetsov:
re[merge]

sosiouxme:
@stevekuznetsov guessing there's still something up with ovs? https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_extended_conformance_install_update/17/

stevekuznetsov:
@sosiouxme yeah we should make an issue in origin repo to track this ... node is still racing with ovs to start 

```
 systemd[1]: Starting Open vSwitch Database Unit...
 systemd[1]: Starting Open vSwitch...
 systemd[1]: Started Open vSwitch.
 systemd[1]: Starting Origin Node...
 systemd[1]: origin-node.service: main process exited, code=exited, status=255/n/a
 systemd[1]: Failed to start Origin Node.
 systemd[1]: Unit origin-node.service entered failed state.
 systemd[1]: origin-node.service failed.
 systemd[1]: Started Open vSwitch Database Unit.
```

sosiouxme:
Created test flake issue https://github.com/openshift/origin/issues/13713

So let's try another [merge]

PR #3650, opened 2017-03-14
Add router svcacct cluster-reader role

mtnbikenc:
Fixes 1431583

### DIFF
--- a/roles/openshift_hosted/tasks/router/router.yml
+++ b/roles/openshift_hosted/tasks/router/router.yml
@@ -37,6 +37,15 @@
     resource_name: hostnetwork
   with_items: "{{ openshift_hosted_routers }}"
 
+- name: Set additional permissions for router service account
+  oc_adm_policy_user:
+    user: "system:serviceaccount:{{ item.namespace }}:{{ item.serviceaccount }}"
+    namespace: "{{ item.namespace }}"
+    resource_kind: cluster-role
+    resource_name: cluster-reader
+  when: item.namespace == 'default'
+  with_items: "{{ openshift_hosted_routers }}"
+
 - name: Create OpenShift router
   oc_adm_router:
     name: "{{ item.name }}"


roles/openshift_hosted/tasks/router/router.yml
```diff
@@ -37,6 +37,15 @@
     resource_name: hostnetwork
   with_items: "{{ openshift_hosted_routers }}"
 
+- name: Set additional permissions for router service account
+  oc_adm_policy_user:
+    user: "system:serviceaccount:{{ item.namespace }}:{{ item.serviceaccount }}"
+    namespace: "{{ item.namespace }}"
+    resource_kind: cluster-role
+    resource_name: cluster-reader
+  when: item.namespace == 'default'
```
mtnbikenc: The only thing I'm not sure about is whether this needs to be run for all shards, or only on the 'default' namesapce.

kwoodson: @mtnbikenc, good question here.  How we run it today is with the same service account and in the same namespace so a single call would probably suffice.

sdodson:
aos-ci-test

sdodson:
[merge]

mtnbikenc:
@sdodson should I be expecting this testing error?  I've not changed anything in that area.  Or do I need to rebase?
```
No package matching 'origin-docker-excluder' found available, installed or updated
```

sdodson:
@mtnbikenc I think the origin excluder is broken and I've opened a PR to fix it. I think we can also work around it in openshift-ansible too.

sdodson:
@mtnbikenc can you pick this into release-1.5 too?

smarterclayton:
Whoa whoa whoa - why is the router a cluster-reader?

smarterclayton:
@liggitt

kwoodson:
@smarterclayton, I think it is in response to this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1431583

mtnbikenc:
@smarterclayton The above bug linked to this this bug comment which provided the guidance. https://bugzilla.redhat.com/show_bug.cgi?id=1332510#c1  Please correct me if there is newer guidance.

sdodson:
Looks like we're also granding cluster-reader to nuage, logging, metrics, and manageiq integrations.

smarterclayton:
We shouldn't given cluster reader to those - that's way too much permission.  We should be giving out a much more restricted role.  It's not good to give infrastructure components cluster-reader by default (maybe manageiq), because that makes those components have way more privileges than we want.

In general, if we need a generic role that lets some data be accessible, we should describe it as a new role - not as giving them cluster-reader.

Can we get the various folks involved in those components @jcantrill @mwringe @simon3z @knobunc to describes what permissions they actually need? 

This is a potential security issue, so I'd like to get resolution on steps forward very shortly.  In the meantime, we need to hold off on merging the backport of this.

mwringe:
For the metrics components, its only Heapster that currently requests cluster-reader permissions. I believe it needs cluster reader permissions because that was the only way it could access /stats and summary information from the nodes.

Heapster also needs to be able to read pod definitions from each namespace, but that can be granted using a role with a lower level of authorisation.

liggitt:
> For the metrics components, its only Heapster that currently requests cluster-reader permissions. I believe it needs cluster reader permissions because that was the only way it could access /stats and summary information from the nodes.

`system:node-reader` gives read access to node API objects, metrics, stats, and spec info from the nodes

kwoodson:
This is referring to the router.  Is there a lesser permission we should be using to grant the router access to read namespaces?

knobunc:
All the router needs is read-namespace so it can read the labels on the namespace to see if they match the subset the router is looking for.  And that permission is only needed when the router is sharded (i.e. when the env NAMESPACE_LABELS is set to a label selector).

jcantrill:
The fluentd component of logging uses cluster-reader to get pod spec info (e.g. annotations, labels) for enriching log messages.

mtnbikenc:
@knobunc Could you point me to where `read-namespace` is documented?  I'm trying to follow-up on implementation and want to make sure I'm doing it right.

knobunc:
@mtnbikenc sorry Russell.  I mean that generically.  It needs to be able to read the namespace objects.  I'm not sure if we have something else we can grant yet if they want to do sharding.

knobunc:
@mtnbikenc Here's the doc where we tell them to grant the cluster-reader permission: https://docs.openshift.com/enterprise/3.2/install_config/install/deploy_router.html#creating-the-router-service-account

@smarterclayton is there a finer-graind role we can use?  Or are you advocating that we make a new one?

PR #3654, opened 2017-03-14
Do not update properties when they are not passed in.

kwoodson:


kwoodson:
aos-ci-test

kwoodson:
[merge]

mtnbikenc:
@detiber @rhcarvalho @sdodson Coverage decreased by 0.008%.  I think this is an example of the challenges in getting code merged.  In this example, with these code changes, what is the expected path forward?  And where do we have these coverage expectations documented?

rhcarvalho:
@mtnbikenc thanks for bring this up.

Yeah, that doesn't seems very reasonable, right? That was my first impression... like "what? really? 0.008%?!"...

I took a deep breath, and went to look at coveralls (disclaimer: I never used it before)...
https://coveralls.io/builds/10611944/source?filename=roles%2Flib_openshift%2Flibrary%2Foc_project.py#L1528

It actually tells us that there are 4 new uncovered lines introduced by this PR.
I searched for a suspicious chain of ifs introduced in this PR and found it marked as uncovered, the link above.

So while it seems pedantic, in this particular case it is revealing what seems to be unnecessary ifs that are untested. I'll comment on the relevant lines of code with a suggestion.

rhcarvalho:
A quick general look at the code and commit message suggests that the new code tries to handle a new scenario, and there are no accompanying tests for that scenario.

mtnbikenc:
aos-ci-test

detiber:
> @detiber @rhcarvalho @sdodson Coverage decreased by 0.008%. I think this is an example of the challenges in getting code merged. In this example, with these code changes, what is the expected path forward? And where do we have these coverage expectations documented?

The main goal I had with using coveralls was to give us a better idea of how we were trending and more unified visualization of the coverage information. I purposely didn't make coveralls a required test, since I do not think it's overly reasonable at this point to require everything to provide full test coverage for changes made. I wanted to try and start a conversation of defining what types of coverage requirements we wanted to require as a team to get us away from the very limited coverage requirements that we have in place in the tox-based testing (coverage reports and requirements are limited to a single tox environment, so we weren't seeing combined coverage from different test invocations or different python versions).

I think it'd be a good idea to discuss what is acceptable for coverage requirements at the next architecture meeting. Coveralls has the ability to specify a certain percentage of lost coverage before failing, which if there can be agreement on what is an allowable percentage would solve this immediate problem.

@sdodson @rhcarvalho another solution that could help is to potentially use codecov instead of coveralls, which has some compelling features that would allow setting different coverage requirements for code changed/added with a PR vs the existing code base.

sdodson:
I'm happy to have trending. I think coveralls is fine for now unless the team suggests a better solution.

kwoodson:
aos-ci-test

kwoodson:
[test]

kwoodson:
[test]

kwoodson:
[test]

kwoodson:
[test]

roles/lib_openshift/src/test/unit/test_oc_project.py
```diff
@@ -108,3 +109,172 @@ def test_adding_a_project(self, mock_cmd, mock_write, mock_tmpfile_copy, mock_lo
...
+        self.assertFalse(results['changed'])
+
+        # Making sure our mock was called as we expected
+        mock_cmd.assert_has_calls([
+            mock.call(['oc', 'get', 'namespace', 'operations', '-o', 'json'], None),
+        ])
+
+    @mock.patch('oc_project.locate_oc_binary')
+    @mock.patch('oc_project.Utils.create_tmpfile_copy')
+    @mock.patch('oc_project.Utils._write')
```
rhcarvalho: As I did in https://github.com/openshift/openshift-ansible/pull/3677, you could patch `oc_project.Utils` as a whole in one shot.

https://github.com/openshift/openshift-ansible/pull/3677/files#diff-2746a5dd3c12cb661bdc37e4dd89d053R112

Might also be helpful to use the `spec=SomeClass` to limit the mock object's API.

mtnbikenc:
aos-ci-test

rhcarvalho:
Could probably squash the first 4 commits into one? They seem to attempts to fix and refix the same things?

mtnbikenc:
@rhcarvalho Squashing is something I've thought about bringing up to the team.  I don't care for all the fix commits because it makes looking back in history a bit more difficult to see what was actually being done.  My preference is that we'd squash before merge, but that would also mean that all the tests would have to re-run, and then we'd have to wait all over again after finally getting it working.  Or we squash each time before pushing, but then it gets tricky to see what was changed after the last time you reviewed the PR.  Sometimes it's also nice to have individual commits for specific chunks of work which could be individually picked or rolled back.  I'm sure everyone else has thoughts on pros/cons, so I'll see about getting this put on an agenda somewhere.

cc: @sdodson

sdodson:
Yes, unless there's actual value in each individual commit and each commit will produce a working install on its own we should be squashing. We've gotten lax about this while dashing towards the finish line but we should be squashing out bad commits.

kwoodson:
I have a fix for the py35 tests here: https://github.com/openshift/openshift-ansible/pull/3683

I will rebase once it is merged.

kwoodson:
aos-ci-test [test]

mtnbikenc:
aos-ci-test

kwoodson:
@rhcarvalho, @stevekuznetsov, This seems like another flake unrelated to this PR.
`18:13:58     "msg": "Unable to start service origin-node: Job for origin-node.service failed because the control process exited with error code. See \"systemctl status origin-node.service\" and \"journalctl -xe\" for details.\n"`

Any ideas?

sdodson:
[merge]

sdodson:
flake https://github.com/openshift/origin/issues/13386
[merge]

PR #3656, opened 2017-03-14
Refactor openshift_projects

mtnbikenc:
The openshift_projects role has been refactored down to one task.  That one task has been moved into the openshift_hosted role since that was the only role calling the projects role.  However, It may be better to move this task somewhere else, given that openshift_hosted will be eventually split into _registry and _router roles.


detiber:
@mtnbikenc There is potential that other components could make use of a common method of creating and managing projects. @ewolinetz would it make sense to share project management tasks for logging and metrics?

There are also the efforts to deploy integrated projects/projects such as ManageIQ/CloudForms, GlusterFS, etc which would have similar requirements.

mtnbikenc:
@detiber In those instances where projects need to be created, would it not make sense to just call the `oc_project` module to create them as needed, instead of calling a role that has one task?  Additionally, that one task would need to essentially default all the options of the `oc_project` module, an interface to an interface.  The logic in the original role was limited to idempotently creating a project, which is handled within the module.

mtnbikenc:
aos-ci-test
[test]

mtnbikenc:
This PR depends on https://github.com/openshift/openshift-ansible/pull/3654 being merged.

mtnbikenc:
aos-ci-test
[test]

mtnbikenc:
aos-ci-test

mtnbikenc:
[merge]

PR #3657, opened 2017-03-14
Openshift logging & metrics bzs

ewolinetz:
Addresses:
  https://bugzilla.redhat.com/show_bug.cgi?id=1430627
  https://bugzilla.redhat.com/show_bug.cgi?id=1430626
  https://bugzilla.redhat.com/show_bug.cgi?id=1430628

### DIFF
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -530,8 +530,8 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # list of options please see roles/openshift_logging/README.md
 #
 # Configure loggingPublicURL in the master config for aggregate logging, defaults
-# to https://kibana.{{ openshift_master_default_subdomain }}
-#openshift_hosted_logging_hostname=https://kibana.example.com
+# to kibana.{{ openshift.common.dns_domain }}
+#openshift_hosted_logging_hostname=kibana.example.com
 # Configure the number of elastic search nodes, unless you're using dynamic provisioning
 # this value must be 1
 #openshift_hosted_logging_elasticsearch_cluster_size=1
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -530,7 +530,7 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # list of options please see roles/openshift_logging/README.md
 #
 # Configure loggingPublicURL in the master config for aggregate logging, defaults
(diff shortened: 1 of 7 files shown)


ewolinetz:
This likely needs a cherry-pick into master too...

roles/openshift_logging/tasks/generate_pvcs.yaml
```diff
@@ -1,4 +1,8 @@
 ---
+- debug: var=es_pvc_size
+- debug: var=es_cluster_size
+- debug: var=es_dc_names
```
sdodson: intentionally left in?

inventory/byo/hosts.origin.example
```diff
@@ -530,8 +530,8 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
 # list of options please see roles/openshift_logging/README.md
 #
 # Configure loggingPublicURL in the master config for aggregate logging, defaults
-# to https://kibana.{{ openshift_master_default_subdomain }}
-#openshift_hosted_logging_hostname=https://kibana.example.com
+# to kibana.{{ openshift.common.dns_domain }}
```
sdodson: hmm, dns_domain is usually 'cluster.local' so internal dns not public dns. I think you want `kibana.{{ openshift_master_default_subdomain }}`

sdodson: About this we probably need to fix https://github.com/openshift/openshift-ansible/blame/master/roles/openshift_logging/defaults/main.yml#L25 too

sdodson: https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/openshift_hosted.yml#L23 is the default logic used elsewhere

roles/openshift_logging/tasks/generate_pvcs.yaml
ewolinetz: nope, thanks for catching that...

inventory/byo/hosts.origin.example
ewolinetz: that is what we use in the openshift_logging role though. 
from defaults/main.yaml
`openshift_logging_kibana_hostname: "{{ openshift_hosted_logging_hostname | default('kibana.{{openshift.common.dns_domain}}') }}"`

Its in the case where they are using the playbooks/byo/openshift-cluster/openshift-logging.yaml playbook. That may not be set

ewolinetz:
@sdodson should we be failing/setting a default value of `openshift_master_default_subdomain` then within the `openshift_logging` role?

PR #3620, opened 2017-03-10
Enable docker during installation and upgrade by default

ingvagabund:
If the docker-excluder is installed and enabled by a user, it must by enabled even during installation, resp. upgrade. If a user unexclude the excluder, the gets excluded at the end of the installation/upgrade.

Bug 1430612

PR #3658, opened 2017-03-14
Revert "Enable docker during installation and upgrade by default"

stevekuznetsov:
Reverts openshift/openshift-ansible#3620


sdodson:
@ingvagabund FYI, we reverted this because it broke the origin merge queue because the origin excluders are broken. https://github.com/openshift/origin/pull/13387

ingvagabund:
We can skip the origin and deploy excluders only for openshit-atomic until the issue is resolved for origin.

stevekuznetsov:
@ingvagabund you'll also want to work with @jhadvig to make sure this won't break the upgrade scenario -- install cluster using 3.4 installer, upgrade using master HEAD

PR #3659, opened 2017-03-14
Revert "Revert "Enable docker during installation and upgrade by default""

stevekuznetsov:
Reverts openshift/openshift-ansible#3658

aos-ci-test

[test]


sdodson:
I think we still want this, but will let @ingvagabund open a new pr if necessary.

PR #3660, opened 2017-03-14
add check to ensure OpenShift image data does not exceed size limit in etcd

juanvallejo:
Adds check to make sure that the total size of OpenShift image data does not exceed a defined limit in an etcd cluster by measuring leaf node values under `/openshift.io/images/...` and `/openshift.io/imagestreams/...`. Size limit for image data defaults to 50% of total space used by an etcd host.

~Differs from https://github.com/openshift/openshift-ansible/pull/3630 as that PR checks the size of the volume in which the etcd cluster is running. I can merge both PRs into one, however if they are similar enough. Thought it would be easier, for now, to review these two checks as separate PRs, however.~

Depends on the ansible host having the [python-etcd](https://pypi.python.org/pypi/python-etcd) package installed.

TODO

- [x] add tests

### DIFF
--- a/roles/openshift_health_checker/library/etcdkeysize.py
+++ b/roles/openshift_health_checker/library/etcdkeysize.py
@@ -1,0 +1,122 @@
+#!/usr/bin/python
+"""Ansible module that recursively determines if the size of a key in an etcd cluster exceeds a given limit."""
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+try:
+    import etcd
+
+    IMPORT_EXCEPTION_MSG = None
+except ImportError as err:
+    IMPORT_EXCEPTION_MSG = str(err)
+
+    from collections import namedtuple
+    EtcdMock = namedtuple("etcd", ["EtcdKeyNotFound"])
+    etcd = EtcdMock(KeyError)
+
+
+# pylint: disable=too-many-arguments
+def check_etcd_key_size(client, key, size_limit, total_size=0, depth=0, depth_limit=1000, visited=None):
+    """Check size of an etcd path starting at given key. Returns tuple (string, bool)"""
+    if visited is None:
+        visited = set()
+
+    if key in visited:
+        return 0, False
+
+    visited.add(key)
+
+    try:
+        result = client.read(key, recursive=False)
+    except etcd.EtcdKeyNotFound:
+        return 0, False
+
+    size = 0
+    limit_exceeded = False
+
+    for node in result.leaves:
+        if depth >= depth_limit:
+            raise Exception("Maximum recursive stack depth ({}) exceeded.".format(depth_limit))
+
+        if size_limit and total_size + size > size_limit:
+            return size, True
+
+        if not node.dir:
+            size += len(node.value)
+            continue
+
+        key_size, limit_exceeded = check_etcd_key_size(client, node.key,
+                                                       size_limit,
+                                                       total_size + size,
+                                                       depth + 1,
+                                                       depth_limit, visited)
+        size += key_size
+
+    max_limit_exceeded = limit_exceeded or (total_size + size > size_limit)
+    return size, max_limit_exceeded
+
+
+def main():  # pylint: disable=missing-docstring,too-many-branches
+    module = AnsibleModule(
+        argument_spec=dict(
+            size_limit_bytes=dict(type="int", default=0),
+            paths=dict(type="list", default=["/openshift.io/images"]),
+            host=dict(type="str", default="127.0.0.1"),
+            port=dict(type="int", default=4001),
+            protocol=dict(type="str", default="http"),
+            version_prefix=dict(type="str", default=""),
+            allow_redirect=dict(type="bool", default=False),
+            cert=dict(type="dict", default=""),
+            ca_cert=dict(type="str", default=None),
+        ),
+        supports_check_mode=True
+    )
+
+    module.params["cert"] = (
+        module.params["cert"]["cert"],
+        module.params["cert"]["key"],
+    )
+
+    size_limit = module.params.pop("size_limit_bytes")
+    paths = module.params.pop("paths")
+
+    limit_exceeded = False
+
+    try:
+        # pylint: disable=no-member
+        client = etcd.Client(**module.params)
+    except AttributeError as attrerr:
+        msg = str(attrerr)
+        if IMPORT_EXCEPTION_MSG:
+            msg = IMPORT_EXCEPTION_MSG
+            if "No module named etcd" in IMPORT_EXCEPTION_MSG:
+                # pylint: disable=redefined-variable-type
+                msg = ('Unable to import the python "etcd" dependency. '
+                       'Make sure python-etcd is installed on the host.')
+
+        module.exit_json(
+            failed=True,
+            changed=False,
+            size_limit_exceeded=limit_exceeded,
+            msg=msg,
+        )
+
+        return
+
+    size = 0
+    for path in paths:
+        path_size, limit_exceeded = check_etcd_key_size(client, path, size_limit - size)
+        size += path_size
+
+        if limit_exceeded:
+            break
+
+    module.exit_json(
+        changed=False,
+        size_limit_exceeded=limit_exceeded,
+    )
+
+
+if __name__ == '__main__':
+    main()
--- a/roles/openshift_health_checker/openshift_checks/etcd_imagedata_size.py
+++ b/roles/openshift_health_checker/openshift_checks/etcd_imagedata_size.py
@@ -1,0 +1,84 @@
+"""
+Ansible module for determining if the size of OpenShift image data exceeds a specified limit in an etcd cluster.
+"""
+
+from openshift_checks import OpenShiftCheck, OpenShiftCheckException, get_var
+
+
+class EtcdImageDataSize(OpenShiftCheck):
+    """Check that total size of OpenShift image data does not exceed the recommended limit in an etcd cluster"""
+
+    name = "etcd_imagedata_size"
+    tags = ["etcd"]
+
+    def run(self, tmp, task_vars):
+        etcd_mountpath = self._get_etcd_mountpath(get_var(task_vars, "ansible_mounts"))
+        etcd_avail_diskspace = etcd_mountpath["size_available"]
+        etcd_total_diskspace = etcd_mountpath["size_total"]
+
+        etcd_imagedata_size_limit = get_var(task_vars,
+                                            "etcd_max_image_data_size_bytes",
+                                            default=int(0.5 * float(etcd_total_diskspace - etcd_avail_diskspace)))
+
+        etcd_is_ssl = get_var(task_vars, "openshift", "master", "etcd_use_ssl", default=False)
+        etcd_port = get_var(task_vars, "openshift", "master", "etcd_port", default=2379)
+        etcd_hosts = get_var(task_vars, "openshift", "master", "etcd_hosts")
+
+        config_base = get_var(task_vars, "openshift", "common", "config_base")
+
+        cert = task_vars.get("etcd_client_cert", config_base + "/master/master.etcd-client.crt")
+        key = task_vars.get("etcd_client_key", config_base + "/master/master.etcd-client.key")
+        ca_cert = task_vars.get("etcd_client_ca_cert", config_base + "/master/master.etcd-ca.crt")
+
+        for etcd_host in list(etcd_hosts):
+            args = {
+                "size_limit_bytes": etcd_imagedata_size_limit,
+                "paths": ["/openshift.io/images", "/openshift.io/imagestreams"],
+                "host": etcd_host,
+                "port": etcd_port,
+                "protocol": "https" if etcd_is_ssl else "http",
+                "version_prefix": "/v2",
+                "allow_redirect": True,
+                "ca_cert": ca_cert,
+                "cert": {
+                    "cert": cert,
+                    "key": key,
+                },
+            }
+
+            etcdkeysize = self.module_executor("etcdkeysize", args, task_vars)
+
+            if etcdkeysize.get("rc", 0) != 0 or etcdkeysize.get("failed"):
+                msg = 'Failed to retrieve stats for etcd host "{host}": {reason}'
+                reason = etcdkeysize.get("msg")
+                if etcdkeysize.get("module_stderr"):
+                    reason = etcdkeysize["module_stderr"]
+
+                msg = msg.format(host=etcd_host, reason=reason)
+                return {"failed": True, "changed": False, "msg": msg}
+
+            if etcdkeysize["size_limit_exceeded"]:
+                limit = self._to_gigabytes(etcd_imagedata_size_limit)
+                msg = ("The size of OpenShift image data stored in etcd host "
+                       "\"{host}\" exceeds the maximum recommended limit of {limit:.2f} GB. "
+                       "Use the `oadm prune images` command to cleanup unused Docker images.")
+                return {"failed": True, "msg": msg.format(host=etcd_host, limit=limit)}
+
+        return {"changed": False}
+
+    @staticmethod
+    def _get_etcd_mountpath(ansible_mounts):
+        valid_etcd_mount_paths = ["/var/lib/etcd", "/var/lib", "/var", "/"]
+
+        mount_for_path = {mnt.get("mount"): mnt for mnt in ansible_mounts}
+        for path in valid_etcd_mount_paths:
+            if path in mount_for_path:
+                return mount_for_path[path]
+
+        paths = ', '.join(sorted(mount_for_path)) or 'none'
+        msg = "Unable to determine a valid etcd mountpath. Paths mounted: {}.".format(paths)
+        raise OpenShiftCheckException(msg)
+
+    @staticmethod
+    def _to_gigabytes(byte_size):
+        return float(byte_size) / 10.0**9
--- a/roles/openshift_health_checker/test/etcd_imagedata_size_test.py
+++ b/roles/openshift_health_checker/test/etcd_imagedata_size_test.py
@@ -1,0 +1,328 @@
+import pytest
+
+from collections import namedtuple
+from openshift_checks.etcd_imagedata_size import EtcdImageDataSize, OpenShiftCheckException
+from etcdkeysize import check_etcd_key_size
+
+
+def fake_etcd_client(root):
+    fake_nodes = dict()
+    fake_etcd_node(root, fake_nodes)
+
+    clientclass = namedtuple("client", ["read"])
+    return clientclass(lambda key, recursive: fake_etcd_result(fake_nodes[key]))
+
+
+def fake_etcd_result(fake_node):
+    resultclass = namedtuple("result", ["leaves"])
+    if not fake_node.dir:
+        return resultclass([fake_node])
+
+    return resultclass(fake_node.leaves)
+
+
+def fake_etcd_node(node, visited):
+    min_req_fields = ["dir", "key"]
+    fields = list(node)
+    leaves = []
+
+    if node["dir"] and node.get("leaves"):
+        for leaf in node["leaves"]:
+            leaves.append(fake_etcd_node(leaf, visited))
+
+    if len(set(min_req_fields) - set(fields)) > 0:
+        raise ValueError("fake etcd nodes require at least {} fields.".format(min_req_fields))
+
+    if node.get("leaves"):
+        node["leaves"] = leaves
+
+    nodeclass = namedtuple("node", fields)
+    nodeinst = nodeclass(**node)
+    visited[nodeinst.key] = nodeinst
+
+    return nodeinst
+
+
+@pytest.mark.parametrize('ansible_mounts,extra_words', [
+    ([], ['none']),  # empty ansible_mounts
+    ([{'mount': '/mnt'}], ['/mnt']),  # missing relevant mount paths
+])
+def test_cannot_determine_available_mountpath(ansible_mounts, extra_words):
+    task_vars = dict(
+        ansible_mounts=ansible_mounts,
+    )
+    check = EtcdImageDataSize(execute_module=fake_execute_module)
+
+    with pytest.raises(OpenShiftCheckException) as excinfo:
+        check.run(tmp=None, task_vars=task_vars)
+
+    for word in 'determine valid etcd mountpath'.split() + extra_words:
+        assert word in str(excinfo.value)
+
+
+@pytest.mark.parametrize('ansible_mounts,tree,size_limit,should_fail,extra_words', [
+    (
+        # test that default image size limit evals to 1/2 * (total size in use)
+        [{
+            'mount': '/',
+            'size_available': 40 * 10**9,
+            'size_total': 80 * 10**9,
+        }],
+        {"dir": False, "key": "/", "value": "1234"},
+        None,
+        False,
+        [],
+    ),
+    (
+        [{
+            'mount': '/',
+            'size_available': 40 * 10**9,
+            'size_total': 48 * 10**9,
+        }],
+        {"dir": False, "key": "/", "value": "1234"},
+        None,
+        False,
+        [],
+    ),
+    (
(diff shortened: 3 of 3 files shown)


juanvallejo:
@ncdc I ended up basing this PR on [etcdstats](https://github.com/ncdc/etcdstats), although a slight difference in this implementation is that I am retrieving all non-dir etcd leaf nodes in a [single request](https://github.com/openshift/openshift-ansible/pull/3660/files#diff-bb47d47e77d2a02ef94208bc5897ab12R33), rather than making [multiple requests](https://github.com/ncdc/etcdstats/blob/master/etcdstats.go#L161) to obtain them. 

Although a single request could be problematic when dealing with fairly large etcd clusters, I was hoping to get your input on how well measuring the size of a cluster through multiple requests would scale.

Any feedback or thoughts on this are appreciated. Thanks

cc @openshift/scale-performance-review

ncdc:
@juanvallejo I would expect multiple requests would scale better (use less memory).

Why did you hardcode the default max size to 700mb?

jeremyeder:
@ncdc is this binary and python-etcd going to be packaged with openshift?

ncdc:
@jeremyeder this PR doesn't use my etcdstats binary. It only uses python-etcd. I'll defer to @abutcher @sdodson etc on what the best approach is here.

juanvallejo:
@ncdc 
> Why did you hardcode the default max size to 700mb?

This was just something to test the implementation, ideally the limit would be defined by the user

sosiouxme:
Just so I'm clear... this ansible module runs on the host, pulls down the entire local etcd contents and measures them?

juanvallejo:
@sosiouxme 
> Just so I'm clear... this ansible module runs on the host, pulls down the entire local etcd contents and measures them?

Yes, although the better approach (and also what his [etcdstats](https://github.com/ncdc/etcdstats/blob/master/etcdstats.go#L161) util does) would be to measure the value of each non-dir key one request per key at a time

rhcarvalho:
> Yes, although the better approach would be to measure the value of each non-dir key one request per key at a time

[thinking aloud] hmm, depending on the write rate and how long this operation to compute the size out take, we can get a skewed result [/thinking aloud]


---

Still couldn't find anything in the docs about how to get the current storage size used, but found that

> To prevent performance degradation or unintentionally overloading the key-value store, etcd enforces a 2GB default storage size quota, configurable up to 8GB. To avoid swapping or running out of memory, the machine should have at least as much RAM to cover the quota.
https://github.com/coreos/etcd/blob/master/Documentation/faq.md#system-requirements

Why can't we just look at the storage device on the multiple nodes of the cluster?!

juanvallejo:
@rhcarvalho 
> Why can't we just look at the storage device on the multiple nodes of the cluster?!

That still won't guarantee that we are only measuring etcd object data. Doing `df -h /var/lib/etcd` on an ec2 instance gives me:

```
Filesystem      Size  Used Avail Use% Mounted on
/dev/xvda2       25G  7.1G   18G  29% /
```

but measuring the size of my etcd cluster with `etcdstats` gives (in bytes):

```
Filesystem     1K-blocks    Used Available Use% Mounted on
/dev/xvda2      26202092 7407396  18794696  29% /
```

Also, directly measuring `/var/lib/etcd` will not work either, as I would be measuring snapshots, etc.:

```
$ du -h /var/lib/etcd

37M     /var/lib/etcd/member/snap
306M    /var/lib/etcd/member/wal
342M    /var/lib/etcd/member
342M    /var/lib/etcd
```

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,64 @@
...
+'''
+Ansible module to determine the size of an etcd cluster
+'''
+
+import etcd
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def get_size(client, key, seen=None):
```
stevekuznetsov: Had this conversation in person but wanted to post it here for broader audience ... asking the database to recursively list all keys will not only generate a huge amount of IO but also blow out the caches it's got set up, not to mention that asking to do a full recursive list on a production etcd close to the limit may cripple it. 

I'm not sure this is a super useful check to be doing with a client like this. If the etcd project has a maximum safe storage size, they should expose a server endpoint for monitoring this.

juanvallejo:
@stevekuznetsov thanks for the feedback. I have updated this check to only focus on measuring OpenShift image data stored in etcd.

roles/openshift_health_checker/openshift_checks/etcd_imagedata_size.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def run(self, tmp, task_vars):
+        ansible_mounts = get_var(task_vars, "ansible_mounts")
+
+        etcd_mountpath = self.get_etcd_mountpath(ansible_mounts)
+        etcd_avail_diskspace = etcd_mountpath["size_available"]
+        etcd_total_diskspace = etcd_mountpath["size_total"]
+
+        etcd_imagedata_size_limit = get_var(task_vars,
+                                            "cli_etcd_max_image_data_size_bytes",
```
juanvallejo: @rhcarvalho per the discussion in https://github.com/openshift/openshift-ansible/commit/d60ac18a64c4fd33e3a7c8d0fb3ce94f79dfa742#commitcomment-21856259, I have updated this to be a bit more intuitive, however will still try to get feedback for a final variable name

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,93 @@
...
+'''
+Ansible module to determine the size of an etcd cluster up to a specified limit
+'''
+
+import etcd
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def get_size(client, key, total=0, limit=0, depth=0, depth_limit=1000, visited=None):
```
juanvallejo: @rhcarvalho added a depth limit here that defaults to `1000`, I cannot foresee a stable etcd setup having any of its subtrees with a bigger height than this, but feel free to add your thoughts

enj: So `depth=0, depth_limit=1000` is pretty pointless here.  We control the etcd keyspace, specifically we control exactly where objects are stored.  I wrote the most complicated logic for storing an object in etcd, which had a depth of `5`, so lets just pretend the absurd worst case is `10`.  Thus python will never even come close to a stackoverflow.  Therefore all this depth checking is just adding pointless complexity for no benefits.

juanvallejo:
@rhcarvalho will add tests if no further comments, thanks

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,105 @@
...
+        supports_check_mode=True
+    )
+
+    module.params["cert"] = (
+        module.params["cert"]["cert"],
+        module.params["cert"]["key"],
+    )
+
+    size_limit = module.params["size_limit_bytes"]
+    module.params.pop("size_limit_bytes", None)
```
sosiouxme: Equivalent to:

    size_limit = module.params.pop("size_limit_bytes")
￼
I'd like all these transform-module-params lines to be in one visual block...

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,105 @@
...
+
+    limit_exceeded = False
+
+    paths = module.params["paths"]
+    module.params.pop("paths", None)
```
sosiouxme: same deal here

roles/openshift_health_checker/openshift_checks/etcd_imagedata_size.py
sosiouxme: Guess it should be `r_openshift_health_checker_etcd_max_image_data_size_bytes`

Which something in the role should set from `openshift_health_etcd_image_data_bytes`

Whee...

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/python
+# vim: expandtab:tabstop=4:shiftwidth=4
```
sosiouxme: take this out

roles/openshift_health_checker/openshift_checks/etcd_imagedata_size.py
```diff
@@ -0,0 +1,84 @@
...
+    def run(self, tmp, task_vars):
+        etcd_mountpath = self.get_etcd_mountpath(get_var(task_vars, "ansible_mounts"))
+        etcd_avail_diskspace = etcd_mountpath["size_available"]
+        etcd_total_diskspace = etcd_mountpath["size_total"]
+
+        etcd_imagedata_size_limit = get_var(task_vars,
+                                            "cli_etcd_max_image_data_size_bytes",
+                                            default=int(0.5 * float(etcd_total_diskspace - etcd_avail_diskspace)))
+
+        etcd_is_ssl = get_var(task_vars, "openshift", "master", "etcd_use_ssl", default=False)
```
sosiouxme: and technically every last one of these would be translated to an `r_openshift_health_checker_` var too

roles/openshift_health_checker/openshift_checks/etcd_imagedata_size.py
```diff
@@ -0,0 +1,84 @@
...
+        etcd_port = get_var(task_vars, "openshift", "master", "etcd_port", default=2379)
+        etcd_hosts = get_var(task_vars, "openshift", "master", "etcd_hosts")
+
+        config_base = get_var(task_vars, "openshift", "common", "config_base")
+
+        cert = task_vars.get("etcd_client_cert", "".join([config_base, "/master/master.etcd-client.crt"]))
```
sosiouxme: Surely this:

        cert = task_vars.get("etcd_client_cert", config_base + "/master/master.etcd-client.crt")

is clearer than a join? Is there some Python practice I'm missing here...

roles/openshift_health_checker/openshift_checks/etcd_imagedata_size.py
```diff
@@ -0,0 +1,84 @@
...
+                if etcdstats.get("module_stderr"):
+                    reason = etcdstats["module_stderr"]
+
+                msg = msg.format(host=etcd_host, reason=reason)
+                return {"failed": True, "changed": False, "msg": msg}
+
+            if etcdstats["size_limit_exceeded"]:
+                limit = self.to_gigabytes(etcd_imagedata_size_limit)
+                msg = ("The size of OpenShift image data stored in etcd host "
+                       "\"{host}\" exceeds the maximum recommended limit of {limit} GB")
```
sosiouxme: Do we have some advice we'd like them to follow? E.g. image pruning...

roles/openshift_health_checker/openshift_checks/etcd_imagedata_size.py
```diff
@@ -0,0 +1,84 @@
...
+                "ca_cert": ca_cert,
+                "cert": {
+                    "cert": cert,
+                    "key": key,
+                },
+            }
+
+            etcdstats = self.module_executor("etcdstats", args, task_vars)
+
+            if etcdstats.get("rc", 0) > 0:
```
sosiouxme: Could the rc be negative?... how about `!= 0` ?

stevekuznetsov: > Could the rc be negative?

Sooo......

Technically RC is always 0-255. However today I saw Ansible reporting an RC of -11 for a real RC of 139. So no idea WTF is going on inside of Ansible somewhere but ... you may want to check that they're following spec before trusting that the field "rc" will be >= 0.

roles/openshift_health_checker/openshift_checks/etcd_imagedata_size.py
juanvallejo: Will suggest image pruning

roles/openshift_health_checker/openshift_checks/etcd_imagedata_size.py
juanvallejo: So, I agree with prefixing our check vars with this, however, in this case, all of the `etcd_..` task_vars used are valid and exist under "ansible_facts.openshift.master" or "openshift.common"

juanvallejo:

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,121 @@
...
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+
+IMPORT_EXCEPTION = None
+
+try:
+    import etcd
+# pylint: disable=invalid-name
+except ImportError as IMPORT_EXCEPTION:
```
rhcarvalho: Hmm, `IMPORT_EXCEPTION` is invalid here, right?

ALL_CAPS is used for constants, and this is not precisely one...

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,121 @@
...
+    # Allows for test mocking
+    from collections import namedtuple
+    EtcdMock = namedtuple("etcd", ["EtcdKeyNotFound"])
+    etcd = EtcdMock(KeyError)
+
+
+# pylint: disable=too-many-arguments
+def check_etcd_key_size(client, key, total_size=0, size_limit=0, depth=0, depth_limit=1000, visited=None):
+    """Recursively checks that total size of an etcd path starting at a given key does not exceed a given limit"""
```
rhcarvalho: https://www.python.org/dev/peps/pep-0257/#one-line-docstrings:
> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

> (...) The preferred form for such a docstring would be something like:
>
> def function(a, b):
>     """Do X and return a list."""
> (Of course "Do X" should be replaced by a useful description!)

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,121 @@
...
+            continue
+
+        key_size, limit_exceeded = check_etcd_key_size(client, node.key,
+                                                       total_size + size,
+                                                       size_limit,
+                                                       depth + 1,
+                                                       depth_limit, visited)
+        size += key_size
+
+    return size, limit_exceeded or (total_size + size > size_limit)
```
rhcarvalho: This is a confusing line.
Suggestion: give a name to the return value (create a variable), then `return size, name_of_bool_var`.

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,121 @@
...
+    limit_exceeded = False
+
+    try:
+        # pylint: disable=no-member
+        client = etcd.Client(**module.params)
+    except AttributeError as err:
+        if IMPORT_EXCEPTION:
+            err = IMPORT_EXCEPTION
+            if "No module named etcd" in IMPORT_EXCEPTION:
+                # pylint: disable=redefined-variable-type
```
rhcarvalho: Hmm, this looks like a legit problem.

`err` should consistently be either a string or an exception, not sometimes one sometimes the other...

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,121 @@
...
+                err = ("Unable to import the python \"etcd\" dependency. "
```
rhcarvalho: If we use single quotes to delimit the string, we wouldn't need to escape the double quotes.

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,121 @@
...
+                       "Make sure python-etcd is installed on the host.")
+
+        module.exit_json(
+            changed=False,
+            size_limit_exceeded=limit_exceeded,
+            rc=1,
```
rhcarvalho: Why does this return `rc=1`?!
And why not `failed=True`?

roles/openshift_health_checker/test/etcd_imagedata_size_test.py
```diff
@@ -0,0 +1,335 @@
...
+])
+def test_check_etcd_key_size_calculates_correct_limit(ansible_mounts, tree, size_limit, failed, extra_words):
+    def execute_module(module_name, args, tmp=None, task_vars=None):
+        if module_name != "etcdstats":
+            return {
+                "changed": False,
+            }
+
+        client = fake_etcd_client(tree)
+        s, limit_exceeded = check_etcd_key_size(client, tree["key"], size_limit=args["size_limit_bytes"])
```
rhcarvalho: The arguments a caller is supposed to pass in should be the first ones, so that we don't need to pass `size_limit` as a named arg. We can order the args more conveniently.

roles/openshift_health_checker/test/etcd_imagedata_size_test.py
```diff
@@ -0,0 +1,335 @@
...
+
+        return {"size_limit_exceeded": limit_exceeded}
+
+    task_vars = dict(
+        r_openshift_health_checker_etcd_max_image_data_size_bytes=size_limit,
```
rhcarvalho: @sosiouxme I still don't see how this will work out...

We'll need to convert some other variable into `r_openshift_health_checker_etcd_max_image_data_size_bytes`. Do we want to have different variable names depending on where the variable is (supposed to be) declared?
It would be cumbersome to rename variables and do manual priority resolution (from the multiple possible ways to define the variable, which one has more priority)... that feels like a poor form solution.

How are your thoughts on this so far?

roles/openshift_health_checker/test/etcd_imagedata_size_test.py
```diff
@@ -0,0 +1,335 @@
...
+            master=dict(etcd_hosts=["localhost"]),
+            common=dict(config_base="/var/lib/origin")
+        )
+    )
+    if size_limit is None:
+        task_vars.pop("r_openshift_health_checker_etcd_max_image_data_size_bytes")
+
+    check = EtcdImageDataSize(execute_module=execute_module).run(tmp=None, task_vars=task_vars)
+
+    if failed:
```
rhcarvalho: Naming... how about "should_fail"? At first look the current name got me confused.

Perhaps even more clear if we have all cases that "should fail" in one place, and ones that "should not fail" in another test.

We could extract the common parts into a function, so that tests would have a shape like:

```python
@pytest.mark.parametrize(... list of cases that should not fail)
def test_name(...):
    result = somefunc(...)
    assert not result.get('failed', False)

@pytest.mark.parametrize(... list of cases that should fail)
def test_name(...):
    result = somefunc(...)
    assert result.get('failed', False)
    for word in words:
         assert word in result['msg']

def somefunc(...):
    ...
````

File: roles/openshift_health_checker/test/etcd_imagedata_size_test.py [hunk 20]
```diff
@@ -0,0 +1,335 @@
...
+    check = EtcdImageDataSize(execute_module=execute_module).run(tmp=None, task_vars=task_vars)
+    assert not check.get("failed", False)
+
+
+@pytest.mark.parametrize('rc,err_msg,extra_words', [
+    (
+        1,
+        "failure",
+        ["Failed to retrieve stats", "failure"]
+    )
```
rhcarvalho: Why parametrize if there is only one set of params?!

roles/openshift_health_checker/test/etcd_imagedata_size_test.py
```diff
@@ -0,0 +1,335 @@
...
+            ]
+        },
+        "/",
+        16,
+        [],
+    ),
+])
+def test_etcd_key_size_check_calculates_correct_size(ansible_mounts, tree, root_path, expected_size, extra_words):
+    def execute_module(module_name, args, tmp=None, task_vars=None):
+        if module_name != "etcdstats":
```
rhcarvalho: We should probably update this name to something more appropriate, it is not collecting "etcd stats" for some time already :-)

roles/openshift_health_checker/library/etcdstats.py
```diff
@@ -0,0 +1,121 @@
...
+
+from ansible.module_utils.basic import AnsibleModule
+
+IMPORT_EXCEPTION = None
+
+try:
+    import etcd
+# pylint: disable=invalid-name
+except ImportError as IMPORT_EXCEPTION:
+    # Allows for test mocking
```
juanvallejo: I thought it might be good to leave a reason here explaining why we catch an import exception, since it is something that is not really done in any other check

juanvallejo:
@rhcarvalho thanks, review comments addressed.

roles/openshift_health_checker/library/etcdstats.py
rhcarvalho: The problem is not the comment, but the assignment of etcd to a fake in this block. We should only construct fakes in the testing code. If production code is calling this module and the import fails, the module execution should fail with an appropriate error message.

juanvallejo:
@rhcarvalho 
> The problem is not the comment, but the assignment of etcd to a fake in this block. We should only construct fakes in the testing code. If production code is calling this module and the import fails, the module execution should fail with an appropriate error message.

While I agree that this should be the case, in order to have tests for this module without requiring the control host to have `python-etcd` installed, we can't have this import fail

juanvallejo:
Updated all task vars to use the internal role variable names. See https://github.com/openshift/openshift-ansible/pull/3630/files#diff-abe713fff903d16dd5b984239abde97f

roles/openshift_health_checker/test/etcd_imagedata_size_test.py
sosiouxme: I have come to think that the whole idea of role variables as parameters scoped to the role won't work out. If we were very strict about it then every role include would have a raft of parameters due to needing to supply everything that dependencies need as well. It would be unmanageable.

juanvallejo:
@sosiouxme wondering if there are any more comments on this patch? Thanks!

sosiouxme:
aos-ci-test

roles/openshift_health_checker/library/etcdkeysize.py
```diff
@@ -0,0 +1,122 @@
...
+
+    try:
+        # pylint: disable=no-member
+        client = etcd.Client(**module.params)
+    except AttributeError as attrerr:
+        msg = str(attrerr)
+        if IMPORT_EXCEPTION_MSG:
+            msg = IMPORT_EXCEPTION_MSG
+            if "No module named etcd" in IMPORT_EXCEPTION_MSG:
+                # pylint: disable=redefined-variable-type
```
sosiouxme: think you don't need this anymore...

roles/openshift_health_checker/library/etcdkeysize.py
```diff
@@ -0,0 +1,122 @@
...
+                                                       total_size + size,
+                                                       depth + 1,
+                                                       depth_limit, visited)
+        size += key_size
+
+    max_limit_exceeded = limit_exceeded or (total_size + size > size_limit)
+    return size, max_limit_exceeded
+
+
+def main():  # pylint: disable=missing-docstring,too-many-branches
```
sosiouxme: doubt it has too many branches now and... give it a crappy docstring?

roles/openshift_health_checker/library/etcdkeysize.py
```diff
@@ -0,0 +1,122 @@
...
+try:
+    import etcd
+
+    IMPORT_EXCEPTION_MSG = None
+except ImportError as err:
+    IMPORT_EXCEPTION_MSG = str(err)
+
+    from collections import namedtuple
+    EtcdMock = namedtuple("etcd", ["EtcdKeyNotFound"])
+    etcd = EtcdMock(KeyError)
```
sosiouxme: rodolfo didn't like this.

sosiouxme:
fatal: [10.8.168.166]: FAILED! => {"changed": false, "failed": true, "msg": "OCP rpm version 3.5.5.17 is different from OCP image version 3.5.5.18"}

sosiouxme:
aos-ci-test

sosiouxme:
[merge]

PR #3666, opened 2017-03-15
don't assume openshift_upgrade_target is in a form d.d

ingvagabund:
Just in case the ``openshift_upgrade_target`` is set to something like ` 1.5.0-999.alpha.3+1eb44c7.379.x86_64`


ingvagabund:
aos-ci-test

jhadvig:
Checked the change and if I disable the excluder manually then it works

sdodson:
[merge]

PR #3669, opened 2017-03-15
Removing dependency on master facts for master_public_url default

ewolinetz:
To address issues in origin-aggregated-logging CI where we see 
```
AnsibleUndefinedVariable: {{ openshift_hosted_logging_master_public_url | default('https://' + openshift.common.public_hostname + ':' + openshift.master.api_port) }}: 'dict object' has no attribute 'master'
```

@jcantrill @richm

### DIFF
--- a/roles/openshift_logging/defaults/main.yml
+++ b/roles/openshift_logging/defaults/main.yml
@@ -3,7 +3,7 @@ openshift_logging_image_prefix: "{{ openshift_hosted_logging_deployer_prefix | d
 openshift_logging_image_version: "{{ openshift_hosted_logging_deployer_version | default('latest') }}"
 openshift_logging_use_ops: "{{ openshift_hosted_logging_enable_ops_cluster | default('false') | bool }}"
 openshift_logging_master_url: "https://kubernetes.default.svc.{{ openshift.common.dns_domain }}"
-openshift_logging_master_public_url: "{{ openshift_hosted_logging_master_public_url | default('https://' + openshift.common.public_hostname + ':' + openshift.master.api_port) }}"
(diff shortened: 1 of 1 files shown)


jcantrill:
My only concern is if this mitigates the QE issue we had where the port was different from the master and not being propagated correctly.  Otherwise LGTM

ewolinetz:
@jcantrill we may need to update the inventory as part of this PR then

For reference, BZ in question: https://bugzilla.redhat.com/show_bug.cgi?id=1425312

jcantrill:
@ewolinetz Are you suggesting users may need to manually update their inventory?

richm:
There are two use cases 
* initial install of OpenShift + logging at the same time - installing OpenShift, and in the same inventory, setting both `openshift_hosted_logging_deploy=true` and `openshift_logging_install_logging=true` so that logging will also be deployed - in this case, how should the default value of `openshift_logging_master_public_url` be derived?
* running ansible again, and getting facts via openshift_facts and openshift_master_facts - in this case, which fact values should be used to construct `openshift_logging_master_public_url`

ewolinetz:
@richm we shouldn't need to set both 'openshift_logging_install_logging=true' if we are setting 'openshift_hosted_logging_deploy=true' since the default for the former is already 'true'.

@jcantrill sorry, i was referring to the example inventory... in that bz I feel like QE was setting the incorrect variable to properly change the master api url. IMO the fix is that their inventory should have `openshift_hosted_logging_master_public_url=https://$master:443` or `openshift_logging_master_public_url=https://$master:443` to correct set the public api.  Since they are running playbooks/byo/openshift-cluster/openshift-logging.yaml `public_master_url` wouldn't impact anything from the cluster.

sdodson:
aos-ci-test

ewolinetz:
[merge]

ewolinetz:
lets try to [merge] again shall we

sdodson:
manually merging.

PR #3672, opened 2017-03-15
Set auto_expand_replicas to Default of 'false'

jcantrill:
to avoid https://github.com/elastic/elasticsearch/issues/1873.

This will require users to explicitly configure the option and avoid OpenShift clusters unnecessarily wasting resources based on what is described in the issue.


jcantrill:
cc @portante

jcantrill:
Additional Ref: https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#dynamic-index-settings

ewolinetz:
this is a duplicate of https://github.com/openshift/openshift-ansible/pull/3580

PR #3673, opened 2017-03-15
Remove openshift_excluders status tasks

sdodson:
Since we no longer care about the current status and instead just install/enable/disable based on enable_excluders variables we can remove calls to status which were attempting to upgrade packages that were not installed and call the status function on non existent binaries.

Previously with enable_excluders=false we were installing them via the status.yml
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1432345


sdodson:
@jhadvig please try this

ingvagabund:
In that case, we can remove all `enable_docker_excluder_override` and `disable_openshift_excluder_override` occurrences as well.

sdodson:
Ok, do you think this makes sense in general though?

ingvagabund:
I was wondering a while ago if we really need the status at this point and I don't think we do.

ingvagabund:
The remaining variables can be removed later,. They are set in the status only anyway. If not set, they are set to some default value.

sdodson:
aos-ci-test

sdodson:
[merge]

PR #3674, opened 2017-03-15
Fix containerized openvswitch race

sdodson:
We wait until openvswitch is started, but openvswitch doesn't wait until ovsdb-server.service has started so we must do that on our own.

See https://github.com/openshift/origin/issues/13318


sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
[merge]

PR #3677, opened 2017-03-16
Do not monkey-patch unnecessary objects

rhcarvalho:
Cleanup patching of unnecessary / fine-grained objects in tests.

Unit tests that know too much about the implementation details of a certain code being tested makes for code that is hard to refactor. We should be able to change implementation details without needing to change tests.

---
Previous purpose:

This shows a path forward to testing previously untested code when making changes to existing files, allowing us to gradually improve code coverage.

Note: I'm leaving implementing meaningful tests as an exercise to another maintainer.

This PR contains the commits from #3654 squashed into one, plus my PoC in a new commit.

### DIFF
--- a/roles/lib_openshift/src/test/unit/test_oc_project.py
+++ b/roles/lib_openshift/src/test/unit/test_oc_project.py
@@ -37,11 +37,9 @@ class OCProjectTest(unittest.TestCase):
         'description': 'All things operations project',
     }
 
-    @mock.patch('oc_project.locate_oc_binary')
-    @mock.patch('oc_project.Utils.create_tmpfile_copy')
-    @mock.patch('oc_project.Utils._write')
+    @mock.patch('oc_project.Utils', autospec=True)
     @mock.patch('oc_project.OCProject._run')
-    def test_adding_a_project(self, mock_cmd, mock_write, mock_tmpfile_copy, mock_loc_oc_bin):
+    def test_adding_a_project(self, mock_cmd, mock_utils):
         ''' Testing adding a project '''
 
         params = copy.deepcopy(OCProjectTest.params)
@@ -83,14 +81,6 @@ class OCProjectTest(unittest.TestCase):
             (0, project_results, ''),  # fetch it
         ]
 
-        mock_tmpfile_copy.side_effect = [
-            '/tmp/mocked_kubeconfig',
-        ]
-
-        mock_loc_oc_bin.side_effect = [
-            'oc',
-        ]
-
         # Act
         results = OCProject.run_ansible(params, False)
 
@@ -110,11 +100,9 @@ class OCProjectTest(unittest.TestCase):
 
         ])
 
-    @mock.patch('oc_project.locate_oc_binary')
-    @mock.patch('oc_project.Utils.create_tmpfile_copy')
-    @mock.patch('oc_project.Utils._write')
+    @mock.patch('oc_project.Utils', autospec=True)
     @mock.patch('oc_project.OCProject._run')
-    def test_modifying_a_project_no_attributes(self, mock_cmd, mock_write, mock_tmpfile_copy, mock_loc_oc_bin):
+    def test_modifying_a_project_no_attributes(self, mock_cmd, mock_utils):
         ''' Testing adding a project '''
         params = copy.deepcopy(self.params)
         params['display_name'] = None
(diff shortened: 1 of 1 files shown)


rhcarvalho:
Repurposing PR as a follow up to #3654.

rhcarvalho:
aos-ci-test

rhcarvalho:
@kwoodson, is this okay for you?

kwoodson:
@rhcarvalho, Could you explain how this improves the test cases?  I'm fine with the approach.  That being said could you explain how using autospec=True and removing the `side_effects` of the locate oc binary calls improves tests?

rhcarvalho:
> Could you explain how this improves the test cases? I'm fine with the approach. That being said could you explain how using autospec=True and removing the side_effects of the locate oc binary calls improves tests?

Sure.

1. Autospeccing

https://docs.python.org/3/library/unittest.mock.html#auto-speccing
"Autospeccing [...] limits the api of mocks to the api of an original object (the spec), [...] so that attributes of mocks only have the same api as the attributes of the spec. In addition mocked functions / methods have the same call signature as the original so they raise a TypeError if they are called incorrectly."

See the discussion in the link with an explanation of problems this solves.

2. Locate oc binary

Hmm, may I invert the question, could you explain why it was added? It seems it is not needed -- I removed it and tests still pass.

I went on a bit further now, and see that even without mocking it is returning `"oc"`, which might be a bug: if the binary is not found in any path in PATH, the function returns `"oc"`. The expected behavior is undocumented AFAIK.

kwoodson:
@rhcarvalho, at some point the TC were raising exceptions as the `oc` binary was not found and it may not be found locally as its not installed.  The point of the mock was to return an `oc` string to make it look like it was in the path.

If it is returning `oc` by default I would tend to believe that it is installed or the locate function is finding it somewhere.  If it is not installed, not in path, and is still returning `oc`, then I'd argue that its a bug.  I didn't author that code but I'm willing to dive in and fix it.

I think that mocking all of the commands that are essential to producing a _good_ test are needed.  If you feel that finding the `oc` command is not needed, I'm fine with that as all tests now flow through that path.  If removing more mocks makes things easier to test, gives us the _same_ coverage, and is more flexible I'm a big fan.

rhcarvalho:
@kwoodson in some way, the more things we mock, the less things we're really testing :)
(because we're monkey-patching them with the mocks)

Testing after-the-fact is hard in the sense that it's so tempting to add in things that seem to be useful, or copy-paste from other places, without finding what is actually necessary. I believe it also often leads to a "no tests" -> "green tests" flow, which doesn't tell us much about the quality of the tests.

As an example of that, one can refer to the discussion I linked above in the mock framework docs... you could have typos in the assertion methods -- or forget to make assertions even, and you'd have a lot of test files, green test run, code coverage and... nothing really being tested ;-)

In automated testing, the most exciting and truth telling transition is when we see a "failed test" -> "green test" by adding a piece of code, because then we say that it is that piece of code and nothing else that is sufficient and required for the test to pass.

kwoodson:
@rhcarvalho, I went and reread the issue https://github.com/openshift/openshift-ansible/issues/3410 that I created and Jason's solution.  I think it was intentional to return `'oc'` as a string from the locate_oc_binary call.  I'm unsure if the test was in place before or after or which implementation of the location of the binary was in place at the time.  Its a fair point to re-evaluate the tests being written each time we run them.

We can remove the oc binary mocking from our tests as it is working as intended with the string `'oc'` and that can be assumed in our tests cases.

rhcarvalho:
@kwoodson, I'm closing this PR.
It has a risk of introducing test failures depending on the environment:

```
$ touch ~/bin/oc
$ tox -e py27-unit -- roles/lib_openshift/src/test/unit/test_oc_env.py
...
E       AssertionError: Calls not found.
E       Expected: [call(['oc', 'get', 'dc', 'router', '-o', 'json', '-n', 'default'], None)]
E       Actual: [call(['/home/rodolfo/bin/oc', 'get', 'dc', 'router', '-o', 'json', '-n', 'default'], None)]
...
```

Better to take a more holistic approach to removing side-effects and simplifying tests, but that's a separate effort.

roles/lib_openshift/src/test/unit/test_oc_project.py
```diff
@@ -37,11 +37,9 @@ class OCProjectTest(unittest.TestCase):
         'description': 'All things operations project',
     }
 
-    @mock.patch('oc_project.locate_oc_binary')
-    @mock.patch('oc_project.Utils.create_tmpfile_copy')
-    @mock.patch('oc_project.Utils._write')
+    @mock.patch('oc_project.Utils', autospec=True)
```
rhcarvalho: @kwoodson, speccing and autospeccing is still a good thing, we can look into adding it to new code during code reviews.

PR #3461, opened 2017-02-23
add docker_image_availability check

juanvallejo:
Something to get a rough idea of how this check might be implemented.

cc @rhcarvalho @brenton

PR #3680, opened 2017-03-16
Misc fixes

rhcarvalho:
Some cleans ups when I touched this area for #3677.

### DIFF
--- a/roles/lib_openshift/library/oc_project.py
+++ b/roles/lib_openshift/library/oc_project.py
@@ -1547,9 +1547,9 @@ class OCProject(OpenShiftCLI):
     def run_ansible(params, check_mode):
         '''run the idempotent ansible code'''
 
-        _ns = None
+        node_selector = None
         if params['node_selector'] is not None:
-            _ns = ','.join(params['node_selector'])
+            node_selector = ','.join(params['node_selector'])
 
         pconfig = ProjectConfig(
             params['name'],
@@ -1560,7 +1560,7 @@ class OCProject(OpenShiftCLI):
                 'admin_role': {'value': params['admin_role'], 'include': True},
                 'description': {'value': params['description'], 'include': True},
                 'display_name': {'value': params['display_name'], 'include': True},
-                'node_selector': {'value': _ns, 'include': True},
+                'node_selector': {'value': node_selector, 'include': True},
             },
         )
 
--- a/roles/lib_openshift/library/oc_project.py
+++ b/roles/lib_openshift/library/oc_project.py
@@ -1551,15 +1551,18 @@ class OCProject(OpenShiftCLI):
         if params['node_selector'] is not None:
             _ns = ','.join(params['node_selector'])
 
-        pconfig = ProjectConfig(params['name'],
-                                'None',
-                                params['kubeconfig'],
-                                {'admin': {'value': params['admin'], 'include': True},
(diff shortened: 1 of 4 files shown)


rhcarvalho:
aos-ci-test

rhcarvalho:
aos-ci-test

rhcarvalho:
@kwoodson PTAL

rhcarvalho:
@kwoodson okay if I merge this?

roles/lib_openshift/library/oc_project.py
```diff
@@ -1543,19 +1543,20 @@ def needs_update(self):
...
+            'None',
+            params['kubeconfig'],
+            {
+                'admin': {'value': params['admin'], 'include': True},
+                'admin_role': {'value': params['admin_role'], 'include': True},
+                'description': {'value': params['description'], 'include': True},
+                'display_name': {'value': params['display_name'], 'include': True},
+                'node_selector': {'value': node_selector, 'include': True},
+            },
+        )
```
rhcarvalho: Maybe one day we'll have something like https://github.com/google/yapf (or any alternative) in place and stop talking about formatting...

I'm not actively looking for code and reformatting, but since I touched this file and simplified the expression above, there it goes.

I explained the reasoning in another PR I was reviewing recently: https://github.com/openshift/openshift-ansible/pull/3461#discussion_r106629152
Or, in case the comment link stops working, http://blog.rodolfocarvalho.net/2017/03/indentation-styles.html.

kwoodson: I'm fine with the formatting as long as its readable and this is.

roles/lib_openshift/library/oc_project.py
```diff
@@ -1543,19 +1543,20 @@ def needs_update(self):
...
-        pconfig = ProjectConfig(params['name'],
-                                'None',
-                                params['kubeconfig'],
-                                {'admin': {'value': params['admin'], 'include': True},
-                                 'admin_role': {'value': params['admin_role'], 'include': True},
-                                 'description': {'value': params['description'], 'include': True},
-                                 'display_name': {'value': params['display_name'], 'include': True},
-                                 'node_selector': {'value': _ns, 'include': True},
-                                })
+        node_selector = ','.join(params.get('node_selector', [])) or None
```
kwoodson: I'm not sure this code will work the same way.  I like the shorter notation but the `params` dictionary will always have a `'node_selector'`.  It will either be `None` or a `list` of `key=value` strings.  I believe this will fail when `None` is not iterable.

`>>> params = {'node_selector': None}`
`>>> node_selector = ','.join(params.get('node_selector', [])) or None`
`Traceback (most recent call last):`
 ` File "<stdin>", line 1, in <module>`
`TypeError: can only join an iterable`

rhcarvalho: You're right, thanks for catching it. I've updated the patch.

I saw this gets used later on in `ProjectConfig.stringify` and there we only test the truthiness of the value, whereas in `OCProject.needs_update` we test whether the value is `None`.

Let's step back and keep the original 3 lines to avoid introducing any unintended bug.

rhcarvalho:
aos-ci-test

rhcarvalho:
Thanks @kwoodson 

[merge]

PR #3681, opened 2017-03-16
enable excluders during node/master scaling up

ingvagabund:
Bug 1432868


sdodson:
CI jobs won't touch these code paths.

sdodson:
Fixed include paths, tested just fine.

PR #3683, opened 2017-03-16
Fixing variable naming for 35 scoping.

kwoodson:


kwoodson:
Tested locally on py35 with  `tox -e py35-unit`.

results:
`========================================================== 602 passed, 52 skipped in 9.89 seconds ===========================================================`

ashcrow:
Testing quickly but it looks good.

kwoodson:
[merge]

kwoodson:
aos-ci-test

kwoodson:
@stevekuznetsov, @sdodson, is there something I can do to speed up the tests on this?

ashcrow:
@kwoodson I'm not authoritative on this but I don't believe so :frowning:

stevekuznetsov:
What do you mean by "speed up tests"? The same set of tests will be run on this PR as on any other, so it will likely take the same amount of time as usual

ashcrow:
@stevekuznetsov I believe @kwoodson means if there is a way to set this PR's priority higher so it would get merged faster.

sdodson:
pre-empting the merge queue as there are other prs that depend on this.

stevekuznetsov:
No, our current system doesn't have any concept of priority other than the age of the code in the PR -- if it's older, you get priority as you did the work first. We are looking into how we could move to use other tools like the k8s Prow system, but at the moment there are no committed cards for that.

PR #3687, opened 2017-03-16
Turn off console logging; add file logging

portante:


ewolinetz:
I think it should be configurable which a user gets. 
We had BZs opened because customers did not want logs contained on the pod. Do we have a recommended best practice when it comes to how we should handle pod logs?

There is also the separate issue of a feedback loop within ES receiving its own logs which would be addressed by this, but once the pod disappears how would we retrieve its logs? The advantage to simply writing to console is that Docker can then pick up the logs and make them available off-pod

portante:
> I think it should be configurable which a user gets.

Do we really want to enable a configuration option that can cripple the Elasticsearch instance?  Does not seem like an option we want to support.

> We had BZs opened because customers did not want logs contained on the pod.
> Do we have a recommended best practice when it comes to how we should handle pod logs?

Moving the logs to the persistent storage volume, and purging old ones periodically (which would be a good option to have for how long to keep logs) would be fine.

> There is also the separate issue of a feedback loop within ES receiving its own logs which would be
> addressed by this

That is the main focus, actually.  The secondary consideration is that at a few customer sites now, debugging 3.4 and later deployed Elasticsearch is quite difficult we because logs for Elasticsearch are in itself.  So when problems occur, even if we don't hit the feedback loop, we are stuck.  This actually has hidden a problem with Searchguard for a while now.

> but once the pod disappears how would we retrieve its logs?

Using the persistent volume already in place ...

> The advantage to simply writing to console is that Docker can then pick up the logs and make them
> available off-pod

If we can find a way to make sure Docker does not make those logs available to fluentd, and stores them persistently for a configurable period of time, that would be an acceptable solutions (albeit more complex).

> Also, if we aren't writing to standard out, I don't think we'd see anything if we try to 'oc logs' for the ES pod
> which breaks a lot of current debugging steps

That is okay, we just need to document those changes and why we are changing it.

jcantrill:
I think its acceptable we don't allow users to config via ansible; they can manually update the configmap.  And ultimately we do not want to the feedback loop.

My preference is to modify fluentd to ignore ES logs and maybe revisit adding a config option then.  Also, at some point maybe ansible can tweak the logging options and how they are emitted based on if an ops cluster exists.

oc logs being broken is less then desirable but I think if we have a persistent volume, proper documentation, and the capacity to turn on stdout via the configmap, it seems worth the change.

roles/openshift_logging/files/elasticsearch-logging.yml
```diff
@@ -28,32 +39,24 @@ logger:
 additivity:
   index.search.slowlog: false
   index.indexing.slowlog: false
+  deprecation: false
 
 appender:
-  console:
```
jcantrill: I think we should leave the appender  so we can turn on console logging if we have to.  And if we can make it possible to modify fluentd to ignore in future

ewolinetz:
> Do we really want to enable a configuration option that can cripple the Elasticsearch instance? Does not seem like an option we want to support.

How would it cripple the instance if its just writing to console?  I think given the alternative is a PV that would need to be configured in addition to the storage PV, users should have the option to use either configuration. It should be simple enough for them to pick between given the logic would be encompassed all within a role.

> That is the main focus, actually. The secondary consideration is that at a few customer sites now, debugging 3.4 and later deployed Elasticsearch is quite difficult we because logs for Elasticsearch are in itself. So when problems occur, even if we don't hit the feedback loop, we are stuck. This actually has hidden a problem with Searchguard for a while now.

I think we should also update Fluentd to ignore all logs from the logging namespace. If we want to avoid loop-back from ES wouldn't we also want to avoid loop-back from the other components as well? (Esp Kibana I would think...)

roles/openshift_logging/files/elasticsearch-logging.yml
portante: Why can't we add an ansible script that modifies the configmap to add this back?

portante:
> My preference is to modify `fluentd` to ignore ES logs and maybe revisit adding a config option then.

This means complication.  Instead of being sure we don't have a feedback look for Elasticsearch, while also ensuring we have local logs for debugging, we defer the prevention of the feedback loop to a `fluentd`.  If you get it wrong there, we have a potential problem.

I recommend we do NOT pursue that root.

> Also, at some point maybe ansible can tweak the logging options and how they are emitted based on if an ops cluster exists.

Here is a [diagram](https://www.lucidchart.com/invitations/accept/de866bab-54d7-426d-95c3-a349ee036788) that discusses what we have today, and where we want to go and why, showing how to break the cycle.

> oc logs being broken is less then desirable but I think if we have a persistent volume, proper documentation, and the capacity to turn on stdout via the configmap, it seems worth the change.

If we can prevent the pod from logging to `journald`, where we would need to filter at the `fluentd` pod, e.g. using `json-file` option for Docker for those pods, that would be acceptable.

But again, moving this filtering to `fluentd` is prone to error.

portante:
@ewolinetz ...

>> Do we really want to enable a configuration option that can cripple the Elasticsearch instance? Does not seem like an option we want to support.

> How would it cripple the instance if its just writing to console?

In the current system, output from the console of Elasticsearch goes back into Elasticsearch, which is the feedback loop.  That can lead to a crippled system.

If we are talking about breaking the loop, but still having console output of Elasticsearch configured, then we need to make sure where we break the loop is supportable.  I don't think filtering logs at `fluentd` is the supportable way to do that.

> I think given the alternative is a PV that would need to be configured in addition to the storage PV, users should have the option to use either configuration. It should be simple enough for them to pick between given the logic would be encompassed all within a role.

Sure, though more options is more for us to verify and support.

>> That is the main focus, actually. The secondary consideration is that at a few customer sites now, debugging 3.4 and later deployed Elasticsearch is quite difficult we because logs for Elasticsearch are in itself. So when problems occur, even if we don't hit the feedback loop, we are stuck. This actually has hidden a problem with Searchguard for a while now.

> I think we should also update Fluentd to ignore all logs from the logging namespace. If we want to avoid loop-back from ES wouldn't we also want to avoid loop-back from the other components as well? (Esp Kibana I would think...)

The feedback loop is only with Elasticsearch and fluentd logs.  Curator and Kibana are not able to generate a feedback loop because they don't log based on incoming logs from apps.

jcantrill:
@ewolinetz This changes seems reasonable to me given the condition we are trying to avoid.  We can always advise users to turn console logging back on if we truly require it.  There should be an additional documentation PR based on: https://github.com/openshift/openshift-docs/pull/3918 which notes the change.  Also, we further need to add a PR(s) to make the file log to a persistent volume that is rotated.

jcantrill:
@portante @ewolinetz It would also be nice in the documentation to add maybe an 'oc exec' command or similiar to get the logs from the container.  Ideally, this would come in conjunction with the PV

roles/openshift_logging/files/elasticsearch-logging.yml
ewolinetz: It should be fairly simple to make this a template instead and then we can have some `{% if %} {% endif %}` blocks

ewolinetz:
I personally think there needs to be a larger discussion on this to go over the impacts of this.. or in the least there should be an easy way to change the configuration if someone wants to maintain using console logs only without needing to manually edit the configmap after deployment

jcantrill:
@ewolinetz We had a discussion during the Arch meeting [1] to:
* accept this PR
* add work to configure PVC for file based logging.
* Add documentation about changes.

[1] https://docs.google.com/document/d/1m9-LXr-UThs9WhRfkFRbwf1Lppxa5Opw_PYRY1wuDbY/edit#heading=h.n33m14qzrf7m

jcantrill:
Closing in favor of #3777

PR #3688, opened 2017-03-16
Bump version to 3.6.0

smunilla:
Bump up the version to make our daily builds happy.


smunilla:
@sdodson PTAL i think this does the right thing.

PR #3691, opened 2017-03-16
Add external provisioners playbook starting with aws efs

wongma7:
Notes:
This doesn't hook up with the cluster installer by default. I don't know if we want to do that or if we can just keep a separate doc for installing these exclusively after a user has installed openshift
We will want to allow users to only install/deploy certain provisioners because not all make sense, currently I do 'when:openshift_provisioners_efs | bool' and soon there will be an e.g. openshift_provisoiners_cephfs. Hope there is a nicer way to do this

lots of things copied from openshift_logging.

~TODO : use volumename and claimref for the pv~

### DIFF
--- a/roles/openshift_provisioners/tasks/generate_clusterrolebindings.yaml
+++ b/roles/openshift_provisioners/tasks/generate_clusterrolebindings.yaml
@@ -1,0 +1,19 @@
+---
+- name: Generate ClusterRoleBindings
+  template: src=clusterrolebinding.j2 dest={{mktemp.stdout}}/templates/{{obj_name}}-clusterrolebinding.yaml
+  vars:
+    acct_name: provisioners-{{item}}
+    obj_name: run-provisioners-{{item}}
+    labels:
+      provisioners-infra: support
+    crb_usernames: ["system:serviceaccount:{{openshift_provisioners_project}}:{{acct_name}}"]
+    subjects:
+      - kind: ServiceAccount
+        name: "{{acct_name}}"
+        namespace: "{{openshift_provisioners_project}}"
+    cr_name: "system:persistent-volume-provisioner"
+  with_items:
+    # TODO
+    - efs
+  check_mode: no
+  changed_when: no
--- a/roles/openshift_provisioners/tasks/generate_secrets.yaml
+++ b/roles/openshift_provisioners/tasks/generate_secrets.yaml
@@ -1,0 +1,14 @@
+---
+- name: Generate secret for efs
+  template: src=secret.j2 dest={{mktemp.stdout}}/templates/{{obj_name}}-secret.yaml
+  vars:
+    name: efs
+    obj_name: "provisioners-efs"
+    labels:
+      provisioners-infra: support
+    secrets:
+      - {key: aws-access-key-id, value: "{{openshift_provisioners_efs_aws_access_key_id}}"}
+      - {key: aws-secret-access-key, value: "{{openshift_provisioners_efs_aws_secret_access_key}}"}
+  check_mode: no
+  changed_when: no
+  when: openshift_provisioners_efs | bool
--- a/roles/openshift_provisioners/tasks/generate_serviceaccounts.yaml
+++ b/roles/openshift_provisioners/tasks/generate_serviceaccounts.yaml
@@ -1,0 +1,12 @@
+---
+- name: Generating serviceaccounts
+  template: src=serviceaccount.j2 dest={{mktemp.stdout}}/templates/{{obj_name}}-sa.yaml
+  vars:
+    obj_name: provisioners-{{item}}
+    labels:
+      provisioners-infra: support
+  with_items:
+  # TODO
+  - efs
+  check_mode: no
+  changed_when: no
--- a/roles/openshift_provisioners/tasks/install_efs.yaml
+++ b/roles/openshift_provisioners/tasks/install_efs.yaml
@@ -1,0 +1,70 @@
+---
+- name: Check efs current replica count
+  command: >
+    {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig get dc provisioners-efs
+    -o jsonpath='{.spec.replicas}' -n {{openshift_provisioners_project}}
+  register: efs_replica_count
+  when: not ansible_check_mode
+  ignore_errors: yes
+  changed_when: no
+
+- name: Generate efs PersistentVolumeClaim
+  template: src=pvc.j2 dest={{mktemp.stdout}}/templates/{{obj_name}}-pvc.yaml
+  vars:
+    obj_name: "provisioners-efs"
+    size: "1Mi"
+    access_modes:
+      - "ReadWriteMany"
+    pv_selector:
+      provisioners-efs: efs
+  check_mode: no
+  changed_when: no
+
+- name: Generate efs PersistentVolume
+  template: src=pv.j2 dest={{mktemp.stdout}}/templates/{{obj_name}}-pv.yaml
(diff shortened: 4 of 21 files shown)


erinboyd:
@wongma7 Seems like might be a good opportunity to create a separate directory in the BYO playbooks specifically for storage. @childsb thoughts?

erinboyd:
@scollier Where do yours exist for the vsphere reference architecture? Also in the byo, but under vsphere? If this is so, @wongma7 we might want to go with adding this to the AWS folder and labeling it AWS-EFS. @childsb

sdodson:
@wongma7 Have any general documentation to help us understand how this works in general?

wongma7:
@sdodson yes, the idea is to run these external provisioners https://github.com/kubernetes-incubator/external-storage#what-is-an-external-provisioner as DCs and they will watch PVCs of a certain StorageClass then create PVs when needed. They are controllers in and do it the same way the PV-controller does it. E.g. by default this EFS one is for storage classes that have "provisioner: openshift.org/aws-efs" so somebody would deploy this, create a storage class "fast" with "provisioner: openshift.org/aws-efs", then a user would create a PVC asking for "fast," and then the DC's pod would automatically create a PV for that PVC.

All provisioners will need a serviceaccount + clusterrolebinding to "system:persistent-volume-provisioner" because that clusterrole grants the necessary (usually admin-restricted) permissions to create PVs 
Most provisioners (gluster block, cephfs, this) will need secrets generated for them, e.g. here the EFS one needs a secret with AWS credentials
Some provisioners may need extra specific things, e.g. here the EFS provisioner needs a PVC+PV.

wongma7:
sample invocation I used to test

sudo env "PATH=$PATH" ansible-playbook ./playbooks/common/openshift-cluster/openshift_provisioners.yml \
-e openshift_provisioners_install_provisioners=True \
-e openshift_provisioners_efs_fsid=fs-xxxxxx \
-e openshift_provisioners_efs_region=us-west-2 \
-e openshift_provisioners_efs_aws_access_key_id=xxxx \
-e openshift_provisioners_efs_aws_secret_access_key=xxxx \
-e openshift_provisioners_image_prefix=quay.io/external_storage/ \
-e openshift_provisioners_efs=True \
-e deployment_type=origin

roles/openshift_provisioners/tasks/generate_clusterrolebindings.yaml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Generate ClusterRoleBindings
+  template: src=clusterrolebinding.j2 dest={{mktemp.stdout}}/templates/{{obj_name}}-clusterrolebinding.yaml
```
sdodson: Stick to yaml style ie:
```
template:
  src: clusterrolebinding.j2
  dest: {{mktemp.stdout}}/templates/{{obj_name}}-clusterrolebinding.yaml
vars:
...
```

roles/openshift_provisioners/tasks/generate_serviceaccounts.yaml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Generating serviceaccounts
+  template: src=serviceaccount.j2 dest={{mktemp.stdout}}/templates/{{obj_name}}-sa.yaml
```
sdodson: Stick to yaml style

roles/openshift_provisioners/tasks/generate_secrets.yaml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Generate secret for efs
+  template: src=secret.j2 dest={{mktemp.stdout}}/templates/{{obj_name}}-secret.yaml
```
sdodson: Stick to yaml style

roles/openshift_provisioners/tasks/install_efs.yaml
```diff
@@ -0,0 +1,70 @@
...
+  command: >
+    {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig get dc provisioners-efs
+    -o jsonpath='{.spec.replicas}' -n {{openshift_provisioners_project}}
+  register: efs_replica_count
+  when: not ansible_check_mode
+  ignore_errors: yes
+  changed_when: no
+
+- name: Generate efs PersistentVolumeClaim
+  template: src=pvc.j2 dest={{mktemp.stdout}}/templates/{{obj_name}}-pvc.yaml
```
sdodson: Stick to yaml style

roles/openshift_provisioners/tasks/install_efs.yaml
```diff
@@ -0,0 +1,70 @@
...
+    size: "1Mi"
+    access_modes:
+      - "ReadWriteMany"
+    pv_selector:
+      provisioners-efs: efs
+  check_mode: no
+  changed_when: no
+
+- name: Generate efs PersistentVolume
+  template: src=pv.j2 dest={{mktemp.stdout}}/templates/{{obj_name}}-pv.yaml
```
sdodson: Stick to yaml style

roles/openshift_provisioners/tasks/install_support.yaml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Check for provisioners project already exists
+  command: >
```
sdodson: Can you use oc_project here?

sdodson:
aos-ci-test

sdodson:
@wongma7 Can you open this against release-1.5 too and do a quick test to verify that everything works there? Based on when you filed this PR I imagine you don't depend on any of the recent changes in master but we should run through it on top of release-1.5 before we build it.

wongma7:
Opened release-1.5 pr here https://github.com/openshift/openshift-ansible/pull/3861 and worked fine for me.

I will address feedback in the next pr as suggested

wongma7:
TASK [openshift_node : Start and enable openvswitch service] *******************
fatal: [10.8.168.171]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to start service openvswitch.service: Job for openvswitch.service failed because the control process exited with error code. See \"systemctl status openvswitch.service\" and \"journalctl -xe\" for details.\n"}

is the test failure

sdodson:
@wongma7 Yeah, 3.6 tests are failing now due to problems with that image and there was a bug in the CI bot that overwrites one of our required statuses. That's since been fixed, re-running.

sdodson:
aos-ci-test

childsb:
[merge]

sdodson:
aos-ci-test

sdodson:
yum needed -y flag to install python-pip, this has already been fixed in aos-cd-jobs
[merge]

roles/openshift_provisioners/tasks/install_provisioners.yaml
```diff
@@ -0,0 +1,55 @@
+---
+- name: Check that EFS File System ID is set
+  fail: msg='the openshift_provisioners_efs_fsid variable is required'
+  when: (openshift_provisioners_efs | bool) and openshift_provisioners_efs_fsid is not defined
```
rhcarvalho: This is the type of verification that would be nice to run at the very beginning of the installation, in order to fail fast (and not half-way through the install).

I don't see a way how to run a "pre-condition" check for multiple roles automatically unless we settle on a name convention and write some code to discover pre-condition-checking files.

Another possibility is to move this precondition checking to a single role that runs early on, e.g. `openshift_sanitize_inventory`. The problem with that approach is the perhaps poor factorization of the code.

@sdodson what are the current thoughts on how to handle this, running pre-condition checks early on?

PR #3694, opened 2017-03-16
enable docker excluder since the time it is installed

ingvagabund:
Once again.

If the docker-excluder is installed and enabled by a user, it must by enabled even during installation, resp. upgrade. If a user unexclude the excluder, the gets excluded at the end of the installation/upgrade.

Bug 1430612

### DIFF
--- a/playbooks/common/openshift-cluster/initialize_openshift_version.yml
+++ b/playbooks/common/openshift-cluster/initialize_openshift_version.yml
@@ -23,6 +23,9 @@
   vars:
     # the excluders needs to be disabled no matter what status says
     with_status_check: false
+    # Only openshift excluder needs to be temporarily disabled
+    # So ignore the docker one
+    enable_docker_excluder: false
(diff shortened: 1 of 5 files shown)


ingvagabund:
@jhadvig how far are we from making this PR work in origin? Can you test it again?

ingvagabund:
aos-ci-test

sdodson:
Looks like we never reverted this change on release-1.5 so no need to pick it back there.

stevekuznetsov:
This did _not_ pass the Origin CI tests, why was it merged?

sdodson:
I should not have merged it.

ashcrow:
@sdodson it happens.

jhadvig:
Tested the change and its breaks the install process. Its fine after reverting the change

sdodson:
@jhadvig should be fine now that excluders were fixed on origin

PR #3703, opened 2017-03-17
Adding oc_group to lib_openshift.

kwoodson:


ashcrow:
Looks like there are some ``flake8`` and ``pylint`` stuff that needs updating. Other than that it looks good.

roles/lib_openshift/src/ansible/oc_group.py
```diff
@@ -0,0 +1,35 @@
...
+            debug=dict(default=False, type='bool'),
+            name=dict(default=None, type='str'),
+            namespace=dict(default='default', type='str'),
+            # addind users to a group is handled through the oc_users module
+            #users=dict(default=None, type='list'),
+        ),
+        supports_check_mode=True,
+    )
+
+    gconfig = GroupConfig(module.params['name'],
```
ashcrow: Looks like this snuck in again. :smile:

roles/lib_openshift/src/class/oc_group.py
```diff
@@ -0,0 +1,152 @@
...
+            return True
+
+        return False
+
+    def get(self):
+        '''return group information '''
+        result = self._get(self.kind, self.config.name)
+        if result['returncode'] == 0:
+            self.group = Group(content=result['results'][0])
+        elif 'groups \"%s\" not found' % self.config.name in result['stderr']:
```
ashcrow: nit(not_required): I'd recommend moving away from ``%s``'s and towards string formatting like so:

(assuming the \'s are required in the output check)

```python
elif 'groups \"{}\" not found'.format(self.config.name) in result['stderr']:
```

ashcrow:
@kwoodson don't worry about coveralls :smile:

kwoodson:
@ashcrow, heh, thanks.

PR #3707, opened 2017-03-20
Do not require python-six via openshift_facts

rhcarvalho:
It turns out the code in
`roles/openshift_health_checker/openshift_checks/__init__.py` had a
mistake in the import statement, and not the case that `add_metaclass`
is not available in the embedded `six` from `ansible.module_utils`.

The fix is to correct the import and do not require python-six to be
installed on target hosts.

Note: now that we have some unit tests in the role, errors like that
would have been caught.

Follows up #3641.

### DIFF
--- a/roles/openshift_facts/vars/main.yml
+++ b/roles/openshift_facts/vars/main.yml
@@ -2,7 +2,6 @@
 required_packages:
   - iproute
   - python-dbus
-  - python-six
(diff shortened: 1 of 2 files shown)


rhcarvalho:
aos-ci-test

rhcarvalho:
[merge]

PR #3709, opened 2017-03-20
Implement namespace-reader role for router shards

mtnbikenc:
Current documented process for using router shards involves adding the `cluster-reader` role to the router service account in order to read namespace labels.
https://docs.openshift.com/container-platform/3.4/install_config/router/default_haproxy_router.html#using-router-shards

This PR takes suggestions for creating a new role `namespace-reader` to provide the only necessary access for the service account.  The role will be created when either `NAMESPACE_LABELS` or `ROUTE_LABELS` is specified in the router configuration.

@knobunc @smarterclayton

### DIFF
--- a/roles/openshift_hosted/tasks/router/router.yml
+++ b/roles/openshift_hosted/tasks/router/router.yml
@@ -37,13 +37,39 @@
     resource_name: hostnetwork
   with_items: "{{ openshift_hosted_routers }}"
 
-- name: Set additional permissions for router service account
+- name: Create the role namespace-reader when using Router Shards
+  oc_obj:
+    kind: ClusterRole
+    name: namespace-reader
+    namespace: "{{ item.namespace }}"
+    content:
+      path: /tmp/osh_namespace_reader
+      data:
+        apiVersion: v1
+        kind: ClusterRole
+        metadata:
+          name: namespace-reader
+        rules:
+        - apiGroups:
+          - ""
+          attributeRestrictions: null
+          resources:
+          - namespaces
+          - namespaces/status
+          verbs:
+          - get
+          - list
+          - watch
+  when: ('NAMESPACE_LABELS' in (item | string )) or ('ROUTE_LABELS' in (item | string ))
+  with_items: "{{ openshift_hosted_routers }}"
+
+- name: Set permissions for router service account when using Router Shards
   oc_adm_policy_user:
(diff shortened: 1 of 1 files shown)


mtnbikenc:
@knobunc @smarterclayton Following up to get your feedback on this implementation.

smarterclayton:
@liggitt @deads2k for security review, this roughly looks like what I expect.

roles/openshift_hosted/tasks/router/router.yml
```diff
@@ -37,13 +37,39 @@
     resource_name: hostnetwork
   with_items: "{{ openshift_hosted_routers }}"
 
-- name: Set additional permissions for router service account
+- name: Create the role namespace-reader when using Router Shards
+  oc_obj:
+    kind: ClusterRole
+    name: namespace-reader
+    namespace: "{{ item.namespace }}"
```
deads2k: ClusterRoles can't have namespaces.  Did you mean Role?

roles/openshift_hosted/tasks/router/router.yml
```diff
@@ -37,13 +37,39 @@
...
+        kind: ClusterRole
+        metadata:
+          name: namespace-reader
+        rules:
+        - apiGroups:
+          - ""
+          attributeRestrictions: null
+          resources:
+          - namespaces
+          - namespaces/status
```
deads2k: I don't think you need access to status.

roles/openshift_hosted/tasks/router/router.yml
```diff
@@ -37,13 +37,39 @@
...
+          - get
+          - list
+          - watch
+  when: ('NAMESPACE_LABELS' in (item | string )) or ('ROUTE_LABELS' in (item | string ))
+  with_items: "{{ openshift_hosted_routers }}"
+
+- name: Set permissions for router service account when using Router Shards
   oc_adm_policy_user:
     user: "system:serviceaccount:{{ item.namespace }}:{{ item.serviceaccount }}"
     namespace: "{{ item.namespace }}"
```
deads2k: Is this trying to bind at a namespace scope?  If you need list or watch on namespaces, this won't let you do it.  You'll need the permission at the cluster scope.

smarterclayton:
To be able to select on labels, this needs cluster scope.

mtnbikenc:
Will come back to this once it is prioritized.
https://trello.com/c/y1mZ7hK3

PR #3711, opened 2017-03-20
Adding option to specify different cert/key/ca for kibana ops route

ewolinetz:
looks like we missed this feature from the deployer and QE hasn't reported it yet.


sdodson:
3.5 blocker or just nice to have?

sdodson:
rebased to get rid of the merge commit, based on standup call we'd like to get this in ahead of 3.5. so merging.

ewolinetz:
aos-ci-test

PR #3721, opened 2017-03-21
Adding oc_configmap to lib_openshift.

kwoodson:


roles/lib_openshift/src/class/oc_configmap.py
```diff
@@ -0,0 +1,183 @@
...
+        return ["--from-literal={}={}".format(key, value) for key, value in self.from_literal.items()]
+
+    def get(self):
+        '''return a configmap by name '''
+        results = self._get('configmap', self.name)
+        if results['returncode'] == 0 and results['results'][0]:
+            self.configmap = results['results'][0]
+
+        if results['returncode'] != 0 and '"{}" not found'.format(self.name) in results['stderr']:
+            results['returncode'] = 0
```
stevekuznetsov: Why is this necessary?

roles/lib_openshift/src/class/oc_configmap.py
```diff
@@ -0,0 +1,183 @@
...
+
+        results = self.openshift_cmd(cmd, output=output)
+
+        return results
+
+    def update(self):
+        '''run update configmap '''
+        return self._replace_content('configmap', self.name, self.inc_configmap)
+
+    def needs_update(self):
```
stevekuznetsov: Why is there logic here to do these checks? It's pretty inherently racy and `oc apply` will do this for you instead of you having to break out `update`, `create`, `delete`

roles/lib_openshift/src/class/oc_configmap.py
kwoodson: @stevekuznetsov, when doing an `oc get <kind> foo` its returncode is 1 and stderr has the message above.  When in reality from an api standpoint, when fetching something we don't want to error.  Doing a get from an API isn't an error, it should return empty.  This piece of code sets the returncode to 0.

stevekuznetsov: Why is the semantic incorrect in Origin? A 404 is an error

roles/lib_openshift/src/class/oc_configmap.py
kwoodson: @stevekuznetsov, Last time I attempted to use `oc apply`, it has been a while I admit, it was not idempotent.  `oc apply` takes the changes whether they were needed or not.  If I am wrong then I will update.

I apologize, I do not understand your comment.
> and oc apply will do this for you instead of you having to break out update, create, delete
Are you stating that oc apply will delete objects?  I didn't see that in the documentation.
I understand that `oc create`, `oc replace`, and `oc delete` are necessary and used appropriately.

roles/lib_openshift/src/doc/configmap
```diff
@@ -0,0 +1,72 @@
...
+  debug:
+    description:
+    - Turn on debug output.
+    required: false
+    default: False
+    aliases: []
+  name:
+    description:
+    - Name of the object that is being queried.
+    required: false
```
ewolinetz: I think this should be required

roles/lib_openshift/src/test/integration/oc_configmap.yml
```diff
@@ -0,0 +1,94 @@
...
+  - debug: var=cmout
+
+  - name: assert configmaptest exists
+    assert:
+      that:
+      - cmout.results.results[0].metadata.name == 'configmaptest'
+      - cmout.results.results[0].data.foo == 'bar'
+  ###### end create test ###########
+
+  ###### update test ###########
```
ewolinetz: Can we also add an update test to verify that we can overwrite existing keys correctly? (Just for sanity)

roles/lib_openshift/src/doc/configmap
```diff
@@ -0,0 +1,72 @@
...
+    default: None
+    aliases: []
+  namespace:
+    description:
+    - The namespace where the object lives.
+    required: false
+    default: str
```
ewolinetz: default: default

roles/lib_openshift/src/class/oc_configmap.py
```diff
@@ -0,0 +1,183 @@
...
+        return results
+
+    def delete(self):
+        '''delete a configmap by name'''
+        return self._delete('configmap', self.name)
+
+    def create(self, dryrun=False, output=False):
+        '''Create a configmap
+
+           :dryrun: Product what you would have done. default: False
```
ewolinetz: Should this be exposed as a parameter to the module, or is this automatically populated in checkmode?

roles/lib_openshift/src/doc/configmap
kwoodson: thanks, will fix.

roles/lib_openshift/src/class/oc_configmap.py
kwoodson: @ewolinetz, good question.  `create` is inside of the OCConfigMap class and can be called by instantiating the class.  The pattern we have been using is calling a `staticmethod` named `run_ansible` that receives all of the parameters it needs to complete the desired ansible steps.  I do not expose `create` to the module.  This allows us to write python tests and debug in pdb.  

The reason you see the `dryrun` and `output` parameters is because I am attempting to accomplish idempotency by reusing the same code path for `create` but only changing the output so that I can see what would have happened in the `state: present` and the object already exists.  At that point, I compare the in-memory version vs what would have been created to determine if I need to update.

Hopefully this makes sense.  After re-reading your question, I think you were asking if check_mode would return what would or could have happened.  I don't currently do that but that is a good idea.

If you want to discuss, please ping and we'll chat in irc.

roles/lib_openshift/src/test/integration/oc_configmap.yml
kwoodson: Thanks, updated.

roles/lib_openshift/src/doc/configmap
kwoodson: I will add this.

kwoodson:
@ewolinetz, Updated with your suggestions.

roles/lib_openshift/src/test/integration/oc_configmap.yml
```diff
@@ -0,0 +1,95 @@
...
+
+  - debug: var=cmout
+
+  - name: assert configmaptest exists
+    assert:
+      that:
+      - cmout.results.results[0].metadata.name == 'configmaptest'
+      - cmout.results.results[0].data.deployment_type == 'online'
+      - cmout.results.results[0].data.foo == 'notbar'
+  ###### end create test ###########
```
ewolinetz: semantics -- 'end update test'

kwoodson: will fix

roles/lib_openshift/src/class/oc_configmap.py
stevekuznetsov: I understand that this module is very similar to the other CRUD type modules, so I do not expect a change to be made here but rather in a larger refactor, but I see the following problems with this approach:
 - you are defining a different semantic for error than normal REST
 - you are defining a different semantic for error then Origin
 - you are propagating the `result` structure from this `oc` call up the module level -- I assume this is why the change is necessary, otherwise the module could fail
 - you are keying off of a human-readable text which has no backwards compatibility guarantee (and would fail for instance when we internationalize the client)

roles/lib_openshift/src/class/oc_configmap.py
stevekuznetsov: `oc apply` attempts to drive the server state of objects to the state you declare you want to exist. It should be idempotent in that if the server state already matches what you need it will take no action.

There are certain places where the strategic merge patches created internally on the client inside of `oc apply` may not correctly honor rules about types, but that will be worked on as `oc apply` is moved into the server.

`oc apply` _can_ delete objects, that is why the `--prune` flag exists.

As I mentioned in another comment, I understand this module to be very similar to many others so I would not expect these types of comments to incur change in this PR, but further architectural process should definitely happen as we continue to refine what this interface looks like. I think in specific, two things can improve:
 - use of `PATCH` instead of `UPDATE` to clobber less
 - generic `oc CRUD` module backbone for different types

roles/lib_openshift/src/class/oc_configmap.py
ashcrow: To follow the ansible patterns already set using a ```returncode``` of ```0``` makes sense. It is true that a 404 Not Found is technically an HTTP error code, but from the REST stance it's a response which means the object/data does not exist. That placed into ansible land would be a valid response IMHO.

stevekuznetsov: Right, my comment is that the propagation of the `returncode` of this specific `oc` call to the module level is incorrect. The `oc get` here is one component and directly returning the `results` as the `module.fail_json()` is an incorrect handling of the output.

roles/lib_openshift/src/class/oc_configmap.py
```diff
@@ -0,0 +1,183 @@
...
+        # Delete
+        ########
+        if state == 'absent':
+            if not Utils.exists(api_rval['results'], params['name']):
+                return {'changed': False, 'state': 'absent'}
+
+            if check_mode:
+                return {'changed': True, 'msg': 'CHECK_MODE: Would have performed a delete.'}
+
+            api_rval = oc_cm.delete()
```
ashcrow: Is there a need to check the ``api_rval`` to ensure the delete was successful?

roles/lib_openshift/src/class/oc_configmap.py
kwoodson: @stevekuznetsov,  Good catch on the `--prune` although this feels very strange to apply a prune.

>  It should be idempotent in that if the server state already matches what you need it will take no action.

The question comes in then, how does an administrator determine when things are actually changed?  With what you are proposing to switch to `oc apply` means that every task that is created, updated, deleted, or noop'd (because it exists or does not exist) would return that it changed because `oc apply` will simply  "do the right thing" and always report changed.

Here is an example:
```bash
[root@ip-172-31-60-178 ~]# oc apply -f mysql.yaml  && echo $?
deploymentconfig "mysql" configured
0
[root@ip-172-31-60-178 ~]# oc apply -f mysql.yaml  && echo $?
deploymentconfig "mysql" configured
0
[root@ip-172-31-60-178 ~]# vim mysql.yaml
[root@ip-172-31-60-178 ~]# oc apply -f mysql.yaml  && echo $?
deploymentconfig "mysql" configured
0
[root@ip-172-31-60-178 ~]# oc apply -f mysql.yaml  && echo $?
deploymentconfig "mysql" configured
0
```

Notice the vim edit in the middle.  I modified the yaml inside and as you stated, `oc apply` certainly applied it but it also applied it when it didn't need to.  This is the reason why `oc apply` is a show stopper.

I am happy to discuss the reasons why we need to see that `state: present` and `changed: True|False`.  These are imperative to the way ansible works and how operations manages clusters.

roles/lib_openshift/src/class/oc_configmap.py
kwoodson: @ashcrow, I'll add that, thanks.

roles/lib_openshift/src/class/oc_configmap.py
stevekuznetsov: Very good question.

Today, with `apply` being client side you could try doing `oc apply --dry-run` and seeing if it determined if changes would be necessary or not, but ideally this would be a good topic to be brought up in the architecture discussion as well as in the discussion of the apply migration to the server. It is not unreasonable to ask if the apply API on the server would return to you what patch would be generated and/or if the operation changed something, as the inherent race between "oc apply --dry-run` and `oc apply` could render the result of the first "check" call to be invalid.

stevekuznetsov: As mentioned above, though, there are clearly some gaps with `apply` in specific scenarios, but in the meanwhile there can be progress made on re-architecting these CRUD wrappers. I think in specific, two things can improve:
 - use of `PATCH` instead of `UPDATE` to clobber less and more markedly communicate user intent where incremental updates are what the user expects
 - generic `oc CRUD` module backbone for different types

kwoodson: @stevekuznetsov, Thanks for the conversation.  The `oc patch` has the same behavior of `oc apply`.
```bash
[root@ip-172-31-60-178 ~]# oc patch dc/mysql -p '{"spec":{"replicas": 2}}'
"mysql" patched
[root@ip-172-31-60-178 ~]# echo $?
0
[root@ip-172-31-60-178 ~]# oc patch dc/mysql -p '{"spec":{"replicas": 2}}'
"mysql" patched
[root@ip-172-31-60-178 ~]# echo $?
0
```
If we can push this behavior upstream so that we can reliably determine changes actually occurred I'm fully on board with updating to match those tools.

The other issue I have with patch is that it takes slices of an item.  Most of the time we want to manage state and when doing so we want multiple edits to occur to an object.

As for the example, `oc patch node k8s-node-1 -p '{"spec":{"unschedulable":true}}'` this requires passing a `-p` and a specific or `strategic` slice into the file.  I did not research more into this per the issue of it always resulting in a returncode of 0 and saying, "patched".

Does patch also allow multiple cuts as a single transaction?  Does it take an entire object?  Until then, I think the `oc replace` for an update, which does a `put` under the covers is exactly what we want.  Its a single transaction and it we get the behavior that we want.

I'm definitely open for discussion when it comes to architecting the modules and improving them.  I'm not sold on `PATCH` over `oc replace`.

stevekuznetsov: The biggest problem with `oc replace` (doing an `UPDATE`) is that the resource version is not required, and when it is set to `0` or omitted, as is often with serialized config on disk you want to use to drive state, the operation clobbers data on the server and can be dangerous.

In cases where the installer or the user or whatever is driving the client knows for certain that it controls the entire document and can determine the state of the entire document at once, there is no problem. However, once the user is attempting to convey the intent of "I want to make this edit" it is not going to work as you would expect. Consider this case:

 - user `kenny` reads a resource off of the server
 - user `kenny` makes some local edits to the resource on his local machine
 - user `steve` updates the resource on the server
 - user `kenny` uses `oc replace` with his edited version

If the intent here was that `kenny` would make an edit to the object, this is the wrong flow. In this case, even if `steve` made edits to a completely unrelated portion of the object, `oc replace` will clobber the entirety of the object with what user `kenny` posted. 

Your points re: the "did I update" semantic are valid and should be brought up in Kubernetes CLI SIG. Perhaps also the question should be fielded on how the upstream Ansible modules are working with this. But in either case when a user wants to update a portion of an object, and does not know for certain that they control the entirety of the object, the `oc replace` is a dangerous and incorrect tool to use. Providing an explicit resource version can help detect places where unknown clobber would occur but in general you'd still be re-pulling, re-editing and re-pushing (and may race with someone else again) but if you just submitted a patch you'd be fine.

kwoodson: @stevekuznetsov,  thanks for the explanations on how `patch` works.  I researched early on in the 3.1-3.2 days and decided with its behavior it wasn't going to meet our needs.

When lib_openshift makes the changes it does use the resource version of an object.  We have run into issues when these have been modified and the module will fail when openshift refuses to replace with an old resource version.  This is undesirable but is something we will have to handle.  Currently we do not handle this and force the user to rerun to apply state.  This is not a good user experience but it is safer to guarantee state.  I imagine we will want to build in retry loops and timeouts.  This is the first version of the modules and they will continue to develop.

The questions about `patch` still remain.
Can `patch` perform multiple edits at a single time?  If not, are you proposing we run multiple `patch` commands to achieve state?  Isn't this susceptible to the race conditions you mentioned above?
How does `patch` handle race conditions?  It seems to me that the situation you described would result in whoever made the edits last would win.  This is kind of the same behavior as `replace` except we attempt to obey resource versions.  I'm not sure we can guarantee state if multiple administrators are attempting to modify the same object if its not done with a transaction.  As soon as `patch` makes changes and another call to `patch` occurs you run the risk of someone making a different change and now the module cannot, with confidence, return that the desired "state" was achieved.

Good discussion going here.

stevekuznetsov: >  I imagine we will want to build in retry loops and timeouts.

Again, I think this really depends on what you want to do here -- if your intent is that you have the _complete_ and _correct_ version of state for an object, do not send the resource version and ensure that your version is what exists on the server. If you mean to _edit_ specific fields of an object, use `oc patch`.

> Can patch perform multiple edits at a single time?

Yes, you can for example do 
```
oc patch type/name --patch='{"spec":{"field": "value","otherfield": "whateverelse"}}"
```

>  Isn't this susceptible to the race conditions you mentioned above?

No. The race I spoke of is for changes to unrelated fields in the object.

> It seems to me that the situation you described would result in whoever made the edits last would win. This is kind of the same behavior as replace except we attempt to obey resource versions. 

Right, the person who makes the last edits does win. The critical difference is that with `patch` you say "I wan _this_ field to change" and thereby you honor edits others make to _unrelated_ fields. When you `oc replace` you say "I want the entire object to look like this".

>  As soon as patch makes changes and another call to patch occurs you run the risk of someone making a different change and now the module cannot, with confidence, return that the desired "state" was achieved.

This will always happen no matter the means by which objects are changed on the API server. This is intended. When multiple people edit in a sequence the last edit will be the one that is reflected in the current state of the object. That is not the racy behavior I mentioned above.
```

stevekuznetsov: So to be clear, the previous statement about `patch` as it could be used today is not meant to be "we should always use `patch`". There are use-cases for `patch` and use-cases for `replace`. These are separate use-cases. I would argue that a `replace` in a re-try loop is a very suspect operation. What would the re-try loop look like?

 1. read object state from server
 2. make edit
 3. try to replace object state on server

When you pass resource version on the `replace` call and get a failure, it means someone else changed the object between steps 1 and 3. So there are a couple cases:

 1. you know the full and correct object state and mean with `oc replace` that the entirety of your YAML or JSON is correct and nothing else should be on the server
 2. you know some subset of correct object state and mean with `oc replace` to edit that portion of the object

For the first option, it does not matter what the other person edited between steps 1 and 3 ... you know the full and correct state. Pass no resource version and ensure that your object is on the server as you intend.

For the second option, the user making an edit between steps 1 and 3 could either change the fields you mean to edit, or unrelated fields. If they change unrelated fields, this means that in your step 2 you'll make a new version of the object with their edits and yours, then submit that to `oc replace`. This is the use-case where `oc patch` is what you really want to use, because the process of reading an object, patching it and re-introducing it into the server is what `oc patch` does... but you are able to encode your intent to change a field and the server can see that the other unrelated change won't impact you.

In the other case, the other unrelated user made a change to the field you meant to change and you either know that your field is always correct or sometimes correct. If you know that you have the correct values for the fields in your `patch` you should be fine to use `oc patch` whenever. If you only want to do the patch sometimes... then realistically you want `oc apply` with pre-conditions on your action. This is not yet implemented but you can expect it to be.

kwoodson:
aos-ci-test
[test]

kwoodson:
@ashcrow, @ewolinetz, could I get a :+1: ?

kwoodson:
[merge]

kwoodson:
aos-ci-test

kwoodson:
[test]

kwoodson:
@sdodson, I ran the aos-ci-test and I think it exploded.  At this point, can we just merge or do we need to do the song and dance?  This code is not referenced anywhere.  Brand new but CICD and logging were looking at using it.

sdodson:
Most other jobs recently have passed, I agree the error is entirely outside of your control but a re-run should clear things up.

sdodson:
aos-ci-test

kwoodson:
[merge]

sdodson:
flake https://github.com/openshift/openshift-ansible/issues/3765

sdodson:
@kwoodson created a flake for the situation where merge happens before aos-ci-test tests were green, can you try to merge again?

kwoodson:
[merge]

kwoodson:
flake #3765 [merge]

stevekuznetsov:
Yeah so today we can only refer to flakes in one repo and that has been configured to be Origin  ...  if someone has cycles to extend the logic to contain multiple repos that would be very useful.

stevekuznetsov:
In the meanwhile I'll rectify this PR merge

sdodson:
@stevekuznetsov thanks

PR #3725, opened 2017-03-21
Update Dockerfile.rhel7 to reflect changes to Dockerfile

codificat:
The Dockerfile in the repo was updated to build upon playbook2image in #3382 

This change updates de product version to match that change. Also updating the labels to the new format.

Also installs the `oc` client.

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -1,26 +1,41 @@
-FROM rhel7
+FROM openshift3/playbook2image
 
-MAINTAINER Troy Dawson <tdawson@redhat.com>
+MAINTAINER OpenShift Team <dev@lists.openshift.redhat.com>
 
-LABEL Name="openshift3/installer"
-LABEL Vendor="Red Hat" License=GPLv2+
-LABEL Version="v3.1.1.901"
-LABEL Release="6"
-LABEL BZComponent="aos3-installation-docker"
-LABEL Architecture="x86_64"
-LABEL io.k8s.description="Ansible code and playbooks for installing Openshift Container Platform." \
-      io.k8s.display-name="Openshift Installer" \
-      io.openshift.tags="openshift,installer"
+LABEL name="openshift3/openshift-ansible" \
+      summary="OpenShift's installation and configuration tool" \
+      description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain and check an OpenShift cluster" \
+      url="https://github.com/openshift/openshift-ansible" \
+      io.k8s.display-name="openshift-ansible" \
+      io.k8s.description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain and check an OpenShift cluster" \
+      io.openshift.expose-services="" \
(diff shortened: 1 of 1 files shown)


Dockerfile.rhel7
```diff
@@ -1,26 +1,43 @@
-FROM rhel7
+FROM playbook2image
```
aweiteka: Is this a placeholder image name for a specific build environment?

Dockerfile.rhel7
```diff
@@ -1,26 +1,43 @@
...
-    rpm -V $INSTALL_PKGS && \
-    yum clean all
+# Playbooks, roles and their dependencies are installed from packages.
+# Unlike in Dockerfile, we don't invoke the 'assemble' script here
+# because all content is already provided. Instead, we just point
+# $APP_HOME to the location of openshift-ansible.
+# The 'run' script needs write access to $APP_HOME, hence the chown.
+USER root
+RUN INSTALL_PKGS="atomic-openshift-utils atomic-openshift-clients" && \
+    yum repolist > /dev/null && \
```
aweiteka: Is this for build debugging?

Dockerfile.rhel7
codificat: Indeed, until we clarify `playbook2image`'s final location downstream

Dockerfile.rhel7
codificat: No, it's needed so that `yum-config-manager` works (see below).

Alternatively we can just run `yum --enablerepo=... install` instead of using `yum-config-manager`

https://access.redhat.com/solutions/1443553
**NOTE:** Until the first yum command is run, `/etc/yum.repos.d/redhat.repo` contains no repositories, so `yum-config-manager` will not enable/disable anything.

aweiteka: Ah, right. Thanks for the reminder!

Dockerfile.rhel7
codificat: I'll reference the image as `openshift3/playbook2image` for the time being - maybe still not final, but at least it will have a namespace...

Dockerfile.rhel7
```diff
@@ -1,26 +1,43 @@
...
+    yum-config-manager --enable rhel-7-server-ose-3.4-rpms && \
+    yum install -y $INSTALL_PKGS && \
+    yum clean all && \
+    chown ${USER_UID} /usr/share/ansible/openshift-ansible && \
+    rm -rf ${APP_HOME} && \
```
codificat: No need to do that I think, `playbook2image` knows how to work from a different location using `WORK_DIR`, will update the PR using that

sdodson:
aos-ci-test

sdodson:
[merge]

rhcarvalho:
The file being changed is unrelated to the test failure, must've been a flake.
I could not properly decipher the output -- https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/106/console.

Let's try to [merge] again.

PR #3726, opened 2017-03-21
Update imageConfig.format on upgrades to match oreg_url

sdodson:
If someone (like @jhadvig) had updated oreg_url value when upgrading we should update that.


jhadvig:
Tested the change with the upgrade playbook and the format is updated correctly now.
Thanks @sdodson :+1:

sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
flake https://github.com/openshift/origin/issues/9999
[merge]

PR #3730, opened 2017-03-22
 Bug 1434300 - Log entries are generated in ES after deployed logging stacks via ansible, but can not be found in kibana.

richm:


ewolinetz:
@richm this should opened against 'release-1.5' if this is intended to fix the 3.5 release per the bz referenced.

Will keep this open as the fix should also be in master

sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
> @ewolinetz can this be merged since the release-1.5 version of this was already merged?

👍 I realize we're in a mad dash to get 3.5 wrapped up, but we really shouldn't be merging anything to the branches without having the same change merged into master already. The risk of forgetting something and then regressing in the future is too high.

PR #3731, opened 2017-03-22
openshift_logging calculate min_masters to fail early on split brain

jcantrill:
This PR was discussed with @portante to avoid split brain scenarios.  This should cause the start of the cluster to fail early in such situations, as seen in BZ 1424637.

cc @sdodson this is one we should consider cherry-picking into 1.5

### DIFF
--- a/roles/openshift_logging/vars/main.yaml
+++ b/roles/openshift_logging/vars/main.yaml
@@ -1,6 +1,8 @@
 ---
 openshift_master_config_dir: "{{ openshift.common.config_base }}/master"
 es_node_quorum: "{{openshift_logging_es_cluster_size|int/2 + 1}}"
+es_min_masters_default: "{{ (openshift_logging_es_cluster_size | int / 2 | round(0,'floor') + 1) | int }}"
+es_min_masters: "{{ (openshift_logging_es_cluster_size == 1) | ternary(1, es_min_masters_default)}}"
 es_recover_after_nodes: "{{openshift_logging_es_cluster_size|int - 1}}"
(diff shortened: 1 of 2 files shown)


roles/openshift_logging/vars/main.yaml
```diff
@@ -1,6 +1,8 @@
 ---
 openshift_master_config_dir: "{{ openshift.common.config_base }}/master"
 es_node_quorum: "{{openshift_logging_es_cluster_size|int/2 + 1}}"
+es_min_masters_default: "{{ (openshift_logging_es_cluster_size | int / 2 | round(0,'floor') + 1) | int }}"
```
portante: This looks wonky: `round(0,'floor')`?  Won't that always be 0?  Or am I missing something?

sdodson: http://jinja.pocoo.org/docs/2.9/templates/#round the first argument is actually the data passed in via the pipe

sdodson: so `round( openshift_logging_es_cluster_size | int / 2, precision=0, method='round')`

sdodson:
@jcantrill please open a pr against release-1.5 too

sdodson:
aos-ci-test

roles/openshift_logging/vars/main.yaml
jcantrill: @sdodson Are you asking me to change it?  @portante the first arg is to 'what decimal you round to' so this basically says round to the 'ones place' and then use 'floor' for rounding instead of default

sdodson: No, just explaining how the filters work. LGTM

PR #3733, opened 2017-03-22
Introduce Elasticsearch readiness probe

lukas-vlcek:
Depends on https://github.com/openshift/origin-aggregated-logging/pull/358

### DIFF
--- a/roles/openshift_logging/templates/es.j2
+++ b/roles/openshift_logging/templates/es.j2
@@ -95,6 +95,13 @@ spec:
               readOnly: true
             - name: elasticsearch-storage
               mountPath: /elasticsearch/persistent
+          readinessProbe:
+            exec:
+              command:
+              - "/usr/share/elasticsearch/probe/readiness.sh"
+            initialDelaySeconds: 5
+            timeoutSeconds: 4
+            periodSeconds: 5
       volumes:
         - name: elasticsearch
           secret:


lukas-vlcek:
@jcantrill @ewolinetz please add "Do not merge" label for now.

lukas-vlcek:
@ewolinetz @jcantrill @richm PTAL
Once https://github.com/openshift/origin-aggregated-logging/pull/358 is merged we are ready to merge this PR as well.

richm:
lgtm, but I don't know anything about readiness probes

roles/openshift_logging/templates/es.j2
```diff
@@ -95,6 +95,13 @@ spec:
               readOnly: true
             - name: elasticsearch-storage
               mountPath: /elasticsearch/persistent
+          readinessProbe:
+            exec:
+              command:
+              - "/usr/share/elasticsearch/probe/readiness.sh"
+            initialDelaySeconds: 5
+            timeoutSeconds: 5
+            periodSeconds: 5
```
ewolinetz: This does not appear to be a configuration option for readiness probes:
https://docs.openshift.org/latest/dev_guide/application_health.html

lukas-vlcek: @ewolinetz I found `periodSeconds` example here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#defining-readiness-probes

ewolinetz: ack +1
it looks like its further explained here: https://kubernetes.io/docs/api-reference/v1.6/#probe-v1-core

lukas-vlcek: @ewolinetz however, if this is not applicable here we shall remove it, WDYT?

lukas-vlcek:
@ewolinetz may be we can continue with this PR as needed https://github.com/openshift/origin-aggregated-logging/pull/358 has been merged

ewolinetz:
aos-ci-test

ewolinetz:
i will also need to test this locally since the ci test doesn't test logging yet

ewolinetz:
aos-ci-test

roles/openshift_logging/templates/es.j2
```diff
@@ -95,6 +95,13 @@ spec:
               readOnly: true
             - name: elasticsearch-storage
               mountPath: /elasticsearch/persistent
+          readinessProbe:
+            exec:
+              command:
+              - "/usr/share/elasticsearch/probe/readiness.sh"
+            initialDelaySeconds: 5
+            timeoutSeconds: 5
```
rhcarvalho: Beware that the timeout doesn't work for "exec" probes: https://github.com/kubernetes/kubernetes/issues/26895

lukas-vlcek: @rhcarvalho Thanks! Interesting...
Let me ask you then how exactly this works (/me confused a bit 😮 ).

- If `timeoutSeconds` is not present, is any default values used for it instead?
- Does it mean that readiness probe works the same regardless `timeoutSeconds` being setup at all?
- If the `command` points to shell script which has [internal timeout](https://github.com/openshift/origin-aggregated-logging/pull/358/files#diff-e0db40ac0a1f0750c954836bcba1c905) (specifically: `curl --max-time ...`) does it make sense to bother with probe `timeoutSeconds` at all? Can we just rely on the internal script timeout instead?

rhcarvalho: @lukas-vlcek, [here's the API type definition of a Probe](https://github.com/openshift/origin/blob/e37455a407f05ac584f5c462effc2b8c3626f39d/vendor/k8s.io/kubernetes/pkg/api/v1/types.go#L1300-L1326).

> If timeoutSeconds is not present, is any default values used for it instead?

Yes, 1 second. But, as I linked above, there's a long-standing issue that timeouts are not enforced for "exec" probes. So in practice, for the time being, setting this value has no effect.

> Does it mean that readiness probe works the same regardless timeoutSeconds being setup at all?

Yes, precisely. At least the Docker Remote API never supported timing out processes executed via the equivalent of a "docker exec ...". There is some ground work done to support this in other container backends, but AFAIK it is not a reality for most people/installations.

> If the command points to shell script which has internal timeout (specifically: curl --max-time ...) does it make sense to bother with probe timeoutSeconds at all? Can we just rely on the internal script timeout instead?

My recommendation would be setting a timeout on the command as you say, yes.

lukas-vlcek: @rhcarvalho Interestingly the API you linked defines the default value for `timeoutSeconds` to be 10 seconds :-) But I think in practice it is 1 second as you say.

So, as of now, the only timeout that can lead to readiness probe failure is the one that is implemented internally within the check script itself. Probably it makes sense then to remove the `timeoutSeconds` settings from this PR.

lukas-vlcek: @ewolinetz @jcantrill @richm I am going to remove the `timeoutSeconds` from DeploymentConfig in this PR. This means we will have to rely on timeout implemented directly in the check itself, which is by default set to 1 sec.

rhcarvalho: @lukas-vlcek 10s is the default for PeriodSeconds.
But those are only comments... the actual defaulting happens [here](https://github.com/openshift/origin/blob/e37455a407f05ac584f5c462effc2b8c3626f39d/vendor/k8s.io/kubernetes/pkg/api/v1/defaults.go#L184-L197).

lukas-vlcek: @rhcarvalho thanks!

rhcarvalho: > I am going to remove the timeoutSeconds from DeploymentConfig in this PR. This means we will have to rely on timeout implemented directly in the check itself, which is by default set to 1 sec.

No matter what value you set or not set, there is no timeout enforcement.

I suggest you pass a timeout to `curl` that is less than `periodSeconds`, otherwise you could have probes overlapping each other.

lukas-vlcek: @rhcarvalho  my only worry is then this (taken from `man curl`):

````
 -m, --max-time <seconds>
              Maximum time in seconds that you allow the whole operation to take.  This is useful for preventing your batch jobs from hanging
              for  hours  due to slow networks or links going down.  Since 7.32.0, this option accepts decimal values, but the actual timeout
              will decrease in accuracy as the specified timeout increases in decimal precision.  See also the --connect-timeout option.

              If this option is used several times, the last one will be used.
````

Notice:  "`Since 7.32.0, this option accepts decimal values, ...`"

I only hope there will be always curl that can interpret decimal values for this option correctly.

lukas-vlcek: Created a new ticket to tune timeout value in the ES readiness probe script: https://github.com/openshift/origin-aggregated-logging/issues/370

rhcarvalho: I don't get why we need to use decimal values?

lukas-vlcek:
Now, once https://github.com/openshift/origin-aggregated-logging/issues/372 is merged we can merge this PR as well.

Apology for the confusion. (_Readiness probes are buggy and not well documented._)
Kudos to @rhcarvalho for helping to clear this out.

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

ewolinetz:
[merge]

PR #3734, opened 2017-03-22
openshift_repos: ensure {openshift_,}deployment_type is defined

sosiouxme:
We really want to set openshift_deployment_type, but users will likely
still have just deployment_type, so accept both. And don't make every
playbook default openshift_deployment_type to deployment_type.


roles/openshift_repos/tasks/main.yaml
```diff
@@ -1,13 +1,20 @@
 ---
+- set_fact:
+    deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
+    openshift_deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
```
sosiouxme: @detiber wanted to get your opinion on whether this is the right place to ensure these both default to each other. `openshift_facts` and `openshift_repos` do not depend on each other at this time. Maybe do it in both places?

sosiouxme:
aos-ci-test [aos-ci-test]

sosiouxme:
aos-ci-test

roles/openshift_repos/tasks/main.yaml
detiber: Instead of using set_fact here, you should be able to provide a default for openshift_deployment_type:
roles/openshift_repos/defaults/main.yaml:
```
openshift_deployment_type: "{{ deployment_type | default(None) }}"
```
If openshift_deployment_type is defined elsewhere it will override the value set here.

roles/openshift_repos/tasks/main.yaml
```diff
@@ -1,13 +1,20 @@
 ---
+- set_fact:
+    deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
```
detiber: Instead of defining both deployment_type and openshift_deployment_type, you can nest default filters:
```
deployment_type: "{{ openshift_deployment_type | default(deployment_type | default(None)) }}"
```

roles/openshift_facts/tasks/main.yml
```diff
@@ -64,8 +64,7 @@
     role: common
     local_facts:
       debug_level: "{{ openshift_debug_level | default(2) }}"
-      # TODO: Deprecate deployment_type in favor of openshift_deployment_type
-      deployment_type: "{{ openshift_deployment_type | default(deployment_type) }}"
+      deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
```
detiber: This will fail if deployment_type is not set...  you will need to nest the second default filter similar to my other comment.

roles/openshift_repos/tasks/main.yaml
```diff
@@ -1,13 +1,20 @@
...
+    openshift_deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
+
+- name: Ensure a valid deployment type has been given
+  fail:
+    msg: "Please set openshift_deployment_type to one of:\n{{ known_openshift_deployment_types | join(', ') }}"
+  when: >
+    openshift_deployment_type is not defined
```
detiber: openshift_deployment_type will always be defined because of set_fact above (or if moving to defaults).

tbielawa:
@sosiouxme @detiber  Can anyone clarify these `openshift_deployment_type` vs `deployment_type` differences/objectives for me?

> We really want to set `openshift_deployment_type`

Why is that?

> but users will likely still have just `deployment_type`, so accept both
> And don't make every playbook default `openshift_deployment_type` to `deployment_type`.

In our [Advanced Installation](https://docs.openshift.com/container-platform/latest/install_config/install/advanced_install.html) documentation we strictly reference `deployment_type` (5 times) and do not mention `openshift_deployment_type` at all. This pattern goes all the way back to the [3.0 docs](https://docs.openshift.com/enterprise/3.0/install_config/install/advanced_install.html).

I guess what I want clarified/answered is this:

We want to use `openshift_deployment_type` in playbooks/roles. However, the problem is that **currently** users are instructed through online documentation (above) and our example inventories to use the `deployment_type`. As a result of this mis-match extra work has been required in to the past to reassign `deployment_type` to `openshift_deployment_type` and then more work to get the defaulting corrected depending on which variable was provided. This PR tries to remove the excessive repeated reassignments (ex: `openshift_deployment_type: "{{ deployment_type }}"`) and corrects identified tasks to rely on `openshift_deployment_type` instead. Additionally, this PR updates the example inventory files to only show use of the preferred `openshift_deployment_type` variable.

Is that correct? If so I believe we should tag in the docs team on this PR (add docs label) so that the online documentation reflects our preferred approach.

Followup on that (probably work for a different PR/card): Similar to how we have pre-req playbooks, we could introduce some kind of *deprecation checks* playbook. A great example of that playbook would be checking if `deployment_type` is defined and clearly and **loudly** notifying the user that `deployment_type` will be completely deprecated in an upcoming release and that they must switch to `openshift_deployment_type` in the future.

sosiouxme:
On Thu, Mar 23, 2017 at 11:38 AM, Tim Bielawa <notifications@github.com>
wrote:

> @sosiouxme <https://github.com/sosiouxme> @detiber
> <https://github.com/detiber> Can anyone clarify these
> openshift_deployment_type vs deployment_type differences/objectives for
> me?
>
> We really want to set openshift_deployment_type
>
> Why is that?
>

To me, consistency with other variables which mostly all begin with
`openshift_` now. And various points in the code that use both.


> but users will likely still have just deployment_type, so accept both
> And don't make every playbook default openshift_deployment_type to
> deployment_type.
>
> In our Advanced Installation
> <https://docs.openshift.com/container-platform/latest/install_config/install/advanced_install.html>
> documentation we strictly reference deployment_type (5 times) and do not
> mention openshift_deployment_type at all. This pattern goes all the way
> back to the 3.0 docs
> <https://docs.openshift.com/enterprise/3.0/install_config/install/advanced_install.html>
> .
>

If you look carefully, while `deployment_type` is *mentioned* in the
examples, it is never described in the parameter listings (and the docs do
not link to the great example inventories that do describe it). Given
installs fail without this specified, it seemed like a significant
oversight. This is one thing I wanted to bring up with docs, but I figured
as long as that's the case it might be nice to standardize the name before
describing it, and allow users to continue to specify the old name in their
existing inventories.


> I guess what I want clarified/answered is this:
>
> We want to use openshift_deployment_type in playbooks/roles. However, the
> problem is that *currently* users are instructed through online
> documentation (above) and our example inventories to use the
> deployment_type. As a result of this mis-match extra work has been
> required in to the past to reassign deployment_type to
> openshift_deployment_type and then more work to get the defaulting
> corrected depending on which variable was provided. This PR tries to remove
> the excessive repeated reassignments (ex: openshift_deployment_type: "{{
> deployment_type }}") and corrects identified tasks to rely on
> openshift_deployment_type instead. Additionally, this PR updates the
> example inventory files to only show use of the preferred
> openshift_deployment_type variable.
>
> Is that correct? If so I believe we should tag in the docs team on this PR
> (add docs label) so that the online documentation reflects our preferred
> approach.
>
Sounds right. I also would like to improve the messaging when this
parameter is missing or invalid (tried to do so in this PR).

> Followup on that (probably work for a different PR/card): Similar to how
> we have pre-req playbooks, we could introduce some kind of *deprecation
> checks* playbook. A great example of that playbook would be checking if
> deployment_type is defined and clearly and *loudly* notifying the user
> that deployment_type will be completely deprecated in an upcoming release
> and that they must switch to openshift_deployment_type in the future.
>

Actually I was starting to think about this in terms of introducing an
`openshift_inventory` role that would validate and standardize the
variables specified prior to anything else. With this variable it would
accept either name and make sure the one we want is defined and valid. With
`openshift_release` for example it would strip off a leading `v` and
perhaps check that (if specified) it's even a possible version given the
deployment type. Without going down the slippery slope of bringing in lots
of business logic here that belongs in other roles, it seems like we could
do some minimal things up front.

Deprecation warnings could certainly go here, although the ansible
experience isn't so great for expressing anything that's not succeed/fail.
Perhaps when there are deprecation warnings, the play could pause or
require user acknowledgement.

roles/openshift_facts/tasks/main.yml
sosiouxme: Not in my experience - it succeeds as written even if neither variable is defined. I wonder if the behavior is different between ansible versions? I don't mind rewriting it, just looked a little cleaner to me this way.

roles/openshift_repos/tasks/main.yaml
sosiouxme: only if `openshift_facts` role comes before `openshift_repos` which is not guaranteed as it's not a dependency. Also it could have been defined as None...

I wanted to do this check here because it's where `known_openshift_deployment_types` is defined. But this could all be moved into a shared `openshift_inventory` role as mentioned below.

roles/openshift_repos/tasks/main.yaml
sosiouxme: I don't see `openshift_deployment_type` defined anywhere, except as defaulting to `deployment_type`.

sosiouxme: Now I see what you're saying. Actually, I want `openshift_deployment_type` to be set globally (i.e. for all depending roles and plays), not just in the `openshift_repos` role. There doesn't seem to be a way other than `set_facts` to do that. Am I missing something?

sosiouxme:
@detiber @tbielawa perhaps like this second commit... what do you think?

rhcarvalho:
Will we keep compatibility with `deployment_type` for at least some time (with a deprecation warning)?

roles/openshift_inventory/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+    set_fact:
+      deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
+      openshift_deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
+
+  - name: Normalize openshift_release
+    set_fact:
+      openshift_release: "{{ openshift_release | string | regex_replace('^v', '') }}"
+    when: openshift_release is defined
+
+- name: Ensure a valid deployment type has been given |
```
tbielawa: Nit pick: why the trailing `|` character?

roles/openshift_inventory/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+  when: >
+    openshift_deployment_type is not defined
+    or openshift_deployment_type not in known_openshift_deployment_types
+  fail:
+    msg: |-
+      Please set openshift_deployment_type to one of:
+      {{ known_openshift_deployment_types | join(', ') }}
```
tbielawa: I'm curious about your use of the YAML `>` character for indicating *folded-style* string content in this `when`. That's just so you could write the condition of the `when` across multiple lines, right?

And the `|-` character here I am not 100% certain of the exact side-effects. I understand that the `|` pipe character is used to indicate a literal string in which new-line characters **are** preserved. I am not sure what the `-` hyphen after the pipe indicates.

roles/openshift_inventory/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+---
+- no_log: True  # keep task descriptions legible
+  block:
+
+  - name: Standardize on latest variable names
```
tbielawa: I'm glad we're working to automatically standardize variable names but without an explanatory comment block near by I think the reasoning for this fact setting could be lost on the casual reader. Also, clearly expressing what the intended end state is supposed to be would be nice.

roles/openshift_repos/meta/main.yml
```diff
@@ -11,4 +11,5 @@ galaxy_info:
     - 7
   categories:
   - cloud
-dependencies: []
+dependencies:
+- role: openshift_inventory
```
tbielawa: I see role dependency proliferation happening if we continue down this path (adding `openshift_inventory` as a dep in module meta files). Is there any avenue available for us to make this a standard pre-run action instead?

roles/openshift_inventory/tasks/main.yml
sosiouxme: Agree that further explanation is desirable here.

roles/openshift_inventory/tasks/main.yml
sosiouxme: Ansible jams on all of the parameters in the output after the name. So it's the difference in readability between:

`TASK [openshift_inventory : Ensure a valid deployment type has been given msg=Please set openshift_deployment_type to one of:`

and:

`TASK [openshift_inventory : Ensure a valid deployment type has been given | msg=Please set openshift_deployment_type to one of:`

In the second case it's visually clearer where you can stop reading. If there were some way (aside from no_log which also suppresses useful debug output) to not show the action parameters in the header I would want to do that. No one else seems too concerned about the output ansible gives the user but... to me, it would be nice if it were as clear as possible.

roles/openshift_inventory/tasks/main.yml
sosiouxme: There are a lot of ways to define multiline strings in YAML: http://stackoverflow.com/a/21699210/262768
I guess it doesn't matter much which I use here.

roles/openshift_repos/meta/main.yml
sosiouxme: I don't see a way to add this globally via ansible.cfg or something. I agree it would seem cleaner but on the other hand it introduces spooky action at a distance, so maybe it's best that it doesn't seem possible.

roles/openshift_inventory/tasks/main.yml
tbielawa: Cool. Yes, the second one is easier to parse. :+1:

sosiouxme: To be fair, a period at the end is almost as clear and less surprising to see in the YAML.

roles/openshift_inventory/tasks/main.yml
tbielawa: Good reference post, I couldn't find a summary that succinct before

> /me adds to `DOCS` bookmark folder

From you reference I now understand that the `|` pipe followed by a `-` hyphen means 
in
>  **Literal style** - turns newlines within the string into literal newlines, and adds one at the end.

With the `-` *chomping indicator* the `strip` action happens:

> **strip**: remove the line feed, remove the trailing blank lines.
So that will remove the last new-line character at the end of the string (recall from above that the `|` literal appends a `\n` to the string) and trim off any trailing blank lines.

*Oh YAML, you so crazy some times*

roles/openshift_repos/meta/main.yml
tbielawa: Yeah. Let's avoid that distant spooky.

roles/openshift_inventory/tasks/main.yml
rhcarvalho: I second that the ` |` is surprising. The "name" should be something semantically reasonable, a period would be better.


tl;dr

"Improving" the output by adding that suffix is not that great, it is like sticking \<font\> tags in HTML5.
For example, if we send a patch upstream to improve the appearance of tasks in the default callback plugin, we could end up getting something like `TASK ... been given || msg=...`. Also, we would be improving one message at the expense of lacking consistency with the rest.

sosiouxme: I'll make it a period for now. I think it would be better in the long run to improve the output upstream as you say, just seems like there would be a lot of bikeshedding to get through.

sosiouxme:
@tbielawa seems cautiously optimistic, any opinion from @detiber or @sdodson on whether this is a good direction to go?

roles/openshift_inventory/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+    set_fact:
+      # goal is to deprecate deployment_type in favor of openshift_deployment_type.
```
tbielawa: :+1:

roles/openshift_inventory/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+      # both will be accepted for now, but code should refer to the new name.
+      # TODO: once this is well-documented, add deprecation notice if using old name.
+      deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
+      openshift_deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
+
+  - name: Normalize openshift_release
+    set_fact:
+      # normalize release if provided, e.g. "v3.5" => "3.5"
```
tbielawa: :+1:

roles/openshift_inventory/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+      openshift_release: "{{ openshift_release | string | regex_replace('^v', '') }}"
+    when: openshift_release is defined
+
+- name: Ensure a valid deployment type has been given.
+  # this variable is required; complain early and clearly if it is invalid.
```
tbielawa: :+1:

roles/openshift_inventory/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+- no_log: True  # keep task descriptions legible
+  block:
+
+  - name: Standardize on latest variable names
+    set_fact:
+      # goal is to deprecate deployment_type in favor of openshift_deployment_type.
+      # both will be accepted for now, but code should refer to the new name.
+      # TODO: once this is well-documented, add deprecation notice if using old name.
+      deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
+      openshift_deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
```
rhcarvalho: Why do we default to `None` if that is not a valid value?

roles/openshift_inventory/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+  - name: Normalize openshift_release
+    set_fact:
+      # normalize release if provided, e.g. "v3.5" => "3.5"
+      openshift_release: "{{ openshift_release | string | regex_replace('^v', '') }}"
+    when: openshift_release is defined
+
+- name: Ensure a valid deployment type has been given.
+  # this variable is required; complain early and clearly if it is invalid.
+  when: |
+    openshift_deployment_type is not defined
```
rhcarvalho: How can it ever be "not defined" if we define it in a `set_fact` above?

roles/openshift_repos/meta/main.yml
rhcarvalho: Does `openshift_inventory` get called multiple times during a run of the install playbook?

roles/openshift_version/meta/main.yml
```diff
@@ -12,6 +12,7 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
+- role: openshift_inventory
```
rhcarvalho: Why both `openshift_version` and `openshift_repos` depend on the new `openshift_inventory`?

BTW, maybe naming it `openshift_inventory_validation` would be more specific? So that we avoid turning it into other things like a dynamic inventory role or anything of the like.

roles/openshift_inventory/tasks/main.yml
sosiouxme: Good point, I was thinking `None` did not count as defined but python/jinja2 do not see it that way. Can remove this line.

roles/openshift_version/meta/main.yml
sosiouxme: They both depend on the facts being correct before they execute.
I'm certainly open to a more descriptive name. Normalization, defaults, and validation is what I have in mind; not sure if there might be yet more we would want to do up front with the inventory before other roles execute.

For some reason github isn't letting me respond to the previous question about `openshift_inventory` getting called multiple times. http://docs.ansible.com/ansible/playbooks_roles.html#role-dependencies says "By default, roles can also only be added as a dependency once - if another role also lists it as a dependency it will not be run again."

roles/openshift_repos/meta/main.yml
sosiouxme: http://docs.ansible.com/ansible/playbooks_roles.html#role-dependencies says "By default, roles can also only be added as a dependency once - if another role also lists it as a dependency it will not be run again."

sdodson:
aos-ci-test

sosiouxme:
how about a [merge] then

sdodson:
aos-ci-test

roles/openshift_inventory/tasks/main.yml
rhcarvalho: I'm not super confident in my understanding of how this YAML-programming actually works without doing local tests, but here's the logic I'm following:

- `set_fact` sets a variable
- `deployment_type` should be a string (conceptually)
- `deployment_type: {{ openshift_deployment_type | default(deployment_type) }}` means set `deployment_type` to `openshift_deployment_type` or to the value of itself?! I don't know what happens if we access a variable that is undefined... I verified it locally, it fails with a rather hard to grasp message:
```
TASK [set_fact] ****************************************************************
    "failed": true
}

MSG:

the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'deployment_type' is undefined

The error appears to have been in '/home/vagrant/src/github.com/openshift/openshift-ansible/foo.yml': line 4, column 5, but may

  tasks:
  - set_fact:
    ^ here
```

- So it seems that `default` allows us to reference undefined names without throwing errors...

My suggestion would be the empty string or no argument... but I don't know what different it makes. I tried all that in this test playbook and `default()`, `default(''`) and `default(None)` all print the same output!

```
$ cat foo.yml 
- hosts: localhost
  connection: local
  tasks:
  - set_fact:
      deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default() }}"
  - debug:
      msg: "deployment_type = {{ deployment_type }}"
[vagrant@localhost openshift-ansible]$ ansible-playbook -i hosts foo.yml 

ok: [localhost]

TASK [set_fact] ****************************************************************
ok: [localhost]

TASK [debug] *******************************************************************
ok: [localhost] => {}

MSG:

deployment_type = 


localhost                  : ok=3    changed=0    unreachable=0    failed=0  
```

---

So, a suggestion, perhaps we could do the same as we do below for `openshift_release` and mark it explicitly as a string?

```yaml
set_fact:
  openshift_deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default | string }}"
  deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default | string }}"
  openshift_release: "{{ openshift_release | default | string | regex_replace('^v', '') }}"
```

roles/openshift_inventory/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+      # both will be accepted for now, but code should refer to the new name.
+      # TODO: once this is well-documented, add deprecation notice if using old name.
+      deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
+      openshift_deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default(None) }}"
+
+  - name: Normalize openshift_release
+    set_fact:
+      # normalize release if provided, e.g. "v3.5" => "3.5"
+      openshift_release: "{{ openshift_release | string | regex_replace('^v', '') }}"
+    when: openshift_release is defined
```
rhcarvalho: Isn't `openshift_release` required to be defined? Is it unlike `openshift_deployment_type`?
If I had to guess, then I could imagine we figure out `openshift_release` based on available RPM/image versions, is that the case?

roles/openshift_repos/meta/main.yml
rhcarvalho: Yes, I've read that documentation before and still observed different behavior when we had the checks implemented using multiple roles. The role that was a common dependency got executed multiple times, despite of what the docs say... and I think we have the same problem with `openshift_facts`?

roles/openshift_version/meta/main.yml
rhcarvalho: I see, normalization is different than validation in that it mutates the inventory variables, validation would imply no changes... `openshift_invetory_preprocessing`?! idk...

roles/openshift_repos/meta/main.yml
sosiouxme: Yes it seems fixing https://github.com/openshift/openshift-ansible/pull/3734 hasn't been a priority so we get multiple invocations of role dependencies. That would be nice to get fixed in general, but I don't think it should block this... this role is quick and idempotent (although showing multiple deprecation notices would be pretty annoying, to be sure).

roles/openshift_inventory/tasks/main.yml
sosiouxme: Yeah, alright, that's fine with me.

I don't want to unconditionally set `openshift_release` though due to logic in other roles that checks whether it's defined.

roles/openshift_repos/meta/main.yml
rhcarvalho: > Yes it seems fixing #3734 hasn't been a priority so we get multiple invocations of role dependencies.

#3734 is this PR, I believe you mean some Ansible issue?

Perhaps a viable alternative are plugins, just like we use https://github.com/openshift/openshift-ansible/blob/master/callback_plugins/aa_version_requirement.py to verify Ansible's version before anything else runs.

roles/openshift_version/meta/main.yml
sosiouxme: `openshift_sanitize_inventory`? `openshift_prepare_inventory`?

rhcarvalho: We don't need to bikeshed on this, `openshift_sanitize_inventory` sounds reasonable. Gives out a bit more of the intention without one needing to resort to the README in the role.

roles/openshift_inventory/tasks/main.yml
sosiouxme: Need to provide some clear comments here and documentation elsewhere about the expectations for `openshift_release`. The official docs don't link [this explanation](https://github.com/openshift/openshift-ansible/blob/403b5c5545b9b07e7a4416e5a63609512ddaa224/inventory/byo/hosts.ose.example#L28) yet, which is what @dgoodwin considers authoritative.

It's not currently required to be defined for an install (looks like upgrades do require it though). Not-defined essentially gets you "latest" although it's a little more complicated than that in `openshift_version`, and in the case of containerized origin you _do_ need to specify `openshift_release` (so we avoid installing alpha builds by default).

Surprisingly, the main use of `openshift_release` doesn't seem to be determining what version to install, but rather verifying that the version to be installed (or upgraded to) matches it. `openshift_version` can figure out what is available in most cases without it, and does not update it according to its findings. When I discussed this with @dgoodwin I came to the conclusion that other variables are eventually defined that determine the actual installation parameters, and we should preserve `openshift_release` as the user's expression of what they wanted, even if they don't specify it.

But it would certainly be clearer if we just required it to be set. It's not documented in the official docs, and the change would break the usage of existing playbooks, but the fix would be pretty easy and obvious.

roles/openshift_repos/meta/main.yml
sosiouxme: Argh, mispaste, I meant https://github.com/ansible/ansible/issues/20169

roles/openshift_inventory/tasks/main.yml
dgoodwin: I would rephrase one part here slightly, the main use of openshift_release is for the user to specify generically what major release they want to install or upgrade to, without having to know x.y.z.w levels of precision. That was the intent anyhow. It's mainly just a seed value for cluster setup/upgrade, after which we pin the version to the first master and/or what's already installed.

sdodson:
[merge]

playbooks/byo/openshift-cluster/config.yml
```diff
@@ -7,5 +7,4 @@
   vars:
     openshift_cluster_id: "{{ cluster_id | default('default') }}"
     openshift_debug_level: "{{ debug_level | default(2) }}"
-    openshift_deployment_type: "{{ deployment_type }}"
     openshift_deployment_subtype: "{{ deployment_subtype | default(none) }}"
```
rhcarvalho: Hmm, so there is a `deployment_subtype` as well!

rhcarvalho: Rarely used, only ever set to `None` or `'registry'`. According to the changelog in openshift-ansible.spec:

> Secure registry for atomic registry deployment (deployment_subtype=registry)

rhcarvalho:
"Conformance test" failed... is it just me or the Jenkins logs are painfully hard to understand?! https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_extended_conformance_install/85/console

This was probably a flake, let's try to re-run the tests and [merge] again.

rhcarvalho:
aos-ci-test

rhcarvalho:
@sosiouxme have you seen this code in openshift_facts?
https://github.com/openshift/openshift-ansible/blob/8e26f8cf9594693702fd1e545b47116cbbff34f0/roles/openshift_facts/library/openshift_facts.py#L1741-L1771

It refers to `deployment_type` in the existing facts, perhaps it needs an update as well?


Got it from a comment: https://github.com/openshift/openshift-ansible/pull/3785#discussion_r108790854.

sosiouxme:
On Thu, Mar 30, 2017 at 5:43 AM, Rodolfo Carvalho <notifications@github.com>
wrote:

> @sosiouxme <https://github.com/sosiouxme> have you seen this code in
> openshift_facts?
>


> It refers to deployment_type in the existing facts, perhaps it needs an
> update as well?
>
>
Yeah, but it's referring to openshift.common.deployment_type which is that
structure openshift_facts itself populates after setup. I didn't see a need
to disturb it just because it has the same name.

rhcarvalho:
@sosiouxme makes sense. Thanks!

Indeed the previous failure was a test flake. Merged now anyway :fireworks:

PR #3735, opened 2017-03-22
Bug 1434300 - Log entries are generated in ES after deployed logging …

richm:
…stacks via ansible, but can not be found in kibana.

https://bugzilla.redhat.com/show_bug.cgi?id=1434300
Add support for common data model index pattern files.
Depends on https://github.com/ViaQ/elasticsearch-templates/pull/36
and
https://github.com/openshift/origin-aggregated-logging/pull/357

@ewolinetz ptal

PR #3736, opened 2017-03-22
Revert "Bug 1434300 - Log entries are generated in ES after deployed logging …"

sdodson:
Reverts openshift/openshift-ansible#3735


ewolinetz:
@sdodson there isn't a need to revert. the changes this would revert do nothing until the new ES images are available.

PR #3737, opened 2017-03-22
Upgrade specific rpms instead of just master/node.

dgoodwin:
Dependencies for these, particularly the SDN package, can cause the
entire transaction to jump to a newer openshift than we requested, if
something newer is available in the repositories. By being specific for
multiple packages, we avoid this problem and get the actual version we
require.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
@@ -1,7 +1,26 @@
 ---
 # We verified latest rpm available is suitable, so just yum update.
-- name: Upgrade packages
-  package: "name={{ openshift.common.service_type }}-{{ component }}{{ openshift_pkg_version }} state=present"
+
+# Master package upgrade ends up depending on node and sdn packages, we need to be explicit
+# with all versions to avoid yum from accidentally jumping to something newer than intended:
+- name: Upgrade master packages
+  package: name={{ item }} state=present
+  when: component == "master"
+  with_items:
+  - "{{ openshift.common.service_type }}{{ openshift_pkg_version }}"
+  - "{{ openshift.common.service_type }}-master{{ openshift_pkg_version }}"
+  - "{{ openshift.common.service_type }}-node{{ openshift_pkg_version }}"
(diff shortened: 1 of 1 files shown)


sdodson:
LGTM, just ping me when you know this works for you.

dgoodwin:
Had to just add the base package to the list, now pushed, otherwise seems to be working fine. Ready for merge and ty for the quick response!

NOTE: no need to run big test suites unless they cover upgrade now.

playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
```diff
@@ -1,7 +1,26 @@
...
+  - "{{ openshift.common.service_type }}-clients{{ openshift_pkg_version }}"
+
+- name: Upgrade node packages
+  package: name={{ item }} state=present
+  when: component == "node"
+  with_items:
+  - "{{ openshift.common.service_type }}{{ openshift_pkg_version }}"
+  - "{{ openshift.common.service_type }}-node{{ openshift_pkg_version }}"
+  - "{{ openshift.common.service_type }}-sdn-ovs{{ openshift_pkg_version }}"
+  - "{{ openshift.common.service_type }}-clients{{ openshift_pkg_version }}"
```
sdodson: I'm wondering if we should just simplify things to upgrade all the packages at once no matter which component is being updated. Do you remember why we did that?

I say this because if for some reason we updated node before master, which we don't today, I think we may get into the same situation  where we go to the latest due to dependencies so perhaps it's best to just do them all at once, it would also simplify things a bit here.

dgoodwin: This does actually do them all at once, they're just separate blocks so we don't have to include "master" in the list when upgrading a node. Even before when you upgraded master it would upgrade node rpm because it had to, so I think that remains the same with this. 

If there's a better way to skip the master rpm on nodes let me know, I couldn't think of one.

sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
flake https://github.com/openshift/origin/issues/12235
[merge]

PR #3742, opened 2017-03-22
Update metrics-deployer.yaml

ebalsumgo:
So as the templates are pushed out now, there is a problem in the fact that according to https://access.redhat.com/containers/#/registry.access.redhat.com/openshift3/metrics-deployer/images the 3.4.0 tag is not actually getting all the latest releases of the images. Switching this to tag v3.4 fixes this.


ewolinetz:
resolved by #3906

PR #3743, opened 2017-03-22
Update logging-deployer.yaml

ebalsumgo:
So as the templates are pushed out now, there is a problem in the fact that according to https://access.redhat.com/containers/#/registry.access.redhat.com/openshift3/logging-deployer/images the 3.4.0 tag is not actually getting all the latest releases of the images. Switching this to tag v3.4 fixes this.

### DIFF
--- a/roles/openshift_hosted_templates/files/v1.4/enterprise/logging-deployer.yaml
+++ b/roles/openshift_hosted_templates/files/v1.4/enterprise/logging-deployer.yaml
@@ -223,7 +223,7 @@ items:
   -
     description: 'Specify version for logging components; e.g. for "registry.access.redhat.com/openshift3/logging-deployer:3.4.0", set version "3.4.0"'
     name: IMAGE_VERSION
-    value: "3.4.0"
+    value: "v3.4"
   -
     description: "(Deprecated) Specify the name of an existing pull secret to be used for pulling component images from an authenticated registry."
     name: IMAGE_PULL_SECRET


richm:
This is a dup of https://github.com/openshift/openshift-ansible/pull/3744 ?

ebalsumgo:
this one was su[pposed to be for the metrics... seems somehow my PRs got
pushed for the logging-deployer twice. the PR is the exact same though,
just for metrics-deployer.yaml.


-B

On Mon, Mar 27, 2017 at 8:51 PM, Richard Megginson <notifications@github.com

richm:
LGTM

ewolinetz:
@sdodson is this change necessary? shouldn't there eventually be a `3.4.0` image?

ebalsumgo:
@ewolinetz the reason i am pushed this PR is due to the fact that if you look at registry.access.redhat.com the metrics and logging images that are tagged as 3.4.x are not linking to latest releases of said images. so basically for the last 4 releases of 3.4.x i have been installing and re-installing the 3.4.0 image of the logging stack rather than the latest one that has bugfixes, etc etc etc. I am not saying this is the THE solution but based on the tags that are available in the resgitry the v3.x tag seems to be the correct one that will get you the latest release of the appropriate 3.x version.

ewolinetz:
resolved by https://github.com/openshift/openshift-ansible/pull/3906

PR #3744, opened 2017-03-22
Update logging-deployer.yaml

ebalsumgo:
So as the templates are pushed out now, there is a problem in the fact that according to https://access.redhat.com/containers/#/registry.access.redhat.com/openshift3/logging-deployer/images the 3.4.0 tag is not actually getting all the latest releases of the images. Switching this to tag v3.4 fixes this.

### DIFF
--- a/roles/openshift_hosted_templates/files/v1.5/enterprise/logging-deployer.yaml
+++ b/roles/openshift_hosted_templates/files/v1.5/enterprise/logging-deployer.yaml
@@ -223,7 +223,7 @@ items:
   -
     description: 'Specify version for logging components; e.g. for "registry.access.redhat.com/openshift3/logging-deployer:3.4.0", set version "3.4.0"'
     name: IMAGE_VERSION
-    value: "3.4.0"
+    value: "v3.4"
   -
     description: "(Deprecated) Specify the name of an existing pull secret to be used for pulling component images from an authenticated registry."
     name: IMAGE_PULL_SECRET


sdodson:
@ebalsumgo you've tested this? I just want to verify that the deployer is smart enough to pull the rest of the corresponding images as I don't see v3.4 tags for all the other images in the logging stack

ebalsumgo:
Yes this and the metrics templates as well


-B

-- Please excuse spelling and grammar errors, this has been sent from my iPhone/iPad

ebalsumgo:
for what its worth I have tested with v3.4 and with passing in the specific latest image tag... but I imagine that you dont want to be updating the templates every time there is a new image? I guess what I am saying is there might be some more need of tuning of the image build/release process

roles/openshift_hosted_templates/files/v1.5/enterprise/logging-deployer.yaml
```diff
@@ -223,7 +223,7 @@ items:
   -
     description: 'Specify version for logging components; e.g. for "registry.access.redhat.com/openshift3/logging-deployer:3.4.0", set version "3.4.0"'
     name: IMAGE_VERSION
-    value: "3.4.0"
+    value: "v3.4"
```
ewolinetz: shouldn't this be `v3.5` given this is a 1.5 file?

ewolinetz:
The 1.5 templates for logging is no longer needed. Closing

PR #3745, opened 2017-03-22
Add unit tests for health checks

rhcarvalho:
Still WIP

- Add several unit tests to cover the existing code base
- Rename module_executor -> execute_module for better readability (still supports the old name to avoid breaking in-flight PRs)

@brenton, @codificat, @juanvallejo, @sosiouxme FYI

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/__init__.py
+++ b/roles/openshift_health_checker/openshift_checks/__init__.py
@@ -63,6 +63,21 @@ class OpenShiftCheck(object):
                 yield subclass
 
 
+LOADER_EXCLUDES = (
+    "__init__.py",
+    "mixins.py",
+)
+
+
+def load_checks():
+    """Dynamically import all check modules for the side effect of registering checks."""
+    return [
+        import_module(__package__ + "." + name[:-3])
+        for name in os.listdir(os.path.dirname(__file__))
+        if name.endswith(".py") and name not in LOADER_EXCLUDES
+    ]
+
+
 def get_var(task_vars, *keys, **kwargs):
     """Helper function to get deeply nested values from task_vars.
 
@@ -78,15 +93,3 @@ def get_var(task_vars, *keys, **kwargs):
             return kwargs["default"]
         raise OpenShiftCheckException("'{}' is undefined".format(".".join(map(str, keys))))
     return value
-
-
-# Dynamically import all submodules for the side effect of loading checks.
-
-EXCLUDES = (
-    "__init__.py",
-    "mixins.py",
-)
-
-for name in os.listdir(os.path.dirname(__file__)):
-    if name.endswith(".py") and name not in EXCLUDES:
-        import_module(__package__ + "." + name[:-3])
--- a/roles/openshift_health_checker/openshift_checks/__init__.py
+++ b/roles/openshift_health_checker/openshift_checks/__init__.py
@@ -21,8 +21,13 @@ class OpenShiftCheckException(Exception):
 class OpenShiftCheck(object):
     """A base class for defining checks for an OpenShift cluster environment."""
 
-    def __init__(self, module_executor):
-        self.module_executor = module_executor
+    def __init__(self, execute_module=None, module_executor=None):
+        if execute_module is module_executor is None:
+            raise TypeError(
+                "__init__() takes either execute_module (recommended) "
+                "or module_executor (deprecated), none given")
+        self.execute_module = execute_module or module_executor
+        self.module_executor = self.execute_module
 
     @abstractproperty
     def name(self):
--- a/roles/openshift_health_checker/openshift_checks/docker_image_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/docker_image_availability.py
@@ -15,6 +15,9 @@ class DockerImageAvailability(OpenShiftCheck):
 
     skopeo_image = "openshift/openshift-ansible"
 
+    # FIXME(juanvallejo): we should consider other possible values of
+    # `deployment_type` (the key here). See
+    # https://github.com/openshift/openshift-ansible/blob/8e26f8c/roles/openshift_repos/vars/main.yml#L7
     docker_image_base = {
         "origin": {
             "repo": "openshift",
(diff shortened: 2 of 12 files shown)


roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -57,6 +65,14 @@ def subclasses(cls):
             for subclass in subclass.subclasses():
                 yield subclass
 
+    @staticmethod
```
sosiouxme: Feels like it would be slightly cleaner with:

```
     @classmethod
     def load_checks(cls):
         """Dynamically import all check modules for the side effect of registering checks."""
         for name in os.listdir(os.path.dirname(__file__)):
             if name.endswith(".py") and name not in cls.LOADER_EXCLUDES:
                 import_module(__package__ + "." + name[:-3])
```

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -21,8 +21,16 @@ class OpenShiftCheckException(Exception):
...
     """A base class for defining checks for an OpenShift cluster environment."""
 
-    def __init__(self, module_executor):
-        self.module_executor = module_executor
+    LOADER_EXCLUDES = (
+        "__init__.py",
+        "mixins.py",
+    )
+
+    def __init__(self, execute_module=None, module_executor=None):
```
rhcarvalho: Yeah, was not the original intention, but I thought it would help @juanvallejo not having to change inflight PRs, and me not having to rebase all the time and update.

For instance, https://github.com/openshift/openshift-ansible/pull/3461 merged yesterday with the old name. Of course the intention is to remove the old name as soon as we have the unit tests for all checks and require new checks to be tested (so that CI would alert us of a bug).

This is a known way on how to do safe refactors :)

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: Yes.

You made me think about this again. I added it here in an attempt to be able to write `OpenShiftCheck.load_checks()` 
and facilitating imports as we talked elsewhere -- but in fact because of inheritance we're also making possible `SomeRandomCheck.load_checks()`.
That latter will likely crash, because this ugly thing here uses things like `__package__` and relative paths to `__file__`...

The reason it should not be in the top level as it was is to avoid the side-effect from simply importing this module...
It is better to explicitly load checks in the action plugin, and we don't need to "accidentally" load checks when we import this module in tests.

Let's possibly bring it back to the package level, we don't want this inherited in every check subclass.

rhcarvalho:
@sosiouxme @juanvallejo PTAL / review.

I'd like to get this in so we all can work with tests in our PRs.

I'll defer to a follow up adding tests for our action and callback plugins to keep the diff reasonable for reviewing and to allow me more time to finish those.

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: @sosiouxme I've brought the code back to the module level, since it depends on module/file-level stuff like `__package__` and `__file__`. Please let me know what you think of the current state.

sosiouxme: Yeah, makes sense to me.

roles/openshift_health_checker/test/mixins_test.py
```diff
@@ -0,0 +1,23 @@
...
+
+@pytest.mark.parametrize('task_vars,expected', [
+    (dict(openshift=dict(common=dict(is_containerized=False))), True),
+    (dict(openshift=dict(common=dict(is_containerized=True))), False),
+])
+def test_is_active(task_vars, expected):
+    assert NotContainerizedCheck.is_active(task_vars) == expected
+
+
+def test_is_active_missing_config():
```
sosiouxme: I think a comment to the effect about why we expect this to throw an exception would help here.

e.g. "is_containerized requests an undefined task_var, resulting in an exception that should be caught and re-raised as OpenShiftCheckException"

sosiouxme:
aos-ci-test

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -62,9 +65,15 @@ def run(self, tmp, task_vars):
 
     def required_images(self, task_vars):
         deployment_type = get_var(task_vars, "deployment_type")
+        # FIXME(juanvallejo): we should handle gracefully with a proper error
+        # message when given an unexpected value for `deployment_type`.
```
rhcarvalho: Yes, your PR will take care of ensuring this is set to some know value.

In this case, we accept/handle less than all the currently known values. We should not raise `KeyError` when `deployment_type` is e.g. `online` or some future new entry to the list of known values (we need not to force a lock-step update of the list of known values here and to the rest of openshift-ansible). Hmm.. or.... would it make sense to read from the YAML files that defines the known values?
Anyway, this is a discussion for a future PR and commenting here will probably be like writing to /dev/null :-)

Anyway, @juanvallejo FYI ^^^^

roles/openshift_health_checker/test/mixins_test.py
rhcarvalho: > e.g. "is_containerized requests an undefined task_var, resulting in an exception that should be caught and re-raised as OpenShiftCheckException"

This is an implementation detail. Too much specificity makes the tests brittle.

The intention here is more on what the test name says: if we call `is_active` without proving `task_vars`, we expect an `OpenShiftCheckException` that talks about `'is_containerized'`.

I don't know what I had in mind when I wrote "missing config"... perhaps `...missing_task_vars` is better?

rhcarvalho:
aos-ci-test

rhcarvalho:
Only renamed a test after approval, so let's [merge] this.

rhcarvalho:
aos-ci-test

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
juanvallejo: > Hmm.. or.... would it make sense to read from the YAML files that defines the known values?

I think, if possible, this idea would make more sense than maintaining a list of values in one of the checks

PR #3747, opened 2017-03-23
Stop all services before upgrading openvswitch

sdodson:
It appears that openshift was holding a lock or perhaps openvswitch on
its own does not restart cleanly when upgrading the package so just
stop it before updating the package.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1420636

### DIFF
--- a/roles/openshift_node_upgrade/tasks/main.yml
+++ b/roles/openshift_node_upgrade/tasks/main.yml
@@ -51,24 +51,28 @@
   failed_when: false
   when: openshift.common.is_containerized | bool
 
+- name: Stop rpm based services
+  service:
+    name: "{{ item }}"
+    state: stopped
+  with_items:
+  - "{{ openshift.common.service_type }}-node"
+  - openvswitch
+  failed_when: false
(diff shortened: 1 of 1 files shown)


roles/openshift_node_upgrade/tasks/main.yml
```diff
@@ -51,6 +51,16 @@
...
 
+- name: Stop rpm based services
+  service:
+    name: "{{ item }}"
+    state: stopped
+  with_items:
+  - "{{ openshift.common.service_type }}-node"
+  - openvswitch
+  failed_when: false
+  when: not openshift.common.is_containerized | bool
```
sdodson: I was going to just remove the conditional on the previous task, but that would've meant that master services got restarted on rpm based installs but that's not necessary and they wouldn't be restarted. So just add a non containerized stop here.

sdodson:
QE has marked the bug as confirmed based on the backport to release-1.5 merging this.

sdodson:
aos-ci-test

sdodson:
[merge]

PR #3749, opened 2017-03-23
Generate wildcard router certificate

enoodle:
Conditioned on openshift_hosted_router_create_certificate=true.
Covers *.{{openshift_master_default_subdomain}},
in particular makes hawkular-metrics route have good cert out of the box.

resubmitting of https://github.com/openshift/openshift-ansible/pull/3226

### DIFF
--- a/roles/lib_openshift/library/oc_adm_ca_server_cert.py
+++ b/roles/lib_openshift/library/oc_adm_ca_server_cert.py
@@ -1453,9 +1453,12 @@ class CAServerCert(OpenShiftCLI):
                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
 
         x509output, _ = proc.communicate()
+
         if proc.returncode == 0:
             regex = re.compile(r"^\s*X509v3 Subject Alternative Name:\s*?\n\s*(.*)\s*\n", re.MULTILINE)
             match = regex.search(x509output)  # E501
+            if not match:
+                return False
             for entry in re.split(r", *", match.group(1)):
                 if entry.startswith('DNS') or entry.startswith('IP Address'):
                     cert_names.append(entry.split(':')[1])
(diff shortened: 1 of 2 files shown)


cben:
cc @mtnbikenc @detiber

enoodle:
@mtnbikenc @detiber can you please take a look?
@kwoodson I had to touch the ansible module to prevent it from crushing on me, can you talk a look please?

roles/lib_openshift/library/oc_adm_ca_server_cert.py
```diff
@@ -1453,9 +1453,12 @@ def exists(self):
                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
 
         x509output, _ = proc.communicate()
+
         if proc.returncode == 0:
             regex = re.compile(r"^\s*X509v3 Subject Alternative Name:\s*?\n\s*(.*)\s*\n", re.MULTILINE)
             match = regex.search(x509output)  # E501
+            if not match:
```
kwoodson: I made these changes under my branch.

enoodle:
@kwoodson I need this change to allow SSO to kibana from ManageIQ. In ManageIQ we ask the user to give us the main Openshift CA if he wan't to authenticate with it. If the router's certificates do not include a wildcard to use with the logging routes, ManageIQ won't be able to authenticate the router through it and we won't be able to securely initiate the SSO.

I would be happy for you to take over with your branch. you seem to know your way around here better than I do :)
Could you please check in your branch if you don't need something similar to [1] to use the certificates default location? (using the new facts that you created? :+1: ).

[1]https://github.com/enoodle/openshift-ansible/blob/e8aeeef35608b06cd92de9b2f88d6e3a84d5f60e/roles/openshift_hosted/tasks/router/router.yml#L72-L74

kwoodson:
@enoodle, I can take this over.

> Could you please check in your branch if you don't need something similar to [1] to use the certificates default location? (using the new facts that you created? :+1: ).

In theory the set_fact in the wildcard router task include sets the `openshift_hosted_router_certificates` that are used for the `openshift_hosted_routers` and thus this will wind up taking the `('/etc/origin/master/' ~ (item.certificates.certfile | basename)) ` which should result in what we want.

PR #3752, opened 2017-03-23
openshift_node: Use meta/main.yml for role dependencies

mtnbikenc:
If the role uses oc_* modules, lib_openshift should be added to the dependencies in meta/main.yml.

cc: @giuseppe

### DIFF
--- a/playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
+++ b/playbooks/common/openshift-cluster/redeploy-certificates/registry.yml
@@ -48,10 +48,6 @@
 
   # Replace dc/docker-registry certificate secret contents if set.
   - block:
-    - name: Load lib_openshift modules
-      include_role:
-        name: lib_openshift
-
     - name: Retrieve registry service IP
       oc_service:
         namespace: default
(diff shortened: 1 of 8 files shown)


mtnbikenc:
aos-ci-test

giuseppe:
@mtnbikenc I did the same mistake in `roles/openshift_master/tasks/system_container.yml`, I can fix that separately if you prefer

mtnbikenc:
aos-ci-test

mtnbikenc:
@sdodson any idea what's up with the bot comments here?

sdodson:
should be fixed now, but they're all green.

sdodson:
Is this how we should do this everywhere? does this pr fix all the instances?

sdodson:
roles/etcd/tasks/system_container.yml looks like the only other place you've missed

mtnbikenc:
@sdodson Role dependencies should be put in meta/main.yml.  In other words, don't use `include_role:` inside roles.  `include_role:` should still be used in `pre_tasks:` where the role has not yet been loaded.  I've updated the PR to include other incorrect uses.

sdodson:
[merge]

PR #3753, opened 2017-03-23
Increase rate limiting in journald.conf

soltysh:
@sdodson ptal, this is to address issues from https://github.com/openshift/origin/issues/12558

@smarterclayton @stevekuznetsov fyi

### DIFF
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -180,6 +180,28 @@
 - name: Install the systemd units
   include: systemd_units.yml
 
+- name: Checking for journald.conf
+  stat: path=/etc/systemd/journald.conf
+  register: journald_conf_file
+
+- name: Update journald setup
+  replace:
+    dest: /etc/systemd/journald.conf
+    regexp: '^(\#| )?{{ item.var }}=\s*.*?$'
+    replace: ' {{ item.var }}={{ item.val }}'
+    backup: yes
+  with_items: "{{ journald_vars_to_replace | default([]) }}"
+  when: journald_conf_file.stat.exists
+  register: journald_update
+
+# I need to restart journald immediatelly, otherwise it gets into way during
+# further steps in ansible
+- name: Restart journald
+  systemd:
+    name: systemd-journald
+    state: restarted
+  when: journald_update | changed
+
 - name: Install Master system container
   include: system_container.yml
   when:
--- a/roles/openshift_master/vars/main.yml
+++ b/roles/openshift_master/vars/main.yml
@@ -20,3 +20,22 @@ openshift_master_valid_grant_methods:
 - deny
 
 openshift_master_is_scaleup_host: False
+
+# These defaults assume forcing journald persistence, fsync to disk once
+# a second, rate-limiting to 10,000 logs a second, no forwarding to
+# syslog or wall, using 8GB of disk space maximum, using 10MB journal
+# files, keeping only a days worth of logs per journal file, and
+# retaining journal files no longer than a month.
+journald_vars_to_replace:
+- { var: Storage, val: persistent }
+- { var: Compress, val: yes }
+- { var: SyncIntervalSec, val: 1s }
+- { var: RateLimitInterval, val: 1s }
+- { var: RateLimitBurst, val: 10000 }
+- { var: SystemMaxUse, val: 8G }
+- { var: SystemKeepFree, val: 20% }
+- { var: SystemMaxFileSize, val: 10M }
+- { var: MaxRetentionSec, val: 1month }
+- { var: MaxFileSec, val: 1day }
+- { var: ForwardToSyslog, val: no }
+- { var: ForwardToWall, val: no }


sdodson:
https://bugzilla.redhat.com/show_bug.cgi?id=1380458 much discussion on this topic in this BZ

roles/openshift_master/tasks/main.yml
```diff
@@ -131,6 +131,23 @@
...
+  register: journald_conf_file
+
+- name: Increase rate limits in journald
+  lineinfile:
+    dest: /etc/systemd/journald.conf
+    regexp: "{{ item.regexp }}"
+    line: "{{ item.line }}"
+  with_items:
+    - { regexp: '^.*RateLimitInterval.*$', line: 'RateLimitInterval=1s' }
+    - { regexp: '^.*RateLimitBurst.*$', line: 'RateLimitBurst=10000' }
```
rhcarvalho:
```yaml
- { regexp: '^RateLimitBurst=.*$', line: 'RateLimitBurst=10000' }
```

would be more conservative?

What if the existing values were already higher than the ones we're setting? This could have an unintended effect of doing the opposite of what the task name says.

soltysh: > '^RateLimitBurst=.*$'

Isn't good enough, because that line might be commented out, but in that case I could add `\#?`. The remaining part is a good point, lemme think about the right approach to fix it.

sdodson:
So, normal logging default is 33/sec per service. We're cramming up to 1000 pods + node + docker into a single service. So should we go to 33333/sec to provide the same level of service that default logging would provide if they were all split out?

soltysh:
I've made the regexp more strict and fixed the travis failure. I'm not sure about that 33333/sec. Let's start with this setting and we'll increase more if needed. My ansible-foo isn't good enough to do the conditional change, sorry @rhcarvalho. I hope other stuff can be fixed as a followup.

smarterclayton:
Here's my concern.  ANY global rate limiting potentially breaks end user apps that expect to get logs.  NO global rate limiting potentially breaks the node.  Administrators need to make a choice.

The correct behavior would be that rate limiting is not global - journald is not in the correct position to determine rate limits, nor should logging from one container have anything to do with logging in another container.

The upstream Kubernetes direction on this is pretty clear - each container is going to log to its own rate limited devices.  That means separation of responsibilities from host logging - containers shouldn't be able to blow up host logs and vice versa, nor should containers be able to DOS other containers.

We need to get to a place where host logging != container logging as fast as possible.  We need the rate limits on the containers to be separate from the host rate limits.  We still need limits - but the limit should penalize the app, not vice versa.

smarterclayton:
Since rate limits are per service, the best we can do is increase Docker's rate limit a reasonable multiple of the expected volume.

smarterclayton:
We need to get the containers team and the logging team to present a unified plan forward here - @jcantrill and @rhatdan are the ones who can help coordinate.

roles/openshift_master/tasks/main.yml
```diff
@@ -131,6 +131,23 @@
...
+  stat: path=/etc/systemd/journald.conf
+  register: journald_conf_file
+
+- name: Increase rate limits in journald
+  lineinfile:
+    dest: /etc/systemd/journald.conf
+    regexp: "{{ item.regexp }}"
+    line: "{{ item.line }}"
+  with_items:
+    - { regexp: '^#?RateLimitInterval=.*$', line: 'RateLimitInterval=1s' }
```
portante: Are we also using persistent volumes?  We have the potential to rapidly grow `/run/log/journal` to the default 4GB limit.  May not be desirable.

portante:
There are only so many pipes for data to flow on a box.  At some point, no matter how isolated the rate-limits may be for an app, its logging rate, or the combined logging rate of all the apps, can affect the system.  Having individual rates is great, but it is going to be very hard to get that balance right to cover most situations so that host logging is not affected by app logging.

Before that, though, don't we need to stop using `/run/log/journal` (`tmpfs`) and start using `/var/log/journal` (on-disk)?

With a configuration that properly curates the on-disk journal files to a reasonable size and retention policy, using on-disk storage for logs will prevent large memory use of `/run`, while giving us the ability to review the host logs for debugging on a crash; something like:

```
[Journal]
Storage=permanent
Compress=yes

# journald will fsync() once a second to persist logs to disk
SyncIntervalSec=1s

# applies to each logging entity, not to all logs flowing through journald
RateLimitInterval=1s
RateLimitBurst=40000

# Only keep 4GB of logs max
SystemMaxUse=4G
# ... but only if there is 40 GB of free disk space
SystemKeepFree=40G
# Store journal files up to 10 MB each so that when rotated, we don't lose too big of a time period of logs.
SystemMaxFileSize=10M
```

smarterclayton:
The fact that these are all under the same journald service is the primary
problem.  If we can improve the logging driver to that point, most of the
immediate problem goes away.

Nodes should be logging to disk by default for sure.  However, the quota
needs to be per pod, not shared.

portante:
Regarding quota per pod, agreed.

Let's solve the logging to disk first.  Perhaps see if could have two instances of journald running, one that listens to `/dev/log` for node level logs, and another that the container daemon connects logs to from apps would be a start.

If the kubernetes logging path would consider including the metadata lost by NOT using journald, that might be sufficient.

portante:
And, when we start persisting logs to disk via journald, we should turn off rsyslog if it is on and persisting logs to disk as well.

richm:
> And, when we start persisting logs to disk via journald, we should turn off rsyslog if it is on and persisting logs to disk as well.

If so, will need to ensure that fluentd is configured to use journald as input, because no /var/log/messages any more.  Also want to make sure that customer is aware that other services provided by rsyslog by default will no longer be available:
```
# Log all the mail messages in one place.
mail.*                                                  -/var/log/maillog
# Log cron stuff
cron.*                                                  /var/log/cron
# Everybody gets emergency messages
*.emerg                                                 :omusrmsg:*
# Save news errors of level crit and higher in a special file.
uucp,news.crit                                          /var/log/spooler
# Save boot messages also to boot.log
local7.*                                                /var/log/boot.log
# Log cloudinit generated log messages to file
:programname, isequal, "cloud-init" /var/log/cloud-init.log
```

bogdando:
The alternative fix for the persistent logs issue could be never log to files or journald from containers and disable it for docker daemon as well. Instead, follow 12factor apps best practices:
* only log to stdout/stderr from containers. It implies there is only one subject of interests runs inside
  of each container and we don't want other logs from other background processes, if any (like dumb-init).
* ship and persist logs with fluentd/ES or the like.

This would remove a lots of "old school issues" and moving parts, like rate limiting, rotation et al.

rhcarvalho:
> only log to stdout/stderr from containers.

IIUC, this is already the case. Contained processes write to stdout/stderr, but then we need to do something with those streams, that's where journald or another solution plugs in.

portante:
@bogdando, does factor 11 of the [12 Factor App](https://12factor.net/logs) say an app should only write to `stdout`/`stderr`?  Actually, where does it mention `stderr`?

It seems that the 12 Factor App is about keeping the app writers concerns separate from the scaffolding of the environment in which the app runs.  Would you agree?

`journald` is just an entity that the docker deamon can be configured to send the `stdout`/`stderr` stream of an app to, where the app has no knowledge of or concern about `journald` or how it operates, which seems to fulfill the spirit of the 11th factor.

The 11th Factor seems to acknowledge that an app has other events streams flowing from it in production environments, as it mentions "In staging or production deploys, each process’ stream will be captured by the execution environment, collated together with all other streams from the app, and routed to one or more final destinations for viewing and long-term archival."

So it does not seem the 11th Factors directs us in a different path here.

Finally, if we don't use docker's `journald` driver, then we need to use the `json-file` driver, which we believe has its own set of scaling issues.

It would seem our best bet is to invest in `journald` to address the scaling we need.

roles/openshift_master/tasks/main.yml
portante: @soltysh, any thoughts here?

roles/openshift_master/tasks/main.yml
```diff
@@ -131,6 +131,23 @@
...
+    - { regexp: '^#?RateLimitBurst=.*$', line: 'RateLimitBurst=10000' }
```
portante: Also, this hardcodes the rate we want to be 10,000 log entries a second.  It would be best if the user could deploy and update this setting with a value they can configure to accommodate their needs.  In the future 10,000/s might be too small, while in other scenarios 10,000/s might be too large.

soltysh: I don't see any problems making it more configurable, I've opened this PR to have a discussion happening, since I've stumbled on this problem.

bogdando:
@portante @rhcarvalho You're right, either journald, or syslog, or fluentd et al for *docker daemon* would 100% fit the manifesto for its workloads: "A twelve-factor app never concerns itself with routing or storage of its output stream. It should not attempt to write to or manage logfiles. Instead, each running process writes its event stream, unbuffered, to stdout".

I was thinking of the docker daemon as a part of the application stack. Because of the shared local host placement with its workloads, it can not be scaled well, and the rate limiting issue confirms that, I'm afraid. The next issue to arise shortly after that might stand for cgroups not handling the limits, and more fun. So my previous comment was badly written, perhaps I should've written it as: Only non local to workloads logs processing solutions would make it to scale well and journald is not the case, unless it can be configured as a dummy logs shipper, like if being told: "please use no rate limiting, no logs rotation, no persistence, nothing, just ship the logs to a given network endpoint."

smarterclayton:
I'd like an option to disable rate limiting on the docker daemon.  We would use that in specific clusters.  I'm fine with higher limits, but the disable is what we need right now to get back to working.

I'd like a fix for this in soon so we can kill these flakes.  Having the option to selectively disable rate limiting on docker journald is the minimum requirement to address openshift/origin-ci-tool#133

smarterclayton:
Let's push this forward so we can turn it on in CI

portante:
What are we going to do about the persistence question?   `/run/log/journal` can blow up in its memory usage under certain conditions.

jcantrill:
cc @pweil-

jcantrill:
@portante With regards to storage, are we not at the mercy of what's available on disk for the host?  I don't believe journal runs in a container so there is no persistent volume to be configured.

jcantrill:
aos-ci-test

richm:
@jcantrill yes, we are at the mercy of what's available on the host.  When persistence is enabled in journald, it will by default use `/var/log/journal`.  Note that the `/var` or `/var/log` file systems usually have a lot of space dedicated to them, much more than the amount of RAM on the system (which is what `/run/log/journal` uses).

In addition, see man journald.conf(5), specifically the discussion around `SystemMaxUse=` and `RuntimeMaxUse=` and `SystemKeepFree=` and `RuntimeKeepFree=` - an excerpt:
```
...
The first pair defaults to 10% and the second to 15% of the size of
the respective file system, but each value is capped to 4G. If the
file system is nearly full and either SystemKeepFree= or
RuntimeKeepFree= are violated when systemd-journald is started, the
limit will be raised to the percentage that is actually free.
...
```

jcantrill:
@richm Are you saying we still have open questions regarding persistence or can we merge this PR?

richm:
@jcantrill I think what @portante is saying is that if you increase the rate limits, the very next thing you are likely to run into is that RAM gets full and records begin dropping from `/run/log/journal`, so increasing the rate limits should be accompanied by using `Storage=persistent`.  IMO, we should merge this PR first, then have another PR which configures journald to use `Storage=persistent` by default.

portante:
Does it really make sense to increase the rate-limiting without also addressing the use of an in-memory `tmpfs` file system, like `/run`?  Having them as two separate commits in separate PRs can lead to the rate-limiting cherry-picked without addressing the persistence issues, right?

soltysh:
I've applied changes as suggested by @portante, ptal

sdodson:
aos-ci-test

sdodson:
[test]

soltysh:
I've fixed yamllint complains. Let's see what we'll get now.

soltysh:
All the test failed on `Import Centos Image streams` task. 

[test]

sdodson:
[test]

soltysh:
I've tried setting up manually and I've stumbled upon exactly same error that the api server was restarted due to inactivity. The only culprit that I found close-by was the journald being restarted. So I've switched from using handler to a task restarting the service right after the configuration modifications. Let's see how this turns out working.

roles/openshift_master/tasks/main.yml
```diff
@@ -180,6 +180,29 @@
...
+
+- name: Update journald setup
+  replace:
+    dest: /etc/systemd/journald.conf
+    regexp: '^(\#| )?{{ item.var }}=\s*.*?$'
+    replace: ' {{ item.var }}={{ item.val }}'
+    backup: yes
+  with_items: "{{ journald_vars_to_replace | default([]) }}"
+  when:
+  - journald_conf_file.stat.exists
```
sdodson: `  register: journal_update`  to record whether or not the file got changed

roles/openshift_master/tasks/main.yml
```diff
@@ -180,6 +180,29 @@
...
+
+# I need to restart journald immediatelly, otherwise it gets into way during
+# further steps in ansible
+- name: Restart journald
+  systemd:
+    name: systemd-journald
+    state: restarted
+  when:
+  - journald_conf_file.stat.exists
```
sdodson: `when: journal_update | changed`

soltysh:
@sdodson comments addressed, now let's see the results.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

soltysh:
I had a typo in conditional check. Should be good now.

sdodson:
aos-ci-test

soltysh:
aos-ci-test

sdodson:
[test]

soltysh:
The failures in `fedora/25/atomic` look like flakes on k8s side of things. @sdodson I think we can safely merge this in, wdyt?

sdodson:
Yes, but since this isn't addressing a blocker bug I'd prefer waiting until Monday when the branch opens up.

soltysh:
@sdodson it is addressing a significant flake to our merge queue, where we fail because of missing data in journald. @smarterclayton any objections for merging this asap?

soltysh:
https://github.com/openshift/origin/issues/14785 for reference

soltysh:
The error doesn't look like it's related to this PR. It was reported long time ago (https://github.com/openshift/origin/issues/9624 and https://github.com/openshift/origin/issues/7291) and closed due to inactivity. So it might be F25 specific issue.

sdodson:
The first time running the install since this merged it failed to restart journald on my host and I got the following in my logs. By the time I looked at the host it had started successfully but the playbook run failed.

```
Sep 20 16:17:13 ose3-master ansible-systemd: Invoked with no_block=False name=systemd-journald enabled=None daemon_reload=False state=restarted user=False masked=None
Sep 20 16:17:13 ose3-master systemd: Stopping Flush Journal to Persistent Storage...
Sep 20 16:17:13 ose3-master journal: Journal stopped
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master systemd: Stopping Flush Journal to Persistent Storage...
Sep 20 16:17:13 ose3-master journal: Journal stopped
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master systemd: Stopping Flush Journal to Persistent Storage...
Sep 20 16:17:13 ose3-master journal: Journal stopped
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:13 ose3-master journal: Permanent journal is using 8.0M (max allowed 8.0G, trying to leave 1.4G free of 2.5G available → current limit 1.0G).
Sep 20 16:17:15 ose3-master journal: Permanent journal is using 176.0M (max allowed 8.0G, trying to leave 1.4G free of 2.3G available → current limit 866.9M).
Sep 20 16:17:15 ose3-master systemd-journald[460]: Received SIGTERM from PID 1 (systemd).
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to create new runtime journal: No such file or directory
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Failed to set file attributes: Operation not supported
Sep 20 16:17:15 ose3-master systemd-journald[10177]: Assertion 'f' failed at src/journal/journal-file.c:132, function journal_file_close(). Aborting.
Sep 20 16:17:15 ose3-master journal: Journal started
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
Sep 20 16:17:15 ose3-master rsyslogd: imjournal: journal reloaded... [v8.24.0 try http://www.rsyslog.com/e/0 ]
```

soltysh:
I did a bit of googling and it looks like there are 2 required actions for this to work properly:
1. create the directory and
2. make sure it has the appropriate selinux context set. 
I think we're failing the 2nd step. Let's sync when you're online.

roles/openshift_master/tasks/main.yml
```diff
@@ -180,6 +180,28 @@
 - name: Install the systemd units
   include: systemd_units.yml
 
+- name: Checking for journald.conf
+  stat: path=/etc/systemd/journald.conf
+  register: journald_conf_file
+
+- name: Update journald setup
+  replace:
```
dlbewley: Nice timing, @soltysh  I was just looking into this problem in my env.

Would it have been more clean to drop a templated file into `/etc/systemd/journald.conf.d` instead?

soltysh: @dlbewley not sure, I'd defer to @sdodson as the main ansible installer maintainer.

sdodson: @dlbewley Yeah, that's probably a nice improvement. PR for that ? :-) Given this just merged I wouldn't worry about attempting to migrate anyone from how this PR modified things to the dropin file

sdodson: Other drop ins we've named 'openshift-ansible.conf' FWIW

dlbewley: i noticed that in `dnsmasq.d`, @sdodson i'll try to find some time to do that PR. don't hold your breath :)

PR #3757, opened 2017-03-23
skip "yum update" sanity check if not running on rhel

juanvallejo:
Related Trello card: https://trello.com/c/5pr2WcQH/43-diagnostics-additional-pre-flight-checks
Detect ansible host's OS using output of `docker info` and skip check if it is not being run on RHEL.

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/package_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/package_availability.py
@@ -9,6 +9,10 @@ class PackageAvailability(NotContainerizedMixin, OpenShiftCheck):
     name = "package_availability"
     tags = ["preflight"]
 
+    @classmethod
+    def is_active(cls, task_vars):
+        return super(PackageAvailability, cls).is_active(task_vars) and task_vars["ansible_pkg_mgr"] == "yum"
+
     def run(self, tmp, task_vars):
         rpm_prefix = get_var(task_vars, "openshift", "common", "service_type")
         group_names = get_var(task_vars, "group_names", default=[])
--- a/roles/openshift_health_checker/test/package_availability_test.py
+++ b/roles/openshift_health_checker/test/package_availability_test.py
@@ -3,6 +3,20 @@ import pytest
 from openshift_checks.package_availability import PackageAvailability
 
 
+@pytest.mark.parametrize('pkg_mgr,is_containerized,is_active', [
+    ('yum', False, True),
+    ('yum', True, False),
+    ('dnf', True, False),
+    ('dnf', False, False),
+])
+def test_is_active(pkg_mgr, is_containerized, is_active):
+    task_vars = dict(
+        ansible_pkg_mgr=pkg_mgr,
+        openshift=dict(common=dict(is_containerized=is_containerized)),
+    )
+    assert PackageAvailability.is_active(task_vars=task_vars) == is_active
+
+
 @pytest.mark.parametrize('task_vars,must_have_packages,must_not_have_packages', [
     (
         dict(openshift=dict(common=dict(service_type='openshift'))),


roles/openshift_health_checker/openshift_checks/package_update.py
```diff
@@ -9,6 +9,24 @@ class PackageUpdate(NotContainerizedMixin, OpenShiftCheck):
...
-        return self.module_executor("check_yum_update", args, tmp, task_vars)
+
+        os = self.get_operating_system(task_vars)
+        if os and self.is_os_supported(os):
+            return self.module_executor("check_yum_update", args, tmp, task_vars)
+
+        return {"changed": False, "msg": "Skipping check on unsupported OS: {os}".format(os=os)}
+
+    def get_operating_system(self, task_vars):
+        info = self.module_executor("docker_info", {}, task_vars)
```
juanvallejo: @sosiouxme per your comment [on the Trello card](https://trello.com/c/5pr2WcQH/43-diagnostics-additional-pre-flight-checks#comment-58d42e29cd83526685047ea1), since this is a pre-flight check, perhaps it would be better to use something like `cat /proc/os-release` on the ansible host?

rhcarvalho: We don't need to call any module. The operating system / package manager is readily available in `task_vars`, provided by http://docs.ansible.com/ansible/setup_module.html

rhcarvalho:
```console
$ ansible -i hosts -m setup master1 | grep yum
        "ansible_pkg_mgr": "yum", 
```

All we need to do is implement `is_active` to look at `task_vars['ansible_pkg_mgr']`.

rhcarvalho:
@juanvallejo note that this is not about "not running on RHEL", but about the package manager... e.g., CentOS also uses "yum" and is perfectly compatible with the existing check.

roles/openshift_health_checker/openshift_checks/mixins.py
```diff
@@ -10,9 +10,13 @@ def is_active(cls, task_vars):
         return (
             # This mixin is meant to be used with subclasses of OpenShiftCheck.
             super(NotContainerizedMixin, cls).is_active(task_vars) and
-            not cls.is_containerized(task_vars)
+            not cls.is_containerized(task_vars) and cls.has_yum_pkg_mgr(task_vars)
```
sosiouxme: Shouldn't this be in a separate mixin, like say:

```
class NeedsYumMixin(object):
     """Mixin for checks that are only active for hosts where yum package manager is present."""
 
     @classmethod
     def is_active(cls, task_vars):
          # This mixin is meant to be used with subclasses of OpenShiftCheck.
          return super(NeedsYumMixin, cls).is_active(task_vars) and cls.has_yum_pkg_mgr(task_vars)

Comment (rhcarvalho) [inline roles/openshift_health_checker/openshift_checks/mixins.py]: Certainly

File: roles/openshift_health_checker/openshift_checks/package_availability.py [hunk 3]
```diff
@@ -1,9 +1,9 @@
 # pylint: disable=missing-docstring
 from openshift_checks import OpenShiftCheck, get_var
-from openshift_checks.mixins import NotContainerizedMixin
+from openshift_checks.mixins import NotContainerizedMixin, NeedsYumMixin
 
 
-class PackageAvailability(NotContainerizedMixin, OpenShiftCheck):
+class PackageAvailability(NotContainerizedMixin, NeedsYumMixin, OpenShiftCheck):
```
stevekuznetsov: On Fedora you can technically still use `yum` but instead of ignoring the check shouldn't you just use `dnf`?

rhcarvalho: This is about having the Python package `yum` available/installed. Not that we couldn't support DNF, just that we don't have code for it at the moment.

This PR makes the check be skipped instead of giving an spurious error.

roles/openshift_health_checker/openshift_checks/mixins.py
```diff
@@ -16,3 +16,23 @@ def is_active(cls, task_vars):
...
+
+class NeedsYumMixin(object):
+    """Mixin for checks that are only active when 'yum' is available as the package manager on the host."""
+
+    @classmethod
+    def is_active(cls, task_vars):
+        return (
+            # This mixin is meant to be used with subclasses of OpenShiftCheck.
+            super(NeedsYumMixin, cls).is_active(task_vars) and
+            not cls.is_containerized(task_vars) and cls.has_yum_pkg_mgr(task_vars)
```
rhcarvalho: @juanvallejo this mixin should do one thing -- it has nothing to do with 'containerized'.
We can remove this part of the condition here and remove the `is_containerized` method.

Note how we use the two mixins together to compose their effects:

```python
class PackageAvailability(NotContainerizedMixin, NeedsYumMixin, OpenShiftCheck):
   ...
```

roles/openshift_health_checker/openshift_checks/mixins.py
```diff
@@ -16,3 +16,23 @@ def is_active(cls, task_vars):
...
+        )
+
+    @staticmethod
+    def is_containerized(task_vars):
+        return get_var(task_vars, "openshift", "common", "is_containerized")
+
+    @staticmethod
+    def has_yum_pkg_mgr(task_vars):
+        return task_vars["ansible_pkg_mgr"] == "yum"
```
rhcarvalho: We could simply inline this:

```python
@classmethod
def is_active(cls, task_vars):
    return super(NeedsYumMixin, cls).is_active(task_vars) and task_vars["ansible_pkg_mgr"] == "yum"
```

rhcarvalho:
@juanvallejo this could be done similar to #3887.

roles/openshift_health_checker/test/package_availability_test.py
```diff
@@ -3,6 +3,21 @@
 from openshift_checks.package_availability import PackageAvailability
 
 
+@pytest.mark.parametrize('pkg_mgr,is_containerized,is_active', [
+    ('yum', False, True),
+    ('yum', True, False),
+    ('dnf', False, False),
+    ('dnf', False, False),
```
rhcarvalho: The two lines above are identical :-)

roles/openshift_health_checker/test/package_availability_test.py
```diff
@@ -3,6 +3,21 @@
...
+])
+def test_is_active(pkg_mgr, is_containerized, is_active):
+    task_vars = dict(
+        ansible_pkg_mgr=pkg_mgr,
+        group_names=["masters", "nodes"],
```
rhcarvalho: This is irrelevant to determine `is_active`, right? If so, we can have one less line to read ;)

juanvallejo:

rhcarvalho:
aos-ci-test

rhcarvalho:
[merge]

rhcarvalho:
Flake https://github.com/openshift/origin/issues/13731.

[merge] again

PR #3759, opened 2017-03-23
Specify enterprise defaults for logging and metrics images

sdodson:
picks #3760


sdodson:
@ewolinetz @jcantrill ptal

sdodson:
My only concern is if we ever call something outside of main.yml directly this doesn't work.

PR #3760, opened 2017-03-23
Specify enterprise defaults for logging and metrics images

sdodson:


sdodson:
aos-ci-test

roles/openshift_logging/tasks/generate_deploymentconfigs.yaml
```diff
@@ -4,8 +4,8 @@
   vars:
```
ewolinetz: this file really needs to be deleted, im pretty sure it isn't called at all.... TODO

roles/openshift_logging/tasks/main.yaml
```diff
@@ -3,11 +3,17 @@
...
   when: "{{ openshift_logging_fluentd_nodeselector.keys() | count }} > 1"
 
-- name: Set default image variables based on deployment_type
-  include_vars: "{{ item }}"
-  with_first_found:
-    - "{{ openshift_deployment_type | default(deployment_type) }}.yml"
-    - "default_images.yml"
+- name: Set image config defaults
+  set_fact:
+    logging_image_prefix: "{{ openshift_logging_image_prefix | openshift_hosted_logging_deployer_prefix | default('docker.io/openshift/origin-') }}"
```
ewolinetz: this seems like it could work. I was thinking more that since we'd keep the first found but use the 'private variable' idea where we either use `openshift_logging_image_prefix` if its provided otherwise we use `__openshift_logging_image_prefix`

sdodson: Ok, refactored things considerably, what do you think?

roles/openshift_metrics/templates/pvc.j2
```diff
@@ -4,7 +4,7 @@ metadata:
   name: "{{obj_name}}"
 {% if labels is not defined %}
   labels:
-    logging-infra: support
+    metrics-infra: support
```
ewolinetz: :+1:

sdodson:
aos-ci-test

sdodson:
[merge]

PR #3764, opened 2017-03-24
Fix quoting

lukas-vlcek:
Back-porting https://github.com/openshift/openshift-ansible/pull/3763
@ewolinetz @richm PTAL

PR #3766, opened 2017-03-24
fix quoting issues in logging es template

richm:
make sure double quotes are used properly


ewolinetz:
using this in favor of https://github.com/openshift/openshift-ansible/pull/3764

PR #3769, opened 2017-03-24
Version comparison filters

abutcher:
Usage:

```
{{ '1.4' | oo_version_gte_3_3_or_1_3('origin') }}
```

```
{{ openshift_version | oo_version_gte_3_3_or_1_3(openshift.common.deployment_type) }}
```

Starting with `3.6`, there is no need to test against the `deployment_type`.

```
{{ '3.7' | oo_version_gte_3_6 }}
```

```
{{ openshift_version | oo_version_gte_3_6 }}
```

### DIFF
--- a/filter_plugins/openshift_version.py
+++ b/filter_plugins/openshift_version.py
@@ -1,0 +1,98 @@
+#!/usr/bin/python
+
+# -*- coding: utf-8 -*-
+# vim: expandtab:tabstop=4:shiftwidth=4
+"""
+Custom version comparison filters for use in openshift-ansible
+"""
+
+# pylint can't locate distutils.version within virtualenv
+# https://github.com/PyCQA/pylint/issues/73
+# pylint: disable=no-name-in-module, import-error
+from distutils.version import LooseVersion
+
+
+def gte_function_builder(name, versions):
+    """
+    Build and return a version comparison function.
+
+    Ex: name = 'oo_version_gte_3_1_or_1_1'
+        versions = {'enterprise': '3.1', 'origin': '1.1'}
+
+        returns oo_version_gte_3_1_or_1_1, a function which based on the
+        version and deployment type will return true if the provided
+        version is greater than or equal to the function's version
+    """
+    enterprise_version = versions['enterprise']
+    origin_version = versions['origin']
+
+    def _gte_function(version, deployment_type):
+        """
+        Dynamic function created by gte_function_builder.
+
+        Ex: version = '3.1'
+            deployment_type = 'openshift-enterprise'
+            returns True/False
+        """
+        version_gte = False
+        if 'enterprise' in deployment_type:
+            if str(version) >= LooseVersion(enterprise_version):
+                version_gte = True
+        elif 'origin' in deployment_type:
+            if str(version) >= LooseVersion(origin_version):
+                version_gte = True
+        return version_gte
+    _gte_function.__name__ = name
+    return _gte_function
+
+
+# pylint: disable=too-few-public-methods
+class FilterModule(object):
+    """
+    Filters for version checking.
+    """
+    #: The major versions to start incrementing. (enterprise, origin)
+    majors = [(3, 1)]
+
+    #: The minor version to start incrementing
+    minor = 3
+    #: The number of iterations to increment
+    minor_iterations = 10
+
+    def __init__(self):
+        """
+        Creates a new FilterModule for ose version checking.
+        """
+        self._filters = {}
+        # For each major version
+        for enterprise, origin in self.majors:
+            # For each minor version in the range
+            for minor in range(self.minor, self.minor_iterations):
+                # Create the function name
+                func_name = 'oo_version_gte_{}_{}_or_{}_{}'.format(
+                    enterprise, minor, origin, minor)
+                # Create the function with the builder
+                func = gte_function_builder(
+                    func_name, {
+                        'enterprise': '{}.{}.0'.format(enterprise, minor),
+                        'origin': '{}.{}.0'.format(origin, minor)
+                    })
+                # Add the function to the mapping
+                self._filters[func_name] = func
+
+        # Create filters with special versioning requirements
+        self._filters['oo_version_gte_3_1_or_1_1'] = gte_function_builder('oo_version_gte_3_1_or_1_1',
+                                                                          {'enterprise': '3.0.2.905',
+                                                                           'origin': '1.1.0'})
+        self._filters['oo_version_gte_3_1_1_or_1_1_1'] = gte_function_builder('oo_version_gte_3_1_or_1_1',
+                                                                              {'enterprise': '3.1.1',
+                                                                               'origin': '1.1.1'})
+        self._filters['oo_version_gte_3_2_or_1_2'] = gte_function_builder('oo_version_gte_3_2_or_1_2',
+                                                                          {'enterprise': '3.1.1.901',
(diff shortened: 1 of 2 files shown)


abutcher:
@php-coder Thinking something like this for `openshift_node_certificates` and possibly broader use since `openshift_version` and `openshift_deployment_type` or `openshift.common.deployment_type` are available.

filter_plugins/openshift_version.py
```diff
@@ -0,0 +1,98 @@
...
+# pylint: disable=no-name-in-module, import-error
+from distutils.version import LooseVersion
+
+
+def gte_function_builder(name, versions):
+    """
+    Build and return a version comparison function.
+
+    Ex: name = 'oo_version_gte_3_1_or_1_1'
+        versions = {'enterprise': '3.1', 'origin': '1.1'}
```
dgoodwin: FWIW per Clayton's email to openshift-dev recently, it looks like origin is adopting the 3.x versioning as well, which will be a nice improvement for stuff like this.

filter_plugins/openshift_version.py
```diff
@@ -0,0 +1,98 @@
...
+        # Create filters with special versioning requirements
+        self._filters['oo_version_gte_3_1_or_1_1'] = gte_function_builder('oo_version_gte_3_1_or_1_1',
+                                                                          {'enterprise': '3.0.2.905',
+                                                                           'origin': '1.1.0'})
+        self._filters['oo_version_gte_3_1_1_or_1_1_1'] = gte_function_builder('oo_version_gte_3_1_or_1_1',
+                                                                              {'enterprise': '3.1.1',
+                                                                               'origin': '1.1.1'})
+        self._filters['oo_version_gte_3_2_or_1_2'] = gte_function_builder('oo_version_gte_3_2_or_1_2',
+                                                                          {'enterprise': '3.1.1.901',
+                                                                           'origin': '1.2.0'})
```
dgoodwin: Do we actually need to keep filters around this far back? I know there's been talk of multi-version upgrades and master going back to supporting versions way back, I assume that's a go and the reason for this?

filter_plugins/openshift_version.py
```diff
@@ -0,0 +1,98 @@
...
+                func_name = 'oo_version_gte_{}_{}_or_{}_{}'.format(
+                    enterprise, minor, origin, minor)
+                # Create the function with the builder
+                func = gte_function_builder(
+                    func_name, {
+                        'enterprise': '{}.{}.0'.format(enterprise, minor),
+                        'origin': '{}.{}.0'.format(origin, minor)
+                    })
+                # Add the function to the mapping
+                self._filters[func_name] = func
```
dgoodwin: Is there time to get a little unit test module plugged in somewhere that nose would pick up? This module looks very easy to test, would be nice to see proof it's picking up all the combinations expected.

filter_plugins/openshift_version.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    Filters for version checking.
+    """
+    #: The major versions to start incrementing. (enterprise, origin)
+    majors = [(3, 1)]
+
+    #: The minor version to start incrementing
+    minor = 3
+    #: The number of iterations to increment
+    minor_iterations = 10
```
dgoodwin: Per the new versioning mentioned earlier, that's coming into effect for 3.6, I'd be tempted to make sure we don't let a oo_version_gte_3_6_or_1_6 ever be available. Could we either adapt this for 3.6+ or maybe just shut down the minor iterations at 5.

abutcher: Ooo I'll see if I can work that in

filter_plugins/openshift_version.py
abutcher: Also gonna work on this

filter_plugins/openshift_version.py
abutcher: Not sure really, I was just duplicating the functionality from the `openshift.common.version_gte*` facts.

tbielawa:
Unit tests and or a handfull of examples using existing tasks (links to original code appreciated) that show how to replace existing comparison with the new comparison filters.

I think this a cool thing. However, if we accept it we might break uniformity in version checking across the code base. If this is a drop-in replacement for our current comparison routines then I'd consider it *accepted technical debt*. I'd say we should create a card/issue to manage and track the technical debt as we audit the code base and replace the old comparison routines with the new ones (so that we are doing things uniformly again).

ashcrow:
Still looks good to me with the changes (once the test failure is fixed).

ashcrow:
@tbielawa part of this change is that the filters will be generated instead of manually hardcoding each time. The hardcoding you see is primarily the legacy versions which were manually being added (or accidentally forgotten about) in the past :-)

In terms of usage, I believe it's exactly the same as before ... but I'm :+1: on 'how to use; if there hasn't ever been docs on that.

filter_plugins/openshift_version.py
```diff
@@ -46,50 +46,82 @@ def _gte_function(version, deployment_type):
...
-                                                                           'origin': '1.1.0'})
-        self._filters['oo_version_gte_3_1_1_or_1_1_1'] = gte_function_builder('oo_version_gte_3_1_or_1_1',
-                                                                              {'enterprise': '3.1.1',
-                                                                               'origin': '1.1.1'})
-        self._filters['oo_version_gte_3_2_or_1_2'] = gte_function_builder('oo_version_gte_3_2_or_1_2',
-                                                                          {'enterprise': '3.1.1.901',
-                                                                           'origin': '1.2.0'})
+        # Create filters with special versioning requirements.
+        # Treat all Origin 1.x as special case.
+        legacy_filters = [{'name': 'oo_version_gte_3_1_or_1_1',
```
tbielawa: I am unsure about something: is this **data** we could extract from the **code**? Is it possible that we could update an external data source (some kind of serialized file, json, yaml, w/e) and read from that instead? Would that be too difficult/out of scope of this PR?

I bring this up because I'm a fan of separating data from code. You really shouldn't have to update code files to update data in an *ideal* situation. I understand if that's not something we can do right now though. Ansible is its own beast.

Same comment applies to `test/openshift_version_tests.py:OpenShiftVersionTests.legacy_gte_tests`.

ashcrow: IMHO the legacy stuff is static at this point, but the ``major``, ``minor`` iteration could possibly be keyed off something. :+1:

Not sure if that would be within this scope though.

tbielawa: re @ashcrow yeah, if those hard-coded ones are just for legacy stuff and won't be subject to change, then ignore my comment. I was thinking of the old way we did things.

abutcher: Yeah, these will be static. It seemed kind of silly to add a loop for the three versions that could be iterated over in these legacy filters.

abutcher:
@tbielawa These filters are all new but are intended to be replacements for [these facts](https://github.com/openshift/openshift-ansible/blob/2e37c702c8d4c3778d172d8e886f68e23d4aeebc/roles/openshift_facts/library/openshift_facts.py#L882-L919). I'd like to not replace all of the current usage (at least initially) and just allow these to be used moving forward. We have a card related to untangling `openshift_facts` and I think these filters work towards that for version comparison facts. Replacement could be done as a part of that card.

abutcher:
aos-ci-test

abutcher:
[merge]

PR #3776, opened 2017-03-27
Updated metrics role to work with HA masters

dak1n1:
This solves https://bugzilla.redhat.com/show_bug.cgi?id=1436147

### DIFF
--- a/roles/openshift_metrics/README.md
+++ b/roles/openshift_metrics/README.md
@@ -90,9 +90,10 @@ Example Playbook
 
 ```
 - name: Configure openshift-metrics
-  hosts: oo_first_master
+  hosts: oo_masters_to_config
   roles:
   - role: openshift_metrics
+  run_once: True
 ```
 
 License
--- a/roles/openshift_metrics/tasks/install_metrics.yaml
+++ b/roles/openshift_metrics/tasks/install_metrics.yaml
@@ -1,5 +1,6 @@
 ---
 - include: pre_install.yaml
+  run_once: True
 
 - name: Install Metrics
   include: "{{ role_path }}/tasks/install_{{ include_file }}.yaml"
@@ -11,19 +12,23 @@
   loop_control:
     loop_var: include_file
   when: "not {{ openshift_metrics_heapster_standalone | bool }}"
+  run_once: True
 
 - name: Install Heapster Standalone
   include: install_heapster.yaml
   when: "{{ openshift_metrics_heapster_standalone | bool }}"
+  run_once: True
 
 - find: paths={{ mktemp.stdout }}/templates patterns=*.yaml
   register: object_def_files
   changed_when: no
+  run_once: True
 
 - slurp: src={{item.path}}
   register: object_defs
   with_items: "{{object_def_files.files}}"
   changed_when: no
+  run_once: True
 
 - name: Create objects
   include: oc_apply.yaml
@@ -33,8 +38,10 @@
     file_name: "{{ item.source }}"
     file_content: "{{ item.content | b64decode | from_yaml }}"
   with_items: "{{ object_defs.results }}"
+  run_once: True
 
 - include: update_master_config.yaml
+  run_once: False
 
 - command: >
     {{openshift.common.client_binary}}
@@ -45,13 +52,16 @@
     -n {{openshift_metrics_project}}
   register: existing_metrics_rc
   changed_when: no
+  run_once: True
 
 - name: Scaling down cluster to recognize changes
   include: stop_metrics.yaml
   when: "{{ existing_metrics_rc.stdout_lines | length > 0 }}"
+  run_once: True
 
 - name: Scaling up cluster
   include: start_metrics.yaml
   tags: openshift_metrics_start_cluster
   when:
     - openshift_metrics_start_cluster | default(true) | bool
+  run_once: True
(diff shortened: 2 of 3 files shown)


sdodson:
@ewolinetz i thought we fixed this somewhere else? Maybe we need to pick something back to master?

ewolinetz:
@dak1n1 how are you using the role? We currently have it configured to run from the byo using the host `oo_first_master` [1]. The intent was that it would be run against a single master.


[1] https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/openshift_hosted.yml#L14

dak1n1:
The problem with using `oo_first_master` is that only the first master gets `metricsPublicURL` set. That causes an issue for clusters running more than one master.

dak1n1:
This is how I'm calling the role, where `aos_3.5_roles` is a symlink to a vendored rpm of openshift-ansible-3.5.40.

```
  - name: Install OpenShift Metrics
    include_role:
      name: aos_3.5_roles/openshift_metrics
    vars:
      openshift_metrics_hawkular_hostname: "metrics.{{ cli_clusterid }}.openshift.com"
      openshift_metrics_project: openshift-infra
      openshift_metrics_image_prefix: registry.ops.openshift.com/openshift3/
      openshift_metrics_image_version: "{{ g_metrics_image_version }}"
      openshift_metrics_hawkular_cert: "{{ scf_retval_openshift_cert }}"
      openshift_metrics_hawkular_key: "{{ scf_retval_openshift_key }}"
      openshift_metrics_hawkular_ca: "{{ scf_retval_chain_cert }}"
      openshift_metrics_hawkular_replicas: 1
      openshift_metrics_cassandra_replicas: 2
      openshift_metrics_cassandra_storage_type: dynamic
      openshift_metrics_cassandra_pvc_size: 100
      openshift_metrics_heapster_standalone: false
      openshift_metrics_startup_timeout: 500
      openshift_metrics_duration: 7
      openshift_metrics_resolution: 30s
      openshift_metrics_cassandra_limits_memory: 2G
      openshift_metrics_cassandra_requests_memory: 1G
      openshift_metrics_hawkular_limits_memory: 3G
      openshift_metrics_hawkular_requests_memory: 2G
      openshift_metrics_heapster_limits_memory: 3.75G
      openshift_metrics_heapster_requests_memory: 0.9375G
      openshift_metrics_hawkular_requests_cpu: 100
    run_once: True
```

For me, this is working in testing. The URL shows up on all 3 of my masters, while the other tasks are run only once.

ewolinetz:
> The problem with using `oo_first_master` is that only the first master gets `metricsPublicURL` set. That causes an issue for clusters running more than one master.

Hmm, this should be setting that for every master with the byo playbook:
https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/openshift_hosted.yml#L49-L70

@sdodson which is the preferred method? I personally like the `run_once: true` pattern and then we can let this run against any master. The only downside is for our cert generation, to not regenerate certificates we would want to point at the same master each time...

sdodson:
Lets talk about this approach and the general pattern of supporting our roles outside of the context of openshift-ansible during this week's arch call. I'm not sure how far we want to take this. To me this is blurring the lines of a role and a playbook a bit.

roles/openshift_metrics/tasks/install_metrics.yaml
```diff
@@ -11,19 +12,23 @@
   loop_control:
     loop_var: include_file
   when: "not {{ openshift_metrics_heapster_standalone | bool }}"
+  run_once: True
 
 - name: Install Heapster Standalone
   include: install_heapster.yaml
   when: "{{ openshift_metrics_heapster_standalone | bool }}"
+  run_once: True
```
mtnbikenc: This will apply to every task in install_heapster.yaml, which makes the run_once: True on generate_heapster_certificates.yaml redundant.  Additionally every task in that include will be run_once, which it looks like is not desired.

roles/openshift_metrics/README.md
```diff
@@ -90,9 +90,10 @@ Example Playbook
 
 ```
 - name: Configure openshift-metrics
-  hosts: oo_first_master
   roles:
   - role: openshift_metrics
+  run_once: True
```
mtnbikenc: I'm not a fan of using run_once at the play level as shown in the example in the README.  The run_once will be applied to **every task** in the entire play (and role tasks) which may not be desired outcome.  However, including run_once at the Play level superceeds the need to add it to each task in the role task file as shown.  Only the task which needs run_once: False should be updated.  run_once: True should be either at the Play/Role level or at the task (or include) level.

So, I don't like the idea of a play/role needing to be called with a run_once in order for it to operate properly.

sdodson: From the arch call we were all in agreement, this should be removed.

roles/openshift_metrics/tasks/install_metrics.yaml
sdodson: I think reading through install_heapster.yaml and generate_heapster_certificates.yaml they should all be run_once: true so we should remove it within install_heapster.yaml.

@jcantrill can you confirm?

jcantrill: @sdodson I would agree we only need to generate certificates and run 'installing heapster' once.  I think removing it from the calling task is correct.

dak1n1:
I'm going to close this since @mwoodson is now driving our metrics changes, so I believe this is no longer needed.

PR #3770, opened 2017-03-24
fix es config merge so template does not need quoting. gen then merge

jcantrill:
This PR fixes

* Merge failures for es config for values that were not quoted in the jinja template
* Executes template task before merging hash

PR #3777, opened 2017-03-27
bug 1432607.  Allow configuration of ES log destination

jcantrill:
cc @portante
This PR:

* turns console logging off by default.
* allows users to turn it back on via ansible inventory
* encompasses #3687  which should be closed.

### DIFF
--- a/roles/openshift_logging/tasks/generate_configmaps.yaml
+++ b/roles/openshift_logging/tasks/generate_configmaps.yaml
@@ -1,10 +1,20 @@
 ---
 - block:
-    - copy:
-        src: elasticsearch-logging.yml
+    - fail:
+        msg: "The openshift_logging_es_log_appenders '{{openshift_logging_es_log_appenders}}' has an unrecognized option and only supports the following as a list: {{es_log_appenders | join(', ')}}"
+      when:
+        - es_logging_contents is undefined
+        - "{{ openshift_logging_es_log_appenders | list | difference(es_log_appenders) | length != 0 }}"
+      changed_when: no
+
+    - template:
+        src: elasticsearch-logging.yml.j2
         dest: "{{mktemp.stdout}}/elasticsearch-logging.yml"
+      vars:
+        root_logger: "{{openshift_logging_es_log_appenders | join(', ')}}"
       when: es_logging_contents is undefined
       changed_when: no
+      check_mode: no
 
     - local_action: >
         copy content="{{ config_source | combine(override_config,recursive=True) | to_nice_yaml }}"
--- a/roles/openshift_logging/templates/elasticsearch-logging.yml.j2
+++ b/roles/openshift_logging/templates/elasticsearch-logging.yml.j2
@@ -1,13 +1,24 @@
 # you can override this using by setting a system property, for example -Des.logger.level=DEBUG
 es.logger.level: INFO
-rootLogger: ${es.logger.level}, console, file
+rootLogger: ${es.logger.level}, {{root_logger}}
 logger:
   # log action execution errors for easier debugging
   action: WARN
+
+  # deprecation logging, turn to DEBUG to see them
+  deprecation: WARN, deprecation_log_file
+
   # reduce the logging for aws, too much is logged under the default INFO
   com.amazonaws: WARN
+
   io.fabric8.elasticsearch: ${PLUGIN_LOGLEVEL}
   io.fabric8.kubernetes: ${PLUGIN_LOGLEVEL}
+
+  # aws will try to do some sketchy JMX stuff, but its not needed.
+  com.amazonaws.jmx.SdkMBeanRegistrySupport: ERROR
+  com.amazonaws.metrics.AwsSdkMetrics: ERROR
+
+  org.apache.http: INFO
 
   # gateway
   #gateway: DEBUG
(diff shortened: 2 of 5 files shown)


jcantrill:
Blocked on #3770 because of merging es config issues.

roles/openshift_logging/templates/elasticsearch.yml.j2
```diff
@@ -6,8 +6,6 @@ script:
   indexed: on
 
 index:
-  number_of_shards: {{ es_number_of_shards | default ('1') }}
```
portante: Why are we dropping these?

jcantrill: We should be dropping these... this file should not be in the PR.  was work around for #3770

jcantrill: fixed and pushed

roles/openshift_logging/tasks/generate_configmaps.yaml
```diff
@@ -1,10 +1,20 @@
 ---
 - block:
-    - copy:
-        src: elasticsearch-logging.yml
+    - fail:
```
ewolinetz: can we move this fail to the top of `install_logging` instead?
we also want to verify that a valid value was provided

roles/openshift_logging/templates/elasticsearch-logging.yml.j2
```diff
@@ -28,13 +39,14 @@ logger:
 additivity:
   index.search.slowlog: false
   index.indexing.slowlog: false
+  deprecation: false
 
 appender:
```
ewolinetz: don't we also need to have a `{% if %}` block for the appenders?

roles/openshift_logging/vars/main.yaml
```diff
@@ -8,3 +8,5 @@ es_recover_expected_nodes: "{{openshift_logging_es_cluster_size|int}}"
 es_ops_node_quorum: "{{openshift_logging_es_ops_cluster_size|int/2 + 1}}"
 es_ops_recover_after_nodes: "{{openshift_logging_es_ops_cluster_size|int - 1}}"
 es_ops_recover_expected_nodes: "{{openshift_logging_es_ops_cluster_size|int}}"
+
+es_log_appenders: ['file','console']
```
ewolinetz: can we rename this to `valid_es_log_appenders` so its more clear?

roles/openshift_logging/templates/elasticsearch-logging.yml.j2
```diff
@@ -44,16 +56,13 @@ appender:
...
-  # For more information see https://logging.apache.org/log4j/extras/apidocs/org/apache/log4j/rolling/RollingFileAppender.html
-  #file:
-    #type: extrasRollingFile
-    #file: ${path.logs}/${cluster.name}.log
-    #rollingPolicy: timeBased
-    #rollingPolicy.FileNamePattern: ${path.logs}/${cluster.name}.log.%d{yyyy-MM-dd}.gz
-    #layout:
-      #type: pattern
-      #conversionPattern: "[%d{ISO8601}][%-5p][%-25c] %m%n"
+  deprecation_log_file:
```
ewolinetz: is a deprecation log appender necessary if we have `deprecation: false` above?

roles/openshift_logging/tasks/generate_configmaps.yaml
jcantrill: Moved to `install_logging`.  The validation already occurs in the when statement with taking the difference of the list against the acceptable values as defined in `vars/main.yaml`

roles/openshift_logging/templates/elasticsearch-logging.yml.j2
jcantrill: I don't think so.  The appenders are static blocks that will only be used when a logger is told to use them, unless I am missing something.

roles/openshift_logging/templates/elasticsearch-logging.yml.j2
jcantrill: I think this is only here to support it being true and match the configured `WARN`.  I defer to @portante  since it was from his original PR

roles/openshift_logging/vars/main.yaml
jcantrill: done

roles/openshift_logging/tasks/generate_configmaps.yaml
ewolinetz: ack, i see it now

roles/openshift_logging/tasks/install_logging.yaml
```diff
@@ -15,6 +15,14 @@
   fail: msg="The openshift_logging_es_ops_cluster_size may only be scaled down manually. Please see official documentation on how to do this."
   when: openshift_logging_facts.elasticsearch_ops.deploymentconfigs | length > openshift_logging_es_ops_cluster_size|int
 
+- name: Validate Elasticsearch log appenders
+  fail:
+    msg: "openshift_logging_es_log_appenders must be defined as a list and only supports the following entries: {{valid_es_log_appenders | join(', ')}}"
+  when:
+    - es_logging_contents is undefined
+    - "{{ openshift_logging_es_log_appenders | list | difference(valid_es_log_appenders) | length != 0 }}"
+  changed_when: no
```
ewolinetz: remove the `changed_when`

roles/openshift_logging/templates/elasticsearch-logging.yml.j2
portante: Keeping this close to the default `logging.yml` shipped with Elasticsearch 2.x is helpful for folks supporting it.

jcantrill:
aos-ci-test

jcantrill:
[merge]

jcantrill:
[merge] flake?

portante:
It leaves folks exposed to support console logging, as currently that re-introduces a feedback loop within the OpenShift environment when enabled.  Are there any warnings regarding this situation in comments or in a README somewhere?  Or will this be documented somewhere?

PR #3781, opened 2017-03-28
Make the OCP available version detection excluder free

ingvagabund:
When detecting available OCP version via repoquery,
use yum.conf file with exclude= set to an empty array.
So the detection is independent of the OCP excluder.

### DIFF
--- a/roles/openshift_excluder/tasks/disable.yml
+++ b/roles/openshift_excluder/tasks/disable.yml
@@ -1,6 +1,5 @@
 ---
 # input variables
-# - with_status_check
 # - excluder_package_state
 # - docker_excluder_package_state
 - include: init.yml
@@ -35,6 +34,6 @@
       unexclude_docker_excluder: false
       # disable openshift excluder is never overrided to be enabled
       # disable it if the docker excluder is enabled
-      unexclude_openshift_excluder: true
+      unexclude_openshift_excluder: "{{ openshift_excluder_on | bool }}"
   when:
   - not openshift.common.is_atomic | bool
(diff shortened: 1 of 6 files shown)


ingvagabund:
aos-ci-test

ingvagabund:
@jhadvig just you know I am messing with excluders again :)

roles/openshift_excluder/tasks/disable.yml
```diff
@@ -35,6 +34,6 @@
       unexclude_docker_excluder: false
       # disable openshift excluder is never overrided to be enabled
       # disable it if the docker excluder is enabled
-      unexclude_openshift_excluder: true
+      unexclude_openshift_excluder: "{{ openshift_excluder_on | bool }}"
```
ashcrow: :+1:

sdodson:
[merge]

roles/openshift_version/tasks/set_version_rpm.yml
```diff
@@ -5,14 +5,41 @@
...
+- name: Create temporary yum.conf file
+  command: mktemp -d /tmp/yum.conf.XXXXXX
+  register: yum_conf_temp_file_result
+
+- set_fact:
+    yum_conf_temp_file: "{{yum_conf_temp_file_result.stdout}}/yum.conf"
+
+- name: Copy yum.conf into the temporary file
+  copy:
+    src: /etc/yum.conf
```
mtnbikenc: By default the `copy` module copies files from the control machine _to_ the remote machine.  In this instance you are trying to copy the local machines' yum.conf which is probably not what was intended.  Adding `remote_src: True` allows copying from the remote host.

PR #3782, opened 2017-03-28
Add etcd_debug and etcd_log_package_levels variables

sdodson:
Help debug https://github.com/openshift/origin/issues/13067

### DIFF
--- a/roles/etcd/templates/etcd.conf.j2
+++ b/roles/etcd/templates/etcd.conf.j2
@@ -60,3 +60,9 @@ ETCD_PEER_CA_FILE={{ etcd_peer_ca_file }}
 ETCD_PEER_CERT_FILE={{ etcd_peer_cert_file }}
 ETCD_PEER_KEY_FILE={{ etcd_peer_key_file }}
 {% endif -%}
+
+#[logging]
+ETCD_DEBUG="{{ etcd_debug | default(false) | string }}"
+{% if etcd_log_package_levels is defined %}
+ETCD_LOG_PACKAGE_LEVELS="{{ etcd_log_package_levels }}"
+{% endif %}
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -91,6 +91,10 @@ openshift_release=v1.5
 
 # Specify exact version of etcd to configure or upgrade to.
 # etcd_version="3.1.0"
+# Enable etcd debug logging, defaults to false
+# etcd_debug=true
+# Set etcd log levels by package
+# etcd_log_package_levels="etcdserver=WARNING,security=DEBUG"
 
 # Upgrade Hooks
 #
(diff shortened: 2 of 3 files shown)


sdodson:
aos-ci-test

roles/etcd/templates/etcd.conf.j2
```diff
@@ -60,3 +60,10 @@ ETCD_PEER_CA_FILE={{ etcd_peer_ca_file }}
 ETCD_PEER_CERT_FILE={{ etcd_peer_cert_file }}
 ETCD_PEER_KEY_FILE={{ etcd_peer_key_file }}
 {% endif -%}
+
+#[logging]
+ETCD_DEBUG="{{ 'true' if etcd_debug | default(false) else 'false' }}"
```
stevekuznetsov: Just cast to a string?

```
ETCD_DEBUG="{{ etcd_debug | default(false) | string }}"
```

roles/etcd/templates/etcd.conf.j2
```diff
@@ -60,3 +60,10 @@ ETCD_PEER_CA_FILE={{ etcd_peer_ca_file }}
...
+# examples for -log-package-levels etcdserver=WARNING,security=DEBUG
+{% if etcd_log_package_levels is defiend %}
```
stevekuznetsov: `s/defiend/defined`

roles/etcd/templates/etcd.conf.j2
sdodson: Ok, we've done it the original way elsewhere but it's not exactly a pattern.

sdodson: Hmm, etcd_debug=0 yields 'false' or '0'?

ingvagabund:
Complying to [1].

[1] https://github.com/coreos/etcd/blob/20490caaf0dcd96bb4a95e40625559def8ef5b04/Documentation/op-guide/configuration.md#logging-flags

roles/etcd/templates/etcd.conf.j2
```diff
@@ -60,3 +60,10 @@ ETCD_PEER_CA_FILE={{ etcd_peer_ca_file }}
 ETCD_PEER_CERT_FILE={{ etcd_peer_cert_file }}
 ETCD_PEER_KEY_FILE={{ etcd_peer_key_file }}
 {% endif -%}
+
+#[logging]
+ETCD_DEBUG="{{ etcd_debug | default(false) | string }}"
+# examples for -log-package-levels etcdserver=WARNING,security=DEBUG
```
stevekuznetsov: Maybe remove this, just clutter -- someone should be in the `etcd` doc if they're touching this.

sdodson:
aos-ci-test

sdodson:
[merge]

PR #3783, opened 2017-03-28
Add system-only annotation to hawkular-metrics-admin

enj:
https://bugzilla.redhat.com/show_bug.cgi?id=1436620

cc @jwforres @benjaminapetersen

### DIFF
--- a/roles/openshift_manageiq/vars/main.yml
+++ b/roles/openshift_manageiq/vars/main.yml
@@ -15,6 +15,8 @@ manageiq_metrics_admin_clusterrole:
   apiVersion: v1
   kind: ClusterRole
   metadata:
+    annotations:
+      authorization.openshift.io/system-only: "true"
     name: hawkular-metrics-admin
   rules:
   - apiGroups:


enj:
aos-ci-test

enj:
@sdodson from reading https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_manageiq/tasks/main.yaml I can see that this change will not be honored if the role already exists.

In origin we handle this through reconciliation - something similar needs to happen here.

enj:
cc @tbielawa @simon3z

sdodson:
> In origin we handle this through reconciliation - something similar needs to happen here.

Do the policy reconcilers handle this today? We call those during upgrade, everytime.

enj:
> Do the policy reconcilers handle this today? We call those during upgrade, everytime.

No, only the bootstrap cluster roles and bindings (embedded in `oc`) are reconciled.

simon3z:
cc @cben @enoodle

simon3z:
cc @moolitayer

PR #3785, opened 2017-03-28
Update docker image availability check

juanvallejo:
Addresses FIXME comments added in https://github.com/openshift/openshift-ansible/pull/3745

cc @rhcarvalho @sosiouxme @brenton

### DIFF
--- a/roles/openshift_health_checker/meta/main.yml
+++ b/roles/openshift_health_checker/meta/main.yml
@@ -2,3 +2,4 @@
 dependencies:
   - role: openshift_facts
   - role: openshift_repos
+  - role: openshift_version
--- a/roles/openshift_health_checker/openshift_checks/docker_image_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/docker_image_availability.py
@@ -13,7 +13,7 @@ class DockerImageAvailability(OpenShiftCheck):
     name = "docker_image_availability"
     tags = ["preflight"]
 
-    skopeo_image = "openshift/openshift-ansible"
+    dependencies = ["skopeo", "python-docker-py"]
 
     deployment_image_info = {
         "origin": {
@@ -35,44 +35,32 @@ class DockerImageAvailability(OpenShiftCheck):
         return super(DockerImageAvailability, cls).is_active(task_vars) and has_valid_deployment_type
 
     def run(self, tmp, task_vars):
+        msg, failed, changed = self.ensure_dependencies(task_vars)
+
+        # exit early if Skopeo update fails
+        if failed:
+            if "No package matching" in msg:
+                msg = "Ensure that all required dependencies can be installed via `yum`.\n"
+            return {
+                "failed": True,
+                "changed": changed,
+                "msg": (
+                    "Unable to update or install required dependency packages on this host;\n"
+                    "These are required in order to check Docker image availability:"
+                    "\n    {deps}\n{msg}"
+                ).format(deps=',\n    '.join(self.dependencies), msg=msg),
+            }
+
         required_images = self.required_images(task_vars)
         missing_images = set(required_images) - set(self.local_images(required_images, task_vars))
 
         # exit early if all images were found locally
         if not missing_images:
-            return {"changed": False}
-
-        msg, failed, changed = self.update_skopeo_image(task_vars)
-
-        # exit early if Skopeo update fails
-        if failed:
-            if "Error connecting: Error while fetching server API version" in msg:
-                msg = (
-                    "It appears Docker is not running.\n"
-                    "Please start Docker on this host before running this check.\n"
-                    "The full error reported was:\n  " + msg
-                    )
-
-            elif "Failed to import docker-py" in msg:
-                msg = (
-                    "The required Python docker-py module is not installed.\n"
-                    "Suggestion: install the python-docker-py package on this host."
-                    )
-            else:
-                msg = "The full message reported by the docker_image module was:\n" + msg
-            return {
-                "failed": True,
-                "changed": changed,
-                "msg": (
-                    "Unable to update the {img_name} image on this host;\n"
-                    "This is required in order to check Docker image availability.\n"
-                    "{msg}"
-                    ).format(img_name=self.skopeo_image, msg=msg),
-            }
+            return {"changed": changed}
 
         registries = self.known_docker_registries(task_vars)
         if not registries:
-            return {"failed": True, "msg": "Unable to retrieve any docker registries."}
+            return {"failed": True, "msg": "Unable to retrieve any docker registries.", "changed": changed}
 
         available_images = self.available_images(missing_images, registries, task_vars)
         unavailable_images = set(missing_images) - set(available_images)
@@ -81,9 +69,9 @@ class DockerImageAvailability(OpenShiftCheck):
             return {
                 "failed": True,
                 "msg": (
-                    "One or more required images are not available: {}.\n"
+                    "One or more required Docker images are not available:\n    {}\n"
                     "Configured registries: {}"
-                ).format(", ".join(sorted(unavailable_images)), ", ".join(registries)),
+                ).format(",\n    ".join(sorted(unavailable_images)), ", ".join(registries)),
                 "changed": changed,
             }
 
@@ -94,23 +82,46 @@ class DockerImageAvailability(OpenShiftCheck):
         image_info = self.deployment_image_info[deployment_type]
 
         openshift_release = get_var(task_vars, "openshift_release", default="latest")
-        openshift_image_tag = get_var(task_vars, "openshift_image_tag", default=openshift_release)
-        if openshift_image_tag and openshift_image_tag[0] != 'v':
-            openshift_image_tag = 'v' + openshift_image_tag
-
+        openshift_image_tag = get_var(task_vars, "openshift_image_tag")
         is_containerized = get_var(task_vars, "openshift", "common", "is_containerized")
-        images = set(self.non_qualified_docker_images(image_info["namespace"], image_info["name"], openshift_release,
-                                                      is_containerized))
+
+        images = set(self.required_docker_images(
+            image_info["namespace"],
+            image_info["name"],
+            ["registry-console"] if "enterprise" in deployment_type else [],  # include enterprise-only image names
+            openshift_release,
+            is_containerized,
+        ))
 
         # append images with qualified image tags to our list of required images.
         # these are images with a (v0.0.0.0) tag, rather than a standard release
         # format tag (v0.0). We want to check this set in both containerized and
         # non-containerized installations.
         images.update(
-            self.qualified_docker_images(image_info["namespace"], image_info["name"], openshift_image_tag)
+            self.required_qualified_docker_images(
+                image_info["namespace"],
+                image_info["name"],
+                openshift_image_tag,
+            ),
         )
 
         return images
+
+    @staticmethod
+    def required_docker_images(namespace, name, additional_image_names, version, is_containerized):
+        if is_containerized:
+            return ["{}/{}:{}".format(namespace, name, version)] if name else []
+
+        # include additional non-containerized images specific to the current deployment type
+        return ["{}/{}:{}".format(namespace, img_name, version) for img_name in additional_image_names]
+
+    @staticmethod
+    def required_qualified_docker_images(namespace, name, version):
+        # pylint: disable=invalid-name
+        return [
+            "{}/{}-{}:{}".format(namespace, name, suffix, version)
+            for suffix in ["haproxy-router", "docker-registry", "deployer", "pod"]
+        ]
 
     def local_images(self, images, task_vars):
         """Filter a list of images and return those available locally."""
@@ -126,27 +137,25 @@ class DockerImageAvailability(OpenShiftCheck):
 
         return bool(result.get("images", []))
 
-    def known_docker_registries(self, task_vars):
-        result = self.module_executor("docker_info", {}, task_vars)
-        if result.get("failed", False):
-            return False
+    @staticmethod
+    def known_docker_registries(task_vars):
+        docker_facts = get_var(task_vars, "openshift", "docker")
+        regs = set(docker_facts["additional_registries"])
 
-        docker_info = result.get("info", {})
-        return [registry.get("Name", "") for registry in docker_info.get("Registries", {})]
+        deployment_type = get_var(task_vars, "openshift_deployment_type")
+        if deployment_type == "origin":
+            regs.update(["docker.io"])
+        elif "enterprise" in deployment_type:
+            regs.update(["registry.access.redhat.com"])
+
+        return list(regs)
 
     def available_images(self, images, registries, task_vars):
         """Inspect existing images using Skopeo and return all images successfully inspected."""
         return [
             image for image in images
-            if self.is_image_available(image, registries, task_vars)
+            if any(self.is_available_skopeo_image(image, registry, task_vars) for registry in registries)
         ]
-
-    def is_image_available(self, image, registries, task_vars):
-        for registry in registries:
-            if self.is_available_skopeo_image(image, registry, task_vars):
-                return True
-
-        return False
 
     def is_available_skopeo_image(self, image, registry, task_vars):
         """Uses Skopeo to determine if required image exists in a given registry."""
@@ -156,32 +165,15 @@ class DockerImageAvailability(OpenShiftCheck):
             image=image,
         )
 
-        args = {
-            "name": "skopeo_inspect",
-            "image": self.skopeo_image,
-            "command": cmd_str,
-            "detach": False,
-            "cleanup": True,
-        }
-        result = self.module_executor("docker_container", args, task_vars)
-        return not result.get("failed", False)
+        args = {"_raw_params": cmd_str}
+        result = self.module_executor("command", args, task_vars)
+        return not result.get("failed", False) and result.get("rc", 0) == 0
 
-    @staticmethod
-    def non_qualified_docker_images(namespace, name, version, is_containerized):
-        if is_containerized:
-            return ["{}/{}:{}".format(namespace, name, version)] if name else []
+    # ensures that the skopeo and python-docker-py packages exist
+    # check is skipped on atomic installations
+    def ensure_dependencies(self, task_vars):
+        if get_var(task_vars, "openshift", "common", "is_atomic"):
+            return "", False, False
 
-        return ["{}/{}:{}".format(namespace, name, version)] if name else []
-
-    @staticmethod
-    def qualified_docker_images(namespace, name, version):
-        return [
-            "{}/{}-{}:{}".format(namespace, name, suffix, version)
-            for suffix in ["haproxy-router", "docker-registry", "deployer", "pod"]
-        ]
-
-    # ensures that the skopeo docker image exists, and updates it
-    # with latest if image was already present locally.
-    def update_skopeo_image(self, task_vars):
-        result = self.module_executor("docker_image", {"name": self.skopeo_image}, task_vars)
-        return result.get("msg", ""), result.get("failed", False), result.get("changed", False)
+        result = self.module_executor("yum", {"name": self.dependencies, "state": "latest"}, task_vars)
+        return result.get("msg", ""), result.get("failed", False) or result.get("rc", 0) != 0, result.get("changed")
--- a/roles/openshift_health_checker/openshift_checks/docker_image_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/docker_image_availability.py
@@ -15,19 +15,24 @@ class DockerImageAvailability(OpenShiftCheck):
 
     skopeo_image = "openshift/openshift-ansible"
 
-    # FIXME(juanvallejo): we should consider other possible values of
-    # `deployment_type` (the key here). See
-    # https://github.com/openshift/openshift-ansible/blob/8e26f8c/roles/openshift_repos/vars/main.yml#L7
-    docker_image_base = {
+    deployment_image_info = {
         "origin": {
-            "repo": "openshift",
-            "image": "origin",
+            "namespace": "openshift",
+            "name": "origin",
         },
         "openshift-enterprise": {
-            "repo": "openshift3",
-            "image": "ose",
+            "namespace": "openshift3",
+            "name": "ose",
         },
     }
+
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip hosts with unsupported deployment types."""
+        deployment_type = get_var(task_vars, "openshift_deployment_type")
+        has_valid_deployment_type = deployment_type in cls.deployment_image_info
+
+        return super(DockerImageAvailability, cls).is_active(task_vars) and has_valid_deployment_type
 
     def run(self, tmp, task_vars):
         required_images = self.required_images(task_vars)
@@ -41,13 +46,34 @@ class DockerImageAvailability(OpenShiftCheck):
 
         # exit early if Skopeo update fails
         if failed:
+            if "Error connecting: Error while fetching server API version" in msg:
+                msg = (
+                    "It appears Docker is not running.\n"
+                    "Please start Docker on this host before running this check.\n"
+                    "The full error reported was:\n  " + msg
+                    )
+
+            elif "Failed to import docker-py" in msg:
+                msg = (
+                    "The required Python docker-py module is not installed.\n"
+                    "Suggestion: install the python-docker-py package on this host."
+                    )
+            else:
+                msg = "The full message reported by the docker_image module was:\n" + msg
             return {
                 "failed": True,
                 "changed": changed,
-                "msg": "Failed to update Skopeo image ({img_name}). {msg}".format(img_name=self.skopeo_image, msg=msg),
+                "msg": (
+                    "Unable to update the {img_name} image on this host;\n"
+                    "This is required in order to check Docker image availability.\n"
+                    "{msg}"
+                    ).format(img_name=self.skopeo_image, msg=msg),
             }
 
         registries = self.known_docker_registries(task_vars)
+        if not registries:
+            return {"failed": True, "msg": "Unable to retrieve any docker registries."}
+
         available_images = self.available_images(missing_images, registries, task_vars)
         unavailable_images = set(missing_images) - set(available_images)
 
@@ -64,31 +90,24 @@ class DockerImageAvailability(OpenShiftCheck):
         return {"changed": changed}
 
     def required_images(self, task_vars):
-        deployment_type = get_var(task_vars, "deployment_type")
-        # FIXME(juanvallejo): we should handle gracefully with a proper error
-        # message when given an unexpected value for `deployment_type`.
-        image_base_name = self.docker_image_base[deployment_type]
+        deployment_type = get_var(task_vars, "openshift_deployment_type")
+        image_info = self.deployment_image_info[deployment_type]
 
-        openshift_release = get_var(task_vars, "openshift_release")
-        # FIXME(juanvallejo): this variable is not required when the
-        # installation is non-containerized. The example inventories have it
-        # commented out. We should handle gracefully and with a proper error
-        # message when this variable is required and not set.
-        openshift_image_tag = get_var(task_vars, "openshift_image_tag")
+        openshift_release = get_var(task_vars, "openshift_release", default="latest")
+        openshift_image_tag = get_var(task_vars, "openshift_image_tag", default=openshift_release)
+        if openshift_image_tag and openshift_image_tag[0] != 'v':
+            openshift_image_tag = 'v' + openshift_image_tag
 
         is_containerized = get_var(task_vars, "openshift", "common", "is_containerized")
-
-        if is_containerized:
-            images = set(self.containerized_docker_images(image_base_name, openshift_release))
-        else:
-            images = set(self.rpm_docker_images(image_base_name, openshift_release))
+        images = set(self.non_qualified_docker_images(image_info["namespace"], image_info["name"], openshift_release,
+                                                      is_containerized))
 
         # append images with qualified image tags to our list of required images.
         # these are images with a (v0.0.0.0) tag, rather than a standard release
         # format tag (v0.0). We want to check this set in both containerized and
         # non-containerized installations.
         images.update(
-            self.qualified_docker_images(self.image_from_base_name(image_base_name), "v" + openshift_image_tag)
+            self.qualified_docker_images(image_info["namespace"], image_info["name"], openshift_image_tag)
         )
 
         return images
@@ -109,13 +128,10 @@ class DockerImageAvailability(OpenShiftCheck):
 
     def known_docker_registries(self, task_vars):
         result = self.module_executor("docker_info", {}, task_vars)
-
         if result.get("failed", False):
-            return []
+            return False
 
-        # FIXME(juanvallejo): wrong default type, result["info"] is expected to
-        # contain a dictionary (see how we call `docker_info.get` below).
-        docker_info = result.get("info", "")
+        docker_info = result.get("info", {})
         return [registry.get("Name", "") for registry in docker_info.get("Registries", {})]
 
     def available_images(self, images, registries, task_vars):
@@ -148,29 +164,21 @@ class DockerImageAvailability(OpenShiftCheck):
             "cleanup": True,
         }
         result = self.module_executor("docker_container", args, task_vars)
-        return result.get("failed", False)
-
-    def containerized_docker_images(self, base_name, version):
-        return [
-            "{image}:{version}".format(image=self.image_from_base_name(base_name), version=version)
-        ]
+        return not result.get("failed", False)
 
     @staticmethod
-    def rpm_docker_images(base, version):
-        return [
-            "{image_repo}/registry-console:{version}".format(image_repo=base["repo"], version=version)
-        ]
+    def non_qualified_docker_images(namespace, name, version, is_containerized):
+        if is_containerized:
+            return ["{}/{}:{}".format(namespace, name, version)] if name else []
+
+        return ["{}/{}:{}".format(namespace, name, version)] if name else []
 
     @staticmethod
-    def qualified_docker_images(image_name, version):
+    def qualified_docker_images(namespace, name, version):
         return [
-            "{}-{}:{}".format(image_name, component, version)
-            for component in "haproxy-router docker-registry deployer pod".split()
+            "{}/{}-{}:{}".format(namespace, name, suffix, version)
+            for suffix in ["haproxy-router", "docker-registry", "deployer", "pod"]
         ]
-
-    @staticmethod
-    def image_from_base_name(base):
-        return "".join([base["repo"], "/", base["image"]])
 
     # ensures that the skopeo docker image exists, and updates it
     # with latest if image was already present locally.
--- a/roles/openshift_health_checker/test/docker_image_availability_test.py
+++ b/roles/openshift_health_checker/test/docker_image_availability_test.py
@@ -3,26 +3,177 @@ import pytest
 from openshift_checks.docker_image_availability import DockerImageAvailability
 
 
-@pytest.mark.xfail(strict=True)  # TODO: remove this once this test is fully implemented.
-@pytest.mark.parametrize('task_vars,expected_result', [
-    (
-        dict(
-            openshift=dict(common=dict(
-                service_type='origin',
-                is_containerized=False,
-            )),
-            openshift_release='v3.5',
-            deployment_type='origin',
-            openshift_image_tag='',  # FIXME: should not be required
-        ),
-        {'changed': False},
-    ),
-    # TODO: add more parameters here to test the multiple possible inputs that affect behavior.
+@pytest.mark.parametrize('deployment_type,is_active', [
+    ("origin", True),
+    ("openshift-enterprise", True),
+    ("enterprise", False),
+    ("online", False),
+    ("invalid", False),
+    ("", False),
 ])
-def test_docker_image_availability(task_vars, expected_result):
+def test_is_active(deployment_type, is_active):
+    task_vars = dict(
+        openshift_deployment_type=deployment_type,
+    )
+    assert DockerImageAvailability.is_active(task_vars=task_vars) == is_active
+
+
+@pytest.mark.parametrize("is_containerized", [
+    True,
+    False,
+])
+def test_all_images_available_locally(is_containerized):
+    def execute_module(module_name, args, task_vars):
+        assert module_name == "docker_image_facts"
+        assert 'name' in args
+        assert args['name']
+        return {
+            'images': [args['name']],
+        }
+
+    result = DockerImageAvailability(execute_module=execute_module).run(tmp=None, task_vars=dict(
+        openshift=dict(common=dict(
+            service_type='origin',
+            is_containerized=is_containerized,
+        )),
+        openshift_deployment_type='origin',
+        openshift_release='v3.4',
+        openshift_image_tag='3.4',
+    ))
+
+    assert not result.get('failed', False)
+
+
+@pytest.mark.parametrize("module_failure", [
+    True,
+    False,
+])
+def test_all_images_available_remotely(module_failure):
+    def execute_module(module_name, args, task_vars):
+        return {
+            'docker_image_facts': {'images': [], 'failed': module_failure},
+            'docker_info': {'info': {'Registries': [{'Name': 'docker.io'}, {'Name': 'registry.access.redhat.com'}]}},
+        }.get(module_name, {'changed': False})
+
+    result = DockerImageAvailability(execute_module=execute_module).run(tmp=None, task_vars=dict(
+        openshift=dict(common=dict(
+            service_type='origin',
+            is_containerized=False,
+        )),
+        openshift_deployment_type='origin',
+        openshift_release='3.4'
+    ))
+
+    assert not result.get('failed', False)
+
+
+def test_all_images_unavailable():
     def execute_module(module_name=None, module_args=None, tmp=None, task_vars=None):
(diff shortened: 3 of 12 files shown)


rhcarvalho:
I'm surprised that Travis is green... There are still assertions missing and code to be added, e.g. in https://github.com/openshift/openshift-ansible/pull/3785/files#diff-79ee78cd0af05985a9f6cbb264338c96L24

Aha! Because all tests are marked as "expect fail" -- https://travis-ci.org/openshift/openshift-ansible/jobs/216077684#L267
You need to remove this line https://github.com/openshift/openshift-ansible/pull/3785/files#diff-79ee78cd0af05985a9f6cbb264338c96L6

rhcarvalho:
@juanvallejo, working on this, these are commands you will need all the time:

```
# run only the tests in our role of interest
tox -e py27-unit -- roles/openshift_health_checker/test

# run tests with higher verbosity, e.g. to see complete diffs on error
tox -e py27-unit -- roles/openshift_health_checker/test -vv

# see coverage report on the browser
# after you run tests, refresh the page relative to the file you
# are testing, you should see the coverage level increase
# testing after is worse than testing first, be careful not to rely
# on implementation details in the tests
xdg-open cover/index.html

# run tests on python 3.5
tox -e py35-unit -- roles/openshift_health_checker/test

# run linters, etc
tox -l
tox -e py27-flake8
tox -e py27-pylint
tox -e py35-flake8
tox -e py35-pylint
```

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -15,22 +15,37 @@ class DockerImageAvailability(OpenShiftCheck):
...
 
-    # FIXME(juanvallejo): we should consider other possible values of
-    # `deployment_type` (the key here). See
-    # https://github.com/openshift/openshift-ansible/blob/8e26f8c/roles/openshift_repos/vars/main.yml#L7
     docker_image_base = {
         "origin": {
             "repo": "openshift",
             "image": "origin",
         },
+        "online": {
```
rhcarvalho: Is online really using openshift/origin?

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -65,16 +80,13 @@ def run(self, tmp, task_vars):
 
     def required_images(self, task_vars):
         deployment_type = get_var(task_vars, "deployment_type")
-        # FIXME(juanvallejo): we should handle gracefully with a proper error
-        # message when given an unexpected value for `deployment_type`.
-        image_base_name = self.docker_image_base[deployment_type]
+        image_base_name = self.docker_image_base.get(deployment_type)
+        if not image_base_name:
+            msg = "Unable to determine required images for deployment type {deployment}"
+            return set(), msg.format(deployment=deployment_type)
```
rhcarvalho: Hmm, this function sometimes return 2 values, sometimes 1.

I think in this case we can probably use the `OpenShiftCheckException` to simplify control flow.

```python
if deployment_type not in self.docker_image_base:
    raise OpenShiftCheckException(...msg here...)
image_base_name = self.docker_image_base[deployment_type]
...
```

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
rhcarvalho: I found this comment which I think relates to the RPM package name...
https://github.com/openshift/openshift-ansible/blob/8e26f8cf9594693702fd1e545b47116cbbff34f0/roles/openshift_repos/vars/main.yml#L3

juanvallejo: > I found this comment which I think relates to the RPM package name...

Thanks, although I am still not sure what is meant by `online currently uses 'openshift' packages`. Would this use the hub.docker.io registry? and is `openshift` the namespace, or part of the image name (e.g. `openshift/openshift-pod`)

rhcarvalho: I believe the comment I linked refers to RPM packages with a prefix of "openshift-"... I don't know / didn't look for where we define what images are used when `deployment_type` is `online`.

Once we know that we can also update https://github.com/openshift/openshift-ansible/pull/3799.

Why did you start with setting this to the same as origin? I thought you may had found it somewhere :)

juanvallejo: > Why did you start with setting this to the same as origin? I thought you may had found it somewhere :)

Ah, I was hoping someone would know the right values and steer me in the right direction upon I opening the PR :)

juanvallejo: After talking with @sdodson, he linked me to https://github.com/openshift/openshift-ansible/blob/8e26f8cf9594693702fd1e545b47116cbbff34f0/roles/openshift_facts/library/openshift_facts.py#L1741-L1771 which is where required images for each deployment type are defaulted. He recommended not worrying about `atomic-enterprise`, and it looks like the cases for `origin` and `online` look to be the same

rhcarvalho: Both fall into the `else` clause... could have been an oversight? Anything that doesn't fall into the explicit `if` clauses end up treated as "origin"...

When I was looking for docs about deployment types, I saw "online" in the list of known values, but it is not documented in DEPLOYMENT_TYPES.md.

Would be good to investigate how is it used and what is the intention of setting `deployment_type=online` in the inventory file.

juanvallejo: @rhcarvalho 
> Would be good to investigate how is it used and what is the intention of setting deployment_type=online in the inventory file.

After speaking with @sallyom, she clarified that "online" is an old type that was valid until release 1.3 of openshift-ansible. When using "online", `openshift3/ose-*` images were used as well. I have updated the PR to address this. 

We could remove `online` altogether from this PR and just treat it as an invalid deployment type if you think this would be the better approach.

rhcarvalho: > We could remove online altogether from this PR and just treat it as an invalid deployment type if you think this would be the better approach.

Yes, let's remove it. We don't need to try to support something that is not supported by the installer. Thanks for investigating.

BTW - if it is not valid any longer, perhaps we should remove it from https://github.com/openshift/openshift-ansible/blob/6051550cb/roles/openshift_sanitize_inventory/vars/main.yml#L7? (/cc @sosiouxme )

rhcarvalho: I'm thinking we may actually sometimes support something that works only on older versions, but this specific case don't sound very useful if the online team is not using this deployment type anymore (you may try to confirm that).

sosiouxme: Makes sense to me to remove it

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -27,6 +24,10 @@ class DockerImageAvailability(OpenShiftCheck):
             "repo": "openshift3",
             "image": "ose",
         },
+        "enterprise": {
+            "repo": "openshift3",
+            "image": "ose",
+        },
```
rhcarvalho: As per recent discussion and documentation update in #3799, `enterprise` is not in use at the moment, and was used for 3.0.x.

@brenton do you think checks should go as far as supporting OSE 3.0?

In this case I'd probably have support for only `origin` and `openshift-enterprise` (for OSE/OCP 3.1+).

If need be, for this or other checks, we could also explicitly disable the checks based on the target version... but in practice the format of the inventory we expect is a current one, we don't go back to support what might an inventory have looked like back in 3.0..., so it might even be a challenge to handle incompatibilities in inventories -- if there was ever any backward-incompatible change in name of variables for example (and I am aware of a recent change from deployment_type -> openshift_deployment_type).

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -65,16 +69,14 @@ def run(self, tmp, task_vars):
 
     def required_images(self, task_vars):
         deployment_type = get_var(task_vars, "deployment_type")
-        # FIXME(juanvallejo): we should handle gracefully with a proper error
-        # message when given an unexpected value for `deployment_type`.
-        image_base_name = self.docker_image_base[deployment_type]
+        if deployment_type not in self.docker_image_base:
+            msg = "Unable to determine required images for deployment type {deployment}"
+            raise OpenShiftCheckException(msg.format(deployment=deployment_type))
```
rhcarvalho: @juanvallejo @sosiouxme I'm thinking here, would it be best to fail a check when we meet an unexpected deployment_type, or should we simply skip the check (by implementing `is_active`)?

In your opinion, what would be the best user experience?

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -65,16 +69,14 @@ def run(self, tmp, task_vars):
...
+
+        image_base_name = self.docker_image_base.get(deployment_type)
 
         openshift_release = get_var(task_vars, "openshift_release")
-        # FIXME(juanvallejo): this variable is not required when the
-        # installation is non-containerized. The example inventories have it
-        # commented out. We should handle gracefully and with a proper error
-        # message when this variable is required and not set.
-        openshift_image_tag = get_var(task_vars, "openshift_image_tag")
+        openshift_image_tag = get_var(task_vars, "openshift_image_tag", default=openshift_release[1:])
```
rhcarvalho: Why are we slicing out the first character? It may be a 'v', or it may not.... @sosiouxme has done some work with normalizing the release version.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -148,17 +142,17 @@ def is_available_skopeo_image(self, image, registry, task_vars):
             "cleanup": True,
         }
         result = self.module_executor("docker_container", args, task_vars)
-        return result.get("failed", False)
+        return not result.get("failed", False)
```
rhcarvalho: The logic was reversed?!

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
juanvallejo: I think it would be consistent with cases such as [this one](https://github.com/openshift/openshift-ansible/pull/3518/files#diff-dd45df49348ab2716e213a0ecc5b0f4dR24) that we skip the check if a deployment type is not supported

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
juanvallejo: The logic was wrong. The method's name is `is_available_skopeo_image`. We expect a return value of `true` if the module_executor's result did *not* fail

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
juanvallejo: I can update this to only slice the first char if it `== 'v'`. Unless it is now safe to assume that the value of `openshift_image_tag` will always begin with `v`?

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
brenton: Definitely no need to support OSE 3.0 at this point.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -102,20 +106,12 @@ def local_images(self, images, task_vars):
 
     def is_image_local(self, image, task_vars):
         result = self.module_executor("docker_image_facts", {"name": image}, task_vars)
-        if result.get("failed", False):
```
juanvallejo: The module returns an empty list for the `images` key if it is unable to fetch info about a given image. This evaluates to `False` anyway, so I deemed this branch to be redundant

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -102,20 +106,12 @@ def local_images(self, images, task_vars):
...
-            return False
-
         return bool(result.get("images", []))
 
     def known_docker_registries(self, task_vars):
         result = self.module_executor("docker_info", {}, task_vars)
 
-        if result.get("failed", False):
```
juanvallejo: Similar reason as the previous comment, we return an empty list if there is a failure below, so this seemed redundant to have

juanvallejo:

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -15,20 +15,27 @@ class DockerImageAvailability(OpenShiftCheck):
...
     skopeo_image = "openshift/openshift-ansible"
 
-    # FIXME(juanvallejo): we should consider other possible values of
-    # `deployment_type` (the key here). See
-    # https://github.com/openshift/openshift-ansible/blob/8e26f8c/roles/openshift_repos/vars/main.yml#L7
     docker_image_base = {
         "origin": {
             "repo": "openshift",
             "image": "origin",
+            "has_registry_console": False,
```
rhcarvalho: This seems to stretch a lot what `docker_image_base` holds... can we even describe what this structure is?

Both methods `containerized_docker_images` and `rpm_docker_images` return a list with a single item based on the values in `docker_image_base`. Wouldn't it be simpler and clearer to just store a flat list of images for each case (origin / ocp)?

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -15,20 +15,27 @@ class DockerImageAvailability(OpenShiftCheck):
...
             "repo": "openshift3",
             "image": "ose",
+            "has_registry_console": True,
         },
     }
 
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip hosts with unsupported deployment types."""
+        deployment_type = get_var(task_vars, "deployment_type")
```
rhcarvalho: Perhaps here we should be reading `openshift_deployment_type`, though older configurations/inventories will be still defining only `deployment_type`... @sosiouxme, any suggestion?

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -65,16 +75,10 @@ def run(self, tmp, task_vars):
 
     def required_images(self, task_vars):
         deployment_type = get_var(task_vars, "deployment_type")
-        # FIXME(juanvallejo): we should handle gracefully with a proper error
-        # message when given an unexpected value for `deployment_type`.
-        image_base_name = self.docker_image_base[deployment_type]
+        image_base_name = self.docker_image_base.get(deployment_type)
```
rhcarvalho: This means that `image_base_name` may be `None`, and then in the `image_from_base_name` method we treat it as a dict and may get a `TypeError` exception.

Since we already check the deployment type in `is_active`, here we can access it directly as it was before.

The handling of unexpected value for `deployment_type` from the FIXME comment has been resolved by skipping the check (though we do not currently produce messages to explain why a check was skipped, potential future feature).

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -65,16 +75,10 @@ def run(self, tmp, task_vars):
...
 
-        openshift_release = get_var(task_vars, "openshift_release")
-        # FIXME(juanvallejo): this variable is not required when the
-        # installation is non-containerized. The example inventories have it
-        # commented out. We should handle gracefully and with a proper error
-        # message when this variable is required and not set.
-        openshift_image_tag = get_var(task_vars, "openshift_image_tag")
+        openshift_release = get_var(task_vars, "openshift_release", default="latest")
+        openshift_image_tag = get_var(task_vars, "openshift_image_tag", default=openshift_release)
```
rhcarvalho: Are the `openshift_release` and `openshift_image_tag` Ansible vars guaranteed to be in the same format? E.g., presence or not of a 'v' suffix.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
rhcarvalho: The fact that the module doesn't return `failed` explicitly today is an implementation detail. We should not rely on that on client code.
If for instance a module will raise an exception, let's say because of a missing dependency or programming error, Ansible captures it and would wrap the result with failed: True. It is useful to handle the case when module executions do fail.

roles/openshift_health_checker/test/docker_image_availability_test.py
```diff
@@ -3,26 +3,171 @@
...
+        openshift_release='v3.4',
+        openshift_image_tag='3.4',
+    ))
+
+    assert not result.get('failed', False)
+
+
+def test_all_images_available_remotely():
+    def execute_module(module_name, args, task_vars):
+        if module_name == "docker_image_facts":
```
rhcarvalho: The current code is fine, but I'd like to show an alternative that shows one "trick" to eliminate chains of ifs:

```python
def execute_module(module_name, args, task_vars):
    return {
        'docker_image_facts': {'images': []},
        'docker_info': {'info': {'Registries': [{'Name': 'docker.io'}, {'Name': 'registry.access.redhat.com'}]}},
    }.get(module_name, {'changed': False})
```

Because it is unlikely that we add more items to the list of Registries in the docker_info output, it is rather okay to have it in a single line, though it could it one registry per line also if we want to make that easier to read.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
sosiouxme: Both will be set.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
sosiouxme: `openshift_release` should be sanitized at this point to `X.Y[.Z[.W]]. If the `openshift_version` role has run then it should be standardized to `vX.Y.Z.W` otherwise there is no guarantee it is even defined.

juanvallejo: @rhcarvalho 
> Are the openshift_release and openshift_image_tag Ansible vars guaranteed to be in the same format? E.g., presence or not of a 'v' suffix.

Not from what I have seen so far. Will add check below this line for this

roles/openshift_health_checker/meta/main.yml
```diff
@@ -2,3 +2,4 @@
 dependencies:
   - role: openshift_facts
   - role: openshift_repos
+  - role: openshift_version
```
juanvallejo: @sosiouxme Thanks for the feedback, per our conversation, I will keep the `openshift_version` dependency as part of this PR, and open a new PR to improve validation in `openshift_version`

sosiouxme: I had a look through `openshift_version` and I can live with it. If it fails, there's a pretty fundamental problem and the user should probably fix that first. The messaging doesn't look too bad coming through our callback plugin, but I drew up a [PR](https://github.com/openshift/openshift-ansible/pull/3949) to improve it a bit.

We do need to improve our messaging when docker and/or docker-py aren't available but that's unrelated to this line of code :)

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
juanvallejo: @sosiouxme 
> openshift_release should be sanitized at this point to X.Y[.Z[.W]]. If the openshift_version role has run then it should be standardized to vX.Y.Z.W otherwise there is no guarantee it is even defined.

Since we are keeping `openshift_version` as a dep, do you think it is still worth having a check for a leading `v` in the value of `openshift_image_tag`?

sosiouxme: Well, either `openshift_version` fails before this or `openshift_image_tag` should be standardized to have the "v". So practically speaking, not needed. There's probably some argument for keeping it as a protective measure, asserting your assumptions, etc. but I think it would mainly just be clutter.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
juanvallejo: @rhcarvalho 
> Both methods containerized_docker_images and rpm_docker_images return a list with a single item based on the values in docker_image_base. Wouldn't it be simpler and clearer to just store a flat list of images for each case (origin / ocp)?

Should this list be stored in the `docker_image_base` dict? Also, should it only apply to the values returned by `containerized_docker_images` and `rpm_images`, or should `qualified_docker_images` also follow this?

rhcarvalho: My thought is that it may be clearer, in this particular case, to have a literal list of the images instead of doing some form of computation to get to them.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -15,20 +15,29 @@ class DockerImageAvailability(OpenShiftCheck):
 
     skopeo_image = "openshift/openshift-ansible"
 
-    # FIXME(juanvallejo): we should consider other possible values of
-    # `deployment_type` (the key here). See
-    # https://github.com/openshift/openshift-ansible/blob/8e26f8c/roles/openshift_repos/vars/main.yml#L7
-    docker_image_base = {
+    deployment_image_info = {
```
rhcarvalho: Expanding on a previous comment about making this a flat list, I was thinking something along the lines of https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_health_checker/openshift_checks/package_availability.py#L31-L70

I haven't tried to see the implications, but had a feeling it could be simpler.

juanvallejo: hm, in the case of a flat list here, we would need to handle cases for images with qualified image tags, non-qualified image tags (`vX.Y` instead of `vX.Y.Z`), and of those two, cases where images exist for one deployment but not for another depending on the install type (only case so far is the `registry-console` image not existing for `origin` deployments in non-containerized installs, and `openshift/origin:v...` existing in containerized installs, but being formatted as `openshift3/ose:v...` for the enterprise deployment). I was trying to avoid returning multiple lists under different conditions in these methods, but will look at this once more and try to simplify further

juanvallejo:
@rhcarvalho PTAL

sosiouxme:
@juanvallejo Could you rebase this on master? Tests won't run locally with old tox config.

sosiouxme:
I created https://github.com/juanvallejo/openshift-ansible/pull/2 to improve the messaging around failures from the `docker_image` Ansible module.

However there is a fundamental concern left. I don't know that it has to be addressed for this PR to proceed but it has to be addressed soon for UX to be acceptable. That is, docker has to be running and docker-py needs to be installed for this check to pass.

The way things are today, if you're doing an install on an RPM-based system and you don't have docker (at all, or not running), the install playbook will install it for you (and respect excluders in doing so) and start it. If you don't have docker-py on the system, nothing else requires it right now so it's no problem. If we insert this check with other preflight checks early in the install playbook then we're going to be forcing users to do these things manually up front (or to skip the check).

Installing docker-py is maybe something we should just add to the [required packages](https://github.com/openshift/openshift-ansible/blob/d5ec349106fc4f80cdc1dfc5feb0036c6fa9ebb9/roles/openshift_facts/vars/default.yml). It seems a little odd to add a requirement solely for the checks but it doesn't seem too much of a stretch from existing practice.

Installing and starting Docker before the check would be harder to swallow (not least because it's a major system mutation). For OCP users, it's not wholly bad to expect that to be done as a prerequisite, as they should need to configure Docker anyway for acceptable storage options. For CI and Origin "just try it out" users this seems like a major pain.

I wonder if we could just skip the check for Origin, or make it non-blocking. Thinking more broadly, I wonder if all of the checks should be non-blocking for Origin by default; perhaps keeping the nice summary of failures (and perhaps pausing for user feedback), but not actually failing the install. Or maybe that would be better for *all* users running these checks via install or upgrade. "We found these problems, do you want to go ahead and run the rest of the playbook anyway? Y/N"

@brenton @rhcarvalho @codificat @juanvallejo what do you think?

juanvallejo:
@sosiouxme 
> Could you rebase this on master? Tests won't run locally with old tox config.

Went ahead and rebased with latest master

> Installing docker-py is maybe something we should just add to the required packages. It seems a little odd to add a requirement solely for the checks but it doesn't seem too much of a stretch from existing practice.

I agree with adding docker-py to the list of required packages

> I wonder if we could just skip the check for Origin, or make it non-blocking.

I think making this check non-blocking for Origin makes sense

brenton:
@sosiouxme, in general I'm fine with skipping checks in Origin whenever we don't have a clear way of making the experience better in the community.  I'd probably lean to installing python-docker-py for now until we know more.

rhcarvalho:
- Installing `docker-py`

  I'm in favor of installing `docker-py` if necessary, but against doing so by via `openshift_facts`, since `openshift_facts` doesn't depend on it.

  In this particular case, it is well likely that we DO NOT need neither Docker nor docker-py for this check. If you don't have Docker installed, it is obvious that the images are not available locally, thus all we need to do is use `skopeo` to inspect remote images.

  Does `skopeo` depend on a local Docker?

  What to do if `skopeo` is not available? My suggestion is to skip the check. It is available in our image, and if you are using a docker image, then you do have docker ;)

- Required x warning only

  This is indeed something we need to sort out as we integrate with install and upgrade flows. Note that making it interactive is not always an option (e.g., won't work for CI), and that _warning only_ means the output will likely be ignored in the middle of a wall of text, defeating the purpose of the checks. We need to fine tune annoyance x usefulness.

brenton:
Yeah, I should have mentioned that openshift_facts definitely shouldn't be the thing that installs docker-py.  The current package installs in that module cause enough problems already.

juanvallejo:
@rhcarvalho 
> Does skopeo depend on a local Docker?

Yes, [we use the `docker_container` ansible module to run `skopeo` from our `openshift-ansible` image in a container](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_health_checker/openshift_checks/docker_image_availability.py#L150). I believe we decided on this approach so as to not require another yum package on the host machine.

sosiouxme:
>
>    -
>
>    Installing docker-py
>
>    I'm in favor of installing docker-py if necessary, but against doing
>    so by via openshift_facts, since openshift_facts doesn't depend on it.
>
>
We can do it in the checks action plugin or define a task in the
`openshift_health_checker` role (probably clearest).


>
>    -
>
>    Does skopeo depend on a local Docker?
>
>    What to do if skopeo is not available? My suggestion is to skip the
>    check. It is available in our image, and if you are using a docker image,
>    then you do have docker ;)
>
>
This goes back to the bootstrapping discussion on skopeo. Atomic includes
it, and a skopeo pkg seems to be available for Fedora, CentOS and RHEL (in
extras). Perhaps we can make it another check dependency like docker-py.
Then no, it wouldn't depend on Docker. I like that.


>
>    -
>
>    Required x warning only
>
>    This is indeed something we need to sort out as we integrate with
>    install and upgrade flows. Note that making it interactive is not always an
>    option (e.g., won't work for CI), and that *warning only* means the
>    output will likely be ignored in the middle of a wall of text, defeating
>    the purpose of the checks. We need to fine tune annoyance x usefulness.
>
> Agreed. Need to see what is possible.

rhcarvalho:
Suggestion:

- If `docker` is not available, skip the check. Rationale: we can't do much without "starting the install" and we could as well don't do that. The thing is that if we plug this in at the right stage in the install playbooks, you'd already have docker installed anyway, so this check would always be able to be run.

- If `docker` is available but `docker-py` is not, try to install it in the context of the check. Rationale: hmm, not a strong argument, but `docker-py` is less intrusive than `docker`, so I'd be fine with installing it if not done yet by the installer.

Installing it in the role level is too global -- you can use the role which is a "framework for running checks" without intending to run that one particular check that depends on X.

Thoughts?

brenton:
I like this suggestion.  Especially if there's a decent way to present a 'warning' to the admin that we couldn't run all the checks because docker was missing.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -48,14 +48,20 @@ def run(self, tmp, task_vars):
...
+                    "It appears Docker is not running. This check will be skipped.\n"
                     "Please start Docker on this host before running this check.\n"
                     "The full error reported was:\n  " + msg
                     )
+
+                return {
+                    "failed": False,
+                    "changed": changed,
+                    "msg": msg,
+                }
```
rhcarvalho: This is silencing the error instead of skipping. Normally in Ansible when `failed` is not present or `False` we don't ever look for a `msg`.

Instead, if possible, I'd use the mechanism we have for skipping checks -- implementing `is_active`. When it is not active, we automatically set `skipped: True` in the return value, and then we could present skipped checks (with perhaps a reason, which is something we may still need to add) in the UI.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -48,14 +48,20 @@ def run(self, tmp, task_vars):
...
             elif "Failed to import docker-py" in msg:
                 msg = (
                     "The required Python docker-py module is not installed.\n"
-                    "Suggestion: install the python-docker-py package on this host."
+                    "Suggestion: install the python docker-py package on this host."
```
rhcarvalho: Add least in Fedora the RPM package is indeed called `python-docker-py`.

If you don't mean the RPM package name, then _Python_ should be capitalized.

sosiouxme:
We could install the skopeo package inside the check itself (call out to the pkg manager module). Then we would not need docker-py, docker, or a docker pull of the image - much more lightweight. And go ahead and fail or skip the check if skopeo can't be installed.

juanvallejo:
@sosiouxme 
> We could install the skopeo package inside the check itself (call out to the pkg manager module). Then we would not need docker-py, docker, or a docker pull of the image - much more lightweight. And go ahead and fail or skip the check if skopeo can't be installed.

While I like the sound of this approach, by relying on installing skopeo via pkg manager module, won't we run into issues if the host:

- does not have the repo with skopeo enabled
- has no internet connectivity

brenton:
In general I think as long as a tool does everything it can reasonably do to workaround a limitation I'm fine if it simply fails with an error that says "Skopeo is needed for .... Please see ___ for more information".

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -140,40 +154,26 @@ def is_available_skopeo_image(self, image, registry, task_vars):
...
-    @staticmethod
-    def image_from_base_name(base):
-        return "".join([base["repo"], "/", base["image"]])
-
     # ensures that the skopeo docker image exists, and updates it
     # with latest if image was already present locally.
     def update_skopeo_image(self, task_vars):
-        result = self.module_executor("docker_image", {"name": self.skopeo_image}, task_vars)
-        return result.get("msg", ""), result.get("failed", False), result.get("changed", False)
+        result = self.module_executor("yum", {"name": "skopeo", "state": "latest"}, task_vars)
```
juanvallejo: @sosiouxme went ahead and removed `docker-py` dep for using skopeo, however [this line](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_health_checker/openshift_checks/docker_image_availability.py#L50) will execute before the skopeo check, to retrieve known docker registries. Which means we will still fail without `docker-py` before even hitting the skopeo part of this check. 

A possible solution would be to have a default `docker.io` registry, so that in case no registries are found, we just use that one (a warning here explaining what happened would most likely be in order, however it might be something that ultimately gets lost in a wall of text.

Another solution could be to try to look in `ansible_facts` for `additional_registries` under `ansible_facts.openshift.docker`

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -107,16 +112,16 @@ def is_image_local(self, image, task_vars):
...
-        result = self.module_executor("docker_info", {}, task_vars)
+    @staticmethod
+    def known_docker_registries(task_vars):
+        docker_facts = get_var(task_vars, "openshift", "docker")
+        regs = set(docker_facts["additional_registries"])
 
-        if result.get("failed", False):
-            return []
+        deployment_type = get_var(task_vars, "openshift_deployment_type")
+        if deployment_type == "origin" and "docker.io" not in regs:
```
sosiouxme: Don't need to check `"docker.io" not in regs`, update is idempotent

Can we do the same for enterprise? https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.ose.example#L74

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -140,40 +145,26 @@ def is_available_skopeo_image(self, image, registry, task_vars):
...
+        return not result.get("failed", False) and result.get("rc", 0) == 0
 
     @staticmethod
-    def rpm_docker_images(base, version):
-        return [
-            "{image_repo}/registry-console:{version}".format(image_repo=base["repo"], version=version)
-        ]
+    def non_qualified_docker_images(namespace, name, version, is_containerized):
+        if is_containerized:
+            return ["{}/{}:{}".format(namespace, name, version)] if name else []
```
sosiouxme: what am I missing here, these two returns look the same...

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -140,40 +145,26 @@ def is_available_skopeo_image(self, image, registry, task_vars):
...
+            for suffix in ["haproxy-router", "docker-registry", "deployer", "pod"]
         ]
 
-    @staticmethod
-    def image_from_base_name(base):
-        return "".join([base["repo"], "/", base["image"]])
-
     # ensures that the skopeo docker image exists, and updates it
     # with latest if image was already present locally.
     def update_skopeo_image(self, task_vars):
```
sosiouxme: Err, unless we don't need docker-py at all any more... which seems to be the case :)

juanvallejo: @sosiouxme 
> And for that matter, why not install python-docker-py at this point?

So we do still need `docker-py` in order to use `docker_image_facts` to determine which images are already available locally, however the method for that is called before this one. I guess we could always have a call to this method be the first thing on the check? wdyt

sosiouxme: I think, yeah... attempt to ensure both deps exist at the top of the method, and fail out if they don't. That seems cleanest.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -107,16 +117,18 @@ def is_image_local(self, image, task_vars):
...
+            regs.update(["registry.access.redhat.com"])
 
-        # FIXME(juanvallejo): wrong default type, result["info"] is expected to
-        # contain a dictionary (see how we call `docker_info.get` below).
-        docker_info = result.get("info", "")
-        return [registry.get("Name", "") for registry in docker_info.get("Registries", {})]
+        return list(regs)
 
     def available_images(self, images, registries, task_vars):
         """Inspect existing images using Skopeo and return all images successfully inspected."""
```
sosiouxme: To reduce the proliferation of tiny methods here, this could be replaced with:

        return [
            image for image in images
            if any(self.is_available_skopeo_image(image, registry, task_vars) for registry in registries)
        ]

roles/openshift_health_checker/test/docker_image_availability_test.py
```diff
@@ -3,26 +3,171 @@
...
+
+@pytest.mark.parametrize("module_failure", [
+    True,
+    False,
+])
+def test_all_images_available_remotely(module_failure):
+    def execute_module(module_name, args, task_vars):
+        return {
+            'docker_image_facts': {'images': [], 'failed': module_failure},
+        }.get(module_name, {'changed': False})
```
sosiouxme: I love a good hash as much as anyone, but this seems like a kinda convoluted way to say:

        if module_name == 'docker_image_facts':
            return {'images': [], 'failed': module_failure}
        return {'changed': False}

roles/openshift_health_checker/test/docker_image_availability_test.py
```diff
@@ -3,26 +3,171 @@
...
+
+
+@pytest.mark.parametrize("module_failure", [
+    True,
+    False,
+])
+def test_all_images_available_remotely(module_failure):
+    def execute_module(module_name, args, task_vars):
+        return {
+            'docker_image_facts': {'images': [], 'failed': module_failure},
```
sosiouxme: Calling it `module_failure` is a little confusing here because this handles executing several modules and the others aren't failing. I think it would be clearer if named `available_locally` and this was `'failed': not available_locally`

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -13,82 +13,92 @@ class DockerImageAvailability(OpenShiftCheck):
...
 
+        openshift_release = get_var(task_vars, "openshift_release", default="latest")
+        openshift_image_tag = get_var(task_vars, "openshift_image_tag")
         is_containerized = get_var(task_vars, "openshift", "common", "is_containerized")
 
-        if is_containerized:
-            images = set(self.containerized_docker_images(image_base_name, openshift_release))
-        else:
-            images = set(self.rpm_docker_images(image_base_name, openshift_release))
+        images = set(self.non_qualified_docker_images(
```
sosiouxme: I find the logic here a little hard to follow, especially since it's not obvious how images are "qualified" or not. A comment might help, some renaming might help, and it would be nice if the `qualified_docker_images` and `non_qualified_docker_images` methods were right under this one.

roles/openshift_health_checker/test/docker_image_availability_test.py
```diff
@@ -3,26 +3,171 @@
...
+        openshift_image_tag='v3.4',
+    ))
+
+    assert not result.get('failed', False)
+
+
+def test_all_images_unavailable():
+    def execute_module(module_name=None, module_args=None, tmp=None, task_vars=None):
+        if module_name == "command":
+            return {
```
sosiouxme: Why spread out?

    return {'failed': True}

roles/openshift_health_checker/test/docker_image_availability_test.py
```diff
@@ -3,26 +3,171 @@
...
+                'failed': True,
+            }
+
+        return {
+            'changed': False,
+        }
+
+    check = DockerImageAvailability(execute_module=execute_module)
+    actual = check.run(tmp=None, task_vars=dict(
```
sosiouxme: For unit testing it seems a little complicated/fragile to run the entire check for every test. Then to make changes/debug you have to follow the logic of the entire check "let's see, this will invoke the X, Y, and Z module so I'll have my stub return something different in each case..."

Not saying you have to refactor the tests for this PR, but something to keep in mind for the future.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -140,40 +152,26 @@ def is_available_skopeo_image(self, image, registry, task_vars):
...
-        return "".join([base["repo"], "/", base["image"]])
-
-    # ensures that the skopeo docker image exists, and updates it
-    # with latest if image was already present locally.
-    def update_skopeo_image(self, task_vars):
-        result = self.module_executor("docker_image", {"name": self.skopeo_image}, task_vars)
-        return result.get("msg", ""), result.get("failed", False), result.get("changed", False)
+    # ensures that the skopeo and python-docker-py packages exist
+    def ensure_dependencies(self, task_vars):
+        result = self.module_executor("yum", {"name": self.dependencies, "state": "latest"}, task_vars)
```
juanvallejo: hm, maybe we could use skopeo from our image only if it is a containerized installation?

juanvallejo:

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
sosiouxme: It's not that skopeo wouldn't be there, it's just that I'm not sure what happens if Ansible tries to yum update to latest on atomic.

Also I need to recreate and file that issue with using the `package` module.

juanvallejo: Ah, okay. For now, I have updated this method to skip using the yum package when running on a containerized install

sosiouxme: Whoah, though: you might still need it on a containerized install. Those can be run against non-Atomic hosts. The issue is specifically with Atomic hosts (is_atomic?).

juanvallejo: Thanks, this is fixed now to use the value of `get_var(task_vars, "openshift", "common", "is_atomic")`

juanvallejo:
@sosiouxme wondering if this is good to merge

sosiouxme:
aos-ci-test

sosiouxme:
aos-ci-test

sosiouxme:
[merge]

sosiouxme:
integration test failed. oops...

juanvallejo:
aos-ci-test

sosiouxme:
aos-ci-test

sosiouxme:
[test] with integration test updates

sosiouxme:
yamllint is going to fail the test/merge this time. maybe we'll get a clean run of aos-ci-test to work with though.

codificat:

codificat:
aos-ci-test

sosiouxme:
should be ready to [merge] again

PR #3787, opened 2017-03-28
Add docker storage check

juanvallejo:
Adds docker storage check to ensure that a docker host is using a supported storage driver,
that its thinpool usage is not close to a specified limit, and that Loopback device is not being used (if devicemapper).

Also updates `docker_image_availability` to reuse code for running only on docker hosts and installing `docker-py` if needed


brenton:
I'll be honest, I don't know the difference between overlay and overlay2.  Were you able to confirm we intend to support both of them?

roles/openshift_health_checker/openshift_checks/docker_storage_check.py
```diff
@@ -0,0 +1,137 @@
...
+            return None
+
+        lv = thinpool_data[0].get("lv")
+        if not lv or not lv[0]:
+            return None
+
+        return lv[0]
+
+    def exec_cmd(self, cmd_str, task_vars):
+        return self.module_executor("command", {
```
rhcarvalho: For new checks you can use `self.execute_module` instead.

juanvallejo:
@brenton 
> Were you able to confirm we intend to support both of them?

I reached out to @stevekuznetsov and he was able to confirm that we will only be supporting `overlay2`. I have updated this check accordingly

roles/openshift_health_checker/openshift_checks/docker_storage_check.py
```diff
@@ -0,0 +1,137 @@
...
+            return True, msg.format(threshold=data_percent)
+
+        if metadata_percent > self.max_thinpool_metadata_usage_percent:
+            msg = "thinpool metadata usage above maximum threshold of {threshold}%%"
+            return True, msg.format(threshold=metadata_percent)
+
+        return False, ""
+
+    def get_lvs_data(self, task_vars):
+        lvs_cmd = "/sbin/lvs --select vg_name=docker --select lv_name=docker-pool --report-format json"
```
juanvallejo: I attempted to look for a core module to use in place of this command, however the closest one I found was http://docs.ansible.com/ansible/lvol_module.html, which does not seem to return the information needed for this check

juanvallejo:
@rhcarvalho @sosiouxme PTAL

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,113 @@
...
+
+        failed, data_percent = self.get_thinpool_data_usage(lv)
+        if failed:
+            return failed, no_data_msg
+
+        failed, metadata_percent = self.get_thinpool_metadata_usage(lv)
+        if failed:
+            return failed, no_data_msg
+
+        if data_percent > float(self.max_thinpool_data_usage_percent):
```
rhcarvalho: When we read `max_thinpool_data_usage_percent` from `task_vars`, as the tests show, it might be a string... and we can't always convert a string to float (not tested).
We could be a bit more careful and handle `TypeError`.

I'd suggest doing that early on when we read from `task_vars`, not here.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,113 @@
...
+            return True, msg.format(threshold=self.max_thinpool_data_usage_percent)
+
+        if metadata_percent > float(self.max_thinpool_metadata_usage_percent):
+            msg = "thinpool metadata usage above maximum threshold of {threshold}%"
+            return True, msg.format(threshold=self.max_thinpool_metadata_usage_percent)
+
+        return False, ""
+
+    def get_lvs_data(self, task_vars):
+        lvs_cmd = "/sbin/lvs --select vg_name=docker --select lv_name=docker-pool --report-format json"
```
rhcarvalho: Can we rely on the names here, `docker` and `docker-pool`?

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,113 @@
...
+        result = self.exec_cmd(lvs_cmd, task_vars)
+        failed = False
```
rhcarvalho: Why do we need this? Looks like it is only used in the last line of the method.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if result.get("failed", False):
+            return True, {}, result.get("msg", "")
```
rhcarvalho: This method returns 3 non-obvious values and they do undocumented...
I'm guessing "did it fail?", "the lvs data I asked for" and "some error message if any". It's a weird API to have failed as the first argument, and might even be unnecessary to have it.

Alternatively, these methods could return what they are meant to return, and if not raise an exception, that would remove a lot of the conditionals.

roles/openshift_health_checker/openshift_checks/docker_storage_driver.py
```diff
@@ -0,0 +1,50 @@
+# pylint: disable=missing-docstring
+from openshift_checks import OpenShiftCheck, get_var
+
+
+class DockerStorageDriver(OpenShiftCheck):
+    """Check Docker storage driver compatibility.
+
+    This check ensures that Docker is using a supported storage driver,
+    and that Loopback is not being used (if using devicemapper).
```
rhcarvalho: Written version of a mental note: this is another check (after memory and disk checks) that will annoy people creating toy clusters...

roles/openshift_health_checker/openshift_checks/docker_storage_driver.py
```diff
@@ -0,0 +1,50 @@
...
+    name = "docker_storage_driver"
+    tags = ["preflight"]
+
+    storage_drivers = ["devicemapper", "overlay2"]
+
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip non-containerized installations."""
+        is_containerized = get_var(task_vars, "openshift", "common", "is_containerized")
+        return super(DockerStorageDriver, cls).is_active(task_vars) and is_containerized
```
rhcarvalho: Looks like both checks ask for a

```python
class ...(Containerized, OpenShiftCheck):
    ...
```

?

roles/openshift_health_checker/openshift_checks/docker_storage_driver.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def run(self, tmp, task_vars):
+        info = self.execute_module("docker_info", {}, task_vars).get("info", {})
+
+        if not self.is_supported_storage_driver(info):
+            msg = "Unsupported Docker storage driver detected. Supported storage drivers: {drivers}"
+            return {"failed": True, "msg": msg.format(drivers=self.storage_drivers)}
```
rhcarvalho: This would print something like `...Supported storage drivers: ['foo', 'bar']`.
We can make it nicer with `', '.join(self.storage_drivers)`.

roles/openshift_health_checker/openshift_checks/docker_storage_driver.py
```diff
@@ -0,0 +1,50 @@
...
+
+        if self.is_using_loopback_device(info):
+            msg = "Use of loopback devices is discouraged. Try running Docker with `--storage-opt dm.thinpooldev`"
+            return {"failed": True, "msg": msg}
+
+        return {"changed": False}
```
rhcarvalho: If we don't focus on the `changed` behavior (we not always return it), maybe it would be more consistent to simply don't ever mention it?

roles/openshift_health_checker/test/docker_storage_test.py
```diff
@@ -0,0 +1,195 @@
...
+        max_thinpool_data_usage_percent=90.0
+    )
+
+    check = DockerStorage(execute_module=execute_module).run(tmp=None, task_vars=task_vars)
+    assert "no thinpool usage data" in check["msg"]
+
+
+
+
+
```
rhcarvalho: Lots of empty lines?

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,113 @@
...
+import json
+
+from openshift_checks import OpenShiftCheck, get_var
+
+
+class DockerStorage(OpenShiftCheck):
+    """Check Docker storage sanity.
+
+    This check ensures that thinpool usage is not
+    close to 100% for a containerized installation.
```
rhcarvalho: The percentage threshold is arbitrary, this comment is too specific.

roles/openshift_health_checker/openshift_checks/docker_storage.py
juanvallejo: hm, upon looking at other installs, `vg_name` and `lv_name` seemed to vary slightly.

A point @sdodson brought up was that we should not be concerned about the total size of a logical volume, as it is automatically expanded once it reaches a certain usage threshold. Instead, we should be looking at the total size of the volume group containing the docker LV.
 
An idea that comes to mind would be to use `vgdisplay` (or the more convenient `vgs`), iterate through all of the volume groups in the output (in my case, I only had one -- `docker`), until we come across a VG  with `docker` in its name. We can then alter this check to only warn when the VG is nearing capacity rather than a specific logical volume under it. WDYT?

cc @sosiouxme

roles/openshift_health_checker/openshift_checks/docker_storage.py
juanvallejo: Thanks, will remove. was leftover after I updated this method

roles/openshift_health_checker/openshift_checks/docker_storage_driver.py
juanvallejo: cc @brenton @sosiouxme

stevekuznetsov: If you're setting up "toy clusters" with multiple nodes, my opinion is that you're out of the "toy cluster" space -- that's more for `oc cluster up`.

roles/openshift_health_checker/openshift_checks/docker_storage_driver.py
juanvallejo: I am not sure what you mean by this, should we create a new mixin rather than read the `is_containerized` value from task_vars?

juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/docker_storage_driver.py
rhcarvalho: That's what I meant, for symmetry with NonContainerized.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,109 @@
...
+
+    name = "docker_storage"
+    tags = ["preflight"]
+
+    max_thinpool_data_usage_percent = 90.0
+    max_thinpool_meta_usage_percent = 90.0
+
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip non-containerized installations."""
```
sosiouxme: Why would we not do this for non-containerized installations too? Everything that's a node needs supportable Docker storage.

juanvallejo:
@sosiouxme thanks for the feedback. Review comment addressed

sosiouxme:
pyunit test is failing right now.

juanvallejo:
@sosiouxme all checks now seem to be passing

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,110 @@
...
+            return msg.format(threshold=self.max_thinpool_data_usage_percent)
+
+        if metadata_percent > self.max_thinpool_meta_usage_percent:
+            msg = "thinpool metadata usage above maximum threshold of {threshold}%"
+            return msg.format(threshold=self.max_thinpool_meta_usage_percent)
+
+        return ""
+
+    def get_lvs_data(self, task_vars):
+        lvs_cmd = "/sbin/lvs --select vg_name=docker --select lv_name=docker-pool --report-format json"
```
sosiouxme: This doesn't seem to work at all on CentOS:

```
# rpm -qf /usr/sbin/lvs
lvm2-2.02.130-5.el7_2.5.x86_64
# /sbin/lvs --select vg_name=docker --select lv_name=docker-pool --report-format json
  Option -S/--select may not be repeated.
  Error during parsing of command line.
# /sbin/lvs --select vg_name=docker  --report-format json
/sbin/lvs: unrecognized option '--reportformat'
  Error during parsing of command line.
```

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,110 @@
...
+        result = self.exec_cmd(lvs_cmd, task_vars)
+
+        if result.get("failed", False):
+            msg = "no thinpool usage data returned by the host: {}"
+            raise OpenShiftCheckException(msg.format(result.get("msg", "")))
+
+        try:
+            data_json = json.loads(result.get("stdout", ""))
+        except ValueError as err:
+            raise OpenShiftCheckException("Invalid JSON value returned by lvs command: {}".format(str(err)))
```
sosiouxme: Isn't this what you get if there are any surprises with lvs? Like, if the VG/LV doesn't even exist?

I got: "Invalid JSON value returned by lvs command: No JSON object could be decoded"

The error message needs to be meaningful to the user. I think we should have a clear error here (or at least not a frustrating or misleading one) if they're using loopback storage or don't have the expected LVM setup.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,110 @@
...
+
+        data = data_json.get("report")
+        if not data:
+            raise OpenShiftCheckException("no thinpool usage data returned by the host.")
```
sosiouxme: Couple of problems here.

1. This isn't too meaningful to the end user. What were you expecting to get and what did you actually get? What should the user do about it?
2. Exact same error message many times here... it won't be obvious which one it balked on.

roles/openshift_health_checker/openshift_checks/docker_storage_driver.py
sosiouxme: I dunno about a `Containerized` mixin but I think this check should be active under the same conditions as the other one.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class DockerStorage(OpenShiftCheck):
+    """Check Docker storage sanity.
+
+    Check for thinpool usage during a containerized installation
+    """
+
+    name = "docker_storage"
+    tags = ["preflight"]
```
sosiouxme: This seems to me like a useful `"health"` check too.

roles/openshift_health_checker/openshift_checks/docker_storage_driver.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def run(self, tmp, task_vars):
+        info = self.execute_module("docker_info", {}, task_vars).get("info", {})
+
+        if not self.is_supported_storage_driver(info):
+            msg = "Unsupported Docker storage driver detected. Supported storage drivers: {drivers}"
+            return {"failed": True, "msg": msg.format(drivers=', '.join(self.storage_drivers))}
+
+        if self.is_using_loopback_device(info):
+            msg = "Use of loopback devices is discouraged. Try running Docker with `--storage-opt dm.thinpooldev`"
```
sosiouxme: s/discouraged/unsupported/

as in, customer support will close your ticket if you haven't set this up right.

I think I would direct them to `docker-storage-setup`

roles/openshift_health_checker/test/docker_storage_test.py
```diff
@@ -0,0 +1,243 @@
+import pytest
+import json
+
+
+from openshift_checks.docker_storage import DockerStorage, OpenShiftCheckException
+
+
+@pytest.mark.parametrize('is_containerized,is_active', [
```
sosiouxme: should test with node group membership too

roles/openshift_health_checker/test/docker_storage_test.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+@pytest.mark.parametrize('stdout,message,failed,extra_words', [
+    (None, "", True, ["no thinpool usage data"]),
+    ("", "", False, ["Invalid JSON value returned by lvs command"]),
+    (None, "invalid response", True, ["invalid response"]),
+    ("invalid", "invalid response", False, ["Invalid JSON value"]),
+])
+def test_get_lvs_data_with_failed_response(stdout, message, failed, extra_words):
+    def execute_module(module_name, args, tmp=None, task_vars=None):
```
sosiouxme: since you use this in every test it seems like you could extract it to a fake method maker.

```
def fake_execute_module(message, failed, stdout):
    def execute_module(module_name, args, tmp=None, task_vars=None):
    ...
    return execute_module
```

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: Actually if I upgrade to lvm2-2.02.166-1.el7_3.4.x86_64 then it works. Hmm... we're certainly expecting RHEL 7.3+ so this ought to be ok, but again, the error message needs to be clear if their only mistake is starting with a 7.2 image.

juanvallejo: Thanks for checking, will handle this error message better

juanvallejo:

sosiouxme:
@juanvallejo unit test is failing, bizarrely only with python3

juanvallejo:
@sosiouxme  will take a look

juanvallejo:
aos-ci-test

sosiouxme:
@juanvallejo I pushed a commit to show you what I meant about more helpful user messages based on lvs output.

sosiouxme:
aos-ci-tests

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: Somehow I did not notice this message. Err... yeah, technically the VG can be named anything and that's probably what we should be looking at. Might want to look in docker config for what it's using. I'll poke around this tomorrow.

sosiouxme:
aos-ci-test

sosiouxme:
@juanvallejo I kept finding things to fiddle with along the way. See commit msg. Review appreciated...

Worthy of note, I'm pretty sure any docker-related checks should only run on hosts that actually use Docker for some reason, so I also updated `docker_image_availability`.

@rhcarvalho review also would be appreciated

sosiouxme:
aos-ci-test

juanvallejo:
@sosiouxme thanks, I am unable to approve this PR, but https://github.com/openshift/openshift-ansible/pull/3787/commits/d63d245c648ad66eaef188c877bf872bb3dc7f0b looks good to me.

roles/openshift_health_checker/openshift_checks/mixins.py
```diff
@@ -8,8 +7,44 @@
...
+
+    def ensure_dependencies(self, task_vars):
+        """
+        Ensure that docker-related packages exist, but not on atomic hosts which should have them.
+        Returns: msg, failed, changed
+        """
+        if get_var(task_vars, "openshift", "common", "is_atomic"):
+            return "", False, False
+
+        # NOTE: we would like to use "package" module but that fails in a way we can't debug just yet.
```
sosiouxme: https://github.com/ansible/ansible/issues/25404

sosiouxme:
@rhcarvalho any concerns with this or are we clear to merge?

rhcarvalho:
> @rhcarvalho any concerns with this or are we clear to merge?

Taking a look now.

roles/openshift_health_checker/openshift_checks/mixins.py
rhcarvalho: I believe it's because "package" is not a module but yet another action plugin (or both?). I've run into this before, looks like the method we invoke to call a module can only resolve to actual modules and not any action plugin, which is a pity.

What we could try is to import and run the action plugin directly.

roles/openshift_health_checker/openshift_checks/mixins.py
```diff
@@ -8,8 +7,44 @@
...
+        result = self.module_executor("yum", {"name": self.dependencies, "state": "present"}, task_vars)
+        msg = result.get("msg", "")
+        if result.get("failed"):
+            if "No package matching" in msg:
+                msg = "Ensure that all required dependencies can be installed via `yum`.\n"
+            msg = (
+                "Unable to install required packages on this host:\n"
+                "    {deps}\n{msg}"
+            ).format(deps=',\n    '.join(self.dependencies), msg=msg)
+        return msg, result.get("failed") or result.get("rc", 0) != 0, result.get("changed")
```
rhcarvalho: Adding one of two variables here would make this more readable.

```python
# ...
failed = result.get("failed") or result.get("rc", 0) != 0
changed = result.get("changed")
return msg, failed, changed
```

rhcarvalho:
@juanvallejo @sosiouxme please update the PR description to include the "bonus" that is coming additional to the new health check ;)

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,181 @@
...
+        if info.get("failed"):
+            return {"failed": True, "changed": changed,
+                    "msg": "Failed to query Docker API. Is docker running on this host?"}
+        if not info.get("info"):  # this would be very strange
+            return {"failed": True, "changed": changed,
+                    "msg": "Docker API query missing info:\n{}".format(json.dumps(info))}
+        info = info["info"]
+
+        # check if the storage driver we saw is valid
+        driver = info.get("Driver", "[NONE]")
```
rhcarvalho: Why `"[NONE]"`?

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,181 @@
...
+            if self._is_dm_using_loopback_device(driver_status):
+                msg = (
+                    "Use of loopback devices with the Docker devicemapper storage driver\n"
+                    "(the default storage configuration) is unsupported in production.\n"
+                    "Please use docker-storage-setup to configure a backing storage volume.\n"
+                    "See http://red.ht/2rNperO for further information."
+                )
+                return {"failed": True, "changed": changed, "msg": msg}
+            result = self._check_dm_usage(driver_status, task_vars)
+            result["changed"] = changed
```
rhcarvalho: Shouldn't this be something like:

```python
result['changed'] = result.get('changed', False) or changed
```

?

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,181 @@
...
+        """
+        vals = dict(
+            vg_free=self._get_vg_free(driver_status.get("Pool Name"), task_vars),
+            data_used=driver_status.get("Data Space Used"),
+            data_total=driver_status.get("Data Space Total"),
+            metadata_used=driver_status.get("Metadata Space Used"),
+            metadata_total=driver_status.get("Metadata Space Total"),
+        )
+
+        for key, value in vals.copy().items():
```
rhcarvalho: Why `.copy()`?

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,181 @@
...
+    def _is_dm_using_loopback_device(driver_status):
+        return bool(driver_status.get("Data loop file"))
+
+    def _check_dm_usage(self, driver_status, task_vars):
+        """
+        For background: under thinpool-backed devicemapper, docker-storage-setup creates
+        an LVM2 LV as a backing storage pool. The LV is "thin" because it does not use
+        all available storage from its VG, instead expanding as needed; so in considering
+        available space, it is necessary to include space remaining in the VG (which is
+        unfortunately not reported by the storage driver in docker info).
```
rhcarvalho: This sounds like a rant, and refers to things that can easily become false statements or be hard to follow.
E.g.:

- "What is 'docker info'?" -- I think this is about our custom Ansible module, but a reader of this code may have no idea what that refers too. In particular, it is NOT about the `docker info` CLI command.
- "unfortunately not reported..." -- might be not reported now, but when people will read this, we can't guarantee that is still true. Instead of talking about what something does or doesn't today, focus on what is required. In the future, someone might understand the requirements and find a way how to implemented it differently.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,181 @@
...
+
+        try:
+            vals["data_threshold"] = float(get_var(task_vars, "max_thinpool_data_usage_percent",
+                                                   default=self.max_thinpool_data_usage_percent))
+            vals["metadata_threshold"] = float(get_var(task_vars, "max_thinpool_metadata_usage_percent",
+                                                       default=self.max_thinpool_meta_usage_percent))
+        except ValueError as err:
+            return {
+                "failed": True,
+                "msg": "Unable to convert thinpool data usage limit to float: {}".format(str(err))
```
rhcarvalho: Hmm, with this we cannot tell if the problem was within `max_thinpool_data_usage_percent` or `max_thinpool_metadata_usage_percent`.

Also, users may not care about "float". Perhaps a better message would explain that the value in the configuration is invalid (and include the bad value in the message).

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,181 @@
...
+                "Docker thinpool data usage percentage {:.1f} "
+                "is higher than threshold {:.1f}."
+            ).format(vals["data_pct_used"], vals["data_threshold"])
+        if vals["metadata_threshold"] < vals["metadata_pct_used"]:
+            vals["failed"] = True
+            msg = msg + "\n" if msg else ""
+            msg += (
+                "\nDocker thinpool metadata usage percentage {:.1f} "
+                "is higher than threshold {:.1f}."
+            ).format(vals["metadata_pct_used"], vals["metadata_threshold"])
```
rhcarvalho: This block/method would benefit from a refactor into smaller methods with a clear name.
At a quick glance it is a confusing wall of text :/

Inspiration:

```python
# ...

if self.usage_below_thresholds(...):
    return {'msg': 'Thinpool usage is below thresholds [include the actual values here]'}

# ...
```

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,181 @@
...
+
+        vals["msg"] = msg or "Thinpool usage is below thresholds."
```
rhcarvalho: Please include the configured threshold and offending value in the message.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,181 @@
...
+        return vals
+
+    def _get_vg_free(self, pool, task_vars):
+        # pool looks like "foo--vg-docker--pool"; vg and lv joined by hyphen, inner hyphens doubled!
+        match = re.match(r'((?:[^-]|--)+)-(?!-)', pool)  # matches up to the first single hyphen
+        if not match:
+            raise OpenShiftCheckException(
+                "Host's docker says it is using LVM storage pool with invalid name '{}'".format(pool)
```
rhcarvalho: What makes it valid / invalid?!

"docker says" sounds strange to me ;)

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -0,0 +1,181 @@
...
+            g=1024**3,
+            t=1024**4,
+            p=1024**5,
+        )
+        string = string or ""
+        match = re.match(r'(\d+(?:\.\d+)?)\s*(\w)?', string)  # float followed by optional unit
+        if not match:
+            raise ValueError("Cannot convert to a byte size: " + string)
+        number, unit = match.groups()
+        return float(number) * units.get(str(unit).lower(), 1)
```
rhcarvalho: If the unit is anything that is not in the `units` dict we may incorrectly assume it is bytes.
I'd rather check explicitly. Empty = bytes, one of the units in the dicts = multiplication factor, otherwise error.

roles/openshift_health_checker/test/docker_storage_test.py
```diff
@@ -0,0 +1,214 @@
+import pytest
+
+
```
rhcarvalho: Too many blank lines, should be at most one separating groups of imports.

roles/openshift_health_checker/test/docker_storage_test.py
```diff
@@ -0,0 +1,214 @@
...
+])
+def test_is_active(is_containerized, group_names, is_active):
+    task_vars = dict(
+        openshift=dict(common=dict(is_containerized=is_containerized)),
+        group_names=group_names,
+    )
+    assert DockerStorage.is_active(task_vars=task_vars) == is_active
+
+
+non_atomic_tv = {"openshift": {"common": {"is_atomic": False}}}
```
rhcarvalho: What does `tv` stand for?! Ah, task_vars!

Please let's use the full name. It's enough cognitive load to have to know what is "task_vars", let alone figure out `_tv` -> `_task_vars` on something that is used only once...

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: *if* the docker info came back without a driver (which seems unlikely, but you know, code defensively) then we're going to say to the user "Docker is using the X driver which isn't supported". What should X be? I don't really care, just thought it would be good to report something distinctly bogus in this unlikely event.

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: because we're modifying `vals` in the loop and I heard it was a bad idea to iterate while changing

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: the bad value is included in the error which is included in the msg. you're right it doesn't tell you which value failed, but i thought the user could figure it out from that. but i suppose that's just lazy; I just didn't want to duplicate the entire try/except for two almost identical things

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: That's the success message. Which isn't really displayed to the user BTW

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: Fair enough :)

This "should not happen" and it would be hard to see how it would be the user's fault if it does, so I blamed it on docker...

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: ok

roles/openshift_health_checker/openshift_checks/mixins.py
sosiouxme: Yep that's what I learned is happening: package is really just an action plugin.

I couldn't readily find a way to run an action plugin from another; it may be easier to port the logic from that plugin.

roles/openshift_health_checker/openshift_checks/mixins.py
sosiouxme: ok

roles/openshift_health_checker/test/docker_storage_test.py
sosiouxme: do we have to pylint test files :)

roles/openshift_health_checker/test/docker_storage_test.py
sosiouxme: ok

sosiouxme:
Yes this will annoy people on dev clusters where they likely will have devicemapper backed by a loopback device since it's the default setup. They can disable the check the same as any check.

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: I whittled it down to a smaller block. Is it better?

sosiouxme:
@rhcarvalho review issues hopefully addressed

sosiouxme:
aos-ci-test

roles/openshift_health_checker/openshift_checks/docker_storage.py
rhcarvalho: Got it.

roles/openshift_health_checker/openshift_checks/docker_storage.py
rhcarvalho: Aha, didn't realize that... I'm used to see a `'msg'` key accompanying a `'failed': True`, didn't think of the other case ;-)

roles/openshift_health_checker/openshift_checks/docker_storage.py
rhcarvalho: I get we (don't / might not) even use the value of 'changed' for checks right now, but we "code defensively" in lots of places, why wouldn't we do it here? If that assumption changes (that `_check_dm_usage` never sets the `'changed'` key), it is unlikely that we'll go look at every caller to stop overriding the value.

roles/openshift_health_checker/openshift_checks/docker_storage.py
rhcarvalho: That's correct...

On a second look, I wonder why we mutate `vals`, and why do we add a key with the `_bytes` prefix? I couldn't find anywhere that actually reads back that key?!

rhcarvalho: Ah, I see I see... we later read "data_used_bytes" and so on.

roles/openshift_health_checker/openshift_checks/mixins.py
```diff
@@ -50,4 +50,6 @@ def ensure_dependencies(self, task_vars):
                 "Unable to install required packages on this host:\n"
                 "    {deps}\n{msg}"
             ).format(deps=',\n    '.join(self.dependencies), msg=msg)
-        return msg, result.get("failed") or result.get("rc", 0) != 0, result.get("changed")
+        failed = result.get("failed", False) or result.get("rc", 0) != 0
+        changed = result.get("changed", False)
+        return msg, failed, changed
```
sosiouxme: sorry, i see i missed doing this one. thanks.

sosiouxme:
aos-ci-test

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: yes, and vals is merged into the result, so that the details are all in the ansible spew if you want to debug what you got

sosiouxme:
[merge]

PR #3791, opened 2017-03-28
Removing initial setting of metrics image prefix/version

ewolinetz:
This was setting the `openshift_hosted_metrics_deployer_prefix` and version to `""` which was then used as the value within `openshift_metrics`

The value should be evaluated only within the role now.


ewolinetz:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1436106

ewolinetz:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

PR #3793, opened 2017-03-28
Integrate GlusterFS storage

jarrpa:
This PR seeks to introduce an enhancement that allows the openshift-ansible installer to either deploy a hoted GlusterFS cluster or access a standalone GlusterFS cluster, both with minimal configuration. It would also allow the automatic creation and use of a GlusterFS volume for storage of a hosted Docker registry.

Depends on: #3832, #3833, #3840


roles/openshift_storage_glusterfs/tasks/main.yml
```diff
@@ -0,0 +1,192 @@
+---
+- name: Create temp directory for doing work in
+  command: mktemp -d /tmp/openshift-glusterfs-ansible-XXXXXX
+  register: mktemp
+  changed_when: False
+  check_mode: no
+
+- name: Copy the admin client config(s)
```
ashcrow: If you are using kubeconfig via the ``oc_*``  modules you no longer need to do this.

roles/openshift_storage_glusterfs/tasks/main.yml
```diff
@@ -0,0 +1,192 @@
...
+    name: "{{ openshift_storage_glusterfs_namespace }}"
+  when: openshift_storage_glusterfs_is_native or openshift_storage_glusterfs_heketi_is_native
+
+- include: glusterfs_deploy.yml
+  when: openshift_storage_glusterfs_is_native
+
+- name: Make sure heketi-client is installed
+  package: name=heketi-client state=present
+
+- name: Delete pre-existing heketi resources
```
ashcrow: Would this be expected by folks installing? To remove previous resources first? I'm not familiar with heketi honestly.

roles/openshift_storage_glusterfs/tasks/main.yml
```diff
@@ -0,0 +1,192 @@
...
+- name: Copy heketi service account resource file
+  copy:
+    src: "{{ openshift.common.examples_content_version }}/{{ item }}"
+    dest: "{{ mktemp.stdout }}/{{ item }}"
+  with_items:
+  - "heketi-service-account.yml"
+  when: openshift_storage_glusterfs_heketi_is_native
+
+- name: Create heketi service account
+  oc_obj:
```
ashcrow: ``oc_serviceaccount`` should be used here if possible. If not, add a comment to note why so folks don't try to port it later.

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
```diff
@@ -0,0 +1,108 @@
+---
+- fail:
+    msg: Only one GlusterFS nodeselector key pair should be provided
+  when: "openshift_storage_glusterfs_nodeselector.keys() | count > 1"
+
+- fail:
+    msg: There must be at least three GlusterFS nodes specified
+  when: "groups.oo_glusterfs_to_config | count < 3"
+
+- name: Delete pre-existing GlusterFS resources
```
abutcher: Is it necessary to remove resources, volumes, configs etc during another application of `openshift_storage_glusterfs`? I'd be worried about the impact these would have on an existing cluster where `playbooks/byo/config.yml` was reran to apply other configuration changes and glusterfs gets reset.

roles/openshift_storage_glusterfs/tasks/heketi_deploy_part2.yml
```diff
@@ -0,0 +1,89 @@
+---
+- name: Create heketi DB volume
+  command: "heketi-cli -s http://{{ openshift_storage_glusterfs_heketi_url }} setup-openshift-heketi-storage --listfile {{ mktemp.stdout }}/heketi-storage.json"
+  register: setup_storage
+  failed_when: False
+
+- name: Copy heketi DB to GlusterFS volume
+  command: "oc create -f {{ mktemp.stdout }}/heketi-storage.json"
```
abutcher: Could this be done with `oc_obj`? If not update to `"{{ openshift.common.client_binary }} create"` and specify the namespace in the arguments.

playbooks/common/openshift-cluster/evaluate_groups.yml
```diff
@@ -29,6 +29,10 @@
       msg: The nfs group must be limited to one host
     when: "{{ (groups[g_nfs_hosts] | default([])) | length > 1 }}"
 
+  - fail:
```
abutcher: Just noting that this matches the previous `fail` conditions in this file for other groups.

jarrpa: I wasn't sure when I started writing this, I was mostly following the patterns the NFS functionality had done. I talked to @abutcher about this, but I've already forgotten the distinction between `l_`, `g_`, and `oo_` variables, my apologies. :) Is there any meaningful distinction?

Realistically this task could probably be removed since the definition of the var being tested is `g_glusterfs_hosts: "{{ groups.glusterfs | default([]) }}"`, unless this playbook is intended to be able to run without first running `cluster_hosts.yml`

roles/lib_openshift/src/lib/base.py
```diff
@@ -397,7 +399,11 @@ def exists(results, _name):
     def find_result(results, _name):
         ''' Find the specified result by name'''
         rval = None
-        for result in results:
+        if len(results) > 0 and 'kind' in results[0] and results[0]['kind'] == 'List':
```
jarrpa: I can understand checking if it's a list, but is the `.get` change necessary?

roles/openshift_storage_glusterfs/tasks/heketi_deploy_part2.yml
jarrpa: I believe it can't be because `heketi-storage.json` is a List of objects, and you can't use `oc_obj` to create without a name. I'll add the namespace, though.

roles/openshift_storage_glusterfs/tasks/main.yml
jarrpa: Kind of. This is to account for trying to re-run the installer if something has gone wrong with the heketi deployment, in which case the main thing to be done is to destroy and recreate the resources. This is part of the attempt to make this playbook vaguely idempotent. The variable `openshift_storage_glusterfs_heketi_wipe` in `when` is False by default, so this task is not normally run unless you know what you're doing.

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
jarrpa: By default they are not run, as `openshift_storage_glusterfs_wipe` is False. However, if something went wrong during the deployment of GlusterFS (e.g. you specified the wrong nodes or the wrong storage devices) there's no easy way to clean that up without recreating the resources.

Further, if something goes wrong while deploying heketi AFTER it has already formatted the storage devices, subsequent attempts to deploy heketi will fail unless the storage devices are wiped clean.

roles/openshift_storage_glusterfs/tasks/main.yml
jarrpa: There are a few `oc` commands where I can't use existing ansible modules, so I'll need this in for now.

roles/openshift_storage_glusterfs/tasks/glusterfs_registry.yml
```diff
@@ -0,0 +1,51 @@
...
+    kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
+    namespace: "{{ openshift.hosted.registry.namespace | default('default') }}"
+    state: present
+    kind: service
+    name: glusterfs-registry-endpoints
+    files:
+    - "{{ mktemp.stdout }}/glusterfs-registry-service.yml"
+
+- name: Check if GlusterFS registry volume exists
+  command: "heketi-cli -s http://{{ openshift_storage_glusterfs_heketi_url }} volume list"
```
jarrpa: Best I know it's always http, yeah.

roles/openshift_storage_glusterfs/tasks/main.yml
ashcrow: Understood. Please add a comment to note that so no one tries to port it until missing items are added.

roles/lib_openshift/src/lib/base.py
ashcrow: The ``.get()`` is not, but it removes the need for ``'kind' in results[0]``.

ashcrow:
bot, add author to whitelist

roles/lib_openshift/src/lib/base.py
jarrpa: Ah! Right. :)

jarrpa:
@ashcrow @abutcher Updated the PR with some additional patches addressing most of your review comments. If the non-GlusterFS parts of this PR look good enough to stand on their own, let me know and I'll create separate PRs for them and mention the dependency on them in this PR.

filter_plugins/oo_filters.py
```diff
@@ -140,7 +140,7 @@ def oo_collect(data_list, attribute=None, filters=None):
             filters   = {'z': 'z'}
             returns [1, 2, 3]
 
-        This also deals with lists of dict as elements.
+        This also deals with lists of lists with dict as elements.
```
ashcrow: 👍

filter_plugins/oo_filters.py
```diff
@@ -171,8 +171,8 @@ def oo_collect(data_list, attribute=None, filters=None):
 
     if filters is not None:
         if not isinstance(filters, dict):
-            raise errors.AnsibleFilterError("oo_collect expects filter to be a"
-                                            " dict")
+            raise errors.AnsibleFilterError(
```
ashcrow: 👍

roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -132,7 +132,10 @@
       kind: pod
       selector: "{{ openshift_hosted_registry_name }}={{ openshift_hosted_registry_namespace }}"
     register: registry_pods
-    until: "registry_pods.results.results[0]['items'] | count > 0 and registry_pods.results.results[0]['items'] | oo_collect(attribute='status.conditions') | oo_collect(attribute='status', filters={'type': 'Ready'}) | map('bool') | select | list | count == openshift_hosted_registry_replicas | int"
+    until:
+    - "registry_pods.results.results[0]['items'] | count > 0"
```
ashcrow: I'm not sure if ``until`` supports lists. If it does 👍

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
```diff
@@ -1,15 +1,14 @@
 ---
-- fail:
+- assert:
```
ashcrow: 👍

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
```diff
@@ -46,11 +44,12 @@
   shell: "{% for line in item.stdout_lines %}{% set fields = line.split() %}{% if fields | count > 1 %}vgremove -fy {{ fields[1] }}; {% endif %}pvremove -fy {{ fields[0] }}; {% endfor %}"
   delegate_to: "{{ item.item }}"
   with_items: "{{ devices_info.results }}"
-  when: openshift_storage_glusterfs_wipe and item.stdout_lines | count > 0
+  when:
```
ashcrow: 👍

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
```diff
@@ -97,12 +93,14 @@
...
-    kubeconfig: "{{ mktemp.stdout }}/admin.kubeconfig"
     namespace: "{{ openshift_storage_glusterfs_namespace }}"
     kind: pod
     state: list
     selector: "glusterfs-node=pod"
   register: glusterfs_pods
-  until: "glusterfs_pods.results.results[0]['items'] | count > 0 and glusterfs_pods.results.results[0]['items'] | oo_collect(attribute='status.conditions') | oo_collect(attribute='status', filters={'type': 'Ready'}) | map('bool') | select | list | count == groups.oo_glusterfs_to_config | count"
+  until:
+  - "glusterfs_pods.results.results[0]['items'] | count > 0"
+  # There must be as many pods with 'Ready' staus  True as there are nodes expecting those pods
```
ashcrow: 👍

roles/openshift_storage_glusterfs/tasks/main.yml
```diff
@@ -135,48 +113,60 @@
...
     src: "{{ openshift.common.examples_content_version }}/topology.json.j2"
     dest: "{{ mktemp.stdout }}/topology.json"
-  when: openshift_storage_glusterfs_is_native and openshift_storage_glusterfs_heketi_topology_load
+  when:
+  - openshift_storage_glusterfs_is_native
+  - openshift_storage_glusterfs_heketi_topology_load
 
 - name: Load heketi topology
-  command: "heketi-cli -s http://{{ openshift_storage_glusterfs_heketi_url }} topology load --json={{ mktemp.stdout }}/topology.json 2>&1"
+  command: "heketi-cli -s http://{{ openshift_storage_glusterfs_heketi_url }} --user admin --secret '{{ openshift_storage_glusterfs_heketi_admin_key }}' topology load --json={{ mktemp.stdout }}/topology.json 2>&1"
```
ashcrow: You may have noted earlier, but if the endpoint could be over TLS it may be nice to store the variable as the full URL.

ashcrow: NM I saw the comment from earlier that it's believed to always be ``http``.

ashcrow:
```('Generated content does not match for /home/travis/build/openshift/openshift-ansible/roles/lib_openshift/src/../library/oc_objectvalidator.py',)```

Looks like the generate.py files need to be executed.

**Comment by jarrpa** [comment]
Oops. Thus regenerated. :)

**Comment by kwoodson** [comment]
aos-ci-test

File: playbooks/common/openshift-glusterfs/config.yml [hunk 16]
```diff
@@ -0,0 +1,20 @@
...
+      port: "24008/tcp"
+    - service: glusterfs_bricks
+      port: "49152-49251/tcp"
+  roles:
+  - os_firewall
+
+- name: Configure GlusterFS
+  hosts: oo_first_master
+  roles:
+  - role: openshift_storage_glusterfs
```
abutcher: We should add a `when` condition here so that `openshift_storage_glusterfs` is not applied to masters when glusterfs isn't being configured. Maybe:

```
when: groups.oo_glusterfs_to_config | default([]) | count > 0
```

jarrpa: @abutcher Good call. Do we need the default given [this](https://github.com/jarrpa/openshift-ansible/blob/86ec8c92ba5f99a2914f863d7a56a7aab406a64d/playbooks/common/openshift-cluster/evaluate_groups.yml#L127-L134)?

abutcher: I think `groups.oo_glusterfs_to_config` won't exist since `g_glusterfs_hosts` is an empty list.

jarrpa: Ah, okay.

jarrpa:
@abutcher @ashcrow @kwoodson Thank you for all the review thus far. :) I've updated the commits again. Since the CI passed, I'll start creating separate PRs for my non-GlusterFS work.

roles/lib_openshift/library/oc_obj.py
```diff
@@ -1533,14 +1533,11 @@ def run_ansible(params, check_mode=False):
...
 
-        if not params['name']:
-            return {'failed': True, 'msg': 'Please specify a name when state is absent|present.'}  # noqa: E501
-
         ########
         # Delete
         ########
         if state == 'absent':
-            if not Utils.exists(api_rval['results'], params['name']):
+            if not api_rval['results'] or not api_rval['results'][0]:
```
ashcrow: If ``api_rval['results']`` could be a true statement and also not a list then this could cause errors.

If that's the case, maybe something like:

```python
if not api_rval['results'] or (isinstance(api_rval['results'], list) and not api_rval['results'][0]):
```

Otherwise it's fine. Just double checking.

roles/lib_openshift/library/oc_adm_ca_server_cert.py
```diff
@@ -919,11 +919,13 @@ def _create(self, fname):
...
         return self.openshift_cmd(['create', '-f', fname])
 
-    def _delete(self, resource, rname, selector=None):
+    def _delete(self, resource, rname=None, selector=None):
         '''call oc delete on a resource'''
-        cmd = ['delete', resource, rname]
+        cmd = ['delete', resource]
         if selector:
             cmd.append('--selector=%s' % selector)
+        elif rname:
```
ashcrow: Are ``rname`` and ``selector`` mutually exclusive? If not:

```python
if rname:
    cmd.append(rname)
if selector:
    cmd.append('--selector={}'.format(selector))
```

filter_plugins/oo_filters.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 # vim: expandtab:tabstop=4:shiftwidth=4
+# pylint: disable=too-many-lines
```
ashcrow: This is indeed a problem, but not one that you introduced.

/cc @sdodson can you flag the fact this is required as tech debt in the backlog?

roles/lib_openshift/library/oc_obj.py
jarrpa: Best I can tell, as the return of `get()` the value of `api_rval['results']` should always be a list.

roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -123,3 +123,56 @@
...
+      state: list
+      kind: pod
+      selector: "{{ openshift_hosted_registry_name }}={{ openshift_hosted_registry_namespace }}"
+    register: registry_pods
+    until:
+    - "registry_pods.results.results[0]['items'] | count > 0"
+    # There must be as many matching pods with 'Ready' status True as there are expected replicas
+    - "registry_pods.results.results[0]['items'] | oo_collect(attribute='status.conditions') | oo_collect(attribute='status', filters={'type': 'Ready'}) | map('bool') | select | list | count == openshift_hosted_registry_replicas | int"
+    delay: 10
+    retries: "{{ (600 / 10) | int }}"
```
ashcrow: nit: Why not 60? 😃

roles/lib_openshift/library/oc_adm_ca_server_cert.py
jarrpa: Yes. if you try to specify both a name and a selector `oc` commands will give you an error. The order I chose was based on the code I saw in `_get()`.

roles/openshift_hosted/tasks/registry/registry.yml
jarrpa: I went with the hard-coded default specified [here](https://github.com/jarrpa/openshift-ansible/blob/b20075e050633445883e89860a5a1029b2fdb15a/roles/openshift_hosted/tasks/registry/registry.yml#L53). ;)

roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -123,3 +123,56 @@
     volume_mounts: "{{ openshift_hosted_registry_volumes }}"
     edits: "{{ openshift_hosted_registry_edits }}"
     force: "{{ True|bool in openshift_hosted_registry_force }}"
+
+- block:
```
kwoodson: I'd prefer this move to a gluster.yml file under the registry directory.

```
- block:
  - include: gluster.yml
  when:  openshift.hosted.registry.storage.kind == 'glusterfs'
```

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
```diff
@@ -0,0 +1,106 @@
...
+  delegate_to: "{{ item }}"
+  with_items: "{{ groups.oo_glusterfs_to_config }}"
+  when: openshift_storage_glusterfs_wipe
+
+- name: Get GlusterFS storage devices state
+  command: "pvdisplay -C --noheadings -o pv_name,vg_name {% for device in hostvars[item].glusterfs_devices %}{{ device }} {% endfor %}"
+  register: devices_info
+  delegate_to: "{{ item }}"
+  with_items: "{{ groups.oo_glusterfs_to_config }}"
+  failed_when: False
```
kwoodson: Does this need to be here?

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
```diff
@@ -0,0 +1,106 @@
...
+  when: openshift_storage_glusterfs_wipe
+
+- name: Clear GlusterFS storage device contents
+  shell: "{% for line in item.stdout_lines %}{% set fields = line.split() %}{% if fields | count > 1 %}vgremove -fy {{ fields[1] }}; {% endif %}pvremove -fy {{ fields[0] }}; {% endfor %}"
```
kwoodson: This makes me a bit nervous.  Would an actual shell script be better here?

roles/openshift_storage_glusterfs/tasks/glusterfs_registry.yml
```diff
@@ -0,0 +1,48 @@
...
+  oc_obj:
+    namespace: "{{ openshift.hosted.registry.namespace | default('default') }}"
+    state: present
+    kind: service
+    name: glusterfs-registry-endpoints
+    files:
+    - "{{ mktemp.stdout }}/glusterfs-registry-service.yml"
+
+- name: Check if GlusterFS registry volume exists
+  command: "heketi-cli -s http://{{ openshift_storage_glusterfs_heketi_url }} --user admin --secret '{{ openshift_storage_glusterfs_heketi_admin_key }}' volume list"
```
kwoodson: Looks like an opportunity for a `heketi-cli` module.  Thoughts?

roles/openshift_storage_glusterfs/tasks/heketi_deploy_part1.yml
```diff
@@ -0,0 +1,41 @@
...
+      HEKETI_ADMIN_KEY: "{{ openshift_storage_glusterfs_heketi_admin_key }}"
+
+- name: Wait for deploy-heketi pod
+  oc_obj:
+    namespace: "{{ openshift_storage_glusterfs_namespace }}"
+    kind: pod
+    state: list
+    selector: "glusterfs=deploy-heketi-pod,deploy-heketi=support"
+  register: heketi_pod
+  until:
```
kwoodson: We should create a `oc_watch` or an `oc_wait` or build in the logic to spin until something on `oc_obj`.  I've seen this pattern in a few places now.

roles/openshift_storage_glusterfs/tasks/heketi_deploy_part2.yml
```diff
@@ -0,0 +1,109 @@
...
+# This is used in the subsequent task
+- name: Copy the admin client config
+  command: >
+    cp {{ openshift.common.config_base }}/master/admin.kubeconfig {{ mktemp.stdout }}/admin.kubeconfig
+  changed_when: False
+  check_mode: no
+
+# Need `command` here because heketi-storage.json contains multiple objects.
+- name: Copy heketi DB to GlusterFS volume
+  command: "{{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig create -f {{ mktemp.stdout }}/heketi-storage.json -n {{ openshift_storage_glusterfs_namespace }}"
```
kwoodson: I think `oc_obj` would work here.  Much prefer it over command.  Your comment above states that command is needed.  What happens with oc_obj is used?

roles/openshift_hosted/tasks/registry/registry.yml
jarrpa: I thought about that too. If I did, I don't think I'd need a block statement, yeah?

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
jarrpa: If we're wiping the storage devices intended for GlusterFS use, then yes. We need to know if a device is marked as a PV (which it may not), and if said PV is part of a VG (which it may not be).

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
jarrpa: I'm not sure it would make things any easier, unless using the `command` module provides any tangible benefit here. Also we'd be passing multi-line input into the script which could be hard to properly test.

roles/openshift_storage_glusterfs/tasks/glusterfs_registry.yml
jarrpa: I think that'd be useful as well, but to do it properly would be somewhat trickier. heketi-cli is just a heketi REST client implemented in golang, thus I would want to write an ansible module that uses it's own python-based REST client. But I also wanted to get something out the door, and it's only ever used in three places thus far, so it doesn't seem like the payoff would be worth the effort.

roles/openshift_storage_glusterfs/tasks/heketi_deploy_part1.yml
jarrpa: Agreed. Didn't want to try and do that here given the timeline I'm on, but that's what I'd rather do. The tricky part is that a watch/wait operation might have to be different for every kind of resource.

roles/openshift_storage_glusterfs/tasks/heketi_deploy_part2.yml
jarrpa: `heketi-storage.json` is a generated file from the above heketi-cli command, which contains a List of multiple resources. Currently, `oc_obj` requires a name for every create operation and thus is not written to handle more than one resource per create. I think it's possible to enhance `oc_obj` to do that, but I wasn't certain if I'd break things if I tried. :)

jarrpa:
@kwoodson Updated patches with the GlusterFS registry stuff in its own file. :)

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
kwoodson: I was specifically referring to the `failed_when`.  What happens when this does fail?

jarrpa: Oh, sorry! pvdisplay will throw an error if one or more of the devices provided is not a PV, but still print valid information for those that are. I wanted to make sure I captured the possible scenario where something went wrong partway through formatting the storage devices.

jarrpa:
@abutcher @ashcrow @kwoodson I added another commit because I tried to run the GlusterFS playbook by itself but realized I was using a filter plugin from the openshift_logging role. I have currently copied the filter I need into the GlusterFS role, but am wondering if there's a better place for it now that it's being used by more than one role.

kwoodson:
@jarrpa, this discussion happened for us (operations) today during some vendoring of openshift-ansible.  The shared location is openshift-ansible/filter_plugins.  The problem I have with putting it there is that the role is no longer self-contained.  This is an issue for consumers as they have to put the specific filter_plugins in the path with precedence.  I'd prefer it if we have them in the individual roles that consume them.

The good:
- Self contained

The bad:
- code duplication

Be curious what the others have to say.

ashcrow:
@jarrpa @kwoodson  IMHO if a filter is used across multiple roles it makes sense to live in ``openshift-ansible/filter_plugins/``. If it's specific to functionality only one module would ever want/need then it should live in it's own filter_plugin location.

At some point it may make sense to create a role that does nothing but hold general filters.

ashcrow:
@jarrpa here is the Travis test failure:

```
py27-flake8 runtests: commands[0] | flake8
./roles/openshift_storage_glusterfs/filter_plugins/openshift_storage_glusterfs.py:5:1: E302 expected 2 blank lines, found 1
./roles/openshift_storage_glusterfs/filter_plugins/openshift_storage_glusterfs.py:13:1: E302 expected 2 blank lines, found 1
```

You can run this locally to test with:

```
$ tox -e py35-flake8
```

jarrpa:
D'oh, forgot to run tox at all on these latest changes. Found a few other small things. Pushed. :)

kwoodson:
@ashcrow @jarrpa, Regarding shared filter_plugins and libraries. 

The role idea that @ashcrow mentioned is something we did recently in the vendoring process of openshift-ansible.  When a role refers to a filter_plugin, it is difficult to get the specific versions (3.3, 3.4, 3.5, 3.6) of filter_plugins and libraries matched up with the correct version of roles.  Operations created a role to house the `/library` and the `/filter_plugins`.  We also add an include to the newly created role that houses the filter_plugins and library directories.  By doing this, when each role refers to a `filter_plugin` or module in the `/library` directory they resolve for us with the correct version.

My vote would be to create a role for these and then if they are used then document it and include it in the `/meta/main.yml` for the role.  This makes the role mostly self contained as well as documented.  Where as the global approach leaves the consumer to deal with the dependencies and ensuring the correct filter_plugins and library directories are available somewhere and in ansible's path.  Then we also run the risk of calling 3.3 or 3.6 filter_plugins when referring to 3.4 roles.

roles/openshift_storage_glusterfs/tasks/heketi_deploy_part2.yml
kwoodson: @jarrpa, 

I believe oc_process could do the same thing here.  That way we move away from copying the kubeconfig step and running command module.  The modules attempt to idempotently handle the operations.

Something like this:
```yaml
- name: Create the gluster template 
  oc_obj:
     name: glustertemplate
     namespace: "{{ openshift_storage_glusterfs_namespace }}"
     files:
     - "{{ mktemp.stdout }}/heketi-storage.json"
     kind: template

-  name: Create all objects from the glustertemplate
   oc_process:
     namespace: "{{ openshift_storage_glusterfs_namespace }}"
     template_name: glustertemplate
     state: present
     create: True
     reconcile: True
```
I'd prefer the oc_obj but if that doesn't work you could try without the oc_obj:  
```yaml
- slurp:
    src: "{{ mktemp.stdout }}/heketi-storage.json"
  register: glusterobjects
-  name: Create all objects from the glustertemplate
   oc_process:
     namespace: "{{ openshift_storage_glusterfs_namespace }}"
     template_name: glustertemplate
     state: present
     create: True
     reconcile: True
     content: "{{ glusterobjects | b64decode }}"
```
Of course, I haven't tested this but we do similar operations in other places.

jarrpa: It's not a Template, though, it's a List.

jarrpa:
@ashcrow @kwoodson If we come to a general concensus on what the ideal strategy is, do we want to implement this now as its own PR before we merge this PR? Also, will it impact the ability to run the GlusterFS playbook by itself (e.g. someone wanting to deploy hosted GlusterFS on an already-running installation)?

ashcrow:
@jarrpa I believe it's out of scope for this work. I feel like this PR will have enough of it's own changes that adding more things to it will slow down the ability to merge. However, the refactor work should be noted in our backlog as a card so we actually remember to do it 😄

jarrpa:
All righty, removed the WIP status and pushed what should be or be close to the final set of commits. :)

ashcrow:
**Note**: Prerequisite PR's all merged.

playbooks/common/openshift-glusterfs/config.yml
ashcrow: Just a reminder note 😄

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
ashcrow: I suggest a comment above this task giving an example of what the shell command would look like. Obviously it won't  be 100% accurate but it should give administrators an idea of what to expect.

roles/openshift_storage_glusterfs/tasks/glusterfs_registry.yml
ashcrow: Agreed. A module would be great but I think that would be a different PR.

ashcrow:
bot, add author to whitelist

ashcrow:
aos-ci-test

ashcrow:
From: openshift-ansible-3-test-matrix/CONTAINERIZED

```
TASK [openshift_node : Start and enable node] **********************************
FAILED - RETRYING: TASK: openshift_node : Start and enable node (1 retries left).
fatal: [10.8.170.33]: FAILED! => {"attempts": 1, "changed": false, "failed": true, "msg": "Unable to start service atomic-openshift-node: Job for atomic-openshift-node.service failed because the control process exited with error code. See \"systemctl status atomic-openshift-node.service\" and \"journalctl -xe\" for details.\n"}
```

roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml
jarrpa: Done!

jarrpa:
@ashcrow @kwoodson @abutcher Can someone shove the CI bot? :)

ashcrow:
aos-ci-test

ashcrow:
[test]

kwoodson:
aos-ci-test

ashcrow:
bot, add author to whitelist

sdodson:
bot, add author to whitelist

ashcrow:
bot, @jarrpa should be whitelisted

ya dummy

kwoodson:
aos-ci-test

kwoodson:
[test]

kwoodson:
[merge]

kwoodson:

kwoodson:
[merge]

smarterclayton:
This broke origin-gce:

```
TASK [openshift_storage_glusterfs : Delete pre-existing GlusterFS config] ******
task path: /usr/share/ansible/openshift-ansible/roles/openshift_storage_glusterfs/tasks/glusterfs_deploy.yml:27
Tuesday 18 April 2017  01:12:47 +0000 (0:00:00.051)       0:15:04.308 ********* 
fatal: [ci-primg212-ig-m-lx3t]: FAILED! => {
    "failed": true, 
    "msg": "'dict object' has no attribute 'oo_glusterfs_to_config'"
}
```

https://ci.openshift.redhat.com/jenkins/job/zz_origin_gce_image/212/consoleFull#-4070161885849c75fe4b0052e114d805c

Can I get an ETA on a fix?

jarrpa:
Just to track the conversation, I opened up #3945 in response to this thread.

PR #3798, opened 2017-03-29
Update six module path to module_utils.six to stay in line with Ansible

dymurray:
A recent PR to Ansible (https://github.com/ansible/ansible/pull/22855) removes the copy of the six module from compat and moves it to module_utils. In openshift-ansible there were references to six in ansible.module_utils and ansible.compat, but this is now obsolete.


sdodson:
@dymurray I think this is no longer necessary? You have to_text where as the current changes don't.

PR #3470, opened 2017-02-23
Add insecure edge termination policy for kibana.

wshearn:
@ewolinetz @jcantrill

PR #3802, opened 2017-03-29
logging routes insecure edge termination policy defaults empty

enoodle:
The logging routes are not edge terminated by default (AFAIK). they are reencrypt and so the can't have their insecure edge termination policy to be "Redirect".

This is the error that I am getting from ansible log when it creates the routes with "Redirect" as the intended InsecureEdgeTerminationPolicy:
```
STDERR:

The Route "logging-kibana" is invalid: spec.tls.insecureEdgeTerminationPolicy: Invalid value: "Redirect": InsecureEdgeTerminationPolicy is only allowed for edge-terminated routes
```

### DIFF
--- a/roles/openshift_logging/defaults/main.yml
+++ b/roles/openshift_logging/defaults/main.yml
@@ -27,7 +27,7 @@ openshift_logging_kibana_proxy_debug: false
 openshift_logging_kibana_proxy_cpu_limit: null
 openshift_logging_kibana_proxy_memory_limit: null
 openshift_logging_kibana_replica_count: 1
-openshift_logging_kibana_edge_term_policy: Redirect
+openshift_logging_kibana_edge_term_policy: null
 
 openshift_logging_kibana_nodeselector: "{{ openshift_hosted_logging_kibana_nodeselector | default('') | map_from_pairs }}"
 openshift_logging_kibana_ops_nodeselector: "{{ openshift_hosted_logging_kibana_ops_nodeselector | default('') | map_from_pairs }}"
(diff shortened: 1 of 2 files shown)


enoodle:
@wshearn  @ewolinetz PTAL, this is related to https://github.com/openshift/openshift-ansible/pull/3470

roles/openshift_logging/defaults/main.yml
```diff
@@ -27,7 +27,7 @@ openshift_logging_kibana_proxy_debug: false
 openshift_logging_kibana_proxy_cpu_limit: null
 openshift_logging_kibana_proxy_memory_limit: null
 openshift_logging_kibana_replica_count: 1
-openshift_logging_kibana_edge_term_policy: Redirect
+openshift_logging_kibana_edge_term_policy: null
```
ewolinetz: Shouldn't this be "None" to reflect the README? or should the README be updated?

enoodle: `None` didn't work for me, I got a message that it is not an allowed value. I will update the readme.

enoodle:
ping @ewolinetz can you take another look?

ewolinetz:
LGTM
@wshearn thoughts on this? You had made the original change

wshearn:
This should work for 3.5 (see https://github.com/openshift/origin/pull/11953 ) is this not the case?

enoodle:
> This should work for 3.5 (see openshift/origin#11953 ) is this not the case?

Great! I guess master is not intended to work with the released version and should always be used with latest available Openshift.

PR #3808, opened 2017-03-30
Support unicode output when dumping yaml

mtnbikenc:
When using `allow_unicode=True` in yaml.dump statements, if the output contains a unicode character it will cause the module to fail.  Using `six.u` to properly pass through all unicode characters. This can have the side effect of passing unicode characters even if they were not intentionally included, such as non-breaking spaces in URLs.  See bug 1416877.

### DIFF
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -21,12 +21,15 @@ import pkg_resources
 import yaml
 
 from ansible import errors
-# pylint no-name-in-module and import-error disabled here because pylint
-# fails to properly detect the packages when installed in a virtualenv
-from ansible.compat.six import string_types  # pylint:disable=no-name-in-module,import-error
-from ansible.compat.six.moves.urllib.parse import urlparse  # pylint:disable=no-name-in-module,import-error
-from ansible.module_utils._text import to_text
 from ansible.parsing.yaml.dumper import AnsibleDumper
+
+# ansible.compat.six goes away with Ansible 2.4
+try:
+    from ansible.compat.six import string_types, u
+    from ansible.compat.six.moves.urllib.parse import urlparse
+except ImportError:
+    from ansible.module_utils.six import string_types, u
+    from ansible.module_utils.six.moves.urllib.parse import urlparse
 
 HAS_OPENSSL = False
 try:
@@ -655,11 +658,11 @@ def to_padded_yaml(data, level=0, indent=2, **kw):
         return ""
 
     try:
-        transformed = yaml.dump(data, indent=indent, allow_unicode=True,
-                                default_flow_style=False,
-                                Dumper=AnsibleDumper, **kw)
+        transformed = u(yaml.dump(data, indent=indent, allow_unicode=True,
+                                  default_flow_style=False,
+                                  Dumper=AnsibleDumper, **kw))
         padded = "\n".join([" " * level * indent + line for line in transformed.splitlines()])
-        return to_text("\n{0}".format(padded))
+        return "\n{0}".format(padded)
     except Exception as my_e:
         raise errors.AnsibleFilterError('Failed to convert: %s' % my_e)
 
--- a/roles/openshift_master_facts/filter_plugins/openshift_master.py
+++ b/roles/openshift_master_facts/filter_plugins/openshift_master.py
@@ -14,9 +14,12 @@ from distutils.version import LooseVersion  # pylint: disable=no-name-in-module,
 from ansible import errors
 from ansible.parsing.yaml.dumper import AnsibleDumper
 from ansible.plugins.filter.core import to_bool as ansible_bool
-# pylint import-error disabled because pylint cannot find the package
-# when installed in a virtualenv
-from ansible.compat.six import string_types  # pylint: disable=no-name-in-module,import-error
+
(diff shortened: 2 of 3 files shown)


mtnbikenc:
aos-ci-test

filter_plugins/oo_filters.py
```diff
@@ -657,7 +657,7 @@ def to_padded_yaml(data, level=0, indent=2, **kw):
     try:
         transformed = yaml.dump(data, indent=indent, allow_unicode=True,
                                 default_flow_style=False,
-                                Dumper=AnsibleDumper, **kw)
+                                Dumper=AnsibleDumper, **kw).decode('utf-8')
```
ashcrow: Sadly, this won't work across versions

### Python 2
```
$ python2 -c "import yaml; print(yaml.dump('{}', allow_unicode=True).decode('utf-8'))"
'{}'
```

### Python 3

```
$ python3 -c "import yaml; print(yaml.dump('{}', allow_unicode=True).decode('utf-8'))"
  File "<string>", line 1, in <module>
AttributeError: 'str' object has no attribute 'decode
```

ashcrow: A possible solution is using ```six.u()```:

```
$ python2 -c "import six, yaml; print(six.u(yaml.dump('{}', allow_unicode=True)))
'{}'
$ python3 -c "import six, yaml; print(six.u(yaml.dump('{}', allow_unicode=True)))
'{}'
```

mtnbikenc:
Wrapping ansible.compat.six in a try/except block because ansible.compat.six goes away with Ansible 2.4.

mtnbikenc:
aos-ci-test

ashcrow:
[test]

ashcrow:
[merge]

ashcrow:
I have a hard time believing this is not a flake.

```
      the server has asked for the client to provide credentials (get builds)
  not to have occurred

  /go/src/github.com/openshift/origin/test/extended/builds/new_app.go:39
```

ashcrow:
@mtnbikenc you may want to try a rebase in case this is a flake from an earlier merge.

mtnbikenc:
aos-ci-test

mtnbikenc:
[merge]

ashcrow:
```
device mapper flake (does not always cause a test failure) https://github.com/openshift/origin/issues/13271
```

ashcrow:
aos-ci-test

ashcrow:
[merge]

sdodson:
[merge]

smarterclayton:
This appears to have resulted in #3978 - what's the change I need to make to how my identity provider var is specified here: https://github.com/openshift/release/blob/master/cluster/test-deploy/data/vars.yaml#L39

Or alternatively, is this unrelated and just looks like it?

ashcrow:
@smarterclayton @mtnbikenc I think we have an issue with PyYAML using ``dump`` instead of ``safe_dump``:

```
In [11]: yaml.dump(u'tag:yaml.org,2002:map')
Out[11]: "!!python/unicode 'tag:yaml.org,2002:map'\n"

Out[19]: "- {!!python/unicode 'challenge': !!python/unicode 'true', !!python/unicode 'kind': !!python/unicode 'AllowAllPasswordIdentityProvider',\n  !!python/unicode 'login': !!python/unicode 'true', !!python/unicode 'name': !!python/unicode 'allow_all'}\n"

```

vs

```
In [15]: yaml.safe_dump('tag:yaml.org,2002:map')
Out[15]: 'tag:yaml.org,2002:map\n...\n'

In [20]: yaml.safe_dump(d)
Out[20]: "- {challenge: 'true', kind: AllowAllPasswordIdentityProvider, login: 'true', name: allow_all}\n"

```

smarterclayton:
Will try locally.

smarterclayton:
The ansible dumper doesn't have the same semantics - what would a proposed patch look like that I can try?

ashcrow:
It looks like the ansible dumper, at least via code invocation, doesn't like Unicode by default:

```
In [1]: d = [{u'name': u'allow_all', u'login': u'true', u'challenge': u'true', u'kind': u'AllowAllPasswordIdentityProvider'}]

In [2]: from ansible.parsing.yaml import dumper

In [3]: dumper.yaml.dump(d)
Out[3]: "- {!!python/unicode 'challenge': !!python/unicode 'true', !!python/unicode 'kind': !!python/unicode 'AllowAllPasswordIdentityProvider',\n  !!python/unicode 'login': !!python/unicode 'true', !!python/unicode 'name': !!python/unicode 'allow_all'}\n"
```
But is happy with it's ``safe_dump``

```
In [9]: dumper.yaml.safe_dump(d)
Out[9]: "- challenge: 'true'\n  kind: AllowAllPasswordIdentityProvider\n  login: 'true'\n  name: allow_all\n"
```

@mtnbikenc 
- [This](http://stackoverflow.com/questions/1950306/pyyaml-dumping-without-tags) post shows a couple possible solutions.
- [Here](http://pyyaml.org/wiki/PyYAMLDocumentation#Python3support) is the results of calls when using PyYAML (not sure about ruamel).

I'll see if I can create a quick diff that *should* work for testing.

ashcrow:
Switching over to the new issue.

PR #3813, opened 2017-03-30
Creating initial proposal doc for review

ewolinetz:
@abutcher @kwoodson for first pass and then we can add the other members of the team

### DIFF
--- a/docs/proposals/role_decomposition.md
+++ b/docs/proposals/role_decomposition.md
@@ -27,7 +27,8 @@ A role should be considered for decomposition if it:
 side by side.
 1) Has different entry points for upgrading and installing a product
 
-Large roles should be responsible for:
+Large roles<sup>1</sup> should be responsible for:
+> 1 or composing playbooks
 
 1) Composing smaller roles to provide a full solution such as an Openshift Master
 1) Ensuring that smaller roles are called in the correct order if necessary
@@ -312,4 +313,41 @@ roles should be able to fall within these guidelines.
 
 # Additional considerations
 
-Playbooks including playbooks (link to rteague's presentation?)
+## Playbooks including playbooks
+In some circumstances it does not make sense to have a composing role but instead
+a playbook would be best for orchestrating the role flow. Decisions made regarding
+playbooks including playbooks will need to be taken into consideration as part of
+defining this process.
+Ref: (link to rteague's presentation?)
+
+## Role dependencies
+We want to make sure that our roles do not have any extra or unnecessary dependencies
+in meta/main.yml without:
+
+1. Proposing the inclusion in a team meeting or as part of the PR review and getting agreement
+1. Documenting in meta/main.yml why it is there and when it was agreed to (date)
+
+## Avoiding overly verbose roles
+When we are splitting our roles up into smaller components we want to ensure we
+avoid creating roles that are, for a lack of a better term, overly verbose. What
+do we mean by that? If we have `openshift_master` as an example, and we were to
+split it up, we would have a component for `etcd`, `docker`, and possibly for
+its rpms/configs. We would want to avoid creating a role that would just create
+certificates as those would make sense to be contained with the rpms and configs.
+Likewise, when it comes to being able to restart the master, we wouldn't have a
+role where that was its sole purpose.
+
+The same would apply for the `etcd` and `docker` roles. Anything that is required
+as part of installing `etcd` such as generating certificates, installing rpms,
+and upgrading data between versions should all be contained within the single
+`etcd` role.
+
+## Enforcing standards
+Certain naming standards like variable names could be verified as part of a Travis
+test. If we were going to also enforce that a role either has tasks or includes
+(for example) then we could create tests for that as well.
+
+## CI tests for individual roles
+If we are able to correctly split up roles, it should be possible to test role
+installations/upgrades like unit tests (assuming they would be able to be installed
+independently of other components).
--- a/docs/proposals/role_decomposition.md
+++ b/docs/proposals/role_decomposition.md
@@ -45,6 +45,7 @@ being installed after other base components such as OCP)
 1) Fail fast when facts it requires are not available or are invalid
 1) "Make it so" based on provided variables and anything that may be required
 as part of doing such (this should include data migrations)
+1) Have a minimal set of dependencies in meta/main.yml, just enough to do its job
 
 ### Example using decomposition of openshift_logging
 
@@ -115,12 +116,40 @@ openshift_logging/tasks/install_logging.yaml
     name: openshift_logging_kibana
   vars:
     generated_certs_dir: "{{openshift.common.config_base}}/logging"
+    openshift_logging_kibana_namespace: "{{ openshift_logging_namespace }}"
+    openshift_logging_kibana_master_url: "{{ openshift_logging_master_url }}"
+    openshift_logging_kibana_master_public_url: "{{ openshift_logging_master_public_url }}"
+    openshift_logging_kibana_image_prefix: "{{ openshift_logging_image_prefix }}"
+    openshift_logging_kibana_image_version: "{{ openshift_logging_image_version }}"
+    openshift_logging_kibana_replicas: "{{ openshift_logging_kibana_replica_count }}"
+    openshift_logging_kibana_es_host: "{{ openshift_logging_es_host }}"
+    openshift_logging_kibana_es_port: "{{ openshift_logging_es_port }}"
+    openshift_logging_kibana_image_pull_secret: "{{ openshift_logging_image_pull_secret }}"
 
 - include_role:
     name: openshift_logging_kibana
   vars:
     generated_certs_dir: "{{openshift.common.config_base}}/logging"
     openshift_logging_kibana_ops_deployment: true
+    openshift_logging_kibana_namespace: "{{ openshift_logging_namespace }}"
+    openshift_logging_kibana_master_url: "{{ openshift_logging_master_url }}"
+    openshift_logging_kibana_master_public_url: "{{ openshift_logging_master_public_url }}"
+    openshift_logging_kibana_image_prefix: "{{ openshift_logging_image_prefix }}"
+    openshift_logging_kibana_image_version: "{{ openshift_logging_image_version }}"
+    openshift_logging_kibana_image_pull_secret: "{{ openshift_logging_image_pull_secret }}"
+    openshift_logging_kibana_es_host: "{{ openshift_logging_es_ops_host }}"
+    openshift_logging_kibana_es_port: "{{ openshift_logging_es_ops_port }}"
+    openshift_logging_kibana_nodeselector: "{{ openshift_logging_kibana_ops_nodeselector }}"
+    openshift_logging_kibana_cpu_limit: "{{ openshift_logging_kibana_ops_cpu_limit }}"
+    openshift_logging_kibana_memory_limit: "{{ openshift_logging_kibana_ops_memory_limit }}"
+    openshift_logging_kibana_hostname: "{{ openshift_logging_kibana_ops_hostname }}"
+    openshift_logging_kibana_replicas: "{{ openshift_logging_kibana_ops_replica_count }}"
+    openshift_logging_kibana_proxy_debug: "{{ openshift_logging_kibana_ops_proxy_debug }}"
+    openshift_logging_kibana_proxy_cpu_limit: "{{ openshift_logging_kibana_ops_proxy_cpu_limit }}"
+    openshift_logging_kibana_proxy_memory_limit: "{{ openshift_logging_kibana_ops_proxy_memory_limit }}"
+    openshift_logging_kibana_cert: "{{ openshift_logging_kibana_ops_cert }}"
+    openshift_logging_kibana_key: "{{ openshift_logging_kibana_ops_key }}"
+    openshift_logging_kibana_ca: "{{ openshift_logging_kibana_ops_ca}}"
   when:
   - openshift_logging_use_ops | bool
 
@@ -130,12 +159,25 @@ openshift_logging/tasks/install_logging.yaml
     name: openshift_logging_curator
   vars:
     generated_certs_dir: "{{openshift.common.config_base}}/logging"
+    openshift_logging_curator_namespace: "{{ openshift_logging_namespace }}"
+    openshift_logging_curator_master_url: "{{ openshift_logging_master_url }}"
+    openshift_logging_curator_image_prefix: "{{ openshift_logging_image_prefix }}"
+    openshift_logging_curator_image_version: "{{ openshift_logging_image_version }}"
+    openshift_logging_curator_image_pull_secret: "{{ openshift_logging_image_pull_secret }}"
 
 - include_role:
     name: openshift_logging_curator
   vars:
     generated_certs_dir: "{{openshift.common.config_base}}/logging"
     openshift_logging_curator_ops_deployment: true
+    openshift_logging_curator_namespace: "{{ openshift_logging_namespace }}"
+    openshift_logging_curator_master_url: "{{ openshift_logging_master_url }}"
+    openshift_logging_curator_image_prefix: "{{ openshift_logging_image_prefix }}"
+    openshift_logging_curator_image_version: "{{ openshift_logging_image_version }}"
+    openshift_logging_curator_image_pull_secret: "{{ openshift_logging_image_pull_secret }}"
+    openshift_logging_curator_cpu_limit: "{{ openshift_logging_curator_ops_cpu_limit }}"
+    openshift_logging_curator_memory_limit: "{{ openshift_logging_curator_ops_memory_limit }}"
+    openshift_logging_curator_nodeselector: "{{ openshift_logging_curator_ops_nodeselector }}"
   when:
   - openshift_logging_use_ops | bool
 
@@ -147,6 +189,45 @@ openshift_logging/tasks/install_logging.yaml
     generated_certs_dir: "{{openshift.common.config_base}}/logging"
 
 - include: update_master_config.yaml
+```
+
+openshift_logging_elasticsearch/meta/main.yaml
+```
+---
+galaxy_info:
+  author: OpenShift Red Hat
+  description: OpenShift Aggregated Logging Elasticsearch Component
+  company: Red Hat, Inc.
+  license: Apache License, Version 2.0
+  min_ansible_version: 2.2
+  platforms:
+  - name: EL
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- role: lib_openshift
+```
+
+openshift_logging/meta/main.yaml
+```
+---
+galaxy_info:
+  author: OpenShift Red Hat
+  description: OpenShift Aggregated Logging
+  company: Red Hat, Inc.
+  license: Apache License, Version 2.0
+  min_ansible_version: 2.2
+  platforms:
+  - name: EL
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- role: lib_openshift
+- role: openshift_facts
 ```
 
 openshift_logging/tasks/install_support.yaml [old]
--- a/docs/proposals/role_decomposition.md
+++ b/docs/proposals/role_decomposition.md
@@ -89,12 +89,29 @@ providing the location of the generated certificates to the individual roles.
 
 #### Snippets
 
-openshift_logging/tasks/install_logging.yaml
-```
-- name: Install logging
-  include: "{{ role_path }}/tasks/install_support.yaml"
-  when: openshift_hosted_logging_install | default(true) | bool
+[openshift_logging/tasks/install_logging.yaml](https://github.com/ewolinetz/openshift-ansible/blob/logging_component_subroles/roles/openshift_logging/tasks/install_logging.yaml)
+```yaml
+- name: Gather OpenShift Logging Facts
+  openshift_logging_facts:
+    oc_bin: "{{openshift.common.client_binary}}"
+    openshift_logging_namespace: "{{openshift_logging_namespace}}"
 
+- name: Set logging project
+  oc_project:
+    state: present
+    name: "{{ openshift_logging_namespace }}"
+
+- name: Create logging cert directory
+  file:
+    path: "{{ openshift.common.config_base }}/logging"
+    state: directory
+    mode: 0755
+  changed_when: False
+  check_mode: no
+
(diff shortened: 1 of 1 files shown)


docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,241 @@
...
+  check_mode: no
+```
+
+# Limitations
+
+There will always be exceptions for some of these rules, however the majority of
+roles should be able to fall within these guidelines.
+
+# Additional considerations
+
```
ewolinetz: role dependencies

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+1) Be deployed independently of other products (this is different than requiring
+being installed after other base components such as OCP)
+1) Be self contained and able to determine facts that it requires to complete
+1) Fail fast when facts it requires are not available or are invalid
+1) "Make it so" based on provided variables and anything that may be required
+as part of doing such (this should include data migrations)
+1) Have a minimal set of dependencies in meta/main.yml, just enough to do its job
+
+### Example using decomposition of openshift_logging
```
tbielawa: Is it possible, maybe with the assistance of the esteemed @mtnbikenc to show graphs visualizing the 'pre-decomposition' and 'post-decomposition' relationships?

Sometimes things are easier to grok when you have pictures.

tbielawa:
General comments I remember from the arch meeting this morning

* Smaller roles that are not `include_role`ing were preferred. They are composable and easier to test.
* *meta-roles*, i.e., roles that only include other roles had a warm reception
  * but I don't think the room came to consensus on that topic

docs/proposals/role_decomposition.md
ewolinetz: Not sure I'm entirely following, from what entry point would you like to see this and at what level of detail? (e.g. task includes, role includes, etc)

tbielawa: Oh. That's more than I considered.

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Small roles
+
+A small role should be able to:
+
+1) Be deployed independently of other products (this is different than requiring
+being installed after other base components such as OCP)
+1) Be self contained and able to determine facts that it requires to complete
+1) Fail fast when facts it requires are not available or are invalid
+1) "Make it so" based on provided variables and anything that may be required
```
ashcrow: Engage. 🌠 🌟 🌠

docs/proposals/role_decomposition.md
ashcrow: I think we create a process. Something like:
```
No extra or unnecessary dependencies in meta/main.yml without

1. Proposing the inclusion in a team meeting, getting agreement.
2. Documenting in the meta/main.yml why it is there and when the team agreed to it
```

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+going forward.
+
+### Large roles
+
+A role should be considered for decomposition if it:
+
+1) Configures/installs more than one product.
+1) Can configure multiple variations of the same product that can live
+side by side.
+1) Has different entry points for upgrading and installing a product
```
ingvagabund: This can be tricky. E.g. we have a role for master installation. We have tasks that should go inside a new master upgrade role. Lot of the tasks are common (configuration files and system services logic, restart handlers, installation/updating deps(rpms, images), etc.). Given the current ansible has still some issues wrt. resource sharing (e.g can not include handlers from different roles properly), it is non-trivial task to break install/upgrade role into two install and upgrade roles without increasing redundancy.

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+
+Large roles should be responsible for:
+
+1) Composing smaller roles to provide a full solution such as an Openshift Master
```
ingvagabund: On one hand I like this idea. It will help to decrease redundancy and role complexity. On the other hand, instead of one `openshift-master` role, we will have:
- `openshift-master-configure` (responsible for various configs, certs, ...) 
- `openshift-master-system-services` (responsible for systemd services a.k.a pulling and running images) 
- `openshift-master-restart` (responsible for restarting and waiting for services/nodes to come back/alive)
- `openshift-master-rest` (responsible for the rest of the tasks)

This is just an illustrative example, not saying it needs to be done this way. Just have the master installation/upgrade tasks in mind as there is a lot of tasks common.

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+1) Ensuring that smaller roles are called in the correct order if necessary
+1) Calling smaller roles with their required variables
```
ingvagabund: We should implement a script that will generate a list of all `free` variables a role have. Plus a list of default variables. It would check the variable has the correct name (e.g. `rolename_variable_name` pattern). And optionally warn as if we are not setting a free variable.

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+1) Performing prerequisite tasks that small roles may depend on being in place
+(openshift_logging certificate generation for example)
+
+### Small roles
+
+A small role should be able to:
+
+1) Be deployed independently of other products (this is different than requiring
+being installed after other base components such as OCP)
+1) Be self contained and able to determine facts that it requires to complete
```
ingvagabund: Maybe, each role should have `check_facts.yml` file that will do exactly that. Would be great to run all the `check_facts.yml` of all roles at the very beginning so we fail ASAP. However, given the `include_task` runs all dependent roles, it is not possible. Plus some variables are set in the middle of the execution.

Maybe, just for testing purpose and to see how much the facts checking needs to be postponed, we could run all the `check_facts` at the beginning or certain times. Just to analyse what is a delay of each role between start and the time the role is actually used. (I guess the paragraph does not make much sense :))

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+1) Fail fast when facts it requires are not available or are invalid
+1) "Make it so" based on provided variables and anything that may be required
+as part of doing such (this should include data migrations)
+1) Have a minimal set of dependencies in meta/main.yml, just enough to do its job
```
ingvagabund: Definitely. E.g. I have been running `etcd_common` role all over and in my environment I don't really need to run `openshift_repos` all over again each time I run it (I always comment the dep out). It saves some time. We should definitely investigate a minimal set of deps for each role so we don't run what we don't need. Alternatively, where reasonable, make the dependency conditional.

docs/proposals/role_decomposition.md
ingvagabund: Again, I really like the idea. I am just sceptic wrt. how we are going to decompose the current roles, not how we are going to compose the current roles (possible new) into large roles. I have noticed some pattern in naming of roles:
- `etcd ` && `openshift_etcd`
- `etcd_client_certificates` && `openshift_etcd_client_certificates`
- `etcd_server_certificates` && `openshift_etcd_server_certificates`
- ...

A role with `openshift_` prefix just adds `openshift_facts` into the `meta/main.yml`, nothing else.
Depending on when the role is used it can have multiple wrappers, each with a list of additional dependencies.

docs/proposals/role_decomposition.md
ingvagabund: Additionaly, the `self-contained` property should be supported by CI.

docs/proposals/role_decomposition.md
kwoodson: I agree that this is the tricky part.  Knowing when and where to separate a role so that non-shared parts are abstracted is difficult.  That being said the ability to configure smaller pieces of a master (your example) IMHO, far outweigh the monolithic approach which requires logic, special variables, hard coded data, etc.  

> Given the current ansible has still some issues wrt. resource sharing (e.g can not include handlers from different roles properly)

Ansible has attempted to make this easier with `include_role` but you are correct that there are issues with it.
> it is non-trivial task to break install/upgrade role into two install and upgrade roles without increasing redundancy.

agreed.

docs/proposals/role_decomposition.md
kwoodson: I think we would come together and chat about where the divisions would be.  There would also have to be some policies setup to define naming schemes for roles.

I like where you are going with this.  Gets the brain cranking away at how to do this.

docs/proposals/role_decomposition.md
kwoodson: :+1: to this idea of having the team agree to dependencies.  I like the idea of using them when necessary and this policy would ensure that this is the case.

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+some requirements that prevent it from utilizing the OCP cert generation service.
+In order to make sure that all components were able to trust one another as they
+did previously, until the cert generation service can be used, the certificate
+generation is being handled within the top level `openshift_logging` role and
+providing the location of the generated certificates to the individual roles.
+
+#### Snippets
+
+openshift_logging/tasks/install_logging.yaml
+```
```
tbielawa: Please switch to using **```yaml** so the presentation format is easier to read.

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+Currently the Aggregated Logging stack generates its own certificates as it has
+some requirements that prevent it from utilizing the OCP cert generation service.
+In order to make sure that all components were able to trust one another as they
+did previously, until the cert generation service can be used, the certificate
+generation is being handled within the top level `openshift_logging` role and
+providing the location of the generated certificates to the individual roles.
+
+#### Snippets
+
+openshift_logging/tasks/install_logging.yaml
```
tbielawa: Can this be a link to that file?

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+- include_role:
+    name: openshift_logging_fluentd
+  vars:
+    generated_certs_dir: "{{openshift.common.config_base}}/logging"
+
+- include: update_master_config.yaml
+```
+
+openshift_logging_elasticsearch/meta/main.yaml
+```
```
tbielawa: **```yaml**

link to file

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- role: lib_openshift
+```
+
+openshift_logging/meta/main.yaml
+```
```
tbielawa: **```yaml**

link to file

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+    - 7
+  categories:
+  - cloud
+dependencies:
+- role: lib_openshift
+- role: openshift_facts
+```
+
+openshift_logging/tasks/install_support.yaml [old]
+```
```
tbielawa: **```yaml**

link to file

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,322 @@
...
+
+- include: generate_clusterrolebindings.yaml
+
+- include: generate_serviceaccounts.yaml
+
+- include: generate_routes.yaml
+```
+
+openshift_logging/tasks/install_support.yaml [new]
+```
```
tbielawa: **```yaml**

link to file

docs/proposals/role_decomposition.md
ewolinetz: You mean like we can/should test each component can install/run on its own, similar to how we do for modules?

ewolinetz: I like the idea of having a commonly named task like `check_facts.yml`.
Would there be a chance of variables changing between roles being called that would potentially alter the original fact value for something like a hosted role? In my mind we would want to gather our facts right before we change anything so we know the current state, but I'm not sure if this is a valid concern or not.

docs/proposals/role_decomposition.md
ewolinetz: I agree, this may be a "pie in the sky" goal and we may not be able to achieve it with all of our roles. Some of our hosted roles like Logging and Metrics will likely be able to follow this, but for `openshift_master` might work better to be a playbook to orchestrate work flow (e.g. mark a node `unschedulable`, pull it out of a HA cluster, etc) and then usesthe smaller component roles to install and configure the parts that make it up

docs/proposals/role_decomposition.md
rhcarvalho: :+1: to the idea as well, with one remark: if proposing on a team meeting would be the only way, this would make it less likely to have contributors external to the CL team propose PRs that involve code in this area.

It could be part of some PR check-list similar to the items currently under `docs/` (style guide, best practices, ...), and non-CL-team contributors could be able to propose the dependency directly on a PR or issue.

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,315 @@
...
+
+## How?
+
+Creating a guide on how to approach breaking up a large role into smaller,
+component based, roles. Also describe how to develop new roles, to avoid creating
+large roles.
+
+## Proposal
+
+Create a new guide or append to the current contributing guide a process for
```
rhcarvalho: Isn't this the guide itself?

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,315 @@
+# Scaffolding for decomposing large roles
```
rhcarvalho: s/Scaffolding/Guide?!

docs/proposals/role_decomposition.md
ingvagabund: > You mean like we can/should test each component can install/run on its own, similar to how we do for modules?

For some roles it makes sense. And if not a role then a subset of roles, e.g. run etcd role (and its dependent etcd-common, certs, ...) to verify the etcd cluster is correctly deployed and running).

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,315 @@
...
+some requirements that prevent it from utilizing the OCP cert generation service.
+In order to make sure that all components were able to trust one another as they
+did previously, until the cert generation service can be used, the certificate
+generation is being handled within the top level `openshift_logging` role and
+providing the location of the generated certificates to the individual roles.
+
+#### Snippets
+
+[openshift_logging/tasks/install_logging.yaml](https://github.com/ewolinetz/openshift-ansible/blob/logging_component_subroles/roles/openshift_logging/tasks/install_logging.yaml)
+```
```
tbielawa: The **```yaml** string needs to be at the beginning of the code block

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,315 @@
...
+- include_role:
+    name: openshift_logging_fluentd
+  vars:
+    generated_certs_dir: "{{openshift.common.config_base}}/logging"
+
+- include: update_master_config.yaml
+```yaml
+
+[openshift_logging_elasticsearch/meta/main.yaml](https://github.com/ewolinetz/openshift-ansible/blob/logging_component_subroles/roles/openshift_logging_elasticsearch/meta/main.yaml)
+```
```
tbielawa: The **```yaml** string needs to be at the beginning of the code block

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,315 @@
...
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- role: lib_openshift
+```yaml
+
+[openshift_logging/meta/main.yaml](https://github.com/ewolinetz/openshift-ansible/blob/logging_component_subroles/roles/openshift_logging/meta/main.yaml)
+```
```
tbielawa: The **```yaml** string needs to be at the beginning of the code block

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,315 @@
...
+    - 7
+  categories:
+  - cloud
+dependencies:
+- role: lib_openshift
+- role: openshift_facts
+```yaml
+
+[openshift_logging/tasks/install_support.yaml - old](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/tasks/install_support.yaml)
+```
```
tbielawa: The **```yaml** string needs to be at the beginning of the code block

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,315 @@
...
+
+
+## Fluentd
+- include_role:
+    name: openshift_logging_fluentd
+  vars:
+    generated_certs_dir: "{{openshift.common.config_base}}/logging"
+
+- include: update_master_config.yaml
+```yaml
```
tbielawa: Should just be **```** at the end of the comment block

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,315 @@
...
+  min_ansible_version: 2.2
+  platforms:
+  - name: EL
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- role: lib_openshift
+```yaml
```
tbielawa: Should just be **```** at the end of the comment block

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,315 @@
...
+  platforms:
+  - name: EL
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- role: lib_openshift
+- role: openshift_facts
+```yaml
```
tbielawa: Should just be **```** at the end of the comment block

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,315 @@
...
+- include: generate_clusterroles.yaml
+
+- include: generate_rolebindings.yaml
+
+- include: generate_clusterrolebindings.yaml
+
+- include: generate_serviceaccounts.yaml
+
+- include: generate_routes.yaml
+```yaml
```
tbielawa: Should just be **```** at the end of the comment block

docs/proposals/role_decomposition.md
ewolinetz: This is more of a proposal so we can all come to a consensus first. We would likely want to have a contribution guide come from this

docs/proposals/role_decomposition.md
ewolinetz: whoops, will fix

docs/proposals/role_decomposition.md
ewolinetz: This is part of what I was hoping to outline with the Small roles section. I agree, we would need to have a defined method of splitting up and composing roles so that we do not have scenarios where, as Jan suggests, we are overly verbose.

My thought is we would have to identify components that make up the larger role and split those out. So if we had a role for `etcd` for example, that role would be responsible for doing everything that was required to get an `etcd` instance stood up -- creating its certificates, ensuring the correct version was installed, performing any upgrades that may be necessary if moving to a newer version, etc.

I would like to get some more feedback/input on how we would identify these components and how we would avoid being overly verbose. Perhaps we would want to have a role for example that contained all commonly used handlers so we could avoid copy/pasta for handlers such as restarting masters (this is used in a couple different places). It could then be clearly defined as a dependency in meta/main.yml

ewolinetz: I'm adding a section under Additional considerations to echo this

docs/proposals/role_decomposition.md
ewolinetz: Would this make sense as a travis test?

ewolinetz:
I tried to encompass/address some of the discussions in "additional considerations"

docs/proposals/role_decomposition.md
ingvagabund: Definitely. More checks we run before a PR is merged, more potential issues postponed.

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,353 @@
...
+Currently the Aggregated Logging stack generates its own certificates as it has
+some requirements that prevent it from utilizing the OCP cert generation service.
+In order to make sure that all components were able to trust one another as they
+did previously, until the cert generation service can be used, the certificate
+generation is being handled within the top level `openshift_logging` role and
+providing the location of the generated certificates to the individual roles.
+
+#### Snippets
+
+[openshift_logging/tasks/install_logging.yaml](https://github.com/ewolinetz/openshift-ansible/blob/logging_component_subroles/roles/openshift_logging/tasks/install_logging.yaml)
```
ewolinetz: update snippet location (point to master)

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,353 @@
...
+## Fluentd
+- include_role:
+    name: openshift_logging_fluentd
+  vars:
+    generated_certs_dir: "{{openshift.common.config_base}}/logging"
+
+- include: update_master_config.yaml
+```
+
+[openshift_logging_elasticsearch/meta/main.yaml](https://github.com/ewolinetz/openshift-ansible/blob/logging_component_subroles/roles/openshift_logging_elasticsearch/meta/main.yaml)
```
ewolinetz: update snippet location (point to master)

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,353 @@
...
+  - name: EL
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- role: lib_openshift
+```
+
+[openshift_logging/meta/main.yaml](https://github.com/ewolinetz/openshift-ansible/blob/logging_component_subroles/roles/openshift_logging/meta/main.yaml)
```
ewolinetz: update snippet location (point to master)

docs/proposals/role_decomposition.md
```diff
@@ -0,0 +1,353 @@
...
+    versions:
+    - 7
+  categories:
+  - cloud
+dependencies:
+- role: lib_openshift
+- role: openshift_facts
+```
+
+[openshift_logging/tasks/install_support.yaml - old](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/tasks/install_support.yaml)
```
ewolinetz: link is dead now

ewolinetz:
@mwoodson @kwoodson we had just gone through this in the latest arch meeting, any other thoughts on this?

ewolinetz:
@sdodson lets merge this bad boy

PR #3816, opened 2017-03-30
add preflight checks integration tests

sosiouxme:
Adds some integration tests that can run preflight checks against inventory made up of throwaway docker containers, and test for expected results.


sosiouxme:
@rhcarvalho The golang part of this is really just running playbooks (in parallel), checking for expected outcomes, and displaying the results nicely. It does seem like the sort of thing that wouldn't be hard to do with pytest instead, and prettier.

Re boilerplate, I think I've reduced the setup part about as much as possible. It is probably worth exploring a callback plugin for teardown rather than having the exception-handling block in every test.

I don't know that we need to block this PR in order to finish that work though. Just a couple small things I'd like to do first:

- rip out the example tests
- get TestPackageAvailabilitySucceeds fixed or skipped

rhcarvalho:
> The golang part of this is really just running playbooks (in parallel), checking for expected outcomes, and displaying the results nicely. It does seem like the sort of thing that could be done with pytest instead, and prettier.

The experiment to be done is to use pytest + pytest-xdist. And look out for flakiness.

sosiouxme:
So, yeah... did those little things.

sosiouxme:
The tests all pass if I run `go test ./test/integration/...`

However with `tox -e integration` all fail immediately with

                     Task:     openshift_sanitize_inventory : Ensure a valid deployment type has been given.
                     Message:  Please set openshift_deployment_type to one of:
                               origin, online, enterprise, atomic-enterprise, openshift-enterprise

I don't really understand what the difference is between the two.

sosiouxme:
Digging a little deeper, tox seems to be running its own ansible, which doesn't work quite the same as the system ansible I get from my path when I run ansible-playbook more directly. I get the same result if I run the vendored ansible. I bumped up debugging and had a closer look. The setup step runs a little differently, and afterward the facts seem to be wiped out. Not sure what's happening here:

```
                Using module file /root/openshift-ansible/.tox/integration/lib/python2.7/site-packages/ansible/modules/core/system/setup.py
                <openshift_ansible_test_84089141491733> ESTABLISH DOCKER CONNECTION FOR USER: root
                <openshift_ansible_test_84089141491733> EXEC ['/bin/docker', 'exec', '-i', u'openshift_ansible_test_84089141491733', u'/bin/sh', '-c', u"/bin/sh -c 'echo ~ && sleep 0'"]
                <openshift_ansible_test_84089141491733> EXEC ['/bin/docker', 'exec', '-i', u'openshift_ansible_test_84089141491733', u'/bin/sh', '-c', u'/bin/sh -c \'( umask 77 && mkdir -p "` echo /root/.ansible
/tmp/ansible-tmp-1491254488.22-118473355611698 `" && echo ansible-tmp-1491254488.22-118473355611698="` echo /root/.ansible/tmp/ansible-tmp-1491254488.22-118473355611698 `" ) && sleep 0\'']
                <openshift_ansible_test_84089141491733> PUT /tmp/tmpGaXB9Q TO /root/.ansible/tmp/ansible-tmp-1491254488.22-118473355611698/setup.py
                <openshift_ansible_test_84089141491733> EXEC ['/bin/docker', 'exec', '-i', u'openshift_ansible_test_84089141491733', u'/bin/sh', '-c', u"/bin/sh -c 'chmod u+x /root/.ansible/tmp/ansible-tmp-14912
54488.22-118473355611698/ /root/.ansible/tmp/ansible-tmp-1491254488.22-118473355611698/setup.py && sleep 0'"]
                <openshift_ansible_test_84089141491733> EXEC ['/bin/docker', 'exec', '-i', u'openshift_ansible_test_84089141491733', u'/bin/sh', '-c', u'/bin/sh -c \'/usr/bin/python /root/.ansible/tmp/ansible-tm
p-1491254488.22-118473355611698/setup.py; rm -rf "/root/.ansible/tmp/ansible-tmp-1491254488.22-118473355611698/" > /dev/null 2>&1 && sleep 0\'']
                ok: [openshift_ansible_test_84089141491733]

                TASK [openshift_sanitize_inventory : Standardize on latest variable names] *****
                task path: /root/openshift-ansible/roles/openshift_sanitize_inventory/tasks/main.yml:2
                ok: [openshift_ansible_test_84089141491733] => {
                    "ansible_facts": {
                        "deployment_type": "", 
                        "openshift_deployment_type": ""
                    }, 
                    "changed": false, 
                    "invocation": {
                        "module_args": {
                            "deployment_type": "", 
                            "openshift_deployment_type": ""
                        }, 
                        "module_name": "set_fact"
                    }
                }
```

sosiouxme:
Tests run if I modify requirements.txt to have `ansible==2.2.1.0` instead of `ansible>=2.2` - argh! Ansible regression apparently...

requirements.txt
```diff
@@ -1,4 +1,6 @@
-ansible>=2.2
+# ansible 2.2.2.0 inexplicably fails integration tests under tox;
```
sosiouxme: I did get tests to run fine from a source checkout of ansible-2.2.2.0 but tox uses what pip gets so it seems there is still some difference.

sosiouxme:
@sdodson this adds integration tests to tox. However they're not running in travis and I assume they should be. How do we make that happen? How do we get them running in the jenkins test/merge also/instead?

Also how do we ensure the local requirements of the tests (docker + golang) are available?

rhcarvalho:
> However they're not running in travis and I assume they should be. How do we make that happen? How do we get them running in the jenkins test/merge also/instead?

Hey, we will probably **not** run them on Travis. Our usage of Travis within the `openshift/*` organization repos should be and stay light, as we have only (IIRC) 2 concurrent builds for all repos.

As integration tests start taking time to run, we don't want to be affecting Travis builds across all projects.

sosiouxme:
Right, makes sense. So I'm curious why travis isn't trying to run integration tests now? They're just another tox environment.

rhcarvalho:
> Right, makes sense. So I'm curious why travis isn't trying to run integration tests now? It's just another environment in tox.ini added in this PR.

Can't see to find it... I have a vague memory that `tox` only run environments matching the list of defaults:
http://tox.readthedocs.io/en/latest/example/basic.html#a-simple-tox-ini-default-environments

So to run an environment "foo" you have to explicitly select it: `tox -e foo`; `tox -e integration`.

test/integration/openshift_health_checker/preflight/playbooks/package_update_dep_missing.yml
```diff
@@ -0,0 +1,24 @@
...
+
+        - include: tasks/enable_repo.yml
+          vars: { repo_name: "break-yum" }
+
+        - action: openshift_health_check
+          args:
+            checks: [ 'package_update' ]
+
+      always:  # destroy the container whether check passed or not
+        - include: ../../teardown_container.yml
```
rhcarvalho: We need to isolate setup and teardown from the actual test, to avoid copy-pasting this snippet all the time.

test/integration/openshift_health_checker/preflight/preflight_test.go
```diff
@@ -0,0 +1,95 @@
+package preflight
+
+import (
+	"testing"
+
+	. ".."
```
rhcarvalho: FTR, this is something I wrote and I don't like it; but to use absolute imports we need to have the code checked out appropriately within a GOPATH.

test/integration/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+```
+tox -e integration
+```
+
+You can also run them more directly, for example to run specific checks
+or increase concurrency:
+
+```
+go test ./test/integration/... -p 8 -run TestPackageUpdateDepMissing
```
sosiouxme: did not know that, interesting. Actually the test script will provide this option, so I'll just leave it out of the README.

requirements.txt
sosiouxme: described at https://github.com/openshift/openshift-ansible/pull/3816#issuecomment-291328973

I really don't want this in here so I'll keep fiddling with it as the CI work comes along.

test/integration/openshift_health_checker/preflight/playbooks/package_update_dep_missing.yml
sosiouxme: Agreed about teardown, I'm pretty sure that can go in a callback plugin.

For setup, you're always going to need some way to specify the various parameters for the host, which is most of what's copy-pasted. It could be a little cleaner if there were a setup action or something like that.

rhcarvalho: We might have better luck using the test framework for that, not YAML.

sosiouxme:
[test]

rhcarvalho:
The unit test failures (incl. Travis) were caused by the way we instantiate action plugins in tests, something that manifested itself only in Ansible 2.3 and fixed in #3919.

[test]

sosiouxme:
docker-py is needed on the host. thinking of running this under tox after all...

sosiouxme:
If I add integration back to tox, it gets run along with all the other tox environments, and before the images to support it are built. Trying something else...

sosiouxme:
@rhcarvalho any idea why pylint is going nuts here all of the sudden? https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_tox/57/console

sosiouxme:
Maybe because pylint released version 1.7

sosiouxme:
pylint 1.7 apparently introduced some new checks. pinning to 1.6.5 that was running before, things pass.

Starting to think we need to pin everything and change versions consciously.

sosiouxme:
So now the integration tests actually run, and fail the same way they do locally for me. So it's not just something nuts about my setup. Progress!

sosiouxme:
The test runner fails under ansible 2.2.2+ because it no longer sets facts on the test hosts the way it's supposed to. I do not think we want to drag the rest of the tests back to ansible 2.2.1 so I created a separate tox config under `test/integration` until this bug is fixed. It's not ideal but it keeps this PR moving.

sosiouxme:
Tests ran fine at https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_tox/60/console

test/integration/run-tests.sh
```diff
@@ -0,0 +1,84 @@
...
+# ./run-tests.sh --prefix=docker.io/openshift/ansible-integration- --parallel=16
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+STARTTIME=$(date +%s)
+source_root=$(dirname "${0}")
+
+prefix="${PREFIX:-openshift-ansible-integration-}"
```
rhcarvalho: I noticed we're defaulting this value in multiple places -- 5 occurrences of `openshift-ansible-integration-` in this PR. What's the case when we would actually use a different prefix?!

test/integration/run-tests.sh
```diff
@@ -0,0 +1,84 @@
...
+fi
+
+# Run the tests. NOTE: "go test" requires a relative path for this purpose.
+# The PWD trick below will only work if cwd is in/above where this script lives.
+retval=0
+IMAGE_PREFIX="${prefix}" env -u GOPATH \
+  go test ./${source_root#$PWD}/... ${gotest_options} \
+  || retval=$?
+
+ENDTIME=$(date +%s); echo "$0 took $(($ENDTIME - $STARTTIME)) seconds"
```
rhcarvalho: I think this is a copy-paste from scripts in Origin... `go test` already prints how long tests took, we could simplify the script and its output by removing this. WDYT?

test/integration/tox.ini
```diff
@@ -0,0 +1,14 @@
+[tox]
```
rhcarvalho: I see we had a different requirement on the Ansible version because of the issue with `delegate_to`, but we could probably still get away with a single tox.ini file in the repository. Earlier this year we've unified /tox.ini and /utils/tox.ini to avoid having to maintain two sets of configuration files.

rhcarvalho:
> Tests ran fine at https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_tox/60/console

@sosiouxme do you know if there's any flag to make the output less verbose?

The interesting line is buried in a ton of useless minutia:

```
ok  	_/data/src/github.com/openshift/openshift-ansible/test/integration/openshift_health_checker/preflight	140.469s
```

More interesting than a GREEN test is to see how RED tests look like, and to make sure that when tests fail people can actually understand the output.

test/integration/run-tests.sh
sosiouxme: That case is when we're building the images separately and pushing them to a shared registry. I wrote this originally expecting that's what we would do, then I learned the shared registry isn't ready yet. Assuming that becomes available, it should just require setting parameters in the Jenkins jobs.

test/integration/run-tests.sh
sosiouxme: Would be fine with me; FWIW this also captures the time to set up the virtualenv. *shrug*

test/integration/tox.ini
sosiouxme: Would be happy to do that. I'm not an expert on tox config :) but it looked to me like that would mean specifying ansible version individually in tox.ini on each environment. Otherwise, if I left the ansible requirement in shared requirements.txt and tried to specify a different one just for integration, it complained about the mismatch. Is there a way to do it?

sosiouxme:
On Tue, Apr 18, 2017 at 7:58 AM, Rodolfo Carvalho <notifications@github.com>
wrote:

> @sosiouxme <https://github.com/sosiouxme> do you know if there's any flag
> to make the output less verbose?
>

Most of the console output is from the ansible run for setup/teardown,
which is from the Jenkins job. If the tests succeeded, it's pretty hard to
find the actual test output, but then, you hardly need to. If the tests
fail, you can generally find a link or some red text. I wish it could be
better sectioned out, perhaps indexed at the top by Jenkins stage.

More interesting than a GREEN test is to see how RED tests look like, and
> to make sure that when tests fail people can actually understand the output.
>

Here's a test failure:
https://ci.openshift.redhat.com/jenkins/job/test_pull_
request_openshift_ansible_tox/58/
I think our output there is pretty good, but we may want to expand the
number of lines of output shown on a failing test.

sosiouxme:
aos-ci-test

sosiouxme:
aos-ci-test

test/integration/run-tests.sh
sosiouxme: ripped out

sosiouxme:
aos-ci-test

sosiouxme:
@rhcarvalho can I get a LGTM or other advice?

test/integration/tox.ini
rhcarvalho: It's good to keep ansible in `requirements.txt`, because for a python developer that's the fist place she would look into and setup runtime requirements in a virtual env (unrelated to tox).

The mismatch is most likely related to how pip works so far... I remember seeing an open issue to allow it to resolve required versions when two sources ask for different versions. The workaround is to call pip twice.

sosiouxme: How do you get tox to call pip twice with different reqs? Or I suppose we could call it manually to update ansible after tox, though that seems odd to me...

sosiouxme: Have pip install ansible==2.2.1.0 be the integration test "command" :)

rhcarvalho: Yes, that's the workaround. I think we've done that before. Odd, yes :)

sosiouxme:
aos-ci-test

test/integration/tox.ini
sosiouxme: @rhcarvalho OK, done! Testing at https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_tox/61/ but local tests went fine so I expect it to work. Travis already passed.

rhcarvalho:
@sosiouxme this needs a rebase.

I haven't had time to try out using pytest-xdist as a test runner. Instead of blocking on this for longer, I'm okay with getting what we have in and following up later.

Please let's do our best to make sure our integration tests don't end up blocking the merge queue.

rhcarvalho:
also, there seems to be room for a `git rebase -i` to combine fixup commits.

sosiouxme:
@rhcarvalho I rebased, and rearranged commits a bit. I think the ones left are the right level of granularity. Will make sure the tests pass, need an approved review to merge...

tox.ini
```diff
@@ -22,3 +24,9 @@ commands =
     generate_validation: python setup.py generate_validation
     # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
     ansible_syntax: python setup.py ansible_syntax
+    # ansible 2.2.2+ unfortunately breaks the integration test runner
+    # because it can no longer set facts on the test docker hosts.
+    # So for now, install separate ansible version for integration.
+    # PR that fixes it: https://github.com/ansible/ansible/pull/23599
+    # Once that PR is available, drop this and use same ansible.
+    integration: pip install ansible==2.2.1.0
```
rhcarvalho: :+1:

tox.ini
```diff
@@ -12,6 +13,7 @@ deps =
     -rrequirements.txt
     -rtest-requirements.txt
     py35-flake8: flake8-bugbear==17.3.0
+    integration: docker-py==1.10.6
```
rhcarvalho: :+1:

sosiouxme:
[merge]

sosiouxme:
aos-ci-test

sosiouxme:
[merge] at last

sosiouxme:
Flake https://github.com/openshift/origin/issues/8571
re[merge]

PR #3817, opened 2017-03-31
Refactor etcd roles

ingvagabund:
- introduce block and simplify when conditions
- introduce config.yml so the self-standing etcd role can be run (e.g. to test etcd cluster deployment)
- remove empty task files
- skip dependency on openshift_etcd_ca
- replace cert paths with their appropriate variables


ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
@ashcrow @mtnbikenc trivial or simple changes only

ashcrow:
@ingvagabund 

```
Unable to start service origin-master: Job for origin-master.service failed because the control process exited with error code.
See "systemctl status origin-master.service" and "journalctl -xe" for details.
```

playbooks/byo/openshift-etcd/config.yml
```diff
@@ -0,0 +1,34 @@
+---
+- name: Create initial host groups for localhost
```
mtnbikenc: Instead of putting these two tasks in this playbook, please include the byo/openshift-cluster/initialize_groups.yml.  This standardizes the tasks, and also flags this playbook as an entry point, which then will run CI syntax checks.

roles/etcd/tasks/etcdctl.yml
```diff
@@ -4,9 +4,9 @@
   when: not openshift.common.is_atomic | bool
 
 - name: Configure etcd profile.d alises
-  copy:
+  template:
+    dest: "/etc/profile.d/etcdctl.sh"
     src: etcdctl.sh
```
mtnbikenc: Can't template from a shell script.

roles/etcd/tasks/etcdctl.yml
```diff
@@ -4,9 +4,9 @@
   when: not openshift.common.is_atomic | bool
 
 - name: Configure etcd profile.d alises
-  copy:
+  template:
```
mtnbikenc: Can't template from a shell script.  The source file needs to be updated to a jinja template if that was the intention.

playbooks/byo/openshift-etcd/config.yml
ingvagabund: Yeah, I had a historical reason for that. Now, it is not needed.

roles/etcd/tasks/etcdctl.yml
ingvagabund: I have moved the file into template directory. But right, it should be renamed to etcdctl.sh.j2

ingvagabund:
aos-ci-test

ingvagabund:
Just squashing

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

roles/etcd/tasks/etcdctl.yml
```diff
@@ -4,9 +4,9 @@
   when: not openshift.common.is_atomic | bool
 
 - name: Configure etcd profile.d alises
-  copy:
-    src: etcdctl.sh
-    dest: /etc/profile.d/etcdctl.sh
+  template:
+    dest: "/etc/profile.d/etcdctl.sh"
```
sdodson: @ingvagabund Were there problems with the static file? this change makes the template depend on a lot of variables which may not be set when running via playbooks/common/openshift-cluster/upgrades/etcd/main.yml

PR #3822, opened 2017-03-31
Example containerized image usage: scheduled certificate check

codificat:
Here is a usage example of the container image for openshift-ansible to perform a certificate check on a regular basis.

There's an additional certificate expiration playbook that  uploads the generated reports to a location in the masters, and an example ScheduledJob that uses it to regularly generate certificate expiration reports.

Usage documentation for the container image is updated to reference them

### DIFF
--- a/README_CONTAINER_IMAGE.md
+++ b/README_CONTAINER_IMAGE.md
@@ -38,4 +38,6 @@ Here is an example of how to run a containerized `openshift-ansible` playbook th
            -e PLAYBOOK_FILE=playbooks/certificate_expiry/default.yaml \
            openshift/openshift-ansible
 
-The [playbook2image examples](https://github.com/aweiteka/playbook2image/tree/master/examples) provide additional information on how to use an image built from it like this one.
+Further usage examples are available in the [examples directory](examples/).
+
+Additional usage information for images built from `playbook2image` like this one can be found in the [playbook2image examples](https://github.com/aweiteka/playbook2image/tree/master/examples).
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,93 @@
+# openshift-ansible usage examples
+
+The primary use of `openshift-ansible` is to install, configure and upgrade OpenShift clusters.
+
+This is typically done by direct invocation of Ansible tools like `ansible-playbook`. This use case is covered in detail in the [OpenShift advanced installation documentation](https://docs.openshift.org/latest/install_config/install/advanced_install.html)
+
+For OpenShift Container Platform there's also an installation utility that wraps `openshift-ansible`. This usage case is covered in the [Quick Installation](https://docs.openshift.com/container-platform/latest/install_config/install/quick_install.html) section of the documentation.
+
+The usage examples below cover use cases other than install/configure/upgrade.
+
+## Container image
+
+The examples below run [openshift-ansible in a container](../README_CONTAINER_IMAGE.md) to perform certificate expiration checks on an OpenShift cluster from pods running on the cluster itself.
+
+You can find more details about the certificate expiration check roles and example playbooks in [the openshift_certificate_expiry role's README](../roles/openshift_certificate_expiry/README.md).
+
+### Job to upload certificate expiration reports
+
+The example `Job` in [certificate-check-upload.yaml](certificate-check-upload.yaml) executes a [Job](https://docs.openshift.org/latest/dev_guide/jobs.html) that checks the expiration dates of the internal certificates of the cluster and uploads HTML and JSON reports to `/etc/origin/certificate_expiration_report` in the masters.
+
+This example uses the [`easy-mode-upload.yaml`](../playbooks/certificate_expiry/easy-mode-upload.yaml) example playbook, which generates reports and uploads them to the masters. The playbook can be customized via environment variables to control the length of the warning period (`CERT_EXPIRY_WARN_DAYS`) and the location in the masters where the reports are uploaded (`COPY_TO_PATH`).
+
+The job expects the inventory to be provided via the *hosts* key of a [ConfigMap](https://docs.openshift.org/latest/dev_guide/configmaps.html) named *inventory*, and the passwordless ssh key that allows connecting to the hosts to be availalbe as *ssh-privatekey* from a [Secret](https://docs.openshift.org/latest/dev_guide/secrets.html) named *sshkey*, so these are created first:
+
+    oc new-project certcheck
+    oc create configmap inventory --from-file=hosts=/etc/ansible/hosts
+    oc secrets new-sshauth sshkey --ssh-privatekey=$HOME/.ssh/id_rsa
+
+Note that `inventory`, `hosts`, `sshkey` and `ssh-privatekey` are referenced by name from the provided example Job definition. If you use different names for the objects/attributes you will have to adjust the Job accordingly.
+
+To create the Job:
+
+    oc create -f examples/certificate-check-upload.yaml
+
+### Scheduled job for certificate expiration report upload
+
+**Note**: This example uses the [ScheduledJob](https://docs.openshift.com/container-platform/3.4/dev_guide/scheduled_jobs.html) object, which has been renamed to [CronJob](https://docs.openshift.org/latest/dev_guide/cron_jobs.html) upstream and is still a Technology Preview subject to further change.
+
+The example `ScheduledJob` in [scheduled-certcheck-upload.yaml](scheduled-certcheck-upload.yaml) does the same as the `Job` example above, but it is scheduled to automatically run every first day of the month (see the `spec.schedule` value in the example).
+
+The job definition is the same and it expects the same configuration: we provide the inventory and ssh key via a ConfigMap and a Secret respectively:
+
+    oc new-project certcheck
+    oc create configmap inventory --from-file=hosts=/etc/ansible/hosts
+    oc secrets new-sshauth sshkey --ssh-privatekey=$HOME/.ssh/id_rsa
+
+And then we create the ScheduledJob:
+
+    oc create -f examples/scheduled-certcheck-upload.yaml
+
+### Job and ScheduledJob to check certificates using volumes
+
+There are two additional examples:
+
+ - A `Job` [certificate-check-volume.yaml](certificate-check-volume.yaml)
+ - A `ScheduledJob` [scheduled-certcheck-upload.yaml](scheduled-certcheck-upload.yaml)
+
+These perform the same work as the two examples above, but instead of uploading the generated reports to the masters they store them in a custom path within the container that is expected to be backed by a [PersistentVolumeClaim](https://docs.openshift.org/latest/dev_guide/persistent_volumes.html), so that the reports are actually written to storage external to the container.
+
+These examples assume that there is an existing `PersistentVolumeClaim` called `certcheck-reports` and they use the  [`html_and_json_timestamp.yaml`](../playbooks/certificate_expiry/html_and_json_timestamp.yaml) example playbook to write timestamped reports into it.
+
+You can later access the reports from another pod that mounts the same volume, or externally via direct access to the backend storage behind the matching `PersistentVolume`.
+
+To run these examples we prepare the inventory and ssh keys as in the other examples:
+
+    oc new-project certcheck
+    oc create configmap inventory --from-file=hosts=/etc/ansible/hosts
+    oc secrets new-sshauth sshkey --ssh-privatekey=$HOME/.ssh/id_rsa
+
+Additionally we allocate a `PersistentVolumeClaim` to store the reports:
+
+	oc create -f - <<PVC
+	---
+	apiVersion: v1
+	kind: PersistentVolumeClaim
+	metadata:
+	  name: certcheck-reports
+	spec:
+	  accessModes:
+		- ReadWriteOnce
+	  resources:
+		requests:
+		  storage: 1Gi
+	PVC
+
+With that we can run the `Job` once:
+
+    oc create -f examples/certificate-check-volume.yaml
+
+or schedule it to run periodically as a `ScheduledJob`:
+
+    oc create -f examples/scheduled-certcheck-volume.yaml
+
--- a/examples/certificate-check-upload.yaml
+++ b/examples/certificate-check-upload.yaml
@@ -1,0 +1,47 @@
+# An example Job to run a certificate check of OpenShift's internal
+# certificate status from within OpenShift.
+#
+# The generated reports are uploaded to a location in the master
+# hosts, using the playbook 'easy-mode-upload.yaml'.
+#
+# This example uses the openshift/openshift-ansible container image.
+# (see README_CONTAINER_IMAGE.md in the top level dir for more details).
+#
+# The following objects are xpected to be configured before the creation
+# of this Job:
+#   - A ConfigMap named 'inventory' with a key named 'hosts' that
+#     contains the the Ansible inventory file
+#   - A Secret named 'sshkey' with a key named 'ssh-privatekey
+#     that contains the ssh key to connect to the hosts
+# (see examples/README.md for more details)
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: certificate-check
+spec:
+  containers:
+  - name: openshift-ansible
+    image: openshift/openshift-ansible
+    env:
+    - name: PLAYBOOK_FILE
+      value: playbooks/certificate_expiry/easy-mode-upload.yaml
+    - name: INVENTORY_FILE
+      value: /tmp/inventory/hosts       # from configmap vol below
+    - name: ANSIBLE_PRIVATE_KEY_FILE    # from secret vol below
+      value: /opt/app-root/src/.ssh/id_rsa/ssh-privatekey
+    - name: CERT_EXPIRY_WARN_DAYS
+      value: "45"      # must be a string, don't forget the quotes
+    volumeMounts:
+    - name: sshkey
+      mountPath: /opt/app-root/src/.ssh/id_rsa
+    - name: inventory
+      mountPath: /tmp/inventory
+  volumes:
+  - name: sshkey
+    secret:
+      secretName: sshkey
+  - name: inventory
+    configMap:
+      name: inventory
+  restartPolicy: Never
--- a/examples/certificate-check-volume.yaml
+++ b/examples/certificate-check-volume.yaml
@@ -1,0 +1,54 @@
+# An example Job to run a certificate check of OpenShift's internal
+# certificate status from within OpenShift.
+#
+# The generated reports are stored in a Persistent Volume using
+# the playbook 'html_and_json_timestamp.yaml'.
+#
+# This example uses the openshift/openshift-ansible container image.
+# (see README_CONTAINER_IMAGE.md in the top level dir for more details).
+#
+# The following objects are xpected to be configured before the creation
+# of this Job:
+#   - A ConfigMap named 'inventory' with a key named 'hosts' that
+#     contains the the Ansible inventory file
+#   - A Secret named 'sshkey' with a key named 'ssh-privatekey
+#     that contains the ssh key to connect to the hosts
+#   - A PersistentVolumeClaim named 'certcheck-reports' where the
+#     generated reports are going to be stored
+# (see examples/README.md for more details)
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: certificate-check
+spec:
+  containers:
+  - name: openshift-ansible
+    image: openshift/openshift-ansible
+    env:
+    - name: PLAYBOOK_FILE
+      value: playbooks/certificate_expiry/html_and_json_timestamp.yaml
+    - name: INVENTORY_FILE
+      value: /tmp/inventory/hosts       # from configmap vol below
+    - name: ANSIBLE_PRIVATE_KEY_FILE    # from secret vol below
+      value: /opt/app-root/src/.ssh/id_rsa/ssh-privatekey
+    - name: CERT_EXPIRY_WARN_DAYS
+      value: "45"      # must be a string, don't forget the quotes
+    volumeMounts:
+    - name: sshkey
+      mountPath: /opt/app-root/src/.ssh/id_rsa
+    - name: inventory
+      mountPath: /tmp/inventory
+    - name: reports
+      mountPath: /var/lib/certcheck
+  volumes:
+  - name: sshkey
+    secret:
+      secretName: sshkey
+  - name: inventory
+    configMap:
+      name: inventory
+  - name: reports
+    persistentVolumeClaim:
+      claimName: certcheck-reports
+  restartPolicy: Never
--- a/examples/scheduled-certcheck-upload.yaml
+++ b/examples/scheduled-certcheck-upload.yaml
@@ -1,0 +1,53 @@
+# An example ScheduledJob to run a regular check of OpenShift's internal
+# certificate status.
+#
+# Each job will upload new reports to a directory in the master hosts
+#
+# The Job specification is the same as 'certificate-check-upload.yaml'
+# and the expected pre-configuration is equivalent.
+# See that Job example and examples/README.md for more details.
+#
+# NOTE: ScheduledJob has been renamed to CronJob in upstream k8s recently. At
+# some point (OpenShift 3.6+) this will have to be renamed to "kind: CronJob"
+# and once the API stabilizes the apiVersion will have to be updated too.
+---
+apiVersion: batch/v2alpha1
+kind: ScheduledJob
+metadata:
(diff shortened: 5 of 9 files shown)


tbielawa:
Looks like some simple linting errors

```
playbooks/certificate_expiry/easy-mode-upload.yaml:
  31:3      error    wrong indentation: expected 4 but found 2  (indentation)
  40:5      error    wrong indentation: expected 6 but found 4  (indentation)
yammlint issues found
ERROR: InvocationError: '/home/travis/build/openshift/openshift-ansible/.tox/py27-yamllint/bin/python setup.py yamllint'
```

codificat:
Lint and rebase done

README_CONTAINER_IMAGE.md
```diff
@@ -38,4 +44,20 @@ Here is an example of how to run a containerized `openshift-ansible` playbook th
            -e PLAYBOOK_FILE=playbooks/certificate_expiry/default.yaml \
            openshift/openshift-ansible
 
-The [playbook2image examples](https://github.com/aweiteka/playbook2image/tree/master/examples) provide additional information on how to use an image built from it like this one.
+#### Scheduled job
+
+The certificate expiration checks in the [standalone container example](#standalone-container) above can also run automatically from within OpenShift on a regular basis using the [example](playbooks/certificate_expiry/scheduled-certcheck.yaml) [CronJob](https://docs.openshift.org/latest/dev_guide/cron_jobs.html) available in the certificate expiration playbooks directory, which will run a playbook that generates certificate expiration reports and upload them to the masters.
```
rhcarvalho: Hmm, might be confusing to talk about CronJob here, but have ScheduledJob in the example YAML? Maybe just like in the YAML we could have a note here to clarify the feature rename.

tbielawa:
> Side note: Apparently the files were already there that way, but I noted that playbooks 
> such as https://github.com/codificat/openshift-ansible/blob/1d4fe6cd0a48b2ea05ec1b267932bf70d554adac/playbooks/certificate_expiry/default.yaml
> have the extension ".yaml", when the Ansible practice and this repo's practice is to use ".yml".

* Ansible docs show an [example directory layout](http://docs.ansible.com/ansible/playbooks_best_practices.html#directory-layout) and they use `yml` there exclusively
* Our [style guide](https://github.com/openshift/openshift-ansible/blob/master/docs/style_guide.adoc#ansible-yaml-file-extension) states repository YAML files **MUST** have a `.yml` extension

I think we can overlook it in this case, as it is consistent with the other contents of the `playbooks/certificate_expiry/` directory.

But looking forward we may need to normalize our repo because we are **not** in compliance with our own style guide:

```bash
$ find . -name '*.yaml' -o -name '*.yml' | grep -v 'tox' | grep -c -E 'yaml$'
143
$ find . -name '*.yaml' -o -name '*.yml' | grep -v 'tox' | grep -c -E 'yml$'
442
```

README_CONTAINER_IMAGE.md
```diff
@@ -38,4 +44,20 @@ Here is an example of how to run a containerized `openshift-ansible` playbook th
            -e PLAYBOOK_FILE=playbooks/certificate_expiry/default.yaml \
            openshift/openshift-ansible
 
-The [playbook2image examples](https://github.com/aweiteka/playbook2image/tree/master/examples) provide additional information on how to use an image built from it like this one.
+#### Scheduled job
```
tbielawa: Is it possible to add some cross referencing to the documentation?

The [cert expiry role README](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_certificate_expiry/README.md) would be improved by adding some sort of section describing what you've done here and linking to your full documentation on the scheduled jobs.

README_CONTAINER_IMAGE.md
codificat: Yes, thanks for noting.
Longer story: as upstream is already at the CronJob level I started this with CronJob - later tests showed that the practical approach at this point would be ScheduledJob and I updated the object but forgot to update the doc.

README_CONTAINER_IMAGE.md
codificat: Good point, will add that x-ref.

About the file extension: I used `.yaml` to match the other files in the directory... but the style guide is quite clear about using `.yml`, and if we look under `playbooks/` there's only 11 files with `.yaml` vs 150+ with `yml`.

As a step towards compliance with the style guide: any objections if I use this PR to rename all the playbooks inside `playbooks/certificate_expiry/` from `.yaml` to `.yml`? (would update the readme accordingly)

rhcarvalho:
> As a step towards compliance with the style guide: any objections if I use this PR to rename all the playbooks inside playbooks/certificate_expiry/ from .yaml to .yml? (would update the readme accordingly)

What's the likelihood of external references to the files?

codificat:
> What's the likelihood of external references to the files?

I'm aware of the `README.md` in the certificate expire roles that @tbielawa referenced above (and that I'll update for x-ref at least), a [reference in openshift-docs](https://github.com/openshift/openshift-docs/blob/master/install_config/redeploying_certificates.adoc), and a [knowledge base entry](https://access.redhat.com/node/2868251)

README_CONTAINER_IMAGE.md
tbielawa: Please please please please please **do not change the extensions** in the `certificate_expiry` directory. Those files referenced in KB articles and online documentation. Changing those extensions will cause support cases to roll in as customers can not run the expiration checks as documented.

tbielawa:
@rhcarvalho @codificat 

:-1: **do not change the file extensions**

recap from review comment above

> Those files referenced in KB articles and online documentation. Changing those extensions will cause support cases to roll in as customers can not run the expiration checks as documented.

codificat:
| :-1: do not change the file extensions

Ok :)  I created issue #3847 to discuss adherence to the style guide, and in this PR I'll use the same extensionions (`.yaml`) as the other files do for consistency. Also, the OpenShift resource definitions are also `.yaml` following the convention in Origin.

I updated the PR with the suggested changes so far:

- created an `examples` directory to host the examples, with docs
- cross references to/from there in the certificate check and top level READMEs.

The examples also include two additional example playbooks around certificate checks: one that uploads reports to the masters and one that uses custom path locations; these are used in the example containerized checks.

I documented these two playbooks in the `README` in the cert expiry role. As the file is getting long I added some headings to it.

@tbielawa PTAL

playbooks/certificate_expiry/html_and_json_timestamp.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+- name: Check cert expirys
+  hosts: nodes:masters:etcd
+  become: yes
+  gather_facts: no
+  vars:
+    openshift_certificate_expiry_generate_html_report: yes
+    openshift_certificate_expiry_save_json_results: yes
+    openshift_certificate_expiry_show_all: yes
+    timestamp: "{{ lookup('pipe', 'date +%Y%m%d') }}"
```
tbielawa: I :heart: this addition, `timestamp`. I wondered about that the first time I gave this PR a review.

roles/openshift_certificate_expiry/README.md
```diff
@@ -104,35 +139,71 @@ $ ansible-playbook -v -i HOSTS /usr/share/ansible/openshift-ansible/playbooks/ce
...
+Uploading a copy of the reports to the masters is one way to make it easy to
+access the reports. Alternatively you can use the
+[role variables](#role-variables) that control the path of the generated reports
+to point to a container volume.
 
-Default behavior:
+With the container use case in mind, this playbook allows control over some
+options via environment variables:
+
+ - `CERT_EXPIRY_WARN_DAYS`: sets `openshift_certificate_expiry_warning_days`,
```
tbielawa: This customization hook is great. I comment about it though because it diverges from the roles default warning window. The role parameters document `openshift_certificate_expiry_warning_days` as having a default value of `30`. Would you consider changing the defaults from `45` to match the `30` for the role? I'm open to hearing why you might not want to as well.

roles/openshift_certificate_expiry/README.md
```diff
@@ -69,11 +68,47 @@ Using the `easy-mode.yaml` playbook will produce:
...
+**From git:**
+```
+$ ansible-playbook -v -i HOSTS playbooks/certificate_expiry/default.yaml
+```
+**From openshift-ansible-playbooks rpm:**
+```
+$ ansible-playbook -v -i HOSTS /usr/share/ansible/openshift-ansible/playbooks/certificate_expiry/default.yaml
+```
+
+> [View This Playbook](../../playbooks/certificate_expiry/default.yaml)
```
tbielawa: I appreciate very much you taking the time to follow the established pattern :smile_cat:

playbooks/certificate_expiry/easy-mode-upload.yaml
```diff
@@ -0,0 +1,40 @@
+# This example enables HTML and JSON reports
+#
+# All certificates (healthy or not) are included in the results
+#
+# Copies of the generated HTML and JSON reports are uploaded to the masters,
+# which is particularly useful when this playbook is run from a container.
+#
+# Optional environment variables to alter the behaviour of the playbook:
+# CERT_EXPIRY_WARN_DAYS:  Lenght of the warning window in days (45)
```
sosiouxme: Lenght => Length

examples/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Here are usage examples to run [openshift-ansible in a container](../README_CONTAINER_IMAGE.md) to perform certificate expiration checks on an OpenShift cluster from pods running on the cluster itself.
+
+You can find more details about the certificate expiration check roles and example playbooks in [the openshift_certificate_expiry role's README](../roles/openshift_certificate_expiry/README.md).
+
+### Job to upload certificate expiration reports
+
+The example `Job` in [certificate-check-upload.yaml](certificate-check-upload.yaml) executes a [Job](https://docs.openshift.org/latest/dev_guide/jobs.html) that checks the expiration dates of the internal certificates of the cluster and uploads HTML and JSON reports to `/etc/origin/certificate_expiration_report` in the masters.
+
+This example use the [`easy-mode-upload.yaml`](../playbooks/certificate_expiry/easy-mode-upload.yaml) example playbook, which generates reports and uploads them to the masters. The playbook can be customized via environment variables to control the length of the warning period (`CERT_EXPIRY_WARN_DAYS`) and the location in the masters where the reports are uploaded (`COPY_TO_PATH`).
```
sosiouxme: This example use => uses

examples/README.md
```diff
@@ -0,0 +1,85 @@
...
+### Job and ScheduledJob to check certificates using volumes
+
+There are two additional examples:
+
+ - A `Job` [certificate-check-volume.yaml](certificate-check-volume.yaml)
+ - A `ScheduledJob` [scheduled-certcheck-upload.yaml](scheduled-certcheck-upload.yaml)
+
+These perform the same work as the two examples above, but instead of uploading the generated reports to the masters they store them in a local path that is expected to correspond to a [PersistentVolumeClaim](https://docs.openshift.org/latest/dev_guide/persistent_volumes.html).
+
+The examples assume that there is a `PersistentVolumeClaim` called `certcheck-reports` and use the  [`html_and_json_timestamp.yaml`](../playbooks/certificate_expiry/html_and_json_timestamp.yaml) example playbook to write timestamped reports into that `PersistentVolume`.
```
sosiouxme: use => uses

examples/README.md
codificat: Thanks, fixing.

examples/README.md
codificat: I think it's "use" here ("the examples [...] use"). I will s/the/these/ and add "they use" for clarity.

playbooks/certificate_expiry/easy-mode-upload.yaml
codificat: thanks, fixing

roles/openshift_certificate_expiry/README.md
codificat: On one hand I wanted to show how to change the default - I could do this by setting the value to the same as the default, but I think it looks more clear if the value differs.

But more importantly, I believe that the default of 30 might be a bit too short in some cases. For example: this playbook is used in the [scheduled certificate check example](../blob/cronjob-example/examples/scheduled-certcheck-upload.yaml) which has a default schedule of "every 1st day of the month", which I think is reasonable. If your CA cert expires on April 1st, the check that runs on March 1st with a default window of 30 days won't warn you, and chances are that you'll have a fun April Fools' day... 

Even in less extreme cases, I believe cluster admins prefer to have a fairly advanced warning in order to have time to plan cert redeployment. In fact I was wondering if the role's default should be increased, with this example's value set even higher :)   What do you think?

Slightly related to this, I think we could improve the role by allowing env var overrides on all defaults - this would allow runtime customization without the need to touch playbooks, which would be very welcome in the container use case... but that's beyond scope of this PR, so if you're ok with that I'll create a separate issue/PR.

tbielawa: You make an excellent point about the default warning window there. I did not think of that condition you described.

:+1: to keeping the new values at 45.

> On one hand I wanted to show how to change the default - I could do this by setting the value to
> the same as the default, but I think it looks more clear if the value differs.

:+1: 

> Even in less extreme cases, I believe cluster admins prefer to have a fairly advanced warning in order
> to have time to plan cert redeployment. In fact I was wondering if the role's default should be increased,
> with this example's value set even higher :) What do you think?

Totally. I'm opening another PR so we can track this for later.

https://github.com/openshift/openshift-ansible/issues/3856

codificat:
Updated with typos fixed and some minor text fixes

examples/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Note that `inventory`, `hosts`, `sshkey` and `ssh-privatekey` are referenced by name from the provided example Job definition. If you use different names for the objects/attributes you will have to adjust the Job accordingly.
+
+To create the Job:
+
+    oc create -f examples/certificate-check-upload.yaml
+
+### Scheduled job for certificate expiration report upload
+
+**Note**: This example uses the [ScheduledJob](https://docs.openshift.com/container-platform/3.4/dev_guide/scheduled_jobs.html) object, which has been renamed to [CronJob](https://docs.openshift.org/latest/dev_guide/cron_jobs.html) upstream and is still a Technology preview subject to further change.
```
rhcarvalho: I see the term "Tech Preview" used more often (with that capitalization), though I don't know if it is official nomenclature...

examples/README.md
```diff
@@ -0,0 +1,85 @@
...
+    oc create -f examples/scheduled-certcheck-upload.yaml
+
+### Job and ScheduledJob to check certificates using volumes
+
+There are two additional examples:
+
+ - A `Job` [certificate-check-volume.yaml](certificate-check-volume.yaml)
+ - A `ScheduledJob` [scheduled-certcheck-upload.yaml](scheduled-certcheck-upload.yaml)
+
+These perform the same work as the two examples above, but instead of uploading the generated reports to the masters they store them in a local path that is expected to correspond to a [PersistentVolumeClaim](https://docs.openshift.org/latest/dev_guide/persistent_volumes.html).
```
rhcarvalho: This is only useful for small / all-in-one clusters, correct?
Normally storing data on a node's local storage is not the way to go.

examples/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+To run these examples we prepare the inventory and ssh keys as in the other examples:
+
+    oc new-project certcheck
+    oc create configmap inventory --from-file=hosts=/etc/ansible/hosts
+    oc secrets new-sshauth sshkey --ssh-privatekey=$HOME/.ssh/id_rsa
+
+Additionally we allocate a `PersistentVolumeClaim` to store the reports:
+
+	cat <<PVC | oc create -f -
```
rhcarvalho: [UUOC](https://en.wikipedia.org/wiki/Cat_%28Unix%29#Useless_use_of_cat)?

```shell
$ oc create -f - <<EOF
{
    ...
}
EOF
```

examples/certificate-check-upload.yaml
```diff
@@ -0,0 +1,40 @@
...
+    image: openshift/openshift-ansible
+    env:
+    - name: PLAYBOOK_FILE
+      value: playbooks/certificate_expiry/easy-mode-upload.yaml
+    - name: INVENTORY_FILE
+      value: /tmp/inventory/hosts       # from configmap vol below
+    - name: ANSIBLE_PRIVATE_KEY_FILE    # from secret vol below
+      value: /opt/app-root/src/.ssh/id_rsa/ssh-privatekey
+    - name: CERT_EXPIRY_WARN_DAYS
+      value: "45"      # must be a string, don't forget the quotes
```
rhcarvalho: The ever curious in me... why must it be a string? Oh my, YAML "magic" types + the way how the library we use parses YAML to fit into a Go string...

examples/certificate-check-volume.yaml
```diff
@@ -0,0 +1,48 @@
...
+#
+# It uses the openshift/openshift-ansible container image.
+# (see README_CONTAINER_IMAGE.md in the top level dir for more details).
+#
+# It expects:
+#   - the Ansible inventory file from the 'hosts' key of a ConfigMap
+#     named 'inventory'
+#   - the ssh key in the 'ssh-privatekey' key of a Secret named 'sshkey'
+#   - a PersistentVolumeClaim named 'certcheck-reports' where the
+#     generated reports are stored
```
rhcarvalho: This header comment in this file and the one in `examples/certificate-check-upload.yaml` look very similar, if not the same in terms of content, but with an apparent unnecessary difference in formatting?

examples/certificate-check-volume.yaml
```diff
@@ -0,0 +1,48 @@
...
+kind: Job
+metadata:
+  name: certificate-check
+spec:
+  containers:
+  - name: openshift-ansible
+    image: openshift/openshift-ansible
+    env:
+    - name: PLAYBOOK_FILE
+      value: playbooks/certificate_expiry/html_and_json_timestamp.yaml
```
rhcarvalho: Is this the only different between the two files? If it is, maybe it should be a single file and passing this value as an argument? Well well, when we have a single object in the YAML we cannot use parameters, but if we use a template object, then we could. Not sure what would be best for usability/maintenance. What do you think?

examples/scheduled-certcheck-upload.yaml
```diff
@@ -0,0 +1,54 @@
...
+# It uses the openshift/openshift-ansible container image.
+#
+# It expects the Ansible inventory file to be available from the 'hosts'
+# key of a ConfigMap named 'inventory' and the ssh key in the 'ssh-privatekey'
+# key of a Secret named 'sshkey'
+# (see README_CONTAINER_IMAGE.md in the top level dir for more details).
+#
+# NOTE: ScheduledJob has been renamed to CronJob in upstream k8s recently. At
+# some point (OpenShift 3.6+) this will have to be renamed to "kind: CronJob"
+# and once the API stabilizes the apiVersion will have to be updated too.
```
rhcarvalho: Oh, actually 3 files that look similar...

examples/scheduled-certcheck-volume.yaml
```diff
@@ -0,0 +1,62 @@
...
+# It expects:
+#   - the Ansible inventory file from the 'hosts' key of a ConfigMap
+#     named 'inventory'
+#   - the ssh key in the 'ssh-privatekey' key of a Secret named 'sshkey'
+#   - a PersistentVolumeClaim named 'certcheck-reports' where the
+#     generated reports are stored
+#
+# NOTE: ScheduledJob has been renamed to CronJob in upstream k8s recently. At
+# some point (OpenShift 3.6+) this will have to be renamed to "kind: CronJob"
+# and once the API stabilizes the apiVersion will have to be updated too.
```
rhcarvalho: 4 similar files!

playbooks/certificate_expiry/easy-mode-upload.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - role: openshift_certificate_expiry
+
+- name: Upload reports to master
+  hosts: masters
+  gather_facts: no
+  vars:
+    destination_path: "{{ lookup('env', 'COPY_TO_PATH') | default('/etc/origin/certificate_expiration_report', true) }}"
+    timestamp: "{{ lookup('pipe', 'date +%Y%m%d') }}"
+  tasks:
+    - name: Create directory in masters
```
rhcarvalho: nit: I think the ansible logs will have "masters" 2x in the same line?
The `hosts` key selects the hosts, I think the task has no business in trying to tell where it is running (easy to get task name and target hosts out of sync).

playbooks/certificate_expiry/easy-mode-upload.yaml
```diff
@@ -0,0 +1,40 @@
...
+      file:
+        path: "{{ destination_path }}"
+        state: directory
+    - name: Copy the reports to the masters
```
rhcarvalho: same as above wrt including the intended host in the task name

roles/openshift_certificate_expiry/README.md
```diff
@@ -69,11 +68,47 @@ Using the `easy-mode.yaml` playbook will produce:
...
 > instead
 
+## Run from a container
+
+The example playbooks that use this role are packaged in the
+[container image for openshift-ansible](../../README_CONTAINER_IMAGE.md), so you
+can run any of them by setting the `PLAYBOOK_FILE` environment variable when
+running an openshift-ansible container.
+
+There are several sample [examples](../../examples/README.md) in the `examples` directory that run certificate check playbooks from a container running on OpenShift.
```
rhcarvalho: s/sample examples/examples ?

examples/README.md
```diff
@@ -0,0 +1,85 @@
+# openshift-ansible usage examples
+
+## Container image for certificate checks
```
codificat: Right, I'm adding an introduction here to clarify.

examples/README.md
codificat: I think "Technology Preview" is the common term, e.g. in the [release notes](https://docs.openshift.com/container-platform/latest/release_notes/ocp_3_4_release_notes.html#ocp-34-technology-preview) and [scope of coverage](https://access.redhat.com/support/offerings/techpreview) documentation.

I'll capitalize the P.

examples/README.md
codificat: What I meant here is a path within the container - backed by a PV... I'll try to clarify that. It's supposed to be useful for big clusters too...

examples/README.md
codificat: yup, thanks

examples/certificate-check-volume.yaml
codificat: there's also an extra volume in the container spec.

I thought about providing a template to cover various use cases - it does get tricky though and it requires more though... for example, the template would be expected to pack all the required resources, and this would include the configmap and secrets, right?

In other words: I left the template for later...

roles/openshift_certificate_expiry/README.md
codificat: :+1:

playbooks/certificate_expiry/easy-mode-upload.yaml
codificat: Well, the `hosts` key of this play restricts the execution to the `masters` group - it should not get out of sync, and I wrote it explicitly for clarity (some hostnames are harder to identify as masters/differenciate from other type of hosts).

Should I remove that anyway?

examples/README.md
rhcarvalho: Ah, I see.

examples/certificate-check-volume.yaml
rhcarvalho: From a usability perspective, having less steps is probably better, in particular if the intermediate steps have no use. But if that means also some other form of complication, what we have is already good, never mind creating a template.

codificat:
Latest comments addressed

examples/README.md
```diff
@@ -0,0 +1,93 @@
+# openshift-ansible usage examples
+
+The primary use of `openshift-ansible` is to install, configure and upgrade OpenShift clusters.
+
+This is typically done by direct invocation of Ansible tools like `ansible-playbook`. This use case is covered in detail in the [OpenShift advanced installation documentation](https://docs.openshift.org/latest/install_config/install/advanced_install.html)
+
+For OpenShift Container Platform there's also an installation utility that wraps `openshift-ansible`. This usage case is covered in the [Quick Installation](https://docs.openshift.com/container-platform/latest/install_config/install/quick_install.html) section of the documentation.
+
+The usage examples below cover use cases other than install/configure/upgrade.
```
rhcarvalho: Great!

rhcarvalho:
aos-ci-test

rhcarvalho:
[merge]

PR #3804, opened 2017-03-29
Wait for firewalld polkit policy to be defined

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1436964
Fixes https://github.com/openshift/openshift-ansible/issues/3213

Issue #3213, opened 2017-01-27
firewalld failure: Action org.fedoraproject.FirewallD1.config.info is not registered


smarterclayton:
Firewalld changes are failing frequently in runs.   I see this 1/12 times in our runs (since we start 4 nodes I see it a lot).

https://ci.openshift.redhat.com/jenkins/job/test_pull_requests_origin_gce/203/consoleFull#-12808334605849c75fe4b0052e114d805c

```
TASK [os_firewall : Add firewalld allow rules] *********************************
Friday 27 January 2017  19:20:01 +0000 (0:00:00.042)       0:12:55.481 ******** 
changed: [ci-prtest203-ig-m-blwm] => (item={u'port': u'10250/tcp', u'service': u'Kubernetes kubelet'})
changed: [ci-prtest203-ig-n-pk5d] => (item={u'port': u'10250/tcp', u'service': u'Kubernetes kubelet'})
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: dbus.exceptions.DBusException: org.freedesktop.PolicyKit1.Error.Failed: Action org.fedoraproject.FirewallD1.config.info is not registered
[...]

PR #3831, opened 2017-04-03
Restart polkitd to workaround a bug in polkitd

sdodson:
Friday I ran into a local case where firewalld was failing as detailed in #3213 even after the changes in #3804 we were able to determine that restarting polkit made the problems go away. Unfortunately polkit lacks a reload function so we must restart it. Polkit should get notifications whenever a new policy is deployed but it looks like maybe it's attempting to load the file while it's still being written to?


sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
[merge]

PR #3833, opened 2017-04-03
oc_obj: allow absent/delete with no name

jarrpa:
Requires a change to oc_base.

### DIFF
--- a/roles/lib_openshift/library/oc_obj.py
+++ b/roles/lib_openshift/library/oc_obj.py
@@ -912,11 +912,13 @@ class OpenShiftCLI(object):
         '''call oc create on a filename'''
         return self.openshift_cmd(['create', '-f', fname])
 
-    def _delete(self, resource, rname, selector=None):
+    def _delete(self, resource, rname=None, selector=None):
         '''call oc delete on a resource'''
-        cmd = ['delete', resource, rname]
+        cmd = ['delete', resource]
         if selector:
             cmd.append('--selector=%s' % selector)
+        elif rname:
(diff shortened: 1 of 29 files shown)


roles/lib_openshift/library/oc_obj.py
```diff
@@ -1531,14 +1533,11 @@ def run_ansible(params, check_mode=False):
         if state == 'list':
             return {'changed': False, 'results': api_rval, 'state': 'list'}
 
-        if not params['name']:
```
ewolinetz: Is there a way we can have this fail if not 'name' or 'selector' ? If we don't have a name we should have a selector, otherwise we wouldn't know what to delete.

jarrpa: I followed the logic found in `_get()`. My guess was that since `oc get` (and similarly `oc delete`) would produce an error if both name and selector were missing that error gets reported back by `return self.openshift_cmd(cmd)`.

kwoodson:
@jarrpa @abutcher @ewolinetz 

I took this work and placed it here: https://github.com/openshift/openshift-ansible/pull/3840

abutcher:
@jarrpa Can we close this PR?

jarrpa:
@abutcher Been waiting for the other one to finish going through. :)

abutcher:
@jarrpa Word. :metal:

kwoodson:
@jarrpa @abutcher  me too =/

jarrpa:
#3840 merged, closing this one.

PR #3830, opened 2017-04-03
Fixed a bug when oc_process executes fail_json.

kwoodson:

PR #3839, opened 2017-04-03
Fixed a bug. Ansible requires a msg param when module.fail_json.

sdodson:
Picks #3830


sdodson:
@kwoodson is this safe to pick into release-1.5, @jupierce is looking for it there

sdodson:
aos-ci-test

PR #3849, opened 2017-04-04
Revert back to iptables as the default for 3.5

sdodson:
Since we don't have a migration strategy to migrate iptables rules to
firewalld during upgrade we should not alter the default.


sdodson:
aos-ci-test

mtnbikenc:
Need to flip the bools around in the README.  Otherwise, LGTM.

abutcher:
aos-ci-test

PR #3851, opened 2017-04-04
Handle iptables.service absence

sdodson:


sdodson:
aos-ci-test

mtnbikenc:
What bug are you seeing here?  Or are you just wanting to switch from `...ignoring` a failed task to reporting `ok:`?  Also, `ignore_errors: true` can be removed when using `failed_when: false`.

sdodson:
Yes, QE filed a bug about it because they said the upgrade failed. But from their logs as far as I can tell everything ran to completion but perhaps the return from ansible-playbook is non zero when there's at least one failure?

sdodson:
https://bugzilla.redhat.com/show_bug.cgi?id=1431976 is the bug

mtnbikenc:
That bug is actually about docker-excluder.  https://bugzilla.redhat.com/show_bug.cgi?id=1431976#c3
Only one host failed, and it didn't stop the install from completing, but there is only one failure noted in the PLAY RECAP.

mtnbikenc:
@sdodson You can close this PR and I'll followup with my PR to get this change in.  The bug mentioned above is not related to these changes.

https://github.com/openshift/openshift-ansible/pull/3871

mtnbikenc:
#3871 has been merged so closing this one.

smarterclayton:
Is this related to https://ci.openshift.redhat.com/jenkins/job/test_branch_origin_extended_conformance_install_update/261/consoleFull#-14246886155849c75fe4b0052e114d805c

?  We're failing install_update right now

sdodson:
Yes, though that doesn't actually fail the playbook and ansible should be returning 0. We've just backported that to release-1.5 https://github.com/openshift/openshift-ansible/pull/3959

PR #3859, opened 2017-04-05
openshift_sanitize_inventory: validate release

sosiouxme:
Complain early if those wacky users use wacky values for openshift_release.

I also took out the no_logs. I haven't seen ansible adding so much annoying stuff to the task description lately, not sure what changed.

### DIFF
--- a/roles/openshift_sanitize_inventory/tasks/main.yml
+++ b/roles/openshift_sanitize_inventory/tasks/main.yml
@@ -1,6 +1,5 @@
 ---
 - name: Standardize on latest variable names
-  no_log: True  # keep task description legible
   set_fact:
     # goal is to deprecate deployment_type in favor of openshift_deployment_type.
     # both will be accepted for now, but code should refer to the new name.
@@ -8,8 +7,15 @@
     deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default | string }}"
     openshift_deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default | string }}"
 
+- name: Abort when deployment type is invalid
+  # this variable is required; complain early and clearly if it is invalid.
+  when: openshift_deployment_type not in known_openshift_deployment_types
+  fail:
+    msg: |-
+      Please set openshift_deployment_type to one of:
+      {{ known_openshift_deployment_types | join(', ') }}
+
 - name: Normalize openshift_release
-  no_log: True  # keep task description legible
   set_fact:
     # Normalize release if provided, e.g. "v3.5" => "3.5"
     # Currently this is not required to be defined for all installs, and the
@@ -19,10 +25,11 @@
     openshift_release: "{{ openshift_release | string | regex_replace('^v', '') }}"
   when: openshift_release is defined
 
-- name: Ensure a valid deployment type has been given.
-  # this variable is required; complain early and clearly if it is invalid.
-  when: openshift_deployment_type not in known_openshift_deployment_types
+- name: Abort when openshift_release is invalid
+  when:
+    - openshift_release is defined
+    - not openshift_release | match('\d+(\.\d+){1,3}$')
   fail:
     msg: |-
-      Please set openshift_deployment_type to one of:
-      {{ known_openshift_deployment_types | join(', ') }}
+      openshift_release is "{{ openshift_release }}" which is not a valid version string.
+      Please set it to a version string like "3.4".


roles/openshift_sanitize_inventory/tasks/main.yml
```diff
@@ -19,10 +25,11 @@
     openshift_release: "{{ openshift_release | string | regex_replace('^v', '') }}"
   when: openshift_release is defined
 
-- name: Ensure a valid deployment type has been given.
-  # this variable is required; complain early and clearly if it is invalid.
-  when: openshift_deployment_type not in known_openshift_deployment_types
+- name: Validate openshift_release looks like a version
+  when: |
+    openshift_release is defined
+    and not openshift_release | match('\d+(\.\d+){1,3}$')
```
mtnbikenc: This could be written as:
```
when:
- openshift_release is defined
- not openshift_release | match('\d+(\.\d+){1,3}$')
```

sosiouxme: ah yes, I guess that's more idiomatic.

speaking of which... was there agreement on fail vs assert?

mtnbikenc: My preference is assert, in the sense that it is declaring the desired truthful state, vice failing on the inverse.  Many times this removes the not's.

sosiouxme: assert doesn't improve things here, it actually forces me to use an or. And the output is more uselessly verbose when it succeeds (which it mostly will).

fail:
```
TASK [openshift_sanitize_inventory : Ensure a valid deployment type has been given] ***
skipping: [localhost]
```

vs assert:
```
TASK [openshift_sanitize_inventory : Ensure a valid deployment type has been given] ***
ok: [localhost] => {
    "changed": false
}
```

tbielawa: I wish `assert` weren't so needlessly verbose. Both of these solutions are less than ideal to me. The skipped `fail` task can be vague or ambiguous when reading a log. You can read this as 'ensuring a valid deployment type was given' was skipped. As in, no verification happened, because it was skipped.

I wish assert didn't need to print out all that garbage.

With the fail though I think the title could be adjusted to make it read clearer

Original
```
TASK [openshift_sanitize_inventory : Ensure a valid deployment type has been given] ***
skipping: [localhost]
```
^ Reads to me as "We skipped checking this"


Revision
```
TASK [openshift_sanitize_inventory : Abort if an invalid deployment type has been given] ***
skipping: [localhost]
```
^ Reads to me as "We checked this, you're good, we're going to skip aborting"

sosiouxme: Point taken and names updated. Thanks!

sosiouxme:
aos-ci-test

roles/openshift_sanitize_inventory/tasks/main.yml
```diff
@@ -1,15 +1,21 @@
...
 - name: Standardize on latest variable names
-  no_log: True  # keep task description legible
   set_fact:
     # goal is to deprecate deployment_type in favor of openshift_deployment_type.
     # both will be accepted for now, but code should refer to the new name.
     # TODO: once this is well-documented, add deprecation notice if using old name.
     deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default | string }}"
     openshift_deployment_type: "{{ openshift_deployment_type | default(deployment_type) | default | string }}"
 
+- name: Abort when deployment type is invalid
```
tbielawa: :+1: to improved readability

roles/openshift_sanitize_inventory/tasks/main.yml
```diff
@@ -19,10 +25,11 @@
     openshift_release: "{{ openshift_release | string | regex_replace('^v', '') }}"
   when: openshift_release is defined
 
-- name: Ensure a valid deployment type has been given.
-  # this variable is required; complain early and clearly if it is invalid.
-  when: openshift_deployment_type not in known_openshift_deployment_types
+- name: Abort when openshift_release is invalid
```
tbielawa: :+1: to improved readability

tbielawa:
aos-ci-test

sosiouxme:
Let's try this again.

aos-ci-test

sosiouxme:
anyone... anyone... bueller... where's the bot?

sosiouxme:
aos-ci-test

sosiouxme:
[merge]

PR #3860, opened 2017-04-05
Copy v3 data dir when performing backup

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1433272

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
+++ b/playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
@@ -4,6 +4,7 @@
   vars:
     embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
     etcdctl_command: "{{ 'etcdctl' if not openshift.common.is_containerized or embedded_etcd else 'docker exec etcd_container etcdctl' if not openshift.common.is_etcd_system_container else 'runc exec etcd etcdctl' }}"
+    timestamp: "{{ lookup('pipe', 'date +%Y%m%d%H%M%S') }}"
   roles:
   - openshift_facts
   tasks:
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
```diff
@@ -13,6 +12,8 @@
       role: etcd
       local_facts: {}
     when: "'etcd' not in openshift"
+  - set_fact:
+      timestamp: "{{ lookup('pipe', 'date +%Y%m%d%H%M%S') }}"
```
sdodson: I did this because without this it was actually re-evaluating the lookup each time `{{ timestamp }}` was used below which meant that the backup location it told you existed usually didn't unless the set of tasks happened within the same millisecond, on my test vm at least that never happened but maybe that's only because i've added additional backup steps.

playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
```diff
@@ -77,6 +78,19 @@
...
       --backup-dir={{ openshift.common.data_dir }}/etcd-backup-{{ backup_tag | default('') }}{{ timestamp }}
 
+  # According to the docs change you can simply copy snap/db
+  # https://github.com/openshift/openshift-docs/commit/b38042de02d9780842dce95cfa0ef45d53b58bc6
+  - name: Check for v3 data store
+    stat:
+      path: "{{ openshift.etcd.etcd_data_dir }}/member/snap/db"
+    register: v3_db
+
+  - name: Copy etcd v3 data store
```
ingvagabund: It is mentioned in [1] as well: ``"Recovering a cluster first needs a snapshot of the keyspace from an etcd member. A snapshot may either be taken from a live member with the etcdctl snapshot save command or by copying the member/snap/db file from an etcd data directory."``

[1] https://github.com/coreos/etcd/blob/master/Documentation/op-guide/recovery.md#snapshotting-the-keyspace

ingvagabund:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
flake on https://github.com/openshift/origin/issues/13067
[merge]

PR #3867, opened 2017-04-06
verify available excluders only when enabled

ingvagabund:
The excluder verification belongs to the `openshift_excluder` role. Let's make it official.

Bug 1434679

### DIFF
--- a/roles/openshift_excluder/tasks/verify_upgrade.yml
+++ b/roles/openshift_excluder/tasks/verify_upgrade.yml
@@ -1,0 +1,15 @@
+---
+# input variables
+# - repoquery_cmd
+# - openshift_upgrade_target
+- include: init.yml
+
+- include: verify_excluder.yml
+  vars:
+    excluder: "{{ openshift.common.service_type }}-docker-excluder"
+  when: docker_excluder_on
+
+- include: verify_excluder.yml
+  vars:
+    excluder: "{{ openshift.common.service_type }}-excluder"
+  when: openshift_excluder_on
--- a/playbooks/common/openshift-cluster/upgrades/disable_excluder.yml
+++ b/playbooks/common/openshift-cluster/upgrades/disable_excluder.yml
@@ -3,14 +3,10 @@
   hosts: oo_masters_to_config:oo_nodes_to_config
   gather_facts: no
   tasks:
-  - include: pre/validate_excluder.yml
-    vars:
-      #repoquery_cmd: repoquery_cmd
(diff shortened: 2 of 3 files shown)


playbooks/common/openshift-cluster/upgrades/pre/validate_excluder.yml
```diff
@@ -27,3 +30,5 @@
     - "{{ excluder_version.stdout.split('.')[0:2] | join('.') | version_compare(openshift_upgrade_target.split('.')[0:2] | join('.'), '>', strict=True) }}"
   when:
   - not openshift.common.is_atomic | bool
+  # validate excluders only if at least one of them is enabled
+  - (enable_excluders | default(false)) or (enable_docker_excluder | default(false)) or (enable_openshift_excluder | default(false))
```
sdodson: Should we not make `(enable_excluders | default(true))` since we also default it to true in the role?

ingvagabund: It should. Thought, I don't like the same decision logic on two places.

ingvagabund:
aos-ci-test

ingvagabund:
Tested locally.

sdodson:
aos-ci-test

roles/openshift_excluder/tasks/verify_upgrade.yml
```diff
@@ -0,0 +1,15 @@
...
+
+- include: verify_excluder.yml
+  vars:
+    #repoquery_cmd: repoquery_cmd
+    #openshift_upgrade_target: openshift_upgrade_target
+    excluder: "{{ item }}"
+  with_items:
+  - "{{ openshift.common.service_type }}-docker-excluder"
+  - "{{ openshift.common.service_type }}-excluder"
+  when: docker_excluder_on or openshift_excluder_on
```
jiajliu: I think docker_excluder and openshift_excluder should be independent. So "docker_excluder_on or openshift_excluder_on" means any of excluder is set to true, then need verify both of excluders?

ingvagabund: Given repositories provided to a customer will always ship both excluders, the assumption is valid. Unless, there is a plan to ship individual rpms in different repositories some of which may be enabled and some disabled. @sdodson can you prove me wrong?

ingvagabund:
aos-ci-test

ingvagabund:
Closed in favor of https://github.com/openshift/openshift-ansible/pull/3966

PR #3876, opened 2017-04-07
Fix BZ#1422541

wkshi:


wkshi:
sorry for wrong PR, will open another one.

PR #3880, opened 2017-04-07
Removing docker-registry route in cockpit-ui as its handled in openshift_hosted.

kwoodson:


kwoodson:
@mwoodson @joelddiaz 

Here is the fix for the issue we are seeing with the passthrough route being setup.  Cockpit-ui was overriding it after the re-encrypt route was laid down.  This should fix it.

mwoodson:
I would consider this to be a ops blocker for 3.5 at this point.  To be explicit, this needs to be in 3.5 as well.

Thanks!

kwoodson:
aos-ci-test

abutcher:
aos-ci-test

kwoodson:
[merge]

PR #3881, opened 2017-04-07
Stop all services prior to upgrading, start all services after

sdodson:
If the node service was not stopped before we attempted to upgrade docker then the node service may cause docker to restart due to service dependencies. So just stop all related services as soon as the node has been drained.

### DIFF
--- a/roles/openshift_node_upgrade/tasks/main.yml
+++ b/roles/openshift_node_upgrade/tasks/main.yml
@@ -9,6 +9,28 @@
 # - openshift_release
 
 # tasks file for openshift_node_upgrade
+
+- name: Stop node and openvswitch services
+  service:
+    name: "{{ item }}"
+    state: stopped
+  with_items:
+  - "{{ openshift.common.service_type }}-node"
+  - openvswitch
+  failed_when: false
+
+- name: Stop additional containerized services
+  service:
+    name: "{{ item }}"
+    state: stopped
+  with_items:
+    - "{{ openshift.common.service_type }}-master"
+    - "{{ openshift.common.service_type }}-master-controllers"
+    - "{{ openshift.common.service_type }}-master-api"
+    - etcd_container
+  failed_when: false
+  when: openshift.common.is_containerized | bool
+
 - include: docker/upgrade.yml
   vars:
     # We will restart Docker ourselves after everything is ready:
@@ -16,7 +38,6 @@
   when:
   - l_docker_upgrade is defined
   - l_docker_upgrade | bool
-  - not openshift.common.is_containerized | bool
 
 - include: "{{ node_config_hook }}"
   when: node_config_hook is defined
@@ -37,51 +58,14 @@
 - include: containerized_node_upgrade.yml
   when: openshift.common.is_containerized | bool
 
-- name: Ensure containerized services stopped before Docker restart
-  service:
-    name: "{{ item }}"
-    state: stopped
-  with_items:
-  - etcd_container
-  - openvswitch
-  - "{{ openshift.common.service_type }}-master"
-  - "{{ openshift.common.service_type }}-master-api"
-  - "{{ openshift.common.service_type }}-master-controllers"
-  - "{{ openshift.common.service_type }}-node"
-  failed_when: false
-  when: openshift.common.is_containerized | bool
-
-- name: Stop rpm based services
-  service:
-    name: "{{ item }}"
-    state: stopped
-  with_items:
-  - "{{ openshift.common.service_type }}-node"
-  - openvswitch
-  failed_when: false
-  when: not openshift.common.is_containerized | bool
-
 - name: Upgrade openvswitch
   package:
     name: openvswitch
     state: latest
   when: not openshift.common.is_containerized | bool
 
-- name: Restart openvswitch
-  systemd:
-    name: openvswitch
-    state: started
-  when:
-  - not openshift.common.is_containerized | bool
-
-# Mandatory Docker restart, ensure all containerized services are running:
-- include: docker/restart.yml
-
-- name: Restart rpm node service
-  service:
-    name: "{{ openshift.common.service_type }}-node"
-    state: restarted
-  when: not openshift.common.is_containerized | bool
+# Restart all services
+- include: restart.yml
 
 - name: Wait for node to be ready
   oc_obj:
--- a/roles/openshift_node_upgrade/tasks/restart.yml
+++ b/roles/openshift_node_upgrade/tasks/restart.yml
@@ -12,7 +12,7 @@
   openshift_facts:
     role: docker
 
-- name: Restart containerized services
+- name: Start services
   service: name={{ item }} state=started
   with_items:
     - etcd_container
@@ -22,7 +22,6 @@
     - "{{ openshift.common.service_type }}-master-controllers"
     - "{{ openshift.common.service_type }}-node"
   failed_when: false
-  when: openshift.common.is_containerized | bool
 
 - name: Wait for master API to come back online
   wait_for:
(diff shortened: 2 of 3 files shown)


roles/openshift_node_upgrade/tasks/main.yml
```diff
@@ -9,14 +9,34 @@
...
+
+- name: Stop additional containerized services
+  service: name={{ item }} state=stopped
+  with_items:
+    - "{{ openshift.common.service_type }}-master"
+    - "{{ openshift.common.service_type }}-master-api"
+    - "{{ openshift.common.service_type }}-master-controllers"
+    - etcd_container
+  failed_when: false
+  when: openshift.common.is_containerized | bool
```
dgoodwin: Node and openvswitch will now remain running for containerized installs it looks like, is that desired? If not I think you may want to drop the when on the above condition so we always stop those.

roles/openshift_node_upgrade/tasks/restart.yml
```diff
@@ -22,7 +22,6 @@
     - "{{ openshift.common.service_type }}-master-controllers"
     - "{{ openshift.common.service_type }}-node"
   failed_when: false
-  when: openshift.common.is_containerized | bool
```
dgoodwin: Ideally we should not be restarting master services unnecessarily, for rpm based installations here they should be independent of docker afaik, so we should leave them running if at all possible.

roles/openshift_node_upgrade/tasks/main.yml
```diff
@@ -9,14 +9,34 @@
...
+
+- name: Stop rpm based services
+  service:
+    name: "{{ item }}"
+    state: stopped
+  with_items:
+  - "{{ openshift.common.service_type }}-node"
+  - openvswitch
+  failed_when: false
+  when: not openshift.common.is_containerized | bool
```
dgoodwin: Assuming this task here is the key part of the fix?

sdodson:
The failure occurred inside tasks/docker/upgrade.yml, based on the journal docker was restarted several times right before we got to stop it.

```
TASK [openshift_node_upgrade : Check Docker image count] ***********************
task path: openshift-ansible-3.5.48/roles/openshift_node_upgrade/tasks/docker/upgrade.yml:35
skipping: [stg-node-infra-5b7b9] => {
    "changed": false, 
    "skip_reason": "Conditional check failed", 
    "skipped": true
}

TASK [openshift_node_upgrade : debug] ******************************************
task path: openshift-ansible-3.5.48/roles/openshift_node_upgrade/tasks/docker/upgrade.yml:40
skipping: [stg-node-infra-5b7b9] => {
    "changed": false, 
    "skip_reason": "Conditional check failed", 
    "skipped": true
}

TASK [openshift_node_upgrade : service] ****************************************
task path: openshift-ansible-3.5.48/roles/openshift_node_upgrade/tasks/docker/upgrade.yml:43
Using module file /usr/lib/python2.7/site-packages/ansible/modules/core/system/systemd.py
<54.152.213.232> ESTABLISH SSH CONNECTION FOR USER: root
<54.152.213.232> SSH: EXEC ssh -o ControlMaster=auto -o ControlPersist=600s -o StrictHostKeyChecking=no -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=root -o ConnectTimeout=10 -o ControlPath=/.ansible/cp/ansible-ssh-%h-%p-%r 54.152.213.232 '/bin/sh -c '"'"'/usr/bin/python && sleep 0'"'"''
fatal: [stg-node-infra-5b7b9]: FAILED! => {
    "changed": false, 
    "failed": true, 
    "invocation": {
        "module_args": {
            "daemon_reload": false, 
            "enabled": null, 
            "masked": null, 
            "name": "docker", 
            "state": "stopped", 
            "user": false
        }
    }, 
    "msg": "Unable to stop service docker: Job for docker.service canceled.\n"
}
```

roles/openshift_node_upgrade/tasks/restart.yml
sdodson: Cleaning up the task name to say `Start services` as we're not actually restarting but only starting and tolerating services that don't exist or are masked because of HA installs.

sdodson: openvswitch and openshift-node are stopped first thing in the role, then containerized master and etcd services are stopped only if containerized.

roles/openshift_node_upgrade/tasks/main.yml
sdodson: updated to remove the conditional so it now stops node and openvswitch no matter what, then conditionally stops additional services for containerized

sdodson:
aos-ci-test

sdodson:
@ingvagabund can you take a look when you get a chance?

roles/openshift_node_upgrade/tasks/restart.yml
ingvagabund: In the non-containerized deployment master services are never stopped. However, given the `systemctl start` over running services has no effect, it is completely fine to remove the when condition.

roles/openshift_node_upgrade/tasks/main.yml
```diff
@@ -9,14 +9,35 @@
...
+    - "{{ openshift.common.service_type }}-master-api"
+    - "{{ openshift.common.service_type }}-master-controllers"
+    - etcd_container
+  failed_when: false
+  when: openshift.common.is_containerized | bool
+
 - include: docker/upgrade.yml
   vars:
     # We will restart Docker ourselves after everything is ready:
     skip_docker_restart: True
```
ingvagabund: `skip_docker_restart` is not used anywhere in the ``docker/upgrade.yml`` so it can be removed

ingvagabund:
openvswitch.docker.service
```
After=docker.service
```

openshift.docker.node.service:
```
After={{ openshift.common.service_type }}-master.service
After=docker.service
After=openvswitch.service
```

master.docker.service:
```
After=docker.service
After=etcd_container.service
```

atomic-openshift-master-api.service
```
After=etcd_container.service
Before={{ openshift.common.service_type }}-node.service
After=docker.service
```

atomic-openshift-master-controllers
```
After={{ openshift.common.service_type }}-master-api.service
After=docker.service
```

etcd.docker:
```
After=docker.service
```

Deps:
```
docker -> openvswitch
master -> node, docker -> node, openvswitch -> node
docker -> master, etcd -> master,
etcd -> master-api, master-api -> node
master-api -> master-ctrl, docker -> master-ctrl
docker -> etcd
```

Given the Hasse diagram the following order of disabling of service must be preserved:
1. node, master-ctrl (in any order)
2. openvswitch, master, master-api (in any order)
3. etcd
4. docker

The following order is valid as well:
1. node
2. openvswitch
3. master
4. master-controller
5. master-api
6. etcd
7. docker

roles/openshift_node_upgrade/tasks/main.yml
```diff
@@ -9,14 +9,35 @@
...
+  failed_when: false
+
+- name: Stop additional containerized services
+  service:
+    name: "{{ item }}"
+    state: stopped
+  with_items:
+    - "{{ openshift.common.service_type }}-master"
+    - "{{ openshift.common.service_type }}-master-api"
+    - "{{ openshift.common.service_type }}-master-controllers"
```
ingvagabund: Given `master-controller` requires `master-api` [1], the `master-controller` should be stopped before the `master-api` (assuming all items are processed in the written order).

[1] https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_master/templates/docker-cluster/atomic-openshift-master-controllers.service.j2#L5

sdodson: Agreed, updated.

PR #3882, opened 2017-04-07
Fix private CA

Fale:
When private (aka not publicly trusted) CAs are used, and a registry from outside the cluster itself is used, it is not possible to retrieve images from it (without this fix).

This is only a problem of the containerized version, since in the RPM version, these files are loaded by design.

I think that also this adds no possible threats, since we are adding to the master-api containers CAs that are already trusted by the nodes, which have a higher security criticality.


tbielawa:
```
-v /etc/pki/ca-trust:/etc/pki/ca-trust
```

I suppose that would work. It should always exist.

Now, it is a directory and it's going to mount and obscure anything that's in the container at that point already.

Additionally, we run the command `update-ca-trust` in

```
$ grep -r update-ca-trust *
playbooks/common/openshift-cluster/redeploy-certificates/ca.yml:    command: update-ca-trust
roles/openshift_node_certificates/handlers/main.yml:  command: update-ca-trust
```

So **everything** your host has trusted could be added to the trust chain in the container. And I think there's a chance that if anything is added into the container it would reflect any changes in `/etc/pki/ca-trust/` back to the host it is mounted from.

I'm no security expert here so I won't drop a review on this PR. Those ^ are just the things that stick out to me.

Fale:
@tbielawa if I understood correctly your message, you are raising two points:

1. Mounting a folder means obscuring anything that would be there in the Pods. Is this safe?
2. Such mount means that all CAs trusted by the nodes will be trusted by the Openshift Master Pods. Is this safe?

So, if that is right:
On 1, I think this is not a problem, since we should not have anything in the Openshift Master Pod CA folder that is not in the node.
On 2, I think that is not a problem, since the node is far more critical than the Openshift Master Pod (aka: if the node is compromised, the Pod is compromised as well, while the opposite is not true).

Also, I'd like to point out that this patch makes the Containerized version to behave _exactly like_ the RPM version.

tbielawa:
[test]

tbielawa:
@Fale Thanks for clearing that up for me. I triggered the test system. I'm going to tag in @ashcrow and ask him to give the :+1:/:-1:

ashcrow:
@tbielawa @Fale I think the question comes down to who owns the certificate trust for applications running in containers. The use case of an organization having their own CA seems likely to me.

I don't see a problem with the mount in this PR, but it may make sense to have the mount as part ``EnvironmentFile`` mounts. I'll leave that up to you guys.

The only issue I can think of would be over selinux context, but I think ``system_u:object_r:cert_t:s0`` should be readable.

Fale:
@ashcrow yeah, I agree on deciding where the ownership lays, and I think that in all cases (at least the ones I were able to think about) this solution applies in a rational way.

As for the SELinux part, it actually works properly from that point of view. We hot-patched this change on a OCP 3.4 containerized cluster and works as expected with SELinux in enforcing mode

abutcher:
aos-ci-test

abutcher:
[test]

ashcrow:
```
fatal: [10.8.168.130]: FAILED! => {"changed": false, "failed": true, "msg": "No package matching 'atomic-openshift-docker-excluder-3.6.35*' found available, installed or updated", "rc": 126, "results": ["No package matching 'atomic-openshift-docker-excluder-3.6.35*' found available, installed or updated"]}
```

Fale:
@ashcrow I have not done anything related to that. Do you have any idea why that got flagged?

ashcrow:
@fale not really, no. @ingvagabund any ideas on this?

ingvagabund:
Since there is no ``atomic-openshift-docker-excluder-3.6.35`` build of the excluders?

Fale:
Yeah, that's what the error suggests it's the problem. What I don't understand is why it is failing since I have not changed anything related to it. I guess the only option is that the master at the commit where I forked already had the problem.

ingvagabund:
Yeah, that is not related to your PR.

Fale:
I have just merged master, let's see if this fixes the problem

abutcher:
[test]

abutcher:
aos-ci-test

Fale:
Still same problem :(

    fatal: [10.8.170.12]: FAILED! => {"changed": false, "failed": true, "msg": "No package matching 'atomic-openshift-docker-excluder-3.5.5.7*' found available, installed or updated", "rc": 126, "results": ["No package matching 'atomic-openshift-docker-excluder-3.5.5.7*' found available, installed or updated"]}

Fale:
I've just rebased it, hopefully now the package thing got solved

ingvagabund:
aos-ci-test

ingvagabund:
Just hit that issue with the latest master, investigating

ingvagabund:
@ashcrow can you give your blessing for the PR?

giuseppe:
@ashcrow thanks for tagging me, we will need to do the same change in the system container once this is merged.

Anyway, do we need write access to `/etc/pki/ca-trust` in the container?

Could it be `-v /etc/pki/ca-trust:/etc/pki/ca-trust:ro` ?

This should also help with the point @tbielawa brought up.

Fale:
Thanks guys for the feedback!

I have rebased and included @giuseppe feedback :)

giuseppe:
thanks for new version.  Having looked at it again more in details I am confused now: we already had a `-v/ /etc/pki:etc/pki:ro`.  Why doesn't it work without this patch?

Fale:
Great catch!

I think that it got added by a commit in the meantime. If we already have this, we can close it!

giuseppe:
@Fale heh, my poor memory, it seems I have even reviewed the change:

https://github.com/openshift/openshift-ansible/pull/4414

PR #3884, opened 2017-04-07
Disable swap space on nodes at install and upgrade

mtnbikenc:
https://trello.com/c/vGmZYJ79/296-3-disable-swap-at-install-and-upgrade

### DIFF
--- a/roles/openshift_node/tasks/main.yml
+++ b/roles/openshift_node/tasks/main.yml
@@ -34,6 +34,33 @@
         dns_ip: "{{ openshift_dns_ip | default(none) | get_dns_ip(hostvars[inventory_hostname])}}"
         env_vars: "{{ openshift_node_env_vars | default(None) }}"
 
+# https://docs.openshift.com/container-platform/3.4/admin_guide/overcommit.html#disabling-swap-memory
+- name: Check for swap usage
+  command: grep "^[^#].*swap" /etc/fstab
+  # grep: match any lines which don't begin with '#' and contain 'swap'
+  # command: swapon --summary
+  # Alternate option, however if swap entries are in fstab, swap will be
+  # enabled at boot. Grepping fstab should catch a condition when swap was
+  # disabled, but the fstab entries were not removed.
+  changed_when: false
+  failed_when: false
+  register: swap_result
+
+  # Disable Swap Block
+- block:
+
+    - name: Disable swap
+      command: swapoff --all
+
+    - name: Remove swap entries from /etc/fstab
+      lineinfile:
+        dest: /etc/fstab
+        regexp: 'swap'
+        state: absent
+
+  when: swap_result.stdout_lines | length > 0
+  # End Disable Swap Block
+
 # We have to add tuned-profiles in the same transaction otherwise we run into depsolving
 # problems because the rpms don't pin the version properly. This was fixed in 3.1 packaging.
 - name: Install Node package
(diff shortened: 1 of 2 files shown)


mtnbikenc:
aos-ci-test

sdodson:
How much of a challenge do you think it'd be to add a task that checks to ensure that we have enough free memory versus swap in use before doing this and perform that as a pre-upgrade check? I believe the formula we'd use is to grab MemFree, SwapTotal, SwapFree from /proc/meminfo and abort if `(MemFree < (SwapTotal - SwapFree) )`

The output of `free` may be easier too.

sdodson:
That said, if we disable swap while the node is evacuated the memory usage will be vastly different than what it is when prior to performing the upgrade and it's difficult to determine how evacuation will affect usage.

roles/openshift_node/tasks/main.yml
```diff
@@ -34,6 +34,32 @@
         dns_ip: "{{ openshift_dns_ip | default(none) | get_dns_ip(hostvars[inventory_hostname])}}"
         env_vars: "{{ openshift_node_env_vars | default(None) }}"
 
+# https://docs.openshift.com/container-platform/3.4/admin_guide/overcommit.html#disabling-swap-memory
+- name: Check for swap usage
+  command: grep swap /etc/fstab
```
sdodson: `grep -v "^#" /etc/fstab | grep swap` or something better to avoid false positives on commented lines?

mtnbikenc:
Updated grep command to ignore commented lines.

mtnbikenc:
aos-ci-test

sdodson:
aos-ci-test

mtnbikenc:
aos-ci-test

mtnbikenc:
[merge]

mtnbikenc:
aos-ci-test

mtnbikenc:
Flake https://github.com/openshift/origin/issues/13271

mtnbikenc:
[merge]

mtnbikenc:
Flake https://github.com/openshift/origin/issues/12629 Downward API volume
Flake https://github.com/openshift/origin/issues/13705 Dumping container logs
re[merge]

rhcarvalho:
> How much of a challenge do you think it'd be to add a task that checks to ensure that we have enough free memory versus swap in use before doing this and perform that as a pre-upgrade check? I believe the formula we'd use is to grab MemFree, SwapTotal, SwapFree from /proc/meminfo and abort if (MemFree < (SwapTotal - SwapFree) )
> 
> The output of free may be easier too.

@mtnbikenc @sdodson the builtin Ansible "setup" module provides memory-related facts such that we may not need to call `free` ourselves. We've been using it in https://github.com/openshift/openshift-ansible/pull/3518.
In particular, we look at `ansible_memtotal_mb`:
https://github.com/openshift/openshift-ansible/pull/3518/files#diff-7cb3c3806a4642e46d907eeed0c8873eR28

sdodson:
So, based on that if we choose to add it, if ( ansible_memfree_mb - ansible_memory_mb.swap.used > 0 ) we're good to go? Anyway, given we're performing this either at install time or during upgrade when the node is already drained I'm relatively confident that we won't run into problems.

```
        "ansible_memory_mb": {
            "nocache": {
                "free": 3507, 
                "used": 284
            }, 
            "real": {
                "free": 2012, 
                "total": 3791, 
                "used": 1779
            }, 
            "swap": {
                "cached": 0, 
                "free": 0, 
                "total": 0, 
                "used": 0
            }
        }, 
        "ansible_memtotal_mb": 3791, 
```

PR #3893, opened 2017-04-10
Updating kibana-proxy secret key for server-tls entry

ewolinetz:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1440720


sdodson:
Do we need this on master too?

ewolinetz:
@sdodson yeah, i'll get that picked after i try to address https://github.com/openshift/openshift-ansible/issues/3896

PR #3898, opened 2017-04-10
Store backups in /var/lib/etcd/openshift-backup-$timestamp

sdodson:
Because containerized installs don't mount /var/lib/origin and we
switched to running the backup inside the container that meant that we
were backing up the etcd data into a directory inside the container
filesystem. Since we have no other volume mounted we need to backup into
/var/lib/etcd.

We want to perform the backup inside the container because we cannot be sure that etcd on the host matches the version in the container.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
+++ b/playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
@@ -4,6 +4,7 @@
   vars:
     embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
     etcdctl_command: "{{ 'etcdctl' if not openshift.common.is_containerized or embedded_etcd else 'docker exec etcd_container etcdctl' if not openshift.common.is_etcd_system_container else 'runc exec etcd etcdctl' }}"
+    timestamp: "{{ lookup('pipe', 'date +%Y%m%d%H%M%S') }}"
   roles:
   - openshift_facts
   tasks:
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
```diff
@@ -71,12 +62,12 @@
     package:
       name: etcd
       state: latest
-    when: ( embedded_etcd | bool or openshift.common.is_containerized ) and not openshift.common.is_atomic
+    when: embedded_etcd | bool
```
sdodson: We no longer rely on host etcd binaries for containerized installs.

playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
```diff
@@ -53,30 +43,22 @@
...
+  - name: Install latest etcd for embedded
     package:
       name: etcd
       state: latest
-    when: ( embedded_etcd | bool or openshift.common.is_containerized ) and not openshift.common.is_atomic
+    when:
+    - embedded_etcd | bool
+    - not openshift.common.is_atomic | bool
 
   - name: Generate etcd backup
```
ashcrow: nit: Not related to your PR but this really should be ``Execute etcd backup command``.

playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
```diff
@@ -53,30 +43,22 @@
...
     command: >
       {{ etcdctl_command }} backup --data-dir={{ openshift.etcd.etcd_data_dir }}
-      --backup-dir={{ openshift.common.data_dir }}/etcd-backup-{{ backup_tag | default('') }}{{ timestamp }}
+      --backup-dir={{ etcd_backup_dir }}
```
ashcrow: 👍

ashcrow:
aos-ci-test

ashcrow:
@sdodson Bah. The 'out-of-date' warning just popped up.

PR #3899, opened 2017-04-10
Remove unnecessary folder refs

mtnbikenc:


mtnbikenc:
aos-ci-test

mtnbikenc:
@kwoodson It is likely that this playbook was moved from another location or that there was copy/paste for these tasks and the paths were just an artifact from the original.

mtnbikenc:
[merge]

Issue #3896, opened 2017-04-10
Openshift logging facts not collecting PVs and not reusing


jupierce:
#### Description

After an upgrade to our dev-preview-stg cluster (3.4->3.5), logging did not correctly initialize and it did not reuse the existing PVs. In addition, though logging was configured for a scale of 2, 3 PVCs were created. 

https://atomic-e2e-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/aos-cd/job/cluster%252Foperation/72/

##### Version

OpenShift v3.5.5.3
OpenShift Ansible: v3.5.52

##### Expected Results
PVs should have been reused.

jupierce:
@ewolinetz

ewolinetz:
Looking through the module that collects the list of PVCs that contain the label `logging-infra=support`.

While that would work for PVCs created by the deployer or this role, it would not address the extra PVCs being created nor cover scenarios where a PVC was created externally and then the ES dc was patched afterwards to use it

PR #3909, opened 2017-04-11
Updating to only create PVC when necessary and attach to the DS, othe…

ewolinetz:
…rwise leave it be

This should resolve the issues seen in #3896

CC @mwoodson @jupierce


sdodson:
master branch too please

sdodson:
PR makes sense to me, there's a few yaml lint errors we need to fix though

PR #3916, opened 2017-04-12
Deleting fluentd pods after DS patching to ensure pods are updated

ewolinetz:
rollingUpdate isn't supported pre kube 1.6 so we need to delete the fluentd pods so that they use the most recent daemonset spec.

The fix for this on master can take advantage of the fact that `oc_obj` with a `state: absent` doesn't require a name.

cc @dgoodwin

### DIFF
--- a/roles/openshift_logging/tasks/install_logging.yaml
+++ b/roles/openshift_logging/tasks/install_logging.yaml
@@ -48,6 +48,25 @@
     loop_var: file
   when: not ansible_check_mode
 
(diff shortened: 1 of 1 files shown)


ewolinetz:
aos-ci-test

dgoodwin:
Should we skip these steps if we're >= 3.6? It might get a little wonky if this starts running while k8s is simultaneously trying to rolling restart.

Otherwise looks great and ty for the quick fix!

sdodson:
Looks like a release-1.5 branch only fix then based on previous comment.

PR #3917, opened 2017-04-12
tox tests: pin test requirement versions (release-1.5)

sosiouxme:
Tests started failing once ansible 2.3 was released. It seems wise to
pin the dependencies to particular versions until we make a conscious
decision to change them (and have tested with them).


sosiouxme:
aos-ci-test

sdodson:
travis is happy, those other failures are transient

PR #3919, opened 2017-04-13
Fix instantiation of action plugin test to support Ansible 2.3

rhcarvalho:
In Ansible 2.3+, the base action plugin class' run method accesses attributes (check_mode) of its play_context.

In older versions play_context was not involved in run, and thus None was passed in.

~~**Note**: the second commit, reverting the pinning of test requirements, is meant to demonstrate how we can run with Ansible 2.3 just fine, and may be dropped later.~~

### DIFF
--- a/roles/openshift_health_checker/test/action_plugin_test.py
+++ b/roles/openshift_health_checker/test/action_plugin_test.py
@@ -1,5 +1,7 @@
 import pytest
 
+from ansible.playbook.play_context import PlayContext
+
 from openshift_health_check import ActionModule, resolve_checks
 from openshift_checks import OpenShiftCheckException
 
@@ -34,7 +36,7 @@ def fake_check(name='fake_check', tags=None, is_active=True, run_return=None, ru
 @pytest.fixture
 def plugin():
     task = FakeTask('openshift_health_check', {'checks': ['fake_check']})
-    plugin = ActionModule(task, None, None, None, None, None)
+    plugin = ActionModule(task, None, PlayContext(), None, None, None)
     return plugin
 
 


requirements.txt
```diff
@@ -1,6 +1,6 @@
-ansible==2.2.2.0
```
rhcarvalho: As noted in the PR description, unpinning is just to see green tests with latest Ansible (2.3), to demonstrate that the original issue is being addressed.

I'm not opposed to leaving versions pinned. One observation, though, is that it would be reasonable to verify that the versions we pin are available on the systems we support (RHEL, CentOS, ...), because even though we install these using `pip` in tests, we, in many contexts, expect users to install the dependencies via `rpm`.

sosiouxme: Agreed; actually I think it would be wise to pin the versions that we ship in the official openshift-ansible image (and change them in lockstep as much as possible). Eventually maybe customers will just use that and all this Ansible version madness will be over.

sdodson:
> Agreed; actually I think it would be wise to pin the versions that we ship in the official openshift-ansible image (and change them in lockstep as much as possible). Eventually maybe customers will just use that and all this Ansible version madness will be over.

Yeah, +1 to making this PR bump the rpm requirement too.

Should we leave this WIP until each team has done some more testing on their own and if necessary added commits to this PR for any changes they've found necessary to move to 2.3

rhcarvalho:
As discussed on IRC, I dropped the second commit so that we stay pinned on Ansible 2.2.2.0. Next we'll make sure other parts work well with 2.3, then pin on 2.3 for tests and on the RPM spec.

rhcarvalho:
[test]

mtnbikenc:
aos-ci-test

sosiouxme:
aos-ci-test

rhcarvalho:
I see failures related to missing excluder packages https://aos-ci.s3.amazonaws.com/openshift/openshift-ansible/jenkins-openshift-ansible-test-matrix-CONTAINERIZED=_containerized,OSE_VER=3.5,PYTHON=System-CPython-2.7,TOPOLOGY=openshift-cluster-containerized,TargetBranch=master,nodes=openshift-ansible-slave-32/590d6c12528654bf4e391a7f47662b038b5601a8.txt

rhcarvalho:
[merge]

rhcarvalho:
Seems to be another case of this flake https://github.com/openshift/origin/issues/13731

[merge]

rhcarvalho:
Flake https://github.com/openshift/origin/issues/13814

[merge]

rhcarvalho:
https://github.com/openshift/origin/issues/13833, apparently bug in the CI code.

Given

- Code reviewed and approved
- At least 2 annotated flakes trying to merge
- No code changes since then
- aos-ci-test and Travis passed
- Only health checks role involved

Will use the temporary exception to merge this manually.

Issue #3921, opened 2017-04-13
openshift_metrics fails when htpasswd is not found.


mwoodson:
### <HTPASSWD_AUTH>

We are aware of the current issues related to htpasswd_auth failures
Please downgrade to ansible 2.2.0.0 until a fix is released.
You can track the status of the bug fix in this issue:
https://github.com/openshift/openshift-ansible/issues/3111
Please erase this <HTPASSWD_AUTH> section if it does not apply to you.

Thanks - 2017-01-31

#### Description

openshift_metrics requires htpasswd and fails with a message.  I believe that openshift_metrics should install the dependency (httpd-tools) if it can't find it, not fail.

I just dug deeper, the httpd-tools needs to be installed from the host that is running the role/playbook.  I think it would be better to have htpasswd called from the node installing and configuring openshift_metrics.

##### Version

Please put the following version information in the code block
indicated below.

* Your ansible version per `ansible --version`

If you're operating from a **git clone**:

* The output of `git describe`

If you're running from playbooks installed via RPM or
`atomic-openshift-utils`

* The output of `rpm -q atomic-openshift-utils openshift-ansible`

Place the output between the code block below:

```
openshift-ansible-3.5

```

##### Steps To Reproduce
1. run the openshift_metrics roles without htpasswd on the host you are running from.

##### Expected Results
```
Install htpasswd if it's not there
```

##### Observed Results
Describe what is actually happening.

```
it fails hard now
```

For long output or logs, consider using a [gist](https://gist.github.com/)

##### Additional Information

Provide any additional information which may help us diagnose the
issue.

* Your operating system and version, ie: RHEL 7.2, Fedora 23 (`$ cat /etc/redhat-release`)
* Your inventory file (especially any non-standard configuration parameters)
* Sample code, etc

```

```

ewolinetz:
per our IRC discussion, the issue seems to be that `openshift_metrics` is expecting `htpasswd` to be available on the control host. There does not appear to be a clear need to run `htpasswd` on the control, it should be possible to accomplish the same on the target.

sdodson:
I think for now I'd press for the workaround of just intalling http-tools on the control host, if our RPM packages were installed rather than vendoring this would be the case.

ewolinetz:
@sdodson but there isn't a need to generate the files locally as they are then immediately copied to the remote host. it seems like an extra, unnecessary step/hop

Issue #1071, opened 2015-12-16
Containerization Followup: htpasswd on atomic host when using htpasswd auth provider


sdodson:
Followup for #808 
In RHEL installs we install http-utils for users so that when they're using htpasswd auth it's easy to generate password files. It'd be cool if we were able to do this on atomic host as well perhaps using rhscl/httpd-24-rhel7

ibotty:
Isn't that image a little heavy?

sdodson:
It was the only httpd image on registry.access.redhat.com when I looked. Did you have another image in mind?

I guess as an alternative to pulling an apache httpd image we could find a python implementation of htpasswd and copy that to /usr/local/bin/.

ibotty:
> It was the only httpd image on registry.access.redhat.com when I looked. Did you have another image in mind?

I did not have one in mind. I would have rolled a minimal one or
let users use rhel-tools (or fedora-tools) and install it there.

> I guess as an alternative to pulling an apache httpd image we could find a python implementation of htpasswd and copy that to /usr/local/bin/. 

Most seem to use passlib, which is not installed on atomic hosts afaict.

cgwalters:
Doesn't need to live on the host system, FWIW I just did:

```
atomic run rhel7/rhel-tools
yum -y install /usr/bin/htpasswd
htpasswd -c /host/etc/origin/master/htpasswd
```

Note the use of `/host`.

sdodson:
hmm, a generic container that's got an entrypoint that takes commands and
arguments, uses yum to install them on the fly and then run the provided
command...

detiber:
@sdodson for one-shot tasks like the htpasswd file I agree...  For something like the oc/oadm client I think we'd want to stick with the client container.

vrutkovs:
`htpasswd` is now being run locally, fixed in #3922

PR #3922, opened 2017-04-13
make htpasswd tasks remote actions instead of local

ewolinetz:
to address #3921 

cc @mwoodson

### DIFF
--- a/roles/openshift_metrics/tasks/install_support.yaml
+++ b/roles/openshift_metrics/tasks/install_support.yaml
@@ -1,11 +1,11 @@
 ---
-- name: Check control node to see if htpasswd is installed
-  local_action: command which htpasswd
+- name: Check target node to see if htpasswd is installed
+  command: which htpasswd
   register: htpasswd_check
   failed_when: no
   changed_when: no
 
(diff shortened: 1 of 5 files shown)


sdodson:
Currently we install httpd-tools via the rpm dependencies for openshift-ansible[1]. We could change this[2] so that it's not conditional based on the IDP in use but we still wouldn't have htpasswd on atomic host. How should this be handled on atomic host?

1 - https://github.com/openshift/openshift-ansible/blob/master/openshift-ansible.spec#L26
2 - https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_master/tasks/main.yml#L74-L78

sdodson:
https://github.com/openshift/openshift-ansible/issues/1071 related, discusses some solutions for htpasswd on atomic host

roles/openshift_metrics/tasks/install_support.yaml
```diff
@@ -1,11 +1,11 @@
 ---
 - name: Check control node to see if htpasswd is installed
```
jcantrill: Change name

ewolinetz:
Updated logic to use docker run instead for generating a htpasswd file @sdodson

ewolinetz:
aos-ci-test

PR #3924, opened 2017-04-13
tox tests: pin test requirement versions

sosiouxme:
This makes test req versions explicit so they only change deliberately.


sosiouxme:
@sdodson this fixes tox requirements if you'd like to expedite.

PR #3926, opened 2017-04-13
Fix default image tag for enterprise

sdodson:
We're just going to hardcode this for now until we can verify that a tag
like v3.6 will always exist in the future


sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
[merge]

PR #3927, opened 2017-04-13
tox tests: pin test requirement versions (release-1.5)

sosiouxme:
This makes req versions explicit so they only change deliberately.


sosiouxme:
https://github.com/openshift/openshift-ansible/pull/3924 for release-1.5

PR #3928, opened 2017-04-13
tox tests: pin test requirement versions (release-1.4)

sosiouxme:
This makes req versions explicit so they only change deliberately.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,8 +8,7 @@ skip_missing_interpreters=True
 [testenv]
 deps =
     -rtest-requirements.txt
-    py35-flake8: flake8-bugbear
-    ansible22: ansible~=2.2
+    py35-flake8: flake8-bugbear==17.3.0
 
 commands =
     flake8: flake8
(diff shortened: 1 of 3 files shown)


sosiouxme:
https://github.com/openshift/openshift-ansible/pull/3924 for release-1.4

sosiouxme:
@sdodson unfortunately https://travis-ci.org/openshift/openshift-ansible/builds/221829943 failed; I'm not sure why yet, looks like it pulled down the right versions of everything.

tox.ini
```diff
@@ -8,8 +8,7 @@ skip_missing_interpreters=True
 [testenv]
 deps =
     -rtest-requirements.txt
-    py35-flake8: flake8-bugbear
-    ansible22: ansible~=2.2
```
sdodson: @sosiouxme maybe this line was inadvertently removed?

sosiouxme: Seems likely that's the problem. I thought it was superfluous with ansible specified in the requirements.txt. Apparently not. Will fix.

sdodson: Hmm, it's not in other branches. I'm not sure what's up.

PR #3929, opened 2017-04-13
tox tests: put back ansible22 that was apparently necessary

sosiouxme:


sosiouxme:
https://travis-ci.org/openshift/openshift-ansible/builds/221857197 failed because there's a whole parallel tox in `utils/` that needs to have the same thing done.

sosiouxme:
@sdodson this should finally fix the 1.4 jobs.

PR #3930, opened 2017-04-13
Start master services in parallel

abutcher:
Removes staggered master start and two 15 second pauses. Need to find link to the controllers race condition this was originally working around but I believe it has been fixed.


sdodson:
aos-ci-test

abutcher:
[test]

PR #3932, opened 2017-04-14
Logging: Add support for mux - secure_forward service

richm:
See this PR for details:
https://github.com/openshift/origin-aggregated-logging/pull/378
@jcantrill @ewolinetz @nhosoi @lukas-vlcek @t0ffel @portante

### DIFF
--- a/roles/openshift_logging/tasks/generate_certs.yaml
+++ b/roles/openshift_logging/tasks/generate_certs.yaml
@@ -45,6 +45,21 @@
     - procure_component: kibana-internal
       hostnames: "kibana, kibana-ops, {{openshift_logging_kibana_hostname}}, {{openshift_logging_kibana_ops_hostname}}"
 
+- include: procure_server_certs.yaml
+  loop_control:
+    loop_var: cert_info
+  with_items:
+    - procure_component: mux
+      hostnames: "logging-mux, {{openshift_logging_mux_hostname}}"
+  when: openshift_logging_use_mux
+
+- include: procure_shared_key.yaml
+  loop_control:
+    loop_var: shared_key_info
+  with_items:
+    - procure_component: mux
+  when: openshift_logging_use_mux
+
 - name: Copy proxy TLS configuration file
   copy: src=server-tls.json dest={{generated_certs_dir}}/server-tls.json
   when: server_tls_json is undefined
@@ -84,6 +99,14 @@
     - system.admin
   loop_control:
     loop_var: node_name
+
+- name: Generate PEM cert for mux
+  include: generate_pems.yaml component={{node_name}}
+  with_items:
+    - system.logging.mux
+  loop_control:
+    loop_var: node_name
+  when: openshift_logging_use_mux
 
 - name: Creating necessary JKS certs
   include: generate_jks.yaml
--- a/roles/openshift_logging/tasks/install_mux.yaml
+++ b/roles/openshift_logging/tasks/install_mux.yaml
@@ -1,0 +1,67 @@
+---
+- set_fact: mux_ops_host={{ (openshift_logging_use_ops | bool) | ternary(openshift_logging_es_ops_host, openshift_logging_es_host) }}
+  check_mode: no
+
+- set_fact: mux_ops_port={{ (openshift_logging_use_ops | bool) | ternary(openshift_logging_es_ops_port, openshift_logging_es_port) }}
+  check_mode: no
+
+- name: Check mux current replica count
+  command: >
+    {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig get dc/logging-mux
+    -o jsonpath='{.spec.replicas}' -n {{openshift_logging_namespace}}
+  register: mux_replica_count
+  when: not ansible_check_mode
+  ignore_errors: yes
+  changed_when: no
+
+- name: Generating mux deploymentconfig
+  template: src=mux.j2 dest={{mktemp.stdout}}/templates/logging-mux-dc.yaml
+  vars:
+    component: mux
+    logging_component: mux
+    deploy_name: "logging-{{component}}"
+    image: "{{openshift_logging_image_prefix}}logging-fluentd:{{openshift_logging_image_version}}"
+    es_host: logging-es
+    es_port: "{{openshift_logging_es_port}}"
+    ops_host: "{{ mux_ops_host }}"
+    ops_port: "{{ mux_ops_port }}"
+    mux_cpu_limit: "{{openshift_logging_mux_cpu_limit}}"
+    mux_memory_limit: "{{openshift_logging_mux_memory_limit}}"
+    replicas: "{{mux_replica_count.stdout | default (0)}}"
+    mux_node_selector: "{{openshift_logging_mux_nodeselector | default({})}}"
+  check_mode: no
+  changed_when: no
+
+- name: "Check mux hostmount-anyuid permissions"
+  command: >
+    {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig
+    get scc/hostmount-anyuid -o jsonpath='{.users}'
+  register: mux_hostmount_anyuid
+  check_mode: no
+  changed_when: no
+
+- name: "Set hostmount-anyuid permissions for mux"
+  command: >
+    {{ openshift.common.admin_binary}} --config={{ mktemp.stdout }}/admin.kubeconfig policy
+    add-scc-to-user hostmount-anyuid system:serviceaccount:{{openshift_logging_namespace}}:aggregated-logging-fluentd
+  register: mux_output
+  failed_when: "mux_output.rc == 1 and 'exists' not in mux_output.stderr"
+  check_mode: no
+  when: mux_hostmount_anyuid.stdout.find("system:serviceaccount:{{openshift_logging_namespace}}:aggregated-logging-fluentd") == -1
+
+- name: "Check mux cluster-reader permissions"
+  command: >
+    {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig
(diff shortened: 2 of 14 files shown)


roles/openshift_logging/templates/mux.j2
```diff
@@ -0,0 +1,157 @@
...
+          value: "{{ops_host}}"
+        - name: "OPS_PORT"
+          value: "{{ops_port}}"
+        - name: "OPS_CLIENT_CERT"
+          value: "{{openshift_logging_es_ops_client_cert}}"
+        - name: "OPS_CLIENT_KEY"
+          value: "{{openshift_logging_es_ops_client_key}}"
+        - name: "OPS_CA"
+          value: "{{openshift_logging_es_ops_ca}}"
+        - name: "ES_COPY"
```
ewolinetz: Should we remove the ES copy vars?

ewolinetz:
@richm for my knowledge with the different component roles, this would best be contained within the `openshift_logging_fluentd` role, correct?

roles/openshift_logging/templates/mux.j2
richm: @ewolinetz Are we going to support ES COPY at all in 3.6?

richm:
> @richm for my knowledge with the different component roles, this would best be contained within the openshift_logging_fluentd role, correct?

Not sure.  I think we will need a new role for openshift_logging_mux.  But there are several files shared with fluentd, so I'm not sure how that will work.  Can we have a "openshift_logging_common" or "openshift_logging_shared" or something like that?  Or can openshift_logging_mux use files/templates from openshift_logging_fluentd?

roles/openshift_logging/templates/fluentd.j2
```diff
@@ -59,6 +59,11 @@ spec:
         - name: dockercfg
           mountPath: /etc/sysconfig/docker
           readOnly: true
+{% if openshift_logging_use_mux_client %}
```
t0ffel: If you're adding the keys for the mux, here, where do you use them in fluentd collector?

roles/openshift_logging/tasks/install_mux.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+- name: "Check mux privileged permissions"
+  command: >
+    {{ openshift.common.client_binary }} --config={{ mktemp.stdout }}/admin.kubeconfig
+    get scc/privileged -o jsonpath='{.users}'
+  register: mux_privileged
+  check_mode: no
+  changed_when: no
+
+- name: "Set privileged permissions for mux"
```
t0ffel: why does it need to be privileged?

roles/openshift_logging/tasks/install_mux.yaml
```diff
@@ -0,0 +1,67 @@
...
+  command: >
+    {{ openshift.common.admin_binary}} --config={{ mktemp.stdout }}/admin.kubeconfig policy
+    add-scc-to-user privileged system:serviceaccount:{{openshift_logging_namespace}}:aggregated-logging-fluentd
+  register: mux_output
+  failed_when: "mux_output.rc == 1 and 'exists' not in mux_output.stderr"
+  check_mode: no
+  when: mux_privileged.stdout.find("system:serviceaccount:{{openshift_logging_namespace}}:aggregated-logging-fluentd") == -1
+
+- name: "Check mux cluster-reader permissions"
```
t0ffel: cluster-reader shouldn't be needed for mux imho

roles/openshift_logging/templates/mux.j2
```diff
@@ -0,0 +1,157 @@
...
+          value: "{{ openshift_logging_mux_port }}"
+        - name: USE_MUX
+          value: "true"
+        - name: USE_MUX_SERVICE
+          value: "{{ openshift_logging_use_mux_service| default('false') }}"
+      volumes:
+      - name: config
+        configMap:
+          name: logging-mux
+      - name: certs
```
t0ffel: not needed in mux?

roles/openshift_logging/templates/mux.j2
t0ffel: but most of these variables are not needed in mux, right?

roles/openshift_logging/templates/mux.j2
```diff
@@ -0,0 +1,157 @@
...
+        secret:
+          secretName: logging-fluentd
+      - name: dockerhostname
```
t0ffel: not needed in mux?

roles/openshift_logging/templates/mux.j2
```diff
@@ -0,0 +1,157 @@
...
+        hostPath:
+          path: /etc/hostname
+      - name: localtime
```
t0ffel: not needed in mux?

roles/openshift_logging/templates/mux.j2
ewolinetz: @richm no, we should remove it by then since we state that its deprecated (as of 3.4) and we provide `secure-forward`

ewolinetz:
@richm the changes seem to only be with fluentd for this (sans another svc and secret). I think it would make sense to contain it in the fluentd role. I'd like to avoid creating common roles because that can quickly become a tangle of dependencies.

Is it possible to deploy 'mux' without fluentd in the context of this PR?

roles/openshift_logging/templates/fluentd.j2
richm: fluentd collector needs the mux ca cert and mux shared_key for the client side of secure_forward - see https://github.com/openshift/origin-aggregated-logging/pull/378/files#diff-5c5eff90645c105043514f4377984270

roles/openshift_logging/tasks/install_mux.yaml
richm: I don't know.  mux does need to mount /etc/localtime and /etc/docker-hostname from the host.  perhaps it is required for that?

roles/openshift_logging/tasks/install_mux.yaml
richm: ok - I supposed this relates to privileged as well

roles/openshift_logging/templates/mux.j2
richm: needed in mux - mux needs these to talk to Elasticsearch

roles/openshift_logging/templates/mux.j2
richm: I don't know why docker-hostname is needed at all, but if it is needed, it is needed in mux in order to get the pipeline metadata hostname.

roles/openshift_logging/templates/mux.j2
richm: needed in mux in order to get the timezone

richm:
> Is it possible to deploy 'mux' without fluentd in the context of this PR?

@ewolinetz I'm not sure.  If it is possible to deploy logging without fluentd, then it is possible to deploy mux without fluentd.
Do we want to do that?  I suppose there could be a case where you deploy openshift + logging _only_ for the purpose of being an external logging service, _and_ you do not want to collect logs from inside the logging cluster, you will rely on looking at the log files inside the pod.  This will not be a common case though.

roles/openshift_logging/tasks/install_mux.yaml
richm: @ewolinetz is it required to have cluster-reader permission in order to talk to the api server?  That is, in order to use the k8s meta plugin from fluentd?

richm: yes, this is required in order to use the k8s meta plugin

ewolinetz:
aos-ci-test

richm:
aos-ci-test

richm:
I have no idea why these tests are failing - doesn't look related to mux or logging

ewolinetz:
@richm i guess my question was more of it is a separate component all together, or another type of fluentd deployment?

I think that after learning more about this it would make sense to be a separate component (in the scheme of splitting up components into individual roles) since the deployment mechanism differs from the current Fluentd. The only common thing between the two is the image used.

richm:
@ewolinetz 
> I think that after learning more about this it would make sense to be a separate component (in the scheme of splitting up components into individual roles) since the deployment mechanism differs from the current Fluentd.

Do I need to do that in this PR, or is that something I should help with when you split things into individual roles?

> The only common thing between the two is the image used.

Right.

I suppose if you have a very small deployment, you could make the fluentd daemonset also a mux service for requests coming from outside the cluster, sort of a fluentd-mux all-in-one.

ewolinetz:
@richm that would be a change incorporated with the subrole PR. I would appreciate assistance with it if you're willing to help. It was dropped in priority to next sprint but I'm currently in a holding pattern with my other card so it was just regression testing. There should be a fairly well defined pattern to follow at this point

ewolinetz:
aos-ci-test

richm:
@ewolinetz so what do I need to do to get this PR merged?

richm:
aos-ci-test

ewolinetz:
aos-ci-test

roles/openshift_logging/tasks/generate_certs.yaml
```diff
@@ -45,6 +45,21 @@
     - procure_component: kibana-internal
       hostnames: "kibana, kibana-ops, {{openshift_logging_kibana_hostname}}, {{openshift_logging_kibana_ops_hostname}}"
 
+- include: procure_server_certs.yaml
+  loop_control:
+    loop_var: cert_info
+  with_items:
+    - procure_component: mux
+      hostnames: "logging-mux, {{openshift_logging_mux_hostname}}"
+  when: openshift_logging_use_mux
```
ewolinetz: i should have asked this earlier... but as a follow up can we add `| bool` to these `when` statements?

ewolinetz:
[merge]

PR #3937, opened 2017-04-17
Correct openshift_node_upgrade role dependencies

mtnbikenc:
`yedit` was added as a command to the the openshift_node_upgrade role, however the lib_utils was not added to the the role dependencies.

PR #3936, opened 2017-04-17
BZ 1442942: Fix paths for file includes

mtnbikenc:
`ansible-playbook --syntax-check` revealed path issues which are fixed.

Fixes 1442942

PR #3938, opened 2017-04-17
Add Ansible syntax checks to tox

mtnbikenc:
Adds syntax checking for entry point playbooks to standard tox checks run by CI.  An entry point playbook is identified as being a .yml/.yaml file containing `initialize_groups.yml`.

This PR will fail Travis CI until #3936 and #3937 are merged.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@ import os
 import fnmatch
 import re
 import sys
+import subprocess
 import yaml
 
 # Always prefer setuptools over distutils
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -199,6 +200,48 @@ def run(self):
...
+                        try:
+                            subprocess.check_output(
+                                ['ansible-playbook', '-i localhost,',
+                                 '--syntax-check', yaml_file]
+                            )
+                        except subprocess.CalledProcessError as cpe:
+                            print('Execution failed: %s' % cpe)
+                            has_errors = True
+        if has_errors:
+            print('Ansible syntax-check issues found')
```
tbielawa: I ran this test (`python setup.py ansible_syntax`) and got the error message above ^

**nit**: `Ansible syntax-check issues found` is a pretty mundane looking output message. Could we prefix that with **ERROR:** or something? I almost missed it (though, there were obvious errors printed in the messages above that lint).

I won't block on it though.

mtnbikenc: I'm fine with dropping that message altogether.  It is not really providing much value.  I'd rather have the lines with `Execution failed` printed in red.  Whenever I'm looking through Travis logs, I scroll down looking for red.  The only thing that is reported in red for the syntax check is the final `InvocationError:`, then you have to scroll back up to see all the individual errors.

mtnbikenc:
@tbielawa Made a few changes regarding the output style.  Colorized the individual error messages and dropped the less useful print statements.

tbielawa:
@mtnbikenc looks great. I was thinking red too. Nice touch. :+1:

mtnbikenc:
aos-ci-test

mtnbikenc:
[merge]

PR #3948, opened 2017-04-18
Create openshift-metrics entrypoint playbook

mtnbikenc:
Standardized entry point to byo/openshift-cluster/openshift-metrics.yml.


mtnbikenc:
aos-ci-test

sdodson:
that's one way to use fewer api calls...

PR #3949, opened 2017-04-18
openshift_version: improve messaging

sosiouxme:
This role often provides the first failure messages when users'
inventory parameters are not going to work out. So, make sure the task
names and messages are informative to someone who may not have very much
context for what this role does (or, indeed, how to read Ansible output).

### DIFF
--- a/roles/openshift_version/tasks/main.yml
+++ b/roles/openshift_version/tasks/main.yml
@@ -7,8 +7,13 @@
 # Block attempts to install origin without specifying some kind of version information.
 # This is because the latest tags for origin are usually alpha builds, which should not
 # be used by default. Users must indicate what they want.
-- fail:
-    msg: "Must specify openshift_release or openshift_image_tag in inventory to install origin. (suggestion: add openshift_release=\"1.2\" to inventory)"
+- name: Abort when we cannot safely guess what Origin image version the user wanted
+  fail:
+    msg: |-
+      To install a containerized Origin release, you must set openshift_release or
+      openshift_image_tag in your inventory to specify which version of the OpenShift
+      component images to use. You may want the latest (usually alpha) releases or
+      a more stable release. (Suggestion: add openshift_release="x.y" to inventory.)
   when:
   - is_containerized | bool
   - openshift.common.deployment_type == 'origin'
@@ -27,7 +32,10 @@
   when: openshift_release is defined
 
 # Verify that the image tag is in a valid format
-- block:
+- when:
+  - openshift_image_tag is defined
+  - openshift_image_tag != "latest"
+  block:
 
   # Verifies that when the deployment type is origin the version:
   # - starts with a v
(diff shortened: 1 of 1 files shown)


sosiouxme:
aos-ci-test

ashcrow:
[test]

ashcrow:
```
2017-04-18 15:06:29,870 - taskrunner - ERROR - Caught exception while running 'GetNodes'
```

ashcrow:
aos-ci-test

roles/openshift_version/tasks/main.yml
```diff
@@ -7,8 +7,13 @@
...
 # be used by default. Users must indicate what they want.
-- fail:
-    msg: "Must specify openshift_release or openshift_image_tag in inventory to install origin. (suggestion: add openshift_release=\"1.2\" to inventory)"
+- name: Abort when we cannot safely guess what Origin image version the user wanted
+  fail:
+    msg: |-
+      To install a containerized Origin release, you must set openshift_release or
+      openshift_image_tag in your inventory to specify which version of the OpenShift
+      component images to use. You may want the latest (usually alpha) releases or
+      a more stable release. (Suggestion: add openshift_release="x.y" to inventory.)
```
tbielawa: I :heart: suggestions. Much nicer than just telling people they dun messed up.

roles/openshift_version/tasks/main.yml
```diff
@@ -27,37 +32,40 @@
...
-  - name: Verify Origin openshift_image_tag is valid
+  - name: (Origin) Verify openshift_image_tag is valid
+    when: openshift.common.deployment_type == 'origin'
     assert:
       that:
       - "{{ openshift_image_tag|match('(^v?\\d+\\.\\d+\\.\\d+(-[\\w\\-\\.]*)?$)') }}"
-      msg: "openshift_image_tag must be in the format v#.#.#[-optional.#]. Examples: v1.2.3, v3.5.1-alpha.1"
-    when: openshift.common.deployment_type == 'origin'
+      msg: |-
+        openshift_image_tag must be in the format v#.#.#[-optional.#]. Examples: v1.2.3, v3.5.1-alpha.1
```
tbielawa: :+1:

sosiouxme:
The flakes are eating us. Let's try a [merge].

roles/openshift_version/tasks/main.yml
```diff
@@ -119,30 +127,42 @@
 
 - fail:
     msg: openshift_version role was unable to set openshift_version
```
rhcarvalho: Suggestion: expand the messages in this section to suggest what to do (set openshift_{version,image_tag,pkg_version} in the inventory)?

roles/openshift_version/tasks/main.yml
```diff
@@ -119,30 +127,42 @@
...
+  assert:
+    that:
+    - openshift_version.startswith(openshift_release) | bool
+    msg: |-
+      You requested openshift_release {{ openshift_release }}, which is not matched by
+      the latest OpenShift RPM we detected as {{ openshift.common.service_type }}-{{ openshift_version }}
+      on host {{ inventory_hostname }}.
+      We will only install the latest RPMs, so please ensure you are getting the release
+      you expect. You may need to adjust your Ansible inventory, modify the repositories
+      available on the host, or run the appropriate OpenShift upgrade playbook.
```
rhcarvalho: :+1:

roles/openshift_version/tasks/main.yml
sosiouxme: Actually these three are more of a "that shouldn't happen" kind of check so I'm not really sure what to recommend.

rhcarvalho: Are those variables set by the role, not the user?

sosiouxme: They are all supposed to be set by the role. The inventory can set `openshift_image_tag` and `openshift_pkg_version` which are described in https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.ose.example#L35 but they are only generally needed if the role can't figure it out some other way. `openshift_version` really shouldn't be set by the user at all, though nothing stops them and some inventories probably do.

You're right, though, having some advice would be useful. It would be good to know if there are any likely situations where the user might hit these so we could address them.

sosiouxme:
Looks like [two flakes for the price of one](https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/253/): https://github.com/openshift/origin/issues/13821 and https://github.com/openshift/origin/issues/11114

re[merge]

PR #3954, opened 2017-04-19
Etcd migrate

ingvagabund:
**Entry playbook:**
``playbooks/byo/openshift-etcd/migrate.yml``

**TODO**:
- [x] migration over rpm based etcd cluster
- [x] migration over containerized etcd cluster
- [x] trivial validation of migrated data
- [x] restore TTL of events via leases?
- [x] find the proper location for the attachlease binary
- [x] backup the etcd v2 data before the migration
- [ ] (optional) backup the etcd v3 data after the migration
- [ ] (optional) minimize the number of dependencies needed for the migration (it takes about 3 minutes to reach the etcd migration role).
- [x] the migration is not destructive so even if the migration fails (either due to etcdctl migrate failure, validation failure or attach lease) the masters need to be restarted (if there is any failure, the master configuration file must stay untouched).
- [x] make the validation faster (e.g. re-implement the python code in Go)
- ...

Deployments tested:
- 3-member etcd cluster + 1 master (all rpm based)
- 3-member etcd clsuter + 3 masters (HA) (all rpm based)
- 3-member containerized etcd cluster + 3 containerized masters (HA)

When starting master-api, resp. master-controllers service, they do not start at the first time. Thus, introducing the until construct:
```yaml
TASK [Start masters] ***********************************************************
changed: [10.8.172.213] => (item=atomic-openshift-master-api)
changed: [10.8.172.211] => (item=atomic-openshift-master-api)
changed: [10.8.172.216] => (item=atomic-openshift-master-api)
changed: [10.8.172.216] => (item=atomic-openshift-master-controllers)
FAILED - RETRYING: TASK: Start masters (5 retries left).
changed: [10.8.172.211] => (item=atomic-openshift-master-controllers)
changed: [10.8.172.213] => (item=atomic-openshift-master-controllers)
```
resp.
```yaml
TASK [Start master services] ***************************************************
changed: [10.8.172.216] => (item=atomic-openshift-master-api)
changed: [10.8.172.213] => (item=atomic-openshift-master-api)
FAILED - RETRYING: TASK: Start master services (5 retries left).
changed: [10.8.172.216] => (item=atomic-openshift-master-controllers)
FAILED - RETRYING: TASK: Start master services (5 retries left).
changed: [10.8.172.211] => (item=atomic-openshift-master-api)
changed: [10.8.172.211] => (item=atomic-openshift-master-controllers)
changed: [10.8.172.213] => (item=atomic-openshift-master-controllers)
```

The service fails randomly. However, it happens only from time to time.

Etcd restore instructions: https://docs.openshift.org/latest/admin_guide/backup_restore.html#cluster-restore-single-member-etcd-clusters

### DIFF
--- a/playbooks/byo/openshift-etcd/migrate.yml
+++ b/playbooks/byo/openshift-etcd/migrate.yml
@@ -1,0 +1,124 @@
+---
+- include: ../openshift-cluster/initialize_groups.yml
+  tags:
+  - always
+
+- include: ../../common/openshift-cluster/evaluate_groups.yml
+  tags:
+  - always
+
+- name: Run pre-checks
+  hosts: oo_etcd_to_config
+  tags:
+  - always
+  roles:
+  - role: etcd_migrate
+    r_etcd_migrate_action: check
+    etcd_peer: "{{ ansible_default_ipv4.address }}"
+
+# TODO(jchaloup): replace the std_include with something minimal so the entire playbook is faster
+# e.g. I don't need to detect the OCP version, install deps, etc.
+- include: ../../common/openshift-cluster/std_include.yml
+  tags:
+  - always
+
+- name: Backup v2 data
+  hosts: oo_etcd_to_config
+  gather_facts: no
+  tags:
+  - always
+  roles:
+  - role: openshift_facts
+  - role: etcd_common
+    r_etcd_common_action: backup
+    r_etcd_common_etcd_runtime: "{{ openshift.common.etcd_runtime }}"
+    r_etcd_common_backup_tag: pre-migration
+    r_etcd_common_embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
+    r_etcd_common_backup_sufix_name: "{{ lookup('pipe', 'date +%Y%m%d%H%M%S') }}"
+
+- name: Gate on etcd backup
+  hosts: localhost
+  connection: local
+  become: no
+  tasks:
+  - set_fact:
+      etcd_backup_completed: "{{ hostvars
+                                 | oo_select_keys(groups.oo_etcd_to_config)
+                                 | oo_collect('inventory_hostname', {'r_etcd_common_backup_complete': true}) }}"
+  - set_fact:
+      etcd_backup_failed: "{{ groups.oo_etcd_to_config | difference(etcd_backup_completed) }}"
+  - fail:
+      msg: "Migration cannot continue. The following hosts did not complete etcd backup: {{ etcd_backup_failed | join(',') }}"
+    when:
+    - etcd_backup_failed | length > 0
+
+- name: Prepare masters for etcd data migration
+  hosts: oo_masters_to_config
+  tasks:
+  - set_fact:
+      master_services:
+      - "{{ openshift.common.service_type + '-master' }}"
+  - set_fact:
+      master_services:
+      - "{{ openshift.common.service_type + '-master-controllers' }}"
+      - "{{ openshift.common.service_type + '-master-api' }}"
+    when:
+    - (openshift_master_cluster_method is defined and openshift_master_cluster_method == "native") or openshift.common.is_master_system_container | bool
+  - debug:
+      msg: "master service name: {{ master_services }}"
+  - name: Stop masters
+    service:
+      name: "{{ item }}"
+      state: stopped
+    with_items: "{{ master_services }}"
+
+- name: Migrate etcd data from v2 to v3
+  hosts: oo_etcd_to_config
+  gather_facts: no
+  tags:
+  - always
+  roles:
+  - role: etcd_migrate
+    r_etcd_migrate_action: migrate
+    etcd_peer: "{{ ansible_default_ipv4.address }}"
+
+- name: Gate on etcd migration
+  hosts: oo_masters_to_config
+  gather_facts: no
+  tasks:
+  - set_fact:
+      etcd_migration_completed: "{{ hostvars
+                                 | oo_select_keys(groups.oo_etcd_to_config)
+                                 | oo_collect('inventory_hostname', {'r_etcd_migrate_success': true}) }}"
+  - set_fact:
+      etcd_migration_failed: "{{ groups.oo_etcd_to_config | difference(etcd_migration_completed) }}"
+
+- name: Configure masters if etcd data migration is succesfull
+  hosts: oo_masters_to_config
+  roles:
+  - role: etcd_migrate
+    r_etcd_migrate_action: configure
+    when: etcd_migration_failed | length == 0
+  tasks:
+  - debug:
+      msg: "Skipping master re-configuration since migration failed."
+    when:
+    - etcd_migration_failed | length > 0
+
+- name: Start masters after etcd data migration
+  hosts: oo_masters_to_config
+  tasks:
+  - name: Start master services
+    service:
+      name: "{{ item }}"
+      state: started
+    register: service_status
+    # Sometimes the master-api, resp. master-controllers fails to start for the first time
+    until: service_status.state is defined and service_status.state == "started"
+    retries: 5
+    delay: 10
+    with_items: "{{ master_services[::-1] }}"
+  - fail:
+      msg: "Migration failed. The following hosts were not properly migrated: {{ etcd_migration_failed | join(',') }}"
+    when:
+    - etcd_migration_failed | length > 0
--- a/roles/etcd_migrate/README.md
+++ b/roles/etcd_migrate/README.md
@@ -1,0 +1,53 @@
+Role Name
+=========
+
+Offline etcd migration of data from v2 to v3
+
+Requirements
+------------
+
+It is expected all consumers of the etcd data are not accessing the data.
+Otherwise the migrated data can be out-of-sync with the v2 and can result in unhealthy etcd cluster.
+
+The role itself is responsible for:
+- checking etcd cluster health and raft status before the migration
+- checking of presence of any v3 data (in that case the migration is stopped)
+- migration of v2 data to v3 data (including attaching leases of keys prefixed with "/kubernetes.io/events" and "/kubernetes.io/masterleases" string)
+- validation of migrated data (all v2 keys and in v3 keys and are set to the identical value)
+
+The migration itself requires an etcd member to be down in the process. Once the migration is done, the etcd member is started.
+
+Role Variables
+--------------
+
+TBD
+
+Dependencies
+------------
+
+- etcd_common
+- lib_utils
+
+Example Playbook
+----------------
+
+```yaml
+- name: Migrate etcd data from v2 to v3
+  hosts: oo_etcd_to_config
+  gather_facts: no
+  tasks:
+  - include_role:
+      name: openshift_etcd_migrate
+    vars:
+      etcd_peer: "{{ ansible_default_ipv4.address }}"
+```
+
+License
+-------
+
+Apache License, Version 2.0
+
+Author Information
+------------------
+
+Jan Chaloupka (jchaloup@redhat.com)
--- a/roles/etcd_migrate/defaults/main.yml
+++ b/roles/etcd_migrate/defaults/main.yml
@@ -1,0 +1,3 @@
+---
+# Default action when calling this role, choices: check, migrate, configure
+r_etcd_migrate_action: migrate
--- a/roles/etcd_migrate/meta/main.yml
+++ b/roles/etcd_migrate/meta/main.yml
@@ -1,0 +1,17 @@
+---
+galaxy_info:
+  author: Jan Chaloupka
+  description: Etcd migration
+  company: Red Hat, Inc.
+  license: Apache License, Version 2.0
+  min_ansible_version: 2.1
+  platforms:
+  - name: EL
+    versions:
+    - 7
+  categories:
+  - cloud
+  - system
+dependencies:
+- { role: etcd_common }
+- { role: lib_utils }
--- a/roles/etcd_migrate/tasks/check.yml
+++ b/roles/etcd_migrate/tasks/check.yml
@@ -1,0 +1,55 @@
+---
+# Check the cluster is healthy
+- include: check_cluster_health.yml
+
+# Check if the member has v3 data already
+# Run the migration only if the data are v2
+- name: Check if there are any v3 data
+  command: >
+    etcdctl --cert {{ etcd_peer_cert_file }} --key {{ etcd_peer_key_file }} --cacert {{ etcd_peer_ca_file }} --endpoints 'https://{{ etcd_peer }}:2379' get "" --from-key --keys-only -w json --limit 1
+  environment:
+    ETCDCTL_API: 3
+  register: l_etcdctl_output
+
+- fail:
+    msg: "Unable to get a number of v3 keys"
+  when: l_etcdctl_output.rc != 0
+
+- fail:
+    msg: "The etcd has at least one v3 key"
+  when: "'count' in (l_etcdctl_output.stdout | from_json) and (l_etcdctl_output.stdout | from_json).count != 0"
+
+
+# TODO(jchaloup): once the until loop can be used over include/block,
+#                 remove the repetive code
+# - until loop not supported over include statement (nor block)
+#   https://github.com/ansible/ansible/issues/17098
+# - with_items not supported over block
+
+# Check the cluster status for the first time
+- include: check_cluster_status.yml
+
+# Check the cluster status for the second time
+- block:
+  - debug:
+      msg: "l_etcd_cluster_status_ok: {{ l_etcd_cluster_status_ok }}"
+  - name: Wait a while before another check
+    pause:
+      seconds: 5
+    when: not l_etcd_cluster_status_ok | bool
+
+  - include: check_cluster_status.yml
+    when: not l_etcd_cluster_status_ok | bool
+
+
+# Check the cluster status for the third time
+- block:
+  - debug:
+      msg: "l_etcd_cluster_status_ok: {{ l_etcd_cluster_status_ok }}"
+  - name: Wait a while before another check
+    pause:
+      seconds: 5
+    when: not l_etcd_cluster_status_ok | bool
+
+  - include: check_cluster_status.yml
+    when: not l_etcd_cluster_status_ok | bool
--- a/roles/etcd_migrate/tasks/check_cluster_health.yml
+++ b/roles/etcd_migrate/tasks/check_cluster_health.yml
@@ -1,0 +1,23 @@
+---
+- name: Check cluster health
+  command: >
+    etcdctl --cert-file /etc/etcd/peer.crt --key-file /etc/etcd/peer.key --ca-file /etc/etcd/ca.crt --endpoint https://{{ etcd_peer }}:2379 cluster-health
+  register: etcd_cluster_health
+  changed_when: false
+  failed_when: false
+
+- name: Assume a member is not healthy
+  set_fact:
+    etcd_member_healthy: false
+
+- name: Get member item health status
+  set_fact:
+    etcd_member_healthy: true
+  with_items: "{{ etcd_cluster_health.stdout_lines }}"
+  when: "(etcd_peer in item) and ('is healthy' in item)"
+
+- name: Check the etcd cluster health
+  # TODO(jchaloup): should we fail or ask user if he wants to continue? Or just wait until the cluster is healthy?
+  fail:
+    msg: "Etcd member {{ etcd_peer }} is not healthy"
+  when: not etcd_member_healthy
--- a/roles/etcd_migrate/tasks/check_cluster_status.yml
+++ b/roles/etcd_migrate/tasks/check_cluster_status.yml
@@ -1,0 +1,32 @@
+---
+# etcd_ip originates from etcd_common role
+- name: Check cluster status
+  command: >
+    etcdctl --cert /etc/etcd/peer.crt --key /etc/etcd/peer.key --cacert /etc/etcd/ca.crt --endpoints 'https://{{ etcd_peer }}:2379' -w json endpoint status
+  environment:
+    ETCDCTL_API: 3
+  register: l_etcd_cluster_status
+
+- name: Retrieve raftIndex
+  set_fact:
+    etcd_member_raft_index: "{{ (l_etcd_cluster_status.stdout | from_json)[0]['Status']['raftIndex'] }}"
+
+- block:
+  # http://docs.ansible.com/ansible/playbooks_filters.html#extracting-values-from-containers
+  - name: Group all raftIndices into a list
+    set_fact:
+      etcd_members_raft_indices: "{{ groups['oo_etcd_to_config'] | map('extract', hostvars, 'etcd_member_raft_index') | list | unique }}"
+
+  - name: Check the minimum and the maximum of raftIndices is at most 1
+    set_fact:
+      etcd_members_raft_indices_diff: "{{ ((etcd_members_raft_indices | max | int) - (etcd_members_raft_indices | min | int)) | int }}"
+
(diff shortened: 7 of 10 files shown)


playbooks/byo/openshift-etcd/config.yml
```diff
@@ -0,0 +1,34 @@
+---
+- name: Create initial host groups for localhost
```
mtnbikenc: Please use the openshift-cluster/initialize_groups.yml include here instead of running these tasks.  It will flag this as an entry point and run the syntax-check in tox tests.

For example: https://github.com/openshift/openshift-ansible/blob/master/playbooks/byo/openshift-etcd/restart.yml#L2

roles/openshift_etcd_migrate/templates/configuremaster.py
```diff
@@ -0,0 +1,19 @@
+import yaml
```
kwoodson: Is this still needed since you moved to `yedit` module call?

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
```diff
@@ -0,0 +1,91 @@
...
+			v2value = json.loads(so.split("\n")[0])
+		except ValueError:
+			v2value = so.split("\n")[0]
+			is_json=False
+
+		# check if they key is in v3 data
+		so, se, rc = runCommand("ETCDCTL_API=3 %s get %s -w json" % (ectl3, v2key))
+		# The rc or the se does not report if the key is missing or not.
+		# Either the v3value is empty or it is not
+		# TODO(jchaloup) catch the exception if the json.loads fails
```
kwoodson: Is there something you need to do here?

roles/openshift_etcd_migrate/library/inline_python.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/python
```
kwoodson: Where is this module used?

playbooks/byo/openshift-etcd/config.yml
ingvagabund: The first commit actually belongs to https://github.com/openshift/openshift-ansible/pull/3817. Once that get merged, the config.yml will be the way as you suggest.

ingvagabund: The first commit actually belongs to https://github.com/openshift/openshift-ansible/pull/3817. Once that get merged, the config.yml will be the way as you suggest.

The entry point is at ``playbooks/byo/openshift-etcd/migrate.yml`` which already encompasses the suggestion.

playbooks/byo/openshift-etcd/config.yml
```diff
@@ -0,0 +1,34 @@
...
+  tasks:
+  - include_vars: ../openshift-cluster/cluster_hosts.yml
+
+- include: ../../common/openshift-cluster/std_include.yml
+  tags:
+  - always
+
+- include: ../../common/openshift-etcd/config.yml
+  vars:
+    openshift_cluster_id: "{{ cluster_id | default('default') }}"
```
ingvagabund: It is a copy from https://github.com/openshift/openshift-ansible/blob/master/playbooks/byo/openshift-cluster/config.yml#L12-L14

roles/openshift_etcd_migrate/.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+
+  # Create ansible.cfg with correct roles_path
+  - printf '[defaults]\nroles_path=../' >ansible.cfg
+
+script:
+  # Basic role syntax check
+  - ansible-playbook tests/test.yml -i tests/inventory --syntax-check
+
+notifications:
+  webhooks: https://galaxy.ansible.com/api/v1/notifications/
```
ingvagabund: Removed.

roles/openshift_etcd_migrate/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+# defaults file for openshift_etcd_migrate
```
ingvagabund: Removed

roles/openshift_etcd_migrate/README.md
```diff
@@ -0,0 +1,38 @@
+Role Name
+=========
+
+A brief description of the role goes here.
```
ingvagabund: Updated

roles/openshift_etcd_migrate/library/inline_python.py
ingvagabund: It was used to run the master-config.yaml python code. Removed as it is replace by the yedit.

roles/openshift_etcd_migrate/templates/configuremaster.py
ingvagabund: Removed

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
ingvagabund: Just to error with a user friendly message instead of the call stack. I will update it later. Most likely I will replace the python code with the go so I don't have to run the commands through shell.

roles/openshift_etcd_migrate/tests/test.yml
```diff
@@ -0,0 +1,5 @@
+---
+- hosts: localhost
+  remote_user: root
+  roles:
+    - openshift_etcd_migrate
```
ingvagabund: Removed

roles/openshift_etcd_migrate/vars/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+# vars file for openshift_etcd_migrate
```
ingvagabund: Removed

sdodson:
Hmm, we'll need to figure out a delivery mechanism for the go binary. Should we move that to origin packaging?

ashcrow:
```Files changed 2,814``` 😱

File: playbooks/byo/openshift-etcd/migrate.yml [hunk 11]
```diff
@@ -0,0 +1,78 @@
...
+      - "{{ openshift.common.service_type + '-master-api' }}"
+    when:
+    -  openshift.common.is_containerized
+  - set_fact:
+      master_services:
+      - "{{ openshift.common.service_type + '-master-controllers' }}"
+      - "{{ openshift.common.service_type + '-master-api' }}"
+    when:
+    - openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
+    - not openshift.common.is_master_system_container | bool
```
sdodson: @ashcrow can you vet the service names here with respect to openshift.common.is_master_system_container ? Reading through roles/openshift_master/tasks/sytem_container.yml it looks like it'll always be openshift.common.service_type + '-master-{api,controllers}' there, am I correct?

sdodson: I think we want
```
  - set_fact:
      master_services:
      - "{{ openshift.common.service_type + '-master' }}"
  - set_fact:
      master_services:
      - "{{ openshift.common.service_type + '-master-controllers' }}"
      - "{{ openshift.common.service_type + '-master-api' }}"
    when:
    - openshift_master_cluster_method == "native" or openshift.common.is_master_system_container | bool
```

playbooks/byo/openshift-etcd/migrate.yml
```diff
@@ -0,0 +1,78 @@
...
+    include_role:
+      name: openshift_etcd_migrate
+      tasks_from: configure_master
+  - set_fact:
+      master_service: "{{ openshift.common.service_type + '-master-api' if openshift.common.is_containerized else openshift.common.service_type + '-master' }}"
+  - set_fact:
+      master_service: "{{ openshift.common.service_type }}-master-api"
+    when:
+    - openshift.master.ha is defined and openshift.master.ha | bool and openshift_master_cluster_method == "native"
+    - not openshift.common.is_master_system_container | bool
```
sdodson: The service list here should match whatever we stop above.

sdodson:
https://github.com/sdodson/openshift-ansible/tree/etcd-migrate rebases this PR on master and removes 2.3 warnings, no other changes.

roles/openshift_etcd_migrate/tasks/check_cluster_health.yml
```diff
@@ -0,0 +1,23 @@
...
+    etcd_member_healthy: false
+
+- name: Get member item health status
+  set_fact:
+    etcd_member_healthy: true
+  with_items: "{{ etcd_cluster_health.stdout_lines }}"
+  when: "(etcd_peer in item) and ('is healthy' in item)"
+
+- name: Check the etcd cluster health
+  # TODO(jchaloup): should we fail or ask user if he wants to continue? Or just wait until the cluster is healthy?
```
ashcrow: Can we loop X times with a wait and recheck?

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
```diff
@@ -0,0 +1,91 @@
+from subprocess import PIPE, Popen
```
ashcrow: I have to admit that generating code via template makes me a little uneasy. What about passing input into the code instead?

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
```diff
@@ -0,0 +1,91 @@
...
+		# Either the v3value is empty or it is not
+		# TODO(jchaloup) catch the exception if the json.loads fails
+		etcddata = json.loads(so)
+		if not keyExists(etcddata):
+			logging.error("Key '%s' not found in v3 data" % v2key)
+			exit(3)
+
+		# v3 value from the json is base64 encoded
+		if is_json:
+			v3value = json.loads(base64.b64decode(etcddata["kvs"][0]["value"]))
```
ashcrow: This may fail as ``dumps`` expects a string and ``b64decode`` will return ``str`` on Python 2 and ``bytes`` on Python 3.

ashcrow: There are some others in this file that may fail as well.

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
sdodson: I think we can get rid of the templating, can we not just use /etc/etcd/peer.crt /etc/etcd/peer.key and /etc/etcd/ca.crt I think those will be the same regardless of configuration? and then just use `hostname` for the etcdctl commands

sdodson:
If we decide that we need to perserve TTL then we'll integrate that source code into the origin repo and rely on that tool being delivered via openshift 3.6 packaging. As such lets remove the golang code for attachlease from this repo, open a PR against openshfit/origin to add it there. For the time being until that's settled we'll leave the golang binary in this repo for ease of testing.

playbooks/byo/openshift-etcd/migrate.yml
ashcrow: I believe so. Tagging @giuseppe as well.

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
ashcrow: @sdodson 👍. Also, this may be cleaner for getting hostname in a portable way https://docs.python.org/3/library/socket.html#socket.gethostname

playbooks/byo/openshift-etcd/migrate.yml
giuseppe: yes, it is always openshift.common.service_type + '-master-{api,controllers}' for HA

sdodson:
@smarterclayton here's the ansible side of things

smarterclayton:
Yeah, `oc adm migrate etcd-ttl` will be exposed with the same flags (except that ttl-prefix is required to prevent people from accidentally deleting their entire database by default)

smarterclayton:
@liggitt also note that if we reset TTL for all tokens, does the token still enforce the timeout stored in the token itself?

liggitt:
> @liggitt also note that if we reset TTL for all tokens, does the token still enforce the timeout stored in the token itself?

Yes, you just lose the auto-pruning

smarterclayton:
So we want to set the TTL for tokens `>=` the max timeout for tokens on migration.  IF customers have very long default TTLs, we can't afford to set this too low.  Could be dangerous for very long term tokens.

ingvagabund:
Rebasing

roles/openshift_etcd_migrate/tasks/check_cluster_health.yml
ingvagabund: We need to run block of tasks in order to do that. Which is not yet implemented in Ansible. I can include the `check_cluster_health` as many times as is needed and then fail.

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
ingvagabund: Getting rid of the template...

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
ashcrow: An easy way to provide the right type is to use ``six.b``. This will be a ``str`` on Python 2 and ``bytes`` on Python 3.

```python
v3value = json.loads(base64.b64decode(six.b(etcddata["kvs"][0]["value"])))
```

playbooks/byo/openshift-etcd/migrate.yml
ingvagabund: `openshift_master_cluster_method` is not always set so it could be:
```yaml
when:
    - (openshift_master_cluster_method is defined and openshift_master_cluster_method == "native") or openshift.common.is_master_system_container | bool
```

I was not sure about the condition. You are saying we want the condition for both `set_facts`? I.e. lines 42-54? Like merge both set_facts into one with the condition?

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
```diff
@@ -0,0 +1,77 @@
...
+import base64
+
+#logging.basicConfig(level=logging.INFO)
+
+def runCommand(cmd):
+	process = Popen(cmd, stderr=PIPE, stdout=PIPE, shell=True)
+	stdout, stderr = process.communicate()
+	rt = process.returncode
+
+	return stdout, stderr, rt
```
ashcrow: Remember that returning ``stdout`` and``stderr`` will be either ``str`` (Python 2) or ``bytes`` (Python 3).

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
```diff
@@ -0,0 +1,77 @@
...
+
+# TODO(jchaloup): once done ping kwoodson to test the etcd migration role
+
+if __name__ == "__main__":
+
+	ectl2="/usr/bin/etcdctl --cert-file {{ etcd_peer_cert_file }} --key-file {{ etcd_peer_key_file }} --ca-file {{ etcd_peer_ca_file }} -C https://{{ etcd_peer }}:2379"
+	ectl3="/usr/bin/etcdctl --cert {{ etcd_peer_cert_file }} --key {{ etcd_peer_key_file }} --cacert {{ etcd_peer_ca_file }} --endpoints 'https://{{ etcd_peer }}:2379'"
+
+	so, se, rc = runCommand("%s ls -r | sort" % ectl2)
+	for line in so.split("\n"):
```
ashcrow: This will fail on Python 3 as you can not split ``bytes`` this way. If you want to always use ``stdout`` as a ``str`` you will want to use ``.decode('utf-8')`` in your ``runCommand`` method.

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
```diff
@@ -0,0 +1,77 @@
...
+	return True
+
+# TODO(jchaloup): once done ping kwoodson to test the etcd migration role
+
+if __name__ == "__main__":
+
+	ectl2="/usr/bin/etcdctl --cert-file {{ etcd_peer_cert_file }} --key-file {{ etcd_peer_key_file }} --ca-file {{ etcd_peer_ca_file }} -C https://{{ etcd_peer }}:2379"
+	ectl3="/usr/bin/etcdctl --cert {{ etcd_peer_cert_file }} --key {{ etcd_peer_key_file }} --cacert {{ etcd_peer_ca_file }} --endpoints 'https://{{ etcd_peer }}:2379'"
+
+	so, se, rc = runCommand("%s ls -r | sort" % ectl2)
```
ashcrow: nit: ``formt()`` is the more modern way to format strings :-)

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
```diff
@@ -0,0 +1,77 @@
...
+	for line in so.split("\n"):
+		if line == "":
+			continue
+
+		v2key = line
+
+		so, se, rc = runCommand("%s get %s 2>/dev/null" % (ectl2, v2key))
```
ashcrow: nit: ``format()``

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
```diff
@@ -0,0 +1,77 @@
...
+		# if a rc of etcdctl get is 4 => the key does not exist
+		if rc == 1:
+			logging.info("Skipping directory %s" % v2key)
+			continue
+
+		if rc != 0:
+			logging.error("Unable to read %s key from v2 data: %s" % (v2key, se))
+			exit(2)
+
+		v2value = json.loads(so.split("\n")[0])
```
ashcrow: Same warning as above on ``stdout`` being ``bytes`` on Python 3.

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
```diff
@@ -0,0 +1,77 @@
...
+
+		# check if they key is in v3 data
+		so, se, rc = runCommand("ETCDCTL_API=3 %s get %s -w json" % (ectl3, v2key))
+		# The rc or the se does not report if the key is missing or not.
+		# Either the v3value is empty or it is not
+		# TODO(jchaloup) catch the exception if the json.loads fails
+		etcddata = json.loads(so)
```
ashcrow: Same warning as above about ``stdout`` being ``bytes`` on Python 3.

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
```diff
@@ -0,0 +1,77 @@
...
+		if not keyExists(etcddata):
+			logging.error("Key '%s' not found in v3 data" % v2key)
+			exit(3)
+
+		# v3 value from the json is base64 encoded
+		v3value = json.loads(base64.b64decode(etcddata["kvs"][0]["value"]))
+
+		# compare the jsons
+		if v2value != v3value:
+			logging.error("Key '%s' v2 value is different from v3 value:\n'%s'\n\n'%s'" % (v2key, json.dumps(v2value), json.dumps(v3value)))
```
ashcrow: Using ``%s`` is correct here, BUT it should be done between a ``,``. Example:

```python
logging.error("Key '%s' v2 value is different from v3 value:\n'%s'\n\n'%s'", v2key, json.dumps(v2value), json.dumps(v3value))
```

The reason for this is to avoid the creation of said string if the loggger does not need to create the statement. It doesn't save much overhead in this case but it's good to get into the habit.

roles/openshift_etcd_migrate/templates/validatemigration.py.j2
```diff
@@ -0,0 +1,77 @@
...
+			exit(1)
+
+		logging.info("OK %s" % v2key)
```
ashcrow: Same ``logging`` note as above.

ingvagabund:
Updating the role so it complies with Russel's action schema.

roles/openshift_etcd_migrate/files/validatemigration.py
```diff
@@ -0,0 +1,135 @@
...
+	)
+
+	return parser.parse_args()
+
+def runCommand(cmd):
+	process = Popen(cmd, stderr=PIPE, stdout=PIPE, shell=True)
+	stdout, stderr = process.communicate()
+	rt = process.returncode
+
+	return stdout.decode('utf-8'), stderr.decode('utf-8'), rt
```
ashcrow: 👍

roles/openshift_etcd_migrate/files/validatemigration.py
```diff
@@ -0,0 +1,135 @@
...
+	return True
+
+# TODO(jchaloup): once done ping kwoodson to test the etcd migration role
+
+if __name__ == "__main__":
+
+	options, args = setOptions()
+	if options.etcdpeer == "":
+		logging.error("--etcd-peer options not set")
+		exit(1)
```
ashcrow: nit(not required): ``raise SystemExit(1)``

roles/openshift_etcd_migrate/files/validatemigration.py
```diff
@@ -0,0 +1,135 @@
...
+	for line in so.split("\n"):
+		if line == "":
+			continue
+
+		v2key = line
+		if options.skipttlkeys and (v2key.startswith("/kubernetes.io/events") or v2key.startswith("/kubernetes.io/masterleases")):
+			logging.info("Skipping event key %s", v2key)
+			continue
+
+		so, se, rc = runCommand("{} get {} 2>/dev/null".format(ectl2, v2key))
```
ashcrow: 👍

roles/openshift_etcd_migrate/files/validatemigration.py
```diff
@@ -0,0 +1,135 @@
...
+		# check if the key is a directory
+		# if a rc of etcdctl get is 1 => the key is a directory
+		# if a rc of etcdctl get is 4 => the key does not exist
+		if rc == 1:
+			logging.info("Skipping directory %s", v2key)
```
ashcrow: 👍

roles/openshift_etcd_migrate/files/validatemigration.py
```diff
@@ -0,0 +1,135 @@
...
+# TODO(jchaloup): once done ping kwoodson to test the etcd migration role
+
+if __name__ == "__main__":
+
+	options, args = setOptions()
+	if options.etcdpeer == "":
+		logging.error("--etcd-peer options not set")
+		exit(1)
+
+	ectl2="/usr/bin/etcdctl --cert-file {} --key-file {} --ca-file {} -C https://{}:2379".format(options.certfile, options.keyfile, options.cafile, options.etcdpeer)
```
ashcrow: With ``shell=True`` in ``runCommand()`` do you need to give the full path to ``etcdctl``?

@giuseppe will this cause any issues with your etcd container work?

ashcrow:
@mtnbikenc has @ingvagabund's work addressed your review?

ashcrow:
@ingvagabund ``roles/etcd_migrate/files/vendor/`` will likely need to be added to the ignore list for ``flake8``.

ingvagabund:
> @ingvagabund roles/etcd_migrate/files/vendor/ will likely need to be added to the ignore list for flake8.

Entire `roles/etcd_migrate/files` will go away once https://github.com/openshift/origin/pull/14559 is merged and available in rpms.

ingvagabund:
Need to update the master config configuration based on https://github.com/openshift/openshift-ansible/pull/4435#discussion_r122117399

ingvagabund:
aos-ci-test

ingvagabund:
It needs to be run over 3.6.111-1 at least (due to new `oadm migrate etcd-ttl` command). I can add check for that later if needed.

ingvagabund:
aos-ci-test

sdodson:
[merge][severity: bug]

PR #3955, opened 2017-04-19
Add /etc/sysconfig/etcd to etcd_container

Fale:
The `/etc/sysconfig/etcd` file is often used to allow the usage of additional variables passed to the etcd daemon.
Example of this is to make some new nodes join an existing cluster.

Having two files simplify the automation around it, since in the `/etc/sysconfig/etcd` is possible to put temporary variables and then nuke the file as soon as they are not used anymore.

PR #3958, opened 2017-04-19
Add /etc/sysconfig/etcd to etcd_container

Fale:
The `/etc/sysconfig/etcd` file is often used to allow the usage of additional variables passed to the etcd daemon.
Example of this is to make some new nodes join an existing cluster.

Having two files simplify the automation around it, since in the `/etc/sysconfig/etcd` is possible to put temporary variables and then nuke the file as soon as they are not used anymore.

### DIFF
--- a/roles/etcd/tasks/main.yml
+++ b/roles/etcd/tasks/main.yml
@@ -37,6 +37,12 @@
     mode: 0700
   when: etcd_is_thirdparty | bool
 
+- name: Create configuration file in sysconfig
+  file:
+    path: /etc/sysconfig/etcd
+    state: touch
+  when: etcd_is_thirdparty | bool
+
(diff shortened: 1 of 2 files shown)


Fale:
This is the 3.5 backport of #3955

sdodson:
bot, test pull request

sdodson:
aos-ci-test

sdodson:
I don't know why the CI jobs aren't hitting this, but QE reported that with this merged if /etc/sysconfig/etcd doesn't exist then the container fails. Is there a way to make the file optional in docker run ? If not maybe we should touch the file to make sure it exists and then mount it in as a volume instead?

ingvagabund:
aos-ci-test

ingvagabund:
[merge]

PR #3960, opened 2017-04-20
Fix private CA

Fale:
When private (aka not publicly trusted) CAs are used, and a registry from outside the cluster itself is used, it is not possible to retrieve images from it (without this fix).

This is only a problem of the containerized version, since in the RPM version, these files are loaded by design.

I think that also this adds no possible threats, since we are adding to the master-api containers CAs that are already trusted by the nodes, which have a higher security criticality.

This is a backport #3882.


ingvagabund:
aos-ci-test

sdodson:
I don't know why the CI jobs aren't hitting this, but QE reported that with this merged if /etc/sysconfig/etcd doesn't exist then the container fails. Is there a way to make the file optional in `docker run` ? If not maybe we should touch the file to make sure it exists and then mount it in as a volume instead?

Fale:
Closed because it has already been implemented. As #3882

PR #3961, opened 2017-04-20
Container image documentation improvements

codificat:
Feedback from testing the container image highlighted the need to clarify a few points:

- that the image expects to be run as non-root
- the path to the ssh keys (related to the user the container runs as)
- the need and consequences of SELinux flags on container mount points

These changes attempt to address these gaps.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,11 +1,10 @@
-# openshift-ansible RPM Build instructions
+# openshift-ansible build instructions
+
+## Build openshift-ansible RPMs
 
 We use tito to make building and tracking revisions easy.
 
 For more information on tito, please see the [Tito home page](https://github.com/dgoodwin/tito "Tito home page").
-
-
-## Build openshift-ansible
 
 - Change into openshift-ansible
 ```
@@ -24,3 +23,22 @@ tito tag
 ```
 tito build --rpm
 ```
+
+## Build an openshift-ansible container image
+
+To build a container image of `openshift-ansible` using standalone **Docker**:
+
+        cd openshift-ansible
+        docker build -t openshift/openshift-ansible .
+
+Alternatively this can be built using on **OpenShift** using a [build and image stream](https://docs.openshift.org/latest/architecture/core_concepts/builds_and_image_streams.html) with this command:
+
+        oc new-build docker.io/aweiteka/playbook2image~https://github.com/openshift/openshift-ansible
+
+The progress of the build can be monitored with:
+
+        oc logs -f bc/openshift-ansible
+
+Once built, the image will be visible in the Image Stream created by the same command:
+
+        oc describe imagestream openshift-ansible
--- a/README_CONTAINER_IMAGE.md
+++ b/README_CONTAINER_IMAGE.md
@@ -1,23 +1,10 @@
 # Containerized openshift-ansible to run playbooks
 
-The [Dockerfile](Dockerfile) in this repository uses the [playbook2image](https://github.com/aweiteka/playbook2image) source-to-image base image to containerize `openshift-ansible`. The resulting image can run any of the provided playbooks.
+The [Dockerfile](Dockerfile) in this repository uses the [playbook2image](https://github.com/aweiteka/playbook2image) source-to-image base image to containerize `openshift-ansible`. The resulting image can run any of the provided playbooks. See [BUILD.md](BUILD.md) for image build instructions.
 
-**Note**: at this time there are known issues that prevent to run this image for installation/upgrade purposes (i.e. run one of the config/upgrade playbooks) from within one of the hosts that is also an installation target at the same time: if the playbook you want to run attempts to manage the docker daemon and restart it (like install/upgrade playbooks do) this would kill the container itself during its operation.
-
-## Build
-
-To build a container image of `openshift-ansible`:
-
-1. Using standalone **Docker**:
-
-        cd openshift-ansible
-        docker build -t openshift/openshift-ansible .
 The image is designed to **run as a non-root user**. The container's UID is mapped to the username `default` at runtime. Therefore, the container's environment reflects that user's settings, and the configuration should match that. For example `$HOME` is `/opt/app-root/src`, so ssh keys are expected to be under `/opt/app-root/src/.ssh`. If you ran a container as `root` you would have to adjust the container's configuration accordingly, e.g. by placing ssh keys under `/root/.ssh` instead. Nevertheless, the expectation is that containers will be run as non-root; for example, this container image can be run inside OpenShift under the default `restricted` [security context constraint](https://docs.openshift.org/latest/architecture/additional_concepts/authorization.html#security-context-constraints).
 
-1. Using an **OpenShift** build:
-
-        oc new-build docker.io/aweiteka/playbook2image~https://github.com/openshift/openshift-ansible
-        oc describe imagestream openshift-ansible
+**Note**: at this time there are known issues that prevent to run this image for installation/upgrade purposes (i.e. run one of the config/upgrade playbooks) from within one of the hosts that is also an installation target at the same time: if the playbook you want to run attempts to manage the docker daemon and restart it (like install/upgrade playbooks do) this would kill the container itself during its operation.
 
 ## Usage
 
--- a/README_CONTAINER_IMAGE.md
+++ b/README_CONTAINER_IMAGE.md
@@ -12,6 +12,7 @@ To build a container image of `openshift-ansible`:
 
         cd openshift-ansible
         docker build -t openshift/openshift-ansible .
+The image is designed to **run as a non-root user**. The container's UID is mapped to the username `default` at runtime. Therefore, the container's environment reflects that user's settings, and the configuration should match that. For example `$HOME` is `/opt/app-root/src`, so ssh keys are expected to be under `/opt/app-root/src/.ssh`. If you ran a container as `root` you would have to adjust the container's configuration accordingly, e.g. by placing ssh keys under `/root/.ssh` instead. Nevertheless, the expectation is that containers will be run as non-root; for example, this container image can be run inside OpenShift under the default `restricted` [security context constraint](https://docs.openshift.org/latest/architecture/additional_concepts/authorization.html#security-context-constraints).
 
 1. Using an **OpenShift** build:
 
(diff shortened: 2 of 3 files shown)


README_CONTAINER_IMAGE.md
```diff
@@ -2,42 +2,48 @@
...
+**Note**: at this time there are known issues that prevent to run this image for installation/upgrade purposes (i.e. run one of the config/upgrade playbooks) from within one of the hosts that is also an installation target at the same time: if the playbook you want to run attempts to manage the docker daemon and restart it (like install/upgrade playbooks do) this would kill the container itself during its operation.
 
 ## Usage
 
 The `playbook2image` base image provides several options to control the behaviour of the containers. For more details on these options see the [playbook2image](https://github.com/aweiteka/playbook2image) documentation.
 
-At the very least, when running a container using an image built this way you must specify:
+At the very least, when running a container you must specify:
+
+1. An **inventory** file. This can be mounted inside the container as a volume and specified with the `INVENTORY_FILE` environment variable. Alternatively you can serve the inventory file from a web server and use the `INVENTORY_URL` environment variable to fetch it, or `DYNAMIC_SCRIPT_URL` to download a script that provides a dynamic inventory.
```
aweiteka: How about "**Inventory**: This can be mounted..." so it's clear they have 3 options (one of which is an actual file).

README_CONTAINER_IMAGE.md
```diff
@@ -2,42 +2,48 @@
...
+
+* `-v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z` mounts your ssh key (`$HOME/.ssh/id_rsa`) under the `default` user's `$HOME/.ssh` in the container (as explained above, `/opt/app-root/src` is the `$HOME` of the `default` user in the container). If you mount the ssh key into a non-standard location you can add an environment variable with `-e ANSIBLE_PRIVATE_KEY_FILE=/the/mount/point` or set `ansible_ssh_private_key_file=/the/mount/point` as a variable in the inventory to point Ansible at it.
+
+  Note that the ssh key is mounted with the `:Z` flag: this is also required so that the container can read the ssh key from its restricted SELinux context; this means that *your original ssh key file will be re-labeled* to something like `system_u:object_r:container_file_t:s0:c113,c247`. For more details about `:Z` please check the `docker-run(1)` man page. Please keep this in mind when providing these mount specifications because this could have unexpected consequences: for example, if you mount (and therefore re-label) your whole `$HOME/.ssh` directory you will block `sshd` from accessing your keys.
```
aweiteka: I experienced this relabeling pain and never found a workaround. Are there any alternatives or is this the best we can do?

README_CONTAINER_IMAGE.md
```diff
@@ -2,42 +2,48 @@
...
+
+* `-v /etc/ansible/hosts:/tmp/inventory` and `-e INVENTORY_FILE=/tmp/inventory` mount the Ansible inventory file into the container as `/tmp/inventory` and set the corresponding environment variable to point at it respectively. The example uses `/etc/ansible/hosts` as the inventory file as this is a default location, but your inventory is likely to be elsewhere so please adjust as needed. Note that depending on the file you point to you might have to handle SELinux labels in a similar way as with the ssh keys, e.g. by adding a `:z` flag to the volume mount.
+
+  You might prefer to copy your ssh key and inventory files to a separate dedicated location before mounting them into the container in order to leave the original files' SELinux labels untouched.
```
aweiteka: So here's the workaround. Can we clarify this is the **host**? "You might prefer to copy your ssh key and inventory files to a different location on the host before mounting..."

README_CONTAINER_IMAGE.md
codificat: Good catch. I'll remove that *file* reference and reword a bit to not imply that mounting a volume is a requiment for INVENTORY_FILE (i.e. for when it can point to a path that is already part of the image)

README_CONTAINER_IMAGE.md
codificat: Thanks, I'll make that part a bit more clear by moving it out of an individual option...

codificat:
@aweiteka thanks for the comments, hopefully addressed now.

@wkshi could you take a look and see if this addresses the concerns you raised?

CC @rhcarvalho and @sosiouxme also FYI if you have additional comments, thanks.

BUILD.md
```diff
@@ -24,3 +24,17 @@ tito tag
...
+
+* Using standalone **Docker**:
+
+        cd openshift-ansible
+        docker build -t openshift/openshift-ansible .
+
+* Using an **OpenShift** build:
+
+        oc new-build docker.io/aweiteka/playbook2image~https://github.com/openshift/openshift-ansible
+        oc describe imagestream openshift-ansible
```
rhcarvalho: I see it was already there and just being moved over to this file. What is this step intended for?

BUILD.md
```diff
@@ -24,3 +24,17 @@ tito tag
 ```
 tito build --rpm
 ```
+
+## Build openshift-ansible container image
```
rhcarvalho: This file starts with

```markdown
# openshift-ansible RPM Build instructions
 
We use tito to make building and tracking revisions easy.
```

Thus without reading through all of the file, I'd think of it as containing instructions for building an RPM.

Maybe we can update the introduction above, or simply keep all the information about the image in a single file?

README_CONTAINER_IMAGE.md
```diff
@@ -2,42 +2,48 @@
 
 The [Dockerfile](Dockerfile) in this repository uses the [playbook2image](https://github.com/aweiteka/playbook2image) source-to-image base image to containerize `openshift-ansible`. The resulting image can run any of the provided playbooks.
 
-**Note**: at this time there are known issues that prevent to run this image for installation/upgrade purposes (i.e. run one of the config/upgrade playbooks) from within one of the hosts that is also an installation target at the same time: if the playbook you want to run attempts to manage the docker daemon and restart it (like install/upgrade playbooks do) this would kill the container itself during its operation.
-
-## Build
```
rhcarvalho: Even if we move the build part to another file, it might make sense to link to that from this file?

README_CONTAINER_IMAGE.md
```diff
@@ -2,42 +2,48 @@
...
-Here is an example of how to run a containerized `openshift-ansible` playbook that will check the expiration dates of OpenShift's internal certificates using the [`openshift_certificate_expiry` role](roles/openshift_certificate_expiry). The inventory and ssh keys are mounted as volumes (the latter requires setting the uid in the container and SELinux label in the key file via `:Z` so they can be accessed) and the `PLAYBOOK_FILE` environment variable is set to point to an example certificate check playbook that is already part of the image:
+Here is an example of how to run a containerized `openshift-ansible` playbook that will check the expiration dates of OpenShift's internal certificates using the [`openshift_certificate_expiry` role](roles/openshift_certificate_expiry):
 
     docker run -u `id -u` \
            -v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z \
            -v /etc/ansible/hosts:/tmp/inventory \
            -e INVENTORY_FILE=/tmp/inventory \
-           -e OPTS="-v" \
            -e PLAYBOOK_FILE=playbooks/certificate_expiry/default.yaml \
+           -e OPTS="-v" -t \
```
rhcarvalho: Probably irrelevant, but maybe the `-t` would be more "visible" in the first line? At least my brain reads these lines as "here we're calling docker, pay attention to the args, here's a volume mount, here's another one, and then here are extra env vars" -- I'd totally overlook anything else in the same line.

I mean:

```
docker run -t -u `id -u` \
...
```

BUILD.md
codificat: Oops, good catch.

I believe that moving the build instructions into `BUILD.md` makes sense, so I'd vote for updating the title/description in that file and link to it from the container image readme as you mention below.

If that's ok, I'll move the reference to `tito` into the section about building an RPM, and move "`RPM`" from the level 1 heading into the first level 2 heading

BUILD.md
codificat: I added the `oc describe is` to show how to find the built image... I thought that the `oc new-build` alone was missing some information/feedback... probably we could add some more info around it? It's only that an explanation of how OpenShift builds and imagestreams work is out of scope here.

I'll just add a link to the OpenShift docs around that if that's ok.

README_CONTAINER_IMAGE.md
codificat: ack, will add a link.

README_CONTAINER_IMAGE.md
codificat: I fully see your point, I actually wrote it that way initially (my brain seems to work that way too ;) but then I moved it down next to the `-v` option to more or less "group" the related options together, so that the detailed explanation that follows was a bit more coherent.

Or in other words: looking at the command invocation one might wonder "why does that need `-t`?" - and I think that having it next to `-e OPTS...` makes it slightly more clear...

BUILD.md
rhcarvalho: Maybe "Once the build is finished, you can get more information about the built image with: ..."?

As for _feedback_, I think the best way is to use `oc logs`? Or IIRC there's a flag to new-build to follow the logs...

BUILD.md
rhcarvalho: > I believe that moving the build instructions into BUILD.md makes sense, so I'd vote for updating the title/description in that file and link to it from the container image readme as you mention below.

:+1: 

Just note that there are other references pointing to BUILD.md, e.g. I know one in README.md, so those might need some adjustment too.

README_CONTAINER_IMAGE.md
rhcarvalho: From the perspective of the explanation, I understand that. As I said, not a very relevant point ;) whatever you choose is good, the explanation itself is more important than the argument order.

BUILD.md
sosiouxme: I don't think there's a flag on new-build to follow the logs.

BUILD.md
codificat: I didn't find a reference in README.md, there's one in CONTRIBUTING.md that I'll update to mention both RPM and container image.

BUILD.md
codificat: Right, there's no flag for new-build. For clarity I'll mention both `oc log` and point to the ImageStream...

README_CONTAINER_IMAGE.md
codificat: Ok, I'll leave it that way then :)

codificat:
Latest comments addressed

wkshi:
@codificat It's good to me,  much clearly.

BUILD.md
rhcarvalho: Right, indeed I'm moving it from CONTRIBUTING to README in #3983.

rhcarvalho:
@sdodson what are the `fedora/*` jobs?

I noticed that it may be impossible to determine what caused a failure with the output it provides.
- https://s3.amazonaws.com/aos-ci/ghprb/openshift/openshift-ansible/fcc1ef69b5b007374ab9bfe3aa10a0f6e90252dc.0.1493068819663035231/index.html
- https://s3.amazonaws.com/aos-ci/ghprb/openshift/openshift-ansible/fcc1ef69b5b007374ab9bfe3aa10a0f6e90252dc.1.1493068823182445637/index.html

In both cases the logs point to artifacts that are not available:
```
Unable to start service origin-master: Job for origin-master.service failed because the control process exited with error code.
See "systemctl status origin-master.service" and "journalctl -xe" for details.
```

sdodson:
@rhcarvalho @jlebon is the poc for the fedora jobs. I think though in general we should consider dumping logs whenever the services fail to start and then `fail:`

codificat:
aos-ci-test

jlebon:
Going down the list of tests on the master branch commits, it looks like this started breaking at https://github.com/openshift/openshift-ansible/pull/3955. @sdodson does that sound plausible?

Yup, we can tweak the `.redhat-ci.yml` so that we also upload the journal on failure.

sdodson:
@jlebon yeah, that makes sense, i'm going to revert that PR

jlebon:
I opened a PR to upload journal logs on failures: https://github.com/openshift/openshift-ansible/pull/4005.

codificat:
This should be good to [merge]

PR #3968, opened 2017-04-21
oc_obj: Allow for multiple kinds in delete

jarrpa:
This is to allow "kind" to include a string like "svc,ep" to delete both Services and Endpoints matching a given name or selector.

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

### DIFF
--- a/roles/lib_openshift/library/oc_obj.py
+++ b/roles/lib_openshift/library/oc_obj.py
@@ -98,7 +98,7 @@ options:
     aliases: []
   kind:
     description:
-    - The kind attribute of the object. e.g. dc, bc, svc, route
+    - The kind attribute of the object. e.g. dc, bc, svc, route. May be a comma-separated list, e.g. "dc,po,svc".
     required: True
     default: None
     aliases: []
@@ -1539,12 +1539,9 @@ class OCObject(OpenShiftCLI):
         # Delete
         ########
         if state == 'absent':
-            # if we were passed a name, verify its not in our results
-            if params['name'] is not None and not Utils.exists(api_rval['results'], params['name']):
-                return {'changed': False, 'state': state}
-
-            # verify results are empty for the selector
-            if params['selector'] is not None and len(api_rval['results']) == 0:
+            # verify its not in our results
+            if (params['name'] is not None or params['selector'] is not None) and \
+               (len(api_rval['results']) == 0 or len(api_rval['results'][0].getattr('items', [])) == 0):
                 return {'changed': False, 'state': state}
 
             if check_mode:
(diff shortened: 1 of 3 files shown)


kwoodson:
@jarrpa, @ashcrow, would it make sense to make the kind a list so the interface is a bit cleaner?
```yaml
kind: 
- ep
- svc
```
I think this makes the api a bit cleaner.  Thoughts?

jarrpa:
@kwoodson @ashcrow Maybe. I just went with the `svc,ep` format because that's what the command line format is. It'd be fairly trivial to do something like `self.kind.join(',')`, but I'm not sure what implications that would have on the rest of `oc_obj`. As the comma-separated format works with everything else as-is, maybe consider a cleanup of the API as a near future PR (this PR would also include the required change to the GlusterFS code)? I could probably take that on (since this is my fauilt ;) ), I just foresee a lot of careful review being required and would rather that not hold up my current work.

ashcrow:
I think a list would be cleaner, but I think a ``'[ ]?,'`` splittable string is good enough for now.

tbielawa:
I think the comma-separated string format could be OK. That's how the command line expects to take arguments, right? So it's consistent with that.

But, this is a computer program, and a user might expect to provide a series of items as a `list` (data structure). So it might not be consistent with what users expect.

I'm good with either way. Obviously if it changes to accepting a `list` instead of a comma separated string more changes will be required.

If we stick with the comma separated string then I would like to see `roles/lib_openshift/src/doc/obj` updated with a note indicating a comma separated list of values is also acceptable input.

jarrpa:
Updated the documentation as requested by @tbielawa.

ashcrow:
```
Checking generated module code: roles/lib_openshift/src/generate.py
('Generated content does not match for /home/travis/build/openshift/openshift-ansible/roles/lib_openshift/src/../library/oc_obj.py',)
Checking generated module code: roles/lib_utils/src/generate.py
```

jarrpa:
@ashcrow Yup, just caught that. :) Addressed and updated.

roles/lib_openshift/library/oc_obj.py
```diff
@@ -1539,12 +1539,9 @@ def run_ansible(params, check_mode=False):
...
         ########
         if state == 'absent':
-            # if we were passed a name, verify its not in our results
-            if params['name'] is not None and not Utils.exists(api_rval['results'], params['name']):
-                return {'changed': False, 'state': state}
-
-            # verify results are empty for the selector
-            if params['selector'] is not None and len(api_rval['results']) == 0:
+            # verify its not in our results
+            if (params['name'] is not None or params['selector'] is not None) and \
```
kwoodson: I actually prefer the logic in 2 separate steps as its more clear to me what's happening.  

I won't block on this.

kwoodson:
aos-ci-test

kwoodson:
[merge]

PR #3971, opened 2017-04-21
Fix issue where the password would not be passed to the htpasswd command

mwringe:
If we don't use the interactive mode with the docker command, then it doesn't accept the password being passed.

This means the htpasswd file is created with an empty password instead of an actual password.


mwringe:
@ewolinetz can you take a look please?

mwringe:
Is there anything we can do to get this merged?

ewolinetz:
aos-ci-test

ewolinetz:
[test]

PR #3972, opened 2017-04-21
Update byo hosts examples with stronger warnings about Docker subnet overlap

lpsantil:
Takes warnings from:
https://docs.openshift.com/container-platform/3.4/install_config/install/advanced_install.html#configuring-host-variables
https://docs.openshift.com/container-platform/3.4/admin_guide/sdn_troubleshooting.html#debugging-steps

and makes them stronger and provides potential workaround.

### DIFF
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -571,9 +571,16 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # network blocks should be private and should not conflict with network blocks
 # in your infrastructure that pods may require access to. Can not be changed
 # after deployment.
+#
+# WARNING : Do not pick subnets that overlap with the default Docker bridge subnet of
+# 172.17.0.0/16.  Your installation will fail and/or your configuration change will
+# cause the Pod SDN or Cluster SDN to fail.
+#
+# WORKAROUND : If you must use an overlapping subnet, you can configure a non conflicting
+# docker0 CIDR range by adding '--bip=192.168.2.1/24' to DOCKER_NETWORK_OPTIONS
+# environment variable located in /etc/sysconfig/docker-network.
 #osm_cluster_network_cidr=10.128.0.0/14
 #openshift_portal_net=172.30.0.0/16
-
 
 # ExternalIPNetworkCIDRs controls what values are acceptable for the
 # service external IP field. If empty, no externalIP may be set. It
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -572,9 +572,16 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # network blocks should be private and should not conflict with network blocks
 # in your infrastructure that pods may require access to. Can not be changed
 # after deployment.
+#
+# WARNING : Do not pick subnets that overlap with the default Docker bridge subnet of
+# 172.17.0.0/16.  Your installation will fail and/or your configuration change will
+# cause the Pod SDN or Cluster SDN to fail.
+#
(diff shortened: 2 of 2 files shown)


sdodson:
@openshift/networking Is this the best workaround?

sdodson:
bot, test pull request

sdodson:
Looks fine to me assuming networking is happy with the workaround. I squashed it down to one commit.

inventory/byo/hosts.origin.example
```diff
@@ -571,10 +571,18 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
 # network blocks should be private and should not conflict with network blocks
 # in your infrastructure that pods may require access to. Can not be changed
 # after deployment.
+#
+# WARNING : Do not pick subnets that overlap with the default Docker bridge subnet of
+# 172.16.0.0/16.  Your installation will fail and/or your configuration change will
```
danwinship: 172.**17**.0.0/16

sdodson: That contradicts what https://docs.openshift.com/container-platform/3.4/install_config/install/advanced_install.html#configuring-host-variables says, which is correct?

```
Defaults to 172.30.0.0/16, and cannot be re-configured after deployment. If changing
from the default, avoid 172.16.0.0/16, which the docker0 network bridge uses by
default, or modify the docker0 network.
```

inventory/byo/hosts.origin.example
```diff
@@ -571,10 +571,18 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
...
+# cause the Pod SDN or Cluster SDN to fail.
+#
+# WORKAROUND : If you must use an overlapping subnet, you can relocate the Docker
+# default subnet by doing the following before installing OCP:
+#     $ echo 'DOCKER_NETWORK_OPTIONS="--bip=192.168.2.1/24"' >> /etc/sysconfig/docker-network
```
danwinship: That will break things if they were already assigning to DOCKER_NETWORK_OPTIONS in that file.

Do we really need to spell out how to do this here anyway? Just add "docker's private network" to the list of things not to conflict with in the existing paragraph.

inventory/byo/hosts.origin.example
danwinship: Those docs are wrong. The docker default is 172.17.

inventory/byo/hosts.origin.example
sdodson: Yeah, I'd just mention the flag is --bip and leave it up to them to choose where to set it.

DanyC97:
ouch, been bitten by this one... glad i came across it, thx guys

lpsantil:
I've corrected the docker0 subnet.

inventory/byo/hosts.origin.example
lpsantil: I hate docs that say "here's a problem" and never point you to a (full) solution/workaround.  Also, this particular problem/workaround is not well recognized via google-fu.

sdodson: The concern is that this would overwrite anything else in DOCKER_NETWORK_OPTIONS. Ideally everyone would mentally interpret that and go edit the file by hand but if we give someone something to copy and paste then there's a good change they'll just do that and won't notice they've just blown away any other settings in that variable already.

How about lets just say "Configure a non conflicting docker0 CIDR range by adding '--bip=192.168.2.1/24' to DOCKER_NETWORK_OPTIONS environment variable located in /etc/sysconfig/docker-network" ?

lpsantil: Thanks for the suggestion.  Corrected.

sdodson:
aos-ci-test

sdodson:
Docs only change, avoiding CI queue and merging manually.

PR #3973, opened 2017-04-21
Removing resourceVersion when replacing objects.

kwoodson:


kwoodson:
@ashcrow @sdodson,

I wrote a test to patch a resource 3 times in a tight loop and then run oc_edit in a loop externally.
- Before the change we were seeing the error > 50-60% of the time where the ` the object has been modified`.  
- After the change I was able to see this occur 0-2% of the time in the same test.

I think this is the best solution to reduce the race condition.

ashcrow:
[test]

kwoodson:
aos-ci-test

kwoodson:
[test]

kwoodson:
aos-ci-test

kwoodson:
[test]

kwoodson:
[merge]

PR #3974, opened 2017-04-21
Allowing CA cert to still be generated is we are doing standalone hea…

ewolinetz:
…pster

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1443741

@jcantrill 

Confirmed that we only generate the CA cert once when openshift_metrics_heapster_standalone is True or False


ewolinetz:
cc @mwringe

ewolinetz:
[test]

PR #3976, opened 2017-04-21
Update swap disable tasks

mtnbikenc:
* Lines are commented out vice being removed
* Comment is added indicating why the change happened
* Variable openshift_disable_swap added to allow user control

https://trello.com/c/vGmZYJ79

### DIFF
--- a/roles/openshift_node_upgrade/tasks/main.yml
+++ b/roles/openshift_node_upgrade/tasks/main.yml
@@ -99,10 +99,6 @@
 - name: Check for swap usage
   command: grep "^[^#].*swap" /etc/fstab
   # grep: match any lines which don't begin with '#' and contain 'swap'
-  # command: swapon --summary
-  # Alternate option, however if swap entries are in fstab, swap will be
-  # enabled at boot. Grepping fstab should catch a condition when swap was
-  # disabled, but the fstab entries were not removed.
   changed_when: false
   failed_when: false
   register: swap_result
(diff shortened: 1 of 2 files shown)


mtnbikenc:
aos-ci-test

mtnbikenc:
aos-ci-test

roles/openshift_node_upgrade/tasks/main.yml
```diff
@@ -114,12 +110,21 @@
...
       dest: /etc/fstab
-      regexp: 'swap'
-      state: absent
+      line: '# OpenShift-Ansible Installer disabled swap per overcommit guidelines'
+      state: present
 
-  when: swap_result.stdout_lines | length > 0
+  when:
+  - swap_result.stdout_lines | length > 0
+  - openshift_disable_swap | default(true)
```
sdodson: I think I've encountered travis complaining when the when conditions weren't indented, any idea why it seems to be inconsistent? Here they're not intended, but in openshift_node/tasks/main.yml they are.

mtnbikenc: The yamllint checks to make sure the _file_ is consistent.  So, if it starts out indented in the file, then it needs to be the same through the rest of the file.  When we started doing yamllint checks, we didn't specify that it _had_ to be one way or the other because there was quite a mix throughout the code base.
https://github.com/openshift/openshift-ansible/blob/master/.yamllint#L53

sdodson: Thanks for the explanation.

mtnbikenc: Out of curiosity I checked to see what the failures would be if we forced `true` or `false` for `indent-sequences`

yamllint errors with:
`indent-sequences: false`: 200
`indent-sequences: true`: 1345

mtnbikenc:
[merge]

sdodson:
flake on https://github.com/openshift/origin/issues/8427

sdodson:
[merge]

PR #3987, opened 2017-04-24
BACKPORT 1.5: Removing resource version to avoid race conditions in lib_openshift.

kwoodson:


kwoodson:
@mwoodson, This is the fix for the race condition.

sdodson:
aos-ci-test

PR #3990, opened 2017-04-24
Replace custom router certificate during redeploy

abutcher:
The default service serving router certificate shares a secret name (`router-certs`) with a custom router certificate specified during install so we should differentiate between the two when replacing them.

These changes allow the router certificate specified by `openshift_hosted_router_certificate` to be replaced when redeploying router certificates.


sdodson:
CI unfortunately doesn't call these playbooks, merging.

PR #3992, opened 2017-04-24
Allowing CA cert to still be generated is we are doing standalone hea…

ewolinetz:
…pster

picks #3974


ewolinetz:
apparently this file does not exist in master, closing

PR #3994, opened 2017-04-25
Add python-boto requirement

codificat:
The [AWS inventory](https://github.com/openshift/openshift-ansible/blob/master/inventory/aws/hosts/ec2.py) uses the Boto library.

In order for containerized openshift-ansible to be able to use that inventory, the container image needs to include that package.

By adding it to `requirements.txt`, the container image build will install it at build time.

For the productized version: adding a dependency on `python-boto` to the specfile does not sound like a good idea: it would introduce a dependency on the *Common* RPM repo - and right now we don't even package the inventories (other than the sample static inventories that are part of the *-docs* package). Therefore, just install the RPM into the image.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 # Versions are pinned to prevent pypi releases arbitrarily breaking
 # tests with new APIs/semantics. We want to update versions deliberately.
 ansible==2.2.2.0
+boto==2.45.0
 click==6.7
 pyOpenSSL==16.2.0
 # We need to disable ruamel.yaml for now because of test failures
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -20,9 +20,10 @@ LABEL name="openshift3/openshift-ansible" \
 # because all content and dependencies (like 'oc') is already
 # installed via yum.
 USER root
-RUN INSTALL_PKGS="atomic-openshift-utils atomic-openshift-clients" && \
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -1,6 +1,7 @@
 # Versions are pinned to prevent pypi releases arbitrarily breaking
 # tests with new APIs/semantics. We want to update versions deliberately.
 ansible==2.2.2.0
+boto==2.45.0
```
rhcarvalho: This is like an "optional" requirement, right?
I think we don't have a good way to express that.

What if we just install boto in both images, without adding it to `requirements.txt`?

codificat: The thing is that right now the Dockerfile uses playbook2image's `assemble` script to put things together, including python package install via `pip install -r requirements.txt`. In other words: I think that the way to install Boto into the image in a "nice" way that fits how things are set up now is to add it to `requirements.txt`.

I didn't think of a negative impact here (unlike the Dockerfile.rhel7 case)... how bad is it to have that non-hard requirement in `requirements.txt`?

rhcarvalho: Perhaps not that bad. It means that whoever uses `pip install -r requirements.txt` will get `boto` even if they don't use it... but then, we might already have or will have other cases like this -- dependencies that are related only to a specific usage of the project.

:+1: to adding it to `requirements.txt` and keeping the image build simple.

codificat:
aos-ci-test

codificat:
aos-ci-test

rhcarvalho:
[merge]

PR #3998, opened 2017-04-25
Revert "Add /etc/sysconfig/etcd to etcd_container"

sdodson:
Reverts openshift/openshift-ansible#3955


sdodson:
@fale FYI, we need to come up with something to make sure that if the file doesn't exist the container starts properly.

Fale:
I think another option would be to add a ExecStartPre that touches the file if it does not exists, would this be better?

Fale:
ie:

    if [[ ! -f /etc/sysconfig/etcd ]]; then touch /etc/sysconfig/etcd; fi

Issue #3997, opened 2017-04-25
New Metrics/Logging Roles Do Not Restart HA Services


dgoodwin:
#### Description

The new metrics/logging role update_master_config.yml tasks have a restart handler only for single master deployments, in an HA deployment calling the role will not restart any services.

If this is intentional, we need some kind of contract defined such that the caller can identify whether or not a restart is necessary. Right now we have to blindly restart as the roles do not expose any variable / register indicating if a restart is required.

##### Version

openshift-ansible-3.5.52

PR #4004, opened 2017-04-25
Updating logging and metrics to restart api, ha and controllers when …

ewolinetz:
…updating master config

Ideally we would have a common role that contains handlers that we could make a dependency, but this should work for now.

Also noticed a bug while testing in logging regarding updating a route (fix in oc_apply.yaml)

cc @dgoodwin


ewolinetz:
Should address #3997

ewolinetz:
Handler taken from https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_master/handlers/main.yml

roles/openshift_logging/tasks/oc_apply.yaml
```diff
@@ -49,4 +49,4 @@
     failed_when: "'error' in generation_apply.stderr"
     changed_when: generation_apply.rc == 0
     when: "'field is immutable' in generation_apply.stderr"
-  when: file_content.kind == "Service"
+  when: file_content.kind in ["Service", "Route"]
```
sdodson: unrelated changes?

roles/openshift_metrics/tasks/update_master_config.yaml
```diff
@@ -4,6 +4,9 @@
     dest: "{{ openshift.common.config_base }}/master/master-config.yaml"
     yaml_key: assetConfig.metricsPublicURL
     yaml_value: "https://{{ openshift_metrics_hawkular_hostname}}/hawkular/metrics"
-  notify: restart master
+  notify:
+  - restart master
+  - restart master api
```
dgoodwin: Suggest dropping "HA" into these handler descriptions just for readability

roles/openshift_logging/tasks/oc_apply.yaml
ewolinetz: its to address a bug I came across while doing some quick local tests... changing the kibana hostname after deploying initially ran into the same issue we see with services

roles/openshift_metrics/tasks/update_master_config.yaml
ewolinetz: i was thinking since its currently copy/pasta, just to make it easier to rip it all out later, i'd keep the name... but i'm open to it

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

ewolinetz:
[merge]

PR #3986, opened 2017-04-24
Bump ansible rpm dependency to 2.2.2.0

sdodson:
Ansible 2.2.2.0 should be widely available in Fedora and Centos and this version contains critical fixes we depend on.

PR #4005, opened 2017-04-25
.redhat-ci.yml: also publish journal logs

jlebon:
- Update to only target v3.6.0-alpha.1 (will open another PR to target
  the newly released 1.5.0 on the 1.5 branch).
- Move test script to its own file.
- Upload journal logs on failure.

/cc @sdodson

### DIFF
--- a/.redhat-ci.sh
+++ b/.redhat-ci.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# F25 currently has 2.2.1, so install from pypi
+pip install ansible==2.2.2.0
+
+# do a simple ping to make sure the nodes are available
+ansible -vvv -i .redhat-ci.inventory nodes -a 'rpm-ostree status'
+
+upload_journals() {
+  mkdir journals
+  for node in master node1 node2; do
+    ssh ocp-$node 'journalctl --no-pager || true' > journals/ocp-$node.log
+  done
+}
+
+trap upload_journals ERR
+
+# run the actual installer
+ansible-playbook -vvv -i .redhat-ci.inventory playbooks/byo/config.yml
+
+# run a small subset of origin conformance tests to sanity
+# check the cluster NB: we run it on the master since we may
+# be in a different OSP network
+ssh ocp-master docker run --rm --net=host --privileged \
+  -v /etc/origin/master/admin.kubeconfig:/config fedora:25 sh -c \
+    '"dnf install -y origin-tests && \
+      KUBECONFIG=/config /usr/libexec/origin/extended.test --ginkgo.v=1 \
+        --ginkgo.noColor --ginkgo.focus=\"Services.*NodePort|EmptyDir\""'
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -18,28 +18,13 @@ packages:
   - openssl-devel
   - redhat-rpm-config
 
-context: 'fedora/25/atomic | origin/v1.5.0-rc.0'
+context: 'fedora/25/atomic | origin/v3.6.0-alpha.1'
 
 env:
-  OPENSHIFT_IMAGE_TAG: v1.5.0-rc.0
+  OPENSHIFT_IMAGE_TAG: v3.6.0-alpha.1
 
 tests:
-  - pip install ansible==2.2.2.0  # F25 currently has 2.2.1, so install from pypi
-  - ansible -vvv -i .redhat-ci.inventory nodes -a 'rpm-ostree status'
-  - ansible-playbook -vvv -i .redhat-ci.inventory playbooks/byo/config.yml
-  # run a small subset of origin conformance tests to sanity check the cluster
-  # NB: we run it on the master since we may be in a different OSP network
-  - ssh ocp-master docker run --rm --net=host --privileged
-    -v /etc/origin/master/admin.kubeconfig:/config fedora:25 sh -c
-    '"dnf install -y origin-tests &&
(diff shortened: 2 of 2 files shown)


sdodson:
aos-ci-test

sdodson:
[merge]

.redhat-ci.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# F25 currently has 2.2.1, so install from pypi
+pip install ansible==2.2.2.0
```
rhcarvalho: Shouldn't we use `pip install -r requirements.txt` instead?
It's going to be one more place to maintain when bumping the ansible version as we're doing in https://github.com/openshift/openshift-ansible/pull/3986.

.redhat-ci.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# do a simple ping to make sure the nodes are available
+ansible -vvv -i .redhat-ci.inventory nodes -a 'rpm-ostree status'
```
rhcarvalho: If we just want to know the hosts are available, why not use the simpler [`ping`](http://docs.ansible.com/ansible/ping_module.html) command?

.redhat-ci.sh
```diff
@@ -0,0 +1,29 @@
...
+
+upload_journals() {
+  mkdir journals
+  for node in master node1 node2; do
+    ssh ocp-$node 'journalctl --no-pager || true' > journals/ocp-$node.log
```
rhcarvalho: Why not use ansible and the inventory file here?
If we change the inventory, this code gets out of sync.

.redhat-ci.yml
```diff
@@ -18,28 +18,13 @@ packages:
...
-  - ansible -vvv -i .redhat-ci.inventory nodes -a 'rpm-ostree status'
-  - ansible-playbook -vvv -i .redhat-ci.inventory playbooks/byo/config.yml
-  # run a small subset of origin conformance tests to sanity check the cluster
-  # NB: we run it on the master since we may be in a different OSP network
-  - ssh ocp-master docker run --rm --net=host --privileged
-    -v /etc/origin/master/admin.kubeconfig:/config fedora:25 sh -c
-    '"dnf install -y origin-tests &&
-      KUBECONFIG=/config /usr/libexec/origin/extended.test --ginkgo.v=1
-      --ginkgo.noColor --ginkgo.focus=\"Services.*NodePort|EmptyDir\""'
+  - sh .redhat-ci.sh
```
rhcarvalho: Why do we set `/bin/bash` in the shebang and here call it with `sh`? Even though both will end up calling `bash` in Fedora, `sh` triggers some compatibility mode.

Suggestion: make `.redhat-ci.sh` executable, and here make the command `./.redhat-ci.sh`.

.redhat-ci.yml
```diff
@@ -18,28 +18,13 @@ packages:
   - openssl-devel
   - redhat-rpm-config
 
-context: 'fedora/25/atomic | origin/v1.5.0-rc.0'
+context: 'fedora/25/atomic | origin/v3.6.0-alpha.1'
 
 env:
-  OPENSHIFT_IMAGE_TAG: v1.5.0-rc.0
+  OPENSHIFT_IMAGE_TAG: v3.6.0-alpha.1
```
rhcarvalho: Sneaky change ;-)

.redhat-ci.sh
jlebon: Ahh good catch. Didn't notice the `requirements.txt` in the repo. Will open a PR to change that.

.redhat-ci.sh
jlebon: Well, it's both a ping as well as getting information about the hosts to know what they're running, which canonically on AH is `rpm-ostree status`. I'll update the comment to make this more obvious.

.redhat-ci.sh
jlebon: We could, though this is nice and concise. The `ocp-$node` hostnames are set by redhat-ci and are independent from the inventory.

.redhat-ci.yml
jlebon: Heh, I mention it in the first bullet point of the commit message. Though yeah, maybe it should have been a separate commit.

.redhat-ci.yml
jlebon: It's just a habit, I find `sh<space>` easier to type than `./` and it has the advantage of handling non-executable script files (though `.redhat-ci.sh` is already executable).

> Even though both will end up calling bash in Fedora, sh triggers some compatibility mode.

True, but is that actually an issue? It's essentially a short path to a `fork` & `exec`, right?

jlebon:
> How have we tested that the journal logs are actually published on error, have we forced a failure run in this PR?

Yup, I tested it manually before submitting the PR. The other comments are addressed in https://github.com/openshift/openshift-ansible/pull/4011.

jlebon:
And the `release-1.5` equivalent: https://github.com/openshift/openshift-ansible/pull/4012.

.redhat-ci.sh
rhcarvalho: Thanks, makes sense. I just learned something :-)

.redhat-ci.yml
rhcarvalho: > It's essentially a short path to a fork & exec, right?

Not according to `man bash`.
I don't know by heart the differences, but I usually prefer to play safe. Under the BUGS section of the man page:

> There are some subtle differences between bash and traditional versions of sh, mostly
> because of the POSIX specification.

Issue #4010, opened 2017-04-26
Installer installs obsolete packages of cockpit


jkaurredhat:
Installer installs obsolete packages of cockpit and gives below error : 
`````
rhel-7-server-satellite-tools-6.2-rpms                                                                                                                                                      | 2.1 kB  00:00:00
Package cockpit-shell is obsoleted by cockpit-system, trying to install cockpit-system-131-3.el7.noarch instead
Resolving Dependencies
--> Running transaction check
---> Package cockpit-bridge.x86_64 0:131-3.el7 will be installed
---> Package cockpit-docker.x86_64 0:131-3.el7 will be installed
---> Package cockpit-kubernetes.x86_64 0:135-4.el7 will be installed
---> Package cockpit-system.noarch 0:131-3.el7 will be installed
---> Package cockpit-ws.x86_64 0:131-3.el7 will be installed
--> Finished Dependency Resolution

Dependencies Resolved
--
---
Transaction check error:
  file /usr/libexec/cockpit-stub conflicts between attempted installs of cockpit-ws-131-3.el7.x86_64 and cockpit-kubernetes-135-4.el7.x86_64
`````
This is observed in 3.5 with below packages : 
```
openshift-ansible-3.5.53-1.git.0.8ade9f2.el7.noarch      
atomic-openshift-utils-3.5.53-1.git.0.8ade9f2.el7.noarch 
```

PR #4013, opened 2017-04-26
Fixes issue #4010

jkaurredhat:
Installer fails to install cockpit  due to obsolete package  'cockpit-shell'. This patch should fix the issue.l

### DIFF
--- a/roles/cockpit/tasks/main.yml
+++ b/roles/cockpit/tasks/main.yml
@@ -3,7 +3,7 @@
   action: "{{ ansible_pkg_mgr }} name={{ item }} state=present"
   with_items:
     - cockpit-ws
-    - cockpit-shell
+    - cockpit-system
     - cockpit-bridge
     - cockpit-docker
     - "{{ cockpit_plugins }}"


mtnbikenc:
In reviewing package dependencies, it appears that `cockpit` requires `cockpit-ws`, `cockpit-system` and `cockpit-bridge`.  It may be better to replace those three items with the one `cockpit`.  Also, `"{{ cockpit_plugins }}"` defaults to `cockpit-kubernetes` which still has a dependency on `cockpit-shell`, so we will still see the "obsoleted" message when that package is installed.

jkaurredhat:
cocpit-kubernetes is available in newer versions and it seems like there is no more dependency on cockpit-shell. Please see the below output : 
````
 yum install cockpit-kubernetes
Loaded plugins: product-id, search-disabled-repos, subscription-manager
Resolving Dependencies
--> Running transaction check
---> Package cockpit-kubernetes.x86_64 0:131-3.el7 will be updated
---> Package cockpit-kubernetes.x86_64 0:135-4.el7 will be an update
--> Finished Dependency Resolution

Dependencies Resolved

 Package              Arch     Version        Repository                    Size
Updating:
 cockpit-kubernetes   x86_64   135-4.el7      rhel-7-server-ose-3.4-rpms   3.3 M
`````

In one of my other systems I was able to install it and only below packages were installed : 
```
rpm -qa | grep cockpit
cockpit-system-135-4.el7.noarch
cockpit-bridge-135-4.el7.x86_64
cockpit-ws-135-4.el7.x86_64
cockpit-kubernetes-135-4.el7.x86_64
cockpit-135-4.el7.x86_64
````

sdodson:
@dperpeet Can you confirm that in order to install the cockpit web interface and cockpit-kubernetes the only packages we should install are cockpit and cockpit-kubernetes, the rest will come via dependencies?

michaelgugino:
Closing and re-opening to trigger new CI tests.

vrutkovs:

PR #4020, opened 2017-04-26
Adding defaults for es_indices

ewolinetz:
Resolves issue being seen where `es_indices` was not being set with the above `set_fact`

cc @mazzystr


mazzystr:
Thanks!

PR #4024, opened 2017-04-27
Place nuage master role in openshift-master playbooks

rparulek:
Place nuage master role in openshift-master playbooks


rparulek:
This was against released branch 1.5

Issue #4023, opened 2017-04-27
Check available docker images and rpms versions


ingvagabund:
During the containerized deployment the docker images version and rpm versions can be out of sync. We should check both versions and stop the deployment if they are not the same at the very beginning of the installation.

sdodson:
Is this a general problem or just a problem that the CI jobs were having earlier today?

The CD team builds RPMS, creates a private yum repo, then builds images from that repo. Once that's done they push those images to their final destination and publish the yum repo at roughly the same time.

The CI jobs earlier today were failing because we were pulling directly from the repo where they build their images so I've changed it to point to where the images get published. Hopefully that will fix the CI jobs we've been having trouble with. It has at least for the problems we saw earlier today.

ingvagabund:
AFAIK, CI problem only. The purpose of the issue is to just report the OCP version of the docker images and the excluder rpm. So we can see if they are different or not. Followed by a pretty message that says what is really going on instead of excluders not being available. Checking the flow it is not a trivial problem to check the version at the beginning if the deployment is containerized. So I would just report the message right before the excluders are installed. The role itself fails if the excluders are not available anyway.

jhcook:
Hi there. Why does this fail a check if this is a CI issue only? I've downgraded the excluder and rerun the pre_upgrade before upgrade. The excluder is upgraded during the pre_upgrade and the checks fail all over again. 

The automated documentation states this is a dry run. I am trying to go from 3.4.1.7 to 3.4.1.44 with zero success. The excluder is being updated to the latest RPM which supersedes the Docker image version and the update bombs.

ingvagabund:
Hi, which pre_upgrade playbook did you run?

jhcook:
openshift-ansible/playbooks/byo/openshift-cluster/upgrades/v3_4/upgrade.yml

ingvagabund:
If the rpm version and docker image version differs, you can set the following variables explicitly:
```
openshift_pkg_version=-3.4.1.44
openshift_version=3.4...
```

The `openshift_pkg_version` corresponds to available rpm version and `openshift_version` to the docker one.

PR #4033, opened 2017-04-28
[RFE] report and fail if rpm version differs from docker image version

ingvagabund:
Fixes: #4023 

E.g.
```yaml
TASK [openshift_version : Report rpm version and docker image version] *********
ok: [10.8.172.211] => {}

MSG:

openshift_rpm_version: 3.5.5.10, openshift_containerized_version: 3.5.5.9
```

### DIFF
--- a/roles/openshift_version/tasks/main.yml
+++ b/roles/openshift_version/tasks/main.yml
@@ -86,8 +86,16 @@
   include: set_version_rpm.yml
   when: not is_containerized | bool
 
-- name: Set openshift_version for containerized installation
-  include: set_version_containerized.yml
+- block:
+  - name: Set openshift_version for containerized installation
+    include: set_version_containerized.yml
+  - name: Determine openshift rpm version
+    include: rpm_version.yml
+  - name: Fail if rpm version and docker image version are different
+    fail:
+      msg: "OCP rpm version {{ openshift_rpm_version }} is different from OCP image version {{ openshift_version }}"
+    # Both versions have the same string representation
+    when: openshift_rpm_version != openshift_version
   when: is_containerized | bool
 
 # Warn if the user has provided an openshift_image_tag but is not doing a containerized install
--- a/roles/openshift_version/tasks/rpm_version.yml
+++ b/roles/openshift_version/tasks/rpm_version.yml
@@ -1,0 +1,44 @@
+---
+# input_variables:
+# - repoquery_cmd
+# - openshift.common.service_type
+# output_variables:
+# - openshift_rpm_version
+
+# if {{ openshift.common.service_type}}-excluder is enabled,
+# the repoquery for {{ openshift.common.service_type}} will not work.
+# Thus, create a temporary yum,conf file where exclude= is set to an empty list
+- name: Create temporary yum.conf file
+  command: mktemp -d /tmp/yum.conf.XXXXXX
+  register: yum_conf_temp_file_result
+
+- set_fact:
+    yum_conf_temp_file: "{{yum_conf_temp_file_result.stdout}}/yum.conf"
+
+- name: Copy yum.conf into the temporary file
+  copy:
+    src: /etc/yum.conf
+    dest: "{{ yum_conf_temp_file }}"
+    remote_src: True
+
+- name: Clear the exclude= list in the temporary yum.conf
+  lineinfile:
+    # since ansible 2.3 s/dest/path
+    dest: "{{ yum_conf_temp_file }}"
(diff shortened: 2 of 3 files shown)


ingvagabund:
aos-ci-test

ingvagabund:
```yaml
TASK [openshift_version : Report rpm version and docker image version] *********
ok: [10.8.170.141] => {
    "msg": "openshift_rpm_version: 3.6.50, openshift_containerized_version: 3.6.52"
}
ok: [10.8.170.15] => {
    "msg": "openshift_rpm_version: 3.6.50, openshift_containerized_version: 3.6.52"
}
```

ingvagabund:
aos-ci-test

roles/openshift_version/tasks/main.yml
```diff
@@ -86,8 +86,14 @@
...
 
-- name: Set openshift_version for containerized installation
-  include: set_version_containerized.yml
+- block:
+  - name: Set openshift_version for containerized installation
+    include: set_version_containerized.yml
+  - name: Determine openshift rpm version
+    include: rpm_version.yml
+  - name: Report rpm version and docker image version
+    debug:
```
mtnbikenc: The use of `openshift_containerized_version` makes it look like that should be a var name somewhere, when it is actually not.  Would it also make sense to fail/assert here if they are not the same?

roles/openshift_version/tasks/main.yml
```diff
@@ -86,8 +86,14 @@
...
+      msg: "openshift_rpm_version: {{ openshift_rpm_version }}, openshift_containerized_version: {{ openshift_version }}"
```
mtnbikenc: The use of `openshift_containerized_version` makes it look like that should be a var name somewhere, when it is actually not.  Would it also make sense to fail/assert here if they are not the same?

ingvagabund: > The use of `openshift_containerized_version` makes it look like that should be a var name somewhere, when it is actually not.

Maybe `OCP rpm version: ..., OCP image version: ...`?

> Would it also make sense to fail/assert here if they are not the same?

This is relevant only when the excluders are enabled. So something like this could work:
```yaml
- name: Fail if rpm version != image version
  fail:
     msg: "Bla bla bla"
  when:
  - openshift_rpm_version != openshift_version
  - (enable_docker_excluder | default(enable_excluders) | bool) or (enable_openshift_excluder | default(enable_excluders) | bool)
```

However, the enable_excluder logic is duplicated.

ingvagabund: IMHO, both versions should equal no matter if excluders are enabled or not.

ingvagabund:
> I see there is output noting the versions but nothing is done to fail if there is a problem. Assuming this is the expected result 👍
>
>If something is supposed to happen if they do not equal then 👎 😆

Given the deployment can be partially containerized and partly rpm based, one can end up with containerized nodes and rpm based master. In which case I would like both master and node components be of the same version. So it is not just excluder-specific issue.

ashcrow:
@ingvagabund so it works as expected, yes?

ingvagabund:
aos-ci-test

ingvagabund:
@ashcrow now it is

roles/openshift_version/tasks/main.yml
```diff
@@ -86,8 +86,16 @@
...
   when: not is_containerized | bool
 
-- name: Set openshift_version for containerized installation
-  include: set_version_containerized.yml
+- block:
+  - name: Set openshift_version for containerized installation
+    include: set_version_containerized.yml
+  - name: Determine openshift rpm version
+    include: rpm_version.yml
+  - name: Fail if rpm version and docker image version are different
```
ashcrow: We've been moving towards ``assert`` rather than ``fail``, but I won't block on it.

ingvagabund: I see the `main.yml` has additional 5 occurrences of the `fail`. So I guess this should be fixed with one huge `replace-fail-with-assert` PR.

ingvagabund:
Busted!!!
```yaml
TASK [openshift_version : Fail if rpm version and docker image version are different] ***
fatal: [10.8.170.32]: FAILED! => {"changed": false, "failed": true, "msg": "OCP rpm version 3.5.5.13 is different from OCP image version 3.5.5.14"}
```

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
[merge]

PR #4035, opened 2017-04-28
bug 1439451. Fix kibana-auth-proxy inability to find secrets

jcantrill:
This will most likely require backport to 1.5 though I dont know if a 1.5 image exists yet that has the corresponding change to the openshift-auth-proxy where this was introduced

### DIFF
--- a/roles/openshift_logging/templates/kibana.j2
+++ b/roles/openshift_logging/templates/kibana.j2
@@ -103,6 +103,22 @@ spec:
             -
              name: "OAP_DEBUG"
              value: "{{openshift_logging_kibana_proxy_debug}}"
+            -
+             name: "OAP_OAUTH_SECRET_FILE"
+             value: "/secret/oauth-secret"
+            -
+             name: "OAP_SERVER_CERT_FILE"
+             value: "/secret/server-cert"
+            -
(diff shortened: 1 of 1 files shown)


ewolinetz:
Are we sure that https://github.com/openshift/openshift-ansible/commit/af7c4d3dda33783ddcc347ef099d3648c7bf7cbe didn't resolve the issue? The bug was opened on 3.6.15-1 and that commit was merged in 3 days ago

richm:
lgtm, modulo @ewolinetz 's comment

jcantrill:
@richm @ewolinetz  This PR is needed to fix kibana memory limits PR that i have for 1.4 and 1.5.   Looking at the attachments to the issue: I do not see those secrets in the configmap; ansible install issues?  Also, I'm not sure how after review you would get this issue given 1.6 does not pull in my change to the auth proxy where I introduced this problem.

Probably should hold on doing anything until we understand why the secrets are not in the map. @ewolinetz  any thoughts here?

ewolinetz:
The original bz you reference has an issue regarding the kibana proxy crash looping because it couldn't find `secret/server-tls.json` which should be resolved by the commit I pasted above.

I'm not sure what issue you are referring to regarding memory limits

PR #4039, opened 2017-04-28
bug 1441369. Kibana memory limits

jcantrill:
Needs to merge with https://github.com/openshift/origin-aggregated-logging/pull/396

### DIFF
--- a/roles/openshift_logging/templates/kibana.j2
+++ b/roles/openshift_logging/templates/kibana.j2
@@ -44,15 +44,19 @@ spec:
 {% if kibana_cpu_limit is not none %}
               cpu: "{{kibana_cpu_limit}}"
 {% endif %}
-{% if kibana_memory_limit is not none %}
-              memory: "{{kibana_memory_limit}}"
-{% endif %}
+              memory: "{{kibana_memory_limit | default('736Mi') }}"
 {% endif %}
           env:
             - name: "ES_HOST"
               value: "{{es_host}}"
             - name: "ES_PORT"
               value: "{{es_port}}"
+            -
+              name: "KIBANA_MEMORY_LIMIT"
+              valueFrom:
+                resourceFieldRef:
+                  containerName: kibana
+                  resource: limits.memory
           volumeMounts:
             - name: kibana
               mountPath: /etc/kibana/keys
@@ -67,9 +71,7 @@ spec:
 {% if kibana_proxy_cpu_limit is not none %}
               cpu: "{{kibana_proxy_cpu_limit}}"
 {% endif %}
-{% if kibana_proxy_memory_limit is not none %}
-              memory: "{{kibana_proxy_memory_limit}}"
-{% endif %}
+              memory: "{{kibana_proxy_memory_limit | default('96Mi') }}"
 {% endif %}
           ports:
             -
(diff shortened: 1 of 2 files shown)


jcantrill:
Tested locally to confirm it works.

sdodson:
master branch too?

roles/openshift_logging/templates/kibana.j2
```diff
@@ -53,6 +53,14 @@ spec:
               value: "{{es_host}}"
             - name: "ES_PORT"
               value: "{{es_port}}"
+{% if kibana_memory_limit is defined and kibana_memory_limit is not none %}
```
sdodson: since we have defaults is this really conditional?

jcantrill: doesnt have to be

jcantrill:
@sdodson this will be needed in master if i have not already added it

roles/openshift_logging/templates/kibana.j2
```diff
@@ -44,15 +44,19 @@ spec:
 {% if kibana_cpu_limit is not none %}
               cpu: "{{kibana_cpu_limit}}"
 {% endif %}
-{% if kibana_memory_limit is not none %}
```
ewolinetz: was this intentionally removed?

roles/openshift_logging/templates/kibana.j2
```diff
@@ -67,10 +71,8 @@ spec:
 {% if kibana_proxy_cpu_limit is not none %}
               cpu: "{{kibana_proxy_cpu_limit}}"
 {% endif %}
-{% if kibana_proxy_memory_limit is not none %}
```
ewolinetz: Same for here

roles/openshift_logging/templates/kibana.j2
```diff
@@ -44,15 +44,19 @@ spec:
 {% if kibana_cpu_limit is not none %}
               cpu: "{{kibana_cpu_limit}}"
 {% endif %}
-{% if kibana_memory_limit is not none %}
               memory: "{{kibana_memory_limit}}"
```
ewolinetz: Should we also have a default value specified here since we don't have one when this template is used (e.g. what if a user specifies the value as null)

roles/openshift_logging/templates/kibana.j2
```diff
@@ -67,10 +71,8 @@ spec:
 {% if kibana_proxy_cpu_limit is not none %}
               cpu: "{{kibana_proxy_cpu_limit}}"
 {% endif %}
-{% if kibana_proxy_memory_limit is not none %}
               memory: "{{kibana_proxy_memory_limit}}"
```
ewolinetz: Should we also have a default value specified here since we don't have one when this template is used (e.g. what if a user specifies the value as null)

roles/openshift_logging/templates/kibana.j2
jcantrill: yes.  but i probably didnt read it fully.

roles/openshift_logging/templates/kibana.j2
jcantrill: ill default it to the values in the efaults

jcantrill:
@ewolinetz @sdodson   updated again... lets get this merged!

roles/openshift_logging/templates/kibana.j2
sdodson: @ewolinetz I'd suggested it under the assumption that since it had changed from null to an actual default it was no longer necessary because it would always have a value. I'm fine with the more recent changes too.

PR #4044, opened 2017-04-28
Move clean repo cache from upgrade to repos role

mtnbikenc:
The clean repo cache command is moved from the upgrade init.yml to the
openshift_repos role to remove duplication.  The role previously would
only run the 'clean all' command when the role changed repo files,
however there are situation where manually changed repo files can also
cause issues.  The task is run regardless to ensure a clean cache.

### DIFF
--- a/roles/openshift_repos/tasks/main.yaml
+++ b/roles/openshift_repos/tasks/main.yaml
@@ -40,4 +40,21 @@
     - openshift_deployment_type == 'origin'
     - openshift_enable_origin_repo | default(true) | bool
 
+  # Singleton block
+  - when: r_osr_first_run | default(true)
+    block:
+    - name: Ensure clean repo cache in the event repos have been changed manually
+      debug:
+        msg: "First run of openshift_repos"
+      changed_when: true
+      notify: refresh cache
+
+    - name: Set fact r_osr_first_run false
+      set_fact:
+        r_osr_first_run: false
+
+  # Force running ALL handlers now, because we expect repo cache to be cleared
+  # if changes have been made.
+  - meta: flush_handlers
+
   when: not ostree_booted.stat.exists
--- a/playbooks/common/openshift-cluster/upgrades/init.yml
+++ b/playbooks/common/openshift-cluster/upgrades/init.yml
@@ -10,17 +10,6 @@
 
 - include: ../initialize_facts.yml
 
-- name: Ensure clean repo cache in the event repos have been changed manually
-  hosts: oo_all_hosts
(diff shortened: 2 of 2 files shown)


sdodson:
I like the intention, but given how frequently we call openshift_repos this is going to be costly. Maybe add a fact to record that this has happened once already and avoid doing it again?

As an aside, of the two failed CI/CD upgrade jobs on Friday, the first actually ran `yum clean all` and then the one that ran a few hours later aborted before that point so it wasn't run. Both suffered the failure to download yum repodata though.

mtnbikenc:
Good point.  byo/config.yml calls `openshift_repos` 39 times...

mtnbikenc:
aos-ci-test

mtnbikenc:
@sdodson I changed this PR to have a singlton block for triggering the clean cache handler on the first run.  This also adds the flush_handlers meta task to make sure it is done before leaving the role.

roles/openshift_repos/tasks/main.yaml
```diff
@@ -40,4 +40,21 @@
     - openshift_deployment_type == 'origin'
     - openshift_enable_origin_repo | default(true) | bool
 
+  # Singleton block
+  - when: r_osr_first_run | default(true)
+    block:
+    - name: Ensure clean repo cache in the event repos have been changed manually
```
sdodson: Should this even be a handler or just a task?

mtnbikenc: I left it as a handler because the previous tasks will trigger it only if they change something.  I'm guessing there is a reason why we call openshift_repos 39 times, so I was not going to just make it not do anything on subsequent runs.  Maybe we need to step back and understand why it is being called so many times in the first place.  The original intent here was to clear the cache at least the first time in case there were manual changes just prior to the playbook being run.

sdodson: Ahh, that makes sense. Trigger it whenever they're modified or on first run.

sdodson:
[merge]

mtnbikenc:
https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible/101/
Flake: https://github.com/openshift/origin/issues/14048 idling and unidling

mtnbikenc:
https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/334/
Flake https://github.com/openshift/origin/issues/13977 Daemon set

sdodson:
[merge]

PR #4047, opened 2017-05-01
[3.5] Bump ansible rpm dependency to 2.2.2.0

sdodson:
Remove check for unpatched 2.2.1.0

Backports #3986

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ you are not running a stable release.
     ***
 
     Requirements:
-    - Ansible >= 2.2.0
+    - Ansible >= 2.2.2.0
     - Jinja >= 2.7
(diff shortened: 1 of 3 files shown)


dgoodwin:
Is 2.2.2.0 going to become a requirement for next minor release of 3.5? User in #openshift-dev today uses these github releases but the package isn't avail yet for RHEL.

sdodson:
@dgoodwin Yes, 2.2.3.0 is shipping with the next errata, 2.2.2.0 is sufficient but the repo will go from 2.2.1.0-2.el7 to 2.2.3.0.

dgoodwin:
Thanks! User is set now I think just using older tags for time being.

PR #4049, opened 2017-05-01
System container docker

ashcrow:
This change allows for the use of either the traditional package install
of docker OR a system container install of docker.

Two new inventory options, ```openshift_docker_use_system_container``` and
```openshift_docker_systemcontainer_image_registry_override```, have been
added which are be used to install with a system container. By default
this option is commented out.

One new fact has been added:
- docker.service_name: docker by default, container-engine-docker for system container

### CODE REVIEW

**Comment by ashcrow** [comment]
@tbielawa / @abutcher ready for an initial review.

**Comment by abutcher** [comment]
@ashcrow PR looks great to me aside from the temporary `docker_image` location :+1:

File: roles/docker/tasks/systemcontainer_docker.yml [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+# Configure container-enginer using the daemon.json file
+- name: Configure Container Engine
+  template:
+    dest: "{{ docker_conf }}/daemon.json"
+    src: daemon.json
+
+# Enable and start the container-engine service
+- name: Start the Container Enginer service
+  systemd:
+    name: container-engine
```
ganhuang: The name is not `"{{ openshift.docker.service_name }}"`?

roles/docker/tasks/systemcontainer_docker.yml
```diff
@@ -0,0 +1,77 @@
...
+- name: Ensure /etc/docker directory exists
+  file:
+    path: "{{ docker_conf_dir }}"
+    state: directory
+
+- name: Download Container Enginer System Container
+  command: atomic pull --storage ostree {{ openshift.common.docker_image }}:{{ openshift_image_tag }}
+
+- name: Install Container Enginer System Container
+  command: atomic install --system --system-package no {{ openshift.common.docker_image }}:{{ openshift_image_tag }}
```
giuseppe: can we use the `oc_atomic_container` module here?  It will also manage updates.

for example for openvswitch we have now:
```
- name: Install or Update OpenVSwitch system container
  oc_atomic_container:
    name: openvswitch
    image: "{{ openshift.common.system_images_registry }}/{{ openshift.node.ovs_system_image }}:{{ openshift_image_tag }}"
    state: latest
```

roles/docker/tasks/systemcontainer_docker.yml
```diff
@@ -0,0 +1,77 @@
...
+    path: "{{ docker_systemd_dir }}"
+    state: directory
+
+- name: Ensure /etc/docker directory exists
+  file:
+    path: "{{ docker_conf_dir }}"
+    state: directory
+
+- name: Download Container Enginer System Container
+  command: atomic pull --storage ostree {{ openshift.common.docker_image }}:{{ openshift_image_tag }}
```
giuseppe: I'd suggest to move this block before `Disable Docker` and rename it to `Pre-pull Container Enginer System Container image`.  It is the same pattern used for other containers

ganhuang: I suspect we need set `http_proxy` and `https_proxy` in `/etc/atomic.conf` to access the container registry in a proxy installation.

roles/docker/tasks/systemcontainer_docker.yml
```diff
@@ -0,0 +1,77 @@
...
+
+# Make sure docker is disabled Errors are ignored as docker may not
+# be installed.
+- name: Disable Docker
+  systemd:
+    name: docker
+    enabled: no
+    state: stopped
+    daemon_reload: yes
+  ignore_errors: True
```
giuseppe: should we add a check for doing this only when the running version is not the system container?  Otherwise it will stop/disable the running system container the second time we run the installation

roles/docker/tasks/systemcontainer_docker.yml
ashcrow: Ohh, yes. Good idea!

ashcrow: @giuseppe am I using ``values`` correctly:

```yaml
- name: Install Container Enginer System Container
  oc_atomic_container:
    name: container-engine-docker
    image: "{{ openshift.common.docker_image }}:{{ openshift_image_tag }}"
    state: latest
    values:
      - "--system-package no"
```

roles/docker/tasks/systemcontainer_docker.yml
ashcrow: The system container version will be called ``container-engine-docker``. I don't think there will be a time we use the same name based on feedback Monday.

roles/docker/tasks/systemcontainer_docker.yml
```diff
@@ -0,0 +1,106 @@
...
+- name: Set http(s) proxies on atomic host
+  block:
+
+    - name: Add http_proxy to /etc/atomic.conf
+      lineinfile:
+        path: /etc/atomic.conf
+        line: "http_proxy={{ openshift.common.http_proxy | default('') }}"
+      when:
+        - openshift.common.http_proxy is defined
+        - openshift.common.http_proxy != ''
```
giuseppe: should this be done in any case (without `when` block) so that openshift.common.http_proxy can be unset as well?

roles/docker/tasks/systemcontainer_docker.yml
```diff
@@ -0,0 +1,106 @@
...
+
+    - name: Add https_proxy to /etc/atomic.conf
+      lineinfile:
+        path: /etc/atomic.conf
+        line: "https_proxy={{ openshift.common.https_proxy | default('') }}"
+      when:
+        - openshift.common.https_proxy is defined
+        - openshift.common.https_proxy != ''
```
giuseppe: same here

roles/docker/tasks/systemcontainer_docker.yml
giuseppe: should be:

```yaml


- name: Install Container Enginer System Container
  oc_atomic_container:
    name: container-engine-docker
    image: "{{ openshift.common.docker_image }}:{{ openshift_image_tag }}"
    state: latest
    values:
      - system-package=no
```

roles/docker/tasks/systemcontainer_docker.yml
ashcrow: I don't believe so. We are assuming it has not been set and are adding the line to the file. In the future we may want to lay down a templetized ``atomic.conf`` so we control all the variables (or do something in ``/etc/atomic.d/`` if that's a possibility. WDYT?

giuseppe: I am fine with it, just wondering what happens if you set the proxy run the installer, unset it and run the installer again.  Wouldn't you still use the old setting for the proxy even if the installer has not such setting?  I have not checked it but atomic should be able to handle an empty line.

ashcrow: If you set the proxy with the installer, remove the proxy and run the installer again it will add the proxy line.  If you set the proxy manually then run the installer it will have 2 proxy lines and could be problematic. However, I don't believe there should be a case where an admin goes and manually configured the proxy and then expects the installer to to remove it and replace it with it's own.

Though I do agree, I'd like a more solid solution later on if possible.

ashcrow:
I don't believe this is related to my change as the actual install case should have followed the original pattern.

```
FAILED - RETRYING: TASK: openshift_node : Start and enable node (1 retries left).Result was: {
    "attempts": 1, 
    "failed": true, 
    "invocation": {
        "module_args": {
            "daemon_reload": false, 
            "enabled": true, 
            "masked": null, 
            "name": "origin-node", 
            "state": "started", 
            "user": false
        }, 
        "module_name": "systemd"
    }, 
    "retries": 2
```

ashcrow:
Adding @dgoodwin to the review list for good measure 😺

ashcrow:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
I had to push some changes to CI jobs and the previous job.

ashcrow:
Updated based on @giuseppe's ideas in IRC.

ashcrow:
aos-ci-test

roles/docker/tasks/systemcontainer_docker.yml
```diff
@@ -0,0 +1,126 @@
...
+        l_docker_image_prepend: "registry.fedoraproject.org/"
+      when: ansible_distribution == 'Fedora'
+
+    - name: Set the full image name
+      set_fact:
+        l_docker_image: "{{ l_docker_image_prepend }}container-engine-docker:latest"
+
+
+- name: Pre-pull Container Enginer System Container image
+  command: "atomic pull --storage ostree {{ l_docker_image }}"
```
dgoodwin: Anything we can do here to not have this always report "changed"?

roles/docker/tasks/systemcontainer_docker.yml
```diff
@@ -0,0 +1,126 @@
...
+- name: Ensure /etc/docker directory exists
+  file:
+    path: "{{ docker_conf_dir }}"
+    state: directory
+
+- name: Install Container Enginer System Container
+  oc_atomic_container:
+    name: container-engine-docker
+    image: "container-engine-docker:latest"
+    state: latest
```
dgoodwin: Is latest redundant in one of these lines?

roles/docker/tasks/systemcontainer_docker.yml
ashcrow: I could add ``change_when: false``. That what you were thinking?

roles/docker/tasks/systemcontainer_docker.yml
ashcrow: Will update

roles/docker/tasks/systemcontainer_docker.yml
dgoodwin: At minimum, do we care about accurately reporting if the pull changed anything?

ashcrow:
@dgoodwin updated ⬆️

roles/docker/tasks/systemcontainer_docker.yml
ashcrow: At least right now, we don't.

dgoodwin: If an operations group was using this role for ongoing config management of their systems, and here we pull down a new docker sys container image, should that not show as a change?

Or would that show as a change elsewhere later in the play?

ashcrow:
aos-ci-test

roles/docker/tasks/systemcontainer_docker.yml
ashcrow: In another release we are planning on using the ```openshift_image_tag`` but for now ``latest`` was the best version we could use.

ashcrow: We'll probably also use https://github.com/openshift/openshift-ansible/pull/3109 as a location for reporting changed in the future as well.

ashcrow:
Adding @sdodson. I need to get at least one 👍 or 👎 review 😆

roles/docker/tasks/systemcontainer_docker.yml
```diff
@@ -0,0 +1,126 @@
...
+  package:
+    name: atomic
+    state: present
+
+# At the time of writing the atomic command requires runc for it's own use. This
+# task is here in the even that the atomic package ever removes the dependency.
+- name: Ensure runc is installed
+  package:
+    name: runc
+    state: present
```
sdodson: when: not openshift.common.is_atomic for these package invocations, weird weird errors if you don't and this runs on atomic host.

roles/docker/tasks/systemcontainer_docker.yml
sdodson: Unfortunately we don't handle set, unset very well currently elsewhere. It'd be nice to revisit in the future but I'm fine with this as written.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -59,11 +59,12 @@ def migrate_docker_facts(facts):
...
             'insecure_registries',
             'blocked_registries',
-            'options'
+            'options',
+            'use_system_container',
         ),
         'node': (
             'log_driver',
-            'log_options'
+            'log_options',
```
sdodson: is this correcting an existing lint error or unintended?

roles/docker/tasks/systemcontainer_docker.yml
ashcrow: Can't believe I didn't think about that!

roles/openshift_facts/library/openshift_facts.py
ashcrow: Leftover from when I updated the section but removed since then. I'll remove it. It's aesthetics only.

ashcrow:
@sdodson squashed and updated. ⬆️

ashcrow:
aos-ci-test

ashcrow:
bot, retest this please

roles/docker/tasks/systemcontainer_docker.yml
```diff
@@ -0,0 +1,128 @@
...
+      when: ansible_distribution == 'RedHat'
+
+    - name: Use Fedora Registry for image when distribution is Fedora
+      set_fact:
+        l_docker_image_prepend: "registry.fedoraproject.org/"
+      when: ansible_distribution == 'Fedora'
+
+    - name: Set the full image name
+      set_fact:
+        l_docker_image: "{{ l_docker_image_prepend }}container-engine-docker:latest"
```
ganhuang: In case that we need to test in a internal registry or not able to access the public network at all, is it possible to make `l_docker_image_prepend` customizable?

ashcrow: @ganhuang I'll throw something together this morning so you can override the image.

ashcrow:
Nice: ```Internal Jenkins error during provisioning the jenkins slave.```

roles/docker/tasks/systemcontainer_docker.yml
ashcrow: Added ```openshift_docker_systemcontainer_image_registry_override``` in inventory.

ashcrow:
aos-ci-test

ashcrow:
```fatal: [10.8.168.9]: FAILED! => {"changed": false, "failed": true, "msg": "No package matching 'atomic-openshift-docker-excluder-3.6.63*' found available, installed or updated", "rc": 126, "results": ["No package matching 'atomic-openshift-docker-excluder-3.6.63*' found available, installed or updated"]}```

ashcrow:
aos-ci-test

ashcrow:
bot, retest this please

ashcrow:
@abutcher I think this is the best CI will give us.

abutcher:
[merge]

Issue #4046, opened 2017-05-01
Metrics installation fails when docker volume dir is immutable


mwoodson:
#### Description

When deploying metrics on an Online cluster, we are unable to create the htpasswd.  The cod that is being executed is here:

https://github.com/openshift/openshift-ansible/blob/release-1.5/roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml#L43-L48

The problem comes from Online clusters having an immutable /var/lib/docker/volumes.  This is because we do NOT want Online customers creating ephemeral volumes on the host.

The error we are seeing is this:

    "stderr": "/usr/bin/docker-current: Error response from daemon: create 189984d7bea8c4687b3a8edb8871aa7faf268cf1ac54711dff9296a0c16abc06: mkdir /var/lib/docker/volumes/189984d7bea8c4687b3a8edb8871aa7faf268cf1ac54711dff9296a0c16abc06: permission denied.\nSee '/usr/bin/docker-current run --help'.", 

##### Version

openshift 3.5
openshift-ansible-3.5.62

* Your ansible version per `ansible --version`

```
ansible 2.2
```

##### Steps To Reproduce
1. Make /var/lib/docker/volume immutable (this is done in Online)
2. Run the metrics playbooks

##### Observed Results

docker run is failing because the volume is immutable

ewolinetz:
@mwoodson this should now be resolved with #4054 being merged

ewolinetz:
The issue still exists, it is stemming from the image having a `Volume` definition within it. We should be able to switch to using the `htpasswd` module provided by Ansible

ewolinetz:
re addressing in https://github.com/openshift/openshift-ansible/pull/4063

PR #4054, opened 2017-05-01
Updating htpasswd logic to not mount a volume for creating file

ewolinetz:
Addresses #4046 

cc @mwoodson

### DIFF
--- a/roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
+++ b/roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
@@ -40,12 +40,18 @@
   - hawkular-metrics
   - hawkular-jgroups-keystore
 
-- name: generate htpasswd file for hawkular metrics
+- name: generate htpasswd content for hawkular metrics
   shell: >
-    docker run -i -a STDIN -a STDOUT -v "{{ mktemp.stdout }}"
-    "{{ openshift_metrics_httpd_image }}" htpasswd -ni hawkular
-    > '{{ mktemp.stdout }}/hawkular-metrics.htpasswd'
-    < '{{ mktemp.stdout }}/hawkular-metrics.pwd'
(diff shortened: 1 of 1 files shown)


roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
```diff
@@ -40,12 +40,19 @@
...
 
-- name: generate htpasswd file for hawkular metrics
+- name: generate htpasswd content for hawkular metrics
   shell: >
-    docker run -i -a STDIN -a STDOUT -v "{{ mktemp.stdout }}"
+    cat "{{ mktemp.stdout}}/hawkular-metrics.pwd" | cat > metrics.pwd && docker run -i -a STDIN -a STDOUT
     "{{ openshift_metrics_httpd_image }}" htpasswd -ni hawkular
-    > '{{ mktemp.stdout }}/hawkular-metrics.htpasswd'
-    < '{{ mktemp.stdout }}/hawkular-metrics.pwd'
+    < metrics.pwd
```
sdodson: Why not just the following avoiding cat from tmp to pwd?

```
docker run -i -a STDIN -a STDOUT
"{{ openshift_metrics_httpd_image }}" htpasswd -ni hawkular
< "{{ mktemp.stdout}}/hawkular-metrics.pwd"
```

If we stick with this as is do we need to clean up metrics.pwd?

ewolinetz: i'll double check that i wasn't doing something else odd when i was testing that... but IIRC it wouldn't validate doing it that way. let me confirm

ewolinetz: yeah, i must have been testing it incorrectly. it validated using the approach you pasted above

ewolinetz:
aos-ci-test

sdodson:
Tested locally, CI won't test this. Merging.

PR #4057, opened 2017-05-01
[3.5] Use openshift_ca_host's hostnames to sign the CA

sdodson:
If for some reason oo_first_master and openshift_ca_host are different
we could've signed the CA with the wrong hostnames.


sdodson:
aos-ci-test

PR #4063, opened 2017-05-01
Updating openshift_metrics to use htpasswd module instead of making a…

ewolinetz:
… docker call. Updating requirement to include python-passlib

cc @mwoodson

### DIFF
--- a/roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
+++ b/roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
@@ -33,25 +33,26 @@
   changed_when: no
 
 - name: generate password for hawkular metrics and jgroups
-  copy:
-    dest: "{{ mktemp.stdout }}/{{ item }}.pwd"
-    content: "{{ 15 | oo_random_word }}"
+  local_action: copy dest="{{ local_tmp.stdout }}/{{ item }}.pwd" content="{{ 15 | oo_random_word }}"
   with_items:
   - hawkular-metrics
   - hawkular-jgroups-keystore
 
-- name: generate htpasswd content for hawkular metrics
-  shell: >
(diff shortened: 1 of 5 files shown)


ewolinetz:
the htpasswd module requires the python-passlib package. Is that the correct way to add it as a requirement?

sdodson:
Can you run this as a local action because we need to ensure that python-passlib is installed on the host and we can't do that on atomic. Honestly we're right back where we were at the beginning. We have to make sure that whatever process online uses to vendor our code ensures that rpm dependencies are met where openshift-ansible is run.

ewolinetz:
@sdodson sure, i'll give it a shot...

ewolinetz:
@sdodson updated to use a local action instead.

cc @mwringe

ewolinetz:
aos-ci-test

roles/openshift_metrics/tasks/generate_hawkular_certificates.yaml
```diff
@@ -33,26 +33,27 @@
...
   no_log: true
-  register: metrics_htpasswd
 
-- name: create htpasswd file with generated content
-  copy:
-    content: "{{ metrics_htpasswd.stdout }}"
-    dest: "{{ mktemp.stdout }}/hawkular-metrics.htpasswd"
+- name: generate htpasswd file for hawkular metrics
+  local_action: htpasswd path="{{ local_tmp.stdout }}/hawkular-metrics.htpasswd" name=hawkular password="{{ hawkular_metrics_pwd.content | b64decode }}"
   no_log: true
```
sdodson: This eats the error about passlib not being installed if it's not installed.

ewolinetz: that isn't the best... but i wanted to avoid the password being displayed if `-vvv` was used

PR #4064, opened 2017-05-01
add open vswitch version validation

juanvallejo:
Related Trello card: https://trello.com/c/ZdkKPVh9/105-ovs-version-validation-diagnostics?menu=filter&filter=[diagnostics]

Ensures versions of OpenShift >= 3.5 have ovs 2.6, and that any other versions have ovs 2.4.
Depends on the remote host having `ovs-vsctl` installed

@sosiouxme @rhcarvalho @brenton


rhcarvalho:
> Depends on the remote host having ovs-vsctl installed

Is that installed from the same package?
If not, could we use `rpm` instead? That'd be one less dependency.

roles/openshift_health_checker/meta/main.yml
```diff
@@ -2,3 +2,4 @@
 dependencies:
   - role: openshift_facts
   - role: openshift_repos
+  - role: openshift_version
```
rhcarvalho: Why do we need this dependency?

Note: missing newline char.

roles/openshift_health_checker/openshift_checks/ovs_version.py
```diff
@@ -0,0 +1,77 @@
+# pylint: disable=missing-docstring
+import json
+
+from openshift_checks import OpenShiftCheck, OpenShiftCheckException, get_var
+
+
+class OVSVersion(OpenShiftCheck):
```
rhcarvalho: Wouldn't it be easier to implement if this was part of the existing `package_version` check?

roles/openshift_health_checker/meta/main.yml
juanvallejo: I need the `openshift_image_tag` task_var to be set in order to determine which version of ovs to check for

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -52,32 +55,41 @@ def main():
...
         module.fail_json(msg=str(excinfo))
     module.exit_json(changed=False)
 
 
+def _to_dict(pkg_list):
+    d = dict()
+    for pkg in pkg_list:
+        d[pkg[LIST_PKG_NAME]] = pkg
+
+    return d
```
sosiouxme: Concise and clearer as a comprehension...

    return {pkg[LIST_PKG_NAME]: pkg for pkg in pkg_list}

roles/openshift_health_checker/meta/main.yml
sosiouxme: It's being added anyway for the docker img availability check.

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -9,6 +9,9 @@ class PackageVersion(NotContainerizedMixin, OpenShiftCheck):
     name = "package_version"
     tags = ["preflight"]
 
+    recommended_ovs_version_os_3_5 = "2.6"
+    recommended_ovs_version = "2.4"
```
rhcarvalho: This seems hard to evolve when newer releases are out.

Which makes me think how would we proceed once 3.6 is out. Are we expecting to be revisiting the checks every now and then and updating the table of version numbers?

rhcarvalho:
@juanvallejo do you mind me borrowing your PR to cleanup old comments in `aos_version.py`? :)

juanvallejo:
@rhcarvalho 
> do you mind me borrowing your PR to cleanup old comments in aos_version.py? :)

This would definitely be appreciated, thanks!

roles/openshift_health_checker/openshift_checks/package_version.py
juanvallejo: hm, I'll go ahead and turn this into a map, but not sure how future versions will be handled. cc @brenton

brenton: We can only expect the combinations of ovs and ocp to become more complicated over time.

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -9,6 +9,19 @@ class PackageVersion(NotContainerizedMixin, OpenShiftCheck):
...
+    openshift_to_ovs_version = {
+        "3.6": "2.6",
+        "3.5": "2.6",
+        "3.4": "2.4",
+    }
+
+    openshift_to_docker_version = {
+        "3.1": "1.8",
+        "3.2": "1.10",
+        "3.3": "1.10",
```
juanvallejo: @brenton I could not find an excluder script in `origin.spec` in the origin repo, however I did see that releases `1.2 - 1.5` define the [required docker version](https://github.com/openshift/origin/blob/release-1.5/origin.spec#L12) as 1.9.1. I am not sure if it is safe to assume the same docker version requirements as ose, but for now, an `origin` deployment type maps to the same required version of docker as an `enterprise` deployment type

brenton: I think in origin the excluder work was never backported so it's only on the master branch (which makes sense).  You can safely assume the same docker version for the corresponding ocp version. eg, ocp 3.2 and origin 1.2 use docker 1.10.

juanvallejo:
@sosiouxme 
> So, yesterday in/after scrum we mentioned that with the changes in subscription management we were less worried about people installing the wrong version now and more worried about detecting it already having happened. That would put this more in health check territory than an expansion of this preflight check. I think it could be good to have a separate check that can retroactively declare that what you have installed is problematic for whatever reason we discover in the future. Thoughts?

Will open a separate check with `health` tag to address this using the `rpm` module. Thanks!

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -17,9 +30,86 @@ def is_active(cls, task_vars):
...
+        return self.parse_version(openshift_version)
+
+    @staticmethod
+    def parse_version(version):
+        components = version.split(".")
+        if not components or len(components) < 2:
+            msg = "Invalid version format returned by host: {}"
+            raise OpenShiftCheckException(msg.format(version))
+
+        return float('.'.join(components[:2]))
```
sosiouxme: Why convert this to float? Just use a string everywhere (and floatify where needed for comparison).

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -27,7 +25,12 @@
 try:
     import yum  # pylint: disable=import-error
 except ImportError as err:
-    IMPORT_EXCEPTION = err  # in tox test env, yum import fails
+    IMPORT_EXCEPTION = err
+
+
+LIST_PKG_NAME = 0
+LIST_PKG_VERSION = 1
+LIST_PKG_MULTI_MINOR_CHECK = 2
```
sosiouxme: I really do not like non-named tuples as an API to this module. Could it be a list of dict instead? It would be clearer on both sides to see:

               dict(
                    name="docker",
                    version=self.get_required_docker_version(task_vars),
                    check_multi=False,
                ),

and

    p['name']

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -9,6 +9,19 @@ class PackageVersion(NotContainerizedMixin, OpenShiftCheck):
     name = "package_version"
     tags = ["preflight"]
 
+    openshift_to_ovs_version = {
+        "3.6": "2.6",
+        "3.5": "2.6",
+        "3.4": "2.4",
```
sosiouxme: Are origin versions like 1.x being translated to 3.x somewhere? I don't see where that would happen.

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -17,9 +30,86 @@ def is_active(cls, task_vars):
...
+    def get_required_ovs_version(self, task_vars):
+        """Return the correct Open vSwitch version for the current OpenShift version.
+        If the current OpenShift version is >= 3.5, ensure Open vSwitch version 2.6,
+        Else ensure Open vSwitch version 2.4"""
+        openshift_version = self.get_openshift_version(task_vars)
+
+        if openshift_version < 3.5:
+            return self.openshift_to_ovs_version["3.4"]
+
+        if not self.openshift_to_ovs_version.get(str(openshift_version)):
```
sosiouxme: nit but I'd like the logic better as:

        ovs_version = self.openshift_to_ovs_version.get(str(openshift_version))
        if ovs_version:
            return ovs_version

        msg = "There is no recommended version of Open vSwitch for the current version of OpenShift: {}"
        raise OpenShiftCheckException(msg.format(openshift_version))

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -17,9 +30,86 @@ def is_active(cls, task_vars):
...
+                ),
+            ],
         }
+
         return self.execute_module("aos_version", args, tmp, task_vars)
+
+    def get_required_ovs_version(self, task_vars):
+        """Return the correct Open vSwitch version for the current OpenShift version.
+        If the current OpenShift version is >= 3.5, ensure Open vSwitch version 2.6,
+        Else ensure Open vSwitch version 2.4"""
```
sosiouxme: So, if OCP 3.7 comes out, this breaks and we presumably fix it prior to release. That's fine. What if ovs-2.7 comes out? I don't think that's gated on openshift-ansible working, so it would cause this check to fail and have to be skipped, right? That seems brittle but I'm not sure there's a better path.

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -17,9 +30,86 @@ def is_active(cls, task_vars):
...
+        """Return the correct Docker version for the current OpenShift version.
+        If the OpenShift version is 3.1, ensure Docker version 1.8.
+        If the OpenShift version is 3.2 or 3.3, ensure Docker version 1.10.
+        If the current OpenShift version is >= 3.4, ensure Docker version 1.12."""
+        openshift_version = self.get_openshift_version(task_vars)
+
+        if openshift_version >= 3.4:
+            return self.openshift_to_docker_version["3.4"]
+
+        if not self.openshift_to_docker_version.get(str(openshift_version)):
```
sosiouxme: similarly prefer non-negating logic here

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -17,9 +30,86 @@ def is_active(cls, task_vars):
...
+        if openshift_version and openshift_version[0] == 'v':
+            openshift_version = openshift_version[1:]
+
+        return self.parse_version(openshift_version)
+
+    @staticmethod
+    def parse_version(version):
+        components = version.split(".")
+        if not components or len(components) < 2:
+            msg = "Invalid version format returned by host: {}"
```
sosiouxme: I don't see how a user would ever see this in practice, but if they did, wouldn't this message be confusing, saying the host returned a version? Maybe: "Invalid OpenShift version found for this host:"

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -27,7 +25,12 @@
 try:
     import yum  # pylint: disable=import-error
 except ImportError as err:
-    IMPORT_EXCEPTION = err  # in tox test env, yum import fails
+    IMPORT_EXCEPTION = err
```
juanvallejo: :) will re-add

roles/openshift_health_checker/library/aos_version.py
juanvallejo: I agree, will update

roles/openshift_health_checker/openshift_checks/package_version.py
juanvallejo: Thanks, handling this [here](https://github.com/openshift/openshift-ansible/pull/4064/files#diff-7a702243830dedaa3bae183335f1745aR122) now

juanvallejo:
@sosiouxme thanks for the feedback. Review comments addressed

roles/openshift_health_checker/library/aos_version.py
sosiouxme: eh, ok, no big deal.

roles/openshift_health_checker/meta/main.yml
sosiouxme: of course the integration tests will fail with this until they're fixed.

sosiouxme:
aos-ci-test

sosiouxme:
[merge]

sosiouxme:
integration tests will fail :(

juanvallejo:
aos-ci-test

sosiouxme:
aos-ci-test

sosiouxme:
[test] to get integration tests

sosiouxme:
no idea what happened with [OS_3.6_containerized](https://aos-ci.s3.amazonaws.com/openshift/openshift-ansible/jenkins-openshift-ansible-test-matrix-CONTAINERIZED=_containerized,OSE_VER=3.6,PYTHON=System-CPython-2.7,TOPOLOGY=openshift-cluster-containerized,TargetBranch=master,nodes=openshift-ansible-slave-469/a87c736edf1c2d6a5a2cea00f82604e4ca0a24e9.txt)

sosiouxme:
aos-ci-test

sdodson:
aos-ci-test

sosiouxme:
test only failed because ci tests weren't done yet. this can [merge]

sosiouxme:
merge failed by missing devcut... try again when master reopens.

sosiouxme:
[merge]

bparees:
[merge]

PR #4067, opened 2017-05-02
Add selector matchLabels storage-type support in Cassandra pvc template in openshift_metrics role

harobed:
I need to use Ceph PV for Cassandra.

Example in PVC:

```
selector:
   matchLabels:
   storage-type: {{ storage_type }}
```

With this patch, I can specify a selector matchLabels storage-type  in Cassandra pvc template in openshift_metrics role.

Best regards,
Stéphane

### DIFF
--- a/roles/openshift_metrics/templates/pvc.j2
+++ b/roles/openshift_metrics/templates/pvc.j2
@@ -25,3 +25,8 @@ spec:
   resources:
     requests:
       storage: {{size}}
+{% if storage_type is defined and storage_type %}
+  selector:
+    matchLabels:
+      storage-type: {{ storage_type }}
+{% endif %}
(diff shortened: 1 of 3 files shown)


harobed:
May be I need to go towards https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/defaults/main.yml#L90 ?

I think it's better for the consistency.

roles/openshift_metrics/templates/pvc.j2
```diff
@@ -25,3 +25,8 @@ spec:
   resources:
     requests:
       storage: {{size}}
+{% if storage_type is defined and storage_type %}
+  selector:
+    matchLabels:
+      storage-type: {{ storage_type }}
```
ewolinetz: could we make this more generic and allow the passing of a node selector? then one could provide "storage-type" as the selector key?

ewolinetz:
This should now be possible with https://github.com/openshift/openshift-ansible/pull/4302

PR #4074, opened 2017-05-02
Switch back to a static files for etcdctl.sh

sdodson:
One possible fix for https://github.com/openshift/origin/issues/14000

I first tried switching to the following but ended up with a bunch of undefined variables and it attempted to run the ca roles.

playbooks/common/openshift-commo
```
include_role:
  name: etcd
  tasks_from: etcdctl
```


ingvagabund:
https://github.com/openshift/openshift-ansible/pull/4077

PR #4077, opened 2017-05-03
move etcdctl.yml from etcd to etcd_common role

ingvagabund:
Fixes: https://github.com/openshift/origin/issues/14000


ingvagabund:
aos-ci-test

ingvagabund:
Running the 3.5 upgrade locally without any issues.

roles/etcd/tasks/main.yml
```diff
@@ -120,7 +120,9 @@
     enabled: yes
   register: start_result
 
-- include: etcdctl.yml
+- include_role:
+    name: etcd_common
+    tasks_from: etcdctl.yml
```
sdodson: Why not just make etcd_common/tasks/main.yml call etcd_common/tasks/etcdctl.yml and drop this from etcd role which already requires etcd_common via roles/etcd/meta/main.yml?

ingvagabund: Since the `etcdctl.yml` installs etcd and that should be done right after the etcd cluster is actually installed and configured. Setting `etcd_common/tasks/main.yml` -> `etcd_common/tasks/main.yml ` will first install etcd [1] and then run the etcd role.

[1] https://github.com/openshift/openshift-ansible/blob/ba5c6594ade6679d055aaecc7c38836444f0f791/roles/etcd/tasks/etcdctl.yml#L2

ingvagabund: Given
```yaml
- name: Install etcd
  package: name=etcd{{ '-' + etcd_version if etcd_version is defined else '' }} state=present
```

is in `etcd_server_certificates` as well, the etcd gets installed even before the `etcd` role itself is started. I will update the PR.

ingvagabund: For me, it is ok to keep the `etcdctl.yml` inside the `etcd_common` as it is a common task to do. However, I don't think it is a common tasks to do everytime the `etdc_common` role is run.

I would keep it the way it is. Once we have more "reasonable" code related to etcd client, we can stack it into the `etcd_client` role that installs the etcdctl command and drops the `etcdctl.sh` into the profiles.

I would like to introduce new etcd-client rpm later that installs just the etcdctl command. Once done, it makes sense to have `etcd_client` role.

sdodson:
I dislike include_role inside a role but I can come up with no better solution. I think we should remove playbooks/common/openshift-cluster/upgrades/etcd/files/etcdctl.sh since it's now a dangling symlink here too.

ingvagabund:
Just for reference, removing etcdctl.sh via https://github.com/openshift/openshift-ansible/pull/4124

PR #4079, opened 2017-05-03
health check playbooks: relocate and expand

sosiouxme:
We are moving toward having adhoc post-install checks and so the
"preflight" designation needs to be widened.
Updated location, added health check playbook, and updated README.

Opening this PR for discussion. @brenton @juanvallejo @codificat @rhcarvalho


playbooks/common/openshift-checks/health.yml
```diff
@@ -11,4 +11,4 @@
     - action: openshift_health_check
       args:
         checks:
-          - '@preflight'
+          - '@health'
```
rhcarvalho: I didn't get the idea for this file. What would match the "health" tag?

playbooks/common/openshift-checks/README.md
```diff
@@ -36,8 +28,14 @@ against your inventory file. Here is the step-by-step:
...
+
+    ```console
+    $ ansible-playbook -i <inventory file> playbooks/common/openshift-checks/pre-install.yml
+    ```
+
+    or
 
     ```console
-    $ ansible-playbook -i <inventory file> playbooks/byo/openshift-preflight/check.yml
+    $ ansible-playbook -i <inventory file> playbooks/common/openshift-checks/health.yml
```
rhcarvalho: I'm not against having playbooks under `common/` but I don't understand how our minds shifted away from `byo/`.

playbooks/common/openshift-checks/README.md
```diff
@@ -1,29 +1,21 @@
...
 run.
 
----
-
-*Note that currently the `check.yml` playbook is only useful for RPM-based
-installations. Containerized installs are excluded from checks for now, but
-might be included in the future if there is demand for that.*
-
----
-
```
rhcarvalho: :+1:

playbooks/common/openshift-checks/README.md
```diff
@@ -1,29 +1,21 @@
...
+an install, as well as health checks to run on installed systems.
 
 Ansible's default operation mode is to fail fast, on the first error. However,
 when performing checks, it is useful to gather as much information about
 problems as possible in a single run.
 
-The `check.yml` playbook runs a battery of checks against the inventory hosts
-and tells Ansible to ignore intermediate errors, thus giving a more complete
+The playbooks run a battery of checks against the inventory hosts
+and have Ansible gather intermediate errors, thus giving a more complete
```
rhcarvalho: :+1:

playbooks/common/openshift-checks/README.md
```diff
@@ -1,29 +1,21 @@
-# OpenShift preflight checks
+# OpenShift health checks
 
 Here we provide an Ansible playbook for detecting potential roadblocks prior to
-an install or upgrade.
+an install, as well as health checks to run on installed systems.
```
rhcarvalho: :+1:

playbooks/common/openshift-checks/README.md
```diff
@@ -1,29 +1,21 @@
-# OpenShift preflight checks
+# OpenShift health checks
```
rhcarvalho: :+1:

playbooks/common/openshift-checks/README.md
codificat: I was wondering the same... the health check playbooks don't seem to fully fit the others under `common/`... although they don't fully fit under `byo/` either.

Brainstorming a bit here but: what about moving to `playbooks/openshift-checks/`? and move the playbooks from `certificate_expiry` in there too?

playbooks/common/openshift-checks/health.yml
codificat: I think that the idea here is for checks like #3660 (etcd image metadata size) to match that tag... i.e. checks that are meaningful for a running cluster, like certificate expiration...

sosiouxme: Yes, exactly. It makes no sense to run most preflight checks on an ongoing basis, or to run most health checks at preflight, so just separating them out.

playbooks/common/openshift-checks/README.md
sosiouxme: Exactly the sort of conversation I opened this PR to have :) 

I had my own concept of what the directories are for but consulting the README:

- `byo` (_Bring Your Own_ hosts) has the most actively maintained
  playbooks for installing, upgrading and performing others tasks on OpenShift
  clusters.
- `common` has a set of playbooks that are included by playbooks in
  `byo` and others.
- others are basically un-maintained / deprecated

`byo` strikes me as `for install`. And specifically for installing when you already have hosts. I guess our preflight checks are a good match for that. Health checks really aren't though. I thought it might make sense to have the various check playbooks in the same place ... but maybe not?

juanvallejo:
Related Trello card: https://trello.com/c/Yk0BdYRU/92-integrate-openshift-certificate-expiry-with-openshift-health-checker-diagnostics?menu=filter&filter=[diagnostics]

playbooks/common/openshift-checks/README.md
sosiouxme: Well, in the absence of further comments, I think @codificat 's idea (`playbooks/openshift-checks/`) is a pretty good one, will tweak the PR some.

sosiouxme:
Preflight checks in `pre-install.yml` (feel free to criticize name, but I thought we might need to do something separate for upgrades and "flight" could refer to either).  Diagnostic checks in `health.yml` - for a running cluster (checks tagged with `health` which should be appearing soon). And I moved the expiry playbooks under `openshift-checks` and moved that to the top level of playbooks. What do you think of these locations?

We need to think a bit more about what "integration" of the expiry checks into the diagnostic checks would mean. Right now they're separate playbooks of course and they have nice reporting features, so I'm not sure we can just pair them with the other checks in a playbook. Frankly we should consider reporting features for all our health checks; e.g. the EFK stack check should gather a bundle of pod logs and API definitions if there is anything wrong. Anytime something is wrong, unless you can diagnose exactly what to do (and maybe even then) give the user a bundle of files for their support rep to help diagnose the problem.

codificat:
> Preflight checks in pre-install.yml (feel free to criticize name, but I
> thought we might need to do something separate for upgrades and "flight" could
> refer to either). Diagnostic checks in health.yml - for a running cluster
> (checks tagged with health which should be appearing soon). And I moved the
> expiry playbooks under openshift-checks and moved that to the top level of
> playbooks. What do you think of these locations?

They look good to me; @rhcarvalho, @juanvallejo what do you think?

> I left behind symlinks for existing playbooks. Do we think these should stay
> around for a transition period? Or we can just make a clean break and not
> leave clutter around.

Yes, I believe we do need that safety net for a bit: the [readme in the cert expiry roles](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_certificate_expiry/README.md) needs to be updated, but beyond that there are also references to this path in [the docs about certificate redeployment](https://docs.openshift.com/container-platform/latest/install_config/redeploying_certificates.html#install-config-cert-expiry-running-playbooks). These need to be updated, and by now this means we need to grant a transition period.

> We need to think a bit more about what "integration" of the expiry checks into
> the diagnostic checks would mean. Right now they're separate playbooks of
> course and they have nice reporting features, so I'm not sure we can just pair
> them with the other checks in a playbook.

Agreed!

To begin with, I'm not even sure how much sense it makes to group all the *health* (i.e. running cluster) checks together, because it's likely that each will need to be run at different times/intervals and/or with different targets. For example: the etcd health check will probably run daily, while the cert expiry will run maybe monthly. 

When/if we group them, should start tagging tasks accordingly? E.g. should we add the *@health* tag to the tasks in the certchecks playbooks? As the action-plugin based checks have their own groups/tags this can get potentially confusing, specially if they start to diverge - in the sense that *@health* or some other tag might not necessarily mean exactly the same when it's part of a task vs. when it's inside a check, right?

> Frankly we should consider reporting features for all our health checks;
> e.g. the EFK stack check should gather a bundle of pod logs and API
> definitions if there is anything wrong. Anytime something is wrong, unless you
> can diagnose exactly what to do (and maybe even then) give the user a bundle
> of files for their support rep to help diagnose the problem.

That's also a good point, certainly worth reviewing - probably on a different PR though.

rhcarvalho:
> What do you think of these locations?

👍 to renaming preflight → pre-install, and all the other considerations about new locations.

The playbooks will have a structure where it makes sense to think what tasks / use cases / when they get to be used.

As for the checks themselves (Python files), they'd still live within the role, and we may add more structure there as we see need for it.

> I left behind symlinks for existing playbooks. Do we think these should stay around for a transition period?

I agree with @codificat, let's keep symlinks, for a while or forever...

sosiouxme:
On Wed, May 10, 2017 at 10:30 AM, Pep Turró Mauri <notifications@github.com>
wrote:

> the readme in the cert expiry roles
> <https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_certificate_expiry/README.md>
> needs to be updated
>
>
Thanks, I missed that... fixed


> , but beyond that there are also references to this path in the docs
> about certificate redeployment
> <https://docs.openshift.com/container-platform/latest/install_config/redeploying_certificates.html#install-config-cert-expiry-running-playbooks>.
> These need to be updated, and by now this means we need to grant a
> transition period.
>
>
Definitely. OK, and that's a followup task once this merges...


> To begin with, I'm not even sure how much sense it makes to group all the
> *health* (i.e. running cluster) checks together, because it's likely that
> each will need to be run at different times/intervals and/or with different
> targets. For example: the etcd health check will probably run daily, while
> the cert expiry will run maybe monthly.
>

True, we might end up with a lot of check tags, based on how customers want
to run these (manually, hourly, daily, off-hours, ...). I think we'll
eventually end up taking an extra param for specifying which checks you
want in a particular run.


> When/if we group them, should start tagging tasks accordingly? E.g. should
> we add the *@health <https://github.com/health>* tag to the tasks in the
> certchecks playbooks? As the action-plugin based checks have their own
> groups/tags this can get potentially confusing, specially if they start to
> diverge - in the sense that *@health <https://github.com/health>* or some
> other tag might not necessarily mean exactly the same when it's part of a
> task vs. when it's inside a check, right?
>

I don't think task tags will be very useful to us. They filter *all* tasks,
including dependencies like openshift_facts. So they have to be present on
all those tasks we need, and I don't think we're going to proliferate tags
throughout the codebase just for checks. We're building an alternate tag
system for selecting checks and while I'm not crazy about doing that, it
does allow us to subvert ansible's usual mode and run things in parallel /
collate multiple failures. Of course we're breaking that immediately by
having expiry checks be separate :) That's why I was saying we need to
think about reporting requirements - so we could generalize what the expiry
playbooks are doing and bring everything under the one check action.

sosiouxme:
aos-ci-test

rhcarvalho:
We got green from aos-ci-test, but Travis did not run. Pushed a new commit with docs changes to re-trigger Travis.

playbooks/openshift-checks/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+2. Diagnostic playbook ([health.yml](health.yml)) - check an existing cluster
+   for known signs of problems.
+
+3. Certificate expiry playbooks ([certificate_expiry](certificate_expiry)) -
+   check that certificates in use are valid and not expiring soon.
+
+## Running
+
+With an installation of Ansible 2.3 or greater, run the playbook directly
```
rhcarvalho: We probably better not mention anything about versions here. As we can see, this easily drifts behind the actual requirements, and apart from Ansible we also need other dependencies to be installed on the host...
Probably better to link to the /README.md or somewhere else and avoid duplicating the instructions how to run playbooks from this repo.

playbooks/openshift-checks/README.md
```diff
@@ -0,0 +1,65 @@
...
+    or
+
+    ```console
+    $ ansible-playbook -i <inventory file> playbooks/openshift-checks/health.yml
+    ```
+
+    or
+
+    ```console
+    $ ansible-playbook -i <inventory file> playbooks/openshift-checks/certificate_expiry/default.yaml -v
```
rhcarvalho: Why do we include `-v` in this example?

playbooks/openshift-checks/health.yml
```diff
@@ -0,0 +1,15 @@
+---
+- include: ../openshift-cluster/initialize_groups.yml
+
+- name: Run OpenShift health checks
+  hosts: OSEv3
+  # Temporarily reverting to OSEv3 until group standardization is complete
```
rhcarvalho: I think we can get rid of this line, it explains something out of context, that only makes sense if we go look at the Git history.

playbooks/openshift-checks/pre-install.yml
```diff
@@ -0,0 +1,15 @@
+---
+- include: ../openshift-cluster/initialize_groups.yml
+
+- hosts: OSEv3
+  # Temporarily reverting to OSEv3 until group standardization is complete
```
rhcarvalho: Same here, this comment is not useful unless someone is reading the Git history, and if so, the commit logs and PRs where the changes happened tell a better story.

playbooks/openshift-checks/pre-install.yml
```diff
@@ -0,0 +1,15 @@
...
+  name: run OpenShift pre-install checks
+  roles:
+    - openshift_health_checker
+  post_tasks:
+    # NOTE: we need to use the old "action: name" syntax until
+    # https://github.com/ansible/ansible/issues/20513 is fixed.
```
rhcarvalho: https://github.com/ansible/ansible/issues/20513 is stale, we can probably just get rid of this comment as well (here and in the other playbook).
The only purpose it serves is to stop someone from "fixing" the YAML to use the other syntax and breaking the playbook. Is it a worth cause?

rhcarvalho:
Syntax-checking the `playbooks/byo/openshift-preflight/check.yml` playbook in Python 2.7 is not finding the roles dir, missing symlink?

https://travis-ci.org/openshift/openshift-ansible/jobs/231469149#L444-L445

rhcarvalho:
> missing symlink?

Yes, re-added the symlink in https://github.com/sosiouxme/openshift-ansible/pull/1, the syntax check goes green for me locally.

playbooks/openshift-checks/README.md
codificat: Because without `-v` the playbook doesn't show any result... maybe at some point we should add a callback plugin to the certificate expiration checks, but at the moment it's the only way of seeing useful information (beyond looking at the generated reports on disk - which this particular playbook does not generate).

rhcarvalho: Aha, so is it coming from a "debug" action or similar?

playbooks/openshift-checks/health.yml
sosiouxme: Well git history will be confusing as we're moving things around... but you're right, this is more a commit message.

playbooks/openshift-checks/pre-install.yml
sosiouxme: Until we have something in CI that tests the playbooks, I think it is worth a quick link like:

    - action: openshift_health_check  # NOTE: https://github.com/ansible/ansible/issues/20513

Also reminds us to check the issue and "fix" the YAML when it's ready.

sosiouxme:
Actually the playbook file should not be a symlink, just needs to include
the other playbook. Then no role symlink is needed.

sosiouxme:
Thanks for the comments, I believe I've addressed them.

Keep in mind that it's rather painful to move locations as we see here... perhaps we should get a few opinions from outside Customer Success to help ensure we don't do this again. @sdodson @ingvagabund any thoughts/concerns on this playbook location?

rhcarvalho:
> Actually the playbook file should not be a symlink, just needs to include
> the other playbook. Then no role symlink is needed.

That should work as well :+1:

sosiouxme:
@mtnbikenc just wanted to verify that the reorg here is what you think best for consistency, as far as `playbooks/byo/openshift-preflight` -> `playbooks/byo/openshift-checks` + `playbooks/common/openshift-checks`

(Most of this PR is just READMEs following that change and the `certificate_expiry` move)

sosiouxme:
aos-ci-test

sosiouxme:
aos-ci-test

sosiouxme:
borky CI :-1:

sosiouxme:
@juanvallejo @codificat it passed CI test, can I get an approved review?

sosiouxme:
Thanks all!

[merge]

sosiouxme:
@codificat can you make any sense of this error?

https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/419/consoleFull#118965008358b6e51eb7608a5981914356 (look a little above)

```
file /usr/share/ansible/openshift-ansible/playbooks/certificate_expiry from install of openshift-ansible-playbooks-3.6.69-1.git.0.6d125ae.el7.noarch conflicts with file from package openshift-ansible-playbooks-3.5.58-1.git.1.49cb4ad.el7.noarch
```
I can't see how an RPM can conflict with itself during an upgrade but maybe it's something to do with the build script changes?

codificat:
> Or, more likely I suppose, with moving the certificate_expiry directory and replacing with a symlink.

Indeed. We hit an [old limitation of RPM](https://bugzilla.redhat.com/show_bug.cgi?id=447156), where an upgrade can't replace a dir with a symlink. Joy.

A workaround using a `pretrans` scriptlet is documented in the [packaging guidelines](https://fedoraproject.org/wiki/Packaging:Directory_Replacement)

sosiouxme:
aos-ci-test

sdodson:
aos-ci-test

sosiouxme:
alright now [merge]

sdodson:
[merge]

Issue #4083, opened 2017-05-03
es_number_of_shards and es_number_of_replicas aren't used in roles/openshift_logging/templates/es.j2 template


harobed:
In `roles/openshift_logging/tasks/set_es_storage.yaml`:

* [es_number_of_shards](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/tasks/set_es_storage.yaml#L79)
* and [es_number_of_replicas](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/tasks/set_es_storage.yaml#L80)

aren't used in [roles/openshift_logging/templates/es.j2](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/templates/es.j2) template.

Why? It's a mistake?

cc @ewolinetz 

Best regards,
Stéphane

ewolinetz:
It is a mistake -- that should be set for the configmap template. 
That must have been overlooked on the change to introduce the ability to configure shard counts.
Thank you for pointing that out!

harobed:
Do you prefer a pull request for `release-1.5` branch and `master` or you can do that today (I need to install it)?

ewolinetz:
@harobed #4084 is the 1.5 fix, i will also cherry pick it for master afterwards. Which version is a higher priority for you?

ewolinetz:
#4085 is the cherry pick pr

ewolinetz:
Should now be resolved by #4084 and #4085

PR #4084, opened 2017-05-03
Correctly setting the primary and replica shard count settings

ewolinetz:
As pointed out in #4083 we don't use the shard count settings in the creation of the ES DC, it is set in the configmap


ewolinetz:
aos-ci-test

ewolinetz:
[merge]

ewolinetz:
merge flaked on https://github.com/openshift/origin/issues/12264

ewolinetz:
[merge]

Issue #4082, opened 2017-05-03
upgrade 1.4.1 to 1.5.0 error (Check for invalid namespaces and SDN errors)


wenskys:
Hi,
I have a problem when i upgrage openshift.It looks like UnicodeDecodeError bug?
```lang=bash
TASK [Check for invalid namespaces and SDN errors] *****************************
fatal: [op1.example.com]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Shared connection to op1.example.com closed.\r\n", "module_stdout": "
Traceback (most recent call last):\r\n  File \"/tmp/ansible_w0XVgT/ansible_module_oc_objectvalidator.py\", line 1475, in <module>\r\n    main()\r\n  File \"/tmp/ansible_w0XVgT/ansible_module_oc_objectvalidator.py\", line 1468, in main\r\n    rval = OCObjectValidator.run_ansible(module.params)\r\n  File \"/tmp/ansible_w0XVgT/ansible_module_oc_objectvalidator.py\", line 1430, in run_ansible\r\n   
 success, rval, invalid = objectvalidator.get_invalid(resource, invalid_filter)\r\n  File \"/tmp/ansible_w0XVgT/ansible_module_oc_objectvalidator.py\", line 1379, in get_invalid\r\n    rval = self._get(kind)\r\n  File \"/tmp/ansible_w0XVgT/ansible_module_oc_objectvalidator.py\", line 907, in _get\r\n    rval = self.openshift_cmd(cmd, output=True)\r\n  File \"/tmp/ansible_w0XVgT/ansible_module_oc_objectvalidator.py\", line 1033, in openshift_cmd\r\n    returncode, stdout, stderr = self._run(cmds, input_data)\r\n  File \"/tmp/ansible_w0XVgT/ansible_module_oc_objectvalidator.py\", line 1008, in _run\r\n    return proc.returncode, stdout.decode(), stderr.decode()\r\nUnicodeDecodeError: 'ascii' codec can't decode byte 0xe6 in position 521: ordinal not in range(128)\r\n", "msg": "MODULE FAILURE"}
	to retry, use: --limit @/root/openshift-ansible/playbooks/byo/openshift-cluster/upgrades/v3_5/upgrade.retry
```
![error-1](https://cloud.githubusercontent.com/assets/12937669/25665454/3ba2c58e-3050-11e7-8f31-5ad62b0683b5.png)

such as this：https://access.redhat.com/solutions/3020271
thanks.

[...]

PR #4089, opened 2017-05-03
Allow oc_ modules to pass unicode results

mtnbikenc:
The JSON output of the oc command could contain Unicode characters based on the regional naming of various objects.

One line change to base.py _run.

Related to:
https://bugzilla.redhat.com/show_bug.cgi?id=1444806
https://bugzilla.redhat.com/show_bug.cgi?id=1446471
#4082

### DIFF
--- a/roles/lib_openshift/library/oc_adm_ca_server_cert.py
+++ b/roles/lib_openshift/library/oc_adm_ca_server_cert.py
@@ -1080,7 +1080,7 @@ class OpenShiftCLI(object):
 
         stdout, stderr = proc.communicate(input_data)
 
-        return proc.returncode, stdout.decode(), stderr.decode()
+        return proc.returncode, stdout.decode('utf-8'), stderr.decode('utf-8')
 
     # pylint: disable=too-many-arguments,too-many-branches
     def openshift_cmd(self, cmd, oadm=False, output=False, output_type='json', input_data=None):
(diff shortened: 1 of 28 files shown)


mtnbikenc:
aos-ci-test

roles/lib_openshift/library/oc_adm_ca_server_cert.py
```diff
@@ -1080,7 +1080,7 @@ def _run(self, cmds, input_data):
 
         stdout, stderr = proc.communicate(input_data)
 
-        return proc.returncode, stdout.decode(), stderr.decode()
+        return proc.returncode, stdout.decode('utf-8'), stderr.decode('utf-8')
```
ashcrow: ## 2
### str
Becomes ``unicode``
### unicode
Stays ``unicode``

## 3
### bytes
Become ``str``
### str (which shouldn't be returned in stdout/stderr)
Will error.

ashcrow: ## Python 2
```
In [1]: data = '0xc3'

In [2]: print(data.decode('utf-8'))
0xc3
```

## Python 3
```
In [6]: data = b'0xc3'

In [7]: print(data.decode('utf-8'))
0xc3
```

mtnbikenc:
[merge]

Issue #4090, opened 2017-05-03
oc_service is unable to take externalIPs


ewolinetz:
Per [1] we should be able to create a service with an external IP however the module does not have a parameter to specify one.

This is required for the openshift_logging_mux role being worked on currently

[1] https://kubernetes.io/docs/concepts/services-networking/service/#external-ips

ewolinetz:
@kwoodson

PR #4091, opened 2017-05-04
add support for oc_service for labels, externalIPs

richm:
Add support for `labels` and `externalIPs` in Services
@ewolinetz @jcantrill PTAL


richm:
fixes https://github.com/openshift/openshift-ansible/issues/4090

richm:
@kwoodson

jcantrill:
LGTM

sdodson:
What's the minimum version of OCP required to use this?

richm:
@sdodson not sure - label and externalIPs support has been in since at least 1.4, probably earlier

sdodson:
Cool, 3.4 or larger is fine.

ashcrow:
Will review after changes made.

richm:
@kwoodson I can't figure out how registry and router process the labels parameter.  There is one reference to labels in ansible/oc_adm_registry.py, in ansible/oc_adm_router.py, class/oc_adm_registry.py, in class/oc_adm_router.py - and in none of those places is the labels argument actually used.  Also, why list?  For example:
```
$ oc get pod router-1-z4knc -o yaml
...
  labels:
    deployment: router-1
    deploymentconfig: router
    router: router
```
This looks like a `dict` to me.  If it is a `list`, what is the format in the yaml representation?

richm:
```
# oadm router --help|grep labels
      --labels='router=<name>': A set of labels to uniquely identify the router and its components.
# oadm router -o yaml
    labels:
      router: router
```

richm:
I think I figured out how to do this, and with testing too. PTAL.

kwoodson:
@richm,  apologies, I thought this was Eric's PR.  Comments below.

I'm not sure we've tested labels with the router/registry creation.  I just don't think that's something we needed to modify.  Now with that being said, we should support it and the idea behind the params being a list is that it would get sent to the command line like this:
```--labels=key=val,key1=val1```

You are correct that there is no processing of labels.  This is why my comment above.  There should be a step before it gets sent to the ServiceConfig.  Should be building a the structure like this:
```python
options['labels'] = (['{}={}'.format(key, val) for key,val in params['labels'].items()]).join(',')
```

Does this make sense?

Thanks for picking this up and adding support.

richm:
@kwoodson What about `--label` or `--labels` being a multi-valued flag e.g. `--label key1=value1 --label key2=value2 ...`

ashcrow:
Travis noted:
```
./roles/lib_openshift/src/test/unit/test_oc_adm_registry.py:221:47: E231 missing whitespace after ':'
```

roles/lib_openshift/library/oc_adm_registry.py
```diff
@@ -2049,6 +2056,7 @@ class Service(Yedit):
     cluster_ip = "spec.clusterIP"
     selector_path = 'spec.selector'
     kind = 'Service'
+    external_ips = "spec.externalIPs"
```
ashcrow: Not related to this PR directly, but at some point these attributes should probably be renamed to follow a pattern such as ``cluster_ip_key``, etc..

roles/lib_openshift/library/oc_adm_registry.py
```diff
@@ -2110,6 +2118,53 @@ def add_portal_ip(self, pip):
         '''add cluster ip'''
         self.put(Service.portal_ip, pip)
 
+    def get_external_ips(self):
+        ''' get a list of external_ips '''
+        return self.get(Service.external_ips) or []
```
ashcrow: Is there a reason to not use this instances attribute: ``self.external_ips``? Or are you following the pre set pattern in the file?

richm: I'm just following the pre set pattern in the file.

richm:
@kwoodson 
> There should be a step before it gets sent to the ServiceConfig. Should be building a the structure like this:
```
options['labels'] = (['{}={}'.format(key, val) for key,val in params['labels'].items()]).join(',')
```

Where?  I'm looking at the code for oc_adm_registry and router.  It looks like the function `to_option_list` is responsible for transforming the internal structure into the proper format for the cli.  This function would take `labels: {"foo":"bar", "baz":"biff"}`.  AFAICT, the best place to do this would be in `to_option_list`, to make it convert the `labels` value `dict` into a comma delimited list as you have above.

richm:
@kwoodson @ashcrow ptal

ashcrow:
Looks like there are some formatting issues https://travis-ci.org/openshift/openshift-ansible/jobs/236488145#L184

richm:
@kwoodson @ashcrow any clues about what I should do about coverage?  Is that a merge blocker?  The problem is that I changed the `to_option_list()` code which `generate.py` includes in every single python file - a lot more code changes + very little coverage change == small percentage decrease in coverage.

ashcrow:
@richm coverage is not a blocker. I'll kick the tests off again now.

ashcrow:
aos-ci-test

richm:
> error: aos-ci-jenkins/OS_3.6_containerized for a1cfec9 (logs)

flake?

sdodson:
aos-ci-test

sdodson:
Yes, you got hit by container images being newer/older than rpm repo because a build was in flight.

sdodson:
aos-ci-test

sdodson:
we'll manually merge tomorrow morning if this fails again

richm:
> we'll manually merge tomorrow morning if this fails again

Thanks!

richm:
@sdodson ready to merge, or do you want to try to aos-ci-test again?

ashcrow:
@richm it should be merged now 🎆

richm:
@ashcrow yep, sorry, I jumped the gun

PR #4093, opened 2017-05-04
mux startup is broken without this fix

richm:
I've tested this both with and without mux, install and delete logging.
I was never able to reproduce the original problem, not sure if this
ansible logic will prevent it.
@ewolinetz ptal

### DIFF
--- a/roles/openshift_logging/tasks/start_cluster.yaml
+++ b/roles/openshift_logging/tasks/start_cluster.yaml
@@ -36,7 +36,7 @@
     name: "{{ object }}"
     namespace: "{{openshift_logging_namespace}}"
     replicas: "{{ openshift_logging_mux_replica_count | default (1) }}"
-  with_items: "{{ mux_dc.results.results[0]['items'] if 'results' in mux_dc else {} | map(attribute='metadata.name') | list }}"
+  with_items: "{{ mux_dc.results.results[0]['items'] | map(attribute='metadata.name') | list if 'results' in mux_dc else [] }}"
   loop_control:
     loop_var: object
   when:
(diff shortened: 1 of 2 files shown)


roles/openshift_logging/tasks/start_cluster.yaml
```diff
@@ -36,7 +36,7 @@
     name: "{{ object }}"
     namespace: "{{openshift_logging_namespace}}"
     replicas: "{{ openshift_logging_mux_replica_count | default (1) }}"
-  with_items: "{{ mux_dc.results.results[0]['items'] if 'results' in mux_dc else {} | map(attribute='metadata.name') | list }}"
+  with_items: "{{ mux_dc.results.results[0]['items'] | map(attribute='metadata.name') | list if 'results' in mux_dc else [] }}"
```
ewolinetz: is there any difference in having the `if` before or after `map(...) | list` ?

richm: `mux_dc.results.results[0]['items'] if 'results' in mux_dc else {} | map(attribute='metadata.name') | list`
In this case, the `if` is before the `map(...)`.  What this does is set `object` to `mux_dc.results.results[0]['items']`, which is the entire definition of the dc, which is passed to the oc scale command line, which causes a failure.

`mux_dc.results.results[0]['items'] | map(attribute='metadata.name') | list if 'results' in mux_dc else []`
In this case, the `if` is after the `map(...)`.  What this does is set `object` to the evaluation of the entire pipeline `mux_dc.results.results[0]['items'] | map(attribute='metadata.name') | list`, which is the list of dc names, which is the argument expected by oc scale.

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

PR #4094, opened 2017-05-04
Ensure good return code for specific until loops

ashcrow:
This simply adds a check for ``returncode`` to the ``until``.

@sdodson @jupierce


ingvagabund:
aos-ci-test

PR #4098, opened 2017-05-04
[1.5] Allow oc_ modules to pass unicode results

mtnbikenc:
Backports #4089

Fixes bug 1444806
Fixes bug 1446471

### DIFF
--- a/roles/lib_openshift/library/oadm_manage_node.py
+++ b/roles/lib_openshift/library/oadm_manage_node.py
@@ -1019,7 +1019,7 @@ class OpenShiftCLI(object):
 
         stdout, stderr = proc.communicate(input_data)
 
-        return proc.returncode, stdout.decode(), stderr.decode()
+        return proc.returncode, stdout.decode('utf-8'), stderr.decode('utf-8')
 
     # pylint: disable=too-many-arguments,too-many-branches
     def openshift_cmd(self, cmd, oadm=False, output=False, output_type='json', input_data=None):
--- a/roles/lib_openshift/library/oc_adm_ca_server_cert.py
+++ b/roles/lib_openshift/library/oc_adm_ca_server_cert.py
@@ -1027,7 +1027,7 @@ class OpenShiftCLI(object):
 
         stdout, stderr = proc.communicate(input_data)
 
-        return proc.returncode, stdout.decode(), stderr.decode()
+        return proc.returncode, stdout.decode('utf-8'), stderr.decode('utf-8')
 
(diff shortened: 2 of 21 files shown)


mtnbikenc:
aos-ci-test

enj:
We need some test data that has unicode characters to actually exercise this.

enj:
Also, it does not really work:

```
>>> import subprocess
>>> proc = subprocess.Popen(['cat', '/tmp/dev_random_data.txt'], stdout=subprocess.PIPE)
>>> stdout, stderr = proc.communicate()
>>> stdout.decode('utf-8')
  File "/usr/lib64/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0x95 in position 1: invalid start byte
```

We need to actually deal with errors:

```
decode(...)
    S.decode([encoding[,errors]]) -> object
    
    Decodes S using the codec registered for encoding. encoding defaults
    to the default encoding. errors may be given to set a different error
    handling scheme. Default is 'strict' meaning that encoding errors raise
    a UnicodeDecodeError. Other possible values are 'ignore' and 'replace'
    as well as any other name registered with codecs.register_error that is
    able to handle UnicodeDecodeErrors.
```

ashcrow:
When it comes to random data I think ``ignore`` is fine. When it comes to data we expect to be valid strings/unicode I would expect an error to raise if it is, in fact, not valid.

@mtnbikenc @enj thoughts?

abutcher:
I don't think we'll encounter any random data so this should be sufficient.

```python
>>> beef = "ÐÊÄÐ-ßÈËF"
>>> beef.decode()
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 0: ordinal not in range(128)
>>> beef.decode('utf-8')
>>> print beef.decode('utf-8')
ÐÊÄÐ-ßÈËF
```

ashcrow:
Added the unicode used in @abutcher's example to a file, and used subprocess to check out ``decode('utf-8')``:

```python
In [1]: import subprocess

In [2]: p = subprocess.Popen(["cat", "data"], stdout=subprocess.PIPE)

In [3]: o, e = p.communicate()

In [4]: o
Out[4]: b'\xc3\x90\xc3\x8a\xc3\x84\xc3\x90-\xc3\x9f\xc3\x88\xc3\x8bF\n'

In [5]: o.decode('utf-8')
Out[5]: 'ÐÊÄÐ-ßÈËF\n'

In [6]: 
```

PR #4113, opened 2017-05-08
bug 1448860. Change recovery_after_nodes to match node_quorum

jcantrill:


jcantrill:
[aos-ci-test]

jcantrill:
@portante does this need to be cherry-picked into 1.5?

sdodson:
aos-ci-test

sdodson:
that bot is very picky

roles/openshift_logging/templates/elasticsearch.yml.j2
```diff
@@ -32,7 +32,7 @@ discovery:
 
 gateway:
   expected_master_nodes: ${NODE_QUORUM}
```
portante: Shouldn't `expected_master_nodes` be the total cluster size?  If you keep this as a quorum value, then for a `>=` three node cluster, it will start recovering the state of the cluster when 2 nodes are present.  This means it will start to move/copy data to restore replicas and rebalance the cluster.  We only want this to happen when we exceed the `recover_after_time` value.

portante: And since we currently only support master nodes serving data, we should probably drop this entirely, and just use the `expected_nodes` parameter.

roles/openshift_logging/vars/main.yaml
```diff
@@ -3,10 +3,8 @@ openshift_master_config_dir: "{{ openshift.common.config_base }}/master"
 es_node_quorum: "{{openshift_logging_es_cluster_size|int/2 + 1}}"
 es_min_masters_default: "{{ (openshift_logging_es_cluster_size | int / 2 | round(0,'floor') + 1) | int }}"
```
portante: This calculation should end up being the same as `es_node_quorum`.  So we should probably merge this with `node_quorum` and use that instead eliminating this parameter.

roles/openshift_logging/templates/elasticsearch.yml.j2
jcantrill: dropped expected_master_nodes

roles/openshift_logging/vars/main.yaml
jcantrill: removed

roles/openshift_logging/vars/main.yaml
```diff
@@ -1,12 +1,8 @@
...
 openshift_master_config_dir: "{{ openshift.common.config_base }}/master"
-es_node_quorum: "{{openshift_logging_es_cluster_size|int/2 + 1}}"
-es_min_masters_default: "{{ (openshift_logging_es_cluster_size | int / 2 | round(0,'floor') + 1) | int }}"
-es_min_masters: "{{ (openshift_logging_es_cluster_size == 1) | ternary(1, es_min_masters_default)}}"
-es_recover_after_nodes: "{{openshift_logging_es_cluster_size|int - 1}}"
-es_recover_expected_nodes: "{{openshift_logging_es_cluster_size|int}}"
-es_ops_node_quorum: "{{openshift_logging_es_ops_cluster_size|int/2 + 1}}"
-es_ops_recover_after_nodes: "{{openshift_logging_es_ops_cluster_size|int - 1}}"
-es_ops_recover_expected_nodes: "{{openshift_logging_es_ops_cluster_size|int}}"
+es_node_quorum: "{{openshift_logging_es_cluster_size | int/2 + 1}}"
```
portante: Do we need this `round(0,'floor')` trick?

roles/openshift_logging/templates/elasticsearch.yml.j2
ewolinetz: We're bringing the ability to have client nodes with https://github.com/openshift/openshift-ansible/pull/3509

So I would like to keep this in there, it doesn't do any harm being there now, correct?

roles/openshift_logging/vars/main.yaml
jcantrill: no.  my test with a python interpreter '1/2 =0' .  Because it is an int it drops the precision

roles/openshift_logging/templates/elasticsearch.yml.j2
portante: Except that with those new parameters we'll still have to make changes to accommodate the new way to specify their values.  Seems better to apply this to the current code base, and then have that patch make the appropriate changes given the new parameters.

ewolinetz: I think using environment variables will be best as we can then maintain a single configmap and have the deployment config (and eventually statefulset) dictate what should be used.  So there wouldn't be a change

roles/openshift_logging/templates/elasticsearch.yml.j2
```diff
@@ -28,11 +28,10 @@ cloud:
 discovery:
   type: kubernetes
   zen.ping.multicast.enabled: false
-  zen.minimum_master_nodes: {{es_min_masters}}
+  zen.minimum_master_nodes: {{es_node_quorum}}
```
jcantrill: @ewolinetz I presume we should maybe convert this to rely on the env variable?  or modify line 34 to use the variable.  Thoughts?  Should these all use vars given they are unlikely to change during the deployment? @portante thoughts?

portante: So as an operator, I want to be able to update my deployment configuration for my ES pod and redeploy the pod.  If I add nodes, I want to be able to add the new node, and then gracefully update the other nodes to have the same configuration.  As long as I can redeploy one pod at a time to pick up the new configuration, we're good.  So if that means keepings these as environment variables as provided by the deployment configuration, great.

jcantrill: They are originally ENV VARS because of the deployer and exist now because of legacy.  We should ask the question regarding all these ENV vars and if they should remain or become an ansible variable.  I assume the threshold is whether or not we desire to tweak the value after deployment.  I assume node quorum and min masters we would prefer to peg those based on ansible inventory where timeouts would be nice to be able to change and tweak?  @ewolinetz ?

portante: I would think we'd want them as ansible variables so that via ansible we could update the deployment properly.

But that would be coordinating how to do that in ansible in a graceful way.

ewolinetz: i think they should remain as env vars, especially since we use a single configmap for ES (and ops). Having this value in a variable would allow us to scale ops and non-ops differently and not cause one to go haywire waiting for a number of master nodes that may never become available

jcantrill: I will update to use ENV var for consistency and we can change later if that's the path we choose.

jcantrill:
aos-ci-test

portante:
This all looks good.  What is the hold-up?

jcantrill:
[merge]

PR #4117, opened 2017-05-08
Flatten package install and upgrade into one transaction

sdodson:
Without this we were pulling in unbounded dependencies and upgrading
to the latest version available in a repo.

Fixes  https://github.com/openshift/origin/issues/13998

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
@@ -1,27 +1,39 @@
 ---
-# We verified latest rpm available is suitable, so just yum update.
+# When we update package "a-${version}" and a requires b >= ${version} if we
+# don't specify the version of b yum will choose the latest version of b
+# available and the whole set of dependencies end up at the latest version.
+# Since the package module, unlike the yum module, doesn't flatten a list
+# of packages into one transaction we need to do that explicitly. The ansible
+# core team tells us not to rely on yum module transaction flattening anyway.
 
-# Master package upgrade ends up depending on node and sdn packages, we need to be explicit
-# with all versions to avoid yum from accidentally jumping to something newer than intended:
+# TODO: If the sdn package isn't already installed this will install it, we
+# should fix that
+
 - name: Upgrade master packages
-  package: name={{ item }} state=present
-  when: component == "master"
-  with_items:
-  - "{{ openshift.common.service_type }}{{ openshift_pkg_version }}"
-  - "{{ openshift.common.service_type }}-master{{ openshift_pkg_version }}"
-  - "{{ openshift.common.service_type }}-node{{ openshift_pkg_version }}"
-  - "{{ openshift.common.service_type }}-sdn-ovs{{ openshift_pkg_version }}"
-  - "{{ openshift.common.service_type }}-clients{{ openshift_pkg_version }}"
+  package: name={{ master_pkgs | join(',') }} state=present
+  vars:
+    master_pkgs:
+      - "{{ openshift.common.service_type }}{{ openshift_pkg_version }}"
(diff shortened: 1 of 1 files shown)


sdodson:
@stevekuznetsov this will solve all your problems

stevekuznetsov:
I thought the default `package` was supposed to do single transactions for `with_items`?

stevekuznetsov:
Might be good to comment as to why this looks so non-idiomatic for future reference.

playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
```diff
@@ -3,24 +3,35 @@
...
+
 - name: Upgrade master packages
-  package: name={{ item }} state=present
+  package: name={{ master_pkgs | join(',') }} state=present
+  vars:
+    master_pkgs:
+      - "{{ openshift.common.service_type }}{{ openshift_pkg_version }}"
+      - "{{ openshift.common.service_type }}-master{{ openshift_pkg_version }}"
+      - "{{ openshift.common.service_type }}-node{{ openshift_pkg_version }}"
+      - "{{ openshift.common.service_type }}-sdn-ovs{{ openshift_pkg_version }} if openshift_use_openshift_sdn | default(true) else omit"
```
sdodson: @abutcher what's the correct syntax for this? I don't want to emit this item if we're not using the sdn.

abutcher: You may have to do something like this since `omit` will add a string to this list.

```yaml
- name: Upgrade master packages
  package: name="{{ master_pkgs | difference(['remove_me']) | join(',') }}" state=present
  vars:
    master_pkgs:
    - "{{ openshift.common.service_type }}{{ openshift_pkg_version }}"
    - "{{ openshift.common.service_type }}-master{{ openshift_pkg_version }}"
    - "{{ openshift.common.service_type }}-node{{ openshift_pkg_version }}"
    - "{{ openshift.common.service_type }}-sdn-ovs{{ openshift_pkg_version }} if openshift_use_openshift_sdn | default(true) else 'remove_me'"
```

sdodson:
> I thought the default package was supposed to do single transactions for with_items?

No, yum module does but package doesn't. dnf may too but doesn't mention it in the docs for the module, not sure.

playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
sdodson: It's broken in the current code, I guess i'll leave it broken in my fix.

stevekuznetsov:
@sdodson you can trigger this job for this PR manually here: https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_extended_conformance_install_update/

playbooks/common/openshift-cluster/upgrades/rpm_upgrade.yml
```diff
@@ -1,27 +1,39 @@
...
-# We verified latest rpm available is suitable, so just yum update.
+# When we update package "a-${version}" and a requires b >= ${version} if we
+# don't specify the version of b yum will choose the latest version of b
+# available and the whole set of dependencies end up at the latest version.
+# Since the package module, unlike the yum module, doesn't flatten a list
+# of packages into one transaction we need to do that explicitly. The ansible
+# core team tells us not to rely on yum module transaction flattening anyway.
+
+# TODO: If the sdn package isn't already installed this will install it, we
+# shouldn fix that
```
stevekuznetsov: We should or should not fix this?

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
[merge]

PR #4118, opened 2017-05-08
Temporarily revert to OSEv3 host group usage

mtnbikenc:
Group normalization has caused an issue when calling new group names that have not been populated.  Temporarily reverting host groups to resolve issues while working out a long-term path forward.

Fixes bug 1447939

PR #4133, opened 2017-05-09
Fix regression, 'no hosts matched', in rhel_sub

tbielawa:
The rhel_sub playbook broke again for me. Discovered after rebuilding my OS1 nodes. Running the playbook would generate all of the `g_*` groups, but once the second play runs the facts are lost. This references the plays correctly now.

Though I am not sure if this is the ideal solution. Please comment if you have a better idea.


mtnbikenc:
I have a fix in #4118 
https://github.com/openshift/openshift-ansible/pull/4118/files#diff-377065f5e66159003edb0fc50ed24ee4R8

PR #4135, opened 2017-05-09
Fix issue with Travis-CI using old pip version

mtnbikenc:
Travis 2.7 test environments use pip 6.0.7 which is very old.  This causes a failure when installing the 2.14 (latest) version of `requests`.  Adding a step to upgrade pip before performing other installs.

See: https://github.com/kennethreitz/requests/issues/4006

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ python:
   - "3.5"
 
 install:
+  - pip install --upgrade pip
   - pip install tox-travis coveralls
 
 script:


mtnbikenc:
@sdodson Once we get another thumbs up on this, I recommend foregoing all the standard testing and merging this to unblock the Travis-CI tests.

sdodson:
It made travis go green and that's the only thing that this would affect so merged. Can you pick it into release branches?

rhcarvalho:
👍

PR #4136, opened 2017-05-09
use docker_log_driver and /etc/docker/daemon.json to determine log driver

richm:
@jcantrill @ewolinetz @nhosoi @lukas-vlcek @ashcrow PTAL
Basically, use `docker_log_driver` to determine if journald is used, but
at the same time, mount /etc/docker/daemon.json in the fluentd pod just
in case.

### DIFF
--- a/roles/openshift_logging/defaults/main.yml
+++ b/roles/openshift_logging/defaults/main.yml
@@ -72,7 +72,7 @@ openshift_logging_fluentd_nodeselector: "{{ openshift_hosted_logging_fluentd_nod
 openshift_logging_fluentd_cpu_limit: 100m
 openshift_logging_fluentd_memory_limit: 512Mi
 openshift_logging_fluentd_es_copy: false
-openshift_logging_fluentd_use_journal: "{{ openshift_hosted_logging_use_journal | default('') }}"
+openshift_logging_fluentd_use_journal: "{{ openshift_hosted_logging_use_journal if openshift_hosted_logging_use_journal is defined else (docker_log_driver == 'journald') | ternary(True, False) if docker_log_driver is defined else (openshift.docker.log_driver == 'journald') | ternary(True, False) if openshift.docker.log_driver is defined else openshift.docker.options | search('--log-driver=journald') if openshift.docker.options is defined else default(omit) }}"
 openshift_logging_fluentd_journal_source: "{{ openshift_hosted_logging_journal_source | default('') }}"
 openshift_logging_fluentd_journal_read_from_head: "{{ openshift_hosted_logging_journal_read_from_head | default('') }}"
 openshift_logging_fluentd_hosts: ['--all']
--- a/roles/openshift_logging/templates/fluentd.j2
+++ b/roles/openshift_logging/templates/fluentd.j2
@@ -59,6 +59,9 @@ spec:
         - name: dockercfg
           mountPath: /etc/sysconfig/docker
           readOnly: true
+        - name: dockerdaemoncfg
+          mountPath: /etc/docker
+          readOnly: true
 {% if openshift_logging_use_mux_client | bool %}
         - name: muxcerts
           mountPath: /etc/fluent/muxkeys
@@ -154,6 +157,9 @@ spec:
       - name: dockercfg
         hostPath:
           path: /etc/sysconfig/docker
+      - name: dockerdaemoncfg
+        hostPath:
+          path: /etc/docker
 {% if openshift_logging_use_mux_client | bool %}
       - name: muxcerts
         secret:


richm:
@stevekuznetsov ptal

jcantrill:
@richm probably will kill the 1.5 CI too since it uses the 3.6 amis

roles/openshift_logging/defaults/main.yml
```diff
@@ -72,7 +72,7 @@ openshift_logging_fluentd_nodeselector: "{{ openshift_hosted_logging_fluentd_nod
 openshift_logging_fluentd_cpu_limit: 100m
 openshift_logging_fluentd_memory_limit: 512Mi
 openshift_logging_fluentd_es_copy: false
-openshift_logging_fluentd_use_journal: "{{ openshift_hosted_logging_use_journal | default('') }}"
+openshift_logging_fluentd_use_journal: "{{ openshift_hosted_logging_use_journal if openshift_hosted_logging_use_journal is defined else (docker_log_driver|default('') == 'journald') | ternary(True, False) }}"
```
ewolinetz: where is `docker_log_driver` being set? and is that preferable to using the docker option from openshift_facts? 
e.g. {{ openshift.docker.options | search('journald') }}

richm: @ewolinetz https://github.com/openshift/openshift-ansible/blob/97cf14a03d2b392e2355478c74882d7be9f653d2/playbooks/common/openshift-cluster/initialize_oo_option_facts.yml#L20 and https://github.com/openshift/openshift-ansible/blob/bc5eb65f9bcd4147130e062ef2e4499145811c90/roles/openshift_docker_facts/tasks/main.yml#L29
the latter is docker.log_driver - I'm not sure where/how this is set, but there is a reference to it in the facts module https://github.com/openshift/openshift-ansible/blob/bc5eb65f9bcd4147130e062ef2e4499145811c90/roles/openshift_facts/library/openshift_facts.py#L65

ashcrow:
@richm overall 👍 but one worry ... if ``/etc/docker/daemon.json`` doesn't exist (like if they are using an older version of the docker package) will the mount cause problems? Or will it mount as null?

richm:
@ashcrow good question - I don't know - In my other patch for fluentd, I do something like this to check:
https://github.com/richm/origin-aggregated-logging/commit/36d335571b5c6b2f302d0a2d24921ba26fee630e
```
+    # if "log-driver" is set in /etc/docker/daemon.json, assume that it is
+    # authoritative
+    # otherwise, look for /etc/sysconfig/docker
+    # also note the unintuitive logic - in this case, a 0 return means true, and a 1
+    # return means false
+    if grep -q '^[^#].*"log-driver":' /etc/docker/daemon.json 2> /dev/null ; then
+        if grep -q '"log-driver":.*journald' /etc/docker/daemon.json 2> /dev/null ; then
+            return 0
+        fi
+    elif grep -q "^OPTIONS='[^']*--log-driver=journald" /etc/sysconfig/docker 2> /dev/null ; then
+        return 0
+    fi
+    return 1 
```
In this case, if /etc/docker/daemon.json doesn't exist, or is not readable, or doesn't specify log-driver: journald, it will fall through to look in /etc/sysconfig/docker.

ashcrow:
> In this case, if /etc/docker/daemon.json doesn't exist, or is not readable, or doesn't specify log-driver: journald, it will fall through to look in /etc/sysconfig/docker.

👍 to the test.

The only other idea I have would be to mount /etc/docker/ instead of the file directly as /etc/docker should always exist.

richm:
trying something else suggested by vgopal

roles/openshift_logging/defaults/main.yml
ewolinetz: So it looks like we can use `openshift.docker.log_driver` ?

On the other hand, the less we need to rely on `openshift_facts` the better. I would just be worried about `docker_log_driver` being set incorrectly compared to what's actually being used.
I think the risk of that is likely minimal though... so LGTM as is

richm:
never mind.  looks like using the docker api/socket is a non-starter.  Can we please merge this PR ASAP so we can un-break logging CI?

ashcrow:
aos-ci-test

richm:
With the latest commit, I actually got logging CI to pass - so please merge this PR ASAP.

roles/openshift_logging/defaults/main.yml
```diff
@@ -72,7 +72,7 @@ openshift_logging_fluentd_nodeselector: "{{ openshift_hosted_logging_fluentd_nod
 openshift_logging_fluentd_cpu_limit: 100m
 openshift_logging_fluentd_memory_limit: 512Mi
 openshift_logging_fluentd_es_copy: false
-openshift_logging_fluentd_use_journal: "{{ openshift_hosted_logging_use_journal | default('') }}"
+openshift_logging_fluentd_use_journal: "{{ openshift_hosted_logging_use_journal if openshift_hosted_logging_use_journal is defined else (docker_log_driver == 'journald') | ternary(True, False) if docker_log_driver is defined else (openshift.docker.log_driver == 'journald') | ternary(True, False) if openshift.docker.log_driver is defined else openshift.docker.options | search('--log-driver=journald') if openshift.docker.options is defined else default(omit) }}"
```
jcantrill: ugly.  I'm not suggesting changing it for this PR but maybe we could replace this with a filter.

sdodson:
[merge]

richm:
@sdodson I don't understand the merge failure - is this a flake?  If so, can you [merge] again?

ashcrow:
[merge]

ashcrow:
@richm I'm not sure either but I did a remerge command in hopes it passes. @sdodson ^^

richm:
@ashcrow Thanks!

sdodson:
flake on https://github.com/openshift/origin/issues/11114

sdodson:
It's still 4th in the merge queue so it'll be a while.

sdodson:
aos-ci-test

richm:
> error: "aos-ci-jenkins/OS_3.5_containerized, aos-ci-jenkins/OS_3.5_containerized_e2e_tests" for 02a5df1 (logs)

another flake?

sdodson:
The logs say it's good but the status says it's bad... w-t-f

```
[EnvInject] - Injecting environment variables from a build step.
[EnvInject] - Injecting as environment variables the properties content 
GH_STATE=success
GH_DESC="e2e tests passed"
GH_CONTEXT=aos-ci-jenkins/OS_3.5_containerized_e2e_tests
```

PR #4145, opened 2017-05-10
Parameterized Calico/Node Arguments

VincentS:
Parameterized Calico/Node configuration to be able to change selected arguments in the Ansible Inventory and documented these arguments in the README files of the Calico Roles.

### DIFF
--- a/roles/calico/defaults/main.yaml
+++ b/roles/calico/defaults/main.yaml
@@ -12,3 +12,9 @@ calico_etcd_key_file: "/etc/origin/calico/calico.etcd-client.key"
 
 calico_url_cni: "https://github.com/projectcalico/cni-plugin/releases/download/v1.5.5/calico"
 calico_url_ipam: "https://github.com/projectcalico/cni-plugin/releases/download/v1.5.5/calico-ipam"
+
+calico_ipv4pool_ipip: "always"
+calico_ipv4pool_cidr: "192.168.0.0/16"
+
+calico_log_dir: "/var/log/calico"
+calico_node_image: "calico/node:v1.1.0"
--- a/roles/calico/templates/calico.service.j2
+++ b/roles/calico/templates/calico.service.j2
@@ -10,7 +10,8 @@ ExecStart=/usr/bin/docker run --net=host --privileged \
  --name=calico-node \
  -e WAIT_FOR_DATASTORE=true \
  -e FELIX_DEFAULTENDPOINTTOHOSTACTION=ACCEPT \
- -e CALICO_IPV4POOL_IPIP=always \
+ -e CALICO_IPV4POOL_IPIP={{ calico_ipv4pool_ipip }} \
+ -e CALICO_IPV4POOL_CIDR={{ calico_ipv4pool_cidr }} \
  -e FELIX_IPV6SUPPORT=false \
  -e ETCD_ENDPOINTS={{ etcd_endpoints }} \
  -v /etc/origin/calico:/etc/origin/calico \
@@ -18,10 +19,11 @@ ExecStart=/usr/bin/docker run --net=host --privileged \
  -e ETCD_CERT_FILE={{ calico_etcd_cert_file }} \
  -e ETCD_KEY_FILE={{ calico_etcd_key_file }} \
  -e NODENAME={{ openshift.common.hostname }} \
- -v /var/log/calico:/var/log/calico \
+ -v {{ calico_log_dir }}:/var/log/calico\
  -v /lib/modules:/lib/modules \
  -v /var/run/calico:/var/run/calico \
- calico/node:v1.1.0
+ {{ calico_node_image }}
+
 
 ExecStop=-/usr/bin/docker stop calico-node
 
--- a/roles/calico/README.md
+++ b/roles/calico/README.md
@@ -20,6 +20,15 @@ To install, set the following inventory configuration parameters:
 * `openshift_use_openshift_sdn=False`
 * `os_sdn_network_plugin_name='cni'`
 
+## Additional Calico/Node and Felix Configuration Options
+
+Additional parameters that can be defined in the inventory are:
+
+| Environment | Description | Schema | Default |   
+|---------|----------------------|---------|---------|
+|CALICO_IPV4POOL_CIDR|	The IPv4 Pool to create if none exists at start up. It is invalid to define this variable and NO_DEFAULT_POOLS.	|IPv4 CIDR	| 192.168.0.0/16 |
+| CALICO_IPV4POOL_IPIP | IPIP Mode to use for the IPv4 POOL created at start up.	| off, always, cross-subnet	| always |
+| CALICO_LOG_DIR | Directory on the host machine where Calico Logs are written.| String	| /var/log/calico |
 
 ### Contact Information
 
--- a/roles/calico_master/README.md
+++ b/roles/calico_master/README.md
@@ -21,6 +21,18 @@ To install, set the following inventory configuration parameters:
 * `os_sdn_network_plugin_name='cni'`
 
 
+
+## Additional Calico/Node and Felix Configuration Options
+
+Additional parameters that can be defined in the inventory are:
+
+
+| Environment | Description | Schema | Default |   
+|---------|----------------------|---------|---------|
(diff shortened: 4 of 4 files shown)


djosborne:
Could someone more familiar with OpenShift comment on whether or not it'd be sensible to get this value from the existing `osm_cluster_network_cidr` field?
https://docs.openshift.org/latest/install_config/install/advanced_install.html#configuring-cluster-variables
It sounds like though it is currently only used by OpenShift SDN, it should be readable from the Calico playbook role

roles/calico/templates/calico.service.j2
```diff
@@ -10,8 +10,12 @@ ExecStart=/usr/bin/docker run --net=host --privileged \
...
  -e WAIT_FOR_DATASTORE=true \
  -e FELIX_DEFAULTENDPOINTTOHOSTACTION=ACCEPT \
- -e CALICO_IPV4POOL_IPIP=always \
- -e FELIX_IPV6SUPPORT=false \
+ -e CALICO_IPV4POOL_IPIP={{ calico_ipv4pool_ipip }} \
+ -e CALICO_IPV4POOL_CIDR={{ calico_ipv4pool_cidr }} \
+ {% if felix_ipv6support is not defined or true %}
+ -e CALICO_IPV6POOL_CIDR={{ calico_ipv6pool_cidr }} \
+ {% endif %}
+ -e FELIX_IPV6SUPPORT={{ felix_ipv6support }} \
```
ingvagabund: What will happen when felix_ipv6support is not defined? Should not it be:
```
{% if felix_ipv6support is not defined or true %}
-e CALICO_IPV6POOL_CIDR={{ calico_ipv6pool_cidr }} \
{% else %}
-e FELIX_IPV6SUPPORT={{ felix_ipv6support }} \
{% endif %}
```
instead? Or set `IPV6SUPPORT` to an empty string? I.e. ``-e FELIX_IPV6SUPPORT={{ felix_ipv6support | default("") }} \``?

roles/calico/templates/calico.service.j2
```diff
@@ -10,8 +10,12 @@ ExecStart=/usr/bin/docker run --net=host --privileged \
  --name=calico-node \
  -e WAIT_FOR_DATASTORE=true \
  -e FELIX_DEFAULTENDPOINTTOHOSTACTION=ACCEPT \
- -e CALICO_IPV4POOL_IPIP=always \
- -e FELIX_IPV6SUPPORT=false \
+ -e CALICO_IPV4POOL_IPIP={{ calico_ipv4pool_ipip }} \
+ -e CALICO_IPV4POOL_CIDR={{ calico_ipv4pool_cidr }} \
```
ingvagabund: Maybe
```
{% if calico_ipv4pool_cidr is defined %}
-e CALICO_IPV4POOL_CIDR={{ calico_ipv4pool_cidr }} \
{% endif %}
```

and comment the `calico_ipv4pool_cidr: "192.168.0.0/16"` in the `main.yml`?

roles/calico/defaults/main.yaml
```diff
@@ -8,3 +8,10 @@ cni_bin_dir: "/opt/cni/bin/"
 calico_etcd_ca_cert_file: "/etc/origin/calico/calico.etcd-ca.crt"
 calico_etcd_cert_file: "/etc/origin/calico/calico.etcd-client.crt"
 calico_etcd_key_file: "/etc/origin/calico/calico.etcd-client.key"
+
+
+calico_ipv6pool_cidr: "fd80:24e2:f998:72d6::/64"
+calico_ipv4pool_ipip: "always"
+calico_ipv4pool_cidr: "192.168.0.0/16"
+
+felix_ipv6support: "false"
```
ingvagabund: Could be worth extending the main.yml with

```yaml
calico_log_dir: "/var/log/calico"
calico_var_dir: "/var/run/calico"
calico_image: "calico/node:v1.1.0"
modules_dir: "/lib/modules"
```

and updating the `calico.service.j1` file.

roles/calico/templates/calico.service.j2
djosborne: This shouldn't even be an option. Neither Kubernetes nor OpenShift support IPv6 for pods.

djosborne:
WRT `osm_cluster_network_cidr`, I spoke with @VincentS  offline and concluded that we'll hold off on reading that value and use a calico namespaced variable instead to follow suit with the other networking plugins and keep a finer line of separation between Calico and those other plugins.

VincentS:
Changed and rebased pull request based on feedback

roles/calico/templates/calico.service.j2
```diff
@@ -10,18 +10,22 @@ ExecStart=/usr/bin/docker run --net=host --privileged \
  --name=calico-node \
  -e WAIT_FOR_DATASTORE=true \
  -e FELIX_DEFAULTENDPOINTTOHOSTACTION=ACCEPT \
- -e CALICO_IPV4POOL_IPIP=always \
- -e FELIX_IPV6SUPPORT=false \
+ -e CALICO_IPV4POOL_IPIP={{ calico_ipv4pool_ipip }} \
+ {% if calico_ipv4pool_cidr is defined %}
```
ozdanborne: Isn't `calico_ipv4pool_cidr` always defined since it's declared as a default var? Shouldn't we just nix the `{% if calico_ipv4pool_cidr is defined %}` and always set `-e CALICO_IPV4POOL_CIDR={{ calico_ipv4pool_cidr }}`?

roles/calico/templates/calico.service.j2
```diff
@@ -10,18 +10,22 @@ ExecStart=/usr/bin/docker run --net=host --privileged \
...
  -v /etc/origin/calico:/etc/origin/calico \
  -e ETCD_CA_CERT_FILE={{ calico_etcd_ca_cert_file }} \
  -e ETCD_CERT_FILE={{ calico_etcd_cert_file }} \
  -e ETCD_KEY_FILE={{ calico_etcd_key_file }} \
  -e NODENAME={{ openshift.common.hostname }} \
- -v /var/log/calico:/var/log/calico \
- -v /lib/modules:/lib/modules \
- -v /var/run/calico:/var/run/calico \
- calico/node:v1.1.0
+ -v {{ calico_log_dir }}:{{ calico_log_dir }} \
```
ozdanborne: The right side of this volume mount (the container's log dir) needs to remain `/var/log/calico` as that is where the container will dump logs.

roles/calico/templates/calico.service.j2
```diff
@@ -10,18 +10,22 @@ ExecStart=/usr/bin/docker run --net=host --privileged \
...
+ -v {{ calico_modules_dir }}:{{ calico_modules_dir }} \
```
ozdanborne: I don't think this needs to be templated. Can we leave it out in favor of simplicity?

roles/calico/templates/calico.service.j2
```diff
@@ -10,18 +10,22 @@ ExecStart=/usr/bin/docker run --net=host --privileged \
...
+ -v {{ calico_var_dir }}:{{ calico_var_dir }} \
```
ozdanborne: Can we leave this out as well? I don't think this usually needs to be changed by users...

roles/calico/templates/calico.service.j2
ozdanborne: I would be interested in hearing if you have personally encountered a use case where you need to change this though

VincentS:
WRT I discussed with @ozdanborne offline the arguments that should be exposed. The current PR is the result of this discussion.

ingvagabund:
aos-ci-test

ingvagabund:
[merge]

roles/calico/defaults/main.yaml
```diff
@@ -12,3 +12,9 @@ calico_etcd_key_file: "/etc/origin/calico/calico.etcd-client.key"
 
 calico_url_cni: "https://github.com/projectcalico/cni-plugin/releases/download/v1.5.5/calico"
 calico_url_ipam: "https://github.com/projectcalico/cni-plugin/releases/download/v1.5.5/calico-ipam"
+
+calico_ipv4pool_ipip: "always"
+calico_ipv4pool_cidr: "192.168.0.0/16"
+
+calico_log_dir: "/var/log/calico"
+calico_image: "calico/node:v1.1.0"
```
ozdanborne: Can we do `calico_node_image`? We may offer the option to change image names for other components in the future, so this would more clearly leave space for it

ingvagabund:
aos-ci-test

ingvagabund:
[merge]

PR #4151, opened 2017-05-10
1.5 - Updating registry-console image version as part of post_control_plane…

ewolinetz:
… upgrade

will need to also pick this for master and 1.4 since we can install when `openshift.common.version_gte_3_3_or_1_3`

Should address https://bugzilla.redhat.com/show_bug.cgi?id=1421987


ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

PR #4152, opened 2017-05-10
Default to iptables on master

sdodson:
We did this in 3.5 but never on master but we never came back to add
migration support. So we'll revert this on master and if/when we add
migration support we'll switch the default.

Forward ports #3849


sdodson:
aos-ci-test

sdodson:
[merge]

PR #4155, opened 2017-05-10
3.6 -Updating registry-console image version during a post_control_plane u…

ewolinetz:
…pgrade

Should address https://bugzilla.redhat.com/show_bug.cgi?id=1421987

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
+++ b/playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
@@ -9,6 +9,8 @@
                          replace ( '${version}', openshift_image_tag ) }}"
     router_image: "{{ openshift.master.registry_url | replace( '${component}', 'haproxy-router' ) |
                       replace ( '${version}', openshift_image_tag ) }}"
+    registry_console_image: "{{ openshift.master.registry_url | replace ( '${component}', 'registry-console') |
+                                replace ( '${version}', openshift.common.short_version ) }}"
 
   pre_tasks:
   - name: Load lib_openshift modules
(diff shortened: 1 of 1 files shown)


abutcher:
I don't think `openshift_image_tag` will cover the images we currently cut for `registry-console`. E.g. `v3.5.5.8` /cc @sdodson 

```
$ curl -sk http://registry.access.redhat.com/v1/repositories/openshift3/registry-console/tags | python -m json.tool | grep \" | cut -d \" -f2
3.3
3.3-1
3.3-10
3.3-13
3.3-3
3.3-4
3.3-6
3.3-8
3.4
3.4-2
3.4-4
3.4-6
3.4-8
3.5
3.5-5
3.5-7
latest
v3.3
v3.4
v3.5
```

ewolinetz:
If we aren't going to cut tags to match for those images I can update it to just use `X.Y` (for 3.4-3.6)

sdodson:
> If we aren't going to cut tags to match for those images I can update it to just use X.Y (for 3.4-3.6)

I think that's what we need to do for OCP, for origin there appear to be no other tags than 'latest'

https://hub.docker.com/r/cockpit/kubernetes/tags/

ewolinetz:
I could probably use `openshift.common.short_version` that would give me `3.5` even if the `openshift_image_tag` will be `3.5.5.8`

ewolinetz:
updated to use the short_version

sdodson:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/post_control_plane.yml
```diff
@@ -61,6 +63,23 @@
...
+  - name: Update registry-console image to current version
+    oc_edit:
+      kind: dc
+      name: registry-console
+      namespace: default
+      content:
+        spec.template.spec.containers[0].image: "{{ registry_console_image }}"
+    when:
+    - _registry_console.results.results[0] != {}
+
```
abutcher: Since the origin image is always latest (`image: cockpit/kubernetes:latest`) we won't have to update the dc when `openshift.common.deployment_type != 'origin'`.

ewolinetz:
aos-ci-test

sdodson:
[merge]

PR #4156, opened 2017-05-10
Fix for yedit custom separators

mwoodson:


mwoodson:
Yedit custom separators were broken.  This happened during this commit:

https://github.com/openshift/openshift-ansible/commit/b7af36569f4f02a6320833759b48e15df25a9b06

I'm hoping @kwoodson can comment.

It appears the change from old python way of formatting (%s) was updated to ".format".  In doing so, the string (%s) was never removed and changed to the new way that .format uses ({}).

```
-    re_key = r"(?:\[(-?\d+)\])|([0-9a-zA-Z%s/_-]+)"
￼+    re_key = r"(?:\[(-?\d+)\])|([0-9a-zA-Z{}/_-]+)"
```

I changed this in yedit, the re-generated all of the yedit and lib_openshift.

If this gets approved, we need a backport to at least 3.6.  I don't know if this is in 3.5.

ashcrow:
aos-ci-test

sdodson:
[merge]

sdodson:
@mwoodson can you open against release-1.5 too if you need it there? just patch the src file and regenerate then push that up as there's probably other changes missing from release-1.5

PR #4160, opened 2017-05-11
[1.5] Fix Origin version variant

mtnbikenc:
Variants were previously corrected for openshift-enterprise and Registry installs, however, the Origin version variant was skipped.  This PR correct the Origin install version variant.


mtnbikenc:
aos-ci-test

PR #4166, opened 2017-05-11
Use v3.5 OCP images on the release-1.5 branch

sdodson:
For origin we have no generic tags that point at 1.5, 3.6, so just keep
using latest there


sdodson:
aos-ci-test

ingvagabund:
[merge]

PR #4168, opened 2017-05-12
Use v3.4 OCP images on the release-1.4 branch

sdodson:
For origin we have no generic tags that point at 1.5, 3.6, so just keep
using latest there


ingvagabund:
aos-ci-test

PR #4170, opened 2017-05-12
WIP: introduce etcd_upgrade role

ingvagabund:
TODO:
* [x] - encompass all etcd upgrade stuff there is in etcd_upgrade
* [x] - move backup machinery into the role as well
* [ ] - refactor etcd_common and etcd so they can be used with etcd_upgrade
* [ ] - move some vars from defaults/main.yml to vars/main.yml
* [ ] - merge etcd_upgrade and etcd roles

QUESTIONS:
* is `etcd_client_port` public variable? See `playbooks/common/openshift-master/config.yml:28`
* not sure about `etcd_hostname` and `etcd_ip` vars vs. openshift_etcd_facts's ones

POTENTIAL PUBLIC VARS:
* ``etcd_client_port``
* ``etcd_version``

RELATED PRs:
* Move etcd upgrade code into role: https://github.com/openshift/openshift-ansible/pull/4217

DO-NOT-MERGE


ingvagabund:
aos-ci-test

ewolinetz:
This seems like its adding an extra role that is too specific. Would it be possible to add this functionality instead to the `etcd` role so that if we see that we are going from an earlier version to a later one we would do the upgrade steps? (e.g. backup, migrate, etc)

ingvagabund:
> This seems like its adding an extra role that is too specific. Would it be possible to add this functionality instead to the etcd role so that if we see that we are going from an earlier version to a later one we would do the upgrade steps? (e.g. backup, migrate, etc)

Agree on that, though, etcd role refactoring will take more time. I will get to it at some point.

ingvagabund:
@ewolinetz getting a headache after renaming etcd like varibles into `r_etcd_common_*` and `r_etcd_*` like names. I will first prepare the new etdc_upgrade role. Once it is merged, I will merge it with etcd. This will take some time to settle down all changes. So to minimize the friction, it is better to do it in more PRs. At the same time I don't want anyone to blow his head reviewing > 100 files.

michaelgugino:
Closing due to inactivity.

PR #4114, opened 2017-05-08
Moving Dockerfile content to images dir

jupierce:
In order to reduce extraneous files in dist-git, it is important to keep image building contributions out of the root.

@sdodson @tdawson

PR #4175, opened 2017-05-12
Fix container image build references

codificat:
In #4114 the location of the Dockerfiles changed. Build-related scripts and documentation need some updates accordingly.

The "worst part" is the section about building the image on OpenShift - it seems there's no easy way to have `oc new-app` look for a Dockerfile under a specific path (i.e. set `dockerfilePath`), which makes things a bit clumsy.

Also a couple of build-related updates:
  - playbook2image is now available from the openshift org, so update the references
  - downstream image is building with boto 2.34 (current EL7 version) so use that version upstream too.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 # Versions are pinned to prevent pypi releases arbitrarily breaking
 # tests with new APIs/semantics. We want to update versions deliberately.
 ansible==2.2.2.0
-boto==2.45.0
+boto==2.34.0
 click==6.7
 pyOpenSSL==16.2.0
 # We need to disable ruamel.yaml for now because of test failures
--- a/BUILD.md
+++ b/BUILD.md
@@ -29,17 +29,34 @@ tito build --rpm
 To build a container image of `openshift-ansible` using standalone **Docker**:
 
         cd openshift-ansible
-        docker build -t openshift/openshift-ansible .
+        docker build -f images/installer/Dockerfile -t openshift/openshift-ansible .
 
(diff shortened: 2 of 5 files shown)


sosiouxme:
aos-ci-test

requirements.txt
```diff
@@ -1,7 +1,7 @@
 # Versions are pinned to prevent pypi releases arbitrarily breaking
 # tests with new APIs/semantics. We want to update versions deliberately.
 ansible==2.2.2.0
-boto==2.45.0
+boto==2.34.0
```
sosiouxme: Is this change supposed to be here or an accident? Not sure why it would be associated with this.

sdodson:
aos-ci-test

sosiouxme:
aos-ci-test

requirements.txt
codificat: It was intentional... I tried to explain it (briefly) in the PR description: I initially picked 2.45.0 just because it was what I had in the env where I was smoke-testing this. However, 2.34 is the version that is used in RHEL7 / will be used in the downstream image, which means it should be a better candidate version than "what Pep had around" ;)

I'm not aware of a policy/way to pick and maintain the versions pinned here, so if there's a better approach please let me know.

sosiouxme: Oh, sorry, I completely forgot that by the time I got to the end of the code. I don't know of any particular policy either so your approach seems sensible.

sosiouxme:
[merge]

codificat:
The failure was caused by yamllint errors that should have now been fixed via c3f96b1f9f77cbd105d082a216b4761ca040bcf7, so let's try to [merge] again

sdodson:
[merge]

sdodson:
[merge]

PR #4176, opened 2017-05-12
Rework openshift_excluders role

mtnbikenc:
The structure and usage of the `openshift_exlcuder` role has been completely reworked.

- The role has been converted to run as a standard role with a `tasks/main.yml` entry-point
- Playbook usage of the role has been converted from an `include_role:` task to a `role:` role
- Role variables have been standardized on `r_<role_name>_`
- Role variable are now passed in when calling the role
- Role tasks to be performed are now controlled by an `_action` variable
- README.md has been revamped
-- Specifies Inventory Variables consumed by the role
-- Documents role variables
-- Provides example playbook for all supported options
- Removed `openshift_facts` and `openshift_repos` role dependencies
- Added `name:` to all tasks
- Role can be run directly without any other playbooks/roles being run first, provided that the variables in the Example Playbook are set properly


mtnbikenc:
aos-ci-test

mtnbikenc:
aos-ci-test

mtnbikenc:
aos-ci-test

roles/openshift_excluder/defaults/main.yml
```diff
@@ -1,6 +1,15 @@
...
-openshift_excluder_package_state: present
-docker_excluder_package_state: present
+r_openshift_excluder_package_state: present
+r_openshift_excluder_docker_package_state: present
 
-enable_excluders: true
+# Inventory variables
+r_openshift_excluder_enable_excluders: "{{ openshift_enable_excluders | default(true) }}"
+r_openshift_excluder_enable_openshift_excluder: "{{ openshfit_enable_openshift_excluder | default(r_openshift_excluder_enable_excluders) }}"
+r_openshift_excluder_enable_docker_excluder: "{{ openshift_enable_docker_excluder | default(r_openshift_excluder_enable_excluders) }}"
```
sdodson: Need to preserve backwards compat with the old variables, can we add that here via additional defaulting?

sdodson:
aos-ci-test

roles/openshift_excluder/defaults/main.yml
mtnbikenc: I'll update to include the legacy variables.

mtnbikenc:
aos-ci-test

mtnbikenc:
@sdodson Added the legacy inventory variables to defaults/main.yml  PTAL

playbooks/common/openshift-cluster/config.yml
```diff
@@ -8,10 +8,10 @@
...
   - always
   gather_facts: no
-  tasks:
-  - include_role:
-      name: openshift_excluder
-      tasks_from: disable
+  roles:
+  - role: openshift_excluder
+    r_openshift_excluder_action: disable
+    r_openshift_excluder_service_type: "{{ openshift.common.service_type }}"
```
mwoodson: I would like to discuss include_role vs roles with the team.  include_role seems much more useful, to me, than role. I'd like to further this discussion.

roles/openshift_excluder/README.md
```diff
@@ -1,50 +1,69 @@
...
+==================
 
 Manages the excluder packages which add yum and dnf exclusions ensuring that
-the packages we care about are not inadvertantly updated. See
+the packages we care about are not inadvertently updated. See
 https://github.com/openshift/origin/tree/master/contrib/excluder
 
 Requirements
 ------------
-openshift_facts
```
mwoodson: +1

roles/openshift_excluder/tasks/disable.yml
```diff
@@ -1,47 +1,38 @@
 ---
-# input variables
-# - excluder_package_state
-# - docker_excluder_package_state
-- include: init.yml
+- when: r_openshift_excluder_verify_upgrade
```
mwoodson: This is stylistic, but i like how you put when at the top, instead of the bottom.  It's much clearer to me when the block should be run.

sdodson:
aos-ci-test

mtnbikenc:
Why do the job logs say `success` but the bot comments say `error`? :(

ingvagabund:
aos-ci-test

playbooks/common/openshift-cluster/config.yml
ingvagabund: It depends on where the role is used. I am not against combining both where reasonable. Still, I would incline to use role instead or include_role. With the role there is only one place where it can be used so it is easy to find it. Include_role can be basically used anywhere inside `tasks`. (`grep` will discover that but still the code should be self-obvious). We should definitely talk about using (or not using) `include_role` inside a role.

playbooks/common/openshift-cluster/config.yml
```diff
@@ -50,7 +50,7 @@
...
   - always
   gather_facts: no
-  tasks:
-  - include_role:
-      name: openshift_excluder
-      tasks_from: enable
+  roles:
+  - role: openshift_excluder
+    r_openshift_excluder_action: enable
+    r_openshift_excluder_service_type: "{{ openshift.common.service_type }}"
```
ingvagabund: Whenever we pull ``opensihft.common`` out of any role, it is a good sign.

roles/openshift_excluder/README.md
```diff
@@ -1,50 +1,69 @@
...
-| enable_excluders           | None             | Enable all excluders
+| Name                                 | Default Value              | Description                            |
+---------------------------------------|----------------------------|----------------------------------------|
+| openshift_enable_excluders           | True                       | Enable all excluders                   |
+| openshift_enable_docker_excluder     | openshift_enable_excluders | Enable docker excluder. If not set, the docker excluder is ignored. |
+| openshift_enable_openshift_excluder  | openshift_enable_excluders | Enable openshift excluder. If not set, the openshift excluder is ignored. |
 
 Role Variables
 --------------
-None
```
ingvagabund: I would like to see this table generated based on variables defined in `defaults/main.yml`

roles/openshift_excluder/meta/main.yml
```diff
@@ -12,6 +12,4 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
-- { role: openshift_facts }
-- { role: openshift_repos }
```
ingvagabund: IIRC, `openshift_repos` dependency was introduced due to CentOS

roles/openshift_excluder/tasks/disable.yml
```diff
@@ -1,47 +1,38 @@
...
-  when:
-  - not openshift.common.is_atomic | bool
+    unexclude_openshift_excluder: "{{ r_openshift_excluder_enable_openshift_excluder }}"
 
 # Install any excluder that is enabled
-- include: install.yml
-  vars:
-    # Both docker_excluder_on and openshift_excluder_on are set in openshift_excluder->init task
-    install_docker_excluder: "{{ docker_excluder_on | bool }}"
-    install_openshift_excluder: "{{ openshift_excluder_on | bool }}"
```
ingvagabund: The idea of introducing `install_openshift_excluder` was to be able to call `install.yml` out of the excluders role. However, given that is not going to happen (unless new entry points of the role are introduced), I am fine with the change.

roles/openshift_excluder/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+      msg: "r_openshift_excluder_upgrade_target must be provided when using this role for upgrades"
+    when:
+    - r_openshift_excluder_verify_upgrade | bool
+    - r_openshift_excluder_upgrade_target is not defined
+
+  - name: Include main action task file
+    include: "{{ r_openshift_excluder_action }}.yml"
+
+  when:
+  - not ostree_booted.stat.exists | bool
```
ingvagabund: +1 for removing `openshift.common.is_atomic`

ingvagabund:
aos-ci-test

mtnbikenc:
The bot has reported `error:` 16 times.  I'm at a loss on how to get this through CI.

From the last error [logs](https://aos-ci.s3.amazonaws.com/openshift/openshift-ansible/jenkins-openshift-ansible-test-matrix-CONTAINERIZED=_NOT_containerized,OSE_VER=3.5,PYTHON=System-CPython-2.7,TOPOLOGY=openshift-cluster,TargetBranch=master,nodes=openshift-ansible-slave-380/c5c222eddb0b8950995105c4c406f5a5d0bfbacd.txt).

```
GH_STATE=error
GH_DESC="Internal Jenkins error during provisioning an 3.5 cluster"
GH_CONTEXT=aos-ci-jenkins/OS_3.5_NOT_containerized
...
GH_STATE=error
GH_DESC="Error prepping the 3.5 cluster"
GH_CONTEXT=aos-ci-jenkins/OS_3.5_NOT_containerized
...
GH_STATE=error
GH_DESC="openshift-ansible install failed"
GH_CONTEXT=aos-ci-jenkins/OS_3.5_NOT_containerized
```

ashcrow:
@mtnbikenc FWIW I've seen that error happen a lot over the last 3 days ☹️ It's a failure and not related to the codebase, but not a "flake" in how we define it.

playbooks/common/openshift-cluster/config.yml
twiest: +1 to include_role. Let's discuss today in the meeting.

playbooks/common/openshift-cluster/config.yml
twiest: +1

roles/openshift_excluder/README.md
twiest: Should the role care about / advertise the inventory variables?

It seems like it should only care about / advertise it's own role variables since that's how you interact with the role.

roles/openshift_excluder/tasks/disable.yml
twiest: Yeah, that's a really interesting way of getting the "if <condition>" feel in ansible.

+1

roles/openshift_excluder/tasks/exclude.yml
```diff
@@ -1,30 +1,22 @@
...
-    stat:
-      path: /sbin/{{ openshift.common.service_type }}-docker-excluder
-    register: docker_excluder_stat
-  - name: Enable docker excluder
-    command: "{{ openshift.common.service_type }}-docker-excluder exclude"
-    when:
-    - exclude_docker_excluder | default(false) | bool
-    - docker_excluder_stat.stat.exists
+- name: Enable docker excluder
+  command: "{{ r_openshift_excluder_service_type }}-docker-excluder exclude"
```
twiest: Is the intention to eventually write an ansible module for the excluder?

roles/openshift_excluder/README.md
mtnbikenc: I stopped short of requiring every variable to be defined when calling the role.  It was a trade-off between having everything being passed in and usability.

If we required everything to be passed in, it would look something like this, with three extra `_enable_` vars passed each time the role is called even though they are not necessarily providing value to what is being done.  We either put the inventory variables only in role defaults, or we put them in every role call.

```yaml
- name: Demonstrate OpenShift Excluder usage
  roles:
  # Disable all excluders
  - role: openshift_excluder
    r_openshift_excluder_enable_excluders: "{{ openshift_enable_excluders | default(enable_excluders) | default(true) }}"
    r_openshift_excluder_enable_openshift_excluder: "{{ openshift_enable_openshift_excluder | default(enable_openshift_excluder) | default(r_openshift_excluder_enable_excluders) }}"
    r_openshift_excluder_enable_docker_excluder: "{{ openshift_enable_docker_excluder | default(enable_docker_excluder) | default(r_openshift_excluder_enable_excluders) }}"
  # Enable all excluders
  - role: openshift_excluder
    r_openshift_excluder_action: enable
    r_openshift_excluder_enable_excluders: "{{ openshift_enable_excluders | default(enable_excluders) | default(true) }}"
    r_openshift_excluder_enable_openshift_excluder: "{{ openshift_enable_openshift_excluder | default(enable_openshift_excluder) | default(r_openshift_excluder_enable_excluders) }}"
    r_openshift_excluder_enable_docker_excluder: "{{ openshift_enable_docker_excluder | default(enable_docker_excluder) | default(r_openshift_excluder_enable_excluders) }}"
  # Disable all excluders and verify appropriate excluder packages are available for upgrade
  - role: openshift_excluder
    r_openshift_excluder_enable_excluders: "{{ openshift_enable_excluders | default(enable_excluders) | default(true) }}"
    r_openshift_excluder_enable_openshift_excluder: "{{ openshift_enable_openshift_excluder | default(enable_openshift_excluder) | default(r_openshift_excluder_enable_excluders) }}"
    r_openshift_excluder_enable_docker_excluder: "{{ openshift_enable_docker_excluder | default(enable_docker_excluder) | default(r_openshift_excluder_enable_excluders) }}"
    r_openshift_excluder_verify_upgrade: true
    r_openshift_excluder_upgrade_target: "{{ openshift_upgrade_target }}"
    r_openshift_excluder_package_state: latest
    r_openshift_excluder_docker_package_state: latest
```

mtnbikenc:
aos-ci-test

mtnbikenc:
[test]

mtnbikenc:
Flake https://github.com/openshift/origin/issues/13827

playbooks/common/openshift-cluster/config.yml
sdodson: I feel like every time we've reached for include_role we've ended up creating problems for ourselves like those @ingvagabund mentioned and @mtnbikenc has previously hi-lighted. I like calling the role with a control variable as proposed here.

roles/openshift_excluder/meta/main.yml
sdodson: Or any environment that makes use of `openshift_additional_repos`

roles/openshift_excluder/tasks/main.yml
ingvagabund: Just a nit for the future (not a blocker). `ostree_booted` is not a role variable (as a to-outside-visible-variable) but a local one. Pre-pending the var with `l_` will make it clear.

ingvagabund: Or maybe, if the local var is defined in file FILE.yml, it could be l_FILE_varname

roles/openshift_excluder/README.md
ingvagabund: Definitely vote for minimizing the number of arguments

mtnbikenc:
[merge]

roles/openshift_excluder/defaults/main.yml
```diff
@@ -1,6 +1,19 @@
...
+r_openshift_excluder_package_state: present
+r_openshift_excluder_docker_package_state: present
 
-enable_excluders: true
+# Legacy variables are included for backwards compatibility with v3.5
+# Inventory variables                   Legacy
+# openshift_enable_excluders            enable_excluders
+# openshift_enable_openshift_excluder   enable_openshift_excluder
+# openshift_enable_docker_excluder      enable_docker_excluder
+r_openshift_excluder_enable_excluders: "{{ openshift_enable_excluders | default(enable_excluders) | default(true) }}"
```
sosiouxme: Wait... does this actually work? I didn't think vars files could use jinja and this would just come out as literally what's in the quotes. Now I have to try it...

mtnbikenc: I was just confirming this, and it does.  I'm using master and can get true/false as desired through inventory vars.

mtnbikenc: Here is my test playbook:

```yaml
- name: Test Playbook
  hosts: masters[0]
  vars:
    openshift_enable_excluders: false
    # enable_excluders: false
    openshift:
      common:
        service_type: atomic-openshift

  roles:
    - role: openshift_excluder
      r_openshift_excluder_action: enable
```

sosiouxme: Yeah it works; I'll have to revisit my actual use case which involves an action plugin to see if that's different somehow.

roles/openshift_excluder/README.md
twiest: @mtnbikenc Anything not passed into the role is the role requiring global variables to be set.

So if I want to call this role outside of the installer / upgrader, I now have to populate my inventory to look like what this role is expecting.

This pattern is very analogous to global variables being used inside a function or class. If course it's easier, but it makes your API very difficult to work with by external consumers.

@sdodson Can we discuss this in the next openshift-ansible arch meeting please?  The pattern used here is really going against the intention of the "don't use global variables in roles" rule that we agreed to previously, so we clearly need more discussion and buy off from the team on it.

mtnbikenc: @twiest Thanks for the feedback.  This is why I opened the PR, to continue the discussion.  For this role to work, you don't have to put any special variables in your inventory, they are optional.  In this instance, we had a feature request of allowing a user to not enable excluders, and to support that we have a variable which gives the user control of when excluders would be enabled.  At some point, the user would need to provide that variable, whether it is done through their own group_vars, inventory, or at the command line.

I'm happy to continue iterating on this work until we have something that is agreeable by all concerned. :smile:

roles/openshift_excluder/defaults/main.yml
sdodson: One thing you may have run into is that it's just in time evaluation. set_fact has to be used to force immediate evaluation unless there's some better way.

sosiouxme: Yes that seems to be the issue. I put debug tasks before and after running the action plugin and they got the evaluated result, while the action plugin saw the jinja. There's probably some hook I can call in the plugin to trigger evaluation, will look into it.

PR #4181, opened 2017-05-12
Improve configurability of registry and registry-console hosts

dmsimard:
This adds missing docs on the registry host route variable that can be specified to customize the host on the registry route.
Additionally, it adds the ability to customize the registry-console route host which allows to get something other than the provided default.

### DIFF
--- a/roles/openshift_hosted/tasks/registry/secure.yml
+++ b/roles/openshift_hosted/tasks/registry/secure.yml
@@ -55,6 +55,7 @@
     - "{{ docker_registry_service_ip.results.clusterip }}"
     - "{{ openshift_hosted_registry_name }}.default.svc"
     - "{{ openshift_hosted_registry_name }}.default.svc.{{ openshift.common.dns_domain }}"
+    - "{{ openshift_hosted_registry_routehost | default(omit) }}"
     - "{{ docker_registry_route_hostname }}"
     cert: "{{ openshift_master_config_dir }}/registry.crt"
     key: "{{ openshift_master_config_dir }}/registry.key"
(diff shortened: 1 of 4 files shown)


dmsimard:
@abutcher as per discussed -- to improve the use case around standalone registry.

The SSL certificate bits are a bit more involved and I can't do it right now but it would definitely be something we'd need.

roles/openshift_hosted/tasks/registry/secure.yml
```diff
@@ -55,6 +55,7 @@
     - "{{ docker_registry_service_ip.results.clusterip }}"
     - "{{ openshift_hosted_registry_name }}.default.svc"
     - "{{ openshift_hosted_registry_name }}.default.svc.{{ openshift.common.dns_domain }}"
+    - "{{ openshift_hosted_registry_routehost | default(omit) }}"
```
abutcher: `omit` fails in the negative case since we expect each item in this list to be a string. My proposed fix is to filter non-strings from the list, PR here https://github.com/openshift/openshift-ansible/pull/4197

dmsimard:
Rebased my PR on top of master with https://github.com/openshift/openshift-ansible/pull/4197 which merged.

dmsimard:
bot, test pull request

abutcher:
aos-ci-test

abutcher:
aos-ci-test

dmsimard:
If you wait a little bit, I'll rework this PR with SSL support for registry-console - it's a work in progress.

abutcher:
@dmsimard copy that

dmsimard:
Osboleted by https://github.com/openshift/openshift-ansible/pull/4256 .. going to close this.

PR #4184, opened 2017-05-14
Add docker package for container-engine install

ashcrow:
The package is required so that openshift_facts is able to use the
docker client to gain information on the versions.


ashcrow:
aos-ci-test

ashcrow:
This should set the version in ``openshift_facts`` properly.

ashcrow:
[merge]

PR #4187, opened 2017-05-15
uninstall: handle container-engine

giuseppe:
Try to restart the docker service only when there is not already
a running container-engine service.

### DIFF
--- a/playbooks/adhoc/uninstall.yml
+++ b/playbooks/adhoc/uninstall.yml
@@ -305,8 +305,15 @@
   - shell: systemctl daemon-reload
     changed_when: False
 
+  - name: restart container-engine
+    service: name=container-engine state=restarted
+    ignore_errors: true
+    register: container_engine
+
   - name: restart docker
     service: name=docker state=restarted
+    ignore_errors: true
+    when: "container_engine.state != 'started'"
 
   - name: restart NetworkManager
     service: name=NetworkManager state=restarted


giuseppe:
aos-ci-test

ganhuang:
@giuseppe @ashcrow What about removing `container-engine` directly? I suspect there might some issues while the user needs to fall back to use rpm docker instead after running `uninstall` playbook against a docker system container environment. This is also what has been written in my test case.

ashcrow:
@ganhuang fair point. Though I'm not sure if we'd want to remove it as they may want to continue using containers inside container-engine. True, they could replace it with the docker package (if the versions we compatible) but that would be hairy to keep track of.

@giuseppe WDYT?

ganhuang:
@ashcrow Thanks, makes sense to me. I'll update the test case.

giuseppe:
I think we should not uninstall it completely, people could have installed directly with `atomic install --system ...` and not as part of openshift-ansible

ashcrow:
The error is unrelated to the code AFAICT though I couldn't find the reason for the failure.

```
GH_STATE=success
GH_DESC="e2e tests passed"
GH_CONTEXT=aos-ci-jenkins/OS_3.5_NOT_containerized_e2e_tests
```

sdodson:
aos-ci-test

ashcrow:
[merge]

PR #4173, opened 2017-05-12
Add no_proxy to atomic.conf

ashcrow:
Requires: https://github.com/projectatomic/atomic/pull/999

PR #4192, opened 2017-05-15
container-engine: Use dest instead of path for lineinfile

ashcrow:
path has been changed to dest to support pre 2.3 releases of ansible.
For more information see: http://docs.ansible.com/ansible/lineinfile_module.html


ashcrow:
aos-ci-test

ashcrow:
Moving this to https://github.com/openshift/openshift-ansible/pull/4173 since a rebase would be required anyway.

PR #4195, opened 2017-05-15
Creation of service_catalog and broker roles

ewolinetz:
Based on https://github.com/benjaminapetersen/origin-web-console/wiki/Catalog-&-broker-in-a-Vagrant-VM
these roles will be used to configure and install the Service Catalog and Broker and their required components on top of OCP.

FYI @jpeeler


ewolinetz:
Tested installing this on a 3.6 system (installed 3.5 and upgraded to 3.6) and the api server seems to keep restarting due to the following
```
I0519 20:16:08.932865       1 request.go:991] Response Body: {"kind":"ConfigMap","apiVersion":"v1","metadata":{"name":"extension-apiserver-authentication","namespace":"kube-system","selfLink":"/api/v1/namespaces/kube-system/configmaps/extension-apiserver-authentication","uid":"5ac5486a-3cce-11e7-803b-52540022ee11","resourceVersion":"2123","creationTimestamp":"2017-05-19T20:04:24Z"},"data":{"client-ca-file":"-----BEGIN CERTIFICATE-----\nMIIC6jCCAdKgAwIBAgIBATANBgkqhkiG9w0BAQsFADAmMSQwIgYDVQQDDBtvcGVu\nc2hpZnQtc2lnbmVyQDE0OTUyMTk4MjcwHhcNMTcwNTE5MTg1MDI3WhcNMjIwNTE4\nMTg1MDI4WjAmMSQwIgYDVQQDDBtvcGVuc2hpZnQtc2lnbmVyQDE0OTUyMTk4Mjcw\nggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQC9VQlF9/BmFDASUX4/wSFC\nbf94mTadivxBHtBvC4ZzWlj7ooxPEhjrlyaIfZAvmV2tQLVTIWsTFGgqoYAIcNbB\n2Z6C5dgv+XCbGTBcDdZUkRkCWbYNW8Ir6jBYYIG5fewA0rgzP54WEc9TRy1g5O8Y\nbXBUWeeiyc3uZMInr8emstQg8+RCGJJ4t0vJhZJqmo9lpaMVE6XQ0wMouQOlzlUu\ndxY8rjxfDtRO1Hg+FgNvpNy0UK4HTzHzE/6R5V4NEF05dtFDZacWV36++aYH0joo\n7uFelO3dqDa0U9UG7wkQPdV7y8mDX0zaxtYjRBrcJD6EXhFLsBGn+NOlMYLo/xOj\nAgMBAAGjIzAhMA4GA1UdDwEB/wQEAwICpDAPBgNVHRMBAf8EBTADAQH/MA0GCSqG\nSIb3DQEBCwUAA4IBAQCD4TG1YmTM72Qw4rlZ88hs+ZyBrKmIhqfXubs4Nr8tw98y\nG0EZCkwVbtY+KlAANEWEMsVJqG2QDhl/fuMuJvZvCvZi+ulugLPowEnXUeQXZ1dX\nxyAdKh6OEPLKf5olnOBiHbcCOFtBdQ0HKJSWJQaelcwp+llFex9TCoKr4H7s1OLr\nnTB8EMS2MbPN4TV4C34prIi5synI4u7daZUKpFNbVpgzjVcszWepZKZwA00n/YLO\nrkwhIeqxXgPldq2RtxbsLeW1pIL2dCFFhlxbYFL83mL7sdB95v1+rTGmIavsTT42\nkVv8goRL5QNt0Y6L1O/XQCCWCJTRuLoek1gqZ1Kf\n-----END CERTIFICATE-----\n"}}
F0519 20:16:08.933004       1 server.go:122] error running server (cluster doesn't provide requestheader-client-ca-file)
```

The manager is restarting because the apiserver is in a crash loop due to the above.
Otherwise this seems to install correctly, pending testing integrating with the web console

jpeeler:
@ewolinetz a few comments:
Because our version of the service catalog container is going to have all the binaries (rather than separate images upstream), I think the command will need to be passed in manually in the respective spec sections.

I pushed an unofficial image to docker hub that should resemble what the final image looks like at docker.io/jpeeler/origin-service-catalog:latest.

roles/openshift_service_catalog/tasks/install.yml
```diff
@@ -0,0 +1,142 @@
...
+  register: get_kube_system
+  changed_when: no
+
+- command: >
+    oc create policybinding kube-system -n kube-system
+  when: "'not found' in get_kube_system.stdout"
+
+# TODO: use the correct module here
+- command: >
+    oc adm policy add-scc-to-user privileged -z apiserver
```
pmorie: the 'hostpath' scc should be all that is required for the API server

roles/openshift_service_catalog/files/catalog.yaml
```diff
@@ -0,0 +1,186 @@
...
+        maxUnavailable: 1
+      type: RollingUpdate
+    template:
+      metadata:
+        labels:
+          app: apiserver
+      spec:
+        containers:
+        - args:
+          - --storage-type
```
pmorie: you should add the following lines here:

```
        - --admission-control
        - "KubernetesNamespaceLifecycle"
```

roles/openshift_service_catalog/files/catalog.yaml
```diff
@@ -0,0 +1,186 @@
...
+            protocol: TCP
+          - containerPort: 8081
+            protocol: TCP
+          resources: {}
+          terminationMessagePath: /dev/termination-log
+          volumeMounts:
+          - mountPath: /var/run/kubernetes-service-catalog
+            name: apiserver-ssl
+            readOnly: true
+        - env:
```
pmorie: If we use using the core's etcd, we do not need and should not have an etcd container in this pod.

ewolinetz: this file should be removed, its leftover and not being used...

playbooks/common/openshift-cluster/config.yml
```diff
@@ -45,6 +45,10 @@
   tags:
   - hosted
 
+- include: service_catalog.yml
```
ewolinetz: @sdodson like this?

sdodson: Yeah that's fine, it's easy to move around until we come out of tech preview anyway.
```
  when:
  - openshift_enable_service_catalog | default(false)
```

ewolinetz: should we use that var for opting in? currently I'm using the following in the roles:
```yaml
openshift_service_catalog_remove=True
openshift_service_catalog_broker_remove=True
```

I'm not sure if they are mutually exclusive or not,  @pmorie ?

pmorie: @ewolinetz I'm not sure that I understand the question

sdodson: I think in playbooks/common/openshift-cluster/config.yml we should optionally include the playbook or not.
I think if they're calling the service catalog playbook directly that's an intentional act and we'd have to assume they want to either install, remove, or upgrade components.

I'm also fine with completely foregoing inclusion in playbooks/common/openshift-cluster/config.yml until we come out of 'tech preview'

ewolinetz:
I had to reinstall my environment and now see the following:
```yaml
"the server doesn't have a resource type \"apiservices\" in group \"apiregistration.k8s.io\""
```
any thoughts @jpeeler @pmorie ?

ewolinetz:
i'm noticing something odd happening with my local env when installing the service catalog after... before i try to install `oc get ds` will return no resources found. but after i try i see that the server "doesn't have a resource type 'daemonsets'"

ewolinetz:
Still will need changes from https://github.com/ewolinetz/openshift-ansible/pull/26 for the ansible service broker changes
Otherwise @jpeeler @pmorie @sdodson @bparees  PTAL

roles/openshift_service_catalog/files/kubeservicecatalog_roles_bindings.yml
```diff
@@ -0,0 +1,214 @@
...
+- kind: ClusterRoleBinding
+  apiVersion: v1
+  metadata:
+    name: template-broker-caller-binding
+  roleRef:
+    name: system:openshift:templateservicebroker-client
+  groupNames:
+  - system:unauthenticated
+  - system:authenticated
+  - system:anonymous
```
bparees: you definitely do NOT want this role.  in general i think you're working off an old version of my template.  see the updated PR for the final iteration of the roles:
https://github.com/openshift/origin/pull/14630

roles/openshift_service_catalog/files/kubesystem_roles_bindings.yml
```diff
@@ -0,0 +1,32 @@
...
+metadata:
+  name: kube-system-service-catalog
+objects:
+
+# this isn't being created in the correct namespace....
+- apiVersion: v1
+  kind: Role
+  metadata:
+    name: extension-apiserver-authentication-reader
+    namespace: kube-system
```
bparees: you need to parameterize this field, even if you never intend to use a different value.  that's why my template did so:
https://github.com/openshift/origin/pull/14630/files#diff-2469e304b4201010f7a68ff63eac6b5bR132

it's due to legacy behavior in the template processor:  if you hardcode a namespace value, we strip it and ignore it.  if you use a parameter, we will respect it.

bparees: (that's why it's not being created in the correct namespace)

roles/openshift_service_catalog/files/kubesystem_roles_bindings.yml
```diff
@@ -0,0 +1,32 @@
...
+    resources:
+    - configmaps
+    verbs:
+    - get
+
+- kind: RoleBinding
+  apiVersion: v1
+  metadata:
+    name: extension-apiserver-authentication-reader-binding
+    namespace: kube-system
```
bparees: same thing for this one. (use a parameter reference)

bparees:
Note that the current iteration of the template in my cluster up PR does NOT enable user access to the TSB.

admins must manually allow unauthenticated access to the TSB to enable it.  not sure how much you care about that for this work.

bparees:
also i didn't dig through this PR to confirm, but if you are not somehow ensuring that users are given permission to CRUD instances+bindings in their own projects, the SC will not work.

in the cluster up PR, we do this by adding those permissions to the admin+edit roles, which every user has on their own projects, when the SC is enabled.  But if you try to do that in the installer, I suspect @deads2k will hunt you down.

roles/openshift_service_catalog/files/kubesystem_roles_bindings.yml
ewolinetz: sorry, that comment about it not being created in the right namespace was leftover before i moved it to its current location... the oc_process module should specify the module this is intended to be created in. If we don't need to specify a namespace in the metadata then I can remove this all together most likely (will double check)

ewolinetz:
Are we going to be installing the TSB as part of this initially? I thought we were going to use the Ansible Service Broker.

bparees:
> Are we going to be installing the TSB as part of this initially? I thought we were going to use the Ansible Service Broker.

those are not mutually exclusive options.  but i believe the plan was to NOT enable the TSB out of the box.  Not sure who's call that is, though.

roles/openshift_service_catalog/templates/api_server.j2
```diff
@@ -0,0 +1,86 @@
...
+        - --storage-type
+        - etcd
+        - --secure-port
+        - "6443"
+        - --insecure-bind-address
+        - 0.0.0.0
+        - --insecure-port
+        - "8081"
+        - --etcd-servers
+        - http://localhost:2379
```
ewolinetz: this will need to point to the core's etcd instead. TODO: remove embedded etcd container too

ewolinetz:
aside from trying to figure out why my apiserver pod can't resolve the cluster hostname for connecting to the core etcd this seems to install correctly and if I use the internal IP instead of the hostname for etcd I can `oc get brokers` correctly

roles/openshift_service_catalog/templates/api_server.j2
```diff
@@ -0,0 +1,80 @@
...
+{% endfor %}
+      containers:
+      - args:
+        - --storage-type
+        - etcd
+        - --secure-port
+        - "6443"
+        - --etcd-servers
+# TODO: for some reason the dns isn't resolving on this....
+        - https://{{ openshift.common.hostname }}:{{ openshift.master.etcd_port }}
```
ewolinetz: is `openshift.common.ip` a valid fact to use instead? @sdodson

ewolinetz:
After merging @fabianvf changes into this branch will need to remove my old broker role (will not be used) and also need to make updates so we can generate a PV for the ansible_service_broker's etcd pvc

ewolinetz:
aos-ci-test

ewolinetz:
we can now generate a pv to back the `openshift_ansible_service_broker` etcd pvc with the following added to our inventory
```yaml
openshift_hosted_etcd_storage_kind=nfs
openshift_hosted_etcd_storage_access_modes=['ReadWriteOnce']
openshift_hosted_etcd_storage_host=nfs.example.com
openshift_hosted_etcd_storage_nfs_directory=/exports
openshift_hosted_etcd_storage_volume_name=etcd
openshift_hosted_etcd_storage_volume_size=1Gi
openshift_hosted_etcd_storage_labels={'storage': 'etcd'}
```

sdodson:
@fabianvf https://github.com/sdodson/openshift-ansible/commit/d7ce993097fb482b0255471ecaa888118ce723bc seemed to have worked for me

ewolinetz:
@sdodson @fabianvf I picked from that commit into this PR branch

ewolinetz:
aos-ci-test

sdodson:
[merge][severity:blocker]

PR #4196, opened 2017-05-15
Don't set-up origin repositories if they've already been configured

dmsimard:
CentOS ships the 'centos-release-openshift-origin' package which is
used to set up the repository and the key. If this package has been
installed prior to running openshift-ansible (i.e, to install
ansible or openshift-ansible from RPM), it leads to the repository
being set up twice in different files.

We're now matching the name of the files and since they are identical,
it will stay idempotent while not configuring the repositories twice.

### DIFF
--- a/roles/openshift_repos/tasks/main.yaml
+++ b/roles/openshift_repos/tasks/main.yaml
@@ -24,15 +24,19 @@
     - openshift_additional_repos | length == 0
     notify: refresh cache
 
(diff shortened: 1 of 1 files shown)


abutcher:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
[merge]

Issue #2338, opened 2016-08-22
containerized installation fails when using openstack cloud provider with an encrypted API


maci0:
In the setup here OSP has an encrypted API. The key is signed by a corporate CA, which is in the host system's trustdb. 
However the openshift3/ose and openshift3/node container images don't have access to that.
Therefore the installation fails during master startup, as it tries to contact OSP and gets an error that the cert is not trusted.
Looking at kubernetes' openstack.go implementation there is no config option to simply ignore that.
Would it be possible to simply bind mount the certstore into the containers? if so that should be done in the systemd unit files by default when doing containerized installations.

maci0:
i can also confirm addign `-v /etc/pki/ca-trust:/etc/pki/ca-trust:ro` to the startupline in `/etc/systemd/system/atomic-openshift-master-api` does fix this issue

sdodson:
Thanks, checking w/ containers team to see if this is something that should be done automatically in a similar way to how subscription certs are made available to rhel containers or not. If not we can just mount it as you suggest.

sdodson:
@jwhonce Is this something that should ultimately be solved by containers team?

jwhonce:
@sdodson ack, going to discuss with team.

liveaverage:
I also hit this exact problem -- internal OSP8/OSP9 stack, API cert signed by internal CA, and failed playbook execution (for 1.4/3.4) when attempting to access the cloud provider. Is a fix scheduled for a future release? I'm attempting a 1.5/3.5 deployment now to see if it's addressed.

The same workaround works for me (-v /etc/pki/ca-trust:/etc/pki/ca-trust). Is it best to simply add this modification to a openshift_master_upgrade_pre_hook play (provided it would execute on first deployment)?

donnydavis:
I can confirm the fix from @liveaverage works

cooktheryan:
@sdodson and @jwhonce do we have any update on this?

sdodson:
This was fixed in release-1.4 via ea29fe1b4753a8c7ab08ec8a1f49f8293eba0e19 release-1.5 via a9ece184eeea7de99a947f799bb902524250b6ab and release-3.6 via 8132949a5f55320e01e7f95765092eb1b2633fcc where we mount the entire /etc/pki into the container

jwhonce:
@cooktheryan It should come in via a bind mount. Like /usr/share/rhel does today.

PR #4198, opened 2017-05-15
Update origin service templates to reference ca-trust

liveaverage:
#2338 Update service templates to reference ca-trust as data volume for origin-master and origin-node containers


michaelgugino:
Closing due to inactivity.

PR #4199, opened 2017-05-16
Default image policy on new clusters to on

smarterclayton:
Will allow for default image resolution to be used in openshift/origin#13210. We should have had this on from 1.4 onwards.


abutcher:
@smarterclayton Are we okay with user provided `openshift_master_admission_plugin_config` completely overriding this or would it be better to bake it in as a default which gets merged with whatever user gives us?

abutcher:
aos-ci-test

smarterclayton:
I think it has to be a default because existing config may have that variable set, and we can't change their behavior.

abutcher:
[merge]

Issue #4203, opened 2017-05-16
openshift_logging: don't create local tmpdir with `become`


stevekuznetsov:
Local tmp dir is created without `local_action` and `become: no`. This causes permissions errors when the control host is in the cluster and the tmpdir is created as root but the user running the playbook isn't root.

irozzo-1A:
@stevekuznetsov Same issue has not been fixed for [release-1.5](https://github.com/openshift/openshift-ansible/blob/release-1.5/roles/openshift_logging/tasks/generate_jks.yaml#L23). In addition there are other `local_action`s which should have `become: no` as well (i.e.. [roles/openshift_logging/tasks/generate_jks.yaml](https://github.com/openshift/openshift-ansible/blob/release-1.5/roles/openshift_logging/tasks/generate_jks.yaml#L29-L96)). I detected it because I launched the playbook from a container where `sudo` is not available. Are you ok if I submit a PR to fix it?

stevekuznetsov:
@irozzo-1A absolutely -- please submit a PR!

sdodson:
The cherry-pick failed, @irozzo-1A can you open a backport PR?

stevekuznetsov:
@kargakis why no comment here?

kargakis:
The bot is not handling issues.

stevekuznetsov:
D'oh... this is an issue

PR #4205, opened 2017-05-16
Don't escalate privileges in local tmpdir creation

stevekuznetsov:
When running `ansible-playbook` as a non-root user, if a local temporary
directory is created without explicitly not escalating privileges, the
directory will not be readable to the user running the playbook on the
controller host, which will lead to tasks like `fetch` failing to place
files into the directory.

/cc @jcantrill @ewolinetz @abutcher

### DIFF
--- a/roles/openshift_logging/tasks/main.yaml
+++ b/roles/openshift_logging/tasks/main.yaml
@@ -28,6 +28,7 @@
   register: local_tmp
   changed_when: False
   check_mode: no
+  become: no
 
 - debug: msg="Created local temp dir {{local_tmp.stdout}}"
 


stevekuznetsov:
aos-ci-test

stevekuznetsov:
[test]

stevekuznetsov:
fixes https://github.com/openshift/openshift-ansible/issues/4203

roles/openshift_logging/tasks/main.yaml
```diff
@@ -28,6 +28,8 @@
   register: local_tmp
   changed_when: False
   check_mode: no
+  delegate_to: localhost
+  become: no
```
sdodson: @ewolinetz @mtnbikenc Do we have a general strategy for use of local_action versus delegate_to? Feels like we should be using one or the other rather than mixing them.

stevekuznetsov: I copied this from a snippet @abutcher linked me somewhere else in the repo

abutcher: `delegate_to` isn't doing anything here (and everywhere else we combine `local_action` and `delegate_to`), the big thing is `become: no/False`. We should remove the `delegate_to` and I'll remove them from the other `local_action` tasks we have.

stevekuznetsov:
@abutcher @sdodson updated to remove the `delegate_to` re[test]

stevekuznetsov:
@sdodson can we do an AOS CI test on this

abutcher:
aos-ci-test

ewolinetz:
@stevekuznetsov +1 we probably will need to do this for the metrics role too

stevekuznetsov:
What's up with these statuses? AOS CI looks like it succeeded but never posted status?

sdodson:
He only thing I can think of is API throttling? Last time I saw this I
think the parent job had errors about that.

stevekuznetsov:
Can we re-trigger please?

abutcher:
aos-ci-test

abutcher:
[merge]

abutcher:
aos-ci-test

stevekuznetsov:
:(

screenshot from 2016-06-09 08-11-52

stevekuznetsov:
Failed to report success _again_ ????

sdodson:
aos-ci-test

stevekuznetsov:
It's so much worse when they're not your flakes...

```
GH_STATE=success
GH_DESC="e2e tests passed"
GH_CONTEXT=aos-ci-jenkins/OS_3.6_containerized_e2e_tests
```

FFS I'm about to hit the PR status API to update that damn thing myself

stevekuznetsov:
@arilivigni can we do anything about the AOS CI being more resilient when trying to update that status? Any way to trigger a re-sync? The job succeeds at running tests but fails to update status seemingly with high probability

PR #4216, opened 2017-05-17
fix etcd_container_version detection

ingvagabund:
The register variable `etcd_container_version` is modified even if the `"Record containerized etcd version"` task is skipped.

Plus, removing the `"Record containerized etcd version"` which will fail deployments without docker.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/etcd/upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/etcd/upgrade.yml
@@ -2,43 +2,61 @@
 - name: Determine etcd version
   hosts: oo_etcd_hosts_to_upgrade
   tasks:
-  - name: Record RPM based etcd version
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/upgrades/etcd/upgrade.yml
```diff
@@ -2,43 +2,61 @@
...
   tasks:
-  - name: Record RPM based etcd version
-    command: rpm -qa --qf '%{version}' etcd\*
-    args:
-      warn: no
-    register: etcd_rpm_version
-    failed_when: false
+  - block:
+    - name: Record RPM based etcd version
+      command: rpm -qa --qf '%{version}' etcd\*
```
ingvagabund: We should use `repoquery` instead. But nothing that needs to be done right now.

ingvagabund:
aos-ci-test

ingvagabund:
[merge]

PR #4221, opened 2017-05-17
Fix Centos PaaS SIG GPG path

sdodson:
#4196 changed where we drop the gpg key but the repo wasn't similarly updated


sdodson:
@smarterclayton can you test this?

sdodson:
Reproduced the problem, patch fixes it. None of the CI processes test this, clearly, so merging.

PR #4224, opened 2017-05-18
Fix creation of tmp Directory on localhost

VincentS:
When I run an openshift-ansible deployment with `ansible-playbook -b ....` Ansible tries to create an  tmp directory on the localhost with `root` which isnt necessary / fails if the user starting the Ansible Run doesn't have `root` privileges on the localhost he is using.

This patches solves this issue by not becoming `root` for creating a `tmp` directory on the localhost.


abutcher:
aos-ci-test

sdodson:
[merge]

PR #4225, opened 2017-05-18
relax check for rpm version and docker image version equality

ingvagabund:
If `openshift_pkg_version` or `openshift_image_tag` are defined, user gives basically a permission both rpm version and docker image version can be different.


ingvagabund:
aos-ci-test

ingvagabund:
@sdodson @ashcrow \O

PR #4229, opened 2017-05-18
Fix for Metrics & Logging pv-pvc crisscross problem

mazzystr:


roles/openshift_persistent_volumes/templates/persistent-volume.yml.j2
```diff
@@ -7,6 +7,8 @@ items:
   kind: PersistentVolume
   metadata:
     name: "{{ volume.name }}"
+    labels:
+      name: "{{ volume.name }}"
```
ewolinetz: can we make this more generic instead? i think we should be able to provide any key and value for a label

roles/openshift_metrics/templates/pvc.j2
```diff
@@ -25,3 +25,8 @@ spec:
   resources:
     requests:
       storage: {{size}}
+{% if openshift_metrics_cassandra_pvc_prefix is defined and openshift_metrics_cassandra_pvc_prefix %}
+  selector:
+    matchLabels:
+      name: {{ openshift_metrics_cassandra_pvc_prefix }}
```
ewolinetz: can we make this more generic, similar to logging?

roles/openshift_logging/tasks/generate_pvcs.yaml
```diff
@@ -16,7 +16,7 @@
     obj_name: "{{claim_name}}"
     size: "{{es_pvc_size}}"
     access_modes: "{{ es_access_modes | list }}"
-    pv_selector: "{{es_pv_selector}}"
+    pv_selector: "{{es_pv_selector | default(['name': openshift_hosted_logging_storage_volume_name])}}"
```
ewolinetz: if we can correctly provide a label to use when generating the PV we shouldnt need to use this as a default...

filter_plugins/oo_filters.py
```diff
@@ -749,7 +749,7 @@ def oo_persistent_volumes(hostvars, groups, persistent_volumes=None):
                         size = params['volume']['size']
                         access_modes = params['access']['modes']
                         persistent_volume = dict(
-                            name="{0}-volume".format(volume),
+                            name="{0}".format(volume),
```
ewolinetz: can we make this a separate PR? I think we should evaluate dropping this and the potential impact it will have with existing systems/customers

mazzystr:
Sorry guys.  I'm taken off the RHMAP / OpenShift project

ewolinetz:
@mazzystr mind if we close this in favour of https://github.com/openshift/openshift-ansible/pull/4302?

Issue #4233, opened 2017-05-19
oreg_url behaviour


tobru:
We have a use case where we only want to configure `imageConfig.format` in `master-config.yaml`. Setting `oreg_url` configures this unconditionally in the node and the master configuration. It would be great to have a way to influence this behavior, f.e. by having a parameter `oreg_url_master` and/or `oreg_url_node`.

PR #4234, opened 2017-05-19
allow to configure oreg_url specifically for node or master. refs #4233

tobru:
This commit allows to specify imageConfig.format specifically for master
or for nodes.

One use case of this could be if you want to use customer builder
images. In this case imageConfig.format only needs to be changed in the
master-config.yml but not in the node-config.yml.

### DIFF
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -127,6 +127,10 @@ openshift_release=v3.6
 
 
 # Alternate image format string, useful if you've got your own registry mirror
+# Configure this setting just on node or master
+#oreg_url_master=example.com/openshift3/ose-${component}:${version}
+#oreg_url_node=example.com/openshift3/ose-${component}:${version}
+# For setting the configuration globally
 #oreg_url=example.com/openshift3/ose-${component}:${version}
 # If oreg_url points to a registry other than registry.access.redhat.com we can
 # modify image streams to point at that registry by setting the following to true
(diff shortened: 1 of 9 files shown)


sdodson:
bot, test pull request

sdodson:
Fixes #4233

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -74,7 +74,7 @@
       ldap_ca: "{{ openshift_master_ldap_ca | default(lookup('file', openshift_master_ldap_ca_file) if openshift_master_ldap_ca_file is defined else None) }}"
       openid_ca: "{{ openshift_master_openid_ca | default(lookup('file', openshift_master_openid_ca_file) if openshift_master_openid_ca_file is defined else None) }}"
       request_header_ca: "{{ openshift_master_request_header_ca | default(lookup('file', openshift_master_request_header_ca_file) if openshift_master_request_header_ca_file is defined else None) }}"
-      registry_url: "{{ oreg_url | default(None) }}"
+      registry_url: "{{ oreg_url | default(oreg_url_master) | default(None) }}"
```
sdodson: This would mean that if oreg_url and oreg_url_master are both set then you'll get oreg_url value. To me I'd expect this to work in the opposite manner, the more specific value should be used. What do you think?

inventory/byo/hosts.origin.example
```diff
@@ -128,6 +128,9 @@ openshift_release=v3.6
 
 # Alternate image format string, useful if you've got your own registry mirror
 #oreg_url=example.com/openshift3/ose-${component}:${version}
+# Specifically set oreg_url on the master or on nodes
+#oreg_url_master=example.com/openshift3/ose-${component}:${version}
```
sdodson: Based on the decision we make in my other comment lets make sure to mention the variable precedence here so it's clear.

roles/openshift_master_facts/tasks/main.yml
tobru: You're completely right. Makes more sense that the more specific setting overrides the other. I'll update the PR to reflect this change.

sdodson:
aos-ci-test

sdodson:
[merge]

tobru:
Any chance to get this merged soon?

sdodson:
[merge]

sdodson:
thanks for the reminder!

tobru:
Why does this jenkins merge job always fail?

sdodson:
flake on https://github.com/openshift/origin/issues/13105
[merge][severity: bug]

PR #4239, opened 2017-05-19
Add a readiness probe to the Kibana container

stevekuznetsov:
In order to ensure that the Kubernetes machinery can determine when the
Kibana Pods are becoming ready, we need to add a readiness probe to the
Containers that make up those pods. The Kibana readiness probe simply
hits the base URL at `http://localhost:5601/` and expects a 200.

/cc @richm @ewolinetz  @lukas-vlcek 

depends on https://github.com/openshift/origin-aggregated-logging/pull/423/files

### DIFF
--- a/roles/openshift_logging/templates/kibana.j2
+++ b/roles/openshift_logging/templates/kibana.j2
@@ -61,6 +61,13 @@ spec:
             - name: kibana
               mountPath: /etc/kibana/keys
               readOnly: true
+          readinessProbe:
+            exec:
+              command:
+              - "/usr/share/kibana/probe/readiness.sh"
+            initialDelaySeconds: 5
+            timeoutSeconds: 4
+            periodSeconds: 5
         -
           name: "kibana-proxy"
           image: {{proxy_image}}


ewolinetz:
@stevekuznetsov what is the priority of getting this merged in? Is it possible to wait until after https://github.com/openshift/openshift-ansible/pull/3509 merges?

stevekuznetsov:
@ewolinetz the major reason for this is it separates concerns about testing for us -- right now there are about five different ways we test for Kibana readiness in the test suite so it is not super urgent -- the more we want to rely on Kubernetes-based tools like `kubectl rollout status` the more we'll need this. It can wait.

roles/openshift_logging/templates/kibana.j2
```diff
@@ -61,6 +61,13 @@ spec:
             - name: kibana
               mountPath: /etc/kibana/keys
               readOnly: true
+          readinessProbe:
+            exec:
+              command:
+              - "/usr/share/kibana/probe/readiness.sh"
+            initialDelaySeconds: 5
```
jcantrill: Do we think there will be any need to tweak these at deployment?

stevekuznetsov:
Changes folded into https://github.com/openshift/openshift-ansible/pull/3509

PR #4252, opened 2017-05-19
Tolerate failures in the node upgrade playbook

sdodson:
In large environments it's likely that some percentage of nodes may fail the upgrade procedures. As such we should allow the admin to configure the max_fail_percentage.

### DIFF
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -785,6 +785,31 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 #
 #etcd_ca_default_days=1825
 
+# Upgrade Control
+#
+# By default nodes are upgraded in a serial manner one at a time and all failures
+# are fatal
+#openshift_upgrade_nodes_serial=1
+#openshift_upgrade_nodes_max_fail_percentage=0
+#
+# You can specify the number of nodes to upgrade at once. We do not currently
+# attempt to verify that you have capacity to drain this many nodes at once
+# so please be careful when specifying these values. You should also verify that
+# the expected number of nodes are all schedulable and ready before starting an
+# upgrade. If it's not possible to drain the requested nodes the upgrade will
+# stall indefinitely until the drain is successful.
+#
+# If you're upgrading more than one node at a time you can specify the maximum
+# percentage of failure within the batch before the upgrade is aborted. Any
+# nodes that do fail are ignored for the rest of the playbook run and you should
+# take care to investigate the failure and return the node to service so that
+# your cluster.
+#
+# The percentage must exceed the value, this would fail on two failures
+# openshift_upgrade_nodes_serial=4 openshift_upgrade_nodes_max_fail_percentage=49
+# where as this would not
+# openshift_upgrade_nodes_serial=4 openshift_upgrade_nodes_max_fail_percentage=50
+
 # host group for masters
 [masters]
 ose3-master[1:3]-ansible.test.example.com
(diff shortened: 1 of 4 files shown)


sdodson:
@jupierce FYI

inventory/byo/hosts.ose.example
```diff
@@ -785,6 +785,25 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
...
+# more than one node at a time you can specify the maximum percentage of failure
+# within the batch before the upgrade is aborted.
+#
+# Be careful if you set openshift_upgrade_nodes_serial to a percentage and your
+# node counts will change. Also be mindful if you have a failure you should
+# return the failed node to service so that you don't exhaust your cluster
+# capacity.
+#
+# The following configuration would abort if two nodes in a batch of four failed.
+#openshift_upgrade_nodes_serial=4
```
mtnbikenc: Outside of these examples, should we indicate the defaults of these variables? A standard practice is to have commented variables defined to their defaults.

```
#openshift_upgrade_nodes_serial=1
#openshift_upgrade_nodes_max_fail_percentage=0
```

sdodson: Updated the documentation a bit. Hopefully it's clearer now?

inventory/byo/hosts.ose.example
```diff
@@ -785,6 +785,31 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
...
+# so please be careful when specifying these values. You should also verify that
+# the expected number of nodes are all schedulable and ready before starting an
+# upgrade. If it's not possible to drain the requested nodes the upgrade will
+# stall indefinitely until the drain is successful.
+#
+# If you're upgrading more than one node at a time you can specify the maximum
+# percentage of failure within the batch before the upgrade is aborted. Any
+# nodes that do fail are ignored for the rest of the playbook run and you should
+# take care to investigate the failure and return the node to service so that
+# your cluster.
```
sdodson: We should probably make the drain operation async and set a time limit for how long we're willing to wait to drain a node. I'll see if I can get some data from the ops team as to what the longest successful drain operation is in their environment.

sdodson:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -248,7 +248,7 @@
   # This var must be set with -e on invocation, as it is not a per-host inventory var
   # and is evaluated early. Values such as "20%" can also be used.
   serial: "{{ openshift_upgrade_nodes_serial | default(1) }}"
-  any_errors_fatal: true
+  max_fail_percentage: "{{ openshift_upgrade_nodes_max_fail_percentage | default(0) }}"
```
mwoodson: @sdodson I noticed that you used the same variable for both controls_plane and nodes.  I think it could be useful if these were separate values.

I would like to be able to fail on any control plane error.  IT's much more important than the nodes.  I understand the argument about numbers.  Like, if i have 3 (or even 10) masters, one fail would fail it.  I still like the idea of separate values.

Thoughts?

sdodson: Yeah, I think we need separate values for both the serial and max_fail_percentage.

sdodson: @mwoodson updated, hows it look now.

sdodson:
@mtnbikenc If any host in the batch gets an UNREACHABLE error then all subsequent batches are skipped and no additional plays are executed. Do you know of any way to prevent one unreachable host from skipping remaining batches?

sdodson:
It looks like the run that failed had two batches with unreachable hosts and then the final batch had a host that failed the task.

sdodson:
@mtnbikenc @abutcher or @mwoodson can i get another review on this in hopes of merging this sprint?

mwoodson:
@sdodson this seems legit. LGTM

sdodson:
aos-ci-test

sdodson:
flake on https://github.com/openshift/origin/issues/14041
[merge]

PR #4253, opened 2017-05-19
[Docs] Ansible Role Variable Rule

mtnbikenc:
This PR is a proposal intended to provide agreement/validation of architecture discussions/decisions.  I would like the team to collectively provide feedback or approval as the see fit to ensure we are all in agreement.  If there are aspects missing or misrepresented, please comment.  Please add any reviewers I have missed.

### DIFF
--- a/docs/style_guide.adoc
+++ b/docs/style_guide.adoc
@@ -1,5 +1,3 @@
-// vim: ft=asciidoc
-
 = openshift-ansible Style Guide
 
 The purpose of this guide is to describe the preferred coding conventions used in this repository (both in ansible and python).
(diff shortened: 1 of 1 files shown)


ashcrow:
First pass I noticed that it seems like ``cli`` items are covered under ``g_`` and ``cli_``. While not related to the changes you made I think we should clarify if it should be ``cli_`` for ``cli`` items or ``g_cli_`` which is what I **think** it's pointing at right now.

docs/style_guide.adoc
```diff
@@ -96,79 +94,24 @@ g_environment: someval
...
-
-.3 word example:
-* Role name: made_up_role
-* Prefix: mur
-[source]
-----
-mur_var1: value_one
-----
+.Purpose
+Name variables to avoid name collision between roles or playbooks.
```
ingvagabund: **Other requirements**:

If a role `roleB` depends on `roleA`, all role variables from `roleA` that affects value of role variables in `roleB` must have their equivalents defined in  `roleB`.
E.g.
* `roleA` defines among others `r_roleA_runtime` and `r_roleA_conf_dir` variables
* `roleB` defines `r_roleB_service_file` which depends on `r_roleA_runtime` variable
* then, `roleB` must defined `r_roleB_runtime` variable

tbielawa: With the intended side effect of this being that we have consistent variable naming between roles, right?

mtnbikenc:
@ashcrow I read through the other sections and intentionally didn't make any other suggestions outside of the role variables.  The part about `cli_` does not make a lot of sense to me, given that Ansible variables can be defined in a multitude of places.  Why would be designate a particular variable as something passed at the cli, when it could just as easily end up in group_vars/host_vars/inventory/etc.?  I wanted to leave this discussion to another PR unless we ended up in a situation where we could not discuss one without the other.

ashcrow:
@mtnbikenc understood.

docs/style_guide.adoc
ashcrow: @tbielawa that's my understanding.

ashcrow: I like @ingvagabund's example. I feel like having that (or something similar) added could help clarify for those reading this document for the first time.

ewolinetz: Do we need to have a `r_roleB_runtime` in this case or could we just update how `r_roleB_service_file` is set?

To me, if we are needing to set `r_roleB_service_file` and it is dependent on `r_roleA_runtime` then when we are calling `roleB` we would set the var like 
```yaml
r_roleB_service_file: "/etc/{{ r_roleA_runtime }}/service.file"
``` 
if we are calling `roleB` from within `roleA` otherwise we would use the variable available to us that `r_roleA_runtime` was set using...

ingvagabund: I hit very similar situation when refactoring etcd roles:

etcd_common/default/main.yml
```yaml
r_etcd_common_service: "{{ 'etcd' if r_etcd_common_runtime == 'host' else 'etcd_service' }}"
```

etcd/tasks/main.yml
```yaml
- name: Restart etcd
  service:
    name: "{{ r_etcd_common_service }}"
    state: restarted
```

Then, when calling role `etcd` one needs to set `r_etcd_common_runtime`. One can either:
```yaml
- roles:
    - role: etcd
      r_etcd_common_runtime: "docker"
```
in which case we must be aware the `etcd` depends on `etcd_common` (but that can change over time). Or we can have `etcd` role define `r_etcd_runtime` varible and have `etcd` set `r_etcd_common_runtime`:
```
```yaml
- roles:
    - role: etcd
      r_etcd_runtime: "docker"
```

The latter has advantage of us not knowing about etcd dependencies. If `etcd_common` gets renamend to `etcd_dog`, API of `etcd` will remain untouched.

ingvagabund: If we go with @kwoodson's suggestion:
```yaml
- name: A play
  tasks:
  - include_role:
      name: etcd_common
    vars:
      r_etcd_common_etcd_runtime: "{{ openshift.common.etcd_runtime }}"
  - include_role:
      name: etcd_upgrade
    vars:
      r_etcd_upgrade_action: upgrade
      r_etcd_upgrade_mechanism: rpm
      r_etcd_upgrade_version: "3.1.9"
      r_etcd_upgrade_etcd_runtime: "{{ openshift.common.etcd_runtime }}"
      r_etcd_upgrade_service: "{{ r_etcd_common_service }}"
      r_etcd_upgrade_service_file: "{{ r_etcd_common_service_file }}"
      r_etcd_upgrade_etcdctlv2: "{{ r_etcd_common_etcdctlv2 }}"
```

with moving some variables from `etcd_common/defaults/main.yml` and setting them in `etcd_common/tasks/main.yml` via `set_fact` so all "exported" etcd_common role variables are evaluated, my previous comments can be ignored.

mtnbikenc:
Closing this as it is no longer relevant given recent architecture discussions.  Updating the style_guide will be handled by a future PR.

PR #4254, opened 2017-05-19
Refactor openshift_hosted's docker-registry route setup

dmsimard:
We have identified an issue where a docker-registry service set up
as 'reencrypt' with a provided certificate and a self-signed certificate
on the pod does not authorize users to push images.
If the docker-registry service is set up as 'passthrough' with the
same provided certificate, everything works. This is filed as an
Origin bug here: https://github.com/openshift/origin/issues/14249

In light of this, this commit essentially adds support for configuring
provided certificates with a passthrough route while maintaining backwards
compatibility with the other use cases.
The default remains 'passthrough' with self-generated certificates.

Other miscellaneous changes include:
- Move fact setup that were only used in secure.yml there
- Omit the hostname for the route if there are none to configure,
  oc_route takes care of handling the default
- Replace hardcoded /etc/origin/master by openshift_master_config_dir

PR #4256, opened 2017-05-21
Refactor registry-console setup and add support for SSL

dmsimard:
This is a refactor to significantly improve the configurability
of the registry console service and route in order to bring it
in line with the way we configure docker-registry.

Effectively, this adds the ability to select a hostname for the
registry console and adds support for securing the registry-console
route with a provided SSL certificate either with passthrough or
reencrypt termination.

We maintain backwards compatibility by keeping the same default
which provides a default registry-console hostname and self-signed
certificates on a passthrough route.

It uses the same approach as with the docker registry refactor in this
other PR: https://github.com/openshift/openshift-ansible/pull/4254


dmsimard:
@abutcher, @aweiteka: FYI for the standalone registry use case.

dmsimard:
Note the usage of "openshift_hosted" prefix for variables instead of "cockpit_ui". I was on the verge of "merging" the cockpit and cockpit-ui roles into the openshift_hosted role but I decided against it because I am not knowledgeable enough of the role composition to understand what are the implications.

Would like opinions on that.

sdodson:
I'd previously almost renamed things `openshift_hosted_cockpit` or something along those lines. We'll probably split openshift_hosted up into openshift_hosted_registry and openshift_hosted_router. Basically openshift_hosted_ would be the prefix for anything that we install atop a running openshift environment.

abutcher:
We prefix variable names with the name of the role they are used in although there are older roles where this rule isn't followed. I think we would ultimately want to rename the `cockpit_ui` role to be something like `openshift_hosted_registry_console` if our plan is to also split `openshift_hosted_registry` and `openshift_hosted_router` into their own roles.

dmsimard:
@abutcher @sdodson so at least it looks like we agree that cockpit/cockpit-ui could/should be merged together and renamed to something more appropriate ?

I don't know enough about cockpit/cockpit-ui (are they used for anything else than the registry console? [It looks like it...](https://blog.openshift.com/monitoring-openshift-cluster-using-cockpit/)) That's why I didn't move it.

I guess it would also be a backwards-incompatible change (changing role name, changing var names) and I don't know how okay this is.
And lastly, I get lost very quickly in the playbooks and apparently endless symlinks -- what roles are included where or untangle the implicit role (meta) dependencies. I'd probably need help figuring out where things needs renamed if we go that route.

roles/cockpit-ui/tasks/route/passthrough.yml
```diff
@@ -0,0 +1,46 @@
...
+    set_fact:
+      docker_registry_console_cert_path: "{{ openshift_master_config_dir }}/named_certificates/{{ openshift_hosted_registry_console_routecertificates['certfile'] | basename }}"
+      docker_registry_console_key_path: "{{ openshift_master_config_dir }}/named_certificates/{{ openshift_hosted_registry_console_routecertificates['keyfile'] | basename }}"
+      docker_registry_console_self_signed: false
+
+  # Since we end up bundling the cert, cacert and key in a .pem file, the 'cafile'
+  # is optional
+  - name: Configure provided ca certificate file path
+    set_fact:
+      docker_registry_console_cacert_path: "{{ openshift_master_config_dir }}/named_certificates/{{ openshift_hosted_registry_console_routecertificates['cafile'] | basename }}"
```
abutcher: Same question re: keeping these in the `named_certificates` directory. For more context, we currently only keep SNI certificates specified via `openshift_master_named_certificates` in this directory and there are other tasks which will clean this dir out based on setting `openshift_master_overwrite_named_certificates=true` so I'd be worried about losing them if we want to keep them but not sure that we want to keep them.

abutcher:
@dmsimard I'll dig into the renaming and see if I can list everything we'd need to account for.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -45,17 +43,68 @@
...
+      {%-   set _ = files.append(docker_registry_console_cacert) %}
+      {%- endif %}
+      {%- set _ = files.append(docker_registry_console_key_path) %}
+      {{- files -}}
+    register: certificate_files
+
+  - name: Generate certificate bundle
+    copy:
+      content: "{{ certificate_files.results | map(attribute='content') | map('b64decode') | join('') }}"
+      dest: "{{ openshift_master_config_dir }}/named_certificates/registry-console.pem"
```
dmsimard: Not a bad idea to put it somewhere temporary. That said, I guess we don't need to keep *any* certificate around once they have been configured either directly on the route (passthrough) or as a secret (mounted as a volume).

That said, I guess you sort of lose some idempotency since you'll need to upload certs in tmpdir/generate bundle/check against what is configured/delete tmpdir.
The fact that the bundle is stored means the task should not be "changed" because the source files are already there and the bundle ends up the same.

roles/cockpit-ui/tasks/main.yml
```diff
@@ -45,17 +43,68 @@
...
+
+  - name: Create the secret for the registry-console certificate
+    oc_secret:
+      name: registry-console-certificate
+      namespace: "{{ openshift_hosted_registry_console_namespace }}"
+      files:
+      - name: registry-console.cert
+        path: "{{ openshift_master_config_dir }}/named_certificates/registry-console.pem"
```
dmsimard: There shouldn't be a new certificate every run, no, given that oc_adm_ca_server_cert and oc_secret are idempotent (from my understanding).

roles/cockpit-ui/tasks/route/passthrough.yml
dmsimard: Unless mistaken, at least in the case of the registry generated certificate (prior to my pull request), the certificate was stored in /etc/origin/master, not in /etc/origin/master/named_certificates. I moved it to named_certificates but we can keep it somewhere else (if we want to keep it)

abutcher:
Looks like we only call into the current `cockpit-ui` role in one place so the rename change was pretty small. https://github.com/abutcher/openshift-ansible/commit/68fe744a4b40710bdf9932e249caf75380aa7ca0

roles/cockpit-ui/tasks/main.yml
abutcher: Makes sense to keep the files for idempotency. `/etc/origin/master/named_certificates` isn't a good folder due to the overwrite var so sticking with convention and using `/etc/origin/master` would be fine for files like `/etc/origin/master/registry-console.*`.

I'd be worried about cluttering `/etc/origin/master` storing files that could have names defined by the user - also potential file name conflicts like ca.crt overwriting the OpenShift CA. What if we add a new directory for `registry-console` and store everything related there? @sdodson Thoughts on that?

dmsimard: Ack. Note that I've taken the same approach in the docker-registry refactor here https://github.com/openshift/openshift-ansible/pull/4254 so whatever solution we agree on would likely apply to both.

dmsimard:
@abutcher ok, I'll include https://github.com/abutcher/openshift-ansible/commit/68fe744a4b40710bdf9932e249caf75380aa7ca0 in my next rebase/commit, thanks.

roles/cockpit-ui/tasks/main.yml
sdodson: another directory is fine with me

dmsimard:
Nothing in particular yet, just rebasing this on top of current master, it was getting stale. I'll try to take a new look at this in the near future.

jtudelag:
Hi @dmsimard, great job.  It would be really helpful to document this in the openshift-docs repo:
https://github.com/openshift/openshift-docs 

Also this https://github.com/openshift/openshift-ansible/pull/4254

Im gonna open two issue there to keep track of both, just FYI.

dmsimard:
Hi @jtudelag, thanks for reminding me I still needed to land this... I just rebased on top of current master and addressed @sdodson and @abutcher's comments about moving certificates from /etc/origin/master/named_certificates to another directory, /etc/origin/master/registry_console_certificates.

I have very little available bandwidth and asking me to write docs will likely leave you waiting for a long time -- if anyone else wants to pick that up, that's fine with me.

dmsimard:
This now includes the rename from 'cockpit-ui' to 'openshift_hosted_registry_console' as per discussed in this PR.

jtudelag:
@dmsimardno no worries, I just created the issue in openshift-docs in order not to forget the documentation ;).

michaelgugino:
@dmsimard Are you planning on rebasing this at some point?  These roles have changed dramatically since last update.

dmsimard:
@abutcher @michaelgugino I got the pull request rebase fatigue.. I sent this more than a year ago now and we're still using it for our deployments. I can rebase once more if we're still interested in this.

michaelgugino:
@dmsimard Thanks for the update.  I don't think we'll be merging this any time soon.  We're in the middle of a bug-fix sprint for the 3.10 release, so this wouldn't land until earliest 3.11.

Conan-Kudo:
@michaelgugino I hope this could be picked up and pulled into OpenShift-Ansible soon. I'm pulling in a variant of this change for my fork of this to deploy Origin 3.9 with this for our use, and it'd be handy to have it in mainline.

Conan-Kudo:
@dmsimard Could you please rebase this for the current master?

Conan-Kudo:
@dmsimard Now that Origin 3.10 is out, can you please rebase this for current git master (which will become 3.11)?

vrutkovs:

PR #4257, opened 2017-05-21
Allow a hostname to resolve to 127.0.0.1 during validation

dmsimard:
127.0.0.1 does not end up in "ansible_all_ipv4_addresses" and should
be considered valid.
For example, a hostname could be resolving to a public address that is
not bound to the actual server.
Or the host could resolve to 127.0.0.1 from the perspective of
the host since the hostname is set up in /etc/hosts.


sdodson:
aos-ci-test

sdodson:
[merge]

abutcher:

abutcher:
[merge]

PR #4261, opened 2017-05-22
Add Kibana readiness probe, stop using `oc scale` for cluster start

stevekuznetsov:
Create logging deployments with non-zero replica counts

When we currently create the set of logging `DeploymentConfig`s, we
create them with zero desired replicas. This causes the deployment to
immediately succeed as there is no work to be done. This inhibits our
ability to use nice CLI UX features like `oc rollout status` to monitor
the logging stack deployments. Instead, we should can create the configs
with the correct number of replicas in the first place and stop using
`oc scale` to bring them up after the fact.

---

Add a readiness probe to the Kibana container

In order to ensure that the Kubernetes machinery can determine when the
Kibana Pods are becoming ready, we need to add a readiness probe to the
Containers that make up those pods. The Kibana readiness probe simply
hits the base URL at `http://localhost:5601/` and expects a 200.

---


stevekuznetsov:
Wrong target branch!

sdodson:
You can edit target branches on prs now. It's unintuitively hidden under the edit button to edit the title of a pr.

stevekuznetsov:
Cool info, but I also needed to target a different fork :(

PR #4269, opened 2017-05-24
openshift_logging: use deployments instead of DCs

wozniakjan:
Change the deploymentconfigs to kubernetes deployments.

NOTE: after the #3509 got merged to master, additions in oc_scale.py proposed here are no longer used by any playbook, should they be removed from this PR?

### DIFF
--- a/roles/lib_openshift/library/oc_scale.py
+++ b/roles/lib_openshift/library/oc_scale.py
@@ -1757,6 +1757,352 @@ spec:
 
 # -*- -*- -*- End included fragment: lib/deploymentconfig.py -*- -*- -*-
 
+# -*- -*- -*- Begin included fragment: lib/deployment.py -*- -*- -*-
+
+
+# pylint: disable=too-many-public-methods
+class Deployment(Yedit):
+    ''' Class to model an openshift Deployment'''
+    default_deployment_config = '''
+apiVersion: extensions/v1beta1 
+kind: Deployment
+metadata:
+  name: default_deployment
+  namespace: default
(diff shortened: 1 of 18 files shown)


wozniakjan:
@jcantrill, @richm could you please take a look at this PR? 

I am afraid I can't add a reviewer properly according to [this page](https://help.github.com/articles/requesting-a-pull-request-review/)

stevekuznetsov:
Is there some background on why we'd prefer `Deployment`s over `DeploymentConfig`s ?

wozniakjan:
I can't provide sufficient answer but there is a trello card for it :)

https://trello.com/c/fJr858nz

ewolinetz:
What is the impact on users that are currently using DCs and then running the playbook with these changes? Will they end up with duplicates e.g. a Kibana DC and a Kibana Deployment?

jcantrill:
We prefer to use deployments mostly to rely on upstream constructs where it makes sense.  Currently, there is nothing about a DC that we utilize.  In regards to  how this will impact current users, we will have to account for an 'upgrade' path.  Also, I wonder  if we have concerns of how this might affect associated tooling.

Should we punt this change to 3.7 where we can add more thought around it?

stevekuznetsov:
I would wait until upstream Origin has a clear upgrade path from DC --> D

wozniakjan:
> Will they end up with duplicates e.g. a Kibana DC and a Kibana Deployment?

yes, I haven't thought that through very well, thanks!

PR #4271, opened 2017-05-24
disable become for local actions

madddi:
When running the playbooks an a bastion host as an unprivileged user, local actions with become will fail or create permission problems.

### DIFF
--- a/roles/openshift_logging/tasks/generate_certs.yaml
+++ b/roles/openshift_logging/tasks/generate_certs.yaml
@@ -21,6 +21,7 @@
     --key={{generated_certs_dir}}/ca.key --cert={{generated_certs_dir}}/ca.crt
     --serial={{generated_certs_dir}}/ca.serial.txt --name=logging-signer-test
   check_mode: no
+  become: false
   when:
     - not ca_key_file.stat.exists
     - not ca_cert_file.stat.exists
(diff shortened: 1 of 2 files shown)


sdodson:
bot, test pull request

sdodson:
aos-ci-test

abutcher:
Neither of these tasks are `local_action` so I'm not sure that `become: false` is required here. Both tasks would be ran against masters rather than localhost.

madddi:
@abutcher Yes, that's right. But if you're using become and create the temp directory on the amster as root, the fetch step in generate_jks.yaml fails:

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/tasks/generate_jks.yaml#L43

Generating the certs as root has the samt effect as the created files won't be readable by the unprivileged user.

abutcher:
@madddi Makes sense to me. I had not seen the fetch.

madddi:
Thanks. It looks like something went wrong with the Jenkins build. Could you run it again so that the changes get merged?

sdodson:
[merge][severity: bug]

stevekuznetsov:
This is the wrong fix. If you are running on the master, and as an unprivileged user, you do not have access to `/etc/origin/master`. Temporary directories _should_ be created as the normal user without `become` as I have done in https://github.com/openshift/openshift-ansible/commit/6985739c03297f5d0ead06d5193a82155db3a661.

PR #4278, opened 2017-05-24
Use local openshift.master.loopback_url when generating loopback kubeconfig.

abutcher:
Fixes BZ#1454321 https://bugzilla.redhat.com/show_bug.cgi?id=1454321

When we're generating certificates for an HA cluster, we first create a set of certificates for the first master using `oc adm create-master-certs` which generates every certificate/key the first master will need. We pass `--master` and `--public-master` arguments matching the `openshift.master.{public_api_url, api_url}` facts which means that the kubeconfigs (`admin.kubeconfig`, `openshift-master.kubeconfig`) will be configured to talk to the potentially load balanced cluster endpoints.

```
# oc adm create-master-certs -h
...
      --master='https://localhost:8443': The API server's URL.
      --public-master='': The API public facing server's URL (if applicable).
```

The `openshift-master.kubeconfig` is used by the controllers service to talk to the master API so we want it to talk to the local master (loopback) rather than the load balanced endpoints since the controllers are going to talk to the API a lot. So, once we've created the set of certificates for the first master, check to see if the `openshift-master.kubeconfig` contains the loopback server address (it won't right after they are created) and if it doesn't then we create a new `openshift-master.kubeconfig` which does.

Then when we're generating loopback kubeconfigs for additional masters we just pass the loopback API urls to begin with since these configs are generated for additional masters individually.

### DIFF
--- a/roles/openshift_ca/tasks/main.yml
+++ b/roles/openshift_ca/tasks/main.yml
@@ -108,6 +108,38 @@
   delegate_to: "{{ openshift_ca_host }}"
   run_once: true
 
+- name: Test local loopback context
+  command: >
+    {{ hostvars[openshift_ca_host].openshift.common.client_binary }} config view
+    --config={{ openshift_master_loopback_config }}
+  changed_when: false
+  register: loopback_config
+  delegate_to: "{{ openshift_ca_host }}"
+  run_once: true
+
+- name: Generate the loopback master client config
+  command: >
+    {{ hostvars[openshift_ca_host].openshift.common.client_binary }} adm create-api-client-config
+      {% for named_ca_certificate in openshift.master.named_certificates | default([]) | oo_collect('cafile') %}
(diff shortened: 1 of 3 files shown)


abutcher:
aos-ci-test

sdodson:
LGTM, sorry for delay in looking at this.

abutcher:
Hit an issue with this and cert redeploy due to a  missing a variable from calling `openshift_ca` directly vs. through the `openshift_master` role. New commit otw

abutcher:
aos-ci-test

abutcher:
aos-ci-test

sdodson:
LGTM still

abutcher:
[merge]

abutcher:
flake https://github.com/openshift/origin/issues/14129

abutcher:
[merge]

PR #4134, opened 2017-05-09
Adding assert to check for python-passlib on control host

ewolinetz:
to address #4111

PR #4277, opened 2017-05-24
Updating python-passlib assert

ewolinetz:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1455310
Fixes https://github.com/openshift/openshift-ansible/issues/4111
Fixes https://github.com/openshift/openshift-ansible/issues/4287

PR #4285, opened 2017-05-25
Fix passlib check for metrics installation

metal3d:
see #4134 that is not correct on Centos7, Fedora, and other distribution where pacakge is named python2-passlib.


metal3d:
Duplicated #4277 where I gave fixes

PR #4288, opened 2017-05-25
openshift_logging: increasing *_elasticsearch_* default CPU and memory

wozniakjan:
During my testing, I couldn't ever get a healthy ES pod. Eric has described to me how to fix the defaults to respect the old settings. This works well in my case.

NOTE: openshift_logging_elasticsearch_cpu_limit should also be done in similar fashion as other defaults, I couldn't figure out how to make it work when openshift_logging_es_cpu_limit is undefined.

### DIFF
--- a/roles/openshift_logging_elasticsearch/defaults/main.yml
+++ b/roles/openshift_logging_elasticsearch/defaults/main.yml
@@ -5,10 +5,10 @@ openshift_logging_elasticsearch_image_version: "{{ openshift_hosted_logging_depl
 openshift_logging_elasticsearch_image_pull_secret: "{{ openshift_hosted_logging_image_pull_secret | default('') }}"
 openshift_logging_elasticsearch_namespace: logging
 
-openshift_logging_elasticsearch_nodeselector: ""
-openshift_logging_elasticsearch_cpu_limit: 100m
-openshift_logging_elasticsearch_memory_limit: 512Mi
-openshift_logging_elasticsearch_recover_after_time: 5m
+openshift_logging_elasticsearch_nodeselector: "{{ openshift_logging_es_nodeselector | default('') }}"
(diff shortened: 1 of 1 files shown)


wozniakjan:
@richm was interested to know what was causing the slow ES pod on my machine, @ewolinetz please confirm I understood your explanation well

richm:
@wozniakjan This fixes the CI problem I saw.  @ewolinetz @jcantrill @stevekuznetsov We need to get this PR merged ASAP

stevekuznetsov:
This stops the hangs? I triggered a job for this here:

https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_aggregated_logging_openshift_ansible/17/console

stevekuznetsov:
Looks like the test job I ran did not hang -- why would we have expected this to fix that?

stevekuznetsov:
This does seem to fix the issue. I would like to see hard limits set on validating user input here -- we should make it impossible for a customer to crash their kernel by resource-starving Elasticsearch on accident with at typo in their inventory, etc.

richm:
> Looks like the test job I ran did not hang -- why would we have expected this to fix that?

This commit makes it so that the elasticsearch pod has reasonable cpu and memory, instead of the severely under resourced elasticsearch that is somehow able to cause a kernel panic.

stevekuznetsov:
Looks like kernels newer than a year should be fine -- we still see ES fail because of cpu restrictions, but we don't see the kernel lockup. Anything left on this or can we merge?

abutcher:
aos-ci-test

abutcher:
[merge]

sdodson:
[merge][severity:bug]

stevekuznetsov:
Failure was https://github.com/openshift/origin/issues/14140

sdodson:
flake https://github.com/openshift/origin/issues/14140

PR #4289, opened 2017-05-25
fix issue 4287 by using regexp in rpm search

pruan-rht:
@ewolinetz, @sdodson, PTAL.  Thanks!


pruan-rht:
closing it since it's a duplicate of https://github.com/openshift/openshift-ansible/pull/4277

PR #4291, opened 2017-05-25
WIP: introduce trivial tests for openshift_excluder role

ingvagabund:
This is my hacking effort with molecule. Former effort done at https://github.com/openshift/openshift-ansible/pull/2998.

Given the molecule itself does not support scenarios yet (e.g. running various combination of input ansible variables), I have introduce `gas` command, that takes `molecule.yml.j2` template and for each scenario in `scenarios.yml` file generates one `molecule.yml`. At the moment all scenarios are run in series. In future, it make sense to run the scenarios in parallel (need to figure out how to present tests logs in a transparent way).

The `gas` command itself is just a wrapper and is meant to be replace by molecule v2 once the scenarios are implemented. Meantime, the `gas` executable will inhabit `test` or `hack` directory.

Each role will have `molecule` directory with (possible) the following structure:
```sh
molecule/
├── ansible.cfg
├── Dockerfile
├── molecule.yml.j2
├── playbooks
├── playbook.yml
├── scenarios.yml
└── tests
```

* ansible.cfg (this should go away eventually)
* Dockerfile (basically just state image to use to override the default one due to https://github.com/metacloud/molecule/pull/409)
* molectule.yml.j2 template for molecule.yml
* playbooks: various playbooks used in testing
* playbook.yml: default playbook that gets used if not specified
* scenarios.yml: various combination of input ansible variable
* tests: tests run by the molecule itself

**Currently the solution is not runnable unless you build the docker image with proper repositories and rebuilt ansible with https://github.com/ansible/ansible/pull/24331**

* Molecule v2 with scenario support: https://github.com/openshift/openshift-ansible/pull/4354

**TODO**:
- [ ] - v2 issue: https://github.com/metacloud/molecule/issues/845
- [ ] - openshift excluder no longer excludes docker excluder? Should it be this way? https://github.com/openshift/origin/pull/12081
- [ ] - build docker image and store it into internal registry (with all needed yum repositories), if needed multiple docker images can be provided
- [ ] - split one repo file with a set of repos into more repo files with single repos (due to inifile so one can disable individual repos based on repo's filename)
- [ ] - make an exhausting list of excluder tests (cover various scenarios from installation to upgrade)
- [x] - find a way how to pass variables into ansible playbook
- [x] - parametrize `molecule.yml` by templating `molecule.yaml.j2`
- [x] - store all combination of input variables at one place a generate all possible `molecule.yml`
- [ ] - template `scenario.yml` if necessary (to avoid duplications), e.g. with `with_items` statement (the same as in ansible itself)
- [ ] - `ansible-lint` (https://github.com/willthames/ansible-lint) fails when run, run the lint over all ansible files to see what is wrong
- [ ] - extend openshift_excluder validation to check at most one of `["enable_excluders", "r_openshift_excluder_enable_excluders"]` is set (the same for `(enable_docker_excluder, openshift_enable_docker_excluder)` and `(enable_openshift_excluder, openshift_enable_openshift_excluder)` pair.
- [ ] backport https://github.com/ansible/ansible/pull/24331 to all RHEL (https://bugzilla.redhat.com/show_bug.cgi?id=1456602) and Fedora (https://bugzilla.redhat.com/show_bug.cgi?id=1456603)
- [ ] - junit, xunit testing

**NICE TO HAVE**
- [ ] generate test table from `scenarios.yml` to see which combination of input variables are covered (maybe as part of the README.md file)

**INFO**:
* molecule `1.24.0` needs docker installed (instead of docker-py), see https://github.com/metacloud/molecule/issues/771 #4354


roles/openshift_excluder/molecule.yml
```diff
@@ -0,0 +1,40 @@
...
+      dockerfile: Dockerfile
+    - name: openshift_excluder_enable
+      image: rhel7-ansible
+      image_version: latest
+      ansible_groups:
+        - openshift_excluder_enable
+      dockerfile: Dockerfile
+ansible:
+  config_file: ./ansible.cfg
+  extra-vars: >
```
ingvagabund: This will get populated with ansible variables (one combination per each generated `molecule.yml`)

roles/openshift_excluder/tests/ansible_vars.py
```diff
@@ -0,0 +1,3 @@
+# r_openshift_excluder_enable_excluders
+# r_openshift_excluder_enable_openshift_excluder
+r_openshift_excluder_enable_docker_excluder=False
```
ingvagabund: This will get generated from a `ansible_vars.py.j2` template

roles/openshift_excluder/tests/test_disable.py
```diff
@@ -0,0 +1,29 @@
+import testinfra.utils.ansible_runner
+import ansible_vars
```
ingvagabund: Current ansible variables

ingvagabund:
@sdodson @ashcrow @mtnbikenc @ewolinetz @giuseppe @detiber @abutcher @tbielawa 

Though, still PoC we can already discuss the methodology.

ashcrow:
At first glance this seems like an awful lot of lines for testing. Don't get me wrong, I'm all for getting unit like testing for our ansible code but I fear that if it takes more yaml, python, configuration and wiring for each area people will not spend the time to write solid tests.

ingvagabund:
> At first glance this seems like an awful lot of lines for testing. Don't get me wrong, I'm all for getting unit like testing for our ansible code but I fear that if it takes more yaml, python, configuration and wiring for each area people will not spend the time to write solid tests.

I completely agree. I don't like the way the yaml gets complicated and the additional python code around molecule. Though, there will always be some overhead. I hope there will be a lot of testing code to share between roles.

ashcrow:
Looking through the molecule code things make more sense. I the most important thing before we start using something like this is to decide on the layout of tests and, in the case of molecule, which verifier we will use. In this PR ``testinfra`` is used, and I think that's a good choice, though ``goss`` looks interesting as well.

In terms of layout I think having molecule setup for each role is fine, but I wonder if there is a way we can avoid some duplication. For instance, I have a feeling the images will be the same for most tests. Symlinks or preprocessing of the yaml to provide always up to date configurations for reusable configuration. In any case, we will need to document what we expect so that we can easily follow testing across roles.

I think the ``gas`` script is a fine start to a runner. With some tweaking we can make it something that can run tests in different paths so we don't need to link/copy it around.

roles/openshift_excluder/gas
```diff
@@ -0,0 +1,95 @@
...
+        return parser
+
+
+def renderTemplate(searchpath, template_file, template_vars):
+    templateLoader = jinja2.FileSystemLoader( searchpath=searchpath )
+    templateEnv = jinja2.Environment( loader=templateLoader )
+    template = templateEnv.get_template( template_file )
+    content = template.render( template_vars )
+    return content
+
```
ashcrow: pep8: needs another newline

roles/openshift_excluder/gas
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    output = renderTemplate(".", "molecule.yml.j2", vars)
+    # yaml formatting
+    output = yaml.load(output)
+    with open("molecule.yml", "w") as f:
+         yaml.dump(output, f, default_flow_style=False)
+
+    return call("molecule test", shell=True)
+
```
ashcrow: pep8: Needs another newline.

roles/openshift_excluder/gas
```diff
@@ -0,0 +1,95 @@
...
+        parser.add_option(
+            "", "-n", "--name", dest="name", default = "",
+            help = "Run specific task"
+        )
+
+        return parser
+
+
+def renderTemplate(searchpath, template_file, template_vars):
+    templateLoader = jinja2.FileSystemLoader( searchpath=searchpath )
```
ashcrow: pep8: Extra spaces were added around calls for the next few lines.

roles/openshift_excluder/gas
```diff
@@ -0,0 +1,95 @@
+#!/bin/python
```
ashcrow: nit: ```/usr/bin/env python``` to avoid forcing the system python in all cases 😄

roles/openshift_excluder/gas
```diff
@@ -0,0 +1,95 @@
...
+import logging
+from subprocess import call
+import optparse
+
+RED = '\033[91m'
+GREEN = '\033[92m'
+BLUE = '\033[94m'
+ENDC = '\033[0m'
+
+def setOptions():
```
ashcrow: nit: ``_`` are generally more accepted in Python. IE:

```python
def set_options():
    ...
```

ashcrow:
Fixes and enhancements for ``gas``: https://gist.github.com/ashcrow/498608161bcb73fe3c6c2b54f8d2ab16

rhcarvalho:
If this moves forward, I suggest moving the test runner into a more general path, not within a particular role (assuming it would actually be used for running tests in arbitrary roles across the repo).
BTW what does "gas" stand for?

ingvagabund:
> If this moves forward, I suggest moving the test runner into a more general path, not within a particular role (assuming it would actually be used for running tests in arbitrary roles across the repo).

Agree. I am still experimenting with molecule v2 in https://github.com/openshift/openshift-ansible/pull/4354. Making a list of missing features in both v1 and v2.

> BTW what does "gas" stand for?

Scenarios is a set of molecules. And gas is a set of molecules :).

michaelgugino:
Closing due to inactivity.

PR #4294, opened 2017-05-26
add ability to expose Elasticsearch as an external route

richm:
This makes the logging es route code work with the new
openshift_logging_elasticsearch role

This adds the ability to expose Elastisearch as a route outside of the
cluster.
- `openshift_logging_es_allow_external`: True (default is False) - if this is
  True, Elasticsearch will be exposed as a Route
- `openshift_logging_es_ops_hostname`: The external facing hostname to use for
  the route and the TLS server certificate (default is "es." +
  `openshift_master_default_subdomain`)

There are other similar parameters for the TLS server cert, key, and CA cert.
There are other similar parameters for when the OPS cluster is deployed e.g.
`openshift_logging_es_ops_allow_external`, etc.
@jcantrill @ewolinetz @noodle PTAL

### DIFF
--- a/roles/openshift_logging_elasticsearch/tasks/main.yaml
+++ b/roles/openshift_logging_elasticsearch/tasks/main.yaml
@@ -269,6 +269,75 @@
     - "{{ tempdir }}/templates/logging-es-dc.yml"
     delete_after: true
 
+- name: Retrieving the cert to use when generating secrets for the {{ es_component }} component
+  slurp:
+    src: "{{ generated_certs_dir }}/{{ item.file }}"
+  register: key_pairs
+  with_items:
+  - { name: "ca_file", file: "ca.crt" }
+  - { name: "es_key", file: "system.logging.es.key" }
+  - { name: "es_cert", file: "system.logging.es.crt" }
+  when: openshift_logging_es_allow_external | bool
+
+- set_fact:
+    es_key: "{{ lookup('file', openshift_logging_es_key) | b64encode }}"
+  when:
+  - openshift_logging_es_key | trim | length > 0
+  - openshift_logging_es_allow_external | bool
+  changed_when: false
+
+- set_fact:
+    es_cert: "{{ lookup('file', openshift_logging_es_cert) | b64encode  }}"
+  when:
+  - openshift_logging_es_cert | trim | length > 0
+  - openshift_logging_es_allow_external | bool
+  changed_when: false
+
+- set_fact:
+    es_ca: "{{ lookup('file', openshift_logging_es_ca_ext) | b64encode  }}"
+  when:
+  - openshift_logging_es_ca_ext | trim | length > 0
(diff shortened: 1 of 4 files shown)


roles/openshift_logging_elasticsearch/tasks/main.yaml
```diff
@@ -269,6 +269,75 @@
...
+  changed_when: false
+
+- set_fact:
+    es_ca: "{{ key_pairs | entry_from_named_pair('ca_file') }}"
+  when:
+  - es_ca is not defined
+  - openshift_logging_es_allow_external | bool
+  changed_when: false
+
+- name: Generating Elasticsearch {{ es_component }} route template
```
ewolinetz: is it possible to use the `oc_route` module instead of generating the template and then using `oc_obj` ?

roles/openshift_logging_elasticsearch/tasks/main.yaml
```diff
@@ -269,6 +269,75 @@
     - "{{ tempdir }}/templates/logging-es-dc.yml"
     delete_after: true
 
+- name: Retrieving the cert to use when generating secrets for the {{ es_component }} component
```
ewolinetz: if we switch to `oc_route` we should be able to point to the file destination instead of slurping them in

roles/openshift_logging_elasticsearch/tasks/main.yaml
richm: @ewolinetz not possible to use oc_route - no support for labels - I suppose once https://github.com/openshift/openshift-ansible/pull/4091 we can add support for labels to oc_route.

richm: We should also change kibana to use `oc_route`.

richm:
> track that the change to oc_route should be taken on as tech-debt

Done.

richm:
What's going on with this PR?  The checks have been pending for several days . . . can we cancel the pending tasks and kick off new ones?  Will they eventually go through?  How long will it take?

sdodson:
aos-ci-test

sdodson:
Didn't see anything pending, just needed to be kicked off as far as I can tell.

richm:
Is there anything I can do to speed up the merge?

sdodson:
aos-ci-test

sdodson:
PRs to openshift-ansible need the following, 1a and 2 you should have rights to do.

1)
  1a) aos-ci-test results to be green, basically everything that shows up as Required in the test list. pushing any commits/rebases invalidates test results and unfortunately they must be manually re-triggered
  1b) positive review (ewolinetz provided that)
2) [ merge ] tag

Sorry this is a pain, we'll hopefully do away with the two testing infra soon.

richm:
[merge]

sdodson:
[merge][severity: bug]

PR #4308, opened 2017-05-29
openshift_logging: prevent under resourced ES pod

wozniakjan:
The #4288 fixed default CPU and memory values for ES pod. This fixes the outdated README and enforces lower bounds for user selected CPU and memory limitations to prevent faulty behavior.

Available units and their explanation is taken from https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-ram-container/


wozniakjan:
@bmcelvee the desription for `openshift_logging_es_cpu_limit` and  `openshift_logging_es_memory_limit` may need to change in the docs if this gets merged

https://docs.openshift.org/latest/install_config/aggregate_logging.html

ewolinetz:
Would the throttle filter be useful for the other components as well?

wozniakjan:
that is very well possible, I could not find any other component needing it or using anything similar. Could you advise more suitable location for the code?

ewolinetz:
it might make sense to move it to the top level repo `filter_plugins` directory and rename to `oo_*` under `oo_filters.py`

wozniakjan:
done, thanks!

ewolinetz:
@sdodson or @abutcher any issues with the new filter plugins?

@wozniakjan can we also use this filter with the other logging components in this PR (e.g. kibana, fluentd, curatator)?

wozniakjan:
@ewolinetz yes, we can. But given the other logging components are not structured the same way ES is and [ansible variable precedence](http://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence), there are two options

1) put the filter in `tasks` (not to defaults `defaults` like in ES) where the values are passed as different variables to the dc generators
2) introduce one more layer of variables in each component just like there is with `*_es_*` and `*_elasticsearch_*` to `defaults`, and then have the filters in the same location for each logging component

I will start with the first option for now unless someone thinks it is not a good idea

ewolinetz:
@wozniakjan sorry, I'm not following.. Why would we be unable to do the same for defaults in the other components similar to how you're doing with ES? 

I may be reading too much into this though... @portante is it necessary to apply this for the other components or is it good to go as is?

roles/openshift_logging/README.md
```diff
@@ -63,8 +63,8 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_es_client_key`: The location of the client key Fluentd uses for openshift_logging_es_host. Defaults to '/etc/fluent/keys/key'.
 
 - `openshift_logging_es_cluster_size`: The number of ES cluster members. Defaults to '1'.
-- `openshift_logging_es_cpu_limit`:  The amount of CPU limit for the ES cluster.  Unused if not set
-- `openshift_logging_es_memory_limit`: The amount of RAM that should be assigned to ES. Defaults to '8Gi'.
+- `openshift_logging_es_cpu_limit`:  The amount of CPU limit for the ES cluster. Defaults to no limitation, unused if not set and lower bound is 1 core.
```
portante: Consider having this read: "The upper limit of CPU usage allowed for a given Elasticsearch pod. Defaults to unlimited; unused if not set; lower bound is 1 core."

filter_plugins/oo_filters.py
```diff
@@ -994,6 +994,60 @@ def oo_random_word(length, source='abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRS
     return ''.join(random.choice(source) for i in range(length))
 
 
+def norm(value, prefix_map):
+    ''' Normalize value based on the prefix_map '''
+    value = value.strip(' ')
+    match = re.search('([0-9.]*)(.*)', value)
+    val = match.group(1)
+    unit = match.group(2).strip(' ')
+    if unit not in prefix_map:
```
portante: Not that this is performance sensitive, but using try/except is better than a double test:

```
try:
    return float(val) * prefix_map[unit]
except KeyError:
    raise ValueError('Unit ' + unit + ' not supported')
```

wozniakjan:
> sorry, I'm not following.. Why would we be unable to do the same for defaults in the other components similar to how you're doing with ES?

@ewolinetz because both inventory and command line can override the defaults. For elasticsearch, user sets `*_es_*` according to what he reads in documentation, filter is executed on `*_elasticsearch_*` and these longer named variables are later used in the role. If user was to set `*_elasticsearch_*`, the [variable precedence](http://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable) would prevent us from any restrictions or filtering. Other logging components don't have this 'two layered variables design' in the defaults from what I was able to find out.

> I am a bit torn about lower bounds. Could we have a warning issued instead, reporting what the lower bound is, and how far below the lower bound the value was found to be?

@portante of course, and would you like the playbook to only report the warning and continue successfully even if user throttles the pod too much?

PR #4309, opened 2017-05-30
disable docker excluder before it is updated to remove older excluded packages

ingvagabund:
When updating excluders from 3.3 to 3.4 version the /etc/yum.conf still excludes `docker-1.12*` packages even if the excluding expressing is no longer present in 3.4 docker excluders.
It happens when the docker excluder is still excluding. The older excluding expressions will never get removed after the upgrade.

### DIFF
--- a/roles/openshift_excluder/tasks/disable.yml
+++ b/roles/openshift_excluder/tasks/disable.yml
@@ -5,10 +5,12 @@
     include: verify_upgrade.yml
 
 # unexclude the current openshift/origin-excluder if it is installed so it can be updated
-- name: Disable OpenShift excluder so it can be updated
+- name: Disable excluders before the upgrade to remove older excluding expressions
   include: unexclude.yml
   vars:
-    unexclude_docker_excluder: false
+    # before the docker excluder can be updated, it needs to be disabled
+    # to remove older excluded packages that are no longer excluded
+    unexclude_docker_excluder: "{{ r_openshift_excluder_enable_docker_excluder }}"
     unexclude_openshift_excluder: "{{ r_openshift_excluder_enable_openshift_excluder }}"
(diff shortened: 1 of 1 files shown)


ingvagabund:
@sdodson @ashcrow @mtnbikenc 

Not a blocker until `docker-1.12*` is removed from docker excluders. Though, this will need backport to all branches.

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

sdodson:
Hmm, which version of the 3.3 excluders did you have installed? I think this is a bug in the 3.3 excluders that we could workaround but I thought this was fixed quite some time ago, if it's not actually fixed in the excluders we need to know.

ingvagabund:
atomic-openshift-docker-excluder-3.3.1.31-1.git.0.f6ab12a.el7.noarch

ingvagabund:
Checking atomic-openshift specs for
```spec
%pretrans excluder
# we always want to clear this out using the last
#   versions script.  Otherwise excludes might get left in
if [ -s /usr/sbin/%{name}-excluder ] ; then
  /usr/sbin/%{name}-excluder unexclude
fi
```

- 3.5 applied
- 3.4 applied
- 3.3 not applied

Yeah, 3.3 is missing https://github.com/openshift/origin/pull/13392/files#diff-3477fbcab168ba3ee773ffeb9323e53eR555

sdodson:
[merge][severity: blocker]

sdodson:
Looks like we also need to fix 3.4 packaging. This PR seems like a decent workaround that will continue to work even after the packages are fixed so merging.

PR #4311, opened 2017-05-30
Skip router/registry cert redeploy when openshift_hosted_manage_{router,registry}=false

abutcher:
Setting the following will cause `playbooks/byo/openshift-cluster/redeploy-certificates.yml` to skip redeploying router and registry certificates. Note that these are the same variables which the installation playbook uses to skip router and registry management.

```
openshift_hosted_manage_router=false
openshift_hosted_manage_registry=false
```
/cc @stenwt


sdodson:
aos-ci-test

abutcher:
aos-ci-test

sdodson:
[merge][severity: blocker]

PR #4312, opened 2017-05-30
Ensure first etcd url listed in master config is local

### LINKED ISSUE #4293: Set first etcd url to be local (to the master)
If etcd is colocated in the master, docs recommend setting the URL in master-config to be the local etcd

https://docs.openshift.org/latest/scaling_performance/host_practices.html

Currently it seems to only rank by the order based on what is listed in the ansible inventory

abutcher:
Ensure the first etcd server listed in master-config.yaml is the local etcd instance when etcd is colocated with the master.

Fixes #4293

### DIFF
--- a/roles/openshift_master/templates/master.yaml.v1.j2
+++ b/roles/openshift_master/templates/master.yaml.v1.j2
@@ -82,7 +82,7 @@ etcdClientInfo:
   certFile: master.etcd-client.crt
   keyFile: master.etcd-client.key
   urls:
-{% for etcd_url in openshift.master.etcd_urls %}
+{% for etcd_url in openshift.master.etcd_urls | oo_sort_etcd_urls_local_first(openshift.common.hostname) %}
     - {{ etcd_url }}
 {% endfor %}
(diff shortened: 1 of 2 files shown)


sdodson:
@ncdc or @sttts Can you confirm that this actually matters? I think this came up elsewhere and we decided that it didn't actually matter.

abutcher:
Referenced in our docs here https://docs.openshift.org/latest/scaling_performance/host_practices.html#scaling-performance-capacity-host-practices-master

ncdc:
I'm not sure this will make much of a difference. It looks like we use the default endpoint selection mode, which is `EndpointSelectionRandom`. See https://github.com/coreos/etcd/blob/02f4a9a034f65f588599eaf6d756b98b55640c98/client/client.go#L59-L80 for more details.

abutcher:
Closing! Should we remove that bullet point from the docs?

ncdc:
@smarterclayton do you agree with my assessment?

sdodson:
Yeah, I think we should clean up docs. We've had a few people ask about this recently, I think there's a BZ for it too.

PR #4313, opened 2017-05-30
WIP: Idea for adding some registry checks to help with configuration mistakes

screeley44:
@sdodson @jarrpa - this is an idea for helping users that accidentally misconfigure an inventory/hosts file and catch things before the playbook runs for 20 minutes.  For example, if they run a playbook, and forget a key variable setting or group for nfs or glusterfs registry, the playbook will either run the whole way through without any errors but also without installing what they intended or the playbook will run for a long time and eventually fail because of a misconfiguration, so these checks will run first and stop the playbook if something doesn't look right.

With that said, not sure if I added these to the right place or if there is a better way to handle this as I'm an ansible novice, but the sanitize_inventory role seemed like a good fit for this?  Let  me know if you think this is worth while.

### DIFF
--- a/roles/openshift_sanitize_inventory/tasks/main.yml
+++ b/roles/openshift_sanitize_inventory/tasks/main.yml
@@ -46,3 +46,139 @@
     msg: |-
       openshift_release is "{{ openshift_release }}" which is not a valid version string.
       Please set it to a version string like "3.4".
+
+# Add additional registry configuration checks
+# ================================== BOTH NFS AND GLUSTERFS ======================================
+# Inventory/hosts file has both glusterfs_registry and nfs registry enabled - user should 
+# choose which one they really want.
+- name: Abort when both nfs and glusterfs_registry are enabled
+  when:
+    - "'nfs' in group_names"
+    - "'glusterfs_registry' in group_names"
+  fail:
+    msg: |-
+      Conflicting registry types are defined, 'nfs' and 'glusterfs_registry' group_names 
+      are both defined in the inventory/hosts file. Please choose one or the other and rerun the playbook. 
+
+# ============================  GLUSTERFS ====================================
+# Inventory/hosts file indicates that registry storage kind is glusterfs
+# but glusterfs_registry does not exist in the group_names OSEv3
+# Y, NA, N
+- name: Abort when registry_storage_kind is glusterfs but no groups exist
+  when:
+    - openshift_hosted_registry_storage_kind == 'glusterfs'
+    - "'glusterfs_registry' not in group_names"
+    - "'glusterfs' not in group_names"
+  fail:
+    msg: |-
+      openshift_hosted_registry_storage_kind=glusterfs is enabled but there
+      is no 'glusterfs_registry' group_name in OSEv3:children defined in the
+      inventory/hosts file.  Please add 'glusterfs_registry' group_name or
+      remove the [glusterfs_registry] hosts group and rerun the playbook.
+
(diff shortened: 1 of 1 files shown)


sdodson:
Reading roles/openshift_sanitize_inventory/README.md I think that's the intention of this role. We have cards to start moving validation to a centralized location too.

https://trello.com/c/YOSlM0x8/444-2-introduce-openshift-config-validation-role

@sosiouxme any objection to making this role the place to validate config variables?

sosiouxme:
@sdodson that is the intent, yeah!

It does require a little caution since so many things have this as a dependency, and if you were checking something specific to an optional component, you'd want to make sure it was relevant to the present playbook run before complaining something wasn't set. And this requires some tight coupling to know what's going to be configured by other roles. So I'd limit it to the simplest and most effective things you can check for users.

roles/openshift_sanitize_inventory/tasks/main.yml
```diff
@@ -46,3 +46,139 @@
...
+  when:
+    - openshift_hosted_registry_storage_kind == 'glusterfs'
+    - "'glusterfs_registry' not in group_names"
+    - "'glusterfs' not in group_names"
+  fail:
+    msg: |-
+      openshift_hosted_registry_storage_kind=glusterfs is enabled but there
+      is no 'glusterfs_registry' group_name in OSEv3:children defined in the
+      inventory/hosts file.  Please add 'glusterfs_registry' group_name or
+      remove the [glusterfs_registry] hosts group and rerun the playbook.
```
jarrpa: The error message should indicate that the `glusterfs` group would also be valid.

roles/openshift_sanitize_inventory/tasks/main.yml
```diff
@@ -46,3 +46,139 @@
     msg: |-
       openshift_release is "{{ openshift_release }}" which is not a valid version string.
       Please set it to a version string like "3.4".
+
+# Add additional registry configuration checks
+# ================================== BOTH NFS AND GLUSTERFS ======================================
+# Inventory/hosts file has both glusterfs_registry and nfs registry enabled - user should 
+# choose which one they really want.
+- name: Abort when both nfs and glusterfs_registry are enabled
```
jarrpa: This check is wrong. It's entirely possible to have both groups (e.g. configure an NFS registry with a glusterfs_registry group primed for being switched to later (obtuse, I know, but should be possible)).

roles/openshift_sanitize_inventory/tasks/main.yml
```diff
@@ -46,3 +46,139 @@
...
+      is no 'glusterfs_registry' group_name in OSEv3:children defined in the 
+      inventory/hosts file. Please add 'glusterfs_registry' group_name to get an 
+      glusterfs registry or remove the openshift_hosted_registry_storage_kind=glusterfs variable
+      and [glusterfs_registry] hosts group to not install glusterfs registry and rerun the playbook.
+
+# Inventory/hosts file indicates that registry storage kind is glusterfs
+# but [glusterfs_registry] hosts group does not exist
+# +regstorage  -[glusterfs_registry]  +glusterfs_registry
+# Y, N, Y
+- name: Abort when glusterfs_registry group and group_name are defined but registry storage does not match
```
jarrpa: This seems like a duplicate of the first one... even so, it should also be allowed.

michaelgugino:
Closing due to inactivity.

PR #4315, opened 2017-05-30
Removing logging and metrics hosted templates

ewolinetz:
To address https://bugzilla.redhat.com/show_bug.cgi?id=1456348
We no longer need to use/provide deployer templates for logging or metrics in [1|3].5+


sdodson:
Lets get this into master first though, just as matter of process we should always get master first.

ewolinetz:
@sdodson looks like the files are already removed in master

PR #4316, opened 2017-05-30
add check to detect increased etcd traffic

juanvallejo:
Related trello card: https://trello.com/c/7pJdHj4F
Related bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1415839

An update in OSE 3.4 caused an increase in etcd traffic due to increased quorum reads.
This check scans etcd logs (`journalctl -ru <unit>`) and recommends an upgrade to `OSE 3.6` if the message `etcd: sync duration of ..., expected less than 1s` is found (this message is a direct symptom of the increased traffic update). 

Based on [oadm diagnostics](https://github.com/openshift/origin/blob/master/pkg/diagnostics/systemd/analyze_logs.go)

TODO

- [x] add unit tests

### DIFF
--- a/roles/openshift_health_checker/library/search_journalctl.py
+++ b/roles/openshift_health_checker/library/search_journalctl.py
@@ -1,0 +1,150 @@
+#!/usr/bin/python
+"""Interface to journalctl."""
+
+from time import time
+import json
+import re
+import subprocess
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+class InvalidMatcherRegexp(Exception):
+    """Exception class for invalid matcher regexp."""
+    pass
+
+
+class InvalidLogEntry(Exception):
+    """Exception class for invalid / non-json log entries."""
+    pass
+
+
+class LogInputSubprocessError(Exception):
+    """Exception class for errors that occur while executing a subprocess."""
+    pass
+
+
+def main():
+    """Scan a given list of "log_matchers" for journalctl messages containing given patterns.
+    "log_matchers" is a list of dicts consisting of three keys that help fine-tune log searching:
+    'start_regexp', 'regexp', and 'unit'.
+
+    Sample "log_matchers" list:
+
+    [
+      {
+        'start_regexp': r'Beginning of systemd unit',
+        'regexp': r'the specific log message to find',
+        'unit': 'etcd',
+      }
+    ]
+    """
+    module = AnsibleModule(
+        argument_spec=dict(
+            log_count_limit=dict(type="int", default=500),
+            log_matchers=dict(type="list", required=True),
+        ),
+    )
+
+    timestamp_limit_seconds = time() - 60 * 60  # 1 hour
+
+    log_count_limit = module.params["log_count_limit"]
+    log_matchers = module.params["log_matchers"]
+
+    matched_regexp, errors = get_log_matches(log_matchers, log_count_limit, timestamp_limit_seconds)
+
+    module.exit_json(
+        changed=False,
+        failed=bool(errors),
+        errors=errors,
+        matched=matched_regexp,
+    )
+
+
+def get_log_matches(matchers, log_count_limit, timestamp_limit_seconds):
+    """Return a list of up to log_count_limit matches for each matcher.
+
+    Log entries are only considered if newer than timestamp_limit_seconds.
+    """
+    matched_regexp = []
+    errors = []
+
+    for matcher in matchers:
+        try:
+            log_output = get_log_output(matcher)
+        except LogInputSubprocessError as err:
+            errors.append(str(err))
+            continue
+
+        try:
+            matched = find_matches(log_output, matcher, log_count_limit, timestamp_limit_seconds)
+            if matched:
+                matched_regexp.append(matcher.get("regexp", ""))
+        except InvalidMatcherRegexp as err:
+            errors.append(str(err))
+        except InvalidLogEntry as err:
+            errors.append(str(err))
+
+    return matched_regexp, errors
+
+
+def get_log_output(matcher):
+    """Return an iterator on the logs of a given matcher."""
+    try:
+        cmd_output = subprocess.Popen(list([
+            '/bin/journalctl',
+            '-ru', matcher.get("unit", ""),
+            '--output', 'json',
+        ]), stdout=subprocess.PIPE)
+
+        return iter(cmd_output.stdout.readline, '')
+
+    except subprocess.CalledProcessError as exc:
+        msg = "Could not obtain journalctl logs for the specified systemd unit: {}: {}"
+        raise LogInputSubprocessError(msg.format(matcher.get("unit", "<missing>"), str(exc)))
+    except OSError as exc:
+        raise LogInputSubprocessError(str(exc))
+
+
+def find_matches(log_output, matcher, log_count_limit, timestamp_limit_seconds):
+    """Return log messages matched in iterable log_output by a given matcher.
+
+    Ignore any log_output items older than timestamp_limit_seconds.
+    """
+    try:
+        regexp = re.compile(matcher.get("regexp", ""))
+        start_regexp = re.compile(matcher.get("start_regexp", ""))
+    except re.error as err:
+        msg = "A log matcher object was provided with an invalid regular expression: {}"
+        raise InvalidMatcherRegexp(msg.format(str(err)))
+
+    matched = None
+
+    for log_count, line in enumerate(log_output):
+        if log_count >= log_count_limit:
+            break
+
+        try:
+            obj = json.loads(line)
+
+            # don't need to look past the most recent service restart
+            if start_regexp.match(obj["MESSAGE"]):
+                break
+
+            log_timestamp_seconds = float(obj["__REALTIME_TIMESTAMP"]) / 1000000
+            if log_timestamp_seconds < timestamp_limit_seconds:
+                break
+
+            if regexp.match(obj["MESSAGE"]):
+                matched = line
+                break
+
+        except ValueError:
+            msg = "Log entry for systemd unit {} contained invalid json syntax: {}"
+            raise InvalidLogEntry(msg.format(matcher.get("unit"), line))
+
+    return matched
+
+
+if __name__ == '__main__':
+    main()
--- a/roles/openshift_health_checker/openshift_checks/etcd_traffic.py
+++ b/roles/openshift_health_checker/openshift_checks/etcd_traffic.py
@@ -1,0 +1,47 @@
+"""Check that scans journalctl for messages caused as a symptom of increased etcd traffic."""
+
+from openshift_checks import OpenShiftCheck, get_var
+
+
+class EtcdTraffic(OpenShiftCheck):
+    """Check if host is being affected by an increase in etcd traffic."""
+
+    name = "etcd_traffic"
+    tags = ["health", "etcd"]
+
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip hosts that do not have etcd in their group names."""
+        group_names = get_var(task_vars, "group_names", default=[])
+        valid_group_names = "etcd" in group_names
+
+        version = get_var(task_vars, "openshift", "common", "short_version")
+        valid_version = version in ("3.4", "3.5", "1.4", "1.5")
+
+        return super(EtcdTraffic, cls).is_active(task_vars) and valid_group_names and valid_version
+
+    def run(self, tmp, task_vars):
+        is_containerized = get_var(task_vars, "openshift", "common", "is_containerized")
+        unit = "etcd_container" if is_containerized else "etcd"
+
+        log_matchers = [{
+            "start_regexp": r"Starting Etcd Server",
+            "regexp": r"etcd: sync duration of [^,]+, expected less than 1s",
+            "unit": unit
+        }]
+
+        match = self.execute_module("search_journalctl", {
+            "log_matchers": log_matchers,
+        }, task_vars)
+
+        if match.get("matched"):
+            msg = ("Higher than normal etcd traffic detected.\n"
+                   "OpenShift 3.4 introduced an increase in etcd traffic.\n"
+                   "Upgrading to OpenShift 3.6 is recommended in order to fix this issue.\n"
+                   "Please refer to https://access.redhat.com/solutions/2916381 for more information.")
+            return {"failed": True, "msg": msg}
+
+        if match.get("failed"):
+            return {"failed": True, "msg": "\n".join(match.get("errors"))}
+
+        return {}
--- a/roles/openshift_health_checker/test/etcd_traffic_test.py
+++ b/roles/openshift_health_checker/test/etcd_traffic_test.py
@@ -1,0 +1,80 @@
+import pytest
+
+from openshift_checks.etcd_traffic import EtcdTraffic
+
+
+@pytest.mark.parametrize('group_names,version,is_active', [
+    (['masters'], "3.5", False),
+    (['masters'], "3.6", False),
+    (['nodes'], "3.4", False),
+    (['etcd'], "3.4", True),
+    (['etcd'], "3.5", True),
+    (['etcd'], "3.1", False),
+    (['masters', 'nodes'], "3.5", False),
+    (['masters', 'etcd'], "3.5", True),
+    ([], "3.4", False),
+])
+def test_is_active(group_names, version, is_active):
+    task_vars = dict(
+        group_names=group_names,
+        openshift=dict(
+            common=dict(short_version=version),
+        ),
+    )
+    assert EtcdTraffic.is_active(task_vars=task_vars) == is_active
+
+
+@pytest.mark.parametrize('group_names,matched,failed,extra_words', [
+    (["masters"], True, True, ["Higher than normal", "traffic"]),
+    (["masters", "etcd"], False, False, []),
+    (["etcd"], False, False, []),
+])
+def test_log_matches_high_traffic_msg(group_names, matched, failed, extra_words):
+    def execute_module(module_name, args, task_vars):
+        return {
+            "matched": matched,
+            "failed": failed,
+        }
+
+    task_vars = dict(
+        group_names=group_names,
+        openshift=dict(
+            common=dict(service_type="origin", is_containerized=False),
+        )
+    )
+
+    check = EtcdTraffic(execute_module=execute_module)
+    result = check.run(tmp=None, task_vars=task_vars)
+
+    for word in extra_words:
+        assert word in result.get("msg", "")
+
+    assert result.get("failed", False) == failed
+
+
+@pytest.mark.parametrize('is_containerized,expected_unit_value', [
+    (False, "etcd"),
+    (True, "etcd_container"),
+])
+def test_systemd_unit_matches_deployment_type(is_containerized, expected_unit_value):
+    task_vars = dict(
+        openshift=dict(
+            common=dict(is_containerized=is_containerized),
+        )
+    )
+
+    def execute_module(module_name, args, task_vars):
+        assert module_name == "search_journalctl"
+        matchers = args["log_matchers"]
+
+        for matcher in matchers:
+            assert matcher["unit"] == expected_unit_value
+
+        return {"failed": False}
+
+    check = EtcdTraffic(execute_module=execute_module)
+    check.run(tmp=None, task_vars=task_vars)
+
+
+def fake_execute_module(*args):
+    raise AssertionError('this function should not be called')
--- a/roles/openshift_health_checker/test/search_journalctl_test.py
+++ b/roles/openshift_health_checker/test/search_journalctl_test.py
@@ -1,0 +1,157 @@
+import pytest
+import search_journalctl
+
+
+def canned_search_journalctl(get_log_output=None):
+    """Create a search_journalctl object with canned get_log_output method"""
+    module = search_journalctl
+    if get_log_output:
+        module.get_log_output = get_log_output
+    return module
+
+
+DEFAULT_TIMESTAMP = 1496341364
+
+
+def get_timestamp(modifier=0):
+    return DEFAULT_TIMESTAMP + modifier
+
+
+def get_timestamp_microseconds(modifier=0):
+    return get_timestamp(modifier) * 1000000
+
+
+def create_test_log_object(stamp, msg):
+    return '{{"__REALTIME_TIMESTAMP": "{}", "MESSAGE": "{}"}}'.format(stamp, msg)
+
+
+@pytest.mark.parametrize('name,matchers,log_input,expected_matches,expected_errors', [
+    (
+        'test with valid params',
+        [
+            {
+                "start_regexp": r"Sample Logs Beginning",
+                "regexp": r"test log message",
+                "unit": "test",
+            },
+        ],
+        [
+            create_test_log_object(get_timestamp_microseconds(), "test log message"),
+            create_test_log_object(get_timestamp_microseconds(), "Sample Logs Beginning"),
+        ],
+        ["test log message"],
+        [],
+    ),
+    (
+        'test with invalid json in log input',
+        [
+            {
+                "start_regexp": r"Sample Logs Beginning",
+                "regexp": r"test log message",
+                "unit": "test-unit",
+            },
+        ],
+        [
+            '{__REALTIME_TIMESTAMP: ' + str(get_timestamp_microseconds()) + ', "MESSAGE": "test log message"}',
+        ],
+        [],
+        [
+            ["invalid json", "test-unit", "test log message"],
+        ],
+    ),
+    (
+        'test with invalid regexp',
+        [
+            {
+                "start_regexp": r"Sample Logs Beginning",
+                "regexp": r"test [ log message",
+                "unit": "test",
+            },
+        ],
+        [
+            create_test_log_object(get_timestamp_microseconds(), "test log message"),
+            create_test_log_object(get_timestamp_microseconds(), "sample log message"),
+            create_test_log_object(get_timestamp_microseconds(), "fake log message"),
+            create_test_log_object(get_timestamp_microseconds(), "dummy log message"),
+            create_test_log_object(get_timestamp_microseconds(), "Sample Logs Beginning"),
+        ],
+        [],
+        [
+            ["invalid regular expression"],
+        ],
+    ),
+], ids=lambda argval: argval[0])
+def test_get_log_matches(name, matchers, log_input, expected_matches, expected_errors):
+    def get_log_output(matcher):
+        return log_input
+
+    module = canned_search_journalctl(get_log_output)
+    matched_regexp, errors = module.get_log_matches(matchers, 500, 60 * 60)
+
+    assert set(matched_regexp) == set(expected_matches)
+    assert len(expected_errors) == len(errors)
+
+    for idx, partial_err_set in enumerate(expected_errors):
+        for partial_err_msg in partial_err_set:
+            assert partial_err_msg in errors[idx]
+
+
+@pytest.mark.parametrize('name,matcher,log_count_lim,stamp_lim_seconds,log_input,expected_match', [
+    (
+        'test with matching log message, but out of bounds of log_count_lim',
+        {
+            "start_regexp": r"Sample Logs Beginning",
+            "regexp": r"dummy log message",
+            "unit": "test",
+        },
+        3,
+        get_timestamp(-100 * 60 * 60),
+        [
+            create_test_log_object(get_timestamp_microseconds(), "test log message"),
+            create_test_log_object(get_timestamp_microseconds(), "sample log message"),
+            create_test_log_object(get_timestamp_microseconds(), "fake log message"),
+            create_test_log_object(get_timestamp_microseconds(), "dummy log message"),
+            create_test_log_object(get_timestamp_microseconds(), "Sample Logs Beginning"),
+        ],
+        None,
+    ),
+    (
+        'test with matching log message, but with timestamp too old',
+        {
+            "start_regexp": r"Sample Logs Beginning",
+            "regexp": r"dummy log message",
+            "unit": "test",
+        },
+        100,
+        get_timestamp(-10),
+        [
+            create_test_log_object(get_timestamp_microseconds(), "test log message"),
+            create_test_log_object(get_timestamp_microseconds(), "sample log message"),
+            create_test_log_object(get_timestamp_microseconds(), "fake log message"),
+            create_test_log_object(get_timestamp_microseconds(-1000), "dummy log message"),
+            create_test_log_object(get_timestamp_microseconds(-1000), "Sample Logs Beginning"),
+        ],
+        None,
+    ),
+    (
+        'test with matching log message, and timestamp within time limit',
+        {
+            "start_regexp": r"Sample Logs Beginning",
+            "regexp": r"dummy log message",
+            "unit": "test",
+        },
+        100,
+        get_timestamp(-1010),
+        [
+            create_test_log_object(get_timestamp_microseconds(), "test log message"),
+            create_test_log_object(get_timestamp_microseconds(), "sample log message"),
+            create_test_log_object(get_timestamp_microseconds(), "fake log message"),
+            create_test_log_object(get_timestamp_microseconds(-1000), "dummy log message"),
+            create_test_log_object(get_timestamp_microseconds(-1000), "Sample Logs Beginning"),
+        ],
+        create_test_log_object(get_timestamp_microseconds(-1000), "dummy log message"),
+    ),
+], ids=lambda argval: argval[0])
+def test_find_matches_skips_logs(name, matcher, log_count_lim, stamp_lim_seconds, log_input, expected_match):
+    match = search_journalctl.find_matches(log_input, matcher, log_count_lim, stamp_lim_seconds)
+    assert match == expected_match


roles/openshift_health_checker/library/etcdlogs.py
```diff
@@ -0,0 +1,95 @@
...
+            log_count_limit=dict(type="int", default=500),
+            start_match=dict(type="str", default="Starting Etcd Server"),
+            log_matcher=dict(type="dict", required=True),
+            output=dict(type="str", default="json"),
+        ),
+    )
+
+    cmd = [
+        '/bin/journalctl',
+        '-ru', 'etcd',
```
juanvallejo: @sosiouxme will make this a module param. Not sure what the value should be

juanvallejo:
aos-ci-test

juanvallejo:
@sosiouxme will begin adding tests

roles/openshift_health_checker/library/etcdlogs.py
sosiouxme: The unit? It'll be `etcd` for multi-master clusters or the master unit name which depends on the deployment type

roles/openshift_health_checker/library/etcdlogs.py
```diff
@@ -0,0 +1,95 @@
...
+        '--output', module.params["output"],
```
sosiouxme: no real reason to parameterize this, you'll always want json because it's machine-readable that way

roles/openshift_health_checker/library/etcdlogs.py
```diff
@@ -0,0 +1,95 @@
+# pylint: disable=missing-docstring
+
+"""Interface to journalctl"""
```
sosiouxme: hard to believe there isn't a module for journalctl... guess it just shows, no one is doing monitoring/diagnostics with ansible yet

roles/openshift_health_checker/library/etcdlogs.py
```diff
@@ -0,0 +1,95 @@
...
+
+        for line in iter(cmd_output.stdout.readline, ''):
+            log_count += 1
+            if log_count >= log_count_limit:
+                break
+
+            try:
+                obj = json.loads(line.rstrip())
+                if start_matcher.match(obj["MESSAGE"]):
+                    break
```
sosiouxme: `# don't need to look past the most recent service restart`

roles/openshift_health_checker/library/etcdlogs.py
sosiouxme: there's actually nothing about this module that's specific to the task at hand; it could probably be renamed to search_journal.py or some such. although in a more generic implementation you'd probably search for several things at the same time in a single unit's journal. but we can cross that bridge later.

roles/openshift_health_checker/library/etcdlogs.py
sosiouxme: (just because that may not be obvious)

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,35 @@
+# pylint: disable=missing-docstring
```
sosiouxme: don't do that...

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,35 @@
...
+    """Check if host is being affected by an increase in etcd traffic."""
+
+    name = "etcd_traffic"
+    tags = ["health", "etcd"]
+
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip hosts that do not have etcd or masters in their group names."""
+        group_names = get_var(task_vars, "group_names", default=[])
+        active = "masters" in group_names or "etcd" in group_names
```
sosiouxme: so, it's masters if there's only one master; etcd if there are multiple masters. there's a variable for figuring out what's in all the groups - maybe just `groups`?

it's not exactly critical to get this right in `is_active` but we might as well.

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,35 @@
...
+        return super(EtcdTraffic, cls).is_active(task_vars) and active
+
+    def run(self, tmp, task_vars):
+        match = self.module_executor("etcdlogs", {
+            "log_matcher": {
+                "regexp": "etcd: sync duration of \d+\.\d+s, expected less than 1s",  # pylint: disable=anomalous-backslash-in-string
```
sosiouxme: use `r'` so the backslashes aren't misinterpreted - or does that not work across the module invocation? i think it should be fine.

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,35 @@
...
+                "level": "warning",
+            }
+        }, task_vars)
+
+        if match["matched"]:
+            msg = ("Higher than normal etcd traffic detected. "
+                   "OpenShift 3.4 introduced an increase in etcd traffic by at least a factor of 4."
+                   "\nUpgrading to OpenShift 3.6 is recommended in order to fix this issue.")
```
sosiouxme: Pending an upgrade there are a few tricks for mitigating this; can we mention those too? Or best, refer to a kbase if we have one.

roles/openshift_health_checker/library/etcdlogs.py
juanvallejo: > although in a more generic implementation you'd probably search for several things at the same time in a single unit's journal. but we can cross that bridge later.

@sosiouxme yeah, was thinking of implementing this, but was not sure how to handle "callbacks" for different "matchers". Ideally I would send an array of dictionaries, with each dictionary containing a regexp field, and a function to be called if that regex was matched, to the remote module.

sosiouxme: I would just have it pass back what matched which regex and let the caller figure out how to respond.

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
juanvallejo: done, linked https://access.redhat.com/solutions/291638

sosiouxme:
@juanvallejo is this still [WIP]? looks pretty done-ish. got some lint/flake8 stuff to fix though.

juanvallejo:
@sosiouxme thanks, removed WIP tag. Will address lint stuff

juanvallejo:
aos-ci-test

rhcarvalho:
If the requirements get fancier, we may use the official `python-systemd` module to work with the journal:
https://github.com/systemd/python-systemd

roles/openshift_health_checker/test/search_journalctl_test.py
```diff
@@ -0,0 +1,199 @@
...
+    """Create a Curator check object with canned get_log_input method"""
+    module = search_journalctl  # fails if a module is actually invoked
+    if get_log_input:
+        module._get_log_input = get_log_input
+    return module
+
+
+def get_timestamp_ms(modifier=0, saveto=None):
+    t = str(time() + modifier).split(".")[0] + "000000"
+    if type(saveto) is list:
```
rhcarvalho: Not very Pythonic to be comparing types, but anyway this the preferred way of doing it:

```python
if isinstance(saveto, list):
```

See https://stackoverflow.com/questions/1549801/differences-between-isinstance-and-type-in-python

roles/openshift_health_checker/test/search_journalctl_test.py
```diff
@@ -0,0 +1,199 @@
+import pytest
+import search_journalctl
+import sys
+
+from time import time
+
+
+SAVED_TIMESTAMPS = list()
```
rhcarvalho: What's the intention behind this?!

roles/openshift_health_checker/test/search_journalctl_test.py
```diff
@@ -0,0 +1,199 @@
...
+
+if sys.version_info > (3, 0):
+    RE_ERROR_UNEXPECTED_END_REGEXP = "unterminated character set"
+else:
+    RE_ERROR_UNEXPECTED_END_REGEXP = "unexpected end of regular expression"
```
rhcarvalho: Hmm, looks catchy -- why would we be checking error message text based depending on Python version?! Hmm, better not.

roles/openshift_health_checker/test/search_journalctl_test.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def canned_search_journalctl(get_log_input=None):
+    """Create a Curator check object with canned get_log_input method"""
```
rhcarvalho: Curator check object?! What?

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,59 @@
...
+    name = "etcd_traffic"
+    tags = ["health", "etcd"]
+
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip hosts that do not have etcd or masters in their group names."""
+        master_count = get_var(task_vars, "openshift", "master", "master_count", default=1)
+        group_names = get_var(task_vars, "group_names", default=[])
+        active = "masters" in group_names
+        if int(master_count) > 1:
```
rhcarvalho: Is this possibly a non-numeric value?

We're probably missing some error handling here.

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,59 @@
...
+            active = "etcd" in group_names
```
rhcarvalho: In terms of code flow, it is strange that this completely overrides the previous value.

The logical expression seems to be:

```python
return super(EtcdTraffic, cls).is_active(task_vars) and (master_could > 1 and 'etcd' in group_names or 'masters' in group_names)
```

Surely complicated-looking... better to give names:

```python
return parent_is_active and (is_etcd_on_multimaster or is_single_master)
```


Now I'm unsure what the actual rule should be... could you please explain in English?

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,59 @@
...
+
+        return super(EtcdTraffic, cls).is_active(task_vars) and active
+
+    def run(self, tmp, task_vars):
+        master_count = get_var(task_vars, "openshift", "master", "master_count", default=1)
+        deploy_type = get_var(task_vars, "openshift", "common", "service_type")
+        unit = self._get_systemd_unit(master_count, deploy_type)
+
+        log_matchers = [{
+            "start_regexp": r"Starting Etcd Server",
```
rhcarvalho: What is the `start_regexp` for?

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,59 @@
...
+            "regexp": r"etcd: sync duration of \d+\.\d+s, expected less than 1s",
```
rhcarvalho: The variable part here should be probably more flexible:
The `.` is probably optional, and since this is coming from a Go duration value... it may end with `s`, but also `m` for minutes, etc, right?
https://golang.org/pkg/time/#Duration.String

Suggestion:

```python
r'etcd: sync duration of [^,]+, expected less than 1s'
```

Matches anything that is not a comma, then the comma and the rest of the error message.

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,59 @@
...
+        }]
+
+        match = self.module_executor("search_journalctl", {
+            "log_matchers": log_matchers,
+        }, task_vars)
+
+        if match.get("matched"):
+            msg = ("Higher than normal etcd traffic detected. "
+                   "OpenShift 3.4 introduced an increase in etcd traffic by at least a factor of 4."
+                   "\nUpgrading to OpenShift 3.6 is recommended in order to fix this issue."
```
rhcarvalho: What about 3.5, is it also affected?

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,151 @@
...
+
+from time import time
+import json
+import re
+import subprocess
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def _exit_json(module, errors, matched_expressions, msg):
```
juanvallejo: Yeah, tried doing this to methods that did not really need an explanation

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,151 @@
...
+        failed=len(errors),
+        errors=errors,
+        matched=matched_expressions,
+        msg=msg,
+    )
+
+
+def _stamp_too_old(stamp, time_limit):
+    epoch = int(stamp) / 1000000
+    return epoch - (time() - time_limit) < 0
```
juanvallejo: Yeah, loosely based this on its oadm diagnostics counterpart https://github.com/openshift/origin/blob/master/pkg/diagnostics/systemd/analyze_logs.go#L142

roles/openshift_health_checker/test/search_journalctl_test.py
juanvallejo: I can go ahead and remove that string match altogether, however, at the time I was thinking it might be worth detecting a regexp exception as part of the tests

roles/openshift_health_checker/test/search_journalctl_test.py
juanvallejo: I needed a way to retain a timestamp created by calling `time.time()` in order to [create a log entry](https://github.com/openshift/openshift-ansible/pull/4316/files#diff-a515f88fdca679286abb67d0bc994c98R188) that matched [an expected log entry set](https://github.com/openshift/openshift-ansible/pull/4316/files#diff-a515f88fdca679286abb67d0bc994c98R191).

roles/openshift_health_checker/test/search_journalctl_test.py
juanvallejo: Thanks for catching this, will update, re-used some test stuff from the efk check

roles/openshift_health_checker/test/search_journalctl_test.py
```diff
@@ -0,0 +1,199 @@
...
+    module = search_journalctl  # fails if a module is actually invoked
+    if get_log_input:
+        module._get_log_input = get_log_input
+    return module
+
+
+def get_timestamp_ms(modifier=0, saveto=None):
```
juanvallejo: I could address the `saveto` parameter by just creating a global "log entry" at the start of the test that I re-use as part of "expected" sets. Did you also have concerns about the modifier? I use that to [create fake older log entries](https://github.com/openshift/openshift-ansible/pull/4316/files#diff-a515f88fdca679286abb67d0bc994c98R169)

roles/openshift_health_checker/test/search_journalctl_test.py
juanvallejo: thanks, will update

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
juanvallejo: We want this to only be active if `masters` is in `group_names` on a single master setup. If there is more than one master, we only want this to be active if `etcd` is in `group_names`. Reason for this is because on multimaster clusters, etcd tends to be on a host of its own if I recall correctly. @sosiouxme

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
juanvallejo: I loosely based most of the logic in this check on Luke's [implementation in oadm diagnostics](https://github.com/openshift/origin/blob/master/pkg/diagnostics/systemd/systemd.go#L64), which also uses a field to mark the beginning of the log set that we want to parse through. This is just to limit the search to last unit startup.

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
juanvallejo: cc @brenton 
Yes, I believe 3.6 is the target release for this bugfix https://bugzilla.redhat.com/show_bug.cgi?id=1415839

roles/openshift_health_checker/test/search_journalctl_test.py
```diff
@@ -0,0 +1,199 @@
...
+    return '{{"__REALTIME_TIMESTAMP": "{}", "MESSAGE": "{}"}}'.format(stamp, msg)
+
+
+@pytest.mark.parametrize('stamp,time_limit,is_too_old', [
+    ("1496341364241717", 60 * 60, True),
+    (time() * 1000000, 60 * 60, False),
+    (time() * 500000, 10, True),
+])
+def test_stamp_too_old(stamp, time_limit, is_too_old):
+    assert search_journalctl._stamp_too_old(stamp, time_limit) == is_too_old
```
juanvallejo: Thanks, will update

roles/openshift_health_checker/test/search_journalctl_test.py
rhcarvalho: This looks like a side-effect of a harder to test API.

Make the function pure (it should not call `time.time()`, but take two timestamps as input instead) and it gets easier to test.

roles/openshift_health_checker/test/search_journalctl_test.py
rhcarvalho: You could look at the type of the exception instead of the specific textual description.

roles/openshift_health_checker/test/search_journalctl_test.py
rhcarvalho: Pure functions and get rid of state, one apple a day and you'll live a healthier life :-)

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
rhcarvalho: Hmm, "tends to be" -- I think we have a gap to cover here ;-)

- What if there is a single master and a dedicated etcd host?
- What if it is multi-master but with no dedicated etcd host?

juanvallejo:
@rhcarvalho thanks, review comments addressed. Although I am not sure about https://github.com/openshift/openshift-ansible/pull/4316#discussion_r121506296 I am hoping @sosiouxme can give some feedback on this

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
rhcarvalho: Summary of what we learned during our call today / IRC follow-up:

- If there are multiple masters, there should be a host in the `etcd` group. This check should run in the host in the `etcd` group.
- If there is a single master, there might still be a dedicated host in the `etcd` group.

To simplify the logic, we thought of running the check in any node that is either in the `masters` group or `etcd` group (or both).

rhcarvalho: Ah, and run this check only in OpenShift 3.4.x and 3.5.x if there's a good way to detect that...

juanvallejo: Thanks, updated `is_active` with these suggestions

juanvallejo:
aos-ci-test

roles/openshift_health_checker/test/etcd_traffic_test.py
```diff
@@ -0,0 +1,87 @@
+import pytest
+
+from openshift_checks.etcd_traffic import EtcdTraffic, OpenShiftCheckException
+
+
+@pytest.mark.parametrize('group_names,is_containerized,version,is_active', [
+    # ensure module is active if "masters" is in group_names when master_count == 1
+    # or "etcd" is in group_names when master_count > 1
```
rhcarvalho: Outdated comment?

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,163 @@
...
+import re
+import subprocess
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+class InvalidMatcherRegexpException(Exception):
+    """Base exception class for invalid matcher regexp fields"""
+    def __init__(self, message):
+        Exception.__init__(self, message)
```
rhcarvalho: These 2 lines should not be needed.

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def stamp_too_old(stamp_microseconds, stamp_limit_seconds):
+    """Receives a log message time stamp (in microseconds) and determines if it
```
rhcarvalho: For all method/function docstrings in this PR: please change the verb tense to imperative. And please add punctuation where missing.

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,163 @@
...
+def main():
+    """Scans a given list of matchers for journalctl messages matching given patterns"""
+    module = AnsibleModule(
+        argument_spec=dict(
+            log_count_limit=dict(type="int", default=500),
+            log_matchers=dict(type="list", required=True),
+        ),
+    )
+
+    timestamp_limit_seconds = time() - 60 * 60  # 1 hour
```
rhcarvalho: Suggestion: use a single time unit and avoid conversions.

We pass this `timestamp_limit_seconds` around through several method calls, and then in `stamp_too_old` we receive time arguments in two different units and need to convert before comparing...

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,163 @@
...
+        regexp = re.compile(matcher["regexp"])
+        start_regexp = re.compile(matcher["start_regexp"])
+    except re.error as err:
+        msg = "A log matcher object was provided with an invalid regular expression: {}"
+        raise InvalidMatcherRegexpException(msg.format(str(err)))
+
+    log_count = 0
+    matched = None
+
+    for line in log_input:
```
rhcarvalho:
```python
for log_count, line = enumerate(log_input):
    ...
```

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,163 @@
...
+            errors.append(str(err))
+            continue
+
+        log_input, error = _get_log_input(matcher)
+        if error:
+            errors.append(error)
+            continue
+
+        try:
+            matched, error = find_matches(log_input, matcher, log_count_limit, timestamp_limit_seconds)
```
rhcarvalho: Shouldn't we look at `matched` only after checking the `error`?

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,163 @@
...
+    log_count = 0
+    matched = None
+
+    for line in log_input:
+        log_count += 1
+        if log_count >= log_count_limit:
+            break
+
+        try:
+            obj = json.loads(line.rstrip())
```
rhcarvalho: Why `rstrip()`?

If that's because of a newline character at the end of the line, removing it is unnecessary:

```
In [4]: import json

In [5]: json.loads('{"line": 54}\n')
Out[5]: {u'line': 54}

In [6]: json.loads('{"line": 54}\n\n')
Out[6]: {u'line': 54}

In [7]: json.loads('{"line": 54}\n   \n')
Out[7]: {u'line': 54}
```

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,163 @@
...
+            '/bin/journalctl',
+            '-ru', matcher["unit"],
+            '--output', 'json',
+        ]), stdout=subprocess.PIPE)
+
+        return iter(cmd_output.stdout.readline, ''), None
+
+    except subprocess.CalledProcessError as exc:
+        err = '[rc {}] {}\n{}'.format(exc.returncode, ' '.join(exc.cmd), exc.output)
+        return None, str(err)
```
rhcarvalho: `err` is already an `str`?!

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,163 @@
...
+        cmd_output = subprocess.Popen(list([
+            '/bin/journalctl',
+            '-ru', matcher["unit"],
+            '--output', 'json',
+        ]), stdout=subprocess.PIPE)
+
+        return iter(cmd_output.stdout.readline, ''), None
+
+    except subprocess.CalledProcessError as exc:
+        err = '[rc {}] {}\n{}'.format(exc.returncode, ' '.join(exc.cmd), exc.output)
```
rhcarvalho: Do we ever expose this error? We may want to make the message more friendly.

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,163 @@
...
+        return None, str(err)
+
+    except OSError as exc:
+        return None, str(exc)
+
+
+class InvalidMatcherException(Exception):
+    """Base exception class for missing or invalid matcher fields"""
+    def __init__(self, message, missing_fields=None):
+        Exception.__init__(self, message)
```
rhcarvalho: `super`?

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
sosiouxme: Late to the party, but yes - I think rather than try to figure out exactly where etcd is running, which is looking more complicated all the time, just use the shotgun approach and look at both the master and etcd journals on any master or etcd host. It's only slightly inefficient and it's much simpler.

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,72 @@
...
+    name = "etcd_traffic"
+    tags = ["health", "etcd"]
+
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip hosts that do not have etcd or masters in their group names."""
+        group_names = get_var(task_vars, "group_names", default=[])
+        valid_group_names = "masters" in group_names or "etcd" in group_names
+
+        version = get_var(task_vars, "openshift", "common", "version", default="0.0.0")
```
sosiouxme: If I'm not mistaken, `openshift.common.short_version` should be defined at this point and not need parsing for comparison...

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,72 @@
...
+        parsed_version = float(".".join(version.split(".")[0:2]))
+        valid_version = parsed_version == 3.4 or parsed_version == 3.5
```
sosiouxme: Thinking of versions as floats is just generally a dangerous thing to do.
In this case, we can just look at them as strings. For comparison, breaking out to integers would be the way to go (fortunately we don't put letters or unicode in our version strings).

roles/openshift_health_checker/library/search_journalctl.py
juanvallejo: yes, this error is propagated up to `_get_log_matches`, and returned in the overall list of errors. Will improve message

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,72 @@
...
+        return {}
+
+    @staticmethod
+    def _get_master_count(task_vars):
+        master_count = get_var(task_vars, "openshift", "master", "master_count", default=1)
+        try:
+            master_count_int = int(master_count)
+        except ValueError:
+            msg = ("Non-numeric value ({}) found for 'openshift.master.master_count' in ansible_facts."
+                   "Make sure an integer value for this variable has been set in your inventory.")
```
juanvallejo: Was not sure how to word this message. Maybe just leave the suggestion out and simply state that `master_count` was a non-numeric value?

juanvallejo:

sosiouxme:
This should go away if we dispense with the logic for figuring out exactly
where etcd is running.

But if that weren't the case... error messages for "this should not ever
happen" can be less useful than usual, but they ought to say what's wrong
in a way that doesn't leave the user wondering if it's their fault.

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,71 @@
...
+        return super(EtcdTraffic, cls).is_active(task_vars) and valid_group_names and valid_version
+
+    def run(self, tmp, task_vars):
+        master_count_int = self._get_master_count(task_vars)
+        deploy_type = get_var(task_vars, "openshift", "common", "service_type")
+        unit = self._get_systemd_unit(master_count_int, deploy_type)
+
+        log_matchers = [{
+            "start_regexp": r"Starting Etcd Server",
+            "regexp": r"etcd: sync duration of [^,], expected less than 1s",
```
rhcarvalho: Missing a `+` to match more characters that are not a comma. Should be `[^,]+`.

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
rhcarvalho: Suggestion: put the "\n" in the previous line.

roles/openshift_health_checker/test/etcd_traffic_test.py
```diff
@@ -0,0 +1,85 @@
...
+    (['masters'], True, "3.6", False),
+    (['nodes'], False, "3.6", False),
+    (['nodes'], True, "3.4", False),
+    (['etcd'], False, "3.4", True),
+    (['etcd'], True, "3.5", True),
+    (['etcd'], False, "3.5", True),
+    (['masters', 'nodes'], False, "3.5", True),
+    (['masters', 'etcd'], False, "3.3", False),
+    ([], False, "3.4", False),
+    (['lb'], False, "3.6", False),
```
rhcarvalho: Looks like we have more combinations here than needed to cover the logic.
E.g. the test case in this line produces is_active=False because the version is wrong and because the group_names doesn't contain the names we're interested in. We may be able to have less lines with the same effect.

roles/openshift_health_checker/test/etcd_traffic_test.py
```diff
@@ -0,0 +1,85 @@
...
+    )
+    assert EtcdTraffic.is_active(task_vars=task_vars) == is_active
+
+
+@pytest.mark.parametrize('group_names,matched,failed,master_count,extra_words', [
+    (["masters"], True, False, 1, ["Higher than normal etcd traffic detected"]),
+    (["masters", "etcd"], False, True, 1, ["No log messages matched the expression"]),
+    (["etcd"], False, False, 5, []),
+])
+def test_log_matches_high_traffic_msg(group_names, matched, failed, master_count, extra_words):
```
rhcarvalho: This test makes it hard to change code, because it looks for very specific error messages. Any updates to the message would require changes to two files.

roles/openshift_health_checker/test/etcd_traffic_test.py
```diff
@@ -0,0 +1,85 @@
...
+    def module_executor(module_name, args, task_vars):
+        return {
+            "matched": matched,
+            "failed": failed,
+            "msg": "No log messages matched the expression",
```
rhcarvalho: This makes little sense, notice how this is the same string as in the test case...

roles/openshift_health_checker/test/etcd_traffic_test.py
```diff
@@ -0,0 +1,85 @@
...
+        openshift=dict(
+            common=dict(service_type="origin"),
+            master=dict(master_count="non-numeric"),
+        )
+    )
+
+    check = EtcdTraffic(execute_module=fake_execute_module)
+    with pytest.raises(OpenShiftCheckException) as err:
+        check._get_master_count(task_vars)
+        assert "Make sure an integer value for this variable has been set" in str(err.value)
```
rhcarvalho: Don't check for specific messages, if makes code harder to change.

roles/openshift_health_checker/test/search_journalctl_test.py
```diff
@@ -0,0 +1,190 @@
...
+def canned_search_journalctl(get_log_input=None):
+    """Create a search_journalctl object with canned get_log_input method"""
+    module = search_journalctl
+    if get_log_input:
+        module._get_log_input = get_log_input
+    return module
+
+
+def get_timestamp_microseconds(modifier=0):
+    return str(time() + modifier).split(".")[0] + "000000"
```
rhcarvalho: I dislike using this in tests because every time we run the tests we are using different test data.
This is an open door for a flake.

roles/openshift_health_checker/test/search_journalctl_test.py
```diff
@@ -0,0 +1,190 @@
...
+    return '{{"__REALTIME_TIMESTAMP": "{}", "MESSAGE": "{}"}}'.format(stamp, msg)
+
+
+@pytest.mark.parametrize('stamp_seconds,stamp_limit_seconds,is_too_old', [
+    (1496341364.241717, time() - 60 * 60, True),
+    (float(time()), time() - 60 * 60, False),
+    (float(time() * 0.5), time() - 10, True),
+])
+def test_stamp_too_old(stamp_seconds, stamp_limit_seconds, is_too_old):
+    assert search_journalctl.stamp_too_old(stamp_seconds, stamp_limit_seconds) == is_too_old
```
rhcarvalho: This function is really trivial, all it does is compare two numbers... If we are bothering to have it defined and testing it, then we can test it with simple constant input, doesn't need calls to `time()`. We're essentially testing the `<` operator...

```python
import operator

stamp_too_old = operator.lt
```

:smile:

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,71 @@
...
+    tags = ["health", "etcd"]
+
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip hosts that do not have etcd or masters in their group names."""
+        group_names = get_var(task_vars, "group_names", default=[])
+        valid_group_names = "masters" in group_names or "etcd" in group_names
+
+        version = get_var(task_vars, "openshift", "common", "short_version")
+        valid_version = version == "3.4" or version == "3.5"
```
sosiouxme: That is its raison d'être so I would say yes.

Although we should probably include 1.4 and 1.5 for Origin versions. There seems to be some translator code in the current master, what does it show up as in actual usage?

roles/openshift_health_checker/test/etcd_traffic_test.py
sosiouxme: That's a problem we have in general. Of course the tests will catch it if we miss something... probably.

Do you have a suggestion? Generally I would solve this by either introducing an unchanging unique ID in each message (which is more annoying than you might think, see current oadm diagnostics implementation), or turning each message into a record with attrs instead of just a string, or just structuring code and tests in such a way that the test can determine easily which error occurred before it gets translated into something meant for humans.

juanvallejo:
@sosiouxme 
> This should go away if we dispense with the logic for figuring out exactly
> where etcd is running.

> But if that weren't the case... error messages for "this should not ever
> happen" can be less useful than usual, but they ought to say what's wrong
> in a way that doesn't leave the user wondering if it's their fault.

Per today's discussion, I have updated the check to only run if `etcd` exists in `group_names`.
The systemd unit is now also `etcd`, unless it is running on a containerized deployment, in which case it is `atomic-openshift-master`.

https://github.com/openshift/openshift-ansible/pull/4316/commits/1566e456c87e411eb5423fddd5b45316c6521254

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,55 @@
...
+        if match["failed"]:
+            return {"failed": True}
+
+        return {}
+
+
+    @staticmethod
+    def _get_systemd_unit(task_vars):
+        is_containerized = get_var(task_vars, "openshift", "common", "is_containerized")
+        if is_containerized:
```
sosiouxme: No need for 4 lines when 1 will do? 

`return "etcd_container" if is_containerized else "etcd"`

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,55 @@
...
+
+class EtcdTraffic(OpenShiftCheck):
+    """Check if host is being affected by an increase in etcd traffic."""
+
+    name = "etcd_traffic"
+    tags = ["health", "etcd"]
+
+    @classmethod
+    def is_active(cls, task_vars):
+        """Skip hosts that do not have etcd or masters in their group names."""
```
sosiouxme: just etcd

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,55 @@
+"""Module that scans journalctl for messages caused as a symptom of increased etcd traffic"""
```
sosiouxme: nit: Technically this is a check not a module...

roles/openshift_health_checker/test/etcd_traffic_test.py
rhcarvalho: > Do you have a suggestion?

Yes. If we really need to test for a specific failure, create an Exception subclass and expect it to be raised (as opposed to comparing its string representation).

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,155 @@
...
+from time import time
+import json
+import re
+import subprocess
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+class InvalidMatcherRegexpException(Exception):
+    """Base exception class for invalid matcher regexp fields"""
```
rhcarvalho: I guess s/fields//?!

Missing punctuation.

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,155 @@
...
+    pass
+
+
+class InvalidMatcherException(Exception):
+    """Base exception class for missing or invalid matcher fields."""
```
rhcarvalho: None of these two need to be called "Base exception class", as they have no subclasses?

Suggestion: s/Base exception class/Exception

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,155 @@
...
+    errors = []
+
+    for matcher in matchers:
+        try:
+            ensure_matcher_fields(matcher)
+        except InvalidMatcherException as err:
+            errors.append(str(err))
+            continue
+
+        log_input, error = get_log_input(matcher)
```
rhcarvalho: This seems inconsistent, why do we use exceptions in `ensure_matcher_fields` but use a return value for errors in `get_log_input`?!

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,155 @@
...
+        return None, err.format(matcher["unit"], str(exc))
+
+    except OSError as exc:
+        return None, str(exc)
+
+
+def find_matches(log_input, matcher, log_count_limit, timestamp_limit_seconds):
+    """Receive iterable input, a matcher dictionary and scan journalctl at the
+    given systemd module for log messages matching a given regular expression.
+    Return the matched log message (or None) and an error string if any."""
```
rhcarvalho: It would be more natural if it just raise exceptions. Currently it communicates errors both by raising `InvalidMatcherRegexpException` AND return a string out of a `ValueError` it may catch...

roles/openshift_health_checker/test/etcd_traffic_test.py
sosiouxme: Makes sense.

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,155 @@
...
+
+from time import time
+import json
+import re
+import subprocess
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+class InvalidMatcherRegexpException(Exception):
```
sosiouxme: Is there a need to have `Exception` in the name? One of @rhcarvalho 's style videos mentioned naming exceptions better. Couldn't it just be `InvalidMatcherRegexp`? And below, `InvalidMatcher`?

roles/openshift_health_checker/library/search_journalctl.py
juanvallejo: Thanks, updated this to just raise exceptions

roles/openshift_health_checker/library/search_journalctl.py
juanvallejo: Thanks, will update these

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,51 @@
...
+    def run(self, tmp, task_vars):
+        unit = self._get_systemd_unit(task_vars)
+
+        log_matchers = [{
+            "start_regexp": r"Starting Etcd Server",
+            "regexp": r"etcd: sync duration of [^,]+, expected less than 1s",
+            "unit": unit
+        }]
+
+        match = self.module_executor("search_journalctl", {
```
sosiouxme: Please, call it `execute_module` - I want to get rid of this deprecated name.

juanvallejo: Will do

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,163 @@
...
+    """Exception class for errors that occur while executing a subprocess"""
+    pass
+
+
+def main():
+    """Scan a given list of matchers for journalctl messages containing given patterns."""
+    module = AnsibleModule(
+        argument_spec=dict(
+            log_count_limit=dict(type="int", default=500),
+            log_matchers=dict(type="list", required=True),
```
rhcarvalho: Perhaps we should document better what is understood to be a "matcher"?
Reading through the code I think it is a dict with 3 keys 'start_regexp', 'regexp' and 'unit'?

I've always wondered if we can represent more complex structures in the `argument_spec`...

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,163 @@
...
+                matched_regexp.append(matcher["regexp"])
+        except InvalidMatcherRegexp as err:
+            errors.append(str(err))
+        except InvalidLogEntry as err:
+            errors.append(str(err))
+
+    return matched_regexp, errors
+
+
+def ensure_matcher_fields(matcher):
```
rhcarvalho: Why do we need this? Couldn't we consider any missing key to behave as if an empty string was given or something equivalent?

roles/openshift_health_checker/library/search_journalctl.py
rhcarvalho: Why do we take a list of matchers if we're only interested in using it for a single regexp/matcher?
Since this module is self-contained in the role, and not very generic anyway, I'd suggest making it as simple as it could possibly be.

rhcarvalho: If we simplify the interface with the module, we could get rid of some code... e.g.

```python
argument_spec=dict(
    log_count_limit=dict(type='int', default=500),
    start_regexp=dict(type='string', default=''),
    ...
)
```

juanvallejo: @rhcarvalho 
> Why do we take a list of matchers if we're only interested in using it for a single regexp/matcher?
> Since this module is self-contained in the role, and not very generic anyway, I'd suggest making it as simple as it could possibly be.

I think the implementation would be a bit awkward if it's just hardcoded to handle one matcher (I was trying to partially base this on @sosiouxme 's implementation in oadm diagnostics). I would rather like to make this more generic instead (maybe it could be used by future checks as well).

juanvallejo: Thanks, will add a comment explaining "matchers".

> I've always wondered if we can represent more complex structures in the argument_spec...

As far as I'm aware, we could require an arg, or limit [its possible values](http://ansible-docs.readthedocs.io/zh/stable-2.0/rst/developing_modules.html#common-module-boilerplate), but I don't think we can have the arg spec enforce "required" matcher fields for us

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,158 @@
...
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+class InvalidMatcherRegexp(Exception):
+    """Exception class for invalid matcher regexp."""
+    pass
+
+
+class InvalidMatcher(Exception):
```
rhcarvalho: Dead code, not used ;)

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,158 @@
...
+        start_regexp = re.compile(matcher.get("start_regexp", ""))
+    except re.error as err:
+        msg = "A log matcher object was provided with an invalid regular expression: {}"
+        raise InvalidMatcherRegexp(msg.format(str(err)))
+
+    matched = None
+
+    for log_count, line in enumerate(log_input):
+        if log_count >= log_count_limit:
+            break
```
rhcarvalho: No need to change anything, but since I shared the existence of `enumerate` before, here goes a bit more of Python-fu:

```python
import itertools

# ...

for line in itertools.islice(log_input, log_count):
    try:
        # ...
```
https://docs.python.org/2/library/itertools.html#recipes, based on the idea of `take(n, iterable)`.

juanvallejo:
@rhcarvalho @sosiouxme thanks for the feedback. Wondering if this check is good to go?

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,51 @@
...
+            "unit": unit
+        }]
+
+        match = self.execute_module("search_journalctl", {
+            "log_matchers": log_matchers,
+        }, task_vars)
+
+        if match.get("matched"):
+            msg = ("Higher than normal etcd traffic detected. "
+                   "OpenShift 3.4 introduced an increase in etcd traffic by at least a factor of 4.\n"
```
rhcarvalho: Nit: we can probably omit the "by at least a factor of 4"? Feels like it contributes to a "fear factor" in the message.

juanvallejo: thanks, fixed

juanvallejo:
@rhcarvalho @sosiouxme thanks, comments addressed

sosiouxme:
> @sosiouxme this conflicts with the changes to task_vars, we will need to decide what goes first.

@rhcarvalho I think we should try to get real changes to the checks (like this) merged first. It's not hard for me to rebase the refactor and fix conflicts/tests.

sosiouxme:
aos-ci-test

juanvallejo:
ping @sosiouxme

roles/openshift_health_checker/test/search_journalctl_test.py
```diff
@@ -0,0 +1,160 @@
...
+            },
+        ],
+        [
+            create_test_log_object(get_timestamp_microseconds(), "test log message"),
+            create_test_log_object(get_timestamp_microseconds(), "Sample Logs Beginning"),
+        ],
+        ["test log message"],
+        [],
+    ),
+    # test with invalid json in log input
```
rhcarvalho: Make the comment part of the test data and the test case ID, so that it becomes part of the output on failure.

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,51 @@
...
+                   "Please refer to https://access.redhat.com/solutions/2916381 for more information.")
+            return {"failed": True, "msg": msg}
+
+        if match["failed"]:
+            return {"failed": True, "msg": "\n".join(match.get("errors"))}
+
+        return {}
+
+    @staticmethod
+    def _get_systemd_unit(task_vars):
```
rhcarvalho: Do we really need this (static) method? If so, does it need an underscore prefix?
Why not inline this in `run`?

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,151 @@
...
+    pass
+
+
+class InvalidLogEntry(Exception):
+    """Exception class for invalid / non-json log entries."""
+    pass
+
+
+class LogInputSubprocessError(Exception):
+    """Exception class for errors that occur while executing a subprocess"""
```
rhcarvalho: For consistency sake, please end the docstrings with a period. This one is different than the similar classes above.
The module docstring in this file and the one below are also missing punctuation.

roles/openshift_health_checker/test/search_journalctl_test.py
juanvallejo: Will update

roles/openshift_health_checker/test/search_journalctl_test.py
```diff
@@ -0,0 +1,160 @@
...
+    def get_log_input(matcher):
+        return log_input
+
+    module = canned_search_journalctl(get_log_input)
+    matched_regexp, errors = module.get_log_matches(matchers, 500, 60 * 60)
+
+    assert set(matched_regexp) == set(expected_matches)
+    for partial_err_msg in expected_errors:
+        for actual_err_msg in errors:
+            assert partial_err_msg in actual_err_msg
```
juanvallejo: Thanks, addressed this. Errors are not duplicate, but in the case of this test, I was receiving a single one each time, so the test was not failing. Addressed this to handle cases where more than one error might be returned by `get_log_matches`

juanvallejo:

roles/openshift_health_checker/test/search_journalctl_test.py
```diff
@@ -0,0 +1,160 @@
...
+                "regexp": r"test log message",
+                "unit": "test",
+            },
+        ],
+        [
+            create_test_log_object(get_timestamp_microseconds(), "test log message"),
+            create_test_log_object(get_timestamp_microseconds(), "Sample Logs Beginning"),
+        ],
+        ["test log message"],
+        [],
```
juanvallejo: Done

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,151 @@
...
+    timestamp_limit_seconds = time() - 60 * 60  # 1 hour
+
+    log_count_limit = module.params["log_count_limit"]
+    log_matchers = module.params["log_matchers"]
+
+    matched_regexp, errors = get_log_matches(log_matchers, log_count_limit, timestamp_limit_seconds)
+
+    module.exit_json(
+        changed=False,
+        failed=len(errors),
```
rhcarvalho: It will be less confusing for us if `failed` is a boolean. Please make it `bool(errors)` (note you don't need `len` to get what you mean, an empty sequence is false, otherwise true).

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,151 @@
...
+        errors=errors,
+        matched=matched_regexp,
+    )
+
+
+def get_log_matches(matchers, log_count_limit, timestamp_limit_seconds):
+    """Iterate through a list of matchers, verify required fields exist on each one, and
+    retrieve log output for each matcher, appending its match pattern to a collection if
+    a matcher's regex pattern is found.
+    Return a collection of matched regular expressions, or raise a MatchErrorException if
```
rhcarvalho: No `raise` statements in this function body, is this comment true / up-to-date?

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,151 @@
...
+            errors.append(str(err))
+        except InvalidLogEntry as err:
+            errors.append(str(err))
+
+    return matched_regexp, errors
+
+
+def get_log_input(matcher):
+    """Run journalctl with the systemd unit specified in the matcher.
+    Return the command output as an iterator."""
```
rhcarvalho: General suggestion for docstrings: speak about *what* a function does, and *why* somebody would want to call it. Often you can omit *how* it does it, leave it as an implementation detail unless the algorithm has particular importance.
That way, comments will be less prone to getting out of sync when the implementation changes, because the *how* tends to change more predominantly than *why* or *what*.

E.g.:

```python
def get_log_output(matcher):
    """Return an iterator on the logs of a given matcher."""
```

BTW, why is the function called "log_input"? Shouldn't it be "output"?!

roles/openshift_health_checker/library/search_journalctl.py
```diff
@@ -0,0 +1,151 @@
...
+        matched=matched_regexp,
+    )
+
+
+def get_log_matches(matchers, log_count_limit, timestamp_limit_seconds):
+    """Iterate through a list of matchers, verify required fields exist on each one, and
+    retrieve log output for each matcher, appending its match pattern to a collection if
+    a matcher's regex pattern is found.
+    Return a collection of matched regular expressions, or raise a MatchErrorException if
+    any errors occur."""
```
rhcarvalho: (please consider this a continuation to the comment below about docstrings)

Another example of docstring:

```python
def get_log_matches(matchers, log_count_limit, timestamp_limit_seconds):
    """Return a list of up to log_count_limit matches for each matcher.

    Log entries are only considered if older than timestamp_limit_seconds.
    """
```

The current doscstring seems to be spelling out a procedure, enumerating the algorithm that reflects the current implementation. It however does not explain what `log_count_limit` and `timestamp_limit_seconds` are meant for (and maybe my example above is documenting them wrong).

roles/openshift_health_checker/library/search_journalctl.py
rhcarvalho: This docstring is outdated. Please review the docstrings. Some are really verbose and still miss fundamental points like explaining non-obvious arguments. Better to aim for concise strings that are still helpful in understanding the purpose of the functions/methods.

roles/openshift_health_checker/library/search_journalctl.py
juanvallejo: Will update docstring

juanvallejo:
@rhcarvalho thanks, comments addressed

juanvallejo:
aos-ci-test

rhcarvalho:
[merge]

rhcarvalho:
Flake https://github.com/openshift/origin/issues/13977

[merge]

rhcarvalho:
Not merging until the queue is open again.

roles/openshift_health_checker/openshift_checks/etcd_traffic.py
```diff
@@ -0,0 +1,47 @@
...
+        }, task_vars)
+
+        if match.get("matched"):
+            msg = ("Higher than normal etcd traffic detected.\n"
+                   "OpenShift 3.4 introduced an increase in etcd traffic.\n"
+                   "Upgrading to OpenShift 3.6 is recommended in order to fix this issue.\n"
+                   "Please refer to https://access.redhat.com/solutions/2916381 for more information.")
+            return {"failed": True, "msg": msg}
+
+        if match["failed"]:
```
juanvallejo: Done

rhcarvalho:
Code was changed since last test run, need another run.

rhcarvalho:
aos-ci-test

rhcarvalho:
[merge]

PR #4317, opened 2017-05-31
Changed the Openstack playbooks to use the native os_stack module

ggillies:
This changes the openstack playbooks to use the native ansible os_stack
module for creating and managing the heat stack for Openshift, rather than
manually invoking the heat command line client. This is because

* The heat command line client itself is depreciated (in favor of the
  unified python-openstackclient

* The output of the command seems to have changed which breaks the
  current method of scraping the output to obtain the output values

* We get some greater flexibility in using the ansible modules as
  they natively hook onto os-cloud-config for cloud credentials

This means that you need to have python-shade installed (as the ansible
openstack modules depend on it) along with the normal Openstack clients.

### DIFF
--- a/playbooks/openstack/openshift-cluster/launch.yml
+++ b/playbooks/openstack/openshift-cluster/launch.yml
@@ -7,96 +7,33 @@
   vars_files:
   - vars.yml
   tasks:
-  # TODO: Write an Ansible module for dealing with HEAT stacks
-  #       Dealing with the outputs is currently terrible
-
-  - name: Check OpenStack stack
-    command: 'heat stack-show openshift-ansible-{{ cluster_id }}-stack'
-    register: stack_show_result
-    changed_when: false
-    failed_when: stack_show_result.rc != 0 and 'Stack not found' not in stack_show_result.stderr
-
-  - set_fact:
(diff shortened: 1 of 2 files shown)


sdodson:
@lhuard1A mind taking a look at this?

bogdando:
@tomassedovic PTAL. Let's think of integration of this work and the https://github.com/openshift/openshift-ansible-contrib/pull/397

lhuard1A:
It’s indeed much cleaner with the dedicated Ansible module!

tomassedovic:
The `playbooks/openstack` directory was removed from openshift-ansible:

https://github.com/openshift/openshift-ansible/pull/5201

They have been moved under openshift-ansible-contrib:

https://github.com/openshift/openshift-ansible-contrib/tree/master/playbooks/openstack

But to my knowledge, no one is actually interested in maintaining them, so a better bet would be to work with the ones here which are actively developed (and may move into openshift-ansible when they are more mature):

https://github.com/openshift/openshift-ansible-contrib/tree/master/playbooks/provisioning/openstack

ggillies:
Closing as it seems this was done elsewhere

PR #4321, opened 2017-05-31
verify upgrade targets separately for each group (masters, nodes, etcd)

ingvagabund:
So we don't check docker and rpm version on nodes if we upgrade the control plane only.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/pre/verify_docker_upgrade_targets.yml
+++ b/playbooks/common/openshift-cluster/upgrades/pre/verify_docker_upgrade_targets.yml
@@ -1,23 +1,20 @@
 ---
-- name: Verify docker upgrade targets
-  hosts: oo_masters_to_config:oo_nodes_to_upgrade:oo_etcd_to_config
-  tasks:
-  # Only check if docker upgrade is required if docker_upgrade is not
-  # already set to False.
-  - include: ../docker/upgrade_check.yml
-    when: docker_upgrade is not defined or docker_upgrade | bool and not openshift.common.is_atomic | bool
+# Only check if docker upgrade is required if docker_upgrade is not
+# already set to False.
+- include: ../docker/upgrade_check.yml
+  when: docker_upgrade is not defined or docker_upgrade | bool and not openshift.common.is_atomic | bool
 
-  # Additional checks for Atomic hosts:
+# Additional checks for Atomic hosts:
 
-  - name: Determine available Docker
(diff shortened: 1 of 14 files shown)


ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/pre/verify_docker_upgrade_targets.yml
```diff
@@ -1,23 +1,20 @@
...
+  when: docker_upgrade is not defined or docker_upgrade | bool and not openshift.common.is_atomic | bool
 
-  # Additional checks for Atomic hosts:
+# Additional checks for Atomic hosts:
 
-  - name: Determine available Docker
-    shell: "rpm -q --queryformat '---\ncurr_version: %{VERSION}\navail_version: \n' docker"
-    register: g_atomic_docker_version_result
-    when: openshift.common.is_atomic | bool
+- name: Determine available Docker
```
ashcrow: Reminder: This will fail on container-engine instances in the future. However, the same method used to gain docker version on atomic will work.

playbooks/common/openshift-cluster/upgrades/pre/verify_upgrade_targets.yml
```diff
@@ -1,45 +1,41 @@
...
-    when: openshift.common.version is not defined
+- name: Verify containers are available for upgrade
+  command: >
+    docker pull {{ openshift.common.cli_image }}:{{ openshift_image_tag }}
+  register: pull_result
+  changed_when: "'Downloaded newer image' in pull_result.stdout"
+  when: openshift.common.is_containerized | bool
 
-  - name: Verify containers are available for upgrade
+- when: not openshift.common.is_containerized | bool
```
ashcrow: Not sure, but I don't think this would work on atomic either.

playbooks/common/openshift-cluster/upgrades/pre/verify_docker_upgrade_targets.yml
ingvagabund: I have not modified the file, just remove the play part and kept all tasks. I can polish the task files if needed. However, we should fix this as a part of another PR, e.g. "towars atomic compatibility".

ingvagabund: Anyway, I would rather replace the shell with repoquery module and extend the module to be runnable on atomic as well.

ingvagabund:
[merge][severity: blocker]

PR #4336, opened 2017-05-31
Add mtu setting to docker

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1457062

This is apparently an issue introduced in 3.4 and newer because the SDN no longer reconfigures docker. The issue only affects containers that aren't pods so DockerStrategy and such

### DIFF
--- a/roles/docker/tasks/package_docker.yml
+++ b/roles/docker/tasks/package_docker.yml
@@ -102,6 +102,21 @@
   notify:
   - restart docker
 
+- stat: path=/etc/sysconfig/docker-network
+  register: sysconfig_docker_network_check
+
+- name: Configure Docker Network OPTIONS
+  lineinfile:
+    dest: /etc/sysconfig/docker-network
+    regexp: '^DOCKER_NETWORK_OPTIONS=.*$'
+    line: "DOCKER_NETWORK_OPTIONS='\
+      {% if openshift.node is defined and openshift.node.sdn_mtu is defined %} --mtu={{ openshift.node.sdn_mtu }}{% endif %}'"
+  when:
+  - sysconfig_docker_network_check.stat.isreg is defined
+  - sysconfig_docker_network_check.stat.isreg
+  notify:
+  - restart docker
+
 - name: Start the Docker service
   systemd:
     name: docker
--- a/roles/openshift_docker_facts/tasks/main.yml
+++ b/roles/openshift_docker_facts/tasks/main.yml
@@ -17,6 +17,9 @@
       hosted_registry_insecure: "{{ openshift_docker_hosted_registry_insecure | default(openshift.docker.hosted_registry_insecure | default(False)) }}"
       hosted_registry_network: "{{ openshift_docker_hosted_registry_network | default(None) }}"
       use_system_container: "{{ openshift_docker_use_system_container | default(False) }}"
+  - role: node
+    local_facts:
+      sdn_mtu: "{{ openshift_node_sdn_mtu | default(None) }}"
 
 - set_fact:
     docker_additional_registries: "{{ openshift.docker.additional_registries


sdodson:
@danwinship PTAL -- this ensures that docker --mtu is set to the same value that's used for networkConfig.mtu in the node config

roles/docker/tasks/package_docker.yml
```diff
@@ -97,7 +97,8 @@
       {% if docker_log_driver is defined  %} --log-driver {{ docker_log_driver }}{% endif %}\
       {% if docker_log_options is defined %} {{ docker_log_options |  oo_split() | oo_prepend_strings_in_list('--log-opt ') | join(' ')}}{% endif %}\
       {% if docker_options is defined %} {{ docker_options }}{% endif %}\
-      {% if docker_disable_push_dockerhub is defined %} --confirm-def-push={{ docker_disable_push_dockerhub | bool }}{% endif %}'"
+      {% if docker_disable_push_dockerhub is defined %} --confirm-def-push={{ docker_disable_push_dockerhub | bool }}{% endif %}\
+      {% if openshift.node is defined and openshift.node.sdn_mtu is defined %} --mtu={{ openshift.node.sdn_mtu }}{% endif %}'"
```
danwinship: This adds the option to OPTIONS in /etc/sysconfig/docker, but in theory network-specific options are supposed to go into DOCKER_NETWORK_OPTIONS in /etc/sysconfig/docker-network. (I don't know why really, but, eg, openshift-sdn originally didn't do it that way, and someone filed a bug about it, https://github.com/openshift/openshift-sdn/issues/52.)

sdodson: Thanks, moved it to DOCKER_NETWORK_OPTIONS in /etc/sysconfig/docker-network

sdodson:
@giuseppe Will the daemon.json change work as expected?

giuseppe:
@sdodson yes that should work without problems in the Docker container as well.  Adding also @ashcrow to the discussion

roles/docker/tasks/package_docker.yml
```diff
@@ -102,6 +102,19 @@
...
 
+- stat: path=/etc/sysconfig/docker-network
+  register: sysconfig_docker_network_check
+
+- name: Configure Docker Network OPTIONS
+  lineinfile:
+    dest: /etc/sysconfig/docker-network
+    regexp: '^DOCKER_NETWORK_OPTIONS=.*$'
+    line: "DOCKER_NETWORK_OPTIONS='\
+      {% if openshift.node is defined and openshift.node.sdn_mtu is defined %} --mtu={{ openshift.node.sdn_mtu }}{% endif %}'"
```
danwinship: I was going to complain that this blows away any pre-existing DOCKER_NETWORK_OPTIONS but I guess that matches the behavior for OPTIONS, so ok

sdodson: Would I be clobbering anything that openshift-sdn puts there today?

roles/docker/tasks/package_docker.yml
```diff
@@ -102,6 +102,19 @@
...
+  when: sysconfig_docker_network_check.stat.isreg is defined and sysconfig_docker_network_check.stat.isreg
```
ashcrow: nit: Multiline when! 😄

roles/docker/templates/daemon.json
```diff
@@ -14,6 +14,9 @@
 	    "path": "/usr/libexec/docker/docker-runc-current"
 	}
     },
+{% if openshift.node is defined and openshift.node.sdn_mtu is defined %}
+    "mtu": {{ openshift.node.sdn_mtu }},
```
ashcrow: This option can't live in both ```/etc/sysconfig/docker-network``` and ```/etc/docker/daemon.json``` as it will cause docker to fail to start up (even if the options are exactly the same.

Note that container-engine honors ```/etc/sysconfig/docker-network``` via it's service file:

```
[Service]
EnvironmentFile=-/etc/sysconfig/docker-storage
EnvironmentFile=-/etc/sysconfig/docker-network
```

roles/docker/tasks/package_docker.yml
danwinship: No, openshift-sdn doesn't modify any docker configuration any more

sdodson:
@ashcrow ok, got rid of that.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

ashcrow:
@sdodson I think the bot is ignoring you! 😆

sdodson:
[merge]

PR #4318, opened 2017-05-31
Metrics: disable become for htpasswd and keytool check

madddi:
local_actions are also executed with sudo if using Ansible "become". This creates permission problems and requires passwordless sudo on the deployment host.

These commits disable become for all local_actions in the openshift_metrics role.

PR #4343, opened 2017-06-01
Logging: disable become on local_actions

madddi:
Like in #4318, sudo on local_actions breaks things when working with an unprivileged user on a seperate deployment host. This disables "become" on local_action tasks.

### DIFF
--- a/roles/openshift_logging/tasks/generate_jks.yaml
+++ b/roles/openshift_logging/tasks/generate_jks.yaml
@@ -22,21 +22,25 @@
 
 - name: Create placeholder for previously created JKS certs to prevent recreating...
   local_action: file path="{{local_tmp.stdout}}/elasticsearch.jks" state=touch mode="u=rw,g=r,o=r"
+  become: false
   when: elasticsearch_jks.stat.exists
   changed_when: False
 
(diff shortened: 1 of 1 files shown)


sdodson:
bot, test pull request

sdodson:
Before we merge this we should audit the entire codebase and get all the `local_action: become: ` problems fixed once and for all.

madddi:
@sdodson Good idea. Some more issues should be fixed in #4318.

As I'm working with unprivileged users I'm bound to encounter all the local_action + become pitfalls. I'll try to fix them and open PRs.

stevekuznetsov:
[test_logging]

FWIW the user we run in CI has passwordless sudo but is not privileged and whatever code paths we hit, no issues today.

madddi:
@stevekuznetsov My user doesn't have passwordless sudo on my local machine, thus triggering the password prompt and failing the local_action-Ansible tasks. The target machines use another user with passwordless sudo.

stevekuznetsov:
> thus triggering the password prompt and failing the local_action-Ansible tasks

I see, so you're failing on local actions that explicitly escalate. These changes pass the logging set-up in the test above, so they seem reasonable to me.

michaelgugino:
Closing due to inactivity.

Issue #4320, opened 2017-05-31
Support resource requests in Logging role


madddi:
#### Description

I'm able to set resource requests in the metrics deployment role, but not in the logging deployment.

As far as I can see, nobody is working on this yet. If this is the case, I'm going to submit a pull request soon which will add support for resource requests.

ewolinetz:
Have you seen the following?
```yaml
openshift_logging_curator_cpu_limit: 100m
openshift_logging_curator_memory_limit: null
openshift_logging_kibana_cpu_limit: null
openshift_logging_kibana_memory_limit: 736Mi
openshift_logging_kibana_proxy_cpu_limit: null
openshift_logging_kibana_proxy_memory_limit: 96Mi
openshift_logging_fluentd_cpu_limit: 100m
openshift_logging_fluentd_memory_limit: 512Mi
openshift_logging_es_cpu_limit: null
openshift_logging_es_memory_limit: "{{ openshift_hosted_logging_elasticsearch_instance_ram | default('8Gi') }}"

openshift_logging_curator_ops_cpu_limit: 100m
openshift_logging_curator_ops_memory_limit: null
openshift_logging_kibana_ops_cpu_limit: null
openshift_logging_kibana_ops_memory_limit: 736Mi
openshift_logging_kibana_ops_proxy_cpu_limit: null
openshift_logging_kibana_ops_proxy_memory_limit: 96Mi
openshift_logging_es_ops_cpu_limit: null
openshift_logging_es_ops_memory_limit: "{{ openshift_hosted_logging_elasticsearch_ops_instance_ram | default('8Gi') }}"
```

madddi:
@ewolinetz Yes, these options set resource limits, not requests.

PR #4344, opened 2017-06-01
Logging: Add support for resource requests

madddi:
This adds support for resource request configuration in openshift_logging  (#4320).

openshift_logging_elasticsearch seems to be in a state of migration / refactoring so the current and new variables don't match up with the parameters in openshift_logging. Is there anything I can or should do?


roles/openshift_logging/tasks/install_logging.yaml
```diff
@@ -176,11 +176,15 @@
...
     openshift_logging_kibana_memory_limit: "{{ openshift_logging_kibana_ops_memory_limit }}"
+    openshift_logging_kibana_cpu_request: "{{ openshift_logging_kibana_ops_cpu_request }}"
+    openshift_logging_kibana_memory_request: "{{ openshift_logging_kibana_ops_memory_request }}"
     openshift_logging_kibana_hostname: "{{ openshift_logging_kibana_ops_hostname }}"
     openshift_logging_kibana_replicas: "{{ openshift_logging_kibana_ops_replica_count }}"
     openshift_logging_kibana_proxy_debug: "{{ openshift_logging_kibana_ops_proxy_debug }}"
     openshift_logging_kibana_proxy_cpu_limit: "{{ openshift_logging_kibana_ops_proxy_cpu_limit }}"
     openshift_logging_kibana_proxy_memory_limit: "{{ openshift_logging_kibana_ops_proxy_memory_limit }}"
+    openshift_logging_kibana_proxy_cpu_request: "{{ openshift_logging_kibana_ops_proxy_cpu_request }}"
+    openshift_logging_kibana_proxy_memory_request: "{{ openshift_logging_kibana_ops_proxy_memory_request }}"
```
ewolinetz: we're also missing this for the ES and Fluentd `include_role`s

ewolinetz:
bot, test pull request

ewolinetz:
[test_logging]

roles/openshift_logging/tasks/install_logging.yaml
madddi: Yes, as I mentioned in my comment above. The variables in openshift_logging_elasticsearch have been renamed and removed from the include_role tasks. I didn't if someone is currently working on this. I could add the necessary parameters again, if you want me to.

ewolinetz:
[test_logging]

sdodson:
[test_logging]

roles/openshift_logging_curator/templates/curator.j2
```diff
@@ -45,6 +45,11 @@ spec:
 {% if curator_memory_limit is defined and curator_memory_limit is not none and curator_memory_limit != "" %}
               memory: "{{curator_memory_limit}}"
 {% endif %}
+            requests:
+              cpu: "{{curator_cpu_request}}"
+{% if curator_memory_request is defined and curator_memory_request is not none and curator_memory_request != "" %}
+              memory: "{{curator_memory_request}}"
+{% endif %}
```
richm: Why is only `memory` set conditionally?  Why is `cpu` always set but not `memory`?

roles/openshift_logging_kibana/templates/kibana.j2
```diff
@@ -47,6 +47,13 @@ spec:
 {% if kibana_memory_limit is not none and kibana_memory_limit != "" %}
               memory: "{{ kibana_memory_limit }}"
 {% endif %}
+            requests:
+{% if kibana_cpu_request is not none and kibana_cpu_request != "" %}
+              cpu: "{{ kibana_cpu_request }}"
+{% endif %}
+{% if kibana_memory_request is not none and kibana_memory_request != "" %}
+              memory: "{{ kibana_memory_request }}"
+{% endif %}
```
richm: Is it possible to have `requests` with no values in it?  Will this cause problems?

roles/openshift_logging_mux/templates/mux.j2
```diff
@@ -46,6 +46,13 @@ spec:
 {% if mux_memory_limit is not none %}
             memory: "{{mux_memory_limit}}"
 {% endif %}
+          requests:
+{% if mux_cpu_request is not none %}
+            cpu: "{{mux_cpu_request}}"
+{% endif %}
+{% if mux_memory_request is not none %}
+            memory: "{{mux_memory_request}}"
+{% endif %}
```
richm: Is it possible that the `requests` section can have no values?  Is the `not none` test sufficient to exclude the empty string value `''`?

roles/openshift_logging_elasticsearch/templates/es.j2
```diff
@@ -48,7 +48,10 @@ spec:
               cpu: "{{es_cpu_limit}}"
 {% endif %}
             requests:
-              memory: "512Mi"
+              memory: "{{es_memory_request}}"
+{% if es_cpu_request is defined and es_cpu_request is not none %}
+              cpu: "{{es_cpu_request}}"
+{% endif %}
```
richm: Why is `cpu` conditionally defined but not `memory`?

madddi:
@richm An empty `request` section is not valid, I think. I'm going to fix the checks there, thanks.

I just extended the tests that were already there. I'm not quite sure why `cpu` is conditionally defined and `memory` is not. It would be more consistent to conditionally set both values I think. I could change this behaviour as well if preferred.

richm:
> I just extended the tests that were already there. I'm not quite sure why cpu is conditionally defined and memory is not. It would be more consistent to conditionally set both values I think. I could change this behaviour as well if preferred.

Yes, please.

michaelgugino:
Closing due to inactivity.

PR #4352, opened 2017-06-01
show correct default value in inventory

mmckinst:
The inventory suggest the default setting for `openshift_hosted_manage_registry` is `false` when its actually `true` in the yaml file: https://github.com/openshift/openshift-ansible/blob/76e4105b6e52cd1f510460d204d37f2b60a790b6/roles/openshift_hosted/tasks/main.yml#L13  . Same for the router https://github.com/openshift/openshift-ansible/blob/76e4105b6e52cd1f510460d204d37f2b60a790b6/roles/openshift_hosted/tasks/main.yml#L10

The current wording is kind of confusing IMHO. With the default setting of `openshift_hosted_manage_registry=true` now it would seem to imply its going to disable management of the openshift registry when it doesn't. This PR should clear that up.


sdodson:
docs only change, merging manually

sdodson:
thanks, you're correct, all of our examples should reflect the defaults rather than alternate options

PR #4360, opened 2017-06-02
Skip service restarts within ca redeployment playbook when expired certificates are detected.

abutcher:
The `openshift_certificate_expiry` role generates [a report](https://gist.github.com/abutcher/6d9bdaae8ddfca6efd76b6639228cf75) containing certificates identified in configuration files and we base inclusion of the restart playbooks on certificate health from the report. Etcd certificates are stored within their own `etcd` key but master and node certificates will be under the `ocp_certs` key. Refactoring the report could simplify the logic used in this PR but folks may depend on the current format in their monitoring.

Note that we're not checking all certificates per component (etcd, master, node) but this should be enough to determine if certificates were previously expired as managed by openshift-ansible.

Tested by generating an expired certificate https://gist.github.com/abutcher/b5cfa5451c790185d3a34ca1bc1a820f

https://bugzilla.redhat.com/show_bug.cgi?id=1452367

### DIFF
--- a/playbooks/common/openshift-cluster/redeploy-certificates/ca.yml
+++ b/playbooks/common/openshift-cluster/redeploy-certificates/ca.yml
@@ -6,6 +6,18 @@
       msg: "The current OpenShift version is less than 1.2/3.2 and does not support CA bundles."
     when: not openshift.common.version_gte_3_2_or_1_2 | bool
 
+- name: Check cert expirys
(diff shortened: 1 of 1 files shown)


ashcrow:
@abutcher will CI test this change?

abutcher:
@ashcrow It will not.

ashcrow:
@sdodson ⬆️

PR #4249, opened 2017-05-19
inventory: rename certificates->certificate in router example

ashcrow:
Per change 9397727e433cbd9bfd865fb5ad773c7b6b4590e8, router replaced
certificates with certificate.

References:
- https://bugzilla.redhat.com/show_bug.cgi?id=1452012
- https://bugzilla.redhat.com/show_bug.cgi?id=1452013

PR #4365, opened 2017-06-05
inventory: rename certificates->certificate in router example

sdodson:
Per change 9397727e433cbd9bfd865fb5ad773c7b6b4590e8, router replaced
certificates with certificate.


sdodson:
picks https://github.com/openshift/openshift-ansible/pull/4249

PR #4369, opened 2017-06-05
etcd: fix conf_dir when installing a system container

giuseppe:
Force the /var/lib/etcd prefix when installing as a system
container.

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -117,6 +117,7 @@
                                      | oo_collect('openshift.common.hostname')
                                      | default(none, true) }}"
     openshift_master_hosts: "{{ groups.oo_masters_to_config }}"
+    r_etcd_common_etcd_runtime: "{{ openshift.common.etcd_runtime }}"
     etcd_ca_host: "{{ groups.oo_etcd_to_config.0 }}"
     etcd_cert_subdir: "openshift-master-{{ openshift.common.hostname }}"
     etcd_cert_config_dir: "{{ openshift.common.config_base }}/master"


giuseppe:
aos-ci-test

roles/etcd_common/defaults/main.yml
```diff
@@ -3,7 +3,7 @@
 r_etcd_common_etcd_runtime: "docker"
 
 # etcd server vars
-etcd_conf_dir: "{{ '/etc/etcd' if r_etcd_common_etcd_runtime != 'runc' else '/var/lib/etcd/etcd.etcd/etc'  }}"
+etcd_conf_dir: "{{ '/etc/etcd' if not openshift.common.is_etcd_system_container else '/var/lib/etcd/etcd.etcd/etc'  }}"
```
ingvagabund: That should be identical. What play calls the `etcd_common` role? There is `openshift.common.etcd_runtime` that is set to `run` iff `openshift.common.is_etcd_system_container` is true.

ingvagabund: We should try to make all roles as much variable free as possible. Including removing `openshift` variable as well.

giuseppe: Thanks for the review.  I see, I've dropped this patch and proposed a different fix.

playbooks/common/openshift-master/config.yml
```diff
@@ -117,6 +117,7 @@
                                      | oo_collect('openshift.common.hostname')
                                      | default(none, true) }}"
     openshift_master_hosts: "{{ groups.oo_masters_to_config }}"
+    r_etcd_common_etcd_runtime: "{{ openshift.common.etcd_runtime }}"
```
ingvagabund: Would be better to introduce `r_openshift_master_etcd_runtime` and propagate it to the `etcd_common` role. However, given the dependency chain `openshift_master` -> `openshift_etcd_client_certificates` -> `etcd_client_certificates` -> `etcd_common`, the roles needs some refactoring before we will have super cool handling of variables or indirect dependencies.

giuseppe:
aos-ci-test

giuseppe:
[merge]

stevekuznetsov:
[test]

PR #4371, opened 2017-06-05
Add node unit file on upgrade

ashcrow:
When ``rpm_upgrade`` is used for upgrading the node's unit file will be
overridden. Later the restart task file reloads systemd and starts the
unit.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1455843

### DIFF
--- a/roles/openshift_node_upgrade/tasks/rpm_upgrade.yml
+++ b/roles/openshift_node_upgrade/tasks/rpm_upgrade.yml
@@ -12,3 +12,18 @@
 - name: Ensure python-yaml present for config upgrade
   package: name=PyYAML state=present
   when: not openshift.common.is_atomic | bool
+
+- name: Install Node service file
+  template:
+    dest: "/etc/systemd/system/{{ openshift.common.service_type }}-node.service"
+    src: "{{ openshift.common.service_type }}-node.service.j2"
+  register: l_node_unit
+
+# NOTE: This is needed to make sure we are using the correct set
+#       of systemd unit files. The RPMs lay down defaults but
+#       the install/upgrade may override them in /etc/systemd/system/.
+# NOTE: We don't use the systemd module as some versions of the module
+#       require a service to be part of the call.
+- name: Reload systemd units
+  command: systemctl daemon-reload
+  when: l_node_unit | changed
--- a/roles/openshift_node_upgrade/templates/atomic-openshift-node.service.j2
+++ b/roles/openshift_node_upgrade/templates/atomic-openshift-node.service.j2
@@ -1,0 +1,1 @@
+../../openshift_node/templates/atomic-openshift-node.service.j2
--- a/roles/openshift_node_upgrade/templates/origin-node.service.j2
+++ b/roles/openshift_node_upgrade/templates/origin-node.service.j2
@@ -1,0 +1,1 @@
+../../openshift_node/templates/origin-node.service.j2


ashcrow:
@sdodson note I'm not sure if I'm doing this in the right location but it seemed right based on the bug recommendation and grepping.

ashcrow:
aos-ci-test

ashcrow:
I think this failed due to the github outage. The result that came back from github caused:

```
simplejson.scanner.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

ashcrow:
aos-ci-test

ashcrow:
Nothing failed AFAICT from logs ....
aos-ci-test

roles/openshift_node_upgrade/tasks/rpm_upgrade.yml
```diff
@@ -12,3 +12,8 @@
 - name: Ensure python-yaml present for config upgrade
   package: name=PyYAML state=present
   when: not openshift.common.is_atomic | bool
+
+- name: Install Node service file
+  template:
+    dest: "/etc/systemd/system/{{ openshift.common.service_type }}-node.service"
+    src: "{{ openshift.common.service_type }}-node.service.j2"
```
sdodson: Probably better to put this in roles/openshift_node_upgrade/tasks/systemd_units.yml

I see other diffs between roles/openshift_node/tasks/systemd_units.yml and roles/openshift_node_upgrade/tasks/systemd_units.yml and I wonder if those should be pulled over too? @ingvagabund thoughts?

ashcrow: Will update once @ingvagabund throws in his views.

ashcrow: @sdodson Looking more at ``roles/openshift_node_upgrade/tasks/systemd_units.yml`` it seems like it's included as part of containerized upgrades only. It also includes ``docker`` commands for pulling containerized containers. It's possible we may want to rename the ``roles/openshift_node_upgrade/tasks/systemd_units.yml`` in the future, but I think it may not be the right place based on tasks.

sdodson: Ok, lets leave it where you have it but we need to reload systemd do we not? If so something like
```
- name: Install Node service file
  template:
    dest: "/etc/systemd/system/{{ openshift.common.service_type }}-node.service"
    src: "{{ openshift.common.service_type }}-node.service.j2"
  register: node_unit

- name: Reload systemd units
  command: systemctl daemon-reload
  when: node_unit | changed
```

ashcrow: We do. I have that at https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_node_upgrade/tasks/restart.yml#L8-L14 which ensures that, before restarting, the daemon is reloaded. I'll go ahead and throw it in here as well as it's possible this gets included without using ``restart.yml``.

sdodson: Thanks, I missed that task when i looked earlier.

ashcrow: Updated.

sdodson:
aos-ci-test

roles/openshift_node_upgrade/tasks/rpm_upgrade.yml
ingvagabund: The `openshift_node_upgrade` duplicates a lot of files from `openshift_node`. Would be more beneficial to merge both roles into so no files between them are duplicated. So `openshift_node_upgrade`'s duplicated files must be 1:1 with the ones in `openshift_node`. I did not want to introduce any symlinks in case we put the roles into galaxy. However, given the current situation I am for the symlinks before the roles are refactored.

sdodson:
[merge]

PR #4372, opened 2017-06-05
Add health checks to upgrade playbook

juanvallejo:
begin adding memory and disk-space verification to upgrade process

Begin addressing https://trello.com/c/K93Wzz4u

cc @sosiouxme @brenton @mtnbikenc @rhcarvalho

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/init.yml
+++ b/playbooks/common/openshift-cluster/upgrades/init.yml
@@ -23,3 +23,5 @@
     set_fact:
       os_firewall_use_firewalld: false
     when: "'Active: active' in service_iptables_status.stdout"
+
+- include: ./pre/verify_health_checks.yml
--- a/playbooks/common/openshift-cluster/upgrades/pre/verify_health_checks.yml
+++ b/playbooks/common/openshift-cluster/upgrades/pre/verify_health_checks.yml
@@ -1,0 +1,13 @@
+---
+- name: Verify Host Requirements
+  hosts: oo_all_hosts
+  roles:
+  - openshift_health_checker
+  vars:
+  - r_openshift_health_checker_playbook_context: upgrade
+  post_tasks:
+  - action: openshift_health_check
+    args:
+      checks:
+      - disk_availability
+      - memory_availability
--- a/roles/openshift_health_checker/openshift_checks/disk_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/disk_availability.py
@@ -35,6 +35,16 @@ class DiskAvailability(OpenShiftCheck):
         },
     }
 
+    # recommended disk space for each location under an upgrade context
+    recommended_disk_upgrade_bytes = {
+        '/var': {
+            'masters': 10 * 10**9,
+            'nodes': 5 * 10 ** 9,
+            'etcd': 5 * 10 ** 9,
+        },
+    }
+
+    @classmethod
     def is_active(self):
         """Skip hosts that do not have recommended disk space requirements."""
         group_names = self.get_var("group_names", default=[])
@@ -80,9 +90,34 @@ class DiskAvailability(OpenShiftCheck):
             config_bytes = max(config.get(name, 0) for name in group_names) * 10**9
             recommended_bytes = config_bytes or recommended_bytes
 
+            # if an "upgrade" context is set, update the minimum disk requirement
+            # as this signifies an in-place upgrade - the node might have the
+            # required total disk space, but some of that space may already be
+            # in use by the existing OpenShift deployment.
+            context = self.get_var("r_openshift_health_checker_playbook_context", default="")
+            if context == "upgrade":
+                recommended_upgrade_paths = self.recommended_disk_upgrade_bytes.get(path, {})
+                if recommended_upgrade_paths:
+                    recommended_bytes = config_bytes or max(recommended_upgrade_paths.get(name, 0)
+                                                            for name in group_names)
+
             if free_bytes < recommended_bytes:
                 free_gb = float(free_bytes) / 10**9
                 recommended_gb = float(recommended_bytes) / 10**9
+                msg = (
+                    'Available disk space in "{}" ({:.1f} GB) '
+                    'is below minimum recommended ({:.1f} GB)'
+                ).format(path, free_gb, recommended_gb)
+
+                # warn if check failed under an "upgrade" context
+                # due to limits imposed by the user config
+                if config_bytes and context == "upgrade":
+                    msg += ('\n\nMake sure to account for decreased disk space during an upgrade\n'
+                            'due to an existing OpenShift deployment. Please check the value of\n'
+                            '  openshift_check_min_host_disk_gb={}\n'
+                            'in your Ansible inventory, and lower the recommended disk space availability\n'
+                            'if necessary for this upgrade.').format(config_bytes)
+
                 return {
                     'failed': True,
                     'msg': (
--- a/roles/openshift_health_checker/openshift_checks/disk_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/disk_availability.py
@@ -44,7 +44,6 @@ class DiskAvailability(OpenShiftCheck):
         },
     }
 
-    @classmethod
     def is_active(self):
         """Skip hosts that do not have recommended disk space requirements."""
         group_names = self.get_var("group_names", default=[])
--- a/roles/openshift_health_checker/test/disk_availability_test.py
+++ b/roles/openshift_health_checker/test/disk_availability_test.py
@@ -224,8 +224,8 @@ def test_min_required_space_changes_with_upgrade_context(name, group_names, cont
         ansible_mounts=ansible_mounts,
     )
 
-    check = DiskAvailability(execute_module=fake_execute_module)
-    result = check.run(tmp=None, task_vars=task_vars)
+    check = DiskAvailability(fake_execute_module, task_vars)
+    result = check.run()
 
     assert result.get("failed", False) == failed
     for word in extra_words:
--- a/roles/openshift_health_checker/test/disk_availability_test.py
+++ b/roles/openshift_health_checker/test/disk_availability_test.py
@@ -97,8 +97,9 @@ def test_succeeds_with_recommended_disk_space(group_names, configured_min, ansib
     assert not result.get('failed', False)
 
 
-@pytest.mark.parametrize('group_names,configured_min,ansible_mounts,extra_words', [
+@pytest.mark.parametrize('name,group_names,configured_min,ansible_mounts,extra_words', [
     (
+        'test with no space available',
         ['masters'],
         0,
         [{
@@ -108,6 +109,7 @@ def test_succeeds_with_recommended_disk_space(group_names, configured_min, ansib
         ['0.0 GB'],
     ),
     (
+        'test with a higher configured required value',
         ['masters'],
         100,  # set a higher threshold
         [{
@@ -117,6 +119,7 @@ def test_succeeds_with_recommended_disk_space(group_names, configured_min, ansib
         ['100.0 GB'],
     ),
     (
+        'test with 1GB available, but "0" GB space requirement',
         ['nodes'],
         0,
         [{
@@ -126,6 +129,7 @@ def test_succeeds_with_recommended_disk_space(group_names, configured_min, ansib
         ['1.0 GB'],
     ),
     (
+        'test with no space available, but "0" GB space requirement',
         ['etcd'],
         0,
         [{
@@ -135,16 +139,17 @@ def test_succeeds_with_recommended_disk_space(group_names, configured_min, ansib
         ['0.0 GB'],
     ),
     (
+        'test with enough space for a node, but not for a master',
         ['nodes', 'masters'],
         0,
         [{
             'mount': '/',
-            # enough space for a node, not enough for a master
             'size_available': 15 * 10**9 + 1,
         }],
         ['15.0 GB'],
     ),
     (
+        'test failure with enough space on "/", but not enough on "/var"',
         ['etcd'],
         0,
         [{
@@ -158,8 +163,8 @@ def test_succeeds_with_recommended_disk_space(group_names, configured_min, ansib
         }],
         ['0.0 GB'],
     ),
-])
-def test_fails_with_insufficient_disk_space(group_names, configured_min, ansible_mounts, extra_words):
+], ids=lambda argval: argval[0])
+def test_fails_with_insufficient_disk_space(name, group_names, configured_min, ansible_mounts, extra_words):
     task_vars = dict(
         group_names=group_names,
         openshift_check_min_host_disk_gb=configured_min,
@@ -170,7 +175,61 @@ def test_fails_with_insufficient_disk_space(group_names, configured_min, ansible
 
     assert result['failed']
     for word in 'below recommended'.split() + extra_words:
-        assert word in result['msg']
+        assert word in result.get('msg', '')
+
+
+@pytest.mark.parametrize('name,group_names,context,ansible_mounts,failed,extra_words', [
+    (
+        'test without enough space for master under "upgrade" context',
+        ['nodes', 'masters'],
+        "upgrade",
+        [{
+            'mount': '/',
+            'size_available': 1 * 10**9 + 1,
+            'size_total': 21 * 10**9 + 1,
+        }],
+        True,
+        ["1.0 GB"],
+    ),
+    (
+        'test with enough space for master under "upgrade" context',
+        ['nodes', 'masters'],
+        "upgrade",
+        [{
+            'mount': '/',
+            'size_available': 10 * 10**9 + 1,
+            'size_total': 21 * 10**9 + 1,
+        }],
+        False,
+        [],
+    ),
+    (
+        'test with not enough space for master, and non-upgrade context',
+        ['nodes', 'masters'],
+        "health",
+        [{
+            'mount': '/',
+            # not enough space for a master,
+            # "health" context should not lower requirement
+            'size_available': 20 * 10**9 + 1,
+        }],
+        True,
+        ["20.0 GB", "below minimum"],
+    ),
+], ids=lambda argval: argval[0])
+def test_min_required_space_changes_with_upgrade_context(name, group_names, context, ansible_mounts, failed, extra_words):
+    task_vars = dict(
+        r_openshift_health_checker_playbook_context=context,
+        group_names=group_names,
+        ansible_mounts=ansible_mounts,
+    )
+
+    check = DiskAvailability(execute_module=fake_execute_module)
+    result = check.run(tmp=None, task_vars=task_vars)
+
+    assert result.get("failed", False) == failed
+    for word in extra_words:
+        assert word in result.get('msg', '')
 
 
 def fake_execute_module(*args):


playbooks/common/openshift-cluster/upgrades/init.yml
```diff
@@ -23,3 +23,16 @@
     set_fact:
       os_firewall_use_firewalld: false
     when: "'Active: active' in service_iptables_status.stdout"
+
+- name: Verify Host Requirements
+  hosts: OSEv3
+  roles:
+  - openshift_health_checker
+  vars:
+  - r_openshift_health_checker_playbook_context: "upgrade"
```
rhcarvalho: Presumably, we need to handle this in https://github.com/openshift/openshift-ansible/blob/2d4709b5a65c5bb2777277420ee08f8554efb1ab/roles/openshift_health_checker/callback_plugins/zz_failure_summary.py#L98? /cc @sosiouxme

rhcarvalho:
Does `pre/verify_memory_and_diskspace.yml` get run automatically?

rhcarvalho:
> Does pre/verify_memory_and_diskspace.yml get run automatically?

I bet it doesn't, I grepped for `verify_upgrade_targets` (because there is a `pre/verify_upgrade_targets.yml` file), and found multiple places where it is include, for example here:

https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/upgrades/v3_3/upgrade.yml#L84

roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
```diff
@@ -95,7 +95,7 @@ def _print_check_failure_summary(self, failed_checks, context):
             'Variables can be set in the inventory or passed on the\n'
             'command line using the -e flag to ansible-playbook.\n'
         ).format(playbook=self._playbook_file, checks=checks)
-        if context in ['pre-install', 'health']:
+        if context in ['pre-install', 'pre-upgrade', 'health']:
```
rhcarvalho: @sosiouxme when you designed this, do you remember what were the "contexts" in which the summary message below should not be printed?

playbooks/common/openshift-cluster/upgrades/pre/verify_memory_and_diskspace.yml
```diff
@@ -0,0 +1,12 @@
+- name: Verify Host Requirements
+  hosts: OSEv3
```
mtnbikenc: This should be `oo_all_hosts`.  `OSEv3` should not be called in `common` playbooks.

juanvallejo:
@mtnbikenc @rhcarvalho thanks for the feedback, review comments addressed

roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
sosiouxme: @rhcarvalho @juanvallejo I guess a comment would have helped. So what prompted this was that if you're running a playbook like [byo/config.yml](https://github.com/openshift/openshift-ansible/blob/4f129045b3f3c1b22203a7ab68cda15ee4b8d87b/playbooks/byo/openshift-cluster/config.yml#L11) then these checks running may come as a surprise and it needs a little more explanation. If you're running a playbook that's specifically for running checks, it's not about "failing early" as they're the whole point, and so the default preamble above is a little out of place. I could imagine other cases in which we'd like to customize the message or even the outcome based on the user intent/expectations. But for now, there are really just two cases: you're trying to run checks, or you're trying to get something else done and the checks are "in the way".

So a "pre-upgrade" context would mean you're running a playbook specifically to see if your upgrade is gonna work out. An "upgrade" context would mean you're actually trying to run the upgrade.

playbooks/common/openshift-cluster/upgrades/pre/verify_memory_and_diskspace.yml
```diff
@@ -0,0 +1,12 @@
+- name: Verify Host Requirements
+  hosts: oo_all_hosts
+  roles:
+  - openshift_health_checker
+  vars:
+  - r_openshift_health_checker_playbook_context: "pre-upgrade"
```
sosiouxme: So if this is being included in an actual upgrade playbook above, I would say that should be just "upgrade" context. If you want to reuse the playbook it should probably not specify a context, instead relying on including playbooks to set context.

Also I don't like calling this `verify_memory_and_diskspace.yml` -- I assume we're going to add more checks to the same play in later iterations.

juanvallejo: Thanks, will update

playbooks/common/openshift-cluster/upgrades/pre/verify_health_checks.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: Verify Host Requirements
+  hosts: oo_all_hosts
+  roles:
+  - openshift_health_checker
+  vars:
+  - r_openshift_health_checker_playbook_context: "upgrade"
```
rhcarvalho: We could safely drop the quotes here, just like in all other places in this YAML file.

playbooks/common/openshift-cluster/upgrades/init.yml
```diff
@@ -23,3 +23,5 @@
     set_fact:
       os_firewall_use_firewalld: false
     when: "'Active: active' in service_iptables_status.stdout"
+
+- include: ./pre/verify_memory_and_diskspace.yml
```
rhcarvalho: This reference is outdated.

playbooks/common/openshift-cluster/upgrades/pre/verify_health_checks.yml
```diff
@@ -0,0 +1,13 @@
...
+  post_tasks:
+  - action: openshift_health_check
+    args:
+      checks:
+      - disk_availability
```
rhcarvalho: The memory requirements in an upgrade should match the requirements of the newer version... but for disk requirements, we probably have at least two scenarios:

**In-place upgrade**

The free space requirement should likely be different than for an install, after all there is already data in the cluster (e.g., Docker images and installed RPM packages).

I couldn't find any specific guidance about free disk space in the [docs](https://docs.openshift.org/latest/install_config/upgrading/index.html).

**Blue-green**

I believe the free space requirement for the new hosts should be the same as for a fresh install.

---

We may need a way to tell when an upgrade is in-place and when it is blue-green.

juanvallejo: Maybe we could propagate a `context` to our checks?

rhcarvalho: What `context` do you have in mind? We do have `r_openshift_health_checker_playbook_context`, but that's manually set from playbooks.

sosiouxme: If you're just configuring how the checks run per playbook, you can just add another variable (like `r_openshift_health_checker_playbook_context`) on the playbook and read that from the check. The only thing that's special about that context var is that it gets recorded in results so that the callback plugin can see it and adjust output.

juanvallejo: @sosiouxme Thanks, although I don't suppose there is a straightforward way of determining the type of upgrade that's taking place? All I have found was https://docs.openshift.org/latest/install_config/upgrading/automated_upgrades.html#install-config-upgrading-automated-upgrades

juanvallejo: @mtnbikenc wondering if you have any thoughts on this as well?

sosiouxme: @juanvallejo Does a different type of upgrade use a different playbook, or just different parameters? Either way you have something to key off of.

juanvallejo: If I understand https://docs.openshift.org/latest/install_config/upgrading/blue_green_deployments.html#blue-green-deployments-preparing-for-upgrade correctly, it looks like in the case of a `blue-green` upgrade, an `in-place` upgrade is still run on existing `masters` and `etcd` hosts, but not on existing `nodes`. Instead, existing `nodes` are labeled with `color=blue` (or whatever value a user decides on). Then, [entirely new nodes are created with the new version of OpenShift using a playbook](https://docs.openshift.org/latest/install_config/upgrading/blue_green_deployments.html#blue-green-creating-labeling-green-nodes). Once green nodes are ready, pods are evacuated from blue nodes to green.

Therefore, I think that if the upgrade `context` is set, and the host is in the `nodes` group, it is safe to assume that it is going through an `in-place` upgrade (since a blue-green upgrade would be creating entirely new nodes instead). `masters` and `etcd` hosts can be treated as `in-place` for both cases.

sosiouxme: Sounds right.
So essentially there are just two cases, install or upgrade.

juanvallejo: Will push updated `disk_availability` check that takes the `upgrade` context into account; halving the required disk space in that case

juanvallejo:
@sosiouxme @rhcarvalho @mtnbikenc wondering your thoughts on https://github.com/openshift/openshift-ansible/pull/4372/commits/7bd04ce8aee698527c2f149f5f6637e7c8358b3b

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -33,6 +33,14 @@ def run(self, tmp, task_vars):
         configured_min = int(get_var(task_vars, "openshift_check_min_host_disk_gb", default=0)) * 10**9
         min_free_bytes = configured_min or recommended_min
 
+        # if an "upgrade" context is set, lower the minimum disk requirement
+        # as this signifies an in-place upgrade - the node might have the
+        # required total disk space, but some of that space may already be
+        # in use by the existing OpenShift deployment.
+        context = get_var(task_vars, "r_openshift_health_checker_playbook_context", default="")
+        if context == "upgrade":
+            min_free_bytes /= 2
```
sosiouxme: Seems... kinda arbitrary. I think in this context, if we run the check at all, it should just be checking that they have enough to run the upgrade. Not sure what that should be, maybe like 1GB? Or, this could be re-purposed as a "health" check when it's not in a fresh install context.

roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
```diff
@@ -95,7 +95,7 @@ def _print_check_failure_summary(self, failed_checks, context):
             'Variables can be set in the inventory or passed on the\n'
             'command line using the -e flag to ansible-playbook.\n'
         ).format(playbook=self._playbook_file, checks=checks)
-        if context in ['pre-install', 'health']:
+        if context in ['pre-install', 'upgrade', 'health']:
```
sosiouxme: This is the summary for playbooks where the user is trying to run checks. I wouldn't include "upgrade" here so they can have the previous msg with a little more explanation. If there were a "pre-upgrade" playbook just for running checks then I would put it here.

roles/openshift_health_checker/openshift_checks/disk_availability.py
juanvallejo: Not able to find a definite requirement in the docs for disk size during an upgrade.

> Or, this could be re-purposed as a "health" check when it's not in a fresh install context.

That could also work. We could fail the check in this context if disk usage is something like > 90%

roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
juanvallejo: Sounds good, will remove

roles/openshift_health_checker/openshift_checks/disk_availability.py
rhcarvalho: Percentage is not good. 90% usage of a 1TB disk means 100GB free -- more than enough for an install/upgrade. 90% usage of 20GB means 2GB free, risky.

juanvallejo: Still unclear on a minimum disk availability value under "upgrade" context. Updated check to require at least `5.0 GB` for in-place upgrades, for now

playbooks/common/openshift-cluster/upgrades/pre/verify_health_checks.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: Verify Host Requirements
+  hosts: oo_all_hosts
```
rhcarvalho: @sosiouxme shall we limit this also to run only on masters and nodes, or is it enough to filter it on the action plugin level?

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -33,6 +33,24 @@ def run(self, tmp, task_vars):
         configured_min = int(get_var(task_vars, "openshift_check_min_host_disk_gb", default=0)) * 10**9
         min_free_bytes = configured_min or recommended_min
 
+        # if an "upgrade" context is set, lower the minimum disk requirement
+        # as this signifies an in-place upgrade - the node might have the
+        # required total disk space, but some of that space may already be
+        # in use by the existing OpenShift deployment.
+        context = get_var(task_vars, "r_openshift_health_checker_playbook_context", default="")
+        if context == "upgrade":
+            upgrade_min_required_diskspace = 5.0 * 10**9
```
juanvallejo: I'm fine with waiting for #4436 to merge, then adding the decreased `upgrade` requirement on top of that

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -33,6 +33,24 @@ def run(self, tmp, task_vars):
...
+            if free_bytes < upgrade_min_required_diskspace:
+                return {
+                    'failed': True,
+                    'msg': (
+                        'Available disk space ({:.1f} GB) for the volume containing '
+                        '"/var" is below minimum recommended space for an upgrade ({:.1f} GB).'
```
juanvallejo: Not sure what you mean. The `return` statement inside this block prevents the text below from being displayed if this one is.

juanvallejo:
@sosiouxme @rhcarvalho rebased with changes introduced to disk_availability.py in #4436 
PTAL

juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -81,6 +100,33 @@ def run(self, tmp, task_vars):
...
+            context = get_var(task_vars, "r_openshift_health_checker_playbook_context", default="")
+            if context == "upgrade":
+                min_upgrade_rec = self.recommended_disk_upgrade_bytes.get(path, {})
+                if min_upgrade_rec:
+                    min_upgrade_rec = max(min_upgrade_rec.get(name, 0) for name in group_names)
+                else:
+                    # if not able to resolve a recommended value, default to non-upgrade recommendation
+                    min_upgrade_rec = max(recommendation.get(name, 0) for name in group_names)
+
+                if free_bytes < min_upgrade_rec:
```
rhcarvalho: I feel like this duplicates logic unnecessarily. Couldn't we simple use the upgrade values when it's an upgrade and then use the same code paths to compare free / required?

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -35,6 +35,25 @@ class DiskAvailability(OpenShiftCheck):
...
+        },
+        '/usr/local/bin': {
+            'masters': 1 * 10**9,
+            'nodes': 1 * 10**9,
+            'etcd': 1 * 10**9,
+        },
+        tempfile.gettempdir(): {
+            'masters': 0.5 * 10**9,
+            'nodes': 0.5 * 10**9,
+            'etcd': 0.5 * 10**9,
```
rhcarvalho: Granted the 1 GB for /usr/local/bin and temp were both originally a rough guess, why do we make this different here?

roles/openshift_health_checker/test/disk_availability_test.py
```diff
@@ -173,7 +173,62 @@ def test_fails_with_insufficient_disk_space(group_names, configured_min, ansible
...
+            'mount': '/',
+            # not enough space for a master,
+            # "health" context should not lower requirement
+            'size_available': 20 * 10**9 + 1,
+        }],
+        True,
+        ["20.0 GB", "below minimum"],
+    ),
+])
+def test_min_required_space_decreases_with_upgrade_context(group_names, context, ansible_mounts, failed, extra_words):
```
rhcarvalho: Here and other places we say "decrease" or otherwise give suggestions as to what the values are compared to "non-upgrade". I'd avoid doing that type of comment or naming, because what governs the relationship of the numbers is actual data. Change the data and all the names become misleading.

Likewise, on an extreme, it is considered bad practice to name variables like `two = 2` only to see `two = 3` some day in the future...

playbooks/common/openshift-cluster/upgrades/pre/verify_health_checks.yml
sosiouxme: @rhcarvalho sorry i missed this. no, it looks like `oo_all_hosts` comes from `g_all_hosts` which is masters/nodes/etcd only so this should be fine. all our playbooks basically need to work the same as `byo/config.yml`

roles/openshift_health_checker/openshift_checks/disk_availability.py
juanvallejo: Will update to just reduce space requirements for `/var`

juanvallejo:

juanvallejo:
@rhcarvalho or @sosiouxme wondering if this patch is good to go. Thanks!

juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -81,6 +90,19 @@ def run(self, tmp, task_vars):
...
+            # required total disk space, but some of that space may already be
+            # in use by the existing OpenShift deployment.
+            context = get_var(task_vars, "r_openshift_health_checker_playbook_context", default="")
+            if context == "upgrade":
+                recommended_bytes = self.recommended_disk_upgrade_bytes.get(path, {})
+                if recommended_bytes:
+                    recommended_bytes = max(recommended_bytes.get(name, 0) for name in group_names)
+                else:
+                    # if not able to resolve a recommended value, default to non-upgrade recommendation
+                    recommended_bytes = max(recommendation.get(name, 0) for name in group_names)
```
rhcarvalho: I think there are some subtle problems here:

- If the user configured the values, this block is overriding it;
- This line "resets" the value of `recommended_bytes` to "almost" what it was in line https://github.com/openshift/openshift-ansible/pull/4372/files#diff-dd45df49348ab2716e213a0ecc5b0f4dR91, though this time we ignore the user-config. That's because when we see the "upgrade" context we're hard-resetting the variable.

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -81,6 +90,19 @@ def run(self, tmp, task_vars):
             config_bytes = max(config.get(name, 0) for name in group_names) * 10**9
             recommended_bytes = config_bytes or recommended_bytes
 
+            # if an "upgrade" context is set, update the minimum disk requirement
+            # as this signifies an in-place upgrade - the node might have the
+            # required total disk space, but some of that space may already be
+            # in use by the existing OpenShift deployment.
+            context = get_var(task_vars, "r_openshift_health_checker_playbook_context", default="")
+            if context == "upgrade":
+                recommended_bytes = self.recommended_disk_upgrade_bytes.get(path, {})
```
rhcarvalho: Could reuse the existing value of `recommended_bytes` instead of resetting it:

```python
recommended_bytes = self.recommended_disk_upgrade_bytes.get(path, recommended_bytes)
```

juanvallejo: Thanks, updated this to only override the value of `recommended bytes` if `self.recommended_disk_upgrade_bytes.get(path, {})` exists

juanvallejo:
@rhcarvalho 
> Still a problem with how we update the requirements when it's an upgrade.

Thanks, this has been addressed in https://github.com/openshift/openshift-ansible/pull/4372/commits/ec3546cfe4ef8e5a79562862a23a31a2a2fdcf6b

Will go ahead and squash if no further comments

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -81,6 +90,16 @@ def run(self, tmp, task_vars):
...
 
+            # if an "upgrade" context is set, update the minimum disk requirement
+            # as this signifies an in-place upgrade - the node might have the
+            # required total disk space, but some of that space may already be
+            # in use by the existing OpenShift deployment.
+            context = get_var(task_vars, "r_openshift_health_checker_playbook_context", default="")
+            if context == "upgrade":
+                recommended_upgrade_paths = self.recommended_disk_upgrade_bytes.get(path, {})
+                if recommended_upgrade_paths:
+                    recommended_bytes = max(recommended_upgrade_paths.get(name, 0) for name in group_names)
```
rhcarvalho: I think this is still wrong because the user configuration should have precedence over everything else?
If you agree, please add a test case for that, see the fail, then fix the implementation and see the test pass. Commit and push :-)

juanvallejo: While I agree user implementation should take precedence, in this case, we are intentionally modifying existing config in order to account for the fact that an upgrade might be dealing with less space / resources due to the existing deployment. I'm not sure if it's still valid to take user config into account, when that config could have been primarily specified for a fresh install

rhcarvalho: There's indeed no obvious answer.
Think how you'd document / explain to a friend each of the possibilities and that might help us make a decision (picking what is clearer, less surprising, easier to explain and to understand).

juanvallejo:
@rhcarvalho thanks for the feedback; updated check to take user config into account, even during an upgrade. Additionally, added a user warning if the upgrade fails due to a user-configured recommended disk space availability

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -81,9 +90,33 @@ def run(self, tmp, task_vars):
...
+                    'Available disk space in "{}" ({:.1f} GB) '
+                    'is below minimum recommended ({:.1f} GB)'
+                ).format(path, free_gb, recommended_gb)
+
+                # warn if check failed under an "upgrade" context
+                # due to limits imposed by the user config
+                if config_bytes and context == "upgrade":
+                    msg += ('\n\nMake sure to account for decreased disk space during an upgrade\n'
+                            'due to an existing OpenShift deployment. Please check the value of\n'
+                            '  r_openshift_health_checker_playbook_context={}\n'
```
rhcarvalho: The message is good, but I think `r_...` variables should not be exposed to users? /cc @sosiouxme

juanvallejo: Thanks, I believe there was no need to use the role variable at all; I had intended to use `openshift_check_min_host_disk_gb` instead, but mistakenly ended up using the role var. This has been addressed

rhcarvalho:
LGTM pending squash. Maybe @mtnbikenc could give a last pass with an  eye on the integration with the upgrade flow?

rhcarvalho:
thanks @mtnbikenc

rhcarvalho:
aos-ci-test

juanvallejo:
@rhcarvalho ok to merge?

rhcarvalho:
@juanvallejo yes, we've got plenty of approval :-) time for the CI dance

rhcarvalho:
aos-ci-test

juanvallejo:
@rhcarvalho ci tests are green, will tag for merge

[merge]

juanvallejo:
flaked on https://github.com/openshift/origin/issues/10162 re[merge]

rhcarvalho:
Flake https://github.com/openshift/origin/issues/15356, perhaps fixed in https://github.com/openshift/origin/pull/15482.

And apparently something new, https://github.com/openshift/origin/issues/15522.

[merge] one last time.

rhcarvalho:
Flake openshift/origin#15522 again, proceeding to manual merge as per https://github.com/openshift/openshift-ansible/blob/master/docs/pull_requests.md#manual-merges

rhcarvalho:
@juanvallejo please review https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/754/ and open a new PR.

rhcarvalho:
I made a mistake and merged when the error was actually related to the changes introduced by this PR. Subsequently reverted the changes.

Issue #4366, opened 2017-06-05
openshift.hosted.registry: Object Storage Provider: "gcs" is not currently supported


pschiffe:
#### Description

OpenShift hosted registry role doesn't fully support GCS (google cloud storage) object storage provider. While the `registry_config.j2` template file supports it [1], the execution of the role won't pass the assertion [2].

[1] https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/templates/registry_config.j2#L56
[2] https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_hosted/tasks/registry/storage/object_storage.yml#L5

When fixing the assertion to:
```
- openshift.hosted.registry.storage.provider in ['azure_blob', 's3', 'swift', 'gcs']
```
the OpenShift hosted registry role works correctly for me.

##### Version

```
$ git describe 
openshift-ansible-3.5.78-1
```

##### Steps To Reproduce
Try to deploy OpenShift 3.5 with Ansible installer on the Google Cloud Platform with hosted registry. Config I'm trying to use:
```
openshift_hosted_registry_selector: 'role=infra'
openshift_hosted_registry_storage_kind: object
openshift_hosted_registry_storage_provider: gcs
openshift_hosted_registry_storage_gcs_bucket: '{{ some_valid_bucket }}'
openshift_hosted_registry_storage_gcs_rootdirectory: /registry
openshift_hosted_registry_pullthrough: true
openshift_hosted_registry_acceptschema2: true
openshift_hosted_registry_enforcequota: true
```

##### Expected Results
Successfully deployed OpenShift on GCP.

##### Observed Results
Failed assertion `Object Storage Provider: "gcs" is not currently supported`

abutcher:
@pschiffe are you able to successfully configure gcs registry storage if you remove the assertion?

pschiffe:
@abutcher yes, I've tried that and it worked.

abutcher:
@pschiffe Great, let's knock down that barrier then.

pschiffe:
@abutcher awesome, thanks! Would it be possible to merge this commit also to the `release-1.5` branch?

ingvagabund:
1.5 backport here https://github.com/openshift/openshift-ansible/pull/4420

PR #4375, opened 2017-06-06
Remove supported/implemented checks for registry object storage provider

abutcher:
Fixes https://github.com/openshift/openshift-ansible/issues/4366

We have variables for the following providers within the registry config template but we only allow `s3`.
* `azure_blob`
* `gcs`
* `s3`
* `swift`

We could just add all of these to supported/implemented (or at least `gcs`) but I'm not sure it makes sense to have this check. If there are issues with a particular provider we can address them.


abutcher:
aos-ci-test

sdodson:
aos-ci-test

abutcher:
aos-ci-test

abutcher:
[merge]

PR #4379, opened 2017-06-06
kibana - only create oauthclient once

richm:
With this fix, I am able to run kibana against an environment setup
with an ops cluster.
@ewolinetz @jcantrill PTAL

### DIFF
--- a/roles/openshift_logging_kibana/tasks/main.yaml
+++ b/roles/openshift_logging_kibana/tasks/main.yaml
@@ -125,6 +125,7 @@
 - name: Generate proxy session
   set_fact:
     session_secret: "{{ 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789' | random_word(200) }}"
+  when: session_secret is not defined
   check_mode: no
 
 # gen oauth_secret -- if necessary
@@ -132,6 +133,7 @@
 - name: Generate oauth client secret
   set_fact:
     oauth_secret: "{{ 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789' | random_word(64) }}"
+  when: oauth_secret is not defined
   check_mode: no
 
 # create oauth client
@@ -141,7 +143,9 @@
     dest: "{{ tempdir }}/templates/oauth-client.yml"
   vars:
     kibana_hostname: "{{ openshift_logging_kibana_hostname }}"
+    kibana_ops_hostname: "{{ openshift_logging_kibana_ops_hostname }}"
     secret: "{{ oauth_secret }}"
+  when: do_oauth_client is defined and do_oauth_client | bool
 
 - name: Set kibana-proxy oauth-client
   oc_obj:
@@ -152,6 +156,7 @@
     files:
     - "{{ tempdir }}/templates/oauth-client.yml"
     delete_after: true
+  when: do_oauth_client is defined and do_oauth_client | bool
 
 # create Kibana secret
 - name: Set Kibana secret
--- a/roles/openshift_logging_kibana/templates/oauth-client.j2
+++ b/roles/openshift_logging_kibana/templates/oauth-client.j2
@@ -7,6 +7,9 @@ metadata:
 secret: {{secret}}
 redirectURIs:
 - https://{{kibana_hostname}}
+{% if openshift_logging_use_ops %}
+- https://{{kibana_ops_hostname}}
+{% endif %}
 scopeRestrictions:
 - literals:
   - user:info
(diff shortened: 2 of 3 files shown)


roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -141,7 +143,9 @@
     dest: "{{ tempdir }}/templates/oauth-client.yml"
   vars:
     kibana_hostname: "{{ openshift_logging_kibana_hostname }}"
+    kibana_ops_hostname: "{{ openshift_logging_kibana_ops_hostname }}"
```
ewolinetz: we shouldn't need to do this, if the openshift_logging_kibana_hostname is incorrect being passed into the role we should adjust that at the `openshift_logging/tasks/install_logging.yml` level. It breaks the pattern being used

roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -125,13 +125,15 @@
 - name: Generate proxy session
   set_fact:
     session_secret: "{{ 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789' | random_word(200) }}"
+  when: session_secret is not defined
```
ewolinetz: I had been struggling with trying to figure out how we should handle this. What if we install a non-ops and then go back and install with ops enabled? we should reuse the prior session secret but the fact may not exist. Similar for the oauth secret.

Should we write these values to a file if they don't exist and then read from them similar to how we do for our certs?

roles/openshift_logging_kibana/templates/oauth-client.j2
```diff
@@ -7,6 +7,9 @@ metadata:
 secret: {{secret}}
 redirectURIs:
 - https://{{kibana_hostname}}
+{% if openshift_logging_use_ops %}
```
ewolinetz: this goes against the pattern as stated above...

roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -141,7 +143,9 @@
     dest: "{{ tempdir }}/templates/oauth-client.yml"
   vars:
     kibana_hostname: "{{ openshift_logging_kibana_hostname }}"
+    kibana_ops_hostname: "{{ openshift_logging_kibana_ops_hostname }}"
     secret: "{{ oauth_secret }}"
+  when: do_oauth_client is defined and do_oauth_client | bool
```
ewolinetz: ideally we would let the oauth creation be idempotent. If we can ensure that we are reusing the same oauth_secret and session_secret we shouldn't need to do this

roles/openshift_logging_kibana/tasks/main.yaml
jcantrill: I think if we wish to support the case you describe we need to treat as certs and write them to a file, do checks to see if said file exists.

roles/openshift_logging_kibana/tasks/main.yaml
jcantrill: This breaks the pattern because..... we iterate over the role multiple times for ops and non-ops scenarios?

jcantrill:
@richm To understand, we need this change because deployment is failing because we try to create oauthclient twice?

roles/openshift_logging_kibana/tasks/main.yaml
richm: @jcantrill 
> This breaks the pattern because..... we iterate over the role multiple times for ops and non-ops scenarios?

Right.  The first iteration creates the kibana-proxy oauthclient with the kibana hostname.  The second iteration replaces the kibana-proxy oauthclient, using the kibana-ops hostname.

richm: Ideally we would have a module for oc_oauthclient which would be able to figure out that the oauthclient already exists, and just add the kibana-ops hostname to it.

Next, I guess we need to fix oc_obj - either it isn't detecting that the kibana-proxy oauthclient already exists, or it thinks it needs to replace the entire thing.

richm:
@jcantrill 
> To understand, we need this change because deployment is failing because we try to create oauthclient twice?

Right.  And the second time it replaces the first time.

We need this fix in order to use kibana and kibana-ops when you deploy an ops cluster i.e. typical CI configuration.

This PR so far is really just a hack to get it to work so that I could use kibana.

It sounds like there is a _lot_ more work needed to do this correctly.  I don't have the cycles right now to do this, so this PR may be in this state for quite some time . . . in the meantime I'll just deploy ansible either without ops, or using this PR.

roles/openshift_logging_kibana/tasks/main.yaml
ewolinetz: I thought when I tested this it just patches in the other hostname, it shouldn't completely replace the current oauthclient...

ewolinetz:
@richm let me take a stab at addressing this as i'm suggesting including verifying that the oauthclient is getting patched vs replaced

roles/openshift_logging_kibana/tasks/main.yaml
ewolinetz: Looks like that is no longer true... it had worked initially, i wonder if the module changed its functionality

ewolinetz:
@richm this should hopefully all be addressed in https://github.com/openshift/openshift-ansible/pull/4387

richm:
closing in favor of https://github.com/openshift/openshift-ansible/pull/4387

PR #4380, opened 2017-06-07
Fixes to oc_obj delete()

jarrpa:


ashcrow:
👁 👁

roles/lib_openshift/library/oc_obj.py
```diff
@@ -1557,7 +1562,8 @@ def run_ansible(params, check_mode=False):
         if state == 'absent':
             # verify its not in our results
             if (params['name'] is not None or params['selector'] is not None) and \
-               (len(api_rval['results']) == 0 or len(api_rval['results'][0].get('items', [])) == 0):
+               (len(api_rval['results']) == 0 or \
```
ashcrow: nit: (Not required to change) Parenthesis are easier on the eyes than backslashes.

```python
if ((params['name'] is not None or params['selector'] is not None) and 
        (len(api_rval['results']) == 0 or ...)
        ...):
    ...
```

ashcrow:
aos-ci-test

roles/lib_openshift/library/oc_obj.py
jarrpa: The backslash was to prevent the line from getting absurdly long, in particular since I was adding yet another parenthesized clause inside a parenthesized clause.

sdodson:
[merge]

sdodson:
[merge]

PR #4384, opened 2017-06-07
Guard check for container install based on openshift dictionary key

admiyo:


sdodson:
bot, test pull request

sdodson:
@abutcher I guess this is necessary because it's run across host group and we may hit another node that hasn't had facts collected?

admiyo:
Yes, I was seeing this when running on a two node cluster.  Without this change, the expresssion was erroring out with an error indicating that the openshift dictionary had not been created.

sdodson:
aos-ci-test

abutcher:
bot, retest this please

abutcher:
@sdodson Right, I can see this being a problem if one of the hosts were unavailable or had received a previous error. We collect facts for all `oo_nodes_to_config` in the play prior to the changes so I would expect facts to be set.

abutcher:
[merge]

PR #4388, opened 2017-06-08
Etcd container to system container

giuseppe:
Migrate existing etcd storage to the system container location.

Check if there is already an installation of etcd before installing the system container.  If there is one, copy the previous data to the system container location.

Also fixed an issue when the etcd systemd service was previously masked.

### DIFF
--- a/roles/etcd/tasks/system_container.yml
+++ b/roles/etcd/tasks/system_container.yml
@@ -15,6 +15,32 @@
       {%- endif -%}
       {% endfor -%}
 
+- name: Check etcd system container package
+  command: >
+    atomic containers list --no-trunc -a -f container=etcd -f backend=ostree
+  register: etcd_result
+
+- name: Unmask etcd service
+  systemd:
+    name: etcd
+    state: stopped
+    enabled: yes
+    masked: no
+    daemon_reload: yes
+  register: task_result
+  failed_when: task_result|failed and 'could not' not in task_result.msg|lower
+  when: "'etcd' in etcd_result.stdout"
+
+- name: Disable etcd_container
+  systemd:
+    name: etcd_container
+    state: stopped
+    enabled: no
+    masked: yes
+    daemon_reload: yes
+  register: task_result
+  failed_when: task_result|failed and 'could not' not in task_result.msg|lower
+
 - name: Check for previous etcd data store
   stat:
     path: "{{ etcd_data_dir }}/member/"
--- a/roles/etcd/tasks/system_container.yml
+++ b/roles/etcd/tasks/system_container.yml
@@ -85,3 +85,5 @@
       - ETCD_PEER_CA_FILE={{ etcd_system_container_conf_dir }}/ca.crt
       - ETCD_PEER_CERT_FILE={{ etcd_system_container_conf_dir }}/peer.crt
       - ETCD_PEER_KEY_FILE={{ etcd_system_container_conf_dir }}/peer.key
+      - ETCD_TRUSTED_CA_FILE={{ etcd_system_container_conf_dir }}/ca.crt
+      - ETCD_PEER_TRUSTED_CA_FILE={{ etcd_system_container_conf_dir }}/ca.crt
--- a/roles/etcd/tasks/system_container.yml
+++ b/roles/etcd/tasks/system_container.yml
@@ -15,6 +15,30 @@
       {%- endif -%}
       {% endfor -%}
 
+- name: Check for previous etcd data store
+  stat:
+    path: "{{ etcd_data_dir }}/member/"
+  register: src_datastore
+
+- name: Check for etcd system container data store
+  stat:
+    path: "{{ r_etcd_common_system_container_host_dir }}/etcd.etcd/member"
+  register: dest_datastore
+
+- name: Ensure that etcd system container data dirs exist
+  file: path="{{ item }}" state=directory
+  with_items:
+    - "{{ r_etcd_common_system_container_host_dir }}/etc"
+    - "{{ r_etcd_common_system_container_host_dir }}/etcd.etcd"
+
+- name: Copy etcd data store
+  command: >
+    cp -a {{ etcd_data_dir }}/member
+    {{ r_etcd_common_system_container_host_dir }}/etcd.etcd/member
+  when:
+    - src_datastore.stat.exists
+    - not dest_datastore.stat.exists
+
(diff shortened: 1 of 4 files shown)


roles/etcd/tasks/system_container.yml
```diff
@@ -15,6 +15,57 @@
...
+
+- name: Unmask etcd service
+  systemd:
+    name: etcd
+    state: stopped
+    enabled: yes
+    masked: no
+    daemon_reload: yes
+  register: task_result
+  failed_when: task_result|failed and 'could not' not in task_result.msg|lower
```
ingvagabund: Does the `failed_when` works with:
```yaml
failed_when:
- task_result|failed
- 'could not' not in task_result.msg|lower
```
as well?

Plus, should not the second condition be `'could not' in task_result.msg|lower`? If the condition is ok, can be add comment why is that so?

roles/etcd/tasks/system_container.yml
```diff
@@ -15,6 +15,57 @@
...
+    path: "{{ etcd_system_container_host_dir }}/etcd.etcd/member"
+  register: dest_datastore
+
+- name: Ensure that etcd system container data dirs exist
+  file: path="{{ item }}" state=directory
+  with_items:
+    - "{{ etcd_system_container_host_dir }}/etc"
+    - "{{ etcd_system_container_host_dir }}/etcd.etcd"
+
+- block:
```
ingvagabund: Block is not needed here.

roles/etcd_common/defaults/main.yml
```diff
@@ -3,7 +3,8 @@
 r_etcd_common_etcd_runtime: "docker"
 
 # etcd server vars
-etcd_conf_dir: "{{ '/etc/etcd' if r_etcd_common_etcd_runtime != 'runc' else '/var/lib/etcd/etcd.etcd/etc'  }}"
+etcd_conf_dir: '/etc/etcd'
+etcd_system_container_host_dir: /var/lib/etcd/etcd.etcd
```
ingvagabund: s/etcd_system_container_host_dir/r_etcd_common_system_container_host_dir/

roles/etcd/tasks/system_container.yml
giuseppe: I copied this chunk from other rules dealing with `systemd` services:

```bash
$ git grep "failed_when: task_result|failed and 'could not' not in task_result.msg|lower"
roles/etcd/tasks/main.yml:    failed_when: task_result|failed and 'could not' not in task_result.msg|lower
roles/openshift_master/tasks/main.yml:  failed_when: task_result|failed and 'could not' not in task_result.msg|lower
roles/os_firewall/tasks/firewall/firewalld.yml:  failed_when: task_result|failed and 'could not' not in task_result.msg|lower
roles/os_firewall/tasks/firewall/iptables.yml:  failed_when: task_result|failed and 'could not' not in task_result.msg|lower
```

Is it fine to leave it as it is?

giuseppe:
@ingvagabund I've pushed a new version with two of your comments addressed.  I've left the systemd part the same, as there are other occurrences in the codebase.  If you still disagree, I'll change it as well :-)

roles/etcd/tasks/system_container.yml
ingvagabund: It is not a blocker so it is ok. Just a nit.

roles/openshift_etcd_facts/vars/main.yml
```diff
@@ -5,6 +5,6 @@ etcd_hostname: "{{ openshift.common.hostname }}"
 etcd_ip: "{{ openshift.common.ip }}"
 etcd_cert_subdir: "etcd-{{ openshift.common.hostname }}"
 etcd_cert_prefix:
-etcd_cert_config_dir: "{{ '/etc/etcd' if not openshift.common.is_etcd_system_container | bool else '/var/lib/etcd/etcd.etcd/etc' }}"
+etcd_cert_config_dir: "/etc/etcd"
```
ingvagabund: Think it would make sense to add `etcd_system_container_cert_config_dir: /var/lib/etcd/etcd.etcd/etc` var here as well and use it instead of `r_etcd_common_system_container_host_dir` where the certs are handled?

roles/etcd/tasks/system_container.yml
```diff
@@ -15,6 +15,56 @@
...
+    - "{{ r_etcd_common_system_container_host_dir }}/etc"
+    - "{{ r_etcd_common_system_container_host_dir }}/etcd.etcd"
+
+- name: Copy etcd data store
+  command: >
+    cp -a {{ etcd_data_dir }}/member
+    {{ r_etcd_common_system_container_host_dir }}/etcd.etcd/member
+  when:
+    - src_datastore.stat.exists
+    - not dest_datastore.stat.exists
```
ingvagabund: What is supposed to happen if the system container data directory already exists? Ignored it? Or maybe print a warning?

giuseppe: IMO, in that case I think it should be silently used without any warning to the user.

giuseppe:
I've discovered a regression and now when installing directly the system container we get an error.  I'll push a new version as soon as I have a fix for it.

giuseppe:
nevermind my last comment, it was an issue on my environment.

@ingvagabund I've pushed a new version that introduces `etcd_system_container_cert_config_dir`

roles/openshift_etcd_facts/vars/main.yml
```diff
@@ -5,6 +5,7 @@ etcd_hostname: "{{ openshift.common.hostname }}"
...
+etcd_system_container_cert_config_dir: "{{ r_etcd_common_system_container_host_dir }}/etc"
```
ingvagabund: That will not work unless etcd_common is a dependency of openshift_etcd_facts. ``etcd_system_container_cert_config_dir: /var/lib/etcd/etcd.etcd/etc`` will do.

giuseppe: thanks, pushed a fix for that

ingvagabund:
`yammlint` issues found, please fix.

giuseppe:
@ingvagabund thanks, I've just fixed them

giuseppe:
aos-ci-test

ingvagabund:
[test]

giuseppe:
[merge]

PR #4045, opened 2017-05-01
Normalizing groups.

kwoodson:

PR #4390, opened 2017-06-08
skip installation of base package in openshift_ca when scalling up master

ingvagabund:
Backport part of https://github.com/openshift/openshift-ansible/pull/4045 and skipping installation of atomic-openshift on the first master during master scaleup (since excluders are disabled on the new master node only).


ingvagabund:
aos-ci-test

PR #4395, opened 2017-06-08
Add play and role to install ansible-service-broker

fabianvf:
Currently blocked waiting for https://trello.com/c/Ebx11vsc

This is currently a work in progress, I'm blocked for a bit waiting on some changes to the broker and rcm approvals. I just wanted to submit this PR now to get a good sanity check and some review so that once my blockers are removed I'll be good to merge.

A few questions I have:

1. I just kind of stuck my play in the `openshift-cluster/config.yml` playbook because it was convenient for my testing, but I wasn't sure how to determine where it actually belongs
1. So far I only need to add a variable that determines whether or not to install the broker. I wasn't sure where I should put this variable. Currently it's global (so I guess should be prefixed with `g_`), but I figured this is probably bad and it would end up nested somewhere else, like `openshift_common` or something. Wasn't sure how to determine where it fits, and assumed this would probably also be dependent on where the play ends up.
1. My etcd requires a persistent volume, the way I've done this is just creating the pvc, but since I don't think there are any pvs available by default I wasn't sure what the correct way to handle that would be. Should I just create a hostpath volume or is there a more preferred way to handle this?

### DIFF
--- a/playbooks/common/openshift-cluster/config.yml
+++ b/playbooks/common/openshift-cluster/config.yml
@@ -54,3 +54,10 @@
   - role: openshift_excluder
     r_openshift_excluder_action: enable
     r_openshift_excluder_service_type: "{{ openshift.common.service_type }}"
+
+- name: install ansible-service-broker
+  hosts: oo_first_master
+  roles:
+  - role: ansible_service_broker
+    when: install_ansible_service_broker | bool
+    ignore_errors: true
--- a/roles/ansible_service_broker/tasks/main.yml
+++ b/roles/ansible_service_broker/tasks/main.yml
@@ -1,0 +1,140 @@
+---
+
+- name: create openshift-ansible-service-broker project
+  oc_project:
+    name: openshift-ansible-service-broker
+
+- name: create ansible-service-broker serviceaccount
+  oc_serviceaccount:
+    name: asb
+    namespace: openshift-ansible-service-broker
+    state: present
+
+- name: create ansible-service-broker service
+  oc_service:
+    name: asb
+    namespace: openshift-ansible-service-broker
+    labels:
+      app: ansible-service-broker
+      service: asb
+    ports:
+      - name: port-1338
+        port: 1338
+    selector:
+      app: ansible-service-broker
+      service: asb
+
+- name: create etcd service
+  oc_service:
+    name: etcd
+    namespace: openshift-ansible-service-broker
+    ports:
+      - name: etcd-advertise
+        port: 2379
+    selector:
+      app: ansible-service-broker
+      service: etcd
+
+- name: create route for ansible-service-broker service
+  oc_route:
+    name: asb-1338
+    namespace: openshift-ansible-service-broker
+    service_name: asb
+    port: 1338
+
+- name: create persistent volume claim for etcd
+  oc_pvc:
+    state: present
+    name: etcd
+    namespace: openshift-ansible-service-broker
+    volume_capacity: 1G
+    access_modes:
+    - ReadWriteOnce
+
+- name: create etcd deployment
+  oc_obj:
+    state: present
+    name: etcd
+    namespace: openshift-ansible-service-broker
+    kind: Deployment
+    content:
+      path: /tmp/dcout
+      data:
+        apiVersion: extensions/v1beta1
+        kind: Deployment
+        metadata:
+          name: etcd
+          namespace: openshift-ansible-service-broker
+          labels:
+            app: ansible-service-broker
+            service: etcd
+        spec:
+          strategy:
+            type: Recreate
+          replicas: 1
+          template:
+            metadata:
+              labels:
+                app: ansible-service-broker
+                service: etcd
+            spec:
+              restartPolicy: Always
+              containers:
+                - image: "ansibleapp/ansible-service-broker-etcd:latest"
+                  name: main
+                  imagePullPolicy: IfNotPresent
+                  workingDir: /etcd
+                  args:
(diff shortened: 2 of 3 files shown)


fabianvf:
[(8) Update OCP ansible installer to install Ansible Service Broker.[asb]](https://trello.com/c/e3RrZNh0/257-8-update-ocp-ansible-installer-to-install-ansible-service-brokerasb)

roles/ansible_service_broker/tasks/main.yml
```diff
@@ -0,0 +1,140 @@
...
+    access_modes:
+    - ReadWriteOnce
+
+- name: create etcd deployment
+  oc_obj:
+    state: present
+    name: etcd
+    namespace: openshift-ansible-service-broker
+    kind: Deployment
+    content:
```
ewolinetz: I didn't realize we could specify the spec with `content:` very cool!

playbooks/common/openshift-cluster/config.yml
```diff
@@ -54,3 +54,10 @@
   - role: openshift_excluder
     r_openshift_excluder_action: enable
     r_openshift_excluder_service_type: "{{ openshift.common.service_type }}"
+
+- name: install ansible-service-broker
+  hosts: oo_first_master
+  roles:
+  - role: ansible_service_broker
+    when: install_ansible_service_broker | bool
+    ignore_errors: true
```
ewolinetz: why are we ignoring errors on this installation?

ewolinetz:
Overall this looks good to me, should there be a way to specify values for things like the PVC size or the namespace we would install this in?

> I just kind of stuck my play in the openshift-cluster/config.yml playbook because it was convenient for my testing, but I wasn't sure how to determine where it actually belongs

If you want to, you could write a playbook for bolting this on afterwards for testing purposes -- `playbooks/byo/openshift-cluster/openshift-logging.yml` should be a decent starting point for that.

> So far I only need to add a variable that determines whether or not to install the broker. I wasn't sure where I should put this variable. Currently it's global (so I guess should be prefixed with g_), but I figured this is probably bad and it would end up nested somewhere else, like openshift_common or something. Wasn't sure how to determine where it fits, and assumed this would probably also be dependent on where the play ends up.

To me, it would make sense to have the prefix of the variable be the role name since it is specific to that role, something like `ansible_service_broker_install`.  I'll have to defer on whether or not it should be also prefixed with `g_`

> My etcd requires a persistent volume, the way I've done this is just creating the pvc, but since I don't think there are any pvs available by default I wasn't sure what the correct way to handle that would be. Should I just create a hostpath volume or is there a more preferred way to handle this?

Are you asking specifically for testing? In `playbooks/common/openshift-cluster/openshift_hosted.yml` we use a role to generate PVs for our hosted components. That might involves also updating facts so that it can know to pick up information for this role. That would configure a NFS backed PV.

playbooks/common/openshift-cluster/config.yml
fabianvf: Oops, that's just because I'm sloppy. Leftover from some debugging.

fabianvf:
>  should there be a way to specify values for things like the PVC size or the namespace we would install this in?

I'm not sure about that, I think in the catalog installation meeting we talked about having a predetermined namespace, but having low-priority default that you can overwrite makes sense to me. Not sure about the volume size, that might be something that becomes more clear after TP. The 1G size should work for all broker workloads, but I could see wanting to make it smaller.

> If you want to, you could write a playbook for bolting this on afterwards for testing purposes -- playbooks/byo/openshift-cluster/openshift-logging.yml should be a decent starting point for that.

That makes perfect sense. Should I put a `- include:` into the config playbook or leave it to the user to run it manually? Not sure what the plan is for the catalog install, but sharing its behavior probably makes sense.

> To me, it would make sense to have the prefix of the variable be the role name since it is specific to that role, something like ansible_service_broker_install. I'll have to defer on whether or not it should be also prefixed with g_

sounds good

> Are you asking specifically for testing? In playbooks/common/openshift-cluster/openshift_hosted.yml we use a role to generate PVs for our hosted components. That might involves also updating facts so that it can know to pick up information for this role. That would configure a NFS backed PV.

No, I'm set for testing, just wasn't sure how to handle it for the real deal. This makes sense, I'll look into it.

Also, is it possible for me to push a commit without triggering all the tests? Don't want to suck up all these cycles fixing typos and cruft when I know it is going to fail.

ewolinetz:
@fabianvf this is the part of the service catalog work right? My understanding from that meeting was that we'll be installing it with the cluster so we'd want that to remain but the separate playbook makes it easy for testing with an already stood up environment. Unless I'm confusing catalog meetings...

Also, how much overlap does this PR have with https://github.com/openshift/openshift-ansible/pull/4195 ?

fabianvf:
> this is the part of the service catalog work right? My understanding from that meeting was that we'll be installing it with the cluster so we'd want that to remain but the separate playbook makes it easy for testing with an already stood up environment. Unless I'm confusing catalog meetings...

This was my understanding as well, was just double checking

> Also, how much overlap does this PR have with #4195 ?

That's a great question... I guess it depends on whether or not we want the ansible service broker to get installed along with the catalog. If so, it would probably make sense for me to rebase my PR onto that branch and add the `ansible_service_broker` role to the catalog install play, and get rid of the play I added to config.yml.

fabianvf:
Closed in favor of merging changes with service-catalog installer: https://github.com/ewolinetz/openshift-ansible/pull/26

PR #4396, opened 2017-06-08
[WIP] Fix master_certs_missings fact idempotence

abutcher:
Correct the list of certificates checked in `openshift_master_certificates` s.t. masters do not incorrectly report that master certs are missing.

* Remove `etcd.server.crt`. `openshift_master_certificates` is responsible for generating certificates for additional masters after the first and there will be no need to test for `etcd.server.crt` in a multi-master environment.
* Remove version checking for `master.proxy-client.crt`.
* Add `ca.crt` and `ca-bundle.crt` to the list to check.

### DIFF
--- a/roles/openshift_master_certificates/tasks/main.yml
+++ b/roles/openshift_master_certificates/tasks/main.yml
@@ -1,25 +1,17 @@
 ---
-- set_fact:
-    openshift_master_certs_no_etcd:
-    - admin.crt
-    - master.kubelet-client.crt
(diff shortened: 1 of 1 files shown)


abutcher:
aos-ci-test

michaelgugino:
Does this still need to ship?

abutcher:
@michaelgugino this is still a problem but I'll need to revisit it for sure. Marking as WIP.

PR #4397, opened 2017-06-08
pre-install checks: add more during byo install

sosiouxme:
Add the docker and RPM checks to the list that run at install time.
They can be disabled the same as the existing ones.

It seems that cockpit-kubernetes RPM is no longer relevant.
Also I noticed that `docker_image_availability` didn't test for the right tag for containerized installs;
I switched to using the `openshift_image_tag` that's set by `openshift_version` for component images, which is how they're installed.
I also added logic to let `oreg_url` set the template for image names of containers that run on top of OpenShift. https://github.com/openshift/openshift-ansible/issues/4415 raises the issue of how to standardize what image names we get as it's rather haphazard now.
I found that `execute_module` in some cases was being called with incorrect positional args; I never saw a problem locally but under CI this could result in an exception down the call stack where a string was expected for tmp dir instead of a dict. So, I fixed that up.

### DIFF
--- a/playbooks/byo/openshift-cluster/config.yml
+++ b/playbooks/byo/openshift-cluster/config.yml
@@ -15,6 +15,11 @@
       checks:
       - disk_availability
       - memory_availability
+      - package_availability
+      - package_update
+      - package_version
+      - docker_image_availability
+      - docker_storage
 
 - include: ../../common/openshift-cluster/std_include.yml
   tags:
--- a/roles/openshift_health_checker/openshift_checks/docker_image_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/docker_image_availability.py
@@ -1,6 +1,22 @@
-# pylint: disable=missing-docstring
+"""Check that required Docker images are available."""
+
 from openshift_checks import OpenShiftCheck, get_var
 from openshift_checks.mixins import DockerHostMixin
+
+
+NODE_IMAGE_SUFFIXES = ["haproxy-router", "docker-registry", "deployer", "pod"]
+DEPLOYMENT_IMAGE_INFO = {
+    "origin": {
+        "namespace": "openshift",
+        "name": "origin",
+        "registry_console_image": "cockpit/kubernetes",
+    },
+    "openshift-enterprise": {
+        "namespace": "openshift3",
+        "name": "ose",
+        "registry_console_image": "registry.access.redhat.com/openshift3/registry-console",
+    },
+}
 
 
 class DockerImageAvailability(DockerHostMixin, OpenShiftCheck):
@@ -13,25 +29,13 @@ class DockerImageAvailability(DockerHostMixin, OpenShiftCheck):
 
     name = "docker_image_availability"
     tags = ["preflight"]
-
     dependencies = ["skopeo", "python-docker-py"]
-
-    deployment_image_info = {
-        "origin": {
-            "namespace": "openshift",
-            "name": "origin",
-        },
-        "openshift-enterprise": {
-            "namespace": "openshift3",
-            "name": "ose",
-        },
-    }
 
     @classmethod
     def is_active(cls, task_vars):
         """Skip hosts with unsupported deployment types."""
         deployment_type = get_var(task_vars, "openshift_deployment_type")
-        has_valid_deployment_type = deployment_type in cls.deployment_image_info
+        has_valid_deployment_type = deployment_type in DEPLOYMENT_IMAGE_INFO
 
         return super(DockerImageAvailability, cls).is_active(task_vars) and has_valid_deployment_type
 
@@ -70,51 +74,55 @@ class DockerImageAvailability(DockerHostMixin, OpenShiftCheck):
 
         return {"changed": changed}
 
-    def required_images(self, task_vars):
+    @staticmethod
+    def required_images(task_vars):
+        """
+        Determine which images we expect to need for this host.
+        Returns: a set of required images like 'openshift/origin:v3.6'
+
+        The thorny issue of determining the image names from the variables is under consideration
+        via https://github.com/openshift/openshift-ansible/issues/4415
+
+        For now we operate as follows:
+        * For containerized components (master, node, ...) we look at the deployment type and
+          use openshift/origin or openshift3/ose as the base for those component images. The
+          version is openshift_image_tag as determined by the openshift_version role.
+        * For OpenShift-managed infrastructure (router, registry...) we use oreg_url if
+          it is defined; otherwise we again use the base that depends on the deployment type.
+        Registry is not included in constructed images. It may be in oreg_url or etcd image.
+        """
+        required = set()
         deployment_type = get_var(task_vars, "openshift_deployment_type")
-        image_info = self.deployment_image_info[deployment_type]
+        host_groups = get_var(task_vars, "group_names")
+        image_tag = get_var(task_vars, "openshift_image_tag")
+        image_info = DEPLOYMENT_IMAGE_INFO[deployment_type]
+        if not image_info:
+            return required
 
-        openshift_release = get_var(task_vars, "openshift_release", default="latest")
-        openshift_image_tag = get_var(task_vars, "openshift_image_tag")
-        is_containerized = get_var(task_vars, "openshift", "common", "is_containerized")
+        # template for images that run on top of OpenShift
+        image_url = "{}/{}-{}:{}".format(image_info["namespace"], image_info["name"], "${component}", "${version}")
+        image_url = get_var(task_vars, "oreg_url", default="") or image_url
+        if 'nodes' in host_groups:
+            for suffix in NODE_IMAGE_SUFFIXES:
+                required.add(image_url.replace("${component}", suffix).replace("${version}", image_tag))
+            # The registry-console is for some reason not prefixed with ose- like the other components.
+            # Nor is it versioned the same, so just look for latest.
+            # Also a completely different name is used for Origin.
+            required.add(image_info["registry_console_image"])
 
-        images = set(self.required_docker_images(
-            image_info["namespace"],
-            image_info["name"],
-            ["registry-console"] if "enterprise" in deployment_type else [],  # include enterprise-only image names
-            openshift_release,
-            is_containerized,
-        ))
+        # images for containerized components
+        if get_var(task_vars, "openshift", "common", "is_containerized"):
+            components = set()
+            if 'nodes' in host_groups:
+                components.update(["node", "openvswitch"])
+            if 'masters' in host_groups:  # name is "origin" or "ose"
+                components.add(image_info["name"])
+            for component in components:
+                required.add("{}/{}:{}".format(image_info["namespace"], component, image_tag))
+            if 'etcd' in host_groups:  # special case, note it is the same for origin/enterprise
+                required.add("registry.access.redhat.com/rhel7/etcd")  # and no image tag
 
-        # append images with qualified image tags to our list of required images.
-        # these are images with a (v0.0.0.0) tag, rather than a standard release
-        # format tag (v0.0). We want to check this set in both containerized and
-        # non-containerized installations.
-        images.update(
-            self.required_qualified_docker_images(
-                image_info["namespace"],
-                image_info["name"],
-                openshift_image_tag,
-            ),
-        )
-
-        return images
-
-    @staticmethod
-    def required_docker_images(namespace, name, additional_image_names, version, is_containerized):
-        if is_containerized:
-            return ["{}/{}:{}".format(namespace, name, version)] if name else []
-
-        # include additional non-containerized images specific to the current deployment type
-        return ["{}/{}:{}".format(namespace, img_name, version) for img_name in additional_image_names]
-
-    @staticmethod
-    def required_qualified_docker_images(namespace, name, version):
-        # pylint: disable=invalid-name
-        return [
-            "{}/{}-{}:{}".format(namespace, name, suffix, version)
-            for suffix in ["haproxy-router", "docker-registry", "deployer", "pod"]
-        ]
+        return required
 
     def local_images(self, images, task_vars):
         """Filter a list of images and return those available locally."""
@@ -124,7 +132,8 @@ class DockerImageAvailability(DockerHostMixin, OpenShiftCheck):
         ]
 
     def is_image_local(self, image, task_vars):
-        result = self.module_executor("docker_image_facts", {"name": image}, task_vars)
+        """Check if image is already in local docker index."""
+        result = self.execute_module("docker_image_facts", {"name": image}, task_vars=task_vars)
         if result.get("failed", False):
             return False
 
@@ -132,6 +141,7 @@ class DockerImageAvailability(DockerHostMixin, OpenShiftCheck):
 
     @staticmethod
     def known_docker_registries(task_vars):
+        """Build a list of docker registries available according to inventory vars."""
         docker_facts = get_var(task_vars, "openshift", "docker")
         regs = set(docker_facts["additional_registries"])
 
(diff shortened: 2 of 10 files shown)


sosiouxme:
aos-ci-test

sosiouxme:
Expect tests to fail, just want to see how they fail.

[test]

sosiouxme:
aos-ci-test

sosiouxme:
aos-ci-test

sosiouxme:
@sdodson @stevekuznetsov Re [this failure](https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_extended_conformance_install_with_status_check/545/consoleFull#77170979858b6e51eb7608a5981914356) our check seems to be reporting that the images legitimately are not there (locally or on docker.io) at the version determined from the RPM (v3.6.0). I have spent some time trying to figure out whether this is expected or not and I just don't have enough context to know. Do we expect images like `openshift/origin-deployer:v3.6.0` to be available during this job? If so, I need to see why it's reporting they're not. If not, I guess just disable the check.

sosiouxme:
aos-ci-test

stevekuznetsov:
This is expected. We do not tag images as the release version in our builds, but instead tag them with the SHA of the commit used to build  them. We pass this in to OpenShift-Ansible on the build, as well: 

```
+ ansible-playbook ... -e openshift_pkg_version=-3.6.0-0.alpha.2.73.1aded21 -e 'oreg_url=openshift/origin-${component}:1aded21' /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml
```

For future reference, Docker information will be in `generated/docker.info` in the S3 artifacts post-build, so you can see the output of things like `docker images` without having to log into a machine.

sosiouxme:
On Mon, Jun 12, 2017 at 10:13 AM, Steve Kuznetsov <notifications@github.com>
wrote:

> + ansible-playbook ... -e openshift_pkg_version=-3.6.0-0.alpha.2.73.1aded21
> -e 'oreg_url=openshift/origin-${component}:1aded21'
> /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml
>
>
Ah yes... this says to me that the check should consult oreg_url if it
exists :) Thanks!

sosiouxme:
aos-ci-test

sosiouxme:
[test] may pass now that `oreg_url` is consulted.

sosiouxme:
aos-ci-test

sosiouxme:
@sdodson any concern with adding docker and RPM checks to `byo/config.yml`?

stevekuznetsov:
We're pretty much always running bleeding edge Docker releases now in CI so just make sure that we don't hobble our ability to do that by checking for a stable RPM or whatever.

roles/openshift_health_checker/openshift_checks/package_availability.py
```diff
@@ -36,7 +36,6 @@ def master_packages(rpm_prefix):
             "bash-completion",
             "cockpit-bridge",
             "cockpit-docker",
-            "cockpit-kubernetes",
             "cockpit-shell",
             "cockpit-ws",
```
sdodson: All of the cockpit packages are optional, why was cockpit-kubernetes removed?

sdodson:
@sosiouxme looks good to me, fine including it in byo.

roles/openshift_health_checker/openshift_checks/package_availability.py
sosiouxme: The cockpit packages were included a few releases ago because I found that a byo install would expect to install them at that time. May not be the case anymore? To some degree, nearly all packages are optional, but if they're missing from your repos it's a sign you're missing something and likely to break somewhere later (... if you're doing an RPM install). Honestly I expect to have to tweak the `package_availability` check once people are actually running it and seeing problems; just not sure what actual users with actual inventories are going to run into.

The cockpit-kubernetes RPM doesn't seem to be available in Origin and nothing seems to break without it so I removed it. I think it migrated to being run as a container only.

sosiouxme:
@stevekuznetsov a valid concern... it does [check that the docker minor version is supported for the release](https://github.com/openshift/openshift-ansible/blob/e768cf9420b967527d0e7b2323aedc131c9a27a6/roles/openshift_health_checker/openshift_checks/package_version.py#L92) and currently that means it expects 1.12 (specifically) for v3.4+. Are we getting docker 1.13+ anytime soon? I kind of expect the excluders will also have something to say about that.

stevekuznetsov:
Checking for minor version will be OK.

sosiouxme:
@sdodson et al - One concern is that if they don't have docker installed and running, the `docker_storage` check will fail the install, where previously byo would have installed and run docker (albeit with unsupported loopback storage). I'm a little divided on whether to consider this a feature (docker setup is supposed to be a prereq) or one annoyance too much (in which case I could make the check just pass when there is no docker yet). Leaning toward "feature" but I'm biased.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -124,14 +129,18 @@ def local_images(self, images, task_vars):
         ]
 
     def is_image_local(self, image, task_vars):
-        result = self.module_executor("docker_image_facts", {"name": image}, task_vars)
+        """Check if image is already in local docker index."""
+        result = self.execute_module("docker_image_facts",
+                                     module_args={"name": image},
+                                     task_vars=task_vars)
```
rhcarvalho: Since `module_args` is the second argument, we could omit its name.

```python
result = self.execute_module("docker_image_facts", {"name": image}, task_vars=task_vars)
```

Good catch on the wrong argument passing! :+1: :medal_military: 

I don't really know yet when `tmp` != `None`, but we should probably be forwarding the value of `tmp` that `run` receives as argument...

roles/openshift_health_checker/openshift_checks/package_availability.py
```diff
@@ -25,7 +25,7 @@ def run(self, tmp, task_vars):
             packages.update(self.node_packages(rpm_prefix))
 
         args = {"packages": sorted(set(packages))}
-        return self.execute_module("check_yum_update", args, tmp, task_vars)
+        return self.execute_module("check_yum_update", module_args=args, tmp=tmp, task_vars=task_vars)
```
rhcarvalho: It is strange that this check was passing `tmp` and all the other are not?!

roles/openshift_health_checker/test/docker_image_availability_test.py
```diff
@@ -53,6 +53,7 @@ def execute_module(module_name, module_args, task_vars):
         ),
         openshift_deployment_type='origin',
         openshift_image_tag='3.4',
+        group_names=['nodes', 'masters'],
```
rhcarvalho: Hmm all the additions are `['nodes', 'masters']`... what about other values?

roles/openshift_health_checker/test/docker_image_availability_test.py
```diff
@@ -174,6 +178,62 @@ def execute_module(module_name=None, module_args=None, tmp=None, task_vars=None)
...
     assert not actual.get("failed", False)
+
+
+@pytest.mark.parametrize("deployment_type, is_containerized, groups, oreg_url, expected", [
+    (
+        "origin", False, ['nodes'], None,
+        set([
+            'openshift/origin-' + infra + ':latest'
+            for infra in NODE_IMAGE_SUFFIXES
+        ])
```
rhcarvalho: Wouldn't this be easier to read if we'd write the set items explicitly?

It feels like the test case is as complicated as reading the implementation ;)

playbooks/byo/openshift-cluster/config.yml
```diff
@@ -15,6 +15,11 @@
       checks:
       - disk_availability
       - memory_availability
+      - package_availability
+      - package_update
+      - package_version
+      - docker_image_availability
+      - docker_storage
```
sosiouxme: aos-ci-test does not have proper docker storage set up so I had it disabled for now (and it passes). Regular CI showed I needed to take `oreg_url` into account but once I did that it passed fine.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -70,51 +72,54 @@ def run(self, tmp, task_vars):
...
+            return required
+
+        # template for images that run on top of OpenShift
+        image_url = "{}/{}-{}:{}".format(image_info["namespace"], image_info["name"], "${component}", "${version}")
+        image_url = get_var(task_vars, "oreg_url", default="") or image_url
+        if 'nodes' in host_groups:
+            for suffix in NODE_IMAGE_SUFFIXES:
+                required.add(image_url.replace("${component}", suffix).replace("${version}", image_tag))
+            # The registry-console is for some reason not prefixed like the other components.
+            # TODO: figure out when registry-console is needed and name for origin/enterprise.
```
sosiouxme: I'll try to find out if this is important and what to expect. Will remove TODO if not.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
sosiouxme: Good point on the indentation, forgot about that.

It's the second arg now, are we sure will it be for future releases of ansible? Seemed like future-proofing. But would be nice to fit on one line.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -151,13 +160,12 @@ def available_images(self, images, registries, task_vars):
...
-        """Uses Skopeo to determine if required image exists in a given registry."""
+        """Uses Skopeo to determine if required image exists with a given registry."""
 
-        cmd_str = "skopeo inspect docker://{registry}/{image}".format(
-            registry=registry,
-            image=image,
-        )
+        # prepend registry only if image does not already have one
+        if image.count("/") < 2:
+            image = registry + "/" + image
```
sosiouxme: If the image is supposed to come from a specific registry (as with the etcd image or if they've put a registry in `oreg_url`) then it's not going to be accurate to look for it from just any registry... even if it's there, it wouldn't be used.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -151,13 +160,12 @@ def available_images(self, images, registries, task_vars):
...
 
-        args = {"_raw_params": cmd_str}
-        result = self.module_executor("command", args, task_vars)
-        return not result.get("failed", False) and result.get("rc", 0) == 0
+        args = {"_raw_params": "skopeo inspect docker://" + image}
+        result = self.execute_module("command", module_args=args, task_vars=task_vars)
+        return result.get("rc", 0) == 0 and not result.get("failed")
```
sosiouxme: `not` always returns a boolean so I don't see how this would return `None`.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
rhcarvalho: Yeah, after I understood what was going on it finally made sense.
They (Ansible) should not change those arguments or they'd break people... the problem was within our own code apparently we started ignoring `tmp` in some places and only caught it now.

roles/openshift_health_checker/openshift_checks/package_availability.py
sosiouxme: yes...

roles/openshift_health_checker/test/docker_image_availability_test.py
sosiouxme: I didn't want to repeat them for every test case... but I can if you want

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
rhcarvalho: Ah, you are correct, ignore that.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
sosiouxme: Yep. I don't know why that ever worked :)

If you don't want to clutter things up with named args that Ansible will never change, I can change that back.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
rhcarvalho: I mean, we can change this method signature to:

```python
def is_available_skopeo_image(self, image, run_command):
    result = run_command({"_raw_params": "skopeo inspect docker://" + image})
    return result.get("rc", 0) == 0 and not result.get("failed")
```

Where `run_command` knows about `task_vars` and `tmp` (or we pass in those and call `self.execute_module`).

This way `image` is whatever needs to go to the Skopeo call.

---

Aha, there have been no changes to the call sites of `is_available_skopeo_image`... so it means that we may be calling it multiple times with the same `image` and different `registry` and this new code simply makes us ignore the registry for certain images...

Maybe that was your intention?

The only thing to fix perhaps is to avoid calling it N times when the registry is pre-determined.

roles/openshift_health_checker/test/docker_image_availability_test.py
rhcarvalho: I trust your take, perhaps looking at the alternative will be equally bad or worse :)

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
rhcarvalho: The named args is fine, but the related problem here is that we ignored `tmp`...

It possibly worked because `task_vars` was not needed by those modules we call, maybe?!

sosiouxme: I don't think any module we didn't write actually needs task_vars. We also obviously don't need tmp in most places; the problem was that it was getting a dict for tmp in at least one case and trying to use it. Probably most modules never use that either, but I would've expected to notice for the one that did.

roles/openshift_health_checker/test/docker_image_availability_test.py
sosiouxme: I have one test case with etcd. I can add others for giggles but those are the only three that get containerized images [AFAICS](https://docs.openshift.org/latest/install_config/install/rpm_vs_containerized.html#install-config-install-install-methods-containerized).

roles/openshift_health_checker/test/docker_image_availability_test.py
sosiouxme: actually as the edge cases increased I liked your suggestion better.

sosiouxme:
aos-ci-test

sosiouxme:
@rhcarvalho I think I addressed your suggestions, have another look...

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -147,17 +157,21 @@ def available_images(self, images, registries, task_vars):
...
 
-    def is_available_skopeo_image(self, image, registry, task_vars):
-        """Uses Skopeo to determine if required image exists in a given registry."""
+    def is_available_skopeo_image(self, image, registries, task_vars):
+        """Use Skopeo to determine if required image exists in known registry(s)."""
+
+        # if image does already includes a registry, just use that
+        if image.count("/") > 1:
+            registry, image = image.split("/", 1)
+            registries = [registry]
```
rhcarvalho: IMHO ignoring the `registries` argument altogether depending on the value of `image` makes the behavior surprising.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -34,7 +34,7 @@ def run(self, tmp, task_vars):
             }
 
         # attempt to get the docker info hash from the API
-        info = self.execute_module("docker_info", {}, task_vars)
+        info = self.execute_module("docker_info", {}, task_vars=task_vars)
```
rhcarvalho: As I said before, I don't know when Ansible passes to our action plugin `tmp != None`.
I think the correct behavior here and the other similar places would be to forward whatever value of `tmp` `run` has received.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
sosiouxme: Would it be less surprising if I called it `additional_registries` or `potential_registries`?

It just made the calling logic less ugly to have this method take care of figuring out which registry to use.

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: You're right, although I don't think it breaks anything now, if we get a `tmp` we should pass it on in case anything would use it. Do we need to hold up this PR for that or can it be a separate one?

sosiouxme: In fact I'd like to internalize `task_vars` and `tmp` so we don't have to pass them around all the time.

rhcarvalho: > In fact I'd like to internalize task_vars and tmp so we don't have to pass them around all the time.

:+1: 

The "difficulty" is that they are part of the `run` method. In terms of OOP it is undesirable to set instance variables at that point... 

The `run` method exists in the first place to mimic the API of the action plugin... but we may reconsider the whole thing if there's a better way.

sosiouxme: We can set variables at the time the checks are instantiated (which happens inside running our action plugin, so vars should be accurate). We only ever run them once per instantiation. I had a branch where I coded all this out a few months ago but never worked up a proper PR.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
rhcarvalho: The complication seem to arise from the fact that some images must come from a specific registry, while others come from a list of registries... hmm...

Would it make sense to (re)define the return value of `required_images` such that it can be used consistently without special-casing what has a fixed registry and what may come from different registries?

```python
{
    'image': {'registries': ['possible', 'registries']},
    'openshift/origin-pod:vtest': {'registries': ['docker.io', 'registry.access.redhat.com']}, # multiple registries possible
    'openshift3/registry-console': {'registries': ['registry.access.redhat.com']}, # must come from a specific registry
    # ...
}
```

sosiouxme: I genuinely think that would be more confusing.

Also do you want this to make devcut today?

sosiouxme:
I'd like to [merge] this for devcut and save any tweaks for a followup.

PR #4399, opened 2017-06-09
Disable excluder only on nodes that are not masters

ingvagabund:
If a master is node as well, the node upgrade is skipped. So we should not disable excluders on such nodes.


ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

sdodson:
[merge]

ingvagabund:
Very similar fix need to be backported to 1.5, 1.4 and 1.3

PR #4400, opened 2017-06-09
replace repoquery with module

ingvagabund:
SSIA

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/pre/verify_upgrade_targets.yml
+++ b/playbooks/common/openshift-cluster/upgrades/pre/verify_upgrade_targets.yml
@@ -14,24 +14,26 @@
 - when: not openshift.common.is_containerized | bool
   block:
   - name: Check latest available OpenShift RPM version
-    command: >
-      {{ repoquery_cmd }} --qf '%{version}' "{{ openshift.common.service_type }}"
-    failed_when: false
-    changed_when: false
-    register: avail_openshift_version
+    repoquery:
+      name: "{{ openshift.common.service_type }}"
+      ignore_excluders: true
+    register: repoquery_out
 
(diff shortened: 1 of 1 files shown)


ingvagabund:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/pre/verify_upgrade_targets.yml
```diff
@@ -14,24 +14,26 @@
 - when: not openshift.common.is_containerized | bool
   block:
   - name: Check latest available OpenShift RPM version
-    command: >
-      {{ repoquery_cmd }} --qf '%{version}' "{{ openshift.common.service_type }}"
-    failed_when: false
-    changed_when: false
-    register: avail_openshift_version
+    repoquery:
```
sdodson: Needs this to ensure that the lib_utils modules have been imported?
```
- include_role:
    name: lib_utils
```

sdodson:
The tests will tell us for sure if it's not imported already but they won't make it clear what happens if this playbook were called in isolation but given how deep this playbook is I don't think we'd ever expect someone to call it directly. So I guess if the tests pass then it's good enough for me.

playbooks/common/openshift-cluster/upgrades/pre/verify_upgrade_targets.yml
ingvagabund: It should but it is already imported by other play. The verify stuff should go into a role most likely. Then, it will have lib_utils as its dependency.

ingvagabund:
[merge]

ingvagabund:
`rm: cannot remove ‘/tmp/etcd/member’: Permission denied` CI issue

PR #4403, opened 2017-06-09
Add systemctl daemon-reload handler to openshift_node

abutcher:
Notify `reload systemd units` handler within the `openshift_node` role whenever we modify files within `/etc/systemd` instead of reloading every time.

Note that we're using the handler as an alternative to `daemon_reload: yes` via the `systemd` module until we're requiring ansible 2.3+.

I also made some minor cosmetic changes to some existing handlers :nail_care:

### DIFF
--- a/roles/openshift_node/handlers/main.yml
+++ b/roles/openshift_node/handlers/main.yml
@@ -1,6 +1,8 @@
 ---
 - name: restart openvswitch
-  systemd: name=openvswitch state=restarted
+  systemd:
+    name: openvswitch
+    state: restarted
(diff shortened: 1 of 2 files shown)


roles/openshift_node/handlers/main.yml
```diff
@@ -1,6 +1,8 @@
 ---
 - name: restart openvswitch
-  systemd: name=openvswitch state=restarted
+  systemd:
+    name: openvswitch
```
ashcrow: Thank you for multiline.

sdodson:
aos-ci-test

abutcher:
[merge]

PR #4412, opened 2017-06-12
logging: ops cluser to allow cluster-reader to read logs

enoodle:
Requiring cluster-admin to view logs is too much. This will activate the option to allow cluster-reader role to read the logs from ES.

BZ reference from CFME: https://bugzilla.redhat.com/show_bug.cgi?id=1454867


enoodle:
ping @jcantrill can you review please?

ewolinetz:
@enoodle This is just the default and can be changed by providing `openshift_logging_es_ops_allow_cluster_reader=True` in your inventory. Given this has always been the default value, I would be hesitant to change that behaviour since it is more open that a customer would expect.

enoodle:
@ewolinetz thanks, I didn't understand that those can be controlled this way.

PR #4425, opened 2017-06-13
Change vsd user nodes parameter name

rparulek:


roles/nuage_node/templates/vsp-openshift.j2
```diff
@@ -9,7 +9,7 @@ enterpriseName: {{ enterprise }}
 # Name of the domain in which pods will reside
 domainName: {{ domain }}
 # Name of the VSD user in admin group
-vsdUser: {{ vsduser }}
+vsdUser: {{ vsd_user }}
```
sdodson: vsd_user | default(vsd_user) to preserve backwards compatibility? Your call not sure how much backwards compat matters to you.

rparulek: I think the current code snippet in my PR is ok as per the requirement. Thanks for the feedback though @sdodson

rparulek:
Can any one of the admins approve this PR and merge to master?

PR #4426, opened 2017-06-13
[3.4] Disable excluder only on nodes that are not masters

sdodson:
Backports #4399

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/disable_node_excluders.yml
+++ b/playbooks/common/openshift-cluster/upgrades/disable_node_excluders.yml
@@ -1,6 +1,6 @@
 ---
 - name: Disable excluders
-  hosts: oo_nodes_to_upgrade
+  hosts: oo_nodes_to_upgrade:!oo_masters_to_config
   gather_facts: no
   roles:
   - role: openshift_excluder


sdodson:
aos-ci-test

sdodson:
ci unfortunately won't test upgrade code

sdodson:
Here's what I'm seeing without this PR.

```
$ git diff upstream/release-1.4..upstream/master playbooks/common/openshift-cluster/upgrades/disable_node_excluders.yml
diff --git a/playbooks/common/openshift-cluster/upgrades/disable_node_excluders.yml b/playbooks/common/openshift-cluster/upgrades/disable_node_excluders.yml
index bba0385b6..a66301c0d 100644
--- a/playbooks/common/openshift-cluster/upgrades/disable_node_excluders.yml
+++ b/playbooks/common/openshift-cluster/upgrades/disable_node_excluders.yml
@@ -1,6 +1,6 @@
 ---
 - name: Disable excluders
-  hosts: oo_nodes_to_upgrade
+  hosts: oo_nodes_to_upgrade:!oo_masters_to_config
   gather_facts: no
   roles:
   - role: openshift_excluder
```

ingvagabund:
That is needed for 3.5+ where we introduced node upgrades as part of the control plane upgrade in case the node components are on the same host as the master components. With `!oo_masters_to_config` the node components would never get upgraded on master nodes.

sdodson:
ack, thanks for checking.

PR #4434, opened 2017-06-13
Disable negative caching, set cache TTL to 1s

stevekuznetsov:
When running headless services as pods on OpenShift, no ClusterIP is
assigned. In these cases, peer pods are relying on DNS to locate the
service endpoints. When a pod is deleted or another trigger causes the
endpoint to change, the OpenShift DNS is updated immediately. However,
dnsmasq has a default TTL of 30s, so the wrong response is returned
on name resolution. Removing negative caching and turning the TTL to
a very short 1s should resolve the issue.

/cc @sdodson @bparees @csrwng 

[test]

### DIFF
--- a/roles/openshift_node_dnsmasq/templates/origin-dns.conf.j2
+++ b/roles/openshift_node_dnsmasq/templates/origin-dns.conf.j2
@@ -1,3 +1,5 @@
 no-resolv
 domain-needed
 server=/{{ openshift.common.dns_domain }}/{{ openshift.common.kube_svc_ip }}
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
aos-ci-test

sdodson:
[merge]

sdodson:
Do you need this backported to other branches?

stevekuznetsov:
> Do you need this backported to other branches?

Good point, we'll need it for testing 1.5 and 1.4 I'll do those.

PR #4435, opened 2017-06-13
etcd v3 for clean installs

sdodson:
If we have no master config assume that we're a clean install.
If we're a clean install and we're 3.6 or greater use etcd v3 storage.

Ensures that clean installs get the following config
```
kubernetesMasterConfig:
  apiServerArguments:
    storage-backend:
    - etcd3
    storage-media-type:
    - application/vnd.kubernetes.protobuf
```

### DIFF
--- a/roles/openshift_master/defaults/main.yml
+++ b/roles/openshift_master/defaults/main.yml
@@ -1,4 +1,3 @@
 ---
 openshift_node_ips: []
-# TODO: update setting these values based on the facts
-#openshift_version: "{{ openshift_pkg_version | default(openshift_image_tag | default(openshift.docker.openshift_image_tag | default(''))) }}"
+r_openshift_master_clean_install: false
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -164,6 +164,26 @@
     - restart master api
     - restart master controllers
 
+- name: Configure master to use etcd3 storage backend on 3.6 clean installs
+  yedit:
+    src: /etc/origin/master/master-config.yaml
+    key: "{{ item.key }}"
+    value: "{{ item.value }}"
+  with_items:
+    - key: kubernetesMasterConfig.apiServerArguments.storage-backend
+      value:
+        - etcd3
+    - key: kubernetesMasterConfig.apiServerArguments.storage-media-type
+      value:
+        - application/vnd.kubernetes.protobuf
+  when:
+    - r_openshift_master_clean_install
+    - openshift.common.version_gte_3_6
+  notify:
+    - restart master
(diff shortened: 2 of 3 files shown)


roles/openshift_master/defaults/main.yml
```diff
@@ -1,4 +1,3 @@
 ---
 openshift_node_ips: []
-# TODO: update setting these values based on the facts
-#openshift_version: "{{ openshift_pkg_version | default(openshift_image_tag | default(openshift.docker.openshift_image_tag | default(''))) }}"
+r_openshift_master_clean_install: false
```
ingvagabund: +1

ingvagabund:
aos-ci-test

sdodson:
@mtnbikenc @abutcher @ingvagabund How do we feel about this method for determining if it's a clean install or not?

sdodson:
aos-ci-test

sdodson:
[merge]

roles/openshift_master/tasks/main.yml
```diff
@@ -164,6 +164,20 @@
     - restart master api
     - restart master controllers
 
+- name: Configure master to use etcd3 storage backend on 3.6 clean installs
+  yedit:
+    src: /etc/origin/master/master-config.yaml
+    key: kubernetesMasterConfig.apiServerArguments.storage-backend
+    value:
+      - etcd3
```
sdodson: @smarterclayton This is the only key we need to set? Some places it mentioned switching from json to protobuf but the etcd v3 migration document mentions only setting this key.

smarterclayton: No, you must also have     

```
storage-media-type:
- application/vnd.kubernetes.protobuf
```

smarterclayton:
[severity:blocker]

roles/openshift_master/tasks/main.yml
sdodson: @ingvagabund Need to add this to migration too.

sdodson:
Updated to add the additional key.

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
[merge][severity: bug]

sdodson:
aos-ci-test

sdodson:
[test]

sdodson:
aos-ci-test

sdodson:
[merge][severity: bug]

sdodson:
I believe the bulk failures were the result of a failed upgrade due to a change in CI jobs that has since been resolved.
[merge][severity: bug]

Issue #4413, opened 2017-06-12
RFE: disk space checks on root filesystem


dustymabe:
on commit dc12c4b

Was able to get my system in an interesting state because of limited disk space on the root filesystem. I didn't realize it but there was a very limited amount of space available on the root filesystem and when openshift-ansible tried to cp the `/usr/local/bin/openshift` binary out of the container and into the host, it failed.

```
TASK [openshift_cli : Copy client binaries/symlinks out of CLI image for use on the host] ************************************************************************************************************************
*************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: OSError: [Errno 28] No space left on device
fatal: [45.55.51.130]: FAILED! => {
    "changed": false, 
    "failed": true, 
    "module_stderr": "Traceback (most recent call last):\n  File \"/usr/lib64/python3.6/shutil.py\", line 544, in move\n    os.rename(src, real_dst)\nOSError: [Errno 18] Invalid cross-device link: '/tmp/tmp_3v8
ij06/openshift' -> '/usr/local/bin/openshift'\n\nDuring handling of the above exception, another exception occurred:\n\nTraceback (most recent call last):\n  File \"/tmp/ansible_i84p32ax/ansible_module_openshif
t_container_binary_sync.py\", line 134, in <module>\n    main()\n  File \"/tmp/ansible_i84p32ax/ansible_module_openshift_container_binary_sync.py\", line 125, in main\n    binary_syncer.sync()\n  File \"/tmp/an
sible_i84p32ax/ansible_module_openshift_container_binary_sync.py\", line 67, in sync\n    self._sync_binary('openshift')\n  File \"/tmp/ansible_i84p32ax/ansible_module_openshift_container_binary_sync.py\", line
 105, in _sync_binary\n    shutil.move(src_path, dest_path)\n  File \"/usr/lib64/python3.6/shutil.py\", line 558, in move\n    copy_function(src, real_dst)\n  File \"/usr/lib64/python3.6/shutil.py\", line 257, 
in copy2\n    copyfile(src, dst, follow_symlinks=follow_symlinks)\n  File \"/usr/lib64/python3.6/shutil.py\", line 122, in copyfile\n    copyfileobj(fsrc, fdst)\n  File \"/usr/lib64/python3.6/shutil.py\", line 
82, in copyfileobj\n    fdst.write(buf)\nOSError: [Errno 28] No space left on device\n", 
    "module_stdout": "", 
    "rc": 1
}
[...]

PR #4436, opened 2017-06-13
Verify disk availability in multiple paths; enable in containerized installs

rhcarvalho:
Related to #4413.

Update disk availability check (run pre-install) to ensure that there's at least 1 GB available in `/usr/local/bin` and the temp dir (determined as per https://docs.python.org/2/library/tempfile.html#tempfile.tempdir)

Note: currently the disk availability check is not run in containerized installs. This PR enables it in order to implement the RFE from #4413. We need to confirm that the [documented disk requirements](https://docs.openshift.org/latest/install_config/install/prerequisites.html#system-requirements) apply to both types of install (/cc @sdodson). Probably yes, according to [Installing on Containerized Hosts: Storage Requirements](https://docs.openshift.org/latest/install_config/install/rpm_vs_containerized.html#containerized-storage-requirements).

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/disk_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/disk_availability.py
@@ -1,4 +1,6 @@
 # pylint: disable=missing-docstring
+import os.path
+
 from openshift_checks import OpenShiftCheck, OpenShiftCheckException, get_var
 from openshift_checks.mixins import NotContainerizedMixin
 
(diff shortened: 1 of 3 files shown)


rhcarvalho:
@dustymabe FYI

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -12,56 +14,77 @@ class DiskAvailability(NotContainerizedMixin, OpenShiftCheck):
...
-        if free_bytes < min_free_bytes:
-            return {
-                'failed': True,
-                'msg': (
-                    'Available disk space ({:.1f} GB) for the volume containing '
-                    '"/var" is below minimum recommended space ({:.1f} GB)'
-                ).format(float(free_bytes) / 10**9, float(min_free_bytes) / 10**9)
-            }
+            # REVIEW: what to do with this when there are multiple paths to be checked?!
+            # configured_min = int(get_var(task_vars, "openshift_check_min_host_disk_gb", default=0)) * 10**9
```
rhcarvalho: @sosiouxme any suggestions for what to do with this variable going forward?

We could make it a `dict` of "path: size" items but I'd rather only add that extra complexity if we have a use case to set custom values like that instead of disabling the check altogether.

dustymabe:
> @dustymabe FYI

thanks

rhcarvalho:
FWIW, we have a check that verifies the available disk space before an install. Previously it was enabled only for RPM installs, and this PR enables it for containerized installs as well.

With that change in, the `fedora/25/atomic` test fails because it doesn't have the minimum documented space in `/var`: [output.log](https://s3.amazonaws.com/aos-ci/ghprb/openshift/openshift-ansible/8f6d9dd59fdf0801e1299608f451e8f5b6355370.0.1497449882859858155/output.log)

```
Available disk space in "/var" (1.7 GB) is below minimum recommended (15.0 GB)
```

We may disable it in CI... but there might be a better alternative involving the future of the variable `openshift_check_min_host_disk_gb`.

roles/openshift_health_checker/openshift_checks/disk_availability.py
rhcarvalho: Idea: treat `openshift_check_min_host_disk_gb` as

1. if it is an `int`, assume it is related to `'/var'`; otherwise
2. it should be a `dict` of `path: int(gb)` pairs

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -12,56 +14,78 @@ class DiskAvailability(NotContainerizedMixin, OpenShiftCheck):
...
+        ansible_mounts = {mount['mount']: mount for mount in ansible_mounts}
+
+        for path, recommendation in self.recommended_disk_space_bytes.items():
+            free_bytes = self.free_bytes(path, ansible_mounts)
+            recommended_bytes = max(recommendation.get(name, 0) for name in group_names)
 
-        recommended_min = max(self.recommended_disk_space_bytes.get(name, 0) for name in group_names)
-        configured_min = int(get_var(task_vars, "openshift_check_min_host_disk_gb", default=0)) * 10**9
-        min_free_bytes = configured_min or recommended_min
+            # REVIEW: what to do with this when there are multiple paths to be checked?!
```
juanvallejo: Maybe only take into account a user-defined limit for `/var`?

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -1,9 +1,11 @@
 # pylint: disable=missing-docstring
+import os.path
+import tempfile
+
 from openshift_checks import OpenShiftCheck, OpenShiftCheckException, get_var
-from openshift_checks.mixins import NotContainerizedMixin
 
 
-class DiskAvailability(NotContainerizedMixin, OpenShiftCheck):
+class DiskAvailability(OpenShiftCheck):
```
rhcarvalho: The docs loop back into the same reqs as for RPM install: https://docs.openshift.org/latest/install_config/install/rpm_vs_containerized.html#containerized-storage-requirements

roles/openshift_health_checker/openshift_checks/disk_availability.py
rhcarvalho: > It would be kinda cool if it could even add filesystems we don't otherwise check...

Interesting idea!

rhcarvalho: I added a TODO so we can leverage the idea in the future, but won't implement it right now.

roles/openshift_health_checker/openshift_checks/disk_availability.py
```diff
@@ -1,9 +1,11 @@
 # pylint: disable=missing-docstring
```
rhcarvalho: Done

rhcarvalho:
https://s3.amazonaws.com/aos-ci/ghprb/openshift/openshift-ansible/0836d176ac66aeabaa9a40f83ef6e24adeb5e8a5.0.1498134779903179623/index.html
```
Available disk space in "/var" (1.7 GB) is below minimum recommended (15.0 GB)
```

I'll lower the required disk space in `/var`.

rhcarvalho:
aos-ci-test

rhcarvalho:
The aos-ci-test job needs to have a custom value for the disk availability check by setting this variable:

```
openshift_check_min_host_disk_gb=10
```

Where 10 GB is arbitrary, but such that it is enough for an install to succeed.

Last [logs](https://aos-ci.s3.amazonaws.com/openshift/openshift-ansible/jenkins-openshift-ansible-test-matrix-CONTAINERIZED=_containerized,OSE_VER=3.5,PYTHON=System-CPython-2.7,TOPOLOGY=openshift-cluster-containerized,TargetBranch=master,nodes=openshift-ansible-slave-711/ffeac220ea2764c02a4ebd9b5e006e82886150bd.txt) with failure in `aos-ci-jenkins/OS_3.5_containerized`:
```
Available disk space in "/var" (14.7 GB) is below minimum recommended (40.0 GB)
```

Similar error in `aos-ci-jenkins/OS_3.6_containerized` [logs](https://aos-ci.s3.amazonaws.com/openshift/openshift-ansible/jenkins-openshift-ansible-test-matrix-CONTAINERIZED=_containerized,OSE_VER=3.6,PYTHON=System-CPython-2.7,TOPOLOGY=openshift-cluster-containerized,TargetBranch=master,nodes=openshift-ansible-slave-711/ffeac220ea2764c02a4ebd9b5e006e82886150bd.txt).

rhcarvalho:
aos-ci-test

rhcarvalho:
Ready to merge when the queue re-opens.

rhcarvalho:
[merge]

PR #4445, opened 2017-06-14
images: set atomic.run=once label

giuseppe:

### DIFF
--- a/images/installer/Dockerfile
+++ b/images/installer/Dockerfile
@@ -12,7 +12,8 @@ LABEL name="openshift/origin-ansible" \
       io.k8s.display-name="openshift-ansible" \
       io.k8s.description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain and check an OpenShift cluster" \
       io.openshift.expose-services="" \
-      io.openshift.tags="openshift,install,upgrade,ansible"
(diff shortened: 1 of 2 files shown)


codificat:
@giuseppe openshift/playbook2image#3 is about to merge, and after that the openshift-ansible image will inherit system container support (including the atomic.run=once label) from playbook2image... therefore, adding this label will be redundant soon.

On the other hand, we might want to remove other [soon to become] redundant system-container bits from openshift-ansible as they will be already part of the base image... what do you think?

giuseppe:
@codificat that is great!  The only question is if we want the "oneshot" feature for every container that uses `playbook2image`, are there playbooks where it is desiderable to run more than once (e. g., some replacement for cloud-init?).  If that is not an issue, I can close this PR if you are fine with it.

codificat:
I don't really see a use case where you'd prefer the non-oneshot approach... but if there is, one can always use the `docker build` approach to playbook2image and remove that label... the suggestion in the p2i PR was to have *run once* by default and see if there's feedback that suggests we should change it.

So, if you're ok with that let's close this PR, and if playbook2image later changes to remove the "oneshot by default" we can add it to openshift-ansible.

giuseppe:
yes sure, I am fine with it.  Thanks for the help.

PR #4447, opened 2017-06-14
Create disk_availability.py

wmhussain:
Updating /var requirement to 28 GB as it fails in azure with below error. 
Host:     aacl-master-0  (Standard_D4_v2)
     Play:     Verify Requirements
     Details:  check "disk_availability":
               **Available disk space (29.6 GB) for the volume containing "/var" is below minimum recommended space (40.0 GB)**

Azure VM disk size is 32 GB only.

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/disk_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/disk_availability.py
@@ -12,7 +12,7 @@ class DiskAvailability(NotContainerizedMixin, OpenShiftCheck):
     # Values taken from the official installation documentation:
     # https://docs.openshift.org/latest/install_config/install/prerequisites.html#system-requirements
     recommended_disk_space_bytes = {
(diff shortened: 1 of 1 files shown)


sdodson:
Our documentation says 40GiB. I think we should fail if we're less than that. You may of course override the value but we want admins to know that their environment doesn't meet the minimum documented requirements.

sosiouxme:
@wmhussain can Azure not supply additional storage that can be added to the root VG, or used directly as a separate `/var` partition? Does it have to be the VM disk?

If we're not going to actually require the documented requirements then we should get PM to agree and update the documentation and checks.

wmhussain:
In Azure, VM images are provisioned with flat partitions, no LVM is there and all images are with sda of 32 GB. I tired deploying multiple times with 25GB as /var in master. it worked successfully every time.

michaelgugino:
Closing due to inactivity.

PR #4448, opened 2017-06-14
Remove invalid when: from vars:

mtnbikenc:
This PR removes a couple of `when:` statements which were listed in the `vars:` section of a play.  They are essentially doing nothing.  Fixes Ansible WARNING for use of reserved name `when:`.

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -47,8 +47,7 @@
                                                     | union(groups['oo_etcd_to_config'] | default([])))
                                                 | oo_collect('openshift.common.hostname') | default([]) | join (',')
                                                 }}"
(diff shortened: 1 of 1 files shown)


mtnbikenc:
@sdodson I'm not sure about the intent of these when statements.  If they should be moved somewhere else I can do that.  This is really old code so the context of the original change may be lost.

sdodson:
aos-ci-test

sdodson:
Yeah, I think those never worked. It was an attempt to avoid some undefined errors but i think we got everything defaulted to empty sets in the end.

mtnbikenc:
[merge]

mtnbikenc:
[merge]

sdodson:
[merge]

mtnbikenc:
Flake https://github.com/openshift/origin/issues/15356 HAProxy router

mtnbikenc:
[merge]

PR #4457, opened 2017-06-15
Clarifying some language to make it clearer

cloudeigh:
Replaced references to a subnet ID (which is a thing in AWS) with subnet CIDR range, which is what it is asking for. 
Added some helper text around the use of ssh-agent.


michaelgugino:
Code path removed.  Closing.

PR #4458, opened 2017-06-15
Add the other featured audit-config paramters as example

git001:
When the audit_config ist enabled all AUDIT messages
will be logged into the system log.

This example creates a dedicated audit log.


sdodson:
Fixes this? https://bugzilla.redhat.com/show_bug.cgi?id=1447774

sdodson:
LGTM

PR #4459, opened 2017-06-15
Fixed version comparing when upgraded to some minor version

quckly:
Example if you try upgrade to 1.5.1 OpenShift Origin, there are will be fail on this line.
I changed compare like using 2 version_compare above.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/pre/verify_inventory_vars.yml
+++ b/playbooks/common/openshift-cluster/upgrades/pre/verify_inventory_vars.yml
@@ -34,4 +34,4 @@
       msg: >
         openshift_release is {{ openshift_release }} which is not a
         valid release for a {{ openshift_upgrade_target }} upgrade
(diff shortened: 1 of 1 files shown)


dgoodwin:
Could you clarify what relevant values you were setting in inventory? From what I can see openshift_release and openshift_upgrade_target are both supposed to just be "1.5" in this case, if setting openshift_release to "1.5.1" that would be incorrect use of the variable.

quckly:
`deployment_type=origin
openshift_release=v1.5.1
containerized=true`
If I set openshift_release=v1.5, openshift will be updated to 1.5.1?

dgoodwin:
Yeah openshift_release should just be "v1.5", and if you run the 1.5 upgrade playbook by default that will take you up to the *latest* 1.5 release. If you want more fine grained control, you can set both openshift_image_tag and openshift_pkg_version. More details avail here: https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.origin.example

Going to close but thank you for the patch just the same!

PR #4462, opened 2017-06-15
calculate replicas depending on the number of shards

ruromero:
I think the number of replicas should be calculated based on the number of instances - the number of primary shards

### DIFF
--- a/roles/openshift_logging/defaults/main.yml
+++ b/roles/openshift_logging/defaults/main.yml
@@ -97,7 +97,7 @@ openshift_logging_es_nodeselector: "{{ openshift_hosted_logging_elasticsearch_no
 # openshift_logging_es_config is a hash to be merged into the defaults for the elasticsearch.yaml
 openshift_logging_es_config: {}
 openshift_logging_es_number_of_shards: 1
-openshift_logging_es_number_of_replicas: 0
+openshift_logging_es_number_of_replicas: "{{ openshift_logging_es_cluster_size - openshift_logging_es_number_of_shards | default(0) }}"
 
 # for exposing es to external (outside of the cluster) clients
 openshift_logging_es_allow_external: False


portante:
Not entirely convinced we want this change for a couple of reasons.

Typically, replication of data is used to increase the durability of the data in the face of failure.  To that end, we would really want to set the value based on the number of failure domains available to the cluster.  Replicating within a failure domain is not necessarily helpful.

Second, here is a table for some values of cluster size replication count:

```
+---------+---------+---------+
| Cluster | Primary | Replica |
| Size    | Shards  | Count   |
+---------+---------+---------+
|    1    |    1    |    0    |
+---------+---------+---------+
|    2    |    1    |    1    |
+---------+---------+---------+
|    3    |    1    |    2    |
+---------+---------+---------+
|    4    |    1    |    3    |
+---------+---------+---------+
|    5    |    1    |    4    |
+---------+---------+---------+
|    6    |    1    |    5    |
+---------+---------+---------+
|    1    |    2    |    0    |
+---------+---------+---------+
|    2    |    2    |    0    |
+---------+---------+---------+
|    3    |    2    |    1    |
+---------+---------+---------+
|    4    |    2    |    2    |
+---------+---------+---------+
|    5    |    2    |    3    |
+---------+---------+---------+
|    6    |    2    |    4    |
+---------+---------+---------+
```

Note that as nodes are added, we start getting into higher replication counts, which will cause too much overhead.

If we gave the user the ability to specify the number of availability zones they have, and how many ES pods they want per zone, we could then default to auto calculating the replication count.

ewolinetz:
@portante would it be easier to document how to calculate the number of necessary replicas we'd want per shard for an index and how to update it with the deployer instead of putting this in as the default for the installation? This is just the default and could be overridden, to me it makes more sense that we'd just document this.

portante:
@ewolinetz, I suppose it depends on how automated we'd want the installer to be.

ewolinetz:
@portante i think I'd prefer we just document on how to size this correctly given it is so easy to provide your own value anyways. This PR is only proposing how to change the replica count but the table you specify also shows acceptable cases where we would increase the primary shard count and not the replicas. Given there are parameters that can be passed to the installer for both I think it would be better to let users control which way they scale and describe how they can do it.

ruromero:
@ewolinetz The intention here was to provide a default value that was something closer to what we had before with the deployer whit the autoexpand_replicas. However @portante is right with the given example which can add a lot of unnecessary replication.
My suggestion is to get the default value be calculated in the same way but limiting it to 2. Adding it to the documentation is also useful but many users may not know about shards and replicas and trust the default values

ruromero:
I think you are right and is better to let the users decide about their replication as it has less downsides having 0 than any other calculated number.
Closing and thanks for the comments 👍

PR #4466, opened 2017-06-15
Ensure only one ES pod per PV

portante:
Unfortunately, the defaults for Elasticsearch prior to v5 allow more
than one "node" to access the same configured storage volume(s).

This change forces this value to `1` to ensure we don't have an ES pod
starting up accessing a volume while another ES pod is shutting down
when reploying. This can lead to "`1`" directories being created in
`/elasticsearch/persistent/${CLUSTER_NAME}/data/${CLUSTER_NAME}/nodes/`. By
default ES uses a "`0`" directory there when only one node is accessing
it.

Fixes BZ https://bugzilla.redhat.com/show_bug.cgi?id=1460564.

### DIFF
--- a/roles/openshift_logging_elasticsearch/templates/elasticsearch.yml.j2
+++ b/roles/openshift_logging_elasticsearch/templates/elasticsearch.yml.j2
@@ -16,6 +16,7 @@ index:
 node:
   master: ${IS_MASTER}
   data: ${HAS_DATA}
+  max_local_storage_nodes: 1
 
 network:
   host: 0.0.0.0


portante:
@ewolinetz, @jcantrill, please consider this not only for 3.6 release, but also for 3.5 and 3.4. Thanks!

jcantrill:
aos-ci-test

jcantrill:
@portante if you have a bug that can be applied to this it may be more sellable to merge before code freeze and make available for backport

ewolinetz:
@portante which versions of ES is this parameter valid for?

portante:
@jcantrill, the BZ is not a bug?

@ewolinetz, it only applies to Elasticsearch 2.0 and later.

jcantrill:
@portante can you modify the commit message to be something like: 'bug XYZ. fixes....'  this helps identify in the git history and there is some automagic that is applied from git to the BZ

portante:
@jcantrill, sure.

portante:
@jcantrill is that works better?

ewolinetz:
@portante I added a label to backport this for 3.5 as well, the 3.4 backport will need  to happen in the origin-aggregated-logging repo however

ewolinetz:
aos-ci-test

ewolinetz:
[merge][severity: bug]

PR #4474, opened 2017-06-16
Rename cockpit-shell -> cockpit-system

rhcarvalho:
The package name has changed.

See
https://bugzilla.redhat.com/show_bug.cgi?id=1461689
https://bugzilla.redhat.com/show_bug.cgi?id=1419718

### DIFF
--- a/.papr.sh
+++ b/.papr.sh
@@ -18,7 +18,9 @@ upload_journals() {
 trap upload_journals ERR
 
 # run the actual installer
-ansible-playbook -vvv -i .papr.inventory playbooks/byo/config.yml
+# FIXME: override openshift_image_tag defined in the inventory until
+# https://github.com/openshift/openshift-ansible/issues/4478 is fixed.
+ansible-playbook -vvv -i .papr.inventory playbooks/byo/config.yml -e "openshift_image_tag=$OPENSHIFT_IMAGE_TAG"
 
 # run a small subset of origin conformance tests to sanity
 # check the cluster NB: we run it on the master since we may
--- a/playbooks/adhoc/uninstall.yml
+++ b/playbooks/adhoc/uninstall.yml
@@ -103,7 +103,7 @@
         - atomic-openshift-sdn-ovs
         - cockpit-bridge
         - cockpit-docker
-        - cockpit-shell
+        - cockpit-system
         - cockpit-ws
         - kubernetes-client
         - openshift
@@ -346,7 +346,7 @@
     - atomic-openshift-master
     - cockpit-bridge
     - cockpit-docker
-    - cockpit-shell
+    - cockpit-system
     - cockpit-ws
     - corosync
     - kubernetes-client
--- a/roles/cockpit/tasks/main.yml
+++ b/roles/cockpit/tasks/main.yml
@@ -3,7 +3,7 @@
   package: name={{ item }} state=present
   with_items:
     - cockpit-ws
-    - cockpit-shell
+    - cockpit-system
     - cockpit-bridge
     - cockpit-docker
     - "{{ cockpit_plugins }}"
--- a/roles/openshift_health_checker/openshift_checks/package_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/package_availability.py
@@ -36,7 +36,7 @@ class PackageAvailability(NotContainerizedMixin, OpenShiftCheck):
             "bash-completion",
             "cockpit-bridge",
             "cockpit-docker",
-            "cockpit-shell",
+            "cockpit-system",
             "cockpit-ws",
             "etcd",
             "httpd-tools",


rhcarvalho:
aos-ci-test

rhcarvalho:
@sosiouxme FYI there's something funny in the Docker Image Availability check I'm trying to debug...

```
  3. Host:     ocp-master
     Play:     Verify Requirements
     Details:  check "docker_image_availability":
               One or more required Docker images are not available:
                   openshift/node:{{ lookup('env', 'OPENSHIFT_IMAGE_TAG') }},
                   openshift/openvswitch:{{ lookup('env', 'OPENSHIFT_IMAGE_TAG') }},
                   openshift/origin-deployer:{{ lookup('env', 'OPENSHIFT_IMAGE_TAG') }},
                   openshift/origin-docker-registry:{{ lookup('env', 'OPENSHIFT_IMAGE_TAG') }},
                   openshift/origin-haproxy-router:{{ lookup('env', 'OPENSHIFT_IMAGE_TAG') }},
                   openshift/origin-pod:{{ lookup('env', 'OPENSHIFT_IMAGE_TAG') }},
                   openshift/origin:{{ lookup('env', 'OPENSHIFT_IMAGE_TAG') }}
               Configured registries: docker.io
```
https://s3.amazonaws.com/aos-ci/ghprb/openshift/openshift-ansible/2388cd13cafe0ff967014a1e7a05e0a5d8e64ff9.0.1497607399834811633/output.log

Looks like we're reading from some variable that was not processed through jinja.

rhcarvalho:
In particular, the value of `openshift_image_tag`:
https://github.com/openshift/openshift-ansible/blame/45518c62e9bf3f7ed0e0420c613f85315b88ba8c/roles/openshift_health_checker/openshift_checks/docker_image_availability.py#L97

rhcarvalho:
The bad value is coming from:
https://github.com/openshift/openshift-ansible/blame/78894125d25d7a3e7bae849cd1243b1472d9e937/.papr.inventory#L10

I need to check if it is valid to use jinja template in the inventory.  @jlebon FYI

jlebon:
I couldn't find any solid reference confirming that it is supported, so we might indeed be abusing Ansible there. We can also make the inventory a Jinja2 template and render it before starting the installer with something like:

```
python -c 'import os, sys, jinja2; tpl = jinja2.Template(sys.stdin.read()); print(tpl.render(OPENSHIFT_IMAGE_TAG=os.environ["OPENSHIFT_IMAGE_TAG"]))' < inventory.j2 > inventory
```

jlebon:
Actually, [this SO answer](https://stackoverflow.com/a/35009576/308136) gives an even shorter way:

```
$ cat tmp.j2
asdf="{{ env['OPENSHIFT_IMAGE_TAG'] }}"
$ export OPENSHIFT_IMAGE_TAG=foobar
$ python -c 'import os, sys, jinja2; tpl = jinja2.Template(sys.stdin.read()); print(tpl.render(env = os.environ))' < tmp.j2
asdf="foobar"
```

rhcarvalho:
@sosiouxme @jlebon 

Our little investigation today brought us to:

1. It is possible to have jinja templates in the inventory.
2. Variables passed to action plugins seem to be in a state before template evaluation, what led us to discover the "feature".


There are a couple of ways we could proceed. The more complete solution may involve evaluating templates in the action plugin, but I'm not comfortable doing that *right now*.

A more expedient solution is to avoid the environment variable for now, iow:

```diff
diff --git a/.papr.inventory b/.papr.inventory
index 23bc992..ed93ee1 100644
--- a/.papr.inventory
+++ b/.papr.inventory
@@ -7,7 +7,7 @@ etcd
 ansible_ssh_user=root
 ansible_python_interpreter=/usr/bin/python3
 deployment_type=origin
-openshift_image_tag="{{ lookup('env', 'OPENSHIFT_IMAGE_TAG') }}"
+openshift_image_tag=v3.6.0-alpha.1
 openshift_master_default_subdomain="{{ lookup('env', 'RHCI_ocp_node1_IP') }}.xip.io"
 openshift_check_min_host_memory_gb=1.9
 
diff --git a/.papr.yml b/.papr.yml
index 16d6e78..e979fc0 100644
--- a/.papr.yml
+++ b/.papr.yml
@@ -32,9 +32,6 @@ packages:
 
 context: 'fedora/25/atomic'
 
-env:
-  OPENSHIFT_IMAGE_TAG: v3.6.0-alpha.1
-
 tests:
   - ./.papr.sh
```

@jlebon @sosiouxme would you agree with such a patch plus an open issue to handle templates in variables?

jlebon:
@rhcarvalho SGTM, though make sure to also remove the `echo` referencing it from `.papr.sh`.

sosiouxme:
@rhcarvalho Works for me as a way to unblock the test, but that might be a pain to have to update the CI config when version tags change. We could also just... disable the check in this env until we decide on a resolution.

rhcarvalho:
@jlebon @sosiouxme alternatively:

```diff
diff --git a/.papr.sh b/.papr.sh
index fe0b97b..f7ada27 100755
--- a/.papr.sh
+++ b/.papr.sh
@@ -18,7 +18,7 @@ upload_journals() {
 trap upload_journals ERR
 
 # run the actual installer
-ansible-playbook -vvv -i .papr.inventory playbooks/byo/config.yml
+ansible-playbook -vvv -i .papr.inventory playbooks/byo/config.yml -e "openshift_image_tag=$OPENSHIFT_IMAGE_TAG"
 
 # run a small subset of origin conformance tests to sanity
 # check the cluster NB: we run it on the master since we may
```

Then all we need to do is drop the `-e ...` part once we can handle the template evaluation.

jlebon:
Sure, that works. Maybe a TODO comment just above explaining why we're overriding the variable?

rhcarvalho:
Now the fedora/25/atomic failed with something unrelated to health checks:

```
Failure summary:

  1. Host:     ocp-node1
     Play:     Configure nodes
     Task:     openshift_node : Start and enable node
     Message:  Unable to start service origin-node: Job for origin-node.service failed because the control process exited with error code.
               See "systemctl status origin-node.service" and "journalctl -xe" for details.
```
https://s3.amazonaws.com/aos-ci/ghprb/openshift/openshift-ansible/f6127f924597c9fcc7da6435e14524253db64d64.0.1497626270645153045/output.log


Will try to re-run.

rhcarvalho:
bot, retest this please

rhcarvalho:
@sdodson this is OK to merge. `fedora/25/atomic` is green; other jobs passed with the PAPR-only changes.

PR #4485, opened 2017-06-17
openshift_checks: store args in check properties

sosiouxme:
Refactor to internalize `task_vars`.

Get rid of deprecated `module_executor` and move `task_vars` into properties so we don't have to pass it around everywhere. Also move `tmp` into properties and make sure it gets filled in on `execute_module`.

In the process, `is_active` and `get_var` become instance methods, and `task_vars` is basically never used directly outside of test code.


roles/openshift_health_checker/action_plugins/openshift_health_check.py
```diff
@@ -92,7 +92,7 @@ def load_known_checks(self):
                         check_name,
                         cls.__module__, cls.__name__,
                         other_cls.__module__, other_cls.__name__))
-            known_checks[check_name] = cls(execute_module=self._execute_module)
+            known_checks[check_name] = cls(execute_module=self._execute_module, tmp=tmp, task_vars=task_vars)
```
sosiouxme: This is the only place we need to pass `tmp` and `task_vars` to the check. Each check is instantiated once per task and host so the stored values are only ever used on one run.

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -19,15 +19,23 @@ class OpenShiftCheckException(Exception):
...
+    """
 
-    def __init__(self, execute_module=None, module_executor=None):
-        if execute_module is module_executor is None:
-            raise TypeError(
-                "__init__() takes either execute_module (recommended) "
-                "or module_executor (deprecated), none given")
-        self.execute_module = execute_module or module_executor
-        self.module_executor = self.execute_module
+    def __init__(self, execute_module, task_vars=None, tmp=None):
```
sosiouxme: This changes the method signature and so breaks any pending checks, but it's kind of unavoidable if we want to do this. And I think it's really time to get rid of `module_executor`.

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -43,13 +51,12 @@ def tags(self):
         """
         return []
 
-    @classmethod
-    def is_active(cls, task_vars):  # pylint: disable=unused-argument
+    def is_active(self):  # pylint: disable=unused-argument,no-self-use
```
sosiouxme: Is there a better way to get pylint to shut up about this? Seems like it would be common enough that your base instance method wouldn't use self.

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -19,15 +19,23 @@ class OpenShiftCheckException(Exception):
...
-    """A base class for defining checks for an OpenShift cluster environment."""
+    """
+    A base class for defining checks for an OpenShift cluster environment.
+
+    Expect a method execute_module, optional dict task_vars, and optional string tmp.
+    execute_module is expected to have a signature compatible with _execute_module
+    from ansible plugins/action/__init__.py, e.g.:
+    def execute_module(module_name=None, module_args=None, tmp=None, task_vars=None, **kwargs)
+    This is stored so that it can be invoked in subclasses as check.execute_module("name", args)
+    which provides the check's task_vars and tmp by default.
```
sosiouxme: If you're wondering why all the explanation here, it's for future me because I keep forgetting what the method signature should be between plugin `_execute_module` and check `execute_module`.

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -62,6 +69,45 @@ def subclasses(cls):
...
+
+        # fill in unspecified tmp as either positional (4th) or keyword arg
+        if len(args) > 3 and args[3] is None:
+            args[3] = self.tmp
+        else:
+            kwargs["tmp"] = kwargs.get("tmp", self.tmp)
+
+        return self._execute_module(*args, **kwargs)
+
+    def get_var(self, *keys, **kwargs):
```
sosiouxme: I meant to mention: since access to `task_vars` now always goes through this, whatever way we decide to solve the lazy evaluation problem only needs to be implemented in this method.

sosiouxme:
aos-ci-test

sosiouxme:
[test]

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: It is possible to do the refactor gradually without breaking on-going PRs, but if you want to avoid that "exercise" I'm fine with it.

Perhaps remove `module_executor` in a separate commit?

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -19,15 +19,23 @@ class OpenShiftCheckException(Exception):
...
-    def __init__(self, execute_module=None, module_executor=None):
-        if execute_module is module_executor is None:
-            raise TypeError(
-                "__init__() takes either execute_module (recommended) "
-                "or module_executor (deprecated), none given")
-        self.execute_module = execute_module or module_executor
-        self.module_executor = self.execute_module
+    def __init__(self, execute_module, task_vars=None, tmp=None):
+        if execute_module is None:
+            raise TypeError("__init__() requires execute_module; none given")
```
rhcarvalho: We don't need to check this.

Sometimes (e.g. in tests) it is convenient to pass `None` if we expect it never to be called.

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: Have you tried keeping it as a `classmethod`? Ah, the class have no idea what is `task_vars` so it must not be a `classmethod`, I see.

It's unfortunate to fiddle with pylint, but we can probably just let it go here, since we have reasons to go against the rule (this is a base implementation meant to be overridden).

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -62,6 +69,45 @@ def subclasses(cls):
...
+        if len(args) > 2 and args[2] is None:
+            args[2] = self.task_vars
+        else:
+            kwargs["task_vars"] = kwargs.get("task_vars", self.task_vars)
+
+        # fill in unspecified tmp as either positional (4th) or keyword arg
+        if len(args) > 3 and args[3] is None:
+            args[3] = self.tmp
+        else:
+            kwargs["tmp"] = kwargs.get("tmp", self.tmp)
```
rhcarvalho: Couldn't we simply replicate the signature of the Ansible method here and then override `task_vars` and `tmp`?

```python
def execute_module(self, module_name=None, module_args=None, tmp=None, task_vars=None, persist_files=False, delete_remote_tmp=True, wrap_async=False):
    if tmp is not None:
        raise RuntimeError('Usage error: tmp is ignored by this method, use _execute_module instead.')
    if task_vars is not None:
        raise RuntimeError('Usage error: task_vars is ignored by this method, use _execute_module instead.')
    return self._execute_module(module_name, module_args, self.tmp, self.task_vars, persist_files, delete_remote_tmp, wrap_async)

File: roles/openshift_health_checker/openshift_checks/__init__.py [hunk 8]
```diff
@@ -62,6 +69,45 @@ def subclasses(cls):
...
+
+        return self._execute_module(*args, **kwargs)
+
+    def get_var(self, *keys, **kwargs):
+        """Helper function to get deeply nested values from task_vars.
```
rhcarvalho: suggestion: s/Helper function to g/G

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
```diff
@@ -124,53 +124,53 @@ def required_images(task_vars):
...
         return [
             image for image in images
-            if self.is_available_skopeo_image(image, registries, task_vars)
+            if self.is_available_skopeo_image(image, default_registries)
         ]
 
-    def is_available_skopeo_image(self, image, registries, task_vars):
+    def is_available_skopeo_image(self, image, default_registries):
         """Use Skopeo to determine if required image exists in known registry(s)."""
+        registries = default_registries
```
rhcarvalho: Hmm, is the rename of the argument really useful if we do this on the first line?

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: Good point, would simplify even a little further.

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: The ansible method signature is rather longer, but I guess it's less hassle to not deal with overlapping args/kwargs and such.

I thought it might be nice to be able to override `tmp` and `task_vars` per-invocation. Don't know if there's a real use case for it though.

roles/openshift_health_checker/openshift_checks/docker_image_availability.py
sosiouxme: method signature "documentation" versus initialization of local variable...

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: I had a separate commit for it originally, but squashed for the refactor... can break it back out if you like.

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: > I thought it might be nice to be able to override tmp and task_vars per-invocation. Don't know if there's a real use case for it though.

To override, one can always call the original function via `self._execute_module`.

sosiouxme: Good point, that does seem better.

sosiouxme:
> It promotes having more external state instead of explicit dependencies.
> Now when we test a method in a check, we may start unnecessarily mocking or
> setting up task_vars when the method doesn't depend on that.
>
> Looking at a method signature won't be enough to know its dependencies.
>

Adding `task_vars` to a method signature isn't going to tell you anything
about which `task_vars` contents need to be provided anyway. You'll still
have to look at the method. Or just do TDD and let your test failure tell
you what it needs :)

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: Actually what's annoying about this is now all your mock `execute_module` methods have to accept all these parameters.

sosiouxme:
aos-ci-test

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: > now all your mock execute_module methods have to accept all these parameters.

Or use `*args, **kwargs` for what is not interesting?

Alternatively, since now this is a method on OpenShiftCheck, we could also use the "mock" fixture to patch it.

rhcarvalho:
> Adding `task_vars` to a method signature isn't going to tell you anything
> about which `task_vars` contents need to be provided anyway.

I agree, `task_vars` is a bag of everything and anything... having it as arg tells little about dependencies.

sosiouxme:
@rhcarvalho what did I get wrong here...

```
An exception occurred during task execution. The full traceback is:
  File "/usr/lib/python2.7/site-packages/ansible/executor/task_executor.py", line 126, in run
    res = self._execute()
  File "/usr/lib/python2.7/site-packages/ansible/executor/task_executor.py", line 511, in _execute
    result = self._handler.run(task_vars=variables)
  File "/home/jenkins/workspace/openshift-ansible-test-matrix/CONTAINERIZED/_NOT_containerized/OSE_VER/3.5/PYTHON/System-CPython-2.7/TOPOLOGY/openshift-cluster/TargetBranch/master/nodes/openshift-ansible-slave/openshift-ansible/roles/openshift_health_checker/action_plugins/openshift_health_check.py", line 66, in run
    r = check.run()
  File "/home/jenkins/workspace/openshift-ansible-test-matrix/CONTAINERIZED/_NOT_containerized/OSE_VER/3.5/PYTHON/System-CPython-2.7/TOPOLOGY/openshift-cluster/TargetBranch/master/nodes/openshift-ansible-slave/openshift-ansible/roles/openshift_health_checker/openshift_checks/package_availability.py", line 29, in run
    return self.execute_module("check_yum_update", args)
  File "/home/jenkins/workspace/openshift-ansible-test-matrix/CONTAINERIZED/_NOT_containerized/OSE_VER/3.5/PYTHON/System-CPython-2.7/TOPOLOGY/openshift-cluster/TargetBranch/master/nodes/openshift-ansible-slave/openshift-ansible/roles/openshift_health_checker/openshift_checks/__init__.py", line 88, in execute_module
    persist_files, delete_remote_tmp, wrap_async)
TypeError: _execute_module() takes at most 7 arguments (8 given)
fatal: [10.8.168.160]: FAILED! => {
    "failed": true, 
    "msg": "Unexpected failure during module execution.", 
    "stdout": ""
}
```

Method signature differences between ansible versions maybe? It runs fine w/ my ansible 2.3...

sosiouxme:
Looks like `wrap_async` is a relatively recent addition.

https://github.com/ansible/ansible/commit/c86a17b7a003ac22c173cfa832ed2368e15015e6#diff-69a29e19a76e98587550dab380483594R591

I can probably just leave it out.

sosiouxme:
aos-ci-test

sosiouxme:
aos-ci-test

roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
```diff
@@ -40,8 +33,10 @@ def run(self, tmp, task_vars):
         # TODO(lmeyer): run it all again for the ops cluster
         return {"failed": False, "changed": False, "msg": 'No problems found with Elasticsearch deployment.'}
 
-    def _not_running_elasticsearch_pods(self, es_pods):
+    def not_running_elasticsearch_pods(self, es_pods):
         """Returns: list of running pods, list of errors about non-running pods"""
+        # FIXME: the docstring above is clearly wrong.
+        # I don't understand this method enough to fix the doscstring.
```
rhcarvalho: @juanvallejo could you help, please?

juanvallejo: Will take a look

juanvallejo: Looks like we only use the method [to determine which pods are running](https://github.com/openshift/openshift-ansible/pull/4485#diff-cf30713e491fc112ba1639d63f6cb276R55) and to return a list of pods we should warn about being in a "not running" state. I think it would be better to have a `running_es_pods` method, and derive the list of non-running pods based on that

sosiouxme: Docstring should say list of NON running pods.

Normally I don't want to do anything about pods that aren't running as they don't hurt anything and could be transient. What I want to do is filter them out so we don't think we have running pods when we don't. However I probably came to that conclusion after writing the initial draft of this method...

sosiouxme:
aos-ci-test

sosiouxme:
@rhcarvalho @juanvallejo any further input on this? I have some further refactoring coming on top of this, and at this relatively quiet point I think I'd prefer to get it merged and manually fix any affected PRs myself, rather than keep fixing this as PRs merge.

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: For some reason I thought it would be a problem to make this a static method and then override it as an instance method. Apparently it's fine.

sosiouxme:
aos-ci-test

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -19,15 +19,26 @@ class OpenShiftCheckException(Exception):
...
-                "or module_executor (deprecated), none given")
-        self.execute_module = execute_module or module_executor
-        self.module_executor = self.execute_module
+    def __init__(self, execute_module=None, task_vars=None, tmp=None):
+        def _placeholder(*_):
+            raise RuntimeError(
+                self.__class__.__name__ +
+                " invoked execute_module without providing the method at initialization."
+            )
+        self._execute_module = execute_module or _placeholder
```
rhcarvalho: Suggestion:

```python
def __init__(...):
    self._execute_module = execute_module

...

def execute_module(...):
    ...
    if self._execute_module is None:
        raise NotImplementedError("'execute_module' was not defined at instance initialization.")
    return self._execute_module(module_name, module_args, self.tmp, self.task_vars)
```

This would keep `__init__` clean and straightforward.

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -43,13 +54,13 @@ def tags(self):
         """
         return []
 
-    @classmethod
-    def is_active(cls, task_vars):  # pylint: disable=unused-argument
+    @staticmethod
```
rhcarvalho: It should not be a `staticmethod` because typically if will need to access `task_vars` via `self`.

Even though it doesn't depend on `self` here, I'd still make it consistent across the class hierarchy -- easier for readers of the code.

sosiouxme: I would do that too; but pylint really doesn't like it. So it would be:

    def is_active(self):   # pylint: disable=unused-argument,no-self-use
          """Returns true if this check applies to the ansible-playbook run."""
          return True

To me that's more ugly than just letting it be static (which, frankly, it is at the base).

Or if we really wanted to tighten it up:

    is_active = lambda self: True

:-P

That said, I'll change it back if you'd still prefer.

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: Yeah, that does seem cleaner, thanks.

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: It sucks when the tools get in the way :-D
Makes us invent more hacks to workaround the tools:

```python
def is_active(self):
    "Some docstring"
    return self
```

or

```python
def is_active(self):
    "Some docstring"
    return self and True
```

(both aberrations that would probably "make the tool happy")

sosiouxme: return bool(self)

sosiouxme:
aos-ci-test

sosiouxme:
[merge]

sosiouxme:
looks like flake https://github.com/openshift/origin/issues/15356
[merge] again

sosiouxme:
flake https://github.com/openshift/origin/issues/15356 again

@sdodson would you mind merging this?

sdodson:
Sure, I'm chasing someone about getting that "flake" fixed.

sosiouxme:
thx

PR #4486, opened 2017-06-18
Preserve etcd3 storage if it's already in use

sdodson:
This would be the case if for instance they'd upgraded and then
migrated.

I've tested 
- [x] byo/config.yml 3.6 existing install that's using etcd3 storage already -> etcd3 storage
- [x] byo/config.yml 3.6 existing install, no etcd3 -> no etcd3 storage
- [x] byo/config.yml 3.6 clean install -> etcd3 storage
- [x] byo/config.yml 3.5 clean install -> no etcd3 storage

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -27,7 +27,17 @@
 
   - name: Set clean install fact
     set_fact:
-      l_clean_install: "{{ not master_config_stat.stat.exists }}"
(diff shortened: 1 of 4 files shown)


sdodson:
aos-ci-test

sdodson:
[merge][severity: bug]

sdodson:
Adding merge tag because this has fixes to the boolean for clean installs which another of my PRs depends on.

sdodson:
[merge][severity: bug]

PR #4489, opened 2017-06-19
Disable TLS verification in skopeo inspect

rhcarvalho:
Some registries are not configured with valid certificates and thus the
check fails with 'http: server gave HTTP response to HTTPS client'.
Since this is not fetching images, but only checking for existence,
trade security for convenience.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1461675.

**Alternatives**

- Instead of always disabling it require an extra flag to disable it.
- Force users to always use registries with valid certificates.

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/docker_image_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/docker_image_availability.py
@@ -169,7 +169,7 @@ class DockerImageAvailability(DockerHostMixin, OpenShiftCheck):
             registries = [registry]
 
         for registry in registries:
-            args = {"_raw_params": "skopeo inspect docker://{}/{}".format(registry, image)}
+            args = {"_raw_params": "skopeo inspect --tls-verify=false docker://{}/{}".format(registry, image)}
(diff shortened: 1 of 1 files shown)


rhcarvalho:
aos-ci-test

rhcarvalho:
> does the installer have flags for specifying whether the registry is "insecure"? If so maybe we can follow that config.

Good point...

```
# Docker Configuration
# Add additional, insecure, and blocked registries to global docker configuration
# For enterprise deployment types we ensure that registry.access.redhat.com is
# included if you do not include it
#openshift_docker_additional_registries=registry.example.com
#openshift_docker_insecure_registries=registry.example.com
```
https://github.com/openshift/openshift-ansible/blob/e2840e8ab50215eb7603a832f2f729d873e5a811/inventory/byo/hosts.origin.example#L72-L77

I don't know if QE has that repo in their list of insecure registries, if they don't then the "perceived problem" persists.

We would have to selectively pass the `--tls-verify=false` flag per registry, that would also complicate reasoning about bug reports if something goes wrong.

sdodson:
[merge][severity:bug]

PR #4491, opened 2017-06-19
Temporarilly only migrate jobs as we were before

sdodson:


sdodson:
@deads2k let us know when we can revert this

deads2k:
> @deads2k let us know when we can revert this

Can't say I'm a fan of intentionally breaking the queue and then trying to fix it up instead of reverting the queue breaker :(.

Here's an issue: https://github.com/openshift/origin/issues/14743

sdodson:
@deads2k This PR reverted the change that widened the storage migration from just jobs to all items. Are you saying that there's an origin change that should've been reverted instead?

deads2k:
> @deads2k This PR reverted the change that widened the storage migration from just jobs to all items. Are you saying that there's an origin change that should've been reverted instead?

The offender looked different (more lines), but if this does the same thing I'm ok with it.  Doesn't ansible have to pass upgrade tests to merge too?

sdodson:
> The offender looked different (more lines), but if this does the same thing I'm ok with it. Doesn't ansible have to pass upgrade tests to merge too?

Yes, but I manually merged the offending PR in haste.

PR #4492, opened 2017-06-19
validate data after v2->v3 etcd migration

ingvagabund:
SSIA


ingvagabund:
aos-ci-test

ingvagabund:
@smarterclayton do you think it makes sense to incorporate the validation as a command into oadm? Or would it be better to poke coreos guys and have it as a part of the `etcdctl` command?

smarterclayton:
What validation is this again?  If it's that the migrate command worked correctly, I'd probably lean on etcdctl to do that.

ingvagabund:
That all v2 keys are in v3 data store and their values in v2 are identical in v3.

michaelgugino:
Closing due to inactivity.

PR #4495, opened 2017-06-19
Skip nfs and lb groups on openshift_version

ashcrow:
I'm not thrilled with this idea but it seems to be the least invasive. Open for cleaner options.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1458653

### DIFF
--- a/roles/openshift_version/tasks/main.yml
+++ b/roles/openshift_version/tasks/main.yml
@@ -84,115 +84,119 @@
   - openshift_version is not defined
   - openshift_protect_installed_version | bool
 
-- name: Set openshift_version for rpm installation
-  include: set_version_rpm.yml
-  when: not is_containerized | bool
+# The rest of these tasks should only execute on
+# masters and nodes as we can verify they have subscriptions
+- when:
+  - inventory_hostname in groups['oo_masters_to_config'] or inventory_hostname in groups['oo_nodes_to_config']
+  block:
+  - name: Set openshift_version for rpm installation
+    include: set_version_rpm.yml
+    when: not is_containerized | bool
 
-- name: Set openshift_version for containerized installation
-  include: set_version_containerized.yml
-  when: is_containerized | bool
+  - name: Set openshift_version for containerized installation
+    include: set_version_containerized.yml
+    when: is_containerized | bool
 
-- block:
-  - name: Get available {{ openshift.common.service_type}} version
-    repoquery:
-      name: "{{ openshift.common.service_type}}"
-      ignore_excluders: true
-    register: rpm_results
-  - fail:
-      msg: "Package {{ openshift.common.service_type}} not found"
-    when: not rpm_results.results.package_found
-  - set_fact:
-      openshift_rpm_version: "{{ rpm_results.results.versions.available_versions.0 | default('0.0', True) }}"
-  - name: Fail if rpm version and docker image version are different
-    fail:
-      msg: "OCP rpm version {{ openshift_rpm_version }} is different from OCP image version {{ openshift_version }}"
-    # Both versions have the same string representation
+  - block:
+    - name: Get available {{ openshift.common.service_type}} version
+      repoquery:
+        name: "{{ openshift.common.service_type}}"
+        ignore_excluders: true
+      register: rpm_results
+    - fail:
+        msg: "Package {{ openshift.common.service_type}} not found"
+      when: not rpm_results.results.package_found
+    - set_fact:
+        openshift_rpm_version: "{{ rpm_results.results.versions.available_versions.0 | default('0.0', True) }}"
+    - name: Fail if rpm version and docker image version are different
+      fail:
+        msg: "OCP rpm version {{ openshift_rpm_version }} is different from OCP image version {{ openshift_version }}"
+      # Both versions have the same string representation
+      when:
+      - openshift_rpm_version != openshift_version
+      # if openshift_pkg_version or openshift_image_tag is defined, user gives a permission the rpm and docker image versions can differ
+      - openshift_pkg_version is not defined
+      - openshift_image_tag is not defined
     when:
-    - openshift_rpm_version != openshift_version
-    # if openshift_pkg_version or openshift_image_tag is defined, user gives a permission the rpm and docker image versions can differ
-    - openshift_pkg_version is not defined
-    - openshift_image_tag is not defined
-  when:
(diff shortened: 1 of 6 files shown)


ashcrow:
aos-ci-test

ashcrow:
Looks like a one time failure as an image wasn't able to be pulled from the registry. However, this change doesn't do anything to change images. "infra flake"

ashcrow:
aos-ci-test

ashcrow:
The health checks are doing their job. I've been seeing "infra flakes" in the last few hours and this looks to be health check finding that:

```One or more required Docker images are not available:```

Will wait a bit before issuing a retry.

**Comment by sosiouxme** [comment]
The bug shows a `fail` task in the `openshift_version` role. So excluding the NFS host from checks won't actually address this - it needs to be fixed in `openshift_version` - unless `openshift_version` wouldn't otherwise run against NFS hosts?

This might be necessary anyway though as I'm not sure what the checks do if `openshift_version` doesn't initialize the version variables.

**Comment by ashcrow** [comment]
I follow what you're saying Luke.

It looks like there are three locations which include the role:

- roles/openshift_health_checker/meta/main.yml: Noted here
- roles/openshift_common/meta/main.yml: Includes without scoping hosts
- roles/openshift_docker/meta/main.yml: Includes without scoping hosts

I'll look at trying to keep nfs and lbs out of package checking rather than skipping them within the health checks.

File: roles/openshift_version/tasks/main.yml [hunk 1]
```diff
@@ -84,115 +84,121 @@
...
-      openshift_rpm_version: "{{ rpm_results.results.versions.available_versions.0 | default('0.0', True) }}"
-  - name: Fail if rpm version and docker image version are different
-    fail:
-      msg: "OCP rpm version {{ openshift_rpm_version }} is different from OCP image version {{ openshift_version }}"
-    # Both versions have the same string representation
+# The rest of these tasks should execute only on
+# non nfs and lb hosts as these hosts may or may not
+# have subscriptions set up
+- when:
+  - inventory_hostname not in hostvars[groups.oo_nfs_to_config]
```
sdodson: maybe like this from roles/etcd_migrate/tasks/check_cluster_status.yml
  when: inventory_hostname in groups.oo_etcd_to_config[0]

ashcrow:
@sdodson / @kwoodson Looks like that one is a no go as well ...

```
The error was: error while evaluating conditional (inventory_hostname not in groups.oo_nfs_to_config[0]): 'dict object' has no attribute 'oo_nfs_to_config'
```

ashcrow:
aos-ci-test

sosiouxme:
>
>
> The error was: error while evaluating conditional (inventory_hostname not in groups.oo_nfs_to_config[0]): 'dict object' has no attribute 'oo_nfs_to_config'
>
>
At the time that the role `openshift_health_checker` operates in the
playbook (and thus pulls in `openshift_version`), I don't think those
groups have been defined yet. Possibly we should do that.

ashcrow:
@sosiouxme Good call. I pulled in the commit by @kwoodson which should move initialization.

ashcrow:
aos-ci-test

sdodson:
Shall we re-factor that playbook to live in common and pull in playbooks/common/openshift-cluster/evaluate_groups.yml?

ashcrow:
@sdodson I *think* @kwoodson's change will do the trick for us. He added a TODO so we do some refactoring in the future.

sdodson:
[merge][severity: blocker]

ashcrow:
Shouldn't be related:

```
    error: command 'gcc' failed with exit status 4
```

ashcrow:
aos-ci-test

ashcrow:
[merge][severity: blocker]

sdodson:
This has broken the health check unit tests. @sosiouxme can that be fixed?

sosiouxme:
@sdodson presumably. I'll look into what's going on here.

ashcrow:
Timed out waiting on router deploy.

```
     Play:     Create Hosted Resources
     Task:     openshift_hosted : Poll for OpenShift router deployment success
     Message:  One or more items failed
```

ashcrow:
Unrelated failure:
```
     Play:     Configure nodes
     Task:     openshift_node : Install sdn-ovs package
     Message:  |failed expects a string or unicode
```

sdodson:
aos-ci-test

ashcrow:
```
     Play:     Configure nodes
     Task:     openshift_node : Install sdn-ovs package
     Message:  |failed expects a string or unicode
```

This is likely related to items being part of two groups. I'll add functionality in to not skip if lb/nfs+node/master.

roles/openshift_version/tasks/main.yml
```diff
@@ -84,115 +84,121 @@
...
-  - name: Fail if rpm version and docker image version are different
-    fail:
-      msg: "OCP rpm version {{ openshift_rpm_version }} is different from OCP image version {{ openshift_version }}"
-    # Both versions have the same string representation
+# The rest of these tasks should execute only on
+# non nfs and lb hosts as these hosts may or may not
+# have subscriptions set up
+- when:
+  - inventory_hostname in groups['oo_masters_to_config']
+  - inventory_hostname in groups['oo_nodes_to_config']
```
sdodson: This should be an or, i think this may be tripping up the test jobs where it's not working on hosts that are only nodes.

sdodson: Looking at the failure more closely it fails on 2 out of 3 hosts in the inventory so I'm pretty sure this is what's wrong.

ashcrow:
- Switched to ``or``
- Updated comment for inclusive when
- Rearranged commits

ashcrow:
aos-ci-test

sosiouxme:
Looks like a [test flake](https://github.com/openshift/origin/issues/14871)

[merge][severity: blocker]

sosiouxme:
@sdodson mind giving this a bump... AFAICS https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/631/consoleFull#-183799125458b6e51eb7608a5981914356 was some weird build env thing.

sdodson:
[test]

sdodson:
will merge manually based on which tests fail in that one

sdodson:
flake on https://github.com/openshift/origin/issues/14871 again

PR #4497, opened 2017-06-19
oc_atomic_container: Require 1.17.2

ashcrow:
The ``oc_atomic_container`` module  requires features only available in ``atomic``
versions 1.17.2+.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1461662

### DIFF
--- a/roles/lib_openshift/library/oc_atomic_container.py
+++ b/roles/lib_openshift/library/oc_atomic_container.py
@@ -65,8 +65,11 @@ options:
 
 # -*- -*- -*- Begin included fragment: ansible/oc_atomic_container.py -*- -*- -*-
 
-# pylint: disable=wrong-import-position,too-many-branches,invalid-name
+# pylint: disable=wrong-import-position,too-many-branches,invalid-name,no-name-in-module, import-error
 import json
+
+from distutils.version import StrictVersion
+
 from ansible.module_utils.basic import AnsibleModule
 
 
@@ -191,9 +194,15 @@ def main():
         )
 
     # Verify that the platform supports atomic command
-    rc, _, err = module.run_command('atomic -v', check_rc=False)
+    rc, version_out, err = module.run_command('atomic -v', check_rc=False)
     if rc != 0:
         module.fail_json(msg="Error in running atomic command", err=err)
+    # This module requires atomic version 1.17.2 or later
+    atomic_version = StrictVersion(version_out.replace('\n', ''))
+    if atomic_version < StrictVersion('1.17.2'):
+        module.fail_json(
+            msg="atomic version 1.17.2+ is required",
+            err=str(atomic_version))
 
     try:
         core(module)
--- a/roles/lib_openshift/src/ansible/oc_atomic_container.py
+++ b/roles/lib_openshift/src/ansible/oc_atomic_container.py
@@ -1,8 +1,11 @@
 # pylint: skip-file
 # flake8: noqa
 
-# pylint: disable=wrong-import-position,too-many-branches,invalid-name
+# pylint: disable=wrong-import-position,too-many-branches,invalid-name,no-name-in-module, import-error
 import json
+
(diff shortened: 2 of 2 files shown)


ashcrow:
aos-ci-test

roles/lib_openshift/library/oc_atomic_container.py
```diff
@@ -191,9 +191,15 @@ def main():
         )
 
     # Verify that the platform supports atomic command
-    rc, _, err = module.run_command('atomic -v', check_rc=False)
+    rc, version_out, err = module.run_command('atomic -v', check_rc=False)
     if rc != 0:
         module.fail_json(msg="Error in running atomic command", err=err)
+    # This module requires atomic version 1.17.2 or later
+    version_out = version_out.replace('\n', '')
+    if version_out < '1.17.2':
```
giuseppe: should we use distutils.version to compare versions?  atomic won't probably get to 1.17.10, but in that case the string comparison will fail.

ashcrow: That's a good point. I'll switch it over.

ashcrow:
@giuseppe ⬆️

ashcrow:
Fixing ``pylint`` issues using the same workaround other ``StrictVersion`` imports use.

ashcrow:
aos-ci-test

ashcrow:
[merge]

sdodson:
[merge][severity: bug[

ashcrow:
[merge][severity: bug]

ashcrow:
😕 

```
Errors from dependency resolution:
  glibc-2.17-157.el7_3.2.i686 requires glibc-common = 2.17-157.el7_3.2
  glibc-devel-2.17-157.el7_3.2.i686 requires glibc-headers = 2.17-157.el7_3.2
You should resolve these issues before proceeding with an install.
You may need to remove or downgrade packages or enable/disable yum repositories.
```

giuseppe:
by the way, should the version 1.17.1 be fine?  It has --package-build or is there anything else we are missing?
```bash
[cloud-user@atomic-0011 ~]$ atomic --version
1.17.1
[cloud-user@atomic-0011 ~]$ atomic install --help
usage: atomic install [-h] [--storage STORAGE] [-n NAME] [--display]
                      [--user | --system] [--system-package SYSTEM_PACKAGE]
                      [--rootfs REMOTE] [--set SETVALUES]
                      image ...
```

ashcrow:
@giuseppe I'll double check. Some fixes were backported into 1.17.2 which were system container related....

ashcrow:
We need 1.17.2. Double check this for me though as I think ``atomic`` gave you the wrong version 😆 

```
# atomic --version
1.17.1
# rpm -qa | grep atomic-1.
atomic-1.17.2-4
```

ashcrow:
Known infra flake on merge

```
Errors from dependency resolution:
                 glibc-2.17-157.el7_3.2.i686 requires glibc-common = 2.17-157.el7_3.2
                 glibc-devel-2.17-157.el7_3.2.i686 requires glibc-headers = 2.17-157.el7_3.2
```

giuseppe:
yeah!

$ rpm -q atomic
atomic-1.17.2-4.git2760e30.el7.x86_64


Nevermind my comment. @ingvagabund, it seems we can use the fake 1.17.1 :)

ashcrow:
[merge][severity: blocker]

ashcrow:
@sdodson What should I do in this case. It's not an openshift origin flake, but it is a flake.

sdodson:
[merge][severity: blocker]

sdodson:
not a flake but test infra problem, re merging

ashcrow:
Can't find a flake for
```
Error response from daemon: You cannot remove a running container...
```

ashcrow:
[merge][severity: blocker]

PR #4504, opened 2017-06-20
detect etcd service name based on etcd runtime when restarting

ingvagabund:
Bug: 1461664


ingvagabund:
aos-ci-test

giuseppe:
I've tried it on AH and it solves the problem for me

sdodson:
aos-ci-test

PR #4509, opened 2017-06-20
Create ansible role for deploying prometheus on openshift

zgalor:
A new role for installing prometheus on openshift.
Depends on `openshift_hosted_prometheus_deploy` flag

role creates:
- prometheus namespace
- prometheus clusterrolebinding and service account
- pvs for prometheus, alertmanager and alertbuffer for internal nfs
- prometheus pod with prometheus behind oauth-proxy, alertmanager and alert-buffer behind oauth-proxy
- prometheus and alertmanager configmaps
- prometheus and alerts services and direct routes
- prometheus, alertmanager and alert-buffer pvcs

Deployment diagram:
prometheus deployment on openshift

role can be run directly using `playbooks/byo/openshift-cluster/openshift-prometheus.yml`

depends on https://github.com/openshift/openshift-ansible/pull/4887

### DIFF
--- a/playbooks/common/openshift-cluster/openshift_prometheus.yml
+++ b/playbooks/common/openshift-cluster/openshift_prometheus.yml
@@ -1,0 +1,9 @@
+---
+- include: std_include.yml
+
+- name: OpenShift Prometheus
+  hosts: oo_first_master
+  roles:
+  - openshift_prometheus
+  vars:
+    openshift_prometheus_state: present
--- a/roles/openshift_prometheus/README.md
+++ b/roles/openshift_prometheus/README.md
@@ -1,0 +1,95 @@
+OpenShift Prometheus
+====================
+
+OpenShift Prometheus Installation
+
+Requirements
+------------
+
+
+Role Variables
+--------------
+
+For default values, see [`defaults/main.yaml`](defaults/main.yaml).
+
+- `openshift_prometheus_state`: present - install/update. absent - uninstall.
+
+- `openshift_prometheus_namespace`: project (i.e. namespace) where the components will be
+  deployed.
+
+- `openshift_prometheus_replicas`: The number of replicas for prometheus deployment.
+
+- `openshift_prometheus_node_selector`: Selector for the nodes prometheus will be deployed on.
+
+- `openshift_prometheus_image_<COMPONENT>`: specify image for the component 
+
+## Storage related variables
+Each prometheus component (prometheus, alertmanager, alert-buffer, oauth-proxy) can set pv claim by setting corresponding role variable:
+```
+openshift_prometheus_<COMPONENT>_storage_type: <VALUE>
+openshift_prometheus_<COMPONENT>_pvc_(name|size|access_modes|pv_selector): <VALUE>
+```
+e.g
+```
+openshift_prometheus_storage_type: pvc
+openshift_prometheus_alertmanager_pvc_name: alertmanager
+openshift_prometheus_alertbuffer_pvc_size: 10G
+openshift_prometheus_pvc_access_modes: [ReadWriteOnce]
+```
+
+## Additional Alert Rules file variable
+An external file with alert rules can be added by setting path to additional rules variable: 
+```
+openshift_prometheus_additional_rules_file: <PATH> 
+```
+
+File content should be in prometheus alert rules format.
+Following example sets rule to fire an alert when one of the cluster nodes is down:
+
+```
+groups:
+- name: example-rules
+  interval: 30s # defaults to global interval
+  rules:
+  - alert: Node Down
+    expr: up{job="kubernetes-nodes"} == 0
+    annotations:
+      miqTarget: "ContainerNode"
+      severity: "HIGH"
+      message: "{{ '{{' }}{{ '$labels.instance' }}{{ '}}' }} is down"
+```
+
+
+## Additional variables to control resource limits
+Each prometheus component (prometheus, alertmanager, alert-buffer, oauth-proxy) can specify a cpu and memory limits and requests by setting
+the corresponding role variable:
+```
+openshift_prometheus_<COMPONENT>_(limits|requests)_(memory|cpu): <VALUE>
+```
+e.g
+```
+openshift_prometheus_alertmanager_limits_memory: 1Gi
+openshift_prometheus_oath_proxy_requests_cpu: 100
+```
+
+Dependencies
+------------
+
+openshift_facts
+
+
+Example Playbook
+----------------
+
+```
+- name: Configure openshift-prometheus
+  hosts: oo_first_master
+  roles:
+  - role: openshift_prometheus
+```
+
+License
+-------
+
+Apache License, Version 2.0
+
--- a/roles/openshift_prometheus/defaults/main.yaml
+++ b/roles/openshift_prometheus/defaults/main.yaml
@@ -1,0 +1,74 @@
+---
+# defaults file for openshift_prometheus
+openshift_prometheus_state: present
+
+openshift_prometheus_namespace: prometheus
+
+openshift_prometheus_replicas: 1
+openshift_prometheus_node_selector: {"region":"infra"}
+
+# images
+openshift_prometheus_image_proxy: "openshift/oauth-proxy:v1.0.0"
+openshift_prometheus_image_prometheus: "openshift/prometheus:v2.0.0-dev"
+openshift_prometheus_image_alertmanager: "openshift/prometheus-alertmanager:dev"
+openshift_prometheus_image_alertbuffer: "ilackarms/message-buffer"
+
+# additional prometheus rules file
+openshift_prometheus_additional_rules_file: null
+
+# All the required exports
+openshift_prometheus_pv_exports:
+  - prometheus
+  - prometheus-alertmanager
+  - prometheus-alertbuffer
+# PV template files and their created object names
+openshift_prometheus_pv_data:
+  - pv_name: prometheus
+    pv_template: prom-pv-server.yml
+    pv_label: Prometheus Server PV
+  - pv_name: prometheus-alertmanager
+    pv_template: prom-pv-alertmanager.yml
+    pv_label: Prometheus Alertmanager PV
+  - pv_name: prometheus-alertbuffer
+    pv_template: prom-pv-alertbuffer.yml
+    pv_label: Prometheus Alert Buffer PV
+
+# Hostname/IP of the NFS server. Currently defaults to first master
+openshift_prometheus_nfs_server: "{{ groups.nfs.0 }}"
+
+# storage
+openshift_prometheus_storage_type: pvc
+openshift_prometheus_pvc_name: prometheus
+openshift_prometheus_pvc_size: 10G
+openshift_prometheus_pvc_access_modes: [ReadWriteOnce]
+openshift_prometheus_pvc_pv_selector: {}
+
+openshift_prometheus_alertmanager_storage_type: pvc
+openshift_prometheus_alertmanager_pvc_name: prometheus-alertmanager
+openshift_prometheus_alertmanager_pvc_size: 10G
+openshift_prometheus_alertmanager_pvc_access_modes: [ReadWriteOnce]
+openshift_prometheus_alertmanager_pvc_pv_selector: {}
+
+openshift_prometheus_alertbuffer_storage_type: pvc
+openshift_prometheus_alertbuffer_pvc_name: prometheus-alertbuffer
+openshift_prometheus_alertbuffer_pvc_size: 10G
+openshift_prometheus_alertbuffer_pvc_access_modes: [ReadWriteOnce]
+openshift_prometheus_alertbuffer_pvc_pv_selector: {}
+
+# container resources
+openshift_prometheus_cpu_limit: null
+openshift_prometheus_memory_limit: null
+openshift_prometheus_cpu_requests: null
+openshift_prometheus_memory_requests: null
+openshift_prometheus_alertmanager_cpu_limit: null
+openshift_prometheus_alertmanager_memory_limit: null
+openshift_prometheus_alertmanager_cpu_requests: null
+openshift_prometheus_alertmanager_memory_requests: null
+openshift_prometheus_alertbuffer_cpu_limit: null
+openshift_prometheus_alertbuffer_memory_limit: null
+openshift_prometheus_alertbuffer_cpu_requests: null
+openshift_prometheus_alertbuffer_memory_requests: null
+openshift_prometheus_oauth_proxy_cpu_limit: null
+openshift_prometheus_oauth_proxy_memory_limit: null
+openshift_prometheus_oauth_proxy_cpu_requests: null
+openshift_prometheus_oauth_proxy_memory_requests: null
--- a/roles/openshift_prometheus/meta/main.yaml
+++ b/roles/openshift_prometheus/meta/main.yaml
@@ -1,0 +1,19 @@
+---
+galaxy_info:
+  author: OpenShift Development <dev@lists.openshift.redhat.com>
+  description: Deploy OpenShift prometheus integration for the cluster
+  company: Red Hat, Inc.
+  license: license (Apache)
+  min_ansible_version: 2.2
+  platforms:
+  - name: EL
+    versions:
+    - 7
+  - name: Fedora
+    versions:
+    - all
+  categories:
+  - openshift
+dependencies:
+- { role: lib_openshift }
+- { role: openshift_facts }
--- a/roles/openshift_prometheus/tasks/install_prometheus.yaml
+++ b/roles/openshift_prometheus/tasks/install_prometheus.yaml
@@ -1,0 +1,241 @@
+---
+
+# namespace
+- name: Add prometheus project
+  oc_project:
+    state: "{{ state }}"
+    name: "{{ openshift_prometheus_namespace }}"
+    node_selector: "{{ openshift_prometheus_node_selector | oo_selector_to_string_list() }}"
+    description: Prometheus
+
+# secrets
+- name: Set alert and prometheus secrets
+  oc_secret:
+    state: "{{ state }}"
+    name: "{{ item }}-proxy"
+    namespace: "{{ openshift_prometheus_namespace }}"
+    contents:
+      - path: session_secret
+        data: "{{ 43 | oo_random_word }}="
+  with_items:
+    - prometheus
+    - alerts
+
+# serviceaccount
+- name: create prometheus serviceaccount
+  oc_serviceaccount:
+    state: "{{ state }}"
+    name: prometheus
+    namespace: "{{ openshift_prometheus_namespace }}"
+    #    TODO add annotations when supproted
+    #    annotations:
+    #      serviceaccounts.openshift.io/oauth-redirectreference.prom: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"prometheus"}}'
+    #      serviceaccounts.openshift.io/oauth-redirectreference.alerts: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"alerts"}}'
+
+    secrets:
+      - prometheus-secrets
+  changed_when: no
+
+# TODO remove this when annotations are supported by oc_serviceaccount
+- name: annotate serviceaccount
+  command: >
+    {{ openshift.common.client_binary }} annotate --overwrite -n {{ openshift_prometheus_namespace }}
+    serviceaccount prometheus
+    serviceaccounts.openshift.io/oauth-redirectreference.prom='{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"prometheus"}}'
+    serviceaccounts.openshift.io/oauth-redirectreference.alerts='{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"alerts"}}'
+
+
+# create clusterrolebinding for prometheus serviceaccount
+- name: Set cluster-reader permissions for prometheus
+  oc_adm_policy_user:
+    state: "{{ state }}"
+    namespace: "{{ openshift_prometheus_namespace }}"
+    resource_kind: cluster-role
+    resource_name: cluster-reader
+    user: "system:serviceaccount:{{ openshift_prometheus_namespace }}:prometheus"
+
+
+######################################################################
+# NFS
+# In the case that we are not running on a cloud provider, volumes must be statically provisioned
+
+- include: nfs.yaml
+  when: not (openshift_cloudprovider_kind is defined and (openshift_cloudprovider_kind == 'aws' or openshift_cloudprovider_kind == 'gce'))
+
+
+# create prometheus and alerts services
+# TODO join into 1 task with loop
+- name: Create prometheus service
+  oc_service:
+    state: "{{ state }}"
+    name: "{{ item.name }}"
+    namespace: "{{ openshift_prometheus_namespace }}"
+    selector:
+      app: prometheus
+    labels:
+      name: "{{ item.name }}"
+      #    TODO add annotations when supported
+      #    annotations:
+      #      service.alpha.openshift.io/serving-cert-secret-name: "{{item.name}}-tls"
+    ports:
+      - port: 443
+        targetPort: 8443
+  with_items:
+    - name: prometheus
+
+- name: Create alerts service
+  oc_service:
+    state: "{{ state }}"
+    name: "{{ item.name }}"
+    namespace: "{{ openshift_prometheus_namespace }}"
+    selector:
+      app: prometheus
+    labels:
+      name: "{{ item.name }}"
+      #    TODO add annotations when supported
+      #    annotations:
+      #      service.alpha.openshift.io/serving-cert-secret-name: "{{item.name}}-tls"
+    ports:
+      - port: 443
+        targetPort: 9443
+  with_items:
+    - name: alerts
+
+
+# Annotate services with secret name
+# TODO remove this when annotations are supported by oc_service
+- name: annotate prometheus service
+  command: >
+    {{ openshift.common.client_binary }} annotate --overwrite -n {{ openshift_prometheus_namespace }}
+    service prometheus 'service.alpha.openshift.io/serving-cert-secret-name=prometheus-tls'
+
+- name: annotate alerts service
+  command: >
+    {{ openshift.common.client_binary }} annotate --overwrite -n {{ openshift_prometheus_namespace }}
+    service alerts 'service.alpha.openshift.io/serving-cert-secret-name=prometheus-alerts-tls'
+
+# create prometheus and alerts routes
+- name: create prometheus and alerts routes
+  oc_route:
+    state: "{{ state }}"
+    name: "{{ item.name }}"
+    namespace: "{{ openshift_prometheus_namespace }}"
+    service_name: "{{ item.name }}"
+    tls_termination: reencrypt
+  with_items:
+    - name: prometheus
+    - name: alerts
+
+# Storage
+- name: create prometheus pvc
+  oc_pvc:
+    namespace: "{{ openshift_prometheus_namespace }}"
+    name: "{{ openshift_prometheus_pvc_name }}"
+    access_modes: "{{ openshift_prometheus_pvc_access_modes }}"
+    volume_capacity: "{{ openshift_prometheus_pvc_size }}"
+    selector: "{{ openshift_prometheus_pvc_pv_selector }}"
+
+- name: create alertmanager pvc
+  oc_pvc:
+    namespace: "{{ openshift_prometheus_namespace }}"
+    name: "{{ openshift_prometheus_alertmanager_pvc_name }}"
+    access_modes: "{{ openshift_prometheus_alertmanager_pvc_access_modes }}"
+    volume_capacity: "{{ openshift_prometheus_alertmanager_pvc_size }}"
+    selector: "{{ openshift_prometheus_alertmanager_pvc_pv_selector }}"
+
+- name: create alertbuffer pvc
+  oc_pvc:
+    namespace: "{{ openshift_prometheus_namespace }}"
+    name: "{{ openshift_prometheus_alertbuffer_pvc_name }}"
+    access_modes: "{{ openshift_prometheus_alertbuffer_pvc_access_modes }}"
+    volume_capacity: "{{ openshift_prometheus_alertbuffer_pvc_size }}"
+    selector: "{{ openshift_prometheus_alertbuffer_pvc_pv_selector }}"
+
+# create prometheus deployment
+- name: Set prometheus deployment template
+  template:
+    src: prometheus_deployment.j2
+    dest: "{{ tempdir }}/templates/prometheus.yaml"
+  vars:
+    namespace: "{{ openshift_prometheus_namespace }}"
+    prom_replicas: "{{ openshift_prometheus_replicas }}"
+
+- name: Set prometheus deployment
+  oc_obj:
+    state: "{{ state }}"
+    name: "prometheus"
+    namespace: "{{ openshift_prometheus_namespace }}"
+    kind: deployment
+    files:
+      - "{{ tempdir }}/templates/prometheus.yaml"
+    delete_after: true
+
+# prometheus configmap
+# Copy the additional rules file if it is defined
+- name: Copy additional rules file to host
+  copy:
+    src: "{{ openshift_prometheus_additional_rules_file }}"
+    dest: "{{ tempdir }}/prometheus.additional.rules"
+  when:
+    - openshift_prometheus_additional_rules_file is defined
+    - openshift_prometheus_additional_rules_file is not none
+    - openshift_prometheus_additional_rules_file | trim | length > 0
+
+- stat:
+    path: "{{ tempdir }}/prometheus.additional.rules"
+  register: additional_rules_stat
+
+# The kubernetes version impacts the prometheus scraping endpoint
+# so gathering it before constructing the configmap
+- name: get oc version
+  oc_version:
+  register: oc_version
+
+- set_fact:
+    kubernetes_version: "{{ oc_version.results.kubernetes_short | float }}"
+
+- template:
+    src: prometheus.yml.j2
+    dest: "{{ tempdir }}/prometheus.yml"
+  changed_when: no
+
+- template:
+    src: prometheus.rules.j2
+    dest: "{{ tempdir }}/prometheus.rules"
+  changed_when: no
+
+# In prometheus configmap create "additional.rules" section if file exists
+- name: Set prometheus configmap
+  oc_configmap:
+    state: "{{ state }}"
+    name: "prometheus"
+    namespace: "{{ openshift_prometheus_namespace }}"
+    from_file:
+      prometheus.rules: "{{ tempdir }}/prometheus.rules"
+      prometheus.additional.rules: "{{ tempdir }}/prometheus.additional.rules"
+      prometheus.yml: "{{ tempdir }}/prometheus.yml"
+  when: additional_rules_stat.stat.exists == True
+
+- name: Set prometheus configmap
+  oc_configmap:
+    state: "{{ state }}"
+    name: "prometheus"
+    namespace: "{{ openshift_prometheus_namespace }}"
+    from_file:
+      prometheus.rules: "{{ tempdir }}/prometheus.rules"
+      prometheus.yml: "{{ tempdir }}/prometheus.yml"
+  when: additional_rules_stat.stat.exists == False
+
+# alertmanager configmap
+- template:
+    src: alertmanager.yml.j2
+    dest: "{{ tempdir }}/alertmanager.yml"
+  changed_when: no
+
+- name: Set alertmanager configmap
+  oc_configmap:
+    state: "{{ state }}"
+    name: "prometheus-alerts"
+    namespace: "{{ openshift_prometheus_namespace }}"
+    from_file:
+      alertmanager.yml: "{{ tempdir }}/alertmanager.yml"
--- a/roles/openshift_prometheus/tasks/main.yaml
+++ b/roles/openshift_prometheus/tasks/main.yaml
@@ -1,0 +1,26 @@
+---
+
+- name: Create temp directory for doing work in on target
+  command: mktemp -td openshift-prometheus-ansible-XXXXXX
+  register: mktemp
+  changed_when: False
+
+- set_fact:
+    tempdir: "{{ mktemp.stdout }}"
+
+- name: Create templates subdirectory
+  file:
+    state: directory
+    path: "{{ tempdir }}/templates"
+    mode: 0755
+  changed_when: False
+
+- include: install_prometheus.yaml
+  vars:
+    state: "{{ openshift_prometheus_state }}"
+
+- name: Delete temp directory
+  file:
+    name: "{{ tempdir }}"
+    state: absent
+  changed_when: False
--- a/roles/openshift_prometheus/templates/alertmanager.yml.j2
+++ b/roles/openshift_prometheus/templates/alertmanager.yml.j2
@@ -1,0 +1,20 @@
+global:
+
+# The root route on which each incoming alert enters.
+route:
+  # default route if none match
+  receiver: alert-buffer-wh
+
+  # The labels by which incoming alerts are grouped together. For example,
+  # multiple alerts coming in for cluster=A and alertname=LatencyHigh would
+  # be batched into a single group.
+  # TODO:
+  group_by: []
+
+  # All the above attributes are inherited by all child routes and can
+  # overwritten on each.
+
+receivers:
+- name: alert-buffer-wh
+  webhook_configs:
+  - url: http://localhost:9099/topics/alerts
--- a/roles/openshift_prometheus/templates/prometheus.rules.j2
+++ b/roles/openshift_prometheus/templates/prometheus.rules.j2
@@ -1,0 +1,4 @@
+groups:
+- name: example-rules
+  interval: 30s # defaults to global interval
+  rules:
--- a/roles/openshift_prometheus/tests/test.yaml
+++ b/roles/openshift_prometheus/tests/test.yaml
@@ -1,0 +1,5 @@
+---
+- hosts: localhost
+  remote_user: root
+  roles:
+    - openshift_prometheus
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -1024,6 +1024,18 @@ def oo_contains_rule(source, apiGroups, resources, verbs):
     return False
 
 
+def oo_selector_to_string_list(user_dict):
+    """Convert a dict of selectors to a key=value list of strings
+
+Given input of {'region': 'infra', 'zone': 'primary'} returns a list
+of items as ['region=infra', 'zone=primary']
+    """
+    selectors = []
+    for key in user_dict:
+        selectors.append("{}={}".format(key, user_dict[key]))
+    return selectors
+
+
 class FilterModule(object):
     """ Custom ansible filter mapping """
 
@@ -1065,5 +1077,6 @@ class FilterModule(object):
             "oo_openshift_loadbalancer_backends": oo_openshift_loadbalancer_backends,
             "to_padded_yaml": to_padded_yaml,
             "oo_random_word": oo_random_word,
-            "oo_contains_rule": oo_contains_rule
+            "oo_contains_rule": oo_contains_rule,
+            "oo_selector_to_string_list": oo_selector_to_string_list
         }
--- a/playbooks/byo/openshift-cluster/openshift-prometheus.yml
+++ b/playbooks/byo/openshift-cluster/openshift-prometheus.yml
@@ -1,0 +1,4 @@
+---
+- include: initialize_groups.yml
+
+- include: ../../common/openshift-cluster/openshift_prometheus.yml
--- a/playbooks/common/openshift-cluster/openshift_hosted.yml
+++ b/playbooks/common/openshift-cluster/openshift_hosted.yml
@@ -49,6 +49,9 @@
   - role: cockpit-ui
     when: ( openshift.common.version_gte_3_3_or_1_3  | bool ) and ( openshift_hosted_manage_registry | default(true) | bool ) and not (openshift.docker.hosted_registry_insecure | default(false) | bool)
 
+  - role: openshift_prometheus
+    when: openshift_hosted_prometheus_deploy | default(false) | bool
+
 - name: Update master-config for publicLoggingURL
   hosts: oo_masters_to_config:!oo_first_master
   tags:
--- a/roles/openshift_prometheus/files/openshift_prometheus.exports
+++ b/roles/openshift_prometheus/files/openshift_prometheus.exports
@@ -1,0 +1,3 @@
+/exports/prometheus *(rw,no_root_squash,no_wdelay)
+/exports/prometheus-alertmanager *(rw,no_root_squash,no_wdelay)
+/exports/prometheus-alertbuffer *(rw,no_root_squash,no_wdelay)
--- a/roles/openshift_prometheus/tasks/create_pvs.yaml
+++ b/roles/openshift_prometheus/tasks/create_pvs.yaml
@@ -1,0 +1,36 @@
+---
+# Check for existance and then conditionally:
+# - evaluate templates
+# - PVs
+#
+# These tasks idempotently create required Prometheus PV objects. Do not
+# call this file directly. This file is intended to be ran as an
+# include that has a 'with_items' attached to it. Hence the use below
+# of variables like "{{ item.pv_label }}"
+
+- name: "Check if the {{ item.pv_label }} template has been created already"
+  oc_obj:
+    namespace: "{{ openshift_prometheus_namespace }}"
+    state: list
+    kind: pv
+    name: "{{ item.pv_name }}"
+  register: prom_pv_check
+
+# Skip all of this if the PV already exists
+- block:
+    - name: "Ensure the {{ item.pv_label }} template is evaluated"
+      template:
+        src: "{{ item.pv_template }}.j2"
+        dest: "{{ tempdir }}/templates/{{ item.pv_template }}"
+
+    - name: "Ensure {{ item.pv_label }} is created"
+      oc_obj:
+        namespace: "{{ openshift_prometheus_namespace }}"
+        kind: pv
+        name: "{{ item.pv_name }}"
+        state: present
+        delete_after: True
+        files:
+          - "{{ tempdir }}/templates/{{ item.pv_template }}"
+  when:
+    - not prom_pv_check.results.results.0
--- a/roles/openshift_prometheus/tasks/nfs.yaml
+++ b/roles/openshift_prometheus/tasks/nfs.yaml
@@ -1,0 +1,44 @@
+---
+# Tasks to statically provision NFS volumes
+# Include if not using dynamic volume provisioning
+- name: Ensure the /exports/ directory exists
+  file:
+    path: /exports/
+    state: directory
+    mode: 0755
+    owner: root
+    group: root
+
+- name: Ensure the prom-pv0X export directories exist
+  file:
+    path: "/exports/{{ item }}"
+    state: directory
+    mode: 0777
+    owner: nfsnobody
+    group: nfsnobody
+  with_items: "{{ openshift_prometheus_pv_exports }}"
+
+- name: Ensure the NFS exports for Prometheus PVs exist
+  copy:
+    src: openshift_prometheus.exports
+    dest: /etc/exports.d/openshift_prometheus.exports
+  register: nfs_exports_updated
+
+- name: Ensure the NFS export table is refreshed if exports were added
+  command: exportfs -ar
+  when:
+    - nfs_exports_updated.changed
+
+
+######################################################################
+# Create the required Prometheus PVs. Check out these online docs if you
+# need a refresher on includes looping with items:
+# * http://docs.ansible.com/ansible/playbooks_loops.html#loops-and-includes-in-2-0
+# * http://stackoverflow.com/a/35128533
+#
+# TODO: Handle the case where a PV template is updated in
+# openshift-ansible and the change needs to be landed on the managed
+# cluster.
+
+- include: create_pvs.yaml
+  with_items: "{{ openshift_prometheus_pv_data }}"
--- a/roles/openshift_prometheus/templates/prom-pv-alertbuffer.yml.j2
+++ b/roles/openshift_prometheus/templates/prom-pv-alertbuffer.yml.j2
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: prometheus-alertbuffer
+  labels:
+    storage: prometheus-alertbuffer
+spec:
+  capacity:
+    storage: 15Gi
+  accessModes:
+    - ReadWriteOnce
+  nfs:
+    path: /exports/prometheus-alertbuffer
+    server: {{ openshift_prometheus_nfs_server }}
+  persistentVolumeReclaimPolicy: Retain
--- a/roles/openshift_prometheus/templates/prom-pv-alertmanager.yml.j2
+++ b/roles/openshift_prometheus/templates/prom-pv-alertmanager.yml.j2
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: prometheus-alertmanager
+  labels:
+    storage: prometheus-alertmanager
+spec:
+  capacity:
+    storage: 15Gi
+  accessModes:
+    - ReadWriteOnce
+  nfs:
+    path: /exports/prometheus-alertmanager
+    server: {{ openshift_prometheus_nfs_server }}
+  persistentVolumeReclaimPolicy: Retain
--- a/roles/openshift_prometheus/templates/prom-pv-server.yml.j2
+++ b/roles/openshift_prometheus/templates/prom-pv-server.yml.j2
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: prometheus
+  labels:
+    storage: prometheus
+spec:
+  capacity:
+    storage: 15Gi
+  accessModes:
+    - ReadWriteOnce
+  nfs:
+    path: /exports/prometheus
+    server: {{ openshift_prometheus_nfs_server }}
+  persistentVolumeReclaimPolicy: Retain
--- a/roles/openshift_prometheus/templates/prometheus.yml.j2
+++ b/roles/openshift_prometheus/templates/prometheus.yml.j2
@@ -1,0 +1,174 @@
+rule_files:
+  - 'prometheus.rules'
+{% if openshift_prometheus_additional_rules_file is defined and openshift_prometheus_additional_rules_file is not none %}
+  - 'prometheus.additional.rules'
+{% endif %}
+
+
+
+# A scrape configuration for running Prometheus on a Kubernetes cluster.
+# This uses separate scrape configs for cluster components (i.e. API server, node)
+# and services to allow each to use different authentication configs.
+#
+# Kubernetes labels will be added as Prometheus labels on metrics via the
+# `labelmap` relabeling action.
+
+# Scrape config for API servers.
+#
+# Kubernetes exposes API servers as endpoints to the default/kubernetes
+# service so this uses `endpoints` role and uses relabelling to only keep
+# the endpoints associated with the default/kubernetes service using the
+# default named port `https`. This works for single API server deployments as
+# well as HA API server deployments.
+scrape_configs:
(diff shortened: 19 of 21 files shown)


zgalor:
@jcantrill Just a kickoff WIP PR. Please review the role name for start

jcantrill:
@zgalor Looks like a good beginning to me.  The only bit I want to reiterate from our conversation is if there is any need to possibly have to install something like alert manager separately, you may wish to add a second role from that so we could modify, update, or deploy event manager without modifying the prometheus deployment.

playbooks/byo/generate_prometheus.yaml
```diff
@@ -0,0 +1,5 @@
+---
```
zgalor: @jcantrill I didn't know where to put this file in the tree. I just use it to run the rule. It might be obsolete in the future?

jcantrill: I believe this location is fine, but the naming to be consistent should use dashes instead of underscore. You may even wish to name it 'openshift-prometheus' to follow the logging and metrics pattern

roles/openshift_prometheus/tasks/main.yaml
```diff
@@ -1,2 +1,37 @@
...
+    description: Prometheus
+
+- name: generate prometheus deployment
+  template:
+    src: prometheus.j2
+    dest: "{{ prometheus_template_path }}"
+  vars:
+    namespace: "{{openshift_prometheus_namespace}}"
+    image_proxy: "openshift/oauth-proxy:v1.0.0"
+    image_prometheus: "openshift/prometheus:v2.0.0-dev"
```
jcantrill: Should this be a var in defaults/main.yaml?

roles/openshift_prometheus/tasks/main.yaml
```diff
@@ -1,2 +1,37 @@
...
+    image_alertmanager: "prom/alertmanager"
+    image_message_buffer: "ilackarms/message-buffer"
+  register: result
+
+- debug: var=result
+
+- name: Deploying template
+  command: >
+    {{ openshift.common.client_binary }} --config={{ kubeconfig }}
+    new-app -f {{ prometheus_template_path }}
```
jcantrill: @ewolinetz is there a 'new-app' task?

roles/openshift_prometheus/vars/main.yaml
```diff
@@ -1,2 +1,4 @@
 ---
 # vars file for openshift_prometheus
+openshift_prometheus_namespace: prometheus
```
jcantrill: Is the intention to allow users to override these?  If so then they should be in main/default.yaml

jcantrill: Probably should be something like 'openshift-prometheus' or something... @smarterclayton ??

roles/openshift_prometheus/tasks/main.yaml
```diff
@@ -1,2 +1,37 @@
...
+- name: Add prometheus project
+  oc_project:
+    state: present
+    name: "{{ openshift_prometheus_namespace }}"
+    description: Prometheus
+
+- name: generate prometheus deployment
+  template:
+    src: prometheus.j2
+    dest: "{{ prometheus_template_path }}"
```
jcantrill: I do not know if this is required, but in logging we generate the deployments to a tmp folder generated everytime and cleaned up after ansible runs.

zgalor:
@twiest @joelddiaz this is still WIP but this initial version works, if you want to start looking into it

zgalor:
@jcantrill @sdodson can you please review?

roles/openshift_prometheus/defaults/main.yaml
```diff
@@ -1,3 +1,11 @@
 ---
 # defaults file for openshift_prometheus
+openshift_prometheus_install_prometheus: True
+
 kubeconfig: "{{ openshift.common.config_base }}/master/openshift-master.kubeconfig"
+openshift_prometheus_namespace: prometheus
```
jcantrill: we should confirm with @smarterclayton or @twiest what this should be named.  There was talk at one time of having infra namespaces like 'openshift-infra-prometheus'

roles/openshift_prometheus/meta/main.yaml
```diff
@@ -1,5 +1,5 @@
 galaxy_info:
-  author: OpenShift Development <dev@lists.openshift.redhat.com>
+  author: Zohar Galor
```
jcantrill: Wouldn't you prefer this ref the dev list so you are not the sole keeper to the kingdom?

roles/openshift_prometheus/tasks/oc_apply.yaml
```diff
@@ -0,0 +1,32 @@
+---
+- name: Checking generation of {{file_content.kind}} {{file_content.metadata.name}}
```
jcantrill: @ewolinetz Isn't there a wrapper to 'oc apply' now which replaces the need for a special task?

jcantrill:
@zgalor In general it looks like you are making good progress to a role that mirrors logging in a number of ways.  Few questions I have that maybe @smarterclayton can provide additional insight:

1. Can you accomplish the generation of all your resources with a template?
1. Do you have challenging orchestration problems that can not be accomplished with a template?

The logging role is a port of the original logging deployer image.  The role does all the things it does because of the above referenced reasons.  At one point in the evolution of the platform, templates were not an option for logging because certain resources needed to exist before others, or we needed to have things like certs generated and stuffed into a configmap.  If you do not have any of these requirements it may simplify your ansible role greatly

roles/openshift_prometheus/meta/main.yaml
zgalor: :+1:

zgalor:
@jcantrill initially the role constructed all the instruments from a single template, but that posed a problem with updating existing items.
The template displays errors and doesn't update existing items.

Also, the ConfigMap that configures prometheus should allow extension by additional user configurations to add alert rules.

Last, if we want to separate alerts from prometheus core in the future, it will be much easier with the existing division to sub-components.

roles/openshift_prometheus/tasks/oc_apply.yaml
ewolinetz: certain api objects have their own modules and would replace this, yes. You should be able to see examples in the `openshift_logging_<component>` roles

zgalor: @ewolinetz As far as I can tell from openshift logging project components, there are now oc modules for:
serviceaccount, secret, service & configmap.
Other objects (route, deployment, rolebinding) should be created using oc_obj module instead of running "oc apply" with command module. is that correct?

zgalor: I also found `oc_route` module, but I noticed that the existing modules don't have `annotations` parameter, which I can currently set with a template. 
Is this planned? Can I add this ability or is someone already on it? is there a different way to add annotations? I need this on `oc_service` and `oc_serviceaccount` modules
@kwoodson @ewolinetz

joelddiaz:
@zgalor i ran through the role today, and it does look like prometheus is up and running on my cluster. I do have some specific feedback:

The pods from the prometheus namespace are running on non-infra nodes.
The pods have no requests/limits values set.

More general feedback would be: i see a number of ansible templates generating openshift objects, then creating/applying them. A number of these openshift objects should be able to be created with some of the ansible modules available in lib_openshift.  Also, how much of these ansible templates can be used as openshift templates where the install process is more like 1) install template into namespace 2) process template by passing the role variables as parameters to the template

zgalor:
@joelddiaz thanks for the feedback.
1. I am adding nodeselector and cpu, memory limit params. what do you mean by requests?
2. I am currently refactoring everything to use as much oc modules as possible. and also moving deployment to a single pod like in this template: https://github.com/openshift/origin/pull/15078
3. For the rest (like Deployment) I am not sure why working with openshift template and processing it is better than using oc_obj. If I understand correctly, template processing displays errors when trying to create existing objects, and also doesn't update existing configuration (Which we will surely need with the alert rules configuration)

joelddiaz:
@zgalor the requests parameter ties in with the limits parameter. requests are what the OpenShift scheduler uses to decide if a node has capacity to run a pod, and the limit is the hard cgroup limit placed on the pod/container.  The difference between the request and limit values determines the amount that the node resources are oversubscribed. https://docs.openshift.com/container-platform/3.5/dev_guide/compute_resources.html#dev-compute-resources

I only mention the OpenShift templates because they are useful by themselves without having to use Ansible, but if they are cumbersome to use with the automation, the oc_obj is definitely the answer.

zgalor:
Code has been refactored: 

- Install all containers in one pod

- add requests and limits, node selection

- create objects using oc_ modules (service, secert, serviceaccount, route, configmap, oc_obj)


oc_route depends on https://github.com/openshift/openshift-ansible/pull/4840

roles/openshift_prometheus/defaults/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+openshift_prometheus_cpu_requests_alertmanager: null
+openshift_prometheus_memory_requests_alertmanager: null
+openshift_prometheus_cpu_limit_alert_buffer: null
+openshift_prometheus_memory_limit_alert_buffer: null
+openshift_prometheus_cpu_requests_alert_buffer: null
+openshift_prometheus_memory_requests_alert_buffer: null
+openshift_prometheus_cpu_limit_oauth_proxy: null
+openshift_prometheus_memory_limit_oauth_proxy: null
+openshift_prometheus_cpu_requests_oauth_proxy: null
+openshift_prometheus_memory_requests_oauth_proxy: null
```
zgalor: @joelddiaz I wasn't sure if I should add limits and requests for each container and what should be the default values. Currently, I added null defaults for all. could you assist with the defaults?

roles/openshift_prometheus/defaults/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+# defaults file for openshift_prometheus
+openshift_prometheus_state: present
+
+openshift_prometheus_namespace: prometheus
+
+openshift_prometheus_replicas: 1
+openshift_prometheus_node_selector: {"region":"infra"}
+
+# container resources
+openshift_prometheus_cpu_limit_prometheus: null
```
jcantrill: Consider inverting the naming convention to have the 'component' before the 'property' which is what we did with logging.  It allows for a more natural grouping of settings by subcomponent: 
`openshift_prometheus_alert_buffer_cpu_limit`

roles/openshift_prometheus/defaults/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+openshift_prometheus_memory_limit_prometheus: null
```
jcantrill: You could probably just drop 'prometheus' for these values assuming you make the change for better readability

playbooks/byo/openshift-prometheus.yaml
```diff
@@ -0,0 +1,10 @@
+---
+- name: generate prometheus deployment
+  hosts: masters
+  roles:
+  - openshift_prometheus
+  vars:
+    openshift_prometheus_state: present
+    # support kubernetes < 1.7
+    openshift_prometheus_use_default_metrics_path: True
```
zgalor: @joelddiaz this flag is for the default endpoint (before kubernetes 1.7).
remove it if you update to a newer cluster.

playbooks/byo/openshift-prometheus.yaml
```diff
@@ -0,0 +1,10 @@
+---
```
joelddiaz: openshift-ansible is in the middle of refactoring the way they do these top-level byo playbooks. look to https://github.com/openshift/openshift-ansible/blob/master/playbooks/byo/openshift-cluster/config.yml for an example of how they are doing things now.

i'm not 100% familiar with all the details, but someone on the openshift-ansible team should be able to provide more info.

playbooks/byo/openshift-prometheus.yaml
```diff
@@ -0,0 +1,10 @@
...
+- name: generate prometheus deployment
+  hosts: masters
```
joelddiaz: the tasks only need to run against a single master node (not all of them)

i'm seeing failures of this kind (presumably) because of this:

```
TASK [openshift_prometheus : Set alert and prometheus secrets] *****************
Thursday 27 July 2017  13:45:16 +0000 (0:13:20.345)       0:17:53.459 ********* 
changed: [master1] => (item=prometheus)
failed: [master2] (item=prometheus) => {"failed": true, "item": "prometheus", "msg": {"cmd": "/usr/bin/oc secrets new prometheus-proxy session_secret=/
tmp/session_secret-O7JqOq -n prometheus", "results": {}, "returncode": 1, "stderr": "Error from server (AlreadyExists): secrets \"prometheus-proxy\" already 
exists\n", "stdout": ""}}
failed: [master3] (item=prometheus) => {"failed": true, "item": "prometheus", "msg": {"cmd": "/usr/bin/oc secrets new prometheus-proxy session_secret=/t
mp/session_secret-zx9yIP -n prometheus", "results": {}, "returncode": 1, "stderr": "Error from server (AlreadyExists): secrets \"prometheus-proxy\" already e
xists\n", "stdout": ""}}
changed: [master2] => (item=alerts)
failed: [master1] (item=alerts) => {"failed": true, "item": "alerts", "msg": {"cmd": "/usr/bin/oc secrets new alerts-proxy session_secret=/tmp/session_s
ecret-mOz9Ct -n prometheus", "results": {}, "returncode": 1, "stderr": "Error from server (AlreadyExists): secrets \"alerts-proxy\" already exists\n", "stdou
t": ""}}
changed: [master3] => (item=alerts)
```

roles/lib_openshift/library/oc_route.py
```diff
@@ -1707,14 +1704,14 @@ def run_ansible(params, check_mode=False):
         if params['tls_termination'] and params['tls_termination'].lower() != 'passthrough':  # E501
 
             for key, option in files.items():
-                if key == 'destcacert' and params['tls_termination'] != 'reencrypt':
+                if not option['path'] and not option['content']:
```
joelddiaz: two things

1) modifying lib_openshift should be in it's own PR
2) when modifying lib_openshift, you need to modify lib_openshift/src/* files and then generate the lib_openshift/library/* files

roles/openshift_prometheus/templates/prometheus.rules.j2
```diff
@@ -0,0 +1,13 @@
...
+  rules:
+  - alert: Node Down
+    expr: up{job="kubernetes-nodes"} == 0
+    annotations:
+      miqTarget: "ContainerNode"
+      severity: "HIGH"
+      message: "{{ '{{' }}{{ '$labels.instance' }}{{ '}}' }} is down"
+{% if openshift_prometheus_external_rules is defined and openshift_prometheus_external_rules is not none %}
+  - {{ openshift_prometheus_external_rules }}
+{% endif %}
```
zgalor: @simon3z this is the way external alert rules are handled. The user can override the `openshift_prometheus_external_rules` variable with additional rule definition and those will be added to the configmap

roles/lib_openshift/library/oc_route.py
zgalor: @joelddiaz it is in a different PR (#4887). Just cherry-picked it so this PR is complete for itself.
Thanks for the generation note. I fixed the commit in the other PR and rebased.

roles/openshift_prometheus/templates/prometheus.rules.j2
simon3z: @zgalor we probably want to remove the `example-rules` for now (wouldn't it be deployed by default now?).
Let's wait to add official rules when we'll have them.

Also, the `openshift_prometheus_external_rules` could it be a file that is just copied in an additional `ConfigMap`?

zgalor: @simon3z I will condition the creation of the example rules with a variable. (I don't want to lose this example code, and I also want to keep the ability to test).

> Also, the openshift_prometheus_external_rules could it be a file that is just copied in an additional ConfigMap?

Yes. I can change it so the user will pass a rule file path if you think this is better

kwoodson: @zgalor, can this example code to the README for the new role?  Might be a good options to retain the documentation for what the example looks like.

roles/openshift_prometheus/templates/alertmanager.yml.j2
```diff
@@ -0,0 +1,20 @@
...
+  # TODO:
+  group_by: []
+
+  # All the above attributes are inherited by all child routes and can
+  # overwritten on each.
+
+receivers:
+- name: alert-buffer-wh
+  webhook_configs:
+  - url: http://localhost:9099/topics/alerts
```
kwoodson: Need a new line character here.

playbooks/common/openshift-cluster/openshift_prometheus.yml
```diff
@@ -0,0 +1,12 @@
+---
+- include: evaluate_groups.yml
```
kwoodson: This needs to be `std_include.yml` which also does `evaluate_groups.yml`.   We are moving to requiring all entry point playbooks to flow through the `std_include.yml` to ensure proper initialization.

roles/openshift_prometheus/tests/test.yaml
```diff
@@ -0,0 +1,5 @@
+---
+- hosts: localhost
+  remote_user: root
+  roles:
+    - openshift_prometheus
```
kwoodson: new line character here.

roles/openshift_prometheus/vars/main.yaml
```diff
@@ -0,0 +1,2 @@
+---
+# vars file for openshift_prometheus
```
kwoodson: If no vars are used we can remove these files.

joelddiaz:
@zgalor current version now runs against only a single master, so the playbook gets much further. It eventually failed for me with:

TASK [openshift_prometheus : create prometheus and alerts routes] **************
Wednesday 02 August 2017  18:19:01 +0000 (0:00:01.577)       0:06:32.240 ****** 
failed: [master] (item={u'name': u'prometheus'}) => {"failed": true, "item": {"name": "prometheus"}, "msg": {"cmd": "/usr/bin/oc create -f /tmp/promet
heus-nSpTKt -n prometheus", "results": {}, "returncode": 1, "stderr": "The Route \"prometheus\" is invalid: spec.tls.destinationCACertificate: Required valu
e\n", "stdout": ""}, "state": "present"}

This looks to be the same output you'd get from an incomplete 'oc create route':

[root@certtest-master-35c24 ~]# oc create route reencrypt --service docker-registry
The Route "docker-registry" is invalid: spec.tls.destinationCACertificate: Required value
[root@certtest-master-35c24 ~]#

zgalor:
@joelddiaz this happened to me before. This oc route requirement for destination ca certificate was removed in openshift 3.6. According to @smarterclayton `3.6.0.alpha2` is the minimal version for support of this. What's your oc version (version string)?

joelddiaz:
@zgalor here's my cluster's details:

zgalor:
@joelddiaz I don't know how to match versioning to release tags.
alpha2 has released 2 months ago and supports omitting the destination ca certificate field (https://github.com/openshift/origin/releases/v3.6.0-alpha.2).
If your cluster doesn't support this features, seems you need to upgrade it.

I am using 
```
oc v3.6.136
kubernetes v1.6.1+5115d708d7
```

zgalor:
Added support for additional.rules file and updated README.md
@moolitayer @simon3z @kwoodson can you please review this?

roles/openshift_prometheus/tasks/install_prometheus.yaml
```diff
@@ -0,0 +1,223 @@
+---
```
smarterclayton: I don't think this is a sustainable pattern.  This means we are defining config twice - once in origin that uses core kube objects, and once here.  I don't think we should be turning core origin config into roles like this.

I'd like the config to be owned by the team - to be provided as a set of core yaml to ansible.  I'd then like ansible to be customized as necessary for the supported outcomes.  I do not like recreating the config in total in another form for Ansible - that's unsustainable in the long run.  The core config in OpenShift is our supported platform.

@liggitt @sdodson this was our discussion today - I do not want to see this pattern continued, because it is not leveraging the platform as it is intended to be used.

kwoodson: @liggitt @sdodson @smarterclayton, I would love to have a conversation about this, specifically the topics that I'd like to see covered:

- Where the responsibility lies on managing the object(s)/technology being deployed?
- How do we achieve idempotency?
- What pattern is sustainable with idempotency in mind?

joelddiaz:
@zgalor reinstalled my cluster with newer versions that allows route create without requiring the destinationCACertificate field (oc v3.6.173.0.5 / kubernetes v1.6.1+5115d708d7). I pulled to sync up with the latest version of this PR before trying again.

It failed with:
```
TASK [openshift_prometheus : annotate serviceaccount] **************************
Monday 07 August 2017  19:37:07 +0000 (0:00:01.147)       0:01:12.100 ********* 
fatal: [52.15.132.77]: FAILED! => {"changed": true, "cmd": ["oc", "annotate", "--overwrite", "serviceaccount", "prometheus", "serviceaccounts.openshift.io/oauth-redirectreference.primary={\"kind\":\"OAuthRedirectReference\",\"apiVersion\":\"v1\",\"reference\":{\"kind\":\"Route\",\"name\":\"prometheus\"}}"], "delta": "0:00:00.225287", "end": "2017-08-07 19:37:08.175299", "failed": true, "rc": 1, "start": "2017-08-07 19:37:07.950012", "stderr": "Error from server (NotFound): serviceaccounts \"prometheus\" not found", "stdout": "", "stdout_lines": [], "warnings": []}
```
I don't see a '--namespace=prometheus' parameter in that annotate command.

It later will fail for a similar reason on the 'annotate services' task:

```
TASK [openshift_prometheus : annotate services] ********************************
Monday 07 August 2017  19:59:42 +0000 (0:00:01.160)       0:01:44.948 ********* 
failed: [52.15.132.77] (item=prometheus) => {"changed": true, "cmd": ["oc", "annotate", "--overwrite", "service", "prometheus", "service.alpha.openshift.io/serving-cert-secret-name=prometheus-tls"], "delta": "0:00:00.222903", "end": "2017-08-07 19:59:43.043972", "failed": true, "item": "prometheus", "rc": 1, "start": "2017-08-07 19:59:42.821069", "stderr": "Error from server (NotFound): services \"prometheus\" not found", "stdout": "", "stdout_lines": [], "warnings": []}
failed: [52.15.132.77] (item=alerts) => {"changed": true, "cmd": ["oc", "annotate", "--overwrite", "service", "alerts", "service.alpha.openshift.io/serving-cert-secret-name=alerts-tls"], "delta": "0:00:00.223345", "end": "2017-08-07 19:59:43.416370", "failed": true, "item": "alerts", "rc": 1, "start": "2017-08-07 19:59:43.193025", "stderr": "Error from server (NotFound): services \"alerts\" not found", "stdout": "", "stdout_lines": [], "warnings": []}
```
Finally, the pods weren't coming up at all, and after looking at it, I noticed the nodeSelector field (in the deployment object) was set to 'region: infra'. That didn't look right, so I changed it to 'type: infra', and now the status of the replicaset is:
```
message: 'unable to create pods: pods "" is forbidden: pod node label selector
      conflicts with its project node label selector'
```

Changed the namespace object to use a custom node-selector:
```
openshift.io/node-selector: “type=infra"
```
Now the pods finally started up on one of the infra nodes.

Lastly, I tried to log into the prometheus web UI, and after going through the Oauth steps, I ended up with:
```
500 Internal Error
```
I was able to authenticate into the cluster's web console, so the cluster's OAuth should be okay.

zgalor:
@joelddiaz I fixed the annotation issue.
regarding node selection: `region: infra` is the convention I am familiar with, and I had no issues with that and my pod gets deployed on an infra node.
Please see this comment https://github.com/openshift/origin/pull/15078#issuecomment-319074492

Regarding the `500 Internal Error`, I didn't encounter that, but I'll be happy to assist. ping me when you are on it.

roles/openshift_prometheus/tasks/install_prometheus.yaml
```diff
@@ -0,0 +1,223 @@
...
+  oc_project:
+    state: "{{ state }}"
+    name: "{{ openshift_prometheus_namespace }}"
+    description: Prometheus
+
+# secrets
+- name: Set alert and prometheus secrets
+  oc_secret:
+    state: "{{ state }}"
+    name: "{{ item }}-proxy"
```
smarterclayton: All of these names need to be consistent with the openshift origin template.  If we want to further prefix them, they should be prefixed in the origin template (i.e. I don't want drift in how the version in origin works to how the one installed by ansible works).

roles/openshift_prometheus/tasks/install_prometheus.yaml
smarterclayton: Yeah this is a much bigger issue for sure - ultimately we need to be config driven (kube api objects, not layers on top) and then be able to cleanly layer the customizations we need in Ansible.  That really has to start with a set of Kube API config objects (maybe 2 or 3 sets, depending on how we want to break out policy / secrets from those objects).

Recreating Kube config by hand as ansible is going to rapidly become unmaintainable.  Kube will get better at running raw config, and so there will be gravity back to raw config objects that will be difficult and costly to reproduce elsewhere.  We need to be on a track where the core config is copied into a playbook and then further customized before being applied.

Let's set some time up to talk this week, but I think we have enough customizations with the yaml editor module today that this should be "import the upstream template", "parameterize a few things", and "deploy".

roles/openshift_prometheus/tasks/install_prometheus.yaml
zgalor: I agree, I just applied your renaming requests in https://github.com/openshift/origin/pull/15078. Once you approve them, I'll align the naming on this PR.

joelddiaz:
@zgalor tried with your recent annotations changes. that now works. that last thing i manually had to do to get the pods running was to annotate the 'prometheus' namespace/project to use a custom node-selector:

openshift.io/node-selector: “type=infra"

This value should probably just be whatever openshift_prometheus_node_selector is set to.

jcantrill:
[(?) track prometheus productization by cloudforms team](https://trello.com/c/mfX6tHai/553-track-prometheus-productization-by-cloudforms-team)

roles/openshift_prometheus/tasks/install_prometheus.yaml
```diff
@@ -0,0 +1,223 @@
+---
+
+# namespace
+- name: Add prometheus project
+  oc_project:
+    state: "{{ state }}"
+    name: "{{ openshift_prometheus_namespace }}"
+    node_selector: "{{ openshift_prometheus_node_selector | regex_replace(':','=') | regex_replace('{', '[') | regex_replace('}', ']') }}"
```
zgalor: @joelddiaz I added the node selector field to oc_project.
It seems to require specific formating 'region=infra' and not support "region:infra" filter that is supported by deployment object, so I had to hack it a bit.
This works on my cluster. could you check on yours?

joelddiaz: @zgalor it fails to create the project for me with the recent changes. There is a bug in oc_project that was causing the module to run without erroring out (fixed with #5058 ), but even without this fix the project isn't being created. It just makes you believe the project was created.

The next task would fail immediately because the project wasn't created.

With the oc_project fix applied, it gives more useful output:

```
TASK [openshift_prometheus : Add prometheus project] ***************************
Thursday 10 August 2017  20:25:18 +0000 (0:00:00.050)       0:01:42.673 ******* 
fatal: [54.191.165.68]: FAILED! => {
    "changed": false, 
    "failed": true
}

MSG:

{u'returncode': 1, u'cmd': u"/usr/bin/oc adm new-project prometheus --admin-role=admin --description=Prometheus --node-selector=['type'= 'infra']", u'results': {}, u'stderr': u'The Project "prometheus" is invalid: nodeSelector: Invalid value: "[\'type\'= \'infra\']": must be a valid label selector\n', u'stdout': u''}
```

zgalor: I believe my last commit fixed that. @joelddiaz can you please check?

roles/openshift_prometheus/tasks/install_prometheus.yaml
```diff
@@ -0,0 +1,223 @@
+---
+
+# namespace
+- name: Add prometheus project
+  oc_project:
+    state: "{{ state }}"
+    name: "{{ openshift_prometheus_namespace }}"
+    node_selector: "{{ openshift_prometheus_node_selector | replace(':','=') | replace('{', '') | replace('}', '') | replace(' ', '') | replace(\"u\'\", '') | replace(\"\'\", '') }}"
```
zgalor: @kwoodson @jcantrill I am very unhappy with what I did here but it's the only thing that worked.
IIUC, nodeselector labels for deployment are formatted in dictionary mode, while in namespace the equality-based definition is used (https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/).
So I had to convert it somehow, and I didn't want to write an entire module for that.

Do you have a better proposal on how to accomplish this?
@kwoodson does it seem logical that the oc_project module will also accept dictionary based labeling and convert it?

roles/openshift_prometheus/tasks/install_prometheus.yaml
joelddiaz: The changes now allow the project to be created successfully. The role ran to completion.

roles/openshift_prometheus/tasks/install_prometheus.yaml
jcantrill: @zgalor Not sure exactly what you are saying here, but keys in inventory files can accept  a hash that can be directly applied in a template.  If the 'oc_project' module does not have the logic to accept one or the other we should make changes to accomplish that.

joelddiaz: @zgalor you could also write a filter_plugin to to the dict_to_string reformatting of the data. It looks like it is possible to specify more than one criteria for the node selector, and with the way it is written right now, it pretty much assumes only one node selector.

zgalor: @joelddiaz I tested it with multiple criteria: `{"region":"infra", "key":"val"}`

zgalor: dict_to_string is not enough because you also need to strip quotes and braces. it's a very proprietary formatting so I don't this it justifies its own plugin. I think adding support in oc_project is a better way to go.

joelddiaz: @zgalor my mistake, it does work with multiple selectors in its current form.

When I mentioned dict_to_string i had something like this in mind in the filter_plugin

def selector_to_string(user_dict):
  selectors = []
  for key in user_dict:
    selectors.append("{}={}".format(key, user_dict[key]))
  return ",".join(selectors)

zgalor: @joelddiaz do you mean adding a plugin to https://github.com/ansible/ansible/tree/devel/lib/ansible/plugins/filter ?
or internally somehow?

joelddiaz: @zgalor i was suggesting putting this kind of function into a role-specific filter plugin. An example of how we've done it in the past here: https://github.com/openshift/openshift-tools/blob/prod/ansible/roles/openshift_aws_persistent_volumes/filter_plugins/filters.py

zgalor: @joelddiaz cool!
I was unaware of that option.
much cleaner.
I'll check it out

zgalor:
@sdodson I finished my latest refactoring to align with https://github.com/openshift/origin/blob/master/examples/prometheus/prometheus.yaml

Please review

ilackarms:
getting a syntax error on the `install_prometheus.yaml` task. maybe the error is my side?

```
ERROR! Syntax Error while loading YAML.


The error appears to have been in '/home/ilackarms/workspace/openshift-ansible/roles/openshift_prometheus/tasks/install_prometheus.yaml': line 8, column 149, but may

    name: "{{ openshift_prometheus_namespace }}"
    node_selector: "{{ openshift_prometheus_node_selector | replace(':','=') | replace('{', '') | replace('}', '') | replace(' ', '') | replace(\"u\'\", '') | replace(\"\'\", '') }}"
                                                                                                                                                    ^ here
start a value. For instance:

    with_items:
      - {{ foo }}

Should be written as:

    with_items:
      - "{{ foo }}"
```

moolitayer:
The escaping doesn't look right on `\"u\'\"`, maybe outer `"` should not be escaped? not sure what this is doing tough.

zgalor:
@ilackarms @moolitayer this has been working for a while but might be buggy. Going to replace this with filter plugin as @joelddiaz suggested

tbielawa:
Some lint
```
playbooks/common/openshift-cluster/openshift_prometheus.yml:
  10:1      error    too many blank lines (1 > 0)  (empty-lines)
roles/openshift_prometheus/meta/main.yaml:
  1:1       warning  missing document start "---"  (document-start)
roles/openshift_prometheus/tasks/install_prometheus.yaml:
  8:149     error    syntax error: found unknown escape character "'"
```

1. Easy fix
2. Easy fix
3. More interesting fix

I'm seeing that unknown escape character, `"` and looking at this part of the filter pipeline:

```
replace(\"u\'\", '') | replace(\"\'\", '')
```

I think this PR is still failing due to that wonky filter pipeline. Did anything come of that previous discussion in this PR? https://github.com/openshift/openshift-ansible/pull/4509#pullrequestreview-56068452 about adding a `selector_to_string(user_dict)` filter plugin?

tbielawa:
Don't forget, in case you didn't know, you can run `tox -e py27-yamllint` on your local work stations to save back and forth time waiting on Travis to test your PR.

tbielawa:
After I fixed up the filter pipeline I ran tox again locally and more lint creeped up

```
roles/openshift_prometheus/tasks/install_prometheus.yaml:
  30:1      warning  comment not indented like content  (comments-indentation)
  36:5      error    wrong indentation: expected 6 but found 4  (indentation)
  68:1      warning  comment not indented like content  (comments-indentation)
  72:5      error    wrong indentation: expected 6 but found 4  (indentation)
  86:1      warning  comment not indented like content  (comments-indentation)
  90:5      error    wrong indentation: expected 6 but found 4  (indentation)
  158:5     error    wrong indentation: expected 6 but found 4  (indentation)
  168:3     error    wrong indentation: expected 4 but found 2  (indentation)
```

The diff below fixes all of that as well as adds the filter plugin

```diff
diff --git a/filter_plugins/oo_filters.py b/filter_plugins/oo_filters.py
index 36a90a8..2eaa172 100644
--- a/filter_plugins/oo_filters.py
+++ b/filter_plugins/oo_filters.py
@@ -1023,6 +1023,16 @@ def oo_contains_rule(source, apiGroups, resources, verbs):
 
     return False
 
+def oo_selector_to_string_list(user_dict):
+    """Convert a dict of selectors to a key=value list of strings
+
+Given input of {'region': 'infra', 'zone': 'primary'} returns a list
+of items as ['region=infra', 'zone=primary']
+    """
+    selectors = []
+    for key in user_dict:
+        selectors.append("{}={}".format(key, user_dict[key]))
+    return selectors
 
 class FilterModule(object):
     """ Custom ansible filter mapping """
@@ -1065,5 +1075,6 @@ class FilterModule(object):
             "oo_openshift_loadbalancer_backends": oo_openshift_loadbalancer_backends,
             "to_padded_yaml": to_padded_yaml,
             "oo_random_word": oo_random_word,
-            "oo_contains_rule": oo_contains_rule
+            "oo_contains_rule": oo_contains_rule,
+            "oo_selector_to_string_list": oo_selector_to_string_list
         }
diff --git a/playbooks/common/openshift-cluster/openshift_prometheus.yml b/playbooks/common/openshift-cluster/openshift_prometheus.yml
index fc16ad6..a979c0c 100644
--- a/playbooks/common/openshift-cluster/openshift_prometheus.yml
+++ b/playbooks/common/openshift-cluster/openshift_prometheus.yml
@@ -7,4 +7,3 @@
   - openshift_prometheus
   vars:
     openshift_prometheus_state: present
-
diff --git a/roles/openshift_prometheus/meta/main.yaml b/roles/openshift_prometheus/meta/main.yaml
index c323e7f..33188bb 100644
--- a/roles/openshift_prometheus/meta/main.yaml
+++ b/roles/openshift_prometheus/meta/main.yaml
@@ -1,3 +1,4 @@
+---
 galaxy_info:
   author: OpenShift Development <dev@lists.openshift.redhat.com>
   description: Deploy OpenShift prometheus integration for the cluster
diff --git a/roles/openshift_prometheus/tasks/install_prometheus.yaml b/roles/openshift_prometheus/tasks/install_prometheus.yaml
index a9c33bb..a6bed55 100644
--- a/roles/openshift_prometheus/tasks/install_prometheus.yaml
+++ b/roles/openshift_prometheus/tasks/install_prometheus.yaml
@@ -5,7 +5,7 @@
   oc_project:
     state: "{{ state }}"
     name: "{{ openshift_prometheus_namespace }}"
-    node_selector: "{{ openshift_prometheus_node_selector | replace(':','=') | replace('{', '') | replace('}', '') | replace(' ', '') | replace(\"u\'\", '') | replace(\"\'\", '') }}"
+    node_selector: "{{ openshift_prometheus_node_selector | oo_selector_to_string_list() }}"
     description: Prometheus
 
 # secrets
@@ -27,13 +27,13 @@
     state: "{{ state }}"
     name: prometheus
     namespace: "{{ openshift_prometheus_namespace }}"
-#    TODO add annotations when supproted
-#    annotations:
-#      serviceaccounts.openshift.io/oauth-redirectreference.prom: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"prometheus"}}'
-#      serviceaccounts.openshift.io/oauth-redirectreference.alerts: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"alerts"}}'
+    #    TODO add annotations when supproted
+    #    annotations:
+    #      serviceaccounts.openshift.io/oauth-redirectreference.prom: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"prometheus"}}'
+    #      serviceaccounts.openshift.io/oauth-redirectreference.alerts: '{"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"alerts"}}'
 
     secrets:
-    - prometheus-secrets
+      - prometheus-secrets
   changed_when: no
 
 # TODO remove this when annotations are supported by oc_serviceaccount
@@ -65,12 +65,12 @@
       app: prometheus
     labels:
       name: "{{ item.name }}"
-#    TODO add annotations when supported
-#    annotations:
-#      service.alpha.openshift.io/serving-cert-secret-name: "{{item.name}}-tls"
+      #    TODO add annotations when supported
+      #    annotations:
+      #      service.alpha.openshift.io/serving-cert-secret-name: "{{item.name}}-tls"
     ports:
-    - port: 443
-      targetPort: 8443
+      - port: 443
+        targetPort: 8443
   with_items:
     - name: prometheus
 
@@ -83,12 +83,12 @@
       app: prometheus
     labels:
       name: "{{ item.name }}"
-#    TODO add annotations when supported
-#    annotations:
-#      service.alpha.openshift.io/serving-cert-secret-name: "{{item.name}}-tls"
+      #    TODO add annotations when supported
+      #    annotations:
+      #      service.alpha.openshift.io/serving-cert-secret-name: "{{item.name}}-tls"
     ports:
-    - port: 443
-      targetPort: 9443
+      - port: 443
+        targetPort: 9443
   with_items:
     - name: alerts
 
@@ -155,7 +155,7 @@
     namespace: "{{ openshift_prometheus_namespace }}"
     kind: deployment
     files:
-    - "{{ tempdir }}/templates/prometheus.yaml"
+      - "{{ tempdir }}/templates/prometheus.yaml"
     delete_after: true
 
 # prometheus configmap
@@ -165,9 +165,9 @@
     src: "{{ openshift_prometheus_additional_rules_file }}"
     dest: "{{ tempdir }}/prometheus.additional.rules"
   when:
-  - openshift_prometheus_additional_rules_file is defined
-  - openshift_prometheus_additional_rules_file is not none
-  - openshift_prometheus_additional_rules_file | trim | length > 0
+    - openshift_prometheus_additional_rules_file is defined
+    - openshift_prometheus_additional_rules_file is not none
+    - openshift_prometheus_additional_rules_file | trim | length > 0
 
 - stat:
     path: "{{ tempdir }}/prometheus.additional.rules"
```

tbielawa:
You can merge or cherry pick the commit from my branch if you want to save time

tbielawa/openshift-ansible.git#prometheus-role

zgalor:
@tbielawa thanks!
This is all very helpful!

I cherry picked your commit and pushed

zgalor:
PV creation added for internal nfs
@ilackarms can you try using the role now, and validate the issues you had were resolved?

zgalor:
@joelddiaz if your request for changes were met, can you please approve the PR?

tbielawa:
Whoops. I messed up some formatting on the filter plugins and it broke the travis test. Sorry about that.

```
./filter_plugins/oo_filters.py:1026:1: E302 expected 2 blank lines, found 1
./filter_plugins/oo_filters.py:1037:1: E302 expected 2 blank lines, found 1
./roles/openshift_master_facts/filter_plugins/oo_filters.py:1026:1: E302 expected 2 blank lines, found 1
./roles/openshift_master_facts/filter_plugins/oo_filters.py:1037:1: E302 expected 2 blank lines, found 1
```

There needs to be an extra newline before the `oo_selector_to_string_list` definition and one following it.

roles/openshift_prometheus/defaults/main.yaml
```diff
@@ -0,0 +1,74 @@
...
+    pv_label: Prometheus Server PV
+  - pv_name: prometheus-alertmanager
+    pv_template: prom-pv-alertmanager.yml
+    pv_label: Prometheus Alertmanager PV
+  - pv_name: prometheus-alertbuffer
+    pv_template: prom-pv-alertbuffer.yml
+    pv_label: Prometheus Alert Buffer PV
+
+# Hostname/IP of the NFS server. Currently defaults to first master
+openshift_prometheus_nfs_server: "{{ groups.nfs.0 }}"
```
tbielawa: Fix in follow-up PR, I believe this should be `oo_nfs_to_config.0`

roles/openshift_prometheus/tasks/install_prometheus.yaml
```diff
@@ -0,0 +1,241 @@
...
+        targetPort: 9443
+  with_items:
+    - name: alerts
+
+
+# Annotate services with secret name
+# TODO remove this when annotations are supported by oc_service
+- name: annotate prometheus service
+  command: >
+    {{ openshift.common.client_binary }} annotate --overwrite -n {{ openshift_prometheus_namespace }}
```
tbielawa: In a future PR, maybe with some other work required first, we need to work on standardizing how we are applying annotations

zgalor:
@tbielawa I fixed the newline and squashed. Can you please review?

roles/openshift_prometheus/tasks/main.yaml
```diff
@@ -0,0 +1,26 @@
+---
+
+- name: Create temp directory for doing work in on target
+  command: mktemp -td openshift-prometheus-ansible-XXXXXX
+  register: mktemp
+  changed_when: False
+
+- set_fact:
+    tempdir: "{{ mktemp.stdout }}"
```
tbielawa: Should use a more descriptive local name than just `tempdir`. Fix in follow-up.

tbielawa: Also going to assume this should be ran on the local control host, not whatever host you run the install playbook against (like `oo_first_master`). But either will work I suppose. 

Check back on this in follow-up PR

tbielawa:
aos-ci-test

tbielawa:
[test]

joelddiaz:
@zgalor i updated to the latest version of this PR and re-ran a prometheus deployment. I'm now seeing that the PVC has a selector which is causing problems:

[root@certtest-master-251f8 master]# oc get pvc prometheus -o json | jq '.spec.selector'
{
  "matchLabels": {
    "storage": "prometheus"
  }
}

This is causing the PVC to be stuck in pending state:

Failed to provision volume with StorageClass "gp2": claim.Spec.Selector is not supported for dynamic provisioning on AWS

zgalor:
@joelddiaz you can override `openshift_prometheus_pvc_pv_selector` `openshift_prometheus_alertmanager_pvc_pv_selector` and `openshift_prometheus_alertbuffer_pvc_pv_selector` variables with empty selector `{}`.

I will check in other roles, and if the pv selector is empty by default and passed externally, I'll align this role to it.
Check back on this in follow-up PR

joelddiaz:
@zgalor looking at the kubernetes docs, it says we can't have dynamically allocated PVs with a selector:

from https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1:
Note that currently, a PVC with a non-empty selector can’t have a PV dynamically provisioned for it.

zgalor:
@joelddiaz I changed the default to empty selectors. Please recheck

joelddiaz:
@zgalor PVs get created successfully now.

Current issue I see is with the 'kubernetes-service-endpoints' target configs in prometheus.yml. Those endpoints are HTTP basic auth protected, so I had to add a section like this to the  'kubernetes-service-endpoints' job configuration section in prometheus.yml to get those targets to work.

basic_auth:
  username: <username from 'oc get dc router -n default' ... the STATS_USERNAME env var>
  password: <password from 'oc get dc router -n default' ... the STATS_PASSWORD env var>

Without this, those targets were failing when prometheus would attempt to scrape.

zgalor:
@joelddiaz we should check usage with @smarterclayton as this is a new configuration he recently added

tbielawa:
@zgalor or @joelddiaz can one of you please summarize what's changing right now? I just want to make sure I'm staying up to date with this PR.

zgalor:
@tbielawa I simply removed the node_selection on the pvcs to allow binding with dynamically provisioned PVs.
The selector is empty by default, but can be overridden by external parameters.

tbielawa:
Flake https://github.com/openshift/origin/issues/16103

tbielawa:
re[test]

tbielawa:
flake https://github.com/openshift/origin/issues/16005

tbielawa:
re[test]

tbielawa:
aos-ci-test

tbielawa:
aos-ci-test

tbielawa:
Ugh...

```
    "msg": "No package matching 'atomic-openshift-docker-excluder-3.6.173.0.21*' found available, installed or updated", 
    "rc": 126, 
    "results": [
        "No package matching 'atomic-openshift-docker-excluder-3.6.173.0.21*' found available, installed or updated"
    ]
```

@zgalor @joelddiaz current status:

There was a push to this that erased the `aos-ci-test` results that I didn't notice. I didn't realize that until after I had ran the origin-ci test tests three times. The first two times I ran the "[test]" commands it failed on CI flakes. The third time it didn't fail on flakes, rather it failed because `aos-ci-test` results had been erased. The above attempts from me to re-run `aos-ci-test` are failing due to the error quoted in the beginning of this response. There needs to be a new push to the testing respos before that can be resolved. Then we need to re-run `aos-ci-test` again, then we can finally run `[merge]`.

tbielawa:
aos-ci-test

zgalor:
Changed nfs exports ownership to `nfsnobody` to align with other roles exports

zgalor:
@tbielawa I probably should have postponed that small change to a follow-up PR. can you approve it, to re-run tests?

tbielawa:
aos-ci-test

tbielawa:
[test]

sdodson:
aos-ci-test

tbielawa:
aos-ci-test

tbielawa:
Lovin' it

```
Updating status for f1750331199aa9cc6b80d638b0e3a2d25da25a32
Failed to update commit status [HTTP 403]
{'status': '403 Forbidden', 'x-ratelimit-remaining': '0', 'x-github-media-type': 'github.v3; format=json', 'x-content-type-options': 'nosniff', 'content-security-policy': "default-src 'none'", 'access-control-expose-headers': 'ETag, Link, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval', 'transfer-encoding': 'chunked', 'x-github-request-id': 'BC52:3A33:2502161:4CF8F77:59AEAEDB', 'access-control-allow-origin': '*', 'strict-transport-security': 'max-age=31536000; includeSubdomains; preload', 'server': 'GitHub.com', 'content-encoding': 'gzip', 'x-ratelimit-limit': '5000', 'x-runtime-rack': '0.020023', 'x-xss-protection': '1; mode=block', 'date': 'Tue, 05 Sep 2017 14:04:12 GMT', 'x-frame-options': 'deny', 'x-oauth-scopes': 'gist, repo, user', 'content-type': 'application/json; charset=utf-8', 'x-accepted-oauth-scopes': 'repo', 'x-ratelimit-reset': '1504620422'}
{u'documentation_url': u'https://developer.github.com/v3/#rate-limiting', u'message': u'API rate limit exceeded for openshift-bot.'}
Build step 'Execute Python script' marked build as failure
[PostBuildScript] - Execution post build scripts.
```

sdodson:
aos-ci-test

tbielawa:
[test]

tbielawa:
flake https://github.com/openshift/origin/issues/16103

tbielawa:
[test]

sdodson:
aos-ci-test

tbielawa:
flake openshift/origin#16103

tbielawa:
re[test]

roles/openshift_prometheus/defaults/main.yaml
juzhao: Does prometheus only support NFS PV?

roles/openshift_prometheus/defaults/main.yaml
```diff
@@ -0,0 +1,74 @@
...
+openshift_prometheus_alertmanager_pvc_pv_selector: {}
+
+openshift_prometheus_alertbuffer_storage_type: pvc
+openshift_prometheus_alertbuffer_pvc_name: prometheus-alertbuffer
+openshift_prometheus_alertbuffer_pvc_size: 10G
+openshift_prometheus_alertbuffer_pvc_access_modes: [ReadWriteOnce]
+openshift_prometheus_alertbuffer_pvc_pv_selector: {}
+
+# container resources
+openshift_prometheus_cpu_limit: null
```
juzhao: What are the suggested value for CPU and memory limit?

roles/openshift_prometheus/defaults/main.yaml
```diff
@@ -0,0 +1,74 @@
...
+openshift_prometheus_alertmanager_pvc_access_modes: [ReadWriteOnce]
+openshift_prometheus_alertmanager_pvc_pv_selector: {}
+
+openshift_prometheus_alertbuffer_storage_type: pvc
+openshift_prometheus_alertbuffer_pvc_name: prometheus-alertbuffer
+openshift_prometheus_alertbuffer_pvc_size: 10G
+openshift_prometheus_alertbuffer_pvc_access_modes: [ReadWriteOnce]
+openshift_prometheus_alertbuffer_pvc_pv_selector: {}
+
+# container resources
```
juzhao: What are the suggested value for CPU and memory limit?

roles/openshift_prometheus/defaults/main.yaml
zgalor: currently, it doesn't create external nfs, but it can work with externally created nfs pvs.
Creating external nfs is planned for a follow up PR.

zgalor:
@sdodson @tbielawa Sorry, I had to re-push a fix to a permission issue that broke the entire deployment.
Please approve and re-run tests

sdodson:
aos-ci-test

roles/openshift_prometheus/defaults/main.yaml
```diff
@@ -0,0 +1,74 @@
+---
+# defaults file for openshift_prometheus
+openshift_prometheus_state: present
+
+openshift_prometheus_namespace: prometheus
+
+openshift_prometheus_replicas: 1
+openshift_prometheus_node_selector: {"region":"infra"}
+
+# images
```
juzhao: Will the default images be changed to other values in the future?

roles/openshift_prometheus/README.md
```diff
@@ -0,0 +1,95 @@
...
+- `openshift_prometheus_namespace`: project (i.e. namespace) where the components will be
+  deployed.
+
+- `openshift_prometheus_replicas`: The number of replicas for prometheus deployment.
+
+- `openshift_prometheus_node_selector`: Selector for the nodes prometheus will be deployed on.
+
+- `openshift_prometheus_image_<COMPONENT>`: specify image for the component 
+
+## Storage related variables
```
juzhao: I did not find oauth-proxy pv ansible parameters in  openshift-ansible/roles/openshift_prometheus/defaults/main.yaml, is these parameters missed/

roles/openshift_prometheus/defaults/main.yaml
zgalor: Initially, I will change `openshift_prometheus_image_alertbuffer` to official upstream image `openshift/prometheus-alert-buffer` that was created yesterday.
In the future, image tags/versions might change.

roles/openshift_prometheus/README.md
zgalor: We don't create a PV for oauth-proxy. we use the prometheus mounts for data, and local /etc for secrets.

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_prometheus/templates/prometheus_deployment.j2#L64

Same for alerts:
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_prometheus/templates/prometheus_deployment.j2#L136

roles/openshift_prometheus/tasks/install_prometheus.yaml
```diff
@@ -0,0 +1,241 @@
...
+  oc_secret:
+    state: "{{ state }}"
+    name: "{{ item }}-proxy"
+    namespace: "{{ openshift_prometheus_namespace }}"
+    contents:
+      - path: session_secret
+        data: "{{ 43 | oo_random_word }}="
+  with_items:
+    - prometheus
+    - alerts
```
simon3z: @zgalor I think this should have been:

```yaml
  with_items:
    - prometheus
    - prometheus-alerts
```
Are we still in time to fix it?

roles/openshift_prometheus/README.md
```diff
@@ -0,0 +1,95 @@
...
+Each prometheus component (prometheus, alertmanager, alert-buffer, oauth-proxy) can set pv claim by setting corresponding role variable:
+```
+openshift_prometheus_<COMPONENT>_storage_type: <VALUE>
+openshift_prometheus_<COMPONENT>_pvc_(name|size|access_modes|pv_selector): <VALUE>
+```
+e.g
+```
+openshift_prometheus_storage_type: pvc
+openshift_prometheus_alertmanager_pvc_name: alertmanager
+openshift_prometheus_alertbuffer_pvc_size: 10G
```
simon3z: @zgalor I think there is a mismatch with the PV which is 15Gb, is that OK?

zgalor: The PVs defaults is indeed more than 10G, but it will still match that PVC, and bind. I agree that further defaults discussion is required

roles/openshift_prometheus/tasks/install_prometheus.yaml
zgalor: This was intentional, to align with the template. 
Naming is not 100% unified but at least it is consistent. 
Only the configmap and service certificate secret are called `prometheus-alerts`. 
All other entities (service, route, proxy secrets) are called `alerts`.

I see we have a cert sign error from the service that doesn't appear when using the template. I am investigating this

roles/openshift_prometheus/tasks/install_prometheus.yaml
```diff
@@ -0,0 +1,241 @@
...
+    labels:
+      name: "{{ item.name }}"
+      #    TODO add annotations when supported
+      #    annotations:
+      #      service.alpha.openshift.io/serving-cert-secret-name: "{{item.name}}-tls"
+    ports:
+      - port: 443
+        targetPort: 9443
+  with_items:
+    - name: alerts
```
simon3z: @zgalor prometheus-alerts?

roles/openshift_prometheus/tasks/install_prometheus.yaml
```diff
@@ -0,0 +1,241 @@
...
+- name: create prometheus and alerts routes
+  oc_route:
+    state: "{{ state }}"
+    name: "{{ item.name }}"
+    namespace: "{{ openshift_prometheus_namespace }}"
+    service_name: "{{ item.name }}"
+    tls_termination: reencrypt
+  with_items:
+    - name: prometheus
+    - name: alerts
```
simon3z: @zgalor prometheus-alerts?

roles/openshift_prometheus/tasks/install_prometheus.yaml
zgalor: again. intentional to align with template

roles/openshift_prometheus/tasks/install_prometheus.yaml
zgalor: again.

We should have a complete naming discussion

roles/openshift_prometheus/defaults/main.yaml
```diff
@@ -0,0 +1,74 @@
...
+openshift_prometheus_pvc_size: 10G
+openshift_prometheus_pvc_access_modes: [ReadWriteOnce]
+openshift_prometheus_pvc_pv_selector: {}
+
+openshift_prometheus_alertmanager_storage_type: pvc
+openshift_prometheus_alertmanager_pvc_name: prometheus-alertmanager
+openshift_prometheus_alertmanager_pvc_size: 10G
+openshift_prometheus_alertmanager_pvc_access_modes: [ReadWriteOnce]
+openshift_prometheus_alertmanager_pvc_pv_selector: {}
+
```
juzhao: Will the alerts be deleted from alert buffer after they are received by webhook?

zgalor: Alert buffer holds alerts for 3 days. I think this can be configured. @ilackarms keep me honest here

roles/openshift_prometheus/defaults/main.yaml
```diff
@@ -0,0 +1,74 @@
...
+openshift_prometheus_cpu_requests: null
+openshift_prometheus_memory_requests: null
+openshift_prometheus_alertmanager_cpu_limit: null
+openshift_prometheus_alertmanager_memory_limit: null
+openshift_prometheus_alertmanager_cpu_requests: null
+openshift_prometheus_alertmanager_memory_requests: null
+openshift_prometheus_alertbuffer_cpu_limit: null
+openshift_prometheus_alertbuffer_memory_limit: null
+openshift_prometheus_alertbuffer_cpu_requests: null
+openshift_prometheus_alertbuffer_memory_requests: null
```
juzhao: Set resource limits for oauth_proxy
openshift_prometheus_oauth_proxy_cpu_limit=200m
openshift_prometheus_oauth_proxy_memory_limit=200Mi
openshift_prometheus_oauth_proxy_cpu_requests=200m
openshift_prometheus_oauth_proxy_memory_requests=200Mi

Get following pod info, memory limit is not shown in prom-proxy container, is this acceptable?
containers
  name: prom-proxy
    ports:
    - containerPort: 8443
      name: web
      protocol: TCP
    resources:
      limits:
        cpu: 200m
      requests:
        cpu: 200m
        memory: 200Mi


  name: alerts-proxy
    ports:
    - containerPort: 9443
      name: web
      protocol: TCP
    resources:
      limits:
        cpu: 200m
        memory: 200Mi
      requests:
        cpu: 200m
        memory: 200Mi

roles/openshift_prometheus/defaults/main.yaml
```diff
@@ -0,0 +1,74 @@
...
+# images
+openshift_prometheus_image_proxy: "openshift/oauth-proxy:v1.0.0"
+openshift_prometheus_image_prometheus: "openshift/prometheus:v2.0.0-dev"
+openshift_prometheus_image_alertmanager: "openshift/prometheus-alertmanager:dev"
+openshift_prometheus_image_alertbuffer: "ilackarms/message-buffer"
+
+# additional prometheus rules file
+openshift_prometheus_additional_rules_file: null
+
+# All the required exports
```
juzhao: Does prometheus must bound with PV?

zgalor: it is creating pvcs by default but can be overridden if `openshift_prometheus_storage_type` `openshift_prometheus_alertmanager_storage_type` `openshift_prometheus_alertbuffer_storage_type` are set to anything other than `pvc`.
This will become clearer in the nfs PR https://github.com/openshift/openshift-ansible/pull/5459

PR #4518, opened 2017-06-20
Allow intermediate CA for Aggregated Logging

ruromero:
openshift_logging role is not able to generate all the internal certificates when an Intermediate CA is provided. Keytool will fail with the following error

```
echo 'Import back to keystore (including CA chain)'
/bin/keytool -import -file /etc/deploy/scratch/ca.crt -keystore /etc/deploy/scratch/system.admin.jks -storepass kspass -noprompt -alias sig-ca
Import back to keystore (including CA chain)
Certificate was added to keystore
/bin/keytool -import -file /etc/deploy/scratch/system.admin.crt -keystore /etc/deploy/scratch/system.admin.jks -storepass kspass -noprompt -alias system.admin
keytool error: java.lang.Exception: Failed to establish chain from reply
```

### DIFF
--- a/roles/openshift_logging/files/generate-jks.sh
+++ b/roles/openshift_logging/files/generate-jks.sh
@@ -73,6 +73,7 @@ function generate_JKS_chain() {
         -keystore $dir/$NODE_NAME.jks \
         -storepass $ks_pass \
         -noprompt \
+        -trustcacerts \
         -alias $NODE_NAME
 
     echo All done for $NODE_NAME
@@ -132,6 +133,7 @@ function generate_JKS_client_cert() {
         -keystore $dir/$NODE_NAME.jks \
         -storepass $ks_pass \
         -noprompt \
+        -trustcacerts \
         -alias $NODE_NAME
 
     echo All done for $NODE_NAME


roles/openshift_logging/files/generate-jks.sh
```diff
@@ -73,6 +73,7 @@ function generate_JKS_chain() {
         -keystore $dir/$NODE_NAME.jks \
         -storepass $ks_pass \
         -noprompt \
+        -trustcacerts \
```
jcantrill: Im interested in understanding why you believe this resolves the issue.  If I read [1] and [2], it looks to me that this switch enables trust of the systemwide CAs that are installed on the runtime.  This means it will work IFF the cert you provide can be additionally trusted by one of the global CAs

[1] http://docs.oracle.com/javase/6/docs/technotes/tools/solaris/keytool.html
[2] http://docs.oracle.com/javase/6/docs/technotes/tools/solaris/keytool.html#cacerts

sferich888: Correct, adding -trustcacerts will just allow for "system CA" to be considered trusted when using this JKS store! The "system ca's" are stored in a certificates file named "cacerts" resides in the security properties directory, java.home/lib/security (and is a JKS store its self). 

This means, that IM CA's not added to that keystore, will not be trusted, so while allowing -trustcacerts is not a bad thing to have in this case, its not helping to solve the real problem. 

That said, related to this ansible file. There is all kinda of wrong with this script! 

1. Why are we using it an not using https://docs.openshift.com/container-platform/3.5/dev_guide/secrets.html#service-serving-certificate-secrets for the default scripts? 
2. Why does our method of setting up the CA (for a java application) differ from the process that MW uses with its templates (they require that you provide a JKS in a secret). Why does out image not require that, and if one is not provided, why don't we refer back to 1 or generate our own JKS in the image we ship? 

Doing thing in Ansible like this (and not using a common library) means that we end up having (by my count 3 ) different ways to setup / provide certificates to various parts of the system.

jcantrill: @sferich888 To answer most of your questions, its because this script is essentially what the deployer was doing is a direct port from that and was reused in the interest of time and resource constraints.  I am unable to answer the question related to MW, but we have https://trello.com/c/6a9bxQdX/445-2-jks-certs-on-container-start-of-logging-logging which is to address item 2 in your question.  Ultimately we wish to take advantage of a client cert generation service...where is that for us?

ruromero: @jcantrill Indeed it makes no difference, actually I should've added the flag to the command that adds the CA. I couldn't make it work running the deployer but I don't know what could be the reason.

vrutkovs:

PR #4519, opened 2017-06-20
verify sane log times in logging stack

juanvallejo:
Related trello card: https://trello.com/c/7uOluOTW
Based on @richm 's script: https://bugzilla.redhat.com/attachment.cgi?id=1258225

This patch verifies that logs sent from logging pods can be queried on
the Elasticsearch pod within a reasonable amount of time.

*TODO*

- [x] add tests

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/logging/logging.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/logging.py
@@ -20,7 +20,7 @@ class LoggingCheck(OpenShiftCheck):
     @staticmethod
     def is_first_master(task_vars):
         """Run only on first master and only when logging is configured. Returns: bool"""
-        logging_deployed = get_var(task_vars, "openshift_hosted_logging_deploy", default=True)
+        logging_deployed = get_var(task_vars, "openshift_hosted_logging_deploy", default=False)
         # Note: It would be nice to use membership in oo_first_master group, however for now it
         # seems best to avoid requiring that setup and just check this is the first master.
         hostname = get_var(task_vars, "ansible_ssh_host") or [None]
--- a/roles/openshift_health_checker/openshift_checks/logging/logging.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/logging.py
@@ -45,7 +45,7 @@ class LoggingCheck(OpenShiftCheck):
                 raise ValueError()
         except ValueError:
             # successful run but non-parsing data generally means there were no pods in the namespace
-            return None, 'There are no pods in the {} namespace. Is logging deployed?'.format(namespace)
+            return None, 'No pods were found for the "{}" logging component.'.format(logging_component)
 
         return pods['items'], None
 
--- a/roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
@@ -1,0 +1,163 @@
+"""
+Check for ensuring logs from pods can be queried in a reasonable amount of time.
+"""
+
+import json
+import time
+
+from uuid import uuid4
+
+from openshift_checks import get_var, OpenShiftCheckException
+from openshift_checks.logging.logging import LoggingCheck
+
+
+ES_CMD_TIMEOUT_SECONDS = 30
+
+
+class LoggingIndexTime(LoggingCheck):
+    """Check that pod logs are aggregated and indexed in ElasticSearch within a reasonable amount of time."""
+    name = "logging_index_time"
+    tags = ["health", "logging"]
+
+    logging_namespace = "logging"
+
+    def run(self, tmp, task_vars):
+        """Add log entry by making unique request to Kibana. Check for unique entry in the ElasticSearch pod logs."""
+        pod_log_timeout = int(get_var(task_vars, "openshift_check_pod_logs_timeout", default=ES_CMD_TIMEOUT_SECONDS))
+
+        # get all Kibana pods
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default=self.logging_namespace)
+        pods, error = super(LoggingIndexTime, self).get_pods_for_component(
+            self.execute_module,
+            self.logging_namespace,
+            "kibana",
+            task_vars,
+        )
+
+        if error:
+            msg = 'Unable to retrieve pods for the "Kibanna" logging component: {}'
+            return {"failed": True, "changed": False, "msg": msg.format(error)}
+
+        running_kibana_pods = self.running_pods(pods)
+        if not len(running_kibana_pods):
+            msg = ('No Kibana pods were found to be in the "Running" state.'
+                   'At least one Kibana pod is required in order to perform this check.')
+            return {"failed": True, "changed": False, "msg": msg}
+
+        # get all Elasticsearch pods
+        pods, error = super(LoggingIndexTime, self).get_pods_for_component(
+            self.execute_module,
+            self.logging_namespace,
+            "es",
+            task_vars,
+        )
+
+        if error:
+            msg = 'Unable to retrieve pods for the "Elasticsearch" logging component: {}'
+            return {"failed": True, "changed": False, "msg": msg.format(error)}
+
+        running_es_pods = self.running_pods(pods)
+        if not len(running_es_pods):
+            msg = ('No Elasticsearch pods were found to be in the "Running" state.'
+                   'At least one Elasticsearch pod is required in order to perform this check.')
+            return {"failed": True, "changed": False, "msg": msg}
+
+        uuid = self.curl_kibana_with_uuid(running_kibana_pods[0], task_vars)
+        self.wait_until_cmd_or_err(running_es_pods[0], uuid, pod_log_timeout, task_vars)
+        return {}
+
+    def wait_until_cmd_or_err(self, es_pod, uuid, timeout_secs, task_vars):
+        """Wait a maximum of timeout_secs for the uuid logged in Kibana to be
+        found in the Elasticsearch logs. Since we are querying for a message
+        with the uuid that was set earlier, there should only be a single match.
+        Raise an OpenShiftCheckException if timeout is reached finding a match."""
+        now = int(time.time())
+        time_start = now
+        time_end = time_start + timeout_secs
+        interval = 1  # seconds to wait between retries
+
+        while now < time_end:
+            if self.query_es_from_es(es_pod, uuid, task_vars):
+                return
+
+            time.sleep(interval)
+            now = int(time.time())
+            if now < time_start:  # saw a large clock reset: start the timer again
+                time_start = now
+                time_end = now + timeout_secs
+
+        msg = "expecting match in Elasticsearch for message with uuid {}, but no matches were found after {}s."
+        raise OpenShiftCheckException(msg.format(uuid, timeout_secs))
+
+    def curl_kibana_with_uuid(self, kibanna_pod, task_vars):
+        """curl Kibana with a unique uuid."""
+        uuid = self.generate_uuid()
+        pod_name = kibanna_pod["metadata"]["name"]
+        exec_cmd = "exec {pod_name} -c kibana -- curl --max-time 30 -s http://localhost:5601/{uuid}"
+        exec_cmd = exec_cmd.format(pod_name=pod_name, uuid=uuid)
+
+        error_str = self.oc_cmd(exec_cmd, [], task_vars)
+
+        try:
+            error_code = json.loads(error_str)["statusCode"]
+        except KeyError:
+            msg = ('invalid response returned from Kibana request (Missing "statusCode" key):\n'
+                   'Command: {}\nResponse: {}').format(exec_cmd, error_str)
+            raise OpenShiftCheckException(msg)
+        except ValueError:
+            msg = ('invalid response returned from Kibana request (Non-JSON output):\n'
+                   'Command: {}\nResponse: {}').format(exec_cmd, error_str)
+            raise OpenShiftCheckException(msg)
+
+        if error_code != 404:
+            msg = 'invalid error code returned from Kibana request. Expecting error code "404", but got "{}" instead.'
+            raise OpenShiftCheckException(msg.format(error_code))
+
+        return uuid
+
+    def query_es_from_es(self, es_pod, uuid, task_vars):
+        """curl the Elasticsearch pod and look for a unique uuid in its logs."""
+        pod_name = es_pod["metadata"]["name"]
+        exec_cmd = (
+            "exec {pod_name} -- curl --max-time 30 -s -f "
+            "--cacert /etc/elasticsearch/secret/admin-ca "
+            "--cert /etc/elasticsearch/secret/admin-cert "
+            "--key /etc/elasticsearch/secret/admin-key "
+            "https://logging-es:9200/project.{namespace}*/_count?q=message:{uuid}"
+        )
+        exec_cmd = exec_cmd.format(pod_name=pod_name, namespace=self.logging_namespace, uuid=uuid)
+        result = self.oc_cmd(exec_cmd, [], task_vars)
+
+        try:
+            count = json.loads(result)["count"]
+        except KeyError:
+            msg = 'invalid response from Elasticsearch query:\n"{}"\nMissing "count" key:\n{}'
+            raise OpenShiftCheckException(msg.format(exec_cmd, result))
+        except ValueError:
+            msg = 'invalid response from Elasticsearch query:\n"{}"\nNon-JSON output:\n{}'
+            raise OpenShiftCheckException(msg.format(exec_cmd, result))
+
+        return count
+
+    @staticmethod
+    def running_pods(pods):
+        """Returns: list of pods in a running state"""
+        return [
+            pod for pod in pods
+            if pod['status']['phase'] == 'Running'
+        ]
+
+    @staticmethod
+    def generate_uuid():
+        """Wrap uuid generator. Allows for testing with expected values."""
+        return str(uuid4())
+
+    def oc_cmd(self, cmd_str, extra_args, task_vars):
+        """Wrap parent exec_oc method. Allows for testing without actually invoking other modules."""
+        return super(LoggingIndexTime, self).exec_oc(
+            self.execute_module,
+            self.logging_namespace,
+            cmd_str,
+            extra_args,
+            task_vars
+        )
--- a/roles/openshift_health_checker/test/logging_index_time_test.py
+++ b/roles/openshift_health_checker/test/logging_index_time_test.py
@@ -1,0 +1,183 @@
+import json
+
+import pytest
+
+from openshift_checks.logging.logging_index_time import LoggingIndexTime, OpenShiftCheckException
+
+
+SAMPLE_UUID = "unique-test-uuid"
+
+
+def canned_podlogquerytime(exec_oc=None):
+    """Create a check object with a canned exec_oc method"""
+    check = LoggingIndexTime("dummy")  # fails if a module is actually invoked
+    if exec_oc:
+        check.oc_cmd = exec_oc
+    return check
+
+
+plain_running_elasticsearch_pod = {
+    "metadata": {
+        "labels": {"component": "es", "deploymentconfig": "logging-es-data-master"},
+        "name": "logging-es-data-master-1",
+    },
+    "status": {
+        "containerStatuses": [{"ready": True}, {"ready": True}],
+        "phase": "Running",
+    }
+}
+plain_running_kibana_pod = {
+    "metadata": {
+        "labels": {"component": "kibana", "deploymentconfig": "logging-kibana"},
+        "name": "logging-kibana-1",
+    },
+    "status": {
+        "containerStatuses": [{"ready": True}, {"ready": True}],
+        "phase": "Running",
+    }
+}
+not_running_kibana_pod = {
+    "metadata": {
+        "labels": {"component": "kibana", "deploymentconfig": "logging-kibana"},
+        "name": "logging-kibana-2",
+    },
+    "status": {
+        "containerStatuses": [{"ready": True}, {"ready": False}],
+        "conditions": [{"status": "True", "type": "Ready"}],
+        "phase": "pending",
+    }
+}
+
+
+@pytest.mark.parametrize('pods, expect_pods', [
+    (
+        [not_running_kibana_pod],
+        [],
+    ),
+    (
+        [plain_running_kibana_pod],
+        [plain_running_kibana_pod],
+    ),
+    (
+        [],
+        [],
+    )
+])
+def test_check_running_pods(pods, expect_pods):
+    check = canned_podlogquerytime(None)
+    pods = check.running_pods(pods)
+    assert pods == expect_pods
+
+
+@pytest.mark.parametrize('name, json_response, uuid, timeout, expect_exception, extra_words', [
+    (
+        'invalid json response',
+        {
+            "invalid_field": 1,
+        },
+        SAMPLE_UUID,
+        0.001,
+        OpenShiftCheckException,
+        ["invalid response", "Elasticsearch"],
+    ),
+    (
+        'empty response',
+        {},
+        SAMPLE_UUID,
+        0.001,
+        OpenShiftCheckException,
+        ["invalid response", "Elasticsearch"],
+    ),
+    (
+        'valid count in response',
+        {
+            "count": 1,
+        },
+        SAMPLE_UUID,
+        0.001,
+        None,
+        [],
+    ),
+    (
+        'valid response but invalid match count',
+        {
+            "count": 0,
+        },
+        SAMPLE_UUID,
+        0.005,
+        OpenShiftCheckException,
+        ["expecting match", SAMPLE_UUID, "0.005s"],
+    )
+], ids=lambda argval: argval[0])
+def test_wait_until_cmd_or_err(name, json_response, uuid, timeout, expect_exception, extra_words):
+    def exec_oc(exec_cmd, args, task_vars):
+        return json.dumps(json_response)
+
+    check = canned_podlogquerytime(exec_oc)
+
+    error = ""
+    if expect_exception:
+        with pytest.raises(expect_exception) as error:
+            check.wait_until_cmd_or_err(plain_running_elasticsearch_pod, uuid, timeout, None)
+    else:
+        check.wait_until_cmd_or_err(plain_running_elasticsearch_pod, uuid, timeout, None)
+
+    if not expect_exception:
+        assert not error
+
+    for word in extra_words:
+        assert word in str(error)
+
+
+@pytest.mark.parametrize('name, json_response, uuid, expect_exception, extra_words', [
+    (
+        'invalid json response',
+        {
+            "invalid_field": "invalid",
+        },
+        SAMPLE_UUID,
+        OpenShiftCheckException,
+        ["invalid response returned", 'Missing "statusCode" key'],
+    ),
+    (
+        'wrong error code in response',
+        {
+            "statusCode": 500,
+        },
+        SAMPLE_UUID,
+        OpenShiftCheckException,
+        ["Expecting error code", "500"],
+    ),
+    (
+        'correct response code, found unique id is returned',
+        {
+            "statusCode": 404,
+        },
+        "sample unique id",
+        None,
+        ["sample unique id"],
+    ),
+])
+def test_curl_kibana_with_uuid(name, json_response, uuid, expect_exception, extra_words):
+    def exec_oc(exec_cmd, args, task_vars):
+        return json.dumps(json_response)
+
+    check = canned_podlogquerytime(exec_oc)
+    check.generate_uuid = lambda: uuid
+
+    error = ""
+    result = ""
+    if expect_exception:
+        with pytest.raises(expect_exception) as error:
+            check.curl_kibana_with_uuid(plain_running_kibana_pod, None)
+    else:
+        result = check.curl_kibana_with_uuid(plain_running_kibana_pod, None)
+
+    if not expect_exception:
+        assert not error
+
+    for word in extra_words:
+        if error:
+            assert word in str(error)
+        elif result:
+            assert word in result
--- a/roles/openshift_health_checker/test/logging_check_test.py
+++ b/roles/openshift_health_checker/test/logging_check_test.py
@@ -118,7 +118,7 @@ def test_is_active(groups, logging_deployed, is_active):
     (
         'No resources found.',
         None,
-        'There are no pods in the logging namespace',
+        'No pods were found for the "es"',
     ),
     (
         json.dumps({'items': [plain_kibana_pod, plain_es_pod, plain_curator_pod, fluentd_pod_node1]}),


roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -64,6 +64,17 @@ def not_running_pods(pods):
         ]
 
     @staticmethod
+    def running_pods(pods):
+        """Returns: list of pods in a running state"""
+        return [
+            pod for pod in pods
+            if any(
+                condition['type'] == 'Ready' and condition['status'] == 'True'
+                for condition in pod['status']['conditions']
```
rhcarvalho: I didn't recall the `conditions` field... is Ready==True equivalent to Phase==Running?

[vendor/k8s.io/kubernetes/pkg/api/v1/types.go#L2453-L2461](https://github.com/openshift/origin/blob/a192b3e406444e23ac12b91354fd0c0130e6e2ae/vendor/k8s.io/kubernetes/pkg/api/v1/types.go#L2453-L2461)

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+
+        msg = "expecting match in Elasticsearch for message with uuid {}, but no matches were found after {}s."
+        raise OpenShiftCheckException(msg.format(uid, timeout_secs))
+
+    def curl_kibana_with_uuid(self, kibanna_pod, task_vars):
+        """curl Kibanna with a unique uuid to search for in Elasticsearch."""
+        uid = str(uuid4())
+        pod_name = kibanna_pod["metadata"]["name"]
+        exec_cmd = ("exec {pod_name} -c kibana -- curl --connect-timeout 1 -s "
+                    "http://localhost:5601/{uid}")
```
rhcarvalho: Poor indentation style; maybe we could fit all in one longer line?

rhcarvalho: I suspect the port is not user-configurable in the logging stack installation?

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+                    "--connect-timeout 1 -s -k "
+                    "--cert /etc/elasticsearch/secret/admin-cert "
+                    "--key /etc/elasticsearch/secret/admin-key "
+                    "https://{es_host}:9200/{es_index}/_count?q=message:{uid}")
+        exec_cmd = exec_cmd.format(pod_name=pod_name, es_host=es_host, es_index=es_index, uid=uid)
+
+        result = self._exec_oc(exec_cmd, [], task_vars)
+        return json.loads(result)["count"]
+
+    def _exec_oc(self, cmd_str, extra_args, task_vars):
```
rhcarvalho: Why the underscore?

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if error:
+            return {"failed": True, "changed": False, "msg": error}
+
+        kibanna_pods = super(Podlogs, self).running_pods(pods)
+        if not len(kibanna_pods):
+            return {"failed": True, "changed": False, "msg": ('No Kibanna pods were found to be' 
+                                                              'in the "Running" state.'
+                                                              'At least one Kibana pod is required' 
+                                                              'in order to perform this check.')}
```
rhcarvalho: Poor indentation style.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if error:
+            return {"failed": True, "changed": False, "msg": error}
+
+        es_pods = super(Podlogs, self).running_pods(pods)
+        if not len(es_pods):
+            return {"failed": True, "changed": False, "msg": ('No Elasticsearch pods were found to be'
+                                                              'in the "Running" state.'
+                                                              'At least one Kibana pod is required'
+                                                              'in order to perform this check.')}
```
rhcarvalho: PIS ;)

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+        while timeout_secs > 0:
+            time.sleep(interval)
+            timeout_secs -= interval
+
+            total = self.query_es_from_es(es_pod, es_host, es_index, uid, task_vars)
+            if total == 1:
+                return
+
+        msg = "expecting match in Elasticsearch for message with uuid {}, but no matches were found after {}s."
+        raise OpenShiftCheckException(msg.format(uid, timeout_secs))
```
rhcarvalho: We mutate `timeout_secs`, so at this point its value doesn't fit the message.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+
+        result = self._exec_oc(exec_cmd, [], task_vars)
+        return json.loads(result)["count"]
+
+    def _exec_oc(self, cmd_str, extra_args, task_vars):
+        return super(Podlogs, self).exec_oc(self.execute_module,
+                                            self.logging_namespace,
+                                            cmd_str,
+                                            extra_args,
+                                            task_vars)
```
rhcarvalho: PIS

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def query_es_from_es(self, es_pod, es_host, es_index, uid, task_vars):
+        """curl the Elasticsearch pod and look for the unique uuid that was set
+        as part of the error log from the curl request made in the Kibanna pod."""
+        pod_name = es_pod["metadata"]["name"]
+        exec_cmd = ("exec {pod_name} -- curl "
+                    "--connect-timeout 1 -s -k "
+                    "--cert /etc/elasticsearch/secret/admin-cert "
+                    "--key /etc/elasticsearch/secret/admin-key "
+                    "https://{es_host}:9200/{es_index}/_count?q=message:{uid}")
```
rhcarvalho: PIS

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+    def curl_kibana_with_uuid(self, kibanna_pod, task_vars):
+        """curl Kibanna with a unique uuid to search for in Elasticsearch."""
+        uid = str(uuid4())
+        pod_name = kibanna_pod["metadata"]["name"]
+        exec_cmd = ("exec {pod_name} -c kibana -- curl --connect-timeout 1 -s "
+                    "http://localhost:5601/{uid}")
+        exec_cmd = exec_cmd.format(pod_name=pod_name, uid=uid)
+
+        error_str = self._exec_oc(exec_cmd, [], task_vars)
+        error_code = json.loads(error_str).get("statusCode")
```
rhcarvalho: There can be a `ValueError` here if the return value is not valid JSON.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+import json
+import time
+
+from uuid import uuid4
+
+from openshift_checks import get_var, OpenShiftCheckException
+from openshift_checks.logging.logging import LoggingCheck
+
+
+ES_CMD_TIMEOUT_SECONDS = 30
```
rhcarvalho: This is the source of uncertainty and flakiness of this check :-/

Monitoring the time it takes over time would be much better than a point-in-time check. What is "reasonable amount of time" for one cluster is not for another.

This could be user configurable, but then I also would not like us to end up with too many flags for people to tune.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+ES_DEFAULT_HOST = "logging-es"
+ES_DEFAULT_INDEX = "project.logging*"
+
+
+class Podlogs(LoggingCheck):
+    """Check that logs from ElasticSearch pods can be queried within a reasonable amount of time."""
+
+    name = "podlogs"
```
rhcarvalho: Maybe pod_logs (separate words with underscore as we do for other check names)? Or even something more descriptive/verbose.

Then make the class name a CamelCase version of the name.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
rhcarvalho: We may need to fine tune more flags for `curl` e.g. to cover failures (`-f`).

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+
+            total = self.query_es_from_es(es_pod, es_host, es_index, uid, task_vars)
+            if total == 1:
+                return
+
+        msg = "expecting match in Elasticsearch for message with uuid {}, but no matches were found after {}s."
+        raise OpenShiftCheckException(msg.format(uid, timeout_secs))
+
+    def curl_kibana_with_uuid(self, kibanna_pod, task_vars):
+        """curl Kibanna with a unique uuid to search for in Elasticsearch."""
```
rhcarvalho: Don't mix the matters in the docstring. This is probably unrelated to Elasticsearch, even though in a greater context the latter will be involved.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if error_code != 404:
+            msg = "expecting server error code 404, but got {} instead."
+            raise OpenShiftCheckException(msg.format(error_code))
+
+        return uid
+
+    def query_es_from_es(self, es_pod, es_host, es_index, uid, task_vars):
+        """curl the Elasticsearch pod and look for the unique uuid that was set
+        as part of the error log from the curl request made in the Kibanna pod."""
```
rhcarvalho: The knowledge of how ES and Kib. mix together should be in the caller that coordinates the use of these methods, not in the methods themselves. This method is only about ES. A caller does whatever it wants with it.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,135 @@
...
+        pod_name = es_pod["metadata"]["name"]
+        exec_cmd = ("exec {pod_name} -- curl "
+                    "--connect-timeout 1 -s -k "
```
rhcarvalho: Why do we use `-k` and pass `--cert` and `--key`, are those somehow contradictory?

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
rhcarvalho: I'm surprised that apparently we create an entry with a GET request?!

Will this somehow pollute what users see, e.g. in the Web Console? Are they going to start seeing this mysterious UUIDs in their logs?

roles/openshift_health_checker/openshift_checks/logging/logging.py
juanvallejo: I believe that it can be in the `Running` state, but its endpoint not be reachable unless its `condition.status` is `true` https://github.com/kubernetes/kubernetes/issues/30785#issuecomment-240583200

We do the same sort of check for [not_running_pods](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_health_checker/openshift_checks/logging/logging.py#L61)

but thinking about this a bit more, I think for this method, it should be enough to just check if the pod's `Phase == "Running"`

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
juanvallejo: > This is the source of uncertainty and flakiness of this check :-/
>
> Monitoring the time it takes over time would be much better than a point-in-time check. What is > "reasonable amount of time" for one cluster is not for another.

While I agree, how would we report this information without failing the check? If I understand correctly, messages from a check do not always show up at the end of a successful check.

> This could be user configurable, but then I also would not like us to end up with too many flags for people to tune.

I think at least for now we could try this, as well as set the default time to a much higher value.
I don't think this will become a problem of flakes, as this check is not meant to be run as pre-install or pre-upgrade; I was thinking it'd be more along the lines of the [openshift image data check](https://github.com/openshift/openshift-ansible/pull/3660)

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
juanvallejo: thanks, will address

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
juanvallejo: > I'm surprised that apparently we create an entry with a GET request?!

This is how the script I am basing this off works: https://bugzilla.redhat.com/attachment.cgi?id=1258225

> Will this somehow pollute what users see, e.g. in the Web Console? Are they going to start seeing this mysterious UUIDs in their logs?

I believe this will happen, the 404 request is logged in kibana, and later aggregated into the elasticsearch logs. @sosiouxme wondering your thoughts on this as well

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
juanvallejo: Thanks, I currently catch this in the run method; however will catch `KeyError`s as well

juanvallejo:

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
juanvallejo: @rhcarvalho 
> I suspect the port is not user-configurable in the logging stack installation?

I'm not sure, but did not see any config for them in the [sample inventories](https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.origin.example)

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
rhcarvalho: Sorry, I was just thinking aloud how the stuff we do need not necessarily fit in the "check" format.

Let's start with this and iterate as we get feedback.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
rhcarvalho: > I believe this will happen, the 404 request is logged in kibana, and later aggregated into the elasticsearch logs.

Ingenious trick!

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,143 @@
...
+            self.execute_module,
+            self.logging_namespace,
+            "kibana",
+            task_vars,
+        )
+
+        if error:
+            return {"failed": True, "changed": False, "msg": error}
+
+        kibanna_pods = super(PodLogQueryTime, self).running_pods(pods)
```
rhcarvalho: Why not `self.running_pods`?!

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,143 @@
...
+
+        return {"failed": failed, "msg": msg}
+
+    def wait_until_cmd_or_err(self, es_pod, uid, timeout_secs, task_vars):
+        """Wait a maximum of timeout_secs for the uuid logged in Kibana to be found in
+        the Elasticsearch logs. Since we are querying for a message with the the unique
+        uuid from earlier, there should only be a single match.
+        Raise an OpenShiftCheckException if the timeout is reached before a result is found."""
+
+        interval = 1  # amount of time to wait between retries
```
rhcarvalho: s/amount of time/seconds ?

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,143 @@
...
+
+        while timeout_secs > 0:
+            time.sleep(interval)
+            timeout_secs -= interval
+
+            total = self.query_es_from_es(es_pod, uid, task_vars)
+            if total == 1:
+                return
+
+        msg = "expecting match in Elasticsearch for message with uuid {}, but no matches were found after {}s."
```
rhcarvalho: Eventually we may have slept for more than `orig_timeout`, but that's probably irrelevant, agree?

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,143 @@
...
+        if error_code != 404:
+            msg = "expecting server error code 404, but got {} instead."
+            raise OpenShiftCheckException(msg.format(error_code))
+
+        return uid
+
+    def query_es_from_es(self, es_pod, uid, task_vars):
+        """curl the Elasticsearch pod and look for a unique uuid in its logs."""
+        pod_name = es_pod["metadata"]["name"]
+        exec_cmd = ("exec {pod_name} -- curl --connect-timeout 1 -s -k -f "
```
rhcarvalho: My question remain, why do we use `-k` and `--cert` and `--key` together?

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
juanvallejo: Sure, will update

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
richm: Because es requires client cert auth for admin access.  So you can either use `-k` or `--cacert /etc/elasticsearch/secret/admin-ca`, your choice.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
richm: We use the same trick in logging CI https://github.com/openshift/origin-aggregated-logging/blob/master/deployer/scripts/util.sh#L459
```
# $1 - unique value to search for in es
function add_test_message() {
    local kib_pod=`get_running_pod kibana`
    oc exec $kib_pod -c kibana -- curl --connect-timeout 1 -s \
       http://localhost:5601/$1 > /dev/null 2>&1
}
```

juanvallejo:
Thanks for the feedback, will begin adding tests

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
rhcarvalho: Thanks @richm. I see no reason not to use `--cacert /etc/elasticsearch/secret/admin-ca` then.

As for the `--connect-timeout`, 1 second seems too aggressive.
If we're willing to wait up to 10 minutes retrying until this succeeds, I'd suggest being more tolerant with slow connections.
`--connect-timeout 60`?

juanvallejo: I agree, will increase connect timeout

juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
sosiouxme: Right, it just causes Kibana itself to create a log entry. So it's only logs for Kibana that get polluted (only admins will see).

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+
+ES_CMD_TIMEOUT_SECONDS = 600
+ES_DEFAULT_HOST = "logging-es"
+ES_DEFAULT_INDEX = "project.logging*"
+
+
+class PodLogQueryTime(LoggingCheck):
+    """Check that logs from ElasticSearch pods can be queried within a reasonable amount of time."""
+
+    name = "pod_log_query_time"
```
sosiouxme: Shouldn't generally be querying that takes the time, but rather waiting for the log entry to get there so you can query it. Suggest `name = "logging_index_time"`

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+from openshift_checks.logging.logging import LoggingCheck
+
+
+ES_CMD_TIMEOUT_SECONDS = 600
+ES_DEFAULT_HOST = "logging-es"
+ES_DEFAULT_INDEX = "project.logging*"
+
+
+class PodLogQueryTime(LoggingCheck):
+    """Check that logs from ElasticSearch pods can be queried within a reasonable amount of time."""
```
sosiouxme: Check that pod logs are aggregated and indexed in ElasticSearch within a reasonable amount of time.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+        """Check delay of logs and return any errors accordingly"""
+        error = None
+
+        try:
+            uid = self.curl_kibana_with_uuid(kibana_pod, task_vars)
+            self.wait_until_cmd_or_err(es_pod, uid, pod_log_timeout, task_vars)
+        except OpenShiftCheckException as error:
+            error = str(error)
+        except KeyError as error:
+            error = "Invalid json response from pod: {}".format(str(error))
```
sosiouxme: Looking at this I have no idea where this would originate without hunting through the methods being called. I don't think `check_log_delay` should know what this means; whatever method may have caused it should catch it and rethrow as an exception with a meaningful error message (which pod? what query was made? what was the response that was invalid?).

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+        return {}
+
+    def check_log_delay(self, kibana_pod, es_pod, pod_log_timeout, task_vars):
+        """Check delay of logs and return any errors accordingly"""
+        error = None
+
+        try:
+            uid = self.curl_kibana_with_uuid(kibana_pod, task_vars)
+            self.wait_until_cmd_or_err(es_pod, uid, pod_log_timeout, task_vars)
+        except OpenShiftCheckException as error:
```
sosiouxme: Don't catch this, just let it bubble up.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+            error = str(error)
+        except KeyError as error:
+            error = "Invalid json response from pod: {}".format(str(error))
+        except ValueError as error:
+            error = "Invalid response from pod {}".format(str(error))
```
sosiouxme: same thing here

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+    def running_pods(pods):
+        """Returns: list of pods in a running state"""
+        return [
+            pod for pod in pods
+            if pod['status']['phase'] == 'Running'
+        ]
+
+    def wait_until_cmd_or_err(self, es_pod, uid, timeout_secs, task_vars):
+        """Wait a maximum of timeout_secs for the uuid logged in Kibana to be found in
+        the Elasticsearch logs. Since we are querying for a message with the the unique
```
sosiouxme: the the

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
sosiouxme: Queries *should* be pretty quick but in times of trouble they may not be... probably better if the loop condition was how much time had passed, not how many times through the loop. Would be even better if there were an actual timer set so that if something got stuck it wouldn't just sit there, but that may be too complicated for now.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+                return
+
+        msg = "expecting match in Elasticsearch for message with uuid {}, but no matches were found after {}s."
+        raise OpenShiftCheckException(msg.format(uid, orig_timeout))
+
+    def curl_kibana_with_uuid(self, kibanna_pod, task_vars):
+        """curl Kibana with a unique uuid."""
+        uid = str(self.generate_uuid())
+        pod_name = kibanna_pod["metadata"]["name"]
+        exec_cmd = "exec {pod_name} -c kibana -- curl --connect-timeout 60 -s http://localhost:5601/{uid}"
```
sosiouxme: If it takes 60 seconds to connect something is really wrong... I'd say maybe 5? For that matter, probably want to add `--max-time 15` to the `curl` as well so queries can't get stuck.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+            total = self.query_es_from_es(es_pod, uid, task_vars)
+            if total == 1:
+                return
+
+        msg = "expecting match in Elasticsearch for message with uuid {}, but no matches were found after {}s."
+        raise OpenShiftCheckException(msg.format(uid, orig_timeout))
+
+    def curl_kibana_with_uuid(self, kibanna_pod, task_vars):
+        """curl Kibana with a unique uuid."""
+        uid = str(self.generate_uuid())
```
sosiouxme: just curious, would this ever not already be a string?

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+        if error_code != 404:
+            msg = "expecting server error code 404, but got {} instead."
+            raise OpenShiftCheckException(msg.format(error_code))
+
+        return uid
+
+    def query_es_from_es(self, es_pod, uid, task_vars):
+        """curl the Elasticsearch pod and look for a unique uuid in its logs."""
+        pod_name = es_pod["metadata"]["name"]
+        exec_cmd = ("exec {pod_name} -- curl --connect-timeout 60 -s -k -f "
```
sosiouxme: I don't *think* there is a need for `-k` here, should be able to just specify the CA and have a valid TLS connection... remind me to check into it.

sosiouxme:
Hah, I mostly just totally contradicted or repeated earlier feedback didn't I? Can discuss further...

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
+"""
+Module for ensuring logs from pods can be queried in a reasonable amount of time.
```
rhcarvalho: Funny thing is that in one sense it is a "Python module". I agree with @sosiouxme that in this context we better call it a check.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+            uid = self.curl_kibana_with_uuid(kibana_pod, task_vars)
+            self.wait_until_cmd_or_err(es_pod, uid, pod_log_timeout, task_vars)
+        except OpenShiftCheckException as error:
+            error = str(error)
+        except KeyError as error:
+            error = "Invalid json response from pod: {}".format(str(error))
+        except ValueError as error:
+            error = "Invalid response from pod {}".format(str(error))
+
+        return error
```
rhcarvalho: I agree with @sosiouxme re: exception handling here, and add that we better not mix styles. In Go we pass around errors as values and the common idiom is to have a return value of type error. Not that we couldn't do something similar in Python, but if we are raising exceptions, let's use a single mechanism for communicating errors.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
rhcarvalho: > how much time had passed

The detail of doing that right is that we need a monotonic clock/timer.

The total time per loop iteration should be approx. the time slept of at least `interval` + the time running `curl`.
We can limit `curl` a bit more by setting `--max-time`.

I'm okay if this function can run for (a bit) longer than `orig_timeout`, specially if we lower that value as @sosiouxme proposes.

roles/openshift_health_checker/test/podlogs_test.py
```diff
@@ -0,0 +1,190 @@
...
+        check.oc_cmd = exec_oc
+    return check
+
+
+def assert_error(error, expect_error):
+    if expect_error:
+        assert error
+        assert expect_error in error
+    else:
+        assert not error
```
rhcarvalho: Dead code?

roles/openshift_health_checker/test/podlogs_test.py
```diff
@@ -0,0 +1,190 @@
...
+    )
+])
+def test_check_running_pods(pods, expect_pods):
+    check = canned_podlogquerytime(None)
+    pods = check.running_pods(pods)
+    assert pods == expect_pods
+
+
+@pytest.mark.parametrize('json_response, uuid, timeout, expect_exception, extra_words', [
+    # test with invalid json response
```
rhcarvalho: If these comments are to be useful, they should be part of the output when a test fail...
We should not need to manually match complex data in the output to identify which test case failed.

See https://docs.pytest.org/en/latest/parametrize.html#_pytest.python.Metafunc.parametrize.

I couldn't find a clean example right now, but I can type:

```python
@pytest.mark.parametrize('name,json_response,uuid,timeout,exc,extra_words', [
    (
        'invalid json response',
        ...
     ),
    (
        'empty response',
        ...
    ),
    ...
], ids=lambda argval: argval[0])
def test_...(name, json_response, ...):
    ...
```

We also have one example in `openshift-ansible`, look for `roles/openshift_certificate_expiry/test/conftest.py`.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
juanvallejo: @rhcarvalho @sosiouxme sounds good, updated `curl` commands to have a `connect-timeout` of `30s` and a `max-time` of `60s`. Also, updated total time per loop iteration to take into account total `curl` time as well:https://github.com/openshift/openshift-ansible/pull/4519/commits/29149c2a6d246a1dc91e6e751f9312be74c45cfd

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+import json
+import time
+
+from uuid import uuid4
+
+from openshift_checks import get_var, OpenShiftCheckException
+from openshift_checks.logging.logging import LoggingCheck
+
+
+ES_CMD_TIMEOUT_SECONDS = 600
```
juanvallejo: Sounds good to me

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
```diff
@@ -0,0 +1,174 @@
...
+
+class PodLogQueryTime(LoggingCheck):
+    """Check that logs from ElasticSearch pods can be queried within a reasonable amount of time."""
+
+    name = "pod_log_query_time"
+    tags = ["health", "logging"]
+
+    logging_namespace = "logging"
+    pod_es_host = ES_DEFAULT_INDEX
+    pod_es_index = ES_DEFAULT_HOST
```
juanvallejo: Originally I passed them as arguments to `curl_kibana_with_uuid` and `query_es_from_es` methods, but was running into a pylint  `too-many-arguments` error.

roles/openshift_health_checker/test/podlogs_test.py
juanvallejo: Will change, originally based this on the way we have written every other test so far, however

juanvallejo:
@sosiouxme @rhcarvalho thanks, review comments addressed

roles/openshift_health_checker/test/podlogs_test.py
sosiouxme: I agree it's better to have the comments in the output than in the code, and I think I'm the one that started with the code comments so I'll take the blame.

roles/openshift_health_checker/openshift_checks/logging/podlogs.py
sosiouxme: It's a wild guess to some degree, but this is all intra-cluster networking. It would be a bad sign if connecting to a service IP took more than a second even if your cluster were federated around the world (although it could always be a transient network issue, I think it's much more likely to be something more significant). The actual retrieval of the request hits storage and processing so I could see a little more leeway on that, agree that it's a good idea to just use `--max-time` to cover the whole operation.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,175 @@
...
+            time.sleep(interval)
+            timeout_secs -= interval
+
+            time_start = int(time.time())
+            total = self.query_es_from_es(es_pod, uid, task_vars)
+            time_end = int(time.time())
+            if total == 1:
+                return
+
+            timeout_secs -= (time_end - time_start)
```
juanvallejo: Copying discussion from 29149c2a6d246a1dc91e6e751f9312be74c45cfd:

@rhcarvalho 
> This is buggy, because time.time() is not monotonic. time_end - time_start may be negative, the clock may have 
> changed between the two calls.
> See https://github.com/golang/proposal/blob/master/design/12914-monotonic.md#computer-clocks.
>
> We may decide to ignore the bug, but then we could as well ignore the time it takes to run query_es_from_es and 
> know that timeout_secs is not to be taken with precision.

@juanvallejo
> Thanks for the link to that proposal.
> What if we keep the time it takes to run query_es_from_es, but only count it against timeout_secs whenever 
> (time_end - time_start) > -1? That way, we keep a precise duration for timeout_secs 99...% of the time

juanvallejo:
@rhcarvalho @sosiouxme wondering if there are any further comments on this? Thanks!

juanvallejo:
@sosiouxme thanks for the review, comments addressed

juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,161 @@
...
+
+    logging_namespace = "logging"
+
+    def run(self, tmp, task_vars):
+        """Add log entry by making unique request to Kibana. Check for unique entry in the ElasticSearch pod logs."""
+        pod_log_timeout = int(get_var(task_vars, "openshift_check_pod_logs_timeout", default=ES_CMD_TIMEOUT_SECONDS))
+
+        # get all Kibana pods
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default=self.logging_namespace)
+        pods, error = super(LoggingIndexTime, self).get_pods_for_component(
```
sosiouxme: Actually if there are no Kibana pods running, this comes back with:

`There are no pods in the logging namespace. Is logging deployed?`

Which is sort of a universal problem with using this method; if any component is completely down it says there are no pods in the whole namespace. Originally I think it got all the pods and then sorted them out by component and that error really meant there were no pods. Now it's misleading. Could the message be adjusted or handled differently?

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,161 @@
...
+            return {"failed": True, "changed": False, "msg": error}
+
+        running_kibana_pods = self.running_pods(pods)
+        if not len(running_kibana_pods):
+            msg = ('No Kibana pods were found to be in the "Running" state.'
+                   'At least one Kibana pod is required in order to perform this check.')
+            return {"failed": True, "changed": False, "msg": msg}
+
+        # get all Elasticsearch pods
+        pods, error = super(LoggingIndexTime, self).get_pods_for_component(
```
sosiouxme: same issue here

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,161 @@
...
+        if error:
+            return {"failed": True, "changed": False, "msg": error}
+
+        running_es_pods = self.running_pods(pods)
+        if not len(running_es_pods):
+            msg = ('No Elasticsearch pods were found to be in the "Running" state.'
+                   'At least one Elasticsearch pod is required in order to perform this check.')
+            return {"failed": True, "changed": False, "msg": msg}
+
+        uid = self.curl_kibana_with_uuid(running_kibana_pods[0], task_vars)
```
sosiouxme: while I'm at it... is there some reason not to call this `uuid`?

`uid` is user ID, `uuid` is universally unique ID.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,161 @@
...
+        self.wait_until_cmd_or_err(running_es_pods[0], uid, pod_log_timeout, task_vars)
+        return {}
+
+    def wait_until_cmd_or_err(self, es_pod, uid, timeout_secs, task_vars):
+        """Wait a maximum of timeout_secs for the uuid logged in Kibana to be found in
+        the Elasticsearch logs. Since we are querying for a message with the unique
```
sosiouxme: alright, super nit but... "unique" is redundant when describing a uuid.

juanvallejo: :)

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
juanvallejo: Thanks, updated the error returned by `get_pods_for_component` to mention the logging component instead, and added further clarification when handling the error in the check.

juanvallejo:
@sosiouxme Thanks, review comments addressed

sosiouxme:
@juanvallejo Can't believe I'm just noticing this but:

https://github.com/juanvallejo/openshift-ansible/blob/18715362c46e21b044f5edbecefd9b7f32bae85b/roles/openshift_health_checker/openshift_checks/logging/logging.py#L23

I don't think the default is to deploy logging. So all these checks would be running even if logging is not deployed. Could you change that default to false with this PR? Then I think we're good to go.

juanvallejo:
@sosiouxme ping

sosiouxme:
aos-ci-test

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,163 @@
...
+            raise OpenShiftCheckException(msg.format(exec_cmd, result))
+        except ValueError:
+            msg = 'invalid response from Elasticsearch query:\n"{}"\nNon-JSON output:\n{}'
+            raise OpenShiftCheckException(msg.format(exec_cmd, result))
+
+        return count
+
+    @staticmethod
+    def running_pods(pods):
+        """Returns: list of pods in a running state"""
```
rhcarvalho: "Filter out pods that are not running" ?
"Filter pods that are running"

Or at least s/Returns:/Return a

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,163 @@
...
+        return [
+            pod for pod in pods
+            if pod['status']['phase'] == 'Running'
+        ]
```
rhcarvalho: Could be in one line?

roles/openshift_health_checker/test/logging_index_time_test.py
```diff
@@ -0,0 +1,183 @@
...
+
+import pytest
+
+from openshift_checks.logging.logging_index_time import LoggingIndexTime, OpenShiftCheckException
+
+
+SAMPLE_UUID = "unique-test-uuid"
+
+
+def canned_podlogquerytime(exec_oc=None):
```
rhcarvalho: The name here is reminiscent from a recent past... isn't it outdated?

roles/openshift_health_checker/test/logging_index_time_test.py
```diff
@@ -0,0 +1,183 @@
...
+        result = check.curl_kibana_with_uuid(plain_running_kibana_pod, None)
+
+    if not expect_exception:
+        assert not error
+
+    for word in extra_words:
+        if error:
+            assert word in str(error)
+        elif result:
+            assert word in result
```
rhcarvalho: I'd find it easier to understand if "positive" and "negative" tests were separate.

Here, it is unclear what assertions are actually made and, more importantly, what do they mean. Maybe it's just me :/

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,163 @@
...
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default=self.logging_namespace)
+        pods, error = super(LoggingIndexTime, self).get_pods_for_component(
+            self.execute_module,
+            self.logging_namespace,
+            "kibana",
+            task_vars,
+        )
+
+        if error:
+            msg = 'Unable to retrieve pods for the "Kibanna" logging component: {}'
```
stevekuznetsov: `s/Kibanna/Kibana/`

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,163 @@
...
+            msg = ('No Elasticsearch pods were found to be in the "Running" state.'
+                   'At least one Elasticsearch pod is required in order to perform this check.')
+            return {"failed": True, "changed": False, "msg": msg}
+
+        uuid = self.curl_kibana_with_uuid(running_kibana_pods[0], task_vars)
+        self.wait_until_cmd_or_err(running_es_pods[0], uuid, pod_log_timeout, task_vars)
+        return {}
+
+    def wait_until_cmd_or_err(self, es_pod, uuid, timeout_secs, task_vars):
+        """Wait a maximum of timeout_secs for the uuid logged in Kibana to be
```
juanvallejo: > We could really simplify this by simply stating a simpler algorithm: "retry N times with S seconds interval", and never call time.time().

Not sure I follow your thinking here. The reason why I'm using `time.time()` is to ensure that delay between log creation and log aggregation does not exceed an overall time constraint.

juanvallejo:
Did not mean to delete branch - re-opening here: https://github.com/openshift/openshift-ansible/pull/4682

PR #4498, opened 2017-06-19
Verify matched openshift_upgrade_nodes_label

mtnbikenc:
Verifies the provided label matches a set of hosts prior to upgrading.
If the label didn't match hosts, the upgrade would silently proceed with
upgrading all nodes given the logic for creating the oo_nodes_to_upgrade
group.

See bug: https://bugzilla.redhat.com/show_bug.cgi?id=1457914

PR #4522, opened 2017-06-21
Remove package_update from install playbook

rhcarvalho:
We observed in our CI environment that this check can't guarantee that
an install would fail when the check fails, thus unless we can make its
output match the fate of an install we shall keep it disabled.

Example PRs that failed to merge:
- https://github.com/openshift/openshift-ansible/pull/4498

### DIFF
--- a/playbooks/byo/openshift-cluster/config.yml
+++ b/playbooks/byo/openshift-cluster/config.yml
@@ -16,7 +16,6 @@
       - disk_availability
       - memory_availability
       - package_availability
-      - package_update
       - package_version
       - docker_image_availability
       - docker_storage


rhcarvalho:
@sosiouxme do we have any indications that `yum update` would have succeeded (is it even a "must" during the install)?

If the check can't give consistent results I'd argue for even removing it completely.

rhcarvalho:
aos-ci-test

sosiouxme:
@rhcarvalho we don't do an actual blanket `yum update` in the install AFAICS. It's there for the "avoid RPM hell" part of https://trello.com/c/ljdnBTLi because if `yum update` would fail it's a likely indicator that your repos include something weird (like HTB content) or you have something baked into your install image that will cause problems. I don't necessarily agree that check failure must be 100% indication of install failure; they're just red flags for evaluation.

That said, putting it in CI does introduce flakiness because yum repos change and can be inconsistent. A full install against the CI config apparently ran fine, so this particular failure was not material. A `yum update` ran fine too -- don't know if it's because the repo became consistent again or because there's something different to this logic (it's just using the yum libs to do exactly what yum would, theoretically). It seems we're crossing that line from informing to annoying, so perhaps it's best for now to take it out of the install path (but leave it in the pre-install playbook).

rhcarvalho:
@sdodson FYI
Perhaps this can be merged out of the queue?

Issue #4438, opened 2017-06-13
failure when running against Fedora 26 with python3 


dustymabe:
When running against Fedora 26 Atomic Host and using the python3 interpreter I get failures from the attemtps to run the seboolean module to change things. This is related to an [intentional change](https://bugzilla.redhat.com/show_bug.cgi?id=1460818) that was made in selinux. Thus the ansible module needs to be updated ([upstream issue](https://github.com/ansible/ansible/issues/25651)). We are going to have to deal with this in openshift-ansible because we know users are going to be executing from older versions of ansible against newer hosts with new libselinux-python3. 

Let's workaround this if possible.

Here is an example failure from dc12c4b

<details>

```
TASK [openshift_node : Set seboolean to allow nfs storage plugin access from containers] **************************************************************************************************************************************************************************************
task path: /guests/sharedfolder/code/github.com/openshift/openshift-ansible/roles/openshift_node/tasks/storage_plugins/nfs.yml:16
Using module file /usr/lib/python2.7/site-packages/ansible/modules/system/seboolean.py
<45.55.51.130> ESTABLISH SSH CONNECTION FOR USER: root
<45.55.51.130> SSH: EXEC ssh -C -o ControlMaster=auto -o ControlPersist=60s -o StrictHostKeyChecking=no -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=root -o ConnectTimeout=10 -o ControlPath=/home/dustymabe/.ansible/cp/8d524dc87e 45.55.51.130 '/bin/sh -c '"'"'/usr/bin/python3 && sleep 0'"'"''
<45.55.51.130> (1, '\n{"msg": "SELinux boolean virt_use_nfs does not exist.", "failed": true, "invocation": {"module_args": {"persistent": true, "state": true, "name": "virt_use_nfs"}}}\n', '')
[...]

PR #4531, opened 2017-06-21
Workaround #4438 by replacing seboolean module with command.

abutcher:
~This is kind of junky. I can't think of a clean way to identify that the python3 interpreter is being used and only workaround in that case.~

https://github.com/openshift/openshift-ansible/issues/4438

### DIFF
--- a/roles/openshift_node/tasks/storage_plugins/glusterfs.yml
+++ b/roles/openshift_node/tasks/storage_plugins/glusterfs.yml
@@ -3,30 +3,52 @@
   package: name=glusterfs-fuse state=present
   when: not openshift.common.is_atomic | bool
 
-- name: Check for existence of virt_use_fusefs seboolean
-  command: getsebool virt_use_fusefs
-  register: virt_use_fusefs_output
-  when: ansible_selinux and ansible_selinux.status == "enabled"
+- name: Check for existence of fusefs sebooleans
+  command: getsebool {{ item }}
+  register: fusefs_getsebool_status
+  when:
+  - ansible_selinux
+  - ansible_selinux.status == "enabled"
   failed_when: false
   changed_when: false
+  with_items:
+  - virt_use_fusefs
+  - virt_sandbox_use_fusefs
 
 - name: Set seboolean to allow gluster storage plugin access from containers
   seboolean:
-    name: virt_use_fusefs
+    name: "{{ item.item }}"
     state: yes
     persistent: yes
-  when: ansible_selinux and ansible_selinux.status == "enabled" and virt_use_fusefs_output.rc == 0
+  when:
+  - ansible_selinux
+  - ansible_selinux.status == "enabled"
+  - item.rc == 0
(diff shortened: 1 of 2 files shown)


dustymabe:
> I can't think of a clean way to identify that the python3 interpreter is being used and only workaround in that case.

can you use any of this information? 

```
$ ansible localhost -m setup -e 'ansible_python_interpreter=/usr/bin/python3' | grep python
 [WARNING]: provided hosts list is empty, only localhost is available
            "_": "/usr/bin/python3"
        "ansible_python": {
            "executable": "/usr/bin/python3", 
            "type": "cpython", 
        "ansible_python_version": "3.5.3", 
```

would still be nice to only apply this change to platforms that are affected. (which would mean leaving the seboolean module call in place for ones that aren't). If we can use the information above to check for python3 then maybe we just do this when python3 is being used?? WDYT?

abutcher:
@dustymabe `ansible_python_version` will do the trick. I'll workshop this a bit but I agree that it would be nice to only apply the workaround when >= python3.

dustymabe:
> I'll workshop this a bit

i can test it out as soon as you have something

dustymabe:
so this PR worked as is - any chance of getting an updated version I can test?

abutcher:
Sorry for the delay @dustymabe we've got a lot of bugs to squash atm.

dustymabe:
> Sorry for the delay @dustymabe we've got a lot of bugs to squash atm.

no worries - i've been in open source a while so I know how patience works. I mostly just want to have this in place for fedora 26 release, which is july 11th i think .

abutcher:
@dustymabe Pushed an update.

I also hit an issue with the systemd units while I was testing. We'll definitely need a fix for it as well but just mentioning here. Looks like we'll need to force a daemon reload after the units are initially created, maybe by flushing handlers earlier in the role.

```
TASK [openshift_node : Start and enable openvswitch service] **************************************************************************************************************************************************************************
fatal: [atomicf26.abutcher.com]: FAILED! => {
    "changed": false, 
    "failed": true
}

MSG:

Could not find the requested service openvswitch.service: host
```

dustymabe:
@abutcher - thanks - i'll test it out - i didn't hit the issue with systemd units last time I ran things - maybe it is a recent change?

dustymabe:
rebased this PR on latest upstream master at the time (b7104cf) and ran a test on both f25 atomic host and f26 atomic host. works like a charm 

+1

abutcher:
aos-ci-test

abutcher:
[merge]

abutcher:
[merge][severity: bug]

sdodson:
[merge][severity: bug]

zmhassan:
I'm running into same issues here in fedora 27

dustymabe:
> I'm running into same issues here in fedora 27

use python3?

michaelgugino:
@zmhassan 

Please open a new issue report with specific details.

PR #4532, opened 2017-06-21
bugzilla:1463577 - Fix for dynamic pvs when using storageclasses.

kwoodson:


roles/openshift_logging_elasticsearch/templates/pvc.j2
```diff
@@ -25,3 +25,6 @@ spec:
   resources:
     requests:
       storage: {{size}}
+{% if storage_class_name is defined %}
+  storageClassName: ""
```
gnufied: shouldn't this be `storage_class_name` variable rather than being set to `""` ?

kwoodson:
@gnufied @ewolinetz,

This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1463577

When doing dynamic provisioning using storageclasses, the annotation is no longer needed.  According to the docs here: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#aws the option is deprecated:
```In the past, the annotation volume.beta.kubernetes.io/storage-class was used instead of the storageClassName attribute. This annotation is still working, however it will become fully deprecated in a future Kubernetes release.```

This PR makes the assumption that we are using storageclasses.  When we are not or the `openshift_logging_elasticsearch_pvc_dynamic` is set to `false` then we will disable the storageclass inside of the pvc by setting `storageClassName: ""`.  This will cause the pvc to look for a static pv.

I went ahead and did the metrics as well as logging as the same pattern was followed.

File: roles/openshift_logging_elasticsearch/templates/pvc.j2 (hunk 1 above)
Comment (kwoodson) [inline roles/openshift_logging_elasticsearch/templates/pvc.j2]: @gnufied, since we are setting it to `""` only when static it doesn't really matter.  I went ahead and updated that piece of code in case we want it.

**Comment by gnufied** [comment]
lgtm

**Comment by kwoodson** [comment]
@ewolinetz, I tested with these changes for both metrics and logging on 3.6.121.  The pvcs were bound:
```
NAME                  STATUS    VOLUME                                     CAPACITY   ACCESSMODES   STORAGECLASS   AGE
logging-es-0          Bound     pvc-e0d76a90-56b8-11e7-b598-0e5b70ec676a   200Gi      RWO           ebs            3m
logging-es-1          Bound     pvc-e99009c9-56b8-11e7-91fa-0e0370a7dcbc   200Gi      RWO           ebs            2m
metrics-cassandra-1   Bound     pvc-4a5944ad-56b9-11e7-8a33-0e078422088e   100Gi      RWO           ebs            27s
metrics-cassandra-2   Bound     pvc-4b2b08e3-56b9-11e7-8a33-0e078422088e   100Gi      RWO           ebs            26s
```

ewolinetz:
@kwoodson so just to make sure im understanding -- instead of the annotations as before for dynamic PVCs, we now specify the storage class?

Should we be passing the storage class in for the dynamic pvcs then? It looks like we're only setting it to "" for the static (pv backed) pvcs.

gnufied:
@ewolinetz to use default storageclasses, we don't need to specify anything in PVC. By default if no storageclass field is present in the PVC, the PVC will use default storageclass.

Setting StorageClass to `""` is only necessary when we want to force PVC to not use default storageclass and want to statically provision PV for the PVC.

ewolinetz:
@gnufied right, I guess my question is more wondering if we should be passing in these values for when we are creating dynamic PVCs as well. My understanding was this replaces the annotation so to me that would mean the dynamic PVCs would require the parameter

kwoodson:
@ewolinetz, I think there was a change in behavior.  If you do nothing, then you get the default storage class.  If you put something in there, you need to match the correct storage class.

For static pvs, it requires you to disable the storageClassName attribute.  We accomplished this by putting `""` as the value.

For dynamic pvs, nothing is required other than the definition of your storageclass.  The annotation is deprecated as I think the behavior changed to use storageclasses by default.  If we wanted to pass in the name of the storage class, that is possible as well.  That would cover both cases because it would continue to work in the dynamic case.  If you needed static then you would need to pass `""` still.

The only thing that is required is getting the `storageClassName` into the openshift_logging role and then correctly passing it into the other places where it is used, openshift_default_storage_class and openshift_logging_elasticsearch, and then setting it properly in the pvc.j2.  Defaulting it to blank has the same effect.  Requiring this parameter requires more maintenance by passing it and ensuring the default doesn't change when the storage class is created.

Its much simpler to default it to empty so it works without explicitly specifying the `storageClassName`.

sdodson:
aos-ci-test

sdodson:
[test]

ewolinetz:
[test]

sdodson:
[merge][severity:blocker]

PR #4542, opened 2017-06-22
Moving checks down after required initialization happens.

kwoodson:


kwoodson:
@mtnbikenc, @ashcrow,

The healthchecks were added before group initialization was occurring.  This in turn breaks the health checks as they rely on other roles and steps which require our `oo_` groups to be defined.  This PR pushes the checks down two steps further so that group initialization can occur properly.

ashcrow:
Included in  #4495

sdodson:
aos-ci-test

PR #4548, opened 2017-06-22
releases: enable build/push with multiple tags

sosiouxme:
Update the release scripts so that either at build or push time we can specify multiple tags we want the same image to end up with.


sosiouxme:
@sdodson given that our one adhoc jenkins job is AFAIK the only consumer of these release scripts, how would you feel about just merging this... or letting me mark it as a blocker to get through CI?

/cc @brenton @codificat

sdodson:
That's fine with me.

PR #4551, opened 2017-06-22
Adding checks to protect data loss from bz1463393.

kwoodson:


kwoodson:
@sdodson, as per our discussion on [bz1463393](https://bugzilla.redhat.com/show_bug.cgi?id=1463393), this is a start on what we wanted.

Please review and feel free to correct me as I'm not 100% sure if this fixes the issue.

sdodson:
Looks like this would probably work. rpm_upgrade.yml doesn't do anything to restart the service but systemd_units.yml does so just add a restart task for the rpm variant?

kwoodson:
@sdodson I realized I can't do all of the version checks in that way.  I need to or them and then and them together.  I'll work on this a bit more.

sdodson:
Also, given we ship from different branches you can include only the
relevant versions for that branch if that makes it easier. It's just be a
Delta between branches.

kwoodson:
@sdodson, probably need some assistance testing this.

playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml
```diff
@@ -273,6 +273,44 @@
...
+      - name: Update the rpm before we proceed
+        include_role:
+          name: roles/openshift_node_upgrade
+          tasks_from: rpm_upgrade.yml
+        vars:
+          component: node
+
+      # TODO: ansible 2.4 added systemd support for daemon_reload
+      - name: reload systemd unit
+        command: systemctl daemon-reload
```
gnufied: Looks like `rpm_upgrade.yaml` also reloads systemd unit files. Do we need to this separately here?

gnufied:
Mostly looks good to me.  I agree with @sdodson that since we ship openshift-ansible from individual branches, we can perhaps simplify the check. That would require creating separate PRs against each branch though.

Also, we don't need this for 3.6 iirc.

sdodson:
> Also, we don't need this for 3.6 iirc.

If they're upgrading from a bad 3.5 node version directly to 3.6 we do.

sdodson:
We've decided to handle this via release notes documentation rather than making our upgrade playbooks even more complicated than they are today.

PR #4558, opened 2017-06-23
migrate embedded etcd

ingvagabund:
- if the etcd is embedded, re-attach the leases through locally running etcd daemon
- refactor etcd_common role to set various variables propertly for embedded etcd

### DIFF
--- a/playbooks/common/openshift-etcd/migrate.yml
+++ b/playbooks/common/openshift-etcd/migrate.yml
@@ -3,6 +3,20 @@
   tags:
   - always
 
+- name: Check masters for embedded etcd
+  hosts: localhost
+  connection: local
+  become: no
+  gather_facts: no
+  tags:
+  - always
+  tasks:
+  - fail:
+      msg: "Migration of clustered embedded etcd instances is not supported. Currently there are {{ groups.oo_masters_to_config | length }} masters."
+    when:
+    - groups.oo_masters_to_config | length != 1
+    - groups.oo_etcd_to_config | default([]) | length == 0
+
 - name: Run pre-checks
   hosts: oo_etcd_to_migrate
   tags:
--- a/roles/etcd_common/defaults/main.yml
+++ b/roles/etcd_common/defaults/main.yml
@@ -14,7 +14,7 @@ r_etcd_common_embedded_etcd: false
 r_etcd_common_etcdctl_command: "{{ 'etcdctl' if r_etcd_common_etcd_runtime == 'host' or r_etcd_common_embedded_etcd | bool else 'docker exec etcd_container etcdctl' if r_etcd_common_etcd_runtime == 'docker' else 'runc exec etcd etcdctl' }}"
 
 # etcd server vars
-etcd_conf_dir: '/etc/etcd'
+etcd_conf_dir: "{{ '/etc/origin/master' if r_etcd_common_embedded_etcd | bool else '/etc/etcd' }}"
 r_etcd_common_system_container_host_dir: /var/lib/etcd/etcd.etcd
 etcd_system_container_conf_dir: /var/lib/etcd/etc
 etcd_conf_file: "{{ etcd_conf_dir }}/etcd.conf"
@@ -22,8 +22,8 @@ etcd_ca_file: "{{ etcd_conf_dir }}/ca.crt"
 etcd_cert_file: "{{ etcd_conf_dir }}/server.crt"
 etcd_key_file: "{{ etcd_conf_dir }}/server.key"
 etcd_peer_ca_file: "{{ etcd_conf_dir }}/ca.crt"
-etcd_peer_cert_file: "{{ etcd_conf_dir }}/peer.crt"
-etcd_peer_key_file: "{{ etcd_conf_dir }}/peer.key"
+etcd_peer_cert_file: "{{ etcd_conf_dir + '/admin.crt' if r_etcd_common_embedded_etcd | bool else etcd_conf_dir + '/peer.crt' }}"
+etcd_peer_key_file: "{{ etcd_conf_dir + '/admin.key' if r_etcd_common_embedded_etcd | bool else etcd_conf_dir + '/peer.key' }}"
 
 # etcd ca vars
 etcd_ca_dir: "{{ etcd_conf_dir}}/ca"
@@ -55,7 +55,7 @@ etcd_is_thirdparty: False
 etcd_data_dir: "{{ '/var/lib/origin/openshift.local.etcd' if r_etcd_common_embedded_etcd | bool else '/var/lib/etcd/' if openshift.common.etcd_runtime != 'runc' else '/var/lib/etcd/etcd.etcd/' }}"
 
 # etcd ports and protocols
-etcd_client_port: 2379
-etcd_peer_port: 2380
+etcd_client_port: "{{ '4001' if r_etcd_common_embedded_etcd | bool else '2379' }}"
+etcd_peer_port: "{{ '7001' if r_etcd_common_embedded_etcd | bool else '2380' }}"
 etcd_url_scheme: http
 etcd_peer_url_scheme: http
--- a/roles/etcd_migrate/tasks/migrate.yml
+++ b/roles/etcd_migrate/tasks/migrate.yml
@@ -3,10 +3,13 @@
 - set_fact:
     l_etcd_service: "{{ 'etcd_container' if openshift.common.is_containerized else 'etcd' }}"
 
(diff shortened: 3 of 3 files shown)


roles/etcd_migrate/tasks/migrate.yml
```diff
@@ -20,34 +23,62 @@
...
+# Then, stop the etcd daemon (so the embedded etcd can start again later)
+- block:
+  - name: Enable etcd member (for embedded etcd)
+    service:
+      name: etcd
+      state: started
+  - name: Re-introduce leases (as a replacement for key TTLs) for embedded etcd
+    command: >
+      oadm migrate etcd-ttl \
+      --etcd-address 'http://localhost:2379' \
```
ingvagabund: `2379` is for etcd running locally so the port must be set explicitly.

ingvagabund:
aos-ci-test

roles/etcd_migrate/tasks/migrate.yml
```diff
@@ -20,34 +23,62 @@
...
+    with_items:
+    - "/kubernetes.io/events"
+    - "/kubernetes.io/masterleases"
+  when:
+  - not r_etcd_common_embedded_etcd | bool
+
+# For embeddded etcd, start the etcd daemon locally (to re-attach leases).
+# Then, stop the etcd daemon (so the embedded etcd can start again later)
+- block:
+  - name: Enable etcd member (for embedded etcd)
```
ingvagabund: Still not highly confident with this approach. As the assumption `etcd binary is installed on master when etcd is embedded` can stop holding silently.

ingvagabund: Plus, is this going to work at all? The etcd daemon is managing the embedded etcd's working directory without any changes to its membership. Just basically changing/adding some keys. I don't see any reason why the embedded etcd inside the master should not continue running properly after the etcd daemon is stopped.

ingvagabund:
aos-ci-test

playbooks/common/openshift-etcd/migrate.yml
```diff
@@ -0,0 +1,134 @@
...
+- name: Check masters for embedded etcd
+  hosts: localhost
+  connection: local
+  become: no
+  gather_facts: no
+  tags:
+  - always
+  tasks:
+  - fail:
+      msg: "Migration of clustered embedded etcds is not supported. Currently there are {{ groups.oo_masters_to_config | length }} masters."
```
sdodson: Yeah lets get that fixed.

roles/etcd_migrate/tasks/migrate.yml
```diff
@@ -20,34 +23,66 @@
...
+      ETCDCTL_API: 3
+    with_items:
+    - "/kubernetes.io/events"
+    - "/kubernetes.io/masterleases"
+    delegate_to: "{{ groups.oo_first_master[0] }}"
+    when:
+    # run the re-attachment only once per all etcd members
+    - inventory_hostname in groups.oo_etcd_to_migrate[0]
+  when:
+  - not r_etcd_common_embedded_etcd | bool
```
sdodson: `run_once: true` ? I think we only need to do this against the cluster once right?

roles/etcd_migrate/tasks/migrate.yml
sdodson: Could we just start the master and then connect to etcd on 4001 to perform the TTL migration?

sdodson:
Since we're blocked on having etcdctl 3.2 for embedded migration can we separate the fixes for dedicated into a separate PR? We'll push handling of embedded migration out to post 3.6 GA.

ingvagabund:
Once https://github.com/openshift/openshift-ansible/pull/4623 merges, I will rebase the PR.

roles/etcd_migrate/tasks/migrate.yml
ingvagabund: We could but once the master starts running it can write new events into v3 data store and running the lease attachment can overwrite the new keys. It should not cause much trouble thought. @smarterclayton ? Any thoughts on that? If I change TTL of new events, will it cause any trouble?

roles/etcd_migrate/tasks/migrate.yml
ingvagabund: Yeah, good point. Updated in https://github.com/openshift/openshift-ansible/pull/4623

ingvagabund:
Just rebasing

playbooks/common/openshift-etcd/migrate.yml
```diff
@@ -3,6 +3,20 @@
...
+- name: Check masters for embedded etcd
+  hosts: localhost
+  connection: local
+  become: no
+  gather_facts: no
+  tags:
+  - always
+  tasks:
+  - fail:
+      msg: "Migration of clustered embedded etcd instances is not supported. Currently there are {{ groups.oo_masters_to_config | length }} masters."
```
ashcrow: nit: This error message would confuse me. Maybe add something like "There can only be one $deployment_type master when migrating etcd. Currently there.."

roles/etcd_common/defaults/main.yml
```diff
@@ -14,16 +14,16 @@ r_etcd_common_embedded_etcd: false
 r_etcd_common_etcdctl_command: "{{ 'etcdctl' if r_etcd_common_etcd_runtime == 'host' or r_etcd_common_embedded_etcd | bool else 'docker exec etcd_container etcdctl' if r_etcd_common_etcd_runtime == 'docker' else 'runc exec etcd etcdctl' }}"
 
 # etcd server vars
-etcd_conf_dir: '/etc/etcd'
+etcd_conf_dir: "{{ '/etc/origin/master' if r_etcd_common_embedded_etcd | bool else '/etc/etcd' }}"
```
ashcrow: Would this be something to change to a variable in the future?

roles/etcd_common/defaults/main.yml
```diff
@@ -14,16 +14,16 @@ r_etcd_common_embedded_etcd: false
...
 etcd_system_container_conf_dir: /var/lib/etcd/etc
 etcd_conf_file: "{{ etcd_conf_dir }}/etcd.conf"
 etcd_ca_file: "{{ etcd_conf_dir }}/ca.crt"
 etcd_cert_file: "{{ etcd_conf_dir }}/server.crt"
 etcd_key_file: "{{ etcd_conf_dir }}/server.key"
 etcd_peer_ca_file: "{{ etcd_conf_dir }}/ca.crt"
-etcd_peer_cert_file: "{{ etcd_conf_dir }}/peer.crt"
-etcd_peer_key_file: "{{ etcd_conf_dir }}/peer.key"
+etcd_peer_cert_file: "{{ etcd_conf_dir + '/admin.crt' if r_etcd_common_embedded_etcd | bool else etcd_conf_dir + '/peer.crt' }}"
+etcd_peer_key_file: "{{ etcd_conf_dir + '/admin.key' if r_etcd_common_embedded_etcd | bool else etcd_conf_dir + '/peer.key' }}"
```
ashcrow: Same variable question for the above 2 lines.

roles/etcd_common/defaults/main.yml
```diff
@@ -55,7 +55,7 @@ etcd_is_thirdparty: False
 etcd_data_dir: "{{ '/var/lib/origin/openshift.local.etcd' if r_etcd_common_embedded_etcd | bool else '/var/lib/etcd/' if openshift.common.etcd_runtime != 'runc' else '/var/lib/etcd/etcd.etcd/' }}"
 
 # etcd ports and protocols
-etcd_client_port: 2379
-etcd_peer_port: 2380
+etcd_client_port: "{{ '4001' if r_etcd_common_embedded_etcd | bool else '2379' }}"
+etcd_peer_port: "{{ '7001' if r_etcd_common_embedded_etcd | bool else '2380' }}"
```
ashcrow: Same variable question for the above 2 lines.

ingvagabund:
I will rebase and prepare the PR for merging once we have etcd version inside a container synced with etcd version installed on a host.

ashcrow:
@ingvagabund I believe I was speaking to having ``defaults`` which would be different based on variables. I found it confusing and would expect the defaults to be modified in playbooks.

ingvagabund:
The v2->v3 migration of an embedded etcd is depricated. Instead, one needs to run:
1. `playbooks/byo/openshift-etcd/embedded2external.yml` to migrate the embedded etcd to an external one
2. then `playbooks/byo/openshift-etcd/migrate.yml` to migrate the v2 data to v3 data

PR #4561, opened 2017-06-23
Set clean install and etcd storage on first master to fix scaleup

sdodson:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1464011

In scaleup scenarios we need to explicitly gather this state from the first master as it wouldn't be included in oo_masters_to_config. For non scaleup scenarios, ie: re-running config.yml that's the correct host to use anyway.

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -20,25 +20,6 @@
     - node
     - .config_managed
 
-  - name: Check for existing configuration
-    stat:
-      path: /etc/origin/master/master-config.yaml
-    register: master_config_stat
-
-  - name: Set clean install fact
-    set_fact:
-      l_clean_install: "{{ not master_config_stat.stat.exists | bool }}"
-
-  - name: Determine if etcd3 storage is in use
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-master/config.yml
```diff
@@ -88,7 +69,7 @@
         ha: "{{ openshift_master_ha | default(groups.oo_masters | length > 1) }}"
         master_count: "{{ openshift_master_count | default(groups.oo_masters | length) }}"
 
-- name: Determine if session secrets must be generated
+- name: Inspect state of first master session secrets and config
```
tbielawa: I see what you're doing here now. The referenced error (*undefined variabled*) is because `oo_masters_to_config` is having the `l_clean_install` fact set whereas later on we're referencing `hostvars[groups.oo_first_master.0`.

Makes sense to me. :1st_place_medal:

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
[test]

Issue #4523, opened 2017-06-21
docker installed before docker-excluder


ingvagabund:
In order to detect currently available OCP version during containerized deployment one needs to run `openshift_version` role. Given the version is available inside a docker image, one needs to run corresponding image, thus install `docker` rpms. Since the version detection is carried before the `docker-excluder` rpm is installed, one can install `docker` rpms of a version that is excluded by the `docker-excluder`. E.g. in 3.3 deployment one gets `docker-1.12*` installed instead of `docker-1.10*`.

At the same time in order to install the `docker-excluder` one needs to know OCP version (otherwise OpenShift rpms and OpenShift docker images version can differ).

Potential ways how to fix it?
* install docker, detect version, uninstall docker (if its version is not the expected one)
* install the latest docker-excluder available at the beginning, detect the OCP version, then reinstall (e.g. distro-sync) the docker-excluder to the proper version based on the detected OCP version.
* set the `docker_version` variable in the inventory file
* implement different mechanism to detect the containerized OCP version

sdodson:
Sounds like we should just hardcode docker_version and openshift_release in the 3.2, 3.3, 3.4, and 3.5 playbooks to appropriate values for those releases. Only do this for openshift-enterprise.

ingvagabund:
@jhadvig will this cause a trouble to the CI infra?

PR #4562, opened 2017-06-23
[3.3 backport] set default openshift_release and docker_version if not set

ingvagabund:
Fixes: #4523

### DIFF
--- a/playbooks/common/openshift-cluster/initialize_openshift_version.yml
+++ b/playbooks/common/openshift-cluster/initialize_openshift_version.yml
@@ -1,4 +1,17 @@
 ---
+- name: Set openshift_release and docker_version default values if not set
+  hosts: oo_all_hosts
+  gather_facts: no
+  tasks:
+  - set_fact:
+      openshift_release: "v3.3"
+    when:
+    - openshift_release is not defined
+  - set_fact:
+      docker_version: "1.10*"
+    when:
+    - docker_version is not defined
+
 # NOTE: requires openshift_facts be run
 - name: Determine openshift_version to configure on first master
   hosts: oo_first_master


ingvagabund:
aos-ci-test

ingvagabund:
@stevekuznetsov not completely sure where the [1] playbook comes from but it is installing docker-1.12 even before the OpenShift Ansible deployment for 3.3 is run. It should install docker-1.10 if anything. Is there any way to fix it? I.e. force docker version to 1.10.*?

[1] /home/jenkins/workspace/openshift-ansible-test-matrix/CONTAINERIZED/_NOT_containerized/OSE_VER/3.3/PYTHON/System-CPython-2.7/TOPOLOGY/openshift-cluster/TargetBranch/release-1.3/nodes/openshift-ansible-slave/atomic-ci-jobs/project/playbooks/openshift/setup.yml

stevekuznetsov:
Not without some custom work. We don't currently support testing for 3.3 on that CI job. @sdodson has been giving out green button pushes I think.

ingvagabund:
If this PR gets pushed without modifying the CI, every PR will be failing. Anyway, it is not right to test 3.3 with docker-1.12*.

stevekuznetsov:
> If this PR gets pushed without modifying the CI, every PR will be failing.

I don't understand -- is HEAD broken after this PR?

> Anyway, it is not right to test 3.3 with docker-1.12*.

Yes, known issue. It is hard to work on things like this as frequency of backports is very small and the work is large -- if you want to work on this to get CI working, we _do_ have a way to install custom Docker version (and we are doing something similar for `ose`) but you'll need to set all that up for these branches of `openshift-ansible.`

sdodson:
We're revisit this after we make it through 3.6 codefreeze. This can wait.

sdodson:
1.3 is no longer being maintained.

PR #4568, opened 2017-06-23
openshift_cfme: add nfs directory support

simon3z:
This adds the support for configuring the cfme nfs directory.

cc @gtanzillo @bazulay @fbladilo @sdodson @tbielawa

### DIFF
--- a/roles/openshift_cfme/defaults/main.yml
+++ b/roles/openshift_cfme/defaults/main.yml
@@ -28,6 +28,7 @@ openshift_cfme_pv_data:
 openshift_cfme_maxImagesBulkImportedPerRepository: 100
 # Hostname/IP of the NFS server. Currently defaults to first master
 openshift_cfme_nfs_server: "{{ groups.nfs.0 }}"
+openshift_cfme_nfs_directory: "/exports"
 # TODO: Refactor '_install_app' variable. This is just for testing but
 # maybe in the future it should control the entire yes/no for CFME.
 #
--- a/roles/openshift_cfme/templates/miq-pv-db.yaml.j2
+++ b/roles/openshift_cfme/templates/miq-pv-db.yaml.j2
@@ -8,6 +8,6 @@ spec:
   accessModes:
     - ReadWriteOnce
   nfs: 
-    path: /exports/miq-pv01
+    path: {{ openshift_cfme_nfs_directory }}/miq-pv01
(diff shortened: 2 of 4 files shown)


roles/openshift_cfme/templates/miq-pv-region.yaml.j2
```diff
@@ -8,6 +8,6 @@ spec:
   accessModes:
     - ReadWriteOnce
   nfs: 
-    path: /exports/miq-pv02
+    path: {{ openshift_cfme_nfs_directory}}/miq-pv02
```
tbielawa: Nit: can you ensure the variable name has spaces on either side after/before the `{{`/`}}`'s for consistency please?

`path: {{ openshift.... }}`

roles/openshift_cfme/templates/miq-pv-server.yaml.j2
```diff
@@ -8,6 +8,6 @@ spec:
   accessModes:
     - ReadWriteOnce
   nfs: 
-    path: /exports/miq-pv03
+    path: {{ openshift_cfme_nfs_directory}}/miq-pv03
```
tbielawa: Nit: as above

roles/openshift_cfme/templates/miq-pv-db.yaml.j2
```diff
@@ -8,6 +8,6 @@ spec:
   accessModes:
     - ReadWriteOnce
   nfs: 
-    path: /exports/miq-pv01
+    path: {{ openshift_cfme_nfs_directory}}/miq-pv01
```
tbielawa: Nit, same as other

roles/openshift_cfme/defaults/main.yml
```diff
@@ -28,6 +28,7 @@ openshift_cfme_pv_data:
 openshift_cfme_maxImagesBulkImportedPerRepository: 100
 # Hostname/IP of the NFS server. Currently defaults to first master
 openshift_cfme_nfs_server: "{{ groups.nfs.0 }}"
+openshift_cfme_nfs_directory: "/exports"
```
tbielawa: Please add description here to help users understand the implications of this.

The README may also need to be updated to reflect the fact that this is configurable and how it works. So maybe comments in two places are required.

PR #4570, opened 2017-06-23
Add playbook to run adhoc health checks or list existing checks

rhcarvalho:
This is useful while developing/testing health checks.

One step towards making checks more discoverable and easier to run.

### DIFF
--- a/playbooks/byo/openshift-checks/README.md
+++ b/playbooks/byo/openshift-checks/README.md
@@ -26,6 +26,9 @@ callback plugin summarizes execution errors at the end of a playbook run.
 3. Certificate expiry playbooks ([certificate_expiry](certificate_expiry)) -
    check that certificates in use are valid and not expiring soon.
 
+4. Adhoc playbook ([adhoc.yml](adhoc.yml)) - use it to run adhoc checks.
+   See the [next section](#the-adhoc-playbook) for a usage example.
+
 ## Running
 
 With a [recent installation of Ansible](../../../README.md#setup), run the playbook
@@ -57,6 +60,34 @@ against your inventory file. Here is the step-by-step:
     ```console
     $ ansible-playbook -i <inventory file> playbooks/byo/openshift-checks/certificate_expiry/default.yaml -v
     ```
+
+### The adhoc playbook
+
+The adhoc playbook gives flexibility to run any check or a custom group of
+checks. What will be run is determined by the `openshift_checks` variable,
+which, among other ways supported by Ansible, can be set on the command line
+using the `-e` flag.
+
+For example, to run the `docker_storage` check:
+
+```console
+$ ansible-playbook -i <inventory file> playbooks/byo/openshift-checks/adhoc.yml -e openshift_checks=docker_storage
+```
+
+To run more checks, use a comma-separated list of check names:
+
+```console
+$ ansible-playbook -i <inventory file> playbooks/byo/openshift-checks/adhoc.yml -e openshift_checks=docker_storage,disk_availability
+```
+
+To run an entire class of checks, use the name of a check group tag, prefixed by `@`. This will run all checks tagged `preflight`:
+
+```console
+$ ansible-playbook -i <inventory file> playbooks/byo/openshift-checks/adhoc.yml -e openshift_checks=@preflight
+```
+
+It is valid to specify multiple check tags and individual check names together
+in a comma-separated list.
 
 ## Running in a container
 
--- a/playbooks/byo/openshift-checks/README.md
+++ b/playbooks/byo/openshift-checks/README.md
@@ -26,7 +26,8 @@ callback plugin summarizes execution errors at the end of a playbook run.
 3. Certificate expiry playbooks ([certificate_expiry](certificate_expiry)) -
    check that certificates in use are valid and not expiring soon.
 
-4. Adhoc playbook ([adhoc.yml](adhoc.yml)) - use it to run adhoc checks.
+4. Adhoc playbook ([adhoc.yml](adhoc.yml)) - use it to run adhoc checks or to
+   list existing checks.
    See the [next section](#the-adhoc-playbook) for a usage example.
 
 ## Running
@@ -88,6 +89,13 @@ $ ansible-playbook -i <inventory file> playbooks/byo/openshift-checks/adhoc.yml 
 
 It is valid to specify multiple check tags and individual check names together
 in a comma-separated list.
+
+To list all of the available checks and tags, run the adhoc playbook without
+setting the `openshift_checks` variable:
+
+```console
+$ ansible-playbook -i <inventory file> playbooks/byo/openshift-checks/adhoc.yml
+```
 
 ## Running in a container
 
--- a/playbooks/byo/openshift-checks/README.md
+++ b/playbooks/byo/openshift-checks/README.md
@@ -7,15 +7,14 @@ Ansible's default operation mode is to fail fast, on the first error. However,
 when performing checks, it is useful to gather as much information about
 problems as possible in a single run.
 
-Thus, the playbooks run a battery of checks against the inventory hosts and have
-Ansible gather intermediate errors, giving a more complete diagnostic of the
-state of each host. If any check failed, the playbook run will be marked as
-failed.
+Thus, the playbooks run a battery of checks against the inventory hosts and
+gather intermediate errors, giving a more complete diagnostic of the state of
+each host. If any check failed, the playbook run will be marked as failed.
 
 To facilitate understanding the problems that were encountered, a custom
 callback plugin summarizes execution errors at the end of a playbook run.
 
-# Available playbooks
+## Available playbooks
 
 1. Pre-install playbook ([pre-install.yml](pre-install.yml)) - verifies system
    requirements and look for common problems that can prevent a successful
--- a/playbooks/byo/openshift-checks/adhoc.yml
+++ b/playbooks/byo/openshift-checks/adhoc.yml
@@ -1,4 +1,25 @@
 ---
+# NOTE: ideally this would be just part of a single play in
+# common/openshift-checks/adhoc.yml that lists the existing checks when
+# openshift_checks is not set or run the requested checks. However, to actually
+# run the checks we need to have the included dependencies to run first and that
+# takes time. To speed up listing checks, we use this separate play that runs
+# before the include of dependencies to save time and improve the UX.
+- name: OpenShift health checks
+  # NOTE: though the openshift_checks variable could be potentially defined on
+  # individual hosts while not defined for localhost, we do not support that
+  # usage. Running this play only in localhost speeds up execution.
+  hosts: localhost
+  connection: local
+  roles:
+  - openshift_health_checker
+  vars:
+  - r_openshift_health_checker_playbook_context: adhoc
+  pre_tasks:
+  - name: List known health checks
+    action: openshift_health_check
+    when: openshift_checks is undefined or not openshift_checks
+
 - include: ../openshift-cluster/initialize_groups.yml
 
 - include: ../../common/openshift-cluster/std_include.yml
--- a/playbooks/byo/openshift-checks/adhoc.yml
+++ b/playbooks/byo/openshift-checks/adhoc.yml
@@ -1,0 +1,6 @@
+---
+- include: ../openshift-cluster/initialize_groups.yml
+
+- include: ../../common/openshift-cluster/std_include.yml
+
+- include: ../../common/openshift-checks/adhoc.yml
--- a/playbooks/common/openshift-checks/adhoc.yml
+++ b/playbooks/common/openshift-checks/adhoc.yml
@@ -1,0 +1,12 @@
+---
+- name: OpenShift health checks
+  hosts: oo_all_hosts
+  roles:
+  - openshift_health_checker
+  vars:
+  - r_openshift_health_checker_playbook_context: adhoc
+  post_tasks:
+  - name: Run health checks
+    action: openshift_health_check
+    args:
+      checks: '{{ openshift_checks | default([]) }}'
--- a/roles/openshift_health_checker/action_plugins/openshift_health_check.py
+++ b/roles/openshift_health_checker/action_plugins/openshift_health_check.py
@@ -86,12 +86,10 @@ class ActionModule(ActionBase):
             name = cls.name
             if name in known_checks:
                 other_cls = known_checks[name].__class__
-                msg = "non-unique check name '{}' in: '{}' and '{}'".format(
-                    name,
-                    full_class_name(cls),
-                    full_class_name(other_cls),
+                raise OpenShiftCheckException(
+                    "duplicate check name '{}' in: '{}' and '{}'"
+                    "".format(name, full_class_name(cls), full_class_name(other_cls))
                 )
-                raise OpenShiftCheckException(msg)
             known_checks[name] = cls(execute_module=self._execute_module, tmp=tmp, task_vars=task_vars)
         return known_checks
 
--- a/roles/openshift_health_checker/action_plugins/openshift_health_check.py
+++ b/roles/openshift_health_checker/action_plugins/openshift_health_check.py
@@ -113,21 +113,27 @@ def list_known_checks(known_checks):
         'list of check names or a YAML list. Available checks:\n  {}'
     ).format('\n  '.join(sorted(known_checks)))
 
+    tags = describe_tags(known_checks.values())
+
+    msg += (
+        '\n\nTags can be used as a shortcut to select multiple '
+        'checks. Available tags and the checks they select:\n  {}'
+    ).format('\n  '.join(tags))
+
+    return msg
+
+
+def describe_tags(check_classes):
+    """Return a sorted list of strings describing tags and the checks they include."""
     tag_checks = defaultdict(list)
-    for cls in known_checks.values():
+    for cls in check_classes:
         for tag in cls.tags:
             tag_checks[tag].append(cls.name)
     tags = [
         '@{} = {}'.format(tag, ','.join(sorted(checks)))
         for tag, checks in tag_checks.items()
     ]
-
-    msg += (
-        '\n\nTags can be used as a shortcut to select multiple '
-        'checks. Available tags and the checks they select:\n  {}'
-    ).format('\n  '.join(sorted(tags)))
-
-    return msg
+    return sorted(tags)
 
 
 def resolve_checks(names, all_checks):
@@ -157,6 +163,12 @@ def resolve_checks(names, all_checks):
         if unknown_tag_names:
             msg.append('Unknown tag names: {}.'.format(', '.join(sorted(unknown_tag_names))))
         msg.append('Make sure there is no typo in the playbook and no files are missing.')
+        # TODO: implement a "Did you mean ...?" when the input is similar to a
+        # valid check or tag.
+        msg.append('Known checks:')
+        msg.append('  {}'.format('\n  '.join(sorted(known_check_names))))
+        msg.append('Known tags:')
+        msg.append('  {}'.format('\n  '.join(describe_tags(all_checks))))
         raise OpenShiftCheckException('\n'.join(msg))
 
     tag_to_checks = defaultdict(set)
--- a/roles/openshift_health_checker/action_plugins/openshift_health_check.py
+++ b/roles/openshift_health_checker/action_plugins/openshift_health_check.py
@@ -4,6 +4,7 @@ Ansible action plugin to execute health checks in OpenShift clusters.
 # pylint: disable=wrong-import-position,missing-docstring,invalid-name
 import sys
 import os
+import traceback
 from collections import defaultdict
 
 try:
@@ -58,26 +59,12 @@ class ActionModule(ActionBase):
 
         user_disabled_checks = normalize(task_vars.get('openshift_disable_check', []))
 
-        for check_name in resolved_checks:
-            display.banner("CHECK [{} : {}]".format(check_name, task_vars["ansible_host"]))
-            check = known_checks[check_name]
-
-            if not check.is_active():
-                r = dict(skipped=True, skipped_reason="Not active for this host")
-            elif check_name in user_disabled_checks:
-                r = dict(skipped=True, skipped_reason="Disabled by user request")
-            else:
-                try:
-                    r = check.run()
-                except OpenShiftCheckException as e:
-                    r = dict(
-                        failed=True,
-                        msg=str(e),
-                    )
-
+        for name in resolved_checks:
+            display.banner("CHECK [{} : {}]".format(name, task_vars["ansible_host"]))
+            check = known_checks[name]
+            check_results[name] = run_check(name, check, user_disabled_checks)
             if check.changed:
-                r["changed"] = True
-            check_results[check_name] = r
+                check_results[name]["changed"] = True
 
         result["changed"] = any(r.get("changed") for r in check_results.values())
         if any(r.get("failed") for r in check_results.values()):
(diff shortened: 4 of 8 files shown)


roles/openshift_health_checker/action_plugins/openshift_health_check.py
```diff
@@ -25,33 +25,37 @@ class ActionModule(ActionBase):
...
-            result["failed"] = True
-            result["msg"] = "'openshift' is undefined, did 'openshift_facts' run?"
-            return result
 
         try:
             known_checks = self.load_known_checks()
+            args = self._task.args
+            requested_checks = normalize(args.get('checks', []))
+
+            if not requested_checks:
```
juanvallejo: maybe I'm missing something, but wouldn't this cause other playbooks like `health` to fail here if they don't have a `checks` argument passed to them as well?

rhcarvalho: Yes, they would fail if either `checks` is not passed or is passed and is empty.

Did I overlook some genuine case when we want to call the action plugin without arguments?

Note that it fails because that's the easiest way to cause text to go to stdout/stderr (without messing with the verbosity level).

playbooks/byo/openshift-checks/adhoc.yml
```diff
@@ -0,0 +1,16 @@
...
+- name: OpenShift health checks
+  hosts: OSEv3
+  roles:
+  - openshift_health_checker
+  vars:
+  - r_openshift_health_checker_playbook_context: adhoc
+  pre_tasks:
+  - name: Print list of known health checks
+    action: openshift_health_check
+    when: checks is undefined or not checks
```
rhcarvalho: To print a list of known checks:

```
$ ansible-playbook -i <inventory file> playbooks/byo/openshift-checks/adhoc.yml
```

But that's not documented (yet). The problem with the current output is that it repeats for as many hosts as you have in the `OSEv3` group.

I would like to de-duplicate error messages by grouping together hosts that had the same error.

sosiouxme:
You'll want to adjust this to eval groups like https://github.com/openshift/openshift-ansible/pull/4495 once that merges.

rhcarvalho:
@sosiouxme thanks for pointing it out!

playbooks/byo/openshift-checks/adhoc.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: OpenShift health checks
+  hosts: OSEv3
+  roles:
+  - openshift_health_checker
+  vars:
+  - r_openshift_health_checker_playbook_context: adhoc
```
rhcarvalho: Done

playbooks/byo/openshift-checks/adhoc.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: OpenShift health checks
+  hosts: OSEv3
```
sosiouxme: This playbook needs to follow the pattern of the other check playbooks, invoking a parallel `playbooks/common/openshift-checks/adhoc.yml` playbook and evaluating groups along the way. `openshift_version` chokes otherwise. Also at some point we'll want to switch from using BYO group names to the common group names in our `is_active` methods.

playbooks/byo/openshift-checks/roles
```diff
@@ -0,0 +1 @@
+../../../roles
```
sosiouxme: I think it is not necessary when invoking the playbooks from common.

playbooks/byo/openshift-checks/adhoc.yml
```diff
@@ -0,0 +1,16 @@
...
+  - r_openshift_health_checker_playbook_context: adhoc
+  pre_tasks:
+  - name: Print list of known health checks
+    action: openshift_health_check
+    when: checks is undefined or not checks
+  post_tasks:
+  - name: Run health checks
+    action: openshift_health_check
+    args:
+      checks: '{{ checks | default([]) }}'
```
sosiouxme: Debating whether the `checks` inventory variable should be qualified. Inputs for most other things are expected to be `openshift_<whatever>` usually, so maybe `openshift_checks`? It seems kind of pedantic since this is specific to the playbook. But it's consistent.

playbooks/byo/openshift-checks/README.md
```diff
@@ -27,6 +27,9 @@ callback plugin summarizes execution errors at the end of a playbook run.
 3. Certificate expiry playbooks ([certificate_expiry](certificate_expiry)) -
    check that certificates in use are valid and not expiring soon.
 
+4. Adhoc playbook ([adhoc.yml](adhoc.yml)) - useful for running adhoc checks.
+   See the next section for an usage example.
```
rhcarvalho: Fixed, thanks!

playbooks/byo/openshift-checks/adhoc.yml
rhcarvalho: Yes, I agree. I think we went with `checks` for making it easier to type.

Still, I can't say I'm happy with such a long command:

```
$ ansible-playbook -i hosts playbooks/byo/openshift-checks/adhoc.yml -e checks=fluentd
```

In this case `oc diagnostics` (that could have a short alias like `oc diag`) is much easier on the eyes and less keystrokes. Surely keystrokes matter less for scripting.

If you think `openshift_checks` is more appropriate, we can plan a migration to that.

playbooks/byo/openshift-checks/adhoc.yml
rhcarvalho: I see. I have two questions about how that is working now, maybe you could help:

1. Why do we `include: ../openshift-cluster/initialize_groups.yml` in `byo/openshift-checks` and then `include: ../openshift-cluster/evaluate_groups.yml` in `common/openshift-checks`?

2. Why do we have `tags: always` in `include: ../openshift-cluster/evaluate_groups.yml`, but not in the evaluate_groups include?

playbooks/byo/openshift-checks/roles
rhcarvalho: Yes, `common` already has a symlink. I think back when I started this I was exploring how to have less layers to get something started, but definitely we should follow a single pattern.

playbooks/byo/openshift-checks/adhoc.yml
rhcarvalho: I'm noticing those includes make the playbook run considerably slower, though I didn't measure time explicitly :cry: 

For the case when there are no checks to be run (and we print a list of available checks), the different is 6s without includes, 8s with includes.

sosiouxme: How are you getting it to run without those includes at all? I get:

```
     Play:     OpenShift health checks
     Task:     openshift_version : Set rpm version to configure if openshift_pkg_version specified
     Message:  The conditional check 'inventory_hostname in groups['oo_masters_to_config'] or inventory_hostname in groups['oo_nodes_to_config']' failed. The error was: error while evaluating conditional (inventory_hostname in groups['oo_masters_to_config'] or inventory_hostname in groups['oo_nodes_to_config']): Unable to look up a name or access an attribute in template string ({% if inventory_hostname in groups['oo_masters_to_config'] or inventory_hostname in groups['oo_nodes_to_config'] %} True {% else %} False {% endif %}).
               Make sure your variable name does not contain invalid characters like '-': argument of type 'StrictUndefined' is not iterable
               
               The error appears to have been in '/home/lmeyer/go/src/github.com/openshift/openshift-ansible/roles/openshift_version/tasks/set_version_rpm.yml': line 2, column 3, but may
```

rhcarvalho: Yup, I get the same. It worked against an older version of master.

sosiouxme:
>
>
>
>    1.
>
>    Why do we include: ../openshift-cluster/initialize_groups.yml in
>    byo/openshift-checks and then include: ../openshift-cluster/evaluate_
>    groups.yml in common/openshift-checks?
>
>
The first is for translating BYO group names into `g_` names. It's a pretty
simple translation in BYO, more complex for other starting points. Then
these names are fed into the second which further defines `oo_first_master`
and such that are to actually be used in the roles.


>
>    1.
>
>    Why do we have tags: always in include: ../openshift-cluster/evaluate_
>    groups.yml, but not in the evaluate_groups include?
>
> This is probably cruft. Basically nothing actually works if you specify
tags (maybe some very narrow dev use cases).

playbooks/byo/openshift-checks/adhoc.yml
rhcarvalho: With a patch to `logging.py` to return `{}`, what is not even a real check:

```
$ time ansible-playbook -i hosts playbooks/byo/openshift-checks/adhoc.yml -e checks=logging -vvv
...
PLAY RECAP ***************************************************************************************************************************************
localhost                  : ok=8    changed=0    unreachable=0    failed=0   
master1                    : ok=36   changed=2    unreachable=0    failed=0   
master2                    : ok=36   changed=2    unreachable=0    failed=0   
node1                      : ok=36   changed=2    unreachable=0    failed=0   
node2                      : ok=36   changed=2    unreachable=0    failed=0   


real    1m42.949s
user    0m36.358s
sys     0m30.438s
```

So that's more than a minute and a half to run the simplest of the checks :cry: 


But anyway, do you understand why the includes are in different files and why the tag always is only in one of them?

playbooks/byo/openshift-checks/adhoc.yml
sosiouxme: What we pass in as an arg to the action plugin is irrelevant, as that's not an inventory variable. But we don't have a `checks` inventory variable before this PR so there's no migration necessary to change that.

You think that command line is long until you type the commands for running via the container... :)

I think the user should have `openshift_checks` for consistency with `openshift_disable_checks`...

    args:
      checks: '{{ openshift_checks | default([]) }}'

playbooks/byo/openshift-checks/adhoc.yml
sosiouxme: >    Why do we include: ../openshift-cluster/initialize_groups.yml in
>    byo/openshift-checks and then include: ../openshift-cluster/evaluate_
>    groups.yml in common/openshift-checks?

The first is for translating BYO group names into `g_` names. It's a pretty
simple translation in BYO, more complex for other starting points. Then
these names are fed into the second which further defines `oo_first_master`
and such that are to actually be used in the roles.

>    Why do we have tags: always in include: ../openshift-cluster/evaluate_
>    groups.yml, but not in the evaluate_groups include?

This is probably cruft. Basically nothing actually works if you specify
tags (maybe some very narrow dev use cases).

sosiouxme: All the other roles that run before `openshift_health_checker` are a problem :( and not just because it's slow. I wish we could separate them out into roles that gather info and set variables versus those which perform an action. Unfortunately there are a few actions (installs) necessary even to gather info. But it could be that those aren't necessary for our use case, and we could further divide the info-gatherers into fast and slow and just stick to info that can be gathered quickly for our checks. Maybe. I don't really look forward to that refactoring but running the checks is pretty heavy like this.

playbooks/byo/openshift-checks/roles
rhcarvalho: Symlink removed.

playbooks/byo/openshift-checks/adhoc.yml
rhcarvalho: FWIW, the original comment has been addressed and it is now in two separate playbooks, following the same pattern as others.

playbooks/byo/openshift-checks/adhoc.yml
rhcarvalho: Made it `openshift_checks`, thanks for pointing it out.

There is still room for debate when we compare with disable, because the latter is in singular form: `openshift_disable_check`...

rhcarvalho:
@sosiouxme this is ready for another pass of review.

roles/openshift_health_checker/test/zz_failure_summary_test.py
```diff
@@ -0,0 +1,63 @@
+from zz_failure_summary import deduplicate_failures
```
sosiouxme: :+1: it's a start, thanks!

rhcarvalho:
@sosiouxme could you please review the documentation changes? Wondering if that is clear enough. Being inside of a role is not the most visible location, but it is good to at least have a place to point people to. The docs here would be the base for something more in the official docs.

playbooks/byo/openshift-checks/README.md
```diff
@@ -59,6 +62,41 @@ against your inventory file. Here is the step-by-step:
...
+To run more checks, use a comma-separated list of check names:
+
+```console
+$ ansible-playbook -i <inventory file> playbooks/byo/openshift-checks/adhoc.yml -e openshift_checks=docker_storage,disk_availability
+```
+
+To run an entire class of checks, use the name of a check group tag. This will run all checks tagged `preflight`:
+
+```console
+$ ansible-playbook -i <inventory file> playbooks/byo/openshift-checks/adhoc.yml -e openshift_checks=preflight
```
rhcarvalho: There is a bug here, it should be `@preflight`.

roles/openshift_health_checker/action_plugins/openshift_health_check.py
```diff
@@ -1,99 +1,124 @@
...
+        'requested. Set the `openshift_checks` variable to a comma-separated '
+        'list of check names or a YAML list. Available checks:\n  {}'
+    ).format('\n  '.join(sorted(known_checks)))
+
+    tag_checks = defaultdict(list)
+    for cls in known_checks.values():
+        for tag in cls.tags:
+            tag_checks[tag].append(cls.name)
+    tags = [
+        '{} = {}'.format(tag, ','.join(sorted(checks)))
```
rhcarvalho: Should be `@tagname`.

roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
```diff
@@ -1,161 +1,223 @@
-"""
-Ansible callback plugin to give a nicely formatted summary of failures.
-"""
+"""Ansible callback plugin to print a nicely formatted summary of failures.
 
-# Reason: In several locations below we disable pylint protected-access
```
sosiouxme: Yeah but you don't have the requisite block of text...

It is fine with me :)

roles/openshift_health_checker/action_plugins/openshift_health_check.py
```diff
@@ -1,99 +1,124 @@
...
+            known_checks[name] = cls(execute_module=self._execute_module, tmp=tmp, task_vars=task_vars)
         return known_checks
 
 
+def list_known_checks(known_checks):
+    """Return text listing the existing checks and tags."""
+    # TODO: we could include a description of each check by taking it from a
+    # check class attribute (e.g., __doc__) when building the message below.
+    msg = (
+        'This playbook is meant to run health checks, but no checks were '
```
sosiouxme: I know what you mean, it looks bad if your terminal is thinner than the arbitrary point where we've placed newlines so it wraps anyway. I guess the real issue with that is on the check outputs since they're shown as indented blocks - tried to keep those short enough so that most people won't see them wrap. If we wanted to do that right we'd need to know the width of the terminal and have a real formatter on the output...

Just in general though I think for prose it makes sense to keep lines pretty short, like in a book. I may have a terminal that's 200 columns wide to look at logs and code and JSON, but it doesn't mean I want to read text at that width. I suppose ideally we'd have a formatter that broke on whitespace at the shorter of the terminal width or a comfortable-to-read width.

Well, I don't really want to start a typographical debate, just saying personally I would wrap it for legibility and consistency... it stuck out at me.

roles/openshift_health_checker/action_plugins/openshift_health_check.py
```diff
@@ -1,99 +1,124 @@
...
             known_checks = self.load_known_checks(tmp, task_vars)
             args = self._task.args
             requested_checks = normalize(args.get('checks', []))
+
+            if not requested_checks:
+                result['failed'] = True
+                result['msg'] = list_known_checks(known_checks)
+                return result
+
             resolved_checks = resolve_checks(requested_checks, known_checks.values())
```
sosiouxme: Well in that case it's probably not a problem with the code we shipped, but rather the user's input... They can do something about it without looking through our Ansible code. So something like "Make sure you have specified only valid check tags or names. Valid tags and names are as follows..."

roles/openshift_health_checker/action_plugins/openshift_health_check.py
rhcarvalho: I agree with you, though the difficulty is missing the tools to do it right as you say. We write using some `display` object; it might mean writing to stdout, but it could be anything, even a file that is not a terminal.

roles/openshift_health_checker/action_plugins/openshift_health_check.py
rhcarvalho: Right. We could list the existing checks.

I was thinking we may also detect that `etcd` is a tag name and should be spelled `@etcd` -- I made that mistake ;-)

sosiouxme:
@rhcarvalho the docs changes LGTM (although I didn't notice the missing `@` on tags so my brain may not be working).

roles/openshift_health_checker/action_plugins/openshift_health_check.py
rhcarvalho: Update: now the error message when the requested check is unknown includes the list of known checks and tags.

playbooks/common/openshift-checks/adhoc.yml
```diff
@@ -0,0 +1,18 @@
...
+- name: OpenShift health checks
+  hosts: OSEv3
+  roles:
+  - openshift_health_checker
+  vars:
+  - r_openshift_health_checker_playbook_context: adhoc
+  pre_tasks:
+  - name: List known health checks
+    action: openshift_health_check
+    when: openshift_checks is undefined or not openshift_checks
```
rhcarvalho: This is good for the two cases in the condition, but doesn't cover other cases such as when there is a typo in a check name.

I was thinking we could have a "preflight check" for the `openshift_health_check` plugin itself, implemented in the plugin perhaps with a flag like `dry-run: true`. But that doesn't have to be part of this PR.

roles/openshift_health_checker/action_plugins/openshift_health_check.py
```diff
@@ -1,99 +1,124 @@
...
-                raise OpenShiftCheckException(
-                    "non-unique check name '{}' in: '{}.{}' and '{}.{}'".format(
-                        check_name,
-                        cls.__module__, cls.__name__,
-                        other_cls.__module__, other_cls.__name__))
-            known_checks[check_name] = cls(execute_module=self._execute_module, tmp=tmp, task_vars=task_vars)
+            name = cls.name
+            if name in known_checks:
+                other_cls = known_checks[name].__class__
+                msg = "non-unique check name '{}' in: '{}' and '{}'".format(
```
rhcarvalho: Done.

roles/openshift_health_checker/action_plugins/openshift_health_check.py
```diff
@@ -144,3 +169,30 @@ def normalize(checks):
...
+        reason = "Could not determine if check should be run, exception: {}".format(exc)
+        return dict(skipped=True, skipped_reason=reason)
+
+    if not is_active:
+        return dict(skipped=True, skipped_reason="Not active for this host")
+
+    try:
+        return check.run()
+    except Exception as exc:
+        return dict(failed=True, msg=str(exc))
```
rhcarvalho: Update: I made it include a traceback on arbitrary exceptions using the `exception` key as described in http://docs.ansible.com/ansible/latest/common_return_values.html#exception

rhcarvalho:
@sosiouxme @juanvallejo if you have some time, could you please try this out and give feedback?

I'm using it like this:

### List available checks

```console
$ ansible-playbook -i hosts playbooks/byo/openshift-checks/adhoc.yml
```

This is notably unintuitive, and feels equivalent to calling a command without the required arguments and getting a usage error message, though it does serve the purpose of listing checks and tags.

It also made me realize an inconsistency with `pre-install` and `preflight` tags.

### Run a single check

```console
$ ansible-playbook -i hosts playbooks/byo/openshift-checks/adhoc.yml -e openshift_checks=disk_availability
```
### Run multiple checks

#### by tag

```console
$ ansible-playbook -i hosts playbooks/byo/openshift-checks/adhoc.yml -e openshift_checks=@preflight
```

#### by name, comma-separated

```console
$ ansible-playbook -i hosts playbooks/byo/openshift-checks/adhoc.yml -e openshift_checks=docker_storage,disk_availability
```

#### by name, YAML

```console
$ ansible-playbook -i hosts playbooks/byo/openshift-checks/adhoc.yml -e '{"openshift_checks": ["docker_storage", "disk_availability"]}'
```

Except for this last one, these use cases are all described in `openshift-checks/README.md`. I omit the last one for it is not very convenient for command line usage.

rhcarvalho:
aos-ci-test

playbooks/common/openshift-checks/adhoc.yml
```diff
@@ -0,0 +1,30 @@
+---
+# NOTE: ideally this would be just part of a single play that lists the existing
+# checks when openshift_checks is not set or run the requested checks. However,
+# to actually run the checks we need to have the included dependencies to run
+# first and that takes time. To speed up listing checks, we use a separate play
+# that runs before the include to save time and improve the UX.
+- name: OpenShift health checks
+  hosts: OSEv3
```
rhcarvalho: @sosiouxme, in trying to change from OSEv3 to oo_all_hosts (empty group before initialization), I realized we could as well only run this play in localhost. The downside is it will not play nice with host-specific variables.
WDYT?

sosiouxme: I think yes, run it against localhost so we don't have to init groups at this point nor put `OSEv3` in a `common` playbook. The only inconsistency would be if they didn't define `openshift_checks` globally but only on individual hosts, and I think that's not a very important use case.

mtnbikenc: Agreed on not using OSEv3 in common playbooks.

playbooks/common/openshift-checks/adhoc.yml
mtnbikenc: `pre_tasks` always run before `roles`.  Did you mean to use `post_tasks` here?  The order of execution for play stages is:
```yaml
- hosts: foo
  vars:
  pre_tasks:
  roles:
  tasks:
  post_tasks:
```
Putting these items in a different order in the yaml file does not change execution order.  Same with `vars:`, they are defined at the beginning of the play.

playbooks/byo/openshift-checks/adhoc.yml
```diff
@@ -0,0 +1,27 @@
...
+  # NOTE: though the openshift_checks variable could be potentially defined on
+  # individual hosts while not defined for localhost, we do not support that
+  # usage. Running this play only in localhost speeds up execution.
+  hosts: localhost
+  connection: local
+  roles:
+  - openshift_health_checker
+  vars:
+  - r_openshift_health_checker_playbook_context: adhoc
+  pre_tasks:
```
rhcarvalho: cc @mtnbikenc 

At this point in time this could be any of `pre_tasks`, `tasks` or `post_tasks` and the observable end result would be the same.

`pre_tasks` run before any tasks defined in the role (there are none), and seemed to express what we want, to run the conditional as soon as possible.

rhcarvalho:
aos-ci-test

playbooks/byo/openshift-checks/adhoc.yml
```diff
@@ -0,0 +1,27 @@
...
+# openshift_checks is not set or run the requested checks. However, to actually
+# run the checks we need to have the included dependencies to run first and that
+# takes time. To speed up listing checks, we use this separate play that runs
+# before the include of dependencies to save time and improve the UX.
+- name: OpenShift health checks
+  # NOTE: though the openshift_checks variable could be potentially defined on
+  # individual hosts while not defined for localhost, we do not support that
+  # usage. Running this play only in localhost speeds up execution.
+  hosts: localhost
+  connection: local
```
rhcarvalho: @sosiouxme FYI, I made the listing of checks run on localhost only.

playbooks/byo/openshift-checks/adhoc.yml
sosiouxme: `pre_tasks` also runs before any role dependencies from meta, which is also desirable here (though I suspect we can now remove `openshift_facts` from dependencies anyway).

rhcarvalho:
[merge]

rhcarvalho:
Flake https://github.com/openshift/origin/issues/16005

[merge]

sdodson:
flake https://github.com/openshift/origin/issues/11873

PR #4573, opened 2017-06-23
Fix parsing certs with very large serial numbers

tbielawa:
Certificates with large serial numbers (such as those from commercial
vendors) previously failed to parse due to absent logic in the
FakeOpenSSLCertificate parser. This would cause the module to abort
and break any other roles depending on the cert expiry checking
results.

* Identifies large serials in HEX format for decimal conversion
* Update unit tests to parse a large serial

### DIFF
--- a/roles/openshift_certificate_expiry/test/conftest.py
+++ b/roles/openshift_certificate_expiry/test/conftest.py
@@ -23,7 +23,10 @@ VALID_CERTIFICATE_PARAMS = [
     {
         'short_name': 'combined',
         'cn': 'combined.example.com',
-        'serial': 6,
+        # Verify that HUGE serials parse correctly.
+        # Frobs PARSING_HEX_SERIAL in _parse_cert
+        # See https://bugzilla.redhat.com/show_bug.cgi?id=1464240
+        'serial': 14449739080294792594019643629255165375,
         'uses': b'clientAuth, serverAuth',
         'dns': ['etcd'],
         'ip': ['10.0.0.2', '192.168.0.2']
--- a/roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
+++ b/roles/openshift_certificate_expiry/library/openshift_cert_expiry.py
@@ -104,6 +104,7 @@ platforms missing the Python OpenSSL library.
         self.extensions = []
 
         PARSING_ALT_NAMES = False
+        PARSING_HEX_SERIAL = False
         for line in self.cert_string.split('\n'):
             l = line.strip()
             if PARSING_ALT_NAMES:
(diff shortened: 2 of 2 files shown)


tbielawa:
aos-ci-test

tbielawa:
[test]

roles/openshift_certificate_expiry/test/conftest.py
```diff
@@ -23,7 +23,10 @@
     {
         'short_name': 'combined',
         'cn': 'combined.example.com',
-        'serial': 6,
+        # Verify that HUGE serials parse correctly.
+        # Frobs PARSING_HEX_SERIAL in _parse_cert
+        # See https://bugzilla.redhat.com/show_bug.cgi?id=1464240
+        'serial': 14449739080294792594019643629255165375,
```
ashcrow: Is this sufficient? The tested function notes `` # Hex serials arrive colon-delimited``

tbielawa: That may appear misleading, I see your concern here @ashcrow . What pulls this all together is in `test_fakeopensslclasses.py`

```
def fake_valid_cert(valid_cert):
    cmd = ['openssl', 'x509', '-in', str(valid_cert['cert_file']), '-text']
    cert = subprocess.check_output(cmd)
    return FakeOpenSSLCertificate(cert.decode('utf8'))
```

The cert objects in memory are seralized and then plain-text versions are produced as from the actual openssl command. After this processing the hex-delimited form is produced and later parsed in `FakeOpenSSLCertificate._parse_cert`. Again, it can be misleading because you never technically *see* the hex version anywhere.

To be sure though, I just double-checked the coverage report and can verify that the relevant logic branches are all being traversed.

*note* I did not try shoving the hex equivalent into the `serial` field up there directly in my original testing. Just for giggles I did right now though and all tests still passed.

```
$ git diff                    
diff --git a/roles/openshift_certificate_expiry/test/conftest.py b/roles/openshift_certificate_expiry/test/conftest.py
index df948ff..fe7b14e 100644
--- a/roles/openshift_certificate_expiry/test/conftest.py
+++ b/roles/openshift_certificate_expiry/test/conftest.py
@@ -26,7 +26,8 @@ VALID_CERTIFICATE_PARAMS = [
         # Verify that HUGE serials parse correctly.
         # Frobs PARSING_HEX_SERIAL in _parse_cert
         # See https://bugzilla.redhat.com/show_bug.cgi?id=1464240
-        'serial': 14449739080294792594019643629255165375,
+        'serial': 0x0adeeb240475ab563914e95a22e285bf,
+        #'serial': 14449739080294792594019643629255165375,
         'uses': b'clientAuth, serverAuth',
         'dns': ['etcd'],
         'ip': ['10.0.0.2', '192.168.0.2']
```

sdodson:
ci doesn't touch this.

PR #4575, opened 2017-06-23
[3.3] Fix parsing certs with very large serial numbers

tbielawa:
https://github.com/openshift/openshift-ansible/pull/4573

Certificates with large serial numbers (such as those from commercial
vendors) previously failed to parse due to absent logic in the
FakeOpenSSLCertificate parser. This would cause the module to abort
and break any other roles depending on the cert expiry checking
results.

* Identifies large serials in HEX format for decimal conversion
* Update unit tests to parse a large serial


tbielawa:
aos-ci-test

PR #4583, opened 2017-06-26
oc_atomic_container: fix version check for atomic

giuseppe:
Some versions of atomic write their version to stderr,
consider that as well.


giuseppe:
@ashcrow done here ⬆️

ashcrow:
@sdodson I don't believe this will be tested by CI

sdodson:
agreed

Issue #4475, opened 2017-06-16
3.6 upgrade for multinode cluster fails during generating etcd backup 


jhadvig:
When upgrading a multinode cluster using following [inventory](http://pastebin.test.redhat.com/494791), I can see the play to fail during the process of generating etcd backup for nodes that contain etcd

```

TASK [etcd_upgrade : Generate etcd backup] **************************************************************************************************************************************************************************************************
task path: /usr/share/ansible/openshift-ansible/roles/etcd_upgrade/tasks/backup.yml:48
fatal: [54.242.203.97]: FAILED! => {
    "changed": true, 
    "cmd": [
        "etcdctl", 
        "backup", 
        "--data-dir=/var/lib/etcd/", 
        "--backup-dir=/var/lib/etcd//openshift-backup-etcd_backup_tag20170616131638"
    ], 
    "delta": "0:00:00.030114", 
    "end": "2017-06-16 13:16:39.876109", 
    "failed": true, 
    "rc": 1, 
    "start": "2017-06-16 13:16:39.845995"
}

STDERR:

2017-06-16 13:16:39.871697 I | failed creating backup snapshot dir /var/lib/etcd/openshift-backup-etcd_backup_tag20170616131638/member/snap: expected "/var/lib/etcd/openshift-backup-etcd_backup_tag20170616131638/member/snap" to be empty, got ["000000000000000b-0000000000849d10.snap"]

fatal: [54.165.80.128]: FAILED! => {
    "changed": true, 
    "cmd": [
        "etcdctl", 
        "backup", 
        "--data-dir=/var/lib/etcd/", 
        "--backup-dir=/var/lib/etcd//openshift-backup-etcd_backup_tag20170616131638"
    ], 
    "delta": "0:00:00.017860", 
    "end": "2017-06-16 13:16:39.922752", 
    "failed": true, 
    "rc": 1, 
    "start": "2017-06-16 13:16:39.904892"
}

STDERR:

2017-06-16 13:16:39.921966 I | failed creating backup snapshot dir /var/lib/etcd/openshift-backup-etcd_backup_tag20170616131638/member/snap: expected "/var/lib/etcd/openshift-backup-etcd_backup_tag20170616131638/member/snap" to be empty, got ["000000000000000b-0000000000849d10.snap"]

changed: [54.88.176.49] => {
    "changed": true, 
    "cmd": [
        "etcdctl", 
        "backup", 
        "--data-dir=/var/lib/etcd/", 
[...]

PR #4584, opened 2017-06-26
Add hostname to etcd backup folder name

jhadvig:
@ingvagabund PTAL

fixes https://github.com/openshift/openshift-ansible/issues/4475

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
+++ b/playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
@@ -8,7 +8,7 @@
     r_etcd_common_backup_tag: etcd_backup_tag
     r_etcd_common_etcd_runtime: "{{ openshift.common.etcd_runtime }}"
     r_etcd_common_embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
-    r_etcd_common_backup_sufix_name: "{{ lookup('pipe', 'date +%Y%m%d%H%M%S') }}"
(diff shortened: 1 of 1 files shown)


playbooks/common/openshift-cluster/upgrades/etcd/backup.yml
```diff
@@ -8,7 +8,7 @@
     r_etcd_common_backup_tag: etcd_backup_tag
     r_etcd_common_etcd_runtime: "{{ openshift.common.etcd_runtime }}"
     r_etcd_common_embedded_etcd: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
-    r_etcd_common_backup_sufix_name: "{{ lookup('pipe', 'date +%Y%m%d%H%M%S') }}"
+    r_etcd_common_backup_sufix_name: "{{ inventory_hostname + '-' + lookup('pipe', 'date +%Y%m%d%H%M%S') }}"
```
ingvagabund: @sdodson WDYT? Do we have any naming schema for etcd backups?

ingvagabund:
aos-ci-test

sdodson:
I'd still like to understand why this is necessary. /var/lib/etcd should never be mounted across hosts so there should never be any conflict.

jhadvig:
After some investigation I found out that the issue was cause by launching the  upgrade playbook with `--connection local` flag. After removing it the issue cant be reproduced. Closing this one.

sdodson:
now it makes sense! thanks

PR #4592, opened 2017-06-26
add fluentd logging driver config check

juanvallejo:
Related trello card: https://trello.com/c/t6H0D8Ox

Checks to see if the fluentd pod is configured to aggregate logs from `journald` or `json-file` and verifies that the host's Docker config matches the expected logging driver.

*TODO*

- [x] add tests

cc @sosiouxme @rhcarvalho @brenton

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/__init__.py
+++ b/roles/openshift_health_checker/openshift_checks/__init__.py
@@ -105,6 +105,29 @@ class OpenShiftCheck(object):
             raise OpenShiftCheckException("'{}' is undefined".format(".".join(map(str, keys))))
         return value
 
+    @staticmethod
+    def get_major_minor_version(openshift_image_tag):
+        """Parse and return the deployed version of OpenShift as a tuple."""
+        if openshift_image_tag and openshift_image_tag[0] == 'v':
+            openshift_image_tag = openshift_image_tag[1:]
+
+        # map major release versions across releases
+        # to a common major version
+        openshift_major_release_version = {
+            "1": "3",
+        }
+
+        components = openshift_image_tag.split(".")
+        if not components or len(components) < 2:
+            msg = "An invalid version of OpenShift was found for this host: {}"
+            raise OpenShiftCheckException(msg.format(openshift_image_tag))
+
+        if components[0] in openshift_major_release_version:
+            components[0] = openshift_major_release_version[components[0]]
+
+        components = tuple(int(x) for x in components[:2])
+        return components
+
 
 LOADER_EXCLUDES = (
     "__init__.py",
--- a/roles/openshift_health_checker/openshift_checks/logging/curator.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/curator.py
@@ -9,11 +9,11 @@ class Curator(LoggingCheck):
     name = "curator"
     tags = ["health", "logging"]
 
-    logging_namespace = None
-
     def run(self):
+        """Check various things and gather errors. Returns: result as hash"""
+
         self.logging_namespace = self.get_var("openshift_logging_namespace", default="logging")
-        curator_pods, error = super(Curator, self).get_pods_for_component(
+        curator_pods, error = self.get_pods_for_component(
             self.logging_namespace,
             "curator",
         )
@@ -23,7 +23,6 @@ class Curator(LoggingCheck):
 
         if check_error:
             msg = ("The following Curator deployment issue was found:"
-                   "\n-------\n"
                    "{}".format(check_error))
             return {"failed": True, "changed": False, "msg": msg}
 
@@ -39,7 +38,7 @@ class Curator(LoggingCheck):
                 "Is Curator correctly deployed?"
             )
 
-        not_running = super(Curator, self).not_running_pods(pods)
+        not_running = self.not_running_pods(pods)
         if len(not_running) == len(pods):
             return (
                 "The Curator pod is not currently in a running state,\n"
--- a/roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
@@ -12,13 +12,11 @@ class Elasticsearch(LoggingCheck):
     name = "elasticsearch"
     tags = ["health", "logging"]
 
-    logging_namespace = None
-
     def run(self):
         """Check various things and gather errors. Returns: result as hash"""
 
         self.logging_namespace = self.get_var("openshift_logging_namespace", default="logging")
-        es_pods, error = super(Elasticsearch, self).get_pods_for_component(
+        es_pods, error = self.get_pods_for_component(
             self.logging_namespace,
             "es",
         )
@@ -28,7 +26,6 @@ class Elasticsearch(LoggingCheck):
 
         if check_error:
             msg = ("The following Elasticsearch deployment issue was found:"
-                   "\n-------\n"
                    "{}".format(check_error))
             return {"failed": True, "changed": False, "msg": msg}
 
@@ -37,7 +34,7 @@ class Elasticsearch(LoggingCheck):
 
     def _not_running_elasticsearch_pods(self, es_pods):
         """Returns: list of pods that are not running, list of errors about non-running pods"""
-        not_running = super(Elasticsearch, self).not_running_pods(es_pods)
+        not_running = self.not_running_pods(es_pods)
         if not_running:
             return not_running, [(
                 'The following Elasticsearch pods are not running:\n'
@@ -78,7 +75,7 @@ class Elasticsearch(LoggingCheck):
         for pod_name in pods_by_name.keys():
             # Compare what each ES node reports as master and compare for split brain
             get_master_cmd = self._build_es_curl_cmd(pod_name, "https://localhost:9200/_cat/master")
-            master_name_str = self._exec_oc(get_master_cmd, [])
+            master_name_str = self.exec_oc(self.logging_namespace, get_master_cmd, [])
             master_names = (master_name_str or '').split(' ')
             if len(master_names) > 1:
                 es_master_names.add(master_names[1])
@@ -111,7 +108,7 @@ class Elasticsearch(LoggingCheck):
 
         # get ES cluster nodes
         node_cmd = self._build_es_curl_cmd(list(pods_by_name.keys())[0], 'https://localhost:9200/_nodes')
-        cluster_node_data = self._exec_oc(node_cmd, [])
+        cluster_node_data = self.exec_oc(self.logging_namespace, node_cmd, [])
         try:
             cluster_nodes = json.loads(cluster_node_data)['nodes']
         except (ValueError, KeyError):
@@ -138,7 +135,7 @@ class Elasticsearch(LoggingCheck):
         error_msgs = []
         for pod_name in pods_by_name.keys():
             cluster_health_cmd = self._build_es_curl_cmd(pod_name, 'https://localhost:9200/_cluster/health?pretty=true')
-            cluster_health_data = self._exec_oc(cluster_health_cmd, [])
+            cluster_health_data = self.exec_oc(self.logging_namespace, cluster_health_cmd, [])
             try:
                 health_res = json.loads(cluster_health_data)
                 if not health_res or not health_res.get('status'):
@@ -165,7 +162,7 @@ class Elasticsearch(LoggingCheck):
         error_msgs = []
         for pod_name in pods_by_name.keys():
             df_cmd = 'exec {} -- df --output=ipcent,pcent /elasticsearch/persistent'.format(pod_name)
-            disk_output = self._exec_oc(df_cmd, [])
+            disk_output = self.exec_oc(self.logging_namespace, df_cmd, [])
             lines = disk_output.splitlines()
             # expecting one header looking like 'IUse% Use%' and one body line
             body_re = r'\s*(\d+)%?\s+(\d+)%?\s*$'
@@ -201,10 +198,3 @@ class Elasticsearch(LoggingCheck):
                     ))
 
         return error_msgs
-
-    def _exec_oc(self, cmd_str, extra_args):
-        return super(Elasticsearch, self).exec_oc(
-            self.logging_namespace,
-            cmd_str,
-            extra_args,
-        )
--- a/roles/openshift_health_checker/openshift_checks/logging/fluentd.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/fluentd.py
@@ -11,8 +11,6 @@ class Fluentd(LoggingCheck):
     name = "fluentd"
     tags = ["health", "logging"]
 
-    logging_namespace = None
-
     def run(self):
         """Check various things and gather errors. Returns: result as hash"""
 
@@ -27,7 +25,6 @@ class Fluentd(LoggingCheck):
 
         if check_error:
             msg = ("The following Fluentd deployment issue was found:"
-                   "\n-------\n"
                    "{}".format(check_error))
             return {"failed": True, "changed": False, "msg": msg}
 
@@ -147,7 +144,11 @@ class Fluentd(LoggingCheck):
 
     def get_nodes_by_name(self):
         """Retrieve all the node definitions. Returns: dict(name: node), error string"""
-        nodes_json = self._exec_oc("get nodes -o json", [])
+        nodes_json = self.exec_oc(
+            self.logging_namespace,
+            "get nodes -o json",
+            []
+        )
         try:
             nodes = json.loads(nodes_json)
         except ValueError:  # no valid json - should not happen
@@ -158,10 +159,3 @@ class Fluentd(LoggingCheck):
             node['metadata']['name']: node
             for node in nodes['items']
         }, None
-
-    def _exec_oc(self, cmd_str, extra_args):
-        return super(Fluentd, self).exec_oc(
-            self.logging_namespace,
-            cmd_str,
-            extra_args,
-        )
--- a/roles/openshift_health_checker/openshift_checks/logging/fluentd_config.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/fluentd_config.py
@@ -1,0 +1,138 @@
+"""
+Module for performing checks on a Fluentd logging deployment configuration
+"""
+
+from openshift_checks import OpenShiftCheckException
+from openshift_checks.logging.logging import LoggingCheck
+
+
+class FluentdConfig(LoggingCheck):
+    """Module that checks logging configuration of an integrated logging Fluentd deployment"""
+    name = "fluentd_config"
+    tags = ["health"]
+
+    def is_active(self):
+        logging_deployed = self.get_var("openshift_hosted_logging_deploy", default=False)
+
+        try:
+            version = self.get_major_minor_version(self.get_var("openshift_image_tag"))
+        except ValueError:
+            # if failed to parse OpenShift version, perform check anyway (if logging enabled)
+            return logging_deployed
+
+        return logging_deployed and version < (3, 6)
+
+    def run(self):
+        """Check that Fluentd has running pods, and that its logging config matches Docker's logging config."""
+        self.logging_namespace = self.get_var("openshift_logging_namespace", default=self.logging_namespace)
+        config_error = self.check_logging_config()
+        if config_error:
+            msg = ("The following Fluentd logging configuration problem was found:"
+                   "\n{}".format(config_error))
+            return {"failed": True, "msg": msg}
+
+        return {}
+
+    def check_logging_config(self):
+        """Ensure that the configured Docker logging driver matches fluentd settings.
+        This means that, at least for now, if the following condition is met:
+
+            openshift_logging_fluentd_use_journal == True
+
+        then the value of the configured Docker logging driver should be "journald".
+        Otherwise, the value of the Docker logging driver should be "json-file".
+        Returns an error string if the above condition is not met, or None otherwise."""
+        use_journald = self.get_var("openshift_logging_fluentd_use_journal", default=True)
+
+        # if check is running on a master, retrieve all running pods
+        # and check any pod's container for the env var "USE_JOURNAL"
+        group_names = self.get_var("group_names")
+        if "masters" in group_names:
+            use_journald = self.check_fluentd_env_var()
+
+        docker_info = self.execute_module("docker_info", {})
+        try:
+            logging_driver = docker_info["info"]["LoggingDriver"]
+        except KeyError:
+            return "Unable to determine Docker logging driver."
+
+        logging_driver = docker_info["info"]["LoggingDriver"]
+        recommended_logging_driver = "journald"
+        error = None
+
+        # If fluentd is set to use journald but Docker is not, recommend setting the `--log-driver`
+        # option as an inventory file variable, or adding the log driver value as part of the
+        # Docker configuration in /etc/docker/daemon.json. There is no global --log-driver flag that
+        # can be passed to the Docker binary; the only other recommendation that can be made, would be
+        # to pass the `--log-driver` flag to the "run" sub-command of the `docker` binary when running
+        # individual containers.
+        if use_journald and logging_driver != "journald":
+            error = ('Your Fluentd configuration is set to aggregate Docker container logs from "journald".\n'
+                     'This differs from your Docker configuration, which has been set to use "{driver}" '
+                     'as the default method of storing logs.\n'
+                     'This discrepancy in configuration will prevent Fluentd from receiving any logs'
+                     'from your Docker containers.').format(driver=logging_driver)
+        elif not use_journald and logging_driver != "json-file":
+            recommended_logging_driver = "json-file"
+            error = ('Your Fluentd configuration is set to aggregate Docker container logs from '
+                     'individual json log files per container.\n '
+                     'This differs from your Docker configuration, which has been set to use '
+                     '"{driver}" as the default method of storing logs.\n'
+                     'This discrepancy in configuration will prevent Fluentd from receiving any logs'
+                     'from your Docker containers.').format(driver=logging_driver)
+
+        if error:
+            error += ('\nTo resolve this issue, add the following variable to your Ansible inventory file:\n\n'
+                      '  openshift_docker_options="--log-driver={driver}"\n\n'
+                      'Alternatively, you can add the following option to your Docker configuration, located in'
+                      '"/etc/docker/daemon.json":\n\n'
+                      '{{ "log-driver": "{driver}" }}\n\n'
+                      'See https://docs.docker.com/engine/admin/logging/json-file '
+                      'for more information.').format(driver=recommended_logging_driver)
+
+        return error
+
+    def check_fluentd_env_var(self):
+        """Read and return the value of the 'USE_JOURNAL' environment variable on a fluentd pod."""
+        running_pods = self.running_fluentd_pods()
+
+        try:
+            pod_containers = running_pods[0]["spec"]["containers"]
+        except KeyError:
+            return "Unable to detect running containers on selected Fluentd pod."
+
+        if not pod_containers:
+            msg = ('There are no running containers on selected Fluentd pod "{}".\n'
+                   'Unable to calculate expected logging driver.').format(running_pods[0]["metadata"].get("name", ""))
+            raise OpenShiftCheckException(msg)
+
+        pod_env = pod_containers[0].get("env")
+        if not pod_env:
+            msg = ('There are no environment variables set on the Fluentd container "{}".\n'
+                   'Unable to calculate expected logging driver.').format(pod_containers[0].get("name"))
+            raise OpenShiftCheckException(msg)
+
+        for env in pod_env:
+            if env["name"] == "USE_JOURNAL":
+                return env.get("value", "false") != "false"
+
+        return False
+
+    def running_fluentd_pods(self):
+        """Return a list of running fluentd pods."""
+        fluentd_pods, error = self.get_pods_for_component(
+            self.logging_namespace,
+            "fluentd",
+        )
+        if error:
+            msg = 'Unable to retrieve any pods for the "fluentd" logging component: {}'.format(error)
+            raise OpenShiftCheckException(msg)
+
+        running_fluentd_pods = [pod for pod in fluentd_pods if pod['status']['phase'] == 'Running']
+        if not running_fluentd_pods:
+            msg = ('No Fluentd pods were found to be in the "Running" state. '
+                   'At least one Fluentd pod is required in order to perform this check.')
+
+            raise OpenShiftCheckException(msg)
+
+        return running_fluentd_pods
--- a/roles/openshift_health_checker/openshift_checks/logging/kibana.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/kibana.py
@@ -21,13 +21,11 @@ class Kibana(LoggingCheck):
     name = "kibana"
     tags = ["health", "logging"]
 
-    logging_namespace = None
-
     def run(self):
         """Check various things and gather errors. Returns: result as hash"""
 
         self.logging_namespace = self.get_var("openshift_logging_namespace", default="logging")
-        kibana_pods, error = super(Kibana, self).get_pods_for_component(
+        kibana_pods, error = self.get_pods_for_component(
             self.logging_namespace,
             "kibana",
         )
@@ -40,7 +38,6 @@ class Kibana(LoggingCheck):
 
         if check_error:
             msg = ("The following Kibana deployment issue was found:"
-                   "\n-------\n"
                    "{}".format(check_error))
             return {"failed": True, "changed": False, "msg": msg}
 
@@ -118,7 +115,11 @@ class Kibana(LoggingCheck):
         """
 
         # Get logging url
-        get_route = self._exec_oc("get route logging-kibana -o json", [])
+        get_route = self.exec_oc(
+            self.logging_namespace,
+            "get route logging-kibana -o json",
+            [],
+        )
         if not get_route:
             return None, 'no_route_exists'
 
@@ -217,10 +218,3 @@ class Kibana(LoggingCheck):
             ).format(error=error)
             return error
         return None
-
-    def _exec_oc(self, cmd_str, extra_args):
-        return super(Kibana, self).exec_oc(
-            self.logging_namespace,
-            cmd_str,
-            extra_args,
-        )
--- a/roles/openshift_health_checker/test/elasticsearch_test.py
+++ b/roles/openshift_health_checker/test/elasticsearch_test.py
@@ -6,14 +6,6 @@ from openshift_checks.logging.elasticsearch import Elasticsearch
 task_vars_config_base = dict(openshift=dict(common=dict(config_base='/etc/origin')))
 
 
-def canned_elasticsearch(task_vars=None, exec_oc=None):
-    """Create an Elasticsearch check object with canned exec_oc method"""
-    check = Elasticsearch("dummy", task_vars or {})  # fails if a module is actually invoked
-    if exec_oc:
-        check._exec_oc = exec_oc
-    return check
-
-
 def assert_error(error, expect_error):
     if expect_error:
         assert error
@@ -50,10 +42,10 @@ split_es_pod = {
 
 
 def test_check_elasticsearch():
-    assert 'No logging Elasticsearch pods' in canned_elasticsearch().check_elasticsearch([])
+    assert 'No logging Elasticsearch pods' in Elasticsearch().check_elasticsearch([])
 
     # canned oc responses to match so all the checks pass
-    def _exec_oc(cmd, args):
+    def _exec_oc(ns, cmd, args):
         if '_cat/master' in cmd:
             return 'name logging-es'
         elif '/_nodes' in cmd:
@@ -65,7 +57,9 @@ def test_check_elasticsearch():
         else:
             raise Exception(cmd)
 
-    assert not canned_elasticsearch({}, _exec_oc).check_elasticsearch([plain_es_pod])
+    check = Elasticsearch(None, {})
+    check.exec_oc = _exec_oc
+    assert not check.check_elasticsearch([plain_es_pod])
 
 
 def pods_by_name(pods):
@@ -88,8 +82,8 @@ def pods_by_name(pods):
 ])
 def test_check_elasticsearch_masters(pods, expect_error):
     test_pods = list(pods)
-    check = canned_elasticsearch(task_vars_config_base, lambda cmd, args: test_pods.pop(0)['_test_master_name_str'])
-
+    check = Elasticsearch(None, task_vars_config_base)
+    check.execute_module = lambda cmd, args: {'result': test_pods.pop(0)['_test_master_name_str']}
     errors = check._check_elasticsearch_masters(pods_by_name(pods))
     assert_error(''.join(errors), expect_error)
 
@@ -124,7 +118,8 @@ es_node_list = {
     ),
 ])
 def test_check_elasticsearch_node_list(pods, node_list, expect_error):
-    check = canned_elasticsearch(task_vars_config_base, lambda cmd, args: json.dumps(node_list))
+    check = Elasticsearch(None, task_vars_config_base)
+    check.execute_module = lambda cmd, args: {'result': json.dumps(node_list)}
 
     errors = check._check_elasticsearch_node_list(pods_by_name(pods))
     assert_error(''.join(errors), expect_error)
@@ -149,7 +144,8 @@ def test_check_elasticsearch_node_list(pods, node_list, expect_error):
 ])
 def test_check_elasticsearch_cluster_health(pods, health_data, expect_error):
     test_health_data = list(health_data)
-    check = canned_elasticsearch(task_vars_config_base, lambda cmd, args: json.dumps(test_health_data.pop(0)))
+    check = Elasticsearch(None, task_vars_config_base)
+    check.execute_module = lambda cmd, args: {'result': json.dumps(test_health_data.pop(0))}
 
     errors = check._check_es_cluster_health(pods_by_name(pods))
     assert_error(''.join(errors), expect_error)
@@ -174,7 +170,8 @@ def test_check_elasticsearch_cluster_health(pods, health_data, expect_error):
     ),
 ])
 def test_check_elasticsearch_diskspace(disk_data, expect_error):
-    check = canned_elasticsearch(task_vars_config_base, lambda cmd, args: disk_data)
+    check = Elasticsearch(None, task_vars_config_base)
+    check.execute_module = lambda cmd, args: {'result': disk_data}
 
     errors = check._check_elasticsearch_diskspace(pods_by_name([plain_es_pod]))
     assert_error(''.join(errors), expect_error)
--- a/roles/openshift_health_checker/test/fluentd_config_test.py
+++ b/roles/openshift_health_checker/test/fluentd_config_test.py
@@ -1,0 +1,357 @@
+import pytest
+
+from openshift_checks.logging.fluentd_config import FluentdConfig, OpenShiftCheckException
+
+
+def canned_fluentd_pod(containers):
+    return {
+        "metadata": {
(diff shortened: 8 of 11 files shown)


roles/openshift_health_checker/openshift_checks/logging/fluentd.py
```diff
@@ -35,9 +35,72 @@ def run(self, tmp, task_vars):
...
+        This means that, at least for now, if the following condition is met:
+
+            openshift_logging_fluentd_use_journal == True
+
+        then the value of the configured Docker logging driver should be "journald".
+        Otherwise, the value of the Docker logging driver should be "json-file".
+        Returns an error string if the above condition is not met, or None otherwise."""
+
+        pod_name = fluentd_pods[0]["metadata"]["name"]
+        use_journald = self._exec_oc("exec {} /bin/printenv USE_JOURNAL".format(pod_name), [], task_vars)
```
juanvallejo: @sosiouxme I decided to read the value of the env var rather than the ansible variable (`openshift_logging_fluentd_use_journal`) since I believe the ansible variable is not set by default (https://github.com/openshift/openshift-ansible/blob/master/inventory/byo/hosts.origin.example).

juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/logging/fluentd.py
sosiouxme: We have one extra problem here that just occurred to me. This check only runs against one master. It will correctly check whether the docker there has a matching journal driver, but it won't check any of the other hosts. This may be tricky to accomplish.

juanvallejo: Hm, we may have to have a separate fluentd check that does not inherit from `LoggingCheck`

juanvallejo:
@sosiouxme went ahead and moved the fluentd config check to its own file. PTAL

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @classmethod
+    def is_active(cls, task_vars):
+        return super(FluentdConfig, cls).is_active(task_vars)
+
+    def run(self, tmp, task_vars):
+        """Check that Fluentd has running pods, and that its logging config matches Docker's logging config."""
+
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
+        fluentd_pods, error = self.get_pods_for_component(
```
sosiouxme: This and other `oc` calls will only work on master, though. That's what makes this tricky. We could say to just look at the logging var to see what fluentd is supposed to have and not check what it actually has. Otherwise we're gonna have to do some really weird delegation to run `oc` against master while we're looking at everything else.

juanvallejo: I was hesitant to rely on the logging var since its value is not set by default, but I guess for now it might be the easiest starting point for this. Will update

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,83 @@
...
+    def check_logging_config(self, task_vars):
+        """Ensure that the configured Docker logging driver matches fluentd settings.
+        This means that, at least for now, if the following condition is met:
+
+            openshift_logging_fluentd_use_journal == True
+
+        then the value of the configured Docker logging driver should be "journald".
+        Otherwise, the value of the Docker logging driver should be "json-file".
+        Returns an error string if the above condition is not met, or None otherwise."""
+        use_journald = get_var(task_vars, "openshift_logging_fluentd_use_journal", default=True)
```
juanvallejo: @sosiouxme removed use of `ocutil` in favor of using the value of the ansible variable. Defaulted value to `True`, making the check assume `journalctl` as the default logging driver, unless explicitly set by a user to `False`. Not entirely sure about this, however.

juanvallejo:
aos-ci-test

juanvallejo:
@sosiouxme switched back to using `ocutil` if check is running on a master. PTAL

sosiouxme:
Ran into this while testing:

```
     Details:  check "fluentd_config":
               Invalid value received from command exec logging-fluentd-nfwxp /bin/printenv USE_JOURNAL: Unexpected error using `oc` to validate the logging stack components.
               Error executing `oc exec logging-fluentd-nfwxp /bin/printenv USE_JOURNAL`:
               [rc 1] /usr/local/bin/oc --config /etc/origin/master/admin.kubeconfig -n logging exec logging-fluentd-nfwxp /bin/printenv USE_JOURNAL
```

Unfortunately there doesn't seem to be any more helpful debug output.

sosiouxme:
But speaking of the above, why try to exec into the pod? You should have the pod definition right there to look at what the env vars are.

juanvallejo:
@sosiouxme 
> But speaking of the above, why try to exec into the pod? You should have the pod definition right there to look at what the env vars are.

Thanks, updated check / tests to do this instead

roles/openshift_health_checker/openshift_checks/fluentd_config.py
sosiouxme: This is correct

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,190 @@
...
+
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
+        config_error = self.check_logging_config(task_vars)
+        if config_error:
+            msg = ("The following Fluentd logging configuration problem was found:"
+                   "\n-------\n"
+                   "{}".format(config_error))
+            return {"failed": True, "changed": False, "msg": msg}
+
+        return {"failed": False, "changed": False, "msg": 'No problems found with Fluentd logging configuration.'}
```
sosiouxme: It's not, except for showing up in `-v` output.

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,190 @@
...
+    logging_namespace = None
+
+    @classmethod
+    def is_active(cls, task_vars):
+        return super(FluentdConfig, cls).is_active(task_vars)
+
+    def run(self, tmp, task_vars):
+        """Check that Fluentd has running pods, and that its logging config matches Docker's logging config."""
+
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
```
juanvallejo: Thanks, will stick to a default class field

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,190 @@
...
+        config_error = self.check_logging_config(task_vars)
+        if config_error:
+            msg = ("The following Fluentd logging configuration problem was found:"
+                   "\n-------\n"
```
juanvallejo: Right now, we seem to use it for similarly-worded error messages in [the](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py#L36) [logging](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_health_checker/openshift_checks/logging/curator.py#L33) [stack](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_health_checker/openshift_checks/logging/kibana.py#L45) [checks](https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_health_checker/openshift_checks/logging/fluentd.py#L34).

sosiouxme: I think that's a legacy from when the logging check was monolithic and you could get a bunch of different msgs for the different components under one check. Dividers probably don't make sense with them all split out in separate checks.

juanvallejo: > I think that's a legacy from when the logging check was monolithic and you could get a bunch of different msgs for the different components under one check. Dividers probably don't make sense with them all split out in separate checks.

ack, will go ahead and remove them from existing checks as well

roles/openshift_health_checker/openshift_checks/fluentd_config.py
rhcarvalho: I suspect on '-v' output it will appear in the middle of a big json object?
I'd rather have less noise. `failed: False` conveys the same info.

I have also mixed feelings about the "changed" key. I don't feel confident we use it consistently and correctly, and we don't really use it in any meaningful way. I remember I wanted to "be prepared" for supporting check-mode but thus far I feel we're only carrying extra code.

sosiouxme: You're right, I don't see a need for a msg here. No one will look at it. There was some place where it just made the logic cleaner to set one but there's no need most of the time.

"changed" response doesn't really have anything to do with check-mode. It's just a record that something was changed during this task, which gets summed up in the output at the end. Idempotency seems to be a big deal for Ansible users so reporting whether you actually changed something (e.g. installed a dependency) or not may matter to some. It does not seem critical to me but I'm not an admin. If we want a better way to track this we could set an instance variable and have the action plugin look at them to determine whether anything changed (even if an exception was thrown) and just keep all the "changed" response logic out of the actual checks.

juanvallejo: Will remove "changed" and "msg" keys

roles/openshift_health_checker/test/fluentd_config_test.py
```diff
@@ -0,0 +1,325 @@
...
+
+def canned_fluentd_config(exec_oc=None):
+    """Create a Fluentd check object with canned exec_oc method"""
+    check = FluentdConfig("dummy")  # fails if a module is actually invoked
+    if exec_oc:
+        check._exec_oc = exec_oc
+    return check
+
+
+def canned_fluentd_pod(containers=[]):
```
rhcarvalho: You have a flake8 error here. Not a problem in this case, but sometimes mutable containers in arguments cause hairy problems...
Seems that we never call the function without an argument, so why don't simply make it required? (s/=[]//)

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,190 @@
...
+
+    logging_namespace = "logging"
+
+    @classmethod
+    def is_active(cls, task_vars):
+        return super(FluentdConfig, cls).is_active(task_vars)
+
+    def run(self, tmp, task_vars):
+        """Check that Fluentd has running pods, and that its logging config matches Docker's logging config."""
+
```
rhcarvalho: extraneous empty line

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,190 @@
...
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default=self.logging_namespace)
+        config_error = self.check_logging_config(task_vars)
+        if config_error:
+            msg = ("The following Fluentd logging configuration problem was found:"
+                   "\n-------\n"
```
rhcarvalho: Did we decide for or against this separator?

roles/openshift_health_checker/test/fluentd_config_test.py
```diff
@@ -0,0 +1,325 @@
...
+        )],
+        {
+            "failed": True,
+            "result": "unexpected",
+        },
+        "json-file",
+        ['no environment variables'],
+    ),
+], ids=lambda argvals: argvals[0])
+def test_failed_check_logging_config_master(name, pods, response, logging_driver, extra_words):
```
rhcarvalho: What's the intent of this test?

roles/openshift_health_checker/test/fluentd_config_test.py
```diff
@@ -0,0 +1,325 @@
...
+    check.get_pods_for_component = get_pods
+    error = check.check_logging_config(task_vars)
+
+    if error:
+        for word in extra_words:
+            assert word in error
+
+        return
+
+    assert error is None
```
rhcarvalho: Seems more logical to assert this first, and only then look at extra_words?

roles/openshift_health_checker/test/fluentd_config_test.py
```diff
@@ -0,0 +1,325 @@
...
+
+    if failed:
+        assert error is not None
+
+        for word in extra_words:
+            assert word in error
+
+        return
+
+    assert error is None
```
rhcarvalho: Wouldn't it be more clear to have positive tests in one function and negative tests in another?
In this form one needs to interpret the params and the conditional in the test to understand what assertions are made.

I'd say normally we don't wan't conditional assertions in tests, that's a smell.

roles/openshift_health_checker/openshift_checks/fluentd_config.py
juanvallejo: Will remove, I believe we decided against it

juanvallejo:
@rhcarvalho thanks for the review, comments addressed

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,188 @@
...
+        use_journald = get_var(task_vars, "openshift_logging_fluentd_use_journal", default=True)
+
+        # if check is running on a master, retrieve all running pods
+        # and check any pod's container for the env var "USE_JOURNAL"
+        group_names = get_var(task_vars, "group_names")
+        if "masters" in group_names:
+            use_journald = self.check_fluentd_env_var(task_vars)
+
+        docker_info = self.execute_module("docker_info", {}, task_vars)
+        if not docker_info.get("info", False):
```
rhcarvalho:
```python
if "info" not in docker_info:
    ...
```

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,188 @@
...
+            return "No information was returned by the Docker client. Unable to verify the logging driver in use."
+
+        if not docker_info["info"].get("LoggingDriver", False):
```
rhcarvalho:
```python
if "LoggingDriver" not in docker_info["info"]:
    ...
```

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,188 @@
...
+            return "No logging driver information was returned by the Docker client."
+
+        logging_driver = docker_info["info"]["LoggingDriver"]
```
rhcarvalho: Perhaps it is not terribly useful to distinguish when `docker_info` is empty and when only the `LoggingDriver` is missing?

If so, then we can make the whole block shorter:

```python
docker_info = self.execute_module("docker_info", {}, task_vars)
try:
    logging_driver = docker_info["info"]["LoggingDriver"]
except KeyError:
    return "Unable to determine Docker logging driver."
```

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,188 @@
...
+                      '{{ "log-driver": "{driver}" }}\n\n'
+                      'See https://docs.docker.com/engine/admin/logging/json-file '
+                      'for more information.').format(driver=recommended_logging_driver)
+
+        return error
+
+    def check_fluentd_env_var(self, task_vars):
+        """Read and return the value of the 'USE_JOURNAL' environment variable on a fluentd pod."""
+        running_pods = self.running_fluentd_pods(task_vars)
+        pod_containers = running_pods[0]["spec"]["containers"]
```
rhcarvalho: Missing exception handling?

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,188 @@
...
+
+        if not pod_containers:
+            msg = ('There are no running containers on selected Fluentd pod "{}".\n'
+                   'Unable to calculate expected logging driver.').format(running_pods[0]["metadata"]["name"])
+            raise OpenShiftCheckException(msg)
+
+        pod_env = pod_containers[0]["env"]
```
rhcarvalho: Missing exception handling?

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,188 @@
...
+        if not pod_env:
+            msg = ('There are no environment variables set on the Fluentd container "{}".\n'
+                   'Unable to calculate expected logging driver.').format(pod_containers[0]["name"])
+            raise OpenShiftCheckException(msg)
+
+        for env in pod_env:
+            if env["name"] == "USE_JOURNAL":
+                return env["value"] != "false"
```
rhcarvalho: Missing exception handling?

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,188 @@
...
+    def running_fluentd_pods(self, task_vars):
+        """Return a list of running fluentd pods."""
+        fluentd_pods, error = self.get_pods_for_component(
+            self.execute_module,
+            self.logging_namespace,
+            "fluentd",
+            task_vars,
+        )
+        if error:
+            msg = 'Unable to retrieve any pods for the "fluentd" logging component: {}'.format(str(error))
```
rhcarvalho: No need for `str`, the string formatting will do the same.

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,188 @@
...
+            raise OpenShiftCheckException(msg)
+
+        running_fluentd_pods = [pod for pod in fluentd_pods if pod['status']['phase'] == 'Running']
+        if not len(running_fluentd_pods):
```
rhcarvalho: No need for `len`.

```python
if not running_fluentd_pods:
   ...
```

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,188 @@
...
+        try:
+            pod_output = self.exec_oc(
+                execute_module,
+                namespace,
+                "get pods -l component={} -o json".format(logging_component),
+                [],
+                task_vars
+            )
+        except OpenShiftCheckException as err:
+            if "false" not in str(err):
```
rhcarvalho: What is this about?!

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,188 @@
...
+                raise err
+
+        try:
+            pods = json.loads(pod_output)
+            if not pods or not pods.get('items'):
+                raise ValueError()
+        except ValueError:
```
rhcarvalho: Suggestion:

```python
try:
    pods = json.loads(pod_output)
    assert pods and 'items' in pods
except (ValueError, TypeError, AssertionError):
    return ...
```

Though logically equivalent, I believe it is easier to reason about `A and B`, than `not (not A or not B)`?

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,188 @@
...
+            # successful run but non-parsing data generally means there were no pods in the namespace
+            return None, 'There are no pods in the {} namespace. Is logging deployed?'.format(namespace)
+
+        return pods['items'], None
+
+    @staticmethod
+    def exec_oc(execute_module=None, namespace="logging", cmd_str="", extra_args=None, task_vars=None):
```
rhcarvalho: Save for white space differences in the docstring and a bug that was fixed in https://github.com/openshift/openshift-ansible/commit/5f51278dee925c47055aacbb8ffac5e351ea7ee2, this is identical to [LoggingCheck.exec_oc](https://github.com/openshift/openshift-ansible/blob/a41fc449c3e56646365701e5b7a40b29eb9e3b67/roles/openshift_health_checker/openshift_checks/logging/logging.py#L67).

Let's not duplicate this method please, nor reintroduce the bug.

juanvallejo:

roles/openshift_health_checker/test/fluentd_config_test.py
```diff
@@ -0,0 +1,365 @@
...
+    check.get_pods_for_component = get_pods
+
+    with pytest.raises(OpenShiftCheckException) as error:
+        check.check_logging_config(task_vars)
+
+    assert error is not None
+    for word in extra_words:
+        assert word in str(error)
+
+        return
```
rhcarvalho: This looks misplaced, it will exit the loop on the first run.

juanvallejo: Thanks for catching that. Fixed

roles/openshift_health_checker/test/fluentd_config_test.py
```diff
@@ -0,0 +1,365 @@
...
+    check.get_pods_for_component = get_pods
+
+    with pytest.raises(OpenShiftCheckException) as error:
+        check.check_logging_config(task_vars)
+
+    assert error is not None
+    for word in extra_words:
+        assert word in str(error)
+
+    return
```
rhcarvalho: Typically we don't need this. Unless we have reasons to make it clear (e.g. when a function has multiple return points), the language itself doesn't require every function to have a return statement.

The test function's purpose is to raise an exception when the test fails, or do nothing otherwise. The return value is ignored.

juanvallejo: hm, not sure why I still felt compelled to add a return there; mindset was in another language :)

rhcarvalho:
Yes, it is understandable :-)

juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,157 @@
...
+class FluentdConfig(OpenShiftCheck):
+    """Module that checks logging configuration of an integrated logging Fluentd deployment"""
+    name = "fluentd_config"
+    tags = ["health"]
+
+    logging_namespace = "logging"
+
+    @classmethod
+    def is_active(cls, task_vars):
+        logging_deployed = get_var(task_vars, "openshift_hosted_logging_deploy", default=True)
```
rhcarvalho: I believe the default is `False`, see [playbooks/adhoc/openshift_hosted_logging_efk.yaml#L16](https://github.com/openshift/openshift-ansible/blob/e432f8a35323e8603f0f58b70d996f7487703ec1/playbooks/adhoc/openshift_hosted_logging_efk.yaml#L16)

juanvallejo:
@rhcarvalho thanks, review comment addressed

roles/openshift_health_checker/openshift_checks/fluentd_config.py
rhcarvalho: May or may not have any consequence to this PR, but in any case, FWIW I see this variable was deprecated 14 days ago:

- [roles/openshift_logging/README.md#L55](https://github.com/openshift/openshift-ansible/blame/d13c11aceb2c2ccf2c1db1ead9f545d7b0891cfc/roles/openshift_logging/README.md#L55)
- [roles/openshift_logging_fluentd/tasks/main.yaml#L18-L20](https://github.com/openshift/openshift-ansible/blob/d13c11aceb2c2ccf2c1db1ead9f545d7b0891cfc/roles/openshift_logging_fluentd/tasks/main.yaml#L18-L20)

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,157 @@
...
+Module for performing checks on a Fluentd logging deployment configuration
+"""
+
+import json
+
+from openshift_checks import OpenShiftCheck, get_var, OpenShiftCheckException
+from openshift_checks.logging.logging import LoggingCheck
+
+
+class FluentdConfig(OpenShiftCheck):
```
rhcarvalho: Out of curiosity, why isn't this a subclass of the `LoggingCheck`? Perhaps we could avoid e.g. detecting whether logging is enabled and determining the logging namespace in two different files.

roles/openshift_health_checker/openshift_checks/fluentd_config.py
```diff
@@ -0,0 +1,157 @@
...
+
+        pod_env = pod_containers[0].get("env")
+        if not pod_env:
+            msg = ('There are no environment variables set on the Fluentd container "{}".\n'
+                   'Unable to calculate expected logging driver.').format(pod_containers[0].get("name"))
+            raise OpenShiftCheckException(msg)
+
+        for env in pod_env:
+            if env["name"] == "USE_JOURNAL":
+                return env.get("value", "false") != "false"
```
rhcarvalho: I don't know how this was commonly used nor how the env var gets parsed by the main process in the Fluentd container, but perhaps worth asking: aren't there other values that also mean "false"? E.g. 0, no, etc.

juanvallejo: I don't know; I'm unable to find any docs on this. Not sure how feasible it would be to parse every possible indication of "don't use jorurnald" though

cc @sosiouxme

roles/openshift_health_checker/openshift_checks/fluentd_config.py
juanvallejo: Not sure if we also want to limit this check to only run on the first master. If we don't, we would have to override the parent's `is_active` method anyway

cc @sosiouxme

juanvallejo:
From: https://github.com/openshift/openshift-ansible/pull/4592#discussion_r128547968
> May or may not have any consequence to this PR, but in any case, FWIW I see this variable was deprecated 14 days ago:

Don't think it hurts to check for it anyway; that being said, we will always expect Fluentd to use journald, unless the `USE_JOURNAL` env var in the pod says otherwise. Can't remember if that is an assumption we want to stick with. cc @sosiouxme

roles/openshift_health_checker/openshift_checks/fluentd_config.py
sosiouxme: It's not just deprecated, it's obviated... with 3.6 fluentd looks at docker config to use the same driver. That makes this whole check kind of pointless for 3.6. And I'm not sure what we plan to tell customers about running checks against earlier versions?

roles/openshift_health_checker/openshift_checks/fluentd_config.py
sosiouxme: Since individual nodes can have different logging config the idea was to check it on all the nodes; so yeah, it would need to override `is_active`. We could still reuse functionality from `LoggingCheck` in various was if it were more involved than this.

juanvallejo: Went ahead and subclassed FluentdConfig from LoggingCheck

juanvallejo:
@sosiouxme @rhcarvalho added openshift version check to `is_active` method; will skip if `> 3.5`

rhcarvalho:
@juanvallejo pylint error in: https://travis-ci.org/openshift/openshift-ansible/builds/256240040

roles/openshift_health_checker/openshift_checks/logging/curator.py
```diff
@@ -12,12 +12,10 @@ class Curator(LoggingCheck):
     name = "curator"
     tags = ["health", "logging"]
 
-    logging_namespace = None
-
     def run(self, tmp, task_vars):
         """Check various things and gather errors. Returns: result as hash"""
 
-        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
+        super(Curator, self).logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
```
rhcarvalho:
```python
self.logging_namespace
```

Python does the attribute look up, and then we want to change the instance attribute, not the parent class' attribute.

In an interpreter run time where two instances share the same parent class, we don't want the `run` method of one instance to affect the other.

roles/openshift_health_checker/openshift_checks/logging/elasticsearch.py
```diff
@@ -15,12 +15,10 @@ class Elasticsearch(LoggingCheck):
     name = "elasticsearch"
     tags = ["health", "logging"]
 
-    logging_namespace = None
-
     def run(self, tmp, task_vars):
         """Check various things and gather errors. Returns: result as hash"""
 
-        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
+        super(Elasticsearch, self).logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
```
rhcarvalho:
```python
self.logging_namespace
```

`super()` can be a weird thing to understand -- https://docs.python.org/2/library/functions.html#super
Focus on the two moments when you need it:

1. Getting multiple inheritance right;
2. Accessing methods / attributes on the parent class without naming the parent class explicitly, typically when you are overriding a method with the same name (we do this a lot e.g. with `is_active`).

In all other cases, just let the language do the name look up in the standard way, `obj.method(...)` will look for `method` in `obj` and if not defined will recursively traverse up the class hierarchy until it finds `method` or eventually raise `AttributeError`.

roles/openshift_health_checker/openshift_checks/logging/fluentd.py
```diff
@@ -13,12 +13,10 @@ class Fluentd(LoggingCheck):
     name = "fluentd"
     tags = ["health", "logging"]
 
-    logging_namespace = None
-
     def run(self, tmp, task_vars):
         """Check various things and gather errors. Returns: result as hash"""
 
-        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
+        super(Fluentd, self).logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
```
rhcarvalho: `self.`

roles/openshift_health_checker/openshift_checks/logging/fluentd_config.py
```diff
@@ -0,0 +1,144 @@
...
+"""
+
+from openshift_checks import get_var, OpenShiftCheckException
+from openshift_checks.logging.logging import LoggingCheck
+from openshift_checks.ovs_version import OvsVersion
+
+
+class FluentdConfig(LoggingCheck):
+    """Module that checks logging configuration of an integrated logging Fluentd deployment"""
+    name = "fluentd_config"
```
rhcarvalho: @juanvallejo @sosiouxme, now that I noticed we have two checks -- `fluentd` and `fluentd_config` -- is it justifiable to have them separate? Is the distinction clear enough to explain users?

roles/openshift_health_checker/openshift_checks/logging/fluentd_config.py
```diff
@@ -0,0 +1,144 @@
...
+    tags = ["health"]
+
+    @classmethod
+    def is_active(cls, task_vars):
+        logging_deployed = get_var(task_vars, "openshift_hosted_logging_deploy", default=False)
+
+        # determine OpenShift version
+        try:
+            ovs = OvsVersion(execute_module="")
```
rhcarvalho: Really weird dependency :)

rhcarvalho: Man, this got me really confused... Why do we even have a `OvsVersion` check and we check the Open vSwitch version using the `package_version` check?!

roles/openshift_health_checker/openshift_checks/logging/fluentd_config.py
```diff
@@ -0,0 +1,144 @@
...
+            version = float(ovs._get_openshift_version(task_vars))
+        except ValueError:
+            # if failed to parse OpenShift version, perform check anyway (if logging enabled)
+            return logging_deployed
+
+        return logging_deployed and version < 3.6
```
rhcarvalho: Using floats to compare versions is buggy. Version-wise, 3.10 > 3.6, but numeric-wise 3.10 < 3.6. Suggestion: use a version tuple.

roles/openshift_health_checker/openshift_checks/logging/kibana.py
```diff
@@ -22,12 +22,10 @@ class Kibana(LoggingCheck):
     name = "kibana"
     tags = ["health", "logging"]
 
-    logging_namespace = None
-
     def run(self, tmp, task_vars):
         """Check various things and gather errors. Returns: result as hash"""
 
-        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
+        super(Kibana, self).logging_namespace = get_var(task_vars, "openshift_logging_namespace", default="logging")
```
rhcarvalho: `self.`

roles/openshift_health_checker/openshift_checks/logging/fluentd_config.py
juanvallejo: Will move `_get_openshift_version` to the parent class.

> Man, this got me really confused... Why do we even have a OvsVersion check and we check the Open vSwitch version using the package_version check?!

The `OvsVersion` check runs as a diagnostic check, while the `package version` Open vSwitch check runs as part of pre-flight checks; they both check for the same thing though

juanvallejo:
@rhcarvalho thanks for the feedback; review comments addressed

roles/openshift_health_checker/openshift_checks/logging/fluentd_config.py
juanvallejo: I am okay with combining these two checks; will wait for @sosiouxme 's input

rhcarvalho:
> EDIT: not sure it is a good idea to just group the ovs check with the logging checks; will just make ovs openshift_version method static

Since it is static, we could make it a function or on the same direction of moving `task_vars` and `get_var` into `OpenShiftCheck`, make it a static method there.

rhcarvalho:
@juanvallejo PTAL at the Travis failure

juanvallejo:
@rhcarvalho 
> Since it is static, we could make it a function or on the same direction of moving task_vars and get_var into OpenShiftCheck, make it a static method there.

Went ahead and moved the method to `OpenShiftCheck`

juanvallejo:
@rhcarvalho or @sosiouxme wondering if you could give this one last pass, thanks

juanvallejo:
@sosiouxme @rhcarvalho had to rebase with latest master once https://github.com/openshift/openshift-ansible/pull/4485#issuecomment-318162643 merged. Latest commit updates this check to take changes introduced into account. ptal

rhcarvalho:
@juanvallejo Travis is red

roles/openshift_health_checker/test/elasticsearch_test.py
```diff
@@ -5,13 +5,17 @@
...
-    if exec_oc:
-        check._exec_oc = exec_oc
-    return check
+
+# def canned_elasticsearch(task_vars=None, exec_oc=None):
+#     """Create an Elasticsearch check object with canned exec_oc method"""
+#     check = Elasticsearch(None, task_vars or {})  # fails if a module is actually invoked
+#     if exec_oc:
+#         check._exec_oc = exec_oc
+#     return check
```
rhcarvalho: If we don't use this we can as well delete it ;)

roles/openshift_health_checker/test/elasticsearch_test.py
```diff
@@ -50,10 +54,10 @@ def assert_error(error, expect_error):
 
 
 def test_check_elasticsearch():
-    assert 'No logging Elasticsearch pods' in canned_elasticsearch().check_elasticsearch([])
+    assert 'No logging Elasticsearch pods' in elasticsearch().check_elasticsearch([])
```
rhcarvalho: When we make it a Pytest Fixture, we can use it by simply listing it as an argument to the test function: https://docs.pytest.org/en/latest/fixture.html#fixtures-as-function-arguments

But since it seems the intention is to call the function with different args, then perhaps it doesn't even need to be a fixture, but simply a plain function?

roles/openshift_health_checker/test/fluentd_config_test.py
```diff
@@ -4,8 +4,8 @@
 
 
 @pytest.fixture
-def fluentd_config():
-    return FluentdConfig("dummy")
+def fluentd_config(execute_module, task_vars):
```
rhcarvalho: Pytest Fixtures typically do not take arbitrary arguments...
It can take a magical `request` argument https://docs.pytest.org/en/latest/fixture.html#fixtures-can-introspect-the-requesting-test-context, or reference other fixtures https://docs.pytest.org/en/latest/fixture.html#modularity-using-fixtures-from-a-fixture-function

roles/openshift_health_checker/test/fluentd_test.py
```diff
@@ -103,7 +100,7 @@ def assert_error(error, expect_error):
     ),
 ])
 def test_get_fluentd_pods(pods, nodes, expect_error):
-    check = canned_fluentd(exec_oc=lambda cmd, args: json.dumps(dict(items=nodes)))
-
+    check = fluentd()
```
rhcarvalho: Hmm, wouldn't it be way simpler to just write `Fluentd()` (with an uppercase F) here?! What's the point of the fixture?

roles/openshift_health_checker/test/elasticsearch_test.py
juanvallejo: Ah, knew I would forget this :)

roles/openshift_health_checker/test/elasticsearch_test.py
juanvallejo: > But since it seems the intention is to call the function with different args, then perhaps it doesn't even need to be a fixture, but simply a plain function?

Agree, will update

juanvallejo:
@rhcarvalho thanks for the review, comments addressed

roles/openshift_health_checker/test/kibana_test.py
```diff
@@ -68,7 +65,7 @@ def assert_error(error, expect_error):
     ),
 ])
 def test_check_kibana(pods, expect_error):
-    check = canned_kibana()
+    check = kibana()
```
rhcarvalho: Similar to another review today, maybe this is just `Kibana()` and no need for a fixture.

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -105,6 +105,28 @@ def get_var(self, *keys, **kwargs):
...
+        components = openshift_image_tag.split(".")
+        if not components or len(components) < 2:
+            msg = "An invalid version of OpenShift was found for this host: {}"
+            raise OpenShiftCheckException(msg.format(openshift_image_tag))
+
+        if components[0] in openshift_major_release_version:
+            components[0] = openshift_major_release_version[components[0]]
+
+        components = [int(x) for x in components[:2]]
+        return tuple(components)
```
rhcarvalho: If you want to return a tuple, no need to create a list first...

```python
components = tuple(int(x) for x in components[:2])
return components
```

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -105,6 +105,28 @@ def get_var(self, *keys, **kwargs):
...
+
+        components = openshift_image_tag.split(".")
+        if not components or len(components) < 2:
+            msg = "An invalid version of OpenShift was found for this host: {}"
+            raise OpenShiftCheckException(msg.format(openshift_image_tag))
+
+        if components[0] in openshift_major_release_version:
+            components[0] = openshift_major_release_version[components[0]]
+
+        components = [int(x) for x in components[:2]]
```
rhcarvalho: Can't `openshift_image_tag` carry a non-numeric version component, like `rcN` or `beta` or `alpha`, etc?

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -105,6 +105,28 @@ def get_var(self, *keys, **kwargs):
             raise OpenShiftCheckException("'{}' is undefined".format(".".join(map(str, keys))))
         return value
 
+    @staticmethod
+    def get_openshift_version(openshift_image_tag):
```
rhcarvalho: Maybe this should have a more specific name, since it strips out version information? It is returning MAJOR.MINOR and throwing away the rest.

roles/openshift_health_checker/openshift_checks/__init__.py
juanvallejo: not sure
cc @sosiouxme

sosiouxme: Yeah it can pretty much be anything. It's an opaque string.

roles/openshift_health_checker/openshift_checks/logging/fluentd_config.py
sosiouxme: `fluentd` would run on the first master only, while `fluentd_config` could theoretically run on all the nodes, but I'm forgetting now how the impl goes and maybe it doesn't. If is_active is different it makes sense to keep separate...

sosiouxme:
Some refactoring on the way that parallels some of this: https://github.com/openshift/openshift-ansible/pull/4913

roles/openshift_health_checker/openshift_checks/__init__.py
juanvallejo: How should we handle non numeric version values; maybe a try, except?

sosiouxme: I spoke too soon. Actually `openshift_version` asserts that it look like a version string

    roles/openshift_version/tasks/main.yml:      - "{{ openshift_image_tag|match('(^v\\d+\\.\\d+[\\.\\d+]*(-\\d+)?$)') }}"

So we should be confident that it has a v and two numeric components.

(Sorry; I spent too long looking at this at some point in the past, figured it all out, then forgot it.)

juanvallejo: Sounds good, I am only focusing on the first two numeric components of the string as it is. Not sure if there is a better var to base this on, but will take a look at sample inventory files

juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/logging/fluentd_config.py
```diff
@@ -0,0 +1,138 @@
...
+class FluentdConfig(LoggingCheck):
+    """Module that checks logging configuration of an integrated logging Fluentd deployment"""
+    name = "fluentd_config"
+    tags = ["health"]
+
+    def is_active(self):
+        logging_deployed = self.get_var("openshift_hosted_logging_deploy", default=False)
+
+        try:
+            version = self.get_major_minor_version(self.get_var("openshift_image_tag"))
```
rhcarvalho: Note that this can raise `OpenShiftCheckException`. We should protect the action plugin call to `is_active` with a `try... except` block for cases like this, to prevent the action plugin run from blowing up. I can include that in my current work branch.

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: For origin the expression includes arbitrary letters, but the first two components should be numeric.

[`openshift_image_tag|match('(^v?\\d+\\.\\d+\\.\\d+(-[\\w\\-\\.]*)?$)')`](https://github.com/openshift/openshift-ansible/blob/4a442d87006bdd535eb5035d122a6b5226555da2/roles/openshift_version/tasks/main.yml#L52)

We should be all set, thanks.

roles/openshift_health_checker/openshift_checks/logging/fluentd_config.py
sosiouxme: Yeah, good idea. While you're at it, probably best to add an `except Exception` to catch any dumb bug we leave on one check so the others still run.

juanvallejo:
@rhcarvalho tagging this for [merge] per scrum discussion

juanvallejo:
re[test]

sosiouxme:
[ERROR] Status 'fedora/25/atomic' is in the 'failure' state. This status must succeed for a merge.

Wait - what? That's new.

sdodson:
aos-ci-test

sdodson:
bot, retest this

PR #4594, opened 2017-06-26
Adding pv encryption and registry backed encryption for s3.

kwoodson:


sdodson:
Is this a blocker?

BigN:
Not a blocker but the number of dedicated enterprise customers asking for
this feature is quickly increasing. Would be great if we can have this in
3.6 (and not wait until 3.7) so when we upgrade we can enable encryption.

On Jun 26, 2017, at 8:41 PM, Scott Dodson <notifications@github.com> wrote:

Is this a blocker?

—
You are receiving this because your review was requested.
<https://github.com/openshift/openshift-ansible/pull/4594#issuecomment-311220256>,
or mute the thread

joelddiaz:
Just in case it isn't clear: doing an in-place upgrade will not enable encryption for the root/boot volumes.

sdodson:
Since this is still WIP we'll hold this until after we fork 3.6 at code freeze.

sdodson:
aos-ci-test

sdodson:
The 3.6 images and repos are all fouled up.

sdodson:
aos-ci-test

sdodson:
[merge]

PR #4619, opened 2017-06-28
[master] evalute groups when running etcd upgrade from byo/openshift-cluster/upgrades/upgrade_etcd.yml

ingvagabund:
When running etcd upgrade via ``playbooks/byo/openshift-cluster/upgrades/upgrade_etcd.yml``, only the basic groups are initialized. Causing the playbook fail due to `oo_etcd_hosts_to_backup` group missing.

Tested on 3.6 cluster, the playbook ended successfully in 3m6.848s.

Bug: 1460617

3.5 backport: https://github.com/openshift/openshift-ansible/pull/4621
3.4 backport: https://github.com/openshift/openshift-ansible/pull/4622

### DIFF
--- a/playbooks/byo/openshift-cluster/upgrades/upgrade_etcd.yml
+++ b/playbooks/byo/openshift-cluster/upgrades/upgrade_etcd.yml
@@ -1,4 +1,6 @@
 ---
 - include: ../initialize_groups.yml
 
+- include: ../../../common/openshift-cluster/evaluate_groups.yml
+
 - include: ../../../common/openshift-cluster/upgrades/etcd/main.yml


ingvagabund:
aos-ci-test

sdodson:
I was a bit hasty on merging the 1.5 PR so I'll go ahead and merge this too. But I wonder if this shouldn't be moved into `playbooks/common/openshift-cluster/upgrades/etcd/main.yml` instead?

ingvagabund:
> I wonder if this shouldn't be moved into playbooks/common/openshift-cluster/upgrades/etcd/main.yml instead?

I did want to separate `initialize_groups.yml` and `evaluate_groups.yml`. We should run both playbooks at the same file. So we have only one point of failure.

mtnbikenc:
> I did not want to separate initialize_groups.yml and evaluate_groups.yml. We should run both playbooks at the same file. So we have only one point of failure.

I guess this would be good to discuss and document in our style guide.  In the refactoring I've done, I've tried to keep the flow of plays/tasks to do what needed to be done in byo separate from what needed to be done in common.  The idea was to do whatever initialization was needed in byo (or other platform types) and then go to common which would run anything needed there.  With having a second call to common/ playbooks here, we would need to duplicate that second call in any other platform playbooks.  I know this brings up the point about supported/unsupported and the issue of having byo in the first place.  Another topic.

PR #4623, opened 2017-06-28
attach leases via the first master only and only once

ingvagabund:
- move openshift-etcd/migrate.yml from byo to common and keep just the entry point
- replace std_include with essential plays (e.g. no need to detect openshift version)
- delegate the ttl re-attaching to the first master

Bug: 1463494


ingvagabund:
aos-ci-test

sdodson:
CI jobs won't test this code so i've killed that job and merged manually.

PR #4624, opened 2017-06-28
increase implicit 300s default timeout to explicit 600s

ingvagabund:
Due to https://bugzilla.redhat.com/show_bug.cgi?id=1441994 a node can reboot slowing. Causing in timeout in ansible `wait_for` module. The default is 300s which is not always enough.

Bug: 1455836

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/docker/restart.yml
+++ b/playbooks/common/openshift-cluster/upgrades/docker/restart.yml
@@ -24,4 +24,5 @@
     state: started
     delay: 10
     port: "{{ openshift.master.api_port }}"
+    timeout: 600
   when: inventory_hostname in groups.oo_masters_to_config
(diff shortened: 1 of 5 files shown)


ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

sdodson:
aos-ci-test

ingvagabund:
@kwoodson I am not to familiar with the technical details. This is basically the simplest hack. The issue needs to be resolved in the Kubernetes/OpenShift first. Once it is done, we can undo this change or switch to the polling you mentioned.

sdodson:
@kwoodson It already polls, does it not?

kwoodson:
@sdodson, yeah, you are right.  This does poll.  I thought we were just waiting on a timeout.

My question as to why this takes so long is still valid.  We are waiting up to 10 minutes for a port to become available.  That seems like a really long time.

PR #4635, opened 2017-06-28
Add a retry to the docker restart handler

sdodson:
Attempt to workaround this https://gist.github.com/sdodson/5967210c5482de1c31b3685d4378e7d6 which happened against openshift-ansible-3.6.126.1-1

The start task doesn't seem to have been triggered.


mtnbikenc:
Looks good if this is due to some new race condition with restarting docker.  It is unclear as to the underlying reason for needing to try the restart again.

sdodson:
aos-ci-test

PR #4636, opened 2017-06-28
openshift_storage_nfs_lvm: fix with_sequence

jfchevrette:
Workaround known bug with with_sequence https://github.com/ansible/ansible/pull/22989#pullrequestreview-29199242

Should be fixed in a later ansible release but for now we can use the (less readable) string-like parameters


jfchevrette:
Pushed new PR #4683 which include this change and more. Closing this one.

PR #4638, opened 2017-06-28
add check for required installation ports

juanvallejo:
Related Trello card: https://trello.com/c/SlxFU8I5

Adds a pre-flight check that attempts to bind to a given set of
required of required ports. If any ports are already in use, they
are displayed along with the name of the service that requires 
the port.

*Sample Output*
```
Failure summary:

  1. Host:     34.227.83.160
     Play:     Run OpenShift health checks
     Details:  check "port_availability":
               The following ports are required for a successful OpenShift installation,
               but an error occurred attempting to verify their availability:

                 - 53 (required by "dnsmasq")
                   error: Port is in use by another process.

                 - 22 (required by "sample")
                   error: Port is in use by another process.


               Please refer to the following article for more information
               on freeing the required ports listed above:

               https://access.redhat.com/solutions/45294

You may choose to configure or disable failing checks by
setting Ansible variables. To disable those above:

    openshift_disable_check=port_availability

Consult check documentation for configurable variables.
Variables can be set in the inventory or passed on the
command line using the -e flag to ansible-playbook.
```

*TODO*
- [ ] add tests

### DIFF
--- a/roles/openshift_health_checker/library/portavail.py
+++ b/roles/openshift_health_checker/library/portavail.py
@@ -1,0 +1,51 @@
+#!/usr/bin/python
+"""Attempt to bind to a given set of ports, or return errors encountered"""
+
+import socket
+import errno
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def main():
+    """Bind to a given set of ports and return any errors this might cause."""
+    module = AnsibleModule(
+        argument_spec=dict(
+            # ports is a list of dicts consisting of a port_name and a port field:
+            #   [
+            #     {
+            #       "port_name": "mysql",
+            #       "port": 3306,
+            #     }
+            #   ]
+            ports=dict(type="list", required=True),
+        ),
+    )
+
+    errors = list()
+
+    ports = module.params["ports"]
+    for port in ports:
+        if not port.get("port"):
+            continue
+
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        try:
+            sock.bind(('localhost', port["port"]))
+        except socket.error as error:
+            if error.errno == errno.EADDRINUSE:
+                error = "Port is in use by another process."
+            port["error"] = str(error)
+            errors.append(port)
+        finally:
+            sock.close()
+
+    module.exit_json(
+        changed=False,
+        failed=len(errors),
+        error_ports=errors,
+    )
+
+
+if __name__ == '__main__':
+    main()
--- a/roles/openshift_health_checker/openshift_checks/port_availability.py
+++ b/roles/openshift_health_checker/openshift_checks/port_availability.py
@@ -1,0 +1,49 @@
+# pylint: disable=missing-docstring
+from openshift_checks import OpenShiftCheck, get_var
+
+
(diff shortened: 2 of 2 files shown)


roles/openshift_health_checker/openshift_checks/port_availability.py
```diff
@@ -0,0 +1,48 @@
...
+                        '  - {} (required by "{}")'.format(
+                            str(p.get("port", "n/a")),
+                            p.get("port_name", "n/a")
+                        )
+                        for p in failed_ports
+                    ]
+                ),
+                ('Please refer to the following article for more information\n'
+                 'on required OpenShift ports:\n\n'
+                 'https://access.redhat.com/solutions/45294')
```
juanvallejo: @sosiouxme was not sure if I should link a customer portal article, or provide a command line solution to find the process currently using a port. Also, found this alternative customer portal article that may be relevant as well: https://access.redhat.com/solutions/1520653

juanvallejo:
@rhcarvalho @sosiouxme for my initial approach, I looked into using [psutil](http://pythonhosted.org/psutil/) to iterate through all net connections, find the one bound to the required port (if port was in use), and return its process name and id. `psutil` did not seem to be available by default on my ec2 instance however, and I was hesitant about adding a new python dependency.

The other approach could be to avoid installing anything, just use netstat, grep for the port we are looking for, parse out the process id, and retrieve its name from that.

Alternatively, we could leave this patch to what it currently does and simply suggest a customer portal article, or provide a command / script to find the process currently bound to a required port. WDYT?

roles/openshift_health_checker/library/portavail.py
```diff
@@ -0,0 +1,47 @@
...
+    errors = list()
+
+    ports = module.params["ports"]
+    for port in ports:
+        if not port.get("port"):
+            continue
+
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        try:
+            sock.bind(('localhost', port["port"]))
```
rhcarvalho: For privileged ports like 53 we'd need to make sure this is run by root.

roles/openshift_health_checker/library/portavail.py
```diff
@@ -0,0 +1,47 @@
...
+        except socket.error as error:
+            port["error"] = str(error)
+            errors.append(port)
+        sock.close()
```
rhcarvalho: Perhaps this should be

```python
finally:
    sock.close()
```

This check would do worse than no check if it clogs up ports.

juanvallejo:
> A clear case against it is that the install playbook should be idempotent. Running it in a loop should 
> consistently leave the system at a known state.
> With this check in, a second run of the installer would fail because dnsmasq itself should be already 
> installed and running and listening on port 53

Then maybe we could shift the approach to ensure that either:
  - port `53` is open (no installation has taken place)
  - port `53` is in use by the process we expect (dnsmasq in this case)

That way, it can be run pre-installation and warn the user if a required port will cause installation problems, or run as a health check after the fact and ensure that the right process is bound to it.

WDYT?

PR #4642, opened 2017-06-29
upgrade: fix name for the etcd system container

giuseppe:
Fixes this error when running upgrade_etcd.yml:

The conditional check 'etcd_container_version | default('99') |
version_compare(etcd_upgrade_version,'<')' failed. The error was:
Version comparison: LooseVersion instance has no attribute 'version'

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/etcd/upgrade.yml
+++ b/playbooks/common/openshift-cluster/upgrades/etcd/upgrade.yml
@@ -36,7 +36,7 @@
       - not openshift.common.is_etcd_system_container | bool
 
     - name: Record containerized etcd version (runc)
-      command: runc exec etcd_container rpm -qa --qf '%{version}' etcd\*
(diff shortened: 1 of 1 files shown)


giuseppe:
aos-ci-test

playbooks/common/openshift-cluster/upgrades/etcd/upgrade.yml
```diff
@@ -36,7 +36,7 @@
       - not openshift.common.is_etcd_system_container | bool
 
     - name: Record containerized etcd version (runc)
-      command: runc exec etcd_container rpm -qa --qf '%{version}' etcd\*
+      command: runc exec etcd rpm -qa --qf '%{version}' etcd\*
```
ingvagabund: We should read the name from a variable. Given this is not the only occurrence of the literal, it is fine.

ingvagabund:
aos-ci-test

PR #4645, opened 2017-06-29
Correct version comparisons to ensure proper fact evaluation

mtnbikenc:
When a user supplies the openshift_image_tag inventory fact as '3.5',
the evaluation of `version >= LooseVersion('3.5.0')` will return False.
Changing the evaluation against only the X.X release will properly
return True.

PR #4650, opened 2017-06-29
[3.4] Correct version comparisons to ensure proper evaluation

mtnbikenc:
When a user supplies the openshift_image_tag inventory fact as '3.5',
the evaluation of `version >= LooseVersion('3.5.0')` will return False.
Changing the evaluation against only the X.X release will properly
return True.

Backports #4645


mtnbikenc:
aos-ci-test

mtnbikenc:
[test]

PR #4653, opened 2017-06-30
run etcd_container with type:spc_t label

ingvagabund:
When the etcd cluster is dedicated (deployed on non-master nodes), the `docker exec etcdctl backup` fails due to SELinux AVC denial. Labelling the running container with `type:spc_t` no longer generates the denial.

Bug: 1460959

Tested on 3.6 OCP cluster with dedicated etcd cluster.


ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ashcrow:
@ingvagabund To verify, a super privileged container is **required** and not a specific mounting or other context (Example: http://www.projectatomic.io/blog/2016/03/dwalsh_selinux_containers/)

PR #4660, opened 2017-06-30
Updating to use openshift.master.etcd_hosts for etcd servers for apis…

ewolinetz:
…erver

Should address https://bugzilla.redhat.com/show_bug.cgi?id=1466727
We will instead use the list of etcd_hosts which should resolve to actual hostnames not just the IP of the master

### DIFF
--- a/roles/openshift_service_catalog/tasks/install.yml
+++ b/roles/openshift_service_catalog/tasks/install.yml
@@ -98,6 +98,7 @@
     cpu_requests: none
     memory_request: none
     cors_allowed_origin: localhost
+    etcd_servers: "{{ openshift.master.etcd_urls | join(',') }}"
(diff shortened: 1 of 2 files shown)


sdodson:
I think if we're calling this playbook on its own we may need to add a dependency on openshift_master_facts. Which way did you test it? playbooks/byo/config.yml or playbooks/byo/openshift-cluster/service-catalog.yml?

ewolinetz:
will add in fix for https://bugzilla.redhat.com/show_bug.cgi?id=1466233 on here since the same files are touched

ewolinetz:
@sdodson i tested it with config.yml since that's the current intended path... i'll double check with the bolt on playbook as well

ewolinetz:
second commit addresses bz1466233

ewolinetz:
@sdodson verifying the bolt-on scenario now

PR #4664, opened 2017-07-03
Fix service catalog image prefixes

sdodson:
According to dist git the image name for service catalog is `openshift3/ose-service-catalog`
And fully qualify the asb image name


ewolinetz:
aos-ci-test

PR #4665, opened 2017-07-03
verify the etcd_container is running with spc_t label before upgrading to 3.6

ingvagabund:
The `etcd_container` must be run with `--security-opt label=type:spc_t` option in order to run `docker exec etcdctl backup ...` command. Given the option can be set automatically, check if the option is set before running the upgrade at least.

Unfortunately, it takes about 3 minutes to reach the check.


sdodson:
Can or should we just attempt to relabel it?

ingvagabund:
It needs to be done inside the etcd_container.service file. Otherwise the label gets lost once the container is restarted/stopped.

ingvagabund:
Closin in favor of https://github.com/openshift/openshift-ansible/pull/4674

PR #3943, opened 2017-04-18
Stop all services prior to upgrading, start all services after

sdodson:
Forward ports #3881

PR #4666, opened 2017-07-03
pre-pull images before stopping docker

ingvagabund:
For the containerized deployment, during the node upgrade, the docker is stopped before upgrading and started almost at the end of the process. Unfortunately, one (or two) docker images are pulled during the process. Given the docker is stopped, no image can be pulled and the node upgrade fails.

Bug: 1466095

Originally caused by https://github.com/openshift/openshift-ansible/pull/3943 by removing the `not openshift.common.is_containerized | bool` condition when including `docker/upgrade.yml`.

### DIFF
--- a/roles/openshift_node_upgrade/tasks/main.yml
+++ b/roles/openshift_node_upgrade/tasks/main.yml
@@ -31,6 +31,20 @@
   failed_when: false
   when: openshift.common.is_containerized | bool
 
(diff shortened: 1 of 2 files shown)


sdodson:
aos-ci-test

ingvagabund:
aos-ci-test

sdodson:
[merge]

PR #4179, opened 2017-05-12
singletonize some role tasks that repeat a lot

sosiouxme:
AIUI there are things in openshift_facts and other roles that actually do need to run repeatedly. However it seems like at least some of the tasks that we see multiple times could be skipped after the first time. This is an attempt to do that for some of them, to reduce iterative development time if nothing else.

PR #4667, opened 2017-07-03
singletonize some role tasks that repeat a lot

sosiouxme:
Backport to 1.5 from https://github.com/openshift/openshift-ansible/pull/4179


sosiouxme:
aos-ci-test

PR #4669, opened 2017-07-03
Adding support for storageclass in lib_openshift.

kwoodson:


roles/lib_openshift/src/class/oc_storageclass.py
```diff
@@ -0,0 +1,147 @@
...
+        return self._delete(self.kind, self.config.name)
+
+    def create(self):
+        '''create the object'''
+        return self._create_from_content(self.config.name, self.config.data)
+
+    def update(self):
+        '''update the object'''
+        # parameters are currently unable to be updated.  need to delete and recreate
+        self.delete()
```
ashcrow: Makes sense ... just a reminder there is a natural race condition here but there isn't anything we can really do about it.

ashcrow:
```
E       KeyError: 'default_storage_class'
```

kwoodson:
aos-ci-test

kwoodson:
aos-ci-test

ashcrow:
Holding off on merging per @kwoodson

kwoodson:
@sdodson, Please let me know what you'd like to do with this one.  We can hold off if we are too close to final cut.  Otoh it is pretty low risk and allows us to be a bit more flexible.  Thanks.

sdodson:
We'll wait until after code freeze.

roles/lib_openshift/src/class/oc_storageclass.py
```diff
@@ -0,0 +1,147 @@
+# pylint: skip-file
+# flake8: noqa
+
+# pylint: disable=too-many-instance-attributes
+class OCStorageClass(OpenShiftCLI):
+    ''' Class to wrap the oc command line tools '''
+    kind = 'sc'
```
ivanhorvath: 3.5 does not have this short form, had to change it to storageclass so the oc commands work

roles/lib_openshift/src/doc/storageclass
```diff
@@ -0,0 +1,86 @@
...
+- name: create the storageclass
+  oc_storageclass:
+  run_once: true
+    name: gp2
+    parameters:
+      type: gp2
+      encrypted: 'true'
+      kmsKeyId: '<full kms key arn>'
+    provisioner: aws-ebs
+    default_sc: False
```
ivanhorvath: This should be 
default_storage_class: False to avoid confusion

roles/lib_openshift/src/class/oc_storageclass.py
```diff
@@ -0,0 +1,147 @@
...
+    def update(self):
+        '''update the object'''
+        # parameters are currently unable to be updated.  need to delete and recreate
+        self.delete()
+        return self.create()
+
+    def needs_update(self):
+        ''' verify an update is needed '''
+        # check if params have updated
+        if self.storage_class.get_parameters() == self.config.parameters:
```
ivanhorvath: May need to check for annotations too if default_storage_class is changing between true and false.

roles/lib_openshift/src/lib/storageclass.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def create_dict(self):
+        ''' instantiates a storageclass dict '''
+        self.data['apiVersion'] = self.api_version
+        self.data['kind'] = 'StorageClass'
+        self.data['metadata'] = {}
+        self.data['metadata']['name'] = self.name
+
+        self.data['metadata']['annotations'] = {}
+        self.data['metadata']['annotations']['storageclass.beta.kubernetes.io/is-default-class'] = self.default_sc
```
ivanhorvath: Python nameconventions caused this to show up in the storageclass definition as "True", so had to do
str(self.default_sc).lower() to make it work. Is there an easier way to force python to handle bool to str conversions?

roles/lib_openshift/src/class/oc_storageclass.py
kwoodson: Thanks!

updated.

roles/lib_openshift/src/doc/storageclass
kwoodson: Thanks, fixed.

roles/lib_openshift/src/class/oc_storageclass.py
kwoodson: added check for annotations.

roles/lib_openshift/src/lib/storageclass.py
kwoodson: @ivanhorvath, the ansible type for the module parameter is set to string.  Did you set specifically `'false'`?  (notice the quotes)

I can cast it for safety.

kwoodson:
aos-ci-test

kwoodson:
@sdodson, can we merge this now?

sdodson:
[merge][severity:blocker]

PR #4670, opened 2017-07-04
Configure persistence volumes for hawkular-metrics and Cassandra logs.

rubenvp8510:


mwringe:
This needs to be a configurable option.

Do we even want to persist our logs like this? Or should that be up to the OpenShift Logging integration?

mwringe:
It's not as simple as just adding in a PV here, we would need to also use a statefulset for this, which means reconfiguring how our components are deployed.

I also don't see anything in here with regards to what happens if the PV gets full. Does Wildfly automatically take care of this? or are we going to run into problems where the logs get full and we start to lose information?

The other problem is that our PVC is set to recycle, which means if someone deletes the pod, the logs are lost.

rubenvp8510:
@mwringe I'm aware of the statefulset thing, we discussed this on IRC,  regards of the PV gets full, I configured wildly for log rotation.

About the recycle of the PVC , good point, I'll update this PR with new changes.

rubenvp8510:
Just one question, the statefulset are official supported in OpenShift?

mwringe:
We can use statefulsets in OpenShift now

PR #4672, opened 2017-07-04
[1.5] Changed annotation for dynamic provisioning on metrics and logging

zonArt:
provisioning with storageclass is not working when using the alpha version

### DIFF
--- a/roles/openshift_logging/tasks/generate_pvcs.yaml
+++ b/roles/openshift_logging/tasks/generate_pvcs.yaml
@@ -32,7 +32,7 @@
   vars:
     obj_name: "{{claim_name}}"
     annotations:
-      volume.alpha.kubernetes.io/storage-class: "dynamic"
+      volume.beta.kubernetes.io/storage-class: "dynamic"
     size: "{{es_pvc_size}}"
     access_modes: "{{ es_access_modes | list }}"
     pv_selector: "{{es_pv_selector}}"
--- a/roles/openshift_logging/tasks/set_es_storage.yaml
+++ b/roles/openshift_logging/tasks/set_es_storage.yaml
@@ -49,7 +49,7 @@
     vars:
       obj_name: "{{ es_pvc_prefix }}-{{ es_pvc_names_count | int + es_pvc_count | int }}"
       annotations:
-        volume.alpha.kubernetes.io/storage-class: "dynamic"
+        volume.beta.kubernetes.io/storage-class: "dynamic"
       size: "{{ es_pvc_size }}"
       access_modes: "{{ openshift_logging_storage_access_modes }}"
       pv_selector: "{{ es_pv_selector }}"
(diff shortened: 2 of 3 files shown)


zonArt:
aos-ci-jenkins

zonArt:
bot, test pull request

sdodson:
@gnufied @kwoodson PTAL

gnufied:
@zonArt there is an actual storageClass called `dynamic` created before being used here right? I seem to remember - we do have that class but just wanted to double check..

sdodson:
aos-ci-test

zonArt:
@gnufied Actually this is my second concern on the playbook, I'm not sure the storageclass is created. What I personally did was run the config.yml playbook without metrics and logging deployment, once cluster is up, create the storageclass and run the playbook again with metrics/logging deployment.
But I also may have missed something and in this case I would love to know how to set the storageclass (especially the provider part), I'm using ceph

gnufied:
@sdodson  please don't merge this yet, I take back my earlier "approval".  I am looking bit closely.

gnufied:
@zonArt just to confirm by ceph you mean `ceph-rbd` right?

zonArt:
@gnufied yes `ceph-rbd`

gnufied:
I do not think we create any ceph-rbd based storageclasses as part of openshift-ansible steps. We should leave the PVC definition as it is and not change it.

zonArt:
@gnufied I think you're right this storageclass is not created, however as it is moving to beta, I think it's a good idea to correct the annotations in the pvc to reflect the change. This is intended for the 1.5 version of origin no sense to have that on master. But I think you already noticed that.

gnufied:
@zonArt yeah the name of storageclasses has already been changed on master in preparation for 3.6 release. We are also creating default storageclasses in master branch but only for GCE & AWS envs.  @kwoodson is working on another refactor to make default storageclasses more configurable.

Lets close this PR for now.

kwoodson:
@gnufied, @zonArt,  Is there any harm switching from alpha to beta?  If the current behavior doesn't work with the `dynamic` storageclass then it is `no harm no foul`, right?

I'm fine with bumping the version up from `alpha` to `beta` if it is better off in the long run.

If the current `alpha` is working and switching to `beta` breaks something then I would agree, let's not do it.  Please help me understand what I'm missing.

gnufied:
switching from alpha to beta could actually break existing scripts, because alpha annotation is special and it works even if there is no actual storage class present. So,  if a PVC has an alpha annotation for storageclass but the actual storageclass doesn't exist - provisioner tries to make best guess and provisions a volume using cloudprovider information. 

If we switch to beta - it means that, the specified storageclass must exist, otherwise it won't work.

kwoodson:
@gnufied, thanks for clarifying.  Let's hold off on this then.

michaelgugino:
Closing due to inactivity.

PR #4674, opened 2017-07-04
drop etcdctl before the etcd_container service

ingvagabund:
If the etcdctl is dropped after the etcd_container is enabled,
label of /var/lib/etcd directory is set to var_lib_t instead of virt_sandbox_file_t.

Bug: 1460959


ingvagabund:
aos-ci-test

sdodson:
[merge]

PR #4675, opened 2017-07-04
openshift_storage_nfs_lvm: parametrize exported directories mode

jfchevrette:
Depending on the environment and security requirements, a user may want to set a different mode on the nfs-exported directories.

This PR moves the default mode (0700) to a default role variable to allow a user to change it from a playbook.

### DIFF
--- a/roles/openshift_storage_nfs_lvm/defaults/main.yml
+++ b/roles/openshift_storage_nfs_lvm/defaults/main.yml
@@ -6,6 +6,10 @@ osnl_nfs_export_options: "*(rw,sync,all_squash)"
 # mounted as <osnl_mount_dir>/test1g0001 etc.
 osnl_mount_dir: /exports/openshift
 
+# Mode for the NFS-exported directories. Defaults to secure mode which may require additional configs depending on the environment.
+# See https://docs.openshift.com/container-platform/3.5/install_config/persistent_storage/persistent_storage_nfs.html#nfs-volume-security
+r_openshift_storage_nfs_lvm_export_dir_mode: 0700
+
 # Volume Group to use.
 osnl_volume_group: openshiftvg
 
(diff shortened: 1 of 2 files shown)


roles/openshift_storage_nfs_lvm/defaults/main.yml
```diff
@@ -6,6 +6,10 @@ osnl_nfs_export_options: "*(rw,sync,all_squash)"
 # mounted as <osnl_mount_dir>/test1g0001 etc.
 osnl_mount_dir: /exports/openshift
 
+# Mode for the NFS-exported directories. Defaults to secure mode which may require additional configs depending on the environment.
+# See https://docs.openshift.com/container-platform/3.5/install_config/persistent_storage/persistent_storage_nfs.html#nfs-volume-security
+osnl_exported_dir_mode: 0700
```
sdodson: I know it's quite verbose, but our new role variable standards say that this should be `r_openshift_storage_nfs_lvm_export_dir_mode` or something along those lines, it just needs to be r_$rolename_foo

jfchevrette: Makes sense. I updated the commit with the new variable naming convention. Is this documented somewhere?
Happy to look into refactoring the whole role for consistency. However this would break users configs if they are using the current/old variable names. I'm thinking about updating the variables names and use set_fact to set them based on the old variables values if those are found/defined. WDYT?

sdodson: The way we've generally handled deprecations like that in the past is like this in the defaults/main.yml
`r_openshift_storage_nfs_lvm_mount_dir: "{{ osnl_mount_dir | default('/exports/openshift') }}"`

jfchevrette:
Pushed PR #4683 which include this change and refactors the role to use new variable naming convention, sets default values for variables which didn't have one and a couple more things. Closing this one in favor of the new PR

PR #4680, opened 2017-07-05
set the proper label of /var/lib/etcd directory

ingvagabund:
Cause there is no guarantee the /var/lib/etcd will keep its `svirt_sandbox_file_t` label between installation and upgrade.

The /var/lib/etcd will have incorrect `var_lib_t` label until the `etcd_container` service is restarted. Once restarted, the directory is correctly re-labeled.

### DIFF
--- a/roles/etcd_common/tasks/backup.yml
+++ b/roles/etcd_common/tasks/backup.yml
@@ -61,6 +61,21 @@
   - r_etcd_common_embedded_etcd | bool
   - not l_ostree_booted.stat.exists | bool
 
+- name: Check selinux label of '{{ l_etcd_data_dir }}'
+  command: >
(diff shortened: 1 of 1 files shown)


ingvagabund:
`sefcontext` module [1] does not seem to work.

[1] https://docs.ansible.com/ansible/sefcontext_module.html

ingvagabund:
aos-ci-test

ashcrow:
Confirmed, the ``sefcontext`` module doesn't seem to work. However, the ``file`` module can set context:

```yaml
  - name: Make sure the /var/lib/etcd has the proper label
    file:
      path: /var/lib/etcd'
      recurse: true
      setype: svirt_sandbox_file_t
```

ashcrow:
[merge]

PR #4682, opened 2017-07-05
verify sane log times in logging stack

juanvallejo:
Related trello card: https://trello.com/c/7uOluOTW
Based on @richm 's script: https://bugzilla.redhat.com/attachment.cgi?id=1258225

This patch verifies that logs sent from logging pods can be queried on
the Elasticsearch pod within a reasonable amount of time.

*TODO*

- [x] add tests

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/logging/logging.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/logging.py
@@ -12,20 +12,21 @@ class LoggingCheck(OpenShiftCheck):
     """Base class for logging component checks"""
 
     name = "logging"
+    logging_namespace = "logging"
 
     @classmethod
     def is_active(cls, task_vars):
-        return super(LoggingCheck, cls).is_active(task_vars) and cls.is_first_master(task_vars)
+        logging_deployed = get_var(task_vars, "openshift_hosted_logging_deploy", default=False)
+        return super(LoggingCheck, cls).is_active(task_vars) and cls.is_first_master(task_vars) and logging_deployed
 
     @staticmethod
     def is_first_master(task_vars):
-        """Run only on first master and only when logging is configured. Returns: bool"""
-        logging_deployed = get_var(task_vars, "openshift_hosted_logging_deploy", default=True)
+        """Run only on first master. Returns: bool"""
         # Note: It would be nice to use membership in oo_first_master group, however for now it
         # seems best to avoid requiring that setup and just check this is the first master.
         hostname = get_var(task_vars, "ansible_ssh_host") or [None]
         masters = get_var(task_vars, "groups", "masters", default=None) or [None]
-        return logging_deployed and masters[0] == hostname
+        return masters and masters[0] == hostname
 
     def run(self, tmp, task_vars):
         pass
@@ -45,7 +46,7 @@ class LoggingCheck(OpenShiftCheck):
                 raise ValueError()
         except ValueError:
             # successful run but non-parsing data generally means there were no pods in the namespace
-            return None, 'There are no pods in the {} namespace. Is logging deployed?'.format(namespace)
+            return None, 'No pods were found for the "{}" logging component.'.format(logging_component)
 
         return pods['items'], None
 
--- a/roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
@@ -1,0 +1,132 @@
+"""
+Check for ensuring logs from pods can be queried in a reasonable amount of time.
+"""
+
+import json
+import time
+
+from uuid import uuid4
+
+from openshift_checks import get_var, OpenShiftCheckException
+from openshift_checks.logging.logging import LoggingCheck
+
+
+ES_CMD_TIMEOUT_SECONDS = 30
+
+
+class LoggingIndexTime(LoggingCheck):
+    """Check that pod logs are aggregated and indexed in ElasticSearch within a reasonable amount of time."""
+    name = "logging_index_time"
+    tags = ["health", "logging"]
+
+    logging_namespace = "logging"
+
+    def run(self, tmp, task_vars):
+        """Add log entry by making unique request to Kibana. Check for unique entry in the ElasticSearch pod logs."""
+        try:
+            log_index_timeout = int(
+                get_var(task_vars, "openshift_check_logging_index_timeout_seconds", default=ES_CMD_TIMEOUT_SECONDS)
+            )
+        except ValueError:
+            return {
+                "failed": True,
+                "msg": ('Invalid value provided for "openshift_check_logging_index_timeout_seconds". '
+                        'Value must be an integer representing an amount in seconds.'),
+            }
+
+        running_component_pods = dict()
+
+        # get all component pods
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default=self.logging_namespace)
+        for component, name in (['kibana', 'Kibana'], ['es', 'Elasticsearch']):
+            pods, error = self.get_pods_for_component(
+                self.execute_module, self.logging_namespace, component, task_vars,
+            )
+
+            if error:
+                msg = 'Unable to retrieve pods for the {} logging component: {}'
+                return {"failed": True, "changed": False, "msg": msg.format(name, error)}
+
+            running_pods = self.running_pods(pods)
+
+            if not running_pods:
+                msg = ('No {} pods in the "Running" state were found.'
+                       'At least one pod is required in order to perform this check.')
+                return {"failed": True, "changed": False, "msg": msg.format(name)}
+
+            running_component_pods[component] = running_pods
+
+        uuid = self.curl_kibana_with_uuid(running_component_pods["kibana"][0], task_vars)
+        self.wait_until_cmd_or_err(running_component_pods["es"][0], uuid, log_index_timeout, task_vars)
+        return {}
+
+    def wait_until_cmd_or_err(self, es_pod, uuid, timeout_secs, task_vars):
+        """Retry an Elasticsearch query every second until query success, or a defined
+        length of time has passed."""
+        deadline = time.time() + timeout_secs
+        interval = 1
+        while not self.query_es_from_es(es_pod, uuid, task_vars):
+            if time.time() + interval > deadline:
+                msg = "expecting match in Elasticsearch for message with uuid {}, but no matches were found after {}s."
+                raise OpenShiftCheckException(msg.format(uuid, timeout_secs))
+            time.sleep(interval)
+
+    def curl_kibana_with_uuid(self, kibana_pod, task_vars):
+        """curl Kibana with a unique uuid."""
+        uuid = self.generate_uuid()
+        pod_name = kibana_pod["metadata"]["name"]
+        exec_cmd = "exec {pod_name} -c kibana -- curl --max-time 30 -s http://localhost:5601/{uuid}"
+        exec_cmd = exec_cmd.format(pod_name=pod_name, uuid=uuid)
+
+        error_str = self.exec_oc(self.execute_module, self.logging_namespace, exec_cmd, [], task_vars)
+
+        try:
+            error_code = json.loads(error_str)["statusCode"]
+        except KeyError:
+            msg = ('invalid response returned from Kibana request (Missing "statusCode" key):\n'
+                   'Command: {}\nResponse: {}').format(exec_cmd, error_str)
+            raise OpenShiftCheckException(msg)
+        except ValueError:
+            msg = ('invalid response returned from Kibana request (Non-JSON output):\n'
+                   'Command: {}\nResponse: {}').format(exec_cmd, error_str)
+            raise OpenShiftCheckException(msg)
+
+        if error_code != 404:
+            msg = 'invalid error code returned from Kibana request. Expecting error code "404", but got "{}" instead.'
+            raise OpenShiftCheckException(msg.format(error_code))
+
+        return uuid
+
+    def query_es_from_es(self, es_pod, uuid, task_vars):
+        """curl the Elasticsearch pod and look for a unique uuid in its logs."""
+        pod_name = es_pod["metadata"]["name"]
+        exec_cmd = (
+            "exec {pod_name} -- curl --max-time 30 -s -f "
+            "--cacert /etc/elasticsearch/secret/admin-ca "
+            "--cert /etc/elasticsearch/secret/admin-cert "
+            "--key /etc/elasticsearch/secret/admin-key "
+            "https://logging-es:9200/project.{namespace}*/_count?q=message:{uuid}"
+        )
+        exec_cmd = exec_cmd.format(pod_name=pod_name, namespace=self.logging_namespace, uuid=uuid)
+        result = self.exec_oc(self.execute_module, self.logging_namespace, exec_cmd, [], task_vars)
+
+        try:
+            count = json.loads(result)["count"]
+        except KeyError:
+            msg = 'invalid response from Elasticsearch query:\n"{}"\nMissing "count" key:\n{}'
+            raise OpenShiftCheckException(msg.format(exec_cmd, result))
+        except ValueError:
+            msg = 'invalid response from Elasticsearch query:\n"{}"\nNon-JSON output:\n{}'
+            raise OpenShiftCheckException(msg.format(exec_cmd, result))
+
+        return count
+
+    @staticmethod
+    def running_pods(pods):
+        """Filter pods that are running."""
+        return [pod for pod in pods if pod['status']['phase'] == 'Running']
+
+    @staticmethod
+    def generate_uuid():
+        """Wrap uuid generator. Allows for testing with expected values."""
+        return str(uuid4())
--- a/roles/openshift_health_checker/test/logging_index_time_test.py
+++ b/roles/openshift_health_checker/test/logging_index_time_test.py
@@ -1,0 +1,182 @@
+import json
+
+import pytest
+
+from openshift_checks.logging.logging_index_time import LoggingIndexTime, OpenShiftCheckException
+
+
+SAMPLE_UUID = "unique-test-uuid"
+
+
+def canned_loggingindextime(exec_oc=None):
+    """Create a check object with a canned exec_oc method"""
+    check = LoggingIndexTime("dummy")  # fails if a module is actually invoked
+    if exec_oc:
+        check.exec_oc = exec_oc
+    return check
+
+
+plain_running_elasticsearch_pod = {
+    "metadata": {
+        "labels": {"component": "es", "deploymentconfig": "logging-es-data-master"},
+        "name": "logging-es-data-master-1",
+    },
+    "status": {
+        "containerStatuses": [{"ready": True}, {"ready": True}],
+        "phase": "Running",
+    }
+}
+plain_running_kibana_pod = {
+    "metadata": {
+        "labels": {"component": "kibana", "deploymentconfig": "logging-kibana"},
+        "name": "logging-kibana-1",
+    },
+    "status": {
+        "containerStatuses": [{"ready": True}, {"ready": True}],
+        "phase": "Running",
+    }
+}
+not_running_kibana_pod = {
+    "metadata": {
+        "labels": {"component": "kibana", "deploymentconfig": "logging-kibana"},
+        "name": "logging-kibana-2",
+    },
+    "status": {
+        "containerStatuses": [{"ready": True}, {"ready": False}],
+        "conditions": [{"status": "True", "type": "Ready"}],
+        "phase": "pending",
+    }
+}
+
+
+@pytest.mark.parametrize('pods, expect_pods', [
+    (
+        [not_running_kibana_pod],
+        [],
+    ),
+    (
+        [plain_running_kibana_pod],
+        [plain_running_kibana_pod],
+    ),
+    (
+        [],
+        [],
+    )
+])
+def test_check_running_pods(pods, expect_pods):
+    check = canned_loggingindextime(None)
+    pods = check.running_pods(pods)
+    assert pods == expect_pods
+
+
+@pytest.mark.parametrize('name, json_response, uuid, timeout, extra_words', [
+    (
+        'valid count in response',
+        {
+            "count": 1,
+        },
+        SAMPLE_UUID,
+        0.001,
+        [],
+    ),
+], ids=lambda argval: argval[0])
+def test_wait_until_cmd_or_err_succeeds(name, json_response, uuid, timeout, extra_words):
+    def exec_oc(execute_module, ns, exec_cmd, args, task_vars):
+        return json.dumps(json_response)
+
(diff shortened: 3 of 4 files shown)


juanvallejo:
@rhcarvalho thanks for the feedback. Review comments from https://github.com/openshift/openshift-ansible/pull/4519 addressed

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,168 @@
...
+
+        if error:
+            msg = 'Unable to retrieve pods for the "Elasticsearch" logging component: {}'
+            return {"failed": True, "changed": False, "msg": msg.format(error)}
+
+        running_es_pods = self.running_pods(pods)
+        if not len(running_es_pods):
+            msg = ('No Elasticsearch pods were found to be in the "Running" state.'
+                   'At least one Elasticsearch pod is required in order to perform this check.')
+            return {"failed": True, "changed": False, "msg": msg}
```
rhcarvalho: We can avoid duplication and ensure consistency:

```python
self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default=self.logging_namespace)

for component, name in (['kibana', 'Kibana'], ['es', 'Elasticsearch']):
    pods, error = self.get_pods_for_component(
        self.execute_module, self.logging_namespace, component, task_vars,
    )

    if error:
        msg = 'Unable to retrieve pods for the {} logging component: {}'
        return {"failed": True, "changed": False, "msg": msg.format(name, error)}

    running_pods = self.running_pods(pods)

    if not running_pods:
        msg = ('No {} pods in the "Running" state were found.'
               'At least one pod is required in order to perform this check.')
        return {"failed": True, "changed": False, "msg": msg.format(name)}
```

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,168 @@
...
+            "es",
+            task_vars,
+        )
+
+        if error:
+            msg = 'Unable to retrieve pods for the "Elasticsearch" logging component: {}'
+            return {"failed": True, "changed": False, "msg": msg.format(error)}
+
+        running_es_pods = self.running_pods(pods)
+        if not len(running_es_pods):
```
rhcarvalho: s/len(...)/...

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,168 @@
...
+            return {"failed": True, "changed": False, "msg": msg}
+
+        uuid = self.curl_kibana_with_uuid(running_kibana_pods[0], task_vars)
+        self.wait_until_cmd_or_err(running_es_pods[0], uuid, log_index_timeout, task_vars)
+        return {}
+
+    def wait_until_cmd_or_err(self, es_pod, uuid, timeout_secs, task_vars):
+        """Retry an Elasticsearch query every second until query success, or a defined
+        length of time has passed."""
+        now = time.time()
```
rhcarvalho: Original comment: https://github.com/openshift/openshift-ansible/pull/4519#discussion_r125238506

Let me try to explain again.

"Retry ... until a defined length of time has passed" sounds plausible, but is hard to implement without access to a monotonic time source. `time.time()` is not monotonic.

Instead of adding hacks to try to compensate for problems with `time.time()`, what I am proposing is an algorithm that doesn't depend on it: "retry N times with S seconds interval" (`time.sleep` uses a monotonic clock internally).

In code:

```python
interval = 1

for i in range(max_tries):
    if self.query_es_from_es(es_pod, uuid, task_vars):
        return
    if i < max_tries - 1:
        time.sleep(interval)  # note: we actually don't need to sleep after the last try...

msg = ...
raise ...
```

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,168 @@
...
+        return [pod for pod in pods if pod['status']['phase'] == 'Running']
+
+    @staticmethod
+    def generate_uuid():
+        """Wrap uuid generator. Allows for testing with expected values."""
+        return str(uuid4())
+
+    def oc_cmd(self, cmd_str, extra_args, task_vars):
+        """Wrap parent exec_oc method. Allows for testing without actually invoking other modules."""
+        return super(LoggingIndexTime, self).exec_oc(
```
rhcarvalho: Why `super(LoggingIndexTime, self).exec_oc`? Super is used when you want to access the parent's implementation skipping the current class', useful when overriding a method. Here it should probably be `self.exec_oc`.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,168 @@
...
+            return {
+                "failed": True,
+                "msg": ('Invalid value provided in your Ansible inventory file for '
+                        '"openshift_check_logging_index_timeout_seconds". Value must '
+                        'be an integer representing an amount in seconds.'),
+            }
+
+        # get all Kibana pods
+        self.logging_namespace = get_var(task_vars, "openshift_logging_namespace", default=self.logging_namespace)
+        pods, error = super(LoggingIndexTime, self).get_pods_for_component(
```
rhcarvalho: s/super(LoggingIndexTime, self)/self ?

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,168 @@
...
+            log_index_timeout = int(
+                get_var(task_vars, "openshift_check_logging_index_timeout_seconds", default=ES_CMD_TIMEOUT_SECONDS)
+            )
+        except ValueError:
+            return {
+                "failed": True,
+                "msg": ('Invalid value provided in your Ansible inventory file for '
+                        '"openshift_check_logging_index_timeout_seconds". Value must '
+                        'be an integer representing an amount in seconds.'),
+            }
```
juanvallejo: I think that it might be worth the price of the added annoyance by having the check fail. That way, a user is at least aware of why a value supplied for `openshift_check_logging_index_timeout_seconds` was invalid

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
juanvallejo: Although, looking at this a bit more, wouldn't it be better to re-use `get_pods_for_component` from the parent class (`LoggingCheck`)?

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
juanvallejo: Thanks, will update

juanvallejo:
@rhcarvalho thanks for the feedback, review comments addressed

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
sosiouxme: I agree if they went to the trouble to set this, they should be alerted if it isn't valid.

sosiouxme: I agree if they went to the trouble to set this, they should be alerted if it isn't valid. Until we get some kind of "warning" mechanism anyway.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
sosiouxme: @rhcarvalho What I don't like about that is that it doesn't account for the time spent on the query, which involves execing into the pod (could take an unbounded amount of time) and running curl (could take up to the limit we set). Frequently the sleep(1) would be the briefest part of the loop. So you could easily have 30 times through the loop varying from ~30s to ~1000s, especially under adverse conditions (exactly the kind you are looking for in the first place). And that's the tradeoff for avoiding a large-clock-reset edge case that should be very rare. I'd rather detect the clock reset and restart the timer in that rare event than warp the accuracy of the check continuously.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
rhcarvalho: When an instance `inst` of a class `C` doesn't implement / override a given method `m`, calling `inst.m()` is the same as `super(C, self).m()`. However, as soon as you override the method, the two calls produce a different result.

Semantically, we mean `inst.m()` and the Python runtime will lookup what is the implementation of `m` to run. The `super(C, self).m()` for is useful when you need to bypass the lookup, normally in a case where you would introduce an infinite recursion looking up for a method implementation. E.g., this doesn't work:

```python
>>> class Foo:
...     def m(self):
...         return self.m()
... 
>>> Foo().m()
  File "<stdin>", line 3, in m
  File "<stdin>", line 3, in m
  ...
RecursionError: maximum recursion depth exceeded
```

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
rhcarvalho: Well, one thing we need to work on is how users discover they can set a variable `openshift_check_logging_index_timeout_seconds`, and then that this value must be an integer (and not a float). This is a general problem for all checks, their usage and their customization -- unrelated to this PR.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,151 @@
...
+
+            running_component_pods[component] = running_pods
+
+        uuid = self.curl_kibana_with_uuid(running_component_pods["kibana"][0], task_vars)
+        self.wait_until_cmd_or_err(running_component_pods["es"][0], uuid, log_index_timeout, task_vars)
+        return {}
+
+    def wait_until_cmd_or_err(self, es_pod, uuid, timeout_secs, task_vars):
+        """Retry an Elasticsearch query every second until query success, or a defined
+        length of time has passed."""
```
rhcarvalho: Hmm, now this is out of sync with the implementation?

Re-reading the previous conversation in https://github.com/openshift/openshift-ansible/pull/4682#discussion_r126519475 and reconsidering, I get that the intention of this check is look at the _elapsed time_ between writing something to Kibana and being able to read it from Elasticsearch.

Sorry for the back-and-forth, but I'll go with what @sosiouxme said and what you had originally, that we want the whole check to run within a certain timeout, and we try to query ES as many times as possible within that time, always spaced by some interval.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,151 @@
...
+    def running_pods(pods):
+        """Filter pods that are running."""
+        return [pod for pod in pods if pod['status']['phase'] == 'Running']
+
+    @staticmethod
+    def generate_uuid():
+        """Wrap uuid generator. Allows for testing with expected values."""
+        return str(uuid4())
+
+    def oc_cmd(self, cmd_str, extra_args, task_vars):
```
rhcarvalho: Perhaps this belongs in the parent class? It seems to have nothing specific about the logging index time check.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
sosiouxme: One of the things I intend to do is introduce a kwarg on get_var so you can specify what type a var should have. Then it will either come back as that type or as an exception thrown that hopefully informs the user what is expected (in a unified fashion, not per check).

That doesn't help them discover the value in the first place... it would also be good if each check had a spec for which values are configurable and the expected semantics, which we could show in either a check "help" playbook or when the checks fail (as an addendum to the "how to disable" advice). In fact maybe get_var could look at that too :)

juanvallejo:
@rhcarvalho @sosiouxme thanks for the feedback, comments addressed

juanvallejo:
@rhcarvalho @sosiouxme Will go ahead and squash if no further comments on this. Thanks

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,141 @@
...
+        now = time.time()
+        time_start = now
+        time_end = time_start + timeout_secs
+        interval = 1  # seconds to wait between retries
+
+        while now < time_end:
+            if self.query_es_from_es(es_pod, uuid, task_vars):
+                return
+            time.sleep(interval)
+            now = int(time.time())
```
rhcarvalho: Why `int`? I think we don't need to truncate the time.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,141 @@
...
+
+        uuid = self.curl_kibana_with_uuid(running_component_pods["kibana"][0], task_vars)
+        self.wait_until_cmd_or_err(running_component_pods["es"][0], uuid, log_index_timeout, task_vars)
+        return {}
+
+    def wait_until_cmd_or_err(self, es_pod, uuid, timeout_secs, task_vars):
+        """Retry an Elasticsearch query every second until query success, or a defined
+        length of time has passed."""
+        now = time.time()
+        time_start = now
```
rhcarvalho: Seems that this is not used for anything, dead code.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,141 @@
...
+        time_end = time_start + timeout_secs
+        interval = 1  # seconds to wait between retries
+
+        while now < time_end:
+            if self.query_es_from_es(es_pod, uuid, task_vars):
+                return
+            time.sleep(interval)
```
rhcarvalho: We can save 1s in run time by not sleeping in the last loop iteration.

Suggestion:

```python
def wait_until_doc_in_es(self, es_pod, uuid, timeout_secs, task_vars):
    deadline = time.time() + timeout_secs
    interval = 1
    while not self.query_es_from_es(es_pod, uuid, task_vars):
        if time.time() + interval > deadline:
            msg = "Message with UUID={} not found in Elasticsearch after {:.1f}s."
            raise OpenShiftCheckException(msg.format(uuid, timeout_secs))
        time.sleep(interval)
```

The above ignores clock resets, but guarantees that we query Elasticsearch at least once.
It also exits immediately when the deadline expired or the document was found, without the extra sleep in the last loop iteration.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
sosiouxme: It's used later for detecting if the clock reset far back (before the start time).

If the clock only resets a little, nothing special happens. Same if it resets ahead... the loop will not have as long to run.

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
sosiouxme: It would only sleep on the last loop iteration if the query were not passing.

Err... so do you want to ignore clock resets now? It certainly simplifies the code... or was this just to insure against time resets forward?

rhcarvalho: > Err... so do you want to ignore clock resets now? It certainly simplifies the code...

It seems better to have it simple and to guarantee it runs at least once. Trying to implement monotonic time out `time.time()` is not going to work, plus adds bloat to confuse us later.

juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -20,7 +21,7 @@ def is_active(cls, task_vars):
     @staticmethod
     def is_first_master(task_vars):
```
rhcarvalho: This is a bad name, makes the `is_active` expression look like we only care about being first master, when in fact it is a bit more than that.

roles/openshift_health_checker/test/logging_index_time_test.py
```diff
@@ -0,0 +1,184 @@
...
+        return json.dumps(json_response)
+
+    check = canned_loggingindextime(exec_oc)
+    check.generate_uuid = lambda: uuid
+
+    with pytest.raises(OpenShiftCheckException) as error:
+        check.curl_kibana_with_uuid(plain_running_kibana_pod, None)
+
+    for word in extra_words:
+            assert word in str(error)
```
rhcarvalho: Too much indentation

roles/openshift_health_checker/test/logging_index_time_test.py
```diff
@@ -0,0 +1,184 @@
...
+        return json.dumps(json_response)
+
+    check = canned_loggingindextime(exec_oc)
+
+    error = ""
+    if expect_exception:
+        with pytest.raises(expect_exception) as error:
+            check.wait_until_cmd_or_err(plain_running_elasticsearch_pod, uuid, timeout, None)
+    else:
+        check.wait_until_cmd_or_err(plain_running_elasticsearch_pod, uuid, timeout, None)
```
rhcarvalho: Please split into two tests and avoid conditional assertions.
The parameters above mix positive and negative cases, making it harder to understand what is being tested.

rhcarvalho:
Also please squash the commits before we run aos-ci-test.

roles/openshift_health_checker/openshift_checks/logging/logging.py
juanvallejo: Thanks, updated method to only check if first master. Moved `logging_deployed` check to `is_active`

juanvallejo:

roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -12,20 +12,21 @@ class LoggingCheck(OpenShiftCheck):
...
-        """Run only on first master and only when logging is configured. Returns: bool"""
-        logging_deployed = get_var(task_vars, "openshift_hosted_logging_deploy", default=True)
+        """Run only on first master. Returns: bool"""
         # Note: It would be nice to use membership in oo_first_master group, however for now it
         # seems best to avoid requiring that setup and just check this is the first master.
         hostname = get_var(task_vars, "ansible_ssh_host") or [None]
         masters = get_var(task_vars, "groups", "masters", default=None) or [None]
-        return logging_deployed and masters[0] == hostname
+        return masters[0] == hostname
```
rhcarvalho: We may assume there is always at least one master. For robustness and to avoid an `IndexError`, we could do 
```python
return masters and masters[0] == hostname
```

roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -94,3 +95,13 @@ def exec_oc(execute_module=None, namespace="logging", cmd_str="", extra_args=Non
             raise OpenShiftCheckException(msg)
 
         return result.get("result", "")
+
+    def oc_cmd(self, cmd_str, extra_args, task_vars):
+        """Wrap parent exec_oc method. Allows for testing without actually invoking other modules."""
```
rhcarvalho: This docstring is out-of-sync. `exec_oc` is not a parent method, it is defined right here in `LoggingCheck`, and not in the parent class, `OpenShiftCheck`.

Tip: document what it does and why someone would want to call this method and what to expect from the return value. No need to make references to class hierarchy nor to "testing without actually invoking other modules" (typically that is not interesting, you can always monkey-patch things in Python... one could monkey-patch `exec_oc` itself and this method wouldn't even be needed. It seems that its existence is solely to let you pass in 2 arguments less.

Suggestion:

```python
    def oc_cmd(self, cmd_str, extra_args, task_vars):
        """Same as exec_oc with a predefined execute_module and namespace."""
```

I almost wrote that this simplifies calls to `exec_oc`, but then I noted that `exec_oc` has all optional arguments while `oc_cmd` has 3 required args, so maybe it is not all that more convenient to call?! All calls I saw pass and empty `[]` to `extra_args`.

juanvallejo: Thanks, will just remove `oc_cmd`

roles/openshift_health_checker/openshift_checks/logging/logging_index_time.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def query_es_from_es(self, es_pod, uuid, task_vars):
+        """curl the Elasticsearch pod and look for a unique uuid in its logs."""
+        pod_name = es_pod["metadata"]["name"]
+        exec_cmd = (
+            "exec {pod_name} -- curl --max-time 30 -s -f "
+            "--cacert /etc/elasticsearch/secret/admin-ca "
+            "--cert /etc/elasticsearch/secret/admin-cert "
+            "--key /etc/elasticsearch/secret/admin-key "
+            "https://logging-es:9200/project.{namespace}*/_count?q=message:{uuid}"
```
sosiouxme: Technically it should probably retrieve the project's UUID and be more specific about the ES index it's querying - reduce query time and load on ES. However I think it's a small enough concern that we can save this for when it becomes a problem or we're revisiting the code for some reason.

sosiouxme:
@rhcarvalho do you approve the changes?

rhcarvalho:
aos-ci-test

rhcarvalho:
[merge]

rhcarvalho:
Flake https://github.com/openshift/origin/issues/15356

And I don't understand https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible_extended_conformance_install_update/649/, says `Stage VERIFY THE INSTALLED OPENSHIFT-ANSIBLE RELEASE [00h 00m 01s] failed.` -- Looks like it comes from https://github.com/openshift/aos-cd-jobs/blob/master/sjb/config/test_cases/test_branch_origin_extended_conformance_install_update.yml#L136-L143, and according to the console output:

```
+ cd /data/src/github.com/openshift/openshift-ansible
++ git describe --tags --abbrev=0 --exact-match HEAD
+ last_tag=openshift-ansible-3.6.154-1
++ git log -n 1 --pretty=%h
+ last_commit=4513926
+ rpm -V openshift-ansible-3.6.154-1.git.0.4513926.el7
package openshift-ansible-3.6.154-1.git.0.4513926.el7 is not installed
++ export status=FAILURE
++ status=FAILURE
+ set +o xtrace
########## FINISHED STAGE: FAILURE: VERIFY THE INSTALLED OPENSHIFT-ANSIBLE RELEASE [00h 00m 01s] ##########
```


[merge]

rhcarvalho:
Again flake https://github.com/openshift/origin/issues/15356 & `Stage VERIFY THE INSTALLED OPENSHIFT-ANSIBLE RELEASE [00h 00m 01s] failed.`

Merging manually as per https://github.com/openshift/openshift-ansible/blob/master/docs/pull_requests.md#manual-merges.

rhcarvalho:
For the record, the other failure seems to be a CI script error: https://github.com/openshift/origin/issues/15186

sdodson:
thanks for tracking that down, sorry didn't get to respond to you on irc.

PR #4683, opened 2017-07-05
Refactor openshift_storage_nfs_lvm role

jfchevrette:
This is my first semi-large PR to this project. I tried to follow the contribution guidelines as much as possible. I was unable to split this into multiple commits since there are multiple intertwined changes. 

- Update all variables to match new role variable naming convention
- Add new default variables for variables which didn't have one (vol prefix, size, num_start and count)
- Add new variable _r_openshift_storage_nfs_lvm_export_dir_mode_ (same default as previously hardcoded value) to allow setting the mode of the nfs-exported dirs
- Update role README (/w variables table for better readability)
- Include fix from PR #4636 as it was not possible to test the changes without it
- Tested multiple scenarios, with new and old variable names & new defaults
- Tested on a live cluster with existing configs, no changes were applied before/after the role refactor

**Old var prefix**: osnl_
**New var prefix**: r_openshift_storage_nfs_lvm_

**Added vars:**
r_openshift_storage_nfs_lvm_export_dir_mode

**Var name change:**
<var_prefix>_number_of_volumes to <var_prefix>_volume_count

One thing I did not do but is probably worth doing at some point is to change all the occurrences of the old variables in the rest of the playbooks and roles. I can attempt to do this in a separate PR but at the moment old variables are still working and used if found.

Supercedes PR #4636 and #4675

### DIFF
--- a/roles/openshift_storage_nfs_lvm/README.md
+++ b/roles/openshift_storage_nfs_lvm/README.md
@@ -15,47 +15,17 @@ create persistent volumes.
 
 ## Role Variables
 
-```
-# Options of NFS exports.
-osnl_nfs_export_options: "*(rw,sync,all_squash)"
-
-# Directory, where the created partitions should be mounted. They will be
-# mounted as <osnl_mount_dir>/<lvm volume name> 
-osnl_mount_dir: /exports/openshift
-
-# Volume Group to use.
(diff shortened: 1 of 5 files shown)


michaelgugino:
This patch seems to have been abandoned.

PR #4684, opened 2017-07-05
Addressing servicecatalog doesnt have enough permissions and multimas…

ewolinetz:
…ter config for service-catalog

Should address:
 https://bugzilla.redhat.com/show_bug.cgi?id=1467860
 https://bugzilla.redhat.com/show_bug.cgi?id=1467602

### DIFF
--- a/roles/openshift_service_catalog/tasks/install.yml
+++ b/roles/openshift_service_catalog/tasks/install.yml
@@ -6,7 +6,6 @@
   register: mktemp
   changed_when: False
 
-
 - include: wire_aggregator.yml
 
 - name: Set default image variables based on deployment_type
--- a/roles/openshift_service_catalog/tasks/wire_aggregator.yml
+++ b/roles/openshift_service_catalog/tasks/wire_aggregator.yml
@@ -1,16 +1,82 @@
 ---
+- name: Make temp cert dir
+  command: mktemp -d /tmp/openshift-service-catalog-ansible-XXXXXX
+  register: certtemp
+  changed_when: False
+
+- name: Check for First Master Aggregator Signer cert
+  stat:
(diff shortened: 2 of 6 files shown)


ewolinetz:
Should also address https://bugzilla.redhat.com/show_bug.cgi?id=1467798

ewolinetz:
@sdodson I'm playing around with a way to sync the proxy certs and aggregator-front-proxy certs from the first master to other masters

ewolinetz:
@sdodson updated how we do multimaster configs, please rereview

roles/openshift_service_catalog/tasks/install.yml
```diff
@@ -6,7 +6,6 @@
   register: mktemp
   changed_when: False
 
-
 - include: wire_aggregator.yml
```
ewolinetz: should this be removed or is it alright to keep in as a just in case?

roles/openshift_service_catalog/tasks/wire_aggregator.yml
```diff
@@ -1,16 +1,82 @@
...
+  command: mktemp -d /tmp/openshift-service-catalog-ansible-XXXXXX
+  register: certtemp
+  changed_when: False
+
+- name: Check for First Master Aggregator Signer cert
+  stat:
+    path: /etc/origin/master/front-proxy-ca.crt
+  register: first_proxy_ca_crt
+  changed_when: false
+  delegate_to: "{{ first_master }}"
```
sdodson: may need `run_once: true` here otherwise all three hosts will delegate the command to master in parallel and potentially stomp on each other.

sdodson:
going ahead and merging so this makes it into the next build, we can follow up tomorrow on the item i noted.

PR #3821, opened 2017-03-31
Allowing a wildcard cert to be generated.

kwoodson:

PR #4693, opened 2017-07-06
Router wildcard certificate created by default

enoodle:
I suggest setting this parameter default to `True`. Since https://github.com/openshift/openshift-ansible/pull/3821 It seems more useful then using arbitrary hostnames.

cc @kwoodson @simon3z

### DIFF
--- a/roles/openshift_hosted/defaults/main.yml
+++ b/roles/openshift_hosted/defaults/main.yml
@@ -29,7 +29,7 @@ openshift_hosted_routers:
 
 openshift_hosted_router_certificate: {}
 openshift_hosted_registry_cert_expire_days: 730
-openshift_hosted_router_create_certificate: False
+openshift_hosted_router_create_certificate: True
 
 os_firewall_allow:
 - service: Docker Registry Port
--- a/roles/openshift_hosted/tasks/router/router.yml
+++ b/roles/openshift_hosted/tasks/router/router.yml
@@ -37,7 +37,7 @@
         cafile: "{{ openshift_master_config_dir ~ '/ca.crt' }}"
 
   # End Block
-  when: openshift_hosted_router_create_certificate | bool
+  when: ( openshift_hosted_router_create_certificate | bool ) and openshift_hosted_router_certificate == {}
 
 - name: Get the certificate contents for router
   copy:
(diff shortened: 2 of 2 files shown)


enoodle:
ping @kwoodson @jcantrill PTAL

kwoodson:
@mtnbikenc @abutcher,  do you see any reason as to why we should or shouldn't do this?

jcantrill:
@enoodle I don't have an opinion on why we would or would not do this.

abutcher:
@kwoodson @enoodle My only worry would be that defaulting `openshift_hosted_router_create_certificate=true` would stomp a user provided `openshift_hosted_router_certificate`.

kwoodson:
@abutcher, @enoodle, That's my fear as we don't use the default router certs on our operations team.  We supply real certificates and would get default ones which would break our installs unless specifying `openshift_hosted_router_create_certificate=false`. 

I understand the desire to have a default be turned on since a lot of installs do not require real certificates.  I guess this is just a matter of preference unless someone has a good argument to update it.

abutcher:
We could probably default this if that linked section didn't create the default wildcard certificate when `openshift_hosted_router_certificate` was provided.

enoodle:
@abutcher I have added a condition on `openshift_hosted_router_certificate` being empty, PTAL

enoodle:
@abutcher @kwoodson   I have added a condition on openshift_hosted_router_certificate being empty, Can you take another look?

abutcher:
@enoodle The conditional looks great thanks.

I did not notice that the wildcard cert will contain SANs for `*.openshift_master_default_subdomain` and `openshift_master_default_subdomain` without any defaults, so base install is `""`, `"*."`. If this is always on I think we should have a default for this like `*.router.default.svc.cluster.local`. @sdodson @kwoodson sound okay?

kwoodson:
@enoodle, sorry for the slow response.  We've been locked down working on 3.6.

@abutcher, That sounds fine.

enoodle:
@kwoodson @abutcher I have added a default value for the certificate paths. Can you take another look?

abutcher:
aos-ci-test

abutcher:
bot, test this please

sdodson:
[merge]

enoodle:
"Error from server (NotFound): namespaces \"logging\" not found"
"/go/src/github.com/openshift/origin/_output/local/go/src/github.com/openshift/origin/vendor/k8s.io/kubernetes/test/e2e/scheduling/predicates.go:73
I0726 01:51:51.623994   17300 request.go:782] Error in request: resource name may not be empty"
 - I am not sure what jenkins is testing there exactly but it doesn't seem connected.

abutcher:
flake https://github.com/openshift/origin/issues/15356

abutcher:
[merge]

PR #4695, opened 2017-07-06
Test docker restart with retries 3 delay 30

ashcrow:
/cc @sdodson

### DIFF
--- a/playbooks/adhoc/uninstall.yml
+++ b/playbooks/adhoc/uninstall.yml
@@ -314,6 +314,10 @@
     service: name=docker state=restarted
     ignore_errors: true
     when: not (container_engine | changed)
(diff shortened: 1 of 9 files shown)


playbooks/adhoc/uninstall.yml
```diff
@@ -314,6 +314,8 @@
     service: name=docker state=restarted
     ignore_errors: true
     when: not (container_engine | changed)
+    retries: 3
+    delay: 30
```
sdodson:
```
- name: restart docker
  systemd:
    name: "{{ openshift.docker.service_name }}"
    state: restarted
  register: r_docker_restart_docker_result
  until: not r_docker_restart_docker_result | failed
  retries: 1
  delay: 30
```

PR #4696, opened 2017-07-06
Bug 1466152 - Json-file log driver: Neither "openshift_logging_fluentd_use_journal=false" nor omitted collects the log entries

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1466152
Do not set openshift_logging_fluentd_use_journal or USE_JOURNAL at
all unless it is explicitly set as an ansible param.  It is almost
always better to let fluentd figure out which log driver docker
is using.
@jcantrill @ewolinetz PTAL


ewolinetz:
@richm  why don't we just deprecate that parameter instead from the ansible install? If we should always let Fluentd figure out what it should be using, is there a scenario where we would want to allow it to be overridden?  We could update the logic also to print out a message if its being set that its been deprecated and instead it'll be ignored? (alternatively we could just ignore it silently)

jcantrill:
@richm @ewolinetz  I'm inclined to remove it all together and force users to have to add it manually into the DS.  I can not think of a scenario where they would need to specify it.  The logging system is only going to work when all the configuration is in sync anyway.

richm:
@ewolinetz @jcantrill addressed your comments

PR #4697, opened 2017-07-06
Wrap docker stop in retries

sdodson:


sdodson:
@ashcrow PTAL

This should cover all places where we're stopping an individual service. There's a few places where we loop through services and we'll want to investigate how to deal with those in the future but I think this will be good enough for now.

ashcrow:
@sdodson LGTM. Merge away.

ashcrow:
@sdodson I do see a few more places that stop docker ...

sdodson:
@ashcrow I ignored anything in playbooks/adhoc as those shouldn't be used.

PR #4704, opened 2017-07-06
Make pre and post master upgrade storage migrations optional

sdodson:
and non fatal

```
Multiple data migrations take place and if they fail they will fail the upgrade
You may wish to disable these or make them non fatal

# openshift_upgrade_pre_storage_migration_enabled=true
# openshift_upgrade_pre_storage_migration_fatal==true
# openshift_upgrade_post_storage_migration_enabled=true
# openshift_upgrade_post_storage_migration_fatal==true

```


sdodson:
@mwoodson @jupierce FYI

PR #4705, opened 2017-07-06
ansible.cfg: improve ssh ControlPath

sosiouxme:
AWS hosts and user names frequently make the default Ansible ControlPath
too long. This prevents ssh persistent connections, significantly
impacting runtime. More detail: https://github.com/ansible/ansible/issues/11536

This config change shortens the ControlPath to make this less likely. It
would be better to change it to the %C hash, as extremely long hostnames
or usernames could still encounter this problem, but that is not yet
available with RHEL's openssh. This at least improves the situation.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -29,3 +29,7 @@ nocows = True
 # Additional ssh options for OpenShift Ansible
 [ssh_connection]
 pipelining = True
+# shorten the ControlPath which is often too long; when it is,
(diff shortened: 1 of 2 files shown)


tbielawa:
aos-ci-test

sosiouxme:
@tbielawa test failure was nothing... can I get a thumbs up?

sosiouxme:

PR #4708, opened 2017-07-07
Upgrading control_plane fails when user is not system:admin

jkaurredhat:


sdodson:
backports #4709

PR #4713, opened 2017-07-07
Only store failures that were not ignored.

rhcarvalho:
In the past, health checks were implemented with ignore_errors: True in
the playbook level, requiring us to store all failures, ignored or not,
so that we could report on all failed checks.

Now checks are run from a single action plugin entry point, without
ignoring errors (all errors are aggregated via the action plugin).

Since the integration of the openshift_health_checker role with the
install playbook, failure summaries are part of the output of a lot more
calls to ansible-playbook. We shall report only failures that caused the
execution to stop, as ignored failures in the summary only serve to
confuse users.

----

Note: this change leaves room for some clean up, but to avoid introducing a bug at this point I'm deferring the cleanup to a separate PR.

### DIFF
--- a/roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
+++ b/roles/openshift_health_checker/callback_plugins/zz_failure_summary.py
@@ -39,7 +39,8 @@ class CallbackModule(CallbackBase):
 
     def v2_runner_on_failed(self, result, ignore_errors=False):
         super(CallbackModule, self).v2_runner_on_failed(result, ignore_errors)
-        self.__failures.append(dict(result=result, ignore_errors=ignore_errors))
+        if not ignore_errors:
+            self.__failures.append(dict(result=result, ignore_errors=ignore_errors))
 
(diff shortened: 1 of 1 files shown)


sdodson:
I think we should move these to `failed_when:` on these tasks to make sure they don't show up as a failure.  ignore_errors: true means that ansible will return non zero which i think we should consider to be a failure and we should display any of those.

sosiouxme:

rhcarvalho:
After talking with Scott on IRC I was not so sure we wanted to merge this.
Let's leave for the future, I think it is better to include ignored errors, but somehow distinguish them from non-ignored.
In any case, the full output is still above the summary (this callback plugin's addition).

sosiouxme:
Err... too late, it merged last night. But I think it's an improvement, for now, not to have this sort of failure show up with "???" as the message. That was just confusing. If we can do something actually informative with this type of failure then that would be better, though I think the direction Scott indicated is to have them not considered failures at all.

PR #4715, opened 2017-07-07
downstream dockerfile fixes

sosiouxme:
True up with fixes I put directly in dist-git (or want to) for the OCP build. I didn't think this got automatically synced like everything else but it does.

### DIFF
--- a/images/installer/Dockerfile.rhel7
+++ b/images/installer/Dockerfile.rhel7
@@ -20,7 +20,7 @@ LABEL name="openshift3/ose-ansible" \
 # because all content and dependencies (like 'oc') is already
 # installed via yum.
 USER root
-RUN INSTALL_PKGS="atomic-openshift-utils atomic-openshift-clients python-boto" && \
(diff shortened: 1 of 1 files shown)


images/installer/Dockerfile.rhel7
```diff
@@ -2,16 +2,20 @@ FROM openshift3/playbook2image
...
 
+# override env vars from base image
+ENV SUMMARY="OpenShift's installation and configuration tool" \
+    DESCRIPTION="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain and check an OpenShift cluster"
+
 LABEL name="openshift3/ose-ansible" \
-      summary="OpenShift's installation and configuration tool" \
-      description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain and check an OpenShift cluster" \
+      summary="$SUMMARY" \
+      description="$DESCRIPTION" \
```
codificat: Was this change in the labels intentional? 

Asking because I don't know where/how these variables would be processed

sosiouxme: The labels are not actually changed... I just wanted the labels and the env vars to be consistent (the env vars were all about the python scl). AFAIK nothing uses either, they're just visible for user inspection.

sosiouxme:

PR #4718, opened 2017-07-07
Enable JMX reporting of internal metrics

jsanda:
We need to enable jmx reporting of our internal, dropwizard metrics
so that they can be exposed over prometheus endpoint.


ewolinetz:
LGTM, @mwringe any objections on this?

ewolinetz:
We will have to wait on merging this until the queue is opened again for 3.6.1

PR #4722, opened 2017-07-10
Add overlay to supported Docker storage drivers

rhcarvalho:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1467809

As a next step, we can refine under which conditions the overlay driver
is supported.


rhcarvalho:
@sdodson PTAL

rhcarvalho:
Aha, we need to update the unit-tests too :)

sosiouxme:

PR #4726, opened 2017-07-10
Require cluster ID be set when using AWS cloudprovider

sdodson:
Deliberately split into two commits so the addition of the variable can be included in older releases.


sdodson:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
Closing this, if aws.conf does not contain the cluster id then it's going to be read from the label of the master. So rather than adding yet another place where we need to synchronize items lets just rely on the instance being labeled.

PR #4728, opened 2017-07-10
add check for scheduled pods

juanvallejo:
Adds a check that ensures pods have been scheduled (and that containerStatuses exist) before attempting to retrieve pod container statuses.

Addresses BZ: [1468760](https://bugzilla.redhat.com/show_bug.cgi?id=1468760)

cc @sosiouxme @rhcarvalho

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/logging/logging.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/logging.py
@@ -54,12 +54,12 @@ class LoggingCheck(OpenShiftCheck):
         """Returns: list of pods not in a ready and running state"""
         return [
             pod for pod in pods
-            if any(
+            if not pod.get("status", {}).get("containerStatuses") or any(
                 container['ready'] is False
                 for container in pod['status']['containerStatuses']
             ) or not any(
                 condition['type'] == 'Ready' and condition['status'] == 'True'
-                for condition in pod['status']['conditions']
+                for condition in pod['status'].get('conditions', [])
             )
         ]
 
--- a/roles/openshift_health_checker/test/logging_check_test.py
+++ b/roles/openshift_health_checker/test/logging_check_test.py
@@ -50,6 +50,16 @@ plain_kibana_pod = {
     }
 }
 
+plain_kibana_pod_no_containerstatus = {
+    "metadata": {
+        "labels": {"component": "kibana", "deploymentconfig": "logging-kibana"},
+        "name": "logging-kibana-1",
+    },
+    "status": {
+        "conditions": [{"status": "True", "type": "Ready"}],
+    }
+}
+
 fluentd_pod_node1 = {
     "metadata": {
         "labels": {"component": "fluentd", "deploymentconfig": "logging-fluentd"},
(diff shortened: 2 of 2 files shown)


juanvallejo:
aos-ci-test

roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -49,9 +49,10 @@ def get_pods_for_component(self, execute_module, namespace, logging_component, t
 
         return pods['items'], None
 
-    @staticmethod
-    def not_running_pods(pods):
-        """Returns: list of pods not in a ready and running state"""
+    def not_running_pods(self, pods):
+        """Returns: list of pods not in a ready and running state
+        or an OpenShiftCheckException if no pods have been scheduled."""
+        self.ensure_scheduled_pods(pods)
```
sosiouxme: So after thinking about this for a bit, while I like the idea of informing users about specific problems with their pods, this just does not feel like the right place to do it. And I think there could be legitimate reasons why a pod might not be working but not actually impacting anything because there's already a replacement, so I wouldn't want the check to fail on any particular pod having problems, just if there are the wrong number of pods actually in service.

WDYT?

roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -49,9 +49,10 @@ def get_pods_for_component(self, execute_module, namespace, logging_component, t
...
         return [
             pod for pod in pods
             if any(
```
sosiouxme: We could just stick to the goal of filtering out pods that aren't actually contributing, and fix the check breakage like so:

```
if not pod.get('status', {}).get('containerStatuses') or any(
... and later ...
    for condition in pod['status'].get('conditions', [])
```

roles/openshift_health_checker/openshift_checks/logging/logging.py
juanvallejo: @sosiouxme 
> And I think there could be legitimate reasons why a pod might not be working but not actually impacting anything because there's already a replacement, so I wouldn't want the check to fail on any particular pod having problems

I think that is a valid point. Will update this change to only fail if no pods or `> 1` pods are found to be `Ready`

sosiouxme:
More than 1 pods can be valid too... I vote for just the minimal change
here.

juanvallejo:
@sosiouxme thanks, updated check based on https://github.com/openshift/openshift-ansible/pull/4728#discussion_r126720192

roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -51,15 +51,16 @@ def get_pods_for_component(self, execute_module, namespace, logging_component, t
 
     @staticmethod
     def not_running_pods(pods):
-        """Returns: list of pods not in a ready and running state"""
+        """Returns: list of pods not in a ready and running state
+        or an OpenShiftCheckException if no pods have been scheduled."""
```
sosiouxme: doesn't throw an exception now :)

sosiouxme: otherwise LGTM... can you squash commits?

juanvallejo: > doesn't throw an exception now :)

:/ saw this as I started to make changes and knew I would somehow forget to edit it out before pushing

juanvallejo:
@sosiouxme thanks for the feedback. commits squashed

juanvallejo:
aos-ci-test

sosiouxme:

PR #4729, opened 2017-07-10
openshift_checks: fix execute_module params

sosiouxme:
Fix where execute_module was being passed task_vars in place of tmp
param. Most modules don't seem to use either and so this doesn't fail;
but under some conditions (perhaps different per version of ansible?) it
tried to treat the dict as a string and came back with a python stack
trace.


sosiouxme:

rhcarvalho:
:+1:

PR #4736, opened 2017-07-11
Add an SA policy to the ansible-service-broker

rthallisey:
We are not adding a policy to the service account after creation.
The ansible-service-broker will require cluster-admin permissions
because we do things like: creating service accounts, projects,
and pods.

### DIFF
--- a/roles/ansible_service_broker/tasks/install.yml
+++ b/roles/ansible_service_broker/tasks/install.yml
@@ -47,7 +47,7 @@
     state: present
     namespace: "openshift-ansible-service-broker"
     resource_kind: cluster-role
-    resource_name: cluster-admin
+    resource_name: admin
     user: "system:serviceaccount:openshift-ansible-service-broker:asb"
 
 - name: create ansible-service-broker service
--- a/roles/ansible_service_broker/tasks/install.yml
+++ b/roles/ansible_service_broker/tasks/install.yml
@@ -42,6 +42,14 @@
     namespace: openshift-ansible-service-broker
     state: present
 
+- name: Set SA cluster-role
+  oc_adm_policy_user:
+    state: present
+    namespace: "openshift-ansible-service-broker"
+    resource_kind: cluster-role
+    resource_name: cluster-admin
+    user: "system:serviceaccount:openshift-ansible-service-broker:asb"
+
 - name: create ansible-service-broker service
   oc_service:
     name: asb


ewolinetz:
@pmorie is this too open of permissions?

ewolinetz:
@rthallisey we create a role and role binding here for that SA: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_service_catalog/files/kubeservicecatalog_roles_bindings.yml#L93-L148

rthallisey:
@ewolinetz This is for a broker, not the service-catalog. I don't think it makes sense to patch the service-catalog playbook regarding the broker. Can you explain further what you mean?

ewolinetz:
@rthallisey ah my mistake, i didn't read it closely enough. Could we not do something similar though for the service broker though? My concern is that being a `cluster-admin` would be too general.

jwmatthews:
Quick note on why we are using 'cluster-admin' for the service account running the broker.

For 3.6.0 we are creating dynamic service accounts to run APBs, after APB execution we delete the dynamic service account.

The Ansible Service Broker requires a service account that is able to:
- Create a new namespace
- Create a dynamic service account in the new namespace to run APBs
- Run APBs in that project, which will deploy applications inside the namespace, using the dynamic service account
- Delete the dynamic service account

rthallisey:
@ewolinetz For the most part, the broker is effectively acting as an admin. It creates resources like namespaces, projects, service accounts, pods and can read/exec to pods in other namespaces. At most, I think a custom role would land it somewhere in between admin in cluster-admin.

ewolinetz:
Right, and I guess since the exact verbs and api objects are known, I'm wondering if we couldn't create a role that will give exactly that access? I just feel that `cluster-admin` gives too much if what's necessary is as you describe above.

rthallisey:
Ok. I'll build a custom role.

rthallisey:
Definitions:
service-catalog -> Kubernetes API that contacts that broker with work
broker -> Running pod that spawns an APB to complete the work from the service-catalog
APB -> Pod executing an ansible playbook to create/delete an application

While testing the custom role I hit the below error. If I understand the error, it's not allowing the broker to create a service-account with cluster-admin privileges without having the same level of privileges. 
```
Error from server (Forbidden): error when creating "/tmp/asb-resource-files/apb-dfa9a28c-92dd-42e2-b5a3-ddfb2371946b.yaml": rolebinding.authorization.openshift.io "apb-dfa9a28c-92dd-42e2-b5a3-ddfb2371946b" is forbidden: user "system:serviceaccount:ansible-service-broker:asb" cannot grant extra privileges:
{Verbs:["*"], APIGroups:["*"], Resources:["*"]}
{Verbs:["*"], NonResourceURLs:["*"]}
```

The service account the broker creates is for the APB. The APB has to be privilaged in some way because the APB functions like a user with admin privileges creating an application. Since the broker's privilege level depends on the APB, I think the broker should be using cluster-admin for now and I'll work on outlining the maximum level of privileges an APB can have. If I'm able to lower the APB below cluster-admin, I'll follow up with a change to limit the broker's privileges.

I created an issue to track this: https://github.com/openshift/ansible-service-broker/issues/269

Does that work for you @ewolinetz?

jwmatthews:
For 3.6.0 we would like to proceed with the Ansible Service Broker running with 'cluster-admin'.

We have a BZ entered for 3.7.0 to come back and address this.
https://bugzilla.redhat.com/show_bug.cgi?id=1470824

Is this acceptable?  @sdodson @ewolinetz @smarterclayton

ewolinetz:
I'm going to defer to @smarterclayton on whether or not we're alright with this level of access (it seems like its going to be required).

If we're going to go that route, can we update the PR to use the module as initially proposed? Currently we're creating a file and creating and processing... we can likely simplify it @rthallisey

fabianvf:
found an old diff for the module version, in case anyone else forgot what it looked like:

```yaml
- name: Set SA cluster-role
   oc_adm_policy_user:
     state: present
     namespace: "openshift-ansible-service-broker"
     resource_kind: cluster-role
     resource_name: cluster-admin
     user: "system:serviceaccount:openshift-ansible-service-broker:asb"
```

jwmatthews:
Spoke with @smarterclayton 

Current plan is for the broker to switch to using project 'admin' for the dynamic service account, then for the broker's own service account to use admin plus whatever else is needed.

I've asked @fabianvf to help testing this combination of roles to identify what will work for 3.6.0

jwmatthews:
FYI, this is the 3.6.0 BZ tracking this change.
https://bugzilla.redhat.com/show_bug.cgi?id=1470861

ewolinetz:
we'll also need to cherry pick this to the 3.6 branch

PR #4738, opened 2017-07-11
Additional optimization parameters for ansible.cfg

### LINKED ISSUE #4734: ansible.cfg does not follow OpenShift install optimization recommendations
Now that we are shipping ansible.cfg in the repository, we should add [our known best settings](https://docs.openshift.com/container-platform/3.5/scaling_performance/install_practices.html#scaling-performance-install-optimization) to optimize the installation out of the box.

```
# cat /etc/ansible/ansible.cfg
# config file for ansible -- http://ansible.com/
# ==============================================
[defaults]
forks = 20 
host_key_checking = False
remote_user = root
roles_path = roles/
gathering = smart
fact_caching = jsonfile
fact_caching_connection = $HOME/ansible/facts
fact_caching_timeout = 600
log_path = $HOME/ansible.log
nocows = 1
callback_whitelist = profile_tasks

[privilege_escalation]
become = False

[ssh_connection]
ssh_args = -o ControlMaster=auto -o ControlPersist=600s
control_path = %(directory)s/%%h-%%r
pipelining = True 
timeout = 10
```

sjug:
Parameters to match our [documented optimized ansible.cfg](https://docs.openshift.com/container-platform/3.5/scaling_performance/install_practices.html#scaling-performance-install-optimization)

Fixes #4734

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -16,6 +16,13 @@ host_key_checking = False
 retry_files_enabled = False
 retry_files_save_path = ~/ansible-installer-retries
 nocows = True
+remote_user = root
+roles_path = roles/
+gathering = smart
+fact_caching = jsonfile
+fact_caching_connection = $HOME/ansible/facts
+fact_caching_timeout = 600
+callback_whitelist = profile_tasks
 
 # Uncomment to use the provided BYO inventory
 #hostfile = inventory/byo/hosts
@@ -29,7 +36,8 @@ nocows = True
 # Additional ssh options for OpenShift Ansible
 [ssh_connection]
 pipelining = True
+ssh_args = -o ControlMaster=auto -o ControlPersist=600s
+timeout = 10
 # shorten the ControlPath which is often too long; when it is,
 # ssh connection reuse silently fails, making everything slower.
 control_path = %(directory)s/%%h-%%r
-


ansible.cfg
```diff
@@ -26,6 +33,12 @@ nocows = True
 # Uncomment to use the provided AWS dynamic inventory script
 #hostfile = inventory/aws/ec2.py
 
+[privilege_escalation]
+become = False
```
sdodson: Whats the purpose of this?

jeremyeder: Wanted to make sure we weren't escalating by default and playbooks could override it.  Actually it appears that false is the default anyway, so this can probably be dropped.

jeremyeder:
Status of this?

ansible.cfg
```diff
@@ -29,7 +36,8 @@ nocows = True
 # Additional ssh options for OpenShift Ansible
 [ssh_connection]
 pipelining = True
+ssh_args = -o ControlMaster=auto -o ControlPersist=600s
+timeout = 10
```
sdodson: this is the default, I'd remove it

ansible.cfg
```diff
@@ -16,6 +16,13 @@ host_key_checking = False
 retry_files_enabled = False
 retry_files_save_path = ~/ansible-installer-retries
 nocows = True
+remote_user = root
```
sdodson: We tend to control this via ansible_ssh_user being set in the inventory, do you not set that?

sdodson:
@sosiouxme mind taking a look to see if you're concerned about setting any of these? I believe this makes its way into the container so it will affect your defaults there.

jeremyeder:
Will this be in 3.6? I ask because if so, we will update the docs as well.

sdodson:
I'd prefer waiting but I'll defer to your judgement.

PR #4739, opened 2017-07-11
Reduce openshift_facts dependencies.

kwoodson:
The summary of this PR is to remove `openshift_facts/tasks` and move them to `initialize_facts.yml`.  `playbooks/common/openshift-cluster/initilize_facts.yml` is included as part of the `playbooks/common/openshift-cluster/std_include.yml` and **ALL** entry point playbooks should flow through this playbook so that the following occur:
- groups are initialized and our oo_<groups> are defined and populated
- openshift_version is properly determined and set
- openshift_facts are set on hosts upfront before any other roles/playbooks run.

If you are seeing issues with your playbooks outside of common, please ensure that the playbooks include the `playbooks/common/openshift-cluster/std_include.yml` playbook.


playbooks/common/openshift-cluster/initialize_facts.yml
```diff
@@ -6,12 +6,140 @@
...
-  - openshift_facts
   tasks:
-  - openshift_facts:
+  - name: load openshift_facts module
+    include_role:
+      name: openshift_facts
+
+  # TODO: Should this role be refactored into health_checks??
+  - name: Run openshift_sanitize_inventory to set variables
+    include_role:
```
ingvagabund: Can you put both roles into `roles` property and move are tasks under `post_tasks`? As long as we can use `roles` instead of `include_role`, I find the plays more transparent.

playbooks/common/openshift-cluster/initialize_facts.yml
```diff
@@ -6,12 +6,140 @@
...
+    set_fact:
+      l_any_system_container: "{{ l_is_etcd_system_container or l_is_openvswitch_system_container or l_is_node_system_container or l_is_master_system_container }}"
+
+  - name: initialize_facts set fact for l_etcd_runtime
+    set_fact:
+      l_etcd_runtime: "{{ 'runc' if l_is_etcd_system_container else 'docker' if l_is_containerized else 'host' }}"
+
+  # TODO: Should this be moved into health checks??
+  # Seems as though any check that happens with a corresponding fail should move into health_checks
+  - name: Validate python version - ans_dist is fedora and python is v3
```
ingvagabund: Maybe move both `Validate python version - ans_dist is fedora and python is v3` and `Validate python version - ans_dist not Fedora and python must be v2` into a new role called `init-checks`? Given both tasks can be moved up and down, the new role can be added to the `roles` list.

playbooks/common/openshift-cluster/initialize_facts.yml
```diff
@@ -6,12 +6,140 @@
...
+    fail:
+      msg: "openshift-ansible requires Python 2 for {{ ansible_distribution }}"
+    when:
+    - ansible_distribution != 'Fedora'
+    - ansible_python['version']['major'] != 2
+
+  # TODO: Should this be moved into health checks??
+  # Seems as though any check that happens with a corresponding fail should move into health_checks
+  # Fail as early as possible if Atomic and old version of Docker
+  - block:
```
ingvagabund: Good candidate for the `init-checks` role.

playbooks/common/openshift-cluster/initialize_facts.yml
kwoodson: @ingvagabund, `include_role` is the future as per our very own cluster lifecycle meeting with the ansible consultant.

playbooks/common/openshift-cluster/initialize_facts.yml
kwoodson: @ingvagabund, this is an interesting idea.  I'd prefer to have less role dependencies.  If the preflight health_checks are already doing this exact type of checking then I'd prefer if we just did these types of checks inside of the health_checks.  Seems like the best place for them.

playbooks/common/openshift-cluster/initialize_facts.yml
ingvagabund: I will be always opposed to that as long as the original `roles` property of a play can provide the same functionality. Though, it is my own opinion and not a blocker for the PR :).

playbooks/common/openshift-cluster/initialize_facts.yml
kwoodson: Same as above.  Any `fail - msg` type checks are candidates for `health` or `preflight`.  I'd prefer to limit the number of dependency calls.  The attempt here is to do them upfront in the initial startup and then never do them again.  If they go into a role dependency then we have a tendency to include them in `mete/main.yml` for multiple roles.

We can definitely have the discussion about whether `health_checks` vs placing these types of checks into a `role` is superior.  Maybe a discussion for the architecture team.  I can see an argument for both sides but I'd prefer to have checks in one place so that in the future we know exactly where to place code like this.

playbooks/common/openshift-cluster/initialize_facts.yml
ingvagabund: `health_checks` by the name checks health of the cluster. Given it can be anything (with a proper definition of what health is), I agree with you.

playbooks/common/openshift-cluster/initialize_facts.yml
kwoodson: @ingvagabund, heh, I agree that at first the new style appears different but I can assure you the benefits are there.  We discussed this very topic with James from Ansible and he was telling us that when you use `include_role` the playbook sections `pre_tasks`, `roles`, `tasks`, and `post_tasks` ordering goes away.  The flow of the playbook moves top to bottom and the flexibility increases as you can include only tasks from a role that you desire.  With `roles:` you get the entire role if that is desired or not.

I can find the meeting recording and send it to you.  I think its in the architecture notes.

kwoodson:
Currently failing to start atomic-openshift-node with this error:
```
 master has not created a default cluster network, network plugin "redhat/openshift-ovs-subnet" can not start
```

ashcrow:
I like this move overall. It makes a lot of sense!

sdodson:
Is atomic-openshift-sdn-ovs being installed?

kwoodson:
@sdodson, atomic-openshift-sdn-ovs-3.6.74-1.git.0.e6d1637.el7.x86_64

kwoodson:
@sdodson @abutcher pointed me in the right direction.  The master-config.yaml was getting a bad value for the CIDR.  This was caused by an invalid input parameter from my template.  

The install was successful with these changes.

playbooks/common/openshift-cluster/initialize_facts.yml
mtnbikenc: This is more stylistic, but for blocks my preference is for putting the `when:` condition at the top of the block.  Such as,
```yaml
- when: 
  - l_is_atomic | bool
  block:
```

playbooks/common/openshift-cluster/initialize_facts.yml
```diff
@@ -6,12 +6,140 @@
...
+      # See https://access.redhat.com/articles/2317361
+      # and https://github.com/ansible/ansible/issues/15892
+      # NOTE: the "'s can not be removed at this level else the docker command will fail
+      # NOTE: When ansible >2.2.1.x is used this can be updated per
+      # https://github.com/openshift/openshift-ansible/pull/3475#discussion_r103525121
+      - name: Determine Atomic Host Docker Version
+        shell: 'CURLY="{"; docker version --format "$CURLY{json .Server.Version}}"'
+        register: l_atomic_docker_version
+
+      - assert:
```
mtnbikenc: Add a task name.

playbooks/common/openshift-cluster/initialize_facts.yml
```diff
@@ -6,12 +6,140 @@
...
+          msg: Installation on Atomic Host requires Docker 1.12 or later. Please upgrade and restart the Atomic Host.
+          that:
+          - l_atomic_docker_version.stdout | replace('"', '') | version_compare('1.12','>=')
+    when:
+    - l_is_atomic | bool
+
+  - block:
```
mtnbikenc: Same comments about `when:` before `block:`.

roles/openshift_facts/vars/Fedora.yml
```diff
@@ -1,6 +0,0 @@
----
-required_packages:
-  - iproute
-  - python3-dbus
```
mtnbikenc: Need to add back in the need for python3-dbus on Fedora.

roles/openshift_facts/tasks/main.yml
```diff
@@ -1,116 +0,0 @@
...
-  - assert:
-      msg: Installation on Atomic Host requires Docker 1.12 or later. Please upgrade and restart the Atomic Host.
-      that:
-      - l_atomic_docker_version.stdout | replace('"', '') | version_compare('1.12','>=')
-
-  when:
-  - l_is_atomic | bool
-  - r_openshift_facts_ran is not defined
-
-- name: Load variables
```
mtnbikenc: This task didn't make it over to the new file.  Therefore the logic for Fedora is missing and will not install the proper `required_packages`.

playbooks/common/openshift-cluster/initialize_facts.yml
kwoodson: @mtnbikenc, Does assert take a name?

roles/openshift_facts/tasks/main.yml
kwoodson: I missed that small discrepancy between fedora and the default.

playbooks/common/openshift-cluster/initialize_facts.yml
kwoodson: tested it, will add.

roles/openshift_facts/vars/Fedora.yml
kwoodson: I added this back in.

roles/openshift_facts/tasks/main.yml
kwoodson: added back in a the install section.  good catch

kwoodson:
aos-ci-test

kwoodson:
@mtnbikenc, squashed!

kwoodson:
@sdodson, I'd love to get this merged at some point.  Any thoughts?

sdodson:
I think given how bad we were in the past with proliferation of calls to openshift_facts we should hold this until after 3.6 is forked so that we get multiple weeks of soak time. We'll merge this as soon as we fork 3.6.

mtnbikenc:
Now that 3.6 is branched, can we move forward with merging this?
[test]

sdodson:
> Now that 3.6 is branched, can we move forward with merging this?

yeah, lets do it!

kwoodson:
> common.go:54: missing in output: ["check \"package_update\":" "Error getting data from at least one yum repository"]

I'm pretty sure this is unrelated.

sdodson:
[merge]

sdodson:
@abutcher @mtnbikenc @kwoodson Sorry I was tied up with other stuff today, following up on irc and offline discussion I'd say first and foremost our obligation is not to merge code that we know breaks supported playbooks. So we should either fix those in this PR or we should line up separate PRs as pre-requisites for this PR, ensuring that the work is clearly linked will be key if we ever try to backport.

tbielawa:
I see this one is approved, scott said to give it the ol' merge. Does it need another test run, or can we go ahead and merge this in?

Also, @sdodson said

```
Sorry I was tied up with other stuff today, following up on irc and offline discussion I'd say first and 
foremost our obligation is not to merge code that we know breaks supported playbooks. So we should
either fix those in this PR or we should line up separate PRs as pre-requisites for this PR, ensuring
that the work is clearly linked will be key if we ever try to backport.
```

Still relevant?

sdodson:
[merge]

sdodson:
Since this will be the place that people end up when they look at the history for roles/openshift_facts can you make sure that the description is updated to include how one needs to update their playbooks to work with this?

kwoodson:
@sdodson, i updated the description and the advice to include the `std_include.yml`.

ashcrow:
bot, retest this please

ashcrow:
We've gotten two different failures with the atomic test. Kicking it again to see if we get one of the same failures.

bot, retest this please

ashcrow:
I think the following needs to occur to get the proper testing in atomic done:

```diff
diff --git a/.papr.yml b/.papr.yml
index 16d6e78b..6658720e 100644
--- a/.papr.yml
+++ b/.papr.yml
@@ -33,7 +33,7 @@ packages:
 context: 'fedora/25/atomic'
 
 env:
-  OPENSHIFT_IMAGE_TAG: v3.6.0-alpha.1
+  OPENSHIFT_IMAGE_TAG: v3.6.0
 
 tests:
   - ./.papr.sh
```

sdodson:
aos-ci-test

sdodson:
[test]

ashcrow:
With ``papr`` update https://github.com/openshift/openshift-ansible/pull/5007

sdodson:
If those two tests come back green I don't care if fedora/25/atomic fails I'll just merge it.

ashcrow:
bot, retest this please

ashcrow:
@sdodson to clarify, you mean you are not worried about the atomic test failure and believe it to be a flake, correct? 😄

sdodson:
> @sdodson to clarify, you mean you are not worried about the atomic test failure and believe it to be a flake, correct?

yes, just that, the rest i care a lot about

ashcrow:
Same problem with the atomic test happens on master. I agree with @sdodson and it can merge without passing. I also recommend keeping the ``papr`` commit in there as we shouldn't be targeting ```3.6.0.alpha.1``` any longer.

kwoodson:
@sdodson @ashcrow, Merge this early Monday?

kwoodson:
@sdodson, @ashcrow, I rebased which then had a conflict.  The conflict was that the `openshift_facts/tasks/main.yml` was removed.  It should be good now.

kwoodson:
```
--- FAIL: TestPackageUpdateRepoUnreachable (23.23s)
	common.go:54: missing in output: ["check \"package_update\":" "Error getting data from at least one yum repository"]
	common.go:98: 
```
I'm not familiar with why these tests were failing.

kwoodson:
```
Failure summary:

  1. Host:     localhost
     Play:     Verify Requirements
     Details:  check "docker_image_availability":
               One or more required Docker images are not available:
                   openshift/origin-docker-registry:ddb26b4,
                   openshift/origin-haproxy-router:ddb26b4
               Configured registries: docker.io
```

ashcrow:
@kwoodson I noticed this in the logs:

```
++ Could not parse metalink 'https://mirrors.fedoraproject.org/metalink?repo=epel-7'
OPENSHIFT_ANSIBLE_VARS: line 1: Could: command not found
OPENSHIFT_ANSIBLE_VARS: line 1: error: command not found
```

Not sure what the heck would cause that.

sdodson:
[test]

ashcrow:
The error here seems off. It's failing testing because the rest of the tests have not returned yet.

abutcher:
bot, retest this please

sdodson:
@sosiouxme or @rhcarvalho Can either of you help debug what's going on with these integration tests?

sosiouxme:
@sdodson https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible/409/ looks like all the yum interactions failed, see e.g. "Error with yum repository configuration: Cannot find a valid baseurl for repo" - probably a yum bobble at exactly the time this was running?

test/integration/openshift_health_checker/setup_container.yml
```diff
@@ -46,3 +46,4 @@
 
 - include: ../../../playbooks/byo/openshift-cluster/initialize_groups.yml
 - include: ../../../playbooks/common/openshift-cluster/evaluate_groups.yml
+- include: ../../../playbooks/common/openshift-cluster/std_include.yml
```
kwoodson: @sdodson, the `std_include.yml` does the `evaluate_groups.yml` call.

kwoodson:
@sosiouxme, thanks.  I'm trying to figure out why this is happening.  It could be related to the PR but I find it unlikely.  If so, then I need to be able to fix it.  Any advice on testing locally or fixing would be great.

ashcrow:
If it is yum flaking out it's doing it a lot more than I'd expect ☹️. I tend to agree with @kwoodson that the problem isn't something specific to this PR.

sosiouxme:
@kwoodson yeah i think that was a flake but when making fairly fundamental changes like this it's good to run the integration tests locally so they don't surprise you at the end of a test or merge.

https://github.com/openshift/openshift-ansible/blob/master/test/integration/README.md describes how to do that; let me know if you get stuck. I can try running against this PR myself just to see if it was a flake... they should only take 5-10 minutes to run locally.

kwoodson:
@sosiouxme, I ran the test locally and if failed the same way.  I'm not sure where to go from here as my PR didn't touch the repos.

ashcrow:
Looking at the errors again it almost reads as if the expected errors are NOT occurring, hence failure.

sdodson:
The commit I've added made things worse. Drop that commit and we're back to `openshift` undefined which just means the integration playbooks need to be updated to include openshift_facts in a meaningful manner.

ashcrow:
Example:
```
		  1. Host:     openshift_ansible_test_72782670407390
		     Play:     Determine openshift_version to configure on first master
		     Task:     openshift_version : fail
		     Message:  Package atomic-openshift not found
--- FAIL: TestPackageUpdateRepoUnreachable (99.15s)
	common.go:54: missing in output: ["check \"package_update\":" "Error getting data from at least one yum repository"]
	common.go:98: 
		$ (cd /data/src/github.com/openshift/openshift-ansible/test/integration/openshift_health_checker/preflight && ansible-playbook -i /dev/null playbooks/package_update_repo_unreachable.yml)
		...
		localhost                  : ok=25   changed=2    unreachable=0    failed=0   
		openshift_ansible_test_32116651946969 : ok=31   changed=5    unreachable=0    failed=1 
```

which is from:

```go
func TestPackageUpdateRepoUnreachable(t *testing.T) {
        PlaybookTest{
                Path:     "playbooks/package_update_repo_unreachable.yml",
                ExitCode: 2,
                Output: []string{
                        "check \"package_update\":",
                        "Error getting data from at least one yum repository",
                },
        }.Run(t)
}
```

ashcrow:
Yeah, this is a test case looking for a failure and **not** getting the string result it expects.

sosiouxme:
The integration tests run actual playbooks. They probably need to be updated to work with this. I'll take a look.

kwoodson:
@sosiouxme, this change involves updating the dependencies and how we initialize a playbook run.  I have added the `std_include.yml` to the bottom of the `setup_container.yml`.  This is what is resulting in the  error:
```
	common.go:54: missing in output: ["check \"package_update\":" "Could not perform a yum update." "break-yum-update-1.0-2.noarch requires package-that-does-not-exist"]
```

sosiouxme:
Sorry it's taken me a while to get my head back into this. So, at least locally the integration tests are failing because `openshift_version` is crapping out before the test can get to what it's testing. Looking through the changes to see why this is...

sosiouxme:
The integration tests all start without openshift repos. Those are enabled as needed for the test cases, in their playbooks. So `openshift_version` needs to happen after that or it will fail. Either update every playbook, or enable a repo in the container setup just to get past the initialization then disable it... I'll try some things.

kwoodson:
@sosiouxme, I tracked it down to your previous statement.
On this PR branch:
```
[ose-3.2]
name=ose-3.2
baseurl=file:///mnt/localrepo/ose-3.2
enabled=0
gpgcheck=0
```
 vs. Master branch
```
[ose-3.2]
name=ose-3.2
baseurl=file:///mnt/localrepo/ose-3.2
enabled = 1
gpgcheck=0
```

sosiouxme:
I added a commit to handle the integration tests. It actually helped me take out some boilerplate...

PR #4740, opened 2017-07-12
fluentd,mux: make buffer_type configurable; defaults buffer type to "…

nhosoi:
…file".

Parameters for the upper file buffer limit per output.
  openshift_logging_fluentd_file_buffer_limit: 1Gi
  openshift_logging_mux_file_buffer_limit: 2Gi
E.g., if es and es-ops are configured for outputs, fluentd allocates 1Gi * 2;
mux allocates 2Gi * 2, if the size is available.

### DIFF
--- a/roles/openshift_logging/README.md
+++ b/roles/openshift_logging/README.md
@@ -167,8 +167,6 @@ Elasticsearch OPS too, if using an OPS cluster:
   set the value to the file buffer limit.
 - `openshift_logging_mux_file_buffer_storage_type`: Default `[emptydir]` - Storage
   type for the file buffer.  One of [`emptydir`, `pvc`, `hostmount`]
-- `openshift_logging_mux_file_buffer_mount_path`: Default `[/mux/filebufferstorage]`
-  - Mount path for the file buffer; Shared by `pvc` and `hostmount`.
 
 - `openshift_logging_mux_file_buffer_pvc_size`: The requested size for the file buffer
   PVC, when not provided the role will not generate any PVCs. Defaults to `4Gi`.
--- a/roles/openshift_logging/README.md
+++ b/roles/openshift_logging/README.md
@@ -57,6 +57,8 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_fluentd_hosts`: List of nodes that should be labeled for Fluentd to be deployed to. Defaults to ['--all'].
 - `openshift_logging_fluentd_buffer_queue_limit`: Buffer queue limit for Fluentd. Defaults to 1024.
 - `openshift_logging_fluentd_buffer_size_limit`: Buffer chunk limit for Fluentd. Defaults to 1m.
+- `openshift_logging_fluentd_buffer_type`: Fluentd will set the buffer type.  Defaults to 'file'.
+- `openshift_logging_fluentd_file_buffer_limit`: Fluentd will set the value to the file buffer limit.  Defaults to '1Gi' per output.
 
 
 - `openshift_logging_es_host`: The name of the ES service Fluentd should send logs to. Defaults to 'logging-es'.
@@ -160,3 +162,21 @@ Elasticsearch OPS too, if using an OPS cluster:
   need to set this
 - `openshift_logging_mux_buffer_queue_limit`: Default `[1024]` - Buffer queue limit for Mux.
 - `openshift_logging_mux_buffer_size_limit`: Default `[1m]` - Buffer chunk limit for Mux.
+- `openshift_logging_mux_buffer_type`: Mux will set the buffer type.  Defaults to 'file'.
+- `openshift_logging_mux_file_buffer_limit`: Default `[2Gi]` per output - Mux will
+  set the value to the file buffer limit.
+- `openshift_logging_mux_file_buffer_storage_type`: Default `[emptydir]` - Storage
+  type for the file buffer.  One of [`emptydir`, `pvc`, `hostmount`]
+- `openshift_logging_mux_file_buffer_mount_path`: Default `[/mux/filebufferstorage]`
+  - Mount path for the file buffer; Shared by `pvc` and `hostmount`.
+
+- `openshift_logging_mux_file_buffer_pvc_size`: The requested size for the file buffer
+  PVC, when not provided the role will not generate any PVCs. Defaults to `4Gi`.
+- `openshift_logging_mux_file_buffer_pvc_dynamic`: Whether or not to add the dynamic
+  PVC annotation for any generated PVCs. Defaults to 'False'.
+- `openshift_logging_mux_file_buffer_pvc_pv_selector`: A key/value map added to a PVC
+  in order to select specific PVs.  Defaults to 'None'.
+- `openshift_logging_mux_file_buffer_pvc_prefix`: The prefix for the generated PVCs.
+  Defaults to 'logging-mux'.
+- `openshift_logging_mux_file_buffer_storage_group`: The storage group used for Mux.
+  Defaults to '65534'.
--- a/roles/openshift_logging_mux/defaults/main.yml
+++ b/roles/openshift_logging_mux/defaults/main.yml
@@ -54,9 +54,6 @@ openshift_logging_mux_buffer_type: "file"
 # One of ['emptydir', 'pvc', 'hostmount']
 openshift_logging_mux_file_buffer_storage_type: "emptydir"
 
-# mount path: shared by hostmount and pvc
-openshift_logging_mux_file_buffer_mount_path: /var/log/mux/filebufferstorage
-
 # pvc options
 # the name of the PVC we will bind to -- create it if it does not exist
 openshift_logging_mux_file_buffer_pvc_name: "logging-mux-pvc"
--- a/roles/openshift_logging_mux/defaults/main.yml
+++ b/roles/openshift_logging_mux/defaults/main.yml
@@ -47,3 +47,26 @@ openshift_logging_mux_ops_ca: /etc/fluent/keys/ca
 #mux_config_contents:
 #mux_throttle_contents:
 #mux_secureforward_contents:
+
+# Default value of mux BUFFER_TYPE
+openshift_logging_mux_buffer_type: "file"
+
+# One of ['emptydir', 'pvc', 'hostmount']
+openshift_logging_mux_file_buffer_storage_type: "emptydir"
+
+# mount path: shared by hostmount and pvc
+openshift_logging_mux_file_buffer_mount_path: /var/log/mux/filebufferstorage
+
+# pvc options
+# the name of the PVC we will bind to -- create it if it does not exist
+openshift_logging_mux_file_buffer_pvc_name: "logging-mux-pvc"
+
+# required if the PVC does not already exist
+openshift_logging_mux_file_buffer_pvc_size: 4Gi
+openshift_logging_mux_file_buffer_pvc_dynamic: false
+openshift_logging_mux_file_buffer_pvc_pv_selector: {}
+openshift_logging_mux_file_buffer_pvc_access_modes: ['ReadWriteOnce']
+openshift_logging_mux_file_buffer_storage_group: '65534'
+
+openshift_logging_mux_file_buffer_pvc_prefix: "logging-mux"
+openshift_logging_mux_file_buffer_limit:  2Gi
--- a/roles/openshift_logging_mux/tasks/main.yaml
+++ b/roles/openshift_logging_mux/tasks/main.yaml
@@ -177,6 +177,34 @@
   check_mode: no
   changed_when: no
 
+# storageclasses are used by default but if static then disable
+# storageclasses with the storageClassName set to "" in pvc.j2
+- name: Creating Mux storage template
+  template:
+    src: pvc.j2
+    dest: "{{ tempdir }}/templates/logging-mux-pvc.yml"
+  vars:
+    obj_name: "{{ openshift_logging_mux_file_buffer_pvc_name }}"
+    size: "{{ openshift_logging_mux_file_buffer_pvc_size }}"
+    access_modes: "{{ openshift_logging_mux_file_buffer_pvc_access_modes | list }}"
+    pv_selector: "{{ openshift_logging_mux_file_buffer_pvc_pv_selector }}"
+    is_dynamic: "{{ openshift_logging_mux_file_buffer_pvc_dynamic }}"
+    storage_class_name: "{{ openshift_logging_mux_file_buffer_pvc_storage_class_name | default('', true) }}"
+  when:
+  - openshift_logging_mux_file_buffer_storage_type == "pvc"
+
+- name: Set Mux storage
+  oc_obj:
+    state: present
+    kind: pvc
+    name: "{{ openshift_logging_mux_file_buffer_pvc_name }}"
+    namespace: "{{ openshift_logging_mux_namespace }}"
+    files:
+    - "{{ tempdir }}/templates/logging-mux-pvc.yml"
+    delete_after: false
+  when:
+  - openshift_logging_mux_file_buffer_storage_type == "pvc"
+
 - name: Set logging-mux DC
   oc_obj:
     state: present
--- a/roles/openshift_logging_mux/templates/mux.j2
+++ b/roles/openshift_logging_mux/templates/mux.j2
@@ -67,13 +67,7 @@ spec:
           mountPath: /etc/fluent/muxkeys
           readOnly: true
         - name: filebufferstorage
-{% if openshift_logging_mux_file_buffer_storage_type == 'pvc' %}
-          mountPath: {{ openshift_logging_mux_file_buffer_mount_path }}
-{% elif openshift_logging_mux_file_buffer_storage_type == 'hostmount' %}
-          mountPath: {{ openshift_logging_mux_file_buffer_mount_path }}
-{% else %}
           mountPath: /var/log/mux/filebufferstorage
-{% endif %}
         env:
         - name: "K8S_HOST_URL"
           value: "{{openshift_logging_mux_master_url}}"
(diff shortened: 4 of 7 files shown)


roles/openshift_logging_mux/defaults/main.yml
```diff
@@ -47,3 +47,26 @@ openshift_logging_mux_ops_ca: /etc/fluent/keys/ca
...
 #mux_secureforward_contents:
+
+# Default value of mux BUFFER_TYPE
+openshift_logging_mux_buffer_type: "file"
+
+# One of ['emptydir', 'pvc', 'hostmount']
+openshift_logging_mux_file_buffer_storage_type: "emptydir"
+
+# mount path: shared by hostmount and pvc
+openshift_logging_mux_file_buffer_mount_path: /mux/filebufferstorage
```
jcantrill: For POSIX convention sake should we make this something like `/var/log/[fluentd|fluentd/mux|mux]/filebufferstorage`

roles/openshift_logging_mux/defaults/main.yml
```diff
@@ -47,3 +47,26 @@ openshift_logging_mux_ops_ca: /etc/fluent/keys/ca
...
+
+# pvc options
+# the name of the PVC we will bind to -- create it if it does not exist
+openshift_logging_mux_file_buffer_pvc_name: "mux-file-buffer-pvc"
```
jcantrill: We should consider matching this to the DC name in such a way that it is the same pattern you get from a statefulset.  This would ensure in future that if we could or need to move mux to a statefulset that we do not need to worry about migrating the volume or PVC

roles/openshift_logging_mux/tasks/main.yaml
```diff
@@ -177,6 +177,48 @@
   check_mode: no
   changed_when: no
 
+# storageclasses are used by default but if static then disable
+# storageclasses with the storageClassName set to "" in pvc.j2
```
jcantrill: This task looks almost identical to the one which follows it.  The difference appears to be 'storage_class_name'.  Is it possible to reduce to a single task and selectively include the storage_class_name if its provided?

roles/openshift_logging_mux/defaults/main.yml
nhosoi: Thanks!  The paths are updated in the branch/PR.

roles/openshift_logging_mux/defaults/main.yml
nhosoi: Ok!  The deploymentconfig name of mux is:
  name: logging-mux
Then, openshift_logging_mux_file_buffer_pvc_name is supposed to be "logging-mux-pvc"?

roles/openshift_logging_mux/tasks/main.yaml
nhosoi: Yes, I borrowed the code from openshift_logging_elasticsearch/tasks/main.yaml.  Are you suggesting to share the elasticsearch task with the mux (instead of having its own in the mux tasks/main.yaml)?

jcantrill: Im suggesting if we can have a single class where you always pass a value for storageclass and the template has the logic to include or not include the storageclass section based on value passed, that would seem to be preferable to me.  We otherwise have a duplicate task with a singular different value and when conditions

roles/openshift_logging/README.md
```diff
@@ -160,3 +162,21 @@ Elasticsearch OPS too, if using an OPS cluster:
...
+- `openshift_logging_mux_file_buffer_mount_path`: Default `[/mux/filebufferstorage]`
+  - Mount path for the file buffer; Shared by `pvc` and `hostmount`.
+
+- `openshift_logging_mux_file_buffer_pvc_size`: The requested size for the file buffer
+  PVC, when not provided the role will not generate any PVCs. Defaults to `4Gi`.
+- `openshift_logging_mux_file_buffer_pvc_dynamic`: Whether or not to add the dynamic
+  PVC annotation for any generated PVCs. Defaults to 'False'.
+- `openshift_logging_mux_file_buffer_pvc_pv_selector`: A key/value map added to a PVC
+  in order to select specific PVs.  Defaults to 'None'.
+- `openshift_logging_mux_file_buffer_pvc_prefix`: The prefix for the generated PVCs.
```
jcantrill: https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#examining-the-pods-ordinal-index

i think its some prefix + podname

roles/openshift_logging_mux/tasks/main.yaml
nhosoi: Ok.  It makes sense.  Let me move them to the common place (is openshift_logging ok?) to share...

nhosoi: Could there be any way to obtain the DC name or container name in the ansible task in the common place?  I'd like to replace logging-es-pvc.yml with {{ DCname }}-pvc.yml or logging-{{ containerName }}-pvc.yml here...
    src: pvc.j2
    dest: "{{ tempdir }}/templates/logging-es-pvc.yml"

jcantrill: @nhosoi I don't believe im making myself clear.  I'm not suggesting you move the task or template to a different role but to consolidate the two pvc tasks into one within this role.  Regarding the name I think you need to do something like 'set_facts' for the names in a prior step and then consume them in subsequent tasks

ewolinetz:
for value validation you could use something like we do in the component roles for verifying versions:
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/tasks/determine_version.yaml#L17-L19

paired with: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/vars/main.yml#L3

nhosoi:
Thank you @ewolinetz!  

openshift_logging_mux has exact the same pair 
  __allowed_mux_versions: ["3_5", "3_6"]

  - fail:
      msg: Invalid version specified for mux
    when: mux_version not in __allowed_mux_versions

Shall we replace "fail" with "assert"?

roles/openshift_logging/README.md
nhosoi: I borrowed this definition... :p
- `openshift_logging_es_pvc_prefix`: The prefix for the generated PVCs. Defaults to 'logging-es'.
It'd be "The prefix for the generated PVCs followed by podname" or something like that?

nhosoi:
Thank you for your reviews, @jcantrill and @ewolinetz.
I've pushed 2 patches following your comments.

roles/openshift_logging_elasticsearch/tasks/main.yaml
```diff
@@ -34,6 +34,33 @@
     mode: 0755
   changed_when: False
 
+- set_fact:
+    pvc_prefix: "{{ openshift_logging_es_pvc_prefix }}"
+
+- set_fact:
+    openshift_logging_pvc_name: "{{ openshift_logging_elasticsearch_pvc_name }}"
```
jcantrill: @ewolinetz Do we need to be careful here?  If this role is run the same time as the other logging roles, do we not run the risk of overriding dependent vars?  Or is that not a concern in context of running individual rols?

jcantrill:
[(?) Use `file` buffer instead of `memory` buffer for fluentd[logging][epic-ois-agl-perf]](https://trello.com/c/XpreI533/509-use-file-buffer-instead-of-memory-buffer-for-fluentdloggingepic-ois-agl-perf)

roles/openshift_logging_elasticsearch/tasks/main.yaml
ewolinetz: I'm not sure I'm following the concern. Are there other places where we use `openshift_logging_pvc_name` that could be overwritten?

roles/openshift_logging_elasticsearch/tasks/main.yaml
```diff
@@ -34,6 +34,33 @@
     mode: 0755
   changed_when: False
 
+- set_fact:
```
ewolinetz: All of these can be condensed into a single `set_fact` call --

```yaml
- set_fact:
     pvc_prefix: "{{ openshift_logging_es_pvc_prefix }}"
     openshift_logging_pvc_name: "{{ openshift_logging_elasticsearch_pvc_name }}"
     ...

**Comment by sosiouxme** [comment]
On Wed, Jul 12, 2017 at 10:01 AM, Jeff Cantrill <notifications@github.com>
wrote:
>
> @nhosoi <https://github.com/nhosoi> Generally it looks good to me but we
> should probably add some additional validation regarding the storage type
> and buffer type early in the execution of these tasks since they are
> limited to a set of values. Maybe @sosiouxme
> <https://github.com/sosiouxme> or @ewolinetz
> <https://github.com/ewolinetz> can point you to the appropriate place to
> add these checks.
>

We don't have a great approach to early validation right now. That's what
the `openshift_sanitize_inventory` role was intended for, however since
that is a dependency for just about everything you have to be really
careful that anything you put in there is going to be correct and
applicable for all playbooks and hosts, not just the ones your role would
normally be run with. So it's really only useful for very basic settings.

I'd suggest adding validation tasks like this early in the role where
they're used, preferably before you've started deploying anything.

File: roles/openshift_logging_mux/tasks/main.yaml [hunk 7]
```diff
@@ -25,6 +25,33 @@
     mode: 0755
   changed_when: False
 
+- set_fact:
```
ewolinetz: these can all be condensed into a single `set_fact` call

roles/openshift_logging/tasks/main.yaml
```diff
@@ -37,6 +37,55 @@
...
+# storageclasses with the storageClassName set to "" in pvc.j2
+- name: Creating storage template - static
+  template:
+    src: pvc.j2
+    dest: "{{ tempdir }}/templates/{{ pvc_prefix }}-pvc.yml"
+  vars:
+    obj_name: "{{ openshift_logging_pvc_name }}"
+    size: "{{ openshift_logging_pvc_size }}"
+    access_modes: "{{ openshift_logging_pvc_access_modes | list }}"
+    pv_selector: "{{ openshift_logging_pv_selector }}"
```
ewolinetz: are we not going to differentiate between the pvs created for ES and the ones created for fluentd?

roles/openshift_logging/tasks/main.yaml
```diff
@@ -37,6 +37,55 @@
...
+# Storageclasses are used by default if configured
+- name: Creating storage template - dynamic
+  template:
+    src: pvc.j2
+    dest: "{{ tempdir }}/templates/{{ pvc_prefix }}-pvc.yml"
+  vars:
+    obj_name: "{{ openshift_logging_pvc_name }}"
+    size: "{{ openshift_logging_pvc_size }}"
+    access_modes: "{{ openshift_logging_pvc_access_modes | list }}"
+    pv_selector: "{{ openshift_logging_pv_selector }}"
```
ewolinetz: same question regarding es vs fluentd selectors here

ewolinetz:
@nhosoi  I would prefer that we kept the logic for creating PVCs in the different component roles, this allows us to separate out the responsibilities and move towards the ability to only use these roles instead of the overarching `openshift_logging` role. Even if this means a duplication of logic, it keeps us from having interdependencies of roles.

nhosoi:
Thank you for your comments, @ewolinetz.
> I would prefer that we kept the logic for creating PVCs in the different component roles...
Ok. I can revert the last commit 70845ca5c721a5527456c454b06e8be7980d2028 (saving some small changes).  Is it ok with you, @jcantrill?  I guess it will eliminate your concern, which is good?
> If this role is run the same time as the other logging roles, do we not run the risk of overriding dependent vars?

Regarding the validation, I have no candidates to check in my mind...  @richm, do you have any?

nhosoi:
@jcantrill, @ewolinetz, sorry for my misunderstanding.  I reverted the patch "Merged elasticsearch and mux pvc tasks to one in openshift_logging/tasks."", then made a change for "Static and Dynamic Creating Mux storage templates into one".  Thanks.

roles/openshift_logging_mux/templates/mux.j2
```diff
@@ -66,6 +66,14 @@ spec:
         - name: muxcerts
           mountPath: /etc/fluent/muxkeys
           readOnly: true
+        - name: filebufferstorage
+{% if openshift_logging_mux_file_buffer_storage_type == 'pvc' %}
```
jcantrill: Editing my comment all together, there is no need to specify an if/else or even modify the mountPath. The mountPath can be fixed since this path never changes from the perspective of the container.

roles/openshift_logging_mux/templates/mux.j2
```diff
@@ -115,6 +123,18 @@ spec:
             resourceFieldRef:
               containerName: "mux"
               resource: limits.memory
+        - name: "BUFFER_TYPE"
+          value: "{{ openshift_logging_mux_buffer_type | lower }}"
+        - name: "FILE_BUFFER_PATH"
+{% if openshift_logging_mux_file_buffer_storage_type == 'pvc' %}
```
jcantrill: Here you can set the path to `openshift_logging_mux_file_buffer_mount_path` with no if/else since we either specify the value or provide a default.

roles/openshift_logging_mux/templates/mux.j2
nhosoi: openshift_logging_mux_file_buffer_mount_path is removed and the mount path "/var/log/mux/filebufferstorage" is fixed.  Thanks.

roles/openshift_logging/README.md
```diff
@@ -57,6 +57,8 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_fluentd_hosts`: List of nodes that should be labeled for Fluentd to be deployed to. Defaults to ['--all'].
 - `openshift_logging_fluentd_buffer_queue_limit`: Buffer queue limit for Fluentd. Defaults to 1024.
 - `openshift_logging_fluentd_buffer_size_limit`: Buffer chunk limit for Fluentd. Defaults to 1m.
+- `openshift_logging_fluentd_buffer_type`: Fluentd will set the buffer type.  Defaults to 'file'.
```
portante: Do we really have to add another parameter?  Why can't we just switch to `file` buffer and keep our solution simpler?

nhosoi: I thought there could be some customers who want to stick with memory buffer...  And for us, it'd be useful for performance comparison testing?

nhosoi:
Please advise.  I reran ansible-playbook with the same inventory file on top of the already deployed system, which should is supposed to do nothing by ansible's idempotency.
ansible-playbook -vvv -i ansible.inventory playbooks/byo/openshift-cluster/openshift-logging.yml
But it failed with this error in the mux storage.  @jcantrill, @ewolinetz, @richm, could you please tell me what is wrong?
```
TASK [openshift_logging_mux : Set Mux storage] ********************************************************************
***********************
task path: /tmp/tmp.csYZuRMjMK/openhift-ansible/roles/openshift_logging_mux/tasks/main.yaml:196
Using module file /tmp/tmp.csYZuRMjMK/openhift-ansible/roles/lib_openshift/library/oc_obj.py
<openshift> ESTABLISH LOCAL CONNECTION FOR USER: root
<openshift> EXEC /bin/sh -c '/usr/bin/python && sleep 0'
fatal: [openshift]: FAILED! => {
    "changed": false, 
    "failed": true, 
    "invocation": {
        "module_args": {
            "all_namespaces": null, 
            "content": null, 
            "debug": false, 
            "delete_after": false, 
            "files": [
                "/tmp/openshift-logging-ansible-peJFJj/templates/logging-mux-pvc.yml"
            ], 
            "force": false, 
            "kind": "pvc", 
            "kubeconfig": "/etc/origin/master/admin.kubeconfig", 
            "name": "logging-mux-pvc", 
            "namespace": "logging", 
            "selector": null, 
            "state": "present"
        }
    }
}

MSG:

{u'returncode': 1, u'cmd': u'/bin/oc replace -f /tmp/openshift-logging-ansible-peJFJj/templates/logging-mux-pvc.yml -n logging', u'results': {}, u'stderr': u'The PersistentVolumeClaim "logging-mux-pvc" is invalid: spec: Forbidden: field is immutable after creation\n', u'stdout': u''}

PLAY RECAP *******************************************************************************************************************************
localhost                  : ok=3    changed=0    unreachable=0    failed=0   
openshift                  : ok=197  changed=28   unreachable=0    failed=1   
```

portante:
@nhosoi, I don't think Ansible is by nature idempotent, but that the modules folks have written for Ansible are for the most part.  We are responsible for ensuring the procedures we write are idempotent.

roles/openshift_logging/README.md
richm: Having the ability to deploy with memory buffer, even if just for testing/PoC, is useful.  Let's keep this knob, and default it to file.

portante: @richm, if you keep the knob, that means it will be supported in the field, right?  Are we going to verify all combinations work?  Are there not way too many options for fluentd as it is that we don't test all combinations of today?

richm: @portante I want to keep the knob.  I don't care what we have to do, if we have to do it to keep this knob, I want to do it.

nhosoi:
Yes, I agree we are responsible.  I'm looking for the way how to fix in the mux role...
```
{u'returncode': 1, u'cmd': u'/bin/oc replace -f /tmp/openshift-logging-ansible-IhxG5E/templates/logging-mux-pvc.yml -n logging', u'results': {}, u'stderr': u'The PersistentVolumeClaim "logging-mux-pvc" is invalid: spec: Forbidden: field is immutable after creation\n', u'stdout': u''}
```
This error looks to me "/tmp/openshift-logging-ansible-IhxG5E/templates/logging-mux-pvc.yml" cannot be modified once it is done.  Is it because of the rule "delete_after: false"?  And I guess it's intentional?
Thanks.

ewolinetz:
I think what failed was the underlying `oc` command. Its likely that you're trying to rename a pvc that already exists. The file `/tmp/openshift-logging-ansible-IhxG5E/templates/logging-mux-pvc.yml` contains the new spec and it is trying to `oc apply` it. I would inspect that file spec and compare it to the pvc of the same name that already exists `oc get pvc/<name>`

richm:
@ewolinetz is there an `oc_volumes` or similar module we can use rather than constructing files and passing them to oc create/oc apply?

ewolinetz:
@richm unfortunately the `oc_volumes` module is a wrapper for the `oc volumes` command, which I don't think would help in this situation... 

Here are the current modules fwiw https://github.com/openshift/openshift-ansible/tree/master/roles/lib_openshift/library

richm:
I think we should try to use the oc_pvc module: https://github.com/openshift/openshift-ansible/blob/master/roles/lib_openshift/library/oc_pvc.py instead of roles/openshift_logging_mux/templates/pvc.j2, constructing a file using that template, and passing it to oc_obj - would that be more idempotent?

ewolinetz:
It should be as idempotent, however since its specific to pvcs it might be a better fit than creating a template and then passing it to oc_obj, so it may resolve the issues we're seeing. +1 to switching

jcantrill:
closing in favor of #4851

PR #4742, opened 2017-07-12
Remove dependency on playbook2image, rebase directly on OS.

juanvallejo:
Very much work in progress.

Attempts to remove any dependencies on playbook2image by only adding the functionality that we currently use from it directly to our repo.

Draft: https://docs.google.com/document/d/1AlvlI0dw_vGHD_Fx4psWxbAuCXIc7blJjT554bGfdUo
Trello card: not yet

**TODO**

- [x] Build from rhel7 (rhel7.3:7.3-released?) or (upstream) centos7
- [x] Update yum install to include ansible
- [x] Combine o-a + p2i labels and ENV
- [x] Remove ln -s lib hack
- [x] Copy /etc/passwd-munging logic (could we add comments for why it’s there…)
- [x] Copy .s2i run script to images/installer and make it the cmd
- [x] Modify to print usage instead of just exiting when there is no inventory/playbook specified
- [x] Test with installing logging and metrics on multi-host cluster; test check playbooks
- [x] Test that it does what we expect when run with runc
- [x] Test that it does what we expect when run with atomic install (currently being tested by Luke)
- [ ] Merge changes to dist-git aos3-installation-docker and build
- [ ] Have QE take a good look at it
- [ ] Probably lots of room for improvement re ssh key/config handling, inventory generation, output redirection, etc.

**Known issues**

 - Running `playbooks/byo/config.yml` deploys OpenShift in a remote host successfully but fails when deploying logging with: `Failed to get information on remote file (/tmp/openshift-logging-ansible-PERtiY/signing.conf): /bin/sh: sudo: command not found`
    - when `openshift_hosted_logging_deploy=true`

cc @sosiouxme @rhcarvalho @codificat @brenton


rhcarvalho:
Out of curiosity, why do we want to part ways with playbook2image?

images/installer/Dockerfile.rhel7
```diff
@@ -18,6 +18,8 @@ LABEL name="openshift3/ose-ansible" \
       version="v3.6.0" \
       release="1" \
       architecture="x86_64"
+      vcs-url="https://github.com/openshift/openshift-ansible" \
+      vcs-type="git" \
```
juanvallejo: cc @codificat

images/installer/Dockerfile.rhel7
```diff
@@ -18,6 +18,8 @@ LABEL name="openshift3/ose-ansible" \
...
 
 # Playbooks, roles and their dependencies are installed from packages.
 # Unlike in Dockerfile, we don't invoke the 'assemble' script here
```
juanvallejo: Done

images/installer/bin/run
```diff
@@ -0,0 +1,62 @@
...
+  else
+    cp ${INVENTORY_FILE} ${INVENTORY}
+  fi
+elif [[ -v INVENTORY_URL ]]; then
+  curl -o ${INVENTORY} ${INVENTORY_URL}
+elif [[ -v DYNAMIC_SCRIPT_URL ]]; then
+  curl -o ${INVENTORY} ${DYNAMIC_SCRIPT_URL}
+  chmod 755 ${INVENTORY}
+else
+  echo "One of INVENTORY_FILE, INVENTORY_URL or DYNAMIC_SCRIPT_URL must be provided"
```
juanvallejo: Sure, done

sosiouxme:
Most of what playbook2image installs is for operating as a builder, and we
don't need that capability. So we can simplify and drop the size a lot by
leaving it out. But more importantly, the downstream playbook2image is
based on the python SCL which operates with different paths and versions
from the system python packages and from the upstream pip-installed
modules. It seems prudent to keep upstream and downstream for
openshift-ansible as similar as possible - the differences have already
caused some problems. We can cut out the SCL and pip and keep things the
same.

images/installer/Dockerfile
```diff
@@ -12,23 +12,54 @@ LABEL name="openshift/origin-ansible" \
...
+# ansible and pip are in EPEL
+RUN yum install -y epel-release && yum clean all -y
+
+# workaround for https://github.com/openshift/openshift-ansible/issues/3111
+# install ansible via pip to lock version
+RUN yum install -y  --setopt=tsflags=nodocs python-pip python-devel && yum clean all -y
+RUN pip install -Iv ansible==2.2.0.0
+
+COPY ./images/installer/bin /usr/bin
+COPY ./images/installer/bin/user_setup /tmp
```
rhcarvalho: Since we are optimizing the image, we could reduce the number of layers by having a single `COPY` instruction, and placing all the files under `./images/installer/root`. Then, we can group together all `RUN` instructions into a single one.

images/installer/Dockerfile
```diff
@@ -12,23 +12,54 @@ LABEL name="openshift/origin-ansible" \
       io.k8s.display-name="openshift-ansible" \
       io.k8s.description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain and check an OpenShift cluster" \
       io.openshift.expose-services="" \
-      io.openshift.tags="openshift,install,upgrade,ansible"
+      io.openshift.tags="openshift,install,upgrade,ansible" \
+      vcs-url="https://github.com/openshift/openshift-ansible" \
+      vcs-type="git" \
+      version="alpha"
```
rhcarvalho: What is this label used for?

images/installer/Dockerfile
```diff
@@ -12,23 +12,54 @@ LABEL name="openshift/origin-ansible" \
...
 
+# Make home folder writable in order to let playbooks make changes
+RUN chmod a+rwx -R /opt/app-root/src
+
 RUN INSTALL_PKGS="skopeo openssl java-1.8.0-openjdk-headless httpd-tools" && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all
 
 USER ${USER_UID}
```
rhcarvalho: This was defined in p2i, now at this point it is undefined.

images/installer/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Using playbook2image as a base
 # See https://github.com/openshift/playbook2image for details on the image
 # including documentation for the settings/env vars referenced below
-FROM registry.centos.org/openshift/playbook2image:latest
+FROM openshift/base-centos7
```
sosiouxme: uh... why this one? This brings in a [bunch of dev stuff](https://hub.docker.com/r/openshift/base-centos7/) we don't need. Why not just `centos7`?

images/installer/Dockerfile
```diff
@@ -12,23 +12,54 @@ LABEL name="openshift/origin-ansible" \
...
+      vcs-type="git" \
+      version="alpha"
 
 USER root
 
+# ansible and pip are in EPEL
+RUN yum install -y epel-release && yum clean all -y
+
+# workaround for https://github.com/openshift/openshift-ansible/issues/3111
+# install ansible via pip to lock version
```
sosiouxme: this is resolved with later ansible, should be able to use latest ansible from EPEL or origin repos

images/installer/Dockerfile
```diff
@@ -12,23 +12,54 @@ LABEL name="openshift/origin-ansible" \
...
+RUN yum install -y  --setopt=tsflags=nodocs python-pip python-devel && yum clean all -y
+RUN pip install -Iv ansible==2.2.0.0
```
sosiouxme: so we should just be able to yum install ansible

images/installer/Dockerfile
```diff
@@ -40,12 +71,13 @@ ENV PLAYBOOK_FILE=playbooks/byo/openshift_facts.yml \
 # /tmp/src (as per the source-to-image specs) so we import it there
 ADD . /tmp/src
 
-# Running the 'assemble' script provided by playbook2image will install
+# Running the 'assemble' script will install
 # dependencies specified in requirements.txt and install the 'oc' client
 # as per the INSTALL_OC environment setting above
-RUN /usr/libexec/s2i/assemble
+RUN /usr/bin/assemble
```
sosiouxme: we don't need any of this assemble stuff

I'm wondering what we need the `oc` client for either but I guess we should include that in case something is actually using it.

we don't need the `INSTALL_OC` variable if we just hardcode that bit from `assemble`

images/installer/bin/assemble
```diff
@@ -0,0 +1,48 @@
...
+elif [[ -e requirements.txt ]]; then
+  pip install --no-cache-dir --user -r requirements.txt
+fi
+
+if [[ -v INSTALL_OC ]]; then
+  pip install --no-cache-dir --user requests
+  echo "---> Installing 'oc' binary..."
+  TMPDIR=$(mktemp -d)
+  cd ${TMPDIR}
+  OC_BINARY_URL=$(python -c "import requests;releases = requests.get('https://api.github.com/repos/openshift/origin/releases').json();print [s for s in [r for r in releases if not r['prerelease'] and '1.4' in r['name']][0]['assets'] if 'linux-64' in s['browser_download_url']][0]['browser_download_url']")
```
sosiouxme: gets latest origin oc... origin-only of course. err, it's a little unclear to me what this actually ends up with... 1.4 releases?

downstream, RPM deps will pull in oc

images/installer/bin/run
```diff
@@ -0,0 +1,49 @@
...
+  exec /usr/bin/usage
+  exit 1
```
sosiouxme: `exec` doesn't return

images/installer/Dockerfile.rhel7
```diff
@@ -18,26 +18,42 @@ LABEL name="openshift3/ose-ansible" \
       version="v3.6.0" \
       release="1" \
       architecture="x86_64"
```
sosiouxme: needs a `\` at the end

images/installer/Dockerfile.rhel7
```diff
@@ -18,26 +18,42 @@ LABEL name="openshift3/ose-ansible" \
...
 # Playbooks, roles and their dependencies are installed from packages.
 # Unlike in Dockerfile, we don't invoke the 'assemble' script here
-# because all content and dependencies (like 'oc') is already
-# installed via yum.
 USER root
 RUN INSTALL_PKGS="atomic-openshift-utils atomic-openshift-clients python-boto skopeo openssl java-1.8.0-openjdk-headless httpd-tools" && \
     yum repolist > /dev/null && \
     yum-config-manager --enable rhel-7-server-ose-3.6-rpms && \
     yum-config-manager --enable rhel-7-server-rh-common-rpms && \
+    yum install -y --setopt=tsflags=nodocs ansible
```
sosiouxme: needs `&& \`

images/installer/Dockerfile.rhel7
sosiouxme: `      atomic.run="once" \`

Also the last line shouldn't have a `\`

images/installer/Dockerfile
juanvallejo: https://docs.docker.com/engine/reference/builder/#label just part of image metadata; however, I seem to have kept the default used by p2i. Not sure what the value should be here; `1.6.0`?

images/installer/Dockerfile
juanvallejo: Was not sure which one to use but for some reason, `openshift/base-centos7` seemed more inviting. Will switch to just `centos7`

images/installer/Dockerfile
rhcarvalho: Sure, some of those labels are used for example in the web console. It made sense for p2i since it was meant to be used as a builder image (thus the `io.k8s.*` and `io.openshift.*` labels). Some other labels are used in build systems.

If we have no use for a label, perhaps it is a good time to get rid of it instead of copying from p2i?

A `version` label seems to be only one more thing that needs to be kept up-to-date.

rhcarvalho: (If we really need a `version` label, we could have our image build script set the right version on build)

images/installer/Dockerfile
juanvallejo: sounds good, should we still install python / worry about having any other deps that `Dockerfile.rhel7` installs as well?

images/installer/bin/assemble
juanvallejo: sounds good, will remove

images/installer/Dockerfile.rhel7
juanvallejo: Looking at `INSTALL_PKS`, do we still need some of them? (looking at java in particular)

rhcarvalho: I think Java was added to be able to run the logging installer? Not sure. It was added relatively recently, we can check the history.

images/installer/Dockerfile
sosiouxme: nope, don't see any need for them... however we are probably going to need to add RPMs for role dependencies that were pip installed upstream and RPM dependencies downstream.

images/installer/Dockerfile
```diff
@@ -1,51 +1,58 @@
...
+    HOME=/opt/app-root/src \
+    PATH=$PATH:/opt/app-root/bin \
+#
+# The playbook to be run is specified via the PLAYBOOK_FILE env var.
+# This sets a default of openshift_facts.yml as it's an informative playbook
+# that can help test that everything is set properly (inventory, sshkeys)
+    PLAYBOOK_FILE=playbooks/byo/openshift_facts.yml \
+    OPTS="-v"
+
+COPY ./images/installer/bin /usr/bin
```
codificat: I'm being picky, but this copies a README file into `/usr/bin`... I'd suggest moving the `README.md` inside `images/installer/bin` one level up and expand it a bit (see below)

images/installer/bin/README.md
```diff
@@ -0,0 +1,5 @@
+Utils for use during / after installation of image
+==================================================
+
+Contains image entrypoint as well as any other scripts useful
+in aiding an image run.
```
codificat: I believe this file can be moved up one level and briefly document the contents, pointing to other readmes for details... it should make it easier to find and also avoid it ending up in `/usr/bin`

images/installer/Dockerfile
```diff
@@ -1,51 +1,58 @@
...
+# files for running as a system container
+COPY ./images/installer/system-container/root /
+
+RUN mkdir -p /opt/app-root/{etc,bin,src} \
+ && chmod -R ug+x /opt/app-root/{bin,etc} /usr/bin/user_setup \
+#
+# Make home folder writable in order to let playbooks make changes
+ && chmod a+rwx -R /opt/app-root/src \
```
codificat: I'm failing to see why we need this...

images/installer/Dockerfile.rhel7
```diff
@@ -10,46 +10,50 @@ LABEL name="openshift3/ose-ansible" \
       summary="$SUMMARY" \
       description="$DESCRIPTION" \
       url="https://github.com/openshift/openshift-ansible" \
-      io.k8s.display-name="openshift-ansible" \
-      io.k8s.description="$DESCRIPTION" \
-      io.openshift.expose-services="" \
-      io.openshift.tags="openshift,install,upgrade,ansible" \
```
codificat: Ditto about these labels.. they are certainly not required, but I believe it doesn't hurt to have them in this image

images/installer/Dockerfile.rhel7
```diff
@@ -10,46 +10,50 @@ LABEL name="openshift3/ose-ansible" \
...
       com.redhat.component="aos3-installation-docker" \
-      version="v3.6.0" \
```
codificat: This one is actually [required by the build system](https://mojo.redhat.com/docs/DOC-1023066#jive_content_id_Labels)

images/installer/Dockerfile.rhel7
```diff
@@ -10,46 +10,50 @@ LABEL name="openshift3/ose-ansible" \
...
 # Playbooks, roles and their dependencies are installed from packages.
-# Unlike in Dockerfile, we don't invoke the 'assemble' script here
-# because all content and dependencies (like 'oc') is already
-# installed via yum.
 USER root
 RUN INSTALL_PKGS="atomic-openshift-utils atomic-openshift-clients python-boto skopeo openssl java-1.8.0-openjdk-headless httpd-tools" && \
     yum repolist > /dev/null && \
     yum-config-manager --enable rhel-7-server-ose-3.6-rpms && \
     yum-config-manager --enable rhel-7-server-rh-common-rpms && \
+    yum install -y --setopt=tsflags=nodocs ansible && \
```
codificat: I believe `ansible` gets pulled as a dependency - but if we want to install explicitly maybe just add it to `INSTALL_PKGS` above...

images/installer/Dockerfile.rhel7
```diff
@@ -10,46 +10,50 @@ LABEL name="openshift3/ose-ansible" \
...
     yum clean all
 
-# The symlinks below are a (hopefully temporary) hack to work around the fact that this
-# image is based on python s2i which uses the python27 SCL instead of system python,
-# and so the system python modules we need would otherwise not be in the path.
-RUN ln -s /usr/lib/python2.7/site-packages/{boto,passlib} /opt/app-root/lib64/python2.7/
 
+RUN mkdir -p ${APP_HOME} ${APP_ROOT}/etc ${APP_ROOT}/bin
+RUN chmod -R ug+x ${APP_ROOT}/bin ${APP_ROOT}/etc /tmp/user_setup && \
+    /tmp/user_setup
```
codificat: These can be consolidated in a single `RUN`.
More importantly, they reference stuff that is defined in `ENV` and `COPY` statements that are below - order should be fixed

images/installer/Dockerfile.rhel7
```diff
@@ -10,46 +10,50 @@ LABEL name="openshift3/ose-ansible" \
...
     ANSIBLE_CONFIG=/usr/share/atomic-openshift-utils/ansible.cfg \
     WORK_DIR=/usr/share/ansible/openshift-ansible \
     OPTS="-v"
 
-# Add files for running as a system container
-COPY system-container/root /
+WORKDIR ${APP_HOME}
 
-CMD [ "/usr/libexec/s2i/run" ]
+RUN sed "s@${USER_NAME}:x:${USER_UID}:0@${USER_NAME}:x:\${USER_ID}:\${GROUP_ID}@g" /etc/passwd > ${APP_ROOT}/etc/passwd.template
```
codificat: This `RUN` is now replaced by the entrypoint, which isn't defined... so this line should be replaced with `ENTRYPOINT ["/usr/bin/entrypoint"]`

images/installer/Dockerfile
```diff
@@ -1,51 +1,58 @@
...
 
 USER root
 
+# install ansible and deps
+RUN yum install -y epel-release \
+ && yum-config-manager --disable epel \
+ && yum install -y --setopt=tsflags=nodocs --enablerepo=epel ansible \
+ && INSTALL_PKGS="python-lxml pyOpenSSL python2-cryptography skopeo openssl java-1.8.0-openjdk-headless httpd-tools" \
+ && yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS \
+ && rpm -V $INSTALL_PKGS \
```
sosiouxme: I have seen too many instances of doing `yum install X Y Z` and having Z not available for some reason but yum still installs everything else and reports success. This fails if you don't install everything you wanted.

images/installer/Dockerfile.rhel7
sosiouxme: +1 it does.

images/installer/Dockerfile
sosiouxme: although `rpm -q` should work just as well...

images/installer/Dockerfile
juanvallejo: At some point I was getting an error because `.ansible` could not be written to `/opt/app-root/src`. Will test again though

images/installer/Dockerfile
sosiouxme: Case in point! `skopeo` comes from extras and wasn't being installed. And we don't actually need it.

rhcarvalho: What @sosiouxme said. We use `rpm -V` in many of the images we ship with OpenShift, e.g.:

https://github.com/sclorg/postgresql-container/blob/master/9.5/Dockerfile#L44

And yes, it saves time debugging why an image is missing some package we think was installed.

images/installer/Dockerfile
rhcarvalho: Also leveraging the practice of writing Dockerfiles... what @codificat says + make a single top-level dir `root` and place files mirroring what we want to see in the image. Things that need not / should not be in the image should be outside of `root`. E.g.:

https://github.com/sclorg/postgresql-container/blob/master/9.5/Dockerfile#L56

images/installer/Dockerfile
```diff
@@ -1,51 +1,44 @@
...
       url="https://github.com/openshift/openshift-ansible" \
       io.k8s.display-name="openshift-ansible" \
       io.k8s.description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain and check an OpenShift cluster" \
       io.openshift.expose-services="" \
-      io.openshift.tags="openshift,install,upgrade,ansible"
+      io.openshift.tags="openshift,install,upgrade,ansible" \
+      atomic.run="once"
 
-USER root
+ENV USER_UID=1000 \
```
codificat: In `Dockerfile.rhel7` it says 1001 instead - let's pick one and make it the same in both

images/installer/Dockerfile
```diff
@@ -1,51 +1,44 @@
-# Using playbook2image as a base
-# See https://github.com/openshift/playbook2image for details on the image
-# including documentation for the settings/env vars referenced below
-FROM registry.centos.org/openshift/playbook2image:latest
+FROM centos:7
 
 MAINTAINER OpenShift Team <dev@lists.openshift.redhat.com>
 
+USER root
+
```
codificat: Is there any reason to move the `USER` and `RUN` commands before the `LABEL`s? If not, wouldn't it be better to leave the `LABEL`s first to keep the metadata type of content at the top? The caching related benefits of that are probably not relevant in our case, but still wondering if it would make sense from a file organization POV?

images/installer/Dockerfile
```diff
@@ -1,51 +1,44 @@
...
-# as per the INSTALL_OC environment setting above
-RUN /usr/libexec/s2i/assemble
-
-# Add files for running as a system container
-COPY images/installer/system-container/root /
-
-CMD [ "/usr/libexec/s2i/run" ]
+WORKDIR ${WORK_DIR}
+ENTRYPOINT [ "/usr/local/bin/entrypoint" ]
+CMD [ "/usr/local/bin/run" ]
```
codificat: Now that this has been re-arranged I think it would make sense to complete the work by moving the contents under `system-container/root` into `root` so there's only one `root`.

The `README.md` inside `system-container` could be moved higher up as `README_SYSTEM_CONTAINER.md`, or its contents integrated into `README_CONTAINER_IMAGE.md`...

images/installer/Dockerfile.rhel7
```diff
@@ -1,55 +1,48 @@
...
+# Playbooks, roles, and their dependencies are installed from packages.
+RUN INSTALL_PKGS="atomic-openshift-utils atomic-openshift-clients python-boto openssl java-1.8.0-openjdk-headless httpd-tools" \
+ && yum repolist > /dev/null \
+ && yum-config-manager --enable rhel-7-server-ose-3.6-rpms \
+ && yum-config-manager --enable rhel-7-server-rh-common-rpms \
+ && yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS \
+ && rpm -q $INSTALL_PKGS \
+ && yum clean all
 
 LABEL name="openshift3/ose-ansible" \
```
codificat: Same question as with `Dockerfile`: should `LABEL` come before `USER`/`RUN`?

images/installer/Dockerfile
sosiouxme: It's really just that when iterating on changes, it helps to put the slowest thing (the install) first so it can be most likely to be cached. Not a huge deal to me but I don't like having to re-run the install just because I tweaked the labels. Labels are still pretty close to the top... but we can move it if you like.

images/installer/Dockerfile
juanvallejo: Will change to `1001` to match .rhel7. Changed it here to `1000` to test while I finished adding the entrypoint script that patched `/etc/passwd` with a `1000` user entry

images/installer/Dockerfile
codificat: Heh, in the "normal" use case it's usually the other way around: when you build the same image over time, labels change less than the result of 'yum install' - so having the labels before the run should help caching.

However, this doesn't really apply to our build system (I think), so it was more of a comment around the logical structure of the file. To me it looks better to have the metadata first, but up to you.

images/installer/Dockerfile
juanvallejo: Will go ahead and add a commit for this

juanvallejo: Changed in https://github.com/openshift/openshift-ansible/pull/4742/commits/e02bc5dc511ee679917f1014304e9331a91a8429

sosiouxme:
As best I can tell, everything is working. I have never seen the `/bin/sh: sudo: command not found` issue when deploying logging (and metrics). I also don't have or need the `localhost` line in my inventory. But it is disquieting that Juan does and makes me feel like users will manage to run into it. But is that something new?

/cc @ashcrow @giuseppe As best I can tell, the `atomic install --system` container works fine with both origin-ansible and ose-ansible after these changes. Would appreciate your input.

(We don't have the labels for atomic install to operate on in a non-system-container context - something to introduce in the near future hopefully as behavior is very surprising without the `--system`.)

images/installer/Dockerfile
sosiouxme: Our build system doesn't use cache so as far as I'm concerned the only optimizations to consider are for iterative local development. So I vote to leave it as is unless I hear loud complaining :)

images/installer/Dockerfile
```diff
@@ -1,51 +1,45 @@
...
-# This is required since the system-container uses by default the playbook under
-# /usr/share/ansible/openshift-ansible.  With this change we won't need to keep two different
-# configurations for the two images.
-RUN mkdir -p /usr/share/ansible/ && ln -s /opt/app-root/src /usr/share/ansible/openshift-ansible
+# Add entrypoint and other setup scripts
+COPY images/installer/root /
+# Add files for running as a system container
+COPY images/installer/system-container/root /
+# Include playbooks, roles, plugins, etc. from this repo
+COPY . ${WORK_DIR}
```
codificat: It might be a good opportunity to review `.dockerignore`... e.g. should we add `images` and `hack? anything else?

rhcarvalho: Why don't we copy only `COPY root /`? Or is this supposed to copy openshift-ansible top-level contents?

sosiouxme: It's suppose to copy top-level contents - playbooks, roles, plugins, etc. These things are all installed by RPM downstream. We could certainly .ignore a lot of things.

sosiouxme: @codificat we actually need `images/` in origin-ansible :) and ose-ansible is already sparse. We can omit `hack/`, `inventory/` and most of the top-level conf files. Do the README.md files belong in the image? Not sure how useful they are but they do little harm...

sosiouxme: Oh actually... .md files already omitted.

sosiouxme:
I think it's ready once we squash commits.

images/installer/Dockerfile
```diff
@@ -1,51 +1,45 @@
...
 
-# Create a symlink to /opt/app-root/src so that files under /usr/share/ansible are accessible.
-# This is required since the system-container uses by default the playbook under
-# /usr/share/ansible/openshift-ansible.  With this change we won't need to keep two different
-# configurations for the two images.
-RUN mkdir -p /usr/share/ansible/ && ln -s /opt/app-root/src /usr/share/ansible/openshift-ansible
+# Add entrypoint and other setup scripts
+COPY images/installer/root /
+# Add files for running as a system container
+COPY images/installer/system-container/root /
```
juanvallejo: Ah meant to remove this but never pushed the change :/

images/installer/Dockerfile
```diff
@@ -1,51 +1,43 @@
...
 LABEL name="openshift/origin-ansible" \
       summary="OpenShift's installation and configuration tool" \
       description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain and check an OpenShift cluster" \
       url="https://github.com/openshift/openshift-ansible" \
       io.k8s.display-name="openshift-ansible" \
       io.k8s.description="A containerized openshift-ansible image to let you run playbooks to install, upgrade, maintain and check an OpenShift cluster" \
       io.openshift.expose-services="" \
-      io.openshift.tags="openshift,install,upgrade,ansible"
+      io.openshift.tags="openshift,install,upgrade,ansible" \
+      atomic.run="once"
```
ashcrow: :+1:

images/installer/Dockerfile.rhel7
```diff
@@ -1,55 +1,46 @@
...
+      architecture="x86_64" \
+      atomic.run="once"
 
-# The playbook to be run is specified via the PLAYBOOK_FILE env var.
-# This sets a default of openshift_facts.yml as it's an informative playbook
-# that can help test that everything is set properly (inventory, sshkeys).
-# As the playbooks are installed via packages instead of being copied to
-# $APP_HOME by the 'assemble' script, we set the WORK_DIR env var to the
-# location of openshift-ansible.
-ENV PLAYBOOK_FILE=playbooks/byo/openshift_facts.yml \
```
ashcrow: Just curious, why was ``PLAYBOOK_FILE`` removed while other env items were kept?

images/installer/Dockerfile
```diff
@@ -1,51 +1,43 @@
...
-    yum clean all
+RUN /usr/local/bin/user_setup \
+ && rm /usr/local/bin/usage.ocp
 
 USER ${USER_UID}
 
-# The playbook to be run is specified via the PLAYBOOK_FILE env var.
-# This sets a default of openshift_facts.yml as it's an informative playbook
-# that can help test that everything is set properly (inventory, sshkeys)
-ENV PLAYBOOK_FILE=playbooks/byo/openshift_facts.yml \
```
ashcrow: Just curious, why was ``PLAYBOOK_FILE`` removed while other env items were kept?

ashcrow:
It seems like reasonable changes to me. Nothing jumps out at me as being problematic but I suggest letting QE know they should re-test the container and system containers just to be sure.

@giuseppe WDYT?

images/installer/Dockerfile.rhel7
sosiouxme: I want the image to do something helpful if you just run it without reading the docs. To me the most helpful thing seems to print a usage message immediately and exit. If we specify PLAYBOOK_FILE then we're going to run a playbook, and running a "usage" playbook seemed like overkill when we can just... run `usage`.

sosiouxme:
We absolutely want QE to scrutinize the major use cases once we merge and rebuild.

https://trello.com/c/BHJYd3Y5/135-drop-playbook2image-dependency

README_CONTAINER_IMAGE.md
```diff
@@ -1,6 +1,6 @@
 # Containerized openshift-ansible to run playbooks
 
-The [Dockerfile](images/installer/Dockerfile) in this repository uses the [playbook2image](https://github.com/openshift/playbook2image) source-to-image base image to containerize `openshift-ansible`. The resulting image can run any of the provided playbooks. See [BUILD.md](BUILD.md) for image build instructions.
+The [Dockerfile](images/installer/Dockerfile) in this repository uses the [Centos:7](https://hub.docker.com/_/centos/) base image to containerize `openshift-ansible`. The resulting image can run any of the provided playbooks. See [BUILD.md](BUILD.md) for image build instructions.
```
juanvallejo: @sosiouxme @codificat maybe we should just take the line about the `base image` out altogether?

codificat: Agreed that it can be removed

BUILD.md
```diff
@@ -37,9 +37,9 @@ To build a container image of `openshift-ansible` using standalone **Docker**:
 
 To build an openshift-ansible image using an **OpenShift** [build and image stream](https://docs.openshift.org/latest/architecture/core_concepts/builds_and_image_streams.html) the straightforward command would be:
 
-        oc new-build registry.centos.org/openshift/playbook2image~https://github.com/openshift/openshift-ansible
+        oc new-build openshift/origin-ansible~https://github.com/openshift/openshift-ansible
```
juanvallejo: @sosiouxme @codificat wasn't sure if this was the right thing to put here... Would this command actually work?  

Also, per the docs in the lines bellow, technically we are no longer "required" to keep the `images/installer` structure

codificat: that doesn't look right... `openshift/origin-ansible` is not a builder image...

I think it's not that easy to build the image in OpenShift anymore... the Dockerfile is under images/ but the context is the tld... 

Maybe we can just leave out the whole *Building on OpenShift* section for now?

juanvallejo: > Maybe we can just leave out the whole Building on OpenShift section for now?

Okay, I think that might be the best approach while we figure this out

cc @brenton

brenton: works for me.

sosiouxme: I'm sure there's a way to invoke the docker builder but I don't particularly want to try and test it right now, so save that for later.

sosiouxme: We could move Dockerfile back to root ... but we would need to keep `images/installer` for the downstream build anyway. Not inclined to change this just now.

sosiouxme:
aos-ci-test

sosiouxme:
Errr... test flake? Doesn't look related.

sosiouxme:
aos-ci-test

sosiouxme:
[merge] :tada:

sosiouxme:
[flake](https://github.com/openshift/origin/issues/15356) although it looks rather too persistent to be called that. @sdodson should I try the merge again?

sosiouxme:
[merge] again just for protocol

rhcarvalho:
Flake https://github.com/openshift/origin/issues/15332 and https://github.com/openshift/origin/issues/13977

[merge]

sosiouxme:
Just https://github.com/openshift/origin/issues/15356 again... @sdodson can you merge please?

rhcarvalho:
For future reference, do we have the size of the image BEFORE and AFTER this change? @juanvallejo @sosiouxme

rhcarvalho:
Also for our future selves, could you please update the PR description to reflect the current state? Are there still known issues, are the checkboxes in the TODO reasonable?

sosiouxme:
In my testing, ose-ansible's size before was about 1.1G and after, 760M. origin-ansible was 880M and is now 454M. Now that I look at it I'm not sure why the difference from origin to ose is so much... should have nearly the same dependencies, and base image size is about the same.

sosiouxme:
Hm, I should update the origin-ansible job to tag with v3.7 when building from master.

sosiouxme:
There is no 3.7 branch in dist-git aos3-installation-docker yet. I am curious to see if the build tomorrow morning pulls in changes from master or has been re-pointed at release-3.6.

PR #4743, opened 2017-07-12
Created js file for enabling tech preview for console, updated master…

ewolinetz:
…-config for pod presets and console tech preview

Should address https://bugzilla.redhat.com/show_bug.cgi?id=1469986

### DIFF
--- a/roles/openshift_service_catalog/tasks/wire_aggregator.yml
+++ b/roles/openshift_service_catalog/tasks/wire_aggregator.yml
@@ -119,6 +119,11 @@
   when:
   - not front_proxy_kubeconfig.stat.exists
 
+- name: copy tech preview extension file for service console UI
+  copy:
+    src: openshift-ansible-catalog-console.js
+    dest: /etc/origin/master/openshift-ansible-catalog-console.js
+
 - name: Update master config
   yedit:
     state: present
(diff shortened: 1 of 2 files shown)


roles/openshift_service_catalog/tasks/wire_aggregator.yml
```diff
@@ -138,6 +143,10 @@
       value: [X-Remote-Group]
     - key: authConfig.requestHeader.extraHeaderPrefixes
       value: [X-Remote-Extra-]
+    - key: assetConfig.extensionScripts
+      value: [/etc/origin/master/openshift-ansible-catalog-console.js]
+    - key: kubernetesMasterConfig.apiServerArguments.runtime-config
+      value: [apis/settings.k8s.io/v1alpha1=true]
```
sdodson: Does this [] notation ensure that it's only an additive thing or would this overwrite whatever they have now? If it's overwriting I'm fine with this for now but we'll want to make sure it's additive as a followup.

ewolinetz: its to ensure that its a list, if it isn't added as a list item the master will fail to start up

ewolinetz: Im not entirely sure, but i believe it will overwrite currently

sdodson: @kwoodson is there an easy way to ensure yedit is additive only?

spadgett:
openshift-ansible-catalog-console.js LGTM

PR #4746, opened 2017-07-12
Adds null nodeselector when creating logging project

thoraxe:
https://bugzilla.redhat.com/show_bug.cgi?id=1470394

For whatever reason, when the logging project doesn't already exist and you are running the logging installation against an existing cluster where there is a default node selector for projects, the logging install blows up.

This change adds a null nodeselector to the project creation line.

I'm not sure there's a "better" implementation here.

### DIFF
--- a/roles/openshift_hosted_logging/tasks/deploy_logging.yaml
+++ b/roles/openshift_hosted_logging/tasks/deploy_logging.yaml
@@ -25,7 +25,7 @@
 
 - name: "Create logging project"
   command: >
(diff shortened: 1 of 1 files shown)


ewolinetz:
@thoraxe looking at the bz, that is open against 3.5 correct? this fix should be based on the release-1.5 branch (I also believe it would be contained within the `openshift_logging` role and not `openshift_hosted_logging` role)

thoraxe:
@ewolinetz The BZ is filed against 3.5 but this affects any OpenShift environment -- I don't believe nodeselector logic has changed, as it's core Kube stuff.

Are you saying that you want me to check out release-1.5 and then make my fix to that and PR back to release-1.5?

PR #4747, opened 2017-07-12
Adds null nodeselector when creating logging project

thoraxe:


ewolinetz:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1470394

sdodson:
We don't need this on master too?

ewolinetz:
No we don't, we are using a module on master and default it to null if it's
not specified.

On Jul 12, 2017 7:27 PM, "Scott Dodson" <notifications@github.com> wrote:

We don't need this on master too?

—
You are receiving this because you commented.

<https://github.com/openshift/openshift-ansible/pull/4747#issuecomment-314934969>,
or mute the thread

PR #4751, opened 2017-07-13
Redeploy-certificates will fail for registry and router if user is not system:admin

jkaurredhat:
Redeploy-certificates will fail for registry and router if user is not system:admin

### DIFF
--- a/playbooks/common/openshift-cluster/redeploy-certificates/router.yml
+++ b/playbooks/common/openshift-cluster/redeploy-certificates/router.yml
@@ -116,6 +116,7 @@
         tls.crt="{{ mktemp.stdout }}/openshift-hosted-router-certificate.pem"
         tls.key="{{ mktemp.stdout }}/openshift-hosted-router-certificate.key"
         --type=kubernetes.io/tls
+        --config={{ mktemp.stdout }}/admin.kubeconfig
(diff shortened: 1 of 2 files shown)


playbooks/common/openshift-cluster/redeploy-certificates/router.yml
```diff
@@ -116,6 +116,7 @@
         tls.crt="{{ mktemp.stdout }}/openshift-hosted-router-certificate.pem"
         tls.key="{{ mktemp.stdout }}/openshift-hosted-router-certificate.key"
         --type=kubernetes.io/tls
+        --config={{ mktemp.stdout }}/admin.kubeconfig
         --confirm
         -o json | {{ openshift.common.client_binary }} replace -f -
```
abutcher: I think we'll also need to pass the config after the pipe to `oc replace`.

sdodson:
Lets make sure we check all instances for this so we're relatively sure that we've addressed this once and for all.

jkaurredhat:
Thanks for the review. I actually missed it. 

I have made the required changes now.

sdodson:
aos-ci-test

sdodson:
[merge][severity:blocker]

PR #4755, opened 2017-07-13
set KillMode to process in node service file

ingvagabund:
Bug: 1423640

### DIFF
--- a/roles/openshift_node/templates/node.service.j2
+++ b/roles/openshift_node/templates/node.service.j2
@@ -25,6 +25,7 @@ SyslogIdentifier={{ openshift.common.service_type }}-node
 Restart=always
 RestartSec=5s
 OOMScoreAdjust=-999
+KillMode=process
 
 [Install]
 WantedBy=multi-user.target
--- a/roles/openshift_node_upgrade/templates/node.service.j2
+++ b/roles/openshift_node_upgrade/templates/node.service.j2
@@ -25,6 +25,7 @@ SyslogIdentifier={{ openshift.common.service_type }}-node
 Restart=always
 RestartSec=5s
 OOMScoreAdjust=-999
+KillMode=process
 
 [Install]
 WantedBy=multi-user.target


ingvagabund:
aos-ci-test

sdodson:
CI jobs are still broken on image builds, merging.

ingvagabund:
@sdodson backport to 3.5 and/or 3.4?

sdodson:
yeah, sorry i should've added those labels when reviewing.

sjenning:
This has exposed https://bugzilla.redhat.com/show_bug.cgi?id=1473531

derekwaynecarr:
I do not think this is a safe change given the other problems it exposes.

I would like this reverted.

@ingvagabund @sdodson

sdodson:
@jsafrane can you comment on this?

derekwaynecarr:
i am trying to balance the number of workloads this helps versus the number it hurts.  right now, it appears to me from my reading of the linked issues its harm is greater than its good.  we need to fix the issues it exposed before taking this fix.

sjenning:
@sdodson I agree with @derekwaynecarr.  I'll try to get a fix in for cadvisor, but that fix should really exist _before_ this fix.  And long term, we do not want the fuse processes in the same systemd service with the node process.  Setting `KillMode=process` effectively makes these processes unmanageable by systemd, which is no good.

sjenning:
Also, I was talking to @jsafrane before he signed off and he agrees that the whole thing is hacky.  We talked about using systemd-run + `--experimental-mounter-path` to start the fuse processes in their own service.  That would cause service sprawl but would make each one a child of systemd.  Not sure what that would mean for selinux policy.  Having two different fixes for RPM vs containerized isn't ideal either.

sdodson:
Ok, lets revert.

sjenning:
discussion upstream on this topic for future reference
https://github.com/kubernetes/kubernetes/issues/34965

jsafrane:
I'm fixing mounter in Kubernetes to use systemd-run even when running on directly on the host. PR will follow shortly.

jsafrane:
Kubernetes PR: https://github.com/kubernetes/kubernetes/pull/49640

PR #4757, opened 2017-07-13
openshift_checks/docker_storage: overlay/2 support

sosiouxme:
fix https://bugzilla.redhat.com/show_bug.cgi?id=1469197

When Docker is configured with the overlay or overlay2 storage driver, check that it is supported and usage is below threshold.

Also, document and fix `openshift_docker_selinux_enabled` option so that running the check need not result in the `docker` role reconfiguring and restarting docker with `--selinux-enabled` (which will prevent docker from running with these drivers at all).

### DIFF
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -113,6 +113,11 @@ openshift_release=v3.6
 # Downgrades are not supported and will error out. Be careful when upgrading docker from < 1.10 to > 1.10.
 # docker_version="1.12.1"
 
+# Specify whether to run Docker daemon with SELinux enabled in containers. Default is True.
+# Uncomment below to disable; for example if your kernel does not support the
+# Docker overlay/overlay2 storage drivers with SELinux enabled.
+#openshift_docker_selinux_enabled=False
+
 # Skip upgrading Docker during an OpenShift upgrade, leaves the current Docker version alone.
 # docker_upgrade=False
 
--- a/roles/docker/tasks/package_docker.yml
+++ b/roles/docker/tasks/package_docker.yml
@@ -93,7 +93,7 @@
     dest: /etc/sysconfig/docker
     regexp: '^OPTIONS=.*$'
     line: "OPTIONS='\
-      {% if ansible_selinux.status | default(None) == '''enabled''' and docker_selinux_enabled | default(true) %} --selinux-enabled {% endif %}\
+      {% if ansible_selinux.status | default(None) == 'enabled' and docker_selinux_enabled | default(true) | bool %} --selinux-enabled {% endif %}\
       {% if docker_log_driver is defined  %} --log-driver {{ docker_log_driver }}{% endif %}\
       {% if docker_log_options is defined %} {{ docker_log_options |  oo_split() | oo_prepend_strings_in_list('--log-opt ') | join(' ')}}{% endif %}\
       {% if docker_options is defined %} {{ docker_options }}{% endif %}\
--- a/roles/openshift_health_checker/openshift_checks/docker_storage.py
+++ b/roles/openshift_health_checker/openshift_checks/docker_storage.py
@@ -1,5 +1,6 @@
 """Check Docker storage driver and usage."""
 import json
+import os.path
 import re
 from openshift_checks import OpenShiftCheck, OpenShiftCheckException, get_var
 from openshift_checks.mixins import DockerHostMixin
@@ -20,10 +21,27 @@ class DockerStorage(DockerHostMixin, OpenShiftCheck):
     storage_drivers = ["devicemapper", "overlay", "overlay2"]
     max_thinpool_data_usage_percent = 90.0
     max_thinpool_meta_usage_percent = 90.0
+    max_overlay_usage_percent = 90.0
 
-    # pylint: disable=too-many-return-statements
-    # Reason: permanent stylistic exception;
-    #         it is clearer to return on failures and there are just many ways to fail here.
+    # TODO(lmeyer): mention these in the output when check fails
+    configuration_variables = [
+        (
+            "max_thinpool_data_usage_percent",
+            "For 'devicemapper' storage driver, usage threshold percentage for data. "
+            "Format: float. Default: {:.1f}".format(max_thinpool_data_usage_percent),
+        ),
+        (
+            "max_thinpool_meta_usage_percent",
+            "For 'devicemapper' storage driver, usage threshold percentage for metadata. "
+            "Format: float. Default: {:.1f}".format(max_thinpool_meta_usage_percent),
+        ),
+        (
+            "max_overlay_usage_percent",
+            "For 'overlay' or 'overlay2' storage driver, usage threshold percentage. "
+            "Format: float. Default: {:.1f}".format(max_overlay_usage_percent),
+        ),
+    ]
+
     def run(self, tmp, task_vars):
         msg, failed, changed = self.ensure_dependencies(task_vars)
         if failed:
@@ -34,17 +52,17 @@ class DockerStorage(DockerHostMixin, OpenShiftCheck):
             }
 
         # attempt to get the docker info hash from the API
-        info = self.execute_module("docker_info", {}, task_vars=task_vars)
-        if info.get("failed"):
+        docker_info = self.execute_module("docker_info", {}, task_vars=task_vars)
+        if docker_info.get("failed"):
             return {"failed": True, "changed": changed,
                     "msg": "Failed to query Docker API. Is docker running on this host?"}
-        if not info.get("info"):  # this would be very strange
+        if not docker_info.get("info"):  # this would be very strange
             return {"failed": True, "changed": changed,
-                    "msg": "Docker API query missing info:\n{}".format(json.dumps(info))}
-        info = info["info"]
+                    "msg": "Docker API query missing info:\n{}".format(json.dumps(docker_info))}
+        docker_info = docker_info["info"]
 
         # check if the storage driver we saw is valid
-        driver = info.get("Driver", "[NONE]")
+        driver = docker_info.get("Driver", "[NONE]")
         if driver not in self.storage_drivers:
             msg = (
                 "Detected unsupported Docker storage driver '{driver}'.\n"
@@ -53,26 +71,34 @@ class DockerStorage(DockerHostMixin, OpenShiftCheck):
             return {"failed": True, "changed": changed, "msg": msg}
 
         # driver status info is a list of tuples; convert to dict and validate based on driver
-        driver_status = {item[0]: item[1] for item in info.get("DriverStatus", [])}
+        driver_status = {item[0]: item[1] for item in docker_info.get("DriverStatus", [])}
+
+        result = {}
+
         if driver == "devicemapper":
-            if driver_status.get("Data loop file"):
-                msg = (
-                    "Use of loopback devices with the Docker devicemapper storage driver\n"
-                    "(the default storage configuration) is unsupported in production.\n"
-                    "Please use docker-storage-setup to configure a backing storage volume.\n"
-                    "See http://red.ht/2rNperO for further information."
-                )
-                return {"failed": True, "changed": changed, "msg": msg}
-            result = self._check_dm_usage(driver_status, task_vars)
-            result['changed'] = result.get('changed', False) or changed
-            return result
+            result = self.check_devicemapper_support(driver_status, task_vars)
 
-        # TODO(lmeyer): determine how to check usage for overlay2
+        if driver in ['overlay', 'overlay2']:
+            result = self.check_overlay_support(docker_info, driver_status, task_vars)
 
-        return {"changed": changed}
+        result['changed'] = result.get('changed', False) or changed
+        return result
 
-    def _check_dm_usage(self, driver_status, task_vars):
-        """
+    def check_devicemapper_support(self, driver_status, task_vars):
+        """Check if dm storage driver is supported as configured. Return: result dict."""
+        if driver_status.get("Data loop file"):
+            msg = (
+                "Use of loopback devices with the Docker devicemapper storage driver\n"
+                "(the default storage configuration) is unsupported in production.\n"
+                "Please use docker-storage-setup to configure a backing storage volume.\n"
+                "See http://red.ht/2rNperO for further information."
+            )
+            return {"failed": True, "msg": msg}
+        result = self.check_dm_usage(driver_status, task_vars)
+        return result
+
+    def check_dm_usage(self, driver_status, task_vars):
+        """Check usage thresholds for Docker dm storage driver. Return: result dict.
         Backing assumptions: We expect devicemapper to be backed by an auto-expanding thin pool
         implemented as an LV in an LVM2 VG. This is how docker-storage-setup currently configures
         devicemapper storage. The LV is "thin" because it does not use all available storage
@@ -83,7 +109,7 @@ class DockerStorage(DockerHostMixin, OpenShiftCheck):
         could run out of space first; so we check both.
         """
         vals = dict(
-            vg_free=self._get_vg_free(driver_status.get("Pool Name"), task_vars),
+            vg_free=self.get_vg_free(driver_status.get("Pool Name"), task_vars),
             data_used=driver_status.get("Data Space Used"),
             data_total=driver_status.get("Data Space Total"),
             metadata_used=driver_status.get("Metadata Space Used"),
@@ -93,7 +119,7 @@ class DockerStorage(DockerHostMixin, OpenShiftCheck):
         # convert all human-readable strings to bytes
         for key, value in vals.copy().items():
             try:
-                vals[key + "_bytes"] = self._convert_to_bytes(value)
+                vals[key + "_bytes"] = self.convert_to_bytes(value)
             except ValueError as err:  # unlikely to hit this from API info, but just to be safe
                 return {
                     "failed": True,
@@ -131,10 +157,12 @@ class DockerStorage(DockerHostMixin, OpenShiftCheck):
         vals["msg"] = "\n".join(messages or ["Thinpool usage is within thresholds."])
         return vals
 
-    def _get_vg_free(self, pool, task_vars):
-        # Determine which VG to examine according to the pool name, the only indicator currently
-        # available from the Docker API driver info. We assume a name that looks like
-        # "vg--name-docker--pool"; vg and lv names with inner hyphens doubled, joined by a hyphen.
+    def get_vg_free(self, pool, task_vars):
+        """Determine which VG to examine according to the pool name. Return: size vgs reports.
+        Pool name is the only indicator currently available from the Docker API driver info.
+        We assume a name that looks like "vg--name-docker--pool";
+        vg and lv names with inner hyphens doubled, joined by a hyphen.
+        """
         match = re.match(r'((?:[^-]|--)+)-(?!-)', pool)  # matches up to the first single hyphen
         if not match:  # unlikely, but... be clear if we assumed wrong
             raise OpenShiftCheckException(
@@ -163,7 +191,8 @@ class DockerStorage(DockerHostMixin, OpenShiftCheck):
         return size
 
     @staticmethod
-    def _convert_to_bytes(string):
+    def convert_to_bytes(string):
+        """Convert string like "10.3 G" to bytes (binary units assumed). Return: float bytes."""
         units = dict(
             b=1,
             k=1024,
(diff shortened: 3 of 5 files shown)


sosiouxme:
aos-ci-test

inventory/byo/hosts.origin.example
```diff
@@ -113,6 +113,9 @@ openshift_release=v3.6
 # Downgrades are not supported and will error out. Be careful when upgrading docker from < 1.10 to > 1.10.
 # docker_version="1.12.1"
 
+# Specify whether to --enable-selinux on Docker (set false if using overlay/2 storage drivers before RHEL 7.4)
```
rhcarvalho: nit: perhaps `overlay2` as that is the name of the driver in config files, etc?

rhcarvalho: Hmm, instead of suggesting people to manually update their inventory (and they may be starting off an old copy of this example file), wouldn't it be possible and wise to "do the right thing" in the playbook(s) that use this value?

roles/docker/tasks/package_docker.yml
```diff
@@ -93,7 +93,7 @@
     dest: /etc/sysconfig/docker
     regexp: '^OPTIONS=.*$'
     line: "OPTIONS='\
-      {% if ansible_selinux.status | default(None) == '''enabled''' and docker_selinux_enabled | default(true) %} --selinux-enabled {% endif %}\
+      {% if ansible_selinux.status | default(None) == '''enabled''' and docker_selinux_enabled | default(true) | bool %} --selinux-enabled {% endif %}\
```
rhcarvalho: Are those triple quotes necessary?!

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -15,15 +16,28 @@ class DockerStorage(DockerHostMixin, OpenShiftCheck):
...
+            "max_thinpool_data_usage_percent",
+            "For 'devicemapper' storage driver, usage threshold percentage for data. Format: float. Default: 90.0",
+        ),
+        (
+            "max_thinpool_meta_usage_percent",
+            "For 'devicemapper' storage driver, usage threshold percentage for metadata. Format: float. Default: 90.0",
+        ),
+        (
+            "max_overlay_usage_percent",
+            "For 'overlay' or 'overlay2' storage driver, usage threshold percentage. Format: float. Default: 90.0",
```
rhcarvalho: Suggestion: If we define `configuration_variables` below `max_overlay_usage_percent` et al, we could reference the class attributes instead of hardcoding the defaults / duplicating.

```python
In [1]: class Foo:
   ...:     bar = 1
   ...:     doc = 'Default bar: {}'.format(bar)
   ...:     

In [2]: Foo.doc
Out[2]: 'Default bar: 1'
```

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -53,26 +67,34 @@ def run(self, tmp, task_vars):
...
-            result['changed'] = result.get('changed', False) or changed
-            return result
+            result = self.check_devicemapper_support(driver_status, task_vars)
 
-        # TODO(lmeyer): determine how to check usage for overlay2
+        if driver in ['overlay', 'overlay2']:
+            result = self.check_overlay_support(docker_info, driver_status, task_vars)
 
-        return {"changed": changed}
+        result['changed'] = result.get('changed', False) or changed
```
rhcarvalho: :+1:

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -183,3 +208,79 @@ def _convert_to_bytes(string):
...
+                    "For Docker OverlayFS mount point {path},\n"
+                    "usage percentage {pct:.1f} is higher than threshold {thresh:.1f}."
+                ).format(path=mount["mount"], pct=usage, thresh=threshold)
+            }
+
+        return {}
+
+    # TODO(lmeyer): migrate to base class
+    @staticmethod
+    def find_ansible_mount(path, ansible_mounts):
```
rhcarvalho: Don't we have this code somewhere else?!

Suggestion: instead of migrating to the base class (OpenShiftCheck?), staticmethod -> function and import and reuse the function where needed.

inventory/byo/hosts.origin.example
sosiouxme: There are two drivers, `overlay` and `overlay2`. Is `overlay/2` a confusing abbreviation for that? Hm, perhaps...

sosiouxme: There's one task in one role that uses it. And as far as I can see, nothing that specifies what storage driver you're using (when I configure it in `/etc/sysconfig/docker-storage` it doesn't get reconfigured). I don't know of any other reason why you would set this to false, and I suspect no one else does either because it didn't even work.

roles/docker/tasks/package_docker.yml
sosiouxme: Doubt it but they were there when I got here :)

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: Good point.

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: ah but see, after `task_vars` is internalized this can become `find_ansible_mount(path)` if it's an instance method.

juanvallejo:
Aside from https://github.com/openshift/openshift-ansible/pull/4757/files#r127292139 (not sure which direction we should take, but I agree it would make sense to handle in playbook), lgtm

inventory/byo/hosts.origin.example
rhcarvalho: Aha, yes, I did not understand it applied to both.

Better say "overlay / overlay2", then.

roles/docker/tasks/package_docker.yml
rhcarvalho: Yes, I know they were there, might be our chance to clean it up :)

inventory/byo/hosts.origin.example
ashcrow: nit: May be worth noting that it will change ``container-daemon.json`` on container-engine. Eventually everything will move to a json file:

```
    "selinux-enabled": {{ l_docker_selinux_enabled | lower }},
```

sosiouxme: @ashcrow ah, thanks. I overlooked that. Hmm, I see that used here:

```
roles/docker/tasks/systemcontainer_docker.yml:    l_docker_selinux_enabled: "{{ docker_selinux_enabled | default(true) | to_json }}"
roles/docker/templates/daemon.json:    "selinux-enabled": {{ l_docker_selinux_enabled | lower }},
roles/openshift_docker_facts/tasks/main.yml:      selinux_enabled: "{{ openshift_docker_selinux_enabled | default(None) }}"
```

Maybe I should be documenting `openshift_docker_selinux_enabled` instead?

I think I see the drift of your nit... when running Docker as a system container, we're not setting flags on the daemon, but options in a config json. So maybe like this instead:

```
# Specify whether to run Docker daemon with SELinux enabled in containers.
# (Needs to be false to use overlay/overlay2 storage drivers before RHEL 7.4)
# openshift_docker_selinux_enabled=True
```

ashcrow: @sosiouxme that works foe me!

inventory/byo/hosts.origin.example
```diff
@@ -113,6 +113,10 @@ openshift_release=v3.6
 # Downgrades are not supported and will error out. Be careful when upgrading docker from < 1.10 to > 1.10.
 # docker_version="1.12.1"
 
+# Specify whether to set flag --enable-selinux on Docker daemon.
+# (Needs to be false to use overlay/overlay2 storage drivers before RHEL 7.4)
+# docker_selinux_enabled=True
```
ashcrow: Should this be ``openshift_docker_selinux_enabled``?

inventory/byo/hosts.origin.example
```diff
@@ -113,6 +113,10 @@ openshift_release=v3.6
 # Downgrades are not supported and will error out. Be careful when upgrading docker from < 1.10 to > 1.10.
 # docker_version="1.12.1"
 
+# Specify whether to set flag --enable-selinux on Docker daemon.
```
ashcrow: If this is the same as the ``ose`` example don't forget to update the wording.

sosiouxme: aw geez, I edited the file and didn't save/commit.

sosiouxme:
I think I've addressed the feedback, @rhcarvalho feel free to take another pass.

inventory/byo/hosts.origin.example
```diff
@@ -113,6 +113,10 @@ openshift_release=v3.6
 # Downgrades are not supported and will error out. Be careful when upgrading docker from < 1.10 to > 1.10.
 # docker_version="1.12.1"
 
+# Specify whether to run Docker daemon with SELinux enabled in containers.
+# (Needs to be false to use overlay/overlay2 storage drivers before RHEL 7.4)
```
rhcarvalho: nit: the True/False here is case-insensitive, isn't it? I'd suggest dropping the parentheses and make it a "command" statement:

```
Set to False when using the overlay or overlay2 storage driver on RHEL 7.3 or earlier.
```

What about other operating systems? CentOS, Fedora?

Is it clear that the commented-out values are the defaults? I'm not sure one can reason that way in the example inventory.

sosiouxme: Maybe "uncomment" as a command would be more helpful.
```
# Specify whether to run Docker daemon with SELinux enabled in containers. Default is True.
# Uncomment below to disable; for example if your kernel does not support the
# overlay/overlay2 docker storage drivers with SELinux enabled.
#openshift_docker_selinux_enabled=False
```

rhcarvalho: Yes :+1:

roles/docker/tasks/package_docker.yml
```diff
@@ -93,7 +93,7 @@
     dest: /etc/sysconfig/docker
     regexp: '^OPTIONS=.*$'
     line: "OPTIONS='\
-      {% if ansible_selinux.status | default(None) == '''enabled''' and docker_selinux_enabled | default(true) %} --selinux-enabled {% endif %}\
+      {% if ansible_selinux.status | default(None) == 'enabled' and docker_selinux_enabled | default(true) | bool %} --selinux-enabled {% endif %}\
```
rhcarvalho: Let me see if I understand this: without `bool`, setting `docker_selinux_enabled=False` in the inventory would make the variable a string with value `"False"`, which is a truthy value in Python leading to always true no matter what string/value you set in the inventory. Is that right?

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -20,10 +21,27 @@ class DockerStorage(DockerHostMixin, OpenShiftCheck):
...
+            "max_thinpool_meta_usage_percent",
+            "For 'devicemapper' storage driver, usage threshold percentage for metadata. "
+            "Format: float. Default: {:.1f}".format(max_thinpool_meta_usage_percent),
+        ),
+        (
+            "max_overlay_usage_percent",
+            "For 'overlay' or 'overlay2' storage driver, usage threshold percentage. "
+            "Format: float. Default: {:.1f}".format(max_overlay_usage_percent),
+        ),
+    ]
```
rhcarvalho: :+1:

roles/docker/tasks/package_docker.yml
sosiouxme: exactly.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -183,3 +212,79 @@ def _convert_to_bytes(string):
...
+                ).format(version=kernel)
+                return {"failed": True, "msg": msg}
+            # NOTE: we could check for --selinux-enabled here but docker won't even start with
+            # that option until it's supported in the kernel so we don't need to.
+
+        return self.check_overlay_usage(docker_info, task_vars)
+
+    def check_overlay_usage(self, docker_info, task_vars):
+        """Check disk usage on OverlayFS backing store volume. Return: result dict."""
+        path = os.path.join(docker_info.get("DockerRootDir", "/var/lib/docker"), docker_info.get("Driver"))
```
rhcarvalho: It is strange to use `os.path.join` and have strings with an specific path separator. In this case we only ever intend this to work on Linux anyway.(?)

Don't we need a default value for `"Driver"`? I think `os.path.join` will raise an exception if one of the arguments is not a string.

```python
path = "{}/{}".format(docker_info.get("DockerRootDir", "/var/lib/docker"), docker_info.get("Driver", ""))
```

roles/openshift_health_checker/openshift_checks/docker_storage.py
rhcarvalho: Yeah, I looked it up, the code we have is not quite the same but we have at least two variations of processing `ansible_mounts`, in `etcd_imagedata_size.py` and `disk_availability.py`.

I'm okay with improving / generalizing later.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -183,3 +212,79 @@ def _convert_to_bytes(string):
...
+        return {}
+
+    # TODO(lmeyer): migrate to base class
+    @staticmethod
+    def find_ansible_mount(path, ansible_mounts):
+        """Return the mount point for path from ansible_mounts."""
+
+        mount_for_path = {mount['mount']: mount for mount in ansible_mounts}
+        mount_point = path
+        while mount_point not in mount_for_path:
```
rhcarvalho: Suggestion: add a paranoia-compliant way to avoid an infinite loop:

https://github.com/openshift/openshift-ansible/blob/2476720c314f8e20cf53e4c622bcd50e432f8d01/roles/openshift_health_checker/openshift_checks/disk_availability.py#L103

Note that `os.path.dirname(mount_point)` may give the empty string instead of `/`:

```python
In [1]: import os

In [2]: os.path.dirname('foo')
Out[2]: ''
```

roles/openshift_health_checker/test/docker_storage_test.py
```diff
@@ -56,14 +56,35 @@ def test_is_active(is_containerized, group_names, is_active):
...
         False,
         [],
     ),
     (
         dict(info={
             "Driver": "overlay",
+            "DriverStatus": [("Backing Filesystem", "btrfs")],
+        }),
+        True,
+        ["'xfs' as the backing storage"],
```
rhcarvalho: Why is `xfs` in the expected error message if the backing filesystem is `btrfs`?

roles/openshift_health_checker/test/docker_storage_test.py
```diff
@@ -219,6 +241,52 @@ def test_convert_to_bytes(string, expect_bytes):
...
+        "bogus percent",
+        True,
+        ["is not a percentage"],
+    ),
+])
+def test_overlay_usage(ansible_mounts, threshold, expect_fail, expect_msg):
+    check = dummy_check()
+    task_vars = non_atomic_task_vars.copy()
+    task_vars["ansible_mounts"] = ansible_mounts
+    if threshold:
```
rhcarvalho: Suggestion:

```python
if threshold is not None:
```

For clarity of intent and to avoid surprises. Otherwise only "thruthy" thresholds would be interpreted correctly.

roles/openshift_health_checker/test/docker_storage_test.py
```diff
@@ -219,6 +241,52 @@ def test_convert_to_bytes(string, expect_bytes):
...
+    (
+        ansible_mounts_not_enough,
+        "bogus percent",
+        True,
+        ["is not a percentage"],
+    ),
+])
+def test_overlay_usage(ansible_mounts, threshold, expect_fail, expect_msg):
+    check = dummy_check()
+    task_vars = non_atomic_task_vars.copy()
```
rhcarvalho: Perhaps `non_atomic_task_vars` should be a fixture?

```python
@pytest.fixture()
def non_atomic_task_vars():
    return {"openshift": {"common": {"is_atomic": False}}}

def test_overlay_usage(..., non_atomic_task_vars, ...):
```

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: If we got to this point in the code, we have a driver. It's just a question of whether it's `overlay` or `overlay2`.  (Actually if we ever get a docker info with the driver unset then something is totally messed up.)

roles/openshift_health_checker/test/docker_storage_test.py
sosiouxme: because the error message tells you what it's supposed to be as well as what it is 
https://github.com/openshift/openshift-ansible/pull/4757/files#diff-4739a080555d5d9b11dc300197eb46b8R223

roles/openshift_health_checker/test/docker_storage_test.py
sosiouxme: But then it's mentally cluttering up the long args, mixing fixtures with parametrized... *shrug* I can do it if you prefer, it's a weak preference for me.

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: Then I can just check that it's "" or "/" ... there are no other cases it can devolve to, right?

rhcarvalho: Either `'/'` or `''`, yes.

roles/openshift_health_checker/test/docker_storage_test.py
rhcarvalho: Reading the test, how is one to tell what the file systems are?! I saw `"btrfs"` in one place, and was confused seeing `"'xfs' ...` nearby.

roles/openshift_health_checker/openshift_checks/docker_storage.py
rhcarvalho: > If we got to this point in the code, we have a driver. It's just a question of whether it's overlay or overlay2.

If we guarantee we have a Driver, then why not `docker_info['Driver']`?

Similarly, if we are sure about DockerRootDir, we could then write:

```python
path = docker_info['DockerRootDir'] + '/' + docker_info['Driver']
```

or

```python
# not important, but avoids an intermediary string in the previous code
path = '{}/{}'.format(docker_info['DockerRootDir'], docker_info['Driver'])
```

This is possible, but problematic if the values are not strings:

```python
path = '/'.join(docker_info['DockerRootDir'], docker_info['Driver'])
```

I care about this being defensive code, so if there is a remote chance one of those keys are not in `docker_info`, then we can wrap it with a `try: ... except KeyError` block.

roles/openshift_health_checker/test/docker_storage_test.py
rhcarvalho: It's not a big deal as is now. The problem would only show itself as we make changes. If for some reason a test mutates `non_atomic_task_vars`, the tests that run after would start making copies of the mutated value.

If we make it a function that returns the value, we guarantee that we won't introduce that sort of bug.

```python
def non_atomic_task_vars():
    return {"openshift": {"common": {"is_atomic": False}}}

def test_overlay_usage(...):
    # ...
    task_vars = non_atomic_task_vars()
    # ...
```

Making it a fixture or not is orthogonal to that.

sosiouxme: +1 will make it a function.

roles/openshift_health_checker/test/docker_storage_test.py
sosiouxme: I can see how the test looks confusing, will improve

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: It's a remote API call, and while I think the chance is slim we'll ever see it come back missing these keys, I can't completely discount it. The question is how much I want to mess up the code to try to make that remote possibility less confusing to the user if it did happen :) I'll try for something better here.

sosiouxme:
@rhcarvalho updated - good now?

roles/openshift_health_checker/openshift_checks/docker_storage.py
sosiouxme: To explain where I landed...

I have absolute confidence `docker_info["Driver"]` is defined because this overlay-specific function could never be called otherwise.

If `DockerRootDir` ever actually came back missing yet docker is actually up enough to respond (say the API changes radically), I do not feel bad about assuming it's the usual directory and the user being none the wiser for now.

sosiouxme:
aos-ci-test

roles/openshift_health_checker/openshift_checks/docker_storage.py
rhcarvalho: :+1:

roles/openshift_health_checker/test/docker_storage_test.py
```diff
@@ -56,14 +57,35 @@ def test_is_active(is_containerized, group_names, is_active):
...
         False,
         [],
     ),
     (
         dict(info={
             "Driver": "overlay",
+            "DriverStatus": [("Backing Filesystem", "btrfs")],
+        }),
+        True,
+        ["storage is type 'btrfs'", "only supported with\n'xfs'"],
```
sosiouxme: Yeah, it does, which is actually why it wasn't there before. But the `\n` doesn't truly matter, it's the generic problem of having error string outputs instead of defined exceptions. I'd like to refactor as much as possible into exceptions at some point. It's tech debt for now.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -183,3 +212,79 @@ def _convert_to_bytes(string):
...
+            ).format(fs=backing_fs)
+            return {"failed": True, "msg": msg}
+
+        # check support for OS and kernel version
+        o_s = docker_info.get("OperatingSystem", "[NONE]")
+        if "Red Hat Enterprise Linux" in o_s or "CentOS" in o_s:
+            # keep it simple, only check enterprise kernel versions; assume everyone else is good
+            kernel = docker_info.get("KernelVersion", "[NONE]")
+            kernel_arr = [int(num) for num in re.findall(r'\d+', kernel)]
+            if kernel_arr < [3, 10, 0, 514]:  # rhel < 7.3
```
sosiouxme: In my head, sure :)
What do we want to say about completely bogus kernel versions coming through? This will fail the check and they'll be in the output so it should be a pretty good clue what went wrong.

As far as non-numeric portions, we are only looking down to the first release segment, and I feel confident RHEL will continue with the x.y.z-r format with all numbers for the OS releases of interest here. I don't think the first release number even changes during the lifetime of a RHEL release. At the very least, the non-numeric bits (e.g. `el7`) do not break it.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -183,3 +212,79 @@ def _convert_to_bytes(string):
...
+            threshold = float(threshold)
+        except ValueError:
+            return {
+                "failed": True,
+                "msg": "Specified 'max_overlay_usage_percent' is not a percentage: {}".format(threshold),
+            }
+
+        mount = self.find_ansible_mount(path, get_var(task_vars, "ansible_mounts"))
+        free_bytes = mount.get('size_available', 0)
+        total_bytes = mount.get('size_total', 1)
```
sosiouxme: If someone has defined a zero-size volume - well, I don't think it's possible, but if it is (or ansible had a bug and 0 came through somehow) what do we want to say about it? Is it a blocker to merging today?

rhcarvalho: I'd rather fail explicitly if total_bytes is 0 than seeing usage = `100.0 * (1 - free_bytes)`.

What would you think as a user if an error message came out saying you're using `-1.5 GB`?! I think that's a terrible experience, and I could even imagine myself ignoring the negative sign at first out of incredulity.

sosiouxme: Yeah, that would suck, I'm just not seeing how it would happen unless ansible collects completely bogus values... free bytes greater than total??

rhcarvalho: I just want us to be on the safe side. It would be plausible to have both readings = 0, in which case the default value of 1 in this line is useless, and we still get a ZeroDivisionError below.

roles/openshift_health_checker/openshift_checks/docker_storage.py
```diff
@@ -183,3 +212,79 @@ def _convert_to_bytes(string):
...
+
+        usage = 100.0 * (total_bytes - free_bytes) / total_bytes
```
sosiouxme: Updated, how does the error handling look? Anything else?

sosiouxme:
3.6 is a separate branch now, so we can merge to master. To address the blocker bug there is a parallel PR https://github.com/openshift/openshift-ansible/pull/4791

It's currently tracked as a blocker. It is in fact more like a feature but it's a new feature on a new feature so low-risk. You guys OK with proceeding as a blocker?

brenton:
I'm fine with this merging for 3.6.

sosiouxme:
[merge]

rhcarvalho:
Flake https://github.com/openshift/origin/issues/15356

[merge]

PR #4758, opened 2017-07-13
Allow Open vSwitch 2.7 on OpenShift 3.5 and 3.6

rhcarvalho:
Change the package_version check to tolerate either OVS 2.6 or 2.7.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1465882

### DIFF
--- a/roles/openshift_health_checker/library/aos_version.py
+++ b/roles/openshift_health_checker/library/aos_version.py
@@ -19,6 +19,10 @@ the inventory, the version comparison checks just pass.
 '''
 
 from ansible.module_utils.basic import AnsibleModule
+# NOTE: because of the dependency on yum (Python 2-only), this module does not
+# work under Python 3. But since we run unit tests against both Python 2 and
+# Python 3, we use six for cross compatibility in this module alone:
+from ansible.module_utils.six import string_types
 
 IMPORT_EXCEPTION = None
 try:
@@ -122,12 +126,15 @@ def _check_precise_version_found(pkgs, expected_pkgs_dict):
     for pkg in pkgs:
         if pkg.name not in expected_pkgs_dict:
             continue
-        # does the version match, to the precision requested?
-        # and, is it strictly greater, at the precision requested?
-        expected_pkg_version = expected_pkgs_dict[pkg.name]["version"]
-        match_version = '.'.join(pkg.version.split('.')[:expected_pkg_version.count('.') + 1])
-        if match_version == expected_pkg_version:
-            pkgs_precise_version_found.add(pkg.name)
+        expected_pkg_versions = expected_pkgs_dict[pkg.name]["version"]
+        if isinstance(expected_pkg_versions, string_types):
+            expected_pkg_versions = [expected_pkg_versions]
+        for expected_pkg_version in expected_pkg_versions:
+            # does the version match, to the precision requested?
+            # and, is it strictly greater, at the precision requested?
+            match_version = '.'.join(pkg.version.split('.')[:expected_pkg_version.count('.') + 1])
+            if match_version == expected_pkg_version:
+                pkgs_precise_version_found.add(pkg.name)
 
     not_found = []
     for name, pkg in expected_pkgs_dict.items():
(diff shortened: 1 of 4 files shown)


roles/openshift_health_checker/test/package_version_test.py
```diff
@@ -72,36 +72,6 @@ def execute_module(module_name=None, module_args=None, tmp=None, task_vars=None)
...
-            if pkg["name"] == "openvswitch":
-                assert pkg["version"] == expected_ovs_version
-
-        return return_value
-
-    check = PackageVersion(execute_module=execute_module)
-    result = check.run(tmp=None, task_vars=task_vars)
-    assert result is return_value
-
-
```
rhcarvalho: Removing this test because it adds no coverage and tests data instead of logic.
This coupling makes every change to supported versions require the same changes to the tests, which is bad.

roles/openshift_health_checker/test/aos_version_test.py
```diff
@@ -73,7 +73,7 @@
         },  # eggs found with multiple versions
     ),
 ])
-def test_check_pkgs_for_precise_version(pkgs, expect_not_found):
+def test_check_precise_version_found(pkgs, expect_not_found):
```
rhcarvalho: Renamed here and below for the sanity of future readers ;-)

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -10,8 +10,8 @@ class PackageVersion(NotContainerizedMixin, OpenShiftCheck):
     tags = ["preflight"]
 
     openshift_to_ovs_version = {
-        "3.6": "2.6",
-        "3.5": "2.6",
+        "3.6": ["2.6", "2.7"],
+        "3.5": ["2.6", "2.7"],
```
rhcarvalho: Adding 2.7 to both 3.5 and 3.6 as discussed on IRC with @brenton and @sdodson.

rhcarvalho:
I've improved the tests for `aos_version` to cover the cases where the input contains multiple versions.

rhcarvalho:
aos-ci-test

roles/openshift_health_checker/test/aos_version_test.py
```diff
@@ -18,7 +18,43 @@
...
+        [Package('spam', '3.2.1'), Package('eggs', '3.2.1')],
+        expected_pkgs,
+    ),
+    (
+        # found with more specific version
+        [Package('spam', '3.2.1'), Package('eggs', '3.2.1.5')],
+        expected_pkgs,
+    ),
+    (
+        [Package('ovs', '2.6'), Package('ovs', '2.4')],
```
sosiouxme: see this would have been funnier with package `ova` because that's latin for "eggs"...

rhcarvalho: hahaha

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -122,12 +122,15 @@ def _check_precise_version_found(pkgs, expected_pkgs_dict):
...
         if pkg.name not in expected_pkgs_dict:
             continue
-        # does the version match, to the precision requested?
-        # and, is it strictly greater, at the precision requested?
-        expected_pkg_version = expected_pkgs_dict[pkg.name]["version"]
-        match_version = '.'.join(pkg.version.split('.')[:expected_pkg_version.count('.') + 1])
-        if match_version == expected_pkg_version:
-            pkgs_precise_version_found.add(pkg.name)
+        expected_pkg_versions = expected_pkgs_dict[pkg.name]["version"]
+        if isinstance(expected_pkg_versions, basestring):
```
rhcarvalho: `basestring` doesn't exist in py3... and by chance I see that `yum` makes use of it... which means that this can't possibly work on py3...

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -26,6 +27,13 @@
 except ImportError as err:
     IMPORT_EXCEPTION = err
 
+# NOTE: yum does not work with Python 3, but since the openshift-ansible code
+# base is tested against both Python 2 and Python 3, we write the code below for
+# compatibility with the latter:
+PY3 = sys.version_info[0] == 3
+if PY3:
+    basestring = str
```
rhcarvalho: The idea is borrowed from `six`, with an adaptation to define `basestring` instead of `string_types`, so avoiding introducing an extra name.

This is only needed to make the unit tests pass in py3, even though in practice one cannot use py3 with this check because of the dependency on `yum` (py2-only).

rhcarvalho:
aos-ci-test

rhcarvalho:
aos-ci-test

roles/openshift_health_checker/library/aos_version.py
rhcarvalho: Well, that approach causes a `pylint` error because it looks at code statically and thinks that 1) under py2 the builtin name `basestring` is being overwritten; and 2) under py3 the name `basestring` is invalid for a constant. Instead, we can import `six` from `ansible.module_utils`.

rhcarvalho:
Flake on excluders https://github.com/openshift/origin/issues/13814

rhcarvalho:
aos-ci-test

rhcarvalho:
[merge][severity:blocker]

rhcarvalho:
@sdodson could you please help here? The merge failed because of the status of aos-ci-test, though we have green from all jobs but a internal error on the "aos-ci-jenkins/all".

PR #4759, opened 2017-07-13
Check for router service annotations when redeploying router certificates.

abutcher:
Check current router annotations s.t. we can avoid redeploying the router's service serving signer signed certificate when they are not present.


abutcher:
aos-ci-test

PR #4762, opened 2017-07-14
Fix the PROXY variables, once and for all

tbielawa:
This will fix a myriad of proxy setting related bugs. At the very least it will impact these:

* https://bugzilla.redhat.com/show_bug.cgi?id=1470165
~~* https://bugzilla.redhat.com/show_bug.cgi?id=1466783~~ (just kidding, this needs some separate work)
* https://bugzilla.redhat.com/show_bug.cgi?id=1468424

### DIFF
--- a/inventory/byo/hosts.origin.example
+++ b/inventory/byo/hosts.origin.example
@@ -740,6 +740,10 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # Global Proxy Configuration
 # These options configure HTTP_PROXY, HTTPS_PROXY, and NOPROXY environment
 # variables for docker and master services.
+#
+# Hosts in the openshift_no_proxy list will NOT use any globally
+# configured HTTP(S)_PROXYs. openshift_no_proxy accepts domains
+# (.example.com), and hosts (example.com), and IP addresses.
 #openshift_http_proxy=http://USER:PASSWORD@IPADDR:PORT
 #openshift_https_proxy=https://USER:PASSWORD@IPADDR:PORT
 #openshift_no_proxy='.hosts.example.com,some-host.com'
@@ -747,7 +751,11 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true', 
 # Most environments don't require a proxy between openshift masters, nodes, and
 # etcd hosts. So automatically add those hostnames to the openshift_no_proxy list.
 # If all of your hosts share a common domain you may wish to disable this and
-# specify that domain above.
+# specify that domain above instead.
+#
+# For example, having hosts with FQDNs: m1.ex.com, n1.ex.com, and
+# n2.ex.com, one would simply add '.ex.com' to the openshift_no_proxy
+# variable (above) and set this value to False
 #openshift_generate_no_proxy_hosts=True
 #
 # These options configure the BuildDefaults admission controller which injects
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1652,6 +1652,7 @@ def set_proxy_facts(facts):
         )
 
         if not proxy_settings_defined:
+            common['no_proxy'] = ''
             return facts
 
         # As of 3.6 if ANY of the proxy parameters are defined in the
--- a/roles/openshift_facts/library/openshift_facts.py
+++ b/roles/openshift_facts/library/openshift_facts.py
@@ -1643,38 +1643,74 @@ def set_proxy_facts(facts):
     if 'common' in facts:
         common = facts['common']
 
-        # No openshift_no_proxy settings detected, empty list for now
-        if 'no_proxy' not in common:
-            common['no_proxy'] = []
+        ######################################################################
+        # We can exit early now if we don't need to set any proxy facts
+        proxy_params = ['no_proxy', 'https_proxy', 'http_proxy']
+        # If any of the known Proxy Params (pp) are defined
+        proxy_settings_defined = any(
+            [True for pp in proxy_params if pp in common]
+        )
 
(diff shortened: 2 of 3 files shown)


tbielawa:
Moving the hacky test script into this gist https://gist.github.com/tbielawa/7276903c31ad3dfb0f2759d72f2c5547

tbielawa:
Squashed.
Rebased onto latest master
Removed the test script (it's in the linked gist, above)

tbielawa:
[test]

tbielawa:
aos-ci-test

inventory/byo/hosts.origin.example
```diff
@@ -740,14 +740,22 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
 # Global Proxy Configuration
 # These options configure HTTP_PROXY, HTTPS_PROXY, and NOPROXY environment
 # variables for docker and master services.
+#
+# Hosts in the openshift_no_proxy list will NOT use any globally
+# configured HTTP(S)_PROXYs. openshift_no_proxy accepts domains
+# (.example.com), hosts (example.com), and CIDR ranges (10.1.0.0/16)
```
sdodson: But certain implementation's don't support CIDR ranges. Docker doesn't. OpenShift does.

inventory/byo/hosts.origin.example
```diff
@@ -740,14 +740,22 @@ openshift_master_identity_providers=[{'name': 'htpasswd_auth', 'login': 'true',
...
 #
 # Most environments don't require a proxy between openshift masters, nodes, and
 # etcd hosts. So automatically add those hostnames to the openshift_no_proxy list.
 # If all of your hosts share a common domain you may wish to disable this and
-# specify that domain above.
+# specify that domain above instead.
+#
+# For example, having hosts with FQDNs: m1.ex.com, n1.ex.com, and
+# n2.ex.com, one would simply add '.ex.com' to the openshift_no_proxy
+# variable (above) and set this value to False
```
sdodson: 👍

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1643,38 +1643,74 @@ def set_proxy_facts(facts):
...
+        # inventory then we MUST add certain domains to the NO_PROXY
+        # environment variable.
+
+        ######################################################################
+
+        # Spot to build up some data we may insert later
+        raw_no_proxy_list = []
+
+        # Automatic 3.6 NO_PROXY additions if a proxy is in use
+        svc_cluster_name = ['.svc', '.' + common['dns_domain'], common['hostname']]
```
ashcrow: Just verifying, but ``dns_domain`` can never fall back to be an IP address, correct? By it's name I'd assume not but just in case ...

sdodson: I don't think we do anything to ensure that someone doesn't set it to an ip address but I'd never expect that to work. It's always used as a dns suffix so I'd expect everything to blow up violently if it weren't a suffix.

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -1643,38 +1643,74 @@ def set_proxy_facts(facts):
...
+
+        # auto_hosts: Added to NO_PROXY list if any proxy params are
```
ashcrow: nit: Since this logic only is used if ``generate_no_proxy_hosts`` is ``True`` it may be worth moving it to line 1692 so it only is generated and stored if it's used.

tbielawa: Discussed on IRC. Implications of leaving the definition in place are negligible. Going to leave as is.

tbielawa:
[merge][severity:blocker]

tbielawa:
@openshift/team-documentation , @adellape  this PR introduces and solidifies behavior which is new in 3.6 and may differ slightly from the existing documentation

https://docs.openshift.org/latest/install_config/install/advanced_install.html#advanced-install-configuring-global-proxy

> If your hosts require use of a HTTP or HTTPS proxy in order to connect to external hosts, there are many components that must be configured to use the proxy, including masters, Docker, and builds. Node services only connect to the master API requiring no external access and therefore do not need to be configured to use a proxy.

May want to reflect explicitly in the online documentation that if **any** of

* `openshift_no_proxy`
* `openshift_https_proxy`
* `openshift_http_proxy` 

are set, then all cluster hosts will have an automatically generated `NO_PROXY` environment variable injected into several service configuration scripts. I'm not sure how you want to describe that, but the point I'm trying to make it, it should be clear that if you set any proxy variables then you will have a `NO_PROXY` environment variable set up for you.

Setting the `http/s_proxy` variables by themselves will cause all cluster members to be added to the NO_PROXY environment variable. As well as the default `.svc` domain and your clusters `dns_domain` (typically `.cluster.local`).

sdodson:
aos-ci-test

ashcrow:
flake https://github.com/openshift/origin/issues/10162

ashcrow:
[merge][severity:blocker]

tbielawa:

tbielawa:
flake https://github.com/openshift/origin/issues/10162

tbielawa:
aos-ci-test

tbielawa:
I think other parts of the openshift-ansible API were expecting `openshift.common.no_proxy` to either be **defined**, or if `undefined` then they set it to an empty list. I am hoping that by setting `common['no_proxy']` to an empty string in the early-`return` I can avoid those `default([])`ing branches in the playbooks later on.

vikram-redhat:
@tbielawa - Thanks Tim. Added a card to our Trello board to make these updates in the docs and tagged @adellape to it. https://trello.com/c/7IsRsDEo/596-updates-for-proxy-variables

sdodson:
[merge][severity:blocker]

sdodson:
"once and for all" lol

PR #4764, opened 2017-07-14
Switch CI to ansible-2.3.1.0

sdodson:
This only gets tox and friends but I'll rev ansible on the CI slave too.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,9 +24,4 @@ commands =
     generate_validation: python setup.py generate_validation
     # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
     ansible_syntax: python setup.py ansible_syntax
-    # ansible 2.2.2+ unfortunately breaks the integration test runner
-    # because it can no longer set facts on the test docker hosts.
-    # So for now, install separate ansible version for integration.
-    # PR that fixes it: https://github.com/ansible/ansible/pull/23599
-    # Once that PR is available, drop this and use same ansible.
-    integration: pip install ansible==2.2.1.0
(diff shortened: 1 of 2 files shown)


sdodson:
[test]

sosiouxme:
Actually the origin build with playbook2image also uses the `requirements.txt`. However that's changing.

Main concern is integration tests, checking those locally along with the other tox stuff.

tox.ini
```diff
@@ -24,9 +24,6 @@ commands =
     generate_validation: python setup.py generate_validation
     # TODO(rhcarvalho): check syntax of other important entrypoint playbooks
     ansible_syntax: python setup.py ansible_syntax
-    # ansible 2.2.2+ unfortunately breaks the integration test runner
-    # because it can no longer set facts on the test docker hosts.
-    # So for now, install separate ansible version for integration.
     # PR that fixes it: https://github.com/ansible/ansible/pull/23599
     # Once that PR is available, drop this and use same ansible.
-    integration: pip install ansible==2.2.1.0
+    integration: pip install ansible==2.3.1.0
```
sosiouxme: no real reason to keep this or the comment above. could replace with:

`    integration: python -c 'print "run test/integration/run-tests.sh"'`

sdodson: updated

sdodson:
aos-ci-test

sdodson:
[merge][severity:blocker]

sdodson:
I'm considering this a blocker because ansible-2.3.1.0 will release in RHEL 7.4 extras and we need to ensure that we pivot to that version ahead of that.

PR #4765, opened 2017-07-14
Adding in permissions to edit and admin cluster roles

ewolinetz:
Should address installer side of https://bugzilla.redhat.com/show_bug.cgi?id=1471040

There isn't a way to easily add to the current roles using modules (they would replace rather than be additive), hence the round about manner. Should be updated to use `oc_clusterrole` when it can do this.

CC @spadgett @bparees

### DIFF
--- a/roles/openshift_service_catalog/tasks/install.yml
+++ b/roles/openshift_service_catalog/tasks/install.yml
@@ -66,6 +66,40 @@
     template_name: kube-system-service-catalog
     namespace: kube-system
 
+- oc_obj:
+    name: edit
+    kind: clusterrole
+    state: list
+  register: edit_yaml
+
+- name: Generate apply template for clusterrole/edit
+  template:
+    src: sc_role_patching.j2
+    dest: "{{ mktemp.stdout }}/edit_sc_patch.yml"
+  vars:
+    original_content: "{{ edit_yaml.results.results[0] | to_yaml }}"
+
+- name: update edit role for service catalog and pod preset access
+  command: >
+    oc apply -f {{ mktemp.stdout }}/edit_sc_patch.yml
+
+- oc_obj:
+    name: admin
+    kind: clusterrole
+    state: list
+  register: admin_yaml
+
+- name: Generate apply template for clusterrole/admin
+  template:
+    src: sc_role_patching.j2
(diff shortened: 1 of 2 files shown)


roles/openshift_service_catalog/templates/sc_role_patching.j2
```diff
@@ -0,0 +1,26 @@
+{{ original_content }}
```
sdodson: clever :-)

roles/openshift_service_catalog/tasks/install.yml
```diff
@@ -66,6 +66,40 @@
...
+- name: Generate apply template for clusterrole/edit
+  template:
+    src: sc_role_patching.j2
+    dest: "{{ mktemp.stdout }}/edit_sc_patch.yml"
+  vars:
+    original_content: "{{ edit_yaml.results.results[0] | to_yaml }}"
+
+- name: update edit role for service catalog and pod preset access
+  command: >
+    oc apply -f {{ mktemp.stdout }}/edit_sc_patch.yml
```
liggitt: I don't think we want to start using apply with reconciled roles... I'm concerned it will try to do clever things and remove permissions.

liggitt: if you're already fetching the current object to build the patch, can you do `oc replace` with the result instead of `oc apply`?

liggitt: also, doesn't this mean that re-running the task will tack on this permission multiple times?

ewolinetz: i hadn't heard of `oc replace` before, I can investigate using that. in regards to rerunning it, wouldnt `oc apply` resolve duplicate entries (perhaps i incorrectly assumed it would)?

liggitt: `wouldnt oc apply resolve duplicate entries`

definitely not, and it stores a copy of the current object json in an annotation and tries to do fancy three-way diffs against future `apply` calls. I do not recommend it for this object

ewolinetz: updated to use `replace` and conditionally add perms in https://github.com/openshift/openshift-ansible/pull/4772

PR #4772, opened 2017-07-17
Updating to use oc replace and conditionally update edit and admin roles

ewolinetz:
Updating per comments on #4765 from @liggitt

### DIFF
--- a/roles/openshift_service_catalog/tasks/install.yml
+++ b/roles/openshift_service_catalog/tasks/install.yml
@@ -72,16 +72,22 @@
     state: list
   register: edit_yaml
 
+# only do this if we don't already have the updated role info
 - name: Generate apply template for clusterrole/edit
   template:
     src: sc_role_patching.j2
     dest: "{{ mktemp.stdout }}/edit_sc_patch.yml"
   vars:
     original_content: "{{ edit_yaml.results.results[0] | to_yaml }}"
+  when:
+  - not edit_yaml.results.results[0] | oo_contains_rule(['servicecatalog.k8s.io'], ['instances', 'bindings'], ['create', 'update', 'delete', 'get', 'list', 'watch']) or not edit_yaml.results.results[0] | oo_contains_rule(['settings.k8s.io'], ['podpresets'], ['create', 'update', 'delete', 'get', 'list', 'watch'])
 
+# only do this if we don't already have the updated role info
 - name: update edit role for service catalog and pod preset access
   command: >
-    oc apply -f {{ mktemp.stdout }}/edit_sc_patch.yml
+    oc replace -f {{ mktemp.stdout }}/edit_sc_patch.yml
+  when:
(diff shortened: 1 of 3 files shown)


roles/openshift_service_catalog/tasks/install.yml
```diff
@@ -72,33 +72,49 @@
...
+# only do this if we don't already have the updated role info
 - name: Generate apply template for clusterrole/edit
   template:
     src: sc_role_patching.j2
     dest: "{{ mktemp.stdout }}/edit_sc_patch.yml"
   vars:
     original_content: "{{ edit_yaml.results.results[0] | to_yaml }}"
+  when:
+  - '"servicecatalog.k8s.io" not in edit_yaml.results.results[0]| to_yaml'
+  - '"settings.k8s.io" not in edit_yaml.results.results[0]| to_yaml'
```
liggitt: these checks seem pretty fragile, especially the quoting bits and the assumption that anything present from these groups means all intended permissions are present. we don't have a great tool for doing covers checks though...

ewolinetz: would there be a way we could do a `oc policy can-i --as` to confirm that these permissions are available? or would we need a user to be set up first; that might be a better way to verify that these permissions are available yet

ewolinetz:
@liggitt created a filter plugin to evaluate if we have the specific rule contained already within the clusterrole

ewolinetz:
Added commit to resolve https://bugzilla.redhat.com/show_bug.cgi?id=1471881

ewolinetz:
@sdodson can you review the filter plugin and make sure this PR is still up to par?

roles/openshift_service_catalog/filter_plugins/contains_rule.py
```diff
@@ -0,0 +1,29 @@
...
+    '''Return true if the specified rule is contained within the provided source'''
+
+    rules=source['rules']
+
+    if rules:
+        for rule in rules:
+            if rule['apiGroups'].sort() == apiGroups.sort():
+                if rule['resources'].sort() == resources.sort():
+                    if rule['verbs'].sort() == verbs.sort():
+                        if rule['attributeRestrictions'] == attributeRestrictions:
```
liggitt: can drop attributeRestrictions, it's been deprecated and is no longer used

liggitt:
thanks, detection + replace make me feel better about this

sdodson:
@ewolinetz looks fine to me, but also looks like this isn't specific to service catalog, should we move this to the top level filter_plugins directory for general use?

ewolinetz:
@sdodson will do

sdodson:
aos-ci-test

ewolinetz:
@sdodson  hold off on merging this, i think my `when` is incorrect... i dont see it performing the replace for the roles

ewolinetz:
aos-ci-test

ewolinetz:
resolved all travis issues, resolved issue was seeing where it wasn't correctly detecting that the roles weren't added yet

sdodson:
@jlebon know what's happening in the fedora test? looks like a yum error on updates repo?

jlebon:
@sdodson Looks like just regular flaky infra. You can always "bot, retest this please" if it happens.

PR #4773, opened 2017-07-17
fix missing console appending in logging

jcantrill:
bring https://github.com/openshift/openshift-ansible/pull/3777 into refactor of logging roles

### DIFF
--- a/roles/openshift_logging_elasticsearch/tasks/main.yaml
+++ b/roles/openshift_logging_elasticsearch/tasks/main.yaml
@@ -107,9 +107,16 @@
     - openshift_logging_elasticsearch_kibana_index_mode in __kibana_index_modes
     msg: "The openshift_logging_elasticsearch_kibana_index_mode '{{ openshift_logging_elasticsearch_kibana_index_mode  }}' only supports one of: {{ __kibana_index_modes | join(', ') }}"
 
+- assert:
+    that:
+    - "{{ openshift_logging_es_log_appenders | length > 0 }}"
+    msg: "The openshift_logging_es_log_appenders '{{ openshift_logging_es_log_appenders }}' has an unrecognized option and only supports the following as a list: {{ __es_log_appenders | join(', ') }}"
+
 - template:
     src: elasticsearch-logging.yml.j2
     dest: "{{ tempdir }}/elasticsearch-logging.yml"
+  vars:
+    root_logger: "{{openshift_logging_es_log_appenders | join(', ')}}"
   when: es_logging_contents is undefined
   changed_when: no
 
--- a/roles/openshift_logging_elasticsearch/vars/main.yml
+++ b/roles/openshift_logging_elasticsearch/vars/main.yml
@@ -2,7 +2,7 @@
 __latest_es_version: "3_5"
 __allowed_es_versions: ["3_5", "3_6"]
 __allowed_es_types: ["data-master", "data-client", "master", "client"]
-
+__es_log_appenders: ['file', 'console']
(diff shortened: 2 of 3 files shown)


jcantrill:
[aos-ci-test]

jcantrill:
aos-ci-test

roles/openshift_logging_elasticsearch/tasks/main.yaml
```diff
@@ -100,13 +100,22 @@
     files:
     - "{{ tempdir }}/logging-elasticsearch-view-role.yaml"
     delete_after: true
-
 # configmap
+- fail:
```
ewolinetz: can we change this from a `fail` task to an `assert` task? we should eventually convert all the cases of it in the role... this would make one less to convert later

roles/openshift_logging_elasticsearch/tasks/main.yaml
```diff
@@ -100,13 +100,22 @@
...
+    msg: "The openshift_logging_es_log_appenders '{{openshift_logging_es_log_appenders}}' has an unrecognized option and only supports the following as a list: {{es_log_appenders | join(', ')}}"
+  when:
+    - es_logging_contents is undefined
+    - "{{ openshift_logging_es_log_appenders | list | difference(es_log_appenders) | length != 0 }}"
```
ewolinetz: where is `es_log_appenders` coming from? it doesn't seem to be used anywhere else here...
Is there a more simple way to evaluate this?

jcantrill: I believe es_log_appenders is a list in vars/main.yaml.  This is how it was done in the original role, and I believe we added it as a way to ensure you defined atleast one appender

ewolinetz: so is `openshift_logging_es_log_appenders` set as a fact based on that list? couldn't we just evaluate the length of the `openshift_logging_es_log_appenders` list?

jcantrill: Recalled why we choose this approach.  The desire is to restrict them to the options of file and console.

ewolinetz: couldn't we just have a separate assert that the provided options are valid? i think it would be better to fail and have the customer know they did something wrong that be surprised later to find out we did something they didn't plan on

roles/openshift_logging_elasticsearch/vars/main.yml
```diff
@@ -2,6 +2,7 @@
 __latest_es_version: "3_5"
 __allowed_es_versions: ["3_5", "3_6"]
 __allowed_es_types: ["data-master", "data-client", "master", "client"]
+es_log_appenders: ['file', 'console']
```
ewolinetz: `__es_log_appenders`

ewolinetz:
aos-ci-test

ewolinetz:
won't block on this, but if you wouldn't mind formatting variables so spacing is more like `{{ openshift_logging_es_log_appenders | join(', ') }}` where vars are being used

jcantrill:
aos-ci-test

jcantrill:
[merge]

jcantrill:
[merge][severity:blocker]

jcantrill:
aos-ci-test

jcantrill:
aos-ci-test

jcantrill:
aos-ci-test

jcantrill:
[merge]

PR #4780, opened 2017-07-18
Update image-gc-high-threshold value

derekwaynecarr:
A value of 90 may never be crossed when using device mapper.

Update the example to have the default value used in OpenShift as 85.


sdodson:
This is a docs only change, merging.

PR #4783, opened 2017-07-18
1471973- default to bootstrapping the broker on startup

fabianvf:
https://bugzilla.redhat.com/show_bug.cgi?id=1471973

This will cause the broker to discover apbs when it comes up, rather than requiring the user to do a manual bootstrap request.


eriknelson:
LGTM

ewolinetz:
we'll also need to cherry pick this to the 3.6 branch

PR #4788, opened 2017-07-19
Allow to configure oreg_url specifically for node or master

hansmi:
This commit allows to specify imageConfig.format specifically for master
or for nodes.

One use case of this could be if you want to use customer builder
images. In this case imageConfig.format only needs to be changed in the
master-config.yml but not in the node-config.yml.

Backported from commit 0ee8128 in master.


sdodson:
bot, test pull request once

sdodson:
aos-ci-test

PR #4797, opened 2017-07-19
Refactor the firewall workflow.

kwoodson:


kwoodson:
In this particular PR, I wanted to focus on the way we handle firewall rules.  The PR appears to have a lot of files added/modified but I followed a pattern for each role.  I'll explain my train of thought.

Old pattern:
The current pattern calls out to `os_firewall` inside of `meta/main.yml` and sets the iptables/firewalld ports.  This is a role call that includes the proper firewall technology that is being used as well as its dependencies (openshift_facts).  It **also** `SKIPS` the technology that it is not being used.  This is a major problem for slowness and skips when running openshift-ansible.  Skips are one of the slowest factors when running openshift-ansible with each node have around ~1000 skips and the master ~3000.

New Pattern:
The new suggested pattern in this PR is to remove the module code (`/library`) out of `os_firewall` and place it in a new role as `lib_os_firewall`.  This will reduce the need to continually call it and its dependencies since modules can be called as libraries and reduce the skipped tasks.  Inside of each role that was setting firewall rules I modified the following files:
-  meta/main.yml - removed os_firewall and added lib_os_firewall
-  defaults/main.yml - added firewall rules to the defaults so they can be overridden if we need.
-  tasks/firewall.yml - to keep the firewall logic to a single location
-  tasks/main.yml - include the firewall.yml so we get the firewall ports code

Overall, I feel this pattern simplifies the firewall changes by placing them in each of the roles defaults and specifically in the firewall.yml.  Then we reduce the number of skips that occur from having to continually skip the technology that isn't being used.  In testing I saw approximately 15% less skips on all nodes/non-first-masters and for the first master over 500 less skips.  I also saw around a 4:25 second speed up on same hardware/cluster.  

Downsides:
-  There is some duplicate code now in each role for performing the firewall updates.  It would be great if we could overcome this.  I feel that in the future when we move to firewalld then we can eliminate a lot of the duplicate code.
-  We need to ensure that `os_firewall` is called in the beginning of playbooks so that it can set the right iptables/firewalld setup.
Upsides:
- Improved speed
- Reduced dependencies
- Less skips
- Non-hard coded firewall rules in dependencies
- overridable defaults

Comments are welcome.

roles/nuage_master/tasks/firewall.yml
```diff
@@ -0,0 +1,40 @@
+---
+- when: os_firewall_enabled | bool and not os_firewall_use_firewalld | bool
+  block:
+  - name: Add iptables allow rules
+    os_firewall_manage_iptables:
+      name: "{{ item.service }}"
+      action: add
+      protocol: "{{ item.port.split('/')[1] }}"
```
ashcrow: nit: Calling on it once here ... I think it would be cleaner to separate protocol and port in the defaults. This would make it easier on the eyes. Anywhere that needs ``proto/port`` would read better as ``{{ $ITEM_protocol }}/{{ $ITEM_PORT }}`` as well.

roles/openshift_hosted/tasks/registry/registry.yml
```diff
@@ -1,6 +1,10 @@
 ---
-- block:
+- name: setup firewall
+  include: firewall.yml
+  static: yes
 
+- when: openshift.hosted.registry.replicas | default(none) is none
```
ashcrow: 👍 I'm all for top ``when``s on ``block``s

roles/nuage_master/tasks/firewall.yml
kwoodson: @ashcrow, @tbielawa, This syntax is what we currently have and the reason we keep it this way is so that we are compatible with the `firewalld` and our own module `os_firewall_manage_iptables`.

I think the work to refactor that could be done in a separate PR.  I wanted to minimize the change set.  If you feel strongly we can do that work in this PR.

http://docs.ansible.com/ansible/latest/firewalld_module.html

ashcrow: I'm fine with pushing it another PR.

kwoodson:
aos-ci-test

ashcrow:
```
    "msg": "The conditional check 'item.cond' failed. The error was: error while evaluating conditional (item.cond): 'dict object' has no attribute 'cond'\n\nThe error appears to have been in '<snip>/roles/openshift_node/tasks/firewall.yml': line 4, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  block:\n  - name: Add iptables allow rules\n    ^ here\n"
```

kwoodson:
aos-ci-test

kwoodson:
@ashcrow, fixed the errors we were seeing earlier.

kwoodson:
@tbielawa, @ashcrow, @mtnbikenc, @sdodson 

Any of you fine gentlemen willing to give me your approval?  If we want more discussion I can wait and bring this up in the architecture meeting.  I feel at this point that this PR improves our ability to apply firewall rules, reduces skips, and allows us to have a pattern for rules going forward.  I would like to get it merged. 

Thanks.

ashcrow:
@kwoodson I don't see an issue with this but I still don't feel like I have a good enough handle of our firewall module compared to folks like @mtnbikenc.

playbooks/common/openshift-cluster/config.yml
```diff
@@ -26,6 +26,21 @@
   tags:
   - always
 
+- name: Setup firewall
+  hosts: oo_all_hosts
+  tags:
+  - always
+  tasks:
+  # This should move to intialize_facts
+  - name: set os_firewall_enabled
```
mtnbikenc: Why are these fact being set here?  Is there an issue with how the defaults from the role are being applied?

playbooks/common/openshift-cluster/config.yml
```diff
@@ -26,6 +26,21 @@
...
+    set_fact:
+      os_firewall_enabled: true
+      os_firewall_use_firewalld: false
+
+  - name: Set proper firewall settings
+    include_role:
+      name: os_firewall
```
mtnbikenc: With the function of the os_firewall role being changed from installing/configuring the firewall AND managing rules, the README should be updated to reflect the new function of just installation/configuration.

roles/cockpit/meta/main.yml
```diff
@@ -12,7 +12,4 @@ galaxy_info:
   categories:
   - cloud
 dependencies:
-- role: os_firewall
-  os_firewall_allow:
-  - service: cockpit-ws
-    port: 9090/tcp
+- role: lib_os_firewall
```
mtnbikenc: Stating once here:
I'd rather we move to having a single role with our modules than having each in its own lib_<module_name> role.  This is not an issue to hold up this PR, just something to consider.

playbooks/common/openshift-cluster/config.yml
kwoodson: You are correct.  The role defaults are sufficient.  I will remove these.

playbooks/common/openshift-cluster/config.yml
kwoodson: Will update the doc.

roles/cockpit/meta/main.yml
kwoodson: Let's set this work for a refactor.  The README is where the documentation for this module lives.  I'll leave it here until we can merge the README doc into the module doc.

playbooks/common/openshift-cluster/config.yml
kwoodson: @mtnbikenc, instead of executing the code here in openshift-cluster/config.yml, is this a candidate for std_include.yml?  If not, we need a place to initialize this a single time.

Thoughts?

kwoodson:
@mtnbikenc, please review the latest.  Specifically std_include.yml update.

mtnbikenc:
Just need to fix the yamllint.

kwoodson:
aos-ci-test

kwoodson:
aos-ci-test

kwoodson:
aos-ci-test

kwoodson:
aos-ci-test

kwoodson:
[merge]

kwoodson:

PR #4799, opened 2017-07-19
[3.6] Add etcd exports to openshift_storage_nfs

abutcher:
https://bugzilla.redhat.com/show_bug.cgi?id=1467574

### DIFF
--- a/roles/openshift_storage_nfs/tasks/main.yml
+++ b/roles/openshift_storage_nfs/tasks/main.yml
@@ -30,7 +30,7 @@
     - "{{ openshift.hosted.metrics }}"
     - "{{ openshift.hosted.logging }}"
     - "{{ openshift.hosted.loggingops }}"
-
+    - "{{ openshift.hosted.etcd }}"
 
 - name: Configure exports
   template:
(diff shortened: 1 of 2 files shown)


sdodson:
aos-ci-test

fabianvf:
can we make it `openshift.hosted.asb_etcd`? That should help with confusion about broker vs master etcd.

sdodson:
> can we make it openshift.hosted.asb_etcd? That should help with confusion about broker vs master etcd.

or other future etcd hosted volumes.

Really though, the fact that this got so close to working without additional changes is just a happy coincidence. Before merging this and upstream PR we should make sure to mention these variables in inventory examples and also highlight that the proper production method is to use dynamic volumes and that the nfs provisioning is only appropriate for POC use cases.

abutcher:
> can we make it openshift.hosted.asb_etcd? That should help with confusion about broker vs master etcd.

I'll try to get this done but it looks like it will require some refactoring in `openshift_facts`.

sdodson:
Per IRC discussion we'll take this as is and clean it before service catalog moves from tech preview to fully supported.

PR #4802, opened 2017-07-20
Fix for : https://bugzilla.redhat.com/show_bug.cgi?id=1467423

jkaurredhat:
Customer confirmed changing KUBERNETES_MASTER_URL and env parameter MASTER_URL value from default value "https://kubernetes.default.svc.cluster.local" to "https://kubernetes.default.svc" solved the issue.

### DIFF
--- a/roles/openshift_metrics/defaults/main.yaml
+++ b/roles/openshift_metrics/defaults/main.yaml
@@ -49,7 +49,7 @@ openshift_metrics_resolution: 30s
 # overriding the values here
 #####
 
-openshift_metrics_master_url: https://kubernetes.default.svc.cluster.local
+openshift_metrics_master_url: https://kubernetes.default.svc
 openshift_metrics_node_id: nodename
 openshift_metrics_project: openshift-infra
 


sdodson:
This should be fine but it should also be un-necessary, can you provide any example errors to support this change?

sdodson:
@mwringe any ideas? you were tagged on the backport pr. I'd like to make sure we address master branch first before backporting.

jkaurredhat:
If I check metrics code I see that metrics pods are using the kubernetes API available at https://kubernetes.default.svc:443

Github docs for metrics deployment support this point

https://github.com/openshift/origin-metrics/blob/bd54fbe33d340c90301a47097a94f9aabb7ce2a4/docs/deployer_installation.adoc#deployer-template

https://github.com/openshift/origin-metrics/blob/665194b5c3e022dc12c938d36623a67bba63d4f0/hawkular-metrics/hawkular-metrics-wrapper.sh#L68
https://github.com/openshift/origin-metrics/blob/7c43f4d1ac6076b88e0926b462a48c5e2cf92017/deployer/templates/heapster-standalone.yaml#L12

mwringe:
@sdodson https://kubernetes.default.svc/ should be the more correct url. The 'cluster.local' part is modifiable at the cluster level and it shouldn't be assumed that its always going to be this value.

This should be changed to https://kubernetes.default.svc/ at least in master. For 1.6 if its still possible to change it, its probably a good idea to do so.

I don't view this as a high priority item, https://kubernetes.default.svc.cluster.local/ should still work for the majority of users and for those where it doesn't work, they can configure their own value.

sdodson:
aos-ci-test

PR #4807, opened 2017-07-20
Fix broken link to Docker image instructions

rhcarvalho:
And document that one needs to install dependencies before running playbooks.

### DIFF
--- a/playbooks/byo/openshift-checks/README.md
+++ b/playbooks/byo/openshift-checks/README.md
@@ -39,7 +39,9 @@ against your inventory file. Here is the step-by-step:
     $ cd openshift-ansible
     ```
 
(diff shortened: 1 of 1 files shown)


rhcarvalho:
@sdodson @sosiouxme PTAL, I noticed the link to the container image README was missing an extra `../`. Since touching it, also added a note about dependencies.

rhcarvalho:
@sdodson could we get this in without aos-ci-test?

sdodson:
sure

sosiouxme:
Good catch, thanks.

Since it can now be installed as a system container we don't even have to limit it to "anywhere docker runs" :) I should probably take a pass at origin docs for various tweaks like that...

PR #4819, opened 2017-07-20
[WIP] Elasticsearch 5 standalone ansible role.

t0ffel:
Here is my idea of a separate `openshift_elasticsearch` role for deployment of Elasticsearch 5
on OpenShift. It should be used in combination with `openshift_logging` role, though it is not integrated yet. This way hopefully we can transition smoothly to ES5, while keeping ES2.
I'll be submitting a companion pr to `origin_aggregated_logging` soon.

The role support 2 different node types: master and client-data.
The SearchGuard initialization is done via separate elasticsearch-init pod(this is still a TODO item).

/cc @richm @jcantrill @portante @ewolinetz

### DIFF
--- a/roles/openshift_elasticsearch/README.md
+++ b/roles/openshift_elasticsearch/README.md
@@ -1,0 +1,70 @@
+# The role deploys Elasticsearch cluster on top of OpenShift
+
+## Topology
+
+The deployment supports 2 kinds of node roles at the moment:
+* master
+* data-client
+
+We explicitly don't support combining masters with any other node type.
+`es_node_topology` variable holds the description of the cluster node
+topology.
+
+### Masters
+
+`es_node_topology.masters` describes how many masters will run. Single
+DeploymentConfig will be created for all masters.
+`replicas` field define how many masters to spin up (replicas in Kubernetes
+notation).
+`limits` define Kubernetes pod limits
+`nodeSelector` define Kuberenetes pod selector
+
+### Client Data nodes
+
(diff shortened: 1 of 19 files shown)


ewolinetz:
is there a reason we couldn't contain this within the current `openshift_logging_elasticsearch` role and just allow a means to specify the ES version? We could then contain templates and files for both versions by using a version folder structure. We could also leverage the currently used process for loading in variable defaults as we do between enterprise and origin installations but for the two different versions.

I'm slightly adversed to creating yet another role for ES, especially if ES 2.x and 5.x need to be supported at the same time for only a few versions. To me it seems the logic to configure is similar but the specs may be different between the two.

jcantrill:
I agree with @ewolinetz sentiments.  We recently went from a single role to component roles so that we could modify the deployment topology.  I do not think it is in our best interest to support yet another role simply because it is a version bump.  I know there is advisement that we should be backwards compatible but we need to move all these changes into the existing roles.  I suspect the other components will need modifications too.  I do not wish to also make 'versioned' roles for them as well.  We would better served, IMO, to have a feature branch that we regularly sync with master until we can cut it over.

ewolinetz:
I think I would prefer having a versioned role as opposed to syncing over periodically since fixes/features may be lost doing that. We should be able to provide configurations for both an ES 5.x and ES 2.x in the same role at the same time and be able to provide a parameter to toggle between the two, defaulting to ES 2.x. Especially if the plan moving forward is to support the prior 2 versions.

t0ffel:
abandoned in favor of https://github.com/openshift/openshift-ansible/pull/4986

PR #4213, opened 2017-05-17
node, systemd: change Requires to Wants for openvswitch

giuseppe:
Sometimes the node container is not started on a container-engine
restart.  Use a weaker dependency on openvswitch that is causing this issue

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1451192

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

PR #4820, opened 2017-07-20
node: make node service PartOf=openvswitch.service when openshift-sdn is used

dcbw:
This reverts commit 7f805f9a0c41477365dd88b0ac73f0d221bd654a.

The commit causes the behavior seen in
https://bugzilla.redhat.com/show_bug.cgi?id=1453113 because openshift-node
is no longer restarted when openvswitch is.

@giuseppe @sdodson @knobunc 

RE https://github.com/openshift/openshift-ansible/pull/4213 can we get a more detailed explanation of why the various dependencies are not being restarted correctly?

### DIFF
--- a/roles/openshift_node/templates/openshift.docker.node.service
+++ b/roles/openshift_node/templates/openshift.docker.node.service
@@ -6,6 +6,7 @@ PartOf={{ openshift.docker.service_name }}.service
 Requires={{ openshift.docker.service_name }}.service
 {% if openshift_node_use_openshift_sdn %}
 Wants=openvswitch.service
+PartOf=openvswitch.service
 After=ovsdb-server.service
 After=ovs-vswitchd.service
 {% endif %}
--- a/roles/openshift_node_upgrade/templates/openshift.docker.node.service
+++ b/roles/openshift_node_upgrade/templates/openshift.docker.node.service
@@ -6,6 +6,7 @@ PartOf={{ openshift.docker.service_name }}.service
 Requires={{ openshift.docker.service_name }}.service
 {% if openshift_use_openshift_sdn %}
 Wants=openvswitch.service
+PartOf=openvswitch.service
 After=ovsdb-server.service
 After=ovs-vswitchd.service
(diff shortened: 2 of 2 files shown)


dcbw:
Though given the issues in https://github.com/systemd/systemd/issues/1312 perhaps we do need Wants + some other way to ensure that openshift-node is restarted when openvswitch is...  not sure what that should be, unless we do something like watch D-Bus for systemd events and then os.Exit() when we see openvswitch being stopped.

sdodson:
Seems like this may be best and hopefully addresses the problem @giuseppe was fixing?

Requries=openvswitch.service
PartOf=openvswitch.service

giuseppe:
does `Want=openvswitch.service + PartOf=openvswitch.service` fix this issue?  Otherwise I am afraid we will reintroduce https://bugzilla.redhat.com/show_bug.cgi?id=1451192 which was addressed by 7f805f9

sdodson:
@giuseppe can you validate if wants + partof works both the bug you were fixing with your previous change and the bug @dcbw is trying to address here?

giuseppe:
@sdodson from my tests it seems `wants+partof`addresses both issues.

I could not reproduce the first issue when docker is restarted (and I could see it only once in 20-25 attempts with "Requires" so it was quite hard to reproduce anyway) and `wants+partof` also ensures that `node` is restarted after `openvswitch` so the second issue should be addressed as well.

giuseppe:
@dcbw could you update this patch to use `Wants=openvswitch.service` and `PartOf=openvswitch.service`?

dcbw:
@giuseppe done! PTAL thanks.

dcbw:
@sdodson the containerized test keeps failing, and looking at the logs it's that the master-controllers service can't start.  However, there master-controllers journal output in S3 artifacts is basically empty.  Is this a general flake?  I can't see how this PR could affect that.

dcbw:
Rebased to see if it fixes the containerized test...

PR #4826, opened 2017-07-21
GlusterFS fixes 6

jarrpa:
Bug fixes in registry setup and boolean parsing.


sdodson:
aos-ci-test

sdodson:
[merge]

playbooks/common/openshift-glusterfs/config.yml
```diff
@@ -1,6 +1,6 @@
 ---
-- name: Open firewall ports for GlusterFS
-  hosts: oo_glusterfs_to_config
+- name: Open firewall ports for GlusterFS nodes
+  hosts: glusterfs
```
sdodson: This shouldn't be necessary should it?

jarrpa: I'm accounting for the possibility that both `[glusterfs]` and `[gluster_registry]` are specified but that one is native while the other is external.

sdodson: My concern is that we're using byo groups here and we should be using oo_* groups exclusively in playbooks/common but it looks like we'd need to do a bit of refactoring to get us there based on the group mapping logic that sets gluster_registry = glusterfs hosts if gluster_registry group is empty. I'll open an issue for us to track this but we'll take it as is for now.

sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
flake on https://github.com/openshift/origin/issues/15356 in https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/726/

PR #4832, opened 2017-07-22
Use openshift-master-api and openshift-master-controllers setup always

smarterclayton:
The installer now always configures two services on the master for the api server and the controllers. Native clustering is the default configuration mode, even when only one master is configured. A warning is added if there are no etcd groups configured. Finally, the new client based leader election code in the controllers is used as the default, which removes the need for the controllers to have access to etcd.

This will set the stage for running controllers and apiservers as static pods.

@sdodson


smarterclayton:
aos-ci-test

smarterclayton:
Changes required to support this in upgrades:

1. stop and remove the old origin-master.service and sysconfig files if they exist
2. error / warn if the user had those files that scripts and other automation may need to change.

smarterclayton:
This does not include the change to move controllers to static pods because I realized we need to deal with some fundamental ordering problems before that (the node needs to be configured and ready ahead of time, instead of masters first).  Requires the work @kwoodson will be doing to split out node preparation from node config - once that is set up, we may actually want to run node prep on masters, then remove the service units for *both* api and controller and replace them with static pods, and use the kubelet in run-once mode as well.  I don't see much value in system containers for api or controllers, but there is a lot of value in those as static pods, especially with crio present

smarterclayton:
aos-ci-test

smarterclayton:
Tsk tsk, jobs not setting etcd hosts

playbooks/common/openshift-cluster/config.yml
```diff
@@ -18,6 +18,12 @@
       - docker_image_availability
       - docker_storage
 
+- hosts: localhost
+  tasks:
+  - fail:
+      msg: No etcd hosts defined. Running an all-in-one master is deprecated and will no longer be supported in a future upgrade.
+    when: groups.oo_etcd_to_config | default([]) | length == 0 and not openshift_master_unsupported_all_in_one | default(False)
```
tbielawa: Please put `AND`d `when` statements on multiple lines in a list format, e.g. `groups.oo_etcd_to_config | default([]) | length == 0 and not openshift_master_unsupported_all_in_one | default(False)` should be more like this:

```yaml
    when:
      - groups.oo_etcd_to_config | default([]) | length == 0
      - not openshift_master_unsupported_all_in_one | default(False)
```

roles/openshift_cfme/handlers/main.yml
```diff
@@ -6,19 +6,14 @@
...
 
-- name: restart master
-  systemd: name={{ openshift.common.service_type }}-master state=restarted
-  when: (openshift.master.ha is not defined or not openshift.master.ha | bool) and (not (master_service_status_changed | default(false) | bool))
-  notify: Verify API Server
-
 - name: restart master api
   systemd: name={{ openshift.common.service_type }}-master-api state=restarted
-  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  when: (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
```
tbielawa: multi-line `WHEN`

roles/openshift_cfme/handlers/main.yml
```diff
@@ -6,19 +6,14 @@
...
   notify: Verify API Server
 
 - name: restart master controllers
   systemd: name={{ openshift.common.service_type }}-master-controllers state=restarted
-  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  when: (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
```
tbielawa: multi-line `WHEN`

roles/openshift_hosted_metrics/handlers/main.yml
```diff
@@ -1,17 +1,12 @@
 ---
-- name: restart master
-  systemd: name={{ openshift.common.service_type }}-master state=restarted
-  when: (openshift.master.ha is not defined or not openshift.master.ha | bool) and (not (master_service_status_changed | default(false) | bool))
-  notify: Verify API Server
-
 - name: restart master api
   systemd: name={{ openshift.common.service_type }}-master-api state=restarted
-  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  when: (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
```
tbielawa: multi-line `WHEN`

roles/openshift_hosted_metrics/handlers/main.yml
```diff
@@ -1,17 +1,12 @@
...
   notify: Verify API Server
 
 - name: restart master controllers
   systemd: name={{ openshift.common.service_type }}-master-controllers state=restarted
-  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  when: (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
```
tbielawa: multi-line `WHEN`

roles/openshift_logging/handlers/main.yml
```diff
@@ -1,17 +1,12 @@
 ---
-- name: restart master
-  systemd: name={{ openshift.common.service_type }}-master state=restarted
-  when: (openshift.master.ha is not defined or not openshift.master.ha | bool) and (not (master_service_status_changed | default(false) | bool))
-  notify: Verify API Server
-
 - name: restart master api
   systemd: name={{ openshift.common.service_type }}-master-api state=restarted
-  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  when: (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
```
tbielawa: multi-line `WHEN`

roles/openshift_logging/handlers/main.yml
```diff
@@ -1,17 +1,12 @@
...
   notify: Verify API Server
 
 - name: restart master controllers
   systemd: name={{ openshift.common.service_type }}-master-controllers state=restarted
-  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  when: (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
```
tbielawa: multi-line `WHEN`

roles/openshift_master/handlers/main.yml
```diff
@@ -1,17 +1,12 @@
 ---
-- name: restart master
-  systemd: name={{ openshift.common.service_type }}-master state=restarted
-  when: (openshift.master.ha is not defined or not openshift.master.ha | bool) and (not (master_service_status_changed | default(false) | bool))
-  notify: Verify API Server
-
 - name: restart master api
   systemd: name={{ openshift.common.service_type }}-master-api state=restarted
-  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  when: (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
```
tbielawa: multi-line `WHEN`

roles/openshift_master/handlers/main.yml
```diff
@@ -1,17 +1,12 @@
...
   notify: Verify API Server
 
 - name: restart master controllers
   systemd: name={{ openshift.common.service_type }}-master-controllers state=restarted
-  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  when: (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -12,16 +12,16 @@
 # HA Variable Validation
 - fail:
     msg: "openshift_master_cluster_method must be set to either 'native' or 'pacemaker' for multi-master installations"
-  when: openshift_master_ha | bool and ((openshift_master_cluster_method is not defined) or (openshift_master_cluster_method is defined and openshift_master_cluster_method not in ["native", "pacemaker"]))
+  when: openshift.master.ha | bool and ((openshift.master.cluster_method is not defined) or (openshift.master.cluster_method is defined and openshift.master.cluster_method not in ["native", "pacemaker"]))
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -12,16 +12,16 @@
...
 - fail:
     msg: "'native' high availability is not supported for the requested OpenShift version"
-  when: openshift_master_ha | bool and openshift_master_cluster_method == "native" and not openshift.common.version_gte_3_1_or_1_1 | bool
+  when: openshift.master.ha | bool and openshift.master.cluster_method == "native" and not openshift.common.version_gte_3_1_or_1_1 | bool
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -12,16 +12,16 @@
...
 - fail:
     msg: "openshift_master_cluster_password must be set for multi-master installations"
-  when: openshift_master_ha | bool and openshift_master_cluster_method == "pacemaker" and (openshift_master_cluster_password is not defined or not openshift_master_cluster_password)
+  when: openshift.master.ha | bool and openshift.master.cluster_method == "pacemaker" and (openshift_master_cluster_password is not defined or not openshift_master_cluster_password)
 - fail:
     msg: "Pacemaker based HA is not supported at this time when used with containerized installs"
-  when: openshift_master_ha | bool and openshift_master_cluster_method == "pacemaker" and openshift.common.is_containerized | bool
+  when: openshift.master.ha | bool and openshift.master.cluster_method == "pacemaker" and openshift.common.is_containerized | bool
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -12,16 +12,16 @@
...
+  when: openshift.master.ha | bool and ((openshift.master.cluster_method is not defined) or (openshift.master.cluster_method is defined and openshift.master.cluster_method not in ["native", "pacemaker"]))
 - fail:
     msg: "'native' high availability is not supported for the requested OpenShift version"
-  when: openshift_master_ha | bool and openshift_master_cluster_method == "native" and not openshift.common.version_gte_3_1_or_1_1 | bool
+  when: openshift.master.ha | bool and openshift.master.cluster_method == "native" and not openshift.common.version_gte_3_1_or_1_1 | bool
 - fail:
     msg: "openshift_master_cluster_password must be set for multi-master installations"
-  when: openshift_master_ha | bool and openshift_master_cluster_method == "pacemaker" and (openshift_master_cluster_password is not defined or not openshift_master_cluster_password)
+  when: openshift.master.ha | bool and openshift.master.cluster_method == "pacemaker" and (openshift_master_cluster_password is not defined or not openshift_master_cluster_password)
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -172,89 +169,45 @@
...
-    openshift.master.cluster_method == 'native' and
-    not openshift.common.is_containerized | bool
-
 - name: Start and enable master api on first master
   systemd:
     name: "{{ openshift.common.service_type }}-master-api"
     enabled: yes
     state: started
-  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native' and inventory_hostname == openshift_master_hosts[0]
+  when: openshift.master.cluster_method == 'native' and inventory_hostname == openshift_master_hosts[0]
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -172,89 +169,45 @@
...
   register: start_result
   until: not start_result | failed
   retries: 1
   delay: 60
 
 - set_fact:
     master_api_service_status_changed: "{{ start_result | changed }}"
-  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native' and inventory_hostname == openshift_master_hosts[0]
+  when: openshift.master.cluster_method == 'native' and inventory_hostname == openshift_master_hosts[0]
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -172,89 +169,45 @@
...
-  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native'
+  when: openshift.master.ha | bool and openshift.master.cluster_method == 'native'
 
 - name: Start and enable master api all masters
   systemd:
     name: "{{ openshift.common.service_type }}-master-api"
     enabled: yes
     state: started
-  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native' and inventory_hostname != openshift_master_hosts[0]
+  when: openshift.master.cluster_method == 'native' and inventory_hostname != openshift_master_hosts[0]
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -172,89 +169,45 @@
...
 
 - set_fact:
     master_api_service_status_changed: "{{ start_result | changed }}"
-  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native' and inventory_hostname == openshift_master_hosts[0]
+  when: openshift.master.cluster_method == 'native' and inventory_hostname == openshift_master_hosts[0]
 
 - pause:
     seconds: 15
-  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native'
+  when: openshift.master.ha | bool and openshift.master.cluster_method == 'native'
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -172,89 +169,45 @@
...
+  when: openshift.master.cluster_method == 'native' and inventory_hostname != openshift_master_hosts[0]
   register: start_result
   until: not start_result | failed
   retries: 1
   delay: 60
 
 - set_fact:
     master_api_service_status_changed: "{{ start_result | changed }}"
-  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native' and inventory_hostname != openshift_master_hosts[0]
+  when: openshift.master.cluster_method == 'native' and inventory_hostname != openshift_master_hosts[0]
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -275,14 +228,14 @@
   delay: 1
   run_once: true
   changed_when: false
-  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native' and master_api_service_status_changed | bool
+  when: openshift.master.cluster_method == 'native' and master_api_service_status_changed | bool
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -275,14 +228,14 @@
...
 
 - name: Start and enable master controller on first master
   systemd:
     name: "{{ openshift.common.service_type }}-master-controllers"
     enabled: yes
     state: started
-  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native' and inventory_hostname == openshift_master_hosts[0]
+  when: openshift.master.cluster_method == 'native' and inventory_hostname == openshift_master_hosts[0]
```
tbielawa: multi-line `WHEN`

roles/openshift_master/tasks/main.yml
```diff
@@ -291,26 +244,26 @@
...
-  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native'
+  when: openshift.master.cluster_method == 'native'
 
 - name: Start and enable master controller on all masters
   systemd:
     name: "{{ openshift.common.service_type }}-master-controllers"
     enabled: yes
     state: started
-  when: openshift_master_ha | bool and openshift.master.cluster_method == 'native' and inventory_hostname != openshift_master_hosts[0]
+  when: openshift.master.cluster_method == 'native' and inventory_hostname != openshift_master_hosts[0]
```
tbielawa: multi-line `WHEN`

roles/openshift_metrics/handlers/main.yml
```diff
@@ -1,17 +1,12 @@
...
 - name: restart master api
   systemd: name={{ openshift.common.service_type }}-master-api state=restarted
-  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  when: (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
   notify: Verify API Server
 
 - name: restart master controllers
   systemd: name={{ openshift.common.service_type }}-master-controllers state=restarted
-  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  when: (not (master_controllers_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
```
tbielawa: multi-line `WHEN`

roles/openshift_metrics/handlers/main.yml
```diff
@@ -1,17 +1,12 @@
 ---
-- name: restart master
-  systemd: name={{ openshift.common.service_type }}-master state=restarted
-  when: (openshift.master.ha is not defined or not openshift.master.ha | bool) and (not (master_service_status_changed | default(false) | bool))
-  notify: Verify API Server
-
 - name: restart master api
   systemd: name={{ openshift.common.service_type }}-master-api state=restarted
-  when: (openshift.master.ha is defined and openshift.master.ha | bool) and (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
+  when: (not (master_api_service_status_changed | default(false) | bool)) and openshift.master.cluster_method == 'native'
```
tbielawa: multi-line `WHEN`

sdodson:
@smarterclayton they're in gerrit, i've updated them so that the master is an etcd host.

roles/openshift_metrics/handlers/main.yml
smarterclayton: I'd like to do the multi-line when in a follow up, it would make this much harder to review and the re are way more references in these files than just my changes

smarterclayton:
aos-ci-test

smarterclayton:
aos-ci-test

roles/openshift_master_facts/tasks/main.yml
```diff
@@ -32,7 +32,7 @@
   openshift_facts:
     role: master
     local_facts:
-      cluster_method: "{{ openshift_master_cluster_method | default(None) }}"
+      cluster_method: "{{ openshift_master_cluster_method | default('native') }}"
```
ingvagabund: Thanks for that, no more `openshift_master_cluster_method variable not set` after 30 minutes of installation.

roles/openshift_metrics/handlers/main.yml
ingvagabund: I can do all the multi-line when in https://github.com/openshift/openshift-ansible/pull/4947#issuecomment-319255197

ingvagabund:
> Changes required to support this in upgrades:
> 
> 1.  stop and remove the old origin-master.service and sysconfig files if they exist
> 2.  error / warn if the user had those files that scripts and other automation may need to change.

We should not merge this PR until we have a code that supports both changes. Since if anyone tries to upgrade from an existing (e.g. 3.5 or 3.6) cluster to 3.7 with openshift-ansible-3.7.*, the master service will keep running and may cause disruptions during the upgrade.

playbooks/common/openshift-cluster/upgrades/pre/verify_control_plane_running.yml
```diff
@@ -9,23 +9,16 @@
...
-      enabled: yes
-    when: openshift.master.ha is defined and not openshift.master.ha | bool and openshift.common.is_containerized | bool
-
   - name: Ensure HA Master is running
     service:
       name: "{{ openshift.common.service_type }}-master-api"
       state: started
       enabled: yes
-    when: openshift.master.ha is defined and openshift.master.ha | bool and openshift.common.is_containerized | bool
+    when: openshift.common.is_containerized | bool
```
ingvagabund: Assume we have containerized 3.6 cluster. Before this change the `master` docker service is up and running. There is no `master-api` nor `master-controllers` services running. Once this PR is merged, the check will fail on the cluster as only `master` docker service is running. Well, at least nothing gets updated. Though, that is not what we want, am I right? We should make the check conditional. If the cluster version is 3.6 and less and the deployment is containerized, check what services are actually running/installed and based on that decide which services need to be verified to be running.

roles/openshift_master/tasks/systemd_units.yml
```diff
@@ -22,34 +22,12 @@
...
-- name: Install Master service file
-  copy:
-    dest: "/etc/systemd/system/{{ openshift.common.service_type }}-master.service"
-    src: "{{ openshift.common.service_type }}-master.service"
-  register: create_master_unit_file
-  when:
-  - not openshift.common.is_containerized | bool
-  - (openshift.master.ha is not defined or not openshift.master.ha) | bool
-
-- command: systemctl daemon-reload
```
ingvagabund: Be aware this file is included in the control plane upgrade as well. See https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/upgrades/upgrade_control_plane.yml#L94

ingvagabund: You might want to refactor https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-master/restart_services.yml as well.

ingvagabund:
Plus, what you are trying to achieve here is not covered in the CI. Since the upgrade is tested over fresh installation of the same version. I don't think we test upgrade from 3.d to 3.(d+1). @jhadvig prove me wrong.

playbooks/common/openshift-cluster/upgrades/pre/verify_control_plane_running.yml
smarterclayton: I think that *is* what we want - upgrade to be blocked.

Technically, the support path for this old mode (which is unusual) should be removed.  We don't support upgrade from non-HA to HA.

smarterclayton:
> We should not merge this PR until we have a code that supports both changes. Since if anyone tries to upgrade from an existing (e.g. 3.5 or 3.6) cluster to 3.7 with openshift-ansible-3.7.*, the master service will keep running and may cause disruptions during the upgrade.

Part of this PR is explicitly removing support for certain old paths:

1. single master without separate etcd
2. single master with origin-master

For all single master cases, there is only one supported path now - being set up "like HA" on a single master, with etcd, api, and controller all separate.

I'm on the fence about whether we need to support this upgrade, but we definitely do *not* need to support it in order to get the other changes in for bootstrapping in 3.7.  So I'd say that yes, we should probably have an upgrade path before we ship 3.7, but I don't think we need it for this specific PR to unblock the work that we are doing to move to a "bootstrap" model on nodes specifically, and a "master components in static pods" model on the master.  So I don't want to work on the upgrade *before* we make the decision whether to completely remove all non-kubelet units / containerized units.

smarterclayton:
Proposed migration flow (only applies to single master clusters):

1. If HA or multi-master, do nothing
2. Stop and remove the openshift-master service
3. Copy appropriate sysconfig settings to the openshift-master-api and openshift-master-controllers sysconfig files
4. Install the new unit files and start them

If we go to static pods, step 3/4 would be replaced with an equivalent.

sdodson:
aos-ci-test

smarterclayton:
Moved the upgrade to an issue with the details of what has to change for each bit.

sdodson:
[merge]

smarterclayton:
Will also do @tbielawa's cleanup in a follow on

ingvagabund:
Failed to the second time on:
```
++ OS_RPM_RELEASE=0.rc.0.472.8ef052d
++ export OS_RPM_NAME OS_RPM_VERSION OS_RPM_RELEASE OS_RPM_ARCHITECTURE
++ echo origin-3.6.0-0.rc.0.472.8ef052d.x86_64
+ origin_package=origin-3.6.0-0.rc.0.472.8ef052d.x86_64
+ rpm -V origin-3.6.0-0.rc.0.472.8ef052d.x86_64
package origin-3.6.0-0.rc.0.472.8ef052d.x86_64 is not installed
```

smarterclayton:
in theory fixed

[merge]

sdodson:
[merge]

smarterclayton:
[merge] flaked on daemon set extended test

smarterclayton:
[merge] in case this is a flake, otherwise might be a setting openshift-master that is no longer being set in the split mode (like a qps tuning)

smarterclayton:
aos-ci-test

smarterclayton:
[merge] we may be right on the edge of a timeout, or something more subtle is at work.

Should aos-ci-test kick the tests again or did that change?

smarterclayton:
Oh, that was not obvious.  Merge runs upgrade, which is still only calling the 3.6 upgrade (because we have no 3.7 tags available).  So this is going to be blocked on doing the upgrade anyway, and is also blocked on the 3.7 tags becoming available.

sdodson:
> Oh, that was not obvious. Merge runs upgrade, which is still only calling the 3.6 upgrade (because we have no 3.7 tags available). So this is going to be blocked on doing the upgrade anyway, and is also blocked on the 3.7 tags becoming available.

We're no longer blocked on pushing 3.7 tags, right?

smarterclayton:
Yes, something is failing on the origin build side related to RPM versions in the egress router docker image build

smarterclayton:
aos-ci-test

smarterclayton:
Updated with the upgrade script (which simply disables and masks origin-master) and also removes all additional places that start origin-master.

The fedora job failure may be because the origin-master service is still started.  Rerunning in case it picked up the older change.

smarterclayton:
aos-ci-test

smarterclayton:
[test]

sdodson:
registry flake?

sdodson:
aos-ci-test

smarterclayton:
Looks like the tests passed in the one job.  [test] again to see if it works this time.

smarterclayton:
This went green on install_update.  The fedora error seems to be happening on other PRs.  The status_check failure on test was a flake in the job.  Seems to be ok from a test perspective.

sdodson:
[merge]

ingvagabund:
\O/

smarterclayton:
Will do follow up for the other changes tomorrow or Friday - specifically the when changes

mffiedler:
This breaks things like the metrics deployer that assume non-ha uses the atomic-openshift-master system unit.   Will start filing BZs.

PR #4834, opened 2017-07-24
[WIP] Adding Memcached for Kibana SSO

enoodle:
This will add a memcached pod to be used for the auth-proxy pods. This will allow multiple replicas of the auth-proxy to sync on the User/Access token pairs received and allow users to login from a different pod than the one that was used to setup SSO

auth_proxy_with_memcached

This is intended to solve BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1470022

This is also expecting images with the names like: "docker.io/openshift/origin-logging-memcached" to exist.

### DIFF
--- a/roles/openshift_logging_kibana/tasks/main.yaml
+++ b/roles/openshift_logging_kibana/tasks/main.yaml
@@ -95,6 +95,20 @@
     - port: 443
       targetPort: "oaproxy"
 
+- name: Set memcached service
+  oc_service:
+    state: present
+    name: "{{ kibana_name }}-memcached"
+    namespace: "{{ openshift_logging_kibana_namespace }}"
+    selector:
+      component: "{{ kibana_component }}"
+      provider: openshift
+    labels:
+      logging-infra: 'support'
+    ports:
+    - port: 11211
+      targetPort: "memcached"
+
 # create routes
 # TODO: set up these certs differently?
 - set_fact:
@@ -235,6 +249,7 @@
     kibana_proxy_memory_limit: "{{ openshift_logging_kibana_proxy_memory_limit }}"
     replicas: "{{ openshift_logging_kibana_replicas | default (1) }}"
     kibana_node_selector: "{{ openshift_logging_kibana_nodeselector | default({}) }}"
+    memcached_hostname: "{{ kibana_name }}-memcached"
 
 - name: Set Kibana DC
   oc_obj:
@@ -244,6 +259,30 @@
     kind: dc
     files:
     - "{{ tempdir }}/templates/kibana-dc.yaml"
+    delete_after: true
+
+- name: Generate Memcached DC template
+  template:
+    src: memcached.j2
+    dest: "{{ tempdir }}/templates/memcached-dc.yaml"
+  vars:
+    component: "{{ kibana_component }}"
+    logging_component: kibana
+    deploy_name: "memcached-{{ kibana_name }}"
+    image: "{{ openshift_logging_image_prefix }}logging-memcached:{{ openshift_logging_image_version }}"
+    kibana_cpu_limit: "{{ openshift_logging_kibana_cpu_limit }}"
+    kibana_memory_limit: "{{ openshift_logging_kibana_memory_limit }}"
+    kibana_proxy_cpu_limit: "{{ openshift_logging_kibana_proxy_cpu_limit }}"
+    kibana_proxy_memory_limit: "{{ openshift_logging_kibana_proxy_memory_limit }}"
+
+- name: Set Memcached DC
+  oc_obj:
+    state: present
+    name: "memcached-{{ kibana_name }}"
+    namespace: "{{ openshift_logging_namespace }}"
+    kind: dc
+    files:
+    - "{{ tempdir }}/templates/memcached-dc.yaml"
     delete_after: true
 
 # update master configs?
--- a/roles/openshift_logging_kibana/templates/memcached.j2
+++ b/roles/openshift_logging_kibana/templates/memcached.j2
@@ -1,0 +1,55 @@
+apiVersion: "v1"
+kind: "DeploymentConfig"
+metadata:
+  name: "{{ deploy_name }}"
+  labels:
+    provider: openshift
+    component: "{{ component }}"
+    logging-infra: "{{ logging_component }}"
+spec:
+  replicas: 1
+  selector:
+    provider: openshift
+    component: "{{ component }}"
+    logging-infra: "{{ logging_component }}"
+  strategy:
+    rollingParams:
+      intervalSeconds: 1
+      timeoutSeconds: 600
+      updatePeriodSeconds: 1
+    type: Rolling
+  template:
+    metadata:
+      name: "{{ deploy_name }}"
+      labels:
+        logging-infra: "{{ logging_component }}"
+        provider: openshift
+        component: "{{ component }}"
+    spec:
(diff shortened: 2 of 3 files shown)


roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -95,6 +95,20 @@
     - port: 443
       targetPort: "oaproxy"
 
+- name: Set memcached service
```
jcantrill: @enoodle I think we need to figure out where this better fits in the logging roles.  @portante has suggested we use memcache for other parts of the logging stack and it would be convenient to set it up correctly the first time.  Maybe it should just live in the 'openshift_logging' role

roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -95,6 +95,20 @@
...
+  oc_service:
+    state: present
+    name: "{{ kibana_name }}-memcached"
```
jcantrill: I think a constant is good enough here 'logging-memcached'

roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -95,6 +95,20 @@
...
+    namespace: "{{ openshift_logging_kibana_namespace }}"
+    selector:
+      component: "{{ kibana_component }}"
```
jcantrill: Does this need to be variable? Can we call it 'logging-memcached'

roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -235,6 +249,7 @@
     kibana_proxy_memory_limit: "{{ openshift_logging_kibana_proxy_memory_limit }}"
     replicas: "{{ openshift_logging_kibana_replicas | default (1) }}"
     kibana_node_selector: "{{ openshift_logging_kibana_nodeselector | default({}) }}"
+    memcached_hostname: "{{ kibana_name }}-memcached"
```
jcantrill: same here?  constant?

roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -246,6 +261,30 @@
...
+- name: Generate Memcached DC template
+  template:
+    src: memcached.j2
+    dest: "{{ tempdir }}/templates/memcached-dc.yaml"
+  vars:
+    component: "{{ kibana_component }}"
+    logging_component: kibana
+    deploy_name: "memcached-{{ kibana_name }}"
+    image: "{{ openshift_logging_image_prefix }}logging-memcached:{{ openshift_logging_image_version }}"
+    kibana_cpu_limit: "{{ openshift_logging_kibana_cpu_limit }}"
```
jcantrill: I believe you only need a single set of cpu and memory limts and we wish to distinguish them from those used by kibana and the proxy

roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -246,6 +261,30 @@
     - "{{ tempdir }}/templates/kibana-dc.yaml"
     delete_after: true
 
+- name: Generate Memcached DC template
+  template:
+    src: memcached.j2
+    dest: "{{ tempdir }}/templates/memcached-dc.yaml"
+  vars:
+    component: "{{ kibana_component }}"
+    logging_component: kibana
```
jcantrill: maybe 'memcached'

roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -246,6 +261,30 @@
...
+    deploy_name: "memcached-{{ kibana_name }}"
```
jcantrill: fixed at logging-memcached

roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -246,6 +261,30 @@
...
+    image: "{{ openshift_logging_image_prefix }}logging-memcached:{{ openshift_logging_image_version }}"
```
jcantrill: There will need to be some logic in defaults and such to allow the image version to be pegged at something specific to avoid unexpected behavior when pods restart and new images are available

roles/openshift_logging_kibana/templates/memcached.j2
```diff
@@ -0,0 +1,55 @@
...
+    type: Rolling
+  template:
+    metadata:
+      name: "{{ deploy_name }}"
+      labels:
+        logging-infra: "{{ logging_component }}"
+        provider: openshift
+        component: "{{ component }}"
+    spec:
+      serviceAccountName: aggregated-logging-kibana
```
jcantrill: Do we need a service account?

roles/openshift_logging_kibana/templates/memcached.j2
```diff
@@ -0,0 +1,55 @@
...
+      containers:
+        -
+          name: "memcached"
+          image: {{ image }}
+          imagePullPolicy: Always
+{% if (kibana_memory_limit is defined and kibana_memory_limit is not none and kibana_memory_limit != "") or (kibana_cpu_limit is defined and kibana_cpu_limit is not none and kibana_cpu_limit != "") %}
```
jcantrill: This should obviously match the vars we choose from my previous comment

roles/openshift_logging_kibana/templates/memcached.j2
```diff
@@ -0,0 +1,55 @@
...
+{% if kibana_cpu_limit is not none and kibana_cpu_limit != "" %}
+              cpu: "{{ kibana_cpu_limit }}"
+{% endif %}
+{% if kibana_memory_limit is not none and kibana_memory_limit != "" %}
+              memory: "{{ kibana_memory_limit }}"
+{% endif %}
+{% endif %}
+          env:
+            -
+              name: "KIBANA_MEMORY_LIMIT"
```
jcantrill: I assume this should match whatever value is expected by a run script for memcache

roles/openshift_logging_kibana/tasks/main.yaml
ewolinetz: i would like to avoid adding more to `openshift_logging` as it may eventually be phased out in exchange for a playbook instead

portante: @ewolinetz, @jcantril, can you two explain more about your comments?  This is being added to the `openshift_logging_kibana` role right now.  Are we suggesting a new role, `openshift_logging_memcached`?  Or something else?  Can you provide some examples of what you two are thinking of?

roles/openshift_logging_kibana/tasks/main.yaml
portante: The key here is we need to have the `requests` and `limits` for the memory size configuration of memcached be reflected properly.

roles/openshift_logging_kibana/tasks/main.yaml
jcantrill: I would prefer not to add yet another role, but if its a component in logging we intend to rely on and use outside of kibana then it should have its own role.

ewolinetz: my argument is just that I would like to avoid adding any new logic to `openshift_logging`, I'm alright with a new role if its a new, separate, stand alone component or with there being redundancy in only a few components if it will be deployed with them (and they will require it to be configured)

PR #4835, opened 2017-07-24
Use absolute path for excluders

tripledes:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1474246

Use absolute path when executing excluder as it's used when checking for excluder.


tripledes:
@sdodson sorry, just noticed the unexclude.yml also use a relative path. Now that this one is merged, should I open a new one for the unexclude? Thanks!

sdodson:
@tripledes yeah that'd be great, thanks

PR #4839, opened 2017-07-24
Use absolute path when un-excluding

tripledes:
* As follow-up on #4835, use absolute path also when un-excluding


sdodson:
aos-ci-test

sdodson:
[merge]

tripledes:
Was checking the failed tests and seem not be because of this PR.

Issue #4838, opened 2017-07-24
oc_route fails when destcacert is empty


zgalor:
#### Description

In 3.6, creating a route with tls_termination=reencrypt doesn't require passing the destination ca certificate, yet the module requires it.

##### Version

```
ansible 2.3.1.0
  config file = /home/zgalor/dev/openshift-ansible/ansible.cfg
  configured module search path = Default w/o overrides
  python version = 2.7.13 (default, May 10 2017, 20:04:28) [GCC 6.3.1 20161221 (Red Hat 6.3.1-1)]

openshift-ansible-3.6.153-1-72-g1b6d8d0
```

##### Steps To Reproduce
Run this task:

- name: create some route
  oc_route:
    name: "{{ item.name }}"
    namespace: test
    service_name: "{{ item.name }}"
    tls_termination: reencrypt
  with_items:
    - name: test_route

##### Expected Results
this task should create the route with empty destinantionCaCertificate
```
"tls": {
                        "destinationCACertificate": "-----BEGIN COMMENT-----\nThis is an empty PEM file created to provide backwards compatibility\nfor reencrypt routes that have no destinationCACertificate. This \ncontent will only appear for routes accessed via /oapi/v1/routes.\n-----END COMMENT-----\n", 
                        "termination": "reencrypt"
                    }
```

##### Observed Results
An Error is displayed:

```
Verify that you pass a value for destcacert
```

##### Additional Information

Provide any additional information which may help us diagnose the
issue.

```
Fedora release 25 (Twenty Five)
```

zgalor:
cc @kwoodson

PR #4840, opened 2017-07-24
Remove test that certificate file exists when termination is reencrypt

zgalor:
In openshift 3.6, providing destination ca certificate to route when tls_termination==reencrypt is not mandatory, so the code that verifies it has been removed.

This resolves issue https://github.com/openshift/openshift-ansible/issues/4838

### DIFF
--- a/roles/lib_openshift/library/oc_route.py
+++ b/roles/lib_openshift/library/oc_route.py
@@ -1663,20 +1663,6 @@ class OCRoute(OpenShiftCLI):
         skip = []
         return not Utils.check_def_equal(self.config.data, self.route.yaml_dict, skip_keys=skip, debug=self.verbose)
 
-    @staticmethod
-    def get_cert_data(path, content):
-        '''get the data for a particular value'''
-        if not path and not content:
-            return None
-
-        rval = None
-        if path and os.path.exists(path) and os.access(path, os.R_OK):
-            rval = open(path).read()
-        elif content:
-            rval = content
-
-        return rval
-
     # pylint: disable=too-many-return-statements,too-many-branches
     @staticmethod
     def run_ansible(params, check_mode=False):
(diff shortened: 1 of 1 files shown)


zgalor:
@kwoodson please review

roles/lib_openshift/library/oc_route.py
```diff
@@ -1704,18 +1690,6 @@ def run_ansible(params, check_mode=False):
                          'content': params['key_content'],
                          'value': None, }, }
 
-        if params['tls_termination'] and params['tls_termination'].lower() != 'passthrough':  # E501
-
-            for key, option in files.items():
-                if key == 'destcacert' and params['tls_termination'] != 'reencrypt':
-                    continue
-
-                option['value'] = OCRoute.get_cert_data(option['path'], option['content'])  # E501
```
kwoodson: I'm not sure this is the right route.  How do we load the data from the certificates if this code is removed?

I understand that reencrypt is your target but what about the other methods?

zgalor: There is no need to load the data from the certificate (at least not in that module). `get_cert_data` is not called from anywhere else.

If I understood the code logic correctly, it only verifies the certificates for reencrypt method.

kwoodson:
```python
  if params['tls_termination'] and params['tls_termination'].lower() != 'passthrough':  # E501

            for key, option in files.items():
                if key == 'destcacert' and params['tls_termination'] != 'reencrypt':
                    continue

                option['value'] = OCRoute.get_cert_data(option['path'], option['content'])  # E501

                if not option['value']:
                    return {'failed': True,
                            'msg': 'Verify that you pass a value for %s' % key}
```
If the `tls_termination` is not `passthrough` then we need the certificate data.
if the `key` is destcacert and `tls_termination` is not `reencrypt`, then we skip.  This only skips `destcacert` if `tls_termination` is not reencrypt`, right?  

What about `cacert`, `key`, and `cert`?

zgalor: Yes, I missed that.
I'll fix the logic.

zgalor: On the other hand, there are 3 termination types: `edge`, `passthrough`, and `reencrypt`.
`passthrough` of course doesn't require any of the files.
When running oc_route on 3.6 with `reencrypt` or `edge`, and no keys or certificates the route object gets created. for `reencrypt` a default `destinationCACertificate` is created, and the route works. for `edge` no certificate is defined. I don't know if the route works, but it does get created. 
The question is what is the modules policy? does it prevent the user from defining objects that the API allows? If it doesn't, I think it should skip the validation and allow creation.

zgalor: @kwoodson do you agree?

zgalor:
closing.
opened a new PR with a better fix https://github.com/openshift/openshift-ansible/pull/4887

PR #4841, opened 2017-07-24
Fix for gce storageclass provisioner default.

kwoodson:


kwoodson:
Should address this issue: https://github.com/openshift/origin/issues/15401#issuecomment-317412670

kwoodson:
aos-ci-test

PR #4846, opened 2017-07-24
libvirt fixups

dcbw:
Fall back to mkisofs if genisoimage isn't available (they use the same arguments).  (thanks to sdodson for the magic on this one)

Update README to document requirement that SSH keys be at specific paths ~/.ssh/id_rsa and ~/id_rsa.pub.

@sdodson


sdodson:
LGTM, unfortunately none of these changes would be tested via our CI tooling so merging manually.

PR #4849, opened 2017-07-24
fix for storageclass provision defaults.

kwoodson:


kwoodson:
@sdodson, I have broken the default provision parameter with the last PR for this.  This one fixes that default variable for the provisioner param.

jeremyeder:
Thanks, that's better now.

PR #4851, opened 2017-07-24
Impl fluentd file buffer

jcantrill:
This PR removes memory buffer and supercedes #4740

### DIFF
--- a/roles/openshift_logging/README.md
+++ b/roles/openshift_logging/README.md
@@ -57,6 +57,7 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_fluentd_hosts`: List of nodes that should be labeled for Fluentd to be deployed to. Defaults to ['--all'].
 - `openshift_logging_fluentd_buffer_queue_limit`: Buffer queue limit for Fluentd. Defaults to 1024.
 - `openshift_logging_fluentd_buffer_size_limit`: Buffer chunk limit for Fluentd. Defaults to 1m.
+- `openshift_logging_fluentd_file_buffer_limit`: Fluentd will set the value to the file buffer limit.  Defaults to '1Gi' per destination.
 
 
 - `openshift_logging_es_host`: The name of the ES service Fluentd should send logs to. Defaults to 'logging-es'.
@@ -160,3 +161,18 @@ Elasticsearch OPS too, if using an OPS cluster:
   need to set this
 - `openshift_logging_mux_buffer_queue_limit`: Default `[1024]` - Buffer queue limit for Mux.
 - `openshift_logging_mux_buffer_size_limit`: Default `[1m]` - Buffer chunk limit for Mux.
+- `openshift_logging_mux_file_buffer_limit`: Default `[2Gi]` per destination - Mux will
+  set the value to the file buffer limit.
+- `openshift_logging_mux_file_buffer_storage_type`: Default `[emptydir]` - Storage
+  type for the file buffer.  One of [`emptydir`, `pvc`, `hostmount`]
+
+- `openshift_logging_mux_file_buffer_pvc_size`: The requested size for the file buffer
+  PVC, when not provided the role will not generate any PVCs. Defaults to `4Gi`.
+- `openshift_logging_mux_file_buffer_pvc_dynamic`: Whether or not to add the dynamic
+  PVC annotation for any generated PVCs. Defaults to 'False'.
+- `openshift_logging_mux_file_buffer_pvc_pv_selector`: A key/value map added to a PVC
+  in order to select specific PVs.  Defaults to 'None'.
+- `openshift_logging_mux_file_buffer_pvc_prefix`: The prefix for the generated PVCs.
+  Defaults to 'logging-mux'.
+- `openshift_logging_mux_file_buffer_storage_group`: The storage group used for Mux.
+  Defaults to '65534'.
--- a/roles/openshift_logging_fluentd/templates/fluentd.j2
+++ b/roles/openshift_logging_fluentd/templates/fluentd.j2
@@ -62,6 +62,8 @@ spec:
         - name: dockerdaemoncfg
           mountPath: /etc/docker
           readOnly: true
+        - name: filebufferstorage
+          mountPath: /var/lib/fluentd
 {% if openshift_logging_use_mux_client | bool %}
         - name: muxcerts
           mountPath: /etc/fluent/muxkeys
@@ -112,6 +114,8 @@ spec:
               resource: limits.memory
         - name: "USE_MUX_CLIENT"
           value: "{{ openshift_logging_use_mux_client | default('false') | lower }}"
+        - name: "FILE_BUFFER_LIMIT"
+          value: "{{ openshift_logging_fluentd_file_buffer_limit | default('1Gi') }}"
       volumes:
       - name: runlogjournal
         hostPath:
@@ -145,3 +149,6 @@ spec:
         secret:
           secretName: logging-mux
 {% endif %}
+      - name: filebufferstorage
+        hostPath:
+          path: "/var/lib/fluentd"
--- a/roles/openshift_logging_mux/tasks/main.yaml
+++ b/roles/openshift_logging_mux/tasks/main.yaml
@@ -177,6 +177,18 @@
   check_mode: no
   changed_when: no
 
+- name: Create Mux PVC
+  oc_pvc:
+    state: present
+    name: "{{ openshift_logging_mux_file_buffer_pvc_name }}"
+    namespace: "{{ openshift_logging_mux_namespace }}"
+    volume_capacity: "{{ openshift_logging_mux_file_buffer_pvc_size }}"
+    access_modes: "{{ openshift_logging_mux_file_buffer_pvc_access_modes | list }}"
+    selector: "{{ openshift_logging_mux_file_buffer_pvc_pv_selector }}"
+    storage_class_name: "{{ openshift_logging_mux_file_buffer_pvc_storage_class_name | default('', true) }}"
+  when:
+  - openshift_logging_mux_file_buffer_storage_type == "pvc"
+
 - name: Set logging-mux DC
   oc_obj:
     state: present
--- a/roles/openshift_logging_mux/templates/mux.j2
+++ b/roles/openshift_logging_mux/templates/mux.j2
@@ -66,6 +66,8 @@ spec:
         - name: muxcerts
           mountPath: /etc/fluent/muxkeys
           readOnly: true
+        - name: filebufferstorage
+          mountPath: /var/lib/fluentd
         env:
         - name: "K8S_HOST_URL"
           value: "{{openshift_logging_mux_master_url}}"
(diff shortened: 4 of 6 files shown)


jcantrill:
aos-ci-test

jcantrill:
This is the companion PR to openshift/openshift-ansible/pull/4851

sdodson:
> This is the companion PR to openshift/openshift-ansible#4851

I'm my own companion too.

roles/openshift_logging_fluentd/vars/main.yml
```diff
@@ -2,3 +2,4 @@
 __latest_fluentd_version: "3_5"
 __allowed_fluentd_versions: ["3_5", "3_6"]
 __allowed_fluentd_types: ["hosted", "secure-aggregator", "secure-host"]
+openshift_logging_fluentd_buffer_type: "file"
```
ewolinetz: vars in here should start with "__" since they are being considered private

roles/openshift_logging_mux/tasks/main.yaml
```diff
@@ -177,6 +177,34 @@
   check_mode: no
   changed_when: no
 
+# storageclasses are used by default but if static then disable
+# storageclasses with the storageClassName set to "" in pvc.j2
+- name: Creating Mux storage template
```
ewolinetz: is it possible to replace this `template` and `oc_obj` logic with a `oc_pvc` call?

roles/openshift_logging_mux/vars/main.yml
```diff
@@ -1,3 +1,4 @@
 ---
 __latest_mux_version: "3_5"
 __allowed_mux_versions: ["3_5", "3_6"]
+openshift_logging_mux_buffer_type: "file"
```
ewolinetz: prefix this with "__" please
we will also need to make sure we're loading and setting the `openshift_logging_mux_buffer_type` fact afterwards too

roles/openshift_logging_fluentd/vars/main.yml
ewolinetz: like below, will need to make sure we load `openshift_logging_fluentd_buffer_type` into a fact afterwards

roles/lib_openshift/src/lib/pvc.py
```diff
@@ -36,12 +40,18 @@ def create_dict(self):
         self.data['spec']['accessModes'] = ['ReadWriteOnce']
         if self.access_modes:
             self.data['spec']['accessModes'] = self.access_modes
+        if self.selector:
+            self.data['spec']['selector'] = {'matchLabels': self.selector}
```
kwoodson: I'm not sure the intention here.  What is `matchLabels`?

roles/lib_openshift/src/lib/pvc.py
```diff
@@ -51,13 +61,29 @@ class PersistentVolumeClaim(Yedit):
     volume_name_path = "spec.volumeName"
     bound_path = "status.phase"
     kind = 'PersistentVolumeClaim'
+    selector_path = "spec.selector.matchLabels"
+    storage_class_name_path = "spec.storageClassName"
 
     def __init__(self, content):
         '''RoleBinding constructor'''
```
kwoodson: If you wouldn't mind renaming this to PersistentVolumeClaim in the doc.  Seems like a copy paste from before.

roles/lib_openshift/src/lib/pvc.py
```diff
@@ -72,6 +98,24 @@ def volume_name(self, data):
         self._volume_name = data
 
     @property
+    def selector(self):
+        ''' selector property '''
+        if self._selector is None:
+            self._selector = self.get_selector()
+            if not isinstance(self._selector, hash):
```
kwoodson: I think you meant `dict`.

roles/lib_openshift/src/lib/pvc.py
```diff
@@ -72,6 +98,24 @@ def volume_name(self, data):
...
+                self._selector = hash(self._selector)
+
+        return self._selector
+
+    @selector.setter
+    def selector(self, data):
+        ''' selector property setter'''
+        if not isinstance(data, hash):
```
kwoodson: `dict`

roles/lib_openshift/src/lib/pvc.py
```diff
@@ -72,6 +98,24 @@ def volume_name(self, data):
...
+            data = hash(data)
```
kwoodson: `dict`

roles/openshift_logging_fluentd/templates/fluentd.j2
```diff
@@ -62,6 +62,8 @@ spec:
         - name: dockerdaemoncfg
           mountPath: /etc/docker
           readOnly: true
+        - name: filebufferstorage
+          mountPath: /var/lib/fluentd/buffer
```
richm: so this should be just `/var/lib/fluentd` ?

roles/openshift_logging_mux/templates/mux.j2
```diff
@@ -66,6 +66,8 @@ spec:
         - name: muxcerts
           mountPath: /etc/fluent/muxkeys
           readOnly: true
+        - name: filebufferstorage
+          mountPath: /var/lib/fluentd/buffer
```
richm: `/var/lib/fluentd` ? too

roles/lib_openshift/src/lib/pvc.py
jcantrill: matchLabels is how you specify the selector.... spec.selector.matchLabels = {}

jcantrill:
Now depends on #4879

roles/openshift_logging/README.md
```diff
@@ -57,6 +57,7 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_fluentd_hosts`: List of nodes that should be labeled for Fluentd to be deployed to. Defaults to ['--all'].
 - `openshift_logging_fluentd_buffer_queue_limit`: Buffer queue limit for Fluentd. Defaults to 1024.
 - `openshift_logging_fluentd_buffer_size_limit`: Buffer chunk limit for Fluentd. Defaults to 1m.
+- `openshift_logging_fluentd_file_buffer_limit`: Fluentd will set the value to the file buffer limit.  Defaults to '1Gi' per output.
```
portante: Perhaps change "per output" to "per output destination"?

roles/openshift_logging/README.md
```diff
@@ -160,3 +161,18 @@ Elasticsearch OPS too, if using an OPS cluster:
   need to set this
 - `openshift_logging_mux_buffer_queue_limit`: Default `[1024]` - Buffer queue limit for Mux.
 - `openshift_logging_mux_buffer_size_limit`: Default `[1m]` - Buffer chunk limit for Mux.
+- `openshift_logging_mux_file_buffer_limit`: Default `[2Gi]` per output - Mux will
+  set the value to the file buffer limit.
```
portante: Not sure, but if "Mux will set the value to the file buffer limit," and this is the option for setting the mux file buffer limit, does this mean to say that it will calculate a default from the available diskspace considering the number of outputs?

jcantrill: The number of outputs is taken into consideration when this value is consumed by run.sh

https://github.com/nhosoi/origin-aggregated-logging/blob/47463c55ccf9dd9f334a8fdc21fab6f36de196c4/fluentd/run.sh#L155-L164

portante: @jcantrill, do you think this comment in the README conveys that?

nhosoi: The value of openshift_logging_mux_file_buffer_limit is set to the environment variable FILE_BUFFER_LIMIT in "templates/mux.j2".

FILE_BUFFER_LIMIT is converted to bytes in run.sh like this:
`TOTAL_LIMIT=$(echo ${FILE_BUFFER_LIMIT:-2Gi} | sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc) || :`
Then, multiplied by $NUM_OUTPUTS to get the total size of file buffer.
`TOTAL_LIMIT=$(expr $TOTAL_LIMIT \* $NUM_OUTPUTS) || :`
This calculation is needed to compare the TOTAL_LIMIT with the available volume size DF_LIMIT.  Then, the TOTAL_LIMIT is divided by NUM_OUTPUTS again to get the real buffer size limit per buffer:
```
# TOTAL_BUFFER_SIZE_LIMIT per buffer
TOTAL_BUFFER_SIZE_LIMIT=$(expr $TOTAL_LIMIT / $NUM_OUTPUTS) || :
```
The TOTAL_BUFFER_SIZE_LIMIT is identical to FILE_BUFFER_LIMIT if DF_LIMIT is larger than TOTAL_LIMIT.  But if the volume size is smaller, the TOTAL_BUFFER_SIZE_LIMIT is smaller than the user specified FILE_BUFFER_LIMIT.  In the case, the following WARNING is issued by fluentd/run.sh:
WARNING: Available disk space ($DF_LIMIT bytes) is less than the user specified file buffer limit ($FILE_BUFFER_LIMIT times $NUM_OUTPUTS)."

ewolinetz:
aos-ci-test

sdodson:
[test]

nhosoi:
Not in this patch, but based on the performance improvement effort by @richm and @portante, the default buffer_chunk_ limit value needs to be 16m, I'd think.
```
diff --git a/roles/openshift_logging/defaults/main.yml b/roles/openshift_logging/defaults/main.yml
index 1c243f9..b3644af 100644
--- a/roles/openshift_logging/defaults/main.yml
+++ b/roles/openshift_logging/defaults/main.yml
@@ -76,7 +76,7 @@ openshift_logging_fluentd_journal_source: "{{ openshift_hosted_logging_journal_s
 openshift_logging_fluentd_journal_read_from_head: "{{ openshift_hosted_logging_journal_read_from_head | default('') }}"
 openshift_logging_fluentd_hosts: ['--all']
 openshift_logging_fluentd_buffer_queue_limit: 1024
-openshift_logging_fluentd_buffer_size_limit: 1m
+openshift_logging_fluentd_buffer_size_limit: 16m
 
 openshift_logging_es_host: logging-es
 openshift_logging_es_port: 9200
diff --git a/roles/openshift_logging_mux/defaults/main.yml b/roles/openshift_logging_mux/defaults/main.yml
index 35fc714..07cfd00 100644
--- a/roles/openshift_logging_mux/defaults/main.yml
+++ b/roles/openshift_logging_mux/defaults/main.yml
@@ -12,7 +12,7 @@ openshift_logging_mux_nodeselector: "{{ openshift_hosted_logging_mux_nodeselecto
 openshift_logging_mux_cpu_limit: 500m
 openshift_logging_mux_memory_limit: 2Gi
 openshift_logging_mux_buffer_queue_limit: 1024
-openshift_logging_mux_buffer_size_limit: 1m
+openshift_logging_mux_buffer_size_limit: 16m
 
 openshift_logging_mux_replicas: 1
```

ewolinetz:
@jcantrill is it possible to update that value? that change LGTM still

PR #4857, opened 2017-07-25
WIP: merge openshift-node and openshift-node-upgrade roles

ingvagabund:
**Individual refactor actions**
* move disable swap tasks into disable_swap.yml
* break node,tuned-profiles installation into with_items style
* install conntrack-tools in the same with_items task as the package has no deps on the node nor tuned-profiles pkgs

**TODO**
* [ ] replace `fail` with `assert`
* [x] should be `openshift.common.use_openshift_sdn` set to true by default (in both roles)? One PR for it
* [ ] can I replace all `openshift.common.node.VAR` with their corresponding variables? E.g. `openshift.node.env_vars` with `openshift_node_env_vars | default(None)`? One PR for it.
* [ ] replace all `openshift.common` like variables with their corresponding role variables
* [ ] create ovs role
* [ ] when generating a table of role variable, make a list of all openshift.common vars as well
* [ ] generate the checks for "a variable is defined"

I don't like this:
```yaml
- role: os_firewall
  os_firewall_allow:
  - service: Kubernetes service NodePort TCP
    port: "{{ openshift_node_port_range | default('') }}/tcp"
  - service: Kubernetes service NodePort UDP
    port: "{{ openshift_node_port_range | default('') }}/udp"
  when: openshift_node_port_range is defined
```

The `openshift_node_port_range` should not default to an empty string, the condition should be
```yaml
when: r_openshift_node_port_range != ""
```
and set to the empty string by default. Then, when the role is called, one just sets:
```yaml
r_openshift_node_port_range: "{{ openshift_node_port_range | default(omit) }}"
```

**Aditional info**
- I can move tasks with `notify` statement from a task yml into another cause all the handlers are triggered at the end of  #a role (unless `meta: flush_handlers` tasks is run explicitly)
- `'dns_ip' in openshift.node` and `openshift.node.dns_ip is defined` are equivalent constructs

**Children PRs**:
* [x] Set the node facts outside of the openshift-node role: https://github.com/openshift/openshift-ansible/pull/4912
* [x] Sync all openshift.common.use_openshift_sdn uses in yaml files: https://github.com/openshift/openshift-ansible/pull/4874
* [x] Backport 'Add systemctl daemon-reload handler to openshift_node' #4403 to openshift_node_upgrade:  https://github.com/openshift/openshift-ansible/pull/4897
* [x] Move common openshift node systemd unit tasks into separete files: https://github.com/openshift/openshift-ansible/pull/4900
* [ ] Introduce r_openshift_node_facts prefixed role variables: https://github.com/openshift/openshift-ansible/pull/5025
* [ ] Refactor openshift node role: https://github.com/openshift/openshift-ansible/pull/5027


roles/openshift_node/tasks/install.yml
```diff
@@ -0,0 +1,217 @@
...
+    dest: /etc/sysconfig/{{ openshift.common.service_type }}-node
+    regexp: "^{{ item.key }}="
+    line: "{{ item.key }}={{ item.value }}"
+    create: true
+  with_dict: "{{ openshift.node.env_vars | default({}) }}"
+  notify:
+    - restart node
+
+- name: NFS storage plugin configuration
+  include: storage_plugins/nfs.yml
```
kwoodson: Should include a when statement for when storage is nfs.

roles/openshift_node/tasks/install.yml
```diff
@@ -0,0 +1,217 @@
...
+
+# The atomic-openshift-node service will set this parameter on
+# startup, but if the network service is restarted this setting is
+# lost. Reference: https://bugzilla.redhat.com/show_bug.cgi?id=1372388
+#
+# Use lineinfile w/ a handler for this task until
+# https://github.com/ansible/ansible/pull/24277 is included in an
+# ansible release and we can use the sysctl module.
+- name: Persist net.ipv4.ip_forward sysctl entry
+  lineinfile: dest=/etc/sysctl.conf regexp='^net.ipv4.ip_forward' line='net.ipv4.ip_forward=1'
```
kwoodson: break into multiline

roles/openshift_node/tasks/install.yml
```diff
@@ -0,0 +1,217 @@
...
+    warn: no
+  register: l_tuned_installed
+  changed_when: false
+  failed_when: false
+
+- name: Set atomic-guest tuned profile
+  command: "tuned-adm profile atomic-guest"
+  when: l_tuned_installed.rc == 0 and openshift.common.is_atomic | bool
+
+- name: Install sdn-ovs package
```
kwoodson: suggest joining this package install with the one above.  Add a cond to the with_items and then place it in the when.

ingvagabund:
> When templating sysconfig files we will stomp any changes customers make. This goes back to editing vs templating. This does not need to be addressed in this PR but it is something to think about.

Main focus of the PR is to sync both roles to minimize file and code duplicates. I am open to your suggestion. Though, that is something we need to to globally. E.g. including openshift_master like roles.

> The other idea that I want to bring up would be to gather all of the is_containerized and place under one section so that there is a single place for that work. That might require duplicate code in some places but I feel it has good benefits:
> 
>     when we move away from rpms then entire code path can be removed or updated easily
>     its very easy to determine code paths when debugging
>     less logic when determining if something runs or not

Right, that already crossed my mind. The original code from which the `openshit_node_upgrade` role was created was going that direction. I am happy to update the PR to make it real.

At the moment I can't wait to test the refactoring itself if it is going to work at all :)
 
> Last point I'll ask is, does it make sense to separate out the openvswitch into their own files? I know openshift node encompasses openvswitch but it might be nice to abstract it into its own files so that when configuring the node we have the flexibility to configure opensvswitch separately or replace it with something else altogether. Thoughts?

Absolutely, the role still needs to be improved. Either to create new roles (e.g. `openvswitch`) or to reorder and break tasks into more files based on their semantics.

It is high likely I will create smaller PR to make the merging more obvious. This PR is growing into a monster eating time that is not easy to review.

mtnbikenc:
Superceeded by https://github.com/openshift/openshift-ansible/pull/6082

PR #4872, opened 2017-07-26
Set the node facts outside of the openshift-node role

ingvagabund:
The role currently refers to various `openshift.node` like variables. Some of the variables like `openshift.node.node_system_image` or `openshift.node.ovs_system_image` are set to default values (hidden under the hood). Some of the variables like `openshift.node.ovs_image` or `openshift.node.node_image`
are set to a user defined variables like `osn_ovs_image` (for the `ovs_image`) or `osn_image` (for the `node_image`).
    
The current `openshift_facts` implementation sets the `node|ovs_system_image` to `node|ovs_image`:

```python
if 'node_image' not in facts['node']:
    facts['node']['node_image'] = node_image
    facts['node']['node_system_image'] = node_image
if 'ovs_image' not in facts['node']:
    facts['node']['ovs_image'] = ovs_image
    facts['node']['ovs_system_image'] = ovs_image
```
Given the node facts are common for more roles, they should be collected outside of a role. Each role can then receive all its variables through role variables. Thus, making the role variable self-contained.

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -47,7 +47,8 @@
                                                     | union(groups['oo_etcd_to_config'] | default([])))
                                                 | oo_collect('openshift.common.hostname') | default([]) | join (',')
                                                 }}"
-
+  pre_tasks:
+  - include: set_node_facts.yml
   roles:
   - role: openshift_node
     openshift_ca_host: "{{ groups.oo_first_master.0 }}"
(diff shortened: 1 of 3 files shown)


ingvagabund:
aos-ci-test

ingvagabund:
Would be great to set the node facts at one place, here: https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-node/config.yml#L16. Though, not sure if there is anything going on in the second and the third play can lead to a different set of values when the node facts are set in the fourth play.

ingvagabund:
aos-ci-test

abutcher:
Any reason not to go with an openshift_node_facts type role that is a dependency of openshift_node?

ingvagabund:
> Any reason not to go with an openshift_node_facts type role that is a dependency of openshift_node?

Can't say no to it.

ingvagabund:
Closing the PR in favor of #4912

PR #4874, opened 2017-07-26
Sync all openshift.common.use_openshift_sdn uses in yaml files

ingvagabund:
Most occurrences are in a form:
    
```yaml
openshift.common.use_openshift_sdn | default(true) | bool
```
    
Let's make all occurences this way given the `use_openshift_sdn` is set to `true` anyway.
See https://github.com/openshift/openshift-ansible/blob/0c350dcc7d06d62be5ba3a8e468dff85cdd96dd7/roles/openshift_facts/library/openshift_facts.py#L2035

### DIFF
--- a/roles/openshift_node/handlers/main.yml
+++ b/roles/openshift_node/handlers/main.yml
@@ -3,7 +3,7 @@
   systemd:
     name: openvswitch
     state: restarted
-  when: (not skip_node_svc_handlers | default(False) | bool) and not (ovs_service_status_changed | default(false) | bool) and openshift.common.use_openshift_sdn | bool
+  when: (not skip_node_svc_handlers | default(False) | bool) and not (ovs_service_status_changed | default(false) | bool) and openshift.common.use_openshift_sdn | default(true) | bool
   register: l_openshift_node_stop_openvswitch_result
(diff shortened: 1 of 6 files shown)


ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

ingvagabund:
[merge]

roles/openshift_node/handlers/main.yml
```diff
@@ -3,7 +3,7 @@
   systemd:
     name: openvswitch
     state: restarted
-  when: (not skip_node_svc_handlers | default(False) | bool) and not (ovs_service_status_changed | default(false) | bool) and openshift.common.use_openshift_sdn | bool
+  when: (not skip_node_svc_handlers | default(False) | bool) and not (ovs_service_status_changed | default(false) | bool) and openshift.common.use_openshift_sdn | default(true) | bool
```
kwoodson: I'd prefer the multi-line when statement here with ands
```
when:
- condition1
- not condition2
- condition3
```

ingvagabund: Definitely. There is more such cases in the `opensift_node` role. I will cover that on the role refactoring.

PR #4876, opened 2017-07-26
Fixing podpresets perms for service-catalog-controller

ewolinetz:
Moving podpresets permissions to the correct permissions api group so that the service-catalog-controller can correctly create/delete/etc podpresets objects.

To address https://bugzilla.redhat.com/show_bug.cgi?id=1471881


ewolinetz:
aos-ci-test

PR #4881, opened 2017-07-26
Add glusterfs hosts to oo_all_hosts so that hosts set initial facts.

abutcher:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1474630

I was able to reproduce with settings similar to the BZ and this ensures that facts are accessible for glusterfs hosts.


sdodson:
aos-ci-test

jarrpa:
Isn't the glusterfs_registry group missing?

wkshi:
glusterfs_registry group missing +1 I think

PR #4882, opened 2017-07-27
Simplify generation of /etc/origin/node/resolv.conf

sdodson:
Don't attempt to copy /etc/resolv.conf whenever there's no watermark,
just write out the list of nameservers that network manager tells us
about like we do for origin-upstream-dns.conf

I think this will also fix the following bug which I think happens on upgrades when the watermark already exists and we weren't writing out the node's /etc/origin/node/resolv.conf
Maybe fixes https://bugzilla.redhat.com/show_bug.cgi?id=1474707


sdodson:
aos-ci-test

sdodson:
[test]

PR #4890, opened 2017-07-27
Simplify generation of /etc/origin/node/resolv.conf

sdodson:
Don't attempt to copy /etc/resolv.conf whenever there's no watermark,
just write out the list of nameservers that network manager tells us
about like we do for origin-upstream-dns.conf

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1474707


sdodson:
Backports #4890

PR #4894, opened 2017-07-27
BZ1469336 - registry dns value should be same with the exsiting masters for the added master

tbielawa:
```
when the OCP v3.6 was upgraded from v3.5 ,the registry dns wasn't enabled by default.  But if we scale
up master against this OCP 3.6 env, the registry dns was enabled on the added master. The new added
master configuration should use same with the existing master.
```

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1469336

----

* Detects if a host is a new master being scaled up into an existing cluster
* Queries first existing master and persists whatever the value of `OPENSHIFT_DEFAULT_REGISTRY` is on the first master. This may be *nothing at all* (i.e., it is not defined), or it may be an actual value.
* If the host is a scale up host, then during templating of the HA master `sysconfig` files (`atomic-openshift-master-controllers`, `atomic-openshift-master-api`) they will insert the persisted value for `OPENSHIFT_DEFAULT_REGISTRY` from the first master (which may mean inserting nothing at all, or an actual value)
* Hosts which are not new scale up hosts are not effected. They will insert a value for `OPENSHIFT_DEFAULT_REGISTRY` as long as `openshift_push_via_dns` has still evaluated to `True`

### DIFF
--- a/playbooks/common/openshift-master/config.yml
+++ b/playbooks/common/openshift-master/config.yml
@@ -5,6 +5,19 @@
     t_oo_option_master_debug_level: "{{ lookup('oo_option', 'openshift_master_debug_level') }}"
 
   pre_tasks:
+  # Per https://bugzilla.redhat.com/show_bug.cgi?id=1469336
+  #
+  # When scaling up a cluster upgraded from OCP <= 3.5, ensure that
+  # OPENSHIFT_DEFAULT_REGISTRY is present as defined on the existing
+  # masters, or absent if such is the case.
+  - name: Detect if this host is a new master in a scale up
+    set_fact:
+      g_openshift_master_is_scaleup: "{{ openshift.common.hostname in ( groups['new_masters'] | default([]) ) }}"
+
+  - name: Scaleup Detection
+    debug:
+      var: g_openshift_master_is_scaleup
+
   - name: Check for RPM generated config marker file .config_managed
     stat:
       path: /etc/origin/.config_managed
@@ -69,7 +82,7 @@
         ha: "{{ openshift_master_ha | default(groups.oo_masters | length > 1) }}"
         master_count: "{{ openshift_master_count | default(groups.oo_masters | length) }}"
 
-- name: Inspect state of first master session secrets and config
+- name: Inspect state of first master config settings
   hosts: oo_first_master
   roles:
   - role: openshift_facts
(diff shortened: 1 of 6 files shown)


tbielawa:
[test]

tbielawa:
aos-ci-test

roles/openshift_master/tasks/systemd_units.yml
```diff
@@ -23,7 +23,7 @@
   when: openshift.common.is_containerized | bool and not openshift.common.is_master_system_container | bool
 
 # workaround for missing systemd unit files
-- name: Create the systemd unit files
+- name: "Create the {{ openshift.common.service_type }} systemd unit file"
```
sdodson: Does this actually evaluate? In the past I don't think it did.

playbooks/common/openshift-master/config.yml
```diff
@@ -98,6 +111,24 @@
...
+  - name: Preserve OPENSHIFT_DEFAULT_REGISTRY master-controllers parameter if present
+    command: awk '/^OPENSHIFT_DEFAULT_REGISTRY/' /etc/sysconfig/atomic-openshift-master-controllers
+    register: l_default_registry_defined_controllers
+
+  - name: Update facts with OPENSHIFT_DEFAULT_REGISTRY value
+    set_fact:
+      l_default_registry_value: "{{ l_default_registry_defined.stdout }}"
+      l_default_registry_value_api: "{{ l_default_registry_defined_api.stdout }}"
+      l_default_registry_value_controllers: "{{ l_default_registry_defined_controllers.stdout }}"
+
```
sdodson: The value will always be the same, at least for the foreseeable future. Can we just check for it to be set in any of the above files and if so then set openshift_push_via_dns=true when calling the role?

roles/openshift_master/tasks/systemd_units.yml
ingvagabund: It does, at least it has always worked for me.

playbooks/common/openshift-master/config.yml
```diff
@@ -5,6 +5,19 @@
...
 
   pre_tasks:
+  # Per https://bugzilla.redhat.com/show_bug.cgi?id=1469336
+  #
+  # When scaling up a cluster upgraded from OCP <= 3.5, ensure that
+  # OPENSHIFT_DEFAULT_REGISTRY is present as defined on the existing
+  # masters, or absent if such is the case.
+  - name: Detect if this host is a new master in a scale up
+    set_fact:
+      g_openshift_master_is_scaleup: "{{ ( openshift.common.hostname in ( groups['new_masters'] | default([]) ) ) | bool }}"
```
ingvagabund: Just a nit:
```
"{{ ( openshift.common.hostname in ( groups['new_masters'] | default([]) ) ) | bool }}"
    ^ are the extra parentheses needed?
```

Plus, should not the in operator return bool by default?

playbooks/common/openshift-master/config.yml
ingvagabund: The first master has been always used as a reference master no matter what is on the second, third, etc. master. User can change the value of `OPENSHIFT_DEFAULT_REGISTRY` to listen the docker registry on a different port. Then, the `OPENSHIFT_DEFAULT_REGISTRY` will be present but when just the `openshift_push_via_dns` gets set to `true `, the port will still be `5000`. Once the value is set, we should not rely on the default one.

sdodson: Ok, that's fine with me. Should we try to get rid of the duplication in the template though or would the additional conditional logic make it more confusing to read?

tbielawa:
aos-ci-test

tbielawa:
[test]

roles/openshift_master/tasks/systemd_units.yml
abutcher: I feel like vars in task names is relatively new. I noticed it working recently too.

tbielawa: Vars in task names have been around for quite a long time. I can confirm because the original bug report for it was against a module I wrote like 3 years ago which turned into a larger "oh snap, we need to evaluate those statements everywhere" bug.

tbielawa: Either way. I've ran this playbook dozens of times now and those have worked each time.

tbielawa: c63b2eb9c9638416db0dce87521375f1

playbooks/common/openshift-master/config.yml
tbielawa: Probably should return bool by default. I was just kinda covering my own butt just in case. I can try it without all of the extra paren's and the extra `...) | bool`

playbooks/common/openshift-master/config.yml
tbielawa: This whole section is to ensure

* if `OPENSHIFT_DEFAULT_REGISTRY=<default_value:here>` in your existing settings, then nothing really changes
* If there is NO VALUE OR REFERENCE TO `OPENSHIFT_DEFAULT_REGISTRY` AT ALL then no lines are added on the scaleup master
* If you have a value that is non-standard, that value is used instead of the default

roles/openshift_master/tasks/systemd_units.yml
abutcher: lol

playbooks/common/openshift-master/config.yml
tbielawa: testing now

tbielawa:
@ingvagabund reduced the extra paren's and extra ` | bool ` statement as you suggested. Tested locally, worked like a charm :cat2:

tbielawa:
@sdodson @abutcher @ingvagabund I have addressed the comments you all gave above

* Using vars in task names: **confirmed works**
* Setting facts for non-ha/api/controllers: @sdodson says "Ok, that's fine with me"
* @ingvagabund You were correct, the extra `(....) | bool` was not required. I have removed it

If there are no more comments then I will squash the last commit down and execute CI testing. Please reply if you :+1: this (and also add you review approved check-mark)

tbielawa:
[test]

tbielawa:
aos-ci-test

tbielawa:
sqashed.

tbielawa:
aos-ci-test

tbielawa:
[test]

roles/openshift_master/tasks/systemd_units.yml
mtnbikenc: Although vars in task names are cool and all, it makes it a bit trickier to find that troubling task in the codebase when you are looking at a task failure in the log.  No task name matches what you are looking for so head scratching ensues.  It may just be my way going about finding stuff.  Some name is better than having no name.

sdodson:
aos-ci-test

sdodson:
[merge]

tbielawa:
aos-ci-test

tbielawa:
[merge]

sdodson:
[merge][severity:blocker]

sdodson:
Please make sure that there's a release-3.6 backport opened up.

PR #4897, opened 2017-07-27
backport 'Add systemctl daemon-reload handler to openshift_node' #4403 to openshift_node_upgrade

ingvagabund:
Cause we want both openshift_node and openshift_node_upgrade to be synced

### DIFF
--- a/roles/openshift_node_upgrade/handlers/main.yml
+++ b/roles/openshift_node_upgrade/handlers/main.yml
@@ -29,3 +29,8 @@
   when:
   - (not skip_node_svc_handlers | default(False) | bool)
   - not (node_service_status_changed | default(false) | bool)
+
+# TODO(jchaloup): once it is verified the systemd module works as expected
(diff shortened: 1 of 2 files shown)


ingvagabund:
aos-ci-test

roles/openshift_node_upgrade/handlers/main.yml
```diff
@@ -29,3 +29,6 @@
   when:
   - (not skip_node_svc_handlers | default(False) | bool)
   - not (node_service_status_changed | default(false) | bool)
+
+- name: reload systemd units
```
kwoodson: Can we add a note here to use the ansible module when it is supported or use it if it is already supported?
http://docs.ansible.com/ansible/latest/systemd_module.html

ingvagabund: Added that as a TODO item.

ingvagabund:
aos-ci-test

ingvagabund:
[merge]

PR #4899, opened 2017-07-27
Misc cleanup and input normalization

rhcarvalho:
Taking some commits out of #4570 to make it easier to review.

### DIFF
--- a/playbooks/common/openshift-checks/health.yml
+++ b/playbooks/common/openshift-checks/health.yml
@@ -1,8 +1,5 @@
 ---
-# openshift_health_checker depends on openshift_version which now requires group eval.
 - include: ../openshift-cluster/evaluate_groups.yml
-  tags:
-  - always
 
 - name: Run OpenShift health checks
   hosts: OSEv3
@@ -11,6 +8,6 @@
   vars:
   - r_openshift_health_checker_playbook_context: "health"
   post_tasks:
-  - action: openshift_health_check  # https://github.com/ansible/ansible/issues/20513
+  - action: openshift_health_check
     args:
       checks: ['@health']
--- a/playbooks/common/openshift-checks/health.yml
+++ b/playbooks/common/openshift-checks/health.yml
@@ -6,7 +6,7 @@
   roles:
   - openshift_health_checker
   vars:
-  - r_openshift_health_checker_playbook_context: "health"
+  - r_openshift_health_checker_playbook_context: health
   post_tasks:
   - action: openshift_health_check
     args:
--- a/roles/openshift_health_checker/action_plugins/openshift_health_check.py
+++ b/roles/openshift_health_checker/action_plugins/openshift_health_check.py
@@ -13,6 +13,7 @@ except ImportError:
     display = Display()
 
 from ansible.plugins.action import ActionBase
+from ansible.module_utils.six import string_types
 
 # Augment sys.path so that we can import checks from a directory relative to
 # this callback plugin.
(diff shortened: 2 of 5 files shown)


playbooks/common/openshift-checks/health.yml
```diff
@@ -1,8 +1,5 @@
 ---
-# openshift_health_checker depends on openshift_version which now requires group eval.
 - include: ../openshift-cluster/evaluate_groups.yml
-  tags:
-  - always
```
rhcarvalho: If `--tags` is passed in the command line, it makes no sense to evaluate groups and then don't run anything else. AFAICT we don't really need tags.

playbooks/common/openshift-checks/health.yml
```diff
@@ -11,6 +8,6 @@
   vars:
   - r_openshift_health_checker_playbook_context: "health"
   post_tasks:
-  - action: openshift_health_check  # https://github.com/ansible/ansible/issues/20513
+  - action: openshift_health_check
```
rhcarvalho: The referenced issue was closed, seems like a wontfix. Just use the old syntax and make no noise about it.

roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -27,7 +30,7 @@ def is_first_master(self):
         return masters[0] == hostname
 
     def run(self):
-        pass
+        return {}
```
rhcarvalho: I believe the more appropriate change would be to make this a `LoggingMixin` or such, and then this class wouldn't be listed as a check. For now, simply making it comply with the API and return a dict, avoiding an unhandled exception.

rhcarvalho:
aos-ci-test

playbooks/common/openshift-checks/health.yml
sosiouxme: Seemed bcoca was saying it would work if there were a module file for the action plugin. Not sure. Anyway this is fine.

roles/openshift_health_checker/action_plugins/openshift_health_check.py
```diff
@@ -134,3 +133,14 @@ def resolve_checks(names, all_checks):
...
+
+def normalize(checks):
+    """Return a clean list of check names.
+
+    The input may be a comma-separated string or a sequence. Leading and
+    trailing whitespace characters are removed. Empty items are discarded.
+    """
+    if isinstance(checks, string_types):
+        checks = checks.split(',')
+    return [name.strip() for name in checks if name.strip()]
```
sosiouxme: works even if they pass in a hash :) Although not if it's a number or list of numbers... If you want to be really paranoid here maybe:

    return [str(name).strip() for name in checks if str(name).strip()]

rhcarvalho: Ah, this made more sense in the context of the adhoc playbook.

`ansible-playbook ... -e openshift_checks=foo,bar,baz` passes in a string.

When working on it, I thought it would be useful to handle YAML/JSON as well, because we can write `-e {"openshift_checks": ["foo","bar","baz"]}` -- not something I would do if I can write a comma-sep string... but this becomes useful if we write it to a file, like `vars/main.yml`:

```yaml
# vars/main.yml
---
openshift_checks:
  - foo
  - bar
  - baz
```

The user needs to try really hard to pass in a list of numbers...

rhcarvalho:
[merge]

sosiouxme:
woo we have a working merge queue again :tada:

PR #4900, opened 2017-07-27
Move common openshift node systemd unit tasks into separete files

ingvagabund:
Sync what we can so it is easier to merge both `openshift_node` and `openshift_node_upgrade` roles.

Depends on https://github.com/openshift/openshift-ansible/pull/4897. Once that one gets merged, I will rebase.

### DIFF
--- a/roles/openshift_node/tasks/config/configure-node-settings.yml
+++ b/roles/openshift_node/tasks/config/configure-node-settings.yml
@@ -1,0 +1,16 @@
+---
+- name: Configure Node settings
+  lineinfile:
+    dest: /etc/sysconfig/{{ openshift.common.service_type }}-node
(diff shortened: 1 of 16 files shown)


ingvagabund:
aos-ci-test

ashcrow:
@ingvagabund looking through it seems like the files between node and upgrade are the same. Is there any reason to keep them separate entities if the contents equal?

ingvagabund:
> @ingvagabund looking through it seems like the files between node and upgrade are the same. Is there any reason to keep them separate entities if the contents equal?

I need to sync both roles as much as I can before I merge both `openshift_node` and `openshift_node_upgrade` roles.

ingvagabund:
aos-ci-test

ingvagabund:
[merge]

PR #4906, opened 2017-07-27
Updating template parameter replica to be more unique to avoid var sc…

ewolinetz:
…ope creeping

This may address the issue being seen where the ES components are being deployed with replica count of 2 instead of 1 seen in https://bugzilla.redhat.com/show_bug.cgi?id=1468731

cc @sosiouxme

### DIFF
--- a/roles/openshift_logging_elasticsearch/templates/es.j2
+++ b/roles/openshift_logging_elasticsearch/templates/es.j2
@@ -8,7 +8,7 @@ metadata:
     deployment: "{{deploy_name}}"
     logging-infra: "{{logging_component}}"
 spec:
-  replicas: {{replicas|default(1)}}
+  replicas: {{es_replicas|default(1)}}
   selector:
     provider: openshift
     component: "{{component}}"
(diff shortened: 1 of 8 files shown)


ewolinetz:
aos-ci-test

roles/openshift_logging_elasticsearch/templates/es.j2
```diff
@@ -8,7 +8,7 @@ metadata:
     deployment: "{{deploy_name}}"
     logging-infra: "{{logging_component}}"
 spec:
-  replicas: {{replicas|default(1)}}
+  replicas: {{es_replicas|default(1)}}
```
sosiouxme: ... when would you even have a case where you wanted something other than 1 ES replica?

ewolinetz: eventually with statefulsets or if you are using host mounts for storage you could just bump up the replicas instead of having multiple DC

ewolinetz:
[merge]

PR #4911, opened 2017-07-27
[3.6] Updating template parameter replica to be more unique to avoid var sc…

ewolinetz:
…ope creeping

Backport of https://github.com/openshift/openshift-ansible/pull/4906


ewolinetz:
aos-ci-test

PR #4912, opened 2017-07-27
Move node facts to new openshift_node_facts role.

abutcher:
Alternative to #4872

### DIFF
--- a/playbooks/common/openshift-node/config.yml
+++ b/playbooks/common/openshift-node/config.yml
@@ -1,25 +1,4 @@
 ---
-- name: Gather and set facts for node hosts
-  hosts: oo_nodes_to_config
-  vars:
-    t_oo_option_node_debug_level: "{{ lookup('oo_option', 'openshift_node_debug_level') }}"
-  pre_tasks:
-  - set_fact:
-      openshift_node_debug_level: "{{ t_oo_option_node_debug_level }}"
-    when: openshift_node_debug_level is not defined and t_oo_option_node_debug_level != ""
-  roles:
-  - openshift_facts
-  tasks:
-  # Since the master is generating the node certificates before they are
-  # configured, we need to make sure to set the node properties beforehand if
-  # we do not want the defaults
-  - openshift_facts:
-      role: node
-      local_facts:
-        labels: "{{ openshift_node_labels | default(None) }}"
-        annotations: "{{ openshift_node_annotations | default(None) }}"
-        schedulable: "{{ openshift_schedulable | default(openshift_scheduleable) | default(None) }}"
-
 - name: Evaluate node groups
   hosts: localhost
   become: no
@@ -32,7 +11,11 @@
       ansible_ssh_user: "{{ g_ssh_user | default(omit) }}"
       ansible_become: "{{ g_sudo | default(omit) }}"
     with_items: "{{ groups.oo_nodes_to_config | default([]) }}"
-    when: hostvars[item].openshift is defined and hostvars[item].openshift.common is defined and hostvars[item].openshift.common.is_containerized | bool and (item in groups.oo_nodes_to_config and item in groups.oo_masters_to_config)
+    when:
+    - hostvars[item].openshift is defined
+    - hostvars[item].openshift.common is defined
+    - hostvars[item].openshift.common.is_containerized | bool
+    - (item in groups.oo_nodes_to_config and item in groups.oo_masters_to_config)
     changed_when: False
 
 - name: Configure containerized nodes
--- a/roles/openshift_node_facts/tasks/main.yml
+++ b/roles/openshift_node_facts/tasks/main.yml
@@ -1,0 +1,34 @@
+---
+- set_fact:
+    openshift_node_debug_level: "{{ lookup('oo_option', 'openshift_node_debug_level') }}"
+  when:
+  - openshift_node_debug_level is not defined
+  - lookup('oo_option', 'openshift_node_debug_level') != ""
+
+- name: Set node facts
+  openshift_facts:
+    role: "{{ item.role }}"
+    local_facts: "{{ item.local_facts }}"
+  with_items:
+  # Reset node labels to an empty dictionary.
+  - role: node
+    local_facts:
+      labels: {}
+  - role: node
+    local_facts:
+      annotations: "{{ openshift_node_annotations | default(none) }}"
+      debug_level: "{{ openshift_node_debug_level | default(openshift.common.debug_level) }}"
+      iptables_sync_period: "{{ openshift_node_iptables_sync_period | default(None) }}"
+      kubelet_args: "{{ openshift_node_kubelet_args | default(None) }}"
+      labels: "{{ lookup('oo_option', 'openshift_node_labels') | default( openshift_node_labels | default(none), true) }}"
+      registry_url: "{{ oreg_url_node | default(oreg_url) | default(None) }}"
+      schedulable: "{{ openshift_schedulable | default(openshift_scheduleable) | default(None) }}"
+      sdn_mtu: "{{ openshift_node_sdn_mtu | default(None) }}"
+      storage_plugin_deps: "{{ osn_storage_plugin_deps | default(None) }}"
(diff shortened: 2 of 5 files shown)


abutcher:
aos-ci-test

roles/openshift_node_facts/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
...
+    openshift_node_debug_level: "{{ lookup('oo_option', 'openshift_node_debug_level') }}"
+  when:
+  - openshift_node_debug_level is not defined
+  - lookup('oo_option', 'openshift_node_debug_level') != ""
+
+- name: Set node facts
+  openshift_facts:
+    role: "{{ item.role }}"
+    local_facts: "{{ item.local_facts }}"
+  with_items:
```
ingvagabund: Don't see much point for the `with_items` here.

playbooks/common/openshift-node/config.yml
```diff
@@ -1,25 +1,4 @@
...
-  pre_tasks:
-  - set_fact:
-      openshift_node_debug_level: "{{ t_oo_option_node_debug_level }}"
-    when: openshift_node_debug_level is not defined and t_oo_option_node_debug_level != ""
-  roles:
-  - openshift_facts
-  tasks:
-  # Since the master is generating the node certificates before they are
-  # configured, we need to make sure to set the node properties beforehand if
-  # we do not want the defaults
```
ingvagabund: First time I was reading this comment I was not sure if we can set all the facts at one place. Reading it for the second time and checking the plays between this play up to plays running the `openshift_node` role I can't see anything touching the node certificates. So I believe the comment is say that when running the `openshift-node/config.yml` playbook, the master has already generated the node certificates and we should update the node facts for that reason. If we do it right here or if we do it once we run the `openshift_node` role, the effect will be the same. Prove me otherwise.

playbooks/common/openshift-node/config.yml
```diff
@@ -32,7 +11,11 @@
       ansible_ssh_user: "{{ g_ssh_user | default(omit) }}"
       ansible_become: "{{ g_sudo | default(omit) }}"
     with_items: "{{ groups.oo_nodes_to_config | default([]) }}"
-    when: hostvars[item].openshift is defined and hostvars[item].openshift.common is defined and hostvars[item].openshift.common.is_containerized | bool and (item in groups.oo_nodes_to_config and item in groups.oo_masters_to_config)
+    when:
+    - hostvars[item].openshift is defined
+    - hostvars[item].openshift.common is defined
+    - hostvars[item].openshift.common.is_containerized | bool
+    - (item in groups.oo_nodes_to_config and item in groups.oo_masters_to_config)
```
ingvagabund: Thanks for that.

ingvagabund:
@abutcher unless there is anything else to be added to the new role, let's merge this.

abutcher:
@kwoodson These aren't being re-ran but just applied when the node role is applied so I would consider this roughly equivalent to the way node facts work now but more in line with the status quo since other roles have their own `_facts` counterparts. Moving the facts away from the node role let's us set them for @ingvagabund's purpose but I guess we ultimately just want these set somewhere else for now so that facts can be applied without running the `openshift_node` role. If we want to set everything up front that would also accomplish the same goal. It's just a matter of when that work will be available and if it blocks whatever @ingvagabund is up to.

ingvagabund:
@kwoodson this is a very useful change that simplifies the `openshift_node` like merging. Plus, the openshift_facts is now called only twice, not three times.

ingvagabund:
> My first concern is that openshift_facts is a dependency on almost everything that runs inside of openshift-ansible. I think we are calling openshift_facts around 156x in our current runs. I'd really like to move away from using it as I feel that it tightly couples roles together. Having said that, if at all possible I'd prefer to reduce the number of calls to openshift_facts if we can.

Currently, the `openshift_facts` role is called the same number of time as before this PR, i.e. once. The `openshift_facts` role is listed in the first play. After the role, a task in the same play collecting node facts is run. Later, in the third and the fourth play the `openshift_node` is run which runs another task collecting the node facts. With the PR applied, the node facts collecting is just moved inside a new `openshift_node_facts` role. And the same holds for the `openshift_facts` role. It can either stay in the first play and out of the new role. Or, it can be moved into the new role as its dependency. At the end, the number of times the `openshift_facts` role is called is the same. I am for decreasing the number of invocation of the `openshift_facts` role. However, this is not the place where we should decide what to do with the role. I would rather create a new PR in which we do that so the first (and hopefully global) change towards moving the `openshift_facts` role invocation and minimizing the number of its invocation is thoroughly tested and justified. After that, we can start massaging the role out of the remaining roles and places. I don't want the change to be made in million PRs. It will be harder to track it. First, one colossal PR. Then, followed by smaller ones. 
 
> One solution to reduce the number of calls to openshift_facts is to call it a single time at the beginning of a run, setup all the facts required, and not require it to be run again. That is something I'm trying to accomplish in #4739. This also removes the tasks from openshift_facts as they aren't required to run over and over. This reduces dependencies as well as the time of a run by upto 6 full minutes on an 8 node cluster. This work also coincides with the work @mtnbikenc is doing to define entry points so that everything is probably established before a run kicks off.

I am taking this PR as a middle step before we can eliminate/minimize invocation of the `openshift_facts` role. This way, we are moving the "hidden" dependency in a play into a visible role dependency. I am not against removing the `openshift_facts` dependency from the new role. However, I would rather keep the dependency as it is intended in the PR rather than keep the `openshift_facts` role in the play and hope that someone will notice the role needs to be called/included before the `openshift_node_facts` is.

> There are two pieces to openshift_facts. One is the tasks list and one is the module that sets facts. The module that sets facts can be called whenever needed as it acts like a library. I would still recommend reducing the number of openshift_facts calls as this goes out and reads/writes files on each host that it runs. This is not a free operation. If this cannot be avoided then I recommend avoiding the cost of running the tasks list in openshift_facts role and just calling the module directly.

I am not against extending the `openshift_facts` with a `noop` action that will "just" include the library instead of running the `main.yml` tasks each time the role is invoked. 

> Can these facts be run up front in the ansible run? If so this method is the preferred one.

> Are these facts being rerun because they are changing often? If facts are changing I'd be curious to know why and if they need to be persisted each time. It seems like we continue to refresh the facts at every role and that they wouldn't be that out-of-date. If they are just required at the time of the role run, then I'm not sure what the benefit is to persist them each time. If openshift_facts is determining something dynamically then that probably adds value. I'd still recommend placing whatever that fact is under the same role so they become self contained.

That is something for a spike. I believe >90% facts will be static.

> I'd really prefer if we remove the openshift_node_facts role in favor of a facts.yml underneath the same role. This would simply be the tasks from openshift_node_role/tasks/main.yml pushed into a single file inside of the same role. Then include that facts.yml at the beginning if needed. This accomplishes the same goal but reducing the number of roles and their dependencies. Thoughts?

I am trying to avoid running:

```yaml
include_role:
  name: openshift_node
  tasks_from: facts.yml
include_role:
  name: openshift_node
```

I still prefer to run the following instead:
```yaml
include_role:
  name: openshift_node_facts
include_role:
  name: openshift_node
```

Plus, it is easier to test all the openshift*facts role at once rather then testing individual role task files. Usually, things are going to change more often inside `openshift_node` roles than in `openshift_facts_node`. If we start role based testing, we only need to test the roles that has changed.

However, why not. Still, something we could discuss on the architecture meeting before we start applying that pattern everywhere.

kwoodson:
@ingvagabund, We can discuss this more in an architecture meeting.  Thanks for answering my questions.  I'm not going to block this PR as I think it is necessary to continue moving forward iteratively but I will state that I'm not in 100% agreement on the current direction.  

I want to see an openshift-ansible with less dependencies and less facts modules.  If facts are needed then they are run a single time and front loaded.  This desire for less dependencies will reduce coupling, reduce dependency loops, reduce undesired dependency chains, increase modularity, increase flexibility, etc.

If you feel like this greatly helps you and your current work then please proceed forward.  Let's make a note of this topic for the architecture meeting and discuss it.

ingvagabund:
[merge]

PR #4913, opened 2017-07-27
openshift_checks: refactor check results

sosiouxme:
Introduced the 'changed' property for checks that can make changes to track whether they did or not. Rather than the check's own logic having to track this and include it in the result hash, just set the property and have the action plugin insert it in the result hash after running (even if there is an exception).

Cleared out a lot of crufty "changed: false" hash entries.

Refactored logging checks:
- Turned failure messages into exceptions.
- Have tests look at exceptions raised instead of text intended for users.
- Turned logging_namespace property into a method.
- Got rid of _exec_oc and just use logging.exec_oc.

### DIFF
--- a/roles/openshift_health_checker/action_plugins/openshift_health_check.py
+++ b/roles/openshift_health_checker/action_plugins/openshift_health_check.py
@@ -68,13 +68,15 @@ class ActionModule(ActionBase):
                         msg=str(e),
                     )
 
+            if check.changed:
+                r["changed"] = True
             check_results[check_name] = r
 
-            if r.get("failed", False):
-                result["failed"] = True
-                result["msg"] = "One or more checks failed"
+        result["changed"] = any(r.get("changed") for r in check_results.values())
+        if any(r.get("failed") for r in check_results.values()):
+            result["failed"] = True
+            result["msg"] = "One or more checks failed"
 
-        result["changed"] = any(r.get("changed", False) for r in check_results.values())
         return result
 
     def load_known_checks(self, tmp, task_vars):
--- a/roles/openshift_health_checker/openshift_checks/__init__.py
+++ b/roles/openshift_health_checker/openshift_checks/__init__.py
@@ -34,6 +34,8 @@ class OpenShiftCheck(object):
         self._execute_module = execute_module
         self.task_vars = task_vars or {}
         self.tmp = tmp
+        # set True when the check makes a change to the host so it can be reported to the user:
+        self.changed = False
 
     @abstractproperty
     def name(self):
--- a/roles/openshift_health_checker/openshift_checks/__init__.py
+++ b/roles/openshift_health_checker/openshift_checks/__init__.py
@@ -13,8 +13,30 @@ from ansible.module_utils.six.moves import reduce  # pylint: disable=import-erro
 
 
 class OpenShiftCheckException(Exception):
-    """Raised when a check cannot proceed."""
-    pass
+    """Raised when a check encounters a failure condition."""
+
+    def __init__(self, name, msg=None):
+        # msg is for the message the user will see when this is raised.
+        # name is for test code to identify the error without looking at msg text.
+        if msg is None:  # for parameter backward compatibility
+            msg = name
+            name = self.__class__.__name__
+        self.name = name
+        super(OpenShiftCheckException, self).__init__(msg)
+
+
+class OpenShiftCheckExceptionList(OpenShiftCheckException):
+    """A container for multiple logging errors that may be detected in one check."""
+    def __init__(self, errors):
+        self.errors = errors
+        super(OpenShiftCheckExceptionList, self).__init__(
+            'OpenShiftCheckExceptionList',
+            '\n'.join(str(msg) for msg in errors)
+        )
+
+    # make iterable
+    def __getitem__(self, index):
+        return self.errors[index]
 
 
 @six.add_metaclass(ABCMeta)
@@ -34,7 +56,8 @@ class OpenShiftCheck(object):
         self._execute_module = execute_module
         self.task_vars = task_vars or {}
         self.tmp = tmp
-        # set True when the check makes a change to the host so it can be reported to the user:
+
+        # set to True when the check changes the host, for accurate total "changed" count
         self.changed = False
 
     @abstractproperty
--- a/roles/openshift_health_checker/openshift_checks/logging/curator.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/curator.py
@@ -1,6 +1,6 @@
 """Check for an aggregated logging Curator deployment"""
 
-from openshift_checks.logging.logging import LoggingCheck
+from openshift_checks.logging.logging import OpenShiftCheckException, LoggingCheck
 
 
 class Curator(LoggingCheck):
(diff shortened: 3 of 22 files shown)


roles/openshift_health_checker/action_plugins/openshift_health_check.py
```diff
@@ -69,13 +69,15 @@ def run(self, tmp=None, task_vars=None):
                         msg=str(e),
                     )
 
+            if check.changed:
+                r["changed"] = True
```
rhcarvalho: This is okay. Have you considered setting `r['changed'] = check.changed` unconditionally?

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -13,8 +13,16 @@
...
 
 class OpenShiftCheckException(Exception):
-    """Raised when a check cannot proceed."""
-    pass
+    """Raised when a check encounters a failure condition."""
+
+    def __init__(self, name, msg=None):
+        # msg is for the message the user will see when this is raised.
+        # name is for test code to identify the error without looking at msg text.
+        if msg is None:  # for parameter backward compatibility
```
rhcarvalho: I think if we swap the order of the arguments we can avoid this little hack?

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -34,6 +42,8 @@ def __init__(self, execute_module=None, task_vars=None, tmp=None):
         self._execute_module = execute_module
         self.task_vars = task_vars or {}
         self.tmp = tmp
+        # set True when the check makes a change to the host so it can be reported to the user:
```
rhcarvalho: s/set True/set to True/ ?

s/so it can be reported to the user// ? "reporting to the user" is kind of optional, we don't need to make promises in this comment, but I'll leave it up to you, I'm fine either way.

roles/openshift_health_checker/openshift_checks/logging/curator.py
```diff
@@ -9,46 +9,33 @@ class Curator(LoggingCheck):
...
-
-        if check_error:
-            msg = ("The following Curator deployment issue was found:"
-                   "\n-------\n"
-                   "{}".format(check_error))
-            return {"failed": True, "changed": False, "msg": msg}
-
+        curator_pods = self.get_pods_for_component("curator")
+        self.check_curator(curator_pods)
         # TODO(lmeyer): run it all again for the ops cluster
```
rhcarvalho: Is this still relevant?

roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -8,11 +8,33 @@
...
+    """Raised when a component has no pods in the namespace."""
+    pass
+
+
+class CouldNotUseOc(OpenShiftCheckException):
+    """Raised when ocutil has a failure running oc."""
+    pass
+
+
+class LoggingErrorList(OpenShiftCheckException):
```
rhcarvalho: Perhaps this doesn't need to have "Logging" in the name? It could as well sit next to "OpenShiftCheckException" and be  an alternative to it? `class OpenShiftCheckExceptionList(OpenShiftCheckException)`?

roles/openshift_health_checker/action_plugins/openshift_health_check.py
sosiouxme: Yes, but I thought it wouldn't hurt to allow the check to continue to return it in the result hash if desired for some reason. But thinking about it, why complicate what is really quite simple... "one way to do it" is probably best here.

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: Of course, but I like to have the concise identifier first and then the blah blah blah when I'm looking at the code raising an exception. Selfish of me? And for once I thought it would be nice not to force a total transformation to the new method signature.

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: set to True :+1: 

"report to user" right that can be worded better. All we're doing is marking the task as having made a change so that the sum total of "changed" tasks can be reported at the end of the run. As if anyone cares (but there's always someone). That's all that's "reported" to the user unless they're looking at `-vvv` output.

        # set to True when the check makes a change to the host so the total "changed" count is accurate
￼

roles/openshift_health_checker/openshift_checks/logging/curator.py
sosiouxme: Sadly yes... these checks do nothing to examine the ops cluster. The main cluster is far more likely to fall over and affect users though so the focus is right. These little reminders are where they are because ideally we would literally just run the same high-level methods against a different set of pods/services/etc to do ops cluster checks.

roles/openshift_health_checker/openshift_checks/logging/logging.py
sosiouxme: I wasn't sure if it would be useful anywhere else. I'm not sure any other checks have a need to compile a list of failures instead of just quitting on the first. But again, thinking about it... it will probably come up. I guess I'll generalize it.

roles/openshift_health_checker/action_plugins/openshift_health_check.py
sosiouxme: Actually, I remembered why I did it this way. I don't want to clutter up every check result in `-vvv` output with "changed" when it won't matter for most. I think I'll leave it as-is.

roles/openshift_health_checker/openshift_checks/logging/logging.py
sosiouxme: That name is too long though :thinking:

sosiouxme:
@rhcarvalho you saw the bits I thought you'd find interesting. Thanks for the feedback.

I might run the "changed" commit in a different PR; the rest of this is more tedious than I expected.

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: I see :+1:

roles/openshift_health_checker/action_plugins/openshift_health_check.py
rhcarvalho: Right. We can omit from checks, but always set it in the task/action plugin level at most once.

Now whether to always include it on the task/action plugin level will depend on how Ansible interpret it when it is missing (default True or default False) ;-)

rhcarvalho:
aos-ci-test

rhcarvalho:
@sosiouxme if this is good to go please remove the [WIP] from the title ;)

roles/openshift_health_checker/action_plugins/openshift_health_check.py
sosiouxme: Pretty sure Ansible defaults to missing == False

For this code, it seemed simplest to always set it at the task level.

sosiouxme:
It can be good to go. There are a lot of other checks to get the same treatment but there's no need to hold up these changes for that.

sosiouxme:
[merge]

sosiouxme:
aos-ci-test

sosiouxme:
[merge]

sosiouxme:
Looks like flakes https://github.com/openshift/origin/issues/14829 and https://github.com/openshift/origin/issues/14898
re[merge]

sosiouxme:
https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/803/ indicates flakes https://github.com/openshift/origin/issues/8571 and https://github.com/openshift/origin/issues/10162

rhcarvalho:
Flake https://github.com/openshift/origin/issues/13977

rhcarvalho:
@sosiouxme okay for us to manually merge this after the recent changes that merged in?

sosiouxme:
@rhcarvalho yes that would be nice... enough test failures already.

rhcarvalho:
Going to merge manually as per https://github.com/openshift/openshift-ansible/blob/master/docs/pull_requests.md#manual-merges

PR #4915, opened 2017-07-27
Stop masters before backing up, block migration for supported configs

sdodson:
- Stop master api/controllers before we perform our backup
- Block migration in supported environments until we isolate a data inconsistency issue

See https://bugzilla.redhat.com/show_bug.cgi?id=1475351 for more info on the data issue

### DIFF
--- a/roles/etcd_migrate/tasks/check.yml
+++ b/roles/etcd_migrate/tasks/check.yml
@@ -1,4 +1,8 @@
 ---
+- fail:
+    msg: "Currently etcd v3 migration is unsupported while we test it more thoroughly"
+  when: not openshift_enable_unsupported_configurations | default(false) | bool
+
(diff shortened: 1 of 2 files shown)


roles/etcd_migrate/tasks/check.yml
```diff
@@ -1,4 +1,8 @@
 ---
+- fail:
+    msg: "Currently etcd v3 migration is unsupported while we test it more thoroughly"
+  when: not openshift_enable_unsupported_configurations | default(false) | bool
```
sdodson: @eparis how's this? We've used this variable elsewhere to ensure that they don't apply configuration variables that we consider to be unsupported.

eparis: that's a good enough name for me.

PR #4916, opened 2017-07-27
use mux_client_mode instead of use_mux_client

richm:
Instead of the `openshift_logging_use_mux_client` boolean parameter,
use `openshift_logging_mux_client_mode` which will allow us to support
different mux client use cases:
The value `full_no_k8s_meta` will cause Fluentd to perform all of the
processing *except* for the Kubernetes metadata processing, which will
be done by mux.
The value `minimal` means that Fluentd will do *no* processing at all,
and send the raw logs to mux for processing.

`MUX_ALLOW_EXTERNAL` is no longer needed in the mux dc.  mux now always
operates to process external logs.  The ansible setting
`openshift_logging_mux_allow_external` is still required in order to
set up the mux service to accept connections from outside of the
cluster.
@jcantrill @nhosoi @portante
This goes with https://github.com/openshift/origin-aggregated-logging/pull/553

### DIFF
--- a/roles/openshift_logging/README.md
+++ b/roles/openshift_logging/README.md
@@ -135,16 +135,23 @@ Elasticsearch OPS too, if using an OPS cluster:
   secure_forward forwarder for the node agent Fluentd daemonsets running in the
   cluster.  This can be used to reduce the number of connections to the
   OpenShift API server, by using `mux` and configuring each node Fluentd to
-  send raw logs to mux and turn off the k8s metadata plugin.
+  send raw logs to mux and turn off the k8s metadata plugin.  This requires the
+  use of `openshift_logging_mux_client_mode` (see below).
 - `openshift_logging_mux_allow_external`: Default `False`.  If this is `True`,
   the `mux` service will be deployed, and it will be configured to allow
   Fluentd clients running outside of the cluster to send logs using
   secure_forward.  This allows OpenShift logging to be used as a central
   logging service for clients other than OpenShift, or other OpenShift
   clusters.
-- `openshift_logging_use_mux_client`: Default `False`.  If this is `True`, the
-  node agent Fluentd services will be configured to send logs to the mux
-  service rather than directly to Elasticsearch.
+- `openshift_logging_mux_client_mode`: Values - `minimal`, `maximal`.
+  Default is unset.  Setting this value will cause the Fluentd node agent to
+  send logs to mux rather than directly to Elasticsearch.  The value
+  `maximal` means that Fluentd will do as much processing as possible at the
+  node before sending the records to mux.  This is the current recommended
+  way to use mux due to current scaling issues.
+  The value `minimal` means that Fluentd will do *no* processing at all, and
+  send the raw logs to mux for processing.  We do not currently recommend using
+  this mode, and ansible will warn you about this.
 - `openshift_logging_mux_hostname`: Default is "mux." +
   `openshift_master_default_subdomain`.  This is the hostname *external*_
   clients will use to connect to mux, and will be used in the TLS server cert
--- a/roles/openshift_logging_fluentd/defaults/main.yml
+++ b/roles/openshift_logging_fluentd/defaults/main.yml
@@ -48,7 +48,6 @@ openshift_logging_fluentd_aggregating_strict: "no"
 openshift_logging_fluentd_aggregating_cert_path: none
 openshift_logging_fluentd_aggregating_key_path: none
 openshift_logging_fluentd_aggregating_passphrase: none
-openshift_logging_use_mux_client: False
 
 ### Deprecating in 3.6
 openshift_logging_fluentd_es_copy: false
--- a/roles/openshift_logging_fluentd/tasks/main.yaml
+++ b/roles/openshift_logging_fluentd/tasks/main.yaml
@@ -23,6 +23,14 @@
     msg: openshift_hosted_logging_use_journal is deprecated.  Fluentd will automatically detect which logging driver is being used.
   when: openshift_hosted_logging_use_journal is defined
 
+- fail:
+    msg: Invalid openshift_logging_mux_client_mode [{{ openshift_logging_mux_client_mode }}], one of {{ __allowed_mux_client_modes }} allowed
+  when: openshift_logging_mux_client_mode is defined and not openshift_logging_mux_client_mode in __allowed_mux_client_modes
+
+- debug:
+    msg: WARNING Use of openshift_logging_mux_client_mode=minimal is not recommended due to current scaling issues
+  when: openshift_logging_mux_client_mode is defined and openshift_logging_mux_client_mode == 'minimal'
+
 - include: determine_version.yaml
 
 # allow passing in a tempdir
--- a/roles/openshift_logging_mux/defaults/main.yml
+++ b/roles/openshift_logging_mux/defaults/main.yml
@@ -28,6 +28,7 @@ openshift_logging_mux_journal_source: "{{ openshift_hosted_logging_journal_sourc
 openshift_logging_mux_journal_read_from_head: "{{ openshift_hosted_logging_journal_read_from_head | default('') }}"
 
 openshift_logging_mux_allow_external: False
+openshift_logging_use_mux: "{{ openshift_logging_mux_allow_external | default(False) }}"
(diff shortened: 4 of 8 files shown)


roles/openshift_logging_mux/defaults/main.yml
```diff
@@ -28,6 +28,7 @@ openshift_logging_mux_journal_source: "{{ openshift_hosted_logging_journal_sourc
 openshift_logging_mux_journal_read_from_head: "{{ openshift_hosted_logging_journal_read_from_head | default('') }}"
 
 openshift_logging_mux_allow_external: False
+openshift_logging_use_mux: "{{ openshift_logging_mux_allow_external | default(False) }}"
```
portante: Why do we reference `openshift_logging_mux_allow_external` instead of `openshift_logging_use_mux`?

roles/openshift_logging/README.md
```diff
@@ -134,16 +134,21 @@ Elasticsearch OPS too, if using an OPS cluster:
   secure_forward forwarder for the node agent Fluentd daemonsets running in the
   cluster.  This can be used to reduce the number of connections to the
   OpenShift API server, by using `mux` and configuring each node Fluentd to
-  send raw logs to mux and turn off the k8s metadata plugin.
+  send raw logs to mux and turn off the k8s metadata plugin.  This requires the
+  use of `openshift_logging_mux_client_mode` (see below).
 - `openshift_logging_mux_allow_external`: Default `False`.  If this is `True`,
```
portante: Aren't we getting rid of `openshift_logging_mux_allow_external`?

jcantrill: See original comment:

> openshift_logging_mux_allow_external is still required in order to
set up the mux service to accept connections from outside of the
cluster.

roles/openshift_logging_mux/defaults/main.yml
richm: because `openshift_logging_use_mux` means "use mux inside the cluster only, no outside connections"

roles/openshift_logging_fluentd/defaults/main.yml
```diff
@@ -48,7 +48,8 @@ openshift_logging_fluentd_aggregating_strict: "no"
 openshift_logging_fluentd_aggregating_cert_path: none
 openshift_logging_fluentd_aggregating_key_path: none
 openshift_logging_fluentd_aggregating_passphrase: none
-openshift_logging_use_mux_client: False
+# this tells the fluentd node agent to use mux instead of sending directly to Elasticsearch
+#openshift_logging_mux_client_mode: "full_no_k8s_meta" or "minimal"
```
jcantrill: @ewolinetz is there a better way to do this other then commenting it out... like setting it to 'none'?  I have not had a good experience knowing how to properly evaluate it in consuming jinja templates

roles/openshift_logging/README.md
richm: right - It was thought that we should be able to use mux without always having it listening and configured for external connections.

roles/openshift_logging_fluentd/defaults/main.yml
jcantrill: We should also probably stuff the values in vars/main.yml and verify them.  Though maybe there is a place to default a value should you pass one that isnt allowed?

richm: I have documented it in the README, but many users will just go straight to the "defaults/main.yml" to see which parameters are available

richm: I can add an explicit check in the fluentd playbook.

ewolinetz: the code is checking if its defined so setting it to `None` makes it defined... commenting it out is valid IMHO

roles/openshift_logging_mux/defaults/main.yml
portante: Doesn't this statement change the value of `openshift_logging_use_mux` to be whatever the value is for `openshift_logging_mux_allow_external` which is hard-coded on the line before?  Why would we do that?

richm: @portante because if the user has set `openshift_logging_mux_allow_external` that implies that the user wants to use mux, so we set `openshift_logging_use_mux`

richm:
@jcantrill added validation for mux_client_mode values

@portante added a warning about the use of `minimal`

ewolinetz:
aos-ci-test

roles/openshift_logging_mux/defaults/main.yml
portante: @richm, so if the user has mistakenly set `openshift_logging_mux_allow_external` to `true` (perhaps not realizing this behavior), and sets `openshift_logging_use_mux` to `false`, what will happen?

richm: I believe the way ansible works with defaults/main.yml is that if you explicitly set a parameter to a value, it will not use the value set in defaults/main.yml.  So in this case, the user will have set two parameters with conflicting values.  But AFAICT, if `openshift_logging_use_mux=false`, the mux component will not be deployed at all, so `openshift_logging_mux_allow_external` will be ignored.

sdodson: > I believe the way ansible works with defaults/main.yml is that if you explicitly set a parameter to a value, it will not use the value set in defaults/main.yml

Yes that's correct.

ewolinetz:
[merge][severity: blocker]

roles/openshift_logging_fluentd/tasks/main.yaml
```diff
@@ -23,6 +23,14 @@
     msg: openshift_hosted_logging_use_journal is deprecated.  Fluentd will automatically detect which logging driver is being used.
   when: openshift_hosted_logging_use_journal is defined
 
+- fail:
+    msg: Invalid openshift_logging_mux_client_mode [{{ openshift_logging_mux_client_mode }}], one of {{ __allowed_mux_client_modes }} allowed
+  when: openshift_logging_mux_client_mode is defined and not openshift_logging_mux_client_mode in __allowed_mux_client_modes
+
+- debug:
```
ewolinetz: Saw this error when testing for the merge:
```
TASK [openshift_logging_fluentd : debug] ***************************************
task path: /usr/share/ansible/openshift-ansible/roles/openshift_logging_fluentd/tasks/main.yaml:30
    "failed": true, 
    "generated_timestamp": "2017-07-31 13:56:09.745136", 
    "msg": "The conditional check 'openshift_logging_mux_client_mode == 'minimal'' failed. The error was: error while evaluating conditional (openshift_logging_mux_client_mode == 'minimal'): 'openshift_logging_mux_client_mode' is undefined\n\nThe error appears to have been in '/usr/share/ansible/openshift-ansible/roles/openshift_logging_fluentd/tasks/main.yaml': line 30, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- debug:\n  ^ here\n"
}
```

ewolinetz: @richm can we change this `debug` and the prior `fail` to be an `assert` call instead?

richm: @ewolinetz does assert abort the run?  I need two different behaviors

- If the user has defined `openshift_logging_mux_client_mode` and the value is not valid, I want to abort the ansible play with an error message
- If the user has defined `openshift_logging_mux_client_mode` and the value is `minimal`, I want to show a warning message, but continue the play

can I do that with `assert`?  If so, how?

ewolinetz: @richm ah, no we can't do that with assert... 
In that case just resolving that error seen will suffice

ewolinetz:
aos-ci-test

ewolinetz:
ideally those `and` separated `when` conditionals would be separate lines but that can be done in a follow up

ewolinetz:
@richm travis errors to address

ewolinetz:
aos-ci-test

ewolinetz:
[test]

ewolinetz:
@sdodson halp

ewolinetz:
[merge]

PR #4919, opened 2017-07-28
Some automation fixes for contiv GUI/auth_proxy & vxlan mode

srampal:
Fixes to the automation for the auth_proxy/ Contiv GUI function and setup needed for VxLan mode. Also requires updating to Contiv binary to 1.1.1 in order to pick these fixes.


srampal:
@sdodson @jcpowermac PTAL

roles/contiv_auth_proxy/files/key.pem
```diff
@@ -0,0 +1,51 @@
+-----BEGIN RSA PRIVATE KEY-----
```
jcpowermac: @srampal shouldn't these keys be generated?

srampal: @jcpowermac Yes, ideally they should and I will open a bug fix PR post the 3.6 relese to make that change. For now, since we are so close to the code cutoff date and really need the rest of these fixes to get work in 3.6, we are fine with using a pre-defined certificate and in addition we will also point the user to our contiv script for generating certificates/ keys which they can just copy into the files directory if they really need new unique certificates. However we really need the auth_proxy to work in the 3.6 release right away (without which the contiv GUI doesnt come up at all in 3.6) hence using the pre-created certificate for now will allow the GUI to work so that Contiv works correctly in the 3.6 release build. In our docs we will also point users to this script for generating new certificates if they need to do so https://github.com/contiv/auth_proxy/blob/master/scripts/generate_certificate.sh
and in a later bug fix PR post the 3.6 release cutoff we will integrate that into the contiv ansible as well.

sdodson:
aos-ci-test

PR #4922, opened 2017-07-28
openshift_checks: enable variable conversion

sosiouxme:
get_var can now do type conversion and raise an error if it doesn't work.

Later this could be integrated with defined check parameters to give the user more info about what was expected. Also check parameters could have indications for whether they're deprecated and defined/used.

Finally somewhere in here we could deal with the lazy evaluation of jinja2 in variables.

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/__init__.py
+++ b/roles/openshift_health_checker/openshift_checks/__init__.py
@@ -10,6 +10,7 @@ from importlib import import_module
 
 from ansible.module_utils import six
 from ansible.module_utils.six.moves import reduce  # pylint: disable=import-error,redefined-builtin
+from ansible.plugins.filter.core import to_bool as ansible_to_bool
 
 
 class OpenShiftCheckException(Exception):
@@ -94,16 +95,59 @@ class OpenShiftCheck(object):
 
         Ansible task_vars structures are Python dicts, often mapping strings to
         other dicts. This helper makes it easier to get a nested value, raising
-        OpenShiftCheckException when a key is not found or returning a default value
-        provided as a keyword argument.
+        OpenShiftCheckException when a key is not found.
+
+        Keyword args:
+          default:
+            On missing key, return this as default value instead of raising exception.
+          convert:
+            Supply a function to apply to normalize the value before returning it.
+            None is the default (return as-is).
+            This function should raise ValueError if the user has provided a value
+            that cannot be converted, or OpenShiftCheckException if some other
+            problem needs to be described to the user.
         """
+        if len(keys) == 1:
+            keys = keys[0].split(".")
+
         try:
             value = reduce(operator.getitem, keys, self.task_vars)
         except (KeyError, TypeError):
-            if "default" in kwargs:
-                return kwargs["default"]
-            raise OpenShiftCheckException("'{}' is undefined".format(".".join(map(str, keys))))
-        return value
+            if "default" not in kwargs:
+                raise OpenShiftCheckException(
+                    "This check expects the '{}' inventory variable to be defined\n"
+                    "in order to proceed, but it is undefined. There may be a bug\n"
+                    "in Ansible, the checks, or their dependencies."
+                    "".format(".".join(map(str, keys)))
+                )
+            value = kwargs["default"]
+
+        convert = kwargs.get("convert", None)
+        try:
+            if convert is None:
+                return value
+            elif convert is bool:  # interpret bool as Ansible does, instead of python truthiness
+                return ansible_to_bool(value)
+            else:
+                return convert(value)
+
+        except ValueError as error:  # user error in specifying value
+            raise OpenShiftCheckException(
+                'Cannot convert inventory variable to expected type:\n'
+                '  "{var}={value}"\n'
+                '{error}'.format(var=".".join(keys), value=value, error=error)
+            )
+
+        except OpenShiftCheckException:  # some other check-specific problem
+            raise
+
+        except Exception as error:  # probably a bug in the function
+            raise OpenShiftCheckException(
+                'There is a bug in this check. While trying to convert variable \n'
+                '  "{var}={value}"\n'
+                'the given converter cannot be used or failed unexpectedly:\n'
+                '{error}'.format(var=".".join(keys), value=value, error=error)
+            )
 
 
 LOADER_EXCLUDES = (
--- a/roles/openshift_health_checker/openshift_checks/logging/logging.py
+++ b/roles/openshift_health_checker/openshift_checks/logging/logging.py
@@ -18,7 +18,7 @@ class LoggingCheck(OpenShiftCheck):
     logging_namespace = "logging"
 
     def is_active(self):
-        logging_deployed = self.get_var("openshift_hosted_logging_deploy", default=False)
+        logging_deployed = self.get_var("openshift_hosted_logging_deploy", convert=bool, default=False)
         return logging_deployed and super(LoggingCheck, self).is_active() and self.is_first_master()
 
     def is_first_master(self):
--- a/roles/openshift_health_checker/test/openshift_check_test.py
+++ b/roles/openshift_health_checker/test/openshift_check_test.py
@@ -81,6 +81,7 @@ def dummy_check(task_vars):
 @pytest.mark.parametrize("keys,expected", [
     (("foo",), 42),
     (("bar", "baz"), "openshift"),
+    (("bar.baz",), "openshift"),
 ])
 def test_get_var_ok(task_vars, keys, expected):
     assert dummy_check(task_vars).get_var(*keys) == expected
@@ -94,3 +95,24 @@ def test_get_var_error(task_vars, missing_keys):
 def test_get_var_default(task_vars, missing_keys):
     default = object()
     assert dummy_check(task_vars).get_var(*missing_keys, default=default) == default
+
+
+@pytest.mark.parametrize("keys, convert, expected", [
+    (("foo",), str, "42"),
+    (("foo",), float, 42.0),
+    (("bar", "baz"), bool, False),
+])
+def test_get_var_convert(task_vars, keys, convert, expected):
+    assert dummy_check(task_vars).get_var(*keys, convert=convert) == expected
+
+
+@pytest.mark.parametrize("keys, convert", [
+    (("bar", "baz"), int),
+    (("bar.baz"), float),
+    (("foo"), "bogus"),
+    (("foo"), lambda a, b: 1),
+    (("foo"), lambda a: 1 / 0),
+])
+def test_get_var_convert_error(task_vars, keys, convert):
+    with pytest.raises(OpenShiftCheckException):
+        dummy_check(task_vars).get_var(*keys, convert=convert)


roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -97,12 +97,39 @@ def get_var(self, *keys, **kwargs):
...
+            keys = keys[0].split(".")
+
         try:
             value = reduce(operator.getitem, keys, self.task_vars)
         except (KeyError, TypeError):
             if "default" in kwargs:
                 return kwargs["default"]
             raise OpenShiftCheckException("'{}' is undefined".format(".".join(map(str, keys))))
+
+        convert = kwargs.get("type")
```
rhcarvalho: Suggestion:

```python
convert = kwargs.get("type", lambda x: x)
try:
    return convert(value)
except ValueError:
    raise OpenShiftCheckException(...)
```

Wherever we use `.get_var(...)`, we'd then always pass a function to the `type` optional argument, and never a string. Passing a string would be more cumbersome anyway.

roles/openshift_health_checker/test/openshift_check_test.py
```diff
@@ -81,6 +81,7 @@ class TestCheck(OpenShiftCheck):
 @pytest.mark.parametrize("keys,expected", [
     (("foo",), 42),
     (("bar", "baz"), "openshift"),
+    (("bar.baz",), "openshift"),
```
rhcarvalho: IIRC we had this before and then removed the "extra feature" in favor of a single way of doing it.
Are you intending to use both `get_var('openshift.foo.bar')` and `get_var('openshift', 'foo', 'bar')`?

roles/openshift_health_checker/test/openshift_check_test.py
```diff
@@ -94,3 +95,22 @@ def test_get_var_error(task_vars, missing_keys):
...
+    (("bar", "baz"), bool, False),
+])
+def test_get_var_convert(task_vars, keys, convert, expected):
+    assert dummy_check(task_vars).get_var(*keys, type=convert) == expected
+
+
+@pytest.mark.parametrize("keys, convert", [
+    (("foo",), "bogus"),
+    (("bar", "baz"), "int"),
+    (("bar.baz"), "float"),
```
rhcarvalho: Funny that this test uses all strings, and the other uses all functions as the `convert` argument. I think dropping strings altogether makes the design and usage simpler.

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: That seems like a good idea. I was thinking we might want logic more complicated to e.g. extract version strings or convert byte suffixes or something, but I guess passing a function would cover that too.

sosiouxme: although, note the logic under bool... part of the reason for doing this is to not have to repeat that everywhere.

I suppose we could still special-case known conversion functions like bool though.

rhcarvalho: > note the logic under bool...

Hmm, why would we add that YAML-like conversion here?

For `bool`, any non-empty string converts to `True`.

Now this gets me thinking that "covert" is normally from A to B, can we say that A is always of type `str`?

roles/openshift_health_checker/test/openshift_check_test.py
sosiouxme: Why not? It's two extra lines and it matches what we do elsewhere in ansible code. Personally I would like the dotted version better but I don't want to go through the code and change them all.

I'm not as set on the "single obvious way to do it" ethos of python...

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: bool("false") is True. Ansible inventory variables typically come in as strings, except where we've set them with facts already. This has already caused multiple bugs. See the change in logging/logging.py - someone who set `openshift_hosted_deploy_logging=False` in their inventory would see the checks expecting logging to be deployed. It's a really easy mistake to make.

sosiouxme:
@rhcarvalho updated

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -107,30 +107,27 @@ def get_var(self, *keys, **kwargs):
                 return kwargs["default"]
             raise OpenShiftCheckException("'{}' is undefined".format(".".join(map(str, keys))))
 
-        convert = kwargs.get("type")
+        convert = kwargs.get("convert", lambda x: x)
```
rhcarvalho: Perhaps the default is really to have strings... in which case `str` as default would be a good deal.

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -107,30 +107,27 @@ def get_var(self, *keys, **kwargs):
...
-            if convert is None:
-                pass
-            elif convert in ["str", str]:
-                return str(value)
-            elif convert in ["int", int]:
-                return int(value)
-            elif convert in ["float", float]:
-                return float(value)
-            elif convert in ["bool", bool]:
+            if convert is bool:
```
rhcarvalho: Suggestion: instead of changing what `bool` (the function) means behind the scenes, we could have an explicit conversion function that handles all the corner cases we want, that would be similar to how API like Django Models and other Python libraries work.

We'd essentially declare `type=SpecialBool` (surely name `SpecialBool` anything different than that...) for the variables where we want the special handling. That also leaves room for us to use `bool` in the future without breaking any then existing code.

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -107,30 +107,27 @@ def get_var(self, *keys, **kwargs):
...
-                )
+                return convert(value)
+
         except ValueError as error:
             raise OpenShiftCheckException(
-                'Cannot convert variable "{var}" to expected type "{type}":\n'
-                '{error}'.format(var=".".join(keys), type=convert, error=str(error))
+                'Cannot convert inventory variable to expected type:\n'
+                '  "{var}={value}"\n'
+                '{error}'.format(var=".".join(keys), value=value, error=str(error))
```
rhcarvalho: I think the `str` around `error` is unnecessary?

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -107,30 +107,27 @@ def get_var(self, *keys, **kwargs):
...
             )
 
-        return value
+        except Exception as error:
+            raise OpenShiftCheckException(
+                'There is a bug in this check. While trying to convert variable \n'
+                '  "{var}={value}"\n'
+                'the given converter cannot be used or failed unexpectedly:\n'
+                '{error}'.format(var=".".join(keys), value=value, error=error)
+            )
```
rhcarvalho: Hmm, this means that if the function returns something that is not a `ValueError` we want to report a slightly different message?!

roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -15,7 +15,7 @@ class LoggingCheck(OpenShiftCheck):
     logging_namespace = "logging"
 
     def is_active(self):
-        logging_deployed = self.get_var("openshift_hosted_logging_deploy", type=bool, default=False)
+        logging_deployed = self.get_var("openshift_hosted_logging_deploy", convert=bool, default=False)
```
rhcarvalho: `convert=bool` reads well, though `convert=str` is a bit strange since we perhaps expect the input to be a string already?! I liked `type`, though it is the name of a builtin function and may make pylint complain ;-)

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: `ValueError` ought to mean the user passed in a wrong value.

Any other kind of error indicates the conversion function writer did not handle the value correctly or at least gracefully (`KeyError` and similar). The user might have done something wrong too, but if it's not obvious to them then they can blame our code.

Are there other error categories that should be included / called out?

roles/openshift_health_checker/openshift_checks/logging/logging.py
sosiouxme: "type" is nice and intuitive for the most of what we're doing now with values coming out of `task_args`. I don't think pylint complains since it's never used as a name. But I think we may come up with more interesting standardized conversion and validation functions that don't just translate to a type. For instance, byte size converters, or validation that `deployment_type` is one of the known values.

So it made sense to me to give it a more general name that doesn't imply a simple typecasting. I'm not set on `convert` though, suggestions welcome... I could easily see `make`, `transform`, `expect`, `ensure` ... or accepting multiple names (`type`, `convert`, `validate`), to support not having to remember which it is :)

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: Now that I think about it, OpenShiftCheckException (from any custom converters we write) should probably just be passed through.

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: There are cases where the user can pass in JSON - disk availability for instance... but it might be best anyway to be explicit about expected values for those.

roles/openshift_health_checker/openshift_checks/__init__.py
```diff
@@ -107,30 +107,27 @@ def get_var(self, *keys, **kwargs):
...
-                pass
-            elif convert in ["str", str]:
-                return str(value)
-            elif convert in ["int", int]:
-                return int(value)
-            elif convert in ["float", float]:
-                return float(value)
-            elif convert in ["bool", bool]:
+            if convert is bool:
                 return str(value).lower() in ["true", "yes", "1"]
```
sosiouxme: I forgot about `y|n` and `on|off`.

Ansible is sort of a mishmash of ini and yaml, since you have the inventory as well as code, group variable files, etc.

My intuition is less about following a standard and more about not surprising the users. Which is a little bit chicken and egg, since following a standard should lead to consistent user expectations, in theory.

In the end, we should probably not implement our own here, and just re-use whatever the jinja `bool` filter already does. Then it will at least be consistent with what the rest of Ansible does, and "not our problem."

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: I really doubt there is an actual use case for using the built-in `bool` as the converter. In this context it is essentially `non_empty` so calling it `bool` would be highly confusing and lead to bugs.

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: > Now that I think about it, OpenShiftCheckException (from any custom converters we write) should probably just be passed through.

I hope we will not get to that? I thought the conversion functions would be more like pure-Python simple functions, just to operate on basic data types.

The `ValueError` + others separation is good, we'll just need to keep it in mind and let our custom functions raise ValueError on bad input and not something else accidental like you mention `KeyError` and similar. Again, I hope perhaps there will be nothing custom apart from the boolean conversion?

roles/openshift_health_checker/openshift_checks/logging/logging.py
rhcarvalho: > or accepting multiple names (type, convert, validate), to support not having to remember which it is :)

Hahahahha :rofl:

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: If it would be possible to delegate/reuse any part of Ansible, jinja, ruamel.yaml, etc, it would be advantageous for us :+1:

roles/openshift_health_checker/openshift_checks/__init__.py
rhcarvalho: Oh my, even JSON... I forgot about that ;)

The best would be to exterminate the INI format and command line flags in the next Ansible version, and made convert it all to Haskell -- hhaha, or maybe not :P

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: Ansible [implements its own bool filter](https://github.com/ansible/ansible/blob/5f73bdc3bf3821c957cddbf81febfc13b3de9af9/lib/ansible/plugins/filter/core.py#L117-L125). It's not exactly a standard but... we can be consistent.

roles/openshift_health_checker/openshift_checks/logging/logging.py
sosiouxme: Still open to other names... `normalize`... `coerce` ....

roles/openshift_health_checker/openshift_checks/__init__.py
sosiouxme: I could see other converters being useful; code we already have could use a standard byte size converter and `deployment_type` validator. I don't think it's too much scope creep / YAGNI to just reraise `OpenShiftCheckException`.

sosiouxme:
aos-ci-test

sosiouxme:
@rhcarvalho you ok with this? Can I get a LGTM?

roles/openshift_health_checker/openshift_checks/logging/logging.py
```diff
@@ -18,7 +18,7 @@ class LoggingCheck(OpenShiftCheck):
     logging_namespace = "logging"
 
     def is_active(self):
-        logging_deployed = self.get_var("openshift_hosted_logging_deploy", default=False)
+        logging_deployed = self.get_var("openshift_hosted_logging_deploy", convert=bool, default=False)
```
rhcarvalho: Is this fixing an existing bug?

sosiouxme: I don't think I filed a bug. But yeah, it's a minor bug that logging is considered always on if defined...

sosiouxme:
[merge]

sosiouxme:
yum flakes https://github.com/openshift/origin/issues/10162 and https://github.com/openshift/origin/issues/8571

[merge]

sosiouxme:
https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/817/

yum flake https://github.com/openshift/origin/issues/8571
also https://github.com/openshift/origin/issues/11114

@rhcarvalho or @sdodson could you merge this manually please.

PR #4925, opened 2017-07-28
Set the openshift_version from the openshift.common.version in case it is empty

ingvagabund:
It can happen the openshift_version is set to an empty string.
Which results in openshift_pkg_version set to "-".
Thus, failing installation of base and excluder packages.

For more info see https://bugzilla.redhat.com/show_bug.cgi?id=1474871


ingvagabund:
aos-ci-test

sdodson:
[merge]

ingvagabund:
[merge]

Issue #4967, opened 2017-08-01
Can't run openshift-ansible against a 3.7.0 origin tag


smarterclayton:
https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_extended_conformance_gce/5209

is failing (contains the v3.7.0-alpha.0 tag) because we haven't added code to ansible to tolerate 3.7.

Need someone to fix this and other places where we require a recognized version.

smarterclayton:
@sdodson @jupierce @stevekuznetsov blocks merging the 3.7.0 tag into master.

stevekuznetsov:
```
  1. Host:     ci-prtest-5a37c28-5209-ig-m-6pzh
     Play:     Verify Requirements
     Task:     openshift_health_check
     Message:  One or more checks failed
     Details:  check "package_version":
               There is no recommended version of Open vSwitch for the current version of OpenShift: 3.7
```

smarterclayton:
Tag is in https://github.com/openshift/origin/pull/15591

smarterclayton:
I suspect there are others, since we have several case statements for versions

sosiouxme:
I'll whip something up for that specific one, can't vouch for anywhere else.

sosiouxme:
aos-ci-test

smarterclayton:
We're still broken:

```
Wednesday 02 August 2017  21:57:35 +0000 (0:00:00.022)       0:08:00.344 ****** 
=============================================================================== 
provision : Provision GCE resources ----------------------------------- 206.80s
openshift_health_check ------------------------------------------------- 27.09s
master-http-proxy : install haproxy ------------------------------------ 23.37s
openshift_facts : Ensure various deps are installed -------------------- 22.31s
openshift_version : Get available origin version ----------------------- 18.89s
/usr/share/ansible/openshift-ansible/roles/os_firewall : need to pause here, otherwise the iptables service starting can sometimes cause ssh to fail -- 10.04s
Gathering Facts --------------------------------------------------------- 7.84s
etcd_server_certificates : Install etcd --------------------------------- 6.78s
openshift_excluder : Install openshift excluder ------------------------- 5.31s
openshift_excluder : Install docker excluder ---------------------------- 3.92s
etcd : Enable etcd ------------------------------------------------------ 2.94s
docker-storage-setup : start docker ------------------------------------- 2.93s
openshift_docker_facts : Set docker facts ------------------------------- 2.57s
[...]

PR #4969, opened 2017-08-01
package_version check: tolerate release version 3.7

sosiouxme:
Addresses issue https://github.com/openshift/openshift-ansible/issues/4967

For now, any version >= 3.6 is handled as if it were 3.6. We may want to
keep that or fine-tune it later.

Also, the ovs_version check is not updated. This is a post-install
health check (does not block install/upgrade) with an update already in
progress so will be addressed there.

### DIFF
--- a/roles/openshift_health_checker/openshift_checks/package_version.py
+++ b/roles/openshift_health_checker/openshift_checks/package_version.py
@@ -1,4 +1,7 @@
 """Check that available RPM packages match the required versions."""
+
+import re
+
 from openshift_checks import OpenShiftCheck, OpenShiftCheckException
 from openshift_checks.mixins import NotContainerizedMixin
 
@@ -9,23 +12,25 @@ class PackageVersion(NotContainerizedMixin, OpenShiftCheck):
     name = "package_version"
     tags = ["preflight"]
 
+    # NOTE: versions outside those specified are mapped to least/greatest
     openshift_to_ovs_version = {
-        "3.6": ["2.6", "2.7"],
-        "3.5": ["2.6", "2.7"],
-        "3.4": "2.4",
+        (3, 4): "2.4",
+        (3, 5): ["2.6", "2.7"],
+        (3, 6): ["2.6", "2.7"],
     }
 
     openshift_to_docker_version = {
-        "3.1": "1.8",
-        "3.2": "1.10",
-        "3.3": "1.10",
-        "3.4": "1.12",
+        (3, 1): "1.8",
+        (3, 2): "1.10",
+        (3, 3): "1.10",
+        (3, 4): "1.12",
+        (3, 5): "1.12",
+        (3, 6): "1.12",
     }
 
-    # map major release versions across releases
-    # to a common major version
-    openshift_major_release_version = {
-        "1": "3",
+    # map major OpenShift release versions across releases to a common major version
+    map_major_release_version = {
+        1: 3,
     }
 
     def is_active(self):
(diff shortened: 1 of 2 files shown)


sosiouxme:
@sdodson should this also be cherry-picked to release-3.6? I.e. do we need to support installing 3.7 with the 3.6 installer... *shrug* maybe not.

sosiouxme:
aos-ci-test

sdodson:
> @sdodson should this also be cherry-picked to release-3.6? I.e. do we need to support installing 3.7 with the 3.6 installer... shrug maybe not.

no, but we should go the otherway around. My hope is that master branch will always support 3.6 and newer so we should be able to merge master back into other branches. Need to formulate more of a plan for how to do that and make sure other teams contributing to the repo are thinking about it.

sdodson:
Travis errors.

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -78,8 +80,10 @@ def get_required_ovs_version(self):
         Else ensure Open vSwitch version 2.4"""
         openshift_version = self.get_openshift_version()
 
-        if float(openshift_version) < 3.5:
+        if _version_lt(openshift_version, "3.5"):
             return self.openshift_to_ovs_version["3.4"]
+        if _version_lt("3.7", openshift_version):
+            return self.openshift_to_ovs_version["3.7"]
```
rhcarvalho: Suggestion: do not hard code "3.7". If the version is greater than the latest version we know about, we can treat it as the latest known. If the version is lower than the earliest version we know about, treat it as the earlier known (i.e., do not hard code the "3.4" either.

That way, whenever a new version comes in, all we have to do is to add 1 line -- or perhaps none, if the version of the deps stays the same, meaning the checks should likely be working on day one after starting a new release cycle.

roles/openshift_health_checker/test/package_version_test.py
```diff
@@ -4,7 +4,7 @@
 
 
 @pytest.mark.parametrize('openshift_release, extra_words', [
-    ('111.7.0', ["no recommended version of Open vSwitch"]),
+    # ('111.7.0', ["no recommended version of Open vSwitch"]),
```
rhcarvalho: We can simply delete this line since the intended behavior now is that newer versions will automatically default to the last known version.

roles/openshift_health_checker/openshift_checks/package_version.py
rhcarvalho: To implement the suggestion, we could stop using strings for the version keys, use tuple of ints instead, and then:

```python
openshift_to_ovs_version = {
    (3, 7): [...],
    ...
}

...

latest = max(openshift_to_ovs_version)
earliest = min(openshift_to_ovs_version)

if version > latest:
    return self.openshift_to_ovs_version[latest]

if version < (3, 0):
    raise ...

if version < earliest:
    return self.openshift_to_ovs_version[earliest]

return ...

File: roles/openshift_health_checker/openshift_checks/package_version.py [hunk 3]
```diff
@@ -124,3 +128,11 @@ def parse_version(self, version):
             components[0] = self.openshift_major_release_version[components[0]]
 
         return '.'.join(components[:2])
+
+
+def _v_to_arr(version):
+    return map(int, re.findall(r'\d+', version))
```
rhcarvalho: `map()` in Python 3 does not return a list... that's what causes the Travis error in `_version_lt` when we try to compare the two objects returned by `_v_to_arr`.

Suggestion:

```python
return tuple(int(component) for component in re.findall(r'\d+', version))
```

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -1,4 +1,5 @@
 """Check that available RPM packages match the required versions."""
+import re
```
rhcarvalho: nit: add an empty line to separate stdlib imports from local imports.

roles/openshift_health_checker/openshift_checks/package_version.py
sosiouxme: I like it

smarterclayton:
This is blocking a lot of other changes - eta? :)

sosiouxme:
Should have it tested locally and updated in less than an hour, then it's a
question of how much CI testing you want to wait for

sosiouxme:
aos-ci-test

sosiouxme:
I do not understand https://travis-ci.org/openshift/openshift-ansible/jobs/260254114 -- seems to have run against the previous commit??

sdodson:
bot, retest this

sosiouxme:
travis passed, should be good to [merge][severity: blocker]

I ran the integration tests locally, those should be good. I also ran the pre-install playbook to exercise this against actual hosts. Up to you guys if you want to hurry up the process any :) @sdodson @smarterclayton

roles/openshift_health_checker/openshift_checks/package_version.py
```diff
@@ -73,54 +78,49 @@ def run(self):
...
+        openshift_version = self.get_openshift_version_tuple()
+
+        earliest = min(self.openshift_to_docker_version)
+        latest = max(self.openshift_to_docker_version)
+        if openshift_version < earliest:
+            return self.openshift_to_docker_version[earliest]
+        if openshift_version > latest:
+            return self.openshift_to_docker_version[latest]
+
+        docker_version = self.openshift_to_docker_version.get(openshift_version)
```
rhcarvalho: For the future: should we explicitly limit `openshift_version` to 2 components, without assuming what "openshift_image_tag" looks like today? That way we'd avoid coming in with `(3, 6, 1)` and then failing to match a key in the version map (for both Docker and OVS).

sosiouxme: [get_openshift_version_tuple returns exactly two ints in the tuple.](https://github.com/sosiouxme/openshift-ansible/blob/65583d894be627c020732cb66e8c126db334560c/roles/openshift_health_checker/openshift_checks/package_version.py#L121-L126)

BTW in testing the failure case I needed to use a bogus tuple like that since everything else would be found.

There's now a very similar method in the base class, I have a todo to combine these. Maybe should be called `openshift_minor_version_tuple` or something to be clearest.

PR #4971, opened 2017-08-02
Add CFME external database app template

bjpritch:
The Red Hat documentation for Cloudforms install on Openshift references a template for provisioning using a database external to the cluster: [Installing CF on Openshift: Deploy with external DB](https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.5/html-single/installing_red_hat_cloudforms_on_openshift_container_platform/#deploying-the-appliance-external-db).

This template provides similar functionality to the cfme-template already included, but with references to ExternalName service instead of a service existing within the cluster.

Edit: Tested against Openshift 3.5

### DIFF
--- a/roles/openshift_examples/files/examples/v1.5/cfme-templates/cfme-template-ext-db.yaml
+++ b/roles/openshift_examples/files/examples/v1.5/cfme-templates/cfme-template-ext-db.yaml
@@ -1,0 +1,414 @@
+apiVersion: v1
+kind: Template
+labels:
+  template: cloudforms-ext-db
+metadata:
+  name: cloudforms-ext-db
+  annotations:
+    description: "CloudForms appliance with persistent storage using an external DB host"
+    tags: "instant-app,cloudforms,cfme"
+    iconClass: "icon-rails"
+objects:
+- apiVersion: v1
+  kind: Secret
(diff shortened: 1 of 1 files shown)


sdodson:
@bjpritch is there an upstream source for this?
@simon3z  @simaishi Can you take a look at this?

sdodson:
We'll want to ensure that this is also placed in v3.6 and v3.7 directories too

simaishi:
The external DB support will be added in the next CFME release, so this will need to wait. This requires a change in the image as well, which hasn't happened yet. It's tracked in [BZ 1474842](https://bugzilla.redhat.com/show_bug.cgi?id=1474842)

sdodson:
@simaishi But the docs reference the template that Ben is adding, where else would they be getting that template from?

simaishi:
Probably the doc should be updated to exclude it then. I wasn't aware the doc includes instruction for external DB... that's not something we currently support.

simaishi:
The reference to the external DB setup has been removed from the documentation. This feature will be added in the next CFME release and I will add the template at that time, along with other template changes.

sdodson:
we'll ensure that external db is supported in our CFME 4.6 work

PR #4973, opened 2017-08-02
logging: introducing event router

wozniakjan:
Initial concept integrating kubernetes events to EFK stack using [eventrouter](https://github.com/heptio/eventrouter)
- **eventrouter** is deployed to logging project, has a service account and its own role to read events
- **eventrouter** watches kubernetes events, marshalls them to JSON and outputs to its STDOUT
- **fluentd** picks them up and inserts to elastic search logging project index

Kubernetes events could be easily stored into different index. Among other ways, it could be achieved either by a fluentd filter plugin or custom eventrouter sink.

Pending actions:
- [x] - get our downstream and upstream images for eventrouter, don't use heptio's

cc: @josefkarasek , @jcantrill , @richm

### DIFF
--- a/roles/openshift_logging/README.md
+++ b/roles/openshift_logging/README.md
@@ -12,13 +12,13 @@ generation for Elasticsearch (it uses JKS) as well as openssl to sign certificat
 As part of the installation, it is recommended that you add the Fluentd node selector label
 to the list of persisted [node labels](https://docs.openshift.org/latest/install_config/install/advanced_install.html#configuring-node-host-labels).
 
-###Required vars:
+### Required vars:
 
 - `openshift_logging_install_logging`: When `True` the `openshift_logging` role will install Aggregated Logging.
 
 When `openshift_logging_install_logging` is set to `False` the `openshift_logging` role will uninstall Aggregated Logging.
 
-###Optional vars:
+### Optional vars:
 - `openshift_logging_purge_logging`: When `openshift_logging_install_logging` is set to 'False' to trigger uninstalation and `openshift_logging_purge_logging` is set to 'True', it will completely and irreversibly remove all logging persistent data including PVC. Defaults to 'False'.
 - `openshift_logging_image_prefix`: The prefix for the logging images to use. Defaults to 'docker.io/openshift/origin-'.
 - `openshift_logging_curator_image_prefix`: Setting the image prefix for Curator image. Defaults to `openshift_logging_image_prefix`.
@@ -90,6 +90,12 @@ When `openshift_logging_install_logging` is set to `False` the `openshift_loggin
 - `openshift_logging_es_nodeselector`: A map of labels (e.g. {"node":"infra","region":"west"} to select the nodes where the pod will land.
 - `openshift_logging_es_number_of_shards`: The number of primary shards for every new index created in ES. Defaults to '1'.
 - `openshift_logging_es_number_of_replicas`: The number of replica shards per primary shard for every new index. Defaults to '0'.
+
+- `openshift_logging_install_eventrouter`: Coupled with `openshift_logging_install_logging`. When both are 'True', eventrouter will be installed. When both are 'False', eventrouter will be uninstalled.
+Other combinations will keep the eventrouter untouched.
+
+Detailed eventrouter configuration can be found in
+- `roles/openshift_logging_eventrouter/README.md`
 
 When `openshift_logging_use_ops` is `True`, there are some additional vars. These work the
 same as above for their non-ops counterparts, but apply to the OPS cluster instance:
--- a/roles/openshift_logging/tasks/delete_logging.yaml
+++ b/roles/openshift_logging/tasks/delete_logging.yaml
@@ -105,3 +105,9 @@
     - logging-elasticsearch
     - logging-fluentd
     - logging-mux
+
+## EventRouter
+- include_role:
+    name: openshift_logging_eventrouter
+  when:
+    not openshift_logging_install_eventrouter | default(false) | bool
--- a/roles/openshift_logging_eventrouter/README.md
+++ b/roles/openshift_logging_eventrouter/README.md
@@ -1,0 +1,20 @@
+Event router
+------------
+
+A pod forwarding kubernetes events to EFK aggregated logging stack.
+
+- **eventrouter** is deployed to logging project, has a service account and its own role to read events
+- **eventrouter** watches kubernetes events, marshalls them to JSON and outputs to its sink, currently only various formatting to STDOUT
+- **fluentd** picks them up and inserts to elasticsearch *.operations* index
+
+- `openshift_logging_install_eventrouter`: When 'True', eventrouter will be installed. When 'False', eventrouter will be uninstalled.
+
+Configuration variables:
+
+- `openshift_logging_eventrouter_image_prefix`: The prefix for the eventrouter logging image. Defaults to `openshift_logging_image_prefix`.
+- `openshift_logging_eventrouter_image_version`: The image version for the logging eventrouter. Defaults to 'latest'.
+- `openshift_logging_eventrouter_sink`: Select a sink for eventrouter, supported 'stdout' and 'glog'. Defaults to 'stdout'.
+- `openshift_logging_eventrouter_nodeselector`: A map of labels (e.g. {"node":"infra","region":"west"} to select the nodes where the pod will land.
+- `openshift_logging_eventrouter_cpu_limit`: The amount of CPU to allocate to eventrouter. Defaults to '100m'.
+- `openshift_logging_eventrouter_memory_limit`: The memory limit for eventrouter pods. Defaults to '128Mi'.
+- `openshift_logging_eventrouter_namespace`: The namespace where eventrouter is deployed. Defaults to 'default'.
--- a/roles/openshift_logging_eventrouter/defaults/main.yaml
+++ b/roles/openshift_logging_eventrouter/defaults/main.yaml
@@ -1,0 +1,9 @@
+---
+openshift_logging_eventrouter_image_prefix: "{{ openshift_logging_image_prefix | default(__openshift_logging_image_prefix) }}"
+openshift_logging_eventrouter_image_version: "{{ openshift_logging_image_version | default('latest') }}"
+openshift_logging_eventrouter_replicas: 1
+openshift_logging_eventrouter_sink: stdout
+openshift_logging_eventrouter_nodeselector: ""
+openshift_logging_eventrouter_cpu_limit: 100m
+openshift_logging_eventrouter_memory_limit: 128Mi
+openshift_logging_eventrouter_namespace: default
--- a/roles/openshift_logging_eventrouter/files/eventrouter-template.yaml
+++ b/roles/openshift_logging_eventrouter/files/eventrouter-template.yaml
@@ -1,0 +1,103 @@
+# this openshift template should match (except nodeSelector) jinja2 template in
+# ../templates/eventrouter-template.j2
+kind: Template
+apiVersion: v1
+metadata:
+  name: eventrouter-template
+  annotations:
+    description: "A pod forwarding kubernetes events to EFK aggregated logging stack."
+    tags: "events,EFK,logging"
+objects:
+  - kind: ServiceAccount
+    apiVersion: v1
+    metadata:
+      name: aggregated-logging-eventrouter
+  - kind: ClusterRole
+    apiVersion: v1
+    metadata:
+      name: event-reader
+    rules:
+    - apiGroups: [""]
+      resources: ["events"]
+      verbs: ["get", "watch", "list"]
+  - kind: ConfigMap
+    apiVersion: v1
+    metadata:
+      name: logging-eventrouter
+    data:
+      config.json: |- 
+        {
+          "sink": "${SINK}"
+        }
+  - kind: DeploymentConfig
+    apiVersion: v1
+    metadata:
+      name: logging-eventrouter
+      labels:
+        component: eventrouter
+        logging-infra: eventrouter
+        provider: openshift
+    spec:
+      selector:
+        component: eventrouter
+        logging-infra: eventrouter
+        provider: openshift
+      replicas: ${REPLICAS}
+      template:
+        metadata:
+          labels:
+            component: eventrouter
+            logging-infra: eventrouter
+            provider: openshift
+          name: logging-eventrouter
+        spec:
+          serviceAccount: aggregated-logging-eventrouter
+          serviceAccountName: aggregated-logging-eventrouter
+          containers:
+            - name: kube-eventrouter
+              image: ${IMAGE}
+              imagePullPolicy: Always
+              resources:
+                limits:
+                  memory: ${MEMORY} 
+                  cpu: ${CPU}
+                requires:
+                  memory: ${MEMORY}
+              volumeMounts:
+              - name: config-volume
+                mountPath: /etc/eventrouter
+          volumes:
+            - name: config-volume
+              configMap:
+                name: logging-eventrouter
+  - kind: ClusterRoleBinding
+    apiVersion: v1
+    metadata:
+      name: event-reader-binding
+    subjects:
+    - kind: ServiceAccount
+      name: aggregated-logging-eventrouter
+      namespace: ${NAMESPACE}
+    roleRef:
+      kind: ClusterRole
+      name: event-reader
+
+parameters:
+  - name: SINK
+    displayName: Sink
+    value: stdout
+  - name: REPLICAS
+    displayName: Replicas
+    value: "1"
+  - name: IMAGE
+    displayName: Image
+    value: "docker.io/openshift/origin-logging-eventrouter:latest"
+  - name: MEMORY
+    displayName: Memory
+    value: "128Mi"
+  - name: CPU
+    displayName: CPU
+    value: "100m"
+  - name: NAMESPACE
+    displayName: Namespace
+    value: default
--- a/roles/openshift_logging_eventrouter/tasks/install_eventrouter.yaml
+++ b/roles/openshift_logging_eventrouter/tasks/install_eventrouter.yaml
@@ -1,0 +1,59 @@
+---
+# initial checks
+- assert:
+    msg: Invalid sink type "{{openshift_logging_eventrouter_sink}}", only one of "{{__eventrouter_sinks}}" allowed
+    that: openshift_logging_eventrouter_sink in __eventrouter_sinks
+
+# allow passing in a tempdir
+- name: Create temp directory for doing work in
+  command: mktemp -d /tmp/openshift-logging-ansible-XXXXXX
+  register: mktemp
+  changed_when: False
+
+- set_fact:
+    tempdir: "{{ mktemp.stdout }}"
+
+- name: Create templates subdirectory
+  file:
+    state: directory
+    path: "{{ tempdir }}/templates"
+    mode: 0755
+  changed_when: False
+
+# create EventRouter deployment config
+- name: Generate EventRouter template
+  template:
+    src: eventrouter-template.j2
+    dest: "{{ tempdir }}/templates/eventrouter-template.yaml"
+  vars:
+    node_selector: "{{ openshift_logging_eventrouter_nodeselector | default({}) }}"
+
+- name: Create EventRouter template
+  oc_obj:
+    namespace: "{{ openshift_logging_eventrouter_namespace }}"
+    kind: template
+    name: eventrouter-template
+    state: present
+    files:
+    - "{{ tempdir }}/templates/eventrouter-template.yaml"
+
+- name: Process EventRouter template
+  oc_process:
+    state: present
+    template_name: eventrouter-template
+    namespace: "{{ openshift_logging_eventrouter_namespace }}"
+    params:
+      IMAGE: "{{openshift_logging_eventrouter_image_prefix}}logging-eventrouter:{{openshift_logging_eventrouter_image_version}}"
+      REPLICAS: "{{ openshift_logging_eventrouter_replicas }}"
+      CPU: "{{ openshift_logging_eventrouter_cpu_limit }}"
+      MEMORY: "{{ openshift_logging_eventrouter_memory_limit }}"
+      NAMESPACE: "{{ openshift_logging_eventrouter_namespace }}"
+      SINK: "{{ openshift_logging_eventrouter_sink }}"
+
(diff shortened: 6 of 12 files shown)


roles/openshift_logging_eventrouter/files/eventrouter-dc.yaml
```diff
@@ -0,0 +1,37 @@
...
+  labels:
+    component: eventrouter
+    logging-infra: eventrouter
+    provider: openshift
+spec:
+  selector:
+    component: eventrouter
+    logging-infra: eventrouter
+    provider: openshift
+  replicas: 2
```
jcantrill: Do we really want 2?  Won't that result in events being shipped twice?

wozniakjan: https://github.com/heptio/eventrouter/blob/master/yaml/eventrouter-deployment.yaml#L21-L23

I based that number on their template and a comment in it. I didn't find duplicate events in the logs but also, I am not entirely sure how this works exactly

wozniakjan: I take that back, you were right and there are duplicates

roles/openshift_logging_eventrouter/templates/eventrouter-dc.j2
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: DeploymentConfig
 metadata:
   name: logging-eventrouter
-  namespace: logging
+  namespace: "{{ namespace }}"
```
jcantrill: I'm not certain we need to specify this here if oc_obj correctly specifies the namespace upon creation.  The namespace will be set by the action of creating the DC

wozniakjan: thanks! fixed

wozniakjan:
[(8) [SCCFSI] flow Master events to the log stack[logging][epic-ois-agl-exp]](https://trello.com/c/RXMD3VJZ/496-8-sccfsi-flow-master-events-to-the-log-stackloggingepic-ois-agl-exp)

roles/openshift_logging/README.md
```diff
@@ -12,14 +12,14 @@ generation for Elasticsearch (it uses JKS) as well as openssl to sign certificat
...
-###Required vars:
+### Required vars:
 
 - `openshift_logging_install_logging`: When `True` the `openshift_logging` role will install Aggregated Logging.
 - `openshift_logging_upgrade_logging`:  When `True` the `openshift_logging` role will upgrade Aggregated Logging.
 
 When both `openshift_logging_install_logging` and `openshift_logging_upgrade_logging` are `False` the `openshift_logging` role will uninstall Aggregated Logging.
 
-###Optional vars:
+### Optional vars:
```
jcantrill: We should add a var to control installing of event router like `openshift_logging_eventrouter_install_eventrouter` default to false maybe?

roles/openshift_logging/README.md
```diff
@@ -80,6 +80,12 @@ When both `openshift_logging_install_logging` and `openshift_logging_upgrade_log
 - `openshift_logging_es_number_of_shards`: The number of primary shards for every new index created in ES. Defaults to '1'.
 - `openshift_logging_es_number_of_replicas`: The number of replica shards per primary shard for every new index. Defaults to '0'.
 
+- `openshift_logging_eventrouter_sink`: Select a sink for eventrouter, supported 'stdout' and 'glog'. Defaults to 'stdout'.
```
jcantrill: Lets move these entries to the role README

roles/openshift_logging/tasks/delete_logging.yaml
```diff
@@ -95,3 +95,36 @@
     - logging-elasticsearch
     - logging-fluentd
     - logging-mux
+
+# delete eventrouter
```
jcantrill: We should move these to the role and figure out how to include them from the parent logging role as a task

roles/openshift_logging/tasks/delete_logging.yaml
```diff
@@ -95,3 +95,36 @@
...
+- name: Delete EventRouter service account
+  oc_serviceaccount:
+    state: absent
+    name: "aggregated-logging-eventrouter"
+    namespace: "{{ openshift_logging_eventrouter_namespace | default('default') }}"
```
jcantrill: I think we should be able to represent this in the defaults/main.yaml so we dont have to repeat everywhere

roles/openshift_logging_eventrouter/README.md
```diff
@@ -0,0 +1,8 @@
+Event router
+------------
+
+A pod forwarding kubernetes events to EFK aggregated logging stack.
+
+- **eventrouter** is deployed to logging project, has a service account and its own role to read events
+- **eventrouter** watches kubernetes events, marshalls them to JSON and outputs to its sink, currently only various formatting on STDOUT
+- **fluentd** picks them up and inserts to elastic search logging project index
```
jcantrill: ...Elasticsearch index..   Note spelling of Elasticsearch

roles/openshift_logging_eventrouter/defaults/main.yaml
```diff
@@ -0,0 +1,7 @@
+---
+openshift_logging_eventrouter_replicas: 1
+openshift_logging_eventrouter_sink: stdout
+openshift_logging_eventrouter_nodeselector: ""
```
jcantrill: Should this default to empty hash?

roles/openshift_logging_eventrouter/templates/eventrouter-dc.j2
```diff
@@ -0,0 +1,53 @@
...
+{% if node_selector is iterable and node_selector | length > 0 %}
+      nodeSelector:
+{% for key, value in node_selector.iteritems() %}
+        {{ key }}: "{{ value }}"
+{% endfor %}
+{% endif %}
+      containers:
+        - name: kube-eventrouter
+          # TODO: reference rh registry once it gets there
+          image: gcr.io/heptio-images/eventrouter:v0.1
```
jcantrill: This should become variables of image and version so can change them accordingly for upstream and downstream, etc

roles/openshift_logging_eventrouter/templates/eventrouter-dc.j2
```diff
@@ -0,0 +1,53 @@
...
+          imagePullPolicy: IfNotPresent
+{% if (memory_limit is defined and memory_limit is not none and memory_limit != "")
+or (cpu_limit is defined and cpu_limit is not none and cpu_limit != "") %}
+          resources:
+            limits:
+{% if memory_limit is defined and memory_limit is not none and memory_limit != "" %}
+              memory: "{{ memory_limit }}"
```
jcantrill: We should also set requests.memory to be the same as limits.memory similar to logging so the pod does not get killed off during memory starvation on the node

roles/openshift_logging_eventrouter/defaults/main.yaml
```diff
@@ -0,0 +1,7 @@
+---
+openshift_logging_eventrouter_replicas: 1
+openshift_logging_eventrouter_sink: stdout
```
jcantrill: We should add the allowable values as a list to vars/main.yaml and validate the choice of sync.  There is a similar pattern in the elasticsearch role for validating the logger outputs (e.g. file, console)

roles/openshift_logging_eventrouter/tasks/main.yaml
```diff
@@ -0,0 +1,88 @@
...
+  oc_obj:
+    state: present
+    name: logging-eventrouter
+    namespace: "{{ openshift_logging_eventrouter_namespace }}"
+    kind: dc
+    files:
+    - "{{ tempdir }}/templates/eventrouter-dc.yaml"
+    delete_after: false
+
+
```
jcantrill: I dont see it hear but I believe in the logging and metrics role we have a task that deletes the tempdir we created initially.

roles/openshift_logging_eventrouter/tasks/main.yaml
```diff
@@ -0,0 +1,88 @@
+# allow passing in a tempdir
+- name: Create temp directory for doing work in
+  command: mktemp -d /tmp/openshift-logging-ansible-XXXXXX
+  register: mktemp
+  changed_when: False
+
+- set_fact:
+    tempdir: "{{ mktemp.stdout }}"
```
jcantrill: Consider change the name to be something more specific to the role.  I believe if this role is run with other roles it will clobber any other use of the name 'tempdir'

roles/openshift_logging/README.md
wozniakjan: I think that is a very good idea, thanks for the suggestion!

roles/openshift_logging/README.md
wozniakjan: Given neither role has its own README, I think it would be better to keep this PR only related to the introduction of eventrouter and then create separate PR which moves relevant README parts to appropriate roles

roles/openshift_logging/tasks/delete_logging.yaml
wozniakjan: Here I would also feel more comfortable with keeping the changes related to the introduction of eventrouter only and do refactoring in a separate PR. Currently, no role has its own uninstall.

roles/openshift_logging_eventrouter/defaults/main.yaml
wozniakjan: majority of roles also default nodeselector to an empty string, I was trying to be consistent with how other roles are implemented.

- https://github.com/openshift/openshift-ansible/blob/d57f27bf71f68ba91996bc4a0b1e79b3be956b7e/roles/openshift_logging_curator/defaults/main.yml#L11
- https://github.com/openshift/openshift-ansible/blob/d57f27bf71f68ba91996bc4a0b1e79b3be956b7e/roles/openshift_logging_elasticsearch/defaults/main.yml#L8
- https://github.com/openshift/openshift-ansible/blob/d57f27bf71f68ba91996bc4a0b1e79b3be956b7e/roles/openshift_logging_kibana/defaults/main.yml#L10


Mux and fluentd don't, should I follow rather this pattern?
- https://github.com/openshift/openshift-ansible/blob/d57f27bf71f68ba91996bc4a0b1e79b3be956b7e/roles/openshift_logging_fluentd/defaults/main.yml#L10
- https://github.com/openshift/openshift-ansible/blob/d57f27bf71f68ba91996bc4a0b1e79b3be956b7e/roles/openshift_logging_kibana/defaults/main.yml#L10

roles/openshift_logging_eventrouter/tasks/main.yaml
wozniakjan: I am not sure I understand, other roles seem to have implemented the creation of tempdir exactly the same way. Could you please elaborate what am I missing? Thank you

- https://github.com/openshift/openshift-ansible/blob/d57f27bf71f68ba91996bc4a0b1e79b3be956b7e/roles/openshift_logging_mux/tasks/main.yaml#L19
- https://github.com/openshift/openshift-ansible/blob/d57f27bf71f68ba91996bc4a0b1e79b3be956b7e/roles/openshift_logging_curator/tasks/main.yaml#L11
- https://github.com/openshift/openshift-ansible/blob/d57f27bf71f68ba91996bc4a0b1e79b3be956b7e/roles/openshift_logging_elasticsearch/tasks/main.yaml#L27
- https://github.com/openshift/openshift-ansible/blob/d57f27bf71f68ba91996bc4a0b1e79b3be956b7e/roles/openshift_logging_fluentd/tasks/main.yaml#L43
- https://github.com/openshift/openshift-ansible/blob/d57f27bf71f68ba91996bc4a0b1e79b3be956b7e/roles/openshift_logging_kibana/tasks/main.yaml#L13

roles/openshift_logging/tasks/delete_logging.yaml
wozniakjan: We have the "per logging component" role defaults but one global uninstall task for logging. Those defaults are not present in this role. Once we refactor the uninstall task to be also per role, the "per logging component" roles defaults will be automatically picked up.

Otherwise, I am not sure how well will the variable precedence behave if we have same named variable defaulted at two different roles.

- https://github.com/wozniakjan/openshift-ansible/blob/b2ab2c642567be9400c2223c688046e812399f5b/roles/openshift_logging_eventrouter/defaults/main.yaml#L9

jcantrill: This is partly why I suggested to move the delete tasks to the event router role.  Tasks can be called from other roles.  A case can be made that we may wish to devise deployment of something like eventrouter->kafka which would mean there is no need for any of the logging role.  Keeping it in logging would mean to uninstall we would have to rely upon the parent logging role.  I generally see event gathering not necessarily tied explicitly to logging

roles/openshift_logging_eventrouter/tasks/main.yaml
jcantrill: I defer to your finding as it appears we use the same pattern without incident

roles/openshift_logging/tasks/delete_logging.yaml
jcantrill: But we have already created a new role.  It seems its simple enough to add a file to it for deleting the event router and include or call the task here.

roles/openshift_logging/tasks/delete_logging.yaml
jcantrill: Per my previous comment, moving all this to the role should alleviate your concerns

roles/openshift_logging_eventrouter/defaults/main.yaml
```diff
@@ -0,0 +1,9 @@
+---
+openshift_logging_eventrouter_image_prefix: "{{ 'gcr.io/heptio-images/' }}"
```
jcantrill: The braces should not be required here or for the image version

roles/openshift_logging_eventrouter/defaults/main.yaml
```diff
@@ -0,0 +1,9 @@
...
+openshift_logging_eventrouter_image_version: "{{ 'v0.1' }}"
+openshift_logging_eventrouter_replicas: 1
+openshift_logging_eventrouter_sink: stdout
```
jcantrill: Not sure you saw my previous comment but we should validate the sink values similiar to how we validate the logging log appendar (file vs console) which are the only 2 supported options.

wozniakjan: thanks! I did miss it but added it now.

Also may be worth mentioning that currently implemented check for `openshift_logging_es_log_appenders` doesn't appear to be checking values, only size
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/tasks/main.yaml#L110-L113

jcantrill: Hmm... I thought it was changed to be like: https://github.com/openshift/openshift-ansible/pull/5048/files#diff-a3144accf55f93bdfdac09292c2ac475R105

roles/openshift_logging_eventrouter/tasks/install_eventrouter.yaml
```diff
@@ -0,0 +1,101 @@
+# initial checks
+- fail:
```
jcantrill: I was advised by @ewolinetz the preferred patter going forward is to use 'assert' instead of fail

roles/openshift_logging_fluentd/templates/fluent.conf.j2
```diff
@@ -42,6 +42,7 @@
   @include configs.d/openshift/filter-pre-*.conf
   @include configs.d/openshift/filter-retag-journal.conf
   @include configs.d/openshift/filter-k8s-meta.conf
+  @include configs.d/openshift/filter-k8s-events.conf
```
jcantrill: Does this need to be conditionally included for cases where we are not deploying eventrouter?

richm: yes
either do it here, or do it in fluentd/run.sh (assuming there is a new env var added to the logging-fluentd ds and logging-mux dc) - then run.sh could replace with a 0 length file if not using the event router

wozniakjan: Since we agreed that the filter will be part of [ViaQ data model plugin](https://github.com/ViaQ/fluent-plugin-viaq_data_model), I will remove this from `fluent.conf` as soon as I have a PR there

richm:
Why are we using a template?

wozniakjan:
@jcantrill and @pweil- suggested using a template instead so it can be easily `oc new-app` for testing purposes

richm:
nack on using a template, unless there is no other way to do it - we'll just have to figure out some other way to make testing easier

pweil-:
> nack on using a template, unless there is no other way to do it - we'll just have to figure out some other way to make testing easier

I was thinking the template would be simple enough to live in examples/event-router.  If the template actually needs pre-processing (I see some munging around the selectors) then maybe that's not the case.  It would be nice to just be able to new-app this on `oc cluster up --logging` rather than have to run a playbook if there is a way to not have to pre-process.

Other suggestions to make testing easy or reasons not to use a template?  If we always have to use Ansible to pre-process args anyway then there really isn't a need to templatize it.

ewolinetz:
I'm actually for using a template. There may be a push/trend to move everything into a template [when able] for the hosted components and have their ansible role simply pull in and process the template (similar to what we had done before with the deployer pod)

richm:
@ewolinetz ok - but that goes against what I thought was the direction of using python ansible modules (like oc_svc), and makes it harder to be idempotent/incrementally changeable

sdodson:
> @ewolinetz ok - but that goes against what I thought was the direction of using python ansible modules (like oc_svc), and makes it harder to be idempotent/incrementally changeable

Our hope is to push for `oc apply` to become idempotent. So openshift-ansible would would process the template eliminating variability such as generated values and then use `oc apply` which would be responsible for ensuring idempotency.

If we're too far from that being a reality today we can continue to push for TSB to be the pathfinder here since we've got a high level of coordination with @deads2k on that one and once that's taken care of we can come back and refactor this.

ewolinetz:
@richm @sdodson right, I don't think it's quite possible to make the jump to that immediately within logging, we are still working through this with TSB (and will try to with ASB after). However, that is my "argument" for using a template in this case as we can then more easily wrap that up later.

richm:
ok - then I have no objections against using a template here, as long as we understand that we may have to do additional work to make it idempotent/incremental for upgrades

roles/openshift_logging_eventrouter/defaults/main.yaml
```diff
@@ -0,0 +1,9 @@
+---
+openshift_logging_eventrouter_image_prefix: gcr.io/heptio-images/
```
jcantrill: We have an Openshift upstream repo to set this to now correct?

roles/openshift_logging_eventrouter/defaults/main.yaml
```diff
@@ -0,0 +1,9 @@
...
+openshift_logging_eventrouter_image_version: v0.1
+openshift_logging_eventrouter_replicas: 1
+openshift_logging_eventrouter_sink: stdout
+openshift_logging_eventrouter_nodeselector: ""
+openshift_logging_eventrouter_cpu_limit: 100m
+openshift_logging_eventrouter_memory_limit: 128Mi
+openshift_logging_eventrouter_namespace: default
```
jcantrill: Do we have any understanding if this should deploy here or the logging namespace?

roles/openshift_logging_eventrouter/files/eventrouter-template.yaml
```diff
@@ -0,0 +1,109 @@
...
+parameters:
+  - name: SINK
+    displayName: Sink
+    value: stdout
+  - name: REPLICAS
+    displayName: Replicas
+    value: "1"
+  - name: IMAGE
+    displayName: Image
+    value: gcr.io/heptio-images/eventrouter:v0.1
```
jcantrill: Same question as above regarding the default value here?  Openshift upstream repo?

roles/openshift_logging_eventrouter/files/eventrouter-template.yaml
```diff
@@ -0,0 +1,109 @@
...
+        metadata:
+          labels:
+            component: eventrouter
+            logging-infra: eventrouter
+            provider: openshift
+          name: logging-eventrouter
+        spec:
+          serviceAccount: aggregated-logging-eventrouter
+          serviceAccountName: aggregated-logging-eventrouter
+#{% if node_selector is iterable and node_selector | length > 0 %}
```
jcantrill: If this is truly an Openshift template this needs to be removed.

roles/openshift_logging_eventrouter/defaults/main.yaml
wozniakjan: That is what was accepted in [the proposal doc](https://github.com/openshift/origin-aggregated-logging/blob/master/docs/proposals/kube_events_design_doc.md)

roles/openshift_logging_eventrouter/files/eventrouter-template.yaml
wozniakjan: And it truly is an OpenShift template. It starts with '#' so it is a comment, but I will remove the comment as well

roles/openshift_logging_eventrouter/defaults/main.yaml
jcantrill: right.  This was so the events ended up in the operations index.

wozniakjan:
@jcantrill , @ewolinetz do you think we can get this merged?

roles/openshift_logging_eventrouter/defaults/main.yaml
smarterclayton: Please do not deploy anything to default.  @sdodson nothing new should be going into default at this point, no matter what kind of project.

Everything deployed by openshift-ansible that is infrastructure component must go into an `openshift-*` namespace, no exceptions.  This allows us to control evolution of these.

Nothing should default to "default".

jcantrill: @smarterclayton This is the only way, at least initially, these events end up in an 'operations' index in Elastic.  Also, are you suggesting we modify the default for logging to be something like 'openshift-infra-logging' or 'openshift-logging'?  We never agreed on what it should be way back when  logging was first released.

PR #4977, opened 2017-08-02
Fixing SA and clusterrole namespaces

ewolinetz:
Was generating SA and clusterroles in the incorrect namespace.
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1477440

### DIFF
--- a/roles/openshift_metrics/tasks/generate_rolebindings.yaml
+++ b/roles/openshift_metrics/tasks/generate_rolebindings.yaml
@@ -41,7 +41,7 @@
 - name: Set hawkular cluster roles
   oc_obj:
     name: hawkular-metrics
-    namespace: "{{ openshift_metrics_hawkular_agent_namespace }}"
+    namespace: "{{ openshift_metrics_project }}"
     kind: clusterrole
     files:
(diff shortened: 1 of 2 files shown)


ewolinetz:
It seems that the change that introduced this isn't in master yet. Will need to create a separate PR for that (not just a simple cherry-pick)

mwringe:
Metrics needs to handle things in two separate namespaces.

The main metric components need to be deployed normally to the 'openshift-infra' namespace, and the Hawkular Agent to the 'default' namespace.

Are you making sure you are handling both these cases?

ewolinetz:
@mwringe does that mean one of the pods should be in `default` and not `openshift-infra`?  This is just to adjust the namespace where these two service-accounts are created... (and the hawkular cluster-role which should actually be namespace independent)

mwringe:
@ewolinetz the agent's pod, roles, permissions, etc need to be done at the 'default' namespace and not the 'openshift-infra' one. While everything else needs to be done at the 'openshift-infra' namespace.

And yeah, cluster roles don't need namespaces but that might have slipped into here somehow :)

ewolinetz:
I verified, the hawkular agent is still only being deployed in `default` and was able to start up

PR #4980, opened 2017-08-02
Switch to migrating one host and scaling etcd members back up

sdodson:
With large datasets where there are many keys with TTLs the expiry was creating a data inconsistency problem. The hope is that by performing the migration once and then forming a new cluster and then scaling up the additional members this problem will go away.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1475351

### DIFF
--- a/playbooks/common/openshift-etcd/migrate.yml
+++ b/playbooks/common/openshift-etcd/migrate.yml
@@ -17,18 +17,14 @@
   tags:
   - always
 
+# TODO: This will be different for release-3.6 branch
 - name: Prepare masters for etcd data migration
   hosts: oo_masters_to_config
   tasks:
   - set_fact:
       master_services:
-      - "{{ openshift.common.service_type + '-master' }}"
-  - set_fact:
-      master_services:
       - "{{ openshift.common.service_type + '-master-controllers' }}"
       - "{{ openshift.common.service_type + '-master-api' }}"
-    when:
-    - (openshift_master_cluster_method is defined and openshift_master_cluster_method == "native") or openshift.common.is_master_system_container | bool
   - debug:
       msg: "master service name: {{ master_services }}"
   - name: Stop masters
@@ -67,8 +63,21 @@
     when:
     - etcd_backup_failed | length > 0
 
-- name: Migrate etcd data from v2 to v3
+- name: Stop etcd
   hosts: oo_etcd_to_migrate
+  gather_facts: no
+  tags:
+  - always
+  pre_tasks:
+  - set_fact:
+      l_etcd_service: "{{ 'etcd_container' if openshift.common.is_containerized else 'etcd' }}"
+  - name: Disable etcd members
+    service:
+      name: "{{ l_etcd_service }}"
+      state: stopped
+
+- name: Migrate data on first etcd
+  hosts: oo_etcd_to_migrate[0]
   gather_facts: no
   tags:
   - always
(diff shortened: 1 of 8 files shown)


roles/etcd_migrate/tasks/migrate.yml
```diff
@@ -42,23 +124,41 @@
...
+      - keys: "/kubernetes.io/events"
+        ttl: "1h"
+      - keys: "/kubernetes.io/masterleases"
+        ttl: "10s"
+      - keys: "/openshift.io/oauth/accesstokens"
+        ttl: "{{ accessTokenMaxAgeSeconds }}s"
+      - keys: "/openshift.io/oauth/authorizetokens"
+        ttl: "{{ authroizeTokenMaxAgeSeconds }}s"
+      - keys: "/openshift.io/leases/controllers"
+        ttl: "{{ controllerLeaseTTL }}s"
```
sdodson: @liggitt PTAL at the lease migration updates here

sdodson: @smarterclayton too

sdodson:
Ok, I think this is finally cleaned up and working.

sdodson:
aos-ci-test

roles/etcd_migrate/tasks/add_ttls.yml
```diff
@@ -0,0 +1,36 @@
+---
+# To be executed on first master
+- delegate_to: "{{ groups.oo_first_master[0] }}"
```
ingvagabund: Using `delegate_to` is not a good practise. That decision should be made on a play level.

playbooks/common/openshift-etcd/migrate.yml
```diff
@@ -93,17 +132,19 @@
   hosts: oo_masters_to_config
   roles:
   - role: etcd_migrate
+    r_etcd_migrate_action: add_ttls
```
ingvagabund: What about to remove `delegate_to` from form the `add_ttls.yml` and move the action to its own play:
```yaml
- name: Add TTLs on the first master
  hosts: groups.oo_first_master[0]
  roles:
  - role: etcd_migrate
    r_etcd_migrate_action: add_ttls
    etcd_peer: "{{ hostvars[groups.oo_etcd_to_migrate.0].ansible_default_ipv4.address }}"
    etcd_url_scheme: "https"
    etcd_peer_url_scheme: "https"
    when: etcd_migration_failed | length == 0
```

roles/etcd_migrate/tasks/main.yml
```diff
@@ -2,7 +2,7 @@
 - name: Fail if invalid r_etcd_migrate_action provided
   fail:
     msg: "etcd_migrate role can only be called with 'check' or 'migrate' or 'configure'"
-  when: r_etcd_migrate_action not in ['check', 'migrate', 'configure']
+  when: r_etcd_migrate_action not in ['check', 'migrate', 'configure','add_ttls','clean_data']
```
ingvagabund: One more space between `'add_ttls',` and `'clean_data'`

ingvagabund: All the role actions should be documented. Though, not a blocker for the PR.

roles/etcd_migrate/tasks/migrate.yml
```diff
@@ -3,62 +3,44 @@
...
+  systemd:
+    name: "{{ l_etcd_service }}"
+    state: restarted
 
-- name: Wait for cluster to become healthy after migration
+- name: Wait for cluster to become healthy after bringing up first member
   command: >
     etcdctl --cert-file {{ etcd_peer_cert_file }} --key-file {{ etcd_peer_key_file }} --ca-file {{ etcd_peer_ca_file }} --endpoint https://{{ etcd_peer }}:{{ etcd_client_port }} cluster-health
   register: l_etcd_migrate_health
-  until: l_etcd_migrate_health.rc == 0
```
ingvagabund: If the `until` is not needed, you can remove the `register: l_etcd_migrate_health` line as well.

playbooks/common/openshift-etcd/migrate.yml
sdodson: I can do that. I was trying to minimize the number of play transitions but I like what you're suggesting better.

roles/etcd_migrate/tasks/migrate.yml
sdodson: I added the until back in. I had removed it because I wanted it to retry several times regardless but I ended up making the add command loop until success. I was seeing that the cluster would go healthy and then when we attempt to add the member it would have lost leader and then immediately regains leader.

sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
aos-ci-test

PR #4986, opened 2017-08-03
Elasticsearch 5 deployment support

t0ffel:
Elasticsearch 5 support is implemented in `openshift_logging_elasticsearch` role, to deploy pass the variable `elasticsearch5_deploy=true`. It should be used in combination with openshift_logging role. This way hopefully we can transition smoothly to ES5, while keeping ES2.
I'll be submitting a companion pr to origin_aggregated_logging soon.

The role support 2 different node types: master and client-data.
Labelling of nodes is not yet done.

/cc @richm @jcantrill @portante @ewolinetz


jcantrill:
I assume the use of 'elasticsearch5_deploy=true' is temporary as we would expect upon merge users to be upgraded to ES5 and not have the option of both

t0ffel:
@jcantrill I hope `elasticsearch5_deploy` switch is temporary. I hope that it'll allow us to merge the PR earlier than all the ES5 work is done, since the role should work with ES2 as well.

ewolinetz:
instead of the flag `elasticsearch5_deploy` can we instead use something like `openshift_logging_elasticsearch_version`? It can be difficult to phase out variables once introduced and I can see the latter having greater reuse.

We can do verification on that variable and limit it to specific values and even default it

t0ffel:
@ewolinetz ok.
based on today's arch discussion the next steps:
1. I'll change the role to support topology in ES2
2. Add client-only role for ES2
3. <merge>
4. Add support for ES5 in the same role enablable via switch `openshift_logging_elasticsearch_version`
5. <merge>
6. depending on the shape we're in the switch can be hardcoded before 3.7 or ES2 can be completely thrown away.

PR #4988, opened 2017-08-03
Update version checks to tolerate 3.7

stevekuznetsov:

addresses https://github.com/openshift/openshift-ansible/issues/4967

/cc @giuseppe @sdodson @smarterclayton


stevekuznetsov:
[test]

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -932,15 +934,19 @@ def set_version_facts_if_unset(facts):
             version_gte_3_4_or_1_4 = True
             version_gte_3_5_or_1_5 = True
             version_gte_3_6 = True
+            version_gte_3_7 = True
```
mtnbikenc: This should be `False`

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -932,15 +934,19 @@ def set_version_facts_if_unset(facts):
...
         facts['common']['version_gte_3_2_or_1_2'] = version_gte_3_2_or_1_2
         facts['common']['version_gte_3_3_or_1_3'] = version_gte_3_3_or_1_3
         facts['common']['version_gte_3_4_or_1_4'] = version_gte_3_4_or_1_4
         facts['common']['version_gte_3_5_or_1_5'] = version_gte_3_5_or_1_5
         facts['common']['version_gte_3_6'] = version_gte_3_6
+        facts['common']['version_gte_3_7'] = version_gte_3_7
 
-        if version_gte_3_6:
+        if version_gte_3_7:
+            examples_content_version = 'v3.7'
```
mtnbikenc: 3.7 examples content has not been pulled into openshift-ansible, therefore the openshift_examples role will fail.

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py
```diff
@@ -53,7 +53,7 @@ def run(self, terms, variables=None, regions_enabled=True, short_version=None,
             short_version = re.sub('^1.', '3.', short_version)
 
         if short_version == 'latest':
-            short_version = '3.6'
+            short_version = '3.7'
```
mtnbikenc: latest should still be 3.6

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py
```diff
@@ -101,7 +101,7 @@ def run(self, terms, variables=None, regions_enabled=True, short_version=None,
                 {'name': 'MatchInterPodAffinity'}
             ])
 
-        if short_version in ['3.5', '3.6']:
+        if short_version in ['3.5', '3.6', '3.7']:
```
mtnbikenc: Has it been confirmed there are no new predicates for 3.7?

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_priorities.py
```diff
@@ -54,7 +54,7 @@ def run(self, terms, variables=None, zones_enabled=True, short_version=None,
             short_version = re.sub('^1.', '3.', short_version)
 
         if short_version == 'latest':
-            short_version = '3.6'
+            short_version = '3.7'
```
mtnbikenc: latest should still be 3.6

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_priorities.py
```diff
@@ -91,7 +91,7 @@ def run(self, terms, variables=None, zones_enabled=True, short_version=None,
                 {'name': 'InterPodAffinityPriority', 'weight': 1}
             ])
 
-        if short_version in ['3.5', '3.6']:
+        if short_version in ['3.5', '3.6', '3.7']:
```
mtnbikenc: Has it been confirmed there are no new priorities for 3.7?

sdodson:
aos-ci-test

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_priorities.py
sdodson: Yeah, this I'm fine with taking as a followup.

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py
sdodson: Not sure I understand this, the latest images will be 3.7 as soon as we fix the installer.

roles/openshift_facts/library/openshift_facts.py
sdodson: updated

roles/openshift_facts/library/openshift_facts.py
sdodson: Added 3.7 content

sdodson:
aos-ci-test

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_priorities.py
sdodson: @derekwaynecarr or @aveshagarwal are there new scheduler predicates and priorities in 3.7 that we need to reconfigure defaults for?

roles/openshift_facts/library/openshift_facts.py
stevekuznetsov: @sdodson do we have automation for this yet

stevekuznetsov:
@mtnbikenc please feel free to push further commits to this branch to update the PR

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py
mtnbikenc: Ok, if we are flipping openshift-ansible to install 3.7 by default then yes, this should be 3.7.

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_priorities.py
mtnbikenc: Ok, if we are flipping openshift-ansible to install 3.7 by default then yes, this should be 3.7.

roles/openshift_facts/library/openshift_facts.py
mtnbikenc: This should, in fact, be `True` if we are flipping openshift-ansible to install 3.7 by default.  I thought the intent was to allow the installer to install 3.7, not set it as the new default.

roles/openshift_examples/files/examples/v3.6/v3.7
```diff
@@ -0,0 +1 @@
+v3.7
```
mtnbikenc: This looks like an unnecessary artifact.

sdodson:
aos-ci-test

roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_priorities.py
aveshagarwal: @sdodson The only new default predicate I see in 3.7, that was not in 3.6 is  NoVolumeNodeConflict. But no change is required in default configuration.

aveshagarwal: @sdodson could you point me where are defaults being set in openshift-ansible, so that i can double check?

sdodson: @aveshagarwal https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py#L104-L115 and https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_master_facts/lookup_plugins/openshift_master_facts_default_predicates.py#L104-L115

aveshagarwal: @sdodson so that means, in 3.7 list of predicates, NoVolumeNodeConflict will be added too.

sdodson:
This is difficult to test without an actual 3.7 binary so go ahead and merge it and hope for the best. The containerized install test is a CI flake.

PR #4993, opened 2017-08-03
fix openshift/origin/issues/15410.  Add oc into origin-ansible image

jcantrill:
This PR add oc back to the image for 'oc cluster up' to fix openshift/origin/issues/15410

### DIFF
--- a/images/installer/Dockerfile
+++ b/images/installer/Dockerfile
@@ -4,8 +4,10 @@ MAINTAINER OpenShift Team <dev@lists.openshift.redhat.com>
 
 USER root
 
+ADD images/installer/origin.repo /etc/yum.repos.d/
+
 # install ansible and deps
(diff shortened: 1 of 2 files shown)


jcantrill:
aos-ci-test

images/installer/Dockerfile
```diff
@@ -4,8 +4,10 @@ MAINTAINER OpenShift Team <dev@lists.openshift.redhat.com>
 
 USER root
 
+ADD images/installer/origin.repo /etc/yum.repos.d/
```
sosiouxme: @jcantrill Actually could you just add the repo definition under images/installer/root/etc/yum.repos.d and move this section to the top:

```
# Add image scripts and files for running as a system container
COPY images/installer/root /
```

sosiouxme: no wait, scratch that. don't want it in OCP.

sosiouxme:
I would prefer to merge https://github.com/openshift/openshift-ansible/pull/5003 instead. I feel better about providing a smaller, signed, released client RPM.

PR #4998, opened 2017-08-04
WIP: generate table of role variables

ingvagabund:
Transform:

```yaml
#: required=true, description="Deployment type", choices=['origin', 'atomic-openshift']
r_openshift_node_deployment_type: "{{ omit }}"
#: description="Set to deploy containerized node"
r_openshift_node_is_containerized: false
#: description="Set to deploy a node over AH"
r_openshift_node_is_atomic: false
```

into (`./hack/generators/generate.py --defaults roles/openshift_node/defaults/main.yml  --table`)

| name                                             | description                                                     | default | required |
|--------------------------------------------------|-----------------------------------------------------------------|---------|----------|
| r_openshift_node_deployment_type                 | Deployment type                                                 |         | true     |
| r_openshift_node_is_containerized                | Set to deploy containerized node                                | false   |          |
| r_openshift_node_is_atomic                       | Set to deploy a node over AH                                    | false   |          |

and generate checks (`./hack/generators/generate.py --defaults roles/openshift_node/defaults/main.yml  --checks`):
```yml
---
- name: "Fail if r_openshift_node_deployment_type is not defined"
  fail:
    msg: "r_openshift_node_deployment_type must be specified for this role"
  when:
  - r_openshift_node_deployment_type is not defined

- name: Fail if invalid r_openshift_node_deployment_type provided
  fail:
    msg: "r_openshift_node_deployment_type can only be set to a single value from ['origin', 'atomic-openshift']"
  when:
  - r_openshift_node_deployment_type is defined
  - r_openshift_node_deployment_type not in ['origin', 'atomic-openshift']
```

The PR is open to debate. This is still PoC.

### Supported annotations
* `name`: optional, used to annotation role variables that needs to be defined before the role gets invoked
* `description`: string describing purpose of a role variable
* `choices`: optional, list of accepted values of a role variable
* `required`: set to `true` if the role variable is required, defaults to `false`
* `required_when`: optional, condition under which a role variable is required (see example below), can be used multiple times

The `required_when` annotation example:
```yaml
# name=r_openshift_node_system_images_registry, \
# description="System image registry to pull images from" \
# required_when=["r_openshift_node_is_containerized | bool", "r_openshift_node_is_node_system_container | bool"]
```

All annotations are on the same line, the `\` symbols is just used to make the example more readable.
The annotation says that the `r_openshift_node_system_images_registry` is optional but needs to set when both `r_openshift_node_is_containerized` and `r_openshift_node_is_node_system_container` are set to `true`.

### Automatic generators
The role variable table, role variable pre checks and additional tasks can be generated automatically via `make generate`. The script finds all `generate.json` files and based on its content carries a list of actions. E.g. Give a file at `roles/openshift_node/generate.json`
```json
{
  "actions": [{
    "action": "role_var_checks",
    "target": "tasks/pre_checks.yml",
    "source": "defaults/main.yml"
  },{
    "action": "role_var_table",
    "source": "defaults/main.yml",
    "target": "README.md"
    }]
}
```

1. The role variable table in the `README.md` gets refreshed. The table in the `README.md` is expected to be accepted by the `Role Variables[^|]*(\|[^\n]*\n)*` regular expression. E.g. in a form (with an additional NL after the table):
```vim
Role Variables
--------------
<NL>
| name                                             | description                         | default | required |
|--------------------------------------------------|-------------------------------------|---------|----------|
| r_openshift_node_deployment_type                 | Deployment type                     |         | true     |
| r_openshift_node_is_containerized                | Set to deploy containerized node    | false   |          |
| r_openshift_node_is_atomic                       | Set to deploy a node over AH        | false   |          |
<NL>
```
2. The role variable checks gets generated into `tasks/pre_checks.yml` file. The checks file can be then included in the `tasks/main.yml` file.

**Role entry-points**
Some roles variables can be only set when certain ansible variables are defined. E.g. the `r_openshift_node_system_images_registry` can be set to `openshift.common.system_images_registry` only when `openshift.common.system_images_registry` is defined. Given the `openshift_node` will usually gets invoked as:

```yaml
- include_role:
    name: openshift_node
  vars:
    openshift_ca_host: "{{ groups.oo_first_master.0 }}"
    r_openshift_node_deployment_type: "{{ deployment_type  }}"
    r_openshift_node_service_type: "{{ openshift.common.service_type }}"
    ...
```

The `Ansible` does not currently allows us to write something like:

```yaml
vars-if-def:
    r_openshift_node_system_images_registry: "{{ openshift.common.system_images_registry }}"
    r_openshift_node_no_proxy: "{{ openshift.common.no_proxy }}"
    r_openshift_node_sdn_network_plugin_name: "{{ openshift.common.sdn_network_plugin_name }}"
```

so each `r_openshift_node_` variable needs to be set before the `include_role` statement as:

```yaml
- set_fact:
    r_openshift_node_system_images_registry: "{{ openshift.common.system_images_registry }}"
  when: openshift.common.system_images_registry is defined

- set_fact:
    r_openshift_node_no_proxy: "{{ openshift.common.no_proxy }}"
  when: openshift.common.no_proxy is defined

- set_fact:
    r_openshift_node_sdn_network_plugin_name: "{{ openshift.common.sdn_network_plugin_name }}"
  when: openshift.common.sdn_network_plugin_name is defined
```

The sequence of the tasks is pretty generic and can be generated. For that reason, if the following content of the `generate.json` file is present in a directory:

```json
{
  "actions": [{
    "action": "set_fact",
    "target": "set_optional_node_role_variables.yml",
    "mapping": {
      "r_openshift_node_dns_ip": "openshift.node.dns_ip",
      "r_openshift_node_sdn_network_plugin_name": "openshift.common.sdn_network_plugin_name",
      "r_openshift_node_ip": "openshift.common.ip",
      "r_openshift_node_min_tls_version": "openshift_node_min_tls_version",
      "r_openshift_node_cipher_suites": "openshift_node_cipher_suites",
      "r_openshift_node_disable_swap": "openshift_disable_swap",
      "r_openshift_node_cloudprovider_aws_access_key": "openshift_cloudprovider_aws_access_key",
      "r_openshift_node_cloudprovider_aws_secret_key": "openshift_cloudprovider_aws_secret_key",
      "r_openshift_node_cloudprovider_kind": "openshift_cloudprovider_kind",
      "r_openshift_node_kubelet_args": "openshift_node_kubelet_args",
      "r_openshift_node_no_proxy": "openshift.common.no_proxy",
      "r_openshift_node_debug_level": "openshift.node.debug_level",
      "r_openshift_node_https_proxy": "openshift.common.https_proxy",
      "r_openshift_node_http_proxy": "openshift.common.http_proxy",
      "r_openshift_node_system_images_registry": "openshift.common.system_images_registry"
    }
  }]
}
```

the sequences of `set_fact-when` tasks gets generated into `set_optional_node_role_variables.yml` file.

**TODO**

* [x] generate `check_role_vars.yml` based on the `defaults/main.yml` so we don't have to update the checks manually each time the `main.yml` files changes.
* [x] update the role variables table in `README.md` automatically without editing (e.g. run `sed`)
* [x] make a list of all supported annotations, currently: `required`, `description`, `choices`


hack/gen_rvar_table.py
```diff
@@ -0,0 +1,119 @@
+#!/bin/python
+
+import re
+
+with open('defaults/main.yml') as f:
```
ingvagabund: Assumed to be run inside a specific role root directory. E.g.
```sh
cd roles/openshift_node
./gen_var_table.py
```

hack/gen_rvar_table.py
```diff
@@ -0,0 +1,119 @@
...
+    lines = f.read().split("\n")
+
+def print_rvar_table(rvar_table):
+    keys = ["name", "description", "default", "required"]
```
ashcrow: nit: Since this isn't updated it should probably be a ``tuple``.

```python
keys = ("name", "description", "default", "required")
```

hack/gen_rvar_table.py
```diff
@@ -0,0 +1,119 @@
...
+
+    sizes = {}
+    for key in keys:
+        sizes[key] = len(key)
+
+    for item in rvar_table:
+        for key in item.keys():
```
ashcrow: nit: ``list(item.keys())`` to ensure the same result on Python 2 and 3.

hack/gen_rvar_table.py
```diff
@@ -0,0 +1,119 @@
...
+            sizes[key] = max(sizes[key], len(item[key]))
+
+    headers = []
+    delimiters = []
+    for key in keys:
+        headers.append(" {}{} ".format(key, " "*(sizes[key]-len(key))))
+        delimiters.append("-{}-".format("-"*(sizes[key])))
+    print "|{}|".format("|".join(headers))
```
ashcrow: ``print()``

hack/gen_rvar_table.py
```diff
@@ -0,0 +1,119 @@
...
+    print "|{}|".format("|".join(delimiters))
```
ashcrow: ``print()``

hack/gen_rvar_table.py
```diff
@@ -0,0 +1,119 @@
...
+    ki_items = {}
+    for item in rvar_table:
+        ki_items[item["name"]] = item
+
+    for key in sorted(ki_items):
+        item = ki_items[key]
+        data = []
+        for key in keys:
+            data.append(" {}{} ".format(item[key], " "*(sizes[key]-len(item[key]))))
+        print "|{}|".format("|".join(data))
```
ashcrow: ``print()``

hack/gen_rvar_table.py
```diff
@@ -0,0 +1,119 @@
...
+    # sort items by variable name
+    ki_items = {}
+    for item in rvar_table:
+        ki_items[item["name"]] = item
+
+    for key in sorted(ki_items):
+        item = ki_items[key]
+        data = []
+        for key in keys:
+            data.append(" {}{} ".format(item[key], " "*(sizes[key]-len(item[key]))))
```
ashcrow: Not required but you can also use the ``ljust()`` or ``rjust()`` string methods for string justification.

hack/gen_rvar_table.py
```diff
@@ -0,0 +1,119 @@
...
+        print "|{}|".format("|".join(data))
+
+def get_delims(line, skip_list=False):
+    # replace all "...", '...' and [...] with # so it is easy to determine
+    # position of the ',' symbol
+    line = re.sub(r'("[^"]*")', lambda l: '#'*len(l.group()) , line)
+    line = re.sub(r"('[^']*')", lambda l: '#'*len(l.group()) , line)
```
ashcrow: Does ``re.sub(r"('[^'\"]*')"`` not work for replacing? Not a big deal but if it does it would be faster.

hack/gen_rvar_table.py
tbielawa: `from __future__ import print_function` at the top would be required for 2/3 compat, correct?

hack/gen_rvar_table.py
```diff
@@ -0,0 +1,119 @@
...
+    if not skip_list:
+        line = re.sub(r"(\[[^\]]*\])", lambda l: '#'*len(l.group()) , line)
+    return re.findall(r"[^,]+", line)
+
+def parse_list(value):
```
tbielawa: I'm wondering if slipping in JSON (yes, JSON embedded in YAML, I know...) would reduce the parsing burden and difficulty. It might also have a side effect of being easier to extend in the future.

```
# {"param": "r_openshift_node_deployment_type", "required": True, "description": "Deployment type", "choices": ["origin", "atomic-openshift"]}
r_openshift_node_deployment_type: "{{ omit }}"
```

tbielawa: Additionally, it makes ordering less important. The doc strings just need to be in the parsed file somewhere. Or even in their own file. (I can see why that might not be ideal, however)

ingvagabund: That already crossed my mind. I rejected that cause I did not like the syntax. You have to specify additional quotes and other stuff. But after we discuss advantages and disadvantages of this approach and conclude it is still way to go, let's do it.

hack/gen_rvar_table.py
ingvagabund: I don't understand. Can you elaborate?

hack/generators/generate.py
```diff
@@ -0,0 +1,320 @@
...
+import re
+import jinja2
+import optparse
+import logging
+import json
+
+
+def setOptions():
+
+    parser = optparse.OptionParser("%prog ")
```
tbielawa: I'd like to see what @ashcrow has to say about this as well, but I know two things:

1. The `optparse` module [is deprecated since 2.7](https://docs.python.org/2/library/optparse.html) in favor of `argparse`
1. We don't use `optparse` anywhere else in our code base, we are using `argparse` (in 7 different locations)

Perhaps switch over to [`argparse`](https://docs.python.org/2/library/argparse.html#module-argparse) instead?

ashcrow: ``argparse`` is the way to go for STDLIB command line arguments.

ingvagabund: Yeah, serving in the next PR I hope unless some crazy idea comes first and I will "just" push another commit.

hack/generators/generate.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+export PATH=$PATH:$(pwd)/hack/generators
+pwd
```
tbielawa: Is this `pwd` just for debugging purposes?

hack/generators/generate.sh
```diff
@@ -0,0 +1,8 @@
...
+for file in $(find . -name generate.json); do
```
tbielawa: `file` is a standard system command. I'm glad this isn't broken but maybe consider using a different name instead.

tbielawa:
In the generated table can it also show the available choices when `choices=[....]` is set? That would be super helpful for enumerated variable types.

ashcrow:
Added four commits to help out at https://github.com/ashcrow/openshift-ansible/commits/generate-list-of-role-variables. @ingvagabund PTAL and cherry-pick if they look good.

ewolinetz:
Will this even parse variables that we may have defined within 'defaults/main' but have commented out because we are using logic like `not is defined` within the role that would use it?

What about patterns where we use a different `var` file based on another parameter, is that something we would be able to encompass with this as well or does that become N times more difficult?

ashcrow:
> Will this even parse variables that we may have defined within 'defaults/main' but have commented out because we are using logic like not is defined within the role that would use it?

No, it would not. If that is needed I'm sure it could be added though. One idea would be to format it like so:

```
#: required=True,description="blah blah"
#i_don't_start_with_a_space="something"
```

But the current implementation would not notice a commented out variable if I read it right.

> What about patterns where we use a different var file based on another parameter, is that something we would be able to encompass with this as well or does that become N times more difficult?

I think that becomes much more difficult. It may be doable by trying to follow ``include``s and ``role``s and then parsing those variables as well, but I'd think that would be something we do after this POC (if needed).

ingvagabund:
> Will this even parse variables that we may have defined within 'defaults/main' but have commented out because we are using logic like not is defined within the role that would use it?

Atm, you can use
```yml
#: name=r_openshift_node_deployment_type, required=true, description="Deployment type"
```
It will get translated into the following check
```yml
- name: "Fail if r_openshift_node_deployment_type is not defined"
  fail:
    msg: "r_openshift_node_deployment_type must be specified for this role"
  when:
  - r_openshift_node_deployment_type is not defined
```

> What about patterns where we use a different var file based on another parameter, is that something we would be able to encompass with this as well or does that become N times more difficult?

Can you provide some example(s)?

ingvagabund:
> In the generated table can it also show the available choices when choices=[....] is set? That would be super helpful for enumerated variable types.

The length of the generated table gets beyond expectations with that. I am wondering if that is something we should put bellow the table. Unless this:

| name                                             | description                                                     | default      | required | choices                     |
|--------------------------------------------------|-----------------------------------------------------------------|--------------|----------|-----------------------------|
| r_openshift_node_cipher_suites                   |                                                                 |              |          |                             |
| r_openshift_node_cloudprovider_aws_access_key    | AWS access key                                                  | ''           |          |                             |
| r_openshift_node_cloudprovider_aws_secret_key    | AWS secret key                                                  | ''           |          |                             |
| r_openshift_node_cloudprovider_kind              | Cloud provider kind, e.g. aws                                   | ''           |          |                             |
| r_openshift_node_config_base                     | Node configuration base directory, e.g. /etc/origin             |              | true     |                             |
| r_openshift_node_data_dir                        |                                                                 |              | true     |                             |
| r_openshift_node_debug_level                     | Set the default node debug level                                | 2            |          |                             |
| r_openshift_node_deployment_type                 | Deployment type                                                 |              | true     | origin,openshift-enterprise |
| r_openshift_node_disable_swap                    | Set to disable a swap to preserve quality of service guarantees | true         |          |                             |
| r_openshift_node_dns_domain                      |                                                                 |              | true     |                             |
| r_openshift_node_dns_ip                          |                                                                 |              |          |                             |
| r_openshift_node_docker_gte_1_10                 |                                                                 | false        |          |                             |
| r_openshift_node_docker_service_name             | Default docker service name, e.g. docker, container-engine      | docker       |          |                             |
| r_openshift_node_env_vars                        | Set node environment variables                                  | {}           |          |                             |
| r_openshift_node_hostname                        |                                                                 |              | true     |                             |
| r_openshift_node_http_proxy                      | HTTP proxy URL address                                          | ''           |          |                             |
| r_openshift_node_https_proxy                     | HTTPS proxy URL address                                         | ''           |          |                             |
| r_openshift_node_image                           | Node image name to pull for containerized deployment            |              |          |                             |
| r_openshift_node_image_tag                       | Node image tag to pull for containerized deployment             |              | true     |                             |
| r_openshift_node_ip                              |                                                                 |              |          |                             |
| r_openshift_node_iptables_sync_period            |                                                                 |              | true     |                             |
| r_openshift_node_is_atomic                       | Set to deploy a node over AH                                    | false        |          |                             |
| r_openshift_node_is_containerized                | Set to deploy containerized node                                | false        |          |                             |
| r_openshift_node_is_node_system_container        | Set to deploy node system container                             | false        |          |                             |
| r_openshift_node_is_openvswitch_system_container | Set to deploy OVS system container                              | false        |          |                             |
| r_openshift_node_kubelet_args                    | Optionall Kubelet arguments                                     | "{{ None }}" |          |                             |
| r_openshift_node_local_quota_per_fsgroup         |                                                                 |              | true     |                             |
| r_openshift_node_master_api_url                  | Master API URL to query                                         |              | true     |                             |
| r_openshift_node_master_sdn_cluster_network_cidr | SDN Cluster network CIDR set by the master component            |              |          |                             |
| r_openshift_node_min_tls_version                 |                                                                 |              |          |                             |
| r_openshift_node_no_proxy                        | No Proxy                                                        | []           |          |                             |
| r_openshift_node_nodename                        |                                                                 |              | true     |                             |
| r_openshift_node_ovs_image                       | OVS image to pull for containerized deployment                  |              |          |                             |
| r_openshift_node_ovs_system_image                | OVS system image to pull for containerized deployment           |              |          |                             |
| r_openshift_node_pkg_version                     | Node package version to install for package based deployment    | ''           |          |                             |
| r_openshift_node_port_range                      |                                                                 |              |          |                             |
| r_openshift_node_portal_net                      | Portal net                                                      |              |          |                             |
| r_openshift_node_proxy_mode                      |                                                                 |              | true     |                             |
| r_openshift_node_registry_url                    |                                                                 |              | true     |                             |
| r_openshift_node_sdn_mtu                         |                                                                 |              | true     |                             |
| r_openshift_node_sdn_network_plugin_name         |                                                                 |              |          |                             |
| r_openshift_node_service_type                    | System service type                                             |              | true     |                             |
| r_openshift_node_set_node_ip                     |                                                                 | true         |          |                             |
| r_openshift_node_storage_plugin_deps             | List if additional storage plugins to deploy                    | []           |          |                             |
| r_openshift_node_system_image                    | Node system image to pull for system containers deployment      |              |          |                             |
| r_openshift_node_system_images_registry          | System image registry to pull images from                       |              |          |                             |
| r_openshift_node_use_calico                      |                                                                 | false        |          |                             |
| r_openshift_node_use_contiv                      |                                                                 | false        |          |                             |
| r_openshift_node_use_dnsmasq                     |                                                                 | false        |          |                             |
| r_openshift_node_use_nuage                       |                                                                 | false        |          |                             |
| r_openshift_node_use_openshift_sdn               | Set to use OpenShift SDN                                        | true         |          |                             |
| r_openshift_node_version_gte_3_3_or_1_3          |                                                                 | true         |          |                             |
| r_openshift_node_version_gte_3_6                 |                                                                 | true         |          |                             |

is still fine.

hack/generators/generate.sh
ingvagabund: :D what is that doing there

ingvagabund: bang, it's gone

tbielawa:
I love the way that table looks @ingvagabund with the `choices` column displayed.

hack/generators/generate.py
```diff
@@ -0,0 +1,348 @@
...
+
+    options = set_options().parse_args()
+    check_options(options)
+
+    if options.json != "":
+
+        with open(options.json) as f:
+            conf = json.load(f)
+
+        # TODO(jchaloup): verify the json file against JSON Schema
```
tbielawa: Just a lil' comment I want to add here: I :heart: JSON Schema.

ingvagabund:
> So we may not define the variables within defaults but we would allow the value to be overwritten if provided in the inventory. Would this follow a similar example for documenting as a variable that we would have commented out (we may just have two defaults since its conditional on the deployment type)?
>
> e.g. #: name=openshift_service_catalog_image_prefix, required=false, description="Service catalog image prefix"

It would as the generator is responsible only for:
* role variable table
* role variable pre-checks

If a role variable is not required, there is no check for it (unless there is the `choice` annotation in which case the variable is checked only when it is defined). And based on the last discussion we agreed on the following pattern:

**When calling a role**
```yaml
- include_role:
    name: <role>
  vars:
    r_openshift_service_catalog_image_prefix: "{{ openshift_service_catalog_image_prefix | default(omit) }}"
    ...
```

**Within a role (inside tasks/pre-checks.yml)**
```yaml
- name: "Set r_openshift_service_catalog_image_prefix to default if omited"
  set_fact:
    r_openshift_service_catalog_image_prefix: "{{ __openshift_service_catalog_image_prefix }}"
  when:
  - __omit_str in r_openshift_service_catalog_image_prefix | string
...
```
**Within a role (inside tasks/main.yml)**:
```yaml
- name: Set default image variables based on deployment_type
  include_vars: "{{ item }}"
  with_first_found:
    - "{{ openshift_deployment_type | default(deployment_type) }}.yml"
    - "default_images.yml"

- include: pre-checks.yml
...
```

So as long as you include a var file that provides the `__` prefix variable, you will get the same effect. Just, one needs to run the generated `pre-check.yml` file after you run the `include_vars`.

[1] https://github.com/ingvagabund/openshift-ansible/blob/2c0bc0dd48208146c846bad90aae55217803c45c/docs/proposals/role-entry-points.md#where-the-role-variables-live-with-annotations

ashcrow:
Let's turn this around a bit since it's been open for a while. Is there anyone who disagrees with this process (keeping in mind that it can change/be improved over time as needed)?

michaelgugino:
I find duplicating all the role defaults into the README.md is not helpful.  I would prefer we spend time documenting the role's higher-level steps.  Users should refer to the defaults/main.yml.

PR #5001, opened 2017-08-04
Moved Elasticsearch 2.x deployment entirely into openshift_logging_elasticsearch role

t0ffel:
Elasticsearch 2.x topology is supported, 2 types of nodes are possible: master, clientdata.
Be default `logging-es` and `logging-es-ops` cluster deploy 1 master and 1 clientdata nodes
The role is launched from `openshift_logging` role.

cc @jcantrill @richm @portante @ewolinetz

### DIFF
--- a/roles/openshift_logging_elasticsearch/defaults/main.yml
+++ b/roles/openshift_logging_elasticsearch/defaults/main.yml
@@ -1,19 +1,43 @@
 ---
+
 ### Common settings
 openshift_logging_elasticsearch_image_prefix: "{{ openshift_logging_image_prefix | default(__openshift_logging_image_prefix) }}"
 openshift_logging_elasticsearch_image_version: "{{ openshift_logging_image_version | default('latest') }}"
 openshift_logging_elasticsearch_image_pull_secret: "{{ openshift_hosted_logging_image_pull_secret | default('') }}"
 openshift_logging_elasticsearch_namespace: logging
-
+openshift_logging_elasticsearch_sa_name: aggregated-logging-elasticsearch
+kubeconfig: /etc/origin/master/openshift-master.kubeconfig
 openshift_logging_elasticsearch_nodeselector: "{{ openshift_logging_es_nodeselector | default('') }}"
 openshift_logging_elasticsearch_cpu_limit: "{{ openshift_logging_es_cpu_limit | default('1000m') }}"
 openshift_logging_elasticsearch_memory_limit: "{{ openshift_logging_es_memory_limit | default('1Gi') }}"
 openshift_logging_elasticsearch_recover_after_time: "{{ openshift_logging_es_recover_after_time | default('5m') }}"
+openshift_logging_elasticsearch_clusterreader: true
 
 openshift_logging_elasticsearch_replica_count: 1
 
 # ES deployment type
 openshift_logging_elasticsearch_deployment_type: "data-master"
+openshift_logging_elasticsearch_node_topology:
+  masters:
+    replicas: 1
+    limits:
+      cpu: 200m
+    requests:
+      memory: 512Mi
+      cpu: 100m
+  clientdata:
+  -
+    node_storage_type: emptydir
+    hostmount_path:
+    - /mnt/data1
+    - /mnt/data2
+    limits:
+      memory: 512Mi
+      cpu: 200m
+    requests:
+      cpu: 100m
+      memory: 256Mi
+
 
 # ES deployment name
 openshift_logging_elasticsearch_deployment_name: ""
@@ -29,7 +53,7 @@ openshift_logging_elasticsearch_hostmount_path: ""
 openshift_logging_elasticsearch_pvc_name: ""
 
 # required if the PVC does not already exist
-openshift_logging_elasticsearch_pvc_size: ""
+openshift_logging_elasticsearch_pvc_size: "10Gi"
 openshift_logging_elasticsearch_pvc_dynamic: false
 openshift_logging_elasticsearch_pvc_pv_selector: {}
 openshift_logging_elasticsearch_pvc_access_modes: ['ReadWriteOnce']
--- a/roles/openshift_logging_elasticsearch/defaults/main.yml
+++ b/roles/openshift_logging_elasticsearch/defaults/main.yml
@@ -17,28 +17,35 @@ openshift_logging_elasticsearch_replica_count: 1
 # ES deployment type
 openshift_logging_elasticsearch_deployment_type: "data-master"
 openshift_logging_elasticsearch_node_topology:
-  masters:
+  clientdatamaster:
     replicas: 1
     limits:
-      cpu: 1000m
-    requests:
-      memory: 1Gi
-  clientdata:
-  -
-    node_storage_type: emptydir
-    hostmount_path:
-    - /mnt/data1
-    - /mnt/data2
-    limits:
       memory: 8Gi
-      cpu: 4000m
     requests:
       memory: 8Gi
-    pv_selector:
-      label: value
-    pvc_size: 10Gi
-    pvc_name: pvc1
-    storage_group: 65534
+
+#  masters:
+#    replicas: 1
+#    limits:
+#      cpu: 1000m
+#    requests:
+#      memory: 1Gi
+#  clientdata:
+#  -
+#    node_storage_type: emptydir
+#    hostmount_path:
+#    - /mnt/data1
+#    - /mnt/data2
+#    limits:
+#      memory: 8Gi
+#      cpu: 4000m
+#    requests:
+#      memory: 8Gi
+#    pv_selector:
+#      label: value
+#    pvc_size: 10Gi
+#    pvc_name: pvc1
+#    storage_group: 65534
 
 # ES deployment name
 openshift_logging_elasticsearch_deployment_name: ""
--- a/roles/openshift_logging_elasticsearch/defaults/main.yml
+++ b/roles/openshift_logging_elasticsearch/defaults/main.yml
@@ -6,7 +6,6 @@ openshift_logging_elasticsearch_image_version: "{{ openshift_logging_image_versi
 openshift_logging_elasticsearch_image_pull_secret: "{{ openshift_hosted_logging_image_pull_secret | default('') }}"
 openshift_logging_elasticsearch_namespace: logging
 openshift_logging_elasticsearch_sa_name: aggregated-logging-elasticsearch
-kubeconfig: /etc/origin/master/openshift-master.kubeconfig
 openshift_logging_elasticsearch_nodeselector: "{{ openshift_logging_es_nodeselector | default('') }}"
 openshift_logging_elasticsearch_cpu_limit: "{{ openshift_logging_es_cpu_limit | default('1000m') }}"
 openshift_logging_elasticsearch_memory_limit: "{{ openshift_logging_es_memory_limit | default('1Gi') }}"
@@ -21,10 +20,9 @@ openshift_logging_elasticsearch_node_topology:
   masters:
     replicas: 1
     limits:
-      cpu: 200m
+      cpu: 1000m
     requests:
-      memory: 512Mi
-      cpu: 100m
+      memory: 1Gi
   clientdata:
   -
     node_storage_type: emptydir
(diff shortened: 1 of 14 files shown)


stevekuznetsov:
[test]

roles/openshift_logging_elasticsearch/library/openshift_logging_elasticsearch_facts.py
```diff
@@ -0,0 +1,351 @@
+'''
```
jcantrill: Do we really need this duplicated from the logging role?  If we have fact gathering for the entire logging role why do we  this here?

roles/openshift_logging_elasticsearch/templates/es.j2
```diff
@@ -66,60 +94,78 @@ spec:
                 fieldRef:
                   fieldPath: metadata.namespace
             -
-              name: "KUBERNETES_TRUST_CERT"
-              value: "true"
+              name: "NODE_NAME"
```
jcantrill: I think this is a bad idea until we can fix how the searchguard index is populated.  This essentially reverts us to before where each ES node has its own index

roles/openshift_logging_elasticsearch/vars/main.yml
```diff
@@ -10,3 +10,6 @@ es_min_masters_default: "{{ (openshift_logging_elasticsearch_replica_count | int
 es_min_masters: "{{ (openshift_logging_elasticsearch_replica_count == 1) | ternary(1, es_min_masters_default) }}"
 es_recover_after_nodes: "{{ openshift_logging_elasticsearch_replica_count | int }}"
 es_recover_expected_nodes: "{{ openshift_logging_elasticsearch_replica_count | int }}"
+es_masters_quorum: "{{ openshift_logging_elasticsearch_node_topology.masters.replicas |int /2 + 1 }}"
```
jcantrill: This is fundamentally the same as es_min_masters, shouldnt we reuse.  I actually think we can remove es_min_masters.

roles/openshift_logging_elasticsearch/vars/main.yml
```diff
@@ -10,3 +10,6 @@ es_min_masters_default: "{{ (openshift_logging_elasticsearch_replica_count | int
...
+es_recover_expected_data_nodes: "{{ openshift_logging_elasticsearch_node_topology.clientdata|length }}"
```
jcantrill: Probably need a default here.

roles/openshift_logging_elasticsearch/defaults/main.yml
```diff
@@ -1,19 +1,44 @@
...
 openshift_logging_elasticsearch_cpu_limit: 1000m
 openshift_logging_elasticsearch_memory_limit: "{{ openshift_logging_es_memory_limit | default('1Gi') }}"
 openshift_logging_elasticsearch_recover_after_time: "{{ openshift_logging_es_recover_after_time | default('5m') }}"
+openshift_logging_elasticsearch_clusterreader: true
 
 openshift_logging_elasticsearch_replica_count: 1
 
 # ES deployment type
 openshift_logging_elasticsearch_deployment_type: "data-master"
+openshift_logging_elasticsearch_node_topology:
```
t0ffel: In the inventory file people can provide JSON with this structure. is that acceptable?
i.e. smth like:
`openshift_logging_elasticsearch_node_topology='{"masters": {"replicas": 1, "limits": {"cpu": "200m"}...'`
i'll think on the translation...

roles/openshift_logging_elasticsearch/tasks/node_provision.yaml
```diff
@@ -0,0 +1,127 @@
...
+    # remove mandatory logging_component
+    # TODO: component below needs to be changed
+    component: "{{ openshift_logging_elasticsearch_clustername }}"
+    es_component: "{{ es_component }}"
+    deploy_name: "{{ es_deploy_name }}"
+    storage_type: "{{ node_storage_type | default(emptydir) }}"
+    image: "{{ openshift_logging_elasticsearch_image_prefix }}{{ openshift_logging_elasticsearch_image_name }}:{{ openshift_logging_elasticsearch_image_version }}"
+  changed_when: false
+
+- name: Set ES dc for role {{ es_role }}
```
t0ffel: After the deployment - user can scale DC w/ emptydir storage via replica settings.

In deployment settings - there is no such option. User must add section to the array
`openshift_logging_elasticsearch_node_topology.clientdata` for each replica. The logic here is that user wants to know where each node's storage land.
Is that critical to add?

roles/openshift_logging_elasticsearch/templates/es.j2
```diff
@@ -66,60 +94,78 @@ spec:
...
-              value: "logging-{{es_cluster_name}}-cluster"
+              value: "{{es_cluster_name}}-cluster"
             -
               name: "CLUSTER_NAME"
-              value: "logging-{{es_cluster_name}}"
+              value: "{{es_cluster_name}}"
             -
               name: "INSTANCE_RAM"
-              value: "{{openshift_logging_elasticsearch_memory_limit}}"
+{% if limits is defined and limits.memory is defined %}
```
t0ffel: I'll take a look

roles/openshift_logging_elasticsearch/templates/es.j2
```diff
@@ -66,60 +94,78 @@ spec:
...
+{% else %}
+              value: "512m"
+{% endif %}
             -
               name: "HEAP_DUMP_LOCATION"
               value: "/elasticsearch/persistent/heapdump.hprof"
             -
-              name: "NODE_QUORUM"
-              value: "{{es_node_quorum | int}}"
+              name: "MASTERS_QUORUM"
```
t0ffel: I'll take a look

roles/openshift_logging_elasticsearch/templates/es.j2
```diff
@@ -66,60 +94,78 @@ spec:
...
+{% for item in hostmount_path %}
+            - name: elasticsearch-storage-{{ loop.index0 }}
+              mountPath: /elasticsearch/persistent/{{ loop.index0 }}
+{% endfor %}
+{% endif %}
       volumes:
         - name: elasticsearch
           secret:
-            secretName: logging-elasticsearch
+            secretName: {{ es_cluster_name }}-certs
```
t0ffel: which health check scripts are you referring to?

roles/openshift_logging_elasticsearch/library/openshift_logging_elasticsearch_facts.py
t0ffel: The idea is that upgrades are done per elasticsearch cluster, so facts for upgrade should be gathered for individual elasticsearch cluster.

roles/openshift_logging_elasticsearch/templates/es.j2
t0ffel: well, not exactly "reverts", rather "keeps".

roles/openshift_logging_elasticsearch/defaults/main.yml
jcantrill: It is cumbersome but we may be at the mercy of the file format

roles/openshift_logging_elasticsearch/tasks/node_provision.yaml
jcantrill: We have similar asks regarding node selectors to control which ES node lands on which Openshift node.  This seems bad.  We should defer to the platform to land us on an infra node.  The specific node and storage volume is immaterial is it not?  As long as the node can find storage the specific storage does not matter.  That's how I understand ES.

With regards to client nodes that have no backing store, can they not be represented with a single DC?  We care they land on an infra node but other then that are they not all the same?  Do they need to have unique DC's

roles/openshift_logging_elasticsearch/templates/es.j2
jcantrill: the openshift health checker in ansible.  Its a specific playbook

roles/openshift_logging_elasticsearch/templates/es.j2
jcantrill: The point is if we base on DC, the seeding of the index probably only happens once where based on the pod name it will happen everytime the pod is restarted.

roles/openshift_logging_elasticsearch/library/openshift_logging_elasticsearch_facts.py
jcantrill: We should then move this to either a common role or most likely an installer library

ewolinetz: I think it's alright to keep this within the elasticsearch specific role. It isn't something that would be used anywhere else, so to me it doesn't make sense to move it into a different role or even the installer library

roles/openshift_logging_elasticsearch/vars/main.yml
t0ffel: done

roles/openshift_logging_elasticsearch/vars/main.yml
t0ffel: done

roles/openshift_logging_elasticsearch/tasks/node_provision.yaml
t0ffel: done, if `openshift_logging_es_node_topology` is not provided, we'll construct the ES cluster topology from the variables. That's default atm

roles/openshift_logging_elasticsearch/templates/es.j2
t0ffel: I didn't do it yet - I want to keep this PR small, and make the change in a subsequent commit.

roles/openshift_logging_elasticsearch/defaults/main.yml
```diff
@@ -1,19 +1,46 @@
...
+  masters:
+    replicas: 1
+    limits:
+      cpu: 200m
+    requests:
+      memory: 512Mi
```
jcantrill: Per comment from @portante  we should set limit and memory to be the same

t0ffel: Same as limits?
what should be the default for clientdata/master nodes?
Right now as I understand default is `8Gi`, so I'll set that as default for clientdata. for master, probably `1Gi`?

roles/openshift_logging_elasticsearch/defaults/main.yml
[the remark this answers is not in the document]
richm: You can also pass in parameters to `ansible-playbook -e@vars.yaml` which will be in yaml format, so you could pass in arguments in this format.

roles/openshift_logging_elasticsearch/templates/es.j2
richm: Is there something we need to do now?  I'd like to get this PR merged ASAP.

roles/openshift_logging_elasticsearch/defaults/main.yml
```diff
@@ -38,7 +36,11 @@ openshift_logging_elasticsearch_node_topology:
     requests:
       cpu: 100m
       memory: 256Mi
-
+    pv_selector:
+      label: value
+    pvc_size: 10Gi
+    pvc_name: pvc1
+    storage_group: 6566
```
richm: where does `6566` come from?

richm:
Have all of the current review comments been addressed?  Also looks like a rebase is needed.

roles/openshift_logging_elasticsearch/defaults/main.yml
jcantrill: We are talking about adding additional node which will consume infra.  We need to be prepared to justify our ask and incase more infra nodes are required.

roles/openshift_logging_elasticsearch/defaults/main.yml
t0ffel: that was a type. added the correct value 65534

roles/openshift_logging_elasticsearch/defaults/main.yml
t0ffel: Added memory limits equal to requests. `8Gi` for clientdata nodes and `1Gi` for master nodes

t0ffel:
[test]

richm:
[test]

richm:
[test]

mtnbikenc:
aos-ci-test

t0ffel:
[test]

ewolinetz:
does https://github.com/openshift/origin-aggregated-logging/pull/733 make it so this possibly isn't needed?

vrutkovs:

PR #5003, opened 2017-08-04
Origin image build: add oc client

sosiouxme:
Fixes https://github.com/openshift/origin/issues/15410

The oc client was left out with the move away from playbook2image.
However it is needed in the image for the `oc cluster up` use case for
installing logging or metrics.

This change adds oc via the latest signed and released client RPM,
which requires adding the centos-openshift-origin repo to the image.


sosiouxme:
@sdodson I had not thought about that. It's as easy as moving `ansible` to the other list if you want to get it from the centos repos; right now it's coming from `epel` (`ansible-2.3.1.0-1.el7` being the highest version), while `python-httplib2` and `python-keyczar` started coming from `centos-openshift-origin` with this change.

As nothing in CI builds or tests the image, how do you feel about using the merge button? There's no particular urgency about this bug fix I suppose, just another thing to have to track.

sosiouxme:
bot, retest this

sdodson:
I'm fine with hitting the merge button.

sdodson:
Oh, do we need to backport this to release-3.6 or any other branches where we removed playbook2image?

sosiouxme:
Yeah, I'll put it into my 3.6.1 PR that's pending.

PR #5004, opened 2017-08-04
Updating PVC generation to only be done if the pvc does not already e…

ewolinetz:
…xist to avoid idempotent issues

Addresses issue seen while trying to create a PVC when installing logging - error such that an immutable field cannot be modified

### DIFF
--- a/roles/openshift_logging_elasticsearch/tasks/main.yaml
+++ b/roles/openshift_logging_elasticsearch/tasks/main.yaml
@@ -192,47 +192,58 @@
     - port: 9200
       targetPort: "restapi"
 
-# storageclasses are used by default but if static then disable
-# storageclasses with the storageClassName set to "" in pvc.j2
-- name: Creating ES storage template - static
-  template:
-    src: pvc.j2
-    dest: "{{ tempdir }}/templates/logging-es-pvc.yml"
-  vars:
-    obj_name: "{{ openshift_logging_elasticsearch_pvc_name }}"
-    size: "{{ openshift_logging_elasticsearch_pvc_size }}"
-    access_modes: "{{ openshift_logging_elasticsearch_pvc_access_modes | list }}"
-    pv_selector: "{{ openshift_logging_elasticsearch_pvc_pv_selector }}"
-    storage_class_name: "{{ openshift_logging_elasticsearch_pvc_storage_class_name | default('', true) }}"
-  when:
-  - openshift_logging_elasticsearch_storage_type == "pvc"
-  - not openshift_logging_elasticsearch_pvc_dynamic | bool
-
-# Storageclasses are used by default if configured
-- name: Creating ES storage template - dynamic
-  template:
-    src: pvc.j2
-    dest: "{{ tempdir }}/templates/logging-es-pvc.yml"
-  vars:
-    obj_name: "{{ openshift_logging_elasticsearch_pvc_name }}"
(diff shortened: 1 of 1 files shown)


ewolinetz:
I'm not sure if this has a BZ attached to it (reaching out to see if it does) but this likely should be a 3.6 blocker since logging installation fails if there is a pre-existing PVC of the same name

To recreate:
1) Install logging using `playbooks/byo/config.yml`
2) Verify PVC bound
3) Reinstall logging using `playbooks/byo/openshift-config/openshift-logging.yml`
4) Observe error regarding immutable field

roles/openshift_logging_elasticsearch/tasks/main.yaml
```diff
@@ -194,6 +194,14 @@
...
 # storageclasses are used by default but if static then disable
 # storageclasses with the storageClassName set to "" in pvc.j2
+- name: Check to see if PVC already exists
+  oc_obj:
+    state: list
+    kind: pvc
+    name: "{{ openshift_logging_elasticsearch_pvc_name }}"
+    namespace: "{{ openshift_logging_elasticsearch_namespace }}"
+  register: logging_elasticsearch_pvc
+
```
sosiouxme: LGTM however IMHO it would look even better if the tasks went in a block instead of duplicating the conditional.

```
   - when:
      - logging_elasticsearch_pvc.results.stderr is defined
      - openshift_logging_elasticsearch_storage_type == "pvc"
     block:
       [...]
```
That way you don't have to look far to see what `logging_elasticsearch_pvc.results.stderr` is all about. I wish there were a more readable way to indicate in the conditional what it's for but I guess there's not much to be done about that.

ewolinetz: yeah... unfortunately `logging_elasticsearch_pvc.results.results | length > 0` returns a false positive... and the RC for being there and not being there is `0` in both cases. I could clean it up and move it into a block, it makes sense to me to do

ewolinetz: I'll add some comments to the when block to make it more obvious what is being checked for and why

ewolinetz:
@sosiouxme addressed comment

ewolinetz:
aos-ci-test

roles/openshift_logging_elasticsearch/tasks/main.yaml
```diff
@@ -192,47 +192,58 @@
...
-    - "{{ tempdir }}/templates/logging-es-pvc.yml"
-    delete_after: true
-  when:
+  register: logging_elasticsearch_pvc
+
+# logging_elasticsearch_pvc.results.results | length > 0 returns a false positive
+# so we check for the presence of 'stderr' to determine if the obj exists or not
+# the RC for existing and not existing is both 0
+- when:
+  - logging_elasticsearch_pvc.results.stderr is defined
```
jcantrill: Little confused here based on the comment.  You are saying that if stderr is defined that means the PVC DOES NOT EXIST and it is therefore OK to create the PVC?

ewolinetz: correct, if `stderr` is a present param on the `results` object then that means the module was not able to find a pvc that matches that name, so it doesn't exist

ewolinetz:
aos-ci-test

PR #5005, opened 2017-08-04
[3.6] Updating PVC generation to only be done if the pvc does not already e…

ewolinetz:
…xist to avoid idempotent issues

Addresses issue seen while trying to create a PVC when installing logging - error such that an immutable field cannot be modified

Backport of https://github.com/openshift/openshift-ansible/pull/5004


ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

PR #5006, opened 2017-08-04
adding check to a yaml dump to work properly with new ruamel lib

ivanhorvath:


roles/lib_openshift/src/class/oc_obj.py
```diff
@@ -50,7 +50,11 @@ def create(self, files=None, content=None):
         if files:
             return self._create(files[0])
 
-        content['data'] = yaml.dump(content['data'])
+        if hasattr(yaml, 'RoundTripDumper'):
+            content['data'] = yaml.dump(content['data'], Dumper=yaml.RoundTripDumper)
```
tbielawa: Without context I find this easily becoming confusing in the future. Would you mind adding a comment explaining why this conditional must exist?

ivanhorvath: The conditional is for safety for the cases when at the top ruamel.yaml is not loaded, thus RoundTripDumper is not present. We have these types of checks in the libraries, this one was missing it. 

The real change is that we had to use the dumper here because from rumael.yaml-0.13 onwards a regular dump won't produce a yaml that the yaml parser in openshift will understand. It is possible that down the line origin will update gopkg.in/yaml.v2 and this will be redundant, but until then we need to make sure openshift can read yamls we create.

kwoodson:
aos-ci-test

ivanhorvath:
aos-ci-test

ivanhorvath:
aos-ci-test

sdodson:
[test]

sdodson:
@tbielawa Needs review and then ready to merge.

tbielawa:
What is going on in those failures? Looks like some of those errors occurred during the period when CI was observed as having no disk space available. And am I searching poorly, or is there no flake for `Network Failure: Unable to connect to a remote server` ? @sdodson

sdodson:
[test]

kwoodson:

kwoodson:
[merge]

sdodson:
If this merge fails ping me to review the failures and I'll merge.

PR #5011, opened 2017-08-07
AWS Provisioning with scale groups.

kwoodson:
AWS Provisioning using scale groups for master, infra, and compute nodes.

Currently the work that is encompassed here:
- roles for aws to create vpc, ssh keys, encrypted amis, s3 bucket, elbs, security groups, launch configuration, and auto scaling groups
- roles to prep an ami
- playbook to build an ami
- playbook to create an entire cluster from provision to finish
- code in openshift_master to perform bootstrap preparation from master
- code added to openshift_hosted to skip the router and registry deployments

Work that will follow in separate PR:
- Only single openshift_node install path.  Merge code from ami prep role into openshift_node
- Need to version or name the launch configs differently.  If an auto scaling group references a launch config then it cannot be removed.  We would want to create a new one and update the scale group with the latest launch config name.
- Role to auto accept the nodes into the cluster from the master
- Role to deploy the autoscaler.  This is more for the FUTURE but I imagine we will want a role to do this work.

Please give thoughts or code review.


roles/lib_utils/library/oo_iam_kms.py
```diff
@@ -0,0 +1,173 @@
...
+                if 'TargetKeyId' in alias and 'KeyId' in key:
+                    if alias['TargetKeyId'] == key['KeyId']:
+                        alias.update(key)
+
+        return aliases
+
+    def get_kms_entry(self, user_alias, alias_list):
+        ''' return single alias details from list of aliases '''
+        for alias in alias_list:
+            if user_alias == alias['AliasName']:
```
ashcrow: It may be worth guarding against a missing ``AliasName`` like so:

```python
if user_alias == alias.get('AliasName', default=False):  # Or some other default ....
```

ashcrow: Assume the same recommendation for other locations checking a key later in this PR.

playbooks/aws/openshift-cluster/build_ami.yml
```diff
@@ -0,0 +1,84 @@
...
+      instance_type: m4.xlarge
+      vpc_subnet_id: "{{ subnetout.subnets[0].id }}"
+      image: "{{ provision.base_image }}"
+      volumes:
+      - device_name: /dev/sdb
+        volume_type: gp2
+        volume_size: 100
+        delete_on_termination: true
+      wait: yes
+      #user_data: "{{ lookup('file', '../cloud-init/user-data-aws') }}"
```
ashcrow: Not needed?

roles/lib_utils/library/oo_iam_kms.py
```diff
@@ -0,0 +1,173 @@
...
+
+        self.module.exit_json(failed=True,
+                              changed=False,
+                              results='Unknown state passed. %s' % state,
+                              state="unknown")
+
+
+if __name__ == '__main__':
+    KMS = AwsIamKms()
+    KMS.main()
```
ashcrow: nit: Why not ```AwsIamKms().main()```?

ashcrow:
I know other cloud providers should not be expected to be part of this PR but how would they fit into this pattern? Would they be another set of playbooks on the side? Conditional includes? etc..

roles/lib_utils/library/oo_iam_kms.py
kwoodson: Will fix.

roles/lib_utils/library/oo_iam_kms.py
kwoodson: I am just porting this over from operations but I'll fix it.

playbooks/aws/openshift-cluster/build_ami.yml
kwoodson: Removed.

roles/lib_utils/library/oo_iam_kms.py
ashcrow: It's fine, total nit 😄

kwoodson:
> I know other cloud providers should not be expected to be part of this PR but how would they fit into this pattern? Would they be another set of playbooks on the side? Conditional includes? etc..

@ashcrow, good question.  The provision.yml playbook lives under the playbooks/aws section.  The roles are all prefixed with openshift_aws_*.  The intent would be for us to support provisioning for the other clouds in their respective playbook/<cloud> and role/openshift_<cloud> directories.  I know that Clayton is going to port over his gce provisioning code into this pattern.  

Each cloud differs with their terminology and components but the end goal would be to provide a way for provisioning to occur with autoscale groups in mind.

cooktheryan:
any thoughts on create an IAM role and assigning an instance profile to the master instances?

kwoodson:
@cooktheryan, I think that instance profiles could definitely be useful.  Adding support for them doesn't look like it would be too much work.  I envision there being a role to create / manage them.

What types of things are you managing with your instance profiles?  What types of data are you passing or which APIs are you calling?

I can see quite a few possibilities but we also need to be mindful of the online first effort.  If our nodes in openshift online become compromised, what does this mean for those instances? This depends on what capabilities are granted in the IAM role.

kwoodson:
aos-ci-test

cooktheryan:
@kwoodson we should be able to use the same exact Ops account roles as they have created now. By using the instance profile/role based access it allows us the ability to not have to roll AWS access and secret keys every so often. Feel free to ping me and IRC with the ops privileges and I can help create an instance profile/role for review

playbooks/aws/openshift-cluster/provision.yml
```diff
@@ -0,0 +1,264 @@
...
+      r_openshift_aws_ssh_keys_region: "{{ provision.region }}"
+
+  - name: create s3 bucket for registry
+    include_role:
+      name: openshift_aws_s3
+    vars:
+      r_openshift_aws_s3_clusterid: "{{ provision.clusterid }}-docker-registry"
+      r_openshift_aws_s3_region: "{{ provision.region }}"
+      r_openshift_aws_s3_mode: create
+    when: provision.openshift_registry_s3 | default(false)
```
twiest: Did you want this `when` clause at the top of the task?

playbooks/aws/openshift-cluster/provision.yml
```diff
@@ -0,0 +1,264 @@
...
+      r_openshift_aws_vpc_tags: "{{ provision.vpc.tags }}"
+
+  - name: create aws ssh keypair
+    include_role:
+      name: openshift_aws_ssh_keys
+    vars:
+      r_openshift_aws_ssh_keys_users: "{{ provision.instance_users }}"
+      r_openshift_aws_ssh_keys_region: "{{ provision.region }}"
+
+  - name: create s3 bucket for registry
```
michaelgugino: This role is pretty small, probably doesn't need to be a role itself.

playbooks/aws/openshift-cluster/build_ami.yml
```diff
@@ -0,0 +1,83 @@
...
+  connection: local
+  become: no
+  tasks:
+  - name: bundle ami
+    ec2_ami:
+      instance_id: "{{ amibase.tagged_instances.0.id }}"
+      region: "{{ provision.region }}"
+      state: present
+      description: "This was provisioned {{ ansible_date_time.iso8601 }}"
+      name: "rhel7-openshift-gi-{{ lookup('pipe', 'date +%Y%m%d%H%M')}}"
```
michaelgugino: base image is configurable, might not be rhel7.

playbooks/aws/openshift-cluster/provision.yml
```diff
@@ -0,0 +1,264 @@
...
+
+    - name: create launch config
+      include_role:
+        name: openshift_aws_launch_config
+      vars:
+        r_openshift_aws_launch_config_clusterid: "{{ provision.clusterid }}"
+        r_openshift_aws_launch_config_region: "{{ provision.region }}"
+        r_openshift_aws_launch_config: "{{ provision.node_group_config }}"
+        r_openshift_aws_launch_config_type: "{{ launch_item }}"
+        r_openshift_aws_launch_config_bootstrap_token: "{{ local_bootstrap['content'] |b64decode }}"
```
michaelgugino: missed a space after |

playbooks/aws/openshift-cluster/vars.yml
```diff
@@ -31,3 +31,100 @@ deployment_vars:
   enterprise: "{{ deployment_rhel7_ent_base }}"
   openshift-enterprise: "{{ deployment_rhel7_ent_base }}"
   atomic-enterprise: "{{ deployment_rhel7_ent_base }}"
+
+provision:
```
michaelgugino: As noted before, setting variables in include_vars files makes them difficult to override for users.  You'll notice, most of the other variables set in this file refer back to other variables.

playbooks/aws/openshift-cluster/vars.yml
```diff
@@ -31,3 +31,100 @@ deployment_vars:
...
+  cert_arn: "arn:aws:iam::000000000000:server-certificate/<clusterid>_<domain>"
+  iam_cert_ca:
+    name: test_openshift
+    cert_path: '/path/to/wildcard.<clusterid>.example.com.crt'
+    key_path: '/path/to/wildcard.<clusterid>.example.com.key'
+    chain_path: '/path/to/cert.ca.crt'
+  yum_repositories:
```
michaelgugino: we probably shouldn't nest all these dictionaries in provision.  That would require someone wanting to override provision itself to hand-set every single dictionary here.  Some of these could be their own top-level variable, such as
"provision_yum_repos"

playbooks/aws/openshift-cluster/provision.yml
```diff
@@ -0,0 +1,264 @@
...
+      osm_custom_cors_origins:
+      - "{{ elbs.results[1].elbs[0].dns_name }}"
+      - "console.{{ provision.clusterid }}.openshift.com"
+      - "api.{{ provision.clusterid }}.openshift.com"
+    with_items: "{{ groups['masters'] }}"
+
+  - name:
+    debug: var=hostvars
+
+- include: ../../common/openshift-cluster/config.yml
```
michaelgugino: I'm still trying to figured out all the different paths through the plays, but this looks like a potential spot to break this play up into 3 (or more) plays:

```
#provision-site.yml
include <this play.yml>
include .....config.yml
include <the rest of this play.yml>
include <what happens next>
```

Just my opinion, fwiw.

playbooks/aws/openshift-cluster/build_ami.yml
```diff
@@ -0,0 +1,83 @@
...
+      timeout: 300
+      search_regex: OpenSSH
+
+  - name: add host to group
+    add_host:
+      name: "{{ amibase.tagged_instances.0.public_dns_name }}"
+      groups: amibase
+
+- hosts: amibase
+  remote_user: root
```
michaelgugino: remote_user needs to be configurable if we support base rhel image to build from, probably should also add support for become_user in args somewhere.

playbooks/aws/openshift-cluster/provision.yml
```diff
@@ -0,0 +1,264 @@
+---
```
kwoodson: I'm not sure this makes sense as a role.  I think that running a playbook that consists of the aws building blocks makes a bit more sense to me.

playbooks/aws/openshift-cluster/provision.yml
kwoodson: don't think the boundary for a role is based on size.  I base roles on the work that it accomplishes.

playbooks/aws/openshift-cluster/provision.yml
kwoodson: The first part of the provision.yml brings up aws components and launches masters.  Once these are up we then need to configure them.  Once configured, we then bring up the nodes in scale groups.  This is required because of the user token that is generated on the masters during install.  

I could see us drawing a boundary around sg, lc, asg because that is something we call for the masters, infra, and then compute nodes.  We could abstract that into a separate playbook.

playbooks/aws/openshift-cluster/provision.yml
kwoodson: Thanks, I'll add it.

playbooks/aws/openshift-cluster/build_ami.yml
kwoodson: We should better parameterize this.

playbooks/aws/openshift-cluster/vars.yml
kwoodson: We could flatten the structure a bit but I'm a fan of using a top level as it provides a namespace for the underlying variables.

cooktheryan:
@kwoodson will we initially be shipping rhel as our golden image AMI platform or would we also be looking at Atomic?

ashcrow:
@cooktheryan @kwoodson hopefully it would be both.

cooktheryan:
@ashcrow we may have to do some fun hacker-y around atomic. I just noticed it when deploying the latest ref arch. Cloud-init doesn't grow root on atomic since the storage is managed from the docker-storage-setup file.

playbooks/aws/openshift-cluster/vars.yml
michaelgugino: Yeah, I'm not against name spacing things and I actually like nested dictionaries for a lot of things.  We should just break out the nest dictionaries to refer back to other variables.  IE:

```
provision_yum_repos: ... some dict ...

provision:
  yum_repositories:  {{ provision_yum_repos }}
```

This will give us maximum flexibility and allow users to override just portions of the larger dictionary.

kwoodson:
@ashcrow, I have updated the README.md file for `playbooks/aws/`.

kwoodson:
aos-ci-test

kwoodson:
squashed

kwoodson:
@ashcrow, fixed up the readme files.

ashcrow:
aos-ci-test

kwoodson:
[merge]

kwoodson:
``` 
   Test Panicked
    runtime error: index out of range
    /usr/local/go/src/runtime/panic.go:489

    Full Stack Trace
    	/usr/local/go/src/runtime/panic.go:489 +0x2cf
    github.com/openshift/origin/test/extended/router.glob..func2.1()
    	/go/src/github.com/openshift/origin/test/extended/router/metrics.go:62 +0x6ae
    github.com/openshift/origin/vendor/github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync(0xc42067fbc0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
    	/go/src/github.com/openshift/origin/test/extended/util/test.go:120 +0x3d4
    github.com/openshift/origin/test/extended.TestExtended(0xc420e181a0)
    	/go/src/github.com/openshift/origin/_output/local/go/src/github.com/openshift/origin/test/extended/extended_test.go:35 +0x40
    testing.tRunner(0xc420e181a0, 0x36198b8)
    	/usr/local/go/src/testing/testing.go:657 +0x96
    created by testing.(*T).Run
    	/usr/local/go/src/testing/testing.go:697 +0x2ca
```

kwoodson:
[merge]

kwoodson:
flake https://github.com/openshift/origin/issues/13975

kwoodson:
[merge]

sdodson:
The only failure is on a known flake.

PR #4960, opened 2017-08-01
add pre-flight checks to ugrade path

juanvallejo:
Followup to https://github.com/openshift/openshift-ansible/pull/4923

cc @rhcarvalho @sosiouxme

PR #5014, opened 2017-08-07
Adding additonal checks upgrade path

juanvallejo:
Depends on https://github.com/openshift/openshift-ansible/pull/4960

TODO

 - Possibly handle `upgrade` playbook context on `etcd_volume` check

cc @sosiouxme @rhcarvalho

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/pre/verify_health_checks.yml
+++ b/playbooks/common/openshift-cluster/upgrades/pre/verify_health_checks.yml
@@ -11,3 +11,4 @@
       checks:
       - disk_availability
       - memory_availability
+      - docker_image_availability


sosiouxme:
@juanvallejo it seems to me the etcd checks can be used as-is. They're both intended for already-installed clusters. Any thoughts what would be different about an upgrade?

sosiouxme:
bot, retest this

juanvallejo:
@sosiouxme 
> it seems to me the etcd checks can be used as-is. They're both intended for already-installed clusters. 
> Any thoughts what would be different about an upgrade?

I was not sure if I was missing any special cases during an upgrade that would need to be handled in the etcd checks. I cannot think of any, so I was hoping I could get enough people to review this and either confirm that nothing else is needed, or catch whatever we're missing

juanvallejo:
[test]

playbooks/common/openshift-cluster/upgrades/pre/verify_health_checks.yml
```diff
@@ -11,3 +11,6 @@
       checks:
       - disk_availability
       - memory_availability
+      - docker_image_availability
+      - etcd_imagedata_size
```
rhcarvalho: Should we block an upgrade if "etcd has too much image data"? That may not be a good idea.

playbooks/common/openshift-cluster/upgrades/pre/verify_health_checks.yml
```diff
@@ -11,3 +11,6 @@
...
+      - etcd_volume
```
rhcarvalho: Similarly, this might be not-so-good. If an upgrade is actually necessary to reduce etcd usage, this check prior to upgrade could block getting the cluster to a healthier state.

playbooks/common/openshift-cluster/upgrades/pre/verify_health_checks.yml
sosiouxme: checks can be disabled...

but I guess the question is really "are we running health checks generally or just looking for issues we think might impact the upgrade?"

to the extent that an upgrade might exercise data in etcd, the etcd checks probably make sense either way. you do not want your upgrade to fail partway through due to running out of etcd space, although it was probably going to fill up imminently anyway, at least you don't want to have to deal with that in the middle of an upgrade.

juanvallejo:
@rhcarvalho Went ahead and removed `etcd_volume` from upgrade path.
Per @sosiouxme 's [comment](https://github.com/openshift/openshift-ansible/pull/5014#discussion_r132222224), I think it would be worth keeping the `etcd_imagedata_size` check in the upgrade path

juanvallejo:
re[test]

juanvallejo:
flaked on https://github.com/openshift/origin/issues/8571
re[test]

juanvallejo:
flake https://github.com/openshift/origin/issues/10162
re[test]

sosiouxme:
Might as well try to [merge] this.

sosiouxme:
aos-ci-test

sosiouxme:
try a [merge]

sosiouxme:
[853](https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/853/)
In addition to https://github.com/openshift/origin/issues/15769 we have https://github.com/openshift/openshift-ansible/pull/5014

[merge] again

sosiouxme:
[Time wasted in the queue](https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/857/)

https://github.com/openshift/origin/issues/15769 again which should be fixed now.

I'm not sure what this is; there doesn't seem to be a flake issue for it:

```
[ERROR] wait_for_fluentd_to_catch_up: not found 1 record project .operations for 2b30b48d-248c-461a-8f27-ca64cf4a72ec after 300 seconds
[ERROR] Checking journal for 2b30b48d-248c-461a-8f27-ca64cf4a72ec...
[ERROR] Found 2b30b48d-248c-461a-8f27-ca64cf4a72ec in journal
No resources found.
error: expected 'logs (POD | TYPE/NAME) [CONTAINER_NAME]'.
POD or TYPE/NAME is a required argument for the logs command
See 'oc logs -h' for help and examples.
[ERROR] PID 119484: hack/testing/test-fluentd-forward.sh:171: `oc logs $ffpod > $ARTIFACT_DIR/test-fluentd-forward.forward-fluentd.log` exited with status 1.
```

Looks like it might be a logging flake, i.e. the logging aggregation didn't start up / catch up fast enough?

I can say [merge] again or @sdodson can save some queue time and merge this since other tests seem fine.

juanvallejo:
[test]

sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
[test] there have been some fixes to the logging job

sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
The test failure seems real, where do we get python-etcd from? We should add it to openshift-ansible.spec as a requirement if it's required on local host.

juanvallejo:
@sdodson 
> The test failure seems real, where do we get python-etcd from? We should add it to openshift-ansible.spec as a requirement if it's required on local host.

So, it is available by default in the "updates" repo on Fedora, but I don't think there is a repo for it on rhel yet.

If it is easier, I could remove the `etcd_imagedata_size` check from this PR, and just add the `docker_image_availability` check to the upgrade path for the time being. The `etcd_imagedata_size` check is not currently on the install path.

cc @sosiouxme

rhcarvalho:
Hmm, looks like aos-ci-test results are missing. Plus possibly some flakes in https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/944/.

rhcarvalho:
aos-ci-test

juanvallejo:
cc @brenton 
I think the test failures were flakes

rhcarvalho:
re-[merge]

rhcarvalho:
Previous merge flake was https://github.com/openshift/origin/issues/16005, and yum.

re-[merge]

juanvallejo:
[test]

juanvallejo:
@sosiouxme or @rhcarvalho tests seem to be passing, mind tagging once more?

rhcarvalho:
@juanvallejo I think we're only merging bug fixes this week

sosiouxme:
I think for the new CI we also need to:
/retest

juanvallejo:
/test tox
test install

sosiouxme:
CI fail
/test install

PR #5020, opened 2017-08-07
Prometheus endpoint config for hawkular-metrics and cassandra

jsanda:
Prometheus endpoints are exposed in hawkular-metrics and in cassandra by default. This PR adds options to disable them.

### DIFF
--- a/roles/openshift_metrics/defaults/main.yaml
+++ b/roles/openshift_metrics/defaults/main.yaml
@@ -60,5 +60,5 @@ openshift_metrics_hawkular_user_write_access: False
 
 openshift_metrics_heapster_allowed_users: system:master-proxy
 
-openshift_metrics_cassandra_disable_prometheus_endpoint: False
-openshift_metrics_hawkular_disable_prometheus_endpoint: False
+openshift_metrics_cassandra_enable_prometheus_endpoint: True
+openshift_metrics_hawkular_enable_prometheus_endpoint: True
--- a/roles/openshift_metrics/defaults/main.yaml
+++ b/roles/openshift_metrics/defaults/main.yaml
@@ -59,3 +59,6 @@ openshift_metrics_cassandra_pvc_access: "{{ openshift_hosted_metrics_storage_acc
 openshift_metrics_hawkular_user_write_access: False
 
 openshift_metrics_heapster_allowed_users: system:master-proxy
+
+openshift_metrics_cassandra_disable_prometheus_endpoint: ""
+openshift_metrics_hawkular_disable_prometheus_endpoint: ""
--- a/roles/openshift_metrics/defaults/main.yaml
+++ b/roles/openshift_metrics/defaults/main.yaml
@@ -60,5 +60,5 @@ openshift_metrics_hawkular_user_write_access: False
 
 openshift_metrics_heapster_allowed_users: system:master-proxy
 
-openshift_metrics_cassandra_disable_prometheus_endpoint: ""
-openshift_metrics_hawkular_disable_prometheus_endpoint: ""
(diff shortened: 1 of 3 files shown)


roles/openshift_metrics/defaults/main.yaml
```diff
@@ -59,3 +59,6 @@ openshift_metrics_cassandra_pvc_access: "{{ openshift_hosted_metrics_storage_acc
 openshift_metrics_hawkular_user_write_access: False
 
 openshift_metrics_heapster_allowed_users: system:master-proxy
+
+openshift_metrics_cassandra_disable_prometheus_endpoint: ""
+openshift_metrics_hawkular_disable_prometheus_endpoint: ""
```
mwringe: Should these not be set to false by default?

jsanda: No because I just check a zero-length string.

jsanda: After further discussion I will change these properties to be false by default.

mwringe:
This looks good to me.

@jcantrill @ewolinetz can one of you please review this?

roles/openshift_metrics/defaults/main.yaml
```diff
@@ -59,3 +59,6 @@ openshift_metrics_cassandra_pvc_access: "{{ openshift_hosted_metrics_storage_acc
 openshift_metrics_hawkular_user_write_access: False
 
 openshift_metrics_heapster_allowed_users: system:master-proxy
+
+openshift_metrics_cassandra_disable_prometheus_endpoint: False
```
jcantrill: Wouldn't it make more sense to invert the language so its easier to understand and read, no double negatives: 'openshift_metrics_cassandra_enable_prometheus_endpoint: True'

jsanda: I agree that double negatives are confusing. I will change this.

roles/openshift_metrics/defaults/main.yaml
```diff
@@ -59,3 +59,6 @@ openshift_metrics_cassandra_pvc_access: "{{ openshift_hosted_metrics_storage_acc
 openshift_metrics_hawkular_user_write_access: False
 
 openshift_metrics_heapster_allowed_users: system:master-proxy
+
+openshift_metrics_cassandra_enable_prometheus_endpoint: True
```
ewolinetz: is there ever a time where we one of these would be true and the other would be false?

mwringe: Someone could want to collect metrics just for one or not the other. Or we may want to disable metrics in one to see if that is causing problems.

ewolinetz: @mwringe ack, would that be a scenario at installation or post installation? I'm just trying to figure out if this should be a single variable or two. If they are mutually exclusive then it makes sense to keep them as two.

mwringe: @ewolinetz sorry, I lost track of this pr, these are separate independent variables, we should have 2.

mwringe:
@ewolinetz is there anything else we need to get done for this issue? I would like to get this closed soon.

ewolinetz:
@mwringe i'll put this through the testing paces today

mwringe:
@ewolinetz anything else we need to do to get this merged?

ewolinetz:
@mwringe with that latest flag the bot should come back around and pick it up and then drop it into the merge queue

sdodson:
Merge gating on kind has been enabled now so from devcut through the end of the sprint we're only supposed to be merging bugfixes.

PR #5023, opened 2017-08-08
Reconcile only needs to run on one master

smarterclayton:
These are operations that apply to the API server and should not be run in
parallel.

@sdodson @liggitt I don't think we should be running reconcile on all masters. That actually reduces the risk of problems by potentially causing conflicts between different reconcilers, since they apply only to etcd.


liggitt:
seems ok

sdodson:
@smarterclayton Can you get rid of all the run_once: true in the play you've modified the hosts for, they're no longer necessary.

abutcher:
aos-ci-test

michaelgugino:
@smarterclayton does this still need to ship?

PR #4789, opened 2017-07-19
Allow to specify docker registry for system containers

ingvagabund:

PR #5024, opened 2017-08-08
Master image pull fails

jkaurredhat:
When using

openshift_use_system_containers=True

it fails  as :

system_images_registry is not available in a dict

### DIFF
--- a/roles/openshift_master/tasks/system_container.yml
+++ b/roles/openshift_master/tasks/system_container.yml
@@ -1,7 +1,7 @@
 ---
 - name: Pre-pull master system container image
   command: >
-    atomic pull --storage=ostree {{ openshift.common.system_images_registry }}/{{ openshift.master.master_system_image }}:{{ openshift_image_tag }}
+    atomic pull --storage=ostree {{ 'docker:' if openshift.common.system_images_registry == 'docker' else openshift.common.system_images_registry + '/' }}{{ openshift.master.master_system_image }}:{{ openshift_image_tag }}
   register: pull_result
   changed_when: "'Pulling layer' in pull_result.stdout"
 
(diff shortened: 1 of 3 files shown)


sdodson:
If this isn't a backport we need to make the change on master first.

giuseppe:
Please specify system_images_registry, we need fully specified image names for system containers. We need to document this

ingvagabund:
This looks like the backport from the master.

ashcrow:
@ingvagabund @jkaurredhat where you folks not setting ``system_images_registry``?

ingvagabund:
It needs to be set only when `openshift_use_system_containers=True`. The #4789 is needed to run the Ci for system containers. And yes, it needs to be backported to 3.6 as well.

ingvagabund:
aos-ci-test

ashcrow:
Ahh ok, it's for CI more than actual usage. In that case I'm OK with this.

ingvagabund:
@jkaurredhat the `docker:` is here to pull the system container images in a case the images are built in docker on the same host. This should not be a valid use case in the production and we should come with a different way how to handle this use case.

ashcrow:
@ingvagabund agreed. I think as @giuseppe pointed out this is more of a lack of proper documentation letting users know they have to set the variable.

ashcrow:
FWIW I'll try to throw some minor documentation together in master to help keep this from happening in the future.

giuseppe:
We could also set proper default values, docker hub for origin and access redhat.com for ose. What do you think?

ashcrow:
@giuseppe I'm not against that, but I think having 3 registries will become harder to maintain. As an example check the bug linked from #5031.

ashcrow:
😕 
```
    "msg": "CA certificate /etc/origin/master/ca.crt doesn't exist on CA host 10.8.168.82. Apply 'openshift_ca' role to 10.8.168.82.\n"
```

ingvagabund:
@ashcrow this PR is not covered by the CI yet. So it can be safely merged. @sdodson ?

ashcrow:
@giuseppe @ingvagabund @jkaurredhat https://github.com/openshift/openshift-ansible/pull/5033

giuseppe:
@ashcrow in this case it will just be needed to reflect what we already do when setting the docker additional registries. Users will still be able to override it if required

PR #5025, opened 2017-08-08
Introduce r_openshift_node_facts prefixed role variables

ingvagabund:
Removing all free variables from the role, replacing all of then with r_openshift_node_facts prefixed variables.

Given the openshift_node consumes all node facts either via openshift.node or free variables, the openshift_node_facts does not need to be openshift_node's dependency as long as we make sure proper node facts are collected before the openshift_node role is run. I am not removing the dependency on the assumption the node facts role gets run ahead, but on the assumption the node facts can change over time and the facts needs to be collected only once and not each time the openshift_node role gets run. At the same time the node facts are set to default values so the openshift_node role should work on its own. Plus, removing the dependency will allow us to set custom node facts during testing to cover more combinations of the node facts.

The `generate.json` file is read by new generators from [1]. it generates the role variable table in `README.md` and the `set_optional_node_facts_role_variables.yml`.

**In depth**
In a process of moving all global variables out of the `openshift_node_facts` role I had to do the following:
1. For each global variable used inside the role create corresponding new variable prefixed with `r_openshift_node_facts_`  string, e.g. for `openshift_node_debug_level` create `r_openshift_node_facts_debug_level`.
2. Given most of the variables has default value attach, I moved the default value of each role variable under `defaults/main.yml`. E.g. the `openshift_node_proxy_mode | default('iptables')` resulted in `r_openshift_node_facts_proxy_mode: iptables` line.
3. Set all role variables to their corresponding global variables. By the nature of `openshift_node_proxy_mode | default('iptables')` expression, one can use the following form:
   ```yaml
   include_role:
     name: openshift_node_facts
   vars:
     r_openshift_node_facts_proxy_mode: "{{ openshift_node_proxy_mode | default('iptables') }}"
   ```

   However, the default value (that is set inside `default/main.yml` gets duplicated). For that reason I decided to use the following form instead:

   ```yaml
   set_fact:
      r_openshift_node_facts_proxy_mode: "{{ openshift_node_proxy_mode }}"
   when: openshift_node_proxy_mode is defined

   include_role:
     name: openshift_node_facts
   ```
   It does not look good but the default value is specified only once. Plus, the `set_fact: ... when: ...` syntax is pretty generic, it can be automatically generated (see `playbooks/common/openshift-node/set_optional_node_facts_role_variables.yml`)
4. Verify role variables: a role has a list of role variables that needs to be defined, a list of variables that can be set to a restricted set of value or a list of variables that depends on other variables or their values. 
This role has only one check (see top of `roles/openshift_node_facts/tasks/main.yml`). For more checks see https://github.com/openshift/openshift-ansible/pull/5027/files#diff-ca0eb3d2f4b4907fcf15212649d257e9

[1] https://github.com/openshift/openshift-ansible/pull/4998

### DIFF
--- a/roles/openshift_node_facts/README.md
+++ b/roles/openshift_node_facts/README.md
@@ -1,0 +1,55 @@
+OpenShift/Atomic Enterprise Node Facts
+================================
+
+Node facts collector
+
+Requirements
+------------
+
+* Ansible 2.2
+
+Role Variables
+--------------
+
+| name                                           | description                              | default  | required | choices |
+|------------------------------------------------|------------------------------------------|----------|----------|---------|
+| r_openshift_node_facts_annotations             | Node annotations                         | None     |          |         |
+| r_openshift_node_facts_debug_level             | Node debug lovel                         |          |          |         |
+| r_openshift_node_facts_dns_ip                  |                                          |          |          |         |
+| r_openshift_node_facts_env_vars                |                                          | None     |          |         |
+| r_openshift_node_facts_iptables_sync_period    |                                          | None     |          |         |
+| r_openshift_node_facts_kubelet_args            |                                          | None     |          |         |
+| r_openshift_node_facts_labels                  | Node labels                              | None     |          |         |
+| r_openshift_node_facts_local_quota_per_fsgroup |                                          | None     |          |         |
+| r_openshift_node_facts_oreg_url                |                                          | None     |          |         |
+| r_openshift_node_facts_osn_image               |                                          | None     |          |         |
+| r_openshift_node_facts_osn_ovs_image           |                                          | None     |          |         |
+| r_openshift_node_facts_osn_storage_plugin_deps | Store plugins to install on the node     | None     |          |         |
+| r_openshift_node_facts_proxy_mode              |                                          | iptables |          |         |
+| r_openshift_node_facts_schedulable             | If set pods can be scheduled on the node | None     |          |         |
+| r_openshift_node_facts_sdn_mtu                 |                                          | None     |          |         |
+| r_openshift_node_facts_set_node_ip             |                                          | None     |          |         |
+
+
+Dependencies
+------------
+
+openshift_facts
+
+Example Playbook
+----------------
+
+Notes
+-----
+
+TODO
+
+License
+-------
+
+Apache License, Version 2.0
+
+Author Information
+------------------
+
+TODO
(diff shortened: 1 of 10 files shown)


ingvagabund:
aos-ci-test

roles/openshift_node_facts/README.md
```diff
@@ -0,0 +1,55 @@
...
+| r_openshift_node_facts_local_quota_per_fsgroup |                                          | "{{ None }}" |          |
+| r_openshift_node_facts_oreg_url                |                                          | "{{ None }}" |          |
+| r_openshift_node_facts_osn_image               |                                          | "{{ None }}" |          |
+| r_openshift_node_facts_osn_ovs_image           |                                          | "{{ None }}" |          |
+| r_openshift_node_facts_osn_storage_plugin_deps | Store plugins to install on the node     | "{{ None }}" |          |
+| r_openshift_node_facts_proxy_mode              |                                          | iptables     |          |
+| r_openshift_node_facts_schedulable             | If set pods can be scheduled on the node | "{{ None }}" |          |
+| r_openshift_node_facts_sdn_mtu                 |                                          | "{{ None }}" |          |
+| r_openshift_node_facts_set_node_ip             |                                          | "{{ None }}" |          |
```
ingvagabund: I need to update the generator to process the `"{{ None }}"` default value correcfly

ingvagabund:
aos-ci-test

ingvagabund:
aos-ci-test

tbielawa:
I recall a great deal of discussion around this PR during the arch meeting on Wednesday. Are there any changes expected as an outcome of that? What's next?

ingvagabund:
> I recall a great deal of discussion around this PR during the arch meeting on Wednesday. Are there any changes expected as an outcome of that? What's next?

The action item was to be more verbose about the changes carried. I extended the PR description with more information so it is aligned with my effort. Feel free to ask any other questions.

ingvagabund:
@kwoodson let's move the discussion about facts in general to a different issue/PR. This PR is purely for introducing role variables.

ashcrow:
FWIW the item I brought up was the creation of defaults for variables and then later checking if the same items were defined as it confused me.

ingvagabund:
> FWIW the item I brought up was the creation of defaults for variables and then later checking if the same items were defined as it confused me.

Only variables that do not have a default value are checked. Can you more elaborate on your comment?

ingvagabund:
It will sit here for a while until I am done with the etcd refactoring, the `openshift_node` and related code is next.

michaelgugino:
I don't believe this is needed any longer.

PR #5028, opened 2017-08-08
More complete discovery of entry point playbooks

mtnbikenc:
This change scans for supported entry point playbooks based on the following conditions:
* In the supported directory of playbooks/byo
* Playbooks not included by any other playbooks

The previous method of discovering entry point playbooks by searching files for `intialize_hosts.yml` was incomplete and would only find properly formatted playbooks.  This method finds all playbooks, as well as new playbooks, which are not included by other playbooks.

Note: Some playbooks which may be a documented entry point, such as byo/config.yml, are not directly identified by this method because of higher up playbooks which also include those playbooks.
* byo/config.yml is documented as an entry point however it is just a placeholder that calls byo/openshift-cluster/config.yml.
* byo/config.yml is also included from byo/vagrant.yml
* byo/openshift-checks/pre-install.yml is an actual entry point, however byo/openshift-preflight/check.yml is maintained as a documented entry point

Test affected:
```
tox -e py27-ansible_syntax
```


setup.py
```diff
@@ -221,27 +221,43 @@ def run(self):
...
-                            print('{}Execution failed: {}{}'.format(
-                                self.FAIL, cpe, self.ENDC))
-                            has_errors = True
-                        # Break for loop, no need to continue looping lines
-                        break
+                for task in yaml.safe_load(contents.read()):
+                    if not isinstance(task, dict):
+                        # Skip yaml files which do not contain plays or includes
+                        continue
+                    if 'include' in task.keys():
```
ashcrow: nit: ``keys`` is not needed.

```python
if 'include' in task:
```

setup.py
```diff
@@ -221,27 +221,43 @@ def run(self):
...
+                        # Add the playbook and capture included playbooks
+                        playbooks.add(yaml_file)
+                        included_file_name = task['include'].split()[0]
+                        included_file = os.path.normpath(
+                            os.path.join(os.path.dirname(yaml_file),
+                                         included_file_name))
+                        included_playbooks.add(included_file)
+                    elif 'hosts' in task.keys():
```
ashcrow: nit: ``keys`` is not needed.

```python
elif 'hosts in task:
```

kwoodson:
@mtnbikenc,  Looking through this code I see that you tied into the syntax check.  Was there a reason why you chose this and not creating a separate check?

I think for efficiency sake, we are already traversing the files, and the logic is quite similar so I think this is probably fine.  If we want to expound or change anything in the future we can break this out.

setup.py
```diff
@@ -221,27 +221,43 @@ def run(self):
         ''' run command '''
 
         has_errors = False
+        playbooks = set()
```
ashcrow: 👍 for using ``set``s!

setup.py
```diff
@@ -221,27 +221,43 @@ def run(self):
...
-                                ['ansible-playbook', '-i localhost,',
-                                 '--syntax-check', yaml_file]
-                            )
-                        except subprocess.CalledProcessError as cpe:
-                            print('{}Execution failed: {}{}'.format(
-                                self.FAIL, cpe, self.ENDC))
-                            has_errors = True
-                        # Break for loop, no need to continue looping lines
-                        break
+                for task in yaml.safe_load(contents.read()):
```
ashcrow: nit: I realize this wasn't something added in this PR but technically ``contents`` is a file instance. It reads a bit off to me to need to ``read`` the ``contents`` to get the contents of the instance.

mtnbikenc:
@kwoodson The original syntax checking was finding playbooks to check only by the existence of 'initialize_hosts.yml` being in the playbook.  This is not a reliable way of finding the actual playbooks to check.  This new method finds all the playbooks that _could_ be run and is not tied to just the inclusion of a particular file.  I'm attempting to improve the completeness of our syntax checking by making this change.

setup.py
mtnbikenc: Actually, `keys` is needed.  I ran into a case when the word `hosts` was merely present in the text of the `dict` as a whole.  This generated a false positive, so I was more specific in looking at the `keys` of the `dict` and not the whole `dict`.

setup.py
mtnbikenc: Same situation here as in `hosts` below, just being specific.  This will also need to be extended once we move to Ansible 2.4 and start switching over to `import_playbook`.

setup.py
mtnbikenc: Updated

setup.py
ashcrow: I've not seen that before, but I believe you ... though I can't reproduce it.


## Python 3
```
In [1]: a = {'a': 'b'}

In [2]: 'a' in a
Out[2]: True

In [3]: 'b' in a
Out[3]: False
```

## Python 2
```
>>> a = {'a': 'b'}
>>> 'a' in a
True
>>> 'b' in a
False
>>> 
```

In the end it's a nit though so 👍 😄

mtnbikenc: I've unraveled the mystery.  The problem was two fold and hidden by the first.  What I was actually encountering was a difference in the type of object being read from the yaml file.  In one instance ([cluster_hosts.yml](https://github.com/openshift/openshift-ansible/blob/master/playbooks/byo/openshift-cluster/cluster_hosts.yml)), I was presented with a `str` and was getting a positive on `hosts` in that `str`.  When I ultimately added a check to skip the task if it was not a `dict`, I was able to bypass that false positive.  So, I don't really need the `keys` as you suggested since I'm already verifying I'm working with a `dict`.  Updated.  :smiley:

ashcrow: @mtnbikenc great debugging!!

mtnbikenc:
aos-ci-test

mtnbikenc:
aos-ci-test

mtnbikenc:
[merge]

mtnbikenc:
Flake: https://github.com/openshift/origin/issues/14898 yum: no mirrors to try

kwoodson:
flake again on yum: https://github.com/openshift/origin/issues/8571

mtnbikenc:
[merge]

PR #5036, opened 2017-08-08
openshift_checks: allow OVS 2.7 on OCP 3.5 and 3.6

Miciah:
`rpm_version`: Allow `package_list` items to specify a list value for `version`.  If a list value is provided for a package, pass the check if any version in that list is found.

`ovs_version`: Specify both 2.6 and 2.7 as allowed versions of OVS for OpenShift versions 3.5 and 3.6.

### DIFF
--- a/roles/openshift_health_checker/library/rpm_version.py
+++ b/roles/openshift_health_checker/library/rpm_version.py
@@ -4,6 +4,7 @@ Ansible module for rpm-based systems determining existing package version inform
 """
 
 from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.six import string_types
 
 IMPORT_EXCEPTION = None
 try:
@@ -82,11 +83,16 @@ def _check_pkg_versions(found_pkgs_dict, expected_pkgs_dict):
             continue
 
         found_versions = [_parse_version(version) for version in found_pkgs_dict[pkg_name]]
-        expected_version = _parse_version(pkg["version"])
-        if expected_version not in found_versions:
+
+        if isinstance(pkg["version"], string_types):
+            expected_versions = [_parse_version(pkg["version"])]
+        else:
+            expected_versions = [_parse_version(version) for version in pkg["version"]]
+
+        if not set(expected_versions) & set(found_versions):
             invalid_pkg_versions[pkg_name] = {
                 "found_versions": found_versions,
-                "required_version": expected_version,
+                "required_versions": expected_versions,
             }
 
     if not_found_pkgs:
@@ -106,7 +112,7 @@ def _check_pkg_versions(found_pkgs_dict, expected_pkgs_dict):
                 "The following packages were found to be installed with an incorrect version: {}".format('\n'.join([
                     "    \n{}\n    Required version: {}\n    Found versions: {}".format(
                         pkg_name,
-                        pkg["required_version"],
+                        ', '.join(pkg["required_versions"]),
                         ', '.join([version for version in pkg["found_versions"]]))
                     for pkg_name, pkg in invalid_pkg_versions.items()
                 ]))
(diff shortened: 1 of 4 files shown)


roles/openshift_health_checker/library/rpm_version.py
```diff
@@ -3,6 +3,8 @@
 Ansible module for rpm-based systems determining existing package version information in a host.
 """
 
+from six import string_types
```
rhcarvalho: Ansible bundles `six`, so it is better to use the bundled version than depending on a standalone version being installed in each host:

```python
from ansible.module_utils.six import string_types
```

roles/openshift_health_checker/library/rpm_version.py
```diff
@@ -106,7 +113,9 @@ def _check_pkg_versions(found_pkgs_dict, expected_pkgs_dict):
                 "The following packages were found to be installed with an incorrect version: {}".format('\n'.join([
                     "    \n{}\n    Required version: {}\n    Found versions: {}".format(
                         pkg_name,
-                        pkg["required_version"],
+                        ' or '.join(pkg["required_versions"]) if len(pkg["required_versions"]) < 3
```
rhcarvalho: Suggestion: keep the message consistent by using either "or" or "," to concatenate lists of versions, otherwise it will look weird.

We may also s/version/versions in the message.

Miciah: This code produces "a" for 1 version, "a or b" for 2 versions, and "a, b, c, d, ..., x, y, or z" for 3 or more versions.  I didn't want to use just ", " because the output could be interpreted as a list of required versions.  I didn't want to use " or " because it would look awkward with more than 2 versions.  The only other option I could think of was to change "Required version:" to "Required version: Any of the following:", but that might be even more awkward than "a or b or c or d or ... or x or y or z".  Thoughts?

roles/openshift_health_checker/library/rpm_version.py
Miciah: Thanks! I'll fix that.

roles/openshift_health_checker/library/rpm_version.py
Miciah: Maybe "a | b | c | d | ... | x | y | z"?

sosiouxme: Agreed, while this is clever, I think just a plain ", " will be clear enough.

Miciah:
Updated per review comments:

• Import `string_types` from `ansible.module_utils.six` instead of from plain `six`.

• Delete the logic to insert an "or" in the list of required versions; instead, only use commas.

• Update `roles/openshift_health_checker/test/ovs_version_test.py`.

The unit test passes locally after these changes, and I have manually tested with all combinations of RHOCP 3.4, 3.5, or 3.6 and openvswitch 2.4 or 2.7.

sosiouxme:
Hm, unit tests failing... forget to commit something?

Miciah:
Pushed changes to `roles/openshift_health_checker/test/rpm_version_test.py`.

sosiouxme:
aos-ci-test

sosiouxme:
if epel doesn't sabotage it maybe a [merge] might even pass

sosiouxme:
Well [that](https://ci.openshift.redhat.com/jenkins/job/merge_pull_request_openshift_ansible/843/)'s a new and interesting failure on all the integration tests.

```
		     Play:     Initialize host facts
		     Task:     os_firewall : Ensure firewalld service is not enabled
		     Message:  Failed to get D-Bus connection: Operation not permitted
```

Can't see how it's related to these changes but it must be somehow, unless it's due to a change in the image we're building from...

sosiouxme:
So [this "flake"](https://github.com/openshift/origin/issues/15769) has come up a lot since so that's a bit of a misnomer, but the job has been passing successfully sometimes so let's give [merge] another shot. Didn't see any other indications of failure except these tests so if this repeats itself without any other test problems, I'd say it's safe to merge.

sosiouxme:
@sdodson looks like this is only failing on https://github.com/openshift/origin/issues/15769 like everything else... can you give it a merge?

PR #5043, opened 2017-08-09
Merge inventory-generator with origin-ansible image

juanvallejo:
Begin adding https://github.com/juanvallejo/dynamic-inventory as a new sub-pkg under `images/`.
the `inventory-generator` is based on [this proposal](https://docs.google.com/a/redhat.com/document/d/1J0-fQhAMaFt22_gGixIIWBKQLaFZPUr0AAr1NzPA-Do/edit?usp=sharing)

Trello card for the creation of this image: https://trello.com/c/314Df06O
Trello card for integrating this image with `openshift-ansible`: https://trello.com/c/ARPorrvv

### Goal

The goal behind having this new image is to:
- Create a script that runs in a pod in a cluster. 
- Script collects information about cluster.
- Script uses collected information to create inventory file for that cluster.
- Our health checks playbook (inside the pod in that same cluster) uses that inventory file to run our checks against the cluster. 

### Building

From the root of the `openshift/ansible` directory
```
$ docker build --rm -t openshift/inventory-gen images/inventory-generator
```

### Implementation

The reason why (currently) both `master-config.yaml` and `admin.kubeconfig` are needed is because we use `master-config.yaml` to extract most of the information about hosts that are part of the current deployment, and then use `admin.kubeconfig` to establish a connection with the `oc` binary, to retrieve any additional information about any remaining nodes through the output of `$ oc get nodes -o yaml`.

I am hoping to use this PR and its feedback to mature the contents of this script.

cc @rhcarvalho @brenton @sosiouxme

### DIFF
--- a/images/installer/README_INVENTORY_GENERATOR.md
+++ b/images/installer/README_INVENTORY_GENERATOR.md
@@ -7,7 +7,7 @@ generate an Ansible inventory file from an existing cluster.
 ## Configure
 
 User configuration helps to provide additional details when creating an inventory file.
-The default location of this file is in `root/etc/inventory-generator-config.yaml`. The
+The default location of this file is in `/etc/inventory-generator-config.yaml`. The
 following configuration values are either expected or default to the given values when omitted:
 
 - `master_config_path`:
@@ -26,7 +26,7 @@ following configuration values are either expected or default to the given value
   - specifies a user to "become" on the remote host. Used for privilege escalation.
   - If a non-null value is specified, `ansible_become` is implicitly set to `yes` in the resulting inventory file.
 
-See the supplied sample user configuration file in `root/etc/inventory-generator-config.yaml` for additional optional inventory variables that may be specified.
+See the supplied sample user configuration file in [`root/etc/inventory-generator-config.yaml`](./root/etc/inventory-generator-config.yaml) for additional optional inventory variables that may be specified.
 
 ## Build
 
--- a/images/installer/README_INVENTORY_GENERATOR.md
+++ b/images/installer/README_INVENTORY_GENERATOR.md
@@ -1,0 +1,89 @@
+Dynamic Inventory Generation
+============================
+
+Script within the openshift-ansible image that can dynamically
+generate an Ansible inventory file from an existing cluster.
+
+## Configure
+
+User configuration helps to provide additional details when creating an inventory file.
+The default location of this file is in `root/etc/inventory-generator-config.yaml`. The
+following configuration values are either expected or default to the given values when omitted:
+
+- `openshift_cluster_user`:
+  - username of account capable of listing nodes in a cluster
+  - used for querying the cluster using `oc` to retrieve additional node information.
+
+- `master_config_path`:
+  - specifies where to look for the bind-mounted `master-config.yaml` file in the container
+  - if omitted or a `null` value is given, its value is defaulted to `/opt/app-root/src/master-config.yaml`
+
+- `admin_kubeconfig_path`:
+  - specifies where to look for the bind-mounted `admin.kubeconfig` file in the container
+  - if omitted or a `null` value is given, its value is defaulted to `/opt/app-root/src/.kube/config`
+
+- `ansible_ssh_user`:
+  - specifies the ssh user to be used by Ansible when running the specified `PLAYBOOK_FILE` (see `README_CONTAINER_IMAGE.md` for additional information on this environment variable).
+  - if omitted, its value is defaulted to `root`
+
+- `ansible_become_user`:
+  - specifies a user to "become" on the remote host. Used for privilege escalation.
+  - If a non-null value is specified, `ansible_become` is implicitly set to `yes` in the resulting inventory file.
+
+See the supplied sample user configuration file in `root/etc/inventory-generator-config.yaml` for additional optional inventory variables that may be specified.
+
+## Build
+
+See `README_CONTAINER_IMAGE.md` for information on building this image.
+
+## Run
+
+Given a master node's `master-config.yaml` file, a user configuration file (see "Configure" section), and an `admin.kubeconfig` file, the command below will:
+
+1. Use `oc` to query the host about additional node information (using the supplied `kubeconfig` file and `openshift_cluster_user` value)
+2. Generate an inventory file based on information retrieved from `oc get nodes` and the given `master-config.yaml` file.
+3. run the specified [openshift-ansible](https://github.com/openshift/openshift-ansible) `health.yml` playbook using the generated inventory file from the previous step
+
+```
+docker run -u `id -u` \
+       -v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z,ro \
+       -v /tmp/origin/master/admin.kubeconfig:/opt/app-root/src/.kube/config:Z \
+       -v /tmp/origin/master/master-config.yaml:/opt/app-root/src/master-config.yaml:Z \
+       -e OPTS="-v --become-user root" \
+       -e PLAYBOOK_FILE=playbooks/byo/openshift-checks/health.yml \
+       -e GENERATE_INVENTORY=true \
+       -e USER=`whoami` \
+       openshift/origin-ansible
+
+```
+
+**Note** In the command above, specifying the `GENERATE_INVENTORY` environment variable will automatically generate the inventory file in an expected location.
+An `INVENTORY_FILE` variable (or any other inventory location) does not need to be supplied when generating an inventory.
+
+## Debug
+
+To debug the `generate` script, run the above script interactively
+and manually execute `/usr/local/bin/generate`:
+
+```
+...
+docker run -u `id -u` \
+       -v ...
+       ...
+       -it openshift/origin-ansible /bin/bash
+
+---
+
+bash-4.2$ cd $HOME
+bash-4.2$ ls
+master-config.yaml
+bash-4.2$ /usr/local/bin/generate $HOME/generated_hosts
+bash-4.2$ ls
+generated_hosts  master-config.yaml
+bash-4.2$ less generated_hosts
+...
+```
+
+## Notes
+
+See `README_CONTAINER_IMAGE.md` for additional information about this image.
--- a/images/installer/README_INVENTORY_GENERATOR.md
+++ b/images/installer/README_INVENTORY_GENERATOR.md
@@ -10,10 +10,6 @@ User configuration helps to provide additional details when creating an inventor
 The default location of this file is in `root/etc/inventory-generator-config.yaml`. The
 following configuration values are either expected or default to the given values when omitted:
 
-- `openshift_cluster_user`:
-  - username of account capable of listing nodes in a cluster
-  - used for querying the cluster using `oc` to retrieve additional node information.
-
 - `master_config_path`:
   - specifies where to look for the bind-mounted `master-config.yaml` file in the container
   - if omitted or a `null` value is given, its value is defaulted to `/opt/app-root/src/master-config.yaml`
@@ -40,7 +36,7 @@ See `README_CONTAINER_IMAGE.md` for information on building this image.
 
 Given a master node's `master-config.yaml` file, a user configuration file (see "Configure" section), and an `admin.kubeconfig` file, the command below will:
 
-1. Use `oc` to query the host about additional node information (using the supplied `kubeconfig` file and `openshift_cluster_user` value)
+1. Use `oc` to query the host about additional node information (using the supplied `kubeconfig` file)
 2. Generate an inventory file based on information retrieved from `oc get nodes` and the given `master-config.yaml` file.
 3. run the specified [openshift-ansible](https://github.com/openshift/openshift-ansible) `health.yml` playbook using the generated inventory file from the previous step
 
--- a/images/installer/root/etc/inventory-generator-config.yaml
+++ b/images/installer/root/etc/inventory-generator-config.yaml
@@ -1,0 +1,23 @@
+---
+# meta config
+master_config_path: "/opt/app-root/src/master-config.yaml"
+admin_kubeconfig_path: "/opt/app-root/src/.kube/config"
+
+# default user configuration
+ansible_ssh_user: ec2-user
+ansible_become: "yes"
+ansible_become_user: "root"
+
+# openshift-ansible inventory vars
+openshift_uninstall_images: false
+openshift_install_examples: true
+openshift_deployment_type: origin
+
+openshift_release: 3.6
+openshift_image_tag: v3.6.0
+openshift_hosted_logging_deploy: null  # defaults to "true" if loggingPublicURL is set in master-config.yaml
+openshift_logging_image_version: v3.6.0
+openshift_disable_check: ""
+
+# openshift cluster-viewer info
+openshift_cluster_user: cluster-lister  # name of user or service account able to list nodes in a cluster
--- a/images/installer/root/etc/inventory-generator-config.yaml
+++ b/images/installer/root/etc/inventory-generator-config.yaml
@@ -18,6 +18,3 @@ openshift_image_tag: v3.6.0
 openshift_hosted_logging_deploy: null  # defaults to "true" if loggingPublicURL is set in master-config.yaml
 openshift_logging_image_version: v3.6.0
 openshift_disable_check: ""
-
-# openshift cluster-viewer info
-openshift_cluster_user: cluster-lister  # name of user or service account able to list nodes in a cluster
--- a/images/installer/root/usr/local/bin/generate
+++ b/images/installer/root/usr/local/bin/generate
@@ -82,22 +82,14 @@ class OpenShiftClient:
         """Execute a remote call using `oc`"""
         cmd = [
             self.oc,
+            '--config',
+            self.kubeconfig
         ] + shlex.split(cmd_str)
         try:
             out = subprocess.check_output(list(cmd), stderr=subprocess.STDOUT)
         except subprocess.CalledProcessError as err:
             raise OpenShiftClientError('[rc {}] {}\n{}'.format(err.returncode, ' '.join(err.cmd), err.output))
         return out
-
-    def login(self, host, user):
-        """Login using `oc` to the specified host.
-        Expects an admin.kubeconfig file to have been provided, and that
-        the kubeconfig file has a context stanza for the given user.
-        Although a password is not used, a placeholder is automatically
-        specified in order to prevent this script from "hanging" in the
-        event that no valid user stanza exists in the provided kubeconfig."""
-        call_cmd = 'login {host} -u {u} -p none --config {c}'
-        return self.call(call_cmd.format(host=host, u=user, c=self.kubeconfig))
 
     def whoami(self):
         """Retrieve information about the current user in the given kubeconfig"""
--- a/images/installer/root/usr/local/bin/generate
+++ b/images/installer/root/usr/local/bin/generate
@@ -1,0 +1,407 @@
+#!/bin/env python
+
+"""
+Attempts to read 'master-config.yaml' and extract remote
+host information to dynamically create an inventory file
+in order to run Ansible playbooks against that host.
+"""
+
+import os
+import re
+import shlex
+import shutil
+import subprocess
+import sys
+import yaml
+
+try:
+    HOME = os.environ['HOME']
+except KeyError:
+    print 'A required environment variable "$HOME" has not been set'
+    exit(1)
+
+DEFAULT_USER_CONFIG_PATH = '/etc/inventory-generator-config.yaml'
+DEFAULT_MASTER_CONFIG_PATH = HOME + '/master-config.yaml'
+DEFAULT_ADMIN_KUBECONFIG_PATH = HOME + '/.kube/config'
+
+INVENTORY_FULL_PATH = HOME + '/generated_hosts'
+USE_STDOUT = True
+
+if len(sys.argv) > 1:
+    INVENTORY_FULL_PATH = sys.argv[1]
+    USE_STDOUT = False
+
+
+class OpenShiftClientError(Exception):
+    """Base exception class for OpenShift CLI wrapper"""
+    pass
+
+
+class InvalidHost(Exception):
+    """Base exception class for host creation problems."""
+    pass
+
+
+class InvalidHostGroup(Exception):
+    """Base exception class for host-group creation problems."""
+    pass
+
+
+class OpenShiftClient:
+    oc = None
+    kubeconfig = None
+
+    def __init__(self, kubeconfig=DEFAULT_ADMIN_KUBECONFIG_PATH):
+        """Find and store path to oc binary"""
+        # https://github.com/openshift/openshift-ansible/issues/3410
+        # oc can be in /usr/local/bin in some cases, but that may not
+        # be in $PATH due to ansible/sudo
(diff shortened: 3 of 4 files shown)


rhcarvalho:
@juanvallejo you have some YAML-lint errors: https://travis-ci.org/openshift/openshift-ansible/jobs/262695660#L669

images/inventory-generator/root/usr/local/bin/user_setup
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+set -x
```
rhcarvalho: I'd recommend:

```bash
set -o errexit
set -o nounset
set -o pipefail
```

images/inventory-generator/root/usr/local/bin/run
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
```
rhcarvalho: I'd recommend:

```bash
set -o errexit
set -o nounset
set -o pipefail
```

images/inventory-generator/root/usr/local/bin/run
```diff
@@ -0,0 +1,15 @@
...
+OUTPUT_INVENTORY="generated_hosts"
+
+# generate inventory script
+/usr/local/bin/generate ${OUTPUT_INVENTORY}
+
+if [[ -z "${PLAYBOOK}" ]]; then
+  echo
```
rhcarvalho: Why this?

images/inventory-generator/root/usr/local/bin/run
```diff
@@ -0,0 +1,15 @@
...
+  cat $HOME/${OUTPUT_INVENTORY}
+  exit 0
+fi
+
+# run ansible playbook using generated inventory
+ansible-playbook -vvv -i ${OUTPUT_INVENTORY} ${PLAYBOOK}  --become --become-user root
```
rhcarvalho: Double space before `--become`; missing new line.

Why do we default to `-vvv` (and hard code it)?

Why do we set `--become-user root`? It is the default.

Why do we need `--become`?

images/inventory-generator/root/usr/local/bin/generate
```diff
@@ -0,0 +1,348 @@
+#!/bin/env python
```
rhcarvalho: Note: the "traditional" path/shebang is `#!/usr/bin/env python`, though this should also work on Fedora/RHEL/CentOS.

images/inventory-generator/root/usr/local/bin/entrypoint
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
```
rhcarvalho: Suggestion: set shell options in the script, not in the shebang -- if the script is executed like `/path/to/shell path/to/script`, the `-e` flag is suddenly gone.

Also, I recommend:

```bash
set -o errexit
set -o nounset
set -o pipefail
```

This is what we do most often (see shell scripts in Origin), each option for a good reason.

images/inventory-generator/root/usr/local/bin/entrypoint
```diff
@@ -0,0 +1,17 @@
...
+#
+# This file serves as the main entrypoint to the inventory-generator image.
+#
+# For more information see the documentation:
+#	https://github.com/juanvallejo/dynamic-inventory/blob/master/README.md
```
rhcarvalho: Update URL?

rhcarvalho:
I did a first pass, but did not review everything.

Is it worth having yet another image, what about simply adding a new script to the existing openshift-ansible image?

images/inventory-generator/root/usr/local/bin/user_setup
```diff
@@ -0,0 +1,39 @@
...
+
+# ensure $HOME exists and is accessible by group 0 (we don't know what the runtime UID will be)
+chown ${USER_UID}:0 ${HOME}
+chmod ug+rwx -R ${HOME}
+
+# runtime user will need to be able to self-insert in /etc/passwd
```
juanvallejo: Ah, will word that comment a bit better. This comment justifies the need to make `/etc/passwd` writable, in order to tie a username to the given uid. This is required for ssh to work from within the container

juanvallejo:
@rhcarvalho 
> Is it worth having yet another image, what about simply adding a new script to the existing openshift-ansible image?

Was not sure to what extent we wanted to couple this with our existing image. But if it makes more sense to have this functionality as part of it, I can go ahead and update the PR.

cc @sosiouxme @brenton

sosiouxme:
One image sounds good to me.

images/inventory-generator/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM centos:7
+
+RUN INSTALL_PKGS="openssh-clients wget git" \
+    && yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS \
+    && EPEL_PKGS="PyYAML ansible python2-boto" \
```
sosiouxme: EPEL packages sometimes shadow official packages, which we would usually prefer. There are two basic ways to prevent this - this is one - the other is to define the EPEL repo but leave it inactive and only enable it for the yum transactions needed. Either way you end up separating out the EPEL dependencies.

The other reason to do this is because it identifies which packages are "gaps" in the official packages. In the downstream these tend to require some different treatment, either shipping packages in another channel, getting them via RHSCL, or making them unnecessary somehow. And if things change and the packages are no longer needed... it becomes obvious you can get rid of EPEL.

images/inventory-generator/root/usr/local/bin/user_setup
```diff
@@ -0,0 +1,39 @@
...
+chmod g+rw /etc/passwd 
+chmod g+rw /opt/app-root/src/.kube/config
+chmod g+rw /opt/app-root/src
+
+# ensure that the ansible content is accessible
+chmod -R g+r ${WORK_DIR} /bin
+find ${WORK_DIR} -type d -exec chmod g+x {} +
+
+# This will redirect us to the latest stable release.
+redirect=$(curl -k https://github.com/openshift/origin/releases/latest)
```
sosiouxme: Agreed, not needed

images/inventory-generator/root/usr/local/bin/user_setup
```diff
@@ -0,0 +1,39 @@
...
+# Openshift follows semantic versioning.
+version=$(echo $redirect | grep -o v[0-9]\.[0-9]*\.[0-9]*)
+git_commit=$(git ls-remote https://github.com/openshift/origin refs/tags/$version^{} | awk '{print $1}' | head -c 7)
```
sosiouxme: `oc` client is available on the o-a image anyway

images/inventory-generator/root/usr/local/bin/user_setup
```diff
@@ -0,0 +1,39 @@
...
+
+mv $oc_src_dir/oc /bin
+rm -r $oc_src_dir
+
+# clone openshift-ansible repository
+git clone https://github.com/openshift/openshift-ansible.git ${HOME}/temp
+cp -r ${HOME}/temp/* ${HOME}/
+rm -rf ${HOME}/temp
+
+# no need for this script to remain in the image after running
```
sosiouxme: Just hygiene...

Image squash does happen in OSBS. It's not like we're trying to hide what's in the script, just avoiding leaving stuff lying around that isn't useful.

juanvallejo:
@rhcarvalho @sosiouxme still testing, but went ahead and merged `inventory-generator` with our image. Its readme file has been updated (`README_INVENTORY_GENERATOR.md`).

images/installer/root/usr/local/bin/generate
```diff
@@ -0,0 +1,347 @@
...
+            e = Host("etcd")
+            e.host_alias(common_host_alias)
+            e.host_name(url)
+            etcd_hosts.append(e)
+
+        host_groups["etcd"] = HostGroup(etcd_hosts)
+
+    # open new inventory file for writing
+    try:
+        inv_file_obj = open(HOME + '/' + INVY, 'w+')
```
sosiouxme: should be configurable

images/installer/root/usr/local/bin/generate
```diff
@@ -0,0 +1,347 @@
...
+    for group in host_groups:
+        inv_file_obj.write("{}\n".format(group))
+    inv_file_obj.write("\n")
+
+    inv_file_obj.write("[OSEv3:vars]\n")
+    inv_file_obj.write("ansible_ssh_user={}\n".format(ansible_ssh_user))
+    inv_file_obj.write("ansible_become={}\n".format(ansible_become))
+    inv_file_obj.write("openshift_uninstall_images={}\n".format(str(openshift_uninstall_images)))
+    inv_file_obj.write("openshift_deployment_type={}\n".format(openshift_deployment_type))
+    inv_file_obj.write("openshift_install_examples={}\n".format(str(openshift_install_examples)))
```
sosiouxme: What about all the other stuff from my config file?

```
openshift_release: 3.6
openshift_image_tag: v3.6.0
openshift_hosted_logging_deploy: True
openshift_logging_image_version: v3.6.0
openshift_disable_check: disk_availability,docker_storage,memory_availability
```

etc.... i think we want to be able to generate as close to the original file as possible.

juanvallejo:
@sosiouxme thanks for the feedback; review comments addressed.
Will only support one master in this iteration. In a future update it would
be nice to add support to multiple masters (and host aliases) through the
user config file.

juanvallejo:
[test]

images/installer/root/usr/local/bin/user_setup
```diff
@@ -6,9 +6,15 @@ mkdir -p ${HOME}
...
 chmod ug+rwx ${HOME}
 
+chown ${USER_UID}:0 ${HOME}/.ssh/config
+
 # runtime user will need to be able to self-insert in /etc/passwd
 chmod g+rw /etc/passwd
 
+# make required config dirs writable
+chown ${USER_UID}:0 ${HOME}/.kube/config
+chmod a+rw ${HOME}/.kube/config
```
rhcarvalho: I suspect this file is bind-mounted, and thus this would have an impact on permissions outside of the container...

images/installer/README_INVENTORY_GENERATOR.md
```diff
@@ -0,0 +1,90 @@
...
+
+- `openshift_cluster_user` (required):
+  - username of account capable of listing nodes in a cluster
+  - used for querying the cluster using `oc` to retrieve additional node information.
+
+- `master_config_path` (required):
+  - specifies where to look for the bind-mounted `master-config.yaml` file in the container
+  - if omitted or a `null` value is given, its value is defaulted to `/opt/app-root/src/master-config.yaml`
+
+- `admin_kubeconfig_path` (required):
```
rhcarvalho: How come it is both required and has a default value?!

images/installer/README_INVENTORY_GENERATOR.md
```diff
@@ -0,0 +1,90 @@
...
+
+User configuration helps to provide additional details when creating an inventory file.
+The default location of this file is in `root/etc/inventory-generator-config.yaml`. The
+following configuration values are either expected or default to the given values when omitted:
+
+- `openshift_cluster_user` (required):
+  - username of account capable of listing nodes in a cluster
+  - used for querying the cluster using `oc` to retrieve additional node information.
+
+- `master_config_path` (required):
```
rhcarvalho: How come it is both required and has a default value?!

images/installer/root/usr/local/bin/user_setup
```diff
@@ -6,9 +6,15 @@ mkdir -p ${HOME}
 chown ${USER_UID}:0 ${HOME}
 chmod ug+rwx ${HOME}
 
+chown ${USER_UID}:0 ${HOME}/.ssh/config
+
 # runtime user will need to be able to self-insert in /etc/passwd
 chmod g+rw /etc/passwd
 
+# make required config dirs writable
+chown ${USER_UID}:0 ${HOME}/.kube/config
```
rhcarvalho: These lines ignore `admin_kubeconfig_path`, and I think it will only work if `admin_kubeconfig_path` has its default value?

images/installer/README_INVENTORY_GENERATOR.md
```diff
@@ -0,0 +1,90 @@
...
+
+- `ansible_become_user`:
+  - specifies a user to "become" on the remote host. Used for privilege escalation.
+  - If a non-null value is specified, `ansible_become` is implicitly set to `yes` in the resulting inventory file.
+
+See the supplied sample user configuration file in `root/etc/inventory-generator-config.yaml` for additional optional inventory variables that may be specified.
+
+### Build
+
+`docker build --rm -t openshift/origin-ansible -f images/installer/Dockerfile .`
```
rhcarvalho: I suspect this must be run from the repository root (to give the correct meaning for `.`) and thus we need to note that?

rhcarvalho: Use triple-backticks for making it a block. Single backtick is used for inline code snippets.

````
```
docker build --rm -t openshift/origin-ansible -f images/installer/Dockerfile .

```
````

File: images/installer/README_INVENTORY_GENERATOR.md [hunk 20]
```diff
@@ -0,0 +1,90 @@
+Dynamic Inventory Generation
+============================
+
+Script within the openshift-ansible image that can dynamically
+generate an Ansible inventory file from an existing cluster.
+
+### Configure
```
rhcarvalho: We are missing the second level of headers, we drop from the first to the third. We could s/###/##/g.

images/installer/root/usr/local/bin/generate
```diff
@@ -0,0 +1,393 @@
...
+    HOME = os.environ['HOME']
+except KeyError:
+    print 'A required environment variable "$HOME" has not been set'
+    exit(1)
+
+DEFAULT_USER_CONFIG_PATH = '/etc/inventory-generator-config.yaml'
+DEFAULT_MASTER_CONFIG_PATH = HOME + '/master-config.yaml'
+DEFAULT_ADMIN_KUBECONFIG_PATH = HOME + '/.kube/config'
+
+INVY_FULL_PATH = HOME + '/generated_hosts'
```
rhcarvalho: `INVY`? Sounds funny :-)

images/installer/README_INVENTORY_GENERATOR.md
rhcarvalho: Hmm, on a second thought, the build process is already described elsewhere, the inventory generation does not need to talk about building images?!

images/installer/inventory-generator-config.yaml
```diff
@@ -0,0 +1 @@
+../installer/root/etc/inventory-generator-config.yaml
```
rhcarvalho: Why the symlink?

images/installer/root/etc/inventory-generator-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+# openshift-ansible inventory vars
+openshift_uninstall_images: false
+openshift_install_examples: true
+openshift_deployment_type: origin
+
+openshift_release: 3.6
+openshift_image_tag: v3.6.0
+openshift_hosted_logging_deploy: null  # defaults to "true" if loggingPublicURL is set in master-config.yaml
+openshift_logging_image_version: v3.6.0
+openshift_disable_check: disk_availability,docker_storage,memory_availability
```
rhcarvalho: Won't this suggest people to disable these checks unintentionally?

images/installer/root/etc/inventory-generator-config.yaml
```diff
@@ -0,0 +1,23 @@
+---
+# meta config
+master_config_path: null  # defaults to /opt/app-root/src/master-config.yaml
+admin_kubeconfig_path: null  # defaults to /opt/app-root/src/.kube/config
```
rhcarvalho: How helpful is it to include these? 

Feels like the default paths are repeated in many places.

images/installer/root/etc/inventory-generator-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+ansible_become_user: "root"
+
+# openshift-ansible inventory vars
+openshift_uninstall_images: false
+openshift_install_examples: true
+openshift_deployment_type: origin
+
+openshift_release: 3.6
+openshift_image_tag: v3.6.0
+openshift_hosted_logging_deploy: null  # defaults to "true" if loggingPublicURL is set in master-config.yaml
```
rhcarvalho: Is it helpful to include this?

juanvallejo: I'm thinking with this and `admin_kubeconfig_path` above, I could at least default them to their default values instead of `null`. I think having these present in the default config file gives a better glance of what can be tweaked

images/installer/root/etc/inventory-generator-config.yaml
juanvallejo: Hm, so I had this for debugging purposes and had meant to take them out :) will update

juanvallejo:
@rhcarvalho thanks for the feedback, review comments addressed

juanvallejo:
aos-ci-test

images/installer/root/usr/local/bin/generate
```diff
@@ -0,0 +1,393 @@
...
+
+    master_public_url = y.get("masterPublicURL")
+    if not master_public_url:
+        msg = "'kubernetesMasterConfig.masterPublicURL' missing from '{}'; Unable to connect to master host..."
+        print msg.format(master_config_path)
+        exit(1)
+
+    # connect to remote host using `oc login...` and extract all possible node information
+    oc = OpenShiftClient(admin_kubeconfig_path)
+    oc.login(master_public_url, openshift_cluster_user)
```
sosiouxme: I don't see why you're still logging in here. You have a kubeconfig. It should already be logged in (if necessary) and ready to use. Trying to login will only confuse things. For me it was broken until I removed this line.

images/installer/README_INVENTORY_GENERATOR.md
```diff
@@ -0,0 +1,90 @@
...
+Given a master node's `master-config.yaml` file, a user configuration file (see "Configure" section), and an `admin.kubeconfig` file, the command below will:
+
+1. Use `oc` to query the host about additional node information (using the supplied `kubeconfig` file and `openshift_cluster_user` value)
+2. Generate an inventory file based on information retrieved from `oc get nodes` and the given `master-config.yaml` file.
+3. run the specified [openshift-ansible](https://github.com/openshift/openshift-ansible) `health.yml` playbook using the generated inventory file from the previous step
+
+```
+docker run -u `id -u` \
+       -v $HOME/.ssh/id_rsa:/opt/app-root/src/.ssh/id_rsa:Z,ro \
+       -v /tmp/aws/ssh/config:/opt/app-root/src/.ssh/config:Z,ro \
```
sosiouxme: ... if that's a thing you want to include... but I wouldn't in this example

images/installer/README_INVENTORY_GENERATOR.md
```diff
@@ -0,0 +1,90 @@
...
+       -v /tmp/origin/master/admin.kubeconfig:/opt/app-root/src/.kube/config:Z \
+       -v /tmp/aws/master-config.yaml:/opt/app-root/src/master-config.yaml:Z \
```
sosiouxme: i don't think i'd mention aws... how about `/tmp/origin/master/master-config.yaml`?

images/installer/README_INVENTORY_GENERATOR.md
```diff
@@ -0,0 +1,90 @@
...
+       -e OPTS="-v --become --become-user root" \
```
sosiouxme: the become part should not be necessary?

images/installer/README_INVENTORY_GENERATOR.md
```diff
@@ -0,0 +1,90 @@
...
+
+```
+
+**Note** In the command above, specifying the `GENERATE_INVENTORY` environment variable will automatically generate the inventory file in an expected location.
+An `INVENTORY_FILE` variable (or any other inventory location) does not need to be supplied when generating an inventory.
+
+## Debug
+
+To debug the `generate` script, run the above script interactively
+and manually execute `/usr/local/bin/generate`:
```
sosiouxme: :+1: useful

rhcarvalho:
@juanvallejo do you need some more help here? It seems the last aos-ci-test failed with something strange, looks like a infra flake:

```
Updating status for 4a5f064eceb41f629ce997a3b25ceacb9ebb3caa
Failed to update commit status [HTTP 403]
```

Some good comments from @sosiouxme since then :-)

images/installer/README_INVENTORY_GENERATOR.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Configure
+
+User configuration helps to provide additional details when creating an inventory file.
+The default location of this file is in `root/etc/inventory-generator-config.yaml`. The
+following configuration values are either expected or default to the given values when omitted:
+
+- `openshift_cluster_user` (required):
+  - username of account capable of listing nodes in a cluster
+  - used for querying the cluster using `oc` to retrieve additional node information.
```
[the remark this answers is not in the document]
juanvallejo: hm, at least when I have tried this, it has always (interactively) prompted me for a username even when providing `--config`

images/installer/README_INVENTORY_GENERATOR.md
```diff
@@ -0,0 +1,90 @@
+Dynamic Inventory Generation
+============================
+
+Script within the openshift-ansible image that can dynamically
+generate an Ansible inventory file from an existing cluster.
+
+## Configure
+
+User configuration helps to provide additional details when creating an inventory file.
+The default location of this file is in `root/etc/inventory-generator-config.yaml`. The
```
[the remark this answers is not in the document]
juanvallejo: yeah, although I was thinking users would first edit this file, then it would get bind-mounted with their changes when they built / ran the image

juanvallejo:
@sosiouxme thanks for the feedback, review comments addressed. Let's see how ci tests fare this time

rhcarvalho:
@juanvallejo please rebase and if there are no other changes we can start aos-ci-test and then tag for merge ;)

juanvallejo:
@rhcarvalho thanks, done!

images/installer/README_INVENTORY_GENERATOR.md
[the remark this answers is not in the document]
sosiouxme: Well I guess I should try using `oct` and you should try an AWS cluster install :)

I don't think it should prompt you unless your token has expired. Or it's not storing tokens for some reason.

juanvallejo:
flaked on https://github.com/openshift/origin/issues/8571 re[test]

sosiouxme:
aos-ci-test

sosiouxme:
I found a few things that still needed addressing, so I did that... works for me now.

sosiouxme:
... why are the test statuses still yellow?

sdodson:
aos-ci-test

sdodson:
If that isn't complete tomorrow i'm fine to merge this manually based on the results of https://ci.openshift.redhat.com/jenkins/job/test_pull_request_openshift_ansible/659/

sdodson:
*sigh* I just noticed no one had added a merge tag, let me know if this should be reverted.

sosiouxme:
@sdodson only reason there was no merge tag was because waiting for aos-ci-test. This is quite low risk so I wouldn't worry about reverting.

PR #5044, opened 2017-08-09
Bug 1471322: logging roles based image versions

wozniakjan:
Allowing to specify an image version for each logging component

https://bugzilla.redhat.com/show_bug.cgi?id=1471322

cc @jcantrill , @portante

### DIFF
--- a/roles/openshift_logging/tasks/install_logging.yaml
+++ b/roles/openshift_logging/tasks/install_logging.yaml
@@ -181,8 +181,6 @@
     openshift_logging_kibana_namespace: "{{ openshift_logging_namespace }}"
     openshift_logging_kibana_master_url: "{{ openshift_logging_master_url }}"
     openshift_logging_kibana_master_public_url: "{{ openshift_logging_master_public_url }}"
-    openshift_logging_kibana_image_prefix: "{{ openshift_logging_image_prefix }}"
-    openshift_logging_kibana_image_version: "{{ openshift_logging_image_version }}"
     openshift_logging_kibana_replicas: "{{ openshift_logging_kibana_replica_count }}"
     openshift_logging_kibana_es_host: "{{ openshift_logging_es_host }}"
     openshift_logging_kibana_es_port: "{{ openshift_logging_es_port }}"
@@ -197,8 +195,6 @@
     openshift_logging_kibana_namespace: "{{ openshift_logging_namespace }}"
     openshift_logging_kibana_master_url: "{{ openshift_logging_master_url }}"
     openshift_logging_kibana_master_public_url: "{{ openshift_logging_master_public_url }}"
-    openshift_logging_kibana_image_prefix: "{{ openshift_logging_image_prefix }}"
-    openshift_logging_kibana_image_version: "{{ openshift_logging_image_version }}"
     openshift_logging_kibana_image_pull_secret: "{{ openshift_logging_image_pull_secret }}"
     openshift_logging_kibana_es_host: "{{ openshift_logging_es_ops_host }}"
     openshift_logging_kibana_es_port: "{{ openshift_logging_es_ops_port }}"
@@ -226,8 +222,6 @@
     openshift_logging_curator_es_host: "{{ openshift_logging_es_host }}"
     openshift_logging_curator_es_port: "{{ openshift_logging_es_port }}"
     openshift_logging_curator_master_url: "{{ openshift_logging_master_url }}"
-    openshift_logging_curator_image_prefix: "{{ openshift_logging_image_prefix }}"
-    openshift_logging_curator_image_version: "{{ openshift_logging_image_version }}"
     openshift_logging_curator_image_pull_secret: "{{ openshift_logging_image_pull_secret }}"
 
 - include_role:
@@ -239,8 +233,6 @@
     openshift_logging_curator_es_port: "{{ openshift_logging_es_ops_port }}"
     openshift_logging_curator_namespace: "{{ openshift_logging_namespace }}"
     openshift_logging_curator_master_url: "{{ openshift_logging_master_url }}"
-    openshift_logging_curator_image_prefix: "{{ openshift_logging_image_prefix }}"
-    openshift_logging_curator_image_version: "{{ openshift_logging_image_version }}"
     openshift_logging_curator_image_pull_secret: "{{ openshift_logging_image_pull_secret }}"
     openshift_logging_curator_cpu_limit: "{{ openshift_logging_curator_ops_cpu_limit }}"
     openshift_logging_curator_memory_limit: "{{ openshift_logging_curator_ops_memory_limit }}"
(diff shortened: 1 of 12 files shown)


roles/openshift_logging/tasks/install_logging.yaml
```diff
@@ -182,7 +182,6 @@
     openshift_logging_kibana_master_url: "{{ openshift_logging_master_url }}"
     openshift_logging_kibana_master_public_url: "{{ openshift_logging_master_public_url }}"
     openshift_logging_kibana_image_prefix: "{{ openshift_logging_image_prefix }}"
```
jcantrill: We should be removing the prefixes too if we are going to defer to the ones defined per role

roles/openshift_logging_curator/defaults/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 ### General logging settings
 openshift_logging_curator_image_prefix: "{{ openshift_hosted_logging_deployer_prefix | default('docker.io/openshift/origin-') }}"
-openshift_logging_curator_image_version: "{{ openshift_hosted_logging_deployer_version | default('latest') }}"
+openshift_logging_curator_image_version: "{{ openshift_logging_image_version }}"
```
jcantrill: We lost the default.  We should retain these in case any of the roles are used separately

jcantrill: @sdodson Is it OK for us to stop using the 'hosted' vars?  This PR will need to be cherry-picked back to 3.5 and 3.6 as well so there is that consideration

sdodson: Not yet, we need to mark them deprecated in 3.7 and remove them in 3.8.
https://trello.com/c/XKOGHSkP/460-3-deprecate-openshifthostedloggingmetrics-installer-variables

wozniakjan: Until then, do we have a defined relationship between these variables? Which of `openshift_hosted_logging_deployer_version`, `openshift_logging_image_version`, `openshift_logging_{{role}}_image_version` takes precedence over which?

jcantrill: I think I would just revert these to the hosted vars and allow @sdodson team to change them as appropriate in subsequent releases.

wozniakjan: Even though `openshift_logging_{{role}}_image_version` vars were never used anywhere? These vars were only defined at one point, but not used. 

Without basing the `openshift_logging_{{role}}_image_version` on the common `openshift_logging_image_version`, I am not sure I understand how to draw the relationship there because if:

- we revert these and keep the rest of the PR - then the user has to define `openshift_hosted_logging_deployer_version` to change the version installed for all roles at once, `openshift_logging_image_version` does nothing, but the user can have fine control over each role image version
- we revert these and invert the relationship between `openshift_logging_image_version` and `openshift_logging_{{role}}_image_version` - all versions are controlled by `openshift_logging_image_version`, we are back to what we had before this PR only slightly more convoluted and confusing

If we can't redefine `openshift_logging_{{role}}_image_version` vars, then I would recommend closing this PR without merge and wait for the 3.8

jcantrill: We need to be able to define specific component versions for 3.6 to avoid the upgrade issues we see by customers defining generic versions like 'v3.5'.  IMO, the price to pay for removing 'hosted' vars is better then having to spend time debugging why a log stack is not working because a restarted pod unintentionally pulled in a new version.

jcantrill: I think as preferred practice we should at least leave in the default('latest')

roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -225,8 +225,8 @@
     component: "{{ kibana_component }}"
     logging_component: kibana
     deploy_name: "{{ kibana_name }}"
-    image: "{{ openshift_logging_image_prefix }}logging-kibana:{{ openshift_logging_image_version }}"
-    proxy_image: "{{ openshift_logging_image_prefix }}logging-auth-proxy:{{ openshift_logging_image_version }}"
+    image: "{{ openshift_logging_kibana_image_prefix }}logging-kibana:{{ openshift_logging_kibana_image_version }}"
```
portante: Shouldn't we have different image prefixes and versions for the Kibana and auth-proxy images?

roles/openshift_logging_mux/defaults/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 ### General logging settings
-openshift_logging_mux_image_prefix: "{{ openshift_hosted_logging_deployer_prefix | default('docker.io/openshift/origin-') }}"
-openshift_logging_mux_image_version: "{{ openshift_hosted_logging_deployer_version | default('latest') }}"
+openshift_logging_mux_image_prefix: "{{ openshift_logging_image_prefix | default(__openshift_logging_image_prefix) }}"
```
portante: I realize that you did not make this change here, but why do we have a mux image prefix and version separate from the fluentd image?

roles/openshift_logging_kibana/tasks/main.yaml
wozniakjan: I can see a reasoning in the fact, that both images are closely coupled. Frequently only matching versions of kibana and kibana-proxy work together.

That being said, change can easily be implemented. Discussion about the design decisions for the role based image versions happened in https://github.com/openshift/openshift-ansible/pull/3509. This PR tries to implement propagation of the image based versions to appropriate places as previously designed.

roles/openshift_logging_mux/defaults/main.yml
wozniakjan: Maybe @ewolinetz could provide an explanation, this PR is based on https://github.com/openshift/openshift-ansible/pull/3509

roles/openshift_logging_kibana/tasks/main.yaml
```diff
@@ -225,8 +225,8 @@
...
+    proxy_image: "{{ openshift_logging_kibana_image_prefix }}logging-auth-proxy:{{ openshift_logging_kibana_image_version }}"
```
richm: The proxy image is completely different.  We may need to introduce brand new parameters for the auth-proxy image prefix and version.

roles/openshift_logging_mux/defaults/main.yml
richm: I suppose we could run into a situation where there is some sort of bug in the fluentd image that only affects mux, so you may want to run with a slightly different version of the fluentd image in mux than the one used for fluentd.  If we don't feel that is important, then I'm fine with getting rid of the separate mux prefix/version parameters.

portante: Do you think its worth considering support for running one version of the fluentd image for mux and another for the collectors?  Wouldn't we be concerned about the testing matrix this might introduce?

roles/openshift_logging_kibana/tasks/main.yaml
wozniakjan: added

roles/openshift_logging_kibana/tasks/main.yaml
wozniakjan: done

jcantrill:
aos-ci-test

sdodson:
[merge]

ewolinetz:
Same question here regarding defaults and the `__*` variables

roles/openshift_logging_mux/defaults/main.yml
richm: As I said "If we don't feel that is important, then I'm fine with getting rid of the separate mux prefix/version parameters." Sounds like it is less important than the testing matrix consideration, so I'm ok with having one version of fluentd for both fluentd and mux.

wozniakjan:
it appears to be set correctly to `docker.io/openshift/origin-` when testing the `origin` deployment type and setting neither `openshift_logging_image_prefix`, nor `openshift_logging_*_image_prefix` vars in the inventory fir 3.6 as well.

PR #5045, opened 2017-08-09
WIP add retries

sosiouxme:
Just a quick PR for stylistic feedback - which do you prefer?

### DIFF
--- a/roles/cockpit/tasks/main.yml
+++ b/roles/cockpit/tasks/main.yml
@@ -1,5 +1,6 @@
 ---
 - name: Install cockpit-ws
+  when: not openshift.common.is_containerized | bool
   package: name={{ item }} state=present
   with_items:
     - cockpit-ws
@@ -7,7 +8,9 @@
     - cockpit-bridge
     - cockpit-docker
     - "{{ cockpit_plugins }}"
-  when: not openshift.common.is_containerized | bool
+  register: l_result
+  retries: 4
+  until: l_result.rc | default(-1) == 0 or l_result.results is defined and 'No package matching' in l_result.results.0
 
 - name: Enable cockpit-ws
   systemd:
--- a/roles/dns/tasks/main.yml
+++ b/roles/dns/tasks/main.yml
@@ -1,7 +1,10 @@
 ---
 - name: Install Bind
-  package: name=bind state=present
   when: not openshift.common.is_containerized | bool
+  package: name=bind state=present
+  register: l_result
+  retries: 4
+  until: l_result.rc | default(-1) == 0
 
 - name: Create docker build dir
   file: path=/tmp/dockerbuild state=directory
--- a/roles/docker/tasks/package_docker.yml
+++ b/roles/docker/tasks/package_docker.yml
@@ -32,8 +32,11 @@
 # Make sure Docker is installed, but does not update a running version.
 # Docker upgrades are handled by a separate playbook.
 - name: Install Docker
-  package: name=docker{{ '-' + docker_version if docker_version is defined else '' }} state=present
   when: not openshift.common.is_atomic | bool
+  package: name=docker{{ '-' + docker_version if docker_version is defined else '' }} state=present
+  register: l_result
+  retries: 4
+  until: l_result.rc is defined and l_result.rc == 0
 
 - block:
   # Extend the default Docker service unit file when using iptables-services
--- a/roles/docker/tasks/systemcontainer_docker.yml
+++ b/roles/docker/tasks/systemcontainer_docker.yml
@@ -13,10 +13,12 @@
       in your inventory file.
 
 - name: Ensure container-selinux is installed
+  when: not openshift.common.is_atomic | bool
   package:
     name: container-selinux
     state: present
-  when: not openshift.common.is_atomic | bool
+  register: l_result
+  until: l_result.rc is defined and l_result.rc == 0
 
 # Used to pull and install the system container
 - name: Ensure atomic is installed


roles/cockpit/tasks/main.yml
```diff
@@ -1,13 +1,16 @@
...
   with_items:
     - cockpit-ws
     - cockpit-system
     - cockpit-bridge
     - cockpit-docker
     - "{{ cockpit_plugins }}"
-  when: not openshift.common.is_containerized | bool
+  register: l_result
+  retries: 4
+  until: l_result.rc | default(-1) == 0 or l_result.results is defined and 'No package matching' in l_result.results.0
```
sosiouxme: This one has logic to avoid retries when the package really just isn't there. However this relies on the output from the backing impl and that seems kind of brittle to me. Plus this is a lot to paste everywhere.

roles/dns/tasks/main.yml
```diff
@@ -1,7 +1,10 @@
 ---
 - name: Install Bind
-  package: name=bind state=present
   when: not openshift.common.is_containerized | bool
+  package: name=bind state=present
+  register: l_result
+  retries: 4
+  until: l_result.rc | default(-1) == 0
```
sosiouxme: I think I like this the best. Concise but pretty obvious what's happening.

roles/docker/tasks/package_docker.yml
```diff
@@ -32,8 +32,11 @@
 # Make sure Docker is installed, but does not update a running version.
 # Docker upgrades are handled by a separate playbook.
 - name: Install Docker
-  package: name=docker{{ '-' + docker_version if docker_version is defined else '' }} state=present
   when: not openshift.common.is_atomic | bool
+  package: name=docker{{ '-' + docker_version if docker_version is defined else '' }} state=present
+  register: l_result
+  retries: 4
+  until: l_result.rc is defined and l_result.rc == 0
```
sosiouxme: Slightly more length but slightly clearer.

roles/docker/tasks/systemcontainer_docker.yml
```diff
@@ -13,10 +13,12 @@
       in your inventory file.
 
 - name: Ensure container-selinux is installed
+  when: not openshift.common.is_atomic | bool
   package:
     name: container-selinux
     state: present
-  when: not openshift.common.is_atomic | bool
+  register: l_result
+  until: l_result.rc is defined and l_result.rc == 0
```
sosiouxme: "retries" is actually optional and defaults to 3. I like having it there so it's obvious what's happening but it is an extra line for each of these...

roles/dns/tasks/main.yml
ashcrow: I prefer this. Unless absolutely needed I don't think we should parse the result string.

ashcrow:
It would be great if we could "wrap" these kinds of tasks to use the same retry logic. In Python it would be something like:

```python
# Wrapper
def retry(task):
    def wrap():
        for x in range(0, 3):
            print("Try #{}".format(x+1))
            if task() == True:
                break
    return wrap


# A task
@retry
def my_task():
    print("I did something!")
    return True


# A task
@retry
def another_task():
    print("Something else!")
    return True


# A task
@retry
def fail_task():
    print("I always fail.")
    return False
```

roles/cockpit/tasks/main.yml
sdodson: I think you could use the failed filter

`until: not l_result | failed`

roles/docker/tasks/package_docker.yml
tbielawa: I'm curious if `until` statements follow the same `AND`'ing pattern that `when` statements have. I.e., separable into a multi-line list

roles/cockpit/tasks/main.yml
sosiouxme: ooh, much better. thanks!

sosiouxme:
Yeah; like if you could just add `my_funky_transform` as an attr to the
task and have it invoke a function to add attributes or whatever to the
task.

Probably the least painful way in this case would be to define an action
plugin `openshift_package` (just copy and tweak from the `package` plugin)
with whatever logic we want to wrap it in. That wouldn't solve the problem
for any other kind of task we might want to retry but the explicit
retries/until logic is available for those, and we could make action
plugins for any that are really common.

ashcrow:
> Probably the least painful way in this case would be to define an action plugin `openshift_package` (just copy and tweak from the `package` plugin) with whatever logic we want to wrap it in. That wouldn't solve the problem for any other kind of task we might want to retry but the explicit retries/until logic is available for those, and we could make action plugins for any that are really common.

Good idea! Or possibly via [strategy plugin](https://www.ansible.com/blog/how-to-extend-ansible-through-plugins) as it sees each task. It could react to having an added attribute and enforce a retry system.

ashcrow:
Based on @sosiouxme's refinement of the wrap idea. This ``ActionModule`` would wrap other modules for retrying. Obviously this is just a simple case version but it could be extended for modules that return non-standard success of failure structures.

## action_plugins/retry_test.py
```python
from __future__ import (absolute_import, division, print_function)
__metaclass__ = type

#import time

from ansible.plugins.action import ActionBase


class ActionModule(ActionBase):

    def run(self, tmp=None, task_vars=None):
        if task_vars is None:
            task_vars = dict()
        super(ActionModule, self).run(tmp, task_vars)
        args = self._task.args.copy()

        mod = list(args.keys())[0]
        mod_args = dict(args[mod])
        result = {}
        prev_results = []

        for iteration in range(0, 5):
            result = self._execute_module(
                module_name=mod,
                module_args=mod_args,
                task_vars=task_vars,
                tmp=tmp)
            # It passes if the result is a list, or the rc is not 0
            if isinstance(result.get('results'), list) and result.get('rc', -1) == 0:
                return result

            print(result)
            # Could be configurable
            #time.sleep(2)
            prev_results.append(result)
        # Don't store the result in the result else we fall into a recursion hole.
        result['previous_failures'] = prev_results[:-1]
        return result
```


##  library/retry_test.py
```python
#!/usr/bin/env python
```

## failure_test.yaml
```yaml
- hosts: 127.0.0.1
  gather_facts: no
  tasks:

  - name: test
    retry_test:
      ping:
        data: crash
```

## probably_work_test.yaml
```yaml
- hosts: 127.0.0.1
  gather_facts: no
  tasks:

  - name: test
    retry_test:
      dnf:
        list: ansible
```

sosiouxme:
@ashcrow neat! Now... the problem is that `package` isn't actually a module (it's really an action plugin that wraps other modules) so you can't `_execute_module` it, at least I remember that not working last I tried it...

ashcrow:
@sosiouxme bah, it looks like you are right. If a module is an action it can't be used by a module that is an action 😦 . Here is an example using ``package``:

```
fatal: [127.0.0.1]: FAILED! => {                                                                                                                                                                                     
    "changed": false,                                                                                                                                                                                                
    "failed": true,                                                                                                                                                                                                  
    "module_stderr": "",                               
    "module_stdout": "",                                                                                                                                                                                             
    "previous_failures": [                                                               
        {                                                                                                               
            "_ansible_parsed": false,                                                     
            "failed": true,                          
            "module_stderr": "",                                                                                                                                                                                     
            "module_stdout": "",                          
            "msg": "MODULE FAILURE",                                                                           
            "rc": 0                                                                                      
        },                                                                                                                                                                                                           
        {                                              
            "_ansible_parsed": false,                                                                                                                                                                                
            "failed": true,                                                              
            "module_stderr": "",                                                                                        
            "module_stdout": "",                                                          
            "msg": "MODULE FAILURE",                 
            "rc": 0                                                                                                                                                                                                  
        },                                                
        {                                                                                                      
            "_ansible_parsed": false,                                                                    
            "failed": true,                                                                                                                                                                                          
            "module_stderr": "",                       
            "module_stdout": "",                                                                                                                                                                                     
            "msg": "MODULE FAILURE",                                                     
            "rc": 0                                                                                                     
        },                                                                                
        {                                            
            "_ansible_parsed": false,                                                                                                                                                                                
            "failed": true,                               
            "module_stderr": "",                                                                               
            "module_stdout": "",                                                                         
            "msg": "MODULE FAILURE",                                                                                                                                                                                 
            "rc": 0                                    
        }                                                                                                                                                                                                            
    ],                                                                                   
    "rc": 0                                                                                                             
}                                                                 
```

sosiouxme:
continued in https://github.com/openshift/openshift-ansible/pull/5125

Issue #5050, opened 2017-08-10
Hangs at: TASK: [os_firewall | Start and enable iptables service] when installing from a master


DenverJ:
#### Description
When installing a multi-master cluster using one of the masters to run the play, it hangs at TASK: [os_firewall | Start and enable iptables service]. Note this is the same/similar as #434 and #747 and we have set ansible_connection=local.

The problem here is attempting to run the task on master2 (when master1 is running the play). It is the starting of iptables on master1 that breaks the ssh connection to master2. It works when rerunning the playbook - I believe because iptables is already running. I will shortly raise a PR with one approach to fix this. 

We are running in AWS and would like to launch masters in an ASG and have them bootstrap themselves which is why we do not want to use a separate ansible control host.

##### Version
```
[user@master1 ~]$ ansible --version
ansible 2.3.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
  python version = 2.7.5 (default, Aug  2 2016, 04:20:16) [GCC 4.8.5 20150623 (Red Hat 4.8.5-4)]
[user@master1 ~]$ rpm -q atomic-openshift-utils openshift-ansible
atomic-openshift-utils-3.5.101-1.git.0.0107544.el7.noarch
openshift-ansible-3.5.101-1.git.0.0107544.el7.noarch
```

##### Steps To Reproduce
1. Two servers (target masters) running RHEL 7 with firewalld enabled and iptables disabled
```[user@master1 ~]$ sudo systemctl status iptables
● iptables.service - IPv4 firewall with iptables
   Loaded: loaded (/usr/lib/systemd/system/iptables.service; enabled; vendor preset: disabled)
   Active: inactive (dead)
[user@master1 ~]$ sudo systemctl status firewalld
● firewalld.service - firewalld - dynamic firewall daemon
   Loaded: loaded (/usr/lib/systemd/system/firewalld.service; enabled; vendor preset: enabled)
   Active: active (running) since Thu 2017-08-10 08:32:33 AEST; 7min ago
   ...
```
2. Run ansible-playbook /usr/share/ansible/openshift-ansible/playbooks/byo/config.yml

##### Expected Results
Playbook to run to completion.

##### Observed Results
The task below hangs trying to start iptables on the remote master (master2).
```
TASK [os_firewall : Start and enable iptables service] *******************************************************
changed: [master1]
[ctrl+c]
```

##### Additional Information
```
[...]

PR #5051, opened 2017-08-10
Start iptables on each master in serial

DenverJ:
This PR fixes the problem we are facing in #5050. The start/enable iptables task hangs when connecting to another master when running the playbook from a master without iptables already enabled.

This is one approach. The other approach is a separate task (and possibly a pause) to start/enable iptables first on the current host if the current host is also a master if that is preferred?

### DIFF
--- a/roles/os_firewall/tasks/firewall/iptables.yml
+++ b/roles/os_firewall/tasks/firewall/iptables.yml
@@ -29,6 +29,9 @@
     masked: no
     daemon_reload: yes
   register: result
+  delegate_to: "{{item}}"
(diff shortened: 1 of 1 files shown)


abutcher:
aos-ci-test

abutcher:
We use this approach when creating node certificates so I think it'll be okay especially since there will be few masters to delegate to.

abutcher:
[merge]

jstuever:
I just ran into the same issue and was planning on solving it differently...

+ async: 30
+ poll: 5

PR #5053, opened 2017-08-10
Change default CFME namespace to use reserved openshift- prefix

### LINKED ISSUE #5049: openshift_cfme default namespace name
The default namespace for the openshift_cfme role is set in defaults/main.yml to
```
openshift_cfme_project: cfme
```
We should be using the reserved namespace prefix 'openshift-'.

tbielawa:
As of OCP 3.6 we are advised to prefix reserved namespaces with an
'openshift-' prefix.

Fixes #5049

----

**Note:** I don't think this should be backported into the `release-3.6` branch. It could produce unexpected results and confuse people on an existing 3.6 install if they suddenly re-ran the playbook and all projects and resources were created in a new namespace.

### DIFF
--- a/roles/openshift_cfme/defaults/main.yml
+++ b/roles/openshift_cfme/defaults/main.yml
@@ -1,6 +1,7 @@
 ---
-# Namespace for the CFME project
-openshift_cfme_project: cfme
+# Namespace for the CFME project (Note: changed post-3.6 to use
+# reserved 'openshift-' namespace prefix)
+openshift_cfme_project: openshift-cfme
 # Namespace/project description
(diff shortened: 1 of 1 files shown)


sdodson:
aos-ci-test

sdodson:
If we actually want to deliver this fix in 3.6 lets open a PR against that branch.

tbielawa:
@sdodson per my original post in this PR

> Note: I don't think this should be backported into the release-3.6 branch. It could produce unexpected results and confuse people on an existing 3.6 install if they suddenly re-ran the playbook and all projects and resources were created in a new namespace.

What are you thoughts on that?

sdodson:
sounds good, sorry i missed that.

PR #5072, opened 2017-08-11
Glusterfs firewall updates.

kwoodson:
@jarrpa, with the landing of a recent firewall refactor I wanted to bring the glusterfs role inline with the new pattern.  I don't have a setup to test glusterfs but I believe these firewall changes would work.

The bulk of the PR encapsulates the firewall rules underneath the glusterfs role.  The openshift-glusterfs/config.yml script will then just apply those rules and call the glusterfs role.  The os_firewall role no longer applies the rules and only installs the proper firewall that is desired.

### DIFF
--- a/playbooks/common/openshift-glusterfs/config.yml
+++ b/playbooks/common/openshift-glusterfs/config.yml
@@ -1,40 +1,26 @@
 ---
 - name: Open firewall ports for GlusterFS nodes
   hosts: glusterfs
-  vars:
-    os_firewall_allow:
-    - service: glusterfs_sshd
-      port: "2222/tcp"
-    - service: glusterfs_daemon
-      port: "24007/tcp"
-    - service: glusterfs_management
-      port: "24008/tcp"
-    - service: glusterfs_bricks
-      port: "49152-49251/tcp"
-  roles:
-  - role: os_firewall
+  tasks:
+  - include_role:
+      name: openshift_storage_glusterfs
+      tasks_from: firewall.yml
     when:
     - openshift_storage_glusterfs_is_native | default(True) | bool
 
 - name: Open firewall ports for GlusterFS registry nodes
   hosts: glusterfs_registry
-  vars:
-    os_firewall_allow:
-    - service: glusterfs_sshd
-      port: "2222/tcp"
-    - service: glusterfs_daemon
-      port: "24007/tcp"
-    - service: glusterfs_management
-      port: "24008/tcp"
-    - service: glusterfs_bricks
-      port: "49152-49251/tcp"
-  roles:
-  - role: os_firewall
+  tasks:
+  - include_role:
+      name: openshift_storage_glusterfs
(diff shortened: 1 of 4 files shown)


kwoodson:
aos-ci-test

kwoodson:
aos-ci-test

mtnbikenc:
[merge]

jarrpa:
@kwoodson Do the new variable names follow some new naming scheme? Can they be brought in line with the existing ones?

What's the point of splitting the rules between allow and deny variables instead of just one that with an action field that defaults to allow?

Also what's the point of item.cond?

Finally, couldn't this just be done from the role with a bunch of delegate_to clauses?

kwoodson:
@jarrpa 
>@kwoodson Do the new variable names follow some new naming scheme? Can they be brought in line with the existing ones?
We would like to name variables with `r_<role name>_<var name>` pattern.  The reason for this is that most variables are set with `openshift_` and it is difficult to know where they are getting set.  The `r_` denotes that it is meant for a role.

> What's the point of splitting the rules between allow and deny variables instead of just one that with an action field that defaults to allow?
This pattern is what was defined originally by someone else.  My PR https://github.com/openshift/openshift-ansible/pull/4797 was designed to localize the firewall rules per role and not change the behavior.  We should revisit this in the future.

> Also what's the point of item.cond?
Many of the rules that were defined in roles had a conditions around them.  The firewall.yml allows you to pass in a condition whether to enable the rule or not.

> Finally, couldn't this just be done from the role with a bunch of delegate_to clauses?
Yes.  There are a few different alternative methods to apply the changes.  What you suggest is a possibility.  It appears that the role wasn't written to run on nodes and would need to be updated.

This pull request moves the glusterfs specific firewall rules into the glusterfs role.  This is the minimal change set to bring the glusterfs role in line with the firewall refactor.

jarrpa:
@kwoodson 

>We would like to name variables with r_<role name>_<var name> pattern. The reason for this is that most variables are set with openshift_ and it is difficult to know where they are getting set. The r_ denotes that it is meant for a role.

Got it.

> This pattern is what was defined originally by someone else. My PR #4797 was designed to localize the firewall rules per role and not change the behavior. We should revisit this in the future.

All righty. Who do I talk to about that? :)

> Many of the rules that were defined in roles had a conditions around them. The firewall.yml allows you to pass in a condition whether to enable the rule or not.

Hmm... is this supposed to allow programmatic toggling of firewall rules? Is that actually any easier than just add or removing rules by `service` name?

> Yes. There are a few different alternative methods to apply the changes. What you suggest is a possibility. It appears that the role wasn't written to run on nodes and would need to be updated.

...oh right. I think I ran into a problem where you can't `delegate_to` a node that's not in the `hosts` for the given play... though now I'm not sure. I should test that again. I'll catch you elsewhere about what the other possibilities are.

sdodson:
[merge]

PR #5074, opened 2017-08-11
Fix duplicate evaluate_groups.yml call during install

mtnbikenc:
The way the service_catalog.yml playbook calls evaluate_groups.yml
causes the evaluation to happen twice during a standard byo/config.yml
installation.  Refactoring the service_catalog.yml playbooks to remove
this duplicate.


mtnbikenc:
aos-ci-test

mtnbikenc:
There are two entry point code paths to common/openshift-cluster/service_catalog.yml:
* From byo/openshift-cluster/service_catalog.yml
* From byo/config.yml

PR #5075, opened 2017-08-11
refactoring os_update_latest 

yocum137:


roles/os_update_latest/tasks/main.yml
```diff
@@ -1,3 +1,10 @@
 ---
+- name: Make room for new kernel on /boot
+  command: /usr/bin/package-cleanup --oldkernels --count=2 -y
```
sdodson: This seems to be solving a very specific environmental problem. If the boot partition isn't large enough to accommodate the default number of installed kernels that's either a bug in the OS because the kernel has become too bloated or the provisioning process which has created a /boot filesystem that's unreasonably small.

roles/os_update_latest/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+oul_package_names: "*"
+oul_exclude: 'atomic-openshift* tuned-profiles-atomic-openshift*'
```
sdodson: This needs to be updated to differentiate between `openshift_deployment_type == 'origin' and 'openshift-enterprise'` as the package names are `*origin*` and `*atomic-openshift*` respectively.

roles/os_update_latest/tasks/main.yml
```diff
@@ -1,3 +1,10 @@
...
+
 - name: Update all packages
-  package: name=* state=latest
+  yum:
```
sdodson: Is the switch from `package` to `yum` intentional? We need to support Fedora installs where `dnf` is used and `package` module provides that abstraction.

roles/os_update_latest/tasks/main.yml
```diff
@@ -1,3 +1,10 @@
...
+    name:  "{{ oul_package_names }}"
+    exclude: "{{ oul_exclude }}"
+    state: latest
+  when: os_update is defined
```
sdodson: Why the when condition? Just don't call this role if you don't intend to update packages?

sdodson:
Can you make sure the description of the pull request is updated to reflect what you're changing and why?

roles/os_update_latest/tasks/main.yml
yocum137: The problem really is that old initramfs images get left in /boot.  Even old ones from 3, 4, 5... versions ago where the kernel has been removed.  This just ensures that those are removed.

roles/os_update_latest/tasks/main.yml
yocum137: Nope, not intentional - I'll fix that.

roles/os_update_latest/tasks/main.yml
yocum137: Wrong condition - I'm still an ansible noob.  :D I'll change that to a bool and an appropriate default.

PR #5080, opened 2017-08-14
Add the ability to specify a timeout for node drain operations

sdodson:
A timeout to wait for nodes to drain pods can be specified to ensure that the upgrade continues even if nodes fail to drain pods in the allowed time. The default value of 0 will wait indefinitely allowing the admin to investigate the root cause and ensuring that disruption budgets are respected. In practice the `oc adm drain` command will eventually error out, at least that's what we've seen in our large online clusters, when that happens a second attempt will be made to drain the nodes, if it fails again it will abort the upgrade for that node or for the entire cluster based on your defined `openshift_upgrade_nodes_max_fail_percentage`.

`openshift_upgrade_nodes_drain_timeout=0` is the default and will wait until all pods have been drained successfully
`openshift_upgrade_nodes_drain_timeout=600` would wait for 600s before moving on to the tasks which would forcefully stop pods such as stopping docker, node, and openvswitch.

### DIFF
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
@@ -35,12 +35,18 @@
 
   - name: Drain Node for Kubelet upgrade
     command: >
-      {{ hostvars[groups.oo_first_master.0]['first_master_client_binary'] }} adm drain {{ openshift.node.nodename | lower }} --config={{ openshift.common.config_base }}/master/admin.kubeconfig --force --delete-local-data --ignore-daemonsets
+      {{ hostvars[groups.oo_first_master.0]['first_master_client_binary'] }} adm drain {{ openshift.node.nodename | lower }}
+      --config={{ openshift.common.config_base }}/master/admin.kubeconfig
+      --force --delete-local-data --ignore-daemonsets
+      --timeout={{ openshift_upgrade_nodes_drain_timeout | default(0) }}s
     delegate_to: "{{ groups.oo_first_master.0 }}"
     register: l_upgrade_nodes_drain_result
     until: not (l_upgrade_nodes_drain_result is failed)
-    retries: 60
-    delay: 60
+    retries: "{{ 1 if ( openshift_upgrade_nodes_drain_timeout | default(0) | int ) == 0 else 0 }}"
+    delay: 5
+    failed_when:
+    - l_upgrade_nodes_drain_result is failed
+    - openshift_upgrade_nodes_drain_timeout | default(0) | int == 0
 
   post_tasks:
   - import_role:
--- a/playbooks/common/openshift-cluster/upgrades/upgrade_scale_group.yml
+++ b/playbooks/common/openshift-cluster/upgrades/upgrade_scale_group.yml
@@ -50,11 +50,11 @@
     delegate_to: "{{ groups.oo_first_master.0 }}"
     register: l_upgrade_nodes_drain_result
     until: not (l_upgrade_nodes_drain_result is failed)
-    retries: "{{ 1 if openshift_upgrade_nodes_drain_timeout | default(0) == '0' else 0  | int }}"
+    retries: "{{ 1 if ( openshift_upgrade_nodes_drain_timeout | default(0) | int ) == 0 else 0 }}"
     delay: 5
     failed_when:
     - l_upgrade_nodes_drain_result is failed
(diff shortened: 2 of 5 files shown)


sdodson:
@mwoodson This is what I was working on. I haven't actually tried it so we should try the following on node before we start.

```
oc adm drain --config=/etc/origin/master/admin.kubeconfig \
--force --delete-local-data --ignore-daemonsets \
--timeout=600s
```

sdodson:
PTAL
/assign kwoodson

sdodson:
@sjenning Any thoughts on how specifying a timeout for node drain operations intersects with respecting pod disruption budgets? I assume there's no reasonable way to apply timeouts and still respect disruption budgets, correct?

kwoodson:
@sdodson, This lgtm.  I'll await the '/' until you are ready.

sjenning:
@sdodson yeah force draining and honoring pod distruption budgets are mutually exclusive

https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/

sdodson:
@kwoodson good to go

sdodson:
Rebased and normalized everything to use {{ hostvars[groups.oo_first_master.0]['first_master_client_binary'] }}

playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
```diff
@@ -35,12 +35,18 @@
...
+      {{ hostvars[groups.oo_first_master.0]['first_master_client_binary'] }} adm drain {{ openshift.node.nodename | lower }}
+      --config={{ openshift.common.config_base }}/master/admin.kubeconfig
+      --force --delete-local-data --ignore-daemonsets
+      --timeout={{ openshift_upgrade_nodes_drain_timeout | default(0) }}s
     delegate_to: "{{ groups.oo_first_master.0 }}"
     register: l_upgrade_nodes_drain_result
     until: not (l_upgrade_nodes_drain_result is failed)
-    retries: 60
-    delay: 60
+    retries: "{{ 1 if openshift_upgrade_nodes_drain_timeout | default(0) == '0' else 0  | int }}"
```
michaelgugino: This appears to be comparing a string to int.

playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
```diff
@@ -35,12 +35,18 @@
...
+    delay: 5
+    failed_when:
+    - l_upgrade_nodes_drain_result is failed
+    - openshift_upgrade_nodes_drain_timeout | default(0) == '0'
```
michaelgugino: Comparing string to int.

playbooks/common/openshift-cluster/upgrades/upgrade_nodes.yml
sdodson: Should be fixed now, thanks.

playbooks/common/openshift-cluster/upgrades/upgrade_scale_group.yml
```diff
@@ -50,11 +50,11 @@
     delegate_to: "{{ groups.oo_first_master.0 }}"
     register: l_upgrade_nodes_drain_result
     until: not (l_upgrade_nodes_drain_result is failed)
-    retries: "{{ 1 if openshift_upgrade_nodes_drain_timeout | default(0) == '0' else 0  | int }}"
+    retries: "{{ 1 if (openshift_upgrade_nodes_drain_timeout | default(0) | int == 0 ) else 0 }}"
```
michaelgugino: The ')' should go after 'int', not after the comparison.

playbooks/common/openshift-cluster/upgrades/upgrade_scale_group.yml
```diff
@@ -50,11 +50,11 @@
...
     delay: 5
     failed_when:
     - l_upgrade_nodes_drain_result is failed
-    - openshift_upgrade_nodes_drain_timeout | default(0) == '0'
+    - openshift_upgrade_nodes_drain_timeout | default(0) | int == 0
```
michaelgugino: This line looks good.

playbooks/common/openshift-cluster/upgrades/upgrade_scale_group.yml
sdodson: ok, updated

PR #5085, opened 2017-08-14
[wip] CA serial syncing BZ1469358

tbielawa:
https://bugzilla.redhat.com/show_bug.cgi?id=1469358

```
/etc/origin/master/ca.serial.txt is only on the origin first master. When the first master
is broken or the sequence of masters is changed in inventory file, the node_certificates
may fail for there isn't /etc/origin/master/ca.serial.txt on the other masters.
```

work in progress

### DIFF
--- a/playbooks/common/openshift-cluster/redeploy-certificates/etcd-ca.yml
+++ b/playbooks/common/openshift-cluster/redeploy-certificates/etcd-ca.yml
@@ -125,8 +125,6 @@
 
 - name: Distribute etcd CA to masters
   hosts: oo_masters_to_config
-  vars:
-    openshift_ca_host: "{{ groups.oo_first_master.0 }}"
   tasks:
   - name: Deploy etcd CA
     copy:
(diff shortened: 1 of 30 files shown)


fosuoy:
@tbielawa / @abutcher - I've ran into this issue.
It looks like no new work has been added to this branch since last summer - am I OK to fork / attempt a rebase / merge from current master and submit a PR to this (tbielawa) wip branch or should I be opening a new PR in line with docs here:
https://github.com/openshift/openshift-ansible/blob/master/CONTRIBUTING.md ?

DanyC97:
@abutcher @sdodson @michaelgugino  what is the plan on this PR ? i just hit this issue on 3.7 branch where i lost master 1, even with back of the files taken before i can't do much since i can't scale up any new node

sdodson:
I've closed that bug wontfix. I don't think we're going to get to this.

> @abutcher @sdodson @michaelgugino what is the plan on this PR ? i just hit this issue on 3.7 branch where i lost master 1, even with back of the files taken before i can't do much since i can't scale up any new node

which files did you backup and restore? You'll need /etc/origin/ca.* and /etc/origin/generated-configs

DanyC97:
@sdodson @michaelgugino thx a bunch for quick response.

so if this PR is dead then let's close it @michaelgugino , you agree ?  thx

@sdodson 

> which files did you backup and restore? You'll need /etc/origin/ca.* and /etc/origin/generated-configs

that is exactly what i've done and the next question is: _how do you restore it?_

Because you can't do a scaleup master - ie add a new node, such that the binaries get lay down and then restore the backup files unless i'm missing something.

sdodson:
If you have master-a, master-b, master-c, and you remove master-a you'd restore those files on master-b, ensure that it's the first master in your inventory, and then you should be able to scale up to add master-d.

DanyC97:
@sdodson , so let's close this PR then. thank you again both !

PR #5087, opened 2017-08-14
Use openshift.node.dns_ip as listening address

sdodson:
If someone has specified openshift_dns_ip address then we should use that when configuring dnsmasq so that it aligns with the dnsIP in the config file.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1481366


sdodson:
aos-ci-test

sdodson:
It's just a configuration value we've allowed users to specify but when we added new related feature we failed to ensure that the config value was used in a common manner.

sdodson:
[merge]

PR #5054, opened 2017-08-10
New tuned profile hierarchy.

jmencak:
This PR aims to restructure the current tuned profile hierarchy and also address RH BZ1459146 and BZ1477518.

Old tuned hierarchy:

                              throughput-performance
                                 /        /\     \______________
       _________________________/        /  \                   \
      |                                 /    \            virtual-guest
      |                 _______________/      \       __________/
      |                /     \                 \     /
      |      virtual-{host,guest}     atomic-{host,guest}
      |                      \                 |   |
      |                       \              BZ1477518: tuned-profiles-atomic-openshift-node is not installed on atomic host
      |                        \
      +_____________________    \
                            \    \
    atomic-openshift-node-{host,guest}
    
Proposed tuned hierarchy.    
    
                          throughput-performance
                             /       /\     \______________
       _____________________/       /  \                   \
      |                            /    \            virtual-guest
      |            _______________/      \       __________/
      |           /     \                 \     /
      |  virtual-{host,guest}     atomic-{host,guest}
      +______________  /______________________/
                     \//
                 openshift (all the existing tunings)
                     |
            ---------+---------------
           |                         |
           |                         |
    openshift-node     openshift-control-plane (sched_migration_cost, kernel.sched_wakeup_granularity_ns=4000000, ...)

I'm placing this under "templates" as the "virtual" in openshift profile will need to be parametrized based on the installation environment.  For Atomic Host, it will be "atomic", for anything else "virtual".

Instatiation of the profiles can be done by placing the tuned profiles into /etc/tuned and running:

    tuned-adm profile `tuned-adm recommend`

once the /etc/origin/node/node-config.yaml file is populated with the region key-value pair.  Ansible code for the actual profile instantiation will be provided by a separate PR.

PR #5089, opened 2017-08-15
Setup tuned profiles in /etc/tuned

jmencak:
This PR is to instantiate the new tuned profile hierarchy outlined in https://github.com/openshift/openshift-ansible/pull/5054.

### DIFF
--- a/roles/openshift_node/tasks/main.yml
+++ b/roles/openshift_node/tasks/main.yml
@@ -70,25 +70,15 @@
     - openshift_disable_swap | default(true) | bool
 # End Disable Swap Block
 
-# We have to add tuned-profiles in the same transaction otherwise we run into depsolving
-# problems because the rpms don't pin the version properly. This was fixed in 3.1 packaging.
 - name: Install Node package
   package:
-    name: "{{ openshift.common.service_type }}-node{{ openshift_pkg_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }},tuned-profiles-{{ openshift.common.service_type }}-node{{ openshift_pkg_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }}"
+    name: "{{ openshift.common.service_type }}-node{{ openshift_pkg_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }}"
     state: present
   when: not openshift.common.is_containerized | bool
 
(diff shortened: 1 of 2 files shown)


jmencak:
@sdodson Could you please take a look?  Thanks.

sdodson:
I think you forgot to commit the actual profiles.

roles/openshift_node/tasks/main.yml
```diff
@@ -70,25 +70,15 @@
     - openshift_disable_swap | default(true) | bool
 # End Disable Swap Block
 
-# We have to add tuned-profiles in the same transaction otherwise we run into depsolving
-# problems because the rpms don't pin the version properly. This was fixed in 3.1 packaging.
 - name: Install Node package
   package:
-    name: "{{ openshift.common.service_type }}-node{{ openshift_pkg_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }},tuned-profiles-{{ openshift.common.service_type }}-node{{ openshift_pkg_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }}"
+    name: "{{ openshift.common.service_type }}-node{{ openshift_pkg_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }}"
```
sdodson: The tuned-profiles-%{name}-node package is still a requirement of the node package in origin. We should probably remove that at the same time. If this doesn't get coordinated will the behavior here still take effect or do we need to wait on that to happen?

jmencak: The actual profiles were included in a separate PR: https://github.com/openshift/openshift-ansible/pull/5054, PTAL.  I agree that the rpm dependency should get removed from the origin repo (origin.spec file), but this is something I wanted to do after the fact (PRs 5054 and 5089 are merged).  The correct profiles (new profile hierarchy) will be used even if you leave the old tuned rpm packages as a dependency.  Tested both on RHEL and RHEL Atomic Host 7.4.

sdodson:
aos-ci-test

sdodson:
[merge]

slowbo:
@sdodson I did switch to ansible 2.3.x after I've noticed that the issue was related to the version but wanted to let you guys know that this problem exists :-) And my cluster is running ever since ;-)

Kind regards,
Slowbo

PR #5092, opened 2017-08-15
Remove unneeded byo/openshift-cluster/config.yml

michaelgugino:
This config.yml is only called by byo/config.yml and
a handful of scripts.  Moving the tasks from that
play directly into byo/config.yml make sense and
is easier to follow what is happening in ansible.


sdodson:
I think this runs counter to what @mtnbikenc has proposed in https://docs.google.com/document/d/1dL77N1O6ZqIVTYa2SAnlFpnmoXvWSw-68foNLhoYV70/edit#heading=h.q1ftoequyklg

michaelgugino:
I'm not sure it 'conflicts' with the proposal, but perhaps where the tasks moves to will change.  I just want to get rid of the oc/config.yml as it's only used in one place.

mtnbikenc:
My thought has been that it should actually be the other way around, removing byo/config.yml and only having byo/openshift-cluster/config.yml.  I've not attempted to address this because byo/config.yml is listed in our docs so that would need to be updated.  Additionally, many of our CI scripts make calls directly to byo/config.yml which would also need to be updated.  Looking back through the history of these files, byo/confg.yml has been left as a place holder so these things continue to function rather than deprecating byo/config.yml.

michaelgugino:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1467790

mtnbikenc:
With the breaking apart of the different components, I still think it is better to keep the complete install in `openshift-cluster/`.  The docs should be updated to deprecate `byo/config.yml` and add `byo/openshift-cluster/config.yml`.  This will match with the others we are creating, `byo/openshift-*/config.yml`.  If we take on a later refactor for getting rid of `byo`, I see all of its contents moving down to `playbooks/`.  However, I see that as a later discussion given the current focus on componentization.

PR #5027, opened 2017-08-08
Refactor openshift node role

ingvagabund:
* move disable swap tasks into `disable_swap.yml`
* introduce `r_openshift_node_` prefixed variables and remove all free variables
* autogenerate table of role variables
* move node installation into its own file

Built on top of https://github.com/openshift/openshift-ansible/pull/5025. Once the underlying PR gets merged, I will rebase.

**In depth**
In a process of moving all global variables out of the `openshift_node` role I had to do the following:
1. For each global variable used inside the role create corresponding new variable prefixed with `r_openshift_node_`  string, e.g. for `openshift_node_sdn_mtu` create `r_openshift_node_facts_sdn_mtu`.
2. Given most of the variables has default value attach, I moved the default value of each role variable under `defaults/main.yml`. E.g. the `openshift.node.env_vars | default({})` resulted in `r_openshift_node_env_vars: {}` line.
3. Set all role variables to their corresponding global variables. By the nature of `openshift.node.env_vars | default({})` expression, one can use the following form:
   ```yaml
   include_role:
     name: openshift_node
   vars:
     r_openshift_node_env_vars: "{{ openshift.node.env_vars | default({}) }}"
   ```

   However, the default value (that is set inside `default/main.yml` gets duplicated). For that reason I decided to use the following form instead:

   ```yaml
   set_fact:
      r_openshift_node_env_vars: "{{ openshift.node.env_vars }}"
   when: openshift.node.env_vars is defined

   include_role:
     name: openshift_node
   ```
   It does not look good but the default value is specified only once. Plus, the `set_fact: ... when: ...` syntax is pretty generic, it can be automatically generated (see `playbooks/common/openshift-node/set_optional_node_facts_role_variables.yml`)
4. Verify role variables: a role has a list of role variables that needs to be defined, a list of variables that can be set to a restricted set of value or a list of variables that depends on other variables or their values. 
The checks can be automatically generated (see roles/openshift_node/tasks/pre_checks.yml).

As you can see in this PR there is about 50 role variables that are being set before the `openshift_node` role gets invoked. The fact may be a signal the role needs to be broken out into smaller roles.

PR #5093, opened 2017-08-15
[proposal] Entry-point based roles

ingvagabund:
Let's shape entry points of all roles to conform to a best practice we have discussed on arch meeting.

Applied at:
* Introduce r_openshift_node_facts prefixed role variables: https://github.com/openshift/openshift-ansible/pull/5025
* Refactor openshift node role: https://github.com/openshift/openshift-ansible/pull/5027

Upstream issue wrt. `omit` for role variables: https://github.com/ansible/ansible/issues/13619

### DIFF
--- a/docs/proposals/role-entry-points.md
+++ b/docs/proposals/role-entry-points.md
@@ -5,13 +5,13 @@
 Each role consists of a set of tasks and variables read/set in some of the tasks.
 Some of the tasks can store their results into registered variables,
 e.g. store stdout of a docker pull command.
-Some of the tasks may be run only under certain conditions that are generated
+Some of the tasks may be run only under certain conditions that are known
 outside of a role, e.g. pull images only if a deployment is containerized.
 Some of the tasks are run only if a certain variable is defined,
 e.g. add additional Kubelet arguments into the `node-config.yml`
 if `kubelet_additional_args` inventory variable is defined.
 
-Some roles work with a lot of variables that from a point of view of a role are uninvited citizens.
+Some roles work with a lot of variables that from a point of view of a role make its maintenance more expensive.
 Some of the variables are set inside a role, some of them are set outside of a role.
 The variables outside of a role are hard to track, hard to determine their value or
 a condition under which they are defined. These unknown facts make a role harder
@@ -19,7 +19,7 @@ to maintain, harder to be closed with respect to variables and to become practic
 
 ## Proposal
 
-Make a role that conforms to the following conditions:
+Make sure roles conform to the following conditions:
 * none of the variables read/set inside a role are global
 * all variables inside a role are `r_` (`r` standing for a role) and/or `__` (for private) prefixed
 * all values are propagated to a role through `r_` prefixed variables only
@@ -69,10 +69,14 @@ The file is to be structured in the following way:
 
 #: description="Set to deploy containerized node"
 r_openshift_node_is_containerized: "{{ __openshift_node_is_containerized }}"
-__openshift_node_is_containerized: false
-
 #: description="Set to deploy a node over AH"
 r_openshift_node_is_atomic: "{{ __openshift_node_is_atomic }}"
+```
+
+`vars/main.yml`:
+```yaml
+# default values for role variables
+__openshift_node_is_containerized: false
 __openshift_node_is_atomic: false
 ```
 
@@ -187,6 +191,55 @@ All variables set inside a role has to be prefixed with `__` string.
 The only exception are role variables that are read outside of a role after
 the role finishes.
 
+If a local variable inside a role is read right after it is set (e.g. register variable),
+it can be prefixed just with `__` string. E.g.:
+
+```yaml
+- name: pull node image
+  command: >
+    docker pull {{ r_openshift_node_image_name }}:{{ r_openshift_node_image_tag }}
+  register: __docker_pull_result
+
+- name: check if the a docker image was pulled
+  debug:
+    msg: "Unable to pull a docker image"
+  fail: __docker_pull_result.rc > 0
+```
+
+After the register variable is set, it is considered valid only within a file it was set.
+If a different role reads the same local variable (e.g. `__docker_pull_result`),
+it is the role responsibility to set it before it is read.
+So one can avoid a situation where one role sets a variable and another role
+reads it without initializing.
+
+If a local variable is read among at least two role files (e.g. task files, template files),
+it needs to be fully prefixed, i.e. `__<role>_<var>`. E.g.
+
+`tasks/main.yml`
+```yaml
+- name: "Set template var"
+  set_fact:
+    __<role>_storage_plugins:
+      - nfs
+      - glusterfs
+      - zfs
+...
+- name: create a configuration file
+  template:
+    src: config.yml.j2
+    dest: config.yml
+```
+
+`templates/config.yml.j2`
+```yaml
+...
+storagePlugins:
+{% for plugin in __<role>_storage_plugins -%}
+- {{ plugin }}
+{% endfor %}
+...
+```
+
 ### Role invocation
 
 Based on a role caller some role variables are always defined,
@@ -223,6 +276,33 @@ The `j` denotes a number of role variables that are set to non-role variables
 that does not have to be defined.
 The `n` denotes the total number of all role variables that are set.
 
+### Outstanding questions
+
+The pattern
+
+```yaml
+- include_role:
+    name: <role>
+  vars:
+    r_<role>_<value>: "{{ inventory_<var> [| default(<default>)] }}"
+```
+
+obscures role's variables. To investigate a role's behavior, one has to look
+inside plays to figure out what's actually being passed into roles.
+
+One should use inventory/group_vars to override the default role variables instead
+of setting the role variables with non-role variables withing plays so:
+
+1) we don't rely on a specific play to setup the role 'just right' in order for the role to do it's work
+2) roles are as independent of the play calling it as possible (plays for the most part should just 'call the role')
+3) we can provide a pattern where users can easily modify the behavior of roles and plays in case the plays are not user-modifiable
+
+Possible resolution:
+
+* drop the concept or `r_<role>_<variable>` role variables in favour of `<role>_<variable>` pattern.
+* make sure a role uses only `<role>_` and `__<role>_` prefixed variables
+* consolidate all relevant roles into action-based roles as discussed in https://github.com/openshift/openshift-ansible/pull/5172 so we minimize a need for setting role variables to values from a different role
+
 ### Example
 
 **Before applying the proposal**:
@@ -246,16 +326,20 @@ The `n` denotes the total number of all role variables that are set.
 
 **After:**
 
-`openshift_node/defaults/main.yml`:
+`openshift_node/vars/main.yml`:
 ```yaml
 __omit_str: "__{{ omit.split('__')[1] }}__"
+# default values for role variables
+__openshift_node_image_name: openshift-node
+__openshift_node_image_tag: v3.6.2
+```
 
+`openshift_node/defaults/main.yml`:
+```yaml
 # description="OpenShift node image name"
 r_openshift_node_image_name: "{{ __openshift_node_image_name }}"
-__openshift_node_image_name: openshift-node
 # description="OpenShift node image tag"
 r_openshift_node_image_tag: "{{ __openshift_node_image_tag }}"
-__openshift_node_image_tag: v3.6.2
 ...
 ```
 
--- a/docs/proposals/role-entry-points.md
+++ b/docs/proposals/role-entry-points.md
@@ -1,38 +1,37 @@
 # Entry-point based roles
 
-## We have a problem
+## Variables inside a role are too open
 
 Each role consists of a set of tasks and variables read/set in some of the tasks.
 Some of the tasks can store their results into registered variables,
 e.g. store stdout of a docker pull command.
-Some of the tasks may be run only under certain condition that are generated
+Some of the tasks may be run only under certain conditions that are generated
 outside of a role, e.g. pull images only if a deployment is containerized.
 Some of the tasks are run only if a certain variable is defined,
 e.g. add additional Kubelet arguments into the `node-config.yml`
 if `kubelet_additional_args` inventory variable is defined.
 
-Some roles work with a lot of variables that from a point of a role are uninvited citizens.
+Some roles work with a lot of variables that from a point of view of a role are uninvited citizens.
 Some of the variables are set inside a role, some of them are set outside of a role.
 The variables outside of a role are hard to track, hard to determine their value or
-a condition under which the variables are defined. These unknown facts make a role harder
-to maintain, harder to make a role closed with respect to variables and to make
-a role practically modular.
+a condition under which they are defined. These unknown facts make a role harder
+to maintain, harder to be closed with respect to variables and to become practically modular.
 
 ## Proposal
 
 Make a role that conforms to the following conditions:
 * none of the variables read/set inside a role are global
-* all variables inside a role are `r_` (`r` standing for a role) and/or `l_`(`l` standing for a local) prefixed
+* all variables inside a role are `r_` (`r` standing for a role) and/or `__` (for private) prefixed
 * all values are propagated to a role through `r_` prefixed variables only
 * all values that are set inside a role are read outside of a role through `r_` prefixed variables only
 * all available `r_` prefixed variables are documented and checked before a main role task is run
 * a table of `r_` prefixed variables in README.md is automatically generated
 
-### Terminilogy
+### Terminology
 
 **Role variables**: All `r_` prefixed variables read/set inside a role
 
-**Local variables**: All `l_` prefixed variables set inside a role
+**Local/Private variables**: All `__` prefixed variables set inside a role
 
 ### Role properties
 
@@ -49,7 +48,7 @@ Well defined API allows automatically generated documentation and variable check
 **Free of outer variables**: only variables consisting the role API or defined locally are allowed.
 Any other variable (e.g. inventory variable or a variable set in a play) are forbidden.
 A role free of any outer variable:
-* can be shared across different Ansible solutions (improves modularity)
+* can be shared across different Ansible solutions (increases modularity)
 * is easier to test as only the role variables need to be set (improves role based testing)
 * independent of variable name changes outside of a role (improves role independence)
 
@@ -69,10 +68,17 @@ The file is to be structured in the following way:
 #: name=r_openshift_node_system_images_registry, description="System image registry to pull images from", required_when=["r_openshift_node_is_containerized | bool", "r_openshift_node_is_node_system_container | bool"], required_when=["r_openshift_node_is_containerized | bool", "r_openshift_node_is_openvswitch_system_container | bool", "r_openshift_node_use_openshift_sdn | bool"]
 
 #: description="Set to deploy containerized node"
-r_openshift_node_is_containerized: false
+r_openshift_node_is_containerized: "{{ __openshift_node_is_containerized }}"
+__openshift_node_is_containerized: false
+
 #: description="Set to deploy a node over AH"
-r_openshift_node_is_atomic: false
+r_openshift_node_is_atomic: "{{ __openshift_node_is_atomic }}"
+__openshift_node_is_atomic: false
 ```
+
+Notice the role variable pairs, e.g. (`r_openshift_node_is_containerized`, `__openshift_node_is_containerized`).
+The former is a role variable available in role API, the latter is a default for the former role variable.
+If the former role variable is omitted (defined later), the default one overwrites the former one.
 
 The role variables can be broken down intro three categories:
 * required role variables but undefined: role variables that must be always defined but without a default value
@@ -88,7 +94,68 @@ All three categories can be annotated with the following keywords:
 
 A table of role variables and a sequence of pre-checks are generated from the specified annotations.
 
-All variables under `var/main.yml` are considered local variables.
+#### Omitted role variables
+
+Once an Ansible variable is defined, it can not be undefined.
+That can be an issue if a play calls a role that has role variables
+set to inventory variables that are optional, i.e. does not have to be defined.
+
+Assume a role that is invoked in the following way:
+```yaml
+- include_role:
+    name: <role>
+  vars:
+    r_<role>_variable: "{{ inventory_variable }}"
+```
+
+If the `inventory_variable` is defined, the role is invoked successfully.
+If the variable is not defined, we either:
+
+* have to check and set the `r_<role>_variable` conditionally, or
+* we need to use the `default` filter.
+
+The former case can lead to a sequence of `set_fact: ... when: ... is defined` tasks.
+Additionally, the role variables are set separately from the role invocation.
+
+The latter case can lead to a sequence of `r_<role>_<variable>: "{{ <inventory_variable> | default(<default>) }}"` lines.
+To apply the `default(<default>)` filter one needs to know a default value of
+corresponding role variable (which leads to value duplication that is already set in the role).
+Thus, to eliminate default value duplication and to keep role variables
+definition in the role invocation expression, one can use the `default(omit)`
+filter form:
+
+```yaml
+- include_role:
+    name: <role>
+  vars:
+    r_<role>_<variable>: "{{ inventory_<variable> | default(omit) }}"
+    r_<role>_<variable>: "{{ inventory_<variable> | default(omit) }}"
+    ...
+```
+
+Currently, the `omit` keyword is correctly interpreted by Ansible only
+when used with module arguments (https://github.com/ansible/ansible/issues/13619).
+For that reason the role has to advocate the following pattern to properly
+interpret "omitted" role variables:
+
+```yaml
+- name: "Set r_openshift_node_is_containerized to default if omited"
+  set_fact:
+    r_openshift_node_is_containerized: "{{ __openshift_node_is_containerized }}"
+  when:
+  - __omit_str in r_openshift_node_is_containerized | string
+```
+
+**Explanation**: The `r_openshift_node_is_containerized` is set to `__openshift_node_is_containerized`
+only when the role variable is set to `omit`. Internally the `omit` corresponds to a string
+accepted by `__omit_place_holder__[\da-e]{40}` regex. Thus, if the role variable
+value contains the `__omit_place_holder__` string (stored inside the `__omit_str` variable),
+the role variable is considered to be "undefined" and thus set to its default value.
+This way, the role variables can be set to non-role variables that are optionally defined
+in a compact way.
+
+Since the `__omit_str` can be used on multiple places (e.g. tasks, templates),
+it is defined in `defaults/main.yml` as `__omit_str: "__{{ omit.split('__')[1] }}__"`.
 
 #### We should check all relevant role variables are properly set
 
@@ -114,68 +181,9 @@ The following checks can be generated:
 
 All the checks can be stored into `tasks/pre_checks.yml` file and included in the main task file.
 
-#### Propagation of values to role dependencies
-
-A role can depend on another role that has its own set of role variables. There are two ways how to propagate values to role dependencies. First approach requires a caller has a knowledge of all role dependency variables:
-
-```yaml
-roles:
-- role: etcd
-  r_etcd_client_port: 2379
-  r_etcd_peer_port: 2380
-  r_etcd_common_etcd_runtime: docker
-  r_etcd_common_embedded_etcd: false
-```
-
-Here the `etcd` role depends on `etcd_common` role.
-The second approach is to duplicate all role dependency variables:
-
-```yaml
-roles:
-- role: etcd
-  r_etcd_client_port: 2379
-  r_etcd_peer_port: 2380
-  r_etcd_etcd_runtime: docker
-  r_etcd_embedded_etcd: false
-```
-
-and write all the duplicated role variables into their corresponding
-dependency variables in `etcd/meta/main.yml`:
-
-```yaml
-dependencies:
-- role: etcd_common
-  r_etcd_common_etcd_runtime: "{{ r_etcd_etcd_runtime }}"
-  r_etcd_common_embedded_etcd: "{{ r_etcd_embedded_etcd }}"
-```
-
-One can use `include_role` instead of `roles`:
-
-```yaml
-include_role:
-  name: etcd
-vars:
-  r_etcd_client_port: 2379
-  r_etcd_peer_port: 2380
-  r_etcd_etcd_runtime: docker
-  r_etcd_embedded_etcd: false
-```
-
-From the point of view of the caller the number of role variables
-to be set does not change in both approaches.
-
-**Advantages** of the second approach:
-* role dependencies and role variables of dependencies can change transparently
-* caller does not need any knowledge of role variables of dependencies
-
-**Disadvantages**:
-* role variables of dependencies has to be duplicated
-* one must not forget to wire all duplicated variables inside `meta/main.yml` file (this step can be automatically generated)
-* optional undefined variables can not be set this way (TODO(jchaloup): or can they? is that even needed, maybe all roles with optional undefined role variables should not be dependencies of another role)
-
 ### Local variables
 
-All variables set inside a role has to be prefixed with `l_` string.
+All variables set inside a role has to be prefixed with `__` string.
 The only exception are role variables that are read outside of a role after
 the role finishes.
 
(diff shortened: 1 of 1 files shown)


docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+The variables outside of a role are hard to track, hard to determine their value or
+a condition under which the variables are defined. These unknown facts make a role harder
+to maintain, harder to make a role closed with respect to variables and to make
+a role practically modular.
+
+## Proposal
+
+Make a role that conforms to the following conditions:
+* none of the variables read/set inside a role are global
+* all variables inside a role are `r_` (`r` standing for a role) and/or `l_`(`l` standing for a local) prefixed
```
michaelgugino: I don't think I like this approach of prefixing variable names.  That might become inflexible and at some point may require a breaking change to change the name of a variable from r_ to l_

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+It means each such variable must have the following information attached:
+* name
+* description
+* default value (if the variable is defined)
+* set of acceptable values
+* if it is required or if it is optional
+
+Well defined API allows automatically generated documentation and variable checks.
+
+**Free of outer variables**: only variables consisting the role API or defined locally are allowed.
```
michaelgugino: I generally agree with this approach.

For example, say the value openshift_version: 3.6 has been set in inventory.  A role should say something like:
my_openshift_version: {{ openshift_version | default(3.6) }}

Or, even better, my_openshift_version: 3.6

And, you override my_openshift_version in inventory/group_vars/extra vars.

docs/proposals/role-entry-points.md
mtnbikenc: `l_` is not really local.  Once a var is set it is available for any subsequent tasks/roles/plays.  There could be unintended collisions of variables because there is no such thing as local vars.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+A role free of any outer variable:
+* can be shared across different Ansible solutions (improves modularity)
+* is easier to test as only the role variables need to be set (improves role based testing)
+* independent of variable name changes outside of a role (improves role independence)
+
+### Role variables
+
+#### Where the role variables live (with annotations?)
+
+All role variables live under `default/main.yml` file.
```
michaelgugino: Yes, everything should live under defaults.  However, some things might not be a good fit for this, especially things like "with_first_found" for having groups of variables across different distros, for example.  Those special cases should be limited and justified, however.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+a condition under which the variables are defined. These unknown facts make a role harder
+to maintain, harder to make a role closed with respect to variables and to make
+a role practically modular.
+
+## Proposal
+
+Make a role that conforms to the following conditions:
+* none of the variables read/set inside a role are global
+* all variables inside a role are `r_` (`r` standing for a role) and/or `l_`(`l` standing for a local) prefixed
+* all values are propagated to a role through `r_` prefixed variables only
```
mtnbikenc: I had already documented the role variable naming in #4253.  I'd like to get consensus from the team on standardization and have it documented in one place.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+* all values that are set inside a role are read outside of a role through `r_` prefixed variables only
+* all available `r_` prefixed variables are documented and checked before a main role task is run
+* a table of `r_` prefixed variables in README.md is automatically generated
+
+### Terminilogy
+
+**Role variables**: All `r_` prefixed variables read/set inside a role
+
+**Local variables**: All `l_` prefixed variables set inside a role
```
mtnbikenc: I think all variables set inside roles should use the role prefix.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+#### Propagation of values to role dependencies
+
+A role can depend on another role that has its own set of role variables. There are two ways how to propagate values to role dependencies. First approach requires a caller has a knowledge of all role dependency variables:
+
+```yaml
+roles:
+- role: etcd
+  r_etcd_client_port: 2379
+  r_etcd_peer_port: 2380
+  r_etcd_common_etcd_runtime: docker
```
mtnbikenc: I don't like passing role variables for a role dependency.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+  r_etcd_common_embedded_etcd: false
+```
+
+Here the `etcd` role depends on `etcd_common` role.
+The second approach is to duplicate all role dependency variables:
```
mtnbikenc: I also don't like duplicating variables just to pass them to a role dependency.  I'd rather see the roles refactored to allow the roles to function independently and be called from the play than putting that dependency logic in meta/main.yml dependencies.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+* description: short and accurate description of a role variable
+* required: set to `false` if not specified, if set to `true` the role variable is required (forced by generated pre-checks tasks)
+* choices: a list of acceptable values of a role variables
+* required_when: a list of conjuncts under which a role variable must be defined
+
+A table of role variables and a sequence of pre-checks are generated from the specified annotations.
+
+All variables under `var/main.yml` are considered local variables.
+
+#### We should check all relevant role variables are properly set
```
michaelgugino: I disagree with this.  There's already a ton of checks in this project, there's no need to check every variable before a role is run.  We should be very much avoiding this situation by properly defining default values wherever possible.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+* caller does not need any knowledge of role variables of dependencies
+
+**Disadvantages**:
+* role variables of dependencies has to be duplicated
+* one must not forget to wire all duplicated variables inside `meta/main.yml` file (this step can be automatically generated)
+* optional undefined variables can not be set this way (TODO(jchaloup): or can they? is that even needed, maybe all roles with optional undefined role variables should not be dependencies of another role)
+
+### Local variables
+
+All variables set inside a role has to be prefixed with `l_` string.
```
mtnbikenc: `l_` is still not clear because it could end up being scoped from anywhere previous in the Ansible execution.  If we want something to be local, the best we could do is use vars/main.yml in the role. See http://docs.ansible.com/ansible/latest/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+```yaml
+- role: openshift_node
+  r_openshift_node_is_containerized: "{{ openshift.common.is_containerized }}"
+```
+
+Some role variables like `r_openshift_node_kubelet_args` are optionally defined
+and can not be set unless their corresponding inventory (or other playbook) variables
+are set. So the proper way to set them is:
+```yaml
+- set_task:
```
mtnbikenc: `set_fact`

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+  when:
+  - r_openshift_node_deployment_type is defined
+  - r_openshift_node_deployment_type not in ['origin', 'openshift-enterprise']
+```
+
+All the checks can be stored into `tasks/pre_checks.yml` file and included in the main task file.
+
+#### Propagation of values to role dependencies
+
+A role can depend on another role that has its own set of role variables. There are two ways how to propagate values to role dependencies. First approach requires a caller has a knowledge of all role dependency variables:
```
michaelgugino: Not a fan of this at all; it's confusing to have a role that utilizes values defined elsewhere.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+
+- include_role:
+    name: openshift_node
+  vars:
+    r_openshift_node_is_containerized: "{{ openshift.common.is_containerized }}"
+```
+
+Thus, in general a role invocation corresponds to the following pattern:
+```yaml
+- set_task:
```
mtnbikenc: `set_fact`

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Proposal
+
+Make a role that conforms to the following conditions:
+* none of the variables read/set inside a role are global
+* all variables inside a role are `r_` (`r` standing for a role) and/or `l_`(`l` standing for a local) prefixed
+* all values are propagated to a role through `r_` prefixed variables only
+* all values that are set inside a role are read outside of a role through `r_` prefixed variables only
+* all available `r_` prefixed variables are documented and checked before a main role task is run
+* a table of `r_` prefixed variables in README.md is automatically generated
```
mtnbikenc: I like automatically generated documentation.

ashcrow: 👍

docs/proposals/role-entry-points.md
ashcrow: @mtnbikenc do you see them being aliased to ``l_`` when passing them along to another role?

docs/proposals/role-entry-points.md
michaelgugino: This is the more appropriate approach

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+roles:
+- role: etcd
+  r_etcd_client_port: 2379
+  r_etcd_peer_port: 2380
+  r_etcd_etcd_runtime: docker
+  r_etcd_embedded_etcd: false
+```
+
+and write all the duplicated role variables into their corresponding
+dependency variables in `etcd/meta/main.yml`:
```
michaelgugino: Definitely don't do this.

docs/proposals/role-entry-points.md
ingvagabund: The concept of a local variable is more like to state "this variable is purely used only within a role and I as a role maintainer has no responsibility over using it outside of my role". So if there are two roles that sets the same `l_` prefixed variable, it is ok as one role is not aware of local variables of another role and each sets its `l_` prefixed variable before it reads it. The most common use case is to use the `l_` prefixed variables with `register: `.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+Some roles work with a lot of variables that from a point of a role are uninvited citizens.
+Some of the variables are set inside a role, some of them are set outside of a role.
+The variables outside of a role are hard to track, hard to determine their value or
+a condition under which the variables are defined. These unknown facts make a role harder
+to maintain, harder to make a role closed with respect to variables and to make
+a role practically modular.
+
+## Proposal
+
+Make a role that conforms to the following conditions:
```
kwoodson: I'd like to see us move in the direction of roles being more stand alone.  In general, defining these are great and I see this as a first step.

docs/proposals/role-entry-points.md
kwoodson: I have never really like l_ variables but I do like how they tell me that vars have been created locally.  It takes me time to process the l because it looks like any of the following chars, 1, l, I, i, L, | and I think it's generally bad practice to call variables by those names.

Could we just use the r_ prefix and still get the same benefit?  This would denote that the variable was passed in, created within the role, or is a role default.  At least this would standardize on variables.

docs/proposals/role-entry-points.md
kwoodson: :+1:

docs/proposals/role-entry-points.md
kwoodson: I agree that the we want to take this in from the inventory.  I'd prefer to not pass in a default since it becomes an artifact the moment it gets merged.  I'd prefer we pass it in and if its not set then explode.  Being explicit is better than implicit for a couple of reasons.  
- If I don't know about a variable then I'm guaranteed to get something installed but it might not be what I desire.  This is sometimes worse than doing anything at all.  I have installed cluster and received example.com as my domain in my config files.  This becomes a debugging nightmare and a reinstall is the only solution.
- fail fast (fail: when a var isn't set) is a great pattern for the roles.  Its a form of defensive programming.  Role variable checking is a good way to ensure that something doesn't get half way configured and then break.  Especially in production clusters.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,264 @@
...
+* name
+* description
+* default value (if the variable is defined)
+* set of acceptable values
+* if it is required or if it is optional
+
+Well defined API allows automatically generated documentation and variable checks.
+
+**Free of outer variables**: only variables consisting the role API or defined locally are allowed.
+Any other variable (e.g. inventory variable or a variable set in a play) are forbidden.
```
kwoodson: This is great.  This stops bleeding into other roles and allows them to be stand alone tools.

docs/proposals/role-entry-points.md
kwoodson: Where defaults are allowed then place defaults values for variables.  Where variables are required then I like testing for those.  I'd prefer a failure message rather than dysfunctional cluster.

docs/proposals/role-entry-points.md
kwoodson: > it's confusing to have a role that utilizes values defined elsewhere.

@michaelgugino, could you clarify your statement?  What specific values are defined elsewhere?  Just so I understand.  Thanks.

docs/proposals/role-entry-points.md
kwoodson: I'm going to agree with @mtnbikenc.  If the role needs to call other roles then the role probably needs to be considered for a refactor.  I'd like it if the orchestration of roles occurs at the playbook level and not through role dependencies.  If we had roles where a calls b and b calls c, then we need to pass data through a to arrive at c.  `a -> b -> c`.  This tightly couples these roles together and obfuscates the role calls with their variables.

I would think that the calling playbook could just do the following:
```
include_role:
  name: a
vars:
  r_a_foo: bar

include_role:
  name: b
vars:
  r_b_foo: czar

include_role:
  name: c
vars:
 r_c_foo: far
```

This is more verbose but now allows us to be a bit more modular.  This approach is very flexible and allows for the ability to call a role when I need to call the role in a tool box fashion.

ashcrow:
I second @kwoodson's statement. Even though there is a lot of different points of view on how to implement this you're doing a great job!

docs/proposals/role-entry-points.md
ingvagabund: If all variables inside a role are prefixed with `r_` it will be hard to distinguish between variables that are meant to be local only and those that are meant to be read outside of a role. Again, we can make a list of all role variables that a role "produces" and document each one properly. However, nothing will stop a user from consuming the remaining ones. Of course, we can state" if you use any of the role variables that we have not declared as `output` variables, it is at your own risk chief". The `l_` prefix is here for that very reason. Of course, we don't have to use the `l_` prefix and can use `__some_variable_not_meant_to_be_used_outside_of_a_role` style instead. I would like to distinguish the concept of variables meant to be used within/outside a role on a syntactic level.

ingvagabund: If you have to change `r_` to `l_` you have used the prefix incorrectly. The `l_` prefix is really meant to be used for variables that are purely local like registered variables or gate variables`. If one needs to share registered variables outside of a role, the one should use:
```yaml
set_fact:
  r_something_finished: "{{ l_something_finished[.stdout] }}"
```

docs/proposals/role-entry-points.md
ingvagabund: We can defined what a role variable is in #4253 and once merged, I can reference the document in this PR.

docs/proposals/role-entry-points.md
ingvagabund: I don't mind that as long as we make clear which role variables are meant to be consumed outside of a role and which are not. We can always go with `r_rolename_l_localvariablename`. In any case the question is how are we going to distinguish between both categories on a syntactical level.

docs/proposals/role-entry-points.md
ingvagabund: Some non-role variables has to be set everytime, like the `openshift_version`. So either it is set inside an inventory or it is set by `openshift_version` role. Other non-role variables can be undefined unless set in an inventory file like `openshift_node_kubelet_args`. If the variable is not set, I should not set corresponding `r_openshift_node_kubelet_args` to anything when invoking a role. If the variable is set, I should set `r_openshift_node_kubelet_args: "{{ openshift_node_kubelet_args }}"`

The question is how can I do that in the following form for both cases:
```yaml
- include_role:
    name: openshift_node
  vars:
    r_openshift_node_kubelet_args: "{{ openshift_node_kubelet_args }}"
```

without using the `default({})` construct? I.e.

```yaml
- include_role:
    name: openshift_node
  vars:
    r_openshift_node_kubelet_args: "{{ openshift_node_kubelet_args | default({}) }}"
```

I see only one way to do it and it is:
```yaml
- set_fact:
    r_openshift_node_kubelet_args: "{{ openshift_node_kubelet_args }}"
  when: openshift_node_kubelet_args is defined

- include_role:
    name: openshift_node
  vars:
    ...
```

Or, we need to make sure that all inventory variables are always set to some default value even if they are not specified in an inventory file.

docs/proposals/role-entry-points.md
ingvagabund: I intentionally choose the "relevant" keyword. At the moment I can just checking:
- if all required role variables are defined
- if all conditionally required variables are defined
- if all role variables with a list of accepted values are set to one of the accepted values

docs/proposals/role-entry-points.md
ingvagabund: As long as a role does not consume role variables of its dependency, I agree with your proposal.

docs/proposals/role-entry-points.md
ingvagabund: The `openshift_node` has a list of role variables that are set to the following non-role variables than do not have to be defined:
* openshift.node.dns_ip
* openshift.common.sdn_network_plugin_name
* openshift.common.ip
* openshift_node_min_tls_version
* openshift_node_cipher_suites
* openshift_disable_swap
* openshift_cloudprovider_aws_access_key
* openshift_cloudprovider_aws_secret_key
* openshift_cloudprovider_kind
* openshift_node_kubelet_args
* openshift.common.no_proxy
* openshift.node.debug_level
* openshift.common.https_proxy
* openshift.common.http_proxy
* openshift.common.system_images_registry
* openshift.node.env_vars
* openshift.docker.gte_1_10

All of them can be set to some default value. The question is which default value is the proper one. The only thing I am afraid of is duplication (and syncing) of default values among role variables and their corresponding non-role variables. On the other hand it is completely fine for a role variable to have a different value than its corresponding inventory variable. Since a default role variable should be always chosen from a point of view of a role and default inventory variable from a point of view of entire deployment.

So, if I can assume all of the variables are always going to be set to some default value, I no longer have to use the `set_fact: ...when: var is defined` sequence for all the above mentioned variables.

ingvagabund: Which rises a another question: What is the proper use of undefined variable? Should I be working with a concept of undefined variables or should I define a variable as soon as I find out it is not defined?

docs/proposals/role-entry-points.md
kwoodson: @ingvagabund  I agree that it would be more difficult to tell whether it was passed or created locally but it would be easy for me to determine that it is only used in this role because of the prefix of r_<role name>_<var name>.

I think what @mtnbikenc was saying is that if you set a variable like, `l_is_atomic`  then that is now set globally across all roles.   That means that other roles just inherit that variable with no prior knowledge of it being set.  This encourages the behavior of using it if it is defined.  The issue with this is that the role no longer stands alone.  If we go with a prefix with r_<role_name> then we continue to localize variables to roles and we have a standard variable naming scheme.

I'd probably be ok with another naming scheme if the prefix retained r_<role_name> then added _local_ or something.  If not, I'm not a fan of l_<var name>.

docs/proposals/role-entry-points.md
michaelgugino: I'm against using set_fact whenever possible.

I also don't care this pattern:
`
- include_role:
    name: openshift_node
  vars:
    r_openshift_node_kubelet_args: "{{ openshift_node_kubelet_args | default({}) }}"
`
The reason is because it's obscuring the role's variables.  If someone wants to investigate a role's behavior, they have to look all over our plays to figure out what's actually being passed into the role.

IMO, the only solution to these problems is to define sensible defaults and have values set in inventory/group_vars.  Why inventory/group_vars?  Because
1) We shouldn't have to rely on a specific play to setup the role 'just right' in order for the role to do it's work.  The role needs to be as independent of the play calling it as possible, plays for the most part should just 'call the role.'
2) Since we're providing plays to the user, plays are not user-modifiable.  If we were only providing a few roles, then users would create their own plays and can define variables however they want.  Since plays are not user-modifiable, we need to provide a pattern where users can easily modify the behavior of roles and plays, and this is done through inventory/group/host_vars/extra_vars.

I prefer inventory group and host_vars because they are implicitly tied to inventory, and I think it better fits the pattern of how ansible should be consumed.

docs/proposals/role-entry-points.md
michaelgugino: A role should not utilize variables not defined inside the role whenever possible.
eg, inside the role, role_x, all variables are role_x_something, role_x_something_else.  Never role_y_something inside of role_x.  If role_x and role_y need the same value for some variable, those variables should be set in inventory or elsewhere individually, possibly pointing to the same meta variable.

If role_y sets a fact that role_x needs to consume, and the only way to have that data is via role_x consuming role_y's variable directly, we should consider if we in fact need two separate roles, as we probably need one role.

docs/proposals/role-entry-points.md
ingvagabund: @kwoodson lemme write another proposal for it.

docs/proposals/role-entry-points.md
ingvagabund: So you are saying we should set `r_` prefixed variables in an inventory file?

kwoodson: @michaelgugino, @mtnbikenc @ingvagabund, for variable passing I liked the approach from the ansible roles for fun and profit video.  
```
# call the role
- include_role:
    name: openshift_node
  vars:
    r_openshift_node_kubelet_args: "{{ openshift_node_kubelet_args | default('UNDEFINED'}}"
```
Then inside of the role we check for `UNDEFINED` and do the appropriate action.
```
- set_fact:
    r_openshift_node_kubelet_args: "{{ __r_openshift_node_kubelet_args }}"
  when:  r_openshift_node_kube_args == 'UNDEFINED'
```
This approach gives us the most flexibility by handling `undefined` inventory vars and defaulting to the role's variables.  This was mentioned in the talk at [ansiblefest](https://www.ansible.com/ansible-roles-by-acquia) @8:45 by Jeff as a good way to pass variables in and maintain default variables as well.

This localizes all of the special knowledge of the variables to our role instead of the calling playbook.  This does introduce yet another `__` prefix.  Thoughts on this approach?

docs/proposals/role-entry-points.md
sosiouxme: It does seem to me that `r_<role_name>` would be best reserved for inputs and outputs from a role. (You could *even* consider a different prefix for outputs, say `o_<role_name>`. Or not.)

I think there are two use cases for "local" variables (though there's no such thing as a local variable). One is the `l_result` pattern where a task sets the variable and then immediately other tasks in the same task file use it, and it's really easy to see that the variable won't accidentally get a value from somewhere unexpected. The other is the pattern where a variable is set internally and meant to be used in other places in the role, for example in a template, but it's not as simple to determine (visually) whether and where it was set, and there's reasonable risk that you might include tasks from another role or even the same role that could use the same name and overwrite what you thought you were getting.

So here's an idea. As in the python tradition, prefix local variables with `_`. If it's a variable that has some risk of conflicting with another "local" variable, further prefix it with the role name, so `_<role_name>_internal_value`. We could consider prefixing the "super local" variables with another `_` so e.g. `__result`.

sosiouxme:
I need to throw a wrench in things here. There's a lot in this proposal and I haven't finished reading it, but it's very relevant to the problem of role decomposition. I would like to stop thinking of roles as monolithic things that you run, and consider how we can run them in phases; to enable completing a phase across all roles before moving on to run all the roles in a different phase. Here are some phases I think we need:

* **Informational** - determine values and set facts accordingly (either as local/private variables or role variable outputs). In some plays (e.g. for checks) this might be the only phase you want to run.
* **Pre-validation** - determine whether the things the role is going to need, for example RPMs / images / tools on localhost, are present, and whether the variables provided actually make sense (are defined, well-formed, come from an acceptable set, etc.) for the purpose of failing fast before any roles make any changes.
* **Run** - make changes that the role is intended for
* **Post-validation** - check whether the actions the role took actually worked; for instance, if the role defines a router, check that the routing pod(s) are actually running, so that we don't declare success and leave the user to figure out nothing is actually running (this happens to me all the time). This phase could run repeatedly until success, defined failure, or timeout. We do this after the run phase for roles where it's OK for things to come up in parallel and achieve "eventual consistency" so that we don't have to wait synchronously for e.g. the router, registry, logging, metrics to actually come up at the time we define them. If we did need to wait, e.g. for the master API to start, we would do that in the run phase.

We may come up with more. Hopefully the need for this phased approach is clear (but... discuss if not). However implementing it is very complicated. We still want the roles to remain at similar granularity as today, where each encapsulates a logical feature. But we want to include different bits of the same role at different phases, and still have all the dependencies work out. To make it worse, the informational/pre-validation phases for some roles might depend on the run phase of others; good examples are `openshift_repos` and `docker` - currently we need to have repos configured and docker running before we can even determine what `openshift_version` should be.

Naively we might just refactor away `tasks/main.yml`, declare standard task file names for the different phases, and define big loops that run the roles in different phases:

    - set_fact:
        roles:
          - role1
          - role2

    - include_role:
        name: "{{ item }}"
        tasks_from: bootstrap.yml
      with_items: "{{ roles }}"

    - include_role:
        name: "{{ item }}"
        tasks_from: pre-validation.yml
      with_items: "{{ roles }}"

    - include_role:
        name: "{{ item }}"
        tasks_from: run.yml
      with_items: "{{ roles }}"

    # TODO: some kind of logic for repeating until success/failure
    - include_role:
        name: "{{ item }}"
        tasks_from: post-validation.yml
      with_items: "{{ roles }}"

This could probably be generalized into some kind of `role_runner` plugin so that playbooks just define a play with the roles needed and this logic occurs. And cross-phase dependencies could be handled by having roles explicitly include other roles' phases as part of their tasks, or some other playbook task that runs outside of this logic.

However this requires externalizing the role dependencies that are currently (mostly) encapsulated in the roles, which means when they change you get to make that change everywhere. And if you're having to pass in a bunch of different variables for each role, it is plainly going to be a nightmare to have to do so multiple times under this structure, individually for every play that includes the role.

So, I don't want to completely derail the conversation here, but I would like to hear thoughts on whether the need for phases can be made compatible with this proposal for exactly how to define and use role variables.

docs/proposals/role-entry-points.md
ingvagabund: As discussed offline, we could use `omit` instead of `UNDEFINED`:

```yaml
# call the role
- include_role:
    name: openshift_node
  vars:
    r_openshift_node_kubelet_args: "{{ openshift_node_kubelet_args | default(omit) }}"
    r_openshift_node_system_registry: "{{ openshift_system_registry | default(omit) }}"
```
Then inside of the role we check for `__omit_place_holder` and do the appropriate action.
```yaml
# this assumes the value of omit in __omit_place_holder__[\da-e]{40}
- set_fact:
    __omit_str: "__{{ omit.split('__')[1] }}__"

- set_fact:
    r_openshift_node_kubelet_args: "{{ __r_openshift_node_kubelet_args }}"
  when:  __omit_str in r_openshift_node_kube_args
- set_fact:
    r_openshift_node_system_registry: "{{ __r_openshift_node_system_registry }}"
  when:  __omit_str in r_openshift_node_system_registry

```

docs/proposals/role-entry-points.md
ingvagabund: > It does seem to me that r_<role_name> would be best reserved for inputs and outputs from a role. (You could even consider a different prefix for outputs, say o_<role_name>. Or not.)

The `r_<role_name>` form is fine for both inputs and outputs as even an input can become an output.

> We could consider prefixing the "obviously private" variables with another _ so e.g. __result.

> The other is the pattern where a variable is set internally and meant to be used in other places in the role, for example in a template

@mtnbikenc @ashcrow @kwoodson WDYT? `__` prefix is more transparent than the `l_` prefix, providing the same functionality. We should really not define `r_<role_name>` prefixed variables that are for internal role purposes only.

docs/proposals/role-entry-points.md
michaelgugino: @kwoodson not 'ideal' but I think a good approach considering how we currently do things and I +1.

@ingvagabund +1 for default(omit)

michaelgugino:
@sosiouxme I really like your suggestion.  It would also be helpful to have local and global skip_pre_check and skip_validation as supported options in this scenario as it could possibly improve speed significantly.

I agree we should have a whole additional discussion on this topic.

ingvagabund:
> I would like to stop thinking of roles as monolithic things that you run, and consider how we can run them in phases; to enable completing a phase across all roles before moving on to run all the roles in a different phase.

This perfectly suites the "action-based roles" principle. I will include it in my proposal.

docs/proposals/role-entry-points.md
ewolinetz: Shouldn't our private variables be defined within `vars/some_file.yml`? From the **Roles for Fun and Profit** video the idea was that variables in `var` are ones we don't advertise or want users to change easily, `default` are ones we are fine with users specifying in inventories.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,302 @@
...
+  when:
+  - r_openshift_node_deployment_type is defined
+  - r_openshift_node_deployment_type not in ['origin', 'openshift-enterprise']
+```
+
+All the checks can be stored into `tasks/pre_checks.yml` file and included in the main task file.
+
+### Local variables
+
+All variables set inside a role has to be prefixed with `__` string.
```
ewolinetz: Isn't this the format for private vars? Shouldn't our local vars be prefixed with the role name to limit scope issues?

e.g. `__image_name` in one role could conflict with `__image_name` in a different role (there was an issue similar to this with logging and the variable `replicas` being used with the router/registry)

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,302 @@
...
+...
+- role: openshift_node
+...
+```
+
+**After:**
+
+`openshift_node/defaults/main.yml`:
+```yaml
+__omit_str: "__{{ omit.split('__')[1] }}__"
```
ewolinetz: to discourage users from overriding this value, would it make more sense for this to be in `vars/main.yml`?

docs/proposals/role-entry-points.md
michaelgugino: I endorse @ewolinetz on this topic.

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,302 @@
+# Entry-point based roles
+
+## Variables inside a role are too open
+
+Each role consists of a set of tasks and variables read/set in some of the tasks.
+Some of the tasks can store their results into registered variables,
+e.g. store stdout of a docker pull command.
+Some of the tasks may be run only under certain conditions that are generated
```
michaelgugino: There's lots of things dynamically generated, like 'openshift.common.x'.  Should we consume openshift.common.x?  No, we shouldn't, but the way we are currently using openshift_facts kind of forces our hand.

docs/proposals/role-entry-points.md
sosiouxme: IMHO, if it is not completely obvious that a private variable is set locally before being used, it should have the role name in the prefix to prevent just this kind of collision. In practice, this would probably mean you only use `__foo` within a single task file where the first thing you do is set it. Otherwise `__role_name_foo`.

docs/proposals/role-entry-points.md
ingvagabund: As @michaelgugino says. I can rephrase the "conditions generated outside of a role" to something like "conditions a role has no control over since variables inside the condition are non-role variables, e.g. inventory variables, variables set in a play or generated by the `openshift_facts`".

docs/proposals/role-entry-points.md
```diff
@@ -0,0 +1,302 @@
...
+outside of a role, e.g. pull images only if a deployment is containerized.
+Some of the tasks are run only if a certain variable is defined,
+e.g. add additional Kubelet arguments into the `node-config.yml`
+if `kubelet_additional_args` inventory variable is defined.
+
+Some roles work with a lot of variables that from a point of view of a role are uninvited citizens.
```
ingvagabund: Anything that makes role maintenance harder. Like a global variables inside a C function, or like a variable over which I have only a limited knowledge (e.g. don't know which values are accepted, if the variable is always defined or just sometimes, or where is the variable set.

docs/proposals/role-entry-points.md
ingvagabund: Yeah, I did not want to separate `r_<role>_<var>` variable from its `_<role>_var` default value. However, with that justification in mind, it makes perfect sense to move all the constant "private" variables under `var/main.yml`.

docs/proposals/role-entry-points.md
ingvagabund: The pure `__` prefix without the <role> is meant to be used for variables that are set via `set_fact` or via `register` and consumed right after it. Given both keywords evaluate the variable value right away (please prove me wrong if I am mistaken), it should not collide with any "private" variable in another role as long as the `__` prefixed variable is not set in `defaults/<file>.yml` nor in `var/<file>.yml`. If the private variable is used among multiple files of the same role it make prefect sense to use the full `__<role>_<variable>` form.

docs/proposals/role-entry-points.md
ingvagabund: Absolutely. None of the `default` and `vars` is a good place for it as it is a hack. Would be great to make it part of the Ansible itself. Even better, make it a filter like `is_omitted` and used it as:
```yaml
- set_fact:
    r_openshift_node_kubelet_args: "{{ __openshift_node_kubelet_args }}"
  when: r_openshift_node_kube_args | is_omitted
```

ingvagabund:
@sosiouxme if you don't mind I took subset of your comments and put them into https://github.com/openshift/openshift-ansible/pull/5172

docs/proposals/role-entry-points.md
ingvagabund: Proposal here: https://github.com/openshift/openshift-ansible/pull/5157

docs/proposals/role-entry-points.md
ingvagabund: > IMO, the only solution to these problems is to define sensible defaults and have values set in inventory/group_vars. Why inventory/group_vars? Because

That would require to:
* drop the concept or `r_<role>_<variable>` role variables in favour of `<role>_<variable>` pattern.
* make sure a role uses only `<role>_` and `__<role>_` prefixed variables
* consolidate all relevant roles into action-based roles as discussed in https://github.com/openshift/openshift-ansible/pull/5172 so we minimize a need for setting role variables to values from a different role
* anything else I forget?

Even after doing all of that, there will be some places where we will have to set role variables from within a play. However, if we align with a proper practices of writing (and maintaining) Ansible roles that should be minimized.

ingvagabund: @michaelgugino I have add a new section "Outstanding questions" to collect all question we need to resolve before the proposal is merged.

docs/proposals/role-entry-points.md
ingvagabund: @ewolinetz @sosiouxme I have extended the "Local variables" section.

docs/proposals/role-entry-points.md
ingvagabund: I have moved all `__` prefixed vars under `vars/main.yml`

sosiouxme:
Guessing this is no longer relevant given the 3.9 refactor.

PR #5101, opened 2017-08-16
Add dnf support

maxamillion:
Adding support for dnf where yum was used and removing one instance of yum where we can just use rpm otherwise.

### DIFF
--- a/roles/lib_openshift/src/lib/base.py
+++ b/roles/lib_openshift/src/lib/base.py
@@ -464,13 +464,12 @@ class Utils(object):
     @staticmethod
     def openshift_installed():
         ''' check if openshift is installed '''
-        import yum
+        import rpm
 
-        yum_base = yum.YumBase()
-        if yum_base.rpmdb.searchNevra(name='atomic-openshift'):
-            return True
+        transaction_set = rpm.TransactionSet()
+        rpmquery = transaction_set.dbMatch("name", "atomic-openshift")
 
-        return False
+        return rpmquery.count() > 0
 
     # Disabling too-many-branches.  This is a yaml dictionary comparison function
     # pylint: disable=too-many-branches,too-many-return-statements,too-many-statements
--- a/roles/openshift_health_checker/library/aos_version.py
+++ b/roles/openshift_health_checker/library/aos_version.py
@@ -24,11 +24,19 @@ from ansible.module_utils.basic import AnsibleModule
 # Python 3, we use six for cross compatibility in this module alone:
 from ansible.module_utils.six import string_types
 
-IMPORT_EXCEPTION = None
+YUM_IMPORT_EXCEPTION = None
+DNF_IMPORT_EXCEPTION = None
+PKG_MGR = None
 try:
     import yum  # pylint: disable=import-error
+    PKG_MGR = "yum"
 except ImportError as err:
-    IMPORT_EXCEPTION = err
+    YUM_IMPORT_EXCEPTION = err
+try:
+    import dnf  # pylint: disable=import-error
+    PKG_MGR = "dnf"
+except ImportError as err:
+    DNF_IMPORT_EXCEPTION = err
 
 
 class AosVersionException(Exception):
@@ -47,8 +55,11 @@ def main():
         supports_check_mode=True
     )
 
-    if IMPORT_EXCEPTION:
-        module.fail_json(msg="aos_version module could not import yum: %s" % IMPORT_EXCEPTION)
+    if YUM_IMPORT_EXCEPTION and DNF_IMPORT_EXCEPTION:
+        module.fail_json(
+            msg="aos_version module could not import yum or dnf: %s %s" %
+            (YUM_IMPORT_EXCEPTION, DNF_IMPORT_EXCEPTION)
+        )
 
     # determine the packages we will look for
     package_list = module.params['package_list']
@@ -83,9 +94,6 @@ def _to_dict(pkg_list):
 
 
 def _retrieve_available_packages(expected_pkgs):
-    # search for package versions available for openshift pkgs
-    yb = yum.YumBase()  # pylint: disable=invalid-name
-
     # The openshift excluder prevents unintended updates to openshift
     # packages by setting yum excludes on those packages. See:
     # https://wiki.centos.org/SpecialInterestGroup/PaaS/OpenShift-Origin-Control-Updates
@@ -94,17 +102,34 @@ def _retrieve_available_packages(expected_pkgs):
     # attempt to determine what packages are available via yum they may
     # be excluded. So, for our purposes here, disable excludes to see
     # what will really be available during an install or upgrade.
-    yb.conf.disable_excludes = ['all']
 
-    try:
-        pkgs = yb.pkgSack.returnPackages(patterns=expected_pkgs)
-    except yum.Errors.PackageSackError as excinfo:
-        # you only hit this if *none* of the packages are available
-        raise AosVersionException('\n'.join([
-            'Unable to find any OpenShift packages.',
-            'Check your subscription and repo settings.',
-            str(excinfo),
-        ]))
+    if PKG_MGR == "yum":
+        # search for package versions available for openshift pkgs
+        yb = yum.YumBase()  # pylint: disable=invalid-name
+
+        yb.conf.disable_excludes = ['all']
+
+        try:
+            pkgs = yb.pkgSack.returnPackages(patterns=expected_pkgs)
+        except yum.Errors.PackageSackError as excinfo:
+            # you only hit this if *none* of the packages are available
+            raise AosVersionException('\n'.join([
+                'Unable to find any OpenShift packages.',
+                'Check your subscription and repo settings.',
+                str(excinfo),
+            ]))
+    elif PKG_MGR == "dnf":
+        dbase = dnf.Base()  # pyling: disable=invalid-name
+
+        dbase.conf.disable_excludes = ['all']
+        dbase.read_all_repos()
+        dbase.fill_sack(load_system_repo=False, load_available_repos=True)
+
+        dquery = dbase.sack.query()
+        aquery = dquery.available()
+
+        pkgs = list(aquery.filter(name=expected_pkgs))
+
     return pkgs
 
 
--- a/roles/openshift_health_checker/library/aos_version.py
+++ b/roles/openshift_health_checker/library/aos_version.py
@@ -130,6 +130,14 @@ def _retrieve_available_packages(expected_pkgs):
 
         pkgs = list(aquery.filter(name=expected_pkgs))
 
+        if not pkgs:
+            # pkgs list is empty, raise because no expected packages found
+            raise AosVersionException('\n'.join([
+                'Unable to find any OpenShift packages.',
+                'Check your subscription and repo settings.',
+                str(excinfo),
+            ]))
+
     return pkgs
 
 
--- a/roles/openshift_health_checker/library/aos_version.py
+++ b/roles/openshift_health_checker/library/aos_version.py
@@ -135,7 +135,6 @@ def _retrieve_available_packages(expected_pkgs):
             raise AosVersionException('\n'.join([
                 'Unable to find any OpenShift packages.',
                 'Check your subscription and repo settings.',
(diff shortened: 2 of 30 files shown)


maxamillion:
Also note, I will squash once I get a +1.

roles/lib_openshift/src/lib/base.py
```diff
@@ -464,13 +464,19 @@ def add_custom_versions(versions):
...
 
-        yum_base = yum.YumBase()
-        if yum_base.rpmdb.searchNevra(name='atomic-openshift'):
-            return True
+        transaction_set = rpm.TransactionSet()
+        rpmquery = transaction_set.dbMatch("name", "atomic-openshift")
 
-        return False
+        results = []
+        for rpm_hdr in rpmquery:
```
kwoodson: Could this be made simpler by doing?
```
return rpmquery.count() > 0
```

maxamillion: Absolutely, will update.

roles/lib_openshift/src/lib/base.py
```diff
@@ -464,13 +464,12 @@ def add_custom_versions(versions):
     @staticmethod
     def openshift_installed():
         ''' check if openshift is installed '''
-        import yum
+        import rpm
```
ashcrow: 👍 

rpm/yum libs are not available on Atomic Host but I believe this never fires when using Atomic.

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -94,17 +102,34 @@ def _retrieve_available_packages(expected_pkgs):
...
+        try:
+            pkgs = yb.pkgSack.returnPackages(patterns=expected_pkgs)
+        except yum.Errors.PackageSackError as excinfo:
+            # you only hit this if *none* of the packages are available
+            raise AosVersionException('\n'.join([
+                'Unable to find any OpenShift packages.',
+                'Check your subscription and repo settings.',
+                str(excinfo),
+            ]))
+    elif PKG_MGR == "dnf":
```
ashcrow: 👍

maxamillion:
Thanks all! Will squash.

ashcrow:
I did notice this:

https://travis-ci.org/openshift/openshift-ansible/jobs/265289887#L806

but it's odd that Python 3 was fine yet 2 was not. I'm thinking it was some kind of weird flake in generated file testing.

ashcrow:
The generation test failed again. I'm going to test it locally to ensure it's a true failure.

maxamillion:
@ashcrow thanks - let me know what (if anything) needs fixing and I'll be happy to.

ashcrow:
@maxamillion looks like it's a legit failure. Regeneration needs to be rerun:

```
python roles/lib_openshift/src/generate.py
```

This will regenerate code using the new ``base.py``.

maxamillion:
@ashcrow will do, thanks

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -94,17 +102,34 @@ def _retrieve_available_packages(expected_pkgs):
...
+        dbase = dnf.Base()  # pyling: disable=invalid-name
+
+        dbase.conf.disable_excludes = ['all']
+        dbase.read_all_repos()
+        dbase.fill_sack(load_system_repo=False, load_available_repos=True)
+
+        dquery = dbase.sack.query()
+        aquery = dquery.available()
+
+        pkgs = list(aquery.filter(name=expected_pkgs))
```
michaelgugino: In the yum version, we're testing if pkgs has something (returnPackages will raise if no matching packages are found).  I'm unfamiliar with the dnf lib, but we should probably test for an empty list here if aquery doesn't raise.  In any case, we should catch the exception if raised as we do in yum as this function is expected to raise AosVersionException if no packages are found.

ashcrow: That's fair. FWIW it ``dnf`` doesn't raise if nothing is found:

```
In [1]: import dnf

In [2]: dbase = dnf.Base()

In [3]: dbase.conf.disable_excludes = ['all']

In [4]: dbase.read_all_repos()

In [5]: dbase.fill_sack(load_system_repo=False, load_available_repos=True)
Out[5]: <dnf.sack.Sack at 0x7fc979c16990>

In [6]: dquery = dbase.sack.query()

In [7]: aquery = dquery.available()

In [8]: pkgs = list(aquery.filter(name=['idonotexist']))

In [9]: pkgs
Out[9]: []
```

maxamillion: Since dnf doesn't raise I didn't really consider an error case. Is it preferred to raise if nothing is found?

michaelgugino: It appears that this module is intended to check if the packages supplied as parameters are valid packages.  If the search for the provided packages yields nothing, we should raise as we do in yum, the calling function (main) will catch that specific exception and call module.fail to inform ansible of the status.

maxamillion: @michaelgugino sounds good, will do. :+1:

michaelgugino: I've looked at this a little more closely.

It seems the yum version is also broken, and that this plugin only works as intended if you're using versioned packages.  It seems an error is only raised for yum if there's a problem connection to the repo; it will happily return a list of zero packages.

A list of zero packages is not checked anywhere unless we're using versioned packages.

Additionally, checking the list for only for zero-length results in an error case as well.

For instance:
```
pkgs = yb.pkgSack.returnPackages(patterns=['yum', 'gcc99x'])
```
gcc99x is not a real package.  pkgs will return a list of yum packages; so the list is greater than zero length, but we didn't actually find all the packages we were looking for.

We're probably not hitting or seeing this bug currently because the packages are probably in the same repo together, or we might be using versions, IDK.  We should probably iterate through the package list, create a dictionary with available packages with key = desired_package, and return that to the calling function; calling function or this function can check if any of the values for the dictionary keys (package names) is zero length.  This will allow us to ensure that each desired package is available, and possibly each desired version of each package.

Maybe that's all oustide the scope of this PR.

ashcrow: @michaelgugino I like your idea. I believe we are passing versioned packages with this module but that doesn't mean we will always be doing so. IMHO that's a bug unrelated to this set of work and, as you noted, it could be fixed within another PR by a core team member.

roles/openshift_health_checker/library/aos_version.py
```diff
@@ -130,6 +130,14 @@ def _retrieve_available_packages(expected_pkgs):
 
         pkgs = list(aquery.filter(name=expected_pkgs))
 
+        if not pkgs:
+            # pkgs list is empty, raise because no expected packages found
+            raise AosVersionException('\n'.join([
+                'Unable to find any OpenShift packages.',
+                'Check your subscription and repo settings.',
+                str(excinfo),
```
ashcrow: Almost there! I don't believe ``excinfo`` is valid in this scope. I recommend removing it and just raising the ``AosVersionException`` with the error message.

maxamillion: @ashcrow ha! sorry about that, totally lazy on my side. I just copied and didn't even notice that (interestingly, Syntastic didn't yell at me for it).

roles/openshift_health_checker/library/aos_version.py
michaelgugino: @ashcrow I agree.  This lPR ooks good for now, we'll just need to get a bug filed, or if @maxamillion feels inclined to send another PR for that bug, that would be good too.

ashcrow:
aos-ci-test

ashcrow:
[test]

ashcrow:
If tests come back 👍 then this should be merged. A backport should be created for ``v3.6`` as well.

maxamillion:
@ashcrow I'm not sure what failed the CI, I looked at the error and I'm not sure how that could have come from these changes.

```
+ tito tag --offline --accept-auto-changelog
Creating output directory: /tmp/tito
Tagging new version of openshift-ansible: 3.7.0-0.101.0 -> 3.7.1-1
  File "/usr/bin/tito", line 23, in <module>
    CLI().main(sys.argv[1:])
  File "/usr/lib/python2.7/site-packages/tito/cli.py", line 202, in main
    return module.main(argv)
  File "/usr/lib/python2.7/site-packages/tito/cli.py", line 666, in main
    return tagger.run(self.options)
  File "/usr/lib/python2.7/site-packages/tito/tagger/main.py", line 114, in run
    self._tag_release()
  File "/usr/lib/python2.7/site-packages/tito/tagger/main.py", line 136, in _tag_release
    self._check_tag_does_not_exist(self._get_new_tag(new_version))
  File "/usr/lib/python2.7/site-packages/tito/tagger/main.py", line 501, in _check_tag_does_not_exist
    raise Exception("Tag %s already exists!" % new_tag)
Exception: Tag openshift-ansible-3.7.1-1 already exists!
```

sdodson:
that problem with the CI jobs should be gone now
[merge]

maxamillion:
thanks!

ashcrow:
[merge]

ashcrow:
The failures look like they were unrelated to this change.

roles/lib_openshift/src/lib/base.py
sosiouxme: rpm libs aren't available? Atomic still has RPM and yum even if they're only informational...

maxamillion:
@ashcrow so if the failures are unrelated but blocking the merge, what do we do?

ashcrow:
Also unrelated to this PR:
```
TASK [openshift_node_dnsmasq : Install dnsmasq configuration] ******************
task path: /usr/share/ansible/openshift-ansible/roles/openshift_node_dnsmasq/tasks/main.yml:28
    "changed": false, 
    "failed": true, 
    "generated_timestamp": "2017-08-21 22:48:29.755175", 
    "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute 'dns_ip'"
}
```

@sdodson we seem to be having some issues testing .... are there any known issues in the CI over the last day?

sdodson:
The CI failure should be fixed momentarily, we should wait on that and try again. Pretty much every time I manually merge something because CI jobs are blocking us we're introducing bugs. I don't think there's anything wrong with this PR but I want to get us back to treating CI job failures as critical items

ashcrow:
@sdodson understood. Ping me when we should try again.

sdodson:
[merge]

roles/openshift_health_checker/library/aos_version.py
sosiouxme: I wrote this originally, some of my first real python/ansible code, so it's far from perfect :)

The results from `_retrieve_available_packages` are used as input for further logic about which versions are available. If you don't get back all the packages you asked for, that would be noted by the further logic. The comment may be inaccurate as it looks like `returnPackages` returns an empty list without exception. But if yum does raise, it converts that exception into one that's meaningful to the caller.

I don't think it is a problem for this method to return an empty list. So the dnf case doesn't need to raise an exception on an empty list. However it should convert any exception it does raise into `AosVersionException` so that it is handled instead of resulting in a nondescript MODULE FAILURE.

maxamillion: To round back on this, is there anything needed from me here or is everyone discussing improvements as general statements that are outside the scope of this pull request? (Either is fine, I just want to make sure nobody is expecting a change from me and I missed it along the way).

ashcrow: @maxamillion I don't believe there is anything you need to do for this PR unless you get a negative review. You've subsequently added a change to raise ``AosVersionException``.

ashcrow:
```
    raise Exception("Tag %s already exists!" % new_tag)
```

Error is unrelated to this PR.

ashcrow:
[merge]

roles/openshift_health_checker/library/aos_version.py
maxamillion: @ashcrow thanks :+1:

sdodson:
The system containers job is busted. Working on fixing that.

PR #5102, opened 2017-08-16
only evaluate openshift_cfme_nfs_server if using nfs

ilackarms:
Currently if no nfs groups are defined, `openshift_cfme_nfs_server` is being evaluated in openshift-cfme role's `defaults.yml`

This variable is not needed if we are not using NFS for deployment, and can cause error if the host's inventory file does not contain an `nfs` hosts section.

This change modifies `openshift_cfme_nfs_server` to be set via `set_fact` rather than as a  default variable, running only as a part of the `nfs.yml` tasks (and onyl when `openshift_cfme_nfs_server` has not been set by the user)

### DIFF
--- a/roles/openshift_cfme/tasks/nfs.yml
+++ b/roles/openshift_cfme/tasks/nfs.yml
@@ -1,6 +1,13 @@
 ---
 # Tasks to statically provision NFS volumes
 # Include if not using dynamic volume provisioning
+
+- name: Set openshift_cfme_nfs_server fact
+  when: openshift_cfme_nfs_server is not defined
+  set_fact:
+    # Hostname/IP of the NFS server. Currently defaults to first master
+    openshift_cfme_nfs_server: "{{ oo_nfs_to_config.0 }}"
+
 - name: Ensure the /exports/ directory exists
   file:
     path: /exports/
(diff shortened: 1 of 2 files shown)


ilackarms:
@tbielawa can you confirm that this is a proper fix for #4556 ?

@simon3z cc

simon3z:
LGTM :+1: 
cc @tbielawa

sdodson:
aos-ci-test

sdodson:
[test]

roles/openshift_cfme/tasks/nfs.yml
```diff
@@ -1,6 +1,13 @@
 ---
 # Tasks to statically provision NFS volumes
 # Include if not using dynamic volume provisioning
+
+- name: Set openshift_cfme_nfs_server fact
+  when: openshift_cfme_nfs_server is not defined
+  set_fact:
+    # Hostname/IP of the NFS server. Currently defaults to first master
+    openshift_cfme_nfs_server: "{{ groups.nfs.0 }}"
```
tbielawa: I think the correct value for `openshift_cfme_nfs_server` should be `oo_nfs_to_config.0` That might also help cut down on failures due to un-defined groups. That is because `playbooks/byo/openshift-cfme/config.yml` calls the `initialize_groups.yml` and `evaluate_groups.yml` task files which define our canonical group names (and members) so that playbooks/roles can run independent of how things are named in inventories.

tbielawa:
I will re-run the tests (CI exploded) after I hear back from you. Thanks!

ilackarms:
@tbielawa fixed that variable

tbielawa:
[test]

tbielawa:
aos-ci-test

PR #5107, opened 2017-08-16
Remove openshift_repos dependencies

mtnbikenc:
With the current refactoring of openshift_repos, it is no longer dependent on openshift_sanitize_inventory because that role is already run as part of initialization.


mtnbikenc:
aos-ci-test

mtnbikenc:
[merge]

PR #5110, opened 2017-08-16
Allow admins to add additional CA certs to the master bundle

smarterclayton:
Injects one or more CAs into the final cluster


smarterclayton:
aos-ci-test

abutcher:
We'll either need to copy these additional CAs up to masters temporarily or permanently so that we can access them during CA/bundle creation. Since multiple roles will need them maybe we go with a permanent home in `/etc/origin/master`? Something like https://github.com/abutcher/openshift-ansible/commit/8c42b291a0cda73a4ab810520124bbef00f63a14 @smarterclayton

smarterclayton:
Another option - slam them all into the trusted CA bundle.  We don't ever recalculate that, do we?

abutcher:
@smarterclayton We do recalculate it during a redeploy and pull in CAs that were copied to a legacy-ca directory.

PR #5037, opened 2017-08-09
Bump Calico to v2.4.1

ozdanborne:

PR #5111, opened 2017-08-17
Use sdn_cluster_network_cidr as default calico pool

ozdanborne:
This PR removes `calico_ipv4pool_cidr` and uses the value of `osm_cluster_network_cidr` (which is actually stored as `openshift.master.sdn_cluster_network_cidr`) as the default pool for Calico.

Without this PR, the Calico pool and cluster-cidr can easily be mismatched, which causes some unexpected behavior with service IPs & kube-proxy.

This PR is follow up from @sdodson 's comment: https://github.com/openshift/openshift-ansible/pull/5037#discussion_r132076206 . His point was true - I confirmed that `osm_cluster_network_cidr` is undefined in ansible if undefined in inventory. I found that `openshift.master.sdn_cluster_network_cidr` does however have the correct value.

### DIFF
--- a/roles/calico/README.md
+++ b/roles/calico/README.md
@@ -32,7 +32,6 @@ Additional parameters that can be defined in the inventory are:
 
 | Environment | Description | Schema | Default |   
 |---------|----------------------|---------|---------|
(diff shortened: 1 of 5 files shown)


sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
travis is happy, i'm happy

PR #5112, opened 2017-08-17
Adding oc_adm_csr to lib_openshift for node approvals

kwoodson:
This pull request coincides with the node bootstrapping.  I have introduced `oc_adm_csr.py` to the `lib_openshift` modules.  This allows us to approve nodes in an automated fashion.  I have added a few parameters in attempt to facilitate auto acceptance.
- Added a timeout in case waiting takes forever.
- Added a parameter for a node list so that we can pass in an expected list of nodes that should be approved
- Added a flag to approve_all nodes.  This is more for testing.

### DIFF
--- a/roles/lib_openshift/src/class/oc_adm_csr.py
+++ b/roles/lib_openshift/src/class/oc_adm_csr.py
@@ -1,0 +1,197 @@
+# pylint: skip-file
+# flake8: noqa
+
+
+class OCcsr(OpenShiftCLI):
+    ''' Class to wrap the oc adm certificate command line'''
+    kind = 'csr'
+
+    # pylint: disable=too-many-arguments
+    def __init__(self,
+                 nodes=None,
+                 approve_all=False,
+                 service_account=None,
+                 kubeconfig='/etc/origin/master/admin.kubeconfig',
+                 verbose=False):
+        ''' Constructor for oc adm certificate '''
+        super(OCcsr, self).__init__(None, kubeconfig, verbose)
+        self.service_account = service_account
+        self.nodes = self.create_nodes(nodes)
+        self._csrs = []
+        self.approve_all = approve_all
+        self.verbose = verbose
+
+    @property
+    def csrs(self):
+        '''property for managing csrs'''
+        # any processing needed??
+        self._csrs = self._get(resource=self.kind)['results'][0]['items']
+        return self._csrs
+
+    def create_nodes(self, nodes):
+        '''create a node object to track csr signing status'''
+        nodes_list = []
+
+        if nodes is None:
+            return nodes_list
+
+        results = self._get(resource='nodes')['results'][0]['items']
+
+        for node in nodes:
+            nodes_list.append(dict(name=node, csrs={}, accepted=False, denied=False))
+
+            for ocnode in results:
+                if node in ocnode['metadata']['name']:
+                    nodes_list[-1]['accepted'] = True
+
+        return nodes_list
+
+    def get(self):
+        '''get the current certificate signing requests'''
+        return self.csrs
+
+    @staticmethod
+    def action_needed(csr, action):
+        '''check to see if csr is in desired state'''
+        if csr['status'] == {}:
+            return True
+
+        state = csr['status']['conditions'][0]['type']
+
+        if action == 'approve' and state != 'Approved':
+            return True
+
+        elif action == 'deny' and state != 'Denied':
+            return True
+
+        return False
+
+    def match_node(self, csr):
+        '''match an inc csr to a node in self.nodes'''
+        for node in self.nodes:
+            # we have a match
+            if node['name'] in csr['metadata']['name']:
+                node['csrs'][csr['metadata']['name']] = csr
+
+                # check that the username is the node and type is 'Approved'
+                if node['name'] in csr['spec']['username'] and csr['status']:
+                    if csr['status']['conditions'][0]['type'] == 'Approved':
+                        node['accepted'] = True
+                # check type is 'Denied' and mark node as such
+                if csr['status'] and csr['status']['conditions'][0]['type'] == 'Denied':
+                    node['denied'] = True
+
+                return node
+
+        return None
+
+    def finished(self):
+        '''determine if there are more csrs to sign'''
+        # if nodes is set and we have nodes then return if all nodes are 'accepted'
+        if self.nodes is not None and len(self.nodes) > 0:
+            return all([node['accepted'] or node['denied'] for node in self.nodes])
+
+        # we are approving everything or we still have nodes outstanding
+        return False
+
+    def manage(self, action):
+        '''run openshift oc adm ca create-server-cert cmd and store results into self.nodes
+
+           we attempt to verify if the node is one that was given to us to accept.
+
+           action - (allow | deny)
+        '''
+
+        results = []
+        # There are 2 types of requests:
+        # - node-bootstrapper-client-ip-172-31-51-246-ec2-internal
+        #   The client request allows the client to talk to the api/controller
+        # - node-bootstrapper-server-ip-172-31-51-246-ec2-internal
+        #   The server request allows the server to join the cluster
+        # Here we need to determine how to approve/deny
+        # we should query the csrs and verify they are from the nodes we thought
+        for csr in self.csrs:
+            node = self.match_node(csr)
+            # oc adm certificate <approve|deny> csr
+            # there are 3 known states: Denied, Aprroved, {}
+            # verify something is needed by OCcsr.action_needed
+            # if approve_all, then do it
+            # if you passed in nodes, you must have a node that matches
+            if self.approve_all or (node and OCcsr.action_needed(csr, action)):
+                result = self.openshift_cmd(['certificate', action, csr['metadata']['name']], oadm=True)
+                # client should have service account name in username field
+                # server should have node name in username field
+                if node and csr['metadata']['name'] not in node['csrs']:
+                    node['csrs'][csr['metadata']['name']] = csr
+
+                    # accept node in cluster
+                    if node['name'] in csr['spec']['username']:
+                        node['accepted'] = True
+
+                results.append(result)
+
+        return results
+
+    @staticmethod
+    def run_ansible(params, check_mode=False):
+        '''run the idempotent ansible code'''
+
+        client = OCcsr(params['nodes'],
+                       params['approve_all'],
+                       params['service_account'],
+                       params['kubeconfig'],
+                       params['debug'])
+
+        state = params['state']
+
+        api_rval = client.get()
+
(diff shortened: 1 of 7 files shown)


kwoodson:
aos-ci-test

ashcrow:
```
************* Module oc_adm_csr

roles/lib_openshift/library/oc_adm_csr.py:163: [E0602(undefined-variable), OpenShiftCLI._replace_content] Undefined variable 'Yedit'

roles/lib_openshift/library/oc_adm_csr.py:181: [E0602(undefined-variable), OpenShiftCLI._replace] Undefined variable 'Yedit'

roles/lib_openshift/library/oc_adm_csr.py:194: [E0602(undefined-variable), OpenShiftCLI._create_from_content] Undefined variable 'Yedit'

roles/lib_openshift/library/oc_adm_csr.py:241: [E0602(undefined-variable), OpenShiftCLI._process] Undefined variable 'Yedit'

roles/lib_openshift/library/oc_adm_csr.py:720: [R0913(too-many-arguments), OCcsr.__init__] Too many arguments (6/5)

roles/lib_openshift/library/oc_adm_csr.py:764: [R0201(no-self-use), OCcsr.action_needed] Method could be a function

roles/lib_openshift/library/oc_adm_csr.py:36: [W0611(unused-import), ] Unused import copy

roles/lib_openshift/library/oc_adm_csr.py:39: [W0611(unused-import), ] Unused import re
```

roles/lib_openshift/src/class/oc_adm_csr.py
```diff
@@ -0,0 +1,192 @@
...
+        return self._csrs
+
+    def create_nodes(self, nodes):
+        '''create a node object to track csr signing status'''
+        nodes_list = []
+
+        if nodes is None:
+            return nodes_list
+
+
```
ashcrow: Extra spacing here. Likely flake8 will complain.

roles/lib_openshift/src/class/oc_adm_csr.py
```diff
@@ -0,0 +1,192 @@
...
+                if params['timeout'] == 0:
+                    if not params['approve_all']:
+                        ctr = 0
+
+                if ctr * 2 > params['timeout']:
+                    timeout = True
+                    break
+
+                time.sleep(2)
```
ashcrow: I recommend adding a comment on why sleep is needed. Not required, but I think it would be helpful.

roles/lib_openshift/src/doc/csr
```diff
@@ -0,0 +1,47 @@
...
+    required: false
+    default: False
+    aliases: []
+  nodes:
+    description:
+    - The names of the nodes in which to accept the certificates
+    required: false
+    default: None
+    aliases: []
+author:
```
ashcrow: Missing docs for ``timeout``, ``approve_all``,  ``service_account``.

roles/lib_openshift/src/class/oc_adm_csr.py
```diff
@@ -0,0 +1,192 @@
...
+                return node
+
+        return None
+
+    def finished(self):
+        '''determine if there are more csrs to sign'''
+        # TODO RECONCILE NODES ALREADY IN CLUSTER
+
+        # no nodes were passed and we are accepting everything.  Keep going
+        if self.approve_all and len(self.nodes) == 0:
```
ashcrow: I think this could be simplified with:

```python
# If nodes is set and it has contents then return specific data
if self.nodes is not None and len(self.nodes) > 0:
    return all([node['accepted'] for node in self.nodes])

# Otherwise we don't have any node data to return
return False
```

roles/lib_openshift/src/class/oc_adm_csr.py
kwoodson: will do

roles/lib_openshift/src/class/oc_adm_csr.py
kwoodson: got it!

roles/lib_openshift/src/doc/csr
kwoodson: yeah, realized that last night when I pushed.

roles/lib_openshift/src/class/oc_adm_csr.py
kwoodson: looks good.  Simplify the logic

kwoodson:
aos-ci-test

kwoodson:
@ashcrow ready for review again, please.

roles/lib_openshift/src/class/oc_adm_csr.py
```diff
@@ -0,0 +1,194 @@
...
+        # we should query the csrs and verify they are from the nodes we thought
+        for csr in self.csrs:
+            node = self.match_node(csr)
+            # oc adm certificate <approve|deny> csr
+            # there are 3 known states: Denied, Aprroved, {}
+            # verify something is needed by OCcsr.action_needed
+            # if approve_all, then do it
+            # if you passed in nodes, you must have a node that matches
+            if self.approve_all or (node and OCcsr.action_needed(csr, action)):
+                result = self.openshift_cmd(['certificate', action, csr['metadata']['name']], oadm=True)
```
ashcrow: If this ends up having an action of denied should the node still be accepted and returned in the results?

kwoodson: I need to test with 'deny'.

kwoodson: @ashcrow, I added a couple of statements to exit when the deny flag was passed. I'm not in love with this method but the acceptance can be quite complex and in the future will be handled by the cloud-provider/k8s.  This script is meant to be temporary until something comes into the product to support the same functionality.

ashcrow: @kwoodson I'm not seeing where the exit is occurring. Mind pointing it out for me?

ashcrow:
```
************* Module oc_adm_csr

roles/lib_openshift/library/oc_adm_csr.py:1493: [W0311(bad-indentation), ] Bad indentation. Found 24 spaces, expected 20

error: lint error 4.
```

roles/lib_openshift/src/class/oc_adm_csr.py
ashcrow: If ``action`` is set to ``denied, ``node`` is populated, ``OCcsr.action_needed(csr, action))`` is called noting action is needed to ``deny`` .... doesn't line 128 still set the node to ``accepted``?

kwoodson: @ashcrow, I'll try and explain.  I'm not in love with this method of approving but I'm not sure how else to accomplish this task.

The node first comes up with a kubeconfig placed on the host through cloud user_data.  The host uses the kubeconfig (token) to talk to the master API/controllers.  
1. This first node request is from the service account to talk to the API from the node.  Inside of the csr (oc get csr) on the master it returns a field named "username".  This "username" field has the service account in it.  The admin calls `oc adm certificate approve/deny <csr>` and the certificate is approved or denied.
2. Once the service account is approved to talk to the API/controllers, a joining request is issued to the node.  The second csr that comes into the master is for the node itself to join the cluster.  The "username" field has the name of the node in it.  Thus, line 128 must match the username in order for us to consider the node accepted only after both 1 and 2 have been approved.

Its kind of confusing that there are 2 csr's per node.  Only after the 2nd one is approved do we consider the node `node['accepted']=True`.

Hopefully this sheds some light on it.  If not, we can walk through it tomorrow and I can demonstrate it better.

If we deny the 1st request then it will never issue the 2nd.  I can't think of a scenario where you would allow the server API access but then deny the node into the cluster.  It would probably be a security issue if this were to happen.  Once denied I was unable to then re-approve the node.  It never exited the denied state.

kwoodson:
aos-ci-test

roles/lib_openshift/src/class/oc_adm_csr.py
ashcrow: The way you describe the flow makes sense. It's not that unlike Commissaire's flow but using certificates instead of using key generation. I guess I'm just having a hard time following it in the code. The way I read the code is that no matter if the node is marked denied or accepted it will be accepted as accepted is hardcoded in the call. However, based on your testing and explanation, it sounds like I'm incorrect and it's just a complicated code path.

roles/lib_openshift/src/class/oc_adm_csr.py
```diff
@@ -0,0 +1,197 @@
...
+
+            return dict(changed=len(all_results) > 0,
+                        results=all_results,
+                        nodes=client.nodes,
+                        state=state,
+                        finished=finished,
+                        timeout=timeout)
+
+        return {'failed': True,
+                'msg': 'Unknown state passed. %s' % state}
```
ashcrow: Don't forget to move to ``format`` over time!

ashcrow:
@sdodson @kwoodson I'm keen on merging this. Any objections?

kwoodson:
@ashcrow, since there exists only a single code path currently calling this, I think the impact is very low that it affects anything else.

sdodson:
> @sdodson @kwoodson I'm keen on merging this. Any objections?

I haven't had a chance to look, but I trust you two to decide when it's ready.

ashcrow:
[merge]

ashcrow:
```
    Test Panicked
    runtime error: index out of range
    /usr/local/go/src/runtime/panic.go:489

    Full Stack Trace
    	/usr/local/go/src/runtime/panic.go:489 +0x2cf
    github.com/openshift/origin/test/extended/router.glob..func2.1()
    	/go/src/github.com/openshift/origin/test/extended/router/metrics.go:62 +0x6ae
    github.com/openshift/origin/vendor/github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync(0xc4213170e0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
    	/go/src/github.com/openshift/origin/test/extended/util/test.go:120 +0x3d4
    github.com/openshift/origin/test/extended.TestExtended(0xc42089bba0)
    	/go/src/github.com/openshift/origin/_output/local/go/src/github.com/openshift/origin/test/extended/extended_test.go:35 +0x40
    testing.tRunner(0xc42089bba0, 0x36198b8)
    	/usr/local/go/src/testing/testing.go:657 +0x96
    created by testing.(*T).Run
    	/usr/local/go/src/testing/testing.go:697 +0x2ca
```

ashcrow:
[merge]

kwoodson:
flake openshift/origin#13975

kwoodson:
[merge]

sdodson:
The only failure in the last test was a known flake. merging.

PR #5119, opened 2017-08-17
logging set memory request to limit

jcantrill:
Per conversation with @portante this PR sets the requests to the limit for memory of logging components so they do not get OOM killed.

### DIFF
--- a/roles/openshift_logging_curator/templates/curator.j2
+++ b/roles/openshift_logging_curator/templates/curator.j2
@@ -44,6 +44,8 @@ spec:
               cpu: "{{curator_cpu_limit}}"
 {% if curator_memory_limit is defined and curator_memory_limit is not none and curator_memory_limit != "" %}
               memory: "{{curator_memory_limit}}"
+            requests:
+              memory: "{{curator_memory_limit}}"
 {% endif %}
           env:
             -
--- a/roles/openshift_logging_elasticsearch/templates/es.j2
+++ b/roles/openshift_logging_elasticsearch/templates/es.j2
@@ -48,7 +48,7 @@ spec:
               cpu: "{{es_cpu_limit}}"
 {% endif %}
             requests:
-              memory: "512Mi"
+              memory: "{{es_memory_limit}}"
           ports:
             -
               containerPort: 9200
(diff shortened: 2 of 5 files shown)


jcantrill:
aos-ci-test

sdodson:
[merge]

jcantrill:
aos-ci-test

jcantrill:
[merge] flake?

jcantrill:
[merge]

jcantrill:

richm:
This breaks CI - the machine does not have enough RAM to run es-ops:
```
1h         1h          26        logging-es-ops-data-master-mptptw56-1-brb9n    
Pod                                                              Warning   FailedScheduling              default-scheduler                       No nodes are available that match all of the following predicates:: Insufficient memory (1).
```

richm:
@jcantrill @sdodson @stevekuznetsov so should we use a larger machine for logging testing?  revoke this commit?  lower the memory limit?

sdodson:
I'll revert for now and discuss tomorrow

portante:
We need this memory limit change, can we not revert this and fix the test environment?  This is exactly what we want to have happen when an infra node is not provisioned properly for logging.

richm:
@portante fix test environment == larger machine?

richm:
or - we change the logging CI to use a 4Gi memory limit for es and es-ops

sdodson:
I know nothing of the sizing of the hosts we're using, if that will make it fit and is a reasonable default then lets do that in a pr that re-introduces this code.

portante:
@sdodson, the defaults are reasonable already.  They have been in play for many versions.  This change is to make sure the requests matches the limits.

The testing environment is much smaller than we normally deploy to, so either we have to beef up the test environment, or parameterize the tests to accommodate the restricted test environment.

portante:
I should add that the tests are working as is, too!  They showed the exact behavior we wanted: when there isn't enough room to run logging on a node, the scheduler fails to run the pod instead of getting an OOM kill at some later point in time.

sdodson:
@richm can you open a pr to re-introduce the original changes in this PR and we'll rely on merge/test to validate that the changes unbreak the queue after https://github.com/openshift/aos-cd-jobs/pull/603 merges

PR #5124, opened 2017-08-17
bug 1482661. Preserve ES dc nodeSelector and supplementalGroups

jcantrill:
After discussion with the logging team, we decided preserving nodeSelectors and supplementalGroups is in the same spirit of preserving hostMount volumes if a user made edits.  This PR:

* Preserves ES nodeSelector if found in the facts
* Preserves ES supplementalGroup if found in the facts

@anton similar changes are required in 3.7 but I would ask you to make the necessary changes in the role refactoring you are doing

### DIFF
--- a/roles/openshift_logging/library/openshift_logging_facts.py
+++ b/roles/openshift_logging/library/openshift_logging_facts.py
@@ -171,12 +171,17 @@ class OpenshiftLoggingFacts(OCBaseCommand):
             if comp is not None:
                 spec = dc_item["spec"]["template"]["spec"]
                 facts = dict(
+                    name=name,
                     selector=dc_item["spec"]["selector"],
                     replicas=dc_item["spec"]["replicas"],
                     serviceAccount=spec["serviceAccount"],
                     containers=dict(),
                     volumes=dict()
                 )
+                if "nodeSelector" in spec:
+                    facts["nodeSelector"] = spec["nodeSelector"]
+                if "supplementalGroups" in spec["securityContext"]:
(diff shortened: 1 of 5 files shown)


jcantrill:
aos-ci-test

jcantrill:
[merge]

jcantrill:
aos-ci-test

jcantrill:
@richm thoughts?  failure related to the logging tests not being run against a 3.6 cluster?

ewolinetz:
I'm weary of implementing this still, our ansible roles should be created such that they're declarative and currently logging seems to be the exception to this.

Won't the need to do this go away with statefulSets?

Is there an issue or bz that this is resolving?

richm:
@ewolinetz There is a bz https://bugzilla.redhat.com/show_bug.cgi?id=1482661 - so if there is a better way to preserve the existing nodeSelector and security information, that would be good to have.

ewolinetz:
@richm thanks. I'm not sure there's a better way to preserve it. My thought would be that it should be specified as an inventory variable and use that rather than reuse what is currently defined in the dc.

The problem though in that bz is that each DC has unique selectors. I'm not sure if that's a case we should try to support as part of this role, again because once we transition to statefulSets we wouldn't have multiple DCs... we can preserve the replica count and then give the customer a chance to patch the DC after the upgrade to add in the custom selectors then scale up after. 

Just my $.02, I won't block this.

jcantrill:
aos-ci-test

jcantrill:
aos-ci-test

jcantrill:

PR #4264, opened 2017-05-23
Flatten all rpm packages other than atomic-openshift-utils

sdodson:
I can think of no situation where someone would want to individually install roles, playbooks, filter_plugins, or other subpackages without installing all the rest. The same applies to upgrading each of those. So in the interest of simplicity lets squash the packaging back down to just the quick installer (atomic-openshift-utils) and ansible code (openshift-ansible).

Fixes #3469

PR #5125, opened 2017-08-17
add retry logic to flaky commands

sosiouxme:
Introduces the `openshift_package` action plugin which adds retry logic to package tasks.
Also modifies the `repoquery` plugin with retry logic.
Updates `openshift_checks` python code with retry logic.

NOTE: in doing this with an action plugin, we require that action plugin to be loaded pretty much everywhere. This will require using an `ansible.cfg` that refers to the repo's `action_plugin/` directory when running playbooks.

### DIFF
--- a/lookup_plugins/oo_option.py
+++ b/lookup_plugins/oo_option.py
@@ -25,7 +25,7 @@ try:
 except ImportError:
     # ansible-1.9.x
     class LookupBase(object):
-        def __init__(self, basedir=None, runner=None, **kwargs):
+        def __init__(self, basedir=None, runner=None, **_):
             self.runner = runner
             self.basedir = self.runner.basedir
 
@@ -40,23 +40,11 @@ except ImportError:
 class LookupModule(LookupBase):
     ''' oo_option lookup plugin main class '''
 
-    # Reason: disable unused-argument because Ansible is calling us with many
-    #     parameters we are not interested in.
-    #     The lookup plugins of Ansible have this kwargs “catch-all” parameter
-    #     which is not used
-    # Status: permanently disabled unless Ansible API evolves
-    # pylint: disable=unused-argument
-    def __init__(self, basedir=None, **kwargs):
+    def __init__(self, basedir=None, **_):
         ''' Constructor '''
         self.basedir = basedir
 
-    # Reason: disable unused-argument because Ansible is calling us with many
-    #     parameters we are not interested in.
-    #     The lookup plugins of Ansible have this kwargs “catch-all” parameter
-    #     which is not used
-    # Status: permanently disabled unless Ansible API evolves
-    # pylint: disable=unused-argument
-    def run(self, terms, variables, **kwargs):
+    def run(self, terms, variables, **_):
         ''' Main execution path '''
 
         ret = []


sosiouxme:
I'm not really happy with the need to add symlinks everywhere for playbooks to enable the plugin. Why not put all the common plugins in an `openshift_common` role (with no tasks) and just have every playbook invoke that? It's not pretty either but it solves the problem once and for all.

sdodson:
oh man! I've never been this excited before! 99% of all the problems we see in upgrading the starter clusters are related to flaky yum commands.

ashcrow:
Will review shortly.

ashcrow:
Looks like travis caught a typo:

```
Syntax checking playbook: /home/travis/build/openshift/openshift-ansible/playbooks/byo/vagrant.yml

ERROR! no action detected in task. This often indicates a misspelled module name, or incorrect module path.

The error appears to have been in '/home/travis/build/openshift/openshift-ansible/roles/openshift_repos/tasks/main.yaml': line 9, column 5, but may

  block:

  - name: Ensure libselinux-python is installed

    ^ here

The error appears to have been in '/home/travis/build/openshift/openshift-ansible/roles/openshift_repos/tasks/main.yaml': line 9, column 5, but may

  block:

  - name: Ensure libselinux-python is installed

    ^ here
```

action_plugins/openshift_package.py
```diff
@@ -0,0 +1,98 @@
+"""
```
ashcrow: nit: Unless this is an ansible-ism generally docstrings occur after the license block.

action_plugins/openshift_package.py
```diff
@@ -0,0 +1,98 @@
...
+            if module not in self._shared_loader_obj.module_loader:
+                result['failed'] = True
+                result['msg'] = 'Could not find a module for %s.' % module
+            else:
+                # run the 'package' module
+                new_module_args = self._task.args.copy()
+                if 'use' in new_module_args:
+                    del new_module_args['use']
+
+                display.vvvv("Running %s" % module)
```
ashcrow: nit: Use current string formatting for later 2.x and 3+

```python
display.vvvv("Running {}".format(module))
```

action_plugins/openshift_package.py
```diff
@@ -0,0 +1,98 @@
...
+                        module_args=new_module_args,
+                        task_vars=task_vars,
+                        wrap_async=self._task.async,
+                    )
+                    tries += 1
+                    if tries > 3 or not res.get('failed'):
+                        result.update(res)
+                        break
+                    result['last_failed'] = res
+                    display.v("{} module failed on try {} with result: {}".format(module, tries, res))
```
ashcrow: 👍 for ``format`` and letting the user know how many tries it's attempted!!

action_plugins/openshift_package.py
```diff
@@ -0,0 +1,98 @@
...
+                tries = 0
+                while True:
+                    res = self._execute_module(
+                        module_name=module,
+                        module_args=new_module_args,
+                        task_vars=task_vars,
+                        wrap_async=self._task.async,
+                    )
+                    tries += 1
+                    if tries > 3 or not res.get('failed'):
```
ashcrow: Not required by any means but it may be worth noting via comment for future developers that ``result['last_failed'] = res`` can not be called on the last try else ansible will have a recursion error.

library/openshift_package.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+"""
```
ashcrow: nit: Unless this is an ansible-ism generally docstrings occur after the license block.

sosiouxme:
On Fri, Aug 18, 2017 at 9:54 AM, Stephen Milner <notifications@github.com>
wrote:

> Looks like travis caught a typo:
>
> Syntax checking playbook: /home/travis/build/openshift/openshift-ansible/playbooks/byo/vagrant.yml
>
> ERROR! no action detected in task. This often indicates a misspelled module name, or incorrect module path.
>
>
Not actually a typo; it's ansible not picking up the action plugin so it
doesn't recognize the action. I couldn't immediately see why but it just
underscores why the symlink-everywhere approach sucks... this is the error
you get when you miss one (or something else happens that prevents the
action plugin from loading).

library/openshift_package.py
```diff
@@ -0,0 +1,64 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+ANSIBLE_METADATA = {'metadata_version': '1.0',
+                    'status': ['stableinterface'],
+                    'supported_by': 'core'}
```
ashcrow: nit: I realize this is a slightly modified copy but this is technically inaccurate. I believe it should be ``community``.

library/openshift_package.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+DOCUMENTATION = '''
+---
+module: openshift_package
+version_added: 2.0
+author:
+    - Ansible Inc
```
ashcrow: nit: Not sure if this should be updated or not.

library/openshift_package.py
```diff
@@ -0,0 +1,64 @@
...
+short_description: Generic OS package manager
+description:
+     - Installs, upgrade and removes packages using the underlying OS package manager.
```
ashcrow: 'Installs, upgrade and removes packages using the underlying OS package manager. Attempts up to 3 times on failure' or something like that.

ashcrow:
A helpful **future** enhancement for this action would be to add support for denoting how many times to try and how long to sleep between tries.

ashcrow:
> Not actually a typo; it's ansible not picking up the action plugin so it doesn't recognize the action. I couldn't immediately see why but it just underscores why the symlink-everywhere approach sucks... this is the error you get when you miss one (or something else happens that prevents the action plugin from loading).

Ah, good catch.

mtnbikenc:
@sosiouxme I don't care for the symlinking everywhere either.  I'd rather use the ansible.cfg file for this like we already do for callback_plugins.  We even have `roles_path` in there too so we could do away with all the `roles` symlinks as well.  But, **ALL** playbooks would need to be launched from the root of the repo in order for the ansible.cfg to take effect.  This does require the reliance on a properly configured ansible.cfg, however, we already want that anyway and we are not getting our desired defaults if we let the user launch a playbook from any directory.  These are considerations that are higher level than just this PR so a decision on this approach should not block the addition of this action plugin.

```
callback_plugins = callback_plugins/
action_plugins = action_plugins/  <== maybe add this and not use symlinks
roles_path = roles/
```
I've opened https://github.com/openshift/openshift-ansible/pull/5140 to test some of these ideas.

sosiouxme:
On Fri, Aug 18, 2017 at 11:35 AM, Russell Teague <notifications@github.com>
wrote:

> @sosiouxme <https://github.com/sosiouxme> I don't care for the symlinking
> everywhere either. I'd rather use the ansible.cfg file for this like we
> already do for callback_plugins. We even have roles_path in there too so
> we could do away with all the roles symlinks as well. But, *ALL*
> playbooks would need to be launched from the root of the repo in order for
> the ansible.cfg to take effect. This does require the reliance on a
> properly configured ansible.cfg, however, we already want that anyway and
> we are not getting our desired defaults if we let the user launch a
> playbook from any directory.
>

OK, I have some concerns with that approach but it's surely cleaner.

The difference is, if you miss a callback plugin, nothing really breaks. If
you miss an action plugin, then any task that invokes it comes back as a
syntax error, which is pretty terrible user experience.


> callback_plugins = callback_plugins/
> action_plugins = action_plugins/  <== maybe add this and not use symlinks
>
>
Sure, I can just nuke the symlinks commit :)


>
> roles_path = roles/
>
> I've opened #5140
> <https://github.com/openshift/openshift-ansible/pull/5140> to test some
> of these ideas.
>

Will discuss over there.

action_plugins/openshift_package.py
sosiouxme: Guess I'm a little confused what you mean here. `res` only ever gets placed into `result`, not itself; where would recursion occur?

sosiouxme:
I really cannot figure out why ansible isn't using the action plugin on `playbooks/byo/rhel_subscribe.yml` (and thus `vagrant.yml`). It works for the other playbooks in the same dir. Ansible bug? Something weird about the `rhel_subscribe` role?

michaelgugino:
Is there a reason we can't use ansible's built-in retry?

sosiouxme:
@michaelgugino we can certainly use retries/until and forgo the `openshift_package` plugin. I just thought it would be nice to have the logic in one central place instead of scattered throughout the roles.

action_plugins/openshift_package.py
ashcrow: The way that ansible houses it's result is via a single instance. If a result appends itself it causes a Python recursion error as something similar to:

```python
result['data'] = result['data'] = result['data'] = ...
```

In your code it specifically avoids this, but it's just an idea as it's an easy accident to make when adding features.

test/integration/openshift_health_checker/common.go
```diff
@@ -25,9 +27,26 @@ func (p PlaybookTest) Run(t *testing.T) {
...
+	// we need to run ansible-playbook with the repository base dir as cwd
+	_, thisFile, _, ok := runtime.Caller(0)
+	if !ok {
+		t.Fatal("could not determine base directory")
+	}
+	repoBase := path.Join(path.Dir(thisFile), "../../..")
+
+	// determine the playbook path relative to repoBase
+	playbook, err := filepath.Abs(p.Path)
+	if err != nil {
```
ashcrow: Just an idea: If you reuse this pattern (which is super common in Go) it may be worth functionizing it:

```golang
func failOnError(err Error):
    if err != nil {
        t.Fatal(err)
    }
```

sosiouxme:
Test failures are now all because either `action_plugin` or `library` aren't picked up in config and so the `openshift_package` plugin does not work. This requires fixing how the jobs run.

sosiouxme:
Will base this on changes from https://github.com/openshift/openshift-ansible/pull/4264 so adding action_plugins to the RPM is painless.

Also, since it will break too many things to require running from repo base with the `ansible.cfg`, will punt on that and continue to add symlinks everywhere.

lhuard1A:
Hello,

I’m wondering what kind of yum command flakiness issues are you guys experiencing.

We’re currently regularly have this kind of ansible failures:
```
TASK [openshift_facts : Ensure various deps are installed] ******************************************************************************************************
ok: [ose3-int-a-master-1.node.jawed-1a-eu-central-1.acs] => (item=iproute)
ok: [ose3-int-a-master-3.node.jawed-1a-eu-central-1.acs] => (item=iproute)
ok: [ose3-int-a-master-2.node.jawed-1a-eu-central-1.acs] => (item=iproute)
ok: [ose3-int-a-master-1.node.jawed-1a-eu-central-1.acs] => (item=python-dbus)
ok: [ose3-int-a-master-1.node.jawed-1a-eu-central-1.acs] => (item=python-six)
ok: [ose3-int-a-master-1.node.jawed-1a-eu-central-1.acs] => (item=PyYAML)
failed: [ose3-int-a-master-2.node.jawed-1a-eu-central-1.acs] (item=python-dbus) => {"failed": true, "item": "python-dbus", "msg": "Failure talking to yum: [Errno 2] No such file or directory: '/var/cache/yum/x86_64/7Server/rhel7-optional/gen/primary_db.sqlite'"}
ok: [ose3-int-a-master-1.node.jawed-1a-eu-central-1.acs] => (item=yum-utils)
ok: [ose3-int-a-master-2.node.jawed-1a-eu-central-1.acs] => (item=python-six)
ok: [ose3-int-a-master-2.node.jawed-1a-eu-central-1.acs] => (item=PyYAML)
```

I could understand flakiness caused by the network or by the health of the remote yum server, but I have no evidence that the yum error we get is linked to the remote server or a network glitch.

Is the `yum` error above something that is “known to happen” and supposed to be addressed by this PR?

michaelgugino:
@lhuard1A 

That error is typically due to some other process has a lock on the file; ie a yum process is already running.

This commit is due to network failures, mostly.

If you are having an issue you need help with, and you think it might be caused by ansible, please file an issue in this repository and we will try to assist.

PR #5128, opened 2017-08-18
Add independent registry auth support

michaelgugino:

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1316341

### DIFF
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -170,6 +170,14 @@ openshift_release=v3.6
 # modify image streams to point at that registry by setting the following to true
 #openshift_examples_modify_imagestreams=true
 
+# If oreg_url points to a registry requiring authentication, provide the following:
+#oreg_auth_user=some_user
+#oreg_auth_password='my-pass'
+# NOTE: oreg_url must be defined by the user for oreg_auth_* to have any affect.
+# oreg_auth_pass should be generated from running docker login.
+# To update registry auth credentials, uncomment the following:
+#oreg_auth_credentials_replace: True
+
 # OpenShift repository configuration
 #openshift_additional_repos=[{'id': 'ose-devel', 'name': 'ose-devel', 'baseurl': 'http://example.com/puddle/build/AtomicOpenShift/3.1/latest/RH7-RHOSE-3.0/$basearch/os', 'enabled': 1, 'gpgcheck': 0}]
 #openshift_repos_enable_testing=false
--- a/roles/openshift_master/defaults/main.yml
+++ b/roles/openshift_master/defaults/main.yml
@@ -19,3 +19,8 @@ r_openshift_master_os_firewall_allow:
 - service: etcd embedded
   port: 4001/tcp
   cond: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
+
+oreg_url: ''
+oreg_host: "{{ oreg_url.split('/')[0] if '.' in oreg_url.split('/')[0] else '' }}"
+oreg_auth_credentials_path: "{{ openshift.common.data_dir }}/.docker"
+oreg_auth_credentials_replace: False
--- a/roles/openshift_master/tasks/main.yml
+++ b/roles/openshift_master/tasks/main.yml
@@ -232,6 +232,22 @@
   - restart master controllers
   when: openshift_master_bootstrap_enabled | default(False)
 
+- name: Check for credentials file for registry auth
+  stat:
+    path: "{{oreg_auth_credentials_path }}"
+  when:
+  - oreg_auth_user is defined
+  register: master_oreg_auth_credentials_stat
+
+- name: Create credentials for registry auth
+  command: "docker --config={{ oreg_auth_credentials_path }} login -u {{ oreg_auth_user }} -p {{ oreg_auth_password }} {{ oreg_host }}"
+  when:
+  - oreg_auth_user is defined
+  - (not master_oreg_auth_credentials_stat.stat.exists or oreg_auth_credentials_replace) | bool
+  notify:
+  - restart master api
+  - restart master controllers
+
 - include: set_loopback_context.yml
   when:
   - openshift.common.version_gte_3_2_or_1_2
--- a/roles/openshift_node/defaults/main.yml
+++ b/roles/openshift_node/defaults/main.yml
@@ -21,3 +21,8 @@ r_openshift_node_os_firewall_allow:
 - service: Kubernetes service NodePort UDP
   port: "{{ openshift_node_port_range | default('') }}/udp"
   cond: "{{ openshift_node_port_range is defined }}"
+
+oreg_url: ''
+oreg_host: "{{ oreg_url.split('/')[0] if '.' in oreg_url.split('/')[0] else '' }}"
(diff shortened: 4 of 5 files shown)


michaelgugino:
I have tested the logic of the new statements to ensure the auth file is placed in the suggested location.  I have not yet tested to see if openshift will actually use this auth with a 3rd party registry.

roles/openshift_master/tasks/main.yml
```diff
@@ -218,6 +218,23 @@
   - restart master api
   - restart master controllers
 
+- name: Check for credentials file for registry auth
+  stat:
+    path: "{{ registry_auth_credentials_path }}"
+  when:
+  - registry_auth_user is defined
+  - openshift_examples_modify_imagestreams | bool
```
jianlinliu: Personally I think openshift_examples_modify_imagestreams have nothing with registry auth here.

roles/openshift_master/tasks/main.yml
```diff
@@ -218,6 +218,23 @@
...
+  register: master_registry_auth_credentials_stat
+
+- name: Create credentials for registry auth
+  command: "docker --config={{ registry_auth_credentials_path }} login -u {{ registry_auth_user }} -p {{ registry_auth_token }} {{ oreg_host }}"
```
jianlinliu: where should user define registry_auth_user and registry_auth_token? In inventory host file?

roles/openshift_master/tasks/main.yml
michaelgugino: You could be right.  Perhats registry_auth_user check is enough?

michaelgugino:
so, the code I committed executes as expected.  Also somewhat as expected, openshift still does not try to auth or consume those credentials in any way:

```
Aug 18 11:50:35 ip-172-18-13-138.ec2.internal origin-node[26303]: E0818 11:50:35.665096   26303 kuberuntime_manager.go:619] createPodSandbox for pod "docker-registry-1-deploy_default(579587e8-842b-11e7-bd06-0e81fb6035e4)" failed: rpc error: code = 2 desc = unable to pull sandbox image "registry.reg-aws.openshift.com:443/openshift3/ose-pod:v3.6.0": unauthorized: authentication required
```

michaelgugino:
Looks like the instructions from bugzilla were close.  I'm going to try to  update the location of the .docker/config.json to /root to see if that works.  On origin /root is the home dir.

roles/openshift_master/tasks/main.yml
jianlinliu: It is better to mention that in inventory/byo/hosts.ose.example, give some example to show user how to define registry_auth_user and registry_auth_token.

michaelgugino:
This is currently blocked on:  https://github.com/openshift/origin/pull/15880

roles/openshift_master/defaults/main.yml
```diff
@@ -19,3 +19,6 @@ r_openshift_master_os_firewall_allow:
 - service: etcd embedded
   port: 4001/tcp
   cond: "{{ groups.oo_etcd_to_config | default([]) | length == 0 }}"
+
+registry_auth_credentials_path: "/var/lib/{{ openshift.common.service_type }}/.docker"
```
sdodson: `registry_auth_credentials_path: "{{ openshift.common.data_dir }}/.docker"` instead? This should be pretty stable but that's the var we use elsewhere.

roles/openshift_master/tasks/main.yml
sdodson: Yeah I agree, only need the first condition.

michaelgugino:
How to test:

```
# Change the following to "True" to replace existing credentials on master/nodes.
registry_auth_credentials_replace: False
registry_auth_user: some_user
registry_auth_token: some_token

# For OpenShift registry, registry_auth_token can be generated as follows:
# oc login --config=~/.kube/example.com --username=myuser --password='mypass' https://api.example.com
# oc --config=~/.kube/example.com -t

# port is optional
oreg_host: my.reg.url:443
oreg_url: my.reg.url:443/openshift3/ose-${component}:${version}

openshift_examples_modify_imagestreams: True
```

Currently blocked on 3.6, but works on 1.5.

inventory/byo/hosts.ose.example
```diff
@@ -167,6 +167,14 @@ openshift_release=v3.6
 # modify image streams to point at that registry by setting the following to true
 #openshift_examples_modify_imagestreams=true
 
+# If oreg_url points to a registry requiring authentication, provide the following:
+oreg_host: example.com:443
+registry_auth_user: some_user
+registry_auth_token: 'my-token'
```
sdodson: Just to ensure commonality with most of our variables what about `openshift_registry_auth_host` `openshift_registry_auth_user` and `openshift_registry_auth_password` instead? Password over token because it's more generic, for our use case where it's an openshift registry it happens to be a token but elsewhere it may just be a password.

michaelgugino: how about oreg_* as it matches the bits that we're modifying?

oreg_auth_password seems fine.

sdodson: That's fine.

jianlinliu: Should not we get oreg_host from oreg_url? Seem like oreg_host is a little redundant.

michaelgugino: I thought about this.  I feel it should actually be the other way around, but that's outside the scope of this PR.

I suppose this would also work in something like this:
oreg_url: somehost.com:443/somedir/registry/openshift3/{{ component }}/{{ version }}
oreg_host: somehost.com:443/somedir/registry

What's the best way to parse oreg_url to derive oreg_host in both situations?
Better situation is this:
oreg_host = somehost.example.com
oreg_url = {{ oreg_host }}/{{ oreg_suffix }}

But that would probably break existing users and would require some refactoring.

sdodson: There's a few places that parse oreg_url to get registry host.

`registry_host: "{{ registry_url.split('/')[0] if '.' in registry_url.split('/')[0] else '' }}"`

michaelgugino: Yes, and if someone's registry is sitting at example.com/myregistry (ie, it's sitting behind a reverse proxy or something of that nature) then this will break for this purpose.  For example:  https://docs.docker.com/registry/recipes/nginx/#setting-things-up

I'm happy to change it, but I think we're introducing a bug, and I think it's bad practice to split variables rather than combine them.

inventory/byo/hosts.ose.example
```diff
@@ -167,6 +167,13 @@ openshift_release=v3.6
 # modify image streams to point at that registry by setting the following to true
 #openshift_examples_modify_imagestreams=true
 
+# If oreg_url points to a registry requiring authentication, provide the following:
+oreg_auth_user=some_user
+oreg_auth_password='my-pass'
```
sdodson: Lets comment these and make it clear that you *must* set oreg_url for this to work, otherwise we don't have a registry host to log into. otherwise LGTM.

sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
Needs a release-3.6 branch backport too

sdodson:
aos-ci-test

PR #5129, opened 2017-08-18
Fedora compat fixes for package version checks and dnf bug workaround

maxamillion:
Fix a couple things that don't work with Fedora:
- Don't attempt to version check components, they aren't going to match the hard coded RHEL package versions.
- Work around a [bug in dnf](https://bugzilla.redhat.com/show_bug.cgi?id=1199432)'s python API that throws an error when a package spec has a `*` character in the Release field

Once we're done with review, I'll squash commits.

### DIFF
--- a/playbooks/common/openshift-cluster/config.yml
+++ b/playbooks/common/openshift-cluster/config.yml
@@ -1,7 +1,7 @@
 ---
 # TODO: refactor this into its own include
 # and pass a variable for ctx
-- name: Verify Install Requirements
+- name: Verify Requirements
   hosts: oo_all_hosts
   roles:
   - openshift_health_checker
--- a/playbooks/common/openshift-cluster/config.yml
+++ b/playbooks/common/openshift-cluster/config.yml
@@ -1,14 +1,17 @@
 ---
 # TODO: refactor this into its own include
 # and pass a variable for ctx
-- name: Verify Requirements
+- name: Verify Install Requirements
   hosts: oo_all_hosts
   roles:
   - openshift_health_checker
   vars:
   - r_openshift_health_checker_playbook_context: install
   post_tasks:
-  - action: openshift_health_check
+
+  - name: Verify Requirements - EL
+    when: ansible_distribution != "Fedora"
+    action: openshift_health_check
     args:
       checks:
       - disk_availability
@@ -17,6 +20,12 @@
       - package_version
       - docker_image_availability
       - docker_storage
+  - name: Verify Requirements - Fedora
+    when: ansible_distribution == "Fedora"
+    action: openshift_health_check
+    args:
+      checks:
+      - docker_image_availability
 
 - include: ../openshift-etcd/config.yml
 
--- a/roles/openshift_excluder/tasks/install.yml
+++ b/roles/openshift_excluder/tasks/install.yml
@@ -6,19 +6,46 @@
 
   block:
 
-  - name: Install docker excluder
+  - name: Install docker excluder - yum
     package:
       name: "{{ r_openshift_excluder_service_type }}-docker-excluder{{ openshift_pkg_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) +  '*' }}"
       state: "{{ r_openshift_excluder_docker_package_state }}"
     when:
     - r_openshift_excluder_enable_docker_excluder | bool
+    - ansible_pkg_mgr == "yum"
 
-  - name: Install openshift excluder
+
+  # For DNF we do not need the "*" and if we add it, it causes an error because
+  # it's not a valid pkg_spec
+  #
+  # https://bugzilla.redhat.com/show_bug.cgi?id=1199432
+  - name: Install docker excluder - dnf
+    package:
+      name: "{{ r_openshift_excluder_service_type }}-docker-excluder{{ openshift_pkg_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }}"
+      state: "{{ r_openshift_excluder_docker_package_state }}"
+    when:
+    - r_openshift_excluder_enable_docker_excluder | bool
+    - ansible_pkg_mgr == "dnf"
+
+  - name: Install openshift excluder - yum
     package:
       name: "{{ r_openshift_excluder_service_type }}-excluder{{ openshift_pkg_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) + '*' }}"
       state: "{{ r_openshift_excluder_package_state }}"
     when:
     - r_openshift_excluder_enable_openshift_excluder | bool
+    - ansible_pkg_mgr == "yum"
+
+  # For DNF we do not need the "*" and if we add it, it causes an error because
+  # it's not a valid pkg_spec
+  #
+  # https://bugzilla.redhat.com/show_bug.cgi?id=1199432
+  - name: Install openshift excluder - dnf
+    package:
+      name: "{{ r_openshift_excluder_service_type }}-excluder{{ openshift_pkg_version | default('') | oo_image_tag_to_rpm_version(include_dash=True) }}"
+      state: "{{ r_openshift_excluder_package_state }}"
+    when:
+    - r_openshift_excluder_enable_openshift_excluder | bool
+    - ansible_pkg_mgr == "dnf"
 
   - set_fact:
       r_openshift_excluder_install_ran: True


playbooks/common/openshift-cluster/config.yml
```diff
@@ -17,6 +17,24 @@
...
+  post_tasks:
+  - action: openshift_health_check
+    args:
+      checks:
+      - disk_availability
+      - memory_availability
+      - package_availability
+      - docker_image_availability
+      - docker_storage
+    when: ansible_distribution != "Fedora"
```
sdodson: Not checking on Fedora is probably the right technical solution but I wonder if we should just forego these checks on origin installs where we have less ability to enforce standards around the OS and config and we're more willing to accept workable but not supported configs? @rhcarvalho @sosiouxme what do you say?

sdodson:
aos-ci-test

playbooks/common/openshift-cluster/config.yml
maxamillion: @sdodson - yeah, I think that makes sense. Folks running upstream bits "hot off the press" are going to have a higher potential to have versions outside the known test matrix. :+1: 

If we're waiting for others to chime in I'll hold off on a change, otherwise I'll get a fix in. Just let me know.

sdodson:
[merge]

maxamillion:
@sdodson I don't think the Jenkins failures were related to these changes. Please advise.

sdodson:
[merge]

sdodson:
It should get another round through when it gets to the top of the merge queue.

michaelgugino:
Please see my comment above before merging.  I believe there is a '!=' where there should be an '==' in the Fedora case.

Also, this looks like the role will execute twice against all hosts, only the post tasks are being filtered.

sdodson:
@michaelgugino yeah you're right

maxamillion:
thanks!

playbooks/common/openshift-cluster/config.yml
sosiouxme: `openshift_health_checker` doesn't run any tasks -- none are even available; it just defines plugins. So task(s) use the plugins after the role is included.

sosiouxme: I agree that it isn't too helpful to run checks for supportable resources on Fedora. It would be nice to have the package requirements correct for Fedora but if we're not going to commit to keeping up with that (and I don't feel like we will yet) then best just to leave `package_availability` out too (I think it already exempts itself in some cases?).

I think there is some value in applying the `docker_image_availability` check on Fedora, as that will let you know if your install will end up trying to use images that don't exist, and that seems pretty relevant to me.

There's no particular need to have separate plays here. So maybe:

```
 - name: Verify Install Requirements
   hosts: oo_all_hosts
   roles:
   - openshift_health_checker
   vars:
   - r_openshift_health_checker_playbook_context: install
   post_tasks:

   - name: Verify Requirements - EL
     when: ansible_distribution != "Fedora"
     action: openshift_health_check
     args:
       checks:
       - disk_availability
       - memory_availability
       - package_availability
       - package_version
       - docker_image_availability
       - docker_storage

   - name: Verify Requirements - Fedora
     when: ansible_distribution == "Fedora"
     action: openshift_health_check
     args:
       checks:
       - docker_image_availability
```

playbooks/common/openshift-cluster/config.yml
```diff
@@ -17,6 +17,24 @@
...
+  post_tasks:
+  - action: openshift_health_check
+    args:
+      checks:
+      - disk_availability
+      - memory_availability
+      - package_availability
+      - docker_image_availability
+      - docker_storage
+  when: ansible_distribution == "Fedora"
```
sosiouxme: Still has logic backward :)

This got attached to an outdated commit so repeating here:

I agree that it isn't too helpful to run checks for supportable resources on Fedora. It would be nice to have the package requirements correct for Fedora but if we're not going to commit to keeping up with that (and I don't feel like we will yet) then best just to leave `package_availability` out too (it already exempts itself anyway when yum is not in use).

I think there is some value in applying the `docker_image_availability` check on Fedora, as that will let you know if your install will end up trying to use images that don't exist, and that seems pretty relevant to me.

There's no particular need to have separate plays here. So maybe:

```
 - name: Verify Install Requirements
   hosts: oo_all_hosts
   roles:
   - openshift_health_checker
   vars:
   - r_openshift_health_checker_playbook_context: install
   post_tasks:

   - name: Verify Requirements - EL
     when: ansible_distribution != "Fedora"
     action: openshift_health_check
     args:
       checks:
       - disk_availability
       - memory_availability
       - package_availability
       - package_version
       - docker_image_availability
       - docker_storage

   - name: Verify Requirements - Fedora
     when: ansible_distribution == "Fedora"
     action: openshift_health_check
     args:
       checks:
       - docker_image_availability
```

maxamillion: Yeah, I missed the `package_availability` in my [add dnf support patch](https://github.com/openshift/openshift-ansible/pull/5101). I guess I could add that later but if there's not much value in it then I can just leave it. I'll make them a single play.

maxamillion: @sosiouxme Don't we still want memory and disk checks for Fedora?

sosiouxme: They're as much to let the admin know about support requirements as anything. Those with dev/test clusters mostly turn them off.

maxamillion: @sosiouxme FWIW, we're running in production on Fedora+Origin. :)

sosiouxme: Neat! Well I don't feel strongly either way. Nothing magical happens if your system doesn't meet the requirements, except that things probably won't run as well as you would like at scale, and Red Hat will consider you unsupported. Running checks and gating the install on it is a balancing act between annoying the user and informing them, so you have to consider whether the information is valuable to enough people to risk annoying the rest of them.

maxamillion: @sosiouxme fair point, I'll leave them out for Fedora.

rhcarvalho: Alternatively we could set the default required value to a low number, the benefit being preventing situations as described in this RFE https://github.com/openshift/openshift-ansible/issues/4413, that led to checking available space in some other paths that are actively used during the install.

To set the required value, one can set this ansible variable:

```yaml
vars:
    openshift_check_min_host_disk_gb: 0.1
```

playbooks/common/openshift-cluster/config.yml
```diff
@@ -1,22 +1,32 @@
...
-      - package_availability
-      - package_version
-      - docker_image_availability
-      - docker_storage
+- name: Verify Install Requirements
+ hosts: oo_all_hosts
+ roles:
+ - openshift_health_checker
+ vars:
+ - r_openshift_health_checker_playbook_context: install
```
rhcarvalho: Suggestion:

```yaml
...
roles:
- role: openshift_health_checker
  r_openshift_health_checker_playbook_context: install
...
```

See https://github.com/openshift/openshift-ansible/pull/5136/files#r134479196

playbooks/common/openshift-cluster/config.yml
```diff
@@ -1,22 +1,32 @@
...
-    args:
-      checks:
-      - disk_availability
-      - memory_availability
-      - package_availability
-      - package_version
-      - docker_image_availability
-      - docker_storage
+- name: Verify Install Requirements
+ hosts: oo_all_hosts
```
rhcarvalho: There is an indentation error here, needing to prepend an extra space here and in the rest of the block.

playbooks/common/openshift-cluster/config.yml
maxamillion: @rhcarvalho would that be a Fedora specific check or a check for all distros?

rhcarvalho: The check itself only runs on hosts in either `masters`, `nodes` or `etcd` groups, unless you specify otherwise like here we're using a `when` clause in Ansible. There is nothing distro-specific.

With that configuration in my previous comment, the check should ensure there is at least 0.1 GB free in `/var`, 1 GB in `/usr/local/bin` and 1 GB in `/tmp` (or equivalent temp dir).

Maybe requiring 1 GB for `/var` would even be fair, to avoid surprises related to no disk space left?

rhcarvalho: Note: the values for `/usr/local/bin` and `/tmp` are the default, could also be configured if desired.

The more comprehensive configuration would look like:

```yaml
vars:
  openshift_check_min_host_disk_gb:
    "/var":
      masters: 0.1
      nodes: 1
      etcd: 2
    "/usr/local/bin":
      masters: 0.1
      nodes: 1
      etcd: 2
```

Replacing `0.1`, `1`, `2` with any desired value.

(/cc @sosiouxme, now I realize that configuring the value for the temp dir depends on the value that `tempfile.gettempdir()` evaluates to, which is not really good UX)

maxamillion: @rhcarvalho I understand there's nothing distro specific about the check, what I meant was: Is this a change we want for all distros or just Fedora? Because if it's not Fedora specific, then it's out of scope for this Pull Request.

rhcarvalho: @maxamillion, right. My proposal is specific to Fedora, as an alternative to disabling the check altogether.

playbooks/common/openshift-cluster/config.yml
maxamillion: @rhcarvalho I'm not trying to sanitize or iterate on the tasks as they exists, I'm simply attempting to introduce Fedora compatibility. I think this change is outside the scope of this PR and should be done separately in the event it was deemed necessary for a revert, the Fedora compat would not be lost.

maxamillion:
@sdodson I also don't think this round of failures is related to this change. Is Jekins ever happy? :)

sosiouxme:
>
> configuring the value for the temp dir depends on the value that
> tempfile.gettempdir() evaluates to, which is not really good UX)
>

Yes, but failing halfway through an install because of a temp dir the user
has never heard of is not any better UX than failing up front...

sdodson:
flake on https://github.com/openshift/origin/issues/13975

sdodson:
I didn't realize that this was five commits. Since we need to do another round through both aos-ci-test and the merge queue can we squash this down to the minimum number of logical commits so we don't have unnecessary intermediate states.

maxamillion:
@sdodson rebased and squashed

sdodson:
aos-ci-test

maxamillion:
@sdodson what happened with this?

rhcarvalho:
@maxamillion there is a merge conflict in `playbooks/common/openshift-cluster/config.yml`. Please rebase against master to fix the conflict.

maxamillion:
@rhcarvalho yeah, I got that. Was curious why this was never merged before the conflict occurred.

sdodson:
Hard to tell, test logs are gone at this point. Lets rebase and try to get this merged.

maxamillion:
@sdodson I don't think that failure is related to this PR, can we re-kick the tests?

maxamillion:
@sdodson @michaelgugino please advise

maxamillion:
Rosie doesn't like me anymore ;)

sosiouxme:
Rosie is obsolete :) Now we have a new class of problems...

michaelgugino:
Barring the need to rebase again, I think we should push this through if the merge queue fails.  It's a small, isolated patch that has been outstanding for quite some time.  This should have merged weeks ago.

maxamillion:
@michaelgugino thanks! :+1:

PR #5136, opened 2017-08-18
Standardize openshift-checks initialization patterns

mtnbikenc:
Standardizing the pattern of starting entry point playbooks with:

```yaml
- include: ../openshift-cluster/initialize_groups.yml

- include: ../../common/openshift-cluster/std_include.yml
```
then including the necessary playbook from `common` to do the actual work.

`OSEv3` should not be used as a host group in `common` playbooks.

### DIFF
--- a/playbooks/byo/openshift-checks/health.yml
+++ b/playbooks/byo/openshift-checks/health.yml
@@ -1,3 +1,6 @@
 ---
 - include: ../openshift-cluster/initialize_groups.yml
+
+- include: ../../common/openshift-cluster/std_include.yml
+
 - include: ../../common/openshift-checks/health.yml
--- a/playbooks/common/openshift-checks/health.yml
+++ b/playbooks/common/openshift-checks/health.yml
@@ -1,10 +1,6 @@
 ---
-- include: ../openshift-cluster/std_include.yml
-  tags:
-  - always
-
 - name: Run OpenShift health checks
-  hosts: OSEv3
+  hosts: oo_all_hosts
   roles:
   - openshift_health_checker
   vars:
--- a/playbooks/byo/openshift-checks/pre-install.yml
+++ b/playbooks/byo/openshift-checks/pre-install.yml
@@ -1,3 +1,6 @@
 ---
 - include: ../openshift-cluster/initialize_groups.yml
+
+- include: ../../common/openshift-cluster/std_include.yml
+
 - include: ../../common/openshift-checks/pre-install.yml
--- a/playbooks/common/openshift-checks/pre-install.yml
+++ b/playbooks/common/openshift-checks/pre-install.yml
@@ -1,10 +1,6 @@
 ---
-- include: ../openshift-cluster/std_include.yml
-  tags:
-  - always
-
-- hosts: OSEv3
-  name: run OpenShift pre-install checks
+- name: run OpenShift pre-install checks
+  hosts: oo_all_hosts
   roles:
   - openshift_health_checker
   vars:


mtnbikenc:
aos-ci-test

mtnbikenc:
aos-ci-test

mtnbikenc:
[test]

sdodson:
[merge]

playbooks/byo/openshift-checks/health.yml
```diff
@@ -1,3 +1,10 @@
 ---
 - include: ../openshift-cluster/initialize_groups.yml
+  tags:
+  - always
+
+- include: ../../common/openshift-cluster/std_include.yml
```
sosiouxme: The other one isn't from `common` it's still specific to `byo`.

playbooks/byo/openshift-checks/health.yml
```diff
@@ -1,3 +1,10 @@
 ---
 - include: ../openshift-cluster/initialize_groups.yml
+  tags:
+  - always
```
sosiouxme: Running `ansible-playbook ... --tags=foo` fails with our plays/roles. It's never tested in CI. I doubt there will ever be a use case for it except perhaps in development with playbooks of very limited scope.

There may be some value in running `ansible-playbook ... --skip-tags=foo`; however in that case `tags: always` still seems unnecessary. I too would like to know if there's some reason to keep this that I'm not seeing.

sosiouxme:
I have a task to follow up with using the common group names in our python modules too.

playbooks/byo/openshift-checks/health.yml
rhcarvalho: There are two includes from `common`:
- `../../common/openshift-cluster/std_include.yml`
- `../../common/openshift-checks/health.yml`

sosiouxme: Oh, hm; right. So, the question is would it be better to just put in `common/openshift-checks/health.yml`

    - include: ../openshift-cluster/std_include.yml

This pattern would help in the adhoc playbook.

playbooks/byo/openshift-checks/health.yml
mtnbikenc: Tags are used in various places throughout our playbooks so for consistency in entry point playbook configuration I added them here.  Usage of tags works in some playbooks and others it doesn't so ultimately we should either fix tag usage or remove them altogether.  To answer your specific questions, using `tags: always` allows someone to run the playbook specifying a tag or tags to run a certain subsection of the playbooks, but will always run the plays/tasks with the `always` tag.  Since these playbooks are part of initialization, we always want them to run.

playbooks/byo/openshift-checks/health.yml
mtnbikenc: We have discussed where to put std_include.yml.  I originally wanted to put them in the `common` playbooks for just the reason you are mentioning, however as we walked through the componentization of the common playbooks, we found it was better to move that initialization up.  This would allow us to group different common components together without each of them having to run the initialization section over and over again.  I can explain it further if desired.

playbooks/common/openshift-checks/health.yml
```diff
@@ -1,14 +1,9 @@
...
-
 - name: Run OpenShift health checks
-  hosts: OSEv3
+  hosts: oo_all_hosts
   roles:
-  - openshift_health_checker
-  vars:
-  - r_openshift_health_checker_playbook_context: health
+  - role: openshift_health_checker
+    r_openshift_health_checker_playbook_context: health
```
mtnbikenc: I'm not sure I understand your first sentence.  The play has a `vars:` section and a `roles:` section.  The `vars:` apply to the entire play.  If this play were to have more roles or other sections added in the future, such as 'pre-tasks:`, `tasks:`, then it could be unclear as to where those vars actually need to be specified.  Putting the vars with the call for the role makes it clear that they need to be set for the role.

mtnbikenc:
Flake https://github.com/openshift/origin/issues/13975 Panic in Spec Setup

mtnbikenc:
[merge]

playbooks/byo/openshift-checks/health.yml
rhcarvalho: > using `tags: always` allows someone to run the playbook specifying a tag or tags to run a certain subsection of the playbooks, but will always run the plays/tasks with the `always` tag. Since these playbooks are part of initialization, we always want them to run.

There is a common property between the `openshift-check` playbooks involved here and others such as [playbooks/byo/openshift-cluster/config.yml](https://github.com/openshift/openshift-ansible/blob/b107f677a3563f6db9a1ef17fe10518d321d495c/playbooks/byo/openshift-cluster/config.yml#L1-L14): they all have 3 plays and all of them are includes. The first two plays/includes are related to initialization, while the third is the "main part" that includes a similarly named playbook under `common`.

We add `tags: always` to the first two includes, but NOT to the third one.

IIUC, the effect is that using the `--tags` flag to `openshift-ansible` causes the initialization to run (taking a non-trivial amount of seconds/minutes), and then possibly nothing else runs. In the case of `openshift-check`, certainly nothing else runs. I do not understand how that is a desirable behavior.

sosiouxme: Until recently, running `ansible-playbook playbooks/byo/config.yml --tags foo` would get you an early error on a role include that tried to use some fact that hadn't been set in a previous task (that was skipped due to not being tagged - it was actually a more complicated chain of events). That no longer breaks after the recent refactoring but still it seems unlikely to me that anyone is actually running playbooks with `--tags` because it did not work at all for a long time, and since we're not testing the use of `--tags` in CI it could easily stop working with any small change. So cargo-culting `tags: always` everywhere to me just seems misleading, if we can't certify that you can actually use the `--tags` feature with that playbook and don't have a known use case for it.

mtnbikenc: Agreed.  We need to either fix and test the usage of tags in the repo or we need to remove all usage of tags.  I'll bring it up on the Arch Meeting for consideration.

mtnbikenc: And one other note, we have users running playbooks with using `--tags`.  I've seen bugs related to it not working properly so it definitely needs to be addressed.

playbooks/common/openshift-checks/health.yml
rhcarvalho: I tested this pattern using the `adhoc.yml` playbook from #4570 and found out that it breaks the intended behavior.

We use this variable to communicate in a rather unorthodox way with the callback plugin to print a summary of errors.
For the solution to work, the variable must be visible to the action plugin call in the `post_tasks` section.
When I set the variable only for the role include, the callback plugin comes with `context=None`.

mtnbikenc:
@rhcarvalho, If the variable needs to be used outside of the role, then the variable is not properly named.  We should not be declaring role style (`r_<role_name>`) variables and then use them in other places.  @kwoodson, what's your take on this?

rhcarvalho:
@mtnbikenc, the action plugin (`openshift_health_check`) is part of the role. After loading the role we execute the action plugin in the `post_tasks` block. The action plugin must be able to see that variable.

If the name pattern is incorrect, please let us know what would be the correct pattern and we'll fix it.

kwoodson:
@rhcarvalho @mtnbikenc, this is interesting.  I'm not sure I fully grasp what the expected behavior is doing.  Using a variable to persist data through phases is a new concept to me and I'm not sure I understand why its being used this way.  Is there something specific that needs to happen when `post_tasks` is called?  If there is some external knowledge in how the `openshift_health_checker` should work, outside being external to the role, then I'd say we have to discuss that.  

My impression of the health checker is that it takes a context, e.g. health, validation, etc, then performs a list of tasks pertaining to that context.  If it is 'health' then the health checker should perform a list of dynamic or statically defined tasks,  report results, and take necessary action (exit/continue).

Maybe some elaboration on the need to pass between playbook phases is required.

mtnbikenc:
@rhcarvalho The pattern we are trying to follow is that any variable needed inside of a role is passed directly to that role when it is called.  In the same way, any variable that is needed by a plugin should be passed into that plugin when it is called.  The action_plugin is defined within the role, but it can be called from anywhere in any playbook and is not called from inside the context of the role.  After looking further at the role, I see there are no tasks in the role and the `r_openshift_health_checker_playbook_context` variable was not used, so it is not needed by the role itself, only the action_plugin.  As I understand, the role is being called to load the the python library, action_plugins and callback_plugins.  Then the action_plugin is called in a post_task to do the actual work.

Here are some examples of what I'm trying to convey:
```yaml
---
- name: run OpenShift pre-install checks
  hosts: oo_all_hosts
  roles:
  - role: openshift_health_checker
  post_tasks:
    args:
      context: pre-install
      checks: ['@preflight']
```
```yaml
---
- name: Verify Host Requirements
  hosts: oo_all_hosts
  roles:
  - openshift_health_checker
  post_tasks:
    args:
      context: upgrade
      checks:
      - disk_availability
      - memory_availability
```

Digging further I see that there are quite a bit of direct access to global variables from within the checks.
```python
        is_containerized = self.get_var("openshift", "common", "is_containerized")
```

mtnbikenc:
I've updated this PR to remove the changes for the use of play `vars:` and role vars.  The variable usage changes can be handled in a later PR.

sosiouxme:
> In the same way, any variable that is needed by a plugin should be passed
> into that plugin when it is called. The action_plugin is defined within the
> role, but it can be called from anywhere in any playbook and is not called
> from inside the context of the role.
>

There are no hooks at the time an action_plugin is defined, only when it is
actually run. So how can a variable be passed in at the time the role is
invoked? There's nothing to pass it into at that time.


> Digging further I see that there are quite a bit of direct access to
> global variables from within the checks.
>
>         is_containerized = self.get_var("openshift", "common", "is_containerized")
>
>
Sure; most of this was written prior to all this discussion of role
variables, and frankly we're still waiting for the outcome to be settled
because yes, the role looks at a lot of outside variables, and I'd rather
refactor once-ish.

mtnbikenc:
aos-ci-test

mtnbikenc:
I've updated the PR to remove the tags per the discussion on the arch call. :smile:

sdodson:
[merge]

mtnbikenc:
aos-ci-test

mtnbikenc:
@rhcarvalho Any concerns left on this PR?

rhcarvalho:
@mtnbikenc LGTM, thanks! For #4570 there is a tricky bit which is running a play before the include of `std_include.yml`... it will go a bit off the pattern but we can discuss it there if necessary.

sosiouxme:
"No package matching 'atomic-openshift-docker-excluder-3.6.173.0.5*' found available, installed or updated" is a flake

mtnbikenc:
aos-ci-test

mtnbikenc:
[merge]

PR #5140, opened 2017-08-18
[WIP] Remove symlinks for Ansible components

mtnbikenc:
Removing all the symlinks for Ansible components and referencing them using the `ansible.cfg` file.  This will require running all playbooks from the repo root or ensure the user has a properly configured `ansible.cfg` file in the directory they are running `ansible-playbook`.  We want the user to have these default settings anyway to run callback_plugins and have tweaks for performance improvements.


mtnbikenc:
aos-ci-test

playbooks/common/openshift-master/restart.yml
```diff
@@ -7,7 +7,7 @@
     openshift_master_ha: "{{ groups.oo_masters_to_config | length > 1 }}"
   serial: 1
   handlers:
-  - include: roles/openshift_master/handlers/main.yml
+  - include: ../../../roles/openshift_master/handlers/main.yml
```
mtnbikenc: I don't like that we are pulling handlers out of roles like this, but that is another issue to address.

michaelgugino:
+1, you read my mind.

mtnbikenc:
I expected CI to have a hard time with this.

mtnbikenc:
[test]

mtnbikenc:
A few things to take into consideration when trying to implement something like this.  We don't really want to require users to be in the repo root when they are running playbooks.  Using the `ANSIBLE_CONFIG` environment variable to set the location of the ansible.cfg file could work, however the plugin and library paths are not relative to the ansible.cfg file.  Oddly, the roles_path path IS relative to the ansible.cfg file.  In order to get this to function in CI, we would need to set the `ANSIBLE_CONFIG` env var as a complete path AND add `${ANSIBLE_CONFIG}` to the beginning of each of the plugin paths and library path.  There are some issues in ansible/ansible regarding making paths defined in ansible.cfg relative paths so there may be something we would work with Ansible developers.  Ultimately, it would be nice to not have to symlink everywhere just to get things to work.

```bash
export ANSIBLE_CONFIG=/complete/path/to/openshift-ansible
```
```
# In ansible.cfg
callback_plugins = ${ANSIBLE_CONFIG}/callback_plugins/
library = ${ANSIBLE_CONFIG}/library/
lookup_plugins = ${ANSIBLE_CONFIG}/lookup_plugins/
filter_plugins = ${ANSIBLE_CONFIG}/filter_plugins/
roles_path = roles/
```
It would be better if the paths in ansible.cfg were relative instead of having to include env vars.

michaelgugino:
It is customary in ansible projects to execute plays inside the directory of the project.  It's also customary for roles to be inside the playbooks directory.  This behavior should be expected and encouraged.  All things should be either child directories of the root of the project, child directories of playbooks, or installed in normal ansible locations (eg roles in /etc/ansible/roles)

michaelgugino:
@sosiouxme 

I'm against dropping ansible.cfg into /etc/ansible.  We certainly can't claim that file as our own system-wide.  There may be other ansible projects installed on the same deployment host.

Adding small notes to the documentation regard where the playbooks should be run from is trivial.

`If installing from dnf/yum, please cd into /usr/share... and run ..., otherwise, cd into the cloned source and run ...  This is because we have modified ansible.cfg to support this product.`

ashcrow:
Agreed. Pushing the config on users system wide would be problematic. In the container version of the installer I believe we use a flag to specify the ansible config and we default to the one provided from our repo.

sosiouxme:
On Mon, Aug 21, 2017 at 3:33 PM, Michael Gugino <notifications@github.com>
wrote:
>
> Adding small notes to the documentation regard where the playbooks should
> be run from is trivial.
>
>
True. It's getting users to read them and understand what's happening when
they ignore them that's hard :-D

Nevertheless it does seem like the best course of action.

sosiouxme:
If we're going to tell RPM users to `cd /usr/share/ansible/openshift-ansible && ansible-playbook ...` then we should probably put `ansible.cfg` in `/usr/share/ansible/openshift-ansible` with one of the RPMs. Currently `atomic-openshift-utils` puts `utils/etc/ansible.cfg` (which is different) in `/usr/share/atomic-openshift-utils/ansible.cfg` and it's not clear to me how plugins are referenced at all.

sosiouxme:
You will want my latest commit on https://github.com/openshift/openshift-ansible/pull/5125/commits ("fix integration tests").

Other than that the CI jobs all need to be changed as they're running the playbooks from RPMs, while sitting in the jobs directory and configured with /etc/ansible/ansible.cfg.

Finally, if we get past that and merge this, I'm pretty sure it will totally break QE's automation which is also RPM-based.

mtnbikenc:
Closing this for now.  We'll revisit removing symlinks once we default to the containerized installer.

PR #5141, opened 2017-08-18
Backport to 1.5 - Bug 1471322: logging roles based image versions

wozniakjan:
Allowing to specify an image version for each logging component

https://bugzilla.redhat.com/show_bug.cgi?id=1471322

cc: @jcantrill , @portante

### DIFF
--- a/roles/openshift_logging/defaults/main.yml
+++ b/roles/openshift_logging/defaults/main.yml
@@ -1,4 +1,15 @@
 ---
+openshift_logging_curator_image_prefix: "{{ openshift_logging_image_prefix | default(__openshift_logging_image_prefix) }}"
+openshift_logging_elasticsearch_image_prefix: "{{ openshift_logging_image_prefix | default(__openshift_logging_image_prefix) }}"
+openshift_logging_fluentd_image_prefix: "{{ openshift_logging_image_prefix | default(__openshift_logging_image_prefix) }}"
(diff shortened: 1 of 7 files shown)


roles/openshift_logging/defaults/main.yml
```diff
@@ -1,4 +1,15 @@
 ---
+openshift_logging_curator_image_prefix: "{{ openshift_logging_image_prefix | default(__openshift_logging_image_prefix) }}"
```
ewolinetz: will `__openshift_logging_image_prefix` be set correctly here? I think we may need to load it in in the main task instead

wozniakjan: it appears to be set correctly to `docker.io/openshift/` when testing the `origin` deployment type and setting neither `openshift_logging_image_prefix`, nor `openshift_logging_*_image_prefix` vars. 

What leads you to think that `__openshift_logging_image_prefix` won't be set correctly? Is it [variable precedenece](http://docs.ansible.com/ansible/latest/playbooks_variables.html), because should apply in our favour in this scenario. Or is it something else I may be missing?

ewolinetz: interesting, i didn't think that would have a value... i guess the default variable was lazy loaded after we had already loaded variables from the correct `var/` file. Excellent!

wozniakjan:
@sdodson  may I kindly ask you to merge this backport of https://github.com/openshift/openshift-ansible/pull/5044? I was told this is wanted in the following release. Thanks!

sdodson:
sure np

PR #5142, opened 2017-08-18
Ensure that openshift_node_facts has been called for dns_ip

sdodson:


sdodson:
@csrwng this may fix the issue you're seeing

sdodson:
aos-ci-test

sdodson:
@csrwng did this help? locally it works for me with and without the change

csrwng:
So tried patching the change into the 3.6 branch and of course didn't work. Is it ok to use master to install a 3.6 cluster?

csrwng:
Output:

```
ERROR! the role 'openshift_node_facts' was not found in /home/azureuser/openshift-ansible/playbooks/byo/openshift-cluster/../../common/openshift-cluster/../openshift-node/roles:/etc/ansible/roles:/home/azureuser/openshift-ansible/roles:/home/azureuser/openshift-ansible/playbooks/byo/openshift-cluster/../../common/openshift-cluster/../openshift-node

The error appears to have been in '/home/azureuser/openshift-ansible/roles/openshift_node_dnsmasq/meta/main.yml': line 16, column 3, but may

- role: openshift_common
- role: openshift_node_facts
  ^ here
```

sdodson:
Yeah, master branch should work for 3.6 installs.

csrwng:
Trying that now

csrwng:
@sdodson so this looks good. I got past the error in my deployment. Now I get another one about not having an etcd host, will try to sort it out.

sdodson:
[merge]

sdodson:
Merge would be blocked by the fact that this needs to be fixed on both master and release-3.6 branches. #5147 to fix on release-3.6 branch

jhadvig:
The issue is still reproducable... Our install_upgrade job is failing due to it 
https://ci.openshift.redhat.com/jenkins/job/test_pull_request_origin_extended_conformance_install_update/4460/consoleFull#39147885258b6e51eb7608a5981914356

sdodson:
@jhadvig that job used openshift-ansible-3.6.173.0.7-1.git.0.f6f19ed.el7 which is from before this merged even though the job ran after it. Any idea why?

jhadvig:
@sdodson that version we are using for installation of origin. and then we are upgrading to the latest version.
The idea of that job is to install the latest available version of previous minor release (in this case 3.6.x) and upgrade to the last minor version release (in this case 3.7.x)

PR #5149, opened 2017-08-21
Updating default tag for enterprise installation for ASB

ewolinetz:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1479165
We should be setting the image tag for the ASB as a release rather than using `latest`


sdodson:
In the interest of making master work for multiple releases we probably need to figure out a way to make this work across v3.6 and v3.7 but I'm not sure what we should do in the immediate term. v3.7 versions of this image do exist in our internal registries.

sdodson:
aos-ci-test

sdodson:
[merge]

PR #5156, opened 2017-08-22
hot fix for env variable resolve

mangirdaz:
If we use environment variables in our inventory files (and from what I seen we do this everywhere where We deploy OCP) our fact engine ignores env variables so if my path looks like
```
openshift_hosted_registry_routecertificates={"certfile": "{{inventory_dir}}/../files/certs/wildcard.registry.company.local.crt", "keyfile": "{{inventory_dir}}/../files/certs/wildcard.registry.companylocal.key", "cafile":"{{inventory_dir}}/../files/certs/CompanyLocalRootCA.crt"} 
openshift_hosted_registry_routehost=containers.registry.comany.local
```
the result is: `/../files/certs/RoSLocalRootCA.crt`

We need to fix our fact set in a long run to read Ansible variables. And it was done in the same way with router certificates already.

### DIFF
--- a/roles/openshift_hosted/tasks/registry/secure.yml
+++ b/roles/openshift_hosted/tasks/registry/secure.yml
@@ -1,7 +1,7 @@
 ---
 - name: Configure facts for docker-registry
   set_fact:
-    openshift_hosted_registry_routecertificates: "{{ ('routecertificates' in openshift.hosted.registry.keys()) | ternary(openshift.hosted.registry.routecertificates, {}) }}"
+    openshift_hosted_registry_routecertificates: "{{ ('routecertificates' in openshift.hosted.registry.keys()) | ternary(openshift_hosted_registry_routecertificates, {}) }}"
     openshift_hosted_registry_routehost: "{{ ('routehost' in openshift.hosted.registry.keys()) | ternary(openshift.hosted.registry.routehost, False) }}"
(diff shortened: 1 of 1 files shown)


mangirdaz:
redeploy certificates is broken after refactoring registers part in https://github.com/openshift/openshift-ansible/pull/4254 . If this is fine with guys @sdodson @dmsimard will align redeploy part here too.

sdodson:
@abutcher can you take a look and also make sure we didn't break cert re-deploy in #4254

abutcher:
@sdodson my custom registry/routehost certs are replaced by a generated cert via cert redeploy so that will need to be fixed.

abutcher:
We don't have any logic in `playbooks/common/openshift-cluster/redeploy-certificates/registry.yml` for `openshift_hosted_registry_routecertificates`.

abutcher:
@mangirdaz This looks fine though :+1: -- is the fix for redeploy similar?

mangirdaz:
IT will be similar. Will add changes to same pull request and let you know. latest on weekend as need to test I dont break anything :)

mangirdaz:
@abutcher will not fix redeploy here...  need to "rebase" my client to the new code and do more testing as it's not straight forward.  Will do it in a separate branch and raise an issue with detail description. I should upgrade clients clusters in next week or two and will need to fix re-deploy for them too (it's not yet fixed upstream)

mangirdaz:
@abutcher Ping

abutcher:
aos-ci-test

abutcher:
aos-ci-test

PR #5157, opened 2017-08-22
[proposal] role dependency orchestration

ingvagabund:
Reshape the way we orchestrate role dependencies

### DIFF
--- a/docs/proposals/role_dependency_orchestration.md
+++ b/docs/proposals/role_dependency_orchestration.md
@@ -1,0 +1,171 @@
+# Role dependency orchestration
+
+## Towards loosely-coupled roles
+
+Assume the following situation:
+* role `<role>_common` defines variables that are common for various other roles
+  that deal with upgrade, certificate deployment, migration, and other operations.
+* each of corresponding `<role>_upgrade`, `<role>_migrate`, `<role>_certificates` roles depend
+  on `<role>_common` role and consumes its role variables
+
+Assume we have a play that upgrades a cluster component in the following sequence:
+
+* re-deploy certificates by invoking `<role>_certificates` role
+* upgrade component by invoking `<role>_upgrade` role
+* migrate data of a component to new schema by invoking `<role>_migrate`
+
+I.e.
+```yaml
+roles:
+- role: <role>_certificates
+- role: <role>_upgrade
+- role: <role>_migrate
+```
+
+What are disadvantages of this approach:
+* `<role>_common` role is invoked three times even if it needs to be invoked only once
+  to set common variables.
+* if the `<role>_upgrade` role needs only a small set of common variables,
+  invoking `<role>_common` becomes an expensive operation.
+* there may be various sources that produces proper values for `<role>_upgrade`
+  role variables. It the `<role>_common` is invoked each time right before
+  the upgrade role, it overrides all the common role variable no matter what they
+  were set to before the invocation. Thus decreasing role modularity and re-usability.
+
+## Role dependencies
+
+A role can have two categories of dependencies:
+
+* self-contained dependency: a dependency that has no role variable to set/read
+* parametrized dependency: a dependency that has a set of input role variables
+
+In the latter case there are two ways how to propagate role variables into role dependency:
+
+* with a knowledge of role variables of role dependencies:
+
+  ```yaml
+  roles:
+  - role: etcd
+    r_etcd_client_port: 2379
+    r_etcd_peer_port: 2380
+    r_etcd_common_etcd_runtime: docker
+    r_etcd_common_embedded_etcd: false
+  ```
+
+  Role variables of both roles (`etcd` and `etcd_common`) are set during `etcd`
+  role invocation. Caller of the `etcd` role must be aware of all relevant
+  `r_etcd_common` like variables of the `etcd_common` role.
+
+* without a knowledge of role variables of role dependencies:
+
+  ```yaml
+  roles:
+  - role: etcd
+    r_etcd_peer_port: 2380
+    r_etcd_etcd_runtime: docker
+    r_etcd_embedded_etcd: false
+  ```
+
+  All `r_etcd_common_` like variables are duplicated and wired into their
+  corresponding dependency variables in `etcd/meta/main.yml`:
+
+  ```yaml
+  dependencies:
+  - role: etcd_common
+    r_etcd_common_etcd_runtime: "{{ r_etcd_etcd_runtime }}"
+    r_etcd_common_embedded_etcd: "{{ r_etcd_embedded_etcd }}"
+  ```
+
+Both ways have its advantages and disadvantages. In either way, the `etcd` role
+depends on the `etcd_common` role.
+
+Aim of the proposal is to eliminate role dependencies wherever it is possible
+in favor of applying the following approach:
+
+```yaml
+- include_role:
+    name: etcd_common
+  vars:
+    r_etcd_common_etcd_runtime: docker
+    r_etcd_common_embedded_etcd: false
+
+- include_role:
+    name: etcd
+  vars:
+    r_etcd_peer_port: 2380
+    r_etcd_etcd_runtime: docker
+    r_etcd_runtime: "{{ r_etcd_common_etcd_runtime }}"
+    r_etcd_service: "{{ r_etcd_common_service }}"
+    r_etcd_service_file: "{{ r_etcd_common_service_file }}"
+```
+
+Which resuls in:
+
+* `etcd_common` dependency of `etcd` role is removed
+* all `r_etcd_common_` like variables needed in the `etcd` role are wired
+  through `r_etcd_` like variables
+* both role are handled as functions with a set of parameters and return values
+* each role deals with its own set of role variables, one is independent of the other
+
+In general, if a role A depends on role B both roles are orchestrated via
+the following pattern:
+
+```yaml
+- include_role:
+    name: A
+  vars:
+    r_A_<var>: <value>
+    r_A_<var>: <value>
+    ...
+
+- include_role:
+    name: B
+  vars:
+    r_B_<var>: <value>
+    r_B_<var>: <value>
+    ...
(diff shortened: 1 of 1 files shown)


michaelgugino:
I (somewhat) agree with the method of consuming another role's variables.  It's proper to refer to them externally rather than explicitly consuming them inside the role directly.

This is an example of two roles that should be combined into one.  Something like the default etcd port doesn't need to be shipped around all over the place.  We can just set that in role defaults.

In fact, everywhere that needs 'etcd_port' should probably just define that in defaults as the same variable name.  It's not always helpful or desirable to give ever variable a unique name.  Just calling it etcd_port in each role defaults main.yml allows users to override the port by changing one line in their inventory / group_vars.  If for some reason master's need a different etcd port than nodes, that's what group_vars are for.

ingvagabund:
> This is an example of two roles that should be combined into one. Something like the default etcd port doesn't need to be shipped around all over the place. We can just set that in role defaults.

Disagree. The `etcd_common` is a dependency of at least 4 other etcd like roles. To eliminate duplication of default values all the etcd common values are set in `etcd_common`. Another thing is if all etcd like roles should be merged into one. We have

* etcd_ca
* etcd_client_certificats
* etcd_migrate
* etcd_upgrade
* etcd

that currently depends on `etcd_common`. I am not against revisiting all the mentioned roles for consideration to be merged into one `etcd` role that has the following actions:

* deploy_ca
* deploy_client_certificates
* install
* upgrade
* migrate

The `etcd_migrate` is already action based role with additional actions:

* check
* configure
* add_ttls
* clean_data

However, we can get to a point where we have a list of 10 or more actions which a role can do. I would not like to exceed some reasonable number of actions per role. So a natural way to overcome it is to split the actions into categories like (with additional actions @sosiouxme mentioned):
* certificates: deploy_ca, deploy_client_certificates
* installation: install, pre-check, post-check
* upgrades: upgrade, pre-check, post-check
* migration: migrate, check, configure, add_ttls, clean_data

And introduce a concept of sub-action. We could even do this:
```yaml
roles:
- role: etcd
  r_etcd_action: migrate
  r_etcd_subaction: add_ttls
````

But it is only a matter of time before we start doing this (yeah papa, I am exaggerating the situation):

```yaml
roles:
- role: etcd
  r_etcd_action: migrate
  r_etcd_subaction: add_ttls
  r_etcd_subsubaction: ???
  r_etcd_subsubsubaction: ???
  ...
````

I will leave it for the audience to decide it that is something we want or not.

So to some extend it make sense to have a two or three roles that depend on `<role>_common` role that sets the most important and default variables.

**Comment by ingvagabund** [comment]
> In fact, everywhere that needs 'etcd_port' should probably just define that in defaults as the same variable name. It's not always helpful or desirable to give ever variable a unique name. Just calling it etcd_port in each role defaults main.yml allows users to override the port by changing one line in their inventory / group_vars. If for some reason master's need a different etcd port than nodes, that's what group_vars are for.

**Comment by ingvagabund** [comment]
> In fact, everywhere that needs 'etcd_port' should probably just define that in defaults as the same variable name. It's not always helpful or desirable to give ever variable a unique name. Just calling it etcd_port in each role defaults main.yml allows users to override the port by changing one line in their inventory / group_vars. If for some reason master's need a different etcd port than nodes, that's what group_vars are for.

That's a good point. Though, if done it is no too much different for reading a non-role (e..g inventory) variable inside a role as we have now.

**Comment by michaelgugino** [comment]
> That's a good point. Though, if done it is no too much different for reading a non-role (e..g inventory) variable inside a role as we have now.

It's exactly how variables are supposed to be used in the inventory: to override role defaults and provide undefined values (eg, passwords).  We want to enable users to be able to override as few variables as possible to get the desired outcome.  In the case of something like a standard port number, that variable can be defined as etcd_port in a variety of places, and the user only has to make one change in group_vars/all.yml or in the inventory under an appropriate group to override that value.

As soon as you go down the path of
```
vars:
  - my_role2_var1 = "{{ my_role1_var1 }}"
```
they might as well be the same variable name altogether.  Most likely they should both have the same default value.  If something other than the default needs to be set at some point, you'll be able to set it in once place and have all the desired results.

> And introduce a concept of sub-action. We could even do this:
This is no longer necessary, as @sosiouxme proposed in another thread, I think include_role is the better path forward.  Then we don't need to add the logic in various places.  This method will also help us clean up and consolidate large role-sets like etcd into one or two roles.

**Comment by ewolinetz** [comment]
Is there any advantage to using multiple roles that are granular to what they do vs calling a single role that would have logic contained within it to know when its appropriate to upgrade or install?

To me this seems like is extra overhead to maintain multiple roles for a single 'role'. We should be able to maintain installing multiple versions within a single role and if we know that we have a specific version already installed and we are told to install a different version we could figure out that an 'upgrade' is being asked for. That seems like an easier UX, we should just say we want a specific version installed and the role should do all that is required for us (instead of us checking if we need to do an upgrade or an install)

**Comment by ingvagabund** [comment]
>  In the case of something like a standard port number, that variable can be defined as etcd_port in a variety of places, and the user only has to make one change in group_vars/all.yml or in the inventory under an appropriate group to override that value.

Both `openshift_master` and `openshift_node` roles have a set of shared variables:
* `openshift.common.is_containerized`
* `openshift.common.data_dir`
* `openshift.common.sdn_network_plugin_name`
...

Once both roles are refactored (with a good practise of naming role variables) they will end up like:
* for master:
  - `openshift_master_is_containerized`
  - `openshift_master_data_dir`
  - `openshift_master_sdn_network_plugin_name`
* for node:
  - `openshift_node_is_containerized`
  - `openshift_node_data_dir`
  - `openshift_node_sdn_network_plugin_name`

We already have `containerized` (and possible other) inventory variable that if set, set `openshift.common.is_containerized` to `True`. In order to propagate the inventory variable into both roles:
* we either wire the `openshift.common.is_containerized` into both `openshift_master_is_containerized` and `openshift_node_is_containerized` in a play or when invoking a role
* or we rename both `openshift_master_is_containerized` and `openshift_node_is_containerized` into `openshift_is_containerized` and pretend it is a role variable as well so it can be overwritten by group/inventory vars.

I don't see any other way around to override master/node role variables without breaking the principles in https://github.com/openshift/openshift-ansible/pull/5093 proposal and at the same time complying with one of the Ansible basic principles of overriding role defaults via group/inventory vars.

Even if we merge all `etcd_` like roles into one, we will still end up with at least two roles that consumes `etcd_port` variable, i.e. `etcd` and `master`. Currently, the master consumes `openshift.master.etcd_port` which is basically `etcd_port` just dressed up in a summer coat.

So the
```yaml
- include_role:
    name: A
  vars:
    r_A_<var>: <value>
    r_A_<var>: <value>

- include_role:
    name: B
  vars:
    r_B_<var>: "{{ r_A_<var> }}"
    r_B_<var>: "{{ r_A_<var> }}"
```
pattern is unavoidable with the current state of the art.

However, in overall I agree with you and we should minimize the pattern as much as we can.

ingvagabund:
> Is there any advantage to using multiple roles that are granular to what they do vs calling a single role that would have logic contained within it to know when its appropriate to upgrade or install?

As long as the number of actions within a role is reasonable, there is not. On the other hand it would be a shame to merge two roles into one "just because" they look semantically similar or share a small set of variables. Basically, `etcd` and `etcd_migrate` roles do not share any common tasks or templates. They only share a set of variables from the `etcd_common` role. Not saying they should be merged, just saying we should not merge anything that starts with `etcd_` prefix into a single role just because it feels right.

> To me this seems like is extra overhead to maintain multiple roles for a single 'role'. We should be able to maintain installing multiple versions within a single role and if we know that we have a specific version already installed and we are told to install a different version we could figure out that an 'upgrade' is being asked for. That seems like an easier UX, we should just say we want a specific version installed and the role should do all that is required for us (instead of us checking if we need to do an upgrade or an install)

Wrt. install vs. update or genereate vs. re-generate I agree. For that reason I can wait once node install and node upgrade roles merges. The same for the (master install, master upgrade) and (etcd install, etcd upgrade) pieces. Once done we can then revisit the concept of the install/upgrade decision logic.

michaelgugino:
> I don't see any other way around to override master/node role variables without breaking the principles in #5093 proposal and at the same time complying with one of the Ansible basic principles of overriding role defaults via group/inventory vars.

Here's the cleanest solution that allows variable linking and inventory overrides:
```
# openshift_master/defauts/main.yml
openshift_all_is_containerized: False
openshift_master_is_containerized: "{{ openshift_all_is_containerized }}"

# openshift_node/defaults/main.yml
openshift_all_is_containerized: False
openshift_node_is_containerized: "{{ openshift_all_is_containerized }}"
```

This lets users override the 'all' version, or each individual version at their choosing.  Of course, group_vars could be used as well.  In fact, you could use group_vars to have nodes containerized and not containerized in the same inventory if you really wanted.

michaelgugino:
> As long as the number of actions within a role is reasonable, there is not. On the other hand it would be a shame to merge two roles into one "just because" they look semantically similar or share a small set of variables. Basically, etcd and etcd_migrate roles do not share any common tasks or templates. They only share a set of variables from the etcd_common role. Not saying they should be merged, just saying we should not merge anything that starts with etcd_ prefix into a single role just because it feels right.

I think a good way to determine are the files being edited, and services being started.  Example include config files, service unit files, environment files.  Especially if the roles are intended to run on the same major host groups.

ingvagabund:
> Here's the cleanest solution that allows variable linking and inventory overrides:
```yaml
# openshift_master/defauts/main.yml
openshift_all_is_containerized: False
openshift_master_is_containerized: "{{ openshift_all_is_containerized }}"
```
The `openshift_all_is_containerized` is no longer a role variable by definition.

What we could do is to limit the use of non-role variables to `defaults/main.yml` and state: "Hey, besides the role API you can use this list of `non-role` variables to override the role variables from within an inventory/group vars". Of course, each non-role variable inside the `default/main.yml` would have to be well-documented.

kwoodson:
closing this as I don't think its relevant now.  repopen if you feel the need to.

PR #5163, opened 2017-08-22
[1.5] Add independent registry auth support

michaelgugino:

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1316341
(cherry picked from commit 5815311)
(cherry picked from commit 57abb2f)

### DIFF
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -120,6 +120,14 @@ openshift_release=v3.5
 # modify image streams to point at that registry by setting the following to true
 #openshift_examples_modify_imagestreams=true
 
+# If oreg_url points to a registry requiring authentication, provide the following:
+#oreg_auth_user=some_user
+#oreg_auth_password='my-pass'
+# NOTE: oreg_url must be defined by the user for oreg_auth_* to have any affect.
+# oreg_auth_pass should be generated from running docker login.
+# To update registry auth credentials, uncomment the following:
+#oreg_auth_credentials_replace: True
+
 # Additional yum repos to install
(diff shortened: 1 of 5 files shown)


inventory/byo/hosts.ose.example
```diff
@@ -120,6 +120,14 @@ openshift_release=v3.5
 # modify image streams to point at that registry by setting the following to true
 #openshift_examples_modify_imagestreams=true
 
+# If oreg_url points to a registry requiring authentication, provide the following:
+oreg_host=example.com:443
```
jianlinliu: Personally I think oreg_host is a little redundant here, we could get the value form oreg_url, maybe we could reuse the logic - https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_examples/defaults/main.yml#L28, no need to ask user to input the oreg_host separately again.

roles/openshift_master/tasks/main.yml
```diff
@@ -164,6 +164,22 @@
...
+  register: master_oreg_auth_credentials_stat
+
+- name: Create credentials for registry auth
+  command: "docker --config={{ oreg_auth_credentials_path }} login -u {{ oreg_auth_user }} -p {{ oreg_auth_password }} {{ oreg_host }}"
+  when:
+    - oreg_auth_user is defined
+    - (not master_oreg_auth_credentials_stat.stat.exists or oreg_auth_credentials_replace) | bool
+  notify:
+    - restart master api
+    - restart master controllers
```
sdodson: Unfortunately for 3.6 and older we need to make sure to notify `restart master` for non HA installs and these two for HA installs.

michaelgugino: Done.

sdodson:
aos-ci-test

michaelgugino:
aos-ci-test

PR #5164, opened 2017-08-22
[1.4] Add independent registry auth support

michaelgugino:

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1484068
(cherry picked from commit e3338f37b37a34573d315ebbbd8df85df2eca50e)
(cherry picked from commit f5be106abcb87ad685ad2ace6120d6525631bc32)
backports #5128

### DIFF
--- a/inventory/byo/hosts.ose.example
+++ b/inventory/byo/hosts.ose.example
@@ -99,6 +99,14 @@ openshift_release=v3.4
 # modify image streams to point at that registry by setting the following to true
 #openshift_examples_modify_imagestreams=true
 
+# If oreg_url points to a registry requiring authentication, provide the following:
+#oreg_auth_user=some_user
+#oreg_auth_password='my-pass'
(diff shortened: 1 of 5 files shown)


inventory/byo/hosts.ose.example
```diff
@@ -99,6 +99,14 @@ openshift_release=v3.4
 # modify image streams to point at that registry by setting the following to true
 #openshift_examples_modify_imagestreams=true
 
+# If oreg_url points to a registry requiring authentication, provide the following:
+oreg_host=example.com:443
```
jianlinliu: Personally I think oreg_host is a little redundant here, we could get the value form oreg_url, maybe we could reuse the logic - https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_examples/defaults/main.yml#L28, no need to ask user to input the oreg_host separately again.

sdodson:
aos-ci-test

michaelgugino:
aos-ci-test

PR #5165, opened 2017-08-22
Flatten docker into one role

michaelgugino:
This commit is intended to flatten the various docker roles
into a single docker role and remove all meta-depends of the
various docker roles in all other roles.

This commit also modifies other roles to not have meta-depends
on the docker role.  This requires that docker role be run
specifically on all host groups that require docker.  The
docker role is now called directly during setup of masters,
nodes, and etcd hosts.


playbooks/common/openshift-cluster/config.yml
```diff
@@ -42,6 +42,13 @@
     r_openshift_excluder_action: disable
     r_openshift_excluder_service_type: "{{ openshift.common.service_type }}"
 
+- name: Setup Docker
+  hosts: oo_all_hosts
```
abutcher: We won't have to configure docker on all hosts, such as on loadbalancer/haproxy or etcd hosts (when not containerized).

michaelgugino: I am very much open to suggestion on much of this PR.  Could you suggest an alternative?

michaelgugino:
Removes/Refactors:  15 unnecessary/redundant set_facts

Eliminates 2 redundant roles and their dependent chains:  openshift_docker, openshift_docker_facts

Removes openshift_facts.py openshift.docker entirely (123 LOC!)

Should be 100% drop-in for existing deploys; we might need to relocate where the docker role is called and for what host groups.

playbooks/common/openshift-cluster/config.yml
abutcher: I don't know how to crack this one offhand. It would be nice to do with one play so maybe we can create a single host group of all containerized hosts or create `oo_X_containerized` groups for containerized hosts and merge them together for this play. We'll know which hosts are containerized by the `openshift.common.is_containerized` fact from `openshift_facts` or just `containerized` from inventory.

* All node hosts.
* Containerized master, etcd, glusterfs, loadbalancer and nfs hosts.

roles/openshift_loadbalancer/templates/haproxy.docker.service.j2
```diff
@@ -1,7 +1,7 @@
 [Unit]
-After={{ openshift.docker.service_name }}.service
-Requires={{ openshift.docker.service_name }}.service
-PartOf={{ openshift.docker.service_name }}.service
+After={{ openshift_docker_service_name }}.service
```
michaelgugino: openshift_docker_service_name is one variable that needs to have a fact set.  This is because we cannot assume 'docker' here, and the user does not set a variable directly for this value.

roles/openshift_node/tasks/openvswitch_system_container.yml
```diff
@@ -1,14 +1,11 @@
 ---
 - set_fact:
-    l_use_crio: "{{ openshift_docker_use_crio | default(false) }}"
```
michaelgugino: setting this value is totally redundant, so we'll just consume openshift_docker_use_crio directly.  Also, consuming this variable instead of openshift.docker.use_crio is inconsistent with how most of the other variables are consumed, so bug fixed by virtue of standardizing how we consume variables passed in from users.

roles/openshift_node/tasks/openvswitch_system_container.yml
```diff
@@ -1,14 +1,11 @@
...
-
-- set_fact:
     l_service_name: "cri-o"
-  when: l_use_crio
+  when: openshift_docker_use_crio
```
michaelgugino: This could probably be handled in role defaults as well, but we'll keep it for now.

roles/openshift_node/templates/openshift.docker.node.service
```diff
@@ -21,7 +21,7 @@ EnvironmentFile=/etc/sysconfig/{{ openshift.common.service_type }}-node-dep
...
...n --name {{ openshift.common.service_type }}-node --rm --privileged --net=host --pid=host --env-file=/etc/sysconfig/{{ openshift.common.service_type }}-node -v /:/rootfs:ro,rslave -e CONFIG_FILE=${CONFIG_FILE} -e OPTIONS=${OPTIONS} -e HOST=/rootfs -e HOST_ETC=/host-etc -v {{ openshift.common.data_dir }}:{{ openshift.common.data_dir }}{{ ':rslave' if openshift.docker.gte_1_10 | default(False) | bool else '' }} -v {{ openshift.common.config_base }}/node:{{ openshift.common.config_base }}/node {% if openshift_cloudprovider_kind | default('') != '' -%} -v {{ openshift.common.config_base }}/cloudprovider:{{ openshift.common.config_base}}/cloudprovider {% endif -%} -v /etc/localtime:/etc/localtime:ro -v /etc/machine-id:/etc/machine-id:ro -v /run:/run -v /sys:/sys:rw -v /sys/fs/cgroup:/sys/fs/cgroup:rw -v /usr/bin/docker:/usr/bin/docker:ro -v /var/lib/docker:/var/lib/docker -v /lib/modules:/lib/modules -v /etc/origin/openvswitch:/etc/openvswitch -v /etc/origin/sdn:/etc/openshift-sdn -v /var/lib/cni:/var/lib/cni -v /etc/systemd/system:/host-etc/systemd/system -v /var/log:/var/log -v /dev:/dev $DOCKER_ADDTL_BIND_MOUNTS -v /etc/pki:/etc/pki:ro {{ openshift.node.node_image }}:${IMAGE_VERSION}
```
michaelgugino: the openshift.docker.gte_1_10 is a left over from previous releases.  We need at least 1_12 now in other places, so this check is not necessary.

roles/openshift_node_upgrade/README.md
```diff
@@ -20,7 +20,6 @@ From this role:
 | docker_version                 |                       | Optional inventory var                                 |
 | l_docker_upgrade               |                       |                                                        |
 | node_config_hook               |                       |                                                        |
-| openshift.docker.gte_1_10      |                       |                                                        |
```
michaelgugino: No longer needed.

roles/openshift_version/tasks/set_version_containerized.yml
```diff
@@ -1,7 +1,4 @@
 ---
-- set_fact:
-    l_use_crio: "{{ openshift_docker_use_crio | default(false) }}"
```
michaelgugino: This right here is what I call a 'bad practice'

playbooks/common/openshift-cluster/config.yml
michaelgugino: I like the suggestion.  Yeah, I'm not sure how best to implement that at this point without skipping lots of tasks or creating new meta-groups.

playbooks/common/openshift-cluster/openshift_hosted.yml
```diff
@@ -47,7 +47,7 @@
     openshift_hosted_logging_elasticsearch_ops_pvc_prefix: "{{ 'logging-es-ops' if openshift_hosted_loggingops_storage_kind | default(none) =='dynamic' else '' }}"
 
   - role: cockpit-ui
-    when: ( openshift.common.version_gte_3_3_or_1_3  | bool ) and ( openshift_hosted_manage_registry | default(true) | bool ) and not (openshift.docker.hosted_registry_insecure | default(false) | bool)
+    when: ( openshift.common.version_gte_3_3_or_1_3  | bool ) and ( openshift_hosted_manage_registry | default(true) | bool ) and not (openshift_docker_hosted_registry_insecure | default(false) | bool)
```
ashcrow: Eventually we should change these to be multiline ... doesn't have to happen in this PR.

ashcrow:
I like where this is going.

roles/openshift_node/templates/openshift.docker.node.service
kwoodson: @michaelgugino, I'm a big fan if not the biggest fan of removing cruft.  According to some meetings in the recent past, we were required to support some older versions.

@sdodson, can you verify that we can get rid of the gte_1_10?

Just a safety check here.

kwoodson:
@michaelgugino, I really like this refactor.  Creating default variables and normalizing variables across the roles reduces duplication, dependencies, and a lot tasks.  I like breaking apart openshift_facts and reducing its footprint and its coupling to the roles.

Thanks for looking into this work.

roles/openshift_node/templates/openshift.docker.node.service
sdodson: Yeah safe to remove, we'd never use this codebase on anything older than 3.6 which requires 1.12.

ashcrow:
Rebase is needed here. After which we can kick off testing again.

roles/openshift_version/tasks/set_version_containerized.yml
ashcrow: Could be, but in fairness it is what was previously considered the way to use local facts in this code base.

kwoodson:
@michaelgugino, let's rebase and get this in!

michaelgugino:
aos-ci-test

sdodson:
aos-ci-test

sdodson:
```
09:37:50 TASK [docker : Set registry params] ********************************************
09:37:50 task path: /home/jenkins/workspace/openshift-ansible-test-matrix/CONTAINERIZED/_NOT_containerized/OSE_VER/3.6/PYTHON/System-CPython-2.7/TOPOLOGY/openshift-cluster/TargetBranch/master/nodes/openshift-ansible-slave/openshift-ansible/roles/docker/tasks/package_docker.yml:54
09:37:50 fatal: [10.8.168.251]: FAILED! => {
09:37:50     "failed": true, 
09:37:50     "msg": "|failed expects first param is a list"
09:37:50 }
09:37:50 	to retry, use: --limit @/home/jenkins/workspace/openshift-ansible-test-matrix/CONTAINERIZED/_NOT_containerized/OSE_VER/3.6/PYTHON/System-CPython-2.7/TOPOLOGY/openshift-cluster/TargetBranch/master/nodes/openshift-ansible-slave/openshift-ansible/playbooks/byo/config.retry
09:37:50 
09:37:50 PLAY RECAP *********************************************************************
09:37:50 10.8.168.247               : ok=35   changed=4    unreachable=0    failed=0   
09:37:50 10.8.168.25                : ok=35   changed=4    unreachable=0    failed=0   
09:37:50 10.8.168.251               : ok=125  changed=45   unreachable=0    failed=1   
09:37:50 localhost                  : ok=11   changed=0    unreachable=0    failed=0   
```

michaelgugino:
@sdodson Yeah, I'm looking into this.  I think I have something for that.

michaelgugino:
aos-ci-test

michaelgugino:
aos-ci-test

sdodson:
Once tests pass, please wait to merge this until the start of the next sprint.

ashcrow:
CI is throwing 502's. Failures likely to CI issues.

ashcrow:

mtnbikenc:
I get a failure when running a test install:
```
TASK [docker : Set registry params] **********************************************************************************************************
task path: /home/rteague/dev/clusters/aws-c1/openshift-ansible/roles/docker/tasks/package_docker.yml:60
fatal: [ec2-34-207-136-96.compute-1.amazonaws.com]: FAILED! => {
    "failed": true
}

MSG:

|failed expects first param is a list
```

inventory/byo/hosts.ose.example
```diff
@@ -98,9 +98,9 @@ openshift_release=v3.7
 # Add additional, insecure, and blocked registries to global docker configuration
 # For enterprise deployment types we ensure that registry.access.redhat.com is
 # included if you do not include it
-#openshift_docker_additional_registries=registry.example.com
-#openshift_docker_insecure_registries=registry.example.com
-#openshift_docker_blocked_registries=registry.hacker.com
+#openshift_docker_additional_registries=['registry.example.com']
+#openshift_docker_insecure_registries=['registry.example.com']
+#openshift_docker_blocked_registries=['registry.hacker.com']
```
sdodson: Is this conversion to a list mandatory? We can't force people to change the format unless there's no way around it.

michaelgugino: We discussed this in IRC a bit.

It seems these examples were in error.  Everything, including our unit tests, expects this to be a list.  This should probably be a list, it doesn't appear to support CSV strings at this time either.

sdodson: In my test inventory I've always had the following, I'll run through a quick test with this PR and my standard inventory.

`openshift_docker_additional_registries="registry1.redhat.com,registry2.redhat.com:8888"`

sdodson:
```
TASK [docker : debug] *************************************************************************************************************************************************************************************************************************
Tuesday 19 September 2017  17:40:15 -0400 (0:00:00.077)       0:00:25.384 ***** 
ok: [ose3-master.example.com] => {
    "msg": "insecure_registries registry.ops.openshift.com,brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888"
}

TASK [docker : Set registry params] ***********************************************************************************************************************************************************************************************************
Tuesday 19 September 2017  17:40:15 -0400 (0:00:00.077)       0:00:25.461 ***** 
fatal: [ose3-master.example.com]: FAILED! => {"failed": true, "msg": "|failed expects first param is a list"}
```

This code was responsible for splitting it when it was provided as a comma separated list https://github.com/openshift/openshift-ansible/pull/5165/files#diff-8c83287a125220bc46f58c79f26c6670L2275

michaelgugino: @sdodson I added some new variables in roles/docker/defaults/main.yml to safely convert any csv values (include just a single item) to a list.  If the values is already a list, it will remain as-is.

sdodson:
I mentioned this on IRC but given the size and scope of this PR when reviewing please commit to performing a smoke test with whatever inventories you generally test your work with. These are great changes that really clean up the code base but the risk of regression is pretty high so I'm asking for extra scrutiny here.

mtnbikenc:
Still having an issue which could just be related to my local test cluster.  The excluders are enabled preventing openshift_version from finding an available atomic-openshift package.  Continuing to look into this.

mtnbikenc:
Since we've confirmed there is an issue with the excluders (which was not caught by CI), putting this on hold until we have an update.
/hold

sdodson:
@ingvagabund this is the pr where we're having excluders problems

ingvagabund:
> Still having an issue which could just be related to my local test cluster. The excluders are enabled preventing openshift_version from finding an available atomic-openshift package. Continuing to look into this.

IMHO, the inventory variable `openshift_release` is set to a proper version so the CI passes as there is no need to detect the version. When the variable is not set (which is our case), it is detected from `atomic-openshift` rpm. Given the installation is performed over vanilla new cluster, the rpm is not installed. Now, then this PR is applied, the excluders are installed before the `openshift_version` role is run (which was not happening before the PR). When the `openshift_version` role is run, it installs the `atomic-openshift` rpm. However, instead of installing the latest rpm, it will not install anything (or install the older rpm) as the excluders are excluding. So the `openshift_version` is not detected (or is set to older OCP version).

ingvagabund:
One should enable excluders in the `playbooks/common/openshift-cluster/setup_docker.yml` only when the `{{ openshift.common.service_type }}` rpm is not yet installed during the installation. Once it is installed the excluder can be enabled since you will have the correct version of the rpm installed and the `package` module will not fail.

PR #5167, opened 2017-08-23
Playbook to deploy keepalived on master nodes for native HA

bacek:
Just simple playbook to deploy keepalived on master node for HA clustering without external LoadBlanacer with round-robin DNS only.

Uses ansible-keepalived role from evrardjp/ansible-keepalived to deploy keepalived and couple of iptables rules to enable VRRP traffic.

### DIFF
--- a/playbooks/byo/master_keepalived.yml
+++ b/playbooks/byo/master_keepalived.yml
@@ -1,0 +1,128 @@
+### Playbook to provision keepalived on master nodes for native HA
+#
+# To configure create group keepalived in inventory file with next labels:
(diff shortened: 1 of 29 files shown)


sdodson:
@openshift/networking Should we use something like this or should we use ipfailover pods created by `oadm ipfailover`

bacek:
It's chicken and egg problem. Basically this is substitute for ```lb``` section for external loadbalancer needed to deploy HA master.

bacek:
Closing in favour of #5273. Creating PR from ```master``` branch wasn't a good idea.

PR #5170, opened 2017-08-23
use generic v3.4 and v3.3 tags for metrics and logging deployers

ruromero:
Users are having problems when upgrading using the deployer and having the x.x.0 version
 3.3.0 only gets upgraded to the latest 3.3.0-x version which is outdated. The same for 3.4


sdodson:
This is fine with me but if you want these changes in the 3.3 installer you need to update release-1.3 branch too. Is there an associated bug for this?

ruromero:
Yes, just wanted to have the approval for the 3.4 branch before.
Here is the BZ https://bugzilla.redhat.com/show_bug.cgi?id=1484398

PR #5172, opened 2017-08-23
[proposal] action-based roles

ingvagabund:
Consolidate roles operating under the same component to minimize code duplication and with a similar semantics by defining a concept of action based roles.

Roles can turn into analogy of https://docs.puppet.com/puppet/5.2/bgtm.html

### DIFF
--- a/docs/proposals/action_based_roles.md
+++ b/docs/proposals/action_based_roles.md
@@ -1,0 +1,183 @@
+# Action based roles
+
+## Description
+
+Some of the openshift-ansible roles share common resources and spread responsibility
+for a single action (or operation) among multiple roles. That leads to a code duplication,
+degradation of code usability, increase of maintenance efforts, etc.
+
+All such roles should be consolidated into a single role providing a mechanism
+to invoke a particular piece of a code.
+
+From here on, an action stands for implementation of an operation.
+
+## Rationale
+
+Each component lives through defined life-cycle phases. At the beginning
+a component is installed. As time passes new version of the component is
+available resulting in component upgrade. Under certain conditions one needs
+to carry additional operations to update successfully such as migration,
+backup or post-verification.
+
+Given each of the operations are applied on the same component, some of the operation
+underlying tasks, variables and other Ansible resources are shared. E.g. templates
+of service files, handlers, sequence of installation or configuration tasks.
+
+Additionally, some operations can be unified into one and the proper behavior hidden
+inside a role as implementation details. E.g. component upgrade can be equivalent
+to component installation up to some differences. The role dealing with the installation
+should then decide if the actual installation or rather an update is required.
+
+In practice, it is natural to create a role that deals with an installation as
+the first step. Later, role that deals with an upgrade is implemented. Not rarely
+followed by implementation of other operations. At the end one can end up
+with a set of roles, each responsible for a single operation. Naturally,
+the layout leads to code duplication and various efforts to minimize it.
+
+## Design
+In case a set of roles conforms to the following conditions:
+
+* each role is responsible for a distinct operation (e.g. installation, upgrade, migration)
+* all operations are applied on the same component
+* the operations share common code base and Ansible resources
+
+consolidate all roles into new one, providing a mechanism to select specific
+operation to apply when the new role is invoked. Since Ansible 2.2 one can
+use `include_role` module and perform file-based operation selection:
+
+```yaml
+include_role:
+  name: <role>
+tasks_from: <operation>
+```
+
+Additionally, one can define a minimal set of actions that each action-based
+role should provide:
+* pre-checks: to check if the role has all necessary conditions fulfilled
+* post-checks: to check a role ended successfully
+
+### Layout
+
+Proposed layout of an action-based role:
+
+```sh
+$ ls <role>/tasks
+<action1_dir>
+<action2_dir>
+...
+<action1>.yml
+<action2>.yml
+...
+```
+
+The role `tasks` directory consists of:
+* one directory per each action (action directory)
+* one yml file per each action (action entry-point)
+* additional yml files needed by the role (e.g. task files shared across actions)
+
+Templates and other resources location are not affected by the layout.
+If reasonable, the action directory layout can be duplicated inside the `templates` directory.
+
+**Expected content of an action file**:
+
+```yaml
+# content of tasks/<action>.yml
+
+# tasks common for actions, e.g. pre-checks, initialization
+- name: Include <common> tasks
+  include: <common>.yml
+...
+- <tasks>:
+    <task_param>: "{{ <value> }}"
+    ...
+...
+# sub-actions finishing implementation of the action
+- include: <action>/<subaction1>.yml
+- include: <action>/<subaction2>.yml
+```
+
+The pattern in the action file allows to:
+- further decompose the action into sub-actions (increase re-usability)
+- compose implementation of an action in the role tasks directory from common tasks and sub-actions
+- easy-to-understand tasks flow (a single entry-point for each action in the role tasks directory, no entry-point in the action directory)
+
+If an action directory was created from a previously existing role,
+it contains all the role's tasks files.
+If it was created as a new action, it contains all tasks files needed
+to implement the action.
+
+### Usage
+
+```yaml
+- hosts: <hosts>
+  tasks:
+  ...
+  - include_role:
+      name: <role>
+      tasks_from: <action>
+    vars:
+      <var1>: "{{ <value1> }}"
+      <var2>: "{{ <value2> }}"
+      ...
+```
+
+If the `tasks/main.yml` is present, `include_role` without the `tasks_from`
+can correspond to invocation of a default action.
+The `tasks/main.yml` can provide additional comments about expected usage:
+
+```yaml
+# This role is intended to be used with include_role. E.g.
+# include_role:
+#   name:  <role>
+#   tasks_from: "{{ item }}"
+# with_items:
+#   - pre_checks
+#   - install
+#   - verify
+```
+
+### Application of action-based roles
+
+#### Stratification
+
+A set of non-overlapping role actions can correspond to a set of well-defined phases.
+One can decide to first run the same phase across all roles and wait for completion
+before moving to run all the roles in a different phase. The phases can be broken out
+into the following categories:
+
+* **Informational** - determine values and set facts accordingly (either as local/private
+  variables or role variable outputs). In some plays (e.g. for checks) this might
+  be the only phase to run.
+* **Pre-validation** - determine whether a role has all it needs to run,
+  for example RPMs / images / tools on localhost, and whether the variables
+  provided actually make sense (are defined, well-formed, consistent, come from
+  an acceptable set, etc.) for the purpose of failing fast before any roles make any changes.
+* **Run** - make changes that the role is intended for
+* **Post-validation** - check whether the actions the role took actually worked;
+  for instance, if the role defines a router, check that the routing pod(s) are actually
+  running, so that we don't declare success and leave the user to figure out nothing
+  is actually running. This phase could run repeatedly until success, defined failure,
+  or timeout.
+
+## Checklist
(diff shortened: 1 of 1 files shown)


docs/proposals/action_based_roles.md
```diff
@@ -0,0 +1,115 @@
...
+```yaml
+...
+# run some checks
+- name: Fail if invalid r_<role>_action provided
+  fail:
+    msg: "<role> role can only be called with 'install', 'upgrade', 'migrate' or 'backup'"
+  when: r_<role>_action not in ['install', 'upgrade', 'migrate' or 'backup']
+...
+- name: Include main action task file
+  include: "{{ r_<role>_action }}.yml"
```
sosiouxme: I slightly prefer this method to using `tasks_from`. The advantage is that it makes clear which task files are intended as entry points. Could also enable logic that you want to apply to all entry points. And keeping a `tasks/main.yml` will help those who are new to the codebase.

I would hope this pattern could be abstracted into an action plugin instead of copy-pasted to every role.

ewolinetz:
I like this approach, but am not sure why we wouldn't just do away with the need to provide an action or at least change the default behavior is an action isn't provided (e.g. install if doesn't already exist, otherwise do an upgrade).

Maybe not all roles will be able to do that, but at least the roles where the components it configures have specific versions that we can look up should enable us to build in some business intelligence.

docs/proposals/action_based_roles.md
```diff
@@ -0,0 +1,115 @@
...
+## Proposal
+
+In case a set of roles conforms to the following conditions:
+
+* each role is responsible for a distinct operation (e.g. installation, upgrade, migration)
+* all operations are applied on the same component
+* the operations share common code base and Ansible resources
+
+consolidate all roles into new one, providing a mechanism to select specific
+operation to apply when the new role is invoked. Once can either go with
```
sosiouxme: Cool, I had not thought much about running different actions using the same role but it does make a lot of sense.

docs/proposals/action_based_roles.md
```diff
@@ -0,0 +1,115 @@
...
+Additionally, one can define a minimal set of actions that each action-based
+role should provide:
+* pre-checks: to check if the role has all necessary conditions fulfilled
+* post-checks: to check a role ended successfully
+
+### Layout
+
+Proposed layout an action-based role (with variable-based operation selection):
+
+**default/main.yml**
```
sosiouxme: Hmmm... what are the chances we'll want different defaults for the different actions/phases? That could get hairy. But maybe in practice it won't be needed.

docs/proposals/action_based_roles.md
```diff
@@ -0,0 +1,115 @@
...
+* variable-based operation selection:
+  ```yaml
+  include_role:
+    name: <role>
+  vars:
+    r_<role>_action: <operation>
+  ```
+
+* or with file-based operation selection:
+```yaml
```
michaelgugino: This is the best way forward.  include_role and tasks_from are relatively new to ansible, and thus that pattern isn't found much (if at all) in openshift-ansible.  If I were starting openshift-ansible from scratch today, this is exactly how I do it.  The benefits are clear and highly desirable.

docs/proposals/action_based_roles.md
```diff
@@ -0,0 +1,115 @@
...
+In case a set of roles conforms to the following conditions:
+
+* each role is responsible for a distinct operation (e.g. installation, upgrade, migration)
+* all operations are applied on the same component
+* the operations share common code base and Ansible resources
+
+consolidate all roles into new one, providing a mechanism to select specific
+operation to apply when the new role is invoked. Once can either go with
+
+* variable-based operation selection:
```
michaelgugino: This method is inferior to the second method, IMO.

docs/proposals/action_based_roles.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Layout
+
+Proposed layout an action-based role (with variable-based operation selection):
+
+**default/main.yml**
+```yaml
+...
+# description="action selector" choices=["install", "upgrade", "migrate", "backup"]
+r_<role>_action: install
```
michaelgugino: unnecessary if going with second method, include_role

docs/proposals/action_based_roles.md
```diff
@@ -0,0 +1,115 @@
...
+...
+```
+
+**tasks/main.yml**:
```
michaelgugino: unnecessary if going with second method, include_role

docs/proposals/action_based_roles.md
```diff
@@ -0,0 +1,115 @@
...
+In case the operation selection is file-based, the `tasks/main.yml` is removed
+and all relevant pre-checks included in all the action tasks files.
+
+### Application of action-based roles
+
+#### Stratification
+
+A set of non-overlapping role actions can correspond to a set of well-defined phases.
+One can decide to first run the same phase across all roles and wait for completion
+before moving to run all the roles in a different phase. The phases can be broken out
```
sosiouxme: It's possible you might have a matrix of actions and phases here, e.g. if install and upgrade require different pre-validation logic (as seems likely). But even when that complexity arises, it's hard to see any better way forward than keeping everything in the role to enable sharing common resources.

I was thinking of running phases in generic loops, but given the need to pass different variables to individual roles that's probably not practical anyway. So we would just need to call each role repeatedly for as many actions as are relevant. This could make for some long playbooks and the role docs will have to be very clear about what variables are required in which actions.

docs/proposals/action_based_roles.md
michaelgugino: In my mind, include_role, tasks_from more clearly treats the different task sets as completely independent parts.

I would say, in each role's main.yml, we leave a note "This role is meant to be used in such and such a way"

docs/proposals/action_based_roles.md
michaelgugino: I like the explicit calling of each phase in our playbooks.  Is it a little longer?  Yes, but it's clear what is happening.

docs/proposals/action_based_roles.md
michaelgugino: If you need a different default, it probably shouldn't be named the same thing across each phase.

docs/proposals/action_based_roles.md
sosiouxme: I think your preference is stronger than mine. It is simpler, so there's that.

ingvagabund:
> I am not sure why we wouldn't just do away with the need to provide an action or at least change the default behavior if an action isn't provided (e.g. install if doesn't already exist, otherwise do an upgrade).
> Maybe not all roles will be able to do that, but at least the roles where the components it configures have specific versions that we can look up should enable us to build in some business intelligence.

Yeah. As long as a role deals just with an installation&upgrades this would be the optimal implementation. Once a role can do more and additional decision logic is required out of the role, we can't avoid action selection.

docs/proposals/action_based_roles.md
ingvagabund: >  In my mind, include_role, tasks_from more clearly treats the different task sets as completely independent parts.
> I would say, in each role's main.yml, we leave a note "This role is meant to be used in such and such a way"

We could elevate that by a directory structure in a role:
* `tasks/<action>.yml` for all actions a role provides
* `tasks/_common/<file>.yml` remaining tasks files that are included in the `<action>.yml` files so a role caller can include only properly named action directly from the `tasks` directory.

If not `_common`, we can go with `_<file>.yml` instead to explicitly state `_` prefixed files are not for inclusion.

ingvagabund: >  I slightly prefer this method to using tasks_from. The advantage is that it makes clear which task files are intended as entry points. Could also enable logic that you want to apply to all entry points. And keeping a tasks/main.yml will help those who are new to the codebase.

Yeah, we already discussed that on arch meeting if we should go with variable based (via `r_<role>_action`) selection or with file (via `tasks_from`) based one. Each one has its advantages and disadvantages. I am in favour of the variable based as:
* you have a single entry point so you can run all checks at one place
* you can change names of underlying task files (you can decide `install.yml` is too boring and change it to `deploy.yml` is it more aligns with your thoughts still keeping the "install" action keyword unchanged)
* you don't have to specify the `.yml` suffix of a task file

ingvagabund:
The proposal updated. PTAL

michaelgugino:
I have thought about this a great deal.  I'm still strongly in favor of 'include_role' for the following reasons:

1) I don't want to have to figure out what variable is being set, and where it is being set, and how it is passed into the role, in order to figure out what is going to happen.

2) People will inevitably want to place more than just "- include {{ my_role_action }}" in main.yml.

3) This is what I feel most strongly about.  include_role is called as a task.  This allows more obvious task ordering in plays.  We can get rid of post-task and pre-task, and just have an easy to follow order of operations.

With that said, I'm not strictly opposed to using the 'action-variable' method either.  I think both have merits, it's just down to preference.

docs/proposals/action_based_roles.md
```diff
@@ -0,0 +1,156 @@
...
+If reasonable, the action directory layout can be duplicated inside the `templates` directory.
+
+**Expected content of an action file**:
+
+```yaml
+# content of tasks/<action>.yml
+
+# include a main file from action directory
+- name: Running <action>
+  include: <action>/main.yml
```
michaelgugino: I don't prefer to have a main.yml inside the action folder.

I would prefer the following:
```yaml
# tasks/<action-1>.yml
- include common_thing.yml
- some_task:
    some_param: {{ some_var }}
- include <action-1>/do_something.yml
- include <action-1>/do_next.yml
```

The reason being, for users, I don't want people to open a file that is simply a pass-through, and then have to open <action-1>/main.yml, which may or may not include some tasks, which then may include other files, possibly from the parent directory.  It's too much searching.

Putting your series of includes inside of tasks/action-1.yml allows a high-level overview of what that portion of the role (or as I've been calling them, pseudo-role) is doing, and allows importing of shared elements without having to do relative parent includes (eg. include ../common/some_thing.yml)

docs/proposals/action_based_roles.md
```diff
@@ -0,0 +1,156 @@
...
+  - include_role:
+      name: <role>
+      tasks_from: <action>
+    vars:
+      <var1>: "{{ <value1> }}"
+      <var2>: "{{ <value2> }}"
+      ...
+```
+
+If the `tasks/main.yml` is present, `include_role` without the `tasks_from`
```
michaelgugino: I would prefer that main.yml has contents like the following:
https://github.com/mgugino-upstream-stage/openshift-ansible/blob/86dfcb45e4c1149010d1d460d6ca6e0809063e93/roles/openshift_hosted/tasks/main.yml

It may be confusing to users who are browsing the source of the role to try to figure out what the role does.  They'll look at main.yml, see some tasks, and go "What is all this other stuff, how is it being used?"

docs/proposals/action_based_roles.md
ingvagabund: Though it require to re-understand each role, I like this approach. The included file was just an example but I can mention that in the proposal as well.

docs/proposals/action_based_roles.md
ingvagabund: Good point.

docs/proposals/action_based_roles.md
michaelgugino: For the most part, I think this will be a drop-in change.  You're essentially renaming old_role/main.yml -> new_role/action_1.yml

Then, you just need to move anything that wasn't the old role's main.yml in the action1 directory, and just refactor the include statements from action_1.yml to include the new folder.  Everything else can be exactly the same.

Then, later, after some common things are identified among the roles that are being consolidated, we can refactor those pieces to be used in a common place somewhere.

ingvagabund: Expected content of the action file updated

docs/proposals/action_based_roles.md
ingvagabund: Updated

vrutkovs:

PR #5173, opened 2017-08-23
[3.3] Use generic v3.4 and v3.3 for metrics and logging deployer

ruromero:
Users are having problems when upgrading using the deployer and having the x.x.0 version
3.3.0 only gets upgraded to the latest 3.3.0-x version which is outdated. The same for 3.4


ruromero:
BZ for this PR https://bugzilla.redhat.com/show_bug.cgi?id=1484403

PR #5176, opened 2017-08-23
Creating structure to warn for use of deprecated variables and set th…

ewolinetz:
…em in a single location before they are no longer honored

Implementation of https://trello.com/c/XKOGHSkP/460-3-deprecate-openshifthostedloggingmetrics-installer-variables

Tried to implement this in a way such that we can easily and cleanly add/remove variables in the future for any role we choose.

Addresses:
https://bugzilla.redhat.com/show_bug.cgi?id=1449812
https://bugzilla.redhat.com/show_bug.cgi?id=1464349


roles/openshift_sanitize_inventory/tasks/__logging.yml
```diff
@@ -0,0 +1,25 @@
...
+
+- set_fact:
+    openshift_logging_install_logging: openshift_hosted_logging_deploy | default(false) | bool
+    logging_hostname: "{{ openshift_hosted_logging_hostname | default('kibana.' ~ (openshift_master_default_subdomain | default('router.default.svc.cluster.local', true))) }}"
+    logging_ops_hostname: "{{ openshift_hosted_logging_ops_hostname | default('kibana-ops.' ~ (openshift_master_default_subdomain | default('router.default.svc.cluster.local', true))) }}"
+    logging_elasticsearch_cluster_size: "{{ openshift_hosted_logging_elasticsearch_cluster_size | default(1) }}"
+    logging_elasticsearch_ops_cluster_size: "{{ openshift_hosted_logging_elasticsearch_ops_cluster_size | default(1) }}"
+    openshift_logging_kibana_hostname: "{{ openshift_hosted_logging_hostname | default('kibana.' ~ (openshift_master_default_subdomain | default('router.default.svc.cluster.local', true))) }}"
+
+    openshift_logging_storage_kind: "{{ openshift_hosted_logging_storage_kind }}"
```
sosiouxme: don't all of these need defaults? seeing as how we don't want people to define them anymore

roles/openshift_sanitize_inventory/tasks/deprecations.yml
```diff
@@ -0,0 +1,13 @@
+---
+
+- name: Check for usage of deprecated variables
+  debug: msg="{{ __deprecation_header }} {{ item }} is a deprecated variable and will be no longer be used in the next minor release. Please update your inventory accordingly."
+  when:
+  - lookup('env', item) | bool
```
sosiouxme: doesn't this only look at env vars and not inventory vars?

roles/openshift_sanitize_inventory/tasks/deprecations.yml
```diff
@@ -0,0 +1,13 @@
...
+  with_items: "{{ __warn_deprecated_vars }}"
```
sosiouxme: Cool, could there be a pause after showing this (if any vars were flagged)? Otherwise it will fly past in the output and no one will see it...

Another thought: have we refactored out all the multiple includes of `openshift_sanitize_inventory`? Wouldn't want to pause multiple times. If it does still get included multiple times, we could just singleton-ize it to skip after the first time.

roles/openshift_sanitize_inventory/tasks/deprecations.yml
```diff
@@ -0,0 +1,13 @@
...
+
+# for with_fileglob Ansible resolves the path relative to the roles/<rolename>/files directory
+- name: Assign deprecated variables to correct counterparts
+  include: "{{ item }}"
+  with_fileglob:
+  - "../tasks/__*.yml"
```
sosiouxme: how about making the association more obvious:

    - "../tasks/deprecations_*.yml"

roles/openshift_sanitize_inventory/tasks/deprecations.yml
ewolinetz: I like the idea of a pause if any are flagged.
We have not yet refactored out multiple includes... I've noticed that with two nodes I see double the checks. +1 to singleton-izing

roles/openshift_sanitize_inventory/tasks/deprecations.yml
ewolinetz: I checked the negative case but not the positive case... I'll verify whether or not it can check an inventory var (hopefully it can, otherwise this approach wont work)

ewolinetz:
I'm currently unsure of how to approach the issue with the generation of PVs. Currently we tokenize inventory variables like `openshift_hosted_logging_volume_size` and process them that way... it makes sense to keep NFS storage nested under `hosted` however we want to move away from users using the `openshift_hosted_{logging,metrics}_*` variables to avoid confusion..

Thoughts?

roles/openshift_sanitize_inventory/tasks/__logging.yml
ewolinetz: I need to go through these a bit further, there starts to be issues decoupling this from the generation of NFS PVs within other roles that I'm unsure of how to approach. Ideally variables like this would just have defaults within the `openshift_logging` role, so I would end up removing this fact from this task file...

roles/openshift_sanitize_inventory/tasks/deprecations.yml
ewolinetz: Looks like it doesn't correctly check for an inventory variable like this. I'll need to come up with a different approach to check if a variable is defined

roles/openshift_sanitize_inventory/tasks/deprecations.yml
ewolinetz: So, we still see this called once per host and I'm not sure if we should change that process. If I try to use the pattern to singletonize that you had used for the excluder, I still see it run against each host...

sosiouxme: Oh yeah, that's not multiple includes, that's just the same thing running for each host. That's hard to deduplicate, and arguably you don't want to since different hosts may have different variables set.

I don't think *pause* will run once for each host though... does it?

roles/openshift_sanitize_inventory/tasks/__deprecations_logging.yml
```diff
@@ -0,0 +1,25 @@
+---
```
michaelgugino: I don't like sticking this logic in openshift_sanitize_inventory.  I think this logic should either be placed inside the role or the play in which we call the role.  Placing this logic here is confusing, and it's not specifically tied to the logging role so may go stale.

Also, sticking in some debug messages during the run., nobody's going to see that among the other 40k lines of stuff that scrolls by, even with a 5 second pause.  I open a terminal to kick off installs and then go do something else while it runs.

The advantage to sticking the deprecation parts inside the role are that it lives with the role, and users wanting to make changes will be consulting that role's logic anyway.

sosiouxme: Well, agreed; but the goal is to have this info flash up almost instantly, not half an hour into the install. If you wait until the logging role is running to check its variables, that's quite a way into the install. This is why I've been advocating for running roles in multiple phases, so we can encapsulate role-specific validations within the role, but run them at an appropriate time in the install. Until we have that, this seems expedient.

A longer wait would make the notice more likely to be seen, and provide a little more incentive to actually make the change. Say 30s?

@michaelgugino How would you suggest indicating deprecation of variables? About the only other way that makes sense to me would be a callback plugin that notes them at the end of the run (and I'm not sure how likely that is to be seen either depending on how many other callback plugins operate). Almost the only messaging available in Ansible is to fail a task; there doesn't seem to be a way to provide a nuanced warning that users will actually see.

michaelgugino: Probably should do a couple things.

1)  Let's set a fact like openshift_deprecated_variables = True
2) At the end of the run, if openshift_deprecated_variables = True, print "You're using deprecated variables in your inventory.  Please compare your variables with DEPRECATED.yml" or something along those lines.
3) DEPRECATED.yml (or other filename, in reasonable place) is a ABC sorted list of all deprecated variables
4) Possibly build a script users can run to compare their inventory to items in deprecated.

This has the benefit of 1) each role is in charge of it's own deprecating 2) User is informed at end of run.

ewolinetz: @michaelgugino I'm not really worried about that file going stale since it should be removed when we do our next minor release since we won't support the usage of those variables anymore, we wouldn't want to map them.

For that reason and the fact that this would run before any other plays/roles that could use these variables and we could map it prior to even making it down into the role, it made sense to me to have it in the same location that we are doing inventory sanitization...

michaelgugino: If this is going to live for only one minor release, I suppose it's fine.

sdodson: What about something like this? https://github.com/openshift/openshift-ansible/blob/master/playbooks/common/openshift-cluster/validate_hostnames.yml#L11-L20 it's modal unless you override it

sdodson:
General approach seems fine to me. I think it's best to halt and prompt with an override variable to skip that.

roles/openshift_sanitize_inventory/tasks/__deprecations_logging.yml
ewolinetz: @sdodson would we want to have that flag just skip the pause or the warning message too?

sdodson: Just the pause that way the logs would still have an accurate list of the deprecated variables if they were to look for them.

sdodson: Maybe we can also print a summary at the end too?

ewolinetz: Now that I can get it to run to completion without error I'll try to figure this out

roles/openshift_hosted_facts/tasks/main.yml
```diff
@@ -8,9 +8,10 @@
 
 - name: Set hosted facts
   openshift_facts:
-    role: hosted
+    role: "{{ item }}"
```
michaelgugino: Not much is relying on openshift_hosted_facts at this point.  I don't think this is necessary.  If you're making changes that require this, make them so they don't if you can.  This role is one task away from extinction.

ewolinetz: unfortunately, I had to modify this so that it would correctly set the facts for the `logging` `loggingops` and `metrics` facts so that `openshift_persistent_volumes` had the correct information to create pvs. During my testing, before I updated this it would skip creating the PV because it didn't correctly parse out facts...

roles/openshift_sanitize_inventory/tasks/__deprecations_logging.yml
ewolinetz: @sdodson I'm not sure how to get this to display at the end. any thoughts what approach I can take? A handler doesn't seem to be firing..

ewolinetz:
I need to rebase but otherwise I believe the changes necessary as part of this are done.
PTAL

ewolinetz:
aos-ci-test

roles/openshift_sanitize_inventory/tasks/__deprecations_logging.yml
sosiouxme: A handler would only fire at the end of a play anyway. That's fine as long as your play is the last one... otherwise I think you want a callback plugin.

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

ewolinetz:
aos-ci-test

wozniakjan:
@ewolinetz could you please take a look at this bz https://bugzilla.redhat.com/show_bug.cgi?id=1491636? I think the current issue they are seeing may be related to this PR

PR #5184, opened 2017-08-23
bug 1480878. Default pvc for logging

jcantrill:
This PR fixes bug 1480878 by defaulting a PVC size if one is not specified

### DIFF
--- a/roles/openshift_logging_elasticsearch/tasks/main.yaml
+++ b/roles/openshift_logging_elasticsearch/tasks/main.yaml
@@ -229,7 +229,7 @@
       dest: "{{ tempdir }}/templates/logging-es-pvc.yml"
     vars:
       obj_name: "{{ openshift_logging_elasticsearch_pvc_name }}"
-      size: "{{ openshift_logging_elasticsearch_pvc_size }}"
+      size: "{{ (openshift_logging_elasticsearch_pvc_size | trim | length == 0) | ternary('10Gi', openshift_logging_elasticsearch_pvc_size) }}"
       access_modes: "{{ openshift_logging_elasticsearch_pvc_access_modes | list }}"
       pv_selector: "{{ openshift_logging_elasticsearch_pvc_pv_selector }}"
       storage_class_name: "{{ openshift_logging_elasticsearch_pvc_storage_class_name | default('', true) }}"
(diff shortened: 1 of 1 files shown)


jcantrill:
aos-ci-test

jcantrill:
cc @t0ffel

roles/openshift_logging_elasticsearch/tasks/main.yaml
```diff
@@ -229,7 +229,7 @@
       dest: "{{ tempdir }}/templates/logging-es-pvc.yml"
     vars:
       obj_name: "{{ openshift_logging_elasticsearch_pvc_name }}"
-      size: "{{ openshift_logging_elasticsearch_pvc_size }}"
+      size: "{{ (openshift_logging_elasticsearch_pvc_size | trim | length == 0) | ternary('10Gi', openshift_logging_elasticsearch_pvc_size) }}"
```
ewolinetz: can we use some python magic and do something like `( openshift_logging_elasticsearch_pvc_size | bool ) | ternary(openshift_logging_elasticsearch_pvc_size, '10Gi')`  if its empty it should come back as false...

roles/openshift_logging_elasticsearch/tasks/main.yaml
```diff
@@ -243,7 +243,7 @@
...
```
ewolinetz: same for here...

t0ffel:
Why not set `openshift_logging_elasticsearch_pvc_size` to `10Gi` in `defaults/main.yml` instead?

jcantrill:
@t0ffel We use the size as a point for evaluation of the storage type.  It was easier to set the default in this way instead of refactoring the other logic.

roles/openshift_logging_elasticsearch/tasks/main.yaml
jcantrill: So i tried your suggestion and the result when I provided no inventory var was the same as initially reported by this issue.  The requested size resulted in '0' instead of the default

ewolinetz: Ack, then what you have is fine

jcantrill:
[merge]

jcantrill:
[merge]

jcantrill:
[merge]

jcantrill:

PR #5186, opened 2017-08-23
Remove near-meta role openshift_cli_facts

michaelgugino:
openshift_cli_facts only runs a single task, and otherwise
the only existence is to serve as a link in the dependency
chain towards openshift_facts.

This commit flattens openshift_cli into one role.


michaelgugino:
bot, retest this please

sdodson:
aos-ci-test

sdodson:
[merge]

sdodson:
It looks like openshift_version depends on openshift.common.cli_image being set but I don't see that role actually depending on this.

roles/openshift_cli/tasks/main.yml
```diff
@@ -1,4 +1,9 @@
 ---
+- openshift_facts:
+    role: common
+    local_facts:
+      cli_image: "{{ osm_image | default(None) }}"
```
kwoodson: Is it possible to move this to `playbooks/common/openshift-cluster/initialize_facts.py`?  It seems to me that osm_image is coming from the inventory and we could just assign this to the rest of the facts we are setting in a single place.  

@michaelgugino, thoughts on this?

kwoodson:
aos-ci-test

ashcrow:

PR #5182, opened 2017-08-23
Remove unused upgrade playbook

mtnbikenc:
This playbook was created for upgrades back in v3.3 but has never been used.

PR #5179, opened 2017-08-23
Standardize openshift_provisioners entry point

mtnbikenc:
Adding a `byo` entry point to standardize the playbook flow for openshift_provisioners.

PR #5188, opened 2017-08-23
Test: Fail on entry point playbooks in common

mtnbikenc:
- Added a tox check to fail if an entry point was found in `playbooks/common`
- Moved task include files into `tasks/' directory to separate task file from playbook files.  This allows proper parsing of yaml files as playbooks.  It also enhances the description of the file being a task list instead of a playbook.
- Deleted legacy files used by {aws/gce/libvirt/openstack}

Depends on:
#5179 merged
#5182 merged
#5201 merged


mtnbikenc:
@kwoodson You mentioned something about breaking out the checks when I added the entry point discovery logic.  Maybe I can handle that in this PR to better separate the individual tests that are run on playbooks.

kwoodson:
@mtnbikenc, We have checks scattered throughout most of our repository.  I'd really like to group those together in the prevalidation of our health checker or somewhere similar.  It seems illogical to run for minutes during an install and fail inside of a role/playbook when a version of python is not as expected.  We should have caught something like this early on.

mtnbikenc:
@kwoodson I think my terminology may have thrown you off.  I was referring to the checks we are performing in CI using setup.py to validate code meets our standards.  I previously added the discovery and syntax checking of all entry point playbooks.  In this PR I'm adding a check to make sure entry point playbooks are not added to `common`.

kwoodson:
@mtnbikenc, makes sense now for entry point checking. Thanks for the explanation.

mtnbikenc:
aos-ci-test

mtnbikenc:
[merge]

dobbymoodge:
rekick [merge] b/c restart

mtnbikenc:
Flake https://github.com/openshift/origin/issues/16005

mtnbikenc:
[merge]

mtnbikenc:
Flake pulling from docker.io
`Trying to pull repository docker.io/openshift/origin-release`

PR #5189, opened 2017-08-24
Bug 1467265 - logging: add 'purge' option with uninstall

wozniakjan:
The common uninstall would keep PVC around to prevent unwanted data loss during reinstalls. There is a scenario where the user may want complete removal of EFK stack including PVC.

This change adds an optional `openshift_logging_purge_logging` variable to completely remove the EFK stack. By default, this is set to 'False' and only works if user sets `openshift_logging_install_logging` to 'False' to initiate the uninstallation process.

cc: @jcantrill , @richm 

https://bugzilla.redhat.com/show_bug.cgi?id=1467265

### DIFF
--- a/roles/openshift_logging/README.md
+++ b/roles/openshift_logging/README.md
@@ -15,12 +15,11 @@ to the list of persisted [node labels](https://docs.openshift.org/latest/install
 ###Required vars:
 
 - `openshift_logging_install_logging`: When `True` the `openshift_logging` role will install Aggregated Logging.
-- `openshift_logging_upgrade_logging`:  When `True` the `openshift_logging` role will upgrade Aggregated Logging.
 
-When both `openshift_logging_install_logging` and `openshift_logging_upgrade_logging` are `False` the `openshift_logging` role will uninstall Aggregated Logging.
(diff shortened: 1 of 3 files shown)


jcantrill:
aos-ci-test

jcantrill:
@wozniakjan The situation from which this originated, is it necessary for us to backport this to 3.6?

wozniakjan:
@jcantrill I don't think it is necessary to backport but it can be backported without breaking any backward compatibility. However, I consider this to be a new functionality to completely remove the logging and it should be probably properly documented so I would rather see it introduced in the following minor release

Maybe @ruromero could weigh in as a reporter of that BZ

jcantrill:
I would prefer not to backport if we do not need to.

sdodson:
aos-ci-test

jcantrill:
[merge]

jcantrill:
[merge]
-- 
--
Jeff Cantrill
Senior Software Engineer, Red Hat Engineering
OpenShift Integration Services
Red Hat, Inc.
*Office*: 703-748-4420 | 866-546-8970 ext. 8162420
jcantril@redhat.com
http://www.redhat.com

wozniakjan:
aos-ci-test

ingvagabund:
aos-ci-test

PR #5191, opened 2017-08-24
Backport to 1.5 - bug 1468987: kibana_proxy OOM

wozniakjan:
Backporting https://github.com/openshift/openshift-ansible/pull/4761 to 1.5

We currently set the memory allocated to the kibana-proxy container to be the same as `max_old_space_size` for nodejs. But in V8, the heap consists of multiple spaces.

The old space has only memory ready to be GC and measuring the used heap by kibana-proxy code, there is at least additional 32MB needed in the code space when `max_old_space_size` peaks.

Setting the default memory limit to 256MB here and also changing the default calculation of `max_old_space_size` in the image repository to be only half of what the container receives to allow some heap for other `spaces`.

### DIFF
--- a/roles/openshift_logging/defaults/main.yml
+++ b/roles/openshift_logging/defaults/main.yml
@@ -25,7 +25,7 @@ openshift_logging_kibana_cpu_limit: null
 openshift_logging_kibana_memory_limit: 736Mi
 openshift_logging_kibana_proxy_debug: false
 openshift_logging_kibana_proxy_cpu_limit: null
-openshift_logging_kibana_proxy_memory_limit: 96Mi
+openshift_logging_kibana_proxy_memory_limit: 256Mi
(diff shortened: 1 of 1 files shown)


niconosenzo:
Seems the backport was only applied to the kibana_proxy_memory_limit (not kibana_ops):

[root@master-0 defaults]# grep proxy_memory_limit main.yml 
openshift_logging_kibana_proxy_memory_limit: 256Mi
openshift_logging_kibana_ops_proxy_memory_limit: 96Mi
[root@master-0 defaults]# rpm -q openshift-ansible
openshift-ansible-3.5.120-1.git.0.c60f69a.el7.noarch
[root@master-0 defaults]# pwd
/usr/share/ansible/openshift-ansible/roles/openshift_logging/defaults
[root@master-0 defaults]#

PR #5082, opened 2017-08-14
GlusterFS: add minor README note for #5071

jarrpa:
Closes #5071

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

PR #5196, opened 2017-08-24
Update glusterfs docs for working array definition

dougbtv:
This updates the glusterfs docs to show an example for the `glusterfs_devices` array.

Due to some dynamic creation of variables within loops in jinja2 templates, the array cannot be specified in an inventory (or as playbook `--extra-vars`) as  `'[ "/dev/sdb" ]'` it must be specified without the single quotes, e.g. `[ "/dev/sdb" ]`. 

Previously if used in the fashion shown in the documents, the array would be interpreted as a string literally with the template resulting in "an array of characters", like this:

```
"devices": ["["," ",""","/","d","e","v","/","v","d","c","""," ","]"]
```
This PR enhances the example to show a working specification of the array.

### DIFF
--- a/roles/openshift_storage_glusterfs/README.md
+++ b/roles/openshift_storage_glusterfs/README.md
@@ -55,7 +55,7 @@ defined:
 
 | Name              | Default value | Description                             |
 |-------------------|---------------|-----------------------------------------|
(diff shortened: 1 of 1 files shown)


jarrpa:
This is wrong. See https://github.com/openshift/openshift-ansible/pull/5082

dougbtv:
Thanks @jarrpa -- appreciate the pointers to that PR and the issue describing it. Closing this one.

PR #5198, opened 2017-08-24
Upgrade check for OpenShift authorization objects

### LINKED ISSUE #5021: Run `oadm migrate authorization` pre 3.6 to 3.7 upgrades
Prior to upgrading from 3.6 to 3.7 we must run
`oadm migrate authorization` with no parameters, if the command fails the upgrade should abort. The command may fail in a loaded cluster and should be retried every 15 seconds for up to one minute?

CC: @enj

mtnbikenc:
Adds task to run `oc adm migrate authorization` to ensure OpenShift authorization objects are in sync.

Fixes #5021


mtnbikenc:
aos-ci-test

sdodson:
[test]

PR #5202, opened 2017-08-24
cri-o: add support for additional registries

giuseppe:
Support added to CRI-O with:
https://github.com/kubernetes-incubator/cri-o/commit/a35727c80bd2a26613aae21db00628045cb9be24

### DIFF
--- a/roles/docker/tasks/systemcontainer_crio.yml
+++ b/roles/docker/tasks/systemcontainer_crio.yml
@@ -3,6 +3,15 @@
 - set_fact:
     l_insecure_crio_registries: "{{ '\"{}\"'.format('\", \"'.join(openshift.docker.insecure_registries)) }}"
   when: openshift.docker.insecure_registries
+- set_fact:
+    l_crio_registries: "{{ openshift.docker.additional_registries + ['docker.io'] }}"
+  when: openshift.docker.additional_registries
+- set_fact:
+    l_crio_registries: "{{ ['docker.io'] }}"
+  when: not openshift.docker.additional_registries
+- set_fact:
+    l_additional_crio_registries: "{{ '\"{}\"'.format('\", \"'.join(l_crio_registries)) }}"
+  when: openshift.docker.additional_registries
 
 - name: Ensure container-selinux is installed
   package:
--- a/roles/docker/templates/crio.conf.j2
+++ b/roles/docker/templates/crio.conf.j2
@@ -120,6 +120,11 @@ insecure_registries = [
 {{ l_insecure_crio_registries|default("") }}
 ]
 
+# registries is used to specify a comma separated list of registries to be used
+# when pulling an unqualified image (e.g. fedora:rawhide).
+registries = [
+{{ l_additional_crio_registries|default("") }}
+]
 # The "crio.network" table contains settings pertaining to the
 # management of CNI plugins.
 [crio.network]


mrunalp:
@runcom PTAL
@stevekuznetsov we would need this to allow pulling from internal registries.

roles/docker/tasks/systemcontainer_crio.yml
```diff
@@ -3,6 +3,9 @@
 - set_fact:
     l_insecure_crio_registries: "{{ '\"{}\"'.format('\", \"'.join(openshift.docker.insecure_registries)) }}"
   when: openshift.docker.insecure_registries
+- set_fact:
+    l_additional_crio_registries: "{{ '\"{}\"'.format('\", \"'.join(openshift.docker.additional_registries)) }}"
+  when: openshift.docker.additional_registries
```
runcom: We need to specify "docker.io" here as well. Cri-o is not gonna default to docker like projectatomic/docker does ;)

giuseppe: I can easily add it, but do we really want this?  Not going by default on `docker.io` for each image not found in the specified registries looks like a "feature" to me

runcom: Well, that could still break someone, Docker falls back, so it'd say let's start by having feature parity with it and add the fall back. We could revisit this later as nobody is going to really use this now. If we have people who relies on the fallback they'll get broken by this, let's have happy people testing openshift :)

giuseppe:
@runcom @mrunalp pushed a new version that always adds "docker.io" ⬆️

runcom:
Works for me, thanks!

ashcrow:
@runcom @giuseppe What about defaulting to having ``docker.io`` in the inventory file for ``crio_additional_registries``? That way if someone wants to remove it they can modify it on install rather than waiting till after installation or needing to find the right location in the YAML to modify?

giuseppe:
@ashcrow wouldn't that still break the case of using existing inventory files (except obviously the new variable) as they are?  I am all for it, but @runcom and @mrunalp seem to prefer the other approach

ashcrow:
@giuseppe true, if you use an existing inventory file it wouldn't work. However, our cri-o support is still Tech Preview so if there is a time to make a change like that it would be now. I won't block on hardcoding it in the yaml it just feels a bit heavy.

runcom:
> @runcom @giuseppe What about defaulting to having docker.io in the inventory file for crio_additional_registries? 

I'm fine with this as long as a default installation of cri-o + origin has docker.io :) (I may be missing something but if what you propose is still a default, I'm fine)

ashcrow:
@giuseppe I'll leave it to your discretion as to if using inventory or using YAML is the best way to go. I'm fine either way.

giuseppe:
If I understand correctly, @runcom wants CRI-O to work exactly in the same way as Docker (that always fallback to docker.io) without any additional change to the inventory file.  With the change going into the default inventory file, we would still behave differently with existing inventory files while the (uglier) change to the YAML forces docker.io to be always present, also with existing inventory files that do not have it by default.  If we want "default" to mean also not specified variable, then I think we should go with the YAML change as this PR does

ashcrow:
[test]

ashcrow:

ashcrow:
aos-ci-test

ashcrow:
Infra flake

ashcrow:
aos-ci-test

giuseppe:
aos-ci-test

ashcrow:
@openshift-bot don't ignore me! 💢 :rage4:

ashcrow:
Infra flake

```
hack/testing/entrypoint.sh:40: `fpod=$( get_running_pod fluentd )` exited with status 1.
```

ashcrow:
aos-ci-test

sdodson:
[merge]

ashcrow:
Looks like infra flake
```
[INFO] Pulling the openshift/origin-release:golang-1.8 image to update it...
Trying to pull repository registry.access.redhat.com/openshift/origin-release ... 
Trying to pull repository docker.io/openshift/origin-release ... 
golang-1.8: Pulling from docker.io/openshift/origin-release
Digest: sha256:0f1705f538fc3096a32b8b3111d5c7cb2eabc51b3d4a027c20ca97f46ff39d6d
Write failed: Broken pipe
```

ashcrow:
[merge]

PR #5205, opened 2017-08-24
Move additional/block/insecure registires to /etc/containers/...

michaelgugino:
Move additional/block/insecure registires to /etc/containers/registries.conf

This commit moves additional/block/insecure registries to
/etc/containers/registries.conf and comments existing lines in
/etc/sysconfig/docker.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1460930

### DIFF
--- a/roles/docker/tasks/package_docker.yml
+++ b/roles/docker/tasks/package_docker.yml
@@ -53,22 +53,22 @@
 - stat: path=/etc/sysconfig/docker
   register: docker_check
 
-- name: Set registry params
+- name: Comment old registry params in /etc/sysconfig/docker
   lineinfile:
     dest: /etc/sysconfig/docker
     regexp: '^{{ item.reg_conf_var }}=.*$'
-    line: "{{ item.reg_conf_var }}='{{ item.reg_fact_val | oo_prepend_strings_in_list(item.reg_flag ~ ' ') | join(' ') }}'"
-  when: item.reg_fact_val != '' and docker_check.stat.isreg is defined and docker_check.stat.isreg
+    line: "#{{ item.reg_conf_var }}=''# Moved to {{ containers_registries_conf_path }}"
   with_items:
   - reg_conf_var: ADD_REGISTRY
-    reg_fact_val: "{{ docker_additional_registries | default(None, true)}}"
-    reg_flag: --add-registry
   - reg_conf_var: BLOCK_REGISTRY
-    reg_fact_val: "{{ docker_blocked_registries| default(None, true) }}"
-    reg_flag: --block-registry
   - reg_conf_var: INSECURE_REGISTRY
-    reg_fact_val: "{{ docker_insecure_registries| default(None, true) }}"
-    reg_flag: --insecure-registry
+  notify:
+  - restart docker
+
+- name: Place additional/blocked/insecure registies in /etc/containers/registries.conf
+  template:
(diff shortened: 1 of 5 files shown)


michaelgugino:
Template from:  https://github.com/projectatomic/registries/blob/master/registries.conf

ashcrow:
@michaelgugino does this replace ``daemon.json`` usage for ``docker`` backends?

michaelgugino:
@ashcrow I'm unsure if system containers are reading that file or not.  According to the bugzilla link, docker should be utilizing this new file.  I believe crio will also read from this new file.  We'll have to investigate whether or not the change needs to be made to container-daemon.json for system containers.  In any case, that could be another PR and this PR doesn't make any changes to that part.

ashcrow:
``container-daemon.json`` is docker's ``daemon.json`` configuration file which was to replace the old sysconfig location used by openshift-ansible. If ``/etc/containers/registries.conf`` replaces the registry portion for docker (and container engine which is docker) I think it would be a good idea to simplify into using one file (or set of files) for the configuration for all docker backends. It doesn't have to happen all in one PR, I just want to make sure we don't accidentally have the old sysconfig, the expected replacement of *daemon.json, and a new possibility in /etc/containers.

ashcrow:
[test]

michaelgugino:
@ashcrow unfortunately, this patch only affects registries; it's not a system-wide docker config file.  We're not getting rid of /etc/sysconfig/docker, someone decided to move only portions of the config (ie, the registires) to /etc/containers/registries.conf

daemon.json is only used in the following play (from docker role):
```
- name: Use System Container Docker if Requested
  include: systemcontainer_docker.yml
  when:
  - l_use_system_container
  - not l_use_crio_only
```

We'll need more input or investigation as to what actually consumes daemon.json and /etc/containers/registries.conf

ashcrow:

ashcrow:
[test]

baude:
FYI , for containers-image -> https://github.com/containers/image/pull/322

ashcrow:

ashcrow:
aos-ci-test

ashcrow:
infra flake
```
Unable to find image 'openshift3/ose:v3.6' locally\n/usr/bin/docker-current: Error response from daemon: Invalid hostname \"r\".\nSee '/usr/bin/docker-current run --help'.", 
```

ashcrow:
aos-ci-test

roles/docker/tasks/package_docker.yml
```diff
@@ -51,22 +51,22 @@
...
   register: docker_check
 
-- name: Set registry params
+- name: Comment old registry params in /etc/sysconfig/docker
   lineinfile:
     dest: /etc/sysconfig/docker
     regexp: '^{{ item.reg_conf_var }}=.*$'
-    line: "{{ item.reg_conf_var }}='{{ item.reg_fact_val | oo_prepend_strings_in_list(item.reg_flag ~ ' ') | join(' ') }}'"
-  when: item.reg_fact_val != '' and docker_check.stat.isreg is defined and docker_check.stat.isreg
+    line: "#{{ item.reg_conf_var }}=''# Moved to {{ containers_registries_conf_path }}"
```
ashcrow: 👍

ashcrow:
Please rebase and we can test again.

PR #5208, opened 2017-08-25
Remove openshift common

michaelgugino:


michaelgugino:
bot, retest this please

playbooks/common/openshift-cluster/config.yml
```diff
@@ -42,6 +42,14 @@
     r_openshift_excluder_action: disable
     r_openshift_excluder_service_type: "{{ openshift.common.service_type }}"
 
+- name: Disable excluders
+  hosts: oo_masters_to_config:oo_nodes_to_config
+  tasks:
+  - name: Set hostname
+    command: >
```
kwoodson: Would hostname module be better than command here?

http://docs.ansible.com/ansible/latest/hostname_module.html

playbooks/common/openshift-cluster/std_include.yml
```diff
@@ -7,6 +7,10 @@
   tags:
   - always
 
+- include: sanity_checks.yml
+  tags:
```
kwoodson: Are we removing tags?

playbooks/common/openshift-cluster/std_include.yml
```diff
@@ -7,6 +7,10 @@
   tags:
   - always
 
+- include: sanity_checks.yml
```
kwoodson: @mtnbikenc, this feels like the right place but I want to ensure Russ has some input as we are guarding the `std_include.yml` so that it does not expand without a double :+1:  on it.

roles/openshift_examples/meta/main.yml
```diff
@@ -11,5 +11,4 @@ galaxy_info:
     - 7
   categories:
   - cloud
-dependencies:
-- role: openshift_common
+dependencies: []
```
kwoodson: big :+1:  for removing dependencies

roles/openshift_facts/library/openshift_facts.py
```diff
@@ -590,13 +518,8 @@ def set_dnsmasq_facts_if_unset(facts):
...
 
     if 'common' in facts:
-        if 'use_dnsmasq' not in facts['common']:
-            facts['common']['use_dnsmasq'] = bool(safe_get_bool(facts['common']['version_gte_3_2_or_1_2']))
         if 'master' in facts and 'dns_port' not in facts['master']:
-            if safe_get_bool(facts['common']['use_dnsmasq']):
-                facts['master']['dns_port'] = 8053
-            else:
-                facts['master']['dns_port'] = 53
+            facts['master']['dns_port'] = 8053
```
kwoodson: At some point I'd like to see this type of stuff bubble up to a default variable in a role.  I don't like digging to figure out information like this.

sdodson:
aos-ci-test

roles/openshift_hosted_templates/meta/main.yml
```diff
@@ -11,5 +11,4 @@ galaxy_info:
     - 7
   categories:
   - cloud
-dependencies:
-- role: openshift_common
+dependencies: []
```
kwoodson: :+1:

sdodson:
[test]

roles/openshift_hosted/defaults/main.yml
```diff
@@ -46,4 +46,4 @@ r_openshift_hosted_registry_os_firewall_deny: []
 r_openshift_hosted_registry_os_firewall_allow:
 - service: Docker Registry Port
   port: 5000/tcp
-  cond: "{{ r_openshift_hosted_use_calico }}"
+  cond: "{{ openshift_hosted_use_calico }}"
```
twiest: I think this is what @sdodson was referencing when he said "It goes against some past architecture decisions such as requiring that all role variables are passed in but it's not like this is introducing more of that it's just substituting openshift_facts magic with more defaults."

I do prefer that we don't reference global variables like this.

Instead of just being able to call this role and pass in variables, I now need to make sure `openshift_hosted_use_calico` is set somewhere before I call this role.

Also, to discover this, I need to read through the code of the role.

Still feels like a bad pattern to me.

sdodson: Yeah, that's what I meant. Thanks for clarifying.

michaelgugino: openshift_hosted_use_calico is not a 'role variable,' as it's not specific to any one role.  It's a user-provided inventory variable that controls many roles.

What is the benefit of defining r_openshift_hosted_use_calico without a default in this role?  In reality, this variable is a direct derivative of openshift_hosted_use_calico.  r_openshift_hosted_use_calico provides no meaning without openshift_hosted_use_calico.

I think the r_variable pattern might have been adopted for a valid purpose, but I think it was too widely adopted and misapplied in cases like r_openshift_hosted_use_calico

michaelgugino: Good catch, I will need to provide default for openshift_hosted_use_calico here though.

kwoodson: @mwoodson @twiest, would it be sufficient to place it in the defaults and have a sane default for the value?

playbooks/common/openshift-cluster/std_include.yml
kwoodson: I'm fine if it will all be done in a single PR or if you want to punt.

roles/openshift_facts/library/openshift_facts.py
kwoodson: Ok, maybe make a note.  I'm fine putting this into a separate PR.

playbooks/common/openshift-cluster/config.yml
```diff
@@ -42,6 +42,13 @@
     r_openshift_excluder_action: disable
     r_openshift_excluder_service_type: "{{ openshift.common.service_type }}"
 
+- name: Disable excluders
```
mtnbikenc: Need to update this play name.

playbooks/common/openshift-cluster/config.yml
```diff
@@ -42,6 +42,13 @@
...
+  hosts: oo_masters_to_config:oo_nodes_to_config
+  tasks:
+  - name: Set hostname
+    hostname: "{{ openshift.common.hostname }}"
```
mtnbikenc: `hostname:` needs the `name:` argument
http://docs.ansible.com/ansible/latest/hostname_module.html

:+1: for moving away from `command`

playbooks/common/openshift-cluster/std_include.yml
mtnbikenc: This could be fine for now.  Moving these checks up higher is good, and since these are just `fail` checks they shouldn't add discernable time.

playbooks/common/openshift-cluster/std_include.yml
mtnbikenc: I see tags coming out as we break things apart.  And after we are all done with breaking stuff apart, we come back one last time and sweep up any remaining tags.

playbooks/common/openshift-cluster/config.yml
sdodson: As long as this works on atomic host. I'm pretty sure we found that it didn't. 95223c4b92

kwoodson: @sdodson, did we file any bugs or issues we can track for the hostname module not working?

sdodson: I don't recall.

roles/openshift_hosted/defaults/main.yml
twiest: @michaelgugino "It's a user-provided inventory variable that controls many roles."

Yeah, that's what I was saying. That's why we call varialbes like that "global variables". Once you tie roles to that global variable, the role becomes much less flexible.

We prefer roles to be self contained. This includes not referencing global variables like this.

That way we can call this role and pass in what we'd like the role to use instead of having to ensure that a bunch of global variables are all set up.

Also, I'd prefer that the role variable be named for what is being enabled / disabled. The role is doing some logic with that variable, and thus we should name it based on what that logic actually means.

Then if we want the global variable to be used for that functionality, then we can pass it in for the role variable.

@sdodson Do you have the recording of the meeting where we discussed not using global variables in roles?

I think that would give @michaelgugino more context into the discussion / decision.

@kwoodson assuming you're talking about a role variable, then that sounds fine to me.

michaelgugino: @twiest This is not a 'global variable'.  It is an inventory variable.  Inventory variables override role defaults.  Each role that consumes such an inventory should also specify a role default, except in a very few special cases (e.g. passwords).

Imagine you're building a web app.  You need two roles, app, and mysql.  Does it make sense to give each role it's own unique name for mysql_port?  No, it doesn't.  We probably want mysql_port to be the same in each place.

Similarly, if we have a choice of different database backends, which affect both app and db role, the variable use_mysqldb would make sense.

If we were dynamically deriving some data and using set_fact to populate that data into a variable, then the method you are suggesting makes sense.  In the case we're currently discussing, it is adding zero-value and only introducing a middle-man variable with no value added, only negatives added.

> We prefer roles to be self contained. This includes not referencing global variables like this.
You're only trying to cover up the fact that an inventory variable is indeed being used.  The role is still 100% dependent on that inventory variable, whether you reference it directly or not.

twiest: @michaelgugino 

> You're only trying to cover up the fact that an inventory variable is indeed being used. The role is still 100% dependent on that inventory variable, whether you reference it directly or not.

Well, the point is to decoupling the role from the inventory variable. This allows the role caller to specifically set the variable based on what they want done. The user of the role could use an inventory variable, or they could just set the role variable manually.

The point is, we're not forcing the caller of the role to have their inventory setup in a specific way before being able to call the role.

Also, again, in this specific case, that variable is determining if the registry should have a hole poked in iptables on port 5000. So it would be better to have a role variable with a name that specifically describes what it's being used for.  Something like `r_openshift_hosted_enable_registry_port_in_fw` or whatever (yes I changed the name of the var since my last comment).  :)

twiest: @sdodson At this point, I think we should discuss this this in the arch meeting and get the rest of the team's opinions on this.

Would you mind adding it to the agenda?

sdodson: done.

michaelgugino:
Blocked By:  https://github.com/openshift/openshift-ansible/pull/5247

ashcrow:
```yum flake network unreachable```

**Comment by ashcrow** [comment]
@mtnbikenc  did @michaelgugino address the items you required or are the remaining items still block worthy?

**Comment by mtnbikenc** [comment]
I'm reviewing again now and will attempt an install with this branch.  This is a huge PR with many optional code paths affected so it will be difficult to validate all the changes.

**Comment by mtnbikenc** [comment]
aos-ci-test

**Comment by michaelgugino** [comment]
@mtnbikenc Good catch, thanks.  That was a type, missed the 'r_' prefix.

**Comment by sdodson** [comment]
aos-ci-test

**Comment by mtnbikenc** [comment]
[merge]

File: playbooks/common/openshift-cluster/config.yml [hunk 10]
```diff
@@ -22,6 +22,14 @@
   tags:
   - always
 
+- name: Set hostname
+  hosts: oo_masters_to_config:oo_nodes_to_config
+  tasks:
+  - name: Set hostname
+    hostname:
+      name: "{{ openshift.common.hostname }}"
+    when: openshift_set_hostname | default(True) | bool
```
detiber: @sdodson This appears to be a regression from the previous behavior. Setting the hostname should not be required, with the exception of some certain edge cases. There were many user complaints with the previous default of always setting the hostname.

roles/openshift_version/tasks/main.yml
```diff
@@ -5,6 +5,12 @@
     is_containerized: "{{ openshift.common.is_containerized | default(False) | bool }}"
     is_atomic: "{{ openshift.common.is_atomic | default(False) | bool }}"
 
+- name: Install the base package for versioning
```
abutcher: We're installing the base package on all hosts which will require subs on some hosts we didn't require subs on before such as loadbalancer hosts.

PR #4641, opened 2017-06-29
logging_es: temporarily disable readiness probe

wozniakjan:
The readiness probe, as currently implemented, together with the ES master discovery using the logging-es service, causes a deadlock on clusters with more than one node. Readiness probe will be reintroduced in a later release when the issue is properly addressed.

More information in:
https://bugzilla.redhat.com/show_bug.cgi?id=1459430

PR #5209, opened 2017-08-25
elasticsearch: reintroduce readiness probe

wozniakjan:
Interim readiness probe until multi-role Elasticsearch topology from https://github.com/openshift/openshift-ansible/pull/5001 is introduced.

**Depends on these PR:**
https://github.com/fabric8io/elasticsearch-cloud-kubernetes/pull/98
https://github.com/openshift/origin-aggregated-logging/pull/609

cc: @lukas-vlcek , @jcantrill , @t0ffel , @richm , @portante

### DIFF
--- a/roles/openshift_logging_elasticsearch/templates/es.j2
+++ b/roles/openshift_logging_elasticsearch/templates/es.j2
@@ -90,6 +90,12 @@ spec:
               name: "RECOVER_AFTER_TIME"
               value: "{{openshift_logging_elasticsearch_recover_after_time}}"
             -
+              name: "READINESS_PROBE_TIMEOUT"
+              value: "30"
+            -
+              name: "POD_LABEL"
+              value: "component={{component}}" 
+            -
               name: "IS_MASTER"
               value: "{% if deploy_type in ['data-master', 'master'] %}true{% else %}false{% endif %}"
 
@@ -106,6 +112,13 @@ spec:
               readOnly: true
             - name: elasticsearch-storage
               mountPath: /elasticsearch/persistent
+          readinessProbe:
+            exec:
+              command:
+              - "/usr/share/java/elasticsearch/probe/readiness.sh"
+            initialDelaySeconds: 10
+            timeoutSeconds: 30
+            periodSeconds: 5
       volumes:
         - name: elasticsearch
           secret:
--- a/roles/openshift_logging_elasticsearch/templates/elasticsearch.yml.j2
+++ b/roles/openshift_logging_elasticsearch/templates/elasticsearch.yml.j2
@@ -24,7 +24,8 @@ network:
 
 cloud:
   kubernetes:
-    service: ${SERVICE_DNS}
+    pod_label: ${POD_LABEL}
+    pod_port: 9300
     namespace: ${NAMESPACE}
 
 discovery:


wozniakjan:
[(3) Elasticsearch label discovery [epic-stability][logging]](https://trello.com/c/IxWno3R8/537-3-elasticsearch-label-discovery-epic-stabilitylogging)

roles/openshift_logging_elasticsearch/templates/es.j2
```diff
@@ -90,6 +90,9 @@ spec:
               name: "RECOVER_AFTER_TIME"
               value: "{{openshift_logging_elasticsearch_recover_after_time}}"
             -
+              name: "READINESS_PROBE_TIMEOUT"
+              value: "30"
```
jcantrill: @portante @t0ffel Any thoughts if this timeout is long enough?

richm: I've seen ES take much longer than 30 seconds to become available.

What happens after this 30 second timeout is hit?  Will Elasticsearch continue to recover indices, etc. whatever it is doing in the background?  Let's say it takes 1 hour to become ready - what does the user need to do to make the Elasticsearch service available in the cluster at that point?  Redeploy?

t0ffel: Doesn't https://github.com/openshift/openshift-ansible/pull/5209/files#diff-0b7373a8acdac264abe27e0f9ded4f1aR117 say that readiness probe will timeout after 30 seconds, so if this is 30 seconds, then total time of readiness probe run may exceed 30 seconds and it'll time out?

Should that even be a parameter? imho hardcoding it is fine. If the 1st readiness probe fails, the next one or the one after next will succeed.

wozniakjan: these timeouts were picked from the exact original values that were previously in master - https://github.com/openshift/openshift-ansible/pull/4641/files

jcantrill:
aos-ci-test

wozniakjan:
> I have some concerns about readiness probes in general for Elasticsearch. How was this tested and shown to work properly? What was the methodology?

- discussed here https://github.com/fabric8io/elasticsearch-cloud-kubernetes/issues/90, 
- here https://trello.com/c/IxWno3R8/537-3-elasticsearch-label-discovery-epic-stabilitylogging 
- and QE test coverage here https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitems?query=subteam.KEY:Intservice%20AND%20trello:IxWno3R8 (test cases will be deleted later by QE)

portante:
@wozniakjan, are the trello and polarion links public or private to Red Hat?

wozniakjan:
@portante, I think they are both public, our trello board is available for public to read and the link to the polarion test cases was posted there as well

portante:
I cannot see the Polarion discussion publicly, at least from my Red Hat account.

I also don't see a concise description of the testing.  I was hoping for something like:

 * 1 pod: quorum 1
   * Verify it boots and comes up cleanly
 * 2 pods: quorum 2
   * Verify one pod starts cleanly and waits for other cluster member, while allowing reads of existing data
   * Verify second pod starts and readily forms cluster
   * Verify first pod scaled down does not cause readiness probe is second to fail
 * 3 pods: quorum 2
   * etc.

wozniakjan:
Polarion has similar description of such test cases.

> 2 pods: quorum 2

I believe in this case, pods will be pronounced ready simultaneously. From my testing, what was happening was that ES returns 503 until the "quorum" number of nodes has been successfully discovered. So if you kill one node and don't scale down quorum count, in order to prevent split brain, both nodes return 503 and are unreachable (readiness probe returns 1 and the service prevents reaching any node).

> 3 pods: quorum 2

This test case gets more interesting to observe. Messing around with only one node still allows rest of the cluster to be operational. The awry node is pronounced non-ready (secluded from the service) and remaining two nodes serve the traffic as they can still reach quorum

Also in order to perform these tests, `openshift_logging_es_number_of_replicas` has to be increased from default 0. Because if you kill an ES node that contained shard without any replica, cluster can never become operational without manual resolution (either delete affected index or recover from backup).

wozniakjan:
aos-ci-test

jcantrill:
[merge][severity:blocker]

PR #5213, opened 2017-08-25
fix #5206.  Default ES cpu limit

### LINKED ISSUE #5206: logging es-ops install broken - no cpu limit
@ewolinetz @jcantrill tl;dr logging CI is broken

#### Description

doing a logging install with the ops cluster - get this error when trying to create the es-ops dc:
```
TASK [openshift_logging_elasticsearch : Set ES dc] *****************************
task path: /usr/share/ansible/openshift-ansible/roles/openshift_logging_elasticsearch/tasks/main.yaml:287
    "changed": false,
    "failed": true,
    "generated_timestamp": "2017-08-24 17:54:39.678096",
    "msg": {
        "cmd": "/bin/oc create -f /tmp/openshift-logging-ansible-q5PRfK/templates/logging-es-dc.yml -n logging",
        "results": {},
        "returncode": 1,
        "stderr": "Error from server (BadRequest): error when creating \"/tmp/openshift-logging-ansible-q5PRfK/templates/logging-es-dc.yml\": DeploymentConfig in version \"v1\" cannot be handled as a DeploymentConfig: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'\n",
        "stdout": ""
    }
}
```
Of course, the error message is almost totally unhelpful 👎 . . .
Fortunately, I was able to tell by visual inspection of `tmp/openshift-logging-ansible-q5PRfK/templates/logging-es-dc.yml`:
```
          resources:
            limits:
              memory: "8Gi"
              cpu: ""
```
This commit broke it: https://github.com/openshift/openshift-ansible/commit/6e498385ab56a1ec3dab84972a160c4bc200919b

The reason is that the default value for `openshift_logging_es_ops_cpu_limit` is `null`: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/defaults/main.yml#L128

which I guess translates into a string value of `""` in the template.

The non-ops case is broken too, but it works because it uses the default value:

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging/tasks/install_logging.yaml#L88

Since it doesn't specify `openshift_logging_elasticsearch_cpu_limit`, it uses the default value : https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/tasks/main.yaml#L281

Which is set here: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/defaults/main.yml#L9

Is the setting of `null` for the default value intentional?  Does that mean "omit the cpu limit"?  If so, it would be better to just omit the default, and in the es.j2 template check if the value is defined.  Or does  null mean "use the default value from openshift_logging_elasticsearch_cpu_limit"?

If it is not intentional, then it should be explicitly set to e.g. `1000m` and the value should be always passed in to the openshift_logging_elasticsearch role in openshift_logging/tasks/install_logging.yaml for both the non-ops and the ops case.

Provide a brief description of your issue here. For example:

> On a multi master install, if the first master goes down we can no
> longer scaleup the cluster with new nodes or masters.

##### Version

Please put the following version information in the code block
indicated below.

* Your ansible version per `ansible --version`

If you're operating from a **git clone**:

* The output of `git describe`

If you're running from playbooks installed via RPM or
`atomic-openshift-utils`

* The output of `rpm -q atomic-openshift-utils openshift-ansible`

Place the output between the code block below:

```
VERSION INFORMATION HERE PLEASE
```

##### Steps To Reproduce
1. [step 1]
2. [step 2]

##### Expected Results
Describe what you expected to happen.

```
Example command and output or error messages
```

##### Observed Results
Describe what is actually happening.

```
Example command and output or error messages
```

For long output or logs, consider using a [gist](https://gist.github.com/)

##### Additional Information

Provide any additional information which may help us diagnose the
issue.

* Your operating system and version, ie: RHEL 7.2, Fedora 23 (`$ cat /etc/redhat-release`)
* Your inventory file (especially any non-standard configuration parameters)
* Sample code, etc

```
EXTRA INFORMATION GOES HERE
```

jcantrill:
(cherry picked from commit cd0d6e5c3fd7cded8c70e210f1e7196a7b0a37d9)

backport of #5212

### DIFF
--- a/roles/openshift_logging/defaults/main.yml
+++ b/roles/openshift_logging/defaults/main.yml
@@ -84,7 +84,7 @@ openshift_logging_es_ca: /etc/fluent/keys/ca
 openshift_logging_es_client_cert: /etc/fluent/keys/cert
 openshift_logging_es_client_key: /etc/fluent/keys/key
 openshift_logging_es_cluster_size: "{{ openshift_hosted_logging_elasticsearch_cluster_size | default(1) }}"
-openshift_logging_es_cpu_limit: null
+openshift_logging_es_cpu_limit: 1000m
 # the logging appenders for the root loggers to write ES logs. Valid values: 'file', 'console'
 openshift_logging_es_log_appenders: ['file']
 openshift_logging_es_memory_limit: "{{ openshift_hosted_logging_elasticsearch_instance_ram | default('8Gi') }}"
@@ -125,7 +125,7 @@ openshift_logging_es_ops_ca: /etc/fluent/keys/ca
 openshift_logging_es_ops_client_cert: /etc/fluent/keys/cert
 openshift_logging_es_ops_client_key: /etc/fluent/keys/key
 openshift_logging_es_ops_cluster_size: "{{ openshift_hosted_logging_elasticsearch_ops_cluster_size | default(1) }}"
-openshift_logging_es_ops_cpu_limit: null
+openshift_logging_es_ops_cpu_limit: 1000m
 openshift_logging_es_ops_memory_limit: "{{ openshift_hosted_logging_elasticsearch_ops_instance_ram | default('8Gi') }}"
 openshift_logging_es_ops_pv_selector: "{{ openshift_hosted_loggingops_storage_labels | default('') }}"
 openshift_logging_es_ops_pvc_dynamic: "{{ openshift_hosted_logging_elasticsearch_ops_pvc_dynamic | default(False) }}"


jcantrill:
aos-ci-test

ewolinetz:
aos-ci-test

jcantrill:
[merge]

sdodson:
In the interest of unblocking other jobs I'm going ot merge this manually.

PR #5215, opened 2017-08-25
[3.6] Remove Jinja delimiters from when conditions

mtnbikenc:


playbooks/common/openshift-cluster/evaluate_groups.yml
```diff
@@ -8,7 +8,7 @@
   - name: Evaluate groups - g_etcd_hosts or g_new_etcd_hosts required
     fail:
       msg: This playbook requires g_etcd_hosts or g_new_etcd_hosts to be set
-    when: "{{ g_etcd_hosts is not defined and g_new_etcd_hosts is not defined}}"
+    when: g_etcd_hosts is not defined or g_new_etcd_hosts is not defined
```
sdodson: Seems like there's a logic change here?

mtnbikenc: This was intentional because the failure message said `or`, and it made sense to be `or` instead of `and`.  I can flip it back if this difference is intentional.  Either the msg needs to be updated or the when condition.

mtnbikenc:
aos-ci-test

playbooks/common/openshift-cluster/evaluate_groups.yml
sdodson: I think the message is equivalent to the logic check. If A or B needs to be defined then the check would be to fail if A and B are undefined, right? Just realized i merged other branches ahead of this one so we'll want to align them.

mtnbikenc: If the logic should actually be `and`, we should make this consistent with the next two tasks as well. 
```yaml
   - name: Evaluate groups - g_master_hosts or g_new_master_hosts required
      fail:
       msg: This playbook requires g_master_hosts or g_new_master_hosts to be set
     when: g_master_hosts is not defined or g_new_master_hosts is not defined
 
   - name: Evaluate groups - g_node_hosts or g_new_node_hosts required
     fail:
       msg: This playbook requires g_node_hosts or g_new_node_hosts to be set
     when: g_node_hosts is not defined or g_new_node_hosts is not defined
```

sdodson: @abutcher can you vet these conditions? for new master, new node, new etcd?

michaelgugino: It reads "I need a or b, if you have neither, fail."  I think the current or in when statement is probably incorrect.

mtnbikenc: I've removed the conditional change from this PR and will open a new PR to address all related tasks.

mtnbikenc:
aos-ci-test

mtnbikenc:
[merge]

mtnbikenc:
@michaelgugino With the updates here can I get a :+1: ?

mtnbikenc:
Test was aborted
re[merge]

brenton:
Seeing if I have [merge] powers.

